PR #367, opened 2015-11-16
Reduce ambiguity in select query (#354)

### LINKED ISSUE #354: Reducing ambiguity in query parameters
Say I have a post object which looks like this:

```json
{
  "id": 5,
  "headline": "Hello, world!",
  "author": 2,
  "editor": 7
}
```

Where both `author` and `editor ` link to a row in the `person` table.

The current `select` parameter does not allow us to expand both (one would try something like: `?select=person(*)`). How should we resolve this issue?

I propose we allow for the following: `?select=author(*),editor(*)` and that expands the `author` and `editor` fields respectively. Thus above would become:

```json
{
  "id": 5,
  "headline": "Hello, world!",
  "author": {
    "id": 2,
    "name": "Jen"
  },
  "editor": {
    "id": 7,
    "name": "Sam"
  }
}
```

One scary part of this, however, is that in one request `author` and `editor` return a number, and in the other `author` and `editor` return objects (personally, I don't see this as too big an issue, but it must be considered nonetheless). One solution to this problem is using JSON-LD (#343) where `author` and `editor` always return an object with an `@id` field.

calebmer:
As I write this, the code in this PR does not actually provide a solution for #354. Rather it is a refactor which is meant to facilitate the implementation of a solution to #354. I PR this first half now to start the discussion on the implementation.

The major refactor change is that all of the types involving the `Query` type were rather ambiguous in its location specifications. Either it only had the table name or the column name and `VForeignKey` was really weird. It had the schema and table name for the `Filter` type it was an object of.

All of these types now use a modified `QualifiedIdentifier` type which is now just an alias for `[Text]`. The new `QualifiedIdentifier` is meant to be a path to whatever bit of data the code is looking for.

One strange part is in order to keep the Parser.hs code "pure" (it shouldn't require knowledge of the database) the head of `QualifiedIdentifier` will be an empty string to represent that it is unqualified. Later they will be qualified [here](https://github.com/calebmer/postgrest/blob/df7ac5139b6d4ff53c11c2e0c0952eebb3caa7ec/src/PostgREST/App.hs#L413-L435). I didn't want to use a data type because it was so handy to be able to use all of the list functions on `QualifiedIdentifier`s.

Maybe give this initial refactor a look (or wait until the solution for #354 is implemented to review). Also, would you rather merge this independently of our solution? Finally, definitely give your opinion on how we fix #354 in the issue.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -25,7 +25,7 @@ executable postgrest
   if flag(ci)
     ghc-options:       -Wall -W -Werror
   else
-    ghc-options:       -Wall -W -O2
+    ghc-options:       -Wall -W
 
   main-is:             PostgREST/Main.hs
   default-extensions:  OverloadedStrings, ScopedTypeVariables, QuasiQuotes
@@ -78,7 +78,7 @@ library
   if flag(ci)
     ghc-options:       -Wall -W -Werror
   else
-    ghc-options:       -Wall -W -O2
+    ghc-options:       -Wall -W
 
   default-language:    Haskell2010
   default-extensions:  OverloadedStrings, ScopedTypeVariables, QuasiQuotes
@@ -146,7 +146,7 @@ Test-Suite spec
   if flag(ci)
     ghc-options:       -Wall -W -Werror
   else
-    ghc-options:       -Wall -W -O2
+    ghc-options:       -Wall -W
   Main-Is:             Main.hs
   Other-Modules:       Feature.AuthSpec
                      , Feature.CorsSpec
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -292,7 +292,7 @@ convertJson v = (,) <$> (header <$> normalized) <*> (vals <$> normalized)
         a@(Array _) -> Right a
         _ -> Left invalidMsg
 
-augumentRequestWithJoin :: Schema ->  [Relation] ->  ApiRequest -> Either Text ApiRequest
+augumentRequestWithJoin :: Schema -> [Relation] -> ApiRequest -> Either Text ApiRequest
 augumentRequestWithJoin schema allRels request =
   (first formatRelationError . addRelations schema allRels Nothing) request
   >>= addJoinConditions schema
@@ -407,18 +407,19 @@ parseRequest schema allRels rootTableName httpRequest reqBody =
               else Left "Expecting a sigle CSV line with header or a JSON object"
 
 qualifyRequest :: Schema -> ApiRequest -> ApiRequest
-qualifyRequest schema (Node (q,x@(name,_)) forest) =
+qualifyRequest schema (Node (q,x@(name,rel)) forest) =
   case q of
     Select sel frm whr ord -> Node (Select (quSelectItems sel) (quFrom frm) (quWhere whr) ord, x) updatedForest
     Insert int fds vls     -> Node (Insert (quTable int) (quFields fds) vls, x) updatedForest
     Delete frm whr         -> Node (Delete (quFrom frm) (quWhere whr), x) updatedForest
     Update int hsh whr     -> Node (Update (quTable int) (quHash hsh) (quWhere whr), x) updatedForest
   where
+    tableN = fromMaybe name (tableName . relTable <$> rel)
     updatedForest = map (qualifyRequest schema) forest
 
-    quTable (UnqualifiedIdentifier uqi) = QualifiedIdentifier sch uqi Nothing where sch = if name == sourceSubqueryName then Nothing else Just schema
+    quTable (UnqualifiedIdentifier uqi) = QualifiedIdentifier sch uqi Nothing where sch = if tableN == sourceSubqueryName then Nothing else Just schema
     quTable qi = qi
-    quColumn (UnqualifiedIdentifier uqi) = QualifiedIdentifier sch name (Just uqi) where sch = if name == sourceSubqueryName then Nothing else Just schema
+    quColumn (UnqualifiedIdentifier uqi) = QualifiedIdentifier sch tableN (Just uqi) where sch = if tableN == sourceSubqueryName then Nothing else Just schema
     quColumn qi = qi
 
     quField (qi,jp) = (quColumn qi,jp)
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -140,7 +140,7 @@ app dbStructure conf reqBody req =
             _         -> return $ responseLBS status404 [] ""
 
     (["rpc", proc], "POST") -> do
-      let qi = QualifiedIdentifier schema (cs proc)
+      let qi = [schema, cs proc]
       exists <- doesProcExist schema proc
       if exists
         then do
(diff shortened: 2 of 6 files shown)


ruslantalpa:
There this type
```
type Schema = Text
-- could also be type TableName = Text
```

So what you have her is
```
a = ["public"::Schema, "comments"::TableName] :: QualifiedIdentifier
```
which would compile but is not a good haskel style (having items of different type in a list)

calebmer:
They are not different types. The `type` construct is just an alias for whatever `data` form it represents. I don't see this as an issue at all.

If it is a problem, there are plenty of places in our code where we do `s :: Schema <> t ::TableName` that need to be fixed.

ruslantalpa:
I know the difference between type and data, the point is you have a list that contains items that mean different thing although the underlying data type is the same, like [Day, Month, Year, Timestamp, Distance, Length] things like that should not be in a single list, that what Record structure is for, the original definition of QualifiedIdentifier is correct (but probably will be phased out now that there are things like Table/Column).

calebmer:
The benefit of doing it this way is that nothing is required, schema doesn't have to exist, column doesn't have to exist etc. But I get your concerns, if someone seconds them I will definitely transition back to a record structure.

ruslantalpa:
The things that "don't have to exist" are described with "Maybe" not the length of the list

calebmer:
Ok, `QualifiedIdentifier` now uses a record syntax with `Maybe`s. I think your right, while I might use an array in JavaScript, it makes much more sense in Haskell to use a data type.

ruslantalpa:
This PR as you mentioned is driven by the desire to implement #354 the way you suggested, direction i think it's the wrong one so i'll wait on the direction for #354 before considering what this PR brings to the table.
Another thing to consider, a refactor is meant to make things easier/clearer/smaller, how does adding 40 lines to the codebase accomplish that (without adding any new feature)?
Sorry for being negative nancy here but this is my opinion.
I think you have your specific use case (a few tables/views/fks named in a specific way) and pushing out code fast to fit your project (deadline?) without carefully considering if it a good general interface or implementation.
I am not saying i am always right (or my solutions are better) but you do have to admit you are bending this project to fit your needs (which is a good thing/motivator/ valid thing to want) but you are rushing things a bit (which is also given away a bit by the fact that you do not want a feature freeze).
The point of the feature freeze is to clean up the code, make it shorter so that when new features are added they are added on a solid base and we don't keep piling up code just to add the next feature. When code is solid/short the desired feature might be 10 times easier to implement as opposed to jsut adding it now.

I would suggest you have your private fork/branch that solves the author/editor problem for you so that the rest of the project can move along. And you use that "interface" for a while and when this capability is added to the "master" (be that using your interface or not, your code or not) you just rewrite the way you do your requests. for example i saw you mentioned you currently do something like
```
author: null
editor: null
```
if my interface is implemented, all you have to do really is change to 
```
"author:persons[author_id]":null
"editor:persons[editor_id]":null
```
and the other code does not have to change
Sorry for the blunt talk

calebmer:
Going to go down each thing and then sum up my thoughts in a conclusion at the end.

1. When you say a refactor is meant to make things easier/clearer/smaller, the "/" does not translate to "and", rather it translates to "or". The current refactors makes things easier and clearer. With the implementation on master (as I said in the OP) it is very hard to reason about what schemas are being added to which tables and what columns go to which from statements. This refactor removes that unclearness *entirely* by switching everything to a `QualifiedIdentifier` type. Where are those 40 lines coming from, [here](https://github.com/begriffs/postgrest/pull/367/files#diff-0721c0195214a5855036d6df3d06e076R415) where we switch from `UnqualifiedIdentifier` to `QualifiedIdentifier` (note how about 50% of the lines are aliases for other functions so that the code that actually matters is *clearer* and *easier*). The rest comes from expansions like [this](https://github.com/begriffs/postgrest/pull/367/files#diff-1176a3ab0722bbf445155bf53e2b2ad7R89) striving to make the code more readable. Line number count is generally a very bad indicator for code quality.
- I'm developing against a real world scenario, real problems that and I have provided the best fix *for* those problems. I do not think my situation is unique or specific, as the very first thing I did after installing the new PostgREST version with the `select` parameter was try to use `author_id(*)`. This was my intuitive response to the new technology, which not including support for meant also introducing some ambiguity which I now wish to clear up. We don't get a lot of developer feedback on this, [but when we do](https://github.com/begriffs/postgrest/issues/343#issuecomment-154093899) that feedback is of confusion. Why do I think this my proposal is the best? Because over all other reasons it is *intuitive* it is the *expected* behavior. @ruslantalpa, You can argue that, but remember you developed the technology. It's very hard to then detach yourself and see it from a new devs eyes.
- You say I haven't carefully considered my interface, I take that very personally. What is #354, what about all the times I solicited feedback in the gitter chat? I wanted to test my interface against the ideas of others, and coming away from all of that I am now more confident then ever it is the best solution.
- *"you do have to admit you are bending this project to fit your needs,"* never said I wasn't, but you say it best: *"which is a good thing/motivator/valid thing to want"*. I am against a feature freeze on an open source project not specifically for this project, but as an idea itself. Open source projects are driven by people contributing time which they (often) are not payed for. Contributions are motivated by features, not by an inherent want to improve code quality. Therefore I am fundamentally against the idea of removing that primary motivation. You can refactor without pausing innovation.
- I agree that good refactored code is easier to build implementations off of, what do you think the first couple commits of this PR are? It's a refactor to make the feature implementation easier.
- As for the private fork/branch, is not a PR fundamentally just a branch? I believe the solution in this PR is the best solution, and other people can obtain value from these contributions, therefore I will keep this PR open, even if I'm the only one using the custom build it produces.

To conclude, this PR is designed to solve a specific problem in an intuitive way; and I believe (after much debate) that this way is the best for all PostgREST devs.

I believe the best resolution to this question is to do user research.

* * *

I also want to apologize to everyone for all of the hostility. It can be easy some times to jump into attack mode whenever an idea is challenged. To be completely honest, I do like an argumentative culture as a good idea needs to be rejected just as much as a bad idea, however I do realize that can be a big turnoff for a lot of people.

At the end of the day I want the same thing from PostgREST as everyone else, a definitive answer to the backend question. And I sincerely believe that PostgREST has the opportunity to become really big, as in GraphQL or Falcor big. Not because it does things in an interesting new way, but rather because it is stable, dependable, and has a great intuitive design (ok, the interesting new way might actually be the fact it's written in Haskell of all things 😊).

calebmer:
Ok, just finished implementing select using column names. I tried to fix the nested comments problem, but there are deeper recursion problems that need to be resolved there which will require another big refactor. I know how @ruslantalpa feels about this, but I really want to know how everyone else feels, so @begriffs and @diogob.

Currently the code does not include aliasing and does not support interchanging `()` with `[]` or `{}`. But I can definitely add those features if you all think it would help alleviate some potential issues.

calebmer:
Here's what I'm going to do, I still support this solution but it seems the project is not ready for it. With #377 I want to get the `QualifiedIdentifier` refactor merged, so I'll create a new PR with just the reactors so we can discuss them indecently.

calebmer:
Wanted to give an update on this considering there have been some issues/people in the chat which have mentioned this solution (eg #413).

I lost motivation for maintaining the `QualifiedIdentifier` refactor with the `0.3.x` upgrade, so I'm looking for a better way to implement this with the new code base. Once I find time/motivation/a good implementation I'll open a new PR and close this one. For now this can continue to collect opinions.

To address the number/object issue I have two solutions: 1) if we adopt JSON Schema it's super easy to show both data formats 2) duck type `_id` checks, so an `author{*}` select would resolve to `author_id{*}` (but still output `author` in the JSON). I think both are valid solutions and we shouldn't let this block the intuitive extension of the `select` parameter.

@ruslantalpa

calebmer:
Instead of force pushing I'm going to close this pull request and open another to keep the git history.

PR #361, opened 2015-11-13
Single row routes

calebmer:
Often I find myself performing requests to PostgREST to select a single row which look like this: `/people?id=eq.5`. There's a couple of issues with this approach, first is that I need to know the primary key(s) name (not a huge issue, but one nonetheless) and I need to extract the single object from the array, which is not always easy depending on the technology being used on the client side. Thus having a row route is useful.

The implementation of a router function is also very powerful in that it allows for future nested routes in a clean way. It also cleans up App.hs, maybe a little? (no not really 😊)

The exact feature details are as follows: to deal with compound keys, PostgREST expects a tuple-like syntax. `/people?id=eq.5` can now be replaced by: `/people/5` or `/people/(5)`. The parentheses are optional only when the row primary key is not a compound one. For a row with a compound key, `/table?k1=eq.2&k2=eq.6` becomes `/table/(2,6)`.

One other change I made is that in addition to `/rpc/procedure` a stored procedure can be called using an ampersand, so `/@procedure`. If no one likes it, it can be taken out. To me it makes a lot more sense then the obfuscated `/rpc` sub route.

This PR contains my other PR #353 (that's why you can see all of @ruslantalpa's comments, didn't see that coming…pretty cool that's detected though) and can be rebased once that is merged.

PR #368, opened 2015-11-16
Singular route request

calebmer:
Continuation of #361 against master.

Removed the single row routes, kept the new router (which we should talk about, do we want it at all?) and implemented @begriffs single object prefer header proposal.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -19,7 +19,7 @@ import qualified Data.HashMap.Strict       as HM
 import           Data.List                 (find, sortBy, delete, transpose)
 import           Data.Maybe                (fromMaybe, fromJust, isJust, isNothing, mapMaybe)
 import           Data.Ord                  (comparing)
-import           Data.Ranged.Ranges        (emptyRange)
+import           Data.Ranged.Ranges        (emptyRange, singletonRange)
 import           Data.String.Conversions   (cs)
 import           Data.Text                 (Text, replace, strip)
 import qualified Data.Text                 as T
@@ -37,7 +37,7 @@ import           Network.Wai
 import           Network.Wai.Parse         (parseHttpAccept)
 
 import           Data.Aeson
-import           Data.Aeson.Types (emptyArray)
+import           Data.Aeson.Types          (emptyArray)
 import           Data.Monoid
 import qualified Data.Vector               as V
 import qualified Hasql                     as H
@@ -65,7 +65,8 @@ import           PostgREST.QueryBuilder ( asJson
                                         )
 import           PostgREST.RangeQuery
 import           PostgREST.Types
-import           PostgREST.Auth (tokenJWT)
+import           PostgREST.Auth            (tokenJWT)
+import           PostgREST.Error           (errResponse)
 import           PostgREST.Router
 
 import           Prelude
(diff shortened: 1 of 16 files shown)


diogob:
I like the idea of extracting the router to a separate module. Specially since the App module is so big right now. But I'd like to see how the code would look like if we move the whole ```case verb of``` to the Router.

My suggestion is that we create another datatype called Route, which would have and endpoint and a verb.
Then in the App we can pattern match only against the Route instead of having nested cases.

Another suggestion is to add haddock strings for new modules describing their planned functionality, this would help discussing the code and reasoning about what each module should encapsulate.

src/PostgREST/App.hs
```diff
@@ -64,138 +65,154 @@ import           PostgREST.QueryBuilder ( asJson
...
-    ([table], "OPTIONS") -> do
-      let cols = filter (filterCol schema table) allCols
-          pkeys = map pkName $ filter (filterPk schema table) allPrKeys
-          body = encode (TableOptions cols pkeys)
-      return $ responseLBS status200 [jsonH, allOrigins] $ cs body
-
-    (_, _) ->
-      return $ responseLBS status404 [] ""
+  case route RootE dbStructure schema path of
+    Just endpoint ->
```
begriffs: Could decrease the indentation by doing something like

```haskell
  case route RootE dbStructure schema path of
    Nothing -> notFound
    Just RootE ->
      -- ...
    Just (TableE table) ->
      -- ...
```

src/PostgREST/App.hs
```diff
@@ -363,8 +380,8 @@ instance ToJSON TableOptions where
       "columns" .= tblOptcolumns t
     , "pkey"   .= tblOptpkey t ]
 
-parseRequest :: Schema -> [Relation] -> NodeName -> Request -> BL.ByteString -> Either Text (Text, Text, Bool)
-parseRequest schema allRels rootTableName httpRequest reqBody =
+parseRequest :: Schema -> [Relation] -> NodeName -> Request -> [Filter] -> BL.ByteString -> Either Text (Text, Text, Bool)
+parseRequest schema allRels rootTableName httpRequest extraFilters reqBody =
```
ruslantalpa: where are extraFilters used? i think it's a leftover form single route needs

ruslantalpa:
This PR again introduces lots of other things (extrafilters, rewrite of createStatement, dbFindTable, order by, pgErrResponse) which have noting to do with route functionality.
How about closing this PR and adding the router functionality in a clean branch (baed on current master)

ruslantalpa:
Writing of the "Router" data type should be easier if it's based on the refactoring from #371

calebmer:
@ruslantalpa `extraFilters` is leftover, `createStatement` rewrite adds no code (except for some `isSimgle` stuff) `dbFindTable` could be moved to Router.hs, and `pgErrRespinse` is a needed refactor of non-uniform error handling (some errors were just plaintext strings).

The rest of the PR has a rather low footprint, but if we want to merge #371 first that's fine. I'll just do a rebase.

begriffs:
Other refactor merged. Ready for a rebase of this one to add plurality selection.

calebmer:
Do we want Router.hs? I could rebase or split this into two commits. One for the error handling refactor and another which adds plurality singular.

PierreR:
Is there any reason not to use Servant for `routing` ?

begriffs:
I like dealing with small focused PRs best. Splitting it would be most convenient.

calebmer:
@diogob I like that idea too, however it seems for now we are going to pass on implementing a router until we need one.

Closing this PR in favor of two smaller ones: #374 and #375.

PR #369, opened 2015-11-16
Clean up things like T.Text T.intercalate in QueryBuilder (1)

ruslantalpa:

PR #370, opened 2015-11-16
Add type synonyms to to beter explain things that used to be Text (2)

ruslantalpa:
This should help a bit with understanding what's going on in App.hs
No code changed, just some type synonyms to better explain the parameters.
(should be merged after #369)


begriffs:
That's very helpful, it reads a lot better.

PR #371, opened 2015-11-16
App.hs refactor (3)

ruslantalpa:
This PR is a bit more complicated.
Please review the other two i submitted ( #369 then #370 ) before reviewing this one so that the "diff" does not contain the code from the previous ones and it's easier to look at.

This PR should help with introducing the "Router" if it desired.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -68,20 +68,30 @@ import           PostgREST.Auth (tokenJWT)
 
 import           Prelude
 
-app :: DbStructure -> AppConfig -> BL.ByteString -> Request -> H.Tx P.Postgres s Response
+app :: DbStructure -> AppConfig -> RequestBody -> Request -> H.Tx P.Postgres s Response
 app dbStructure conf reqBody req =
   case (path, verb) of
+    ([table], "OPTIONS") -> do
+      let cols = filter (filterCol schema table) $ dbColumns dbStructure
+          pkeys = map pkName $ filter (filterPk schema table) allPrKeys
+          body = encode (TableOptions cols pkeys)
+          filterCol :: Schema -> TableName -> Column -> Bool
+          filterCol sc tb (Column{colTable=Table{tableSchema=s, tableName=t}}) = s==sc && t==tb
+          filterCol _ _ _ =  False
 
-    ([table], "GET") ->
-      if range == Just emptyRange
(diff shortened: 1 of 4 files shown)


ruslantalpa:
I looked at the generated diff here https://github.com/ruslantalpa/postgrest/commit/0c3545fa09bf6c297dc9a2dec43370503ec9ca3d and it's not nice (more confusion). I suggest looking directly at the [source](https://github.com/ruslantalpa/postgrest/blob/app_refactor/src/PostgREST/App.hs)  (after looking a bit at the diff to get a sense what parts changed)

calebmer:
Looks good. I think after merging this with #368 we remove some of the [intuition](https://github.com/begriffs/postgrest/pull/371/files#diff-0721c0195214a5855036d6df3d06e076R87) disconnect and end up making the `app` function much better.

It might be easier if I just cherry pick this commit into #368, would you like/mind that?

ruslantalpa:
I am not sure i understand the  order of the things you suggest but my view is this:
1.  Merge (1,2,3) with improvements if suggested
2. You create a fresh PR starting from master (without the previous commit attempts) that strictly contains the implementation of Router data type and 1-2 additional helper functions and see if there is improvement in readability of the "app" code added by the Router idea (no refactoring of other parts like error functions and such so that we can clearly see the impact and compare old style to new style)

ruslantalpa:
One thing i did wrong in this PR is combine the refactoring of parseRequest + createStatemnet with the rewrite of the app function, the first part is a step in the right direction, the last part  (the app function, removing explicit GET and nesting of the core logic under ([table], _)) i am not 100% sure so if someone thinks that can be done in a better way i'll close this particular PR and open a separate one with just the parseRequest and createStatement

calebmer:
What I mean by ordering is that #368 and just the 3rd commit here might be better as one PR vs two.

ruslantalpa:
The 3rd commit is about single routes in pg, you know my opinion there. The second commit (router) has value although i think it's better to implement it fresh rather then trying to rebase/merge

calebmer:
Sorry, I meant the 3rd commit in this PR 😊

begriffs:
@ruslantalpa let's break this into pieces like you say. This PR includes commits that were already merged so a clean PR would be good for that reason too (either that or rebasing your PR branch on new master).

PR #373, opened 2015-11-16
Schema templates

begriffs:
Example sql for to make a blog.


begriffs:
I'll add a test suite in another PR.

PR #374, opened 2015-11-16
Small error refactor

calebmer:
Adds a helper function for us to easily make uniform json errors.


begriffs:
This is a nice one! Why didn't we do it earlier... :)

PR #375, opened 2015-11-16
Implement single route selection

calebmer:
Use `Prefer: plurality=singular` to select single objects.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -19,7 +19,7 @@ import qualified Data.HashMap.Strict       as HM
 import           Data.List                 (find, sortBy, delete, transpose)
 import           Data.Maybe                (fromMaybe, fromJust, isJust, isNothing, mapMaybe)
 import           Data.Ord                  (comparing)
-import           Data.Ranged.Ranges        (emptyRange)
+import           Data.Ranged.Ranges        (emptyRange, singletonRange)
 import           Data.String.Conversions   (cs)
 import           Data.Text                 (Text, replace, strip)
 import           Data.Tree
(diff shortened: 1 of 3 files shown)


src/PostgREST/App.hs
```diff
@@ -404,15 +410,17 @@ parseRequest schema allRels rootTableName httpRequest reqBody =
...
     wrapQuery selectQuery [
       if countTable then countAllF else countNoneF,
       countF,
       "null", -- location header can not be calucalted
-      if asCsv then asCsvF else asJsonF
-    ] selectStarF range
+      if asCsv
+        then asCsvF
+        else if isSingle then asJsonSingleF else asJsonF
+    ] selectStarF (if isNothing range && isSingle then Just $ singletonRange 0 else range)
```
ruslantalpa: Not sure this line is correct. It reads:
when we have isSingle and the user did not specify a range then do "limit 1", else just use the range parameter
What about the case when the user sends isSingle=true and also specifies the range. It will probably work but maybe the line should be
```
( if isSingle then Just $ singletonRange 0 else range )
```
singletonRange 0 basically does a LIMIT 1, right ?

ruslantalpa:
:+1:

src/PostgREST/App.hs
calebmer: Yes, it's just a limit 1. I can have `isSingle` override the range parameter.

ruslantalpa: Just remembered, override is necessary because if the range specifies limit > 1 then asJsonSingleF will fail

begriffs:
This behaves differently than I imagined it -- but your way is interesting. The issue is how a response with multiple rows should be treated when the client sends the prefer header. I can see three ways:

1. (your way) The first row is popped out of the array and returned by itself
2. A 400 error indicating that the client assumes the resource is singular but it is not
3. A single object when there is only one row, or the unchanged array if more rows. In other words the header would change `[{foo}]` to `{foo}` and `[]` to 404 but leave `[{foo},{bar}]` unchanged.

Option 1 is nice because if a client sets the header the server will never surprise them with an array. It's bad because accidentally including the header or using it for an inappropriate route will pretend everything is fine and it could be harder to debug.

Option 2 is nice because it helps clients correct their behavior and assumptions. We can detect if the client is filtering on the primary key and if not then we respond with 400 because the resource is not intrinsically singular. This option has the downside that grabbing the first row as an object might be a convenient thing that a client wants to do even on a plural resource.

Option 3 is what I originally imagined because I think of a Prefer header as saying, "you may not be able to do this server, but I would *prefer* if you did." The downside is that the response could sometimes be an object, sometimes an array, which is more work for clients.

Anyone have opinions one way or another?

ruslantalpa:
I am for option 1. This is what i imagined when you suggested it. The implementation is simple/short (the other ones would require more work/not as nice)
For 2 you can get the surprise that it works for a while then you add a row and sudenly you get a 400
For 3 you get different reaponse shape depending on the num of rows, hard to implement in frontend

As a coclusion, this header will most probably be used in nginx config so it will be transparent to the client

calebmer:
Changelog has been notified 😊

PR #376, opened 2015-11-17
Refactor parseRequest, add buildMutateApiRequest function

ruslantalpa:
With this refactor, now select/mutate paths now have the same stages (http->ast->sql->result) and functions similar in structure. Probably now is a good time for @begriffs start splitting App.hs QueryBuilder.hs into smaller modules that probably mirror the stages above


diogob:
:+1:

PR #377, opened 2015-11-19
Disentangle code

begriffs:
This PR separates various concerns into their own modules. It defines a clean interface between modules. I'd like to use the type system better, get logic out of overkill monads, use understandable names, and document heavily.

This is a work in progress.

### DIFF
--- a/src/PostgREST/RequestIntent.hs
+++ b/src/PostgREST/RequestIntent.hs
@@ -99,7 +99,9 @@ userIntent schema req reqBody =
             (JSON.eitherDecode reqBody)
         Right TextCSV ->
           either (PayloadParseError . cs)
-            (PayloadJSON . csvToJson)
+            (\val -> case ensureUniform (csvToJson val) of
+              Nothing -> PayloadParseError "All lines must have same number of fields"
+              Just json -> PayloadJSON json)
             (CSV.decodeByName reqBody)
         Left accept ->
           PayloadParseError $
@@ -177,11 +179,11 @@ type CsvData = V.Vector (M.HashMap T.Text BL.ByteString)
   The reason for its odd signature is so that it can compose
   directly with CSV.decodeByName
 -}
-csvToJson :: (CSV.Header, CsvData) -> UniformObjects
+csvToJson :: (CSV.Header, CsvData) -> JSON.Array
 csvToJson (_, vals) =
-  UniformObjects $ V.map rowToJsonObj vals
+  V.map rowToJsonObj vals
  where
-  rowToJsonObj =
+  rowToJsonObj = JSON.Object .
     M.map (\str ->
         if str == "NULL"
           then JSON.Null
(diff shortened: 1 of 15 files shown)


src/PostgREST/RequestIntent.hs
```diff
@@ -0,0 +1,104 @@
...
+  -- | If client wants created items echoed back
+  , iPreferRepresentation :: Bool
+  -- | If client wants first row as raw object
+  , iPreferSingular :: Bool
+  }
+
+-- | Examines HTTP request and translates it into user intent.
+userIntent :: Schema -> Request -> RequestBody -> Intent
+userIntent schema req _ =
+  let action = case requestMethod req of
```
calebmer: Any reason why this is a let vs in the where?

calebmer:
This is looking really exciting 😊

src/PostgREST/RequestIntent.hs
begriffs: These let's contain more logic than the where's so by putting them in the function I guess I'm trying to call attention to their logic. Let's do kind of suck for indentation though and they require the ...`in` ending.

src/PostgREST/RequestIntent.hs
```diff
@@ -0,0 +1,131 @@
...
+-- then CSV is converted to a JSON array.
+data Payload = PayloadJSON JSON.Array
+             | PayloadParseError BS.ByteString
+
+-- | Describes what the user wants to do. This data type is a
+-- translation of the raw elements of an HTTP request into domain
+-- specific language.  There is no guarantee that the intent is
+-- sensible, it is up to a later stage of processing to determine
+-- if it is an action we are able to perform.
+data Intent = Intent {
```
ruslantalpa: Should probably have something like
iSelect :: String
iFilters :: [(String, String)]
iOrder :: Maybe String

begriffs:
I think my goal for this PR is to do one big thing: finish the Intent module and make App.hs use it for parsing requests. Continued refactors can go into future PRs.

ruslantalpa:
So far so good, waiting for this module to touch App code :)

begriffs:
The good news thus far: memory leak is fixed for bulk insertion of JSON/CSV!

screen shot 2015-11-20 at 12 59 59 pm

It has some quirks though, like if I insert [{good}, {good2}, {bad}, {good3}] then only good, good2, good3 get added to the db and we get no error from the bad one. Relatedly it won't choose a default value for omitted columns, so inserting to a table with auto-incrementing primary key fails.

I commented out the PATCH verb stuff entirely so that I could focus on getting POST to work.

begriffs:
Posting big arrays is very slow, like half a minute for 100,000 objects.

I was worried it is caused by my checking that all keys in the objects match... but that doesn't appear to be the slow part. When I comment out the logic in `ensureUniform` the post is still slow. [confusing profiler output](https://gist.github.com/begriffs/0d0b67d3f8d82e5c5697) Could it be the parsing itself? Was it always this slow? We used to have the memory problem which prevented benchmarking at this size of data.

begriffs:
I confirmed that posting 100,000 CSV objects is very fast, despite being converted into an Aeson Value for submission to postgresql. It's gotta be the json parsing. I'll continue working on "the trick" of selecting the right values from `json_populate_recordset` to get the insert test passing for columns with default values.

PR #378, opened 2015-11-19
Remove range check

calebmer:
Forgot to throw this suggestion in from @ruslantalpa into #375.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -431,7 +431,7 @@ createReadStatement selectQuery range isSingle countTable asCsv =
       if asCsv
         then asCsvF
         else if isSingle then asJsonSingleF else asJsonF
-    ] selectStarF (if isNothing range && isSingle then Just $ singletonRange 0 else range)
+    ] selectStarF (if isSingle then Just $ singletonRange 0 else range)
   ) V.empty True
 
 createWriteStatement :: SqlQuery -> SqlQuery -> Bool -> Bool -> [Text] -> Bool -> B.Stmt P.Postgres


begriffs:
Is there a feature test that illustrates what this does?

calebmer:
This isn't a feature, just a small nuance (which should technically never matter) in how the code works. To quote @ruslantalpa:

> Just remembered, override is necessary because if the range specifies limit > 1 then asJsonSingleF will fail

The override referred to is this change.

begriffs:
I'm assuming we could construct a request with a range that would cause a response failure, but which would work once this PR is merged.

ruslantalpa:
Exactly, sending prefer singular and range 10 will generate a bad query/response

begriffs:
I wrote this test which passes on master (post refactor PR merge).

```haskell
    it "works in the presence of a range header" $
      let headers = ("Prefer","plurality=singular") :
            (rangeHdrs $ ByteRangeFromTo 0 9) in
      request methodGet "/items" headers ""
          matchBody    = Just [json| {"id":1} |]
        , matchStatus  = 200
        , matchHeaders = []
        }
```
Does this mean we're good?

ruslantalpa:
so you say without merging this, the test passes?
My concern was this line
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/QueryBuilder.hs#L297
Not sure why the test passes though ...

begriffs:
Yeah it passes on master just as master is. Did I screw up the test somehow? Try pulling master and adding the test locally.

PR #379, opened 2015-11-19
Refactor QualifiedIdentifier

calebmer:
Currently, when it comes to `QualifiedIdentifier` the code can get very confusing. Column names are stored in weird places, for there to be no schema the value must be set to `""`, `VForeignKey` held a `QualifiedIdentifier` for the `Field` type which contained it, and there was no way to distinguish between un-qualified values and qualified ones. The parser module does not have access to `DbStructure` so it can't add the schema and/or table name, therefore the new `qualifyRequest` function takes an unqualified identifier and qualifies it in one place vs. the dangerous assumptions being made in multiple places throughout the code base.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -147,7 +147,7 @@ app dbStructure conf reqBody req =
             _         -> return notFound
 
     (["rpc", proc], "POST") -> do
-      let qi = QualifiedIdentifier schema (cs proc)
+      let qi = QualifiedIdentifier (Just schema) (cs proc) Nothing
       exists <- doesProcExist schema proc
       if exists
         then do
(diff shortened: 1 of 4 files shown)


ruslantalpa:
QualifiedIdentifier is a leftover from the old codebase, before types like Table and Column.
Some of the old formatting functions at the end of the chain pgFmt ... were using it so i created QI values to feed to them but i think the direction should be to remove this data type altogether rather then making it's presence even stronger. From a value of type Table you can get schema and table name and from a value of type Column you also can get column name, table, schema, so we should use those

calebmer:
The problem with switching to table and column data types is that `WITH` clause things, like `pg_source` are not a table and should not have a schema.

Also procedures and other database constructs are identifiable with `QualifiedIdentifier`.

ruslantalpa:
Then maybe extend the Table data type to also have constructors like RPC and Source (not 100% if it works). It just seems not right because of 2 corner cases (RPC/Source) to add a data type that touches all the functions on all the levels and the info contained in it is basically the same as the on in Table or Column.

begriffs:
QualifiedIdentifier was originally intended to represent the name of a database object whose existence is not yet certain such as those specified by API routes. Many functions accept arguments for schema and table/proc name so I packaged those names together. I agree that once the object has been discovered in the db it would be better referenced by another data structure such as Table.

Hence we might not want to remove QI, rather we ought to use it in the proper situations. (I'm open to counterarguments of course)

ruslantalpa:
Only part it makes sense is parsers, after that it should be Table/Column ...

begriffs:
Sorry I've let this pull request languish. Trying to work through the mega-refactor. Hoping it will be the last mega-PR for a while.

begriffs:
This no longer merges cleanly due to all the changes on master.

@calebmer could you look at the new code and see if your concern still applies?

I'm going to close this PR as it is now.

PR #380, opened 2015-11-20
Get the refactored code to compile and pass tests

ruslantalpa:
Currently code compiles and 5 test failing.
4 of the failing tests are related to RPC calls (maybe you can figure it out faster)
1 of the failing test is when you send a csv with the number of columns in the header not matching the other rows.
The csv parser should return a parser error in such a case

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -4,20 +4,17 @@
 --module PostgREST.App where
 module PostgREST.App (
   app
-, contentTypeForAccept
 ) where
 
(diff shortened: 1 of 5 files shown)


diogob:
-32 lines! :confetti_ball: :tada:

ruslantalpa:
That was just comments i guess, this pr has to be merged then we look at the lines i #377 that's where the real test is :)

begriffs:
:pray: 

Whew, thank you.

Issue #144, opened 2015-02-13
Support RAML description of API


StefanScott:
Hi -

(This isn't really a question about any problem or issue involving PostgREST - but rather an exploratory question about possible opportunities for using PostgREST along with tools for documentation APIs, and possibly even generating simple CRUD front-ends based on a database schema, leveraging PostgREST.)

Recently I saw a free open-source tool called swagger.io which is used to document REST APIs.

http://en.wikipedia.org/wiki/Swagger_(computer_science)

http://swagger.io/

http://editor.swagger.io/#/

https://github.com/swagger-api/swagger-ui

https://github.com/swagger-api/swagger-spec/

http://petstore.swagger.io/#!/pet/getPetById

http://en.wikipedia.org/wiki/Web_API

It seems that swagger.io could be useful in combination with PostgREST. 

They also apparently have some sort of tool which generates client code in different languages:

https://github.com/swagger-api/swagger-codegen

I'd like to use swagger.io with PostgREST. Currently, I'm unsure how to get started, so I'm still reading the documentation. It seems that the REST API has to expose an endpoint like:

https://postgrest.herokuapp.com/swagger.json

which would then provide discovery and documentation of the REST endpoints of the above Postgres database.

I guess first it would be necessary to create the swagger.json document for a given Postgres database.

Ultimately I think it would be interesting if it would be possibly not only to provide discovery and documentation of a given database, but also to generate a simple admin CRUD interface for the database, in a particular client language. 

It looks like someone has already used swagger to develop generator which produces a simple admin CRUD interface from a database definition:

http://www.toobop.net/gform-admin/

https://github.com/stemey/gform-admin

I think at this point in history, it should be possible to combine certain notions of metaprogramming with web development frameworks to create admin CRUD interface generators based solely on a database schema (ie, a basic but modern web client providing standard layouts such as forms and grids supporting CRUD operations - including simple master-detail layouts, and custom field editors - eg data pickers for date fields, and drop-down menus for foreign-key fields). 

This sort of thing has been available (typically in the form of "wizards") in many native (Windows, Linux, Apple) database application development environments (eg, Microsoft Access :-)

Now, with the power of today's web development frameworks and metaprogrammable languages, this sort of thing should also start becoming available for web-based database development as well.

I would like to learn more about how to combine swagger.io and PostgREST to do this sort of thing - to generate a variety of simple admin CRUD front-ends based on a PostgREST database schema, targeting a variety of front-end technologies, such as:

AngularJS:

[...]

PR #381, opened 2015-11-21
Adds Api module to handle HTTP parsing using servant

diogob:
First, this PR is just to discuss the approach, so it's totally incomplete but I wanted to have a green light before spending any more time.

The idea is to isolate the Api design from it's implementation (as we've been discussing), with servant the whole Api shape gets nicely encapsulated in a Api type.

We can extend servant and add any combinators we like to extract info from the HTTP request. There is an example in which I extract the whole queryString to a list here:
```haskell
instance HasServer api => HasServer (QueryString :> api) where
```

More examples of how to extend servant are found here: http://haskell-servant.github.io/extending.html

Servant also solves the content type problem in a very elegant way, for we can provide different instances for our result data type and define in this instance how to serve it.
This can be seen here where I've defined an instance for PlainText:
```haskell
instance MimeRender PlainText ResultSet where
    mimeRender _ = cs . show
```
More can be read here:
http://haskell-servant.github.io/tutorial/server.html#using-content-types-with-your-data-types

Overall it seems a very thin layer, would not add a lot of overhead (although it's relies heavily on several GHC extensions) and would encapsulate a big chunk of our custom code that deals with HTTP.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -61,6 +61,8 @@ executable postgrest
                      , parsec
                      , errors
                      , bifunctors
+                     , servant
+                     , servant-server
   hs-source-dirs:      src
   other-modules:       Paths_postgrest
                      , PostgREST.App
@@ -73,6 +75,7 @@ executable postgrest
                      , PostgREST.QueryBuilder
                      , PostgREST.RangeQuery
                      , PostgREST.Types
+                     , PostgREST.Api
 
 library
   if flag(ci)
@@ -124,6 +127,8 @@ library
                      , wai-extra
                      , wai-middleware-static
                      , warp
+                     , servant
+                     , servant-server
 
   Other-Modules:       Paths_postgrest
   Exposed-Modules:     PostgREST.App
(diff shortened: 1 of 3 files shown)


begriffs:
Sorry I haven't responded sooner about this. I'm grateful you're exploring this. How about we finish the current refactor which is pretty intense, get version 0.3 released, and then think about applying Servant?

diogob:
Sounds good, I'll close this for now and reopen once we have a more stable codebase ;)

PierreR:
I would love to see this issue re-opened ... Particularly now that `servant-swagger` and `swagger2` have been announced (https://www.reddit.com/r/haskell/comments/44gemu/announcing_servantswagger_and_swagger2/).

I would be quite interested in a swagger compatibility layer (alike `Dreamfactory`)

diogob:
@PierreR I'm not sure if this approach is interesting. Once I started coding I realized that for a dynamically generated API such as the one we have in PostgREST the servant types don't make the code much simpler. For we can't define the endpoints at compile time.

PierreR:
@diogob To be honest as a mere user of Postgrest I don't care much about it using Servant or not. But I would love to have a `swagger` compatibility layer. Maybe I should fill in a new issue for this ?

diogob:
There already quite a few issues about this, but the initial target is RAML, take a look at #8  #144 #449

PR #383, opened 2015-11-23
Type refactoring (ApiRequest/DbRequest)

ruslantalpa:
I refactored the types.
Looks a bit better to me but had to split Query data type in two because otherwise it would have a few incomplete pattern matches in functions.
In some places i added "undefined" for patterns that should not normally happen. Maybe someone has a better idea, but at least it signals it's a "bad" branch that should not be reached

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -72,7 +72,7 @@ executable postgrest
                      , PostgREST.DbStructure
                      , PostgREST.QueryBuilder
                      , PostgREST.RangeQuery
-                     , PostgREST.RequestIntent
+                     , PostgREST.ApiRequest
                      , PostgREST.Types
 
 library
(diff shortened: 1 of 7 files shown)


calebmer:
I like the `Intent` name more than `ApiRequest`. We aren't really “requesting” anything in PostgREST, but rather translating to SQL.

PierreR:
@calebmer Don't we use an API to request against Http ? To me `Intent` sounds `vague` enough.

This is just from a million miles above as I haven't had the time to look/browse/appreciate the code much. Please take my comment with a grain of salt ;-)

begriffs:
We were discussing having `ApiRequest` for the http things the client sends us, and translate it into a `DbRequest`. The `DbRequest` is turned into raw sql when it's time to do the work. Just a parity between the names.

PR #355, opened 2015-11-12
Small bug fix introduced by the last JWT PR

ruslantalpa:
status400 was not included

PR #344, opened 2015-11-05
Ensure JWT expires

calebmer:
In order to implement a secure refresh token pattern (I assume Auth0 uses this) with JWT, we need to ensure that tokens are rejected if they have expired. Therefore, these changes.

There is one breaking change, whereas incorrectly formatted JWTs once just fell through (a 404 was returned), now a bad JWT will returns a 400 with the message "Invalid JWT".

PR #392, opened 2015-11-28
Fix time fetch on initialization

### LINKED ISSUE #391: PostgREST request time is actually initialization time.
I feel bad for not catching this earlier, but there is a big security hole in PostgREST. The time we get for invalidating JWTs based on their `exp` claim [here](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Auth.hs#L55) is the time the PostgREST server started, not the time the request is made. To see how we get that time, go [here](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Middleware.hs#L39).

Anyone know how to fix this? I have no clue. Maybe take a look at whatever package we use which outputs the current time to the command line or sets the `Date` header.

calebmer:
There was a big security hole where the server was not fetching the correct date. This PR fixes that. The issue involved the `unsafePerformIO` call, I don't know why I just know that when I removed it things worked 😊 (see more at #344, [this](https://github.com/begriffs/postgrest/pull/344#discussion-diff-44341727) discussion specifically @begriffs).

Closes #391.

And I promise, no bugs 😊 (#355).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Fixed
+- Miscalculation of time used for expiring tokens - @calebmer
+
 ## [0.3.0.1] - 2015-11-27
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
 ### Fixed
 - Miscalculation of time used for expiring tokens - @calebmer
 
--- a/circle.yml
+++ b/circle.yml
@@ -13,4 +13,4 @@ dependencies:
 test:
   post:
     - cabal exec hlint -- -X QuasiQuotes src/**/*.hs test/**/*.hs
-    - cabal exec packdeps postgrest.cabal
+    - cabal exec packdeps postgrest.cabal || true
--- a/src/PostgREST/Main.hs
+++ b/src/PostgREST/Main.hs
@@ -17,6 +17,7 @@ import           Data.Functor.Identity
 import           Data.Monoid                          ((<>))
 import           Data.String.Conversions              (cs)
 import           Data.Text                            (Text)
+import           Data.Time.Clock.POSIX                (getPOSIXTime)
 import qualified Hasql                                as H
 import qualified Hasql.Postgres                       as P
 import           Network.Wai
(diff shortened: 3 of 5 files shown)


ruslantalpa:
Looks like Nikita released a new version of hasql and something is failing.
Try restricting hasql version to the last one we know it was working
https://hackage.haskell.org/package/hasql-0.7.4

begriffs:
That pesky `getPOSIXTime` had a problem after all... What value did it return from inside unsafePerformIO?

I think we still have a bug in this PR code. Although this PR will get the correct time on startup, the timestamp will always stay the same as long as the server runs. Then in

```haskell
jwtClaims :: JWT.Secret -> Text -> NominalDiffTime -> Maybe JWT.ClaimsMap
jwtClaims secret input time =
  case join $ claim JWT.exp of
    Just expires ->
      if JWT.secondsSinceEpoch expires > time
        then customClaims
        else Nothing
    _ -> customClaims
```

It is checking that `expires` is greater than time and time will never change. Hence no tokens will expire. That is my hypothesis at least.

calebmer:
@begriffs from my duck testing, this version works and the `unsafePerformIO` value was a constant (start of server).

@ruslantalpa Restricted the values to the versions I have locally. Still failing.

begriffs:
The test failing is another issue. One of the test commands I'm running is packdeps: https://github.com/begriffs/postgrest/blob/master/circle.yml#L16

Packdeps fails if there are any packages on hackage which have versions newer than we allow in our cabal file. I added the check there so that our program would not fall behind and start having problems building in newer stackage snapshots.

However in this case we're hitting a real problem: the changes in Hasql are big. Our code is incompatible with those changes. So I think we should turn packdeps into a warning rather than error. To do that you could probably add `|| true` to the end of that line in the circle config.

begriffs:
On second thought I think you're right. That `runSettings` calls our code for each client connect and recalculates the time. I thought the warp listening loop was later for some reason.

calebmer:
@begriffs you were right about `packdeps`, it's been changed in this PR and the tests are now passing. 

@ruslantalpa took out the version restrictions, good call.

begriffs:
What a nasty bug, glad you spotted it and got rid of that gross unsafe perform. Throw a line in the changelog and it looks good.

Wish we could test it somehow but like I mentioned before I don't know a way an easy way to mock time in the tests.

calebmer:
Changelog's got it.

begriffs:
Maybe add an

```md
## Unreleased
```
before your fix line. Such a nitpick, sorry. :ant:

calebmer:
Done, hope you don't mind the extra commit. Used the GitHub editor :relaxed:

PR #221, opened 2015-07-05
Qualifies columns of WHERE clauses so we can use computed columns as filters

diogob:
This PR allows the use of [computed columns](http://www.postgresql.org/docs/9.4/interactive/xfunc-sql.html#XFUNC-SQL-COMPOSITE-FUNCTIONS) as filters in GET and PATCH calls without including these fields in the view being queried.

I know that it's not a super elegant solution but it's the least invasive I could think of.

With the current code to query for any function applied to the rows I need to include this function as a field in the view. Sometimes this is undesirable, for these columns can be quite large and I may want to keep a leaner response body.

The case I ran into was filtering by a concatenation of text fields like:
```sql
SELECT * 
FROM projects 
WHERE name || ' ' || author || ' ' || description ILIKE '%search string%';
```

So instead of modelling the above concatenation as a field in my view I want to create a function such as
```sql
CREATE FUNCTION search_text(projects) RETURNS text AS $$
SELECT $1.name || ' ' || $1.author || ' ' || $1.description
$$ STABLE LANGUAGE SQL;
```

AND query it with:
```sql
SELECT * 
FROM projects 
WHERE projects.search_text ILIKE '%search string%';
```

PR #400, opened 2015-12-02
Order by computed column

diogob:
In #221 I forgot to add ORDER BY support, this PR fixes that oversight, making computed columns even more useful.

### DIFF
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -130,7 +130,11 @@ spec =
         [json| [{"text_search_vector":"'baz':1 'qux':2"}] |]
 
     it "matches with computed column" $
-      get "/items?always_true=eq.true" `shouldRespondWith`
+      get "/items?always_true=eq.true&order=id.asc" `shouldRespondWith`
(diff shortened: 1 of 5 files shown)


begriffs:
Docs, tests, changelog: perfection. :gem:

test/Feature/QuerySpec.hs
```diff
@@ -130,7 +130,11 @@ spec =
         [json| [{"text_search_vector":"'baz':1 'qux':2"}] |]
 
     it "matches with computed column" $
-      get "/items?always_true=eq.true" `shouldRespondWith`
+      get "/items?always_true=eq.true&order=id.asc" `shouldRespondWith`
+        [json| [{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}] |]
+
+    it "order by computed column" $
+      get "/items?order=always_true.asc,id.asc" `shouldRespondWith`
```
begriffs: I know that this does cover the functionality because this request would return an error prior to your code changes. However it might be nice to emphasize the sorting itself by creating a computed column such as `anti_id` which is defined to be the negative value of the id. Then order by `anti_id` descending and check that the results come out ascending by id.

diogob: That's a good idea, will do it :D

diogob: I thought of using ```bizarro_id```, but decided to chose readability over the joke :stuck_out_tongue_winking_eye:

PR #404, opened 2015-12-06
Add --max-rows option

begriffs:
Add command line option to limit the number of rows ever returned by any read request. This does not apply to inserting rows with `Prefer: return=representation`. All those kind of rows would be returned possibly in excess of the read row limit.

Also `rangeParse` now always returns a range, no more Maybe. Where it would have returned maybe before it now returns the unbounded range starting from 0. This simplifies other parts of the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Allow order by computed columns - @diogob
+- Set max rows in response with --max-rows - @begriffs
 
 ## [0.3.0.1] - 2015-11-27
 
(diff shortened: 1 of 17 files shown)


calebmer:
So if one were to request 500 rows but the limit is 100 they would only get 100? This should really apply to any `return=representation` request to prevent abuse.

calebmer:
Should max rows also limit how many rows one can bulk insert at once? Or could that be limited db side?

begriffs:
That's a good point about limiting insert size because inserts can actually be much more costly than read requests. We could use the nginx config setting [client_max_body_size](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) to do it more easily than postgrest logic.

begriffs:
OK, we've got tests and docs.

I bet there are still some corner cases though, like a client using `select=*,bigtable{*}` embedding to generate a huge response.

begriffs:
I think this is a good improvement even as it is. It prevents naive requests to a table which could have huge responses. We can tackle limiting embedded data later.

PR #406, opened 2015-12-07
Add brew install instructions back

begriffs:
Waiting for https://github.com/Homebrew/homebrew/pull/46773 to be merged to ensure that our formula upgrade techniques work.

### DIFF
--- a/docs/install/server.md
+++ b/docs/install/server.md
@@ -126,7 +126,7 @@ brew doctor
 brew install postgrest
 ```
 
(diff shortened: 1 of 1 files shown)


docs/install/server.md
```diff
@@ -111,6 +111,29 @@ The possible flags are:
...
+brew update
+
+# Check for any problems with brew's setup
+brew doctor
+
+# Install the postgrest package
+brew install postgrest
+```
+
+This will automatically install PostgreSQL as a dependency (see the [Installing PostgreSQL](#installing-postgresql) section for setup instructions). The process tends to take around 15 minutes to install the package and its dependencies.
```
mikemcquaid: Probably want to say "up to 15 minutes" as if they are installed into `/usr/local` they'll get binary packages which just take download + extraction time.

PR #409, opened 2015-12-08
Release db connections on sigint

begriffs:
When I interrupt the program with ctrl-c I see it print out the message, "Releasing connection pool...". Other than that I'm not sure if it changes the behavior of the program. To manually test I run this command in another terminal:

```bash
watch "psql -d postgrest_test -c 'SELECT sum(numbackends) FROM pg_stat_database;'" 
```

When PostgREST is not running the count is 1 because of psql itself. Once the server starts the count goes to 2. When I ctrl-c the server it goes back down to 1. The decreasing to 1 happens with or without the code in this pull request though.

@monemihir can you check if this pull request helps? I'm not sure how else to test it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Miscalculation of time used for expiring tokens - @calebmer
 - Remove bcrypt dependency to fix Windows build - @begriffs
 - Detect relations event when authenticator does not have rights to intermediate tables - @ruslantalpa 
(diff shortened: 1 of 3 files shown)


diogob:
:confetti_ball: :tada:

begriffs:
Moved the message after the pool release and it worked. Then removed the statement for logging like you said. Note that there is still that "Listening on port 3000..." message on startup which also has an incorrect format.

Also like I said the connection release may happen automatically after the process dies anyway. Not sure the PR does anything useful but it can't hurt.

PR #410, opened 2015-12-08
Refactor tests

diogob:
* Moves several haskell function calls to SQL fixtures
* Recreate database for every suite execution (now the test database is created automatically)
* Separate data fixtures from schema fixtures
* Uses postgrest_authenticator to log in database during suite execution (the test database still belongs to user postgrest_test)


diogob:
@ruslantalpa I followed your suggestions here for the role names. One test failed in the Structure.hs:
```includes all information on views for renamed columns, and raises relations to correct schema```

So I granted all privileges to the postgrest_authenticator role to make them green again. Every other test is passing after the refactor if you remove the last line in ```test/fixtures/schema.sql```

**Comment by ruslantalpa** [comment]
@diogob :scream:  :bow: :metal: 
:+1:

File: test/Feature/QuerySpec.hs [hunk 1]
```diff
@@ -11,18 +11,7 @@ import Text.Heredoc
...
-   . beforeAll (clearTable "complex_items" >> createComplexItems)
-   . beforeAll (clearTable "nullable_integer" >> createNullInteger)
-   . beforeAll (
-       clearTable "no_pk" >>
-       createNulls 2 >>
-       createLikableStrings >>
-       createJsonData)
-   . afterAll_ (clearTable "items" >> clearTable "complex_items" >> clearTable "no_pk" >> clearTable "simple_pk")
-   . around (withApp cfgDefault) $ do
+spec = around (withApp cfgDefault) $ do
```
begriffs: So much better.

test/fixtures/roles.sql
```diff
@@ -1,16 +1,7 @@
...
-BEGIN
-  IF NOT EXISTS (SELECT * FROM pg_roles WHERE rolname = rolename) THEN
-    EXECUTE format('CREATE ROLE %I %s', rolename, opts);
-    RETURN 'CREATE ROLE';
-  ELSE
-    RETURN format('ROLE ''%I'' ALREADY EXISTS', rolename);
-  END IF;
-END;
-$$;
+DROP ROLE IF EXISTS postgrest_authenticator, postgrest_anonymous, test_default_role, postgrest_test_author;
```
begriffs: Good point, just blow them away and recreate...

test/SpecHelper.hs
```diff
@@ -73,26 +67,20 @@ withApp config perform = do
...
-    <- H.acquirePool pgSettings testPoolOpts
-  void . liftIO $ H.session pool $
-    H.tx Nothing $ do
-      H.unitEx [H.stmt| drop schema if exists test cascade |]
-      H.unitEx [H.stmt| drop schema if exists private cascade |]
-      H.unitEx [H.stmt| drop schema if exists postgrest cascade |]
-
+setupDb :: IO ()
+setupDb = do
+  void $ readProcess "psql" ["-a", "-f", "test/fixtures/database.sql"] []
```
begriffs: On my mac it defaults to trying to connect to a database named after my username, which fails:

```
postgrest-0.3.0.1: test (suite: spec)
Progress: 1/2psql: FATAL:  database "j" does not exist
spec: readCreateProcess: psql "-a" "-f" "test/fixtures/database.sql" (exit 2): failed
Test suite failure for package postgrest-0.3.0.1
    spec:  exited with: ExitFailure 1
Logs printed to console
```

If you add `-d postgres` it should be OK assuming we can count on that database existing for all postgres installations.

begriffs:
This is a great PR, the tests really needed your help.

About `postgrest_authenticator`, you could revoke its extra privileges again and just make the failing test pending. After merging we can take a pass at fixing that and other pending tests (and removing the superfluous pending tests for PUT requests).

test/fixtures/roles.sql
```diff
@@ -1,16 +1,7 @@
...
+CREATE ROLE postgrest_authenticator WITH login;
```
begriffs: Should tighten up the security on this line to

```sql
CREATE ROLE postgrest_authenticator WITH login NOINHERIT;
```

Strangely this makes three Range feature specs fail on my machine.

calebmer:
Because roles are global across a PostgREST instance, could we call the user `postgrest_test_authenticator` to avoid collisions with the common name `posthrest_authenticator`?

diogob:
@begriffs regarding the NOINHERIT I think it's better to send on another PR, since it looks like it's going to take some time to get this right.

diogob:
@calebmer I've prefixed all roles used in the suite now with postgrest_test to avoid collisions, since postgrest_ is an obvious prefix to use with real roles and having long role names for the tests is no problem.

begriffs:
Works on CI and works on my mac. Should be good!

Issue #407, opened 2015-12-07
Add tests for weak authenticator role


begriffs:
The whole feature test suite uses an authenticator role with superuser access. In a real deployment the authenticator should not be able to access anything. It should be a `noinherit` role which can `set role` into the other users.

To keep existing tests free of JWT tokens we can ensure the anonymous role is strong, while the auth role is weak.

ruslantalpa:
(note to myself) Another test here is to check that relations are detected by the code in #403

ruslantalpa:
This can be closed, it's been implemented

begriffs:
If I recall, making the auth role noinherit still causes tests to fail.

ruslantalpa:
I'll try and see

ruslantalpa:
Fixed in #427

PR #403, opened 2015-12-04
Detect relations event when authenticator does not have rights to intermediate tables

ruslantalpa:
This issue might also be a problem for functions like allPrimaryKeys, allColumns because they also use information_schema ... this needs investigation.

While debugging this i was confronted with the fact i could not clearly see where the problem is because the the old way of getting all the relations from the query was changed to getting the info in 2 stages.

I am planning on changing this in the future so that the query returns all the relations directly and eliminate the need  for the functions in this line
```
let rels' = (addManyToManyRelations . raiseRelations schema syns . addParentRelations . addSynonymousRelations syns) rels
 ```

so i am going to post the current query here to have it documented
```
    /*
    -- CTE to get the original columns reference in views
    -- it returns results in the folowing format

    -- "src_table_schema","src_table_name","src_column_name","syn_table_schema","syn_table_name","syn_column_name"
    -- "basic_auth","users","email","public","authors","email"
    */
    WITH syns AS (
        WITH synonyms AS (
          /*
          -- CTE to replace the view from information_schema because the information in it depended on the logged in role
          -- notice the commented line
          */
          WITH view_column_usage AS (
            SELECT DISTINCT
                   CAST(current_database() AS character varying) AS view_catalog,
                   CAST(nv.nspname AS character varying) AS view_schema,
                   CAST(v.relname AS character varying) AS view_name,
                   CAST(current_database() AS character varying) AS table_catalog,
                   CAST(nt.nspname AS character varying) AS table_schema,
                   CAST(t.relname AS character varying) AS table_name,
                   CAST(a.attname AS character varying) AS column_name
            FROM pg_namespace nv, pg_class v, pg_depend dv,
                 pg_depend dt, pg_class t, pg_namespace nt,
                 pg_attribute a
            WHERE nv.oid = v.relnamespace
                  AND v.relkind = 'v'
                  AND v.oid = dv.refobjid
                  AND dv.refclassid = 'pg_catalog.pg_class'::regclass
                  AND dv.classid = 'pg_catalog.pg_rewrite'::regclass
                  AND dv.deptype = 'i'
                  AND dv.objid = dt.objid
                  AND dv.refobjid <> dt.refobjid
                  AND dt.classid = 'pg_catalog.pg_rewrite'::regclass
                  AND dt.refclassid = 'pg_catalog.pg_class'::regclass
                  AND dt.refobjid = t.oid
                  AND t.relnamespace = nt.oid
                  AND t.relkind IN ('r', 'v', 'f')
[...]

Issue #107, opened 2014-12-07
Show foreign key references in table options


begriffs:
It is often unable to display the foreign key references because the `constraint_column_usage` shows "Only those columns ... that are contained in a table owned by a currently enabled role."

http://www.postgresql.org/docs/9.3/static/infoschema-constraint-column-usage.html

begriffs:
@ruslantalpa I tried to un-pend this test https://github.com/begriffs/postgrest/blob/3e9f9f300c6144179e943aa2d7ae65dded3203e2/test/Feature/StructureSpec.hs#L253

It almost works except it reports that the `auto_inc_fkey` column references the `insertable_view_with_join` table rather than the `auto_incrementing_pk` table.

```json
{
  "pkey": [
    "id"
  ],
  "columns": [
    {
      "references": null,
      "default": "nextval('\"1\".has_fk_id_seq'::regclass)",
      "precision": 64,
      "updatable": true,
      "schema": "1",
      "name": "id",
      "type": "bigint",
      "maxLen": null,
      "enum": [],
      "nullable": false,
      "position": 1
    },
    {
      "references": {
        "column": "id",
        "table": "insertable_view_with_join"
      },
      "default": null,
      "precision": 32,
      "updatable": true,
      "schema": "1",
      "name": "auto_inc_fk",
      "type": "integer",
      "maxLen": null,
      "enum": [],
      "nullable": true,
      "position": 2
    },
    {
      "references": {
        "column": "k",
        "table": "simple_pk"
      },
      "default": null,
      "precision": null,
      "updatable": true,
      "schema": "1",
      "name": "simple_fk",
      "type": "character varying",
      "maxLen": 255,
      "enum": [],
      "nullable": true,
      "position": 3
    }
  ]
}
```

PR #415, opened 2015-12-09
Test and fixes for #407 #403 #107

ruslantalpa:
This PR adds tests for a weak authenticator #407, and ads a check for detecting relations #403
A good side effect is it fixes also an old issue #107 (test passing, it can be closed).

One thing i noticed is that test got a lot slower from 20s to 90s.
The only major thing that changed is that we rely on CTE instead of information_schema and the new queries take about 500ms to execute. Since the queries in the dbstructure are executed only once in the beginning i don't think it will impact runtime performance however i think that these queries are somehow rerun at each test. 
Can anyone spot why this is happening and maybe fix it in a separate PR?

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -202,6 +202,98 @@ allColumns tabs = do
           info.column_default AS default_value,
           array_to_string(enum_info.vals, ',') AS enum
       FROM (
+          /*
+          -- CTE based on information_schema.columns to remove the owner filter
+          */
+          WITH columns AS (
+              SELECT current_database()::information_schema.sql_identifier AS table_catalog,
+                  nc.nspname::information_schema.sql_identifier AS table_schema,
+                  c.relname::information_schema.sql_identifier AS table_name,
+                  a.attname::information_schema.sql_identifier AS column_name,
+                  a.attnum::information_schema.cardinal_number AS ordinal_position,
+                  pg_get_expr(ad.adbin, ad.adrelid)::information_schema.character_data AS column_default,
+                      CASE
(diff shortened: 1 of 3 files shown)


diogob:
@ruslantalpa during the tests it seems that the DbStructure gets recreated for each test case.
The problem is here: https://github.com/begriffs/postgrest/blob/master/test/SpecHelper.hs#L58
The withApp is called in ```around``` for each case.

ruslantalpa:
Thought so, anyway, this does not impact the runtime so i think that can be fixed with another PR

begriffs:
(from a conversation in the chat room)

I'd rather get the tests fast again before merging. It may take a little while to figure out how to do it and during that time people branching off master would have to deal with slow tests. 

One technique to make it faster would be to make `withApp` accept a `DbStructure` argument so it wouldn't have to run `getDbStructure` inside. However it's unclear how to get that value from Main.hs into each of the specs. We use hspec-discover to find and run each spec. [This article](https://stackoverflow.com/questions/30631737/hspec-discovery-custom-main-and-passing-argument-to-spec) makes it seem like we can't get an extra argument into the spec that way. We could ditch hspec-discover and call the tests manually from Main and pass them the db structure to give to withApp.

Another technique could be to not call withApp so often. It gets called for every spec item. Maybe it could be put into a `beforeAll` or something. We would have to change the signature of the function somehow.

begriffs:
As mentioned in the chatroom, I pushed a patch to make the tests fast on branch spec-cache-dbstructure. One test is failing there due to the changes but the whole suite runs in three seconds.

diogob:
roadrunner

ruslantalpa:
Closing this one since the changes in here are included in #421

PR #417, opened 2015-12-09
Fix #399 insert records in tables with no SELECT privileges

### LINKED ISSUE #399: Unable to insert into table using a write-only role
Assuming I want public writability, but secured readability.

Say
```
CREATE TABLE test (id INTEGER, doc JSON);
GRANT INSERT ON test TO anonymous;
GRANT ALL ON test TO admin;
```
running with
```
$ postgrest postgres://admin@localhost:5432/db --port 3000 --schema public --anonymous anonymous --pool 50 --jwt-secret secret
```
I would expect that
```
$ curl -X POST -H "Content-Type: application/json" -d '{"id":2,"doc":{"foo":"bar"}}' 'http://host:3000/test'
```
returns success instead of
```
{"hint":null,"details":null,"code":"42501","message":"permission denied for relation logs"}
```
What am I doing wrong?

ruslantalpa:
This PR adds the possibility of insterting records in tables with no SELECT privileges if you supply the header
```
Prefer: return=none
```
I don't really like the idea of an extra header but i don't see a way around this. Also this "feature" made the code a bit dirtier, especially the "createWriteStatement"

Let's discuss the issue like so:
Is the extra header the right way to go?
   yes -> can the code that implements it be nicer?
   no -> how do we put an "IF" around the "RETURNING *" in SQL ?

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -33,6 +33,8 @@ data Action = ActionCreate | ActionRead
 data Target = TargetIdent QualifiedIdentifier
             | TargetRoot
             | TargetUnknown [T.Text]
+-- | How to return the inserted data
+data PreferRepresentation = Full | HeadersOnly | None deriving Eq
 -- | Enumeration of currently supported content types for
 -- route responses and upload payloads
 data ContentType = ApplicationJSON | TextCSV deriving Eq
@@ -59,7 +61,7 @@ data ApiRequest = ApiRequest {
   -- | Data sent by client and used for mutation actions
   , iPayload :: Maybe Payload
   -- | If client wants created items echoed back
-  , iPreferRepresentation :: Bool
+  , iPreferRepresentation :: PreferRepresentation
   -- | If client wants first row as raw object
   , iPreferSingular :: Bool
   -- | Whether the client wants a result count (slower)
@@ -119,7 +121,7 @@ userApiRequest schema req reqBody =
   , iTarget = target
   , iAccepts = pickContentType $ lookupHeader "accept"
   , iPayload = relevantPayload
-  , iPreferRepresentation = hasPrefer "return=representation"
+  , iPreferRepresentation = representation
   , iPreferSingular = singular
   , iPreferCount = not $ hasPrefer "count=none"
   , iFilters = [ (k, fromJust v) | (k,v) <- qParams, k `notElem` ["select", "order"], isJust v ]
@@ -138,8 +140,10 @@ userApiRequest schema req reqBody =
   lookupHeader    = flip lookup hdrs
   hasPrefer val   = any (\(h,v) -> h == "Prefer" && v == val) hdrs
   singular        = hasPrefer "plurality=singular"
-
-
+  representation
+    | hasPrefer "return=representation" = Full
+    | hasPrefer "return=minimal" = None
+    | otherwise = HeadersOnly
 
 -- PRIVATE ---------------------------------------------------------------
 
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -52,6 +52,7 @@ import           Data.Scientific         ( FPFormat (..)
                                          , isInteger
                                          )
 import           Prelude hiding          (unwords)
+import           PostgREST.ApiRequest    (PreferRepresentation (..))
 
 type PStmt = H.Stmt P.Postgres
 instance Monoid PStmt where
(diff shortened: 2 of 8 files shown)


calebmer:
Can we invert the option? So the count is off by default but with `Prefer: count=all` it is enabled.

ruslantalpa:
The default for "GET" is to return the count, that's why i chose that as a default for INSERT/UPDATE

calebmer:
Ok, I like inverting the option as it makes some sense a `POST` shouldn't read by default.

Another thought, maybe we can universally swap the option. That's currently what I do in my PostgREST client to eek out some more performance gains. (I actually really like that idea)

src/PostgREST/ApiRequest.hs
```diff
@@ -59,7 +61,7 @@ data ApiRequest = ApiRequest {
   -- | Data sent by client and used for mutation actions
   , iPayload :: Maybe Payload
   -- | If client wants created items echoed back
-  , iPreferRepresentation :: Bool
+  , iPreferRepresentation :: PreferRepresentaion
```
begriffs: Typo: PreferRepresenta[t]ion

src/PostgREST/ApiRequest.hs
```diff
@@ -138,8 +140,10 @@ userApiRequest schema req reqBody =
   lookupHeader    = flip lookup hdrs
   hasPrefer val   = any (\(h,v) -> h == "Prefer" && v == val) hdrs
   singular        = hasPrefer "plurality=singular"
-
-
+  representation
+    | hasPrefer "return=representation" = Full
+    | hasPrefer "return=none" = None
```
begriffs: [RFC7240](https://tools.ietf.org/html/rfc7240#page-9) suggests the string "minimal" rather than "none."

begriffs:
But like you say the broader question is whether we need a header or whether we can detect the role's inability to insert and issue the sql without the returning clause.

begriffs:
Two other things we could do to allow inserts into write only tables without needing an extra header.

1. On startup consult the information schema for select permissions like [this](http://www.simononsoftware.com/how-to-check-postgresql-privileges/) except query for all roles. Add this info to `DbStructure`. The insert query builder can refer to this information to add or withhold `RETURNING *` from the query.
2. Catch the access denied exception and retry the query minus `RETURNING *`. The slowdown would occur on every insert to the table though.

diogob:
My 2 cents: The extra header makes the code simpler and the semantics clearer.

src/PostgREST/QueryBuilder.hs
```diff
@@ -73,23 +74,39 @@ createReadStatement selectQuery range isSingle countTable asCsv =
...
+  ) (V.singleton . B.encodeValue . JSON.Array . V.map JSON.Object $ rows) True
+createWriteStatement qi selectQuery mutateQuery isSingle Count
+                     pKeys _ (PayloadJSON (UniformObjects rows)) =
+  B.Stmt (
+    wrapQuery (mutateQuery <> " RETURNING " <> fromQi qi <> ".*") [
+      countNoneF, -- when updateing it does not make sense
+      countF,
+      if isSingle then locationF pKeys else "null",
+      "null"
+    ] selectQuery
```
ruslantalpa: this probably is wrong, no need for selectQuery when doing only count

ruslantalpa:
@diogob so you are saying you prefer to do it like in this PR?

diogob:
Yes, I'd keep the design as is proposed by this PR.

begriffs:
I do agree with you @diogob, supporting the header to fix this issue is nice and simple. The only confusing thing is that an API consumer needs to realize this quirk of database access that inserting a new row tries to read the row as well.

We should probably modify the access denied error message to add a hint about trying `Prefer: representation=minimal` otherwise I predict it'll be a recurring source of confusion.

src/PostgREST/ApiRequest.hs
```diff
@@ -33,6 +33,8 @@ data Action = ActionCreate | ActionRead
 data Target = TargetIdent QualifiedIdentifier
             | TargetRoot
             | TargetUnknown [T.Text]
+-- | How to return the inserted data
+data PreferRepresentation = Full | Count | None deriving Eq
```
begriffs: Using the term `Count` here is confusing I think. There are three Prefer header settings the client can use: `return`, `count`, and `plurality`. Since the above line deals with `return` and not whether the count is calculated the name `LocationHeader` would probably be a better choice than `Count`.

begriffs:
Things to do before merge:

* Rebase against master
* Change `Count` to `HeadersOnly` in `PreferRepresentation`

ruslantalpa:
Rebased and changed to HeadersOnly

Issue #365, opened 2015-11-15
Statically compiling postgrest


ruslantalpa:
I think we should try and compile v3 like described here
http://begriffs.com/posts/2015-08-11-dev-deploy-haskell-docker.html
https://github.com/nilcons/ghc-musl

PierreR:
Thanks for the pointers. I haven't watched the video yet but wouldn't be much better to have this issue fixed: http://bugs.alpinelinux.org/issues/4257

There is a risk here to get an complicated docker build process which might stop being maintained in the future. To me a much more ideal situation  would be to have native `alpine` ghc and `stack` packages. Then you have the leverage to build a nice small docker image.

PierreR:
I am also waiting for `stack` to offer the possibility to build inside custom small docker image. Right now, only Ubuntu precise is supported (which is not option for a small image). This is a known issue and I read somewhere that it might be fixed in a couple of weeks/months.

PierreR:
Another alternative that does not require docker is described here:
https://getpocket.com/a/read/1084317437

ruslantalpa:
The idea is not the docker thing (that is just the build env)
I was sugesting building the binary so that it does not have dependencies and its compiled with ghc that itself was built against musl so that the resulting bin is as small as posible, i doubt we depend much on things from glibc

begriffs:
I'm OK with building binaries for several platforms at each release. If you figure out how to make a portable statically linked binary then we can distribute that instead, but things are OK for now.

steve-chavez:
This would be specially useful on Windows, users keep reporting problems there.

Currently we require a full PostgreSQL installation in Windows to use pgrst and when the db is in another server that dependency seems too big. I don't know a way of getting just `libpq` on Windows(haven't tried [msys2](https://www.msys2.org/)).

steve-chavez:
Closing in favor of #1295.

PR #419, opened 2015-12-09
Add docker file with configurable env vars

jackfirth:
This pull request adds a simple dockerfile that can be configured to build an image for any PostgREST version not less than 0.3 (due to the change in how flags are passed to postgrest).  It uses a build arg (`ARG POSTGREST_VERSION`) which is only available since docker 1.9. Older docker versions can still run the built images however.

The dockerfile has different defaults than postgrest does, in order to line up with the defaults for the Postgres image. For instance, assuming this is used to build a `postgrest:0.3.0.1` image, the following docker-compose.yml links a postgrest container to a postgres database container:

```yaml
postgrest:
  image: postgrest:0.3.0.1
  links:
    - db:db
db:
  image: postgres
```

This works because the postgrest image has configurable `POSTGREST_DB_USER`, `POSTGREST_DB_PASSWORD`, `POSTGREST_DB_HOST`, `POSTGREST_DB_PORT` env vars that default to values that happen to directly connect to a postgres database running at host "db" (the host inserted by the link) on port `5432` (default postgres port) with username and password both `postgrest` (the official postgres docker image's defaults). You can connect to an arbitrary postgres database (including an amazon one) not in a docker container by changing those variables, with no need for `external_links`.

Future work is to set up a docker hub image repository and push images for each version to it, so that docker users won't need to build anything themselves.

This PR also adds Google Inc. to the license due to legal requirements of my employment. This does not change the underlying license terms in any way, and approval for this PR was given through an automated process - Google has no desire to manage or interfere with this project. I can also request copyright myself for this PR, but that's subject to weeks of review and bureaucratic overhead and may not be approved.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
 Copyright (c) 2014 Joe Nelson
+Modified work Copyright 2015 Google Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,10 +26,10 @@ ENV POSTGREST_CONNECTION_POOL_SIZE 10
 ENV POSTGREST_JWT_SECRET secret
 ENV POSTGREST_SCHEMA public
 
-# Connstring shorthand variable
-ENV CONNSTRING postgres://${POSTGREST_DB_USER}:${POSTGREST_DB_PASS}@${POSTGREST_DB_HOST}:${POSTGREST_DB_PORT}/${POSTGREST_DB_NAME}
+# POSTGREST_CONNSTRING can be set as well. If not set it defaults to a
+# connstring constructed from the user, password, host, port, and db name vars
 
-CMD postgrest ${CONNSTRING} \
+CMD postgrest ${POSTGREST_CONNSTRING:=postgres://${POSTGREST_DB_USER}:${POSTGREST_DB_PASS}@${POSTGREST_DB_HOST}:${POSTGREST_DB_PORT}/${POSTGREST_DB_NAME}} \
               --port 3000 \
               --anonymous ${POSTGREST_ANON_USER} \
               --schema ${POSTGREST_SCHEMA} \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,38 @@
+FROM debian:8.2
+
+ARG POSTGREST_VERSION
+
+# Shorthands to make commands easier to read
+ENV POSTGREST_DOWNLOAD postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
+ENV POSTGREST_RELEASES_URL http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}
+
+RUN apt-get update \
+  && apt-get install -y xz-utils wget libpq-dev \
+  && wget ${POSTGREST_RELEASES_URL}/${POSTGREST_DOWNLOAD} \
+  && tar xf ${POSTGREST_DOWNLOAD} \
+  && mv postgrest /usr/local/bin/postgrest \
+  && rm -f ${POSTGREST_DOWNLOAD} \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Public env vars, can be passed in as params
+ENV POSTGREST_DB_HOST db
+ENV POSTGREST_DB_PORT 5432
+ENV POSTGREST_DB_USER postgres
+ENV POSTGREST_DB_NAME ${POSTGREST_DB_USER}
+ENV POSTGREST_ANON_USER ${POSTGREST_DB_USER}
+ENV POSTGREST_DB_PASS ${POSTGREST_DB_USER}
+ENV POSTGREST_CONNECTION_POOL_SIZE 10
+ENV POSTGREST_JWT_SECRET secret
+ENV POSTGREST_SCHEMA public
+
+# Connstring shorthand variable
+ENV CONNSTRING postgres://${POSTGREST_DB_USER}:${POSTGREST_DB_PASS}@${POSTGREST_DB_HOST}:${POSTGREST_DB_PORT}/${POSTGREST_DB_NAME}
+
+CMD postgrest ${CONNSTRING} \
+              --port 3000 \
+              --anonymous ${POSTGREST_ANON_USER} \
+              --schema ${POSTGREST_SCHEMA} \
+              --jwt-secret ${POSTGREST_JWT_SECRET}
+
+EXPOSE 3000


LICENSE
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2014 Joe Nelson
+Modified work Copyright 2015 Google Inc.
```
ruslantalpa: There is no rush/immediate need for the docker config file, the PR can stay open until there is permission to remove this line.

docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# Public env vars, can be passed in as params
+ENV POSTGREST_DB_HOST db
+ENV POSTGREST_DB_PORT 5432
+ENV POSTGREST_DB_USER postgres
+ENV POSTGREST_DB_NAME ${POSTGREST_DB_USER}
+ENV POSTGREST_ANON_USER ${POSTGREST_DB_USER}
+ENV POSTGREST_DB_PASS ${POSTGREST_DB_USER}
+ENV POSTGREST_CONNECTION_POOL_SIZE 10
+ENV POSTGREST_JWT_SECRET secret
+ENV POSTGREST_SCHEMA public
```
ruslantalpa: All the above parameters need to be configurable, especially jwt_secret, but i would make them all configurable, event the port (3000)

jackfirth: The port can't be made configurable because the configuration happens at container run-time, after the image is built, while the port exposing happens at image build time. It's not a big deal anyway, because container ports are internal to the container - when running them their exposed ports can be re-mapped by Docker.

docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Connstring shorthand variable
+ENV CONNSTRING postgres://${POSTGREST_DB_USER}:${POSTGREST_DB_PASS}@${POSTGREST_DB_HOST}:${POSTGREST_DB_PORT}/${POSTGREST_DB_NAME}
```
ruslantalpa: a connection string has a lot more options/flexibility, why hardcode it?
Why not make it just one parameter POSTGREST_CONNECTION_STRING, basically all the parameters of postgrest would have equivalent docker params

jackfirth: To play nicely with the Postgres container when using docker-compose. When using a postgres container, the host is fixed - it's always the service name - so the container only lets you configure the username, password, and port. If this image only used a connstring, then the user has to figure out that the host comes from the postgres container name and mentally construct the connstring. An explicit connstring that could be provided instead of these env vars would be a good option.

ruslantalpa: I am not sure what was your conclusion above? Is a single evn var for connection string a good idea?
I think it is. I might want to use docker postgrest container but my db is not in a container (most of the time isn't) so this would limit the expressiveness of the connection string.
For the case where everything is in a container maybe the "composition" definition can expose it's own vars (like in this example) and form the connection string internally (don't know docker, maybe this is not possible)

jackfirth: I think the image should work with *both* a connstring and the multiple env vars, and the user chooses which one to use. I don't know how to do that terribly well yet, it'd probably be some ugly bash scripting. Supporting the multiple vars first makes sense to me as that's much easier to use in development (with a db container).

LICENSE
jackfirth: Submitted the request, says I can expect a response in a month.

calebmer: I don't think this is a blocked, but it's @begriffs final call.

docker/Dockerfile
calebmer: Having the option to set the connection string out right would be greatly appreciated. The following should work.

```bash
if [ -z $POSTGREST_DB ]
then
  CONNSTRING=(what's there now)
else
  CONNSTRING=$POSTGREST_DB
fi
```

docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM debian:8.2
```
calebmer: I'm not too docker literate, but might this be easily implemented off of a `scratch` base to reduce size? If it's too hard to get the binary in the container with `scratch` it doesn't need to be changed (although couldn't we just build before a docker push?).

jackfirth: We'd need #365 implemented first - the current binary has too many dynamic links to be easily usable in a minimal image. Ideally the docker image would be built and pushed as part of CI, I'd like to get that working next. With those both done it'd be very easy to make a minimal image: write a dockerfile that just adds the built binary into it during CI, no download necessary.

docker/Dockerfile
jackfirth: Works with a POSTGREST_CONNSTRING env var now, through bash magic

begriffs:
Seldom do I let political concerns interfere with open source work but this is one case where I do.

Google asking you to sign away your copyright on your spare-time independent work is frankly abusive. I will not accept pull requests that assign copyright to Google.

As this hacker news thread suggests, prospective employees should negotiate for their intellectual property rights during the hiring process. https://news.ycombinator.com/item?id=1969979

Sorry it's just a hot-button issue for me and I'm closing this PR.

ruslantalpa:
Sorry for posting in a closed PR, maybe we could have a compromise :)
a) @jackfirth could add the license and the copyright mention to this specific file and it would cover only this file.
b) maybe we should remove configuration files like docker/debian from this repository and open other smaller repositories for them. For example, @jackfirth could open his own repository that will provide docker config files and there he can add the license the way he chooses and we can link to his repository in the "ecosystem" docs page

jackfirth:
I can understand your position. I'd personally prefer not to wait the handful of decades it will take for copyright law to get properly reformed where this situation could be resolved in an ideal fashion, so I'd be more than happy to move this work to my own repo and provide the docker images and infrastructure as a separate body of work. If Google decides to release copyright on this patch, I'll remove it at once. Would that be an amicable solution?

begriffs:
Sure either that or use [file-scope copyright notice](https://softwarefreedom.org/resources/2012/ManagingCopyrightInformation.html#maintaining-file-scope-copyright-notices) for the files you contribute. (Good idea Ruslan, I didn't realize this was an option.)

Probably keeping the docker files in this repo would be simplest for future maintenance. If you want rebase your commits so that `docker/Dockerfile` contains the new copyright and `LICENSE` does not then I'm OK with it.

begriffs:
It's gotten quiet on this PR, hope I didn't scare you away by my reaction to the licensing. Want to add the copyright notice inside the dockerfile itself?

docker/Dockerfile
ptman: May I suggest either `debian:8` or `debian:jessie` instead of locking it down to a specific point-release. That would ensure security updates, and I believe binary compatibility would be maintained. But I'm not that familiar with deploying Haskell programs.

LICENSE
begriffs: @jackfirth any word back on your request? (been about a month)

bobcolner:
I have submitted an independent postgREST Dockerfile implantation: https://github.com/begriffs/postgrest/pull/497

begriffs:
@bobcolner thanks for reviving this effort! We'll switch over to your PR.

PR #422, opened 2015-12-13
`pgFmtIdent` always quotes #388

### LINKED ISSUE #388: Columns named as reserved words
when a column name is a [reserved word](http://www.postgresql.org/docs/7.3/static/sql-keywords-appendix.html) the generated query fails. In my case with a column called "end".
```sql
    CREATE SEQUENCE regions_id_seq;
    CREATE TABLE regions (
	  "id"        integer NOT NULL DEFAULT nextval('regions_id_seq'),
	  "start"     varchar(255),   
	  "end"       varchar(255),
	  "note"      varchar(255),   
	  "id_song"   integer,           
	  "version"   varchar(255),  
	  "position"  integer,
	  Constraint "regions_id_pkey" Primary Key ("id")
    );
```
postges error log
```log
2015-11-27 21:19:33 UTC ERROR:  syntax error at or near "end" at character 111
2015-11-27 21:19:33 UTC STATEMENT:  WITH pg_source AS (UPDATE  public.regions  SET note='hello'::unknown,start='12'::unknown,id_song='2'::unknown,end='22'::unknown,version=null,position='10'::unknown  WHERE  public.regions.id = '17'::unknown  RETURNING public.regions.*) SELECT null, pg_catalog.count(t), null, null FROM ( SELECT  pg_source.* FROM  pg_source   LIMIT ALL OFFSET 0) t
```
In my situation I'll  rename the column to "ending" but I though was good idea to report it.

calebmer:
While I was knocking around the PostgREST codebase I wanted to knock out some of the easy issues. There were two I found. It might be nice to have a tag to denote these issues, anyway this should close #388. From what I know quoting everything shouldn't be a problem, maybe @diogob should weigh in?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Detect relations event when authenticator does not have rights to intermediate tables - @ruslantalpa
 - Ensure db connections released on sigint - @begriffs
 - Fix #396 include records with missing parents - @ruslantalpa
+- `pgFmtIdent` always quotes #388 - @calebmer
 
 ### Added
 - Allow order by computed columns - @diogob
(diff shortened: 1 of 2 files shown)


diogob:
:+1:  Great PR! Simpler code and we should have no problem quoting everything. Actually this will make everything case sensitive (as it should be).

begriffs:
Yes, great work!

PR #423, opened 2015-12-13
Change default schema #389

### LINKED ISSUE #389: Change default schema if argument not provided
It currently defaults to `1` from the days when a single postgrest server did the versioning. A better value would be  `public` because that is the default schema used by postgresql in a new database.

Then update the postgrest.com examples.

calebmer:
This was the second issue. Closes #389 and ends the versioning legacy ☺️

Except for all the API devs that are smart and still version their schemas…

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Detect relations event when authenticator does not have rights to intermediate tables - @ruslantalpa
 - Ensure db connections released on sigint - @begriffs
 - Fix #396 include records with missing parents - @ruslantalpa
+- Default schema, changed from `"1"` to `public` - @calebmer
 
 ### Added
 - Allow order by computed columns - @diogob
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -51,7 +51,7 @@ argParser = AppConfig
 
   <*> option auto  (long "port"       <> short 'p' <> help "port number on which to run HTTP server" <> metavar "PORT" <> value 3000 <> showDefault)
   <*> strOption    (long "anonymous"  <> short 'a' <> help "postgres role to use for non-authenticated requests" <> metavar "ROLE")
-  <*> strOption    (long "schema"     <> short 's' <> help "schema to use for API routes" <> metavar "NAME" <> value "1" <> showDefault)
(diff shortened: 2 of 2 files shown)


diogob:
Now I have mixed feelings about this one, I believe that modelling the API in the public schema is a bad practice. Am I alone on this one?

calebmer:
It might be, but it's a great story for on boarding. Just do `create table people (…);` and `postgrest localhost:5432 -a postgres` and things just work.

I don't think `"1"` is an intelligible default without versioning, but `public` kinda is, even if it's a bad practice.

begriffs:
It is nice for a demo to be able to create some tables in the default place and start the server and have it work. If the default schema is `1` it doesn't complain when that schema does not exist, it just can'f find any tables. That can be confusing for a newcomer who think it's ignoring their tables.

And maybe it's not so bad because if you want to make version 2 routes you can still create another schema for them. Probably doesn't actually matter if the v1 schema is called "1" or "public," it can be superseded by new versions. And if you want to change v1 tables into views you can create a new schema like "private," move the tables in there, and create views with the old names in the public schema.

In short, it's less confusing for newcomers if we use the public schema by default, and doesn't trap people in bad designs. Am I overlooking something?

I hesitated to make the change after releasing 0.3 because it's technically backward incompatible. If it's a part of 0.3.0.2 a command line invocation which used to work in 0.3.0.1 could break if someone actually had their v1 tables in the "1" schema.

diogob:
I agree that is less confusing for newcomers and it does not trap users in bad design indeed. Maybe the public is well suited for an API (given its name) as long as you use other namespaces for private objects. And we can point out this issue about having private schemas in the docs.
:+1: for public as default

calebmer:
True statement about the breaking change…

In my mind this should have been changed when versioning was removed, so we could say this is a bug fix and was always the intention of that breaking change?

Or we could start the `v4` branch.

diogob:
My vote is to use the master branch (I'm not a big fan of branches using version numbers) and release a 0.4 version in time for christmas :santa:

ruslantalpa:
I dont think this merits a major version change. It should go on 0.3.0.2, or maybe 3.1 considering the selection of parent items by fk

begriffs:
It *is* pretty intense to release a major version for this. We can consider it as correcting an omission from the original breaking change of removing versioning.

PR #424, opened 2015-12-13
Enable selection by column

calebmer:
Enables selection by column in a much more concise way for the new `3.x` codebase. I realize this pull request has historically been a source of contention, so if you don't want to merge immediately I understand. I'm going to use this in my application and then report back once I'm fully confident that this feature should be in PostgREST.

This request also includes the capability to embed rows by column without an `_id`, the row is also embedded with the sans `_id` name. This makes it more semantic and fixes the problem for some where say `author_id` could be both an integer and an object. It is the last resort option if no other relations could be found.

Oh, this PR also fixes a small capitalization bug ([see here](https://github.com/calebmer/postgrest/commit/68034f6f5ae640bd786f188ce39ef89b061ca964#diff-76fcef3192f95790fdcada5de992f8b4L227)).

### DIFF
--- a/docs/api/reading.md
+++ b/docs/api/reading.md
@@ -227,10 +227,11 @@ Content-Range → 0-14/*
 
 ### Embedding Foreign Entities
 
-Suppose you have a `projects` table which references `clients` through
-a foreign key called `client_id`. When listing projects through the
-API you can have it embed the client within each project response.
-For example,
+To help you make fewer requests, PostgREST allows the embedding of
+traditional SQL relationships into a response. Suppose you have a
+`projects` table which references `clients` through a foreign key
+called `client_id`. When listing projects through the API you can
+have it embed the client within each project response. For example,
 
 ```HTTP
 GET /projects?id=eq.1&select=id, name, clients{*}
@@ -256,6 +257,33 @@ but the the select query is recursive. You could for instance specify
 ```HTTP
 GET /foo?select=x, y, bar{z, w, baz{*}}
 ```
+
+You can select not only using table names, but also column names!
+To embed the same foreign key row from our client example earlier
+you could do the following:
+
+```HTTP
+GET /projects?id=eq.1&select=id, name, client_id{*}
+```
+
+In the response there will be a `client_id` object containing all
+the data for that row.
+
+However, a `client_id` object doesn't make a lot of sense, so you
+could do one of two things. Create a view which renames `client_id`
+to just `client` (this is the hard way), or just try `client{*}`
+in the select parameter! PostgREST supports smart ducktype checking
+for common foreign key names, so if your column name ends with
+`_id`, `_fk`, or any variation of the two (including camelcase)
+you can embed a row with just the name's beginning.
+
+So for a complete example:
+
+```HTTP
+GET /projects?id=eq.1&select=id, name, client{*}
+```
+
+Would embed in the `client` key the row referenced with `client_id`.
 
 ### Response Format
 
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -143,7 +143,7 @@ addRelations schema allRelations parentNode node@(Node readNode@(query, (name, _
       find (\r -> s == tableSchema (relTable r) && s == tableSchema (relFTable r) && t1 == tableName (relTable r) && t2 == tableName (relFTable r)) allRelations
     findRelationByColumn s t c =
       find (\r -> s == tableSchema (relTable r) && s == tableSchema (relFTable r) && t == tableName (relFTable r) && length (relFColumns r) == 1 && c `colMatches` (colName . head . relFColumns) r) allRelations
-      where n `colMatches` rc = (cs ("^" <> rc <> "(?:|_id|Id)$") :: BS.ByteString) =~ (cs n :: BS.ByteString)
+      where n `colMatches` rc = (cs ("^" <> rc <> "_?(?:|[iI][dD]|[fF][kK])$") :: BS.ByteString) =~ (cs n :: BS.ByteString)
 
 addJoinConditions :: Schema -> ReadRequest -> Either Text ReadRequest
 addJoinConditions schema (Node (query, (n, r)) forest) =
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -120,20 +120,25 @@ createWriteStatement qi selectQuery mutateQuery isSingle Full
       | otherwise = asJsonF
 
 addRelations :: Schema -> [Relation] -> Maybe ReadRequest -> ReadRequest -> Either Text ReadRequest
-addRelations schema allRelations parentNode node@(Node n@(query, (table, _)) forest) =
+addRelations schema allRelations parentNode node@(Node readNode@(query, (name, _)) forest) =
   case parentNode of
-    Nothing -> Node (query, (table, Nothing)) <$> updatedForest
-    (Just (Node (_, (parentTable, _)) _)) -> Node <$> (addRel n <$> rel) <*> updatedForest
+    (Just (Node (Select{from=[parentTable]}, (_, _)) _)) -> Node <$> (addRel readNode <$> rel) <*> updatedForest
       where
-        rel = note ("no relation between " <> table <> " and " <> parentTable)
-            $  findRelation schema table parentTable
-           <|> findRelation schema parentTable table
+        rel = note ("no relation between " <> parentTable <> " and " <> name)
+            $  findRelationTable schema name parentTable
+           <|> findRelationTable schema parentTable name
+           <|> findRelationColumn schema parentTable name
         addRel :: (ReadQuery, (NodeName, Maybe Relation)) -> Relation -> (ReadQuery, (NodeName, Maybe Relation))
-        addRel (q, (t, _)) r = (q, (t, Just r))
+        addRel (q, (n, _)) r = (q {from=fromRelation}, (n, Just r))
+          where fromRelation = map (\t -> if t == n then tableName (relTable r) else t) (from q)
+
+    _ -> Node (query, (name, Nothing)) <$> updatedForest
   where
     updatedForest = mapM (addRelations schema allRelations (Just node)) forest
-    findRelation s t1 t2 =
-      find (\r -> s == (tableSchema . relTable) r && t1 == (tableName . relTable) r && t2 == (tableName . relFTable) r) allRelations
+    findRelationTable s t1 t2 =
+      find (\r -> s == tableSchema (relTable r) && t1 == tableName (relTable r) && t2 == tableName (relFTable r)) allRelations
+    findRelationColumn s t c =
+      find (\r -> s == tableSchema (relTable r) && s == tableSchema (relFTable r) && t == tableName (relFTable r) && length (relFColumns r) == 1 && c == (colName . head . relFColumns) r) allRelations
 
 addJoinConditions :: Schema -> ReadRequest -> Either Text ReadRequest
 addJoinConditions schema (Node (query, (n, r)) forest) =
@@ -218,11 +223,12 @@ requestToCountQuery schema (DbRead (Node (Select _ _ conditions _, (mainTbl, _))
 requestToQuery :: Schema -> DbRequest -> SqlQuery
 requestToQuery _ (DbMutate (Insert _ (PayloadParseError _))) = undefined
 requestToQuery _ (DbMutate (Update _ (PayloadParseError _) _)) = undefined
-requestToQuery schema (DbRead (Node (Select colSelects tbls conditions ord, (mainTbl, _)) forest)) =
+requestToQuery schema (DbRead (Node (Select colSelects tbls conditions ord, (nodeName, maybeRelation)) forest)) =
   query
   where
     -- TODO! the folloing helper functions are just to remove the "schema" part when the table is "source" which is the name
     -- of our WITH query part
+    mainTbl = fromMaybe nodeName (tableName . relTable <$> maybeRelation)
     tblSchema tbl = if tbl == sourceCTEName then "" else schema
     qi = QualifiedIdentifier (tblSchema mainTbl) mainTbl
     toQi t = QualifiedIdentifier (tblSchema t) t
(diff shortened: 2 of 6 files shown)


ruslantalpa:
Back when we had the argument the code looked a bit different (the name of the types), in particular what a ApiRequest meant, when talking about relations it specifically mentioned "table", now it's refactored and it's called "NodeName", so no problem if that string would be used for a table or column. 
The impact on the code is also minimal (i thought it would be much bigger and maybe it would've been in the old structure), basically just one additional function so that is great too.
Overall i am in favour of this pr.
There are a few things that need a bit of change like:
The name of the functions findRelationTable findRelationColumn need to change to findRelationByTable or something similar, right now it suggests it finds a related table or a related column.
Both functions need some comments to explain what they are doing and how they are going about doing that.
Matching _id && Id feels a bit hardcoded, at least one other (commonly used) name would be _fk. and there is no match for _ID.

calebmer:
@ruslantalpa Implemented all your suggestions.

src/PostgREST/QueryBuilder.hs
```diff
@@ -95,20 +96,29 @@ createWriteStatement selectQuery mutateQuery isSingle echoRequested
...
-    findRelation s t1 t2 =
-      find (\r -> s == (tableSchema . relTable) r && t1 == (tableName . relTable) r && t2 == (tableName . relFTable) r) allRelations
+    -- Searches through all the relations and returns a match given the parameter conditions.
+    -- Will only find a relation where both schemas are in the PostgREST schema.
+    -- `findRelationByColumn` also does a ducktype check to see if the column name has any variation of `id` or `fk`. If so then the relation is returned as a match.
+    findRelationByTable s t1 t2 =
+      find (\r -> s == tableSchema (relTable r) && s == tableSchema (relFTable r) && t1 == tableName (relTable r) && t2 == tableName (relFTable r)) allRelations
+    findRelationByColumn s t c =
+      find (\r -> s == tableSchema (relTable r) && s == tableSchema (relFTable r) && t == tableName (relFTable r) && length (relFColumns r) == 1 && c `colMatches` (colName . head . relFColumns) r) allRelations
+      where n `colMatches` rc = (cs ("^" <> rc <> "_?(?:|[iI][dD]|[fF][kK])$") :: BS.ByteString) =~ (cs n :: BS.ByteString)
```
ruslantalpa: sorry, that's worse :) in the sense that it's more obfuscated then before
there should be a way here to say "case insensitive match" with a modifier something similar to
```
/_?(id|fk)/i 
```
(not sure what ?: means in your regexp)
This will also allow the possibility of the client specifying the column name in lower case even if in the db it's with uppercase.

test/Feature/QuerySpec.hs
```diff
@@ -224,7 +224,15 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
 
     it "detect relations in views from exposed schema that are based on tables in private schema and have columns renames" $
       get "/articles?id=eq.1&select=id,articleStars{users{*}}" `shouldRespondWith`
-        [str|[{"id":1,"articlestars":[{"users":{"id":1,"name":"Angela Martin"}},{"users":{"id":2,"name":"Michael Scott"}},{"users":{"id":3,"name":"Dwight Schrute"}}]}]|]
+        [str|[{"id":1,"articleStars":[{"users":{"id":1,"name":"Angela Martin"}},{"users":{"id":2,"name":"Michael Scott"}},{"users":{"id":3,"name":"Dwight Schrute"}}]}]|]
+
+    it "can select by column name" $
+      get "/projects?id=in.1,3&select=id,name,client_id,client_id{id,name}" `shouldRespondWith`
+        "[{\"id\":1,\"name\":\"Windows 7\",\"client_id\":1,\"client_id\":{\"id\":1,\"name\":\"Microsoft\"}},{\"id\":3,\"name\":\"IOS\",\"client_id\":2,\"client_id\":{\"id\":2,\"name\":\"Apple\"}}]"
```
ruslantalpa: Use here the [str||] thing, makes it much more readable

src/PostgREST/QueryBuilder.hs
calebmer: `?:` just means non matching, it's a nice to have I can take it out if its confusing. I didn't use the case insensitive flag because we need to concatenate the relation column with the regular expression to check the ending. With the flag the entire thing becomes case insensitive. This might not matter, but I figured better safe then sorry. Also, I don't know how to do flags in Haskell so someone with more experience would need to tell me what to do ☺️

ruslantalpa: The last thing about lower case, it might not work because in other places we look for tables/columns by their exact name.
Maybe in another PR we should add the possibility of requesting the column  as "authorName" even if in the db it's like "authorname"

test/Feature/QuerySpec.hs
calebmer: Ok, did a quick regular expression find/replace to change this everywhere in the tests.

test/Feature/QuerySpec.hs
```diff
@@ -208,23 +208,31 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
...
+        [str|[{"user_id":2,"task_id":6,"comments":[{"content":"Needs to be delivered ASAP"}]}]|]
 
     it "detect relations in views from exposed schema that are based on tables in private schema and have columns renames" $
       get "/articles?id=eq.1&select=id,articleStars{users{*}}" `shouldRespondWith`
-        [str|[{"id":1,"articlestars":[{"users":{"id":1,"name":"Angela Martin"}},{"users":{"id":2,"name":"Michael Scott"}},{"users":{"id":3,"name":"Dwight Schrute"}}]}]|]
+        [str|[{"id":1,"articleStars":[{"users":{"id":1,"name":"Angela Martin"}},{"users":{"id":2,"name":"Michael Scott"}},{"users":{"id":3,"name":"Dwight Schrute"}}]}]|]
+
+    it "can select by column name" $
+      get "/projects?id=in.1,3&select=id,name,client_id,client_id{id,name}" `shouldRespondWith`
+        [str|[{"id":1,"name":"Windows 7","client_id":1,"client_id":{"id":1,"name":"Microsoft"}},{"id":3,"name":"IOS","client_id":2,"client_id":{"id":2,"name":"Apple"}}]|]
```
begriffs: Is this just demonstrating a corner case? It seems odd that the `client_id` key is duplicated. I guess it's allowed by the JSON standard but discouraged. Many libraries meant to handle JSON will be confused by it.
https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object

test/Feature/QuerySpec.hs
```diff
@@ -208,23 +208,31 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
...
+
+    it "can select by column name sans id" $
+      get "/projects?id=in.1,3&select=id,name,client_id,client{id,name}" `shouldRespondWith`
+        [str|[{"id":1,"name":"Windows 7","client_id":1,"client":{"id":1,"name":"Microsoft"}},{"id":3,"name":"IOS","client_id":2,"client":{"id":2,"name":"Apple"}}]|]
```
begriffs: This test response, on the other hand, makes sense because `client` is not the same key as `client_id`.

test/Feature/QuerySpec.hs
calebmer: In my mind this is valid JSON and if a client can't parse it then it's not following the standard (you also see this pattern when people write unofficial comments in JSON using `"//": …, "//": …`). At worst, I do think this is a corner case as most users won't be selecting both fields at once and/or will use the alternative syntax.

Another note is that `select=id,id,id` will return 3 identical ids.

src/PostgREST/QueryBuilder.hs
begriffs: This id suffix business is ad-hoc, I don't care for it. For instance in my blog demo the foreign key from posts to users is called author. You can argue whether that's a particularly good name but foreign keys can be called anything you want. We should think of a way to handle the general case rather than stripping names based on unspoken conventions.

How about a variation on Ruslan's renaming idea? You would specify the foreign key column and the name to use in the embedding. The renaming would be required. For instance

```
/projects?id=in.1,3&select=id,name,client_id,client_id:client{id,name}
```

This would embed whatever `client_id` points at with the name `client`. To embed it with a different name, like `volunteer`, you'd do this

```
/projects?id=in.1,3&select=id,name,client_id,client_id:volunteer{id,name}
```

Even so, this proposal of embedding with column names doesn't apply to embedding through join tables right? The disparity isn't good.

begriffs:
Here are my general thoughts. I think you've got the right approach, i.e. we should be doing embedding by specifying the foreign key column(s) involved. This solves both the problem of compound foreign keys and of multiple, ambiguous keys like billing_address_id and shipping_address_id on an order. It's not enough to tell it to embed "address".

The final, high quality solution will require two things

* The ability to specify multiple columns in a single embedding to accommodate compound foreign keys. I know these are rare and I'm "taking the bear's side" but the interface is not complete and does not fully support the relational database model without this ability.
  * Not sure of the syntax in the query string. Maybe `/table?select=(fk_a,fk_b):name{*}`
* The ability (necessity even?) of choosing a name for the embedded resource. Tables and foreign keys can be named anything. The only way to prevent name collisions is allowing the client to choose the name for the embedding.
  * Should we try to help out here by choosing defaults? For instance stripping `_?[Ii][Dd]` or choosing the name of the destination table if a client does not specify the embedding name? I'd argue against it.
  * The renaming convention can't always work. Do you try the id stripping, then fall back to the destination name, then throw a 400 error?
  * I suggest *requiring* the client to choose an embedding name. It's clearer what's going on.

In fact I'd say we should be embedding through column names alone and remove the interface where you specify other table names in the select parameter. The only catch is I don't know how to do N-N embedding this way.

calebmer:
I think I agree on the suffixes. I included it because I thought it might ease some earlier problems with this proposal and to see how others felt about it. I also agree with the multiple column embedding and I do like your proposed syntax.

I don't agree with requiring a name however. If a developer really wanted a clean syntax (like myself) they could create a view with the foreign key column `author` in the hopes of requesting `author{*}`. Forcing a rename makes this design decision impractical. The only time where we should force a rename is a compound key (and even then we could use the referenced table name as an intelligent default).

If we are going to allow column renaming at all, we shouldn't care about column name collisions.

Finally, should we also consider how column renaming changes how we approach schema generation?

ruslantalpa:
In short, i think this PR should be merged exactly as it is (with _id thing included) :)

**About specifying an embed once using a fk column and in other cases using table name**:
I think it's a bad idea to remove embedding by specifying table name (especially because you have no solution for n-n)
in this request
```
/projects?select=id,name,client{id,name},tasks{id,name}
```
where client is a fk and tasks is a table i think it makes perfect sense for them to mean different things because they are embeding different type or relations, a foreign key points to a single items, while table points to a collection/array. In addition, this is exactly the shape of the json (the name of the keys) you want to get back. By looking at this request you can clearly see how the json will look.

Consider the case where you would embed by only specifying fk, look at how you would wright the same query
```
/projects?select=id,name,client{id,name},project_id:tasks{id,name}
```
by looking at it it's unclear how the json will look, also notice that the fk project_id that is from tasks table, leaks out of the table and it seems as it's a column from projects table, i find this query very confusing/ugly.

ruslantalpa:
In short i think this PR should be merged as it is.

**Specifying the embeding using only fk**
It's a bad idea, you are considering how an embed looks when referring to parent object, you forgot about child objects and you do not have a solution for embeding n-n
Consider this query
```
/projects?select=id,name,client{id,name},project_id:tasks{id,name}
```
notice how you are forced to specify :tasks which again has to be the table name (and not only a rename) because you do not know from what table to get that key (theyre could be multiple child tables that have the exact same key) so what does this accomplish?
Also notice how the "projects_id" key leaks out of the tasks table and makes it seem like its a column in projects table
The current query looks much nicer and it gives you a clear idea of how the json will look and all the columns are grouped by table.
```
/projects?select=id,name,client{id,name},tasks{id,name}
```
It is ok to embed things some times by specifying fk and some times by specifying tables because they embed different types of relations.
A fk points to a specific item, thus you get a {}, a table means a collection, thus you get a [].


**Using ..._id ..._fk ...Id ...Fk**
It's the conventions used by most of the devs/dba. I don't see why adding this case that coveres most of the situations and makes life easy could not be supported.
This ```client{id,name}``` says:
```Embed here the related object specified by the FK "client"```
If we add the _id thing, the only thing that changes is
```Embed here the related object specified by the FK "client", and by the way, the FK could actually be named client_id but i still want it back as client```
By not supporting this, you will be forcing most of the queries to use the "rename" feature or force the dba to name the FK the way the client expects them
I think this ```client{id,name}``` looks much nicer then ```client_id:client{id,name}```.


**Supporting the rename feature**
Sure, we need it for cases where the fk is not like client_id but something else or when the db uses my_table_name and the frontend wants it like myTableName but this can be done in a separate PR and does not infulence the part about how to detect a relation (FK or table), it's simply menat to be  "AS renamed" adition to the sql.
About the format of the rename, i propose we do this ```rename:fk_name{*}```, graphql does it like this, the FK is always close to the ```{}``` object it references (not 100% sure on why do it exactly this way but i don't see any arguments for the other case)


**Compound keys**
I have no good solution for this that i am sure of, maybe jsut the [key_a,key_b] thing i mentioned before, but we are not in a hury to implement this, the cases are rare

**Comment by calebmer** [comment]
To briefly support @ruslantalpa's `_id` etc argument and answer one of @begriffs questions, the `_id` stuff is last resort, so it is the last thing relation detection will attempt.

**Comment by begriffs** [comment]
Also can you please add a note to the docs? http://postgrest.com/api/reading/#embedding-foreign-entities

**Comment by calebmer** [comment]
Rebased & documented.

File: docs/api/reading.md [hunk 5]
```diff
@@ -227,10 +227,11 @@ Content-Range → 0-14/*
 
 ### Embedding Foreign Entities
 
-Suppose you have a `projects` table which references `clients` through
-a foreign key called `client_id`. When listing projects through the
-API you can have it embed the client within each project response.
-For example,
+To help you make fewer requests, PostgREST allows the embedding of
+traditional SQL relationships into a response. Suppose you have a
```
begriffs: Much better intro.

test/Feature/QuerySpec.hs
```diff
@@ -90,25 +90,25 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
       get "/no_pk?a=is.null" `shouldRespondWith`
         [json| [{"a": null, "b": null}] |]
 
-      get "/nullable_integer?a=is.null" `shouldRespondWith` "[{\"a\":null}]"
+      get "/nullable_integer?a=is.null" `shouldRespondWith` [str|[{"a":null}]|]
```
begriffs: Great idea for quasi-quoting to get rid of escape characters. I'll have to remember this one myself.

begriffs:
Thank you for the contribution!

scott113341:
Hi all, I just wanted to say thank you for resolving this!  I was playing around with PostgREST a couple weeks ago and ran into the exact issue described in #354, and it's so awesome to see that it's been fixed in the meantime.  Thank you!

SebAlbert:
Very sensible discussions here and in #354. In the outcome of this, what's now actually the approach to compound foreign keys? (maybe even several compound references to the same table)

I'm sorry I seem to be "the bear", or in a .01% minority (or both), but in schemas that I am using very frequently, there are lots of compound foreign keys, and many times a subset of a foreign key would be yet another foreign key (i.e., part of the primary key of a referenced table would be a foreign key to a third table). I haven't designed those schemas, but I have to work with them without altering them. (And I might also define my very own schemas similarly, as I don't see a fundamental relational design problem in it apart from this question how to deal with it in PostgREST.)

A cumbersome workaround for me might be creating views with all foreign/primary key columns concatenated into one single additional column on each table, but... well, cumbersome and much worse also on the frontend dev's side. I might rather go the way of multiple requests (not using embedding at all) then.

Also, do the foreign keys have to be expressly contained as FK constraints in the database schema for the embedding to work?

ruslantalpa:
@SebAlbert at this moment there is no support for multiple fk, but stay tunned, it should not take very long to support that (there was no embeding at all 4 months ago). For any relation to be detected they have to be defined as FK as you said (views with cocnatenated columns will not work).

ruslantalpa:
@SebAlbert i think i spoke too soon (forgot all the details)
There is some support for compound foreign keys.
https://github.com/begriffs/postgrest/blob/master/test/Feature/QuerySpec.hs#L221
The problem this PR fixed is when one table contained 2 FK pointing to the same column, but if your FKs point to different columns then the relation should be detected, try and see what happens

Issue #414, opened 2015-12-08
Performance regression for GET requests in 0.3+


diogob:
In the version 0.3 some CTEs were introduced to simplify the query structure. But this generates a barrier for the PostgreSQL optimizer and prevents some nodes from collapsing in the query plan tree.
To demonstrate I'll use a query that can be generated using the postgrest_test database.

First we need some test data for the problems only manifest in large relations:
```sql
insert into test.simple_pk 
select s::text, 'testing for k = ' || s from generate_series(1, 100000) s;
```
Then we generate a query with an ORDER BY:
```sql
WITH pg_source AS ( 
	SELECT  test.simple_pk.* 
	FROM  test.simple_pk 
        ORDER BY k
) 
SELECT 
	(SELECT pg_catalog.count(1) FROM (SELECT * FROM pg_source) a ), 
	pg_catalog.count(t), 
	null, 
	array_to_json(array_agg(row_to_json(t)))::character varying 
FROM 
	(SELECT * FROM pg_source LIMIT 1 OFFSET 0) t
```
The above query is generated by a GET request using some filters and a limit 1. The generated plan (with a total cost of 7116 blks in my test env) is:
<img width="525" alt="screen shot 2015-12-08 at 5 49 33 pm" src="https://cloud.githubusercontent.com/assets/20662/11671149/2619f4ae-9dd4-11e5-8d86-5141a7d48bbd.png">

---

## Possible solution

After trying several different formats and different queries, the best compromise for all different scenarios was the query bellow:

```sql
SELECT 
	(SELECT pg_catalog.count(1) FROM ( 
	SELECT  test.simple_pk.* 
	FROM  test.simple_pk) a ), 
	pg_catalog.count(t), 
	null, 
	array_to_json(array_agg(row_to_json(t)))::character varying 
FROM 
	( 
	SELECT  test.simple_pk.* 
	FROM  test.simple_pk 
        ORDER BY k LIMIT 1 OFFSET 0) t;
```

This query does not use CTEs and also drops the **ORDER BY** clause in the subquery used for counting.

The generated plan is:

<img width="538" alt="screen shot 2015-12-08 at 5 56 34 pm" src="https://cloud.githubusercontent.com/assets/20662/11671266/0bbfabac-9dd5-11e5-9c5f-2559c73b5b99.png">

The cost is 1985 in my test env. Much cheaper than the original.

[...]

PR #425, opened 2015-12-14
#414 revert to separate count query

ruslantalpa:
For now i went with keeping the "pg_source" CTE and inlining the count query.
Besides counting, it's also used when dealing with CSV.
It's not the final version of the code (i want to remove the wrapQuery... functions, at this point i think they are in the way, it made sense when it was only one) but i wanted to PR this so that @diogob can check the performance of the generated query. @diogob please also look at the performance when the requested response needs to be in CSV format
If the query is good, then i will clean up a bit and do another commit here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix #396 include records with missing parents - @ruslantalpa
 - `pgFmtIdent` always quotes #388 - @calebmer
 - Default schema, changed from `"1"` to `public` - @calebmer
+- #414 revert to separate count query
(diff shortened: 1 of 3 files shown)


diogob:
@ruslantalpa this PR solves my problem. I tested with the CSV and the JSON response in the same case that was blocking my upgrade.
Thanks for the fix <3

ruslantalpa:
I think it's ready
I removed some of the nameF  functions and inlined the sql, makes it a bit easier to see how exactly the wrapping sql is being constructed.

begriffs:
I'm losing touch with how this part of the code works, but it passes the tests so I'll take your word for it.

PR #426, opened 2015-12-15
Use schema search path to enable real versioning

begriffs:
I discovered a problem with how we are currently implementing versioning. Because we specify a fixed schema on every request it prevents a search path from working. For instance

```sql
create schema one;
create schema two;
create view one.hi as select 'hi';
SET search_path TO two,one;
select * from hi;
 ?column?
----------
 hi
(1 row)

select * from two.hi;
ERROR:  relation "two.hi" does not exist
```

This pull request removes all mention of schemas and adds new middleware `setSchemaSearchPath` to set the search path based on the command line parameter `--schema`. This parameter is now interpreted as a comma separated list like `--schema v2,v1`.

These changes make the new tests in `VersionSpec` pass, but I took things too far removing the considerations of schema from the structure building functions. Other old tests are failing.

### DIFF
--- a/test/Feature/VersionSpec.hs
+++ b/test/Feature/VersionSpec.hs
@@ -1,0 +1,25 @@
+module Feature.VersionSpec where
+
+-- {{{ Imports
+import Test.Hspec
+import Test.Hspec.Wai
+
+import Hasql as H
+import Hasql.Postgres as P
+
+import SpecHelper
+import PostgREST.Types (DbStructure(..))
+-- }}}
+
+spec :: DbStructure -> H.Pool P.Postgres -> Spec
+spec struct pool = around (withApp cfgTestV2Schema struct pool)
+  $ describe "versioning" $ do
+
+  it "hides tables not in v1 or v2 schemas" $
+    get "/private_schema_view" `shouldRespondWith` 404
+
+  it "can see a new view in v2" $
+    get "/new_v2" `shouldRespondWith` 200
+
+  it "can continue to see a route in v1" $
+    get "/items" `shouldRespondWith` 200
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -17,9 +17,7 @@ import qualified Data.Vector             as V
 import           Network.Wai             (Request (..))
 import           Network.Wai.Parse       (parseHttpAccept)
 import           PostgREST.RangeQuery    (NonnegRange, rangeRequested)
-import           PostgREST.Types         (QualifiedIdentifier (..),
-                                          Schema, Payload(..),
-                                          UniformObjects(..))
+import           PostgREST.Types         (Payload(..), UniformObjects(..))
 import           Data.Ranged.Ranges      (singletonRange)
 
 type RequestBody = BL.ByteString
@@ -30,7 +28,7 @@ data Action = ActionCreate | ActionRead
             | ActionInfo   | ActionInvoke
             | ActionUnknown BS.ByteString deriving Eq
 -- | The target db object of a user action
-data Target = TargetIdent QualifiedIdentifier
+data Target = TargetIdent T.Text
             | TargetRoot
             | TargetUnknown [T.Text]
 -- | How to return the inserted data
@@ -75,8 +73,8 @@ data ApiRequest = ApiRequest {
   }
 
 -- | Examines HTTP request and translates it into user intent.
-userApiRequest :: Schema -> Request -> RequestBody -> ApiRequest
-userApiRequest schema req reqBody =
+userApiRequest :: Request -> RequestBody -> ApiRequest
+userApiRequest req reqBody =
   let action = case method of
                  "GET"     -> ActionRead
                  "POST"    -> if isTargetingProc
@@ -88,10 +86,8 @@ userApiRequest schema req reqBody =
                  other     -> ActionUnknown other
       target = case path of
                  []            -> TargetRoot
-                 [table]       -> TargetIdent
-                                  $ QualifiedIdentifier schema table
-                 ["rpc", proc] -> TargetIdent
-                                  $ QualifiedIdentifier schema proc
+                 [table]       -> TargetIdent table
+                 ["rpc", proc] -> TargetIdent proc
                  other         -> TargetUnknown other
       payload = case pickContentType (lookupHeader "content-type") of
         Right ApplicationJSON ->
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -71,7 +71,7 @@ app dbStructure conf reqBody req =
 
   case (iAction apiRequest, iTarget apiRequest, iPayload apiRequest) of
 
-    (ActionRead, TargetIdent qi, Nothing) ->
+    (ActionRead, TargetIdent tTable, Nothing) ->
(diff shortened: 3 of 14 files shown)


ruslantalpa:
A scenario of what to test
https://gist.github.com/ruslantalpa/4300b3b196102d114ae9
if anyone has more ideas we can add them in that doc

ruslantalpa:
Maybe @begriffs you can copy/paste my gist and create a new issue with it, and while discussing it, you would "edit" the original post to include the comments

begriffs:
Note from the chatroom. We can find the schema in which the search_path resolves a database object.

```sql
select nspname
from pg_catalog.pg_namespace as n
where n.oid = (
  select relnamespace
  from pg_catalog.pg_class
  where relfilenode = to_regclass(?)
);
```

I noticed the computed column tests all fail because `table.column` does not resolve and appears to require full qualification as `schema.table.column`. If needed we can supply that qualification using the query above.

begriffs:
I was wrong about computed columns needing schema qualification. The fact was the functions were still in the public schema which is no longer in the schema search path on this branch.

ruslantalpa:
I've beed thinking about this last night and maybe the current way is better then the proposed one by this PR (i.e. we don't actually want searchpath to work ).
Consider this:
The docs suggest that you do versioning using schemas & views, this way you can decouple a bit your data model from how the api looks and when you have a change, you create a new "version schema".

Suppose all your tables live in "base" schema, and the version schemas are v1,v2,v3 and they all contain view that reference tables from base.
What does it mean to introduce v4, there are two options here
- define copies for all the views from v3 with the applied changes (current way)
- define only the views that changed (PR way using searchpath)

when you would look at a db like this it will look like
a)
```
base
  t1
  t2
v1
  tv1
  tv2
v2
 tv1 -- same definition as in v1
 tv2 --the difinition changed a bit
```

b)
```
base
  t1
  t2
v1
  tv1
  tv2
v2
 tv2 --the difinition changed a bit
```

I would argue that (a) looks better, less confusing. Debugging what is going on in (b) would be very hard (for postgrest user), he will have to go through all the version schemas to track actually what view is being "touched" by the request.

Based on this logic, i propose we stick to the current way. Postgrest exposes one and only one schema only, makes it easy to reason about (both for us and users).

calebmer:
@ruslantalpa but the user can still choose the A upgrade path with this PR. I don't think I'll ever really use this feature (say that now…). But if it can be implemented with little technical debt, and it provides some benefits to some more technically inclined users, I don't see why not.

I do worry about security however. Do we know for sure tables like `pg_roles` can't be accessed? Can we maybe add a test for this?

ruslantalpa:
That's just the thing, it's not a little technical debt, it's a lot. And yes, the db admin has to be much more careful not to expose private data wile in the current version he knows, stuff under schema v1 is exposed.

SebAlbert:
As a new user of PostgREST, I already struggled with the idea that I'll have to create views inside one single exposed schema for everything I want to be accessible, or, alternatively, start one PostgREST instance for each schema and therefore use separate TCP ports for objects from different schemata in my frontend. So I would basically appreciate multiple schemas exposed simultaneously like this.

(Then again, if you wanted to be really general, in my ad-hoc reasoning we would need a way to qualify our RESTful requests with an actual schema. This could mean an even more breaking change such as GET /schemaname/tablename instead of GET /tablename, or with GET /tablename as unqualified fallback, probably using a search path. However, I could of course also put some nginx rewriting in front of several PostgREST instances.)

I don't think we should completely restrict the user from something just because it can be potentially harmful if used inappropriately. With the default setting only exposing one specific schema, and most reasonable db admins creating a separate role (with no default rights on any table/view/function etc.) for an exposed API, I don't think that's a cause.

calebmer:
@SebAlbert well I was on the fence, but you just changed my mind 😊

It's really hard when making decisions to put yourself in the shoes of another developer working with PostgREST.

As for security, I haven't looked at the code yet (I'll do that as soon as possible), but my initial idea was to improve the `QualifiedIdentifier` system (I did a PR on that a bit ago which could be really helpful now) and then simply use the schemas specified in the parameter + our generated `DbStructure` to select all the things. It has the advantage over `search_path` in that we explicitly know what can be selected at all times and the code will be clearer. However it will also be harder to implement.

ruslantalpa:
@calebmer @SebAlbert 
My main reason against this is the code complexity for us, i think it will become unmanageable and hard to implement and we will hit a lot of corner cases. Also the users will have a hard time understanding what is actually exposed, not everyone is a dba with 10 years of experience, and most of us come from mysql which does not even have a schema concept.

Do we really want to go down this road http://petereisentraut.blogspot.ro/2009/07/schema-search-paths-considered-pain-in.html (look on the right side who this guy is)

calebmer:
That article doesn't exactly apply to our use case. We care more about the application of cascading schema selection in an APIs design and not the literal PostgreSQL implementation of `search_path` and its troubles in SQL land.

As for code complexity, the code surrounding schemas at the moment is unnecessarily complex (this was the argument I made back in #379). Theoretically this feature is just replacing a schema string with a schema array in a few places, not that complex. But because of the code's current complexity we must address some more fundamental problems first.

test/Feature/VersionSpec.hs
```diff
@@ -0,0 +1,25 @@
+module Feature.VersionSpec where
```
calebmer: I think this should be called `SchemaSpec` instead. It more accurately represents what the feature is vs the problem the feature aims to solve.

calebmer:
I just looked at the code. I like how it simplifies things in removing schemas everywhere, but I have this nagging feeling that doing this is going to bite us in the future. What's the likeliness at some point in the future we revert this and add schemas back, I think that's relatively high.

I'm for this feature, but I'm against the `search_path` implementation.

begriffs:
I feel like using the search path is a good idea if used hygienically.

* It makes specifying changes to an API easier. You don't have to copy all views over to the new version. 
* If you refactor a view, to make it run faster say, you don't have to propagate the improvement to other copies.
* I'm not seeing why it's hard to debug, but would appreciate an example.
* (appeal to authority:) The Postgres community opted to support the notion of a search path and they are deliberate people.

In the *Schema Search Paths Considered a Pain in the Butt* the author suggests

> The way to deal with this, to the extent that it is dealt with at all, is to either set the schema search path everywhere before accessing any object (for some value of everywhere, such as every script, every session, every function) or to explicitly schema-qualify every object reference.

...and this is actually what we are doing, by using a `SET LOCAL search_path = ...` on every request. The full generality of search paths can be complex, including the default top schema on the path being the user's role name. I don't think those criticisms apply to our more carefully managed situation.

Our code complexity and tech debt is another matter. My gut feeling is that our code has evolved to be more complicated than it has to be -- whether or not we opt to use the search path. That perception may be because I haven't taken the time to understand the full meaning of some of the things I've merged in the `DbStructure` and `QueryBuilder` modules. You guys have been more active contributing there so you have a better idea.

begriffs:
This code is now way out of date due to the hasql 19 upgrade.

SebAlbert:
Too bad. I hope the feature is still going to happen.

begriffs:
Yeah I still want to make it happen. It's just that the code in this PR was an experiment and I think we can do it better.

PR #427, opened 2015-12-16
Fix errors in tests when setting authenticator to noinherit #407

ruslantalpa:
What was going on is this
Range spec was expecting the "menagerie" table to be empty, but resetDb was not doing that (not all tables in fixture had truncate statements).

The reason it worked before is that because authenticator role inherited all the privileges from anonymous, it had the right to "clearTable" that was being executed after each test in "InsertSpec"

### DIFF
--- a/test/Feature/InsertSpec.hs
+++ b/test/Feature/InsertSpec.hs
@@ -23,7 +23,7 @@ import TestTypes(IncPK(..), CompoundPK(..))
 spec :: DbStructure -> H.Pool P.Postgres -> Spec
 spec struct pool = beforeAll_ resetDb $ around (withApp cfgDefault struct pool) $ do
   describe "Posting new record" $ do
-    after_ (clearTable "menagerie") . context "disparate csv types" $ do
+    context "disparate csv types" $ do
       it "accepts disparate json types" $ do
         p <- post "/menagerie"
           [json| {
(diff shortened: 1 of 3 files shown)


begriffs:
That's tricky. Good find.

I notice a lot of `after_ (clearTableAre ...)` are removed. Are you worried about test pollution between specs if we don't clear table modifications after each?

ruslantalpa:
The resetdb is curently doing that, they are not needed anymore

begriffs:
In InsertSpec, `resetDb` is set as `beforeAll` so it won't clean things up between each test, only before the entire suite runs. The call to `clearTable` was meant to stop tests from polluting others in the same suite. (Could call `resetDb` as `before` rather than `beforeAll` but it would run slower.)

So we should probably grant the test role access to truncate tables and keep the clearTables in the InsertSpec. But keep your truncate statements in data.sql too.

ruslantalpa:
I understand their purpose its just that for now they dont polyte each other, they work with different tables. When the time comes and one of them polytes the data for another one, a cleanTtable will be introduced in that place, for now it is not needed

begriffs:
Sure we can merge this. I was just pointing out that the best practice would be for all our tests to clean up after themselves. These tests don't interfere with each other's data and are currently OK but they're not setting a good example of cleanup.

Actually an ideal situation would be to run each test in a transaction and roll it back afterwards. It's complicated because postgrest also opens a transaction for each request so we would probably need to set a savepoint to deal with nested transactions.

Something to think about for later.

PR #433, opened 2015-12-20
Consistent read query escaping

calebmer:
This is a simple bug fix. As I camel case my APIs PostgREST was generating bad queries which look like this:

```sql
SELECT
  "api"."posts".*,
  row_to_json(postTopics.*) AS "postTopic"
FROM
  "api"."posts"
LEFT OUTER JOIN (
  SELECT
    "api"."postTopics".*
  FROM
    "api"."postTopics"
) AS postTopics
ON
  "postTopics"."id" = "posts"."postTopicId";
```

Note how some of the `postTopics` are not quoted. This is because those values are not run through `pgFmtIdent`. This caused an error as `posttopics` and `"postTopics"` do not match. This is a quick PR to fix that.

### DIFF
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -387,3 +387,18 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
       it "returns proper json" $
         post "/rpc/sayhello" [json| { "name": "world" } |] `shouldRespondWith`
           [json| [{"sayhello":"Hello, world"}] |]
+
+  describe "weird requests" $ do
+    it "can query as normal" $ do
+      get "/Escap3e;" `shouldRespondWith`
+        [json| [{"so6meIdColumn":1},{"so6meIdColumn":2},{"so6meIdColumn":3},{"so6meIdColumn":4},{"so6meIdColumn":5}] |]
+      get "/ghostBusters" `shouldRespondWith`
+        [json| [{"escapeId":1},{"escapeId":3},{"escapeId":5}] |]
+
+    it "will embed a collection" $
+      get "/Escap3e;?select=ghostBusters{*}" `shouldRespondWith`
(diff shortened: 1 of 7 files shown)


begriffs:
Good catch. Unquoted identifiers make me worried about sql injection too.

Can you add a test for this change?

calebmer:
Sure. I didn't initially because it would involving adding a couple tables.

calebmer:
Sorry for taking so long, I had other priorities, but the tests are added. It's a couple of weird corner cases that include the error I was hitting. We can expand that suite as we find more corner cases.

test/Feature/QuerySpec.hs
```diff
@@ -382,3 +382,18 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
       it "returns proper json" $
         post "/rpc/sayhello" [json| { "name": "world" } |] `shouldRespondWith`
           [json| [{"sayhello":"Hello, world"}] |]
+
+  describe "weird requests" $ do
+    it "can query as normal" $ do
```
begriffs: This test passes with the code on master, is this what you would expect?

test/Feature/QuerySpec.hs
```diff
@@ -382,3 +382,18 @@ spec struct pool = around (withApp cfgDefault struct pool) $ do
...
+      get "/Escap3e;" `shouldRespondWith`
+        [json| [{"so6meIdColumn":1},{"so6meIdColumn":2},{"so6meIdColumn":3},{"so6meIdColumn":4},{"so6meIdColumn":5}] |]
+      get "/ghostBusters" `shouldRespondWith`
+        [json| [{"escapeId":1},{"escapeId":3},{"escapeId":5}] |]
+
+    it "will embed a collection" $
```
begriffs: This test passes with the code on master.

calebmer:
Yes for both tests. The last one is really the issue I was hitting, but I figured as long as I was writing tests I could write some safeguards for weird escaping.

begriffs:
Can you rebase please? I think the changelog has a merge conflict because Ruslan's change also adds an "Unreleased" section. Next time I make a new release I'll add an empty unreleased section ahead of time so that the next two pull requests don't conflict.

calebmer:
Done, sorry again for taking so long.

PR #435, opened 2015-12-22
Document python requests jwt

davidthewatson:
As discussed with @begriffs, this example documents:

* logging into a PostgREST API using python-requests
* obtain token from response
* use token auth with requests-jwt
* use Range header for pagination of results

### DIFF
--- a/docs/examples/python-requests-jwt.md
+++ b/docs/examples/python-requests-jwt.md
@@ -1,49 +1,40 @@
-## Accessing PostgREST API using Python Requests and Requests-JWT
+## Python Client for PostgREST API
 
-This code relies on setting up the PostgreSQL auth functions and grants correctly first. Follow these instructions:
+### Setup PostgreSQL
 
-http://postgrest.com/examples/users/
+This code relies on setting up the PostgreSQL auth functions and grants correctly first. Follow [these instructions](http://postgrest.com/examples/users/).
 
-After completing the configuration, be sure to create a user with email, password, role, and verified flag. We'll use that user to login in the code below.
+After completing the PostgreSQL configuration, be sure to create a user with email, password, role, and verified flag. We'll use that user to login in the code below.
 
-Install the required libraries
+### Setup PostgREST
 
-    pip install requests
-    pip install requests-jwt
+Next, setup PostgREST according to the documentation [http://postgrest.com/install/server/](here).
 
-Then, from a python interpreter or script:
+### Setup Python Client
 
-    import requests
-    import requests_jwt
-    import json
+Finally, we'll install and configure the python client. Follow the instructions in the [README](https://github.com/davidthewatson/postgrest_python_requests_client/blob/master/README.md). Be sure to set the [credentials](https://github.com/davidthewatson/postgrest_python_requests_client/blob/master/config.in#L3-L5) and [urls](https://github.com/davidthewatson/postgrest_python_requests_client/blob/master/config.in#L7-L9) in config.py.
 
-    # email and password auth through postgrest
-    resp = requests.post('http://localhost:3000/rpc/login', json={"email": "you@yours.com", "pass": "dog"})
-    if resp.status_code != 200:
-        raise Exception()
+### Python Client Functions
 
-    # JWT auth using the token above
-    token = json.loads(resp.text)['token']
-    auth = requests_jwt.JWTAuth(token)
-    if r.status_code != 200:
-        raise Exception()
+There are four primary functions to the python client:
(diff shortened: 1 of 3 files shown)


davidthewatson:
@begriffs I've updated the pull request with usage of the Range header for pagination.

docs/examples/python-requests-jwt.md
```diff
@@ -0,0 +1,49 @@
...
+        raise Exception()
+
+    # JWT auth using the token above
+    token = json.loads(resp.text)['token']
+    auth = requests_jwt.JWTAuth(token)
+    if r.status_code != 200:
+        raise Exception()
+
+    # Make a request to get the range header
+    r = requests.get('http://localhost:3000/posts', auth=auth)
```
begriffs: This might not be a good idea since it would request the entire list of posts which could be big. You can actually request a limited range initially (like 0-9) and the response Content-Range header will still give you the total for all posts.

davidthewatson: I was thinking the same thing, except that my initial thought, intuitively, was: why don't I just make a HEAD request? Then I made the HEAD request, saw that it was neutered, read the HTTP docs at w3, and saw that the behavior with regard to Content-Range and payload was according to the docs and sat there puzzled, since it seemed like the perfect use of a HEAD request, since the count is a lightweight thing and I don't actually need the payload there, or further, may want to do something else if the count is a billion. I didn't actually understand why Fielding et al made that design choice? And why using HEAD this way is not an appropriate choice here?

Could I just make the get request with 0 range?

begriffs: You're right, HEAD seems perfect. The spec forbids sending back Content-Range though?

The smallest range you can request contains one element, for instance `0-0`.

davidthewatson: The spec is pretty confusing in this regard. I've read various versions of the spec and I can't quite find the clarity that I'm seeking. Turns out that the HATEOAS folks and RFC 5988 have produced some thoughtful inquiry on the subject but nothing I'd describe as authoritative.

http://pilhuhn.blogspot.com/2013/02/best-practice-for-paging-in-restful-apis.html

The notion of next and previous links seems usable. I believe that the key issue for clients is: what is the most intuitive way to interact with the API and make decisions about how to proceed through a paginated set? I'm scratching my head wondering how we've gotten this far without a standard for what seem like such rudimentary operations.

davidthewatson: @begriffs I've redone this entire thing minimizing all of the problems with the existing interface here:

https://github.com/davidthewatson/postgrest_python_requests_client

I'd be happy to redo this documentation based on that example. Thoughts?

begriffs: That's really nice, thank you!

* You can add a link to the python client under the "Client Libraries" section of http://postgrest.com/install/ecosystem/
* And then redoing the docs in this pr would be great as well. A nice convention is to add subsections to your doc page to separate the login stuff from the pagination.

davidthewatson: @begriffs I've done both of these in the latest commits:

* You can add a link to the python client under the "Client Libraries" section of http://postgrest.com/install/ecosystem/
* And then redoing the docs in this pr would be great as well. A nice convention is to add subsections to your doc page to separate the login stuff from the pagination.

Thanks for your review!

docs/install/ecosystem.md
```diff
@@ -5,6 +5,8 @@
 * [mithril.postgrest](https://github.com/catarse/mithril.postgrest) - Mithril plugin to create and authenticate requests
 * [lewisjared/postgrest-request](https://github.com/lewisjared/postgrest-request) - node interface to postgrest instances
 * [JarvusInnovations/jarvus-postgrest-apikit](https://github.com/JarvusInnovations/jarvus-postgrest-apikit) - Sencha framework package for binding models/stores/proxies to PostgREST tables
+*
```
begriffs: Not sure if the newline here will mess up the list, maybe combine this and the next line.

begriffs:
Nicely organized!

One final thing is to add your new page into the menu system of the docs page by modifying https://github.com/begriffs/postgrest/blob/master/mkdocs.yml

docs/install/ecosystem.md
davidthewatson: Good catch! My bad. Fixed.

davidthewatson:
@begriffs I killed the extraneous newline and added the example link to the mkdocs.yml. Should be good to go!

PR #439, opened 2015-12-25
Fix bug in many-many relation detection

ruslantalpa:


begriffs:
Oh I see, the reverse relation...

PR #450, opened 2016-01-08
docs/examples/users.md: pgcrypto includes uuidv4 generation

daurnimator:
No need to use the uuid-ossp extension

From http://www.postgresql.org/docs/current/static/uuid-ossp.html :
> Note: If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.

`gen_random_uuid` was only added in to postgres in version 9.4. I'm not sure what version the docs are meant to be compatible with?

### DIFF
--- a/docs/examples/users.md
+++ b/docs/examples/users.md
@@ -28,13 +28,11 @@ value.
 ### Storing Users and Passwords
 
 We create a database schema especially for auth information. We'll
-also need the postgres extensions
-[pgcrypto](http://www.postgresql.org/docs/current/static/pgcrypto.html) and
-[uuid-ossp](http://www.postgresql.org/docs/current/static/uuid-ossp.html).
+also need the postgres extension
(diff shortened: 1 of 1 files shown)


begriffs:
Good to know. I must have learned about generating uuids from some outdated docs.

Have you tried this on your machine to double check if it works?

begriffs:
Thanks for updating this part of the docs.

michalskop:
This did not work for me (pgsql vers. 9.5), I got "function uuid_generate_v4() does not exist" error for `/rpc/signup` and had to do
```
CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
create extension if not exists pgcrypto;
```

begriffs:
Thanks for pointing out the problem. Would you mind sending a pull request to fix up the docs?

michalskop:
Done, PR #525

daurnimator:
The solution should be to update to use `gen_random_uuid`

PR #454, opened 2016-01-11
fix #453 by sending the request to the db (instead of using the cached data)

### LINKED ISSUE #453: Anonymous users can see schema structure they shouldn't
Making a GET request to the root path shows tables that the anonymous user does not have access to read or write.

Steps to reproduce:

```bash
createdb perms
psql -d perms
```

then in psql
```sql
create role perm_auth noinherit login;
create role perm_anon;
create role perm_user;
create table festival
(
  name text not null primary key
);
grant perm_anon to perm_auth;
grant perm_user to perm_auth;
grant usage on schema public to perm_user;
```

Start postgrest
```bash
postgrest postgres://perm_auth:@localhost:5432/perms -a perm_anon --schema public
```

At this point when I ask for `/` it gives me back an empty array which is good. Nothing is visible to the `perm_anon` user. Then I go back into psql and execute `grant all on festival to perm_user;` (not `perm_anon`). Requesting `/` now gives me

```json
[
  {
    "schema": "public",
    "name": "festival",
    "insertable": true
  }
]
```

ruslantalpa:
So the history of this is as follows:
In the beginning the / request used to go to the db (like in this pr).
The we started caching the db structure (along the info of who has access to what) and serving the response using that.
Then during discussions about how to model "users/auth" we figured that when someone wants to use 1 user = 1 role, this way of doing things (caching who has access to what) so we started to migrate back and at the same time there was some dbstructure refactoring and things got mixed up.

PS: @begriffs before merging this, can you test in 3.0.3 my suspicion that this bug only appears when the "grant" command is executed during the time postgrest is running and the bug presents itself but when you restart postgrest, it works again. If that is correct then the docs should say somewhere that on each db structure change, postgrest needs to be restarted

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -178,7 +178,7 @@ app dbStructure conf reqBody req =
         else return notFound
 
     (ActionRead, TargetRoot, Nothing) -> do
-      body <- encode <$> accessibleTables (filter ((== cs schema) . tableSchema) (dbTables dbStructure))
+      body <- encode <$> accessibleTables (cs schema)
       return $ responseLBS status200 [jsonH] $ cs body
 
     (ActionUnknown _, _, _) -> return notFound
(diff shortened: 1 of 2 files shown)


begriffs:
Nice and consolidated, thanks.

Issue #458, opened 2016-01-14
Problem with embedded quotes in JSON


motiz88:
Something appears broken in PostgREST versions 0.2.11.0 through 0.3.0.3 (see version investigations below).

If `"A"` is a `text` column and `"B"` is a `jsonb` column, then:
* PATCHing this works:
```json
{"A": "\"a"}
```

* PATCHing this also works:
```json
{"A": "\"a", "B": {"b": "b"}}
```
* PATCHing this fails with `{
details: "Token \"b\" is invalid.",
code: "22P02",
message: "invalid input syntax for type json"
}`, which AFAICT is a bug:
```json
{"A": "\"a", "B": {"b": "\"b"}}
```

I am using the Ubuntu build in a custom Docker image, doing nothing fancy. The error is the same with both my local Windows Postgres v9.4.3 server and a v9.4.1 server on Amazon RDS.

motiz88:
Another data point is that POST with the exact same content works fine. The issue may be specific to PATCH.

motiz88:
I can now confirm that 0.2.9.0 and 0.2.10.0 work correctly. (0.2.11.0 does not run at all on my setup, for whatever reason.) So the bug was introduced in some commit `0.2.10.0 < x <= 0.2.11.1`.

motiz88:
0.2.11.0 contains the bug, I managed to verify it now. (I just hadn't sent an Accept header.)
Meaning [`0.2.10.0 < x <= 0.2.11.0`](https://github.com/begriffs/postgrest/compare/master@%7B2015-06-03%7D...master@%7B2015-08-28%7D).

motiz88:
From skimming the source code, I believe the bug is likely to be in either [`insertableValue`](https://github.com/begriffs/postgrest/blob/48f9ce114e48c449e808b097042b56ced078a7eb/src/PostgREST/QueryBuilder.hs#L402) or [`pgFmtLit`](https://github.com/begriffs/postgrest/blob/48f9ce114e48c449e808b097042b56ced078a7eb/src/PostgREST/QueryBuilder.hs#L212), but not being fluent in Haskell, I can't really act on this hunch.

motiz88:
[...]

PR #459, opened 2016-01-14
pgFmtLit: Change condition for using E"..." syntax

motiz88:
This may solve #458.
**I'm submitting this PR as a quick way of getting all the tests to run - not sure if it's correct yet!**

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -209,7 +209,7 @@ pgFmtLit x =
  let trimmed = trimNullChars x
      escaped = "'" <> replace "'" "''" trimmed <> "'"
      slashed = replace "\\" "\\\\" escaped in
- if "\\\\" `isInfixOf` escaped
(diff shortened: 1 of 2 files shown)


diogob:
We should add a spec for this change. 
This is the simplest one I could think of (in InsertSpec.hs):
```haskell
      it "can set a column to escaped value" $ do
        _ <- post "/json" [json| { data: {"escaped":"bar"} } |]
        request methodPatch "/json?data->>escaped=eq.bar"
                     [("Prefer", "return=representation")]
                     [json| { "data": { "escaped":" \"bar" } } |]
            matchBody    = Just [json| [{ "data": { "escaped":" \"bar" } }] |]
          , matchStatus  = 200
          , matchHeaders = []
          }
```

begriffs:
Thanks again for debugging the problem and sending a fix.

Can you squash these commits into a single one please? That will help me cherry pick it into the 2.x branch as well.

motiz88:
Done :smile:

Issue #405, opened 2015-12-07
Multiple insert on database


zizou10:
Hello,

I would like to insert in two tables links by one to many or many to many relations with post request.
For example:
I have two tables Subscription & Origin

Subscription is link to origin (one to many relation)
I would like to insert new entry into subscription and origin.

How can I do that ?

Regards

ruslantalpa:
At least for now, inserts/updates work on one table at a time only, meaning you first do a POST on the child (many) table, then do a POST on the parent (one) table.

I think it might be possible to have a before insert trigger on the parent table so that when it notices a value being inserted to a "meta column" it takes that data and first inserts it in the child table then the insert continues on the main table.

ruslantalpa:
Your situation is common (trying to insert data in multiple tables at the same time, for example add a "client" and a few "projects" in one request) so i am interested in adding such functionality in the future or at least provide an easy way/template of how to do this with db triggers

zizou10:
Thank you for your response.

For now, I think the best way is to use sql function which takes a json in entry and insert data into multiple tables.

begriffs:
Sure a stored proc can do it. Another option is making an [auto-updatable](http://www.postgresql.org/docs/9.4/static/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS) join view. A simple join between the parent and child tables will be updatable so you can write to it through postgrest as if it were a table.

begriffs:
@zizou10 have you tried any of these ideas and do they work for you OK?

zizou10:
@begriffs I try the version with Stored Procedure function but I had trouble to call the function.
I follow your tutorial but I have an error (couldn't find my function)

Can you give me an example please ?

Regards

begriffs:
Which database schema contains your stored procedure? And what value are using for the command line option `--schema` when starting the postgrest server?

zizou10:
I have this file postgrest to configure the postgrest service:

POSTGREST_DBPORT=5432

# database to use
POSTGREST_DBNAME=zizou

# database user
POSTGREST_DBUSER=zizou

# database password
POSTGREST_DBPASS=pass

# anonymouse user
POSTGREST_ANON=zizou

# database pool
POSTGREST_POOL=10

# jwt secret
POSTGREST_JWT_SECRET=secret

# default schema
POSTGREST_SCHEMA=public

# secure (use 1 to enable, empty string to disable)
POSTGREST_SECURE=None

[...]

PR #466, opened 2016-01-21
Adds section about multiple insertion/update to api/writing docs [#405]

diogob:


begriffs:
Thank you, docs are so needed. I like the use of the design consideration box as well. :smile:

PR #468, opened 2016-01-21
Rename endpoint for consistent endpoint names

guido4000:
In line 126 the endpoint url is api_token. Therefore it should be the same in line 178.


begriffs:
Good catch, much appreciated!

PR #472, opened 2016-01-22
Upgrade to Hasql 19

begriffs:
Let the code review begin! There are elements of code style and of broken functionality that need help.

Note that this compiles but that all client requests fail with the error message, "another command is already in progress."

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -10,9 +10,15 @@ module PostgREST.DbStructure (
 , doesProcReturnJWT
 ) where
 
+import qualified Hasql.Query             as H
+import qualified Hasql.Encoders          as HE
+import qualified Hasql.Decoders          as HD
+
 import           Control.Applicative
 import           Control.Monad          (join)
+import           Data.Functor.Contravariant (contramap)
 import           Data.Functor.Identity
+import           Text.InterpolatedString.Perl6 (q)
 import           Data.List              (elemIndex, find, subsequences, sort, transpose)
 import           Data.Maybe             (fromMaybe, fromJust, isJust, mapMaybe, listToMaybe)
 import           Data.Monoid
@@ -42,24 +48,36 @@ getDbStructure schema = do
     , dbPrimaryKeys = keys'
     }
 
-doesProc :: forall c s. B.CxValue c Int =>
-            (Text -> Text -> B.Stmt c) -> QualifiedIdentifier -> H.Tx c s Bool
-doesProc stmt qi = do
-  row :: Maybe (Identity Int) <- H.maybeEx $ stmt (qiSchema qi) (qiName qi)
-  return $ isJust row
+encodeQi :: HE.Params QualifiedIdentifier
+encodeQi =
+  contramap qiSchema (HE.value HE.text) <>
+  contramap qiName   (HE.value HE.text)
 
-doesProcExist :: QualifiedIdentifier -> H.Session Bool
-doesProcExist = doesProc [H.stmt|
+decodeTable :: HD.Result Table
+decodeTable =
+  HD.singleRow standardRow
+ where
+  standardRow = Table <$> HD.value HD.text <*> HD.value HD.text
+                      <*> HD.value HD.bool
+
+doesProcExist :: H.Query QualifiedIdentifier Bool
+doesProcExist =
+  H.statement sql encodeQi (HD.singleRow (HD.value HD.bool)) True
+ where
+  sql = [q| SELECT EXISTS (
       SELECT 1
       FROM   pg_catalog.pg_namespace n
       JOIN   pg_catalog.pg_proc p
       ON     pronamespace = n.oid
       WHERE  nspname = ?
       AND    proname = ?
-    |]
+    ) |]
 
-doesProcReturnJWT :: QualifiedIdentifier -> H.Session Bool
-doesProcReturnJWT = doesProc [H.stmt|
+doesProcReturnJWT :: H.Query QualifiedIdentifier Bool
+doesProcReturnJWT =
+  H.statement sql encodeQi (HD.singleRow (HD.value HD.bool)) True
+ where
+  sql = [q| SELECT EXISTS (
       SELECT 1
       FROM   pg_catalog.pg_namespace n
       JOIN   pg_catalog.pg_proc p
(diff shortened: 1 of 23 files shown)


begriffs:
@nikita-volkov I'm having trouble understanding the database client error I keep receiving, "another command is already in progress." Am I not allowed to have more than one `sql` function per `Session` or to mix `sql` and `query` in a `Session`?

nikita-volkov:
Do you use concurrency anywhere there? If not, then I suspect it's a bug in Hasql's implementation of `sql`.

begriffs:
I don't think there's concurrency. When the server starts up it creates a pool (from the `resource-pool` package) like this

```haskell
  pool <- createPool (H.acquire pgSettings)
            (either (const $ return ()) H.release) 1 1 (configPool conf)
```

Then when a new request comes in it tries to grab a connection out of the pool and run the handler which is in a Session (`Request -> H.Session Response`). The handler ends up calling some functions that call `sql` and `query` a few times. I'm seeing the command-in-progress error right away on the first request.

One thing to note is that before serving requests postgrest runs a one-time query to determine the database structure. It does this within another `withResource` so that connection should be properly released and not interfere with the subsequent connection.

I can produce SQL logs if it helps. Thank you for helping to look into this.

nikita-volkov:
Okay. I'll test it out on the Hasql side this weekend and post back here.

BTW, I maintain a Hasql-specialized wrapper over "resource-pool" for convenience: ["hasql-pool"](http://hackage.haskell.org/package/hasql-pool). I must advise you against using the "Hasql.Pool.Handle" module there though, since it's experimental and I'll probably remove it in the next release.

begriffs:
@nikita-volkov I created a simplified program which illustrates the problem https://gist.github.com/begriffs/8eeb25597064fdfe6363

ruslantalpa:
@begriffs looked over the code, look's good, the logic stayed the same, you changed the types to accommodate the new version of hasql.  The use of ```Text.InterpolatedString.Perl6``` looks good.
:+1:

nikita-volkov:
@begriffs Thanks for the reproduction - it's part of the testsuite now. The thing should be fixed in the latest release.

begriffs:
Thanks for the super fast Hasql fix!

The app appears to work well when I manually send requests. But the tests are still a problem. The first test works but subsequent ones cause the already-in-progress error. However here we might have real concurrency if each test is starting a web server which listens for a request on a new thread. All the specs are sharing the same connection.

nikita-volkov:
You can't issue multiple requests on a single connection concurrently. Consider using a pool in this case.

src/PostgREST/DbStructure.hs
```diff
@@ -10,28 +10,32 @@ module PostgREST.DbStructure (
...
 
-getDbStructure :: Schema -> H.Tx P.Postgres s DbStructure
+getDbStructure :: Schema -> H.Session DbStructure
 getDbStructure schema = do
-  tabs <- allTables
-  cols <- allColumns tabs
-  syns <- allSynonyms cols
-  rels <- allRelations tabs cols
-  keys <- allPrimaryKeys tabs
+  tabs <- H.query () $ allTables
```
diogob: The pattern ```H.query ()``` appears in several places, maybe we should define a ```staticQuery``` as
```haskell
staticQuery = H.query ()
```

src/PostgREST/DbStructure.hs
```diff
@@ -44,24 +48,78 @@ getDbStructure schema = do
...
+  tblRow = Table <$> HD.value HD.text <*> HD.value HD.text
+                 <*> HD.value HD.bool
+
+decodeColumns :: [Table] -> HD.Result [Column]
+decodeColumns tables =
+  mapMaybe (columnFromRow tables) <$> HD.rowsList colRow
+ where
+  colRow =
+    (,,,,,,,,,,)
+      <$> HD.value HD.text <*> HD.value HD.text
```
diogob: I don't understand this part of the code very well are there any docs on the new hasql version?
Besides, maybe it would be more readable if we defined ```colRow``` in a higher level:
```haskell
textDecoder = HD.value HD.text
intDecoder = HD.value HD.int4
nullableIntDecoder = HD.nullableValue HD.int4
nullableTextDecoder = HD.nullableValue HD.text
```

src/PostgREST/QueryBuilder.hs
```diff
@@ -57,70 +60,101 @@ import           Data.Scientific         ( FPFormat (..)
...
 createWriteStatement qi _ mutateQuery isSingle HeadersOnly
-                     pKeys _ (PayloadJSON (UniformObjects rows)) =
-  B.Stmt (
-    "WITH " <> sourceCTEName <> " AS (" <> mutateQuery <> " RETURNING " <> fromQi qi <> ".*" <> ") " <>
-    "SELECT " <> intercalate ", " [
+                     pKeys _ (PayloadJSON (UniformObjects _)) =
+  H.statement sql encodeUniformObjs decodeStandardMay True
+ where
+  sql = [qc|
+      WITH {sourceCTEName} AS ({mutateQuery} RETURNING {fromQi qi}.*)
```
diogob: Nice! :sunglasses:

begriffs:
Here's where we are now.

* The tests work when I run each individually!
* Still have that already-in-progress error
    * Interestingly, [one of the tests](https://github.com/begriffs/postgrest/blob/hasql-19/test/Feature/InsertSpec.hs#L359-L366) raises this same error when run singly
    * Maybe it's a clue into the conditions causing the error
* Continuous integration fails long before running the tests
    * It has a cabal dependency conflict (even when I rebuild without cache)
    * I'm wrestling with the question: should we use Stack in CI like [this](https://gist.github.com/dbp/bef96402ea07001dfed2)?
    * The advantage is a consistent build with what we do in development
    * The disadvantage is it could lead to poorer support for people who do not use Stack and who want to use PostgREST as a library in their projects
* I should switch to hasql-pool as Nikita suggests

begriffs:
@nikita-volkov I found another situation that causes the in-progress error and I made a gist out of it. https://gist.github.com/begriffs/c632b5e3abe0984d2483

Basically If an error happens during a connection then attempting to run any future commands on that connection will cause the in-progress bug.

nikita-volkov:
@begriffs Thanks again for the reproduction. It's fixed in the latest release.

src/PostgREST/DbStructure.hs
begriffs: The length of the alias names aren't too much shorter than the originals so maybe it's not worth adding more names.

begriffs:
OK it's rebased on master and passing the tests. Maybe give it one last look and I'll merge tomorrow. One source of errors that kept popping up was that I use decoders which expect values to not be null and they could become null. It's conceivable that I missed dealing with some cases like this and that they were not exercised by the tests. The way I dealt with the ones I fixed was to wrap sql terms in

```sql
coalesce(NULLABLE_VALUE, '')
```

Another approach would be to change the decoders for these kind of values to allow nulls and handle the resulting Maybe values in our Haskell code.

PR #476, opened 2016-01-29
Switch to Stack on CircleCI

begriffs:
The new config has a few benefits

* Builds like we do in development
* Removes helpers like hlint and packdeps from the cabal file and into the circle config
* Newer stack resolver
* More kinds of ghc warnings
* Added checks for
    * cabal file warnings
    * haddock coverage
    * problems creating hackage tarball

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,16 +1,21 @@
-machine:
+dependencies:
+  cache_directories:
+    - "~/.stack"
   pre:
+    - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 575159689BEFB442
+    - echo 'deb http://download.fpcomplete.com/ubuntu precise main'|sudo tee /etc/apt/sources.list.d/fpco.list
+    - sudo apt-get update && sudo apt-get install stack -y
     - createuser --superuser --no-password postgrest_test
     - createdb -O postgrest_test -U ubuntu postgrest_test
-  ghc:
-    version: 7.10.1
-dependencies:
   override:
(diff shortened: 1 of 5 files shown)


circle.yml
```diff
@@ -1,16 +1,23 @@
...
+    - stack build --test --no-run-tests
+
 test:
-  post:
-    - cabal exec hlint -- -X QuasiQuotes src/**/*.hs test/**/*.hs
-    - cabal exec packdeps postgrest.cabal || true
+  override:
+    - stack test
+    - git ls-files | grep '\.l\?hs$' | xargs stack exec -- hlint -X QuasiQuotes "$@"
+    - stack exec -- packdeps *.cabal || true
```
diogob: Doesn't this lose a bit of it's purpose when we build with stack?

diogob:
Rock n'Roll :metal:

circle.yml
begriffs: I was thinking if someone wanted to use the modules in postgrest from their own haskell project that I want to make sure they wouldn't have to hold common dependencies back at old versions. Many of our deps have constraints on their major version.

Packdeps is treated as a warning here (for instance I don't *want* to upgrade to aeson 10 because it has a bug).

begriffs: What would also be good is adding [packunused](https://github.com/hvr/packunused) to help us prune packages we no longer need. At the moment is has [stack problems](https://github.com/hvr/packunused/issues/18) though.

Issue #475, opened 2016-01-28
Restart required for detecting table relations on schema change


sscarduzio:
I have an `event_group` table with fields `group_id` and `event_id` referencing `events.id` and `groups.id`respectively.

The weird thing is that even if hydrating the event IDs  works perfectly: 

```
GET /event_group?select=events{*}
[
  {
    "events": {
      "doc": {
     ...
       }
  }
]
```

When I try to do the same with `groups` , I receive this error:
```
GET  /event_group?select=groups{*}
 {
  "hint": null,
  "details": null,
  "code": "42703",
  "message": "column groups.group_id does not exist"
}
```
why does it expect groups.group_id and not groups.id? I didn't need a event.event_id for events!
 
Here's the SQL:

```sql

CREATE TABLE "1".event_group
(
  group_id character varying(32) NOT NULL,
  event_id character varying(36),
  CONSTRAINT fk_eid FOREIGN KEY (event_id) REFERENCES "1".events (id) MATCH SIMPLE ON UPDATE CASCADE ON DELETE NO ACTION,
  CONSTRAINT fk_groups FOREIGN KEY (group_id) REFERENCES "1".groups (id) MATCH SIMPLE ON UPDATE NO ACTION ON DELETE NO ACTION,
  CONSTRAINT uq_gid_eid UNIQUE (event_id, group_id)
)


CREATE TABLE "1".events
(
  doc jsonb,
  id character varying(36) NOT NULL DEFAULT uuid_generate_v4(),
  CONSTRAINT pkey_eid PRIMARY KEY (id)
)

CREATE TABLE "1".groups
(
  id character varying(32) NOT NULL DEFAULT uuid_generate_v4(),
  name character varying(128) NOT NULL,
  description character varying(800),
  CONSTRAINT group_pkey PRIMARY KEY (id)
)
```

begriffs:
This was fixed by restarting your postgrest server, correct? If so I'll rename this issue to discuss ways for PostgREST to detect schema changes.

sscarduzio:
Restarting PostgREST has fixed my issue indeed. But honestly, before I consider this resolved, I'd love to see some notes on the documentation that makes this (probably temporary) side effect evident to the users.
Do you want to take care of this or shall I?

begriffs:
I'd love your help with the docs if you could.

begriffs:
Here are the options I can see for handling schema change.

* Don't cache knowledge of the schema
[...]

PR #477, opened 2016-01-29
#475

sscarduzio:


begriffs:
Thanks for the doc fix. Hopefully we can improve this restart restriction later on.

sscarduzio:
Yeah, let’s start a discussion about that :)

PR #479, opened 2016-01-31
Detect db schema changes

begriffs:
While the correct approach to detecting schema changes is outlined in #475, I moved ahead on this PR to figure out how to make the code keep a changing schema inside an `MVar`. It currently polls for changes every five minutes.

While we wait for support for postgresql LISTEN in hasql to arrive in the next week I wanted to put this PR out for code review.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -47,6 +47,7 @@ executable postgrest
                      , string-conversions
                      , text
                      , time
+                     , transformers
                      , unordered-containers
                      , vector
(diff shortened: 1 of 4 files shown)


begriffs:
Actually it seems like the polling loop never runs more than once...

begriffs:
I created another PR for this. It is simpler in that it uses an IORef rather than an MVar. Also it waits for an explicit signal to reload the schema.

PR #485, opened 2016-02-02
Document default order direction correctly, fix #484

### LINKED ISSUE #484: Default order direction is documented wrong & not explicitly tested for
The documentation [here](https://github.com/begriffs/postgrest/wiki/Routing#ordering) and on [postgrest.com](http://postgrest.com/api/reading/#ordering) states that `?order=col` will default to sorting in descending order. In fact, there is a [tangentially related test](https://github.com/begriffs/postgrest/blob/b9fd083c772831106f82c055832cc70f15e2ffbd/test/Feature/RangeSpec.hs#L33) that depends on the _opposite_ (`ASC` default) being true.

Also, there's no test dedicated to this detail either way.

(Aside: Is there a preferred way to submit documentation issues?)

motiz88:


begriffs:
Thank you!

PR #486, opened 2016-02-05
Adjust usage message to indicate required arguments

begriffs:
Several people have been confused about which command line arguments are required to run postgrest. This pull request changes the usage message slightly to hopefully make it clear. The new message looks like this:

```
$ postgrest
Missing: DB_URL (-a|--anonymous ROLE)

Usage: postgrest DB_URL [-p|--port PORT] (-a|--anonymous ROLE)
                 [-s|--schema NAME] [-j|--jwt-secret SECRET] [-o|--pool COUNT]
                 [-m|--max-rows COUNT]
  PostgREST 0.3.0.3 / create a REST API to an existing Postgres database

Available options:
  -h,--help                Show this help text
  DB_URL                   (REQUIRED) database connection string
  -p,--port PORT           port number on which to run HTTP
                           server (default: 3000)
  -a,--anonymous ROLE      (REQUIRED) postgres role to use for non-authenticated
                           requests
  -s,--schema NAME         (REQUIRED) schema to use for API
                           routes (default: "public")
  -j,--jwt-secret SECRET   secret used to encrypt and decrypt JWT
                           tokens (default: "secret")
  -o,--pool COUNT          max connections in database pool (default: 10)
  -m,--max-rows COUNT      max rows in response (default: "infinity")
```

@StevanWhite do you think this makes it clear enough?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
+- Improved usage screen - @begriffs
 - Reject non-POSTs to rpc endpoints - @begriffs
 - Throw an error for OPTIONS on nonexistent tables - @calebmer
 
--- a/docs/install/server.md
+++ b/docs/install/server.md
@@ -65,7 +65,7 @@ If you want to run the test suite, stack can do that too: `stack test`.
 ### Running the Server
 
 ```bash
-postgrest postgres://user:pass@host:port/db [flags]
+postgrest postgres://user:pass@host:port/db -s public -a anon_user [other flags]
 ```
 
 The user in the connection string is the "authenticator role," i.e.
@@ -80,11 +80,11 @@ The possible flags are:
 <dd>The port on which the server will listen for HTTP requests.
     Defaults to 3000.</dd>
 
-<dt>-a, --anonymous</dt>
+<dt>-a, --anonymous (required)</dt>
 <dd>The database role used to execute commands for those requests
     which provide no JWT authorization.</dd>
 
-<dt>-s, --schema</dt>
+<dt>-s, --schema (required)</dt>
 <dd>The db schema which you want to expose as an API. For historical
     reasons it defaults to <code>1</code>, but you're more likely
     to want to choose a value of <code>public</code>.</dd>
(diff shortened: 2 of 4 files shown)


eric-brechemier:
It would be clearer to group required parameters together (by moving optional parameter `port` after required parameters) and even clearer to separate two groups:

* Required Parameters
* Optional Parameters and Default Values

Also, the usage command should use brackets consistently for required parameters, and no parenthesis for mandatory parameters:

```
Usage: postgrest DB_URL --anonymous ROLE --schema NAME
                 [--port PORT] [--jwt-secret SECRET]
                 [--pool COUNT] [--max-rows COUNT]
```

Also, including only the long form of the options in the example usage makes it easier to understand.

We should also provide some reference on the format of the database connection string, and guidance for the choice of the secret.

begriffs:
OK, I've reordered the list of arguments to group the required ones together, and also included an example db connection string in the usage. Here's what it looks like now

```
$ postgrest
Missing: DB_URL (-a|--anonymous ROLE)

Usage: postgrest DB_URL (-a|--anonymous ROLE) [-s|--schema NAME]
                 [-p|--port PORT] [-j|--jwt-secret SECRET] [-o|--pool COUNT]
                 [-m|--max-rows COUNT]
  PostgREST 0.3.0.3 / create a REST API to an existing Postgres database

Available options:
  -h,--help                Show this help text
  DB_URL                   (REQUIRED) database connection string, e.g.
                           postgres://user:pass@host:port/db
  -a,--anonymous ROLE      (REQUIRED) postgres role to use for non-authenticated
                           requests
  -s,--schema NAME         schema to use for API routes (default: "public")
  -p,--port PORT           port number on which to run HTTP
                           server (default: 3000)
  -j,--jwt-secret SECRET   secret used to encrypt and decrypt JWT
                           tokens (default: "secret")
  -o,--pool COUNT          max connections in database pool (default: 10)
  -m,--max-rows COUNT      max rows in response (default: "infinity")
```

I can't easily remove the parens from required arguments because this usage screen is generated by the `optparse-applicative` library. But it's now consistent that required options use parens and optional ones use brackets.

eric-brechemier:
OK! Would it make sense to make the anonymous role optional, using the authenticator role name read from the database connection string by default?

begriffs:
It's possible to do it but kind of tricky. Although I must admit I usually use the `postgres` role for auth and anon when playing around in development it's actually not a good practice to make them the same thing in production. You'll want your authenticator role to have very few privileges and your anonymous role to be separate. So I think making them the same by default might send the wrong message to developers.

eric-brechemier:
> So I think making them the same by default might send the wrong message to developers.
Got it.

PR #489, opened 2016-02-11
Apply range headers to rpc

diogob:
I believe this adds consistency for a very small cost.
Using the same mechanism to paginate the result of procedure calls makes it easier to reuse front end pagination code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Throw an error for OPTIONS on nonexistent tables - @calebmer
 - Remove deadlock on simultaneous contentious updates - @ruslantalpa, @begriffs
 
+### Added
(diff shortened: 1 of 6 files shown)


begriffs:
This will be really convenient for the client, and it didn't take much code either.

What do you think about adding a response range header as well? It would create more parity between tables and rpc if the rpc response included a `Content-Range`. See the tests in [RangeSpec.hs](https://github.com/begriffs/postgrest/blob/f3d4d1fb60edce96abfe9b39c38222278a67e9b4/test/Feature/RangeSpec.hs) for examples.

diogob:
Oh sure, forgot about that! Will add tonight ;)

diogob:
I believe it's ready for merging :D

diogob:
@begriffs I just rebased after #503 
Should be good to merge again ;)

begriffs:
Great stuff!

PR #492, opened 2016-02-12
Do not block threads on libpq ffi

### LINKED ISSUE #445: Deadlock when concurrently updating same row over high latency db connection
To reproduce on a mac

* Add 500ms artificial latency on localhost port 5432: https://mop.koeln/blog/2015/06/01/Limiting-bandwidth-on-Mac-OS-X-yosemite/
* `brew install parallel`
* Start postgrest on port 3000
* Modify and run the concurrent patch script below

```bash
patch() { 
  curl -s 'http://localhost:3000/foo?id=eq.1' \
    -X PATCH -H 'Accept-Encoding: gzip, deflate, sdch' \
    -H 'Content-Type: application/json' -H 'Accept: */*' \
    -H 'Connection: keep-alive' \
    -H 'Prefer: return=minimal' \
    -H 'Prefer: count=none' \
    --data-binary '{"col":"val"}' \
    --compressed
}
export -f patch

THREADS=${1:-10}
seq $THREADS | parallel patch -j $THREADS
```

Running with at least three threads will cause a database deadlock because of the update statements inside a `WITH` clause.

Possibile solution: combine multiple sql statements into one. Currently the `BEGIN`, role setting, command running and `COMMIT` happen as separate db calls. With high latency it means the time between attempted update and transaction commit is long. This makes it more likely that two updates can be attempted simultaneously in that window.

Bundling multiple sql statements together requires we upgrade the hasql library dependency. The version we use now, 0.7, detects and prevents multiple statements in a single call, but 0.15 allows it.

Bundling commands may not fix all situations. An even tighter request race could have the same results. The real fix is probably to get the update out of a `with` clause. The postgresql [docs](http://www.postgresql.org/docs/9.4/static/queries-with.html#QUERIES-WITH-MODIFYING) state
> Trying to update the same row twice in a single statement is not supported. Only one of the modifications takes place, but it is not easy (and sometimes not possible) to reliably predict which one. This also applies to deleting a row that was already updated in the same statement: only the update is performed. Therefore you should generally avoid trying to modify a single row twice in a single statement. In particular avoid writing WITH sub-statements that could affect the same rows changed by the main statement or a sibling sub-statement. The effects of such a statement will not be predictable.

This issue was discovered and reported by @timwis.

begriffs:
Fixes #445 

I suspect this will also fix the random deadlocks that @motiz88 experienced.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -24,6 +24,7 @@ Flag CI
 executable postgrest
   main-is:             PostgREST/Main.hs
   default-extensions:  OverloadedStrings, ScopedTypeVariables, QuasiQuotes, LambdaCase
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
   default-language:    Haskell2010
   build-depends:       aeson >= 0.8 && < 0.10
                      , base >= 4.8 && < 5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Improved usage screen - @begriffs
 - Reject non-POSTs to rpc endpoints - @begriffs
 - Throw an error for OPTIONS on nonexistent tables - @calebmer
+- Remove deadlock on simultaneous contentious updates - @ruslantalpa, @begriffs
 
 ## [0.3.0.3] - 2016-01-08
 


postgrest.cabal
```diff
@@ -24,6 +24,7 @@ Flag CI
 executable postgrest
   main-is:             PostgREST/Main.hs
   default-extensions:  OverloadedStrings, ScopedTypeVariables, QuasiQuotes, LambdaCase
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
```
ruslantalpa: i saw some ghc options in the stack file also, not sure of how things work here (where is the place to put them) but maybe they should all be in one place

ruslantalpa: days of banging the head against the wall, fixed by a compile option :) welcome to concurrency. I am really glad we are in the haskell world right now and not in C semaphores and mutexes :D

begriffs: I think `cabal check` will complain about having options such as `-O2` or `-Werror` in the cabal file. This would make hackage reject the package for upload.

Whereas cabal check is happy with these threading options, and I want to be sure even if someone builds with cabal that they will build with `-threaded`. So it seems like an OK split to have some options here and some in the stack file.

ruslantalpa: -O2 used to be in cabal before (all of them were there), only recently you moved them to stack

begriffs: It's pretty easy to experiment and see which options ghc check is OK with. For instance when I add Wall, Werror and O2 I get this

```
$ stack exec -- cabal check
These warnings may cause trouble when distributing the package:
* 'ghc-options: -Wall -Werror' makes the package very easy to break with
future GHC versions because new GHC versions often add new warnings. Use just
'ghc-options: -Wall' instead.

* 'ghc-options: -O2' is rarely needed. Check that it is giving a real benefit
and not just imposing longer compile times on your users.
```

So when I had those things in the cabal file before I was doing something that is discouraged.

diogob:
2-fireworks

PR #496, opened 2016-02-17
Reduces pool resource locking

diogob:
This PR was based on a conversation I had with @ruslantalpa in the gitter and on some ad-hoc tests using ```threadDelay```.

I'm still not sure if I should include some kind of test or benchmark to validate this change.

The change is very simple though, I return the results from withResource function before applying the respond continuation. This ensures that the pool resource is freed as soon as the database operation is complete

### DIFF
--- a/src/PostgREST/Main.hs
+++ b/src/PostgREST/Main.hs
@@ -97,8 +97,9 @@ main = do
     body <- strictRequestBody req
     let handleReq = H.run $ inTransaction ReadCommitted
           (runWithClaims conf time (app dbStructure conf body) req)
-    withResource pool $ \case
-      Left err -> respond $ errResponse HT.status500 (cs . show $ err)
+    res <- withResource pool $ \case
+      Left err -> return $ errResponse HT.status500 (cs . show $ err)
       Right c -> do
         resOrError <- handleReq c
-        either (respond . pgErrResponse) respond resOrError
+        either (return . pgErrResponse) return resOrError
+    respond res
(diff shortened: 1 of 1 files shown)


begriffs:
Makes sense, thanks for the fix.

I bet we could create a regression test case. Could fork a bunch of threads in test that all call a stored procedure which raises an exception. The test would check that it never gets that "current transaction is aborted" error.

src/PostgREST/Main.hs
```diff
@@ -97,8 +97,9 @@ main = do
...
     let handleReq = H.run $ inTransaction ReadCommitted
           (runWithClaims conf time (app dbStructure conf body) req)
-    withResource pool $ \case
-      Left err -> respond $ errResponse HT.status500 (cs . show $ err)
+    res <- withResource pool $ \case
+      Left err -> return $ errResponse HT.status500 (cs . show $ err)
       Right c -> do
         resOrError <- handleReq c
-        either (respond . pgErrResponse) respond resOrError
+        either (return . pgErrResponse) return resOrError
```
begriffs: See if this simplification compiles

```haskell
return $ either pgErrResponse id resOrError
```

begriffs:
Tonight I'm looking at how to apply the suggestion from https://github.com/hspec/hspec-wai/issues/32 to make parallel requests in a test case.

diogob:
Cool, because I would get stuck trying to do that :stuck_out_tongue_winking_eye:

begriffs:
Closing this PR and moving over to the other one with a test added.

diogob:
:+1:

Issue #473, opened 2016-01-27
Select safer default value for jwt-secret


eric-brechemier:
I suggest to use a secret made of random bytes, [with a length of 512 bits](http://stackoverflow.com/a/30520284), similar to the output of the command `openssl rand 512`.

This would make it suitable for development and deployment on a single server, unlike the current default of `'secret'`.

joaoportela:
Do you mean changing the default constant that is in the source code?

What would be the benefits of doing this?

eric-brechemier:
> Do you mean changing the default constant that is in the source code?

Yes.

> What would be the benefits of doing this?

The default secret is not suitable for production. Anyone can encode a JWT Token using 'secret' and pretend that they have a different role, e.g. 'postgres', and the `postgrest` server will accept to change to the requested role.

joaoportela:
> The default secret is not suitable for production.

But a constant is never suitable for production. Event it is random bytes. The user needs to generate it, right?

begriffs:
How about generating a new random value every time postgrest starts?

joaoportela:
@begriffs That sounds reasonable. How would the "user" know generated jwt-secret?

eric-brechemier:
> How about generating a new random value every time postgrest starts?

I now realize it was not explicit, but that's exactly what I had in mind.

> How would the "user" know generated jwt-secret?

1. noone knows the secret (that's a feature)
2. it is written into a file when postgrest starts (in which case you need to specify the path to the file, and thus you could have provided a file yourself)

joaoportela:
@eric-brechemier I'm sorry, I'm not familiar with JWT terminology, hence the confusion.

(Hopefully) using the correct terminology:

First of, I assume we are talking about symmetric jwt-keys, is this correct?

How does the issuer access the jwt-secret to generate new tokens? <- This is what I was trying to ask before.

In this scenario postgrest is the consumer, right?

ruslantalpa:
He probably means the case where pg is the consumer and also the producer (rpc returning a special type in which case pg encrypts it) 
The problem is that this is at most usefull in development. In production you probably want a more complicated producer, so you need to know the secret. Also, a restart of the server (schema change:) will invalidate all current tokens.
I dons see a problem with how things are currently working (and even less when there is support for config file)

joaoportela:
@ruslantalpa I'm inclined to agree with you.

It seems that you wouldn't use this in production anyway and in development you would have a ton more work and you don't care that much about the security.

[...]

PR #497, opened 2016-02-17
PostgREST Dockerfile

bobcolner:
PostgRest Dockerfile with ENV parameter passthrough. This dockerfile builds an image that can be linked to the offical postgresql image and uses ENV vars for postgrest parameter passthrough.

## build postgrest docker image
`docker build -t pgrest .`
## start postgresql container
`docker run --name pg  -p 5432:5432 -e POSTGRES_PASSWORD=sillypass -d postgres:9.5`
## start postgREST API server linked to the db
```
docker run --name pgrest --link pg:pg -p 3000:3000 \
	-e POSTGREST_ANONYMOUS=anon \
	-e POSTGREST_SCHEMA=public \
	-e POSTGREST_JWT_SECRET=thisisnotarealsecret \
	-e POSTGREST_MAX_ROWS=1000000 \
	-e POSTGREST_POOL=200 \
	-d pgrest
```


begriffs:
Really grateful for your contribution, docker support will be great.

I've got some questions (keep in mind I'm new to the docker world!)

* Is this compatible with having `docker-compose` run and connect images for the db and postgrest?
* Will this image allow me to do [automated builds](https://docs.docker.com/docker-hub/builds/) on Docker Hub? (I created a [repo](https://hub.docker.com/r/postgrest/postgrest/) there.) It might be cool to have pre-built docker images for the latest code on master as well as specific tagged versions. Is that a common thing to do?
* If for some reason the postgrest server died in the container would it be restarted?
* Could you add to this pull request some info in the docs about running the docker image? The relevant sections are https://github.com/begriffs/postgrest/blob/master/docs/admin/deployment.md and possibly https://github.com/begriffs/postgrest/blob/master/docs/install/server.md

Finally @ptman, @ChristopherBiscardi or anyone else would you care to review?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ENV POSTGREST_JWT_SECRET thisisnotarealsecret
+ENV POSTGREST_MAX_ROWS 1000000
+ENV POSTGREST_POOL 200
+
+RUN apt-get update
+RUN apt-get install -y tar xz-utils wget libpq-dev
+
+RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}/postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
+RUN tar --xz -xvf postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
+RUN mv postgrest /usr/local/bin/postgrest
```
begriffs: The wget, tar, mv sequence could be a one-liner like this
```sh
curl -L http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}/postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz | tar zx -C /usr/local/bin/postgrest
```

But maybe there's a benefit to splitting commands into multiple RUN statements, not sure.

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:14.04
+
+ENV POSTGREST_VERSION 0.3.0.4
+ENV POSTGREST_SCHEMA public
+ENV POSTGREST_ANONYMOUS postgres
+ENV POSTGREST_JWT_SECRET thisisnotarealsecret
```
begriffs: Is it possible to default the jwt secret to have no value? That would work better for our eventual solution of #473.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ENV POSTGREST_POOL 200
+
+RUN apt-get update
+RUN apt-get install -y tar xz-utils wget libpq-dev
+
+RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}/postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
+RUN tar --xz -xvf postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
+RUN mv postgrest /usr/local/bin/postgrest
+
+CMD postgrest postgres://${PG_ENV_POSTGRES_USER}:${PG_ENV_POSTGRES_PASSWORD}@${PG_PORT_5432_TCP_ADDR}:${PG_PORT_5432_TCP_PORT}/${PG_ENV_POSTGRES_DB} \
```
ruslantalpa: I would suggest having the db connection string as a single parameters, it allows greater flexibility (passing additional connection parameters, connecting through sockets ...)

danielxvu:
@begriffs: Regarding the Docker Compose compatibility concern raised, I can confirm that using this pull request's Dockerfile in my own solution (https://github.com/danielxvu/docker-postgrest-with-postgresql) works as expected with the PostgreSQL image via Docker Compose. (On a slightly related note, I'd love to hand over my Docker Compose config to this repository or add a section in the wiki at least--is one of those options preferable over the other?)

bobcolner:
Thanks for the encouragement @begriffs on my first pull request ever.

For 1) this docker file should be compatible with docker-compose. I can get an example of this and some other examples as well.

For 2) I've never setup an automated docker build but this dockerfile should be fine but will need to bump the postgrest version number and ubuntu version manually.

For 3) We can add `--restart=always` flag to the docker run command or to the docker-compose file for auto-restarts.

Finally, we could combine all the run commands into one command. There may be an advantage to keeping them separate though. From the dockerfile docs: 

> The RUN instruction will execute any commands in a new layer on top of the current image and commit the results. The resulting committed image will be used for the next step in the Dockerfile.
> 
> Layering RUN instructions and generating commits conforms to the core concepts of Docker where commits are cheap and containers can be created from any point in an image’s history, much like source control.

ChristopherBiscardi:
There are a few easy wins in terms of size, which turns the Dockerfile into:

```
FROM debian:jessie

ENV POSTGREST_VERSION 0.3.0.4
ENV POSTGREST_SCHEMA public
ENV POSTGREST_ANONYMOUS postgres
ENV POSTGREST_JWT_SECRET thisisnotarealsecret
ENV POSTGREST_MAX_ROWS 1000000
ENV POSTGREST_POOL 200

RUN apt-get update && \
    apt-get install -y tar xz-utils wget libpq-dev && \
    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}/postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz && \
    tar --xz -xvf postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz && \
    mv postgrest /usr/local/bin/postgrest && \
    rm postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz

CMD postgrest postgres://${PG_ENV_POSTGRES_USER}:${PG_ENV_POSTGRES_PASSWORD}@${PG_PORT_5432_TCP_ADDR}:${PG_PORT_5432_TCP_PORT}/${PG_ENV_POSTGRES_DB} \
              --port 3000 \
              --schema ${POSTGREST_SCHEMA} \
              --anonymous ${POSTGREST_ANONYMOUS} \
              --pool ${POSTGREST_POOL} \
              --jwt-secret ${POSTGREST_JWT_SECRET} \
              --max-rows ${POSTGREST_MAX_ROWS}

EXPOSE 3000
```

### How

##### Ubuntu version (original Dockerfile)

```
ubuntu                   14.04               36248ae4a9ac          188 MB
pgt                      latest              0403d77f66fc          352.3 MB
```

removing the whiteouts by consolidating the apt commands (and cleanup) into a single layer reduces size by ~20mb

```
pgt-1                    latest              712dc71fff2a          330.8 MB
```

Using `debian:jessie` as a base image brings it down another ~80mb

```
debian                   jessie              9b607719a62a        125.1 MB
pgt-3                    latest              2aca79a0bd9d        253.8 MB
```

Merging the wget/tar/cleanup commands reduces another ~20mb

```
pgt-4                    latest              d02b67c0edb4        229.3 MB
```

ChristopherBiscardi:
@begriffs 

> Is this compatible with having docker-compose run and connect images of the db and postgrest?

Looks fine. I'll try to check it out in a real compose setup soon.

> Will this image allow me to do automated builds on Docker Hub? (I created a repo there.) It might be cool to have pre-built docker images for the latest code on master as well as specific tagged versions. Is that a common thing to do?

You can set this up. Building master should work by default (and I think it also builds any other branches by default) although alternate schemes can be set up.

> If for some reason the postgrest server died in the container would it be restarted?

I'd expect the container to restart in this scenario, not the server inside. You can set up a [restart policy](https://docs.docker.com/engine/reference/run/#restart-policies-restart) for that purpose.

ruslantalpa:
@ChristopherBiscardi some time ago i was suggesting using your method to build a very small distribution (and package it in docker based on alpine).
https://github.com/begriffs/postgrest/issues/365
Would it be possible for you to try to build postgrest that way?
The only real dependency for postgrest is libpq and after a superficial search on google, it seems ppl have compiled postgresql agains musl so there should be a way to get libpq compiled agains musl.

ChristopherBiscardi:
@ruslantalpa 

I haven't tried it with postgrest, but it seems likely. If there's interest I could do it. The bigger issue would likely be making sure whatever ghc-musl based build process (for building postgrest itself) is acceptable to the project.

We wouldn't need to compile libpq. The following yields a `4.969 MB` image.

```
FROM alpine
RUN apk add libpq --update && rm -rf /var/cache/apk/*
```

bobcolner:
Thanks @ChristopherBiscardi, sliming down docker images is always nice.

This is a good reference on building postgresql on alpine docker: http://michael.otacoo.com/postgresql-2/minimal-docker-container-postgres/

I think it would be a good idea to keep the interface as close as possible to the official postgresql image. FIY, a key Apline linux maintainer just got hired at Docker -it's possiable that the official images will start to be ported to Alpine this year: https://www.brianchristner.io/docker-is-moving-to-alpine-linux/

As for the docker-compose file, I would propose keeping that in a separate thread/pull-request. The compose format was just changed recently. Also worth considering what services should be included. E.g. I think it would be cool to include a Caddy web-server/load-balancer with auto SSL enabled in addition to postgresql + postgRest services. The new compose features would make it easy to scale-up the number of postgRest containers.

ruslantalpa:
@ChristopherBiscardi there is no "build process standard" and even a guide on how to do it would be very welcome, and maybe later an automated process can be developed but for now a guide to manually build for alpine and have it in docker would be cool.

Dockerfile
bobcolner: The default jwt secret can be overwritten by passing a POSTGREST_JWT_SECRET env variable.

bobcolner:
I've updated the Dockerfile with @ChristopherBiscardi suggestions.

begriffs:
Thanks again for your contribution!

PR #503, opened 2016-02-21
Reduces pool resource locking (2)

begriffs:
Cherry-picked @diogob's commits from PR #496 and added a test of concurrent requests. Sadly the test is still failing.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -7,11 +7,13 @@ module PostgREST.Error (pgErrResponse, errResponse) where
 
 import           Data.Aeson                ((.=))
 import qualified Data.Aeson                as JSON
+import           Data.Maybe                (fromMaybe)
 import           Data.Monoid               ((<>))
 import           Data.String.Conversions   (cs)
 import           Data.Text                 (Text)
 import qualified Data.Text                 as T
 import qualified Hasql.Session             as H
+import qualified Hasql.Pool                as P
 import           Network.HTTP.Types.Header
 import qualified Network.HTTP.Types.Status as HT
 import           Network.Wai               (Response, responseLBS)
@@ -19,9 +21,16 @@ import           Network.Wai               (Response, responseLBS)
 errResponse :: HT.Status -> Text -> Response
 errResponse status message = responseLBS status [(hContentType, "application/json")] (cs $ T.concat ["{\"message\":\"",message,"\"}"])
 
-pgErrResponse :: H.Error -> Response
+pgErrResponse :: P.UsageError -> Response
 pgErrResponse e = responseLBS (httpStatus e)
   [(hContentType, "application/json")] (JSON.encode e)
+
+instance JSON.ToJSON P.UsageError where
+  toJSON (P.ConnectionError e) = JSON.object [
+    "code" .= ("" :: T.Text),
+    "message" .= ("Connection error" :: T.Text),
+    "details" .= (cs (fromMaybe "" e) :: T.Text)]
+  toJSON e = JSON.toJSON e -- H.Error
 
 instance JSON.ToJSON H.Error where
   toJSON (H.ResultError (H.ServerError c m d h)) = JSON.object [
@@ -51,15 +60,17 @@ instance JSON.ToJSON H.Error where
     "message" .= ("Database client error"::String),
     "details" .= (fmap cs d::Maybe T.Text)]
 
-httpStatus :: H.Error -> HT.Status
-httpStatus (H.ResultError (H.ServerError c _ _ _)) =
+httpStatus :: P.UsageError -> HT.Status
+httpStatus (P.ConnectionError _) =
+  HT.status500
+httpStatus (P.SessionError (H.ResultError (H.ServerError c _ _ _))) =
   case cs c of
     '0':'8':_ -> HT.status503 -- pg connection err
     '0':'9':_ -> HT.status500 -- triggered action exception
     '0':'L':_ -> HT.status403 -- invalid grantor
     '0':'P':_ -> HT.status403 -- invalid role specification
-    "23503" -> HT.status409 -- foreign_key_violation
-    "23505" -> HT.status409 -- unique_violation
+    "23503"   -> HT.status409 -- foreign_key_violation
+    "23505"   -> HT.status409 -- unique_violation
     '2':'5':_ -> HT.status500 -- invalid tx state
     '2':'8':_ -> HT.status403 -- invalid auth specification
     '2':'D':_ -> HT.status500 -- invalid tx termination
@@ -76,8 +87,8 @@ httpStatus (H.ResultError (H.ServerError c _ _ _)) =
     'H':'V':_ -> HT.status500 -- foreign data wrapper error
     'P':'0':_ -> HT.status500 -- PL/pgSQL Error
     'X':'X':_ -> HT.status500 -- internal Error
-    "42P01" -> HT.status404 -- undefined table
-    "42501" -> HT.status404 -- insufficient privilege
-    _ -> HT.status400
-httpStatus (H.ResultError _) = HT.status500
-httpStatus (H.ClientError _) = HT.status503
+    "42P01"   -> HT.status404 -- undefined table
+    "42501"   -> HT.status404 -- insufficient privilege
+    _         -> HT.status400
+httpStatus (P.SessionError (H.ResultError _)) = HT.status500
+httpStatus (P.SessionError (H.ClientError _)) = HT.status503
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -30,7 +30,7 @@ instance JSON.ToJSON P.UsageError where
     "code" .= ("" :: T.Text),
     "message" .= ("Connection error" :: T.Text),
     "details" .= (cs (fromMaybe "" e) :: T.Text)]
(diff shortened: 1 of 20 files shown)


begriffs:
Actually it looks like the tests don't even use a db pool. I'm going to find a way that the tests can share code with the rest of the program for running a server. Their divergence isn't good.

ruslantalpa:
The tests fail because as you said, they do not use the Data.Pool abstraction which is used in Main.hs, basically you try to execute 3 queries on the same db connection at the same time
You need to make withApp code used for tests from here https://github.com/begriffs/postgrest/blob/master/test/SpecHelper.hs#L43
more like this code https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Main.hs#L95-L104

Maybe even extract that into a separate function that can be used by tests and app.

Another idea in a completely different direction is that maybe tests should not be an "executable that embeds the postgrest code" but they should connect to a running postgrest instance.

begriffs:
@ruslantalpa I'm almost ready with a new commit. We'll see how you like it.

begriffs:
Pushing this work in progress. It doesn't properly reset the db between tests, and the test suite has some kind of infinite loop which uses 100% CPU. Maybe someone can help me spot the error.

ruslantalpa:
One thing that tests didn't have before is:
inside the psotgrest function you now have 
let middle = logStdout . defaultMiddle
maybe you could take that outside the postgrest function

begriffs:
Tried removing the middleware from the `postgrest` function but it still hangs. I'm going to leave it in because the middleware sets the cors policy and includes `unsupportedAccept`, both of which affect the tests.

src/PostgREST/Error.hs
```diff
@@ -19,10 +21,17 @@ import           Network.Wai               (Response, responseLBS)
...
+pgErrResponse :: P.UsageError -> Response
 pgErrResponse e = responseLBS (httpStatus e)
   [(hContentType, "application/json")] (JSON.encode e)
 
+instance JSON.ToJSON P.UsageError where
+  toJSON (P.ConnectionError e) = JSON.object [
+    "code" .= ("" :: T.Text),
+    "message" .= ("Connection error" :: T.Text),
+    "details" .= (cs (fromMaybe "" e) :: T.Text)]
+  toJSON e = JSON.toJSON e -- H.Error
```
ruslantalpa: The locking comes from the line above

change
```toJSON e = JSON.toJSON e -- H.Error```
to
```toJSON (P.SessionError e) = JSON.toJSON e -- H.Error```

after this the tests will not lock but most of them will fail with "current transaction is aborted, commands ignored until end of transaction block".
that is because on the db connection, a previous test executed a failing query (missing table) and the transaction was not automatically rolled back. Looking if i can pinpoint where exactly

**Comment by ruslantalpa** [comment]
The problem comes from inside hasql.transaction.
https://github.com/nikita-volkov/hasql-transaction/blob/master/library/Hasql/Transaction.hs#L73
In case there is an error, there is no "commit" call at the end.
In Main.postgrest, when you get back an error, you need to call ```Queries.abortTransaction``` (from hasql.transaction) before returning the error to http client.

FYI, these symtoms are also visible in the main executable, if you hit something that returns an error, the next call (even if valid) will fail

**Comment by begriffs** [comment]
I don't think I'm able to call `abort` at the appropriate times. A `SessionError` is only available to me once `Pool.use` has finished, but I can't then reliably get access to that same db connection in order to run the sql `abort`. I think this may be something that has to be fixed inside the hasql library. /cc @nikita-volkov

**Comment by nikita-volkov** [comment]
> A SessionError is only available to me once Pool.use has finished

It's available to you inside of `Session` through the `MonadError` interface.

**Comment by nikita-volkov** [comment]
> The problem comes from inside hasql.transaction.
https://github.com/nikita-volkov/hasql-transaction/blob/master/library/Hasql/Transaction.hs#L73
In case there is an error, there is no "commit" call at the end.
In Main.postgrest, when you get back an error, you need to call Queries.abortTransaction (from hasql.transaction) before returning the error to http client.

I don't explicitly abort transaction because AFAIK Postgres aborts it automatically when there is an error. Am I not correct?

**Comment by ruslantalpa** [comment]
@nikita-volkov i don't think you are, i think you explicitly need to have explicitly have a "rollback" statement.
I think this is exactly what psql is doing with ON_ERROR_ROLLBACK

**Comment by ruslantalpa** [comment]
The reason for doing this manually is probably because you have multiple options, completely rollback the transaction or rollback to a specific savepoint.
As for the hasql.transaction i think it's the correct implementation, then client ()postgrest needs to decide what to do with the error, which in our case is rollback.

**Comment by begriffs** [comment]
@nikita-volkov thanks for the tip about `MonadError`. I got all the tests passing!

However I then increased the number of connections in my concurrent test so that they outnumber slots in the db pool. The test now reliably fails with this:

```json
{
  "hint":null,
  "details":null,
  "code":"26000",
  "message":"prepared statement \"12\" does not exist"
}
```

What this test [does](https://github.com/begriffs/postgrest/blob/one-tx-per-client/test/Feature/ConcurrentSpec.hs#L23-L34) is concurrently hit a route that causes a db error. The error should result in postgrest checking the sql error code and issuing an http 404. When each request is served with a fresh pool connection that's exactly what happens. However when a request has to reuse a connection in the pool which previously had an error then we get this other message about the prepared statement not being found.

Not sure if I'm using hasql correctly.

nikita-volkov:
@ruslantalpa

> The reason for doing this manually is probably because you have multiple options, completely rollback the transaction or rollback to a specific savepoint.
As for the hasql.transaction i think it's the correct implementation, then client ()postgrest needs to decide what to do with the error, which in our case is rollback.

No. `Hasql.Transaction` is meant to abstract over the transaction controlling statements as well as the conflict errors. Expecting the user to issue a rollback would make it a leaky abstraction.

I've just released an update, which does the roll-back.

Thanks for spotting this.

nikita-volkov:
@begriffs Seems like you've spotted another bug in Hasql. This one is caused by the recent updates to the way the connections, which had errors, get handled. I think I know what to do. I'll post back in an hour or two.

Thanks for spotting this!

nikita-volkov:
@begriffs
I can't reproduce the bug. Here's the test that I've tried: https://github.com/nikita-volkov/hasql/blob/6cb86412cabab2bf0d19ea7fbfc4a6c82a036187/tasty/Main.hs#L25-L48

It tries to reexecute a prepared statement after an error and succeeds.

Can you please provide a Hasql-specialized reproduction of the issue or, better, make a PR with such a test?

src/PostgREST/Main.hs
```diff
@@ -55,50 +55,50 @@ main = do
...
-        resOrError <- handleReq c
-        either (respond . pgErrResponse) respond resOrError
+
+    let handleReq = runWithClaims conf time (app dbStructure conf body) req
+    resp <- either pgErrResponse id <$> P.use pool
+      (rollbackOnError $ HT.run handleReq HT.ReadCommitted HT.Write)
+    respond resp
+
+ where
+  rollbackOnError :: H.Session a -> H.Session a
```
ruslantalpa: This is not needed anymore after the last commit to hasql.transaction, just upgrade the version to 0.4.3

begriffs:
@nikita-volkov I reproduced the prepared statement error in a simpler case.
https://github.com/begriffs/hasql-stress/tree/2e0c8e2ced0b337b75278564108335e2b28977c9

Run the app and watch your postgres logs.

nikita-volkov:
@begriffs Thanks. I've reproduced it. The problem is that an erroneous statement fails to get prepared.

nikita-volkov:
@begriffs The prepared statement issue should be fixed in 0.19.9.

begriffs:
@nikita-volkov thanks, I'm grateful for your super fast bugfix.

RobinBressan:
Hi everybody, thanks all for your work on this project! 
@begriffs I am facing with this issue when writing my functionnal tests for erroneous queries. When do you plan a new release with this fix?

begriffs:
I'll build a 0.3.1.0 release for you. Just waiting to include another fix in the release that @ruslantalpa is working on.

In the meantime you could [build from source](http://postgrest.com/install/server/#building-from-source) which is pretty easy actually.

RobinBressan:
@begriffs Thanks a lot for that. I tried to build but I need to do it in a docker build process and it fails in it.

begriffs:
Can you tell me more about how the build fails?

ruslantalpa:
@begriffs i encountered a build error myself, i fixed it in the PR
it looked like this
`Warning: output was redirected with -o, but no output will be generated`
it might be @RobinBressan had the same thing

RobinBressan:
@begriffs This is what I get:

```sh
network-info-0.2.0.8: download
mtl-2.2.1: copy/register
happy-1.19.5: download
stack: c_poll: permission denied (Operation not permitted)
cmdargs-0.10.13: copy/register
network-2.6.2.1: build
network-info-0.2.0.8: configure
network-info-0.2.0.8: build
happy-1.19.5: configure
network-info-0.2.0.8: copy/register
happy-1.19.5: build
network-2.6.2.1: copy/register
happy-1.19.5: copy/register
sendWakeup: invalid argument (Bad file descriptor)
stack: ioManagerDie: write: Bad file descriptor
The command '/bin/sh -c cd postgrest && sudo stack install --install-ghc --local-bin-path /usr/local/bin' returned a non-zero code: 1
```

PR #508, opened 2016-02-29
Test that binary builds, not just that suite passes

begriffs:
Regarding comment in #503


begriffs:
This is ready for code review.

* Now CI will fail if the binary fails to build
* Improved .cabal file hygiene
* Allow aeson 11 (to satisfy packdeps)

PR #511, opened 2016-03-06
Use `CMD exec` in Dockerfile

Recmo:
Without exec the `postgrest` process is not run with PID 1 (it
is a child process of the shell that starts it). This means
signals send to the docker (like `docker stop` or ^C) will
not be handled correctly.

This commit adds `exec` to resolve this problem, as per the
recommendation in the Dockerfile documentation:

<https://docs.docker.com/engine/reference/builder/#shell-form-entrypoint-example>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
+- Terminate gracefully on SIGTERM - @recmo
 
 ## [0.3.1.0] - 2016-02-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
-- Terminate gracefully on SIGTERM - @recmo
+* Terminate gracefully on SIGTERM (for use in Docker) - @recmo
 
 ## [0.3.1.0] - 2016-02-28
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VER
     mv postgrest /usr/local/bin/postgrest && \
     rm postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
 
-CMD postgrest postgres://${PG_ENV_POSTGRES_USER}:${PG_ENV_POSTGRES_PASSWORD}@${PG_PORT_5432_TCP_ADDR}:${PG_PORT_5432_TCP_PORT}/${PG_ENV_POSTGRES_DB} \
+CMD exec postgrest postgres://${PG_ENV_POSTGRES_USER}:${PG_ENV_POSTGRES_PASSWORD}@${PG_PORT_5432_TCP_ADDR}:${PG_PORT_5432_TCP_PORT}/${PG_ENV_POSTGRES_DB} \
               --port 3000 \
               --schema ${POSTGREST_SCHEMA} \
               --anonymous ${POSTGREST_ANONYMOUS} \
--- a/src/PostgREST/Main.hs
+++ b/src/PostgREST/Main.hs
@@ -60,10 +60,11 @@ main = do
 
 #ifndef mingw32_HOST_OS
   tid <- myThreadId
-  void $ installHandler keyboardSignal (Catch $ do
-      P.release pool
-      throwTo tid UserInterrupt
-    ) Nothing
+  forM_ [sigINT, sigTERM] $ \sig ->
+    void $ installHandler sig (Catch $ do
+        P.release pool
+        throwTo tid UserInterrupt
+      ) Nothing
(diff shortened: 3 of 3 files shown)


Recmo:
Sorry I didn't post this before the Dockerfile branch got merged, I just discovered postgrest and as I was playing with the dockerfile branch it got merged.

Some minor points of note:

* Proper signal handling is necessary for docker-compose to close postgrest. Without it it will timeout and sigkill postgrest.

* If the configuration is read from environment variables or a config file, the exec version of `CMD` can be used. This is how respectively the official postgres and nginx dockerfiles handle configuration.

* The `wget` install can be replaced by a `ADD <url>` command. Shaving a tiny bit of space. (Not too much because `libpq-dev` still drags in all the TLS libraries).

* I'm working on a nginx ↔ postgrest ↔ postgresql docker-compose setup using only the official images. I'd love to share the results and do a little write-up if there is interest.

Recmo:
It turns out just adding `exec` is not enough to have the docker container properly shutdown.

begriffs:
Thanks for pointing out the `docker stop` problem.

The [docs](https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/) suggest that if you use an array syntax for CMD then it will execute the command without an intervening shell:

```dockerfile
CMD ["executable","param1","param2"]
```

Recmo:
True, but the array notation does not expand shell variables. The current Dockerfile uses shell expansion to pass the environment variables as arguments. Since the only way to pass configuration to postgrest is via the command line arguments, I do not see a way arround this.

The official Nginx Dockerfile doesn't have this problem, since it simply reads the configuration from a predefined location that is easily augmented with a volume mount.

The official Postgres Dockerfile uses environment variables a lot for configuration. It starts a shell script that sets everything up and then exec's the postgress process. This also results in Postgres running as PID 1.

The patch I propose in this merge request mimics the Postgres method, but does it in a one-liner instead of a shell script.

However, I submitted it prematurely. `docker stop` still doesn't work for me.

I think it has something to do with Linux treating the PID 1 process as special and installing *ignore* signal handlers by default for `SIGTERM`. Since postgrest does not set it's own signalhandlers (and really why should it?) it will keep the default and ignore SIGTERM when ran as PID 1.

I'll dig some more into this.

begriffs:
Actually postgrest does set a signal handler for CTRL-C

https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Main.hs#L61-L67

It uses the handler to close out all db connections in its pool.

Recmo:
Indeed it does! It did not handle sigTERM, so I've added a handler. These two should be sufficient according to [1].

This also solves the `docker stop` problem, since now it overrides the default handlers.

And since there are no zombie processes to reap, this means we no longer need the `tini` process, reducing complexity.

Would you like me to rebase to remove the tini commit and it's revert?

[1]: https://stackoverflow.com/questions/4042201/how-does-sigint-relate-to-the-other-termination-signals

begriffs:
Thanks for the fix! A rebase would be great, as well as this small style nitpick that the linter pointed out:

```
src/PostgREST/Main.hs:63:3: Suggestion: Use forM_
Found:
  flip mapM_
Why not:
  forM_
```

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
+- Terminate gracefully on SIGTERM - @recmo
+- Use exec in Dockerfile - @recmo
```
ruslantalpa: At least until now, the changelog contained info strictly about the executable, docker file is more like documentation/example (that should be moved somewhere else, in time). What i am trying to say is the line about docker should not be here :)

begriffs:
Could mention something like

```
* Terminate gracefully on SIGTERM (for use in Docker) - @recmo
```

(Don't forget the asterisk to make it a markdown list)

CHANGELOG.md
Recmo: Thanks for pointing it out. I didn't feel good about occupying two changelog lines for something so trivial.

begriffs:
For a test I modified it to print a message when it traps a signal and verified that CTRL-C and `kill` get caught. So that looks good.

However when I use this Dockerfile to make a container it takes 10 seconds for `docker stop` to work. Was that your original problem or does it mean the problem was fixed?

Recmo:
> However when I use this Dockerfile to make a container it takes 10 seconds for docker stop to work. Was that your original problem or does it mean the problem was fixed?

That should not happen! `docker stop` should finish nearly instantly. (In particular, it should not depend on the timeout argument `-t` which defaults to 10 sec.)

Did you make sure the updated executable ended up in the docker image?

By default the docker image will download version `0.3.1.0` from the releases, which doesn't handle SIGTERM. This needs to be overriden by the new updated version. You can do so by adding the line

```
ADD ./my-build-dir/postgrest /usr/local/bin/postgrest
```

It needs to be added after line 18 in the Dockerfile.

begriffs:
Right, I forgot that it downloads the most recent release. After copying in a fresh binary it works perfectly.

PR #514, opened 2016-03-07
Minor changes in documentation

Recmo:
* Do not build under `sudo` (which fails).
* Use `gen_random_uuid()` instead of `uuid_generate_v4()` in example.


begriffs:
Thank you, should save other people from confusion!

PR #515, opened 2016-03-07
Multi api example

nicklasaven:
An example of how to setup PostgREST with 2 apis and nginx in front.

### DIFF
--- a/docs/multi_api_example/index.md
+++ b/docs/multi_api_example/index.md
@@ -1,6 +1,6 @@
 #Multi API example#
 
-How to set up PostgREST  with multiple apis and nginx in front on Debian
+##How to set up PostgREST  with multiple apis and nginx in front on Debian##
 
 This is written for Debian, and tested on Debian 8, Jessie
 
@@ -9,12 +9,14 @@ We will do this with PostgREST. PostgREST handles multiple apis by letting you d
 
 So, first install PostgREST as describe in the doc (link do installation doc here)
 
+###Database Setup###
 For this example, we have the following setup in the database:
 
-First we create a database in PostgreSQL that we call test_db
+Start by creating a database that we call **test_db**
 
-Users and roles:
-/*authenticator user, we use postgrest*/
+```
+/*Users and roles:
+authenticator user, we use postgrest*/
 CREATE ROLE postgrest NOINHERIT LOGIN;
 
 /*For anonymous lgoin*/
@@ -67,13 +69,13 @@ GRANT USAGE ON SCHEMA v2 TO anon;
 
 GRANT SELECT ON v1.plates TO anon;
 GRANT SELECT ON v2.plates TO anon;
-
-Summary of database
+```
+#####Summary of database#####
 As you see we have no keys or constraints on those tables. 
 
 We just have prepared two schema's for two different apis using the same data.
 
-About the unix user running PostgREST
+###About the unix user running PostgREST###
 PostgreSQL have a feature that a local user o the server can connect directly to the database without logging in through unix-socket. That is very convenient since that mean you don't have to deal with passwords in your setup. It is also as secure as your server since it is not usable without access to the server.
 
 To make it secure from any user on the system use the peer authentication method.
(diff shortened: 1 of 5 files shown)


ruslantalpa:
The title is making me think that the APIs are unrelated, maybe something closer to 
`How to serve multiple versions of the API with Postgrest and Nginx`

The tutorial is structured in terms of "here is this custom config, copy it to this location". I would suggest you modify the current  config+init script to make them general (details below) then change the tutorial to "copy the default files to these locations, rename to this and edit these lines" and the default files should be added/changed in /debian diretory with this PR

Config file and init scripts, the good part is that you eliminated the wrapper thing and added the parameters needed to run multiple services based on the same binary.
The part that i don't think is a good direction is that you eliminated the default values from the init script
and made them mandatory in the config file, i would go for something like this in the init script
```
POSTGREST_DBPOOL=${POSTGREST_DBPOOL:-10}
```
and this in the config file
```
# database pool
#POSTGREST_POOL=10
```

and the if statements here https://github.com/begriffs/postgrest/blob/master/debian/postgrest.init.d#L36-L52 are useless since there is always a value, POSTGREST_OPTS can be constructed in one line.
I would also eliminate the separate config values for the db connection and just add "connection_string", this will simplify the init script and also make it more flexible to specify the connection.

ruslantalpa:
Also maybe it's time to eliminate the `POSTGREST_` part from each variable since it's obviouse :+1:

nicklasaven:
Thanks @ruslantalpa 
Good points

Yes, this was written more in a blog post manner, I agree. The purpose of simplifying the init script, was that I wanted to make it clearer what is really happening. But then I should of course have removed also the if-parts.

Maybe there could be both on different parts of the documentation. Both simplified versions of init files and one more multipurpose, that catches all cases.

The reason I say so is that people like me who is not very familiar with init-scripts, and bash syntax in general have a hard time to understand all the parts of a multi purpose init script.

The "POSTGREST_DBPOOL=${POSTGREST_DBPOOL:-10}" is very nice and compact but I had a hard time googling that syntax. It is quite obvious what is happening, but anyway a little confusing.

So, why being able to hard code in init script? Doesn't that just make things confusing about where to search for where a parameter is set? I think PostgREST documentation could recommend one of the paths. All in init or all in config.


Maybe also the nginx part should be in a separate doc? I put things together just because I was in the blog post mood. I don't know if that is the right way. 

I separate part about nginx could also say some words about https and let's encrypt.
I found this document from digital ocean very good and it helped me to set up https quite easy.
https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-14-04

ruslantalpa:
The reason i asked for a general init script is precisely because ppl do not want to mess with that, so ideally it's just a script they copy to some location and do not care what's inside.
And for this specific case (multiple api versions) they edit just one line to specify the config file location.

About the default values, even though you can omit parameters like --pool when staring postgrest by hand, those parameters still have default values inside the binary (basically every parameter if i am not mistaken has a default value).
So instead of generating a "dynamic parameter list" command for staring pg, i think it would be best if the generated command line contains all the parameters, and the parameters that have a default value in the binary, should have that default also set in the init script. This will make the init script simpler (without a lot of ifs) and also the config file will be in the form of
```
#key=defaultvalue
```
that makes it clear what are the configuration options and the current default value if the line is commented.

docs/multi_api_example/index.md
```diff
@@ -0,0 +1,220 @@
...
+
+Now you can use a  web browser to see the output from the two apis using:
+http://localhost:3000/plates
+and
+http://localhost:3001/plates
+
+You can stop postgREST with ctrl-c
+
+###Make services of this###
+Now we want to put our two instances in services so they start on reboot andjust works in the background.
```
kenhahn85: `andjust` -> `and just`

opensrcken: `andjust` -> `and just`

docs/multi_api_example/index.md
```diff
@@ -0,0 +1,220 @@
...
+
+We just have prepared two schema's for two different apis using the same data.
+
+###About the unix user running PostgREST###
+PostgreSQL have a feature that a local user o the server can connect directly to the database without logging in through unix-socket. That is very convenient since that mean you don't have to deal with passwords in your setup. It is also as secure as your server since it is not usable without access to the server.
+
+To make it secure from any user on the system use the peer authentication method.
+
+Then we have two things to do
+1) Create a system user named postgres
```
Velmont: postgres*t*.  A kinda bad typo considering what the postgres user is ;)

begriffs:
So I read these docs again and my thought is they have nice material but should probably be split up and used to fill out existing empty places in the docs. For instance a discussion of init scripts in http://postgrest.com/admin/deployment/#debian and schema versioning in http://postgrest.com/admin/versioning/

I'm going to close this PR as it is but will consult its contents when updating the actual sections of the docs. Thanks for researching this stuff @nicklasaven, especially the init scripts which I know very little about.

PR #522, opened 2016-03-11
Allow SQL functions to generate registered JWT claims

### LINKED ISSUE #198: Add iss, sub, aud, exp, nbf, iat, jti support to JWT
The Haskell JWT library doesn't offer an implementation at the time (even though there is [this issue](https://bitbucket.org/ssaasen/haskell-jwt/issue/9/make-it-easy-to-verify-exp-nbf-and-iat))

I'm not sure what's the best way to deal with the problem, maybe it's better to contribute upstream the changes and keep most of the changes outside postgrest?

The [Ruby library](https://github.com/progrium/ruby-jwt) by progrium is a good reference

### LINKED ISSUE #500: postgREST does not return exp field in jwt token
Creating new issue for better visibility.

I have updated the login function from blog user-management example to include an `exp` token expiration field. I'm useing postgREST version 3.0.4

```
create type basic_auth.jwt_claims AS (role text, email text, exp integer);

create or replace function
login(email text, pass text) returns basic_auth.jwt_claims
  language plpgsql
  as $$
declare
  _role name;
  result basic_auth.jwt_claims;
begin
  select basic_auth.user_role(email, pass) into _role;
  if _role is null then
    raise invalid_password using message = 'invalid user or password';
  end if;
  -- TODO; check verified flag if you care whether users
  -- have validated their emails
  select _role as role, login.email as email,
         (round(extract(epoch from now()))+5*60)::int as exp
         into result;
  return result;
end;
$$;
```
I can confirm that the new psql login function returns an jwt_claims type with the exp integer field added. However, the actual jwt token returned by the postgrest login endpoint does not include the exp field, only email and role. As a result my token is effectively valid forever.

Perhaps I missing something obvious?

begriffs:
* Fixes #500
* Fixes #198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
+* Allow SQL functions to generate registered JWT claims - @begriffs
 * Terminate gracefully on SIGTERM (for use in Docker) - @recmo
 
 ## [0.3.1.0] - 2016-02-28
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -40,6 +40,8 @@ executable postgrest
                      , http-types
                      , interpolatedstring-perl6
                      , jwt
+                     , lens >=3.8 && < 5.0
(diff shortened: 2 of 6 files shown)


bobcolner:
great work.

PR #523, opened 2016-03-13
Prevent duplicate call to stored procs

### LINKED ISSUE #513: Execute RPCs only once
In the user management example, the request `POST /rpc/signup` with `{ "email": "foo@bar.com", "pass": "foobar" }` fails. The log shows the following:

```
ERROR:  duplicate key value violates unique constraint "users_pkey"
DETAIL:  Key (email)=(foo@bar.com) already exists.
CONTEXT:  SQL function "signup" statement 1
STATEMENT:   SELECT
                    (select pg_catalog.count(1)  from "public"."signup"("email":='foo@bar.com'::unknown,"pass":='foobar'::unknown)  c)  as countTotal,
                   pg_catalog.count(t) as countResult,
                   array_to_json(
                     coalesce(array_agg(row_to_json(t)), '{}')
                   )::character varying
                   from (select *  from "public"."signup"("email":='foo@bar.com'::unknown,"pass":='foobar'::unknown)  LIMIT 1000000 OFFSET 0) t 
```

It appears the RPC stored procedure is called twice for one request.

A workarround is to add a `Prefer: count=none` to the request headers.

This is (in a way) a special case of #273. I file it separately because the documentation explicitly mentions that RPCs do not have to be idempotent, but postgrest fails to treat them as such.

begriffs:
Fixes #513

### DIFF
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -262,6 +262,18 @@ $_$;
 
 
 --
+-- Name: callcounter(); Type: FUNCTION; Schema: test; Owner: -
+--
+
+CREATE SEQUENCE callcounter_count START 1;
+
+CREATE FUNCTION callcounter() RETURNS bigint
+    LANGUAGE sql
+    AS $_$
+    SELECT nextval('test.callcounter_count');
+$_$;
+
+--
 -- Name: test_empty_rowset(); Type: FUNCTION; Schema: test; Owner: -
 --
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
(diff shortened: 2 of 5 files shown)


test/fixtures/schema.sql
```diff
@@ -262,6 +262,18 @@ $_$;
...
 --
+-- Name: callcounter(); Type: FUNCTION; Schema: test; Owner: -
+--
+
+CREATE SEQUENCE callcounter_count START 1;
+
+CREATE FUNCTION callcounter() RETURNS bigint
+    LANGUAGE sql
+    AS $_$
+    SELECT nextval('test.callcounter_count');
```
diogob: :+1: very clever way to test number of calls

diogob:
LGTM :D

test/fixtures/schema.sql
begriffs: :smile: thanks! Good old postgresql, always ready with a trick.

PR #524, opened 2016-03-13
Preserve unicode in requests and responses

### LINKED ISSUE #516: Encoding issue
Tested by java code:
```
        String json = "{\"id\": 2,  \"name\": \"￥\"}";
		HttpPatch patch = new HttpPatch("http://localhost:5430/event?id=eq.2");
		patch.addHeader("Content-Type","application/json;charset=UTF-8");
		HttpEntity entity = new StringEntity(json, ContentType.APPLICATION_JSON);
		patch.setEntity(entity);
		CloseableHttpClient client = HttpClients.createDefault();
		try {
			CloseableHttpResponse resp = client.execute(patch);
			System.out.println("response : " + (resp.getEntity() != null ? EntityUtils.toString(resp.getEntity()):""));
			
			client.close();
		} catch (Exception e) {
			e.printStackTrace();
		}
```
and return error:
```
response : {"hint":null,"details":null,"code":"22021","message":"invalid byte sequence for encoding \"UTF8\": 0xe5 0x27 0x3a"}
```
it worked before v0.3.0.4

begriffs:
Fixes #516

It seems like the interpolated strings are safe in the case when return-type polymorphism to the `Text` type causes `[qc| ... ]` to preserved unicode. Because interpolated strings are pleasant to read I've left them in but annotated all occurrences with `:: Text`. On one hand it's kind of touchy because future uses of interpolated strings without the Text annotation will cause problems, we have to be vigilant. On the other this PR adds tests that the operations of insert, update, and procedure call work with Unicode so it protects against regressions in these operations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Fixed
+- Preserve unicode values in insert,update,rpc (regression) - @begriffs
 - Prevent duplicate call to stored procs (regression) - @begriffs
 - Allow SQL functions to generate registered JWT claims - @begriffs
 - Terminate gracefully on SIGTERM (for use in Docker) - @recmo
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -41,8 +41,8 @@ data PreferRepresentation = Full | HeadersOnly | None deriving Eq
 -- route responses and upload payloads
 data ContentType = ApplicationJSON | TextCSV deriving Eq
 instance Show ContentType where
-  show ApplicationJSON = "application/json"
-  show TextCSV         = "text/csv"
+  show ApplicationJSON = "application/json; charset=utf-8"
+  show TextCSV         = "text/csv; charset=utf-8"
 
 {-|
   Describes what the user wants to do. This data type is a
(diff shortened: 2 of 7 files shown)


ruslantalpa:
There seems to be a common theme with these changes that you first annotate it with Text and immediately convert the result with cs to ByteString for `H.statement`.
Maybe if you define a custom function to replace ```H.statement`` like the one below, it will eliminate the need for ```::Text && (cs sql)``` all over the place.
```
-- type definition copied from hasql
myStatement :: Text -> Encoders.Params a -> Decoders.Result b -> Bool -> Query.Query a b
myStatement sql = H.statement $ cs sql
```

Also note that DbStructure also uses these interpolated string so it would be good for this pr also to fix that file.

ruslantalpa:
The definition of H.statement is here https://github.com/nikita-volkov/hasql/blob/master/library/Hasql/Query.hs#L54 and it also has some annotations for the compiler, maybe you would  need to add those also for performance.

dnaq:
If we know that we always will call H.statement (cs sql) with sql :: Text, couldn't we instead use a monomorphic version of cs, i.e. Data.Text.Encoding.encodeUtf8 and let type inference do most parts of the job. Basically what @ruslantalpa suggested.

So H.statement (T.encodeUtf8 sql) ...

begriffs:
How do you like this `unicodeStatement` approach? Should I merge?

dnaq:
Looks fine to me, thanks.

PR #525, opened 2016-03-14
adding uuid-ossp extension into documentation

michalskop:
I got "function uuid_generate_v4() does not exist" error for `/rpc/signup` and for direct inserting into `basic_auth.users` (using pgsql 9.5):
```
insert into basic_auth.users VALUES  ('michal@example.com','123','author',true);
```
and had to do
```
CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
create extension if not exists pgcrypto;
```
(the second line is taken from the documentation)

### DIFF
--- a/docs/examples/users.md
+++ b/docs/examples/users.md
@@ -28,10 +28,11 @@ value.
 ### Storing Users and Passwords
 
 We create a database schema especially for auth information. We'll
-also need the postgres extension
-[pgcrypto](http://www.postgresql.org/docs/current/static/pgcrypto.html).
+also need the postgres extensions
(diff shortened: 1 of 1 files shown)


begriffs:
Reading this again I think a better choice would be to use `gen_random_uuid()` [[1]](http://www.postgresql.org/docs/9.5/static/pgcrypto.html#AEN171443) from pgcrypto. The uuid-ossp docs say

> Note: If you only need randomly-generated (version 4) UUIDs, consider using the gen_random_uuid() function from the pgcrypto module instead.

PR #528, opened 2016-03-15
Fix typo-d flag in the installation docs.

rowdypixel:
Just fixing a typo in the docs where it lists -p twice instead of -p and -o


begriffs:
Thank you!

PR #534, opened 2016-03-25
Regression test for read/write unicode table names

begriffs:
Test (suggested by @ruslantalpa) to ensure that non-English table/schema names are accessible.

I verified that it fails when cherry-picked onto v0.3.1.0:
```
status mismatch:
  expected: 200
  but got:  404
body mismatch:
  expected: []
  but got:  {"hint":null,"details":null,"code":"42P01","message":"relation \"*3*.EH'1/\" does not exist"}
```

### DIFF
--- a/test/Feature/UnicodeSpec.hs
+++ b/test/Feature/UnicodeSpec.hs
@@ -1,0 +1,20 @@
+module Feature.UnicodeSpec where
+
+import Test.Hspec
+import Test.Hspec.Wai
+import Test.Hspec.Wai.JSON
+import Network.Wai (Application)
+import Control.Monad (void)
+
+spec :: SpecWith Application
+spec =
+  describe "Reading and writing to unicode schema and table names" $
+    it "Can read and write values" $ do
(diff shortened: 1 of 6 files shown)


test/Feature/UnicodeSpec.hs
```diff
@@ -0,0 +1,20 @@
...
+
+spec :: SpecWith Application
+spec =
+  describe "Reading and writing to unicode schema and table names" $
+    it "Can read and write values" $ do
+      get "/%D9%85%D9%88%D8%A7%D8%B1%D8%AF"
+        `shouldRespondWith` "[]"
+
+      void $ post "/%D9%85%D9%88%D8%A7%D8%B1%D8%AF"
+        [json| { "هویت": 1 } |]
```
begriffs: Ha, it looks like the right-to-left language actually messes up the Github diff preview itself.

begriffs:
Well, nobody has added any comments so I'll take that as LGTM.

PR #538, opened 2016-03-29
Update postgrest version to 0.3.1.1 in Dockerfile

jpierre03:


begriffs:
Thank you! I've added a note to my release checklist so that I'll remember to increase the version in the Dockerfile next time.

PR #347, opened 2015-11-06
Basic tables/procs for blog demo

begriffs:
Still need the user management views.

PR #543, opened 2016-04-03
Avoid slow PL/pgSQL exception handling in example

begriffs:
Suggestion by @diogob in #347.

### DIFF
--- a/docs/examples/users.md
+++ b/docs/examples/users.md
@@ -456,15 +456,16 @@ Here's a function to get the email of the currently authenticated
 user.
 
 ```sql
+-- Prevent current_setting('postgrest.claims.email') from raising
+-- an exception if the setting is not present. Default it to ''.
+ALTER DATABASE :database_name SET postgrest.claims.email TO '';
+
 create or replace function
 basic_auth.current_email() returns text
   language plpgsql
   as $$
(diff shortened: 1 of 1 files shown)


docs/examples/users.md
```diff
@@ -456,15 +456,16 @@ Here's a function to get the email of the currently authenticated
 user.
 
 ```sql
+-- Prevent current_setting('postgrest.claims.email') from raising
+-- an exception if the setting is not present. Default it to ''.
+ALTER DATABASE :database_name SET postgrest.claims.email TO '';
```
diogob: Maybe it's better to use a fictitious database name instead of using a variable `:database_name`.

docs/examples/users.md
```diff
@@ -456,15 +456,16 @@ Here's a function to get the email of the currently authenticated
 user.
 
 ```sql
+-- Prevent current_setting('postgrest.claims.email') from raising
+-- an exception if the setting is not present. Default it to ''.
+ALTER DATABASE your_db_name SET postgrest.claims.email TO '';
```
begriffs: @diogob this look OK?

diogob: Perfect :D

PR #555, opened 2016-04-07
Adding check for verified flag before login into Users example

michalskop:
I am not sure if it is the best sql way, but it works ok for me.


begriffs:
Thanks for the doc contribution. I've been hesitating whether the verification detail makes the example harder to understand, or whether it will be useful because someone will probably want the logic in there if they are using this for a real API...

begriffs:
@michalskop hey sorry for the delayed response.  I'm thinking of merging this and just wondering if you've tried running the code to be sure it works?

PR #560, opened 2016-04-10
Set role only once, and set it before other GUC vars

begriffs:
I verified in the sql logs that this fixes the order of `set local` statements. From the outside there is no change of behavior so I don't know what regression test I could add.

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -12,8 +12,7 @@ In the test suite there is an example of simple login function that can be used 
 very simple authentication system inside the PostgreSQL database.
 -}
 module PostgREST.Auth (
-    setRole
-  , claimsToSQL
+    claimsToSQL
   , jwtClaims
   , tokenJWT
   ) where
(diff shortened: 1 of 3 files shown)


src/PostgREST/Auth.hs
```diff
@@ -71,8 +70,8 @@ jwtClaims secret input time =
   value2map _          = M.empty
 
 {-| Receives the name of a role and returns a SET ROLE statement -}
-setRole :: Text -> BS.ByteString
-setRole r = "set local role " <> cs (pgFmtLit r) <> ";"
+setRole :: Value -> BS.ByteString
```
begriffs: I'm not pleased with this setRole function, I feel it could be rolled into claimsToSQL if the claims were properly prepared (e.g. non-role claims would get prepended by `postgrest.claims.` and all claims could be unioned with the singleton map {role: anonymous} to create a default value for the role).

begriffs: Then again if `setRole` handles all claims including the role then we'll have to sort role to be the first one, so the code might be less understandable.

src/PostgREST/Middleware.hs
```diff
@@ -42,15 +40,16 @@ runWithClaims conf time app req = do
         if M.null claims && not (null tokenStr)
           then clientErr "Invalid JWT"
           else do
-            let cmdBatch = mconcat $ claimsToSQL claims
-            unless (BS.null cmdBatch) (H.sql cmdBatch)
+            H.sql . mconcat $
+              setRole (
+                fromMaybe anon (M.lookup "role" claims)
```
diogob: shouldn't this ordering and default setting a responsibility of claimsToSQL ? After all it returns an array of commands to be executed, why not return the array with everything already set properly?

begriffs:
Anyone want to take another pass through the code?

ruslantalpa:
It works as expected so good.
While i would make it more explicit that when there is no jwt, only set anon happens, it's not that important and it's not that complicated as it is, ... so merge.

PR #564, opened 2016-04-18
Update reading.md to have note on foreign-keys

davidkuep:
After v0.3.1.1, views containing sub-selected columns will not have any foreign keys automatically found in them, before this, some foreign keys were found (though not always), updated docs to reflect this.

### DIFF
--- a/docs/api/reading.md
+++ b/docs/api/reading.md
@@ -287,7 +287,8 @@ Would embed in the `client` key the row referenced with `client_id`.
 
 <div class="admonition note">
     <p class="admonition-title">Design Consideration</p>
-    <p>In order for this feature to work as expected after a schema change, PostgREST currently requires to be restarted.</p>
+    <p>In order for this feature to work as expected after a schema change, PostgREST currently requires a restart.</p>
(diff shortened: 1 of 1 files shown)


docs/api/reading.md
```diff
@@ -287,7 +287,8 @@ Would embed in the `client` key the row referenced with `client_id`.
 
 <div class="admonition note">
     <p class="admonition-title">Design Consideration</p>
-    <p>In order for this feature to work as expected after a schema change, PostgREST currently requires to be restarted.</p>
+    <p>In order for this feature to work as expected after a schema change, PostgREST currently requires a restart.</p>
+    <p>Additionally, in versions >=0.3.1.1 views containing sub-select statements as columns will not have foreign key relations matched. Sub-select statements may occur in a WHERE clause following a JOIN condition. In previous versions, foreign key relations in views with columns defined with sub-selects were sometimes incidentally recognized.  </p>
```
ruslantalpa: Let's not introduce confusion by mentioning versions and my saying something like "it used to work", it didn't :). That was a bug, while sometimes it made a correct match (by pure chance), most of the time, the relations were badly detected (between tables that did not have any relation).

Let's just say something like "Relation detection (based on foreign keys) are always reliably detected for tables. Views have a limitation, relation detection only works then there are no sub-selects in SELECT/FROM/JOIN part of the query, they are allowed in WHERE clause"

begriffs:
@davidkuep what do you think about Ruslan's suggestion?

begriffs:
(bump)

Issue #554, opened 2016-04-07
row-level security problem in the example from documentation


michalskop:
I implemented the row-level security as described here http://postgrest.com/examples/blog/ (with some changes in my case, but I "translated" the following examples into the documentation example). It works ok for a post with id=1 and author author@example.com if sending:

```
PATCH /posts?id=eq.1
```

with data

```
{"title": "new title"}
```

(it updates)

And it also works ok for another registered author hacker@example.com:

```
PATCH /posts?id=eq.1
```

with data

```
{"title": "hacking"}
```

(it does not updates)

However, it does not work correctly if the other registered author hacker@example.com requests

```
PATCH /posts?id=eq.1
```

with data 

```
{"title": "hacking", "author": "hacker@example.com"}
```

as it updates the post and changes its author to hacker@example.com

Using postgrest/0.3.1.1

begriffs:
😱 Crap. We need a way to tell it to check the original value of `author` against `basic_auth.current_email()` and not whatever value of `author` was just passed in. I'm new to row-level security actually so we'll have to look into it. There's a way with triggers to reject inappropriate changes, but I'd like to be able to use RLS if we can.

begriffs:
I merged the changes by @opensrcken and deployed the docs. Can you check whether it fixes the security hole you discovered?

spinus:
@michalskop did you check that by any chance?
@begriffs  is it still the case in 0.4.0.0?

begriffs:
The auth docs have been stripped down in 0.4. Want to check for me whether you see the problem in https://postgrest.readthedocs.io/en/v0.4/auth.html#sql-user-management ?

begriffs:
Closing this. If you notice that the docs for 0.4 suggest anything insecure please open an issue on the [docs repo](https://github.com/begriffs/postgrest-docs)

PR #579, opened 2016-05-01
fix row level security readme per https://github.com/begriffs/postgre…

opensrcken:
This addresses the issue in #554. The actual security filter needs to be put in the `using` block.

### DIFF
--- a/docs/examples/blog.md
+++ b/docs/examples/blog.md
@@ -84,6 +84,18 @@ drop policy if exists comments_select_unsecure on comments;
 create policy comments_select_unsecure on comments for select
   using (true);
 
+drop policy if exists authors_eigencreate on posts;
+create policy authors_eigencreate on posts for insert
+  with check (
+    author = basic_auth.current_email()
+  );
+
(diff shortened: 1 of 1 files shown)


begriffs:
Thanks for continuing to debug this problem.

This change will have the side effect or preventing people from even reading each others' posts though, right? So it would prevent anonymous users from reading anybody's posts?

kenhahn85:
You are correct. I have updated the branch with the full set of commands I needed to enable `anon` read access to posts and comments. Furthermore, I added row-level security on deletion of blogs and comments. In the spirit of the existing code, I named this policy `authors_eigendelete`.

In updating these examples, I did not consider the case of a superuser, who might have access to editing any post / comment.

begriffs:
Do you think we have to protect `insert` as well as `update` so users cannot insert new posts impersonating other users?

opensrcken:
I just came here to comment on that. :) I'll add that later tonight.

begriffs:
Thanks, you're the best! Glad you figured out how to fix the security of this example.

PR #582, opened 2016-05-05
Accept JWT secrets that have crazy binary values

### LINKED ISSUE #495: Invalid JWT signature created by PostgREST with binary jwt-secret
The verification on [jwt.io](http://jwt.io/) of the JSON Web Tokens generated by PostgREST fails when the configured `jwt-secret` is binary.

I am using latest stable version of PostgREST, `v0.3.0.4`, running on Mac OSX 10.11.1 El Capitan. I have the same issue on a server running Ubuntu 14.04 LTS.

Here are the steps to reproduce:

1. generate a binary secret in a file using openssl (only 8 bytes to keep it simple, but it is unsafe):
  ```sh
openssl rand 8 > jwt.secret
```

2. open the page [jwt.io](http://jwt.io/)

3. encode the binary secret as `base64url` for use in jwt.io:
   ```sh
base64UrlEncode()
{
  base64 |
  sed -e 's/+/-/g' -e 's#/#_#g' -e 's/=//g' # base64 to base64url
}
cat jwt.secret |
base64UrlEncode |
less
```

4. copy the JWT secret encoded in `base64url` format
(or use `pbcopy` instead of `less` at previous step)

5. in the page [jwt.io](http://jwt.io/), paste the JWT secret encoded in `base64url` format in the box labelled "VERIFY SIGNATURE" in "Decoded" section on the right of the page

6. under the input for the JWT secret, check the box ☑ `secret base64 encoded`

7. define a data type and a stored procedure to prepare a JSON Web Token for PostgREST (I have isolated them in a schema named `issue495`):
  ```sql
CREATE SCHEMA issue495;

CREATE TYPE issue495.jwt_claims 
AS ( role TEXT ) 
;

CREATE FUNCTION issue495.get_jwt_for_role(role TEXT)
RETURNS issue495.jwt_claims
RETURNS NULL ON NULL INPUT
LANGUAGE sql
IMMUTABLE
AS 'SELECT role AS role;'
;
```

8. start a new instance of PostgREST which connects to the schema `issue495` and makes the stored procedure available on port 4495 (for example) of localhost, using the binary `jwt-secret`:
  ```sh
dbuser="$(whoami)";
postgrest "postgres://$dbuser@localhost:5432/egull" \
  --anonymous "$dbuser" \
  --schema issue495 \
  --port 4495 \
  --jwt-secret "$(cat jwt.secret)"
```

9. POST a request to get the JSON Web Token for the role `issue495`:
  ```sh
curl -s -X POST \
  -d '{"role":"issue495"}'  \
  'http://localhost:4495/rpc/get_jwt_for_role' |
cut -d'"' -f4 | # cut the value of the "token" property
less
```

10. copy the JSON Web Token
(or use `pbcopy` instead of `less` at previous step)

11. paste the JSON Web Token in the input labelled "Encoded" on the left of the page

The result of the validation is displayed in a red bar below: "invalid signature".

begriffs:
This PR converts the secret right to ByteString rather than to Text so that sequences of binary values which are invalid unicode are allowed. For instance the output of `openssl rand 8`.

This does not yet have a regression test.

Fixes #495.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -39,7 +39,7 @@ executable postgrest
                      , hasql-transaction >= 0.4.3 && < 0.5
                      , http-types
                      , interpolatedstring-perl6
-                     , jwt
+                     , jwt >= 0.7.1 && < 8
                      , lens >=3.8 && < 5.0
                      , lens-aeson >= 1.0.0.0 && < 1.1.0.0
                      , mtl
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -32,7 +32,7 @@ import           Options.Applicative
 import           Paths_postgrest             (version)
 import           Prelude
 import           Safe                        (readMay)
-import           Web.JWT                     (Secret, secret)
+import           Web.JWT                     (Secret, binarySecret)
 
 -- | Data type to store all command line options
 data AppConfig = AppConfig {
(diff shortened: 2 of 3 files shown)


begriffs:
This pull request is on hold as we discuss problems with this approach from within the linked issue.

begriffs:
It turns out this will not be sufficient because any particular string conversion I pick will fail on certain inputs. It will either mangle valid unicode, or try parsing random bytes as if they were unicode. The solution is reading the secret out of a file, byte for byte. Changing the jwt parameter is backwards incompatible and will be left for the next major version.

PR #586, opened 2016-05-10
Ability to rename columns/nodes in the output and support "-" in column names

### LINKED ISSUE #310: GraphQL revisited
meme7b7b94473d7dcc19

I am quite interested in the new (frontend) technologies coming out of FB and GraphQL + Relay + React is a very powerful combination.

I want to propose to steer the capabilities of postgresql to be isomorphic to a graphql server.
After the addition of &select with it's nested capabilities, the "read part" has the same power as a GQL server

compare this
```
{
  projects (active: true) {
    id,
    name,
    client {
        id,
        name
    },
    tasks (completed: false) {
        id,
        name
    }
  }
}
``` 

to this
```
/projects?select=id,name,client(id,name),tasks(id,name)&active=eq.true&tasks.completed=eq.false
```

The other lacking part is "mutations", in GQL you can say "make this mutation then return this data".
In postgrest you can say "Prefer:return=representation", we could extend it to something like "return=id,name,client(id) ..."

With this additional capability, i think (more or less) postgresql can act as a graphql server (has the core capabilities).
Now the main question is, how to make postgresql understand a graphql query.
The first approach would be to code it in through a parser which would not be too hard to do but i hope it would be unnecessary.

When looking at relay (frontend client that talks to a graphql server) i noticed a part of code called RelayNetworkLayer.
I think it provides the capabilities to intercept a GQL query and translate it into a postgrest request.

Please look at [this video] (https://www.youtube.com/watch?v=Pxdgu2XIAAg) to get a sense of why i'm advocating graphql support. 

PS: Did someone get the reference of the image? :)

ruslantalpa:
With this PR now you can do this
`/projects?id=eq.1&select=myId:id, name, project_client:client_id{*}, project_tasks:tasks{id, name}`

to get this type of output
```
{
   "myId":1,
   "name":"Windows 7",
   "project_client":{
      "id":1,
      "name":"Microsoft"
   },
   "project_tasks":[
      {
         "id":1,
         "name":"Design w7"
      },
      {
         "id":2,
         "name":"Code w7"
      }
   ]
}
```

This can also close #310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Reload database schema on SIGHUP - @begriffs
+- Support "-" in column names - @ruslantalpa
 
 ### Fixed
+- Set transaction mode to READ when possible to support connecting to read replicas - @ruslantalpa
 
 - Omit Content-Type header for empty body - @begriffs
 - Prevent role from being changed twice - @begriffs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
-
 - Reload database schema on SIGHUP - @begriffs
 - Support "-" in column names - @ruslantalpa
+- Support column/node renaming `alias:column` - @ruslantalpa
 
 ### Fixed
-- Set transaction mode to READ when possible to support connecting to read replicas - @ruslantalpa
-
 - Omit Content-Type header for empty body - @begriffs
 - Prevent role from being changed twice - @begriffs
 - Use read-only transaction for read requests - @ruslantalpa
--- a/docs/api/reading.md
+++ b/docs/api/reading.md
@@ -258,7 +258,8 @@ but the the select query is recursive. You could for instance specify
 GET /foo?select=x, y, bar{z, w, baz{*}}
 ```
 
-You can select not only using table names, but also column names!
+You can select not only using table names, but also foreign key column names!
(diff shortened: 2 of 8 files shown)


begriffs:
Looks like this PR includes duplication from the existing commit 5aadfba84bdd15723423316984b15a13b1599b01 for the read-only transaction stuff. If you rebase on master I think those commits will become empty and you can remove them.

ruslantalpa:
this branch is in sync with my master, the master is the problem.
That happened because the previous pr that i submitted (and you manually included it on your side and then i closed the pr) was done on my master, so that commit you see is from that pr.
Not sure how to remove it, don't want to mess things up, can we just merge this as it is (after the review), it will be just one "commit" without anything

begriffs:
Try this

```bash
git remote add upstream https://github.com/begriffs/postgrest.git
git fetch upstream

# on your pr branch
git rebase upstream/master
# now force push the results to your remote pr branch
git push -f
```

ruslantalpa:
did what you asked, there are only 3 commits now, a bit cleaner, but there is still that old commit (all it's changes were overwritten, it's not needed) but not sure how to just "delete" a commit

begriffs:
Try `git rebase -i master` and your editor will pop up with the commits. Delete the line containing the offending commit and it will disappear from your branch. Then you can force push again.

ruslantalpa:
can not squash without a previous commit ...

ruslantalpa:
finally ...

begriffs:
This is really nice, tests and everything. Could you include a note in the [docs](https://github.com/begriffs/postgrest/blob/master/docs/api/reading.md) in this PR as well about the feature?

ruslantalpa:
Sure, tomorrow :)

PR #590, opened 2016-05-15
Return proper 401/403 when access denied

### LINKED ISSUE #584: Add flag to reveal access denied errors
Currently PostgreSQL error code 42501 (insufficient privilege) is mapped to a 404 http response.
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Error.hs#L91

This was designed to hide endpoints from potential attackers. However unexpected 404s can be confusing when debugging, and perhaps security by obscurity is not very robust anyway.

Suggestion: add a `--reveal-forbidden` command line flag. When the flag is present and the database raises an access denied, have postgrest return
* **401 Unauthorized** if no JWT token was provided with the request
* **403 Forbidden** if a JWT was provided

begriffs:
When PostgreSQL returns code `42501` for insufficient privilege we now return an access denied status code rather than a 404. If the request did not provide a JWT token with role claim then we return 401 Unauthorized with a `WWW-Authenticate` header. When the request does have a JWT with role then we return 403 Forbidden.

Fixes #584

As decided in the referenced issue we are changing the behavior unconditionally, no need for a special command-line flag.

### DIFF
--- a/test/fixtures/privileges.sql
+++ b/test/fixtures/privileges.sql
@@ -17,6 +17,7 @@ GRANT ALL ON TABLE
     , comments
     , complex_items
     , compound_pk
+    , empty_table
     , has_count_column
     , has_fk
     , insertable_view_with_join
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Support column/node renaming `alias:column` - @ruslantalpa
 
 ### Fixed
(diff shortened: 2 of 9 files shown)


test/fixtures/privileges.sql
```diff
@@ -17,6 +17,7 @@ GRANT ALL ON TABLE
     , comments
     , complex_items
     , compound_pk
+    , empty_table
```
begriffs: The previous 404/401 subtlety had even caused a problem with a test which was getting 404 because of an access denied rather than directly as a result of there being no data.

PR #592, opened 2016-05-16
Accept POST requests from HTML forms

begriffs:
Allows inserts or RPC calls via HTML forms like

```html
<form method="post" action="https://myapi.com/people">
  <input type="text" name="fullname" />
  <input type="text" name="dob" />
  <!-- the other columns -->
  <input type="submit" value="Create New">
</form>
```

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -122,6 +122,9 @@ userApiRequest schema req reqBody =
               Nothing -> PayloadParseError "All lines must have same number of fields"
               Just json -> PayloadJSON json)
             (CSV.decodeByName reqBody)
+        -- This is a Left value because form-urlencoded is not a content
+        -- type which we ever use for responses, only something we handle
+        -- just this once for requests
         Left "application/x-www-form-urlencoded" ->
           PayloadJSON . UniformObjects . V.singleton . M.fromList
                       . map (cs *** JSON.String . cs) . parseSimpleQuery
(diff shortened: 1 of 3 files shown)


begriffs:
One thing which makes this functionality somewhat awkward is that after submitting the form the browser will arrive at a blank page because the default API response has no body.

My original motivation for this feature is a simple email list signup form.

src/PostgREST/ApiRequest.hs
```diff
@@ -122,6 +122,10 @@ userApiRequest schema req reqBody =
               Nothing -> PayloadParseError "All lines must have same number of fields"
               Just json -> PayloadJSON json)
             (CSV.decodeByName reqBody)
+        Left "application/x-www-form-urlencoded" ->
```
ruslantalpa: This seems like it should be handled in pickContentType and get a "Right" value from there

begriffs: That's an interesting point and one I wrestled with. The issue is that `ContentType` is used not only for code involving the request Content-Type header as this PR is, but also code that checks the Accept header and chooses the *response* Content-Type header. Extending `ContentType` with a `FormUrlEncoded` constructor would mean I have to add a case in the show instance for it like

```haskell
data ContentType = ApplicationJSON | TextCSV | FormUrlEncoded deriving Eq
instance Show ContentType where
  show ApplicationJSON = "application/json; charset=utf-8"
  show TextCSV         = "text/csv; charset=utf-8"
  show FormUrlEncoded  = "application/x-www-form-urlencoded; charset=utf-8"
```

but this seems misleading, like we may be planning to use it in responses or something. It is possible to do this and leave the `unsupportedAccept` middleware as it is now which would prevent people from choosing form urlencoding in an Accept header. However the code communicates the wrong intention. Either way seems awkward and I think I got myself into a corner with the types as I have defined them.

ruslantalpa: got it.
In this case i think it is ok as it is now, maybe just a comment for this special case and why the choice was made.
It's not worth putting in more work/lines of code to support this feature and it's only a minimal change.

PR #593, opened 2016-05-16
Fix Windows build issue 589

dbc60:
A small change in Main.hs to fix the Windows build. I moved Data.IORef outside of the block of code covered by `ifndef mingw32_HOST_OS` and deleted line 14, `Control.Monad.IO.Class` because that generated a warning about the import being redundant and the warning was treated as an error.

### DIFF
--- a/src/PostgREST/Main.hs
+++ b/src/PostgREST/Main.hs
@@ -11,7 +11,6 @@ import           PostgREST.Config                     (AppConfig (..),
 import           PostgREST.DbStructure
 
 import           Control.Monad
-import           Control.Monad.IO.Class               (liftIO)
 import           Data.Monoid                          ((<>))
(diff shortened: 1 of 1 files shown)


begriffs:
Thanks for diving into Haskell to help!

One thing I notice is that this PR fails on CI with:

```
/home/ubuntu/postgrest/src/PostgREST/Main.hs:82:9:
    Not in scope: ‘liftIO’
```

Rather than removing the following include, I bet it will work if you put the import inside the `ifndef`:

```haskell
import Control.Monad.IO.Class (liftIO)
```

dbc60:
Yes, thanks! Moving `import Control.Monad.IO.Class (liftIO)` to inside the `ifndef` was exactly the problem.

PR #594, opened 2016-05-16
Fix include entities from the same parent table using two different foreign keys

ruslantalpa:
This type of query was generating a bad sql
`/orders?id=eq.1&select=id, name, billing_address_id{id, name}, shipping_address_id{id, name}`
The PR fixes that


begriffs:
Love it, more tests passing with less code!

Issue #588, opened 2016-05-14
Non–URL-safe characters in Location header for 201 Created


league:
I'm just toying with postgrest for now, trying the [film festival tutorial](http://postgrest.com/examples/start/). I built from source using stack on the git master (0dc33db).

I was surprised to find that the `Location` header returned in a 201 Created response for a single entity was not properly URL-encoded, and therefore could not directly be used for retrieval. Here's a transcript using `curl`.

````
% curl -v --crlf --data-binary @- -H "Content-Type: text/csv" http://localhost:3000/festival
name
New Haven Arts & Ideas
^D
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 3000 (#0)
> POST /festival HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.48.0
> Accept: */*
> Content-Type: text/csv
> Content-Length: 28
> 
* upload completely sent off: 28 out of 28 bytes
< HTTP/1.1 201 Created
< Transfer-Encoding: chunked
< Date: Sat, 14 May 2016 13:18:05 GMT
< Server: postgrest/0.3.1.1
< Location: /festival?name=eq.New Haven Arts & Ideas
< 
* Connection #0 to host localhost left intact
````

When I try to `GET` using that provided `Location`, it doesn't work:

````
% curl 'http://localhost:3000/festival?name=eq.New Haven Arts & Ideas'
Bad Request
````

I need to turn the spaces into `+` and also encode the ampersand for it to work:

````
% curl 'http://localhost:3000/festival?name=eq.New+Haven+Arts+%26+Ideas'
[{"name":"New Haven Arts & Ideas"}]
````

Seems to me that the value of the `Location` header should be ready to use as it is. Of course, this doesn't occur with numeric primary keys that are used most often in practice, but since this was the example in the tutorial, I happened to notice it. Thanks!

league:
P.S., I have Haskell-fu so if there's agreement that this is a bug, I could take a shot at a PR.

ruslantalpa:
I think you'll need some sql foo to fix this since the location string is calculated in the db :) (if i remember correctly) but yes, it's a bug

league:
Unfortunately, Postgres doesn't seem to have a urlencode function built in or in any standard extension. Here are two (slow) implementations using plpgsql:
  - http://stackoverflow.com/a/33616365
  - http://stackoverflow.com/a/10322075

[...]

PR #595, opened 2016-05-17
URL-encode Location header in 201 response (#588)

league:


src/PostgREST/App.hs
```diff
@@ -24,6 +24,7 @@ import           Text.ParserCombinators.Parsec (parse)
 
 import           Network.HTTP.Types.Header
 import           Network.HTTP.Types.Status
+import           Network.HTTP.Types.URI
```
ruslantalpa: maybe specify exactly the imported functions

test/Feature/InsertSpec.hs
```diff
@@ -145,7 +145,7 @@ spec = do
           `shouldRespondWith` ResponseMatcher {
             matchBody    = Just inserted
           , matchStatus  = 201
-          , matchHeaders = ["Location" <:> [str|/json?data=eq.{"foo":"bar"}|]]
+          , matchHeaders = ["Location" <:> "/json?data=eq.%7B%22foo%22%3A%22bar%22%7D"]
```
ruslantalpa: You mentioned somewhwere that there is a problem getting back the record with the value encoded, does it work now as it is?

ruslantalpa:
Although this quick fix works for now, and since the location string is not calculated entirely on the db side maybe it would make sense to get back from sql an array of some sorts for the location (instead of a string) and have hasql decode it into haskell types and then later recombine them into an url.
That is if there is a significant speedup or correctness to worth the effort.

To summarise, if the returned location string can always be used to get back the record then this PR looks good.

test/Feature/InsertSpec.hs
league: No, JSON keys aren't working for me, whether it's URL-encoded or not, and whether it's a list or an object. I'll add tests for those.

ruslantalpa:
I just checked on your branch why the tests fail,
short version is that `json` type does not provide a equality operator ... i.e. it's useless as a primary key :)
the generated query is this
```
WITH pg_source AS (
        SELECT  "test"."json".* 
        FROM  "test"."json"  
        WHERE  "test"."json"."data" = '{"foo":"bar"}'::unknown  
        ) 
SELECT (SELECT pg_catalog.count(1) FROM  "test"."json" WHERE  "test"."json"."data" = '{"foo":"bar"}'::unknown ) AS total_result_set, pg_catalog.count(t) AS page_total, NULL::text[] AS header, coalesce(array_to_json(array_agg(row_to_json(t))), '[]')::character varying AS body
FROM ( SELECT * FROM pg_source LIMIT ALL OFFSET 0) t ",100,,"
```

which is correct.
I tried just changing the column to jsonb but that breaks other tests so i say we should just remove the additional tests that check you can get back the record by location and merge this PR (if everything looks good, i noticed there is a new commit and i did not look at it)

ruslantalpa:
Looked over the latest commit. Looks good.
I say remove the new failing part of the tests and it's good to merge

league:
Great. I rebased to current master and squashed some of the churn – hope that worked okay, I'm not an expert yet on git-rebase! 0a5ddc4 is the one we want, but the PR is showing other commits from master too, maybe that's normal.

begriffs:
@league try this

```bash
git remote add upstream https://github.com/begriffs/postgrest.git
git fetch upstream
git rebase upstream/master
```

(If that doesn't work we can try a `reset --hard` and a `cherry-pick`, but let's see)

ruslantalpa:
For rebase add -i parameter, it will give you the chance to squash all commits into one

ruslantalpa:
Oh, you already did that :)

league:
Yeah, I think this is clean now, and squashed to just the one commit. The other 5 it's showing are already in master. Not sure if there was a way to fetch upstream and rebase that would have prevented that.

begriffs:
OK here's a way to fix up your branch

```bash
git remote add upstream https://github.com/begriffs/postgrest.git
git fetch upstream
git reset --hard upstream/master
git cherry-pick 0a5ddc4
git push -f origin master
```

league:
Ah, that's more like it. I'm not sure what went wrong the first time. Possibly I did a pull instead of fetch from upstream, before doing the rebase -i.

ruslantalpa:
you should add a CHANGELOG line

src/PostgREST/QueryBuilder.hs
```diff
@@ -61,9 +61,22 @@ import           Data.Scientific         ( FPFormat (..)
 import           Prelude hiding          (unwords)
 import           PostgREST.ApiRequest    (PreferRepresentation (..))
 
+{-| The generic query result format used by API responses. The location header
+    is represented as a list of strings containing variable bindings like
+    @"k1=eq.42"@. If unused, it's null/Nothing rather than the empty list
+    because 'PostgreSQL.Binary.Decoder.arrayDimension' cannot decode an empty
+    array!
```
begriffs: Very helpful comment.

Do you know if this is a bug in `arrayDimension`? It looks like an intentional library decision: https://github.com/nikita-volkov/postgresql-binary/blob/da82acf62855012ddecdcd6378d8fba69ca77b31/library/PostgreSQL/Binary/Decoder.hs#L499-L502

@nikita-volkov can you give us more context about this function?

src/PostgREST/App.hs
```diff
@@ -235,6 +235,14 @@ app dbStructure conf apiRequest =
                                           status = rangeStatus frm to (toInteger <$> tableTotal)
                                       in (status, contentRange)
 
+splitKeyValue :: BS.ByteString -> (BS.ByteString, BS.ByteString)
+splitKeyValue kv = (k, BS.tail v)
+  where (k, v) = BS.break (== '=') kv
```
begriffs: Initially I worried about `=` signs in the value itself but it's fine, since `break` splits around the first one only. Good to know about this function, I hadn't noticed it in base.

begriffs:
@league thanks for this work, it's a significant improvement for inserts. Also really like your Haskell coding style. I'll merge this as is and we can improve the situation with `arrayDimension` later if the postgresql-binary package changes.

src/PostgREST/QueryBuilder.hs
nikita-volkov: Can please provide some context of the issue?

league: Hi @nikita-volkov, we were doing a `SELECT array_agg(...)` and decoding it into `[ByteString]`. But sometimes the query will produce zero rows, so the aggregator function should return an empty array `{}::text[]`. I expected that to decode to the empty list, but instead it produces `failure "A missing dimension length"` in `PostgreSQL.Binary.Decoder.arrayDimension`.

In the commit shown here, we just worked around it by decoding to `Maybe [ByteString]` instead, and that works fine.

We're using `Hasql.Decoders` (`HD`) which in turn calls `PostgreSQL.Binary.Decoder`. Something like this:

```haskell
-- works, producing Maybe [ByteString]:
HD.nullableValue $ HD.array $ HD.arrayDimension replicateM $ HD.arrayValue HD.bytea
-- fails with "A missing dimension length" when aggregated result set is empty:
HD.value $ HD.array $ HD.arrayDimension replicateM $ HD.arrayValue HD.bytea
```

nikita-volkov: Okay. Thanks. I'll look into that and post back.

nikita-volkov: It seems like you just need to upgrade your Hasql dependency bounds, because the latest releases have included [a fix for that](https://github.com/nikita-volkov/hasql/issues/46).

begriffs: @nikita-volkov perfect!

@league I'll make a PR to update all deps to their newest version including hasql stuff. Once that's merged could you send a PR to refactor this part of the code?

league: @begriffs Yeah, ping me from that dependency PR, if you remember.

PR #597, opened 2016-05-18
Tweak .cabal to avoid unneeded recompilation

league:
Previously when making a change and running `stack test`, it would
compile each module 3 times: for the library, the executable, and the
test suite.

This change more cleanly segregates the hs-source-dirs for each target,
which avoids recompilation. The only source change is moving Main.hs
into its own directory (but it's otherwise unchanged). See also:
<http://stackoverflow.com/questions/6711151/how-to-avoid-recompiling-in-this-cabal-file>


begriffs:
Oh no way this is great! I've been doing this wrong in every project, thanks for the help.

PR #602, opened 2016-05-20
Ability to order embedded items (closes #509)

### LINKED ISSUE #509: Ordering of foreign entities
In [embedding foreign entities](/begriffs/postgrest/blob/master/docs/api/reading.md#embedding-foreign-entities) we learn that we can “ask for a client and all of their projects” with `GET /clients?id=eq.42&select=id,name,projects{*}` .

In [ordering](/begriffs/postgrest/blob/master/docs/api/reading.md#ordering) the url syntax `GET /people?order=age.desc,height.asc` is presented and references only columns from the main entity. It is then extended to `GET /people?order=age.desc.nullslast` suggesting that using the dot to drill down into foreign entities' columns is not going to work.

What is the story for ordering embedded foreign entities? I hope I am not missing something simple.

Thank you.

ruslantalpa:
like so
`/projects?select=id,name,tasks{id,name}&order=id.asc&tasks.order=name.ask`

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -77,8 +77,8 @@ data ApiRequest = ApiRequest {
   , iFilters :: [(String, String)]
   -- | &select parameter used to shape the response
   , iSelect :: String
-  -- | &order parameter
-  , iOrder :: Maybe String
+  -- | &order parameters for each level
+  , iOrder :: [(String,String)]
   -- | Alphabetized (canonical) request query string for response URLs
   , iCanonicalQS :: String
   -- | JSON Web Token
@@ -147,11 +147,11 @@ userApiRequest schema req reqBody =
   , iPreferRepresentation = representation
   , iPreferSingular = singular
   , iPreferCount = not $ singular || hasPrefer "count=none"
-  , iFilters = [ (k, fromJust v) | (k,v) <- qParams, k `notElem` ["select", "order"], isJust v ]
+  , iFilters = [ (cs k, fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn "order" k) ]
   , iSelect = if method == "DELETE"
               then "*"
               else fromMaybe "*" $ fromMaybe (Just "*") $ lookup "select" qParams
-  , iOrder = join $ lookup "order" qParams
+  , iOrder = [(cs k, fromJust v) | (k,v) <- qParams, isJust v, endingIn "order" k ]
   , iCanonicalQS = urlEncodeVars
      . sortBy (comparing fst)
      . map (join (***) cs)
@@ -181,6 +181,9 @@ userApiRequest schema req reqBody =
   tokenStr = case T.split (== ' ') (cs auth) of
     ("Bearer" : t : _) -> t
     _                  -> ""
+  endingIn:: T.Text -> T.Text -> Bool
+  endingIn word key = word == lastWord
+    where lastWord = last $ T.split (=='.') key
 
 -- PRIVATE ---------------------------------------------------------------
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -280,10 +280,10 @@ augumentRequestWithJoin schema allRels request =
 
 buildReadRequest :: [Relation] -> ApiRequest -> Either Text ReadRequest
 buildReadRequest allRels apiRequest  =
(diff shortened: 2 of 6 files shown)


src/PostgREST/ApiRequest.hs
```diff
@@ -77,8 +77,8 @@ data ApiRequest = ApiRequest {
   , iFilters :: [(String, String)]
   -- | &select parameter used to shape the response
   , iSelect :: String
-  -- | &order parameter
-  , iOrder :: Maybe String
+  -- | &order parameters for each level
```
begriffs: Thanks for updating the comments, they tend to get out of date.

src/PostgREST/ApiRequest.hs
```diff
@@ -147,11 +147,11 @@ userApiRequest schema req reqBody =
   , iPreferRepresentation = representation
   , iPreferSingular = singular
   , iPreferCount = not $ singular || hasPrefer "count=none"
-  , iFilters = [ (k, fromJust v) | (k,v) <- qParams, k `notElem` ["select", "order"], isJust v ]
+  , iFilters = [ (cs k, fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn "order" k) ]
   , iSelect = if method == "DELETE"
               then "*"
               else fromMaybe "*" $ fromMaybe (Just "*") $ lookup "select" qParams
-  , iOrder = join $ lookup "order" qParams
+  , iOrder = [(cs k, fromJust v) | (k,v) <- qParams, isJust v, endingIn "order" k ]
```
begriffs: That's pretty elegant. I also like the `foo.order=bar` syntax since `order=baz` is then a special case and it makes sense.

src/PostgREST/App.hs
```diff
@@ -280,10 +280,10 @@ augumentRequestWithJoin schema allRels request =
 
 buildReadRequest :: [Relation] -> ApiRequest -> Either Text ReadRequest
 buildReadRequest allRels apiRequest  =
-  augumentRequestWithJoin schema rels =<< first formatParserError (foldr addFilter <$> (addOrder <$> readRequest <*> ord) <*> flts)
+  augumentRequestWithJoin schema rels
+  =<< first formatParserError (foldr addFilter <$> (foldr addOrder <$> readRequest <*> ords) <*> flts)
```
begriffs: I actually didn't know that this indentation is valid Haskell. Might be clearer with the second line indented like

```haskell
augumentRequestWithJoin schema rels =<<
  first formatParserError
    (foldr addFilter <$> (foldr addOrder <$> readRequest <*> ords) <*> flts)
```

I suspect that last bit might benefit from EitherT or something but I don't know enough about how that works.

ruslantalpa: i can move the `=<<` up if you want but i think it should not be indented since `argument ...` and `first` are two computations happening at the same level, it's not like `first` is inside `argument`

ruslantalpa: but yes, that line i guess could be written in a nicer way if ai knew any haskell :)
But i think it's clear enough what's happening on that line so maybe let's leave the refactoring for some other day since any way i see some patterns emerging that would allow some overall refactoring.

ruslantalpa: as a sidenote, this is what's going in this line
`ReadRequest` is a tree that comes out of parsing the &select parameter.
Then we first add to that tree the `order` parameters, then we add `filters` and at the end, we look at the "relations" between the tree nodes and we add the necessary filters/conditions in order for the joins to work.
At the end we have a tree that we can just turn into a query, all the data is in it

ruslantalpa: and if you look at `=<<` as nothing more then monadic function composition, then you see why the indentation works as is :) (at least that's how i look at it)

begriffs: Hm, OK. The way it's written things are kind of broken apart, like `flts` is far apart from `addFilter`. In any case seems to work fine.

begriffs:
Great feature and love all the tests.

src/PostgREST/App.hs
begriffs: The main thing is that beginning a new line with `=<<` at the same indentation as the previous line looks strange. I think it's better as one line even, or with some kind of indentation. Up to you.

ruslantalpa: i'll move the arrow up

nerfpops:
This is great, can't wait for the next release.

PR #603, opened 2016-05-21
jwtClaims should always return Left for invalid JWT

diogob:
It's a tiny refactor but I believe the code is better like this.
Before this PR `jwtClaims` was returning a Right value in some invalid JWT cases, so we had to double check in our `Middleware` using a clumsy `if`.
It makes more sense to always return a `Left` value in error cases.
The PR also improves readability as the case for empty JWTs is now handled separately.

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -25,7 +25,7 @@ import           Data.Aeson.Types        (parseMaybe, emptyObject, emptyArray)
 import qualified Data.ByteString         as BS
 import qualified Data.Vector             as V
 import qualified Data.HashMap.Strict     as M
-import           Data.Maybe              (fromMaybe, maybeToList)
+import           Data.Maybe              (fromMaybe, maybeToList, fromJust)
 import           Data.Monoid             ((<>))
 import           Data.String.Conversions (cs)
 import           Data.Text               (Text)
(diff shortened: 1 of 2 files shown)


src/PostgREST/Auth.hs
```diff
@@ -52,22 +52,20 @@ claimsToSQL claims = roleStmts <> varStmts
 {-|
   Receives the JWT secret (from config) and a JWT and
   returns a map of JWT claims
-  In case there is any problem decoding the JWT it returns Nothing.
+  In case there is any problem decoding the JWT it returns an error Text
```
begriffs: Oh thanks

src/PostgREST/Middleware.hs
```diff
@@ -30,10 +30,7 @@ runWithClaims :: AppConfig -> Either Text (M.HashMap Text Value) ->
 runWithClaims conf eClaims app req =
   case eClaims of
     Left e -> clientErr e
-    Right claims ->
-      if M.null claims && not (null $ iJWT req)
-        then clientErr "Invalid JWT"
-        else do
+    Right claims -> do
```
begriffs: Definitely see what you mean, If we've got a `Right` things should be right. 😄

src/PostgREST/Auth.hs
```diff
@@ -52,22 +52,20 @@ claimsToSQL claims = roleStmts <> varStmts
...
-jwtClaims secret input time =
-  case mClaims of
-    Nothing -> Right M.empty
-    Just claims -> do
-      let mExp = claims ^? key "exp" . _Integer
-          expired = fromMaybe False $ (<= time) . fromInteger <$> mExp
-      if expired
-        then Left "JWT expired"
-        else Right (value2map claims)
+jwtClaims _ "" _ = Right M.empty
```
begriffs: Yeah good idea to just take care of this up here...

PR #604, opened 2016-05-21
Run GC every 2s rather than 0.3s

### LINKED ISSUE #565: Strange increase in CPU-usage over time
We have a server running on Debian with PostgreSQL 9.5 and latest (I think) postgrest.

The server is more or less just idling since it is a production server not yet in production.

Over time I have seen that the cpu-usage is increasing. After some weeks it is using more than 50 % of 400 % in total (4 cores). I can see on the graphs that the usage is increasing very slowly day by day. 

I have 2 instances of postgrest runing on the server.

I see the same thing on our test server, but that one is under some load during this test phase. But also when it is idling it uses a lot of CPU.

When I restart the postgrest instances things go back to normal.

begriffs:
Fixes #565

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use read-only transaction for read requests - @ruslantalpa
 - Include entities from the same parent table using two different foreign keys - @ruslantalpa
 - Ensure that Location header in 201 response is URL-encoded - @league
+- Fix garbage collector CPU leak - @ruslantalpa et al.
 
 ## [0.3.1.1] - 2016-03-28
 
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -24,7 +24,10 @@ Flag CI
 executable postgrest
   main-is:             Main.hs
   default-extensions:  OverloadedStrings, ScopedTypeVariables, QuasiQuotes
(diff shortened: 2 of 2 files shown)


begriffs:
Looks like cabal check is confusing the RTS option with a ghc-option:

```
stack exec -- cabal check
The following warnings are likely affect your build negatively:
* Instead of 'ghc-options: -I2"' use 'include-dirs: 2"'

Hackage would reject this package.
```

Or maybe grouped quoted options is not valid in a cabal file. I'll consult the docs...

begriffs:
This builds and passes `cabal check`. Looking at its eventlog the GC doesn't run much as I use the program. Although going back to the original rtsopts the GC didn't seem to run much either.

Also if I change it to `"-with-rtsopts=-N -K8"` it will not crash the program whereas running with `+RTS -K8 -RTS` will. So I'm not sure it's actually using this configuration.

begriffs:
Turns out I had to run `stack clean` for the rtsopts to take effect in a new build. Merging.

eric-brechemier:
> Turns out I had to run `stack clean` for the rtsopts to take effect in a new build. Merging.

Is it required for those of us building PostgREST from source?

ruslantalpa:
my guess is yes

begriffs:
To pick up changed GHC flags it's better to clean to be on the safe side.

PR #605, opened 2016-05-21
Replace lens dependency for microlens

diogob:
We don't need the whole Lens library. The difference in compilation time is minimal though, but if you guys don't think it's worth changing I totally understand.
Compiling from scratch with lens:
screen shot 2016-05-21 at 2 26 06 pm

Compiling from scratch with microlens:
screen shot 2016-05-21 at 2 04 31 pm

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -40,8 +40,8 @@ executable postgrest
                      , http-types
                      , interpolatedstring-perl6
                      , jwt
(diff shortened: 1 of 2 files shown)


ruslantalpa:
the less dependencies the better

begriffs:
Seems great to me and it passes CI so merging!

PR #606, opened 2016-05-22
Newdeps

begriffs:
Upgrade the deps so @league is able to refactor sql array decoding.

One thing I did here is pin the hasql dependencies to exact versions so that `packdeps` can warn us in CI when newer versions come out on hackage. Will this be a problem for other programs which may want to use postgrest as a library, or will it (hopefully) affect only building the postgrest binary?

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -34,9 +34,9 @@ executable postgrest
                      , containers
                      , contravariant
                      , errors
-                     , hasql >= 0.19.9 && < 0.20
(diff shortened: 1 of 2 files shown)


league:
My limited understanding is that tight version bounds here _would_ impact other programs building on postgrest as a library. Although since it's (arguably) primarily an executable, maybe that's a small niche and we can wait for someone to complain. :)

begriffs:
Good point, we can fix the version constraints if it causes anyone problems.

PR #607, opened 2016-05-22
Simplify serialization of location header

league:
Possible after bug fix in a Hasql that we picked up with the new
dependency bounds in #606.


league:
**Wait** — better delay this for a moment. I realized that this change might be generating the Location header even when it's not appropriate, when there are no fields to include in it, because previously there were cases where `header` could be `[]`. Give me a bit to look into that, and maybe add a test case that ensures there **is no** location header when there shouldn't be.

league:
Okay, ready now. The initial fix did _always_ generate the Location header. Now it omits it when the field list is empty.

begriffs:
Nice thorough work and test, thanks.

PR #608, opened 2016-05-24
Limit embeded items

ruslantalpa:
This PR allows limiting embedded items and also enforces globally `--max-rows` setting.
In addition, it's possible now to specify the top level limit/range by using `&limit` and `&offset` (but Range header overrides them).
Just like for ordering, you specify the limit using `tree.path.limit=10`, however `offset` can only be specified for the top level (it does not make sense for the embedded items)

The code still needs a lot of cleanup (mainly because i don't fully understand the Range datatype, i just "guest" how it works, so i "improvised" some functions to fit my needs) but it works as it should and the tests pass.

Mainly i would like help with cleaning up/simplifying things in `ApiRequest.hs` and `RangeQuery.hs`, the rest of the files i will cleanup.

### DIFF
--- a/docs/api/reading.md
+++ b/docs/api/reading.md
@@ -172,9 +172,9 @@ GET /people?order=age.nullsfirst
 GET /people?order=age.desc.nullslast
 ```
 
-To filter the embedded items, you need to specify the tree path for the order param like so. 
+To order the embedded items, you need to specify the tree path for the order param like so.
 ```HTTP
-GET /projects?select=id,name,tasks{id,name}&order=id.asc&tasks.order=name.ask
+GET /projects?select=id,name,tasks{id,name}&order=id.ask&tasks.order=name.ask
 ```
 
 
@@ -213,6 +213,15 @@ Range: 0-4
 
 You can also use open-ended ranges for an offset with no limit:
 `Range: 10-`.
+
+In addition to the `Range` header, you can use `&limit` and `&offset` parameters
+to achieve the same result.
+
+You can also set a limit (but not offset) for the embedded items like so
+```HTTP
+/posts?select=id,title,body,comments{id,email,body}&limit=10&comments.limit=3
+```
+The above request will return the first 10 posts and for each of the posts, 3 comments at most
 
 #### Suppressing Counts
 
@@ -309,6 +318,13 @@ GET /orders?id=eq.1&select=orderId:id, customer:customer_id{customerId:id, custo
   }
 ]
 ```
+
+If you want to apply filters to the embedded items, you can do that like so:
+```HTTP
+GET /clients?id=eq.42&select=id,name,projects{id,name,is_active}&projects.is_active=eq.true
+```
+The above request will return the client with id=42 and all the projects for that client that are still active
+
 
 <div class="admonition note">
     <p class="admonition-title">Design Consideration</p>
--- a/docs/api/reading.md
+++ b/docs/api/reading.md
@@ -174,7 +174,7 @@ GET /people?order=age.desc.nullslast
 
 To order the embedded items, you need to specify the tree path for the order param like so.
 ```HTTP
-GET /projects?select=id,name,tasks{id,name}&order=id.ask&tasks.order=name.ask
+GET /projects?select=id,name,tasks{id,name}&order=id.asc&tasks.order=name.asc
 ```
 
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -21,7 +21,7 @@ import           Network.HTTP.Types.Header (hAuthorization)
 import           Network.HTTP.Types.URI    (parseSimpleQuery)
 import           Network.Wai               (Request (..))
 import           Network.Wai.Parse         (parseHttpAccept)
-import           PostgREST.RangeQuery      (NonnegRange, rangeRequested)
+import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, limitToRange, toRange )
 import           PostgREST.Types           (QualifiedIdentifier (..),
                                             Schema, Payload(..),
                                             UniformObjects(..))
@@ -60,7 +60,7 @@ data ApiRequest = ApiRequest {
   -- | Similar but not identical to HTTP verb, e.g. Create/Invoke both POST
     iAction :: Action
   -- | Requested range of rows within response
-  , iRange  :: NonnegRange
+  , iRange  :: M.HashMap String NonnegRange
   -- | The target, be it calling a proc or accessing a table
   , iTarget :: Target
   -- | The content type the client most desires (or JSON if undecided)
@@ -140,18 +140,20 @@ userApiRequest schema req reqBody =
 
   ApiRequest {
     iAction = action
-  , iRange  = if singular then singletonRange 0 else rangeRequested hdrs
   , iTarget = target
+  , iRange = setTopLevelRange headerRange $
+             setTopLevelRange urlRange $
+             M.fromList [(cs k, limitToRange $ cs $ fromJust v) | (k,v) <- qParams, isJust v, endingIn ["limit"] k ]
   , iAccepts = pickContentType $ lookupHeader "accept"
   , iPayload = relevantPayload
   , iPreferRepresentation = representation
   , iPreferSingular = singular
   , iPreferCount = not $ singular || hasPrefer "count=none"
-  , iFilters = [ (cs k, fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn "order" k) ]
+  , iFilters = [ (cs k, fromJust v) | (k,v) <- qParams, isJust v, k /= "select", k /= "offset", not (endingIn ["order", "limit"] k) ]
   , iSelect = if method == "DELETE"
               then "*"
               else fromMaybe "*" $ fromMaybe (Just "*") $ lookup "select" qParams
-  , iOrder = [(cs k, fromJust v) | (k,v) <- qParams, isJust v, endingIn "order" k ]
+  , iOrder = [(cs k, fromJust v) | (k,v) <- qParams, isJust v, endingIn ["order"] k ]
   , iCanonicalQS = urlEncodeVars
      . sortBy (comparing fst)
      . map (join (***) cs)
@@ -181,9 +183,17 @@ userApiRequest schema req reqBody =
   tokenStr = case T.split (== ' ') (cs auth) of
     ("Bearer" : t : _) -> t
     _                  -> ""
-  endingIn:: T.Text -> T.Text -> Bool
-  endingIn word key = word == lastWord
+  endingIn:: [T.Text] -> T.Text -> Bool
+  endingIn xx key = lastWord `elem` xx
     where lastWord = last $ T.split (=='.') key
+
+  headerRange = if singular then Just (singletonRange 0) else rangeRequested hdrs
+  urlRange = toRange (join $ lookup "limit" qParams) (join $ lookup "offset" qParams)
+
+  setTopLevelRange :: Maybe NonnegRange -> M.HashMap String NonnegRange -> M.HashMap String NonnegRange
+  setTopLevelRange Nothing ranges = ranges
+  setTopLevelRange (Just r) ranges = M.insert "limit" r ranges
+
 
 -- PRIVATE ---------------------------------------------------------------
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -15,17 +15,18 @@ import           Data.Monoid               ((<>))
 import           Data.Ord                  (comparing)
 import           Data.String.Conversions   (cs)
 import qualified Data.Text                 as T
+import           Text.Read                 (readMaybe)
 import qualified Data.Vector               as V
 import           Network.HTTP.Base         (urlEncodeVars)
 import           Network.HTTP.Types.Header (hAuthorization)
 import           Network.HTTP.Types.URI    (parseSimpleQuery)
 import           Network.Wai               (Request (..))
 import           Network.Wai.Parse         (parseHttpAccept)
-import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, limitToRange, toRange )
+import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, restrictRange, rangeGeq, allRange)
 import           PostgREST.Types           (QualifiedIdentifier (..),
                                             Schema, Payload(..),
                                             UniformObjects(..))
-import           Data.Ranged.Ranges        (singletonRange)
+import           Data.Ranged.Ranges        (singletonRange, rangeIntersection)
 
 type RequestBody = BL.ByteString
 
@@ -141,9 +142,8 @@ userApiRequest schema req reqBody =
   ApiRequest {
     iAction = action
   , iTarget = target
-  , iRange = setTopLevelRange headerRange $
-             setTopLevelRange urlRange $
-             M.fromList [(cs k, limitToRange $ cs $ fromJust v) | (k,v) <- qParams, isJust v, endingIn ["limit"] k ]
+  , iRange = M.insert "limit" (rangeIntersection headerRange urlRange) $
+      M.fromList [ (cs k, restrictRange (readMaybe =<< v) allRange) | (k,v) <- qParams, isJust v, endingIn ["limit"] k ]
   , iAccepts = pickContentType $ lookupHeader "accept"
   , iPayload = relevantPayload
   , iPreferRepresentation = representation
@@ -187,13 +187,12 @@ userApiRequest schema req reqBody =
   endingIn xx key = lastWord `elem` xx
     where lastWord = last $ T.split (=='.') key
 
-  headerRange = if singular then Just (singletonRange 0) else rangeRequested hdrs
-  urlRange = toRange (join $ lookup "limit" qParams) (join $ lookup "offset" qParams)
-
-  setTopLevelRange :: Maybe NonnegRange -> M.HashMap String NonnegRange -> M.HashMap String NonnegRange
-  setTopLevelRange Nothing ranges = ranges
-  setTopLevelRange (Just r) ranges = M.insert "limit" r ranges
-
+  headerRange = if singular then singletonRange 0 else rangeRequested hdrs
+  urlOffsetRange = rangeGeq . fromMaybe (0::Integer) $
+    readMaybe =<< join (lookup "offset" qParams)
+  urlRange = restrictRange
+    (readMaybe =<< join (lookup "limit" qParams))
+    urlOffsetRange
 
 -- PRIVATE ---------------------------------------------------------------
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -36,6 +36,8 @@ import           Data.Time.Clock.POSIX     (getPOSIXTime)
 import qualified Data.Vector               as V
 import qualified Hasql.Transaction         as H
 
+import qualified Data.HashMap.Strict       as M
+
 import           PostgREST.ApiRequest   (ApiRequest(..), ContentType(..)
                                             , Action(..), Target(..)
                                             , PreferRepresentation (..)
@@ -101,7 +103,7 @@ app dbStructure conf apiRequest =
         Left e -> return $ responseLBS status400 [jsonH] $ cs e
         Right (q, cq) -> do
           let singular = iPreferSingular apiRequest
-              stm = createReadStatement q cq range singular
+              stm = createReadStatement q cq singular
                     shouldCount (contentType == TextCSV)
           respondToRange $ do
             row <- H.query () stm
@@ -187,7 +189,7 @@ app dbStructure conf apiRequest =
           let p = V.head payload
               jwtSecret = configJwtSecret conf
           respondToRange $ do
-            row <- H.query () (callProc qi p range shouldCount)
+            row <- H.query () (callProc qi p topLevelRange shouldCount)
             returnJWT <- H.query qi doesProcReturnJWT
             let (tableTotal, queryTotal, body) = fromMaybe (Just 0, 0, emptyArray) row
                 (status, contentRange) = rangeHeader queryTotal tableTotal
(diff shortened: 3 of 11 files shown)


begriffs:
Cool man, thanks. I'll take a look at this tomorrow. Also 👍 for cleanup wherever we can.

ruslantalpa:
I also added a few lines to the docs mentioning filtering of embedded items, not detailed at all but at least it will hit that it's possible.

docs/api/reading.md
```diff
@@ -172,7 +172,7 @@ GET /people?order=age.nullsfirst
 GET /people?order=age.desc.nullslast
 ```
 
-To filter the embedded items, you need to specify the tree path for the order param like so. 
+To order the embedded items, you need to specify the tree path for the order param like so.
 ```HTTP
 GET /projects?select=id,name,tasks{id,name}&order=id.asc&tasks.order=name.ask
```
begriffs: Oh looks like the `.ask` was a previous typo, want to include a fix here while you're at it?

test/Feature/RangeSpec.hs
```diff
@@ -142,6 +143,50 @@ spec = do
...
+          , matchHeaders = ["Content-Range" <:> "0-14/15"]
+          }
+      it "top level limit with parameter" $
+        get "/items?select=id&order=id.asc&limit=3"
+          `shouldRespondWith` ResponseMatcher {
+            matchBody    = Just [str|[{"id":1},{"id":2},{"id":3}]|]
+          , matchStatus  = 206
+          , matchHeaders = ["Content-Range" <:> "0-2/15"]
+          }
+      it "headers override get parameters" $
```
begriffs: I was going to ask about what happens in this case. I like that you documented it in a test.

src/PostgREST/ApiRequest.hs
```diff
@@ -181,10 +183,18 @@ userApiRequest schema req reqBody =
...
-  endingIn:: T.Text -> T.Text -> Bool
-  endingIn word key = word == lastWord
+  endingIn:: [T.Text] -> T.Text -> Bool
+  endingIn xx key = lastWord `elem` xx
     where lastWord = last $ T.split (=='.') key
 
+  headerRange = if singular then Just (singletonRange 0) else rangeRequested hdrs
+  urlRange = toRange (join $ lookup "limit" qParams) (join $ lookup "offset" qParams)
+
+  setTopRange :: Maybe NonnegRange -> M.HashMap String NonnegRange -> M.HashMap String NonnegRange
```
begriffs: This name confused me because I thought it was referring to limiting the upper value of a range (the range's "top value") but I guess it refers to ranges for top levels of the query tree you're building?

src/PostgREST/RangeQuery.hs
```diff
@@ -38,12 +41,13 @@ rangeParse range = do
       rangeIntersection lower upper
     Nothing -> rangeGeq 0
 
-rangeRequested :: RequestHeaders -> NonnegRange
-rangeRequested = rangeParse . fromMaybe "" . lookup hRange
+rangeRequested :: RequestHeaders -> Maybe NonnegRange
```
begriffs: Rather than `Maybe NonnegRange` what about using `NonnegRange` with a value of `fullRange` in place of `Nothing`?

src/PostgREST/App.hs
```diff
@@ -286,53 +288,78 @@ augumentRequestWithJoin schema allRels request =
   (first formatRelationError . addRelations schema allRels Nothing) request
   >>= addJoinConditions schema
 
-buildReadRequest :: [Relation] -> ApiRequest -> Either Text ReadRequest
-buildReadRequest allRels apiRequest  =
-  augumentRequestWithJoin schema rels =<<
-  first formatParserError (foldr addFilter <$> (foldr addOrder <$> readRequest <*> ords) <*> flts)
+addFiltersOrdersRanges :: ApiRequest -> Either ParseError (ReadRequest -> ReadRequest)
+addFiltersOrdersRanges apiRequest = foldr1 (liftA2 (.)) [
```
begriffs: Trying to think of a more prosaic way to write this function but not sure how to do it.

src/PostgREST/RangeQuery.hs
```diff
@@ -38,12 +41,13 @@ rangeParse range = do
...
+rangeRequested headers = rangeParse <$> lookup hRange headers
```
begriffs: so something like this
```haskell
fromMaybe (rangeGeq 0) $ rangeParse <$> lookup hRange headers
```

src/PostgREST/RangeQuery.hs
```diff
@@ -66,3 +70,17 @@ rangeGeq n =
 rangeLeq :: Integer -> NonnegRange
 rangeLeq n =
   Range BoundaryBelowAll (BoundaryAbove n)
+
+limitToRange :: BS.ByteString -> NonnegRange
+limitToRange l = rangeParse ("0-" <> cs (show (l' - 1)))
+  where l' = fromMaybe 0 (readMaybe $ cs l)::Integer
+
+toRange :: Maybe String -> Maybe String -> Maybe NonnegRange
+toRange Nothing Nothing = Nothing
```
begriffs: Same idea here, try the ≥0 range

src/PostgREST/RangeQuery.hs
```diff
@@ -38,12 +41,13 @@ rangeParse range = do
       rangeIntersection lower upper
     Nothing -> rangeGeq 0
 
-rangeRequested :: RequestHeaders -> NonnegRange
-rangeRequested = rangeParse . fromMaybe "" . lookup hRange
+rangeRequested :: RequestHeaders -> Maybe NonnegRange
+rangeRequested headers = rangeParse <$> lookup hRange headers
 
-restrictRange :: Maybe Integer -> NonnegRange -> NonnegRange
+restrictRange :: Maybe Integer -> Maybe NonnegRange -> Maybe NonnegRange
```
begriffs: If you go back to manipulating straight NonnegRange then this function could return to how it was?

src/PostgREST/ApiRequest.hs
begriffs: After reading more of the code I see that this `top` name is a common pattern so it makes more sense. Maybe `setTopLevelRange` would be less ambiguous.

docs/api/reading.md
ruslantalpa: done

src/PostgREST/ApiRequest.hs
ruslantalpa: done

src/PostgREST/RangeQuery.hs
ruslantalpa: The Maybe ties into this function https://github.com/begriffs/postgrest/pull/608/files#diff-02f231cf09412413f8f0e9d5e5d7a06eR193
and i use Nothing to "not set" the limit for the top level range if it's not specified, so if there is no value for it, there is actually nothing in the sql, before this, it used to be "Limit all"

src/PostgREST/App.hs
ruslantalpa: Unless there is high level function that can make this shorter i think it would have something similar underneath. I say that because i am thinking like:
We are in the context `Either ParserError a` so when we apply a value in that context to a function taking two parameters using `<$>` we get ourselves things like `Either ParserError (ReadRequest -> ReadRequest)`
With no context, we would use simple compose since their types align nicely, but since they are in a context, we need to compose them in that context ... and the only way to do that is applicative (i think), i'll add a comment to make it a bit more clear what's going on and maybe later if some haskell guru looks over it, might suggest something better. My feeling is that if something will be possible, it will be because of the functions in question are folds and not because of the applicative bit

src/PostgREST/RangeQuery.hs
ruslantalpa: explained above why i deed the Nothing bit returned

src/PostgREST/RangeQuery.hs
ruslantalpa: same as above, i think i need the Nothing value coming out of these functions

ruslantalpa:
So, is this pr good for merging?

src/PostgREST/RangeQuery.hs
begriffs: You could modify that function to also operate without a Maybe:

```haskell
setTopLevelRange :: NonnegRange -> M.HashMap String NonnegRange -> M.HashMap String NonnegRange
setTopLevelRange r ranges
  | r == rangeGeq 0 = ranges
  | otherwise       = M.insert "limit" r ranges
```

ruslantalpa: I don't particularly like this solution, it's like using (return -1) to signal Nothing in C for the sake of making the code a few lines shorter.

begriffs: Actually won't `limitF` use `limit ALL` if `rangeLimit` returns Nothing? You may be able to unconditionally insert the range in `setTopLevelRange`.

ruslantalpa: no, https://github.com/begriffs/postgrest/pull/608/files#diff-9165bde0d08f724b3d81b8a6ff4b784eR292

begriffs:
OK here's a patch that represents ranges internally as ranges rather than Maybe ranges. It nets a twenty-line reduction. My reasoning is that headers/params may or may not include the ranges, but we should interpret input at the edges of the program and inside the program use the meaning of the input. A missing range in the request means "all items" aka the range 0 to infinity.

This passes all your feature specs, but does change one thing. If ranges are supplied in both the query params and headers then this code will take their intersection rather than having one override the other. So a request with `Range: 0-9` and `?limit=5` would return items 0-4. Similarly `Range: 5-24` and `?limit=10` would return 5-9 rather than 5-14 as someone might imagine.

I don't think it would be too hard to change the intersection logic back to an override if this behavior seems too weird.

```patch
From 337c736fc3f1238a52b3eb1b4b71eecc56383759 Mon Sep 17 00:00:00 2001
From: Joe Nelson <cred+github@begriffs.com>
Date: Fri, 27 May 2016 12:27:28 -0700
Subject: [PATCH] Manipulate ranges internally, not Maybe ranges

---
 src/PostgREST/ApiRequest.hs   | 22 ++++++++++------------
 src/PostgREST/App.hs          |  2 +-
 src/PostgREST/Parsers.hs      |  6 +++---
 src/PostgREST/QueryBuilder.hs |  2 +-
 src/PostgREST/RangeQuery.hs   | 27 +++++----------------------
 src/PostgREST/Types.hs        |  2 +-
 6 files changed, 21 insertions(+), 40 deletions(-)

diff --git a/src/PostgREST/ApiRequest.hs b/src/PostgREST/ApiRequest.hs
index acabbf1..be5260e 100644
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -21,11 +21,11 @@ import           Network.HTTP.Types.Header (hAuthorization)
 import           Network.HTTP.Types.URI    (parseSimpleQuery)
 import           Network.Wai               (Request (..))
 import           Network.Wai.Parse         (parseHttpAccept)
-import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, limitToRange, toRange )
+import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, restrictRange, rangeGeq)
 import           PostgREST.Types           (QualifiedIdentifier (..),
                                             Schema, Payload(..),
                                             UniformObjects(..))
-import           Data.Ranged.Ranges        (singletonRange)
+import           Data.Ranged.Ranges        (singletonRange, rangeIntersection)
 
 type RequestBody = BL.ByteString
 
@@ -141,9 +141,8 @@ userApiRequest schema req reqBody =
   ApiRequest {
     iAction = action
   , iTarget = target
-  , iRange = setTopLevelRange headerRange $
-             setTopLevelRange urlRange $
-             M.fromList [(cs k, limitToRange $ cs $ fromJust v) | (k,v) <- qParams, isJust v, endingIn ["limit"] k ]
+  , iRange = M.insert "limit" (rangeIntersection headerRange urlRange) $
+      M.fromList [ (cs k, restrictRange (read <$> v) (rangeGeq 0)) | (k,v) <- qParams, isJust v, endingIn ["limit"] k ]
   , iAccepts = pickContentType $ lookupHeader "accept"
   , iPayload = relevantPayload
@@ -187,13 +186,12 @@ userApiRequest schema req reqBody =
   endingIn xx key = lastWord `elem` xx
 
-  headerRange = if singular then Just (singletonRange 0) else rangeRequested hdrs
-  urlRange = toRange (join $ lookup "limit" qParams) (join $ lookup "offset" qParams)
-
-  setTopLevelRange :: Maybe NonnegRange -> M.HashMap String NonnegRange -> M.HashMap String NonnegRange
-  setTopLevelRange Nothing ranges = ranges
-  setTopLevelRange (Just r) ranges = M.insert "limit" r ranges
-
+  headerRange = if singular then singletonRange 0 else rangeRequested hdrs
+  urlOffsetRange = rangeGeq . fromMaybe (0::Integer) $
+    read <$> join (lookup "offset" qParams)
+  urlRange = restrictRange
+    (fmap read (join $ lookup "limit" qParams))
+    urlOffsetRange
 
 -- PRIVATE ---------------------------------------------------------------
 
diff --git a/src/PostgREST/App.hs b/src/PostgREST/App.hs
index 52ef98a..cc1b685 100644
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -379,7 +379,7 @@ addOrder :: (Path, [OrderTerm]) -> ReadRequest -> ReadRequest
 addOrder = addProperty addOrderToNode
 
 addRangeToNode :: NonnegRange -> ReadRequest -> ReadRequest
-addRangeToNode r (Node (q,i) f) = Node (q{range_=Just r}, i) f
+addRangeToNode r (Node (q,i) f) = Node (q{range_=r}, i) f
 
 addRange :: (Path, NonnegRange) -> ReadRequest -> ReadRequest
 addRange = addProperty addRangeToNode
diff --git a/src/PostgREST/Parsers.hs b/src/PostgREST/Parsers.hs
index bb12fc7..e10dfa9 100644
--- a/src/PostgREST/Parsers.hs
+++ b/src/PostgREST/Parsers.hs
@@ -11,7 +11,7 @@ import           Data.Tree
 import           PostgREST.QueryBuilder        (operators)
 import           PostgREST.Types
 import           Text.ParserCombinators.Parsec hiding (many, (<|>))
-import           PostgREST.RangeQuery      (NonnegRange)
+import           PostgREST.RangeQuery      (NonnegRange, rangeGeq)
 
 
 pRequestSelect :: Text -> Parser ReadRequest
@@ -19,7 +19,7 @@ pRequestSelect rootNodeName = do
   fieldTree <- pFieldForest
   return $ foldr treeEntry (Node (readQuery, (rootNodeName, Nothing, Nothing)) []) fieldTree
   where
-    readQuery = Select [] [rootNodeName] [] Nothing Nothing
+    readQuery = Select [] [rootNodeName] [] Nothing (rangeGeq 0)
     treeEntry :: Tree SelectItem -> ReadRequest -> ReadRequest
     treeEntry (Node fld@((fn, _),_,alias) fldForest) (Node (q, i) rForest) =
       case fldForest of
@@ -27,7 +27,7 @@ pRequestSelect rootNodeName = do
         _  -> Node (q, i) newForest
           where
             newForest =
-              foldr treeEntry (Node (Select [] [fn] [] Nothing Nothing, (fn, Nothing, alias)) []) fldForest:rForest
+              foldr treeEntry (Node (Select [] [fn] [] Nothing (rangeGeq 0), (fn, Nothing, alias)) []) fldForest:rForest
 
 pRequestFilter :: (String, String) -> Either ParseError (Path, Filter)
 pRequestFilter (k, v) = (,) <$> path <*> (Filter <$> fld <*> op <*> val)
diff --git a/src/PostgREST/QueryBuilder.hs b/src/PostgREST/QueryBuilder.hs
index d8664e1..a06bcbb 100644
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -289,7 +289,7 @@ requestToQuery schema (DbRead (Node (Select colSelects tbls conditions ord range
       unwords joins,
       ("WHERE " <> intercalate " AND " ( map (pgFmtCondition qi ) conditions )) `emptyOnNull` conditions,
       orderF (fromMaybe [] ord),
-      fromMaybe "" $ limitF <$> range
+      limitF range
       ]
     orderF ts =
         if null ts
diff --git a/src/PostgREST/RangeQuery.hs b/src/PostgREST/RangeQuery.hs
index 1ef75d6..d5e70eb 100644
--- a/src/PostgREST/RangeQuery.hs
+++ b/src/PostgREST/RangeQuery.hs
@@ -6,14 +6,11 @@ module PostgREST.RangeQuery (
 , restrictRange
 , rangeGeq
 , NonnegRange
-, limitToRange
-, toRange
 ) where
 
 
 import           Control.Applicative
 import           Network.HTTP.Types.Header
-import           Data.Monoid               ((<>))
 
 import           Data.Ranged.Boundaries
@@ -41,13 +38,13 @@ rangeParse range = do
       rangeIntersection lower upper
     Nothing -> rangeGeq 0
 
-rangeRequested :: RequestHeaders -> Maybe NonnegRange
-rangeRequested headers = rangeParse <$> lookup hRange headers
+rangeRequested :: RequestHeaders -> NonnegRange
+rangeRequested headers = fromMaybe (rangeGeq 0) $
+  rangeParse <$> lookup hRange headers
 
-restrictRange :: Maybe Integer -> Maybe NonnegRange -> Maybe NonnegRange
+restrictRange :: Maybe Integer -> NonnegRange -> NonnegRange
 restrictRange Nothing r = r
-restrictRange (Just limit) Nothing = Just $ rangeIntersection (rangeGeq 0) (rangeLeq (limit - 1))
-restrictRange (Just limit) (Just r) = Just $
+restrictRange (Just limit) r =
   rangeIntersection r $
     Range BoundaryBelowAll (BoundaryAbove $ rangeOffset r + limit - 1)
 
@@ -70,17 +67,3 @@ rangeGeq n =
 rangeLeq :: Integer -> NonnegRange
 rangeLeq n =
   Range BoundaryBelowAll (BoundaryAbove n)
-
-limitToRange :: BS.ByteString -> NonnegRange
-limitToRange l = rangeParse ("0-" <> cs (show (l' - 1)))
-  where l' = fromMaybe 0 (readMaybe $ cs l)::Integer
-
-toRange :: Maybe String -> Maybe String -> Maybe NonnegRange
-toRange Nothing Nothing = Nothing
-toRange Nothing (Just o) = Just $ rangeParse $ cs $ show o' <> "-"
-  where o' = fromMaybe 0 (readMaybe $ cs o)::Integer
-toRange (Just l) Nothing = Just $ limitToRange $ cs l
-toRange (Just l) (Just o) = Just $ rangeParse $ cs $ show o' <> "-" <> show (o' + l' - 1)
-  where
-    l' = fromMaybe 0 (readMaybe $ cs l)::Integer
-    o' = fromMaybe 0 (readMaybe $ cs o)::Integer
diff --git a/src/PostgREST/Types.hs b/src/PostgREST/Types.hs
index 3337622..b4a225c 100644
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -112,7 +112,7 @@ type Cast = Text
 type NodeName = Text
 type SelectItem = (Field, Maybe Cast, Maybe Alias)
 type Path = [Text]
-data ReadQuery = Select { select::[SelectItem], from::[TableName], flt_::[Filter], order::Maybe [OrderTerm], range_::Maybe NonnegRange } deriving (Show, Eq)
+data ReadQuery = Select { select::[SelectItem], from::[TableName], flt_::[Filter], order::Maybe [OrderTerm], range_::NonnegRange } deriving (Show, Eq)
 data MutateQuery = Insert { in_::TableName, qPayload::Payload }
                  | Delete { in_::TableName, where_::[Filter] }
                  | Update { in_::TableName, qPayload::Payload, where_::[Filter] } deriving (Show, Eq)
-- 
2.8.2

```

begriffs:
One thing I just noticed: my diff includes the `read` function which is unsafe. It throws an exception if it can't parse the input. Probably `readMay` would be better.

cutso:
hi begriffs, so glad to see this PR released in next version that we can get the result  in one time with a nested query. thx ruslantalpa.

ruslantalpa:
@cutso it's been possible to get the data in one go using &select parameter for a long time now, this pr only introduces the ability to limit the number of returned items on each level.

ruslantalpa:
Done implementing your diff

src/PostgREST/ApiRequest.hs
```diff
@@ -140,18 +141,19 @@ userApiRequest schema req reqBody =
 
   ApiRequest {
     iAction = action
-  , iRange  = if singular then singletonRange 0 else rangeRequested hdrs
   , iTarget = target
+  , iRange = M.insert "limit" (rangeIntersection headerRange urlRange) $
+      M.fromList [ (cs k, restrictRange (read <$> v) allRange) | (k,v) <- qParams, isJust v, endingIn ["limit"] k ]
```
begriffs: I see you imported `readMaybe` above, what about doing `(join $ readMaybe <$> v)` to be on the safe side?

ruslantalpa:
done

test/Feature/RangeSpec.hs
begriffs: Maybe it would be more accurate to say "header and url parameter ranges intersect"

src/PostgREST/RangeQuery.hs
```diff
@@ -63,6 +65,9 @@ rangeGeq :: Integer -> NonnegRange
 rangeGeq n =
   Range (BoundaryBelow n) BoundaryAboveAll
 
+allRange :: NonnegRange
+allRange = rangeGeq 0
```
begriffs: This is nice, I like how how it standardizes the rest of the code to say `allRange`.

src/PostgREST/RangeQuery.hs
```diff
@@ -39,13 +40,14 @@ rangeParse range = do
     Nothing -> rangeGeq 0
 
 rangeRequested :: RequestHeaders -> NonnegRange
-rangeRequested = rangeParse . fromMaybe "" . lookup hRange
+rangeRequested headers = fromMaybe (rangeGeq 0) $
```
begriffs: What about using `allRange` internally in this file as well?

src/PostgREST/RangeQuery.hs
ruslantalpa: done

test/Feature/RangeSpec.hs
steve-chavez: This tripped me up, thought for a long time Range overrode limit/offset.

PR #615, opened 2016-05-28
Add Github issue template

begriffs:
When people open a new issue they will see these questions. Should help automate getting useful information. Any suggestions for improvements? I got it from https://www.talater.com/open-source-templates/#/page/98 and modified it slightly

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,31 @@
+<!--- Provide a general summary of the issue in the Title above -->
+
+## Context
+<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
+
+## Expected Behavior
+<!--- Tell us what should happen -->
+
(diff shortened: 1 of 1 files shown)


ruslantalpa:
Style wise .. the ## seems to make the text too big :)
will the comments < ... > be visible when creating the issue (otherwise the titles don't say much)
not every issue is a "bug", most of them are "questions/features" so if possible, it would be nice to be able to allow ppl to chose a tag and a template for that tag (not sure if possible).
For the bug template (which this PR seems to implement) it would help if steps to reproduce would be like: provide sample schema dump, provide curl request, provide result output, provide desired output ...

As a conclusion, what i find most lacking in the process is the tagging (question/bug/feature) and not that ppl do not provide information.

begriffs:
Yeah you're probably right. I'll close this for now. We can just remember for later that we can add an issue template if need be.

Issue #600, opened 2016-05-19
Why is result returned as a dictionary with the function name as key?


nerfpops:
A sample response from this function:

    CREATE OR REPLACE FUNCTION public.nearest_droppoint(
        longitude gis_longitude,
        latitude gis_latitude)
      RETURNS jsonb AS
    $BODY$
        select to_jsonb(row) from (select *, distance_from_to(droppoint_longitude, droppoint_latitude, longitude, latitude) as distance_km from droppoints order by distance_km limit 1) row
    $BODY$
      LANGUAGE sql STABLE SECURITY DEFINER;

Returns

    [
      {
        "nearest_droppoint": {
          "droppoint_latitude": 55.696037,
          "droppoint_postal_code": "2100",
          "distance_km": 0.0070047270467817,
          "droppoint_street1": "Blegdamsvej 82",
          "droppoint_city": "K\u00f8benhavn \u00d8",
          "courier": "dao",
          "id": 470,
          "droppoint_name": "F\u00e6lledparkens Kiosk",
          "droppoint_external_id": "97487",
          "droppoint_longitude": 12.571179
        }
      }
    ]

Is there any reason to wrap the result in a dictionary by the name of the function?

begriffs:
It's a side effect of the sql output from our `callProc` function. I bet it's possible to modify to get rid of the function name in the results.

We may want to think when to make the change though as it would be backwards incompatible with existing clients. (Like save it for the release after the current one in progress)

begriffs:
One other note, if we remove the function name from the response then in certain situations it will not be valid JSON. For instance suppose `myfunc` returns a number like 2. Then the response would be

```json
[ { "myfunc": 2 } ]
```
Removing the function name from the output would give
```json
[2]
```
and including request header `Prefer: plurality=singular` would give back
```
2
```
which, while a javascript value, is not a valid response for the mime type `application/json`. See [this](https://stackoverflow.com/questions/7487869/is-this-simple-string-considered-valid-json) discussion for more detail.

So it might be a bad idea to implement this proposal. Thoughts?

nerfpops:
The latest JSON spec supports strings, numbers and booleans:

From http://stackoverflow.com/a/33440554/1193236
As of 2014, RFC 7159 obsoletes the older JSON RFCs, and declares that any JSON value is valid JSON text and valid application/json content - including strings. However, it also points out the incompatibility issue with older JSON implementations:

[...]

Issue #599, opened 2016-05-19
Support plurality=singular for functions


nerfpops:
Either that, or just always assume singular if function doesn't return setof T.

Setting prefer: plurality=singular doesn't seem to affect functions.

ruslantalpa:
This should be part of a bit of refactoring for RPC path, besides plurality support, &select parameter should also be supported

nerfpops:
Oh yes, select and filtering on function return values would be a killer feature.

begriffs:
It's all possible now.

PR #620, opened 2016-05-31
Rpc refactor

ruslantalpa:
This PR introduces a breaking change, for functions that return scalar types, the output will be directly that value, it used to be wrapped in `[{functionname:...}]`.
This particular change, coupled with the fact that before running the PRC we know the output type, we can generate different queries (for example binary output). These issues are addressed #600 #599.

The other (more interesting) features are the fact that RPC path now joins the elites :) (GET/POST/PATH/DELETE) and supports filtering and shaping/embedding with &select

Enjoy

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -203,29 +203,41 @@ addJoinConditions schema (Node nn@(query, (n, r, a)) forest) =
     addCond query' con = query'{flt_=con ++ flt_ query'}
 
 type ProcResults = (Maybe Int64, Int64, JSON.Value)
-callProc :: QualifiedIdentifier -> JSON.Object -> NonnegRange -> Bool -> H.Query () (Maybe ProcResults)
-callProc qi params range countTotal =
+callProc :: QualifiedIdentifier -> JSON.Object -> SqlQuery -> SqlQuery -> NonnegRange -> Bool -> Bool -> H.Query () (Maybe ProcResults)
+callProc qi params selectQuery countQuery _ countTotal isSingle =
   unicodeStatement sql HE.unit decodeProc True
   where
     sql = [qc|
-            WITH t AS (select * {_callSql})
+            WITH {sourceCTEName} AS ({_callSql})
             SELECT
-              {_countExpr} as countTotal,
-              pg_catalog.count(1) as countResult,
-              array_to_json(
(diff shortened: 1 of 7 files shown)


test/Feature/QuerySpec.hs
```diff
@@ -454,6 +454,38 @@ spec = do
...
+    context "shaping the response returned by a proc" $ do
+      it "returns a project" $
+        post "/rpc/getproject" [json| { "id": 1} |] `shouldRespondWith`
+          [json|[{"id":1,"name":"Windows 7","client_id":1}]|]
+
+      it "can filter proc results" $
+        post "/rpc/getallprojects?id=gt.1&id=lt.5&select=id" [json| {} |] `shouldRespondWith`
+          [json|[{"id":2},{"id":3},{"id":4}]|]
+
+      it "can limit proc results" $
```
diogob: ❤️  I actually was going to send a PR just for this feature!

src/PostgREST/QueryBuilder.hs
```diff
@@ -203,29 +203,41 @@ addJoinConditions schema (Node nn@(query, (n, r, a)) forest) =
...
             SELECT
-              {_countExpr} as countTotal,
-              pg_catalog.count(1) as countResult,
-              array_to_json(
-                coalesce(array_agg(row_to_json(r)), '\{}')
-              )::character varying
-            FROM (select * from t {limitF range}) r;
+              {countResultF} AS total_result_set,
+              pg_catalog.count(t) AS page_total,
+              case when pg_catalog.count(1) > 1
```
diogob: let's try to rewrite using something like:
```sql
case 
  when pg_catalog.count(1) > 1 then 
    {bodyF}
  else
    coalesce(((array_agg(row_to_json(t)))[1]->{_procName}), {bodyF})
end) as body
```

ruslantalpa:
This is good for merging now

begriffs:
Should we merge it and start working toward releasing v0.4?

ruslantalpa:
I vote yes

begriffs:
Heeeere goes! 🏁

begriffs:
@hudayou now that we're working toward a new major version you could make the swagger response the default for the root route (without a special accept header). Also having the swagger document the available functions/arguments for RPC routes would be amazing as well.

PR #625, opened 2016-06-02
Implement select/return representation for DELETE queries (fix #518)

### LINKED ISSUE #518: Support RETURNING functionality on DELETE queries
I'm working on a client side cache/data store front-end to sit in front of postgrest, and this functionality would save me a lot of trouble in terms of cache invalidation logic.

The prefer: return=representation header with a select request arg (like insertion) would be fine in my opinion.

ruslantalpa:
Something i missed when implementing &select for mutate requests.
Would be good to include in the upcoming release


begriffs:
Beautiful. Glad you double-checked about this functionality.

PR #626, opened 2016-06-02
Update operator documentation

edofic:


ruslantalpa:
Here is the list of all operators available
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/QueryBuilder.hs#L231

as far as i know @> and <@ are not exclusive to arrays so it would be good to mention that in the documentation or not mention arrays at all (only maybe for the example)

edofic:
@ruslantalpa you are right about `@>` and `<@` - can also be used for other types. I updated my PR accordingly. I also added other missing operators and adjusted order to match the source.

begriffs:
Thank you @edofic

PR #628, opened 2016-06-06
Include allow header in options response

### LINKED ISSUE #96: Include http verb information in OPTIONS
Including the `Allow` and `Accept-Patch` headers.

http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.7
http://tools.ietf.org/html/rfc5789#page-7

begriffs:
If the table is updatable it includes all the verbs, otherwise just GET.

Fixes #96

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -183,7 +183,7 @@ app dbStructure conf apiRequest =
               filterCol :: Schema -> TableName -> Column -> Bool
               filterCol sc tb Column{colTable=Table{tableSchema=s, tableName=t}} = s==sc && t==tb
               filterCol _ _ _ =  False
-              acceptH = (hAllow, if tableInsertable table then "GET,POST,PUT,PATCH,DELETE" else "GET") in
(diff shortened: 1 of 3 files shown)


src/PostgREST/App.hs
```diff
@@ -173,16 +173,18 @@ app dbStructure conf apiRequest =
...
+      case mTable of
+        Nothing -> return notFound
+        Just table ->
+          let cols = filter (filterCol tSchema tTable) $ dbColumns dbStructure
+              pkeys = map pkName $ filter (filterPk tSchema tTable) allPrKeys
+              body = encode (TableOptions cols pkeys)
+              filterCol :: Schema -> TableName -> Column -> Bool
+              filterCol sc tb Column{colTable=Table{tableSchema=s, tableName=t}} = s==sc && t==tb
+              filterCol _ _ _ =  False
+              acceptH = (hAllow, if tableInsertable table then "GET,POST,PUT,PATCH,DELETE" else "GET") in
```
ruslantalpa: Why PUT?

Issue #631, opened 2016-06-08
JWT role invalidation


daurnimator:
I'm trying to figure out how to deactivate (but not delete) user accounts. e.g. if an employee leaves.
First step will be adding an 'active' column to my users table. 

Using this, you could check that the user is active when you fetch their user id:

```sql
create function core.current_setting_no_throw(s text) returns text
	language plpgsql as $$
	begin
		return current_setting(s);
	exception
		when undefined_object then return NULL;
	end
	$$;

create function core.postgrest_user() returns uuid
	language plpgsql as $$
	declare
		_id uuid;
	begin
		_id = core.current_setting_no_throw('postgrest.claims.user_id')::uuid;
		if _id is null then
			return null;
		elsif exists(select 1 from core.users where id = _id and active = false) then
			raise invalid_password using message = 'your user is inactive';
		end if;
		return _id;
	end
	$$;
```

However this does nothing to change/remove the `role` that is restored by postgrest from their JWT token.
Postgrest needs a way to jump in and validate a role/JWT before it's used.

djk447:
Just mentioned this in the gitter, but check out https://github.com/begriffs/postgrest/issues/529 has a similar feel. I solved this for now by adding a database role for every user, then granting group membership to certain users. I can share some of the functions etc that I made to make the process of role creation etc manageable if that would be helpful. It just feels a bit awkward because of the whole role/cluster thing. But it does allow you to do individual GRANT/REVOKE for the users, so even if the employee doesn't leave but their permissions change it makes it easier. If your app isn't horribly large the role system should be a perfectly viable way of handling it. I still think that a hook before the SET ROLE statement would be useful for other reasons and make this a little cleaner, but this way definitely works.

daurnimator:
> I solved this for now by adding a database role for every user, then granting group membership to certain users.

For my original design I discussed this in the #postgresql irc room, and they said its a bad architecture from their perspective: something about how the user catalogue has locks around access and is shared between databases.
Not to mention, I think I ran into issues inheriting from a system catalogue; which makes adding extra user attributes error prone.

djk447:
[...]

Issue #529, opened 2016-03-15
Option to Call Function Rather than SET ROLE <user>


davidkuep:
In Auth.hs lines 74 & 75 appear to be the only place that determine how auth works in the entire system, and it only works by setting role to a role as provided by JWT. 

`{-| Receives the name of a role and returns a SET ROLE statement -}
setRole :: Text -> BS.ByteString
setRole r = "set local role " <> cs (pgFmtLit r) <> ";"
`

If you want to do session authentication, or set a role based on a parameter other than the role provided in JWT, it's very hard because of this. While I understand that one can do session authorization etc outside of Postgres/PostgREST, it would be much, much easier to instead provide a command line option that allowed for a function to be called by the authenticator user, taking no inputs and using the JWT claims exposed by PostgREST for instance, you could include a command line argument that is either --role-auth or --function-auth. The user would then be required to create a function in the schema in which PostgREST was running called postgrest_auth(). This would allow the user full control over login straight from Postgres, they could just write a function to check any auth tables, etc using the claims encoded in the JWT. The function would be required to return void and issue an error if there was an error authenticating the user.  

For instance, right now, even if I had a session auth system, it would have to be separate from Postgres etc because the only claim that matters in the JWT is the Role claim, so then I'd have to implement something in Nginx or Redis or something else to handle token revocation etc, it would be easier to, instead of issuing a claim in my JWT called Role, simply issue a claim called sessionid and then look it up and switch into the role listed in the field for that session as long as it wasn't expired. I don't think it would add much overhead either. Another way to do this would be to say that role is the reserved key in the JWT for PostgREST, if you issue a role claim in JWT it will switch to that role, if not, (and if you've used a certain command line argument when deploying), it will instead call the postgrest_auth function after setting the other claims in the JWT which it will expect to have the side effect of switching the session to another role.

I suppose this is more of a feature request, and I'm not sure if this would be more complicated to implement than I think, but if it is as easy as I think it would be a nice feature. While it might not be optimal for eventual speed of session auth etc, it is a starting point for that that could work well and securely for a while for a developer, if they're scaling perhaps they'd switch to a quicker option or something, but the complexity involved with this is orders of magnitude less than other options for session auth, which seems like a relatively common use case.

eric-brechemier:
It could be helpful to have a set of extension points defined in SQL, at the interface between PostgREST and PostgreSQL:

* an optional function to set role based on JWT (this issue)
* an optional function to create the JWT from the jwt_claims
[...]

PR #632, opened 2016-06-08
Add ability to provide jwt_claims verification.

daurnimator:
Related to #631 and #529

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -40,11 +40,16 @@ import qualified Web.JWT                 as JWT
   statement.
 -}
 claimsToSQL :: M.HashMap Text Value -> [BS.ByteString]
-claimsToSQL claims = roleStmts <> varStmts
+claimsToSQL claims = varStmts <> [validateStmt] <> roleStmts
  where
+  validateStmt = "do $$declare begin \
+    \if exists(select 1 from pg_catalog.pg_proc where proname = 'validate_jwt_claims') then \
+      \perform validate_jwt_claims(); \
+    \end if; \
+  \end$$;"
   roleStmts = maybeToList $
     (\r -> "set local role " <> r <> ";") . cs . valueToVariable <$> M.lookup "role" claims
-  varStmts = map setVar $ M.toList (M.delete "role" claims)
+  varStmts = map setVar $ M.toList claims
   setVar (k, val) = "set local " <> cs (pgFmtIdent $ "postgrest.claims." <> k)
                     <> " = " <> cs (valueToVariable val) <> ";"
   valueToVariable = pgFmtLit . unquoted


daurnimator:
To accomplish #529, you could do the `set role` in an `else` (and then it could be up to `validate_jwt_claims` to set the role)

daurnimator:
As the code currently is, it works, but the raised error's message isn't coming through:

```sql
create function validate_jwt_claims() returns void
	language plpgsql as $$
	declare
		_id uuid;
		_active boolean;
	begin
		_id := core.postgrest_user();
		if _id is null then return; end if;

		select active from core.users
			where id = _id
			into _active;
		if _active is null then
			raise invalid_password using message = 'invalid user';
		elsif not _active then
			raise invalid_password using message = 'account not active';
		end if;
	end
	$$;
```

Postgres logs:
```
LOG:  statement: SET client_encoding = 'UTF8';SET client_min_messages TO WARNING;
LOG:  execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ ONLY
LOG:  statement: set local "postgrest.claims.role" = 'anon';set local "postgrest.claims.user_id" = '81a38f2a-2d4d-11e6-99e7-dfa0729831ab';do $$declare begin if exists(select 1 from pg_catalog.pg_proc where proname = 'validate_jwt_claims') then perform validate_jwt_claims(); end if; end$$;set local role 'anon';
ERROR:  account not active
CONTEXT:  SQL statement "SELECT validate_jwt_claims()"
        PL/pgSQL function inline_code_block line 1 at PERFORM
STATEMENT:  set local "postgrest.claims.role" = 'anon';set local "postgrest.claims.user_id" = '81a38f2a-2d4d-11e6-99e7-dfa0729831ab';do $$declare begin if exists(select 1 from pg_catalog.pg_proc where proname = 'validate_jwt_claims') then perform validate_jwt_claims(); end if; end$$;set local role 'anon';
```

HTTP response:
```
HTTP/1.1 503 Service Unavailable
Transfer-Encoding: chunked
Date: Wed, 08 Jun 2016 07:57:35 GMT
Server: postgrest/0.3.1.1
Content-Encoding: gzip
Access-Control-Allow-Origin: http://localhost:8080
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: Content-Encoding, Content-Location, Content-Range, Content-Type, Date, Location, Server, Transfer-Encoding, Range-Unit

{"details":"another command is already in progress\n","message":"Database client error"}
```i

**Comment by diogob** [comment]
This counters the purpose of using JWT, relying on external authentication system and assuming that the token comes with valid credentials.
You can setup a very short expiration and require the user to request the token again before each request if the concern is solving #631

**Comment by djk447** [comment]
@diogob I definitely get your position on this, and I think there are some nice ways of implementing what #631 and others want without this feature...but I think the point is that it does counter the purpose of JWT because some people have use cases that JWT isn't as nicely suited for and in which the older, not as useful in some cases, session type auth is better for them, this just allows for that to be done significantly more easily for people with DB experience and potentially with significantly less complexity. 

There are also other use cases that don't involve JWT re-auth that would be helped immensely by this sort of thing. Suppose I wanted to log users transactions etc but they all fall under the same role, so when I look at logs, I basically see that that role is running a query but not the specific user, one could imagine a function that just logged a line to an activity table with other postgrest claims information and the current time and then set role. It's better design practice to have a single role for classes of users and not create them all in the database, but I might still want to record when a user was active or something. 

Another, perhaps more interesting/important use case would be to set session variables for particular users having a hook on SET LOCAL ROLE would also allow you to set other session variables depending on the user. You can do some of that with other claims in the JWT, but not all of it. 

This could also be fixed if  if Postgres provided the ability to have a trigger fire when a SET LOCAL ROLE statement was encountered...but I can't quite figure out how to do that, anyone know if there's a way to do that? The event triggers seem to only be for DDL and normal triggers are only on relations so not quite sure if it's possible.

**Comment by ruslantalpa** [comment]
@djk447 if what you need is sessions then you should generate session ids (with some other system), store for that session the associated jwt token (without sending it to the user) and on each request, replace the session id header with the jwt header and send it upstream to postgrest.
When you need to logout the user, you delete the session, the jwt is lost (even if it was still valid) and the user can no longer access the service.

So far i have not seen an example of usecase that is impossible to implement using openresty+postgrest. Most of the examples provided revolve around "i want to revoke a jwt (aka delete session) and i want postgrest to do it for me automatically" :) or in other words i want jwt's to act as sessions. 

From my point of view PostgREST is a single tool in a toolbox that should do only one thing and do it well  and not try to implement internally things that are better handle a layer above (nginx) or a layer below (db). 
As for this custom function hook, it might be useful in some situations (that i have not heard of yet, they might exist) but using it as a feature for this particular case (use jwt as sessions) is not the right way to do it (session -> jwt switching should be implemented in nginx)

**Comment by daurnimator** [comment]
> You can setup a very short expiration and require the user to request the token again before each request if the concern is solving #631 
> Why is the expiration approach not sufficient?

First off: that's terribly inefficient: it's using two different db connections for every transaction.
Second: it opens up a TOCTOU race.



> I think the point is that it does counter the purpose of JWT because some people have use cases that JWT isn't as nicely suited for and in which the older, not as useful in some cases, session type auth is better for them,

That wasn't what I was going for here.

The issue I'm concerned about is de-privileging a user.
e.g. imagine I fire an employee from my company. I'd like to immediately change their role to a non-privileged user without waiting for expiration.



Furthermore, this allows me to update jwt validation rules on database changes. e.g. if I add a new field to the jwt claim, I can check for it and raise an error that asks the user to re-login.



> So far i have not seen an example of usecase that is impossible to implement using openresty+postgrest. Most of the examples provided revolve around "i want to revoke a jwt (aka delete session) and i want postgrest to do it for me automatically" :) or in other words i want jwt's to act as sessions.
> 
> From my point of view PostgREST is a single tool in a toolbox that should do only one thing and do it well and not try to implement internally things that are better handle a layer above (nginx) or a layer below (db).

I reject the idea that postgrest cannot stand-alone as a database communication middle-man.

**Comment by ruslantalpa** [comment]
@daurnimator here is the confusion that is made (me including). Jwt is supposed to answer the question "who this user is" and not "what rights does he have", and having "role=admin" or "role=employee" in the jwt is trying to encode the ACL into the jwt and that is wrong.
the correct way to do it is have the jwt {role: "webuser", user_id: 1} ... this will never change. Then on the view level (or RLS) have checks like this https://github.com/ruslantalpa/postgrest-rls-example/blob/master/sql/api_schema.sql#L9-L10 but change this function https://github.com/ruslantalpa/postgrest-rls-example/blob/master/sql/functions.sql#L3 from returning the current_user to making a select and get the user type from the db (using user_id comming from jwt)

**Comment by ruslantalpa** [comment]
Thought about this a bit more. so the classic usecase in postgrest is to have 'role' in the jwt.
That works as long as the
a) role=user
b) role=group if you can live with not being able to immediately revoke token or have a blacklist/jwt gate in front of postgrest (nginx)

If the desire is however to use jwt as a session, then it can not contain any "rights", it should only identify the user (jwt containing only user_id field). 
to work with this, one approach (besides middleware in nginx) is to have "smart views" but that can quickly become a pain.
It would be nice to have a jwt that only contains the user_id, the views not to have any logic except for the rls if needed and also have the ability to use "roles as groups".

here is a scenario that i think can solve this problem using a function like this pr is trying to implement

- user starts postgrest with parameter `--role-switching-function-name switch_role` (takes no parameters)
- a request comes in with jwt {user_id:10, some_other_claim: "xxx"}
- postgrest sets all the claims
- it sees there is no 'role' claim so it immediately switches to 'anon' role
- since it was started with that parameter, we make a call to `switch_role()` while being anon user (notice no case statement, no nothing, just call that function)
- the request goes on as normal

About the `switch_role` function: while we would not make any checks, the only thing this function should do is selects (select role from users where id = claims.user_id) and as the last step 'set role to admin/employee ...'
It should never under any circumstances raise an error.
If this function desires to 'deny access' then it should jsut switch to a (existing) role that has no privileges at all so when the main query runs, it gets 'access denied'

**Comment by diogob** [comment]
@ruslantalpa @daurnimator why not trust the JWT to pick the role switching function instead of a parameter?

Let's say I encode in the JWT: ```{switch_role: switch_role, user_id: 10}```
In this case ```switch_role``` could be a special claim (like role is) that points to any function without parameters that returns a `name` or `text` with the role to be set. If the function returns NULL then we keep the anonymous role. If the function is not found we return an error code (I'm not sure which).

**Comment by daurnimator** [comment]
@ruslantalpa sounds like you're understanding a bit better now :)

> user starts postgrest with parameter --role-switching-function-name switch_role (takes no parameters)

I'm happy for this to be configurable, but considering that the `jwt_claims` type name is not configurable, I saw this as fitting into the same catergory of 'special' names.

> it sees there is no 'role' claim so it immediately switches to 'anon' role
> since it was started with that parameter, we make a call to switch_role() while being anon user (notice no case statement, no nothing, just call that function)

I think the function should run as `authenticator`. That way it is just like the 'inbuilt' automatic role switching.

> About the switch_role function: while we would not make any checks, the only thing this function should do is selects (select role from users where id = claims.user_id) and as the last step 'set role to admin/employee ...'

I don't see why it shouldn't be able to do other things. e.g. maybe you want to update the 'last seen' column of a user.

> It should never under any circumstances raise an error.

Why not? I think raising different errors like "unknown user" vs "invalid credential" vs "session conflict", etc is useful. The error should be converted to a http status code where possible.

**Comment by ruslantalpa** [comment]
@diogob good idea only one thing, the function should not return, it should switch the role itselef, simplifies the generated sql and it can't generate errors or the connection will be "poisoned".
@daurnimator i understood from the start, all of this can be done outside postgrest.
whatever role it runs with, that role will need additional access to for example users table (to do the select) ... you might be right.
You can do other things, it's just for that you will need more "rights" for the authenticator (less secure)  and there is a greater possibility to trigger errors, and you can't generate errors in this function.

Generating errors - Because then we'd have to add 'catch' logic in runWithClaims function and that seems too much for this niche feature

**Comment by begriffs** [comment]
* We already [translate](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Error.hs#L73-L97) error codes to http codes so we shouldn't need new logic in `runWithClaims`. The validator function would just choose a postgres code from [the list](https://www.postgresql.org/docs/9.5/static/errcodes-appendix.html) to raise. This assumes we can find a way to get the exception message to bubble all the way up to the user rather than showing "another command is already in progress." (Maybe that is the new logic mentioned)
* Allowing the JWT to include a claim about which validator function to use could be nice, since different users could then be able to undergo different checks, or logging or whatever.
* The authenticator role may not need extra permissions if the validator function is created with `SECURITY DEFINER`.
* I kind of like the idea that if JWT specifies custom validator function then *everything* about the login is up to that function, including executing the `set role` statement. A JWT without a `validator` claim falls back to the current behavior which is the "default validator." It makes the responsibility of the function clear.

File: src/PostgREST/Auth.hs [hunk 1]
```diff
@@ -40,11 +40,16 @@ import qualified Web.JWT                 as JWT
...
 -}
 claimsToSQL :: M.HashMap Text Value -> [BS.ByteString]
-claimsToSQL claims = roleStmts <> varStmts
+claimsToSQL claims = varStmts <> [validateStmt] <> roleStmts
  where
+  validateStmt = "do $$declare begin \
+    \if exists(select 1 from pg_catalog.pg_proc where proname = 'validate_jwt_claims') then \
+      \perform validate_jwt_claims(); \
+    \end if; \
+  \end$$;"
```
begriffs: There's a nice quasi-quoter you can use to prettify multiline strings

```haskell
import Text.InterpolatedString.Perl6 (q)

validateStmt = [q|
    do $$declare begin
      if exists(select 1 ...etc
  |]

**Comment by ruslantalpa** [comment]
@begriffs @diogob 
let's not call it `validator_function`, let's call it `role_switching_function` makes things more clear.
i don't like the function name being included in jwt. you really don't need that "flexibility", it's always going to be one function (and if you really really need to, include a custom claim, and based on that, switch inside the function).
The function name should be a command line parameter. 
Having it like that will allow in future to support `session` concept in the classical way.

To summarise, it should be really simple
command line parameter `role-switching-function`
if jwt comes in, and has `role` in it ... use that and do a `set role`
if it does not, call that function and trust that it will switch to the appropriate role

if you want to support that function throwing errors, you will need a few lines of logic inside `runWithClaims` to detect that an error occured and not call the `continuation` and just return the error

**Comment by diogob** [comment]
@ruslantalpa I don't mind having the parameter instead of the JWT defined. My vote is for whatever is simpler to implement. But regarding the error handling it seems important to me, because if the exception raising was not intentional would be a great help to see an error status other than 500 and a message like: 
> Exception during role switching function "name_of_the_function".

**Comment by begriffs** [comment]
@daurnimator want to give this a try?

**Comment by daurnimator** [comment]
I'm not sure how to wrap the exception in haskell/your libraries.
I think it's best you take this from here :)

**Comment by diogob** [comment]
@begriffs @daurnimator I can take a stab at it this week

**Comment by begriffs** [comment]
@diogob thank you!

**Comment by ruslantalpa** [comment]
@diogob i have this working locally if you are willing to wait a few days for the cleanup.

**Comment by diogob** [comment]
@ruslantalpa sure, I actually don't need this feature, but I thought about a cleaner design:
I was going to revert the changes in `claimsToSQL` and just call the function passed as parameter after the claims had been set.
This would have the benefit of not changing anything about our `JWT` semantics and we just need to add  the procedure call and error handling code to `runWithClaims`.

**Comment by ruslantalpa** [comment]
@diogob 
it's not that simple
you don't want to switch role before calling the function and you don't call the function if the jwt contains role
i've implemented it like this
```
if jwt & role then
  set role from jwt
else if fn then
  call fn()
else
  set role anon
end
```

**Comment by ruslantalpa** [comment]
per discussion with @diogob 
it seems you can do multiple `set role`
in this case the code becomes simpler since you can just the function after the role was switched and it can execute other set role statements.

I'll add a PR once #649 #620 are merged since my local repo has those already merged.

**Comment by ruslantalpa** [comment]
Ok, so this feature is ready with tests. I just need the other PRs to be merged before i can submit a PR for this because i worked on my master on a few things at a time and it has multiple things there (including RPC).
Need to rebase, fix conflicts, then the only thing that will remain will be this feature.

**Comment by begriffs** [comment]
There might be an issue that `authenticator` has privileges to switch into both `A` and `B` but `A` doesn't have access to switch into `B`. So `set role A; set role B;` would fail even though `set role B;` would pass. Not sure about it, want to check it in your tests?

**Comment by ruslantalpa** [comment]
@begriffs. That is not true, it will allways succeed regardless of a/b privileges

**Comment by daurnimator** [comment]
> @begriffs. That is not true, it will allways succeed regardless of a/b privileges

huh? how is that possible?

**Comment by ruslantalpa** [comment]
i was surprised to
pg has the notion of current user and session user
set local role ... changes the current user
and to determine if it has the right to switch, it looks at the privileges of the session user.
that's why you can randomly jump between users

**Comment by daurnimator** [comment]
I don't suppose you accidentally had `INHERIT` on?

**Comment by ruslantalpa** [comment]
no, on what? if i had, it would be on authenticator role and that would should have no effect on switching from A to B or B to A

**Comment by ruslantalpa** [comment]
just try it in your db ...

**Comment by ruslantalpa** [comment]
from the docs https://www.postgresql.org/docs/current/static/sql-set-role.html
SET ROLE has effects comparable to SET SESSION AUTHORIZATION, but the privilege checks involved are quite different. Also, SET SESSION AUTHORIZATION determines which roles are allowable for later SET ROLE commands, **whereas changing roles with SET ROLE does not change the set of roles allowed to a later SET ROLE**.

**Comment by ruslantalpa** [comment]
So i've hit a snag :)
The main use for this feature (for me) is to be able to implement a custom "role switching" function.
for example it would look up a claim SESSID, based on that, look up the session row, user id, get that users 'role' and run a 'set role ...' command.

The problem is that if you want that function to have access to sessions table, you make it a security definer function. The minute you do that ... you are no longer allowed to execute `set role` inside that function.

how can we reconcile this problem?
if we don't make it a 'security definer' function, it means it runs with `anonymous` privileges.
somehow `anonymous` would need full access to sessions table but at the same time no one should be able to access that endpoint from the api.

i know i can give access to `anonymous` to `data.sessions` table (only the `api` schema is exposed ) and since we qualify everything http would not have access but it does not feel right go grant that access

**Comment by djk447** [comment]
So, in this case, I'd use a security definer function just to access the sessions table in the way you want it accessed, so your final function would look something like (and I haven't debugged this so don't trust it completely, but it's the general idea:
(also I'm assuming you mean you'd grant access to authenticator, and anonymous is the default role you switch into otherwise)
```SQL
create function get_user_role(sess_id int) returns text as $$
declare
l_user_id int;
begin
select user from data.sessions where session_id=sess_id into l_user_id;
return select role from data.users where user_id = l_user_id;
end; $$ language plpgsql  volatile security definer;

create function role_switcher() returns void as $$
declare
    user_role text;
begin
    select coalesce(get_user_role( current_setting('postgrest.claims.sessid'), 'anonymous') into user_role;
    execute format('set role %I', user_role);
end; $$ language plpgsql volatile;

grant execute on function get_user_role(int) to authenticator, anonymous;
--you're given a session id and it's basically assumed that if you have a session id you can select stuff about that session, so make it a uuid or something maybe, not an int like I have here...
grant execute on function role_switcher() to authenticator; 
--only authenticator needs to have access to this function, and also if you were to try to access otherwise it would fail because you haven't granted the role you're trying to switch into to other users
```

So that's how I'd do that. I wouldn't give access on the sessions table to anyone outside of the functions used to access it and maybe reset it. You could even grant anonymous access to a rpc that could be run without any jwt to establish a new session with user/pass payload or something. The only access to that table likely should be through functions, unless you're an admin user, this makes sure that you're only accessing the rows you're allowed to access without too much craziness with either rls or security barrier views.

djk447:
Also you could make that function then update something about sessions, like update last_used (would have to mark the function volatile) but then you wouldn't have to worry about update grants on the table, just pack all that logic in the one function you call here.

ruslantalpa:
@djk447 right, should work 👍  thanks

begriffs:
@ruslantalpa did that technique end up working for you?

ruslantalpa:
Yes, it works nicely

begriffs:
@ruslantalpa now that the RPC pull request is merged are you just waiting on #649?

ruslantalpa:
yes, i have a bunch of changes fixes on a private repo so it would be easier for me to to first rebase the master the push a single commit containing everything.
Maybe you should just merge #649 and add the changelog yourself, and i'll try to rebase and PR today/tomorrow

begriffs:
OK, sounds good.

begriffs:
Random other thought: we'll want to prevent people from randomly calling `validate_jwt_claims` via `/rpc`!

ruslantalpa:
thought of that
the function is placed in public or somewhere in the searchpath and we call
it like `thefunction()` and let the searchpath do the work, it's never in
the `api` schema

begriffs:
Since Ruslan has this feature working and tested on another branch I'm closing this PR.

daurnimator:
Which branch is that?

Sigura:
validate_jwt_claims is nice way to add expiration feature manually,

how faster this changes will added to release?

begriffs:
Opening this issue again since a fix hasn't yet made it into master.

Sigura:
Yeah got error: 

"Database client errordb" & "current transaction is aborted, commands ignored until end of transaction block"

on `raise invalid_password using message = '...';`

in other request

begriffs:
Thanks everyone for the discussion. We're adding a command line argument to postgrest which will specify what (if any) stored procedure to run for each request. This proc can raise an error or set a different role or whatever it wants.

daurnimator:
Looks like the fix is still on an unmerged branch? (i.e. "jwt")

begriffs:
Yep, soon to be merged.

Issue #627, opened 2016-06-03
Prefer: return=representation AND Prefer: plurality=singular


nerfpops:
When patching, I'd like the patched row returned as an object instead of an object in a list.

Using fetch in JavaScript:

    var headers = new Headers()
    headers.append("prefer", "return=representation")
    headers.append("prefer", "plurality=singular")
    fetch("/api/row?id=eq.123", { headers: headers, method: "PATCH", body: {} })

Prefer is then merged into a single header separated by a comma:

    prefer:return=representation,plurality=singular

But this results in 204 No Content.

ruslantalpa:
That's because the value of the header should be

`Prefer: return=representation;plurality=singular`

nerfpops:
Now it returns the row in a list.

ruslantalpa:
Looks like this line https://github.com/begriffs/postgrest/blob/master/src/PostgREST/App.hs#L148
hardcodes the fact that when doing update, you always get a list ... that is probably done because we never know how many rows will be updated.
More thought should be put into the idea that we must consider "singular" header ... i.e. you could issue a PATHC, send singular header but actually update a lot more rows

nerfpops:
Personally I don't mind if the separator is comma or semi-colon, but accept-encoding and accept-language uses comma as their seperator and the [fetch spec](https://fetch.spec.whatwg.org/#concept-header-list-append) states that additional values for the same headers should be separated by a comma.

Since a SELECT call might also return additional rows, and plurality=singular in that case just returns the first, I don't think it would be big issue allowing plurality=singular on a patch call that might update multiple rows.

PR #634, opened 2016-06-09
Singular response for PATCH request

ehamberg:
Implement support for `plurality=singular` for `PATCH` requests

This is an attempt at implementing support for `Prefer: return=representation;plurality=singular` as requested in #627. In cases when you update a single entity (i.e. `pkey=eq.1`) this would be convenient since e.g. inserting a new blog post and updating one would both return a post that can be merged with the existing data.


ehamberg:
I added two tests, where one asks for a singular result even after updating two rows. It expects the “first” one as the response – which in practice will work, but either should really be an acceptable response, really.

I'm open to discussing whether this is something we want or not, but personally I have encountered several instances where I would like the response from `POST` and `PATCH` (against a unique id) to have the same shape.

What's still missing from this is handling the case where no rows are updated. This should return a `404` (like it does without `plurality=singular`), but this test:

```haskell
      it "returns 404 when not found" $ do
        request methodPatch
          "/addresses?id=eq.999"
          [("Prefer", "return=representation;plurality=singular")]
          [json| { address: "B Street" } |]
        `shouldRespondWith` 404
```

returns a 500 with `{"details":"Row number 0","message":"Row error: unexpected null"}`. Any suggestions for what needs to be done/how to proceed to debug this?

ruslantalpa:
The code looks good but now is the time to decide how Create/Update should behave in relation to prefer=singular

Create/POST path at this moment ignores the header and decides on the output format based on the number of rows you are inserting, while this PR has the Update/PATCH path checking the prefer header for that case.

I think both cases should behave consistently in relation to prefer header which is a breaking change (probably can be merged only when switching to 4.0.0)
So let's take some time to think about this and the implications both for inserting and updating mainly the question is:

Is it ok (when having prefer=singular) to return the first row even if multiple rows were inserted / updated?

ehamberg:
> Is it ok (when having prefer=singular) to return the first row even if multiple rows were inserted / updated?

Yeah, that's really the core issue. I'm leaning towards yes, but I would consider, e.g. touching several rows with a `PATCH` with `plurality=singular` as a client bug in most cases.

I guess returning an error and rolling back is a bit harsh when you said you *prefer* a singular response. :)

ruslantalpa:
I don't think there is a valid case where you would want to insert/update multiple rows while returning only one. I would speculate that such (runtime) cases are mistakes (forgot to specify the PK condition), thus it would be a good solution to return an error and rollback the transaction (not sure how to exactly do that)

ruslantalpa:
i.e. what is a situation that you would actually want to update multiple rows but get back only the first one?

ehamberg:
I agree. I don't see why one would want to update multiple rows and get only the first row (there isn't really an order, so it's really just *a* row).

An error would be nice in my opinion, but I'm not sure what the error should be, nor whether the semantics of the `Prefer` header allow turning an otherwise acceptable request into one that returns an error.

ruslantalpa:
the error would be bad request (wording is not that important, something like "you indicated using prefer header that you expect your request to update one row but it was about to change x rows so we fixed that by reverting :) ") 
as for semantics ... i don't see a problem here, it's not like we are breaking an RFC or something

The only thing is i do not know how exactly to to accomplish the above :)

diogob:
My 2 cents: I prefer having the header (instead of guessing) and in case there is a mismatch between header and response we return an error.

ehamberg:
Had another (brief) look at this now, but I'm not sure what the best way to do this is. In `App.app` for the `ActionUpdate` case, it is too late to detect a mismatch between rows returned and *plurality* since `QueryBuilder.createWriteStatement` has already created a query that returns only one row.

ruslantalpa:
yeah, that's the problem with generating the error. It needs to happen in sql something like:
once the CTE is ready with the updates, the part where the count is made has to have something like
`if pg_catalog.count(1) > 1 then raise error else return count = 1`

ruslantalpa:
this might help a bit http://stackoverflow.com/questions/11916838/how-to-execute-postgresql-raise-command-dynamically

begriffs:
Actually the `app` function is running inside the `Transaction` monad from the [hasql-transaction](https://hackage.haskell.org/package/hasql-transaction) package. I found that if you run this in the `ActionUpdate` handler it causes the whole thing to be rolled back:

```haskell
H.sql "DO $$ BEGIN RAISE EXCEPTION cardinality_violation USING MESSAGE = 'foo bar'; END $$;"
```

You could try running it conditionally depending on the value of `queryTotal`.

ruslantalpa:
@begriffs YES! :) raise an error after the main query since it's still in transaction ... doh! 👍

ehamberg:
> Actually the app function is running inside the Transaction monad from the hasql-transaction package. I found that if you run this in the ActionUpdate handler it causes the whole thing to be rolled back:

Yep. That was the plan.

> You could try running it conditionally depending on the value of queryTotal.

Unless I'm misunderstanding something – it's too late to do this after the query is run. The query generated by `createWriteStatement ` when `isSingle` is `True` will only return one row, so `queryTotal` is never > 1, so something like this – which was my original plan – won't work:

```haskell
let (_, queryTotal, _, body) = extractQueryResult row
when (singular && queryTotal /= 1) raiseSomeErrorHere
```

Unfortunately I haven't had, and won't have, much time to look at this this week, but I think `createWriteStatement` needs to be changed to achieve this.

ruslantalpa:
@ehamberg that's he beauty of transactions.
although the main query was already executed, things where changed, you got a total back, you are still in transaction context, it's not yet closed, so if you generate an error, everything will be rolled back.

ehamberg:
Sorry, I wasn't being clear. I don't mean it's too late as in “the change has been committed”, I know the whole transaction will be rolled back if I throw an error in the `ActionUpdate` handler in `app`. :)

What I meant was that the query that was generated by `createWriteStatement` when `isSingle` is `True` already has returned a singular result after the `H.query` call. In other words, `queryTotal` will be `1`, even if several records were updated.

ruslantalpa:
change `(_, queryTotal, _ , body)` to `(tableTotal, queryTotal, _ , body)` and work with tableTotal
(Have to look a bit more at the code to make sure that is being returned correctly but it should work)

begriffs:
@ehamberg does Ruslan's suggestion make sense to you?

begriffs:
@ehamberg I hope you're still interested in looking into this, it would be a valuable feature.

ehamberg:
Hi. Sorry about the radio silence. I'm still interested in this feature. :)

I'm currently travelling and won't be able to look at this until the last week of July, however.

I should have commented on Ruslan's suggestion earlier as I did try it. `tableTotal` also returned a count of one, as far as I can remember.

ruslantalpa:
I looked a bit at the code and you should ignore the `tableTotal` and look at the `queryTotal `, if you updated more rows, you should get something bigger then 1, that number is not the thing that is returned but the number of updated rows.

Can you check again and say for sure that you update multiple rows but in that variable you get 1?

ehamberg:
You were right, @ruslantalpa, `queryTotal` has the correct value. What confused me is that this didn't happen in the test I wrote. When I create the same table (`addresses` from `text/fixtures/schema.sql`) in an actual database and insert the same data (from `test/fixtures/data.sql`), I *do* get the 400 response I wanted, but in the test (ddfe1be) I get a 200 response with the first updated entry back:

```
  test/Feature/InsertSpec.hs:414:
  1) Feature.InsertSpec, Patching record, in a table, raises an error when attempting to update multiple entities with plurality=singular
       expected: Status {statusCode = 400, statusMessage = "Bad Request"}
        but got: Status {statusCode = 200, statusMessage = "OK"}
```

Any idea what's happening here? `trace`ing the execution reveals that `singular` is `True` and `queryTotal` is `1` when running the test case…

ruslantalpa:
Just above the test, you deleted the two lines that where inserting the two records you are testing against. Basically you are doing an update on a empty table.

ehamberg:
Don't think that's it as the table should still have the four original rows inserted in `test/fixtures/data.sql` (line 275–278). The body returned in the failing test case is `{"id":1, "address": "zzz"}` which seems to me to confirm this.

Any way, I undid the deletion of the explicitly inserted lines now, so the test doesn't depend on the behaviour of other tests. I still get the same issue. Any idea about why the test fails but doing it manually returns a `400`?

ruslantalpa:
I think the queries from the entire spec are executed within the same transaction + i think there is some laziness into play (since you are using applicative with`when`) so the error is raised a bit later, after you got back the 200 response.
How about trying a quick thing (not final), turn that `when` into an `if` and make two big branches and see how that works.

ehamberg:
Good suggestion.

I tried that, but a 200 was still returned, and the test failed.

I then removed all tests but the failing one and watched the debug output from postgresql to see what's different between running the test and doing the `PATCH` request manually. Here are the relevant log sections:

Running the test:
```
LOG:  execute 6: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
LOG:  statement: set local role 'postgrest_test_anonymous';
DEBUG:  parse 9:
	      WITH pg_source AS (UPDATE  "test"."addresses"  SET "address"='zzz'::unknown  WHERE  "test"."addresses"."id" > '0'::unknown  RETURNING "test"."addresses".*)
	      SELECT '' AS total_result_set, pg_catalog.count(t) AS page_total, ( WITH s AS (SELECT row_to_json(ss) as r from pg_source as ss  limit 1) SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null')) FROM s, json_each_text(s.r) AS json_data), coalesce(string_agg(row_to_json(t)::text, ','), '')::character varying  AS body
	      FROM (SELECT  "pg_source".* FROM  "pg_source"    LIMIT 1 OFFSET 0) t
DEBUG:  bind <unnamed> to 9
LOG:  execute 9:
	      WITH pg_source AS (UPDATE  "test"."addresses"  SET "address"='zzz'::unknown  WHERE  "test"."addresses"."id" > '0'::unknown  RETURNING "test"."addresses".*)
	      SELECT '' AS total_result_set, pg_catalog.count(t) AS page_total, ( WITH s AS (SELECT row_to_json(ss) as r from pg_source as ss  limit 1) SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null')) FROM s, json_each_text(s.r) AS json_data), coalesce(string_agg(row_to_json(t)::text, ','), '')::character varying  AS body
	      FROM (SELECT  "pg_source".* FROM  "pg_source"    LIMIT 1 OFFSET 0) t
DETAIL:  parameters: $1 = '[{"address":"zzz"}]'
```

Doing a HTTP request from the command line:
```
LOG:  execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
LOG:  statement: set local role 'postgrest_test_anonymous';
DEBUG:  parse 1:
	      WITH pg_source AS (UPDATE  "test"."addresses"  SET "address"='zzz'::unknown  WHERE  "test"."addresses"."id" > '0'::unknown  RETURNING "test"."addresses".*)
	      SELECT '' AS total_result_set, pg_catalog.count(t) AS page_total, ( WITH s AS (SELECT row_to_json(ss) as r from pg_source as ss  limit 1) SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null')) FROM s, json_each_text(s.r) AS json_data), ''
	      FROM (SELECT 1 FROM pg_source) t
DEBUG:  bind <unnamed> to 1
LOG:  execute 1:
	      WITH pg_source AS (UPDATE  "test"."addresses"  SET "address"='zzz'::unknown  WHERE  "test"."addresses"."id" > '0'::unknown  RETURNING "test"."addresses".*)
	      SELECT '' AS total_result_set, pg_catalog.count(t) AS page_total, ( WITH s AS (SELECT row_to_json(ss) as r from pg_source as ss  limit 1) SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null')) FROM s, json_each_text(s.r) AS json_data), ''
	      FROM (SELECT 1 FROM pg_source) t
DETAIL:  parameters: $1 = '[{"address":"zzz"}]'
LOG:  statement:  DO $$
	                          BEGIN RAISE EXCEPTION cardinality_violation
	                          USING MESSAGE =
	                            'plurality=singular specified, but more than one object would be updated';
	                          END $$;

ERROR:  plurality=singular specified, but more than one object would be updated
STATEMENT:   DO $$
	                          BEGIN RAISE EXCEPTION cardinality_violation
	                          USING MESSAGE =
	                            'plurality=singular specified, but more than one object would be updated';
	                          END $$;
```

It's hard to see here, but if you open them with a diff viewer, it turns out the only real difference is the following line from running the test:

```
	      FROM (SELECT  "pg_source".* FROM  "pg_source"    LIMIT 1 OFFSET 0) t
```

vs. the same line when doing the HTTP request manually:

```
	      FROM (SELECT 1 FROM pg_source) t
```

Digging further reveals that the `LIMIT` + `OFFSET` comes from `limitF` in `QueryBuilder`:

```haskell
limitF :: NonnegRange -> SqlFragment
limitF r  = if r == allRange
  then ""
  else "LIMIT " <> limit <> " OFFSET " <> offset
  where
    limit  = maybe "ALL" (cs . show) $ rangeLimit r
    offset = cs . show $ rangeOffset r
```

As a quick test, I tried to hard-code `limitF` to return `""`. This “worked” and the test succeeds.

I don't have any more time to debug right now. The open questions at this point:

- Is `limitF` the function that is used when specifying a range in a request?
- Are there range headers specified for the tests?
- Why doesn't setting `Range:  0-0` and `Range-Unit:  items` when doing a request from the command line trigger the same behaviour?
- A range in a `PATCH` request shouldn't affect the `queryTotal`, right?

ruslantalpa:
figured it out i think.
Take a look at this line
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/ApiRequest.hs#L150
and this line
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/ApiRequest.hs#L193
what happens is this.
the output of limitF is governed by the `range` parameter. Until now we used that only for read requests so mutate was not covered.
the range is calculated like this (for the first level off the tree, which is our case)
first we look at the headers, we get nothing, so the range is set to `allRange` (which is good, we need all the updated rows back) but then we check if we want only a single row, so we turn that to `singletonRange 0` (line 193).
We need to do that only in the case of read requests

try this quick fix (a bit ugly)

turn this
```
iRange = M.insert "limit" (rangeIntersection headerRange urlRange) $
```

into

```
iRange = M.insert "limit" (if action == ActionRead then (rangeIntersection headerRange urlRange) else allRange) $
```

if we get tests to pass then we can think about the logic cleanup

ehamberg:
Ah, interesting. The “a bit ugly” change above leads to my test succeeding, but quite a few other failures as you probably saw.

However, with the following change, all tests pass:

```diff
-  headerRange = if singular then singletonRange 0 else rangeRequested hdrs
+  headerRange = if singular && method == "GET" then singletonRange 0 else rangeRequested hdrs
```

The thinking here is basically that the singleton range should only be applied to read requests. Does something like this make sense? If so we should perhaps move the `action` variable from the `let … in …` block down to the `where` block so that we can use it directly in the test instead of comparing `method` against the bytestring `"GET"`:

```diff
diff --git i/src/PostgREST/ApiRequest.hs w/src/PostgREST/ApiRequest.hs
index ad5de09..7d9c392 100644
--- i/src/PostgREST/ApiRequest.hs
+++ w/src/PostgREST/ApiRequest.hs
@@ -92,58 +92,6 @@ data ApiRequest = ApiRequest {
 -- | Examines HTTP request and translates it into user intent.
 userApiRequest :: Schema -> Request -> RequestBody -> ApiRequest
 userApiRequest schema req reqBody =
-  let action =
-        if isTargetingProc
-          then
-            if method == "POST"
-               then ActionInvoke
-               else ActionInappropriate
-          else
-            case method of
-               "GET"     -> ActionRead
-               "POST"    -> ActionCreate
-               "PATCH"   -> ActionUpdate
-               "DELETE"  -> ActionDelete
-               "OPTIONS" -> ActionInfo
-               _         -> ActionInappropriate
-      target = case path of
-                 []            -> TargetRoot
-                 [table]       -> TargetIdent
-                                  $ QualifiedIdentifier schema table
-                 ["rpc", proc] -> TargetProc
-                                  $ QualifiedIdentifier schema proc
-                 other         -> TargetUnknown other
-      payload = case pickContentType (lookupHeader "content-type") of
-        Right ApplicationJSON ->
-          either (PayloadParseError . cs)
-            (\val -> case ensureUniform (pluralize val) of
-              Nothing -> PayloadParseError "All object keys must match"
-              Just json -> PayloadJSON json)
-            (JSON.eitherDecode reqBody)
-        Right TextCSV ->
-          either (PayloadParseError . cs)
-            (\val -> case ensureUniform (csvToJson val) of
-              Nothing -> PayloadParseError "All lines must have same number of fields"
-              Just json -> PayloadJSON json)
-            (CSV.decodeByName reqBody)
-        Right oa@OpenAPI ->
-          PayloadParseError $ "Content-type not acceptable: " <> cs (show oa)
-        -- This is a Left value because form-urlencoded is not a content
-        -- type which we ever use for responses, only something we handle
-        -- just this once for requests
-        Left "application/x-www-form-urlencoded" ->
-          PayloadJSON . UniformObjects . V.singleton . M.fromList
-                      . map (cs *** JSON.String . cs) . parseSimpleQuery
-                      $ cs reqBody
-        Left accept ->
-          PayloadParseError $
-            "Content-type not acceptable: " <> accept
-      relevantPayload = case action of
-        ActionCreate -> Just payload
-        ActionUpdate -> Just payload
-        ActionInvoke -> Just payload
-        _            -> Nothing in
-
   ApiRequest {
     iAction = action
   , iTarget = target
@@ -190,13 +138,66 @@ userApiRequest schema req reqBody =
   endingIn xx key = lastWord `elem` xx
     where lastWord = last $ T.split (=='.') key
 
-  headerRange = if singular then singletonRange 0 else rangeRequested hdrs
+  headerRange = if singular && action == ActionRead then singletonRange 0 else rangeRequested hdrs
   urlOffsetRange = rangeGeq . fromMaybe (0::Integer) $
     readMaybe =<< join (lookup "offset" qParams)
   urlRange = restrictRange
     (readMaybe =<< join (lookup "limit" qParams))
     urlOffsetRange
 
+  action =
+    if isTargetingProc
+      then
+        if method == "POST"
+           then ActionInvoke
+           else ActionInappropriate
+      else
+        case method of
+           "GET"     -> ActionRead
+           "POST"    -> ActionCreate
+           "PATCH"   -> ActionUpdate
+           "DELETE"  -> ActionDelete
+           "OPTIONS" -> ActionInfo
+           _         -> ActionInappropriate
+  target = case path of
+             []            -> TargetRoot
+             [table]       -> TargetIdent
+                              $ QualifiedIdentifier schema table
+             ["rpc", proc] -> TargetProc
+                              $ QualifiedIdentifier schema proc
+             other         -> TargetUnknown other
+  payload = case pickContentType (lookupHeader "content-type") of
+    Right ApplicationJSON ->
+      either (PayloadParseError . cs)
+        (\val -> case ensureUniform (pluralize val) of
+          Nothing -> PayloadParseError "All object keys must match"
+          Just json -> PayloadJSON json)
+        (JSON.eitherDecode reqBody)
+    Right TextCSV ->
+      either (PayloadParseError . cs)
+        (\val -> case ensureUniform (csvToJson val) of
+          Nothing -> PayloadParseError "All lines must have same number of fields"
+          Just json -> PayloadJSON json)
+        (CSV.decodeByName reqBody)
+    Right oa@OpenAPI ->
+      PayloadParseError $ "Content-type not acceptable: " <> cs (show oa)
+    -- This is a Left value because form-urlencoded is not a content
+    -- type which we ever use for responses, only something we handle
+    -- just this once for requests
+    Left "application/x-www-form-urlencoded" ->
+      PayloadJSON . UniformObjects . V.singleton . M.fromList
+                  . map (cs *** JSON.String . cs) . parseSimpleQuery
+                  $ cs reqBody
+    Left accept ->
+      PayloadParseError $
+        "Content-type not acceptable: " <> accept
+  relevantPayload = case action of
+    ActionCreate -> Just payload
+    ActionUpdate -> Just payload
+    ActionInvoke -> Just payload
+    _            -> Nothing
+
+
 -- PRIVATE ---------------------------------------------------------------
 
 {-|
```

What are your thoughts?

ruslantalpa:
the code i suggested (where many tests fail) and the code you suggested (with all tests passing) have the same idea at the core.
I tried to add the `if` at a higher level, that's all. My thinking was, unless it's a GET requests, don't look at all at range headers and limit parameters, but i guess i missed something.

I would say add the line so that all the tests pass, then we can take a look at it again.
At this stage i would suggest leaving the App.hs mostly like it is in your PR, and add a comment about the need for a consistent handling of range/limit across all code paths

ehamberg:
Yep, it's the same idea. :)

I'll go ahead and add that line.

ehamberg:
… and we're green again. :)

begriffs:
@ehamberg thanks for your continued work on this long running pull request.

I tried out the changes locally and it works nicely. One thing I noticed is that this behavior of raising an error for singular preference when results are plural does not match how we treat POST or GET.

With `return=representation;plurality=singular` bulk POST returns `[ {...} ]`. GET returns `{...}`. Looks like we could either add the error feature to POST and GET or make PATCH and POST act like GET and happily truncate results if asked.

ehamberg:
I added the error feature to `POST` requests too. Inserting several rows and asking for a singular response is most likely an error. I'm less sure about `GET`: `return=representation` is a no-op for `GET` requests (I assume), so the question is whether `plurality=singular` for a `GET` request should be a bad request. Maybe it should, actually? That's also indicative of a client error…

Btw, I just realized that it is possible to specify an ordering for `POST` that affects what's returned when `return=representation` is set (but not for `PATCH`, it seems). In that case the returned value with `plurality=singular` *would* be deterministic. The same is also of course the case for `GET`.

src/PostgREST/App.hs
```diff
@@ -21,6 +21,8 @@ import qualified Hasql.Transaction         as HT
 import           Text.Parsec.Error
 import           Text.ParserCombinators.Parsec (parse)
 
+import qualified Text.InterpolatedString.Perl6 as P6
```
ruslantalpa: import just `q` here, no need for entire module

ehamberg:
Fixed now. Only `q` is imported. The module is still imported qualified since `q` is used as a variable name throughout the file.

ruslantalpa:
Oh, i did not notice/consider that. I was just thinking of using it like in other modules. Looks good

begriffs:
Looks like this is almost ready to merge! One last thing, could you update the OpenAPI response in `src/PostgREST/ApiRequest.hs` to mention the plurality header for PATCH/POST? It should be a matter of copying and pasting from the GET request section.

ehamberg:
Yep. Let's tie up the last, loose ends. :)

I had a quick look at `ApiRequest.hs` looking for the OpenAPI code, but it wasn't obvious at all to me where the OpenAPI response is generated. Could you point me to where exactly the responses for `GET` are generated?

ruslantalpa:
it starts here https://github.com/begriffs/postgrest/blob/master/src/PostgREST/App.hs#L205-L216

ehamberg:
Ok. I added the new possible `Prefer` header value `return=representation;plurality=singular` for `POST` and `PATCH` requests now.

I verified this by creating the `addresses` table from the test suite and running `$ http localhost:3000 'Accept: application/openapi+json' | jq '.paths["/addresses"].post.parameters'` (same for `patch`).

begriffs:
@ehamberg thanks again, really appreciate your work!

PR #636, opened 2016-06-10
Remove unix dependency for tests

begriffs:
Allows `stack test` to build and pass on Windows.


ruslantalpa:
Merge
(Why was it there anyway?)

begriffs:
Not sure how it got there, but just discovered that it breaks tests on windows so it's gone now.

Issue #461, opened 2016-01-17
Ability to set addresses to listen on


user318:
Now postgrest accepts connections on all addresses. It will be good if there will be a way to supply it with specific addresses. So it can be run to listen on localhost only for example.

begriffs:
Sure, I can add a `--host` command line parameter. How high priority is this issue for you?

user318:
For me it is not high, because I now use it in test environment and can filter access to it using iptables. But in perspective, this option will be helpful for me and perhaps others. Thank you.

SpiritQuaddicted:
For anyone ending up here in 2024, you can set e.g. `server-host = "127.0.0.1"` in your config.

steve-chavez:
@SpiritQuaddicted Wonder if we should improve the docs to make that more clear.

Was https://docs.postgrest.org/en/v12/references/configuration.html#server-host not enough? Perhaps an example there?

SpiritQuaddicted:
An example would be great, I think. If I look back, I seem to have googled for `"postgrest" listen localhost` and there Google shows only a generic result for the configuration page, which I did not end up visiting and reading. Maybe an example would make search engines show it right away? Might be stupid silly search engine optimisation of perfectly fine documentation but I'd try it.

Maybe simply another 1-2 lines in the list of examples with `127.0.0.1` and one from https://datatracker.ietf.org/doc/html/rfc5737?

SpiritQuaddicted:
Just by looking at "In addition to the usual address options" I would not have felt comfortable to try `server-host = "127.0.0.1"` though to be honest. Explicit examples are super useful.

steve-chavez:
Hm, I see no harm in adding an example since we have some for other configs too:

- https://docs.postgrest.org/en/v12/references/configuration.html#server-unix-socket
- https://docs.postgrest.org/en/v12/references/configuration.html#jwt-role-claim-key


Done in https://github.com/PostgREST/postgrest/commit/28ebe37e07c8c5ff2f5307be5c5b0b0608a6dfed

PR #639, opened 2016-06-14
Provide a swagger2 spec for the dynamic API

hudayou:
Related issue: #144


begriffs:
Wow, this is great. I'll try it out tomorrow.

main/Main.hs
```diff
@@ -46,7 +47,9 @@ main = do
   hSetBuffering stdin  LineBuffering
   hSetBuffering stderr NoBuffering
 
-  conf <- readOptions
+  conf' <- readOptions
+  host  <- getHostName
```
ruslantalpa: The name that is resolved is rarely the public name for your api so probably better to have a config option with a default to 'localhost' ... i.e. no Network.BSD (might even break on windows)

postgrest.cabal
```diff
@@ -44,8 +44,8 @@ executable postgrest
                      , http-types
                      , interpolatedstring-perl6
                      , jwt
-                     , microlens >= 0.4.2 && < 0.5
-                     , microlens-aeson >= 2.1.1 && < 2.2
+                     , lens >=3.8 && < 5.0
```
ruslantalpa: lens was dropped in favore of microlense becasue it pull is a lot of dependencies. Can you refactor your code to only use microlense?

postgrest.cabal
```diff
@@ -122,6 +130,7 @@ library
                      , PostgREST.RangeQuery
                      , PostgREST.ApiRequest
                      , PostgREST.Types
+                     , PostgREST.ApiSpec
```
ruslantalpa: How about Swagger, ApiSpec implies like this is the only description that we want to support

src/PostgREST/App.hs
```diff
@@ -206,7 +205,7 @@ app dbStructure conf apiRequest =
         else return notFound
 
     (ActionRead, TargetRoot, Nothing) -> do
-      body <- encode <$> H.query schema accessibleTables
```
ruslantalpa: The current behaviour should remain unchanged and a custom header (Prefer: format=swagger) show switch which 'encoding function' is used

src/PostgREST/App.hs
```diff
@@ -206,7 +205,7 @@ app dbStructure conf apiRequest =
...
+      body <- (encodeApi . toTableInfo) <$> H.query schema accessibleTables
```
ruslantalpa: Both of these functions should live in the Swagger module

test/Feature/StructureSpec.hs
```diff
@@ -13,54 +13,6 @@ import Network.Wai.Test (SResponse(simpleHeaders))
 spec :: SpecWith Application
 spec = do
 
-  describe "GET /" $ do
```
ruslantalpa: This test should remain and pass

src/PostgREST/App.hs
```diff
@@ -220,8 +219,23 @@ app dbStructure conf apiRequest =
     (_, _, _) -> return notFound
 
  where
+  toTableInfo :: [Table] -> [(Table, [Column], [Text])]
+  toTableInfo = map (\t ->
```
ruslantalpa: While this function (along with encodeApi) tries to use available functionality (and it's the right thing to do at this point). the way it's implemented is quite inefficient (2 queries for each table).
In the future it should expect to receive a single "DbStructure" object based on which the schema is generated (for different output formats)

ruslantalpa:
As a general comment, can you try make a few changes so that you do not need to touch any file but Main/Config/App and the changes be minimal, mostly dropping the lens dependency and moving all the (new) functions inside the swagger module?


(those are the things that i think need to change, @begriffs thoughts?)

postgrest.cabal
hudayou: That's a dependency dragged in by swagger2:
https://github.com/haskell-servant/servant-swagger/issues/39

ruslantalpa: Ok :(

hudayou:
@ruslantalpa , besides the lens dependency, others could be done

ruslantalpa:
I am not familiar with swagger, does it have the notion or related types of objects?
Can it fully describe this type of queries?
`/projects?select=id,name,client{id,name},tasks{id,name}`
Without it, swagger will be able to describe only the most basic queries postgrest can do

src/PostgREST/App.hs
begriffs: The standard way to request a certain format is with the Accept header, so in this case

```HTTP
Accept: application/swagger+json
```

test/Feature/StructureSpec.hs
begriffs: ...and of course we'll want tests of the swagger output as well.

hudayou:
Needs to figure out how to do swagger validation later :-)

begriffs:
I just tried running the server and exploring it with [Swagger UI](https://github.com/swagger-api/swagger-ui) and the results are SO COOL!

hudayou:
Thanks, all the cool from the good foundation that postgrest provides.

src/PostgREST/App.hs
```diff
@@ -206,7 +204,9 @@ app dbStructure conf apiRequest =
         else return notFound
 
     (ActionRead, TargetRoot, Nothing) -> do
-      body <- encode <$> H.query schema accessibleTables
+      body <- if contentType == OpenAPI
```
ruslantalpa: can be a bit more clean
```
body <- encodeFn <$> H.query schema accessibleTables
where encodeFn = if contentType == OpenAPI then encodeApi . toTableInfo else encode
```
and maybe also the encodeFn definition can embed these lines
```
encodeApi ti = encodeOpenAPI ti host port
host = configHost conf
port = toInteger $ configPort conf
```

src/PostgREST/ApiRequest.hs
```diff
@@ -123,6 +124,8 @@ userApiRequest schema req reqBody =
               Nothing -> PayloadParseError "All lines must have same number of fields"
               Just json -> PayloadJSON json)
             (CSV.decodeByName reqBody)
+        Right OpenAPI ->
+          PayloadParseError "Content-type not acceptable: application/openapi+json; charset=utf-8"
```
begriffs: To avoid repeating the mime type you could do

```haskell
Right oa@OpenAPI ->
  PayloadParseError $ "Content-type not acceptable: " <> cs (show oa)
```

src/PostgREST/Config.hs
```diff
@@ -51,6 +52,7 @@ argParser = AppConfig
   <$> argument str (help "(REQUIRED) database connection string, e.g. postgres://user:pass@host:port/db" <> metavar "DB_URL")
   <*> strOption    (long "anonymous"  <> short 'a' <> help "(REQUIRED) postgres role to use for non-authenticated requests" <> metavar "ROLE")
   <*> strOption    (long "schema"     <> short 's' <> help "schema to use for API routes" <> metavar "NAME" <> value "public" <> showDefault)
+  <*> strOption    (long "hostname"   <> short 'n' <> help "hostname on which the HTTP server is running" <> metavar "HOSTNAME" <> value "localhost" <> showDefault)
```
begriffs: Might this conflict with the eventual solution of #461? In that issue the hostname parameter would mean the address the server should bind on, whereas in this one it means the public DNS by which clients will know the server. Maybe they'll be the same?

src/PostgREST/App.hs
```diff
@@ -206,7 +204,9 @@ app dbStructure conf apiRequest =
...
+                 then (encodeApi . toTableInfo) <$> H.query schema accessibleTables
+                 else encode <$> H.query schema accessibleTables
       return $ responseLBS status200 [jsonH] $ cs body
```
begriffs: This is kind of pedantic, but if when the swagger is returned then it should respond with a `Content-Type: application/openapi+json` header.

begriffs:
You have good Haskell style, I enjoyed reading the code in `OpenAPI.hs`.

In addition to the notes above here are some general things I noticed to fix:

* When I request a non-root route with `Accept: application/openapi+json` it succeeds and responds with JSON. I think it should be a an HTTP 415 (unsupported media type).
* When a user requests a resource using a Range header or limit param causing the result to be truncated then postgrest uses the status code 206 rather than 200 but the swagger result currently lists only 200.
* In addition to return=representation and return=minimal there is a return=none which is not documented in the swagger output. (That option is used to suppress even a Location header in the response to creating a new item for use in inserting to a write-only table)
* Add an entry to CHANGELOG.md
* How about a few basic tests? Such as
  * Requesting the root with the right Accept header returns a body containing the string `"swagger": "2.0"`
  * Attempting to request other routes in the swagger format fails with http 415

As Ruslan mentioned the resource embedding capability (such as `/movies?include=*,actor{*}`) is a powerful part of postgrest APIs. Do you think it's possible to document that kind of thing given the swagger format? I understand if you want to save that for a separate pull request since it's probably pretty hard, just wondering if you think it's possible.

One other thing is that we don't yet document the RPC calls (what are the function names, what arguments do they take) so that's another thing for us to look at later.

src/PostgREST/Config.hs
hudayou: Tried to combine these two, but met some problems:
https://hackage.haskell.org/package/warp-3.2.6/docs/Network-Wai-Handler-Warp.html#t:HostPreference
Special values such as the below are supported and also ip addresses:
* means HostAny
*4 means HostIPv4
!4 means HostIPv4Only
*6 means HostIPv6
!6 means HostIPv6Only

setHost seems accept FQDN only, localhost and other will cause an error.

hudayou:
Also in the morning was trying to work on a solution to [validate](https://github.com/seagreen/hjsonschema)
the swagger output against the [schema](https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/schemas/v2.0/schema.json).

Haven't succeeded yet, going to try that and also address some of the easy ones commented above.

src/PostgREST/Config.hs
ruslantalpa: i don't think it's a problem, we just need to document exactly what this 'hostname', and make it explicit that it's used when outputing swagger. Maybe it's not even ok to name it hostname, it's more of a url because to postgrest it might be someip:3000 but to outside world, the api i accessed using https:/some domain.com/api/v1/ so actually this parameters shoudl be the full url prefix.
the other thing is almost all the time an IP

hudayou:
Agreed, the generation of select pattern should be possible, but haven't got a solution yet, could be done in another pr.

Going to focus on adding test specs in the following days.

src/PostgREST/Config.hs
hudayou: Yes, the other one in #461 is actually an IP, though it is called setHost.

hudayou:
Done adding tests for OpenAPI validation which uses https://github.com/seagreen/hjsonschema/pull/38.
The added JSON fixtures are modified version of JSON draft v4 schema and OpenAPI schema to be able to be used together with https://github.com/seagreen/hjsonschema/.
Also tried to align with the final solution to set address to listen on.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
+- Ability to generate an OpenAPI spec - @mainx07, @hudayou, @ruslantalpa, @begriffs
```
ruslantalpa: it should really be just you here :)

src/PostgREST/App.hs
```diff
@@ -206,8 +204,13 @@ app dbStructure conf apiRequest =
         else return notFound
 
     (ActionRead, TargetRoot, Nothing) -> do
-      body <- encode <$> H.query schema accessibleTables
-      return $ responseLBS status200 [jsonH] $ cs body
+      let encodeApi ti = encodeOpenAPI ti host port
```
ruslantalpa: host and port here are the ones from command line which are used to define interface/port to bind to
usually that is going to be an ip and usually, postgrest is going to sit behind a proxy (nginx) and the api will be accessible using a nice url (https://mydomain.com/api/v1).
When you are going to request the openapi schema, it's not going to match the public url, it's gonna say `192.168.10.10:3000`

ruslantalpa:
Aside from host/port thing (and changelog line) everything looks great.
I would say sqash this all into a single commit and it should be merged

CHANGELOG.md
hudayou: Thanks, dude, I appreciate that, but I really see that as a community effort for all the time all the people spent on it. 👍

src/PostgREST/App.hs
hudayou: I did some thought on this. If we were introducing something like x-openapi-uri in the proxy to take precedence over that host / port. It is actually the same as:
 location /api/v1 {
        proxy_pass http://Host:Port/;
 }

hudayou:
@ruslantalpa , I'll squash them, thanks for the advice.

src/PostgREST/App.hs
ruslantalpa: I am not sure i follow, if you have a location like you mentioned with proxy_pass (that forwards requests to postgrest)
and someone requests the swagger schema, doesn't that response (json) contain the internal string 'host:port'? and then in the proxy, you would have to preprocess the response and only after that forward it to the user?

hudayou: Right, something was blocking my mind on that.
A header such as X-Forwarded-URI should take precedence over host and port for feeding information to the generated swagger api.
location /api/v1 {
proxy_pass http://Host:Port/;
proxy_set_header X-Forwarded-URI $request_uri;
}

ruslantalpa: Yes, that will work

ruslantalpa: Just realised that `X-Forwarded-URI` will help also with the generation of the Location header so maybe leave that for another PR and let's merge this one as is.

The next PR will address swagger/location behind a proxy separatly

begriffs:
Actually no need to manually squash; Github now allows a squash from the merge button.

And I'm thinking I might like to merge this one with its separate commits intact because they're self-contained and help someone understand the history of this feature.

ruslantalpa:
20+ commits in history for a feature seems excessive and harder to view a diff of what it took to add it

ehamberg:
No strong opinions on the subject, but I will say that by creating a merge commit with a message summarizing the changes and keeping the commits, someone looking at it later can *choose* whether to view it as one, big change or to dive into the separate commits.

PR #640, opened 2016-06-14
Name custom types and domains on OPTIONS requests

martingms:
I use quite a lot of custom types in my schema, and would like to get the actual typenames when doing an OPTIONS request on a table. Right now, postgrest returns 'USER-DEFINED' if it is a custom type, and the base type if it is a domain. I had a look at the code and patched it to allow my usecase. I couldn't think of any reason to treat domains differently than other types in this context, or if outputting user-defined types in general is a bad idea?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Output names of used-defined types (instead of 'USER-DEFINED') - @martingms
+
 ### Fixed
 
 ## [0.3.2.0] - 2016-06-10
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -285,18 +285,8 @@ allColumns tabs =
                         ELSE 'YES'::text
                     END::information_schema.yes_or_no AS is_nullable,
                     CASE
-                        WHEN t.typtype = 'd'::"char" THEN
-                        CASE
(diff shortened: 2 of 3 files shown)


martingms:
I am not quite sure whether it should be `WHEN bt.typelem <> 0::oid AND bt.typlen = (-1) THEN 'ARRAY'::text` or `WHEN t.typelem <> 0::oid AND t.typlen = (-1) THEN 'ARRAY'::text` however, doesn't seem to make a difference in my testing.

ruslantalpa:
This query comes from the view definition of information_schema.columns.

how about changin only the 'ELSE' part and leave the rest unchanged
`ELSE 'USER-DEFINED'::text` to `ELSE format_type(a.atttypid, a.atttypmod)`

your changes should not break this test
 -    "type": "character varying",
 +   "type": "character varying(255)"

ruslantalpa:
this value is not used anywhere internaly i think, only for output

begriffs:
This PR makes a lot of sense. I'll review it more carefully tomorrow.

martingms:
@ruslantalpa Sure, I can fix that. I just didn't see the reason for treating domains with a specific case? But, as long as I get actual names for both domains and custom types, I'm happy, so whatever you decide.

begriffs:
The original query has so much in it I feel like there must be a reason it deals with the domain case vs the other case. I'd say you should probably keep both cases and just modify them to use `format_type(...)` rather than `'USER-DEFINED'`.

martingms:
Looks good?

ruslantalpa:
Yes

begriffs:
Looks good, can you add an entry to the changelog as well? Then I'll merge!

PR #645, opened 2016-06-16
how to install pre-built packages ?

waghanza:
Binaries are automatically built by https://github.com/waghanza/plunder.

We can procedure to the documentation, but how to specify that packages :

+ are built by http://stackage.org
+ are not **official** packages

### DIFF
--- a/docs/install/server.md
+++ b/docs/install/server.md
@@ -1,5 +1,37 @@
 ## Installation
 
+### Install pre-built packages
+
+Some linux packages are availables on http://packagecloud.io
+
++ **debian**
++ **ubuntu**
++ **fedora**
++ **enterprise linux** (`centos`, `rhel`, `scientific linux`)
+
+:warning: Package are built for `amd64` architecture :warning:
+
+#### debian / ubuntu
+
+```sh
+curl -s https://packagecloud.io/install/repositories/waghanza/plunder/script.deb.sh | sudo bash
+apt-get install postgrest
+```
+
+#### enterprise linux
+
+```sh
+curl -s https://packagecloud.io/install/repositories/waghanza/plunder/script.rpm.sh | sudo bash
+yum install postgrest
+```
+
+#### fedora
+
+```sh
+curl -s https://packagecloud.io/install/repositories/waghanza/plunder/script.rpm.sh | sudo bash
+dnf install postgrest
+```
+
 ### Installing from Pre-Built Release
 
 The [release page](https://github.com/begriffs/postgrest/releases/latest)


begriffs:
Thanks for your work building postgrest packages on various platforms. I'm glad you opened a pull request because it's a good place to talk about our strategy for packaging.

Let's compare the proposed way to install on Ubuntu vs the current way.

#### Proposed method
```sh
curl -s https://packagecloud.io/install/repositories/waghanza/plunder/script.deb.sh | sudo bash
apt-get install postgrest
```

#### Existing method
```sh
curl -L https://github.com/begriffs/postgrest/releases/download/v0.3.2.0/postgrest-0.3.2.0-ubuntu.tar.xz
tar zxf postgrest-0.3.2.0-ubuntu.tar.xz
```

These commands are about the same length. The advantage of your method is that it puts the postgrest binary into a system directory in the path. However it has a few disadvantages:
* requires sudo
* requires running a mystery script
* requires keeping the packages up to date on packagecloud

I think the disadvantages outweigh the advantage, what do you think?

Another option would be to get postgrest into the official package repositories for linux distributions. (We have it in Homebrew already.) That would be super convenient for people because the installation would then look like

#### "Official" method
```sh
sudo apt-get install postgrest
```

Do you think you could help with this?

waghanza:
In fact with [plunder](https://github.com/waghanza/plunder), I _initially_ planned to provide **official** packages (supported by each _linux communities_). 

However, I _think_ that the first step is to **avoid** packaging manual works :wink:.  I want to facilitate application packaging, and even created a https://github webhooks to distribute oss treasures when a release is created. (To grant more **development** time than packaging time)

+ In one hand, I privileged created binary (cause I need my app to do othe stuff).
+ In the other hand, **official** _linux_ packages (specially the most use **debian**) is very hard / long
   + Packages needs to be source package
   + Packages needs to be compiled with supported libs (so no [stackage](https://www.stackage.org) I think, I'm currently learning how to provide **haskell** packages onto **debian** communities)
    + Process is extremely long (especially for **debian**), integration in 
      + testing
      + unstable
      + jessie

Official packaging is planned, but I think in some month :wink:

PS : It will perhaps be a good idea to backlink https://packagrecloud.io doc on what is made by the mysterious script :man_with_gua_pi_mao:

begriffs:
You're right about those extra benefits, I hadn't thought of those.

However like I said the very best thing will be to have official distribution packages. I think we can do it, including the building from source part. We can use a script similar to how we do it for Homebrew: https://github.com/Homebrew/homebrew-core/blob/master/Formula/postgrest.rb

If official debian packages are difficult to make, are other distros easier? How hard are RPMs to build? I feel like if we're going to have packages we should have the real thing.

waghanza:
In fact, **debian** packages are **NOT** difficult to distribute _officially_ (only **very long**) but it's not the main problem here, I think.

:warning: `plunder` aims not to provide **official** packages, only facilitate opensource  **distribution** :warning:

The problem is that we have creating a matrix (_plunder_ or an other), but :
* we need **sources** packages to add _postgrest_ in **official** repositories (e.g : `apt-get install postgrest` for **debian** like distro)
* we need to use the be compatible with all dependencies and compiler versions to add _postgrest_ in **official** repositories

For example, we can use `docker` to build for **debian** as **docker** image is a **debian**, but this provide a **binary** so not compatible with **official** contributions.

~~~dockerfile
FROM haskell:latest
ENV DEBIAN_FRONTEND noninterractive
RUN apt-get -y update
RUN apt-get -y install libpq-dev git
RUN git clone https://github.com/begriffs/postgrest /opt/app
WORKDIR /opt/app
RUN git tag | grep -v beta | sort -V | tail -n1 | xargs git checkout
RUN cabal update
RUN cabal install postgrest.cabal
RUN cabal build
~~~

For me we need, to distribute `postgrest` in several steps :

* [x] Create a matrix to provide binaries
* [x] Update the doc to allow users to install `postgrest` binaries
* [ ] Add control scripts / startup scripts / configuration values ...
* [ ] Update the doc to allow users to install `postgrest` binaries with 
* [ ] Create source package
* [ ] Ask permission :man: to provide in official way
* [ ] Update the doc

begriffs:
Thanks for looking into this. I'm interested in packaging postgrest but only as official packages for each distribution.

waghanza:
@begriffs still working on the packaging stuff ...

**official** packages could neither be provided for :

* (**stable*) `debian` 
* `centos`

but could be made for 
* `ubuntu`
* `fedora`

PS : I am targetting **linux** (I'll see after for **bsd** ...)

begriffs:
Thanks for your efforts with this. What's the issue with debian and centos?

waghanza:
Thé stable version are still on 7.6 version officielle ghc

PS : I try tout build with ghc / cabal, sincebuilding with stack will remove ability tout provide communiquer packages

waghanza:
Hi @begriffs, 

Still working on packaging this app. 

I have changed my worflow a little. In plan to upload a binary package (currently only **debian**) as a release (on https://github.com).

After, uploading on https://github.com, I will ask debian community mentors to add **officially** ad d _postgrest_ on **debian**.

What do you think ?

begriffs:
Sorry, I really don't know much about Debian packaging. Has the Debian community accepted the official package?

waghanza:
@begriffs In fact I have not asked (yet). I'll make a clean **deb** and propose to **debian** `mentors` (only a `mentor` could accept your package)

begriffs:
Any news with the packaging saga?

waghanza:
Nop. 

In fact, I'm still working on it ...

When I've finished, I'll push my work to **debian** community and make an update in this issue.

PS : the update will probably come in September

begriffs:
I saw some packaging updates on twitter and thought it was yours for a second, but I guess it's Arch.

https://twitter.com/arch_update_bot/status/913465748860813312
https://twitter.com/arch_update_bot/status/913467203504058368

waghanza:
@begriffs Hi, I'm actually very silent on gitter / github, but still packaging this. To be honest with you I want to let you (or repository maintainer) to release app on github so as a auto-packaging tool can upload binaries on github

PR #649, opened 2016-06-24
Add proxy awareness for OpenAPI spec generation

hudayou:
Proxy command line option over manually specified host, port

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Ability to generate an OpenAPI spec - @mainx07, @hudayou, @ruslantalpa, @begriffs
+- Ability to generate an OpenAPI spec behind a proxy- @hudayou
 - Ability to set addresses to listen on - @hudayou
 - Filtering, shaping and embedding with &select for the /rpc path - @ruslantalpa
 - Output names of used-defined types (instead of 'USER-DEFINED') - @martingms
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -8,6 +8,7 @@ import           PostgREST.Config                     (AppConfig (..),
                                                        minimumPgVersion,
                                                        prettyVersion,
                                                        readOptions)
+import           PostgREST.OpenAPI                    (isMalformedProxyUri)
 import           PostgREST.DbStructure
 
 import           Control.Monad
@@ -50,11 +51,15 @@ main = do
   conf <- readOptions
   let host = configHost conf
       port = configPort conf
+      proxy = configProxyUri conf
       pgSettings = cs (configDatabase conf)
       appSettings = setHost (fromString host)
                   . setPort port
                   . setServerName (cs $ "postgrest/" <> prettyVersion)
                   $ defaultSettings
+
+  when (isMalformedProxyUri proxy) $ error
+    "Malformed proxy uri, a correct example: https://example.com:8443/basePath"
 
   unless (secret "secret" /= configJwtSecret conf) $
     putStrLn "WARNING, running in insecure mode, JWT secret is the default value"
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -67,6 +67,7 @@ executable postgrest
                      , warp >= 3.1.0
                      , insert-ordered-containers >= 0.1.0.1
                      , swagger2 >= 2.1
+                     , network-uri >= 2.6.1.0
                      , HTTP
                      , Ranged-sets
                      , protolude >= 0.1.5 && < 0.2.0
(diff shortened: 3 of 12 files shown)


src/PostgREST/ApiRequest.hs
```diff
@@ -152,6 +157,7 @@ userApiRequest schema req reqBody =
   , iPreferRepresentation = representation
   , iPreferSingular = singular
   , iPreferCount = not $ singular || hasPrefer "count=none"
+  , iProxy = pickProxy $ lookupHeader "x-forwarded-uri"
```
ruslantalpa: since this is used only for openAPI, how about string here only the raw value of the header and call the pickProxy only when generating the openapi output, and while doing that ... move all these new functions in the openapi module

src/PostgREST/Middleware.hs
```diff
@@ -50,9 +51,16 @@ unsupportedAccept app req respond =
...
+  where proxyUri = lookup "x-forwarded-uri" $ requestHeaders req
+        malformedProxyUriRespond = respond $ errResponse status400 $ "Malformed X-Forwarded-URI header: " <> cs (fromJust proxyUri)
+
 defaultMiddle :: Application -> Application
 defaultMiddle =
     gzip def
   . cors corsPolicy
   . staticPolicy (only [("favicon.ico", "static/favicon.ico")])
   . unsupportedAccept
+  . malformedProxyUri
```
ruslantalpa: while behind the proxy, there is going to be a proxy uri header on each request, and it's going to get checked and parsed on each request just to confirm it's correct even though you do nothing with it later.
there could be an argument for using this in the location header but it seems too much computation with not much benefit. the location header can be easily adjusted by the proxy with
`add_header Content-Location  /prefix/goes/here/$upstream_http_content_location;`

src/PostgREST/ApiRequest.hs
```diff
@@ -271,3 +277,85 @@ ensureUniform arr =
...
+  A valid proxy uri should be an absolute uri without query and user info,
+  only http(s) schemes are valid, port number range is 1-65535.
+
+  For example
+  http://postgrest.com/openapi.json
+  https://postgrest.com:8080/openapi.json
+-}
+isMalformedProxyUri :: Maybe BS.ByteString -> Bool
+isMalformedProxyUri Nothing =  False
+isMalformedProxyUri (Just uri)
```
ruslantalpa: considering the fact that this affects only the json output for the schema and that most of the time (if not all) the person deploying postgrest will also be the one in control of the proxy, what's the downside of just trusting the uri is valid ... treat it as a string?

ruslantalpa:
so in the end ... isn't it simpler to just take in a command line parameter --public-uri or something like that and use it for openapi json?

hudayou:
@ruslantalpa A command line option seems to be much better if the proxy URI is not anywhere else.

ruslantalpa:
Lets leave this open for a while and think is there are usecases where you can not use the command line and have to work with the headers

begriffs:
The command line parameter does sound easier. I guess if you have more than one proxy going to a single postgrest server then they might each use a different public url, but that seems like an uncommon case.

hudayou:
Trying to see whether there is a workaround for Circle CI.

@begriffs  Also I suspect Circle CI's CPU quota is low
```
	Tue Jun 28 09:44 2016 Time and Allocation Profiling Report  (Final)

	   spec +RTS -N -sstderr -p -RTS

	total time  =       11.99 secs   (11989 ticks @ 1000 us, 1 processor)
	total alloc = 13,430,147,528 bytes  (excludes profiling overheads)

COST CENTRE             MODULE                                  %time %alloc

toJSON.mapMaybe         Data.JsonSchema.Draft4.Schema            31.7   31.0
parseJSON               Data.Validator.Draft4.Any                11.1   19.1
internalSchemaHashMap   Data.JsonSchema.Draft4.Schema            10.4    6.8
parseJSON.checkUnique   Data.Validator.Draft4.Any                 8.9   14.3
unstream/resize         Data.Text.Internal.Fusion                 5.1    5.3
internalSchemaHashMap.f Data.JsonSchema.Draft4.Schema             4.5    1.3
parseJSON.\             Data.JsonSchema.Draft4.Schema             3.7    4.4
parseJSON               Data.Validator.Utils                      2.4    3.9
parseJSON               Data.Validator.Draft4.Object.Properties   2.2    3.2
toJSON                  Data.JsonSchema.Draft4.Schema             2.2    1.6
parseJSON               Data.Validator.Draft4.Array               1.1    1.9
parseJSON.checkUnique   Data.Validator.Draft4.Object              1.1    1.3
```

ruslantalpa:
The new simplified proxy url from command line looks nice and clearly separated in it's own module, validation is done only once on startup.
Nice work 👍

hudayou:
@begriffs, Really have no idea about the behavior of Circle, can we report a bug to them  :-p?
In the following build, the test costs only about 25 seconds:
https://circleci.com/gh/begriffs/postgrest/1241

Unless the CI is fixed, the only workable workaround I found is to Skip the OpenAPI specs on Circle CI.

begriffs:
Tried this out locally and it works nicely. Could you add a note to the changelog? Then I'll merge. (Skipping the open api test on ci is fine for now.)

PR #655, opened 2016-06-28
A helper script that makes it simpler to spin up PostgREST in Docker containers

deinspanjer:
I've been making lots of use out of the Docker image, but I needed to share a script with others on my team to make it easier for them to quickly spin up containers.

The script is a little quirky, but I was aiming for simple to write and straightforward rather than nice getopts and usage docs and such.  I figured it couldn't hurt to share it upstream in case it is of interest to either include or provides ideas about a better way to do something like it.

### DIFF
--- a/run_docker_postgres.sh
+++ b/run_docker_postgres.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+#This helper script lets you quickly and easily run one or more PostgREST containers connected to any Postgres DB and schema.
+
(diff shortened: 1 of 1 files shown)


begriffs:
Thanks for sharing this. Maybe the best way to incorporate docker information would be to add a Docker section to our docs. https://github.com/begriffs/postgrest-docs/blob/master/docs/admin/deployment.md Would you be interested in writing this?

deinspanjer:
Absolutely.  I'll put together some info and can just provide the shell script in a pre-formatted code block there.

PR #656, opened 2016-06-29
Remove unnecessary string conversions

diogob:
While I was working in another refactor when I came across this code that can be simplified right away.
As it turns out it maps a `[Text]` to another `[Text]` and a `Maybe Text` to another `Maybe Text` without changing the value.


begriffs:
Nice, thanks.

PR #657, opened 2016-06-29
Protolude first fugues

diogob:
This is the first of a series of PRs to banish String from our code and replace the default `Prelude` provided by haskell with a custom library.
I chose the [protolude](https://github.com/sdiehl/protolude/) for the sake of simplicity (it has less dependencies than [classy-prelude](https://github.com/snoyberg/classy-prelude) and focus only on very basic functionality). 

Some advantages we get from this:
- We have a default setup that will be easier for newcomer to work safely, for we won't have any more partial functions included by default. 
- We get rid of `string-conversions` as the `Show` class becomes polymorfic.
- We have much less work converting strings since most of `Wai` data comes in `ByteString` already.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,7 +17,7 @@ dependencies:
 
 test:
   override:
-    - stack test
+    - stack test --test-arguments "--skip \"returns a valid openapi\""
     - git ls-files | grep '\.l\?hs$' | xargs stack exec -- hlint -X QuasiQuotes -X NoPatternSynonyms "$@"
     - stack exec -- cabal update
     - stack exec --no-ghc-package-path -- cabal install --only-d --dry-run
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -4,13 +4,9 @@
 
 module PostgREST.Error (pgErrResponse, errResponse) where
 
-
+import           Protolude
 import           Data.Aeson                ((.=))
 import qualified Data.Aeson                as JSON
-import           Data.Maybe                (fromMaybe)
-import           Data.Monoid               ((<>))
-import           Data.String.Conversions   (cs)
-import           Data.Text                 (Text)
 import qualified Data.Text                 as T
 import qualified Hasql.Pool                as P
 import qualified Hasql.Session             as H
@@ -19,7 +15,7 @@ import qualified Network.HTTP.Types.Status as HT
 import           Network.Wai               (Response, responseLBS)
 
 errResponse :: HT.Status -> Text -> Response
-errResponse status message = responseLBS status [(hContentType, "application/json")] (cs $ T.concat ["{\"message\":\"",message,"\"}"])
+errResponse status message = responseLBS status [(hContentType, "application/json")] (toS $ T.concat ["{\"message\":\"",message,"\"}"])
 
 pgErrResponse :: Bool -> P.UsageError -> Response
 pgErrResponse authed e =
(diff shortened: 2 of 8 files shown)


diogob:
@begriffs circle is failing because of a timeout, but the build is green ;)

ruslantalpa:
it fails because of a specific test, there is a PR that disables that test in Circle so maybe copy that line from there to make your PR pass

src/PostgREST/Error.hs
```diff
@@ -4,13 +4,9 @@
...
 module PostgREST.Error (pgErrResponse, errResponse) where
 
-
+import           Protolude
 import           Data.Aeson                ((.=))
 import qualified Data.Aeson                as JSON
-import           Data.Maybe                (fromMaybe)
-import           Data.Monoid               ((<>))
-import           Data.String.Conversions   (cs)
-import           Data.Text                 (Text)
```
begriffs: Yeah! Bye bye long import list.

circle.yml
```diff
@@ -17,7 +17,7 @@ dependencies:
 
 test:
   override:
-    - stack test
+    - stack test --test-arguments "--skip \"returns a valid openapi\""
```
begriffs: I agree, this is fine to skip on CI. We run the tests locally all the time anyway and this one is unlikely to break and comes from a library that probably has its own test suite.

src/PostgREST/Types.hs
```diff
@@ -54,12 +54,12 @@ data PrimaryKey = PrimaryKey {
 } deriving (Show, Eq)
 
 data OrderDirection = OrderAsc | OrderDesc deriving (Eq)
-instance Show OrderDirection where
+instance S.Show OrderDirection where
```
begriffs: I'm assuming we'll change this to the protolude Show instance once other modules are converted to protolude?

diogob: Protolude does not define a new typeclass, it only defines a polymorphic show function and does not expose the `show` from Prelude.
The only reason I've used this qualified was to be able to define new instances of Show since it is not imported in protolude.
But since then I've inspected the `GHC.Show` module and realized that would be silly to qualify it, since it does not conflict with anything, so I'm adding a commit to do that.
We can add GHC.Show to our custom Prelude later on if we want.

begriffs:
This is a great start. Merging!

src/PostgREST/Error.hs
```diff
@@ -35,41 +31,41 @@ instance JSON.ToJSON P.UsageError where
...
 instance JSON.ToJSON H.Error where
   toJSON (H.ResultError (H.ServerError c m d h)) = JSON.object [
-    "code" .= (cs c::T.Text),
-    "message" .= (cs m::T.Text),
-    "details" .= (fmap cs d::Maybe T.Text),
-    "hint" .= (fmap cs h::Maybe T.Text)]
+    "code" .= (toS c::T.Text),
+    "message" .= (toS m::T.Text),
+    "details" .= (fmap toS d::Maybe T.Text),
+    "hint" .= (fmap toS h::Maybe T.Text)]
```
ruslantalpa: shouldn't this be `Text` instead of `T.Text`?

src/PostgREST/Error.hs
```diff
@@ -19,7 +15,7 @@ import qualified Network.HTTP.Types.Status as HT
 import           Network.Wai               (Response, responseLBS)
 
 errResponse :: HT.Status -> Text -> Response
-errResponse status message = responseLBS status [(hContentType, "application/json")] (cs $ T.concat ["{\"message\":\"",message,"\"}"])
+errResponse status message = responseLBS status [(hContentType, "application/json")] (toS $ T.concat ["{\"message\":\"",message,"\"}"])
```
ruslantalpa: also here `concat` should work instead of `T.concat`

begriffs:
Looks like I merged prematurely. Well you can include these changes in the next protolude pull request.

diogob:
Sure, I'll continue working on this during the upcoming week and I'll send in the next batch.
I'm expecting at least 2 more PRs.

PR #662, opened 2016-07-07
Translate raise_exception to http 400

### LINKED ISSUE #661: /rpc: return 400 for raise 'exception'
I use a lot of `/rpc` functions. Inside of those, I'm raising exceptions and want to return `400` if the client sent syntactically or semantically invalid data. 

Right now `postgrest` returns `500` for most exceptions, especially for the standard `P0001` that is thrown when using `raise 'mymessage'`. 

I think `postgrest` should return `400` for `P0001`:

* instead of the verbose `raise data_exception using message = 'my message'` a simple `raise 'my message'` would work.
* Also I don't think using `data_exception` (which `postgrest` converts to `400`) is correct in my case, but I'm not sure which one to use of the ones listed in the [documentation](https://www.postgresql.org/docs/9.3/static/errcodes-appendix.html).

begriffs:
Fixes #661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Do not apply limit to parent items - @ruslantalpa
 
+### Changed
+- Use HTTP 400 for raise\_exception - @begriffs
+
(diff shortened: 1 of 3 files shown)


ruslantalpa:
exactly what i did on my branch.

PR #665, opened 2016-07-08
nested jwt

schmittr:
as described in https://github.com/begriffs/postgrest/issues/550

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ codex.tags
 .stack-work
 tags
 site
+
+# intellij
+.idea
+.iml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
+- Ability to return mixed type json claims, which include JWT - @schmittr
 - Ability to generate an OpenAPI spec - @mainx07, @hudayou, @ruslantalpa, @begriffs
 - Ability to set addresses to listen on - @hudayou
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -201,7 +201,7 @@ app dbStructure conf apiRequest =
             (if returnsJWT
             then returnJson $ "{\"token\":\"" <> cs (tokenJWT jwtSecret body) <> "\"}"
             else do
-              let containsJWT = fromMaybe False $ isInfixOf "jwt_claims" <$> returnType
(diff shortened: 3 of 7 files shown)


ruslantalpa:
The goal of this PR can be achieved in the db with sql using this https://github.com/michelp/pgjwt :)

diogob:
@begriffs @ruslantalpa @schmittr maybe we should move all JWT related logic to the database using this pgjwt.
As it depends only on pgcrypto is pretty save for Heroku/RDS and similar cloud services.
It would be one less special case to deal with and we can drop the jwt_claims type requirement for the `0.4`.

Thoughts?

begriffs:
@schmittr thanks for the thought and time you put into this pull request.

@ruslantalpa pgjwt looks wonderful, and only 63 lines long internally! I'm always surprised at the ingenious tools people make for postgres.

@diogob you're right, we should remove the magic `jwt_claims` return type in 0.4 and have people use pgjwt. In fact due to the new postgrest RPC changes, functions can return unadorned JWT.

Schmittr, can you double-check that pgjwt allows you do do the things that you created this pull request to do?

ruslantalpa:
jwt_claims can be removed since that is output as response body can be controlled from db however things that are headers/response code ... will need at this point a `magic` return type.

schmittr:
In case of nested JWT this wouldn't really be an issue. One could include the headerinformation into the JWT json. As it would also be possible like that to use several token with different properties (although this would probably not very common to use), which means more DOF. Further, using the psql approach enables not only a one-depth nested token as my approach does. One could build a completely custom json tree with multiple token with different algos (which is maybe very far away from the standard :smile:).

The open question would be how to apply the header information in case of an unnested token (if we want to match the standard). But as I saw in https://github.com/begriffs/postgrest/pull/664 you guys already think about propagating header info from psql to postgrest which sounds pretty neat to me :wink:

But that's all theoretically, I didn't practically test pgjwt, yet. What do you think?

begriffs:
Let's try accomplishing this with pgjwt instead. I'll close this for now.

schmittr:
Yep, I think that's the best. But another thought: wouldn't it make sense with pgjwt to propagate the secret into let's say postgrest.secret? Surely, it's possible to do so by env-variables but seems to me more sober by creating an postgest variable.

ruslantalpa:
it will surely be convenient, we can de a `set` on postgrest startup, not hard ... arguments against?

schmittr:
What's the current state of the header-sending feature? As far as I understand this was the only issue, which separates this approach here from the standard. Imho, it would be nice to match the standard instead of using some custom workaround.

PR #677, opened 2016-07-20
Protolude second movement

diogob:
This PR continues replacing Prelude for Protolude in several modules.
I believe once everything is converted we will drop a lot of of the `toS` calls for every single string will be represented as `Text`.

Modules converted in this PR:
 * Middleware
 * App
 * Main
 * RangeQuery
 * OpenAPI

Modules still missing:
* ApiRequest
* Config
* DbStructure
* Parsers
* QueryBuilder

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -2,7 +2,7 @@
 
 module Main where
 
-import           Prelude
+import           Protolude
 import           PostgREST.App
 import           PostgREST.Config                     (AppConfig (..),
                                                        minimumPgVersion,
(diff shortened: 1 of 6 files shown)


begriffs:
So happy to see this cleanup! I'll review in more detail later tonight or tomorrow.

begriffs:
Great stuff.

begriffs:
I noticed if I use the stack nightly-2016-07-19 resolver which is GHC 8.0.1 it causes an error:

```
postgrest/src/PostgREST/App.hs:121:25: error:
    Ambiguous occurrence ‘<>’
    It could refer to either ‘Data.Monoid.<>’,
                             imported from ‘Data.Monoid’ at src/PostgREST/App.hs:32:1-28
                          or ‘Protolude.<>’,
                             imported from ‘Protolude’ at src/PostgREST/App.hs:65:1-73
                             (and originally defined in ‘Data.Semigroup’)
```

The error went away when I removed the import of `Data.Monoid`. No big deal I guess we'll deal with it when we upgrade the resolver.

Issue #56, opened 2014-10-07
Move table structure body out of the OPTIONS


begriffs:
Use rel="describedby" link header in the regular get response

https://www.iana.org/assignments/link-relations/link-relations.xml
https://www.mnot.net/blog/2012/10/29/NO_OPTIONS

Also include rel="describes" in the description url link headers.

begriffs:
This information has been moved out of the OPTIONS for individual endpoints and consolidated in the root API description.

begriffs:
Reopening this issue because people have been asking how to find out the primary and foreign keys of a relation in order to do resource embedding, and the openapi description is proving insufficient.

As mentioned above, I'm considering having each endpoint include a [describedby](https://www.mnot.net/blog/2012/10/29/NO_OPTIONS#what-should-i-use-instead) http header link. It would go to another endpoint that provides metadata about the original. Kind of like our old OPTIONS way but without abusing the options verb.

ruslantalpa:
They can create a custom view that describes their schema with all the details they need.
Here is a base query to start from
https://gist.github.com/ruslantalpa/b2f10eb1b5f6dd0fc1c154e071a1c91b

steve-chavez:
As mentioned in https://github.com/begriffs/postgrest/issues/790#issuecomment-381645515 having the users provide their own query is redundant work, the ideas here are worth considering, let's follow this issue on #790.

PR #693, opened 2016-08-13
Per-route content negotiation, and OpenAPI only for root

begriffs:
This is a backward-incompatible change for 0.4. Requesting `/` used to return the API schema using a custom JSON format, or return an OpenAPI description depending on Accept headers. This PR removes support for the non-OpenAPI response. We still have a custom response for the OPTIONS verb on specific tables.

This also allows each route to customize content negotiation for the Content-Types it supports. It will properly reject Accept headers that are unsatisfiable.

Two concerns about the changes:
* I removed a test for authorization which checks that the schema includes only the tables that the currently authenticated user can see. The OpenAPI equivalent should be added.
* It strips charset encoding from Accept headers and uses UTF-8 for responses. This might not be correct HTTP semantics. For instance if a client said `Accept: application/json; charset=ISO-8859-1` we would not reply with HTTP 415 (Unsupported media type) but instead would succeed with `Content-Type: application/json; charset=UTF-8`.


ruslantalpa:
i would say remove `options` for each table. there is not much value in it if the openapi describes all routes and all tables and fields/columns

begriffs:
I removed the API description from OPTIONS responses, but kept the handlers there to deal with CORS requests. Unfortunately we've lost test coverage of foreign key stuff, and the test that users with less db permissions are unable to see tables they cannot access.

ruslantalpa:
This part https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Types.hs#L134-L163 is also probably not needed anymore but you are right about those tests, they are needed.
Any way to add a completely new type of tests to does not go through `warp` code but just tests the output of functions like `dbColumns`?

begriffs:
Closes #56
Closes #144

begriffs:
@ruslantalpa could you help with adding tests for the relations in another PR?

begriffs:
I'm going to merge this. Let's think of how we might add more test coverage for relations.

PR #698, opened 2016-08-19
Use comma rather than semicolon to split header values

### LINKED ISSUE #627: Prefer: return=representation AND Prefer: plurality=singular

begriffs:
This is an addon to #634 which addresses @nerfpops' comment about the Prefer header separator. [RFC7240](https://tools.ietf.org/html/rfc7240) specifies that comma rather than semicolon should separate multiple Prefer values.

Closes #627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Use HTTP 400 for raise\_exception - @begriffs
 - Remove non-OpenAPI schema description - @begriffs
+- Use comma rather than semicolon to separate Prefer header values - @begriffs
 
 ## [0.3.2.0] - 2016-06-10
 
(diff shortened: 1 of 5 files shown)


SebAlbert:
I don't actually see how RFC7240 says that comma be preferred over semicolon. It just seems to sometimes use one and sometimes the other. Especilaly the ABNF useses semicolon, as far as I understand. It seems we'd better support just both...

SebAlbert:
Ah, never mind. I found the answer in the [Errata](https://www.rfc-editor.org/errata_search.php?rfc=7240) - so it's rather RFC7230 specifying the use of a comma :)

PR #700, opened 2016-08-28
No count by default

### LINKED ISSUE #273: Get rid of count(1) by default
(This is related to #255)

In MANY cases the app is designed in the way where count is not needed at all. For example if you query items and show them to user in twitter/pinterest/facebook-like "infinite" list, when at the first step, only Top-N items are selected, then, for the second portion, again, Top-N items are selected but with some "shift" (using `id` or `created` column usually), and this process can be repeated many times.

This very common task. And usually you have many millions items in the table and query should be executed in milliseconds.

So in such cases count(1) is not needed at all. Moreover, estimated count is not needed either since the app doesn't show user any information about it (no items count, no page number). I implemented solutions with estimated count in a couple of large projects (dozends Ms of users) and unfortunately such approaches fail very fast – in cases when you have some filters where Postgres has no good statistics and makes error estimations (not to mention that it does require many additional milliseconds – see below). In many cases, it's better to show nothing to user than to show wrong information. So in my personal cases I decided to stop using estimated count for large datasets and design apps in a way where this info is not needed.

However, in some cases count(1) is useful of course and should be provided.

The main thing here is that for years Postgres has slow count(1). This is very serious problem and people who use Postgres know how to handle it and how to live with it. 

What I want to say is that by default, count(1) is not needed. I propose to skip it by default and show only if user asks for it. This is separate issue than #255 since we would have 3 options: 
 - no count
 - exact count
 - estimated count

.. and large projects *need* the first one to be used by default, otherwise performance issues will be real headache. 

"Estimated count by default" is also no-good, since in loaded server it's better to avoid additional planner executions on *every* request. Here is a quick example for one real production server with 5k TPS, dozens of M of rows in the table:
```
XXXXXX=# explain select * from user where birthday <= '1980-01-01';
                                      QUERY PLAN
---------------------------------------------------------------------------------------
 Seq Scan on "user"  (cost=10000000000.00..10000273915.79 rows=1926654 width=405)
   Filter: (birthday <= '1980-01-01'::date)
(2 rows)

Time: 83.222 ms
XXXXXX=# explain select * from user where birthday <= '1980-01-01';
                                      QUERY PLAN
---------------------------------------------------------------------------------------
 Seq Scan on "user"  (cost=10000000000.00..10000273915.79 rows=1926654 width=405)
   Filter: (birthday <= '1980-01-01'::date)
(2 rows)

Time: 0.295 ms
XXXXXX=# explain select * from user where birthday > '2000-01-01';
                                     QUERY PLAN
-------------------------------------------------------------------------------------
 Seq Scan on "user"  (cost=10000000000.00..10000273915.79 rows=49839 width=405)
   Filter: (birthday > '2000-01-01'::date)
(2 rows)

Time: 24.530 ms
XXXXXX=# explain select * from user where birthday > '2000-01-01';
                                     QUERY PLAN
-------------------------------------------------------------------------------------
 Seq Scan on "user"  (cost=10000000000.00..10000273915.79 rows=49839 width=405)
   Filter: (birthday > '2000-01-01'::date)
(2 rows)

Time: 0.338 ms
```
– no one would be happy to have additional 20-80 ms (even though 2nd call is quicker, but it should be the same call and what if you have many possible filters?).

So I propose to disable count(1) by default. This will help to adopt this solution in mid-sized and big apps. The project itself is great and well designed and according my twitter, more and more people are interested in such solutions.

begriffs:
This is a backward incompatible change which computes total table size only when `Prefer: count=exact` is specified. By default postgrest will now get an efficiency boost by no longer performing count(1).

I have opened this PR as a WIP. There are a number of things to consider still

* When not counting we don't know the total number of rows hence do not know when to return status 206 rather than 200
* RPC endpoints still return an exact count. I have to see if this hurts performance and should be suppressed like the rest or whether it's information we can include for free. Also even if it's free to include should we hide the count for consistency when the Prefer header is absent?
* The PATCH verb also includes a count. Similar question to RPC: does it slow things down, should it be included without the Prefer header?
* The when ranges are requested and we do not count we are unable to error with 416 when the user requests a range starting beyond the last element because we don't know what the last element is.

Closes #273


ruslantalpa:
i would do it like this.
- always return 200 (unless it's count and we know the total)
- RPC, i think the cost is free, but still hide it so it's the same as get
- PATCH, i think the cost is free, but still hide it so it's the same as get
- return 200 with empty array

begriffs:
Currently POST doesn't return a count even if you supply `Prefer: count=exact` so I'll add that, but not tonight.

begriffs:
I think all the verbs deal with count correctly now. Ready for a last review on this PR.

ruslantalpa:
looks good

Issue #567, opened 2016-04-20
~~Clarify supported algorithms for jwt~~  Adopt support for more jwt algorithms


drawks:
I can find no reference in the docs as to what algos are implemented/supported for jwt verification. All references in docs and examples and the default value are plain ascii strings, which makes sense for symmetric schemes such as HS256, but there is no mention of support for asymmetric schemes such as RS256/512 or ES265/512.

I'm not quiet haskell literate enough to determine if these are supported and if they are what encoding should be used for the public key when used as a secret for postgrest.

drawks:
After digging through the code I found my answer. The JWT library in use here appears to be **the only** jwt library in existence that has only implemented support for HS256.

Seems like it would be a good idea to switch to the much more well featured https://github.com/frasertweedale/hs-jose which supports the full set of algorithms as defined in the JWT spec. 

Not being able to use asymmetric key pairs really reduces any security gains you get from having process separation of authentication and authorization.

begriffs:
This will be the next feature I tackle because I think it will also fix #495. Given how things go I'm guessing this will be complete by the middle of next week.

begriffs:
OK, looking more at the `jose` library I see that using it will cause a big change in how the JWT "secret" (aka `KeyMaterial`) must be specified. The `--secret` flag would probably become JSON so that these various `FromJSON` [parsers](https://hackage.haskell.org/package/jose-0.4.0.0/docs/src/Crypto-JOSE-JWA-JWK.html) could work on them. Each type of key requires different parameters inside the JSON.

As a first step I'm upgrading the jwt library we use right now to fix a bug related to secrets containing gnarly binary values. When that is fixed I can investigate this issue again. Keep in mind that supporting all these types of keys looks like a backward incompatible change for the command line arguments of the program.

drawks:
There should only be one other parameter besides "secret" that would need to be changed and that is the "algo" which is currently hard coded to "HS256" I'm a little confused what the "different parameters" are that you are talking about. The command line args shouldn't need to change in any backwards incompatible way. You simple add an optional "--algo" flag that defaults to "HS256" which is the current behavior.

begriffs:
The elliptic curve algo requires parameters `kty` `x`, `y`, and `d`. The RSA algo requires parameters `kty`, `n`, `e` and optionally `d`. The symmetric key algo requires `kty` and `k`.

[...]

Issue #505, opened 2016-02-24
Accept more than one JWT secret for authentication


eric-brechemier:
In order to allow a smooth transition from one JWT secret to the next, without invalidating all existing authentication tokens at once, more than one JWT secret should be tried for authentication, typically:
- the JWT secret currently used for the signature of tokens
- the JWT secret used previously for the signature of tokens

In other words, the configuration should allow at least two JWT secrets, one for signature (and authentication), and one or more just for authentication.

begriffs:
Makes sense, I can add that.

begriffs:
Actually looking at this again is the complexity worth it? What if you just invalidated existing tokens? Then clients would request a new token and continue.

eric-brechemier:
> Actually looking at this again is the complexity worth it?

Yes. There is no other way when using JWT for authentication (rather than password authentication over JWT).

Each JWT corresponds to a user role. Before we grant a new role to a user, with a new JWT, we need the current JWT to authenticate the user.

> What if you just invalidated existing tokens? 

Then existing users could only access the public API using the anonymous role.

> Then clients would request a new token and continue.

When doing so, they would simply become new users and would lose access to existing data.

ruslantalpa:
@eric-brechemier especially your last comment about "become new users" does not make sense, why calling `login` a second time would make the user lose his data?

as a general note, i don't see why you would need to transition between two secrets.
You only change the jwt secret when you suspect it has been compromised in which case all the existing tokens are invalidated and each client needs to login again.

eric-brechemier:
@ruslantalpa This is not merely `login`, as you assume. Generating a new secret is the equivalent of `signup` when one JWT = one identity.

> as a general note, i don't see why you would need to transition between two secrets.
You only change the jwt secret when you suspect it has been compromised (...)

We need to transition to a new secret in order to upgrade to newer versions of PostgREST at some point in the future, because the version that we are currently using has corrupted our secret (cf #495).

ruslantalpa:
I think you are using the jwt in a strange way.
Let's assume there is this feature you are asking for. Your jwt secret get's compromised and you need to change it?
What do you do, since you said you need to have the old and the new one working at the same time because switching directly to a new one will make the old clients lose their data, at the same time, if you leave the old one in place, it means the attacker can create fake (but valid) tokens as long as he likes.

[...]

PR #703, opened 2016-09-04
Big bag of JWT fixes for accumulated issues

begriffs:
Taking the opportunity to make some breaking changes for v0.4

This is a work in progress. It removes the default value of `"secret"` for JWTs, and responds with an error if someone tries authenticating against a postgrest server which has not been configured with a secret. The new code also adds the convention that a secret beginning with `@` means a filename from which to read the real secret. Reading from a file should allow unicode and binary secrets alike (but this requires testing). Reading from a file also removes sensitive information from the unix process list.

Two other JWT requested features not yet addressed are #505 and #567. Are these issues something we still want to fix?

I'll also be incorporating the custom validation discussed in #632...

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -7,7 +7,8 @@ module PostgREST.App (
 ) where
 
 import           Control.Applicative
-import qualified Data.ByteString.Char8   as BS
+import           Control.Monad             ((>>))
+import qualified Data.ByteString.Char8     as BS
 import           Data.IORef                (IORef, readIORef)
 import           Data.List                 (delete, lookup)
 import           Data.Maybe                (fromJust)
@@ -45,7 +46,7 @@ import           PostgREST.ApiRequest   (ApiRequest(..), ContentType(..)
                                             , ctToHeader
                                             , userApiRequest
                                             , toHeader)
-import           PostgREST.Auth            (tokenJWT, jwtClaims, containsRole)
+import           PostgREST.Auth            (jwtClaims, containsRole)
 import           PostgREST.Config          (AppConfig (..))
 import           PostgREST.DbStructure
 import           PostgREST.Error           (errResponse, pgErrResponse)
@@ -97,6 +98,8 @@ transactionMode _ = HT.Write
 
 app :: DbStructure -> AppConfig -> ApiRequest -> H.Transaction Response
 app dbStructure conf apiRequest =
+  exposeSecretToSQL (configJwtSecret conf) >>
+
   case (iAction apiRequest, iTarget apiRequest, iPayload apiRequest) of
 
     (ActionRead, TargetIdent qi, Nothing) ->
@@ -213,10 +216,6 @@ app dbStructure conf apiRequest =
      Just (PayloadJSON (UniformObjects payload))) -> do
         let p = V.head payload
             singular = iPreferSingular apiRequest
-            jwtSecret = secret <$> configJwtSecret conf
-            returnType = lookup (qiName qi) $ dbProcs dbStructure
-            returnsJWT = fromMaybe False $
-              isInfixOf "jwt_claims" . pdReturnType <$> returnType
         serves [CTApplicationJSON] (iAccepts apiRequest) $ \_ -> case readSqlParts of
           Left e -> return $ responseLBS status400 [jsonH] $ toS e
           Right (q,cq) -> respondToRange $ do
@@ -224,14 +223,7 @@ app dbStructure conf apiRequest =
             let (tableTotal, queryTotal, body) =
                   fromMaybe (Just 0, 0, emptyArray) row
                 (status, contentRange) = rangeHeader queryTotal tableTotal
-            return $ case (returnsJWT, jwtSecret) of
-              (True, Nothing) ->
-                errResponse status500 "Server lacks JWT secret"
-              (True, Just s) ->
-                responseLBS status [jsonH, contentRange] $
-                  "{\"token\":\"" <> toS (tokenJWT s body) <> "\"}"
-              (False, _) ->
-                responseLBS status [jsonH, contentRange] (toS . encode $ body)
+            return $ responseLBS status [jsonH, contentRange] (toS . encode $ body)
 
     (ActionRead, TargetRoot, Nothing) -> do
       let host = configHost conf
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -7,7 +7,6 @@ module PostgREST.App (
 ) where
 
 import           Control.Applicative
-import           Control.Monad             ((>>))
 import qualified Data.ByteString.Char8     as BS
 import           Data.IORef                (IORef, readIORef)
 import           Data.List                 (delete, lookup)
@@ -98,8 +97,6 @@ transactionMode _ = HT.Write
 
 app :: DbStructure -> AppConfig -> ApiRequest -> H.Transaction Response
 app dbStructure conf apiRequest =
-  exposeSecretToSQL (configJwtSecret conf) >>
-
   case (iAction apiRequest, iTarget apiRequest, iPayload apiRequest) of
 
     (ActionRead, TargetIdent qi, Nothing) ->
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -80,7 +80,8 @@ postgrest conf refDbStructure pool =
 
     let schema = toS $ configSchema conf
         apiRequest = userApiRequest schema req body
-        eClaims = jwtClaims (secret $ configJwtSecret conf) (iJWT apiRequest) time
+        eClaims = jwtClaims
+          (secret <$> configJwtSecret conf) (iJWT apiRequest) time
         authed = containsRole eClaims
         handleReq = runWithClaims conf eClaims (app dbStructure conf) apiRequest
         txMode = transactionMode $ iAction apiRequest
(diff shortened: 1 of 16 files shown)


src/PostgREST/App.hs
```diff
@@ -202,20 +204,24 @@ app dbStructure conf apiRequest =
...
+                  fromMaybe (Just 0, 0, emptyArray) row
                 (status, contentRange) = rangeHeader queryTotal tableTotal
-              in
-              return $ responseLBS status [jsonH, contentRange]
-                      (if returnsJWT
-                      then "{\"token\":\"" <> toS (tokenJWT jwtSecret body) <> "\"}"
-                      else toS $ encode body)
+            return $ case (returnsJWT, jwtSecret) of
+              (True, Nothing) ->
+                errResponse status500 "Server lacks JWT secret"
```
ruslantalpa: 500 seems quite bad, maybe find something less scary

src/PostgREST/App.hs
```diff
@@ -202,20 +204,24 @@ app dbStructure conf apiRequest =
...
+              (True, Just s) ->
+                responseLBS status [jsonH, contentRange] $
+                  "{\"token\":\"" <> toS (tokenJWT s body) <> "\"}"
```
ruslantalpa: i thought the idea was to completley remove the "jwt_claims" type support and ask ppl to use the jwt lib from sql

begriffs: You're right, I forgot about that. I'll just remove this code (yay!) and create another pull request to allow stored procs to customize the http status code and headers so that they can use https://github.com/michelp/pgjwt

ruslantalpa: besides this code, make sure to remove code from dbstructure that looks at the proc to see if it returns jwt.
4.0 can go out without the feature to customise return code since that would technically be a new feature so it can go in 4.1

src/PostgREST/App.hs
diogob: One alternative might be `422` which means: The request was well-formed but was unable to be followed due to semantic errors.
But I believe 500 is also appropriate in this case.

begriffs:
I'm concerned about storing the secret in a variable as well because it makes the worst case security failure of a proc much worse. It used to be that a proc could only do what the role of the current request allowed. Now it could accidentally leak the jwt secret and allow people to escalate their permissions. It's kind of dangerous. Also as you mention the secret appears in the db logs.

I don't think hardcoding the secret in the proc would be good either. The secret used for checking auth attempts has to match the hardcoded one. They could get out of sync, and it's slow to change the hardcoded one if necessary.

ruslantalpa:
Hardcoding is just an option.
Another option would be to have that secret in memory table, and the proc does a select from that table. Doing it like that, you can have the postgrest init script, before starting postgrest, saving the secret to that table.
Or just like here, it can be a variable, and it can be stored in a small sql file that you execute each time before starting postgrest.
I guess what i am saying is that doing it like it's now, does not give the user the flexibility in regards to security, having the secret from the pg param and the one in the db in sync is not a hard problem, its basically an additional command in the postgrest init script that takes the secret and sends it to the db.

diogob:
Another option is shipping `pgjwt` with our binary and upon postgrest initialization we (only when the secret option is passed):
* check for pgcrypto, if we can create the extension we also create the `pgjwt` with an extra procedure that postgrest will use to generate JWT.
* if we cannot create pgcrypto we display an error message saying that the functionality to generate tokens is disabled, so the secret is only being used to validate tokens.

Advantages of this approach:
* We have full control over the JWT workflow, since this seems to be a sensitive issue this might avoid a lot of user support.
* We know exactly which procedure is being used to create JWTs so we can hardcode its name in the haskell code so the secret gets passed without the user having to pass in the HTTP request.

diogob:
Actually the simplest approach possible might be to keep the magic type `jwt_token` or something to that effect to signal that the function receives the secret as a parameter. But I still like the idea of shipping the `pgjwt`, I feel that it gives a better out of the box experience and, therefore, less user support will be necessary.

ruslantalpa:
@diogob the "shipping pgjwt" approach i think i better handled by a "starter kit/sample" then bundling the lib with the binary. In the next few days i am thinking of writing a "thinking in PostgREST" tutorial and with that i will maybe create a "starter kit" repo. The current "users" example i think is unnecessarily complex.
For the purpose of this pr, i still think the best approach is to not send from Postgrest to the db the secret in any way, and have the user implement additional scripts (but we provide samples on how to do that) that will send the same secret both to postgrest command line and to the db env

begriffs:
Logically there are a few alternatives for generating JWT:

* in SQL
  * The secret can be in plaintext somewhere for SQL to access (either as a GUC or the fancy in-memory thing)
  * Or there could be specially initialized security definer function designed to generate JWT which internally has privileges to read the secret
  * Or there could be no communication of the secret from postgrest, but implicit knowledge in the person writing the SQL
* in Haskell
  * Using the return type naming convention as before
* in a proxy
  * encode JWT for routes chosen by the user - could be made into an nginx plugin or something

Looking back on it the approach of creating JWT in Haskell had a number of good properties. It is pretty simple, doesn't require installing an extension like pgcrypto or a special sql helper library. It keeps the secret hidden from the db, and updates the secret immediately if the server is restarted with a new one. The two drawbacks things are that inspecting the name of a return type is freaky and magical, and being able to generate JWT in SQL is more flexible in terms of choosing the signing algorithm etc.

👍 to having a simpler example of auth than what's in the guide right now. But requiring a starter kit just to enable the feature isn't as smooth as a magic return type. Are there more arguments against doing it in Haskell that I didn't think of?

ruslantalpa:
the problem with doing it in haskell is that that magical function can return only a json like `{token: ...}` and usually ppl want more (user id, user name ...)
doing it in proxy has the same problem as in sql, you need somehow to feed the same secret to postgrest and to the proxy (both reading from a file)
the good part about having postgrest not handle jwt generation at all (not even sending the secret) is that it leaves room for flexibility. Some might want to generate in db. some might want to generate in proxy and both cases are possible if you leave that door open.

about the starter kit, if you leave the jwt in the state it is now, it's not like everything just magically works because of the ability to encode jwt in haskell, ppl still start with "users" example ... that is already a starter kit ... might as well make it official (and simpler).
Another thing about jwt, it's not like jwt is absolutely required to have an auth system. you can do auth with sessions (my unfinished PR) in which case jwt does not come into play at all so that is another reason to externalise this feature (jwt generation)

pgcrypt is standart and is available everywhere so i don't think it's a problem.
The only thing that needs to be solved/decided for the starter kit that generates jwt in sql is a way to get the secret from a file to the db land and secure it somehow there. 
Let's see if this is possible then go from there

ruslantalpa:
Thinking a bit about this, a table holding the secret value seems to be a good solution (for the starter kit):
- login is a rare operation so it's not a big problem reading the secret from any kind of table
- access to the secret can be secured through usual mechanisms (only postgres user should have access to that table)
- postgrest startup script (which any production system should have and not start pg by hand) can be altered to send the same secret both postgrest and db (by doing an insert), for the starter kit, we can have a hardcoded `psql` command that reads the secret from a file and inserts it in the db table
- there is no magic involved, everything is in the open (sql) and the method can be completely removed/altered to the users liking

Thoughts?

eric-brechemier:
* Reading from a small table should never be a problem. And I agree that PostgreSQL permissions are the way to go to secure access to the table.

* I do not see the need for PostgREST server to have the JWT secret if the crypto is done by a procedure running inside the database.

* It would make sense to package a few SQL scripts with PostgREST at this point, that can be used as is or customized by advanced users.

ruslantalpa:
Postgrest still needs to validate the incoming request jwt ... hence it needs the secret
You usually package scripts that are required for a app to work, nothing from the "starter kit" is required, that is just a sample app, maybe your system does not even need auth, it jsut servers an open schema so it does not use at all anything in relation to jwt.

eric-brechemier:
> Postgrest still needs to validate the incoming request jwt ...

OK... but how is PostgREST going to validate a request signed using a custom algorithm? Isn't it one of the reasons for moving JWT generation to SQL? I am slightly confused.

ruslantalpa:
You are right, it wont be able to decode.
Th reasons are not to have more algorithms available.
The reason is to eliminate magic return type and to allow custom payload in reaponse for login functions (return more data then just the token)

begriffs:
How about this:

* Postgrest no longer sets `postgrest.jwt_secret`. Whoever runs the server and the db is responsible for synchronizing the secret.
* Postgrest decodes JWT signed with HMAC-SHA256 and no other algorithms
* There is no magic return type, procs have to use pgjwt
* Postgrest auth code will look for and execute a `validate_jwt_claims` if it exists for custom validation. (Should we require that this function live in a special schema to avoid it being visible as an external RPC call?)

ruslantalpa:
All except the past point. You are adding back a magic function and that function si very lomited in scope. It's better handled by a start param with a function name and that fn is executed on each request and it can be used for everything (including session based auth)

diogob:
Agreed, if the lack of secret sharing generates too many tickets we may revisit this idea in a future relesase.

begriffs:
I pushed the latest work to support calling a custom pre-request proc which can be used to conditionally raise auth errors. It's currently breaking tests -- we'll have to look at the db logs and see what's wrong.

ruslantalpa:
Look at my unfinished pr, it has this custom fn thing and tests for it

begriffs:
Sounds good, I'll check it out tomorrow.

begriffs:
OK I've decided that the open issues about having a backup secret and about validating multiple kinds of encryption are out of scope for this PR.

Can I get some final code review on the new `--pre-request` parameter and its tests? Do you all like the convention?

begriffs:
Oh there is one more little thing I'd like to add, using [auto-update](http://www.yesodweb.com/blog/2014/08/announcing-auto-update) to avoid redundant calls to `getPOSIXTime` when checking JWT expiration. Should improve performance a little.

begriffs:
Caching getPOSIXTime may be a premature optimization, but it probably saves a little time under heavy traffic. Internally it calls [clock_gettime](https://linux.die.net/man/3/clock_gettime).

src/PostgREST/App.hs
begriffs: The `5xx` errors are for things caused by the server or its configuration, and the `4xx` for request errors, so I believe this is the canonical case for a 500.

PR #705, opened 2016-09-08
Use table alias unlikely to conflict with a column name

### LINKED ISSUE #704: PostgREST cannot display tables with a column named t
Create a table with a column named `t`. Try to see it in the browser with http://localhost:3000/yourtable 

An error will show up
```
{"hint":"No function matches the given name and argument types. You might need to add explicit type casts.","details":null,"code":"42883","message":"function row_to_json(text) does not exist"}
```

Only happens when displaying json output. `t` is also the name of an internal variable used when building the SQL query sent to postgres, so it looks like a variable capture problem.
https://github.com/begriffs/postgrest/blob/6f737056a2a73698badffc9ca450c019100a7531/src/PostgREST/QueryBuilder.hs#L100

begriffs:
Fixes #704


ruslantalpa:
you're on a roll :)

PR #706, opened 2016-09-11
Include RPC endpoints in OpenAPI output

### LINKED ISSUE #694: Include RPC endpoints in self-generated API docs
I see: https://github.com/begriffs/postgrest/pull/639 - I see that is merged into master - I am running 3.2.0 - is there a release coming soon? If I did start using master (instead of what comes down from a ```brew install postgrest```), to get the endpoint data, do you just make a request with ```Accept: application/swagger+json``` or ```Accept: application/openapi+json```

I also see from the API reading doc page: "As well as issuing a GET / to obtain a list of the tables, views, and stored procedures available, you can get more information about any particular endpoint."

Should I be able to see functions like /rpc/login, /rpc/signup? (like described in the example auth - http://postgrest.com/examples/users/ ) I don't see those in a GET request to the root endpoint. For example, an "anon" user, making a GET request to the root endpoint, sees only the views they have been given permission to (anon does not have access to tables). They don't see anything related to any /rpc methods in that GET request, but can POST to those endpoints to sign up and log in. If I try to make an OPTIONS (any request other than a POST) request on /rpc/login, I get ```<Response body is empty>```. Is there something we need to do when creating the /rpc endpoints so this information get exposed?

Is there someway to get a list of all open/accessible endpoints and their available request methods? (and ideally what those endpoints expect for a successful request - body, headers, ?)

begriffs:
I still have to add more details to the OpenAPI response but the majority of the work is done. Thought I'd share the code at this stage for your feedback.

One big problem is there are no tests for what I've done. I've been testing the results by hand in [Swagger UI](https://github.com/swagger-api/swagger-ui). It's challenging to write the tests because we load a huge database fixture for all tests and it generates a big OpenAPI description. I would/will have to write tests that dig into that JSON structure.

The huge fixture has been a pain for testing for a long time. I would prefer to have each test create the necessary structure in a database transaction and then roll back afterwards. The big problem here is that PostgREST caches the db structure on start, and I'm not sure how to reload the server (or send it SIGHUP) on every test to refresh that cache.

Fixes #694 

Big thanks to @LogvinovLeon who helped me kick off this PR at our local SF Haskell hack night.


rkoberg:
Could you post here some of the responses you get back from a GET and OPTIONS request for the docroot? I could send you a ddl for a db with the authentication functions setup similar to the examples, permissions, as well as some crazy views.

begriffs:
@rkoberg thanks, I figured out how to use aeson-lens to pick apart the openapi response in a test. I'll add a few more tests and this PR should be good to go. We'll save the fixture refactor for later.

begriffs:
Able to parse more complicated proc args now. However it will fail for args with spaces or escaped quotes in their names. Any suggestions for a more robust way to parse this?

begriffs:
As discussed in Gitter room we'll just file an issue about arguments with names containing spaces or escape sequences. It's an edge case and we'll see if it causes any problems.

Tomorrow I'll probably finish this PR by adding OpenAPI information about the headers that RPC calls accept and any other relevant info.

begriffs:
Actually the headers reported by swagger need an update across the board. I'll do that in a separate PR and include headers for RPC there as well.

PR #711, opened 2016-10-04
Allow using nulls order without explicit order direction, fix #552

### LINKED ISSUE #552: Cannot order with "nullsfirst" without explicitly specifying an ascending order
In the docs for reading from the API, you have the example `GET /people?order=age.nullsfirst`. When I try to do this, I still get nulls last:

    CREATE TABLE foo (num integer)
    INSERT INTO foo values (1), (5), (2), (NULL), (6)

    GET "/foo?order=num.nullsfirst"
    "[{\"num\":1},{\"num\":2},{\"num\":5},{\"num\":6},{\"num\":null}]"

However, if I change the query to explicitly order ascending, I get the correct response:

    GET "/foo?order=num.asc.nullsfirst"
    "[{\"num\":null},{\"num\":1},{\"num\":2},{\"num\":5},{\"num\":6}]"

steve-chavez:
Fixes #552

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Do not apply limit to parent items - @ruslantalpa
 - Customize content negotiation per route - @begriffs
+- Allow using nulls order without explicit order direction - @steve-chavez
 
 ### Changed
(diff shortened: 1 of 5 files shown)


begriffs:
This is beautiful, tests, changelog entry and everything. Thank you!

(I also verified the new test fails on the master branch)

PR #712, opened 2016-10-05
fix relation detection bug and allow offset on lower levels

ruslantalpa:


begriffs:
Woo hoo, thanks for the fix!

One curious thing is that the modified test "limit works on all levels" passes even against the master branch. Any idea why?

ruslantalpa:
this pr adds "offests" not "limits", limits were added some time ago.
At the time i thought it made no sense to add offset to lower levels, but there is a use case, particularity when in the first level you have only one item (a post) fetched by id, and you also want to include the children (comments) but be able to paginate/load them progressively.

src/PostgREST/ApiRequest.hs
```diff
@@ -191,11 +191,19 @@ userApiRequest schema req reqBody =
     where lastWord = last $ T.split (=='.') key
 
   headerRange = if singular && method == "GET" then singletonRange 0 else rangeRequested hdrs
-  urlOffsetRange = rangeGeq . fromMaybe (0::Integer) $
-    readBSMaybe =<< join (lookup "offset" qParams)
-  urlRange = restrictRange
-    (readBSMaybe =<< join (lookup "limit" qParams))
-    urlOffsetRange
+  replaceLast x s = T.intercalate "." $ L.init (T.split (=='.') s) ++ [x]
```
ruslantalpa: then the line would get longer since i would need an additional fromMaybe, not sure

src/PostgREST/QueryBuilder.hs
```diff
@@ -339,7 +382,7 @@ requestToQuery schema isParent (DbRead (Node (Select colSelects tbls conditions
     --the following is just to remove the warning
     --getQueryParts is not total but requestToQuery is called only after addJoinConditions which ensures the only
     --posible relations are Child Parent Many
-    getQueryParts (Node (_,(_,Nothing,_)) _) _ = undefined
+    getQueryParts _ _ = undefined --error "undefined getQueryParts"
```
ruslantalpa: it's possible bit it's a lot of refactoring, a lot ...., all the core types would need to change :)

ruslantalpa: this specific case is never reached since this function does not get called when the previous code cant find a relation between to nodes, and this undefined part is here to match a node where the relation type is `Nothing`

ruslantalpa: but sure, a global refactoring is needed sometime in the future

src/PostgREST/ApiRequest.hs
begriffs: Actually initSafe is a convenience method that doesn't use Maybes. It just returns `[]` for `[]`.

ruslantalpa: oh, in that case ok

src/PostgREST/QueryBuilder.hs
begriffs: OK, as long as you feel the current setup is safe we can wait on the refactor.

src/PostgREST/ApiRequest.hs
ruslantalpa: On the second thought, if for some reason there is a situation where the array is empty and `L.init` blows, isn't int better to have it like that (the logs would show and undefined error) then quietly hiding the error and making it hard to detect the logical error (empty array) in the first place?

ruslantalpa: the server would not die, just the current request

begriffs: Actually I think we're safe here either way because split always returns a list with at least one item, for instance `split (=='.') ""` returns `[""]`. In this case replaceLast will just concatenate a dot and another value to the original string.

ruslantalpa:
Anything more for this to be merged?

begriffs:
We can merge it.

I keep going back to review the QueryBuilder module and I get overwhelmed by it. I'm fairly certain there are still bugs in there since we keep finding them and the code is complicated. However a big refactor prior to 0.4 will take a long time so I'm OK postponing it. The new release will be about as buggy as 0.3 in this respect but it's been tolerable.

As I mentioned in the chat my priority is release -> much improved docs -> refactor -> new features.

PR #714, opened 2016-10-06
Read settings from config file

### LINKED ISSUE #474: Read from config file to hide sensitive info from process list
The method suggested in the documentation, [Hiding Password from Process List](http://postgrest.com/install/server/#running-the-server) by setting an environment variable, leaves the value visible in `ps` listing, since the parameter `$PASS` is expanded and replaced with its value before running the command.

```sh
JWT_SECRET='abc123'
postgrest postgres://postgrest@localhost:5432/db (...) --jwt-secret "$JWT_SECRET"
ps aux | grep jwt-secret
```

```
(...) postgrest postgres://postgrest@localhost:5432/egull (...) --jwt-secret abc123
```

begriffs:
Fixes #474

Removes all the command line arguments except an argument `-c` to override path to configuration file. The default config file path is `~/.config/postgrest/conf` or `$XDG_CONFIG_HOME/postgrest/config` if that environment variable is set.

As of this first commit the PR is somewhat lacking because it doesn't provide help text that guides the user to the expected format of the config file.

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -26,6 +26,7 @@ import           Control.Applicative
 import qualified Data.ByteString.Char8       as BS
 import qualified Data.CaseInsensitive        as CI
 import qualified Data.Configurator           as C
+import qualified Data.Configurator.Types     as C
 import           Data.List                   (lookup)
 import           Data.Text                   (strip, intercalate)
 import           Data.Text.IO                (hPutStrLn)
@@ -124,23 +125,24 @@ readOptions = do
     (C.load [C.Required $ caConfig args])
     configNotfoundHint
 
-  -- db ----------------
-  cDbUri    <- C.require conf "db-uri"
-  cDbSchema <- C.require conf "db-schema"
-  cDbAnon   <- C.require conf "db-anon-role"
-  cPool     <- C.lookupDefault 10 conf "db-pool"
-  -- server ------------
-  cHost     <- C.lookupDefault "*4" conf "server-host"
-  cPort     <- C.lookupDefault 3000 conf "server-port"
-  cProxy    <- C.lookup conf "server-proxy-uri"
-  -- jwt ---------------
-  cJwtSec   <- C.lookup conf "jwt-secret"
-  -- safety ------------
-  cMaxRows  <- C.lookup conf "max-rows"
-  cReqCheck <- C.lookup conf "pre-request"
+  handle missingKeyHint $ do
+    -- db ----------------
+    cDbUri    <- C.require conf "db-uri"
+    cDbSchema <- C.require conf "db-schema"
+    cDbAnon   <- C.require conf "db-anon-role"
+    cPool     <- C.lookupDefault 10 conf "db-pool"
+    -- server ------------
+    cHost     <- C.lookupDefault "*4" conf "server-host"
+    cPort     <- C.lookupDefault 3000 conf "server-port"
+    cProxy    <- C.lookup conf "server-proxy-uri"
+    -- jwt ---------------
+    cJwtSec   <- C.lookup conf "jwt-secret"
+    -- safety ------------
+    cMaxRows  <- C.lookup conf "max-rows"
+    cReqCheck <- C.lookup conf "pre-request"
 
-  return $ AppConfig cDbUri cDbAnon cProxy cDbSchema cHost cPort
-        cJwtSec cPool cMaxRows cReqCheck False
+    return $ AppConfig cDbUri cDbAnon cProxy cDbSchema cHost cPort
+          cJwtSec cPool cMaxRows cReqCheck False
 
  where
   opts = info (helper <*> argParser) $
@@ -158,6 +160,13 @@ readOptions = do
       "Cannot open config file:",
       "\t" <> show e,
       "\nUse the --help flag to learn how to fix this."]
+    exitFailure
+
+  missingKeyHint :: C.KeyError -> IO a
+  missingKeyHint (C.KeyError n) = do
+    hPutStrLn stderr $
+      "Required config parameter \"" <> n <> "\" is missing or of wrong type.\n" <>
+      "Try the --example-config option to see how to configure PostgREST."
     exitFailure
 
 data CmdArgs = CmdArgs {
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
+
 {-|
 Module      : PostgREST.Config
 Description : Manages PostgREST configuration options.
@@ -31,7 +33,6 @@ import           Options.Applicative
 import           Paths_postgrest             (version)
 import           Protolude hiding            (intercalate
                                              , (<>))
-import           System.Environment.XDG.BaseDir
 
 -- | Config file settings for the server
 data AppConfig = AppConfig {
@@ -75,13 +76,21 @@ corsPolicy req = case lookup "origin" headers of
 prettyVersion :: Text
 prettyVersion = intercalate "." $ map show $ versionBranch version
 
+-- | Default OS-specific path to config file
+defConf :: FilePath
+defConf =
+#ifdef mingw32_HOST_OS
+  "%PROGRAMDATA%\\postgrest\\postgrest.conf"
+#else
+  "/etc/postgrest.conf"
+#endif
+
 -- | Function to read and parse options from the command line
 readOptions :: IO AppConfig
 readOptions = do
   confArg <- customExecParser parserPrefs opts
-  defConf <- getUserConfigFile "postgrest" "config"
 
-  conf <- C.load [C.Required $ fromMaybe defConf confArg]
+  conf <- C.load [C.Required confArg]
   -- db ----------------
   cDbUri    <- C.require conf "db.uri"
   cDbSchema <- C.require conf "db.schema"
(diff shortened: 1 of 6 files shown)


ruslantalpa:
my question is why the XDG thing and why the ~/ thing?
This placement of config files is typical of user applications, not services.
usually on linux systems every config is under /etc so why don have the path like /etc/postgrest.conf then look for similar paths for other OS.

maybe have the logic like, look in /etc, if not there, then look in the directory where the postgrest binary is located.

statik:
I'm using postgrest via the docker container, would this mean I need to build docker containers with the configuration file baked in?

ruslantalpa:
@statik you either bake it in or your start the container with a `-v` and feed it your config file which lives outside the continer

begriffs:
@ruslantalpa good point about XDG. I always run postgrest under my own own user directory but the normal use case is running postgrest as global system service. Using a global config path like `/etc/postgrest.conf` is the way to go. I'll investigate how to use the right global config path per OS by default.

begriffs:
The basic value reading works now. There are a few more concerns:

* Auto-reloads. The `configurator` library allows us to run code when the config file changes, but I'm not sure how to handle it. Different changes require different actions. Changing the port requires relaunching the web server, changing db parameters require flushing out the connection pool. Altering the jwt secret or anon role would probably be easier to deal with. I propose we don't support config auto reload in 0.4, and save it for some other *0.4.x*
* Broken Docker and Heroku config. For docker we'll want an argument to copy in a config file from outside the container. The config file format supported by configurator allows one file to import another. So the config file inside the container should import the one that was copied in and override some things:

   ```
   import "$(HOME)/imported.conf"

    # override these params for safety to match Dockerfile
    server {
      host = "*4"
      port = 3000
    }
   ```
The [heroku buildpack](https://github.com/begriffs/postgrest-heroku) will have to create a config file if it detects a requested version of 0.4 or above.
* Showing the user the possible config file options. The `--help` screen no longer shows program options since the only option is `-c`. As it is I can imagine someone getting completely lost understanding how to make a config file. Here are ways we could deal with it:
    1. Add an example config file to the help screen
    1. Bundle an example config file in the release tarball
    1. Add a URL to the help screen output pointing to a sample config
    1. Add another command line option which will generate a config

begriffs:
Here's how to use the updated Dockerfile:

```bash
docker run -v my.conf:/etc/postgrest.conf begriffs/postgrest
```

Hard to test though because the Dockerfile downloads releases from github rather than building from source, and the 0.4 release is obviously not available yet for download.

ruslantalpa:
For the config file option i saw you went with (i)
I would suggest/vote for (II)+(III) at the same time

- Bundle an example config file in the release tarball
- Add a URL to the help screen output pointing to a sample config (right now maybe just a link to the main site until the docs are ready)

Here is a scenario of how the whole process would go:
- download the tar file containing bin + conf
- if you start directly, the exe should look for /etc/postgrest.conf (not found) then look at ./postgrest.conf (which exists)
- on startup it should say in logs (starting with config from /..../postgrest.conf)
- obviously it should fail right from the start and say (could not connect to the db)
- the config file should have a comment right at the beginning saying (this file should be placed in /etc/postgrest.conf)
- the user would then move the file there and make a few changes (db config)
- if the user makes a mistake, when loading the config file, it should say the exact line where the error is (i think configurator supports this)
- the help switch should only display an info about the -c param and say "path to the config file, for a sample look in the release tarball or this url. The default path is /etc/ then ./ when -c is not provided"

docker file is good enough, no need to worry to much about it, it's very easy to create your own if you have any problem with the default way (config file), you can use old way of passing env vars to the container and have an "entrypoint" that turns those to a config file on startup

ruslantalpa:
One more thing to discuss/think about is the 
`db {} server {}` groupings, what is the benefit of this grouping, can't this be implemented with comments and empty new lines like
```
# DB
host = ...
anon = ...

# SERVER
```
eliminating this grouping makes it easier to programatically alter a config file (grep/sed/awk) and the order of the config statements would not matter.

for example in a docker situation, if you have this behaviour + the last statement overwrites the one above then you could just do
```
echo "param_i_want_changed" >> /etc/postgrest.conf
```
and things should work

begriffs:
Good logic about the config groups, I'll remove them.

itmitica:
I would second the possibility of a config file in the `~/.config/postgrest/` folder to read settings from.
The `/etc/postgrest.config` is global, not every user will want the same settings, will they?
For example, pgAdmin uses the home folder: `~/.config/pgAdmin Development Team`.

ruslantalpa:
Postgrest is not a user faceing software laike pgadmin, its a service. Services are 99% of the time are a system wide thing. The only time you will have multiple instances running on the same box is in a dev env, and you can use -c for that to specify a custom location

itmitica:
I do not have to root the postgrest instance but I have to root my config file? Just saying.
Also, having the config file in the same directory as the binary or in the /etc is a mix between Windows and GNU/Linux, it's not consistent.

begriffs:
@itmitica I had the same initial feeling as you, but then I started thinking of other server software like Apache or Nginx and I think they tend to use global configuration by default.

Also this pull request includes a `-c` option to load the configuration file from a custom place if you don't have access to `/etc` or don't want to store the config globally.

```bash
postgrest -c ~/.config/postgrest/conf
```

ruslantalpa:
@begriffs maybe it would be a good idea to have the -c parameter mandatory and have the default path be decided by the startup script?

itmitica:
I, for example, have put the postgrest binary in the PostgreSQL bin folder: `/opt/PostgreSQL/9.6/bin`. The `/etc` path for the config file would probably make sense in this case.

I might want to test the unreleased version. In this case, looking for a config file in the `~/.config` directory first and then look for the global config would make the most sense.

The use cases above align with the convention-over-configuration scenario.


Having a `-c` option aligns with the configuration-over-convention scenario. In this case, regardless where the unreleased binary is moved to, carrying the config file with it would most likely contribute to the file clutter.


Should both scenarios be covered? As @begriffs says, should the configuration-over-convention be the only one?


It's hard to resist having an easy way of controlling the place for the config file, but I'd argue ditching the `-c` option is probably best. I'd argue this option belongs to testing versions in binaries, not in development or production.

PierreR:
To broad the discussion would you also consider package manager like 'nix' ? Assuming any global relevance of a global /etc folder is an anti-pattern in such setting.

begriffs:
@itmitica it's currently a mix of both ways. By default it will look for `/etc/postgrest.conf` and the optional `-c` argument allows overrides. (The idea is "support a convention but allow configuration" I guess.)

The question in my mind is what is the best default? I can imagine somebody downloading postgrest and experimenting with it on their development machine from the home directory, and storing the config file there too. However the "major" case where the default location applies would be running postgrest on a production server with a global config. I don't have a strong opinion here, but that seems plausible.

@PierreR what is the best practice for Nix-compatible application configuration files?

itmitica:
@begriffs
the `/etc/postgrest.conf` seems good. `/etc/postgrest/postgrest.conf` to me looks better.

For some reason `-c` seems to me half backed solution for a problem that doesn't exist: just fall back to `~/.config/postgrest/postgrest.conf` for overrides, as everyone expects, outside the "major" case scenario.

Maybe I'm overthinking this. I'm sure `-c` it's a nice feature to have for many while I'm taking time to over analyse. Sorry.

ruslantalpa:
There is a db, this db has 3 schemas, `api_v1`, `api_v2`, `api_v3` since the developer has to support 3 different versions (or maybe 2 separate apis for different audiences)
then he has the box where postgrest lives and he has to start 3 different instances that each connect to different schemas.
@itmitica  Please tell me how he can do that if the config path is hardcoded in the binary and how -c parameters is a "half baked solution for a problem that does not exist"

itmitica:
@ruslantalpa If the separation takes place at the db level, then keep there: build a forth db schema to handle the three and keep a single postgrest instance running pointing at the forth db schema.

Question: Having multiple postgrest instances running, how would you know which instance to kill for a specific api_?

Question: Having multiple postgrest instances running, the load balance for these instances is handled by ... ?

src/PostgREST/Config.hs
```diff
@@ -88,18 +81,105 @@ corsPolicy req = case lookup "origin" headers of
...
-                    <> progDesc (
-                    "PostgREST "
-                    <> toS prettyVersion
-                    <> " / create a REST API to an existing Postgres database"
-                    )
-    parserPrefs = prefs showHelpOnError
+readOptions = do
+  args <- customExecParser parserPrefs opts
+
+  when (caExample args) $ do
```
ruslantalpa: what might be useful is a `--check-config` option if `configurator` allows adding such functionality easily

begriffs: Configurator can check whether a mandatory key is missing from the config or whether a value has the wrong type. It doesn't say which line is wrong but the key name should be enough information to debug the issue. For instance when I use an integer value for `db-anon-role` it warns me:

```
$ stack exec -- postgrest -c test.conf
Required config parameter "db-anon-role" is missing or of wrong type.
Try the --example-config option to see how to configure PostgREST.
```

ruslantalpa: i guess it's good enough and it's cool that it works each time you load the config.
So maybe have the binary just document the `-c` option and also mention the paths it tries to look for the config file, and maybe eventually a link to docs/sample

begriffs:
Given what we discussed, I've adjusted the code to make the `-c` argument mandatory, no default location.

* There doesn't appear to be consensus about the default config file location - system folder vs home vs XDG vs Nix
* There are times when you need multiple postgrest servers running at the same time with different parameters. Most notably servers pointing at different db schemas and running on different ports (to be routed by nginx for API versioning)

PR #715, opened 2016-10-08
Add port in environment

pimuzzo:
At the moment I have 2 databases on the same postgres, but postgrest needs to specify the `PG_ENV_POSTGRES_DB` to work.
An alternative could be without `PG_ENV_POSTGRES_DB` changing url like /database/table instead of /table

Thank you

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ ENV POSTGREST_ANONYMOUS postgres
 ENV POSTGREST_JWT_SECRET thisisnotarealsecret
 ENV POSTGREST_MAX_ROWS 1000000
 ENV POSTGREST_POOL 200
+ENV POSTGREST_PORT 3000
 
 RUN apt-get update && \
     apt-get install -y tar xz-utils wget libpq-dev && \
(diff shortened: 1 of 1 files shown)


begriffs:
Since a single postgrest server runs in each instance of the docker container couldn't you start two instances? If I'm understanding it right, the port with value 3000 is an internal detail of the container that the outside world shouldn't need to know about.

pimuzzo:
At the moment I'm running postgres and two postgrest in the same docker compose, so they share the network and only exit through nginx

begriffs:
What does your docker compose file look like? I think you can alias ports in it like `HOST:CONTAINER`.

https://docs.docker.com/compose/compose-file/#/ports

pimuzzo:
I don't want to expose nothing except nginx to the host :(

my compose is something like that:
```
version: '2'

services:
  postgres:
    restart: always
    container_name: postgres
    image: postgres:9.5
    environment:
      POSTGRES_PASSWORD: xxx
      POSTGRES_DB: xxx
    volumes:
      - "./pg_data:/var/lib/postgresql/data"

  nginx:
    restart: always
    container_name: nginx
    image: nginx
    volumes:
      - "./nginx.conf:/etc/nginx/nginx.conf:ro"
    ports:
      - "80:80"

  postgrest_1:
    restart: always
    container_name: postgrest_1
    image: pimuzzo/postgrest
    environment:
      PG_ENV_POSTGRES_USER: xxx
      PG_ENV_POSTGRES_PASSWORD: xxx
      PG_PORT_5432_TCP_ADDR: postgres
      PG_PORT_5432_TCP_PORT: 5432
      PG_ENV_POSTGRES_DB: blabla1
      POSTGREST_PORT: 3000

  postgrest_2:
    restart: always
    container_name: postgrest_2
    image: pimuzzo/postgrest
    environment:
      PG_ENV_POSTGRES_USER: xxx
      PG_ENV_POSTGRES_PASSWORD: xxx
      PG_PORT_5432_TCP_ADDR: postgres
      PG_PORT_5432_TCP_PORT: 5432
      PG_ENV_POSTGRES_DB: blabla2
      POSTGREST_PORT: 3001
```

If both postgrest use port 3000 one of them doesn't work

begriffs:
I see what you mean about not opening those ports to the host. I'm unfamiliar with docker-compose [networking](https://docs.docker.com/compose/networking/), but as I read about it more it appears that `postgrest_1` and `postgrest_2` will be different hostnames for those containers. Maybe nginx could reference `postgrest_1:3000`/`postgrest_2:3000` rather than `localhost:3000`/`localhost:3001`.

Let me know if that works.

ruslantalpa:
@begriffs you are right, they are separate hosts on the same network, there is no conflict

pimuzzo:
You right, I did that try before, but maybe I done a typo or another error because now it's working all perfectly :D Thank you for the time spent

PR #720, opened 2016-10-18
Set request.jwt.claim.*

### LINKED ISSUE #674: Consider adopting `PostgreSQL JSON Web Token Serialization Specification`
The [PostGraphQL project](https://github.com/calebmer/postgraphql) has proposed a standardized spec for handling JWT's in Postgresql API servers:

[PostgreSQL JSON Web Token Serialization Specification](https://github.com/calebmer/postgraphql/blob/master/docs/pg-jwt-spec.md)

As Postgrest is the most mature postgresql API server, I think it would benefit the greater community to consider adopting this spec. The spec is itself based off (but a little difference naming conventions) the existing postgrest implementation.

The big win for end-users would be the ability to expose both a REST API *and* a graphQL API from the same postgresql DB -using a single authentication mechanism.

begriffs:
Fixes #674 

Saving `request.cookie` and `request.header` for another day after the 0.4 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - No more reserved `jwt_claims` return type - @begriffs
 - HTTP 401 rather than 400 for expired JWT - @begriffs
 - Remove default JWT secret - @begriffs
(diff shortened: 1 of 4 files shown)


ruslantalpa:
The important part is the consistency (and path forward)
Merge!

PR #721, opened 2016-10-18
Remove things to lighten support burden

begriffs:
Just trying to simplify here.

Removes:

* Blogging demo sql. This should be maintained in docs not here.
* Debian packaging scripts. The PostgREST distribution is a single binary file, very easy to install. Hope to keep it that way!
* Old docs placeholder that points at docs repo -- postgrest.com now points at docs.


ruslantalpa:
One thing from the deleted files is valuable and should be moved somewhere, init script (although it needs some work).
It's a non trivial thing to properly install postgrest as a system service (and not just a binary you manually start)

begriffs:
Good point. We can research and document ways to make the server a linux service.

PR #722, opened 2016-10-19
Bring packages up to date

begriffs:


postgrest.cabal
```diff
@@ -29,17 +29,17 @@ executable postgrest
     -rtsopts
     "-with-rtsopts=-N -I2"
   default-language:    Haskell2010
-  build-depends:       aeson (>= 0.8 && < 0.10) || (>= 0.11 && < 0.12)
+  build-depends:       aeson (>= 0.11 && < 0.12) || (>= 1.0 && < 1.1)
```
diogob: Why not:
`aeson (>= 0.8 && < 0.12)` ?

stack.yaml
```diff
@@ -1,12 +1,12 @@
-resolver: lts-6.7
+resolver: lts-7.4
 extra-deps:
   - Ranged-sets-0.3.0
-  - bytestring-tree-builder-0.2.7
-  - hasql-0.19.12
+  - bytestring-tree-builder-0.2.7.1
```
diogob: we don't need this line, bytestring-tree-builder-0.2.7.1 is included in LTS 7.4
Other packages that could be removed from extra-deps because they are included are:
* hasql-0.19.15.2
* jwt-0.7.2
* postgresql-binary-0.9.1.1
* binary-parser-0.5.2
* placeholders-0.1

postgrest.cabal
```diff
@@ -29,17 +29,17 @@ executable postgrest
...
                      , bytestring
-                     , bytestring-tree-builder == 0.2.7
+                     , bytestring-tree-builder == 0.2.7.1
                      , case-insensitive
                      , cassava
                      , containers
                      , contravariant
                      , errors
-                     , hasql == 0.19.12
+                     , hasql == 0.19.15.2
```
diogob: wouldn't it be nicer `hasql  >= 0.19.15 && < 0.19.16` ?

begriffs: I think I made the direct match so that `packdeps` would warn me when hasql has been upgraded beyond the strict bounds. But I guess with your proposal it still would warn me and it would be a more meaningful version change. So good idea.

postgrest.cabal
begriffs: The 0.10 series was actually known to be buggy so I'd like to skip it. But I could do something like

```
aeson (>= 0.8 && < 0.10) || (>= 0.11 && < 1.1)
```

stack.yaml
begriffs: Oh excellent!

begriffs:
What do you think of this? I removed version constraints from most packages, trusting Stackage to pick compatible versions. The exception is strict version equality for packages which are also specified in the extra-deps section of stack.yaml. This way packdeps will complain on CI as soon as any extra-dep goes out of date (otherwise nothing would warn us). Also most packages are in the Stackage LTS now so I could clean out extra-deps (which will help us stay automatically up to date as we upgrade the LTS).

My reasoning to remove the version constraints from the cabal file is that postgrest is not a library that other packages usually include. The main thing we're making is an executable. However our CI does include

```bash
stack exec --no-ghc-package-path -- cabal install --only-d --dry-run
```

to check that old fashioned cabal-install is able to solve for package versions without Stack's help. This may help ensure that other packages will be able to include postgrest.

diogob:
Makes sense to me. And it's easy to revert in the future if we find out that is better to have the constraints. 👍

PR #723, opened 2016-10-23
Update readme

begriffs:
Misc updates to be more accurate for v4

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,6 @@ Finally it uses the database efficiently with the
 [Hasql](https://nikita-volkov.github.io/hasql-benchmarks/) library
 by
 
-* Reusing prepared statements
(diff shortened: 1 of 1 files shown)


ruslantalpa:
I would remove the "Future Features" section, almost all the things listed there don't seem to be on the radar in the near future and some of them (caching) i would (have done already) do in another layer (nginx, easyer and more flexible)

ruslantalpa:
and also the line about prepared statements, it wont work like that after fe change `True` to `False` in the main query

begriffs:
Sure I can remove the future features since much of that info is captured in the issues.

Also I hope at some point after we disable prepared statements for the memory problem that we can re-enable them and use postgres query params.

PR #726, opened 2016-10-26
add nix.packages to stack.yaml

phanimahesh:
postgrest depends on postgres and zlib. When using stack's nix
integration, these dependencies must be specified explicitly.

Note that `--nix` must be passed to stack, or `nix.enable:true` must
be set in `stack.yaml` to enable using nix while building.

---

This change is [Reviewable](https://reviewable.io/reviews/begriffs/postgrest/726)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -5,3 +5,5 @@ extra-deps:
   - hasql-transaction-0.4.5.1
 ghc-options:
   postgrest: -O2 -Werror -Wall -fwarn-identities
+nix:
+  packages: [postgresql, zlib]


ruslantalpa:
does `postgresql` mean pull in the entire postgresql server?
is there a package something like `libpq` or `libpq-dev` because PostgREST does not need the actual database to work or compile?
I see there is something like `libpqxx`, can that work?

phanimahesh:
yes, postgresql will pull in the server and client. I'll try to narrow it down.

phanimahesh:
`postgresql-libpq` requires `pg_config` while building, and `libpqxx` does not provide it.

`nix.packages` only specifies the build dependencies. When creating a nix derivation runtime dependencies can be limited to a subset of this.

begriffs:
I wonder how hard it would be to create a new nix package just for libpq? [This comment](https://stackoverflow.com/questions/1244778/where-do-i-get-libpq-source) says if you download the postgres source and then build a certain folder that it'll create libpq alone.

phanimahesh:
Runtime dependency can be restricted to `postgresql.lib` nix package, but for building looks like we also need `postgresql.bin`, with `pg_config` in there.

begriffs:
Sounds good, we can refine the dependencies later if desired.

PR #728, opened 2016-10-29
Fix fatal error on postgres unsupported version, also fix #577 format…

### LINKED ISSUE #577: Misleading formatting of version number in error message "PostgREST needs at least 90300"
I was getting this error message when I started up PostgREST:

`Cannot run in this PostgreSQL version, PostgREST needs at least 90300`

Thinking that it meant version 9.0.3, I disregarded it because I am running 9.2 and went off the garden path looking for some other cause of the error. 

User ruslantalpa put me right, but it may help some other poor soul if `minimumPgVersion` could be better formatted to show the major and minor versions more sensibly.

steve-chavez:
… of supported version

I tested postgREST on a 9.2.19 postgres and it didn't panic and instead gave the following error:

```
"{\"hint\":\"No function matches the given name and argument types. You might need to add explicit type casts.\",\"details\":null,\"code\":\"42882\",\"message\":\"function pg_relation_is_updatable(regclass, boolean) does not exist\"}"
```

I did a `trace` of [ver](https://github.com/begriffs/postgrest/blob/master/main/Main.hs#L36) and it was giving a 808792116 value so it was a casting problem.

Also I thought of formatting 90300 into "9.3" but after seeing how libpq [does it](https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-exec.c#L978-L1011) I thought is better to just hardcode it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix bug in relation detection when selecting parents two levels up by using the name of the FK - @ruslantalpa
 - Customize content negotiation per route - @begriffs
 - Allow using nulls order without explicit order direction - @steve-chavez
+- Fatal error on postgres unsupported version, format supported version in error message - @steve-chavez
 
 ### Changed
 - Use HTTP 400 for raise\_exception - @begriffs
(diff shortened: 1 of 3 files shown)


begriffs:
That's subtle, thank you for the fix.

PR #732, opened 2016-10-30
Remove misleading comment

jhr007:


begriffs:
Thanks!

You'll also want to pass an argument to postgrest, like

```bash
postgrest -c /etc/postgrest/postgrest.conf
```
if you think that's the best path for the conf file.

begriffs:
Just to reiterate, the following will most definitely fail because postgrest needs a config file path.

```dockerfile
CMD exec postgrest
```

begriffs:
I'm closing this in light of the more comprehensive changes you're working on.

PR #733, opened 2016-10-31
Reduce app function

diogob:
This PR is a small refactor in preparation for a series of refactors with the goal of reducing and simplifying the big `app` function.

It extracts the content type negotiation to a separate function (`servesMatchingContentTypes `) that later could be entirely removed and coded in terms of a new type called `InvalidResponse` that would encompass several error cases closer to the domain level.

It also extracts some common cases that checked for valid db requests into the functions `servesMutateRequest` and `servesReadRequest'`. This reduces duplication and improves a bit the readability of the big case body in the `app` function.

As part of readability improvements some lines were moved to bring definitions closer to their respective usage.

There is also some naming improvements, `ctToHeader` is not `toHeader` and the old `toHeader` is `toMime` as these names seem more precise.

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -1,4 +1,12 @@
-module PostgREST.ApiRequest where
+module PostgREST.ApiRequest ( ApiRequest(..)
+                            , ContentType(..)
+                            , Action(..), Target(..)
+                            , PreferRepresentation (..)
+                            , mutuallyAgreeable
+                            , ctToHeader
+                            , userApiRequest
+                            , toHeader
+                            ) where
 
 import           Protolude
 
@@ -281,6 +289,3 @@ ensureUniform arr =
   if (V.length objs == V.length arr) && areKeysUniform
     then Just (UniformObjects objs)
     else Nothing
-
-readBSMaybe :: Read a => ByteString -> Maybe a
-readBSMaybe = readMaybe . toS
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -54,6 +54,7 @@ data Target = TargetIdent QualifiedIdentifier
             | TargetProc  QualifiedIdentifier
             | TargetRoot
             | TargetUnknown [Text]
+            deriving Eq
 -- | How to return the inserted data
 data PreferRepresentation = Full | HeadersOnly | None deriving Eq
                           --
@@ -120,9 +121,9 @@ userApiRequest schema req reqBody =
                else ActionInappropriate
           else
             case method of
-               "GET"     -> case target of
-                              TargetRoot -> ActionInspect
-                              _ -> ActionRead
+               "GET"     -> if target == TargetRoot
+                              then ActionInspect
+                              else ActionRead
                "POST"    -> ActionCreate
                "PATCH"   -> ActionUpdate
                "DELETE"  -> ActionDelete
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -1,11 +1,16 @@
+{-|
+Module      : PostgREST.ApiRequest
+Description : PostgREST functions to translate HTTP request to a domain type called ApiRequest.
+-}
 module PostgREST.ApiRequest ( ApiRequest(..)
                             , ContentType(..)
-                            , Action(..), Target(..)
+                            , Action(..)
+                            , Target(..)
                             , PreferRepresentation (..)
                             , mutuallyAgreeable
-                            , ctToHeader
-                            , userApiRequest
                             , toHeader
+                            , userApiRequest
+                            , toMime
                             ) where
 
 import           Protolude
@@ -41,6 +46,7 @@ type RequestBody = BL.ByteString
 data Action = ActionCreate | ActionRead
             | ActionUpdate | ActionDelete
             | ActionInfo   | ActionInvoke
+            | ActionInspect
             | ActionInappropriate
             deriving Eq
 -- | The target db object of a user action
@@ -56,16 +62,16 @@ data ContentType = CTApplicationJSON | CTTextCSV | CTOpenAPI
                  | CTAny | CTOther BS.ByteString deriving Eq
 
 -- | Convert from ContentType to a full HTTP Header
-ctToHeader :: ContentType -> Header
-ctToHeader ct = (hContentType, toHeader ct <> "; charset=utf-8")
+toHeader :: ContentType -> Header
+toHeader ct = (hContentType, toMime ct <> "; charset=utf-8")
 
 -- | Convert from ContentType to a ByteString representing the mime type
-toHeader :: ContentType -> ByteString
-toHeader CTApplicationJSON = "application/json"
-toHeader CTTextCSV         = "text/csv"
-toHeader CTOpenAPI         = "application/openapi+json"
-toHeader CTAny             = "*/*"
-toHeader (CTOther ct)      = ct
+toMime :: ContentType -> ByteString
+toMime CTApplicationJSON = "application/json"
+toMime CTTextCSV         = "text/csv"
+toMime CTOpenAPI         = "application/openapi+json"
+toMime CTAny             = "*/*"
+toMime (CTOther ct)      = ct
 
 {-|
   Describes what the user wants to do. This data type is a
@@ -114,7 +120,9 @@ userApiRequest schema req reqBody =
                else ActionInappropriate
           else
             case method of
-               "GET"     -> ActionRead
+               "GET"     -> case target of
+                                       TargetRoot -> ActionInspect
+                                       _ -> ActionRead
                "POST"    -> ActionCreate
                "PATCH"   -> ActionUpdate
                "DELETE"  -> ActionDelete
@@ -147,7 +155,7 @@ userApiRequest schema req reqBody =
                       . map (toS *** JSON.String . toS) . parseSimpleQuery
                       $ toS reqBody
         ct ->
-          PayloadParseError $ "Content-Type not acceptable: " <> toHeader ct
+          PayloadParseError $ "Content-Type not acceptable: " <> toMime ct
       relevantPayload = case action of
         ActionCreate -> Just payload
         ActionUpdate -> Just payload
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -121,8 +121,8 @@ userApiRequest schema req reqBody =
           else
             case method of
                "GET"     -> case target of
-                                       TargetRoot -> ActionInspect
-                                       _ -> ActionRead
+                              TargetRoot -> ActionInspect
+                              _ -> ActionRead
                "POST"    -> ActionCreate
                "PATCH"   -> ActionUpdate
                "DELETE"  -> ActionDelete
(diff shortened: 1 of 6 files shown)


src/PostgREST/ApiRequest.hs
```diff
@@ -104,7 +120,9 @@ userApiRequest schema req reqBody =
                else ActionInappropriate
           else
             case method of
-               "GET"     -> ActionRead
+               "GET"     -> case target of
```
begriffs: Maybe an if statement would be more compact rather than the sub case.

```haskell
"GET" ->
  if target == TargetRoot
    then ActionInspect
    else ActionRead
```

src/PostgREST/App.hs
```diff
@@ -101,37 +102,31 @@ app dbStructure conf apiRequest =
...
     (ActionRead, TargetIdent qi, Nothing) ->
-      serves [CTApplicationJSON, CTTextCSV] (iAccepts apiRequest) $ \contentType ->
-      case readSqlParts of
-        Left e -> return $ responseLBS status400 [jsonH] $ toS e
-        Right (q, cq) -> do
-          let singular = iPreferSingular apiRequest
-              stm = createReadStatement q cq singular
-                    shouldCount (contentType == CTTextCSV)
-          respondToRange $ do
+      servesMatchingContentTypes' $ \contentType ->
```
begriffs: The tick mark naming convention confused me. Maybe it's worthwhile to use the non-tick version and pass in the arguments repetitively.

src/PostgREST/App.hs
```diff
@@ -101,37 +102,31 @@ app dbStructure conf apiRequest =
...
+        servesReadRequest' $ \q cq -> do
+            let singular = iPreferSingular apiRequest
```
begriffs: Happily this let statement and the statements in the do block can be dedented.

src/PostgREST/ApiRequest.hs
```diff
@@ -47,15 +61,17 @@ data PreferRepresentation = Full | HeadersOnly | None deriving Eq
...
+toHeader ct = (hContentType, toMime ct <> "; charset=utf-8")
 
-toHeader :: ContentType -> ByteString
-toHeader CTApplicationJSON = "application/json"
-toHeader CTTextCSV         = "text/csv"
-toHeader CTOpenAPI         = "application/openapi+json"
-toHeader CTAny             = "*/*"
-toHeader (CTOther ct)      = ct
+-- | Convert from ContentType to a ByteString representing the mime type
+toMime :: ContentType -> ByteString
```
begriffs: Much better names!

src/PostgREST/App.hs
```diff
@@ -210,40 +201,36 @@ app dbStructure conf apiRequest =
...
 
-    (ActionInvoke, TargetProc qi,
-     Just (PayloadJSON (UniformObjects payload))) -> do
-        let p = V.head payload
-            singular = iPreferSingular apiRequest
-        serves [CTApplicationJSON] (iAccepts apiRequest) $ \_ -> case readSqlParts of
-          Left e -> return $ responseLBS status400 [jsonH] $ toS e
-          Right (q,cq) -> respondToRange $ do
+    (ActionInvoke, TargetProc qi, Just (PayloadJSON (UniformObjects payload))) ->
+        servesMatchingContentTypes' $ \_ ->
```
begriffs: dedent

src/PostgREST/App.hs
```diff
@@ -261,38 +248,55 @@ app dbStructure conf apiRequest =
...
-  mutateDbRequest = DbMutate <$> buildMutateRequest apiRequest
-  selectQuery = requestToQuery schema False <$> readDbRequest
-  countQuery = requestToCountQuery schema <$> readDbRequest
-  mutateQuery = requestToQuery schema False <$> mutateDbRequest
-  readSqlParts = (,) <$> selectQuery <*> countQuery
-  mutateSqlParts = (,) <$> selectQuery <*> mutateQuery
-  respondToRange response = if topLevelRange == emptyRange
-                            then return $ errResponse status416 "HTTP Range error"
-                            else response
   rangeHeader queryTotal tableTotal = let lower = rangeOffset topLevelRange
```
begriffs: Even though this function hasn't been modified, it should probably have the let statement on the next line and indented to avoid the extreme indentation.

src/PostgREST/App.hs
```diff
@@ -261,38 +248,55 @@ app dbStructure conf apiRequest =
...
+servesDbRequest selectQuery query resp =
+  case (,) <$> selectQuery <*> query of
+    Left e -> return $ responseLBS status400 [toHeader CTApplicationJSON] $ toS e
+    Right (sq,q) -> resp sq q
+
+servesMatchingContentTypes :: Monad m => [ContentType] -> Action -> (ContentType -> m Response) -> m Response
+servesMatchingContentTypes accepts action = serves contentTypesForRequest accepts
+  where
+    contentTypesForRequest =
+      case action of
```
begriffs: It's nice to see these all in one place to compare them better.

ruslantalpa:
The construct
```
servesMatchingContentTypes' $ \contentType ->
        servesMutateRequest $ \sq mq -> do
```
is confusing to me and looks a bit like js callback style. Only after looking carefully inside the functions i could understand what is going on.
I understand that it tries to be continuation passing style but somehow it (function signature in this style) does not seem right to me.
In my mind CPS is similar in "feel" to `bind =<<` and that interface works as long as there is a clear `thing` being piped through a chain of functions.
For these couple of functions i do not see that `thing` and i can not see how each function transforms or does something with that thing.
Maybe i am missing something or maybe the naming of the functions is not that great. Can you explain your thinking behind this?

For the following functions i have the same comment as above, i don't "feel" the CPS here
```
servesDbRequest :: Monad m => Either Text SqlQuery -> Either Text SqlQuery -> (SqlQuery -> SqlQuery -> m Response) -> m Response
servesReadRequest :: Monad m => NonnegRange -> Either Text SqlQuery -> Either Text SqlQuery -> (SqlQuery -> SqlQuery -> m Response) -> m Response
```
Another comment is to maybe remove the `Either` from the types and use aaplicative when calling the functions, and maybe even combine `selectQuery` and `countQuery` in a (,) since they are kind of inseparable and might make using applicative when calling a bit nicer

diogob:
@ruslantalpa I also didn't like that callback style. I was just trying to isolate the error responses and make the data dependencies explicit, something I can definitely do using `ifs`. 

I'll work more on this during this week.

ruslantalpa:
Yes i saw how you isolated the errors and that's a progress, and now that you've done that this idea emerges (not sure if it's worth doing)
if you look at `servesMatchingContentTypes servesDbRequest servesReadRequest`, they either call errResponse or the continuation.
Maybe they can all be refactored to use the `Either (httpstatus,errorstring)` monad and use bind (hop i haven't said said something stupid).
and in the end, the calling function for a `Left` result calls `errResponse` and for the `Right` it would be calling the continuation passed down by warp

ruslantalpa:
This would should also eliminate the `either` type from most of the parameters of these functions (except the last one of course)

src/PostgREST/App.hs
```diff
@@ -252,47 +239,61 @@ app dbStructure conf apiRequest =
...
+        ActionRead   -> [CTApplicationJSON, CTTextCSV]
+        ActionCreate -> [CTApplicationJSON, CTTextCSV]
+        ActionUpdate -> [CTApplicationJSON, CTTextCSV]
+        ActionDelete -> [CTApplicationJSON, CTTextCSV]
+        ActionInvoke -> [CTApplicationJSON]
+        ActionInspect -> [CTOpenAPI]
+        _ -> []
```
begriffs: Maybe you should handle `ActionInfo` and `ActionInappropriate` explicitly and remove the wildcard match. That way any new future actions will give a compiler warning of non-exhaustive pattern match.

diogob:
@ruslantalpa @begriffs I gave up on those serves* functions as they really seemed to complicate and make the code more cryptic.

But I moved the content type verification to a single case before the main one.

I also pushed the error responses down the function chain and now the `buildMutateRequest`and `buildReadRequest` are named `mutateRequestOrError` and `readRequestOrError` respectively and both return a `Response` on their `Left` case.

diogob:
@ruslantalpa I'm still considering the idea of using the Either monad to simplify all the nested Either cases, but this will be a future refactor, I want to avoid too much churn in a single PR.

src/PostgREST/App.hs
```diff
@@ -372,8 +378,13 @@ treeRestrictRange maxRows_ request = pure $ nodeRestrictRange maxRows_ `fmap` re
     nodeRestrictRange :: Maybe Integer -> ReadNode -> ReadNode
     nodeRestrictRange m (q@Select {range_=r}, i) = (q{range_=restrictRange m r }, i)
 
-buildReadRequest :: Maybe Integer -> [Relation] -> [(Text, Text)] -> ApiRequest -> Either Text ReadRequest
-buildReadRequest maxRows allRels allProcs apiRequest  =
+mapLeft :: (a -> b) -> Either a c -> Either b c
+mapLeft f (Left x) = Left $ f x
+mapLeft _ (Right x) = Right x
+
+readRequestOrError :: Maybe Integer -> [Relation] -> [(Text, Text)] -> ApiRequest -> Either Response ReadRequest
```
ruslantalpa: I think the old names are better, the type is enough to convey there is a "error" case (plus the fact that `<$>` is used).

diogobnulogy: true, but may we drop the build prefix and keep it just `readRequest`? Build makes me think we are going to return some sort of Builder or yield something.

diogob: true, but may we drop the build prefix and keep it just readRequest? Build makes me think we are going to return some sort of Builder or yield something.

diogob:
I think this is good to merge.

begriffs:
I'll be away from computers for the next few days but will review and merge this first thing when I'm back.

src/PostgREST/App.hs
```diff
@@ -372,11 +378,16 @@ treeRestrictRange maxRows_ request = pure $ nodeRestrictRange maxRows_ `fmap` re
     nodeRestrictRange :: Maybe Integer -> ReadNode -> ReadNode
     nodeRestrictRange m (q@Select {range_=r}, i) = (q{range_=restrictRange m r }, i)
 
-buildReadRequest :: Maybe Integer -> [Relation] -> [(Text, Text)] -> ApiRequest -> Either Text ReadRequest
-buildReadRequest maxRows allRels allProcs apiRequest  =
+mapLeft :: (a -> b) -> Either a c -> Either b c
```
begriffs: From the lens [docs](https://hackage.haskell.org/package/lens-4.15.1/docs/Control-Lens-Prism.html) it seems like `mapLeft foo` is the same as `& _Left %~ foo` but the lens way might be gratuitously terse.

src/PostgREST/App.hs
```diff
@@ -372,11 +378,16 @@ treeRestrictRange maxRows_ request = pure $ nodeRestrictRange maxRows_ `fmap` re
...
+mapLeft f (Left x) = Left $ f x
+mapLeft _ (Right x) = Right x
+
+readRequest :: Maybe Integer -> [Relation] -> [(Text, Text)] -> ApiRequest -> Either Response ReadRequest
```
begriffs: The return type for this function strikes me as odd somehow, mixing an HTTP response with a ReadRequest. I recognize that it reduces duplication in the code though. I'm not seeing another way to do it without duplication.

PR #739, opened 2016-11-13
Introduce new 'Prefer' header, rpcParams=asJson

dsimunic:
At present, the parameters from an urlencoded form POST are first converted to JSON, and then this json is expaned into named parameters in [`callProc`](https://github.com/begriffs/postgrest/blob/41c1cf6e018b1ced2f720219fd4928cdc2259fa6/src/PostgREST/QueryBuilder.hs#L261). 

This flag instructs `callProc` to pass all parameters as a single JSON object to a stored procedure. If the request contains the header `Prefer: rpcParams=asJson`, then the call assumes a stored procedure with a single input JSON parameter and passes the complete params object as JSON.

This is super-useful for letting the database dispatch on more complicated HTML forms, or when the application has dynamically generated forms. Instead of creating multiple stored procedure for every possible combination of fields, a single procedure can take care of inspecting, modifying, and dispatching the request as needed.

For static HTML forms, this requires merely adding the header on the reverse proxy, without having to manipulate the request any further.

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -238,8 +238,8 @@ addJoinConditions schema (Node nn@(query, (n, r, a)) forest) =
     addCond query' con = query'{flt_=con ++ flt_ query'}
 
 type ProcResults = (Maybe Int64, Int64, JSON.Value)
-callProc :: QualifiedIdentifier -> JSON.Object -> SqlQuery -> SqlQuery -> NonnegRange -> Bool -> Bool -> H.Query () (Maybe ProcResults)
-callProc qi params selectQuery countQuery _ countTotal isSingle =
+callProc :: QualifiedIdentifier -> JSON.Object -> SqlQuery -> SqlQuery -> NonnegRange -> Bool -> Bool -> Bool -> H.Query () (Maybe ProcResults)
+callProc qi params selectQuery countQuery _ countTotal isSingle paramsAsJson =
   unicodeStatement sql HE.unit decodeProc True
   where
     sql = [qc|
@@ -258,7 +258,9 @@ callProc qi params selectQuery countQuery _ countTotal isSingle =
           |]
           -- FROM (select * from {sourceCTEName} {limitF range}) t;
     countResultF = if countTotal then "("<>countQuery<>")" else "null::bigint" :: Text
-    _args = intercalate "," $ map _assignment (HM.toList params)
+    _args = if paramsAsJson
+                then insertableValueWithType "json" $ JSON.Object params
+                else intercalate "," $ map _assignment (HM.toList params)
     _procName = pgFmtLit $ qiName qi
     _assignment (n,v) = pgFmtIdent n <> ":=" <> insertableValue v
     _callSql = [qc|select * from {fromQi qi}({_args}) |] :: Text
@@ -497,6 +499,9 @@ emptyOnNull val x = if null x then "" else val
 insertableValue :: JSON.Value -> SqlFragment
 insertableValue JSON.Null = "null"
 insertableValue v = (<> "::unknown") . pgFmtLit $ unquoted v
+
+insertableValueWithType :: Text -> JSON.Value -> SqlFragment
+insertableValueWithType t v = (<> "::" <> t) . pgFmtLit $ unquoted v
 
 whiteList :: Text -> SqlFragment
 whiteList val = fromMaybe
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -501,7 +501,8 @@ insertableValue JSON.Null = "null"
 insertableValue v = (<> "::unknown") . pgFmtLit $ unquoted v
 
 insertableValueWithType :: Text -> JSON.Value -> SqlFragment
-insertableValueWithType t v = (<> "::" <> t) . pgFmtLit $ unquoted v
+insertableValueWithType t v =
+  pgFmtLit (unquoted v) <> "::" <> t
 
 whiteList :: Text -> SqlFragment
 whiteList val = fromMaybe
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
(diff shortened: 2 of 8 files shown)


ruslantalpa:
couldn't you just have the function with just a single parameter that is of type jsonb then have the entire payload be under that single key?

dsimunic:
It would not work for posting URL encoded forms.

ruslantalpa:
... which could have a few lines of js to take the form and send it as a single json value

dsimunic:
This removes the need for javascript for static forms. It complements nicely the already existing support for accepting POSTs from url-encoded forms. Together, these two features enable elegant migration of legacy system backends with a trivial redirect on the proxy server. The pull request is inspired by a real-world production scenario.

Besides, one can make the same "few lines of javascript" argument for the already existing support for url-encoded forms.

ruslantalpa:
Forgot that is already supported. Not every real world scenario has to be supported by postgrest however in this case it does not seem too bad. I agree this is useful but i still feel a bit uneasy because it seems like this allows for improperly calling exposed functions, but then again so does sending json body that does not fit.
As a conclusion, feature is good, code is good but this  (and the support of url encoding) needs a bot more consideration from the security point of view before merging it. Can this be misused somehow for sql injection or does it create an additional burden on the ppl writing the rpc function and by that i mean if i write a function, do i need to pay extra attention to the input parameters and sanitise data instead of relying on the input type?

dsimunic:
I don't think there are any new security implications that this specific patch introduces. The same goes for support for url encoding, for posting JSON, or for posting anything though postgREST. In other words, this is as secure as postgREST itself. Every request one sends to postgREST will go through exactly same mechanisms.

Furthermore, this flag just tells postgREST to pass the already-existing JSON as such instead of unpacking the first level of the object into parameters. And it even adds the explicit `::json` type for the parameter (in contrast with the rest of the code that uses `::unknown`), making sure postgres does its own check on JSON. 

So, no additional burden on the developer--it does not increase the risk of injection compared to any other rpc request where the the first level of parameters object gets expanded.

In general, I don't see postgREST introducing any new risks or hidden gotchas for developers. To the contrary, the sole purpose of the product is helping developers reduce the risks by not having to write brittle and boring data transformations between web server and the database. When it comes to rpc support, it is transparent from the perspective of the procedure writer: one just writes procedures as one normally would, knowing that if the call succeeds, the parameters are validated by postgres.

ruslantalpa:
Actually there is no validation going on in postgrest, you can create an invalid call to a rpc and the db will try to call a function that matches the signature. I am not 100% sure as you that this is safe (has nothing to do with this particular pr though)

dsimunic:
That is as safe as it gets:

```
POST http://localhost:3000/rpc/non_existent_function

...

LOG:  execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
LOG:  statement: set local role 'anon';
ERROR:  function api.non_existent_function(json) does not exist at character 47
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
STATEMENT:  
	            WITH pg_source AS (select * from... 
```

Postgrest does the right thing: pass the call to pg and let it do its job. Not sure what else one could have done.

dsimunic:
If I understood you well, there is nothing wrong with this  pr. Can you please merge it?

ruslantalpa:
yes, nothing wrong. begriffs is the one merging.
this particular PR does not change anything security wise, it's the same as before, all i am saying is that it feels a bit awkward one is able to get an invalid query to be executed (and the db raise an error because of the function signature mismatch)
and there is no big chance of someone putting in the time to think if this can be exploited :)
PS: from my point of view this can be merged

begriffs:
I'll review this later tonight. Thanks for the contribution!

begriffs:
Makes sense for handling dynamically generated forms. You may also be able to use named parameters with default values to handle the possible combinations of fields in a form. In any case I trust that your situation is chaotic with dynamic forms and that your PR makes life easier on the database side. As another benefit, the code change is nice and small.

Here are a few notes:
* I noticed that existing IANA multi-word [http preferences](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#preferences) use dashes rather than camel case. What about `Prefer: params=single-object`?
* Please modify the OpenAPI output to include the new Prefer header in [makeProcParam](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/OpenAPI.hs#L185-L190). You can copy the use of `makePreferParam` in the function prior.
* Add tests to the test suite, both for sending JSON and `application/x-www-form-urlencoded`. Here's how:
  * Add your new function that accepts a single json parameter to `fixtures/schema.sql`
  * Add new test cases to `QuerySpec.hs`. There are RPC tests in there you can copy and modify
* Create a CHANGELOG entry to note your awesome contribution!

jackfirth:
It may be worthwhile to name the preference `postgrest-params` or `vnd.postgrest.params`, to make it clear this is a Postgrest-specific preference as opposed to a service-agnostic preference.

dsimunic:
@ruslantalpa Thanks.

@begriffs Thanks for your feedback, will do.

@jackfirth Makes sense. I am neutral as to which one to use. Maybe prefixing with `postgrest-`  would be more in line with the existing preference names?

src/PostgREST/QueryBuilder.hs
```diff
@@ -498,6 +500,9 @@ insertableValue :: JSON.Value -> SqlFragment
 insertableValue JSON.Null = "null"
 insertableValue v = (<> "::unknown") . pgFmtLit $ unquoted v
 
+insertableValueWithType :: Text -> JSON.Value -> SqlFragment
+insertableValueWithType t v = (<> "::" <> t) . pgFmtLit $ unquoted v
```
begriffs: Might be written a little more simply as
```haskell
insertableValueWithType t v =
  pgFmtLit (unquoted v) <> "::" <> t
```

dsimunic: Thanks, changed. Running tests.

dsimunic: Done.  BTW, I didn't prefix the key with `postgres-` to remain consistent with all other `Prefer` keys.

begriffs:
Great feature; really appreciate your diving into Haskell to contribute this.

As for the name, `params` seems fine. It's concise and unlikely to clash with standards any time soon so no need for a vendor prefix. Also our `plurality` preference has no prefix. I think we're good to merge! 🏁

PR #741, opened 2016-11-13
Support conditional requirement for -c argument

### LINKED ISSUE #736: The `--example-config` param doesn't output anything
On the master branch the `-c` flag is marked as mandatory so passing `--example-config` alone is treated as a usage error. The solution is to remove `optparse-applicative` and do simpler command line parsing ourselves. This is desirable because the command line options have become much simpler now that we have a config file.

begriffs:
Fixes #736

The `readOptions` function is getting ungainly. I added comments to help explain it. The notion of having a command line argument that is conditionally mandatory does not fit Haskell's notion of Applicative, hence using optparse-applicative to parse the args is awkward. However that library is pretty convenient with help text formatting etc that I kept it and am working around it a little.

Code reviews appreciated.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -47,6 +47,7 @@ library
   default-language:    Haskell2010
   default-extensions:  OverloadedStrings, QuasiQuotes, NoImplicitPrelude
   build-depends:       aeson
+                     , ansi-wl-pprint
                      , base >= 4.8 && < 6
                      , bytestring
                      , case-insensitive
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -28,15 +28,15 @@ import qualified Data.CaseInsensitive        as CI
 import qualified Data.Configurator           as C
 import qualified Data.Configurator.Types     as C
 import           Data.List                   (lookup)
-import           Data.Text                   (strip, intercalate)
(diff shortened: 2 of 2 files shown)


ruslantalpa:
Looking at the example from https://github.com/pcapriotti/optparse-applicative
it seems you can have a mandatory config but also a help switch.
Why not have this behaviour (which seems to be supported by opt parse)
- no arguments -- display help
- help argument -- display help
- -c argument -- normal usage

help should show only info for 2 arguments -c and -h (help) as parameters for for command lines and if you want also display a sample config when help is invoke rather then having a special parameter to extract info about a sample config, include it in the help screen

ruslantalpa:
`progDesc` could be used for the sample config file

begriffs:
Good point, the help text doesn't have much to say so might as well output the example config file too. I changed it to include the example at the end of the message (as a footer). I removed the `-c` option. There is only one argument now, just the config filename. All other flags can go into the config file, keeping the command line simple.

One nice property of the old way was that `postgrest --example-config > pg.conf` outputs exactly the lines necessary for the config file whereas in this new way the user has to split and dedent them from the usage message. Not too big of a deal though.

My final final thought is that the usage text might get super long if the config file gets a lot of new options. At that point we could choose to output a minimal config with a link to full documentation as a comment.

PR #745, opened 2016-11-23
Moves request error handling to user api request

diogob:
The idea behind this refactor is to extract error cases that were inside the `ApiRequest` type (namely as the `ActionInappropriate` and `PayloadParseError` constructors).

Now this error cases are handled by the `userApiRequest` that returns an `Either ApiRequestError ApiRequest`.

The type `ApiRequestError` has a function inside the error module that converts this type into an HTTP response. This function is used in the `postgrest` function making the `app` function slightly simpler as well.

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -36,8 +36,8 @@ import           Network.Wai.Parse         (parseHttpAccept)
 import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, restrictRange, rangeGeq, allRange, rangeLimit, rangeOffset)
 import           Data.Ranged.Boundaries
 import           PostgREST.Types           (QualifiedIdentifier (..),
-                                            Schema, Payload(..),
-                                            UniformObjects(..))
+                                            Schema,
+                                            PayloadJSON(..))
 import           Data.Ranged.Ranges        (Range(..), singletonRange, rangeIntersection, emptyRange)
 
 type RequestBody = BL.ByteString
@@ -95,7 +95,7 @@ data ApiRequest = ApiRequest {
   -- | Content types the client will accept, [CTAny] if no Accept header
   , iAccepts :: [ContentType]
   -- | Data sent by client and used for mutation actions
-  , iPayload :: Maybe Payload
+  , iPayload :: Maybe PayloadJSON
   -- | If client wants created items echoed back
   , iPreferRepresentation :: PreferRepresentation
   -- | If client wants first row as raw object
@@ -150,13 +150,13 @@ userApiRequest schema req reqBody
       CTApplicationJSON ->
           either Left (\val -> case ensureUniform (pluralize val) of
             Nothing -> Left "All object keys must match"
-            Just json -> Right $ PayloadJSON json) (JSON.eitherDecode reqBody)
+            Just json -> Right json) (JSON.eitherDecode reqBody)
       CTTextCSV ->
           either Left (\val -> case ensureUniform (csvToJson val) of
             Nothing -> Left "All lines must have same number of fields"
-            Just json -> Right $ PayloadJSON json) (CSV.decodeByName reqBody)
+            Just json -> Right json) (CSV.decodeByName reqBody)
       CTOther "application/x-www-form-urlencoded" ->
-        Right . PayloadJSON . UniformObjects . V.singleton . M.fromList
+        Right . PayloadJSON . V.singleton . M.fromList
                     . map (toS *** JSON.String . toS) . parseSimpleQuery
                     $ toS reqBody
       ct ->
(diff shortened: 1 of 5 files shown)


src/PostgREST/QueryBuilder.hs
```diff
@@ -113,7 +113,6 @@ createReadStatement selectQuery countQuery isSingle countTotal asCsv =
 createWriteStatement :: QualifiedIdentifier -> SqlQuery -> SqlQuery -> Bool ->
                         PreferRepresentation -> [Text] -> Bool -> Payload ->
                         H.Query UniformObjects (Maybe ResultsWithCount)
-createWriteStatement _ _ _ _ _ _ _ (PayloadParseError _) = undefined
```
ruslantalpa: i like how these are going away :)

src/PostgREST/ApiRequest.hs
```diff
@@ -113,84 +117,79 @@ data ApiRequest = ApiRequest {
...
-     . parseSimpleQuery
-     $ rawQueryString req
-  , iJWT = tokenStr
-  }
-
+userApiRequest :: Schema -> Request -> RequestBody -> Either ApiRequestError ApiRequest
+userApiRequest schema req reqBody
+  | isTargetingProc && method /= "POST" = Left ErrorActionInappropriate
+  | topLevelRange == emptyRange = Left ErrorInvalidRange
+  | shouldParsePayload && isLeft payload = either (Left . ErrorInvalidBody . toS) undefined payload
```
begriffs: This is good! Lewis Stern just told us in gitter about a problem that turned out to be caused because postgrest didn't issue a descriptive error for a malformed post body. Your change will prevent other people from getting confused.

Could you add a test that posting a malformed body returns an error? Lewis' particular problem was sending a JSON object quoted as a string rather than as a real object. I'm not sure your code would catch that case because a string is valid json. Postgrest will probably treat it as an empty object, which is confusing for the user.

src/PostgREST/ApiRequest.hs
```diff
@@ -113,84 +117,79 @@ data ApiRequest = ApiRequest {
...
+  topLevelRange = fromMaybe allRange $ M.lookup "limit" ranges
+  action =
+    if isTargetingProc
+      then ActionInvoke
+      else
+        case method of
+            "GET"     -> if target == TargetRoot
+                          then ActionInspect
+                          else ActionRead
+            "POST"    -> ActionCreate
```
begriffs: What about putting `if isTargetingProc` inside this POST case to avoid indenting the whole case statement?

src/PostgREST/ApiRequest.hs
```diff
@@ -113,84 +117,79 @@ data ApiRequest = ApiRequest {
...
+  target = case path of
+              []            -> TargetRoot
+              [table]       -> TargetIdent
+                              $ QualifiedIdentifier schema table
+              ["rpc", proc] -> TargetProc
+                              $ QualifiedIdentifier schema proc
+              other         -> TargetUnknown other
+  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke]
+  relevantPayload = if shouldParsePayload
+                      then case payload of
```
begriffs: Check out `rightToMaybe` in the either package.

src/PostgREST/App.hs
```diff
@@ -81,17 +80,18 @@ postgrest conf refDbStructure pool getTime =
...
+        let eClaims = jwtClaims
+              (secret <$> configJwtSecret conf) (iJWT apiRequest) time
+            authed = containsRole eClaims
+            handleReq = runWithClaims conf eClaims (app dbStructure conf) apiRequest
+            txMode = transactionMode $ iAction apiRequest
 
-    resp <- either (pgErrResponse authed) id <$> P.use pool
-      (HT.run handleReq HT.ReadCommitted txMode)
-    respond resp
+        resp <- either (pgErrResponse authed) id <$> P.use pool
```
begriffs: `mapLeft` from either package

src/PostgREST/QueryBuilder.hs
begriffs: +1

src/PostgREST/Types.hs
```diff
@@ -111,9 +110,7 @@ unUniformObjects (UniformObjects objs) = objs
 -- | When Hasql supports the COPY command then we can
 -- have a special payload just for CSV, but until
 -- then CSV is converted to a JSON array.
-data Payload = PayloadJSON UniformObjects
-             | PayloadParseError BS.ByteString
-             deriving (Show, Eq)
+data Payload = PayloadJSON UniformObjects deriving (Show, Eq)
```
begriffs: Is there a reason to keep Payload as a type now that there is only one constructor? Perhaps we could use UniformObjects throughout the code instead.

diogob: True, I actually only refrained from doing this to avoid too many steps in one PR 😛 
I'll include this

src/PostgREST/ApiRequest.hs
diogob: Sure, I'll check that out.

src/PostgREST/ApiRequest.hs
diogob: I like moving the if inside the case, it also creates a nice symmetry with the GET case.
If we need a different mapping later one (1 Action -> Multiple verbs) we probably will want to rethink this section of the code anyway.

src/PostgREST/ApiRequest.hs
diogob: @begriffs from what I've seen we would still accept an invalid PayloadJSON as long as is a valid JSON.
I'd love to add that constraint but I believe that would be more suited to another pull request, since this is becoming quite large already.

begriffs:
Thank you!

PR #747, opened 2016-11-23
Use PGHOST to Set Postgresql Test Host

dsimunic:
Specify the postgresql host during `stack test` through the `PGHOST` environment variable. This simplifies testing on macOS docker:

    PGHOST=192.168.64.1 PGUSER=postgres stack test --work-dir .stack-work-docker

Equally we can choose to specify the host in the container invocation (assuming we built a container with all prerequisites for testing):

    docker run --rm -it -v `pwd`:`pwd` -e PGHOST=192.168.64.1 -e PGUSER=postgres -w="`pwd`" postgrest-test bash -c 'stack test --work-dir .stack-work-docker'

If the variable is not set, the connection defaults to `localhost` as before.


ruslantalpa:
1. in your example you say PGUSER PGHOST but in the code it's PGHOST PGPORT
2. Are these env vars standart in pg setups? if not then i would suggest having a single env var called POSTGREST_TEST_DB that takes in the entire connection string (or some other env var name)

dsimunic:
PGUSER influences psql, so that when the database scripts are run, psql knows which database to connect to. 

The only variables needed to form the test connection string are PGHOST and PGPORT. It would make no sense to introduce a new env var, since you have to set PGHOST for the psql part to work. (That worked the same before, but the problem was the hard-coded localhost in the `testDbConn`)

ruslantalpa:
psql is not influenced by those vars https://github.com/begriffs/postgrest/blob/master/test/SpecHelper.hs#L87
and that's a good thing, you would not want part of the code connecting (test executing) in a different way to the db then the other part (db resetting)

i say it makes sense a lot to have a single connection string env var to completley override the location of the database, it's name, and the user/password with which to connect to it.
The way it is right now (and the way it would still be after your pr) is that to run the test i should know how to name the database and how to name the user that has access to it and that that particular user needs to be able to login without a password.
It's much simple to have a default string in the docs (or maybe when you try to run the tests) that says "Hey, we tried to connect to the db using this connection string but we failed, either check and make sure that you can connect with psql using the same string or provide your own connection string by setting this env var ..."

dsimunic:
Psql is influenced by those vars because `readProcess` inherits the current user's environment.

See for yourself:

```
$ PGHOST=somehost ghci
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help
Prelude> :m +System.Process
Prelude System.Process> readProcess "psql" [] []
psql: could not translate host name "somehost" to address: nodename nor servname provided, or not known
*** Exception: readCreateProcess: psql (exit 2): failed
Prelude System.Process> readProcess "env" [] []
...

```

Using the standard `PG*` variables is the way to go. No need to introduce new env vars. 

This pr is for people developing postgrest, not its users. Anyone able to run the test from vagrant or docker who wants to run with Postgres on another host knows that psql takes `PG*` env vars, or will quickly find out. Every standard postgres install comes with the user `postgres`, so it's always safe to specify `PGUSER=postgres`, and possibly a password. 

The tests are always run with `postgrest_test_authenticator` user, and the database is always called `postgrest_test`, that's already hard-coded in test scripts and the code, no need to change that.

Database creation already worked with `PGHOST` and `PGUSER`, and it was the first thing I tried (and I suspect others that wanted to connect to an existing postgres). This pr fixes the expectation that the tests will honor the host setting.

dsimunic:
I see why you thought that `psql` is not influenced by environment variables: if you don't provide either env vars or the `.pgpass` file, it defaults to `localhost` and the current user. 

And since the current test setup requires postgresql to be local for the tests to work, local postgresql will always have the current user set up as well, so you won't notice any errors on connection.

ruslantalpa:
run this
`PGHOST=192.168.64.1 PGUSER=inexistentuser stack test --work-dir .stack-work-docker`
the tests will still run, that var is ignored because it is overridden by command line parameters.

dsimunic:
You're right. So, the PR is good as is: one only needs to specify `PGHOST` and all tests will run correctly on any Postgres host. 

Not specifying PGHOST will default to running tests on a locally installed Postgresql.

ruslantalpa:
ok, i am not making myself understood, I don't think the pr is good as is :), it only fixes one specific use case

There are 2 separate issues here:

1. Ability to specify every aspect of the connection string, not just host/port but also user/password/database. 
If this is going to be done, might as well do it all the way.
The way things are now (and even with this pr), before someone is able to run the tests he has to manually create the db, add the user, edit pg_hba.conf file to allow passwordleless auth and that is especially bad when the db you want to use is not your localhost (and the user has to look in the source code to find out the name of the user and the name of the db). All these problems go away if he is able to configure all the params.

2. The method of specifying the connect params, use the PG**** params that might be present on the users box or have a single custom env var. 
The problem with PG** vars is that it's more verbose to specify each thing individually then have a single param with a connection string. Another problem is that if the user has those setup he probably wants to to connect to the db of his current project fast (what if he decides those vars are for his production db) and if you use those default vars silently, he will suddenly wake up to a bunch of postgrest test tables that have nothing with his project. It's much more simple and clear to say in the docs (and not even have a default connection string to eliminate mistakes): "You can run the tests like this"
```
POSTGREST_TEST_DB=postgrest://user:password@host:port/db stack test
```
then have a bunch of info on how postgrest used the default PG vars and how you can look them up on the postgresql docs page and so on.

dsimunic:
OK, I get your point. It indeed makes sense to not have to touch postgres at all.

How about this compromise: running tests without touching the db server settings, but still using PG* variables. 

Covers the following scenarios:

(All scenarios assume that stack and the prerequisites are installed on the development machine, and the starting point is `git clone postgrest; cd postgrest; stack setup`; iow no need to set up anything special on the database server).

a) Locally installed postgres, having the same superuser name as the user running the test, ident auth for localhost (current scenario on circle ci):

```
$ apt-get install -y postgresql
$ stack test
```

b) Using an existing postgres server without the need to touch the server configuration (assuming Postgres.app on a mac, which might or might not have a superuser password):

```
$ PGHOST=localhost PGUSER=postgres PGPASSWORD=existing_superuser_pwd stack test
```

c) Using disposable postgres from the official postgres container (assuming another postgres already taking port 5432, and docker is accessible as localhost): 

```
$ export PGPASSWORD=pass PGUSER=postgres PGHOST=localhost PGPORT=5433 
$ docker run --name pg -e POSTGRES_PASSWORD=$PGPASSWORD -p 5433:5432 -d postgres
$ stack test
```

d) Running `stack test` in a container, with local postgres as in (b), assuming development happens on the host, and one already has a container with stack set up:

```
$ host_ip=$(ifconfig en0 | grep 'inet ' | cut -f 2 -d' ')
$ docker run -e PGHOST=$host_ip -e PGUSER=postgres PGPASSWORD=existing_superuser_pwd -v `pwd`:`pwd` -v ~/.stack-linux:/root/.stack --rm -it pgst-test bash -c 'stack test --work-dir .stack-work-docker'
```
This scenario is currently needed on macOS Sierra, as the test build fails with ghc 8.0.1.

e) Combo of (c) and (d):

```
$ docker run --name pg -e POSTGRES_PASSWORD=pass -d postgres
$ docker run --link pg:pg  -e PGHOST=pg -e PGUSER=postgres  -e PGPASSWORD=pass -v `pwd`:`pwd` -v ~/.stack-linux:/root/.stack --rm -it pgst-test bash -c 'stack test --work-dir .stack-work-docker'
```
Obviously, one can set this up in docker-compose, so it gets even easier to run the test.

--

As it stands, tests need two users: a superuser to create the test database and one test runner account to create different fixtures in the test database. If we don't want to touch the database server, then we only want to specify an existing superuser account, and let the tests do the rest. 

The test runner account (`postgrest_test`) is hard-coded and doesn't need ever to be specified, the password is randomly generated and only known for the duration of the test run. (This is not the case in existing code, `postgrest_test` is presently created without a password).

My main reason for not liking the URL connection string is that it's a pain to parse correctly. If you know of some library that can do it, then we can use that.

I'm working on the changes that support these five scenarios, will amend the pr later today.

dsimunic:
I spoke too fast. It's a standard URI and we can parse it with `Network.URI`. I'll give it a go.

ruslantalpa:
There is no need to parse that url, it is passed as a parameter as is to both psql and hasql lib. this is a standart postgresql way of specifying the connection string [32.1.1.2. Connection URIs](https://www.postgresql.org/docs/current/static/libpq-connect.html#AEN45347)

ruslantalpa:
it would be ideal if the test system would not need at least some super user privileges but it does since it needs to create different roles, so assume in the scenarios that the database is already created and the connection string is for a user that has the ability to to create other roles and do anything within that db

dsimunic:
We can meet that requirement with two different variables: `POSTGREST_TEST_CONNECTION` and `POSTGREST_TEST_ROLE`--the former does the whole business, and the latter requires the specified database to exist. If both are specified, the former takes precedence.

ruslantalpa:
i am not sure i understand what you mean by `POSTGREST_TEST_ROLE `?
The `POSTGREST_TEST_CONNECTION` contains the db name as well.
The more parameters the harder to understand. I think it's simple enough to say:

Run the test suit like this and make sure the user has enough privileges to create other users and grant them access to the specified database
`POSTGREST_TEST_CONNECTION=postgresql://user:password@host:port/dbname stack test`

It's more or leas clear/implied that both the database and the user need to exist before running the test suit.

dsimunic:
Yes, not the best choice of names. 

I like the idea of having the database created and dropped by the test runner, and not having to manually create it in advance. (This also supports the existing way of testing, so no need to make a breaking change).

Your suggestion of being able to specify an existing database is also great.

I'll add an additional parameter in the connection uri to signal that I want the database needs to be created: 

    POSTGREST_TEST_CONNECTION=postgres://user@host/dbname?createdb=yes

ruslantalpa:
if you do that you are back in the land of parsing the connection string, reading the special parameter, splitting the url to extract the db name then connecting to postgres, and not to that specific db since it does not exist, running separatley a "create database" statement, disconnecting and only after that you can use the full url to connect. To much complication only for the benefit of not requiring the user to create a database.
Just assume you get a connection string to the database for a user that has the right to do anything  with that db.

ruslantalpa:
Looks good, the only part i did not properly look at is how the user creation and their priviliges was moved to a .sh script
(maybe rename those 2 scripts with .sh in the end)
@begriffs take a look, this PR certainly does a better job at having the whole test process easier to pull off, especially the option to run everything in docker

begriffs:
Nice work making this elegant, guys! I confirmed that it also builds in Circle CI without a cache of the previous build.

However I am having problems getting it to work locally on my Mac.

* My postgres installation (9.6.1 built from source on on OS X 10.11.6) initially did not have a database of the same name as my user. This breaks the script.
   ```
   $ test/create_test_db "postgres://foo@localhost" postgrest_test
   psql: FATAL:  database "foo" does not exist
   psql: FATAL:  database "postgrest_test" does not exist
   psql: FATAL:  database "postgrest_test" does not exist
   psql: FATAL:  database "postgrest_test" does not exist
   psql: FATAL:  database "postgrest_test" does not exist
   postgres://postgrest_test_postgrest_test:n3vD9VnkxAQpwvyYa@localhost/postgrest_test
   ```
* However `createdb foo` it gets past this. Probably not a big deal since the standard installation process creates a user and matching db I think.
* My logging settings cause confirmation statements to get mixed with the create_test_db output
   ```
   $ echo $POSTGREST_TEST_CONNECTION
   CREATE EXTENSION ALTER DATABASE SET SET SET SET SET SET CREATE
   . . .
   postgres://postgrest_test_postgrest_test:PSZLyXUtU61e7890a@localhost/postgrest_test
   ```
  the sql command executions in that script should probably be redirected from stdout.

dsimunic:
I've added the test to check for any errors during connection and print a friendly message.

Redirecting all errors from running scripts to `/dev/null` would cause grief as you wouldn't see that something is wrong. In normal conditions it should not go wrong, so if it does, better to be aware of it.

I can't repro the stuff with statements going into the create_test_db output. The output is supposed to be already suppressed with `PGOPTIONS='-c client_min_messages=WARNING'`. Do you happen to have a `.psqlrc`or something else that might override this?

dsimunic:
@begriffs, you most likely had an `.psqlrc` somewhere that caused those notices to get added to the output. Weirdly, the settings in `.psqlrc` take precedence over the settings in the env vars. I've added the `-X` to ignore the `.psqrc` settings in the commit yesterday.

begriffs:
You were right, the `-X` removed those extra statements in the output. I am debugging another local problem and will provide more details later.

begriffs:
Actually after happening to restart my computer this all seems to work fine.

Love your thorough docs too. I'm going to move them into https://github.com/begriffs/postgrest-docs to keep the readme tidy in this repo. I'll let you know when that's done and we can remove them from here and merge this!

begriffs:
OK, all set! https://postgrest.readthedocs.io/en/v0.4/install.html#postgrest-test-suite

dsimunic:
OK. Removed the readme.

begriffs:
Thank you for improving our testing process! Really nice.

ruslantalpa:
I know this is already merged and it's nice, maybe one thing that can be simplified is the creation of the test database.
Maybe we can ditch (in the future) the database setup shell scripts in favour of something  like http://ephemeralpg.org/ since it maps really nice to how this PR uses a single env var for a connections string.

dsimunic:
I don't think the two are exclusive. For example, when developing new tests, it's good to be able to tweak the database before committing the updated sql, without the fear that the work in progress will be garbage collected during lunch.

I'd say the changes needed to support `pg_tmp` are to concat the SQL scripts into a single file,  to document its use as yet another alternative for testing, and modify `create_test_db` to use the unified SQL file.

Issue #751, opened 2016-11-27
Tighten cabal constraints to fix vanilla cabal build


SkinyMonkey:
Hi!

I tried to build from the master to check the OpenAPI feature.

Here is my environment:
OS: macosx
Version : 10.10.5 (Yosemite)
GHC version : 8.0.1
cabal-install version: 1.24.0.0

First the build failed telling me that the happy binary was missing.
I installed it with cabal by hand and now it gives me this error:

```
[ 1 of 13] Compiling Paths_postgrest  ( dist/dist-sandbox-301e6d5c/build/autogen/Paths_postgrest.hs, dist/dist-sandbox-301e6d5c/build/Paths_postgrest.o )
[ 2 of 13] Compiling PostgREST.Config ( src/PostgREST/Config.hs, dist/dist-sandbox-301e6d5c/build/PostgREST/Config.o )

src/PostgREST/Config.hs:118:12: error:
    • Variable not in scope:
        (<>) :: ParserInfo FilePath -> InfoMod a3 -> t3
    • Perhaps you meant one of these:
        ‘>>’ (imported from Protolude), ‘<|>’ (imported from Protolude),
        ‘<*>’ (imported from Protolude)
      Perhaps you want to remove ‘<>’ from the hiding clauses
      in one of these imports:
        ‘Protolude’ (src/PostgREST/Config.hs:(41,1)-(42,52))
        ‘Text.PrettyPrint.ANSI.Leijen’ (src/PostgREST/Config.hs:39:1-59)

src/PostgREST/Config.hs:120:16: error:
    • Variable not in scope: (<>) :: [Char] -> t5 -> t4
    • Perhaps you meant one of these:
        ‘>>’ (imported from Protolude), ‘<|>’ (imported from Protolude),
        ‘<*>’ (imported from Protolude)
      Perhaps you want to remove ‘<>’ from the hiding clauses
      in one of these imports:
        ‘Protolude’ (src/PostgREST/Config.hs:(41,1)-(42,52))
        ‘Text.PrettyPrint.ANSI.Leijen’ (src/PostgREST/Config.hs:39:1-59)

src/PostgREST/Config.hs:121:16: error:
    • Variable not in scope: (<>) :: t4 -> [Char] -> GHC.Base.String
    • Perhaps you meant one of these:
        ‘>>’ (imported from Protolude), ‘<|>’ (imported from Protolude),
        ‘<*>’ (imported from Protolude)
[...]

PR #752, opened 2016-11-28
Add constraint for vanilla cabal builds

begriffs:
In #751 @SkinyMonkey discovered that regular cabal had a build error. The cause was a breaking change in optparse-applicative 0.13.0.0. This is an error we would have seen when upgrading to Stack LTS 7.10 as well actually.

I'm wondering whether I should examine the package versions chosen in the current LTS snapshot and add major version constraints to the packages in the cabal file. This will ensure that people who use cabal can also comfortably contribute to the project.

While I have gotten used to Stack and rather enjoy it, I know some other Haskellers prefer using the constraint solver for finer grained control. Cabal now supports `new-build` as well which can reuse build artifacts so it's really not so bad.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -68,7 +68,7 @@ library
                      , lens
                      , lens-aeson
                      , network-uri
-                     , optparse-applicative
+                     , optparse-applicative >= 0.12.0.0 && < 0.13.0.0
                      , parsec
                      , protolude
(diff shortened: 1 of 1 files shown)


begriffs:
Adding constraints for everything is probably not worth the effort. Merging this because it does fix the cabal build. Stack is our main way to build but I'm happy to address cabal problems if they pop up in the future.

PR #758, opened 2016-12-05
Downgrade postgrest version in Dockerfile until new version is released

begriffs:
Dockerhub is getting confused


begriffs:
Closing this in favor of @deinspanjer's idea to better configure docker hub.

PR #759, opened 2016-12-06
WiP of dobi Docker build system

deinspanjer:
This is not ready yet, but I need some review on the Haskell aspects of the build.

Dobi is a Docker based build system.  All build actions happen inside isolated Docker containers which makes it much easier to have repeatable builds with only the dobi executable and Docker as dependencies.

The Dobi executable can be installed from https://github.com/dnephin/dobi/releases

The dobi.yaml file is the heart of the build, and it should be relatively easy to read through.

The parts that I need help with start in the dockerfiles/Dockerfile.build file.
I ran `dobi shell` and then went through each step of the dependencies one by one:
`cabal update` -- very quick to run and it leaves me with empty ~/.stack and ./.stack-work folders.
`stack install --allow-different-user hlint packdeps` -- very slow, but it seems to deal with only haskell related libraries and libpg-dev. I didn't see any references to postgrest itself in the log output.
`stack build --allow-different-user --fast` -- I wish it was fast!  I think this was the one that said it was 130 steps. This seemed to log a bunch of lines very similar to the previous command, but at the end, it also did specifically compile postgrest stuff at the end.
`stack build --allow-different-user --fast --test --no-run-tests` -- Another slow build from scratch and again, at the end, it built postgrest stuff.

So, I think my biggest question is: Are all of these commands are required for just building/installing dependencies?

The flow I was aiming for with this build setup is to have a builder image that contains the necessary build tools such as Haskell / stack / cabal and the system library dependencies such as lib-pg-dev.  Then, have the image be run as a container with the .stack-work and ~/.stack folders on the host system mapped in as volumes so they are inspectable and maybe even sharable with a host-native build (that depends on the cross-platform compatibility in Haskell).  Once the expensive part of dependency management is done, I was hoping subsequent builds would be only against the postgrest source files themselves and fairly quick without having to spend much time scanning or verifying any dependencies since dobi manages that with source and artifact timestamps.

Thoughts?


deinspanjer:
Sorry, a couple of further notes..

What does work so far is running either:
`dobi dependencies`
or
`dobi shell`

the shell command will drop you into a bash prompt that will look and feel like your regular postgrest dev directory, and you should be able to run various commands such as cabal or stack.

deinspanjer:
I am happy to report I have a much nicer way of doing this.  closing in favor of a new PR coming up shortly.

PR #761, opened 2016-12-06
Update DbStructure.hs

fab1an:
added regexp so that views that reference public tables are found.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -606,7 +606,7 @@ allSynonyms cols =
                 select case when match is not null then coalesce(match[8], match[7], match[4]) end
                 from regexp_matches(
                     CONCAT('SELECT ', SPLIT_PART(vcu.view_definition, 'SELECT', 2)),
-                    CONCAT('SELECT.*?((',vcu.table_name,')|(\w+))\.(', vcu.column_name, ')(\s+AS\s+("([^"]+)"|([^, \n\t]+)))?.*?FROM.*?',vcu.table_schema,'\.(\2|',vcu.table_name,'\s+(as\s)?\3)'),
+                    CONCAT('SELECT.*?((',vcu.table_name,')|(\w+))\.(', vcu.column_name, ')(\s+AS\s+("([^"]+)"|([^, \n\t]+)))?.*?FROM.*?(',vcu.table_schema,'\.|)(\2|',vcu.table_name,'\s+(as\s)?\3)'),
                     'nsi'
                 ) match
             ) as view_column_name


ruslantalpa:
from the gitter
```
Fabian Zeindl @fab1an 16:11
hmmm
I just realized that if you do 'set search_path='' ' before running the query, it will append public. to the views and everything will work
```

I would prefer to go that way if possible because we would not be touching this sensitive regexp and also because we would be matching against a fully qualified item `public.tablename` vs `tablename` which makes the regex a bit more stricter, thus better and not returning bad rows.
We've had in the past where it would return false relations and i think it's better to have less results but 100% correct then more results but some of them bad.

fab1an:
I tried setting the search path of "authenticator" by doing "alter role authenticator set search_path=''", but then other functions that are called by anonymous don't work anymore.

I think that maybe the updated search_path is carried into the new role, after "set role anonymous", so it breaks those users search_path as well.

ruslantalpa:
I that case i guess it's simpler to change the regexp (and it's actually the clean way to do it but a bit more dangerous).

@fab1an do you have time to do a few manual tests on this new regexp?
Here is how to do it.
Have a clean db with 4 tables, 2 in public schema, 2 in separate schema, have those tables related 2 by 2.

have an api schema, place there as many views as you can think of with those tables as sources.
in the definitions of those views, use these `sql features`

in the select section, have columns like this
```sql
select id, id as ID, tbl.name, tbl.name as NAME, tablename.name, tablename.name as NAME2 from tablename as tbl
select id, id as ID, tbl.name, tbl.name as NAME, tablename.name, tablename.name as NAME2 from tablename tbl

```
in the FROM section, have the table specified as

```
tablename, tablename as tbl, tablename tbl
```

also add a few views that do JOINS

When all those views are there, you can have a single query that compares the results generated by both queries each using a different version of the regexp. 
Publish that in a gist, i'll try to also come up with example views for it if i'll see cases missing.

fab1an:
@ruslantalpa Why not save the original search_path, set it to `''` before doing the query and restore the original search_path again?

ruslantalpa:
- it's more work
- the idea to go with search_path was more out of fear to break things then it being the correct way
- regexp fix is much much shorter, it's jus needs extra care/tests

fab1an:
Hm, I think search_path is the cleaner way, because a search_path of `''` should give completely normalized results. And, I haven't written Haskell in years, but it's only two db-commands, I think.

Unfortunately I don't have any time to do more testing at the moment, that might change in a couple of weeks, though. Thank you for your efforts with this!

ruslantalpa:
I would prefer a query that works reliably (and in isolation) regardless of the database settings which can defer from instance to instance.

fab1an:
@ruslantalpa That is exactly my point. There are so many different ways to set the search_path in postgresql, that I think that clearing it would give the only "standard"-query.

diogob:
@fab1an as a general principle the operation of setting a variable scoped to the transaction can be achieved by `SET LOCAL`.
I prefer the regex solution thought, it has less dependency on shared state. Also we should have good testing around this regex anyways.

fab1an:
@diogob I have to say, that I don't mind how this is achieved, but I honestly don't understand why the regexp fix is better (even though I wrote it ;) ).

I understand that additional db-commands means more to worry about, but if the search_path interferes with the results of the reflective query like that, that implies that you can break postgrest in a couple of different ways, which won't seem obvious to the user at all:

https://dba.stackexchange.com/questions/56023/what-is-the-search-path-for-a-given-database-and-user

A pity that we can't test this against real-world installations, maybe the solution would give the same results anyway.

ruslantalpa:
@fab1an for me, the regexp way is better because you can take out the query and test in isolation, in addition to that, it also allows someone to do a custom build with a completely custom query instead of the current one (that returns the same columns) and he does not have to worry about the rest of the code. what i mean by that is this. Say for someone the regexp does not work at all, what he can do is replace this query, with a `select * from customtable` and have a manual way of saying what view column comes from what table column.
That's the main reason i would like to keep this query isolated/self contained

fab1an:
@ruslantalpa That seems reasonable to me.

It would require some work, but what'd be nice would be some kind of "postgrest doctor" script, that would output debugging information about schemata that could be used in bug reports and to help fixing these kind of problems.

ruslantalpa:
The debug script as you call it, can be a single sql query that can be executed and return the exact view of the database structure with which postgrest works, the query is possible but a lot of work is needed in order to have postgrest queries unified into a single one, i.e. a lot of refactoring.

fab1an:
Okay, then it's probably not worth it.

begriffs:
My take

- We're already using a regex approach so continuing to adjust the regex is not making the code more fragile
- If it passes our test suite then we should probably conclude it's fine and if we find a problem we'll add that to the test suite and fix
- Speaking of tests, we should also include a test that illustrates the need for the change
- In the future perhaps we can switch to a non-regex approach but I haven't investigated how hard that would be

@fab1an do you have time to add a test or should I look into it?

fab1an:
I don't have time nor the haskell-experience :).

begriffs:
No problem, I really appreciate your debugging the regex issue. I'll add a test and merge unless anybody has objections.

begriffs:
I'm having trouble making a test case that passes on this branch but fails on master. Everything I try passes on master too. Can you explain how you discovered the problem?

fab1an:
Try this:

```SQL
create schema api;

create table consumers (
    id                  serial             not null unique,
    name                text               not null check (name <> ''),
    primary key (id)
);

create table orders (
    id                  serial             not null unique,
    consumer            integer            not null references consumers(id),
    number              integer            not null,
    primary key (id)
);

create view api.orders as
    select * from orders;

create view api.consumers as
    select * from consumers;
```

Query: `/consumers?select=*, orders{*}`

begriffs:
Thanks, I'll try this.

begriffs:
I applied your change in another PR and included a test.

PR #762, opened 2016-12-06
New Docker build system

deinspanjer:
The begriffs/postgrest-empty container needs to be built at least once.  It only needs to be rebuilt when something changes in it.
It is built with `docker build -t begriffs/postgrest-empty dockerfiles/`

To build a new Docker image containing a new postgrest binary, just run:
`stack --docker image container`

If you want to do a build of the binary without putting it in a new Docker image:
`stack --docker build`

All the magic is in the stack.yaml file.  Most of the options in there are actually
the defaults, but I put them in with comments for knowledge transfer.

Future ideas:
* We could build static and even compress the image and put it in an alpine image that
  has been tested more extensively with stack / Haskell apps:
  https://www.fpcomplete.com/blog/2016/10/static-compilation-with-stack
* We could add scripts to the build tree to build the postgrest-empty image and run the build
* We could add a script to tag the resulting image and push it to hub.docker.com
* Handle the version better? Right now there is an ENV variable in the Dockerfile.


deinspanjer:
I implemented a wrapper script and put it in the PR as well.

The only downside to it is that it isn't tied in any strong way to the configuration options of postgrest itself. If there were some new option added in the code, the wrapper script wouldn't automatically pass it on unless it was modified.

If we had a canonical config file such as postgrest.conf.dist or something else that described all the options, then I could modify the wrapper script to take it into account.
_

begriffs:
Looks like Stack can do a lot for us!

Am I missing a step? Here's what happens on my mac:

```
$ docker --version
Docker version 1.13.0-rc2, build 1f9b3ef

$ stack --docker image container
Running /usr/local/bin/docker inspect fpco/stack-build:lts-7.4 exited with ExitFailure 1

[]

Error: No such object: fpco/stack-build:lts-7.4
```

deinspanjer:
Hmm..  I remember seeing something in the stack docs about needing to do a pull first but for some reason I didn't seem to need to do it.  Maybe I did it but forgot?

Try the following:
`stack docker pull`
`stack --docker build`

jhr007:
No problems stand out. Complements/praise on the coverage of the wrapper script, comments.  

I didn't know about the [reasons behind](https://www.fpcomplete.com/blog/2016/10/docker-demons-pid1-orphans-zombies-signals) `ENTRYPOINT: ["/sbin/pid1"]` so link for future readers. 

Please don't read this next line as "What were you thinking?!" 😅 

But to better understand this PR, what problem were you focused on solving while making these changes? 

I fear the rest of this comment is at best off topic with the problem you're solving, at worst, [bikeshedding](http://bikeshed.com/), either way don't let me hold this back. 

> The begriffs/postgrest-empty container needs to be built at least once. It only needs to be rebuilt when something changes in it.

[Cache invalidation](http://martinfowler.com/bliki/TwoHardThings.html) is hard. I read the above quote to mean that mental space will be needed later to remember and determine if the `begriffs/postgrest-empty` image is stale and needs updating, but I also noticed your point: 

> We could add scripts to the build tree to build the postgrest-empty image and run the build

I'd like to see this in the PR so it doesn't have to be remembered by the maintainer during a context switch between projects, but don't go recoding it just yet. (It's not my project :smile: )

I had a cache issue in my build as well, if the docker.tar exists load it and don't rebuild. But the only way to invalidate it is to click the 'build without cache' on CircleCI. 

I don't think dockerizing the builds adds that much value to the CircleCI process. Circle's build environment is already isolated between builds. I went down the build-in-docker road when I started with my changes, and found them to be less performant. Ultimately switching back because I didn't trust myself to fix the issues I ran into with the tests(see below). 

Dockerizing the the build on a developer's machine does add value by preventing the it-does/doesn't-work-for-me problem. 

I was "rebuilding without cache" on two branches on circleci to compare the invalidated cache build times, but jhr_circle_ci(build in docker) failed. https://circleci.com/gh/jhr007/postgrest the jhr_circleci_docker branch is the build-in-docker). 

FWIW, I had an issue where the tests running in the docker wouldn't break the circle ci build because the docker command would exit normally. Same goes for when I moved the tests in a single bash script. I also ran into a wall getting Postgres inside docker by sharing Circle's Postgres instance inside.  At the time I didn't explore solving these issues too deeply(felt like a huge yak to shave) and just returned to the everything-in-circle.yaml approach.

deinspanjer:
> Please don't read this next line as "What were you thinking?!" 😅
Okay.. :)
 
> But to better understand this PR, what problem were you focused on solving while making these changes?
My company is building a multi-tenant SaaS analytics application.  We are using Postgrest for the API layer.  We have several instances of Postgrest, at least one per tenant, potentially more when we need to support multiple schema versions simultaneously.  We are deploying to Kubernetes.

As such, having a solid Docker image for Postgrest is important to me.  Having something that can easily be built by a developer who is forking or that builds a nightly image is also very beneficial.


> > The begriffs/postgrest-empty container needs to be built at least once. It only needs to be rebuilt when something changes in it.
> Cache invalidation is hard. I read the above quote to mean that mental space will be needed later to remember and determine if the begriffs/postgrest-empty image is stale and needs updating, but I also noticed your point: 
> > We could add scripts to the build tree to build the postgrest-empty image and run the build
> I'd like to see this in the PR so it doesn't have to be remembered by the maintainer during a context switch between projects, but don't go recoding it just yet. (It's not my project 😄 )

The postgrest-empty is appropriately named.  It is just a simple reference image in which to plop the built binary.  That is more of an artifact of how stack integrates Docker than anything else.

Rather than rebuilding it from scratch on every commit which would trigger the re-fetching of the apt-get and such, I figured it would be something better refreshed around beta or release candidate time just to ensure that there aren't any gotchas with some system package that gets updated.  I'm open to ideas though.

> I don't think dockerizing the builds adds that much value to the CircleCI process. Circle's build environment is already isolated between builds. I went down the build-in-docker road when I started with my changes, and found them to be less performant. Ultimately switching back because I didn't trust myself to fix the issues I ran into with the tests(see below).

I agree.  Since the entire project isn't microservice-centric, it isn't a requirement to give Docker CI building priority treatment.

> Dockerizing the the build on a developer's machine does add value by preventing the it-does/doesn't-work-for-me problem.

This one is definitely a big plus for people who aren't that familiar with Haskell in general.

deinspanjer:
So do you think this can go in with the 4.0 items?

begriffs:
Improving the Docker situation for deployment is rad, and I appreciate your work on this. I'd like to focus on getting 0.4 released before merging this though. There are a lot of big changes at once, so let's introduce them a little more gradually. We can get to this PR shortly thereafter.

themightychris:
I'm not sure if this is helpful for this effort or there is a better place to share, but I've been building postgrest with [habitat](https://www.habitat.sh/), which can be used to build docker containers: https://github.com/JarvusInnovations/habitat-plans/tree/master/postgrest

begriffs:
@themightychris that looks interesting. The `postgrest/plan.sh` file is pretty clean too. Let's see what we can do with straight Docker first though. If it ends up being too painful then we can try a less mainstream containerization system.

begriffs:
@deinspanjer OK, I gave this another try tonight and `stack --docker build` works fine on my machine. Want to rebase this on master and update for version 0.4?

> If we had a canonical config file such as postgrest.conf.dist or something else that described all the options, then I could modify the wrapper script to take it into account.

Now postgrest uses a config file rather than command line arguments as documented in https://postgrest.readthedocs.io/en/v0.4/admin.html#configuration

deinspanjer:
I understand that postgrest is using a config file now, but config files aren't as easy to work with inside a docker container as commandline arguments and environment variables, so I wrote the env script dockerfiles/postgrest-env which can take either commandline args or ENV variables and write them into a postgrest.conf file which is then used by the postgrest binary.

My comment was that it would be nice to have a canonical config file such as postgrest.conf.dist or something else that described all the options.  If such a file existed, it would be useful for anyone using postgrest outside Docker, but I could also modify this wrapper script to use it to perform validation and such.

begriffs:
Can I get your opinion about #808?

begriffs:
Now that we found a simplified way to make a Dockerfile with config that interpolates environment variables, how about reducing this pull request to only the Docker details in stack.yaml? I'd say maybe fewer comments in that file too, but you could include a link to the Stack docs about the options as a comment.

deinspanjer:
Sure thing.

deinspanjer:
Working on a v2 of the PR in a new branch.

Issue #753, opened 2016-11-28
return=representation returns singular object if you insert an array with 1 element


yuhe00:
The following works as expected:
```
echo '[{"name":"Hyundai"},{"name":"Mazda"}]' | http POST localhost:3337/makes "Prefer:return=representation"
HTTP/1.1 201 Created
Content-Encoding: gzip
Content-Type: application/json; charset=utf-8
Date: Mon, 28 Nov 2016 09:58:49 GMT
Server: postgrest/0.3.2.0
Transfer-Encoding: chunked

[
    {
        "id": "f4a99b15-2144-4563-8b5a-41a1456fb692", 
        "name": "Hyundai", 
        "status": "UNPUBLISHED"
    }, 
    {
        "id": "21205052-77a8-4e1c-b78d-a49c42c41307", 
        "name": "Mazda", 
        "status": "UNPUBLISHED"
    }
]
```
...while the following returns a single object:
```
> echo '[{"name":"Hyundai"}]' | http POST localhost:3337/makes "Prefer:return=representation"
HTTP/1.1 201 Created
Content-Encoding: gzip
Content-Type: application/json; charset=utf-8
Date: Mon, 28 Nov 2016 09:52:20 GMT
Location: /makes?id=eq.f6adee2d-c2a2-4bf9-afbb-d8ecc6ead2a6
Server: postgrest/0.3.2.0
Transfer-Encoding: chunked

{
    "id": "f6adee2d-c2a2-4bf9-afbb-d8ecc6ead2a6", 
    "name": "Hyundai", 
    "status": "UNPUBLISHED"
}
```
Not sure if this is intended behavior, but it was a bit unexpected. I think it should return an array if you send an array, no matter the amount of elements in the array.

ruslantalpa:
This is happening [here](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/App.hs#L136), not sure why that was decided

ruslantalpa:
i think this is just a remanence from the fact that you can post like this
`{"name":"Hyundai"}`  and also like this `[{"name":"Hyundai"},{"name":"Mazda"}]`
we'd need to look in the old issues to see if it was an oversight or there is a specific reason

NikolayS:
Hi, we have the same issue.

[...]

PR #763, opened 2016-12-07
Use a mime type to request singular JSON responses

### LINKED ISSUE #748: Output error when doing GET with plurality=singular and more than one row is returned
postgREST current behavior is to do a ```LIMIT 1 OFFSET 0``` when doing a GET with ```plurality=singular```, I think it will be more appropriate to give an error when more than one row would be returned.

This would help with client errors, for example a while back I got a case of a duplicate key and postgREST could have helped me there since I mostly use ```plurality=singular``` when explicitly requesting a single row with filters or a limit.

Also this would be a more consistent behavior since POST and PATCH already give an error when returning more than one mutated row.

begriffs:
Fixes #748

This PR will remove support for `Prefer: plurality=singular` and add support for a special mime type in the Accept header that is like JSON except it encodes objects only, not arrays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use specific columns in the RETURNING section - @ruslantalpa
 
 ### Changed
-- Replace `Prefer: plurality=singular` with custom mime type - @begriffs
+- Replace `Prefer: plurality=singular` with `Accept: application/vnd.pgrst.object` - @begriffs
 - Standardize arrays in responses for `Prefer: return=representation` - @begriffs
 - Calling unknown RPC gives 404, not 400 - @begriffs
 - Use HTTP 400 for raise\_exception - @begriffs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Allow test database to be on another host - @dsimunic
-- New `Prefer` header value: `params=single-object` to pass all form values as a single json object to a stored procedure - @dsimunic
+- `Prefer: params=single-object` to treat payload as single json argument in RPC - @dsimunic
 - Ability to generate an OpenAPI spec - @mainx07, @hudayou, @ruslantalpa, @begriffs
 - Ability to generate an OpenAPI spec behind a proxy - @hudayou
 - Ability to set addresses to listen on - @hudayou
@@ -30,6 +30,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use specific columns in the RETURNING section - @ruslantalpa
 
 ### Changed
+- Replace `Prefer: plurality=singular` with custom mime type - @begriffs
+- Standardize arrays in responses for `Prefer: return=representation` - @begriffs
+- Calling unknown RPC gives 404, not 400 - @begriffs
 - Use HTTP 400 for raise\_exception - @begriffs
 - Remove non-OpenAPI schema description - @begriffs
 - Use comma rather than semicolon to separate Prefer header values - @begriffs
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -38,7 +38,7 @@ import           Data.Ranged.Boundaries
 import           PostgREST.Types           (QualifiedIdentifier (..),
                                             Schema,
                                             PayloadJSON(..))
-import           Data.Ranged.Ranges        (Range(..), singletonRange, rangeIntersection, emptyRange)
+import           Data.Ranged.Ranges        (Range(..), rangeIntersection, emptyRange)
 
 type RequestBody = BL.ByteString
 
@@ -98,8 +98,6 @@ data ApiRequest = ApiRequest {
   , iPayload :: Maybe PayloadJSON
   -- | If client wants created items echoed back
   , iPreferRepresentation :: PreferRepresentation
-  -- | If client wants first row as raw object
-  , iPreferSingular :: Bool
   -- | Pass all parameters as a single json object to a stored procedure
   , iPreferSingleObjectParameter :: Bool
   -- | Whether the client wants a result count (slower)
@@ -130,9 +128,8 @@ userApiRequest schema req reqBody
         map decodeContentType . parseHttpAccept <$> lookupHeader "accept"
       , iPayload = relevantPayload
       , iPreferRepresentation = representation
-      , iPreferSingular = singular
       , iPreferSingleObjectParameter = singleObject
-      , iPreferCount = not singular && hasPrefer "count=exact"
+      , iPreferCount = hasPrefer "count=exact"
       , iFilters = [ (toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn ["order", "limit", "offset"] k) ]
       , iSelect = toS $ fromMaybe "*" $ fromMaybe (Just "*") $ lookup "select" qParams
       , iOrder = [(toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, endingIn ["order"] k ]
@@ -194,7 +191,6 @@ userApiRequest schema req reqBody
     where
         split :: BS.ByteString -> [Text]
         split = map T.strip . T.split (==',') . toS
-  singular        = hasPrefer "plurality=singular"
   singleObject    = hasPrefer "params=single-object"
   representation
     | hasPrefer "return=representation" = Full
@@ -208,7 +204,7 @@ userApiRequest schema req reqBody
   endingIn xx key = lastWord `elem` xx
     where lastWord = last $ T.split (=='.') key
 
-  headerRange = if singular && method == "GET" then singletonRange 0 else rangeRequested hdrs
+  headerRange = rangeRequested hdrs
   replaceLast x s = T.intercalate "." $ L.init (T.split (=='.') s) ++ [x]
   limitParams :: M.HashMap ByteString NonnegRange
   limitParams  = M.fromList [(toS (replaceLast "limit" k), restrictRange (readMaybe =<< (toS <$> v)) allRange) | (k,v) <- qParams, isJust v, endingIn ["limit"] k]
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -59,6 +59,7 @@ data PreferRepresentation = Full | HeadersOnly | None deriving Eq
                           --
 -- | Enumeration of currently supported response content types
 data ContentType = CTApplicationJSON | CTTextCSV | CTOpenAPI
+                 | CTSingularJSON
                  | CTAny | CTOther BS.ByteString deriving Eq
 
 data ApiRequestError = ErrorActionInappropriate
@@ -75,6 +76,7 @@ toMime :: ContentType -> ByteString
 toMime CTApplicationJSON = "application/json"
 toMime CTTextCSV         = "text/csv"
 toMime CTOpenAPI         = "application/openapi+json"
+toMime CTSingularJSON    = "application/vnd.pgrst.object+json"
 toMime CTAny             = "*/*"
 toMime (CTOther ct)      = ct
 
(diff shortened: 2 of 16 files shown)


ruslantalpa:
Why did you remove all those tests?
They should have just been converted to send the special accept header instead of the prefer.
Maybe it's not ready yet but i was expecting changes only on the top level (ApiRequest) that changes reading `singular` from accept header instead of prefer and not deep changes to the logic ...

begriffs:
This is still a work in progress and I will ensure that the new functionality will be fully tested.

Patience, my son. 🙏

ruslantalpa:
(how do you get those cool emojies :))
ok, i'll wait :) just sayin' ... i think you are doing more work then needed so i was just giving you a heads up before you spend more time on it.
basically i think of this feature as this (the way to implement it), maybe you have a better way:

- convert all the tests that send prefer header, to send the accept header (and have them fail at first) but the result/body of the test should not be altered.
- edit the ApiRequest to read the preferSingular var from the accept header and not the prefer header (maybe change the var name along with that)
- change the App code for the GET path to issue a error when more then one items are returned.

this way i think is way less code (no need for new tests) and a bit safer in terms of breaking existing functionality then going the way of "remove completely the functionality then add it back together"

begriffs:
That's a good point. I feel like I responded poorly above so sorry about that. I interpreted your first suggestion as a lack of trust that I would adequately test the code, but you're just pointing out that it's more efficient to convert the tests (allowing them to fail for a while) and then add the new feature. I agree and will do it that way.

As for the emojis, get ready for all the possibilities: http://www.webpagefx.com/tools/emoji-cheat-sheet/

ruslantalpa:
no need to be sorry :) i can see how my wording would make you believe that, so that is my fault (sorry).
These things might tend to happen in the future with my comments, so whenever you see a comment from me like the one above ( or like for example the one in gitter ... "don't do that") most of the time you should read it like:
- it's a second language so the `proper tone` might be lost in translation
- it's never a demand but a request to reconsider/think a bit
- i usually don't put a lot of thought into `sounding right`, i just start typing what i am thinking since i don't think i need some kind of filter as long as i am criticising the code/idea and not the person/his abilities .. just too old to waste time on that :)

PS: and i don't get offended that easily, and always feel free to argue as long as necessary if yo think my idea is bad or your way is better, i can be convinced to change my mind. :)

src/PostgREST/ApiRequest.hs
```diff
@@ -75,6 +76,7 @@ toMime :: ContentType -> ByteString
 toMime CTApplicationJSON = "application/json"
 toMime CTTextCSV         = "text/csv"
 toMime CTOpenAPI         = "application/openapi+json"
+toMime CTSingularJSON    = "application/vnd.pgrst.object+json"
```
ruslantalpa: don't know the specs, can it be `application/json+vnd.pgrst.object`

test/Feature/SingularSpec.hs
```diff
@@ -0,0 +1,80 @@
...
+        _ <- post "/addresses" [json| { id: 99, address: "yyy" } |]
+        p <- request methodPatch
+          "/addresses?id=gt.0"
+          [("Prefer", "return=representation"), singular]
+          [json| { address: "zzz" } |]
+        liftIO $ simpleStatus p `shouldBe` status400
+
+    context "when creating rows" $ do
+
+      it "works for one row" $ do
```
ruslantalpa: Maybe this too should be an error.
If you send "singular" then don't put your payload in an array.

ruslantalpa:
Overall it looks good, maybe one more pass to see if all the previous deleted tests have an equivalent in the new spec file.

begriffs:
I also want to add a few more tests to make sure the array/object/blank body behavior is consistent across the verbs.

* Always array for empty set and singleton results in json
  * Always empty bodies when no return=rep
  * Otherwise, [] [a], [a,b..]
    * inserting zero, single or multiple items
    * deleting no, one, or some rows
    * updating no, one, or some rows
  * Updates/deletes that match no rows should return 404
* With vnd.pgrst.object with return=rep
  * All verbs: multiple rows must return 406
  * No matching rows continues to return 404, blank body

begriffs:
:drum: Whew, I think this is ready for review. There are three main changes in here

* Using `application/vnd.pgrst.object` rather than the prefer header to request singular object representations
* Standardizing the use of arrays in responses (Fixes #753)
* Changes in HTTP status codes for some cases

There also are some internal testing and code quality changes not reflected in the changelog:

* Remove unused parameter to `createWriteStatement` and refactore it to avoid complicated pattern matching
* Avoid parsing/reserializing RPC bodies
* Consolidate tests for the singular response cases
  * Sadly these tests, like others in our test suite, rely on lots of implicit knowledge of the fixture. This is a big area for a refactor some day.
  * I use `ROLLBACK` to cancel requests after a singularity problem is detected, and it works fine except the test runner output is now littered by `WARNING:  there is no transaction in progress` because my Hasql transaction still calls `COMMIT` after the roll back.
* Fixes and regression tests for when error messages are not in standard JSON format (not in changelog because it was a recent bug not in the previous release)

Here are aspects of the new behavior that may be surprising or worth discussing

* If you try updating or deleting `/unknownroute` you'll get a 404, but `/knownroute?condition=eq.impossible` will not return 404. The idea is that the thing you are filtering is a known route, but your conditions may be such that no rows get updated.
  * To detect that no rows were updated, the client must either request `return=rep` and look for the `[]` response, or check the `Content-Range` response header. This header is of the form `X/Y` where `X` will be `*` when no rows are updated, or a numerical range when some are.
* RPCs returning simple values like 2, or "hello" will succeed under `application/vnd.pgrst.object` even though that response is not a json object. It is easier in the code to allow these cases. Is this cool (in that these are "singular" responses) or do we want to go the extra mile to make it reject all non-objects?
* `Prefer: return=minimal` (or non prefer=rep really) makes `Accept: application/vnd.pgrst.object` always work. We're treating a blank body as a valid singular response. It's kind of convenient for the client, but maybe points to the name of our content type as misleading. It's not so much object as it is non-array.

Dear god I want to finish this and get the new release out the door. I do intend to be thorough and careful though so please help me review these changes!

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use specific columns in the RETURNING section - @ruslantalpa
 
 ### Changed
+- Replace `Prefer: plurality=singular` with custom mime type - @begriffs
```
ruslantalpa: specify the exact mime type here

src/PostgREST/App.hs
```diff
@@ -109,96 +106,104 @@ app dbStructure conf apiRequest =
...
-              return . responseLBS status201 headers $
-                if iPreferRepresentation apiRequest == Full
-                  then toS body else ""
-
-        (ActionUpdate, TargetIdent qi, Just payload) ->
+              if contentType == CTSingularJSON
+                 && not isSingle
+                 && iPreferRepresentation apiRequest == Full
+               then do
+                  HT.sql "ROLLBACK;"
```
ruslantalpa: Here we are in the context of a hasql transaction abstraction, if you do rollback here, won't the abstraction try to do a `commit` after this thing and maybe "poison" the connection?
Is this this thing tested?
I mean, the rollback might succeed, but will the following request/transaction succeed?

to reiterate, this is what i think will happen on the postgresql side, will it work without issues?

```sql
-- first request
begin; --hasql transaction
main query;
rollback; -- the line above
commit; --hasql transaction

-- second request
begin; -- is the connection in a broken state here because of the rollback followed by a commit?
...
```

src/PostgREST/App.hs
```diff
@@ -318,6 +327,17 @@ contentRangeH lower upper total =
       totalNotZero  = fromMaybe True ((/=) 0 <$> total)
       fromInRange   = lower <= upper
 
+singularityError :: Integer -> Response
```
ruslantalpa: should this be moved to Error.hs?

src/PostgREST/App.hs
```diff
@@ -326,15 +346,15 @@ formatParserError e = formatGeneralError message details
        $ showErrorMessages "or" "unknown parse error" "expecting" "unexpected" "end of input" (errorMessages e)
 
 formatGeneralError :: Text -> Text -> Text
-formatGeneralError message details = message <> ", " <> details
+formatGeneralError message details = toS . encode $
```
ruslantalpa: should this be moved to Error.hs

src/PostgREST/App.hs
ruslantalpa: i see you mentioned this in your last comment.
Unless there is a way to force hasql transaction abstraction to do a rollback without doing this with a raise exception, the previous way was better imo.

ruslantalpa:
On first look and after reading your long comment the only thing i do not like is the rollback-commit thing.
The other thing worth discussing is the 200 on delete where no rows were affected, no opinion about this just seems worth thinking a bit more about this and the usecases

src/PostgREST/App.hs
ruslantalpa: looking haql transaction
if we somehow could return a proper `Left Error` then the abstraction would to the rollback itself
the status is checked here
https://github.com/nikita-volkov/hasql-transaction/blob/master/library/Hasql/Transaction/Private/Transaction.hs#L63
and
https://github.com/nikita-volkov/hasql/blob/master/library/Hasql/Session.hs
seems to export some error type constructors

jackfirth:
> The other thing worth discussing is the 200 on delete where no rows were affected, no opinion about this just seems worth thinking a bit more about this and the usecases

Just chiming in from the http standards side of things: returning 200 on no-op delete plays nicely with ETags and optimistic locking headers. You can use the etag header `If-Match` to explicitly say a request should only succeed if the current representation matches some etag, so you can mark certain delete requests as "only perform this delete if the resource is in this state before deletion". This lets clients opt-in to different response codes for normal delete vs. no-op delete. It's even possible to implement this in a proxy on top of postgrest; the delete request would first trigger a GET from the proxy to postgrest which checks that the resource to delete exists and then calculates an etag.

begriffs:
@ruslantalpa I experimented with finding a way to get Hasql's Transaction to rollback rather than commit. One way is to actually generate a database error or exception, but then the application cannot return the custom 406 http response since execution jumps to our database error handler which generates a generic response.

I think the `Left error` Hasql code you linked to deals with Either only because `tryError` above it converts Hasql exceptions to Either for easier processing. So the only thing that triggers the Left, if I understand correctly, is exceptions due to database or connection errors. I don't know how to make our own code trigger the rollback in the absence of causing an intentional db error.

However we might be OK to leave in the current `rollback; commit;` behavior. I set `db-pool = 1` in the postgrest config and verified that I can cause the rollback and then execute another request successfully. It causes the warning in the db logs, but does not poison future transactions on the connection that I can see. I will try to encode this check for connection poisoning into the test suite.

@jackfirth that's good to know, glad our return code coincides with good http practice for ETags reasons.

ruslantalpa:
@begriffs 
Probably it's not possible to have our own code return an error, i was thinking somehow wrapping this `HS.ResultError (HS.ServerError "a" "b" Nothing Nothing)` and returning it without actually going to the db (maybe @nikita-volkov can tell us if it's possible to trigger an error within a transaction without going to the db)

As for your comment about doing it the old way (going to the db).
I don't see why you can not control the response code if the flow ends up in Error.hs
Since you are the one generating the error (with a hardcoded query), you know the db error code, and the message body that is going to come back from the database, so knowing that, you can request more detail here (by replacing _ with a variable)
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Error.hs#L81
```
httpStatus authed (P.SessionError (H.ResultError (H.ServerError c _ _ _)))
```
 to get the message body
then on this line 
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Error.hs#L104
```
'P':'0':_ -> HT.status500
```
 instead of returning the status code directly, have an if that checks the message body and if it's the exact string you sent, then return something else then status500

It's not a haskell oneliner and matches on a string (i get that it's a bit of a hack) but at least it's not generating bad queries.
I get that the connection is not poisoned and it works but the db is telling us (with the warning) that we are doing something wrong and i for one would like to not intentionally generating warnings at the db level.

begriffs:
(As an update for those of you following this thread, the conversation is ongoing in hasql-transaction about  the best way to handle rollbacks.)

PR #767, opened 2016-12-08
Fix embeded relations error message

diogob:
The `formatGeneralError` function was being used in the same context as other plain text error messages. It seems to me that we don't need that JSON formatting for those error messages.

If we would keep the message/details structure for that message we would need a more extensive code change.

I also fixed a bug with our `errResponse` that would generate messages using text concatenation which could create an invalid json body.


begriffs:
Thanks for the fix. It's no problem that we drop the details here, just as long as it's JSON with a message key, which it is.

PR #768, opened 2016-12-09
Fix #718

### LINKED ISSUE #718: Switch to normal query from prepared statements on GET requets
On this line https://github.com/begriffs/postgrest/blob/master/src/PostgREST/App.hs#L112 we run a prepared statement that has no parameters, everything is hardcoded (for now) in the actual query string.

This creates a suboptimal case for two reasons.
- hasql (probably) calculates a hash of the statement
- postgresql treats this a prepared statement and caches it

The problem with this is that when a server is "hammered" with a lot of requests that seem to be the same thing (only value for filters change) are actually considered different prepared statements and are cached separately thus consuming memory on the database server.
What's worse is that each connection maintains it's own cache thus making the problem worse (imagine 10k useless prepared statements cached per connection, it adds up quite fast).
Until the generated query will be a true prepared statement, we should change how we run queries.

ruslantalpa:
1fmtis


begriffs:
Haha, nice.

Think it merits a note in the changelog? Like, "Prevent space leak of prepared statements" or something.

PR #769, opened 2016-12-09
Fix #746

### LINKED ISSUE #746: Returning a subset of the full object from an INSERT
Is there anyway to do this kind of statement in Postgrest?

```sql
INSERT INTO "1".users (firstName,lastName,password)
VALUES ("mark", "jacob", "123")
RETURNING firstName, lastName
```

I tried the `Prefer: return=firstName,lastName` but that doesn't seem to work correctly on 0.3.2 anyway.

For context, this is useful when you want to lock down selects for an anon user but still allow them to create users (and not expose all your columns)

ruslantalpa:
This fixes the insert problem in a table to which the user does not have select access to all the columns.
Took way longer then expected


begriffs:
I appreciate your work on this. To be honest one reason it might have taken longer than you expected is because the code in `QueryBuilder` has gotten crazy. The functions have evolved to be poorly decomposed, having incomplete pattern matches and frequently ignoring a majority of their arguments in various cases.

I believe we should not continue to complicate this file. We should get the release out the door, and then aggressively refactor QueryBuilder. Once that file has clean code then we add more features or fixes such as the one in this pull request.

ruslantalpa:
You prefer to release with a known bug instead merging code that works?
The refactoring will not involve only that file but the whole core starting from the types, and that is not going to come any time soon

ruslantalpa:
Another way to put it, with the current core types, this is the best thing possible

begriffs:
I see your point, it does fix a known bug. OK we can merge it but for real we have to refactor that file before it keeps getting more complex. I'll give the refactor a try after the release.

begriffs:
Can you tell me about the `create_test_db` and `SpecHelper` changes? What are those about at a high level?

ruslantalpa:
about the refactoring:
Sure it's needed, code has gotten quite ugly, but it works. Everything evolved gradually (select was introduced step by step to each path) and the original types no longer contain all the information needed in order to have nicer code.
The refactoring needs to be a lot deeper then you think, touching core types, which means almost every file in the project will require changes/rewrite. For me personally a small refactor of the QueryBuilder does not make since this work will be thrown away when the big refactor happens (whenever that may be)

about test env changes:

After the last couple of commits the tests ran correctly only on a
fresh db, if you run the second time with `stack tests`, a couple of tests will fail (try it).
In addition, the tables/views/roles within the db were not created/dropped
on each execution, only on the creating of the db with that shell script but we need a clean db created from scratch on each execution.

diogob:
+1 for refactoring.
We should require constant refactor with any feature work on these files. I believe even small refactors could do a lot of help.

A safe way to start this work is by redesigning signatures for the QueryBuilder module improving the input types and write translation functions that could be composed without changing any core types. Therefore we would be able to design a clean API without breaking any code, and defer the bigger refactor of types for a second step.

For instance the function 
```haskell
returningF :: Target -> PreferRepresentation -> DbRequest -> SqlFragment
```
has too many responsibilities. 

It signature probably could be something like:
```haskell
returningF :: Target -> SelectColumns -> FkColumns -> SqlFragment
```

The type `PreferRepresentation` seems to be closer to the HTTP side than to the database side, so maybe another function could decide if we need to use `returningF` eliminating one input parameter and one branch for this function.

ruslantalpa:
I am not against the small refactor (and for the big one, in time), but i am not very motivated to spend time on the small refactors of the current code since i think the big one will mostly remove all the current code, kind of like what happened when going from 2 to 3.

But if someone wants to ... :)

begriffs:
More background on why code cleanliness is important to me, lest I seem like a fanatic. This project serves as an ambassador for the Haskell language. People use it who never have any contact with the language. During this first introduction I worry people will look at the code and say, "What is this nonsense, give me back my Javascript!"

Clear code is also important for getting new contributors. (In general I'd rather use dumb but clear code than take advantage of fancy Haskell features.) The QueryBuilder stuff confuses *me*, so it must be a big barrier for new contributions in that part of the codebase. I'm not trying to blame anybody for how the code is, it just evolved to handle a lot of situations. Luckily the tests are good and are insulated from the code behind simple HTTP calls so we're in good shape to refactor.

All this to say: a refactor doesn't have to happen right this moment, but it should soon have high priority because the problem is only getting worse.

___

Also @dsimunic do the testing changes in this PR match what you have in mind for re-running tests? Any opinion here?

dsimunic:
@ruslantalpa why did you want to bring back the database creation in the test code? (function `setupDb`). I remember you were rightly advising me to avoid doing that.

Can you post the repro for repeated run problems?

ruslantalpa:
i am not creating the database on each test execution, i am only dropping/recreating all the entities inside the database, not just on the shell script execution

the tests fail if you do it like this

```
export POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://ruslantalpa@localhost" postgrest_test)
stack test (ok)
stack test (fail)
```

entities inside the database (tables/views/users) need to be dropped/created on each test execution because the workflow is:
- i have setup my test db by running once `export PO...`
- from now on i only want to do `stack test`
- now i am developing a new feature for which i need an additional table/view/role
- as things are now, i am not able to only add that object to the sql files and run stack test, i have to re-execute your shell script (which will drop the db and recreate everything, assuming i run it with the exact same parameters)

dsimunic:
So far the assumption was that one always runs the db script before running tests, but you are absolutely right that it makes sense to run tests repeatedly while you are developing new features or refactoring.

What I don't like about the change is that we are tying again database setup with the test code. My longer term vision is that these create scripts are used to set up a starter database for demo purposes, not just for testing. Sadly, I only planned to work on that next year.

The best would be to ensure that every call to `resetDb` brings the database to the initial state. Ideally, we'd create such a test-supporting function in the database, so that the test code does not have to know about the location of sql files at all.

Why do you need to drop and recreate all schemas and tables?

begriffs:
Some day I'd like to refactor the test suite so that each test creates and populates the db schema in its own transaction which gets rolled back between each test. That way we can avoid having a big fixture that developers have to cross-reference.

Until that day... is the testing behavior introduced into this PR good to merge for 0.4, or any tweaks left to clean it up?

ruslantalpa:
no tweaks left, and actually there is no new behaviour introduced, the test function the way they used to before with the exception that the database is not dropped on each run, only the objects within the db.

begriffs:
Tested this locally and the tests pass even after repeated runs which is good. However at the top of each run I see:

```
psql:test/fixtures/roles.sql:2: ERROR:  2BP01: role "postgrest_test_anonymous" cannot be dropped because some objects depend on it
DETAIL:  42 objects in database postgrest_test
42 objects in database funfun
LOCATION:  DropRole, user.c:1045
psql:test/fixtures/roles.sql:3: ERROR:  42710: role "postgrest_test_anonymous" already exists
LOCATION:  CreateRole, user.c:337
psql:test/fixtures/roles.sql:4: ERROR:  42710: role "postgrest_test_default_role" already exists
LOCATION:  CreateRole, user.c:337
psql:test/fixtures/roles.sql:5: ERROR:  42710: role "postgrest_test_author" already exists
LOCATION:  CreateRole, user.c:337
```

Can you adjust this PR to remove that error?

ruslantalpa:
That is because you have your old postgrest_test database which uses the same roles so the error you are seeing is because those roles are used in two different databases.

Delete your old "postgrest_test" database and let the script create the new one

begriffs:
That was exactly the problem. I had a number of test databases that needed to be dropped.

Note that if I run `test/create_test_db` more than once with two different db names then the warnings will come back. That's not the intended use of that script though. Something we may improve later.

dsimunic:
The warnings come back because the test user is hard-coded again, and the user is shared for the whole server. 

The correct behavior is to create a random user name for each new database, and with a random password to increase security in case someone uses non-ephemeral server and doesn’t drop the test db/user after running tests.

begriffs:
The tests are passing OK on CI and it works well locally when I re-use the same test db, so how about I merge this PR. @dsimunic if you want to improve it with a random user that's fine by me, as long as calling `stack test` repeatedly or `stack test --file-watch` keeps working smoothly.

dsimunic:
I’ll take a look next year.

ruslantalpa:
the error @begriffs described will not go away if you make the authenticator role random.
there is no additional security if the username is random since the password is already random but having the authenticator username random means on each execution of the script you will create a new one, thus leaving "junk logins" behind

Issue #660, opened 2016-07-01
JWT invalid with Auth0 tokens


Orbmancer:
Hi !

I'm using the postgres binary build from ubuntu, and I'm trying to plug the external authentication Auth0.

They use a JWT whose sub claim is in the form
```javascript
{
    ...
    sub: "{connection}|{user_id}",
    ...
}
```

The pipe `|` makes postgrest returns a
```javascript
{
  "message": "Invalid JWT"
}
```

I don't know if it comes from postgrest or even higher in the _Web.JWT_ library.

ruslantalpa:
i think is from Web.JWT, PostgREST does not look as `sub`

darkyen:
Adding support for this would be wonderful, Auth0 and Postgrest look like a match made in heaven.

Orbmancer:
The `sub` claim must be in `StringOrURI` format according to the [RFC-7519](https://tools.ietf.org/html/rfc7519#section-4.1.2)
However, there is no mention of particular forbidden character for the String.

ruslantalpa:
Have not tested this a all, just glanced how the JWT lib works,
the main logic for this issue i think is here
https://bitbucket.org/ssaasen/haskell-jwt/src/c8b79eda623fb16061a19fa0196a9a528dcb523f/src/Web/JWT.hs?at=master&fileviewer=file-view-default#JWT.hs-495
```
instance FromJSON StringOrURI where
    parseJSON (String s) | URI.isURI $ T.unpack s = return $ U $ fromMaybe URI.nullURI $ URI.parseURI $ T.unpack s
    parseJSON (String s)                          = return $ S s
    parseJSON _                                   = mzero
```

so if `URI.isURI $ T.unpack s` would return true but `URI.parseURI $ T.unpack s` would actually fail, then i think the problem will surface.

@Orbmancer care to make a small test and check if the logic is this line works?
`parseJSON (String s) | URI.isURI $ T.unpack s = return $ U $ fromMaybe URI.nullURI $ URI.parseURI `

or maybe make a small test and see if Web.JWT fails to parse your token (maybe somehow i missed it and the problem is in postgrest)?
i.e.
a) test web.jwt fails to parse your token
b) if a is true (it fails), then test if the logic in `isURI / parseUri` is correct

Orbmancer:
I'm on it 👍

darkyen:
[...]

PR #772, opened 2016-12-10
Binary secrets working and tested

TrevorBasinger:
Backstory: Auth0 generates binary secrets and gives the user a base64URL encoded value to use. I was decoding that payload into a file and then trying to run postgrest with `jwt-secret = "@path/to/key"`. When postgrest would read that key file, it would read the contents as a string `[Char]`, convert to `Text`, and attempt to decode the jwt using this bastardized secret.

My solution was to have postgrest optionally accept base64 encoded secrets. All secrets are then converted to a `ByteString` and then decoded with the `binarySecret` function included in the jwt library.

The one "gotcha" is that if a service like Auth0 gives you a base64URL encoded value, you would need to convert* that into a base64 value when you insert that into your configuration file or `@secretfile`.

Auth0 is working great now for me.

I'd also be happy to write the supporting docs for Auth0. It requires you to write a rule to add the role into the scope of the token.

#660 and #495 also elaborate on parts of this.

* Replace `-` with `+` and `_` with `/` as referenced [here](https://en.wikipedia.org/wiki/Base64#URL_applications)

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -17,7 +17,7 @@ import           Control.AutoUpdate
 import           Data.ByteString.Base64               (decode)
 import           Data.String                          (IsString (..))
 import           Data.Text                            (stripPrefix, pack)
-import           Data.Text.Encoding                   (encodeUtf8)
+import           Data.Text.Encoding                   (encodeUtf8, decodeUtf8)
 import           Data.Text.IO                         (hPutStrLn, readFile)
 import           Data.Function                        (id)
 import           Data.Time.Clock.POSIX                (getPOSIXTime)
@@ -103,7 +103,7 @@ main = do
 loadSecretFile :: AppConfig -> IO AppConfig
 loadSecretFile conf = extractAndTransform mSecret
   where
-      mSecret   = configJwtSecretOrFile   conf
+      mSecret   = decodeUtf8 <$> configJwtSecret conf
       isB64     = configJwtSecretIsBase64 conf
 
       extractAndTransform :: Maybe Text -> IO AppConfig
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -109,9 +109,10 @@ loadSecretFile conf = extractAndTransform mSecret
     extractAndTransform :: Maybe Text -> IO AppConfig
     extractAndTransform Nothing  = return conf
     extractAndTransform (Just s) =
-      case stripPrefix "@" s of
-        Nothing       -> setSecret <$> transformString isB64 s
-        Just filename -> fmap setSecret $ transformString isB64 =<< readFile (toS filename)
+      fmap setSecret $ transformString isB64 =<<
+        case stripPrefix "@" s of
+            Nothing       -> return s
+            Just filename -> readFile (toS filename)
 
     transformString :: Bool -> Text -> IO ByteString
     transformString False t = return . encodeUtf8 $ t
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -14,9 +14,11 @@ import           PostgREST.OpenAPI                    (isMalformedProxyUri)
 import           PostgREST.DbStructure
 
 import           Control.AutoUpdate
+import           Data.ByteString.Base64               (decode)
 import           Data.String                          (IsString (..))
-import           Data.Text                            (stripPrefix)
-import           Data.Text.IO                         (hPutStrLn)
+import           Data.Text                            (stripPrefix, pack)
+import           Data.Text.Encoding                   (encodeUtf8)
+import           Data.Text.IO                         (hPutStrLn, readFile)
 import           Data.Function                        (id)
 import           Data.Time.Clock.POSIX                (getPOSIXTime)
 import qualified Hasql.Query                          as H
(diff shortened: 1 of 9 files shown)


src/PostgREST/Config.hs
```diff
@@ -43,17 +44,19 @@ import           Protolude hiding            (intercalate
...
-  , configMaxRows   :: Maybe Integer
-  , configReqCheck  :: Maybe Text
-  , configQuiet     :: Bool
+    configDatabase          :: Text
+  , configAnonRole          :: Text
+  , configProxyUri          :: Maybe Text
+  , configSchema            :: Text
+  , configHost              :: Text
+  , configPort              :: Int
+  , configJwtSecretOrFile   :: Maybe Text
```
begriffs: Is it possible to do everything with just `configJwtSecret`? After processing, `configJwtSecretOrFile` gets turned into Nothing. Wondering if the secret or file could be stored in `configJwtSecret` and then this data field be overwritten by the properly encoded value?

TrevorBasinger: Yeah... interesting idea. I think that'd work. Let me test...

TrevorBasinger:
@begriffs Great idea. Works fine.

test/SpecHelper.hs
```diff
@@ -54,25 +54,35 @@ getEnvVarWithDefault var def = do
   varValue <- getEnv (toS var) `E.catchIOError` const (return $ toS def)
   return $ toS varValue
 
+_baseCfg :: AppConfig
+_baseCfg =  AppConfig -- Connection Settings
+                      mempty "postgrest_test_anonymous" Nothing "test" "localhost" 3000
```
TrevorBasinger: No worries, it's all a preference thing. I was trying to get the comments to align. What's your preference? Perhaps a comment line between the type sig and definition?

main/Main.hs
```diff
@@ -99,9 +101,24 @@ main = do
...
 loadSecretFile :: AppConfig -> IO AppConfig
-loadSecretFile conf = do
-  let s = configJwtSecret conf
-  real <- case join (stripPrefix "@" <$> s) of
-            Nothing -> return s -- the string is the secret, not a filename
-            Just filename -> sequence . Just $ readFile (toS filename)
-  return conf { configJwtSecret = real }
+loadSecretFile conf = extractAndTransform mSecret
+  where
+      mSecret   = decodeUtf8 <$> configJwtSecret conf
```
begriffs: :ant: indenting by two rather than four under the `where` would match the surrounding style a little better.

test/SpecHelper.hs
begriffs: Maybe

TrevorBasinger:
I think I like this better:
```

_baseCfg :: AppConfig
_baseCfg =  -- Connection Settings
  AppConfig mempty "postgrest_test_anonymous" Nothing "test" "localhost" 3000
            -- Jwt settings
            (Just $ encodeUtf8 "safe") False
            -- Connection Modifiers
            10 Nothing (Just "test.switch_role")
            -- Debug Settings
            True
```

Thoughts?

begriffs:
Thanks for this pull request, it's going to make people's day when they try using Auth0.

Some more ideas:

* For convenience what if postgrest replaced the nonstandard b64 characters (if there are any) after loading the string? Only if configJwtSecretIsBase64 is set of course. Seems like a small thing we can do to accommodate those nonstandard b64 strings to reduce confusion among our users.
* Add a line to the [changelog](https://github.com/begriffs/postgrest/blob/master/CHANGELOG.md) as well please
* I'd appreciate a pull request to update the [configuration](https://postgrest.readthedocs.io/en/v0.4/admin.html) section of the docs. The new docs live on branch v0.4 of https://github.com/begriffs/postgrest-docs
* If you're feeling ambitious I'd *love* your help completing the [Auth0 section](https://postgrest.readthedocs.io/en/v0.4/auth.html#jwt-from-auth0) of the new docs.

TrevorBasinger:
Sounds great. I'll get it done and ping you again in a bit.

begriffs:
(Your idea for the `_baseCfg` change looks fine.)

TrevorBasinger:
@begriffs ping.

Let me know if there's anything I can do.

main/Main.hs
```diff
@@ -99,9 +101,26 @@ main = do
...
-            Just filename -> sequence . Just $ readFile (toS filename)
-  return conf { configJwtSecret = real }
+loadSecretFile conf = extractAndTransform mSecret
+  where
+    mSecret   = decodeUtf8 <$> configJwtSecret conf
+    isB64     = configJwtSecretIsBase64 conf
+
+    extractAndTransform :: Maybe Text -> IO AppConfig
+    extractAndTransform Nothing  = return conf
+    extractAndTransform (Just s) =
```
begriffs: Might be a little simpler to say

```haskell
extractAndTransform (Just s) =
  fmap setSecret $ transformString isB64 =<<
    case stripPrefix "@" s of
      Nothing       -> return s
      Just filename -> readFile (toS filename)
```

begriffs:
Thanks for your great work, tests and docs and everything.

TrevorBasinger:
You're welcome. Fun fact: I was a pupil of yours not too long ago when you and Brian did the functional Frontend Masters course. Thanks for teaching that course and helping me level up.

PR #779, opened 2016-12-24
Fix alias in select for mutate queries and small refactor

steve-chavez:
Currently when doing this request:

```
POST /projects?select=pId:id,name,client_id
Prefer: return=representation
{"id":7,"name":"New Project","client_id":2}
```

This response is obtained:

```json
{"hint":null,"details":null,"code":"42703","message":"column pg_source.id does not exist"}
```

The error comes from the ```returningF``` function that allows the RETURNING to use alias and the subsequent SELECT made to shape the response doesn't consider the renamed columns.

This PR fixes that and also I refactored because I think that the RETURNING section fits better in the ```MutateQuery``` type since it's a part of the sql syntax for each of the sql mutate statements.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -421,8 +421,8 @@ mutateRequest apiRequest readReq = mapLeft (errResponse status400) $
         _ -> undefined
     fieldNames :: ReadRequest -> PreferRepresentation -> [FieldName]
     fieldNames _ None = []
-    fieldNames (Node (Select colSelects _ _ _ _, (_, _, _)) forest) _ =
-      map (fst . view _1) colSelects ++ map colName fks
+    fieldNames (Node (sel, _) forest) _ =
+      map (fst . view _1) (select sel) ++ map colName fks
       where
         fks = concatMap (fromMaybe [] . f) forest
         f (Node (_, (_, Just Relation{relFColumns=cols, relType=Parent}, _)) _) = Just cols
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -7,6 +7,8 @@ module PostgREST.App (
 ) where
 
 import           Control.Applicative
+import           Control.Lens.Getter       (view)
+import           Control.Lens.Tuple        (_1)
 import qualified Data.ByteString.Char8     as BS
 import           Data.IORef                (IORef, readIORef)
 import           Data.List                 (delete, lookup)
@@ -58,7 +60,6 @@ import           PostgREST.QueryBuilder ( callProc
                                         , createReadStatement
                                         , createWriteStatement
                                         , ResultsWithCount
-                                        , returningF
                                         )
 import           PostgREST.Types
 import           PostgREST.OpenAPI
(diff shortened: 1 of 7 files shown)


begriffs:
Woo hoo, I always love refactoring in the QueryBuilder jungle. 🦁 

Would you mind if I finish and merge #763 first and then you can rebase on it? I think we both discovered some of the same things like useless arguments.

steve-chavez:
Sure, no problem.

src/PostgREST/App.hs
```diff
@@ -423,6 +423,15 @@ mutateRequest apiRequest = mapLeft (errResponse status400) $
...
         _ -> undefined
+    fieldNames :: ReadRequest -> PreferRepresentation -> [FieldName]
+    fieldNames _ None = []
+    fieldNames (Node (Select colSelects _ _ _ _, (_, _, _)) forest) _ =
+      map (fst . view _1) colSelects ++ map colName fks
+      where
+        fks = concatMap (fromMaybe [] . f) forest
+        f (Node (_, (_, Just Relation{relFColumns=cols, relType=Parent}, _)) _) = Just cols
+        f _ = Nothing
+    fieldNames' = fieldNames readReq (iPreferRepresentation apiRequest)
```
ruslantalpa: For clarity, maybe name this `returningFieldNames'` or something like that

ruslantalpa:
Good job!

diogob:
This makes a lot of sense. Nice refactor! 🍻

src/PostgREST/App.hs
steve-chavez: Thanks, I've renamed it to ```returnings```.

begriffs:
OK, @steve-chavez I merged the other one. Ready for your rebase.

steve-chavez:
@begriffs Done. Let me know if it's ok.

src/PostgREST/App.hs
```diff
@@ -419,6 +419,15 @@ mutateRequest apiRequest = mapLeft (errResponse status400) $
       case target of
         (TargetIdent (QualifiedIdentifier _ t) ) -> t
         _ -> undefined
+    fieldNames :: ReadRequest -> PreferRepresentation -> [FieldName]
+    fieldNames _ None = []
+    fieldNames (Node (Select colSelects _ _ _ _, (_, _, _)) forest) _ =
```
begriffs: There are a lot of underscores, what about matching `(Node sel forest)` and then using the `select` accessor from Select to obtain what you call `colSelects` here?

test/Feature/InsertSpec.hs
```diff
@@ -277,6 +289,17 @@ spec = do
                             "Location" <:> "/no_pk?a=is.null&b=eq.foo"]
           }
 
+      it "only returns the requested column header with his associated data" $
```
begriffs: :ant: small English nitpick: "its" rather than "his."

src/PostgREST/App.hs
steve-chavez: I've improved the function to match ```Node (sel, _) forest```. Thanks.

test/Feature/InsertSpec.hs
steve-chavez: Fixed.

PR #787, opened 2017-01-18
Move DbRequest building functions to new module

steve-chavez:
I noticed similar signatures and intention in these functions.


begriffs:
It's great to consolidate the code, thanks. Especially good to clean App out. The names QueryBuilder and DbRequestBuilder are confusingly similar but I can't think of anything better.

PR #794, opened 2017-02-03
Increase lower and upper bound for optparse-applicative

### LINKED ISSUE #778: Support optparse-applicative 0.13
Hi,

I noticed the constraint added in #752. Would it be possible to support optparse-applicative 0.13?

PierreR:
Fix #778

- Compilation OK both with stack and nix (tested with ghc-8 only)

- Adding `import Data.Monoid` should be useless whenever Protolude releases a new version (https://github.com/sdiehl/protolude/commit/0fb7730b0c540c1be28cc18601a39cea50677eb7)

- I didn't update to a more recent LTS-7 version (to minimize PR changes)

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -70,7 +70,7 @@ library
                      , lens
                      , lens-aeson
                      , network-uri
-                     , optparse-applicative >= 0.12.0.0 && < 0.13.0.0
+                     , optparse-applicative >= 0.13 && < 0.14
                      , parsec
                      , protolude
                      , Ranged-sets == 0.3.0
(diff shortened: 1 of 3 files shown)


begriffs:
Oh...didn't realize that's all it takes to fix! A while ago I tried bumping the version but got actual compilation problems with my code using optparse. But that code has been simplified by reading the config out of a file and looks like the new version works great now.

Thank you!

PierreR:
On that matter I am a bit disappointed that command line args was not kept to override the configuration

begriffs:
Your concern is ease of use in Docker or Heroku?

PierreR:
My immediate concern is more about compatibility. I will need to change the puppet scripts to install the new version.

Then I have been using tools that have both args on the command line and a configuration file and it has proven to be useful so I am a bit biased ;-)

Anyhow not a big deal really

PR #798, opened 2017-02-08
built with nix, worked on my box

NotBrianZach:
to build,
$ nix-build
symlink to executable will be ./result/bin/postgrest

to drop into shell with postgrest dependencies,
$  nix-shell
from nix-shell, should be able to build with,
$ cabal configure
$ cabal build
should also be able to import postgrest dependencies in ghci from nix-shell.

also swagger2.nix and hasql.nix generated e.g.
cabal2nix cabal://swagger2-2.1.3 > ./nixExternalLibs/swagger2.nix  
in the future this might be improved by moving this into default.nix with callHackage or another superior helper method see discussion here:
https://github.com/Gabriel439/haskell-nix/issues/6

also just noticed I included some crufty commented out code, sorry :P.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,14 +6,16 @@ dependencies:
     - curl -L https://github.com/commercialhaskell/stack/releases/download/v1.1.2/stack-1.1.2-linux-x86_64.tar.gz | tar zx -C /tmp
     - sudo mv /tmp/stack-1.1.2-linux-x86_64/stack /usr/bin
     - sudo apt-get update; sudo apt-get install --only-upgrade binutils
-    - curl https://nixos.org/nix/install | sh && . /home/ubuntu/.nix-profile/etc/profile.d/nix.sh
-
+    - curl https://nixos.org/nix/install | sh 
   override:
     - stack setup
     - rm -fr $(stack path --dist-dir) $(stack path --local-install-root)
     - stack install hlint packdeps cabal-install
     - stack build --fast
     - stack build --fast --test --no-run-tests
+  post:
+    . /home/ubuntu/.nix-profile/etc/profile.d/nix.sh
+
 compile:
   post:
     - nix-build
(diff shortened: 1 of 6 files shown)


begriffs:
Thanks @NotBrianZach. I notice these files declare dependencies which have a danger of falling out of date, do you know a way to verify the Nix build in Circle CI?

@wavewave can you review the Nix expressions? You know way more than I do about this than I do.

NotBrianZach:
@begriffs 
 verify the build in Circle CI:
You mean make sure the executable was generated, and run existing test suite against it? I could probably figure out how to do that.

begriffs:
Just that the executable was generated would be fine. I trust that if the tests work ordinarily then they'll work on that build as well.

begriffs:
Thinking about this again, what if you submitted your package to https://github.com/NixOS/nixpkgs instead? Then it would show up on https://nixos.org/nixos/packages.html

I'm worried that if I make the Nix build part of CI for this project then it could block development on core postgrest during any times when the Nix build is temporarily broken.

Your contribution is really cool and I'd love to link to it in the postgrest docs as well. Just thinking it might be better to keep heavy packaging things out of this repo.

NotBrianZach:
that's a fine plan by me, postgrest might build on nixpkgs https://github.com/NixOS/nixpkgs/pull/22367 as soon as they update to ghc 8.0.2 (possible still might need the hasql override, will have to look into it to tell if I need to do a pr there (which I might not, since this is working for me for now, other folks seem to be on it))

jwiegley:
@NotBrianZach Hi, just to add a note, this package can be built without using any custom scripts:

``` bash
cabal2nix . > default.nix
nix-build -E 'let pkgs = import <nixpkgs> {}; in pkgs.stdenv.lib.callPackageWith (pkgs // pkgs.haskellPackages) ./default.nix {}'
```

PR #799, opened 2017-02-09
Update hspec-wai to 0.8.0

diogob:
Stack resolver is now nightly-2017-01-31.

I actually started this upgrade to be able to compile on sierra (since GHC 8.0.1 was broken and the stackage nightly is using 8.0.2).

But to do so, hspec-wai had to be updated to 0.8.0 which changes the type for the `ResponseMatcher`.

Some caveats:

* When matching json now hspec assumes the header `Content-Type: application/json` when no `matchHeaders` is given, so I needed to specify it manually in some cases.
* When using `ResponseMatcher` I had to change the data constructor to use the `IsString` instance instead of the explicit `ResponseMatcher`.
* We cannot match against a body `Nothing` so now the body always has to be given as something that can be converted to a `ByteString`.
* In `SingularSpec.hs` line 145 the comment seems to be incorrect, since the delete fails and the rows are still there. I might be overlooking something though...

### DIFF
--- a/test/Feature/AuthSpec.hs
+++ b/test/Feature/AuthSpec.hs
@@ -18,43 +18,39 @@ spec = describe "authorization" $ do
   let single = ("Accept","application/vnd.pgrst.object+json")
 
   it "denies access to tables that anonymous does not own" $
-    get "/authors_only" `shouldRespondWith` ResponseMatcher {
-        matchBody = Just [json| {
+    get "/authors_only" `shouldRespondWith` [json| {
           "hint":null,
           "details":null,
           "code":"42501",
           "message":"permission denied for relation authors_only"} |]
-      , matchStatus = 401
+      { matchStatus = 401
       , matchHeaders = ["WWW-Authenticate" <:> "Bearer"]
       }
 
   it "denies access to tables that postgrest_test_author does not own" $
     let auth = authHeaderJWT "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoicG9zdGdyZXN0X3Rlc3RfYXV0aG9yIiwiaWQiOiJqZG9lIn0.y4vZuu1dDdwAl0-S00MCRWRYMlJ5YAMSir6Es6WtWx0" in
     request methodGet "/private_table" [auth] ""
-      `shouldRespondWith` ResponseMatcher {
-        matchBody = Just [json| {
+      `shouldRespondWith` [json| {
           "hint":null,
           "details":null,
           "code":"42501",
           "message":"permission denied for relation private_table"} |]
-      , matchStatus = 403
+      { matchStatus = 403
       , matchHeaders = []
       }
 
   it "returns jwt functions as jwt tokens" $
     request methodPost "/rpc/login" [single]
       [json| { "id": "jdoe", "pass": "1234" } |]
-      `shouldRespondWith` ResponseMatcher {
-          matchBody = Just [json| {"token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xuYW1lIjoicG9zdGdyZXN0X3Rlc3RfYXV0aG9yIiwiaWQiOiJqZG9lIn0.P2G9EVSVI22MWxXWFuhEYd9BZerLS1WDlqzdqplM15s"} |]
-        , matchStatus = 200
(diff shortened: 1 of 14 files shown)


begriffs:
That's a hefty change, I'll look it over this weekend.

begriffs:
Looks like LTS 8.0 is out now which uses GHC 8.0.2

test/Feature/AuthSpec.hs
```diff
@@ -94,9 +89,8 @@ spec = describe "authorization" $ do
   it "hides tables from users with invalid JWT" $ do
     let auth = authHeaderJWT "ey9zdGdyZXN0X3Rlc3RfYXV0aG9yIiwiaWQiOiJqZG9lIn0.y4vZuu1dDdwAl0-S00MCRWRYMlJ5YAMSir6Es6WtWx0"
     request methodGet "/authors_only" [auth] ""
-      `shouldRespondWith` ResponseMatcher {
-          matchBody = Nothing
-        , matchStatus = 401
+      `shouldRespondWith` [json| {"message":"JWT invalid"} |]
```
begriffs: It's kind of awkward that the new version doesn't make it easy to ignore the body and match the other parts. The `Test.Hspec.Wai.Matcher` module even has this internal helper:

```haskell
matchAny :: MatchBody
matchAny = MatchBody (\_ _ -> Nothing)
```

On the other hand maybe asserting the particular output in the body provides better test coverage anyway...

test/Feature/InsertSpec.hs
```diff
@@ -259,10 +252,11 @@ spec = do
             |12,0.1,a string,true,1929-10-01,12,bar
             |]
         request methodPost "/menagerie" [("Content-Type", "text/csv"), ("Accept", "text/csv"), ("Prefer", "return=representation")] inserted
-
-           `shouldRespondWith` ResponseMatcher {
-             matchBody    = Just inserted
-           , matchStatus  = 201
+           `shouldRespondWith` [str|integer,double,varchar,boolean,date,money,enum
+            |13,3.14159,testing!,false,1900-01-01,$3.99,foo
```
begriffs: Is it necessary to repeat the value of `inserted` here?

test/Feature/QuerySpec.hs
```diff
@@ -30,68 +30,49 @@ spec = do
...
-        , matchStatus  = 200
-        , matchHeaders = ["Content-Range" <:> "0-2/*"]
-        }
+        `shouldRespondWith` [json| [{"id":1},{"id":3},{"id":5}] |]
+        { matchHeaders = ["Content-Range" <:> "0-2/*"] }
 
     it "matches nulls using not operator" $
       get "/no_pk?a=not.is.null" `shouldRespondWith`
         [json| [{"a":"1","b":"0"},{"a":"2","b":"0"}] |]
+        { matchHeaders = ["Content-Type" <:> "application/json; charset=utf-8"] }
```
begriffs: This occurs so often maybe you should have a *let* alias for `["Content-Type" <:> "application/json; charset=utf-8"]`.

test/Feature/SingularSpec.hs
```diff
@@ -144,9 +144,8 @@ spec =
 
         -- the rows should not exist, either
```
begriffs: I think this comment was incorrectly copied and pasted from earlier. As you point out, it's not accurate here. The rows were not deleted so they *should* still exist.

test/Feature/InsertSpec.hs
diogob: I'm not sure, will investigate. I just duplicated it because I was having problems type checking, but this should be revisited.

test/Feature/QuerySpec.hs
diogob: True, I'll refactor this.

test/Feature/RangeSpec.hs
```diff
@@ -121,57 +114,51 @@ spec = do
...
-            }
+            `shouldRespondWith` [json| [{"id":1}] |]
+            { matchHeaders = ["Content-Range" <:> "0-0/*"] }
 
     context "with limit/offset parameters" $ do
       it "no parameters return everything" $
         get "/items?select=id&order=id.asc"
           `shouldRespondWith` ResponseMatcher {
-            matchBody    = Just [str|[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}]|]
+            matchBody    = [str|[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}]|]
```
diogob: sure 😄

diogob:
I'll try to re-target the PR against LTS 8.0 and apply the suggested changes during this week

begriffs:
(Sorry, merged something under your PR, you'll need to do a quick rebase, shouldn't be bad)

diogob:
Done! Should be good to go.

diogob:
Sorry, just noticed onde last fix on the cabal.

Issue #506, opened 2016-02-27
Support raw binary output (and input)


nicklasaven:
Would it be possible to find a good way to output raw binary data from PostgREST?

It would mean returning data like application/octet-stream
https://www.iana.org/assignments/media-types/application/octet-stream

The main problem is that it cannot be put in a json object which is text.

In PostgreSQL this type of data is stored as bytea.

eric-brechemier:
@nicklasaven What is your use case?

nicklasaven:
Hi, sorry for late reply

My use case might be rare, so maybe I should solve it in some other way.

But dealing with json is in many cases heavy and not the way to go.

I use PostGIS and are testing solutions with the new twkb export format.

That is a format that compresses the geoemtries by delta-encoding the coordinates and stores them as VarInts, the same way as protobuffer do.

So, what I want to do is getting thetwkb geometries out through the rest api.

But maybe with nginx in front I should get them directly from the db instead. But I like the level of security PostgREST adds by just accepting right formatted requests. It is an easy way to avoid any sql-injection problems.

begriffs:
Can you explain a little more? You're saying, for instance, that one of your tables has a column with big binary values, and you would like to return one of these values directly without it being embedded in JSON?

geoervin:
My use case is pretty simple but I have not been to find a solution. I have a content type of application/pdf (blob) that I would like to write (post) to the database using the postgres bytea binary type.   My attempts at using with text datatype with base64 encoding got me close but eventually became too complex to justify for my skill set.

nicklasaven:
My use case might be better to solve outside postgrest as discussed above, but it might be nice to ba able to do it inside postgrest too.

I saw in #257 that @begriffs suggests a one row return limitation. For my use case it would be better if the result was just concatenated. The format includes information about each parts length and have a defined ending for each geoemtry. So getting a long binary string scrubbed from the bytea meta-data would be the best. 

Then the job is up to the client to rip it apart and render it on a map.

In my case the binary format also includes an ID so the geometry can be linked to attribute data client side.  

But I am not sure this type of usage has a generic usage, so probably better to do in other ways.

An alternative could be a possibility to add a delimiter to the call, like the postgresql function string_agg(string, delimiter). In this case it should be an ascii sequence (or hex string or whatever)  so there is no encoding problems when used as binary.

diogob:
@nicklasaven would encoding the bytea field as base64 and keeping the json be a possible solution?
[...]

Issue #257, opened 2015-08-10
Directly serve individual field data somehow?


motiz88:
My current project generates some binary data on the client, then stores it as part of a database. This got me wondering, could PostgREST just serve this data back over HTTP upon request? Not encoded as part of a result set - just the one plain value.

So this is me thinking out loud about this feature, just trying to get an interesting discussion going.
1. There would have to be a way to unambiguously & idiomatically request a particular value out of a row, as opposed to a representation of a row / set. I feel that's probably doable using a URL and optionally some headers.
   For example, `GET /table/col1?col1=eq.123` could be defined to mean "the value of `table.col1` in the result of `GET /table?col1=eq.123`, provided that _that_ query returns precisely one row".
   To play it extra safe, PostgREST can try to guarantee that ahead of time (looking at keys, uniqueness constraints and any `Range` header) and error out appropriately.
   (Bike shed issue: `GET /table/columns/col1` vs. `GET /table/col1` vs ...)
2. `Content-Type` is a problem - how would PostgREST know it in the general case? I don't know the ins and outs of Postgres' type system nearly well enough, but perhaps we could use a form of reflection and define a map from Postgres types to MIME types, which we would let the user extend for their own types as part of the schema. Along with basic `Accept` handling and some sensible defaults - `text/plain` for text values, `application/octet-stream` for `bytea`, `application/json` for JSON etc - this might even work well.

Yes, the line between database and general-purpose web server has to be drawn somewhere, but insofar as individual values within a row can be thought of as resources, and particular ways of serving them thought of as representations, I think this idea has some merit in a REST context. Hopefully some of you will agree :)

begriffs:
Holy crap @motiz88 I apologize for not responding to this question for **months**.

Returning a single "cell" from the db seems similar to what @nicklasaven just asked about in the context of PostGIS binary data in #506.

I think this feature is feasible and we could enable it if the client does three things in their request

1. Includes header `Prefer: plurality=singular` to single out one row
2. Selects a single column `?select=col` 
3. Includes header `Accept: application/octet-stream` to indicate they do not want the JSON wrapping

begriffs:
This functionality is now in master.

PR #802, opened 2017-02-12
Proposal for binary output 

steve-chavez:
This feature is similar to what @begriffs proposed in #257, to request an unwrapped value the client must:

- Select a single column ```?select=col```
- Include a header ```Accept: application/octet-stream```, if this is specified wihout selecting a single column a ```406 not acceptable``` is returned.

Constrasting with @begriffs proposal, more than one row can be returned and if this happens the results are concatenated with no delimiter ```''```, I don't think this is a problem since most binary formats have headers so this also addresses the use case for the TWKB format in #506.


begriffs:
Wow this feature is going to make a lot of people very happy, thank you! Add a line in the changelog and I'll take a pass through the code and review it. The multiple row concatenation behavior seems fine, like you said it has its uses.

src/PostgREST/App.hs
```diff
@@ -93,22 +101,26 @@ app dbStructure conf apiRequest =
...
-                if contentType == CTSingularJSON && queryTotal /= 1
-                  then singularityError (toInteger queryTotal)
-                  else responseLBS status
-                    [toHeader contentType, contentRange,
-                      ("Content-Location",
-                        "/" <> toS (qiName qi) <>
-                          if BS.null canonical then "" else "?" <> toS canonical
-                      )
-                    ] (toS body)
+            Right (q, cq) ->
```
begriffs: One idea to reduce the nesting of these cases:

```haskell
(ActionRead, TargetIdent qi, Nothing) ->
  let partsFields = (,) <$> readSqlParts
        <*> (binaryField contentType =<< fldNames) in
  case partsFields of
    Left errorResponse -> return errorResponse
    Right ((q, cq), bField) -> do
```

Sadly it is more dense and symbolic and introduces another variable name but it does prevent nested case statements dealing with Either.

src/PostgREST/App.hs
```diff
@@ -285,6 +305,15 @@ responseContentTypeOrError accepts action = serves contentTypesForRequest accept
             "None of these Content-Types are available: " <> failed
         Just ct -> Right ct
 
+binaryField :: ContentType -> [FieldName] -> Either Response (Maybe FieldName)
+binaryField CTOctetStream fieldNames = 
+  if length fieldNames == 1 && fromJust fieldName /= "*"
```
begriffs: What about `fieldName /= Just "*"`? I know in this case they are logically equivalent but I like to avoid partial functions like `fromJust`.

src/PostgREST/App.hs
```diff
@@ -257,8 +269,16 @@ app dbStructure conf apiRequest =
 
       mapSnd f (a, b) = (a, f b)
       readReq = readRequest (configMaxRows conf) (dbRelations dbStructure) (map (mapSnd pdReturnType) $ dbProcs dbStructure) apiRequest
+      fldNames = fieldNames <$> readReq
+      fieldNames :: ReadRequest -> [FieldName]
```
begriffs: This feels out of place here somehow. The surrounding statements are making helper variables and this is a full function. Maybe put it lower in the file or in another module?

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -247,8 +246,8 @@ toSourceRelation mt r@(Relation t _ ft _ _ rt _ _)
   | Just mt == (tableName <$> rt) = Just $ r {relLTable=(\tbl -> tbl {tableName=sourceCTEName}) <$> rt}
   | otherwise = Nothing
 
-mutateRequest :: ApiRequest -> ReadRequest -> Either Response MutateRequest
-mutateRequest apiRequest readReq = mapLeft (errResponse status400) $
+mutateRequest :: ApiRequest -> [FieldName] -> Either Response MutateRequest
```
begriffs: I like how this type exposes more about what mutateRequest uses. 👍

src/PostgREST/App.hs
steve-chavez: I'll move ```fieldNames``` to DbRequestBuilder.hs, I think is related to the module since is used to build ```mutateRequest```.

src/PostgREST/App.hs
steve-chavez: I'll take that idea as it is since I don't have enough experience in Haskell(just started learning it to contribute) to propose another one.

begriffs:
Could you send a PR to add a section to the docs in https://github.com/begriffs/postgrest-docs/blob/master/api.rst#response-format as well?

src/PostgREST/App.hs
begriffs: You're just beginning with Haskell? Wow that's impressive, you jumped right into this codebase.

begriffs:
Anyone else have any comments before I merge this?

steve-chavez:
Sure, I'll document this feature.

fab1an:
Does this work for stored procedures as well?

begriffs:
No, just get requests on tables and views.

fab1an:
I see… I have a stored procedure returning `bytea` (a PDF). Is there some way I can render that as bytestream using `application/pdf`, so it get's display in the browser properly?

begriffs:
I imagine the technique in this pull request can be extended in another PR to enable outputting `application/octet-stream` for stored procedures as well, although I don't know how postgrest would know to label the result as PDF in particular.

fab1an:
I think one could use nginx to do that. 

Another idea: Does it make sense to provide the Accept Header to the procedure and set the response ContentType to the same value?

steve-chavez:
I'll add the binary output ability to stored procedures in another PR, for now to display the pdf in the browser with GET maybe you can use something like:

```
	location /api/documents/ {
		proxy_pass http://127.0.0.1:3000/documents;
		proxy_set_header Accept application/octet-stream;
		proxy_hide_header Content-Type;
		add_header  Content-Type application/pdf;
		add_header  Content-Disposition inline;
	}
```
And then visit the url ```http://localhost/api/documents?select=content&name=eq.some_sample.pdf```.

fab1an:
@steve-chavez Thanks.

steve-chavez:
@fab1an Could you open an issue with your use case regarding binary output for stored procedures? I'd like to start working on this.

fab1an:
@steve-chavez I gave up using postgrest for generating these PDFs for now, so from my side there's no use-case anymore :/.

begriffs:
@steve-chavez thanks for your help with this. I notice you forked the docs repo today too so assuming you're documenting the octet-stream stuff. 👍

nicklasaven:
This is great!
THanks!!

PR #803, opened 2017-02-13
Public view failure test

begriffs:
A version of #761 with a test.


begriffs:
Had to deal with some tricky behavior in the test fixture with public schema ownership but got it to work. Also verified that the new test fails on master.

Issue #808, opened 2017-02-17
Standardize and simplify docker workflow


begriffs:
It's high time we get Docker figured out. The existing Dockerfile was created in the 0.3 era of postgrest and as written won't work with 0.4. There are more than one way to fix it. In particular there are two PRs currently open (#738, #762), and I appreciate the work that went into them. I'd like to propose a simpler way to use Docker though.

**Proposal**: provide images on Docker Hub for tagged version releases only, remove nightlies, have a "latest" alias for the newest tagged version.

Here's why:

* Docker support is crucial for people deploying postgrest to Kubernetes, but are these people deploying nightlies?
* I would prefer that support requests from Docker users be on a known version. It's easier than debugging a problem in a random build.
* We can avoid complicated build machinery and its maintenance.
* Stack builds postgrest from source reliably on all platforms that I'm aware of, so contributors can already play with the newest build without needing Docker.

Also, for the design of the Dockerfile:

* Continue having the Dockerfile download a binary from the Github releases. I build these binaries anyway, so no extra work there.
* Require mapping in a configuration file rather than accepting env vars. The dockerfiles of nginx and apache do it this way, and postgrest is a web server too so is it that weird?

ruslantalpa:
All except the last point about  env vars.
This is a situation where you can totally have your cake and eat it too :)
start with an image that has the binary and a default conf file and the CMD is postgrest -c /....

then add an entrypoint script that checks if an env var like `POSTGREST_ANON` exists then it does something like 

`echo "anon-user=$POSTGREST_ANON" > /etc/postgrest.conf`

with this way the user will have the option to either bake the config file in, provide it with a mapping or override a specific config parameter using a env var

deinspanjer:
> Proposal: provide images on Docker Hub for tagged version releases only, remove nightlies, have a "latest" alias for the newest tagged version.
 
> Docker support is crucial for people deploying postgrest to Kubernetes, but are these people deploying nightlies?

Only if they are developers contributing to the postgrest project in the context of their own applications.  I think it makes sense to try to make things easy for developers both in and out of the Docker environment.

> I would prefer that support requests from Docker users be on a known version. It's easier than debugging a problem in a random build.

Agreed.

> We can avoid complicated build machinery and its maintenance.

Agreed.

> Stack builds postgrest from source reliably on all platforms that I'm aware of, so contributors can already play with the newest build without needing Docker.

[...]

PR #810, opened 2017-02-18
Dockerfile that works with postgrest 0.4

begriffs:
Updated the Dockerfile command to point postgrest at a config file. Verified that docker compose could spin it up locally when using a local image I built from the Dockerfile. After merging this I can generate a new "latest" on dockerhub.

This supports the direction I talked about in #808

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,6 +13,6 @@ RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VER
 
 # PostgREST reads /etc/postgrest.conf so map the configuration
 # file in when you run this container
-CMD exec postgrest
+CMD exec postgrest /etc/postgrest.conf
 
 EXPOSE 3000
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -13,6 +13,6 @@ RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VER
 
 # PostgREST reads /etc/postgrest.conf so map the configuration
 # file in when you run this container
-CMD exec postgrest
+CMD exec postgrest /etc/postgrest.conf
 
 EXPOSE 3000
```
begriffs: It's possible for the user to mess this up by choosing a different port in the config file they map in.

docker/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+stgrest:
+  image: begriffs/postgrest:latest
```
begriffs: At this stage this refers to 0.3.2.1 on docker hub. To test this docker-compose you'll need to build an image locally and refer to that until we tag this repo with `v0.4.0.0-1` and trigger a docker hub build.

Issue #822, opened 2017-03-01
JWT secret shall not be allowed to be empty


eric-brechemier:
I recently encountered this issue on PostgREST v0.3.0.3: while there is a check in place to ensure that the parameter `--jwt-secret` is present, there was no complaint when the provided value was `''` (the empty value being the result of an error, not a conscious choice).

The goal of this ticket is twofold:

1. warn users of older versions of PostgREST
2. kindly request a check (if not in place already) to ensure it does not happen with the latest version

begriffs:
Thanks for the note. Can you check whether 0.4 is susceptible to the problem as well?

As an unrelated observation, 0.3.0.3 is really really old, have you considered using a newer 0.3.x version? Look at all the bugfixes for later releases: https://github.com/begriffs/postgrest/releases

eric-brechemier:
No actually, we only considered migrating to 0.4, but I see now that upgrading to latest 0.3.x would be a good stepping stone. Do you keep a list of issues fixed in each new release? I have looked for additional details for several issues listed in the change log, but failed to identify the issue which matched the description, e.g. ("Prevent role from being changed twice" and "Prevent query error from infecting later connection").

eric-brechemier:
> Can you check whether 0.4 is susceptible to the problem as well?

Yes, I confirm that the issue is present on `master`.

I left the `jwt-secret` empty:

```
## choose a secret to enable JWT auth   
## (use "@filename" to load from separate file)   
jwt-secret = ""        
# secret-is-base64 = true
```

and started the server without getting any complaint.

I generated a JWT using the empty secret `''` and checked that it was accepted by this instance of the server.

begriffs:
> Prevent role from being changed twice

This one was a duplicate `SET ROLE ...` statement sent during authentication. It was harmless but just a little inefficient.

> Prevent query error from infecting later connection

This one happened after calling a stored proc that raised an exception. The next http request could sometimes end up using the previous postgresql connection and getting an error result.

___________

In the future I'll include issue numbers to the changelog so it's easier to get more details.

eric-brechemier:
Thanks.

steve-chavez:
Looks like this was already fixed in https://github.com/PostgREST/postgrest/commit/d6a848c8859f3ec7c37d5efab6bc02d5d60ec5cb but github somehow didn't close the issue.

Issue #819, opened 2017-02-28
docker-compose dosn't load postgrest.conf file


xamika:
I was trying to use posgrest in a docker container. I used The instruction on this page: https://github.com/begriffs/postgrest/tree/master/docker but it isn't working. I mounted the config.conf file to /etc/postgrest.conf. This is working but postgrest doesn't load this config file from etc/postgrest.conf.

ruslantalpa:
make sure you are using the latest version (4.0)
looking at the tags in docker i think you need to use "nightly" tag

xamika:
Thank you for you response. I updated my docker-compose file:
```
  postgrest:
    image: begriffs/postgrest:nightly
    ports:
     - '3000:3000'
    links:
     - postgres
    volumes:
     - ./conf/postgrest.conf:/etc/postgrest.conf
```
but now I receive another error XD:
`postgrest_1  | Missing: FILENAME
postgrest_1  | 
postgrest_1  | Usage: postgrest FILENAME
postgrest_1  |   PostgREST 0.4.0.0 / create a REST API to an existing Postgres database
postgrest_1  | 
postgrest_1  | Available options:
postgrest_1  |   -h,--help                Show this help text
postgrest_1  |   FILENAME                 Path to configuration file
postgrest_1  | 
`

ruslantalpa:
can you try and build the image on your computer from this definition?
https://github.com/begriffs/postgrest/blob/master/docker/Dockerfile

i am not sure what exactly how that nightly image was built
(the docker for the latest version is not quite right yet, there is still a discussion of how to get it up to date)

xamika:
I solved my problem with a workarround (I'm overrwriting the postgrest command):

```
postgrest:
    image: begriffs/postgrest:nightly
    ports:
     - '3000:3000'
    links:
     - postgres
    volumes:
     - ./conf/postgrest.conf:/etc/postgrest.conf
    command: postgrest /etc/postgrest.conf
```
Please add the path to your command. https://hub.docker.com/r/begriffs/postgrest/builds/bwxqmc8rj3ygeycorkarnq8/

begriffs:
Good workaround. The `latest` image on docker hub is not working right for 0.4 because we're still deciding the best way to do things for the new version.

begriffs:
[...]

PR #826, opened 2017-03-06
Use env interpolation in config

begriffs:
Work in progress to use a combination of config file and environment variables in a Dockerfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,6 +11,21 @@ RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VER
     mv postgrest /usr/local/bin/postgrest && \
     rm postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
 
+COPY postgrest.conf /etc/postgrest.conf
+
+
+ENV PGRST_DB_URI= \
+    PGRST_DB_SCHEMA=public \
+    PGRST_DB_ANON_ROLE= \
+    PGRST_DB_POOL=100 \
+    PGRST_SERVER_HOST=*4 \
+    PGRST_SERVER_PORT=3000 \
+    PGRST_SERVER_PROXY_URL= \
+    PGRST_JWT_SECRET= \
+    PGRST_SECRET_IS_BASE64=false \
+    PGRST_MAX_ROWS= \
+    PGRST_PRE_REQUEST=
+
 # PostgREST reads /etc/postgrest.conf so map the configuration
 # file in when you run this container
 CMD exec postgrest /etc/postgrest.conf
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
 {-|
 Module      : PostgREST.Config
 Description : Manages PostgREST configuration options.
@@ -28,8 +29,10 @@ import qualified Data.ByteString.Char8       as BS
 import qualified Data.CaseInsensitive        as CI
 import qualified Data.Configurator           as C
 import qualified Data.Configurator.Parser    as C
+import           Data.Configurator.Types     (Value(..))
 import           Data.List                   (lookup)
 import           Data.Monoid
+import           Data.Scientific             (floatingOrInteger)
 import           Data.Text                   (strip, intercalate, lines)
 import           Data.Text.Encoding          (encodeUtf8)
 import           Data.Text.IO                (hPutStrLn)
@@ -105,11 +108,11 @@ readOptions = do
           <*> C.key "server-proxy-uri"
           <*> C.key "db-schema"
           <*> (fromMaybe "*4" <$> C.key "server-host")
-          <*> (fromMaybe 3000 <$> C.key "server-port")
+          <*> (fromMaybe 3000 . join . fmap coerceInt <$> C.key "server-port")
           <*> (fmap encodeUtf8 <$> C.key "jwt-secret")
           <*> (fromMaybe False <$> C.key "secret-is-base64")
-          <*> (fromMaybe 10 <$> C.key "db-pool")
-          <*> C.key "max-rows"
+          <*> (fromMaybe 10 . join . fmap coerceInt <$> C.key "db-pool")
+          <*> (join . fmap coerceInt <$> C.key "max-rows")
           <*> C.key "pre-request"
           <*> pure False
 
@@ -121,6 +124,11 @@ readOptions = do
       return appConf
 
  where
+  coerceInt :: (Read i, Integral i) => Value -> Maybe i
+  coerceInt (Number x) = rightToMaybe $ floatingOrInteger x
+  coerceInt (String x) = readMaybe $ toS x
+  coerceInt _ = Nothing
+
   opts = info (helper <*> pathParser) $
            fullDesc
            <> progDesc (
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -110,7 +110,7 @@ readOptions = do
           <*> C.key "db-schema"
           <*> (fromMaybe "*4" <$> C.key "server-host")
           <*> (fromMaybe 3000 . join . fmap coerceInt <$> C.key "server-port")
-          <*> (fmap encodeUtf8 <$> C.key "jwt-secret")
+          <*> (fmap encodeUtf8 . mfilter (/= "") <$> C.key "jwt-secret")
           <*> (fromMaybe False <$> C.key "secret-is-base64")
           <*> (fromMaybe 10 . join . fmap coerceInt <$> C.key "db-pool")
           <*> (join . fmap coerceInt <$> C.key "max-rows")
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -41,6 +41,7 @@ import           Network.Wai
 import           Network.Wai.Middleware.Cors (CorsResourcePolicy (..))
 import           Options.Applicative hiding  (str)
 import           Paths_postgrest             (version)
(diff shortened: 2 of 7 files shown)


begriffs:
Relates to the conversation in #808

ruslantalpa:
What a difference a good lib makes :)

begriffs:
There's still a problem. The `docker/postgrest.conf` uses environment variables for all settings, but our config library requires the env variables be present for interpolation or else it dies. So what's the problem, we set defaults in the Dockerfile with `ENV`, right? Well yes, but there is one big problem and one annoyance:

* The annoyance is that we are specifying default values twice, like host and port. Once in the Dockerfile and once inside postgrest. A user can always specify the desired value so it's not really a problem, just inelegant.
* The bigger problem. In most cases there is not a sensible default value for environment variables, so I assign them the empty string. However postgrest does not treat these as missing values, it treats them as the user's choice. The `db-uri` parameter for instance treats the empty string as an indication to use the default unix socket which will never work in our Docker container.

I propose we add logic to our configuration module that treats the empty string as a missing parameter. Interestingly this same problem of empty strings just arose in #822.

begriffs:
Hang on, looks like there is a successor to the configurator library: https://hackage.haskell.org/package/configurator-ng

Let's see what this one can do for us...

deinspanjer:
I have a PR with new docker build stuff ready to test, but need an env aware config file.  How is this testing coming?

begriffs:
I'll try to finish that up this weekend. Thanks for your patience with all this.

begriffs:
Note to self: once this branch is fixed and merged, check whether it helps #819.

begriffs:
@deinspanjer I ran into a problem and opened issue in the configurator-ng project about it. Hopefully they will have a suggestion.

docker/Dockerfile
```diff
@@ -11,6 +11,21 @@ RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VER
     mv postgrest /usr/local/bin/postgrest && \
     rm postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz
 
+COPY postgrest.conf /etc/postgrest.conf
+
+
+ENV PGRST_DB_URI= \
```
FGRibreau: \o/ :heart:

begriffs: Yeah this will work pretty well as a mix of env variables and mapping files into Docker if we can just get past the stumbling block of parsing quoted integers - https://github.com/lpsmith/configurator-ng/issues/4

ronalddddd:
Hope this is relevant. My current workaround to have an env vars-only docker image: 

```
# https://github.com/ronalddddd/docker-postgrest/blob/master/entrypoint.sh
# ...
cat > /etc/postgrest.conf <<-EOF
db-uri = "postgres://${DBUSER}:${DBPASS}@${DBHOST}:${DBPORT}/${DBNAME}"
db-schema = "${DBSCHEMA}"
db-anon-role = "${ANONUSER}"
EOF

postgrest /etc/postgrest.conf
```

Docker image can be found here: https://hub.docker.com/r/ronalddddd/postgrest/

begriffs:
@ronalddddd, interesting, does it allow for all the options listed in the [configuration](https://postgrest.com/en/v0.4/admin.html) section?

begriffs:
@roman I just read about Haskell-etc on the Haskell Cafe mailing list. It may be just what we're looking for in this pull request! Maybe you can advise.

The background is we read a simple [config file](https://postgrest.com/en/v0.4/admin.html) on server startup. We would like to interpolate environment variables in the config file to allow it to be used in Docker without the need to map a config file into the container. I was trying the [configurator-ng](https://github.com/lpsmith/configurator-ng) library, but it's unable to parse parse environment vars as integers, only as strings. Doesn't work with settings such as `server-port` which is an integer in our Haskell code.

roman:
Hi @begriffs,

I guess the first thing is to define a spec file for your `postgrest` executable to use, in this spec file you would say which env vars align to your config keys. My first attempt would be something like this:

```yaml
etc/entries:
  db-uri:
    etc/spec:
      env: PGRST_DB_URI

  db-schema:
    etc/spec:
      env: PGRST_DB_SCHEMA

  db-anon-role:
    etc/spec:
      env: PGRST_DB_ANON_ROLE

  db-pool:
    etc/spec:
      env: PGRST_DB_POOL


  server-host:
    etc/spec:
      env: PGRST_SERVER_HOST

  server-port:
    etc/spec:
      env: PGRST_SERVER_PORT


  server-proxy-url:
    etc/spec:
      env: PGRST_SERVER_PROXY_URL

  jwt-secret:
    etc/spec:
      env: PGRST_JWT_SECRET

  secret-is-base64:
    etc/spec:
      env: PGRST_SECRET_IS_BASE64

  max-rows:
    etc/spec:
      env: PGRST_MAX_ROWS

  pre-request:
    etc/spec:
      env: PGRST_PRE_REQUEST
```

And the Haskell side of things would look something like:

```haskell
import Data.Aeson ((.:), (.:?))
import qualified Data.Aeson as JSON
import qualified System.Etc as Etc

parseConfiguration :: Etc.Config -> Maybe AppConfig
parseConfiguration config =
  let
    getValue key =
      Etc.getConfigValue [key] config
  in
    AppConfig
      <$> getValue "db-uri"
      <*> getValue "db-annon-value"
      <*> getValue "server-proxy-uri"
      <*> getValue "db-schema"
      <*> (fromMaybe "*4" <$> getValue "server-host")
      <*> (fromMaybe 3000 <$> getValue "server-port")
      <*> (fmap encodeUtf8 <$> getValue "jwt-secret")
      <*> (fromMaybe False <$> getValue "secret-is-base64")
      <*> (fromMaybe 10 <$> getValue "db-pool")
      <*> getValue "max-rows"
      <*> getValue "pre-request"
      <*> pure False

main :: IO ()
main = do
  spec <- Etc.readConfigSpec "spec.yaml"
  config <- Etc.resolveEnv spec
  parseConfiguration config
```
_NOTE 1_ you may also try using an `Aeson.FromJSON` instance for your `AppConfig` type, you'll need the AppConfig to be a sub-map in your config though.

_NOTE 2_ If dependency count is something you are mindful of, maybe having the configuration as JSON might be a better idea (sadly the `yaml` package adds _many_ dependencies).

Let me know if you have more questions or if you want to pair to flesh things out, this API is still in testing and things might need some iterations to be fleshed out into a nicer interface.

Cheers, and thanks for considering `etc`.

jesseadams:
This would be extremely useful.

begriffs:
@roman sorry I didn't acknowledge your careful explanation, thank you for researching that. I'm going to keep fighting with the existing library though because I want to continue supporting the simple `key = val` style config file for backward compatibility.

ruslantalpa:
Instead of fighting the lib to get env vars to work how about this.
The use of "env vars" in config is a "docker thing".
This can be easily solved with an entry point script even with the current version.
IMO, the only thing that needs change is the Dockerfile to use a entrypoint script like this.

```bash
#!/bin/bash
set -e


if [[ $SERVER_PROXY_URI != "" ]]; then SERVER_PROXY_URI="server-proxy-uri = \"${SERVER_PROXY_URI}\""; fi
if [[ $JWT_SECRET != "" ]]; then JWT_SECRET="jwt-secret = \"${JWT_SECRET}\""; fi
if [[ $MAX_ROWS != "" ]]; then MAX_ROWS="max-rows = ${MAX_ROWS}"; fi
if [[ $PRE_REQUEST != "" ]]; then PRE_REQUEST="pre-request = \"${PRE_REQUEST}\""; fi

CONF=/etc/postgrest.conf
touch $CONF
cat >> $CONF <<-EOS
db-uri = "$DB_URI"
db-schema = "$DB_SCHEMA"
db-anon-role = "$DB_ANON_ROLE"
db-pool = $DB_POOL

server-host = "$SERVER_HOST"
server-port = 3000

## base url for swagger output
# server-proxy-uri = ""
$SERVER_PROXY_URI

## choose a secret to enable JWT auth
## (use "@filename" to load from separate file)
# jwt-secret = "foo"
$JWT_SECRET
secret-is-base64 = $SECRET_IS_BASE64

## limit rows in response
# max-rows = 1000
$MAX_ROWS

## stored proc to exec immediately after auth
# pre-request = "stored_proc_name"
$PRE_REQUEST

EOS

exec postgrest /etc/postgrest.conf

```

begriffs:
That's an option for sure. I'm currently hacking on this PR a little this afternoon and let's just see if I can make a breakthrough.

begriffs:
YEAH! It seems to be working.

src/PostgREST/Config.hs
```diff
@@ -95,30 +98,37 @@ readOptions = do
...
+  case mAppConf of
+    Nothing -> do
+      forM_ errs $ hPutStrLn stderr . show
+      exitFailure
+    Just appConf ->
+      return appConf
 
  where
+  coerceInt :: (Read i, Integral i) => Value -> Maybe i
+  coerceInt (Number x) = rightToMaybe $ floatingOrInteger x
```
begriffs: This function works for Int or Integer, but Haskell throws a warning on this line because it defaults `floatingOrInteger`'s left value to Double. I tried to cast it as `Either Double i` but GHC didn't catch on that I meant the same `i` as in the signature for this function. Experimented with `forall i.` but didn't do the trick. Don't know how to clean that up.

src/PostgREST/Config.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```
begriffs: Needed this for the `floatingOrInteger` thing later on. Would rather not suppress a warning though.

begriffs:
Going to take the silence as a :shipit:

Issue #757, opened 2016-12-05
Invalid JSON for resource embedding error


begriffs:
I accidentally tried to embed an incompatible table in another and the JSON of the resulting error message is invalid. Not only is there a message inside a message, but it lacks proper escaping even for that.

```
{"message":"{"details":"no relation between foo and bar","message":"could not find foreign keys between these entities"}"}
```

diogob:
this is probably a regression related to the latest refactoring around the error message. I'll take a look and add some regression testing.

begriffs:
Closing in favor of the more general fix described in #791

PR #829, opened 2017-03-12
Refactor error type and fix nested error message [fix #791]

### LINKED ISSUE #791: Malformed nested JSON error
@diogob this one seems like your specialty, a bug whose solution involves a nice refactor.

`/foo?id=0` throws a weird parse error with json inside json:

```
{"message":"{\"details\":\"unexpected \\\"0\\\" expecting \\\"not.\\\" or operator (eq, gt, ...)\",\"message\":\"\\\"failed to parse filter (0)\\\" (line 1, column 1)\"}"}
```

One way to attack at the root of the problem is changing the type of `errResponse` in App.hs to take a structured error type rather than text. Something like

```haskell
GeneralError {
  message :: Text,
  details :: Maybe Text,
  hint :: Maybe Text,
  code :: Maybe Text
}
```

Then follow the type errors throughout app/querybuilder etc and then all the malformed error anomalies should be fixed.

diogob:
This PR make fixes the problem of double encoding JSON errors and avoids whenever possible using Text values to represent error information. 

This is related to #791 and #757

Besides the fix there are some extra goals:

* Avoid exposing the `ParseError` type to better encapsulate the `Parsec` module inside our `Parsers`.
* Avoid exposing any formatting error functions from `Error` module so the error representation becomes more opaque.
* Add more constructors to `ApiRequestError` so it can describe in details more error conditions.
* Move the `ApiRequestError` type and some helper functions to the more generic `Types` module since they are used in several modules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- #791, malformed nested JSON error - @diogob
 - Resource embedding in views referencing tables in public schema - @fab1an
 
 ## [0.4.0.0] - 2017-01-19
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -9,7 +9,7 @@ import           PostgREST.Config                     (AppConfig (..),
                                                        minimumPgVersion,
                                                        prettyVersion,
                                                        readOptions)
-import           PostgREST.Error                      (prettyUsageError)
+import           PostgREST.Error                      (encodeError)
 import           PostgREST.OpenAPI                    (isMalformedProxyUri)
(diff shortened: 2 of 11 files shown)


begriffs:
Nice and clean! 🌈 

There's a tiny linter error in CI. If you could fix that and add a changelog entry then I'll get this merged.

For the changelog, I'd like to start referencing issue numbers, like

```markdown
### Fixed

- #791, malformed nested JSON error - @diogob
```

The reason is that somebody asked me on Gitter what the context is behind some of the changelog entries. Referencing an issue number will help people learn about the changes.

diogob:
:D Done!
I also removed the catch-all patterns from the `apiRequestError` and wrote an appropriate error status for each case.

PR #831, opened 2017-03-13
Fix proc resource embedding issue with search_path and refactor return type

steve-chavez:
I created a new db, role with login, tables ```clients```, ```projects``` and proc ```getproject(id int) returns setof projects```, all defined just as in ```fixtures/schema.sql```.

I tested the proc embedding feature with ```POST rpc/getproject?select=id,name,clients{id,name}``` and the relation was not found.

This happened because the tables were created in the```public``` schema and by default the role has that schema in his ```search_path``` and the [detection of the return type only works if the schema is qualified](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/DbRequestBuilder.hs#L52).

This [test](https://github.com/begriffs/postgrest/blob/master/test/Feature/QuerySpec.hs#L515) already fails when adding the ```ALTER ROLE postgrest_test_authenticator set search_path = 'public', 'test';```, that's why I didn't thought of adding a new one.

### DIFF
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -50,8 +50,7 @@ readRequest maxRows allRels allProcs apiRequest  =
           where
             retType = lookup proc allProcs
             tName = case retType of
-              Just (SetOfQI qi) -> qiName qi
-              Just (SetOf item) -> item
+              Just (SetOf (Composite qi)) -> qiName qi
               _ -> proc
 
         _ -> Nothing
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -51,6 +51,7 @@ readRequest maxRows allRels allProcs apiRequest  =
             retType = lookup proc allProcs
             tName = case retType of
               Just (SetOf (Composite qi)) -> qiName qi
+              Just (Single (Composite qi)) -> qiName qi
               _ -> proc
 
         _ -> Nothing
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 module PostgREST.DbRequestBuilder (
-  readRequest
+readRequest
 , mutateRequest
 , fieldNames
 ) where
@@ -11,7 +11,7 @@ import           Control.Lens.Tuple        (_1)
 import qualified Data.ByteString.Char8     as BS
 import           Data.List                 (delete, lookup)
 import           Data.Maybe                (fromJust)
-import           Data.Text                 (isInfixOf, dropWhile, drop)
+import           Data.Text                 (isInfixOf)
 import           Data.Tree
 import           Data.Either.Combinators   (mapLeft)
 
@@ -35,7 +35,7 @@ import           Protolude                hiding (from, dropWhile, drop)
 import           Text.Regex.TDFA         ((=~))
 import           Unsafe                  (unsafeHead)
 
-readRequest :: Maybe Integer -> [Relation] -> [(Text, Text)] -> ApiRequest -> Either Response ReadRequest
+readRequest :: Maybe Integer -> [Relation] -> [(Text, RetType)] -> ApiRequest -> Either Response ReadRequest
 readRequest maxRows allRels allProcs apiRequest  =
   mapLeft apiRequestError $
   treeRestrictRange maxRows =<<
@@ -46,13 +46,12 @@ readRequest maxRows allRels allProcs apiRequest  =
       let target = iTarget apiRequest in
       case target of
         (TargetIdent (QualifiedIdentifier s t) ) -> Just (s, t)
-        (TargetProc  (QualifiedIdentifier s p) ) -> Just (s, t)
+        (TargetProc  (QualifiedIdentifier s proc) ) -> Just (s, tName)
           where
-            returnType = fromMaybe "" $ lookup p allProcs
-            -- we are looking for results looking like "SETOF schema.tablename" and want to extract tablename
-            t = if "SETOF " `isInfixOf` returnType
-              then drop 1 $ dropWhile (/= '.') returnType
-              else p
+            retType = lookup proc allProcs
+            tName = case retType of
+              Just (SetOfQI qi) -> qiName qi
+              _ -> proc
 
         _ -> Nothing
 
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -9,7 +9,7 @@ import           Control.Applicative
 import           Control.Lens.Getter       (view)
 import           Control.Lens.Tuple        (_1)
 import qualified Data.ByteString.Char8     as BS
-import           Data.List                 (delete, lookup)
+import           Data.List                 (delete)
 import           Data.Maybe                (fromJust)
 import           Data.Text                 (isInfixOf)
 import           Data.Tree
@@ -35,7 +35,7 @@ import           Protolude                hiding (from, dropWhile, drop)
 import           Text.Regex.TDFA         ((=~))
 import           Unsafe                  (unsafeHead)
 
-readRequest :: Maybe Integer -> [Relation] -> [(Text, RetType)] -> ApiRequest -> Either Response ReadRequest
+readRequest :: Maybe Integer -> [Relation] -> M.HashMap Text ProcDescription -> ApiRequest -> Either Response ReadRequest
 readRequest maxRows allRels allProcs apiRequest  =
   mapLeft apiRequestError $
   treeRestrictRange maxRows =<<
@@ -48,7 +48,7 @@ readRequest maxRows allRels allProcs apiRequest  =
         (TargetIdent (QualifiedIdentifier s t) ) -> Just (s, t)
         (TargetProc  (QualifiedIdentifier s proc) ) -> Just (s, tName)
           where
-            retType = lookup proc allProcs
+            retType = pdReturnType <$> M.lookup proc allProcs
             tName = case retType of
               Just (SetOf (Composite qi)) -> qiName qi
               Just (Single (Composite qi)) -> qiName qi
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 module PostgREST.DbRequestBuilder (
-readRequest
+  readRequest
 , mutateRequest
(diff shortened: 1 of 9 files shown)


test/fixtures/privileges.sql
```diff
@@ -62,9 +62,12 @@ GRANT USAGE ON SEQUENCE
...
 -- Privileges for non anonymous users
+GRANT USAGE ON SCHEMA test TO postgrest_test_default_role;
 GRANT USAGE ON SCHEMA test TO postgrest_test_author;
 GRANT ALL ON TABLE authors_only TO postgrest_test_author;
 
 GRANT SELECT (article_id, user_id) ON TABLE limited_article_stars TO postgrest_test_anonymous;
 GRANT INSERT (article_id, user_id) ON TABLE limited_article_stars TO postgrest_test_anonymous;
 GRANT UPDATE (article_id, user_id) ON TABLE limited_article_stars TO postgrest_test_anonymous;
+
+ALTER ROLE postgrest_test_authenticator set search_path = 'public', 'test';
```
ruslantalpa: Why is this needed, the search_patth of the authenticator should not have any influence on any request?

test/fixtures/privileges.sql
```diff
@@ -62,9 +62,12 @@ GRANT USAGE ON SEQUENCE
 TO postgrest_test_anonymous;
 
 -- Privileges for non anonymous users
+GRANT USAGE ON SCHEMA test TO postgrest_test_default_role;
```
ruslantalpa: Why is this needed?

ruslantalpa:
Overall this is a good quick fix, the only thing needed here is to add a few tests (a few different procs with different return types) and check that the relation detection on them works.

As a mid term goal i think there is too much logic in haskell for parsing the return types.
The reflection query should provide a description of the proc in a more reliable way without needing to parse strings.

Here is a query that can be built upon

```sql
with compound_types as (
        select t.oid as ooid, attname as att_name, a.oid::regtype as att_type
        from pg_type t
        join pg_class on (reltype = t.oid)
        join pg_attribute on (attrelid = pg_class.oid)
        join pg_type a on (atttypid = a.oid)
    )
    select
        n.nspname as "schema_name",
        p.proname as "procedure_name",
        (SELECT array_to_json(array_agg(cast(n as regtype))) FROM unnest(p.proargtypes) AS n) as "arg_types",
        (SELECT array_to_json(array_agg(cast(n as regtype))) FROM unnest(p.proallargtypes) AS n) as "all_arg_types",        
        array_to_json(p.proargnames) as "arg_names",
        array_to_json(p.proargmodes) as "arg_modes",
        p.prorettype::regtype as "return_type",
        (
            select array_to_json(array_agg(row_to_json(ttt)))
            from
            (
                select c.att_name, c.att_type
                from compound_types as c
                where c.ooid = p.prorettype
            ) ttt
        ) as return_type_components,
        d.description as "description",
        case
            when p.provolatile = 'i' then false
            when p.provolatile = 's' then false
        else true
        end as "is_mutation",
        p.proisstrict as "is_strict",
        p.proretset as "returns_set"
        
    from pg_proc as p
    left join pg_namespace as n on n.oid = p.pronamespace
    left join pg_description as d on d.objoid = p.oid and d.objsubid = 0
    where p.proisagg = false -- no aggregate functions
        and p.proiswindow = false -- no windows functions
        and n.nspname not in ('pg_catalog', 'information_schema')
```

test/fixtures/privileges.sql
steve-chavez: For the tests I proved that It actually does, for example if you do a ```alter role postgrest_test_authenticator set search_path=' '``` some tests fail because they depend on the ```public``` schema being present on the ```search_path```.

I set the ```search_path``` to ```public, test``` to prove that the proc return type now doesn't depend on schema qualification. Being that ```clients``` is in the test schema.

steve-chavez: I fixed that dependency on the ```search_path``` on some of the tests just by schema qualifying some of the definitions on ```fixtures/schema.sql``` but there were some others tests related to computed columns that didn't work, so I leaved that as it is to maybe fix it in another issue.

ruslantalpa: Maybe if you write the tests i will understand better but until now everything is kind of built not not depend on the search schema at all, every component in the query is fully qualified, and search schema should not play a role anywhere. It's especially weird that the search path for the authenticator plays a role since this is a user that should work with absolutley no privileges, or settings.

Maybe add the tests as you have it now then i can better understand what is going on and how to remove this search path dependency

test/fixtures/privileges.sql
steve-chavez: This [test](https://github.com/begriffs/postgrest/blob/master/test/Feature/AuthSpec.hs#L128) failed, it responded with ``` {"hint":null,"details":null,"code":"42501","message":"permission denied for schema test"}```.

Seems that when switching roles the ```postgrest_test_default_role``` had the privilege of only using the ```public``` schema and not the api exposed ```test``` schema.

ruslantalpa:
Nice! looks good to me

ruslantalpa:
(please open an issue in relation to the search path bug in relation to tests and how it's tied to the authenticator role, wee need to clean that up)

ruslantalpa:
The only thing missing are tests. Here is what i think should be tested
If you don't have time for all these tests, maybe add the 2-3 that you think are the most important and i think it's good to merge

- return a scalar
- return a "row" with two scalars
- return a custom type (similar to row, just a few scalars, but this time the type is defined) and the custom type is defined in the exposed schema
- return a custom type, the custom type is defined in a hidden schema
- tables that are exposed in the api schema
  - return a single row from a table (relations  should work as long as the related table is also exposed)
  - return multiple rows from a table (relations should work as long as the related table is also exposed)
  - return a row from a table that has relations to an unexposed table, relations should not work
- the function should return rows from a hidden table (that has relations) but that table and the related tables are exposed as views, this case should transparently work just as if the tables themselves are exposed

steve-chavez:
Thanks for the hints @ruslantalpa, I already included the tests.

ruslantalpa:
awesome work!

begriffs:
Thanks @steve-chavez, I'll look this over tomorrow but guessing it'll be good given the adjustments you guys have already done.

src/PostgREST/DbStructure.hs
```diff
@@ -23,7 +23,7 @@ import           PostgREST.Types
 import           Text.InterpolatedString.Perl6 (q)
 
 import           GHC.Exts                      (groupWith)
-import           Protolude
+import           Protolude
```
begriffs: Looks like some whitespace got in here somehow.

src/PostgREST/DbStructure.hs
```diff
@@ -118,14 +119,30 @@ accessibleProcs =
        else Just $
          PgArg (dropAround (== '"') name) (strip typ) (T.null def)
 
+  parseRetType :: Text -> Text -> Bool -> Char -> RetType
```
begriffs: I like how you've captured this as a new type!

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -35,7 +35,7 @@ import           Protolude                hiding (from, dropWhile, drop)
 import           Text.Regex.TDFA         ((=~))
 import           Unsafe                  (unsafeHead)
 
-readRequest :: Maybe Integer -> [Relation] -> [(Text, Text)] -> ApiRequest -> Either Response ReadRequest
+readRequest :: Maybe Integer -> [Relation] -> [(Text, RetType)] -> ApiRequest -> Either Response ReadRequest
 readRequest maxRows allRels allProcs apiRequest  =
```
begriffs: The `allProcs` argument name seems kind of inaccurate, it's more like `procRetTypes`. Also maybe it's inefficient to do a list lookup each time. What if this argument was prepared ahead of time as a `Map Text RetType` rather than `[(Text, RetType)]`?

I realize the code was like this before your changes, but thought it would be a nice addition to the refactoring. if changing types to Map turns out to be pretty difficult you can skip that. :)

begriffs: In fact what if this argument had type `Map Text ProcDescription`? Would that help for getting tName without conditional logic?

test/Feature/QuerySpec.hs
```diff
@@ -512,25 +512,74 @@ spec = do
...
       it "returns empty json array" $
         post "/rpc/test_empty_rowset" [json| {} |] `shouldRespondWith`
           [json| [] |]
           { matchHeaders = [matchContentTypeJson] }
 
-    context "a proc that returns plain text" $ do
-      it "returns proper json" $
-        post "/rpc/sayhello" [json| { "name": "world" } |] `shouldRespondWith`
-          [json|"Hello, world"|]
+    context "proc return types" $ do
```
begriffs: This extra test coverage is just great.

begriffs:
Look into the `readRequest` argument type and see if my suggestion makes sense. After we decide what to do there then this is good to merge!

src/PostgREST/DbRequestBuilder.hs
steve-chavez: I refactored the ```allProcs``` to ```HashMap Text ProcDescription``` but it didn't help to reduce the tName logic, I thought of working that by adding a ``` [Table] ``` param to ```accesibleProcs``` but the refactor will be larger. I'm going to work on the binary output for procs next, maybe It'll be better to continue the refactor on that PR.

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -46,13 +46,13 @@ readRequest maxRows allRels allProcs apiRequest  =
       let target = iTarget apiRequest in
       case target of
```
begriffs: This is getting better. I tried to simplify it to the following but it causes two test failures:

```haskell
case target of
  (TargetIdent (QualifiedIdentifier s t) ) ->
    Just (s, t)
  (TargetProc  (QualifiedIdentifier s proc) ) ->
    Just (s, fromMaybe proc $ pdName <$> M.lookup proc allProcs)
  _ -> Nothing
```

Is `pdName` sometimes getting set to the incorrect value? The two test failures talk about `pg_source`, like "No relation found between pg_source and client" and "No relation found between pg_source and articleStars."

steve-chavez: For the relation detection to work the ```pdReturnType``` is needed, not the ```pdName```.
The pattern matching on ```pdReturnType``` is also needed since we only need Composite, it includes pg relations as documented in [pg_type](https://www.postgresql.org/docs/9.6/static/catalog-pg-type.html), not Scalars or Pseudo types.

steve-chavez: Though I think that ```toSourceRelation``` already discriminates the retType values to only Tables so the detection might still work for any ```pdReturnType``` but I think that is more clear with the pattern match.

PR #832, opened 2017-03-19
Accept empty string as empty json body when calling an rpc

### LINKED ISSUE #777: Empty body is enforced when calling a non-parameterized rpc?
I'm not sure if this is a feature or a bug. When making an rpc call to a store procedure which does not accept parameters, I have to pass an empty body, otherwise, Postgrest returns an exception.

```sql
CREATE FUNCTION version() RETURNS text
AS $$
  select version();
$$ LANGUAGE sql;
```

This works fine (empty body passed by `-d '{}'`):

```sh
curl -X POST -H "Content-Type: application/json" -d '{}' "http://localhost:3333/rpc/version"
```

This does not:

```sh
curl -X POST -H "Content-Type: application/json" "http://localhost:3333/rpc/version"
```

```json
{"details":"Error in $: not enough input","message":"Cannot parse request payload"}
```

koulakis:
Fixes #777. I added the non-parametric stored procedure mentioned in the issue to the test database and wrote a test that simulates the curl command which makes an rpc call without a json body. Then located the call of the rpc in the request API and mapped the empty ByteString for rpc calls to an empty JSON body.

I added two imports in order to use empty ByteStrings and JSON Values, since I could find no elegant way to build them from the existing libraries. Would be happy to change that if there is some alternative I am missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #791, malformed nested JSON error - @diogob
 - Resource embedding in views referencing tables in public schema - @fab1an
+- #777, Empty body is enforced when calling a non-parameterized RPC - @koulakis
 
 ## [0.4.0.0] - 2017-01-19
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -13,6 +13,7 @@ module PostgREST.ApiRequest ( ApiRequest(..)
 
 import           Protolude
 import qualified Data.Aeson                as JSON
+import           Data.Aeson.Types          (emptyObject)
 import qualified Data.ByteString           as BS
 import qualified Data.ByteString.Internal  as BS (c2w)
 import qualified Data.ByteString.Lazy      as BL
@@ -125,8 +126,11 @@ userApiRequest schema req reqBody
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
       CTApplicationJSON ->
           either Left (\val -> case ensureUniform (pluralize val) of
-            Nothing -> Left "All object keys must match"
-            Just json -> Right json) (JSON.eitherDecode reqBody)
+              Nothing -> Left "All object keys must match"
+              Just json -> Right json)
+            (if BL.null reqBody && isTargetingProc
+              then Right emptyObject
+              else JSON.eitherDecode reqBody)
       CTTextCSV ->
           either Left (\val -> case ensureUniform (csvToJson val) of
             Nothing -> Left "All lines must have same number of fields"
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -125,13 +125,13 @@ userApiRequest schema req reqBody
   payload =
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
       CTApplicationJSON ->
-        (if BL.null reqBody && isTargetingProc
-           then Right emptyObject
-           else JSON.eitherDecode reqBody)
-         >>= note "All object keys must match" . ensureUniform . pluralize
+        note "All object keys must match" . ensureUniform . pluralize
+          =<< if BL.null reqBody && isTargetingProc
+               then Right emptyObject
+               else JSON.eitherDecode reqBody
       CTTextCSV ->
-        CSV.decodeByName reqBody
-        >>= note "All lines must have same number of fields" . ensureUniform . csvToJson
+        note "All lines must have same number of fields" . ensureUniform . csvToJson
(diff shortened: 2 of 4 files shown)


begriffs:
Thank you! That issue has been open for a while, so I appreciate your help.

Could you add an entry to the changelog as well? I'll look at your PR in more detail tomorrow.

koulakis:
Just added it. Thanks for your response!

ruslantalpa:
let's first merge @steve-chavez  work and make sure this one works after

src/PostgREST/ApiRequest.hs
```diff
@@ -125,8 +126,11 @@ userApiRequest schema req reqBody
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
       CTApplicationJSON ->
           either Left (\val -> case ensureUniform (pluralize val) of
-            Nothing -> Left "All object keys must match"
-            Just json -> Right json) (JSON.eitherDecode reqBody)
+              Nothing -> Left "All object keys must match"
+              Just json -> Right json)
+            (if BL.null reqBody && isTargetingProc
+              then Right emptyObject
+              else JSON.eitherDecode reqBody)
```
begriffs: Your change got me thinking...what if we use the monad instance of Either to clean this code up?

```haskell
CTApplicationJSON -> do
  val <- if BL.null reqBody && isTargetingProc
           then Right emptyObject
           else JSON.eitherDecode reqBody
  case ensureUniform (pluralize val) of
    Nothing -> Left "All object keys must match"
    Just json -> Right json
```

What's your opinion? Do you like it better written that way? If so you could conver the `CTTextCSV` case below to this style as well.

koulakis: I think it is a nice refactoring. How do you feel about extending this to the following:
```
import qualified Data.Either.Utils as EU (maybeToEither)

CTApplicationJSON ->
    (if BL.null reqBody && isTargetingProc
         then Right emptyObject
         else JSON.eitherDecode reqBody)
     >>= EU.maybeToEither "All object keys must match" . ensureUniform . pluralize
CTTextCSV ->
    CSV.decodeByName reqBody
    >>= EU.maybeToEither "All lines must have same number of fields" . ensureUniform . csvToJson
```

?

ruslantalpa: there is also this which kind of looks the same, donno which is better
`note :: a -> Maybe b -> Either a b`

ruslantalpa: actually `note` is already there since we use protolude
you should use that and not import anything else

https://github.com/sdiehl/protolude/blob/3994639b2b442f9ee2b0d3a197c5e2ecb579d3b7/src/Exceptions.hs#L30

koulakis: note is definitely better, since it does not add any extra dependency.

src/PostgREST/ApiRequest.hs
```diff
@@ -124,13 +125,13 @@ userApiRequest schema req reqBody
   payload =
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
       CTApplicationJSON ->
-          either Left (\val -> case ensureUniform (pluralize val) of
-            Nothing -> Left "All object keys must match"
-            Just json -> Right json) (JSON.eitherDecode reqBody)
+        (if BL.null reqBody && isTargetingProc
```
begriffs: Nice work with the bind, and also I love that `note` function, been wanting one like that for a while!

What if we reverse the direction of the bind? I think it makes the values appear to flow from right to left cleanly through the composition, plus putting the note first like a kind of comment:

```haskell
CTApplicationJSON ->
  note "All object keys must match" . ensureUniform . pluralize =<<
    if BL.null reqBody && isTargetingProc
       then Right emptyObject
       else JSON.eitherDecode reqBody
CTTextCSV ->
  note "All lines must have same number of fields" . ensureUniform . csvToJson =<<
    CSV.decodeByName reqBody
```

koulakis: Sure, this spares us also the parentheses. I find it a bit more readable though to have the reverse bind on the second line e.g.:
```
note "All lines must have same number of fields" . ensureUniform . csvToJson
          =<< CSV.decodeByName reqBody
```
This makes it immediately clear which operator combines the two lines.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #791, malformed nested JSON error - @diogob
 - Resource embedding in views referencing tables in public schema - @fab1an
+- #777, Empty body is enforced when calling a non-parameterized RPC - @koulakis
```
begriffs: Seems more like the empty body is "allowed" rather than "enforced"

begriffs:
OK, this needs a quick rebase now that I merged another PR and then it's good to go!

PR #834, opened 2017-03-20
Increase Warp connection timeout

### LINKED ISSUE #827: postgrest: Thread killed by Warp's timeout reaper
I noticed these strange logs while running PostgREST `master` (0.4.x):

```
postgrest: Thread killed by Warp's timeout reaper
```

I ran PostgREST with:

```
stack exec postgrest ../postgrest-config/localhost.config
```

on macOS Sierra (version 10.12.2).

majorcode:
I was able to find a few references to "Thread killed by Warp's timeout reaper" in other projects. One (Hets) resolved the issue by increasing the Warp connection timeout to one day. I'll explain and attempt to justify this below.

If accepted, this pull request resolves #827 in the same way as referred to in a Yesod [issue](https://github.com/yesodweb/wai/issues/351) and [resolved](https://github.com/spechub/Hets/pull/1520) in the Hets project.

---

Some browsers keep TCP connections open for a while, probably to improve the speed of subsequent requests to the same site. This is a reasonable practice.

Warp apparently has a feature to kill (reap) long running connections. The default connection timeout is 60 seconds. I can't find code listing the 60 second default--I see a [default of 30 seconds](https://github.com/yesodweb/wai/blob/3736ea965049777f8a5d95e6f645c04a940608d4/warp/Network/Wai/Handler/Warp/Settings.hs#L133). Perhaps it's being applied twice inside Warp? The Hets project also [observed a timeout message](https://github.com/spechub/Hets/issues/1482) after 60 seconds.

Connection timeouts cause error messages printed to stderr. The messages are harmless notifications from Warp stating that it closed a connection after it's timeout expired. Affected clients/browsers can simply re-connect as needed. PostgREST's ability to serve requests is unaffected. But the messages sound scary and are logged as errors, possibly implying that action is needed to resolve a problem.

Why would Warp do this? It's one step in its protection from a Slow Loris attack (or the like) where an attacker keeps a large number of connections open in an attempt to deny service through connection starvation.

Slow Loris protection is a good thing. But I'm not certain PostgREST needs this feature.

First, with a need for SSL termination on the web, you really should have PostgREST behind a proxy--like Nginx. The proxy has to have Slow Loris protection. In the case of NginX, you can limit the number of requests from the same IP (and do more to [protect it](https://www.nginx.com/blog/mitigating-ddos-attacks-with-nginx-and-nginx-plus/)).

Second--and this is where I have technical questions about how Warp uses Haskell for connection threads--I think the thread overhead in Haskell is so small that it would be very difficult to trigger connection exhaustion. It seems to me that the ulimit would be reached first.

Finally, in my informal tests, I can't get PostgREST to fall over under load. I can get it to use all of my CPU. But, if I'm requesting any end point that hits the database, the database dominates resource usage. Postgres would probably kick the bucket first. Consequently, more exotic DDoS protection provided by a proxy is in order. I believe DDoS prevention features are out of scope for PosgREST.

---

### Is this solution acceptable?
I can not find a way to disable the timeout completely. So a high limit is the next best thing.

### Is 24 hours a good number? 
Would a lower number be better? 24 hours should eliminate all error logs, unless you have a long running dashboard or IoT thing using PostgREST. The best number would be just over the duration of your max site visit. But that varies. Is there a reference to the timeout value used by browsers? I cold not find a reliable source.

### Should it be configurable?
I don't think so because of the need for other DDoS protection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- #827, Avoid Warp reaper, extend socket timeout to 1 hour - @majorcode
 - #791, malformed nested JSON error - @diogob
 - Resource embedding in views referencing tables in public schema - @fab1an
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -57,7 +57,7 @@ main = do
       appSettings = setHost ((fromString . toS) host)
                   . setPort port
                   . setServerName (toS $ "postgrest/" <> prettyVersion)
-                  . setTimeout 86400
+                  . setTimeout 3600
                   $ defaultSettings
 
   when (isMalformedProxyUri $ toS <$> proxy) $ panic
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -57,6 +57,7 @@ main = do
       appSettings = setHost ((fromString . toS) host)
                   . setPort port
                   . setServerName (toS $ "postgrest/" <> prettyVersion)
+                  . setTimeout 86400
                   $ defaultSettings
 
   when (isMalformedProxyUri $ toS <$> proxy) $ panic
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-8.0
+resolver: lts-8.5
 extra-deps:
   - Ranged-sets-0.3.0
   - hasql-pool-0.4.1


ruslantalpa:
at first, when i saw 24h i was like woo, hold on there cowboy :) but you are right in everything you said
- it needs to be behind nginx
- a new connection is not that expensive in warp


this also makes it a bit better when you have it behind nginx because
to get good performance there, one would have
```
proxy_set_header  Connection ""; #optimise communication with upstream (keep alive)
proxy_http_version 1.1;
```

and also in the upstream configuration
```
keepalive 64;
```

so it's a good thing that the connections nginx keeps open to postgrest do not get killed every 30 sec.

This probably won't yield any performance increase but it's probably ok to have it merged, but probably not 24h, maybe like 1h for starters would be enough

i don't think we need to complicate the config with this minor thing.

majorcode:
An hour sounds reasonable too. A client would have to be inactive for that long to cause a message. You may see it in development if you leave Postman open. I don't really know how long Postman (Chrome) keeps the connection alive.

I just think 24h sends a message that timeouts are purposely turned off.

steve-chavez:
This seems already fixed in warp 3.2.11.1, see the [PR](https://github.com/yesodweb/wai/issues/613), wouldn't be better to just upgrade warp version?

steve-chavez:
I compiled with a 8.5 lts in https://github.com/begriffs/postgrest/blob/master/stack.yaml#L1, and I didn't get the ```Thread killed by Warp's timeout reaper```, I think 8.3 could work too.

majorcode:
Wouldn't that mean we're bumping the Haskell version too? Do we have a green light on that? there may be other regressions.

majorcode:
As @steve-chavez says, the reaper error message doesn't appear if we upgrade Haskell and pull in the latest Warp. I'm new to Haskell. But I'm assuming that's what changing to `lts-8.5` accomplishes. The test suite runs successfully too.

ruslantalpa:
While the error going a way is a good thing, the change from 30s to 1h might still be valuable.
or at least something like 5min

majorcode:
I'm not well versed in Haskell (yet) or Warp's code. But, in reading the PR @steve-chavez referenced, the error message problem may have been from a regression in Warp that stops logging the error.

I can't be certain because there should still be some protection from Slow Loris. It's possible they still terminate the threads. But they've stopped emitting the error. That would mean that raising the timeout is still a good idea to cover things like @ruslantalpa's Nginx keepalive config.

I can change it to 3600, if you would approve.

Also, was I correct about the `lts-8.5` change meaning that we're changing the compiler version? Like I said, I'm only a part-time week into learning Haskell. I know the tests run. But I've been bitten by compiler changes and have thick C++ scar tissue.

begriffs:
Actually both LTS-8.0 and LTS-8.5 use the same underlying GHC version of 8.0.2. The only difference in the LTS versions is that they change the pinned versions of some libraries, like warp-3.2.11 vs warp-3.2.11.1. So the upgrade should be safe.

Nice measurements of load etc, @majorcode, sounds like we could raise the timeout to 3600 like you said.

ruslantalpa:
since this pr is about that error, the lts upgrade should probably go here also?

majorcode:
Should I make a second PR to change stack.yml yo use `lts-8.5`?

majorcode:
@begriffs I though @ruslantalpa was already answering my question about 8.5. Now I think he may have just been asking you the same question. I can roll this back if need be.

begriffs:
Bumping the version is fine. Thanks for all the research about this issue everyone.

PR #836, opened 2017-03-23
Upgrade to fast new hasql

begriffs:
Upgrade for the speed improvements prior to releasing a new postgrest version.

@ruslantalpa note that I had to use a stackage snapshot plus hold some other packages' versions back. I'd kind of like to wait for hasql to appear in stackage again and work cleanly with the other packages...

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
     - stack setup
     - rm -fr $(stack path --dist-dir) $(stack path --local-install-root)
     - stack install hlint packdeps cabal-install
-    - stack build --fast
+    - stack build --fast -j1
     - stack build --fast --test --no-run-tests
 
 test:
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -142,7 +142,7 @@ Test-Suite spec
                      , hasql-pool
                      , heredoc
                      , hjsonpointer
-                     , hjsonschema
+                     , hjsonschema == 1.5.0.1
                      , hspec
                      , hspec-wai >= 0.7.0
                      , hspec-wai-json
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TupleSections #-}
---module PostgREST.App where
+
 module PostgREST.App (
   postgrest
 ) where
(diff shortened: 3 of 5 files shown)


ruslantalpa:
yes, i also am for waiting on 8.6
let's leave this PR open, maybe i'll have a day soonish to create a simple performance test specific for postgrest (i.e. simplify/refactor what i have)

begriffs:
Just checked again today and LTS 8.11 is still back at hasql 0.19.16

begriffs:
Getting close now! it works with the latest nightly. An LTS shouldn't be too far away.

begriffs:
Gr, why is the OpenAPI spec failing again? I've rebased this on master where it was fixed...

ldesgoui:
OpenAPI package returns valid json from what I'm seeing, the problem might come from this (haven't looked in depth yet, will later today):
https://github.com/seagreen/hjsonschema/compare/1.5.0.1...master#diff-64ecbe36b1efdb402197771823df789a

begriffs:
Checked Stackage again today and even LTS-8.21 is still stuck at `hasql-0.19.16`!

begriffs:
The good news: LTS 9.0 now has the newest hasql and rebasing makes this a simple change.

The bad news: building the whole thing on a new LTS times out consistently in CI.

ldesgoui:
It seems to hit the 4G memory limit, maybe adding -j1 to `stack build` might help ?
commercialhaskell/stack#1910

begriffs:
@ldesgoui nice find! That got past the build failure.

Now we're hitting a fake OpenAPI validation failure...

ruslantalpa:
the openapai validation is don i think against a hard-coded schema.
meybe with the new lts, a new openapi lib came in that produces a slightly different output.
it might be a case of just adding the changed parts to this file https://github.com/begriffs/postgrest/blob/master/test/fixtures/openapi.json

begriffs:
OK, this is unsatisfying but it works: pin hjsonschema back at 1.5.0.1. I validated output from the latest version of the library online and it looks fine, but somehow fails the tests. Perhaps we need to update helper files like @ruslantalpa suggests.

The changelog for the package mentions this, if it's helpful in tracking down the problem:

_____
+ Fix defect where validators alongside "$ref" weren't ignored.
+ Fix defect where local references would fail if an "id" key had set resolution scope to start from a different document.
+ Vendor latest tests.
+ Remove `ReferencedSchemas`.
+ Create `Scope` to hold information that changes during validation.
+ Use a sum type for "type" values. Thanks to Philip Weaver (GitHub @pheaver).
_____

We could also just merge this as it is because it will keep the hjsonschema version unchanged but improve the hasql version.

ruslantalpa:
i vote for merging as is and opening an issue to remember to unpin the openapi lib version.
and since this release is all about performance, hold off for 1-2 days before making it, i might add a PR with some query speed improvements.

Issue #547, opened 2016-04-04
Postgrest error when connected to read replica database


paulmin55:
Hello, 

I'm trying to run the lastest Postgrest v0.3.1.1 on Debian GNU/Linux 8 to use a Postgres 9.4.5 read replica db for read only queries but I'm getting this message when curling postgrest at:

**Steps to reproduce:**
``` 
   $curl http://localhost:3000 
{"hint":null,"details":null,"code":"0A000","message":"cannot set transaction read-write mode during recovery"}.
```

I have no trouble using postgrest against the master db running Postgres 9.4.5, however I was hoping I can use Postgrest with the read replica db to perform read-only operations.   

I've chatted with @ruslantalpa on glitter and he suspects it might be an issue with how postgrest issues the 'begin transaction' statement and requested that I put the postgres query logs as shown below.  I appreciate any advise on this.  Thank you!

**Postgres read replica db query log when  running curl http://localhost:3000:**
```
[28104]:LOG: execute 6: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
2016-04-04 15:21:11 UTC:xxx.xxx.xxx.xxx(42946):read_only@database:[28104]:ERROR: cannot set transaction read-write mode during recovery
```

**Successful postgres query logs when running curl http://localhost:3000: on master db:**
```
2016-04-04 15:04:31 UTC:xxx.xxx.xxx.xxx(36131):read_only@database:[21057]:DETAIL: parameters: $1 = 'public'
2016-04-04 15:04:31 UTC:xxx.xxx.xxx.xxx(36131):read_only@database:[21057]:LOG: execute 8: commit
2016-04-04 15:04:31 UTC:xxx.xxx.xxx.xxx(36131):read_only@database:[21057]:LOG: execute 6: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
2016-04-04 15:04:31 UTC:xxx.xxx.xxx.xxx(36131):read_only@database:[21057]:LOG: statement: set local role ‘read_only’;
2016-04-04 15:04:31 UTC:xxx.xxx.xxx.xxx(36131):read_only@database:[21057]:LOG: execute 7: 
select
n.nspname as table_schema,
relname as table_name,
c.relkind = 'r' or (c.relkind IN ('v', 'f')) and (pg_relation_is_updatable(c.oid::regclass, false) & 8) = 8
or (exists (
select 1
from pg_trigger
[...]

PR #844, opened 2017-03-28
Read only proc tx

begriffs:
A read-only standby postgresql server does not support read-write transactions. Our code used to ask for these kind of transactions to call stored procs, but this PR checks the volatility class of the function and uses read-only transactions when possible.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -110,7 +110,7 @@ accessibleProcs =
                             HD.value HD.bool <*>
                             HD.value HD.char)
                         <*> (parseVolatility <$>
-                            HD.value HD.text)
+                            HD.value HD.char)
       )
     ) True
  where
@@ -140,10 +140,10 @@ accessibleProcs =
         'p' -> Pseudo name
         _   -> Scalar qi -- 'b'ase, 'd'omain, 'e'num, 'r'ange
 
-  parseVolatility :: Text -> ProcVolatility
-  parseVolatility "i" = Immutable
-  parseVolatility "s" = Stable
-  parseVolatility "v" = Volatile
+  parseVolatility :: Char -> ProcVolatility
+  parseVolatility 'i' = Immutable
+  parseVolatility 's' = Stable
+  parseVolatility 'v' = Volatile
   parseVolatility _   = Volatile -- should not happen, but be pessimistic
 
   sql = [q|
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -100,10 +100,19 @@ decodeSynonyms cols =
 accessibleProcs :: H.Query Schema (M.HashMap Text ProcDescription)
 accessibleProcs =
   H.statement sql (HE.value HE.text)
-    (M.fromList . map addName <$> HD.rowsList (ProcDescription <$> HD.value HD.text
-                                <*> (parseArgs <$> HD.value HD.text)
-                                <*> (parseRetType <$> HD.value HD.text <*> HD.value HD.text <*>
-                                                      HD.value HD.bool <*> HD.value HD.char))) True
+    (M.fromList . map addName <$>
+      HD.rowsList (
+        ProcDescription <$> HD.value HD.text
+                        <*> (parseArgs <$> HD.value HD.text)
+                        <*> (parseRetType <$>
+                            HD.value HD.text <*>
+                            HD.value HD.text <*>
+                            HD.value HD.bool <*>
+                            HD.value HD.char)
+                        <*> (parseVolatility <$>
+                            HD.value HD.text)
+      )
+    ) True
  where
   addName :: ProcDescription -> (Text, ProcDescription)
   addName pd = (pdName pd, pd)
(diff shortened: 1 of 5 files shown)


jhannah:
Thank you @begriffs!! Like #547 I can build from source with these patches to see if the problem disappears. :)

src/PostgREST/DbStructure.hs
```diff
@@ -124,25 +133,32 @@ accessibleProcs =
...
+    where
       qi = QualifiedIdentifier schema name
-      pgType = case typ of 
+      pgType = case typ of
         'c' -> Composite qi
         'p' -> Pseudo name
         _   -> Scalar qi -- 'b'ase, 'd'omain, 'e'num, 'r'ange
-          
+
+  parseVolatility :: Text -> ProcVolatility
```
steve-chavez: Instead of ```Text``` a ```Char``` wouldn't be better for this case?

jhannah:
It appears my problem isn't `/rpc/` specific, and is further back in history than this branch. v0.3.2.0 runs against our AWS RDS read-only slave just fine (`/rpc/` untested), but v0.4.0.0 is unusable:

```
$ curl http://localhost:3001/
{"hint":null,"details":null,"code":"0A000","message":"cannot set transaction read-write mode during recovery"}
```

I'm attempting a manual "git bisect" style investigation of building from source to try to find which commit between v0.3.2.0 and v0.4.0.0 triggers the difficulty. (See #547 for the original report against this environment.)

begriffs:
The root route translates into what we call internally `ActionInspect`, and that falls back into a read-write transaction. Does accessing a table like `/foo` work properly? Also does the current branch fix the behavior for RPC? I'll make a change to deal with the root request too.

jhannah:
@begriffs progress! On d9efbc1: `/` throws swagger stuff (yay!), `/foo` works (yay!), but `/rpc/foo` is still unhappy:

```
curl -v -d '{"strategy_id":77}' \
   http://localhost:3001/rpc/dimensions_for_strategy_id
{
    "code": "0A000",
    "details": null,
    "hint": null,
    "message": "cannot set transaction read-write mode during recovery"
}

WARNING:  there is no transaction in progress
127.0.0.1 - - [31/Mar/2017:19:30:42 +0000] "POST /rpc/dimensions_for_strategy_id HTTP/1.1" 400 - "" "curl/7.38.0"
```
I'm digging around trying to figure out what I can do where to give you more information.

jhannah:
Hmm... so I built an AWS lab w/ a master RDS, a slave RDS, and an EC2 for compiling PostgREST. I compiled  d9efbc1 and set `log_statement=all`, but it doesn't appear to me that it's spewing very helpful stuff:
```
curl -v -d '{"id":1}' \
   http://localhost:3000/rpc/hello

WARNING:  there is no transaction in progress
127.0.0.1 - - [31/Mar/2017:23:35:12 +0000] "POST /rpc/hello HTTP/1.1" 400 - "" "curl/7.38.0"

2017-03-31 23:35:13 UTC:172.30.0.221(29645):awsuser@jaytest:[29581]:LOG: statement: SET client_encoding = 'UTF8';SET client_min_messages TO WARNING;
2017-03-31 23:35:13 UTC:172.30.0.221(29645):awsuser@jaytest:[29581]:LOG: execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
2017-03-31 23:35:13 UTC:172.30.0.221(29645):awsuser@jaytest:[29581]:ERROR: cannot set transaction read-write mode during recovery
2017-03-31 23:35:13 UTC:172.30.0.221(29645):awsuser@jaytest:[29581]:STATEMENT: BEGIN ISOLATION LEVEL READ COMMITTED READ WRITE
2017-03-31 23:35:13 UTC:172.30.0.221(29645):awsuser@jaytest:[29581]:LOG: execute 1: ABORT
2017-03-31 23:35:13 UTC:172.30.0.221(29645):awsuser@jaytest:[29581]:WARNING: there is no transaction in progress
```

begriffs:
Have you tried applying all the logging settings mentioned here? https://postgrest.com/en/v0.4/admin.html#debugging

jhannah:
@begriffs in AWS RDS I think only `log_statement=all` applies because it's running in their magic cloud, not on a Linux box logging to disk. 
I also tried these and don't see anything helpful in there:
* sudo ngrep -d lo port 3000 (redundant with curl output) 
* sudo ngrep -d eth0 port 5432 (mostly encrypted talking to DB)

jhannah:
In case it matters, my schema is:
```
CREATE TABLE names (
  id SERIAL NOT NULL PRIMARY KEY,
  fname varchar(20),
  lname varchar(20)
);
INSERT INTO names (fname, lname) values ('Jay', 'Hannah');
CREATE FUNCTION hello(name_id integer)
  RETURNS TABLE(fname char(20), lname char(20)) AS $$
SELECT fname, lname
FROM names
WHERE id = $1
$$ LANGUAGE SQL;
```

jhannah:
@begriffs Happy to grant you ssh & psql access to this AWS lab, if you want it. I don't know what else I can provide at this point, other than debugging the Haskell itself, which I can try to wade into. #newb :)

steve-chavez:
I think that by default a function is VOLATILE, maybe try creating a function and marking it as IMMUTABLE or STABLE.

jhannah:
!!! THANK YOU @steve-chavez! `CREATE FUNCTION ... STABLE` fixes all known problems against 00c4288! I didn't know STABLE was a thing! Thanks! :)

begriffs:
I verified in the logs that this branch of postgrest is now using the correct transaction mode for functions. I changed one of the functions in our test fixture to be immutable just to test that the new code doesn't break the RPC. It's not a full test of the functionality here, but it's a basic sanity check.

jhannah:
( These commits were all squased into 56bdf59e. )

begriffs:
Yep, is it working OK for you?

Issue #800, opened 2017-02-10
Access HTTP headers inside PSQL


tjunussov:
Like in Oracle PL/SQL via UTL_HTTP.GET_HEADER,  Is it possible to pass all header information to access them via psql function, smth like :
```
current_settings("http_host");
current_settings("http_referer");

```

begriffs:
I thought there was a github issue about this already, but I can't find it...

Anyway, this is a good idea. We could prefix the name of the variables `request.header.http_referer` so that it is similar to how we store JWT claims as e.g. `request.jwt.claim.email`. One thing is we'll want to avoid making an extra roundtrip to the database to set the variables if we can help it, but it's probably doable.

julesbou:
We could also set IP address of current request. It can be useful to save IPs in the database, for example when you insert a comment on a forum. Seems like it's not possible to do that with postgrest yet (I might be wrong).

eldadj:
@ruslantalpa fyi

ruslantalpa:
@eldadj see #849 
until it's merged, you can use the patch.

spinus:
Looks like it's implemented and released in 0.4.1, should we close the ticket?
@tjunussov maybe you would like to verify?

begriffs:
Closing. Can re-open if anybody runs into a problem with it.

NikolayS:
For those (like me) who wonder how to use it, this one works:
```
current_setting('request.header.user-agent', true)
```

PR #849, opened 2017-03-29
Make HTTP headers available as GUCs #800

ruslantalpa:
This PR makes request headers available as GUC vars for each request
It also sets `request.jwt.claims.role`, it useful when doing RLS through views.

Please look at runWithClaims from middleware if it needs some nicer style/formatting

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -97,7 +97,7 @@ data ApiRequest = ApiRequest {
   -- | HTTP request headers
   , iHeaders :: [(Text, Text)]
   -- | Request Cookies
-  , iCookies :: Maybe [(Text, Text)]
+  , iCookies :: [(Text, Text)]
   }
 
 -- | Examines HTTP request and translates it into user intent.
@@ -126,7 +126,7 @@ userApiRequest schema req reqBody
         $ rawQueryString req
       , iJWT = tokenStr
       , iHeaders = [ (toS $ CI.foldedCase k, toS v) | (k,v) <- hdrs, k /= hAuthorization, k /= hCookie]
-      , iCookies = parseCookiesText <$> lookupHeader "Cookie"
+      , iCookies = fromMaybe [] $ parseCookiesText <$> lookupHeader "Cookie"
       }
  where
   isTargetingProc = fromMaybe False $ (== "rpc") <$> listToMaybe path
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -40,6 +40,7 @@ import           PostgREST.Types           ( QualifiedIdentifier (..)
                                            , ApiRequestError(..)
                                            , toMime)
 import           Data.Ranged.Ranges        (Range(..), rangeIntersection, emptyRange)
+import qualified Data.CaseInsensitive      as CI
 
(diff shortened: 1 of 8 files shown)


begriffs:
Thanks Ruslan, that's a long-requested feature. I'll review the code in a day or two.

eldadj:
@ruslantalpa Thanks for this!

src/PostgREST/Middleware.hs
```diff
@@ -48,6 +55,13 @@ runWithClaims conf eClaims app req =
       ]
       (toS $ "{\"message\":\""<>message<>"\"}")
 
+envVarToSql :: Text -> [(Text, Text)] -> [ByteString]
```
begriffs: Good idea to extract this into its own function. Maybe it makes more sense to put it in the QueryBuilder module, with return type `[SqlFragment]`?

Also `envVarToSql` sounds confusingly like unix environment variables. What about `setSqlVars` or something?

Adding a comment describing the motivation for the function would be nice, like how it's used for setting the role, jwt claims, and http request headers as GUCs.

src/PostgREST/Middleware.hs
```diff
@@ -48,6 +55,13 @@ runWithClaims conf eClaims app req =
...
+envVarToSql prefix vars = varStmts
```
begriffs: This is minor, but you could inline `varStmts` as `map setVar vars` here. The extra name doesn't get us too much.

src/PostgREST/Auth.hs
```diff
@@ -12,8 +12,8 @@ In the test suite there is an example of simple login function that can be used
 very simple authentication system inside the PostgreSQL database.
 -}
 module PostgREST.Auth (
-    claimsToSQL
-  , containsRole
+    -- claimsToSQL
```
begriffs: You could remove this comment.

ruslantalpa:
the build fail is due to some temporary network issue. Trigger rebuild in CircleCI and it should work

src/PostgREST/ApiRequest.hs
```diff
@@ -92,6 +94,10 @@ data ApiRequest = ApiRequest {
   , iCanonicalQS :: ByteString
   -- | JSON Web Token
   , iJWT :: Text
+  -- | HTTP request headers
+  , iHeaders :: [(Text, Text)]
+  -- | Request Cookies
+  , iCookies :: Maybe [(Text, Text)]
```
begriffs: What's the difference in meaning here between `Nothing` and `Just []`?

ruslantalpa: No meaning, it was kind of an effect of the return type of the function used. Probably [] will be better

begriffs:
Looks good! I made a note for myself to document this new behavior for the v0.4.1 docs

pocman:
@begriffs Hi, I can't find the documentation for this :/
`current_setting(request.header.x-my-header)` point to an undefined_object.

begriffs:
@pocman I added a little section about it to the docs: https://postgrest.com/en/v4.1/api.html#accessing-request-headers-cookies

In your case are you able to read some headers but not others, or are all headers coming back undefined?

pocman:
Hi, it was just a typo 😉
Thanks. It's working fine.

PR #852, opened 2017-04-03
Update with empty json

### LINKED ISSUE #493: Updating with empty JSON object generates invalid sql
Patching with `{}` should be a 400 Bad Request not a server error.

koulakis:
Fixes #493. 

While running the new tests noticed that the table `items`, which is being used by several tests of the update requests, changes between those tests. Made some corrections but could not discover why `id:16` is being added to it. For the time being I just ignored that.


begriffs:
This is really nice. Sorry the tests can be fragile and interact with one another. You might want to create a new table in the fixture just for this test, that way the server response won't have stray data from other tests.

begriffs:
Could you do a quick rebase to resolve the simple conflict in the changelog? (I opened #859 to prevent these kind of conflicts in the future)

koulakis:
Done!

PR #853, opened 2017-04-03
Adds application/json to inspect type

feynmanliang:
The latest version of swagger-ui https://github.com/swagger-api/swagger-ui/tree/v3.0.4 cannot consume the OpenAPI spec provided at `/api/`, throwing a content type error in the browser:
```
Failed to load resource: the server responded with a status of 415 (Unsupported Media Type)
GET http://localhost/api/ 415 (Unsupported Media Type)
```

The response sent by `postgrest` shows

```
{"message":"None of these Content-Types are available: application/json"}
```

This is caused by the [Content-Type request by swagger-ui](https://github.com/swagger-api/swagger-ui/blob/e23dada9fed460a5e4a88809db362ee75ccd08aa/src/core/plugins/download-url.js#L18).

This PR adds support for this.

---
This change is [Reviewable](https://reviewable.io/reviews/begriffs/postgrest/853)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Allow requesting binary output on GET - @steve-chavez
+- Accept clients requesting `Content-Type: application/json` from / - @feynmanliang
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


begriffs:
Thanks for the contribution!

I believe the fact of Swagger only accepting `application/json` was a regression that will be fixed (https://github.com/swagger-api/swagger-ui/issues/2798). But it seems cool for us to be more permissive and serve the result as JSON if the client is picky.

Anybody else have an objection?

FGRibreau:
> Anybody else have an objection?

No objection at all :+1:

begriffs:
Cool. @feynmanliang can you add a line to the changelog as well? Then I'll merge.

feynmanliang:
Of course :)

PR #854, opened 2017-04-05
Fix #701: allow quoted values for IN operator, refactor and Fix #641: allow IN filter to have no values

### LINKED ISSUE #641: Allow 'in.' filter to have no items
When `in.` is not followed by any characters, treat it as the empty set.

---

Related to https://github.com/marmelab/admin-config/issues/71

I'm using the code:

``` js

                if (params._filters[filter] instanceof Array) {
                    params[filter] = "in." + params._filters[filter].join(",");
                } else {
                    params[filter] = "eq." + params._filters[filter];
                }
```

However, if the filter list is empty, it ends up requesting the path `/organisation?id=in.`. This comes back from postgrest as a 400 error:

``` json
{"hint":null,"details":null,"code":"22P02","message":"invalid input syntax for uuid: \"\""}
```

### LINKED ISSUE #701: "in" filter cannot express strings containing commas
Assume I have a relation "public.names", containing a single text column, "name".
I want to retrieve two names `"x,y"` and `"a,b"`, using the in operator, so I make this request:

```
GET /names?name=in.x,y,a,b
```

This returns nothing, because `"x,y"` is parsed as two values. It seems there is no way to specify a string
containing a comma.

I have three ideas for solutions:
1. Define escaping rules for strings like "\" means backslash, and "\," means literal comma. This would break backwards compatibility (unless a header could be added)
2. Allow the separator character to be set in a HTTP header
3. Allow filters to be specified as JSON data

For 3, that could look something like this:

```
[
    {"column": "name", "operator": "in", "value": ["x,y", "a,b"]},
    {"column": "age",    "operator": "lt", "value": 3}
]
```

Personally, I feel like 3 is better - it could be a simple "filter DSL", which would have two frontends - the existing URL strings, and a JSON representation.

However, it might be nice to also have a patch for 2 simply allowing the separator for the "in" operator to be set via a header - it's much less work to implement and solves the immediate problem very quickly. I can probably implement this if needed!

Hopefully it's possible already and I wasted my time opening this issue though :-)

steve-chavez:
While working on the ```where``` param I noticed that it may suffer from this lacking #701 as well, for example in ```people?where=and{name=in.Thompson, Larry,or{age=gte.18,height=lte.180}}``` or even ```..eq.Thompson, Larry,or{age..``` I'll have to consider that later.

Because of that I thought on working this first and also do some refactoring, the ```where``` param may combine different parsers so I'm trying to be cautious and not make my implementation a source of bugs or maintainability problems.

### DIFF
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -79,9 +79,7 @@ data Column =
     , colDefault   :: Maybe Text
     , colEnum      :: [Text]
     , colFK        :: Maybe ForeignKey
-    }
-  | Star { colTable :: Table }
-  deriving (Show, Ord)
+    } deriving (Show, Ord)
 
 type Synonym = (Column,Column)
 
@@ -253,7 +251,6 @@ instance Eq Table where
 
 instance Eq Column where
   Column{colTable=t1,colName=n1} == Column{colTable=t2,colName=n2} = t1 == t2 && n1 == n2
-  _ == _ = False
 
 -- | Convert from ContentType to a full HTTP Header
 toHeader :: ContentType -> Header
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -1,6 +1,7 @@
 module PostgREST.Types where
 import           Protolude
 import qualified GHC.Show
+import qualified GHC.Read
 import           Data.Aeson
 import qualified Data.ByteString.Lazy as BL
 import           Data.HashMap.Strict  as M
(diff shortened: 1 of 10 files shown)


src/PostgREST/Types.hs
```diff
@@ -134,8 +133,66 @@ data Proxy = Proxy {
...
+    "like" -> [(Like, "")]
+    "ilike" -> [(ILike, "")]
+    "in" -> [(In, "")]
+    "notin" -> [(NotIn, "")]
+    "isnot" -> [(IsNot, "")]
+    "is" -> [(Is, "")]
+    "@@" -> [(TSearch, "")]
+    "@>" -> [(Contains, "")]
+    "<@" -> [(Contained, "")]
+    _ -> []
```
steve-chavez: This doesn't respect the readsPrec method implementation but I thought of leaving it that way since I only use ```read``` for single values, previously I had:

```haskell
instance Read Operator where
  readsPrec _ op =  case op of
    'e':'q':r -> [(Equals, r)]
    'g':'t':'e':r -> [(Gte, r)]
    'g':'t':r -> [(Gt, r)]
    'l':'t':'e':r -> [(Lte, r)]
    'l':'t':r -> [(Lt, r)]
    'n':'e':'q':r -> [(Neq, r)]
    'l':'i':'k':'e':r -> [(Like, r)]
    'i':'l':'i':'k':'e':r -> [(ILike, r)]
    'i':'n':r -> [(In, r)]
    'n':'o':'t':'i':'n':r -> [(NotIn, r)]
    'i':'s':'n':'o':'t':r -> [(IsNot, r)]
    'i':'s':r -> [(Is, r)]
    '@':'@':r -> [(TSearch, r)]
    '@':'>':r -> [(Contains, r)]
    '<':'@':r -> [(Contained, r)]
    _ -> []
```

I couldn't come up with a way of making this more readable.

ruslantalpa:
Very nice refactoring.
Only thing that seems repetitive is the operator related stuff types.
I see basically 3 forms to represent an operator
`like / Like / LIKE` and the definitions and transitions between them are spread between a few places.

Would it be possible to have the definition somehow only in a single place
with an array of `(Like, "like", "LIKE")` and then have a function that that translates form one representation to another and use that function in the definition of the Read/Show instances and the other function that does some translation for the parsers?

Having it this way, will make adding new operators easier if the need arises.

@begriffs what do you think?

steve-chavez:
@ruslantalpa By defining the ```Operator``` type and its constructors I meant to make the ```Like``` constructor in your example the only representation of the querystring operator ```like``` that postgREST sees. The Read/Show instances are just for obtaining that representation in the Parsers, and the only transition I saw was from that querystring ```Operator``` to his SQL operator. I could make this clear maybe by renaming ```Operator``` to ```QSOperator``` and adding comments. 

I get that there are 3 functions now to handle that instead of the ```operators``` tuples but my main motivation for converting this to types was to avoid having the [pgFmtOperator](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/QueryBuilder.hs#L470-L473) that lookups the operators and instead just pattern match to convert the ```Operator``` to SQL like in [opToSqlFragment](https://github.com/steve-chavez/postgrest/blob/in-quoted/src/PostgREST/Types.hs#L177), I thought it was better to have that transition closer to the types.

if I where to do the ```(Like, "like", "LIKE")``` array I think the ```Like``` constructor wouldn't add any value and maybe it'll be better to just delete it and go back to the ```operators``` tuples.

begriffs:
I merged Ruslan's PR because I'd like to begin some work on top of it. Could you rebase please?

steve-chavez:
Done.

src/PostgREST/Types.hs
```diff
@@ -138,8 +137,66 @@ data Proxy = Proxy {
 , proxyPath       :: Text
 } deriving (Show, Eq)
 
-type Operator = Text
-data FValue = VText Text | VForeignKey QualifiedIdentifier ForeignKey deriving (Show, Eq)
+data Operator = Equals | Gte | Gt | Lte | Lt | Neq | Like | ILike | Is | IsNot |
+                TSearch | Contains | Contained | In | NotIn deriving (Eq, Enum)
+
+instance Show Operator where
```
begriffs: Can `show` and `readsPrec` both consult a common list of tuples? Their definitions have duplicate code.

begriffs: Oops, I see you already discussed the tradeoffs of this and decided the pattern match is better.

begriffs:
Looks good. Adding a note for myself to update the docs to mention quoting the IN values.

PR #795, opened 2017-02-05
Update DbStructure.hs

krakrjak:
Backport column name detection in public schema for 0.3.2.1

PR #856, opened 2017-04-05
0.3.x

krakrjak:
Backport fixes from 4.0 to 0.3.x line with the exception of jwt handling code. The current status should not be merged unless you accept the following test failures:
```
  test/Feature/InsertSpec.hs:65: 
  1) Feature.InsertSpec, Posting new record, requesting full representation, can rename and cast the selected columns
       body mismatch:
         expected: [{"pId":"7","pName":"New Project","cId":"2"}]
         but got:  {"pId":"7","pName":"New Project","cId":"2"}

  test/Feature/InsertSpec.hs:258: 
  2) Feature.InsertSpec, CSV insert, requesting full representation, only returns the requested column header with its associated data
       body mismatch:
         expected: "id\n8\n9"
         but got:  "id,name,client_id\n8\n9"

  test/Feature/QuerySpec.hs:592: 
  3) Feature.QuerySpec, remote procedure call, improper input, responds with 400 on an unexisting proc
       status mismatch:
         expected: 400
         but got:  404

  test/Feature/QuerySpec.hs:615: 
  4) Feature.QuerySpec, remote procedure call, executes the proc exactly once per request
       body mismatch:
         expected: 1
         but got:  4

  test/Feature/UnicodeSpec.hs:15: 
  5) Feature.UnicodeSpec, Reading and writing to unicode schema and table names, Can read and write values
       body mismatch:
         expected: []
         but got:  [{"هویت":1}]
```

This PR is to replace #795 .


begriffs:
Wow this is... intense!

I wonder whether any other backward-incompatible changes are in here? We can't have that because semantic versioning would dictate a major version bump and obviously 0.4 is already taken. Only bug fixes and added features can be in this release which would be 0.3.3.0.

begriffs:
Is the motivation for this work that don't want to generate JWT in SQL (and share the JWT secret in the db)?

krakrjak:
Absolutely! Removing the JWT secret sharing to the DB is the motivation. I understand the desire and need to share the secret between the database and the web server, but my code currently depends on not sharing this secret with anyone.

Ultimately I am flexible and intend to forward port my application to the new way of sharing the secret, but it will take a while and I'd like to pick up a few fixes as I attempt to cross the rubicon.

begriffs:
@ruslantalpa, just a thought: what if we actually added back the ability for postgrest to encode JWT based on a special function return type? People who want to do the manual in-SQL way could still do that, but I've noticed a lot of people are finding it annoying to do the secret management when a very modest amount of postgrest code could help them.

begriffs:
@krakrjak I respect the work you put in here, but think it would be better to avoid this extensive work on the 0.3.x series. Feel free to make an issue with suggestions about JWT handling in future versions and we can discuss the merits.

PR #857, opened 2017-04-06
Fix swagger-ui when "server-proxy-uri" is not set

feynmanliang:
When `server-proxy-uri` is not set in `postgrest.conf`, the `swagger.json` would report a `Base url: 0.0.0.0:3000/`, which results in the generated API calls for the examples to have an extra slash i.e. the "//" in:

```
curl -X GET http://0.0.0.0:3000//skills -H  "accept: application/json" -H  "content-type: application/json" -H  "Range: undefined" -H  "Range-Unit: items" -H  "Prefer: undefined"
```

This results in a failure when trying out the API, see:

This PR removes this extraneous tailing "/" for the case where `server-proxy-uri` is not set.

---
This change is [Reviewable](https://reviewable.io/reviews/begriffs/postgrest/857)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Resource embedding in views referencing tables in public schema - @fab1an
 - #777, Empty body is allowed when calling a non-parameterized RPC - @koulakis
 - #831, Fix proc resource embedding issue with search_path - @steve-chavez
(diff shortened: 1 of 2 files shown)


begriffs:
Just reproduced locally, thanks for the fix! Want to include a line in the changelog too please?

feynmanliang:
:+1:

begriffs:
Actually I reverted this PR because it was actually a bug with swagger-ui, not postgrest. https://github.com/swagger-api/swagger-ui/issues/819

Issue #790, opened 2017-01-21
OPTIONS no longer works in v0.4.0.0


RB14:
In the new version released yesterday, it doesn't seem possible to get the schema definition for a particular table/view using OPTIONS request.

It is possible to get the schema for the entire db by issuing a GET request on the root node, but that seems like an overkill, and too verbose.

It is still noted in the docs though, that OPTIONS is a valid verb.

Also, without OPTIONS, it doesn't seem possible to get information about foreign key constraints (GET request on the root node does not provide this info).

\* Using Windows binary

begriffs:
Yeah you're right, OPTIONS no longer provides API documentation about a route. However it is still a supported verb and responds with CORS and Allow headers.

The API information has been moved to the OpenAPI description available at the root path. Foreign key information is not yet included in the OpenAPI output though. That will have to be a later feature.

Are you working on a front-end admin panel or other generic code that needs to learn about the foreign keys between tables?

RB14:
Yes, I was working on some sort of admin panel, but eventually I decided to just hard-code the relationships, as it is more appropriate for my use case.

I still think it would be better to be able to consult each node for its relationships rather than having to fetch the entire schema from the root path (which can get pretty huge).

majorcode:
As someone who is not fully familiar with the OpenAPI spec, may I ask: Is the absence of information about the foreign key or embedded resources due to limitations in the OpenAPI specification? Or is it something you have not had an opportunity to implement?

In addition to UIs that auto-discover which resources can be embedded, I think this information is really useful to someone learning to use PostgREST. First, appearance of foreign key / embedded resource data during development confirms that the relationship is correct, and that PostgREST picked up on the change (after a HUP). Second, it's obvious that showing the structure makes it easier to explore and use PostgREST.

@begriffs In my opinion, I really liked your use of OPTIONS to discover the API. However, I can understand the switch to OpenAPI. I'd love to help reinstate this functionality. Alas, my Haskell foo is too weak. Would you be open to a Haskell noob attempting to assist?

begriffs:
To be honest I don't know the best way to structure this information in OpenAPI. I have a feeling there's a way to do it though. Are you interested in researching this and providing your recommendation?

I agree that it's useful functionality, for instance for building admin panels like in #717.

(I'm also happy to help you with any Haskell questions as you begin learning. A good starting place would be to try building the project from source, explained here: https://postgrest.com/en/v0.4/install.html#build-from-source)

[...]

PR #864, opened 2017-04-19
Add and/or params for doing complex boolean logic, Fix #652

### LINKED ISSUE #652: Why aren't "OR" conditions currently supported?
It seems like a non-trivial limitation on queries, but perhaps there is a good reason for it. Out of curiosity, why aren't OR conditions supported currently?

From http://postgrest.com/api/reading/#embedding-foreign-entities 

> For more complicated filters (such as those involving condition 1 OR condition 2) you will have to create a new view in the database.

steve-chavez:
The ```and/or``` params are working as discussed in #652 though this is a work in progress(only working for GET for now), I wanted to show how this currently works and propose some solutions to some conflicts I've bumped into, so right now:

```http
GET /people?not.or={student.is.true,not.and{age.gte.18,height.lte.180}}
```
Gets translated in sql to:
```sql
WHERE NOT ( student IS TRUE OR NOT ( age => 18 AND height <= 180))
```

I've added the optional ```not``` to the and/or, also since in #652 no one objected to the idea of replacing ```=``` with ```.``` I went ahead and implemented it like that to avoid the need to urlencode.

Also added the ability to combine ```and``` with ```or``` params, in addition to the ability to combine those with traditional filters:  

```http
GET /people?or={student.is.true,age.lte.14}&and={height.gt.180,weight.lt.100}&driver=is.true
```

They get concatenated with AND in sql as you can see:

```sql
WHERE ( student IS TRUE OR age <= 14 ) AND ( height > 180 AND weight < 100 ) AND driver IS TRUE
```

```and/or``` also works for embedded levels, you can see that on the [tests](https://github.com/steve-chavez/postgrest/blob/and-or-virtual-cols/test/Feature/QuerySpec.hs#L825-L834).

## Conflicts

Now there are three operators that don't work with the ```and/or``` params: 

#### IN operator:

```http
GET /people?or={id.in.1,2,3,age.eq.20}
```

There's no clear separation with ```,``` maybe the IN value could be forced to have quotes when inside             ```and/or``` but the quotes are also used for values that have comma and I don't know if this: ```
in."x,"x,y","w,z""``` could be parsed.

How about changing the separator inside ```and/or``` to another url safe character like ```+```, this could look like:

```http
GET /people?or={id.in.1,2,3+and{height.lte.180+age.eq.20}}
```

#### @> and <@ operators:

```http
GET /complex_items?or={id.eq.1,arr_data.@>.{1}}
GET /people?or={id.eq.1,arr_data.<@.{1,2,3}}
```

For these two there's a conflict between PostgreSQL array literal ```{1,2,3}``` and the ```and/or``` block delimiter ```}```.  

How about replacing the ```{}``` block delimiters to parentheses ```()```? This would avoid that conflict, I know that this replacement was planned for later to avoid urlencoding, but maybe this would be a good time to introduce them just in the ```and/or``` params.

The rest of the operators are working and I'll add tests for those, but before proceeding I would like to solve those conflicts. Thoughts?

### DIFF
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -180,7 +180,7 @@ addFiltersOrdersRanges apiRequest = foldr1 (liftA2 (.)) [
   of type (ReadRequest->ReadRequest) that are in (Either ParseError a) context
   -}
   where
-    filters :: Either ApiRequestError [(Path, Filter)]
+    filters :: Either ApiRequestError [(EmbedPath, Filter)]
     filters = mapM pRequestFilter flts
       where
         action = iAction apiRequest
@@ -188,30 +188,30 @@ addFiltersOrdersRanges apiRequest = foldr1 (liftA2 (.)) [
           | action == ActionRead = iFilters apiRequest
           | action == ActionInvoke = iFilters apiRequest
           | otherwise = filter (( "." `isInfixOf` ) . fst) $ iFilters apiRequest -- there can be no filters on the root table whre we are doing insert/update
-    orders :: Either ApiRequestError [(Path, [OrderTerm])]
+    orders :: Either ApiRequestError [(EmbedPath, [OrderTerm])]
     orders = mapM pRequestOrder $ iOrder apiRequest
-    ranges :: Either ApiRequestError [(Path, NonnegRange)]
+    ranges :: Either ApiRequestError [(EmbedPath, NonnegRange)]
     ranges = mapM pRequestRange $ M.toList $ iRange apiRequest
 
 addFilterToNode :: Filter -> ReadRequest -> ReadRequest
 addFilterToNode flt (Node (q@Select {flt_=flts}, i) f) = Node (q {flt_=flt:flts}, i) f
 
-addFilter :: (Path, Filter) -> ReadRequest -> ReadRequest
+addFilter :: (EmbedPath, Filter) -> ReadRequest -> ReadRequest
 addFilter = addProperty addFilterToNode
 
 addOrderToNode :: [OrderTerm] -> ReadRequest -> ReadRequest
 addOrderToNode o (Node (q,i) f) = Node (q{order=Just o}, i) f
 
-addOrder :: (Path, [OrderTerm]) -> ReadRequest -> ReadRequest
+addOrder :: (EmbedPath, [OrderTerm]) -> ReadRequest -> ReadRequest
 addOrder = addProperty addOrderToNode
 
 addRangeToNode :: NonnegRange -> ReadRequest -> ReadRequest
 addRangeToNode r (Node (q,i) f) = Node (q{range_=r}, i) f
 
-addRange :: (Path, NonnegRange) -> ReadRequest -> ReadRequest
+addRange :: (EmbedPath, NonnegRange) -> ReadRequest -> ReadRequest
 addRange = addProperty addRangeToNode
 
-addProperty :: (a -> ReadRequest -> ReadRequest) -> (Path, a) -> ReadRequest -> ReadRequest
+addProperty :: (a -> ReadRequest -> ReadRequest) -> (EmbedPath, a) -> ReadRequest -> ReadRequest
 addProperty f ([], a) n = f a n
 addProperty f (path, a) (Node rn forest) =
   case targetNode of
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE DuplicateRecordFields    #-}
 module PostgREST.DbRequestBuilder (
   readRequest
 , mutateRequest
@@ -6,6 +7,7 @@ module PostgREST.DbRequestBuilder (
 ) where
 
 import           Control.Applicative
+import           Control.Arrow             ((***))
 import           Control.Lens.Getter       (view)
 import           Control.Lens.Tuple        (_1)
 import qualified Data.ByteString.Char8     as BS
@@ -173,7 +175,8 @@ addFiltersOrdersRanges :: ApiRequest -> Either ApiRequestError (ReadRequest -> R
 addFiltersOrdersRanges apiRequest = foldr1 (liftA2 (.)) [
     flip (foldr addFilter) <$> filters,
     flip (foldr addOrder) <$> orders,
-    flip (foldr addRange) <$> ranges
+    flip (foldr addRange) <$> ranges,
+    flip (foldr addLogicTree) <$> logicForest
   ]
   {-
   The esence of what is going on above is that we are composing tree functions
@@ -182,12 +185,13 @@ addFiltersOrdersRanges apiRequest = foldr1 (liftA2 (.)) [
   where
     filters :: Either ApiRequestError [(EmbedPath, Filter)]
     filters = mapM pRequestFilter flts
-      where
-        action = iAction apiRequest
-        flts
-          | action == ActionRead = iFilters apiRequest
-          | action == ActionInvoke = iFilters apiRequest
-          | otherwise = filter (( "." `isInfixOf` ) . fst) $ iFilters apiRequest -- there can be no filters on the root table whre we are doing insert/update
+    logicForest :: Either ApiRequestError [(EmbedPath, LogicTree)]
+    logicForest = mapM pRequestLogicTree logFrst
+    action = iAction apiRequest
+    -- there can be no filters on the root table when we are doing insert/update/delete
+    (flts, logFrst)
+      | action == ActionRead || action == ActionInvoke = (iFilters apiRequest, iLogic apiRequest)
+      | otherwise = join (***) (filter (( "." `isInfixOf` ) . fst)) (iFilters apiRequest, iLogic apiRequest)
     orders :: Either ApiRequestError [(EmbedPath, [OrderTerm])]
     orders = mapM pRequestOrder $ iOrder apiRequest
     ranges :: Either ApiRequestError [(EmbedPath, NonnegRange)]
@@ -210,6 +214,12 @@ addRangeToNode r (Node (q,i) f) = Node (q{range_=r}, i) f
 
 addRange :: (EmbedPath, NonnegRange) -> ReadRequest -> ReadRequest
 addRange = addProperty addRangeToNode
+
+addLogicTreeToNode :: LogicTree -> ReadRequest -> ReadRequest
+addLogicTreeToNode t (Node (q@Select{logic=l},i) f) = Node (q{logic=t:l}::ReadQuery, i) f
+
+addLogicTree :: (EmbedPath, LogicTree) -> ReadRequest -> ReadRequest
+addLogicTree = addProperty addLogicTreeToNode
 
 addProperty :: (a -> ReadRequest -> ReadRequest) -> (EmbedPath, a) -> ReadRequest -> ReadRequest
 addProperty f ([], a) n = f a n
@@ -248,8 +258,8 @@ mutateRequest :: ApiRequest -> [FieldName] -> Either Response MutateRequest
 mutateRequest apiRequest fldNames = mapLeft apiRequestError $
   case action of
     ActionCreate -> Right $ Insert rootTableName payload returnings
-    ActionUpdate -> Update rootTableName <$> pure payload <*> filters <*> pure returnings
-    ActionDelete -> Delete rootTableName <$> filters <*> pure returnings
+    ActionUpdate -> Update rootTableName <$> pure payload <*> filters <*> logic_ <*> pure returnings
+    ActionDelete -> Delete rootTableName <$> filters <*> logic_ <*> pure returnings
     _        -> Left UnsupportedVerb
   where
     action = iAction apiRequest
@@ -261,7 +271,11 @@ mutateRequest apiRequest fldNames = mapLeft apiRequestError $
         _ -> undefined
     returnings = if iPreferRepresentation apiRequest == None then [] else fldNames
     filters = map snd <$> mapM pRequestFilter mutateFilters
-      where mutateFilters = filter (not . ( "." `isInfixOf` ) . fst) $ iFilters apiRequest -- update/delete filters can be only on the root table
+    logic_ = map snd <$> mapM pRequestLogicTree logicFilters
+    -- update/delete filters can be only on the root table
+    mutateFilters = onlyRoot $ iFilters apiRequest
+    logicFilters = onlyRoot $ iLogic apiRequest
+    onlyRoot = filter (not . ( "." `isInfixOf` ) . fst)
 
 fieldNames :: ReadRequest -> [FieldName]
 fieldNames (Node (sel, _) forest) =
--- a/src/PostgREST/Parsers.hs
+++ b/src/PostgREST/Parsers.hs
@@ -78,8 +78,9 @@ pFieldForest :: Parser [Tree SelectItem]
 pFieldForest = pFieldTree `sepBy1` lexeme (char ',')
 
 pFieldTree :: Parser (Tree SelectItem)
-pFieldTree = try (Node <$> pSimpleSelect <*> between (char '{') (char '}') pFieldForest)
-          <|>     Node <$> pSelect <*> pure []
+pFieldTree =  try (Node <$> pSimpleSelect <*> between (char '{') (char '}') pFieldForest)
+          <|> try (Node <$> pSimpleSelect <*> between (char '(') (char ')') pFieldForest)
+          <|> Node <$> pSelect <*> pure []
 
 pStar :: Parser Text
 pStar = toS <$> (string "*" *> pure ("*"::ByteString))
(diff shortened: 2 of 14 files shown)


begriffs:
I'll review this tonight and also set up your permissions for the Waffle board.

begriffs:
First of all...wow, thanks for this big step forward! Now about your questions.

For `in`, what if we required the values to be surrounded with `{}` -- or with `()` if we're switching to that:

```http
GET /people?or=(id.in.(1,2,3),age.eq.20)
```

I believe we already support quoted `in` values which allows strings containing commas as possible values.

Same for `<@`?

```http
GET /people?or=(id.eq.1,arr_data.<@.(1,2,3))
```

Is it possible to parse this format and turn it into a SQL array?

ruslantalpa:
On the http format i'll let you two decide which one is best. I want to comment on the implementation.

The current path was probably chosen to implement the feature with as little changes as possible but i think it leads to complications down the line when this is extended to work everywhere and on all levels. It think it's allso best to 'bite the bullet' and implement it everywhere in all paths (read/write/sublevels), otherwise you would have to maintain duplicate functions that implement the same logic but work for different data types. 

The problem starts at the type level. These record fields `flt_::[Filter], logic::[LogicTree]` both hold the same type of data. They represent information that naturally is represented by a single tree, it's just that the `flt_` holds the first level of the tree that is later merged into the `logic` when creating the final query.

Side note, the name `flt_` was chosen because at the time GHC did not support the same property name for two different record types but now i think that is allowed, so it should really be named `where_`

The better data types i think are
```
data LogicOperator = And | Or deriving Eq
data Filter = Filter { field::Field, operation::Operation } deriving (Show, Eq)
data Condition = F Filter | O LogicOperator deriving (Show, Eq)
data ReadQuery = Select { select::[SelectItem], from::[TableName], where_::Tree Condition, order::Maybe [OrderTerm], range_::NonnegRange } deriving (Show, Eq)
``` 
The specific naming is not important (better names are possible), the point is show how i think this should be encoded in the types.

Some specifics:
all the capabilities of `[Filter]` (the old type) are covered by a tree where the root node is `O And` and all the elements in the forest are of type `F Filter`. In fact, the root node of `where_` is always going to be of type `O And` and then all the filtering conditions, be that normal columns or virtual are added as items of its forest.

steve-chavez:
@begriffs For the <@, @> if you already agreed on changing the block delimiters ```{}``` to parens ```()``` then the PostgreSQL array literal can stay as it is, so I can make this:

```http
GET /people?or=(id.eq.1,arr_data.<@.{1,2,3})
```

Work normally, though I would have to add another parser to handle the ```{1,2,3}``` since the comma ```,``` is still our separator, but that is doable, I agree and I'll change the parser block delimiters to parens.

The IN surrounded with parens ```id.in.(1,2,3)``` is also doable and solves the problem, but is inconsistent. While testing the ```and/or```params I've noticed that it's easy to make mistakes and I would prefer not to have this "gotcha". I'd like to know what's wrong with ```+``` because I think that's a more clear separator.

@ruslantalpa The implementation was already working for the sublevels and now I've added the feature for insert/update/delete, I didn't add duplicate functions and I've mostly reused some of the logic and functions already available so I don't see much complications.

I get your point on the types, both the Filter and LogicTree should be represented with a single type, that would definitely simplify the code, I would like to refactor to that but I think it'll be more appropriate to do that on another PR otherwise I'll extend this PR for too long.

ruslantalpa:
@steve-chavez you are probably right about having the first version of this implemented as it is now since it does not "touch" the old code and the new feature is kind of separate in the code.
The next release could add this feature as "experimental", ppl will try it out, comment on the interface and the release after that, will commit to the interface and also rewrite the internal types.

the `and/or` columns could start using `()` directly in this experimental version, this can be separate from the `{}` used in `select`. However, i think it should be possible for the `select` parser to understand both `{}` and `()` at the same time so that we have a nice transition period. Maybe this PR could implement this.

jmealo:
I'm impressed with how expressive the API has become. Is this expressiveness worth the complexity and maintenance challenges it presents?

@begriffs: Should their be REST and GraphQL endpoints that share common plumbing? I wouldn't expect a REST API to support these things... That doesn't mean they can't/shouldn't.

I'm interested in what happens when you separate introspection and query building primitives from transports (REST/GraphQL/websockets/webhooks/pubsub) and offer optional change data capture (real-time) and auth plugins. That way you can customize what you want and selectively expose resources over various transports with custom auth.

ruslantalpa:
I am building the graphql endpont on top :) (with websockets and such) i doubt someone is going to take on the challange of building it with an MIT license and have the quality of postgrest. You have to remember it took almost two years to have the interface of postgrest be capable of supporting graphql on top

begriffs:
@jmealo that's a thoughtful question. In this case I like the new interface. It's compatible with the old behavior (except the change from curly braces to parens). It's pretty simple to use, and people can use as much or little of it as they like. For instance you can choose use custom views instead of the rich query params. Also it seems to be expressive enough that you can make a graphql layer on top like Ruslan did, or even another query language.

As for introspection, I reopened #56 because of the conversation in #790

begriffs:
@steve-chavez I see what you mean about inconsistencies between operators in top-level query params and those inside and/or blocks. I agree now that we shouldn't use parens for the `in` values, but should use a `+` separator.

ruslantalpa:
And/or dont need to support {}, only (). This is a new feature. Nothing will break. Only select needs to support both

ruslantalpa:
as simple as that (the parser can probably be nicer, but why, it will be in this sate only for a few releases)
```haskell
pFieldTree :: Parser (Tree SelectItem)
pFieldTree = try (Node <$> pSimpleSelect <*> between (char '{') (char '}') pFieldForest)
          <|>     (Node <$> pSimpleSelect <*> between (char '(') (char ')') pFieldForest)
          <|>     Node <$> pSelect <*> pure []
```

begriffs:
True that only select really needs to preserve the old style for backward compatibility. Although for consistency the and/or ones should probably be equally permissive.

ruslantalpa:
the support for `{}` is a curtesy to legacy code, the correct way to write a requests (in the new version) is to use `()` everywhere. If someone is in the position that he is writing a query string and using `or/and` it means he should be using the new format everywhere. I don't see why a new feature should be consistent with a deprecated format especially since it introduces complications because of things like `in.{1,2,4}`.

ruslantalpa:
maybe i did not explain it completely. I am thinking of the next version in general, not only about this pr. I think we agreed that the new version will use `()` as delimiters, so from that perspective, users starting working with postgrest, will see no inconsistencies since they will use `()` everywhere (and the docs will say to use that).
People already using postgrest will be able to upgrade however, since select will also support `{}`.
Now, if old users, start adding new features to their applications and want to use `and/or`, having support only for `()` will push them towards moving to the new format everywhere.

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -173,45 +175,53 @@ addFiltersOrdersRanges :: ApiRequest -> Either ApiRequestError (ReadRequest -> R
...
   ]
   {-
   The esence of what is going on above is that we are composing tree functions
   of type (ReadRequest->ReadRequest) that are in (Either ParseError a) context
   -}
   where
-    filters :: Either ApiRequestError [(Path, Filter)]
+    filters :: Either ApiRequestError [(EmbedPath, Filter)]
     filters = mapM pRequestFilter flts
-      where
```
begriffs: Nice call getting rid of the nested where blocks.

src/PostgREST/QueryBuilder.hs
```diff
@@ -279,27 +286,31 @@ requestToQuery schema _ (DbMutate (Insert mainTbl (PayloadJSON rows) returnings)
...
+requestToQuery schema _ (DbMutate (Update mainTbl (PayloadJSON rows) conditions logic_ returnings)) =
   case rows V.!? 0 of
     Just obj ->
       let assignments = map
             (\(k,v) -> pgFmtIdent k <> "=" <> insertableValue v) $ HM.toList obj in
       unwords [
         "UPDATE ", fromQi qi,
         " SET " <> intercalate "," assignments <> " ",
-        ("WHERE " <> intercalate " AND " ( map (pgFmtFilter qi ) conditions )) `emptyOnNull` conditions,
+        if null conditions && null logic_
```
begriffs: The repeated `if null conditions && null logic_` makes me think we could have an `emptyOnFalse` function. It would handle this logic as well as replace the `emptyOnNull` that occurs in other places in the code (with slight modification).

ruslantalpa: could that be a separate pr, since it touches places unrelated to this specific feature?

src/PostgREST/Types.hs
```diff
@@ -137,79 +137,57 @@ data Proxy = Proxy {
...
-  NotIn -> "NOT IN"
-  IsNot -> "IS NOT"
-  Is -> "IS"
-  TSearch -> "@@"
-  Contains -> "@>"
-  Contained -> "<@"
-
+type Operator = Text
+operators :: M.HashMap Operator SqlFragment
+operators = M.fromList [
```
begriffs: Nice! The consolidated operators.

test/Feature/StructureSpec.hs
```diff
@@ -19,7 +19,8 @@ spec :: SpecWith Application
 spec = do
 
   describe "OpenAPI" $ do
-    it "root path returns a valid openapi spec" $
+    it "root path returns a valid openapi spec" $ do
```
begriffs: This can be rebased away now that the test works on master again.

steve-chavez:
@begriffs There was an issue with using ```+``` as a separator, the current [queryString](http://hackage.haskell.org/package/wai-3.2.1.1/docs/Network-Wai-Internal.html#v:queryString) function we use for getting the query params parses the ```+``` to space ``` ```, I've seen there's a bool option to turn this off in [urlEncode](http://hackage.haskell.org/package/http-types-0.9.1/docs/Network-HTTP-Types-URI.html#v:urlEncode) but I guess internally the ```queryString``` sets that to True.

Couldn't get a clean way to do that using [rawQueryString](http://hackage.haskell.org/package/wai-3.2.1.1/docs/Network-Wai-Internal.html#v:rawQueryString) so instead I decided to change the separator to ```.``` so the current url is looking like this:

```http
GET /entities?and=(arr.not.@>.{1,2,3}.and(id.not.eq.2.name.not.like.*3))
```

You can see more examples in the tests, the advantage of that is that now all operators are working. Before trying to make the current and/or compatible with ```{}``` I want to know if you are ok with this change.

begriffs:
OK final TODOs and then I'll merge

* Change the delimiter inside and/or to `+` as you suggested
* Use `()` to surround and/or clauses
* Allow `()` to surround select clauses as well, making `{}` supported too but deprecated
* Tests for any edge cases you can think of like `or.()`
* Changelog entry
* Rebase on master

src/PostgREST/Parsers.hs
```diff
@@ -171,7 +171,7 @@ pOrderTerm =
   <|> OrderTerm <$> pField <*> pure Nothing <*> pure Nothing
 
 pLogicTree :: Parser LogicTree
-pLogicTree = try (pNotLogicOp <*> (char '{' *> pLogicTree) <*> (char ',' *> pLogicTree <* char '}'))
+pLogicTree = try (pNotLogicOp <*> (char '(' *> pLogicTree) <*> (pDelimiter *> pLogicTree <* char ')'))
```
ruslantalpa: why the switch to `.` wasn't the discussion about `+`? although i don't see whats wrong with `,`
`id.eq.1.id.eq.2` is a very ugly format.
PS: the parser also needs to support spaces before/after the separator or maybe the space should be the separator like

`/entities?or=(id.eq.1 id.eq.2)&select=id,name` but if you do it like this, there is in inconsistency in regard to `select`

so again, i don't see why it's not implemented like this
`/entities?or=(id.eq.1, id.eq.2,id.eq.3)&select=id,name`

steve-chavez: See my comment on top https://github.com/begriffs/postgrest/pull/864#issuecomment-298828830

begriffs: Steve noted that the haskell url parse turns `+` into a space. And yeah I also find the dot separators to be hard to read. What about a semicolon?

```
/entities?or=(id.eq.1;id.eq.2;id.eq.3)&select=id,name
```

ruslantalpa:
@begriffs all your points are good, especially the decision to have `and/or` support only `()` format (while select having the backward compatibility)

the only comment i have it for `+` which is now changed to `.` (which makes it even worse).
The way i understand it is that the reason to not using `,` was because of the parsers complications. I would suggest we better use more complex/smarter parsers then have an ugly interface with `+` or `.` as a separator, which is distinct from how `select` is separating items.

If using `,` is that complicated, maybe we can unify `select` and `or/and` by using ` ` <- spaces as separators (while having select support `,` also)

so like `/table?select=a b c&or=(id.eq.1 id.eq.2 )` 
having said that, it might also be ok to have it like this
`/table?select=a,b,c&or=(id.eq.1 id.eq.2 )` since in SQL you use `,` in the select section and spaces in the where section

ruslantalpa:
@steve-chavez make sure in places where you use the separators ( be that `,` `+` `.`) that it supports whitespace before/after the separator (and have a test for that)

begriffs:
What about semicolon?

```
/table?select=a,b,c&or=(id.eq.1;id.eq.2)
```

steve-chavez:
@ruslantalpa I don't think that whitespace is a good separator since that would have to be urlencoded %20.

@begriffs I thought that ```;``` was a not safe character according to rfc. I've seen js encodeURI not encode the ```;``` but in python's urllib it does encode it to %3B, so I guess it could cause issues to some clients.

I agree that the current ```.``` is hard to read though, think it would be better to just go back to ```,``` and force the ```in.1,2,3``` be inside of parentheses ```in.(1,2,3)```.

ruslantalpa:
Had time to think this through while on the way to office. Here is a solution that works
The problem arises not in the `and/or` context but in the `in` operator context, because the format is very permissive (`a,b,c,1,2,3`)
- change the separator back to `,` in the `and/or` context
- change the behaviour of `in` operator to be like `col=in.[1,2,3,a,"some, val"]` (this requires only adding `[]` separators as an option)
- change the docs to specify `[]` as the only option for the `in` operator
- preserve the old format of `in` for backward compatibility (this is very similar to what will happen with `select`)
- if someone uses `in` inside `and/or`, they will have to use `[]` and the parsers parsers will consume the correct string

while in this compatibility mode, it will be possible for someone to use the old format for the `in` value and the parser will consume more then it should, but the user will get a failed request (in development) and he will correct the query (which by they way should never happen since the docs will list `[]` as the only correct way)

ruslantalpa:
@begriffs @steve-chavez do you see problems with the above solution? does it seem like it will work?

ruslantalpa:
cool on the `()` separator, but for supporting only `()` format inside or, ... don't go into the woods.
to have it like that it would meant that you would need to define separate `condition` parsers, one to support both formats, and one to support only `()` format but they would both be parsing the same thing only in different context. It's possible you would also have to touch other parsers.
i don't see a big problem in letting the `in condition` parser "eating" more input then it's suppose to, it will fail anyway.
If you can pull it off as you said with very minimal overhead (1-2 lines) then by all means, enforce the `()`, but don't do it if it overcomplicates the code, remember this is just a backward compatibility thing what will be removed at some point ( in condition without the `()` )

steve-chavez:
So I found an edge case, using ```,)``` inside and ```eq``` and other operators besides ```in``` was not allowed. I've used the same quoting mechanism as used in ```in``` to allow those kind of values, like in:

```http
GET /entities?or=(name.eq."(entity,4)",name.eq."(entity,5)")
```

Added tests for those, also whitespace between expressions is tested to work. Changed the separator to ```,``` and added the transition to parentheses in ```select``` and top level ```in``` as we agreed, also there are tests for that.

@begriffs I think this is ready for another review.

test/Feature/AndOrParamsSpec.hs
```diff
@@ -0,0 +1,121 @@
...
+          get "/grandchild_entities?or=(id.in.(\"1\",\"2\"),id.in.(\"3\",\"4\"))&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }, { "id": 2 }, { "id": 3 }, { "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
+
+      it "allows whitespace" $
+        get "/entities?and=( and ( id.in.( 1, 2, 3 ) , id.eq.3 ) , or ( id.eq.2 , id.eq.3 ) )&select=id" `shouldRespondWith`
+          [json|[{ "id": 3 }]|] { matchHeaders = [matchContentTypeJson] }
+
+    context "used with POST" $
+      it "includes related data with filters" $
+        request methodPost "/child_entities?entities.or=(id.eq.2,id.eq.3)&select=id,entities{id}"
```
begriffs: This is surprising functionality to filter the results of the newly inserted records, but I guess it is consistent with other requests and could be powerful in the right circumstances. I'll have to be sure to document this.

test/Feature/QuerySpec.hs
```diff
@@ -799,8 +799,31 @@ spec = do
         get "/items_with_different_col_types?time_data=in." `shouldRespondWith`
           [json| [] |] { matchHeaders = [matchContentTypeJson] }
 
+    it "returns an empty result for notin and not.in when no value is present" $ do
```
begriffs: Shouldn't "not.in nothing" actually match everything? I'm thinking of it as being the complement of the empty set. It's a weird situation though, glad there's a test for it, whatever we decide.

begriffs:
Nice stuff and a lot of good tests. How about adding tests for some malformed query params? Just to verify they are sending back 4xx error codes and meaningful messages. Stuff like `/foo?or=)(` etc.

After that and figuring out the not.in question I think it's good to merge!

steve-chavez:
@begriffs I've improved some parser error messages by reducing ```try``` scopes, added tests for those, however in one case I was not able to improve the message. When doing this:
```http
GET /clients?or=(xor(id.eq.1,id.eq.2),id.eq.2)
```
Ideally the error message should be something like:
```json
{ "details": "unexpected \"x\" expecting logic operator (and, or)" }
```
But it gives this:
```json
{ "details": "unexpected \"(\" expecting letter, digit, \"-\", \"->\", \"->>\" or delimiter (.)", }
```

A similar error message works for ```not.xor(..)``` in this [test](https://github.com/steve-chavez/postgrest/blob/and-or-virtual-cols/test/Feature/AndOrParamsSpec.hs#L168-L172), for now I'm not sure why this happens, could clear that out in a later PR if that's ok with you.

Also fixed the ```not.in``` issue you pointed, added a ```jsonb``` col test with ```and/or``` and more ```not``` tests.
 
The PR commits got inconsistent, I don't think it will be good to squash all of them into one, so I've squashed them into 3 commits, maybe it would be better to leave them like that in master to make the history clearer.

begriffs:
This is great. It's fine about the and/or error message, I'll just merge this and preserve the three commits for logical history like you suggest. Thanks again for this feature.

PR #869, opened 2017-04-30
Add connection retrying on startup and SIGHUP, Fix #742

### LINKED ISSUE #742: Postgrest does not handle temporary database failure well
If the database goes down temporarily, Postgrest (version 3.2.0 in my tests) begins responding to requests with `503 Service Unavailable`. However, once the database issue is resolved Postgres does not self-heal and continues serving 503 errors. This requires a manual restart of Postgrest, since in addition to not reconnecting it doesn't die either (which could trigger a restart by an external monitor.)

Postgrest should handle database connectivity issues more gracefully. Simply retrying the connection every X seconds can cause high bandwidth usage for long outages if X is small, but if X is large then Postgrest doesn't quickly come back online when the database comes back online. Exponential backoff (wait X seconds, then 2X, then 4X, then 8X, etc.) fares better but long outages won't get resolved quickly. If the database is down for an hour then in the worst case Postgrest might be down for N hours, where N is the exponential growth factor. A more robust option, "cyclic exponential backoff", looks like somewhat this:

1. Wait X seconds
2. Wait 2X seconds
3. Wait 4X seconds
4. Wait 8X seconds
...
5. Wait (2^N)X seconds, where N is the maximum amount of "lag time" you want between when the database comes back up and when Postgrest should come back up (ideally no more than a few minutes)
6. Wait X seconds
7. Wait 2X seconds
... repeat indefinitely

And so on. Essentially, the strategy is to _never give up_ and _never go down_, although clients can specify a timeout period they're willing to wait. Postgrest could attempt retries in this manner in the background, but automatically respond with 503 to any requests that take longer than five seconds to complete.

This affects how Postgrest starts up too. Postgrest would listen to requests immediately while trying to connect to the database in the background. Until the connection is successful, all requests are met with 503. This has a handful of nice benefits:

- No need for worrying about service startup order. It no longer matters whether you start Postgrest first or the database first.
- It's easy to add health checking. Just fire off a `GET /` request until you get a success code and you know Postgrest is successfully listening to requests and talking to the database. Docker 1.12+ has native support for health checks, giving you neat features like the ability to do rolling deploys of Postgrest where traffic isn't switched over to new Postgrest instances until they're healthy.
- Avenues for adding more complex request logic around failures and timing. The [HTTP "wait" preference](https://tools.ietf.org/html/rfc7240#section-4.3) gives a standard way for clients to ask a server to set a deadline on how long processing the request will take. If a client isn't interested in waiting more than a second or two for a response, Postgrest can eagerly close the connection and respond with 503 when the database isn't available. Conversely, if a client _is_ willing to wait then Postgrest can give them a longer request deadline, giving the client a greater chance of success in the face of intermittent network failures between Postgrest and the database.

steve-chavez:
This adds a capped exponential backoff(base is 1 second, cap is 32 seconds, arbitrarily picked, could be improved) for retrying the db connection in a background thread.

This thread is started at postgREST startup and leaves it rejecting all requests with a 503 with no parsing or further processing, it's in an ```unavailable``` state until connection succeeds; this thread can also be started on SIGHUP.

I considered that postgREST is in an ```available``` state when the dbStructure is successfully obtained, and that's what the ```connectionWorker``` works on, I'll explain why:

At first I had an ```available``` IORef for when the connection succeeded but got some issues when the dbStructure was not obtained, for example postgREST succeeded with requests like:

```http
GET /clients
GET /clients?select=id,name
```

But gived an error with requests like:

```http
GET /
GET /clients?select=id,name,projects{*}
```

For what I've seen the first requests escaped the sections of the code that operated on dbStructure, so if the dbStructure is not obtained then postgREST is in a broken state, so instead of having two IORefs to maintain I decided that the IORef dbstructure should indicate the ```available``` state.

I think this is useful as it is but I don't know an easy way to add tests. Thoughts?

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -45,6 +45,7 @@ import           PostgREST.DbRequestBuilder( readRequest
 import           PostgREST.Error           ( simpleError, pgError
                                            , apiRequestError
                                            , singularityError, binaryFieldError
+                                           , connectionLostError
                                            )
 import           PostgREST.RangeQuery      (allRange, rangeOffset)
 import           PostgREST.Middleware
@@ -62,7 +63,7 @@ import           Data.Function (id)
 import           Protolude              hiding (intercalate, Proxy)
 import           Safe                   (headMay)
 
-postgrest :: AppConfig -> IORef DbStructure -> P.Pool -> IO POSIXTime ->
+postgrest :: AppConfig -> IORef (Maybe DbStructure) -> P.Pool -> IO POSIXTime ->
              Application
 postgrest conf refDbStructure pool getTime =
   let middle = (if configQuiet conf then id else logStdout) . defaultMiddle in
@@ -70,20 +71,22 @@ postgrest conf refDbStructure pool getTime =
   middle $ \ req respond -> do
     time <- getTime
     body <- strictRequestBody req
-    dbStructure <- readIORef refDbStructure
-
-    response <- case userApiRequest (configSchema conf) req body of
-      Left err -> return $ apiRequestError err
-      Right apiRequest -> do
-        let jwtSecret = binarySecret <$> configJwtSecret conf
-            eClaims = jwtClaims jwtSecret (iJWT apiRequest) time
-            authed = containsRole eClaims
-            handleReq = runWithClaims conf eClaims (app dbStructure conf) apiRequest
-            txMode = transactionMode dbStructure
-              (iTarget apiRequest) (iAction apiRequest)
-        response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
-        return $ either (pgError authed) identity response
-    respond response
+    maybeDbStructure <- readIORef refDbStructure
+    case maybeDbStructure of
+      Nothing -> respond connectionLostError
+      Just dbStructure -> do
+        response <- case userApiRequest (configSchema conf) req body of
+          Left err -> return $ apiRequestError err
+          Right apiRequest -> do
+            let jwtSecret = binarySecret <$> configJwtSecret conf
+                eClaims = jwtClaims jwtSecret (iJWT apiRequest) time
+                authed = containsRole eClaims
+                handleReq = runWithClaims conf eClaims (app dbStructure conf) apiRequest
+                txMode = transactionMode dbStructure
+                  (iTarget apiRequest) (iAction apiRequest)
+            response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
+            return $ either (pgError authed) identity response
+        respond response
 
 transactionMode :: DbStructure -> Target -> Action -> H.Mode
 transactionMode structure target action =
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -64,8 +64,8 @@ import           Protolude              hiding (intercalate, Proxy)
 import           Safe                   (headMay)
 
 postgrest :: AppConfig -> IORef (Maybe DbStructure) -> P.Pool -> IO POSIXTime ->
-             Application
-postgrest conf refDbStructure pool getTime =
+             IO () -> Application
+postgrest conf refDbStructure pool getTime worker =
   let middle = (if configQuiet conf then id else logStdout) . defaultMiddle in
 
   middle $ \ req respond -> do
@@ -86,7 +86,14 @@ postgrest conf refDbStructure pool getTime =
                   (iTarget apiRequest) (iAction apiRequest)
             response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
             return $ either (pgError authed) identity response
-        respond response
+        if isResponse503 response then do
+          worker
+          respond response
+        else
+          respond response
+
+isResponse503 :: Response -> Bool
+isResponse503 resp = statusCode (responseStatus resp) == 503
 
 transactionMode :: DbStructure -> Target -> Action -> H.Mode
 transactionMode structure target action =
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -86,11 +86,8 @@ postgrest conf refDbStructure pool getTime worker =
                   (iTarget apiRequest) (iAction apiRequest)
             response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
             return $ either (pgError authed) identity response
-        if isResponse503 response then do
-          worker
-          respond response
-        else
-          respond response
+        when (isResponse503 response) worker
+        respond response
 
 isResponse503 :: Response -> Bool
 isResponse503 resp = statusCode (responseStatus resp) == 503
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -8,6 +8,7 @@ module PostgREST.Error (
 , simpleError
 , singularityError
 , binaryFieldError
+, connectionLostError
 , encodeError
 ) where
 
@@ -72,6 +73,10 @@ binaryFieldError :: Response
 binaryFieldError =
   simpleError HT.status406 (toS (toMime CTOctetStream) <>
   " requested but a single column was not selected")
+
+connectionLostError :: Response
+connectionLostError =
+  simpleError HT.status503 "Database connection lost, retrying the connection."
 
 encodeError :: JSON.ToJSON a => a -> LByteString
 encodeError = JSON.encode
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -133,7 +133,7 @@ instance JSON.ToJSON H.Error where
     "details" .= (fmap toS d::Maybe Text)]
 
 httpStatus :: Bool -> P.UsageError -> HT.Status
-httpStatus _ (P.ConnectionError _) = HT.status500
+httpStatus _ (P.ConnectionError _) = HT.status503
 httpStatus authed (P.SessionError (H.ResultError (H.ServerError c _ _ _))) =
   case toS c of
     '0':'8':_ -> HT.status503 -- pg connection err
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #742, Add connection retrying on startup and SIGHUP - @steve-chavez
+
 ### Fixed
 
 ## [0.4.1.0] - 2017-04-25
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -153,6 +153,7 @@ main = do
     defaultUpdateSettings { updateAction = getPOSIXTime }
 
   runSettings appSettings $ postgrest conf refDbStructure pool getTime
+    (connectionWorker mainTid pool (configSchema conf) refDbStructure refIsWorkerOn)
 
 loadSecretFile :: AppConfig -> IO AppConfig
 loadSecretFile conf = extractAndTransform mSecret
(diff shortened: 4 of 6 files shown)


ruslantalpa:
this gets us to the "set it and forget it" state :)

steve-chavez:
Btw, to manually test this what I do is to set the pool timeout to a higher value, like 3600, here https://github.com/begriffs/postgrest/blob/master/main/Main.hs#L68, otherwise postgREST reconnects fast on its own, after that you can just stop or restart your PostgreSQL instance and check the reconnection process output, whether at startup or at SIGHUP.

begriffs:
This is helpful to make the server more robust, and to deal with docker-compose races too. Very nice.

I'll review the specifics of the code soon, and will think about how to test as well...

jackfirth:
Does this also apply when PostgREST is running normally but the database temporarily goes down?

steve-chavez:
For now just at startup and at SIGHUP, but if it works well as it is I could extend the connection retrying to work automatically whenever an user request gets a 503 response.

ajnsit:
@steve-chavez Wouldn't it make more sense to have some sort of a heartbeat process instead of waiting for a user request to fail?

jackfirth:
I consider recovering from temporary database connection failures necessary for closing #742. I could open a separate issue for that however.

steve-chavez:
@ajnsit If the db is down the reconnection is not guaranteed to be fast, could take a long time depending on the recovery of the db, so the user request would still fail, having a heartbeat process would not avoid that.

Also I think that the hearbeat would be more wasteful since to do that I'll have to do a background thread that sends something like ```SELECT 1``` in intervals and waits for failure all the time, better to just run a thread eventually.

@jackfirth I'll see if I can add the automatic connection retrying in this PR.

src/PostgREST/App.hs
```diff
@@ -62,28 +63,30 @@ import           Data.Function (id)
...
-            eClaims = jwtClaims jwtSecret (iJWT apiRequest) time
-            authed = containsRole eClaims
-            handleReq = runWithClaims conf eClaims (app dbStructure conf) apiRequest
-            txMode = transactionMode dbStructure
-              (iTarget apiRequest) (iAction apiRequest)
-        response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
-        return $ either (pgError authed) identity response
-    respond response
+    maybeDbStructure <- readIORef refDbStructure
+    case maybeDbStructure of
```
diogob: This is where I see that having an `Either ApiRequestError DbStructure` type would be useful, making composition easier

src/PostgREST/Error.hs
```diff
@@ -73,6 +74,10 @@ binaryFieldError =
   simpleError HT.status406 (toS (toMime CTOctetStream) <>
   " requested but a single column was not selected")
 
+connectionLostError :: Response
```
diogob: Using the either type we could also remove this function and use only `apiRequestError`

begriffs:
Note it's getting too late for me to review this one tonight, but I'll get to it tomorrow night.

begriffs:
Tried this locally and it works nicely when I kill the db and send postgrest a SIGHUP. During the interim it sends back 503, and reconnects when the db is back.

I agree with @jackfirth that detecting disconnection on a failed web request is important. When I kill the db and issue a request (no SIGHUP) postgrest currently sends back a lackluster error:

HTTP 500
```json
{
  "details":"FATAL:  role \"postgrest_test_authenticator\" does not exist\n",
  "code":"",
  "message":"Database connection error"
}
```

The message of the error is good, but the details are misleading and HTTP 503 would be preferable to indicate the temporary nature of the problem.

Having trouble coming up with tests for the test suite. We'd need a way to mock the database or something and make it die while expecting reconnection attempts. Sounds hard and un-haskelly. I guess I'm good if we smoke test it and it works.

begriffs:
I'll wait to look at the code in greater depth until after you've worked on the non-sighup disconnection detection in case that work causes reorganization of the code.

ruslantalpa:
@begriffs the code in this pr is already useful in a lot of cases, especially docker. Let's have it reviewed and merged and leave the "mid production disconnect" for another PR. It already took a long time having this implemented.

begriffs:
@ruslantalpa I tentatively disagree.

The automatic reconnection detection seems like the primary feature actually. Sighup handling is a special case that would be unnecessary if postgrest could auto reconnect on lost connection. (Well maybe it would still be useful to trigger an immediate retry if postgrest had reached a very large wait value for exponential backoff...) I'm imagining the reconnection attempt on startup to be a special case of reconnecting *whenever* the connection drops.

Merging this feature in the current state might be dangerous in that it's sorta kinda good enough for some cases and then we'd be less likely to revisit it. If we add features to postgrest I want us to do it right.

Thoughts?

ruslantalpa:
The initial connect failure is the important feature. Once the db structure is introspected, the db can go down and PostgREST will recover because db connections are released from the pool after 1s (if not used). Without this feature, if the db is not present at startup, PostgREST never recovers, so you need to revert to tricks to control the boot order.
Also, this PR introduces the primitives (which will not need changing later) that will be used to implement the mid process disconnect.
The two features while in the same category are not dependent on one another, they can be added separately, in stages. I can understand your reasoning if somehow this code would later need a rewrite, but i don't think that is the case, the mid disconnect case was considered from the very beginning and the primitives were created in such a way to later support that feature (hence the global ioref and the 503). 
Lastly, not merging a feature because it will be "less likely to revisit" (a second related feature)  does not increase chances of the second feature being implemented (but rather decrease them :) )

jackfirth:
@ruslantalpa I'm not sure I understand what you mean, so I'd like to clarify what I think is necessary for closing #742. Specifically, I think this flow should be possible:

- User sends request to PostgREST, receives normal response.
- Database goes down, all attempts to connect fail.
- User sends request to PostgREST, receives error response.
- Database comes back up, new connections can now be opened.
- User sends request to PostgREST, receives normal response.

This PR as far as I know only addresses how PostgREST behaves at startup. Are you saying that due to the way PostgREST pools connections, this currently happens already? Or are you saying that PostgREST currently does not handle this situation, but this PR somehow interacts with connection pooling to address it?

And when you say "mid process disconnect", do you mean when the database temporarily fails *while* PostgREST is processing a request? So retrying a temporary failure there would prevent the user from seeing a failure at the cost of greater latency?

ruslantalpa:
Yes, because of how pool connections work, this happens already, provided more then 1s passes between http requests so that the db connections in the pool are closed and the new ones will work.
This does not happen if there is a constant stream of requests (more then 1 per second)

ruslantalpa:
by mid process i mean the situation you describe, not mid request

jackfirth:
So in the flow I described, if PostgREST gets more than one query per second in total traffic continuously it will not recover from a temporary failure? That seems like a pretty big deal, I would not call #742 fixed in that case.

ruslantalpa:
it's a big deal, and no one is calling #742 fixed

jackfirth:
Okay good, thanks for the clarification.

steve-chavez:
@begriffs I've added the automatic connection retrying, this happens whenever an user gets a 503. Also corrected that error code status you mentioned to 503. Now it needs review and manual testing.

begriffs:
@steve-chavez you are such a rockstar! 🤘 I'll give it a look this evening and try hitting it with a bunch of concurrent requests/sighups while bringing the db up and down just to check for any concurrency issues.

begriffs:
I was able to find odd behavior. To reproduce, create these scripts:

sighup.sh:
```bash
while [ 1 ]
do
    killall -HUP postgrest
    sleep 0.23
done
```

getrequest.sh:
```bash
while [ 1 ]
do
    curl -I http://localhost:3000 -s | head -n 1 | cut -d$' ' -f2
    sleep 0.1
done
```

1. Start the db and then start postgrest
2. Run getrequest.sh, which should give a bunch of 200s
3. In another terminal run sighup.sh
4. Getrequest will show alternating 200 and 503. Maybe this is because it takes time to reload the db schema and during the reload it temporarily wipes out the structure cache?

steve-chavez:
Yes, it's because of the time it takes to reload the dbStructure, but that was deliberate, the idea was to put PostgREST in an 'unavailable' state and respond quickly with 503, avoiding parsing etc., while retrying the connection.

steve-chavez:
Accidentally deleted my previous comment with Vimium. Yes, it's because the time it takes to obtain the ```dbStructure```, but that was deliberate, the idea was to put PostgREST in an unavailable state and respond with 503 quickly, without parsing ,etc., while the connection retrying is happening.
Do you think that his behavior should be improved?. I think is good enough, even if getting the dbStructure is avoided and we just retry the connection it would still give 503 to some concurrent requests.

begriffs:
Makes sense. Repeatedly sending sighup is not even a realistic thing, and a brief 503 should be fine. Just wanted to double check it with you.

src/PostgREST/App.hs
```diff
@@ -62,28 +63,37 @@ import           Data.Function (id)
...
+          Right apiRequest -> do
+            let jwtSecret = binarySecret <$> configJwtSecret conf
+                eClaims = jwtClaims jwtSecret (iJWT apiRequest) time
+                authed = containsRole eClaims
+                handleReq = runWithClaims conf eClaims (app dbStructure conf) apiRequest
+                txMode = transactionMode dbStructure
+                  (iTarget apiRequest) (iAction apiRequest)
+            response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
+            return $ either (pgError authed) identity response
+        if isResponse503 response then do
```
begriffs: Might be a good use for the [when](https://www.stackage.org/haddock/lts-8.13/protolude-0.1.10/Monad.html#v:when) function:

```haskell
when (isResponse503 response) worker
respond respose
```

src/PostgREST/Error.hs
steve-chavez: A special message for the reconnection state is needed. I tried to make your suggestion work by adding a new constructor for ```ApiRequestError``` with this message, all good, the problem was that by making the ```IORef (Either ApiRequestError DbStructure)``` I had to change two function type signatures, namely ```app``` and ```transactionMode```, to avoid that I could have unwrapped the Left of the IORef, but that would be the same non composable way as of the Maybe. In the end I was not able to make the code simpler with your suggestion.

begriffs:
It's looking good, I verified that
- concurrent sighup and requests work when taking db down and up
- everything still shuts down properly if isServerVersionSupported returns false
- it retries for connection on startup if needed and gets it successfully

My only question about functionality at this point is about the retry timing and sighup. Let's say postgrest has tried reconnecting a number of times and the next try won't be for quite a while. It would be nice if sighup could reset the backoff and make it check right away. That way if I notice the db is down I can bring it up and get postgrest reconnected fast.

Things are working great right now and that sighup improvement is totally optional. If you think it's a pain to make that change I can just merge this as it is, let me know.

steve-chavez:
Right now the code is so that only a connectionWorker thread can run at a time no matter how many SIGHUP's are made, this also works when doing SIGHUP's and the connectionWorker is already on because of a failed request with 503; adding that timer reset and handling it concurrently would make this feature more complex.

The retrying is done in exponential backoff and when it hits 32 seconds, it doesn't keep increasing and the next retrying is 32s, next 32s, and so on(capped at 32s), so if the idea is not making the user wait for too long one option could be to reduce this cap, but for now I don't know what number will be more appropriate.

Other option would be, and this was the original intention, to do a cyclic exp backoff so if it hits 32s it restarts again from 1s, that would make the wait shorter, but I couldn't get that to work with the ```retry``` lib, see https://github.com/Soostone/retry/issues/47, @begriffs maybe you could check that I may have complicated myself and turns out its a one liner and then I could add that to this PR, if it's not simple then I think it'll be better if I give it more time and implement this option in another PR.

jackfirth:
As a frequent PostgREST and constant Docker user, I'm really glad to see this implemented. Thanks a ton @steve-chavez!

begriffs:
I overlooked that the wait time maxes out at 32 seconds. That's not so bad.

I think it's merge time! 🚀

PR #872, opened 2017-05-09
Postgrest Websockets [WIP]

diogob:
This is a first draft just to start discussing the approach.

Things to be done before this PR is ready to merge:
- [x] Expose a nicer interface from `postgrest-ws` to make the middleware setup in one line.
- [x] Integrate the ws middleware connection with the connection worker (or have a separate connection worker and abstract commonalities between both cases).
- [x] Do some tests to understand the memory consumption behaviour as the number of websockets clients increase.
- [ ] Add documentation.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -13,15 +13,20 @@ import           PostgREST.Error                      (encodeError)
 import           PostgREST.OpenAPI                    (isMalformedProxyUri)
 import           PostgREST.DbStructure
 import           PostgREST.Types                      (DbStructure, Schema)
+import           PostgRESTWS
+import           PostgRESTWS.Broadcast
+import           PostgRESTWS.HasqlBroadcast
 
 import           Control.AutoUpdate
 import           Control.Retry
 import           Data.ByteString.Base64               (decode)
+import           Data.Function                        (id)
 import           Data.String                          (IsString (..))
 import           Data.Text                            (stripPrefix, pack, replace)
 import           Data.Text.Encoding                   (encodeUtf8, decodeUtf8)
 import           Data.Text.IO                         (hPutStrLn, readFile)
 import           Data.Time.Clock.POSIX                (getPOSIXTime)
+import qualified Hasql.Connection                     as H
 import qualified Hasql.Query                          as H
 import qualified Hasql.Session                        as H
 import qualified Hasql.Decoders                       as HD
@@ -127,6 +132,9 @@ main = do
 
   pool <- P.acquire (configPool conf, 10, pgSettings)
 
+  conOrError <- H.acquire pgSettings
+  let con = either (panic . show) id conOrError :: H.Connection
+
   refDbStructure <- newIORef Nothing
 
   -- Helper ref to make sure just one connectionWorker can run at a time
@@ -152,8 +160,13 @@ main = do
   getTime <- mkAutoUpdate
     defaultUpdateSettings { updateAction = getPOSIXTime }
 
-  runSettings appSettings $ postgrest conf refDbStructure pool getTime
-    (connectionWorker mainTid pool (configSchema conf) refDbStructure refIsWorkerOn)
+  multi <- newHasqlBroadcaster con
+  void $ relayMessagesForever multi
+
+  runSettings appSettings $
+    postgrestWsMiddleware (configJwtSecret conf) getTime pool multi $
+    postgrest conf refDbStructure pool getTime
+      (connectionWorker mainTid pool (configSchema conf) refDbStructure refIsWorkerOn)
 
 loadSecretFile :: AppConfig -> IO AppConfig
 loadSecretFile conf = extractAndTransform mSecret
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -158,9 +158,8 @@ main = do
   wsMiddleware <-
         if configWebsockets conf
           then do
-            multiOrError <- newHasqlBroadcasterOrError pgSettings
-            let multi = either (panic . show) id multiOrError
-                secret = fromMaybe (panic "You need to set a JWT secret to enable websockets") $ configJwtSecret conf
+            multi <- newHasqlBroadcaster pgSettings
+            let secret = fromMaybe (panic "You need to set a JWT secret to enable websockets") $ configJwtSecret conf
             return $ postgrestWsMiddleware secret getTime pool multi
           else return id
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -14,7 +14,6 @@ import           PostgREST.OpenAPI                    (isMalformedProxyUri)
 import           PostgREST.DbStructure
 import           PostgREST.Types                      (DbStructure, Schema)
 import           PostgRESTWS
-import           PostgRESTWS.Broadcast
 import           PostgRESTWS.HasqlBroadcast
 
 import           Control.AutoUpdate
@@ -26,7 +25,6 @@ import           Data.Text                            (stripPrefix, pack, replac
 import           Data.Text.Encoding                   (encodeUtf8, decodeUtf8)
 import           Data.Text.IO                         (hPutStrLn, readFile)
 import           Data.Time.Clock.POSIX                (getPOSIXTime)
-import qualified Hasql.Connection                     as H
 import qualified Hasql.Query                          as H
 import qualified Hasql.Session                        as H
 import qualified Hasql.Decoders                       as HD
@@ -161,11 +159,10 @@ main = do
   wsMiddleware <-
         if configWebsockets conf
           then do
-            conOrError <- H.acquire pgSettings
-            let con = either (panic . show) id conOrError :: H.Connection
-            multi <- newHasqlBroadcaster con
-            void $ relayMessagesForever multi
-            return $ postgrestWsMiddleware (configJwtSecret conf) getTime pool multi
+            multiOrError <- newHasqlBroadcasterOrError pgSettings
+            let multi = either (panic . show) id multiOrError
+                secret = fromMaybe (panic "You need to set a JWT secret to enable websockets") $ configJwtSecret conf
+            return $ postgrestWsMiddleware secret getTime pool multi
           else return id
 
   runSettings appSettings $
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -132,9 +132,6 @@ main = do
 
   pool <- P.acquire (configPool conf, 10, pgSettings)
 
-  conOrError <- H.acquire pgSettings
-  let con = either (panic . show) id conOrError :: H.Connection
-
   refDbStructure <- newIORef Nothing
 
   -- Helper ref to make sure just one connectionWorker can run at a time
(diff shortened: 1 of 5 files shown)


diogob:
@begriffs @ruslantalpa as the connection pool is shared, I only need to re-estabilish the listener connection. I am planning to implement some sort of self-healing capability in the postgrest-ws library, so the PR code wouldn't change much. 

This branch already works, but the websockets functionality is lost if postgresql is not available during startup or if the connection is broken during operation.

I'd appreciate any early feedback.

begriffs:
The code seems fine, and is pretty short. I think your checklist is exactly right. Curious if the self-healing can leverage Steve's recent reconnection code.

diogob:
@begriffs after much consideration I gave up on reusing the same code. Since postgrest-ws uses a `Hasql.Connection` instead of a pool, and already hides this connection as an implementation detail, it made more sense to add a retry functionality to the function that creates the multiplexer.

For more details check this PR: https://github.com/diogob/postgrest-ws/pull/15

diogob:
The first set of benchmarks is ready. They are really simple, feel free to suggest aditional scenarios.
In this first round I just wanted to establish if the presence of the additional middleware would have any impact on standard HTTP requests. So this benchmarks are mostly testing the WAI plumbing:

All tests are run against the PostgREST test database using [artillery](https://artillery.io)
The http tests were performed with the command:

    artillery quick --duration 60 --rate 10 -n 20 http://localhost:3000

## Scenarios:

* master at 2017-05-27 (commit 50de0536ded1a560e2f405fa52e8dc6e18c404cc)

    all scenarios completed
    Complete report @ 2017-05-27T15:57:26.241Z
      Scenarios launched:  600
      Scenarios completed: 600
      Requests completed:  12000
      RPS sent: 198.22
      Request latency:
        min: 3.7
        max: 112.7
        median: 7
        p95: 12.6
        p99: 16.3
      Scenario duration:
        min: 152.5
        max: 434.7
        median: 171.2
        p95: 216.7
        p99: 341.2
      Scenario counts:
        0: 600 (100%)
      Codes:
        200: 12000

* postgrest-ws branch with websockets enabled in the config

    all scenarios completed
    Complete report @ 2017-05-27T15:40:32.491Z
      Scenarios launched:  600
      Scenarios completed: 600
      Requests completed:  12000
      RPS sent: 198.25
      Request latency:
        min: 3.7
        max: 423.8
        median: 7.3
        p95: 19.8
        p99: 50.3
      Scenario duration:
        min: 110.4
        max: 1439.1
        median: 173.3
        p95: 466.2
        p99: 1105.5
      Scenario counts:
        0: 600 (100%)
      Codes:
        200: 12000

* postgrest-ws branch with websockets disabled in the config

    all scenarios completed
    Complete report @ 2017-05-27T15:46:59.921Z
      Scenarios launched:  600
      Scenarios completed: 600
      Requests completed:  12000
      RPS sent: 198.25
      Request latency:
        min: 3.7
        max: 283.2
        median: 7
        p95: 18.8
        p99: 40.5
      Scenario duration:
        min: 126.5
        max: 1233.1
        median: 168.7
        p95: 486.6
        p99: 927.8
      Scenario counts:
        0: 600 (100%)
      Codes:
        200: 12000

diogob:
During all the tests above the memory consumption was fairly stable (< 20MB) for all scenarios.

This is evidence that this PR would not cause any regression for people that are not using the websockets feature.

I'm going to proceed with some memory profiling and tests against websockets connections.

diogob:
I found two problems so far:

* There is a memory leak somewhere, I still haven't found the culprit.
*  It seem that there is a lot of idling during websockets tests, I don't see any visible bottleneck in the CPU, maybe this is some STM contention (due to the sharing of one single database connection).

I'll investigate further and keep updating this thread.

diogob:
@ruslantalpa @begriffs as it turns out the problems I posted in the previous comment were bad reading of the events on my part 😛 

There was no leak, it was just the memory ramping up because of new connections being established. When I proceeded with further testing the memory stabilized. As old clients disconnect and new ones come in memory seems to be behaving well.

Also the idling was probably due to the lack of concurrency. Since there is a big overhead in the handshake for the claims validation to improve the throughput I needed longer sessions so the websockets would have more concurrent activity. After tweaking that I was able to create bottlenecks.

In the chart bellow you can see 2 contiguous sessions with an interval were memory is properly reclaimed:

screen shot 2017-05-28 at 1 33 05 pm

In the chart above I've reached ~500 simultaneous clients, using 16 database connections (from the postgrest pool which was set to max at 50) for the `NOTIFY` commands.

If anyone wants the files to reproduce the results I can upload to a new repo.

I consider this good enough to merge (performance-wise) so I'll proceed to write the docs 😉

diogob:
BTW, just as a side note praising haskell, STM rocks! 🍻

ruslantalpa:
@diogob how about having some kind of compile flag that would include/exclude the ws functionality at compile time?
this way you could have in your repository ws enabled postgrest distributions.

diogob:
@ruslantalpa sounds good, but I would probably just keep postgrest-ws as is now (with a sample server) since I'm not using this in production.

diogob:
@ruslantalpa @begriffs let's just decide if this is going to be merged before I write any docs 😄 

@ruslantalpa is there any other concern that I could address to bring your vote to yes?

My only issue with the compile time switch is that is does not seem to bring any advantage over the runtime switch (which is False by default).

ruslantalpa:
i don't have any concerns performance wise.
My main issue is that this functionality is useful in a limited number of cases (postgres pub/sub is very limited), so it will not be used by most of the people and it will just sit there even if it does not have  a big performance hit. 
I would vote for shipping 2 distributions, with and without the the ws part.

PierreRochard:
I'm just a hobbyist but my view is that change feeds are crucial for building a modern web application and this WS extension is a great starting place.

ruslantalpa:
@PierreRochard  change feeds are important and i am not debating that, i am just saying that that postgres pub/sub, and by extension postgres-ws is very limited.
Take the simple case of an blogging system where each user can login and edit his (unpublished) articles. Whenever he makes a change, there should be an event propagated in the system, but that event should only be see by him, the owner of the article. This means that the only way to do it is to have a distinct channel for each user that connects to the system. Simple example but already you can see the limitations, i don't think postgres pub/sub was optimised for tens of thusands of distinct channels. Then what about a single event that should be sent to multiple users, this means you need to have logic to publish the same event to multiple channels or have the clients listen on multiple channels.
What you want to have is a way for the database to emit events to a handful of channels, funnel those events into a dedicated messaging server (rabbit/Kafka) and then, within that system create complex routing topologies for the messages originating from the database.

diogob:
@ruslantalpa what do you think of having a run time plugin system for these scenarios? Like the `contrib` in PostgreSQL. We could use something like the [plugin package](https://hackage.haskell.org/package/plugins). It greatly increases the complexity of this specific solution, but adds another feature that might be very valuable for the future.

ruslantalpa:
a plugin system would be ideal but as you said it's a lot of work, that's why i suggested the first stage to be a compile time flag.
If you want to work on a plugin system i am all for it.
The only suggestion i would make is for ppl to pitch ideas of plugins, see if there are a lot of possibilities/uscases so that you don't develop a plugin system just for one plugin

ruslantalpa:
actually i just though of a usecase.
a great plugin would be one where you could configure by which columns a user can filter using which operators. Sort of like whitelist/blacklist kind of thing. For example one could deny filtering on a big text field using `like`.

if this plugin system is in place, then middleware things like jwt/logging/gzip can all be extracted into plugins and selectively enabled/disabled (for example i don't find the logging thing very useful since the proxy is logging all the requests).

diogob:
Yeah, I have another couple of plugins in mind. I'll ask on gitter later to gather some feedback.
@begriffs what do you think of a plugin system?

begriffs:
@ruslantalpa this change seems pretty good for a number of reasons:

* Binary goes from 45 to 46 mb on mac (after I run `strip` on it like I do for the releases)
* The additional compilation time was minimal
* @diogob's benchmarks show no problems
* Code changes are +24/-2 lines, very small
* Functionality can be completely disabled at runtime

It seems like your objections are:

a) The feature does not do enough, or cover enough cases
b) Adding a listener per distinct user may case performance problems in the db

My reaction is

a) It's optional, and the upvotes on this issue indicate interest in even this basic level of functionality
b) Hard to say how efficient postgres is at managing listeners, and whether the distinct channel per user is even a requirement for all systems. What would be a good way to measure this so we make an informed decision?

I'd also say it's premature to build a plugin system, unless we definitely have some other features we want to conditionally enable/disable. @diogob let me know what extras you have in mind.

ruslantalpa:
I am not debating the usefulness of this module or it's merits, it's a completely valid/useful/good piece of code for some usecases, however, it's place is not in PostgREST core.

- the code in postgrest and ws have very little overlap and does not need to interact (the control does not jump from one thing to another). It's like there is some top level code which decides what module handles the current request. This means postgrest-ws can be a completely separate executable on it's own, and the integration can be done by a proxy.
- even the names suggest these are two separate components of a system that do not interact with each other REST/Websockets. The only thing they have in common is they are both listening on port 80. No need to push all the functionality in a single executable.
- just because it's included as a remote module, does not mean it's only 24 lines. This is the same as if you copied all the files in the current repo. You need to also count the lines in that module since they are now part of the core postgrest functionality/logic, so it's more like 500 loc, which is about 16% increase in the loc (postgrest currently has 2800). 
- Users are always going to vote in favour of a new feature without dwelling to much on the implications.

diogob:
@begriffs the additional plugins I've imagined (I'm not planning to implement any of these, but they have potential) are:

* Use data constraints to provide better error messaging (this would be done in a very oppinionated way, so I think it does not belong to the core code base).
* Instrumentation plugins (think newrelic, datadog, that sort of stuff).
* Different logging plugins (logstash).
* A rails session integration using [rails-session](http://hackage.haskell.org/package/rails-session)
* A GraphQL frontend using [graphql-api](http://hackage.haskell.org/package/graphql-api).
* We could also extract the pre-request procedure call, the OpenAPI docs and the JWT claims verification into plugins.

I know that having run time plugins is a much taller order, but I'd be down to give it a shot.

begriffs:
You're right, that does make sense. OK should we close this PR then and open an issue to discuss a plugin architecture?

diogob:
Sure, I'm closing it. I'll start to do research on the run time library loading.

diogob:
meanwhile, if anyone stumbles across this PR and was hoping to find some websockets functionality you can check the [postgrest-ws](https://github.com/diogob/postgrest-ws) project. It offers a stand-alone server and a library with the WAI middleware.

ruslantalpa:
@diogob i've been meaning to write here. I was not agains ws and it's certainly a very useful tool to have that solves a lot of use cases (though not all). My reluctance to including it in the core was because it's a complementary tool and does not relate to the task postgrest is designed for (REST).
I think it can stand very well on it's own and integrated in a system using the proxy in front, or as a plugin.

Even though this PR did not get merged, i think a very good thing will come out of it, the beginning of the plugin ecosystem!

If we think a little more ahead, we can think even of PostgREST as a plugin, at the same level as postgrest-ws, each of them handling calls to different urls and something servant based at the top, integrating them both. Something to think about ...

diogob:
@ruslantalpa I certainly agree! No hard feelings here. Besides, doing those tests on memory was really useful and instructive regardless of the PR. I'm currently testing the plugins package and thinking of some refactors that would make the whole plugin ecosystem simpler to develop. 🍻

krakrjak:
@ruslantalpa 

> If we think a little more ahead, we can think even of PostgREST as a plugin, at the same level as postgrest-ws, each of them handling calls to different urls and something servant based at the top, integrating them both. Something to think about ...

This is exactly how I'm building apps with postgrest. I can see a bright future in this line of thinking.

Issue #861, opened 2017-04-15
OpenAPI doesn't show keys


timwis:
Is it just me, or does the OpenAPI page not show the keys? I believe it's meant to be at `definitions.{tablename}.key[]`, but it looks like this isn't supported in [the code](https://github.com/begriffs/postgrest/blob/f3b79bcc402e201b263dac9fdb9bf26ff01d8636/src/PostgREST/OpenAPI.hs#L53). I'm trying to update a row using its primary key as the condition, but I'm not sure how I'd figure out the primary key without this?

May be related to #790

ldesgoui:
Hello @timwis ,
I'm currently working on a PR to make the OpenAPI generation a little better.
Would you mind telling me more about this "key" attribute you speak of ? I've looked it up in the OpenAPI Spec and I can't seem to find it. I've also looked for terms relating to "primary key", "identifier", but that didn't yield anything either.
Thanks!

timwis:
Wow, I must be losing my mind. I swear I was referring to openapi documentation . . . maybe I got mixed up with json table schema? Either way, the real problem here is that there's no way to find out a table's primary keys. But the issues go a bit beyond that: there's also no way to see the default value (and thus determine whether a field is auto-incrementing), whether it's nullable, whether it's a foreign key, etc.

Here's what I ended up using for getting the schema for my app:
https://github.com/timwis/dataface/blob/master/db/get-schema.sql

It's a stored procedure that returns `name`, `type`, `length`, `default`, `null`, `constraint`, and `custom`. `custom` is actually the column's `COMMENT` converted to JSON, allowing me to store arbitrary meta data (like if it's a foreign key, which column from the foreign table should be used for display purposes).

ldesgoui:
Thanks for the answer.
Now that I'm knee deep in it, I will probably add something really similar using the ability to extend the spec using `x-custom-field` and/or written in descriptions.

begriffs:
Not to muddy the waters, but as I was just commenting in #790 we might consider a simpler format than api description format than openapi, whether it be a big official standard or no. Just thinking what will actually be simple and usable for people. Feel free to jump in on that discussion.

ruslantalpa:
This seems to be implemented with #885

priyank-purohit:
Looks like it's still not showing keys, should be re-opened.

Issue #817, opened 2017-02-26
Add schema to GET responses in OpenApi definition


samuba:
Right now the OpenApi definition does not provide a schema for the GET responses.
Those definitions already exists (they are e.g. used for POST parameters) and it would be nice to also have them included in the GET responses.
This would make the API definition more complete and enable clients, for example swagger-ui, to always show the model or even example results of the endpoint. 
See for example: http://petstore.swagger.io/#!/pet/getPetById 

Currently the OpenApi definition for a GET endpoint response looks like this:
```
"responses": {
    "206": {
        "description": "Partial Content"
     },
     "200": {
         "description": "OK"
     }
 }
```

Including the schema would look like this:
```
"responses": {
    "206": {
        "description": "Partial Content"
     },
     "200": {
         "description": "OK",
         "schema": {
             "$ref": "#/definitions/film"
          }
     }
 }
```

mpwang:
I think the change is necessary, otherwise when using swagger-codegen, the generated library will return none for any get method

ruslantalpa:
fixed by #885

Issue #717, opened 2016-10-14
Document foreign key relationships in autogenerated API docs


begriffs:
@adambaker requested this in order to make an admin panel for postgrest APIs.

ruslantalpa:
what would this admin panel "administer"?

begriffs:
It would allow the user to see the list of all tables and edit their contents. It would also want to know about fk relationships, so that if you're editing a `Store`, say, the program will know it has `Product`s and will show links to add and edit them.

ruslantalpa:
This can be implemented outside PostgREST as explained here https://github.com/begriffs/postgrest/issues/13#issuecomment-319890673

Issue #222, opened 2015-07-07
Expose comments on views and columns


sveniu:
This is a feature request to expose PG's [object comments](http://www.postgresql.org/docs/9.4/static/sql-comment.html) for views and their columns.

Several meta attributes are already exposed via the OPTIONS method, for both views and columns. For views, the comment can be fetched with `obj_description(object_oid, catalog_name)`. For columns within a view, the view OID must be accompanied by the column's position; this is already fetched – presumably via the information_schema.columns.ordinal_position column – and then fed to `col_description(table_oid, column_number)`.

PG itself mixes the terms "comment" and "description", so I'm not sure which one would make most sense to use in the API.

begriffs:
Very cool idea for documenting the API. :+1:

How would you say the resulting JSON response from OPTIONS should look when extended with the new information?

sveniu:
For GET or OPTIONS requests to `/`, I imagine simply adding a "description" attribute to each view's json object. This description would then be extracted verbatim from the comment added to the view in postgres.

For OPTIONS requests to a `/view`, I imagine that the view description is available as a top-level "description" attribute, and a column "description" attribute is available in each column's json object.

Since the comment is a plain text string, it probably makes sense to make it available verbatim, without any processing.

diogob:
+1

dweinstein:
👍 

would be awesome to see the comments for table and columns in the OPTIONS response and calling it `description` in each object.

would this be a good first contribution for someone?

diogob:
That would be a good issue to get acquainted with the code indeed. @dweinstein if you need any help feel free to ping me ;)

FGRibreau:
:+1: this could be used in the OpenAPI/Swagger spec (Postgraphql does that as well for GraphQL).

ruslantalpa:
Implemented in #885 plus you can have your custom view describing the api in the format you want by using a query/view like this one https://gist.github.com/ruslantalpa/b2f10eb1b5f6dd0fc1c154e071a1c91b

PR #885, opened 2017-06-03
Update OpenAPI (SQL COMMENT to description, constraints, cleaning up)

ldesgoui:
Greetings,

This PR addresses the idea raised in #222. 
I launched the PR to ask feedback: what do you feel deserves change in the current OpenAPI generation ? 

To do before completion:
- [x]  Schema COMMENT to info.description
- [x]  Table COMMENT to path..description and definitions..description
- [x]  Procedure COMMENT to path..description and definitions..description
- [x]  Column COMMENT to path...parameters..description and definitions..parameters..description
- [x]  ( #861  #717 ) Display table constraints (pk, fk, etc)
- [x]  ( #817 ) Link schemas in path responses to matching definitions
- [x]  ( #790 ) Clean repetitive parameters/headers
- [x]  Link to Postgrest end user manual or include parts of manual to describe the shared 
- [x]  Update tests
- [x]  Write changelog

Most of this seems trivial and should be accomplished soon :)

Side note:
I've updated the ToJSON instances but I don't think they're used anywhere, maybe they could get removed? (EDIT: removed now)
What about fluffing up the hard-coded OpenAPI options a little more while I'm at it ? They're not the most welcoming in the world and could definitely redirect developers to postgrest.com documentation

Links:
[Postgres SQL Comments](https://www.postgresql.org/docs/9.6/static/sql-comment.html)
[PostgraphQL's introspection query](https://github.com/postgraphql/postgraphql/blob/master/resources/introspection-query.sql)


begriffs:
You know I'm really not an OpenAPI expert. I think wherever you want the comments to appear in the output is probably a great start. Perhaps other people can chime in here. Just doing the hard part of obtaining those comments and putting them somewhere is wonderful, and it will be easier to tweak later on as people have suggestions as they use it. Having some repetition seems OK. The output can be pretty huge anyway, and I'd rather have API consumers notice the comments than overlook them.

I do like the idea of allowing a schema comment to override the generic API description placeholder. Also regarding your side note, feel free to try removing unused instances and cleaning up the code. My philosophy is that if the existing tests keep passing then we should be good.

Thanks for jumping in and working on this! I know people have wanted it for a while. Once you add some tests and a changelog entry we can look into getting this merged.

Anybody with opinions please add a comment.

begriffs:
I learned from previous work that when the openapi test times out it is actually failing because of malformed openapi output. Not sure why it does't just return an error. In the past I copied the output into an online validator to get a useful error message.

begriffs:
I know the test suite isn't very helpful when it fails for OpenAPI validation. Let's see if we can work together to figure out the problem. I checked out your branch, built it, and ran postgrest against the database schema created by our test suite. I was able to get more information about the validation failure by copying the openapi output into http://bigstickcarpet.com/swagger-parser/www/index.html :

```
Swagger schema validation failed. 
  Too few properties defined (0), minimum 1 at #/paths//withUnique/delete/responses
 
JSON_OBJECT_VALIDATION_FAILED

Error: Swagger schema validation failed. 
  Too few properties defined (0), minimum 1 at #/paths//withUnique/delete/responses
 
JSON_OBJECT_VALIDATION_FAILED
    at http://bigstickcarpet.com/swagger-parser/www/js/bundle.min.js:11:318
    at http://bigstickcarpet.com/swagger-parser/www/js/bundle.min.js:50:316
    at <anonymous> 

SyntaxError: Swagger schema validation failed. 
  Too few properties defined (0), minimum 1 at #/paths//withUnique/delete/responses
 
JSON_OBJECT_VALIDATION_FAILED
    at Function.syntax (http://bigstickcarpet.com/swagger-parser/dist/swagger-parser.min.js:284:318)
    at validateSchema (http://bigstickcarpet.com/swagger-parser/dist/swagger-parser.min.js:20:254)
    at http://bigstickcarpet.com/swagger-parser/dist/swagger-parser.min.js:8:1921
    at <anonymous> 

z-schema validation error: JSON_OBJECT_VALIDATION_FAILED
    at ZSchema.getLastError (http://bigstickcarpet.com/swagger-parser/dist/swagger-parser.min.js:2047:3479)
    at validateSchema (http://bigstickcarpet.com/swagger-parser/dist/swagger-parser.min.js:20:158)
    at http://bigstickcarpet.com/swagger-parser/dist/swagger-parser.min.js:8:1921
    at <anonymous>
```

Does this error help?

ldesgoui:
Oh, that seems easy to fix, `deleteOp` in `OpenAPI.makePathItem` doesn't contain a "at" field
Thanks a lot for the help. I'll be continuing work on the PR this week, unless you'd rather discuss #790 further.

begriffs:
That discussion would affect another major version anyway, and we still may keep the openapi description but use content negotiation (the Accept request header) to show it rather than show it by default. Also extracting sql comments into our in-memory data structure will be useful no matter how we output the api description so carry on!

ldesgoui:
I tried using the tool you linked but I had no chance with it, when I click "Validate it!", a pretty-print of the description shows up in a box that shakes for a second and nothing else happens/shows up, where did you get those error messages?

ldesgoui:
How the output looks after the removal of all duplication: https://gist.github.com/ldesgoui/faecf9fec35e150f67be1cb7843c1f7e

ldesgoui:
Misclick, sorry :frowning:

begriffs:
The new output looks really nice, I can actually read it directly rather than the old avalanche of text. Great work!

Did you manage to figure out the validator I was using? Looks like the tests are failing on the most recent commit except this time they are giving a real error message rather than just hanging.

ldesgoui:
I did, yes, the output is valid now. The tests fail because they need to be updated to reflect changes

ldesgoui:
Hey, I'm done with this PR, here's a recap:
- Dead ToJSON intances were remvoed 
- ProcDescription, Table, Column received a new Maybe Text field for descriptions
- Introspection received an additional SQL query to get the Schema COMMENT
- RPC body schema was inlined as it's only used once
- All Table path properties were moved to Parameter Definitions as they're all used more than once
- OpenAPI links to PostgREST docs

src/PostgREST/OpenAPI.hs
```diff
@@ -36,30 +37,48 @@ toSwaggerType "boolean"   = SwaggerBoolean
...
+      , fk <$> colFK c
+      ]
+    d =
+      if length n > 1 then
+        Just $ append (fromMaybe "" ((`append` "\n\n") <$> colDescription c)) (intercalate "\n" n)
+      else
+        colDescription c
+    s =
+      (mempty :: Schema)
+        & default_ .~ (decode . TL.encodeUtf8 . TL.fromStrict =<< colDefault c)
```
begriffs: Is it possible to use the more generic `toS` to convert the string types? Something like

```haskell
default_ .~ (toS =<< colDefault c)
```

Not sure if it can infer the types, but thought I'd ask.

begriffs:
Thanks for all your work to research this and finish it!

ldesgoui:
My pleasure, postgrest is pure joy to use in projects :)

test/Feature/StructureSpec.hs
```diff
@@ -27,18 +27,82 @@ spec = do
...
+                    "format": "integer",
+                    "type": "integer"
+                  },
+                  "name": {
+                    "description": "child_entities name comment",
+                    "format": "text",
+                    "type": "string"
+                  },
+                  "parent_id": {
+                    "description": "Note:\nThis is a Foreign Key to `entities.id`.<fk table='entities' column='id'/>",
```
steve-chavez, line 89: @ldesgoui I've always wondered.. Where does the `<fk>` and `<pk>` tags come from?

steve-chavez:
@ldesgoui I see, thanks for the quick reply!

According to https://github.com/OAI/OpenAPI-Specification/issues/587, another way of specifying the pk/fk would be with [OpenAPI extensions](https://swagger.io/docs/specification/openapi-extensions/).
This could be a future enhancement.

Issue #884, opened 2017-06-02
Broken link in CONTRIBUTING.md


eric-brechemier:
In [CONTRIBUTING.md](https://github.com/begriffs/postgrest/blob/10f6ce45dbf8b1652d07446aba98743b071aef6c/CONTRIBUTING.md), the link [building from source](http://postgrest.com/install/server/#building-from-source) leads to a 404 page.

PR #886, opened 2017-06-03
Fix broken link to 'building from source' in CONTRIBUTING

ldesgoui:
Fixing #884


begriffs:
Thanks, good catch. An even better link might be https://postgrest.com/en/stable/install.html#build-from-source because it will track the newest tagged version as it changes.

Issue #858, opened 2017-04-07
Improve internal documentation


begriffs:
Haddock warns about many top level functions and types that lack documentation. Some of these functions speak for themselves, but others would benefit from comments explaining their motivation and context.

```
Haddock coverage:
   0% (  0 /  9) in 'PostgREST.RangeQuery'
  Missing documentation for:
    Module header
    rangeParse (src/PostgREST/RangeQuery.hs:28)
    rangeRequested (src/PostgREST/RangeQuery.hs:40)
    rangeLimit (src/PostgREST/RangeQuery.hs:50)
    rangeOffset (src/PostgREST/RangeQuery.hs:56)
    restrictRange (src/PostgREST/RangeQuery.hs:44)
    rangeGeq (src/PostgREST/RangeQuery.hs:62)
    allRange (src/PostgREST/RangeQuery.hs:66)
    NonnegRange (src/PostgREST/RangeQuery.hs:26)
   7% (  4 / 54) in 'PostgREST.Types'
  Missing documentation for:
    Module header
    ApiRequestError (src/PostgREST/Types.hs:17)
    DbStructure (src/PostgREST/Types.hs:26)
    PgArg (src/PostgREST/Types.hs:34)
    PgType (src/PostgREST/Types.hs:40)
    RetType (src/PostgREST/Types.hs:42)
    ProcVolatility (src/PostgREST/Types.hs:44)
    ProcDescription (src/PostgREST/Types.hs:47)
    Schema (src/PostgREST/Types.hs:54)
    TableName (src/PostgREST/Types.hs:55)
    SqlQuery (src/PostgREST/Types.hs:56)
    SqlFragment (src/PostgREST/Types.hs:57)
    RequestBody (src/PostgREST/Types.hs:58)
    Table (src/PostgREST/Types.hs:60)
    ForeignKey (src/PostgREST/Types.hs:66)
    Column (src/PostgREST/Types.hs:68)
    Synonym (src/PostgREST/Types.hs:85)
    PrimaryKey (src/PostgREST/Types.hs:87)
    OrderDirection (src/PostgREST/Types.hs:92)
     (src/PostgREST/Types.hs:93)
    OrderNulls (src/PostgREST/Types.hs:97)
     (src/PostgREST/Types.hs:98)
    OrderTerm (src/PostgREST/Types.hs:102)
    QualifiedIdentifier (src/PostgREST/Types.hs:108)
    RelationType (src/PostgREST/Types.hs:114)
[...]

PR #888, opened 2017-06-08
comments for Main.hs

Skyfold:
This is first part to solving the issue #858. I am planning on doing a separate pull request for each file I comment so I can get feedback while I'm doing this. Mostly this is to get a sense both for what the community wants in terms of documentation and to catch any misunderstandings I have of the code.

I did make a minor code change, `s` became `secret` for readability. I will refrain from making more complex changes until I have a good understanding of the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-### Fixed
-
-## [0.4.2.0] - 2017-06-11
-
 ### Added
 
 - #889, Allow more than two conditions in a single and/or - @steve-chavez
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+## [0.4.2.0] - 2017-06-11
+
 ### Added
 
 - #742, Add connection retrying on startup and SIGHUP - @steve-chavez
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #889, Allow more than two conditions in a single and/or - @steve-chavez
+
+### Fixed
+
+- #896, Boolean env var interpolation in config file - @begriffs
+
+## [0.4.2.0] - 2017-06-11
+
+### Added
+
 - #742, Add connection retrying on startup and SIGHUP - @steve-chavez
 - #652, Add and/or params for complex boolean logic - @steve-chavez
 - #808, Env var interpolation in config file (helps Docker) - @begriffs
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -2,160 +2,151 @@
 
 module Main where
 
-import PostgREST.App (postgrest)
-import PostgREST.Config
-       (AppConfig(..), PgVersion(..), minimumPgVersion, prettyVersion,
-        readOptions)
-import PostgREST.DbStructure (getDbStructure)
-import PostgREST.Error (encodeError)
-import PostgREST.OpenAPI (isMalformedProxyUri)
-import PostgREST.Types (DbStructure, Schema)
-import Protolude
+import           PostgREST.App            (postgrest)
+import           PostgREST.Config         (AppConfig (..),
+                                           PgVersion (..),
+                                           minimumPgVersion,
+                                           prettyVersion, readOptions)
+import           PostgREST.DbStructure    (getDbStructure)
+import           PostgREST.Error          (encodeError)
+import           PostgREST.OpenAPI        (isMalformedProxyUri)
+import           PostgREST.Types          (DbStructure, Schema)
+import           Protolude
 
-import Control.AutoUpdate
-import Control.Retry
-import Data.ByteString.Base64 (decode)
-import Data.IORef (IORef, atomicWriteIORef, newIORef, readIORef)
-import Data.String (IsString(..))
-import Data.Text (pack, replace, stripPrefix)
-import Data.Text.Encoding (decodeUtf8, encodeUtf8)
-import Data.Text.IO (hPutStrLn, readFile)
-import Data.Time.Clock.POSIX (getPOSIXTime)
-import qualified Hasql.Decoders as HD
-import qualified Hasql.Encoders as HE
-import qualified Hasql.Pool as P
-import qualified Hasql.Query as H
-import qualified Hasql.Session as H
-import Network.Wai.Handler.Warp
-       (defaultSettings, runSettings, setHost, setPort, setServerName,
-        setTimeout)
-import System.IO (BufferMode(..), hSetBuffering)
+import           Control.AutoUpdate       (defaultUpdateSettings,
+                                           mkAutoUpdate, updateAction)
+import           Control.Retry            (RetryStatus, capDelay,
+                                           exponentialBackoff,
+                                           retrying, rsPreviousDelay)
+import           Data.ByteString.Base64   (decode)
+import           Data.IORef               (IORef, atomicWriteIORef,
+                                           newIORef, readIORef)
+import           Data.String              (IsString (..))
+import           Data.Text                (pack, replace, stripPrefix)
+import           Data.Text.Encoding       (decodeUtf8, encodeUtf8)
+import           Data.Text.IO             (hPutStrLn, readFile)
+import           Data.Time.Clock.POSIX    (getPOSIXTime)
+import qualified Hasql.Decoders           as HD
+import qualified Hasql.Encoders           as HE
+import qualified Hasql.Pool               as P
+import qualified Hasql.Query              as H
+import qualified Hasql.Session            as H
+import           Network.Wai.Handler.Warp (defaultSettings,
+                                           runSettings, setHost,
+                                           setPort, setServerName,
+                                           setTimeout)
+import           System.IO                (BufferMode (..),
+                                           hSetBuffering)
 #ifndef mingw32_HOST_OS
-import System.Posix.Signals
+import           System.Posix.Signals
 #endif
+
+{-|
+	Used by connectionWorker to know if it should throw an error and kill the
+  main thread.
+-}
+isServerVersionSupported :: H.Session Bool
+isServerVersionSupported = do
+  ver <- H.query () pgVersion
+  return $ ver >= pgvNum minimumPgVersion
+ where
+  pgVersion =
+    H.statement "SELECT current_setting('server_version_num')::integer"
+      HE.unit (HD.singleRow $ HD.value HD.int4) False
 
 {-|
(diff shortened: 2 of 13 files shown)


ldesgoui:
I've noted a few typos:
L59: caling -> calling
L62: runnig -> running
L193: SIGNUP -> SIGHUP

Great work

Skyfold:
Fixed with a rebase. (I'll make sure spell check is on next time...)

Thanks

main/Main.hs
```diff
@@ -82,7 +99,15 @@ connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
             atomicWriteIORef refIsWorkerOn False
             putStrLn ("Connection successful" :: Text)
 
--- | Connect to pg server if it fails retry with capped exponential backoff until success
+{-|
+  Only used by 'connectionWorker' to check if the provided db-uri lets
```
steve-chavez: Why is valuable to comment the scope of the function(Only used by..), I think that's better expressed by the code.

main/Main.hs
```diff
@@ -2,39 +2,46 @@
...
+                                           setPort, setServerName,
+                                           setTimeout)
+import           System.IO                (BufferMode (..),
+                                           hSetBuffering)
 #ifndef mingw32_HOST_OS
 import           System.Posix.Signals
 #endif
 
+{-|
+	Used by connectionWorker to know if it should throw an error and kill the
```
steve-chavez: Same thing as before(Used by..).

Skyfold: What do you think about merging isServerVersionSupported and connectingSucceeded into connectionWoker? I can re-write the code a bit to make the merged code more redable.

steve-chavez: Sure, go ahead. I actually tried that before but made the code less readable because of nested where's, maybe you can come up with a better way.

Skyfold:
Sorry I took so long to get back to this. 

I merged the three functions in what I hope is a clear manner. Hindent does not work well with `#ifndef`; I had to remove that block, call hindent, then put it back in. Please let me know if hindent made changes you don't like or if merging the three functions needs to be done more elegantly.

Skyfold:
Strangely, putting a space before `#ifndef` and `#endif` on OSX compiles fine, but on Ubuntu it does not. That took me while to figure out what was wrong; sorry for all the failed tests on CircleCI.

begriffs:
Thanks for this! I'll look at it in more detail in the next few days.

main/Main.hs
```diff
@@ -2,180 +2,290 @@
...
+      appSettings =
+        setHost ((fromString . toS) host) -- Warp settings
+         .
+        setPort port .
+        setServerName (toS $ "postgrest/" <> prettyVersion) .
+        setTimeout 3600 $
+        defaultSettings
+  --
+  -- Check if the connection string to the database is in the right format
+  when (isMalformedProxyUri $ toS <$> proxy) $
```
steve-chavez: Is this comment correct? The doc in ```isMalformedProxyUri``` in OpenAPI.hs module says it does a proxy uri check not a db uri check.

main/Main.hs
```diff
@@ -2,180 +2,290 @@
...
+            " seconds..."
+          return itShould
+
+
+{-|
+  This is where everything starts.
 
+  Things that can throw errors: loadSecretFile
+                              , readOptions
+                              , when isMalformedProxyUri
```
steve-chavez: I'm working on a way to detect the used libpq version and this may throw another error, this kind of comment can get outdated pretty quickly and is easy to forget to update it.
I think it's better to just leave the code to express this.

main/Main.hs
```diff
@@ -2,180 +2,290 @@
...
+import Data.Time.Clock.POSIX (getPOSIXTime)
+import qualified Hasql.Decoders as HD
+import qualified Hasql.Encoders as HE
+import qualified Hasql.Pool as P
+import qualified Hasql.Query as H
+import qualified Hasql.Session as H
+import Network.Wai.Handler.Warp
+       (defaultSettings, runSettings, setHost, setPort, setServerName,
+        setTimeout)
+import System.IO (BufferMode(..), hSetBuffering)
```
steve-chavez: Why the change in the imports indentation? I find the previous way more readable. First thing I did before contributing to PostgREST was to read all the import declarations to learn the used libraries. I think is important to keep readability here.

main/Main.hs
```diff
@@ -2,180 +2,290 @@
...
-              <> pgvName minimumPgVersion)
-            killThread mainTid
-          dbStructure <- getDbStructure schema
-          liftIO $ atomicWriteIORef refDbStructure $ Just dbStructure
+        result <-
+          P.use pool $ do
+            supported <-
+              fmap
+                (\ver -> ver >= pgvNum minimumPgVersion)
+                (H.query () pgVersion)
```
steve-chavez: The ```isServerVersionSupported``` was more concise here(though maybe it could be renamed to ```isPgVersionSupported```). Previous ```connectionWorker``` was already a long function, by merging the 3 functions it got longer, the indentation increased and there are nested where's, overall I think this is less clear than before. Maybe it would be better to go back to having these as top-level functions.

main/Main.hs
begriffs: I agree about the old alignment being nice. The suggestions in [this article](https://hackernoon.com/4-steps-to-a-better-imports-list-in-haskell-43a3d868273c) about imports seem pretty good. Maybe you could review our import lists and get them all to follow those guidelines.

main/Main.hs
```diff
@@ -2,180 +2,290 @@
...
+  Things that can throw errors: loadSecretFile
+                              , readOptions
+                              , when isMalformedProxyUri
+-}
 main :: IO ()
 main = do
+  --
+  -- LineBuffering: the entire output buffer is flushed whenever a newline is 
+  --                output
+  --              , the buffer overflows
```
begriffs: Maybe just use regular english formatting rather than separate lines for each comma clause.

main/Main.hs
```diff
@@ -2,180 +2,290 @@
...
-                  . setTimeout 3600
-                  $ defaultSettings
-
-  when (isMalformedProxyUri $ toS <$> proxy) $ panic
-    "Malformed proxy uri, a correct example: https://example.com:8443/basePath"
-
+      pgSettings = toS (configDatabase conf) -- is the db-uri
+      appSettings =
+        setHost ((fromString . toS) host) -- Warp settings
+         .
```
begriffs: This composition dot floating on its own line looks kind of odd. There's probably a more uniform way to align the composition symbol on each of these lines.

Skyfold:
Thanks for the feedback! Very helpful for giving me a better idea of what you want in terms of commenting and formatting. Please let me know if I am bikeshedding at any point. 

- Let me poke around with some other Haskell code formating tools. I agree that hindent isn't formating the imports very nicely. If I cannot find a tool that matches your linked article I'll just do it manually
- I'll remove the comments about functions that throw errors, it would be too easy to miss one (or forget to add one)
- I'll pull out the comment formatting for comments like the one on Line Buffering
- Sounds like I made a mistake with `isMalformedProxyUri`, will fix
- I will un-merge the `connectionWorker` function

begriffs:
@Skyfold thanks, looking forward to your changes.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+## [0.4.2.0] - 2017-06-11
```
begriffs: This appears to be a duplicate, see line 20.

main/Main.hs
```diff
@@ -45,14 +55,30 @@ isServerVersionSupported = do
       HE.unit (HD.singleRow $ HD.value HD.int4) False
 
 {-|
+  The purpose of this worker is to fill the refDbStructure created in 'main'
+  with the 'DbStructure' returned from calling 'getDbStructure'. This method
+  is meant to be called by multiple times by the same thead, but does nothing if
+  the previous invocation has not terminated. In all cases this method does not
+  halt the caling thead, the work is proformed in a separate thread.
```
begriffs: caling -> calling, proformed -> performed

main/Main.hs
```diff
@@ -45,14 +55,30 @@ isServerVersionSupported = do
...
+
+  Note: 'atomicWriteIORef' is essentialy a lazy semaphore that prevents two
```
begriffs: essentialy -> essentially

Does your editor have a spell checker? I usually run my code/docs through one just in case, since it helps catch little typos.

main/Main.hs
```diff
@@ -45,14 +55,30 @@ isServerVersionSupported = do
...
+  threads from runnig 'connectionWorker' at the same time.
```
begriffs: runnig -> running

(I'll stop pointing these little things out, just give the comments a quick spell check)

main/Main.hs
```diff
@@ -103,79 +138,152 @@ connectingSucceeded pool =
...
+{-|
+  This is where everything starts.
+-}
 main :: IO ()
 main = do
+  --
+  -- LineBuffering: the entire output buffer is flushed whenever a newline is 
+  -- output, the buffer overflows, a hFlush is issued or the handle is closed
+  --
+  -- no-buffering: output is written immediately and never stored in the buffer
```
begriffs: Maybe list as `NoBuffering` rather than `no-buffering` to match the camel case style of `LineBuffering` in this comment.

main/Main.hs
```diff
@@ -103,79 +138,152 @@ connectingSucceeded pool =
...
   forM_ [sigINT, sigTERM] $ \sig ->
     void $ installHandler sig (Catch $ do
         P.release pool
         throwTo mainTid UserInterrupt
       ) Nothing
 
-  void $ installHandler sigHUP (
-      Catch $ connectionWorker mainTid pool (configSchema conf) refDbStructure refIsWorkerOn
-   ) Nothing
+  void $ installHandler sigHUP (Catch $ connectionWorker mainTid pool (configSchema conf) refDbStructure refIsWorkerOn) Nothing
```
begriffs: This is a pretty long line, are you sure you don't want to break it into a few lines?

main/Main.hs
Skyfold: Finally noticed that `set spell` was not in my vimrc. (Goes over all the other comments I have written)

Skyfold:
Fixed the changelog (when I merged with upstream I missed the duplication)

There should be no more spell errors.....

I am now using `stylish-haskell -c .stylish-haskell.yaml -i src/PostgREST/App.hs` to format the files. All the setting for it are in the `.stylish-haskell.yaml` file.

What do you think about this commenting structure where I put a block comment before each function then smaller comments throughout each function?

begriffs:
Looking good! I like the indentation and your comment convention for before and inside functions.

If you can take care of my comment earlier about the lonely "." character and align the composition better then I think this should be good to merge.

Skyfold:
Fixed the lonely ".", (sorry I missed that comment) and aligned the composition much like how lists are aligned.

I hope commenting the rest of the codebase will be swifter; I apologise for the long iteration times.

begriffs:
Thanks for your help, no need to apologize. Takes time to find an agreeable style for comments/indentation.

PR #891, opened 2017-06-10
Dockerfile: Clean up to save space

purpleKarrot:
Running postgrest requires the libpq library runtime, not the -dev
package.

Make sure that tools that are used to download and extract the postgrest
package do not remain installed in the image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,15 +1,24 @@
 FROM debian:jessie
 
-ENV POSTGREST_VERSION 0.4.1.0
+# Install libpq5
+RUN apt-get -qq update && \
+    apt-get -qq install -y --no-install-recommends libpq5 && \
+    apt-get -qq clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-RUN apt-get update && \
-    apt-get install -y tar xz-utils wget libpq-dev && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
-RUN wget http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}/postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz && \
-    tar --xz -xvf postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz && \
+# Install postgrest
+RUN POSTGREST_VERSION="0.4.1.0" \
+    BUILD_DEPS="curl ca-certificates xz-utils" && \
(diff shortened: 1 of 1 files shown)


purpleKarrot:
After: 176MB
Before: 239MB

ruslantalpa:
any reason why you two `RUN` statements and not one?

purpleKarrot:
I consider it cleaner with two `RUN` statements. In the first step we install the runtime dependencies. In the second one we install postgrest (we temporarily install some stuff and remove it in the same step again, but this is an implementation detail; the only side effect is that postgrest is installed afterwards).

Combining the two steps would make the `Dockerfile` only slightly smaller, and only slightly harder to maintain. It would not make a difference on the image size.

docker/Dockerfile
```diff
@@ -1,15 +1,24 @@
...
+    apt-get -qq clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Install postgrest
+RUN POSTGREST_VERSION="0.4.1.0" \
+    BUILD_DEPS="curl ca-certificates xz-utils" && \
+    apt-get -qq update && \
+    apt-get -qq install -y --no-install-recommends $BUILD_DEPS && \
+    cd /tmp && \
+    curl -SLO http://github.com/begriffs/postgrest/releases/download/v${POSTGREST_VERSION}/postgrest-${POSTGREST_VERSION}-ubuntu.tar.xz && \
```
begriffs: While you're at it you could switch to https

docker/Dockerfile
```diff
@@ -1,15 +1,24 @@
...
+# Install libpq5
+RUN apt-get -qq update && \
+    apt-get -qq install -y --no-install-recommends libpq5 && \
+    apt-get -qq clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Install postgrest
+RUN POSTGREST_VERSION="0.4.1.0" \
+    BUILD_DEPS="curl ca-certificates xz-utils" && \
+    apt-get -qq update && \
```
begriffs: What's the reason for this second apt-get update? Just because this RUN statement is a separate stage from the previous one and may be run later on independently?

purpleKarrot: `apt-get update` doesn't update any software.  It updates the package lists, which are deleted at the end of the `RUN` with `rm -rf /var/lib/apt/lists/*`. We do not want to keep the lists between runs. Everything that we preserve between runs will take space forever.

begriffs: Ah right you're deleting the package lists.

Issue #766, opened 2016-12-07
Support for custom output format on GET


steve-chavez:
I have a table: 

```sql
  create table workers(
    id char(8) primary key,
    first_name text not null,
    last_name text not null,
    birth_date date
  )
```

I need to download a file with this table rows in the following format:

```
09310817|JOHN|DOE|15/04/88|
42152780|FRED|BLOGGS|20/02/85|
43006541|OTTO|NORMALVERBRAUCHER|01/07/90|
02452492|ERIKA|MUSTERMANN|11/01/80|
```

Sort of a pipe separated values format with that extra pipe at the end, this is a requirement for a local goverment entity.

Currently there's no way to output that format with just postgREST I would need some custom application code, this is a problem since I also need the security and filtering that postgREST provides. 

I can construct the format in a column of a custom view, it would be convenient to have a way to tell postgREST that the JSON or CSV wrapping is not required and just output that column.

Seeing the following issues #257, #506 I see there's some overlap with my use case. 

My proposal is similar to this [one](https://github.com/begriffs/postgrest/issues/257#issuecomment-205047876) differing in that the required parameter only be selecting a single column ```select=col``` when doing an ```Accept: application/octet-stream``` if no single col is selected a ```406 Not Acceptable``` response could be applied.

The single valued rows of the response could be concatenated taking advantage of the ```string_agg``` pg function, that supports text and bytea.  

For my text case it's appropriate that postgREST concats the rows with a ```'\n'``` but not sure if this is good for the binary output, if that's the case then we could just concat with ```''``` and let the client decide the delimiter; though it would be better if the users of multiple row binary output clarify this.

I already have a working custom build of postgREST with this feature, I think it would be great to support this flexibility. Glad to do a PR if we reach an agreement.

ruslantalpa:
@steve-chavez i think you are overcomplicating things with this approach.
Maybe in some other cases your proposed method with one column would be needed but in this case i think it's not a good approach.
The way to do this is to postprocess the response from postgrest at the proxy level.
Since you are talking gov/security ... you need nginx in front of postgrest for security, since it's already there, go for openresty, once you have openresty, do this

- make a normal request to postgrest and request back a CSV format.
[...]

PR #895, opened 2017-06-25
Binary output support for RPC

### LINKED ISSUE #883: Add binary output to rpc calls
Please permit binary output for rpc `POST` requests and an `Accept: application/octet-stream` request header to use with it.
Right now I'm getting a response from a function returning `BYTEA` data, but it doesn't seem to be in the same binary form, that it is returned from a function in postgresql. `application/octet-stream` header is not set on response and request.
 
So I have to separate API from binary rpc calls and use [this nginx module](https://github.com/konstruxi/ngx_postgres/) instead of postgrest, which adds some inconveniences. 

My use case is quite simple: I'm serving `GEOMETRY` and other binary data out of my DB and would really like to use http rpc calls from postgrest endpoint.

Possible?

steve-chavez:
Fixes #883. By default, when requesting ```Accept: application/octet-stream``` the user has to do 
```?select=single_col``` similar to binary output on GET, but since a stored procedure can return a scalar    I considered that is not necessary to force the user to do a ```/rpc_name?select=rpc_name``` so in the case of scalar I'm not applying that restriction.  

To do that I changed ```callProc``` to make it consider if the proc is scalar by consulting the ```ProcDescription```, this also avoids doing a [count](https://github.com/begriffs/postgrest/blob/b00742814216be685fddcd3ed61c2ce81894fbf0/src/PostgREST/QueryBuilder.hs#L158-L161) to determine this. 

A disadvantage of that is that now for a proc to return the scalar properly(e.g. a rpc that returns an integer should return ```4``` not ```[ { "rpc" : 4 } ]```, this was #600) it needs the ```DbStructure``` up to date, similar to the resource embedding feature, maybe this needs to be documented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - #889, Allow more than two conditions in a single and/or - @steve-chavez
+- #883, Binary output support for RPC - @steve-chavez
 
 ### Fixed
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -238,8 +238,12 @@ app dbStructure conf apiRequest =
               let p = V.head payload
                   singular = contentType == CTSingularJSON
                   paramsAsSingleObject = iPreferSingleObjectParameter apiRequest
+                  proc = M.lookup (qiName qi) allProcs 
+                  returnsScalar = case proc of
+                    Just ProcDescription{pdReturnType = (Single (Scalar _))} -> True
+                    _ -> False
               row <- H.query () $
-                callProc qi p q cq topLevelRange shouldCount singular
+                callProc qi p returnsScalar q cq topLevelRange shouldCount singular
                          paramsAsSingleObject (contentType == CTTextCSV)
               let (tableTotal, queryTotal, body) =
                     fromMaybe (Just 0, 0, "[]") row
@@ -257,7 +261,7 @@ app dbStructure conf apiRequest =
               uri Nothing = ("http", host, port, "/")
               uri (Just Proxy { proxyScheme = s, proxyHost = h, proxyPort = p, proxyPath = b }) = (s, h, p, b)
               uri' = uri proxy
-              encodeApi ti = encodeOpenAPI (M.elems $ dbProcs dbStructure) ti uri'
+              encodeApi ti = encodeOpenAPI (M.elems allProcs) ti uri'
           body <- encodeApi . toTableInfo <$> H.query schema accessibleTables
           return $ responseLBS status200 [toHeader CTOpenAPI] $ toS body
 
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by begriffs** [comment]
Sorry for the delay, I'll review this tonight!

**Comment by begriffs** [comment]
Didn't get time to test this out today. But got a long weekend coming up so I'll get it done for sure.

**Comment by begriffs** [comment]
OK, @steve-chavez I tried it out and the behavior makes perfect sense not to require selecting a column for functions that return scalars. Over all it works really nicely. I did encounter a problem when attempting to get an octet-stream for multiple bigint values. Using the data from our text fixture you should be able to reproduce:

```http
GET http://localhost:3000/rpc/getitemrange?select=id
Accept: application/octet-stream
```

Post body:
```json
{"min": 0, "max":100}
```

Result HTTP 404 with message:
```json
{
  "hint": "No function matches the given name and argument types. You might need to add explicit type casts.",
  "details": null,
  "code": "42883",
  "message": "function string_agg(bigint, unknown) does not exist"
}
```

This appears to be caused by the type of the column to be concatenated. For instance `http://localhost:3000/rpc/getallprojects?select=name` can return an octet-stream just fine.

begriffs:
(Verified the same bug already exists for regular requests like `GET http://localhost:3000/items?select=id` with octet-stream)

steve-chavez:
@begriffs You can make those requests work if you cast the column to ```text``` as in:
```http
GET http://localhost:3000/items?select=id::text&limit=1
Accept: application/octet-stream

1
```
That works because the ```string_agg``` pg function is only defined for ```bytea``` and ```text``` type arguments. Also without ```limit```ing it would concat all ids and give ```123456789101112131415```, though I don't think an user would want that kind of response. 

Now this is the way it is because when I implemented the binary output(more like an unwrapped value) I took into account the similar issues related to requesting an unwrapped ```bytea```(#506, #257) and also requesting unwrapped ```text```(#766)(not an use case for me anymore and there were alternatives).

I can see a couple of options around that unclear error you showed:

1. Improve the error message by restricting the ```Accept: application/octet-stream``` to work with just the ```bytea``` type(col or proc return), we would lose the unwrapped ```text``` functionality here though this was undocumented and not the main use case more like an added one. 
2. Make the ```Accept: application/octet-stream``` work for all types by casting every ```non-bytea``` to ```text```, no functionality is lost here but I think is not semantically correct since a ```text/plain``` would fit better for this kind of response, also not sure right now if every pg type is castable to text.

Overall I'm in favor of adding the 1st option, if the need for getting an unwrapped ```text``` response arises again we could later see a more appropriate way to do it, what do you think?

steve-chavez:
I think this could be merged as it is, since fixing that behavior is not straightforward and requires some discussion, I can open a PR later(or maybe if an user hits the problem he'll open an issue before) with my proposal. Also since there are 2 fixes annotated in the commits and they are self-contained it maybe better to not squash them.

begriffs:
I see your point that concatenating the decimal representation of numbers isn't likely to be useful the way string concatenation is, so I guess it's just the obscurity of the error message that could be improved. But even there the corner case is not that big of a deal, just thought I'd point out what I noticed.

I'll merge this and preserve the individual commits in git history.

PR #911, opened 2017-07-14
Put traditional Filters inside [LogicTree]

steve-chavez:
This is for not having ```flt_::[Filter], logic::[LogicTree]``` and instead just a single 
```where_:[LogicTree]``` in the ```ReadQuery/MutateQuery``` types, this was mentioned in https://github.com/begriffs/postgrest/pull/864#issuecomment-296352326


begriffs:
Is this purely an implementation detail, or are there any tests that would be relevant to add to the test suite?

ruslantalpa:
I think it's an implementation detail/refactoring/cleanup.
The tests are already there of all the way to specify the filters (with boolean logic and without)

steve-chavez:
It's just a refactor for making the query types a bit more clear, the tests are already there for traditional filters and and/or, no functionality is changed.

begriffs:
OK, sounds good.

PR #913, opened 2017-07-14
Change singularityError zero rows status to 404

### LINKED ISSUE #908: 406 status code returned when no records found requesting a singular object
According to the documentation, a status code should be 404 https://postgrest.com/en/v4.1/api.html#singular-or-plural.

@ruslantalpa tells:

> https://github.com/begriffs/postgrest/blob/master/src/PostgREST/App.hs#L133
> should probably be
> `if contentType == CTSingularJSON && queryTotal > 1`
> and then a special case for total = 0 with a 404

steve-chavez:
Fixes #908. I think is more semantically correct(also in sync with the docs) to return 404 with no body when getting zero rows in:
```http
GET /projects?id=eq.0 
Accept: application/vnd.pgrst.object+json
```

But I'm not sure if that's the better response for POST/PATCH/DELETE with ```application/vnd.pgrst.object+json```, maybe for those cases it could be better to show the failed operation with the usual error message:
```json
{
    "details": "Results contain 0 rows, application/vnd.pgrst.object+json requires 1 row",
    "message": "JSON object requested, no rows returned"
}
```
Thoughts?

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -55,10 +55,13 @@ pgError authed e =
 
 singularityError :: Integer -> Response
 singularityError numRows =
-  responseLBS HT.status406
+  if numRows == 0
+    then responseLBS HT.status404 [toHeader CTSingularJSON] ""
+  else
+    responseLBS HT.status406
     [toHeader CTSingularJSON]
     $ toS . formatGeneralError
-      "JSON object requested, multiple (or no) rows returned"
+      "JSON object requested, multiple rows returned"
       $ unwords
         [ "Results contain", show numRows, "rows,"
         , toS (toMime CTSingularJSON), "requires 1 row"
--- a/test/Feature/SingularSpec.hs
+++ b/test/Feature/SingularSpec.hs
@@ -22,7 +22,7 @@ spec =
     context "with GET request" $ do
       it "fails for zero rows" $
         request methodGet  "/items?id=gt.0&id=lt.0" [singular] ""
-          `shouldRespondWith` 406
+          `shouldRespondWith` 404
 
       it "will select an existing object" $ do
         request methodGet "/items?id=eq.5" [singular] ""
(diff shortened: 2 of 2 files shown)


ruslantalpa:
i think this is correct. that error was meant for when th user thinks he is requesting one object but actually, the filers return more than one. It did not account for the missing case when requesting by ID

begriffs:
Actually the docs in this case are out of date. We used to use a custom header to request a singular response, which used the 404 response code when no rows were returned. Now that we use content negotiation for the response type, an HTTP 406 is the appropriate status code. This is because the empty results cannot be represented as a JSON object (application/vnd.pgrst.object+json). Also 406 will work fine with clients because it is a 4xx code and I think client libraries will still treat it as an error.

Contrast this content type error with the user requesting `GET /sdfjsdfjsdfj` which gives a 404 due to a totally missing route.

ruslantalpa:
I think this was closed prematurely.
What people really want is to do something like `GET /items/1` and the way to do this in postgrest is with this Accept header.

Now leaving postgrest aside for a minute, what is the default method for handling this rest call when the item is missing?
```
GET /items/1
Accept: application/json
```

i would say people expect a 404 and the response body does not matter, but it could be a `{}`

Why would it be wrong to do it like in this PR, maybe with the addition of replying with `{}` for the body to make it a valid json?

begriffs:
The thing is, it's strange to change an http code to 404 on a resource just because the client specified a different format. It is after all the same resource. Like it would be odd for postgrest to issue a 404 when the client requested `/foo` as csv, but 200 when `/foo` was requested as json. Same logic goes for requesting it as application/vnd.pgrst.object.

Also the client should be able to specify `Accept:  application/vnd.pgrst.object+json, application/json` to allow a fallback if the singular response is not available. That's just [content negotiation](https://www.w3.org/Protocols/rfc2616/rfc2616-sec12.html).

If you want to make a new route using nginx like `/foo/1` then nginx should set the client Accept header and remap the response 406 to 404. It's up to nginx to provide the illusion of a separate route.

ruslantalpa:
The fact that for the same request with simple json you get `200 []` but for single object you get `404 {}` makes sense for it to be a problem. 
Nginx indeed can provide the illusion, but for that it needs the count header (to see that it's 0/*) which I guess is available.
i guess i was too quick to ask Steve to work on this (sorry @steve-chavez for wasting a few hours on this :) )

PR #914, opened 2017-07-15
fix(doc): server-proxy-uri -- URI, not URL

petermetz:
There was this typo in the ./docker/postgrest.conf example file, hoping to spare others from having to chase this straight back to the source code.


begriffs:
Thank you! There was a similar typo in the postgrest Heroku buildpack that I fixed recently, so glad you caught this one.

PR #916, opened 2017-07-19
strip newline before Base64 decoding

eric-brechemier:
I applied the fix [suggested by @begriffs on May 30th](https://github.com/begriffs/postgrest/issues/877#issuecomment-304778972).

I wrote unit tests for Input/Output in shell script. Only parts of these tests pass with this PR, namely all the tests related to Base64-encoded secrets.


begriffs:
Wow, this is thorough work! I could learn a thing or two from your detailed style of commit messages.

How about adding your test runner to [circle.yml](https://github.com/begriffs/postgrest/blob/master/circle.yml#L16) too so it can run as part of CI?

test/io-tests/configs/secret-from-file.config
```diff
@@ -0,0 +1,10 @@
+db-uri = "postgres:///postgrest_test"
+db-schema = "test"
+db-anon-role = "postgrest_test_anonymous"
+db-pool = 1
+server-host = "*4"
+server-port = 49421
+
+# Read secret from a file: /dev/stdin (alias for standard input)
+jwt-secret = "@/dev/stdin"
```
begriffs: Would "-" work to indicate stdin, or is that just a bash-ism?

test/io-tests.sh
```diff
@@ -0,0 +1,98 @@
...
+# Unit Test Templates
+readSecretFromFile(){
+  case "$1" in
+    *.b64)
+      pgrConfig="base64-secret-from-file.config";;
+    *)
+      pgrConfig="secret-from-file.config";;
+  esac
+  pgrStartRead "./configs/$pgrConfig" "./secrets/$1"
+  sleep 1 # leave enough time for the server to start
```
begriffs: A faster way may be to poll and check whether postgrest has bound to its port:

```bash
while ! nc -z localhost 49421; do   
  sleep 0.1
done
```

test/io-tests/configs/secret-from-file.config
eric-brechemier: `-` is a convention which has to be implemented by the software (hint: PostgREST doesn't).
`/dev/stdin` is a symbolic link which points to the standard input of the current process; this is an OS feature.

test/io-tests.sh
eric-brechemier: Good idea. I also need to detect faster when PostgREST fails to start and never binds to the port.

eric-brechemier: Checking that the port is bound is not enough:

```
  while pgrStarted && ! nc -z localhost "$pgrPort" >/dev/null 2>&1
  do
    # wait for the server to start
    sleep 0.1 \
    || sleep 1 # fallback: subsecond sleep is not standard and may fail
  done
```

all tests fail with HTTP status `503` because PostgREST has not connected to the database yet.

eric-brechemier: I replaced `nc` with `curl` to `/`, waiting for a status `200`. It works and cuts the duration of the execution from 12+ seconds to 4+ seconds for 12 tests. Not bad.

eric-brechemier:
> How about adding your test runner to circle.yml too so it can run as part of CI?

Done.

begriffs:
This PR is also nice because it gives people another place to write a certain kind of test if they need to.

Add an entry to the changelog and I think this is good to merge.

eric-brechemier:
Thanks!

PR #918, opened 2017-07-22
Ability to embed using a specific relation when there are multiple between tables, fixes #907

### LINKED ISSUE #907: Resource Embedding Unexpected Behavior: table is being treated as pivot
```sql
-- users
create table data.users ( -- df: mult=3.0
  id serial primary key,
....
-- forms
create table data.forms ( -- df: mult=10.0
  id serial primary key,
  user_id integer references data.users(id) default request.user_id(),
....
-- pitches
create table data.pitches ( -- df: mult=100.0
  id serial primary key,
  user_id integer references data.users(id) default request.user_id(),
  form_id integer references data.forms(id) not null,
....
```
http://localhost:8080/rest/users?select=name,forms(name)
creates this query
```sql
SELECT  "api"."users"."name", COALESCE ((SELECT array_to_js on(array_agg(row_to_json("forms"))) 
FROM (SELECT  "api"."forms"."name" FROM  "api"."pitches", "api"."forms"  
WHERE  "api"."forms"."id" = "api"."pitches"."form_id" 
AND  "api"."users"."id" = "api"."pitches"."user_id"  ) "forms"), '[]') 
AS "forms" FROM  "api"."users"
```

### intended behavior:
users (one to many) forms
### recieved:
users (many to many through pitches) forms

ruslantalpa:
*Well, this bothered me more than it should :)*

When embedding parents we had the ability to say which FK to use by using it directly instead of the table name ( `billing_address_id`, `shipping_address_id` )

but when requesting children, there is no such possibility.
This PR adds the possibility to "lock" the path in the other direction like this
`/addresses?select=id,billed:orders.billing_address_id{id},shipped:orders.shipping_address_id{id}`
*the aliases before `:` are not mandatory when there is no collision*

In this case, the part after the `.` is the name of the FK

When there is a many to many relation, it's also possible to select it like so
`/tasks?select=id,users.users_tasks{id}`

In this case the part after the `.` is the name of the "link" table.

PS: I'd like to thank the Academy, Haskell language, and GHC's type-checker without which these types of PRs would not be possible :)

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -199,7 +199,7 @@ pgFmtLit x =
 
 requestToCountQuery :: Schema -> DbRequest -> SqlQuery
 requestToCountQuery _ (DbMutate _) = undefined
-requestToCountQuery schema (DbRead (Node (Select _ _ logicForest _ _, (mainTbl, _, _)) _)) =
+requestToCountQuery schema (DbRead (Node (Select _ _ logicForest _ _, (mainTbl, _, _, _)) _)) =
  unwords [
    "SELECT pg_catalog.count(*)",
    "FROM ", fromQi qi,
@@ -215,7 +215,7 @@ requestToCountQuery schema (DbRead (Node (Select _ _ logicForest _ _, (mainTbl, 
    filteredLogic = filter nonFKRoot logicForest
 
 requestToQuery :: Schema -> Bool -> DbRequest -> SqlQuery
-requestToQuery schema isParent (DbRead (Node (Select colSelects tbls logicForest ord range, (nodeName, maybeRelation, _)) forest)) =
+requestToQuery schema isParent (DbRead (Node (Select colSelects tbls logicForest ord range, (nodeName, maybeRelation, _, _)) forest)) =
   query
   where
     mainTbl = fromMaybe nodeName (tableName . relTable <$> maybeRelation)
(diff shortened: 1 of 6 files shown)


ruslantalpa:
Let's merge #885 first then i'll rebase. Those openAPI tests take a long time that that pr does a great job of fixing that.

begriffs:
Nice! What happens if somebody puts a nonexistent column name after the dot, or a column that is not part of an FK? Curious how friendly the error will be.

(Also I merged the openapi PR so ready for your rebase)

ruslantalpa:
the error will be could not find relation ...

ruslantalpa:
rebased, CI is running now

begriffs:
Nice feature, nice tests. 🚢  Looks good to me, so you can click squash and merge if you're happy with it.

Too bad it introduces more underscores in the pattern matches, but fixing that doesn't seem straightforward.

src/PostgREST/Types.hs
```diff
@@ -182,7 +182,8 @@ type Field = (FieldName, Maybe JsonPath)
 type Alias = Text
 type Cast = Text
 type NodeName = Text
-type SelectItem = (Field, Maybe Cast, Maybe Alias)
+type RelationDetail = Text
```
steve-chavez: This type is not clear to me, Relation is already a confusing name since PostgreSQL calls tables and views "relations". Could this name be more precise?

ruslantalpa:
By the time the logic gets to the QueryBuilder module, that last param i added is not used since the information is contained in the `Maybe Relation` type, so yeah it's not very nice, but i think i have an idea of how to refactor so that querybuilder module is not touched.

something like

```haskell
data Rel = R Relation | D RelationDetail
type ReadNode = (ReadQuery, (NodeName, Maybe Rel, Maybe Alias))
```

so that the middle component can hold the info either in `R` or `D` format since it kind of relates to the same thing.

@steve-chavez i'll think about the naming, my thinking was: this information is used in determining the `Relation` component so i called it `RelationDetail`.  Here relation is used in the sens of "what is the relation between this tree node and its parent"
While changing RelationDetail part is relatively easey, i would not go and change the name for Relation type since it's all over the place ...

steve-chavez:
@ruslantalpa The Relation type is definitely not easy to refactor, I was thinking that just a rename of ```RelationDetail``` could be made, into a more obvious type name, something like ```EmbedFK/FKCol/FKName/JoinColumn```, or maybe just leave the type with a comment.

src/PostgREST/QueryBuilder.hs
```diff
@@ -218,7 +218,11 @@ requestToQuery :: Schema -> Bool -> DbRequest -> SqlQuery
 requestToQuery schema isParent (DbRead (Node (Select colSelects tbls logicForest ord range, (nodeName, maybeRelation, _)) forest)) =
   query
   where
-    mainTbl = fromMaybe nodeName (tableName . relTable <$> maybeRelation)
+    -- data RelationInfo = R Relation | D RelationDetail;
+    rir (R r) = Just r
+    rir _ = Nothing
+
+    mainTbl = fromMaybe nodeName (tableName . relTable <$> join (rir <$> maybeRelation))
```
ruslantalpa: anyone has a better idea how to write this (without the rir function ...)?

ruslantalpa:
do the types look better now a bit compared to the previous version?

steve-chavez:
Though there are less wildcard pattern matches now there are unwrapping functions(`rir`, `r`) the former being more easy to understand, the added constructors `R` and `D` are not that meaningful and the QueryBuilder is still modified, overall I think the previous version was clearer, maybe it'll be better to just go back to the previous `ReadNode`(withouth `RelationInfo`) type and add a comment for the `RelationDetail` since is not easy to come with a self-explanatory name for it.

ruslantalpa:
@begriffs feel the same way? I'll reveret. Also not particularily liking these types

begriffs:
Yeah I do think the R and D stuff is pretty terse.

Also I agree with Steve's point that Relation is a pretty overloaded term. Perhaps `FkRelation` in the context of foreign keys would make it more explicit.

ruslantalpa:
OK, i'll revert but leave the Relation type name in places, it would be too much unrelated change for this RP

ruslantalpa:
reverted back

ruslantalpa:
about relationDetail, happy to rename just not sure what it should be.
It does not always hold the name of the FK, sometimes, in many to many it holds the name of the join table, that's why i named it RelationDetail.

any suggestions before i merge this? (adding a comment about the type now)

begriffs:
The new comments are helpful. I'm OK if you merge, unless Steve has any other thoughts.

steve-chavez:
I'm OK as well, the new comments are good enough.

PR #919, opened 2017-07-26
Support asymmetric JWK

### LINKED ISSUE #567: ~~Clarify supported algorithms for jwt~~  Adopt support for more jwt algorithms

begriffs:
Uses the [jose](https://hackage.haskell.org/package/jose) library to support signing JWTs with many [algorithms](https://hackage.haskell.org/package/jose-0.6.0.3/docs/Crypto-JOSE-JWS.html#t:Alg) including RSA.

The `jwt-secret` configuration parameter previously used to accept only an HMAC-SHA256 secret. Now it accepts any JSON Web Key (JWK). It's backward compatible in that if you provide a string it can't parse as JWK then it treats it as an HMAC-SHA secret.

To use PostgREST with asymmetric crypto you use the private key to sign the JWT, and put the public key inside the PostgREST config to verify the signature. I added a test case for a RS256 keypair. Note that the JWT spec requires the key size to be at least 2048 bits.

There is technically one breaking change: if you use a plain string secret this library requires it to be at least 32 characters long. Shorter secrets will cause the JWT validation to always fail. This length requirement is actually mandated by the JWT standard for proper security, but our previous Haskell library was lenient and didn't require it. So for people using PostgREST with a shorter secret, they would need to change the secret prior to upgrading.

Also note that jose supports validating with more than one key, so this PR paves the way for passing a json array in the `jwt-secret` parameter as desired in #505 .

Fixes #567


begriffs:
@eric-brechemier wondering if you can help me out -- the secrets used in your io-tests are too short for the new jose library. Could you send a pull request to lengthen them? I can then rebase on it.

ruslantalpa:
So i am looking at the fact that the `exp` validation is done internally in the jwt lib
and it starts with this line
https://github.com/frasertweedale/hs-jose/blob/master/src/Crypto/JWT.hs#L437
The definition of the currentTime is actually `ask` from inside the `MonadTime ` which i think is this

https://hackage.haskell.org/package/monad-time-0.2/docs/src/Control-Monad-Time.html#currentTime

so in the end, what gets called is Data.Time.getCurrentTime

and the type of the function also tells us there is IO
https://github.com/begriffs/postgrest/blob/6fe32f1bdc5cd2c733bbe9f7bcd54aa359724dad/src/PostgREST/Auth.hs#L45
`jwtClaims :: Maybe JWK -> BL.ByteString -> IO JWTAttempt`

so what i am getting at is that or each JWT validation, each request, we are doing a syscall with `Data.Time.getCurrentTime` which is an expensive call.

Merging it this way big regression in speed.
It's too many monads for me to be comforatable with so i don't know where exactly this should/can be fixed, postgrest code, or JOSE but this (doing expensive sys calls on every request) does not work well in webserver settings

ruslantalpa:
it also seems that every other validation function is making a call to currentTime and i don't think that is being cached somewhere, it will go all the way down to the OS.

begriffs:
I remember when we pulled the time check out so that it wouldn't get called on every request, but do you remember if checking the time adds significant delay?

ruslantalpa:
it might seem small in human terms, but when you want high performance (thousands of rps) those things add up.  I specifically took that out after reading a blog post from warp describing this specific thing (doing get time + formatting it) as a reason for low performance and a thing they had to optimize.

eric-brechemier:
> @eric-brechemier wondering if you can help me out -- the secrets used in your io-tests are too short for the new jose library. Could you send a pull request to lengthen them? I can then rebase on it.

@begriffs sure, can you give me more details?

begriffs:
Oops, yeah. 😄 

The secret strings need to be at least 32 bits, so some of the unicode ones are squeaking by, but most are too short. When the new JWT library tries to validate a token signed with a short string, it treats it as a validation failure.

eric-brechemier:
@begriffs done in d824d81 in https://github.com/eGullGolf/postgrest/tree/issue919

diogob:
@begriffs @ruslantalpa I'm also sad to see auto-update go.
It seems to me that jose already does the job of using a polymorphic monad so we can provide a custom instance instead of using IO. 

Something like the [wai-transformers](https://hackage.haskell.org/package/wai-transformers) package might make it easier.

begriffs:
@diogob interesting idea, do you have the time/desire to look into it and paste a code snippet of how it might work?

diogob:
@begriffs I'd love to. I'll try to write something.

begriffs:
OK, as we discussed in the chat room, the getCurrentTime is probably not that much overhead, and we'll merge this as is.

The jose library is going to add a new validation function for us where we can pass in the time ourselves, so if the current way turns out to be too slow we can more easily fix it later.

PR #923, opened 2017-07-31
Add autorelease for linux distros, osx and windows

steve-chavez:
This PR generates releases for the linux distros specified in the Release Checklist wiki(CentOS 6/7, Ubuntu 16 x64/x32), it does this by using docker in CircleCI, a result of this can be seen in https://github.com/steve-chavez/postgrest/releases/tag/v0.4.3.0 (after bumping version and pushing tag).

Still missing autoreleases for windows(I'm running into a compile issue in AppVeyor) and freebsd/osx(was planning to do those on travis-ci), but for now we can discuss if the approach is good for the linux distros.

The building of the docker images and the setup of stack for each one of those will make the first CircleCI build really slow(couple hours maybe),  but I've cached the docker images(circleci doesn't do it natively) in a `~/docker` dir and also there's a `~/.stack` dir for each one of the distros so after the setup the build times will be just a little more slow(I've seen that restoring the cache takes like 4mins now).


steve-chavez:
A GITHUB_TOKEN env var in CircleCI is required for the autorelease to work.

ruslantalpa:
Checked out the build scripts, very nice setup. CHANGELOG grep also nice.
So far so good :)
Once the win/mac is ready and this is merged, we should try a release :)

ruslantalpa:
@steve-chavez FYI, if you switch to CIrcleCI 2, you have the option to have parallel jobs,
the build for each OS can run in a separate job and at the end you can run ghr to publish all the artifacts.

and example https://github.com/subzerocloud/postgrest-starter-kit/blob/cicrleci/.circleci/config.yml

the biggest change to be aware in circle2 is the executor type for each job https://circleci.com/docs/2.0/executor-types/

In my example i mostly use `machine` executor but i could have done without it and use https://circleci.com/docs/2.0/building-docker-images/ , except for the `test` job where i need `docker-compose`

ruslantalpa:
For windows, maybe this can serve as a guide https://github.com/commercialhaskell/intero/blob/master/appveyor.yml

ruslantalpa:
@begriffs the `GITHUB_TOKEN` needed by ghr is basically a password for your account and even if it would not be visible to ppl forking the repo (a setting in CircleCI), it will be visible to contributors (me/steve) so the best solution for this I think is to create a distinct github account, make it a collaborator on postgrest and use its token in CircleCI.

Will that work?

begriffs:
This is really cool! I created @postgrest-bot who is a collaborator on this project and whose token we can use in CircleCI. What should I do next to hook this up for you?

ruslantalpa:
just tell Steve the password of the bot in chat :)

ruslantalpa:
@begriffs and since we've got you here, we can also have mac building for free
https://circleci.com/pricing/#faq-section-os-x
```
We also offer the Seed plan for OS X open-source projects.
Contact us at billing@circleci.com for access. 
If you are building a bigger open-source project and need more resources, 
let us know how we can help you!
```
all we have to do is ask :)

begriffs:
☑️ Emailed them about the OS X seed plan.

ruslantalpa:
Another feature of dokcer/circleci that we should leverage is multistage builds
https://circleci.com/blog/multi-stage-docker-builds/
i.e. the same docker file containse the steps to build the binary, then the steps to produce lightweight the runtime image that uses that binary (no need to curl from release page).

This way, we can have docker files for each distro that ppl can produce for themselves the "master branch" docker image in one easy step instead of waiting for a release to be able to build the docker image.

begriffs:
(By the way, Circle replied to me saying they enabled the OS X seed plan for us)

steve-chavez:
Added appveyor.yml for windows, it'll only build on tags and won't test, I think is fine since we already test on circle.

steve-chavez:
Does anyone have a guide/example on how to setup a mixed Linux and OS X build on CircleCI?
Been able to do something like this in `circle.yml`:
```yaml
os:
  - linux
  - osx
```

I've seen that as of CircleCI 2.0 this isn't supported(https://discuss.circleci.com/t/how-to-use-circleci-for-linux-xcode-in-one-project/693/4) and they recommend having a separate repo for OS X builds; I don't think that will be a good solution.

Also I don't understand if the seed plan enables that use case and if it does how to configure it in a `circle.yml`, so for now I'll be going with Travis CI just for building the OS X release.

steve-chavez:
The OS X release for Travis has an issue, there's a job timeout at around 50 mins and the stack setup with dependencies takes longer than that.
 
The workaround for that is to [timeout](https://github.com/begriffs/postgrest/pull/923/commits/52e0c06665370427a5e28d4b2068f1266914e208#diff-354f30a63fb0907d4ad57269548329e3R43) the stack setup for the job to exit successfully(builds and releases only done on tags). After that since stack already installed some dependencies and those are cached the job can be restarted manually from Travis and that way complete installing and caching all the dependencies.

That workaround will only have to be done once, after that when pushing a tag it builds and releases in about 7 mins.

steve-chavez:
I think the FreeBSD would be better done manually, Travis doesn't support FreeBSD(previously saw 
 a [guide](https://erouault.blogspot.pe/2016/09/running-freebsd-in-travis-ci.html) that run it through QEMU but doesn't seem feasible to do the release that way), so I'll be installing a local FreeBSD and attach the release manually after the other releases are completed.

For now the osx, linux distros and windows releases are working(see this [example](https://github.com/steve-chavez/postgrest/releases/tag/v0.4.3.0)), and though the migration to Circle 2.0 is missing I think we could merge this as it is now and see if the setup is good enough(circle distro releases are taking like 10 mins after setup is done)  if not the migration to Circle 2.0 could be done in another PR.

Also I've added `GITHUB_TOKEN` env vars from @postgrest-bot to both Circle and Travis, only missing Appveyor, @begriffs maybe you can create an account there to add the last token.

begriffs:
I created an Appveyor account and will send you the API token privately.

.travis.yml
```diff
@@ -0,0 +1,63 @@
...
+  if test ! "$TRAVIS_TAG"
+  then
+    echo 'No tag pushed. Skipping release.'
+  else
+    OWNER="$(echo "$TRAVIS_REPO_SLUG" | cut -f1 -d/)"
+    REPO="$(echo "$TRAVIS_REPO_SLUG" | cut -f2 -d/)"
+    START=$(echo $TRAVIS_TAG | cut -c2-)
+    END='## \['
+    BODY=$(sed -n "1,/$START/d;/$END/q;p" CHANGELOG.md)
+    tar cjf postgrest-$TRAVIS_TAG-osx.tar.xz postgrest
```
ruslantalpa: you need to run the `strip` command here to make the binary smaller before zipping it

steve-chavez:
Moved to CircleCI v2, the build and test of previous v1 version somehow is faster(takes like 4 mins after cached dependencies), the release step is also faster(5 mins with cache) since the distros binaries are built in parallel.

.circleci/config.yml
```diff
@@ -0,0 +1,178 @@
...
+          at: /tmp/workspace
+      - checkout
+      - run:
+          name: add body and tars to github release
+          command: |
+            go get -u github.com/tcnksm/ghr
+            START=$(echo $CIRCLE_TAG | cut -c2-)
+            END='## \['
+            BODY=$(sed -n "1,/$START/d;/$END/q;p" CHANGELOG.md)
+            ghr -t $GITHUB_TOKEN -u $CIRCLE_PROJECT_USERNAME -r $CIRCLE_PROJECT_REPONAME -b "$BODY" --replace v0.4.3.0 /tmp/workspace/bin
```
ruslantalpa: after this command, you could add the docker build step, quite easy.
change this var from hardcoded to an `ARG`
https://github.com/begriffs/postgrest/blob/master/docker/Dockerfile#L10

then you can do this
```
docker build --build-arg POSTGREST_VERSION=CIRCLE_TAG -t postgrest ./docker/
docker tag postgrest postgrest/postgrest:$CIRCLE_TAG
docker push postgrest/postgrest:$CIRCLE_TAG
docker tag postgrest postgrest/postgrest:latest
docker push postgrest/postgrest:latest
```

steve-chavez: I would need docker hub credentials to do that, @begriffs does the postgrest organization exists? If it does maybe we could add @postgrest-bot as a member and then use his docker hub credentials as env vars in circle to be able to push the image.

ruslantalpa: it exists, https://hub.docker.com/r/postgrest/postgrest/

steve-chavez:
Added docker image autorelease, some example results can be seen in https://hub.docker.com/r/stevechavez/postgrest/tags/.

As I mentioned before in https://github.com/begriffs/postgrest/pull/923#discussion_r136697479 we need docker hub credentials for this to work and then set [this](https://github.com/steve-chavez/postgrest/blob/linux-auto-release/.circleci/config.yml#L134) env vars on CircleCI.

begriffs:
I just created a "postgrestbot" user on dockerhub (they don't allow dashes in the name) and linked him to the github account of "postgrest-bot." I also added postgrestbot to the collaborators list on the postgrest organization, and gave admin rights to collaborators. (I could turn that down to write access if admin is more than we need.)

So all I need to do is set `$DOCKER_USER` and `$DOCKER_PASS` on Circle?

steve-chavez:
Yes, that would be all is needed for pushing the image to docker hub.

ruslantalpa:
merge and release! (if anything goes wrong it can be deleted like it never happened :) )

steve-chavez:
Aside from the docker hub credentials setup I think this PR is ready to merge.

begriffs:
Cool, I'm coming back from a trip now but I'll review this tonight just to get familiar with how it works and then merge it.

begriffs:
OK, CircleCI env vars are in place. I looked over the changes in this PR and here we go!

Issue #924, opened 2017-07-31
Slow relations initialization


9too:
# Context
PostgREST version : 0.4.2.0
PostgreSQL : 9.6.3
OS : Gentoo & RHEL 6.9

The source of the problem is unclear and I cannot provide a simple reproducible case.
( I provided @ruslantalpa with a _minimal_ database dump )

# Observation
On a particular database, requesting resource embedding for the first time causes a **significant** computation over a long period of time. On the sample provided, it takes ~30sec on my box and PostgREST was using ~40% of CPU load the whole time. Although that was after a cleanup, before I simply didn't wait long enough (> 10min). The query itself seems unimportant, as asking for an invalid relation triggers this problem. Hence, the exposed schema to test this simply contains one table such as
```sql
create table my_api.test ( col int );
```
And the problem happens when issuing a query like
```sh
curl 'http://localhost:3000/test?select=*,test(*)'
```

## Debugging
Everything involves the whole database and does _not_ involve the schema exposed to PostgREST.
- Trying to scope down the cause, it was found that :
   - Dropping one specific table completely removed the delay, albeit the table is nothing special at first glance
   - Dropping a couple views in a particular schema improves the delay gradually
- Requesting for a single entity while PostgREST is under this computation works without any delay.

9too:
With a bit more troubleshooting and a dose of luck, I have found the potential culprit.
The table which caused the problem to disapear contains 2 foreign keys which are exactly the same reference, they only have different names. Of course this is an error and shouldn't be there, but by dropping any of the two, the problem goes away.
Now I tried recreating a simple case with only 2 tables, but it does not cause any noticeable delay.
My guess is that it probably starts to mess up when there are a certain amount of references to be analyzed.

ruslantalpa:
Interesting case, although i think there are valid cases for 2 Fks to the same table so this still is interesting to investigate.

9too:
After more testing it seems my assumption was incorrect, the problem is likely related but not exactly caused by identical FK duplicates.

I could find out by diffing a local database that I had fixed doing this with a preproduction version that was slightly altered from my version, where the problem still showed up after applying the above change.

In the diffing process, I found a FK that after being dropped, fixed the problem again. But this is a valid FK constraint that applies on a different column of the constrained table, but to the same reference.
The FK is on a character(8), that's the only particularity I can see.

Again I'm not able to reproduce with a minimal case ...

[...]

PR #926, opened 2017-08-02
Improve relations initialization time (#924)

9too:
Reduce computation required to generate relations on the first request.

## Sample test case

**PostgREST v0.4.2.0**
```
$ time curl 'http://localhost:3000/x?select=x(*)'

real	0m27.601s
user	0m0.001s
sys	0m0.001s
```
**PostgREST with combinations improvement**
```
$ time curl 'http://localhost:3000/x?select=x(*)'

real	0m0.012s
user	0m0.001s
sys	0m0.001s
```


steve-chavez:
The performance improvement is great, though I can't really say I understand the `combinations` function easily, how about adding the reference implementation in a comment like: 
```haskell
-- From https://wiki.haskell.org/99_questions/Solutions/26
combinations :: ...
```

9too:
I must agree on this, I will add the reference.

begriffs:
Thanks @9too, good work tracing it back to this function. What a difference a good algorithm can make...

PR #929, opened 2017-08-06
Treat blank pre-request config as missing

### LINKED ISSUE #927: Empty pre-request value should indicate that PostgREST should not execute any procedure
The way ```pre-request``` configuration variable is being set up in ```Dockerfile``` (```begriffs/postgrest:v4.2.0```) finally leaves everyone with non-functioning PostgREST container if ```PGRST_PRE_REQUEST``` environment variable is left as it is (empty) because PostgREST is trying to execute statement ```SELECT ()``` right before any query.

begriffs:
Fixes #927


begriffs:
@zalmarge sorry for the insufficient testing that allowed this issue.

Issue #930, opened 2017-08-08
Fill in openapi 'summary' field


daurnimator:
Currently the comment on a table is used for the 'description' but the 'summary' field is left blank.
https://github.com/begriffs/postgrest/blob/2243f4b6539577a6534e11bee66052ae5cda7e71/src/PostgREST/OpenAPI.hs#L188

I propose using the *first* line of the comment as the 'summary', and anything following as the description (sort of like a git commit).

daurnimator:
This should happen for columns (as well as tables/view): if there is no summary, then swagger-ui prefills `<input>` boxes with the whole column description.

PR #931, opened 2017-08-08
OpenAPI: Split table comment into summary and description

daurnimator:
For #930 

I don't know haskell, so be gentle :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #883, Binary output support for RPC - @steve-chavez
 - #885, Postgres COMMENTs on SCHEMA/TABLE/COLUMN are used for OpenAPI - @ldesgoui
 - #907, Ability to embed using a specific relation when there are multiple between tables - @ruslantalpa
+- #930, Split table comment on newline to get OpenAPI operation summary and description - @daurnimator
 
 ### Fixed
 
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -6,17 +6,18 @@ module PostgREST.OpenAPI (
   , pickProxy
   ) where
 
+import           Control.Arrow               ((&&&))
 import           Control.Lens
 import           Data.Aeson                  (decode, encode)
 import           Data.HashMap.Strict.InsOrd  (InsOrdHashMap, fromList)
 import           Data.Maybe                  (fromJust)
 import qualified Data.Set                    as Set
 import           Data.String                 (IsString (..))
-import           Data.Text                   (unpack, pack, init, tail, toLower, intercalate, append)
+import           Data.Text                   (unpack, pack, init, tail, toLower, intercalate, append, dropWhile, breakOn)
 import           Network.URI                 (parseURI, isAbsoluteURI,
                                               URI (..), URIAuth (..))
 
-import           Protolude hiding              ((&), Proxy, get, intercalate)
+import           Protolude hiding              ((&), Proxy, get, intercalate, dropWhile)
 
 import           Data.Swagger
 
@@ -183,9 +184,14 @@ makeRowFilters tn = map (makeRowFilter tn)
 makePathItem :: (Table, [Column], [Text]) -> (FilePath, PathItem)
 makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
+    -- Use first line of table description as summary; rest as description (if present)
+    -- We strip leading newlines from description so that users can include a blank line between summary and description
+    (tSum, tDesc) = fmap fst &&& fmap (dropWhile (=='\n') . snd) $
+                    breakOn "\n" <$> tableDescription t
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      & description .~ tableDescription t
+      & summary .~ tSum
+      & description .~ mfilter (/="") tDesc
     getOp = tOp
       & parameters .~ map ref (rs <> ["select", "order", "range", "rangeUnit", "offset", "limit", "preferCount"])
       & at 206 ?~ "Partial Content"
--- a/test/Feature/StructureSpec.hs
+++ b/test/Feature/StructureSpec.hs
@@ -33,12 +33,26 @@ spec = do
         r <- simpleBody <$> get "/"
 
         let method s = key "paths" . key "/child_entities" . key s
+            childGetSummary = r ^? method "get" . key "summary"
+            childGetDescription = r ^? method "get" . key "description"
             getParameters = r ^? method "get" . key "parameters"
             postResponse = r ^? method "post" . key "responses" . key "201" . key "description"
             patchResponse = r ^? method "patch" . key "responses" . key "204" . key "description"
             deleteResponse = r ^? method "delete" . key "responses" . key "204" . key "description"
 
+        let grandChildGet s = key "paths" . key "/grandchild_entities" . key "get" . key s
+            grandChildGetSummary = r ^? grandChildGet "summary"
+            grandChildGetDescription = r ^? grandChildGet "description"
+
         liftIO $ do
+
+          childGetSummary `shouldBe` Just "child_entities comment"
+
+          childGetDescription `shouldBe` Nothing
+
+          grandChildGetSummary `shouldBe` Just "grandchild_entities summary"
+
+          grandChildGetDescription `shouldBe` Just "grandchild_entities description\nthat spans\nmultiple lines"
 
           getParameters `shouldBe` Just
             [aesonQQ|
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -1221,6 +1221,13 @@ comment on table child_entities is 'child_entities comment';
 comment on column child_entities.id is 'child_entities id comment';
 comment on column child_entities.name is 'child_entities name comment';
 
+comment on table grandchild_entities is
+$$grandchild_entities summary
+
+grandchild_entities description
+that spans
+multiple lines$$;
+
 -- Used for testing that having the same return column name as the proc name
 -- doesn't conflict with the required output, details in #901
 create function test.test() returns table(test text, value int) as $$


src/PostgREST/OpenAPI.hs
```diff
@@ -185,7 +185,8 @@ makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      & description .~ tableDescription t
+      & summary .~ (strip . takeWhile (/= '\n') <$> tableDescription t)
+      & description .~ (strip . dropWhile (/= '\n') <$> tableDescription t)
```
daurnimator: How can we make this not render description at all if the string is empty?

src/PostgREST/OpenAPI.hs
```diff
@@ -185,7 +185,8 @@ makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      & description .~ tableDescription t
+      & summary .~ (strip . takeWhile (/= '\n') <$> tableDescription t)
```
daurnimator: Not sure if we should `strip` or not

src/PostgREST/OpenAPI.hs
begriffs: You can transform `Just ""` to `Nothing` using `mfilter (/="")`

src/PostgREST/OpenAPI.hs
begriffs: Probably doesn't hurt to strip off whitespace on the ends.

daurnimator:
Add `mfilter` to description.

src/PostgREST/OpenAPI.hs
```diff
@@ -185,7 +185,9 @@ makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      & description .~ tableDescription t
+      -- Use first line of table description as summary; rest as description (if present)
+      & summary .~ (strip . takeWhile (/="\n") <$> tableDescription t)
```
begriffs: Try single quoted, `'\n'`, which indicates a single character rather than a string.

src/PostgREST/OpenAPI.hs
```diff
@@ -185,7 +185,9 @@ makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
...
+      & description .~ mfilter (/="") (strip . dropWhile (/="\n") <$> tableDescription t)
```
begriffs: Also try single quoted \n here.

Also double check that strip eats a leading newline in the description since I think the dropWhile would preserve it.

daurnimator: yes the strip does.

src/PostgREST/OpenAPI.hs
```diff
@@ -185,7 +185,9 @@ makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      & description .~ tableDescription t
+      -- Use first line of table description as summary; rest as description (if present)
+      & summary .~ (strip . takeWhile (/='\n') <$> tableDescription t)
```
begriffs: You know, thinking about this again, it might be more efficient to grab both parts of the string in one `breakOn` operation, and then use them later. Like in the `where` clause add a new item:

```haskell
where
  (tSum, tDesc) = fmap fst &&& fmap snd $
                    breakOn "\n" <$> tableDescription t
  -- . . .
  summary .~ mfilter (/="") tSum
  description .~ mfilter (/="") tDesc
```

The weird [&&&](https://hackage.haskell.org/package/base-4.10.0.0/docs/Control-Arrow.html#v:-38--38--38-) thing is letting us move the Maybe inside the tuple.

This may be more efficient than two scans to takeWhile and dropWhile. Do you think the crazy syntax is worth it?

daurnimator: It does sound like a more suitable operator. Don't forget we also need to trip that `\n` off the start of `tDesc` too.

I tried implementing this but failed: I just don't know enough haskell. My attempt: http://sprunge.us/ZSce

begriffs:
> It does sound like a more suitable operator. Don't forget we also need to trip that \n off the start of tDesc too.

I see what you mean:

```haskell
snd $ breakOn "\n" "Hello\nworld"
--> "\nworld"

strip . snd $ breakOn "\n" "Hello\nworld"
--> "world"
```

So we'll want to strip the description.

> I tried implementing this but failed: I just don't know enough haskell. My attempt: http://sprunge.us/ZSce

No problem, you pretty much almost had it. Just need to import the `(&&&)` operator from Control.Arrow. Here's a patch that works locally for me:

```diff
diff --git a/src/PostgREST/OpenAPI.hs b/src/PostgREST/OpenAPI.hs
index a3a3925..5a1bb4e 100644
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -6,13 +6,14 @@ module PostgREST.OpenAPI (
   , pickProxy
   ) where
 
+import           Control.Arrow               ((&&&))
 import           Control.Lens
 import           Data.Aeson                  (decode, encode)
 import           Data.HashMap.Strict.InsOrd  (InsOrdHashMap, fromList)
 import           Data.Maybe                  (fromJust)
 import qualified Data.Set                    as Set
 import           Data.String                 (IsString (..))
-import           Data.Text                   (unpack, pack, init, tail, toLower, intercalate, append, takeWhile, dropWhile, strip)
+import           Data.Text                   (unpack, pack, init, tail, toLower, intercalate, append, breakOn, strip)
 import           Network.URI                 (parseURI, isAbsoluteURI,
                                               URI (..), URIAuth (..))
 
@@ -183,11 +184,14 @@ makeRowFilters tn = map (makeRowFilter tn)
 makePathItem :: (Table, [Column], [Text]) -> (FilePath, PathItem)
 makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
+    -- Use first line of table description as summary;
+    -- rest as description (if present)
+    (tSum, tDesc) = fmap fst &&& fmap (strip . snd) $
+      breakOn "\n" <$> tableDescription t
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      -- Use first line of table description as summary; rest as description (if present)
-      & summary .~ (strip . takeWhile (/='\n') <$> tableDescription t)
-      & description .~ mfilter (/="") (strip . dropWhile (/='\n') <$> tableDescription t)
+      & summary .~ mfilter (/="") tSum
+      & description .~ mfilter (/="") tDesc
     getOp = tOp
       & parameters .~ map ref (rs <> ["select", "order", "range", "rangeUnit", "offset", "limit", "preferCount"])
       & at 206 ?~ "Partial Content"
```

daurnimator:
@begriffs pushed.

src/PostgREST/OpenAPI.hs
```diff
@@ -183,9 +184,14 @@ makeRowFilters tn = map (makeRowFilter tn)
 makePathItem :: (Table, [Column], [Text]) -> (FilePath, PathItem)
 makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
   where
+    -- Use first line of table description as summary; rest as description (if present)
+    -- We strip leading newlines from description so that users can include a blank line between summary and description
+    (tSum, tDesc) = fmap fst &&& fmap (dropWhile (/='\n') . snd) $
```
begriffs: I'm confused by `dropWhile (/='\n')`, we want to remove non-newline characters? Would `strip` be better instead?

src/PostgREST/OpenAPI.hs
```diff
@@ -183,9 +184,14 @@ makeRowFilters tn = map (makeRowFilter tn)
...
+                    breakOn "\n" <$> tableDescription t
     tOp = (mempty :: Operation)
       & tags .~ Set.fromList [tn]
-      & description .~ tableDescription t
+      & summary .~ tSum
```
begriffs: If `tSum` were a blank string would we want set summary, or would a Nothing value be more appropriate?

daurnimator: I think nothing would be more appropriate: if no comment is set, emit no summary.

daurnimator: But we don't want the empty string to be mapped to nothing

src/PostgREST/OpenAPI.hs
daurnimator: Fixed.

src/PostgREST/OpenAPI.hs
begriffs: Hmm, I'm confused. 😕

> if no comment is set, emit no summary

I don't know whether this is true, but I was assuming that the way to do this would be to make the summary a Nothing value, which would cause it to be omitted in the OpenAPI output. Whereas a present value of the empty string for summary would it to be present in the OpenAPI output but empty.

How about adding test cases just to verify that this is causing the intended behavior?

daurnimator: "no comment" is a different case to "comment is the empty string".

begriffs: I see, so in case somebody adds a SQL comment to the object, and that comment is the empty string, then we'll preserve it as such. Even though it's a little weird, it's what the person specified in their SQL.

daurnimator: > Even though it's a little weird, it's what the person specified in their SQL.

Exactly.

daurnimator:
@begriffs  @postgrest-bot  test added.

begriffs:
Great work! Add an entry to the changelog and this should be good to merge.

daurnimator:
Done.

begriffs:
I don't see it yet, maybe it's not pushed.

daurnimator:
Oops sorry; pushed to wrong branch. fixed/pushed now.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- #930, Split table comment on newline to get OpenAPI operation summary and description - @daurnimator
```
begriffs: This should go under "Added" or "Fixed" depending on how you classify the change.

daurnimator: fixed,

Issue #539, opened 2016-03-30
Just suggestions about matching


Sigura:
I think it's will be wonderful to have support regular [expression matching](http://www.postgresql.org/docs/9.5/static/functions-matching.html). May be with `~` operator.

Then request url will looks like: `/table?search_field=~.expression`.

[+] at this time I need to make a lot of views in the ptoject for add `search` field. Thas is why I think - will be wonderful to support `or` operator in url.
May be `/table?or.first_name=~.^Josh&or.last_name=~.^Josh`.

Sigura:
PS: And I think you and comunity make a wondefult work. This project really expired me! Anyway THANK YOU!

begriffs:
Sure we can think about adding a regex filter. Also the `OR` condition has been requested fairly often, we should find a way to add it.

For now try these two steps, they may allow you to do what you need:

1. You can create a computed column that is used for filtering (it won't appear in the table results)

   ```sql
   CREATE FUNCTION search_field(table) RETURNS text AS $$
       SELECT $1.first_name || ' ' || $1.last_name;
   $$ LANGUAGE SQL;
   ```

2. Use the `like` or `ilike` matchers to find a substring in the computed column

   ```
   /table?search_field=like.*Josh*
   ```

Sigura:
Sure ,)

You described chosen solution for `or`  ,) But for regex it was just a sample to figure out my suggestion, my case more advanced ,)

And my suggestion is just way to avoid extra code to support additional `views`, `functions`, `gants` and etc ,)

Anyway thank you!

steve-chavez:
I would like to implement the regex filter feature but I think is not safe because of [ReDoS](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS).
For example this statement:

```sql
  select 'a' ~ '($^)+';
  # 'a' can be any string
```

Makes postgres version 9.4.5 to hang in an infinite loop, this appears to be fixed in 9.5 though.
I'm no postgres regex expert but it seems that just exposing regex expressions like that in the querystring gives a dangerous attack vector.

If this feature is worth it an option would be to do a ```set local statement_timeout``` for each transaction to mitigate the risk, the timeout value could be specified in postgREST config file and that maybe would allow it to provide more costly sql operations like ```group by```, etc.

begriffs:
Thanks, I didn't know about this attack.

For the `statement_timeout` as well as other local vars we could have it read a `locals` section from the config like

```
locals {
[...]

PR #938, opened 2017-08-12
Add support for range operators

russelldavies:
- [x] Update docs

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -417,7 +417,9 @@ pgFmtFilter table (Filter fld (Operation hasNot_ ex)) = notOp <> " " <> case ex 
    (op, VText val) -> pgFmtFieldOp op <> " " <> case op of
      "like"  -> unknownLiteral (T.map star val)
      "ilike" -> unknownLiteral (T.map star val)
+     -- TODO: The '@@' was deprecated, remove in v0.5.0.0
      "@@"    -> "to_tsquery(" <> unknownLiteral val <> ") "
+     "fts"   -> "to_tsquery(" <> unknownLiteral val <> ") "
      "is"    -> whiteList val
      "isnot" -> whiteList val
      _       -> unknownLiteral val
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -163,7 +163,15 @@ operators = M.fromList [
   ("is", "IS"),
   ("@@", "@@"),
   ("@>", "@>"),
-  ("<@", "<@")]
+  ("<@", "<@"),
+  ("cs", "@>"),
+  ("cd", "<@"),
+  ("ov", "&&"),
+  ("sl", "<<"),
+  ("sr", ">>"),
+  ("nxr", "&<"),
+  ("nxl", "&>"),
+  ("adj", "-|-")]
 data Operation = Operation{ hasNot::Bool, expr::(Operator, Operand) } deriving (Eq, Show)
 data Operand = VText Text | VTextL [Text] | VForeignKey QualifiedIdentifier ForeignKey deriving (Show, Eq)
 
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -161,9 +161,7 @@ operators = M.fromList [
   ("notin", "NOT IN"),
   ("isnot", "IS NOT"),
   ("is", "IS"),
-  ("@@", "@@"),
-  ("@>", "@>"),
-  ("<@", "<@"),
+  ("fts", "@@"),
   ("cs", "@>"),
   ("cd", "<@"),
   ("ov", "&&"),
@@ -171,7 +169,11 @@ operators = M.fromList [
   ("sr", ">>"),
   ("nxr", "&<"),
   ("nxl", "&>"),
-  ("adj", "-|-")]
+  ("adj", "-|-"),
+  -- TODO: these are deprecated and should be removed in v0.5.0.0
+  ("@@", "@@"),
+  ("@>", "@>"),
+  ("<@", "<@")]
 data Operation = Operation{ hasNot::Bool, expr::(Operator, Operand) } deriving (Eq, Show)
 data Operand = VText Text | VTextL [Text] | VForeignKey QualifiedIdentifier ForeignKey deriving (Show, Eq)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #924, Improve relations initialization time - @9too
 - #927, Treat blank pre-request as missing - @begriffs
 
+### Changed
+
+- #938, Deprecate symbol operators with mnemonic names. - @russelldavies
+
 ## [0.4.2.0] - 2017-06-11
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #885, Postgres COMMENTs on SCHEMA/TABLE/COLUMN are used for OpenAPI - @ldesgoui
 - #907, Ability to embed using a specific relation when there are multiple between tables - @ruslantalpa
 - #930, Split table comment on newline to get OpenAPI operation summary and description - @daurnimator
+- #938, Support for range operators - @russelldavies
 
 ### Fixed
 
--- a/test/Feature/AndOrParamsSpec.hs
+++ b/test/Feature/AndOrParamsSpec.hs
@@ -76,6 +76,37 @@ spec =
         it "can handle @> and <@" $
           get "/entities?or=(arr.@>.{1,2,3},arr.<@.{1})&select=id" `shouldRespondWith`
             [json|[{ "id": 1 },{ "id": 3 }]|] { matchHeaders = [matchContentTypeJson] }
+
+        it "can handle range operators" $ do
+          get "/ranges?range=eq.[1,3]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=neq.[1,3]&select=id" `shouldRespondWith`
+            [json|[{ "id": 2 }, { "id": 3 }, { "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=lt.[1,10]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=gt.[8,11]&select=id" `shouldRespondWith`
+            [json|[{ "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=lte.[1,3]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=gte.[2,3]&select=id" `shouldRespondWith`
+            [json|[{ "id": 2 }, { "id": 3 }, { "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=cs.[1,2]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=cd.[1,6]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }, { "id": 2 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=ov.[0,4]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }, { "id": 2 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=sl.[9,10]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }, { "id": 2 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=sr.[3,4]&select=id" `shouldRespondWith`
+            [json|[{ "id": 3 }, { "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=nxr.[4,7]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }, { "id": 2 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=nxl.[4,7]&select=id" `shouldRespondWith`
+            [json|[{ "id": 3 }, { "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
+          get "/ranges?range=adj.(3,10]&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }]|] { matchHeaders = [matchContentTypeJson] }
+
         context "operators with not" $ do
           it "eq, @>, like can be negated" $
             get "/entities?and=(arr.not.@>.{1,2,3},and(id.not.eq.2,name.not.like.*3))&select=id" `shouldRespondWith`
--- a/test/Feature/AndOrParamsSpec.hs
+++ b/test/Feature/AndOrParamsSpec.hs
@@ -70,10 +70,16 @@ spec =
         it "can handle is" $
           get "/entities?and=(name.is.null,arr.is.null)&select=id" `shouldRespondWith`
             [json|[{ "id": 4 }]|] { matchHeaders = [matchContentTypeJson] }
-        it "can handle @@" $
+        it "can handle fts" $ do
+          get "/entities?or=(text_search_vector.fts.bar,text_search_vector.fts.baz)&select=id" `shouldRespondWith`
+            [json|[{ "id": 1 }, { "id": 2 }]|] { matchHeaders = [matchContentTypeJson] }
+          -- TODO: remove in 0.5.0 as deprecated
           get "/entities?or=(text_search_vector.@@.bar,text_search_vector.@@.baz)&select=id" `shouldRespondWith`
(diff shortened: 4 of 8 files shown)


russelldavies:
@ruslantalpa I'm not sure if the tests are failing because the new schema and data is not being loaded or there are issues with the tests themselves.

ruslantalpa:
@russelldavies the error message of the failing test is very clear imo :)
permission denied
you need to allow the database user with which postgrest makes the request to access that table,
probably here https://github.com/russelldavies/postgrest/blob/025f4679376263afa110e1816243d7d2a387dcda/test/fixtures/privileges.sql#L13

src/PostgREST/Types.hs
```diff
@@ -163,7 +163,13 @@ operators = M.fromList [
   ("is", "IS"),
   ("@@", "@@"),
   ("@>", "@>"),
-  ("<@", "<@")]
+  ("<@", "<@"),
+  ("&&", "&&"),
```
ruslantalpa: this is a special char in urls, it should be something else.
like notice we have `("gte", ">=")`, i think things like `<<, && ...` need to be changed to something else 
I am not sure to what but after the tests pass, let's discuss the options then you just change the tuples and the tests

russelldavies:
Thanks, for some reason I wasn't getting that as my output when using a local postgres database. Works fine using Docker though.

src/PostgREST/Types.hs
russelldavies: Yeah, as I mentioned in the gitter chat, it requires the client to encode it as `%26`. I've found that most libraries do this already so it's not that inconvenient but it will need to be noted in the documentation.

In terms of replacing the ampersand with another character, the best ASCII equivalent I can think of is `e`since an `&` is written as that letter in handwriting. It looks kind of confusing though. It's also legal to use unicode characters in urls, so I suppose mathematical symbols like the intersection symbol could be used, but this is possibly very unfriendly.

ruslantalpa: i was thinking more like
`dat = @@`
`atgt = @>`
`ltat = <@`
`dlt = <<`
`dgt = >>`
`amlt = &<`
`amgt = &>`
`pipe = -|-`
`dam = &&`

russelldavies: That works but it does introduce a breaking change since it would replace `@@`, which has been implemented for a while now.

My preference would be to also keep using symbols as much as possible for these.

ruslantalpa: the old mode can be supported since you can have 2 tuples resolving to the same operator.
my preference would be to be consistent :), all chars, i think `@` was a mistake.

other opinions?

russelldavies: Good point, in that case I'm fine with what you suggested. If anyone else needs to weigh in on this, add them to the discussion, otherwise I'll update the PR.

ruslantalpa: let's wait a day or two @begriffs @steve-chavez

steve-chavez:
I think renaming the special chars operators `&>, &&, <@` to letter ones would be more consistent and also avoid urlencoding, though it adds more reserved keywords and later more if we manage to support postgis comparison operators.

If the renaming of comparison operators is done I think it should be more mnemonic, for example:
```haskell
cs(contains)   = @>
cd(contained)  = <@
```
Since `<@ @>` mean the same for both the ranges(1) and arrays(2) I think that's fine. 

The issue with this approach comes when renaming `&&` to say `ov`, because that operator is `overlap` for ranges, arrays and geometries(3) but `contains or is contained` for inets(4). 

Same goes for strictly left/right`<</>>` renamed to`sl/sr` meaning the same for ranges and geometries but not for inets (`<</>>` contained/contains).

Overall my proposal for the new operators, ignoring the perhaps rare use case for inets, would be:
```haskell
ov(overlap)                       = &&
sl(strictly left)                 = <<
sr(strictly right)                = >>
nxr(does not extend to right of)  = &<
nxl(does not extend to left of)   = &>
adj(adjancent)                    = -|-
```

Also I think is not necessary to introduce the aliasing of old operators in this PR, maybe we could leave that for another one.

(1): https://www.postgresql.org/docs/9.6/static/functions-range.html
(2): https://www.postgresql.org/docs/9.6/static/functions-array.html
(3): https://www.postgresql.org/docs/9.6/static/functions-geometry.html
(4): https://www.postgresql.org/docs/9.6/static/functions-net.html

ruslantalpa:
@steve-chavez suggestions are good as long as they are documented and mention what the short names stand for.
only one comment, the aliasing of the @>  <@ should be done now so that the upcoming 4.3 version has all the operators right.

russelldavies:
By aliasing the operators it would then be possible to deprecate the symbol operators (and add a notice to the changelog and docs). In that case, should the `@@` also be changed to `fts` (full text search) in this PR?

ruslantalpa:
+1 for doing it, not sure about the name

begriffs:
> If the renaming of comparison operators is done I think it should be more mnemonic,

Agreed. Good idea Steve, these mnemonic names are nice. 

> The issue with this approach comes when renaming && to say ov, because that operator is overlap for ranges, arrays and geometries(3) but contains or is contained for inets(4).

I see what you mean, and it seems OK to go with "ov" for the more common meaning of overlap.

> Same goes for strictly left/right<</>> renamed tosl/sr meaning the same for ranges and geometries but not for inets (<</>> contained/contains).

We could once again pick the more common interpretation as you suggest. Alternately could choose llt and ggt, as being "beefed up" lt and gt. Although that does hint more at their symbolic representation than meaning, so I'll defer it to you guys. Either way is fine by me.

> Also I think is not necessary to introduce the aliasing of old operators in this PR, maybe we could leave that for another one.

We should continue to allow @@, @>, and <@ (in addition to their mnemonic new  names) for backward compatibility, but deprecate them in 4.3 and ultimately remove them in 5.0.

begriffs:
> In that case, should the @@ also be changed to fts (full text search) in this PR?

Yeah, aliasing @@ as "fts" sounds good - it also exists as a geometric operator to get the center of a circle, but the full text usage is the only boolean operator of that name.

steve-chavez:
I've seen the `@@` referred as the full-text search operator before, so I'm fine with `fts` for that, also for `<</>>/&&` my vote is for going with the most common use case(`sl/sr/ov`) and if a need arises for using those operators with inets it can be documented later, it's an inconsistency in PostgreSQL anyway, so I think it will be fine.

begriffs:
Add an entry to the changelog and it looks good to me!

src/PostgREST/QueryBuilder.hs
```diff
@@ -418,7 +418,7 @@ pgFmtFilter table (Filter fld (Operation hasNot_ ex)) = notOp <> " " <> case ex
    (op, VText val) -> pgFmtFieldOp op <> " " <> case op of
      "like"  -> unknownLiteral (T.map star val)
      "ilike" -> unknownLiteral (T.map star val)
-     "@@"    -> "to_tsquery(" <> unknownLiteral val <> ") "
```
steve-chavez: So this means that we are removing support for @@ now and not just deprecating it, right?

This has not been detected by the tests since the old ones have just been updated.

How about adding a test section for the deprecated operators, could be just the old tests unchanged.

begriffs:
> So this means that we are removing support for @@ now and not just deprecating it, right?

I completely overlooked this, good catch.

> How about adding a test section for the deprecated operators, could be just the old tests unchanged.

Yeah we should keep the old tests, and maybe add "(deprecated)" to their descriptions.

src/PostgREST/QueryBuilder.hs
russelldavies: Whoops, will correct that now.

begriffs:
Hmm, this merge conflict should not be happening because the `.gitattributes` file for this project contains `/CHANGELOG.md merge=union`. Maybe the forward slash is messing it up, would you try adding a commit to remove the slash and see if that solves the merge conflict?

src/PostgREST/QueryBuilder.hs
russelldavies: Yeah, is there some issue with the tests that they're not catching this?

ruslantalpa: well, you rewrote all the tests to work with `ts` instead of `@@` so ... :)

ruslantalpa:
for the conflicts, you need to rebase your branch with the latest changes to remove the conflicts

src/PostgREST/QueryBuilder.hs
ruslantalpa: i.e. i would say to copy one of the tests with `ts` and make it with `@@` just to make sure it's still working, i think it's enough.

russelldavies: That's what I've done. I left the original test, copied it, and in the copy changed the operator. I've just checked again and can't spot what error I've made.

ruslantalpa: i am sorry, you are right, i was duped by the renaming of the test title.
looking at why this was not detected

ruslantalpa: is there any chance that 
`col @@ 'lala'::unknow` is the same as `col @@ to_tsquery('lala'::unknown)`?
i.e. postgres is doing the casting on it's own without the need for to_tsquery in which case that line should be removed? or is this autocasting a new feature of 9.5+ and not present in older versions?

ruslantalpa: that the only thing that would explain why the tests are till working

russelldavies: Looks to be. I've asked a postgres developer I know to be sure.

russelldavies: So it depends on what data types are involved. If it's text, `@@ 'asdf'::unknown` will cast it so the `to_tsquery` is not required.

steve-chavez: If you run:
```sql
select to_tsvector('impossible') @@ 'impossible';
```
It'll output `false`, but if you run:
```sql
select to_tsvector('impossible') @@ to_tsquery('impossible');
```
It'll output a `true`, that's because:
```sql
SELECT 'impossible'::tsquery, to_tsquery('impossible');
   tsquery    | to_tsquery
--------------+------------
 'impossible' | 'imposs'
```
The `to_tsquery` is needed to build the lexemes, I think that `@@` just operates against strings and converts them to `tsquery` and for the case of `asdf, lala == to_tsquery(asdf), to_tsquery(lala)` the lexemes are the same.

So the `ts_query` is required the issue is that the tests are not catching that('foo/bar/qux' same as lexemes) so maybe you could improve the tests to catch that difference.

russelldavies: Thanks, I'm updating the fixture data so that the stems are not equal to the words.

russelldavies:
The `tsearch` table fixture data has been updated so that it's no longer possible to accidentally do a full text search match if the string to be searched on has not been transformed by `to_tsquery()`. 

So if `@@` is used as the full text search operator and `"@@"    -> "to_tsquery(" <> unknownLiteral val <> ") "` is missing then the tests will now fail.

daurnimator:
I have a need for the inet contains operator `>>` and contained within `<<`. Are they included in this PR? https://www.postgresql.org/docs/9.1/static/functions-net.html

daurnimator:
I guess I should ask: why are we deprecating the symbol operators? IMO we could allow *any* operator that meets the postgres criteria (https://www.postgresql.org/docs/current/static/sql-createoperator.html). And then add aliases for some common ones e.g. `gt` for `>`.

steve-chavez:
Not every PostgreSQL operator is fast/safe to publicly expose in an api, for example see the issue about the regex operator https://github.com/begriffs/postgrest/issues/539 it posed a ReDos threat https://github.com/begriffs/postgrest/issues/539#issuecomment-255616248.

For the inets `<</>>` you could use `sl/sr` once this PR is merged.

daurnimator:
> Not every PostgreSQL operator is fast/safe to publicly expose in an api, for example see the issue about the regex operator #539 it posed a ReDos threat #539 (comment).

Except that `~` is not only the regex operator, but also the bitwise NOT operator, and the [postgis contains operator](http://postgis.net/docs/manual-2.0/ST_Geometry_Contain.html). The latter two I can imagine wanting to use with postgrest.

This implies we need a way to whitelist not operators, but the *types* they apply to (and perhaps a blacklist rather than a whitelist)

> For the inets <</>> you could use sl/sr once this PR is merged.

Yes, and I will; however I think it's a bit odd that we're inventing aliases here.

daurnimator:
> This implies we need a way to whitelist not operators, but the types they apply to (and perhaps a blacklist rather than a whitelist)

Looks like we can blacklist operators like so:

```sql
revoke execute on function pg_catalog.textregexeq(text,text) from public;
```

Where `pg_catalog.textregexeq` is the function behind the regex `~` operator.

daurnimator:
An alternative way for us to blacklist operators is to shadow the operators:

```sql
create schema operator_blacklist;
create function operator_blacklist.fail(text, text) returns bool as $$begin raise insufficient_privilege; end;$$ language plpgsql;
create operator operator_blacklist.~ (procedure = operator_blacklist.fail, leftarg=text, rightarg=text);
```

Then on each connect to the database, run something like: 
```sql
set search_path = operator_blacklist,pg_catalog;
```

ruslantalpa:
@daurminator it's a discussion worth having but probably not in the context of this PR which looks like it's ready for merging

@begrifs @steve-chavez agreed? Merging?

daurnimator:
> @daurminator it's a discussion worth having but probably not in the context of this PR which looks like it's ready for merging @begrifs @steve-chavez agreed? Merging?

It affects this PR, as I don't think the old operators should be deprecated (this PR adds deprecation notices).
And I also don't think we should be adding all these aliases, I disagree with @steve-chavez's comments above above. For now, we should just be adding the raw operators e.g. `<<` and `>>`

steve-chavez:
@daurnimator The renaming of the range operators is not only about consistency with old ones but useful to avoid urlencoding, even more in the case of the new range operators like `&< &>` that contain a `&`, that will get silently ignored when typed in an browser url address bar since it will interpret it as an empty querystring param.

Also if it's a matter of not having aliases, PostgREST already had aliases for `< > =` for a while now, so I don't see that as a strong reason to block this PR for longer.

About your proposal I also think that is worth discussing it but it'll be better to do that on it's own issue/PR.

daurnimator:
I disagree; but before merging this PR, could we remove the comments about deprecation?

ruslantalpa:
- if it were possible not to have url encoding issues than we would have used the operators directly, but since it is not, all of them should be letters for consistency and to avoid problems.
- `@@ @>` ... was a mistake, now it's being fixed
- the `@@` operator is not deprecated, only it's method of calling
- postgret url format is not trying to be a form of SQL, we are in the REST world and also postgrest is not slowly going in the direction of exposing all the sql capabilities so i don't see a big problem of moving a little away from sql by using aliases for operators.

as before, my vote is for merging as is, with depreciation for direct names of operators.

daurnimator:
> if it were possible not to have url encoding issues than we would have used the operators directly, but since it is not, all of them should be letters for consistency and to avoid problems.

What url encoding issues are you having? I can't say I've ever had any except for  '&' and was only when adding to my browser address bar or on command line (vs when calling api from programming languages or via swagger tooling)

> @> ... was a mistake, now it's being fixed

How was it a mistake? I use it without issue regularly. It's tremendously straight forward to think about when I can just refer to postgres docs for @> instead of indirecting via postgrest manual.

begriffs:
I believe the best design for specifying operators in the command line should be

1. Consistent. If we're using letter abbreviations we should stick to letter abbreviations and not mix symbols.
2. Compatible with URI unreserved characters as specified in [RFC 3986](https://tools.ietf.org/html/rfc3986#section-2.2).

Property 2 means that & and @ are out, and we'll need a text abbreviation. However once we choose a text abbreviation for them, then property 1 would argue that we use text representations for all of them.

I acknowledge that it's a shame to reinvent the postgresql operator names that people already know, but the point of the project is to provide database access through web standards, and the sql operator names don't match that well. I regard the name changes as part of the db-to-http mapping.

@daurnimator have I addressed your concern, or is there something I'm overlooking? Maybe the only danger is that I'm overstating the benefits of consistency, and that we *could* continue having both symbolic and text names for operators at least for the symbols which don't cause as much trouble.

daurnimator:
> Compatible with URI unreserved characters as specified in RFC 3986.

That just means they have to be escaped; not that it's impossible to use them.

>  acknowledge that it's a shame to reinvent the postgresql operator names that people already know

IMO the consistency with postgres is more important than the consistency across operators.

That said: for users allergic to escaping in urls, I can see why you might want aliases for some operators (particularly those containing an ampersand). The aliases should be taken from a space that doesn't conflict with *any* possible postgres operator syntax, the alphanumerics are a fine choice here.

However I don't think we should have aliases for *all* operators, having to remember e.g. that to use `-|-` I have to tell posgret `adj` is pointless trivia that will only trip me up; and more importantly: make it something else you have to know to use a postgrest API: the less you have to know the better.

begriffs:
> That just means they have to be escaped; not that it's impossible to use them.
> ...
> IMO the consistency with postgres is more important than the consistency across operators.

What if we provided both? For every operator, include both its postgres name, and the alphabetical name? Then those who like the pg names (and who don't have a problem with urlencoding) can use one kind, and people who want clean urls can use another?

It seems pretty trivial to include more than one mapping, not a big maintenance burden.

steve-chavez:
For what it's worth when I started using PostgREST I liked the way it worked with the medium and allowed me to explore the filters in my browser address bar and not have to change the url when taking that to a webapp, I didn't consider having `gt/lt/eq` as pointless trivia but a way to embrace web standards with the less friction(no urlencoding), also it wasn't really a big overhead to learn them since the names were mnemonic.

@begriffs I prefer to maintain consistency, and between the burden of us maintaining the plain postgres operators, that is adding them, keep/add tests, adding docs, also support(we had already some confused users regarding urlencoding) and the burden of an user maintaining his own fork with just adding the operators, I'd say the latter is a lighter one, it's trivial to keep in sync, considering the arguments against the aliased operators presented in this issue for me that's the fairer outcome.

As before my vote is to merge this PR as it is, and if it's necessary continue the discussion in another issue/PR.

PR #939, opened 2017-08-14
Fix compatibility with protolude-0.2

felixonmars:


begriffs:
This is nice, thanks. Good to know the changes won't be so bad for protolude 0.2.

Perhaps we should wait to apply this until the Stackage LTS version uses the new version of protolude?

felixonmars:
Not sure if they will update it soon due to the ongoing 8.2.1 move. But it's surely fine to wait :)

(Note: I already applied the patch for Arch package as we have the new protolude already)

begriffs:
Ah I didn't realize the arch situation. Would merging this PR help make your continuing arch support easier?

felixonmars:
It will if you are going to release a new version soon - so I can drop the patch here :)

begriffs:
We are preparing to release v4.3, so I'll merge this change.

Also, not sure it affects your packaging but we're dropping the initial zero from the version, so it will be 4.3.0 rather than 0.4.3.0. The quadruple version numbers is a weird Haskellism and we'd rather use regular semver.

felixonmars:
That will be fine, and I like the change :P

PR #942, opened 2017-08-15
Allow passing URI with database to tests

daurnimator:
Without this, psql will try to use the default database name for the user

The database part of the uri only needs to be stripped for the final part of create_test_db

### DIFF
--- a/test/create_test_db
+++ b/test/create_test_db
@@ -17,8 +17,7 @@ then
 fi
 
 BASEPATH=$( cd $(dirname $0) ; pwd -P )
-#Remove database path from the connection uri--prevents setting up the new database name with PGDATABASE
-URI=$(echo $1 | cut -d'/' -f1-3)
+URI="$1"
 #Extract host and port--we need this to form the new connection string
 HOST_PORT=$(echo $URI | cut -d'/' -f3 | cut -d'@' -f2 )
 DB=$2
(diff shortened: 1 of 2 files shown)


begriffs:
Does this affect the instructions at https://postgrest.com/en/v4.1/install.html#running-the-tests or is it more fixing a bug where the user name for the connection is disregarded?

daurnimator:
> Does this affect the instructions at https://postgrest.com/en/v4.1/install.html#running-the-tests

No.

> or is it more fixing a bug where the user name for the connection is disregarded?

Username is untouched; it's the database name that was getting stripped off.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- #930, Split table comment on newline to get OpenAPI operation summary and description - @daurnimator
```
begriffs: Oh I see, this is applied to the wrong PR

begriffs:
@dsimunic you wrote these scripts initially, and the code seems kind of subtle to me, like how it interacts with PGDATABASE. Does this PR look good to you?

ruslantalpa:
@daurnimator i think i understand what you are saying, code looks fine, out of curiosity, what is the problem with psql making the initial connection to the default db (and not the one in the connection string) sice the script does not interact with it at all?

Is there a case when the user does not have a default db thus psql failing to connect?

dsimunic:
The change looks fine--it improves the flexibility of the script for the edge case when the user doesn't have the default database. The change doesn't impact the description of the testing procedure.

I'm also curious in what scenario there is no default db for the user.

daurnimator:
> Is there a case when the user does not have a default db thus psql failing to connect?

The database I'm testing against is not via a unix socket (connecting via TCP to localhost). There is only a 'postgres' database, not a database with permissions to my current user.

Issue #915, opened 2017-07-19
distinct/unique


daurnimator:
PostgREST should have a way to perform a `SELECT DISTINCT` query.

I suggest a new reserved query word 'distinct'.

  - With no value it would just do a `select distinct` e.g. `/mytable?distinct`
  - Otherwise it is a comma separated list of columns (or computed columns) e.g. `/mytable?distinct=name`

Note that the columns that you are distinct on will also need to have order applied.

begriffs:
This has certainly been a frequently requested feature. My only hesitation is making "distinct" another reserved word, meaning it would block any filters on a column that happened to be called "distinct." However, like "select," this one is a sql reserved word too so it makes sense that we reserve it as a special url parameter.

I hesitate to try messing with the url parsing and query generation part of the code myself, that's more @steve-chavez's domain. Perhaps he can estimate the difficulty of this feature.

steve-chavez:
Not sure if this should be implemented, by allowing `distinct` to be applied to any column unrestricted it could potentially ddos a database.

I've bumped into a slow distinct query in postgresql a while ago and solved it by using a `group by` instead of `distinct`, remember distinct generating a more expensive seq scan, I don't have the details anymore but a quick googling suggest the problem could still persist:

https://dba.stackexchange.com/questions/93158/how-to-speed-up-select-distinct
https://dba.stackexchange.com/questions/63268/postgresql-being-slow-on-count-distinct-for-dates

As mentioned in one of the answers, you can mostly avoid doing `distinct` with better database-design or better queries.

So in general I think `distinct` is a expensive operation(similar to `group by`) and it should be better exposed behind a view or a stored procedure.

ruslantalpa:
@steve-chavez this is a good point against this. PostgrREST should expose only safe/fast operations

russelldavies:
For what it's worth, `DISTINCT` is just syntactic sugar for `GROUP BY` and I think in modern versions of postgres it's even implemented using the same code. (And `DISTINCT ON` is a postgres extension from way back that's a bit of a performance hack). So @steve-chavez's conclusion is correct.

daurnimator:
> As mentioned in one of the answers, you can mostly avoid doing distinct with better database-design or better queries.

I have *many* situations where I've wished for postgrest to have a distinct operator.
[...]

PR #943, opened 2017-08-15
Reserve 'distinct' query parameter

daurnimator:
For #915

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #877, Base64 secret read from a file ending with a newline - @eric-brechemier
 - #896, Boolean env var interpolation in config file - @begriffs
 - #885, OpenAPI repetition reduced by using more definitions- @ldesgoui
+- #915, Reserve 'distinct' query param - @daurnimator
 - #924, Improve relations initialization time - @9too
 - #927, Treat blank pre-request as missing - @begriffs
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -118,7 +118,7 @@ userApiRequest schema req reqBody
       , iPreferRepresentation = representation
       , iPreferSingleObjectParameter = singleObject
       , iPreferCount = hasPrefer "count=exact"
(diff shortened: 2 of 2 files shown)


ruslantalpa:
i don't see the point in this PR.
it's not like it will warn users not to use this keyword, it will just silently drop it from consideration leaving users to wonder what's going on.

In addition to that, it was not decided that a reserved keyword is the way to go here,
distinct is not on the same level as select/where/from/order ... this is something that is a modifyer in the select part.

daurnimator:
> distinct is not on the same level as select/where/from/order ... this is something that is a modifyer in the select part.

distinct is a sql reserved word in every sense see https://www.postgresql.org/docs/current/static/sql-keywords-appendix.html

ruslantalpa:
this is not what i was saying, distinct operates in the `SELECT` space, it does not stand on it's own like WHERE/GROUP/ORDER.

i.e. the place for this feature in in the `&select=...distinct...` space, and not a top level keyword

daurnimator:
> i.e. the place for this feature in in the &select=...distinct... space, and not a top level keyword

I disagree, as noted in https://github.com/begriffs/postgrest/issues/915#issue-243948218 we need to be able to provide key/value syntax for `distinct`, and as a matter of http api taste, I think this belongs at the top level.

ruslantalpa:
and what will you do when the next logical request will be "i need to request distinct entities on the embedded entities"? :)
and i don't see any problems with the value you are saying, it's even more natural
`select=distinct(id),name,subitems(distinct(id),name)`

daurnimator:
> and i don't see any problems with the value you are saying, it's even more natural
> select=distinct(id),name,subitems(distinct(id),name)

Sure; that works too. Though it does look quite a lot like a resource embedding.

begriffs:
I agree with Ruslan here that disregarding a parameter without raising a warning, and not acting on it in any other way, could really leave people scratching their heads.

Also the first item feels more cohesive to me than the second in this list:

* `?select=distinct(id),name,subitems(distinct(id),name)`
* `?select=id,name,subitems(id,name)&distinct=id,subitems.id`

Although I admit that this may be purely subjective.

In any case, I'm closing this pull request as-is because until we act on the distinct param I believe it's confusing to silently remove it.

Issue #328, opened 2015-10-24
rpc: specify argument types (GET requests on RPCs)


LeKovr:
There is incorrect SQL generated when stored procedure has argument of array type:

``` sql
CREATE OR REPLACE FUNCTION "1".test(q INTEGER[]) RETURNS SETOF INTEGER LANGUAGE 'sql' AS $_$
      SELECT unnest($1)
$_$;
```

Call:

```
curl -d '{"q":[1,2,3]}' http://localhost/rpc/test
```

Generated SQL (incorrect):

``` sql
select * from "1".test(q:='[1,2,3]'::unknown)
-- ERROR:  missing "]" in array dimensions
```

Correct SQL:

``` sql
select * from "1".test_text3(q:='{1,2,3}'::unknown);
-- or
select * from "1".test_text3(q:=ARRAY[1,2,3]);
```

Also, when argument has type TEXT[] we got another error (missing dimension value) but both of them might be solved by replacing [] with {}.

begriffs:
Thanks for reporting! I'll look into it.

begriffs:
Appears related to #282

LeKovr:
Yes, string arg from #282 is working:
```
curl -d '{"q":"{1,2,3}"}' http://localhost/rpc/test
```
but real array support would be better than this string join/split method

begriffs:
There may be an ambiguity in this problem. There can be two stored procedures with the same name but taking arguments of different types. What if one of them takes a postgres array and the other takes a json value? I don't think we could tell which is the right one to execute.

We could examine the functions and their argument types like this

```sql
SELECT ns.nspname, p.proname, oidvectortypes(p.proargtypes)
FROM pg_proc p
INNER JOIN pg_namespace ns 
ON (p.pronamespace = ns.oid)
WHERE ns.nspname =  'public';
```

Then if the function in question is not ambiguous we'll know whether to format as array or json. Might just be easier to make all your functions accept a json values rather than arrays. What do you think?

LeKovr:
> What if one of them takes a postgres array and the other takes a json value?

This is the same problem as with int/text args or even text/json. What is the general way to solve it? 

> Might just be easier to make all your functions accept a json values rather than arrays

Will it be any json supported or just json arrays?

LeKovr:
> We could add query params to an RPC post request to specify types, something like

Let's think about example like 
```
/rpc/myproc?arg1=integer&arg2=text&arg3=decimal
```
[...]

PR #946, opened 2017-08-20
Allow GET on RPC

steve-chavez:
Addresses the "GET requests on RPCs" part of #328, I decided not to support the part about specifying argument types, because in my proposal we would do `GET /rpc/name?id=1::int` however this way of defining the types would not be possible for POST requests.

Maybe something like the `argtypes=id::int,gender::bit` proposed in the issue could be used later as an additional query param on GET and POST for those rare cases of proc ambiguity, but for now I think this feature addresses the most common use case and also for the cases of proc ambiguity the user has workarounds like renaming the proc.


src/PostgREST/App.hs
```diff
@@ -227,23 +229,28 @@ app dbStructure conf apiRequest =
               let acceptH = (hAllow, if tableInsertable table then "GET,POST,PATCH,DELETE" else "GET") in
               return $ responseLBS status200 [allOrigins, acceptH] ""
 
-        (ActionInvoke, TargetProc qi, Just (PayloadJSON payload)) ->
+        -- These two are ensured by shouldParsePayload in ApiRequest
+        (ActionInvoke{isReadOnly=False}, _, Nothing) -> pure notFound
```
begriffs: What do these two notFound matchers do? When I remove them locally it still builds and the tests still pass.

src/PostgREST/ApiRequest.hs
```diff
@@ -48,7 +49,7 @@ type RequestBody = BL.ByteString
 -- | Types of things a user wants to do to tables/views/procs
 data Action = ActionCreate | ActionRead
             | ActionUpdate | ActionDelete
-            | ActionInfo   | ActionInvoke
+            | ActionInfo   | ActionInvoke{isReadOnly :: Bool}
```
begriffs: Don't know if it's feasible, but I have an idea for a simplification. What if:

* ActionInvoke went back to not taking the isReadOnly option
* iRpcQParams became iRpcParams which ApiRequest.hs would fill in from either the post body or query params as relevant
* Make a new query attribute iReadOnly and set it true for ActionRead, ActionInfo, ActionInspect, or GET-based ActionInvoke. Refer to iReadOnly in transactionMode

Do you think that would be a net simplification, or just shift complexity around? I'm feeling there would be less pattern matching in other places that way.

src/PostgREST/ApiRequest.hs
```diff
@@ -132,6 +136,14 @@ userApiRequest schema req reqBody
       , iCookies = fromMaybe [] $ parseCookiesText <$> lookupHeader "Cookie"
       }
  where
+  (filters, rpcQParams) =
+    case action of
+      ActionInvoke{isReadOnly=True} -> (filter (liftM2 (||) (isEmbedPath . fst) (hasOperator . snd)) flts,
```
begriffs: Perhaps the [partition](https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-List.html#v:partition) function could clean up the code here where you're splitting a list into two which do and do not satisfy a predicate.

test/Feature/RpcSpec.hs
```diff
@@ -0,0 +1,288 @@
+module Feature.RpcSpec where
```
begriffs: Good idea to make this its own section.

test/Main.hs
```diff
@@ -82,6 +83,7 @@ main = do
     , ("Feature.DeleteSpec"       , Feature.DeleteSpec.spec)
     , ("Feature.InsertSpec"       , Feature.InsertSpec.spec)
     , ("Feature.QuerySpec"        , Feature.QuerySpec.spec)
+    , ("Feature.RpcSpec"          , Feature.RpcSpec.spec)
```
begriffs: The indentation here looks a little different from the other lines.

begriffs:
This is a great feature, thank you! I'm sure you're making many people's day with this one.

src/PostgREST/App.hs
steve-chavez: Previoulsy we had:
```haskell
   (ActionInvoke, TargetProc qi, Nothing) ->
   -- Not explicitly pattern matched and going to notFound through
   _ -> return notFound
```
I was trying to maintain that mapping though I didn't consider to modify the `relevantPayload` function, I think I have a better solution now, with that these pattern matches can be removed.

src/PostgREST/ApiRequest.hs
steve-chavez: An issue I see with the idea is that `iReadOnly` is only used for `ActionInvoke` and don't think it'll be good for separation of concerns to have that attribute in `ApiRequest`, all of his attributes seem more directly related to things that the user wants to get/set.

I've come up with a modification to the `relevantPayload` function to reduce the pattern matches, check my latest commit.

test/Main.hs
steve-chavez: I don't see where the indentation is different, the first `,` has the same indentation as the others and the second `,` too.

src/PostgREST/ApiRequest.hs
steve-chavez: That's an awesome function, it perfectly suits my case since I had a `a or b` and the negation of that is `~(a or b) = ~a and ~b`.

test/Main.hs
begriffs: Oops, I was on a small screen and it was just the line wrapping on github's preview that confused me.

src/PostgREST/Types.hs
```diff
@@ -199,6 +199,10 @@ type Alias = Text
 type Cast = Text
 type NodeName = Text
 
+-- Rpc query param, only used for GET rpcs
+type RpcQParam = (Text, Text)
+data RpcParam = RpcRW Object | RpcR [RpcQParam]
```
ruslantalpa: This data type `RpcParam ` is not needed and is combining into one two types of information,
- is transaction readonly and the name of the type does not suggest that
- the actual params and they have 2 forms for the same data, i.e. `Object` and `[(key,val)]` which is the same as `Object`

I would say this needs to happen:
- remove this type
- `callProc` maintains the `JSON.Object` param as is, which is calculated in the `App.hs`
- `callProc` gains another Bool (isReadOnly) which is calculated in also in `app`
- optional: the callProc signature is getting ugly, can we invent some kind intermediate record data type to hold together a few of those parameters together?

in short, the section in `app` referring to rpc, does a bit of additional work so that inside `callProc` there is no trace of the form in which the params were sent (get/post)

begriffs: I agree that the RPC params should probably be represented in a single way whether they come in from the post body or the query string.

steve-chavez:
If there isn't more feedback I think this could be merged.

ruslantalpa:
let's concentrate on auto release, once that is done, there's nothing stopping us from having a relseas whenever a new feature is merged. no need to rush this particular one so that it makes it into "4.3", it can make it into 4.4, which could be in a week (or day) after 4.3

steve-chavez:
I've disabled Travis builds for PRs, when this PR is merged maybe we could try a new release, and see if the osx release works now.

begriffs:
Let's do this!

steve-chavez:
My bad, I've missed a test case here and there's a bug, currently filters preceded with `not.` are not working:
```http
GET /rpc/get_projects_below?id=5&id=not.gt.2&select=id

{"hint":null,"details":null,"code":"22P02","message":"invalid input syntax for integer: \"not.gt.2\""}
```
The `not.` part is been considered as an arg for the function, the fix for this is simple just adding `M.keys operators ++ ["not"]` [here](https://github.com/begriffs/postgrest/pull/946/files#diff-02f231cf09412413f8f0e9d5e5d7a06eR144).

Another bug I've just noticed because this was merged after #964 is that the new `plain.fts/french.fts` filters aren't going to work with RPC GET because they will get treated as args, this fix is a bit more complicated.

Should I open a new PR for these fixes or could we just revert this commit? I could make the fixes here.

ruslantalpa:
New pr

begriffs:
Good catch before the problem made it into any numbered release.

PR #947, opened 2017-08-20
Allow optparse-applicative 0.14

felixonmars:
It builds and works fine here.


begriffs:
Cool, for a test I added `- optparse-applicative-0.14.0.0` to stack.yaml and that built and ran fine as well. So looks good to me!

Issue #601, opened 2016-05-20
Better full text search capabilities


NikolayS:
(Per discussion in gitter chat)

PostgREST supports `@@` operator for GET queries, but it 
- 1) lacks capability to point out, which full text config should be used during `text`→`tsquery` conversion (e.g.: "english" or "russian", which have different stemmers/ispell dictionaries)
- 2) leaves the task of converting user's input to `tsquery` to the client-side developer (not always a good thing, because client-side developers often don't understand how Postgres works, how to clear user input from special characters, trim it, etc – so there is a high chance to get SQL error from Postgres; at the same time, Postgres already provides `plainto_tsquery` function, which helps to solve this task w/o headache – more about it here: http://www.postgresql.org/docs/current/static/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES)

The only current solution to have both features is to use custom function (`/rpc/xxx`).

It would be great to have both capabilities available for GET queries by default – some form of API call that would be directly transformed to `plainto_tsquery(configname, userinput)`. Maybe a new operator(s), for instance, `@` (for `plainto_tsquery` with default FTS config), `configname@` (explicit choosing config for `plainto_tsquery`), and `configname@@` (for `to_tsquery(configname, query)`):  

``` http
GET /items?tsvector=@urlencoded%20query%20here
GET /items?tsvector=english@urlencoded%20query%20here 
GET /items?tsvector=russian@много%20букв
GET /items?tsvector=english@@apples%20%7C%20bananas
```

…that would be translated correspondingly to:

``` sql
items.tsvector @@ plainto_tsquery('urlencoded query here')
items.tsvector @@ plainto_tsquery('english', 'urlencoded query here')
items.tsvector @@ plainto_tsquery('russian', 'много букв')
items.tsvector @@ to_tsquery('english', 'apples | bananas')
```

@ruslantalpa's thoughts about it:

> there is no PostgREST way of doing that at the moment (except custom rpc), only if there is a way of doing that in the db somehow to turn col @@ 'string' into col @@ plainto_tsquery('russian', 'string') on the fly.
> i can think of a relatively fast way of being able to call functions in the select parameter while maintaining backword compatibility, i am not sure about fitlers
> since col=@@.plainto_tsquery('russian', 'string') can be interpreted as col @@ 'plainto_tsquery(\'russian\', \'string\')' We need a way to differentiate

begriffs:
[...]

PR #964, opened 2017-09-07
Allow specifying dictionary and plain/phrase in full text search

### LINKED ISSUE #887: Extend @@ for plainto_tsquery phraseto_tsquery support
For our API, we use a tsquery interface.  Modifying user input for a good text query can be a bit of a challenge on the client side.  We are building a quick rpc interface, to use the postgres services to do this, but that results in two calls to the server.

```bash
http --json https://wine-api.library.ucdavis.edu/rpc/strto_tsquery str='Moet & Chandon'
#[  {  "phrase": "moet <-> chandon",   "plain": "moet & chandon" } ]
http --json https://dev-wine-api.library.ucdavis.edu/catalogs select==catalog_id q=='@@.moet <-> chandon'
```

It seems like there are more sophisticated questions about tsquery, but what about a simple suffix for plain and phrase queries, so you could do interpret the string directly.

``` bash
http --json https://dev-wine-api.library.ucdavis.edu/catalogs select==catalog_id q=='@@z.Moet & Chandon'
```

I think the changes would be pretty simple? If I'm not missing something.

``` diff
diff --git a/src/PostgREST/QueryBuilder.hs b/src/PostgREST/QueryBuilder.hs
index d01f317..8087b8e 100644
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -421,6 +421,8 @@ pgFmtFilter table (Filter fld (Operation hasNot_ ex)) = notOp <> " " <> case ex
      "like"  -> unknownLiteral (T.map star val)
      "ilike" -> unknownLiteral (T.map star val)
      "@@"    -> "to_tsquery(" <> unknownLiteral val <> ") "
+     "@@z"    -> "phraseto_tsquery(" <> unknownLiteral val <> ") "
+     "@@p"    -> "plainto_tsquery(" <> unknownLiteral val <> ") "
      "is"    -> whiteList val
      "isnot" -> whiteList val
      _       -> unknownLiteral val
diff --git a/src/PostgREST/Types.hs b/src/PostgREST/Types.hs
index 3d2967b..2046cac 100644
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -153,6 +153,8 @@ operators = M.fromList [
   ("isnot", "IS NOT"),
   ("is", "IS"),
   ("@@", "@@"),
+  ("@@z", "@@"),
+  ("@@p", "@@"),
   ("@>", "@>"),
   ("<@", "<@")]
```
 If that sounds reasonable, I could submit a better pull request with tests if that would help.

steve-chavez:
Fixes #887, #601. Now is possible to:

```http
# Use language in fts query
GET /tsearch?ts_vector=french.fts.amusant

# Use plainto_tsquery and phraseto_tsquery
GET /tsearch?ts_vector=plain.fts.The%20Fat%20Cats
GET /tsearch?ts_vector=phrase.fts.The%20Fat%20Rats

# Combine both
GET /tsearch?ts_vector=phrase.english.fts.The%20Fat%20Cats

# "not" also working
GET /tsearch?ts_vector=not.phrase.english.fts.The%20Fat%20Cats
```

There is a catch the `phraseto_tsquery` function is only supported for pg > 9.6, so this should be noted in the docs also that's the reason why I upgraded the circle pg to 9.6.2, otherwise the tests would fail.

### DIFF
--- a/src/PostgREST/Parsers.hs
+++ b/src/PostgREST/Parsers.hs
@@ -1,6 +1,6 @@
 module PostgREST.Parsers where
 
-import           Protolude                     hiding (try, intercalate, replace)
+import           Protolude                     hiding (try, intercalate, replace, option)
 import           Control.Monad                 ((>>))
 import           Data.Foldable                 (foldl1)
 import qualified Data.HashMap.Strict           as M
@@ -21,7 +21,7 @@ pRequestFilter :: (Text, Text) -> Either ApiRequestError (EmbedPath, Filter)
 pRequestFilter (k, v) = mapError $ (,) <$> path <*> (Filter <$> fld <*> oper)
   where
     treePath = parse pTreePath ("failed to parser tree path (" ++ toS k ++ ")") $ toS k
-    oper = parse (pOperation pVText pVTextL) ("failed to parse filter (" ++ toS v ++ ")") $ toS v
+    oper = parse (pOpExpr pSingleVal pListVal) ("failed to parse filter (" ++ toS v ++ ")") $ toS v
     path = fst <$> treePath
     fld = snd <$> treePath
 
@@ -96,7 +96,6 @@ pFieldName = do
     dash :: Parser Char
     dash = isDash *> pure '-'
 
-
 pJsonPathStep :: Parser Text
 pJsonPathStep = toS <$> try (string "->" *> pFieldName)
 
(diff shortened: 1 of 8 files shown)


src/PostgREST/Parsers.hs
```diff
@@ -131,26 +130,36 @@ pFieldSelect = lexeme $
...
   where
-    pExpr :: Parser (Operator, Operand)
-    pExpr =
-          ((,) <$> (toS <$> foldl1 (<|>) (try . ((<* pDelimiter) . string) . toS <$> M.keys notInOps)) <*> parserVText)
-      <|> ((,) <$> (toS <$> foldl1 (<|>) (try . ((<* pDelimiter) . string) . toS <$> M.keys inOps)) <*> parserVTextL)
+    pOperation :: Parser Operation
+    pOperation =
+          Op . toS <$> foldl1 (<|>) (try . ((<* pDelimiter) . string) . toS <$> M.keys ops) <*> pSVal
+      <|> In . toS <$> (string "in" <* pDelimiter) <*> pLVal
+      <|> In . toS <$> (string "notin" <* pDelimiter) <*> pLVal
```
steve-chavez: I've noticed the `notin` operator not being documented on https://postgrest.com/en/v4.1/api.html#horizontal-filtering-rows, but it was documented on https://postgrest.com/en/v3.2/api_reading.html#filtering-rows.

Same case for the `isnot` operator, don't know when those two operators got deprecated but they still remain in the code.

Is it safe to remove them? I think we don't need them since we have `not.in` and `not.is` working.

begriffs: Yeah it's been a while since we deprecated them, so you can remove.

ruslantalpa:
@bgamari a while ago I asked if you could ask well-typed to look over postgrest code, but that is before i looked at what you are working on.
Thank you for the work you do on GHC and thanks to well-typed for paying you for that work and releasing it to FOSS.
That work benefits not only postgrest but every haskell developer.
We prioritized this feature for you as a thank you.

PS: and of course @steve-chavez nice work

bgamari:
Ruslan Talpa <notifications@github.com> writes:

> @bgamari a while ago I asked if you could ask well-typed to look over
> postgrest code, but that is before i looked at what you are working
> on.
>
> Thank you for the work you do on GHC and thanks to well-typed for
> paying you for that work and releasing it to FOSS.
>
Thanks! However, it should be said that the real credit belongs to
Microsoft Research, who funds this work.

> That work benefits not only postgrest but every haskell developer.
> We prioritized this feature for you as a thank you.
>
Thank you!

ruslantalpa:
I knew about Microsoft Research funding ** simonpj**, didn't know it was also your case, well, thank you @microsoft :) 

comparing github pages of microstof/apple/google/aws says a lot about who developers should say thank you to :)

src/PostgREST/Types.hs
```diff
@@ -174,8 +172,15 @@ operators = M.fromList [
   ("@@", "@@"),
   ("@>", "@>"),
   ("<@", "<@")]
-data Operation = Operation{ hasNot::Bool, expr::(Operator, Operand) } deriving (Eq, Show)
-data Operand = VText Text | VTextL [Text] | VForeignKey QualifiedIdentifier ForeignKey deriving (Show, Eq)
+
+data OpExpr = OpExpr Bool Operation deriving (Eq, Show)
+data Operation = Op Operator Text |
```
begriffs: Nice constructors for Operation.

src/PostgREST/Parsers.hs
```diff
@@ -131,26 +130,35 @@ pFieldSelect = lexeme $
     s <- pStar
     return ((s, Nothing), Nothing, Nothing, Nothing)
 
-pOperation :: Parser Operand -> Parser Operand -> Parser Operation
-pOperation parserVText parserVTextL = try ( string "not" *> pDelimiter *> (Operation True <$> pExpr)) <|> Operation False <$> pExpr
+pOpExpr :: Parser Text -> Parser [Text] -> Parser OpExpr
```
begriffs: Is there any more specific type than Text and [Text] we could use here?

steve-chavez:
Seems the enable/disable of the project in Travis did work, also if we all agree on merging this I would like to rebase and maintain the "migrate circle pg to 9.6" and "remove isnot/notin operators" commits, I think is worth preserving those.

begriffs:
The code seems good to me, and lots of tests. And `phraseto_tsquery` is only invoked if the client asks specifically for phrase matching, so it should remain generally compatible with pg 9.5 for the other full text searching cases.

Anyone have any other thoughts about this before we merge?

(I'll be sure to do a regular merge commit rather than a squash so it preserves the commits you create in your interactive rebase)

ruslantalpa:
Merge!

@steve-chavez @begriffs do you have a strong feeling against releasing often? 
Like 1-2 features ... release `4.x.0` or a few bugfixes ... release `4.3.x`?

begriffs:
Releasing more often is fine, especially patches because for minor versions I also advance the docs version to match.

begriffs:
I add a section on the master branch of docs, so the next tagged version there will pick it up. https://postgrest.com/en/latest/api.html#full-text-search

PR #967, opened 2017-09-07
Constrain jose to < 0.6

ilovezfs:
Fixes https://github.com/begriffs/postgrest/issues/965.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -66,7 +66,7 @@ library
                      , http-types
                      , insert-ordered-containers
                      , interpolatedstring-perl6
-                     , jose
+                     , jose < 0.6
                      , lens
                      , lens-aeson
                      , network-uri


ruslantalpa:
0.6 is very old
how come we are able to build locally on our macs?

ilovezfs:
@ruslantalpa probably because you're using Stack? See https://github.com/begriffs/postgrest/blob/master/stack.yaml#L8

Note that the latest release is failing across the board on Hackage: https://matrix.hackage.haskell.org/package/postgrest

begriffs:
@ilovezfs thanks for the fix and letting us know about the hackage failure.

How can we prevent this from happening in the future with other packages? A while ago I added [this line](https://github.com/begriffs/postgrest/blob/master/.circleci/config.yml#L91) to our continuous integration to try and detect whether regular cabal could solve the dependencies of the project without stack, but I guess it's failing to detect the problem. Can you come up with a modification to that line that will properly detect the problem? I'd like to see the test suite fail on master but pass on this PR.

begriffs:
Good catch. In stack.yaml we do pin a few versions:

```yaml
extra-deps:
  - configurator-ng-0.0.0.1
  - critbit-0.2.0.0
  - hasql-pool-0.4.1
  - hjsonpointer-1.1.1
  - hjsonschema-1.5.0.1
  - jose-0.5.0.3
  - Ranged-sets-0.3.0
  - protolude-0.2
```

Among those is the old jose. Perhaps this means there are some more constraints we should be adding to the cabal file.

ilovezfs:
@begriffs hmm ... well since the failure occurs while actually building `postgrest`, not while solving the dependencies, I'm not sure how to catch that with a dry-run. I think we'd actually need to build it.

Also the jose version in stack.yaml should probably be bumped to 0.5.0.5

ilovezfs:
>Perhaps this means there are some more constraints we should be adding to the cabal file.

I believe critbit and hjsonpointer are unpinned in postgrest.cabal.

ilovezfs:
@begriffs thanks for the merge! :heart:

PR #968, opened 2017-09-09
Dockerfile has typo in server proxy uri variable

begriffs:
I'm writing docs today and am adding a section to our installation section about running postgrest from docker. I encountered this:

```bash
$ docker pull postgrest/postgrest
$ docker run postgrest/postgrest
postgrest: ParseError "" "no such variable \"PGRST_SERVER_PROXY_URI\""
```

The problem was that the dockerfile was providing a default value for PGRST_SERVER_PROXY_UR*L* instead.

When we merge this our build process will automatically rebuild the `:latest` tag, right? I guess we'll have to release v4.3.1 as well... 😞


begriffs:
I pushed another commit to ignore the proxy uri if it is present but blank, as it is by default in Docker. Otherwise I get:

```bash
$ docker run -e PGRST_SERVER_PROXY_URI="" postgrest/postgrest
postgrest: FatalError {fatalErrorMessage = "Malformed proxy uri, a correct example: https://example.com:8443/basePath"}
```

steve-chavez:
Why not just manually replace the `v0.4.3.0/latest` docker images with the `PGRST_SERVER_PROXY_URI` fix now, is not working anyway and I see this more as a flaw of the build process, I think it would be better to leave the version bumping for changes in the PostgREST application codebase. After that we could include the `Treat blank proxy uri as missing` commit in v0.4.4.0 with the other recently merged PR's.

ruslantalpa:
yes, rebuild the docker images manually (both tags, just like the build script) and push. no need for a new version

begriffs:
What about my change in Config.hs though?

ruslantalpa:
that can be a 4.3.1, along with other small changes, it's too small to merit a .1 i think.

rvernica:
I just pulled the latest image from Docker Hub and followed the [instructions](https://postgrest.com/en/v4.3/install.html#docker) but I still get the *Malformed proxy uri* error. The Docker Hub image seems to have been pushed 4 days ago. This patch was merged 6 days ago. I guess this patch did not make it into the Docker image.

abbapappa:
I tried with `postgrest/postgrest:latest` (`cd5df30b4a7e435aafe9abeac0e4158b98df03182de5bce93823c843293223c1`), same result. When can this be merged into master?

ruslantalpa:
it is in master, :latest does not mean master, it means the latest release, which is 4.3

abbapappa:
Sorry, my bad. I did not see it was already merged. I'll await the next release :).

PR #970, opened 2017-09-12
Fix some OpenAPI issues

steve-chavez:


ruslantalpa:
Good PR (fixing all the outstanding issues in one go, didn't look at the code yet though)

ruslantalpa:
code looks good

test/Feature/StructureSpec.hs
```diff
@@ -78,46 +85,53 @@ spec = do
...
-                    "description": "child_entities name comment",
-                    "format": "text",
-                    "type": "string"
-                  },
-                  "parent_id": {
-                    "description": "Note:\nThis is a Foreign Key to `entities.id`.<fk table='entities' column='id'/>",
-                    "format": "integer",
-                    "type": "integer"
+      it "includes definitions to tables" $ do
+          r <- simpleBody <$> get "/"
```
begriffs: As long as your changing indentation, these lines should be indented 2 beneath the `it` statement rather than 4.

src/PostgREST/DbStructure.hs
```diff
@@ -100,74 +101,94 @@ decodeSynonyms cols =
...
-    JOIN pg_namespace pn ON pn.oid = p.pronamespace
-    JOIN pg_type t ON t.oid = p.prorettype
-    JOIN pg_namespace tn ON tn.oid = t.typnamespace
-    LEFT JOIN pg_class comp ON comp.oid = t.typrelid
-    LEFT JOIN pg_catalog.pg_description as d on d.objoid = p.oid
-  WHERE  pn.nspname = $1|]
+accessibleProcs :: H.Query Schema (M.HashMap Text ProcDescription)
+accessibleProcs = H.statement sql (HE.value HE.text) decodeProcs True
+  where
+    sql = [q|
```
begriffs: Maybe `allProcs` and `accessibleProcs` could share a chunk of sql? The latter seems to have the same query except with an extra WHERE clause added. It could concatenate the extra clauses to the shared sql?

test/Feature/NonexistentSchemaSpec.hs
```diff
@@ -0,0 +1,15 @@
+module Feature.NonexistentSchemaSpec where
```
begriffs: I like that you added a test for this along with the fix!

test/Feature/StructureSpec.hs
```diff
@@ -29,7 +29,14 @@ spec = do
               (acceptHdrs "application/openapi+json") ""
         `shouldRespondWith` 415
 
-    describe "table" $
+    it "includes postgrest.com current version api docs" $ do
+        r <- simpleBody <$> get "/"
```
begriffs: Indentation a little too deep here.

begriffs:
Three fixes in one stroke, very nice.

How about some entries in the changelog?

steve-chavez:
Changelog updated, corrections added.

Issue #955, opened 2017-08-27
JWT verification fails with audience claim


russelldavies:
In the [jose](https://hackage.haskell.org/package/jose) library if the `aud` claim is present in the JWT then during verification it validates this claim.

Currently, there is no way to disable this or set the audience claim so verification is unsuccessful. Specifically, the `JWTNotInAudience` message is generated during JWT verification.

begriffs:
Can you provide more information please, such as the http request and error message?

russelldavies:
@begriffs I accidentally tabbed and hit return and the form posted before I had a chance to finish typing. I've edited it now.

begriffs:
Oh man that sounds bad, so if `aud` is present then it just lets any claim through??

russelldavies:
Not at all, it just fails with a `JWTNotInAudience` error. In the `defaultJWTValidationSettings` unless the `aud` claim is set to what is in the JWT `aud` claim then the JWT is rejected.

So a solution is to either disable this check or to instruct the jose library what the audience to check should be, through the postgrest configuration file.

will-ob:
^^ So,
```
if `aud`
  verify against value in pg-config
```
That sounds like it wouldn't break anything for existing deployments.

begriffs:
Yeah sounds straightforward. I'll open a PR to implement this.

statik:
I've managed to configure postgrest 0.4.3 with a JWK retrieved from keycloak, and specify a custom mapper in keycloak so that I can force a "role" claim to be included in the JWT to satisfy postgrest. However, I'm now running into the error JWTNotInAudience from postgrest. I have not been able to find any way to configure keycloak to leave the "aud" claim out of the JWT, so I think the workaround for Auth0 isn't working for keycloak.

I'm pretty newbie to haskell but interested in helping with this if the PR is not already written as I have already invested a bunch of time figuring out keycloak and this seems to be the last missing piece to get postgrest working with keycloak.

marsouin:
@statik this is great, I have the same issue, can I update Postgrest with your work?

statik:
@marsouin this was just merged to master an hour ago, I am currently running with a version I built from source until the next release comes out and it's working ok for me.

marsouin:
Oh awesome! Any tips on building it? I'm very new to Haskell...

russelldavies:
@marsouin The [build instructions](https://postgrest.com/en/v4.3/install.html#build-from-source) cover this.

[...]

PR #975, opened 2017-09-15
Allow configurable audience claim

statik:
This is intended to  address #955.

I consulted https://github.com/frasertweedale/hs-jose/issues/24 when working on this. 

Things to contemplate: Would we want to support something like a wildcard to allow all audience claims through? Or a list of multiple claims? I don't know enough about JWT to know if that is a bad idea, but I know for my use case it seems like I would want to allow all audiences because I expect to have many keycloak realms that are not statically defined enough to list in a config file (one per organization that is a customer of my app).

I think it makes sense to start with the simplest thing, and then if people show up with real world use cases of needing to use multiple audience claims we can extend the support.


begriffs:
@statik this is great, I'm so happy you're jumping in to make a Haskell contribution. Let me try building this locally and find what we need to change to make it build.

src/PostgREST/Auth.hs
```diff
@@ -39,16 +39,16 @@ data JWTAttempt = JWTInvalid JWTError
...
-jwtClaims _ "" = return $ JWTClaims M.empty
-jwtClaims secret payload =
+jwtClaims :: Maybe JWK -> Text -> BL.ByteString  -> IO JWTAttempt
+jwtClaims _ "" "" = return $ JWTClaims M.empty
+jwtClaims secret audience payload =
   case secret of
     Nothing -> return JWTMissingSecret
     Just jwk -> do
-      let validation = defaultJWTValidationSettings
+      let validation = set audiencePredicate (==  audience) defaultJWTValidationSettings
```
begriffs: As you pointed out, the error says we're trying to compare StringOrURI with Text, so we want to find a way to convert `audience` so we can compare it.

The first thing to note in debugging this issue is that our cabal file specifies
[jose < 0.6](https://github.com/begriffs/postgrest/blob/master/postgrest.cabal#L69), so we need to be sure when looking at jose code on hackage we look at v0.5.0.5.

My next step was scanning the docs on hackage for a function to convert the types. This one turns up:

https://hackage.haskell.org/package/jose-0.5.0.5/docs/Crypto-JWT.html#v:fromString

Making this change fixes the error you encountered:

```haskell
let validation = set audiencePredicate (==  fromString audience) defaultJWTValidationSettings
```

src/PostgREST/Config.hs
```diff
@@ -172,6 +174,7 @@ readOptions = do
         |## (use "@filename" to load from separate file)
         |# jwt-secret = "foo"
         |# secret-is-base64 = false
+        |# jwt-aud = "yourkeycloakrealm"
```
begriffs: Oh I almost forgot about changing this output, good idea. Maybe we should use a less keycloak-specific example as a string literal?

src/PostgREST/Auth.hs
```diff
@@ -39,16 +39,16 @@ data JWTAttempt = JWTInvalid JWTError
...
 
 {-|
-  Receives the JWT secret (from config) and a JWT and returns a map
+  Receives the JWT secret and audience (from config) and a JWT and returns a map
   of JWT claims.
 -}
-jwtClaims :: Maybe JWK -> BL.ByteString -> IO JWTAttempt
-jwtClaims _ "" = return $ JWTClaims M.empty
-jwtClaims secret payload =
+jwtClaims :: Maybe JWK -> Text -> BL.ByteString  -> IO JWTAttempt
```
begriffs: The audience argument should be `Maybe Text` rather than `Text` though, which will change our logic below slightly.

src/PostgREST/Auth.hs
statik: Thank you very much for the patient help! This was exactly what I needed, and I had been looking at the wrong version of jose docs and getting very confused.

src/PostgREST/Config.hs
statik: Good call, I'll just refer to audience.

src/PostgREST/Auth.hs
statik: I've moved the Maybe handling into the config parser and having it default to "", which simplifies the rest of the handling nicely.

statik:
When I get back from bringing kids to violin lessons I'll work on updating the tests so they pass.

statik:
So I believe the existing tests will pass now and I've manually tested this with and without audience in the config. @begriffs what degree of unit tests would you like to see?

src/PostgREST/Config.hs
```diff
@@ -112,6 +113,7 @@ readOptions = do
           <*> (fromMaybe 3000 . join . fmap coerceInt <$> C.key "server-port")
           <*> (fmap encodeUtf8 . mfilter (/= "") <$> C.key "jwt-secret")
           <*> (fromMaybe False . join . fmap coerceBool <$> C.key "secret-is-base64")
+          <*> (fromMaybe "" . mfilter (/= "") <$> C.key "jwt-aud")
```
begriffs: I'm thinking the mfilter is not necessary here since all it would do is convert `Just ""` to `Nothing` which `fromMaybe` would return as `""`.

ruslantalpa:
Just make sure empty string is an acceptable value so that we dont break docker again:)

begriffs:
To add a test for the audience claim, you can make a new little test file that roughly copies the pattern in [BinaryJwtSecretSpec.hs](https://github.com/begriffs/postgrest/blob/master/test/Feature/BinaryJwtSecretSpec.hs). You can add a function in SpecHelper that constructs a config file with the audience filled in, similar to [this](https://github.com/begriffs/postgrest/blob/master/test/SpecHelper.hs#L92-L96). Finally in Main.hs create an instance of the application with that config, like [here](https://github.com/begriffs/postgrest/blob/cf19ad036926b31f9982e80f79a14e7cd93eaba3/test/Main.hs#L47), and run it like [so](https://github.com/begriffs/postgrest/blob/cf19ad036926b31f9982e80f79a14e7cd93eaba3/test/Main.hs#L70-L72).

statik:
I've added the unit test and removed the mfilter. I believe this is good to go!

test/Feature/AudienceJwtSecretSpec.hs
```diff
@@ -0,0 +1,21 @@
...
+import Network.HTTP.Types
+
+import SpecHelper
+import Network.Wai (Application)
+
+import Protolude hiding (get)
+-- }}}
+
+spec :: SpecWith Application
+spec = describe "server started with binary JWT secret" $
```
begriffs: This description threw me off for a second because its emphasis seems to be about the binary secret rather than the audience claim.

begriffs:
Great work, thanks!

So in short, when someone sets `jwt-aud` in the config, then only requests containing that aud claim (and correct signing of course) will succeed? What will happen if you send a JWT with no aud claim?

src/PostgREST/Config.hs
```diff
@@ -172,6 +174,7 @@ readOptions = do
         |## (use "@filename" to load from separate file)
         |# jwt-secret = "foo"
         |# secret-is-base64 = false
+        |# jwt-aud = "your_audience_claim"
```
begriffs: We'll want to add a new variable to the Docker setup as well, much like how PGRST_JWT_SECRET is handled in the [Dockerfile](https://github.com/begriffs/postgrest/blob/master/docker/Dockerfile#L34) and [postgrest.conf](https://github.com/begriffs/postgrest/blob/master/docker/postgrest.conf#L10). Those two files are used for building our images on Dockerhub.

test/Feature/AudienceJwtSecretSpec.hs
statik: I'll revise to be more clear what this is testing, and add decoded versions of the JWT in comments so the difference is clear to the reader.

src/PostgREST/Config.hs
statik: Good catch! I updated those files as well as app.json.

test/Feature/AudienceJwtSecretSpec.hs
statik: I also extended the unit test to show that if jwt-aud is configured, and the JWT presented does not have an `aud` claim, that the validation will still succeed. Seems like we are great for backward compatibility.

docker/postgrest.conf
```diff
@@ -9,6 +9,7 @@ server-port = "$(PGRST_SERVER_PORT)"
 server-proxy-uri = "$(PGRST_SERVER_PROXY_URI)"
 jwt-secret = "$(PGRST_JWT_SECRET)"
 secret-is-base64 = "$(PGRST_SECRET_IS_BASE64)"
+jwt-aud = "$(PGRST_JWT_AUD"
```
begriffs: Oops, a missing paren.

statik: Doh! Fixing.

docker/postgrest.conf
```diff
@@ -9,6 +9,7 @@ server-port = "$(PGRST_SERVER_PORT)"
 server-proxy-uri = "$(PGRST_SERVER_PROXY_URI)"
 jwt-secret = "$(PGRST_JWT_SECRET)"
 secret-is-base64 = "$(PGRST_SECRET_IS_BASE64)"
+jwt-aud = "$(PGRST_JWT_AUD")
```
russelldavies: Typo here, it should be `jwt-aud = "$(PGRST_JWT_AUD)"`

statik: Thank you! Fixed.

begriffs:
Thanks @statik 

Add a line to CHANGELOG.md in the PR and I'll merge this!

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #907, Ability to embed using a specific relation when there are multiple between tables - @ruslantalpa
 - #930, Split table comment on newline to get OpenAPI operation summary and description - @daurnimator
 - #938, Support for range operators - @russelldavies
+- #955, Support configurable aud claim - @statik
```
begriffs: Actually could you move this up a little higher into the "Added" section under "Unreleased?" rather than "0.4.3.0"? Your change will become part of the next release.

docker/Dockerfile
```diff
@@ -33,6 +33,7 @@ ENV PGRST_DB_URI= \
     PGRST_SERVER_PROXY_URI= \
     PGRST_JWT_SECRET= \
     PGRST_SECRET_IS_BASE64=false \
+    PGRST_JWT_AUD=\
```
begriffs: Actually since I made my other comment about the changelog adjustment let me throw this in too: can you add a space between the equals sign and the backslash?

Sorry to be annoying, the whitespace probably doesn't matter, I just felt kind of OCD about it. 😁

statik:
@begriffs updated as requested, and rebased on master to address merge conflict. Thanks!

PR #978, opened 2017-09-16
Remove the need for pk in &select for parent embed

steve-chavez:
**Edit**: Omit this proposal, no need to upgrade PostgreSQL version, see my comment below.

Main idea to solve this was to include the primary keys on the LEFT JOIN subquery regardless of whether the user requested them or not, and then remove them from the output by using PostgreSQL jsonb `-` operator.

So now when requesting:
```http
GET /projects?select=*,client{name}
```
This query is generated:
```sql
WITH pg_source AS (
   SELECT "test"."projects".*,
          to_jsonb("clients_client".*) - 'id' AS "client"
   FROM "test"."projects"
   LEFT JOIN (
     SELECT "test"."clients"."id",
            "test"."clients"."name"
     FROM "test"."clients"
   ) AS "clients_client" ON "clients_client"."id" = "projects"."client_id")
SELECT NULL AS total_result_set,
       pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header,
       coalesce(array_to_json(array_agg(row_to_json(_postgrest_t))), '[]')::character varying AS body
FROM (SELECT * FROM pg_source) _postgrest_t
```
The cost of this new query remains the same as the previous one in the tests I've made. Also requests that included a rename of the pk:
```http
GET /projects?select=name,client{clientId:id,name}
```
Were not working previously and now this is fixed.

The drawback of this solution is that the used `-` jsonb operator is only available starting from PostgreSQL 9.5, so the required version for PostgREST would have to be increased to 9.5 for this to work.

Maybe it's not worth to increase the version for just this fix, however I think with a similarly crafted query I could modify the `Many/Child` embed queries and as noted in https://github.com/begriffs/postgrest/issues/411#issuecomment-329324791 the query costs for those are really high now, if I can dramatically decrease these costs I think then it'll be worth the jump to 9.5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #996, Fix embedded column conflicts table name - @grotsev
 - #974, Fix RPC error when function has single OUT param - @steve-chavez
 - #1021, Reduce join size in allColumns for faster program start - @nextstopsun
+- #411, Remove the need for pk in &select for parent embed - @steve-chavez
 
 ## [0.4.3.0] - 2017-09-06
 
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -31,7 +31,6 @@ import           PostgREST.ApiRequest   ( ApiRequest(..)
 import           PostgREST.Error           (apiRequestError)
 import           PostgREST.Parsers
 import           PostgREST.RangeQuery      (NonnegRange, restrictRange)
-import           PostgREST.QueryBuilder (getJoinFilters, sourceCTEName)
 import           PostgREST.Types
 
 import           Protolude                hiding (from, dropWhile, drop)
@@ -194,7 +193,7 @@ addJoinFilters :: Schema -> ReadRequest -> Either ApiRequestError ReadRequest
 addJoinFilters schema (Node node@(query, nodeProps@(_, relation, _, _)) forest) =
   case relation of
     Just Relation{relType=Root} -> Node node  <$> updatedForest -- this is the root node
-    Just Relation{relType=Parent} -> Node node <$> updatedForest
+    Just rel@Relation{relType=Parent} -> Node (augmentQuery rel, nodeProps) <$> updatedForest
     Just rel@Relation{relType=Child} -> Node (augmentQuery rel, nodeProps) <$> updatedForest
     Just rel@Relation{relType=Many, relLTable=(Just linkTable)} ->
       let rq = augmentQuery rel in
@@ -204,6 +203,21 @@ addJoinFilters schema (Node node@(query, nodeProps@(_, relation, _, _)) forest) 
     updatedForest = mapM (addJoinFilters schema) forest
     augmentQuery rel = foldr addFilterToReadQuery query (getJoinFilters rel)
     addFilterToReadQuery flt rq@Select{where_=lf} = rq{where_=addFilterToLogicForest flt lf}::ReadQuery
+
+getJoinFilters :: Relation -> [Filter]
+getJoinFilters (Relation t cols ft fcs typ lt lc1 lc2) =
+  case typ of
+    Child  -> zipWith (toFilter tN ftN) cols fcs
+    Parent -> zipWith (toFilter tN ftN) cols fcs
+    Many   -> zipWith (toFilter tN ltN) cols (fromMaybe [] lc1) ++ zipWith (toFilter ftN ltN) fcs (fromMaybe [] lc2)
+    Root   -> undefined --error "undefined getJoinFilters"
+  where
+    s = if typ == Parent then "" else tableSchema t
+    tN = tableName t
+    ftN = tableName ft
+    ltN = fromMaybe "" (tableName <$> lt)
+    toFilter :: Text -> Text -> Column -> Column -> Filter
+    toFilter tb ftb c fc = Filter (colName c, Nothing) (OpExpr False (Join (QualifiedIdentifier s tb) (ForeignKey fc{colTable=(colTable fc){tableName=ftb}})))
 
 addFiltersOrdersRanges :: ApiRequest -> Either ApiRequestError (ReadRequest -> ReadRequest)
 addFiltersOrdersRanges apiRequest = foldr1 (liftA2 (.)) [
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -14,12 +14,10 @@ module PostgREST.QueryBuilder (
     callProc
   , createReadStatement
   , createWriteStatement
-  , getJoinFilters
   , pgFmtIdent
   , pgFmtLit
   , requestToQuery
   , requestToCountQuery
-  , sourceCTEName
   , unquoted
   , ResultsWithCount
   , pgFmtEnvVar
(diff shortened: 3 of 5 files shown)


steve-chavez:
Actually I think I've come up with a better query that doesn't need that jsonb `-` operator and instead just use a  `LEFT JOIN LATERAL`. I'm working on a follow up commit with this modification.

steve-chavez:
So, nevermind the first query this one is better, same request:
```http
GET /projects?select=*,client{name}
```
Now generates this query:
```sql
WITH pg_source AS (
  SELECT "test"."projects".*,
          row_to_json("clients_client".*) AS "client"
  FROM "test"."projects"
  LEFT JOIN LATERAL (
    SELECT "test"."clients"."name"
    FROM "test"."clients"
    WHERE "clients"."id" = "projects"."client_id" ) AS "clients_client" ON TRUE)
SELECT NULL AS total_result_set,
       pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header,
       coalesce(array_to_json(array_agg(row_to_json(_postgrest_t))), '[]')::character varying AS BODY
FROM (SELECT * FROM pg_source) _postgrest_t;
```
The cost remains the same as the parent embed query currently in master.

Also moved some functions because I think that the `DbRequestBuilder` module shouldn't depend on the `QueryBuilder` module, this I believe makes the `Parsers -> DbRequestBuilder -> QueryBuilder` progression more clear.

PR ready for review now.

ruslantalpa:
This type of pr is less about haskell and more about sql.

We should take extreme care with these type of changes.

I suggest a basic test schema with tables 1M+ records each.

Then take this query (and the old one) and compare them with pgbench.

The query should request about 100 random rows and embed the parent, and run each test 5+ minutes.

The explain/cost is not always exact so we need real world tests

It would be great to have the schema as a sql/gist file so that others (probabil just us anyway :) can run the same test 

We should not rush this pr since it has a big potential to break/slow things down

begriffs:
Here's a thought - what if we put the sql code into separate files, and embed them into the Haskell code at compile time with [file-embed](https://hackage.haskell.org/package/file-embed-0.0.10.1/docs/Data-FileEmbed.html)? Then a benchmark could load the sql too, and there'd be no danger that the queries under test would drift from those used by the code since they would be loaded from the same place.

Maybe more trouble than it's worth, but putting it out there.

ruslantalpa:
@begriffs there is no sql to put into files, the queries are generated, all you'd be able to put into files would be words like select/where/join ...

begriffs:
Ah of course. Nevermind. 😁

steve-chavez:
Did the tests, here is the gist with the schema and the old/new queries:  https://gist.github.com/steve-chavez/f4d3b7637e8f8075bfc65fd9a1429e8a. Queries were generated by exposing the schema through PostgREST and doing a:
```http
GET /items?select=id,name,parent_items{id,name}&id=lte.100
```

My results are:
- Explain costs:
  - parent_join_test_query_old.sql: 363.85..363.87
  - parent_join_test_query_new.sql: 363.85..363.87

- pgbench -T 300 example -f parent_join_test_query_old.sql
```rust
    scaling factor: 1
    query mode: simple
    number of clients: 1
    number of threads: 1
    duration: 300 s
    number of transactions actually processed: 592364
    latency average = 0.506 ms
    tps = 1974.546292 (including connections establishing)
    tps = 1974.554440 (excluding connections establishing)
```
- pgbench -T 300 example -f parent_join_test_query_new.sql
```rust
  scaling factor: 1
  query mode: simple
  number of clients: 1
  number of threads: 1
  duration: 300 s
  number of transactions actually processed: 588589
  latency average = 0.510 ms
  tps = 1961.962470 (including connections establishing)
  tps = 1961.970978 (excluding connections establishing)
```

ruslantalpa:
your tests are running the same query each time since the id is hardcoded here
https://gist.github.com/steve-chavez/f4d3b7637e8f8075bfc65fd9a1429e8a#file-parent_join_test_query_new-sql-L11 so the results may be misleading 

try changing the tests to be similar to this one, which generates a different query each time (adjust as needed)

```
\setrandom cid 1 9800
BEGIN ISOLATION LEVEL READ COMMITTED READ ONLY;
select * from clients where id > :cid and id <= (:cid + 100);
COMMIT;
```

steve-chavez:
Updated the gist with the random variable, new results:
- pgbench -T 300 example -f parent_join_test_query_old.sql
```rust
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
duration: 300 s
number of transactions actually processed: 182995
latency average = 1.639 ms
tps = 609.982469 (including connections establishing)
tps = 609.984661 (excluding connections establishing)
```
- pgbench -T 300 example -f parent_join_test_query_new.sql
```rust
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
duration: 300 s
number of transactions actually processed: 182925
latency average = 1.640 ms
tps = 609.747136 (including connections establishing)
tps = 609.749533 (excluding connections establishing)
```

begriffs:
@ruslantalpa are you happy with the new measurements?

I like that `getQueryParts` got a little shorter.

ruslantalpa:
@begriffs the measurements are promising and the code looks ok, however i would like to drag out this PR for longer for the following reasons.

- i have not had the time yet to try the tests myself although they look fine
- unlike other PRs, this one changes a core part of postgrest, the shape of the query, so it needs extreme care, maybe we should event have a 4.4-RC and let ppl test it out on their projects and see if anything breaks
- i would like (hope) more ppl compile this and do the speed test.
- it's not that urgent, the bug it fixes has a simple workaround and there are no speed improvements it brings.

begriffs:
Side note: after I'm done with the current book I'm studying (SQL for Smarties), I want to do a deep dive into profiling and performance measurement. Postgrest would be a great project to try practicing the methodology from these books:

* https://www.goodreads.com/book/show/430782.The_Art_of_Computer_Systems_Performance_Analysis
* https://www.goodreads.com/book/show/18058001-systems-performance

ruslantalpa:
I guess we could merge this and do a 4.4-RC

steve-chavez:
Right now there are many fixes and not potentially breaking changes in master, and the PR about the parametrized query in RPC also changes the core PostgREST queries. So if were gonna do the RC thing I think it would be better to do that for just those two PRs and in a v0.5.0.0-RC, after the v0.4.4.0 release.

begriffs:
OK, so how about we finish up the "safe" PRs, release v4.4, then get these less safe ones into master.

ruslantalpa:
All the current PR's are safe, including #1007 since it changes unreleased functionality.
This week's time is for postgrest so all opened PR's could be finished and merged then there could be a `4.4-RC`

steve-chavez:
For me the RC version is not necessary, I don't see how the change in this PR(particularly the `LEFT JOIN LATERAL`) would break clients and if there's a certain scenario where a failure is suspected to happen it should be a explicit test.

However if were going to do an RC version it would be better not to mix it with many of the fixes already available in master, so.. by following the logic:
>  this one changes a core part of postgrest, the shape of the query, so it needs extreme care

Then #1025 should also be part of an RC I suggest leaving it for v0.5.0.0-RC, then #1022 and #1024 could be part of the v0.4.4.0.

ruslantalpa:
#1025 is actually safer then lateral join from #978 but i think both are ok for 4.4.0 since all the tests pass. RC is probably overkill as you said. let's just finish all, merge and release

steve-chavez:
Ok then. I've rebased, now this is ready for another review @begriffs.

begriffs:
The `LEFT JOIN LATERAL … ON TRUE` thing is kind of mysterious to me, but I verified that the new tests fail on master and pass on this branch. So looks good to me.

PR #980, opened 2017-09-20
Fix RPC GET filters bugs with not/{plain,phrase}fts operators

steve-chavez:
Fixes the bugs mentioned in https://github.com/begriffs/postgrest/pull/946#issuecomment-330617307.

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -39,7 +39,8 @@ import           PostgREST.Types           ( QualifiedIdentifier (..)
                                            , ContentType(..)
                                            , ApiRequestError(..)
                                            , toMime
-                                           , operators)
+                                           , operators
+                                           , FtsMode(..))
 import           Data.Ranged.Ranges        (Range(..), rangeIntersection, emptyRange)
 import qualified Data.CaseInsensitive      as CI
 import           Web.Cookie                (parseCookiesText)
@@ -141,8 +142,16 @@ userApiRequest schema req reqBody
       ActionInvoke{isReadOnly=True} -> partition (liftM2 (||) (isEmbedPath . fst) (hasOperator . snd)) flts
       _ -> (flts, [])
   flts = [ (toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn ["order", "limit", "offset", "and", "or"] k) ]
-  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False $ (<> ".") <$> M.keys operators
+  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False ((<> ".") <$> (M.keys operators ++ ["not", show Plain, show Phrase]))
+                    || hasLanguageFts val
   isEmbedPath = T.isInfixOf "."
+  -- handle "?tsv=english.fts.possible" case
+  hasLanguageFts val = case T.splitOn "." val of
+    [_, "fts", _] -> True
+    [_, "fts"]    -> True
+    [_, "@@", _]  -> True -- TODO: '@@' deprecated
+    [_, "@@"]     -> True
+    _             -> False
   isTargetingProc = fromMaybe False $ (== "rpc") <$> listToMaybe path
   payload =
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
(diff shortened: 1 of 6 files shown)


src/PostgREST/QueryBuilder.hs
```diff
@@ -429,9 +429,9 @@ pgFmtFilter table (Filter fld (OpExpr hasNot oper)) = notOp <> " " <> case oper
 
    Fts mode lang val ->
      pgFmtFieldOp "fts" <> " " <> case mode of
-       Normal -> "to_tsquery("
-       Plain  -> "plainto_tsquery("
-       Phrase -> "phraseto_tsquery("
+       Just Plain  -> "plainto_tsquery("
+       Just Phrase -> "phraseto_tsquery("
+       Nothing     -> "to_tsquery("
```
begriffs: It seems counter-intuitive to model this as `Maybe FtsMode` rather than simply `FtsMode` since even if the user omits an explicit mode we're still choosing some kind of a mode.

src/PostgREST/Parsers.hs
```diff
@@ -146,9 +146,9 @@ pOpExpr pSVal pLVal = try ( string "not" *> pDelimiter *> (OpExpr True <$> pOper
       <|> pFts
       <?> "operator (eq, gt, ...)"
     pFts = do
-      mode <- option Normal $
-              try (string "phrase" *> pDelimiter *> pure Phrase)
-          <|> try (string "plain" *> pDelimiter *> pure Plain)
+      mode <- optionMaybe $
```
begriffs: After reading https://www.postgresql.org/docs/9.6/static/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES I'm wondering whether `to_tsquery` should be the default when no mode is specified, since

> to_tsquery offers access to more features than either plainto_tsquery or phraseto_tsquery, but it is less forgiving about its input

Too bad changing the default would cause a potentially breaking change for people who rely on our current fts behavior. Maybe I'm wrong but the plainto_tsquery function seems more likely to be useful for the strings people might pass via the url.

src/PostgREST/ApiRequest.hs
```diff
@@ -141,8 +142,16 @@ userApiRequest schema req reqBody
...
       _ -> (flts, [])
   flts = [ (toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn ["order", "limit", "offset", "and", "or"] k) ]
-  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False $ (<> ".") <$> M.keys operators
+  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False ((<> ".") <$> (M.keys operators ++ ["not", show Plain, show Phrase]))
+                    || hasLanguageFts val
   isEmbedPath = T.isInfixOf "."
+  -- handle "?tsv=english.fts.possible" case
+  hasLanguageFts val = case T.splitOn "." val of
+    [_, "fts", _] -> True
+    [_, "fts"]    -> True
```
steve-chavez: Don't really like this workaround, in hindsight I would've preferred the `fts` syntax to be like:
```http
-- fts with language
GET /tsearch?ts_vector=fts(french).amusant
-- plainto_tsquery
GET /tsearch?ts_vector=plfts.The%20Fat%20Cats
GET /tsearch?ts_vector=plfts(english).The%20Fat%20Cats
-- phraseto_tsquery
GET /tsearch?ts_vector=phfts.The%20Fat%20Rats
GET /tsearch?ts_vector=phfts(english).The%20Fat%20Rats
```
That would make this workaround unnecessary, also it's more consistent with previous operators and the syntax is shorter.

steve-chavez: Since there hasn't been any release yet, how about if I change the `fts` syntax to the above?

ruslantalpa: what previous opeartors?
how is `fts(french)` shorter then `fts.french`

i don't remember how exactly is the format you implemented but i remember it's modeled after how `not` changes the behaviour of like. `eq.1 vs no.eq.1` and here we might have `fts.test vs french.fts.test`
is it now like `fts.french.test`?

steve-chavez: Right now we have `not.plain.english.fts.impossible`, with the syntax I mentioned above this would be like `not.plfts(english).impossible`, so the syntax would be shorter for the `phrase/plain` fts modes.

The `not.french.fts.test` would be like `not.fts(french).test`, no gain in length there.

ruslantalpa: this pr needs to be merged to fix the current master, but the format change can be discussed before a release happens

steve-chavez: Ok. Agreed.

PR #986, opened 2017-09-26
Add ability to set custom http headers and http status

steve-chavez:
PR adds the following:

- When invoking an rpc is possible to set http headers by doing:
```sql
  SET LOCAL "response.headers" = '[{"X-Test": "Domain=.server.com; Secure; Version=1"}, {"X-Test-2": "TimeoutLoggedIn=login; Version=1"}]';
```
This gets mapped to:
```http
HTTP/1.1 200 OK
Server: postgrest/0.4.3.0
X-Test: Domain=.server.com; Secure; Version=1
X-Test-2: TimeoutLoggedIn=login; Version=1
```
The value for the `response.headers` must be a json array of objects with a single key and a string value for the reason mentioned in https://github.com/begriffs/postgrest/pull/664#issuecomment-231635405, if a specified value doesn't comply with this a `500` response is sent to the client.

- As proposed in https://github.com/begriffs/postgrest/issues/917#issuecomment-320480665, the http status can be customized with:
```sql
RAISE sqlstate 'PT402' using message = 'Payment Required',
           detail = 'Quota exceeded',
           hint = 'Upgrade your plan';
```
I added the status message option, the above gets mapped to a response of:
```http
HTTP/1.1 402 Payment Required
Server: postgrest/0.4.3.0

{"hint":"Upgrade your plan","details":"Quota exceeded"}
```

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -84,11 +84,8 @@ postgrest conf refDbStructure pool worker =
                   (iTarget apiRequest) (iAction apiRequest)
             response <- P.use pool $ HT.transaction HT.ReadCommitted txMode handleReq
             return $ either (pgError authed) identity response
-        when (isResponse503 response) worker
+        when (responseStatus response == status503) worker
         respond response
-
-isResponse503 :: Response -> Bool
-isResponse503 resp = statusCode (responseStatus resp) == 503
 
 transactionMode :: DbStructure -> Target -> Action -> H.Mode
 transactionMode structure target action =
@@ -252,6 +249,7 @@ app dbStructure conf apiRequest =
                          singular paramsAsSingleObject
                          (contentType == CTTextCSV)
                          (contentType == CTOctetStream) _isReadOnly bField
+                         (pgVersion dbStructure)
               let (tableTotal, queryTotal, body, jsonHeaders) =
                     fromMaybe (Just 0, 0, "[]", "[]") row
                   (status, contentRange) = rangeHeader queryTotal tableTotal
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -6,7 +6,7 @@ module PostgREST.App (
 ) where
 
 import           Control.Applicative
-import           Data.Aeson                (toJSON)
+import           Data.Aeson                (toJSON, eitherDecode)
 import qualified Data.ByteString.Char8     as BS
 import           Data.Maybe
 import           Data.IORef                (IORef, readIORef)
@@ -44,7 +44,7 @@ import           PostgREST.DbRequestBuilder( readRequest
 import           PostgREST.Error           ( simpleError, pgError
                                            , apiRequestError
                                            , singularityError, binaryFieldError
-                                           , connectionLostError
+                                           , connectionLostError, gucHeadersError
                                            )
 import           PostgREST.RangeQuery      (allRange, rangeOffset)
 import           PostgREST.Middleware
@@ -248,18 +248,22 @@ app dbStructure conf apiRequest =
                   singular = contentType == CTSingularJSON
                   paramsAsSingleObject = iPreferSingleObjectParameter apiRequest
               row <- H.query () $
-                callProc qi prms returnsScalar q cq topLevelRange shouldCount
+                callProc qi prms returnsScalar q cq shouldCount
                          singular paramsAsSingleObject
                          (contentType == CTTextCSV)
                          (contentType == CTOctetStream) _isReadOnly bField
-              let (tableTotal, queryTotal, body) =
-                    fromMaybe (Just 0, 0, "[]") row
+              let (tableTotal, queryTotal, body, jsonHeaders) =
+                    fromMaybe (Just 0, 0, "[]", "[]") row
                   (status, contentRange) = rangeHeader queryTotal tableTotal
-              if singular && queryTotal /= 1
-                then do
-                  HT.condemn
-                  return $ singularityError (toInteger queryTotal)
-                else return $ responseLBS status [toHeader contentType, contentRange] (toS body)
+                  decodedHeaders = first toS $ eitherDecode $ toS jsonHeaders :: Either Text [GucHeader]
+              case decodedHeaders of
+                Left _ -> return gucHeadersError
+                Right hs ->
+                  if singular && queryTotal /= 1
+                    then do
+                      HT.condemn
+                      return $ singularityError (toInteger queryTotal)
+                    else return $ responseLBS status ([toHeader contentType, contentRange] ++ toHeaders hs) (toS body)
 
         (ActionInspect, TargetRoot, Nothing) -> do
           let host = configHost conf
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -10,6 +10,7 @@ module PostgREST.Error (
 , binaryFieldError
 , connectionLostError
 , encodeError
+, gucHeadersError
 ) where
 
 import           Protolude
@@ -78,6 +79,11 @@ binaryFieldError :: Response
 binaryFieldError =
   simpleError HT.status406 [] (toS (toMime CTOctetStream) <>
   " requested but a single column was not selected")
+
+gucHeadersError :: Response
+gucHeadersError =
+ simpleError HT.status500 []
+ "response.headers guc must be a JSON array composed of objects with a single key and a string value"
 
 connectionLostError :: Response
 connectionLostError =
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -22,6 +22,7 @@ import           Network.HTTP.Types.Header
 import qualified Network.HTTP.Types.Status as HT
 import           Network.Wai               (Response, responseLBS)
 import           PostgREST.Types
+import           Text.Read                 (readMaybe)
 
 apiRequestError :: ApiRequestError -> Response
 apiRequestError err =
@@ -109,11 +110,17 @@ instance JSON.ToJSON P.UsageError where
   toJSON (P.SessionError e) = JSON.toJSON e -- H.Error
 
 instance JSON.ToJSON H.Error where
-  toJSON (H.ResultError (H.ServerError c m d h)) = JSON.object [
-    "code" .= (toS c::Text),
-    "message" .= (toS m::Text),
-    "details" .= (fmap toS d::Maybe Text),
-    "hint" .= (fmap toS h::Maybe Text)]
+  toJSON (H.ResultError (H.ServerError c m d h)) = case toS c of
+    'P':'T':_ ->
+      JSON.object [
+        "details" .= (fmap toS d::Maybe Text),
+        "hint" .= (fmap toS h::Maybe Text)]
+    _ ->
+      JSON.object [
+        "code" .= (toS c::Text),
+        "message" .= (toS m::Text),
+        "details" .= (fmap toS d::Maybe Text),
+        "hint" .= (fmap toS h::Maybe Text)]
   toJSON (H.ResultError (H.UnexpectedResult m)) = JSON.object [
     "message" .= (m::Text)]
   toJSON (H.ResultError (H.RowError i H.EndOfInput)) = JSON.object [
@@ -138,7 +145,7 @@ instance JSON.ToJSON H.Error where
 
 httpStatus :: Bool -> P.UsageError -> HT.Status
 httpStatus _ (P.ConnectionError _) = HT.status503
-httpStatus authed (P.SessionError (H.ResultError (H.ServerError c _ _ _))) =
+httpStatus authed (P.SessionError (H.ResultError (H.ServerError c m _ _))) =
   case toS c of
     '0':'8':_ -> HT.status503 -- pg connection err
     '0':'9':_ -> HT.status500 -- triggered action exception
@@ -166,6 +173,7 @@ httpStatus authed (P.SessionError (H.ResultError (H.ServerError c _ _ _))) =
     "42883"   -> HT.status404 -- undefined function
     "42P01"   -> HT.status404 -- undefined table
     "42501"   -> if authed then HT.status403 else HT.status401 -- insufficient privilege
+    'P':'T':n -> fromMaybe HT.status500 (HT.mkStatus <$> readMaybe n <*> pure m)
     _         -> HT.status400
 httpStatus _ (P.SessionError (H.ResultError _)) = HT.status500
 httpStatus _ (P.SessionError (H.ClientError _)) = HT.status503
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -4,6 +4,7 @@ import           Protolude
 import qualified GHC.Show
 import           Data.Aeson
 import qualified Data.ByteString.Lazy as BL
+import qualified Data.CaseInsensitive as CI
 import qualified Data.HashMap.Strict  as M
 import           Data.Tree
 import qualified Data.Vector          as V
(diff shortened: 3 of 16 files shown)


begriffs:
Steve thank you, you are simply amazing.

Let me review the older PRs first and then I'll look into this one.

(Sorry for my somewhat slow response lately, I've been on a trip.)

ruslantalpa:
code looks good 👍

steve-chavez:
Forgot to mention that I'm using the `current_setting(setting_name [, missing_ok ])` function, the `missing_ok` was added in PostgreSQL 9.6, so maybe this feature should be left for `v0.5.0.0`.

Also, should we allow the user to overwrite the `Content-Type/Content-Range` headers with the `response.headers` guc? Right now the header value in the guc is being appended.

Maybe an use case for that would be returning a binary with the `Content-Type` changed, so the user could have a function like `create function my_pic() returns bytea` then he could request the rpc with an `Accept: application/octet-stream` and in the function body do a `set local "response.headers" = [{"Content-Type": "image/png"}, {"Vary": "Accept"}]` or something like that.

begriffs:
Can you remind me why we prefer an array like

```json
[{"X-Test": "Domain=.server.com; Secure; Version=1"},
 {"X-Test-2": "TimeoutLoggedIn=login; Version=1"}]
```

Over a single object like

```json
{"X-Test": "Domain=.server.com; Secure; Version=1",
 "X-Test-2": "TimeoutLoggedIn=login; Version=1"}
```

> Also, should we allow the user to overwrite the Content-Type/Content-Range headers with the response.headers guc? Right now the header value in the guc is being appended.

I vote that we should allow the override. Then people could make one of these procs return `text/html` if they wanted. Or your example of returning an image. Seems like it opens up a lot of possibilities without much extra trouble on our end.

> Forgot to mention that I'm using the current_setting(setting_name [, missing_ok ]) function, the missing_ok was added in PostgreSQL 9.6, so maybe this feature should be left for v0.5.0.0.

Does it always attempt calling that form of `current_setting` for every RPC call? Meaning it would definitely break backward compatibility on pre 9.6 postgresql for any RPC call?

ruslantalpa:
- because you can have the same header multiple times (set cookie for example)
- yes, called every time. 

about compatibility:
i propose this solution:

- based on the pg version (which we know), use different form of `current_setting`.
- in 9.6 it will just work
- in older versions, to upgrade to the new postgrest version, users will jsut have to run something like `alter database .. set 'response.headers'=''` to set a default value and it will work, i think is a very small change in order to be able to upgrade and use the latest version

steve-chavez:
Added a test for `Set-Cookie` duplicate headers to make more clear the need of the json array structure. Also added the ability to overwrite only the `Content-Type`, couldn't think of a use case that would need overwriting the `Content-Range`.

For not breaking backward compatibility I propose detecting the PostgreSQL version and disabling this feature if it is `< 9.6`. I could just do `null as headers` in `callProc` for disabling it.

I think that most of the users don't need the custom headers, and making the docs say that they need to do `alter database .. set response.headers=''` to enable RPC in upcoming versions doesn't seem right, it seems fairer to me to tell the users that if they need this feature they have to upgrade to `9.6`.

src/PostgREST/Error.hs
```diff
@@ -166,6 +179,7 @@ httpStatus authed (P.SessionError (H.ResultError (H.ServerError c _ _ _))) =
     "42883"   -> HT.status404 -- undefined function
     "42P01"   -> HT.status404 -- undefined table
     "42501"   -> if authed then HT.status403 else HT.status401 -- insufficient privilege
+    'P':'T':n -> fromMaybe HT.status400 (HT.mkStatus <$> readMaybe n <*> pure m)
```
begriffs: If `n` can't be read as a number, maybe we should return a 500 error rather than 400 since it's a problem internal to the server rather than a problem with the request.

src/PostgREST/Types.hs
```diff
@@ -217,6 +218,23 @@ type NodeName = Text
...
+
+instance FromJSON GucHeader where
+  parseJSON (Object o) = case headMay (M.toList o) of
+    Just (k, String s) | M.size o == 1 -> pure $ GucHeader (k, s)
+                       | otherwise     -> mzero
+    _ -> mzero
+  parseJSON _          = mzero
+
+toHeaders :: [GucHeader] -> [Header]
+toHeaders (GucHeader (k, v):ghs) = (CI.mk $ toS k, toS v):toHeaders ghs
```
begriffs: What about
```hs
toHeaders = map $ \(GucHeader (k, v)) -> (CI.mk $ toS k, toS v)
```

src/PostgREST/App.hs
```diff
@@ -252,14 +253,21 @@ app dbStructure conf apiRequest =
...
-                    fromMaybe (Just 0, 0, "[]") row
+              let (tableTotal, queryTotal, body, jsonHeaders) =
+                    fromMaybe (Just 0, 0, "[]", Nothing) row
                   (status, contentRange) = rangeHeader queryTotal tableTotal
-              if singular && queryTotal /= 1
-                then do
-                  HT.condemn
-                  return $ singularityError (toInteger queryTotal)
-                else return $ responseLBS status [toHeader contentType, contentRange] (toS body)
+                  decodedHeaders = first toS <$> (eitherDecode <$> (toS <$> jsonHeaders)) :: Maybe (Either Text [GucHeader])
```
begriffs: Trying to understand this somewhat complex type. What's the difference between `Nothing`, `Left t` and `Right []`?

begriffs:
> For not breaking backward compatibility I propose detecting the PostgreSQL version and disabling the custom headers feature if it is < 9.6. I could just do null as headers in callProc for disabling it.

That sounds fine.

ruslantalpa:
no feature for < 9.6 is ok with me too, i just  thought it was an relatively easy way to bring this feature to ppl that can't upgrade the db (i for example would have preferred to run that sql and have that feature) but of course, it is easier on us to just require a 9.6+

src/PostgREST/App.hs
steve-chavez: If the user doesn't do a `set local "response.headers"` then this gives Nothing, this comes from the `Maybe ByteString` in `type ProcResults` in QueryBuilder.

`Left t` happens when the `response.headers` has bad format, like in `"response.headers" = '{"X-Test": "invalid"}'`(can only be array) or `"response.headers" = '{{}}'`. I tried to use aeson `fail` function in `parseJSON` to specify the error but sometimes it didn't respond with my custom error so that's why I don't use the `t` in `Left t`.  

`Just (Right [])` can only happen if `response.headers = '[]'` is set, I made this strict so even `response.headers='[{"Set-Cookie": null}]'/'[{"Set-Cookie": 666}]'/'[{"Set-Cookie": true}]'` will be a `Left`.

I struggled a bit to keep this part DRY, for now don't know how to make it clearer.

steve-chavez:
Added the PostgreSQL version detection for maintaining backward compatibility, guc headers feature is disabled for `< 9.6`. Also I moved the tests dependent on `9.6` to their own spec(also moved the `phraseto_tsquery` there) to make sure RPC works normally for `< 9.6`. Tried to make `9.3` the default version to run all the tests but `jsonb`(introduced on `9.4`) is used in some of the [tests](https://github.com/begriffs/postgrest/blob/65c9d549c1760cad63c42120ee95a64cff583211/test/Feature/InsertSpec.hs#L196) so default docker pg image for the tests on CircleCI is `9.4`.

ryanfmurphy:
This is great, I was hoping something like this would become available so I could set "Content-Type: text/html" and thus define a Postgres function / RPC endpoint that functions as an HTML view.

Do you guys need/want any help testing this?  I was looking for a way to apply this change and it's not super clear to me in the Github UI how to locate that branch / fork and apply the change.

begriffs:
@ryanfmurphy if you want to give this branch a try before we merge it you can do that in one of two ways:

1. Clone the project as usual, and apply the patch for this change:
   ```bash
   git clone https://github.com/begriffs/postgrest.git
   curl -L https://patch-diff.githubusercontent.com/raw/begriffs/postgrest/pull/986.patch | git apply -
   ```
2. OR, clone Steve's fork and switch to the branch
   ```bash
   git clone https://github.com/steve-chavez/postgrest.git
   git checkout rpc-headers-status
   ```

src/PostgREST/App.hs
begriffs: What about treating two of these cases the same: when the user doesn't set `response.headers` and when they set it to `[]`? We could then simplify the type to `Either Text [GucHeader]` and return `Right []` when `response.headers` is not set.

src/PostgREST/App.hs
```diff
@@ -248,18 +246,26 @@ app dbStructure conf apiRequest =
...
-                then do
-                  HT.condemn
-                  return $ singularityError (toInteger queryTotal)
-                else return $ responseLBS status [toHeader contentType, contentRange] (toS body)
+                  decodedHeaders = first toS $ eitherDecode $ toS jsonHeaders :: Either Text [GucHeader]
+              case decodedHeaders of
+                Left _ -> return gucHeadersError
+                Right hs -> do
+                  let customHeaders = toHeaders hs
+                      contentTypeH | isNothing $ lookup hContentType customHeaders = [toHeader contentType]
```
begriffs: Too bad duplicate keys are significant or we could make customHeaders a map like

```haskell
let customHeaders = M.fromList (toHeaders hs)
      `M.union` M.fromList [toHeader contentType, contentRange]
```

And then use `toList` in `responseLBS`. ...but yeah duplicate keys are allowed.

___

In any case I've got a question about the current logic. It falls back to the server<->client negotiated content type if the GUC has not set a content type, otherwise uses the GUC's choice. However `responseContentTypeOrError` recognizes only a short list of Accept types for rpc calls:

```haskell
ActionInvoke _ ->  [CTApplicationJSON, CTSingularJSON, CTTextCSV, CTOctetStream]
```

If a client calls `/rpc/generate_page` with `Accept: text/html` won't the request die when `responseContentTypeOrError` returns Left -- even though `generate_page` may be capable of producing HTML?

It seems to be only when clients add `*/*` to their Accept header that RPC calls will work with functions returning fancy content types.

begriffs: I was thinking of this situation:

```haskell
it "content negotiation allows nonstandard Accept when function agrees" $
  request methodGet "/rpc/hello_in_text_plain" (acceptHdrs "text/plain") ""
    `shouldRespondWith`
    "hello"
    { matchHeaders = ["Content-Type" <:> "text/plain"] }
```
Which produces
```
status mismatch:
  expected: 200
  but got:  415
missing header:
  Content-Type: text/plain
the actual headers were:
body mismatch:
  expected: hello
  but got:  {"message":"None of these Content-Types are available: text/plain"}
```

My thought is, the client should be able to request the content type that the RPC will ultimately provide.

ruslantalpa: @begriffs you are thinking about this PR from the direction of "let's generate html with postgrest" which is not the goal for it, i know i mentioned in other threads that this type of PR might open the html feature (which i think is a bad way of doing it) but it does not mean that this particular pr should make it possible.

There are lots of other useful headers out there (like set cookie which will enable cookie auth) that have nothing to do with the incoming request headers, and this PR is immensely useful already without having to implement Accept/Content-Type correlation.

Let's merge this and leave the html thing for another time (that is if someone wants to implement it).

begriffs: HTML is one example, but really I was thinking it makes sense that a client can request a certain content type and not be denied if the stored proc does indeed return that type. Having to request `application/octet-stream` and then getting some other type back is nonstandard HTTP. Requesting `Accept: application/octet-stream,*/*` and getting something else back does make sense.

I can see the point that this may not end up being a common case, and it may be hard to fix this aspect of the code because usually the code evaluates the Accept header against a fixed set of types and decides immediately to fail rather than running the stored proc and checking retroactively. We could overlook this in the current pull request, but it does mean our treatment of HTTP is kind of broken.

ruslantalpa: i would not say the current way is broken, it's only that this PR allows a RPC to break it. If the RPC does not touch content-type then everything works as it should, so if it breaks, it's on "them" :)

ruslantalpa: come to think of it, isn't it the same in every other env php/node, the browser sends an accept header while the script can do whatever it wants?

the only thing different here is that we don't allow "unknown" types to fall through to the rpc but this is a completely different topic, how do we know if the rpc is going to take care of the content type or not (probably by inspecting the return type).

or maybe an RPC should not be allowed to directly change the content type header and the content type should be determined by it's return type. like bytea -> octetstream, text -> (json or html), any other -> as it was before.
but as i said, this is a completely different discussion/pr.

steve-chavez: Initially I thought that the responsibility of making the http treatment adequate should be left to the developer and didn't consider the unsupported case where client does an `Accept: text/html` and the response has the same content type returned, was always thinking in terms of an `Accept: application/octet-stream` and then RPC overwrites content-type but as mentioned that is not an adequate treatment of http.

I think I have a solution for this and preserve the most of http semantics.

What if we allow arbitrary `Content-Types` to be requested on RPC? Well the RPC could set the matching content type but the body could be anything, including the default json, we cannot check the body for compliance but really the only way for the body to be say an `text/html` is that the return type of the RPC function is bytea/text, both of these types being supported by the binary output feature, so we could enforce that condition.

So the solution I propose would be:

- Make RPC support other arbitrary content types, with the requirement that when one of these types is requested, the function has to return `bytea/text` or client selects a column of that type for a row, same condition used for binary output.
- Check that the `Accept` content type matches the one provided by the RPC custom content type otherwise fail.

However as @ruslantalpa mentioned the feature request was mostly about setting other headers and not `Content-Type`, @begriffs how about if I remove the content type overwriting from this PR and then continue this discussion on another issue? I could open one with my proposed solution when this is merged.

steve-chavez: I'll go ahead with removing the `Content-Type` overwriting(I'll keep a local branch with the change and I'll make a note for opening another issue with the proposal).

steve-chavez:
I've removed the `Content-Type` overwriting, I'll open another issue on that when this is merged.

@begriffs Let me know if this is good to merge.

begriffs:
Looks good to me!

deinspanjer:
Are there instructions anywhere on how you get from raising an error with a hint and details in postgres to a specific response with a particular HTTP status code and the desired message in the body?

ruslantalpa:
@deinspanjer look at the tests for the PR

deinspanjer:
Okay.. I did look through the code including the files that were added and changed in the /test directory, but I didn't see anything that looked like a straightforward SQL example, and I am not familiar with Haskell.  Don't worry about it though.  If it is obvious to other people then that s great.

ruslantalpa:
HTTP
https://github.com/steve-chavez/postgrest/blob/75620e5799307d188e595f22dcc05463be068842/test/Feature/RpcSpec.hs#L279-L284

SQL
https://github.com/steve-chavez/postgrest/blob/75620e5799307d188e595f22dcc05463be068842/test/fixtures/schema.sql#L1296-L1302

deinspanjer:
Yup, just as I feared.  Very obvious and I just overlooked it. :)  Thanks @ruslantalpa

Orbmancer:
Hey that's an awesome feature !
Adding it to the documentation could be great for other developers as well, I had to search hard to find this :wink:

I'm using the Postgrest Docker container for my production, do the maintainers have planned to publish the newest versions (including this PR) ? Last published is  [v0.4.2.0](https://hub.docker.com/r/begriffs/postgrest/tags/)

I'm going to build my own container in the meanwhile, but would definitely appreciate an official Docker build 😊

steve-chavez:
@Orbmancer The feature is documented in https://postgrest.com/en/v4.4/api.html#errors-and-http-status-codes and https://postgrest.com/en/v4.4/api.html#setting-response-headers.
 
Also the latest docker images are in https://hub.docker.com/r/postgrest/postgrest/tags/.

Orbmancer:
@steve-chavez thanks for the links !
My bad I didn't saw this part in the documentation 👍

Issue #876, opened 2017-05-18
Binary secret cannot be read from a file


eric-brechemier:
Using following secret made of 8 random bytes:

```
$ openssl rand 8 > /tmp/secret.bin
$ cat /tmp/secret.bin | xxd
00000000: 8ea1 d877 40be 5639                      ...w@.V9
```

```
## choose a secret to enable JWT auth
## (use "@filename" to load from separate file)
jwt-secret = "@/tmp/secret.bin"
# secret-is-base64 = false
```

PostgREST v0.4.1.0 fails with following error when the secret is read from a file:

```
postgrest: /tmp/secret.bin: hGetContents: invalid argument (invalid byte sequence)
```

begriffs:
Thanks for the report. I see there are commits referring to this issue -- do you have a PR you can send to fix it?

begriffs:
@eric-brechemier probably base64 encoding would be a workaround, but do you already have an idea of the cause of this problem?

eric-brechemier:
@begriffs Yes.

1. [secret is defined as `Text` instead of `Secret`](https://github.com/eGullGolf/postgrest/commit/2649098b5c9766e51122065e062ce1d1e3fbc28a#diff-d6621ea7eea0dc9e1ea18f49f17cdc6bL42) 
2. [secret is corrupted by the conversion to `Text`](https://github.com/eGullGolf/postgrest/commit/2649098b5c9766e51122065e062ce1d1e3fbc28a#diff-cf0c4471495584a946dc7035028102e7L95) instead of [being saved as a `binarySecret`](https://github.com/eGullGolf/postgrest/commit/0c955cb28fc6f108923cdc12066f063dc6c02998#diff-cf0c4471495584a946dc7035028102e7R96)

ruslantalpa:
4.3 has a lot of new stuff related to jwt which probably fixes this (since you can use base64 to encode it)

eric-brechemier:
@ruslantalpa I'll uncomment the unit tests and check then ;)

eric-brechemier:
@ruslantalpa the tests still fail in v0.4.3.0:

```
$ ./io-tests.sh 
1..12
ok 1 - authentication with simple (no EOL) secret read from a file
not ok 2 - failed to authenticate using JWT for simple secret: 401
ok 3 - authentication with ASCII (no EOL) secret read from a file
not ok 4 - failed to authenticate using JWT for ASCII secret: 401
[...]

PR #988, opened 2017-09-27
Read secret file as binary, without premature parsing as UTF-8 (#876)

eric-brechemier:
This allows to read binary secrets as well as text (UTF-8) or Base64
secrets. Also, the final LF character (`\n`), if any, is now removed.

After this change, all I/O unit tests now pass.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -16,13 +16,14 @@ import           Protolude hiding         (replace, hPutStrLn)
 import           Control.Retry            (RetryStatus, capDelay,
                                            exponentialBackoff,
                                            retrying, rsPreviousDelay)
-import           Data.ByteString.Base64   (decode)
 import           Data.IORef               (IORef, atomicWriteIORef,
                                            newIORef, readIORef)
 import           Data.String              (IsString (..))
 import           Data.Text                (pack, replace, stripPrefix, strip)
 import           Data.Text.Encoding       (decodeUtf8, encodeUtf8)
-import           Data.Text.IO             (hPutStrLn, readFile)
+import           Data.Text.IO             (hPutStrLn)
+import qualified Data.ByteString.Base64   as B64
+import qualified Data.ByteString          as BS
 import qualified Hasql.Decoders           as HD
 import qualified Hasql.Encoders           as HE
 import qualified Hasql.Pool               as P
@@ -230,6 +231,12 @@ main = do
          refDbStructure
          refIsWorkerOn)
 
+chomp :: ByteString -> ByteString
+chomp bs =
+  case BS.stripSuffix "\n" bs of
+    Nothing -> bs
+    Just rest -> rest
+
 {-|
   The purpose of this function is to load the JWT secret from a file if
   configJwtSecret is actually a filepath and replaces some characters if the JWT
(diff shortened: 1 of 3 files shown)


main/Main.hs
```diff
@@ -230,6 +231,9 @@ main = do
          refDbStructure
          refIsWorkerOn)
 
+chomp :: ByteString -> ByteString
+chomp bs = fromMaybe bs (BS.stripSuffix "\n" bs)
```
begriffs: I was trying to think of a shorter way to write this but this seems to be best.

Also how about defining `chomp` in a where clause of `extractAndTransform` rather than as a top level function? The reason I say this is no other functions use this helper.

test/io-tests.sh
```diff
@@ -93,13 +93,13 @@ totalTests=12
 echo "1..$totalTests"
 
 readSecretFromFile word.noeol 'simple (no EOL)'
-skip readSecretFromFile word.txt 'simple'
```
begriffs: Nice, bye bye skip!

begriffs:
By the way, how does it tell the difference between binary and utf-8 when it loads a file? I'm imagining that the same sequence of bytes could yield both interpretations?

eric-brechemier:
UTF-8 is the binary encoding of text.

main/Main.hs
eric-brechemier: > I was trying to think of a shorter way to write this but this seems to be best.

Actually, this short form was suggested by the linter. What I wrote was much longer, using a `case`.

> Also how about defining `chomp` in a where clause of `extractAndTransform` rather than as a top level function? The reason I say this is no other functions use this helper.

The way I see it, chomp should be moved to a separate file as a (potentially) reusable utility. I find that `extractAndTransform` is already hairy enough.

`chomp` is based on [the function of the same name in Perl](http://perldoc.perl.org/functions/chomp.html), which is a top-level utility.

begriffs: Yeah I'm familiar with it from perl. The haskell `Text` type has `strip` which will drop whitespace including newlines from before and after a string, but ByteString doesn't have it.

Trying to think of a good place for this function. Was thinking `Types.hs` for a second, but that's more for our internal types and their helper functions.  Could make a whole new file seeing that this is a general purpose helper, but that seems premature because no other functions in the codebase need chomp. That's why I was suggesting to make it private to the function who uses it.

begriffs:
OK, I was thinking if we're now using raw bytes for the secret that maybe utf-8 would need to be recognized/interpreted somehow, like what if there was a difference bytewise between the string "cat" as ASCII and "cat" as unicode. However I read a [Q&A](http://www.jguru.com/faq/view.jsp?EID=297134) saying:
> UTF-8 represents codepoints between 0-127 with one byte, so it's equivalent to ASCII.

Anyway, it's no problem.

eric-brechemier:
You can consider that `ASCII < UTF-8 < BINARY`.

We previously parsed the file as text, which covered `ASCII < UTF-8` but failed on `BINARY`. The file is now read as `BINARY`, which preserves all the rest as well.

main/Main.hs
eric-brechemier: I am fine with either. I will make it private for now.

PR #996, opened 2017-10-03
Fix embedded column conflicts table name

grotsev:
Error is reported while resource embedding when table name and column name in the table are same.
```
{"hint":"No function matches the given name and argument types. You might need to add explicit type casts."
,"details":null
,"code":"42883"
,"message":"function row_to_json(uuid) does not exist"
}
```
`uuid` is type of column, but it should be `row`.

To name table explicitly use `.*`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #962, OpenAPI don't err on nonexistent schema - @steve-chavez
 - #954, make OpenAPI rpc output dependent on user privileges - @steve-chavez
 - #955, Support configurable aud claim - @statik
+- Fix embedded column conflicts table name - @grotsev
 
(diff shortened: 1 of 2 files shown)


ruslantalpa:
maybe add a change log line for credit :) and it's good to merge!

grotsev:
@ruslantalpa Fixed )

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #962, OpenAPI don't err on nonexistent schema - @steve-chavez
 - #954, make OpenAPI rpc output dependent on user privileges - @steve-chavez
 - #955, Support configurable aud claim - @statik
+- Fix embedded column conflicts table name - @grotsev
```
begriffs: Could you precede it with "#996" please?

begriffs:
Thanks @grotsev !

PR #997, opened 2017-10-04
Update jose to 0.6

PierreR:
To match the version available in stack LTS-9.x.

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -42,20 +42,19 @@ data JWTAttempt = JWTInvalid JWTError
   Receives the JWT secret and audience (from config) and a JWT and returns a map
   of JWT claims.
 -}
-jwtClaims :: Maybe JWK -> Text -> BL.ByteString  -> IO JWTAttempt
+jwtClaims :: Maybe JWK -> StringOrURI -> BL.ByteString  -> IO JWTAttempt
 jwtClaims _ "" "" = return $ JWTClaims M.empty
 jwtClaims secret audience payload =
   case secret of
     Nothing -> return JWTMissingSecret
-    Just jwk -> do
-      let validation = set audiencePredicate (==  fromString audience) defaultJWTValidationSettings
+    Just k -> do
+      let validation = defaultJWTValidationSettings (== audience)
       eJwt <- runExceptT $ do
         jwt <- decodeCompact payload
-        validateJWSJWT validation jwk jwt
-        return jwt
+        verifyClaims validation k jwt
       return $ case eJwt of
         Left e -> JWTInvalid e
-        Right jwt -> JWTClaims . claims2map . jwtClaimsSet $ jwt
+        Right jwt -> JWTClaims . claims2map $ jwt
 
 {-|
   Whether a response from jwtClaims contains a role claim
@@ -89,4 +88,4 @@ hs256jwk key =
     & jwkUse .~ Just Sig
     & jwkAlg .~ (Just $ JWSAlg HS256)
  where
-  km = OctKeyMaterial (OctKeyParameters Oct (Base64Octets key))
+  km = OctKeyMaterial (OctKeyParameters (Base64Octets key))
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -18,16 +18,12 @@ module PostgREST.Auth (
   , parseJWK
   ) where
 
-import           Protolude        hiding ((&))
-import           Control.Lens
-import           Data.Aeson              (Value (..), decode, toJSON)
-import qualified Data.ByteString.Lazy    as BL
-import qualified Data.HashMap.Strict     as M
+import           Control.Lens.Operators
+import           Data.Aeson             (Value (..), decode, toJSON)
+import qualified Data.HashMap.Strict    as M
+import           Protolude
 
-import           Crypto.JOSE.Compact
-import           Crypto.JOSE.JWK
-import           Crypto.JOSE.JWS
-import           Crypto.JOSE.Types
+import qualified Crypto.JOSE.Types      as JOSE.Types
 import           Crypto.JWT
 
 {-|
@@ -42,7 +38,7 @@ data JWTAttempt = JWTInvalid JWTError
   Receives the JWT secret and audience (from config) and a JWT and returns a map
   of JWT claims.
 -}
-jwtClaims :: Maybe JWK -> Maybe StringOrURI -> BL.ByteString  -> IO JWTAttempt
+jwtClaims :: Maybe JWK -> Maybe StringOrURI -> LByteString  -> IO JWTAttempt
 jwtClaims _ Nothing "" = return $ JWTClaims M.empty
 jwtClaims secret audience payload =
   case secret of
@@ -53,7 +49,7 @@ jwtClaims secret audience payload =
         jwt <- decodeCompact payload
         verifyClaims validation s jwt
       return $ case eJwt of
-        Left e -> JWTInvalid e
+        Left e    -> JWTInvalid e
         Right jwt -> JWTClaims . claims2map $ jwt
 
 {-|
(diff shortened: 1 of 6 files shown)


PierreR:
Please review this PR closely (I am a complete newbie with `jose` and `JWT`)

stack.yaml
```diff
@@ -1,11 +1,9 @@
-resolver: lts-9.0
+resolver: lts-9.6
 extra-deps:
   - configurator-ng-0.0.0.1
   - critbit-0.2.0.0
-  - hasql-pool-0.4.1
   - hjsonpointer-1.1.1
   - hjsonschema-1.5.0.1
-  - jose-0.5.0.3
```
begriffs: Nice to be able to remove one of these custom deps

stack.yaml
```diff
@@ -1,11 +1,9 @@
-resolver: lts-9.0
+resolver: lts-9.6
 extra-deps:
   - configurator-ng-0.0.0.1
   - critbit-0.2.0.0
-  - hasql-pool-0.4.1
```
begriffs: Does the new LTS try including a newer version of hasql-pool that conflicted with the old choice of 0.4.1?

src/PostgREST/Auth.hs
```diff
@@ -23,6 +23,7 @@ import           Control.Lens
 import           Data.Aeson              (Value (..), decode, toJSON)
 import qualified Data.ByteString.Lazy    as BL
 import qualified Data.HashMap.Strict     as M
+import           Data.Maybe (fromJust)
```
begriffs: This is nitpicky, but how about indenting fromJust to match the rightmost column of the imports?

src/PostgREST/Auth.hs
```diff
@@ -47,15 +48,15 @@ jwtClaims _ "" "" = return $ JWTClaims M.empty
 jwtClaims secret audience payload =
   case secret of
     Nothing -> return JWTMissingSecret
-    Just jwk -> do
-      let validation = set audiencePredicate (==  fromString audience) defaultJWTValidationSettings
+    Just k -> do
+      let audience' = fromJust $ preview stringOrUri audience
```
begriffs: Huh, `preview` is cool, I never knew about that.

However `fromJust` is an unsafe function that dies on Nothing. Is it possible to make this code more robust?

PierreR: I hope there is one ... It comes directly from the `jose` [example](https://github.com/frasertweedale/hs-jose/blob/master/example/Main.hs#L82). I also find this line quite unsatisfying.

stack.yaml
PierreR: The new LTS includes this [version](https://www.stackage.org/lts-9.6/package/hasql-pool-0.4.1) exactly

stack.yaml
PierreR: I am using `nix` which uses stack LTS by default. It is a bit annoying when an older version is used which motivates me for the change.

src/PostgREST/Auth.hs
PierreR: Just asked here: https://github.com/frasertweedale/hs-jose/issues/58

frasertweedale: If you're dealing with a user-supplied value you should treat the `Nothing` case a user error.
If it is a "known good" value then use `fromJust . preview stringOrUri` without remorse.

src/PostgREST/Auth.hs
PierreR: Done. I would suggest to even go further and to make it 'work' with stylish-haskell (which would involve separate the protolude line). As a note, the protolude line is usually at the bottom instead of the top of the import list IIRC.

src/PostgREST/Auth.hs
PierreR: @begriffs Do you wish to consider the Nothing case as a configuration error or leave it as before this PR. I would argue this is good to merge because it does not modify the old behaviour (but I am not convinced the old behaviour was sane)

begriffs: OK, I figured out how to trigger this edge case. Use a colon in the desired audience, but make sure it's not part of a valid url:

```ini
jwt-aud = "not::a url ? at all"
```

Then make a request with a valid jwt that contains an aud claim. It triggers an internal server error.

On the other hand I tried postgrest on the master branch in this same situation and it responds with HTTP 401:

```json
{
  "message": "JWTNotInAudience"
}
```

I don't know why the new jose library insists on parsing correct urls in the presence of a colon. Must be a more correct interpretation of the spec or something.

Given that the library is acting this way, it's probably best if we change the type of `configJwtAudience` to `StringOrURI`. https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Config.hs#L62

The parsing would need to change here as well - https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Config.hs#L121

frasertweedale: From https://tools.ietf.org/html/rfc7519#page-5:

 >     StringOrURI
 >        A JSON string value, with the additional requirement that while
 >        arbitrary string values MAY be used, any value containing a ":"
 >        character MUST be a URI [RFC3986].  StringOrURI values are
 >        compared as case-sensitive strings with no transformations or
 >        canonicalizations applied.

You can use the `stringOrUri` prism for parsing a `StringOrURI` value.

begriffs: Cool. I do like the correctness of the library, it forces us to do things right, like having a sufficient key length, or now this audience parsing.

src/PostgREST/Config.hs
```diff
@@ -132,15 +140,21 @@ readOptions = do
       return appConf
 
  where
+  parseJwtAudience :: Name -> C.ConfigParserM StringOrURI
+  parseJwtAudience k =
+    preview stringOrUri . fromMaybe ("" :: Text) <$> C.key k >>= \case
```
frasertweedale: If the key is not present, you are assuming `""` (empty string).  This means that the JWT will be required to
have an audience claim being the empty string.  Perhaps not what is intended?

If you want postgrest to *require* the administrator to set the  audience claim, you should probably require the config be specified, rather than defaulting it to empty string.

Otherwise you should propagate the evidence that no audience predicate was specified (e.g. as a
`Nothing :: Maybe StringOrURI`) and if it was not specified, set the audience predicate to `(const True)` to
accept any audience.

PierreR: Done. Thanks for the feedback.

begriffs:
This works great! I tried a bad jwt audience and it failed right away with a helpful error. That's going to help avoid a lot of potential confusion.

One thing that's unrelated to your changes but maybe you could fix in your PR while you're at it...during build time I see:

```
Linking .stack-work/dist/x86_64-osx/Cabal-1.24.2.0/build/postgrest/postgrest ...

Warning: The following modules should be added to exposed-modules or other-modules in ~/dev/hs/postgrest/postgrest.cabal:
    - In spec component:
        Feature.AudienceJwtSecretSpec
```

Otherwise this looks good to merge!

PierreR:
@begriffs Something that seems unrelated to the PR but a bit strange to me (I knew nothing about JWT before this PR) is the fact that if I add a secret on the config file but don't add any `jwt-aud`, on the client side all requests succeed as before without using any token. Is that expected ?

begriffs:
Can you share one of these surprising jwts, the secret used to sign it, and the secret and aud settings in your postgrest config?

PierreR:
Here is my config file:
```
db-uri = "postgres://salt:pepper@localhost:5432/saltstack"
db-schema = "public"
db-anon-role = "salt"
db-pool = 10

server-host = "*4"
server-port = 3000

jwt-secret = "XmguE3f/cN1sdS6q9Ek/ZjfybdDCt8pfgFfCHqOOGLs="
secret-is-base64 = true
```
I only try this with a simple normal `GET` request such as : http://localhost:3000/salt_result. I don't pass any tokens and it displays the result.

begriffs:
If the `salt_result` function is accessible to your `db-anon-role` (in this case `salt`), then that seems normal. An anonymous request comes in -- meaning one not including an `Authorization: Bearer <jwt>` header -- so postgrest falls back to executing the request as the anonymous role.

What *would* be weird is if that request contained a jwt that was signed with a different secret and the system allowed the request to continue.

PierreR:
OK then it is all good !

Issue #966, opened 2017-09-07
How to convert pl/sql  AES code to postgreSQL AES 


abhinaycha:
declare
   input_string       VARCHAR2 (200) := 'Secret Message';
   output_string      VARCHAR2 (200);
   encrypted_raw      RAW (2000);             -- stores encrypted binary text
   decrypted_raw      RAW (2000);             -- stores decrypted binary text
   num_key_bytes      NUMBER := 256/8;        -- key length 256 bits (32 bytes)
   key_bytes_raw      RAW (32);               -- stores 256-bit encryption key 
   encryption_type    PLS_INTEGER :=          -- total encryption type
                            DBMS_CRYPTO.ENCRYPT_AES256
                          + DBMS_CRYPTO.CHAIN_CBC
                          + DBMS_CRYPTO.PAD_PKCS5;
begin
   DBMS_OUTPUT.PUT_LINE ('Original string: ' || input_string);
   key_bytes_raw := DBMS_CRYPTO.RANDOMBYTES (num_key_bytes);
   encrypted_raw := DBMS_CRYPTO.ENCRYPT
      (
         src => UTL_I18N.STRING_TO_RAW (input_string, 'AL32UTF8'),
         typ => encryption_type,
         key => key_bytes_raw
      );
    -- The encrypted value in the encrypted_raw variable can be used here:
   decrypted_raw := DBMS_CRYPTO.DECRYPT
      (
         src => encrypted_raw,
         typ => encryption_type,
         key => key_bytes_raw
      );
   output_string := UTL_I18N.RAW_TO_CHAR (decrypted_raw, 'AL32UTF8');
   DBMS_OUTPUT.PUT_LINE ('Decrypted string: ' || output_string);
end;

ruslantalpa:
this is not a repo for the PostgreSQL database server, this is PostgREST.
the question has nothing to do with this repo

PR #998, opened 2017-10-07
Fix RPC error when function has single OUT param

### LINKED ISSUE #974: rpc function(with OUT parameters) must have return named same as function name
If I have a function
`create function my_rpc( foo int, OUT bar text)`
I get an error like this
`{
  "details": "Row number 0",
  "message": "Row error: unexpected null"
}`
The function works if the OUT parameter is named the same as the function (which is a nice practice anyway). So this works:
`create function my_rpc( foo int, OUT my_rpc text)`

steve-chavez:
Fixes #974, I had to accommodate OpenAPI output because this [test](https://github.com/steve-chavez/postgrest/blob/master/test/Feature/StructureSpec.hs#L23-L25) was failing when including functions with OUT/INOUT parameters in `fixtures/schema.sql`, also I included a test for the recent #996.

Would be better to merge this PR before the others currently opened so the commit with the test can follow up the commit in #966. Also from now on, after reviews, I would like to merge my own PRs if that's fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #954, make OpenAPI rpc output dependent on user privileges - @steve-chavez
 - #955, Support configurable aud claim - @statik
 - #996, Fix embedded column conflicts table name - @grotsev
+- #974, Fix RPC error when function has single OUT param - @steve-chavez
 
 ## [0.4.3.0] - 2017-09-06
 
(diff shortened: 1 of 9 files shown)


begriffs:
Ahh, `select foo()` vs `select * from foo()`. Nice tests too. Looks good to merge! 🚢 

> Also from now on, after reviews, I would like to merge my own PRs if that's fine.

Good point, let's do it that way, that's the normal way come to think of it. 😄

Issue #1001, opened 2017-10-11
Bad csv output on queries with embeds


steve-chavez:
When doing:
```http
GET /clients?select=*,projects(*) 
Accept:text/csv
```
The following is obtained:
```csv
id,name,projects
1,Microsoft,"[{""id"":1,""name"":""Windows 7"",""client_id"":1},{""id"":2,""name"":""Windows 10"",""client_id"":1}]"
2,Apple,"[{""id"":3,""name"":""IOS"",""client_id"":2},{""id"":4,""name"":""OSX"",""client_id"":2}]"
```
I guess that csv queries with embeds aren't been used as much because until now this hasn't been reported. Since a tree structure cannot be done in csv, I think the most appropriate output would be:
```csv
id,name,projects_id,projects_name,projects_client_id
1,Microsoft,1,Windows 7,1
1,Microsoft,2,Windows 10,1
2,Apple,3,IOS,2
2,Apple,4,OSX,2
```
This corresponds more to a single query with joins, unlike our current queries that can have many subqueries with their own joins, so doesn't look like it maps too well to an "embed" operation. It may not be worth the effort to adjust the csv output on embed queries, maybe another option could be restricting csv to non-embed queries.

ruslantalpa:
Dont think its worth the effort, just a 400 bad request would be better

Issue #621, opened 2016-06-01
Avoid using CTEs in read/get requests


ruslantalpa:
@NikolayS pointed to an article that describes some problems CTEs generate (when the result is big)
http://adpgtech.blogspot.ru/2016/05/indiscriminate-use-of-ctes-considered.html?m=1

Currently we use a CTE like this

``` SQL
with pg_source as
    (select projects.id,
            row_to_json(clients.*) as client,
            coalesce(
                         (select array_to_json(array_agg(row_to_json(tasks)))
                          from
                              (select tasks.id
                               from tasks
                               where tasks.project_id = projects.id) tasks), '[]') as tasks
     from projects
     left outer join
         (select clients.id
          from clients) as clients on clients.id = projects.client_id
    )
select
    (select pg_catalog.count(1) from projects) as total_result_set,
    pg_catalog.count(t) as page_total,
    '' as header,
    coalesce(array_to_json(array_agg(row_to_json(t))), '[]')::character varying as body
from
    (select *
     from pg_source 
    ) t;

```

Not that the PR regarding limits is merged, it's possible to remove the CTE like this it like this

``` SQL
select
    (select pg_catalog.count(1) from projects) as total_result_set,
    pg_catalog.count(1) as page_total,
    '' as header,
    coalesce(array_to_json(array_agg(row_to_json(t))), '[]')::character varying as body
from
    (select projects.id,
                    row_to_json(clients.*) as client,
                    coalesce(
                                 (select array_to_json(array_agg(row_to_json(tasks)))
                                  from
                                      (select tasks.id
                                       from tasks
                                       where tasks.project_id = projects.id) tasks), '[]') as tasks
             from projects
             left outer join
                 (select clients.id
                  from clients) as clients on clients.id = projects.client_id

    ) t;

```

ruslantalpa:
Almost all tests pass except 2 related to CSV.
I can not generate the CSV header directly in sql
```SQL
-- current query
with pg_source as (
    select 'a' as column_a, 'b' as column_b
)
select
    -- csv header
   (
    select coalesce(string_agg(a.k, ','), '')
    from
     (select json_object_keys(r)::text as k
      from
        (select row_to_json(hh) as r
         from pg_source as hh
         limit 1
        ) s
     ) a
  ) 
  || E'\n' ||
  --csv body
  coalesce(string_agg(substring(t::text, 2, length(t::text) - 2), ''), '') AS body
from (
    select * from pg_source
) t

-- new query
select
[...]

PR #1002, opened 2017-10-12
Avoid CTE on non-csv GET requests

steve-chavez:
For #621, I've been trying to remove the CTE for GET queries with no success due to the CSV generation in sql.

So I thought of just leaving the CTE for CSV requests, considering #1001 I think that complex queries will most likely use json, so this might be a good enough workaround for now.

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -90,9 +90,15 @@ createReadStatement :: SqlQuery -> SqlQuery -> Bool -> Bool -> Bool -> Maybe Fie
 createReadStatement selectQuery countQuery isSingle countTotal asCsv binaryField =
   unicodeStatement sql HE.unit decodeStandard False
  where
-  sql = [qc|
-      WITH {sourceCTEName} AS ({selectQuery}) SELECT {cols}
-      FROM ( SELECT * FROM {sourceCTEName}) _postgrest_t |]
+  sql =
+    if asCsv
+      then [qc|
+        WITH {sourceCTEName} AS ({selectQuery}) SELECT {cols}
+        FROM ( SELECT * FROM {sourceCTEName}) _postgrest_t |]
+      else [qc|
+        SELECT {cols}
+        FROM ({selectQuery}) _postgrest_t |]
+
   countResultF = if countTotal then "("<>countQuery<>")" else "null"
   cols = intercalate ", " [
       countResultF <> " AS total_result_set",
@@ -321,19 +327,19 @@ unquoted (JSON.Number n) =
 unquoted (JSON.Bool b) = show b
 unquoted v = toS $ JSON.encode v
 
--- private functions
 asCsvF :: SqlFragment
 asCsvF = asCsvHeaderF <> " || '\n' || " <> asCsvBodyF
   where
-    asCsvHeaderF =
-      "(SELECT coalesce(string_agg(a.k, ','), '')" <>
-      "  FROM (" <>
-      "    SELECT json_object_keys(r)::TEXT as k" <>
-      "    FROM ( " <>
-      "      SELECT row_to_json(hh) as r from " <> sourceCTEName <> " as hh limit 1" <>
-      "    ) s" <>
-      "  ) a" <>
-      ")"
+    asCsvHeaderF = [qc|
+      (SELECT coalesce(string_agg(keys.name, ','), '')
+        FROM (
+          SELECT json_object_keys(sample)::text as name
+          FROM (
+            SELECT row_to_json(_) AS sample FROM {sourceCTEName} AS _ limit 1
+          ) _
+        ) keys
+      )
+    |]
     asCsvBodyF = "coalesce(string_agg(substring(_postgrest_t::text, 2, length(_postgrest_t::text) - 2), '\n'), '')"
 
 asJsonF :: SqlFragment


src/PostgREST/QueryBuilder.hs
```diff
@@ -321,19 +327,19 @@ unquoted (JSON.Number n) =
...
   where
-    asCsvHeaderF =
-      "(SELECT coalesce(string_agg(a.k, ','), '')" <>
-      "  FROM (" <>
-      "    SELECT json_object_keys(r)::TEXT as k" <>
-      "    FROM ( " <>
-      "      SELECT row_to_json(hh) as r from " <> sourceCTEName <> " as hh limit 1" <>
-      "    ) s" <>
-      "  ) a" <>
-      ")"
```
ruslantalpa: i was actually thinking of removing `qc` from the code since it's very inefficient when it comes to "big payloads".
We've tracked this down to being the culprit ob big memory usage.
I would say leave this part as it is and in another PR we'll fix the other sections
(there is a issue somehwere with a minimal test case that shows the memory consumption issue)

ruslantalpa:
i think it's a very good workaround since it might enable the use of citus which is huge.
does this change also remove the CTE from the insert/update/delete queries?

ruslantalpa:
still not removing from update/insert/delete since it looks like
```
WITH {sourceCTEName} AS ({mutateQuery})
SELECT {cols}
FROM ({selectQuery}) _postgrest_t
```

i think it's possible to rewrite this part too (although not necessarily in this PR).

@begriffs can you check if the select queries generated by this PR will actually work with citus? just removing ctes might not be enough since we also do subqueries and aggregations, and if that's the case, then there is no point in merging this.

src/PostgREST/QueryBuilder.hs
steve-chavez: In this case there's no input interpolation, no big payload chance there, in the out of memory rpc issue the args to the proc are being interpolated.

steve-chavez:
I'll see if the CTEs for update/delete/insert can be removed, even if Citus doesn't work with subqueries I think is worth removing them because the query planner can work better without them.

ruslantalpa:
"query planner can work better without them" that is a common wisdom however not in this case.
the way we use CTE here does not affect the query planner since it's used only at the last stage when all the data is already pulled in and it only needs to be json encoded.

src/PostgREST/QueryBuilder.hs
ruslantalpa: the idea is if this lib does not work well in all the cases maybe it's better to not include it at all, it only adds a bit of "nice strings", nothing more.
maybe by not using it we drop other dependencies too (though not many i think)

steve-chavez:
Did some tests and for the child/m2m embeds the CTE actually reduces the cost of the query planner to almost half of the subselect, for parent embeds I got the same cost with/without CTE.

```http
-- Child embed
GET /clients?select=*,projects{name}
WITH CTE : cost=31864.78..31864.80
WITHOUT CTE: cost=63646.53..63646.55

-- Many to many embed
GET /tasks?select=id,users{id}
WITH CTE : cost=51284.80..51284.82
WITHOUT CTE : cost=102490.60..102490.61
```
Queries are here: https://gist.github.com/steve-chavez/b9c1c4a444b11960dfde3c5b4b2270b3

src/PostgREST/QueryBuilder.hs
begriffs: The qc library does make the string building easier to read for those queries that aren't in danger of becoming too long with the payload.  I can see the benefit of removing the library so that we never accidentally use it in the wrong place, but if we're careful it may not cause any harm. Maybe there are more efficient text building libraries to investigate too?

begriffs:
I'll give it a try with Citus over the weekend. Really happy to see progress removing unnecessary CTEs!

begriffs:
I spun up a citus cluster and loaded it with this schema - https://docs.citusdata.com/en/v7.0/tutorials/multi-tenant-tutorial.html

I'm able to do more queries than on master, like filtering on a non-distribution column. However the columns that come back in the json are all relabeled as `f1`, `f2`, `f3` etc.

ruslantalpa:
@begriffs can you try on thing?
record the query postgrest generates (even the current release) then let's try variations of it with citus.

for those `f1` things, see if this helps
replace `array_to_json(array_agg(row_to_json(` portions with `json_agg`
it's an optimisation i've been meaning to add for a while, it make the queries faster and it might also help with those f1 things.
Also, please paste here the query and the result you get and we can brainstorm.

begriffs:
OK, gave it a try. it's interesting, I think Citus is rewriting the column names when it uses the "real-time executor" rather than the "router executor." Meaning if it has to break the query into fragments and run them across nodes rather than passing the query unchanged down to a node.

Here are the statements from my logs (I removed the part where it sets a bunch of client headers because it was just noisy).

### `/ads?company_id=eq.8` - Works properly

```
LOG:  execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ ONLY
LOG:  execute <unnamed>: 
	        SELECT null AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header, coalesce(array_to_json(array_agg(row_to_json(_postgrest_t))), '[]')::character varying AS body
	        FROM (SELECT  "public"."ads".* FROM  "public"."ads"  WHERE  "public"."ads"."company_id" = '8'::unknown   ) _postgrest_t 
LOG:  execute 1: COMMIT
```

### `/ads?campaign_id=eq.73` - Renames columns

```
LOG:  execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ ONLY
LOG:  execute <unnamed>: 
	        SELECT null AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header, coalesce(array_to_json(array_agg(row_to_json(_postgrest_t))), '[]')::character varying AS body
	        FROM (SELECT  "public"."ads".* FROM  "public"."ads"  WHERE  "public"."ads"."campaign_id" = '73'::unknown   ) _postgrest_t 
LOG:  execute 1: COMMIT
```

It's actually very easy to reproduce this yourself. [This tutorial](https://docs.citusdata.com/en/v7.0/tutorials/multi-tenant-tutorial.html) includes a Docker image running Citus and the sample data. It will set you up with the same schema and data I was using.

begriffs:
Tried the `json_agg` optimization and it must be a shortcoming of the current Citus version because it says

```json
{
  "hint": null,
  "details": null,
  "code": "XX000",
  "message": "unsupported aggregate function json_agg"
}
```

begriffs:
But to clarify, I don't think we're doing anything wrong, it's more that we're running into an edge case in how citus Citus breaks a query into fragments and reassembles the results. My hypothesis: we're swallowing the columns into json so a part of Citus on the coordinator node that is looking for those column names f1, f2 etc can't find them to rename them back.

Removing the CTE when possible seems like a good idea, so I'm cool with you merging this if you want.

ruslantalpa:
Why?
It still does not work on citus and the query is slower.

begriffs:
Is it slower?

> Did some tests and for the child/m2m embeds the CTE actually reduces the cost of the query planner to almost half of the subselect, for parent embeds I got the same cost with/without CTE.

Unless that comment refers only to planning cost but makes the query slower overall...

ruslantalpa:
from steve's tests https://github.com/begriffs/postgrest/pull/1002#issuecomment-336239333
the whole point of this was to make it compatible with citus (which is more of feel good/confidence thing for the majority users then an actual need right now )
CTE's, especially how we use them, are no problem by themselves, and it seems in this particular case of ours they make some queries twice as fast.

so i see no reason in removing them, we gain nothing and make the queries twice as slow.

begriffs:
Oops you're right, I read the comment with the measurements backwards. 😊 

I'll file an issue on the Citus repo about how columns are being renamed, and attempt to simplify our query to the smallest possible reproducible case for them.

begriffs:
Actually looks like this Citus issue was first reported in July - https://github.com/citusdata/citus/issues/1485

steve-chavez:
I'll close this, removing CTE's didn't show a perf advantage and achieving Citus compatibility is complicated for now.

Issue #956, opened 2017-08-28
Computed columns don't show up in swagger output


daurnimator:

steve-chavez:
What would be the most useful format for computed columns in openapi? Say we have:
```sql
CREATE TABLE items (
    id bigint NOT NULL
);

CREATE FUNCTION always_true(items) RETURNS boolean
    LANGUAGE sql STABLE
    AS $$ SELECT true $$;
```
Then for openapi output I was thinking of having this:
```json
{
  "paths":{
    "items": {
        "properties": {
            "id": {
                "format": "bigint",
                "type": "string",
                "description": "note:\nthis is a primary key.<pk/>"
            },
            "always_true": {
                "format": "boolean",
                "type": "string",
                "description": "note:\nthis is a computed column.<pk/>"
            }
        },
        "type": "object"
    }
  }
}
```

mdr1384:
Steve - I note that in 5.1.0 after using [swagger2markup](https://github.com/Swagger2Markup/swagger2markup) on the openapi output, the text descriptions still have that `<pk/>` and `<fk table... />` stuff in them making the text presented to the user a bit wonky. Is that XML stuff embedded in the text something that is used by Swagger UI?
![swagger-key](https://user-images.githubusercontent.com/32360633/48558291-feb88e80-e8b6-11e8-9e35-15fb936561f5.png)

steve-chavez:
@mdr1384 Not sure. @ldesgoui I've noticed you made that addition in https://github.com/PostgREST/postgrest/pull/885/files#diff-387c8dbc69527febf946f92972537ccaR51, are those tags needed?

ldesgoui:
I don't recall they are? They were a quick hack to encode some info about the schema, assuming the comments would be rendered using Markdown, which would hide them or filter them out

PR #1003, opened 2017-10-13
Add `set schema` to middleware

### LINKED ISSUE #828: Computed Column only works in public-schema
When implementing a "Computed Column" according to the documentation [here](https://postgrest.com/en/v0.4/api.html#computed-columns), it only worked when the function defining the computed column is in the ``public`` schema.

Is this a bug or is the PostgREST version used too old or is this wanted behaviour?

``Tested version is PostgREST v0.3.0 with PostgreSQL 9.4.``

### LINKED ISSUE #835: Test suite setup dependency on postgrest_test_authenticator search_path
I came across this on #831, while wanting to prove in tests the PR fix, I modified the ```search_path``` of the ```postgrest_test_authenticator``` and that resulted in tests, unrelated to the PR, failing.

When doing an ```ALTER ROLE postgrest_test_authenticator SET search_path = 'public', 'test'``` the following fail:

```
  test/Feature/AuthSpec.hs:130: 
  1) Feature.AuthSpec.authorization, custom pre-request proc acting on id claim, able to switch to postgrest_test_default_role (id=2)
       status mismatch:
         expected: 200
         but got:  401
       body mismatch:
         expected: "postgrest_test_default_role"
         but got:  {"hint":null,"details":null,"code":"42501","message":"permission denied for schema test"}

  test/Feature/StructureSpec.hs:42: 
  2) Feature.StructureSpec.OpenAPI.RPC includes a representative function with parameters
       expected: Just (Object (fromList [("required",Array [String "double",String "varchar",String "boolean",String "date",String "money",String "enum"]),("type",String "object"),("properties",Object (fromList [("double",Object (fromList [("format",String "double precision"),("type",String "string")])),("money",Object (fromList [("format",String "money"),("type",String "string")])),("date",Object (fromList [("format",String "date"),("type",String "string")])),("boolean",Object (fromList [("format",String "boolean"),("type",String "boolean")])),("varchar",Object (fromList [("format",String "character varying"),("type",String "string")])),("enum",Object (fromList [("format",String "test.enum_menagerie_type"),("type",String "string")])),("integer",Object (fromList [("format",String "integer"),("type",String "integer")]))]))]))
        but got: Just (Object (fromList [("required",Array [String "double",String "varchar",String "boolean",String "date",String "money",String "enum"]),("type",String "object"),("properties",Object (fromList [("double",Object (fromList [("format",String "double precision"),("type",String "string")])),("money",Object (fromList [("format",String "money"),("type",String "string")])),("date",Object (fromList [("format",String "date"),("type",String "string")])),("boolean",Object (fromList [("format",String "boolean"),("type",String "boolean")])),("varchar",Object (fromList [("format",String "character varying"),("type",String "string")])),("enum",Object (fromList [("format",String "enum_menagerie_type"),("type",String "string")])),("integer",Object (fromList [("format",String "integer"),("type",String "integer")]))]))]))
```

First one can be solved with a GRANT USAGE, in the second one there's a diff between ```test.enum_menagerie_type``` and ```enum_menagierie_type``` seems a problem in the detection of proc schema in openAPI.

When doing an ```ALTER ROLE postgrest_test_authenticator SET search_path = ''``` the following fail:

```
  test/Feature/AuthSpec.hs:43: 
  1) Feature.AuthSpec.authorization returns jwt functions as jwt tokens
       status mismatch:
         expected: 200
         but got:  404
       body mismatch:
         expected: {"token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xuYW1lIjoicG9zdGdyZXN0X3Rlc3RfYXV0aG9yIiwiaWQiOiJqZG9lIn0.P2G9EVSVI22MWxXWFuhEYd9BZerLS
1WDlqzdqplM15s"}
         but got:  {"hint":"No function matches the given name and argument types. You might need to add explicit type casts.","details":null,"code":"42883","m
essage":"function hmac(text, text, text) does not exist"}

  test/Feature/AuthSpec.hs:51: 
  2) Feature.AuthSpec.authorization sql functions can encode custom and standard claims
       status mismatch:
         expected: 200
         but got:  404
       body mismatch:
         expected: {"token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJqb2UiLCJzdWIiOiJmdW4iLCJhdWQiOiJldmVyeW9uZSIsImV4cCI6MTMwMDgxOTM4MCwibmJmIjoxMzAwO
DE5MzgwLCJpYXQiOjEzMDA4MTkzODAsImp0aSI6ImZvbyIsInJvbGUiOiJwb3N0Z3Jlc3RfdGVzdCIsImh0dHA6Ly9wb3N0Z3Jlc3QuY29tL2ZvbyI6dHJ1ZX0.IHF16ZSU6XTbOnUWO8CCpUn2fJwt8P00rlYV
yXQjpWc"}
         but got:  {"hint":"No function matches the given name and argument types. You might need to add explicit type casts.","details":null,"code":"42883","m
essage":"function hmac(text, text, text) does not exist"}

  test/Feature/InsertSpec.hs:398: 
  3) Feature.InsertSpec, Patching record, in a nonempty table, can update based on a computed column
       status mismatch:
         expected: 200
         but got:  400
       missing header:
         Content-Range: */*
       the actual headers were:
       body mismatch:
         expected: []
         but got:  {"hint":null,"details":null,"code":"42703","message":"column items.always_true does not exist"}

  test/Feature/QuerySpec.hs:111: 
  4) Feature.QuerySpec, Filtering response, matches with computed column
       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: [{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15
}]
         but got:  {"hint":null,"details":null,"code":"42703","message":"column items.always_true does not exist"}

  test/Feature/QuerySpec.hs:116: 
  5) Feature.QuerySpec, Filtering response, order by computed column
       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: [{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15
}]
         but got:  {"hint":null,"details":null,"code":"42703","message":"column items.anti_id does not exist"}
```

For the first ones a schema qualification for ```hmac``` in ```jwt.sql``` would be a solution, the other ones are related to #828.

steve-chavez:
Fix #828, breaking change: computed columns now only work if they are on the config schema.

Fix #835, tests now not depend on the search_path of the postgrest_test_authenticator.

### DIFF
--- a/src/PostgREST/Middleware.hs
+++ b/src/PostgREST/Middleware.hs
@@ -32,7 +32,7 @@ runWithClaims conf eClaims app req =
     JWTInvalid e -> return $ unauthed $ show e
     JWTMissingSecret -> return $ simpleError status500 [] "Server lacks JWT secret"
     JWTClaims claims -> do
-      H.sql $ toS.mconcat $ setRoleSql ++ claimsSql ++ headersSql ++ cookiesSql
+      H.sql $ toS.mconcat $ setSchemaSql ++ setRoleSql ++ claimsSql ++ headersSql ++ cookiesSql
       mapM_ H.sql customReqCheck
       app req
       where
@@ -41,6 +41,7 @@ runWithClaims conf eClaims app req =
         claimsSql = map (pgFmtEnvVar "request.jwt.claim.") [(c,unquoted v) | (c,v) <- M.toList claimsWithRole]
         setRoleSql = maybeToList $
           (\r -> "set local role " <> r <> ";") . toS . pgFmtLit . unquoted <$> M.lookup "role" claimsWithRole
+        setSchemaSql = ["set schema " <> pgFmtLit (configSchema conf) <> ";"] :: [Text]
         -- role claim defaults to anon if not specified in jwt
         claimsWithRole = M.union claims (M.singleton "role" anon)
         anon = String . toS $ configAnonRole conf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- #828, Fix computed column only working in public schema - @steve-chavez
+
+### Changed
+
+- Computed columns now only work if they belong to the db-schema - @steve-chavez
+
 ## [0.4.4.0] - 2018-01-08
 
 ### Added
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -171,6 +171,9 @@ spec = do
         [json| [{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}] |]
         { matchHeaders = [matchContentTypeJson] }
 
+    it "cannot access a computed column that is outside of the config schema" $
+      get "/items?always_false=is.false" `shouldRespondWith` 400
+
     it "matches filtering nested items 2" $
       get "/clients?select=id,projects{id,tasks2{id,name}}&projects.tasks.name=like.Design*"
         `shouldRespondWith` [json| {"message":"Could not find foreign keys between these entities, No relation found between projects and tasks2"}|]
--- a/test/Feature/StructureSpec.hs
+++ b/test/Feature/StructureSpec.hs
@@ -173,7 +173,7 @@ spec = do
                     "type": "string"
                   },
                   "enum": {
-                    "format": "test.enum_menagerie_type",
+                    "format": "enum_menagerie_type",
                     "type": "string"
                   },
                   "integer": {
--- a/test/fixtures/jwt.sql
+++ b/test/fixtures/jwt.sql
@@ -4,7 +4,6 @@ set client_min_messages to warning;
 DROP SCHEMA IF EXISTS jwt CASCADE;
 CREATE SCHEMA jwt;
 
-
 CREATE OR REPLACE FUNCTION jwt.url_encode(data bytea) RETURNS text LANGUAGE sql AS $$
     SELECT translate(encode(data, 'base64'), E'+/=\n', '-_');
 $$;
@@ -31,7 +30,7 @@ WITH
       WHEN algorithm = 'HS384' THEN 'sha384'
       WHEN algorithm = 'HS512' THEN 'sha512'
       ELSE '' END)  -- hmac throws error
-SELECT jwt.url_encode(hmac(signables, secret, (select * FROM alg)));
+SELECT jwt.url_encode(public.hmac(signables, secret, (select * FROM alg)));
 $$;
 
 
--- a/test/fixtures/privileges.sql
+++ b/test/fixtures/privileges.sql
@@ -81,3 +81,5 @@ GRANT UPDATE (article_id, user_id) ON TABLE limited_article_stars TO postgrest_t
 
 REVOKE EXECUTE ON FUNCTION privileged_hello(text) FROM PUBLIC; -- All functions are available to every role(PUBLIC) by default
 GRANT EXECUTE ON FUNCTION privileged_hello(text) TO postgrest_test_author;
+
+GRANT USAGE ON SCHEMA test TO postgrest_test_default_role;
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -112,16 +112,19 @@ CREATE TABLE items (
 );
 
 
-SET search_path = public, pg_catalog;
-
---
--- Name: always_true(test.items); Type: FUNCTION; Schema: public; Owner: -
---
-
 CREATE FUNCTION always_true(test.items) RETURNS boolean
     LANGUAGE sql STABLE
     AS $$ SELECT true $$;
 
+CREATE FUNCTION anti_id(test.items) RETURNS bigint
+    LANGUAGE sql STABLE
+    AS $_$ SELECT $1.id * -1 $_$;
+
+SET search_path = public, pg_catalog;
+
+CREATE FUNCTION always_false(test.items) RETURNS boolean
+    LANGUAGE sql STABLE
+    AS $$ SELECT false $$;
 
 create table public_consumers (
     id                  serial             not null unique,
@@ -135,16 +138,6 @@ create table public_orders (
     number              integer            not null,
     primary key (id)
 );
-
---
--- Name: anti_id(test.items); Type: FUNCTION; Schema: public; Owner: -
---
-
-CREATE FUNCTION anti_id(test.items) RETURNS bigint
-    LANGUAGE sql STABLE
-    AS $_$ SELECT $1.id * -1 $_$;
-
-
 
 SET search_path = تست, pg_catalog;
 


begriffs:
By the way, does this affect the pleasant side effect where setting `db-schema = ""` in the config file falls back to the search_path?

ruslantalpa:
from how the code looks, i think it affects it. we should keep that sideffect. when the value of schema is emptty string, there should be no "set schema" statement

ruslantalpa:
one more thing, since these functions live in the api schema, what happens if you call them directly? do they get displayed in the openapi output?

steve-chavez:
The computed columns get displayed in the openapi output, taking `always_true` as the example:
```json
{
  "/rpc/always_true": {
      "post": {
          "tags": [
              "(rpc) always_true"
          ],
          "produces": [
              "application/json",
              "application/vnd.pgrst.object+json"
          ],
          "parameters": [
              {
                  "required": true,
                  "schema": {
                      "type": "object"
                  },
                  "in": "body",
                  "name": "args"
              },
              {
                  "$ref": "#/parameters/preferParams"
              }
          ],
          "responses": {
              "200": {
                  "description": "OK"
              }
          }
      }
  }
}
```
In psql I'm able to call it as a function with both:
```sql
select test.always_true(row(1));
select test.always_true(i) from test.items i;
```   
However through PostgREST I'm not able to call them with a `GET/POST rpc/always_true` since it doesn't take a named argument. 
So it seems it's not useful to list the computed columns as `rpc` in the openapi output, I added a comment on #956 with a proposal for a different format.

steve-chavez:
The `db-schema = ""` doesn't work as before since that would also set the `search_path TO ''` ignoring the `authenticator` search_path.

If the idea of the `db-schema` was to add an extra layer of security allowing it to be `""` seems wrong to me(btw just tested that and it leaves the openapi output empty). I think that is a bug and it should be restricted to a non-empty value.

If we want to allow the user to use a `search_path` instead of a fixed schema I propose that we change the `db-schema` to `db-search-path` this would take values in the usual format like `public, test`, for this to work I'll have to remove the prepending of the config schema, that isn't required now anyway so it's doable without breaking previous functionality.

ruslantalpa:
it is intact doable, and yes that is the "right/flexible" way of doing it but here is the downside.
i guarantee that instead thinking about it and creating a clear separation for the public api, ppl will start just dumping in that config all the schemas and accessing the tables directly. sooner or later that is going to lead to security issues and guess who is going to get the blame.

The PostgREST way of building apis is new and i'd rather people (including us) learn to walk before they start running :) and the reality is there is not much db expertise around for people to truly understand the implications of search_path. 

I would argue that in the majority of the cases, it's not only enough to have one schema as a public api but in fact that is the better method. At least for now it'd rather have this "hack" remain undocumented and explained to users that really need the "feature".

i would say in this pr, jsut change the function to output an empty string/query when the param is empty to preserve the previous "hack".

steve-chavez:
Thing is we're not offering good support for users that need more than one schema with that hack, when doing the `db-schema = ""` and relying on the search_path the openapi output is broken(it doesn't list any tables or rpc's) and it's likely that there'll be more bugs(by the way I just ran the test suite with the config  changed to the empty db-schema with the `postgrest_test_authenticator` search_path changed to `public,test` and there are 61 failures).

Not reasonable to fix the bug's bugs for supporting that use case like that, for me it'd be better to give explicit support with `db-search-path`.

How about supporting both `db-schema` and `db-search-path`? We could make them exclude each other for the config to be valid, and the documentation would have the `db-schema` as the preferred way(the only one included in the tutorials) and `db-search-path` only mentioned as an advanced config option(with a warning), both would be validated to be nonempty.

begriffs:
> it is intact doable, and yes that is the "right/flexible" way of doing it

I think (assuming we're careful as I describe below) we should support the most elegant and flexible solution which is the search path.

> i guarantee that instead thinking about it and creating a clear separation for the public api, ppl will start just dumping in that config all the schemas and accessing the tables directly

Yeah...convenience does tempt people to do regrettable things. I think it's a matter of how we present it though, of guiding people to the safest way by default and only revealing the extra functionality as necessary.

> How about supporting both db-schema and db-search-path?

Actually what about supporting just `db-schema` like we do now, but treating comma separated values as a search path if someone knows about that feature? That way it's backward compatible and the docs can stay the same, but it provides the most flexibility for power users?

In another less prominent docs section about versioning we could describe how setting e.g. `db-schema="v2,v1"` can help reduce view duplication.

ruslantalpa:
i have no strong opinion about this,  if you guys think we should make `db-schema` to actually be the search-path value then i'm fine with it.
Maybe the only thing needed here is something like on boot, if the value has any commas then say something like
"It seems like you are exposing more then one schema as an api, this is discouraged and we hope you know what you are doing" ... and that would be warning enough.

@begriffs you write the message :)

begriffs:
Maybe a message like, "CAUTION: Multiple schemas detected in db-schema parameter. This might leak unintended information."

src/PostgREST/Middleware.hs
```diff
@@ -41,6 +41,7 @@ runWithClaims conf eClaims app req =
         claimsSql = map (pgFmtEnvVar "request.jwt.claim.") [(c,unquoted v) | (c,v) <- M.toList claimsWithRole]
         setRoleSql = maybeToList $
           (\r -> "set local role " <> r <> ";") . toS . pgFmtLit . unquoted <$> M.lookup "role" claimsWithRole
+        setSchemaSql = ["set schema " <> pgFmtLit (configSchema conf) <> ";"] :: [Text]
```
begriffs: The postgres [docs](
https://www.postgresql.org/docs/10/static/sql-set.html) say

> SET SCHEMA 'value' is an alias for SET search_path TO value. Only one schema can be specified using this syntax.

To allow the full generality of comma-separated search path items, we could change `configSchema` to a list that is split at config parsing time, then do something like:

```haskell
setSchemaSql = [
    "SET search_path TO " <>
    (intercalate ',' . map pgFmtIdent $ configSchema conf)) <> ";"
  ] :: [Text]
```
I didn't test this snippet, but you get the idea.

Also after parsing the config we can check whether `configSchema` has more than one item, and issue that warning text discussed in a previous comment.

begriffs: @steve-chavez does this look good to you? If you're busy, I can write the code for detecting multiple schema items and issuing an error.

steve-chavez: The snippet is good but I think just adding the  `SET search_path TO` to the `Middleware.hs` won't work(tests will fail), there's some code to be removed, mainly the `Schema` prepending. I already had an idea of how to do this but didn't have the time, I think I could finish this PR on the coming week, if I cannot finish I'll pass the ball, is that ok?

begriffs: Thanks, and yeah of course.

begriffs:
Insofar as it adds a `set schema` call this PR is good to merge. After it's merged I'll open one to try and remove the schema prefixing from names and support the full search_path.

steve-chavez:
@begriffs I was already working on it, on this branch https://github.com/steve-chavez/postgrest/commits/set-schema-2.

begriffs:
You rock! By all means carry on then. 😄

steve-chavez:
My bad, I thought this PR couldn't be merged because it was incomplete, but reviewing it again it seems is good to merge as it is, and the `search_path` is and additional feature that could be done on another PR.

@begriffs What about if I merge this branch and you continue where I left on this commit https://github.com/steve-chavez/postgrest/commit/e61579f170e0a1b7f06b9064734edd6839fbf711?

steve-chavez:
Since this PR has a breaking change(also the other one regarding memory usage) I think we could release a new version now(v0.4.4.0) and include the current PR changes in v0.5.0.0, that way we would be complying to semver.

begriffs:
Sounds good, let's revisit this after the next minor release.

This PR, and additional work for the more generalized search_path behavior can go into the next major version.

PR #1004, opened 2017-10-16
Remove hjsonpointer

PierreR:
It is a deps of hjsonchema and does not need to be pinned to a specific version.


begriffs:
Thanks!

Issue #690, opened 2016-08-07
Large memory usage for bulk inserts


begriffs:
I measured that PostgREST 0.3.2.0 uses memory proportional to 20x the size of a JSON payload during bulk inserts. This doesn't seem reasonable.

begriffs:
Note to self: try using [weigh](https://www.fpcomplete.com/blog/2016/05/weigh-package) in tests to measure the space used.

ruslantalpa:
Is this still a thing?
I remember it being a thing in #360 but that thread says it was fixed in 3.0

maximsch2:
I'm seeing similar problems with CSV uploads as well.

begriffs:
@maximsch2 would you be interested in helping us track down the problem? The issue that Ruslan linked to includes information about running how to run the binary with profiling enabled. If you're feeling brave you could do a similar thing and include a graph of memory allocation.

steve-chavez:
I have a proposal for avoiding the json payload processing in `INSERT` and clearing this out, I'll use the [articles table](https://github.com/begriffs/postgrest/blob/master/test/fixtures/schema.sql#L380-L384) as a reference.

Currently when receiving a payload like `[{"id": 11, "body": "uno"}, {"id": 12, "body": "dos"}]` we process this in Haskell to check if the json is an `array/object` to adequately use pg `json_populate_recordset/json_populate_record` functions.

This part can be done in pure SQL:
```sql
WITH payload AS(
-- This json would be replaced by $1
  SELECT '[{"id": 11, "body": "uno"}, {"id": 12, "body": "dos"}]'::json AS json_data 
),
vals AS(
  -- With this we make sure that the query works for json arrays or a single json object
  SELECT json_data AS val FROM payload WHERE json_typeof(json_data) = 'array'
  UNION ALL
  SELECT json_build_array(json_data) AS val FROM payload WHERE json_typeof(json_data) = 'object'
)
INSERT INTO articles(id, body) SELECT id, body from json_populate_recordset(null::articles, (select val from vals));
```

Now the other thing that the Haskell processing does is making sure the json has the same columns, not sure but I'm guessing the original purpose of this was to somehow protect against unintentional null values in a bulk insert? Anyway I think this would be better handled by the db not null contraints.

However this part also handles the `<cols>` specifying in  `ÌNSERT INTO table(<cols>) SELECT <cols> FROM ...` and that cannot be done in SQL, this is also necessary for handling column privileges.

[...]

PR #1005, opened 2017-10-22
Change RPC/UPDATE query to a parametrized query, avoid payload json encoding

### LINKED ISSUE #925: Out of memory during RPC call
Hi all,

after finishing my project I am experiencing many oom on my server running:
* postgrest
* postgresql
* openresty (nginx reverse proxy)

```
Aug  1 11:56:09 host-test01 kernel: [219140.781331] Out of memory: Kill process 27360 (postgrest) score 415 or sacrifice child
Aug  1 11:56:09 host-test01 kernel: [219140.781582] Killed process 27360 (postgrest) total-vm:1074990692kB, anon-rss:3395440kB, file-rss:0kB
Aug  1 11:56:09 host-test01 kernel: [219140.832559] systemd invoked oom-killer: gfp_mask=0x201da, order=0, oom_score_adj=0
Aug  1 11:56:09 host-test01 kernel: [219140.832563] systemd cpuset=/ mems_allowed=0
```

this is part of the kernel call trace. it says that postgrest will be sacrified to get free memory.

What am I doing:
* I am writing files to and from postgresql (in a table) chunk-wise
* mostly rpc calls (read_blob, write_blob)

hardware specs:
* 2 vCPUs (cores)
* 8GB RAM
* postgrest, openresty and postgresql in docker containers

any hints on what we can do here? where to start debugging and what to do to solve this problem?

starting the postgrest docker in attached mode shows the usual printing log lines (http requests) and as soon as the kernel kills the container I just drop out the container shell and the container stops running.

steve-chavez:
Fixes #925, the memory usage when posting a big payload on RPC now is practically the same as when doing an INSERT.

I'm using `json_to_record` pg function for this, it's available on pg >= 9.4, it's possible to maintain backwards compatibility by querying the pg version and keeping the previous custom input sanitizing for pg 9.3, but for now not sure if it's worth it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,6 +138,50 @@ jobs:
           name: run tests
           command: POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://circleci@localhost" postgrest_test) stack test
 
+  build-prof-test:
+    docker:
+      - image: circleci/buildpack-deps:trusty
+        environment:
+          - PGHOST=localhost
+          - TERM=xterm
+      - image: circleci/postgres:9.6.2
+        environment:
+          - POSTGRES_USER=circleci
+          - POSTGRES_DB=circleci
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - v1-stack-prof-dependencies-{{ checksum "postgrest.cabal" }}-{{ checksum "stack.yaml" }}
+      - run:
+          name: install stack & dependencies
+          command: |
+            curl -L https://github.com/commercialhaskell/stack/releases/download/v1.1.2/stack-1.1.2-linux-x86_64.tar.gz | tar zx -C /tmp
+            sudo mv /tmp/stack-1.1.2-linux-x86_64/stack /usr/bin
+            sudo apt-get update
+            sudo apt-get install -y libgmp-dev
+            sudo apt-get install -y postgresql-client
+            stack setup
+      - run:
+          name: build with profiling enabled
+          command: |
+            stack build --profile -j1
+      - run:
+          name: run memory usage tests
+          command: |
+            test/create_test_db "postgres://circleci@localhost" postgrest_test
+            psql "postgres:///postgrest_test" -f test/fixtures/database.sql
+            psql "postgres:///postgrest_test" -f test/fixtures/roles.sql
+            psql "postgres:///postgrest_test" -f test/fixtures/schema.sql
+            psql "postgres:///postgrest_test" -f test/fixtures/jwt.sql
+            psql "postgres:///postgrest_test" -f test/fixtures/privileges.sql
+            test/memory-tests.sh
+      - save_cache:
+          paths:
+            - "~/.stack"
+            - ".stack-work"
+          key: v1-stack-prof-dependencies-{{ checksum "postgrest.cabal" }}-{{ checksum "stack.yaml" }}
+
   centos6:
     <<: *build-distro-bin
 
@@ -188,10 +232,15 @@ workflows:
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
+      - build-prof-test:
+          filters:
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*/
       - centos6:
           requires:
             - build-test
             - build-test-9.6
+            - build-prof-test
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
@@ -201,6 +250,7 @@ workflows:
           requires:
             - build-test
             - build-test-9.6
+            - build-prof-test
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
@@ -210,6 +260,7 @@ workflows:
           requires:
             - build-test
             - build-test-9.6
+            - build-prof-test
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
@@ -219,6 +270,7 @@ workflows:
           requires:
             - build-test
             - build-test-9.6
+            - build-prof-test
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - #828, Fix computed column only working in public schema - @steve-chavez
+- #925, Avoid RPC high memory usage by using parametrized query - @steve-chavez
 
 ### Changed
 
 - Computed columns now only work if they belong to the db-schema - @steve-chavez
+- To use RPC now the `json_to_record/json_to_recordset` functions are needed, these are available starting from PostgreSQL 9.4 - @steve-chavez
 
 ## [0.4.4.0] - 2018-01-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Computed columns now only work if they belong to the db-schema - @steve-chavez
 - To use RPC now the `json_to_record/json_to_recordset` functions are needed, these are available starting from PostgreSQL 9.4 - @steve-chavez
+- Overloaded functions now depend on the `dbStructure`, restart/sighup may be needed for their correct functioning - @steve-chavez
 
 ## [0.4.4.0] - 2018-01-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - #828, Fix computed column only working in public schema - @steve-chavez
-- #925, Avoid RPC high memory usage by using parametrized query - @steve-chavez
+- #925, Fix RPC high memory usage by using parametrized query and avoiding json encoding - @steve-chavez
 
 ### Changed
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -95,8 +95,6 @@ data ApiRequest = ApiRequest {
   , iHeaders :: [(Text, Text)]
   -- | Request Cookies
   , iCookies :: [(Text, Text)]
-  -- | Rpc query params e.g. /rpc/name?param1=val1, similar to filter but with no operator(eq, lt..)
-  , iRpcQParams :: [(Text, Text)]
   }
 
 -- | Examines HTTP request and translates it into user intent.
@@ -116,7 +114,6 @@ userApiRequest schema req reqBody
       , iPreferSingleObjectParameter = singleObject
       , iPreferCount = hasPrefer "count=exact"
       , iFilters = filters
-      , iRpcQParams = rpcQParams
       , iLogic = [(toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, endingIn ["and", "or"] k ]
       , iSelect = toS $ fromMaybe "*" $ fromMaybe (Just "*") $ lookup "select" qParams
       , iOrder = [(toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, endingIn ["order"] k ]
@@ -130,6 +127,7 @@ userApiRequest schema req reqBody
       , iCookies = fromMaybe [] $ parseCookiesText <$> lookupHeader "Cookie"
       }
  where
+  -- rpcQParams = Rpc query params e.g. /rpc/name?param1=val1, similar to filter but with no operator(eq, lt..)
   (filters, rpcQParams) =
     case action of
       ActionInvoke{isReadOnly=True} -> partition (liftM2 (||) (isEmbedPath . fst) (hasOperator . snd)) flts
@@ -141,20 +139,22 @@ userApiRequest schema req reqBody
   isEmbedPath = T.isInfixOf "."
   isTargetingProc = fromMaybe False $ (== "rpc") <$> listToMaybe path
   payload =
-    case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
-      CTApplicationJSON ->
-        note "All object keys must match" . consPayloadJSON reqBody
+    case (decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type", action) of
+      (_, ActionInvoke{isReadOnly=True}) ->
+        Right $ PayloadJSON (JSON.encode $ M.fromList $ second JSON.toJSON <$> rpcQParams) PJObject (S.fromList $ fst <$> rpcQParams)
+      (CTApplicationJSON, _) ->
+        note "All object keys must match" . payloadAttributes reqBody
           =<< if BL.null reqBody && isTargetingProc
                then Right emptyObject
                else JSON.eitherDecode reqBody
-      CTTextCSV -> do
+      (CTTextCSV, _) -> do
         json <- csvToJson <$> CSV.decodeByName reqBody
-        note "All lines must have same number of fields" $ consPayloadJSON (JSON.encode json) json
-      CTOther "application/x-www-form-urlencoded" ->
+        note "All lines must have same number of fields" $ payloadAttributes (JSON.encode json) json
+      (CTOther "application/x-www-form-urlencoded", _) ->
         let json = M.fromList . map (toS *** JSON.String . toS) . parseSimpleQuery $ toS reqBody
             keys = S.fromList $ M.keys json in
         Right $ PayloadJSON (JSON.encode json) PJObject keys
-      ct ->
+      (ct, _) ->
         Left $ toS $ "Content-Type not acceptable: " <> toMime ct
   topLevelRange = fromMaybe allRange $ M.lookup "limit" ranges
   action =
@@ -177,9 +177,8 @@ userApiRequest schema req reqBody
               ["rpc", proc] -> TargetProc
                               $ QualifiedIdentifier schema proc
               other         -> TargetUnknown other
-  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke{isReadOnly=False}]
-  relevantPayload | action == ActionInvoke{isReadOnly=True} = Nothing
-                  | shouldParsePayload = rightToMaybe payload
+  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke{isReadOnly=False}, ActionInvoke{isReadOnly=True}]
+  relevantPayload | shouldParsePayload = rightToMaybe payload
                   | otherwise = Nothing
   path            = pathInfo req
   method          = requestMethod req
@@ -274,8 +273,8 @@ csvToJson (_, vals) =
           else JSON.String $ toS str
       )
 
-consPayloadJSON :: BL.ByteString -> JSON.Value -> Maybe PayloadJSON
-consPayloadJSON raw json =
+payloadAttributes :: RequestBody -> JSON.Value -> Maybe PayloadJSON
+payloadAttributes raw json =
   -- Test that Array contains only Objects having the same keys
   case json of
     JSON.Array arr ->
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -13,7 +13,7 @@ module PostgREST.ApiRequest ( ApiRequest(..)
 
 import           Protolude
 import qualified Data.Aeson                as JSON
-import           Data.Aeson.Types          (emptyObject)
+import           Data.Aeson.Types          (emptyObject, emptyArray)
 import qualified Data.ByteString           as BS
 import qualified Data.ByteString.Internal  as BS (c2w)
 import qualified Data.ByteString.Lazy      as BL
@@ -33,14 +33,7 @@ import           Network.Wai               (Request (..))
 import           Network.Wai.Parse         (parseHttpAccept)
 import           PostgREST.RangeQuery      (NonnegRange, rangeRequested, restrictRange, rangeGeq, allRange, rangeLimit, rangeOffset)
 import           Data.Ranged.Boundaries
-import           PostgREST.Types           ( QualifiedIdentifier (..)
-                                           , Schema
-                                           , PayloadJSON(..)
-                                           , ContentType(..)
-                                           , ApiRequestError(..)
-                                           , toMime
-                                           , operators
-                                           , ftsOperators)
+import           PostgREST.Types
 import           Data.Ranged.Ranges        (Range(..), rangeIntersection, emptyRange)
 import qualified Data.CaseInsensitive      as CI
 import           Web.Cookie                (parseCookiesText)
@@ -150,17 +143,17 @@ userApiRequest schema req reqBody
   payload =
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
       CTApplicationJSON ->
-        note "All object keys must match" . ensureUniform . pluralize
+        note "All object keys must match" . consPayloadJSON reqBody
           =<< if BL.null reqBody && isTargetingProc
                then Right emptyObject
                else JSON.eitherDecode reqBody
-      CTTextCSV ->
-        note "All lines must have same number of fields" . ensureUniform . csvToJson
-          =<< CSV.decodeByName reqBody
+      CTTextCSV -> do
+        json <- csvToJson <$> CSV.decodeByName reqBody
+        note "All lines must have same number of fields" $ consPayloadJSON (JSON.encode json) json
       CTOther "application/x-www-form-urlencoded" ->
-        Right . PayloadJSON . V.singleton . M.fromList
-                    . map (toS *** JSON.String . toS) . parseSimpleQuery
-                    $ toS reqBody
+        let json = M.fromList . map (toS *** JSON.String . toS) . parseSimpleQuery $ toS reqBody
+            keys = S.fromList $ M.keys json in
+        Right $ PayloadJSON (JSON.encode json) PJObject keys
       ct ->
         Left $ toS $ "Content-Type not acceptable: " <> toMime ct
   topLevelRange = fromMaybe allRange $ M.lookup "limit" ranges
@@ -270,9 +263,9 @@ type CsvData = V.Vector (M.HashMap Text BL.ByteString)
   The reason for its odd signature is so that it can compose
   directly with CSV.decodeByName
 -}
-csvToJson :: (CSV.Header, CsvData) -> JSON.Array
+csvToJson :: (CSV.Header, CsvData) -> JSON.Value
 csvToJson (_, vals) =
-  V.map rowToJsonObj vals
+  JSON.Array $ V.map rowToJsonObj vals
  where
   rowToJsonObj = JSON.Object .
     M.map (\str ->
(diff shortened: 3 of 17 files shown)


src/PostgREST/App.hs
```diff
@@ -241,11 +241,12 @@ app dbStructure conf apiRequest =
             Right ((q, cq), bField, params) -> do
               let prms = case payload of
                           Just (PayloadJSON pld) -> V.head pld
-                          Nothing -> M.fromList $ second toJSON <$> params -- toJSON is just for reusing the callProc function
+                          Nothing -> M.fromList $ second toJSON <$> params
                   singular = contentType == CTSingularJSON
                   paramsAsSingleObject = iPreferSingleObjectParameter apiRequest
-              row <- H.query () $
-                callProc qi prms returnsScalar q cq shouldCount
+                  specifiedPgArgs = filter (flip M.member prms . pgaName) $ fromMaybe [] (pdArgs <$> proc)
```
begriffs: So `specifiedPgArgs` strips away json keys that don't match the param names that postgrest has cached in dbStructure for the given proc?

Can postgresql overload a single function name with procedures taking different numbers of arguments, and if so would the logic here prevent selecting the desired proc through choice of arguments?

begriffs:
What a relief that you fixed the RPC memory problem!

> I'm using json_to_record pg function for this, it's available on pg >= 9.4

Yeah this is interesting. Half of me thinks that a breaking change like this calls for a major version bump, but the other half says that postgres 9.3 was released back in 2013 and it's up to version 10 now so it's less likely that the change will disrupt anyone.

If we do want to proceed with this change as-is, we should add a "Changed" entry to the changelog similar to how we did in the 4.3 release. Also would it make sense to make the minimum version checker on startup check for >=9.4? That way people would find out quickly that there is a problem rather than getting sql errors later on during RPC calls.

src/PostgREST/App.hs
steve-chavez: My intention with that variable was to avoid specifying types for the parameters that aren't arguments so that functions with DEFAULT parameters work as normal, e.g. if the `filter` was not applied to `pdArgs` when calling this [function](https://github.com/begriffs/postgrest/blob/master/test/fixtures/schema.sql#L1292-L1294)  with `GET /rpc/many_inout_params?num=3&str=smth` it would give `[{"b": null,"num": 3,"str": "smth"}]` instead of the correct `[{"b": true, "num": 3, "str": "smth"}]`.

Now as you noticed this new way of specifying the arguments doesn't work for overloaded functions, I didn't realized this until now(tests would've made it clear, I'll add those if I get it to work), previous version was working due to the `::unknown` casting, for trying something similar in the new query I did:
```sql
with args as (
  select * from json_to_record('{"num":1, "str": "smth", "b": true}') as _(num unknown, str unknown, b unknown)
)
select test.many_inout_params(num := (select num from args), str := (select str from args), b:= (select b from args));
```
But postgres cannot infer the types and gives: `ERROR:  XX000: failed to find conversion function from unknown to integer`, so it cannot be done in that way.

For now, I'll be trying to specify the types as unknowns at the `hasql` level.

steve-chavez:
I noticed that UPDATE also didn't use a parametrized query and had high memory usage for the same tests I did on RPC and INSERT, so I went ahead and changed the UPDATE query to one with the following structure:
```sql
update clients set "id" = _."id", "name" = _."name" from (select * from json_populate_record(NULL::clients,'{"id": 4, "name": "Amazon"}')) _ where "clients".id = 3;
```
The `$1` would be the arg of `json_populate_record`. Now, with the same table and file I provided [here](https://github.com/begriffs/postgrest/issues/925#issuecomment-338330239) I did:
```bash
echo "{\"blob\": \"$(cat 100mb.json)\"}" |  curl -v -H "Content-Type: application/json" --request PATCH -d @- http://localhost:3000/leak\?id\=eq.1
```
And for the old UPDATE query I got a `27,752,281,816 bytes` of total allocation. 
<details>
<summary>Old UPDATE query profile report</summary>
<pre>
     postgrest +RTS -N -I2 -p -h -RTS config

  total time  =        8.44 secs   (8439 ticks @ 1000 us, 1 processor)
  total alloc = 27,752,281,816 bytes  (excludes profiling overheads)

COST CENTRE     MODULE                         SRC                                             %time %alloc

trimNullChars   PostgREST.QueryBuilder         src/PostgREST/QueryBuilder.hs:478:1-38           26.4   38.8
toQQ            Text.InterpolatedString.Perl6  src/Text/InterpolatedString/Perl6.hs:200:5-29    25.1   29.3
unquoted        Data.Aeson.Encoding.Builder    Data/Aeson/Encoding/Builder.hs:119:1-47          20.9   17.9
">>=.\.succ'"   Data.Attoparsec.Internal.Types Data/Attoparsec/Internal/Types.hs:146:13-76       7.7    1.1
append.x        Data.Text                      Data/Text.hs:(464,7)-(468,18)                     5.9    8.2
replace.ixs     Data.Text                      Data/Text.hs:691:5-33                             5.7    0.0
pgFmtLit        PostgREST.QueryBuilder         src/PostgREST/QueryBuilder.hs:(197,1)-(203,15)    2.5    0.0
hashPtrWithSalt Data.Hashable.Class            Data/Hashable/Class.hs:(714,1)-(716,23)           2.5    0.0
encodeUtf8      Data.Text.Encoding             Data/Text/Encoding.hs:(373,1)-(388,37)            0.0    1.5
</pre>
</details>
</br>

With the new query for UPDATE I get `5,871,973,344 bytes` of total allocation, similar to the INSERT and new RPC.
 
<details>
<summary>New UPDATE query profile report</summary>
<pre>
     postgrest +RTS -N -I2 -p -h -RTS config

  total time  =        2.50 secs   (2500 ticks @ 1000 us, 1 processor)
  total alloc = 5,871,973,344 bytes  (excludes profiling overheads)

COST CENTRE           MODULE                            SRC                                                     %time %alloc

unquoted              Data.Aeson.Encoding.Builder       Data/Aeson/Encoding/Builder.hs:119:1-47                  67.2   84.6
">>=.\.succ'"         Data.Attoparsec.Internal.Types    Data/Attoparsec/Internal/Types.hs:146:13-76              26.5    5.3
append.\.\.\          Data.Attoparsec.ByteString.Buffer Data/Attoparsec/ByteString/Buffer.hs:(127,40)-(134,56)    1.6    0.0
receiveloop           Network.Wai.Handler.Warp.Recv     Network/Wai/Handler/Warp/Recv.hs:(115,1)-(129,20)         1.5    0.0
sendQueryPrepared.\.\ Database.PostgreSQL.LibPQ         src/Database/PostgreSQL/LibPQ.hsc:(1705,16)-(1709,69)     0.9    1.8
append.\.\            Data.Attoparsec.ByteString.Buffer Data/Attoparsec/ByteString/Buffer.hs:(110,35)-(134,56)    0.6    4.6
encode                Data.Aeson                        Data/Aeson.hs:128:1-46                                    0.1    1.8
nullableValue.\.\     Hasql.Private.Encoders.Params     library/Hasql/Private/Encoders/Params.hs:51:37-78         0.0    1.8
</pre>
</details>

src/PostgREST/App.hs
steve-chavez: Did a PR on hasql for handling this case, if it gets accepted it would not be necessary to use `json_to_record`, we would just use the `unknown` encoder for passing the args to the function.

ruslantalpa:
the remaining memory consumption comes comes from parsing json into memory.
we don't really do anything with that structure (in case of update and rpc we needed that to go over each key, now we dont). with how things stand in the context of the current pr, the only thing we do wi the json in in case of insert.
we go over each row and validate that the "keys/columns" are the same.
if we can skip that, then there is really no reason to do a "full parse" of the json. we could somehow do a "light" validation that it conforms to some basic structure (no sure exactly what that would mean) and then send it off to the db.

so the question is, do we really need to have in case of insert, all the rows to have the same structure, can the query be in some other way or maybe do we trust the user to send the right format and in case he does not, the query fails at the db level?

src/PostgREST/App.hs
begriffs: Cool!

begriffs:
@ruslantalpa yeah letting the db handle validation for non-uniform keys would certainly save our code that processing.

Maybe there's a cheap way to construct `PayloadJSON` in the JSON case to avoid the uniformity check:
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/ApiRequest.hs#L157-L169

I'm curious to see what the db error looks like too in the case of non-uniform keys. Hopefully it's not too cryptic, and if it is maybe we can intercept a certain error code to improve the message.

src/PostgREST/App.hs
begriffs: Woo hoo, looks like hasql 1.1 is now available and includes your change!

ruslantalpa: while that is a good thing, in this case as we discussed bellow, to take advantage of that, we'd need to parse the json and maybe it makes sense to avoid that (then entire body as a single param).
what hasql 1.1 can help with (in the future) is to transition to real prepared statements.

steve-chavez:
Unfortunately the db doesn't err in the case of non-uniform keys, in the function `json_populate_recordset(null::myrowtype, {json})` the `null::rowtype` coerces the json to its specified columns ignoring the unspecified ones, I've looked into the possibility of enforcing the uniform keys in sql but it seems that approach is too complex.

However what I've noticed is that most of the memory usage doesn't come from the `ensureUniform` haskell function but instead it comes from the `Aeson.encode` function that gets used when encoding the json through Hasql in [encodeUniformObjs](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/QueryBuilder.hs#L87)(Hasql uses `encode` [here](https://github.com/nikita-volkov/postgresql-binary/blob/master/library/PostgreSQL/Binary/Encoding/Builders.hs#L348)), this also can be seen in the previous .prof I posted.

To avoid the `encode` what I did was to use the new hasql `unknown` encoder(so we can pass the plain ByteString without encoding and let PostgreSQL figure it out) but still keep `ensureUniform` function as validation(we still decode but not encode) and have a cached unecoded json ByteString in PayloadJSON, all of this can be seen on this [branch](https://github.com/steve-chavez/postgrest/commits/rpc-memory-no-encode-2)(bear in mind this is not yet finished/refactored, just to show the results), as a result the memory usage for the previous UPDATE test gets reduced to almost `900MB`, great reduction compared to the previous almost `6GB`, the .prof:

<details>
<summary>UPDATE query without encode profile report</summary>
<pre>
	   postgrest +RTS -N -I2 -p -h -RTS config

	total time  =        0.85 secs   (850 ticks @ 1000 us, 1 processor)
	total alloc = 904,808,136 bytes  (excludes profiling overheads)

COST CENTRE               MODULE                            SRC                                                     %time %alloc

">>=.\.succ'"               Data.Attoparsec.Internal.Types    Data/Attoparsec/Internal/Types.hs:146:13-76              72.6   34.5
append.\.\.\              Data.Attoparsec.ByteString.Buffer Data/Attoparsec/ByteString/Buffer.hs:(127,40)-(134,56)    4.9    0.0
sendQueryPrepared.\.\     Database.PostgreSQL.LibPQ         src/Database/PostgreSQL/LibPQ.hsc:(1705,16)-(1709,69)     3.6   11.4
receiveloop               Network.Wai.Handler.Warp.Recv     Network/Wai/Handler/Warp/Recv.hs:(115,1)-(129,20)         3.6    0.0
nullableValue.\.\         Hasql.Private.Encoders.Params     library/Hasql/Private/Encoders/Params.hs:56:37-78         3.5   11.4
app                       PostgREST.App                     src/PostgREST/App.hs:(109,1)-(310,58)                     2.7   11.5
weekDays.(...)            Network.HTTP.Date.Formatter       Network/HTTP/Date/Formatter.hs:83:16-54                   2.4    0.0
append.\.\                Data.Attoparsec.ByteString.Buffer Data/Attoparsec/ByteString/Buffer.hs:(110,35)-(134,56)    1.9   29.8
receive                   Network.Wai.Handler.Warp.Recv     Network/Wai/Handler/Warp/Recv.hs:(96,1)-(99,48)           1.5    0.3
chunkedTransferTerminator Blaze.ByteString.Builder.HTTP     Blaze/ByteString/Builder/HTTP.hs:202:1-54                 1.2    0.0
</pre>
</details>

ruslantalpa:
this is good, with that change i say ok to merge (also check if you can use the unknown encoder) for insert/update (can't remember if they have the same issue as rpc)

begriffs:
Nice research, it's certainly an improvement! Clean up your branch as you think proper, and we can review it and get it merged.

In the future we could maybe look into possibilities for using a haskell Conduit for places postgrest needs to look through the json. I know you said ensureUniform isn't the main memory culprit, but it might benefit from it.

steve-chavez:
I decided to use `json_to_record` function to keep the `900mb` max usage for the test on RPC POST.
This is because I couldn't make Aeson to decode directly to a `ByteString` without doing an `encodeUtf8`(this also contributed to the high memory usage when testing, also see [this](https://github.com/haskell/text/issues/194)), as far as I've seen this isn't possible with Aeson([the ByteString instances were removed](https://github.com/bos/aeson/issues/187) and before the removal [those instances used encodeUtf8](https://github.com/bos/aeson/commit/9ee73f303cacb8ae73b2325f2f47288522607420#diff-7769cc48c0cd5956c0d28805dcf2f1a8L482).

I did some more tests with `1mb` and `10mb` files, in summary the memory usage:

- Without parametrized query(RPC/UPDATE on master): 
```
1mb = 220,565,432 - 10mb = 2,190,870,088 - 100mb = 22,266,823,688 
Around x220 payload size
```
- With Aeson encoding and with parametrized query(INSERT on master): 
```
1mb = 61,873,424 - 10mb = 592,780,848 - 100mb = 5,872,095,496
Around x60 payload size
```
- Without Aeson encoding and with parametrized query(RPC/UPDATE/INSERT on this PR): 
```
1mb = 13,280,792 - 10mb = 101,057,976 - 100mb = 904,757,664
Around x10 payload size
```

@begriffs This is ready for another review, also in #690 you mentioned you wanted to use [weigh](https://www.fpcomplete.com/blog/2016/05/weigh-package), it would be great if you could add some tests using that, that would ensure we don't get performance regressions.

Regarding Conduit, I've seen that `LinearTypes` is being worked on GHC, hopefully it'll arrive soon and that'll provide a cleaner way to have constant memory usage, just linear arrows `->.` instead of more abstractions that I think would make the code more complex.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #974, Fix RPC error when function has single OUT param - @steve-chavez
 - #1021, Reduce join size in allColumns for faster program start - @nextstopsun
 - #411, Remove the need for pk in &select for parent embed - @steve-chavez
+- #925, Fix RPC high memory usage by using parametrized query - @steve-chavez
```
begriffs: Is it also worth noting how you changed the behavior with overloaded procs?

src/PostgREST/ApiRequest.hs
```diff
@@ -184,9 +179,8 @@ userApiRequest schema req reqBody
               ["rpc", proc] -> TargetProc
                               $ QualifiedIdentifier schema proc
               other         -> TargetUnknown other
-  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke{isReadOnly=False}]
-  relevantPayload | action == ActionInvoke{isReadOnly=True} = Nothing
-                  | shouldParsePayload = rightToMaybe payload
+  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke{isReadOnly=False}, ActionInvoke{isReadOnly=True}]
+  relevantPayload | shouldParsePayload = rightToMaybe payload
```
begriffs: Wondering why we still have the notion of `relevantPayload`? AFAICT lazy evaluation would allow us to set `iPayload` directly to `payload`, which would not get evaluated if the app didn't request it.

src/PostgREST/ApiRequest.hs
```diff
@@ -184,9 +179,8 @@ userApiRequest schema req reqBody
               ["rpc", proc] -> TargetProc
                               $ QualifiedIdentifier schema proc
               other         -> TargetUnknown other
-  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke{isReadOnly=False}]
-  relevantPayload | action == ActionInvoke{isReadOnly=True} = Nothing
-                  | shouldParsePayload = rightToMaybe payload
+  shouldParsePayload = action `elem` [ActionCreate, ActionUpdate, ActionInvoke{isReadOnly=False}, ActionInvoke{isReadOnly=True}]
```
begriffs: We *might* think about removing `shouldParsePayload` although that may cause the `InvalidBody` error in cases where the payload was previously ignored, and would force parsing the payload on every request.

src/PostgREST/App.hs
```diff
@@ -305,7 +315,6 @@ app dbStructure conf apiRequest =
       fldNames = fieldNames <$> readReq
       readDbRequest = DbRead <$> readReq
       mutateDbRequest = DbMutate <$> (mutateRequest apiRequest =<< fldNames)
-      rpcQParams = readRpcRequest apiRequest
```
begriffs: It's great that this is unified in the payload now. Probably should delete `iRpcQParams` from the `ApiRequest` data type as well.

src/PostgREST/ApiRequest.hs
```diff
@@ -281,27 +275,26 @@ csvToJson (_, vals) =
           else JSON.String $ toS str
       )
 
--- | Convert {foo} to [{foo}], leave arrays unchanged
--- and truncate everything else to an empty array.
-pluralize :: JSON.Value -> JSON.Array
-pluralize obj@(JSON.Object _) = V.singleton obj
-pluralize (JSON.Array arr)    = arr
-pluralize _                   = V.empty
+consPayloadJSON :: BL.ByteString -> JSON.Value -> Maybe PayloadJSON
```
begriffs: Initially the name `consPayloadJson` confused me because `cons` makes me think of a list operation (like `cons` and `cdr`). I guess it means "construct?" What about another name like `payloadAttributes` or something? Also the first argument could be of type RequestBody

src/PostgREST/App.hs
```diff
@@ -96,8 +95,8 @@ transactionMode structure target action =
     ActionInvoke{isReadOnly=False} ->
       let proc =
             case target of
-              (TargetProc qi) -> M.lookup (qiName qi) $
-                                   dbProcs structure
+              -- TODO: this proc search won't be accurate when the function is overloaded
```
begriffs: Although it's not perfect, you could improve accuracy by creating a set of the `pdVolatility` values for all matching functions, and then choose the minimal value in that set as defined by the Ord instance of `ProcVolatility`. (I think minimal is what I mean…basically Volatile if any are Volatile, going on up the list to Immutable if all are Immutable).

src/PostgREST/App.hs
```diff
@@ -136,22 +135,24 @@ app dbStructure conf apiRequest =
                       )
                     ] (toS body)
 
-        (ActionCreate, TargetIdent (QualifiedIdentifier _ table), Just payload@(PayloadJSON rows)) ->
+        (ActionCreate, TargetIdent (QualifiedIdentifier _ table), Just (PayloadJSON payload pType _)) ->
           case mutateSqlParts of
             Left errorResponse -> return errorResponse
             Right (sq, mq) -> do
-              let isSingle = (==1) $ V.length rows
+              let (isSingle, rows) = case pType of
```
begriffs: Maybe rename `rows` to `nRows` since it's just the count. Makes it clearer later on.

src/PostgREST/App.hs
```diff
@@ -136,22 +135,24 @@ app dbStructure conf apiRequest =
...
-                then return $ singularityError (toInteger $ V.length rows)
+                then return $ singularityError (toInteger rows)
                 else do
                   let pKeys = map pkName $ filter (filterPk schema table) allPrKeys -- would it be ok to move primary key detection in the query itself?
                       stm = createWriteStatement sq mq
                         (contentType == CTSingularJSON) isSingle
                         (contentType == CTTextCSV) (iPreferRepresentation apiRequest)
                         pKeys
-                  row <- H.query payload stm
+                  row <- H.query (toS payload) stm
```
begriffs: At least `toS` is going from strict to lazy bytestring which I don't think uses lots of extra memory. Just puts the string into a single element list I believe.

src/PostgREST/App.hs
```diff
@@ -227,33 +227,43 @@ app dbStructure conf apiRequest =
...
-                          Nothing -> M.fromList $ second toJSON <$> params -- toJSON is just for reusing the callProc function
+            Right ((q, cq), bField) -> do
+              let isObject = case pType of
+                                PJObject  -> True
+                                PJArray _ -> False
                   singular = contentType == CTSingularJSON
-                  paramsAsSingleObject = iPreferSingleObjectParameter apiRequest
-              row <- H.query () $
-                callProc qi prms returnsScalar q cq shouldCount
+                  specifiedPgArgs = filter (flip S.member pKeys . pgaName) $ fromMaybe [] (pdArgs <$> proc)
```
begriffs: This is a minor nitpick, but I find it slightly clearer without `flip`:

```haskell
specifiedPgArgs = filter ((`S.member` pKeys) . pgaName) $ fromMaybe [] (pdArgs <$> proc)
```

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -50,7 +50,8 @@ readRequest maxRows allRels allProcs apiRequest  =
         (TargetIdent (QualifiedIdentifier s t) ) -> Just (s, t)
         (TargetProc  (QualifiedIdentifier s proc) ) -> Just (s, tName)
           where
-            retType = pdReturnType <$> M.lookup proc allProcs
+            -- TODO: this proc search won't be accurate when the function is overloaded
```
begriffs: Could we check whether all matching procs have Single retvals? If so, choose Singular else SetOf? It's not perfect but it could improve the case if the overloaded procs are somewhat similar.

src/PostgREST/QueryBuilder.hs
```diff
@@ -142,35 +134,47 @@ createWriteStatement selectQuery mutateQuery wantSingle wantHdrs asCsv rep pKeys
...
        SELECT
          {countResultF} AS total_result_set,
          pg_catalog.count(_postgrest_t) AS page_total,
          {bodyF} AS body,
          {responseHeaders} AS response_headers
        FROM ({selectQuery}) _postgrest_t;|]
 
+    (argsRecord, args) | paramsAsSingleObject = ("_args_record AS (SELECT NULL)", "$1::json")
+                       | null pgArgs = (ignoredBody, "")
+                       | otherwise = (
```
begriffs: Lots of concatenation here, might use `[qc|` like other code in this file. You could see if it makes it any cleaner.

CHANGELOG.md
steve-chavez: In which way?(I'm assuming changelog is related to functionality for end users)

The detection of the `return type/volatility` is working the same as before(HashMap used to merge duplicates `ProcDescription` into just one) and the `callProc` function considers overloaded functions(that previously were undocumented/untested).

src/PostgREST/App.hs
steve-chavez: I thought of leaving it as a TODO and see if later after a refactor I could pass the `PayloadJSON` keys so I can make the proc search accurate, thought it would make no harm since prior to this change the proc search also didn't consider the overloaded functions case.

But I think I can do as you said for now to somewhat improve the accuracy.

src/PostgREST/App.hs
```diff
@@ -227,33 +227,43 @@ app dbStructure conf apiRequest =
...
               return $ responseLBS status200 [allOrigins, acceptH] ""
 
-        (ActionInvoke _isReadOnly, TargetProc qi, payload) ->
-          let proc = M.lookup (qiName qi) allProcs
+        (ActionInvoke _, TargetProc qi, Just (PayloadJSON payload pType pKeys)) ->
+          let paramsAsSingleObject = iPreferSingleObjectParameter apiRequest
+              procs = M.lookup (qiName qi) allProcs
+              proc = join $
+                -- Handle overloaded functions case
+                (case length <$> procs of
```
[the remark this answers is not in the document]
steve-chavez: It will work but the idea was to immediately get the `ProcDescription` if the function is not overloaded(majority of cases) and avoid the extra steps, creating a Set and comparing two Sets, do you think that is not worth to do that?

src/PostgREST/QueryBuilder.hs
steve-chavez: I came up with this:
```haskell
[qc|_args_record AS (
   SELECT * FROM {(if isObject then "json_to_record" else "json_to_recordset")::Text}($1) 
   AS _( {(intercalate ", " ((\a -> pgaName a <> " " <> pgaType a) <$> pgArgs))::Text} )
 )|]
```
The second expression doesn't compile(`SrcLoc "" 1 55 Parse error in pattern`), when I put that expression in a variable it does compile. It seems unclearer for me in that way, I'll use `unwords` to reduce concatenations.

CHANGELOG.md
begriffs: Nevermind, if it caused no visible change then nothing to document. I wasn't sure how it might affect using overloaded functions.

src/PostgREST/QueryBuilder.hs
begriffs: OK, sounds good.

src/PostgREST/App.hs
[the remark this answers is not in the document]
begriffs: I see what you mean. Well maybe update the comment to explain that we're doing an optimization for the length=1 case.

src/PostgREST/ApiRequest.hs
steve-chavez: I think if we remove that some of the pattern matching relying on a `Nothing` PayloadJSON in `App.hs` would not work. I think that part would be better refactored in another PR as it's not tied directly to this PR changes, also it already took me a lot of work to refactor the PayloadJSON and I'd rather not touch this part for now.

src/PostgREST/App.hs
steve-chavez: On second thought it'll be better if I refactor this part now, that way the proc search is accurate.

src/PostgREST/App.hs
steve-chavez: `iRpcQParams` deleted.

steve-chavez:
Ended up adding memory usage tests in bash, doing it in haskell seemed more difficult.

I just did a [grep](https://github.com/begriffs/postgrest/pull/1005/commits/f3c0bdef9dea9492c54168a93b368eec67ed6020#diff-7b488460bf6e9e6b901065d1626b8500R45) of the generated `postgrest.prof` file `total alloc`, I think this is good enough for now.

begriffs:
Nice work with the automated memory testing, that will give us some peace of mind.

I'll give the rest another review in the next day or two and we can get it wrapped up.

.circleci/config.yml
```diff
@@ -188,10 +232,15 @@ workflows:
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
+      - build-prof-test:
+          filters:
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*/
```
begriffs: Running the memory testing only on a release tag might be too late. By that point won't our build system already have built new binaries?

If the memory test doesn't take too long we should probably have it run all the time on CI to immediately catch any regression.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1021, Reduce join size in allColumns for faster program start - @nextstopsun
 - #411, Remove the need for pk in &select for parent embed - @steve-chavez
 - #1016, Fix anonymous requests when configured with jwt-aud - @ruslantalpa
+- #925, Fix RPC high memory usage by using parametrized query - @steve-chavez
```
begriffs: Anything to add about the handling of overloaded functions as well?

begriffs:
Aside from my questions this looks good to go!

.circleci/config.yml
steve-chavez: It's not descriptive but that config means that on tag push circle will also run the `build-prof-test` job(if `filters` is not added the job will be skipped when releasing), see the CircleCI v2 docs on [that](https://circleci.com/docs/2.0/configuration-reference/#tags).

So the job will run when doing any branch push, you can also see [here](https://circleci.com/gh/steve-chavez/postgrest/590?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that that job ran on the latest commit on this branch.

CHANGELOG.md
steve-chavez: I mentioned earlier [here](https://github.com/begriffs/postgrest/pull/1005#discussion_r155926177) that there wasn't a behavior change related to overloaded functions.

However thinking it again, now the overloaded functions depend on the `dbStructure` to be correctly detected, guess I could add that under a `Changed` header in the CHANGELOG.

steve-chavez:
As it is this PR has a breaking change(using `json_to_record`) so it can't be part of `v0.4.4.0` I think we should release and after that merge this.

Also I want to emphasize that the memory usage improvements on this PR are directly related to a big json value(`{"key": "10mb val"}`) and not to big json arrays like the ones used in bulk inserts(#690), I've measured that the memory usage for that case remains the same when comparing current master and this PR.

Right now with a 2.3M json payload(100000 array values, two cols object) I get around `41GB` memory usage and the uploading takes like `50s`, I've made some improvements on this reducing the memory usage to `12MB` and  `0.349s` however this involves some tradeoffs, so it'll be better to discuss it in another PR, I'll open a new one when the current is merged.

steve-chavez:
Added the change about overloaded functions in the CHANGELOG, @begriffs see if it's clear and then I'll merge.

begriffs:
Yep, the changelog note makes sense.

PR #1007, opened 2017-11-02
Change phrase/plain full text search syntax

steve-chavez:
Changes the syntax for `phrase/plain` introduced in #964.

Previous syntax:
```http
GET /tsearch?tsv=french.fts.amusant
GET /tsearch?tsv=plain.german.fts.Art%20Spass
GET /tsearch?tsv=phrase.english.fts.The%20Fat%20Cats
```
New syntax:
```http
GET /tsearch?tsv=fts(french).amusant
GET /tsearch?tsv=plfts(german).Art%20Spass
GET /tsearch?tsv=phfts(english).The%20Fat%20Cats
```
New syntax is a bit shorter, it's more alike to the pg `{plain/phrase}to_tsquery` functions in that the first arg is the language(previous one has the language between plain/phrase and fts) and also it's more consistent because dots `.` are currently used like separators and the `(language)` fits better as an optional argument.

Also have to mention that having `not.english..` was kinda confusing inside `and/or` operations, I think overall this new syntax is better and in addition it makes the code simpler.

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -40,7 +40,7 @@ import           PostgREST.Types           ( QualifiedIdentifier (..)
                                            , ApiRequestError(..)
                                            , toMime
                                            , operators
-                                           , FtsMode(..))
+                                           , ftsOperators)
 import           Data.Ranged.Ranges        (Range(..), rangeIntersection, emptyRange)
 import qualified Data.CaseInsensitive      as CI
 import           Web.Cookie                (parseCookiesText)
@@ -142,16 +142,10 @@ userApiRequest schema req reqBody
       ActionInvoke{isReadOnly=True} -> partition (liftM2 (||) (isEmbedPath . fst) (hasOperator . snd)) flts
       _ -> (flts, [])
   flts = [ (toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn ["order", "limit", "offset", "and", "or"] k) ]
-  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False ((<> ".") <$> (M.keys operators ++ ["not", show Plain, show Phrase]))
-                    || hasLanguageFts val
+  hasOperator val = any (`T.isPrefixOf` val) $
+                      ((<> ".") <$> "not":M.keys operators) ++
+                      ((<> "(") <$> M.keys ftsOperators)
   isEmbedPath = T.isInfixOf "."
-  -- handle "?tsv=english.fts.possible" case
-  hasLanguageFts val = case T.splitOn "." val of
-    [_, "fts", _] -> True
-    [_, "fts"]    -> True
-    [_, "@@", _]  -> True -- TODO: '@@' deprecated
-    [_, "@@"]     -> True
-    _             -> False
   isTargetingProc = fromMaybe False $ (== "rpc") <$> listToMaybe path
   payload =
     case decodeContentType . fromMaybe "application/json" $ lookupHeader "content-type" of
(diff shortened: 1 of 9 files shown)


begriffs:
I like the new syntax too, it looks clean.

Because this is a breaking change it should go in postgrest 5. (The parameterized callProc pull request should as well because it changes the postgresql minimum version requirement.)

Maybe we should get the other PRs merged first and make a 4.4 release with the new features, then follow it closely with a 5.0 release that contains these two breaking changes?

steve-chavez:
I don't see the breaking change, the `plain/phrase` full text search hasn't been released yet and `fts` still maintains the v0.4.3.0 syntax(the lang in fts(lang) is optional). 

My idea was to avoid having a breaking change in v0.5 by having v0.4.4 with this PR merged and also after merging this to release v0.4.4 leaving my other PR's for v0.5 or V0.5-RC since they change core sql queries of PostgREST.

begriffs:
Good point. I forgot the full text search is not yet part of an actual release.

src/PostgREST/ApiRequest.hs
```diff
@@ -142,16 +142,8 @@ userApiRequest schema req reqBody
       ActionInvoke{isReadOnly=True} -> partition (liftM2 (||) (isEmbedPath . fst) (hasOperator . snd)) flts
       _ -> (flts, [])
   flts = [ (toS k, toS $ fromJust v) | (k,v) <- qParams, isJust v, k /= "select", not (endingIn ["order", "limit", "offset", "and", "or"] k) ]
-  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False ((<> ".") <$> (M.keys operators ++ ["not", show Plain, show Phrase]))
-                    || hasLanguageFts val
+  hasOperator val = foldr ((||) . flip T.isPrefixOf val) False $ ((<> ".") <$> "not":M.keys operators) ++ ((<> "(") <$> M.keys ftsOperators)
```
begriffs: This code is kind of gnarly, would the [any](https://hackage.haskell.org/package/base-4.10.0.0/docs/Prelude.html#v:any) function help simplify it?

begriffs: Like how about this:

```haskell
hasOperator val =
  any (`T.isPrefixOf` val)
    $  map (`snoc` '.') ("not" : M.keys operators)
    ++ map (`snoc` '(') (M.keys ftsOperators)
```

src/PostgREST/QueryBuilder.hs
```diff
@@ -433,12 +433,8 @@ pgFmtFilter table (Filter fld (OpExpr hasNot oper)) = notOp <> " " <> case oper
...
       Nothing    -> emptyValForIn
 
-   Fts mode lang val ->
-     pgFmtFieldOp "fts" <> " " <> case mode of
-       Normal -> "to_tsquery("
-       Plain  -> "plainto_tsquery("
-       Phrase -> "phraseto_tsquery("
-     <> maybe "" (flip (<>) ", " . pgFmtLit) lang <> unknownLiteral val <> ") "
+   Fts op lang val ->
+     pgFmtFieldOp op <> "(" <> maybe "" (flip (<>) ", " . pgFmtLit) lang <> unknownLiteral val <> ") "
```
begriffs: How about a little whitespace and a section rather than flip?

```haskell
pgFmtFieldOp op <> "("
  <> maybe "" ((<> ", ") . pgFmtLit) lang
  <> unknownLiteral val <> ") "
```

begriffs:
A couple minor style suggestions, then I'd say it's good to merge!

src/PostgREST/ApiRequest.hs
steve-chavez: The `any` function is helpful, though the `snoc/map` instead of `<>/<$>` makes the function harder to read for me.

PR #1021, opened 2017-11-21
allColumns query narrowed down to current postgrest schema

nextstopsun:
This gives a significant speedup at startup when you have a lot of tables in DB.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -552,7 +552,10 @@ allPrimaryKeys tabs =
  where
   sql = [q|
     /*
-    -- CTE to replace information_schema.table_constraints to remove owner limit
+    -- CTE based on information_schema.columns 
+    -- changed:
+    -- remove the owner filter
+    -- limit columns to the ones in the api schema or PK/FK columns
     */
     WITH tc AS (
         SELECT current_database()::information_schema.sql_identifier AS constraint_catalog,
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -364,7 +364,10 @@ allColumns tabs =
                AND n.nspname NOT IN ('pg_catalog', 'information_schema', $1)
         ),
         /*
-        -- CTE based on information_schema.columns to remove the owner filter
+        -- CTE based on information_schema.columns 
+        -- changed:
+        -- remove the owner filter
+        -- limit columns to the ones in the api schema or PK/FK columns
         */
         columns AS (
             SELECT current_database()::information_schema.sql_identifier AS table_catalog,
@@ -552,10 +555,7 @@ allPrimaryKeys tabs =
  where
   sql = [q|
     /*
-    -- CTE based on information_schema.columns 
-    -- changed:
-    -- remove the owner filter
-    -- limit columns to the ones in the api schema or PK/FK columns
+    -- CTE to replace information_schema.table_constraints to remove owner limit
     */
     WITH tc AS (
         SELECT current_database()::information_schema.sql_identifier AS constraint_catalog,
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -326,25 +326,24 @@ allColumns tabs =
   H.statement sql (HE.value HE.text) (decodeColumns tabs) True
  where
   sql = [q|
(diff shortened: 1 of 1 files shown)


begriffs:
@nextstopsun you rock for jumping into the code, thank you! 🤘 

I saw a conversation in the chat about changing the query, but can you explain the changes? What is the `typelem`, `typlen` stuff etc?

begriffs:
Oh, actually the way github was showing the diff was confusing to me. Sometimes when indentation changes across many lines it makes the diff big. Adding `?w=1` to the "Files Changed" url in github shows the changes more clearly.

So it looks like you're restricting a join to go on just key columns? Is that the main idea for the optimization?

ruslantalpa:
`AND n.nspname NOT IN ('pg_catalog', 'information_schema', $1)` this is wrong i think, `$1` should not be there

nextstopsun:
@ruslantalpa Why do you think it's wrong? All columns from `api` schema get selected in the next query, so could be excluded from pk/fk list too.

nextstopsun:
@begriffs Basically I've added a cte to fetch all pk and fk columns and restricted main query to just columns in api schema. 
In my case this gives ~1000 times startup speedup.
Sorry for changed indentation, it was autoindented while I was testing parts of the query separately.

ruslantalpa:
It was at first look, i'll look again

ruslantalpa:
a few things to implement/cleanup

- reformat sql query as it was and add the new stuff after (?w=1 works well but i think it's cleaner to have it as before + it's not aligned right now)
- remove the `$1` filter from the key_columns. While it works correctly as is, the naming of the CTE as is, is misleading, it's actually "key_columns_that_are_not_in_the_api_schema" and you only understand the logic behind the filter if you look at the `or` condition in the next CTE. so let's remove the `$1` to make `key_columns` more easy to follow.
- move the top comment
```
       -- CTE based on information_schema.columns to remove the owner filter
```
above the columns CTE and rewrite like this
```
       -- CTE based on information_schema.columns 
       -- changed:
       -- remove the owner filter
       -- limit columns to the ones in the api schema or PK/FK columns
```
- add a comment for this line `AND (nc.nspname = $1 OR kc.r_oid IS NOT NULL)` like so
 `/*--filter only columns that are FK/PK or in the api schema  */`

nextstopsun:
Ok, will do. 
But `remove the $1 filter from the key_columns` makes it a tiny bit slower. 
Also we'll have duplicated columns then. Although they are cleared later in the query.
Is this necessary just for naming purposes?

ruslantalpa:
formatting perfect.
the `$1` is not for naming strictly, it's make the query understandable, it's already complicated as it is, i am sure it is a little slower but probably by a very slow margin but it will be easier to understand for someone looking at it.

nextstopsun:
Well honestly all these massive queries need some love (with nested regex selects etc.) Better split them to understandable CTEs where possible. 

Meanwhile, can we just stick with this comment?
`CTE based on pg_catalog to get only Primary and Foreign key columns outside api schema`

I'll remove `$1` if it looks critical, but the whole query is already complicated enough so imho this tiny bit doesn't really matter.

ruslantalpa:
comment helps, i guess it's ok as is.
yes, those queries need some cleanup, they evolved slowly

src/PostgREST/DbStructure.hs
```diff
@@ -523,7 +552,10 @@ allPrimaryKeys tabs =
  where
   sql = [q|
     /*
-    -- CTE to replace information_schema.table_constraints to remove owner limit
```
ruslantalpa: you replaced the wrong comment, it's a completely different query then the one you were working on

nextstopsun:
Oops. @ruslantalpa you're right!

ruslantalpa:
good to merge 👍

nextstopsun:
@begriffs ping

begriffs:
Thanks for the nudge! I'll give this one last look this evening and get it merged.

begriffs:
@nextstopsun this looks good, and runs fine for me locally as well.

One more thing before I merge. Could you add an entry to [CHANGELOG.md](https://github.com/begriffs/postgrest/blob/master/CHANGELOG.md) noting how you improved startup performance?

begriffs:
Nevermind, I'll just add that for you.

PR #1022, opened 2017-11-26
Include git sha in version

begriffs:
When people submit a bug report for a custom binary they built off master it would be helpful to know exactly which git hash they built from. This helps us debug problems better.

This pull request adds the first seven characters of the git sha to the version output in the Server response header and also the usage screen.

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE LambdaCase, TemplateHaskell #-}
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
 {-|
 Module      : PostgREST.Config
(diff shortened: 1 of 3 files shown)


src/PostgREST/Config.hs
```diff
@@ -42,15 +42,16 @@ import           Data.Scientific              (floatingOrInteger)
 import           Data.String                  (String)
 import           Data.Text                    (dropAround,
                                                intercalate, lines,
-                                               strip)
+                                               strip, take)
 import           Data.Text.Encoding           (encodeUtf8)
 import           Data.Text.IO                 (hPutStrLn)
 import           Data.Version                 (versionBranch)
+import           Development.GitRev
```
ruslantalpa: maybe an explicit import here of `gitHash`

begriffs: Good idea

begriffs:
I also confirmed that the `externalDocs` property has the current version in its url. The `docsVersion` command properly strips out the extra sha. So I think it's good for me to merge.

PR #1024, opened 2017-11-27
Fix #1016

### LINKED ISSUE #1016: Things broken when jwt aud is specified
This line is wrong
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Auth.hs#L42
when aud is specified and a request without jwt comes (anonymous), things break.
Also config should treat empty string for aud as Nothing (for docker.)

I have the code for this, i'll make a pr, just documenting so that we don't forget and do a release

ruslantalpa:


src/PostgREST/Auth.hs
```diff
@@ -39,7 +39,7 @@ data JWTAttempt = JWTInvalid JWTError
   of JWT claims.
 -}
 jwtClaims :: Maybe JWK -> Maybe StringOrURI -> LByteString  -> IO JWTAttempt
-jwtClaims _ Nothing "" = return $ JWTClaims M.empty
+jwtClaims _ _ "" = return $ JWTClaims M.empty
```
begriffs: Given that you now turn `Just ""` into `Nothing` in the config parser, does this pattern match need to be updated too?

ruslantalpa: it would probably work without this change but on it's own `jwtClaims` does not seem right the way it was. The `M.empty` result is determined by the presence (or absence) of the JWT, the `aud` config value should not influence that.

PierreR: If I am not mistaken this is actually changing a very old behavior, isn't ?

ruslantalpa:
You are mistaken :)
Aud is a recent thing, probably not even released and if by bihavior you mean error when configuring with aud and sending an anonymous reques, then yes, this is changing that ;)

Try the new test on the master branch and see it fail

PierreR:
@ruslantalpa I am wondering if you are replying to my deleted comment. Anyhow I believe you are quite right ;-) that is why I have deleted the comment afterward as it didn't bring anything useful.

begriffs:
This PR makes sense. When there is no JWT payload then `jwtClaims` should return an empty claims map regardless of the aud configuration. Looks good to merge!

src/PostgREST/Config.hs
```diff
@@ -148,6 +148,7 @@ readOptions = do
         Nothing -> pure Nothing -- no audience in config file
         Just aud -> case preview stringOrUri (aud :: String) of
           Nothing -> fail "Invalid Jwt audience. Check your configuration."
+          (Just "") -> pure Nothing
```
steve-chavez: Maybe remove the unnecessary parentheses around `Just ""` here, wonder why hlint doesn't catch this.

PR #1025, opened 2017-11-27
Faster queries using json_agg

ruslantalpa:
with these changes a single postgrest joined query with json encoding over 3 tables is as fast as making 3 distinct queries (with `id in (...)` condition) on one table each.

tests changed because json_agg returns a slightly different json format, but the same data.


begriffs:
Nice, I know you've been wanting to fix this one for a while.

My only note is that `json_agg` is broken on Citus, but then again so are other things postgrest uses, so we won't really be much worse off for future compatibility.

Also, when you say this returns a slightly different json format, do you mean it simply re-orders the rows?

ruslantalpa:
i mean new lines between array items

begriffs:
OK, no big deal then. Guessing that testing `[json|…|]` for equality disregards whitespace differences and hence the reason for that change in the tests.

ruslantalpa:
yes, exactly

begriffs:
This looks good to merge!

PR #1027, opened 2017-11-29
Make full text search be explicit with config.

luke-clifton:
`to_tsvector(<lang>,<lhs>)` is now called explicitly before being passed to `@@` if a language configuration has been supplied in the query.

This way the `tsquery` and the `tsvector` match. Not doing this causes weird results to be returned.

If people are happy with this, I can add tests.

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -434,7 +434,9 @@ pgFmtFilter table (Filter fld (OpExpr hasNot oper)) = notOp <> " " <> case oper 
       Nothing    -> emptyValForIn
 
    Fts op lang val ->
-     pgFmtFieldOp op
+       maybe (\x -> x) (\l fmtfld -> "to_tsvector(" <> pgFmtLit l <> ", " <> fmtfld <> ")") lang (pgFmtField table fld)
+       <> " "
+       <> sqlOperator op
        <> "("
        <> maybe "" ((<> ", ") . pgFmtLit) lang
        <> unknownLiteral val


steve-chavez:
PostgREST uses the `@@` to operate on a `tsvector` column type, it seems to me that you are using it to operate on `text` instead.

Could you clarify your use case? I think properly formatting the data on INSERT(having a `tsvector` on the table instead of `text` and then doing `INSERT INTO table VALUES (to_tsvector('english', 'doc..')`) would be more appropriate.

luke-clifton:
Indeed we want to specify a text column, and we create indices on that text
column with a variety of configurations (multiple languages).

https://www.postgresql.org/docs/9.1/static/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX

With this PR we know that the config used to create the tsquery always
matches that used to create the tsvector.

If you are operating on tsvector columns directly, surely it would be best
for postgrest to choose the matching configuration for you based on how the
tsvector was created.

If they don't match you get weird/wrong results.

steve-chavez:
I see, however as it is the change could not be merged since it would break existing clients(that use `fts` to operate on `tsvector`), you can see the `function to_tsvector(unknown, tsvector) does not exist` error in CircleCI, and that's because `to_tsvector` can only operate on `text`.

Not sure for now if it's worth it to accommodate your use case in PostgREST, I think operating on `tsvector` is the most common use case and you could accomplish what you want with an RPC(which now supports GET).

If we would include your change in PostgREST maybe an option could be detecting the type of the column, e.g. in `?doc=fts(english).fatcat` see if `doc` is `text` or `tsvector` and if it's `text` use the `to_tsvector` function, that would make the change backwards compatible.

luke-clifton:
Hmm, it wasn't my intention to change the behaviour of using an undecorated `fts` call. It should just fall back to what it was doing before, and only attempt to do anything different if the `lang` argument is `Just something`.

luke-clifton:
I guess, the way I see it:

The current behaviour (in master) means that there is only ever one valid way to use `fts`, that is, it MUST be used with the same dictionary used to create the `tsvector` (or, if operating on `text` must be used undecorated so that the default dictionary is used) otherwise we get incorrect results. Worse, the client doesn't know what dictionary that is (and can't know except via external documentation I believe?). Worse again, if you change the dictionary used in the database, you need to update all your clients to start using the same dictionary.

My proposed behaviour allows for more flexibility. We assume you are operating on text if you supply a dictionary, and thus allow you to choose the desired dictionary from the client without risk of a mismatch. If you don't supply a dictionary, the default is used, and if you have stored `tsvector` and are operating on that, we assume that that was created using the default dictionary (we can't know any better?).

If you want the existing (in master) behaviour of operating on `tsvector` which was created with a custom dictionary directly, I would suggest you use the RPC mechanism, because otherwise you run the risk of your clients using an incorrect dictionary, which means their results will be garbage.

begriffs:
> We assume you are operating on text if you supply a dictionary … If you don't supply a dictionary, the default is used, and if you have stored tsvector and are operating on that, we assume that that was created using the default dictionary

I think the conditional logic for changing the expected column type may be confusing for people.

Steve's suggestion seems easier to work with because it reacts by detecting the underlying column type rather than the syntax of the url.

Although one other question: would using a computed tsvector column work for your situation?

```sql
create table searchme ( corpus text primary key );

create function corpus_fts(searchme) returns tsvector as $$
  select to_tsvector('english', $1.corpus);
$$ language sql;

create index corpus_fts_idx on searchme using gin(to_tsvector('english', corpus));

insert into searchme values ('search this'), ('also this'), ('another one');
```

Now you can request `/searchme?corpus_fts=fts(english).another`

luke-clifton:
> it reacts by detecting the underlying column type rather than the syntax of the url.

What do you mean by this? It doesn't detect the underlying column type, unless I'm misunderstanding?

> I think the conditional logic for changing the expected column type may be confusing for people.

If you provide a dictionary we use that, if you don't, it behaves like `@@`  from PostgreSQL. My experience watching people using the current fts is that they where very confused by the fact that using a dictionary when trying to search text columns got them back subtly wrong results (a sample size of two people is not much, but they both expected something different to the current behaviour).

Maybe I can change the PR to detect text columns and do the right thing in that case? Would that be possible?

begriffs:
@luke-clifton OK, I'm reading over your initial description again and noticing things I overlooked initially. First you have a single text column, and then have multiple indices on `to_tsvector` expressions for multiple languages. So having the `fts` query filter wrap the text column with `to_tsvector` is convenient because you can switch languages in the query and automatically match one of those indices.

I also understand your point that tsvectors don't expose their language setting to API consumers so there is no clue which language an API request should use in its `fts` filter. 

All this makes sense. But here's what's weird to me: `@@` operates on tsvector, that's how it was designed and what it does. Being able to coerce a text column for `@@` would be totally cool, except the change in this PR of unconditionally wrapping the column with `to_tsvector` *breaks* the use of `@@` on a tsvector, which is its fundamental use.

What about this: expose a special version of `fts` for text columns (`tfts`?) which uses to_tsvector, and leave `fts` to operate on columns which are already tsvectors? Then we could update our docs to discuss the difference between these operators and the scenarios for their use.

luke-clifton:
As long as you don't mind another operator, I'm happy to do that.

begriffs:
Cool, let's do a separate operator. Want to update the PR with that?

luke-clifton:
Will do once I'm back online in the new year!

begriffs:
@luke-clifton hope you can find some time soon to get this one finished. I've been reviewing the open pull requests trying to get them all resolved.

begriffs:
Looks like @luke-clifton lost interest in this, so closing.

Issue #781, opened 2016-12-27
No errors when desc/asc ordering modificators are misspelled


NikolayS:
By mistake, I made a typo and wrote `...?order=mycolumn.dec` in API call – so basically I wrote `dec` instead of `desc`.

PostgREST accepted it, producing no errors... It did ordering with `asc`. This time I was lucky enough to quickly find that typo. But it potentially would be very hard to catch.

I expect that it should report, that only either `asc` or `desc` are accepted in that place.

begriffs:
Makes sense, could be annoying to track that sort of thing down otherwise.

tedkornish:
I'd like to take a stab at this one. The request should probably respond with a `400 Bad Request` and maybe a plaintext payload like:

```
invalid ordering "dec" on column "mycolumn"
```

begriffs:
Thank you, I'd appreciate your help with this one! If you run into problems feel free to open a pull request with what you've got and we can discuss it from there.

As for the error message, we usually try to return a json object with fields:

* message (required)
* details
* hint
* code

You could put the string you suggested into message, and maybe mention the correct possible values in the hint.

steve-chavez:
Fixed in #1104.

PR #1047, opened 2018-01-20
fix parser to show errors when ordering modificators are misspelled

onporat:
Fix for issue #781
The solution:
1. Parse OrderDirection OrderNulls together (not one after the other) so we could tell between misspelling and change of type.
2. use setBy1 so errors are not swallowed

That was a tricky one, thanks for the challenge.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -10,10 +10,10 @@ stgrest:
     PGRST_DB_ANON_ROLE: app_user
 
 postgres:
-  image: postgres
+  image: postgres:9
   ports:
     - "5432:5432"
   environment:
-    POSTGRES_DB: app_db
-    POSTGRES_USER: app_user
+    POSTGRES_DB: postgrest_test
+    POSTGRES_USER: postgrest_test
     POSTGRES_PASSWORD: password
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -115,6 +115,10 @@ data OrderTerm = OrderTerm {
 , otDirection :: Maybe OrderDirection
 , otNullOrder :: Maybe OrderNulls
 } deriving (Show, Eq)
+instance Monoid OrderTerm where
+  mempty = OrderTerm (mempty, mempty) Nothing Nothing
+  OrderTerm nt (Just nd) Nothing `mappend` OrderTerm mt Nothing (Just mn) = OrderTerm (nt `mappend` mt) (Just nd) (Just mn)
+  OrderTerm nt Nothing (Just nn) `mappend` OrderTerm mt (Just md) Nothing = OrderTerm (nt `mappend` mt) (Just md) (Just nn)
 
 data QualifiedIdentifier = QualifiedIdentifier {
   qiSchema :: Schema
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -117,6 +117,9 @@ data OrderTerm = OrderTerm {
 } deriving (Show, Eq)
 instance Monoid OrderTerm where
   mempty = OrderTerm (mempty, mempty) Nothing Nothing
+  OrderTerm ("",Nothing) Nothing nn  `mappend` OrderTerm tm Nothing Nothing = OrderTerm tm Nothing nn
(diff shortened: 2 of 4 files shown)


src/PostgREST/Types.hs
```diff
@@ -115,6 +115,10 @@ data OrderTerm = OrderTerm {
 , otDirection :: Maybe OrderDirection
 , otNullOrder :: Maybe OrderNulls
 } deriving (Show, Eq)
+instance Monoid OrderTerm where
+  mempty = OrderTerm (mempty, mempty) Nothing Nothing
+  OrderTerm nt (Just nd) Nothing `mappend` OrderTerm mt Nothing (Just mn) = OrderTerm (nt `mappend` mt) (Just nd) (Just mn)
```
begriffs: The continuous integration tests are failing at this line with a warning:

```
Pattern match(es) are non-exhaustive
In an equation for ‘mappend’:
    Patterns not matched:
        (OrderTerm _ Nothing Nothing) _
        (OrderTerm _ Nothing (Just _)) (OrderTerm _ Nothing _)
        (OrderTerm _ Nothing (Just _)) (OrderTerm _ (Just _) (Just _))
        (OrderTerm _ (Just _) Nothing) (OrderTerm _ Nothing Nothing)
        ...
```

begriffs:
Thanks for working on this. Could you include some test cases to illustrate the fix as well?

begriffs:
Looks like a few other tests are failing now if you want to take a look.

begriffs:
I tried this branch out locally, and it does produce a nice parse error:

```
/foo?order=a.bla

{
  "details": "unexpected \"b\" expecting \"asc\", \"desc\", \"nullslast\" or \"nullsfirst\"",
  "message": "\"failed to parse order (a.bla)\" (line 1, column 3)"
}
```

However the parser still allows trailing characters. For instance, this is accepted: `/foo?order=a.desc2`

docker/docker-compose.yml
```diff
@@ -10,10 +10,10 @@ stgrest:
     PGRST_DB_ANON_ROLE: app_user
 
 postgres:
-  image: postgres
+  image: postgres:9
```
begriffs: What was the motivation for changing this file?

src/PostgREST/Types.hs
```diff
@@ -115,6 +115,14 @@ data OrderTerm = OrderTerm {
 , otDirection :: Maybe OrderDirection
 , otNullOrder :: Maybe OrderNulls
 } deriving (Show, Eq)
+instance Monoid OrderTerm where
+  mempty = OrderTerm (mempty, mempty) Nothing Nothing
+  OrderTerm ("",Nothing) Nothing nn  `mappend` OrderTerm tm Nothing Nothing = OrderTerm tm Nothing nn
```
begriffs: All the cases in this `mappend` definition are confusing to me, and also I don't think they work right.

For instance, calling `/foo?order=a.asc.desc` blows up in a weird way:

```json
{
  "hint":null,
  "details":null,
  "code":"42601",
  "message":"zero-length delimited identifier at or near \"\"\"\""
}
```

What is the intended behavior? Perhaps something like

```haskell
mappend (OrderTerm at ad ano) (OrderTerm bt bd bno) =
  OrderTerm (at <> bt) (bd <|> ad) (bno <|> ano)
```
Meaning combine the terms, and choose whichever non-Nothing direction and order are last encountered.

begriffs:
I noticed that this branch breaks multi-column sorting, `/foo?order=a.desc,b.asc`:

```json
{
  "details":"unexpected ','",
  "message":"\"failed to parse order (a.desc,b.asc)\" (line 1, column 8)"
}
```

Our docs [describe](https://postgrest.com/en/v4.4/api.html#ordering) this use case but curiously our test suite doesn't include a test for it. Could you fix the behavior on this branch, and add a test for it as well?

begriffs:
@onporat are you still motivated to work on this pull request? I like your improvement and think we're getting close to completing it.

steve-chavez:
I would like to merge this before next release, I'll continue the work in another PR.

steve-chavez:
Closing as #1104 got merged, thanks for the work @onporat.

PR #1048, opened 2018-01-21
Add support for UPSERT

### LINKED ISSUE #256: Bulk upsert
Hi,

Is there a way to `insert` **or** `update` (like `pg 9.5` `upsert`) some rows.

I have a large `CSV` file with some `duplicate` content.

Is there a possibility to `merge` data into a table ?

My idea (if not already `developed`) is to :
- Create a _temporary_ table (same structure than my _target_)
- Fill my _temporary_ table
- `Merge` or `upsert` (for `pg 9.5`) each rows
- `Merge` content from _temporary_ to _target_
- `Delete` _temporary_

My goal is to :
- Avoid duplicates
- Avoid table locks
- Be ready for high availabity

Regards,

steve-chavez:
Closes #256. UPSERT is done with the `ON CONFLICT` clause available on PostgreSQL >= 9.5. 

Bulk UPSERT can be done by adding `Prefer: resolution=ignore-duplicates/merge-duplicates` headers, these do a `INSERT .. ON CONFLICT(<pkCols>) DO NOTHING/DO UPDATE`. 
The `ignore-duplicates/merge-duplicates` Prefer values only work with tables that have a PK, if the table doesn't have a PK the `Prefer` will be ignored(no error).

Single row UPSERT can be done with PUT if the client specifies only PK columns as `eq` filters and the payload is identified by the URI, other restrictions include not allowing `Content-Range/limit/offset`(took into account the previous [implementation](https://github.com/begriffs/postgrest/blob/a34051b1774eb4b589354a469024d4ce2482bc83/src/Dbapi.hs#L137-L158)).

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE NamedFieldPuns #-}
 
 module PostgREST.App (
   postgrest
@@ -79,7 +80,7 @@ postgrest conf refDbStructure pool worker =
 
             let authed = containsRole eClaims
                 proc = case (iTarget apiRequest, iPayload apiRequest, iPreferSingleObjectParameter apiRequest) of
-                  (TargetProc qi, Just PayloadJSON{pjKeys=pKeys}, s) -> findProc qi pKeys s $ dbProcs dbStructure
+                  (TargetProc qi, Just PayloadJSON{pjKeys}, s) -> findProc qi pjKeys s $ dbProcs dbStructure
                   _ -> Nothing
                 handleReq = runWithClaims conf eClaims (app dbStructure proc conf) apiRequest
                 txMode = transactionMode proc (iAction apiRequest)
@@ -144,11 +145,11 @@ app dbStructure proc conf apiRequest =
                       )
                     ] (toS body)
 
-        (ActionCreate, TargetIdent (QualifiedIdentifier _ table), Just (PayloadJSON payload pType _)) ->
-          case mutateSqlParts of
+        (ActionCreate, TargetIdent (QualifiedIdentifier tSchema tName), Just PayloadJSON{pjRaw, pjType}) ->
+          case mutateSqlParts tSchema tName of
             Left errorResponse -> return errorResponse
             Right (sq, mq) -> do
-              let (isSingle, nRows) = case pType of
+              let (isSingle, nRows) = case pjType of
                                         PJArray len -> (len == 1, len)
                                         PJObject -> (True, 1)
               if contentType == CTSingularJSON
@@ -156,17 +157,16 @@ app dbStructure proc conf apiRequest =
                  && iPreferRepresentation apiRequest == Full
                 then return $ singularityError (toInteger nRows)
                 else do
-                  let pKeys = map pkName $ filter (filterPk schema table) allPrKeys -- would it be ok to move primary key detection in the query itself?
-                      stm = createWriteStatement sq mq
+                  let stm = createWriteStatement sq mq
                         (contentType == CTSingularJSON) isSingle
                         (contentType == CTTextCSV) (iPreferRepresentation apiRequest)
-                        pKeys
-                  row <- H.query (toS payload) stm
+                        (tablePKCols dbStructure tSchema tName)
+                  row <- H.query (toS pjRaw) stm
                   let (_, _, fs, body) = extractQueryResult row
                       headers = catMaybes [
                           if null fs
                             then Nothing
-                            else Just (hLocation, "/" <> toS table <> renderLocationFields fs)
+                            else Just (hLocation, "/" <> toS tName <> renderLocationFields fs)
                         , if iPreferRepresentation apiRequest == Full
                             then Just $ toHeader contentType
                             else Nothing
@@ -178,8 +178,8 @@ app dbStructure proc conf apiRequest =
                     if iPreferRepresentation apiRequest == Full
                       then toS body else ""
 
-        (ActionUpdate, TargetIdent _, Just p@(PayloadJSON payload _ _)) ->
-          case (mutateSqlParts, pjIsEmpty p, iPreferRepresentation apiRequest == Full) of
+        (ActionUpdate, TargetIdent (QualifiedIdentifier tSchema tName), Just p@PayloadJSON{pjRaw}) ->
+          case (mutateSqlParts tSchema tName, pjIsEmpty p, iPreferRepresentation apiRequest == Full) of
             (Left errorResponse, _, _) -> return errorResponse
             (_, True, True) -> return $ responseLBS status200 [contentRangeH 1 0 Nothing] "[]"
             (_, True, False) -> return $ responseLBS status204 [contentRangeH 1 0 Nothing] ""
@@ -187,7 +187,7 @@ app dbStructure proc conf apiRequest =
               let stm = createWriteStatement sq mq
                     (contentType == CTSingularJSON) False (contentType == CTTextCSV)
                     (iPreferRepresentation apiRequest) []
-              row <- H.query (toS payload) stm
+              row <- H.query (toS pjRaw) stm
               let (_, queryTotal, _, body) = extractQueryResult row
               if contentType == CTSingularJSON
                  && queryTotal /= 1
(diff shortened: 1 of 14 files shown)


begriffs:
:clinking_glasses: Wow, this is a great feature. People have been wanting this one for a long long time, thank you!

I'll review the code and add notes.

src/PostgREST/QueryBuilder.hs
```diff
@@ -275,23 +275,30 @@ requestToQuery schema isParent (DbRead (Node (Select colSelects tbls logicForest
     --getQueryParts is not total but requestToQuery is called only after addJoinConditions which ensures the only
     --posible relations are Child Parent Many
     getQueryParts _ _ = undefined
-requestToQuery schema _ (DbMutate (Insert mainTbl p@(PayloadJSON _ pType keys) returnings)) =
+requestToQuery schema _ (DbMutate (Insert mainTbl pkCols p@(PayloadJSON _ pType pKeys) onConflct logicForest returnings)) =
```
begriffs: I was confused in another part of the code because `pKeys` made me think "primary keys" rather than "payload keys." Maybe name it `pjKeys`?

src/PostgREST/App.hs
```diff
@@ -205,8 +204,36 @@ app dbStructure proc conf apiRequest =
                     then responseLBS s [toHeader contentType, r] (toS body)
                     else responseLBS s [r] ""
 
-        (ActionDelete, TargetIdent _, Nothing) ->
-          case mutateSqlParts of
+        (ActionSingleUpsert, TargetIdent (QualifiedIdentifier tSchema tName), Just (PayloadJSON payload pType pKeys)) ->
```
begriffs: I was confused below because `pKeys` made me think "primary keys" rather than "payload keys." Maybe name it `pjKeys`?

src/PostgREST/App.hs
```diff
@@ -205,8 +204,36 @@ app dbStructure proc conf apiRequest =
...
+              else if not isSingle
+                then return $ simpleError status405 [] "PUT payload must contain a single row"
+              else if S.fromList colNames /= pKeys
+                then return $ simpleError status405 [] "You must specify all columns in the payload when using PUT"
+              else do
+                row <- H.query (toS payload) $
+                       createWriteStatement sq mq (contentType == CTSingularJSON) False
+                                            (contentType == CTTextCSV) (iPreferRepresentation apiRequest) []
+                let (_, queryTotal, _, body) = extractQueryResult row
+                if queryTotal /= 1
```
begriffs: Nice, `queryTotal /= 1` is a clever way to detect that the PUT is being applied incorrectly, rather than comparing the filters in the request to primary keys etc etc.

src/PostgREST/App.hs
```diff
@@ -205,8 +204,36 @@ app dbStructure proc conf apiRequest =
...
+                  then do
+                    HT.condemn
+                    return $ simpleError status405 [] "The URI must univocally identify the payload when using PUT"
```
begriffs: This error message might be unclear. What about, "PUT must be applied to exactly one resource." Maybe include a hint like, "Request attempted to modify ${queryTotal} table rows"

src/PostgREST/App.hs
begriffs: Looking further into the code, I see that `mutateRequest` checks for `ActionSingleUpsert` that the request contains equality filters on all primary keys. Does that make the `queryTotal /= 1` here unnecessary, or what is the interaction between the logic?

test/Feature/UpsertSpec.hs
```diff
@@ -0,0 +1,183 @@
...
+        request methodPost "/only_pk" [("Prefer", "return=representation"), ("Prefer", "resolution=ignore-duplicates")]
+          [json|[ { "id": 1 }, { "id": 2 }, { "id": 3} ]|]
+          `shouldRespondWith`
+          [json|[ { "id": 3} ]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
+        request methodPost "/only_pk" [("Prefer", "return=representation"), ("Prefer", "resolution=merge-duplicates")]
+          [json|[ { "id": 1 }, { "id": 2 }, { "id": 4} ]|]
+          `shouldRespondWith`
+          [json|[ { "id": 1 }, { "id": 2 }, { "id": 4} ]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
+
+      it "succeeds and ignores the Prefer: resolution header if the table has no PK" $
```
begriffs: What about this alternative: if a table has no pk then make the Prefer header `resolution=ignore-duplicates` translate to `ON CONFLICT DO NOTHING` (which has an empty conflict target). But have `resolution=merge-duplicates` raise an error on tables without a pk because there's no obvious conflict target other than the primary keys. Currently it appears that the entire `ON CONFLICT` clause is dropped for these tables.

I feel like if a client explicitly requests merge-duplicates then it's an error to ignore it on tables without a pk. Is there a use-case you have in mind that I am overlooking?

begriffs:
This is really pleasant and readable code. I just had a few small comments, and a question about the proper behavior on tables lacking a primary key.

test/Feature/UpsertSpec.hs
steve-chavez: I chose to drop the `ON CONFLICT` clause for table with no pks because I thought that `Prefer` semantics meant the condition it's optional and shouldn't give an error, also according to [RFC 7240](https://tools.ietf.org/html/rfc7240):
> A server that does not recognize or is unable to comply with
   particular preference tokens in the Prefer header field of a request
   MUST ignore those tokens and continue processing instead of signaling
   an error.

I thought to later add the `Preference-Applied` feature so the client can check the success of the condition.

src/PostgREST/App.hs
steve-chavez: Is not clear from the code but I'm enforcing a different condition here, so first of all my interpretation of [RFC 7231](https://tools.ietf.org/html/rfc7231#section-4.3.4) is that the PUT uri must identify the payload being inserted so that the next GET can return the same exact payload right? So what I'm doing here is this:

```haskell
-- table numbers is empty
it "fails if the uri doesn't identify the payload" $
  put "/numbers?id=eq.7" [str| [ { "id": 11, "name": "eleven"} ]|] `shouldRespondWith` 405
```

This is actually done by the [sql query](https://github.com/steve-chavez/postgrest/blob/upsert/src/PostgREST/QueryBuilder.hs#L290) that WHEREs the payload according to the querystring parameters, if that WHERE gives false then nothing gets upserted and the `queryTotal` will be 0.

src/PostgREST/App.hs
steve-chavez: Is not clear from the code but I'm enforcing a different condition here, so first of all my interpretation of [RFC 7231](https://tools.ietf.org/html/rfc7231#section-4.3.4) is that the PUT uri must identify the payload being inserted so that the next GET can return the same exact payload right? So what I'm doing here is this:

```haskell
-- table numbers is empty
it "fails if the uri doesn't identify the payload" $
  put "/numbers?id=eq.7" [str| [ { "id": 11, "name": "eleven"} ]|] `shouldRespondWith` 405
```

This is actually done by the [sql query](https://github.com/steve-chavez/postgrest/blob/upsert/src/PostgREST/QueryBuilder.hs#L290) that WHEREs the payload according to the querystring parameters, if that WHERE gives false then nothing gets upserted and the `queryTotal` will be 0.

steve-chavez: I tried to not have the restriction of just allowing primary key cols `eq` filters first, but without that there's no reliable way of identifying the payload with the same sql query,  consider:
```http
-- id is persons pk, there's already many records with the name John
PUT /persons?name=eq.John
{ "id": 12, "first_name": "John", "last_name": "Doe" }
```
The WHERE would give true and the PUT would succeed with a `queryTotal == 1` but since name is not the pk subsequent `GET /persons` would have many records.

steve-chavez: @begriffs Let me know if I'm interpreting the RFC right and if so how can this logic be made more simple/clear.

test/Feature/UpsertSpec.hs
begriffs: Great point that this is only a Prefer header so we shouldn't raise an error if the header can't be applied.

About tables with no primary keys, we could conceivably do `ON CONFLICT` with no columns in parentheses. The docs say:

> If no list of column names is given at all, the default is all the columns of the table in their declared order

Only problem is this might cause an expensive sequential scan for duplicates rather than using indices, so it would put extra load on the server. Your choice to drop the ON CONFLICT for tables without keys is probably the right thing to do.

Would it be difficult to include a `Preference-Applied` response header for resolution? Might help people debug a surprising result.

src/PostgREST/App.hs
steve-chavez: Check my comments in https://github.com/begriffs/postgrest/pull/1048#discussion_r163925372, if I'm misunderstanding something in the RFC I'll change it but for now not sure what would be the clearer message.

begriffs:
I tried this out locally again and it's fun to use. 🎈 

Noticed a few things:

* `"You must specify all columns in the payload when using PUT"`
  * This returns HTTP 405, which is misleading because PUT *is* supported on the resource, it's just that the payload was wrong. Seems more like it should be a 400 error. (405 is perfect for the "Filters must include all and only primary key columns with 'eq' operator" though)
* This was weird, but when I did a POST with a single item that duplicated a row in the table, and used `Prefer: resolution=ignore-duplicates`, then the server had an internal error.
  * ```json
    {
      "details": "Row number 0",
      "message": "Row error: unexpected null"
    }
    ```
  * Seems like a haskell error because running the generated query in psql just gives
    ```
    ┌──────────────────┬────────────┬───────────┬──────────┐
    │ total_result_set │ page_total │ array_agg │ ?column? │
    ├──────────────────┼────────────┼───────────┼──────────┤
    │                  │          0 │ ∅         │          │
    └──────────────────┴────────────┴───────────┴──────────┘
    ```
    Including multiple items in the POST array makes the error go away, even when both items are dups of existing rows.

src/PostgREST/App.hs
begriffs: I'm haven't figured out how to make a request that causes this error message to happen. What do I need to do?

steve-chavez: This test shows how to make the error happen: https://github.com/begriffs/postgrest/pull/1048/files#diff-385e57e2f6697b6568f08e4d151b1d7cR133

begriffs: Thanks, now I see. What about adding a hint to the error message, like

```json
{
  "message": "The URI must univocally identify the payload when using PUT",
  "hint": "Changing primary key values with PUT is not allowed"
}
```

The existing message is fine, but the hint could make it more concrete for people.

begriffs: Or `"hint": "Payload values do not match resource in primary key column(s)"`
whatever you think explains it best.

steve-chavez:
@begriffs Added latest corrections.

test/Feature/UpsertSpec.hs
```diff
@@ -92,6 +92,14 @@ spec =
           `shouldRespondWith`
           [json|[ { "a": "1", "b": "0" } ]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
 
+      it "succeeds if not a single resource is created" $ do
+        request methodPost "/tiobe_pls" [("Prefer", "return=representation"), ("Prefer", "resolution=ignore-duplicates")]
+          [json|[ { "name": "Java", "rank": 1 } ]|] `shouldRespondWith`
+          [json|[]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
```
begriffs: If no records got created, then a [201](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.2) response doesn't seem accurate. Maybe a 200 in this case?

If it's a pain to change then it maybe it's not such a big deal. What do you think?

begriffs:
I added one more note about response codes. Aside from that and rebasing it looks good to merge!

test/Feature/UpsertSpec.hs
steve-chavez: Actually tried that before by applying a `queryTotal == 0` condition but other unrelated tests failed.

Like this [one](https://github.com/begriffs/postgrest/blob/master/test/Feature/InsertSpec.hs#L137), the issue is the `return=minimal` always makes the `queryTotal == 0` despite rows being inserted.

There's another way of detecting the insertion by using the [xmax](https://stackoverflow.com/a/39204667/4692662) system column, tried to use that for PUT as well (since it should respond 200 for update and 201 for insert according to RFC) but it involves a not-so-easy refactoring.

How about if I merge as it is and open an issue for that to correct it later?

steve-chavez: Also some old insert tests like this one https://github.com/begriffs/postgrest/blob/master/test/Feature/InsertSpec.hs#L129-L135 also respond with 201 despite not inserting nothing.

begriffs: OK, since it's an existing issue, we can merge this PR and fix that one in its own.

(Cool xmax trick btw!)

PR #1054, opened 2018-01-31
Add tables aliasing to avoid conflict in embeds with self reference fks

### LINKED ISSUE #987: self-reference foreign keys, can get "parent" but not "child"
Refiling our problem as the last one was poorly worded to our intent (#960 which I'll close in a moment).

For reference sake, here's a FK example with 2 separate tables:
```sql
create table one (
id text not null primary key,
name text not null,
other text);

create table two (
id text not null primary key,
name text not null);

alter table only one add constraint otherptr foreign key (other)  references two(id);

insert into two values ('21', 'twenty-one');
insert into two values ('22', 'twenty-two');

insert into one values ('11', 'eleven', '21');
insert into one values ('12', 'twelve', '22');

grant all on table one to testuser;
grant all on table two to testuser;
```

Some queries:
```bash
# all parents (others) of one members
% curl 'http://localhost:3000/one?select=other(*)'
[{"other":{"id":"21","name":"twenty-one"}},{"other":{"id":"22","name":"twenty-two"}}]

# all children of two members (ones that have others pointing to two)
% curl 'http://localhost:3000/two?select=one.other(*)'
[{"one":[{"id":"11","name":"eleven","other":"21"}]},{"one":[{"id":"12","name":"twelve","other":"22"}]}]

# only 11
% curl 'http://localhost:3000/one?id=eq.11'
[{"id":"11","name":"eleven","other":"21"}]

# parent (other) of 11
% curl 'http://localhost:3000/one?id=eq.11&select=other(*)'
[{"other":{"id":"21","name":"twenty-one"}}]

# only 21
% curl 'http://localhost:3000/two?id=eq.21'
[{"id":"21","name":"twenty-one"}]

# child of 21 (that with other pointing to 21)
% curl 'http://localhost:3000/two?id=eq.21&select=one.other(*)'
[{"one":[{"id":"11","name":"eleven","other":"21"}]}]

# There was some hint you could use the FK as a qualifier, but no (not that it matters)
% curl 'http://localhost:3000/two?select=otherptr(*)'
{"message":"Could not find foreign keys between these entities, No relation found between two and otherptr"}
% curl 'http://localhost:3000/two?select=one.otherptr(*)'
{"message":"Could not find foreign keys between these entities, No relation found between two and one"}
```

Now with self-referencing FKs
Table creation...
```sql
create table family_tree (
id text not null primary key,
name text not null,
parent text);

alter table only family_tree add constraint pptr foreign key (parent) references family_tree(id);

insert into family_tree values ('1', 'Parental Unit', NULL);
insert into family_tree values ('2', 'Kid One', '1');
insert into family_tree values ('3', 'Kid Two', '1');
insert into family_tree values ('4', 'Grandkid One', '2');
insert into family_tree values ('5', 'Grandkid Two', '3');

grant all on table family_tree to testuser;
```

Now some queries against the above:
```bash
# dump of whole tree
% curl 'http://localhost:3000/family_tree'
[{"id":"1","name":"Parental Unit","parent":null},{"id":"2","name":"Kid One","parent":"1"},{"id":"3","name":"Kid Two","parent":"1"},{"id":"4","name":"Grandkid One","parent":"2"},{"id":"5","name":"Grandkid Two","parent":"3"}]

# dump "Kid Two" from middle generation
% curl 'http://localhost:3000/family_tree?id=eq.3'
[{"id":"3","name":"Kid Two","parent":"1"}]

# Get kid two parents:
% curl 'http://localhost:3000/family_tree?id=eq.3&select=parent(*)'
[{"parent":{"id":"1","name":"Parental Unit","parent":null}}]

# Get kid two kids?: (should be Grandkid Two)
% curl 'http://localhost:3000/family_tree?id=eq.3&select=family_tree.parent(*)'
[{"family_tree":[]}]   <--------- Broken case here
```

Does that all make sense?  So we think we should be able to pull in not only the entries pointed AT by the FKs, but grab the entries doing the pointing (the children in this case).  Which works with 2 tables, but not 1.

steve-chavez:
Fixes #987.  Using the `family_tree` TABLE provided in https://github.com/begriffs/postgrest/issues/987#issue-260720899 we do a: 
`GET /family_tree?id=eq.3&select=family_tree.parent(*)` and currently this generates the following query:
```sql
SELECT COALESCE((
  SELECT array_to_json(array_agg(row_to_json("family_tree"))) 
  FROM (
    SELECT  "test"."family_tree".* FROM  "test"."family_tree"  
    WHERE  "test"."family_tree"."parent" = "test"."family_tree"."id") "family_tree"),
 '[]') AS "family_tree" FROM  "test"."family_tree"  WHERE  "test"."family_tree"."id" = '3'::unknown;
-- Gives []
-- It should give [{"id":"5","name":"Grandkid Two","parent":"3"}]   
```
The problem is that the subquery should be referring to the parent `"test"."family_tree"`, to get the correct result the query should be like:
```sql
SELECT COALESCE((
  SELECT array_to_json(array_agg(row_to_json("family_tree"))) 
  FROM (
    SELECT  "test"."family_tree".* FROM  "test"."family_tree"  
    WHERE  "test"."family_tree"."parent" = x."id") "family_tree"),
 '[]') AS "family_tree" FROM  "test"."family_tree" x  WHERE  x."id" = '3'::unknown;
-- Gives [{"id":"5","name":"Grandkid Two","parent":"3"}]
```
So to solve this the parent `"test"."family_tree"` needs to be aliased.
To generalize the fix I initially thought of adding an uuid alias to every FROM but what I've done is to add a `Level` attribute to the `ReadNode` type that represents the level of the tree where the node is at and then when building the query the FROM target can be aliased as `target_{level}`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #828, Fix computed column only working in public schema - @steve-chavez
 - #925, Fix RPC high memory usage by using parametrized query and avoiding json encoding - @steve-chavez
+- #987, Fix embedding with self-reference foreign key - @steve-chavez
 
 ### Changed
 
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -31,7 +31,7 @@ import           PostgREST.ApiRequest   ( ApiRequest(..)
                                         )
 import           PostgREST.Error           (apiRequestError)
 import           PostgREST.Parsers
-import           PostgREST.RangeQuery      (NonnegRange, restrictRange)
+import           PostgREST.RangeQuery      (NonnegRange, restrictRange, allRange)
 import           PostgREST.Types
 
 import           Protolude                hiding (from, dropWhile, drop)
@@ -43,8 +43,10 @@ readRequest maxRows allRels proc apiRequest  =
   mapLeft apiRequestError $
   treeRestrictRange maxRows =<<
   augumentRequestWithJoin schema relations =<<
-  parseReadRequest
+  addFiltersOrdersRanges apiRequest <*>
+  (buildReadRequest <$> pRequestSelect (iSelect apiRequest))
   where
+    action = iAction apiRequest
     (schema, rootTableName) = fromJust $ -- Make it safe
       let target = iTarget apiRequest in
       case target of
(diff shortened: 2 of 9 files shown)


begriffs:
Seems to work fine. The code is complex, but the results are cool.

begriffs:
(When you're comfortable with this branch, it looks OK to merge, at least to my not fully-comprehending pass over it. Passes the tests and all.)

PR #1061, opened 2018-02-09
Include foreign tables in API description

rhyamada:
Added missed relkind for foreign table.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -317,7 +317,7 @@ allTables =
           AND (pg_trigger.tgtype::integer & 69) = 69) ) AS insertable
     FROM pg_class c
     JOIN pg_namespace n ON n.oid = c.relnamespace
-    WHERE c.relkind IN ('v','r','m')
+    WHERE c.relkind IN ('v','r','m','f')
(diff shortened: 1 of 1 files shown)


begriffs:
Thanks, have you built this locally and confirmed that the foreign tables show up?

rhyamada:
Yes, Foreign tables are correctly displayed I test only get methods.

begriffs:
Cool! So how might we add a regression test to prevent this from breaking in the future? Maybe include a CSV file in our test fixtures and use [file_fdw](https://www.postgresql.org/docs/10/static/file-fdw.html) to access it as a table? Are you interested in giving that a try?

begriffs:
@rhyamada want some suggestions about how to modify the test suite, or would you prefer I write the test?

steve-chavez:
I'll add a test for this and merge it before next release.

steve-chavez:
I don't see the foreign table in the OpenAPI description with this change, however if I add the change(`'f'`) to the `accessibleTables` query, I do see the foreign table showing.

steve-chavez:
Closing this since #1101 got merged. Thanks @rhyamada.

PR #1063, opened 2018-02-10
Add the postgREST process' environment to the DB's SESSION

### LINKED ISSUE #1062: Where to store secrets in database?
I've been following along with the excellent documentation at https://postgrest.com/en/v4.4/auth.html, and the suggested place to store app secrets is as a property of the database:

```
-- run this once
ALTER DATABASE mydb SET "app.jwt_secret" TO 'reallyreallyreallyreallyverysafe';

-- then all functions can refer to app.jwt_secret
SELECT sign(
  row_to_json(r), current_setting('app.jwt_secret')
) AS token
FROM ...
```

However, it appears that this is not allowed on Amazon RDS:

```
beagle=> ALTER DATABASE beagle SET "app.test" TO "value";
ERROR:  permission denied to set parameter "app.test"
```

(the same `ALTER DATABASE` call works locally for me).

Do you have any other suggestions for keeping app secrets in a secure place, i.e. not stored insecurely at rest?

canadaduane:
- allows queries to refer to current_setting('app.env.[ENV]') to
  retrieve environment variables
- useful for 12-factor apps (app data can be in environment)
- provides workaround for AWS Relational Database Service (RDS) not
  allowing `ALTER DATABASE SET 'app.[KEY]' TO '[VALUE]'` on database.

Fixes #1062 

UPDATE: Now uses `current_settings('app.settings.*')` and a more generic way of getting data into those settings--rather than dump the entire ENV into the session, the config file can now specify which ENV vars, if any, are added to the db session, along side other strings and app setting data.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -76,13 +76,13 @@ connectionWorker mainTid pool schema settings refDbStructure refIsWorkerOn = do
       connected <- connectingSucceeded pool
       when connected $ do
         result <- P.use pool $ do
-          fillSessionWithSettings settings
           actualPgVersion <- getPgVersion
           unless (actualPgVersion >= minimumPgVersion) $ liftIO $ do
             hPutStrLn stderr
               ("Cannot run in this PostgreSQL version, PostgREST needs at least "
               <> pgvName minimumPgVersion)
             killThread mainTid
+          fillSessionWithSettings settings
           dbStructure <- getDbStructure schema actualPgVersion
           liftIO $ atomicWriteIORef refDbStructure $ Just dbStructure
         case result of
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -64,7 +64,7 @@ connectionWorker
   -> IORef (Maybe DbStructure) -- ^ mutable reference to 'DbStructure'
   -> IORef Bool                -- ^ Used as a binary Semaphore
   -> IO ()
-connectionWorker mainTid pool schema environ refDbStructure refIsWorkerOn = do
+connectionWorker mainTid pool schema settings refDbStructure refIsWorkerOn = do
   isWorkerOn <- readIORef refIsWorkerOn
   unless isWorkerOn $ do
     atomicWriteIORef refIsWorkerOn True
@@ -76,7 +76,7 @@ connectionWorker mainTid pool schema environ refDbStructure refIsWorkerOn = do
       connected <- connectingSucceeded pool
       when connected $ do
         result <- P.use pool $ do
-          fillSessionWithSettings environ
+          fillSessionWithSettings settings
           actualPgVersion <- getPgVersion
           unless (actualPgVersion >= minimumPgVersion) $ liftIO $ do
             hPutStrLn stderr
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -7,7 +7,8 @@ import           PostgREST.App            (postgrest)
 import           PostgREST.Config         (AppConfig (..),
                                            minimumPgVersion,
                                            prettyVersion, readOptions)
-import           PostgREST.DbStructure    (getDbStructure, getPgVersion, fillSessionWithEnvironment)
+import           PostgREST.DbStructure    (getDbStructure, getPgVersion,
+                                           fillSessionWithSettings)
 import           PostgREST.Error          (encodeError)
 import           PostgREST.OpenAPI        (isMalformedProxyUri)
 import           PostgREST.Types          (DbStructure, Schema, PgVersion(..))
@@ -32,7 +33,7 @@ import           Network.Wai.Handler.Warp (defaultSettings,
                                            setTimeout)
 import           System.IO                (BufferMode (..),
                                            hSetBuffering)
-import           System.Environment       (getEnvironment)
+
 #ifndef mingw32_HOST_OS
 import           System.Posix.Signals
 #endif
@@ -59,10 +60,11 @@ connectionWorker
   :: ThreadId -- ^ This thread is killed if pg version is unsupported
   -> P.Pool   -- ^ The PostgreSQL connection pool
   -> Schema   -- ^ Schema PostgREST is serving up
+  -> [(Text, Text)] -- ^ Settings or Environment passed in through the config
   -> IORef (Maybe DbStructure) -- ^ mutable reference to 'DbStructure'
   -> IORef Bool                -- ^ Used as a binary Semaphore
   -> IO ()
-connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
+connectionWorker mainTid pool schema environ refDbStructure refIsWorkerOn = do
   isWorkerOn <- readIORef refIsWorkerOn
   unless isWorkerOn $ do
     atomicWriteIORef refIsWorkerOn True
@@ -70,12 +72,11 @@ connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
   where
     work = do
       atomicWriteIORef refDbStructure Nothing
-      environ <- getEnvironment
       putStrLn ("Attempting to connect to the database..." :: Text)
       connected <- connectingSucceeded pool
       when connected $ do
         result <- P.use pool $ do
-          fillSessionWithEnvironment environ
+          fillSessionWithSettings environ
           actualPgVersion <- getPgVersion
           unless (actualPgVersion >= minimumPgVersion) $ liftIO $ do
             hPutStrLn stderr
@@ -176,6 +177,7 @@ main = do
     mainTid
     pool
     (configSchema conf)
+    (configSettings conf)
     refDbStructure
     refIsWorkerOn
   --
@@ -198,6 +200,7 @@ main = do
               mainTid
               pool
               (configSchema conf)
+              (configSettings conf)
               refDbStructure
               refIsWorkerOn
     ) Nothing
@@ -214,6 +217,7 @@ main = do
          mainTid
          pool
          (configSchema conf)
+         (configSettings conf)
          refDbStructure
          refIsWorkerOn)
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -7,7 +7,7 @@ import           PostgREST.App            (postgrest)
 import           PostgREST.Config         (AppConfig (..),
                                            minimumPgVersion,
                                            prettyVersion, readOptions)
-import           PostgREST.DbStructure    (getDbStructure, getPgVersion)
+import           PostgREST.DbStructure    (getDbStructure, getPgVersion, fillSessionWithEnvironment)
 import           PostgREST.Error          (encodeError)
 import           PostgREST.OpenAPI        (isMalformedProxyUri)
 import           PostgREST.Types          (DbStructure, Schema, PgVersion(..))
@@ -32,6 +32,7 @@ import           Network.Wai.Handler.Warp (defaultSettings,
                                            setTimeout)
 import           System.IO                (BufferMode (..),
                                            hSetBuffering)
+import           System.Environment       (getEnvironment)
 #ifndef mingw32_HOST_OS
 import           System.Posix.Signals
 #endif
@@ -69,10 +70,12 @@ connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
   where
     work = do
       atomicWriteIORef refDbStructure Nothing
+      environ <- getEnvironment
       putStrLn ("Attempting to connect to the database..." :: Text)
       connected <- connectingSucceeded pool
       when connected $ do
         result <- P.use pool $ do
+          fillSessionWithEnvironment environ
           actualPgVersion <- getPgVersion
           unless (actualPgVersion >= minimumPgVersion) $ liftIO $ do
             hPutStrLn stderr
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -168,9 +168,9 @@ readOptions = do
     coerceInt (String x) = readMaybe $ toS x
     coerceInt _          = Nothing
 
-    coerceBool ::  Value -> Maybe Bool
+    coerceBool :: Value -> Maybe Bool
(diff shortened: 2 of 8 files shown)


begriffs:
Clever solution! I'll look through the code and give more detailed feedback.

begriffs:
OK, so I'm running your branch locally, and I see the `set_config` commands in my database log. However when I start up a psql connection to the same db I'm unable to view these values with `current_setting`. I don't think they are visible outside of the session that set them.

Normally we set these kind of variables in a middleware on every request. https://github.com/begriffs/postgrest/blob/516976e32fa3d8b8e8cc6b6e02bc02ee1e4964ca/src/PostgREST/Middleware.hs#L35

You could save the list of environment variables to an IORef the way we do for dbStructure, and then use that value from the middlware instead.

ruslantalpa:
I think the implementation in this PR does it in the correct way.
The env vars are exposed for the benefit of PostgREST and requests coming in through it, hence having them in session scope and not visible to other db clients is better (PostgREST not polluting the database env for everybody). If the user needs some "global settings" he can use other methods (like storing them in tables).
The second comment about setting them is middleware also has a problem. These variables do not change with each request so setting them in middleware will bring in overhead (transaction scope) with each individual request. Setting them in session scope (each pool connection) is better.

The only thing to maybe consider a bit is if we want all env vars to end up in the db or we want a way  restrict them to a limited set (like having a config `expose-env-vars` with possible settings like `all` `none` `A, B, C`)

begriffs:
> I think the implementation in this PR does it in the correct way.

It would certainly be best to set these vars once when grabbing connections from the pool rather than redundantly on every request, I agree. It's just that I had trouble when testing it locally last night. I picked the PATH env var (because I know it exists) then I created a view:

```sql
CREATE VIEW pp AS SELECT current_setting('app.env.PATH');
```

Then I tried to request it:

```
$ curl localhost:3000/pp
{"hint":null,"details":null,"code":"42704","message":"unrecognized configuration parameter \"app.env.PATH\""}
```

It's weird because I had built postgrest on the branch for this PR, so I thought it would have picked up the environment variable. This morning I tried again and found that after restarting postgrest the view works. I don't know why that would be. Could you try a similar experiment and see whether such a view created while postgrest is running will work the first time around, or whether you have to restart the server?

> we want a way restrict them to a limited set

I was thinking the same thing last night when I saw the big sql dump in my db log with all the vars. What if any `expose-*` param in the config file gets exposed to the db, so we could leverage the config file's ability to interpolate environment vars? So you could say

```ini
expose-jwt-secret = "$(POSTGREST_JWT_SECRET)"
```

This could be exposed in the db as `app.conf.jwt-secret`. It is slightly less powerful than what you suggested because there isn't an "all" setting to grab them all, but maybe it's good to be explicit about what vars to expose? On the flip side, this technique allows exposing value literals that aren't present as environment variables.

ruslantalpa:
I like the using the explicit config lines rather then going directly to read the env from the code.
The only comment would be:
- Can the haskell conf lib provide a way to get the conf keys in a array so you can filter them?
- there is a disconnect between `expose-jwt-secret` `app.conf.jwt-secret`, can we make them at least similar in both envs (conf file and db?) like `app.conf.jwt-secret`

begriffs:
Configurator-ng provides the [subassocs](https://hackage.haskell.org/package/configurator-ng-0.0.0.1/docs/Data-Configurator-Parser.html#v:subassocs) function for getting `[(Name, Value)]`.

For the name you're proposing that the config file has the following?

```ini
app.config. jwt-secret = "$(POSTGREST_JWT_SECRET)"
```

I believe configurator-ng also allows this to be written

```
app {
  config {
    jwt-secret = "$(POSTGREST_JWT_SECRET)"
  }
}
```

I do like that the names match up with what a person will request from inside the db, although it feels weird to write "config" from inside a config file. :man_shrugging:

ruslantalpa:
`app.setting.jwt-secret` :)

canadaduane:
Hey, just getting around to reading your comments--thanks for taking note, and yes, I think all of what you've said is the direction I would go as well (keeping the environment vars local to the db session provides a transient/ephemeral and app-local way to pass and keep secrets for functions in the database).

Being explicit about what environment variables are passed through is ok, but I would really like for there to be a way to pass env vars without modifying the config file--especially when running in docker. If we required explicit pass-through, then one would need to modify both the config file AND the environment, which requires more work when setting up this "stateless" server.

If it's important, and no `ALL` option is available, an alternative could be to automatically pass through environment variables that are prefixed with `PGRST_`. It's a bit of an opinionated way to do it, but I'd rather that than configure in two places.

ruslantalpa:
In docker you can have a trivial entrypoint script that takes all env vars and apends them to the internal config file.
The specific config line per env var is better as a general (flexible) solution and the ALL functionality can be accomolished in other ways.

canadaduane:
I think this resolves all currently outstanding issues. Any thoughts on merging?

src/PostgREST/DbStructure.hs
```diff
@@ -747,3 +752,19 @@ getPgVersion = H.query () $ H.statement sql HE.unit versionRow False
   where
     sql = "SELECT current_setting('server_version_num')::integer, current_setting('server_version')"
     versionRow = HD.singleRow $ PgVersion <$> HD.value HD.int4 <*> HD.value HD.text
+
+fillSessionWithSettings :: [(Text, Text)] -> H.Session ()
+fillSessionWithSettings settings =
+    -- combine all environment-setting SQL queries into one Hasql.Session
+    sequence_ $ List.map buildSetConfigQuery settings
```
ruslantalpa: This sequence_ call will generate a roundtrip to the db for each var resulting in a slower "first request". This will be especially painful for small deployments where there is no big stream of requests and the connection is released after 1-2 seconds so basically every request will be a slow "first request".
All the above queries can be done in a single roundtrip like we do in middleware
`H.sql $ toS.mconcat $ setSchemaSql ++ setRoleSql ++ claimsSql ++ headersSql ++ cookiesSql`
so the resulting string will be `select set_config ...; select set_config...;` or even better it can be
`select set_config(), set_config()`.
Ideally you would do it with a prepared statement like `select set_config($1, $2), set_config($3, $4) ...` if you can get hasql to run a dynamically generated prepared statement, if not, then just `H.sql` with the vars in the string directly (piped through our formatting functions) will be ok.

(everything else looks great)

canadaduane: Ok, I'll see what I can do :)

canadaduane:
BTW, I realized I hadn't run tests, and will fix the 1 that I broke.

canadaduane:
I think this is ready...

src/PostgREST/Config.hs
```diff
@@ -145,6 +147,13 @@ readOptions = do
       return appConf
 
   where
+    parsedPairToTextPair :: (Name, Value) -> (Text, Text)
+    parsedPairToTextPair (k, v) = (k, newValue)
+      where
+        newValue = case v of
```
begriffs: `newValue` looks similar to functions like `coerceInt` and `coerceBool`. What if you called it `coerceText` and added it near those others?

Then

```haskell
parsedPairToTextPair = id *** coerceText
```
Unless you feel like this last definition is a little too weird looking.

src/PostgREST/DbStructure.hs
```diff
@@ -747,3 +753,17 @@ getPgVersion = H.query () $ H.statement sql HE.unit versionRow False
   where
     sql = "SELECT current_setting('server_version_num')::integer, current_setting('server_version')"
     versionRow = HD.singleRow $ PgVersion <$> HD.value HD.int4 <*> HD.value HD.text
+
+fillSessionWithSettings :: [(Text, Text)] -> H.Session ()
+fillSessionWithSettings settings =
+    -- Send all of the config settings to the set_config function, using pgsql's `unnest` to transform arrays of values
+    H.query settings $ H.statement "SELECT set_config(unnest($1), unnest($2), false)" encoder HD.unit False
```
begriffs: Wow that's clever.

canadaduane:
I'm ok with that. I'm away from my computer right now. If you'd like to
make the change, I'd be grateful.

src/PostgREST/DbStructure.hs
begriffs: …maybe too clever actually. Look at the logs from CI, they repeat this over and over:
```json
{
  "hint":null,
  "details":null,
  "code":"0A000",
  "message":"functions and operators can take at most one set argument"
}
```

canadaduane: That's odd... it's working for me locally.

begriffs: Maybe I'm misdiagnosing the cause of that message. The tests are getting stuck though.

canadaduane: No, I think you're right. I was seeing the postgresql log and believing it to be successful:

```
2018-02-12 17:46:36.836 MST [18209] LOG:  statement: SET client_encoding = 'UTF8';SET client_min_messages TO WARNING;
2018-02-12 17:46:36.839 MST [18209] LOG:  statement: SELECT 1
2018-02-12 17:46:36.845 MST [18209] LOG:  execute <unnamed>: SELECT set_config(unnest($1), unnest($2), false)
2018-02-12 17:46:36.845 MST [18209] DETAIL:  parameters: $1 = '{app.settings.test,app.settings.test2}', $2 = '{123/hi,/Users/duane/ok}'
```

But it isn't setting the values.

src/PostgREST/Config.hs
canadaduane: I gave up on `id *** coerceText`. If you can apply your haskell fu successfully, I will dispatch my wizard hat forthwith.

canadaduane:
I don't have a complete test database setup locally, so I'll wait on `circleci` to tell me if I was successful. But I think the `FROM` clause solves the previous issue with `unnest`.

begriffs:
Nice, passing tests!

* Could you include a line in the CHANGELOG.md (under the "Added" section) for your change?
* Also can you add a test that shows this feature working?
   * Add a value to `configSettings` in _baseCfg (you have `[]` in it currently)
   * Make a new feature test that posts to `/rpc/get_guc_value` and assert that the new value appears for `app.settings.foo` - example in  https://github.com/begriffs/postgrest/blob/85d9feeeabb7ee3ede9515d7f6a9fba1c0a3124a/test/Feature/QuerySpec.hs#L687-L688

canadaduane:
It appears the tests do not run the same session start code as the main app? I don't have time to look into it right now, but it seems `connectionWorker` is not called by tests?

canadaduane:
Any thoughts on merging this?

main/Main.hs
```diff
@@ -73,6 +76,7 @@ connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
       connected <- connectingSucceeded pool
       when connected $ do
         result <- P.use pool $ do
+          fillSessionWithSettings settings
```
steve-chavez: This should go after the minimum PostgreSQL version validation, 6 lines below, otherwise it'll give an error on older versions because `unnest` with two args is only available on `>= 9.4`.

test/Main.hs
```diff
@@ -58,7 +58,12 @@ main = do
       asymJwkApp           = return $ postgrest (testCfgAsymJWK testDbConn)     refDbStructure pool $ pure ()
       nonexistentSchemaApp = return $ postgrest (testNonexistentSchemaCfg testDbConn)   refDbStructure pool $ pure ()
 
-  let reset = resetDb testDbConn
+  let reset :: IO ()
+      reset = do
+          r <- resetDb testDbConn
+          _ <- P.use pool $ fillSessionWithSettings (configSettings $ testCfg testDbConn)
+          pure r
```
steve-chavez: Could be reduced to:
```haskell
reset = P.use pool (fillSessionWithSettings (configSettings $ testCfg testDbConn)) >> resetDb testDbConn
```

steve-chavez:
As it is this PR would break PostgreSQL `9.3` clients, which we currently support as the minimum version(see [minimumPgVersion](https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Config.hs#L219-L221)), this is because of the `unnest(anyarray, anyarray)` function available on `>= 9.4`.

One way to solve this would be conditionally enabling the feature but since in #1005 the need for `9.4` in RPC was introduced I suggest just increasing the `minimumPgVersion` to:

```haskell
minimumPgVersion :: PgVersion
minimumPgVersion = PgVersion 90400 "9.4"
```

steve-chavez:
Also @begriffs I saw in the docs that `9.5` is required(change [here](https://github.com/begriffs/postgrest-docs/commit/5fc48057f8c4d01f520bb4009051f8aa267cda15)) but that's not accurate, we run tests on CircleCI with the minimum of `9.4`, so maybe the docs should be corrected to `9.4` after this PR is merged.

canadaduane:
@steve-chavez How does that look?

steve-chavez:
It's good, I've tested the validation on a PostgreSQL 9.3 docker container and it's working ok.

canadaduane:
Ready to merge?

canadaduane:
Awesome, thanks!

Just for my own planning purposes, do you plan to release a new version soon?

begriffs:
How about a new version after we get the open PRs resolved? It would then have quite a few nice features.

canadaduane:
Ok, thanks!

revolunet:
Does this mean we can use `SELECT current_settings('app.env.JWT_SECRET');` when process.Env.JWT_SECRET is set on postgrest environment ?

steve-chavez:
@revolunet Only if you have `app.settings.jwt_secret = "$(JWT_SECRET)"` on your config file.

Then you can use `SELECT current_setting('app.settings.jwt_secret')`.

Check http://postgrest.org/en/v7.0.0/configuration.html#app-settings.

revolunet:
Thanks @steve-chavez so the `JWT_SECRET` is not "hardcoded" into the config file which is great 👍

PR #1074, opened 2018-02-26
Fix Parent Relation when having many views

### LINKED ISSUE #1044: Unable to get embedded resource via view in a specific scenario
[instrospection.json.txt](https://github.com/begriffs/postgrest/files/1609752/instrospection.json.txt)

[schema.sql.txt](https://github.com/begriffs/postgrest/files/1609751/schema.sql.txt)
I'm using PostgREST 0.4.3.0 on macosx 10.13.2 and docker postgresql container with version 10.1. I have two similar views referencing the same table, one with join:

```
postgres=# \d+ api.fixtures
                                         View "api.fixtures"
     Column     |           Type           | Collation | Nullable | Default | Storage  | Description
----------------+--------------------------+-----------+----------+---------+----------+-------------
 id             | integer                  |           |          |         | plain    |
 competition_id | integer                  |           |          |         | plain    |
 home_team_id   | integer                  |           |          |         | plain    |
 away_team_id   | integer                  |           |          |         | plain    |
 result_id      | integer                  |           |          |         | plain    |
 date_time      | timestamp with time zone |           |          |         | plain    |
 status         | text                     |           |          |         | extended |
 matchday       | integer                  |           |          |         | plain    |
View definition:
 SELECT fixtures.id,
    fixtures.competition_id,
    fixtures.home_team_id,
    fixtures.away_team_id,
    fixtures.result_id,
    fixtures.date_time,
    fixtures.status,
    fixtures.matchday
   FROM data.fixtures;


postgres=# \d+ api.current_fixtures
                                     View "api.current_fixtures"
     Column     |           Type           | Collation | Nullable | Default | Storage  | Description
----------------+--------------------------+-----------+----------+---------+----------+-------------
 id             | integer                  |           |          |         | plain    |
 competition_id | integer                  |           |          |         | plain    |
 home_team_id   | integer                  |           |          |         | plain    |
 away_team_id   | integer                  |           |          |         | plain    |
 result_id      | integer                  |           |          |         | plain    |
 date_time      | timestamp with time zone |           |          |         | plain    |
 status         | text                     |           |          |         | extended |
 matchday       | integer                  |           |          |         | plain    |
View definition:
 SELECT f.id,
    f.competition_id,
    f.home_team_id,
    f.away_team_id,
    f.result_id,
    f.date_time,
    f.status,
    f.matchday
   FROM data.fixtures f
     LEFT JOIN data.competitions c ON f.competition_id = c.id
  WHERE f.matchday = c.current_matchday;
```

Selecting any embedded resource works on just only one of them if they're both present in schema. If I drop the one which works and restart postgrest, the other one starts working.
So for example this works:
```
GET 'http://localhost:3000/current_fixtures?select=id,competition(id,caption)'
```

But this doesn't:
```
$ GET 'http://localhost:3000/fixtures?select=id,competition(id,caption)'
HTTP/1.1 400 Bad Request
Content-Encoding: gzip
Date: Sun, 07 Jan 2018 12:27:24 GMT
Server: postgrest/0.4.3.0
Transfer-Encoding: chunked

{
"message": "Could not find foreign keys between these entities, No relation found between fixtures and competition"
}
```

But after `drop view api.current_fixtures;` and restarting postgrest
`GET 'http://localhost:3000/fixtures?select=id,competition(id,caption)'` works as expected.

schema and output from instrospection.sql attached.

steve-chavez:
Fixes #1044.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE QuasiQuotes           #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE TypeSynonymInstances  #-}
+{-# LANGUAGE NamedFieldPuns  #-}
 module PostgREST.DbStructure (
   getDbStructure
 , accessibleTables
@@ -18,9 +19,8 @@ import qualified Hasql.Query                   as H
 
 import           Control.Applicative
 import qualified Data.HashMap.Strict           as M
-import           Data.List                     (elemIndex)
-import qualified Data.List                     as List
-import           Data.Maybe                    (fromJust)
+import qualified Data.List                     as L
+import           Data.Set                      as S (fromList)
 import           Data.Text                     (split, strip,
                                                 breakOn, dropAround,
                                                 splitOn)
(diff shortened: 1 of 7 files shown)


ruslantalpa:
that `raiseRelations` has always been a bit cryptic :) glad it's gone. nice job

begriffs:
Thanks Steve, I'll review this tomorrow.

src/PostgREST/DbStructure.hs
```diff
@@ -248,17 +236,71 @@ addForeignKeys rels = map addFk
...
+When having t1_view.c1 and a t2_view.c2 synonyms, we need to add a View to View Relation in addition to the prior
+
+         t1.c1----t2.c2               t1.c1----------t2.c2
+                               ->          \--------/
+                                           /        \
+    t1_view.c1     t2_view.c2     t1_view.c1-------t2_view.c1
+
+The logic for composite pks is similar just need to make sure all the Relation columns have synonyms.
+-}
+addViewRelations :: [Synonym] -> [Relation] -> [Relation]
```
begriffs: Internally this function has the shape of a [concatMap](https://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#v:concatMap), perhaps structuring it around that would simplify it and remove the need for explicit recursion?

begriffs:
Nice, this code does seem more uniform than what we had before. I added one comment about a possible further simplification, see if you find it helpful. Otherwise looks good to me.

jlsphar:
Nice! We ran into this issue today as well. Thanks for the fix.

src/PostgREST/DbStructure.hs
steve-chavez: Changed the `addViewRelations` function and other similar ones to use `concatMap`.

begriffs:
Looks good to me!

PR #1080, opened 2018-03-03
Update stack.yaml to use the latest LTS-10

PierreR:


steve-chavez:
Also noted a dependency issue for GHC 8.2.2 in https://github.com/PostgREST/postgrest/issues/1109#issuecomment-417073687, let's continue this there.

Issue #934, opened 2017-08-10
Upgrade hjsonschema version(or remove it) in cabal file


begriffs:
We held it back because later versions made the test suite fail (even though it appears that they generate valid OpenAPI).

PierreR:
This is still an issue see https://github.com/begriffs/postgrest/pull/1081

Getting the test suite to work (without a docker) is getting quite painful because of this kind of pinning. As a matter of context, I was hoping to have a go with the replacement of `configurator-ng` with `dhall` and these kind of issues are not encouraging me to go further.

steve-chavez:
I've tried upgrading to hjsonschema 1.6.3 and 1.9.0, both cause this [test](https://github.com/PostgREST/postgrest/blob/master/test/Feature/StructureSpec.hs#L24) to fail and somehow it makes the test suite run hang while consuming 100% CPU.

Seeing the lib is now deprecated https://github.com/seagreen/hjsonschema, I think the best would be to try to remove it, we use it in this [part](https://github.com/PostgREST/postgrest/blob/master/test/SpecHelper.hs#L53-L61). 

For now we'll stay with the 1.6.0.1 version.

steve-chavez:
hjsonschema makes our test suite fail often, it gets tedious to manually restart the CircleCI job from time to time.

steve-chavez:
The hjsonschema test times out even more frequently now. I'm going to use [pendingWith](http://hackage.haskell.org/package/hspec-wai-0.9.2/docs/Test-Hspec-Wai.html#v:pendingWith) to comment it.

Using the postgres-json-schema and doing the test in bash would be the solution. Then we can remove hjsonschema.

PR #1081, opened 2018-03-03
Update hjsonschema to >= 1.6

PierreR:


steve-chavez:
Let's continue this in #934, see my latest comment https://github.com/PostgREST/postgrest/issues/934#issuecomment-417323827.

PR #1085, opened 2018-03-16
Allow embeds alias to be used in filters, specifying source column in embed and only alias tables on self join

steve-chavez:
On #1054 aliasing was added to all tables to fix the self join cases, however that made the error messages a bit unclear:
```http
GET "/items?select=id,wrong"
{"hint":null,"details":null,"code":"42703","message":"column items_1.wrong does not exist"}

GET "/clients?select=*,projects(id,name,wrong)"
{"hint":null,"details":null,"code":"42703","message":"column projects_2.wrong does not exist"}

GET "localhost:3000/tasks?select=id,name,project(id,name,client(id,name,wrong))"
{"hint":null,"details":null,"code":"42703","message":"column tasks_3.wrong does not exist"}
```

There may be other cases where the error messages and generated queries are harder to understand.
Self reference fks are rare in db designs, so I thought it'd be better to restrict the aliasing to only self joins.


begriffs:
Thanks for this! I'll add a review in the next day or two.

begriffs:
Actually Steve I trust your judgement. I'll let you do with this PR as you think best.

PR #1103, opened 2018-05-08
Use protolude 0.2.2

felixonmars:
protolude provides headMay since 0.2.2, it can be used instead of safe.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -76,10 +76,9 @@ library
                      , network-uri
                      , optparse-applicative >= 0.13 && < 0.15
                      , parsec
-                     , protolude >= 0.2 && < 0.2.2
+                     , protolude >= 0.2.2
                      , Ranged-sets == 0.3.0
                      , regex-tdfa
-                     , safe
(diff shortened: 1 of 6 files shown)


ilovezfs:
@felixonmars you'll probably want to rebase this now that https://github.com/PostgREST/postgrest/pull/1111 is merged CC @steve-chavez

felixonmars:
Rebased :)

steve-chavez:
Seems the problem is that Protolude added warnings for `undefined` among other functions in https://github.com/sdiehl/protolude/pull/70 and we compile with [-Werror](https://downloads.haskell.org/~ghc/7.0.3/docs/html/users_guide/options-sanity.html).

We could compile with `-fnowarn-warnings-deprecations`, but not sure it's worth giving up on that strictness, I think the best would be to refactor and remove the `undefined`, maybe we should hold upgrading Protolude until that it's done.

steve-chavez:
Thank you @felixonmars, the less dependencies the better.

PR #1104, opened 2018-05-08
Fix accepting misspellings in order syntax

steve-chavez:
Continuing the work on #1047, also included a test for multi-column sorting.


steve-chavez:
Fixes #781.

Issue #1110, opened 2018-05-15
GHC 8.2.2 build failure


ilovezfs:
### Environment

* Version of postgrest: 0.5.0.0
* Operating system: macOS
* Running in docker:
  * [ ] yes
  * [x] no

### Description of issue

```
==> cabal install --jobs=3 --max-backjumps=100000 --prefix=/usr/local/Cellar/postgrest/0.5.0.0
clang: warning: -Wl,-headerpad_max_install_names: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/opt/ghc@8.2/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/opt/openssl/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/opt/readline/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries' [-Wunused-command-line-argument]
Resolving dependencies...
Notice: installing into a sandbox located at
/private/tmp/postgrest-20180515-57214-r5uljl/postgrest-0.5.0.0/.cabal-sandbox
Configuring postgrest-0.5.0.0...
Building postgrest-0.5.0.0...
Failed to install postgrest-0.5.0.0
Build log ( /private/tmp/postgrest-20180515-57214-r5uljl/postgrest-0.5.0.0/.cabal-sandbox/logs/ghc-8.2.2/postgrest-0.5.0.0-EdQoU8JjeOk6Mne6HM7fqc.log ):
cabal: Entering directory '.'
Configuring postgrest-0.5.0.0...
clang: warning: -Wl,-headerpad_max_install_names: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/opt/ghc@8.2/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/opt/openssl/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/opt/readline/lib' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/lib' [-Wunused-command-line-argument]
[...]

PR #1111, opened 2018-05-15
postgrest.cabal: fix constraints on protolude and hasql-transaction

ilovezfs:
Fixes https://github.com/PostgREST/postgrest/issues/1110.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -64,7 +64,7 @@ library
                      , gitrev
                      , hasql
                      , hasql-pool
-                     , hasql-transaction
+                     , hasql-transaction < 0.6
                      , heredoc
                      , HTTP
                      , http-types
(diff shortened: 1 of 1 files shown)


ilovezfs:
@begriffs @steve-chavez any thoughts?

steve-chavez:
Seems ok, thanks. One question though why when using `cabal` the build fails and not when using `stack`?

ilovezfs:
You have them pinned for stack:
https://github.com/PostgREST/postgrest/blob/master/stack.yaml#L8
https://github.com/PostgREST/postgrest/blob/master/stack.yaml#L11

steve-chavez:
Forgot that, all good. Merging.

ilovezfs:
Thanks!

Issue #945, opened 2017-08-18
Slow query, possibly due to regex, affects start-up time severely


krueger71:
The sql-query in the function 
```haskell
allSynonyms :: [Column] -> H.Query () [(Column,Column)]` 
```
in [src/PostgREST/DbStructure.hs](https://github.com/begriffs/postgrest/blob/build-0.4.2.0/src/PostgREST/DbStructure.hs) runs very slowly in 0.4.1.0/0.4.2.0 compared to version 0.4.0.0 and 0.3.2.0.

The single change of the query between 0.4.0.0 and 0.4.1.0/0.4.2.0 is the following expression containing a regex:

```sql
diff sql-0.4.0.0.sql sql-0.4.1.0.sql
50c50
<                     CONCAT('SELECT.*?((',vcu.table_name,')|(\w+))\.(', vcu.column_name, ')(\s+AS\s+("([^"]+)"|([^, \n\t]+)))?.*?FROM.*?',vcu.table_schema,'\.(\2|',vcu.table_name,'\s+(as\s)?\3)'),
---
>                     CONCAT('SELECT.*?((',vcu.table_name,')|(\w+))\.(', vcu.column_name, ')(\s+AS\s+("([^"]+)"|([^, \n\t]+)))?.*?FROM.*?(',vcu.table_schema,'\.|)(\2|',vcu.table_name,'\s+(as\s)?\3)'),
```
Analyzing that line further one can see that the new version has only one change, which is this capture group expression `(',vcu.table_schema,'\.|)`.  

Query times on my particular database schema is around 11 s with 0.3.2.0, 15 s with 0.4.0.0 and a huge slow-down to 430 s in 0.4.1.0/0.4.2.0. Postgrest is unresponsive or throwing errors until this query has finished.

ruslantalpa:
@krueger71 can you attach a sample (big) schema that exposes this problem?
you can try your schema and rename to gibberish all the table/column names. 
the important part is to see how this query performs on the schema

cbandy:
The change made matching on `vcu.table_schema` optional. If that's okay, then I think we can just remove this group completely. The preceding `.*?` will cover it, and only the column name and alias are needed from the captures.

steve-chavez:
[...]

PR #1115, opened 2018-05-16
Improve the allSynonyms query

steve-chavez:
For #945, the change to the regex included in #761 seems to generate excessive backtracks, I've removed that and simply void the `search_path` for the query to have a normalized output, this was also recommended in #761 as a better alternative, the counterarguments presented there against this are at best nice-to-haves when taking into account the huge performance impact.

This is still a work in progress, I believe the query can be improved further, it's pending to generate a test case for this(programatically creating a big schema) and the `SET schema ''` is affecting a whole session, this won't be an issue because we also do a `SET schema <db-schema>` on each request, but it'd be better to isolate it in a transaction. 

Edit: This approach was discarded, see below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - #1113, Fix UPSERT failing when having a camel case PK column - @steve-chavez
+- #945, Fix slow start-up time on big schemas - @steve-chavez
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #1113, Fix UPSERT failing when having a camel case PK column - @steve-chavez
 - #945, Fix slow start-up time on big schemas - @steve-chavez
+- #1129, Fix view embedding when table is capitalized - @steve-chavez
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #1099, Add support for getting json/jsonb by array index - @steve-chavez
 - #1145, Add materialized view columns to OpenAPI output - @steve-chavez
+- #709, Allow embedding on views with subselects/CTE - @steve-chavez
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - #1099, Add support for getting json/jsonb by array index - @steve-chavez
+- #1145, Add materialized view columns to OpenAPI output - @steve-chavez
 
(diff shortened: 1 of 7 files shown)


steve-chavez:
Seeing that we now have a new issue with the current allSynonyms query https://github.com/PostgREST/postgrest/issues/1129 and to solve this the current regex would have to get more complex, I decided to discard the current query and try a different approach, some of the ones I considered:

1) Use `information_schema.view_column_usage` and insert into `pg_depend`, insert the view alias position in `pg_depend.refobjsubid` that way we would have the view columns with aliases, user would have to manually insert these rows though(thought we could offer a custom function for making the job easier), however this had an additional problem, with `view_column_usage` we can't know what are the columns/aliases at the root level, it gives all the columns used in the view(in subselects, joins, etc), so a possible solution was just to insert the view aliases used at the top and leave the rest to 0. An error prone approach.

2) Specify the view join columns in a custom option in the view, similar to what [PipelineDB does](https://www.pipelinedb.com/blog/pipelinedb-0-9-9-one-more-release-until-pipelinedb-is-a-postgresql-extension) with `CREATE VIEW continuous_view WITH (action=materialize, ...)`, with that we could have a way to specify PostgREST metadata for the views without polluting the query/view comment(discarded these approaches as well), that would enable us to specify the join columns in a `WITH (pgrst_join_cols=..)` or smth like that.
The downside is that we'd have to create a custom extension(`CREATE EXTENSION postgrest`) and define these custom options, which is troubling for cloud environments.

3) Parse [pg_rewrite](https://www.postgresql.org/docs/10/static/catalog-pg-rewrite.html) output, this includes the pg internal query tree representation in `ev_action`, the advantage is that this is more regular than the currently used `pg_get_viewdef`, when seeing this output I noticed that the `:targetList` contains all the info needed for getting the table columns used by the views, thought of parsing this in Haskell.

steve-chavez:
I ended up going with 3), and also while doing this I noticed that I could parse what was needed with just SQL. An explanation of how the query was developed is here https://gist.github.com/steve-chavez/7ee0e6590cddafb532e5f00c46275569.

The advantages of this are that now we can do embedding in views with subselects, and the query is a lot faster, running this query in barbalex's schema gave the result in 406.407 ms, previous query took around 8 mins.

steve-chavez:
Wanted a big schema to leave a perf test on our test suite, but couldn't find a sample database complex enough for this, best I could find was [AdventureWorks for Postgres](https://github.com/lorint/AdventureWorks-for-Postgres/blob/master/install.sql) but that ony has 88 views and they're spread accross different schemas.

Programatically generating a big schema seems like a lot of work, so I'm just gonna rely on the good numbers that barbalex and me got after manually running the new query.

ruslantalpa:
@steve-chavez forgot to comment on this PR, this i would say is one of the hard problems you solved (other PRs are just work, this one is research/innovation) and you solved it the right way (figuring out the right query) not removing the existing features. good job!

steve-chavez:
Thanks Ruslan. Hopefully this new allSynonyms query handles most users' views, if complicated edge cases come up I think we should be looking at doing this detection in C and propose a function in pgsql-hackers, it'll be faster than parsing and the structs are already there.

PR #1120, opened 2018-05-25
Add role-claim-key to docker postgrest.conf

theomega:
With #1091, the role-claim-key was added, but it is not configurable in docker containers as the template config doesn't allow it.
This patch adds the necessary line to the postgrest.conf of the docker container.


steve-chavez:
Thank you @theomega , I overlooked this in the latest docker release.

steve-chavez:
I've pushed an image with the change to docker hub as `postgrest/postgrest:v0.5.0.0-f6c1ff8`.

steve-chavez:
Seems we missed to add `PGRST_ROLE_CLAIM_KEY` to the `ENV`, make sure to pull `postgrest/postgrest:latest` as it contains a fix for that.

Issue #336, opened 2015-10-31
Cannot filter by JSON equality


begriffs:
e.g. `GET http://localhost:3000/json?data=eq.{"foo":"bar"}`. It returns

``` json
{
  "hint": "No operator matches the given name and argument type(s). You might need to add explicit type casts.",
  "details": null,
  "code": "42883",
  "message": "operator does not exist: json = unknown"
}
```

Bug is present in 0.2.12 and 0.3 candidate. Url encoding the json does not help.

ruslantalpa:
This is a combination of a few issue, basically what is happening here is this
```
select * from "1".complex_items where settings = '{"a":1}'
```
and it fails
but if you do
```
select * from "1".complex_items where settings::jsonb = '{"a":1}'
```
it works.

I would say it works as expected with a few mentions:
The users can check fore equality for any part inside the json with ->> and it works but not the whole json column.
Whenever we return a location header where there are no keys but there is a json column, the url will not work.
I would clasiffy this as low priority since there are workarounds (use jsonb instead of json, have keys on table, filter based on data within json column)

begriffs:
Sort of low priority, I agree. Although I noticed it when testing the urls appearing in the Location header for newly posted json data. They are of this form and are broken.

Interestingly when I altered one of the tables to use an actual jsonb column rather than doing a typecast in the query it still fails.

ruslantalpa:
the location header issue was addressed and since the root of the problem here comes from the fact that the data type does not support equality, i would say this is not a bug, it's the same as saying `select 'a'+'b'` not being supported is a bug. ... close :)

begriffs:
For good measure we should update [this test](https://github.com/begriffs/postgrest/blob/master/test/Feature/InsertSpec.hs#L168-L178) to not only insert a jsonb object but attempt to GET the url provided in the resulting location header and check it succeeds.

steve-chavez:
I think this issue can be closed by:

- Adding a test to ensure that a `jsonb` column can be filtered by equality.
- Enable the `Location` header only for tables with a PK. If the table has no PK the `Location` will be misleading, also don't think `json/jsonb` columns are good candidates for PK so I'll add a test with a different data type for this.

@begriffs Let me know what you think.

steve-chavez:
Closing this since #1128 got merged, it included a test for jsonb equality in https://github.com/PostgREST/postgrest/pull/1128/commits/610ac4b74340badfb95905c3294104936ebdd11d.

PR #1128, opened 2018-06-08
Add support for getting json/jsonb by array index

### LINKED ISSUE #1099: Add array index support to jsonb
### Environment

* Version of postgrest: v0.4.4.0
* Operating system: RHEL 7.4
* Running in docker:
  * [ ] yes
  * [x] no

### Description of issue

I have this output where `ret` is of type `jsonb`:

http://nixpaste.lbr.uno/BJw5QeGX?json

In the example, I would like to be able to filter the output from all `ret` records that have a 'fun' equal to `
saltutils.find_job`.

### Support level

In addition to free community support, are you interested in priority consulting time from [begriffs](https://begriffs.com/)?

* [ ] yes
* [x] no

steve-chavez:
Closes #1099, json/jsonb array indexes can now be specified as:
```http
GET /json_arr?select=data->2->key->>1 
```
Negative indexes(only for pg >= 9.5), e.g. `select=data->-4`, also work as per the PostgreSQL [docs](https://www.postgresql.org/docs/9.5/static/functions-json.html).

Only thing missing for fully compliant `->/->>` operators would be to allow finishing a json path with `->` and not restrict it to finish with `->>`. As it is now, for obtaining json the user needs to cast like `data->>key::json`.

The docs used to say: `The final arrow must be the double kind, ->>, or else PostgREST will not attempt to look inside the JSON`, but not sure why it was done this way, @ruslantalpa maybe you can shed some light into this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - #1099, Numbers in json path `?select=data->1->>key` now get treated as json array indexes instead of keys - @steve-chavez
+- #1128, Allow finishing a json path with a single arrow `->`. Now a json can be obtained without resorting to casting, Previously: `/json_arr?select=data->>2::json`, now: `/json_arr?select=data->2` - @steve-chavez
 
 ## [0.5.0.0] - 2018-05-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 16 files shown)


steve-chavez:
I'll consider the `->>` finishing as a limitation of the feature at the time it was added, don't see a compelling reason to keep it. 

I think it'll be better to parse the operators and let pg do the validation, right now a user can do `/json?select=data->>a->b` and the `->b` part will be ignored. However when doing `select data->>'a'->'b'` on pg it'll fail with `operator does not exist: text -> unknown`, this seems preferable for most cases.

steve-chavez:
Also added a test for https://github.com/PostgREST/postgrest/issues/336, that can be closed when this is merged.

PR #1148, opened 2018-07-23
OpenAPI: add required section for the non-nullable columns

laughedelic:
This PR adds the `definitions.{definition}.required` list with the names of the columns that have `colNullable` set to `False`.

For example, if I have a table

```sql
create table organisms (
  name   text primary key,
  tax_id text not null
);
```

Then the corresponding definition in the generated OpenAPI specification will be

```json
  "definitions": {
    "organisms": {
      "required": [
        "name",
        "tax_id"
      ],
      "properties": {
        "name": {
          "format": "text",
          "type": "string",
          "description": "Note:\nThis is a Primary Key.<pk/>"
        },
        "tax_id": {
          "format": "text",
          "type": "string"
        }
      },
      "type": "object"
    }
  }
```

Notice the `required` list: `name` column is the primary key, so it's not null and `tax_id` has an explicit `not null` annotation.

Motivation: I'm using the Swagger/OpenAPI specification provided by PostgREST and [swagger-codegen](https://github.com/swagger-api/swagger-codegen#prerequisites) to generate client-side code (Elm), so if the `required` list is omitted, all fields are marked as optional (`Maybe ...`) which is very inconvenient, taking into account that the DB (and PostgREST) knows which columns are non-optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1099, Add support for getting json/jsonb by array index - @steve-chavez
 - #1145, Add materialized view columns to OpenAPI output - @steve-chavez
 - #709, Allow embedding on views with subselects/CTE - @steve-chavez
+- #1148, OpenAPI: add `required` section for the non-nullable columns - @laughedelic
 
 ### Fixed
 
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -42,7 +42,8 @@ makeTableDef pks (t, cs, _) =
       (tn, (mempty :: Schema)
         & description .~ tableDescription t
         & type_ .~ SwaggerObject
-        & properties .~ fromList (map (makeProperty pks) cs))
+        & properties .~ fromList (map (makeProperty pks) cs)
+        & required .~ map colName (filter (not . colNullable) cs))
 
 makeProperty :: [PrimaryKey] -> Column -> (Text, Referenced Schema)
 makeProperty pks c = (colName c, Inline s)
--- a/test/Feature/StructureSpec.hs
+++ b/test/Feature/StructureSpec.hs
@@ -113,7 +113,10 @@ spec = do
                     "format": "integer",
                     "type": "integer"
                   }
-                }
+                },
+                "required": [
+                  "id"
+                ]
               }
             |]
 


steve-chavez:
Thank you @laughedelic. Add an entry to the "Added" section in the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md) and then I'll merge.

laughedelic:
Done 👌

laughedelic:
Hi @ruslantalpa. You're right, I didn't take defaults into account. I tried to use them and discovered #1150.

Apart from that I think that there are two different perspectives on it:

* When you are _reading_ some rows with columns that have (non-null) defaults, it means that you're always getting a value of the declared type, no point in wrapping it in `Maybe`. This corresponds to the `#/definitions/foo` part of the spec which is returned in the `GET` response and my PR was focused on that. But another thing that I didn't take into account and I think you're talking about is that

* When you are _writing_, you don't need to submit the whole row, only those fields that are non-null and don't have defaults. This corresponds to the `#/parameters/body.foo` which is used in `POST` and `PATCH` requests. Currently its schema just refers to `#/definitions/foo` and I think it's wrong, because it's actually two different things.

What do you think about it?

ruslantalpa:
@laughedelic if this PR only refers to the GET part and not the POST then it's 100% correct

laughedelic:
@ruslantalpa the intention was to improve the GET part, but unfortunately this change also affects the POST part, because it reuses the same definition: `#/parameters/body.foo` refers to (the changed) `#/definitions/foo` as its schema.

I have an idea how to fix it though. _If_ OpenAPI v2 supports `allOf`, it could be done like this:
* remove `required` from the `#/definitions/foo` (i.e. as it was before)
* add it to the `#/paths/foo/get/responses/200` schema like this:
    ```json
    {
      "type": "array",
      "items": {
        "allOf": [
          { "$ref": "#/definitions/foo" },
          { "required": [ 
              "a", 
              "b", 
              "... non-nullable fields of foo" 
            ] 
          }
        ]
      }
    }
    ```

While this maybe pushes the problem on the side of the code generators, it's a more correct OpenAPI specification.

ruslantalpa:
@laughedelic i am not familiar with openapi spec (nor very interested in it :) ), i was just pointing out something i thought you missed. Whatever @steve-chavez and you decide to do is ok.

steve-chavez:
@laughedelic Your proposal looks good to me, also seems OpenAPI v2 does support `allOf` see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism.

PR #1149, opened 2018-07-23
OpenAPI: Change GET response type to array

laughedelic:
The [docs](http://postgrest.org/en/v5.0/api.html#singular-or-plural) say:

> By default PostgREST returns all JSON results in an array, even when there is only one item. For example, requesting /items?id=eq.1 returns
> 
> ```
> [
>   { "id": 1 }
> ]
> ```

But in the generated OpenAPI specification `GET` returns a single item:

```json
"get": {
  "responses": {
    "200": {
      "schema": {
        "$ref": "#/definitions/foo"
      },
    },
  }
}
```

After this PR, it would generate 

```json
"get": {
  "responses": {
    "200": {
      "schema": {
        "items": {
          "$ref": "#/definitions/cartridges"
        },
        "type": "array"
      },
    },
  }
}
```


steve-chavez:
Would be better to have a test for this, in https://github.com/PostgREST/postgrest/blob/master/test/Feature/StructureSpec.hs#L39, could be something like `it "includes an array type for GET responses"` and then validate that using `child_entities` and `grancdhild_entities`.

laughedelic:
@steve-chavez is 6891542 what you meant? I'm having trouble setting up the test DB locally and I haven't written any spec tests in Haskell before, so I don't really know what I'm doing 😅

steve-chavez:
@laughedelic Looks good, I think that'll do. Also, to run the tests locally I use:
```bash
test/destroy_test_db "postgres://postgres@localhost" postgrest_test
export POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://postgres@localhost" postgrest_test)
stack test
```
This is mentioned in the [docs](https://postgrest.org/en/v5.0/install.html#running-the-tests) as well.

laughedelic:
@steve-chavez Thanks. I read the docs, but couldn't create a test DB. I guess my problem is related to some general Postgres setup: when I try to create the test DB, I get

```
ERROR: Please specify the user with 'Create DB' permissions, and ensure that the default database for the username exists.
```

steve-chavez:
Maybe you can fix that with:

```bash
sudo -u postgres psql -c 'ALTER ROLE "<your-user>" CREATEDB'  
sudo -u postgres psql -c 'CREATE DATABASE "<your-user>"'
```

laughedelic:
It didn't help, so I just reinstalled Postgres from another source and now it works and I can run tests 👌 

Thanks for support!

PR #1153, opened 2018-07-26
Quote identifier names in RPC function arguments

mdr1384:
Your proposed change did not do it, there was another spot that also needed pgFmtIdent()


steve-chavez:
Thanks a lot @mdr1384, it would be good to have a regression test to prevent this kind of errors.

A test for this function might be enough:
```sql
create or replace function "quotedFunction"("user" text, "fullName" text, "SSN" text) 
returns json AS $$
  select format('{ "user": "%s", "fullName": "%s", "SSN": "%s"}', "user", "fullName", "SSN")::json;
$$ language sql;
```

PR #1157, opened 2018-07-31
Update LICENSE

Deryew:
bump copyright year

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014 Joe Nelson
+Copyright (c) 2018 Joe Nelson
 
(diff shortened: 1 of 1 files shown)


steve-chavez:
Appreciate you wanting to contribute to the project, but I don't think this is right, from https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated:

**The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work.**

Maybe we should keep the year as it is.

Issue #899, opened 2017-06-28
v0.4.2.0 does not run on CentOS 7


monemihir:
Following error is received:
```
[mihir@vmcentos7 Downloads]$ ./postgrest test.conf 
Listening on port 30000
Attempting to connect to the database...
postgrest: getAddrInfo: does not exist (Name or service not known)
[mihir@vmcentos7 Downloads]$ 
```

This is my conf file
```
db-uri = "postgres://testuser:test123@localhost:5432/postgrest_test_db"
db-schema = "public"
db-anon-role = "postgres"
db-pool = 100

server-host = "localhost"
server-port = 30000
```
I am running on CentOS 7
```
[mihir@vmcentos7 Downloads]$ hostnamectl
   Static hostname: vmcentos7
         Icon name: computer-vm
           Chassis: vm
        Machine ID: 27ac7b066d974238a1fffac65ae7f0c8
           Boot ID: 97f051ea7d46425ca801eff22b4379e4
    Virtualization: vmware
  Operating System: CentOS Linux 7 (Core)
       CPE OS Name: cpe:/o:centos:centos:7
            Kernel: Linux 3.10.0-514.el7.x86_64
      Architecture: x86-64
[mihir@vmcentos7 Downloads]$ 
```
PostgreSql version is 9.5. I can connect to the Postgres server using **psql** as well as **pgadmin3**.

joehealy:
Not sure I understand the resolution completely, but we were able to make it work using "*4" rather than localhost as the server-host parameter.

I would have thought that localhost would have been treated as a hostname as I believe server-host is a HostPreference (https://www.stackage.org/haddock/lts-8.5/warp-3.2.11.1/Network-Wai-Handler-Warp.html#t:HostPreference)

begriffs:
It appears that `localhost` is traditionally set in `/etc/hosts`, and perhaps it is missing from that file on your system.

gitfirenz:
I'm on centos7  (digitalocean) and it works too with listen "*4"
I add issues however with postgrest using IP V6 to connect to postgresql (and so have to fix pg_hba).

begriffs:
Can you explain what you had to do so I can document it?

begriffs:
I'm not sure from that Haskell library doc whether `*4` will be accessible to the outside network if your firewall allows the port through. For safety's sake wondering if 127.0.0.1 works on centos. Can you check?

monemihir:
Yes, 127.0.0.1 works.

[...]

PR #1165, opened 2018-08-09
Default to 127.0.0.1 instead of *4, add SIGUSR1 as alias for SIGHUP

### LINKED ISSUE #724: 500 Error with node
In safari, chrome and postman works ok, but node http client works fail.

MacOS Sierra 10.12
PostgREST 0.3.2.0
node v6.9.1

``` sh
postgrest postgres://localhost:5432/mydb -a myaccount -m 12
```

``` js
// Request: 'GET / HTTP/1.1\r\nHost: localhost:3000\r\nConnection: close\r\n\r\n'
// Response: 'Something went wrongHTTP/1.0 500 Internal Server Error\r\nDate: Sun, 23 Oct 2016 05:17:37 GMT\r\nServer: postgrest/0.3.2.0\r\nContent-Type: text/plain; charset=utf-8\r\n\r\nSomething went wrong'
```

``` js
http.get('http://localhost:3000/', res=> console.log(res.statusCode))

fetch = require('isomorphic-fetch')
fetch('http://localhost:3000/').then(res => console.log(res.status)).catch(err => console.log(err))

postgrest = require('postgrest-request')
postgrest('http://localhost:3000').get('/').end((err,data) => console.log(data))

request = require('request')
request('http://localhost:3000/', (err, res, body) => console.log(body))
```

steve-chavez:
Fixes #724.

The default of *4 was added because `localhost` was not in `/etc/hosts` on CentOS(this was #899). Don't know why *4 was preferred to `127.0.0.1`, but the latter would be better to prevent errors as the one in #724. 

Using SIGHUP as reload signal causes pgrst to be in a broken state when closing the terminal window or when running pgrst on a ssh session and this session drops. These errors could be prevented by using a `nohup` or having pgrst as daemon, but in development one is usually starting/stopping pgrst manually.

The idea is to deprecate SIGHUP(remove it in v6) and introduce SIGUSR1 as it's replacement.

Any thoughts on why this would be a bad idea? Seems like the simpler fix for me(leaving SIGHUP with it's default behavior).


docker/Dockerfile
```diff
@@ -28,7 +28,7 @@ ENV PGRST_DB_URI= \
     PGRST_DB_SCHEMA=public \
     PGRST_DB_ANON_ROLE= \
     PGRST_DB_POOL=100 \
-    PGRST_SERVER_HOST=*4 \
+    PGRST_SERVER_HOST="127.0.0.1" \
```
ruslantalpa: this is not a good default for docker (it should stay *4). This will mean if you start postgrest container it will listen only internally and you will not be able to actually work with it.

steve-chavez: I overlooked that, though I believe that's only true for macos/windows that run the docker host in a vm, in linux it works fine. I'll make the change.

ruslantalpa: i don't think it's about the vm. have you actually built the image? (i don't think you could since the build process downloads the binary from the release). 
regardless of the os, in docker each container gets assigned an external ip/interface, that is the interface for which docker bridges the external/internal port, this (127...) will not work on any os (but yes, it failed me on mac)

steve-chavez: It can be tested by overriding the `PGRST_SERVER_HOST`(not necessary to build the image), on linux it works with `127.0.0.1`.

steve-chavez: Used this https://gist.github.com/steve-chavez/77d5ef4e8b0a51120bc020b96d6553f5.

ruslantalpa: that gist works only because you have `network_mode: host` there, which is not the standard way of running docker containers, if you remove that line you'll see the problem

PR #1197, opened 2018-10-21
Config: apply CorsResourcePolicy to PUT requests

bkylerussell:
PUT requests are part of the public upsert API, so if postgrest
runs at a different origin, these requests will fail because they
don't pass CORS.


steve-chavez:
This slipped me by when adding PUT. 

Thanks a lot!

PR #1198, opened 2018-10-25
Reference secrets via files when using Docker

msalvadores:
While working with postgrest and docker I prefer to manage secrets via files instead of env vars because it plays better with docker secrets. 

I have added a `start` script that allows to optionally set these via files that can be then managed with docker secrets. This allows me to have an environment file like ...

```env
PGRST_DB_SCHEMA=app
PGRST_DB_ANON_ROLE=none
PGRST_SERVER_PORT=3000
PGRST_JWT_SECRET_FILE=/run/secrets/jwt_api_secret
PGRST_DB_URI_FILE=/run/secrets/pgrest_db_uri
```
That then is used in compose with ...

```yaml
version: "3.3"
services:
  pgrest:
    image: myregistry.azurecr.io/postgrest:latest
    ports:
      - "30009:3000"
    depends_on:
      - postgres
    env_file:
      - envs/pgrest.env
    secrets:
      - jwt_api_secret
      - pgrest_db_uri
    networks:
      - app_network
    deploy:
      replicas: 1

secrets:
  jwt_api_secret:
    file: ./secrets/jwt_api.scrt
  pgrest_db_uri:
    file: ./secrets/pgrest_db_uri.scrt
```

I figured that these changes could be of benefit to others. If there are other workarounds to what I was trying to achieve with this please let me know. Thanks.


docker/Dockerfile
```diff
@@ -21,25 +21,22 @@ RUN BUILD_DEPS="curl ca-certificates xz-utils" && \
     apt-get -qq clean && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-COPY postgrest.conf /etc/postgrest.conf
-
+COPY start .
```
steve-chavez: Usually this file is called `entrypoint.sh` in Dockerfiles, how about if you change it and then have this line as:
```Dockerfile
COPY entrypoint.sh /entrypoint.sh
```

docker/Dockerfile
```diff
@@ -21,25 +21,22 @@ RUN BUILD_DEPS="curl ca-certificates xz-utils" && \
...
     PGRST_MAX_ROWS= \
     PGRST_PRE_REQUEST= \
     PGRST_ROLE_CLAIM_KEY=".role"
 
-# PostgREST reads /etc/postgrest.conf so map the configuration
-# file in when you run this container
-CMD exec postgrest /etc/postgrest.conf
-
-EXPOSE 3000
+CMD ./start
```
steve-chavez: Then have this line as:
```Dockerfile
ENTRYPOINT ["/entrypoint.sh"]
``` 

Also, not sure why but with `CMD` I cannot quit the container with `Ctrl + C` with `ENTRYPOINT` I can.

docker/start
```diff
@@ -0,0 +1,36 @@
...
+
+if [[ $PGRST_JWT_SECRET_FILE != "" ]]; then 
+    JWT_SECRET=$(<$PGRST_JWT_SECRET_FILE); 
+else
+    JWT_SECRET=${PGRST_JWT_SECRET};
+fi
+
+CONF=/etc/postgrest-secrets.conf
+touch $CONF
+cat > $CONF <<-EOS
```
steve-chavez: You could replace this 3 lines with:
```bash
cat <<EOF > postgrest.conf
```

Also let's keep the name as `postgrest.conf`, not all the config values are secrets.

steve-chavez:
Thank you @msalvadores. See my comments above, after those corrections add an entry to the CHANGELOG and then I'll merge.

ruslantalpa:
This direction is a bad idea, the entire config can be thought of as a secret and mounted in place. Secrets config i think has a setting to tell it where to mount

ruslantalpa:
I.e. there is no need to change the image

msalvadores:
Thanks for the comments.

@ruslantalpa re.
> the entire config can be thought of as a secret and mounted in place. 

Interesting take. I had not thought of that. 

The solution proposed supports both having independent secret management and, as you suggest, put everything in one secret. I'd rather manage secrets independently because I can track their provenance better and allows me to change the keys in one single place.

@steve-chavez I'll address your comments. Thanks for the suggestions.

ruslantalpa:
as currently implemented it does not, you are overwriting the config each time from the entrypoint so someone simply mapping using -v the config direcly into the image will break (it might probably break the docker image for a lot of ppl relying on the current method)

the jwt-secret param already supports reading from a file so all the code related to it is unneeded.
https://github.com/PostgREST/postgrest/blob/master/test/io-tests/configs/secret-from-file.config

the correct way to implement this is to have the same functionality implemented for db-uri param and i skimmed over the code and this functionality i think comes from the `conf` module so it might very wall be that you can read the db uri string from a file by specifying the path and prepanding `@` like in the file i linked

the entrypoint.sh method is unneded complexity for the default/official image (now besides maintaining the postgrest.conf file, the entrypoint.sh needs maintainace)

msalvadores:
@ruslantalpa 

> the jwt-secret param already supports reading from a file

I was unaware of this.

> the correct way to implement this is to have the same functionality implemented for db-uri param ...

That does sound like the best solution. I second that as a feature request ;) ... not sure I can get into haskell level modifications.

msalvadores:
Closing this and filing a feature request here #1203

PR #1205, opened 2018-10-31
Add support for parsing JSON Web Key Sets

### LINKED ISSUE #505: Accept more than one JWT secret for authentication

russelldavies:
This should resolve #505.

The configuration option `jwt-secret` is a bit overloaded now as it's doing three things so perhaps there should be thought about whether one option works (also perhaps the name needs to be updated).

I'll update the documentation once this feature has been reviewed and accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
+- #1205, Add support for parsing JSON Web Key Sets -@russelldavies
 
 ### Fixed
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -35,7 +35,7 @@ import           PostgREST.ApiRequest   ( ApiRequest(..), ContentType(..)
                                         , mutuallyAgreeable
                                         , userApiRequest
                                         )
-import           PostgREST.Auth            (jwtClaims, containsRole, parseJWK)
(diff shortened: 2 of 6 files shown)


steve-chavez:
@russelldavies Thanks for implementing this long wanted feature!(Sorry for the late review btw)

The code looks good and the comment is really helpful :+1: . 

As you mention the `jwt-secret` option is overloaded but I don't know if I can come up with a better name(`jwk-or-secret`? do you have any suggestions?), maybe we can just make this clear in the docs.

Also it would be good to have this clarification on the CLI when invoking `postgrest` with no config, in this part [here](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/Config.hs#L212-L214), and don't forget to add an entry in the CHANGELOG :).

russelldavies:
@steve-chavez In regard to `jwt-secret` I think just `auth`.

steve-chavez:
`auth` seems a bit broad/generic, for now I think we can just clarify `jwt-secret` usage in the docs.

Thanks a lot for this feature :). I'll merge it now.

russelldavies:
@steve-chavez Thanks, I'll update the docs accordingly.

PR #1218, opened 2018-12-07
Add db-extra-search-path config

### LINKED ISSUE #1200: Explicit qualification issues, add extra config param for extensions schema
We had some issues with [Explicit qualification](http://postgrest.org/en/v5.1/api.html#explicit-qualification) lately: 

- PostGIS functions/types calls https://github.com/PostgREST/postgrest/issues/1125 (solvable by adding search_path)
- ltree operators https://github.com/PostgREST/postgrest/issues/1183 (solvable by adding extension to `pg_catalog` or wrapping access in a function)
- [Heroku postgres issue](https://help.heroku.com/OZ4RPNQS/how-do-i-resolve-function-get_xmlbinary-does-not-exist-errors-in-heroku-connect) reported on gitter (no option but to downgrade to v0.4.4.0)

Now, let's assume https://github.com/PostgREST/postgrest/issues/1106 is implemented. Users would do `db-schema = public, api` and this would solve the above issues without workarounds. But this would also have the unwanted side effect of exposing all the extensions(postgis, ltree, etc) internals. So I think that feature is not the solution, unless later we add a config for hiding a schema but this seems too complicated.

So, how about if instead we add an extra config param(say `db-ext-schema`) for an additional extension schema that only gets added to the `search_path` that PostgREST sets but doesn't get exposed?(we keep qualifying all db objects with the `db-schema`, like `api.table`, so is not possible to access `public.internal` through `GET /internal`)

This `db-ext-schema` would be `public` by default so most workflows will continue to work as usual.

Thoughts? Any feedback would be appreciated.

(Also considered just dropping the `set schema` change and not mess with the `search_path` but then we would have https://github.com/PostgREST/postgrest/issues/828 again)

steve-chavez:
Fixes #1200. The config value has a list format: 
```rust
db-extra-search-path = ["public", "extensions"]
```
By default the config is `["public"]` .

Is the list format appropriate? Or perhaps I should change it to maintain the format used by pg `search_path`, like `db-extra-search-path = "public, extensions"`? I just reused `configurator-ng` functionality for now, but I guess I could parse the value and make it comma separated.

**Edit**: Now the config value has the following format
```rust
db-extra-search-path = "public, extensions"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
-- #1205, Add support for parsing JSON Web Key Sets -@russelldavies
+
+- #1205, Add support for parsing JSON Web Key Sets - @russelldavies
 - #1203, Add support for reading db-uri from a separate file - @zhoufeng1989
+- #1200, Add db-extra-search-path config for adding schemas to the search_path, solves issues related to extensions created on the public schema - @steve-chavez
 
 ### Fixed
 
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -124,6 +124,7 @@ Test-Suite spec
                      , Feature.ConcurrentSpec
                      , Feature.CorsSpec
                      , Feature.DeleteSpec
+                     , Feature.ExtraSearchPathSpec
                      , Feature.InsertSpec
                      , Feature.JsonOperatorSpec
                      , Feature.NoJwtSpec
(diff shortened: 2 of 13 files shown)


ruslantalpa:
Why the parsing, why not a simple string. Do you get any benefit from knowing individual schemas?

steve-chavez:
Come to think of it, shouldn't we validate the config value? Should we worry about injecting sql code in the config values?

ruslantalpa:
Once an attacker can modify a config file (etc dir) it’s game over. Validating to guard against “stupid mistakes” however might make sense but i guess it’s just more (defensive) code with limited benefits (a misconfiguration here will manifest itself very quickly)

ruslantalpa:
But if it’s just a line for a “parser” expression (chars, numbers, spaces) then i see no harm in having a very basic “validation”

steve-chavez:
Thanks for the feedback Ruslan. I added a simple comma splitting and quoting of the values, I think it's good enough.

Issue #1228, opened 2019-01-15
db-extra-search-path configuration parameter is missing from postgrest.conf of dockerized postgREST v5.2


vpapako:
### Environment

* PostgreSQL version: docker: postgres:9.6-alpine
* PostgREST version: docker: 5.2

### Description of issue

The configuration parameter of db-extra-search-path is missing from `/etc/postgrest.conf` of dockerized postgREST v5.2

##### (Assuming that the environment variable PGRST_DB_EXTRA_SEARCH_PATH has already been set when started the container)

1. Create a new Bash session in the container of postgREST `docker exec -it <CONTAINER> /bin/bash`
2. Add the config parameter in postgres.conf 
`echo 'db-extra-search-path = "$(PGRST_DB_EXTRA_SEARCH_PATH)"' >> /etc/postgrest.conf`
3. Restart the postgREST container

steve-chavez:
Closing since #1229 got merged.

PR #1229, opened 2019-01-15
Add db-extra-search-path config param to dockerized postgREST

vpapako:
This PR is expected to solve issue #1228


steve-chavez:
@vpapako Thank you :smiley:! Now I'm going to push a new docker image with the fix.

Issue #1109, opened 2018-05-15
GHC 8.4.2 build failure


ilovezfs:
### Environment

* Version of postgrest: 0.5.0.0
* Operating system: macOS
* Running in docker:
  * [ ] yes
  * [x] no

### Description of issue

```
Failed to install Ranged-sets-0.3.0
Build log ( /private/tmp/postgrest-20180514-7152-1j56srv/postgrest-0.5.0.0/.cabal-sandbox/logs/ghc-8.4.2/Ranged-sets-0.3.0-H49IfYeWOqP168H0mI98jo.log ):
cabal: Entering directory '/tmp/cabal-tmp-7430/Ranged-sets-0.3.0'
Configuring Ranged-sets-0.3.0...
clang: warning: -Wl,-headerpad_max_install_names: 'linker' input unused
clang: warning: argument unused during compilation: '-L/usr/local/opt/openssl/lib'
clang: warning: argument unused during compilation: '-L/usr/local/opt/readline/lib'
clang: warning: argument unused during compilation: '-L/usr/local/lib'
clang: warning: argument unused during compilation: '-L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries'
Preprocessing library for Ranged-sets-0.3.0..
Building library for Ranged-sets-0.3.0..
[1 of 4] Compiling Data.Ranged.Boundaries ( Data/Ranged/Boundaries.hs, dist/dist-sandbox-b7a9e73c/build/Data/Ranged/Boundaries.o )
[2 of 4] Compiling Data.Ranged.Ranges ( Data/Ranged/Ranges.hs, dist/dist-sandbox-b7a9e73c/build/Data/Ranged/Ranges.o )
[3 of 4] Compiling Data.Ranged.RangedSet ( Data/Ranged/RangedSet.hs, dist/dist-sandbox-b7a9e73c/build/Data/Ranged/RangedSet.o )

Data/Ranged/RangedSet.hs:75:10: error:
    • Could not deduce (Semigroup (RSet a))
        arising from the superclasses of an instance declaration
      from the context: DiscreteOrdered a
        bound by the instance declaration
        at Data/Ranged/RangedSet.hs:75:10-45
    • In the instance declaration for ‘Monoid (RSet a)’
   |
75 | instance DiscreteOrdered a => Monoid (RSet a) where
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1237, opened 2019-02-03
Relax some version bounds

felixonmars:
The build and tests went smoothly here with the updated dependencies (base-4.12.0.0, protolude-0.2.3, hjsonschema-1.9.0).

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -32,12 +32,12 @@ executable postgrest
     "-with-rtsopts=-N -I2"
   default-language:    Haskell2010
   build-depends:       auto-update
-                     , base >= 4.8 && < 4.10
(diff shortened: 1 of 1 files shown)


steve-chavez:
`protolude` is probably fine, but upgrading `base` and `hjsonschema` makes the build fail when not using stack(since it pins versions in LTS, cabal will not work).

See https://github.com/PostgREST/postgrest/issues/934#issuecomment-417323827 and #1109. Also https://wiki.haskell.org/Base_package (we restrict `base` versions based on ghc compatibility).

robx:
#1319 is an updated version of this.

steve-chavez:
Bounds are updated since #1319 got merged.

PR #1241, opened 2019-02-13
Faster bulk insert with ?columns query arg

### LINKED ISSUE #690: Large memory usage for bulk inserts

### LINKED ISSUE #1234: Ignore extra information on JSON
### Environment

* PostgreSQL version: 11.1
* PostgREST version: 5.1
* Operating system: Debian 8 and MacOS 10.14

### Description of issue

I'm not sure if this has been raised before, I tried my best to find out but couldn't find anything.

I wonder if there is a header that tells PostgREST to ignore any extra properties in a json that doesn't match the table columns, and continue the INSERT or UPDATE with the matching ones.

PostgREST usual those and returns an error as its default behaviour, that is the correct behaviour, but in some cases, we may need to use the matching properties, ignore the rest, and continue executing the INSERT or UPDATE.

For example, in my current project I have a table similar to this:
```sql
create table datasets(
  id serial primary key,
  source text,
  publication_date timestamp,
  figure bigint
)
```

I allow data from multiple sources to be stored in this table, each source may have some extra fields that I don't want, for example:
```json
{
  "source": "Natural Disaster Prevention and Control (NDPC)",
  "publication_date": "...",
  "figure": 1100,
  "location": "...",
  "comment": "...",
  "lat": "...",
  "lon": "...",
}
```

Usually, if it was few records, I would clean up the data and reformat it, but in my case I have thousands of records coming from hundred of different sources in different formats.

Thanks in advance.

steve-chavez:
Fixes #690. Also fixes #1234.

As proposed in https://github.com/PostgREST/postgrest/issues/690#issuecomment-378363946, now a `columns` query parameter can be specified to delegate all the json processing to PostgreSQL when doing POSTs(also RPC and PATCH). This is analogous to how we've been doing GET until now.

This new mechanism makes our perf test of bulk inserting 100000 json array values improve on memory consumption, from 1.15G to 20M, around 1% of what it used to be.

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -143,9 +143,9 @@ createWriteStatement selectQuery mutateQuery wantSingle isInsert asCsv rep pKeys
 
 type ProcResults = (Maybe Int64, Int64, ByteString, ByteString)
 callProc :: QualifiedIdentifier -> [PgArg] -> Bool -> SqlQuery -> SqlQuery -> Bool ->
-            Bool -> Bool -> Bool -> Bool -> Maybe FieldName -> Bool -> PgVersion ->
+            Bool -> Bool -> Bool -> Bool -> Maybe FieldName -> PgVersion ->
             H.Statement ByteString (Maybe ProcResults)
-callProc qi pgArgs returnsScalar selectQuery countQuery countTotal isSingle paramsAsSingleObject asCsv asBinary binaryField isObject pgVer =
+callProc qi pgArgs returnsScalar selectQuery countQuery countTotal isSingle paramsAsSingleObject asCsv asBinary binaryField pgVer =
   unicodeStatement sql (HE.param HE.unknown) decodeProc True
   where
     sql =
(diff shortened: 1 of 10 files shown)


src/PostgREST/QueryBuilder.hs
```diff
@@ -322,6 +321,22 @@ requestToQuery schema _ (DbMutate (Delete mainTbl logicForest returnings)) =
...
+-- These CTEs convert a json object into a json array, this way we can use json_populate_recordset for all json payloads
+-- Otherwise we'd have to use json_populate_record for json objects and json_populate_recordset for json arrays
+-- We do this in SQL to avoid processing the JSON in application code
+normalizedBody :: SqlFragment
+normalizedBody =
+  unwords [
+    "pgrst_payload AS (SELECT $1::json AS json_data),",
+    "pgrst_body AS (",
+      "SELECT json_data AS val FROM pgrst_payload WHERE json_typeof(json_data) = 'array'",
+      "UNION ALL",
```
ruslantalpa: could this be written in form of `CASE`?, it would be easier to understand.

test/Feature/InsertSpec.hs
```diff
@@ -427,21 +480,21 @@ spec = do
             matchHeaders = ["Content-Range" <:> "*/*"]
           }
 
-        get "/items" `shouldRespondWith`
```
ruslantalpa: were these tests failing after the change or is it just cleanup since it's a GET in a section about POST?

ruslantalpa:
Have you tried to see the impact this thing has on the database (although there is a small chance this will have a big negative impact since it only adds a "check the type of payload" logic into sql)

src/PostgREST/QueryBuilder.hs
ruslantalpa: another benefit would be that `json_type` would be called only once on this big payload

test/Feature/InsertSpec.hs
steve-chavez: It's just a cleanup. Didn't consider this test relevant and I would have had to update it with more `{"id": ..}` to keep it working.

src/PostgREST/QueryBuilder.hs
steve-chavez: CTE is much simpler now, thanks.

steve-chavez:
@ruslantalpa I haven't measured the impact but as you mention is only adding one json_typeof call and I think that would be negligible.

What do you think about removing the haskell json processing and making the POSTs be for all table columns by default? I've seen a busy system with lots of inserts(not bulk inserts, just single inserts with payloads with many columns) showing considerable memory consumption(around 500mb).

It'd be a breaking change for another version. I think we shouldn't be processing the json and validating that all columns are the same when the db can do a better job enforcing no columns are missing with not null constraints.

steve-chavez:
I meant make the breaking change in a future PR.

steve-chavez:
There's still some unnecessary copying, I just did a test with:
```
yes "{\"time\": \"`date +'%F %T-05'`\"}" | head -n 1000000 | jo -a | curl -H "Content-Type: application/json" -d @- "http://localhost:3000/conditions?columns=time"
```
And the resulting .prof is:
```
	Sat Feb 16 12:10 2019 Time and Allocation Profiling Report  (Final)

	   postgrest +RTS -N -I2 -p -h -RTS support/configs/config

	total time  =        0.06 secs   (61 ticks @ 1000 us, 1 processor)
	total alloc = 112,343,744 bytes  (excludes profiling overheads)

COST CENTRE                 MODULE                                 SRC                                                        %time %alloc

app                         PostgREST.App                          src/PostgREST/App.hs:(119,1)-(337,132)                      19.7   30.5
receiveloop                 Network.Wai.Handler.Warp.Recv          Network/Wai/Handler/Warp/Recv.hs:(115,1)-(129,20)           19.7    0.0
sendQueryPrepared.\.\       Database.PostgreSQL.LibPQ              src/Database/PostgreSQL/LibPQ.hsc:(1736,16)-(1740,69)       18.0   30.3
nullableValue.\.encoder     Hasql.Private.Encoders.Params          library/Hasql/Private/Encoders/Params.hs:(29,5)-(30,47)     18.0   30.3
receive                     Network.Wai.Handler.Warp.Recv          Network/Wai/Handler/Warp/Recv.hs:(96,1)-(99,48)              3.3    0.7
MAIN                        MAIN                                   <built-in>                                                   1.6    0.1
```

Don't know why libpq and hasql do allocations. I wonder if zero-copy would be possible :thinking:.

steve-chavez:
Maybe we can get zero-copy when [linear types](https://ghc.haskell.org/trac/ghc/wiki/LinearTypes) get into ghc. For now I think this is good to merge.

PR #1255, opened 2019-03-13
Add Healthcheck

anton-yurchenko:
Shall we add at least basic healthcheck? 🙃

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,6 +2,9 @@ FROM debian:jessie
 
 ARG POSTGREST_VERSION
 
+HEALTHCHECK --interval=30s --timeout=30s --start-period=10s --retries=3 \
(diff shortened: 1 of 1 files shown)


ruslantalpa:
this accomplishes nothing really, when replying to `/` all the info comes from memory, db is not accessed so what this tests is only that postgrest executable running and not that it can communicate with the db.  IF it would not reply to that url then it would mean the exe/container crashed and by default it would be restarted (in docker-compose, when you have restart:always). Crashed postgrest == crashed container => container is restarted.

Also, when the connection to the db goes down, postgrest periodically tries to reconnect. This means that even you would have a url that makes a call to the db and you detect that postgrest can not connect to the database (assuming the connection info is correct), it is still not the best approach to restart the container since postgrest tries to reconnect to the db by itself.

anton-yurchenko:
@ruslantalpa, thanks for the exaplanation. (closing the PR)

PR #1256, opened 2019-03-14
Allows resource embedding on materialized views

vitorbaptista:
Currently, it's only possible to do resource embedding on non-materialized views and tables. This PR enables support for materialized views.

Fixes #1239

PR #1170, opened 2018-08-16
Add summary to OpenAPI doc for RPC functions

mdr1384:
For #1158.

I note that there is a second `description` under `parameters.schema` -- it includes both summary and description. I am not sure where/how to fix that (if it needs fixing.)

PR #1259, opened 2019-03-24
Add numeric and character types to OpenAPI spec

### LINKED ISSUE #1238: smallint gets the string in OpenGraph API description
### Environment

* PostgreSQL version: 11.1
* PostgREST version: 5.2.0 (1e732ac)
* Operating system: `Linux FancyHostname 4.20.6-arch1-1-ARCH #1 SMP PREEMPT Thu Jan 31 08:22:01 UTC 2019 x86_64 GNU/Linux`

### Description of issue

The format `smallint` gets type `string`, but it should get type `integer`. This bug occurs in the API in the API description. In the actual API it works fine.

### How to reproduce

Use this sql code: 
```sql
CREATE TABLE smallint_test (
    some_number smallint
);
```

Using the defaults in the config go to: http://localhost:3000/

The response should be contain this: 
```json
{ "smallint_test":{"properties":{"some_number":{"format":"smallint","type":"string"}},"type":"object"} }
```

Do especially note `"type":"string"` for `some_number`, where it should be `"type":"integer"` instead.

fpusch:
At the moment smallint and other numeric types get `string` types in the OpenAPI spec. This attempts to fix this by adding some missing types to `toSwaggerType`. Types are based on https://www.postgresql.org/docs/current/datatype.html and https://swagger.io/docs/specification/data-models/data-types/ .

This should fix #1238 .

Note: I'm not at all familiar with Haskell, so point me in the right direction if a test or something else is missing or necessary to implement this.

Edit: I've missed the RPC test in `StructureSpec.hs`. Test should pass now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1223, Fix incorrect OpenAPI externalDocs url - @steve-chavez
 - #1221, Fix embedding other resources when having a self join - @steve-chavez
 - #1242, Fix embedding a view having a select in a where - @steve-chavez
+- #1238, Fix PostgreSQL to OpenAPI type mappings for numeric and character types - @fpusch
 
 ## [5.2.0] - 2018-12-12
 
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -41,7 +41,6 @@ toSwaggerType "boolean"           = SwaggerBoolean
 toSwaggerType "smallint"          = SwaggerInteger
 toSwaggerType "integer"           = SwaggerInteger
 toSwaggerType "bigint"            = SwaggerInteger
-toSwaggerType "decimal"           = SwaggerNumber
 toSwaggerType "numeric"           = SwaggerNumber
 toSwaggerType "real"              = SwaggerNumber
 toSwaggerType "double precision"  = SwaggerNumber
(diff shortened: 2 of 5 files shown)


steve-chavez:
@fpusch Thanks for jumping into the code!

It'd be better to have an explicit test for the mapping between postgres-openapi types. Steps for creating this test would be:

-  `CREATE TABLE openapi_types(...)` in our [fixtures](https://github.com/PostgREST/postgrest/blob/master/test/fixtures/schema.sql)
- Add GRANT here https://github.com/PostgREST/postgrest/blob/master/test/fixtures/privileges.sql#L103
- And then have a test like [this one](https://github.com/PostgREST/postgrest/blob/48803806d59881c911b5e32438e00c186e975ddd/test/Feature/StructureSpec.hs#L224)(you can pretty much copy and paste some parts) that inspects the json and verifies the mapping.

For running the tests locally there's some extra help here https://github.com/PostgREST/postgrest/pull/1170#issuecomment-414026322.

Just ping me if you need any help.

fpusch:
@steve-chavez Thanks for the tips!

I've added test cases to try to cover the mapping. Please have a look at let me know what needs fixing. I've made every mapping it's own testcase, maybe this is not appropriate.

steve-chavez:
@fpusch Looks good :smiley:! I like having one test case for each mapping.

Add an entry to the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md) and I'll merge!

fpusch:
@steve-chavez Done! Thanks for your help.

It seems like some of the tests are flaky and run into timeouts. Similar to what was seen here https://github.com/PostgREST/postgrest/pull/1256#issuecomment-473077166

steve-chavez:
This is a great contribution @fpusch. Thank you!

Issue #1260, opened 2019-03-25
Bump Docker base container version


pixie79:
Hi,

The base container image for docker has many known security issues. Whilst I appreciate probably none of the affected libs are actively used by postgrest like systemd, it might be better to update the base to either a newer debian version. Or probably better still switch to Alpine which is smaller and does not include these extras by default.

If you would like i am happy to submit a sample patch for an alpine base container instead?

Regards

Mark

steve-chavez:
Regarding alpine, see https://github.com/PostgREST/postgrest/issues/1143.

About upgrading debian, I think that'd be good. You could open a PR and I'll review it.

ruslantalpa:
I would say there is no point in having an "official" image until Haskell/ghc itself officially supports alpine.
This looks like the place it's tracked https://gitlab.haskell.org/ghc/ghc/issues/14502 and it won't be any time soon.

steve-chavez:
Copy pasting.

Switching to `stretch-slim` would be a good idea. A while ago I got a `Failed to fetch http://deb.debian.org/debian/dists/jessie-updates/main/binary-amd64/Packages  404  Not Found` when trying to build the image. It seems debian changed jessie mirror urls recently:

  - https://unix.stackexchange.com/questions/508724/failed-to-fetch-jessie-backports-repository
  - https://stackoverflow.com/questions/55386246/w-failed-to-fetch-http-deb-debian-org-debian-dists-jessie-updates-main-binary
  - https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html

steve-chavez:
Closing since https://github.com/PostgREST/postgrest/pull/1297 got merged.

PR #1262, opened 2019-03-27
switch to buster base

pixie79:
Updated the base image to buster - https://github.com/PostgREST/postgrest/issues/1260

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:jessie
+FROM debian:buster-slim
 
 ARG POSTGREST_VERSION
 


steve-chavez:
Buster doesn't seem to be stable https://www.debian.org/releases/buster/index.en.html.

I think we should stick to LTS releases listed here https://wiki.debian.org/LTS.

So `stretch` would be the best choice.

pixie79:
I considered that, stretch also has a lot of CVEs too. 
Might check CentOS and see if there base is better

steve-chavez:
CentOS would change almost all the Dockerfile. 

What are the CVEs you're concerned about? Are you sure they would affect us?

steve-chavez:
Switching to `stretch-slim` would be a good idea. A while ago I got a `Failed to fetch http://deb.debian.org/debian/dists/jessie-updates/main/binary-amd64/Packages  404  Not Found` when trying to build the image. It seems debian changed jessie mirror urls recently:

  - https://unix.stackexchange.com/questions/508724/failed-to-fetch-jessie-backports-repository
  - https://stackoverflow.com/questions/55386246/w-failed-to-fetch-http-deb-debian-org-debian-dists-jessie-updates-main-binary
  - https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html

steve-chavez:
Closing since https://github.com/PostgREST/postgrest/pull/1297 got merged.

PR #1267, opened 2019-04-02
The process should not run as root within docker

tfluehmann:
The docker image should not run as root by default.
I think it is best practice to create a separate user to run the process. Therefore I added a user and modified the docker image to use this user.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -39,6 +39,12 @@ ENV PGRST_DB_URI= \
     PGRST_PRE_REQUEST= \
     PGRST_ROLE_CLAIM_KEY=".role"
 
+RUN groupadd -g 1000 postgrest && \
+    useradd -r -u 1000 -g postgrest postgrest && \
+    chown postgrest:postgrest /etc/postgrest.conf
+
+USER 1000
+
 # PostgREST reads /etc/postgrest.conf so map the configuration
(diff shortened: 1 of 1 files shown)


steve-chavez:
Running as a minimally privileged user seems like a good idea.

Wouldn't it be better to also create a dedicated group instead of using `nogroup`? According to the [adduser manual](https://manpages.debian.org/jessie/adduser/adduser.8.en.html), `nogroup` is being assigned. We could do something like:
```bash
RUN groupadd -g 1000 postgrest && \
    useradd -r -u 1000 -g postgrest postgrest
USER postgrest
```

Perhaps also chowning the config file(`/etc/postgrest.conf`) for that user?

steve-chavez:
Also I've thought about using the `nobody` user like [Prometheus does](https://github.com/prometheus/prometheus/blob/master/Dockerfile#L14) but maybe is not such a good idea https://github.com/prometheus/prometheus/issues/3441.

A dedicated group/user seems like the best choice.

tfluehmann:
Yes, this was my intention too.
Adding a group as well and changing the postgres.conf seems to be even better.

steve-chavez:
Regarding the user id being a number for k8s, I think that's good. The more use cases we can cover the better.

@tfluehmann Thank you! Merged :smiley:.

PR #1268, opened 2019-04-03
Fix tests expected responses due to changes in pg11

Qu4tro:
Hey there.

While running PostgREST test suit, I found 5 tests that failed. Investigating it further, it seems there was a change in postgres error messages, making the expected outputs in a way, obsolete.

Reproducing this issue by running the tests as described in the [documentation](http://postgrest.org/en/v5.1/install.html?highlight=test#testing-with-docker):
```
$ docker run --name db-scripting-test -e POSTGRES_PASSWORD=pwd -p 5434:5432 -d postgres:latest
$ POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://postgres:pwd@localhost:5434" test_db) stack test
```
Notice the flag as latest. Omitting it or using the tag `:11` or even `:11.2` would fetch the same postgres image. To get the tests running we can simply pick the tag `:10` or anything under (and above PostgREST minimum postgres version).
I specifically picked 11.2 (which is currently latest), because both `:11.0` and `11.1`, tests fail with another issue.

This PR simply adds the `pgVersion112` value and if expressions on the affects tests to get the correct expected value.

Let me know if there's anything you would like me to improve on this PR.

Cheers

### DIFF
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -374,6 +374,9 @@ pgVersion96 = PgVersion 90600 "9.6"
 pgVersion100 :: PgVersion
 pgVersion100 = PgVersion 100000 "10"
 
+pgVersion112 :: PgVersion
+pgVersion112 = PgVersion 110002 "11.2"
+
 sourceCTEName :: SqlFragment
 sourceCTEName = "pg_source"
 
--- a/test/Feature/AuthSpec.hs
+++ b/test/Feature/AuthSpec.hs
@@ -11,16 +11,27 @@ import Network.Wai (Application)
 
 import Protolude hiding (get)
 
-spec :: SpecWith Application
-spec = describe "authorization" $ do
+import PostgREST.Types (PgVersion, pgVersion112)
+
+spec :: PgVersion -> SpecWith Application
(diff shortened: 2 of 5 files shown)


steve-chavez:
Hey @Qu4tro. I've reproduced the issue in a local pg11. Your solution is good!

To double check, we could have a test suite running on pg11 in CircleCI. Doing this would be basically copying this [config for pg10](https://github.com/PostgREST/postgrest/blob/master/.circleci/config.yml#L141-L171) and changing it for pg11(CircleCI has an [image available](https://discuss.circleci.com/t/postgresql-11-circleci-docker-image-available/25928)), then adding this job [here](https://github.com/PostgREST/postgrest/blob/master/.circleci/config.yml#L255-L327) in a similar way it's done for the other pg versions.

Also, as you might have noticed, some tests were timing out(I had to restart circleci manually). You can rebase against master to avoid this. I added a temporary solution in https://github.com/PostgREST/postgrest/pull/1266.

steve-chavez:
Just added a CircleCI test suite for pg11. Thank you!

Qu4tro:
Ah thanks! Didn't have much time over the weekend.

Issue #1038, opened 2017-12-12
Some tests fail on PG 10


ruslantalpa:
If the local db where the tests run is 10, these tests fail, everything works ono 9.6.
all the failing tests relate to `text_search_vector`

(there is a small chance lines numbers might be a bit off since this is my custom build and i might have added a few test)
```
Failures:

  test/Feature/QuerySpec.hs:99: 
  1) Feature.QuerySpec, Filtering response, Full text search operator, finds matches with to_tsquery
       body mismatch:
         expected: [{"text_search_vector":"'fun':5 'imposs':9 'kind':3"}]
         but got:  [{"text_search_vector":"'do':7 'fun':5 'impossible':9 'it':1 'kind':3 'of':4 's':2 'the':8 'to':6"}]

  test/Feature/QuerySpec.hs:104: 
  2) Feature.QuerySpec, Filtering response, Full text search operator, can use lexeme boolean operators(&=%26, |=%7C, !) in to_tsquery
       body mismatch:
         expected: [{"text_search_vector":"'also':2 'fun':3 'possibl':8"}]
         but got:  [{"text_search_vector":"'also':2 'but':1 'do':5 'fun':3 'is':7 'possible':8 'to':4 'what':6"}]

  test/Feature/QuerySpec.hs:117: 
  3) Feature.QuerySpec, Filtering response, Full text search operator, finds matches with plainto_tsquery
       body mismatch:
         expected: [{"text_search_vector":"'ate':3 'cat':2 'fat':1 'rat':4"}]
         but got:  []

  test/Feature/QuerySpec.hs:140: 
  4) Feature.QuerySpec, Filtering response, Full text search operator, can be negated with not operator
       body mismatch:
         expected: "[{\"text_search_vector\":\"'fun':5 'imposs':9 'kind':3\"},{\"text_search_vector\":\"'also':2 'fun':3 'possibl':8\"},{\"text_search_vector\":\"'amus':5 'fair':7 'impossibl':9 'peu':4\"},{\"text_search_vector\":\"'art':4 'spass':5 'unmog':7\"}]"
[...]

PR #1269, opened 2019-04-03
Fix tests with postgres instances with a different locale

Qu4tro:
Hey again,

My locale is `en_GB.UTF-8` and when running the tests I get 7 tests that fail due to `LC_MONETARY` (pounds instead of dollars).  

This issue can be reproduced on a postgres instance by altering this PR, the `LC_MONETARY` line by the following one:
`ALTER DATABASE $DB SET LC_MONETARY = "en_GB.UTF-8";`
You postgres installation and system must support already the `en_GB.UTF-8` locale, although any other, whose symbol is not `$` will fail as well and can be used to reproduce the issue.

This PR tries to control this aspect, by creating a test database with all locale properties set to `POSIX`, which is standardized.

Let me know, if there's something to improve/change.
Cheers.


steve-chavez:
Can reproduce. Reminds me of a similar issue related to fts https://github.com/PostgREST/postgrest/issues/1038.

Issue #1257, opened 2019-03-19
404 when PATCH/update didn't change anything


sevagh:
Hello. I'm doing a PATCH to update a row that doesn't exist. There's a GitHub issue (https://github.com/PostgREST/postgrest/issues/182) and pull request/changelog entry from 2015 (https://github.com/PostgREST/postgrest/commit/709e70561f8b527998b8ed8ac0916e92b1a5e17b) that indicates a PATCH that didn't change any rows should return 404.

Right now in my tests, I'm getting `204` whether or not any rows were updated. Could it be that this "404 when no rows were updated" behavior got lost somewhere along the line?

I'm using `postgREST v0.5.1` and Postgres 9.

steve-chavez:
Not sure why that behavior changed in https://github.com/PostgREST/postgrest/pull/763, maybe it was just something that slipped by.

I also think that 404 is the correct status code. Thanks for the report.

Issue #1070, opened 2018-02-21
Respond with HTTP 200 instead of 201 when no rows inserted


steve-chavez:
Currently this is happening with POST when the payload is `[]`, see: https://github.com/begriffs/postgrest/blob/108f3cd651a448fad065faf81a2090cee8a8d9fb/test/Feature/InsertSpec.hs#L129-L135

And when `Prefer: resolution=ignore-duplicates` is specified and all rows in the payload are duplicates: https://github.com/begriffs/postgrest/blob/108f3cd651a448fad065faf81a2090cee8a8d9fb/test/Feature/UpsertSpec.hs#L95-L101.

Also according to [RFC](https://tools.ietf.org/html/rfc7231#section-4.3.4), PUT should respond with 200 when the row is updated and with 201 when the row is inserted, currently it's always responding with 200.

These cases can be corrected by using the [xmax](https://www.postgresql.org/docs/10/static/ddl-system-columns.html) system column, as described [here](https://stackoverflow.com/questions/39058213/postgresql-upsert-differentiate-inserted-and-updated-rows-using-system-columns-x/39204667#39204667).

steve-chavez:
The `xmax` trick wouldn't work with auto-updatable VIEWs since only TABLEs have them.

wolfgangwalther:
[This](https://stackoverflow.com/a/34768232) could work for both tables and views, I think.

taimoorzaeem:
@steve-chavez The design for this fix/feature seems rather vague. I am thinking of implementing it the following way:

- Add `inserted` and `updated` fields in `ResultSet` here.
https://github.com/PostgREST/postgrest/blob/57fa2719ddade4c25aa12acea355ca6ee93a466e/src/PostgREST/Query/Statements.hs#L37-L53

- Add the local GUCs `pgrst.inserted` and `pgrst.updated` in the query here and then read them with `AS` in `prepareWrite` statement.
https://github.com/PostgREST/postgrest/blob/57fa2719ddade4c25aa12acea355ca6ee93a466e/src/PostgREST/Query/QueryBuilder.hs#L85-L105

- Then, use `Hasql.decoders` to decode `pgrst.inserted` and `pgrst.updated` variables. Compare them and set `rsGucStatus` accordingly.

I am not very familiar with PostgreSQL IO yet, let me know other things that need to be changed for this fix.

steve-chavez:
@taimoorzaeem Looks good. Go ahead :+1:

[...]

PR #1270, opened 2019-04-04
Fix invalid SQL generated for bulk upsert with empty array

### LINKED ISSUE #1265: invalid SQL generated for bulk upsert with empty array
### Environment

* PostgreSQL version: 9.6.11
* PostgREST version: 5.2.0 (1e732ac)
* Operating system: Debian 9.8

### Description of issue

When upserting an empty array of items, PostgREST reports the error

    syntax error at or near "RETURNING"

while non-empty upserts or plain inserts are handled fine. This seems to be irrespective of table structure. E.g.:

```
$ curl -X POST -H 'Content-type: application/json' -d '[]' -H 'Prefer: resolution=merge-duplicates' localhost:3000/articles
{"hint":null,"details":null,"code":"42601","message":"syntax error at or near \"RETURNING\""}
```

while this succeeds:

```
$ curl -X POST -H 'Content-type: application/json' -d '[]' localhost:3000/articles
```

Qu4tro:
Simple fix and respective test that resolves #1265.

I've added a simple check to check what columns are selected. If none are (the reason behind the syntax error), then it creates a `DO NOTHING` as suggested in the issue.

### DIFF
--- a/test/Feature/UpsertSpec.hs
+++ b/test/Feature/UpsertSpec.hs
@@ -44,7 +44,7 @@ spec =
             , matchHeaders = ["Preference-Applied" <:> "resolution=merge-duplicates", matchContentTypeJson]
             }
 
-        it "succeeds when the payload has no elements" $ do
+        it "succeeds when the payload has no elements" $
           request methodPost "/articles" [("Prefer", "return=representation"), ("Prefer", "resolution=merge-duplicates")]
             [json|[]|] `shouldRespondWith`
             [json|[]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
--- a/test/Feature/UpsertSpec.hs
+++ b/test/Feature/UpsertSpec.hs
@@ -44,6 +44,12 @@ spec =
             , matchHeaders = ["Preference-Applied" <:> "resolution=merge-duplicates", matchContentTypeJson]
             }
 
+        it "succeeds when the payload has no elements" $ do
+          request methodPost "/articles" [("Prefer", "return=representation"), ("Prefer", "resolution=merge-duplicates")]
+            [json|[]|] `shouldRespondWith`
+            [json|[]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
+
+
       context "when Prefer: resolution=ignore-duplicates is specified" $ do
         it "INSERTs and ignores rows on pk conflict" $
           request methodPost "/tiobe_pls" [("Prefer", "return=representation"), ("Prefer", "resolution=ignore-duplicates")]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1221, Fix embedding other resources when having a self join - @steve-chavez
 - #1242, Fix embedding a view having a select in a where - @steve-chavez
 - #1238, Fix PostgreSQL to OpenAPI type mappings for numeric and character types - @fpusch
+- #1265, Fix query generated on bulk upsert with an empty array - @qu4tro
 
 ## [5.2.0] - 2018-12-12
 
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -295,8 +295,10 @@ requestToQuery schema _ (DbMutate (Insert mainTbl iCols onConflct putConditions 
       IgnoreDuplicates ->
         "DO NOTHING"
       MergeDuplicates  ->
-        "DO UPDATE SET " <> intercalate ", " (pgFmtIdent <> const " = EXCLUDED." <> pgFmtIdent <$> S.toList iCols)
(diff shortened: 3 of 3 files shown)


test/Feature/UpsertSpec.hs
```diff
@@ -44,6 +44,12 @@ spec =
             , matchHeaders = ["Preference-Applied" <:> "resolution=merge-duplicates", matchContentTypeJson]
             }
 
+        it "succeeds when the payload has no elements" $
+          request methodPost "/articles" [("Prefer", "return=representation"), ("Prefer", "resolution=merge-duplicates")]
+            [json|[]|] `shouldRespondWith`
+            [json|[]|] { matchStatus = 201 , matchHeaders = [matchContentTypeJson] }
```
steve-chavez: HTTP 201 Created is a bit misleading, an HTTP 200 OK would be better. You could conditionally change the status by comparing if `queryTotal = 0` I believe.

src/PostgREST/App.hs
```diff
@@ -197,9 +197,10 @@ app dbStructure proc conf apiRequest =
               else do
                 let r = contentRangeH 0 (toInteger $ queryTotal-1)
                           (toInteger <$> if shouldCount then Just queryTotal else Nothing)
-                    s = if iPreferRepresentation apiRequest == Full
+                    s_200 = if iPreferRepresentation apiRequest == Full
                           then status200
                           else status204
+                    s = if queryTotal == 0 then status404 else s_200
```
steve-chavez: Is this related to https://github.com/PostgREST/postgrest/issues/1257?

steve-chavez:
You could avoid the test timeouts by rebasing to master.

src/PostgREST/App.hs
Qu4tro: Oh. That was a mistake, wrong branch. Sorry about that.

Yes. I was trying to fix it, but my current approach, breaks a few other tests.

I can create a PR and we can discuss it there.

test/Feature/UpsertSpec.hs
Qu4tro: I agree with you that HTTP 201 is misleading.

I've tried to not to de-rail, from current behavior and I had found this test:
```
-- test/Feature/InsertSpec.hs (lines 131-137)
it "returns empty array when no items inserted, and return=rep" $ do
  p <- request methodPost "/no_pk"
               [("Prefer", "return=representation")]
               [json| [] |]
  liftIO $ do
    simpleBody p `shouldBe` [json| [] |]
    simpleStatus p `shouldBe` created201
```

Which I would argue is similar to the one created in this PR. 
Let me know, which option do you think it's more fitting: Change both behaviors to 200, change just this PR to 200 or keep it as is for consistency.

steve-chavez: Both behaviors to 200 seems correct. My reasoning is that with [201](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201) we can include a `Location` header for identifying the resource, and with `[]` as the response we can't do that.

src/PostgREST/App.hs
steve-chavez: Thank you. That makes reviewing easier.

Qu4tro:
Changed return code to 200 and both tests to expect 200 as discussed.

A few others tests started to fail as a result, which I changed them as well.
These fall under two fields:

`return=minimal` - When using return=minimal, the query that it is built is hardcoded with a zero. I'm guessing that's for performance reasons and I don't think that's bad, since it's explicit. 
`resolution=ignore-duplicates` when no row is actually inserted. 

I think both cases are fitted for a 200 as well. In this second case, no resource is created, similarly to the case in the original discussion.

In the `return=minimal` the request is already being handicapped and we don't actually know if a resource was created (i.e. ignore duplicates above), so I think a 200 OK (we ran the query and everything went fine) is appropriate.

steve-chavez:
@Qu4tro Just realized I bumped into the `return=minimal` problem in a previous PR https://github.com/PostgREST/postgrest/pull/1048#discussion_r169307092, opened an issue for that here https://github.com/PostgREST/postgrest/issues/1070.

I think `return=minimal` responding with 200 is not accurate when a resource is created, it should be `201`. We could make it work by having the same query for this [two](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/QueryBuilder.hs#L111-L117) and returning [xmax](https://www.postgresql.org/docs/11/ddl-system-columns.html) in this [line](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/DbRequestBuilder.hs#L342).

This is getting away from the original https://github.com/PostgREST/postgrest/issues/1265 though. So how about if you open another PR for #1070?
You could remove your last commit and then I'll merge this PR.

steve-chavez:
Also, don't forget to add an entry to the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md).

Qu4tro:
Cool, all done.

PR #1271, opened 2019-04-08
include error message with custom raised errors

ruslantalpa:
currently only details/hint shows up, error (the main message) is skipped

there is a test to fix, trying to do it directly from GH interface, don't have local clone configured, if it does not work we'll delete it later

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -129,6 +129,7 @@ instance JSON.ToJSON H.CommandError where
   toJSON (H.ResultError (H.ServerError c m d h)) = case toS c of
     'P':'T':_ ->
       JSON.object [
+        "message" .= (toS m::Text),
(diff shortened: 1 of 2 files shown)


steve-chavez:
The main message does show in a `"error": "Payment Required"` json, but not sure where this is coming from.

I think to maintain consistency it'll be better to show the message in the `"message"` key.

steve-chavez:
Now that https://github.com/PostgREST/postgrest/pull/1282 got merged, I think you can rebase and do the same change [here](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/Error.hs#L112) and `message` will show.

steve-chavez:
Now that I've remembered, in the case of `PT***` errors, the `message` is included as part of the status(e.g. 402 Payment Required). It's not included in the json body, it could be included, but it would be redundant.

PR #1272, opened 2019-04-08
Fix wrong status 404 when PATCH request didn't change anything

### LINKED ISSUE #1257: 404 when PATCH/update didn't change anything

Qu4tro:
This is a simple attempt to fix the reported bug described in title.
So, citing #1257 and confirmed by this test in this PR:

> Right now in my tests, I'm getting 204 whether or not any rows were updated. 

This PR can be tweaked to take care of some of the failing tests, but like in the previous PR, I would argue that some of those failing tests should be tweaked as well.

Let me know what you think.

This resolves #1257 .

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -162,8 +162,7 @@ app dbStructure proc conf apiRequest =
                     , if iPreferRepresentation apiRequest == Full
                         then Just $ toHeader contentType
                         else Nothing
-                    , Just . contentRangeH 1 0 $
-                        if shouldCount then Just queryTotal else Nothing
+                    , Just $ contentRangeH shouldCount 1 0 queryTotal
                     , if null pkCols
                         then Nothing
                         else (\x -> ("Preference-Applied", show x)) <$> iPreferResolution apiRequest
@@ -190,8 +189,7 @@ app dbStructure proc conf apiRequest =
               let (_, queryTotal, _, body) = extractQueryResult row
 
                   updateIsNoOp       = S.null $ pjKeys pJson
-                  contentRangeHeader = contentRangeH 0 (queryTotal - 1) $
-                                          if shouldCount then Just queryTotal else Nothing
+                  contentRangeHeader = contentRangeH shouldCount 0 (queryTotal - 1) queryTotal
                   minimalHeaders     = [contentRangeHeader]
                   fullHeaders        = toHeader contentType : minimalHeaders
 
@@ -205,7 +203,7 @@ app dbStructure proc conf apiRequest =
 
                 (CTSingularJSON, _, Full) ->
                   do HT.condemn
-                     return $ singularityError 0
+                     return $ singularityError queryTotal
 
                 (_, _, Full) ->
                   return $ responseLBS status fullHeaders (toS body)
@@ -255,8 +253,7 @@ app dbStructure proc conf apiRequest =
                     (iPreferRepresentation apiRequest) []
               row <- H.statement mempty stm
               let (_, queryTotal, _, body) = extractQueryResult row
-                  r = contentRangeH 1 0 $
-                    if shouldCount then Just queryTotal else Nothing
+                  r = contentRangeH shouldCount 1 0 queryTotal
               if contentType == CTSingularJSON
                  && queryTotal /= 1
                  && iPreferRepresentation apiRequest == Full
@@ -332,7 +329,7 @@ app dbStructure proc conf apiRequest =
       rangeHeader queryTotal tableTotal =
         let lower = rangeOffset topLevelRange
             upper = lower + toInteger queryTotal - 1
-            contentRange = contentRangeH lower upper (toInteger <$> tableTotal)
+            contentRange = contentRangeH (isJust tableTotal) lower upper $ (toInteger . fromJust) tableTotal
             status = rangeStatus lower upper (toInteger <$> tableTotal)
         in (status, contentRange)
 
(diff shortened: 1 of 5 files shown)


Qu4tro:
Added a few changes:

Using a empty payload no longer touches the db. They keep returning the same status as before.
PATCHs whose conditions result in no records found (and therefore no records updated) return 404.

1º was a necessary condition for 2º, because we need to distinguish between, declaring no changes and not founding records to change, which return different status codes.

What do you think?

src/PostgREST/App.hs
```diff
@@ -188,21 +188,47 @@ app dbStructure proc conf apiRequest =
...
-                          then status200
-                          else status204
-                return $ if iPreferRepresentation apiRequest == Full
-                  then responseLBS s [toHeader contentType, r] (toS body)
-                  else responseLBS s [r] ""
+
+                  contentRangeHeader = contentRangeH 0 (toInteger $ queryTotal-1)
+                        (toInteger <$> if shouldCount then Just queryTotal else Nothing)
+
+                  updateIsNoOp = case mutationDbRequest tSchema tName of
```
steve-chavez: This data can be obtained with `pjKeys pJson`(see PayloadJSON in Types.hs), without passing through `mutateRequest`.

steve-chavez:
> we need to distinguish between, declaring no changes and not founding records to change, which return different status codes.

Yes. This is correct :+1: 

However, the wildcard pattern matchs are a bit hard to follow. I think not hitting the database should not be a concern for now and the code can be a bit more concise(the added variables are good though). It could be something like:
```haskell
let
  contentRangeHeader = contentRangeH 0 (toInteger $ queryTotal - 1)
        (toInteger <$> if shouldCount then Just queryTotal else Nothing)
  minimalHeaders = [contentRangeHeader]
  fullHeaders = toHeader contentType : minimalHeaders
  updateIsNoOp = S.null $ pjKeys pJson
  status | queryTotal == 0 && not updateIsNoOp = status404
         | iPreferRepresentation apiRequest == Full = status200
         | otherwise = status204
return $ if iPreferRepresentation apiRequest == Full
  then responseLBS status fullHeaders (toS body)
  else responseLBS status minimalHeaders mempty
```

Qu4tro:
A few more commits.

- `pjKeys`
- Simplified the case statement. I've kept it over the `if`s expression, to better handle the `CTSingularJSON` possibility.
- Made contentRangeH polymorphic, so we can reduce the number of `toInteger` calls.

Qu4tro:
Additional commit for some whitespace. I think it reads better.

src/PostgREST/App.hs
```diff
@@ -188,21 +188,31 @@ app dbStructure proc conf apiRequest =
...
+                         | iPreferRepresentation apiRequest == Full = status200
+                         | otherwise                                = status204
+
+              case (contentType, queryTotal, iPreferRepresentation apiRequest) of
+                (CTSingularJSON, 1, Full) ->
+                  return $ responseLBS status fullHeaders (toS body)
+
+                (CTSingularJSON, _, Full) ->
+                  do HT.condemn
+                     return $ singularityError 0
```
steve-chavez: Previously we had:
```haskell
return $ singularityError (toInteger queryTotal)
```
It'd be better to keep the queryTotal instead of always returning 0.

I'm surprised this wasn't included in any test. Perhaps you could add one for making sure we don't miss this again.

src/PostgREST/App.hs
```diff
@@ -381,7 +392,7 @@ rangeStatus lower upper (Just total)
   | (1 + upper - lower) < total = status206
   | otherwise               = status200
 
-contentRangeH :: Integer -> Integer -> Maybe Integer -> Header
+contentRangeH :: (Integral a, Show a) => a -> a -> Maybe a -> Header
```
steve-chavez: Nice! :+1:

src/PostgREST/App.hs
Qu4tro: Ah, nice catch, thanks!

Qu4tro:
Further commit that fixes the issue mentioned above. I've also hardcoded the tests, so it doesn't only check if it is an error, but which as well.

Curious thing happened on 
https://github.com/PostgREST/postgrest/pull/1272/files#diff-d36f1ad401857a200ca8d27e932681f8L64
Re-running the tests (without destroying db), yields different results (6 or 7).
Kept previous check instead, as the case its tested further down.

src/PostgREST/App.hs
```diff
@@ -188,21 +187,30 @@ app dbStructure proc conf apiRequest =
...
+                  updateIsNoOp       = S.null $ pjKeys pJson
+                  contentRangeHeader = contentRangeH shouldCount 0 (queryTotal - 1) queryTotal
+                  minimalHeaders     = [contentRangeHeader]
+                  fullHeaders        = toHeader contentType : minimalHeaders
+
+                  status | queryTotal == 0 && not updateIsNoOp      = status404
+                         | iPreferRepresentation apiRequest == Full = status200
+                         | otherwise                                = status204
+
+              case (contentType, queryTotal, iPreferRepresentation apiRequest) of
```
steve-chavez: queryTotal is only used by 1 match, so how about expressing this as:
```haskell
case (contentType, iPreferRepresentation apiRequest) of
  (CTSingularJSON, Full)
    | queryTotal == 1 -> return $ responseLBS status fullHeaders (toS body)
    | otherwise -> HT.condemn >> return (singularityError queryTotal)

  (_, Full) ->
    return $ responseLBS status fullHeaders (toS body)

  (_, _) ->
    return $ responseLBS status minimalHeaders mempty
```

src/PostgREST/App.hs
```diff
@@ -322,7 +329,7 @@ app dbStructure proc conf apiRequest =
       rangeHeader queryTotal tableTotal =
         let lower = rangeOffset topLevelRange
             upper = lower + toInteger queryTotal - 1
-            contentRange = contentRangeH lower upper (toInteger <$> tableTotal)
+            contentRange = contentRangeH (isJust tableTotal) lower upper $ (toInteger . fromJust) tableTotal
```
steve-chavez: We should [avoid partial functions](https://wiki.haskell.org/Avoiding_partial_functions) as much as we can. There are some of those in the codebase, but I've been removing them in any chance I get. Could you express this in other way and avoid `fromJust`?

Qu4tro: Yes, I understand.

I've changed to this form, because it was shorter and `total` was never instantiated if `shouldCount` was false. 

Anyway, to avoid this form, which is I agree, is not ideal, I just reverted to the previous form.

src/PostgREST/App.hs
Qu4tro: Changed to suggested.

steve-chavez:
@Qu4tro Thank you! Sorry the review took so long.

PR #1275, opened 2019-04-11
[WIP] Correct status code on upserts

### LINKED ISSUE #1070: Respond with HTTP 200 instead of 201 when no rows inserted

Qu4tro:
As discussed in #1270 . Quoting my description of the changes made:

> Changed return code to 200 and both tests to expect 200 as discussed. A few others tests started to fail as a result, which I changed them as well. These fall under two fields: 
> `return=minimal` - When using return=minimal, the query  that it is built is hardcoded with a zero. I'm guessing that's for  performance reasons and I don't think that's bad, since it's explicit. 
`resolution=ignore-duplicates` when no row is actually inserted.

@steve-chavez , also quoting your description of an additional step that will also fix #1070 :
> I think `return=minimal` responding with 200 is not accurate when a resource is created, it should be `201`. We could make it work by having the same query for this [two](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/QueryBuilder.hs#L111-L117) and returning [xmax](https://www.postgresql.org/docs/11/ddl-system-columns.html) in this [line](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/DbRequestBuilder.hs#L342).

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -167,7 +167,7 @@ app dbStructure proc cols conf apiRequest =
 
                 (CTSingularJSON, _, Full) ->
                   do HT.condemn
-                     return $ singularityError (toInteger queryTotal)
+                     return . errorResponseFor . singularityError $ toInteger queryTotal
 
                 (_, 0, _) ->
                   successResponse
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -154,16 +154,24 @@ app dbStructure proc cols conf apiRequest =
                         then Nothing
                         else (\x -> ("Preference-Applied", show x)) <$> iPreferResolution apiRequest
                     ]
-              if contentType == CTSingularJSON
-                 && queryTotal /= 1
-                 && iPreferRepresentation apiRequest == Full
-                then do
-                  HT.condemn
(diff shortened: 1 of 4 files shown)


steve-chavez:
Code here is outdated by now. Closing.. original issue still open.

PR #1280, opened 2019-04-20
Add config option for pool timeout

### LINKED ISSUE #1278: Allow for specifying the connection pool timeout value
### Environment

* PostgREST version: 5.2

### Description of issue

Currently the ttl value of a pooled connection is hard coded to 10 seconds:

https://github.com/PostgREST/postgrest/blob/0d6d112b38e7b51f31c5ba9cf25ad344cf801c96/main/Main.hs#L167

It would be great if this could be set by an environment variable.

Qu4tro:
Resolves #1278 

Added to config, key `db-pool-timeout` that accepts float/int.

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -35,6 +35,7 @@ import qualified Data.Configurator.Parser     as C
 import           Data.Configurator.Types      as C
 import           Data.List                    (lookup)
 import           Data.Monoid
+-- import           Data.Time.Clock.UTC          (NominalDiffTime)
 import           Data.Scientific              (floatingOrInteger)
 import           Data.String                  (String)
 import           Data.Text                    (dropWhileEnd, dropEnd,
@@ -73,7 +74,7 @@ data AppConfig = AppConfig {
   , configJwtAudience       :: Maybe StringOrURI
 
   , configPool              :: Int
-  , configPoolTimeout       :: Int
+  , configPoolTimeout       :: Float
   , configMaxRows           :: Maybe Integer
   , configReqCheck          :: Maybe Text
   , configQuiet             :: Bool
@@ -137,7 +138,7 @@ readOptions = do
           <*> (fromMaybe False . join . fmap coerceBool <$> C.key "secret-is-base64")
           <*> parseJwtAudience "jwt-aud"
           <*> (fromMaybe 10 . join . fmap coerceInt <$> C.key "db-pool")
-          <*> (fromMaybe 10 . join . fmap coerceInt <$> C.key "db-pool-timeout")
+          <*> (fromMaybe 10.0 . join . fmap coerceFloat <$> C.key "db-pool-timeout")
           <*> (join . fmap coerceInt <$> C.key "max-rows")
           <*> (mfilter (/= "") <$> C.key "pre-request")
           <*> pure False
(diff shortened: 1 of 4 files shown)


ruslantalpa:
What’s the usefullness for supporting float? (Without it you do net need an aditional coerce function)

Qu4tro:
Not exactly sure. It has definitely never been a use-case of mine.

I've added the second commit (support float vs int), because the underlying Wai function receives a `NominalDiffTime` which is a fractional number and not just an integer.

This is also indicated by the `Settings` on [Hasql.Pool](https://hackage.haskell.org/package/hasql-pool-0.5/docs/Hasql-Pool.html#t:Settings):
> Timeout. An amount of time for which an unused resource is kept open. The smallest acceptable value is 0.5 seconds.

steve-chavez:
Tried the `int` version and though it compiled just fine, it failed at runtime with:
```
postgrest: Data.Pool.pool : invalid idle time 0.00000000001s
CallStack (from HasCallStack):
  error, called at ./Data/Pool.hs:393:5 in resource-pool-0.2.3.2-GmoeVeJ8L5Y51jMJ5dPk3U:Data.Pool
```

So supporting `float` seems correct.

Now, I've tried setting a `db-pool-timeout = 100` and `db-pool-timeout = 100.0` and the connection seems to always timeout at `10` seconds, so maybe the `coerceFloat` is always resulting in `Nothing`.

To test this use a query like:
```sql
select * from pg_stat_activity where usename = 'postgrest_test_authenticator';
```
It should keep showing a row for more than `10` seconds.

Qu4tro:
Sorry, could you detail the testing? I'm not exactly sure on how to.

I think you're right, about the `Nothing`. I made a small fix on a hunch.

Qu4tro:
I've also an added an additional `coerceNum` func. This generalizes the other two and as such they can be made simpler.

steve-chavez:
Now it works! 

Regarding the testing, I just manually started pgrst with different timeouts and then ran the query above. We could have a bash test for this, but I think we would be testing hasql pool. So it's fine to manual test.

src/PostgREST/Config.hs
```diff
@@ -73,6 +74,7 @@ data AppConfig = AppConfig {
   , configJwtAudience       :: Maybe StringOrURI
 
   , configPool              :: Int
+  , configPoolTimeout       :: Float
```
steve-chavez: Hmm. I don't think the users would need a timeout like `11.7` seconds.
And when I try a float(`30.5`), the timeout always gets the default `10` seconds.

Converting this into `Int` would be better.

Qu4tro: Hm, kind of curious to know where things aren't going as expected.

But, I won't push further. As I've mentioned above, I had added float support merely for that piece in hasql docs. 
I've added a commit to use `Int` instead.

Qu4tro:
I understand this test is not trivial to automate. I'm still wrapping my head around io-tests and memory-tests. I guess this can be added in a later PR.

steve-chavez:
@Qu4tro Add an entry to the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md) and I'll merge!

Also, please add an entry for the previous https://github.com/PostgREST/postgrest/pull/1272.

Qu4tro:
@steve-chavez Added changelog.

Thanks for the feedback here.

steve-chavez:
@Qu4tro Thank YOU for helping us! :smiley:

PR #1282, opened 2019-04-21
Small refactor around error responses and more tests

### LINKED ISSUE #880: Some errors are overwriting the details field
Did they mean to have a hint property in these rather than duplicate details?

https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Error.hs#L115-L133

### LINKED ISSUE #1285: Reread configuration or crash on DB connection failure
### Environment

* PostgreSQL version: Not relevant
* PostgREST version: 5.2.0
* Operating system: Not relevant

### Description of issue

Some operations team will rotate passwords on a schedule, with affected infrastructure detecting those changes by being disconnected and having to reconnect with the updated password (fetched from a central secure storage).

PostgREST will try to reconnect using the wrong password forever without either rereading the configuration or crashing. If PostgREST would crash, a `systemd` unit could refetch configuration from a central system whenever it gets started up which would allow for a password change to be discovered.

If PostgREST were to reload the configuration file on failure, another mechanism could update the configuration file on password change.

Expected behaviour would be for either PostgREST to crash on incorrect password (or have that be a configurable option) or reread the configuration.

Qu4tro:
Hey there.

This is a PR to clean out a bit the error responses code and improve coverage for it. I've tried to keep it from noticeable run-time changes.

Changes from this PR:
- Created typeclass `PostgRestError`, which is used for all error responses. It also requires the error to be an instance of `toJSON` to render the body. This was already the case for the pgerrors and apiRequestError.

- Moved `ApiRequestError` from `Types.hs` to `Error.hs`.
- Created instance of `PostgRestError` for `ApiRequestError`.

- Created `PgError` datatype. It bundles the hasql error and the `authed` var.
- Created instance of `PostgRestError` for `PgError`.
- Created instance of `toJSON` for `PgError`. This only calls `toJSON` on the hasql error, using the already implemented instances to render the final error.
- Fix #880 and fix #1285. 

- Create `SimpleError` datatype. Aggregated all responses that used `simpleError` function into this + singularityError.
- Created instance of `PostgRestError` for `SimpleError`.
- Created instance of `toJSON` for `SimpleError`.

- Hardcode several tests.
- Created a few more tests.

- Some alignment.

There are some things I think this could be improved further, but I think this is the gist of this PR and wanted some feedback as well:

- Test cases for #880. I'm not sure on how to reproduce the `RowError`s from hasql.
- SimpleError. Currently, it's a `misc` error datatype. Maybe we could extract the errors inside, into smaller, more specific error types.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -75,7 +75,7 @@ connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
       putStrLn ("Attempting to connect to the database..." :: Text)
       connected <- connectionStatus pool
       case connected of
-        FatalError reason         -> hPutStrLn stderr reason 
+        ServerFatalError reason         -> hPutStrLn stderr reason 
                                       >> killThread mainTid  -- Fatal error when connecting
         NotConnected              -> return ()               -- Unreachable
         Connected actualPgVersion -> do                      -- Procede with initialization
@@ -114,12 +114,12 @@ connectionStatus pool =
           let err = PgError False e
           hPutStrLn stderr . toS $ errorPayload err
           case checkIsFatal err of
-            (Just reason) -> return $ FatalError reason
+            (Just reason) -> return $ ServerFatalError reason
             Nothing       -> return NotConnected
 
         Right version ->
           if version < minimumPgVersion
-             then return . FatalError $ "Cannot run in this PostgreSQL version, PostgREST needs at least " <> pgvName minimumPgVersion
+             then return . ServerFatalError $ "Cannot run in this PostgreSQL version, PostgREST needs at least " <> pgvName minimumPgVersion
              else return . Connected  $ version
 
     shouldRetry :: RetryStatus -> ConnectionStatus -> IO Bool
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -7,9 +7,9 @@ import           PostgREST.App              (postgrest)
 import           PostgREST.Config           (AppConfig (..), configPoolTimeout',
                                              prettyVersion, readOptions)
 import           PostgREST.DbStructure      (getDbStructure, getPgVersion)
-import           PostgREST.Error            (errorPayload, isFatal, PgError(PgError))
+import           PostgREST.Error            (errorPayload, checkIsFatal, PgError(PgError))
 import           PostgREST.OpenAPI          (isMalformedProxyUri)
-import           PostgREST.Types            (DbStructure, Schema, PgVersion(..), minimumPgVersion)
+import           PostgREST.Types            (DbStructure, Schema, PgVersion(..), minimumPgVersion, ConnectionStatus(..))
 import           Protolude                  hiding (hPutStrLn, replace)
 
 
@@ -28,7 +28,6 @@ import           Data.Text.Encoding         (decodeUtf8, encodeUtf8)
 import           Data.Text.IO               (hPutStrLn, readFile)
 import           Data.Time.Clock            (getCurrentTime)
 import qualified Hasql.Pool                 as P
-import qualified Hasql.Session              as H
 import qualified Hasql.Transaction.Sessions as HT
 import           Network.Wai.Handler.Warp   (defaultSettings,
                                              runSettings, setHost,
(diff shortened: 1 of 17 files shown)


Qu4tro:
Added commit that resolves #1285

main/Main.hs
```diff
@@ -74,25 +73,23 @@ connectionWorker mainTid pool schema refDbStructure refIsWorkerOn = do
...
+        NotConnected                -> return ()               -- Unreachable
+        Connected actualPgVersion   -> do                      -- Procede with initialization
+          result <- P.use pool $ do
+            dbStructure <- HT.transaction HT.ReadCommitted HT.Read $ getDbStructure schema actualPgVersion
+            liftIO $ atomicWriteIORef refDbStructure $ Just dbStructure
+          case result of
+            Left e -> do
+              putStrLn ("Failed to query the database. Retrying." :: Text)
+              hPutStrLn stderr . toS . errorPayload $ PgError False e
+
```
steve-chavez: There was a `work` in this line in the previous version. Better leave it to retry just in case.

src/PostgREST/Error.hs
```diff
@@ -127,69 +108,151 @@ instance JSON.ToJSON H.QueryError where
...
+        'P':'T':n -> fromMaybe HT.status500 (HT.mkStatus <$> readMaybe n <*> pure m)
+        _         -> HT.status400
+
+    _                       -> HT.status500
+
+checkIsFatal :: PgError -> Maybe Text
+checkIsFatal (PgError _ (P.ConnectionError e))
+  | isAuthFailureMessage = Just "Password authentication failed"
+  | otherwise = Nothing
+  where isAuthFailureMessage = "FATAL:  password authentication failed" `isPrefixOf` toS (fromMaybe "" e)
```
steve-chavez: Instead of parsing the error response from PostgreSQL, could the error code somehow be obtained?

It should be `28P01` according to https://www.postgresql.org/docs/current/errcodes-appendix.html.

src/PostgREST/Error.hs
```diff
@@ -127,69 +108,151 @@ instance JSON.ToJSON H.QueryError where
...
+        "42P01"   -> HT.status404 -- undefined table
+        "42501"   -> if authed then HT.status403 else HT.status401 -- insufficient privilege
+        'P':'T':n -> fromMaybe HT.status500 (HT.mkStatus <$> readMaybe n <*> pure m)
+        _         -> HT.status400
+
+    _                       -> HT.status500
+
+checkIsFatal :: PgError -> Maybe Text
+checkIsFatal (PgError _ (P.ConnectionError e))
+  | isAuthFailureMessage = Just "Password authentication failed"
```
steve-chavez: It'd be better to leave the original pg error, could have more details.

src/PostgREST/Types.hs
```diff
@@ -413,3 +405,11 @@ sourceCTEName = "pg_source"
 type JSPath = [JSPathExp]
 -- | jspath expression, e.g. .property, .property[0] or ."property-dash"
 data JSPathExp = JSPKey Text | JSPIdx Int deriving (Eq, Show)
+
+
+
+-- | Current database connection status data ConnectionStatus
```
steve-chavez: Missed `data ConnectionStatus =` line.

src/PostgREST/Error.hs
```diff
@@ -29,73 +27,41 @@ import           Network.Wai               (Response, responseLBS)
...
- "response.headers guc must be a JSON array composed of objects with a single key and a string value"
-
-connectionLostError :: Response
-connectionLostError =
-  simpleError HT.status503 [] "Database connection lost, retrying the connection."
-
-encodeError :: JSON.ToJSON a => a -> LByteString
-encodeError = JSON.encode
+
+class (JSON.ToJSON a) => PostgRestError a where
```
steve-chavez: `PostgRestError` could be shortened to `PgrstError`. We use `pgrst` as a shortand for PostgREST in some places.

steve-chavez:
@Qu4tro Good idea centralizing the error messages in `Error.hs` and adding more details to the test suite.

See my comments and make sure to leave the test suite green so I can give it a second review.

steve-chavez:
@Qu4tro I was thinking to merge this PR before your other ones since it contains a big change.

Hope you can find some time soon to finish this one. Let me know if I can help in any way.

Qu4tro:
Hello there. I was offline for pretty much all of last week. I'm trying to catch up with everything now. I'll try to pick it up between today and tomorrow.

src/PostgREST/Error.hs
Qu4tro: I think we can't do it with the current interface of `Hasql.Pool` `use`. I think we would have to go a bit over hasql internals for this.

`use` returns an Either (as usual, left error, right data). The error given is an `UsageError`, which is itself a union type that encapsulates either `ConnectionError` or `QueryError`. 

There's also a bit of type rabbit-hole on `QueryError`, where eventually error codes may live, but the point is that bad password is a `ConnectionError`, which is simply defined as `type ConnectionError = Maybe ByteString`. The string doesn't contain any code.

Example with `stack ghci`:
```haskell
> import qualified Hasql.Pool as P
> P.acquire (1, 10, "postgres://goodUser:badPasswor@goodHost:goodPort/goodDbName") >>= (\pool -> P.use pool getPgVersion)

< Left (ConnectionError (Just "FATAL:  password authentication failed for user \"goodUser\"\nFATAL:  no pg_hba.conf entry for host \"goodHost\", user \"goodUser\", database \"goodDbName\", SSL off\n"))
```

steve-chavez: @Qu4tro True. I've just noticed that. Parsing the error should be good for now.

steve-chavez: Just opened an issue for exposing the error code in `ConnectionError`: https://github.com/nikita-volkov/hasql/issues/112.

Qu4tro:
Okay, I think all issues with this one are fixed. Let me know if there was something else that needs changing.

##### Not incredibly related to PR
One good thing, we might add, is the setup necessary for automatic styling. I'm thinking a small pre-commit config and CircleCI check.
Consistent styling across commits would help with the reading of some bigger diffs and `.stylish-haskell.yaml` already exists. TBH I don't if it needs tweaking either, but I'm fairly sure that's what my autoformatter is picking up on.

steve-chavez:
@Qu4tro LGTM. Add an entry to the CHANGELOG for #1285 and I'll merge.

It'd be nice to have a bash test for the db retrying, doing `systemctl stop/start postgresql` and checking it reconnects. But there wasn't a previous test for this so we can leave it for another PR(this one is getting a bit long).

Regarding stylish haskell, I have used it sometimes(manually), automating that would be great.

Qu4tro:
Added entry to changelog.

PR #1284, opened 2019-04-26
Fix reference to deprecated SIGHUP call

dsernst:
This updates the [`CONTRIBUTING.md`](https://github.com/PostgREST/postgrest/blob/e2bd91623777b2ade2ddd78f48fa7cc115556533/.github/CONTRIBUTING.md) instructions about sending a `SIGHUP` call (deprecated) to `SIGUSR1`, and links to relevant docs: http://postgrest.org/en/v5.2/admin.html#schema-reloading

Per the docs:

> As of PostgREST v5.1 reloading with SIGHUP is deprecated, it’s still supported but will be removed in v6.0. SIGUSR1 should be used instead.

See [5.1.0 CHANGELOG](https://github.com/PostgREST/postgrest/blob/d32f373e1ee95ba68141a3f3a48ab8c94ea2453a/CHANGELOG.md#deprecated) or https://github.com/PostgREST/postgrest/issues/724

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -32,7 +32,7 @@ your contributions.
   then [find your logs](http://blog.endpoint.com/2014/11/dear-postgresql-where-are-my-logs.html).
 
 * If your database schema has changed while the PostgREST server is running,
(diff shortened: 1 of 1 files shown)


steve-chavez:
@dsernst Thank you!

PR #1286, opened 2019-04-28
Haskell memory tests

Qu4tro:
Hey there,

This PR creates a new test-suit. This test suit is meant to measure memory-usage in the likes of the current `memory-tests.sh`. It uses the [weigh](https://github.com/fpco/weigh) package to provide the necessary tooling.

One thing of note is that the results don't match to bash tests. I'm not sure why that it is, yet.

Additionally and TODO: Check status code of requests.

### DIFF
--- a/test/Memory.hs
+++ b/test/Memory.hs
@@ -1,20 +1,19 @@
 module Main where
 
+import           Data.Aeson           ((.=))
 import qualified Data.Aeson           as JSON
-import           Data.Map             (fromList)
 import           Network.Wai          (Application)
 import           SpecHelper           (posgrestTestApp, requestAction,
-                                       testCfg)
+                                       testCfgMemory)
 
-import           Data.ByteString.Lazy (ByteString, replicate)
+import           Data.ByteString.Lazy as BS hiding (take)
 import           Network.Wai.Test     (SResponse)
 import           Test.Hspec.Wai       (WaiSession, get, patch, post)
 import           Weigh                (Weigh, mainWith, maxAllocs,
-                                       validateAction)
+                                       validateAction, wgroup)
 
 
-import           Protolude            hiding (ByteString, get,
-                                       replicate)
+import           Protolude            hiding (get)
 
 
 newtype MegaBytes = MegaBytes Double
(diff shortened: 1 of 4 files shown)


Qu4tro:
Tests are currently failing, due to different measurements, which I guess stem from hardware differences. This could be easily changed to accommodate Circle.

On both circle and locally the results seems to be consistent, though. I've also tried `circleci local execute --job build-test-9.4` and the tests pass.

Keeping it that way for until further discussion.

steve-chavez:
@Qu4tro Cool usage of weigh. Is the intention to replace the bash test suite?

Also, I've ran the new test suite several times and somehow it takes like 5-6 minutes to complete. Does the same happen to you?

Qu4tro:
Yes, the intention would be to get rid of the bash tests. I think having them in Haskell, would incentivize the creation of more memory tests.

Do you mean both test-suits or just the memory one? 
I got 4 minutes for both (and some compiling) and just under a minute with the memory suit alone. 
Same thing for both docker/local postgres.

steve-chavez:
Only running `stack test postgrest:memory` takes 5-6 minutes for me. Not sure why.

test/Memory.hs
```diff
@@ -0,0 +1,107 @@
...
+jsonKeyTest1M :: MemoryCaseTester -> Weigh ()
+jsonKeyTest1M test = wgroup "json key test - 1M blob" $ do
+    test "post rpc request"
+         (post "/rpc/leak?columns=blob"           $ jsonBlob 1) $
+         MegaBytes 10
+    test "post request"
+         (post "/leak?columns=blob"               $ jsonBlob 1) $
+         MegaBytes 10
+    test "patch request"
+         (patch "/leak/leak?id=eq.1&columns=blob" $ jsonBlob 1) $
```
steve-chavez: `/leak/leak?id=eq..` is wrong here. Should be just `/leak?id=eq..`

Qu4tro:
Ah thanks. Will fix.

I should also be able to get the status codes of the responses, so that these mistakes are easily found .

test/Memory.hs
Qu4tro: Fixing this also made the patch requests consume around the same memory as the other requests resolving that question.

Qu4tro:
@steve-chavez if it helps, the output of inxi on my laptop:
```
❯ inxi -b
System:
  Host: thinkpad4 Kernel: 5.0.10-arch1-1-ARCH x86_64 bits: 64 
  Desktop: i3 4.16.1-107-g41264e54 Distro: Arch Linux 
Machine:
  Type: Laptop System: LENOVO product: 20LS001APG v: ThinkPad L480 
  serial: <root required> 
  Mobo: LENOVO model: 20LS001APG v: SDK0J40697 WIN serial: <root required> 
  UEFI: LENOVO v: R0QET41W (1.18 ) date: 04/03/2018 
Battery:
  ID-1: BAT0 charge: 39.7 Wh condition: 39.7/45.0 Wh (88%) 
CPU:
  Quad Core: Intel Core i5-8250U type: MT MCP speed: 800 MHz min/max: 400/3400 MHz 
Graphics:
  Device-1: Intel UHD Graphics 620 driver: i915 v: kernel 
  Display: tty server: X.Org 1.20.4 driver: i915 resolution: 1920x1080~60Hz 
  OpenGL: renderer: Mesa DRI Intel UHD Graphics 620 (Kabylake GT2) 
  v: 4.5 Mesa 19.0.3 
Network:
  Device-1: Intel Ethernet I219-V driver: e1000e 
  Device-2: Intel Wireless 8265 / 8275 driver: iwlwifi 
Drives:
  Local Storage: total: 238.47 GiB used: 114.55 GiB (48.0%) 
Info:
  Processes: 215 Uptime: 10m Memory: 15.43 GiB used: 2.98 GiB (19.3%) Shell: zsh 
  inxi: 3.0.33 
```

steve-chavez:
We're now getting a lot of [out of memory errors](https://circleci.com/gh/PostgREST/postgrest/3582?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) when compiling with `stack build --profile`.

@Qu4tro I wanted to ask you, using weigh requires compiling with [profiling](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/profiling.html) enabled?

From this PR I'm left with the impression that it's not required. I'm wondering how `weigh` works in that case. Perhaps you can elaborate on this.

steve-chavez:
Hmm. This kind of explains that weigh doesn't need ghc profiling enabled https://github.com/fpco/weigh/issues/14.

steve-chavez:
Still like the idea of having the memory tests in Haskell. The code here is really good so I'll leave the PR open for someone to pick it up.

steve-chavez:
An all-haskell test suite doesn't look close at the moment: we even have [IO tests in python](https://github.com/PostgREST/postgrest/blob/5c75f0dcc295e6bd847af6d9703fad5b9c3d76c9/test/io-tests/test_io.py) now.

We might get more attempts at this after https://github.com/PostgREST/postgrest/issues/1804 is done.

PR #1288, opened 2019-05-01
Breaking changes

steve-chavez:
Breaking changes for the upcoming version. 

- Remove SIGHUP
- Change server-host default of 127.0.0.1 to !4
- Change symbol for disambiguating resource embedding(undocumented featured added in #918) from dot '.' to plus '+'. Used to be `/clients?select=id,projects:projects.client_id(id,tasks(id))`, now is `/clients?select=id,projects:project+client_id(id,tasks(id))`.
  + Dot is not clear since we already use the dot '.' for representing embedded filters.
  + Dot in pg implies dot notation(e.g. `table.col`) which is not the case when we use the dot for disambiguating many to many relationships.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,15 +91,6 @@ jobs:
       - run:
           name: run linter
           command: git ls-files | grep '\.l\?hs$' | xargs stack exec -- hlint -X QuasiQuotes -X NoPatternSynonyms "$@"
-      - run:
-          name: extra checks
-          command: |
-            stack exec -- cabal update
-            stack exec --no-ghc-package-path -- cabal install --only-d --dry-run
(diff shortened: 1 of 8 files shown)


ruslantalpa:
`.` and `+` should both work for a period of time (about a year/till 6.x) so that ppl can upgrade and slowly change their rest call to reflect the new format

steve-chavez:
I just considered removing `.`  because the feature was never documented AFAICT, but I'm not opposed to instead deprecate it.

steve-chavez:
Some extra cabal checks were making CircleCI give an out of memory error. Not sure why this is happening now and I haven't figured out how to solve it(`-j1` doesnt' work https://github.com/haskell/cabal/issues/5505), so I'm just removing the extra checks for now.

PR #1289, opened 2019-05-01
Re-read configuration and schema on SIGUSR2

### LINKED ISSUE #1119: [Question] SIGHUP behavior to reload jwt-secret
### Environment

* PostgreSQL version: 10
* PostgREST version: (if using docker, specify the image) 4.4
* Operating system: Linux

### Description of issue

(Expected behavior vs actual behavior)
I source my JWT secret from a file:

```
## choose a secret to enable JWT auth
## (use "@filename" to load from separate file)
jwt-secret = "@/secrest/jwt.secret"
```

When I change this file and send Postgrest a SIGHUP, it reloads itself:

```
Listening on port 3000
Connection successful
```

However I don't think it re-reads the contents of the JWT secret file. Is it supposed to?

My goal is to rotate the JWT signing key (for a sort of primitive "revocation" of pre-existing tokens).

Qu4tro:
This adds a signal handler that allows re-reading configuration and schema, by closing all open connections and restarting.

One thing I noticed, is that if the server is inside `connectionWorker` trying to connect, there still won't be a connection to close and it will keep its attempts.

Resolves #1119

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -23,7 +23,8 @@ import Network.Wai.Handler.Warp (defaultSettings, runSettings,
                                  setGracefulShutdownTimeout, setHost,
                                  setInstallShutdownHandler, setPort,
                                  setServerName)
-import System.IO                (BufferMode (..), hSetBuffering)
+
+import System.IO (BufferMode (..), hSetBuffering)
 
 import PostgREST.App         (postgrest)
 import PostgREST.Config      (AppConfig (..), configPoolTimeout',
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -20,7 +20,9 @@ import Data.Text.Encoding       (decodeUtf8, encodeUtf8)
 import Data.Text.IO             (hPutStrLn, readFile)
 import Data.Time.Clock          (getCurrentTime)
 import Network.Wai.Handler.Warp (defaultSettings, runSettings,
-                                 setHost, setPort, setServerName)
+                                 setGracefulShutdownTimeout, setHost,
+                                 setInstallShutdownHandler, setPort,
+                                 setServerName)
 import System.IO                (BufferMode (..), hSetBuffering)
 
 import PostgREST.App         (postgrest)
@@ -34,7 +36,6 @@ import PostgREST.Types       (ConnectionStatus (..), DbStructure,
                               PgVersion (..), Schema,
                               minimumPgVersion)
 import Protolude             hiding (hPutStrLn, replace)
-
 
 #ifndef mingw32_HOST_OS
 import System.Posix.Signals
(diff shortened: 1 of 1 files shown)


steve-chavez:
@Qu4tro Good idea having a different signal(`SIGUSR2`) for the re-reading.

> One thing I noticed, is that if the server is inside connectionWorker trying to connect, there still won't be a connection to close and it will keep its attempts.

Yes, as `connectionWorker` spawns a new thread. In fact I've done many `kill -USR2` and it spawns many connectionWorkers each one trying to reconnect and they finish at different times. 

---

Now, regarding #1119, I think the usefulness of having signals is to help reducing downtime. With the current `SIGUSR2` handler, where not offering something that `systemctl restart postgrest` wouldn't do.
 
Reloading `jwt-secret` shouldn't require a reconnection to the database, nor reloading some of the other settings that are not related to the connection.

So, ideally I think `SIGURS2` should only re-read these settings:

- db-schema
- db-anon-role
- server-proxy-uri
- jwt-secret
- secret-is-base64
- jwt-aud
- max-rows
- pre-request
- role-claim-key
- db-extra-search-path

The settings that are not worth re-reading(would require a `restart`) would be:

- db-uri
- db-pool
- db-pool-timeout
- server-host
- server-port

Not sure if it's too complex to code that, but it would be the ideal behavior. What do you think?

Qu4tro:
@steve-chavez Yep, that makes sense.

And what if the new config is not a valid one? Should we keep the old values?

steve-chavez:
@Qu4tro On second thought, we should keep the PR focused to #1119 and just reload the JWT settings on SIGUSR2, these would be:

- jwt-secret
- secret-is-base64
- jwt-aud
- role-claim-key

If some of these values are invalid we could just let the JWT Middleware fail with JWTMissingSecret, JWTInvalid, etc.

steve-chavez:
@Qu4tro Hey! :smiley:. I'm going to release a new version in the next few weeks. It'd be great to have this feature. Let me know if I can help you finishing this PR.

I was thinking we should store the config file path when postgrest is started, and then use that path to reload the config. If the config is missing from that path, then we quit with an error.

Qu4tro:
Hey there. 

Sorry I haven't picked up the work on this PR. The few different ways of doing this and my unease around signals and haskell multi-threading added some friction. 

Yeah, that sounds good.

steve-chavez:
PR contains a useful discussion on the implementation but the code is out dated now. Closing..

awalias:
Hi guys, was this feature of re-loading config on `SIG<X>` implemented in the end? Particularly interested in reloading `db-schema=` without having to restart postgrest

steve-chavez:
@awalias This one didn't got implemented. Regarding db-schema, there's a discussion on https://github.com/PostgREST/postgrest/issues/1521.

PR #1291, opened 2019-05-13
Update Dockerfile base image

russelldavies:
Debian Jessie is quite old now and causes some issues with the `sources.list` when extending the image. `debian-stretch` also works in case there is a preference for that.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:jessie
+FROM ubuntu:18.04
 
 ARG POSTGREST_VERSION
 


steve-chavez:
@russelldavies Also ran into this issue https://github.com/PostgREST/postgrest/pull/1262#issuecomment-479960205 with `jessie`. I was thinking we should switch to `debian:stretch-slim`(lighter image).

russelldavies:
In that case, I'd go for `debian:stretch-slim` if it's lighter and there's no functional difference. Closing because this is a duplicate of #1262.

steve-chavez:
@russelldavies If you'd like you could make the change to `stretch-slim` in this PR and I'll merge. #1262 hasn't been updated for a while.

russelldavies:
@steve-chavez Can you re-open this issue, please. It seems that because it's closed Github won't see that I've updated the branch. Thanks.

ruslantalpa:
github does not allow it, the button is grayed out and says "branch was force-pushed or recreated"

russelldavies:
From reading a few other posts about this situation there doesn't seem to be much I can do apart from create a new pull request. If no one has any objections, I'll do that.

steve-chavez:
@russelldavies Opening a new PR would be great. Thank you.

Issue #1186, opened 2018-09-29
Unix socket listening support request


akemrir:
### Environment

* PostgreSQL version: 10.5
* PostgREST version: 5.1.0 (3cef4b7)
* Operating system: Archlinux

### Description of issue

Right now the only way to send requests to PostgREST is to make tcp call over port.
There is no way to do it via unix socket. To utilize it with nginx for example.

There are many benchmarks proving that unix socket is more efficient than tcp connection.
Maybe this options could make it`s way into PostgREST?

For example with new listening option in configuration:
`server-unix-socket = unix:/tmp/socktest.sock`

ruslantalpa:
nginx can very well be in front of postgrest as a reverse proxy, and as for efficiency ... optimising this is almost irrelevant, in the total time it takes to process a request, the tcp/http overhead between nginx/postgrest on the same box i would say is well under 1% (just a feeling). 
You will better spend your time optimising indexes on your db :)

steve-chavez:
A single box with postgresql-unixsocket-postgrest-unixsocket-nginx looks like a solid production setup, would be interesting to benchmark it and see if the unix socket performance boost is really that negligible.

ruslantalpa:
i benchmarked (in aws, on small boxes with limited resources) running a query on the db directly and trough postgrest. the entire postgrest in the chain adds 2ms, imagine how much of that time is the tcp overhead... and now think that an average query might be say 50-100ms, add in the network time (browser to postgrest) of 50-100ms then you end up optimising 0.2ms out of a 100ms request ...
in terms of CPU used on the box ... a few cents a month maybe?

this is really a feature for setups with really high loads that are trying to get the most throughput out of their hardware and in that case maybe they should sponsor it? :)

steve-chavez:
Just saw how to implement this and I think it should be a really simple change!

Currently we use [runSettings](http://hackage.haskell.org/package/warp-1.2.1.1/docs/Network-Wai-Handler-Warp.html#v:runSettings) here https://github.com/PostgREST/postgrest/blob/master/main/Main.hs#L217.
There's another function [runSettingsSocket](http://hackage.haskell.org/package/warp-1.2.1.1/docs/Network-Wai-Handler-Warp.html#v:runSettingsSocket) that receives a Socket as a parameter. We could have a `server-socket` config and apply that function conditionally.

A PR would be welcomed :)

Dansvidania:
Hello, I would like to work on this if its ok.

[...]

PR #1293, opened 2019-05-14
added conditional use of runSettingsSocket in Main.hs

Dansvidania:
#1186 added unix socket config option in config module and example config file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1186, Add support for user defined unix socket
 - #690, Add `?columns` query parameter for faster bulk inserts, also ignores unspecified json keys in a payload - @steve-chavez
 - #1239, Add support for resource embedding on materialized views - @vitorbaptista
 - #1264, Add support for bulk RPC call - @steve-chavez
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -4,38 +4,52 @@ module Main where
 
 
 import           PostgREST.App              (postgrest)
-import           PostgREST.Config           (AppConfig (..), configPoolTimeout',
-                                             prettyVersion, readOptions)
-import           PostgREST.DbStructure      (getDbStructure, getPgVersion)
+import           PostgREST.Config           (AppConfig (..),
+                                             configPoolTimeout',
+                                             prettyVersion,
+                                             readOptions)
+import           PostgREST.DbStructure      (getDbStructure,
+                                             getPgVersion)
 import           PostgREST.Error            (encodeError)
 import           PostgREST.OpenAPI          (isMalformedProxyUri)
-import           PostgREST.Types            (DbStructure, Schema, PgVersion(..), minimumPgVersion)
-import           Protolude                  hiding (hPutStrLn, replace)
+import           PostgREST.Types            (DbStructure,
+                                             PgVersion (..), Schema,
+                                             minimumPgVersion)
+import           Protolude                  hiding (hPutStrLn,
+                                             replace)
 
 
 import           Control.AutoUpdate         (defaultUpdateSettings,
-                                             mkAutoUpdate, updateAction)
+                                             mkAutoUpdate,
+                                             updateAction)
+import           Control.Exception          (catch)
 import           Control.Retry              (RetryStatus, capDelay,
                                              exponentialBackoff,
-                                             retrying, rsPreviousDelay)
+                                             retrying,
+                                             rsPreviousDelay)
 import qualified Data.ByteString            as BS
 import qualified Data.ByteString.Base64     as B64
 import           Data.IORef                 (IORef, atomicWriteIORef,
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1186, Add support for user defined unix socket
```
steve-chavez: Would be better to specify the new `server-unix-socket` config option.

main/Main.hs
```diff
@@ -213,21 +227,33 @@ main = do
...
-      refDbStructure
-      pool
-      getTime
-      (connectionWorker
-         mainTid
-         pool
-         (configSchema conf)
-         refDbStructure
-         refIsWorkerOn)
+  let postgrestApplication = postgrest
```
steve-chavez: nitpick: To reduce indentation could you move the variable assignment to the next line? Like:
```haskell
let postgrestApplication =
  postgrest
    conf
    ...
```

src/PostgREST/Config.hs
```diff
@@ -139,6 +142,7 @@ readOptions = do
           <*> C.key "db-schema"
           <*> (fromMaybe "127.0.0.1" . mfilter (/= "") <$> C.key "server-host")
           <*> (fromMaybe 3000 . join . fmap coerceInt <$> C.key "server-port")
+          <*> C.key "server-unix-socket"
```
steve-chavez: I've noticed you can pass a config like `server-unix-socket=""` and it works. Do you think that should be allowed?

I've tested that with `echo -e "GET / HTTP/1.0\n" | nc -U ""` and the output seems correct.

steve-chavez: Further research lead me to this https://stackoverflow.com/questions/29427769/how-to-use-unix-domain-socket-without-creating-a-socket-file.

So perhaps we should allow empty value as a unix socket.

main/Main.hs
```diff
@@ -213,21 +227,33 @@ main = do
...
+                                   refIsWorkerOn)
+   in case maybeSocketAddr of
+        Nothing -> do
+            -- run the postgrest application
+            putStrLn $ ("Listening on port " :: Text) <> show (configPort conf)
+            runSettings appSettings postgrestApplication
+        Just socketAddr -> do
+            sock <- createAndBindSocket socketAddr
+            listen sock maxListenQueue
+            putStrLn $ ("Listening on socket " :: Text) <> show socketAddr
```
steve-chavez: I think `Listening on unix socket` would be more precise.

src/PostgREST/Config.hs
```diff
@@ -223,6 +227,7 @@ readOptions = do
           |
           |## base url for swagger output
           |# server-proxy-uri = ""
+          |# server-unix-socket = ""
```
steve-chavez: Could be confusing to offer the `""` value(since it's valid as I mentioned above) as an example. Perhaps the example should be: ` server-unix-socket = pgrst.sock`.

Also, when `server-port` and `server-unix-socket` are specified we prefer the unix socket. Perhaps you could add a comment to the config like: `If specified, server-unix-socket has precedence over server-port`.

steve-chavez:
@Dansvidania Sorry for the late review. Code looks good!

Regarding the failed test don't worry about it. If you rebase against master it'll be fixed.

Adding a test case for unix socket would be great, but I'm not sure how to use the unix socket with [hspec](http://hackage.haskell.org/package/hspec-wai-0.9.2/docs/Test-Hspec-Wai.html).

Another option would be to add a bash test to [io-tests.sh](https://github.com/PostgREST/postgrest/blob/master/test/io-tests.sh), checking that a request like `curl --unix-socket pgrst.sock http://localhost/projects ` succeeds would be enough.

main/Main.hs
```diff
@@ -290,6 +316,19 @@ loadDbUriFile conf = extractDbUri mDbUri
...
+
+createAndBindSocket :: FilePath -> IO Socket
+createAndBindSocket filePath = do
+  deleteSocketFileIfExist filePath
+  sock <- socket AF_UNIX Stream defaultProtocol
+  bind sock $ SockAddrUnix filePath
+  return sock
+  where
+    deleteSocketFileIfExist path = removeFile path `catch` handleDoesNotExist
+      where
```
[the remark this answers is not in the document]
Dansvidania: very true. removed.

src/PostgREST/Config.hs
Dansvidania: I did not know anything about sockets before so no opinion regarding this. I have corrected the example but will leave the empty string as a valid option to allow the usage shown on stack overflow.

Thanks for taking the time to research this.

src/PostgREST/Config.hs
Dansvidania: Thanks for the feedback. Both good points. I will update the PR asap.

Dansvidania:
messed the original repository up, will open anew

PR #1299, opened 2019-05-21
Update stylish haskell config; apply all; add CI config

Qu4tro:
Added a few things in support of using stylish-haskell and keeping using it.

Commits:
 - Apply stylish-haskell to all files.
 - Update config default with `stylish-haskell --defaults > .stylish-haskell.yaml`. Kept non-defaults: `columns: 70` and `language_extensions` .
- Remove redundant import. Re-ordering the imports revealed this. What is used from `Network.HTTP.Types.Headers` are also exported by `Network.HTTP.Types`.
- Add CircleCI config. Since `stylish-haskell` doesn't have an option to check format correctness (https://github.com/jaspervdj/stylish-haskell/issues/107), I use a small hack: Run with `-i` to format all source files inplace and then check if the git repo has any changes (the formatted files).

Edit:
 - Last commit was to match stylish-haskell version with our setup. Instead of the original command, used: `stack exec -- stylish-haskell --defaults > .stylish-haskell.yaml`. Kept non-defaults.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ jobs:
             sudo apt-get install -y postgresql-client
             stack setup
             rm -rf $(stack path --dist-dir) $(stack path --local-install-root)
-            stack install hlint packdeps cabal-install
+            stack install hlint packdeps cabal-install stylish-haskell
       - run:
           name: build src and tests
           command: |
@@ -91,6 +91,9 @@ jobs:
       - run:
           name: run linter
           command: git ls-files | grep '\.l\?hs$' | xargs stack exec -- hlint -X QuasiQuotes -X NoPatternSynonyms "$@"
+      - run:
+          name: run styler
+          command: git ls-files | grep '\.l\?hs$' | xargs stack exec -- stylish-haskell -i && git diff-index --quiet HEAD --
       - save_cache:
           paths:
             - "~/.stack"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ jobs:
           command: git ls-files | grep '\.l\?hs$' | xargs stack exec -- hlint -X QuasiQuotes -X NoPatternSynonyms "$@"
       - run:
           name: run styler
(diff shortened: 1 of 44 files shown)


Qu4tro:
To solve conflicts I've created a new branch off of master, where I cherry picked all the commits from this branch (slightly different order) except the actual formatting commit. Replaced that commit with a new one where I ran the fmt command. And push -f.

Qu4tro:
Also see [here](https://github.com/Qu4tro/postgrest/pull/1) for a suggestion on changing the import align option to `group` instead of `global`. 

I've grouped imports into: `import qualified`, `import Module (a, b, c)`, `import Module`, local modules and protolude. It was mostly to have a standard way to arrange them.  
It also means imports aren't all aligned with the qualified keyword, which didn't help readability. 

I've setup the PR to merge to this branch, so lmk what you think.

.circleci/config.yml
```diff
@@ -91,6 +91,9 @@ jobs:
       - run:
           name: run linter
           command: git ls-files | grep '\.l\?hs$' | xargs stack exec -- hlint -X QuasiQuotes -X NoPatternSynonyms "$@"
+      - run:
+          name: run styler
+          command: git ls-files | grep '\.l\?hs$' | xargs stack exec -- stylish-haskell -i && git diff-index --quiet HEAD --
```
steve-chavez: Maybe use `--exit-code` instead of  `--quiet`. This is a new workflow for me, if there's an error I'd like to know in which file there's a discrepancy.

steve-chavez:
@Qu4tro Looks good!

I don't have a strong opinion on `global` vs `group`. I'll let you decide that one.
I'm already happy with enforcing a shared style for the codebase .

.circleci/config.yml
Qu4tro: :100:

Qu4tro:
Ok, I'll merge the PR to this branch then. 

Like I've mentioned before, IMO the import section becomes more readable. We might rethink the actual grouping method I picked, later on, but I think it's a improvement at the moment.

PR #1303, opened 2019-05-26
Add server-unix-socket option to enable trafic to postgrest server via unix socket

Dansvidania:
Targeting #1186 

Hello, I have corrected the issues from my previous pull request and had to make some changes to make it merge-able after the styling changes

The only thing that is missing is a test in io-test.sh, which I am trying to work on. If anyone is able to give me some pointers on how the bash tests are organized that would be very helpful.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,11 @@ jobs:
           keys:
           - v1-stack-dependencies-{{ checksum "postgrest.cabal" }}-{{ checksum "stack.yaml" }}
       - run:
+          name: install ncat
+          command: |
+            # utility needed to test socket connection with curl <  7.40
+            sudo apt-get install nmap
+      - run:
           name: install stack & dependencies
           command: |
             curl -L https://github.com/commercialhaskell/stack/releases/download/v1.1.2/stack-1.1.2-linux-x86_64.tar.gz | tar zx -C /tmp
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1186, Add support for user defined unix socket via `server-unix-socket` config option
 - #690, Add `?columns` query parameter for faster bulk inserts, also ignores unspecified json keys in a payload - @steve-chavez
 - #1239, Add support for resource embedding on materialized views - @vitorbaptista
 - #1264, Add support for bulk RPC call - @steve-chavez
--- a/test/io-tests.sh
+++ b/test/io-tests.sh
@@ -181,6 +181,30 @@ ensureAppSettings(){
   pgrStop
 }
 
+getSocketStatus() {
+  curl -sL -w "%{http_code}\\n" -o /dev/null localhost:54321
+}
+
+socketConnection(){
+  # map port 54321 traffic to unix socket as workaround for curl below 7.40
+  # not supporting --unix-socket flag
+  ncat -vlk 54321 -c 'ncat -U /tmp/postgrest.sock' &
+  pgrStart "./configs/unix-socket.config"
+  while pgrStarted && test "$( getSocketStatus )" -ne 200
+  do
+    # wait for the server to start
+    sleep 0.1 \
+    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+  done
+  if test $( getSocketStatus ) -eq 200
+  then
+    ok "Succesfully connected through unix socket"
+  else
+    ko "Failed to connect through unix socket"
+  fi
+  pgrStop
+}
+
 # PRE: curl must be available
 test -n "$(command -v curl)" || bailOut 'curl is not available'
 
@@ -190,6 +214,8 @@ psql -l 1>/dev/null 2>/dev/null || bailOut 'postgres is not running'
 setUp
 
 echo "Running IO tests.."
+
+socketConnection
 
 readSecretFromFile word.noeol 'simple (no EOL)'
 readSecretFromFile word.txt 'simple'
@@ -223,6 +249,7 @@ invalidRoleClaimKey 1234
 
 ensureIatClaimWorks
 ensureAppSettings
+
 
 cleanUp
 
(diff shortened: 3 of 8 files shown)


Qu4tro:
Hey. Sorry about those changes. I'm sure they conflicted with all PRs that were open ahah. 

You can see the function `ensureAppSettings()` on https://github.com/PostgREST/postgrest/blob/master/test/io-tests.sh#L165 . You can basically take that as a template, change the config file for a new one with a `server-unix-socket` key, modify the curl command and check if the results are as expected. I think a simple request should do fine, as the purpose of the test is to determine if the server is reachable and responding in parity to normal usage. Finally, call this new function before `cleanUp` and you should be gtg.

Dansvidania:
No worries actually I also use stylish-haskell and I was worrying I would have to revert all those lines before the PR

I have tried to do just that and then run the circleCi jobs to test, but I am encountering some problems with my bash code (file not found, --unix-socket option not avaliable)

On the other hand, I am having issues running the tests by just running the bash file. I keep getting psql auth bail outs. Am I missing something silly?

Dansvidania:
It seems like the curl version in circleci images is below 7.40 and the --unix-socket option is not available.

what is the preferred way to fix this? would it be ok to download a newer version of curl in the command section on .circleci/config.yaml ?

Qu4tro:
Yep, I don't see any problem with testing with a more recent version of curl.

.circleci/config.yml
```diff
@@ -66,6 +66,11 @@ jobs:
       - restore_cache:
           keys:
           - v1-stack-dependencies-{{ checksum "postgrest.cabal" }}-{{ checksum "stack.yaml" }}
+      - run:
+          name: install ncat
```
steve-chavez: Is `nc` available in CircleCI? Maybe we could omit this dependency, try:

```bash
echo -e "GET / HTTP/1.0\n" | nc -U "/tmp/postgrest.sock"
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1186, Add support for user defined unix socket via server-unix-socket config option
```
steve-chavez: Add surrounding backticks for  `server-unix-socket` so it's more noticeable.

test/io-tests.sh
```diff
@@ -181,6 +181,31 @@ ensureAppSettings(){
   pgrStop
 }
 
+getSocketStatus() {
+  curl -sL -w "%{http_code}\\n" -o /dev/null localhost:54321
```
steve-chavez: If `nc` is available. Then you could do:
```bash
  echo -e "GET / HTTP/1.0\n" | nc -U "/tmp/postgrest.sock" | head -n 1 | awk '{ print $2 }'
```

test/io-tests.sh
```diff
@@ -181,6 +181,31 @@ ensureAppSettings(){
...
+}
+
+socketConnection(){
+  # map port 54321 traffic to unix socket as workaround for old curl version
+  # not supporting --unix-socket flag
+  ncat -vlk 54321 -c 'ncat -U /tmp/postgrest.sock' &
```
steve-chavez: This wouldn't be needed if the above works(`nc`).

test/io-tests.sh
```diff
@@ -181,6 +181,31 @@ ensureAppSettings(){
...
+  pgrStart "./configs/unix-socket.config"
+  while pgrStarted && test "$( getSocketStatus )" -ne 200
+  do
+    # wait for the server to start
+    sleep 0.1 \
+    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+  done
+  sleep 11
```
steve-chavez: `sleep 11` shouldn't be needed if the `while` above is working, try removing it.

test/io-tests.sh
```diff
@@ -181,6 +181,31 @@ ensureAppSettings(){
...
+getSocketStatus() {
+  curl -sL -w "%{http_code}\\n" -o /dev/null localhost:54321
+}
+
+socketConnection(){
+  # map port 54321 traffic to unix socket as workaround for old curl version
+  # not supporting --unix-socket flag
+  ncat -vlk 54321 -c 'ncat -U /tmp/postgrest.sock' &
+  pgrStart "./configs/unix-socket.config"
+  while pgrStarted && test "$( getSocketStatus )" -ne 200
```
steve-chavez: If the change to `nc` is possible, maybe `test -ne` could fail, you could try `test "$( getSocketStatus )" != "200"`

.circleci/config.yml
Dansvidania: @steve-chavez I have tried this; for some reason this line works from the command line but I keep getting 400 bad requests when I execute it in a script. 

I have found a reference to [this askubuntu question](https://askubuntu.com/questions/493862/netcat-not-working-at-least-not-that-i-can-tell) but it does not seem to apply for io-tests.sh

any pointers?

steve-chavez: @Dansvidania  Oh. I think it's related to the shell. 

Are you executing the script like `./io-tests.sh`? If so, does it work if you try with `bash io-tests.sh`?

Dansvidania: @steve-chavez unfortunately not.

steve-chavez: @Dansvidania I see. I think for now we can leave the tests with nmap. Check my other comment about `sleep`.

steve-chavez: @Dansvidania Also, please  `rebase` against master. Just merged another PR that changed some of the files. Sorry about that.

steve-chavez:
@Dansvidania Looks good to me!

@Qu4tro What do you think?

steve-chavez:
@Dansvidania Thanks a lot for the work you put on this PR.

I'll merge it now. Soon we'll have a new release :rocket:.

Dansvidania:
@steve-chavez thank you for the feedback. Its a pleasure to be a small part of this great project :)

Qu4tro:
Sorry, been away for a while. This looks great!

Read this the other day that remind me of it: 
https://jvns.ca/blog/2015/11/21/why-you-should-understand-a-little-about-tcp

Haven't actually checked if it's an issue, just something I thought of bringing up :smile:

PR #1304, opened 2019-05-27
Change tests to use postgres-json-schema instead of hjsonschema.

### LINKED ISSUE #934: Upgrade hjsonschema version(or remove it) in cabal file

robx:
Fixes #934.

Embeds https://github.com/gavinwahl/postgres-json-schema.

Feels like gavinwahl/postgres-json-schema should be credited more properly, but I didn't find a good place to do so, and `test/fixtures/jwt.sql` does things the same.

(The tests pass with some other local changes and stackage LTS-13.23, haven't been able to test with cabal or older stackage, but I don't expect problems there.)

### DIFF
--- a/test/SpecHelper.hs
+++ b/test/SpecHelper.hs
@@ -5,14 +5,12 @@ import qualified Data.ByteString.Char8  as BS
 import qualified Data.ByteString.Lazy   as BL
 import qualified Data.Map.Strict        as M
 import qualified Data.Set               as S
-import qualified JSONSchema.Draft4      as D4
 import qualified System.IO.Error        as E
 
 import Control.Monad        (void)
-import Data.Aeson           (Value (..), decode)
+import Data.Aeson           (Value (..), decode, encode)
 import Data.CaseInsensitive (CI (..))
 import Data.List            (lookup)
-import Data.Maybe           (fromJust)
 import Network.Wai.Test     (SResponse (simpleBody, simpleHeaders, simpleStatus))
 import System.Environment   (getEnv)
 import System.Process       (readProcess)
@@ -45,17 +43,19 @@ validateOpenApiResponse headers = do
     let respHeaders = simpleHeaders r in
     respHeaders `shouldSatisfy`
       \hs -> ("Content-Type", "application/openapi+json; charset=utf-8") `elem` hs
-  liftIO $
-    let respBody = simpleBody r
-        schema :: D4.Schema
-        schema = D4.emptySchema { D4._schemaRef = Just "openapi.json" }
-        schemaContext :: D4.SchemaWithURI D4.Schema
-        schemaContext = D4.SchemaWithURI
-          { D4._swSchema = schema
-          , D4._swURI    = Just "test/fixtures/openapi.json"
-          }
-       in
-       D4.fetchFilesystemAndValidate schemaContext ((fromJust . decode) respBody) `shouldReturn` Right ()
+  let Just body = decode (simpleBody r)
+  Just schema <- liftIO $ decode <$> BL.readFile "test/fixtures/openapi.json"
+  let args :: M.Map Text Value
+      args = M.fromList
+        [ ( "schema", schema )
+        , ( "data", body ) ]
+      hdrs = acceptHdrs "application/json"
+  request methodPost "/rpc/validate_json_schema" hdrs (encode args)
+      `shouldRespondWith` "true"
+      { matchStatus = 200
+      , matchHeaders = []
+      }
+
 
 getEnvVarWithDefault :: Text -> Text -> IO Text
 getEnvVarWithDefault var def = toS <$>
@@ -130,6 +130,7 @@ setupDb dbConn = do
   loadFixture dbConn "roles"
   loadFixture dbConn "schema"
   loadFixture dbConn "jwt"
+  loadFixture dbConn "jsonschema"
   loadFixture dbConn "privileges"
   resetDb dbConn
 
--- a/test/fixtures/jsonschema.sql
+++ b/test/fixtures/jsonschema.sql
@@ -1,0 +1,280 @@
+-- from gavinwahl/postgres-json-schema commit 5a257e19a1569a77b82e9182b0b7d9fc8b6f6382
+
+/*
+Copyright (c) 2016, Gavin Wahl
+
+Permission to use, copy, modify, and distribute this software and its
+documentation for any purpose, without fee, and without a written agreement is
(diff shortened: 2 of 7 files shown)


robx:
The failure in in `build-test-9.4` was an hlint error.

The failures in `build-test-{10,11}`  seem real:

       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: true
         but got:  {"hint":null,"details":"Token \"swagger\" is invalid.","code":"22P02","message":"invalid input syntax for type json"}

robx:
My best guess at this point is that `postgres-json-schema` is incompatible with some change in postgresql 10.

There are some JSON-related changes in https://www.postgresql.org/docs/10/release-10.html, but nothing stands out...

robx:
The test failures with PostgreSQL >= 10 appear to be due to bug related to the parsing of embedded JSON arguments, due to a change in the behavior of `record_to_json` in PostgreSQL.

Qu4tro:
Hey there. I've tested this patch with both 9.4 and 11. Can you give this a go?

```
From 36a39c5dd98031bda1eabe97742cf786e75d174a Mon Sep 17 00:00:00 2001
From: Xavier Francisco
Date: Mon, 27 May 2019 20:53:44 +0100
Subject: [PATCH] Fix json encoding

---
 test/SpecHelper.hs | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/test/SpecHelper.hs b/test/SpecHelper.hs
index b576c8e..c3f476e 100644
--- a/test/SpecHelper.hs
+++ b/test/SpecHelper.hs
@@ -15,11 +15,11 @@ import Data.List            (lookup)
 import Network.Wai.Test     (SResponse (simpleBody, simpleHeaders, simpleStatus))
 import System.Environment   (getEnv)
 import System.Process       (readProcess)
+import Test.Hspec           (shouldSatisfy)
 import Text.Regex.TDFA      ((=~))
 
 
 import Network.HTTP.Types
-import Test.Hspec
 import Test.Hspec.Wai
 import Text.Heredoc
 
@@ -47,10 +47,10 @@ validateOpenApiResponse headers = do
   liftIO $
     (decode (simpleBody r) :: Maybe Value) `shouldSatisfy` isJust
   schema <- liftIO $ T.readFile "test/fixtures/openapi.json"
-  let args :: M.Map Text Text
+  let args :: M.Map Text Value
       args = M.fromList
-        [ ( "schema", schema )
-        , ( "data", decodeUtf8 . BL.toStrict . simpleBody $ r ) ]
+        [ ( "schema", fromMaybe "{}" . decode . toS $ schema )
+        , ( "data",   fromMaybe "{}" . decode . toS . decodeUtf8 . BL.toStrict . simpleBody $ r ) ]
       hdrs = acceptHdrs "application/json"
   request methodPost "/rpc/validate_json_schema" hdrs (encode args)
       `shouldRespondWith` "true"
-- 
2.21.0

```

test/SpecHelper.hs
```diff
@@ -46,16 +45,19 @@ validateOpenApiResponse headers = do
     respHeaders `shouldSatisfy`
```
Qu4tro: Not really related to the PR, but could these checks be replaced by a single `shouldRespondWith` ?

robx:
> Hey there. I've tested this patch with both 9.4 and 11. Can you give this a go?

Thanks! I'm a bit suspicious of it...

```
-  let args :: M.Map Text Text
+  let args :: M.Map Text Value
       args = M.fromList
-        [ ( "schema", schema )
-        , ( "data", decodeUtf8 . BL.toStrict . simpleBody $ r ) ]
+        [ ( "schema", fromMaybe "{}" . decode . toS $ schema )
+        , ( "data",   fromMaybe "{}" . decode . toS . decodeUtf8 . BL.toStrict . simpleBody $ r ) ]
```

- The `"{}"` is of type `Value` via a JSON `IsString` instance? Is that the JSON `"{}"` or `{}`?
- Regardless, I fear this might silently hide errors when the arguments are not valid JSON, since validating an empty JSON object or against an empty JSON schema might trivially succeed.

Is the idea to pass the JSON arguments as unquoted JSON? In that case, could I ask for your input on https://github.com/PostgREST/postgrest/pull/1306? Perhaps the bug there is not what I think it is.

test/SpecHelper.hs
robx: My understanding is that `shouldRespondWith` requires specifying the full response body, which is why it's avoided here.

Qu4tro:
You're right in both points (and it's unintentionally `"{}"`). Either way, I agree that having a default is not what we want here. I think we should check the decode call instead and fail the test if either decoding fails.

Yes, the idea is to pass a `Aeson.Value` instead of the quoted json.
Looking.

robx:
Let's see then, pushed a change to not quote the arguments.

robx:
I've rebased this onto master; the Postgres 9.6 vs Postgres 10 issue was resolved by embedding the JSON unquoted as suggested by @Qu4tro (though I'm still not sure what's really "right" here, but that can be discussed in #1306).

test/fixtures/jsonschema.sql
```diff
@@ -0,0 +1,260 @@
+-- from gavinwahl/postgres-json-schema commit 5a257e19a1569a77b82e9182b0b7d9fc8b6f6382
+
```
steve-chavez: Regarding credit, how about if we copy the contents of https://github.com/gavinwahl/postgres-json-schema/blob/master/LICENSE here.

steve-chavez:
@robx Great PR! I've been wanting to include `postgres-json-schema` for a while.

Been playing with it and doing tests like:
```sql
\set openapi $$`cat test/fixtures/openapi.json`$$::jsonb
\set openapi_pgrst $$`curl -s localhost:3000`$$::jsonb

select test.validate_json_schema(:openapi, :openapi_pgrst); -- true
select test.validate_json_schema(:openapi, '{"swagger": "2.0", "info": null, "paths": null}'); -- false
select test.validate_json_schema(:openapi, '{"swagger": "2.0", "info":{"version":"5.2.0","title":"PostgREST API"}, "paths": {}}'); -- true
```
And it works really well! I think we can safely replace hjsonschema.

See my review and after that I think we can merge.

test/fixtures/jsonschema.sql
robx: Sounds good. Done.

Issue #1292, opened 2019-05-14
switch postgrest configurator-ng to dhall configure for latest ghc build


clojurians-org:
the depends on configurator-ng(and it's further depend on critbit) is too old to adapt to new ghc environment.
it's only used for simple config parser  function.

i currently try to use nix to build static build for postgrest, there's a little problem to solve.
but dynamic build is relative easy on postgrest for latest GHC after adopting dhall instead of configurator-ng.

https://github.com/clojurians-org/nix-haskell-build/blob/master/postgrest/src/PostgREST/Config.hs

steve-chavez:
Thank you for the suggestion. Do you know if there's a way for dhall to accept our current configuration syntax(without curly brackets and commas)?

robx:
While I don't object at all to moving away from `configurator-ng` (it seems quite abandoned), as a less invasive short-term fix I tried updating `critbit` and `configurator-ng` for current GHC: https://github.com/lpsmith/configurator-ng/pull/9 https://github.com/bos/critbit/pull/93

It's unclear to me whether those will make it to hackage anytime soon though.

Qu4tro:
I fully support the move to `dhall`. Not only is well supported, it also brings both flexibility and ease with it.

@steve-chavez I think the conversion from currents `configurator-ng` config to `dhall` ones, can be automated. Would this settle worries about backwards incompatibility?
We would then provide said tool, so that people could use to effortlessly update their config to the new version.

This would also simplify configuration code-wise. The example from @clojurians-org already simplifies the `Config.hs` module so much. `dhall` also supports reading from ref files, so we should be able to remove `loadSecretFile` and `loadDbUriFile` from `Main.hs` as well.

ruslantalpa:
Unless dhall can support the exact config format @robx has the right approach, fix upstream. Maintainers should be contacted over email and asked respectfully to merge (when they like the code) the pr and if they are willing, until the change makes it into hackage, postgrest should use directly the branches with the fixes.

There is no need for complex configuration capabilities, having a simple configuration file is a goal for postgrest to have and it does not get any simpler the key=value and its a format everyone knows and understands as opposed to dhall

Qu4tro:
I agree that fixing upstream should be the approach to take to fix this issue. Contacting the maintainers and use the fixed branch in the meantime.

[...]

PR #1305, opened 2019-05-27
Vendor orphaned configuration dependencies critbit and configurator-ng

robx:
To get unstuck on these orphaned and stale dependencies, I suggest bundling them with PostgREST as an intermediate step. This will allow easily updating them to build with newer GHC/base versions, and kicking them back out once a better alternative is found.

### DIFF
--- a/deps/Data/Configurator.hs
+++ b/deps/Data/Configurator.hs
@@ -1,0 +1,527 @@
+{-# LANGUAGE CPP, BangPatterns, OverloadedStrings, RecordWildCards,
+    ScopedTypeVariables, TupleSections #-}
+
+-- |
+-- Module:      Data.Configurator
+-- Copyright:   (c) 2011 MailRank, Inc.
+--              (c) 2015-2016 Leon P Smith
+-- License:     BSD3
+-- Maintainer:  Leon P Smith <leon@melding-monads.com>
+-- Stability:   experimental
+-- Portability: portable
+--
+-- A simple (yet powerful) library for working with configuration
+-- files.
+--
+-- Note that while the "Data.Configurator.Parser" and
+-- "Data.Configurator.FromValue" should be quite stable at this point,
+-- this module is likely to be subjected to significant breaking changes in
+-- subsequent versions of configurator-ng.   So please do file an issue
+-- if you have any opinions or especially needs with regards to
+-- configuration (re)loading,  change notifications,  etc.
+
+module Data.Configurator
+    (
+    -- * Configuration file format
+    -- $format
+
+    -- ** Binding a name to a value
+    -- $binding
+
+    -- *** Value types
+    -- $types
+
+    -- *** String interpolation
+    -- $interp
+
+    -- ** Grouping directives
+    -- $group
+
+    -- ** Importing files
+    -- $import
+
+    -- * Types
+      Worth(..)
+    -- * Loading configuration data
+    , autoReload
+    , autoReloadGroups
+    , autoConfig
+{--
+    -- * Lookup functions
+
+    , lookup
+    , lookupDefault
+    , require
+--}
+    -- * Notification of configuration changes
+    -- $notify
+    , prefix
+    , exact
+    , subscribe
+    -- * Low-level loading functions
+    , load
+    , loadGroups
+    , reload
+    , addToConfig
+    , addGroupsToConfig
+    -- * Helper functions
+    , display
+    , readConfig
+    ) where
+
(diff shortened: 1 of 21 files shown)


robx:
Some open questions:
- cut the embedded libraries down to the minimum that's used by PostgREST?
- should I add fixes for new GHC to the code directly? (I'm not quite sure how to deal with unused import warnings for Semigroup/Monoid across GHC versions)
- I added the original LICENSE files (witih copyright info updated from the cabal files) within deps; should something be done at top-level?

steve-chavez:
@robx Totally get why we need more control over those dependencies, as they seem unmaintained.

But instead of bundling them in this repo(5k lines added, big impact in the codebase), couldn't we have them forked and then published to hackage(with different names, like pgrst-critbit or smth) so we can instead reference them?

ruslantalpa:
Or without publishing to hackage and just referencing the git

https://stackoverflow.com/questions/38377521/haskell-stack-install-package-dependency-from-github

ruslantalpa:
Could be a repo under postgrest org in github

robx:
Sure that would work for me too. I don't particularly care either way, just wanted to help make progress. And the advantage with this approach is that we see integration test results. :)

(Also it seemed like a less invasive change than putting up a fork of a deprecated package on hackage that's meant to be abandoned again soon.)

If it's really just the impact on the repository, I'm happy to cut it down to size. If there's some way I can help with the forking/uploading, I'm also happy to do that.

steve-chavez:
> (Also it seemed like a less invasive change than putting up a fork of a deprecated package on hackage that's meant to be abandoned again soon.)

@robx You're right about hackage. Just forking and referencing git as Ruslan suggested would be better.

robx:
Referencing a git revision in a local stack.yaml or whatever build file you use is all well and good, but doesn't do anything about the fact that postgrest isn't buildable from hackage with recent ghc versions.

Then you might as well deprecate it on hackage and note that you should get binary/source releases from the github repo / project homepage, no?

My personal motivation here is that I'd like to get postgrest built in nix/guix, and while it's possible to work around such issues there, it would be much more straightforward and clean to build directly from hackage. (As an aside, it would also give less of a stale impression if postgrest were buildable out of the box with recent ghc versions; I know it's actively maintained, but still.)

robx:
Tell you what though, I'll give reimplementing the config parser a shot, seems like a fun exercise.

ruslantalpa:
I a not opposed to having a postgrest specific config parser since we don't require a complicated logic or fancy features and while it seems a bit like reinventing the weel, it's a good tradeof to have postgrest easely bildable/installable.
having these features seems enough to me for any future uses:
- `key = value` `key=value` `key="value with spaces"` basic structure
- support comments with `#`
- `key=@file` read the value from file
- key=$(ENV_VAR)
- `key.sub1=a` `key.sub2=b` a map


this is a good overview of what's being used now https://github.com/PostgREST/postgrest/blob/master/docker/postgrest.conf

other oppinions?

Qu4tro:
I've already gave my opinion on #1292 and I still have the same concern, but I'll expand them a bit.
I understand the motivation for reducing dependencies. I think #1304 is great! Yes, dependencies either get stale or get upgraded and have backwards-compatibility issues. I definitely can't say for certain that it won't happen to any package we add. On the other hand, we're responsible for more code that can go wrong. 

I think any approach that will fix the build is a good one at the moment. Either the upload of the fixed version to hackage (has anyone emailed the mantainers?) or the in-house implementation of it. So no objection on my part.

Still I think a bit more long term (after have builds working, after adding deprecation warnings, after adding migration options), we should stay away from a "simple", ad-hoc configuration file. I definitely don't agree that it's great for future use.
To be clear, when I mention `dhall` I'm trying to advocate to depend on a standardized configuration language. It's just that is one of the few that fit the bill. It also has the advantage of being available on stackage and hackage. I understand that `dhall` is full of fancy features that are not needed for pgrst, but reading environmental variables and mostly reading from files are not common features either.
The advantages of the use of a standardized config language are numerous. Editor syntax files (a big red warning if you mess up), tools can be more easily built to manipulate configuration files, instead of trying to provide an almost perfect printer/parser for the ad-hoc version (which is kind of nice, if you have to administrate +20 servers - even if they're not all pgrst) and well.. less code.

Anyway, that's my opinion.

Qu4tro:
I assume no one yet has emailed the maintainers of configurator-ng for the lack of responses.
I'll send one later tonight.

robx:
Closing in favor of #1312

Issue #837, opened 2017-03-24
Inserting composite type fails


rokkie:
I've created the following schema:

```sql
CREATE TYPE public.material AS ENUM ('aluminium','steel');

CREATE TYPE public.cylinder AS
(
  capacity smallint,
  material public.material
);

CREATE TABLE public.dive(
    id integer NOT NULL,
    cylinder public.cylinder NOT NULL,
    CONSTRAINT pk_dive PRIMARY KEY (id)
);
```

When i try to insert a record with the following request
```http
POST /dive HTTP/1.1
Host: localhost:3000
Content-Type: application/json

{
	"id": 123,
	"cylinder": "(12,'steel')"
}
```

It fails with the following message:
```json
{
  "hint": null,
  "details": null,
  "code": "22P02",
  "message": "invalid input value for enum material: \"'steel'\""
}
```

ruslantalpa:
when you select a value of that type, how is it displayed?
i.e. when pg casts it to "text" how does it look?

try using that representation as a value

rokkie:
Ok, i feel really silly now.
I could have sworn i've tried it without quotes, which works just fine.
Thanks for pointing me in the right direction!

russelldavies:
For non-ENUM composite types there is the problem that the representation returned in a GET request is different from the representation that must be sent in a POST/PATCH/PUT request. 
Example:
```sql
CREATE TYPE mytype AS (a int, b int);
CREATE TABLE foo (id serial primary key, type mytype);
INSERT INTO foo (type) VALUES ((1,2,))
```
A GET request to `/foo` will return:
```json
[
  {
    "id": 1,
    "type": {
      "a": 1,
      "b": 2
    }
  }
]
```
However, sending this as the body of the POST request will fail:
```json
{
  "type": {
    "a": 3,
    "b": 4
  }
}
```
with the error message
```json
{
  "hint": null,
  "details": "Missing left parenthesis.",
  "code": "22P02",
  "message": "malformed record literal: \"{\"a\":3,\"b\":4}\""
}
```
and what Postgres logs:
```
ERROR:  malformed record literal: "{"a":3,"b":4}"
DETAIL:  Missing left parenthesis.
STATEMENT:  
	      WITH pg_source AS (INSERT INTO "public"."foo" ("type")SELECT "type" FROM json_populate_recordset(null:: "public"."foo" , $1) RETURNING  "public"."foo".*)
[...]

PR #1306, opened 2019-05-27
Fix parsing of embedded quoted JSON with PostgreSQL version 10 and later

robx:
* Add test cases to check whether RPCs accept quoted and raw embedded JSON (some of these are expected to fail while it's unclear what the desired behavior should be)
* Add a shim for PostgreSQL >= 10 to ensure postgREST keeps accepting quoted JSON fields

The underlying PostgreSQL change seems like a regression as a side-effect to changes to `record_to_json` to better deal with composite values, as mentioned in the release notes: https://www.postgresql.org/docs/10/release-10.html#id-1.11.6.13.5.

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -165,25 +165,11 @@ callProc qi pgArgs returnsScalar selectQuery countQuery countTotal isSingle para
       | paramsAsSingleObject = ("_args_record AS (SELECT NULL)", "$1::json")
       | null pgArgs = (ignoredBody, "")
       | otherwise = (
-          let explicit = pgVer >= pgVersion100
-              rawType a = if explicit
-                then case pgaType a of
-                       "json"  -> "text"
-                       "jsonb" -> "text"
-                       t       -> t
-                else pgaType a
-              conversion = unwords [
-                "SELECT",
-                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> "::" <> pgaType a) <$> pgArgs),
-                "FROM ("]
-          in
           unwords [
             normalizedBody <> ",",
             "_args_record AS (",
-            if explicit then conversion else "",
-            "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
-                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> rawType a) <$> pgArgs) <> ")",
-            if explicit then ") _x" else "",
+              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
+                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> pgaType a) <$> pgArgs) <> ")",
             ")"]
          , intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " := _args_record." <> pgFmtIdent (pgaName a)) <$> pgArgs))
 
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -165,11 +165,25 @@ callProc qi pgArgs returnsScalar selectQuery countQuery countTotal isSingle para
       | paramsAsSingleObject = ("_args_record AS (SELECT NULL)", "$1::json")
       | null pgArgs = (ignoredBody, "")
       | otherwise = (
+          let explicit = pgVer >= pgVersion100
+              rawType a = if explicit
+                then case pgaType a of
+                       "json"  -> "text"
+                       "jsonb" -> "text"
+                       t       -> t
+                else pgaType a
+              conversion = unwords [
+                "SELECT",
+                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> "::" <> pgaType a) <$> pgArgs),
+                "FROM ("]
+          in
           unwords [
             normalizedBody <> ",",
             "_args_record AS (",
-              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
-                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> pgaType a) <$> pgArgs) <> ")",
+            if explicit then conversion else "",
+            "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
+                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> rawType a) <$> pgArgs) <> ")",
+            if explicit then ") _x" else "",
             ")"]
          , intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " := _args_record." <> pgFmtIdent (pgaName a)) <$> pgArgs))
 
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -288,7 +288,7 @@ requestToQuery schema _ (DbMutate (Insert mainTbl iCols onConflct putConditions 
     "INSERT INTO ", fromQi qi, if S.null iCols then " " else "(" <> cols <> ")",
     unwords [
       "SELECT " <> cols <> " FROM",
-      "json_populate_recordset", "(null::", fromQi qi, ", " <> selectBody <> ") _",
+      "jsonb_populate_recordset", "(null::", fromQi qi, ", " <> selectBody <> ") _",
       -- Only used for PUT
       ("WHERE " <> intercalate " AND " (pgFmtLogicTree (QualifiedIdentifier "" "_") <$> putConditions)) `emptyOnFalse` null putConditions],
     maybe "" (\(oncDo, oncCols) -> (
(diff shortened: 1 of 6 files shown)


robx:
This feels like a regression as a side-effect to changes to `record_to_json` to better deal with composite values, as mentioned in the release notes: https://www.postgresql.org/docs/10/release-10.html#id-1.11.6.13.5

src/PostgREST/QueryBuilder.hs
```diff
@@ -165,11 +165,25 @@ callProc qi pgArgs returnsScalar selectQuery countQuery countTotal isSingle para
...
+          in
           unwords [
             normalizedBody <> ",",
             "_args_record AS (",
-              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
-                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> pgaType a) <$> pgArgs) <> ")",
+            if explicit then conversion else "",
+            "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
+                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> rawType a) <$> pgArgs) <> ")",
+            if explicit then ") _x" else "",
```
robx: I believe I have to give the subquery a name, is `_x` a good choice?

robx:
The fix has the side-effect of accepting both variants with pg >= 10 (if it's unquoted JSON, it will be converted to text and back now). It feels like it might be better to reject the unquoted variant.

On the other hand, it's likely it would be possible to add a similar shim to pg <=  9.6 to make it accept both variants.

robx:
Also it's possible that I see this all the wrong way around. Is it rather an accident that quoted JSON works with PostgreSQL <=9.6?

Qu4tro:
I think so.

This fails on both pg11 and pg9.4:
```
select '"{"a": 2}"'::jsonb                               
invalid input syntax for type json
LINE 1: select '"{"a": 2}"'::jsonb
               ^
DETAIL:  Token "a" is invalid.
CONTEXT:  JSON data, line 1: "{"a...
```

I don't think that's a mistake from postgres. A quoted json is just a string. Sounds good to me.

It's a worthy question to ask whatever needs fixing. 
I'm leaning on no, mainly because I think that was an odd behavior from pg<10 and that it won't be available on the next versions. Unquoted json can still be used, so I don't think we're losing anything.
On the other hand, PostgREST will have a different behavior, depending on pg version, which is a bit of a pain.

I'm actually more inclined to disallow the quoted json or keep the current behavior, but I'll let others chime in as well.

Finally, that's only related to the changes on `QueryBuilder`. 
The rest of the PR looks great.

steve-chavez:
Sorry. I'm confused about this. Not sure what the issue is.

@Qu4tro example fails for me in all pg versions from 9.4.21 to 11.

Let's see, `select '"smth"'::jsonb` is valid since that's a **json string**. Qu4tro's example basically is `'"{"a'` and of course the **a token** makes it an invalid json string, hence the pg error.

A `'"{\"key\": 3}"'` is also a valid **json string**, because quotes are escaped.

All of the PR new test cases work for me from pg 9.4.21 to 11 without the change to `QueryBuilder.hs`. So not sure why the change is needed.

Extra tests are good though.

Also, perhaps this is related to https://github.com/PostgREST/postgrest/issues/837?

steve-chavez:
I get it now. Check this(on pg10):
```sql
select *  from  json_to_recordset('[{"double": 3.1, "varchar": "hello", "boolean": true, "date": "20190101", "money": 0, "enum": "foo", "integer": 43, "jsonb": "{\"this is embedded\": 5}" }]') as x("double" double precision, "varchar" varchar, "boolean" boolean, "date" date, "money" money, "enum" enum_menagerie_type, "integer" integer, "jsonb" jsonb);

ERROR:  22P02: invalid input syntax for type json
DETAIL:  Token "this" is invalid.
CONTEXT:  JSON data, line 1: "{"this...
```
`json_to_recordset` fails. But `jsonb_to_recordset` succeeds:
```sql
select *  from  jsonb_to_recordset('[{"double": 3.1, "varchar": "hello", "boolean": true, "date": "20190101", "money": 0, "enum": "foo", "integer": 43, "jsonb": "{\"this is embedded\": 5}" }]') as x("double" double precision, "varchar" varchar, "boolean" boolean, "date" date, "money" money, "enum" enum_menagerie_type, "integer" integer, "jsonb" jsonb);
 double | varchar | boolean |    date    | money | enum | integer |            jsonb
--------+---------+---------+------------+-------+------+---------+-----------------------------
    3.1 | hello   | t       | 2019-01-01 | $0.00 | foo  |      43 | "{\"this is embedded\": 5}"
```
What's weird is that just casting to `json` works:
```sql
select '{ "double": 3.1, "varchar": "hello", "boolean": true, "date": "20190101", "money": 0, "enum": "foo", "integer": 43, "jsonb": "{\"this is embedded\": 5}" }'::json;
```
So yeah it seems is a bug in the impl of `json_to_recordset`.

src/PostgREST/QueryBuilder.hs
```diff
@@ -165,11 +165,25 @@ callProc qi pgArgs returnsScalar selectQuery countQuery countTotal isSingle para
...
+                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> "::" <> pgaType a) <$> pgArgs),
+                "FROM ("]
+          in
           unwords [
             normalizedBody <> ",",
             "_args_record AS (",
-              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
-                intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> " " <> pgaType a) <$> pgArgs) <> ")",
+            if explicit then conversion else "",
+            "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <>
```
steve-chavez: Try making the change to `jsonb_to_recordset`. Maybe tests pass without the casting to `text` logic above.

Not sure right now if there will be some downsides to using `jsonb_to_recordset`.

robx:
I'm also thoroughly confused, sorry for bringing that across here. :)

The core issue is:

PostgREST with Postgres 9.6 accepts `{ "embedded": "{ \"field\" : 1 }" }` as `(embedded jsonb)`
while PostgREST with Postgres 10 doesn't.

I originally ran into this with the hjsonschema tests because I was embedding JSON quoted as above, and got test failures on Postgres 10 but not 9.6; at the time it seemed clear to me that embedding the JSON as a string was the right thing to do.

By now I'm unclear on
- what's the intended behavior, if any: should PostgREST accept both a quoted json field and an unquoted json field or both?
- consequently, if the change here is good: It was developed on my original understanding to allow PostgREST/Postgres 10 to accept quoted json fields, but I'm not that clear on side-effects to other scenarios.
- should this be considered a bug (in Postgres or PostgREST)? My intuition is, yes, behavior shouldn't change with Postgres upgrade, but if embedding quoted json is wrong, perhaps the upgrade just fixes an unintended loop-hole.

robx:
Note that the tests aren't particularly great: They just test that postgrest accepts the parameters, they don't care whether the embedded quoted json makes it through as a json string or a json object.

In particular, I see

```
# Postgres 9.6
> select * from jsonb_to_record('{"out": "{\"key\": 1}"}') as (out jsonb);
    out     
------------
 {"key": 1}

# Postgres 10
      out       
----------------
 "{\"key\": 1}"
```

That difference would not be caught by these test cases.

robx:
I've reported this as a bug to postgreql, let's see what they say: https://www.postgresql.org/message-id/D6921B37-BD8E-4664-8D5F-DB3525765DCD%40vllmrt.net

steve-chavez:
@robx Sorry about the wrong hint on `jsonb_to_recordset`. Saw it broke a lot of tests.

I do think this is a bug in pg and we shouldn't try to patch it in PostgREST. Instead we should just document this shortcoming and keep the extra tests as proof of this behavior.

So, how about if you use `actualPgversion` to make sure the tests are ignored/applied depending on the pg version. This way we'd be green on CI. 
https://github.com/PostgREST/postgrest/blob/e34669b13789e64d44ca865bd5ae7ea29aef2c3a/test/Feature/RpcSpec.hs#L19

robx:
Closing in favor of https://github.com/PostgREST/postgrest/pull/1314

robx:
> @robx Sorry about the wrong hint on jsonb_to_recordset. Saw it broke a lot of tests.

No worries. It did require changing things to `jsonb` elsewhere, and curiously `jsonb` is serialized with slightly different spacing, which caused most of the test failures. The only real failure in the end was with Postgres 9.4, where some `jsonb_*` function was not yet availabe. I'm a bit curious whether this change to `jsonb` across the board has any performance implications.

steve-chavez:
@robx Seems postgres 11.4 patched this error https://www.postgresql.org/docs/release/11.4/.

Just got a [failure on CircleCI](https://circleci.com/gh/steve-chavez/postgrest/1604?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) since we're using the docker image circleci/postgres:11(meant 11.3 at the time of the PR, now 11.4)

robx:
@steve-chavez Thanks for the heads-up, possible fix in #1331.

Qu4tro:
It was fixed in 10.9 as well.

```
  test/Feature/RpcSpec.hs:279:11: 
  1) Feature.RpcSpec, remote procedure call, proc argument types, fails to parse quoted JSON arguments (Postgres >= 10, < 11.4)
       status mismatch:
         expected: 400
         but got:  200
       body mismatch:
         expected: {"hint":null,"details":"Token \"key\" is invalid.","code":"22P02","message":"invalid input syntax for type json"}
         but got:  "string"
```

PR #1312, opened 2019-05-29
Replace configurator-ng by configurator-pg (via stack.yaml/git)

robx:
I put together a variant of configurator-ng at https://github.com/robx/configurator-pg that should make a fine replacement for PostgREST's purposes. I believe it supports all features of configurator-ng's that PostgREST uses.

Test coverage is not great, but it does pass the relevant tests of the original configurator-ng, as well as a few extra ones. I also didn't manage to break it by hand. It seems to do fine across ghc versions: https://travis-ci.org/robx/configurator-pg

~~With the exception of proper dependency bounds,~~ the package should be ready to be published. I'd be happy for it to live under the PostgREST github org, but can also put it on hackage from where it's at right now. What do you prefer?

Errors come out of the API a little less structured; on the other hand PostgREST's config errors are nicer now.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -59,7 +59,7 @@ library
                      , bytestring
                      , case-insensitive
                      , cassava
-                     , configurator-ng == 0.0.0.1
+                     , configurator-pg >= 0.1 && < 0.2
                      , containers
                      , contravariant
                      , contravariant-extras
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -33,17 +33,17 @@ executable postgrest
   default-language:    Haskell2010
   build-depends:       auto-update
                      , base >= 4.8 && < 4.10
+                     , base64-bytestring
+                     , bytestring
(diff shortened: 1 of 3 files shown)


robx:
Hmm circleci now failing with

```
   Configuring configurator-pg-0.1.0.0...
    setup: This package description follows version 2.4 of the Cabal
    specification. This tool only supports up to version 1.24.2.0.
```

Any idea what I need to change in https://github.com/robx/configurator-pg/blob/master/configurator-pg.cabal? (I tried putting a lower cabal-version in before, but then cabal complained...)

steve-chavez:
@robx Great work on `configurator-pg`. I've tested manually the env var interpolation and since it passes our `io-tests.sh` I think all should be fine. 

Regarding publishing, Hackage would be fine. Since our needs for the config file are really simple, we would rarely need to change `configurator-pg`.

robx:
It's up on hackage now: http://hackage.haskell.org/package/configurator-pg

Glad you like it!

PR #1314, opened 2019-06-01
Test current behavior with respect to JSON RPC arguments

robx:
In particular, check buggy behavior around embedded quoted JSON
and Postgres versions, compare

https://www.postgresql.org/message-id/D6921B37-BD8E-4664-8D5F-DB3525765DCD%40vllmrt.net
   
This adds version-bounded tests for the handling of quoted JSON,
and a pending test that documents the assumption that Postgres >=10
intends to parse quoted JSON as a string (similar to how jsonb works
now).


robx:
Regarding documentation, the note at

> For versions prior to PostgreSQL 10, to pass a PostgreSQL native array you need to quote it as a string:

would seem to be a good spot to add a note on this.

PR #1317, opened 2019-06-03
Add option for overriding root spec

steve-chavez:
Continuing the work on https://github.com/PostgREST/postgrest/pull/1302.

I've kept it simpler and leave Content-Type override alone. Only added the option to request `Accept: application/openapi+json` for the root proc spec.


steve-chavez:
Not sure if it's correct to let the user specify a schema for `root-spec`. A proc like `private.root_spec` could be added and this would encourage to open access to the private schema? Maybe I should just let this spec be determined by the `search_path`. 

I think it's no problem if the spec is available at `/rpc/spec` besides being available at root `/`.

PR #1319, opened 2019-06-05
Update cabal bounds

robx:
I went over the PostgREST dependencies now that the problematic hjsonschema and configurator-ng dependencies are gone, with the results in this PR.

The important changes are:
- Add `base >= 4.9` to force GHC >= 8.0 for the `DuplicateRecordFields` extension.
- Add `swagger2 < 2.4` since it has incompatible changes.
- Relax the base upper bound to `< 4.13`, allowing GHC 8.6.
- Allow Ranged-sets 0.4.0 and protolude 0.2.3.

The current restrictive upper bounds seem to be:
- swagger2 < 2.4 as mentioned above.
- hasql < 1.4 due to incompatible changes in 1.4.
- jose < 0.8 due to an error message change in jose-0.8 that causes a unit test failure.

~~I've included stack configurations that I've built edge cases successfully with.~~ I compiled with tests but didn't actually run the tests in all cases.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -32,16 +32,16 @@ executable postgrest
     "-with-rtsopts=-N -I2"
   default-language:    Haskell2010
   build-depends:       auto-update
-                     , base >= 4.9 && < 4.10
+                     , base >= 4.9 && < 4.13
                      , base64-bytestring
                      , bytestring
                      , directory
                      , hasql >= 1.3 && < 1.4
                      , hasql-pool >= 0.5 && < 0.6
                      , hasql-transaction >= 0.7 && < 0.8
-                     , network == 2.6.3.2
+                     , network < 2.9
                      , postgrest
-                     , protolude == 0.2.2
+                     , protolude >= 0.2.2 && < 0.3
                      , retry
                      , text
                      , time
@@ -56,7 +56,7 @@ library
   default-extensions:  OverloadedStrings, QuasiQuotes, NoImplicitPrelude
   build-depends:       aeson
                      , ansi-wl-pprint
-                     , base >= 4.9 && < 4.10
+                     , base >= 4.9 && < 4.13
                      , base64-bytestring
                      , bytestring
                      , case-insensitive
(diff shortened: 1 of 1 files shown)


robx:
I should note that besides the stack builds, I tested cabal-based builds in two settings:

- GHC 8.6.3 and new-style cabal.
- GHC 8.2.2 and old-style cabal.

steve-chavez:
@robx Been able to use newer GHCs is great!

Having a more restricted .cabal is good for me. But not sure what to do with the extra `stack-<number>.yaml` files, shouldn't we decide what new stack LTS version to pick and override our `stack.yaml`?

Which LTS version would make postgrest more easily buildable with nix?

robx:
I mostly put them all in because I assembled them and didn't want to lose them before it was clear what should happen. And to provide some proof that these cabal bounds actually work. :)

I'm happy to update the primary `stack.yaml`, it doesn't really matter to me.

Building on macos with nix, I was typically using a recent GHC 8.6 LTS version, but my file looks a bit different:

```
resolver: nightly-2019-06-06 # configurator-pg included since last night
extra-deps:
- hasql-1.3.0.6
- hasql-transaction-0.7.1
- jose-0.7.0.0
ghc-options:
  postgrest: -O2 -Werror -Wall -fwarn-identities -fno-warn-redundant-constraints -optP-Wno-nonportable-include-path
nix:
  enable: true
  packages: [pcre, pkgconfig, zlib]
```

`-optP-Wno-nonportable-include-path` is in there because I get warnings otherwise due to the weird case-insensitivity in my filesystem. And I use nix to provide the non-Haskell dependencies, only.

But really, it depends on what `stack.yaml` is there for at all. I didn't want to touch it without being sure of that. As far as I understand, it's used for the circle-ci tests and for your local development?

I'll go ahead and push an updated version just to see what the tests do.

robx:
I've rebased this on to of the `network` change. Also I've put in an updated `stack.yaml` with some speculative nix changes -- would you test those on NixOS?

robx:
It builds successfully now on travis against various GHC versions: https://travis-ci.org/robx/postgrest/builds/542387944. I'll include the travis config, might be nice to ensure that these builds don't break.

steve-chavez:
@robx Seems like `stack build --profile` is broken due to a GHC issue https://github.com/commercialhaskell/stack/issues/4275. And it hasn't been fixed yet https://gitlab.haskell.org/ghc/ghc/issues/12932.

Maybe we could try GHC 8.2.2(LTS 11.21) instead of the latest one and see if it works.

Regarding `travis.yml`, it's not stale, we use that to generate OSX binaries on new releases. I don't have other way right now to generate osx binaries so it'd be better to leave travis like it was.

robx:
@steve-chavez It seems the 8.4 profiling build works now. The error I saw in the circle-ci logs when building with 8.6 also looked different from the ticket you linked -- I interpreted them to mean an out-of-memory condition.

And thanks for the heads-up about travis. I'll go head and rebase this to squash the stack mess and remove the travis change.

robx:
I've split out some of the independent changes into other PRs:
- #1320 for the hlint warnings
- #1321 for the independent circle-ci changes

Those two and this should be mergeable individually. I'll put the actual stackage update into a separate PR that pulls them all together.

robx:
Here's a good travis run with cabal builds of various GHC versions: https://travis-ci.org/robx/postgrest/builds/543067761

PR #1321, opened 2019-06-08
Tweaks to circle-ci profiling build

robx:
- update stack version (required to build with GHC 8.6 stackage releases)
- remove some unused packages (packdeps, cabal-install) (these aren't used and cause trouble with newer stackage releases)
- avoid rebuilding in `memory-tests.sh` (no idea why this was there)

Split out from #1319 which was getting unwieldy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,8 +74,8 @@ jobs:
       - run:
           name: install stack & dependencies
           command: |
-            curl -L https://github.com/commercialhaskell/stack/releases/download/v1.1.2/stack-1.1.2-linux-x86_64.tar.gz | tar zx -C /tmp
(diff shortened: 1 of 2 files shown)


steve-chavez:
> avoid rebuilding in memory-tests.sh (no idea why this was there)

That's strange, when I try to do `stack exec -- postgrest test/memory-tests/config +RTS -p -h` locally, the following error is shown:
```
postgrest: the flag -p requires the program to be built with -prof
postgrest:
postgrest: Usage: <prog> <args> [+RTS <rtsopts> | -RTS <args>] ... --RTS <args>
```
That's why I included the `stack build --profile` in the bash script.

steve-chavez:
D'oh. Nevermind, I forgot that `stack build --profile` is already on the CI yaml.

PR #1322, opened 2019-06-08
Update stackage GHC to 8.4.4

robx:
Based on #1319, #1320, #1321.

This updates stack.yaml to build with ~~GHC 8.6.5~~ GHC 8.4.4, affecting both stack users and circle-ci integration tests.

The `stack.yaml` nix build is simplified in a way that works for me on macos with nix, but needs to be verified to work on NixOS.

~~Since the profiling build on circle-ci fails due to an apparent out-of-memory condition with GHC 8.6.5, this adds a separate `stack-prof.yaml` file that's used for that test.~~


steve-chavez:
Hmm. Doing memory tests in a different GHC could be misleading, how about if we use GHC 8.4.4 for now.

Once https://github.com/PostgREST/postgrest/pull/1286 is merged maybe we'll not get the out of memory error and we could switch to GHC 8.6.5

robx:
That sounds good, I changed it to use GHC 8.4.4 everywhere.

stack.yaml
```diff
@@ -1,17 +1,12 @@
...
-  - wai-middleware-static-0.8.2
-  - configurator-pg-0.1.0.1
+- Ranged-sets-0.4.0
+- configurator-pg-0.1.0.3
+- http-types-0.12.3
 ghc-options:
   postgrest: -O2 -Werror -Wall -fwarn-identities -fno-warn-redundant-constraints
 nix:
-  shell-file: shell.nix
+  packages: [pcre, pkgconfig, zlib]
```
steve-chavez: Packages needs postgresql as well, like `packages: [postgresql, pcre, pkgconfig, zlib]`.

steve-chavez: Otherwise it doesn't build on NixOS.

PR #1324, opened 2019-06-11
Support jose 0.8.0

robx:
Not sure this is necessarily a good change, but I'd like to put it up for discussion. While raising the cabal bounds, I ran into a test failure with jose 0.8.0 because the `Show` instance of a certain error message changes. (Compare https://github.com/frasertweedale/hs-jose/issues/72.)

This generalizes the corresponding test to pass with both versions. Thoughts:
- I'd assume that PostgREST doesn't want to make any guarantees about precise wording of error messages, hence testing for precise error messages is rather because that's easier to test than because we want to check the precise error message
- the generalised test code is annoyingly complicated
- maybe it would be better to make some better test helper that understands how postgrest errors look `shouldRespondWithError request (checkErrorValue :: Text -> Bool)`?
- it's also perfectly fine to just decide to keep the upper bound and just switch over to `>= 0.8 && < 0.9` at some point when needed

I ran into this because current stackage includes jose 0.8.0.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -63,7 +63,7 @@ library
                     , http-types                >= 0.12.2 && < 0.13
                     , insert-ordered-containers >= 0.1 && < 0.3
                     , interpolatedstring-perl6  >= 1 && < 1.1
-                    , jose                      >= 0.7 && < 0.8
+                    , jose                      >= 0.8 && < 0.9
                     , lens                      >= 4.14 && < 4.18
                     , lens-aeson                >= 1.0.1 && < 1.1
                     , network-uri               >= 2.6.1 && < 2.7
--- a/stack.yaml
+++ b/stack.yaml
@@ -12,6 +12,7 @@ extra-deps:
 - text-builder-0.6.5.1
 - deferred-folds-0.9.10.1
 - primitive-0.6.4.0
+- jose-0.8.0.0
 ghc-options:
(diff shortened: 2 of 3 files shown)


steve-chavez:
Tests should be really simple to read, sometimes we even instruct users to read them to understand PostgREST functionality.

So I think it'd be preferable to upgrade to the newer jose and check the `JWSError (CompactDecodeError CompactDecodeError: Expected 3 parts; got 2)` error than to complicate the test code.

Also, in a way we do want to check the precise error messages to see if they're end-user understandable.

allen370:
從我的 iPhone 傳送

robx:
Done. Note that the message will probably change again with the next `jose` release: https://github.com/frasertweedale/hs-jose/commit/787b2e75a68ab1d81f4836fbd1a0ea20b372e895

robx:
Oh nice, postgrest would build with stackage nightly-2019-06-28 with this change.

test/SpecHelper.hs
```diff
@@ -32,7 +32,16 @@ matchContentTypeJson = "Content-Type" <:> "application/json; charset=utf-8"
 matchContentTypeSingular :: MatchHeader
 matchContentTypeSingular = "Content-Type" <:> "application/vnd.pgrst.object+json; charset=utf-8"
 
-validateOpenApiResponse :: [Header] -> WaiSession ()
+shouldRespondWith' :: WaiSession SResponse
```
steve-chavez: This function is unused right?

robx: Yes thanks for catching it.

PR #1325, opened 2019-06-11
Upgrade hasql to 1.4

robx:
Another PR that's more meant to document/not lose work, not necessarily to be merged. Perhaps at some point it will be desirable to use the latest hasql version?

hasql version 1.4 changes how nullable/non-nullable values are de- and encoded. Perhaps the new API has advantages, but for now it seems to just add verbosity. For now, this PR simply emulates the old API (redundantly in both affected modules).

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -56,9 +56,9 @@ library
                     , cookie                    >= 0.4.2 && < 0.5
                     , either                    >= 4.4.1 && < 5.1
                     , gitrev                    >= 1.2 && < 1.4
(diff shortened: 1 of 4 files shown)


steve-chavez:
Thank you @robx. I think we definitely want to stay up to date with hasql.

Right now I'd like to use http://hackage.haskell.org/package/hasql-dynamic-statements to see if it helps me with calling PostgreSQL 11 [stored procedures](https://www.postgresql.org/docs/11/sql-createprocedure.html).

Please rebase and I'll merge.

PR #1326, opened 2019-06-11
Upgrade swagger to 2.4

robx:
Third of the dependency update PRs. This is another one that might better wait. On top of the stackage update.

Two changes:
- tags are now an insert-ordered hashset
- `type_ .~` -> `type_ ?~` with some magic type inference

This is an incompatible change, hence the lower bound bump. And it pulls in a bunch of dependency updates via the required update of insert-ordered-containers.


robx:
Rebased onto current master. This one should probably wait until swagger2 2.4 makes it to stackage.

robx:
@steve-chavez This could be merged now. I added `-fno-spec-constr` as a work-around for the out-of-memory issues which seems to help (though I didn't re-test without). That flag was recommended here: https://gitlab.haskell.org/ghc/ghc/issues/16017#note_219304

robx:
Oh good, I missed that PR. In that case maybe the `-fno-spec-constr` should be dropped again?

robx:
> What would be the disadvantage of having `-fno-spec-constr`?

It's just that it adds an extra option to make GHC act in a non-default way. I assume that in principle `spec-constr` is something useful, deactivating it might mean some optimizations are missed. But the primary downside I see is a (very small) bit of extra complexity for postgrest as a whole.

> I just thought it'd be nice to not fight out-of-memory issues again.

I agree with that part. :)

I'll make a small PR to document those options in `stack.yaml`, that would mostly address my concerns.

PR #1331, opened 2019-06-24
Update json rpc argument tests for bugfix in PostgreSQL 11.4

robx:
See https://www.postgresql.org/docs/release/11.4/.

PR #1332, opened 2019-06-25
Work around macos case sensitivity issue

robx:
See also https://github.com/haskell/cabal/issues/4739.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -13,6 +13,6 @@ extra-deps:
 - deferred-folds-0.9.10.1
 - primitive-0.6.4.0
 ghc-options:
-  postgrest: -O2 -Werror -Wall -fwarn-identities -fno-warn-redundant-constraints
+  postgrest: -O2 -Werror -Wall -fwarn-identities -fno-warn-redundant-constraints -optP-Wno-nonportable-include-path
 nix:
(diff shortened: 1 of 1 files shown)


robx:
@steve-chavez The offending macos build is only triggered on tags I suppose?

robx:
This fixes the macos travis release build error: https://travis-ci.org/robx/postgrest/builds/550137527 (still fails because my postgrest travis isn't set up for release upload)

robx:
The circleci failure in build-test-11 is the one fixed by #1331. The circleci failure in build-prof-test looks like the one we saw before when building with ghc 8.6.5, an apparent out of memory issue while compiling postgrest. It seems unlikely but not impossible that this was directly caused by this change. Perhaps it's because everything needs to be recompiled for the change in GHC options?

steve-chavez:
@robx Actually travis(osx build) was triggered on every push to master. It builds dependencies for a limited time(`gtimeout 1800`) to not surpass the travis max build time and err. It's an ugly workaround that was working until now to build releases for osx.
 
> Perhaps it's because everything needs to be recompiled for the change in GHC options

Yes. That happens when compiling with profiling enabled. Not sure why.

PR #1333, opened 2019-06-25
Fix release process

steve-chavez:
Trying to fix the release process and trying to build the binaries manually for each platform. PR not necessarily to be merged, I'm mostly documenting the workarounds I'm trying.

The [ubuntu failure](https://circleci.com/gh/PostgREST/postgrest/3311?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) is a fresh one reported in https://github.com/commercialhaskell/stack/issues/4888. I think It could be solved by pinning the stack version instead of using the latest one in the container.

Right now I'm trying manual builds on an ec2 instance but I keep getting `Unable to commit 1048576 bytes of memory` errors. Seems there's [no](https://www.reddit.com/r/haskell/comments/83ks19/unsure_how_to_deal_with_this_error/) [fix](https://gitlab.haskell.org/ghc/ghc/issues/15500) for this error, so I'll have to upgrade the ec2 to another that has more ram.

Appveyor was timing out(60 mins build time max) so I'm using something like we do for travis with the `timeout` command to not surpass that limit. However the `45 mins` cap doesn't seem to be enough for building `text-printer` https://ci.appveyor.com/project/steve-chavez/postgrest. I'll try seeing how to solve this on a Windows machine soon.

Centos images had some outdate urls so I've fixed them, however now I'm getting another error `/home/centos/.stack/indices/Hackage/01-index.cache: /home/centos/.stack/indices/Hackage/01-index.cache: openBinaryFile: does not exist`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,7 @@ jobs:
       - image: circleci/buildpack-deps:trusty
         environment:
           - PGHOST=localhost
-      - image: circleci/postgres:11
+      - image: circleci/postgres:11.3
         environment:
           - POSTGRES_USER=circleci
           - POSTGRES_DB=circleci
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,10 @@
+## Travis is used for building an OSX binary on new tag releases,
+## it also compiles the project on OSX for each commit to master.
+##
+## Building the whole project can take longer than 50 minutes. Since Travis has a global timeout of 50 minutes
+## we compile for 30 minutes tops(see `gtimeout 1800` below) and quit compiling with no error.
+## Since we CACHE the compile results we can continue compiling from where we left off
+## on the next commit.
 language: generic
 
 sudo: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,4 @@
+## AppVeyor is used for building a Windows binary on new tag releases
 platform: x64
 
 cache:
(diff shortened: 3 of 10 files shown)


steve-chavez:
stack doesn't support ghc 8.4.4 for centos :disappointed: , see https://github.com/commercialhaskell/stack/issues/4161.

steve-chavez:
A showstopper for GHC 8.4 on Windows https://github.com/commercialhaskell/stackage/issues/3385. Seems the error could go if we upgrade to GHC 8.6.

steve-chavez:
Tried compiling with GHC 8.6 on windows but now I get the same error as in https://ci.appveyor.com/project/steve-chavez/postgrest/builds/25506383.

Reported that here https://github.com/mvv/text-printer/issues/1

steve-chavez:
While trying to manually build ubuntu 32 bit it fails with the same error as: https://circleci.com/gh/PostgREST/postgrest/3313?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link. Seems there's an issue with text-printer.

steve-chavez:
I'm unable to build haddock documentation with the new version, so I've just uploaded the source to https://hackage.haskell.org/package/postgrest

All it's missing is ubuntu 32 bit and windows, which don't compile in the current version. centos 6 users can use the static binary.

I'll open an issue for windows, for now I'll merge some of the PR fixes for the release process.

Issue #1249, opened 2019-02-28
Documentation inaccurate for ov operator


kairan77:
### Environment

* PostgreSQL version: (if using docker, specify the image)
11.X
* PostgREST version: (if using docker, specify the image)
5.2
* Operating system:
WIN10
### Description of issue

The online documentation http://postgrest.org/en/v5.2/api.html states that the syntax for using the operator ov is:
     /api?a=ov.['x','y'] ,
however upon testing the working syntax is:
     /api?a=ov.{'x','y'}
(Expected behavior vs actual behavior)
correct the documentation? I figure curly bracket is inline with cs/cd hence more intuitive.

(Steps to reproduce: Include a minimal SQL definition plus how you make the request to PostgREST and the response body)

![ovdocumentation](https://user-images.githubusercontent.com/48101363/53587717-0d0c8b00-3bc6-11e9-995a-15314e4ab2b4.png)

steve-chavez:
What's the type of `permissions`? Could you share a sample schema?

We have a test for `ov.[0, 4]` working: https://github.com/PostgREST/postgrest/blob/c9b2830e52184ad1abefa83619bb1b328213a14e/test/Feature/AndOrParamsSpec.hs#L103-L104

https://github.com/PostgREST/postgrest/blob/c9b2830e52184ad1abefa83619bb1b328213a14e/test/fixtures/schema.sql#L1225-L1228

kairan77:
create table data.permission (
	uuid		int GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
	issue_date	timestamp default now(),
	title		text default ''
);

kairan77:
create table data.department (
	uuid		int GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
	version		integer default 0,
	issue_date	timestamp default now(),
	last_modified timestamp default now(),
	title		text default '',
	dept_type	data.department_type default '',
	company		integer REFERENCES data.company(uuid) on delete cascade not null,
	members		integer[] default '{}',
	permissions	integer[] default '{}'
);

kairan77:
is ov designed only for range and not meant to be used for int[]?

steve-chavez:
> is ov designed only for range and not meant to be used for int[]?

It's fine to use `ov` for arrays, though seems the bracket notation `[]` only works for ranges, I see now why you got a syntax error.

[...]

PR #1337, opened 2019-07-04
Add tests for overlap operator in array types

herulume:
After reading #1249 I've noticed that there are no tests for the `ov` operator being used in array types. This PR addresses that. 

I was hesitant to put the tests in the `AndOrParamsSepc` file and in the `operators inside and/or` context, but since the `range` tests were also there, I've followed that convention.

Besides adding the `ov` tests I've also added tests for the rest of the supported array types operations.

The proper documentation was added in another [PR](https://github.com/PostgREST/postgrest-docs/pull/225).


steve-chavez:
@herulume Nice! More tests are always good.

Regarding organization, the tests just evolved to be that way and can definitely be reorganized. If you'd like to add a dedicated file for a spec(like `PgOperatorsSpec`), please feel free to do so.

PR #1338, opened 2019-07-04
Change CircleCI config to test in all supported major versions

Qu4tro:
Hey.

Currently `postgres:10.9` which is also `postgres:10` is failing. This PR is meant to make it easier to check our test-suit in postgres latest supported major versions. That is if I did the CircleCI magic right. (The goal is pg10 red and the others green).

@steve-chavez I've target it to master, but I'm thinking this could be a separate branch, which an empty-commit could be pushed to trigger circle. This also creates a pathway to automating this process (testing pgrst on new pg releases).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,9 +34,7 @@ jobs:
             stack build --fast --test --no-run-tests
       - run:
           name: run tests
-          command: |
-            POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://circleci@localhost" postgrest_test) stack test
-            test/io-tests.sh
+          command: POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://circleci@localhost" postgrest_test) stack test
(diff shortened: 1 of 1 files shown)


steve-chavez:
Now that https://github.com/PostgREST/postgrest/pull/1340 was merged, there shouldn't be an error on pg 10 I think.

Not pinning the minor version could be worth it, also testing `latest`(though this means 11.4? maybe we should pin 12 instead). That way we can be ahead if a breaking change happens.

I'd prefer this to be on master though, it's easy to forget doing a check and that happened in this new release with windows(only checked it on new tags).

So I think it's fine removing the minor versions but we should leave `9.4` pinned(we'll have to drop support for it eventually).

steve-chavez:
Postgres 9.4 will reach its End Of Life [this December](https://endoflife.software/applications/databases/postgresql), maybe that's a good reason for us to drop support as well.

PR #1339, opened 2019-07-05
Add websearch_to_tsquery support

### LINKED ISSUE #1243: websearch_to_tsquery support
Starting on PostgreSQL 11 there is a new function `websearch_to_tsquery` (a combination of `{plain,phrase}to_tsquery`) See: [textsearch-controls](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES) and [textsearch](https://www.postgresql.org/docs/current/functions-textsearch.html#TEXTSEARCH-FUNCTIONS-TABLE)... at the end of the section and the 7th row, respectively.

This is yet not supported by PostgREST. Reference: [Types.hs](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/Types.hs#L237)

On the side: as temporary workaround, is there a way to "expose" those not-yet-implemented functions on the API without creating custom views?

Thank you!

herulume:
This PR aims to add `websearch_to_tsquery` support and close #1243.

Since `websearch_to_tsquery` is only supported in `pg` 11.x, I was not sure how to struct the tests.
I've decided to pass the `PgVersion` where needed but, as @steve-chavez said on #1337, I think a `PgOperatorsSpec` test file would be ideal and that would contain most of the conditional tests. Didn't take that route here to keep this PR focused on the new feature.

One thing I've noticed is that `postgrest` gives a `404` error when a `pg` function is not implemented/found. Not sure if that's intentional or not, but I didn't find it intuitive since `pg` functions are more like operators than resources per se.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1285, Abort on wrong database password - @qu4tro
 - #790, Allow override of OpenAPI spec through `root-spec` config option - @steve-chavez
 - #1308, Accept `text/plain` and `text/html` for raw output - @steve-chavez
+- #1339, Add websearch_to_tsquery support - @herulume
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- #1243, Add websearch_to_tsquery support - @herulume
+
 ### Added
 
 ### Fixed
@@ -21,7 +23,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1285, Abort on wrong database password - @qu4tro
 - #790, Allow override of OpenAPI spec through `root-spec` config option - @steve-chavez
 - #1308, Accept `text/plain` and `text/html` for raw output - @steve-chavez
-- #1339, Add websearch_to_tsquery support - @herulume
 
 ### Fixed
 
--- a/test/Feature/AndOrParamsSpec.hs
+++ b/test/Feature/AndOrParamsSpec.hs
@@ -80,18 +80,18 @@ spec actualPgVersion =
               {"text_search_vector": "'amus':5 'fair':7 'impossibl':9 'peu':4" },
               {"text_search_vector": "'art':4 'spass':5 'unmog':7"}
             ]|] { matchHeaders = [matchContentTypeJson] }
-          get "/tsearch?or=(text_search_vector.plfts(german).Art,text_search_vector.plfts(french).amusant,text_search_vector.not.wfts(english).impossible)"
-            `shouldRespondWith` (
-                if actualPgVersion >= pgVersion112 then
-                    [json|[
-                        {"text_search_vector": "'also':2 'fun':3 'possibl':8" },
-                        {"text_search_vector": "'ate':3 'cat':2 'fat':1 'rat':4" },
-                        {"text_search_vector": "'amus':5 'fair':7 'impossibl':9 'peu':4" },
-                        {"text_search_vector": "'art':4 'spass':5 'unmog':7" }
-                    ]|] { matchHeaders = [matchContentTypeJson] }
-                else
-                    [json|{"hint":"No function matches the given name and argument types. You might need to add explicit type casts.","details":null,"code":"42883","message":"function websearch_to_tsquery(unknown, unknown) does not exist"}|] { matchStatus = 404 , matchHeaders = [matchContentTypeJson] }
-            )
+
+        when (actualPgVersion >= pgVersion112) $
+          it "can handle fts" $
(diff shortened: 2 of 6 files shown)


test/Feature/AndOrParamsSpec.hs
```diff
@@ -80,6 +80,19 @@ spec =
...
+          get "/tsearch?or=(text_search_vector.plfts(german).Art,text_search_vector.plfts(french).amusant,text_search_vector.not.wfts(english).impossible)"
+            `shouldRespondWith` (
+                if actualPgVersion >= pgVersion112 then
+                    [json|[
+                        {"text_search_vector": "'also':2 'fun':3 'possibl':8" },
+                        {"text_search_vector": "'ate':3 'cat':2 'fat':1 'rat':4" },
+                        {"text_search_vector": "'amus':5 'fair':7 'impossibl':9 'peu':4" },
+                        {"text_search_vector": "'art':4 'spass':5 'unmog':7" }
+                    ]|] { matchHeaders = [matchContentTypeJson] }
+                else
```
steve-chavez: You can use [when](https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Monad.html#v:when) instead of having to check the `else` case for version of pg  that don't support `websearch_to_tsquery`.

That would simplify all of the new tests.

steve-chavez: An example:

https://github.com/PostgREST/postgrest/blob/64dc6ab9acc8cf8242697e817cf6dcf22a41026f/test/Feature/RpcSpec.hs#L259-L265

steve-chavez:
@herulume Regarding 404 for not found operators(a function), what do you think about this, quoting [Roy Fielding](https://roy.gbiv.com/untangled/2008/rest-apis-must-be-hypertext-driven#comment-743):

*"Resources are not storage items (or, at least, they aren’t always equivalent to some storage item on the back-end). The same resource state can be overlayed by multiple resources, just as an XML document can be represented as a sequence of bytes or a tree of individually addressable nodes. Likewise, a single resource can be the equivalent of a **database stored procedure**, with the power to abstract state changes over any number of storage items"*.

herulume:
Rebased with master.

herulume:
> @herulume Regarding 404 for not found operators(a function), what do you think about this, quoting [Roy Fielding](https://roy.gbiv.com/untangled/2008/rest-apis-must-be-hypertext-driven#comment-743):
> 
> _"Resources are not storage items (or, at least, they aren’t always equivalent to some storage item on the back-end). The same resource state can be overlayed by multiple resources, just as an XML document can be represented as a sequence of bytes or a tree of individually addressable nodes. Likewise, a single resource can be the equivalent of a **database stored procedure**, with the power to abstract state changes over any number of storage items"_.

For some reason, I can't think of PG's functions as resources. 

I guess my problem is that I'm thinking in REST terms (?).
`foo?id=eq.13&name=eq.bar` uniquely identifies a resource, but `eq` by itself does not.
`rpc/boo` identifies a resource (a storage procedure) but `boo` by itself means nothing. 
I admit this doesn't differ much from `eq` identifying a function and therefore a "resource" but it causes me some confusion.

I guess I need some time to reorder my thoughts.
Thank you for that link!

test/Feature/AndOrParamsSpec.hs
```diff
@@ -80,6 +80,19 @@ spec =
               {"text_search_vector": "'amus':5 'fair':7 'impossibl':9 'peu':4" },
               {"text_search_vector": "'art':4 'spass':5 'unmog':7"}
             ]|] { matchHeaders = [matchContentTypeJson] }
+
+        when (actualPgVersion >= pgVersion112) $
+          it "can handle fts" $
```
steve-chavez: Above this test there's a `it "can handle fts"` test, that makes this one look like a duplicate.   

Maybe rename this test as `it "can handle wfts(websearch_to_tsquery)"`

herulume: Well spotted, thanks.

steve-chavez:
@herulume All looking good now!

Please add an entry to the [CHANGELOG.md](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md), then I'll merge.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1285, Abort on wrong database password - @qu4tro
 - #790, Allow override of OpenAPI spec through `root-spec` config option - @steve-chavez
 - #1308, Accept `text/plain` and `text/html` for raw output - @steve-chavez
+- #1339, Add websearch_to_tsquery support - @herulume
```
steve-chavez: Move it to the `Unreleased` section`above pls.

https://github.com/PostgREST/postgrest/blob/ea970554491557d79408ee2193d8a5f08d2984ea/CHANGELOG.md#added

steve-chavez: Also change the issue from #1339 to #1243.

herulume: Sorry for the confusion.

PR #1342, opened 2019-07-05
Windows build

robx:
This pulls in a fixed text-printer and should result in a successful Windows build.

It's branched before the jose 0.8 update since jose 0.8 causes Windows build trouble, too: https://github.com/frasertweedale/hs-jose/issues/83

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,14 +13,6 @@ environment:
 
 test: off
 
-skip_non_tags: true
-
-skip_branch_with_pr: true
-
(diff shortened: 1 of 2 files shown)


robx:
Seems appveyor refuses to build my branch here. Here's my build log: https://ci.appveyor.com/project/robx/postgrest/builds/25782811

robx:
seems this should be fixed by pulling in text-printer-0.5.0.1 instead

Issue #1336, opened 2019-07-03
Error when testing on Chrome/Firefox: text/html requested but a single column was not selected


steve-chavez:
PostgREST version: v6.0.0

### Problem

Some users might use regular browsers(chrome, firefox, etc) to explore and make quick tests of the API, but since the addition of #1330 this is no longer possible:

```
GET /projects

{"message":"text/html requested but a single column was not selected"}
```

This would also have happened for chrome if #1077 was implemented, since this is the `Accept` header that it sends by default:
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3
```
(See the `image/webp` and `image/apng` in there)

Having pgrst behave like this by default is bad for UX. So we shouldn't have `image/*` or `text/html` as alias for binary output by default. Though I think `text/plain` is probably safe.

Allowing `Content-Type` overriding isn't a solution since that doesn't make pgrst do raw output.

### Proposal

My proposal for now would be to add a new config option that lets the user specify the media types for which it would like the raw output and not have `text/html` by default. Something like:

`raw-output-media-types = [ "image/png", "image/jpeg", "text/html" ]`

This would solve this issue and #1077.

Dansvidania:
I'd like to try to implement this. As a clarification, is the intended - current - use to accept text/html and select one and only one column?

steve-chavez:
@Dansvidania Actually the use case for `text/html` was just for pg functions that returned text. It was a mistake on my part to enable it by default for all kind of binary outputs. Getting an html column wasn't intended.

So the idea as I proposed is to remove text/html from here:

https://github.com/PostgREST/postgrest/blob/ea970554491557d79408ee2193d8a5f08d2984ea/src/PostgREST/Types.hs#L61-L63

And instead let the user specify other media types on the config option I mentioned.

Also adding tests with the `Accept` header that chrome/firefox send would be good so this kind of mistakes don't happen again.

We might do a quick release for `v6.0.1` once this is fixed.

Issue #1077, opened 2018-03-01
Get binary data with image/* Accept header that is used by browsers for img tags


christiaanwesterbeek:
Thanks to #802 we can get binary output using the `Accept: application/octet-stream` header. Works fine. I can fetch my binary, which is an image using Postman:
![image](https://user-images.githubusercontent.com/465989/36872853-fcf80110-1da6-11e8-9eaf-712fd0569581.png)

But now, I want to get that image in an [img tag](https://github.com/begriffs/postgrest/issues/506#issuecomment-216045702) like so:

`<img src="http://server:3000/images?select=binary_data&id=eq.f6deb9ff-839f-426c-baa0-0f05ebbab691">`

But the result is a 200 with an empty response. Probably because the Accept header is not `application/octet-stream`, but `image/webp,image/apng,image/*,*/*;q=0.8`

![image](https://user-images.githubusercontent.com/465989/36873034-92c7b9ce-1da7-11e8-9b47-9a6a8d34253b.png)
![image](https://user-images.githubusercontent.com/465989/36873069-a955ae80-1da7-11e8-8bd9-f338a214b5d9.png)

I know I can use  a base64 data img src and I know about the pros and cons. In need the url's as img src. Later I'll setup a CDN to cache the images from the api. But for now;

What can I do to get binary output from the api with the image/* header that the browser is currently sending along with the img request?

begriffs:
Regardless of whether it's a good idea to serve images from the database, and many people advise against it, this does seem like a postgrest bug. Your Accept header included `*/*` so the server is free to make a response with `application/octet-stream`. The blank response is odd.

christiaanwesterbeek:
It would indeed be a very bad idea to serve every image for every person from the database, but I'm going to put a CDN in front of it, so it'll be every image just once within the expiration period from the database and the rest from the CDN.

> Your Accept header included */* so the server is free to make a response with application/octet-stream.

Would be great if it could. I'm sorry to say I have zero experience in Haskell. Otherwise, I'd give it a go. I'd need to find some free time first to learn the language.

> The blank response is odd.

Just did a retest right after each other.
![image](https://user-images.githubusercontent.com/465989/36909949-b6749b10-1e3f-11e8-86fd-4699c98b4cfa.png)
![image](https://user-images.githubusercontent.com/465989/36909970-c2b59f3c-1e3f-11e8-855d-ed4cedd43500.png)
[...]

PR #1349, opened 2019-07-11
Introduced raw-media-types config option 

Dansvidania:
Should solve #1336 and  #1077

It seems functional but needs tests, for now I have only been able to extract rawContentTypes to Config.hs from Types.hs

It was used as a global value in App.hs so I had to make minor changes to a few helper functions.

Posting the WIP here to get feedback on the code while I work on the tests suggested by @steve-chavez

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -66,28 +66,29 @@ import Protolude         hiding (concat, hPutStrLn, intercalate, null,
 
 -- | Config file settings for the server
 data AppConfig = AppConfig {
-    configDatabase          :: Text
-  , configAnonRole          :: Text
-  , configProxyUri          :: Maybe Text
-  , configSchema            :: Text
-  , configHost              :: Text
-  , configPort              :: Int
-  , configSocket            :: Maybe Text
+    configDatabase            :: Text
+  , configAnonRole            :: Text
+  , configProxyUri            :: Maybe Text
+  , configSchema              :: Text
+  , configHost                :: Text
+  , configPort                :: Int
+  , configSocket              :: Maybe Text
 
-  , configJwtSecret         :: Maybe B.ByteString
-  , configJwtSecretIsBase64 :: Bool
-  , configJwtAudience       :: Maybe StringOrURI
+  , configJwtSecret           :: Maybe B.ByteString
+  , configJwtSecretIsBase64   :: Bool
+  , configJwtAudience         :: Maybe StringOrURI
 
-  , configPool              :: Int
-  , configPoolTimeout       :: Int
-  , configMaxRows           :: Maybe Integer
-  , configReqCheck          :: Maybe Text
-  , configQuiet             :: Bool
-  , configSettings          :: [(Text, Text)]
-  , configRoleClaimKey      :: Either ApiRequestError JSPath
-  , configExtraSearchPath   :: [Text]
+  , configPool                :: Int
+  , configPoolTimeout         :: Int
+  , configMaxRows             :: Maybe Integer
+  , configReqCheck            :: Maybe Text
+  , configQuiet               :: Bool
+  , configSettings            :: [(Text, Text)]
+  , configRoleClaimKey        :: Either ApiRequestError JSPath
+  , configExtraSearchPath     :: [Text]
 
-  , configRootSpec          :: Maybe QualifiedIdentifier
+  , configRootSpec            :: Maybe QualifiedIdentifier
+  , configRawOutputMediaTypes :: [B.ByteString]
   }
 
 configPoolTimeout' :: (Fractional a) => AppConfig -> a
@@ -168,6 +169,7 @@ readOptions = do
         <*> (maybe (Right [JSPKey "role"]) parseRoleClaimKey <$> optValue "role-claim-key")
         <*> (maybe ["public"] splitExtraSearchPath <$> optValue "db-extra-search-path")
         <*> ((\x y -> QualifiedIdentifier x <$> y) <$> dbSchema <*> optString "root-spec")
+        <*> (fmap encodeUtf8 <$> optionalListOfText "raw-output-media-types")
 
     parseJwtAudience :: C.Key -> C.Parser C.Config (Maybe StringOrURI)
     parseJwtAudience k =
@@ -177,6 +179,12 @@ readOptions = do
           Nothing -> fail "Invalid Jwt audience. Check your configuration."
           (Just "") -> pure Nothing
           aud' -> pure aud'
+
+    optionalListOfText :: C.Key -> C.Parser C.Config [Text]
+    optionalListOfText k =
+      C.optional k (C.list C.string) >>= \case
+        Nothing -> pure []
+        Just types -> pure types
 
     reqString :: C.Key -> C.Parser C.Config Text
     reqString k = C.required k C.string
@@ -273,6 +281,9 @@ readOptions = do
           |## stored proc that overrides the root "/" spec
           |## it must be inside the db-schema
           |# root-spec = "stored_proc_name"
+          |
+          |## content types to produce raw output
+          |#raw-output-media-type=["image/png","image/jpg"]
           |]
 
 pathParser :: Parser FilePath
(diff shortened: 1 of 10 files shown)


src/PostgREST/Config.hs
```diff
@@ -168,6 +169,7 @@ readOptions = do
         <*> (maybe (Right [JSPKey "role"]) parseRoleClaimKey <$> optValue "role-claim-key")
         <*> (maybe ["public"] splitExtraSearchPath <$> optValue "db-extra-search-path")
         <*> ((\x y -> QualifiedIdentifier x <$> y) <$> dbSchema <*> optString "root-spec")
+        <*> (fmap encodeUtf8 <$> C.required "raw-output-media-type" (C.list C.string))
```
Dansvidania: I made it a required config key to make the behavior as explicit as possible, but I am not sure whats the preferred approach

steve-chavez: It definitely should not be required. Empty list [] should be default.

Also, I think `text/plain` should remain included since we always have had support for raw output for a `text` type and users have been using `application/octet-stream` for getting this.

test/io-tests/configs/app-settings.config
```diff
@@ -6,3 +6,4 @@ server-host = "127.0.0.1"
 server-port = 49421
 
 app.settings.external_api_secret = "0123456789abcdef"
+raw-output-media-types=["image/jpeg","image/png"]
```
steve-chavez: New feature should not affect other tests configs. One dedicated config should be added.

steve-chavez: You should probably have one new config with `text/html` added to the list. We already have a test for that type somewhere, and that one should pass.

src/PostgREST/Config.hs
Dansvidania: Ok. Do you mean text/plain and application/octet-stream should be included by default even if the config variable is left empty?

steve-chavez: @Dansvidania Yes, correct.

Dansvidania: @steve-chavez ok sounds good, thanks.

test/SpecHelper.hs
```diff
@@ -81,6 +81,8 @@ _baseCfg =  -- Connection Settings
             []
             -- No root spec override
             Nothing
+            -- Raw output media types
+            ["text/html"]
```
steve-chavez: What we should do is to have another config(smth like raw-media-types.config) that includes this setting, rest of the tests should run with raw-media-types of `[]`.

Then you should move this test https://github.com/PostgREST/postgrest/blob/master/test/Feature/RpcSpec.hs#L467-L482 to its own Spec and have this spec use the previous config.

steve-chavez:
Really important is that you add a couple of tests for the `Accept` header that chrome/firefox sends and make sure the result is json. That way we make sure we don't have a regression.

I mentioned that chrome sends:
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3
```
In that case pgrst would pick the `*/*` and send json accordingly.

If you have other browsers default Accept headers it would be great to test them as well.

test/SpecHelper.hs
Dansvidania: @steve-chavez yep, I am working on it at the moment. I am hoping to push this before end of day.

test/Feature/RawOutputTypesSpec.hs
```diff
@@ -0,0 +1,24 @@
...
+import Test.Hspec
+import Test.Hspec.Wai
+
+import Protolude
+import SpecHelper (acceptHdrs)
+
+spec :: SpecWith Application
+spec = describe "When raw-output-media-types config variable is missing or left empty" $ do
+  let firefoxAcceptHdrs = acceptHdrs "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+      chromeAcceptHdrs = acceptHdrs "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3"
```
steve-chavez: Really nice! This is what I also had in mind.

One thing though, I think it'd be better to test an endpoint such as `/items?id=lte.3` and make sure we get the JSON payload. Root `/` follows some different rules.

To be extra sure we could also test a GET on rpc, an example `/rpc/get_projects_below?id=3`.

src/PostgREST/Config.hs
```diff
@@ -273,6 +281,9 @@ readOptions = do
           |## stored proc that overrides the root "/" spec
           |## it must be inside the db-schema
           |# root-spec = "stored_proc_name"
+          |
+          |## content types to produce raw output
+          |#raw-output-media-type=["image/png","image/jpg"]
```
steve-chavez: Sorry, a nitpick. The config should be plural, end in `media-types` since it accepts a list.

Also, I think the name I proposed is too long. Ultimately we could shorten it to just `raw-media-types`.

Dansvidania: @steve-chavez makes sense. Will fix asap.

test/Feature/RawOutputTypesSpec.hs
Dansvidania: Ok. I'll be "afk" till the 21st but will try to fix them asap.

Thanks for the feedback :)

test/Feature/RawOutputTypesSpec.hs
```diff
@@ -0,0 +1,33 @@
...
+import Network.HTTP.Types
+import Test.Hspec
+import Test.Hspec.Wai
+import Test.Hspec.Wai.JSON
+
+import Protolude
+import SpecHelper (acceptHdrs)
+
+spec :: SpecWith Application
+spec = describe "When raw-output-media-types config variable is missing or left empty" $ do
```
steve-chavez: `raw-output-media-types` should be `raw-media-types` here.

test/Feature/HtmlRawOutputSpec.hs
```diff
@@ -0,0 +1,30 @@
...
+import Network.HTTP.Types
+import Test.Hspec         hiding (pendingWith)
+import Test.Hspec.Wai
+import Text.Heredoc
+
+import Protolude  hiding (get)
+import SpecHelper (acceptHdrs)
+
+spec :: SpecWith Application
+spec = describe "When raw-output-media-types is set to \"text/html\"" $
```
steve-chavez: Ditto. `s/raw-output-media-types/raw-media-types`

Dansvidania: @steve-chavez ah, sorry, missed those occurrencies. Fixed them now.

test/Feature/HtmlRawOutputSpec.hs
```diff
@@ -0,0 +1,30 @@
...
+import Test.Hspec.Wai
+import Text.Heredoc
+
+import Protolude  hiding (get)
+import SpecHelper (acceptHdrs)
+
+spec :: SpecWith Application
+spec = describe "When raw-media-types is set to \"text/html\"" $
+  it "can get raw output with Accept: text/html" $
+           request methodGet "/rpc/welcome.html" (acceptHdrs "text/html") ""
```
steve-chavez: Reduce indentation here please.

test/SpecHelper.hs
```diff
@@ -172,6 +177,8 @@ authHeaderJWT :: BS.ByteString -> Header
 authHeaderJWT token =
   (hAuthorization, "Bearer " <> token)
 
+acceptHeader :: BS.ByteString -> Header
```
steve-chavez: It seems this function is unused. There's also a similar one called `acceptHdrs` above.

steve-chavez:
Just realized that `text/html` should be gone from inside the codebase. So these references to `CTTextHtml` should be deleted:

https://github.com/PostgREST/postgrest/blob/c77ac249a9f8d1afb33934c0d7d8129c993fbe37/src/PostgREST/Types.hs#L26-L27

https://github.com/PostgREST/postgrest/blob/c77ac249a9f8d1afb33934c0d7d8129c993fbe37/src/PostgREST/Types.hs#L40

https://github.com/PostgREST/postgrest/blob/c77ac249a9f8d1afb33934c0d7d8129c993fbe37/src/PostgREST/Types.hs#L53

Dansvidania:
@steve-chavez thanks again for the feedback. All points should be addressed.

steve-chavez:
@Dansvidania Great work on this PR. I'm going to merge it now.

PR #1351, opened 2019-07-13
bump jose dependency to 0.8.1, to fix windows trouble

robx:
compare https://github.com/frasertweedale/hs-jose/issues/83

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -63,7 +63,7 @@ library
                     , http-types                >= 0.12.2 && < 0.13
                     , insert-ordered-containers >= 0.1 && < 0.3
                     , interpolatedstring-perl6  >= 1 && < 1.1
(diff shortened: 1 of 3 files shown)


steve-chavez:
Nice! We got a new version of jose.

Only thing missing would be to correct this test:
https://github.com/PostgREST/postgrest/blob/e639c77aa2aea8f225f2245952b084e33189d96e/test/Feature/AuthSpec.hs#L120

robx:
> Nice! We got a new version of jose.
> 
> Only thing missing would be to correct this test:
> 
> https://github.com/PostgREST/postgrest/blob/e639c77aa2aea8f225f2245952b084e33189d96e/test/Feature/AuthSpec.hs#L120

Oops, missed that. Pushed a fix!

PR #1354, opened 2019-07-21
Fix windows build

### LINKED ISSUE #1334: Unable to compile v6.0.0 on windows
I wasn't able to compile latest version on windows. This seems related to the GHC upgrade we had recently.

- Error for GHC 8.4.4: https://ci.appveyor.com/project/postgrest/postgrest/builds/25457530
- Error for GHC 8.6.5: https://ci.appveyor.com/project/steve-chavez/postgrest/builds/25506383 

These errors are on AppVeyor, but I also hit the same errors when trying to compile on a Windows Server 2019 in an ec2 instance.

Any help would be greatly appreciated. Trying to [build from source](http://postgrest.org/en/v5.2/install.html#build-from-source) on Windows and reporting here would help.

steve-chavez:
Fixes #1334. Also enabling Travis and Appveyor on every PR to ensure builds succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,5 @@
-## Travis is used for building an OSX binary on new tag releases,
-## it also compiles the project on OSX for each commit to master.
-##
-## Building the whole project can take longer than 50 minutes. Since Travis has a global timeout of 50 minutes
-## we compile for 30 minutes tops(see `gtimeout 1800` below) and quit compiling with no error.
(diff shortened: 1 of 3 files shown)


steve-chavez:
Seems appveyor is not reusing the cache.

steve-chavez:
appveyor is successfully generating the binary though I'm not sure why some libraries are being rebuilt. I guess for now this is good enough.

Issue #280, opened 2015-09-10
Extending the capabilities of &select parameter (and filtering along with it)


ruslantalpa:
What do you guys think about implementing this https://github.com/ruslantalpa/skin (ignore the code quality and the specific implementation details)
The details are in the README

Right now you can do this
```
&select=col1,col2::int,col3->>sub1::json,col3->sub1->>sub2::int
```

I propose to add the capabilities to do this
```
&select=id,name,projects(id,name,tasks(id,name))
```
and this
```
&id=eq.1&projects.tasks.completed=eq.true
```

This would allow the capability of getting back results that look like this
```
[  
   {  
      "id":1,
      "name":"Windows 7",
      "clients":{  
         "id":1,
         "name":"Microsoft"
      },
      "tasks":[  
         {  
            "id":1,
            "name":"Design w7"
         },
         {  
            "id":2,
            "name":"Code w7"
         }
      ],
      "users":[  
         {  
            "id":1,
            "name":"Angela Martin"
         },
         {  
            "id":3,
            "name":"Dwight Schrute"
         }
      ]
   }
]

```

begriffs:
Wow, this is really great work! Would love to add this kind of functionality.

Some questions:
* If we're getting a list of movies say and each movie has actors, do we allow the user to include the actors inside each movie? The response could get big. Or do we allow embedding only when requesting a single movie? (Possible by checking if primary key is specified in filter)
* How do we limit the count of embedded resources? Maybe interpret a request range header as applying to the embedded resource rather than a top-level list. This also argues for allowing embeds only on a request for a single top level result.

ruslantalpa:
1. Allow embedding of actors. Yes the response is bigger, but if the data is needed It would have to be fetched anyway (in multiple requests). I would say it's the responsibility of the client to find the best/optimal way to get the data. For example, in case of projects/tasks structure, there is no duplication of data (tasks are unique to project) so it can be efficiently fetched in one request. In case of movies/actors i would make two requests, a) all actors, b) all movies, and include actor info, but only the id, then combine the data on the client side.
2. The limit would refer to the top level items. Most of the time, the use case is you are displaying a list of items (projects, movies), the list is big so you need to paginate it (this is where limit comes in) but each item has some related info that you need (category of the movie, client of the project) so it is great to be able to get that in one request. Anyway, it would be next ti impossible (i think) to get the count of nodes in the tree and limit/offset does not even make sense in a tree like structure.

[...]

PR #1358, opened 2019-07-24
Support PATCH for self-referencing entities

### LINKED ISSUE #1301: PATCH fails with self-referencing entity
Trying to work out whether this is a bug or me being an idiot.

Given an entity like:

```
create table website_category (
    id uuid primary key default uuid_generate_v1 ()
    , name text
    , parent_category_id uuid references website_category(id)
);
```

A GET request like `/website_category?id=eq.a9b44532-5cf6-11e9-bad0-0242c0a8f008&select=parent_category(id,name)` works perfectly, returning the nested parent category as expected.

However, the same PATCH, lets say with body `{"name": "new name"}`, fails with:

```
{"message":"Could not find foreign keys between these entities, No relation found between pg_source and parent_category"}
```

This is the first time I've seen this and my app has tonnes of these types of calls that work fine.  This is the first entity I have that references itself, so I'm wondering if that is tripping up Postgrest.

herulume:
Hey
This PR aims to close #1301.

### Code duplication
I know there's a bit of code duplication but I think the `findRelation` function is easier to read this way. 

### Doubt about Relation
I was a bit confused about the `Relation` values in the `findRelation` function. 
When patching from something like `/wc?id=eq.0&select=parent(id, name)`, where `parent_id` refers to another table, say table `xyz`, the relation that is found by `findRelation` has the following:
- relTable name: xyz
- relFTable name: pg_source
- FColumns table name: wc
- relType: Parent

Shouldn't this be `relType: Child`? Since the foreign key is in `wc`. 
I'm pretty sure I'm confusing basic stuff but wanted to clear this. Is there any resource/PR/issue to read more about the `Relation` type?

### DIFF
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -162,6 +162,14 @@ findRelation schema allRelations nodeTableName parentNodeTableName relationDetai
           length relFColumns == 1 &&
           -- match common foreign key names(table_name_id, table_name_fk) to table_name
           (toS ("^" <> colName (unsafeHead relFColumns) <> "_?(?:|[iI][dD]|[fF][kK])$") :: BS.ByteString) =~ (toS nodeTableName :: BS.ByteString)
+        ) ||
+
+        -- self relations
+        (
+          tableName relTable == tableName relFTable &&
+          length relFColumns == 1 &&
+          -- match common foreign key names(table_name_id, table_name_fk) to table_name
+          (toS ("^" <> colName (unsafeHead relFColumns) <> "_?(?:|[iI][dD]|[fF][kK])$") :: BS.ByteString) =~ (toS nodeTableName :: BS.ByteString)
         )
 
         -- (request)        => project_id { ..., client_id{...} }
--- a/test/Feature/InsertSpec.hs
+++ b/test/Feature/InsertSpec.hs
@@ -482,23 +482,14 @@ spec actualPgVersion = do
           [json| [{ a: "keepme", b: null }] |]
           { matchHeaders = [matchContentTypeJson] }
 
-      it "meh" $ do
-        g <- get "/s_web_content?id=eq.0"
-        liftIO $ simpleHeaders g
-          `shouldSatisfy` matchHeader "Content-Range" "0-0/*"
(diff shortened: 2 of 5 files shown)


steve-chavez:
@herulume Hey!

Hmm. The change you made is not fixing the bug for me.

This is the test that should succeed:

```bash
http PATCH "http://localhost:3000/web_content?id=eq.0&select=name,web_content(name)" "Prefer:return=representation" <<JSON
{"name": "tardis-patched"}
JSON
```
And the response should be:

```json
[{"name":"tardis-patched","web_content":[{"name":"fezz"}]}]
```

Right now the response is:

```
{
    "message": "Could not find foreign keys between these entities, No relation found between pg_source and web_content"
}
```

steve-chavez:
> Is there any resource/PR/issue to read more about the Relation type?

Tbh, that part needs a refactor, definitely not easy to understand. I'm constantly confused by the term Relation since it conflicts with PostgreSQL/Relational Algebra terminology. I think Relationship would've been a better name.

Now, I think this is the first issue regarding the Relation code https://github.com/PostgREST/postgrest/issues/280, from there I'd recommend looking at https://github.com/ruslantalpa/skin, perhaps also ask @ruslantalpa for further explanation.

herulume:
I'm going to investigate further. 
Thank for those links!

steve-chavez:
@herulume Sorry for not being more helpful here.

I've taken some of your work and with that opened #1391.

Basically, `findRelation` wasn't the root cause of the bug since it already worked for self-reference entities. Some tests(for GET) are available here:

https://github.com/PostgREST/postgrest/blob/186381bab29b8dc8e00dafb978f2b8d7830323b8/test/Feature/QuerySpec.hs#L581-L687

The root cause was the `toSourceRelation` function:
https://github.com/PostgREST/postgrest/blob/ae9e27a0c71c6793d998c79141c4bdaccd6edeeb/src/PostgREST/DbRequestBuilder.hs#L95-L105

That's what I ended up modifying to solve the issue in https://github.com/steve-chavez/postgrest/commit/88d64d115ec11588fda0e6741229de5a7574a0c8.

I've added comments to the functions in #1391.  If you have any questions, we can discuss it further there.

steve-chavez:
Superseded by #1391. Thanks for the work done here @herulume.

Issue #1353, opened 2019-07-18
Enum array support


bwbroersma:
Enum looks nice in Swagger, enum arrays do not.
Here the changes that work for me for nice Swagger output, but it might not be correct for the OpenAPI standard:
https://github.com/openstate/postgrest/commit/74c5d9704233687efffda407abe7d5ef778b544f

bwbroersma:
Ok the previous commit / provided solution is not the correct way to do it, this will also result in an
**Example Value**
```json
{
  "the-enum-array-field": "first-value"
}
```
were we would want more something like
```json
{
  "the-enum-array-field": ["first-value"]
}
```

bwbroersma:
The correct Swagger to output is:
```json
{
  "format": "schema.typename[]",
  "type": "array",
  "items": {
    "type": "string",
    "enum": ["first-value", "second-value"]
  }
}
```

kavitang-sa:
Any update on this issue? We have an enum array column and would want the above mentioned response from the OpenApi response.

laurenceisla:
@kavitang-sa I don't think this will be fixed here (in the core repository), but it will be addressed in the [`postgrest-openapi` ](https://github.com/PostgREST/postgrest-openapi) project.

kavitang-sa:
Thanks @laurenceisla I'll create an issue there to address this.

kavitang-sa:
> > I'll create an issue there to address this.
> 
> No need to do that, we will track it here.

Great, I just read this after creating an issue. I'll remove it, thank you @wolfgangwalther !

PR #1362, opened 2019-07-24
[WIP] Enum arrays - see #1353.

bwbroersma:
See #1353.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -447,12 +447,14 @@ allColumns tabs =
                     CASE
                         WHEN t.typtype = 'd'::"char" THEN
                         CASE
+                            WHEN bt.typelem <> 0::oid AND bt.typlen = (-1) AND a.atttypmod = (-1) THEN format_type(a.atttypid, a.atttypmod)
                             WHEN bt.typelem <> 0::oid AND bt.typlen = (-1) THEN 'ARRAY'::text
                             WHEN nbt.nspname = 'pg_catalog'::name THEN format_type(t.typbasetype, NULL::integer)
                             ELSE format_type(a.atttypid, a.atttypmod)
                         END
                         ELSE
                         CASE
+                            WHEN t.typelem <> 0::oid AND t.typlen = (-1) AND a.atttypmod = (-1) THEN format_type(a.atttypid, a.atttypmod)
                             WHEN t.typelem <> 0::oid AND t.typlen = (-1) THEN 'ARRAY'::text
                             WHEN nt.nspname = 'pg_catalog'::name THEN format_type(a.atttypid, NULL::integer)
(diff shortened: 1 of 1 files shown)


src/PostgREST/DbStructure.hs
```diff
@@ -555,7 +557,7 @@ allColumns tabs =
         JOIN pg_enum e ON t.oid = e.enumtypid
         JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace
         GROUP BY s,n
-    ) AS enum_info ON (info.udt_name = enum_info.n)
+    ) AS enum_info ON (trim(leading '_' from info.udt_name) = enum_info.n)
```
steve-chavez: Hmm. Would like to see an example output of this, not sure why the `trim leading` is needed.

bwbroersma: The `udt_name` (actually `pg_type`.`typname` [pg_type docs](https://www.postgresql.org/docs/current/catalog-pg-type.html)) for enum arrays has a leading `_`, see the [`CREATE TYPE` docs](https://www.postgresql.org/docs/current/sql-createtype.html#id-1.9.3.94.5.9):
> Whenever a user-defined type is created, PostgreSQL automatically creates an associated array type, whose name consists of the element type's name prepended with an underscore, and truncated if necessary to keep it less than `NAMEDATALEN` bytes long. (If the name so generated collides with an existing type name, the process is repeated until a non-colliding name is found.) This implicitly-created array type is variable length and uses the built-in input and output functions `array_in` and `array_out`. The array type tracks any changes in its element type's owner or schema, and is dropped if the element type is.

But your comment about 'Example output' is right in another way at least for the Swagger UI: https://github.com/PostgREST/postgrest/issues/1353#issuecomment-515175906

bwbroersma: Ok before Postgresql `8.3` this was a valid solution, but my solution is no longer valid, see the [`CREATE TYPE` docs the Notes section](https://www.postgresql.org/docs/current/sql-createtype.html#SQL-CREATETYPE-NOTES):
> Before PostgreSQL version 8.3, the name of a generated array type was always exactly the element type's name with one underscore character (`_`) prepended. (Type names were therefore restricted in length to one less character than other names.) While this is still usually the case, the array type name may vary from this in case of maximum-length names or collisions with user type names that begin with underscore. Writing code that depends on this convention is therefore deprecated. Instead, use `pg_type.typarray` to locate the array type associated with a given type.
>
> It may be advisable to avoid using type and table names that begin with underscore. While the server will change generated array type names to avoid collisions with user-given names, there is still risk of confusion, particularly with old client software that may assume that type names beginning with underscores always represent arrays.

monacoremo:
Seeing as this would only work for Postgres < 8.3, I'm closing this PR for now and leaving the issue open.

Issue #1369, opened 2019-08-02
Change `raw-media-types` parameter to a comma separated text instead of list


steve-chavez:
Currently `raw-media-types` cannot be used in Docker. This is because of https://github.com/PostgREST/postgrest/issues/1367 and https://github.com/robx/configurator-pg/issues/3.

Simplest solution would be to change the param to:
```
raw-media-types = "text/html, image/png"
```
Instead of having the current:
```
raw-media-types = ["text/html", "image/png"]
```

We'll have to make sure commas don't conflict with media types syntax, otherwise we'd need a different separator. From a quick glance at https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types and https://en.wikipedia.org/wiki/Media_type it seems commas are safe.

PR #1370, opened 2019-08-04
Change raw-media-types to accept a string of comma separated MIME types

Dansvidania:
should solve #1369

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -167,9 +167,9 @@ readOptions = do
         <*> pure False
         <*> (fmap (fmap coerceText) <$> C.subassocs "app.settings" C.value)
         <*> (maybe (Right [JSPKey "role"]) parseRoleClaimKey <$> optValue "role-claim-key")
-        <*> (maybe ["public"] splitExtraSearchPath <$> optValue "db-extra-search-path")
+        <*> (maybe ["public"] splitOnCommas <$> optValue "db-extra-search-path")
         <*> ((\x y -> QualifiedIdentifier x <$> y) <$> dbSchema <*> optString "root-spec")
-        <*> (fmap encodeUtf8 <$> optionalListOfText "raw-media-types")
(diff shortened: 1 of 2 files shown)


src/PostgREST/Config.hs
```diff
@@ -180,11 +180,11 @@ readOptions = do
...
 
-    optionalListOfText :: C.Key -> C.Parser C.Config [Text]
-    optionalListOfText k =
-      C.optional k (C.list C.string) >>= \case
+    optionalCommaSeparatedList :: C.Key -> C.Parser C.Config [Text]
+    optionalCommaSeparatedList k =
+      C.optional k C.string >>= \case
         Nothing -> pure []
-        Just types -> pure types
+        Just commaSeparatedValues -> pure $ strip <$> splitOn "," commaSeparatedValues
```
steve-chavez: Down below there's a `splitExtraSearchPath` function which has the same logic. Perhaps that function can be renamed and reused here.

src/PostgREST/Config.hs
```diff
@@ -283,7 +283,7 @@ readOptions = do
           |# root-spec = "stored_proc_name"
           |
           |## content types to produce raw output
-          |# raw-media-types=["image/png","image/jpg"]
+          |# raw-media-types="image/png,image/jpg"
```
steve-chavez: nitpick: Add a space between the types.`"image/png, image/jpg"`

src/PostgREST/Config.hs
Dansvidania: I was in doubt rregarding whether refactoring that code would be out of the scope of this PR, will fix. Thanks for the feedback

src/PostgREST/Config.hs
Dansvidania: Ok

steve-chavez:
LGTM! Regarding refactoring, I think we should do it whenever we can.

Merging..

PR #1372, opened 2019-08-06
Reword error message for BinaryFieldError

Dansvidania:
I was confused by the wording of the error and I hope this improves it.

@steve-chavez would the extraction of error messages to a resource file be welcomed? 

It would make it unnecessary for simple changes like this to make also changes on existing tests

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -224,7 +224,7 @@ instance JSON.ToJSON SimpleError where
   toJSON GucHeadersError           = JSON.object [
     "message" .= ("response.headers guc must be a JSON array composed of objects with a single key and a string value" :: Text)]
   toJSON (BinaryFieldError ct)          = JSON.object [
-    "message" .= ((toS (toMime ct) <> " requested but a single column was not selected") :: Text)]
+    "message" .= ((toS (toMime ct) <> " requested but more than one column was selected") :: Text)]
   toJSON ConnectionLostError       = JSON.object [
     "message" .= ("Database connection lost, retrying the connection." :: Text)]
 
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -950,7 +950,7 @@ spec actualPgVersion = do
     it "fails if a single column is not selected" $ do
       request methodGet "/images?select=img,name&name=eq.A.png" (acceptHdrs "application/octet-stream") ""
         `shouldRespondWith`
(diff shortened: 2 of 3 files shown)


steve-chavez:
Maybe having a link to the docs in the error would help even more. But perhaps that should be part of another PR. Change looks good.

> @steve-chavez would the extraction of error messages to a resource file be welcomed?
It would make it unnecessary for simple changes like this to make also changes on existing tests

You mean you want to make error messages configurable by the user?

steve-chavez:
Or is the idea of the resource file to make it more easier for us postgrest developers to keep the codebase DRY?

Dansvidania:
> Or is the idea of the resource file to make it more easier for us postgrest developers to keep the codebase DRY?

@steve-chavez This. The idea would be to extract error messages to values in a separate module, and have Errors.hs and also tests use those values. Ultimately, as you said, it is for DRY. 

That is unless you see a need for user configurable error messages. So far in my use of PostgREST I have not encountered such a need.

Dansvidania:
> Maybe having a link to the docs in the error would help even more. But perhaps that should be part of another PR. Change looks good.

This indeed is even better. If you approve the initiative described above, I could start working on DRY-ing the error messages and also including links to the docs.

steve-chavez:
Check this example of error messages here: https://github.com/PostgREST/postgrest/blob/f3546a15bec1a0ad441c427cd93eca1eadcb85f0/test/Feature/AndOrParamsSpec.hs#L238-L260

Would we have to also put PostgreSQL messages in the resource file?

steve-chavez:
I imagine that if we'd have every error message in a different file that would make tests less readable. Since for every test we'd have to jump to another file to see what the error really is. That kind of reminds me of some big java projects where every user message is in another resource file, and those "indirections" made projects harder to maintain.

Don't get me wrong, I welcome every change to improve the codebase. But we'd have to think if the pro of DRY(we rarely change error messages, and that only leads us to correct a few tests) would really outweigh the con of having a harder to understand test suite(and sometimes we even recommend end users reading them to understand some functionality).

steve-chavez:
Hmmm. Or maybe we could just limit ourselves to putting postgrest error messages in the resources file.

Anyway, let me know what you think.

Dansvidania:
I was definitely only thinking about PostgREST errors, but I think you might be right about the limited advantages of DRY for something that changes rarely. 

I would close this change here then, if that works for you.

steve-chavez:
Change here is good. Also adding links to docs in Error.hs would be helpful. Could be done in other PR.

Merging..

PR #1375, opened 2019-08-09
Allow select in OpenAPI for POST requests

nbouscal:
If you use `return=representation`, `select` is useful in `POST` requests.
It already works correctly, but is missing from the OpenAPI spec.

### DIFF
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -218,7 +218,7 @@ makePathItem (t, cs, _) = ("/" ++ unpack tn, p $ tableInsertable t)
         )
       )
     postOp = tOp
-      & parameters .~ map ref ["body." <> tn, "preferReturn"]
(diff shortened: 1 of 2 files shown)


steve-chavez:
I see your point. Could you add a test to https://github.com/PostgREST/postgrest/blob/master/test/Feature/StructureSpec.hs to confirm?

nbouscal:
I've changed the commit to include this in the tests, but I haven't managed to get the test suite running locally yet (I'm getting `spec: fd:35: hGetContents: invalid argument (invalid byte sequence)`). Hopefully I got it right anyway, if not I'll fix it tomorrow. Thanks!

steve-chavez:
@nbouscal Looking good now!

I remember fixing the hGetContents error some time ago by setting the LANG env var as mentioned here https://stackoverflow.com/a/30436634/4692662

PR #1380, opened 2019-08-19
Go back to converting plus "+" to space

### LINKED ISSUE #1348: PostgREST 6.0.0 breaks filtering by strings with space as +
### Environment

* PostgreSQL version: postgrest/postgrest
* PostgREST version: 6.0.0
* Operating system: Windows Home 64-bit

### Description of issue

I use Axios to query PostgREST from Node.JS and browser. In 5.2.0, this request (made by Axios) works, i.e. it returns a row that has systemName='Foo - Bar':

```
http://localhost:3000/some_table?systemName=eq.Foo+-+Bar
```

I have tried upgrading to 6.0.0 today and in 6.0.0 this stops working and returns an empty result set. If you replace the "+" by %20, it works again.

Maybe this was caused by #1288? 

Anyhow, I'm not sure if this is an issue with Axios encoding the request wrongly, but according to this, it should be valid: https://stackoverflow.com/questions/2678551/when-to-encode-space-to-plus-or-20

Any ideas?

steve-chavez:
Fixes #1348. Not doing this conflicts with some http clients and proxies.

Use the alternative url-safe character `!` instead for disambiguating resource embedding(https://github.com/PostgREST/postgrest/pull/918).

I chose `!` because is part of the group of url-safe characters according to [RFC 1738](https://www.ietf.org/rfc/rfc1738.txt).
 
> Thus, only alphanumerics, the special characters "$-_.+!*'(),", and
> reserved characters used for their reserved purposes may be used
> unencoded within a URL.

Currently we use all of the special characters except for `$` and `!`.
We can't use `$`  because it can be part of a [pg identifier](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)(having a column named `my$col$` is admitted without quoting) so that only left us with `!`.


steve-chavez:
After merging this I think we can release a `6.0.2` version, which should be stable now.

Issue #1378, opened 2019-08-14
Estimated count


LorenzHenk:
> In future versions we will support Prefer: count=estimated to leverage the PostgreSQL statistics tables for a fast (and fairly accurate) result.
> -- Cite from http://postgrest.org/en/v5.2/admin.html

We would really like to have that feature implemented!

#### `count=estimated`

To extract the estimated count from a explain call, you can use `EXPLAIN (FORMAT JSON) $query`.
`FORMAT JSON` makes extracting the planned row count easier (i.e. `explain_result["QUERY PLAN"][0]["Plan"]["Plan Rows"]`).
The explain call would have to be done *without* `limit` and `offset`.

#### `count=estimated-overfetched`

When you are interested in the count, the relative error is important.
If you have an estimated count of `1000000` and the real count is `1001000`, the error is small enough to be ignored. 
But with an estimated count of `7`, a real count of `28` would be a huge misprediction.

In general, when having smaller row-counts, the estimated count should be as close to the real count as possible. 
How about actually counting as long as the row count is `<= 1000` and using the estimated value from `EXPLAIN` only if the row count exceeds `1000`?
Note: `1000` is just a placeholder threashold - it just has to be small enough to not significantly increase the response time. This number would have to be adapted to the `limit` used for the query.

This approach would have the following implications:
- The query would not use the given limit, but instead use our threshold (`1000` from above) to check how many rows actually exist. If the resource you want to query is a view containing a function call, this function will be called more often than expected by the user supplying a limit. Thus, this `estimated-overfetched` operation must only be used by users who know the implications
- `EXPLAIN` would only have to be called if the threshold is reached

Related issues: #255 #273

steve-chavez:
Definitely a long promised feature.

@LorenzHenk I'll think about this and return with some feedback once I'm done with the docs for our latest version https://github.com/PostgREST/postgrest-docs/pull/239.

begriffs:
> To extract the estimated count from a explain call, you can use
> `EXPLAIN (FORMAT JSON) $query`.  `FORMAT JSON` makes extracting the
> planned row count easier (i.e. `explain_result["QUERY
> PLAN"][0]["Plan"]["Plan Rows"]`).  The explain call would have to be
> done *without* `limit` and `offset`.

I agree it makes sense to use EXPLAIN to get an estimate, since it'll
handle queries with arbitrary filters.

> When you are interested in the count, the relative error is important.
> If you have an estimated count of `1000000` and the real count is
[...]

PR #1383, opened 2019-08-31
Add HEAD http method support

steve-chavez:
For #1378.

- ~~Added support for `Prefer: count=estimate` for obtaining an EXPLAIN statement "Plan Rows". This only works for `GET /table` now. It's possible to extend it and support  `GET /rpc/func` later(it would involve changing how #1266 was implemented).~~ This will be finished on a next PR.

- Added support for HEAD requests. Basically a headers-only GET. This is for better HTTP compliance and will help later on when we focus on caching. Also helps in debugging headers with `curl -I`.

Also includes some reorganization around modules and some changes to the test suite to make it run faster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1383, Add support for HEAD request - @steve-chavez
+
 ### Fixed
 
 ## [6.0.2] - 2019-08-22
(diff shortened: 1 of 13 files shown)


steve-chavez:
Btw, I refrained from having some headers work with HEAD and not GET because according to [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4):

**The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request.**

SHOULD means optional, but I thought it'd be better to follow the recommendation.

steve-chavez:
I think this is fine. I'd like to make the `count=estimated` feature uniform and also support it for GET RPC but it will involve some unrelated changes.

So I'll add the HEAD support in the CHANGELOG and finish the `count=estimated` on another PR.

Issue #1264, opened 2019-03-28
Stored Procedures RPC call receiving CSV upload doesn't work


dalthon:
### Environment

* PostgreSQL version: postgres:11.0-alpine
* PostgREST version: postgrest/postgrest:v5.2.0-3900baa
* Operating system: macOS Mojave

### Description of issue

I'm not sure if it's not supported and I'm trying to do something wrong, but I was trying to call a stored procedure in a bulk way, like in the insert.

But it didn't worked as I expected. So, to reproduce it:

As described in [Stored Procedures doc](http://postgrest.org/en/v5.2/api.html#stored-procedures), I created the following `add_them` procedure:

```sql
CREATE FUNCTION add_them(a integer, b integer)
RETURNS integer AS $$
 SELECT a + b;
$$ LANGUAGE SQL;
```

Then tried to call it in a bulk, like on bulk inserts with this HTTP request:

```
POST /rpc/add_them HTTP/1.1
Host: 0.0.0.0:3000
Accept: application/json, application/vnd.pgrst.object+json
Content-Type: text/csv
cache-control: no-cache

a,b
1,2
3,4
5,6
```

Then I received this response:

```javascript
{
    "hint": null,
    "details": null,
    "code": "21000",
    "message": "more than one row returned by a subquery used as an expression"
}
```

Looking at postgres log, I saw this error message:

```
2019-03-28 14:20:50.294 UTC [281] ERROR:  more than one row returned by a subquery used as an expression
2019-03-28 14:20:50.294 UTC [281] STATEMENT:
	       WITH _args_record AS ( SELECT * FROM json_to_recordset($1) AS _("a" integer, "b" integer) ),
	       pg_source AS (
	         SELECT "public"."add_them"("a" := (SELECT "a" FROM _args_record), "b" := (SELECT "b" FROM _args_record))
	       )
	       SELECT
	         null::bigint AS total_result_set,
	         1 AS page_total,
	         (row_to_json(_postgrest_t)->'add_them')::character varying AS body,
	         coalesce(nullif(current_setting('response.headers', true), ''), '[]') AS response_headers
	       FROM (SELECT "pg_source".* FROM "pg_source"    ) _postgrest_t;
```

So looking at this quey on debug, I believe that the correct query should be this:

```sql
WITH _args_record AS ( SELECT * FROM json_to_recordset($1) AS _("a" integer, "b" integer) ),
pg_source AS (
[...]

PR #1385, opened 2019-09-09
Make costly bulk call query optional. Add query cost tests.

steve-chavez:
While adding support for https://github.com/PostgREST/postgrest/issues/1264 I didn't realize the EXPLAIN cost of calling a proc increased by almost two orders of magnitude(see [this](https://github.com/steve-chavez/postgrest/blob/optional-bulk-call/test/QueryCost.hs#L49-L54) and [this](https://github.com/steve-chavez/postgrest/blob/optional-bulk-call/test/QueryCost.hs#L30-L34)). This happened basically because bulk RPC call generates a LATERAL query and this was made the default for all RPC calls. That's not good considering bulk calls are rare.

So I've introduced the `Prefer: params=multiple-objects` headers to make bulk calls explicit. If a json array payload is sent to the procedure, only the first element will be considered by default(this prevents the SQL error mentioned in #1264).

Also added a `spec-querycost` test suite that will come in handy for preventing costly queries. The other sql queries in `QueryBuilder.hs` can be added there later.

Moved the `Hasql` decoder/encoder/statement logic to its own module. Now the QueryBuilder module is only in charge of generating `SqlQuery` types(executable sql) and its private module should only generate `SqlFragment`(not executable sql). I'm thinking all of this reorganization could help find ways to simplify the code later.


src/PostgREST/ApiRequest.hs
```diff
@@ -130,10 +131,12 @@ userApiRequest schema rootSpec req reqBody
...
       , iPreferRepresentation = representation
-      , iPreferSingleObjectParameter = singleObject
+      , iPreferParameters = if hasPrefer (show SingleObject)    then Just SingleObject else
+                            if hasPrefer (show MultipleObjects) then Just MultipleObjects
+                            else Nothing
       , iPreferCount = hasPrefer "count=exact"
-      , iPreferResolution = if hasPrefer (show MergeDuplicates) then Just MergeDuplicates
-                            else if hasPrefer (show IgnoreDuplicates) then Just IgnoreDuplicates
+      , iPreferResolution = if hasPrefer (show MergeDuplicates)  then Just MergeDuplicates else
+                            if hasPrefer (show IgnoreDuplicates) then Just IgnoreDuplicates
```
Dansvidania: nitpick : wouldn't guards be better for iPreferParameters and iPreferResolution?
```
iPreferResolution
  | hasPrefer (show MergeDuplicates) = Just MergeDuplicates
  | hasPrefer (show IgnoreDuplicates) = Just IgnoreDuplicates
  | otherwise = Nothing
```

src/PostgREST/Types.hs
```diff
@@ -56,11 +56,26 @@ decodeContentType ct = case BS.takeWhile (/= BS.c2w ';') ct of
...
+type SqlFragment = Text
+
 data PreferResolution = MergeDuplicates | IgnoreDuplicates deriving Eq
 instance Show PreferResolution where
   show MergeDuplicates  = "resolution=merge-duplicates"
   show IgnoreDuplicates = "resolution=ignore-duplicates"
 
+data PreferParameters =
+  -- | Pass all parameters as a single json object to a stored procedure
+  SingleObject |
```
Dansvidania: sorry another nitpick: why not the as follows?
reason: readability
``` 
data PreferParameters
  = SingleObject -- Pass all parameters as a single json object to a stored procedure
  | MultipleObjects -- Pass an array of json objects as params to a stored procedure
   deriving Eq

instance Show PreferParameters where
  show SingleObject    = "params=single-object"
  show MultipleObjects = "params=multiple-objects"
```

src/PostgREST/QueryBuilder.hs
```diff
@@ -145,9 +134,54 @@ requestToQuery schema _ (DbMutate (Delete mainTbl logicForest returnings)) =
...
+
+    sourceBody :: SqlFragment
+    sourceBody
+      | paramsAsMulitpleObjects =
+          if returnsScalar
+            then "SELECT " <> callIt <> " AS pgrst_scalar FROM pgrst_args"
+            else unwords [
+              "SELECT pgrst_lat_args.*",
+              "FROM pgrst_args,",
+              "LATERAL ( SELECT * FROM " <> callIt <> " ) pgrst_lat_args" ]
```
Dansvidania: again nitpick for readability, I hope you do not mind
``` 
else unwords [ "SELECT pgrst_lat_args.*"
                      , "FROM pgrst_args,"
                      , "LATERAL ( SELECT * FROM " <> callIt <> " ) pgrst_lat_args"
                      ]
```

src/PostgREST/ApiRequest.hs
steve-chavez: I see what you mean. Unfortunately GHC doesn't allow using guards in [record syntax](https://en.wikibooks.org/wiki/Haskell/More_on_datatypes#Named_Fields_.28Record_Syntax.29) constructor fields.

I wanted to keep the logic close to the named field since it's short and that way avoid having an extra variable.

Another choice could be using [MultiWayIf](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-MultiWayIf). It would look like:

```haskell
iPreferResolution = 
  if | hasPrefer (show MergeDuplicates)  -> Just MergeDuplicates
     | hasPrefer (show IgnoreDuplicates) -> Just IgnoreDuplicates
     | otherwise                         -> Nothing
```

What do you think, does that make it clearer? Or it'd be better to just have another variable and use regular guards?

Dansvidania: I was not aware of that limitation of the record syntax.

I do think the MultiWayIf is clearer and I would prefer it over the if-then-else syntax.

src/PostgREST/Types.hs
Dansvidania: I came across the conventions only after posting the comment, again thanks for the link, another thing I learned :)

src/PostgREST/QueryBuilder.hs
steve-chavez: I think here you mean aligning to comma first(your snippet is somehow misaligned). I agree, I'll change it.

Dansvidania: Correct, sorry for the unclear comment. I still cannot get the commas to align, for some reason.

steve-chavez:
Also made the change on the latest version of the docs http://postgrest.org/en/latest/api.html#bulk-call.

steve-chavez:
A really slow query reported thanks to the LATERAL https://gitter.im/begriffs/postgrest?at=5dd3c8cb68ad1c4a0ff3b97b.

It was a good idea to add the query cost tests.

PR #1386, opened 2019-09-14
Add support for `Prefer: count=planned/estimated` on GET /table

steve-chavez:
This should solve #1378.

For including some reproducible tests when using EXPLAIN, I've used ROWS from [CREATE FUNCTION](https://www.postgresql.org/docs/11/sql-createfunction.html) and wrapped the function in a VIEW.


steve-chavez:
Proposed a change of the headers names in https://github.com/PostgREST/postgrest/issues/1378#issuecomment-531506803.

Issue #1389, opened 2019-09-23
M2M embedding not working with a stored procedure


steve-chavez:
Reported in https://gitter.im/begriffs/postgrest?at=5d71ac3a50508949d3e416f5.

For reproducing this with our [test schema](https://github.com/PostgREST/postgrest/blob/master/test/fixtures/schema.sql) create the following stored procedure:

```sql
CREATE FUNCTION getallusers() RETURNS SETOF users AS $$
SELECT * FROM test.users;
$$ LANGUAGE sql;
```

And then do:

```json
http GET "localhost:3000/rpc/getallusers?select=*,tasks(*)"

{
    "code": "42P01",
    "details": null,
    "hint": "There is an entry for table \"pg_source\", but it cannot be referenced from this part of the query.",
    "message": "invalid reference to FROM-clause entry for table \"pg_source\""
}
```

PR #331, opened 2015-10-26
Code refactoring POST/PUT/PATCH to be more like GET and use internal data type

ruslantalpa:
So a lot has changed and it is hard to see from the diff what exactly so here is the summary.

The point of all this is to have all the paths use the internal data type in order to generate queries. This will later allow extending the capabilities of the insert/update paths.
While implementing this i also came across a way to make all the queries (the one that wraps the core query) be similar and that gave the ability to reuse some sql fragments (count,as json, as csv).

Currently only the GET/INSERT paths have been rewritten and UPDATE will come soon.
A few new capabilities for insert, ability so send bulk insert in json format also and if return representation is set, the new records will be sent back in json/csv format as requested.

If a single record is inserted, the location header is also calculated directly in the db eliminating the need to decode the response in order to generate it.

I will also add the capability for the &select parameter to shape the response for insert in the same way it does for get.
After adding the ability to return the inserted result in csv format i encountered a few corner cases which will need work, specifically the NULL values and Boolean (the cause for one test failing at the moment)

When this is done, most of the "query transform" functions will be gone, their place will be taken by "query fragments" that are mixed and matched directly in App.hs

Please start reviewing this and see if some things can be implemented in a nicer way

PR #1391, opened 2019-09-24
Allow self join resource embedding on PATCH

### LINKED ISSUE #1301: PATCH fails with self-referencing entity

steve-chavez:
Fixes #1301. Continuing the work on #1358.

The main problem was in the `toSourceRelation` function, which was added in this PR https://github.com/PostgREST/postgrest/pull/331. Specifically in https://github.com/PostgREST/postgrest/commit/f02b8381eaeb8b1a26872e8be4e1c9f8830201db#diff-0721c0195214a5855036d6df3d06e076R488-R493.

The first line of that function contained some logic to replace the source tableName from the `Relation` to `pg_source` which was making the self reference case not be found by `findRelation`.

The third line of the function seems to contain some logic for handling the M2M case, however since we have https://github.com/PostgREST/postgrest/issues/1389 reported it probably means that logic never worked.

So I removed the unnecessary lines in that function(which actually didn't cause any test to fail) and just left it with the logic that converted the Relation fk tableName to `pg_source`(which actually makes sense to me). My guess is that the other lines weren't thoroughly reviewed and just something that slipped by.

Also did some refactoring and added comments, the `DbRequestBuilder` module is probably what needs it the most now. Seeing the `mapMs` in there makes me think that it could be somehow simplified with `Traversable`. I'm also hoping to refactor the `Relation` type in a next PR, it also seems backward to me as pointed in https://github.com/PostgREST/postgrest/pull/1358#issue-300720791.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #1383, Add support for HEAD request - @steve-chavez
 - #1378, Add support for `Prefer: count=planned` and `Prefer: count=estimated` on GET /table - @steve-chavez
+- #1301, Allow self join resource embedding on PATCH - @herulume, @steve-chavez
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #1383, Add support for HEAD request - @steve-chavez
 - #1378, Add support for `Prefer: count=planned` and `Prefer: count=estimated` on GET /table - @steve-chavez
-- #1301, Allow self join resource embedding on PATCH - @herulume, @steve-chavez
+- #1301, Fix self join resource embedding on PATCH - @herulume, @steve-chavez
+- #1389, Fix many to many resource embedding on RPC/PATCH - @steve-chavez
 
 ### Changed
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -104,7 +104,7 @@ data ApiRequest = ApiRequest {
   -- | &and and &or parameters used for complex boolean logic
   , iLogic                :: [(Text, Text)]
   -- | &select parameter used to shape the response
-  , iSelect               :: Text
+  , iSelect               :: Maybe Text
   -- | &columns parameter used to shape the payload
   , iColumns              :: Maybe Text
   -- | &order parameters for each level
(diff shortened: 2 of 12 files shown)


ruslantalpa:
the reason tests didn't fail is because there are none or very few.
this part that you removed deals with embedding after an insert/update/delete is done
from the birds eye view, with your current code, if you update an item and try to embed one of it's m2m relations, it will fail.

before merging this you need 3 tests:
insert and test you can embed parent/child/m2m
update and test you can embed parent/child/m2m
delete and test you can embed parent/child/m2m

steve-chavez:
Tests for parent/child embbeding on mutations are already available.

I was thinking of doing m2m on another PR but I guess the fix can be included here.

ruslantalpa:
i've looked and i could find only a single test for insert and embed a parent (no children) and a similar test for delete.
there are no tests for update/upsert at all as far as i can see and there are no test for all mutations embedding children (one level) or tests for embedding children/parents 2 levels, and of course test for m2m. Though my guess is these type of tests will work fine with this PR but it would not hurt to have them (except for m2m relations, logic for which has been removed, so the tests will fail in this PR but not in the master). 
this above is of course from memory, not sure why there are no tests for m2m in mutations, it's possible i never tested that path (mutate/embed m2m)

ruslantalpa:
👍

steve-chavez:
Fixed #1389. As it can be seen the fix didn't involve changing `toSourceRelation` so M2M on mutations didn't work before.

Regarding tests, this is what I think:

- INSERT and embed child/m2m don't make sense since the related entities don't exist. INSERT and embed parent does and a test already exists.

- DELETE and embed child/m2m doesn't makes sense. The relationships would be broken(ON DELETE RESTRICT would give an error, CASCADE would not return back any related data? Not sure, @ruslantalpa maybe you can try adding tests for this in another PR). A test for embed parent already exists. 

- PATCH and all types of embedding should be tested. All of these are available now.

steve-chavez:
We need a git pre-commit hook. Otherwise it's easy to forget running `make lint/style`.

steve-chavez:
Rebased to summarize the refactoring commits.

PR #1392, opened 2019-09-29
Recommend doing make check pre-commit

steve-chavez:
Helps with forgetting to run the styler/linter before doing a PR.


steve-chavez:
Argh, seems `make style` is not good for working locally. Trying to do a new commit will always fail with `make check-commit`. The `git diff-index` needs some tweaking to consider the staged changes(`--cached` or something else). I recommend not enforcing it in `pre-commit` it for now.

Issue #818, opened 2017-02-27
A relational insert


Taximan:
Having a (simplified) model with a many to many relation like so

| customers  |      |
|------------|-----:|
| id         | uuid |
| first_name | text |
| last_name  | text | 

| bank    |      |
|---------|-----:|
| id      | uuid |
| name    | text |

| customer_bank_accounts |        |
|-----------------------|-------:|
| account_number        | bigint |
| customer_id           |   uuid |
| bank_id               |   uuid |

Now lets assume that in my application I have a form that lets you add a new customer.  On that form, besides creating a new customer I also want to immediately assign him to a known bank. Of course I could just: (in pseudo code)

1. ` newCustomerId =  POST /customers?select=id { "first_name": "John", "last_name": "Doe" }`
2.  `POST /customer_bank_accounts { "account_number": 123456788901234556152, "customer_id": ${newCustomerId}, "bank_id": "some_preexisting_id_of_a_bank_that_the_user_selected" ` 

However this of course is not trasactional and the problem here is that if step 2 fails I would like to rollback and discard the creation of a user. But boom tss, the user has already been created! 

Is there an api that would let me to do this like so? 

`POST http://localhost:3000/customers?select=*,customer_bank_accounts{account_number, bank_id}`
```json
 {  
   "first_name":"John",
   "last_name":"Doe",
   "customer_bank_accounts":[  
      {  
         "account_number":5851859590151,
         "bank_id":"some_preexisting_id_of_a_bank_that_the_user_selected"
      }
   ]
}
```
From the docs I figure that the only way to achieve such would be to manually write a procedure and then call it via rpc, would that be correct?

ruslantalpa:
Yes, that is correct, for now, RPCs are the only* way to do that.
It might be also possible to have the same effect using views and triggers.
for example if customers is a view and it has a column "customer_bank_accounts" which is a json, it might be possible for you to write an trigger that will allow you to do a POST to this view in the form that you specified above. I have no specific example but i remember someone implementing just this in one of the issues.

Taximan:
@ruslantalpa Eh thats a bit of a bummer. Both solution sound a bit more compilicated than it should be. There really should be a just streamlined way of doing transactions. But reading through other issues I see that there are already some ideas to implement them. Guess I will have to wait for them before adopting postgREST.

steve-chavez:
I have a proposal for this. Once #690 is solved, we could take advantage of the new query parameter(`columns`) to do this:
```sql
[...]

PR #1401, opened 2019-11-02
Err embedding when multiple relationships found 

steve-chavez:
When having one-to-many relationships like:

person

(Where `person_detail` is a view of person)

This request:
```http
GET /message?select=*,sender(*)
```

Is ambiguous. Both `person` or `person_detail` could be embedded.
Until now we have returned the first detected relationship but with this PR we would return a `300 Multiple Choices` error with this body:
```json
{
  "details": [
  {
    "cardinality": "Parent",
    "source": "test.person[id]",
    "target": "test.message[sender]"
  },
  {
    "cardinality": "Parent",
    "source": "test.person_detail[id]",
    "target": "test.message[sender]"
  }
  ],
  "hint": "Disambiguate by choosing a relationship from the `details` key",
  "message": "More than one relationship was found for message and sender"
}
```

So, for the request to succeed, the user would have to disambiguate by following the `details` info:

```http
GET /message?select=*,person_detail!sender(*)
```

Or:

```http
GET /message?select=*,person!sender(*)
```

This gets more helpful when there are more complex relationships. The aim of this PR is to formalize the disambiguation capabilities we have.

As it is now, the disambiguation syntax is hard to grok and the duck typing stuff is hard to explain and document. We get frequent questions about this on gitter chat.

This will also help for future features like #818, were a wrong embed would result in a more serious error.

This PR would imply a breaking change.


steve-chavez:
The aim is to have this syntax for disambiguation:

```http
GET /message?select=*,<table>!<fk_column>!<cardinality>(*)
```

Specifying the cardinality is needed for cases like self reference relationships where there would always be a parent and child relationship. A prototype of this can be seen [here](https://github.com/steve-chavez/postgrest/commit/844b693c72204278d82096c65593827c79a45502#diff-6d48609f00923849d6c7617d46b7c192R201-R224). The cardinality values would be: `o2m`, `m2m` and `m2o`.

I think that syntax would cover all of the ambiguous cases(fk_column could take a list later).

For that we'd have to remove duck typing. Currently that feature makes up for some confusing errors:
```http
GET /tasks?select=id,name,project(id)
```

```json
  {
   "hint":"Disambiguate by choosing a relationship from the `details` key",
   "details":[
     {"cardinality":"Child","source":"test.projects[id]","target":"test.tasks[project_id]"},
     {"cardinality":"Child","source":"test.materialized_projects[id]","target":"test.tasks[project_id]"},
     {"cardinality":"Child","source":"test.projects_view[id]","target":"test.tasks[project_id]"},
     {"cardinality":"Child","source":"test.projects_view_alt[t_id]","target":"test.tasks[project_id]"}],
    "message":"More than one relationship was found for tasks and project"
  }
```

Using `select=projects(*)` instead of `select=project(*)` in this case results in a single relationship found.

This enhancements would be done in another PR.

steve-chavez:
I'll continue with the disambiguation improvements in another PR.

For now I think the error message is good for merging.

Issue #1408, opened 2019-11-11
Set permissions on unix socket


dafvid:
### Environment

* PostgreSQL version: 11.5
* PostgREST version: 6.0.2
* Operating system: FreeBSD 12.0-RELEASE-p7

### Description of issue
I am using an unix socket to connect to the server with the option 'server-unix-socket'.
The created unix socket gets the default permissions of 755 but I need 660 since the web server needs to be able to use the socket. Please add the ability to set the socket permissions. And preferably the user/group of the socket as well.

steve-chavez:
Hey @Dansvidania, perhaps you may have some advice for the unix socket permissions?

Dansvidania:
From a first attempt at researching the issue, I have not -yet- found a way to assign a custom permission set to the socket created in the Network.Socket API. I need to investigate this further.

I apologize as I momentarily do not have access to a UNIX system and cannot verify this myself; this might be a stupid question, but is specifying a custom owner and group possible? 

unless Postgrest code runs as root, I do not think we can change the ownership of the socket to an arbitrary user:group pair (referencing [this for example](https://askubuntu.com/questions/95985/is-it-possible-to-change-ownership-of-a-file-without-root-access?rq=1) )

steve-chavez:
@Dansvidania I think you're right. Maybe we shouldn't try messing with permissions.

I had a look at this SO question: https://stackoverflow.com/questions/20171747/how-to-create-unix-domain-socket-with-a-specific-permissions-in-c.

@dafvid Can you try creating a dir with your desired permissions and pointing postgrest to create the socket file inside that dir?

dafvid:
@steve-chavez well I already do that actually. But that only solves the problem with the group membership. I need both read and write for the group.

My setup is this. I run postgrest with a postgrest user. That user and the www user are both in the www group. I run nginx with the www user. That allows the web server to read/write over the socket. I really don't want everyone to be able to read the socket.

It's true you need to run as root to change user so that wasn't perhaps such a good idea. But the postgrest user should be able to set group permissions to groups i belongs to. 

Looking at `man chgrp`: 
> The user invoking chgrp must belong to the specified group and be the owner of the file, or be the super-user.

But creating the socket in a dir with the correct group owner solves that part anyway so maybe that's the way to go.

As for socket permissions I looked at this https://github.com/grafana/grafana/issues/8358#issuecomment-303499537
and this
[...]

PR #1415, opened 2019-11-17
added config option server-unix-socket-mode

Dansvidania:
started working on solving #1408 

Currently I apply a default (changed to 660) whenever the file mode is not specified, or (I think this part is problematic) whenever the mode provided is invalid (not in octal)

I also still need to validate that the octal provided is not above 777

 @steve-chavez is this the preferred way to set the permission, or should we rather go with a string in the more readable format "rwxrwxrwx" ?

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -31,6 +31,8 @@ import Network.Wai.Handler.Warp (defaultSettings, runSettings,
 import System.Directory         (removeFile)
 import System.IO                (BufferMode (..), hSetBuffering)
 import System.IO.Error          (isDoesNotExistError)
+import System.Posix.Files       (setFileMode)
+import System.Posix.Types       (FileMode)
 
 import PostgREST.App         (postgrest)
 import PostgREST.Config      (AppConfig (..), configPoolTimeout',
@@ -163,6 +165,7 @@ main = do
       port = configPort conf
       proxy = configProxyUri conf
       maybeSocketAddr = configSocket conf
+      maybeSocketFileMode = configSocketMode conf
       pgSettings = toS (configDatabase conf) -- is the db-uri
       roleClaimKey = configRoleClaimKey conf
       appSettings =
@@ -249,8 +252,9 @@ main = do
              putStrLn $ ("Listening on port " :: Text) <> show (configPort conf)
              runSettings appSettings postgrestApplication
          Just socketAddr -> do
+             let socketFileMode = fromMaybe 432 maybeSocketFileMode
              -- run postgrest application with user defined socket
-             sock <- createAndBindSocket (unpack socketAddr)
+             sock <- createAndBindSocket (unpack socketAddr) socketFileMode
              listen sock maxListenQueue
              putStrLn $ ("Listening on unix socket " :: Text) <> show socketAddr
              runSettingsSocket appSettings sock postgrestApplication
@@ -324,11 +328,12 @@ loadDbUriFile conf = extractDbUri mDbUri
         Just filename -> strip <$> readFile (toS filename)
     setDbUri dbUri = conf {configDatabase = dbUri}
 
-createAndBindSocket :: FilePath -> IO Socket
-createAndBindSocket filePath = do
-  deleteSocketFileIfExist filePath
+createAndBindSocket :: FilePath -> FileMode -> IO Socket
+createAndBindSocket socketFilePath socketFileMode = do
+  deleteSocketFileIfExist socketFilePath
   sock <- socket AF_UNIX Stream defaultProtocol
-  bind sock $ SockAddrUnix filePath
+  bind sock $ SockAddrUnix socketFilePath
+  setFileMode socketFilePath socketFileMode
   return sock
   where
     deleteSocketFileIfExist path = removeFile path `catch` handleDoesNotExist
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -12,6 +12,7 @@ import Control.AutoUpdate       (defaultUpdateSettings, mkAutoUpdate,
 import Control.Retry            (RetryStatus, capDelay,
                                  exponentialBackoff, retrying,
                                  rsPreviousDelay)
+import Data.Either.Combinators  (whenLeft)
 import Data.IORef               (IORef, atomicWriteIORef, newIORef,
                                  readIORef)
 import Data.String              (IsString (..))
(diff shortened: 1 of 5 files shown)


steve-chavez:
> @steve-chavez is this the preferred way to set the permission, or should we rather go with a string in the more readable format "rwxrwxrwx" ?

@Dansvidania The string is more readable but the octal is more standard(what most unix devs would expect). So I think we should go with octal.

660 default is good. Making sure the value is indeed octal would be even better. Like we do [here for other configs](https://github.com/PostgREST/postgrest/blob/master/main/Main.hs#L175-L181)(we crash if they're invalid).

Dansvidania:
I have added the validation but I am not a huge fan of the code... I am looking forward to your feedback on it

main/Main.hs
```diff
@@ -324,11 +332,12 @@ loadDbUriFile conf = extractDbUri mDbUri
...
-createAndBindSocket :: FilePath -> IO Socket
-createAndBindSocket filePath = do
-  deleteSocketFileIfExist filePath
+createAndBindSocket :: FilePath -> FileMode -> IO Socket
+createAndBindSocket socketFilePath socketFileMode = do
+  deleteSocketFileIfExist socketFilePath
   sock <- socket AF_UNIX Stream defaultProtocol
-  bind sock $ SockAddrUnix filePath
+  bind sock $ SockAddrUnix socketFilePath
+  setFileMode socketFilePath socketFileMode
```
steve-chavez: How about if we just leave it with the default of `755`(also no breaking change) and let the `setFileMode` be optional:

```haskell
when (isJust socketFileMode) $
  setFileMode socketFilePath $ fromJust socketFileMode
```

Then you could call this function like: `createAndBindSocket (unpack socketAddr) (rightToMaybe socketFileMode)`.

This way you could delete the `432` parts and the validation would be similar for all configs options in the `main` func.

Dansvidania: @steve-chavez thank you for your feedback, the suggestion made a lot of sense.

Dansvidania: FYI I ended up changing it to forM_ to comply with linters and then again to mapM_ for better readability

steve-chavez: Oh, that's better. Lately I've been noticing that there are parts of the code that could be simplified in terms of `Traversable`: 
```hs
setFileMode socketFilePath `traverse_` maybeSocketFileMode
```

But in this case I think the `mapM_` is more readable, so it's good to leave it as it is.

I'll give the PR another review.

main/Main.hs
```diff
@@ -171,6 +175,9 @@ main = do
         . setServerName (toS $ "postgrest/" <> prettyVersion) $
         defaultSettings
 
+  when (isLeft socketFileMode) $
+    panic $ (fromJust . leftToMaybe) socketFileMode
```
steve-chavez: You could use [whenLeft](http://hackage.haskell.org/package/either-5.0.1.1/docs/Data-Either-Combinators.html#v:whenLeft) here(`whenLeft socketFileMode panic`). You'll have to add the `either` lib to the executable in the cabal file but that's fine.

test/SpecHelper.hs
```diff
@@ -66,6 +66,8 @@ _baseCfg =  -- Connection Settings
   AppConfig mempty "postgrest_test_anonymous" Nothing "test" "localhost" 3000
             -- No user configured Unix Socket
             Nothing
+            -- No user configured Unix Socket file mode (defaults to 660)
+            (Right 432)
```
steve-chavez: `493` here by default. Also the comment should be `defaults to 755`·

src/PostgREST/Config.hs
```diff
@@ -248,6 +266,9 @@ readOptions = do
           |## unix socket location
           |## if specified it takes precedence over server-port
           |# server-unix-socket = "/tmp/pgrst.sock"
+          |## unix socket file mode
+          |## if server-unix-socket is used, but an invalid mode or none is provided, 755 is applied
+          |# server-unix-socket-mode = "777"
```
steve-chavez: Example value could be `server-unix-socket-mode = "755"` to better signal the default.

Also, the comment about `invalid mode` should be removed since now we crash when the value is invalid.

steve-chavez:
See my latest comments, once we clear that I think it's good to merge.

Also, could you help me with updating the docs for this new config option? 

The change would have to be done in https://github.com/PostgREST/postgrest-docs/blob/master/configuration.rst.

Dansvidania:
@steve-chavez thanks for checking again, I should have corrected all the mistakes and changed to `whenLeft` . Is the version condition with a lower boundary only ok for `either` ? i had no idea what would be a reasonable upper boundary

postgrest.cabal
```diff
@@ -97,6 +97,7 @@ executable postgrest
                     , base64-bytestring >= 1 && < 1.1
                     , bytestring        >= 0.10.8 && < 0.11
                     , directory         >= 1.2.6 && < 1.4
+                    , either            >= 5.0.1.1
```
steve-chavez: The library uses `either >= 4.4.1 && < 5.1`. You could copy that.

main/Main.hs
```diff
@@ -171,13 +176,16 @@ main = do
         . setServerName (toS $ "postgrest/" <> prettyVersion) $
         defaultSettings
 
+  whenLeft socketFileMode $
```
steve-chavez:
```suggestion
  whenLeft socketFileMode panic
```

main/Main.hs
```diff
@@ -171,13 +176,16 @@ main = do
...
+    panic $ (fromJust . leftToMaybe) socketFileMode
+
   -- Checks that the provided proxy uri is formated correctly
   when (isMalformedProxyUri $ toS <$> proxy) $
     panic
       "Malformed proxy uri, a correct example: https://example.com:8443/basePath"
 
   -- Checks that the provided jspath is valid
-  when (isLeft roleClaimKey) $
+  whenLeft roleClaimKey $
```
steve-chavez: I think you changed an unrelated line here.

main/Main.hs
steve-chavez: That would allow you to drop the next line: `panic $ (fromJust . leftToMaybe) socketFileMode`.

main/Main.hs
Dansvidania: I refactored based on the `whenLeft` suggestion you made, to keep it consistent

main/Main.hs
Dansvidania: I was not aware of this usage of panic, thanks.

steve-chavez:
LGTM! :+1: 

Great work done here @Dansvidania. Merging...

dwagin:
The default 755 doesn't make sense, should be 777 or 770.

PR #1417, opened 2019-11-25
Improvement singular/plural logic

dwagin:
Hello

I get Incorrect statuses by using "Accept: application/vnd.pgrst.object+json" with and without "Prefer: return=representation".

for multiple entities on
POST expected 406, receive 201
PATCH expected 406, receive 204
DELETE expected 406, receive 204

for zero entities on
PATCH expected 404, receive 406 
DELETE expected 404, receive 204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1417, Improvement singular/plural logic - @dwagin
 - #1415, Add support for user defined socket permission via `server-unix-socket-mode` config option - @Dansvidania
 - #1383, Add support for HEAD request - @steve-chavez
 - #1378, Add support for `Prefer: count=planned` and `Prefer: count=estimated` on GET /table - @steve-chavez
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- #1417, Improvement singular/plural logic - @dwagin
+- #1417, `Accept: application/vnd.pgrst.object+json` behavior is now enforced for POST/PATCH/DELETE regardless of `Prefer: return=representation/minimal` - @dwagin
 - #1415, Add support for user defined socket permission via `server-unix-socket-mode` config option - @Dansvidania
 - #1383, Add support for HEAD request - @steve-chavez
 - #1378, Add support for `Prefer: count=planned` and `Prefer: count=estimated` on GET /table - @steve-chavez
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -251,7 +251,7 @@ app dbStructure proc cols conf apiRequest =
                     (iPreferRepresentation apiRequest) []
               row <- H.statement mempty stm
               let (_, queryTotal, _, body) = row
-                  r = contentRangeH 1 0 $
+                  contentRangeHeader = contentRangeH 1 0 $
                         if shouldCount then Just queryTotal else Nothing
               if contentType == CTSingularJSON
                  && queryTotal /= 1
@@ -260,7 +260,7 @@ app dbStructure proc cols conf apiRequest =
                   return . errorResponseFor . singularityError $ queryTotal
                 else
                   return $ if iPreferRepresentation apiRequest == Full
-                    then responseLBS status200 [toHeader contentType, r] (toS body)
+                    then responseLBS status200 [toHeader contentType, contentRangeHeader] (toS body)
                     else responseLBS status204 [r] ""
 
         (ActionInfo, TargetIdent (QualifiedIdentifier tSchema tTable), Nothing) ->
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -175,7 +175,6 @@ app dbStructure proc cols conf apiRequest =
                     ]
               if contentType == CTSingularJSON
                  && queryTotal /= 1
-                 && iPreferRepresentation apiRequest == Full
                 then do
                   HT.condemn
                   return . errorResponseFor . singularityError $ queryTotal
@@ -193,28 +192,24 @@ app dbStructure proc cols conf apiRequest =
                     (iPreferRepresentation apiRequest) []
               row <- H.statement (toS $ pjRaw pJson) stm
               let (_, queryTotal, _, body) = row
-
-                  updateIsNoOp       = S.null cols
-                  contentRangeHeader = contentRangeH 0 (queryTotal - 1) $
-                                          if shouldCount then Just queryTotal else Nothing
-                  minimalHeaders     = [contentRangeHeader]
-                  fullHeaders        = toHeader contentType : minimalHeaders
-
+                  updateIsNoOp = S.null cols
+                  r = contentRangeH 0 (queryTotal - 1) $
+                        if shouldCount then Just queryTotal else Nothing
+                  headers | iPreferRepresentation apiRequest == Full = [toHeader contentType, r]
+                          | otherwise                                = [r]
                   status | queryTotal == 0 && not updateIsNoOp      = status404
                          | iPreferRepresentation apiRequest == Full = status200
                          | otherwise                                = status204
-
-              case (contentType, iPreferRepresentation apiRequest) of
-                (CTSingularJSON, Full)
-                      | queryTotal == 1 -> return $ responseLBS status fullHeaders (toS body)
-                      | otherwise       -> HT.condemn >> (return . errorResponseFor . singularityError) queryTotal
-
-                (_, Full) ->
-                  return $ responseLBS status fullHeaders (toS body)
-
-                (_, _) ->
-                  return $ responseLBS status minimalHeaders mempty
-
+              if contentType == CTSingularJSON
+                 && queryTotal > 1
+                then do
+                  HT.condemn
+                  return . errorResponseFor . singularityError $ queryTotal
+              else
+                return . responseLBS status headers $
+                  if iPreferRepresentation apiRequest == Full
+                     && queryTotal > 0
+                    then toS body else ""
 
         (ActionSingleUpsert, TargetIdent (QualifiedIdentifier tSchema tName), Just ProcessedJSON{pjRaw, pjType, pjKeys}) ->
           case mutateSqlParts tSchema tName of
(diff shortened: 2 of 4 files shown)


steve-chavez:
Hmm.. Several tests are failing though that's mostly due to renaming the error message.

Could you add tests so I can better understand your proposal?

steve-chavez:
@dwagin Hey there.

Hmm.. Several tests are failing though that's mostly due to renaming the error message.

Could you add tests so I can better understand your proposal?

dwagin:
I must have misunderstood. Some changes need to be made.

dwagin:
https://github.com/PostgREST/postgrest/pull/763#issuecomment-269425834

> We're treating a blank body as a valid singular response. It's kind of convenient for the client, but maybe points to the name of our content type as misleading. It's not so much object as it is non-array.

dwagin:
IMHO now 'Accept: application/vnd.pgrst.object + json' does not work correctly, namely it returns an error only if the header 'Prefer: return = representation' is present. When interacting with not equal to one object, an error should be returned.
I prepared a Pull Request that solves this problem, but maybe I'm missing something essential.

steve-chavez:
@dwagin I think I understand your point a bit better now. Basically, right now we issue an error when the following happens:
```bash
http POST localhost:3000/projects "Accept:application/vnd.pgrst.object+json" "Prefer: return=representation" <<JSON
[{"id": 113, "name": "x"}, {"id": 114, "name": "y"}]
JSON
HTTP/1.1 406 Not Acceptable
```
Without `Prefer: return=representation` the same request succeeds:
```bash
http POST localhost:3000/projects "Accept:application/vnd.pgrst.object+json" <<JSON
[{"id": 113, "name": "x"}, {"id": 114, "name": "y"}]
JSON
HTTP/1.1 201 Created
```
Perhaps we should enforce that if content negotiation(`Accept`) is desired, then `return=representation` is implicit for POST? What do you think?

dwagin:
In my opinion, there is incorrect behavior when lack of some other flags of request which tagged as singular, made differing result. It should have 406 status and new multiple entities should not been added to database to maintain uniformity of REST API behavior.

steve-chavez:
@dwagin I totally agree we have incorrect behavior when singular is requested and we do not return 406 when multiple entities are inserted.

test/Feature/SingularSpec.hs
```diff
@@ -56,7 +56,7 @@ spec =
         _ <- post "/addresses" [json| { id: 98, address: "xxx" } |]
         _ <- post "/addresses" [json| { id: 99, address: "yyy" } |]
         p <- request methodPatch "/addresses?id=gt.0"
-                [("Prefer", "return=representation"), singular]
+                [singular]
```
steve-chavez: It'd be better to not modify previous tests that have both `Prefer:return=representation` and `singular`. Instead you could add new tests for POST/PATCH/DELETE that only have the `singular` header and fail with 406.

test/Feature/SingularSpec.hs
```diff
@@ -100,25 +100,13 @@ spec =
       it "raises an error when attempting to create multiple entities" $ do
         p <- request methodPost
           "/addresses"
-          [("Prefer", "return=representation"), singular]
+          [singular]
           [json| [ { id: 200, address: "xxx" }, { id: 201, address: "yyy" } ] |]
         liftIO $ simpleStatus p `shouldBe` notAcceptable406
 
-        -- the rows should not exist, either
-        get "/addresses?id=eq.200" `shouldRespondWith` "[]"
```
steve-chavez: Was this line removed on purpose? Seems it should still be correct.

test/Feature/SingularSpec.hs
```diff
@@ -100,25 +100,13 @@ spec =
...
-
-      it "return=minimal allows request to create multiple elements" $
-        request methodPost "/addresses"
-          [("Prefer", "return=minimal"), singular]
```
steve-chavez: Instead of deleting this test, perhaps we can rephrase it as `singular behavior holds priority over return=minimal` and confirm it returns 406.

steve-chavez: Hmm.. I'm hesitating about this part now, I think it merits more discussion. See my above comment.

dwagin: I think that expectation to have success code on creation multiple objects when you obviously set singular entity header is oddly and in real life no projects do this.

dwagin:
I think that expectation to have success code on creation multiple objects when you obviously set singular entity header is oddly and in real life no projects do this.

steve-chavez:
So, if client specifies `Prefer: return=minimal` and `Accept: application/vnd.pgrst.object+json` in the same POST request we'd assume it's an error and only apply the latter.

it does sound reasonable but I'm not sure if this is correct from an http standpoint(@jackfirth perhaps you could help us to clear this doubt?).

@dwagin Also, with your current patch, the following happens:

```bash
http POST localhost:3000/projects "Accept:application/vnd.pgrst.object+json" <<JSON
[{"id": 1115, "name": "x"}]
JSON
HTTP/1.1 201 Created
## No body
```
Returning an empty body on an Accept request wouldn't be correct. This brings my previous question, should we assume `Prefer: return=representation` when `Accept` is specified?

fortyanov:
We setup unix socket on postrest
server-unix-socket = "/tmp/.s.PostgREST"
server-unix-socket-mode = "777"
Problem still happens.

When we setup NGINX keepalive_timeout value up
```
{
  server unix:/tmp/.s.PostgREST;
  keepalive 64;
  keepalive_timeout 120s;
}
```
to make it bigger then keep-alive in postgrest the problem not happens. Default  keepalive timeout is 60 second in nginx and postgrest too (I think) . Seems like there is race condition on closing session between postgrest and nginx.

fortyanov:
Im bad in haskel no there is could be information about library that used in postgrest [https://github.com/yesodweb/wai/issues/734#issuecomment-555441539](url)

fortyanov:
We setup unix socket on postrest
```
server-unix-socket = "/tmp/.s.PostgREST"
server-unix-socket-mode = "777"
```
Problem still happens.

When we setup NGINX keepalive_timeout value up
```
{
  server unix:/tmp/.s.PostgREST;
  keepalive 64;
  keepalive_timeout 120s;
}
```
to make it bigger then keep-alive in postgrest the problem not happens. Default  keepalive timeout is 60 second in nginx and postgrest too (I think) . Seems like there is race condition on closing session between postgrest and nginx.

fortyanov:
Im bad in haskel no there is could be information about library that used in postgrest
[https://github.com/yesodweb/wai/issues/734#issuecomment-555441539](url)

dwagin:
Why you think that empty body in response is incorrect when we don't use header `Prefer: return=representation` or `Prefer: return=minimal`.

When we use `Accept: application/vnd.pgrst.object+json` we expect to work with single object in first of all. If there is zero or many objects we expect to get error status. In postgres there is no other way to specify this.

steve-chavez:
@dwagin Sorry, I phrased that wrong. What it doesn't look correct is to enforce the `Accept: application/vnd.pgrst.object+json` condition and don't return a body. `Accept` acts upon the response body and here we enforce the condition when no body is present.

steve-chavez:
I've gotten feedback from the "apis you won't hate" slack channel. I think they're right to point that:

> Accept is used for the client to tell the server the set of content-types that it is capable of understanding.
It does not mean that the server **must** return a response. It just means that if the server does return a response it should be in one of these formats.

So to answer my previous doubt: It's fine to enforce the `Accept: singular` behavior despite not returning any body.

dwagin:
> So to answer my previous doubt: It's fine to enforce the `Accept: singular` behavior despite not returning any body.

Good news!

What about the pull request?

steve-chavez:
@dwagin Sorry for the delay. I'll give it a review now.

src/PostgREST/App.hs
```diff
@@ -193,28 +192,23 @@ app dbStructure proc cols conf apiRequest =
...
-                (CTSingularJSON, Full)
-                      | queryTotal == 1 -> return $ responseLBS status fullHeaders (toS body)
-                      | otherwise       -> HT.condemn >> (return . errorResponseFor . singularityError) queryTotal
-
-                (_, Full) ->
-                  return $ responseLBS status fullHeaders (toS body)
-
-                (_, _) ->
-                  return $ responseLBS status minimalHeaders mempty
-
```
steve-chavez: This previous case expression was much easier to read for me. Could you keep the same structure and apply the new logic? I think it'd be a matter of wildcard matching the `iPreferRepresentation`:

```hs
case (contentType, iPreferRepresentation apiRequest) of
  (CTSingularJSON, _) -> .... 
```

src/PostgREST/App.hs
```diff
@@ -193,28 +192,23 @@ app dbStructure proc cols conf apiRequest =
                     (iPreferRepresentation apiRequest) []
               row <- H.statement (toS $ pjRaw pJson) stm
               let (_, queryTotal, _, body) = row
-
-                  updateIsNoOp       = S.null cols
-                  contentRangeHeader = contentRangeH 0 (queryTotal - 1) $
-                                          if shouldCount then Just queryTotal else Nothing
-                  minimalHeaders     = [contentRangeHeader]
-                  fullHeaders        = toHeader contentType : minimalHeaders
-
```
steve-chavez: Ditto. Previous variables were more clear.

test/Feature/SingularSpec.hs
```diff
@@ -79,44 +101,68 @@ spec =
         p <- request methodPost
           "/addresses"
           [("Prefer", "return=representation"), singular]
-          [json| [ { id: 100, address: "xxx" } ] |]
-        liftIO $ simpleBody p `shouldBe` [str|{"id":100,"address":"xxx"}|]
+          [json| [ { id: 102, address: "xxx" } ] |]
+        liftIO $ simpleBody p `shouldBe` [str|{"id":102,"address":"xxx"}|]
```
steve-chavez: Hmm.. why are you changing the `id` value here? Makes me think that one of the previous test cases succeeded in patching when it shouldn't have.

test/Feature/SingularSpec.hs
```diff
@@ -79,44 +101,68 @@ spec =
...
+        p <- request methodPost
+          "/addresses"
+          [("Prefer", "return=representation"), singular]
+          [json| [ { id: 202, address: "xxx" }, { id: 203, address: "yyy" } ] |]
+        liftIO $ simpleStatus p `shouldBe` notAcceptable406
+
+        -- the rows should not exist, either
+        get "/addresses?id=eq.202" `shouldRespondWith` "[]"
+
+      it "singular behavior holds priority over return=minimal" $ do
```
steve-chavez: Now that I cleared my doubt, I think it's not about priority but more about consistent behavior. `singular` will enforce behavior regardless of `return=minimal` or `return=rep`.

So maybe this test case can be rephrased as `it "raises an error regardless of return=minimal"`.

steve-chavez:
@dwagin The new `singular` logic looks good, check my review(mostly about readibility). 

Also, don't forget to add an entry to the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md).

test/Feature/SingularSpec.hs
dwagin: Separate `id` from other tests

src/PostgREST/App.hs
dwagin: This variant of code is more corresponding with other parts.

steve-chavez: @dwagin The other parts could be refactored later. Variables like `r` are not really meaningful and it's important that we keep improving readability.

test/Feature/SingularSpec.hs
steve-chavez: Oh, I see. That's better.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- #1417, Improvement singular/plural logic - @dwagin
```
steve-chavez: A more descriptive entry would be better for the users. How about:
```suggestion
- #1417, `Accept: application/vnd.pgrst.object+json` behavior is now enforced for POST/PATCH/DELETE regardless of `Prefer: return=representation/minimal` - @dwagin
```

src/PostgREST/App.hs
dwagin: Naming changed.

dwagin: IMHO switch/case in this case less readable and requires code duplication.

steve-chavez: It reads better now. Thanks!

steve-chavez:
@dwagin Thanks a lot for your work here. Our singular behavior is more consistent now.

PR #1423, opened 2019-12-04
Change default server-unix-socket-mode 755 -> 777

dwagin:
The default 755 doesn't make sense, should be 777 or 770. IMHO default 777 is better, because PostgREST does authorization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- #1423, Changed default Unix Socket file mode from 755 to 660 - @dwagin
 - #1369, Change `raw-media-types` to accept a string of comma separated MIME types - @Dansvidania
 - #1368, Fix long column descriptions being truncated at 63 characters in PostgreSQL 12 - @amedeedaboville
 - #1348, Go back to converting plus "+" to space " " in querystrings by default - @steve-chavez
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,12 +22,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - #1385, bulk RPC call now should be done by specifying a `Prefer: params=multiple-objects` header - @steve-chavez
 - #1401, resource embedding now outputs an error when multiple relationships between two tables are found - @steve-chavez
+- #1423, default Unix Socket file mode from 755 to 660 - @dwagin
 
 ## [6.0.2] - 2019-08-22
 
 ### Fixed
 
-- #1423, Changed default Unix Socket file mode from 755 to 660 - @dwagin
 - #1369, Change `raw-media-types` to accept a string of comma separated MIME types - @Dansvidania
 - #1368, Fix long column descriptions being truncated at 63 characters in PostgreSQL 12 - @amedeedaboville
 - #1348, Go back to converting plus "+" to space " " in querystrings by default - @steve-chavez
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -267,8 +267,8 @@ readOptions = do
           |## if specified it takes precedence over server-port
           |# server-unix-socket = "/tmp/pgrst.sock"
           |## unix socket file mode
-          |## when none is provided, 777 is applied by default
-          |# server-unix-socket-mode = "777"
+          |## when none is provided, 660 is applied by default
+          |# server-unix-socket-mode = "660"
           |
           |## base url for swagger output
           |# server-proxy-uri = ""
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -177,7 +177,7 @@ readOptions = do
     parseSocketFileMode :: C.Key -> C.Parser C.Config (Either Text FileMode)
     parseSocketFileMode k =
       C.optional k C.string >>= \case
-        Nothing -> pure $ Right 493 -- return default 755 mode if no value was provided
+        Nothing -> pure $ Right 511 -- return default 777 mode if no value was provided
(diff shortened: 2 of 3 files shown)


steve-chavez:
Could you elaborate more on why `777` is better than `755`?

dwagin:
> Could you elaborate more on why `777` is better than `755`?

755 does not make sense, it is the same as 700 in our case. 700 means that only the postgres user has access, this is inconvenient to use.

steve-chavez:
@Dansvidania What do you think?

@dwagin If we change the default, the docs should be updated https://github.com/PostgREST/postgrest-docs/blob/master/configuration.rst (`755` to `777`). Also, it'd need an entry on the CHANGELOG under the "Changed" section.

Dansvidania:
@steve-chavez i think the point raised by @dwagin makes sense, but I would vote against making 777 the default for the following reasons:

My intent for the 755 default was to 
A- not change the behaviour of postgrest instances that were started by people before we provided the ```server-unix-socket-mode``` option
B- provide a default that would make it possible to test and play with the application with minimal setup, but that (because of the limitations pointed out by @dwagin) would encourage users to setup their own filemode overriding the default.

dwagin:
> A- as @steve-chavez pointed out, 755 does not change the behaviour of Postgrest instances that were started by people before we provided the `server-unix-socket-mode` option

The default setting 755 same as 700. For this reason it was necessary to add a setting `server-unix-socket-mode`.
 
> B- we currently provide a default that makes it possible to test and play with the application with minimal setup, but that (because of the limitations pointed out by @dwagin) encourages users to setup their own filemode overriding the default once they move out of the "playing around"/discovery phase.

Everything is exactly the opposite. The default setting (755) does not allow you to try this mode. The web server runs under its own user and does not have access to postgrest socket. After "playing around"/discovery phase a reasonable setting is 770.

> Moving the default to 777 is against both lines of thought.

I think this is wrong.

> If we want to disregard both A and B, I would strongly suggest to change to a more restrictrive filemode. Given the fact that the users are allowed to customise the filemode, there is no reason not to follow the _principle of least privilege_ by default.

This is an imaginary defense that makes the program not friendly.

I have not seen any programs that have default 700 mode.

Dansvidania:
@dwagi 777 by default would allow users to ignore or forget this option, leaving them with an unnecessarily permissive file mode, which is IMO a code smell. 

My preference would have been to make ```server-unix-socket-mode``` mandatory whenever ```server-unix-socket``` is used, but that is a potentially breaking change, so I decided against it.

As a compromise, I would rather have a default that allows you to start the application ignoring the ```server-unix-socket-mode``` option but that forces users to think about what mode they need in production and apply their own, instead of a very permissive default possibly creating security holes.

dwagin:
> @dwagin 777 by default would allow users to ignore or forget this option, leaving them with an unnecessarily permissive file mode, which is IMO a code smell.
> 
> My preference would have been to make `server-unix-socket-mode` mandatory whenever `server-unix-socket` is used, but that is a potentially breaking change, so I decided against it.
> 
> As a compromise, I would rather have an inconvenient default that allows starting the application ignoring the `server-unix-socket-mode` option but that forces users to think about what mode they need in production and apply their own, instead of a very permissive default possibly creating security holes.

On such logic listen on 127.0.0.1, too, not safely.

steve-chavez:
Having the socket be world-readable/writable by default is definitely not a good practice.
If this is needed, the `server-unix-socket-mode=777` option is available(just one line of config, I don't see a major inconvenience there).

If we were to change the default, the `660` proposed in #1415 would be more sensible.

dwagin:
Сhoosing 660 by default is much better than 700, it's a working setting.

dwagin:
I'll fix the pull request in the near future.

Dansvidania:
> If we were to change the default, the `660` proposed in #1415 would be more sensible(I'd also have preferred to not introduce a breaking change but since we're here, we might just as well pick a better default).

if we are to make breaking changes, how about making ```server-unix-socket-mode``` mandatory when ```server-unix-socket``` is in use?

steve-chavez:
@Dansvidania It seems a bit too strict for me. I haven't seen another web server enforce being explicit about the mode. For example, nginx creates the socket with `666` as the default and doesn't even let you modify it through its config(AFAICT).

> instead of changing the default "under their feet"

Since we follow [Semver](https://semver.org/), we'll increase the MAJOR version next release and also announce the new default mode. Proper care is always needed when migrating to a new MAJOR version and that's why we encourage seeing the CHANGELOG(we even have two, one on the docs and one on the repo).

Instead I think we should try to be as clear as possible with breaking changes on the CHANGELOG.
What do you think?

dwagin:
Why do you all say that the default change of rights 755 -> 660 or 755 -> 777 is "breaking changes"?

dwagin:
`server-unix-socket` and` server-unix-socket-mode` are not yet released.

src/PostgREST/Config.hs
```diff
@@ -267,8 +267,8 @@ readOptions = do
           |## if specified it takes precedence over server-port
           |# server-unix-socket = "/tmp/pgrst.sock"
           |## unix socket file mode
-          |## when none is provided, 755 is applied by default
-          |# server-unix-socket-mode = "755"
+          |## when none is provided, 777 is applied by default
+          |# server-unix-socket-mode = "777"
```
steve-chavez: `660` change is missing here.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- #1423, Changed default Unix Socket file mode from 755 to 660 - @dwagin
```
steve-chavez: The entry should go under the `Changed` section.

Issue #1230, opened 2019-01-17
Resource embedding with view works with some tables but not others


matt-snider:
### Environment

* PostgreSQL version: docker postgres:11 (4dcf1ba7e5e2)
* PostgREST version: docker postgrest/postgrest:v5.2.0 (643538cf996d)
* Operating system: Arch Linux 4.19.11

### Description of issue

There appears to be an issue with resource embedding when using foreign keys with views. @steve-chavez was [able to reproduce](https://github.com/PostgREST/postgrest/issues/1179#issuecomment-454878513) the issue.

Here is the schema and minimal dataset:
```sql
-- Schema
CREATE TABLE projects (
  project_id  serial  PRIMARY KEY,
  name varchar(150)
);

CREATE TABLE sites (
  site_id     serial  PRIMARY KEY,
  name varchar(150)
);

CREATE TABLE jobs (
  job_id      uuid    PRIMARY KEY,
  site_id     int     NOT NULL REFERENCES sites,
  project_id  int     NOT NULL REFERENCES projects
);

CREATE VIEW stats AS
  SELECT
    site_id,
    project_id,
    count(job_id) AS number_of_jobs
  FROM sites
  JOIN jobs USING (site_id)
  JOIN projects USING (project_id)
  GROUP BY (site_id, project_id);

-- Data 
INSERT INTO projects (project_id, name) 
VALUES (1, 'project x'), 
       (2, 'project y');

INSERT INTO sites (site_id, name) 
VALUES (1, 'site x'), 
       (2, 'site y');

INSERT INTO jobs (job_id, site_id, project_id) 
VALUES ('bc5d5362-b881-438f-b9f5-7417e08704ed', 1, 1),
       ('3bd52697-033b-4edd-8a28-46a9c04b7c1e', 1, 2);
```

When attempting to embed a `site` object, I seem to receive list of jobs instead:
```
GET /stats?select=*,sites:site_id(*)
[
    {
        "site_id": 1,
        "project_id": 1,
        "number_of_jobs": 1,
        "sites": [
            {
                "job_id": "bc5d5362-b881-438f-b9f5-7417e08704ed",
                "site_id": 1,
                "project_id": 1
            },
            {
                "job_id": "3bd52697-033b-4edd-8a28-46a9c04b7c1e",
                "site_id": 1,
                "project_id": 2
            }
        ]
    },
    ...
]
```

[...]

PR #1430, opened 2019-12-18
Allow specifying the constraint name to disambiguate an embedding

steve-chavez:
This makes the previous duck typing rules(undocumented) unnecessary, allows composite FK disambiguation and supports a more flexible naming scheme. 

The previous duck typing can give unexpected matches: see https://github.com/PostgREST/postgrest/issues/1406#issue-520233773 (how it allows embedding `smallmolec`) and this [previous test](https://github.com/PostgREST/postgrest/blob/2e6c78d72349cc5fd6214ee08c4f5f5daf19fac2/test/Feature/EmbedDisambiguationSpec.hs#L41-L131)(which shows how long the error message can get). These make the disambiguation error less helpful.

A migration path(without changing the clients/frontends) for deployments that used the duck typing feature would be renaming the foreign key:
```sql
alter table projects rename constraint projects_client_id_fkey to client;
```
Then this request can be done as before:
```http
GET /projects?select=*,client(*)
```
This is also more flexible since you can rename the fk as `project_client` or any other format that makes the embedding more clear(also FK renaming is easy compared to renaming columns since they're not used in queries). FK names are also namespaced by table, which makes this mechanism convenient since two different tables can have an FK named `client`.

This would leave the disambiguation options like:
```abnf
; GET /origin?select=target(*)
request = "/" origin "?select=" target "(*)"
; GET /origin?select=target!hint(*)
request-with-hint = "/" origin "?select=" target "!" hint "(*)"
origin = table / view
target = table / view / constraint / column-from-origin
hint = constraint / column-from-origin / column-from-target / junction
```

(no breaking changes to column name disambiguation, self reference embeds are also handled in a special way as before)

I'll start working on docs for disambiguation to see how clear it can be explained.

### DIFF
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -4,7 +4,7 @@ import Network.Wai      (Application)
 import Network.Wai.Test (SResponse (simpleHeaders))
 
 import Network.HTTP.Types
-import Test.Hspec
+import Test.Hspec          hiding (pendingWith)
 import Test.Hspec.Wai
 import Test.Hspec.Wai.JSON
 
@@ -265,18 +265,8 @@ spec actualPgVersion = do
         [json|[{"id":1,"name":"Windows 7","clients":{"id":1,"name":"Microsoft"},"tasks":[{"id":1,"name":"Design w7"},{"id":2,"name":"Code w7"}]}]|]
         { matchHeaders = [matchContentTypeJson] }
 
-    it "requesting parent without specifying primary key" $
-      get "/projects?select=name,client(name)" `shouldRespondWith`
-        [json|[
-          {"name":"Windows 7","client":{"name": "Microsoft"}},
-          {"name":"Windows 10","client":{"name": "Microsoft"}},
-          {"name":"IOS","client":{"name": "Apple"}},
-          {"name":"OSX","client":{"name": "Apple"}},
-          {"name":"Orphan","client":null}
-        ]|]
-        { matchHeaders = [matchContentTypeJson] }
-
-    it "requesting parent and renaming primary key" $
+    it "requesting parent and renaming primary key" $ do
+      pendingWith "duck typing removed: see what to do with single fk column embed"
       get "/projects?select=name,client(clientId:id,name)" `shouldRespondWith`
         [json|[
           {"name":"Windows 7","client":{"name": "Microsoft", "clientId": 1}},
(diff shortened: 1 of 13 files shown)


test/Feature/QuerySpec.hs
```diff
@@ -265,19 +265,8 @@ spec actualPgVersion = do
         [json|[{"id":1,"name":"Windows 7","clients":{"id":1,"name":"Microsoft"},"tasks":[{"id":1,"name":"Design w7"},{"id":2,"name":"Code w7"}]}]|]
         { matchHeaders = [matchContentTypeJson] }
 
-    it "requesting parent without specifying primary key" $
```
steve-chavez: This test has been moved to [EmbedDisambiguationSpec](https://github.com/PostgREST/postgrest/pull/1430/files#diff-6db2a928de7cd07d05747e73acfef7e8R144-R148)(which github hides from the default diff). It's shown to work with the FK rename.

test/Feature/QuerySpec.hs
```diff
@@ -295,13 +284,8 @@ spec actualPgVersion = do
         [json|[{"id":1,"commenter_id":1,"user_id":2,"task_id":6,"content":"Needs to be delivered ASAP","users_tasks":{"taskId": 6}}]|]
         { matchHeaders = [matchContentTypeJson] }
 
-    it "embed data with two fk pointing to the same table" $
```
steve-chavez: Same here. Moved to [EmbedDisambiguationSpec](https://github.com/steve-chavez/postgrest/blob/5e7d7ea8efa8ea32641c82ed5b6ca11472a27a78/test/Feature/EmbedDisambiguationSpec.hs#L243-L246). The QuerySpec tests were getting big and unrelated.

steve-chavez:
With this change I can start wrapping my head around solving #1230. The disambiguation error now should be precise enough to see where the embedding is going wrong.

test/Feature/QuerySpec.hs
```diff
@@ -265,19 +265,8 @@ spec actualPgVersion = do
         [json|[{"id":1,"name":"Windows 7","clients":{"id":1,"name":"Microsoft"},"tasks":[{"id":1,"name":"Design w7"},{"id":2,"name":"Code w7"}]}]|]
         { matchHeaders = [matchContentTypeJson] }
 
-    it "requesting parent without specifying primary key" $
-      get "/projects?select=name,client(name)" `shouldRespondWith`
-        [json|[
```
steve-chavez: This test has been moved to [EmbedDisambiguationSpec](https://github.com/steve-chavez/postgrest/blob/5e7d7ea8efa8ea32641c82ed5b6ca11472a27a78/test/Feature/EmbedDisambiguationSpec.hs#L144-L148)(which github hides from the default diff). It's shown to work with the FK rename.

test/Feature/QuerySpec.hs
```diff
@@ -295,13 +284,8 @@ spec actualPgVersion = do
         [json|[{"id":1,"commenter_id":1,"user_id":2,"task_id":6,"content":"Needs to be delivered ASAP","users_tasks":{"taskId": 6}}]|]
         { matchHeaders = [matchContentTypeJson] }
 
-    it "embed data with two fk pointing to the same table" $
-      get "/orders?id=eq.1&select=id, name, billing_address_id(id), shipping_address_id(id)" `shouldRespondWith`
-        [str|[{"id":1,"name":"order 1","billing_address_id":{"id":1},"shipping_address_id":{"id":2}}]|]
```
steve-chavez: Same here. Moved to [EmbedDisambiguationSpec](https://github.com/steve-chavez/postgrest/blob/5e7d7ea8efa8ea32641c82ed5b6ca11472a27a78/test/Feature/EmbedDisambiguationSpec.hs#L243-L246). The QuerySpec tests were getting big and unrelated.

steve-chavez:
While documenting this functionality, I've realized that the following notation:
```http
GET /projects?select=*,client_id(*)
```
Has always been bad(with duck typing, column name or fk name) for client stability. That's because adding a VIEW of `clients` will always result in an ambiguous request. Once the VIEW is added the client would need patching.

The other longer notation with the hint:
```http
GET /projects?select=*,clients_view!client_id(*)
```
Would be stable since the view/table is specified.

I'm thinking whether documenting only the latter notation or documenting both and recommending the latter.

steve-chavez:
I think stability should be more valued than having a "shorthand" for specifying the embedding.
 
The reasons why we support the **target** disambiguation are: backwards compatibility and to allow self reference embedding(which could be solved with the cardinality hint I proposed on #1419).

steve-chavez:
I think I now realize what the best way to solve this would be: deprecate the **target** disambiguation notation without causing breaking changes. 

I can remove the ability to specify the FK on the **target** and we can reintroduce the duck typing. We can avoid documenting this **target** notation and only document/encourage the **hint** disambiguation since it's more stable.

steve-chavez:
> I think I now realize what the best way to solve this would be: deprecate the target disambiguation notation without causing breaking changes.

This idea was too strict and reduced practicality. I've left the PR as it is and improved the error message a bit.

This are the docs resulting from this PR: https://postgrest.org/en/preview/api.html#embedding-disambiguation.

I've tried to make them short and not put all of the possible cases, instead only mention the disambiguation capabilities.

Issue #1327, opened 2019-06-17
Upsert with "Prefer: resolution=merge-duplicates" on a specified column


kasugaicrow:
According to the PostgREST's documentation, if I want to make an upsert on a table I need to specify the "Prefer: resolution=merge-duplicates" string on the header of the request. The documentation also says 

> UPSERT operates based on the primary key columns

But what if I want to made an upsert on a table with a specified column rather than the primary key. It is possible with the latest version of PostgREST?

For instance, if I have a table like the following, the upsert will work with the primary key **id_user** but in this case the field type is UUID so the values ​​will always be different and the upsert will "fail". I would like to upsert using unique_code column that it's unique anyways.
```sql
CREATE TABLE IF NOT EXISTS user (
	id_user UUID PRIMARY KEY,
        unique_code TEXT,       
	first_name TEXT
	last_name TEXT,
	age TEXT,
	address TEXT,
);
```

steve-chavez:
I see your point. Right now it's not possible to specify a column different than the PK.

The UPSERT feature is possible thanks to [PostgreSQL ON CONFLICT clause](https://www.postgresql.org/docs/11/sql-insert.html#SQL-ON-CONFLICT).
Maybe we could offer an additional query param that makes explicit use of this clause.

That way you could specify your UNIQUE column(must have a constraint). It could be like:

```http
POST /user?on_conflict=unique_code
```

What do you think?

ChristianPredebon:
Ok great, that's exactly what I was looking for. A friend of mine could implement it, what do you think @NSilv ?

steve-chavez:
This feature would involve changes in many pgrst modules, but I think it should be straightforward.

`on_conflict=cola,colb` just needs a list of columns and pg will only admit it if a constraint is defined on those columns, so validation on our part would not be needed. This new query param should complement `merge-duplicates/ignore-duplicates` as a way to override the on conflict target.

A rough guideline for implementing this would be:

1. Add the type for the on conflict values(actually this is already available)
https://github.com/PostgREST/postgrest/blob/ea970554491557d79408ee2193d8a5f08d2984ea/src/PostgREST/Types.hs#L378-L385

2. Define the querystring param in(should only be allowed for POST, otherwise ignored)
https://github.com/PostgREST/postgrest/blob/ea970554491557d79408ee2193d8a5f08d2984ea/src/PostgREST/ApiRequest.hs#L76

[...]

Issue #1106, opened 2018-05-10
Proposal for allowing multiple schemas in a single instance


steve-chavez:
This is needed both for **api versioning** and **schema based multitenancy**, currently it can be done with many PostgREST instances but it's unwieldy as there's considerable ops overhead, I'll make my proposal from the api versioning perspective.

## Specifying the API version

Traditionally this has been done by having a prefix url path like `/v1/projects`, other options include a header that specifies the version or a date, like Stripe does, or a mime type that appends the version(`application/vnd.com.v2+json`).

All of this approaches are global and if we want to make them work with just SQL it would imply having to  copy all the unmodified views from `v1` schema to `v2`, ideally only the views that contain breaking changes would be in `v2`.

We can reuse PostgreSQL dot notation in the url to version the resources, this also has the advantage of being able to define embeds and specify the target resource version:
```http
GET /v1.projects?select=id,name,v2.clients(id,name,active)&clients.active=is.true
```
(Notice that in the filter `v2.clients` is just referred as `clients`, the schema can only be specified at the url path and in the `select` part, this will ensure that we don't break our current filter syntax) 

## Maintaning backwards compatibility

I think that REST API evolution can be handled in a similar way to regular software packages evolution,   
we can use schema names that follow SemVer, or maybe just a MAJOR MINOR convention(like `v1_2`, underscore used to avoid conflicts with `.`) to represent additions/breaking changes, comments on columns/views/schemas can notify deprecations and maybe release date. 

This combined with the schemas in the `search_path`, specified through `db-schema` as [previously discussed](https://github.com/PostgREST/postgrest/pull/1003#issuecomment-338557113), would allow us to maintain backwards compatibility:
```sql
-- For maintaing backward compatibility on breaking change
SET search_path TO v1_0, v2_0;
GET /projects -- This would query FROM v1_0.projects
GET /v2_0.projects -- Client would have to specify the schema to use the latest version

-- For maintaing backward compatibility on new addition
SET search_path TO v1_1, v1_0, v2_0;
GET /projects -- This would query FROM v1_1.projects
``` 
Dropping support can be achieved by removing the schema from the `search_path` and maybe doing a redirect with a 3xx to the new url, this part would have to be done in the proxy layer(maybe later when adding #1088 this could be done in the `pre-request` function).

## Security considerations

[...]

PR #1432, opened 2019-12-28
Upsert with "Prefer: resolution=merge-duplicates" on specified columns (#1327)

ykst:
This introduces new optional query syntax  `on_conflict=key1,key2,...` for POST, as discussed on #1327.  It enables upsert on keys with unique constraint other than primary keys. 

## Usage

Given unique constraint on `unique_key`;

```
POST /employees?on_conflict=unique_key HTTP/1.1
Prefer: resolution=merge-duplicates
[
  { "unique_key": 1, "value": "A" }
]
```

performs upsert with `unique_key` instead of primary keys.

For compound unique keys, comma-separated list of column names should be given.

When multiple `on_conflict` terms are given, only the first one is used (like `select`).

## Compatibility

Comply with the semantics of `merge-duplicates/ignore-duplicates` in request headers, and `Preference-Applied` in response headers as with upsert on primary keys.

### DIFF
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -324,7 +324,11 @@ addProperty f (targetNodeName:remainingPath, a) (Node rn forest) =
 mutateRequest :: Schema -> TableName -> ApiRequest -> S.Set FieldName -> [FieldName] -> ReadRequest -> Either Response MutateRequest
 mutateRequest schema tName apiRequest cols pkCols readReq = mapLeft errorResponseFor $
   case action of
-    ActionCreate -> Right $ Insert qi cols ((,) <$> iPreferResolution apiRequest <*> Just pkCols) [] returnings
+    ActionCreate -> do
+        confCols <- case iOnConflict apiRequest of
+            Nothing -> pure pkCols
+            Just param -> pRequestOnConflict param
+        pure $ Insert qi cols ((,) <$> iPreferResolution apiRequest <*> Just confCols) [] returnings
     ActionUpdate -> Update qi cols <$> combinedLogic <*> pure returnings
     ActionSingleUpsert ->
       (\flts ->
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -326,7 +326,7 @@ mutateRequest schema tName apiRequest cols pkCols readReq = mapLeft errorRespons
   case action of
     ActionCreate -> do
         confCols <- case iOnConflict apiRequest of
-            Nothing -> pure pkCols
+            Nothing    -> pure pkCols
             Just param -> pRequestOnConflict param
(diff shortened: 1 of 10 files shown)


src/PostgREST/DbRequestBuilder.hs
```diff
@@ -324,7 +324,11 @@ addProperty f (targetNodeName:remainingPath, a) (Node rn forest) =
 mutateRequest :: Schema -> TableName -> ApiRequest -> S.Set FieldName -> [FieldName] -> ReadRequest -> Either Response MutateRequest
 mutateRequest schema tName apiRequest cols pkCols readReq = mapLeft errorResponseFor $
   case action of
-    ActionCreate -> Right $ Insert qi cols ((,) <$> iPreferResolution apiRequest <*> Just pkCols) [] returnings
+    ActionCreate -> do
+        confCols <- case iOnConflict apiRequest of
+            Nothing -> pure pkCols
```
steve-chavez: The [error](https://circleci.com/gh/PostgREST/postgrest/3968?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) on CI it's related to alignment. Try:
```suggestion
            Nothing    -> pure pkCols
```

src/PostgREST/Parsers.hs
```diff
@@ -30,6 +30,12 @@ pRequestSelect :: Text -> Either ApiRequestError [Tree SelectItem]
 pRequestSelect selStr =
   mapError $ parse pFieldForest ("failed to parse select parameter (" <> toS selStr <> ")") (toS selStr)
 
+pRequestOnConflict :: Text -> Either ApiRequestError  [FieldName]
+pRequestOnConflict oncStr =
+  mapError $ parse columns ("failed to parse on_conflict parameter (" <> toS oncStr <> ")") (toS oncStr)
+  where
+    columns = pFieldName `sepBy1` char ','
```
steve-chavez: You could omit this `where` if you reuse the `pColumns` function:
https://github.com/PostgREST/postgrest/blob/e12c1319b605dea2074ad521b0b74f869da67de4/src/PostgREST/Parsers.hs#L237-L238

steve-chavez:
@ykst Really nice PR with tests included. Check my comments for some minor corrections.

Also, edit the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md#added) with the addition.

ErwinJanssen:
I'm happy to see that this feature is implemented! I would like to use this feature in my projects, so I was wondering if there is an estimation for when a new version of PostgREST will be released.

steve-chavez:
@ErwinJanssen Release should be in this month's last week. I want to include the [multiple schemas enhancement](https://github.com/PostgREST/postgrest/issues/1106) for this release.

If you need the feature now, you could [build from source](https://postgrest.org/en/latest/install.html#build-source).

Issue #1422, opened 2019-12-01
Workaround for GET /rpc/function to STABLE function while setting response headers


francois:
### Environment

* PostgreSQL version:  PostgreSQL 12.1 on x86_64-apple-darwin18.7.0, compiled by Apple clang version 11.0.0 (clang-1100.0.33.12), 64-bit
* PostgREST version: PostgREST 6.0.2 (UNKNOWN)
* Operating system: Mac OS X Mojave (10.14.6 (18G103))

### Description of issue

I wanted to serve IMG blobs straight from the database. I created the following function to be called from `/rpc/get_blob?blob_slug=aslug`

```sql
CREATE OR REPLACE FUNCTION api.get_blob(blob_slug text) RETURNS bytea AS $$
DECLARE
  res RECORD;
  headers TEXT;
BEGIN
  SELECT content_type, bytes INTO res
  FROM public.blobs
  WHERE blobs.blob_slug = $1;

  headers := CONCAT(
    '[{"Cache-Control": "public"}, {"Cache-Control": "max-age=259200", "Content-Type": "', res.content_type, '"}]'
  );

  SET LOCAL "response.headers" = headers;

  RETURN res.bytes;
END
$$ LANGUAGE 'plpgsql' STABLE;
```

PostgREST returns a 400 Bad Request when calling the function:

```
$ curl -sv -HAccept:"image/webp,*/*" 'http://localhost:3000/rpc/get_blob?blob_slug=cffikqsz' | strings
> GET /rpc/get_blob?blob_slug=cffikqsz HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.54.0
> Accept:image/webp,*/*
>
< HTTP/1.1 400 Bad Request
< Transfer-Encoding: chunked
< Date: Sun, 01 Dec 2019 02:27:53 GMT
< Server: postgrest/6.0.2 (UNKNOWN)
< Content-Type: application/json; charset=utf-8
<
{ [114 bytes data]
{"hint":null,"details":null,"code":"0A000","message":"SET is not allowed in a non-volatile function"}
```

I wanted to be able to set the `Content-Type` header so that the browser would be correctly informed of the actual content of the response's body. Notice the `Accept` header in curl above: this is the same header that Firefox sends. If I remove `SET LOCAL` in the function, the response becomes:

```
$ curl -sv -HAccept:"image/webp,*/*" 'http://localhost:3000/rpc/get_blob?blob_slug=cffikqsz' | strings
[...]

PR #1435, opened 2020-01-15
Add response.headers GUC for GET/POST/PATCH/PUT/DELETE

### LINKED ISSUE #1088: Add headers to all responses through "response.headers" GUC
Our development  is trying to add the response headers per latest rest  documentations.
Our client tools see no addition to the headers.

 pkg info | grep rest
postgrest-0.4.4.0              RESTful API for PostgreSQL

FreeBSD 11.1

i.e 

I am suspicious about your test for version_num and version ( 9.6.8 vs  9.6 )   I am not familiar enough with haskell to claim that definitively but if this version test fails that would explain the lack of the header.

Our values:

 SELECT current_setting('server_version_num')::integer, current_setting('server_version')
ace_db-> ;
 current_setting | current_setting
-----------------+-----------------
           90608 | 9.6.8

The response.headers variable is being setup by the ALTER role set "response.header" = .....

 select coalesce(nullif(current_setting('response.headers', true), ''), '[]') :: Text ;
                           coalesce
---------------------------------------------------------------
 [{"Cache-Control": "public"}, {"Cache-Control": "max-age=2"}]

I am thinking the header is not sent because the version comparison is failing ? 

Any suggestions/Comments ?

Thanks 

Dave

### LINKED ISSUE #1168: How to access the authorization header value?
### Environment

* PostgreSQL version: 9.6.8
* PostgREST version: 0.5.0.0
* Operating system: FreeBSD

### Description of issue

Trouble accessing the `request.header.authorization` variable.  I can access every other header in the request, even application-specific ones like `postman-token`.  But for some reason I cannot access the `authorization` header.   I verified that it is being sent - it has to be since `request.jwt.claim.role` is working.  I need some way to access the entire token string -- I tried everything I could think of. Exactly what variables are set up anyway?  Is there a list somewhere?  Is there some other way to access the raw authorization token?

Schema:
```
CREATE OR REPLACE FUNCTION api.test1() RETURNS VOID AS $$
BEGIN
    RAISE WARNING 'request.header.authorization = %', current_setting('request.header.authorization');
EXCEPTION WHEN OTHERS THEN
    RAISE WARNING 'request.header.authorization = (no way)';
END;
BEGIN
    RAISE WARNING 'request.header = %', current_setting('request.header');
EXCEPTION WHEN OTHERS THEN
    RAISE WARNING 'request.header = (no way)';
END;
BEGIN
    RAISE WARNING 'request.headers = %', current_setting('request.headers');
EXCEPTION WHEN OTHERS THEN
    RAISE WARNING 'request.headers = (no way)';
END;
BEGIN
    RAISE WARNING 'request.jwt = %', current_setting('request.jwt');
EXCEPTION WHEN OTHERS THEN
    RAISE WARNING 'request.jwt = (no way)';
END;
BEGIN
    RAISE WARNING 'request = %', current_setting('request');
EXCEPTION WHEN OTHERS THEN
    RAISE WARNING 'request = (no way)';
END;

RAISE WARNING 'request.header.accept = %', current_setting('request.header.accept');
RAISE WARNING 'request.header.accept-encoding = %', current_setting('request.header.accept-encoding');
RAISE WARNING 'request.header.cache-control = %', current_setting('request.header.cache-control');
RAISE WARNING 'request.header.content-length = %', current_setting('request.header.content-length');
RAISE WARNING 'request.header.content-type = %', current_setting('request.header.content-type');
RAISE WARNING 'request.header.user-agent = %', current_setting('request.header.user-agent');
RAISE WARNING 'request.header.postman-token = %', current_setting('request.header.postman-token');
RAISE WARNING 'request.jwt.claim.role = %', current_setting('request.jwt.claim.role');
$$ LANGUAGE plpgsql;
```
Request:
```
GET
https://xxx.yyy/api/v1/rpc/test1
Request Headers:
   content-type:"application/json"
   cache-control:"no-cache"
   postman-token:"b12193c7-2e96-4674-96dc-87d9c2afde13"
   authorization:"Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIiA6ICJyZXN0X3VzZXJfbWF4IiwgInJlZGNvbV9zaWdtYV9yZXN0X3VzZXJfaWQiIDogMSwgInJlZGNvbV9zaWdtYV9yZXN0X21vZGVfaWQiIDogMCwgInJlZGNvbV9zaWdtYV9yZXN0X3ZzX2lkIiA6IDAsICJqdGkiIDogImpWL1hqQTlwMzVkbWlwOGt3RmE2SUIydG9vWnFrT3Z2ZWtVbnc5OXpReHM9IiwgImlzcyIgOiAiZmxvcmlkYS5yZWRjb20uY29tIiwgImlhdCIgOiAxNTM0Mjc3MjY2LCAibmJmIiA6IDE1MzQyNzcyMzYsICJleHAiIDogMTUzNDU3NzI2Nn0.ci2vzUiez7kiIYbtThf9ySnksNk2-yp96bdm1_BfveY"
   user-agent:"PostmanRuntime/7.2.0"
   accept:"*/*"
   accept-encoding:"gzip, deflate"
   content-length:55
```
Output:
```
2018-08-14 ...  [3-1] WARNING:  request.header.authorization = (no way)
2018-08-14 ...  [4-1] WARNING:  request.header = (no way)
2018-08-14 ...  [5-1] WARNING:  request.headers = (no way)
2018-08-14 ...  [6-1] WARNING:  request.jwt = (no way)
2018-08-14 ...  [7-1] WARNING:  request = (no way)
2018-08-14 ...  [8-1] WARNING:  request.header.accept = */*
2018-08-14 ...  [9-1] WARNING:  request.header.accept-encoding = gzip, deflate
2018-08-14 ...  [10-1] WARNING:  request.header.cache-control = no-cache
2018-08-14 ...  [11-1] WARNING:  request.header.content-length = 55
2018-08-14 ...  [12-1] WARNING:  request.header.content-type = application/json
2018-08-14 ...  [13-1] WARNING:  request.header.user-agent = PostmanRuntime/7.2.0
2018-08-14 ...  [14-1] WARNING:  request.header.postman-token = 03d33f80-9d2d-4c83-8c47-22d5bb18c86b
2018-08-14 ...  [15-1] WARNING:  request.jwt.claim.role = rest_user_max
```

### LINKED ISSUE #1427: Override unsuitable Location header values
### Environment

* PostgreSQL version: (if using docker, specify the image)
  11.5
* PostgREST version: (if using docker, specify the image)
  6.0.2
* Operating system:
  FreeBSD 11.3-RELEASE-p5
### Description of issue

For some of our endpoints the Location header does not use the value we want. In some cases there is no Location header returned, and in other cases it uses an unsuitable combination of columns.

In the case where it does not return a Location header, we were able to overcome this by putting the following in our `INSTEAD OF INSERT` trigger function:
```sql
    -- Provide a value for the Location header
    IF NEW.id IS NOT NULL
    THEN
        PERFORM SET_CONFIG(
                'response.headers',
                FORMAT('[{"Location": "/%s?id=eq.%s"}]', TG_TABLE_NAME, NEW.id),
                FALSE
            );
    END IF;
```

The trouble is for the case where PostgREST uses an unsuitable value. In this case the above code does not seem to override the value chosen by PostgREST.  This is especially problematic when the chosen columns contain huge amounts of data, which exceed the 4KB limit for header info.

So far we have not been able to do anything to overcome this, except disable the Location header completely in the web server configs for the affected endpoints.

If PostgREST were to only set the Location header value if it was not already set (assuming PostgREST is setting it _after_ the INSERT happens) -- then it seems there would be a way to override it when the default value was unsuitable.

I looked at the code (src/PostgREST/QueryBuilder/WriteStatement.hs) a bit but Haskell is like egyptian hieroglyphics to me -- I did not see how to do it...

Also maybe consider NOT setting the Location header if the resulting string is more than X bytes long (2000? 3000?).

steve-chavez:
Custom headers improvements.

- Fixes #1088.
- Fixes #1168.
- Aiming to Fix #1427. Also thinking to allow overriding Content-Type(requested in https://github.com/PostgREST/postgrest/issues/1422) and Content-Range(requested in chat https://gitter.im/begriffs/postgrest?at=5dcacd06e75b2d5a19bcfb2c).

### DIFF
--- a/BACKERS.md
+++ b/BACKERS.md
@@ -42,6 +42,7 @@ PostgREST ongoing development is only possible thanks to our Sponsors and Backer
 - Christopher Reid
 - Nathan Bouscal
 - Daniel Rafaj
+- David Fenko
 
 ## Former Backers
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1378, Add support for `Prefer: count=planned` and `Prefer: count=estimated` on GET /table - @steve-chavez
 - #1327, Add support for optional query parameter `on_conflict` to upsert with specified keys for POST - @ykst
 - #1430, Allow specifying the foreign key constraint name(`/source?select=fk_constraint(*)`) to disambiguate an embedding - @steve-chavez
-- #1168, Allow access to the Authorization header through the request.header.authorization GUC - @steve-chavez
-- #1435, Add request.method and request.path GUCs - @steve-chavez
+- #1168, Allow access to the `Authorization` header through the `request.header.authorization` GUC - @steve-chavez
+- #1435, Add `request.method` and `request.path` GUCs - @steve-chavez
+- #1088, Allow adding headers to GET/POST/PATCH/PUT/DELETE responses through the `response.headers` GUC - @steve-chavez
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1378, Add support for `Prefer: count=planned` and `Prefer: count=estimated` on GET /table - @steve-chavez
 - #1327, Add support for optional query parameter `on_conflict` to upsert with specified keys for POST - @ykst
 - #1430, Allow specifying the foreign key constraint name(`/source?select=fk_constraint(*)`) to disambiguate an embedding - @steve-chavez
+- #1168, Allow access to the Authorization header through the request.header.authorization GUC - @steve-chavez
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1168, Allow access to the `Authorization` header through the `request.header.authorization` GUC - @steve-chavez
 - #1435, Add `request.method` and `request.path` GUCs - @steve-chavez
 - #1088, Allow adding headers to GET/POST/PATCH/PUT/DELETE responses through the `response.headers` GUC - @steve-chavez
+- #1427, Allow overriding provided headers(Location, Content-Type, etc) through the `response.headers` GUC - @steve-chavez
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - #1327, Add support for optional query parameter `on_conflict` to upsert with specified keys for POST - @ykst
 - #1430, Allow specifying the foreign key constraint name(`/source?select=fk_constraint(*)`) to disambiguate an embedding - @steve-chavez
 - #1168, Allow access to the Authorization header through the request.header.authorization GUC - @steve-chavez
+- #1435, Add request.method and request.path GUCs - @steve-chavez
 
 ### Fixed
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -94,6 +94,8 @@ data ApiRequest = ApiRequest {
   , iJWT                  :: Text                             -- ^ JSON Web Token
   , iHeaders              :: [(Text, Text)]                   -- ^ HTTP request headers
   , iCookies              :: [(Text, Text)]                   -- ^ Request Cookies
+  , iPath                 :: ByteString                       -- ^ Raw request path
+  , iMethod               :: ByteString                       -- ^ Raw request method
   }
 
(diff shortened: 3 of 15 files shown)


steve-chavez:
Added the ability to override all postgrest provided headers.

steve-chavez:
I remember we got a question in gitter chat regarding setting the http status through a `response.status` GUC, can't find it though. We currently support setting status through raise http://postgrest.org/en/v6.0/api.html#errors-and-http-status-codes, though that's more meant for errors and will cancel the body pgrst generates.

Should we allow manipulating the status for successful responses as well? This would make our feature set for customizable http logic complete.

Setting the `response.status` GUC also seems more intuitive than setting the sqlstate to our special `PT***` code.

steve-chavez:
I'll leave `response.status` for another time. Raising the `PT***` error should be enough for most use cases.

I think this is good to merge now.

PR #1436, opened 2020-01-19
multiple schemas in one instance

mahmoudkassem:


test/Feature/JsonOperatorSpec.hs
```diff
@@ -26,7 +26,7 @@ spec actualPgVersion = describe "json and jsonb operators" $ do
 
     it "fails on bad casting (data of the wrong format)" $
       get "/complex_items?select=settings->foo->>bar::integer"
-        `shouldRespondWith` [json| {"hint":null,"details":null,"code":"22P02","message":"invalid input syntax for integer: \"baz\""} |]
+        `shouldRespondWith` [json| {"hint":null,"details":null,"code":"22P02","message":"invalid input syntax for type integer: \"baz\""} |]
```
steve-chavez: What PostgreSQL version are you using? pg 12? If so you could conditionally validate the output according to the version. There's a similar logic below:

https://github.com/PostgREST/postgrest/blob/9b1224827a4c4b71ffe924e54f024b08df93fecf/test/Feature/JsonOperatorSpec.hs#L54-L64 

This would make the test runs go green.

test/Feature/QuerySpec.hs
```diff
@@ -886,7 +886,7 @@ spec actualPgVersion = do
     it "only returns an empty result set if the in value is empty" $
       get "/items_with_different_col_types?int_data=in.( ,3,4)"
         `shouldRespondWith`
-        [json| {"hint":null,"details":null,"code":"22P02","message":"invalid input syntax for integer: \"\""} |]
+        [json| {"hint":null,"details":null,"code":"22P02","message":"invalid input syntax for type integer: \"\""} |]
```
steve-chavez: Ditto.

steve-chavez:
My bad, just merged #1435 and it caused some conflicts. Could you rebase?

Also see my comments above.

test/Feature/JsonOperatorSpec.hs
steve-chavez: You'll need to add pg12 version in https://github.com/PostgREST/postgrest/blob/9b1224827a4c4b71ffe924e54f024b08df93fecf/src/PostgREST/Types.hs#L506-L513

mahmoudkassem: Yes, I'm running version 12.1 adjusted the test case accordingly, thanks for pointing this out. I just saw it being red and fixed it according to the expected output. I assumed it to be a typo.

test/Feature/QuerySpec.hs
mahmoudkassem: adjusted accordingly, thanks again for pointing this out and for the detailed description on how to resolve this

steve-chavez:
Hmm.. the merge commits makes the PR a bit hard to review for me. When you think it's ready for review, please try to rebase(`git rebase -i master`) the branch.

(another choice maybe would be to `cherry-pick` each commit you made in a new branch started from upstream/master and then rename this new branch to master, finally `push origin -f`)

src/PostgREST/ApiRequest.hs
```diff
@@ -208,13 +208,20 @@ userApiRequest schema rootSpec req reqBody
       "DELETE"  -> ActionDelete
       "OPTIONS" -> ActionInfo
       _         -> ActionInspect{isHead=False}
+  schema = case lookupHeader "Accept-Version" of
+             Nothing      -> case schemas of
+                               []                  -> ""
```
steve-chavez: An empty schema list shouldn't be possible. Maybe you could change the `configSchema` type to a [Data.List.NonEmpty](http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List-NonEmpty.html#t:NonEmpty).

mahmoudkassem: Yes the config parser makes sure this isn't possible. I wasn't aware of this type that's why I put the empty list case so the compiler would't complain. Thanks for introducing this, it is nice to be able to be this precise.

mahmoudkassem:
As for when I would say this is ready, there are 2 more things to do form what I can tell. First is add the Open API tests and second implement a check that forbids cross schema queries to go through unless both source and target schema are explicitly added in the db-schema parameter(and a test for that as well). Once I have these I can clean up to exclude the merge commits.

steve-chavez:
@mahmoudkassem Would you like me to help you with the OpenAPI tests? I think if you check `Allow edits from maintainers` in the PR I can push to this branch.

mahmoudkassem:
@steve-chavez Sorry didn't have any time to work on this(new job, new apartment and some more stuff). Things are currently cooling down. I should be able to resume working on this during this week.

The tests should be fine, one for the first test schema as default and one where it is passed explicitly in the header and maybe add one more table to the second test schema to distinguish them and one more 404 test with an unknown schema. What is unclear to me is how to restrict cross schema access in the q parameter to schemas in the db-schema parameter and fail the request otherwise.

The "Allow edits from maintainers" checkbox was already set. Support especially on the second point would be much appreciated.

main/Main.hs
```diff
@@ -175,6 +176,9 @@ main = do
         . setServerName (toS $ "postgrest/" <> prettyVersion) $
         defaultSettings
 
+  -- the first value in the db-schema configuration parameter is used as default
+  let defaultSchema = head $ configSchemas conf
```
steve-chavez: I've noted that taking the head of the schemas would make the procs only work for a single schema(not the whole list). This is because we obtain some info about procs in DbStructure.
https://github.com/PostgREST/postgrest/blob/b75e7cef906b1703eb14ffcc32723eacff30ee11/src/PostgREST/DbStructure.hs#L47-L55

We would need to pass the whole list here and change some of the DbStructure queries. I can help with this.

steve-chavez:
>  (new job, new apartment and some more stuff)

@mahmoudkassem Congrats!

I can help you with the part you mentioned. Also I've noted there is other stuff we need to modify(DbStructure).

I've excluded the merges on this branch and cherry-picked your commits on https://github.com/steve-chavez/postgrest/commits/mult-schemas. Let me know if that branch is ok and I'll force push it here so we can continue working on this.

mahmoudkassem:
@steve-chavez looks good all non-merge commits are there.

steve-chavez:
@mahmoudkassem Sorry, I made a mistake when force pushing here :disappointed: .

I did `git push mahmoud master -f` instead of `git push mahmoud mahmoud:master -f`  :man_facepalming:. And now I can't push to your master branch.

Could you update your master branch with the commits on https://github.com/steve-chavez/postgrest/commits/multiple-schemas and push it here? (github automatically closed this one). You could also open another PR if there's any trouble.

PR #1441, opened 2020-01-28
Change server-proxy-uri to openapi-server-proxy-uri

### LINKED ISSUE #1413: Change misleading config options names
These config options often cause confusion among users:

- [server-proxy-uri](http://postgrest.org/en/v6.0/configuration.html#server-proxy-uri): the name doesn't imply it's related to the openapi output and misleads into thinking it's a web server setting. It'd be better to change it to `openapi-server-proxy-uri`.

- [db-extra-search-path](http://postgrest.org/en/v6.0/configuration.html#db-extra-search-path)(name was discussed in https://github.com/PostgREST/postgrest/issues/1200): this always seems related to exposing more than one schema. Since the original intent was only to make it easier to work with extensions, I suggest renaming it to `db-extensions-schema`. Default still would be public and it'll keep accepting a schema list.

steve-chavez:
Fixes #1413. 

I've noted that if an unspecified config option(like `xx-yy=33`) is set then the config file is accepted with no error shown. 

That means we'd need to add additional code to make sure a removed config is no longer accepted.

In the case of `server-proxy-uri` just changing the config option name is not that critical. I've left `db-extra-search-path` as it is.


brentybh:
Using binary from `postgrest-v7.0.0-ubuntu.tar.xz`
but `openapi-server-proxy-uri` doesn't work?
the legacy `server-proxy-uri` works fine

steve-chavez:
@brentybh You're right. This somehow slipped by during a rebase https://github.com/PostgREST/postgrest/blame/master/src/PostgREST/Config.hs#L157.

And now we have `server-proxy-uri` https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/Config.hs#L157.

Could you open an issue for this?  Maybe we can release a 7.0.1 with some bug fixes.

PR #1450, opened 2020-02-24
multiple schemas in one instance

steve-chavez:
Continuing the work on #1436.

### DIFF
--- a/test/Feature/MultipleSchemaSpec.hs
+++ b/test/Feature/MultipleSchemaSpec.hs
@@ -54,8 +54,8 @@ spec =
     it "read another_table from other schema(v2)" $
       request methodGet "/another_table" [("Accept-Version", "v2")] "" `shouldRespondWith`
         [json|[
-          {"id":1,"value":"value5"},
-          {"id":2,"value":"value6"}
+          {"id":1,"another_value":"value5"},
+          {"id":2,"another_value":"value6"}
         ]|]
         {
           matchStatus = 200
@@ -173,14 +173,14 @@ spec =
                       "format" : "integer",
                       "type" : "integer"
                     },
-                    "value" : {
+                    "another_value" : {
                       "format" : "text",
                       "type" : "string"
                     }
                   },
                   "required" : [
                     "id",
-                    "value"
+                    "another_value"
                   ],
                   "type" : "object"
                 }
--- a/test/Feature/MultipleSchemaSpec.hs
+++ b/test/Feature/MultipleSchemaSpec.hs
@@ -112,14 +112,46 @@ spec =
             matchStatus = 406
           }
 
+    context "calling procs on different schemas" $ do
+      it "succeeds in calling the default schema proc" $
+        request methodGet "/rpc/get_parents_below?id=6" [] ""
+          `shouldRespondWith`
+          [json|[{"id":1,"name":"parent v1-1"}, {"id":2,"name":"parent v1-2"}]|]
+          {
+            matchStatus = 200
+          , matchHeaders = [matchContentTypeJson, "Content-Profile" <:> "v1"]
+          }
+
+      it "succeeds in calling the v1 schema proc and embedding" $
+        request methodGet "/rpc/get_parents_below?id=6&select=id,name,childs(id,name)" [("Accept-Profile", "v1")] ""
+          `shouldRespondWith`
+          [json| [
+            {"id":1,"name":"parent v1-1","childs":[{"id":1,"name":"child 1"}]},
+            {"id":2,"name":"parent v1-2","childs":[{"id":2,"name":"child 2"}]}] |]
+          {
+            matchStatus = 200
+          , matchHeaders = [matchContentTypeJson, "Content-Profile" <:> "v1"]
+          }
+
+      it "succeeds in calling the v2 schema proc and embedding" $
+        request methodGet "/rpc/get_parents_below?id=6&select=id,name,childs(id,name)" [("Accept-Profile", "v2")] ""
+          `shouldRespondWith`
+          [json| [
+            {"id":3,"name":"parent v2-3","childs":[{"id":1,"name":"child 3"}]},
+            {"id":4,"name":"parent v2-4","childs":[]}] |]
+          {
+            matchStatus = 200
+          , matchHeaders = [matchContentTypeJson, "Content-Profile" <:> "v2"]
+          }
+
     context "OpenAPI output" $ do
       it "succeeds in reading table definition from default schema v1 if no schema is selected via header" $ do
-          r <- request methodGet "/" [] ""
+          req <- request methodGet "/" [] ""
 
           liftIO $ do
-            simpleHeaders r `shouldSatisfy` matchHeader "Content-Profile" "v1"
+            simpleHeaders req `shouldSatisfy` matchHeader "Content-Profile" "v1"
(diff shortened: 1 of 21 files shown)


steve-chavez:
@mahmoudkassem Regarding your issue on https://github.com/PostgREST/postgrest/pull/1436#issuecomment-578444296

> implement a check that forbids cross schema queries to go through unless both source and target schema are explicitly added in the db-schema parameter

What I've done is to reject the request at the http level(it doesn't hit the database, so no need for tailoring an invalid sql query). I've taken into account the [Accept-Profile](https://profilenegotiation.github.io/I-D-Accept--Schema/I-D-accept-schema#rfc.section.3) proposal recommendations:

**If the server cannot process the specified profile, it would answer with an http 406 status code and possibly a list of acceptable profiles.**

 So, if the user specifies an invalid schema he'll get an error:
```
curl -H "Accept-Version: v5" localhost:3000/table
{"message":"The schema must be one of the following: v1, v2"} 
```

I've also solved the dbstructure issue(mentioned in https://github.com/PostgREST/postgrest/pull/1436#discussion_r381535990).

Let me know what you think.

mahmoudkassem:
@steve-chavez great job! Now all that's left is the OpenAPI test, right? Will try to get them done tomorrow.

steve-chavez:
@mahmoudkassem I've noticed that with the `Accept-Version` approach we can GET different schemas resources. But how about a POST or PATCH to a different schema resource?
With the current interface that cannot be done.

However If we fully adopt the [Content Negotiation by Profile](https://www.w3.org/TR/dx-prof-conneg/) spec we could have that working by doing:

```
POST /items
Content-Profile: v2 

{....}
```

That would make an insert to the `v2.items` relation.
I'll have to read the spec more carefully and see if we can do it that way.

For now, you adding the missing OpenAPI tests would be great.

steve-chavez:
To fully adopt the profile negotiation spec, we would need to implement two requests: [list profiles](https://www.w3.org/TR/dx-prof-conneg/#listprofiles) and [get resource by profile](https://www.w3.org/TR/dx-prof-conneg/#getresourcebyprofile).

I'm asking some questions to the workgroup(https://github.com/w3c/dx-connegp/issues/15 and https://github.com/w3c/dx-connegp/issues/18) for how to do this correctly. 

As it is now, I think we can implement only the "get resource by profile" part(the `Accept-Profile`/`Content-Profile` requests, which would solve our issue).

I'd like to move forward and make a new release. We can work on supporting "list profiles" after the release and for now we'll make a note in the docs about it being a work in progress.

mahmoudkassem:
@steve-chavez the test is failing because it is omitting the "value" object in the properties as well as the "value" key in the "required" array, not sure what's wrong.

 
test_error

test/Feature/MultipleSchemaSpec.hs
```diff
@@ -0,0 +1,196 @@
...
+                    },
+                    "value" : {
+                      "format" : "text",
+                      "type" : "string"
+                    }
+                  },
+                  "required" : [
+                    "id",
+                    "value"
+                  ],
```
steve-chavez: The `required` array should only contain `id`. This is because only `id` is non-nullable(primary key) and value is not.

That logic is done in:
https://github.com/PostgREST/postgrest/blob/de218e900b4d30b98d2508a71045dadff25118a3/src/PostgREST/OpenAPI.hs#L60

test/Feature/MultipleSchemaSpec.hs
```diff
@@ -0,0 +1,196 @@
...
```
steve-chavez: Ditto. After correcting all of these the test suite should be green.

mahmoudkassem: In order for the tests to turn green I need to additionally remove the "value" object under "properties", but is the output I get when calling the endpoint
curl_output

mahmoudkassem:
@steve-chavez in my local database(not the one created for the tests) the fields where marked as not null that's why they were showing up in the required field when calling the service with curl. Now that error is gone, but I don't understand why the values are omitted under "properties".

steve-chavez:
My only guess for now is that the DbStructure doesn't contain the columns(`value`) but it does contain the primary keys and somehow this is only happening when running the test suite. There must be something related to the config there. I'd have to `traceShowId` to see what's going on.

steve-chavez:
Ahhh.. here's the culprit:

https://github.com/PostgREST/postgrest/blob/de218e900b4d30b98d2508a71045dadff25118a3/test/Main.hs#L58-L60

In the test suite the DbStructure is only loading the "test" schema. This needs refactoring so it doesn't happen again when testing other schemas.

mahmoudkassem:
@steve-chavez thank you for finding the issue, now that this is addressed its time to take a look at "get resource by profile" implementation.

steve-chavez:
@mahmoudkassem Regarding "get resource by profile", for **GET/HEAD** should be simple. Basically:

```http
GET /table
Accept-Profile: v1

HTTP/1.1 200 OK
Content-Profile: v1

body here
```
That is, when `Accept-Profile` is specified the response must contain a `Content-Profile` reply if it's possible to comply.

For **POST/PATCH/PUT/DELETE**, the implementation should be:
```http
POST /table
Content-Profile: v1

body here

HTTP/1.1 201 Created
```
That should also be straightforward but there's a catch here. 

Suppose we have a `private.table`, and we create a VIEW of this table in `v1.table`, another VIEW in `v2.table` and we expose these `v1, v2` schemas through postgrest.
 
Should we allow to POST a payload to a v1.table and respond with a v2.table representation? In theory, this could be done with the following headers:
```http
POST /table
Content-Profile: v1
Accept-Profile: v2
Prefer: return=representation

body here
```
For now I think we should not allow this and keep it simple. If a POST request with `Content-Profile`, `Accept-Profile` and `Prefer: return=representation` comes, assume the `Content-Profile` schema and add the `Content-Profile: v1` header to the response as well.

steve-chavez:
Come to think of it, if we list profiles with OPTIONS that would also solve https://github.com/PostgREST/postgrest/issues/790. Funny that https://github.com/PostgREST/postgrest/issues/790#issuecomment-312472581 also mentioned the profile spec as a possible solution. Though it was an older one where the profile could only be specified in the media type.

steve-chavez:
I've added a condition for checking profile negotiation is only possible when multiple schemas are specified. This is to ensure that users that would like to keep using a single schema don't see the `Content-Profile: v1` response.

Tests that are missing: POST/PATCH/PUT/DELETE, calling a stored proc and maybe a resource embedding one to ensure it all works normally.

steve-chavez:
I think I've finished my work here. The tests should cover all of the http verbs.

@mahmoudkassem Could you give it a final review and add an entry to the CHANGELOG?

mahmoudkassem:
@steve-chavez Everything looks fine to me, sorry for being so inactive things have been crazy lately and thank you for all your efforts, patience and all that I have learned. Glad to see this feature being seen through. I've added an entry to CHANGELOG.md , if you want to add something please feel free to do so.

steve-chavez:
@mahmoudkassem Thanks a lot for your work and help here. Without your initiative this feature wouldn't have been done by now. So I'm also giving you credit on the CHANGELOG.

I'll merge and make a new release :rocket:

PR #1454, opened 2020-03-07
Update bounds for GHC 8.8 / stackage lts-15

robx:
This relaxes upper bounds to allow building with GHC 8.8 (and the corresponding stackage set).

It also raises configurator-pg lower bound because it turns out that an explicit instance for MonadFail is required.

I've compiled this with stackage lts-15.2 and ghc 8.8.3 succesfully but did not run tests yet. There are some unused import warnings that it seems can't be fixed while staying compatible with older GHC versions.


robx:
In order to build with stackage lts-15.2, this should be combined with #1456

robx:
Ran tests with stackage lts-15.2 via #1455.

robx:
Updated the lower bound for configurator-pg to 0.1.0.6, which includes a fix for building on GHC 7.10.

steve-chavez:
Looks good :+1: .

PR #1455, opened 2020-03-07
Update stack.yaml to GHC 8.8

robx:
This is on top of #1454 and #1456, and probably not suitable for merging.

It fixes redundant import warnings for GHC 8.8, which appears to cause trouble with older GHC versions (I'm currently not precisely sure how far back).

The point of this PR is to

- save the import warnings fix somewhere
- get CI to run with GHC 8.8 to prove #1454 ok


robx:
Test failures appear to be memory-related while building haskell-src-exts. Where that works, tests seem to pass fine.

robx:
Closing since it's done the job of showing that tests still pass, and is not mergeable because some of those imports are required with lower GHC versions.

Issue #1451, opened 2020-02-26
Support hspec-wai 0.10


felixonmars:
### Environment

* PostgREST version: 6.0.2
* Operating system: Arch Linux x86_64

### Description of issue

Currently postgrest's test suite doesn't build with hspec-wai 0.10 series:

```
Building test suite 'spec' for postgrest-6.0.2..
[ 1 of 30] Compiling Feature.ConcurrentSpec ( test/Feature/ConcurrentSpec.hs, dist/build/spec/spec-tmp/Feature/ConcurrentSpec.dyn_o )

test/Feature/ConcurrentSpec.hs:43:30: error:
    • Expecting one more argument to ‘WaiSession’
      Expected kind ‘* -> *’, but ‘WaiSession’ has kind ‘* -> * -> *’
    • In the second argument of ‘MonadBaseControl’, namely ‘WaiSession’
      In the instance declaration for ‘MonadBaseControl IO WaiSession’
   |
43 | instance MonadBaseControl IO WaiSession where
   |                              ^^^^^^^^^^

test/Feature/ConcurrentSpec.hs:52:23: error:
    • Expecting one more argument to ‘WaiSession’
      Expected kind ‘* -> *’, but ‘WaiSession’ has kind ‘* -> * -> *’
    • In the second argument of ‘MonadBase’, namely ‘WaiSession’
      In the instance declaration for ‘MonadBase IO WaiSession’
   |
52 | instance MonadBase IO WaiSession where
   |                       ^^^^^^^^^^
```

It would be nice if the new version could be supported.

steve-chavez:
@felixonmars Fixed in #1456.

felixonmars:
Thanks!

PR #1456, opened 2020-03-07
Upgrade hspec-wai to 0.10

robx:
This updates the tests to use hspec-wai and hspec-wai-json >= 0.10, which adds support for custom state (which we don't use). This came out of trying to get tests to work on GHC 8.8 (#1455), but could be considered independently.

The code changes I made were naive attempts to fix compilation. There might be better ways to do this.


robx:
Oh I just saw there's a corresponding issue: #1451. This should fix that.

steve-chavez:
Looks good to me! I think the code changes are straightforward.

Thanks a lot for helping in keeping the dependencies up-to-date @robx.

PR #1463, opened 2020-03-21
Add docker-compose file to run tests

wolfgangwalther:
Tried to run the tests with docker according to the docs and came up with this docker-compose file.

Tests can be run with:
`docker-compose up test`

Although the first run will fail, while the database container is still created and not accepting connections, yet - calling it again works. Calling `docker-compose up postgres` separately before would as well.

Some notes regarding the other changes:
- Had to add a couple of missing dependencies to `test/Dockerfile.test`, otherwise it wouldn't build for me.
- Changing `-a` to `-q` for the psql command was needed to stop psql from printing non-ascii characters to stdout, which readProcess would fail to read correctly. It would just throw with `spec: fd:14: hGetContents: invalid argument (invalid byte sequence)`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,30 @@
+version: "3.6"
+
+services:
+  test:
+    build:
+      context: .
+      dockerfile: test/Dockerfile.test
+    depends_on:
+      - postgres
+    working_dir: /root/postgrest
+    volumes:
+      - ./:/root/postgrest
+      - stack-linux:/root/.stack
+      - stack-work:/root/postgrest/.stack-work
+    command: bash -c "POSTGREST_TEST_CONNECTION=$$(test/create_test_db 'postgres://postgres:postgres@postgres' test_db) stack test"
+
+  postgres:
+    image: postgres:11
+    environment:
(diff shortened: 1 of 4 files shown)


test/SpecHelper.hs
```diff
@@ -160,7 +160,7 @@ analyzeTable dbConn tableName =
 
 loadFixture :: Text -> FilePath -> IO()
 loadFixture dbConn name =
-  void $ readProcess "psql" ["--set", "ON_ERROR_STOP=1", toS dbConn, "-a", "-f", "test/fixtures/" ++ name ++ ".sql"] []
+  void $ readProcess "psql" ["--set", "ON_ERROR_STOP=1", toS dbConn, "-q", "-f", "test/fixtures/" ++ name ++ ".sql"] []
```
steve-chavez: The `hGetContents: invalid argument (invalid byte sequence)` is something that also happened on non-docker environments so it's good we don't have that anymore.

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: "3.6"
```
steve-chavez: Could you put this file under the `test` directory?

PR #1464, opened 2020-03-21
Computed overload

wolfgangwalther:
I noticed in my project that computed columns were not found when two overloaded functions with the same name (and different table as parameter) were created. One of them would be found, while the other would just throw.

I was able to add some tests to show the problem. One of the two `overloaded computed columns are found #N` tests is failing.

I guess it should be possible to use the correct function for the computed column depending on, on which table it is used?

However I was just able to write the tests so far and have not come up with a solution. I wouldn't even know where to start in the source code and I don't know haskell either...


steve-chavez:
@wolfgangwalther Hey there.

In this case the tests are failing because of a missing privilege on `items2`. Tests would be green if you add it here:
https://github.com/PostgREST/postgrest/blob/0f8838623b1a9bc50e795295b5a43b020fed1f4c/test/fixtures/privileges.sql#L124-L125

You can try adding that to confirm. I've also tried `computed_overload` manually and it worked.

Could it be that you're also missing a privilege grant in your project's schema?

wolfgangwalther:
I guess I was a bit too quick to call my test cases finished. Thanks for the hint. When I do that, the test cases indeed turn green.

However that doesn't solve the issue in my project. The only way I can make it work there is by renaming one of the two overloaded functions - it then works as expected.
Without renaming I am getting this error:
`{"hint":null,"details":null,"code":"42703","message":"column pg_source.computed_overload does not exist"}`

I'll have to investigate how to reproduce this in a test case.

wolfgangwalther:
It seems like the problem is only happening when selecting an overloaded computed column on the result of a rpc - see the new test cases. I hope I haven't missed anything this time (other than putting some of those tests in RpcSpec.hs - but I feel it's easier to understand right now, if those tests are all next to each other).

steve-chavez:
I see what you mean now. However I think this is more of a PostgreSQL limitation. This is the query generated for `/rpc/search?id=1&select=computed_overload`:

```sql
WITH
pg_source AS (
  WITH
  pgrst_payload AS (SELECT '{"id": 1}'::json AS json_data),
  pgrst_args AS ( SELECT * FROM json_to_recordset((SELECT val FROM pgrst_body)) AS _("id" bigint) )
  SELECT * FROM "test"."search"("id" := (SELECT "id" FROM pgrst_args LIMIT 1)))
SELECT
  coalesce(json_agg(_postgrest_t), '[]')::character varying AS body
FROM (SELECT "pg_source"."id", "pg_source"."computed_overload" FROM "pg_source"    ) _postgrest_t;
```

If you `drop function computed_overload(test.items2);`, it will work as usual. It seems pg is not able to resolve and identify the function to call. 
I've tried removing the nested WITH and I get the same result. It's weird because `GET /table` generates a similar query and it works. It might be only happening on functions that return table types.

There might be a workaround for this use case, not overloading the function but using `anyelement`: 

```sql
CREATE FUNCTION computed_overload(anyelement) RETURNS boolean
    LANGUAGE sql STABLE
    AS $$ SELECT true $$;
```

If the tables have the same column name, you could even use something like `$1.id` inside there.

Would that work for you?

wolfgangwalther:
Unfortunately, the tables don't have the same column names.

I simplified the query a bit and ran it manually:
```
WITH pg_source AS (
  SELECT * FROM search(1)
)
SELECT pg_source.id, pg_source.computed_overload FROM pg_source;
-- ERROR: column pg_source.computed_overload does not exist
```

Modifying the query a bit reveals more:

```
WITH pg_source AS (
  SELECT * FROM search(1)
)
SELECT pg_source.id, computed_overload(pg_source) FROM pg_source;
-- ERROR: function computed_overload(record) is not unique
```

So basically the subquery - at least in this case - turns the type into an "anonymous" `record`.

However this works:

```
WITH pg_source AS (
  SELECT id, search.computed_overload FROM search(1)
)
SELECT * FROM pg_source;
-- returns proper result
```

I actually did the same tests with tables instead of the search-function - and it gave the same results. So there was **no difference** between tables and table-type-returning-functions regarding this.

Are you sure the query gets build the same way in both cases? Or does the table case maybe have the select on the inner query, ie. directly at the source? Would that be an option for the rpc-case?

_Knowing this, I could probably work around the whole thing in my case by giving one of the two overloaded functions the argument type `record` - this would be the one, that is called on the rpc. The other one is a table. But this seems hacky at best, as it relies on "undefined behaviour" - at least thats what I would call the fact that in the rpc case the row type is "anonymized" to `record`._

steve-chavez:
> I actually did the same tests with tables instead of the search-function - and it gave the same results.

I can also confirm this:

```sql
WITH pg_source AS (
  SELECT * FROM items
)
SELECT pg_source.id, pg_source.computed_overload FROM pg_source;

-- ERROR:  42703: column pg_source.computed_overload does not exist
-- drop function computed_overload(items2); makes it work
```

> Are you sure the query gets build the same way in both cases? Or does the table case maybe have the select on the inner query, ie. directly at the source?

You're right. The table case has does the select directly at the source(the `{selectQuery}` part):

https://github.com/PostgREST/postgrest/blob/a80eb2ff0efdfc14539f6b84f9c90315a5c686f7/src/PostgREST/Statements.hs#L85-L94

Meanwhile calling the proc has it on a result of the CTE.

https://github.com/PostgREST/postgrest/blob/a80eb2ff0efdfc14539f6b84f9c90315a5c686f7/src/PostgREST/Statements.hs#L127-L134

> Would that be an option for the rpc-case?

I'm not sure for now if I can accommodate the query in a way it doesn't break existing functionality. Maybe it can be done, but it would involve some non-trivial refactoring.

wolfgangwalther:
I see - moving the select part inside might be difficult.

There's another way:
```
WITH pg_source AS (
  SELECT pg_record FROM test.search(1) pg_record
)
SELECT (pg_source.pg_record).id, (pg_source.pg_record).computed_overload FROM pg_source;
```

This works as well. This returns the whole record as one column from the inner query - thereby keeping the type. The only change you would have to make is to add the `(pg_source.pg_record).` instead of `pg_source.` to the selected columns as a prefix. That should work with `(pg_source.pg_record).*` as well.

Do you think that would be easier to do?

wolfgangwalther:
So I learned a bit of Haskell and implemented my suggestion above.

The first couple of commits were just cosmetics and stuff:

1. Found some typos/spelling mistakes while studying the source code (aug**u**mentRequestWithJoin, paramsAsMul**it**pleObjects, childs -> children).
2. Change the prefix for `pg_source` and `pg_source_count` to `pgrst_` to align with all the others (`pgrst_body`, `pgrst_payload`, ...)
3. Fixed one of my computed_overload tests
4. Refactored `procReturnsScalar`, which returned `single && scalar` before into `procReturnsScalar` (`scalar`) and `procReturnsSingle` (`single`). Needed to split this up for changes I had to make to `QueryBuilder.requestToCallProcQuery` later on. Just put this into a separate commit to make it easier to see - it's just a refactor, nothing changed here, yet.

The last commit implements the concept of the following query:

```SQL
WITH pgrst_source AS (
  SELECT pgrst_record FROM test.search(1) pgrst_record
)
SELECT (pgrst_source.pgrst_record).id, (pgrst_source.pgrst_record).computed_overload FROM pgrst_source;
```

At first, the changes in `QueryBuilder.requestToCallProcQuery` make sure that the `pg_source` CTE always returns exactly one column named `pgrst_record`, even for Scalars etc. - this is to make sure that nowhere else the difference between accessing `pgrst_source` or `pgrst_source.pgrest_record` has to be made except here. Everywhere else - as soon as it's a procedure call - we need to access `pgrst_source.pgrest_record` now instead of `pgrst_source` directly. `pgrst_record` keeps the row type we need to make overloaded computed columns work.

Now to make sure that in the proc call case each column is accessed as `(pgrst_source.pgrst_record).xyz` instead of `pgrst_source.xyz`, I changed `DbRequestBuilder.rootWithRels`. This took care of changing from table name to `pgrst_source` before - now it's making that `pgrst_source.pgrst_record`. Without the brackets, yet. To make embedding work again, `DbRequestBuilder.getJoinConditions removeSourceCTESchema` was adjusted as well.

Now because I changed `DbRequestBuilder.rootWithRels`, `QueryBuilder.readRequestToQuery` would not only replace the column references, but also the table in the `FROM`-part to something like `FROM pgrst_source.pgrst_record`. This is obviously wrong - so I changed the function to use just `pgrst_source` as soon as a `pgrst_record` is there. But just for the `FROM`-part. The same thing happens in `QueryBuilder.readRequestToCountQuery`.

Now to the brackets - with the changes before, the columns would be accessed as `pgrst_source.pgrst_record.xyz` - but this will throw. To get the brackets I changed `QueryFragment.fromQi` to put brackets around it as soon as the table name is `pgrst_record`, as this will always be a column.

At last I needed to refactor `QueryFragment.asCsvF` a bit to take the `FROM` part for the header as an argument. This way the function can be called as before from `Statements.createWriteStatement` and `Statements.createReadStatement`, but can be changed accordingly in `Statements.callProcStatement` to generate the headers from `pgrst_source.pgrst_record`.

Now all tests are passing - yeah!

src/PostgREST/Types.hs
```diff
@@ -183,8 +183,14 @@ specifiedProcArgs keys proc =
 procReturnsScalar :: ProcDescription -> Bool
 procReturnsScalar proc = case proc of
   ProcDescription{pdReturnType = (Single (Scalar _))} -> True
+  ProcDescription{pdReturnType = (SetOf  (Scalar _))} -> True
```
wolfgangwalther, line 186: Now my Haskell expertise comes to an end.. how can I make those two lines one? ie pattern match on both constructors at once, while keeping the `(Scalar _)`?

wolfgangwalther:
There's a failed test for build-test-9.4 - styler. What exactly does that want to tell me? I'm not getting anything from those error messages.

wolfgangwalther:
Added some tests for one of the write statements (patch here), because those have the `{selectQuery}` outside the CTE as well. But those are passing already, because the columns get selected in the `RETURNING` clause as well. Nothing to do here right now, just to prevent a regression.

wolfgangwalther:
After writing this and understanding more of the code base, I realized this can be done easier. Closed in favor of #1473.

src/PostgREST/Types.hs
steve-chavez: Oh, that's a GHC limitation. Or patterns are not supported. There was a [proposal](https://github.com/osa1/ghc-proposals/blob/or_patterns/proposals/0000-or-patterns.rst#11motivation) some time ago but it didn't get finished. These cases usually need workarounds. Such as guards, multiway if or adding an extra function.

Issue #977, opened 2017-09-15
JWSError JWSInvalidSignature


rvernica:
It does not seem that the JSON Web Token verification works as intended or maybe the documentation is outdated. I am getting a `401` `JWSError JWSInvalidSignature` on the `POST` requests with authorization token.

I followed [Tutorial 0](https://postgrest.com/en/v4.3/tutorials/tut0.html) and [Tutorial 1](https://postgrest.com/en/v4.3/tutorials/tut1.html) word-by-word and this is what I get:

```bash
> cat postgrest.conf
db-uri = "postgres://postgres:mysecretpassword@localhost/postgres"
db-schema = "api"
db-anon-role = "web_anon"
jwt-secret = "secret"
```

```bash
> ./postgrest postgrest.conf
Listening on port 3000
Attempting to connect to the database...
Connection successful
127.0.0.1 - - [15/Sep/2017:14:58:02 -0700] "GET /todos HTTP/1.1" 200 - "" "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:55.0) Gecko/20100101 Firefox/55.0"
127.0.0.1 - - [15/Sep/2017:14:58:23 -0700] "POST /todos HTTP/1.1" 401 - "" "curl/7.53.1"
```

The token I use is identical with the one in the [tutorial](https://postgrest.com/en/v4.3/tutorials/tut1.html#step-3-sign-a-token).
```bash
> export TOKEN="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoidG9kb191c2VyIn0.IF9PKgLZ_XA70Uz5vm0OxrqYTCQdXxBA0Oz4uK8lBqM"
> curl http://localhost:3000/todos -v -X POST \
     -H "Authorization: Bearer $TOKEN"   \
     -H "Content-Type: application/json" \
     -d '{"task": "learn how to auth"}'

Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 3000 (#0)
> POST /todos HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.53.1
> Accept: */*
[...]

PR #1465, opened 2020-03-21
Improve command line help

mrexodia:
After searching for a while I found that there is a hidden requirement of at least being 32 characters in the `jwt-secret`.

It would be best to fail on config parsing if this requirement is not satisfied, but my Haskell is not good enough, so hopefully this documentation update will prevent people from making the same mistake in the future.

https://github.com/PostgREST/postgrest/issues/977


steve-chavez:
Good idea.

The `jwt-secret` config is overloaded now so it's not easy to do validation. 

This will definitely help. Thank you.

Issue #1461, opened 2020-03-13
Large Location header on POST response


steve-chavez:
When there' no PK detected on a relation, and the relation columns have huge amounts of data(`text`/`bytea`) the `Location` header included in a response to a POST request can exceed proxies header limit size(4K, 8K, etc). An example in https://github.com/PostgREST/postgrest/issues/1427#issuecomment-595409551 shows: 

```http
POST /resource

HTTP/1.1 201 Created
Location: /loc_test?port_id=eq.206&name=eq.Loc%20header%20test&...(2k+ chars)
```
Proposed solution: don't reply with a Location header when the created row cannot be identified with a PK or a UNIQUE column(right now we only consider the PK). 

I also think that this goes more in line with [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.3.2), which says:

>The primary resource created by the request is **identified** by either a Location header field in the response or, if no Location field is received, by the effective request URI.

We cannot **identify** the row when there's no pk or unique constraint.

For fixing this now, users can override the `Location` header with a trigger as shown in https://github.com/PostgREST/postgrest/issues/1427#issuecomment-594242147.

neonerd:
I've also encountered this bug when POSTing to a VIEW of a table that has a "json" type column, the redirect URL returned in Location header seemed to be invalid (as it was trying to do a very convoluted eq.{json_value} filtering).

I'd just like to add here that this is a bit of a problem when using the private / public schema approach suggested in documentation, with only VIEWs in public schema exposed by PostgreSQL, as these don't seem to be inheriting PKs from their parent tables.

If we lose the Location: {url} redirect after POST for every VIEW, how do you then propose to do operations that require to further use the created object? (i.e. I create a new user by POSTing to a view and want to use his auto-generated ID, etc.)

PR #1471, opened 2020-04-05
POST, PATCH, DELETE with ?select= and empty body or return=minimal throw

wolfgangwalther:
Migrating from 6.0.2 to 7.0.0 I encountered the following:

Because the default for DELETE requests is now `return=minimal`, some of my requests without return-header but with `?select=` in the query are now failing with:

```JSON
{
  "hint": null,
  "details":null,
  "code": "42703",
  "message": "column pg_source.selected_column does not exist"
}
```

Instead of throwing this should just return nothing, as expected.

Digging deeper, the same is happening on PATCH and POST calls with return=minimal. I also received this error when trying to PATCH with an empty-body - and using select at the same time. All this worked before the update to 7.0.0.

Some of this is related to #1393, because `RETURNING 1` doesn't return the columns anymore, which `?select=` tries to access.

All covered by the test cases so far. I added some additional test cases about the payload while debugging - those are already passing, but I figured to keep them in. Why delete test cases, eeh? ;)

### DIFF
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -115,15 +115,15 @@ mutateRequestToQuery (Delete mainQi logicForest returnings) =
     returningF mainQi returnings
     ]
 
-requestToCallProcQuery :: QualifiedIdentifier -> [PgArg] -> Bool -> Maybe PreferParameters -> SqlQuery
-requestToCallProcQuery qi pgArgs returnsScalar preferParams =
+requestToCallProcQuery :: QualifiedIdentifier -> [PgArg] -> Bool -> Maybe PreferParameters -> [FieldName] -> SqlQuery
+requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
   unwords [
     "WITH",
     argsCTE,
     sourceBody ]
   where
     paramsAsSingleObject    = preferParams == Just SingleObject
-    paramsAsMulitpleObjects = preferParams == Just MultipleObjects
+    paramsAsMultipleObjects = preferParams == Just MultipleObjects
 
     (argsCTE, args)
       | null pgArgs = (ignoredBody, "")
@@ -134,7 +134,7 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams =
             "pgrst_args AS (",
               "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (\a -> " " <> pgaType a) <> ")",
             ")"]
-         , if paramsAsMulitpleObjects
+         , if paramsAsMultipleObjects
              then fmtArgs (\a -> " := pgrst_args." <> pgFmtIdent (pgaName a))
              else fmtArgs (\a -> " := (SELECT " <> pgFmtIdent (pgaName a) <> " FROM pgrst_args LIMIT 1)")
         )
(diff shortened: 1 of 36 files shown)


steve-chavez:
You're right. I can also reproduce the issue. Thanks for catching this up!

More test cases are always good :)

test/Feature/DeleteSpec.hs
```diff
@@ -27,15 +27,25 @@ spec =
           { matchStatus  = 200
           , matchHeaders = ["Content-Range" <:> "*/1"]
           }
+
+      it "ignores ?select= when return=minimal" $
+        request methodDelete "/items?id=eq.3&select=id" [("Prefer", "return=minimal")] ""
```
steve-chavez, line 32: Since now we default to the `return=minimal` behavior, I think these tests shouldn't contain the `Prefer: return=minimal`. Having the header confused me a little bit when reading the tests, I think the header should only be used when testing POST. Also, maybe the test should be called like `ignores ?select= when no return=representation`. Others could follow the same kind of naming.

What do you think?

wolfgangwalther: I tend to approach it the other way around. There were some tests (I think those with reduced privileges on columns), where a certain default setting for `return` was expected. This confused me! ;)

I think in general it would be best if all tests were specific about `return=representation` or `return=minimal` and would **never** assume a default. Because then those tests would all still work when the default should change at any point.

There should maybe be 1 test to actually test the default itself, to make sure that this doesn't change unintendedly.

src/PostgREST/Statements.hs
```diff
@@ -72,6 +72,11 @@ createWriteStatement selectQuery mutateQuery wantSingle isInsert asCsv rep pKeys
     | wantSingle = asJsonSingleF
     | otherwise = asJsonF
 
+  selectF
+    -- prevent using any of the column names in ?select= when no response is returned from the CTE
+    | rep `elem` [None, HeadersOnly] = "SELECT * FROM " <> sourceCTEName
```
steve-chavez, line 77: Hmm.. here I think a `SELECT null` would suffice for preventing the error. With this you wouldn't need the change made to the `QueryBuilder` I think.

test/Feature/DeleteSpec.hs
steve-chavez: I see what you mean, that makes sense. Adding a test for the default would be great.

src/PostgREST/Statements.hs
wolfgangwalther: I think for the HeadersOnly case you still need the columns at this point to put together the location string or am I missing something right now?

And that makes me wonder: Is it actually guaranteed that the PK columns are always in the `RETURNING` part?

wolfgangwalther: Like when I do a `POST return=rep select=non_pk_only`.

I remember the foreign keys for the embedded columns get added to the `RETURNING`, but I am not sure about the PKs.

steve-chavez: > Is it actually guaranteed that the PK columns are always in the RETURNING part?

No, but maybe that should be a different fix, check https://github.com/PostgREST/postgrest/issues/1461.

src/PostgREST/QueryBuilder.hs
```diff
@@ -94,7 +94,9 @@ mutateRequestToQuery (Insert mainQi iCols onConflct putConditions returnings) =
     cols = intercalate ", " $ pgFmtIdent <$> S.toList iCols
 mutateRequestToQuery (Update mainQi uCols logicForest returnings) =
   if S.null uCols
-    then "WITH " <> ignoredBody <> "SELECT null WHERE false" -- if there are no columns we cannot do UPDATE table SET {empty}, it'd be invalid syntax
+    -- if there are no columns we cannot do UPDATE table SET {empty}, it'd be invalid syntax
+    -- selecting an empty resultset from mainQi gives us the column names to prevent errors when using &select=
+    then "WITH " <> ignoredBody <> "SELECT * FROM " <> fromQi mainQi <> " WHERE false"
```
wolfgangwalther, line 99: > [...] With this you wouldn't need the change made to the `QueryBuilder` I think.

Regarding the `QueryBuilder`: We do need that change here anyways for the `PATCH return=rep empty-body` case.

wolfgangwalther:
Improved the test cases as suggested and also realized that with the empty-body PATCH with ?select= overloaded computed columns would throw again. Added a test for that as well and fixed it in `QueryBuilder`.

This change also has the `SELECT NULL` again for the default case, ie if no ?select= has been used.

src/PostgREST/Statements.hs
wolfgangwalther: added some test cases for that at #1475 
it's not throwing, but just not returning a location header at all it seems. So yes, this is a problem.

src/PostgREST/QueryBuilder.hs
```diff
@@ -105,6 +108,10 @@ mutateRequestToQuery (Update mainQi uCols logicForest returnings) =
         ]
   where
     cols = intercalate ", " (pgFmtIdent <> const " = _." <> pgFmtIdent <$> S.toList uCols)
+    empty_body_returned_columns :: SqlFragment
+    empty_body_returned_columns
+      | null returnings = "NULL"
+      | otherwise       = intercalate ", " (pgFmtColumn (QualifiedIdentifier mempty $ qiName mainQi) <$> returnings)
```
steve-chavez, line 114: This fragment is looking similar to `returned_columns` in `requestToCallProcQuery`. Do you think that could be factored in a function inside [QueryFragment.hs](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/Private/QueryFragment.hs)?

steve-chavez: Everything else is looking good. I was seeing if there could be some other way to do the fix but your solution is the simplest one.

steve-chavez: Can be done later. Thanks for the fix!

Issue #1162, opened 2018-08-04
No Location header when primary key is not specified in select=


steve-chavez:
```bash
psql -c 'CREATE TABLE test_bulk(id serial primary key, code text)'

curl -i -X POST -d '{"code":42}' \
-H 'Prefer:return=representation' \
-H "Content-type: application/json" \
"localhost:3000/test_bulk?select=code"

#HTTP/1.1 201 Created
#Transfer-Encoding: chunked
#Date: Sat, 04 Aug 2018 17:58:57 GMT
#Server: postgrest/0.5.0.0 (fa1e92f)
#Content-Type: application/json; charset=utf-8
#Content-Range: */*

#[{"code":"42"}]

curl -i -X POST -d '{"code":42}' \
-H 'Prefer:return=representation' \
-H "Content-type: application/json" \
"localhost:3000/test_bulk?select=id,code"

#HTTP/1.1 201 Created
#Transfer-Encoding: chunked
#Date: Sat, 04 Aug 2018 18:00:44 GMT
#Server: postgrest/0.5.0.0 (fa1e92f)
#Location: /test_bulk?id=eq.2
#Content-Type: application/json; charset=utf-8
#Content-Range: */*

#[{"id":2,"code":"42"}]
```

PR #1475, opened 2020-04-10
add new test cases for location header

wolfgangwalther:
Just some test cases as a head start on #1461 for whoever wants to work on it (might be me, might be somebody else).

see also https://github.com/PostgREST/postgrest/pull/1471#discussion_r405003346


wolfgangwalther:
@steve-chavez can you have a look at the failed style-check. It suggests "to reduce duplication", just because two different test-cases expect the same results and have 3 rows in common, I guess.

How do I handle that?

test/Feature/InsertSpec.hs
```diff
@@ -139,11 +149,11 @@ spec actualPgVersion = do
           }
 
       context "into a table with no pk" $ do
-        it "succeeds with 201 and a link including all fields" $ do
+        it "succeeds with 201 but no location header" $ do
           p <- post "/no_pk" [json| { "a":"foo", "b":"bar" } |]
```
steve-chavez, line 153: @wolfgangwalther I found a similar suggestion before, I just worked around it by renaming the variable. Like:
```suggestion
          resp <- post "/no_pk" [json| { "a":"foo", "b":"bar" } |]
```
And replace the `p` for `resp` in the lines that follow.

hlint is a bit lacking in this regard(see https://github.com/ndmitchell/hlint/issues/1009). So the workaround would be good.

wolfgangwalther: ok, i see - we're going to trick hlint here. I tried swapping two of those lines, let's see whether that works.

wolfgangwalther:
So I added/fixed some more test cases and implemented a part of it: location headers are currently only created for primary key columns. Not for any other columns anymore, but also not for unique columns, in case a pk is not available. That would still be to do, but requires more work, I guess, because the list of unique columns first needs to be made available there.

Also fixed #1162, by always adding the primary key cols to the `RETURNING` part. This will have to be extended with unique columns, once those are used for location headers as well. I wonder whether we should maybe just do `RETURNING *` all the time, that would simplify that a lot and those unneeded columns will still not be selected with `return=minimal` or `select=x,y,z`, because the outer query does a `SELECT x, y, z FROM ..` anyways... - what do you think @steve-chavez?

Still have two tests failing right now. One is expected (unique column fallback), the other one is not (unicode). @steve-chavez could you have a look at that 2nd one? Basically the response is correct, the json comparison is just in wrong order. No matter in which order I put the keys in the `let payload = ...`, they always come out the same way (wrong!). I also tried creating the response string to match by hand, instead of reusing `payload`, but that somehow generated different escape sequences for those unicode characters, so it failed as well. How do I do that properly?

test/Feature/InsertSpec.hs
```diff
@@ -418,8 +412,8 @@ spec actualPgVersion = do
 
     context "with unicode values" $
       it "succeeds and returns usable location header" $ do
-        let payload = [json| { "a":"圍棋", "b":"￥" } |]
-        p <- request methodPost "/no_pk"
+        let payload = [json| { "k":"圍棋", "extra":"￥" } |]
+        p <- request methodPost "/simple_pk"
```
steve-chavez, line 425: I also get the same error, it's weird. But since the test was meant for `no_pk` Location, and we're changing this behavior, I think we can just delete this one. There's another unicode test below that is working.

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -343,7 +345,7 @@ returningCols rr@(Node _ forest) = returnings
     -- However if the "client_id" is present, e.g. mutateRequest to /projects?select=client_id,name,clients(name)
     -- we would get `RETURNING client_id, name, client_id` and then we would produce the "column reference \"client_id\" is ambiguous"
     -- error from PostgreSQL. So we deduplicate with Set:
-    returnings = S.toList . S.fromList $ fldNames ++ (colName <$> fkCols)
+    returnings = S.toList . S.fromList $ fldNames ++ (colName <$> fkCols) ++ pkCols
```
steve-chavez, line 348: I think the pkCols change is good. But we could get lost later, the returningCols function is getting a bit complicated. A comment explaining the pkCols inclusion would be nice.

test/Feature/InsertSpec.hs
steve-chavez: I think it's also possible to specify a `.hlint.yaml` at the root of the project and ignore this suggestion. But not sure how to do it now.

src/PostgREST/Private/QueryFragment.hs
```diff
@@ -87,9 +87,9 @@ asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldNam
 locationF :: [Text] -> SqlFragment
 locationF pKeys = [qc|(
   WITH data AS (SELECT row_to_json(_) AS row FROM {sourceCTEName} AS _ LIMIT 1)
-  SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null'))
+  SELECT array_agg(json_data.key || '=eq.' || json_data.value)
   FROM data CROSS JOIN json_each_text(data.row) AS json_data
-  {("WHERE json_data.key IN ('" <> intercalate "','" pKeys <> "')") `emptyOnFalse` null pKeys}
+  WHERE json_data.key IN ('{intercalate "','" pKeys}')
```
steve-chavez, line 92: More verbose but more clear I think:
```suggestion
  WHERE json_data.key IN ('{fmtPKeys}')
)|]
  where fmtPKeys = intercalate "','" pKeys
```
I tried the `= ANY ARRAY` form but didn't work somehow.

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -316,16 +316,18 @@ mutateRequest schema tName apiRequest cols pkCols readReq = mapLeft errorRespons
...
     combinedLogic = foldr addFilterToLogicForest <$> logic <*> filters
     -- update/delete filters can be only on the root table
     (mutateFilters, logicFilters) = join (***) onlyRoot (iFilters apiRequest, iLogic apiRequest)
     onlyRoot = filter (not . ( "." `isInfixOf` ) . fst)
 
-returningCols :: ReadRequest -> [FieldName]
-returningCols rr@(Node _ forest) = returnings
+returningCols :: ReadRequest -> [FieldName] -> [FieldName]
+returningCols rr@(Node _ forest) pkCols
+  | fldNames == ["*"] = fldNames
```
steve-chavez, line 329: Comment here would also be nice.

steve-chavez:
@wolfgangwalther Regarding unique columns, I also think it'd be better to leave that addition for later.

>  I wonder whether we should maybe just do RETURNING * all the time, that would simplify that a lot and those unneeded columns will still not be selected with return=minimal or select=x,y,z, because the outer query does a SELECT x, y, z FROM .. anyways

Not sure if `RETURNING *` would pass all the tests, some could fail because of column privileges. But if you could make it work, by all means go ahead and I'll give it a review.

test/Feature/InsertSpec.hs
wolfgangwalther: The other test is for `PATCH`, I think, and doesn't test the location header with unicode in it. This one still does, because the changed table now has the unicode column as PK. So the test does have some value.

Since the expected output was the wrong way around (in regards to key order) - I now just changed the actual ouput's key order by adding a `select=extra,k`. Should pass now.

wolfgangwalther:
Column privileges! Yes, that's a very good reason for what we have right now. I don't think those would work with `RETURNING *`, so I won't even try it.

I removed the tests for "location headers from unique columns" for now, so this should be good to merge.

Resolves #1461, #1162

src/PostgREST/DbRequestBuilder.hs
```diff
@@ -316,16 +316,19 @@ mutateRequest schema tName apiRequest cols pkCols readReq = mapLeft errorRespons
...
     -- update/delete filters can be only on the root table
     (mutateFilters, logicFilters) = join (***) onlyRoot (iFilters apiRequest, iLogic apiRequest)
     onlyRoot = filter (not . ( "." `isInfixOf` ) . fst)
 
-returningCols :: ReadRequest -> [FieldName]
-returningCols rr@(Node _ forest) = returnings
+returningCols :: ReadRequest -> [FieldName] -> [FieldName]
+returningCols rr@(Node _ forest) pkCols
+  -- if * is part of the select, we must not add pk or fk columns manually - otherwise those would be selected and output twice
+  | "*" `elem` fldNames = ["*"]
```
steve-chavez, line 330: Nice comment! Now I understand why the `*` is needed.

test/fixtures/schema.sql
```diff
@@ -566,6 +566,12 @@ CREATE TABLE no_pk (
     b character varying
 );
 
+CREATE TABLE no_pk_but_unique (
+    a character varying,
+    b character varying,
+    unique_column character varying DEFAULT NULL UNIQUE
+);
+
```
steve-chavez, line 574: Looks like this table is not being used now, it could be useful for later when the unique Location works. But for now, I think It'd be better to remove it.

test/fixtures/privileges.sql
```diff
@@ -31,6 +31,7 @@ GRANT ALL ON TABLE
     , materialized_view
     , menagerie
     , no_pk
+    , no_pk_but_unique
```
steve-chavez, line 34: Ditto. Unused.

steve-chavez:
@wolfgangwalther Made one last review, check my above comments.

Also, please add the fixes to the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md#fixed) and then I'll merge!

wolfgangwalther:
Technically one of them is a change not a fix, I guess, because if you were to make a really large TEXT column the PK on the table the actual bug in #1461 (exceeding header size limit) would still not be solved. So the "bug" is still there, also a lot less likely with the changed behaviour ;)

steve-chavez:
@wolfgangwalther Thanks!

> because if you were to make a really large TEXT column the PK on the table

I think it's unlikely to have such a big text PK. So I'm calling #1461 fixed :)

PR #1476, opened 2020-04-12
Add a script for running the test suite in a local temporary database.

monacoremo:
This utility script makes it much simpler to run the tests locally - just run:

```bash
test/with_tmp_db stack test
```

No Docker or a host Postgres cluster needed, and you can always be sure that the tests are run on a fresh database. The only dependencies are the Postgres binaries 'initdb' and 'pg_ctl'.

Happy to provide matching testing docs to postgrest-docs if there is interest! This might also make some CI tasks faster/simpler.

### DIFF
--- a/test/with_tmp_db
+++ b/test/with_tmp_db
@@ -54,7 +54,8 @@ trap stop exit
 # Prepare the database for running the tests.
 psql <<EOF
     create extension pgcrypto;
-    alter database $PGDATABASE set request.jwt.claim.id = '-1';
+    alter database $PGDATABASE set request.jwt.claim.id = '-1';# 
+    alter role $PGUSER set default_text_search_config to english;
 EOF
 
 export POSTGREST_TEST_CONNECTION="postgresql://$PGDATABASE?host=$PGDATA&user=$PGUSER"
--- a/test/with_tmp_db
+++ b/test/with_tmp_db
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+usage() {
+    cat << EOF
+
+USAGE: $0 <command>
+
+Runs the given command with the POSTGREST_TEST_CONNECTION environment variable
+set to a temporary database that is ready for running the PostgREST test suite.
+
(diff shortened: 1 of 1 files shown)


steve-chavez:
Really liking this new way of running tests. It's much faster and simpler.

I don't understand why build-test-9.4 is failing though: https://circleci.com/gh/PostgREST/postgrest/4337.

test/with_tmp_db
```diff
@@ -0,0 +1,66 @@
...
+    rm -rf "$tmpdir"
+}
+
+trap stop exit
+
+# Prepare the database for running the tests.
+psql <<EOF
+    create extension pgcrypto;
+    alter database $PGDATABASE set request.jwt.claim.id = '-1';
+EOF
```
steve-chavez, line 58: There is some other stuff that `test/create_test_db` did:

https://github.com/PostgREST/postgrest/blob/be674eb41d958716e9b4fa0fe186208992f73cb8/test/create_test_db#L50-L58

Do you think adding that would be unnecesary? IIRC, the `ALTER ROLE postgrest_test_authenticator SET default_text_search_config TO english; ` was necessary for users that run in another locale. Otherwise tests failed.

monacoremo: I'm not sure about the default_text_search_config setting, I'll add it just in case!

The `LC_` settings should all be taken care of by the `--no-locale` flag given to `initdb`, which [sets everything to `C`](https://www.postgresql.org/docs/12/app-initdb.html), which in turn is [equivalent to `POSIX`](https://www.postgresql.org/docs/12/locale.html).

monacoremo:
> I don't understand why build-test-9.4 is failing though: https://circleci.com/gh/PostgREST/postgrest/4337.

Not sure either... the pull request just adds that one file and shouldn't impact existing tests at all. Seems to work for some other CI jobs though.

steve-chavez:
@monacoremo Try rebasing with master. Let's see if https://github.com/PostgREST/postgrest/pull/1482 fixed this one.

monacoremo:
@steve-chavez that did it!

PR #1478, opened 2020-04-12
Simplify the DbStructure queries

monacoremo:
First suggestion for some cleanup in the DbStructure queries:

* [x] removed no-op lines
* [x] renamed `information_schema.sql_identifier` type to name, and similar cases
* [x] replaced uses of the `information_schema.yes_or_no` (quite archaic! :-) ) with bool 
* [x] broke long lines with many conditions
* [x] removed redundant type casts
* [x] added explanations on some magic numbers
* [x] removed unnecessary quotes

Overall, this should hopefully make the queries easier to read and understand.

I did not change the case or indentation of most of the queries in order to keep the diff a bit smaller.  

--
Before edit:

While working on a PostgREST bindings generator I noticed that the `DbStructure` type contains a lot of redundant info and that the queries for reading the DbStructure from the database are probably more complex than they need to be. This might negatively impact performance and maintainability, but of course changing this part of the project is highly sensitive - so it might not be worth it to change anything here, to be discussed.

An approach to simplifying the structure could be:
1. Simplify the queries in DbStructure.hs as much as possible (should become much shorter, e.g. see the first commit here)
2. Internally to the DbStructure module: Switch to getting the columns directly with the table query, but still return the current DbStructure type for compatibility with the rest of the library
3. Switch the DbStructure type to a less redundant definition - As noted in https://github.com/PostgREST/postgrest/blob/c5245317845d8c0524800b08b3aded150b8eb6a0/src/PostgREST/Types.hs#L113 the `Table` type could hold it's own columns, for example. This would, of course, impact many other modules.

What do you think? Is this a refactor worth pursuing?

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -596,14 +596,7 @@ allPrimaryKeys tabs =
         SELECT
             c.conname::name AS constraint_name,
             nr.nspname::name AS table_schema,
-            r.relname::name AS table_name,
-                CASE c.contype
-                    WHEN 'c' THEN 'CHECK'
-                    WHEN 'f' THEN 'FOREIGN KEY'
-                    WHEN 'p' THEN 'PRIMARY KEY'
-                    WHEN 'u' THEN 'UNIQUE'
-                    ELSE NULL
-                END::text AS constraint_type
+            r.relname::name AS table_name
         FROM pg_namespace nc,
             pg_namespace nr,
             pg_constraint c,
@@ -611,9 +604,10 @@ allPrimaryKeys tabs =
         WHERE
             nc.oid = c.connamespace
             AND nr.oid = r.relnamespace
-            AND c.conrelid = r.oid AND (c.contype <> ALL (ARRAY['t', 'x']))
+            AND c.conrelid = r.oid
             AND r.relkind = 'r'
             AND NOT pg_is_other_temp_schema(nr.oid)
+            AND c.contype = 'p'
     ),
     -- CTE to replace information_schema.key_column_usage to remove owner limit
     kc AS (
@@ -663,7 +657,6 @@ allPrimaryKeys tabs =
     FROM
         tc, kc
     WHERE
-        tc.constraint_type = 'PRIMARY KEY' AND
         kc.table_name = tc.table_name AND
         kc.table_schema = tc.table_schema AND
         kc.constraint_name = tc.constraint_name AND
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -462,52 +462,8 @@ allColumns tabs =
                         END
                     END::information_schema.character_data AS data_type,
                 information_schema._pg_char_max_length(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.cardinal_number AS character_maximum_length,
-                information_schema._pg_char_octet_length(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.cardinal_number AS character_octet_length,
                 information_schema._pg_numeric_precision(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.cardinal_number AS numeric_precision,
-                information_schema._pg_numeric_precision_radix(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.cardinal_number AS numeric_precision_radix,
-                information_schema._pg_numeric_scale(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.cardinal_number AS numeric_scale,
-                information_schema._pg_datetime_precision(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.cardinal_number AS datetime_precision,
-                information_schema._pg_interval_type(information_schema._pg_truetypid(a.*, t.*), information_schema._pg_truetypmod(a.*, t.*))::information_schema.character_data AS interval_type,
-                NULL::integer::information_schema.cardinal_number AS interval_precision,
-                NULL::character varying::information_schema.sql_identifier AS character_set_catalog,
-                NULL::character varying::information_schema.sql_identifier AS character_set_schema,
-                NULL::character varying::information_schema.sql_identifier AS character_set_name,
-                    CASE
-                        WHEN nco.nspname IS NOT NULL THEN current_database()
-                        ELSE NULL::name
-                    END::information_schema.sql_identifier AS collation_catalog,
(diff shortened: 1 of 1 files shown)


steve-chavez:
> What do you think? Is this a refactor worth pursuing?

Yes, I think it's worth it. That module definitely needs some cleaning.

Removing unused parts should be safe as long as tests keep passing.

monacoremo:
Ok :-) I'll take a go at it and submit pull requests step by step!

src/PostgREST/DbStructure.hs
```diff
@@ -621,63 +583,44 @@ allPrimaryKeys tabs =
...
+                    ELSE NULL
+                END::text AS constraint_type
         FROM pg_namespace nc,
             pg_namespace nr,
             pg_constraint c,
             pg_class r
-        WHERE nc.oid = c.connamespace AND nr.oid = r.relnamespace AND c.conrelid = r.oid AND (c.contype <> ALL (ARRAY['t'::"char", 'x'::"char"])) AND r.relkind = 'r'::"char" AND NOT pg_is_other_temp_schema(nr.oid)
-        /*--AND (pg_has_role(r.relowner, 'USAGE'::text) OR has_table_privilege(r.oid, 'INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER'::text) OR has_any_column_privilege(r.oid, 'INSERT, UPDATE, REFERENCES'::text))*/
-        UNION ALL
-        SELECT current_database()::information_schema.sql_identifier AS constraint_catalog,
```
monacoremo, line 652: This part of the union can never be used, as the `contraint_type` is hardcoded to be 'CHECK' and we filter for 'PRIMARY_KEY' in the final query (`tc.constraint_type = 'PRIMARY KEY'`)

src/PostgREST/DbStructure.hs
```diff
@@ -460,77 +470,42 @@ allColumns tabs =
...
+                LEFT JOIN pg_attrdef ad
+                    ON a.attrelid = ad.adrelid AND a.attnum = ad.adnum
+                JOIN (pg_class c JOIN pg_namespace nc ON c.relnamespace = nc.oid)
+                    ON a.attrelid = c.oid
+                JOIN (pg_type t JOIN pg_namespace nt ON t.typnamespace = nt.oid)
+                    ON a.atttypid = t.oid
+                LEFT JOIN (pg_type bt JOIN pg_namespace nbt ON bt.typnamespace = nbt.oid)
+                    ON t.typtype = 'd' AND t.typbasetype = bt.oid
+                LEFT JOIN (pg_collation co JOIN pg_namespace nco ON co.collnamespace = nco.oid)
+                    ON a.attcollation = co.oid AND (nco.nspname <> 'pg_catalog'::name OR co.collname <> 'default'::name)
```
monacoremo, line 501: This section had quite misleading indentation (see the joins in the nested queries) - this should reflect the logic better

src/PostgREST/DbStructure.hs
```diff
@@ -617,69 +591,36 @@ allPrimaryKeys tabs =
...
-            current_database()::information_schema.sql_identifier AS table_catalog,
-            nr.nspname::information_schema.sql_identifier AS table_schema,
-            r.relname::information_schema.sql_identifier AS table_name,
-                CASE c.contype
-                    WHEN 'c'::"char" THEN 'CHECK'::text
-                    WHEN 'f'::"char" THEN 'FOREIGN KEY'::text
-                    WHEN 'p'::"char" THEN 'PRIMARY KEY'::text
-                    WHEN 'u'::"char" THEN 'UNIQUE'::text
-                    ELSE NULL::text
-                END::information_schema.character_data AS constraint_type,
```
monacoremo, line 636: This was only used to filter for 'PRIMARY KEY' in the outside query, so its much simpler to filter directly here in the CTE

steve-chavez: Yes, and since you filtered the results by pk constraint(`c.contype = 'p'`) the other condition for omitting trigger and exclusion constraints is not needed(`(c.contype <> ALL (ARRAY['t', 'x']))` ).

It's much more clear now!

src/PostgREST/DbStructure.hs
steve-chavez: Yes! I can see why it's unnecessary here since we are only looking for pk constraints.

src/PostgREST/DbStructure.hs
```diff
@@ -228,11 +228,22 @@ accessibleTables =
...
+        -- it's possible to insert into the relation.
+        or (exists (
+          select 1
+          from pg_trigger
+          where
+            pg_trigger.tgrelid = c.oid
+            and (pg_trigger.tgtype::integer & 69) = 69)
+            -- The trigger type `tgtype` is a bitmask where 69 corresponds to
+            -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_INSERT
+            -- in the PostgreSQL source code.
```
steve-chavez, line 245: These comments are really nice. I remember looking what the 69 meant but I long forgot it. The bitmask mention will definitely clear things up for anyone that stumbles into this.

steve-chavez:
LGTM! It's always nice to reduce the LOCs and clear things up. 

As you noted, the `information_schema` contains various legacy types, it should be fine to replace them.

For example the `information_schema.cardinal_number` should be equivalent to `integer`(same size):

```
\dT+ information_schema.cardinal_number

Schema            : information_schema                 
Name              : information_schema.cardinal_number 
Internal name     : cardinal_number                    
Size              : 4                                  
Elements          :                                    
Owner             : postgres                           
Access privileges :                                    
Description       :                                    

\dT+ integer

Schema            : pg_catalog                                      
Name              : integer                                         
Internal name     : int4                                            
Size              : 4                                               
Elements          :                                                 
Owner             : postgres                                        
Access privileges :                                                 
Description       : -2 billion to 2 billion integer, 4-byte storage 
```

steve-chavez:
@monacoremo Here's another great thing that this change brought that I'd like to point out.

The new allColumns query is around x3 times faster than the old allColumns query!

I've attached the queries and explain results here: https://gist.github.com/steve-chavez/d95af03333b4c31f126869e2ea119063

- New query: cost=2690.69..2799.27
- Old query: cost=6005.36..6113.94

I've also tested this on a big schema that was shared to me privately on https://github.com/PostgREST/postgrest/issues/945#issuecomment-389635029. The time:

- New query: 235.870 ms
- Old query: 667.906 ms

The explain costs have been pretty much accurate until now.
Queries could be tested in https://github.com/PostgREST/postgrest/blob/master/test/QueryCost.hs for not missing these great results.

monacoremo:
@steve-chavez Wow, that's awesome! Thank you for sharing! 

I didn't expect such an improvement from those changes, as I thought that Postgres would recognize many of the unneeded parts and optimize them out. But apparently not!

I was playing with a much more aggressive refactoring of `allColumns`, as the CTEs are not strictly needed. This would probably be even faster: https://github.com/monacoremo/postgrest/blob/a662a0e628b54c263b5d3afdde21bd3038620c5d/src/PostgREST/DbStructure.hs#L396 - but we would have to take another detailed look at it, as it would slightly change the semantics.

PR #1480, opened 2020-04-13
Add a way to run the test suite for all Postgres versions locally

monacoremo:
With this addition, it's possible to run e.g. `nix-shell --run postgrest-test-all` in the `test` directory in order to run the test suite against all versions of Postgres locally.

### DIFF
--- a/test/default.nix
+++ b/test/default.nix
@@ -1,0 +1,155 @@
+# Run the PostgREST test suite with several versions of Postgres.
+
+# You'll need Nix (https://nixos.org/download.html) to use the scripts defined
+# here. To run the tests for all versions:
+#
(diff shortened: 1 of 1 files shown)


monacoremo:
This might also be useful for CI, e.g. with Github Actions we could use something like:

```yaml
name: "Tests"
on:
  pull_request:
  push:
jobs:
  tests:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Install application
      run: nix-env -iA postgrestTestAll -f test/default.nix
    - name: Run test suite against all Postgres versions
      run: postgrest-test-all
```

(not tested yet, and should probably add some caching for stack)

monacoremo:
Superseded by #1483

PR #1481, opened 2020-04-13
Make tests pass for Postgres 12.2

monacoremo:
When running the test suite against a 12.2 Postgres database, all but two tests passed. The two failures were due to a changed error message. The new error message is added for the newer versions.


monacoremo:
Ah, I see this was already added in the meantime :-)

Issue #756, opened 2016-12-02
Build with Nix


tomberek:
With `nix-env -iA nixpkgs.haskellPackages.postgrest` I get the following:

```
Configuring postgrest-0.3.2.0...
Setup: Encountered missing dependencies:
bytestring-tree-builder ==0.2.7,
hasql ==0.19.12,
hasql-transaction ==0.4.5,
postgresql-binary ==0.9.0.1
builder for ‘/nix/store/kskmdp16snn3bnp9nc8grdnqwln5m5fr-postgrest-0.3.2.0.drv’ failed with exit code 1
error: build of ‘/nix/store/kskmdp16snn3bnp9nc8grdnqwln5m5fr-postgrest-0.3.2.0.drv’ failed
```

Let me know if this is just a simple cabal2nix and a bump, otherwise I'll try to package it up myself.

begriffs:
I've actually never made a Nix package and would appreciate your advice/assistance.

tomberek:
@begriffs https://github.com/NixOS/nixpkgs/pull/20884

That is just a workaround. The proper solution is to update your .cabal in Hackage (and bump the rest of package if possible?) and test on a fresh cabal2nix expression. Then we can remove that cruft and let Nixpkgs Just-Do-Its-Work.

begriffs:
Sounds good. Let's figure that out after the 0.4 release.

witkamp:
We use the PostgREST nix package as well and would like to see the constraint fixed.
Are you expecting to do this after v0.4?

tomberek:
https://github.com/NixOS/nixpkgs/pull/20884

Any help on fixing it or making it more tobust would be appreciated.

On Sat, Dec 31, 2016, 20:49 Theodore Witkamp <notifications@github.com>
wrote:

We use the PostgREST nix package as well and would like to see the
constraint fixed.
Are you expecting to do this after v0.4?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/begriffs/postgrest/issues/756#issuecomment-269889187>,
or mute the thread

begriffs:
I released 0.4 binaries and pushed the cabal file to hackage. Is there something else I need to do inside the cabal file?

PierreR:
I don't know about `darwin` but I have managed to build ` posgrest` on linux with nix.
[...]

PR #1483, opened 2020-04-15
Nix development and build environment

monacoremo:
Work in progress towards #756 

## Tasks

* [x] `nix-build` works, building PostgREST with Nix
* [x] `nix-shell` development environment
  * [x] includes a Nix development environment for PostgREST (`ghc` etc.) 
  * [x] `cabal` integrated with ghc (build, test etc.)
  * [x] `cabal build` works
  * [x] Testing
    * [x] `postgrest-test-spec` implemented, runs the test suite against a temporary db
    * [x] Pass all tests
    * [x] `postgrest-test-spec-all`: implement a script for running the tests against all Postgres versions
    * [x] integrate other test scripts
      * [x] `io-tests.sh`
      * [x] `memory-tests.sh`
  * [x] Styling: implemented `postgrest-style`, automatically formats Haskell and Nix code so far
  * [x] Linting: implemented `postgrest-lint`, uses `hlint` so far
  * [x] Pinned nixpkgs: implemented `nixpkgs-upgrade`, a script for upgrading the pinned version of nixpkgs to the current nixpkgs-unstable
  
## Possibly for later, tdb

* Docs on postgrest-docs
* CI integration
  * [x] Preferences regarding CircleCI, Github Actions, etc.? --> CircleCI
  * Caching with Cachix (tdb if required, as we track nixpkgs-unstable and most packages have cached builds)
* Static: building static binaries directly from Nix  
* Docker: generate Docker images directly from Nix
* Cleanup of older setup?
  * determine how much should be possible to do without Nix (build, run test suite, lint, ...)
  * `test/`: create_test_db, destroy_test_db etc.
  * Makefile
  * CI
  * `docker/`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,38 +88,18 @@ jobs:
 
   style:
     docker:
-      - image: circleci/buildpack-deps:trusty
+      - image: nixos/nix:2.3
     steps:
       - checkout
-      - restore_cache:
-          keys:
-          - v1-stack-style-dependencies-{{ checksum "stack.yaml" }}
       - run:
-          name: install stack
-          command: |
-            curl -L https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-x86_64.tar.gz | tar zx -C /tmp
-            sudo mv /tmp/stack-2.1.3-linux-x86_64/stack /usr/bin
-            stack setup
+          name: Install linters
+          command: nix-env -iA lint style -f default.nix
       - run:
-          name: install linters
-          command: |
-            rm -rf $(stack path --dist-dir) $(stack path --local-install-root)
-            stack install hlint stylish-haskell
-      - save_cache:
-          paths:
-            - "~/.stack"
-            - ".stack-work"
-          key: v1-stack-style-dependencies-{{ checksum "stack.yaml" }}
+          name: Run linter
+          command: postgrest-lint
       - run:
-          name: Add stack tools to $PATH
-          command: |
-            echo "export PATH=/home/circleci/.local/bin:$PATH" >> $BASH_ENV
-      - run:
-          name: run linter
-          command: make lint
-      - run:
-          name: run styler
-          command: make style
+          name: Run styler
+          command: postgrest-style-check
 
   build-test-9.4:
     docker:
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ site
 *~
 *#*
 .#*
+*.swp
+result*
+dist-newstyle
--- a/default.nix
+++ b/default.nix
@@ -26,22 +26,27 @@ let
     import pinnedPkgs { inherit overlays; };
 
   postgresqlVersions =
-    [
-      pkgs.postgresql_9_5
-      pkgs.postgresql_9_6
-      pkgs.postgresql_10
-      pkgs.postgresql_11
-      pkgs.postgresql_12
-    ];
+    map postgresqlWithPackages
+      [
+        pkgs.postgresql_12
+        pkgs.postgresql_11
+        pkgs.postgresql_10
+        pkgs.postgresql_9_6
+        pkgs.postgresql_9_5
+      ];
+
+  postgresqlWithPackages =
+    postgresql:
+      postgresql.withPackages (p: [ p.pgjwt ]);
 in
 rec {
-  inherit pkgs;
+  inherit pkgs pinnedPkgs;
 
   postgrest =
     pkgs.haskellPackages.callCabal2nixWithOptions name src "--no-check" {};
 
   tests =
-    pkgs.callPackage nix/tests.nix { inherit postgresqlVersions; };
+    pkgs.callPackage nix/tests.nix { inherit postgresqlVersions pinnedPkgs; };
 
   style =
     pkgs.callPackage nix/style.nix {};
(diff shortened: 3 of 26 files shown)


monacoremo:
@steve-chavez was there a special reason for the lower bound on https://github.com/PostgREST/postgrest/blob/f6ce93f2c83178327ae42e5da42dd92a3f3742e2/test/QueryCost.hs#L53 ? This repeatedly broke the tests with `test/QueryCost.hs:53:13: 1) QueryCost, call proc query, params=multiple-objects, should not exceed cost when calling setof composite proc predicate failed on: Nothing`

steve-chavez:
@monacoremo The problem is that `(Nothing < Just 2000) == True`. I left the lower bound as a way to check that `cost` is not giving `Nothing`, because `(Nothing > Just 2000) == False`. This definitely needs refactoring.

So as it is, the QueryCost module is not really working because `cost` is `Nothing`.

steve-chavez:
Regarding CI, I think we should stick with CircleCI. I remember begriffs got us a deal with them for extra resources usage. IIRC I think they increased timeout for builds and enabled building on OSX(we're not taking advantage of this).

Integrating static binary building on CI would be nice. But I think we should focus first on just building locally with Nix and leave CI changes for another PR(this would also help me in reviewing). The CI infra is a bit fragile and we depend on that for releases.

Cleanup of older setup would be good. The `create/destroy_test_db` always felt a bit convoluted. Though we should make sure that `with_tmp_db` works when testing on docker too.

monacoremo:
@steve-chavez that makes sense - looking at the db logs, it was a race condition where cabal ran spec and spec-querycost in parallel and https://github.com/PostgREST/postgrest/blob/f6ce93f2c83178327ae42e5da42dd92a3f3742e2/test/QueryCost.hs#L25 does not hold - added a quick fix on the cabal side, but should probably be fixed properly at some point

monacoremo:
> leave CI changes for another PR

@steve-chavez Absolutely, we should do this step by step - maybe even in a separate branch for now? Changing the stack resolver etc. is relatively intrusive...

robx:
Oh, I missed that you were already working on updating stackage. #1484, 1485 are redundant with this one. But let's at least see if they pass CI.

monacoremo:
> Oh, I missed that you were already working on updating stackage. #1484, #1485 are redundant with this one. But let's at least see if they pass CI.

@robx you did a much cleaner job with the dependencies! Let's get your work merged and I'll rebase this pull request.

monacoremo:
@steve-chavez the style tests previously failed probably because stack ran out of memory when installing hlint and stylish-haskell - The latest commit makes the install more efficient by installing them directly from Nix. Happy to split that into a separate PR if you prefer.

steve-chavez:
@monacoremo I'll give your PR a review now!

Btw, in case anyone stumbles with this error:
```
Downloading lts-15.8 build plan ...
RedownloadInvalidResponse Request {
  host                 = "raw.githubusercontent.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/fpco/lts-haskell/master//lts-15.8.yaml"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 "/home/steve-chavez/.stack/build-plan/lts-15.8.yaml" (Response {responseStatus = Status {statusCode = 404, statusMessage = "Not Found"}
, responseVersion = HTTP/1.1, responseHeaders = [("Connection","keep-alive"),("Content-Length","14"),("Content-Type","text/plain; charse
t=utf-8"),("Content-Security-Policy","default-src 'none'; style-src 'unsafe-inline'; sandbox"),("Strict-Transport-Security","max-age=315
36000"),("X-Content-Type-Options","nosniff"),("X-Frame-Options","deny"),("X-XSS-Protection","1; mode=block"),("Via","1.1 varnish (Varnis
h/6.0)"),("X-GitHub-Request-Id","096A:7607:2850F0:2B45D6:5E9B3127"),("Accept-Ranges","bytes"),("Date","Sat, 18 Apr 2020 16:56:07 GMT"),(
"Via","1.1 varnish"),("X-Served-By","cache-gru17141-GRU"),("X-Cache","HIT, MISS"),("X-Cache-Hits","2, 0"),("X-Timer","S1587228968.876053
,VS0,VE119"),("Vary","Authorization,Accept-Encoding"),("Access-Control-Allow-Origin","*"),("X-Fastly-Request-ID","fb7848c985ee0cb57ce6c6
e113b735c611e7372a"),("Expires","Sat, 18 Apr 2020 17:01:07 GMT"),("Source-Age","0")], responseBody = (), responseCookieJar = CJ {expose
= []}, responseClose' = ResponseClose})
``` 
The fix is to upgrade `stack`. See https://stackoverflow.com/a/57521914/4692662.

steve-chavez:
> @steve-chavez the style tests previously failed probably because stack ran out of memory when installing hlint and stylish-haskell - The latest commit makes the install more efficient by installing them directly from Nix. Happy to split that into a separate PR if you prefer.

@monacoremo Please do, would be easier for me to review changes to CI in another PR(I do note that the build is faster, but I'll have to give the change more thought). It's weird why CircleCI is not picking the cache on that `style` job though. The `stack.yaml` hasn't changed.

I'll see if the style CI job has a fix. If not, I can merge without fixing that here. We'll just let it fail for now.

monacoremo:
Will do - I'll submit the changes in smaller PRs. This here was good to figure out an overall structure that works, but it combined too many different concerns in one go.

#1489 should contain the bare necessities and I'll suggest additional improvements from there. I'll keep this PR as is for reference.

PR #1487, opened 2020-04-16
circleci: Split out dedicated style check and build caching jobs

robx:
Currently, the postgres 9.4 job is also responsible for

- running style checks
- caching the dependencies

This has a couple downsides. E.g. it's not apparent at a glance that your PR tests failed due to linter issues, and you need to wait quite a while for these. Secondly, on a dependency update, most of the test jobs do the full compile work in parallel because the cache doesn't exists, and again it's hard to distinguish between postgres version specific failures and a general postgrest build failure.

This PR aims to address those by

- splitting out a style checking target that doesn't even build
- splitting out a dedicated build and cache target, and having the various postgres-version-specific jobs depend on this

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,9 @@ jobs:
       - image: circleci/buildpack-deps:trusty
     steps:
       - checkout
+      - restore_cache:
+          keys:
+          - v1-stack-style-dependencies-{{ checksum "stack.yaml" }}
       - run:
           name: install stack
           command: |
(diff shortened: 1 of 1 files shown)


steve-chavez:
This one is really good. It will speed up the CI run!

.circleci/config.yml
```diff
@@ -339,26 +388,44 @@ workflows:
...
+      - style:
+          filters:
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*/
       - build-test-9.4:
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
+          requires:
+            - build-cache
```
steve-chavez, line 404: Do you think the pg jobs should also depended on the `style` job? Since they run in parallel maybe `style` could take a while. Maybe it'll be better for contributors to get a quick feedback if somehow they forgot to run the styler locally.

steve-chavez: We'll see about this later. PR it's looking really good as it is.

steve-chavez: Actually the styler runs quickly. So nevermind about this.

PR #1488, opened 2020-04-17
Fix and improvements to with_tmp_db

monacoremo:


test/with_tmp_db
```diff
@@ -1,67 +1,111 @@
...
 set to a temporary database that is ready for running the PostgREST test suite.
 
-You'll need to have the Postgres binaries 'initdb' and 'pg_ctl' on your PATH.
+You'll need to have the Postgres binaries 'initdb', 'pg_ctl' and 'psql' on your
+PATH.
 
 Example:
 
-    $0 stack test
+    $0 cabal v2-test
```
steve-chavez, line 19: For users that won't be able to use nix-shell, I think the `stack test` example should be kept. It's working really good with stack.

steve-chavez: Maybe something like:
```suggestion
    $0 stack v2-test
    
    ## Or for when inside nix-shell
    $0 cabal v2-test
```

monacoremo: Good idea, thanks for the fix!

test/with_tmp_db
```diff
@@ -1,67 +1,114 @@
...
+log "Loading fixtures..."
+psql -v ON_ERROR_STOP=1 >> "$setuplog" << EOF
+  create extension pgcrypto;
+  alter database $PGDATABASE set request.jwt.claim.id = '-1';
+  alter role $PGUSER set default_text_search_config to english;
 
-export POSTGREST_TEST_CONNECTION="postgresql://$PGDATABASE?host=$PGDATA&user=$PGUSER"
+  \i $fixturesdir/database.sql
+  \i $fixturesdir/roles.sql
+  \i $fixturesdir/schema.sql
```
steve-chavez, line 103: Why the fixtures run in the script now? The haskell test helper already runs them:
https://github.com/PostgREST/postgrest/blob/master/test/SpecHelper.hs#L148-L155

Won't this be redundant?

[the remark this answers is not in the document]
monacoremo: Yes, good point - I added this based on playing around with getting the tests running reliably from Nix (upcoming PR :-) ). I encountered the following issues:

* If you forget to specify --j=1 (only one job) when running the tests with stack or cabal, sometimes the QueryCost tests run before the main tests had the chance to run `setupDb`, and then [this assumption](https://github.com/PostgREST/postgrest/blob/24db4a1e256860c66e91d60999e979664f4a0b68/test/QueryCost.hs#L25) does not hold and you get a failure.
* The io-tests and memory-tests assume that the Haskell test suite ran before and ran `setupDb`; so if you want to run them independently, you currently need to set up the fixtures by hand (like here currently: https://github.com/PostgREST/postgrest/blob/24db4a1e256860c66e91d60999e979664f4a0b68/.circleci/config.yml#L345 )

Loading the fixtures fixes those problems and is basically free time-wise. A whole with_tmp_db run takes less than half a second:

```bash
remo@xps> time test/with_tmp_db psql -c 'select 1;'                                                                                                                                             ~/Projects/postgrest
 ?column? 
----------
        1
(1 row)

test/with_tmp_db psql -c 'select 1;'  0,32s user 0,16s system 58% cpu 0,841 total
remo@xps>   
```

To avoid loading the fixtures twice, I think the cleanest solution for now would be to also load them in `create_test_db` and to remove `setupDb` on the Haskell side, as well as the manual setup for memory-tests in CircleCI. Another PR?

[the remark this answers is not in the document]
steve-chavez: Very nice explanation. You're right! I forgot the fixtures also ran in CI.

> To avoid loading the fixtures twice, I think the cleanest solution for now would be to also load them in create_test_db and to remove setupDb on the Haskell side, as well as the manual setup for memory-tests in CircleCI. Another PR?

Yes, I would agree with that. Then the haskell db helper would only load the fixture for data([resetDb](https://github.com/PostgREST/postgrest/blob/master/test/SpecHelper.hs#L157-L158)).

PR #1489, opened 2020-04-19
Basics for building PostgREST with Nix

monacoremo:
This makes it possible to run `nix-build` to build a PostgREST binary. With `nix-shell`, you are dropped into a new shell with the correct GHC and Cabal version for building PostgREST interactively with cached intermediate results.

The version of nixpkgs is pinned to a nixpkgs-unstable version where all dependencies of PostgREST have recently been unbroken - no overrides required anymore! A small utility to upgrade the pinned version is also included - run `nixpkgs-upgrade > nix/nixpkgs-version.nix` in `nix-shell` to upgrade.

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -1,0 +1,77 @@
+# Nix development and build environment
+
+With Nix it's possible to quickly and reliably recreate the full environments
+for developing, testing and building PostgREST.
+
+## Getting started with Nix
+
+You'll need to [get Nix](https://nixos.org/download.html). The installer will
+create your Nix store in the `/nix/` directory, where all build artifacts and
+their dependencies will be stored. It will also link the Nix executables like
+`nix-env`, `nix-build` and `nix-shell` into your PATH. Nix will manage all
+other PostgREST dependencies from here on out. To clean up older build
+artifacts from the `/nix/store`, you can run `nix-collect-garbage`.
+
+## Building PostgREST
+
+To build PostgREST from your local checkout of the repository, run:
+
+```bash
+nix-build --attr postgrest
+
+```
+
+This will create a `result` directory that contains the PostgREST binary at
+`result/bin/postgrest`. The `--attr` parameter (or short: `-A`) tells Nix to
+build the `postgrest` attribute from the Nix expression it finds in our
+`default.nix` (see below for details). Nix will take care of getting the right
+GHC version and all the build dependencies.
+
+## Developing
+
+A development environment for PostgREST is available with `nix-shell`. The
+following command will put you into a new shell that has GHC and Cabal on the
+PATH:
+
+```bash
+nix-shell
+
+```
+
+Within `nix-shell`, you can run Cabal commands as usual.
+
+## Tour
+
+The following is not required for working on PostgREST with Nix, but it will
+give you some more background and details on how it works.
+
+### `default.nix`
+
+[`default.nix`](../default.nix) is our 'respository expression' that pulls all
+the pieces that we define with Nix together. It returns a set (like a dict in
+other programming languages), where each attribute is a derivation that Nix
+knows how to build, like the `postgrest` attribute from earlier.
+
+Internally, our `default.nix` uses the `pkgs.callPackage` function to import
+the modules that we defined in the `nix` directory. It automatically passes the
+arguments those modules require if they are available in `pkgs` (this means
+that `pkgs` is defined in terms of itself, better not to think too much about
+that).
+
+We also use `default.nix` to load our pinned version of the `nixpkgs`
+repository. This set of packages will always be the same, independently from
+where or when you use it. The pinned version can be upgraded with the small
+`nixpkgs-upgrade` utility. Running `nixpkgs-upgrade > nix/nixpkgs-version.nix`
+in `nix-shell` will upgrade the pinned version to the latest `nixpkgs-unstable`
+version.
+
+### `shell.nix`
+
+[`shell.nix`](../shell.nix) defines an environment in which PostgREST can be
+built and developed. It extends the build enviroment from our `postgrest`
+attribute with useful utilities that will be put on the PATH in `nix-shell`.
+
+### `nix/overlays`
+
+Our overlays to the Nix package set are defined here. They allow us to tweak our
+`pkgs` in `default.nix` by adding new packages or overriding existing ones.
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ site
 *~
 *#*
 .#*
+*.swp
+result*
+dist-newstyle
--- a/default.nix
+++ b/default.nix
@@ -2,9 +2,14 @@ let
   name =
     "postgrest";
 
-  # PostgREST source files, filtered based on the rules in the .gitignore files.
+  # PostgREST source files, filtered based on the rules in the .gitignore files
+  # and file extensions. We want to include as litte as possible, as the files
+  # added here will increase the space used in the Nix store and trigger the
+  # build of new Nix derivations when changed.
   src =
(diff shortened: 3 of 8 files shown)


robx:
I don't understand why the style test on this PR didn't find the cache, while a cache was available under the same key on #1490. Perhaps the cache is not shared among all circleci workers?

Anyway I hope #1490 might help with this out-of-memory problem, but it's pretty hard to be sure.

monacoremo:
@robx yes, that's really weird! CircleCI is looking for exactly the same cache key `v1-stack-style-dependencies-KUtJT8...` in both cases, but is somehow not finding it for my pull request...

monacoremo:
All green! That did it, thank you @robx ! 

Somehow the builds from this pull requests seem to access a different cache, this run re-uploaded the cache to the key `KUtJT8...` that should already exist. @steve-chavez maybe that has something to do with the extra resources that CircleCI provides to this project? Maybe those are not accessible for builds triggered by 'outside' PRs?

steve-chavez:
@monacoremo I just triggered the style job and now it picked up the cache. Maybe it didn't restore the cache before because the style job in this PR started in parallel with the style job in master. 

> maybe that has something to do with the extra resources that CircleCI provides to this project? Maybe those are not accessible for builds triggered by 'outside' PRs?

Hmm.. I don't think so. The jobs in PRs have been restoring caches normally until now(the other pg jobs in this PR for example).

nix/README.md
```diff
@@ -0,0 +1,77 @@
...
+`nix-env`, `nix-build` and `nix-shell` into your PATH. Nix will manage all
+other PostgREST dependencies from here on out. To clean up older build
+artifacts from the `/nix/store`, you can run `nix-collect-garbage`.
+
+## Building PostgREST
+
+To build PostgREST from your local checkout of the repository, run:
+
+```bash
+nix-build --attr postgrest
```
steve-chavez, line 20: Hmm.. somehow this is also building Haddock docs(build take longer to complete). Is that intended?

```
[17 of 17] Compiling PostgREST.App    ( src/PostgREST/App.hs, dist/build/PostgREST/App.p_o )
Preprocessing executable 'postgrest' for postgrest-7.0.0..
Building executable 'postgrest' for postgrest-7.0.0..
[1 of 2] Compiling UnixSocket       ( main/UnixSocket.hs, dist/build/postgrest/postgrest-tmp/UnixSocket.o )
[2 of 2] Compiling Main             ( main/Main.hs, dist/build/postgrest/postgrest-tmp/Main.o )
Linking dist/build/postgrest/postgrest ...
haddockPhase
Preprocessing library for postgrest-7.0.0..
Running Haddock on library for postgrest-7.0.0..
Warning: The documentation for the following packages are not installed. No
links will be generated to these packages: mtl-compat-0.2.2, vault-0.3.1.4
Warning: --source-* options are ignored when --hyperlinked-source is enabled.

Haddock coverage:
Warning: 'limit' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
Warning: 'offset' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
   9% (  1 / 11) in 'PostgREST.RangeQuery'
  Missing documentation for:
    rangeParse (src/PostgREST/RangeQuery.hs:33)
    rangeRequested (src/PostgREST/RangeQuery.hs:45)
    rangeLimit (src/PostgREST/RangeQuery.hs:54)
    rangeOffset (src/PostgREST/RangeQuery.hs:60)
    restrictRange (src/PostgREST/RangeQuery.hs:48)
    rangeGeq (src/PostgREST/RangeQuery.hs:66)
    allRange (src/PostgREST/RangeQuery.hs:70)
    NonnegRange (src/PostgREST/RangeQuery.hs:31)
    rangeStatusHeader (src/PostgREST/RangeQuery.hs:77)
    contentRangeH (src/PostgREST/RangeQuery.hs:92)
```

nix/README.md
```diff
@@ -0,0 +1,77 @@
...
+A development environment for PostgREST is available with `nix-shell`. The
+following command will put you into a new shell that has GHC and Cabal on the
+PATH:
+
+```bash
+nix-shell
+
+```
+
+Within `nix-shell`, you can run Cabal commands as usual.
```
steve-chavez, line 41: Is only cabal meant to work? I'm trying building with stack(inside nix-shell) and I get:
```
stack build --nix
error: attribute 'ghc883' missing, at (string):1:74
```

nix/README.md
monacoremo: We could turn that off by adding `dontHaddock` to the `dontCheck` line in [default.nix](https://github.com/PostgREST/postgrest/pull/1489/files#diff-5712e736e0de6ba170577f8472c398e9R35) if we wanted (those helpers are defined in [`haskell-modules/lib.nix`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/lib.nix)).

But I would leave that on for now, as that's the standard behavior of Haskell builds in Nix.

nix/README.md
monacoremo: For now yes - the stack `--nix` integration is a bit more tricky to get right, Cabal seems to be the recommended way as far as I could see. Stack still works as before, both within and without nix-shell, if we don't use '--nix'. You can also try adding `pkgs.stack` to `shell.nix` to get the updated/compatible version.

nix/README.md
steve-chavez: Sorry, still a Nix newbie, how do I add the `dontHaddock`?

```
    pkgs.haskell.lib.dontCheck drv;
```

nix/README.md
steve-chavez: Oh, I see. I'm fine with using cabal inside nix-shell. However, that would mean these [ghc-options defined in stack.yaml](https://github.com/PostgREST/postgrest/blob/master/stack.yaml#L3-L10) won't be used by cabal right? Shouldn't those be added?

nix/README.md
monacoremo: `pkgs.haskell.lib.dontHaddock (pkgs.haskell.lib.dontCheck drv);` should work! 

Or you could also write:

`with pkgs.haskell.lib; dontHaddock (dontCheck drv);` - which is shorter, but it's not so clear anymore from what scope those functions are coming from...

nix/README.md
monacoremo: Yes, that's true! I'll look into adding those.

nix/README.md
steve-chavez: Thanks! It's working good now. I'll save that one for myself. I'm fine with keeping the standard behavior.

nix/README.md
monacoremo: The cleanest way is to add the flags to the .cabal file directly, I think. See https://github.com/PostgREST/postgrest/pull/1489/commits/6f461f4c522370eaf259a00042aaaf9451691f5b for my suggestion.

We could also override the flags directly through Nix. For example, in `default.nix` we'd need to add something like:

```nix
  postgrest =
    pkgs.lib.overrideDerivation drv (
      base: {
        configureFlags =
          base.configureFlags ++ [
            "--ghc-option=-O2"
            "--ghc-option=-Werror"
            "--ghc-option=-Wall"
            ...
          ];
      }
    );
```

(where `overrideDerivation` is a function that takes a derivation and a function that returns a modified derivation)

This seems less maintainable and has the downside that other tools that process Hackage/Stackage would not pick the flags up. E.g. Nixpkgs only looks at the .cabal file of each package and essentially runs 'cabal2nix' on it.

steve-chavez:
I think this is good to merge now! Looking forward to see another PR with how to run the tests with nix-shell.

PR #1490, opened 2020-04-19
circleci: Build linters with memory-saving ghc-options.

robx:
Previously, the linters would use the cached build artefacts from
building postgrest with -fno-spec-constr via stack.yaml. Potentially
passing it to the linter build explicitly will fix recent out-of-memory
issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,9 @@ jobs:
           name: install linters
           command: |
             rm -rf $(stack path --dist-dir) $(stack path --local-install-root)
-            stack install hlint stylish-haskell
+            # -fno-spec-constr may help with out-of-memory issues compiling
+            # haskell-src-exts. Compare stack.yaml.
+            stack install --ghc-options=-fno-spec-constr hlint stylish-haskell
       - save_cache:
           paths:
             - "~/.stack"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -104,7 +104,9 @@ jobs:
           name: install linters
           command: |
             rm -rf $(stack path --dist-dir) $(stack path --local-install-root)
```
robx, line 106: I wonder what's the reason for this line, and whether it affects caching.

robx:
This is meant to address the style build issue as seen in #1489 . 

It seems the trouble was introduced with #1487, but the only explanation I can see for that is that haskell-src-exts is now built unaware of the ghc-options in stack.yaml, which would have (maybe?) been in effect while building it as a postgrest dependency via stack.yaml.

.circleci/config.yml
steve-chavez: You know, I'm not sure either(I just kept it from an old CI config). We can try removing it.
```suggestion
```

steve-chavez: (Commited the change through github UI)

steve-chavez:
The `-fno-spec-constr` change makes sense. Let's see if this fixes #1489 CI run.

PR #1492, opened 2020-04-21
 Run tests against several PostgreSQL versions using Nix

monacoremo:
Now for some Nix magic :-)

This makes it possible to run the Haskell test suite against all supported PostgreSQL versions, locally and without Docker. Simply run `postgrest-test-spec-all` in nix-shell!

This is on top of #1488, which introduces several improvements to the `with_tmp_db` script.

It's not yet possible to install the test scripts with `nix-env -iA tests` and then run them outside nix-shell directly (e.g. `postgrest-test-spec` instead of `nix-shell --run postgrest-test-spec`), as running the tests currently depends on the build environment provided in our `shell.nix`. This would be nice to have for CI, for example, but I think the PR is already useful in this state.

PR #1493, opened 2020-04-21
Integrate stack in the nix-shell environment

monacoremo:


steve-chavez:
Not sure why the appveyor run was failing [before](https://ci.appveyor.com/project/postgrest/postgrest/builds/32342143)(I think the cached build got corrupted) but it seems it's working good now. 

I've tried running the test suite inside nix-shell, but I got:

```bash
./test/with_tmp_db stack test

Stack has not been tested with GHC versions above 8.6, and using 8.8.3, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.0.1.0 was found, this may fail
postgrest> test (suite: spec)

psql:test/fixtures/database.sql:2: ERROR:  42501: must be owner of schema test
LOCATION:  aclcheck_error, aclchk.c:3621
```
Could this because the haskell test setup also runs the fixtures?

steve-chavez:
I think we can discuss the above in https://github.com/PostgREST/postgrest/pull/1492. The change here is minimal and it's good to merge. `stack build` is working good.

monacoremo:
> I think we can discuss the above in #1492. The change here is minimal and it's good to merge. `stack build` is working good.

@steve-chavez figured it out, it's stack trying to be helpful and enabling `--pure` by default. This drops our POSTGREST_TEST_CONNECTION, which causes spec to try the local Postgres server by default and fail. The solution is to run `test/with_tmp_db stack --nix --no-nix-pure test` in nix-shell, which works for me. 

We probably should note in the README that cabal is recommended over stack in nix-shell, as it picks up all the Haskell dependencies that are pinned and built by Nix and usually only ever rebuilds PostgREST itself (unless you do something crazy like --enable-profiling, which also impacts all dependencies).

PR #1495, opened 2020-04-22
Basic CI integration for Nix builds

monacoremo:
This would require that a cachix binary cache named 'postgrest' exists and that the CACHIX_SIGNING_KEY is available in the CircleCI environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -406,7 +406,7 @@ jobs:
             # from a fork. CircleCI will not make the CACHIX_SIGNING_KEY
             # available to those builds, which is required for pushing to the
             # cache.
-            if [ -z "${CI_PULL_REQUEST:-""}" ] || [ "${CIRCLE_PR_USERNAME:-""}" != "PostgREST" ]; then
+            if [ -z "${CI_PULL_REQUEST:-""}" ] && [ "${CIRCLE_PR_USERNAME:-""}" != "PostgREST" ]; then
               echo "Building and caching all derivations..."
               nix-build | cachix push postgrest
             else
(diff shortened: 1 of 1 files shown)


steve-chavez:
@monacoremo Done!

I did `cachix generate-keypair postgrest` and that showed me the `export CACHIX_SIGNING_KEY=....`. I added that env var to CircleCI settings.

Let me know if it works as expected.

monacoremo:
Tha ld


Thanks @steve-chavez, looking good! Other than the linter compile fail again...

Whether `cachix push` works, we'll only see in master, as it's disabled for PRs by design for security - but everything should be set up for it.

steve-chavez:
@monacoremo Should the build-nix CI job fail in PRs? Right now it's failing with: `cachix: FatalError {fatalErrorMessage = "You need to: export CACHIX_SIGNING_KEY=XXX"}`. See https://circleci.com/gh/PostgREST/postgrest/5008.

Also, looks like it's also building haddock docs. Maybe that should be disabled for making the build faster?

monacoremo:
@steve-chavez they should not! Now the logic for pushing should be actually fixed and documented.

Currently the CI job is set up to build everything that we define in default.nix, so potentially later also static builds etc. I think we should do this at some point in CI, to make sure that everything builds properly. With cachix that should also be reasonably quick.

Maybe a compromise would be to build `-A postgrest` (the dynamic executable without anything else) first in a separate step, and then everything else in a second step? That would allow the job to fail fast, but still build everything.

steve-chavez:
LGTM!

> Currently the CI job is set up to build everything that we define in default.nix, so potentially later also static builds etc. I think we should do this at some point in CI, to make sure that everything builds properly. With cachix that should also be reasonably quick.

Oh, I forgot cachix would speed things up. I think doing the whole build is good. But it's good to merge as it is.

PR #1496, opened 2020-04-23
Update to protolude 0.3.0

robx:
The good part is that protolude 0.3.0 builds with GHC 8.10.

The bad part is that this change is a bit painful:
- the default `toS` has changed to no longer convert to and from ByteString
- similarly, `show` no longer outputs ByteString

The changes here are pretty much minimal to keep things compiling; I didn't
see a nice way to work with the new ConvertText class, even though `toUtf8`
seems like it might help if used besides `toS` at just the right spots.

Compare https://github.com/commercialhaskell/stackage/issues/5319.


test/SpecHelper.hs
```diff
@@ -24,7 +24,8 @@ import Text.Heredoc
 
 import PostgREST.Config (AppConfig (..))
 import PostgREST.Types  (JSPathExp (..))
-import Protolude
+import Protolude        hiding (toS)
+import Protolude.Conv   (toS)
```
steve-chavez, line 28: So `Protolude.Conv.toS` is the legacy one right? I'll keep this in mind when working with the new Protolude.

If we later need to add utilitarian functions to workaround breaking changes, we can include them in https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/Private/Common.hs.

steve-chavez:
LGTM :+1: 

Btw, really nice how the stackage issues warns about breaking changes.

test/SpecHelper.hs
robx: Right. It turns out that the legacy `toS` would throw exceptions when trying to decode broken UTF8 bytestrings. That risk is still there. Though unfortunately, the change also means that the non-legacy `toS` can't do the safe conversions between bytestrings or to bytestring. (See also https://github.com/protolude/protolude/issues/117.)

PR #1497, opened 2020-04-23
Disable nix pure by default in stack.yaml

monacoremo:
Fixes the issue noted in https://github.com/PostgREST/postgrest/pull/1493


steve-chavez:
`stack test` works now! 

I always used to run it like: `stack test --nix --no-nix-pure`(forgot about it in https://github.com/PostgREST/postgrest/pull/1493#issuecomment-617976063), but this is much better.

Issue #1227, opened 2019-01-14
FreeBSD port


steve-chavez:
Doing manual releases gets tedious quickly(also my FreeBSD disk partition got damaged recently, cannot boot into it :crying_cat_face:). 

We could try to automate FreeBSD releases with [Cirrus CI](https://cirrus-ci.org/), which recently announced support for FreeBSD https://forums.freebsd.org/threads/cirrus-ci-free-freebsd-ci-testing-for-open-source-projects.68560/.

steve-chavez:
Made some progress on this at https://github.com/steve-chavez/postgrest/blob/cirrus-2/.cirrus.yml

Unfortunately the updated GHC broke the build again. I'm leaving the link here in case someone wants to pick it up.

steve-chavez:
It would be even better if someone can help creating a FreeBSD port. There was an [initiative](https://twitter.com/postgrest_org/status/1189898706826473472) but it didn't got finished.

dwagin:
Main problem FreeBSD ports don't support building via stack, it support only cabal.
But my attempts compile from sources using cabal = failed.

steve-chavez:
@dwagin Oh, I didn't know that. Maybe the archlinux package PKGBUILD can be of help: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/postgrest. Looks like it only uses `runhaskell`.

steve-chavez:
@dwagin Building with cabal is now possible!

I've added a CI for FreeBSD that has the steps to build postgrest. See here: https://github.com/PostgREST/postgrest/blob/master/.cirrus.yml#L30-L34.

The CI will also ensure that we keep building successfully on FreeBSD.

It'd be great if you could help us with a FreeBSD port!

dwagin:
@steve-chavez I think I'm making some progress.
Which architectures are suitable for postgrest? i386, amd64?

dwagin:
[hs-postgrest-7.0.0.log](https://github.com/PostgREST/postgrest/files/4606749/hs-postgrest-7.0.0.log)

steve-chavez:
@dwagin Great news!

I think amd64 would be good enough.

In case you're wondering, here's a complete list of what platforms GHC supports: https://gitlab.haskell.org/ghc/ghc/-/wikis/platforms.

dwagin:
[...]

PR #1498, opened 2020-04-23
Remove ghc options from stack.yaml

monacoremo:
They are redundant after being added to postgrest.cabal in #1489

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,13 +1,5 @@
 # stack is used for circleci and appveyor CI builds
 resolver: lts-15.8
-ghc-options:
-  # -fno-spec-constr may help keep compile time memory use in check,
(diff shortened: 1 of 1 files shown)


monacoremo:
Ok, so the options are still needed by the linters in CI - which we could replace by installing them from Nix soon, so I'll keep this around for a bit

robx:
> Ok, so the options are still needed by the linters in CI - which we could replace by installing them from Nix soon, so I'll keep this around for a bit

That's weird. Chances are rather that it's a matter of luck whether the linters build successfully or not, and the ghc options might have nothing to do with it at all...

steve-chavez:
Looks good.

Btw, another advantage of integrating better with cabal is that we could get a FreeBSD port(only supports cabal and not stack): https://github.com/PostgREST/postgrest/issues/1227#issuecomment-609482399.

PR #1499, opened 2020-04-23
Pin version 0.3 of protolude in nix

monacoremo:
Protolude 0.3, which was added as a dependency with #1496, is not yet available in our pinned version of nixpkgs. This pins the protolude package to fix the issue.

This is a great example I think for how to solve this kind of issues if they should crop up. In the future, this incompatibility would have been automatically flagged to us by the failing Nix build in CI.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -23,6 +23,7 @@ let
   overlays =
     [
       (import nix/overlays/gitignore.nix)
+      (import nix/overlays/haskell-packages)
     ];
 
   pkgs =
--- a/nix/overlays/haskell-packages/README.md
+++ b/nix/overlays/haskell-packages/README.md
@@ -1,0 +1,10 @@
+The `.nix` files in this directory pin specific versions of Haskell packages
+from Hackage. They were generated with `cabal2nix`, e.g.:
+
+```bash
+cabal2nix cabal://protolude > protolude.nix
+
+```
+
+Those overrides might become obsolete as the pinned versions are adopted into
+our pinned version of nixpkgs.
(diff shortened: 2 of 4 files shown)


steve-chavez:
I see what you mean, `cabal v2-build` doesn't work now.

```bash
cabal: Could not resolve dependencies:
[__0] trying: postgrest-7.0.0 (user goal)
[__1] next goal: protolude (dependency of postgrest)
[__1] rejecting: protolude-0.2.4/installed-5gjsiSmpp43VsHbUHeVza (conflict:
postgrest => protolude>=0.3 && <0.4)
[__1] skipping: protolude-0.2.3, protolude-0.2.2, protolude-0.2.1,
protolude-0.2, protolude-0.1.10, protolude-0.1.9, protolude-0.1.8,
protolude-0.1.7, protolude-0.1.6, protolude-0.1.5, protolude-0.1.4,
protolude-0.1.3, protolude-0.1.2, protolude-0.1.1, protolude-0.1.0 (has the
same characteristics that caused the previous version to fail: excluded by
constraint '>=0.3 && <0.4' from 'postgrest')
[__1] fail (backjumping, conflict set: postgrest, protolude)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: postgrest, protolude
```

Would be nice to have this checked in CI.

Btw, doing `cabal2nix cabal://protolude > protolude.nix` and then updating the version is really straightforward.  Really nice way of overcoming these kind of issues.

steve-chavez:
Not sure why the CircleCI style job is failing here, but in master is working normally. I'll merge this now and see how it goes.

PR #1501, opened 2020-04-25
Add scripts for linting and styling with all dependencies managed by Nix

monacoremo:


nix/lint.nix
```diff
@@ -0,0 +1,12 @@
...
+
+writeShellScriptBin "postgrest-lint"
+  ''
+    set -euo pipefail
+
+    rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
+
+    # Lint Haskell files
+    ${silver-searcher}/bin/ag -l -g '\.l?hs$' "$rootdir" \
+      | xargs ${hlint}/bin/hlint -X QuasiQuotes -X NoPatternSynonyms
```
steve-chavez, line 11: Here  you use the `silver-searcher` instead of `git ls-files | grep '\.l\?hs$$'` because the command is shorter right?

I was thinking just using `git` and `grep` would avoid having an extra dependency, but I think is fine since we use Nix here.

monacoremo: There is a small difference, this will also pick up files that are not yet staged in git, while still respecting .gitignore etc. That might be useful for contributors if they are working on adding new files.

steve-chavez: That's better! I've already tested it and it's working good. Merging..

nix/README.md
```diff
@@ -61,6 +61,20 @@ nix-shell --run postgrest-test-spec-postgresql-9.5.21
...
 ```
 
+## Linting and styling code
+
+The nix-shell also contains scripts for linting and styling the PostgREST
+source code:
+
+```bash
+# Linting
+nix-shell --run postgrest-lint
```
steve-chavez, line 71: I didn't know that when inside the `nix-shell` I could just run `postgrest-lint`.

I now understand what you meant about autocompleting with tab when doing `postgrest-test-spec-<tab>`. This is really nice  :)

steve-chavez: (I thought autocomplete was meant to work like `nix-shell --run postgrest-test-spec-<tab>`)

PR #1502, opened 2020-04-25
Relax dependencies to build with GHC 8.10 (base, swagger2)

robx:


steve-chavez:
Looks good. I restarted the style job and it worked this time. Seems there is some randomness as you noted in https://github.com/PostgREST/postgrest/pull/1498#issuecomment-619201709.

PR #1504, opened 2020-04-28
Hook the Nix styling scripts into CI

monacoremo:
This should be faster, more consistent (as the same scripts can be used for local development) and hopefully less random than the stack-based integration :-)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,9 @@ jobs:
           command: nix-env -f default.nix -iA lint style
       - run:
           name: Run linter
-          command: postgrest-lint
+          command: |
+            echo 'Note: For checking this locally, use `make lint`. Or, if using `nix`, run `nix-shell --run postgrest-lint`'
+            postgrest-lint
       - run:
           name: Run style check
           command: postgrest-style-check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
           name: Run style check
           command: |
             echo 'Note: For checking this locally, use `make style`. Or, if using `nix`, run `nix-shell --run postgrest-style`'
-            postgrest-style
+            postgrest-style-check
 
   build-test-9.4:
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -88,39 +88,18 @@ jobs:
...
-          name: Add stack tools to $PATH
-          command: |
-            echo "export PATH=/home/circleci/.local/bin:$PATH" >> $BASH_ENV
+          name: Install linting and styling scripts
+          command: nix-env -f default.nix -iA lint style
       - run:
-          name: run linter
-          command: make lint
+          name: Run linter
+          command: postgrest-lint
```
steve-chavez, line 99: For non-nix users, do you think we could suggest here a way of checking the failure locally?

I'm thinking of having an `echo` like:

```suggestion
          name: Run linter
          command: |
            echo "Note: For checking this locally, use `make lint`. Or, if using `nix`, run `nix-shell --run postgrest-lint`.
            postgrest-lint
```

.circleci/config.yml
```diff
@@ -88,39 +88,18 @@ jobs:
...
       - run:
-          name: run styler
-          command: make style
+          name: Run style check
+          command: postgrest-style-check
```
steve-chavez, line 102: Similarly here:

```suggestion
          name: Run style check
          command: |
            echo 'Note: For checking this locally, use `make style`. Or, if using `nix`, run `nix-shell --run postgrest-style`'
            postgrest-style
```

monacoremo: Yes, that makes a lot of sense! Good idea :-)

.circleci/config.yml
```diff
@@ -88,39 +88,22 @@ jobs:
...
           command: |
-            echo "export PATH=/home/circleci/.local/bin:$PATH" >> $BASH_ENV
-      - run:
-          name: run linter
-          command: make lint
-      - run:
-          name: run styler
-          command: make style
+            echo 'Note: For checking this locally, use `make style`. Or, if using `nix`, run `nix-shell --run postgrest-style`'
+            postgrest-style
```
steve-chavez, line 106: Made a mistake in my suggestion here. It should have been:

```suggestion
            postgrest-style-check
```

PR #1505, opened 2020-04-28
Add tips and tricks to using nix-shell

monacoremo:


nix/README.md
```diff
@@ -42,6 +42,57 @@ Within `nix-shell`, you can run Cabal commands as usual. You can also run
...
+# Note that you need to quote any arguments that you would like to pass to
+# the command to be run in nix-shell:
+$ nix-shell --run "postgrest-foo --bar"
+
+```
+
+A third option is to install utilities that you use very often locally:
+
+```
+$ nix-env -f default -iA style
```
steve-chavez, line 82: This should be `default.nix` right?

Otherwise I get:
```
nix-env -f default -iA postgrest
error: getting information about '/home/steve-chavez/Projects/Haskell/postgrest/default': No such file or directory
```
```suggestion
$ nix-env -f default.nix -iA style
```

monacoremo: Absolutely - good catch, thanks!

steve-chavez:
Very nice elucidation.

For building and running postgrest now I use:

```
nix-env -f default.nix -iA postgrest && postgrest config
```

Another good thing is that I can use the latest postgrest executable in every other shell.

PR #1506, opened 2020-04-28
More consistent loading of the test fixtures

monacoremo:
As discussed on #1488

### DIFF
--- a/test/Main.hs
+++ b/test/Main.hs
@@ -56,7 +56,6 @@ main = do
   getTime <- mkAutoUpdate defaultUpdateSettings { updateAction = getCurrentTime }
 
   testDbConn <- getEnvVarWithDefault "POSTGREST_TEST_CONNECTION" "postgres://postgrest_test@localhost/postgrest_test"
-  resetDb testDbConn
 
   pool <- P.acquire (3, 10, toS testDbConn)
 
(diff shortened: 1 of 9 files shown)


test/Main.hs
```diff
@@ -56,7 +56,7 @@ main = do
   getTime <- mkAutoUpdate defaultUpdateSettings { updateAction = getCurrentTime }
 
   testDbConn <- getEnvVarWithDefault "POSTGREST_TEST_CONNECTION" "postgres://postgrest_test@localhost/postgrest_test"
-  setupDb testDbConn
+  resetDb testDbConn
```
steve-chavez, line 59: Resetting the data at the start is not needed right? The old `setupDb` didn't run `data.sql` as well. I've tested it and I think it's safe to remove.

```suggestion
```

monacoremo: Yes, that would be even cleaner! I think this will require that nobody runs the tests on the same database twice - which is probably reasonable. Maybe we would need to document this for the users of `test/create_test_db`?

I would then also remove the `resetDB` and `loadFixture` functions in `SpecHelper`, no?

monacoremo: Ah, no, those are still needed for the resets in line 93 in Main.hs. It should also be fine for reruns as far as I can see!

steve-chavez: Yes, some tests mutate the db and the `reset` function handles that. I think reruns will be fine.

PR #1507, opened 2020-04-28
Allow base64-bytestring 1.2.

robx:
Seems like a harmless change: http://hackage.haskell.org/package/base64-bytestring-1.1.0.0/changelog

Would a 7.0.1 release to hackage be feasible, since we've got the GHC 8.10 changes together now? (Actually this one isn't required, but we might as well include it.)

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -48,7 +48,7 @@ library
                     , Ranged-sets               >= 0.3 && < 0.5
                     , aeson                     >= 1.4.7 && < 1.5
                     , ansi-wl-pprint            >= 0.6.7 && < 0.7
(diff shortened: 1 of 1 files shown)


robx:
Compare https://github.com/commercialhaskell/stackage/issues/5323

steve-chavez:
Looks good.

> Would a 7.0.1 release to hackage be feasible, since we've got the GHC 8.10 changes together now?

Yes, I'm going to solve some bugs in https://github.com/PostgREST/postgrest/pull/1509 and then I'll make a new patch release.

steve-chavez:
@robx Just released v7.0.1! 

https://hackage.haskell.org/package/postgrest-7.0.1

PR #1510, opened 2020-04-28
Fix the cachix push logic, the obvious way...

monacoremo:
This is the obvious solution with no security downsides, not sure why I didn't use that in the first place!


steve-chavez:
Makes a lot of sense! Thanks for the prompt help :)

PR #1513, opened 2020-05-02
Get DbStructure in a single query

monacoremo:
For discussion - this is a working proof of concept, with currently 5 failing tests. It was 100 % green across all PostgreSQL versions until ddf286c, I would need some help to solve the failures that sprung up by refactoring the procs.

Main changes:
* Factored out the SQL queries in `DbStructure.hs` into on big one in `dbstructure/query.sql`. It's possible to run this query independently from PostgREST with the `dbstructure/show` script (schemas are not parametrized yet, but would be simple to add). This was extremely helpful for debugging and I think it would be very useful for future development and triaging issues ('what is PostgREST seeing?'). The query returns one big JSON object.
* Derived generic Aeson `FromJSON` instances for all relevant DbStructure types (no manual decoders! :-) )
* Decode the query result with Aeson. For example, `table_name` in the query is decoded as the `tableName` field in the `Table` type.
* Added some glue to the current data wrangling functions in `DbStructure`
* Reduced duplicated code by providing the accessible tables and procs in `DbStructure`

The query and the remaining parts of `DbStructure` could be refactored a lot, but I tried to make this as much a 1:1 translation of the current setup as possible.


steve-chavez:
Looks great overall!

I'm going to give it a review now. I had some issues with building locally because of the `file-embed` dependency:

```
postgrest-test-spec

Running spec against postgresql-12.2...

Warning: The package list for 'hackage.haskell.org' is 310 days old.
Run 'cabal update' to get the latest list of available packages.
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: postgrest-7.0.0 (user goal)
[__1] next goal: file-embed (dependency of postgrest)
[__1] rejecting: file-embed-0.0.11 (conflict: postgrest =>
file-embed>=0.0.11.2 && <0.0.12)
[__1] skipping: file-embed-0.0.10.1, file-embed-0.0.10, file-embed-0.0.9.1,
file-embed-0.0.9, file-embed-0.0.8.2, file-embed-0.0.8.1, file-embed-0.0.8,
file-embed-0.0.7.1, file-embed-0.0.7, file-embed-0.0.6, file-embed-0.0.5,
file-embed-0.0.4.9, file-embed-0.0.4.8, file-embed-0.0.4.7,
file-embed-0.0.4.6, file-embed-0.0.4.5, file-embed-0.0.4.4,
file-embed-0.0.4.3, file-embed-0.0.4.2, file-embed-0.0.4.1, file-embed-0.0.4,
file-embed-0.0.3.1, file-embed-0.0.3, file-embed-0.0.2, file-embed-0.0.1,
file-embed-0.0.0 (has the same characteristics that caused the previous
version to fail: excluded by constraint '>=0.0.11.2 && <0.0.12' from
'postgrest')
[__1] fail (backjumping, conflict set: file-embed, postgrest)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: postgrest, file-embed
```

After doing a `cabal update` it all worked fine. However, I thought `nix build` would handle the dependencies and I wouldn't need to handle cabal. In CircleCI, I see only doing `nix build` makes all work. Not sure what I'm missing.

steve-chavez:
I did a perf test(on a big schema) of the new query. There's a noticeable performance regression, but I think it's fixable.

- Old query: Time 409.00 ms
- New query: Time 3110.303 ms

The bulk of the cost is coming from the new `source_columns` CTE.

dbstructure/query.sql
```diff
@@ -0,0 +1,483 @@
...
+                and col.attname = c1.col_name
+       ) source_column,
+       lateral (
+            select c2
+            from columns c2
+            where
+                res.view_schema = c2.col_schema
+                and res.view_name = c2.col_table_name
+                and res.view_colum_name = c2.col_name
+       ) as view_column
```
steve-chavez, line 456: These laterals are what's making the query costly. I think we would get the same time(or better) as the old query if we remove them. Do you think it could be possible to go back to making the matching of the columns and source_columns in Haskell code?

steve-chavez: The number of columns can get big. I don't see a good way for getting them in the query for now.

steve-chavez: (In case you'd like to better test this. I can send you a sample big schema to your email.)

monacoremo: @steve-chavez Very interesting, thank you for investigating!! Do you think it's an issue with the amount of data that we need to transfer with this approach?

I think there is a lot of potential for optimizing the big query, would be exciting to see if we can push through and get this much faster - which should be possible in principle I believe. If there is a larger data set that you can share, that would be very helpful!

dbstructure/query.sql
```diff
@@ -0,0 +1,507 @@
...
+          select
+            matches[1]::smallint as attnum,
+            matches[2] as attname,
+            matches[3]::oid as table_oid,
+            matches[3]::smallint as table_col_position
+          from
+            regexp_matches(
+              r.ev_action,
+              ':resno (\d+) :resname (\S+) :ressortgroupref \d+ :resorigtbl (\d+) :resorigcol (\d+)',
+              'g'
```
steve-chavez, line 149: This looks like it really simplifies my work on https://gist.github.com/steve-chavez/7ee0e6590cddafb532e5f00c46275569.

I've noticed you don't do the cleaning work I did on the previous query. Does this only target the VIEW root select(not the subselects or SELECT in WHERE or other parts)?

monacoremo: Amazing work on this by the way, figuring out how to parse ev_action in sql is a feat!

This is an experiment to simplify the query, not even sure if it's faster yet... your version is already very optimal. The idea here is to make regex simpler (and hopefully faster). But as you note, this skips the more precise cleaning from your version. To compensate, the targetentries are filtered by which ones are valid columns of the view with the same attnum and name. So this would always find all relevant columns, but it might also include some additional ones in some edge cases (e.g. calculated columns). This could only be worth it if it was radically faster, bit that doesn't seem to be the case.

monacoremo:
closing in favor of #1612

PR #1522, opened 2020-05-13
Build Docker image from Nix

monacoremo:
This builds a PostgREST Docker image directly from Nix using the static executable. Usage would look something like this:

```bash
$ nix-build -A docker.load
...

$ result/bin/postgrest-docker-load
Loaded image: postgrest/postgrest:latest
Loaded image: postgrest/postgrest:v7.0.0

$ docker push postgrest/postgrest:latest
The push refers to repository [docker.io/postgrest/postgrest]
9c714c7702db: Pushed                                 
latest: digest: sha256:5dba29a5d2c8c257307378da428e46c7d2096d369c717a05a0056596eea2d291 size: 529
```

I re-tagged the built image a pushed it here for testing: [monacoremo/postgrest:latest](
https://hub.docker.com/layers/monacoremo/postgrest/latest/images/sha256-5dba29a5d2c8c257307378da428e46c7d2096d369c717a05a0056596eea2d291)

Benefits:
* Building the image from Nix does not actually require Docker or root privileges, so it might easier to build it locally or in CI.
* The result is significantly smaller than the current images (13mb vs. 38mb compressed).

Possible issues:
* The resulting image is absolutely bare bones, so it does not include bash or anything else. Those could be added if needed, e.g. to help users to introspect the image with `docker run`. 
* The creation date is set to 1970-01-01, as that keeps the build repeatable. We could specify `created = "now";` in the `dockerTools.buildImage` call to get the current time.

### DIFF
--- a/nix/docker/default.nix
+++ b/nix/docker/default.nix
@@ -8,6 +8,10 @@ let
         name = "postgrest/postgrest";
         contents = postgrest;
 
+        # Set the current time as the image creation date. This makes the build
+        # non-reproducible, but that should not be an issue for us.
+        created = "now";
+
         extraCommands =
           ''
             mkdir etc
--- a/nix/docker/default.nix
+++ b/nix/docker/default.nix
@@ -8,7 +8,7 @@ rec {
       contents = postgrest;
       extraCommands = ''
         mkdir etc
-        cp ${../docker/postgrest.conf} etc/postgrest.conf
+        cp ${./postgrest.conf} etc/postgrest.conf
       '';
       config = {
         Cmd = [ "/bin/postgrest" "/etc/postgrest.conf" ];
--- a/nix/docker/default.nix
+++ b/nix/docker/default.nix
@@ -5,7 +5,7 @@ let
       dockerTools.buildImage {
         inherit tag;
 
-        name = "postgrest";
+        name = "postgrest/postgrest";
         contents = postgrest;
 
         extraCommands =
--- a/nix/docker/default.nix
+++ b/nix/docker/default.nix
@@ -1,45 +1,59 @@
 { postgrest, dockerTools, writeShellScriptBin }:
-
-rec {
+let
   image =
-    dockerTools.buildImage {
-      name = "postgrest";
-      tag = "latest";
-      contents = postgrest;
-      extraCommands = ''
-        mkdir etc
-        cp ${./postgrest.conf} etc/postgrest.conf
-      '';
-      config = {
-        Cmd = [ "/bin/postgrest" "/etc/postgrest.conf" ];
-        Env = [
-          "PGRST_DB_URI=postgresql://?user=postgres"
-          "PGRST_DB_SCHEMA=public"
-          "PGRST_DB_ANON_ROLE="
-          "PGRST_DB_POOL=100"
-          "PGRST_DB_EXTRA_SEARCH_PATH=public"
-          "PGRST_SERVER_HOST=*4"
-          "PGRST_SERVER_PORT=3000"
-          "PGRST_OPENAPI_SERVER_PROXY_URI="
-          "PGRST_JWT_SECRET="
-          "PGRST_SECRET_IS_BASE64=false"
-          "PGRST_JWT_AUD="
-          "PGRST_MAX_ROWS="
-          "PGRST_PRE_REQUEST="
-          "PGRST_ROLE_CLAIM_KEY=.role"
-          "PGRST_ROOT_SPEC="
-          "PGRST_RAW_MEDIA_TYPES="
-        ];
-        User = "1000";
-        ExposedPorts = {
-          "3000/tcp" = {};
+    tag:
+      dockerTools.buildImage {
+        inherit tag;
+
+        name = "postgrest";
+        contents = postgrest;
+
+        extraCommands =
+          ''
+            mkdir etc
+            cp ${./postgrest.conf} etc/postgrest.conf
+          '';
+
+        config = {
+          Cmd = [ "/bin/postgrest" "/etc/postgrest.conf" ];
+          Env = [
+            "PGRST_DB_URI=postgresql://?user=postgres"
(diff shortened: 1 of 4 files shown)


steve-chavez:
@monacoremo Interesting!

The command to build the image should be: `nix-build -A loadDocker` right?(tried `load.docker` but gave an error).

(Been checking [docker/examples.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/docker/examples.nix) and [docker/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/docker/default.nix) to understand this a bit better).

Some questions:
  - How should I run the docker image after creating it with `loadDocker`?
  - 1970-01-01: What would be the downsides of setting this? Leaning on `now` for the meantime.
  - libpq: One other advantage I see with this approach is that I wouldn't have to change the base image for getting a new dependency(had to switch to `debian:buster-slim` recently https://github.com/PostgREST/postgrest/pull/1467). But I don't see `libpq` on the nix files, how should this be specified?

monacoremo:
> The command to build the image should be: `nix-build -A loadDocker` right?(tried `load.docker` but gave an error).

@steve-chavez It should be `docker.load`, the other way round :-) `load` is one of the attribute returned from [nix/docker/default.nix](https://github.com/monacoremo/postgrest/blob/monacoremo-nix-static-docker/nix/docker/default.nix). You should then be able to run `result/bin/postgrest-docker-load`, which runs `docker load -i` on the generated image (the image can also be built directly with `nix-build -A docker.imageLatest`).

After that, running it should be possible with something like `docker run postgrest/postgrest:latest`!

> 1970-01-01: What would be the downsides of setting this? Leaning on now for the meantime.

The builds would not be repeatable, but I don't think that would be an issue for us.

> libpq

It's baked in statically! `nix-store -qR $(nix-instantiate -A postgrestStatic) | grep libpq` tells me its the version `postgresql-libpq-0.9.4.2`.

steve-chavez:
> It should be docker.load, the other way round :-) 

Sorry about that. I grabbed the wrong branch(`monacoremo-nix-docker`) and that command didn't run for me.

`nix-build -A docker.imageLatest` works great!. Now `docker images` shows the updated `postgrest/postgrest`.

(Btw, `cachix use postgrest` made the build really fast)

> libpq It's baked in statically! 

Oh, I forgot that. Is that libpq version `>=` than the debian buster version we were using?: https://packages.debian.org/buster/libpq5. Not sure if the nix static version could break https://github.com/PostgREST/postgrest/pull/1467.

nix/docker/default.nix
```diff
@@ -0,0 +1,63 @@
...
+
+        # Set the current time as the image creation date. This makes the build
+        # non-reproducible, but that should not be an issue for us.
+        created = "now";
+
+        extraCommands =
+          ''
+            mkdir etc
+            cp ${./postgrest.conf} etc/postgrest.conf
+          '';
```
steve-chavez, line 19: How about the non-root group/user? I saw you added those in: https://github.com/monacoremo/postgrest/commit/694fd927a8c71fe819852623ec99bd8c1fa5f98e#diff-5712e736e0de6ba170577f8472c398e9R92-R93.

For reference, we have them in our [Dockerfile](https://github.com/PostgREST/postgrest/blob/master/docker/Dockerfile#L47-L48)

monacoremo:
@steve-chavez Re. libpq: It should be the same, buster is also is on Postgres 11.7!

nix/docker/default.nix
monacoremo: Yes, that should already be the case :-) It runs as non-root based on the line `User = "1000";`. 

In the bare bones image, actually creating that user in `/etc/passwd`, `shadow` etc. does not seem to be necessary, as there is only our static binary in it and nothing else that could possibly inspect those files.

The only case that I can think of where that could make a difference is when `libpq` tries to infer the database login user name from the local user. This would resolve to `postgrest` in the current image and fail here. I don't think its likely that someone relies on this, however, and even if, it would be easily fixed by explicitly setting the username.

monacoremo: Looked into that a bit more, I found a few references that the user does not have to actually exist:
* https://docs.docker.com/engine/reference/run/#user
> When passing a numeric ID, the user does not have to exist in the container.
* [Openshift](https://docs.openshift.com/container-platform/4.4/openshift_images/create-images.html#use-uid_create-images) (more details in the [older docs](https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html#use-uid) run containers under arbitrary user ids that may or may not exist as users in the container

This should work well this way!

monacoremo:
[This branch](https://github.com/monacoremo/postgrest/tree/monacoremo-nix-static-docker-opensslfix) contains a proof of concept for patching the openssl dependency bug, which brings the image size down to 4mb... pretty wild :-)

image

Uploaded here for testing: https://hub.docker.com/repository/docker/monacoremo/postgrest/tags

We could merge this later though, as many packages (including GHC, again...) depend on openssl, so this triggers a huge rebuild. The openssl patching could maybe be done a bit more elegantly also. Submitted a first PR to Nixpkgs here: https://github.com/NixOS/nixpkgs/pull/87879

nix/docker/default.nix
steve-chavez: I see. I forgot we also specify `USER 1000` in the Dockerfile.

So adding the postgrest user/group through `useradd/groupadd` wouldn't be necessary anymore?

monacoremo: Yes, it's not needed in this case as there is only postgrest in the image and any uid has execute permissions on it. Thats exactly enough to run the application correctly!

There is no other program like bash, whoami etc. in the image that would look at /etc/passwd, so we don't actually need it.

steve-chavez:
@monacoremo Awesome!

I'll merge this now. Would be great to have the openssl fix later.

PR #1524, opened 2020-05-14
Release 7.0.1

steve-chavez:


steve-chavez:
Releasing centos7 and ubuntui386 failed:

- [centos7](https://circleci.com/gh/steve-chavez/postgrest/2699): Cannot open: https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-centos10-10-2.noarch.rpm. Skipping.
- [ubuntui386](https://circleci.com/gh/steve-chavez/postgrest/2701): Sorry, currently only 64-bit (x86_64) Linux binary is available(stack)

This is why Nix is important. Ensuring reproducible builds would save this trouble.

steve-chavez:
I'm leaning on just recommending to use the x64 static bin and dropping the centos7/ubuntu binaries.

Not sure about the ubuntui386 for now. Maybe I can build with cabal instead of stack of the 32 bit container.

steve-chavez:
Decided to drop the previous linux releases, including the 32 bit ubuntu. 

The nix static bin is more reliable and simpler to check. Offering a single linux bin would also prevent issues like https://github.com/PostgREST/postgrest/issues/1527.

We'll figure out how to build a 32 bit static bin later.

PR #1528, opened 2020-05-18
Reduce size of the static docker image by patching the Nix openssl package

monacoremo:
As discussed in #1522, this reduces the closure size of the Nix static build by improving the openssl package.

Benefit: The size of the static docker image should be reduced to 4mb.

Issues: This will trigger a rebuild for all our dependencies that depend on static openssl, so this will almost certainly time out in CI... :-)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -81,12 +81,18 @@ rec {
   postgrestStatic =
     lib.justStaticExecutables (lib.dontCheck drvStatic);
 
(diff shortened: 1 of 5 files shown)


steve-chavez:
LGTM :+1:

Did a quick  `nix-build -A docker.load` and ran the container. All's good!

Issue #1117, opened 2018-05-20
Docker image compatible with ARM: is that possible?


cuducos:
### Environment

* PostgreSQL version: `postgres:10.4`
* PostgREST version: `v0.5.0.0`
* Operating system: Raspbian GNU/Linux 9 (stretch)

<details>

```
pi@raspberrypi:~ $ uname -a
Linux raspberrypi 4.14.34-v7+ #1110 SMP Mon Apr 16 15:18:51 BST 2018 armv7l GNU/Linux
pi@raspberrypi:~ $ cat /etc/os-release
PRETTY_NAME="Raspbian GNU/Linux 9 (stretch)"
NAME="Raspbian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

</details>

### Description of issue

Probably related to #846, but I would like to ask about the possibility of a Docker image ready for ARM processors. So far I think the PostgREST Docker image is not compatible:

```
pi@raspberrypi:~ $ docker pull postgrest/postgrest
Using default tag: latest
latest: Pulling from postgrest/postgrest
Digest: sha256:8c40ba14f6400a51be1ed78f83f472132a96ab955fe627ca385286378dd3f88b
Status: Image is up to date for postgrest/postgrest:latest
pi@raspberrypi:~ $ docker run postgrest/postgrest


standard_init_linux.go:190: exec user process caused "exec format error"
```

Also I see only `amd64` in the image inspector:

```
pi@raspberrypi:~ $ docker inspect postgrest/postgrest | grep Architecture
        "Architecture": "amd64",
```

Is this possible and in the radar for short term releases? Is there someway I could help in this particular issue?

steve-chavez:
I've uploaded a PostgREST binary for ARMv7 here https://github.com/steve-chavez/postgrest/releases/download/v0.5.0.0/postgrest-v0.5.0.0-debian-armv7.tar.xz, maybe you can test it so I can later see how to create a docker image for that and make it an official release.

If you'd like to you can also build your own docker image with that binary by following this https://github.com/PostgREST/postgrest/blob/master/docker/Dockerfile.

Let me know how it goes, I don't own an ARM device so what I did was to use qemu and docker to build the binary, basically:

[...]

PR #1529, opened 2020-05-19
Bump nixpkgs and simplify/clean up nix setup

monacoremo:
This is on top of #1528, will rebase. **Update:** rebased.

Now that we have a pretty complete Nix setup and gained some experience, I noticed some opportunities to simplify the Nix setup. Fixes include:
* Explicitly set the default Postgresql version we want to use in an overlay (this will also be the version that `libpq` is taken from)
* Explicitly specify the GHC version being used when accessing haskellPackages (this and the one above makes the setup more independent from changes in Nixpkgs)
* Merge the scripts related to development into one Nix file, `devtools.nix` - no real need to have them in separate files
* Merge the haskellPackages overlay for static and dynamic builds (should more maintainable, as changes don't need to be synced between the two anymore)
* Only have the major versions of the different Postgresql versions in the testing script names (i.e. `9.4` instead of `9.4.27`, easier to use locally and potentially in CI)
* Add a recommendation for `cachix use postgrest` in the README
* Pin protolude using `callHackageDirect`, instead of creating a local file with cabal2nix

I also bumped Nixpkgs to the current nixpkgs-unstable version, which caused no major changes.* I think it's good if we track unstable closely for now, at least until all our patches and already merged PRs have landed in a stable version.

*edit: It did yield new hlint hints and nixpkgs-fmt formatting changes, both added in separate commits.

Those a quite a few changes, but as everything is related to the Nix setup I hope it's still good to review!


steve-chavez:
Looks good!

I gave it a look and also tried the new `postgrest-test-spec` commands, I like the shorter version numbers.

Combining the static and dynamic build into a single config is great. Btw, maybe we can later leverage this to build a static bin for arm64. There was some work for this on https://github.com/PostgREST/postgrest/issues/1117#issuecomment-599060315, but there were some issues.

Merging now!

PR #1532, opened 2020-05-19
cabal: fix upload to hackage

steve-chavez:
When doing:

```
cabal upload dist-newstyle/sdist/postgrest-7.0.1.tar.gz
```

The following error is shown:

```
Error: Invalid package

'ghc-options: -Wall -Werror' makes the package very easy to break with
future GHC versions because new GHC versions often add new warnings.
Use just 'ghc-options: -Wall' instead. Alternatively, if you want to
use this, make it conditional based on a Cabal configuration flag
(with 'manual: True' and 'default: False') and enable that flag
during development.
```

Put -Werror in a cabal flag to work around this restriction.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -52,7 +52,7 @@ let
 
   # Base dynamic derivation for the PostgREST package.
   drv =
-    pkgs.haskellPackages.callCabal2nix name src {};
+    lib.enableCabalFlag (pkgs.haskellPackages.callCabal2nix name src {}) "FailOnWarn";
 
   # Static derivation for the PostgREST executable.
   drvStatic =
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -19,7 +19,7 @@ source-repository head
   type:     git
   location: git://github.com/PostgREST/postgrest.git
 
-flag werror
+flag FailOnWarn
   default:     False
   manual:      True
   description: No warnings allowed
@@ -88,7 +88,7 @@ library
   default-extensions: OverloadedStrings
                       QuasiQuotes
                       NoImplicitPrelude
-  if flag(werror)
+  if flag(FailOnWarn)
     ghc-options: -O2 -Werror -Wall -fwarn-identities
                  -fno-spec-constr -optP-Wno-nonportable-include-path
   else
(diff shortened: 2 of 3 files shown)


steve-chavez:
@monacoremo How can I pass an arg to callCaball2nix?

With plain cabal it would be like:
```
cabal v2-build -f werror 
``` 

I'm guessing the last set [here](https://github.com/PostgREST/postgrest/blob/master/default.nix#L55) should be used, but not sure how.

For reference, see https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/make-package-set.nix#L216.

monacoremo:
@steve-chavez There is a [`callCabal2nixWithOptions`](https://github.com/NixOS/nixpkgs/blob/f5b6ea126f0f85cd2126984607e95151daf9a859/pkgs/development/haskell-modules/make-package-set.nix#L201), but I think in this case it would be best to use something like [`lib.appendConfigureFlag`](https://github.com/NixOS/nixpkgs/blob/698e71db2c47b5d3a206c65288f88ab4320eaf1e/pkgs/development/haskell-modules/lib.nix#L137) to pass a ghc flag.

So a quick fix would be something like:

```
drv =
  lib.appendConfigureFlag (pkgs.haskellPackages.callCabal2nix name src {}) "-ghc-option=-Werror";
```

Looking at this, I just found [`failOnAllWarnings`](https://github.com/NixOS/nixpkgs/blob/698e71db2c47b5d3a206c65288f88ab4320eaf1e/pkgs/development/haskell-modules/lib.nix#L274). So the best solution probably is:

```
drv =
  lib.failOnAllWarnings (pkgs.haskellPackages.callCabal2nix name src {});
```

steve-chavez:
@monacoremo Thanks!

I ended up using [enableCabalFlag](https://github.com/NixOS/nixpkgs/blob/698e71db2c47b5d3a206c65288f88ab4320eaf1e/pkgs/development/haskell-modules/lib.nix#L165). I wanted to keep the cabal file as the source of truth. That way stack and nix would use the same ghc-options. Could be helpful in case we change something in the future.

PR #1533, opened 2020-05-20
fix UNKNOWN being shown as git commit hash (fixes #1530)

### LINKED ISSUE #1530: Git revision shows as 'UNKNOWN' in help text
With the static builds, the Git revision is shown as UNKNOWN:

```
Usage: postgrest FILENAME
  PostgREST 7.0.1 (UNKNOWN) / create a REST API to an existing Postgres database
```

This is because we don't include the .git directory in the source files and `gitrev` cannot find the reference. We could try to
* work around this by including parts of the .git directory (probably quite messy)
* find another way to pass the revision into the build process, or
* remove showing the revision from the help text and remove the dependency on `gitrev`

monacoremo:
Fix for #1530 

This might be a good compromise - it will show the commit hash in the help text if `.git` is available (e.g. when someone built PostgREST from source), but it will not show `UNKNOWN` if it's not.

There are some cases where `.git` simply can't be available (e.g. when cabal or nix build the package from a sdist tarball from Hackage), so this would give us the best of both worlds I think.

Example when built with nix:
```
Usage: postgrest FILENAME
  PostgREST 7.0.1 / create a REST API to an existing Postgres database
```

Example when built with stack:
```
Usage: postgrest FILENAME
  PostgREST 7.0.1 (69516b9) / create a REST API to an existing Postgres database
```

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -127,10 +127,9 @@ prettyVersion =
   intercalate "." (map show $ versionBranch version) <> gitRev
   where
     gitRev =
-      if $(gitHash) == "UNKNOWN" then
-        ""
-      else
-        " (" <> take 7 $(gitHash) <> ")"
+      if $(gitHash) == "UNKNOWN"
+        then mempty
+        else " (" <> take 7 $(gitHash) <> ")"
 
 -- | Version number used in docs
 docsVersion :: Text
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -124,8 +124,13 @@ corsPolicy req = case lookup "origin" headers of
 -- | User friendly version number
 prettyVersion :: Text
 prettyVersion =
-  intercalate "." (map show $ versionBranch version)
-  <> " (" <> take 7 $(gitHash) <> ")"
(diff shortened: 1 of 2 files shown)


src/PostgREST/Config.hs
```diff
@@ -124,8 +124,13 @@ corsPolicy req = case lookup "origin" headers of
...
 prettyVersion =
-  intercalate "." (map show $ versionBranch version)
-  <> " (" <> take 7 $(gitHash) <> ")"
+  intercalate "." (map show $ versionBranch version) <> gitRev
+  where
+    gitRev =
+      if $(gitHash) == "UNKNOWN" then
+        ""
+      else
+        " (" <> take 7 $(gitHash) <> ")"
```
steve-chavez, line 133: A small nitpick :)
```suggestion
      if $(gitHash) == "UNKNOWN"
        then mempty
        else " (" <> take 7 $(gitHash) <> ")"
```

steve-chavez:
That's a clean solution.

I think it should be included as a fix on the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md#fixed) since it clears up the pgrst help text.

monacoremo:
@steve-chavez good idea, I added an entry to the changelog!

PR #1534, opened 2020-05-21
Fix the static build

monacoremo:
Just noticed that I broke the static build with #1529, it resulted in a dynamic build based on Musl :-|

This is because static-haskell-nix apparently only works if the package to be built is already included in the package set that is passed to it. Fixed and added a comment to that effect, and re-added the openssl patch is still required.


steve-chavez:
Oh, I didn't notice that before. ldd now says:

```bash
# After building
$ nix-build -A postgrestStatic
...
copying path '/nix/store/3mhpr1gj5lzpzw62j2dgw14mr3g92ln2-postgrest-7.0.1' from 'https://postgrest.cachix.org'...
/nix/store/3mhpr1gj5lzpzw62j2dgw14mr3g92ln2-postgrest-7.0.1

$ ldd /nix/store/3mhpr1gj5lzpzw62j2dgw14mr3g92ln2-postgrest-7.0.1/bin/postgrest

linux-vdso.so.1 (0x00007ffcf4103000)
libz.so.1 => /nix/store/bzbx09i04l7x6qq54zg72x1l8m1ja2la-zlib-1.2.11/lib/libz.so.1 (0x00007f2aba500000)
libpq.so.5 => /nix/store/1papww45vk8rbk5sg87417587hxkmn5y-postgresql-11.7-lib/lib/libpq.so.5 (0x00007f2aba1c0000)
libgmp.so.10 => /nix/store/jryq2vbdcqkmnay1qn202127xmzg968z-gmp-6.2.0/lib/libgmp.so.10 (0x00007f2aba118000)
libffi.so.7 => /nix/store/fdyqkr9qzwva0f02ppm55vppv4xxns2a-libffi-3.3/lib/libffi.so.7 (0x00007f2aba108000)
libc.so => /nix/store/vwkdykrhjdpbn5gikfns21j313kp128k-musl-1.1.24/lib/libc.so (0x00007f2aba058000)
```

Are the new build results(from this PR) pushed to cachix? Doing `nix-build -A postgrestStatic` is getting me to compile(don't know why haskell conduit is being compiled).

monacoremo:
Some stuff was missing I think, did `nix-store -qR --include-outputs $(nix-instantiate)` and it should now all be cached!

steve-chavez:
@monacoremo Thanks! Cachix is just amazing.

I can confirm the static build is good now.

Merging..

monacoremo:
@steve-chavez nice! :-) 

And I agree - The whole Nix tooling is pretty magical, it's hard to imagine doing some of the stuff we are playing with here without it. Especially static-haskell-nix with "lets just transform a set of thousands of packages and switch everything to musl" in a handful of lines.

Really fun and fullfilling to work with this! And hopefully beneficial for PostgREST long term.

steve-chavez:
> And hopefully beneficial for PostgREST long term.

@monacoremo It definitely will!

Nix has already has improved my workflow. I run `nix-shell`(aliased to [cached-nix-shell](https://github.com/xzfc/cached-nix-shell)) and then do `postgrest-test-spec` for development. Having all the pg versions at your fingertips is also amazing. Can't see myself going back to Docker for this.

PR #1535, opened 2020-05-21
Simplify CircleCI setup

monacoremo:
This is more of a collection of ideas for discussion:

* Implement the release scripts with the help of nix: should make it much easier to test and maintain the release process. This pins the `ghr` tool and wraps the release process in two scripts, `postgrest-release-github` and `postgrest-release-docker`. For example you can run: `nix-env -iA release -f default.nix; DOCKER_REPO=monacoremo result/bin/postgrest-release-docker 
` to test the docker release.
* Keep one stack + docker based test, but run the tests against all PostgreSQL versions with Nix
* The docker directory could be removed, now that all release builds and docker images are done with nix


docker/Dockerfile
```diff
@@ -1,51 +0,0 @@
-# To build use:
```
steve-chavez, line 1: Pretty much agree with removing the other Dockerfiles.

But this one still has a link in docker hub: https://hub.docker.com/r/postgrest/postgrest.

Maybe we should keep it until we make a new release with the new docker image.

nix/release.nix
```diff
@@ -0,0 +1,56 @@
+{ buildEnv
+, ghr
+, writeShellScriptBin
+, postgrest
+, postgrestDocker
+, runCommand
+}:
+let
+  version = postgrest.version;
```
steve-chavez, line 9: Trying to understand a bit. How `postgrest` gets the `version` attribute?

monacoremo: Yes, that's not so obvious - it comes from `postgrest.cabal` via `callCabal2nix`. As I understand, most derivations in nix have that attribute, next to `pname` for the package name, for example.

docker/Dockerfile
monacoremo: Ah, yes - we should definitely keep it then! Does the docker-compose file still have a function?

steve-chavez: Once we make a release with the new image we'll drop this Dockerfile.

Also, we'll probably need some wording on the new image. Maybe a README inside `./nix/docker`? This could then be copy pasted to the docker hub overview.

I'm reading https://nixos.org/nixpkgs/manual/#sec-pkgs-dockerTools. And I think we could put something there about why the Nix approach allows us to get better reproducibility and a leaner image.

A change like this would also merit an addition in the CHANGELOG.

What do you think?

steve-chavez: > Ah, yes - we should definitely keep it then! Does the docker-compose file still have a function?

On second thought, I think I can just edit the docker hub overview and point it to a previous commit. No need for restoring the files.

But we would need to have the above README and entry in the CHANGELOG. We have a lot of dockers users, they'll wonder about the change.

nix/release.nix
steve-chavez: Oh, I see.

I was trying to do a `builtins.trace` to understand that a bit more. But somehow doing:

```
nix-env -iA release -f default.nix
```
Makes my laptop freeze :fearful:(twice now). Never had this before, is there some evaluation in the nix file that could be expensive?

steve-chavez: (After adding a `-v` to the command, I'm seeing it's evaluating a lot of files, maybe it's rebuilding the static bin somehow. I'll see if cachix is set).

steve-chavez:
Nice!

I already did a couple of test releases on https://github.com/steve-chavez/postgrest/releases/tag/v7.0.1 with `postgrest-release-github`.

I might later add uploading to hackage to `release.nix`. This is much nicer than following my checklist.

docker/Dockerfile
monacoremo: @steve-chavez Sounds good!

I've added:
* a [README](https://github.com/monacoremo/postgrest/tree/monacoremo-release/nix/docker) for the docker build
* a first draft for the [description on Docker Hub](https://github.com/monacoremo/postgrest/blob/monacoremo-release/nix/release/docker-hub-full-description.md) (tested [here](https://hub.docker.com/repository/docker/monacoremo/postgrest))
* a script in `nix/release/default.nix` that pushes the description to Docker Hub, `postgrest-release-dockerhubdescription`
* and the CHANGELOG entry

nix/release.nix
monacoremo: Hm, that's weird - everything froze? Evaluating the Nix expression without any cache is probably not cheap, but it should not break a whole machine...

monacoremo:
> `io-tests.sh` and `memory-tests.sh` could be tested with Nix as well.

Yes, I had a look at that back here: https://github.com/monacoremo/postgrest/commit/8267ae9759a26136ce80d80249bf830e1edb99d3 . Should not be too difficult to rebase! That would also make it much easier to run those locally. But I'd submit that in a separate PR, this one is getting a bit out of hand :-)

docker/Dockerfile
steve-chavez: @monacoremo That's some advanced Docker stuff :).

TIL thanks to your README!. The explanation is perfect and the docker hub description release is extra nice.

nix/release.nix
steve-chavez: Yeah, the xserver froze and I couldn't access the other ttys(f1, f2, etc). But it's no biggie since I can use my desktop machine for evaluating the expression if no cache is available.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
  - #1530, Fix how the PostgREST version is shown in the help text when the `.git` directory is not available - @monacoremo
 
+### Changed
+
+ - Docker images are now built from scratch based on a static build of the PostgREST executable.
```
steve-chavez, line 16:
```suggestion
 - #1535, Docker images are now built from scratch based on a static build of the PostgREST executable.
```
Also, maybe mention that the image is optimized now(from 30 mb to 4mb).

It'll be good to merge after this.

steve-chavez:
> But I'd submit that in a separate PR, this one is getting a bit out of hand :-)

I agree, the PR got a bit all over the place. Would be nice to get those tests Nixified later :).

Note: already edited the current docker hub description to the previous Dockerfile: https://hub.docker.com/r/postgrest/postgrest.

.circleci/config.yml
```diff
@@ -367,82 +190,75 @@ jobs:
...
+          command: nix-env -f default.nix -iA tests
+      # Our utility scripts use cabal to run the tests, which will pick up all
+      # the libraries that are built with Nix. So this should only ever build
+      # the PostgREST itself and be reasonably quick. We accelerate that part
+      # further by caching `~/.cabal` and `dist-newstyle`.
+      - restore_cache:
+          keys:
+            - nix-test-{{ checksum "postgrest.cabal" }}-{{ checksum "nix/nixpkgs-version.nix" }}
+            - nix-test-{{ checksum "postgrest.cabal" }}
+            - nix-test
```
steve-chavez, line 203: These ones slipped me by. Are they correct? They look redundant.

monacoremo: @steve-chavez Ah, yes, we missed to discuss that. These additional keys should help to make the test runs a bit faster using the partial cache restore feature of circleci (https://circleci.com/docs/2.0/caching/#restoring-cache). The keys are prefix matched in order. If the more  specific key is not found, it will try the next one and take the newest cache that matches the prefix.

In line 226 we append the current time to the caching key, so a new cache will always be saved (it should never become very large, as all dependencies are pre built by Nix). If the nix pin or cabal file changes, the build will not have to start completely from scratch, it will use the latest cache which is hopefully still a bit similar and accelerare the process.

steve-chavez: @monacoremo Ah, got it. I didn't know that one. Also missed the `epoch` cache key.

Could you add those explanations as comments on the circle config? Maybe on #1538.

steve-chavez: Ah, got it. I didn't know that one. Also missed the epoch cache key. Thanks for the explanation!

Could you add comments with the explanations on the circle config? Maybe on #1538.

PR #1538, opened 2020-05-26
Nixify io and memory tests

monacoremo:
As discussed in #1535, this integrates the io and memory tests with the Nix setup.

A few notes:
* The io and memory scripts are now independent from stack and take the postgrest executable from the PATH (also reflected in the stack CI jobs)
* This PR touches quite a few files because the db URI needed to be parameterized (the `with_tmp_db` will never be in the same location) - otherwise I tried to keep the changes minimal
* The tests still feels a bit hacky/brittle - the nmap dependency could probably be removed (its process not being properly terminated currently), as could the dependency on a hardcoded port. Shellcheck throws many warnings on the two scripts, but I left as much as possible as it was for this PR :-)
* The Nix test scripts now depend on the final executables (both a static and profiled build). This is good on the one hand, as that makes the tests truly end-to-end. On the other hand, it would make the nix-shell startup times very large if we included the tests by default. So I implemented flags to enable the more 'expensive' modules and also reflected that in the README

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -17,15 +17,17 @@ artifacts from the `/nix/store`, you can run `nix-collect-garbage`.
 To build PostgREST from your local checkout of the repository, run:
 
 ```bash
-nix-build --attr postgrest
+nix-build --attr postgrestPackage
 
 ```
 
 This will create a `result` directory that contains the PostgREST binary at
 `result/bin/postgrest`. The `--attr` parameter (or short: `-A`) tells Nix to
-build the `postgrest` attribute from the Nix expression it finds in our
+build the `postgrestPackage` attribute from the Nix expression it finds in our
 `default.nix` (see below for details). Nix will take care of getting the right
 GHC version and all the build dependencies.
+
+## Binary cache
 
 We recommend that you use the PostgREST binary cache on
 [cachix](https://cachix.org/):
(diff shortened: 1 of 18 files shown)


nix/README.md
```diff
@@ -66,17 +68,31 @@ The PostgREST utilities available in `nix-shell` all have names that begin with
...
-postgrest-test-spec                    postgrest-test-spec-postgresql-9.5
-postgrest-test-spec-all                postgrest-test-spec-postgresql-9.6
+postgrest-lint         postgrest-style        postgrest-style-check
 
 [nix-shell]$
 
 ```
 
+Some additional modules like `tests`, `docker` and `release` have large
+dependencies that would need to be built before the shell becomes available,
```
steve-chavez, line 78: Hmm.. couldn't we have the end-to-end tests(running io-tests.sh and memory-tests.sh) in another nix file(like `io-tests.nix`) and avoid building the static and profiled builds on nix-shell entrance?

I was really liking the DX of going inside nix-shell and being able to do `postgrest-test-<tab>`.

steve-chavez: Then we could do `nix-shell --arg io-tests true` in case we need to run the io tests. I think this would be better because those tests aren't run that frequently.

monacoremo: You're right, that would be much nicer! Didn't think about that.

With the new commit the spec tests are available in the nix-shell by default and the other tests can be enabled with flags!

steve-chavez: @monacoremo Thanks! Much better now.

Just something I noticed. Looks like the nix memory tests consume less memory than the stack-test-memory.

nix-tests:

```
Running memory usage tests..
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(9,075,584 bytes) is less than 13M
ok 2 - POST /leak?columns=blob: with a json key of 1M the memory usage(9,193,768 bytes) is less than 13M
```
stack-memory-test:
```
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(12,122,480 bytes) is less than 13M
ok 2 - POST /leak?columns=blob: with a json key of 1M the memory usage(12,257,056 bytes) is less than 13M
```

Not sure why. But it's good nonetheless, the stack memory tests used to break from time to time.

test/io-tests.sh
```diff
@@ -231,8 +234,10 @@ readSecretFromFile ascii.b64 'Base64 (ASCII)'
 readSecretFromFile utf8.b64 'Base64 (UTF-8)'
 readSecretFromFile binary.b64 'Base64 (binary)'
 
-readDbUriFromFile uri.noeol "(no EOL)"
-readDbUriFromFile uri.txt "(EOL)"
+eol=$'\x0a'
```
steve-chavez, line 237: Nice. Thanks for taking the time to refactor these tests! Comments are really good too.

test/memory-tests.sh
```diff
@@ -7,14 +18,9 @@ ko(){ result 'not ok' "- $1"; failedTests=$(( $failedTests + 1 )); }
 
 pgrPort=49421
 
-pgrStopAll(){ pkill -f "$(stack path --profile --local-install-root)/bin/postgrest"; }
-
-pgrStart(){ stack exec --profile -- postgrest test/memory-tests/config +RTS -p -h >/dev/null & pgrPID="$!"; }
+pgrStart(){ postgrest test/memory-tests/config +RTS -p -h > /dev/null & pgrPID="$!"; }
```
steve-chavez, line 21: Using the binary is much better. I remember I had some issues with stack there.

steve-chavez:
Looks good to merge now!

> The tests still feels a bit hacky/brittle - the nmap dependency could probably be removed (its process not being properly terminated currently), as could the dependency on a hardcoded port. Shellcheck throws many warnings on the two scripts, but I left as much as possible as it was for this PR :-)

The tests are definitely hacky, never took the time to run shellcheck on them. It'd be great if you could further improve them on a later PR.

nix/README.md
monacoremo: @steve-chavez Interesting... I don't have much experience with profiling haskell, might it be that some libraries are not included in the nix profiled build? Or different optimization levels?

Issue #1525, opened 2020-05-15
Feature Request: Set HTTP response code and body without needing to throw an exception.


W1M0R:
Currently, in order to set the HTTP response code and response body, a user-defined function must throw an exception. When a function throws an exception, a rollback is triggered, undoing everything that the function did. If a function wants to log data into a table, i.e. let's say failed login attempts, before returning with a HTTP 401/403 status, then the rollback will prevent the failed login attempt to be captured.

If it were possible to set the HTTP status and body without needing to throw an exception, then we would be able to log failed login attempts, while still communicating that an error took place.

The workarounds are unpleasant to work with or rely on system-level changes, i.e. using dblink, relying on pg 11/12 procedures.

@steve-chavez pointed out that maybe a SET LOCAL response.status = 400 could be offered.

Further reading: 
https://github.com/PostgREST/postgrest/issues/917
https://github.com/PostgREST/postgrest/pull/1435
https://github.com/PostgREST/postgrest/pull/986

steve-chavez:
I see the disadvantages about `RAISE`. I don't see why `response.status` cannot be offered as an alternative. It also looks more consistent with our other conventions. So I think it can be added.

A login function to return the 4xx status and the body could be like:

```sql
create or replace function signin(email text, pass text) returns json as $$
begin
  if email = 'john@doe.com' and pass = 'john' then
    return '{"id": 1}';
  else
    perform set_config('response.status', '403', true);
    return json_build_object('message', 'invalid user or password');
  end if;
end;
$$ language plpgsql;
```

W1M0R:
That looks perfect to me.

That solution will also satisfy the use case where a Hasura Action calls a PostgREST endpoint, e.g. /rpc/signin, but in order for Hasura to understand and report the error, it expects a specific response payload, e.g:

```json
{
  "message": "The error text",
  "code": 1212
}
```

grantwparks:
I'm trying to use this to get HTTP 302 response but it keeps yielding 200 OK.  I'm using Docker image postgrest/postgrest:v7.0.1.  Are there any special rules as to how to return from the function?

wolfgangwalther:
This feature is not included in 7.0.1, yet. You'll have to use one of the nightly builds to use it for now.

PR #1541, opened 2020-06-02
Allow http status override through response.status guc

steve-chavez:
Fixes https://github.com/PostgREST/postgrest/issues/1525.

A quick feature for this new batch of enhancements!

I think this should finish the [HTTP Logic](http://postgrest.org/en/v7.0.0/api.html#http-logic) features. I should turn that on its own page, the API reference is getting big.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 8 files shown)


steve-chavez:
Docs added in https://github.com/PostgREST/postgrest-docs/pull/329.

Preview: https://postgrest.org/en/preview/api.html#setting-response-status-code.

Issue #1011, opened 2017-11-10
PostgREST + PgBouncer - How to use different pool methods in PostgREST?


karrapetta:
Hi,

We're using PostgREST with PgBouncer in our project.

PgBouncer pooling supports Session, Transaction or Statement.  (pgbouncer.github.io/usage.html)

We tested only changing the pool mode in pgBouncer to Transaction or Statement, and it occurred the error bellow:
{"hint":null,"details":null,"code":"42P05","message":"prepared statement "0" already exists"}

We need to know how to use transaction or statement pooling in PostgREST, due to the number of requests we made.

Could anyone help us?

Thanks.
Rodrigo C Souza

ruslantalpa:
what are you using pgbouncer for?

begriffs:
It looks like the hasql-postgres library may have a counter it uses to name prepared statements, and that counter isn't getting incremented? https://github.com/nikita-volkov/hasql-postgres/blob/5c815ec13655b7d49b3d727ab4303ea41f18f7b1/library/Hasql/Postgres/Session/Execution.hs#L55

Maybe it's the way we're calling the library, or an interaction with connection pooling.

@nikita-volkov can you advise?

nikita-volkov:
Are you sure you're using "hasql-postgres"? That library was deprecated ages ago in favor of just "hasql". But yes, we do name the prepared statements the same way in the present versions.

It's not clear to me what the issue here is though.

nikita-volkov:
Are you sure you're using "hasql-postgres"? That library was deprecated ages ago in favor of just "hasql". But yes, we do name the prepared statements the same way in the present versions.

It's not clear to me what the issue here is though.

guiuprado:
Hi everyone.
Some good reason it's in production cases with or without cluster or citusdata, the pgbouncer or other pool manger service can help to optimize the use of active and idles connections. Mostly cases when several applications with pool enabled tend to access the PostgresSQL not only the PostgrREST. Actually we currently use exactly this way. So to both work together we need to set the pgbouncer mode to session. The session is less optimized method and tends to maintain the connection open until the caller close or timeout is hit. In critical environment it's not allowed.
Thanks for great job!! PostgREST is amazing. Thanks.

ruslantalpa:
if you have a single database server then i don't see why you need pgbouncer.
PostgREST already maintains it's own connection pool (connections are being reused).

guiuprado:
Hi everyone;

[...]

PR #1542, opened 2020-06-10
Allow schema cache reloading with NOTIFY

### LINKED ISSUE #1512: PostgREST watch mode
The idea is to provide a `watch: true` config option that reloads the schema cache without the need for restarting or sending a SIGUSR1 to pgrst from another process. This would make it more ergonomic for development.

Since our schema cache queries are fast (take 400 ms on a big schema)  I think we can just re-run the queries on every request. It could be possible that some requests respond with a 503 while pgrst is reloading. In this case the request would have to be retried. IIRC postgrest sends a message asking the user to retry as well. So it seems acceptable to behave this way.

watch would print a big warning telling it should not be used for production.

steve-chavez:
Fixes #1512

Helps on environments where you can't send unix signals(Windows, managed
containers). Also provides better UX for schema reloads - NOTIFY
can be send from pg clients(psql, pgadmin).

By default, `NOTIFY pgrst` - with no payload - should be done to reload
the schema cache. Notifications with a payload will be ignored.
The channel name can be configured with `db-channel`.

~~There's limitation for now, if the LISTEN connection is lost the
connection won't be retried.~~

**Edit**: The LISTEN connection is retried if lost now. With exponential backoff, same way as the pool.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -267,6 +267,8 @@ workflows:
           filters:
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
+          requires:
+            - nix-build
       - release:
           requires:
             - style-check
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
  - #1525, Allow http status override through response.status guc - @steve-chavez
+ - #1512, Allow schema cache reloading with NOTIFY - @steve-chavez
 
 ### Fixed
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -89,8 +89,8 @@ connectionWorker mainTid pool schemas refDbStructure refIsWorkerOn (dbChannelEna
       atomicWriteIORef refDbStructure Nothing
       putStrLn ("Attempting to connect to the database..." :: Text)
       connected <- connectionStatus pool
-      when dbChannelEnabled $ -- If this is not done, the putMVar will lock the thread since the MVar will never be consumed with a takeMVar
-        putMVar mvarConnectionStatus connected
+      when dbChannelEnabled $
+        void $ tryPutMVar mvarConnectionStatus connected -- tryPutMVar doesn't lock the thread. It should always succeed since the worker is the only producer.
       case connected of
         FatalConnectionError reason -> hPutStrLn stderr reason >> killThread mainTid -- Fatal error when connecting
         NotConnected                -> return ()                                     -- Unreachable because connectionStatus will keep trying to connect
@@ -227,7 +227,7 @@ main = do
   -- a 'Connection' or a 'ConnectionError'. Does not throw.
   pool <- P.acquire (configPool conf, configPoolTimeout' conf, dbUri)
 
-  -- No connection to the db at first. This is only used if the dbChannelEnabled is true(LISTENer enabled).
+  -- No connection to the db at first. This is only used if the dbChannelEnabled is true(listener enabled).
   mvarConnectionStatus <- newEmptyMVar
 
   -- To be filled in by connectionWorker
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -234,7 +234,7 @@ main = do
   -- a 'Connection' or a 'ConnectionError'. Does not throw.
   pool <- P.acquire (configPool conf, configPoolTimeout' conf, dbUri)
 
-  -- No connection to the db at first. This is only used if the dbChannelEnabled is true(listener enabled).
+  -- Used to sync the listener with the connectionWorker. No connection for the listener at first. Only used if dbChannelEnabled=true.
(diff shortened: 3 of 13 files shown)


dwagin:
How to disable this feature?

steve-chavez:
Hey @dwagin. Right now I'm including it by default with no option to disable it.
But I could add a `db-channel-enable`(default true) config for disabling it, would that work for you?

steve-chavez:
@dwagin I've added the `db-channel-enabled` config option! `db-channel-enabled=false` can be done for no LISTEN.

dwagin:
> @dwagin I've added the `db-channel-enabled` config option! `db-channel-enabled=false` can be done for no LISTEN.

This is very good, because LISTEN/NOTIFY doesn't work with PgBouncer and additional extra connection is not needed in this situation.

nerfpops:
If this is something that'll potentially cause trouble, like with PgBouncer, wouldn't it make sense to have db-channel-enabled=false as default?

steve-chavez:
@dwagin Could you share how are you using PgBouncer?

On [SQL feature map for pooling modes](https://www.pgbouncer.org/features.html#sql-feature-map-for-pooling-modes), they mention that LISTEN/NOTIFY does work with Session pooling(not transaction pooling). And https://github.com/PostgREST/postgrest/issues/1011 left me with the impression that transaction pooling is not working with PostgREST now(only session pooling works), or did you managed to make that work?

steve-chavez:
@nerfpops It was tempting to enable it by default and just tell users to run `NOTIFY pgrst`, but with the ideas in https://github.com/PostgREST/postgrest/issues/1512#issuecomment-642955889 I think we can instead have it as a recommended config in the docs(the EVENT TRIGGER would be an extra step anyway). So yes, I can disable it by default to be on the safe side.

dwagin:
@steve-chavez Session pooling doesn't make sense, we use transaction pooling (with small hasql patch). 


[patch-hasql-1.4.2_library_Hasql_Private_IO.hs](https://github.com/PostgREST/postgrest/files/4799408/patch-hasql-1.4.2_library_Hasql_Private_IO.hs.txt)

steve-chavez:
An update on the implementation. The LISTEN connection is now retried if lost.
Also, I've added a debounce of 1 second on the NOTIFY reloading. This is necessary for migrations as mentioned in https://github.com/PostgREST/postgrest/issues/1512#issuecomment-643152662.

Without the debounce, doing this in bash:

```bash
for i in {1..20}; do psql postgrest_test -c "NOTIFY pgrst"; done
```
Results in 20 schema reloads. The debounce reduces that to one.

Weirdly, I'm not able to reproduce the bad 20 schema reloads with just `psql`:
```sql
DO $$ BEGIN FOR counter IN 1..20 LOOP NOTIFY pgrst; END LOOP; END$$;
```

That always results in one NOTIFY sent, not sure what could be wrong there. But to be on the safe sade, I'll add the integration tests in bash.

steve-chavez:
Adding the bash integration tests to CI is proving to be a lot of work. The `with_tmp_db` shell script would have to be modified and we'd have to interact with `pg_ctl` directly(stop/start). So I've just manual tested that the reloading is good and I've ensured the SIGUSR1 is working as usual.

On the introduction of SIGHUP(https://github.com/PostgREST/postgrest/pull/869) we also did manual bash tests. I have a general idea on how to automate this, but I'll open another issue for that.

So I think this is good to merge now :rocket:

wolfgangwalther:
> Weirdly, I'm not able to reproduce the bad 20 schema reloads with just `psql`:
> 
> ```sql
> DO $$ BEGIN FOR counter IN 1..20 LOOP NOTIFY pgrst; END LOOP; END$$;
> ```
> 
> That always results in one NOTIFY sent, not sure what could be wrong there. But to be on the safe sade, I'll add the integration tests in bash.

I'm pretty sure this is because all those `NOTIFY`s are executed in one transaction - there will be only one notification with the same payload sent per transaction. Citing the pg docs:

> If the same channel name is signaled multiple times from the same transaction with identical payload strings, the database server can decide to deliver a single notification only.

If this was guaranteed (the wording doesn't seem like it is), we wouldn't need the debounce, because a schema migration could just be run in a single transaction. I guess in most cases it would be, anyways.

steve-chavez:
@wolfgangwalther Nice catch! I missed that one.

(For reference: https://www.postgresql.org/docs/current/sql-notify.html. ctrl +f `If the same channel name is signaled multiple times`)

Since the debounce was easy to set up - it didn't require an extra lib and just a few lines of code - I guess we can keep it to be on the safe side. But the pg doc reference could be noted as a comment on the code.

laurenceisla:
@dwagin  @nerfpops  Just a heads up: `database-channel-enabled` setting is now true by default since https://github.com/PostgREST/postgrest/commit/214a92f20779db656f3ee9fb6fd68914501ad40a, and needs to be explicitly set to false when using PgBouncer.

PR #1544, opened 2020-06-15
Enable config file re-reading with SIGUSR2

steve-chavez:
Continuing the work on https://github.com/PostgREST/postgrest/pull/1289. Should also fix https://github.com/PostgREST/postgrest/issues/1119.

Considerations:
- SIGUSR2 only re-reads the config, but doesn't reload the schema cache. This is because `jwt-secret` and other settings don't deal with the db. Reloading the cache on a `jwt-secret` update would be wasteful.
- From the above, if a `db-schema` changes, a SIGUSR2 + SIGUSR1 would be needed for a correct schema cache reload.
- If the config file path has changed, the reload will fail(and show an error) but postgrest will continue to work. The reloading will be done in a different thread. This is probably not an issue in production since services have an absolute path defined for the config.
- If the config file has an invalid config option(like invalid roleClaimKey) the reload will also fail and pgrst will continue to work. In this case the user would have to fix the config for the reload to be successful.

Only these settings will be re-read:

- db-schema
- db-anon-role
- server-proxy-uri
- jwt-secret
- secret-is-base64
- jwt-aud
- max-rows
- pre-request
- role-claim-key
- db-extra-search-path
- app.settings.*

These setting won't be re-read(would require a `restart`):

- db-uri
- db-pool
- db-pool-timeout
- server-host
- server-port
- server-unix-socket
- server-unix-socket-mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
  - #1525, Allow http status override through response.status guc - @steve-chavez
  - #1512, Allow schema cache reloading with NOTIFY - @steve-chavez
+ - #1119, Allow config file reloading with SIGUSR2 - @steve-chavez
 
 ### Fixed
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -3,7 +3,6 @@
 module Main where
 
 import qualified Data.ByteString            as BS
-import qualified Data.ByteString.Base64     as B64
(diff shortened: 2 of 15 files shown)


steve-chavez:
I think this is done!

I've added bash tests for `app.settings.*`, `jwt-secret`, `db-schema` reloading with SIGUSR2. Those are really the main use case for SIGUSR2. Left testing other config options reload as TODOs.

PR #1551, opened 2020-07-03
add test cases for singular minimal

wolfgangwalther:
While debugging something that ultimately turned out to be a bug in my app, I added some test-cases. They already pass, but I guess it makes sense to add those tests cases for `DELETE` and `PATCH`, because one already exists for `POST` as well.

Also fixed another one, probably just a copy-paste mistake.


steve-chavez:
Nice. More tests are always good :+1:

Issue #1469, opened 2020-04-02
Incorrect behavior of header `Prefer: params=single-object` on overloaded function


fortyanov:
* PostgreSQL version: 12.1
* PostgREST version: 6.0.2 (713b214)
* Operating system: Ubuntu 19.04

### Incorrect behavior of header ```Prefer: params=single-object```

We have got overridden function 
```sql 
CREATE TYPE test_type AS
(
    "p1" text,
    "p2" text
);

CREATE OR REPLACE FUNCTION test(a text, b text) RETURNS test_type
    LANGUAGE PLpgSQL
AS
$_$
BEGIN
    RETURN ROW (a, b);
END
$_$;

CREATE OR REPLACE FUNCTION test(json) RETURNS json
    LANGUAGE PLpgSQL
AS
$_$
BEGIN
    RETURN $1;
END
$_$;
```
send request with header 
`Prefer: params=single-object`

```bash 
curl --location --request POST 'http://127.0.0.1:3001/rpc/test' \
--header 'Content-Type: application/json' \
--header 'Prefer: params=single-object' \
--data-raw '{
  "a": "a",
  "b": "b"
}'
```
returns incorrect response
```
[
  {
    "test": {
      "a": "a",
      "b": "b"
    }
  }
]
```
after dropping  first function
```sql
DROP FUNCTION test(text, text);
```
and repeat request i have got correct response
```
{
  "a": "a",
  "b": "b"
}
```

steve-chavez:
Hmm.. I cannot reproduce. I get:

```json
[{"test":{
  "a": "a",
  "b": "b"
}}]
```

Consistently, on both 6.0.2 and master version. Even after `DROP FUNCTION test(text, text);`.

Not saying the above output is right though, I think the top key `test` shouldn't appear in the json.

dwagin:
> Hmm.. I cannot reproduce. I get:
> 
> ```json
> [{"test":{
>   "a": "a",
>   "b": "b"
> }}]
> ```
This is the bug

> 
> Consistently, on both 6.0.2 and master version. Even after `DROP FUNCTION test(text, text);`.
> 
> Not saying the above output is right though, I think the top key `test` shouldn't appear in the json.

The problem is that the type of return value is not determined correctly:
First function return composite type and returnsScalar must be false
Second function return scalar type and returnsScalar must be true

When both functions are present in the database call second function returnsScalar is false, but must be true.
When only second function are present in the database all fine.

[...]

Issue #1405, opened 2019-11-07
OpenAPI generated specs do not support type: array, and use format: ARRAY instead


synerscope:
When generating an OpenAPI spec, array-valued types are annotated as having *type: string* with *format: array*. The OpenAPI specification supports using *type: array* with an *items* subkey to set the type of the items in the array.

This breaks my code when I generate (using openapi-generator) a client that needs to parse array return values, as it postgrest itself correctly returns an array typed object, but the spec indicates it should be interpreted as a string.  I have validated that the generated client code works if I patch the generated openapi spec before generation and replace all format: ARRAY, type: string entries with type: array and then items: - string.

I believe this is the code responsible, but do not know how to get the internal type of the items contained in the array to make the fix.

https://github.com/PostgREST/postgrest/blob/9847e60dcace57b9ae2f4fbba4ba9bcabceae4c6/src/PostgREST/OpenAPI.hs#L81-L87

steve-chavez:
I've fiddled with the [Data.Swagger](http://hackage.haskell.org/package/swagger2-2.4/docs/Data-Swagger.html#t:SwaggerItems) types and this works for getting the string array items:

```patch
diff --git a/src/PostgREST/OpenAPI.hs b/src/PostgREST/OpenAPI.hs
index 1c3c875..97658e7 100644
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -85,6 +85,9 @@ makeProperty pks c = (colName c, Inline s)
         & format ?~ colType c
         & maxLength .~ (fromIntegral <$> colMaxLen c)
         & type_ ?~ toSwaggerType (colType c)
+        & items ?~ (SwaggerItemsObject $ Inline (mempty
+          & type_ ?~ SwaggerString
+        ))
 
 makeProcSchema :: ProcDescription -> Schema
 makeProcSchema pd =
```

That logic should probably be moved to the `toSwaggerType` function. For starters that function could only pattern match the `text[]` type and output something similar to the above.

Open a PR if you'd like to give it a shot. Then we'd see how to add a test.

ntgx:
[...]

Issue #1222, opened 2018-12-13
OpenApi not showing GET method for immutable functions


apanagio:
### Environment

* PostgreSQL version: "PostgreSQL 11.1 (Ubuntu 11.1-1.pgdg18.10+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 8.2.0-7ubuntu1) 8.2.0, 64-bit"
* PostgREST version: 5.1.0 (3cef4b7)
* Operating system: ubuntu linux 18.10

### Description of issue

I have an immutable function that i can call with the get method. However the GET method doesn't show up within the openapi specification (using swagger-ui)

here is the declaration of the function:
```
CREATE OR REPLACE FUNCTION public.whoami(
	)
    RETURNS name
    LANGUAGE 'sql'

    COST 100
    IMMUTABLE STRICT 
AS $BODY$select current_user$BODY$;

maxwell-clarke-intranel:
I am also having this issue

steve-chavez:
Thanks for the report. Actually GET is not being showed for any function. [Here](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/OpenAPI.hs#L237) is why RPCs are only showing POST, we would need to add GET in a way similar to [this](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/OpenAPI.hs#L215).

PR #1552, opened 2020-07-07
fix calling rpc with variadic argument

### LINKED ISSUE #1470: VARIADIC (Functions with variable numbers of arguments) is not supported with any array type
### Environment

* PostgreSQL version: 11
* PostgREST version: 6.0.2
* Operating system: Ubuntu 18.0.4

### Description of issue
URL: http://127.0.0.1:3000/rpc/test?col1=val1&col2=val2
Error Details: You might need to add explicit type casts.","details":null,"code":"42883","message":"function public.callback(col1 => text, col2 => text) does not exist"}
Function Header: CREATE OR REPLACE FUNCTION public.test(VARIADIC data text[])

wolfgangwalther:
Still todo: Update docs

Adding the changelog item is hard on the first commit - because you have to guess the number of the PR. Let's see whether I guessed right.

Resolves #1470

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -149,14 +149,16 @@ decodeProcs =
     parseArgs = mapMaybe parseArg . filter (not . isPrefixOf "OUT" . toS) . map strip . split (==',')
 
     parseArg :: Text -> Maybe PgArg
-    parseArg a =
-      let arg = lastDef "" $ splitOn "INOUT " a
-          (body, def) = breakOn " DEFAULT " arg
+    parseArg arg =
+      let isVariadic = isPrefixOf "VARIADIC " $ toS arg
+          argNoInout = lastDef "" $ splitOn "INOUT " arg
+          argNoVariadic = lastDef "" $ splitOn "VARIADIC " argNoInout
+          (body, def) = breakOn " DEFAULT " argNoVariadic
           (name, typ) = breakOn " " body in
       if T.null typ
          then Nothing
          else Just $
-           PgArg (dropAround (== '"') name) (strip typ) (T.null def)
+           PgArg (dropAround (== '"') name) (strip typ) (T.null def) isVariadic
 
     parseRetType :: Text -> Text -> Bool -> Char -> RetType
     parseRetType schema name isSetOf typ
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -151,9 +151,9 @@ decodeProcs =
     parseArg :: Text -> Maybe PgArg
     parseArg arg =
       let isVariadic = isPrefixOf "VARIADIC " $ toS arg
-          argNoInout = lastDef "" $ splitOn "INOUT " arg
-          argNoVariadic = lastDef "" $ splitOn "VARIADIC " argNoInout
-          (body, def) = breakOn " DEFAULT " argNoVariadic
+          -- argmode can be IN, OUT, INOUT, or VARIADIC
+          argNoMode = lastDef "" $ splitOn (if isVariadic then "VARIADIC " else "INOUT ") arg
+          (body, def) = breakOn " DEFAULT " argNoMode
           (name, typ) = breakOn " " body in
       if T.null typ
          then Nothing
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -137,15 +137,20 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
           unwords [
             normalizedBody <> ",",
             "pgrst_args AS (",
-              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (\a -> " " <> pgaType a) <> ")",
+              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (const "") (\a -> " " <> pgaType a) <> ")",
             ")"]
          , if paramsAsMultipleObjects
-             then fmtArgs (\a -> " := pgrst_args." <> pgFmtIdent (pgaName a))
-             else fmtArgs (\a -> " := (SELECT " <> pgFmtIdent (pgaName a) <> " FROM pgrst_args LIMIT 1)")
+             then fmtArgs varadicPrefix (\a -> " := pgrst_args." <> pgFmtIdent (pgaName a))
+             else fmtArgs varadicPrefix (\a -> " := (SELECT " <> pgFmtIdent (pgaName a) <> " FROM pgrst_args LIMIT 1)")
         )
 
-    fmtArgs :: (PgArg -> SqlFragment) -> SqlFragment
-    fmtArgs argFrag = intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> argFrag a) <$> pgArgs)
+    fmtArgs :: (PgArg -> SqlFragment) -> (PgArg -> SqlFragment) -> SqlFragment
+    fmtArgs argFragPre argFragSuf = intercalate ", " ((\a -> argFragPre a <> pgFmtIdent (pgaName a) <> argFragSuf a) <$> pgArgs)
+
+    varadicPrefix :: PgArg -> SqlFragment
+    varadicPrefix a
+      | pgaVar a  = "VARIADIC "
+      | otherwise = ""
 
     sourceBody :: SqlFragment
     sourceBody
--- a/src/PostgREST/QueryBuilder.hs
+++ b/src/PostgREST/QueryBuilder.hs
@@ -137,7 +137,7 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
           unwords [
             normalizedBody <> ",",
             "pgrst_args AS (",
-              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (const "") (\a -> " " <> pgaType a) <> ")",
+              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (const mempty) (\a -> " " <> pgaType a) <> ")",
             ")"]
          , if paramsAsMultipleObjects
              then fmtArgs varadicPrefix (\a -> " := pgrst_args." <> pgFmtIdent (pgaName a))
@@ -148,9 +148,7 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
     fmtArgs argFragPre argFragSuf = intercalate ", " ((\a -> argFragPre a <> pgFmtIdent (pgaName a) <> argFragSuf a) <$> pgArgs)
 
     varadicPrefix :: PgArg -> SqlFragment
-    varadicPrefix a
-      | pgaVar a  = "VARIADIC "
-      | otherwise = ""
+    varadicPrefix a = if pgaVar a then "VARIADIC " else mempty
 
     sourceBody :: SqlFragment
     sourceBody
--- a/test/Feature/RpcSpec.hs
+++ b/test/Feature/RpcSpec.hs
@@ -394,11 +394,15 @@ spec actualPgVersion =
         get "/rpc/many_inout_params?num=1&str=two&b=false" `shouldRespondWith`
           [json| [{"num":1,"str":"two","b":false}]|] { matchHeaders = [matchContentTypeJson] }
 
-    it "works with variadic parameter" $
+    it "works with variadic parameter" $ do
       post "/rpc/variadic_param"
-          [json| { "v": ["hello"] } |]
+          [json| { "v": ["hi", "hello", "there"] } |]
         `shouldRespondWith`
-          [json|"Hi"|]
+          [json|["hi", "hello", "there"]|]
+          { matchHeaders = [matchContentTypeJson] }
+      get "/rpc/variadic_param?v=%7Bhi,hello,there%7D"
+        `shouldRespondWith`
+          [json|["hi", "hello", "there"]|]
           { matchHeaders = [matchContentTypeJson] }
 
     it "can handle procs with args that have a DEFAULT value" $ do
--- a/test/Feature/RpcSpec.hs
+++ b/test/Feature/RpcSpec.hs
@@ -396,6 +396,47 @@ spec actualPgVersion =
               [json|"string"|]
               { matchHeaders = [matchContentTypeJson] }
 
+    context "repeated params as array" $ do
+      context "in query string for GET" $ do
+        it "without single item overload" $ do
+          get "/rpc/array_argument?arg=a&arg=b&arg=c"
+            `shouldRespondWith`
+              [json|["a", "b", "c"]|]
+          get "/rpc/array_argument?arg=a"
+            `shouldRespondWith` 400
+
+        it "with single item overload" $ do
+          get "/rpc/generic_argument?arg=1&arg=2&arg=3"
+            `shouldRespondWith`
+              [json|[1, 2, 3]|]
+          get "/rpc/generic_argument?arg=1"
+            `shouldRespondWith`
+              [json|[1]|]
+
+      context "in x-www-form-urlencoded body for POST" $ do
+        it "without single item overload" $ do
+          request methodPost "/rpc/array_argument"
+              [("Content-Type", "application/x-www-form-urlencoded")]
+              "arg=a&arg=b&arg=c"
+            `shouldRespondWith`
+              [json|["a", "b", "c"]|]
+          request methodPost "/rpc/array_argument"
+              [("Content-Type", "application/x-www-form-urlencoded")]
+              "arg=a"
+            `shouldRespondWith` 400
+
+        it "with single item overload" $ do
+          request methodPost "/rpc/generic_argument"
+              [("Content-Type", "application/x-www-form-urlencoded")]
+              "arg=1&arg=2&arg=3"
+            `shouldRespondWith`
+              [json|[1, 2, 3]|]
+          request methodPost "/rpc/generic_argument"
+              [("Content-Type", "application/x-www-form-urlencoded")]
+              "arg=1"
+            `shouldRespondWith`
+              [json|[1]|]
+
     context "improper input" $ do
       it "rejects unknown content type even if payload is good" $ do
         request methodPost "/rpc/sayhello"
@@ -496,6 +537,18 @@ spec actualPgVersion =
         get "/rpc/variadic_param?v=%7Bhi,hello,there%7D"
           `shouldRespondWith`
             [json|["hi", "hello", "there"]|]
+            { matchHeaders = [matchContentTypeJson] }
+
+      it "works with repeated params" $
+        get "/rpc/variadic_param?v=hi&v=hello&v=there"
+          `shouldRespondWith`
+            [json|["hi", "hello", "there"]|]
+            { matchHeaders = [matchContentTypeJson] }
+
+      it "works with single param" $
+        get "/rpc/variadic_param?v=hi"
+          `shouldRespondWith`
+            [json|["hi"]|]
             { matchHeaders = [matchContentTypeJson] }
 
     it "can handle procs with args that have a DEFAULT value" $ do
--- a/test/Feature/RpcSpec.hs
+++ b/test/Feature/RpcSpec.hs
@@ -394,16 +394,28 @@ spec actualPgVersion =
         get "/rpc/many_inout_params?num=1&str=two&b=false" `shouldRespondWith`
           [json| [{"num":1,"str":"two","b":false}]|] { matchHeaders = [matchContentTypeJson] }
 
-    it "works with variadic parameter" $ do
-      post "/rpc/variadic_param"
-          [json| { "v": ["hi", "hello", "there"] } |]
-        `shouldRespondWith`
-          [json|["hi", "hello", "there"]|]
-          { matchHeaders = [matchContentTypeJson] }
-      get "/rpc/variadic_param?v=%7Bhi,hello,there%7D"
-        `shouldRespondWith`
-          [json|["hi", "hello", "there"]|]
-          { matchHeaders = [matchContentTypeJson] }
+    context "with variadic parameter" $ do
+      when (actualPgVersion < pgVersion100) $
+        it "works with POST (Postgres < 10)" $
+          post "/rpc/variadic_param"
+              [json| { "v": "{hi,hello,there}" } |]
+            `shouldRespondWith`
+              [json|["hi", "hello", "there"]|]
+              { matchHeaders = [matchContentTypeJson] }
+
+      when (actualPgVersion >= pgVersion100) $
+        it "works with POST (Postgres >= 10)" $
+          post "/rpc/variadic_param"
+              [json| { "v": ["hi", "hello", "there"] } |]
+            `shouldRespondWith`
+              [json|["hi", "hello", "there"]|]
+              { matchHeaders = [matchContentTypeJson] }
+
+      it "works with GET" $
+        get "/rpc/variadic_param?v=%7Bhi,hello,there%7D"
+          `shouldRespondWith`
+            [json|["hi", "hello", "there"]|]
+            { matchHeaders = [matchContentTypeJson] }
 
     it "can handle procs with args that have a DEFAULT value" $ do
       get "/rpc/many_inout_params?num=1&str=two" `shouldRespondWith`
--- a/test/Feature/RpcSpec.hs
+++ b/test/Feature/RpcSpec.hs
@@ -291,6 +291,20 @@ spec actualPgVersion =
             [json|"Hi"|]
             { matchHeaders = [matchContentTypeJson] }
 
+      it "works with inout argument" $
+        post "/rpc/inout_argument"
+            [json| { "arg": true } |]
+          `shouldRespondWith`
+            [json| true |]
+            { matchHeaders = [matchContentTypeJson] }
+
+      it "works with variadic arguments" $
+        post "/rpc/variadic_argument"
+            [json| { "v": ["hello"] } |]
+          `shouldRespondWith`
+            [json|"Hi"|]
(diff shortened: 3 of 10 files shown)


wolfgangwalther:
rebased to current master.

test/Feature/RpcSpec.hs
```diff
@@ -291,6 +291,20 @@ spec actualPgVersion =
             [json|"Hi"|]
             { matchHeaders = [matchContentTypeJson] }
 
+      it "works with inout argument" $
+        post "/rpc/inout_argument"
```
wolfgangwalther, line 295: I just realized that there are already `INOUT` tests in place. I guess I was searching for "argument" but those are labeled with "param", so I didn't find them. So this test can probably be removed.

wolfgangwalther:
So I had a look at the existing `INOUT` tests and I am confused.

This is what the test looks like:

```haskell
it "returns a scalar result when there is a single INOUT param" $
        get "/rpc/single_inout_param?num=2" `shouldRespondWith`
          [json|3|] { matchHeaders = [matchContentTypeJson] }
```

and this is the SQL for the function

```SQL
create function test.single_inout_param(INOUT num int) AS $$
  select num + 1;
$$ language sql;
```

Nowhere in the function definition the volatility is set to `STABLE` or `IMMUTABLE` - so according to the docs it should be marked as `VOLATILE` per default. Yet, the function is called with `GET` and not with `POST`.

I guess I am not understanding the PostgREST docs correctly, where it says:
> The API endpoint supports POST (and in some cases GET) to execute the function.
> [...]
> Stable and immutable functions can be called with the HTTP GET verb if desired.

So does the volatility actually have any effect on PostgREST? Or is it just that function that modify data will fail in the read-only transaction and that's it - independent of the volatility? Does it have an affect on the json schema exposed at the api root?

@steve-chavez Can you help my understanding here?

steve-chavez:
@wolfgangwalther Yes, the volatility is just a hint for the pg planner. For GET we call the function in a read only transaction, disregard the actual volatility and let pg err if the function modifies the database. For POST we do consider the volatility for the transaction.

https://github.com/PostgREST/postgrest/blob/343e41c51d6cad33383043abb4ef15dc54a32245/src/PostgREST/App.hs#L115-L121

(For reference: the idea came from https://github.com/PostgREST/postgrest/issues/328#issuecomment-322502897)

> Does it have an affect on the json schema exposed at the api root?

You mean if the volatility is considered for the openapi output to hint whether a function should be called with GET/POST? I believe this isn't being done now.

> I guess I am not understanding the PostgREST docs correctly, where it says:

You're right about the docs. Perhaps we can clarify and mention that GET enforces non-mutability with the read-only tx, despite the volatility.

wolfgangwalther:
Thanks for clearing that up for me and providing that background - makes a lot more sense now. Improving the docs in that regard might help others as well.

> You mean if the volatility is considered for the openapi output to hint whether a function should be called with GET/POST? I believe this isn't being done now.

Yes, it only shows POST I guess: #1222 #1151 

Now, of course I asked myself what kind of error will be thrown in case of calling a volatile proc with GET. I found this:

```SQL
    context "only for GET rpc" $ do
      it "should fail on mutating procs" $ do
        get "/rpc/callcounter" `shouldRespondWith` 500
        get "/rpc/setprojects?id_l=1&id_h=5&name=FreeBSD" `shouldRespondWith` 500
```

Instead of a generic 500, it would probably be better to return `405 Method Not Allowed`, which seems like it's designed for exactly that case. Do you think the error returned on "mutation inside the read-only transaction" can easily be caught and mapped to 405 instead?

steve-chavez:
>  Do you think the error returned on "mutation inside the read-only transaction" can easily be caught and mapped to 405 instead?

@wolfgangwalther I think it would be a matter of detecting the "25006 read_only_sql_transaction" pg error code in Errors.hs:
https://github.com/PostgREST/postgrest/blob/e8b4e3771cb773302941dde218fd5e96bec24170/src/PostgREST/Error.hs#L178-L187

(For reference https://www.postgresql.org/docs/12/errcodes-appendix.html)

I was thinking if there is a case where it wouldn't make sense to map a 25006 to a Method not allowed but couldn't come up with one. So I think 405 would be good.

src/PostgREST/DbStructure.hs
```diff
@@ -149,14 +149,16 @@ decodeProcs =
...
 
     parseArg :: Text -> Maybe PgArg
-    parseArg a =
-      let arg = lastDef "" $ splitOn "INOUT " a
-          (body, def) = breakOn " DEFAULT " arg
+    parseArg arg =
+      let isVariadic = isPrefixOf "VARIADIC " $ toS arg
+          argNoInout = lastDef "" $ splitOn "INOUT " arg
+          argNoVariadic = lastDef "" $ splitOn "VARIADIC " argNoInout
+          (body, def) = breakOn " DEFAULT " argNoVariadic
```
steve-chavez, line 156: Can be improved to:
```suggestion
          -- argmode can be IN, OUT, INOUT, or VARIADIC
          argNoMode = lastDef "" $ splitOn (if isVariadic then "VARIADIC " else "INOUT ") arg
          (body, def) = breakOn " DEFAULT " argNoMode
```
(Not possible to declare a parameter like `func(INOUT VARIADIC a..`, only one argmode at a time. Check https://www.postgresql.org/docs/current/sql-createfunction.html#id-1.9.3.67.6)

`pgaVar` could take a new type but I think is fine for now.

src/PostgREST/QueryBuilder.hs
```diff
@@ -137,15 +137,20 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
           unwords [
             normalizedBody <> ",",
             "pgrst_args AS (",
-              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (\a -> " " <> pgaType a) <> ")",
+              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (const "") (\a -> " " <> pgaType a) <> ")",
```
steve-chavez, line 140: Nitpick
```suggestion
              "SELECT * FROM json_to_recordset(" <> selectBody <> ") AS _(" <> fmtArgs (const mempty) (\a -> " " <> pgaType a) <> ")",
```

src/PostgREST/QueryBuilder.hs
```diff
@@ -137,15 +137,20 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
...
 
-    fmtArgs :: (PgArg -> SqlFragment) -> SqlFragment
-    fmtArgs argFrag = intercalate ", " ((\a -> pgFmtIdent (pgaName a) <> argFrag a) <$> pgArgs)
+    fmtArgs :: (PgArg -> SqlFragment) -> (PgArg -> SqlFragment) -> SqlFragment
+    fmtArgs argFragPre argFragSuf = intercalate ", " ((\a -> argFragPre a <> pgFmtIdent (pgaName a) <> argFragSuf a) <$> pgArgs)
+
+    varadicPrefix :: PgArg -> SqlFragment
+    varadicPrefix a
+      | pgaVar a  = "VARIADIC "
+      | otherwise = ""
```
steve-chavez, line 153: Another nitpick :--]
```suggestion
    varadicPrefix a = if pgaVar a then "VARIADIC " else mempty
```

test/fixtures/schema.sql
```diff
@@ -1072,6 +1072,11 @@ create function test.many_inout_params(INOUT num int, INOUT str text, INOUT b bo
   select num, str, b;
 $$ language sql;
 
+CREATE FUNCTION test.variadic_param(VARIADIC v TEXT[]) RETURNS text
+LANGUAGE SQL AS $_$
+  SELECT 'Hi'::text
+$_$;
```
steve-chavez, line 1078: More meaningful test what we can point users to:
```suggestion
CREATE FUNCTION test.variadic_param(VARIADIC v TEXT[]) RETURNS text[]
LANGUAGE SQL AS $_$
  SELECT $1
$_$;
```

test/Feature/RpcSpec.hs
```diff
@@ -394,6 +394,13 @@ spec actualPgVersion =
         get "/rpc/many_inout_params?num=1&str=two&b=false" `shouldRespondWith`
           [json| [{"num":1,"str":"two","b":false}]|] { matchHeaders = [matchContentTypeJson] }
 
+    it "works with variadic parameter" $
+      post "/rpc/variadic_param"
+          [json| { "v": ["hello"] } |]
+        `shouldRespondWith`
+          [json|"Hi"|]
+          { matchHeaders = [matchContentTypeJson] }
```
steve-chavez, line 402:
```suggestion
    it "works with variadic parameter" $ do
      post "/rpc/variadic_param"
          [json| { "v": ["hi", "hello", "there"] } |]
        `shouldRespondWith`
          [json|["hi", "hello", "there"]|]
          { matchHeaders = [matchContentTypeJson] }
      get "/rpc/variadic_param?v=%7Bhi,hello,there%7D"
        `shouldRespondWith`
          [json|["hi", "hello", "there"]|]
          { matchHeaders = [matchContentTypeJson] }
```

steve-chavez:
@wolfgangwalther Thought you were going to do the 405 logic here, but it's actually better on another PR. Check my above review.

wolfgangwalther:
Yeah, that 405 stuff wasnt really related to this here - just noticed it while working on the stuff.

But there might be some other code that I would add to this PR instead of creating a new one: I have the "multiple query params with the same name should make an array" in the pipeline as suggested in https://github.com/PostgREST/postgrest/issues/1470#issuecomment-653473143 and realized that this might have implications on how variadic handling (should) work(s) as well.

wolfgangwalther:
So as a first step, I extended the existing tests about RPC params a little bit, because I noticed passing various argument types via query string was not tested for, yet. The same thing about `application/x-www-form-urlencoded`, which was only tested for `Prefer params=single-object`, but not for regular function arguments. Then I added an array argument to those tests as well, because that was not included at all.

I noticed a couple of things - see my comments below.

test/Feature/RpcSpec.hs
```diff
@@ -274,21 +274,102 @@ spec actualPgVersion =
           { matchHeaders = [matchContentTypeJson] }
 
     context "proc argument types" $ do
-      it "accepts a variety of arguments" $
```
wolfgangwalther, line 277: This test seems to be completely broken in it's previous form. @steve-chavez can you try to reproduce the following?

On current master change e.g. `"double": 3.1` to `"double": "x.1"` for this test. I would expect an error along the lines of "invalid input syntax for double precision". But for me, the test still passes.

When I then change the `varied_arguments` function to return the json object (see `schema.sql`) with all argument values - it correctly throws that input syntax error.

I was not able to reproduce that with bare SQL, but apparently we are calling the RPC in a way where that wouldn't throw.

In this state this test is basically useless, because you could pass anything and wouldn't notice whether it's cast correctly - that's why I changed the whole test to return the arguments in a json object as well.

test/Feature/RpcSpec.hs
```diff
@@ -274,21 +274,102 @@ spec actualPgVersion =
...
+                "enum": "foo",
+                "arr": ["a", "b", "c"],
+                "integer": 43,
+                "json": {"some key": "some value"},
+                "jsonb": {"another key": [1, 2, "3"]}
+              } |]
+              { matchHeaders = [matchContentTypeJson] }
+
+      it "accepts a variety of arguments with GET" $
+        -- without JSON / JSONB here, because passing those via query string is useless - they just become a "json string" all the time
```
wolfgangwalther, line 339: So passing JSON / JSONB via query string does not seem to work. At least if I try to pass something that should match an object, it seems like this is just returned as a json string.

Looking at the tests below that ("parses xxx JSON as JSON"), where this is explicitely tested for POST, I wonder: Is this expected behaviour for GET / query string (composing json objects in a query string is really ugly...) or is this to be considered a bug?

test/Feature/RpcSpec.hs
```diff
@@ -274,21 +274,102 @@ spec actualPgVersion =
...
+                "integer": 43,
+                "json": {},
+                "jsonb": {}
+              } |]
+            { matchHeaders = [matchContentTypeJson] }
+
+      it "accepts a variety of arguments from an html form" $
+        request methodPost "/rpc/varied_arguments"
+            [("Content-Type", "application/x-www-form-urlencoded")]
+            "double=3.1&varchar=hello&boolean=true&date=20190101&money=0&enum=foo&arr=%7Ba,b,c%7D&integer=43"
```
wolfgangwalther, line 359: I did not try to pass in JSON here, yet, but I think that would take a slightly different route then the query params. So that might yield a different result.

test/Feature/StructureSpec.hs
```diff
@@ -452,6 +453,10 @@ spec = do
                     "format": "enum_menagerie_type",
                     "type": "string"
                   },
+                  "arr": {
+                    "format": "text[]",
+                    "type": "string"
```
wolfgangwalther, line 458: So I had to add those lines to make the tests pass, because I added the array to the `varied_arguments` function. The result does not seem correct, though - I think this looks like #1405. Now I don't think it makes sense to keep this in here like this, because this is not what we want it to do.

If we just changed that here accordingly - did I accidently create a test for https://github.com/PostgREST/postgrest/issues/1405#issuecomment-551325528 ?

wolfgangwalther:
Now to my suggestion to implement support for `GET /rpc/array_argument?arg=a&arg=b&arg=c` to be passed as an array. On top of https://github.com/PostgREST/postgrest/issues/1470#issuecomment-655029587 I found some more (better) reasons as to why to do this:

- this has actually been asked for in the past already: https://github.com/PostgREST/postgrest/issues/328#issuecomment-161077207

- other libraries do it as well: e.g. [NodeJS ](https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options), [PHP](https://www.php.net/manual/en/faq.html.php#faq.html.arrays) (with different syntax), and probably many others

- browsers send forms with multiple fields with the same name like this. Think multiple `<input type="checkbox" name="x">` or `<select multiple name="y">...`. HTML5 allows it: https://stackoverflow.com/a/37579281

Ok, to be honest - I did the implementation first, because I was curious how the query string was parsed etc. - and only then looked for a reason ;)

So how did I implement this and what did I have in mind?

- query strings and `application/x-www-form-urlencoded` bodies are now parsed, so that all params are put into a list

- all lists with only one element are unpacked again. For non-repeated params, this is the same before. Repeated params are now passed as array to the RPCs.

- to use this properly with a function taking an array, this function would have to be overloaded for the "single item" case, like this:

```SQL
CREATE FUNCTION repeated(arg INT) RETURNS INT[]
LANGUAGE SQL AS $$
  SELECT repeated(ARRAY[arg])
$$;

CREATE FUNCTION repeated(arg INT[]) RETURNS INT[]
LANGUAGE SQL AS $$
  SELECT arg
$$;
```
This could now be called with both `GET /rpc/repeated?arg=1&arg=2` and `GET /rpc/repeated?arg=1`. At least that's the theory, I couldn't get this to work, yet. Right now the tests for the overloaded function fail, because the wrong function is chosen. Not sure whether that's related to #1469 (see next paragraph) or something else we can control. Maybe postgres just chooses the wrong overload consistently...

For `VARIADIC` functions this will lead to another problem: We can't define an overloaded function with `VARIADIC` taking a single element instead of an array (that would be pointless...) - and we can't define the overloaded function without `VARIADIC` as well, because we can't reliably determine whether to actually use the `VARIADIC` keyword when passing the arguments or not. This is somewhat related to #1469 - we are not resolving the "overloading" in haskell code, so don't know beforehand which of the functions is going to be used.

But what we might be able to do is to call the `VARIADIC` function something like this in SQL:

```SQL
SELECT * FROM variadic_func(VARIADIC arg := CASE jsonb_typeof(args.arg) WHEN 'array' THEN args.arg ELSE ARRAY[args.arg] END)
```

If that works, then there would suddenly be a valid use-case for `VARIADIC` functions in the first place: To support that "repeated query params" syntax properly. Not implemented yet, so the test for it still fails.

wolfgangwalther:
> This could now be called with both GET /rpc/repeated?arg=1&arg=2 and GET /rpc/repeated?arg=1. At least that's the theory, I couldn't get this to work, yet. Right now the tests for the overloaded function fail, because the wrong function is chosen. Not sure whether that's related to #1469 (see next paragraph) or something else we can control. Maybe postgres just chooses the wrong overload consistently...

Maybe this can be solved by putting those lists that are generated from the repeated query params into `{...,...}` format instead of JSON array format. Would need to do that for PG < 10 anyways, at least some tests fail for 9.4 right now because of that.

test/Feature/RpcSpec.hs
steve-chavez: > In this state this test is basically useless, because you could pass anything and wouldn't notice whether it's cast correctly - that's why I changed the whole test to return the arguments in a json object as well.

@wolfgangwalther You're right. Looks like with the old test, the parameters are ignored if not used in the body. I tried this:

```

CREATE FUNCTION varied_arguments(
  double double precision,
  "varchar" character varying,
  "boolean" boolean,
  date date,
  money money,
  enum enum_menagerie_type,
  "integer" integer default 42,
  json json default '{}',
  jsonb jsonb default '{}'
) RETURNS text
AS $_$
  SELECT $1::text;
$_$ LANGUAGE sql;

http POST "l:3000/rpc/varied_arguments_plpgsql" <<JSON
{
              "double": "x.1",
              "varchar": "hello",
              "boolean": true,
              "date": "20190101",
              "money": 0,
              "enum": "foo",
              "integer": 43,
              "json": {"some key": "some value"},
              "jsonb": {"another key": [1, 2, "3"]}
            }
JSON

HTTP/1.1 400 Bad Request
{
    "code": "22P02",
    "details": null,
    "hint": null,
    "message": "invalid input syntax for type double precision: \"x.1\""
}
```
If the param is used then the syntax error happens.

Check this out, with PLPGSQL(instead of SQL) the function will fail despite the body:

```
CREATE FUNCTION varied_arguments_plpgsql(
  double double precision,
  "varchar" character varying,
  "boolean" boolean,
  date date,
  money money,
  enum enum_menagerie_type,
  "integer" integer default 42,
  json json default '{}',
  jsonb jsonb default '{}'
) RETURNS text
AS $_$ begin
  return 'Hi'::text;
end; $_$ LANGUAGE plpgsql;

http POST "l:3000/rpc/varied_arguments_plpgsql" <<JSON
{
              "double": "x.1",
              "varchar": "hello",
              "boolean": true,
              "date": "20190101",
              "money": 0,
              "enum": "foo",
              "integer": 43,
              "json": {"some key": "some value"},
              "jsonb": {"another key": [1, 2, "3"]}
            }
JSON

HTTP/1.1 400 Bad Request

{
    "code": "22P02",
    "details": null,
    "hint": null,
    "message": "invalid input syntax for type double precision: \"x.1\""
}
```

I was unable to reproduce with plain SQL though:

```
CREATE FUNCTION hey(
  par float8
) RETURNS text
AS $_$
  SELECT 'Hi'::text;
$_$ LANGUAGE sql;

select hey('x.1');
ERROR:  22P02: invalid input syntax for type double precision: "x.1"
LINE 1: select hey('x.1');
```

So the test change is good. My guess is that it has to be the way we wrap the function in our queries.

test/Feature/RpcSpec.hs
steve-chavez: The transformation from query params to sql params happens in Haskell code. I think it may be possible to fix this, but don't think is worth it. As you mentioned is ugly to deal with json(urlencoding) in the params.

I think users that wanted to pass json on GET params have been served well by `Prefer: params=single-object` http://postgrest.org/en/v7.0.0/api.html#calling-functions-with-a-single-json-parameter.

steve-chavez: @wolfgangwalther Now that I think about it maybe the shortcoming is in the ` second JSON.toJSON` in:

```haskell
rpcPrmsToJson = ProcessedJSON (JSON.encode $ mapFromListWithMultiple $ second JSON.toJSON <$> rpcQParams) (S.fromList $ fst <$> rpcQParams)
```

The fix might could be using something like this?(Not sure):

https://github.com/PostgREST/postgrest/blob/8e4687fb5301f9000c31c70f4fe22a87fe6d4d3f/src/PostgREST/Middleware.hs#L90-L95

wolfgangwalther:
Couple more thoughts about the overloading example above (and the corresponding failing test):

Why does this not work?

```
CREATE FUNCTION repeated(arg INT) RETURNS INT[]
LANGUAGE SQL AS $$
  SELECT repeated(ARRAY[arg])
$$;

CREATE FUNCTION repeated(arg INT[]) RETURNS INT[]
LANGUAGE SQL AS $$
  SELECT arg
$$;
```

I think we currently can't use overloaded functions with arguments with the same name. This is because we are resolving which overloaded function to use in Haskell and Postgres does not have to make any choice itself. This is because of:

https://github.com/PostgREST/postgrest/blob/8e4687fb5301f9000c31c70f4fe22a87fe6d4d3f/src/PostgREST/QueryBuilder.hs#L140

where we cast all the arguments to postgres types (and thereby choosing which function to call). To do this, we are currently resolving the overloaded functions only by argument name, so we will only ever use one of the casts.

To improve this, we would need to either:

- Move this logic to postgres somehow. I'm not sure this is even possible.
- Add the actual value to the "which-overload-to-use" resolving process. That's not easy as well and I'm not sure if it's possible for query strings at all, because we don't have json types there, but just strings.

Will think about this a bit more.

test/Feature/RpcSpec.hs
wolfgangwalther: Ok, that makes a lot of sense. I think the `SQL` functions / statements, are actually parsed at definition time - so postgres knows it won't need the arguments and will not even evaluate them (lazy evaluation). `plpgsql` functions on the other hand are parsed at execution time, so the arguments have to be cast before the functions is executed and the error occurs.

Reproducing this in plain sql queries did not work, I guess, because we were both using constants as arguments to the function - and I guess constants are evaluated and cast even before all of that, when parsing the `select hey('x.1')` query.

steve-chavez:
>  This is because we are resolving which overloaded function to use in Haskell and Postgres does not have to make any choice itself. This is because of:

@wolfgangwalther Yeah, that was needed to fix https://github.com/PostgREST/postgrest/pull/1005. I think it was worth it.

> Now to my suggestion to implement support for GET /rpc/array_argument?arg=a&arg=b&arg=c to be passed as an array

This PR got a bit complicated, how about if we leave the multiple params as array for another PR?

You already did solve https://github.com/PostgREST/postgrest/issues/1470. We could merge the variadic part and improve it later. 

variadic parameters are not common on API functions but being compliant with pg syntax is definitely a win for us.

wolfgangwalther:
Yes, you're right it's gotten a bit more. I have thought about it and I think we can make it work now - at least in parts.

I would definitely keep the overloading out of this (and remove the tests in that direction for now), because that's a lot more complicated - and is an issue we need to solve more generally as well.

I might be able to get the "multiple params as array" working for variadic functions by looking at the function arguments while parsing the query string already. In this case there would be no overloading involved. If it doesn't work, then I will keep the "multiple params as array" out completely for now.

Would that work for you?

I think it would be good if we could include that in the next release. What's your current plan: When do you want to release? So I know by when I have to get this done. Had a bit of other stuff on my plate the last couple of days.

steve-chavez:
> I will keep the "multiple params as array" out completely for now.
> Would that work for you?

Yes, that would work! It would also help me in reviewing the multiple param array idea better.  

> What's your current plan: When do you want to release? So I know by when I have to get this done. Had a bit of other stuff on my plate the last couple of days.

I was thinking to release when we merge this one and #1546. No pressure from my part :)

My current plan is to release what we have - so users can try the new reloading features and see if there's any bugs - add docs, and do PostGIS features for the next release.

wolfgangwalther:
I thought about this in-depth again - and I don't think it would be a good idea to merge the variadic part I did so far as it is now. This is to avoid breaking changes later on.

My reasoning is as follows:

- I strongly believe that it would be good to support some form of "multiple params as array" for the reasons somewhere upthread.

- Whenever we support this, we have to deal with the zero and one params cases as well. It's easy to turn 2+ params of the same name into an array and pass this to a function - but the hard part will be to decide whether to turn zero repetitions into an empty array and one repetition into a 1-item array or not.

- Since we are potentially dealing with multi-dimensional arrays (arrays of arrays), this zero/one-as-array-decision will not be easily possible by checking the function arguments type. Also this would be a major challenge with overloaded functions, because our current approach for choosing the function to call (based on argument names only) would not be able to handle this at all. Even though we probably need to improve on the overloading stuff in general, I'm not sure whether we will be able to do everything we would need to support this.

- Variadic functions seems like a perfect candidate to solve this: We still need to handle the overloading case to be able to decide whether we are dealing with a variadic function or not - but this can use the current implementation (of course with the same drawbacks as currently the case). We can use the variadic marker for the decision to turn non-repeated params into an array or not: Once the argument name is variadic, we will just create an array, no matter how often the param is repeated.

- This would have the implication, however, that we can't call variadic functions with array syntax anymore `?variadic_argument={a,b}` - for the same reason as above (multi-dimensional arrays and the decision whether this is already the variadic array or not). I don't think this is a negative in itself - because if you want to pass an array in one param in the query string, you can easily use a regular array argument already. So "calling variadic with array syntax" doesn't actually add any benefit.

- Now if we implement the "calling variadic with array syntax" now, it would be a breaking change to go back to the "multiple params as array" approach.

To avoid having to break this again, if we decide to support repeated params, I suggest to just implement the variadic in the form of repeated params from the start. This also makes a lot more sense conceptually - because that's exactly what "variadic" is all about.

I do already have a very good idea of how to implement this. I still need to do a smaller refactor, to have the function definition (including arguments + overloading decision) available where the query params are parsed, but that will not be a problem. I will not be able to do it this week, though. so @steve-chavez you might want to do the next release without variadic.

steve-chavez:
@wolfgangwalther Sure, let's think better about variadic for the next release.

Also, check my latest review on https://github.com/PostgREST/postgrest-docs/pull/339. It would be great to merge those additions.

wolfgangwalther:
closed in favor of #1603

Issue #1554, opened 2020-07-09
Inexplicable failed token verification with Keycloak RSA


RNCTX:
I'm following your docs and the previously referenced #1191 in trying to get a Keycloak client to generate RSA tokens for use with Postgrest and not having any luck getting auth to work.

### Environment

* PostgreSQL version: 12
* PostgREST version: 7.0.1
* Operating system: Ubuntu 20.04

### Description of issue
```
db-uri = "postgres://apiuser:cccccccccccccccc@/api"
db-schema = "public"
db-anon-role = "web_anon"
jwt-secret = "{ \"kty\":\"RSA\",\"e\":\"AQAB\",\"n\":\"generated on https://8gwifi.org/jwkconvertfunctions.jsp using public key below\" }"
role-claim-key = ".resource_access.apiclient.roles[0]"
jwt-aud = "apiclient"
```

public RSA key from keycloak is:

```
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAqkRL2PimLMDxvdvY3KqwyyBGPs0kxIhDInXhGHBsKyT815V6VvrPSpd9KjUFsE+hk3qgjLzjgTIqwOc0FmC4qooZRwXDIBXS9BVolFtS7GM7sG9sZdlGbj0M9CTEKBW4KGve6k2jEE0KGhU0ZvQa9VTovG/Rh9PAiK1FndKiBTxawCGLQLnlL6ydYxK2znz8/owFFsBsRlA5otU3+en3KZvsRXDlGpmN40CHKaFhkYEh6eMktR5R7aBTVo2Wwp1ucMN/UlDzC/DqH7Sj7UKAaIVHrY1vX5ltvzPx89KbGrZsGAitgTMI0/E9jEADaEoIHXKAKOc7JHper/ntMpS9lwIDAQAB
```

The above configured Postgrest install is connecting to a schema with three roles besides the owner...

```
CREATE ROLE web_anon NOLOGIN;
GRANT web_anon TO apiuser;

CREATE ROLE web_user NOLOGIN;
GRANT USAGE ON SCHEMA public TO web_user;
GRANT web_user TO apiuser;

CREATE ROLE web_admin NOLOGIN;
GRANT USAGE, CREATE ON SCHEMA public TO web_admin;
GRANT web_admin TO apiuser;

[...]

PR #1558, opened 2020-07-14
Additionally allow "bearer" instead of only "Bearer" in Authorization…

wolfgangwalther:
Related to #1554 

While technically using `Bearer` seems to be correct (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml), I think it should be no problem to allow `bearer` as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1525, Allow http status override through response.status guc - @steve-chavez
  - #1512, Allow schema cache reloading with NOTIFY - @steve-chavez
  - #1119, Allow config file reloading with SIGUSR2 - @steve-chavez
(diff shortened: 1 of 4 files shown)


steve-chavez:
@wolfgangwalther Agree. We can be lenient with the spec and help users. It's a simple change so I don't see why not.

Could you add a test? I think you'd have to change:

https://github.com/PostgREST/postgrest/blob/1f6a824dfb00b417da210a529b0ada3fe964c4d6/test/SpecHelper.hs#L197-L199

And add the test in `test/Feature/AuthSpec.hs`.

Also, don't forget to add an "Added" entry to the CHANGELOG :)

wolfgangwalther:
Added the test and added the "Added" to the ADDlog. ;)

(Already pushed it, but GitHub seems to have trouble, so its not showing in the PR, yet)

Issue #1429, opened 2019-12-18
Read jwt-secret from the database


carragom:
### Environment

* PostgreSQL version: 12.1
* PostgREST version: 6.0.2
* Operating system: MacOS Mojave

### Description of issue

Hi, not sure if this is a question or a feature request but here it goes. I'm using [SQL User Management](http://postgrest.org/en/latest/auth.html#sql-user-management) as described in the docs and everything is working as expected.

Whenever I deploy my database schema using migrations a random JWT secret gets created with a statement like this:

```sql
select set_config('app.jwt.secret', md5(random()::text), true);
```

Since the app is in development the schema gets re-deployed quite often and a new random secret gets installed on the database. Every time this happens I have to edit my `postgrest.conf` and set the `jwt-secret` parameter to the result of the following query:

```sql
select current_setting('app.jwt.secret');
```

If this is not done, the authentication does not work. This is expected of course since the signer (a database function) uses a different secret that PostgREST which is the one that verifies the JWT.

I was wondering if there was a way to tell PostgREST to read the `jwt-secret` from the database instead of having it hard coded in the config.  Or maybe there is some other clever (hopefully simple) way to keep those two values in sync.

Thanks for your time and this great tool.

steve-chavez:
@carragom Hey there.

From your example I'm not sure how your `app.jwt.secret` GUC is being persisted since GUCs have a lifetime scoped to transaction or session(it'll be empty if postgrest tries to read it). That's unless you set the GUC to the database like: `alter database postgrest_test set app.jwt.secret = 'mysecret';` then the GUC will be attached to the db lifetime.

For the latter case, since postgrest supports reading from a file in the [jwt-secret](http://postgrest.org/en/v6.0/configuration.html#jwt-secret) you could do something like:

```bash
## Having the postgrest.conf like
jwt-secret = @./mysecret

## Then on deployment
psql postgrest_test -A -t -X -c "select current_setting('app.jwt.secret');" > mysecret
``` 
Then you'd have to restart postgrest since there's no unix signal yet for reloading the jwt-secret(https://github.com/PostgREST/postgrest/issues/1119).

carragom:
@steve-chavez that makes a lot of sense. Will try it that way.

steve-chavez:
That being said, we could add a way to get the jwt-secret directly from the db. There are deployments that have the jwt-secret in a table, like in https://github.com/PostgREST/postgrest-docs/issues/28.

[...]

PR #1559, opened 2020-07-15
Schema cache reload with zero downtime

steve-chavez:
The current SIGUSR1 schema cache reload causes downtime for some milliseconds(depending on schema size). To test this:

```bash
# Start
postgrest conf
# Reload
killall -SIGUSR1 postgrest & curl localhost:3000/projects
{"message":"Database connection lost, retrying the connection."}
```

This is because the current reload(implemented in https://github.com/PostgREST/postgrest/pull/869) was meant for a single schema and the error message prevented getting a response from a stale schema cache.

(The error message is inaccurate when the db connection is actually up and a reload is done).

But now that we support multiple schemas in `db-schema`, is not appropriate to err a response on an up-to-date schema just because a newly added one can be stale.

So with this PR, responses can come from a stale cache. Embedding and rpc can be innacurate for a short time, but pgrst will be available on this lapse.

### DIFF
--- a/test/io-tests.sh
+++ b/test/io-tests.sh
@@ -251,11 +251,11 @@ checkDbSchemaReload(){
     sleep 0.1 \
     || sleep 1 # fallback: subsecond sleep is not standard and may fail
   done
-  secret="reallyreallyreallyreallyverysafe"
   # add v1 schema to db-schema
   replaceConfigValue "db-schema" "test, v1" ./configs/sigusr2-settings.config
   # reload
   kill -s SIGUSR2 $pgrPID
+  kill -s SIGUSR1 $pgrPID
   httpStatus="$(v1SchemaParentsStatus)"
   if test "$httpStatus" -eq 200
   then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1525, Allow http status override through response.status guc - @steve-chavez
  - #1512, Allow schema cache reloading with NOTIFY - @steve-chavez
  - #1119, Allow config file reloading with SIGUSR2 - @steve-chavez
+ - #1559, No downtime when reloading the schema cache with SIGUSR1 - @steve-chavez
 
 ### Fixed
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -45,153 +45,6 @@ import System.Posix.Signals
 import UnixSocket
 #endif
 
--- Time constants
-_32s :: Int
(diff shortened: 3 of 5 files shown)


steve-chavez:
Should be ready!

I've improved the failed connection error messages. Before we would only get:
```bash
{"message":"Database connection lost, retrying the connection."}
```
But now we can get:
```bash
{"details":"no connection to the server\n","message":"Database client error. Retrying the connection."}

# or
{"details":"could not connect to server: Connection refused\n\tIs the server running on host \"localhost\" (::1) and accepting\n\tTCP/IP connections on port 5432?\ncould not connect to server: Connection refused\n\tIs the server running on host \"localhost\" (127.0.0.1) and accepting\n\tTCP/IP connections on port 5432?\n","code":"",
"message":"Database connection error. Retrying the connection."}
```
Some other error messages arise during recovery now. But they include the "Retrying the connection" message to be clear that pgrst is trying to recover.

test/io-tests.sh
```diff
@@ -251,11 +251,11 @@ checkDbSchemaReload(){
     sleep 0.1 \
     || sleep 1 # fallback: subsecond sleep is not standard and may fail
   done
-  secret="reallyreallyreallyreallyverysafe"
   # add v1 schema to db-schema
   replaceConfigValue "db-schema" "test, v1" ./configs/sigusr2-settings.config
   # reload
   kill -s SIGUSR2 $pgrPID
+  kill -s SIGUSR1 $pgrPID
```
steve-chavez, line 258: This change on the test is ensuring the SIGUSR1 doesn't cause downtime.

wolfgangwalther:
Looking at the distinction between "reloadable" and "non-reloadable" parts of the config file, I wonder whether it would be possible to use global postgres variables on a database level for the "realoadable" parts?

So something like this:

```SQL
ALTER DATABASE xyz SET pgrst.db_schema = 'public';
```

Those could be read together with all the other info about the schema on schema reload.

The main benefit of this would be to have only one source of truth. Schema cache and some of the config options depend on each other - so it would make sense to store them all in code. Should be easier for users and maybe easier to handle in haskell code down the road as well?
Together with `NOTIFY` this could be even more "configure once, ...".

steve-chavez:
@wolfgangwalther Hmm.. how would the `db-schema` in the config file look like in that case?

One drawback of the ALTER DB statement is that it doesn't work unless you're the db owner or superuser - relevant on cloud managed services where you don't have those. Still, it could be worth having.

> Together with NOTIFY this could be even more "configure once, ...".

I agree, all in the db is tempting. If we come up with a solution, I believe that would also solve https://github.com/PostgREST/postgrest/issues/1429.

PR #1560, opened 2020-07-15
Return 405 Method not Allowed for GET of volatile RPC instead of 500

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/pull/1552#issuecomment-657863333

I changed the CHANGELOG to include this changed behaviour, but still need to change the docs, I guess ;)


wolfgangwalther:
(Not sure what's happening with GitHub right now. CI checks probably have to be manually triggered, once everything works again)

wolfgangwalther:
Pushed an empty commit to trigger a ci rebuild. Can be squashed away when merging.

steve-chavez:
Looks good!

> I changed the CHANGELOG to include this changed behaviour, but still need to change the docs, I guess ;)
> Adding the changelog item is hard on the first commit - because you have to guess the number of the PR. Let's see whether I guessed right.(from https://github.com/PostgREST/postgrest/pull/1552#issue-445627522)

Yeah, the CHANGELOG duplication is getting a bit tedious. I've opened https://github.com/PostgREST/postgrest-docs/issues/340 to address this for the upcoming release.

nerfpops:
Love it, these small increments of goodness, especially correct error handling, are super useful.

PR #1579, opened 2020-09-04
Upgrade nixpkgs to current master and add an upgrade checklist

monacoremo:
This upgrades the PostgreSQL packages (including libpq linked into the static binary) to the newest [patch releases from August](https://www.postgresql.org/about/news/2060/) (12.4, 11.9 etc.).

Also added a checklist that covers all necessary steps.


nix/UPGRADE.md
```diff
@@ -0,0 +1,94 @@
...
+# From the root of the repository, enter nix-shell
+nix-shell
+
+# Run the utility script to pin the latest revision in master
+nixpkgs-upgrade > nix/nixpkgs-version.nix
+
+# Exit the nix-shell with Ctrl-d
+
+# Alternatively, in one step
+nix-shell --run nixpkgs-upgrade > nix/nixpkgs-version.nix
```
steve-chavez, line 35: This one gives me the following error in and outside of  `nix-shell`:

```
error: while evaluating anonymous function at /home/steve-chavez/Projects/Haskell/postgrest/shell.nix:12:1, called from undefined position:
while evaluating the attribute 'pkgs' at /home/steve-chavez/Projects/Haskell/postgrest/default.nix:80:10:
syntax error, unexpected $end, at /home/steve-chavez/Projects/Haskell/postgrest/nix/nixpkgs-version.nix:1:1
```

The `nix/nixpkgs-version.nix` file contents get empty. I guess that's what the `unexpected $end` error means.

The `nixpkgs-upgrade > nix/nixpkgs-version.nix` version inside `nix-shell` works good though. 

Is something missing on the command?

If the fix is not easy, maybe we can omit the one liner.

monacoremo: Hm, weird - just tested it and it works for me. Maybe it's the nix-shell version? I'm on 2.3.7

But I guess it doesn't hurt to take out that line for now :-)

steve-chavez: Ah, yes. I'm on `2.2.2`. It's like bash redirection doesn't work on that version. Maybe a bug that got fixed.

PR #1581, opened 2020-09-11
Use nix-build-uncached to accelerate CI

monacoremo:
Currently, `nix-build` downloads all items to be built form the cache if they are already built, which wastes 2-3 minutes before tests can start in CI. By using `nix-build-uncached`, we can avoid those unnecessary downloads.


monacoremo:
That... was not actually faster. Will investigate :-)

monacoremo:
Closing for now, not top priority

PR #1590, opened 2020-09-22
Allow wai-extra 3.1

robx:
See https://github.com/commercialhaskell/stackage/issues/5655.

If this passes CI without changes to source code, would be nice to add to add to hackage as a revision.


steve-chavez:
Revision added :) https://hackage.haskell.org/package/postgrest-7.0.1/revisions/

PR #1593, opened 2020-09-25
Removing single column limit from join table M2M mapping detection.

goteguru:
Restricting M2M mapping detection to single foreign key column seems to be unreasonable. I can't see how can it be a problem (by scanning the code), therefore lifted the unnecessary restriction. All tests run fine. Everything is working fine (at least in my app). 

Added some minor fixes. 
- It is impossible for groupWith to return empty foldable, therefore filtering empty is unreasonable and confusing. 
- Text concatenation is ambiguous. myschema_my.table and myschema.my_table should be different, but concatenating them with "_" they evaluate the same which is undesirable and leads to a nasty bug if both schemas are accessible. (Very rare...but still).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1559, No downtime when reloading the schema cache with SIGUSR1 - @steve-chavez
 
 ### Fixed
-
+ 
+ - #1592, Removed single column restriction to allow composite foreign keys in join tables - @goteguru
  - #1530, Fix how the PostgREST version is shown in the help text when the `.git` directory is not available - @monacoremo
  - #1094, Fix expired JWTs starting an empty transaction on the db - @steve-chavez
  - #1475, Fix location header for POST request with select= without PK (#1162) - @wolfgangwalther
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -363,7 +363,7 @@ addM2MRels rels = rels ++ addMirrorRel (mapMaybe junction2Rel junctions)
       Relation{relTable=jt, relColumns=jc1, relConstraint=const1, relFTable=t,  relFColumns=c},
       Relation{             relColumns=jc2, relConstraint=const2, relFTable=ft, relFColumns=fc}
       ]
-      | jc1 /= jc2 && length jc1 == 1 && length jc2 == 1 = Just $ Relation t c Nothing ft fc M2M (Just $ Junction jt const1 jc1 const2 jc2)
+      | jc1 /= jc2 = Just $ Relation t c Nothing ft fc M2M (Just $ Junction jt const1 jc1 const2 jc2)
       | otherwise = Nothing
     junction2Rel _ = Nothing
     addMirrorRel = concatMap (\rel@(Relation t c _ ft fc _ (Just (Junction jt const1 jc1 const2 jc2))) ->
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -351,7 +351,7 @@ addO2MRels = concatMap (\rel@(Relation t c cn ft fc _ _) -> [rel, Relation ft fc
 addM2MRels :: [Relation] -> [Relation]
 addM2MRels rels = rels ++ addMirrorRel (mapMaybe junction2Rel junctions)
   where
-    junctions = join $ map (combinations 2) $ filter (not . null) $ groupWith groupFn $ filter ( (==M2O). relType) rels
+    junctions = join $ map (combinations 2) $ groupWith groupFn $ filter ( (==M2O). relType) rels
     groupFn :: Relation -> Text
     groupFn Relation{relTable=Table{tableSchema=s, tableName=t}} = s <> "_" <> t
     -- Reference : https://wiki.haskell.org/99_questions/Solutions/26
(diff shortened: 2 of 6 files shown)


wolfgangwalther:
> All tests run fine.

That's good. But it would be better if you were to add a test as well. That test should be breaking without your change and passing with it.

goteguru:
Fair enough. I'll try. (I'm not familiar with the testing system the
project is using, but it's never late to learn...)

steve-chavez:
Hey @goteguru. I can reproduce the issue. The fix is looking good but as Wolfgang mentions, we'd need a test to ensure it works.

Some guidance for adding the test:

- For fixtures, add the issue tables on [test/fixtures/schema.sql](https://github.com/PostgREST/postgrest/blob/17f04a886e50a38a161d97ff7bedcd104d02adf1/test/fixtures/schema.sql). Also on [test/fixtures/privileges.sql](https://github.com/PostgREST/postgrest/blob/17f04a886e50a38a161d97ff7bedcd104d02adf1/test/fixtures/privileges.sql).
- For requests, you can add them here: https://github.com/PostgREST/postgrest/blob/17f04a886e50a38a161d97ff7bedcd104d02adf1/test/Feature/QuerySpec.hs#L309-L322

goteguru:
@steve-chavez @wolfgangwalther : 3 tests added. 3 failures without the patch, all pass with the patch.

However I've realized this is a breaking change. If somebody (probably unintentionally) exploited the wrong M2M detection algorithm before, having new (M2M) joins may lead ambiguous relation paths which *must* be resolved (by hints) now (needs downstream code changes).  Maybe some warning about the issue somewhere?

goteguru:
@steve-chavez @wolfgangwalther I've also added a test case for recursive (self) M2M joins (eg. graphs), but strangely it's not working. Any ideas why is it? The detection algo is able to catch them... I've tested it separately.  May be there is a filter somewhere else in the code fidling with self joins?

steve-chavez:
Nice work on the tests @goteguru.

> However I've realized this is a breaking change. If somebody (probably unintentionally) exploited the wrong M2M detection algorithm before, having new (M2M) joins may lead ambiguous relation paths which must be resolved (by hints)

Do you have an example of this happening? I've also noted a limitation on our disambiguation capabilities for M2M here:

https://github.com/PostgREST/postgrest/blob/17f04a886e50a38a161d97ff7bedcd104d02adf1/test/Feature/EmbedDisambiguationSpec.hs#L101-L105

A test failure could be added on the same spec.

> I've also added a test case for recursive (self) M2M joins (eg. graphs), but strangely it's not working.

Hmm.. did you add the commit? I'm not seeing the test on the PR.

Also, oddly enough, I don't see the CircleCI tests runs. Maybe Circle will fix itself on a next commit.

goteguru:
@steve-chavez  : For me, CirceCI run just fine (2 green marks). Strange. 

> Do you have an example of this happening? 

Consider the following: 

```
------------           FK2                ----------
|          |    >|--------------------|-  | TableA |
| TableJ   |                              ----------
|          |                                | FK1
|          |  >|--------------------o- ----------
------------          FK3              |Table B |
                                       ----------  
```

We have a M2O from TableJ to both TableA and TableB and we have a O2O (or M2O/O2M) between TableA and TableB.   FK1, FK2 and FK3 are correctly detected, but TableJ (M2M) not. Queries targeting /TableA?select=TableB(*) running fine because there is only one path between A-B. But if TableJ (M2M) correctly detected, this is not the case anymore. (downstream must fix the queries).  We are working with (moderately) complex databases, and this happened many places. Not a big deal, but still, you have to touch the code.

> Hmm.. did you add the commit? I'm not seeing the test on the PR.

I didn't send it. I was not sure if you like to see failing tests in your repo :-D  You can check it in my fork of course:
[https://github.com/goteguru/postgrest/compare/master...goteguru:fix-M2M-selfjoin](https://github.com/goteguru/postgrest/compare/master...goteguru:fix-M2M-selfjoin)

> I've also noted a limitation on our disambiguation capabilities for M2M here

Ok, I'll check it out. However I think the message is not entirely correct, since having N FKs on the Join table is not a problem (we can happily handle ternary joins), but having *two* FKs from the Join table to the *same* table is. I think it's not very common.  Anyway this *shouldn't* effect on M2M recursive joins, because there are (usually) 2 FKs only.  Both are pointing to the same table, but this case it is (in theory) *resolvable*! The above test shouldn't fail. Do we have a special case handling multiple FKs between tables?

wolfgangwalther:
> However I've realized this is a breaking change.

It seems to me like this is a thing any time we improve the foreign key detection / embedding algorithms. Whenever we have more embedding options available (doesn't matter which specifically!) there might be users that suddenly get an ambiguous embedding error, that they didn't get before.

Therefore it makes sense to at least lump changes to those algorithms together in the same release to avoid to many "breaking changes" (or "bugs fixed, where people relied on buggy behaviour" ;) ).

I have 1 or 2 improvements for foreign-key detection through views coming up, that I stumbled upon, but haven't been able to finish them up, yet. Just fyi, that there might be something coming up, @steve-chavez.

goteguru:
I think it's ok to introduce "breaking fixes" into the dev branch, especially if you intend to add more of them @wolfgangwalther . Users of branch master will know what they are doing, semantic versioning protects the rest.  But probably it's a good idea to mention the fact somewhere, in the changelog maybe, idk.

wolfgangwalther:
> I have 1 or 2 improvements for foreign-key detection through views coming up, that I stumbled upon, but haven't been able to finish them up, yet. Just fyi, that there might be something coming up, @steve-chavez.

Details in #1607 and #1608.

wolfgangwalther:
Now, that we have the nightly builds, I think it would make sense to merge this as well @steve-chavez?

steve-chavez:
Sorry for the late reply here @goteguru. The change is good, I worried a bit about the breaking change for the newly detected M2M's, but we'll likely have some of those for this release anyway.

> May be there is a filter somewhere else in the code fidling with self joins?

Yes, this is something we also need to improve, see https://github.com/PostgREST/postgrest/issues/1643.

I also had an idea for reducing the M2M's embedding possibilities here: https://github.com/PostgREST/postgrest/pull/1625#issuecomment-729002402

I'll merge now :rocket:

Issue #1258, opened 2019-03-22
Some unicode characters require %22 quotes in request URL


mrussell16:
### Environment

* PostgreSQL version: docker (postgres:11)
* PostgREST version: docker (postgrest/postgrest:v5.2.0)
* Operating system: Ubuntu 18.04

### Description of issue

Some unicode characters in column names require quoting with `%22` in the request URL.

I created a table that has a unicode replacement character (https://www.fileformat.info/info/unicode/char/0fffd/index.htm) in one of the column names:

```sql
CREATE TABLE public_api.test (fine_amount_� BIGINT);
```

Percent encoding this column name in requests to PostgREST without `%22` around the column name returns an error response.

A select request (`curl "https://127.0.0.1/test?select=fine_amount_%EF%BF%BD"`) is returning the following error message:

```json
{"hint":"Perhaps you meant to reference the column \"test.fine_amount_�\".","details":null,"code":"42703","message":"column test.fine_amount_ does not exist"}
```

An order request `curl "https://127.0.0.1/test?order=fine_amount_%EF%BF%BD.asc"` returns the following:

```json
{"details":"unexpected '\\65533' expecting letter, digit, \"-\", \"->>\", \"->\", delimiter (.), \",\" or end of input","message":"\"failed to parse order (fine_amount_�.asc)\" (line 1, column 13)"}
```

I have checked that the percent encoding I am using is exactly the same as the character which is in the database using the following python function:

```sql
CREATE LANGUAGE plpythonu;

CREATE OR REPLACE FUNCTION urlescape(original text) RETURNS text LANGUAGE plpythonu AS $$
    import urllib
    return urllib.quote(original);
$$
IMMUTABLE STRICT;

SELECT column_name, urlescape(column_name) FROM information_schema.columns
    WHERE table_schema = 'public_api' AND table_name   = 'test';
```

The same behaviour happens with a ZERO WIDTH NO-BREAK SPACE character(https://www.fileformat.info/info/unicode/char/feff/index.htm).

### Work Around

It is possible to work around this by manually quoting the column name in the request.  The two commands below return the correct output.

```
curl "https://127.0.0.1/test?select=%22fine_amount_%EF%BF%BD%22"
curl "https://127.0.0.1/test?order=%22fine%20amount%20%EF%BF%BD%22.asc"
```

PR #1599, opened 2020-10-02
Fix error messages on connection failure for postgres on localized Windows

### LINKED ISSUE #1585: postgrest: Cannot decode byte '\xfc': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
I cannot get run postgREST 7.0.1 with PG 12.3. I'm using the Docker-based installation, after starting the Docker I can see the following, private settings replaced by *:
```
Starting REST-API Service...
serving "token" at http://127.0.0.1:3001
tcp6       0      0 [::]:3001               [::]:*                  LISTEN
Connecting to database postgres://authenticator:****@*****:5433/***** ...
Listening on port 3000
postgrest: Cannot decode byte '\xfc': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```
When I start the same Docker connecting to Postgres 11.3, the error message will not be displayed and subsequent requests could be delivered successfully:

Postgres will be installed by locale "German_Germany.1252", but this is also the case for the running PG11.3:
```
Starting REST-API Service...
serving "token" at http://127.0.0.1:3001
tcp6       0      0 [::]:3001               [::]:*                  LISTEN
Connecting to database postgres://authenticator:****@*****:5432/***** ...
Listening on port 3000
Connection successful
````

### PR DESCRIPTION
**Pull request by wolfgangwalther**
Resolves #1585. TLDR: When postgres returns localized error messages on Windows, those are returned with a Windows codepage encoding (here `German_Germany.1252`) instead of UTF8 - independent of `client_encoding` (which defaults to UTF8 for Hasql). Converting this error message to JSON fails because strict conversion to UTF8 is used and some of the special characters fall out of this range.

This seems to be only an issue before the connection is established. Once it is, postgres sends UTF8 according to `client_encoding`. Therefore, I think, the "Database Connection Error" should be the only one that needs this change.

Not sure if/where/how to add tests for this. This is what was happening on my local test-setup before:

```
Listening on port 3000
postgrest: Cannot decode byte '\xfc': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```

This is happening after:

```
Listening on port 3000
{"details":"FATAL:  kein pg_hba.conf-Eintrag f�r Host �192.168.0.2�, Benutzer �authenticator�, Datenbank �postgres�, SSL aus\n","code":"","message":"Database connection error. Retrying the connection."}
```

While this is not converted perfectly, I think this is the best we can do here.

Definitely one of the hardest to debug issues in relation to the length of the fix so far. If it wasn't for the additional import, it would be a 1 character difference... ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1530, Fix how the PostgREST version is shown in the help text when the `.git` directory is not available - @monacoremo
  - #1094, Fix expired JWTs starting an empty transaction on the db - @steve-chavez
  - #1475, Fix location header for POST request with select= without PK (#1162) - @wolfgangwalther
+ - #1599, Fix error messages on connection failure for localized postgres on Windows (#1585) - @wolfgangwalther
 
 ### Changed
 
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -28,7 +28,7 @@ import Network.HTTP.Types.Header
 
 import PostgREST.Types
 import Protolude       hiding (toS)
-import Protolude.Conv  (toS)
+import Protolude.Conv  (toS, toSL)
 
 
 class (JSON.ToJSON a) => PgrstError a where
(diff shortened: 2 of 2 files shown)


steve-chavez:
@wolfgangwalther Whoa, nice catch!

Looks like the [toSL](https://hackage.haskell.org/package/protolude-0.3.0/docs/src/Protolude.Conv.html#toSL) variant uses  the [lenientDecode](https://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text-Encoding-Error.html#v:lenientDecode) function.

> Replace an invalid input byte with the Unicode replacement character U+FFFD.

While `toS` uses [strictDecode](https://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text-Encoding-Error.html#v:strictDecode).

> Throw a UnicodeException if decoding fails.

I agree that adding tests is looking complicated now. So the PR looks good to me as it is.

Add an entry to the CHANGELOG and I'll merge :)

steve-chavez:
(We use `toS` all over the place, I wonder if a similar exception could occur in other ways. I remember https://github.com/PostgREST/postgrest/issues/1258, but not sure if it's related)

PR #1603, opened 2020-10-04
Allow calling variadic functions with repeated parameters

### LINKED ISSUE #1470: VARIADIC (Functions with variable numbers of arguments) is not supported with any array type

wolfgangwalther:
Remake of #1552 - new PR for a clean history and easier discussion. Resolves #1470.

This implements variadic arguments for RPCs by either:
- using JSON arrays in the function body or
- using repeated query params

See the test cases for some examples.

To be able to implement this, I had to do the following in 3 commits
- The first commit just improved the "accepts a variety of arguments" test case to include arrays. I had that done in the other PR already and included them here mainly to make sure I'm not breaking regular arrays with everything that follows.
- The second commit is "just" a refactor. The basic idea is to move the column parsing and `findProc` from App to ApiRequest, to be able to later use the result of `findProc` already when parsing the query string - to make sure to only include repeated params for the variadic case and not for regular functions. This also lead to a change in `TargetProc`, which is now passing the `ProcDescription` instead of only a `QualifiedIdentifier` around. To still support the fallback case (i.e. when the schema cache is stale or something) I added a fallback `ProcDescription` to findProc which basically serves only as a container for the identifier which was passed around in that case before. This change also has the side-effect of providing some defaults (volatility, return type), so the places where those were used do not need to make a decision about fallbacks anymore.
- The third commit is the actual implementation of the variadic stuff. Tests and most stuff is almost identical to the changes in the other PR, where they had already been reviewed. The new stuff is all in ApiRequest's `payloadColumns` and `mapFromListWithMultiple`. The latter just appends all params into a list and passes them either as a JSON Array (if variadic) or takes the first element of that list (if non-variadic). For the variadic decision, the result of `findProc` is needed, which in turn is based on `payloadColumns`. To avoid a circular dependency I had to change `payloadColumns` a bit, so that it does not depend on `payload` (which depends on `mapFromListWithMultiple`...) anymore, at least in the cases where we parse the query string.

Note:
The n=1 and n>1 cases are handled through `mapFromListWithMultiple` or an array in the JSON body. For n=0:
- with a JSON body an empty array can be used for the variadic argument
- to call it without arguments through either the query string or by omitting the argument in the json body, one needs to add a `DEFAULT '{}'` to the functions variadic argument definition. This is in line with the default behaviour of postgres.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1512, Allow schema cache reloading with NOTIFY - @steve-chavez
  - #1119, Allow config file reloading with SIGUSR2 - @steve-chavez
  - #1558, Allow 'Bearer' with and without capitalization as authentication schema - @wolfgangwalther
+ - #1470, Allow calling RPC with variadic argument by passing repeated params - @wolfgangwalther
  - #1559, No downtime when reloading the schema cache with SIGUSR1 - @steve-chavez
  - #504, Add `log-level` config option. The admitted levels are: crit, error, warn and info - @steve-chavez
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -2,8 +2,9 @@
 Module      : PostgREST.ApiRequest
 Description : PostgREST functions to translate HTTP request to a domain type called ApiRequest.
 -}
-{-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE LambdaCase     #-}
+{-# LANGUAGE MultiWayIf     #-}
+{-# LANGUAGE NamedFieldPuns #-}
 
 module PostgREST.ApiRequest (
   ApiRequest(..)
@@ -45,6 +46,7 @@ import Web.Cookie                (parseCookiesText)
 import Data.Ranged.Boundaries
 
 import PostgREST.Error      (ApiRequestError (..))
+import PostgREST.Parsers    (pRequestColumns)
 import PostgREST.RangeQuery (NonnegRange, allRange, rangeGeq,
                              rangeLimit, rangeOffset, rangeRequested,
                              restrictRange)
@@ -63,7 +65,7 @@ data Action = ActionCreate       | ActionRead{isHead :: Bool}
             deriving Eq
 -- | The target db object of a user action
 data Target = TargetIdent QualifiedIdentifier
-            | TargetProc{tpQi :: QualifiedIdentifier, tpIsRootSpec :: Bool}
+            | TargetProc{tProc :: ProcDescription, tpIsRootSpec :: Bool}
             | TargetDefaultSpec{tdsSchema :: Schema} -- The default spec offered at root "/"
             | TargetUnknown [Text]
             deriving Eq
@@ -90,7 +92,7 @@ data ApiRequest = ApiRequest {
   , iLogic                :: [(Text, Text)]                   -- ^ &and and &or parameters used for complex boolean logic
   , iSelect               :: Maybe Text                       -- ^ &select parameter used to shape the response
   , iOnConflict           :: Maybe Text                       -- ^ &on_conflict parameter used to upsert on specific unique keys
-  , iColumns              :: Maybe Text                       -- ^ &columns parameter used to shape the payload
+  , iColumns              :: S.Set FieldName                  -- ^ parsed colums from &columns parameter and payload
   , iOrder                :: [(Text, Text)]                   -- ^ &order parameters for each level
   , iCanonicalQS          :: ByteString                       -- ^ Alphabetized (canonical) request query string for response URLs
   , iJWT                  :: Text                             -- ^ JSON Web Token
@@ -103,12 +105,13 @@ data ApiRequest = ApiRequest {
   }
 
 -- | Examines HTTP request and translates it into user intent.
-userApiRequest :: NonEmpty Schema -> Maybe Text -> Request -> RequestBody -> Either ApiRequestError ApiRequest
-userApiRequest confSchemas rootSpec req reqBody
+userApiRequest :: NonEmpty Schema -> Maybe Text -> DbStructure -> Request -> RequestBody -> Either ApiRequestError ApiRequest
+userApiRequest confSchemas rootSpec dbStructure req reqBody
   | isJust profile && fromJust profile `notElem` confSchemas = Left $ UnacceptableSchema $ toList confSchemas
   | isTargetingProc && method `notElem` ["HEAD", "GET", "POST"] = Left ActionInappropriate
   | topLevelRange == emptyRange = Left InvalidRange
   | shouldParsePayload && isLeft payload = either (Left . InvalidBody . toS) witness payload
+  | isLeft parsedColumns = either Left witness parsedColumns
   | otherwise = Right ApiRequest {
       iAction = action
       , iTarget = target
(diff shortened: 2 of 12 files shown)


src/PostgREST/ApiRequest.hs
```diff
@@ -189,12 +202,26 @@ userApiRequest confSchemas rootSpec req reqBody
...
       (CTOther "application/x-www-form-urlencoded", _) ->
-        let json = M.fromList . map (toS *** JSON.String . toS) . parseSimpleQuery $ toS reqBody
+        let json = mapFromListWithMultiple . map (toS *** JSON.String . toS) . parseSimpleQuery $ toS reqBody
             keys = S.fromList $ M.keys json in
         Right $ ProcessedJSON (JSON.encode json) keys
       (ct, _) ->
         Left $ toS $ "Content-Type not acceptable: " <> toMime ct
-  rpcPrmsToJson = ProcessedJSON (JSON.encode $ M.fromList $ second JSON.toJSON <$> rpcQParams) (S.fromList $ fst <$> rpcQParams)
+  rpcPrmsToJson = ProcessedJSON (JSON.encode $ mapFromListWithMultiple $ second JSON.toJSON <$> rpcQParams) (S.fromList $ fst <$> rpcQParams)
+  mapFromListWithMultiple ls = M.mapWithKey listToValue $ M.fromListWith reverseConcat . map valueToList $ ls
```
steve-chavez, line 211: This function is a tad complicated. We should try to simplify it.

I was remembering this code and I found out that I left this unused type:
https://github.com/PostgREST/postgrest/blob/ccad9eb9bc87b09de668eee8d9fb56aefc8862f9/src/PostgREST/Types.hs#L402

Maybe you could try and make that a `data RpcQParam` and add an [instance ToJSON](https://hackage.haskell.org/package/aeson-1.5.4.1/docs/Data-Aeson.html#t:ToJSON) with some of the `mapFromListWithMultiple` logic.

Also, I've found [jsonAccum](https://hackage.haskell.org/package/aeson-1.5.4.1/docs/Data-Aeson-Parser-Internal.html#v:jsonAccum) through https://github.com/bos/aeson/pull/714. Looks like it could help but unfortunately I haven't been able to find an example snippet of how to use it.

wolfgangwalther: `jsonAccum` didn't help, because it parses a string as json - we already have lists of params at that point.

However the other suggestion turned out to be great - simplifies the logic a lot!

wolfgangwalther:
Rebased on current master and refactored:
- ApiRequest.hs: `mapFromListWithMultiple` renamed to `paramsFromList` with a lot less logic inside. returns `RpcParams`. Moved part of the `isVariadic` function to a helper in Types.hs.
- Types.hs: Added `RpcParams` and `RpcParamValue` types, including `toJSON` instance and some helper functions

src/PostgREST/Types.hs
```diff
@@ -398,8 +407,26 @@ type Alias = Text
...
 
--- Rpc query param, only used for GET rpcs
-type RpcQParam = (Text, Text)
+-- RPC query param, used for POST of form-data and GET requests
+data RpcParamValue = Fixed Text | Variadic [Text]
+
+mergeParams :: RpcParamValue -> RpcParamValue -> RpcParamValue
+mergeParams (Variadic a) (Variadic b) = Variadic $ b ++ a
+-- repeated params for non-variadic arguments are not merged
+mergeParams v _                       = v
```
wolfgangwalther, line 416: need to add a test here, that it keeps the first occurence of a repeated param in the non-variadic case - not the last.

src/PostgREST/ApiRequest.hs
```diff
@@ -189,12 +202,18 @@ userApiRequest confSchemas rootSpec req reqBody
         json <- csvToJson <$> CSV.decodeByName reqBody
         note "All lines must have same number of fields" $ payloadAttributes (JSON.encode json) json
       (CTOther "application/x-www-form-urlencoded", _) ->
-        let json = M.fromList . map (toS *** JSON.String . toS) . parseSimpleQuery $ toS reqBody
+        let json = paramsFromList . map (toS *** toS) . parseSimpleQuery $ toS reqBody
```
wolfgangwalther, line 205: add a test here for variadic arguments passed with form-data.

src/PostgREST/Types.hs
```diff
@@ -157,13 +159,17 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
...
-  case M.lookup qi allProcs of
-    Nothing     -> Nothing
-    Just [proc] -> Just proc           -- if it's not an overloaded function then immediately get the ProcDescription
-    Just procs  -> find matches procs  -- Handle overloaded functions case
+findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> ProcDescription
+findProc qi payloadKeys paramsAsSingleObject allProcs = fromMaybe fallback bestMatch
   where
+    -- instead of passing Maybe ProcDescription around, we create a fallback description here when we can't find a matching function
+    -- args is empty, but because "specifiedProcArgs" will fill the missing arguments with default type text, this is not a problem
+    fallback = ProcDescription (qiSchema qi) (qiName qi) (Just "fallback descriptor") mempty (SetOf $ Composite $ QualifiedIdentifier "" "record") Volatile
```
steve-chavez, line 167: Hmm... how about making it like:
```suggestion
    fallback = ProcDescription (qiSchema qi) (qiName qi) Nothing mempty (SetOf $ Composite $ QualifiedIdentifier "" "record") Volatile
```
The  `(SetOf $ Composite $ QualifiedIdentifier "" "record")` doesn't look right though.
To be more correct, shouldn't we default to returning `void`?

wolfgangwalther: > `SetOf $ Single $`

That doesn't make sense. Did you want to make it `SetOf Scalar` or `Single Composite`?

In any case: We need to default to `SetOf Composite` to at least return all results in any kind of format in this case.

steve-chavez: > "unknown" reads quite well in the list as well, but I understand that type to be associated with scalars of unknown type and not.. records.

Agree. record is good. Just changing the description to Nothing would be fine.

> SetOf $ Single $
> That doesn't make sense.

Yeah, sorry, I messed up there. Disregard that suggestion.

steve-chavez:
Only one thing left to do for me: run a load test and see if the PR has an impact on non-variadic RPC.

(Once https://github.com/PostgREST/postgrest/pull/1609 is finished, we'll check this automatically)

wolfgangwalther:
> Only one thing left to do for me: run a load test and see if the PR has an impact on non-variadic RPC.

In the meantime I will have to dig out the piece of documentation I had already started and update that.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #1530, Fix how the PostgREST version is shown in the help text when the `.git` directory is not available - @monacoremo
+ - #1603, Fix calling RPC with variadic argument by passing repeated params (#1470) - @wolfgangwalther
```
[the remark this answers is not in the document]
steve-chavez: My bad, I didn't notice the #1470 was at the end.

> I have been putting the PR with the change in front and related issues (could be more than one) at the end. Is that alright?

Until now, we've been putting the issue first, that way end users can understand the context of the fix/enhancement.
Sometimes when there's no issue and a direct PR is done, the PR number is put on the CHANGELOG.

> I have been putting the PR with the change in front and related issues (could be more than one) at the end. Is that alright?

Ideally it would be a PR per issue. Easier for users to read and easier for me to review ;)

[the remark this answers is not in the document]
wolfgangwalther: Ok, thanks for clearing that up. I changed that in all my open PRs and also created a small PR for those changelog entries that have already been committed with the wrong issue numbers.



> Ideally it would be a PR per issue. Easier for users to read and easier for me to review ;)

I know, the rpc-return stuff got quite big.... ;)
It's hard to slow myself down, when I'm on a roll and when stuff just depends on each other that leads to bigger PRs.

wolfgangwalther:
rebased on master with merge conflict resolved; docs PR is done.

steve-chavez:
I think I can refactor this a bit, but I'll do it on another PR.

Looks good to merge :rocket:

(Docs on https://github.com/PostgREST/postgrest-docs/pull/357)

Issue #540, opened 2016-03-31
customize log verbosity


ilkosta:
Hi,
Postgrest seems fantastic! 
It would be nice to see logged near the requests submitted to postgrest, also the pg response errors.
What do you think?

begriffs:
We could add that, although the errors will be in your db logs as well depending on your logging settings.

begriffs:
Note there is more documentation now about debugging server problems - https://postgrest.readthedocs.io/en/v0.4/admin.html#debugging

begriffs:
Revisiting this, a log level would be nice in order to both reduce logging (since it might be duplicating the logs of a proxy server), or turn them up.

I propose we allow syslog style [log levels](https://en.wikipedia.org/wiki/Syslog#Severity_level).

trourance:
Hi, are there any progress on that ? 
All other tickets end up here. I would like to have some verbose logging without having to modify the db side logs, for example for SSL connection:
`SSL established: TLSv1.2/ECDHE-RSA-AES256-GCM-SHA384/ECDH=prime256v1`

steve-chavez:
When having several clients that access the database, it can be hard to pinpoint the log entries that belong to a particular pgrst instance.

I think we should recommend using [log_line_prefix](https://www.postgresql.org/docs/9.5/runtime-config-logging.html#GUC-LOG-LINE-PREFIX) and setting [application_name](https://www.postgresql.org/docs/12/runtime-config-logging.html#GUC-APPLICATION-NAME) in the connection uri(on `db-uri`). This would be like:
```
postgres://user@localhost/mydb?application_name=pgrst_mydb_myschema"
```
I was also thinking if we could somehow set the `application_name` ourselves. But in any case this should be documented.

steve-chavez:
I have a counterpoint against doing more detailed logging in the syslog style.

Recently, I helped an user that was in need of an audit trail for PostgREST events, this was needed for compliance with a standard. The idea was to have detailed request information in JSON(payload body, headers, success/error codes, etc) and then send them to Kafka for later processing.

Naturally the first idea that comes to mind was to add more detailed logging in PostgREST itself and construct the required JSON.

However, this detailed logging would introduce code paths that can lead to performance degradation. In the past we have had problems such as #690 and #925(#1206 still open btw). Those were fixed basically by not messing with the payload body in Haskell code and instead rely on directly sending it to pg through parametrized queries.

[...]

Issue #1596, opened 2020-09-28
Feature request: Sentry integration


dsferruzza:
Hi!

There are some cases where PostgREST can return a HTTP 500 response, for example when a stored procedure raises an exception (on purpose or not). In general (in apps I develop), I tend to deal with these unhandled exceptions by adding an integration to [Sentry](https://sentry.io/): it is an open source and SaaS product that ingests exception events (augmented by the integration to provide more context), aggregates them and notifies maintainers. (In my opinion, it is almost inconceivable to operate a real-world production system without this kind of tools; it is **so** useful 💜.)

Do you think this could be relevant to integrate Sentry in PostgREST, so that, if a Sentry DSN was provided in PostgREST's configuration, any failure (either from PostgreSQL or PostgREST itself) would be sent to Sentry?

---

Additional notes:

1. There seems to be a _relatively_ up-to-date package in Hackage to play with Sentry: https://hackage.haskell.org/package/raven-haskell
2. I believe this could also be achievable in the reverse proxy (for example, with [OpenResty](https://openresty.org/) + [raven-lua](https://github.com/cloudflare/raven-lua)) but with a poor access to context (for example, it would be harder to report which user was authenticated)

steve-chavez:
Hey @dsferruzza.
                                                                                                                                                                                                                                                                           
I see the convenience of the Sentry integration. We've had a similar request on [gitter chat](https://gitter.im/begriffs/postgrest?at=5f28ee78a2be9049908deec0) and the recommendation was to use a log collector(logstash/fluentd) to send the logs to Sentry from there.
                                                                                                                                                                                                                                                                           
That being said we've just refined our logging(https://github.com/PostgREST/postgrest/pull/1604) and we could consider supporting Sentry in the future.

dsferruzza:
Hi!

Using a custom log collector that sends errors to Sentry could indeed work, and provide value.
This might be a better approach than doing this in the reverse proxy (even if more complex to setup, I guess) because it would not be on the critical path and might have more detailed error messages.

But from my perspective, the main point of having a native integration is to be able to **provide more context** in Sentry errors (for example, which user was authenticated, or maybe what queries did PostgREST already send to the database).

[...]

PR #1604, opened 2020-10-05
Add log level config

steve-chavez:
Add the `log-level` config option. The admitted values are:

- `crit`: No requests are logged. Only startup and db connection recovery messages are logged.
- `error`: Only failed requests(status >=400) are logged.
- `info`: All requests are logged. This is the only level we've been offering until now.

Partly addresses: https://github.com/PostgREST/postgrest/issues/540 (`debug` level for another PR).

### Perf

According to my load tests, the lower levels give more requests per second:

- `crit`: 208 req/s more on average.
- `error`: 170 req/s more on average.

Which makes sense because our default `info` level doesn't buffer the log messages. It writes them to disk, directly.
The `error` level is a bit slower because of the status filtering.

### Changing the default log level

- Should we change the default log level to `error`? It certainly looks better for production.
- When PostgREST is behind nginx, the `access.log` already contains the logged requests. In these cases the `crit` level could be used. Since we're usually behind a proxy, perhaps we could default to `crit` instead?

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -266,11 +266,12 @@ readAppConfig cfgPath = do
     parseLogLevel :: C.Key -> C.Parser C.Config LogLevel
     parseLogLevel k =
       C.optional k C.string >>= \case
-        Nothing -> pure LogInfo
-        Just "" -> pure LogInfo
-        Just "crit" -> pure LogCrit
-        Just "info" -> pure LogInfo
-        Just _ -> fail "Invalid logging level. Check your configuration."
+        Nothing      -> pure LogInfo
+        Just ""      -> pure LogInfo
+        Just "crit"  -> pure LogCrit
+        Just "error" -> pure LogError
+        Just "info"  -> pure LogInfo
+        Just _       -> fail "Invalid logging level. Check your configuration."
 
     reqString :: C.Key -> C.Parser C.Config Text
     reqString k = C.required k C.string
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -195,7 +195,7 @@ readPathShowHelp = customExecParser parserPrefs opts
           |# raw-media-types="image/png, image/jpg"
           |
           |## logging level, the admitted values are: crit, error, warn and info.
-          |# log-level = "info"
+          |# log-level = "error"
           |]
 
 -- | Parse the config file
@@ -266,8 +266,8 @@ readAppConfig cfgPath = do
     parseLogLevel :: C.Key -> C.Parser C.Config LogLevel
     parseLogLevel k =
       C.optional k C.string >>= \case
-        Nothing      -> pure LogInfo
-        Just ""      -> pure LogInfo
+        Nothing      -> pure LogError
+        Just ""      -> pure LogError
         Just "crit"  -> pure LogCrit
         Just "error" -> pure LogError
         Just "warn"  -> pure LogWarn
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -194,7 +194,7 @@ readPathShowHelp = customExecParser parserPrefs opts
           |## content types to produce raw output
           |# raw-media-types="image/png, image/jpg"
           |
-          |## logging level. The admitted values are: info and crit
+          |## logging level, the admitted values are: crit, error, warn and info.
           |# log-level = "info"
           |]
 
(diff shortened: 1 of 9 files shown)


wolfgangwalther:
>     * When PostgREST is behind nginx, the `access.log` already contains the logged requests. In these cases the `crit` level could be used. Since we're usually behind a proxy, perhaps we could default to `crit` instead?

+1 for making `crit` the default.

Personally I have never used the logging output of postgrest so far, except for everything that's there with `crit`. I'd be using `debug` a lot more as well, but I usually find the request logging quite useless, because I do that on the client-side during development (in the browser) with a lot more detail.

Over in #1596 a distinction between errors in the 4xx and in the 5xx range is made. I think that makes a lot of sense. Imho, everything in the 4xx range is more or less expected (or at least handled properly), while those errors in 5xx deserve a lot more attention. Do you think it's possible to make a difference here?

I would probably turn on `crit+5xx` instead of `crit`, if available.

If this distinction makes sense to you as well, it could be like this:
- `crit`: same thing that you do now.
- `error`: crit + 5xx
- `warning`: error + 4xx
- `info`: all requests
- `debug`: + sql

src/PostgREST/Config.hs
```diff
@@ -190,9 +193,11 @@ readPathShowHelp = customExecParser parserPrefs opts
           |
           |## content types to produce raw output
           |# raw-media-types="image/png, image/jpg"
+          |
+          |## logging level. The admitted values are: info and crit
```
wolfgangwalther, line 197: `error` as well?

steve-chavez:
>  I think that makes a lot of sense. Imho, everything in the 4xx range is more or less expected (or at least handled properly), while those errors in 5xx deserve a lot more attention.

> - crit: same thing that you do now.
> - error: crit + 5xx
> - warning: error + 4xx
> - info: all requests
> - debug: + sql

Good call! I'll add the `warning` level.

> (In that case, I would tend to make error the default)

So do you think we should have `error`(crit + 5xx) as the default?

I'm tempted to make `crit` the default because it plays better with nginx(production setup) and gives more req/s.

wolfgangwalther:
> I'm tempted to make `crit` the default because it plays better with nginx(production setup) and gives more req/s.

My reasoning to make `error` (crit+5xx) the default would be:
- I expect 5xx errors to be something that should NOT happen in regular usage (i.e. when all is good) - when 5xx is thrown, something **bad** happens and I need to know that it does
- I expect the performance to only suffer when logging is actually happening. I might be wrong here - it depends a bit on whether your benchmark had errors in it as well, or not. In any case if only the decision to log an error or not is decreasing performance already - that seems odd to me. Can we do better here?
- If enough 5xx errors pile up to seriously harm performance - I have a way bigger problem than my performance to deliver those errors responses.

src/PostgREST/Config.hs
steve-chavez: Yeah, forgot about that. Solved!

steve-chavez:
@wolfgangwalther Totally agree on the reasoning for making `error` the default. I've made the changes.

> I expect the performance to only suffer when logging is actually happening. I might be wrong here - it depends a bit on whether your benchmark had errors in it as well, or not. 

No errors on the benchmark - [one single req is tested and non-200 are checked](https://github.com/steve-chavez/postgrest-benchmark/blob/master/k6/GETSingle.js#L30-L31). 

> In any case if only the decision to log an error or not is decreasing performance already - that seems odd to me. Can we do better here?

I think the next step for this would be adding the `pgrstFormat` logic on the [wai-extra](https://hackage.haskell.org/package/wai-extra-3.0.29.2/docs/Network-Wai-Middleware-RequestLogger.html) package. We could get feedback there on how to make it faster.

For now I'm happy with the results. Let me know if you agree and I'll merge :)

wolfgangwalther:
LGTM!

steve-chavez:
(ci/circlei: stack-test failed because io-tests are flaky:`Received "segmentation fault" signal`. Should be good to merge).

elimisteve:
`log-level = "debug"` would be great; been trying to fix `invalid input syntax for integer: ""` for a week now. It's seemingly caused by a race condition -- potentially in PostgREST itself -- since the same exact request sometimes triggers the error and sometimes doesn't.

If it'd slow things down too much and just show what's in postgres's access logs anyway, I understand not adding `"debug"` as an option.

steve-chavez:
> log-level = "debug" would be great; been trying to fix invalid input syntax for integer: "" for a week now
>  It's seemingly caused by a race condition -- potentially in PostgREST itself -- since the same exact request sometimes triggers the error and sometimes doesn't.

@elimisteve Sorry, this somehow fell off my radar. I've actually seen that before, but thought it was an user error. Please open an issue if you encounter it again so we can track it.

> If it'd slow things down too much and just show what's in postgres's access logs anyway, I understand not adding "debug" as an option.

`debug` is definitely in the plans. This week we'll release v8, then we'll look into this.

PR #1606, opened 2020-10-05
Upgrade nixpkgs and add Postgres 13 support

monacoremo:
- upgrade nixpkgs to current master
- link against libpq from Postgres 13 by default
- integrate tests against Postgres 13
- Fix tests for a changed error message in Postgres 13

### DIFF
--- a/nix/patches/nixpkgs-revert-ghc-bootstrap.patch
+++ b/nix/patches/nixpkgs-revert-ghc-bootstrap.patch
@@ -1,13 +1,14 @@
-From f9e4b7fd7a93ab643db0b1c84003987400bbe120 Mon Sep 17 00:00:00 2001
-From: Remo <remo>
-Date: Sun, 23 Aug 2020 12:01:48 +0200
+From 9ec726b1d1a231dcff5ada065a858e9faaa7463f Mon Sep 17 00:00:00 2001
+From: monacoremo <monacoremo>
+Date: Sun, 4 Oct 2020 16:16:04 +0200
 Subject: [PATCH] Revert "ghc: 8.6.3-binary -> 8.6.5-binary"
 
 This reverts commit 3c7ef6bcd85e3e133ac6f2ab5e238934d56d902e.
+
 ---
- .../{8.6.5-binary.nix => 8.6.3-binary.nix}    | 13 +++++-----
- pkgs/top-level/haskell-packages.nix           | 25 ++++++++-----------
- 2 files changed, 16 insertions(+), 22 deletions(-)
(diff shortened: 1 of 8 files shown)


wolfgangwalther:
There's one more reference to pg12 in `tests/docker-compose.yml` - that could be updated to pg13 as well, I guess. If somebody used it, it'd make sense to use the latest version.

monacoremo:
> There's one more reference to pg12 in `tests/docker-compose.yml` - that could be updated to pg13 as well,

Good point @wolfgangwalther , updated! I used the `13` tag in favor of `latest` or `13.0`, as it should stay resonably reproducible for local testing while not requiring frequent updates.

steve-chavez:
@monacoremo Looks good!

>  weird that the 'io tests' failed this time

Ah, the `Received "segmentation fault" signal` error has happened to me before. Not sure why it does.. I restart manually on those cases :)

nix/patches/nixpkgs-revert-ghc-bootstrap.patch
```diff
@@ -52,29 +53,29 @@ index 41af279e83f..5853a470efc 100644
...
-
+ 
 -    ghc865Binary = callPackage ../development/compilers/ghc/8.6.5-binary.nix { };
 +    ghc863Binary = callPackage ../development/compilers/ghc/8.6.3-binary.nix { };
-
-     ghc865 = callPackage ../development/compilers/ghc/8.6.5.nix {
-       bootPkgs = packages.ghc822Binary;
-@@ -56,31 +57,25 @@ in {
+ 
+     ghc8102Binary = callPackage ../development/compilers/ghc/8.10.2-binary.nix {
```
steve-chavez, line 76: Currently we're using `ghc883`:
https://github.com/PostgREST/postgrest/blob/d395bb60522aa228a534d9181339519f171fe6b2/default.nix#L5-L6

Can I switch that to `ghc8102Binary` to use GHC 8.10 and expect to all work the same?
GHC 8.10 might contain some memory usage improvements and I was thinking to benchmark it :)

steve-chavez: Oh, actually I've just tried that and I get a nix failure:
```
error: attribute 'haskellCompilerName' missing, at /nix/store/yacxa1kgxalkyw7fyk27xm76vp7gjqwd-source/pkgs/development/haskell-modules/make-package-set.nix:136:32
```

No problem though. That was more of a question.

monacoremo: @steve-chavez it probably works for normal builds, but for the fully static build we need `static-haskell-nix` to catch up: https://github.com/nh2/static-haskell-nix/pull/98

Issue #1399, opened 2019-10-31
At a random point in time stops responding on HTTP requests.


fortyanov:
### Environment
* PostgreSQL version: 12
* PostgREST version: 6.0.2-713b214
* Operating System: FreeBSD 12.1-STABLE

### Description of the problem
At a random point in time, it stops responding to http requests. We tried the program assembled manually and the assembly from github, the problem occurs in both cases.

Our .conf file contains options
db-pool = 15
db-pool-timeout = 1000

I cannot provide more detailed information, because the logs do not contain information about the reasons for stopping the response to HTTP requests. When a problem occurs, reloading the database schema works fine. The only way to resume the process is to restart it.

In the near future I will attach TCP dump, as soon as problem repeats.

fortyanov:
Here is TCP dump [https://www.codepile.net/pile/akPp5Vkp](url)

fortyanov:
Here is TCP dump.
[dump.pcap.zip](https://github.com/PostgREST/postgrest/files/3798460/dump.pcap.zip)

steve-chavez:
I think this could be the same problem as #905. That issue is frequently visited, seems the error persists even on current versions.

Using `tshark` on the pcap file reveals nothing for me.

How about if you add this change https://github.com/steve-chavez/postgrest/commit/110314d84533311717b8d6273cc6d0de1c5fcf40 and recompile from master? Perhaps that could lead to a more meaningful error message on the client.

fortyanov:
I rebuilt the project with a patch 110314d84533311717b8d6273cc6d0de1c5fcf40 but when problem repeats, I still have no log messages. My haskell is very bad. Can you say in which parts of project should I turn my attention.

steve-chavez:
Oh, the patch only adds a client error message for when the empty reply happens again. I'd like to see what's the error message in that case.

steve-chavez:
Also, could it be an out of memory issue?

fortyanov:
Free memory is about 2 GB minimum. Daemon not returns any error messages and do not response on any HTTP requests and do not close TCP connection.

fortyanov:
We set parameter
db-pool-timeout = 1
and problem not repeats about 2 days. Hope this information will helps you.

fortyanov:
Problem repeats again but not stable.  Seems like the problem is somehow related to db-pool-timeout parameter. Do you know some more logging instruments to get more information?

steve-chavez:
[...]

Issue #286, opened 2015-09-13
Atomic updates to join tables


timbod7:
I have a schema capturing events and participants:

```
CREATE TABLE demo.participant(
  id BIGINT DEFAULT,
  name VARCHAR(255) NOT NULL
  );

CREATE TABLE demo.event(
  id BIGINT DEFAULT,
  title VARCHAR(255) NOT NULL
  );

CREATE TABLE demo.event_participant(
  event_id BIGINT NOT NULL REFERENCES demo.event(id),
  participant_id BIGINT NOT NULL REFERENCES demo.participant(id),
  PRIMARY KEY(event_id,participant_id)
  );
```

I can expose these tables directly with 3 postgrest views. However, I required that inserting/updating an event and the participants in that event should be atomic.

I think I need a single view that somehow data from event and event_participant, and is insertable/updateable.

Is there any way to do this with postgrest?

timbod7:
I'm curious that there hasn't been a response to this. How are people handling many-many relationships with postgrest?

begriffs:
Hey sorry for the slow response. I've been away on vacation this week and not on the computer much.

There is currently no general way to batch operations in a single transaction. As a workaround you can create a stored procedure to do complicated updates and call it with the [rpc interface](https://github.com/begriffs/postgrest/wiki/Routing#stored-procedures).

It would be good to invent a nice interface for batch operations. We could take advantage of HTTP/2 [multiplexing](https://http2.github.io/faq/#why-is-http2-multiplexed) and share a single transaction per request. Then the client could send several requests in the same connection to execute atomically.

timbod7:
Thanks for the reply. An interface for batch operations would be awesome.

But apart from the "how can I do this atomically" question, I guess I was wondering if there was a better way to model the data relationship given the capabilities of postgrest now? Given my naive model above, If I want to update an event along with the  list of participants associated, I need to:

      - PUT the new event value
      - DELETE the existing values from event_participant
      - POST a csv to bulk insert the new values in event_participant

Perhaps, as you say the RPC approach is the best one.

begriffs:
The RPC is your best bet at the moment, but we really need a batch interface. I can see that being high priority after the upcoming v0.3 release.

dcominottim:
Is this still a priority? When this is implemented, I will be able to use PostgREST in many of my projects.

ruslantalpa:
[...]

PR #1609, opened 2020-10-05
[for discussion] Integrate a load test into the local dev tools and CI

monacoremo:
This is an experiment on integrating a load test into the local dev tools and CI.

To run the loadtest locally, run `postgrest-loadtest` in nix-shell. This will run for 60s by default.

This is based on https://locust.io/, but happy to try other frameworks out if you have suggestions!

The locustfile could be extended with more complex reads, writes etc. that can then be selected via tags to simulate different workloads.

The Postgres database, PostgREST server, Nginx reverse proxy and the load all run on the same machine with this setup, which makes it easy to quickly run locally and in CI, but it impairs comparability.

Todos:
- [ ] output a human readable report (perhaps with Github checks integration?)
- [ ] validate that the results are sufficiently stable across runs
- [ ] compare performance (e.g. to current master) and potentially fail if a certain threshold is passed?
- [x] make `postgrest-loadtest` an option in nix-shell in order to maintain fast nix-shell startup without cachix
- [ ] add more cases to be tested (get with embed, post etc.)
- ...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,7 +176,7 @@ jobs:
             fi
       - run:
           name: Install testing scripts
-          command: nix-env -f default.nix -iA tests tests.ioTests tests.memoryTests
+          command: nix-env -f default.nix -iA tests tests.ioTests tests.memoryTests loadtest
       - run:
           name: Skip tests on build failure
           command: circleci-agent step halt
@@ -213,6 +213,9 @@ jobs:
           name: Run memory tests
           command: postgrest-test-memory
           when: always
+      - run:
+          name: Run load test
+          command: postgrest-loadtest
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,7 @@ jobs:
             postgrest-release-dockerhub
             postgrest-release-dockerhubdescription
 
-  # Build everything in default.nix, push to the Cachix binary cache and run tests
+  # Build everything in default.nix, run tests and push to the Cachix binary cache
   nix-build-test:
     machine: true
     steps:
@@ -162,18 +162,6 @@ jobs:
           command: |
             nix-env -iA cachix -f https://cachix.org/api/v1/install
             cachix use postgrest
-      - run:
-          name: Build all derivations from default.nix and push results to Cachix
-          command: |
-            # Only push to the cache when CircleCI makes the CACHIX_SIGNING_KEY
-            # available (e.g. not for pull requests).
-            if [ -n "${CACHIX_SIGNING_KEY:-""}" ]; then
-              echo "Building and caching all derivations..."
-              nix-build | cachix push postgrest
-            else
-              echo "Building all derivations (caching skipped for outside pull requests)..."
-              nix-build
-            fi
       - run:
           name: Install testing scripts
           command: nix-env -f default.nix -iA tests tests.ioTests tests.memoryTests loadtest
@@ -216,6 +204,21 @@ jobs:
       - run:
           name: Run load test
           command: postgrest-loadtest
+          when: always
+      - run:
+          name: Build all derivations from default.nix
+          command: nix-build
+      - run:
+          name: Push built artifacts to Cachix
+          command: |
+            # Only push to the cache when CircleCI makes the CACHIX_SIGNING_KEY
+            # available (e.g. not for pull requests).
+            if [ -n "${CACHIX_SIGNING_KEY:-""}" ]; then
+              echo "Pushing to Cachix..."
+              make cachix-push-all
+            else
+              echo "Skipping push to Cachix."
+            fi
 
 workflows:
   version: 2
--- a/nix/loadtest.nix
+++ b/nix/loadtest.nix
@@ -41,7 +41,7 @@ let
 
         ${locust}/bin/locust -f ${../test/loadtest/locustfile.py} \
             -H http://localhost:3080 --users 100 --spawn-rate 20 \
-            --headless -t 60s
+            --headless -t 2m
       '';
 
   postgrestConf =
@@ -65,7 +65,7 @@ let
         events {}
 
         http {
-            access_log access.log;
+            access_log access.log combined buffer=32k flush=5m;
             client_body_temp_path . 1 2;
             proxy_temp_path proxy_temp;
             fastcgi_temp_path fastcgi_temp;
--- a/nix/loadtest.nix
+++ b/nix/loadtest.nix
@@ -1,0 +1,100 @@
+{ postgrest
+, envsubst
+, locust
+, openresty
+, postgresql
+, withTmpDb
+, writeShellScript
+, writeShellScriptBin
+, writeText
+}:
+
+let
+  runner =
+    writeShellScript "postgrest-loadtest-runner"
+      ''
+        set -euo pipefail
+
+        tmpdir=$(mktemp -d)
+        trap "rm -r $tmpdir" exit
+
+        export POSTGREST_TEST_DIR="$tmpdir"
+
+        ${postgrest}/bin/postgrest ${postgrestConf} &
+        postgrestPID=$!
+
+        export POSTGREST_API_URL="http://localhost:3000"
+
+        mkdir -p "$tmpdir"/{logs,conf}
+        touch "$tmpdir"/logs/{error.log,access.log}
+        ${envsubst}/bin/envsubst -i ${nginxConf} \
+          -o "$tmpdir/conf/nginx.conf"
+
+        ${openresty}/bin/openresty -p "$tmpdir" &
+        nginxPID=$!
+
+        cleanup() {
+          kill $postgrestPID $nginxPID || true
+          rm -r "$tmpdir"
+        }
+
+        trap cleanup exit
+
+        ${locust}/bin/locust -f ${../test/loadtest/locustfile.py} \
+            -H http://localhost:3080 --users 100 --spawn-rate 20
+            # --headless -t 60s
+      '';
+
+  postgrestConf =
+    writeText "postgrest.conf"
+      ''
+        db-uri = "$(POSTGREST_TEST_CONNECTION)"
+        db-schema = "test"
+        db-anon-role = "postgrest_test_anonymous"
+        db-pool = 1
+        server-port = 3000
+        jwt-secret = "reallyreallyreallyreallyverysafe"
+      '';
+
+  nginxConf =
+    writeText "nginx.conf"
+      ''
+        daemon off;
+        error_log stderr info;
+        pid nginx.pid;
+
+        events {}
+
+        http {
+            access_log access.log;
+            client_body_temp_path . 1 2;
+            proxy_temp_path proxy_temp;
+            fastcgi_temp_path fastcgi_temp;
+            uwsgi_temp_path uwsgi_temp;
+            scgi_temp_path scgi_temp;
+
+            include ${openresty}/nginx/conf/mime.types;
+            default_type application/ocet-stream;
+
+            gzip off;
+            sendfile on;
+
+            keepalive_timeout 65s;
+
+            server {
+                listen 3080;
+
+                location / {
+                    more_clear_input_headers Accept-Encoding;
+                    proxy_pass $POSTGREST_API_URL;
+                }
+            }
+        }
+      '';
+in
+writeShellScriptBin "postgrest-loadtest"
+  ''
+    set -euo pipefail
+
+    ${withTmpDb postgresql} ${runner}
+  ''
--- a/nix/loadtest.nix
+++ b/nix/loadtest.nix
@@ -8,7 +8,6 @@
 , writeShellScriptBin
 , writeText
 }:
-
 let
   runner =
     writeShellScript "postgrest-loadtest-runner"
@@ -41,8 +40,8 @@ let
         trap cleanup exit
 
         ${locust}/bin/locust -f ${../test/loadtest/locustfile.py} \
-            -H http://localhost:3080 --users 100 --spawn-rate 20
-            # --headless -t 60s
+            -H http://localhost:3080 --users 100 --spawn-rate 20 \
+            --headless -t 60s
       '';
 
   postgrestConf =
--- a/nix/loadtest.nix
+++ b/nix/loadtest.nix
@@ -41,7 +41,7 @@ let
 
         ${locust}/bin/locust -f ${../test/loadtest/locustfile.py} \
             -H http://localhost:3080 --users 100 --spawn-rate 20 \
-            --headless -t 2m
+            --reset-stats --only-summary --headless -t 1m
       '';
 
   postgrestConf =
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ result*
 dist-newstyle
 postgrest.hp
 postgrest.prof
+__pycache__
--- a/default.nix
+++ b/default.nix
@@ -35,6 +35,7 @@ let
       allOverlays.gitignore
       allOverlays.ghr
       (allOverlays.haskell-packages { inherit compiler; })
+      allOverlays.locust
     ];
 
   # Evaluated expression of the Nixpkgs repository.
--- a/default.nix
+++ b/default.nix
@@ -108,9 +108,12 @@ rec {
   nixpkgsUpgrade =
     pkgs.callPackage nix/nixpkgs-upgrade.nix { };
 
+  withTmpDb =
+    pkgs.callPackage nix/withtmpdb.nix { };
+
   # Scripts for running tests.
   tests =
-    pkgs.callPackage nix/tests.nix { inherit postgrest postgrestStatic postgrestProfiled postgresqlVersions; };
+    pkgs.callPackage nix/tests.nix { inherit postgrest postgrestStatic postgrestProfiled postgresqlVersions withTmpDb; };
 
   # Development tools, including linting and styling scripts.
   devtools =
@@ -122,4 +125,7 @@ rec {
       inherit docker;
       postgrest = postgrestStatic;
     };
+
+  loadtest =
+    pkgs.callPackage nix/loadtest.nix { inherit withTmpDb; postgrest = postgrestStatic; };
 }
--- a/nix/overlays/default.nix
+++ b/nix/overlays/default.nix
@@ -2,5 +2,6 @@
   gitignore = import ./gitignore.nix;
   ghr = import ./ghr;
   haskell-packages = import ./haskell-packages.nix;
+  locust = import ./locust;
   postgresql-default = import ./postgresql-default.nix;
 }
--- a/nix/overlays/locust/default.nix
+++ b/nix/overlays/locust/default.nix
@@ -1,0 +1,8 @@
+self: super:
+
+let
+  flask-basicauth = super.python38Packages.callPackage ./flask-basicauth.nix { };
+in
+{
+  locust = super.python38Packages.callPackage ./locust.nix { inherit flask-basicauth; };
+}
--- a/nix/overlays/locust/default.nix
+++ b/nix/overlays/locust/default.nix
@@ -1,5 +1,4 @@
 self: super:
-
 let
   flask-basicauth = super.python38Packages.callPackage ./flask-basicauth.nix { };
 in
--- a/nix/overlays/locust/flask-basicauth.nix
+++ b/nix/overlays/locust/flask-basicauth.nix
@@ -1,0 +1,30 @@
+{ buildPythonPackage
+, fetchFromGitHub
+, flask
+, unittest2
+}:
+
+buildPythonPackage rec {
+  pname = "flask-basicauth";
+  version = "0.2.0";
+
+  src = fetchFromGitHub {
+    owner = "jpvanhal";
+    repo = "flask-basicauth";
+    rev = "v${version}";
+    sha256 = "1xkhw5nkhwwzxcz54mr9wzzpx4657scmdim1b2p7km886cxg9ac5";
+  };
+
+  propagatedBuildInputs =
+    [
+      flask
+    ];
+  checkInputs = [ unittest2 ];
+
+  doCheck = false;
+
+  meta = {
+    homepage = "https://github.com/jpvanhal/flask-basicauth";
+    description = "HTTP basic access authentication for Flask.";
+  };
+}
--- a/nix/overlays/locust/locust.nix
+++ b/nix/overlays/locust/locust.nix
@@ -11,6 +11,7 @@
 , pyquery
 , pyzmq
 , requests
+, requests-unixsocket
 , unittest2
 }:
 
@@ -37,6 +38,7 @@ buildPythonPackage rec {
       pyquery
       pyzmq
       requests
+      requests-unixsocket
     ];
   checkInputs = [ mock unittest2 ];
   # remove file which attempts to do GET request
--- a/nix/overlays/locust/locust.nix
+++ b/nix/overlays/locust/locust.nix
@@ -1,0 +1,54 @@
+{ buildPythonPackage
+, ConfigArgParse
+, fetchFromGitHub
+, flask
+, flask-basicauth
+, gevent
+, geventhttpclient
+, mock
+, msgpack
+, psutil
+, pyquery
+, pyzmq
+, requests
+, unittest2
+}:
+
+buildPythonPackage rec {
+  pname = "locustio";
+  version = "1.2.3";
+
+  src = fetchFromGitHub {
+    owner = "locustio";
+    repo = "locust";
+    rev = version;
+    sha256 = "11y28hi28m98dl26v9wxx4nl5756vpfgvg4crpsm93hrx7w8043q";
+  };
+
+  propagatedBuildInputs =
+    [
+      ConfigArgParse
+      flask
+      flask-basicauth
+      gevent
+      geventhttpclient
+      msgpack
+      psutil
+      pyquery
+      pyzmq
+      requests
+    ];
+  checkInputs = [ mock unittest2 ];
+  # remove file which attempts to do GET request
+  preCheck = ''
+    rm locust/test/test_stats.py
+  '';
+
+  # Running tests will take some more work
+  doCheck = false;
+
+  meta = {
+    homepage = "https://locust.io/";
+    description = "Scalable user load testing tool written in Python";
+  };
+}
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -12,23 +12,10 @@
 , postgrest
 , postgrestStatic
 , postgrestProfiled
-, runtimeShell
-, writeShellScript
 , writeShellScriptBin
+, withTmpDb
 }:
 let
-  # Wrap the `test/with_tmp_db` script with the required dependencies from Nix.
-  withTmpDb =
-    postgresql:
-    writeShellScript "postgrest-test-${postgresql.name}"
-      ''
-        set -euo pipefail
-
-        export PATH=${postgresql}/bin:${git}/bin:${runtimeShell}/bin:"$PATH"
-
-        exec ${../test/with_tmp_db} "$@"
-      '';
-
   # Script to run the Haskell test suite against a specific version of
   # PostgreSQL.
   testSpec =
--- a/nix/withtmpdb.nix
+++ b/nix/withtmpdb.nix
@@ -1,0 +1,15 @@
+# Wrap the `test/with_tmp_db` script with the required dependencies from Nix.
+{ git
+, runtimeShell
+, writeShellScript
+}:
+
+postgresql:
+writeShellScript "postgrest-test-${postgresql.name}"
+  ''
+    set -euo pipefail
+
+    export PATH=${postgresql}/bin:${git}/bin:${runtimeShell}/bin:"$PATH"
+
+    exec ${../test/with_tmp_db} "$@"
+  ''
--- a/shell.nix
+++ b/shell.nix
@@ -27,6 +27,7 @@ lib.overrideDerivation postgrest.env (
         pkgs.cabal-install
         pkgs.cabal2nix
         pkgs.postgresql
+        pkgs.locust
         postgrest.nixpkgsUpgrade
         postgrest.devtools
         postgrest.tests
--- a/shell.nix
+++ b/shell.nix
@@ -31,6 +31,7 @@ lib.overrideDerivation postgrest.env (
         postgrest.nixpkgsUpgrade
         postgrest.devtools
         postgrest.tests
+        postgrest.loadtest
       ]
       ++ lib.optional ioTests postgrest.tests.ioTests
       ++ lib.optional memoryTests postgrest.tests.memoryTests
--- a/shell.nix
+++ b/shell.nix
@@ -9,7 +9,7 @@
 #
 # We highly recommend that use the PostgREST binary cache by installing cachix
 # (https://app.cachix.org/) and running `cachix use postgrest`.
-{ ioTests ? false, memoryTests ? false, docker ? false, release ? false }:
+{ ioTests ? false, memoryTests ? false, docker ? false, release ? false, loadtest ? false }:
 let
   postgrest =
     import ./default.nix;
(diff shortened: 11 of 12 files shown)


wolfgangwalther:
I love the idea. That will really be helpful. I wonder how comparable the loadtest will be in CI from run to run - but that's something that we should be able to find out pretty quickly.

If it were possible to somehow run both the current master build and the PR build at the same time - that could increase comparability a lot and also allow some automatic comparisons, that lead to a test failure above/below a certain threshold?

I have no idea about the whole CI pipeline that is set up (and especially the nix stuff), but there seems to be a bit of caching involved everywhere. Is caching possible across runs? And can you set up your own caches? If yes to both, something like the following could maybe work:
- run all tests and build
- move new build to cache-pr
- only on the master branch: move build to cache-master as well
- in the loadtest part: load builds from cache-pr and cache-master and run both to compare

So basically the current build of master would always be kept in a different cache to be able to run against the current PR build. Even if circle CI performance would be different between runs, running both builds at the same time should eliminate that.

Maybe I'm completely off here and this doesn't work at all... just wanted to throw that out!

wolfgangwalther:
Oh and I'm not sure you know about the stuff steve is doing, see https://github.com/PostgREST/postgrest/pull/1600#issuecomment-703196548 ?

Maybe those can play together?

monacoremo:
@wolfgangwalther Agree, it would be great to get a direct performance comparison! E.g. in nixpkgs, the build bot Ofborg generates a report that is accessible directly in Github: https://github.com/NixOS/nixpkgs/pull/99650/checks?check_run_id=1211150408

No idea how they are doing this though, will need to look into it.

By the way, results from this first run in CI (a bit hidden in the output right now, need to see how to make it more visible):
```
 Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 GET /                                                            621     0(0.00%)  |     423      76    1768     390  |   10.39    0.00
 GET /articles                                                   6304     0(0.00%)  |     295       1    1872     260  |  105.48    0.00
 GET /orders?select=name                                         6373     0(0.00%)  |     295       1    1747     260  |  106.63    0.00
--------------------------------------------------------------------------------------------------------------------------------------------
 Aggregated                                                     13298     0(0.00%)  |     301       1    1872     270  |  222.50    0.00

Response time percentiles (approximated)
 Type     Name                                                              50%    66%    75%    80%    90%    95%    98%    99%  99.9% 99.99%   100% # reqs
--------|------------------------------------------------------------|---------|------|------|------|------|------|------|------|------|------|------|------|
 GET      /                                                                 390    490    560    600    740    860   1100   1100   1800   1800   1800    621
 GET      /articles                                                         260    370    440    480    600    710    860    960   1300   1900   1900   6304
 GET      /orders?select=name                                               260    370    440    490    610    700    830    920   1300   1700   1700   6373
--------|------------------------------------------------------------|---------|------|------|------|------|------|------|------|------|------|------|------|
 None     Aggregated                                                        270    380    450    490    610    710    860    960   1300   1900   1900  13298
```

--> 301 reqs/s and reponse time of 1.9s at 99.99% 

I'll fix the styling fail and we'll see what comes out in a second run :-)

@wolfgangwalther happy to guide you through the Nix and CI setup if you're interested, would also be good to get your pointers where we can add to the documentation of that setup!

monacoremo:
Second run:

```
 Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 GET /                                                            847     0(0.00%)  |     290      69    1175     260  |   14.17    0.00
 GET /articles                                                   8244     0(0.00%)  |     201       1    1405     180  |  137.89    0.00
 GET /orders?select=name                                         8108     0(0.00%)  |     200       1    1222     180  |  135.61    0.00
--------------------------------------------------------------------------------------------------------------------------------------------
 Aggregated                                                     17199     0(0.00%)  |     205       1    1405     190  |  287.67    0.00

Response time percentiles (approximated)
 Type     Name                                                              50%    66%    75%    80%    90%    95%    98%    99%  99.9% 99.99%   100% # reqs
--------|------------------------------------------------------------|---------|------|------|------|------|------|------|------|------|------|------|------|
 GET      /                                                                 260    340    390    420    510    580    700    760   1200   1200   1200    847
 GET      /articles                                                         180    260    300    340    420    490    580    660    890   1400   1400   8244
 GET      /orders?select=name                                               180    250    300    330    420    490    580    690    910   1200   1200   8108
--------|------------------------------------------------------------|---------|------|------|------|------|------|------|------|------|------|------|------|
 None     Aggregated                                                        190    260    310    340    420    500    590    680    920   1300   1400  17199
```

I rebased on the Postgres 13 PR that got merged earlier, so let me re-run once more!

steve-chavez:
Also love the idea!

> Oh and I'm not sure you know about the stuff steve is doing, see #1600 (comment) ?
> Maybe those can play together?

The benchmark I'm doing is mostly about how many req/s we can support on different EC2 instances. It shouldn't conflict with this PR. I use [k6](https://k6.io/) for my load tests, but it's cool to use locust on our CI :).

> compare performance (e.g. to current master) and potentially fail if a certain threshold is passed?

This should definitely be the goal! Failing on the threshold would let us know of drops in performance when changes happen.

Including a couple more scenarios on this PR would be ideal:
- GET with embedding: [example request](https://github.com/steve-chavez/postgrest-benchmark/blob/master/k6/GETSingleEmbed.js#L30).
- POST: [example request](https://github.com/steve-chavez/postgrest-benchmark/blob/master/k6/POSTSingle.js#L30-L47).

The above are done on a different db(Chinook). Using any of the tables from our fixtures for those would be cool.

nix/loadtest.nix
```diff
@@ -0,0 +1,99 @@
...
+    writeText "nginx.conf"
+      ''
+        daemon off;
+        error_log stderr info;
+        pid nginx.pid;
+
+        events {}
+
+        http {
+            access_log access.log;
```
steve-chavez, line 68: I think you may be able to get more req/s if you enable buffering for the logs:

```suggestion
            access_log access.log  main buffer=32k flush=5m;
```

.circleci/config.yml
```diff
@@ -213,6 +201,24 @@ jobs:
           name: Run memory tests
           command: postgrest-test-memory
           when: always
+      - run:
+          name: Run load test
+          command: postgrest-loadtest
+          when: always
+      - run:
+          name: Build all derivations from default.nix
+          command: nix-build
```
wolfgangwalther, line 210: The `nix-build` is now run after all the tests in the pipeline. How are the tests now run against the new build? It looks like they are run against the cached version from cachix, which would always be the last build from master? What am I missing?

monacoremo: Good point @wolfgangwalther , I'll need to add a comment to document this. This setup works correctly and makes the tests run a bit earlier in the job, failing faster if needed.

`nix-env -iA` in the earlier step also uses `nix-build` under the hood. At this point, only the things that are really needed to run the tests are built or pulled from the cache if available. `nix-env -iA` will never install anything stale from the cache. 

This `nix-build` call at  this step then just builds whatever is not yet built or pulled from the cache.

.circleci/config.yml
```diff
@@ -162,21 +162,9 @@ jobs:
...
-            # available (e.g. not for pull requests).
-            if [ -n "${CACHIX_SIGNING_KEY:-""}" ]; then
-              echo "Building and caching all derivations..."
-              nix-build | cachix push postgrest
-            else
-              echo "Building all derivations (caching skipped for outside pull requests)..."
-              nix-build
-            fi
       - run:
           name: Install testing scripts
```
wolfgangwalther, line 166: maybe this should be renamed to mention that it also builds here

.circleci/config.yml
wolfgangwalther: Ah cool. Just had a comment on that previous step, that it should maybe be renamend then. Deleted it again, after I realized there are actually 2 steps that have `nix-env -iA` in them. So a comment + rename of one of the steps would sure help, I guess.

I understand that we could basically run the tests and `nix-build` in parallel after we are done with `Install testing scripts`? Have you tried doing that with circleCI's "workspaces" feature?
It looks like you could split this into 3 jobs, where the test and build jobs run in parallel after the first one. I guess you could just copy the whole `/nix` folder between those jobs via persist_to_workspace / attach_workspace.

See here for reference:
https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/
https://circleci.com/blog/deep-diving-into-circleci-workspaces/

One benefit of that would be that those jobs would both show up separately in the check list on github. I think this could then be extended to all the tests against different pg versions - to have them show up one by one on the check list. Only if that's wanted, of course.

monacoremo: Using workspaces is a cool idea! We'll need to test how this performs, as the `/nix/` folder can easily be a few GB large, having a few GHC versions for the static builds etc. Not sure if 'saving' and 'loading' the workspace could be a bottleneck (it would be with the CircleCI caching feature, but maybe workspaces work differently).

wolfgangwalther: > as the `/nix/` folder can easily be a few GB large, having a few GHC versions for the static builds etc

Shouldn't those be fetched only at the `nix-build` stage?

But yeah performance needs to be tested. Very likely that it won't actually perform better.

wolfgangwalther:
Another question:

Looking at the report at https://app.circleci.com/pipelines/github/PostgREST/postgrest/426/workflows/6fb87532-cdd6-4936-a3ca-bfb181d088b3/jobs/6779.

The step "Run the test specs against PostgreSQL 9.5" took > 2 minutes. Looking at the output it shows that the lib and test-suite are rebuilt completely. However I understood, that this should have happenend in the step before ("Install testing scripts") already, because of `nix-env -iA`?

Is the same thing happening here twice? Or is `nix-env -iA` actually not building anything, but the test command builds it on-demand?

monacoremo:
> Is the same thing happening here twice? Or is `nix-env -iA` actually not building anything, but the test command builds it on-demand?

Will also need to clarify/document that somewhere :-) Nix builds the whole environment required for running the tests (including all library dependencies, ghc, cabal-install and postgres), but the tests themselves are run using `cabal-install` directly.

The `postgrest-test-*` scripts that we define through Nix are intended for incremental testing/development, they don't use Nix to run build/run the tests themselves. Building/running tests with Nix would mean that we can't use the caching and incremental builds that `cabal-install` provides, which would be quite painful. So we use Nix to reproducibly define and build absolutely everything for running the tests, but then run the tests using [`cabal v2-test`](https://github.com/PostgREST/postgrest/blob/eed0d3a66c48f7fec8d3b31708aa8be6f7bbc0be/nix/tests.nix#L49). We ripped out the caching for this last step as it was causing issues (cabal bug?), so this is why you are seeing those 2 min build time currently.

wolfgangwalther:
Thanks, I think I understand it a whole lot better now.

wolfgangwalther:
@monacoremo
I just started `nix-shell --arg loadtest true` on your branch and then switched to current master to run the loadtests there. I assume this should work because the nix shell stuff is loaded into memory by this time and doesn't care that the `.nix` files on disk are changed, right?

I was surprised to see, that when running `postgrest-loadtest` the executable was never re-built. I think this should be triggered automatically, just like with the spec-tests. Maybe it's just because all those builds for those commits have already been cached? It doesn't look like that from the output, though. Just noticed that `postgrest-test-io` behaves exactly the same - starts the tests immediately. However when I do `nix-build` after switching branches, this does load new builds from the cache.

I'm trying `nix-build && postgrest-loadtest` right now and I am getting results that look quite similar across runs... unfortunately across commits where they should not:
- 350 req/s on current master
- 350 req/s on 698fac8ff2fa206df217d147abe37253b16df81b (before the prepared statements commit)
(both values are +- 2req/s across different runs on the same branch, so quite stable)

I expect the performance to be significantly less without prepared statements, as that's what Steves tests have shown.

I guess the results from `nix-build` are not used for the loadtest. I tried with a manual `cabal v2-build` as well, but that didn't work either (I didn't really expect it to).

Is there anything that I need to do differently?

Ah. I did something differently now: I stashed your changes and applied them to the master branch to be able to launch `nix-shell --arg loadtest` on the current branch. This seems to work, now I get 450 req/s for current master and only 350 req/s for the other commit mentioned above.

So this means that the executable that is used with the load/memory/io-tests is actually build when launching nix-shell. That seems to limit the use of `postgrest-loadtets` and the other commands inside `nix-shell` a alot, because I can't test my current changes.

Any way we can change this?

wolfgangwalther:
It looked like the numbers were quite consistent at the beginning. Ranged around +- 5min. I expected to get better numbers, when running longer, so I took a couple of 30 min runs. Variance did not change. That's not what I would have expected, that questions reliability (repeatability) of those tests.

I wonder whether we can implement proper A/B testing where we always compare two postgrest executables with each other. Ideally in an alternating setting, so something like 5 sec A, 5 sec B, ... repeat for 10 times or something. That should be a lot more independent of current machine load and give a delta of req/s as a result. This delta should be much better to infer conclusions from.

> The Postgres database, PostgREST server, Nginx reverse proxy and the load all run on the same machine with this setup, which makes it easy to quickly run locally and in CI, but it impairs comparability.

Another question: Why is there nginx running at all? Can the clients not make the requests directly to PostgREST? I understand that nginx (or another reverse proxy) is probably part of most setups - but we can't do much about it's performance, at least in this repo. Including it feels like taking resources (and focus) off of testing PostgREST itself.

monacoremo:
Haven't looked at this for a while, it's next up on my list.

There are a few interesting options [here](https://github.com/denji/awesome-http-benchmark) for maximum performance tests, like wrk. I got 3000rps on my local machine with it. Thinking about including a balanced load test with locust, and a max unbalanced load test with another tool.

monacoremo:
nginx helped make postgrest more stable with very highly concurrent loads, if I remember correctly - will test again

We will definitely be able to run the incremental build, now that we've figured out cabal v2-exec.

A/B testing could be possible based on our nightly releases - let's think about this.

wolfgangwalther:
I did a few tests as well last week, but haven't been able to write them up. some interesting numbers for sure. not exactly sure what to make of those.

wolfgangwalther:
> Another question: Why is there nginx running at all? Can the clients not make the requests directly to PostgREST? I understand that nginx (or another reverse proxy) is probably part of most setups - but we can't do much about it's performance, at least in this repo. Including it feels like taking resources (and focus) off of testing PostgREST itself.

I tested it without the nginx stuff and it's working fine. Performance is actually the same, so it does not seem to hurt. The log output is a lot better (cleaner), because all the error logging from nginx is gone. For the simple case, I don't think we need nginx.

---

I also experimented a bit with a different number of users. Comparing last week's master branch and the commit right before prepared statements were introduced with 5s each step:

<details>
  <summary>Data set</summary>
  
users | master | pre prep.
--: | --: | --:
0 | 0,0 | 0,0
5 | 34,0 | 35,1
10 | 68,0 | 66,4
15 | 103,0 | 98,5
20 | 133,3 | 131,3
25 | 166,3 | 162,0
30 | 196,1 | 193,4
35 | 232,6 | 213,1
40 | 248,8 | 239,0
45 | 257,4 | 295,1
50 | 300,6 | 346,7
55 | 336,4 | 341,9
60 | 289,8 | 313,6
65 | 317,6 | 358,9
70 | 393,9 | 367,9
75 | 362,3 | 338,0
80 | 410,4 | 296,4
85 | 406,3 | 298,0
90 | 399,0 | 309,9
95 | 421,4 | 352,7
100 | 404,8 | 339,9
105 | 423,5 | 310,4
110 | 395,9 | 333,6
115 | 334,3 | 288,1
120 | 380,5 | 286,0
125 | 353,5 | 322,7
130 | 343,8 | 320,0
135 | 259,9 | 294,1
140 | 316,9 | 269,0
145 | 348,9 | 289,7
150 | 354,6 | 293,4
155 | 315,4 | 256,3
160 | 352,6 | 241,7
165 | 335,9 | 273,3
170 | 330,1 | 280,9
175 | 333,4 | 267,3
180 | 346,0 | 247,0
185 | 254,3 | 287,6
190 | 289,9 | 259,1
195 | 275,0 | 290,9
</details>

<details>
  <summary>Scatter plot (red=master, blue=pre prep)</summary>

locust
</details>

Up to 30 users the numbers are exactly the same. Although a bit more variation, the same trend is still continued up to 55 users. From 60 users on the variation starts to increase massively. It makes no sense that perf first decreases then increases again with more users - so this is random variation for sure. I observed this variation with a lot longer run times as well. The trend is clear, though: The performance improvement with prepared statements is clearly visible across runs. However:
- with lower numbers of users, the variation is low, but both branches show no difference
- with higher numbers of users, the variation is high, but both branches perform differently

From the plot, one can tell that this makes the current locust test basically useless, because even with a huge performance difference you could end up randomly in a spot as shown with `users=135` in the plot, which indicates that prepared statements perform worse than before.

I'm pretty sure that the reason is that with lower numbers of users locust is not able to saturate PostgREST, so we have idle times and all requests can be served. So the number of requests is basically the number of requests that locust can send, not that PostgREST can handle. Clearly locust requires a lot of resources and when we increase the number of users, locust and PostgREST have to share those, since we're running everything on the same machine. The most likely cause is just random scheduling differences between runs, I assume. We can't really fight those. Plus, when we want to run those tests in CI, we only have 2 cores available (I looked that up, I think it was 2 - very few for sure). I tried a few other things as well, to reduce variation - with no success.

There's one other thought I had: Why are we testing with a multi-user setup? We are forcing postgrest to use only 1 database connection with `db-pool=1` already, and that makes sense, so the whole locust machinery of creating multiple users just adds a **huge** overload. If we were to test in a multi-threaded situation - we would also test our underlying http framework big-time, as that provides the multi-threadedness, right? But we don't need to do that.

For our dev and CI tools use-case, where we want to know how our "request to query" code performs - it would be best to just use a single threaded test runner, that performs 1 request, waits for the response and then sends the next request. This would need to be implemented most efficiently - so no python :D. We can then just hammer PostgREST with one query for 5s, then the next query for 5s etc. to get an idea of different code paths.

----

We could do this with a simple `while` loop around curl, I guess, but @monacoremo you posted a nice list:

> There are a few interesting options [here](https://github.com/denji/awesome-http-benchmark) for maximum performance tests, like wrk. I got 3000rps on my local machine with it. Thinking about including a balanced load test with locust, and a max unbalanced load test with another tool.

I think what I asked for above, is what you mean by "unbalanced" load test, right?

I will now narrow down the list of benchmark tools in the linked repo to those that might make sense for us. I am only looking for:
- CLI tool obviously + simple to use
- possibility to run single-threaded (increases the chances, that the other core - if we have two - can be used by PostgREST / Postgres)
- no "slow" comment + written in compiled languages
- supports POST, PUT and PATCH with body
- supports unix socket (see #1399)
- supports HTTP/2 support (see #286)

Only very few of the tools do support http/2, unix sockets and all request methods. Some came close, but only one of the tools looks really promising to solve all of that: https://github.com/tsenart/vegeta. Some highlights of that:

>   -unix-socket string
    	Connect over a unix socket. This overrides the host address in target URLs

This seems to be - if at all supported - trouble everywhere else. Overriding the host address in target URLs is exactly what is needed, though - no fiddling with URIs to support sockets...

>   -cpus int
    	Number of CPUs to use (defaults to the number of CPUs you have)

That will be a good one to limit us to 1 CPU for CI.

> http format
> The http format almost resembles the plain-text HTTP message format defined in RFC 2616 but it doesn't support in-line HTTP bodies, only references to files that are loaded and used as request bodies (as exemplified below).
> Although targets in this format can be produced by other programs, it was originally meant to be used by people writing targets by hand for simple use cases.

If I understand correctly, we can just write our test-cases as plain text files in HTTP message format. This is awesome, because it has exactly no boilerplate code at all.

`vegeta` is also available in nixpkgs already: https://github.com/NixOS/nixpkgs/blob/278d8641c8da8cea2fa506e7a887d887ba38b14c/pkgs/tools/networking/vegeta/default.nix

I will try to give `vegeta` a test run.

steve-chavez:
That's a nice find @wolfgangwalther. I also read about locust being slow before, but I didn't think it'd make that much of a difference.

> I will try to give vegeta a test run

If vegeta has issues, I can vouch for [k6](https://github.com/loadimpact/k6). For my local `db-pool=1` tests, I used it with a single [virtual user](https://k6.io/docs/cloud/cloud-faq/what-are-vus-virtual-users), which for the unprepared/prepared comparison, tended to give `100 req/s` more in favor of prepared statements(don't remember the numbers exactly but it was like `450 req/s` vs `550 req/s`).

wolfgangwalther:
Although `vegeta` is much nicer to handle, the numbers still jump up and down for me. That is `req/s` across the whole run. The more test runs I made, the more it became clear to me that average `req/s` is not the right metric for our devtools and CI.

We have a lot of unrelated factors that contribute to the performance of a single request (other processes running, cpu scheduling, even automatically adjusted cpu frequency and stuff like that) that create a lot of noise. Every once in a while there will be an outlier with huge response times - and that will kill the whole average. I found that the median response time (reported as 50th percentile) was already much more reliable.

After upgrading nix and `vegeta` to the latest version, I also had access to the minimum response time. This is **much more** reliable (see https://github.com/PostgREST/postgrest/pull/1600#issuecomment-739551613 for some numbers). Minimum latency is a measurement of **peak performance**, when all the other conditions are in our favor. As such it is the most direct way to measure the performance of the actual code. The number itself might not have much meaning in a real world use-case, because there those other factors are important - but we don't want to measure the performance of the overall system. Just the code we've written, ideally only the changes we made.

In my tests, I used this PR and made a couple of changes:
- removed nginx
- updated nix
- changed postgrest.conf to create `/tmp/postgrest.socket` (should probably create a tmp file with `mktemp` for this)
- used the following command to run vegeta:
```
${vegeta}/bin/vegeta -cpus 1 \
  attack -max-workers 1 -workers 1 -unix-socket /tmp/postgrest.socket \
  -targets ./test/loadtest.http -rate 0 -duration 30s "$@" \
  | ${vegeta}/bin/vegeta report
```
- tests were just one request so far, so `./test/loadtest.http` looks like this:
```
GET http://postgrest/orders?select=name
```

I didn't test again with locust, but I would assume if we could find some kind of peak performance metric there, this would also be more reliable. However, I think the upside for vegeta as our test-runner is quite high, see my comment above about criteria.

> If vegeta has issues, I can vouch for [k6](https://github.com/loadimpact/k6). For my local `db-pool=1` tests, I used it with a single [virtual user](https://k6.io/docs/cloud/cloud-faq/what-are-vus-virtual-users), which for the unprepared/prepared comparison, tended to give `100 req/s` more in favor of prepared statements(don't remember the numbers exactly but it was like `450 req/s` vs `550 req/s`).

I couldn't find any unix socket support mentioned in the docs. But apart from that, it looks like it supports everything we need here as well. The "1 virtual user" would probably work as well. However, the virtual user concept + the requests written in javascript surely add overhead.

With vegeta I am getting ~ 2400 req/s with a single process and no concurrency. For peak performance measurements this is good, because the more requests we can make, the higher the chance to find that "true minimum". Also, to be fair, I did not run k6 on my machine and don't know how well it performs here.

wolfgangwalther:
> A/B testing could be possible based on our nightly releases - let's think about this.

Nightly builds could be a problem, because this will be the static build - while for the actual test we want to switch to the incremental build, right?

Maybe we can do that on a commit basis. So `postgrest-loadtest <sha1>` would compare the current working directory vs that commit.

This could be done by something like this (inspired from [here](https://stackoverflow.com/a/15250398)), in parts pseudo-code as comments:

```
# create a temp directory: /tmp/postgrest-sha1

git archive  <sha1> | tar -x -C /tmp/postgrest-sha1

cp -r dist-newstyle /tmp/postgrest-sha1 # to benefit from incremental builds for small differences

# build in temp directory
# launch postgrest in temp directory

# build in working directory
# launch postgrest in working directory

# hit both instances hard
```

steve-chavez:
> I found that the median response time (reported as 50th percentile) was already much more reliable.

That'is really interesting @wolfgangwalther. I've just found that the median is pretty stable on my tests on https://github.com/PostgREST/postgrest/pull/1600#issuecomment-739471788. I'll share the k6 output of 5 runs here, see the `http_req_duration med`:

<details>
<summary>med=1.41ms, 578.878085/s</summary>
<br>
<pre>
    data_received..............: 5.3 MB 176 kB/s
    data_sent..................: 1.9 MB 65 kB/s
    failed requests............: 0.00%  ✓ 0   ✗ 17433
    http_req_blocked...........: avg=5.57µs  min=3.21µs   med=5.09µs  max=516.57µs p(90)=6.57µs   p(95)=7.61µs
    http_req_connecting........: avg=23ns    min=0s       med=0s      max=415.78µs p(90)=0s       p(95)=0s
    http_req_duration..........: avg=1.52ms  min=957.23µs med=1.41ms  max=145.69ms p(90)=1.66ms   p(95)=1.76ms
    http_req_receiving.........: avg=98.42µs min=49.37µs  med=97.74µs max=704.7µs  p(90)=125.98µs p(95)=136.43µs
    http_req_sending...........: avg=31.93µs min=14.22µs  med=28.37µs max=527.87µs p(90)=50.95µs  p(95)=58.36µs
    http_req_tls_handshaking...: avg=0s      min=0s       med=0s      max=0s       p(90)=0s       p(95)=0s
    http_req_waiting...........: avg=1.39ms  min=838.26µs med=1.28ms  max=145.33ms p(90)=1.53ms   p(95)=1.61ms
    http_reqs..................: 17433  578.878085/s
    iteration_duration.........: avg=1.7ms   min=1.1ms    med=1.59ms  max=146.79ms p(90)=1.86ms   p(95)=1.95ms
    iterations.................: 17433  578.878085/s
    vus........................: 1      min=1 max=1
    vus_max....................: 1      min=1 max=1                                                    
</pre>
</details>

<details>
<summary>med=1.41ms, 582.899933/s</summary>
<br>
<pre>
    data_received..............: 5.4 MB 178 kB/s
    data_sent..................: 2.0 MB 65 kB/s
    failed requests............: 0.00%  ✓ 0   ✗ 17555
    http_req_blocked...........: avg=5.4µs   min=3.29µs   med=4.89µs  max=503.17µs p(90)=6.38µs  p(95)=7.39µs
    http_req_connecting........: avg=23ns    min=0s       med=0s      max=404.73µs p(90)=0s      p(95)=0s
    http_req_duration..........: avg=1.51ms  min=971.63µs med=1.41ms  max=81.52ms  p(90)=1.66ms  p(95)=1.75ms
    http_req_receiving.........: avg=98.34µs min=49.09µs  med=97.35µs max=669.27µs p(90)=126.5µs p(95)=136.29µs
    http_req_sending...........: avg=31.99µs min=15.15µs  med=28.66µs max=538.74µs p(90)=51.16µs p(95)=57.94µs
    http_req_tls_handshaking...: avg=0s      min=0s       med=0s      max=0s       p(90)=0s      p(95)=0s
    http_req_waiting...........: avg=1.38ms  min=856.02µs med=1.28ms  max=81.39ms  p(90)=1.52ms  p(95)=1.61ms
    http_reqs..................: 17555  582.899933/s
    iteration_duration.........: avg=1.69ms  min=1.1ms    med=1.6ms   max=81.72ms  p(90)=1.85ms  p(95)=1.95ms
    iterations.................: 17555  582.899933/s
    vus........................: 1      min=1 max=1
    vus_max....................: 1      min=1 max=1                                                   
</pre>
</details>

<details>
<summary>med=1.41ms, 570.758333/s</summary>
<br>
<pre>
    data_received..............: 5.2 MB 174 kB/s
    data_sent..................: 1.9 MB 64 kB/s
    failed requests............: 0.00%  ✓ 0   ✗ 17189
    http_req_blocked...........: avg=5.63µs   min=3.19µs   med=5.18µs   max=328.34µs p(90)=6.81µs   p(95)=7.77µs
    http_req_connecting........: avg=12ns     min=0s       med=0s       max=216.94µs p(90)=0s       p(95)=0s
    http_req_duration..........: avg=1.54ms   min=982.29µs med=1.41ms   max=184.67ms p(90)=1.7ms    p(95)=1.86ms
    http_req_receiving.........: avg=101.12µs min=50.62µs  med=100.81µs max=702.54µs p(90)=129.31µs p(95)=140.84µs
    http_req_sending...........: avg=32.45µs  min=14.74µs  med=29.39µs  max=632.58µs p(90)=50.31µs  p(95)=58.57µs
    http_req_tls_handshaking...: avg=0s       min=0s       med=0s       max=0s       p(90)=0s       p(95)=0s
    http_req_waiting...........: avg=1.41ms   min=884.56µs med=1.28ms   max=184.53ms p(90)=1.55ms   p(95)=1.71ms
    http_reqs..................: 17189  570.758333/s
    iteration_duration.........: avg=1.73ms   min=1.11ms   med=1.6ms    max=184.84ms p(90)=1.9ms    p(95)=2.07ms
    iterations.................: 17189  570.758333/s
    vus........................: 1      min=1 max=1
    vus_max....................: 1      min=1 max=1                                                 
</pre>
</details>

<details>
<summary>med=1.4ms, 583.66937/s</summary>
<br>
<pre>
    data_received..............: 5.4 MB 178 kB/s
    data_sent..................: 2.0 MB 65 kB/s
    failed requests............: 0.00%  ✓ 0   ✗ 17577
    http_req_blocked...........: avg=5.56µs  min=3.18µs   med=5.14µs  max=347.51µs p(90)=6.79µs   p(95)=7.74µs
    http_req_connecting........: avg=14ns    min=0s       med=0s      max=261.8µs  p(90)=0s       p(95)=0s
    http_req_duration..........: avg=1.51ms  min=941.61µs med=1.4ms   max=153.5ms  p(90)=1.64ms   p(95)=1.73ms
    http_req_receiving.........: avg=98.86µs min=47.42µs  med=98.56µs max=633.2µs  p(90)=126.63µs p(95)=136.6µs
    http_req_sending...........: avg=32.62µs min=15.17µs  med=29.45µs max=507.07µs p(90)=52.19µs  p(95)=58.61µs
    http_req_tls_handshaking...: avg=0s      min=0s       med=0s      max=0s       p(90)=0s       p(95)=0s
    http_req_waiting...........: avg=1.38ms  min=833.4µs  med=1.26ms  max=153.3ms  p(90)=1.5ms    p(95)=1.59ms
    http_reqs..................: 17577  583.66937/s
    iteration_duration.........: avg=1.69ms  min=1.07ms   med=1.58ms  max=153.7ms  p(90)=1.83ms   p(95)=1.93ms
    iterations.................: 17577  583.66937/s
    vus........................: 1      min=1 max=1
    vus_max....................: 1      min=1 max=1                                           
</pre>
</details>

<details>
<summary>med=1.42ms, 573.985382/s</summary>
<br>
<pre>
    data_received..............: 5.3 MB 175 kB/s
    data_sent..................: 1.9 MB 64 kB/s
    failed requests............: 0.00%  ✓ 0   ✗ 17285
    http_req_blocked...........: avg=5.62µs  min=3.31µs   med=5.17µs  max=399.4µs  p(90)=6.57µs   p(95)=7.66µs
    http_req_connecting........: avg=14ns    min=0s       med=0s      max=251.85µs p(90)=0s       p(95)=0s
    http_req_duration..........: avg=1.54ms  min=951.16µs med=1.42ms  max=286.92ms p(90)=1.67ms   p(95)=1.76ms
    http_req_receiving.........: avg=99.29µs min=49.7µs   med=98.4µs  max=444.76µs p(90)=127.47µs p(95)=137.89µs
    http_req_sending...........: avg=32.45µs min=15.21µs  med=29.05µs max=579.37µs p(90)=51.42µs  p(95)=59µs
    http_req_tls_handshaking...: avg=0s      min=0s       med=0s      max=0s       p(90)=0s       p(95)=0s
    http_req_waiting...........: avg=1.4ms   min=840.34µs med=1.28ms  max=286.77ms p(90)=1.53ms   p(95)=1.62ms
    http_reqs..................: 17285  573.985382/s
    iteration_duration.........: avg=1.72ms  min=1.08ms   med=1.6ms   max=287.18ms p(90)=1.86ms   p(95)=1.96ms
    iterations.................: 17285  573.985382/s
    vus........................: 1      min=1 max=1
    vus_max....................: 1      min=1 max=1                                           
</pre>
</details>

(For my tests, the minimum response time, `http_req_duration min`, doesn't seem to be as stable.)

Prior to the prepared statements change I'm getting a stable `med=2 ms`.

PR #1611, opened 2020-10-06
Allow wai-middleware-static 0.9

robx:
Changelog:
- Only serve static files on HEAD or GET requests.

The old behaviour seemed wrong, but I'm not 100% sure we don't rely on it. Let's see what the tests say. :)

Stackage tracking issue: https://github.com/commercialhaskell/stackage/issues/5672


steve-chavez:
LGTM!

Already made a new revision: https://hackage.haskell.org/package/postgrest-7.0.1/revisions/

PR #1612, opened 2020-10-06
WIP: Get DbStructure in a single query

monacoremo:
This is a new attempt at https://github.com/PostgREST/postgrest/pull/1513

Currently the DbStructure ("schema cache") is loaded based on multiple queries. This has some performance issues for large schemas and the `DbStructure.hs` file has become difficult to maintain. Getting everything we need in one query might make it easier to debug PostgREST (by 'seeing what PostgREST sees') and accelerate the development of new features/improvements. 

# How it works

Differently from the previous attempt, this keeps any changes to the existing schema cache and queries as minimal as possible for now while adding a new way to get the schema cache.

The new way to get the schema cache can be activated via the environment variable `POSTGREST_MEGAQUERY`. For example, run `POSTGREST_MEGAQUERY=1 postgrest path/to/postgrest.conf` to try the new mode.

This means that PostgREST will behave as normal usually, but the new DbStructure path can be switched on for automated or manual testing of the new path

Currently, only part of the data is sourced from the new query, and it falls back to the standard DbStructure data for the other parts I haven’t yet solved (currently version, tables and procs are sourced from the new path, everything else from the old one). This should enable nice incremental development with diffing of results where necessary to fix bugs

If you have an idea for a less silly name, please let me know :-) I would change the name of the new module and the name of the env variable accordingly.

# Next steps tbd
- Develop this step by step, keeping all tests passing until the new path is completely independent from the old one

Then, possibly:
- Merge (earliest after the next release, but don’t think this will progress that fast)
- Integrate new path in CI (should be as simple as adding POSTGREST_MEGAQUERY=1 postgrest-test-spec-all)
- release beta versions and have power users test the new path with the env flag
- Transitional releases (tbd if needed?)
- One release with both paths, old one as default
- One release with both paths, new one as default
- At some point, kill the old path, reap all the simplification gains (and possibly performance, e.g. from more efficient data structures)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -12,7 +12,7 @@ let
   src =
     pkgs.lib.sourceFilesBySuffices
       (pkgs.gitignoreSource ./.)
-      [ ".cabal" ".hs" ".lhs" "LICENSE" ];
+      [ ".cabal" ".hs" ".lhs" ".sql" "LICENSE" ];
 
   # Commit of the Nixpkgs repository that we want to use.
   nixpkgsVersion =
--- a/default.nix
+++ b/default.nix
@@ -111,7 +111,9 @@ rec {
 
   # Scripts for running tests.
   tests =
-    pkgs.callPackage nix/tests.nix { inherit postgrest postgrestStatic postgrestProfiled postgresqlVersions; };
+    pkgs.callPackage nix/tests.nix {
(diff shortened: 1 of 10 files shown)


monacoremo:
Closing for now as main diverged signficantly, I'll take a new stab at it in time.

Issue #1466, opened 2020-03-27
No way to change RPC return template?


ppKrauss:
The JSON return template for [rpc](http://postgrest.org/en/v6.0/api.html#stored-procedures) of a function that returns JSON is <br/> &nbsp; `[{"functionName":R}]` <br/>but sometimes we need only `R`.

Is it possible, by `postgrest.conf`  or another simple adaptation of PostgREST?

PS: on my [`GET /rpc/add_them?a=1&b=2`](http://postgrest.org/en/v6.0/api.html#immutable-and-stable-functions) it is returning `[{"add_them":3}]`.

steve-chavez:
Reload the schema cache http://postgrest.org/en/v6.0/admin.html#schema-reloading.

ppKrauss:
Hi @steve-chavez 
> 
> Reload the schema cache http://postgrest.org/en/v6.0/admin.html#schema-reloading.

How to `killall -SIGUSR1 postgrest` (error "postgrest: no process found") for a Docker process? 

Documentation (your  link) need some alert about Docker's Schema Reloading

jpagex:
You can run either `docker kill -s SIGUSR1 CONTAINER` or `docker-compose kill -s SIGUSR1 SERVICE`.

ppKrauss:
Reopening for v7.  The `SETOF JSONb` is with a bug. Example:

```sql
CREATE or replace FUNCTION API.uri_dispatcher(uri text DEFAULT '') RETURNS JSONb AS $f$
  SELECT '[{"id":1,"test":"t1"},{"id":2,"test":"t2"}]'::jsonb
$f$ language SQL immutable;

CREATE or replace FUNCTION API.uri_dispatcher_tuple(uri text DEFAULT '') RETURNS SETOF JSONb AS $f$
  SELECT '[{"id":1,"test":"t1"},{"id":2,"test":"t2"}]'::jsonb
$f$ language SQL immutable;
```
1. when create functions the return is ugly, with function names.
2. after something as docker's  SIGUSR1  the `rpc/uri_dispatcher` return expected JSON, clean.<br/>**But** `rpc/uri_dispatcher_tuple` not.

PR #1582, opened 2020-09-15
Add `pgrst.accept` setting to RPC

steve-chavez:
Allows setting a custom media type for the [response format](http://postgrest.org/en/v7.0.0/api.html#response-format)(`Accept` and `Content-Type` headers) by using the `pgrst.accept` setting on RPC.

Basically, it serves as an escape hatch for when the user wants an `application/msgpack` or an `image/png` response. Example:

```sql
-- with this function
create or replace function ret_image(name text) returns bytea as $$
  select img from images where name = $1;
$$ language sql
set pgrst.accept = 'image/png';

-- you can obtain an `image/png` by doing:
curl -H "Accept: image/png" localhost:3000/rpc/ret_image?name=A
curl -H "Accept: */*" localhost:3000/rpc/ret_image?name=A
curl localhost:3000/rpc/ret_image?name=A
```

Currently it doesn't handle a wildcard media subtype(like `image/*`) but handles the all media types wildcard(`*/*`). Which is good enough for cases like browser `` requests or right click -> view image requests(check these [tests](https://github.com/steve-chavez/postgrest/blob/259d20f46c30c65dc6864e1073ae3a3794950c2f/test/Feature/RpcSpec.hs#L816-L843) and #1462).
`image/*` can be handled in a later enhancement.

Related issues:

- Fixes #1462 and partially addresses #1548 
- Deprecates [raw-media-types](http://postgrest.org/en/v7.0.0/configuration.html#raw-media-types)
- Continues the work on https://github.com/PostgREST/postgrest/pull/1546
- Also fixes appending an erroneous `charset=uft-8` on binary output. Reported on [gitter chat](https://gitter.im/begriffs/postgrest?at=5e4ed61189f30b12651cad0f).

PR #1615, opened 2020-10-08
Fix RPC returns

### LINKED ISSUE #1469: Incorrect behavior of header `Prefer: params=single-object` on overloaded function

### LINKED ISSUE #1584: text/plain requested but more than one column was selected
### Environment

* PostgreSQL version: 9.5
* PostgREST version: 6.0.1, 7.0.1
* Operating system: CentOS 6

### Description of issue

With header 'Accept: text/plain' set, request of output of a function returning  a set of values from single column, ends up with HTTP 406:

wget   -O- --header='Accept: text/plain'   http://127.0.0.1:5438/rpc/test
**HTTP request sent, awaiting response... 406 Not Acceptable**

when same URL is proxied by nginx and requested via Chrome, user sees this :
**{"message":"text/plain requested but more than one column was selected"}**

The pgrest log contains this:
GET /rpc/test HTTP/1.0" 406 - "" "Wget/1.12 (linux-gnu)

when header 'Accept: text/plain' _IS NOT_ set, the URL returns:
[{"test":"AAAAA\n"}]

### Function DDL SQL

CREATE OR REPLACE FUNCTION "api"."test" ()  RETURNS SETOF text
  IMMUTABLE
AS $dbvis$
begin
	 RETURN query select 'AAAAA'::text||chr(10);
end;
$dbvis$ LANGUAGE plpgsql

-- when return type is not "SETOF TEXT", but rather "TEXT",  the output is according the documentation.

wolfgangwalther:
This PR fixes a couple of things regarding the returns of RPC calls:
1. The 1st commit is something that I discovered while fixing the other stuff and has - to my knowledge - not been reported, so far. This comment tells the whole story: `_   -> Scalar qi -- 'b'ase, 'd'omain, 'e'num, 'r'ange`. Domain types were always treated as `Scalar` here. That, of course, is not correct once a domain has a composite base type (this seems to be possible to do starting with PG 11, which might explain why it was done that way earlier on). Test added, test failed - confirmed.
The solution is to add a recursive CTE to the proc query, to replace the return type characteristics with those of the base type for domains. The recursive query makes sure, that this works also for domains of domains of domains.... Schema and name of the base type are taken as well, to make embedding work against domains of table types. Added a test for that as well.

2. When I tried to add the next test case, I was stuck quite a bit of time because of a strange "No function matches the given name and argument types." error. It turns out I had naively used just `json` without `pg_catalog.` prefix as an argument type, but json was overloaded as a table name.
To prevent others stumbling across the same thing when adding new tests, I decided to rename the json table. The 2nd commit contains this refactor. I also changed all occurences of `pg_catalog.json` to `json` in `schema.sql` to make sure that a lot of tests start failing once anybody tries to come up with another table/type named `json` again....

3. Next up in the 3rd commit is #1469, which, imho, has multiple dimensions (see below). The obvious thing here is, that overloading is broken with `Prefer: params=single-object`. I added the test-case that @steve-chavez suggested in #1469 slightly adapted. The problem here is, that `findProc` looks for a function with the number of arguments==1, but `decodeProcs` throws away unnamed arguments. So the `ProcDescription` for this overloaded function contained 0 arguments, was never matched by `findProc` and everything went downhill from there.
The fix is to allow `decodeProcs` to parse unnamed arguments. This does not fix it 100%, yet, as it would not properly parse `character varying`. But since we only care for unnamed `json` arguments anyway, that should not be a problem. If somebody wanted to create arguments with spaces in their name, that could lead to problems as well... but I guess that's for another day. I left a todo comment for future reference. Resolves #1469.

4. Next, another small refactor of tests, because the next commit below initially failed on some of the tests due to whitespace changes in the json response. To keep the other commit a bit easier to review, I moved this refactor into a separate one.
To properly test for json responses, the `[str|...|]` QuasiQuoter was replaced with `[json|...|` everywhere. `shouldRespondWith` is overloaded when the json qq is used, so that the response status and content type header (defaults to `application/json`) are checked as well. Because of that some of the tests had to be fixed to test for the singular header. The tests that I touched this way, I also reformatted a bit, to match with the format that is used in most of the other tests.

5. Looking again at #1469, there is something else happening as well: Because the overloading detection was broken, the called RPC was not found in the definitions. When the function returns a scalar (like `json` in this case) it returns the kind of "ugly" json object with the function name as a key. This is normally prevented by the `returnsScalar` branch, which is not executed in this case. The same thing happens when the schema cache is stale (e.g. because a new function was added) or the function returns SetOf Scalar (e.g. `SETOF json` as mentioned in https://github.com/PostgREST/postgrest/issues/1466#issuecomment-694119550).
This can be improved on by moving the whole scalar/composite decision into SQL. To do this, I first added a test-case that creates a new RPC on-the-fly without cache reload. This was therefore consistently called without the `returnsScalar` branch, before my changes were made.
Now, the basic idea of this change is to give the function a consistent name with `AS pgrst_scalar` and then, once the result's row has been encoded in a json object, check whether the key `pgrst_scalar` is part of this object (scalar case) or not (composite case). To do this, the encoding of each row and the aggregation in a json array have to separated. Before, both had been done by `json_agg`, now `to_jsonb` is used to first encode in json, then make the scalar/composite decision, then use `json_agg` at last. The intermediary type has to be `jsonb`, because the `?` operator ("does key exist?") is only available for `jsonb` and not `json`. The intermediate jsonb type results in the whitespace changes mentioned in the refactor above.
Additionally `returnsScalar` is replaced with `returnsSingle` to be able to still tell whether an array or a scalar/object should be returned. This decision can not be made yet in SQL, although I already have some ideas for that, too.
I did add a test for #1584 which is passing now as well. Resolves #1584.

The last part changes the return values of RPCs in some ways that break existing tests. I left those tests unchanged for now, to have them shown up with expected and new output in the CI test output. Those broken tests boil down to 3 core changes:
- RPCs with a return type of `Single Composite X`, e.g. `RETURNS table_type`, now return an object instead of an array with exactly one object. This was actually something that bothered me a few times already. When I have a function that never returns more than 1 row per definition (no setof), I don't want to receive an array - and I don't want to have to set the `application/vnd.pgrst.object+json` accept header all the time.
Example:
```
  test/Feature/RpcSpec.hs:283:9:
  5) Feature.RpcSpec, remote procedure call, proc return types, returns single row from table
       body mismatch:
         expected: [{"id":2}]
         but got:  {"id": 2}
```
- RPCs with a return type of `SetOf Scalar Y`, e.g. `RETURNS SETOF int`, now return a json array with those scalars - and not an array of objects with a single key (= function's name). This has been considered a bug in various issues, yet it's coded into some tests that expect exactly this kind of response.
Example:
```
  test/Feature/RpcSpec.hs:242:9:
  2) Feature.RpcSpec, remote procedure call, proc return types, returns setof integers
       body mismatch:
         expected: [{"ret_setof_integers":1},{"ret_setof_integers":2},{"ret_setof_integers":3}]
         but got:  [1, 2, 3]
```
- RPCs with `INOUT / OUT` params now return an object all the time. Before the change, a single `OUT` param would have lead to a scalar output, while multiple `OUT` params would have been an object. This was quite inconsistent.
Example:
```
  test/Feature/RpcSpec.hs:416:9:
  7) Feature.RpcSpec, remote procedure call, procs with OUT/INOUT params, returns a scalar result when there is a single OUT param
       body mismatch:
         expected: 6
         but got:  {"num_plus_one": 6}
```

I am not sure whether those last two bullet points were actually designed that way on purpose or whether they were just accepted as fact with the current implementation. If all those changes are ok to do, I will change the tests accordingly. We will need to mention those explicitely in the changelog / a migration guide / an appropriate place, of course.

There's one more test that is failing right now, where I don't know how we should proceed with that, yet. This is whenever binary output is requested without `select=` from a RPC that returns a composite type. The test output changes as follows:
```
  test/Feature/RpcSpec.hs:655:11:
  13) Feature.RpcSpec, remote procedure call, binary output, Proc that returns rows, fails if a single column is not selected
       status mismatch:
         expected: 406
         but got:  400
       body mismatch:
         expected: {"message":"application/octet-stream requested but more than one column was selected"}
         but got:  {"hint":null,"details":null,"code":"42703","message":"column _postgrest_t.pgrst_scalar does not exist"}
```
Unfortunately, it's not possible to detect this case anymore before making the request, because the scalar/composite decision is made in SQL. There's a couple of ways to go forward with this:
- leave the error message like that (pgrst_scalar does not exist...)
- catch this exact error in `Errors.hs` and rewrite it to the one it was before (not sure whether that's even possible)
- delay the proper err msg until we have the possibility to create SQL functions: In that case we could raise a custom error from the query itself, when we hit that situation.

Since all those changes are closely related and partly dependent on each other, I put this all into one PR. To make reviewing easier, I tried to split it into easy-to-consume commits, despite some refactors inbetween.

### DIFF
--- a/src/PostgREST/Private/QueryFragment.hs
+++ b/src/PostgREST/Private/QueryFragment.hs
@@ -90,11 +90,15 @@ asCsvF = asCsvHeaderF <> " || '\n' || " <> asCsvBodyF
       ")"
     asCsvBodyF = "coalesce(string_agg(substring(_postgrest_t::text, 2, length(_postgrest_t::text) - 2), '\n'), '')"
 
-asJsonF :: SqlFragment
-asJsonF = "coalesce(json_agg(_postgrest_t), '[]')::character varying"
+asJsonF :: Bool -> SqlFragment
+asJsonF returnsScalar
+  | returnsScalar = "coalesce(json_agg(_postgrest_t.pgrst_scalar), '[]')::character varying"
+  | otherwise     = "coalesce(json_agg(_postgrest_t), '[]')::character varying"
 
-asJsonSingleF :: SqlFragment --TODO! unsafe when the query actually returns multiple rows, used only on inserting and returning single element
-asJsonSingleF = "coalesce(string_agg(row_to_json(_postgrest_t)::text, ','), '')::character varying "
+asJsonSingleF :: Bool -> SqlFragment --TODO! unsafe when the query actually returns multiple rows, used only on inserting and returning single element
+asJsonSingleF returnsScalar
+  | returnsScalar = "coalesce(string_agg(to_json(_postgrest_t.pgrst_scalar)::text, ','), '')::character varying"
+  | otherwise     = "coalesce(string_agg(to_json(_postgrest_t)::text, ','), '')::character varying"
 
 asBinaryF :: FieldName -> SqlFragment
 asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldName <> ", ''), '')"
--- a/src/PostgREST/Statements.hs
+++ b/src/PostgREST/Statements.hs
@@ -74,8 +74,8 @@ createWriteStatement selectQuery mutateQuery wantSingle isInsert asCsv rep pKeys
   bodyF
     | rep `elem` [None, HeadersOnly] = "''"
     | asCsv = asCsvF
-    | wantSingle = asJsonSingleF
-    | otherwise = asJsonF
+    | wantSingle = asJsonSingleF False
+    | otherwise = asJsonF False
 
   selectF
     -- prevent using any of the column names in ?select= when no response is returned from the CTE
@@ -108,9 +108,9 @@ createReadStatement selectQuery countQuery isSingle countTotal asCsv binaryField
 
   bodyF
     | asCsv = asCsvF
-    | isSingle = asJsonSingleF
+    | isSingle = asJsonSingleF False
     | isJust binaryField = asBinaryF $ fromJust binaryField
-    | otherwise = asJsonF
+    | otherwise = asJsonF False
 
   decodeStandard :: HD.Result ResultsWithCount
   decodeStandard =
@@ -128,10 +128,10 @@ standardRow = (,,,,,) <$> nullableColumn HD.int8 <*> column HD.int8
 
 type ProcResults = (Maybe Int64, Int64, ByteString, Either SimpleError [GucHeader], Either SimpleError (Maybe Status))
 
-callProcStatement :: Bool -> H.Snippet -> H.Snippet -> H.Snippet -> Bool ->
-                     Bool -> Bool -> Bool -> Bool -> Maybe FieldName -> PgVersion -> Bool ->
+callProcStatement :: Bool -> Bool -> H.Snippet -> H.Snippet -> H.Snippet -> Bool ->
+                     Bool -> Bool -> Bool -> Maybe FieldName -> PgVersion -> Bool ->
                      H.Statement () ProcResults
-callProcStatement returnsScalar callProcQuery selectQuery countQuery countTotal isSingle asCsv asBinary multObjects binaryField pgVer =
+callProcStatement returnsScalar returnsSingle callProcQuery selectQuery countQuery countTotal asSingle asCsv multObjects binaryField pgVer =
   H.dynamicallyParameterized snippet decodeProc
   where
     snippet =
@@ -149,16 +149,12 @@ callProcStatement returnsScalar callProcQuery selectQuery countQuery countTotal 
     (countCTEF, countResultF) = countF countQuery countTotal
 
     bodyF
-     | returnsScalar = scalarBodyF
-     | isSingle     = asJsonSingleF
-     | asCsv = asCsvF
+     | asSingle           = asJsonSingleF returnsScalar
+     | asCsv              = asCsvF
      | isJust binaryField = asBinaryF $ fromJust binaryField
-     | otherwise = asJsonF
-
-    scalarBodyF
-     | asBinary    = asBinaryF "pgrst_scalar"
-     | multObjects = "json_agg(_postgrest_t.pgrst_scalar)::character varying"
-     | otherwise   = "(json_agg(_postgrest_t.pgrst_scalar)->0)::character varying"
+     | returnsSingle
+       && not multObjects = asJsonSingleF returnsScalar
+     | otherwise          = asJsonF returnsScalar
 
     decodeProc :: HD.Result ProcResults
     decodeProc =
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -297,9 +297,8 @@ app dbStructure conf apiRequest =
                 preferParams = iPreferParameters apiRequest
                 pq = requestToCallProcQuery (QualifiedIdentifier pdSchema pdName) (specifiedProcArgs (iColumns apiRequest) proc)
                        (iPayload apiRequest) returnsScalar preferParams returning
-                stm = callProcStatement returnsScalar pq q cq shouldCount (contentType == CTSingularJSON)
-                        (contentType == CTTextCSV) (contentType `elem` rawContentTypes) (preferParams == Just MultipleObjects)
-                        bField pgVer prepared
+                stm = callProcStatement returnsScalar returnsSingle pq q cq shouldCount (contentType == CTSingularJSON)
+                        (contentType == CTTextCSV) (preferParams == Just MultipleObjects) bField pgVer prepared
               row <- H.statement mempty stm
               let (tableTotal, queryTotal, body, gucHeaders, gucStatus) = row
                   gucs =  (,) <$> gucHeaders <*> gucStatus
@@ -351,6 +350,10 @@ app dbStructure conf apiRequest =
           case iTarget apiRequest of
             TargetProc proc _ -> procReturnsScalar proc
             _                 -> False
+        returnsSingle =
+          case iTarget apiRequest of
+            TargetProc proc _ -> procReturnsSingle proc
+            _                 -> False
         pgVer = pgVersion dbStructure
         profileH = contentProfileH <$> iProfile apiRequest
 
@@ -401,7 +404,10 @@ responseContentTypeOrError accepts rawContentTypes action target = serves conten
 -}
 binaryField :: ContentType -> [ContentType] -> Bool -> ReadRequest -> Either Response (Maybe FieldName)
 binaryField ct rawContentTypes isScalarProc readReq
-  | isScalarProc = Right Nothing
+  | isScalarProc =
+      if ct `elem` rawContentTypes
+        then Right $ Just "pgrst_scalar"
+        else Right Nothing
   | ct `elem` rawContentTypes =
       let fieldName = headMay fldNames in
       if length fldNames == 1 && fieldName /= Just "*"
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -226,8 +226,12 @@ procsSqlQuery = [q|
     tn.nspname AS schema,
     COALESCE(comp.relname, t.typname) AS name,
     p.proretset AS rettype_is_setof,
-    -- Only pg pseudo type that is a row type is 'record'
-    (t.typtype = 'c' or t.typtype = 'p' and t.typname = 'record') AS rettype_is_composite,
+    (t.typtype = 'c'
+     -- Only pg pseudo type that is a row type is 'record'
+     or t.typtype = 'p' and t.typname = 'record'
+     -- if any INOUT or OUT arguments present, treat as composite
+     or COALESCE(proargmodes::text[] && '{b,o}', false)
+    ) AS rettype_is_composite,
     p.provolatile
   FROM pg_proc p
   JOIN pg_namespace pn ON pn.oid = p.pronamespace
--- a/src/PostgREST/Types.hs
+++ b/src/PostgREST/Types.hs
@@ -203,7 +203,13 @@ specifiedProcArgs keys proc =
 procReturnsScalar :: ProcDescription -> Bool
 procReturnsScalar proc = case proc of
   ProcDescription{pdReturnType = (Single (Scalar _))} -> True
+  ProcDescription{pdReturnType = (SetOf (Scalar _))}  -> True
   _                                                   -> False
+
+procReturnsSingle :: ProcDescription -> Bool
+procReturnsSingle proc = case proc of
+  ProcDescription{pdReturnType = (Single _)} -> True
+  _                                          -> False
 
 procTableName :: ProcDescription -> Maybe TableName
 procTableName proc = case pdReturnType proc of
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -821,14 +821,22 @@ spec actualPgVersion = do
     it "fails if a single column is not selected" $ do
       request methodGet "/images?select=img,name&name=eq.A.png" (acceptHdrs "application/octet-stream") ""
         `shouldRespondWith`
-        [json| {"message":"application/octet-stream requested but more than one column was selected"} |]
-        { matchStatus = 406
-        , matchHeaders = [matchContentTypeJson]
-        }
-      request methodGet "/images?select=*&name=eq.A.png" (acceptHdrs "application/octet-stream") ""
-        `shouldRespondWith` 406
-      request methodGet "/images?name=eq.A.png" (acceptHdrs "application/octet-stream") ""
-        `shouldRespondWith` 406
+          [json| {"message":"application/octet-stream requested but more than one column was selected"} |]
+          { matchStatus = 406 }
+
+      request methodGet "/images?select=*&name=eq.A.png"
+          (acceptHdrs "application/octet-stream")
+          ""
+        `shouldRespondWith`
+          [json| {"message":"application/octet-stream requested but more than one column was selected"} |]
+          { matchStatus = 406 }
+
+      request methodGet "/images?name=eq.A.png"
+          (acceptHdrs "application/octet-stream")
+          ""
+        `shouldRespondWith`
+          [json| {"message":"application/octet-stream requested but more than one column was selected"} |]
+          { matchStatus = 406 }
 
     it "concatenates results if more than one row is returned" $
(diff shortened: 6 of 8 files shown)


src/PostgREST/Statements.hs
```diff
@@ -128,36 +128,39 @@ standardRow = (,,,,,) <$> nullableColumn HD.int8 <*> column HD.int8
...
-callProcStatement returnsScalar callProcQuery selectQuery countQuery countTotal isSingle asCsv asBinary multObjects binaryField pgVer =
+callProcStatement callProcQuery selectQuery countQuery countTotal isSingle asCsv asBinary multObjects binaryField pgVer =
   unicodeStatement sql (param HE.unknown) decodeProc True
   where
     sql = [qc|
-      WITH {sourceCTEName} AS ({callProcQuery})
+      WITH {sourceCTEName} AS ({callProcQuery}),
+      pgrst_select AS ({selectQuery}),
+      -- using to_json()::jsonb here, because pg9.4 does not have to_jsonb() yet
+      pgrst_to_json AS (SELECT to_json(pgrst_select)::jsonb AS pgrst_json FROM pgrst_select)
```
wolfgangwalther, line 141: Just using `to_jsonb` would be a lot better, probably also for performance.

We might consider dropping support for pg 9.4 - it's final release was in February and it's not supported anymore.

steve-chavez: It'd be better to drop pg 9.4 on another PR, I think there are changes we need to make on nix.

@monacoremo Perhaps you'd like to work on this?

wolfgangwalther: No, the tests are broken on purpose. I admit, my text above is quite lengthy.... so I'm quoting myself here:

> The last part changes the return values of RPCs in some ways that break existing tests. I left those tests unchanged for now, to have them shown up with expected and new output in the CI test output. 

Once we drop support for pg 9.4, we can change `to_json(pgrst_select)::jsonb` to `to_jsonb(pgrst_select)`. So instead of converting twice, we only need to do it once.

I have prepared a branch to drop 9.4 at any point soon. If you're ok with dropping it now, I can open a PR.

wolfgangwalther:
Rebased on the drop-support-pg94 branch and replaced `to_json(...)::jsonb` with `to_jsonb(...)`.

> The last part changes the return values of RPCs in some ways that break existing tests. I left those tests unchanged for now, to have them shown up with expected and new output in the CI test output. 

Added a commit with the changes in expected test output now, CI tests should be all passing now. Will rebase this commit into the one before, later on. Just kept it separate for now, for better visibility what kind of changes this PR introduces.

steve-chavez:
> When I tried to add the next test case, I was stuck quite a bit of time because of a strange "No function matches the given name and argument types." error. It turns out I had naively used just json without pg_catalog.

Great you fixed this one. I also bumped into it a while ago.

> RPCs with a return type of Single Composite X, e.g. RETURNS table_type, now return an object instead of an array with exactly one object.

Agree, this is a more sensible behavior.

> RPCs with a return type of SetOf Scalar Y, e.g. RETURNS SETOF int, now return a json array with those scalars - and not an array of objects with a single key (= function's name)

Yes, this was designed to be this way because when:
```sql
-- you do
select * from test.ret_setof_integers();
-- you get rows, not an array
 ret_setof_integers
--------------------
                  1
                  2
                  3
```
However, as you noted, this is confusing for many users. So I agree with the change, returning arrays is good.

> RPCs with INOUT / OUT params now return an object all the time. Before the change, a single OUT param would have lead to a scalar output, while multiple OUT params would have been an object. This was quite inconsistent.

Hmm.. this was done on purpose, but I can't remember why(maybe a user requested it). I'll have to dig the history to find out.

wolfgangwalther:
Rebased on current master. Moved the changed tests you have approved, @steve-chavez, to the commit before.

The "change expected test output" commit now has only a few changes remaining (single out param and the error message).

wolfgangwalther:
> > RPCs with INOUT / OUT params now return an object all the time. Before the change, a single OUT param would have lead to a scalar output, while multiple OUT params would have been an object. This was quite inconsistent.
> 
> Hmm.. this was done on purpose, but I can't remember why(maybe a user requested it). I'll have to dig the history to find out.

git blame lead me to #974 and this commit https://github.com/PostgREST/postgrest/commit/d1a8c3a6f88cfd70039a2810c49e932f2bca360e.

Looks like that single OUT parameter was not supported before, you fixed the bug and because of how scalar returns worked back then, the result was a scalar return in this case, which you have put in the test-case.

I think changing this makes sense, because whenever you give a return a name (via OUT parameter or TABLE or composite type) this should be reflected in the output. And tbh, I wouldn't know how to prevent this in this patch... :)

wolfgangwalther:
rebased on master.

@steve-chavez does my history digging with git blame answer your question about `OUT` parameter - would you be okay with the change?

Same question for the changed error message - the other remaining test change in the last commit.

wolfgangwalther:
rebased on current master. as only the last two commits are still under discussion here, I branched the first 4 out in #1654, so we can already merge them.

src/PostgREST/QueryBuilder.hs
```diff
@@ -153,23 +153,18 @@ requestToCallProcQuery qi pgArgs returnsScalar preferParams returnings =
...
     sourceBody
       | paramsAsMultipleObjects =
-          if returnsScalar
-            then "SELECT " <> callIt <> " AS pgrst_scalar FROM pgrst_args"
-            else BS.unwords [ "SELECT pgrst_lat_args.*"
-                         , "FROM pgrst_args,"
-                         , "LATERAL ( SELECT " <> returned_columns <> " FROM " <> callIt <> " ) pgrst_lat_args" ]
-      | otherwise =
-          if returnsScalar
-            then "SELECT " <> callIt <> " AS pgrst_scalar"
```
wolfgangwalther, line 163: Note: Having the function call in the select expression list, does also potentially harm performance, because inlining the function is much less likely, given the huge list of conditions:
https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions

Especially this part:
> A [...] function call will be inlined if all of the following conditions are met:
> [...]
> the body contains no aggregate or window function calls, no subqueries, no CTEs, no FROM clause or references to any table or table-like object, none of GROUP BY, HAVING, ORDER BY, DISTINCT, LIMIT, OFFSET, UNION, INTERSECT, EXCEPT

Having the function call in the `FROM` part, will allow much more. So this PR potentially also improves performance, also currently only in theory.

That's, because currently, inlining of non-scalar functions is still prevented by the sub-select in the argument list (`x := (SELECT x FROM pgrst_args LIMIT 1)`, see https://github.com/PostgREST/postgrest/issues/1652#issuecomment-729975165. Once we can fix that and make the change in this PR here, that should lead to better performance for many RPCs.

steve-chavez:
#### OUT param change

> I think changing this makes sense, because whenever you give a return a name (via OUT parameter or TABLE or composite type) this should be reflected in the output.
> @steve-chavez does my history digging with git blame answer your question about the OUT parameter - would you be okay with the change?

@wolfgangwalther Sorry for the late reply. Yes, I agree with that, it makes sense to treat OUT as composite. We should note this on the CHANGELOG in case someone relied on that behavior(which now seems wrong to me).

wolfgangwalther:
Rebased on current #1654, merged the two remaining commits (the second one was just showing the changed tests) and added the `OUT` change to the CHANGELOG.

steve-chavez:
I think this one needs one more rebase(still showing #1654 diff).

wolfgangwalther:
Ah yes, thanks for the heads up. Done!

src/PostgREST/QueryBuilder.hs
wolfgangwalther: Scratch that. Since this is only about scalar functions, those won't be inlined in the `FROM` part at all. So if anything, this change might potentially harm inlining.

I guess I have to re-think the approach in this PR.

wolfgangwalther:
See my comment above about inling:

> Scratch that. Since this is only about scalar functions, those won't be inlined in the FROM part at all. So if anything, this change might potentially harm inlining.
> I guess I have to re-think the approach in this PR.

I'll put this PR on hold for a moment, I will have to think about this again. I am currently doing some testing in regards to RPC performance (#1652 is related) and want to get this right.

wolfgangwalther:
Although I wasn't able to continue the actual performance testing, I was able to think this through a bit more while writing https://github.com/PostgREST/postgrest/pull/1582#issuecomment-740919214.

The current approach in this PR is to query all RPCs (both with scalar and composite return types) the same way and do an "in-SQL" decision about how to format the return type. Most of the issues that this resolves are around the fact that the current code does not treat `SETOF <scalartype>` and single `<compositetype>` properly. There is another way of solving that ("in-haskell"), by changing the current `returnsScalar` to a combination of `returnsScalar` and `returnsSingle`. I already implemented that once, way back in time, .. but lost the code :D. This isn't too hard, though.

The main differences between the two approaches are:
- in-SQL allows better responses with a stale schema cache - although not in all cases
- in-SQL is immune against a faulty `findProc` in haskell
- in-haskell allows a better error message for that one error message that is changed in this PR
- **in-haskell allows to keep scalar functions in the `SELECT` part instead of the `FROM` part of the query, thereby possibly enabling inlining.**

I am not entirely sure about the "single OUT paramter" case, yet, but I'm quite confident that this can still be solved in-haskell as well.

The stale schema cache is really not something that should be priority, more a nice-to-have if possible. Deciding about overloaded functions is a hard business for `findProc` - but we need to rely on that for other things (volatility -> transaction isolation level, returned table type -> embedding, etc.) anway. Inlining is going to give us far more, if it just makes one query for anyone faster, so the in-haskell solution is by far better.

-> Will implement the in-haskell approach again, once I find the time.

wolfgangwalther:
> Will implement the in-haskell approach again,

Done. Test-cases / behaviour are the same, with those exceptions:
- As noted above, I removed the "stale schema cache" test - this won't be possible
- As noted above, I reverted the previously changed error message for one of the binary column errors. Unchanged to master now.
- **new**: After implementing this, I got empty responses for RPC that return `void`. They returned `null` previously. I didn't dig into this deeper, because I felt this was actually better. Nobody needs a return value for a `void` function, so we can as well save the 4 characters in the response.

src/PostgREST/Private/QueryFragment.hs
```diff
@@ -90,11 +90,13 @@ asCsvF = asCsvHeaderF <> " || '\n' || " <> asCsvBodyF
       ")"
     asCsvBodyF = "coalesce(string_agg(substring(_postgrest_t::text, 2, length(_postgrest_t::text) - 2), '\n'), '')"
 
-asJsonF :: SqlFragment
-asJsonF = "coalesce(json_agg(_postgrest_t), '[]')::character varying"
+asJsonF :: Bool -> SqlFragment
+asJsonF False = "coalesce(json_agg(_postgrest_t), '[]')::character varying"
+asJsonF True  = "coalesce(json_agg(_postgrest_t.pgrst_scalar), '[]')::character varying"
```
steve-chavez, line 95: The Bool value here could use some clarification. How about changing the function to:

```suggestion
asJsonF returnsScalar = if returnsScalar
  then "coalesce(json_agg(_postgrest_t.pgrst_scalar), '[]')::character varying"
  else "coalesce(json_agg(_postgrest_t), '[]')::character varying"
```

steve-chavez: It could also be like:

```suggestion
asJsonF returnsScalar | returnsScalar = "coalesce(json_agg(_postgrest_t.pgrst_scalar), '[]')::character varying"
                      | otherwise     = "coalesce(json_agg(_postgrest_t), '[]')::character varying"
```

src/PostgREST/Private/QueryFragment.hs
```diff
@@ -90,11 +90,13 @@ asCsvF = asCsvHeaderF <> " || '\n' || " <> asCsvBodyF
...
 
-asJsonSingleF :: SqlFragment --TODO! unsafe when the query actually returns multiple rows, used only on inserting and returning single element
-asJsonSingleF = "coalesce(string_agg(row_to_json(_postgrest_t)::text, ','), '')::character varying "
+asJsonSingleF :: Bool -> SqlFragment --TODO! unsafe when the query actually returns multiple rows, used only on inserting and returning single element
+asJsonSingleF False = "coalesce(string_agg(to_json(_postgrest_t)::text, ','), '')::character varying"
+asJsonSingleF True  = "coalesce(string_agg(to_json(_postgrest_t.pgrst_scalar)::text, ','), '')::character varying"
```
steve-chavez, line 99: Similarly here:

```suggestion
asJsonSingleF returnsScalar = if returnsScalar
  then "coalesce(string_agg(to_json(_postgrest_t.pgrst_scalar)::text, ','), '')::character varying"
  else "coalesce(string_agg(to_json(_postgrest_t)::text, ','), '')::character varying"
```

steve-chavez:
@wolfgangwalther Liking the new behavior for SETOF scalars. Also for the OUT/INOUT as composite and the void. It overall looks more consistent in code and usage.

It's also great that our new convention for the CHANGELOG leads us to write better commits messages. I'll be doing this myself as well.

Left a couple of reviews above. Besides that, it looks good to merge!

src/PostgREST/Private/QueryFragment.hs
wolfgangwalther: So now you're doing multi-line suggestions... cool :D

I went with the 2nd suggestion, but moved it to a new line. Will push and merge in a sec.

wolfgangwalther:
> It's also great that our new convention for the CHANGELOG leads us to write better commits messages. I'll be doing this myself as well.

Yes, I think that really helps! :)

PR #1616, opened 2020-10-13
make check on current master

wolfgangwalther:
When I run `make check` on the current master branch, I get this diff.

The changes to `stack.yaml.lock` appear whenever I run `make test`.

Seems like the whitespace stuff for imports only appeared after I updated stylish-haskell. Right now, I can't use `make style` because the resulting diff is just too annoying.

Is this something where I have to change my config somehow? Or should we just merge those changes?

### DIFF
--- a/stack.yaml.lock
+++ b/stack.yaml.lock
@@ -31,7 +31,14 @@ packages:
       size: 1644
       sha256: babf32b414f25f790b7a4ce6bae5c960bc51a11a289e7c47335b222e6762560c
   original:
-    hackage: protolude-0.3.0
+    hackage: protolude-0.3.0@sha256:8361b811b420585b122a7ba715aa5923834db6e8c36309bf267df2dbf66b95ef,2693
+- completed:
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
The CI failures don't make any sense to me. I guess they are the result of the changes to stack.yaml.lock?

steve-chavez:
@wolfgangwalther What's your stylish-haskell version? I have:
```
> stylish-haskell --version
stylish-haskell 0.11.0.3
```

> The CI failures don't make any sense to me. 

Weird.. I'm not sure either. It's like the styler messed the files somehow.

steve-chavez:
> make style

Unfortunately that doesn't work for me(`error: attribute 'ghc883' missing`). But that's because I've been using the nix workflow and haven't touched stack for a while.

> 0.12.0.0

Perhaps you could try downgrading to 11.0.3? 12 looks a bit new and there are already new releases with fixes: https://github.com/jaspervdj/stylish-haskell/releases

wolfgangwalther:
Since this passes all tests - any objection to merging this? This would make my life with stack a bit easier.

PR #1617, opened 2020-10-13
Drop support for postgres 9.4

wolfgangwalther:
Postgres 9.4 is EOL since February and #1615 introduces new code, that can be better done without pg 9.4 support. It would make sense to drop pg 9.4 now.

@monacoremo can you have a look at the nix stuff? I just removed everything that looked like 9.4... ;)

@steve-chavez There is a test-suite `PgVersion95Spec`. I understand this as stuff that is only supported from 9.5 on. Since 9.5 is our minimum version now, we can probably move those tests to other files, where they belong. Correct?


steve-chavez:
> There is a test-suite PgVersion95Spec. I understand this as stuff that is only supported from 9.5 on. Since 9.5 is our minimum version now, we can probably move those tests to other files, where they belong. Correct?

@wolfgangwalther Yes, in fact I've already done that, I just didn't merge it(https://github.com/PostgREST/postgrest/pull/1582): https://github.com/PostgREST/postgrest/pull/1582/commits/61a78abfb617ecd8a922954520834eebc4b0a35f

wolfgangwalther:
Now this is interesting. nix-test fails with a strange one. The expected output of the failing test is not touched at all in this PR, but it is changed in https://github.com/PostgREST/postgrest/pull/1603/commits/c9fbcd99b1195c2040bfe08a572138a2517036c0 (that's in another PR!). I noticed the same kind of failure earlier today in another PR (merged already), where it was gone on the next build.

I did make a change to the other PR that triggered a rebuild and then opened this PR - so both pipelines were running at the same time. I think what happens is:
- PR #1 runs nix-build and sends it to cache
- PR #2 runs nix-build and sends it to cache
- PR #1 pulls from cache - but the wrong build from the other PR. That's why it has some random failures.

@monacoremo Can you have a look at this? I think you need we need to tag the builds we're sending to cachix with maybe the commit-id or something?

monacoremo:
@wolfgangwalther Hm, weird. Looks like some issue with the CircleCI caching and/or cabal-install again, which we use to run the test suite interactively. The Nix and Cachix pieces themselves should never create such issues... https://github.com/PostgREST/postgrest/pull/1618 rips out the CircleCI caching and should make the builds completely deterministic. Can you try to rebase on that?

wolfgangwalther:
> Can you try to rebase on that?

I'm not sure whether I would be able to actually reproduce the error in the first place. I'm sure that this PR here would pass easily, when I trigger it and don't trigger any other PRs at the same time. But, when I rebase both PRs now and we get both PRs passing - I would not be convinced that the problem is solved, yet. It could be just coincidence.

wolfgangwalther:
Ok, so it succeeded in both PRs with this. I will now remove the commits again, let's see whether that brings it up again.

wolfgangwalther:
So after I removed the commits, it's now failing again. "Cool!"

wolfgangwalther:
Done. Since it's only those text files that were changed, the CI result will not be different. So ready to merge already.

PR #1618, opened 2020-10-14
Simplify Nix CI jobs

monacoremo:
- Removed CircleCI caching for Nix build jobs. It's finicky (e.g. https://github.com/PostgREST/postgrest/pull/1617#issuecomment-707868410) and is not really worth it build time wise - Nix will always provide all dependencies and we only ever build PostgREST and the test suite with cabal-install, which is relatively quick
- Merged nix-build and nix-test - separating those steps makes things more complicated than they need to be and negatively impacts performance for PRs, as we can't push to cachix for those

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,12 +153,16 @@ jobs:
             postgrest-release-dockerhub
             postgrest-release-dockerhubdescription
 
-  # Build everything in default.nix and push to the Cachix binary cache.
-  nix-build:
-    docker:
-      - image: nixos/nix:2.3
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
I confirmed that this fixes https://github.com/PostgREST/postgrest/pull/1617#issuecomment-707868410.

When I merged this PR into both #1603 and #1617 and they were building at the same time, both were passing.

When I removed those commits again, again building both at the same time, #1617 is now failing again with the error, where the test actually belongs to #1603.

LGTM!

steve-chavez:
Oh. This was a tricky one. Great work guys!

Merging..

PR #1619, opened 2020-10-14
fix appveyor build script

wolfgangwalther:
Seems like the appveyor build is silently throwing errors right now. This is because the error code is swallowed. This change should make it pass on a timeout, but fail on every other error.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@ build_script:
 - stack setup --no-terminal > nul
 # Appveyor has a timeout of 60 mins, building can take longer, limit the time and make sure this succeeds,
 # previous work will get cached and finish on next commit
-- bash -lc "timeout 2700 'C:\projects\postgrest\stack.exe' build -j1 --copy-bins --local-bin-path . || true"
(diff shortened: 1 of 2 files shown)


wolfgangwalther:
So, as expected the appveyor build fails now. 2nd commit tries to fix the windows build error.

wolfgangwalther:
Appveyor is passing again.

@steve-chavez ready to merge.

wolfgangwalther:
The `nix-test` failure in here is just another instance of what is fixed in #1618. So the failure is not related to this PR, still good to merge.

appveyor.yml
```diff
@@ -26,7 +26,7 @@ build_script:
 - stack setup --no-terminal > nul
 # Appveyor has a timeout of 60 mins, building can take longer, limit the time and make sure this succeeds,
 # previous work will get cached and finish on next commit
-- bash -lc "timeout 2700 'C:\projects\postgrest\stack.exe' build -j1 --copy-bins --local-bin-path . || true"
+- bash -lc "timeout 2700 'C:\projects\postgrest\stack.exe' build -j1 --copy-bins --local-bin-path . || (($?==124))"
```
steve-chavez, line 29: Ah, so timeout exits with 124: https://linux.die.net/man/1/timeout

Didn't know that was possible. Great that you have it fixed.

PR #1621, opened 2020-10-14
improve cirrus build time by properly using cabal cache

wolfgangwalther:
So I had a look at the cirrus build, that takes like forever all the time.

Looking at the output, it seems to me like the cabal cache is not updated properly and all dependencies are built again every time. The cache update is based on the contents of `postgrest.cabal` - but since there are version ranges given in this file, I could imagine that cabal still decides to download newer packages and not use those from cache.

I tried removing the `fingerprint_script` line. Now it should default back to `reupload_on_changes: true` and scan the `~/.cabal` folder for changes to upload.

Let's see what happens when I trigger this twice.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,54 +2,32 @@ freebsd_instance:
   image: freebsd-12-1-release-amd64
 
 build_task:
-  env:
-    ASSUME_ALWAYS_YES: YES
+  # caches the freebsd package downloads
+  # saves probably just a couple of seconds, but hey...
+  pkg_cache:
+    folder: /var/cache/pkg
+
   install_script:
-    ## TODO: Some of these workaround steps might be unneeded now
-    ## gmake required, otherwise bsd make is used and some cryptic "No operator" errors are shown
-    - pkg install -y gmake
-    ## gcc required, otherwise warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
-    - pkg install -y gcc
-    - pkg install -y libiconv
-    - pkg install -y gmp
-    - pkg install -y postgresql95-client
-    - pkg install -y pkgconf
-    - pkg install -y ghc
-    - pkg install -y hs-cabal-install
-  cabal_cache:
-    folder: ~/.cabal
+    - pkg install -y postgresql95-client ghc hs-cabal-install
+  
+  # cache the hackage index file and downloads which are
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
Step 1 worked, it uploaded a new cache in this build:

```
SHA for //.cabal is 'b12f27b95be88d8bf978e2c527b3a04c957943c4eab000472cdc154f43612c27'
cabal cache size is 270Mb.
Uploading cache cabal...
```

Let's see whether that helps to speed it up by triggering another build somewhere...

wolfgangwalther:
I found the problem: The cache folder was set to `~/.cabal`. Apparently `~` is somehow resolved to `/` in this environment. This lead to some strange paths like `//.cabal` - and the cache upload step was completely broken. It uploaded the cache into a subdirectory `.cabal` and when it was downloaded again, this resulted in a directory structure like `/.cabal/.cabal/....`.

So I changed the cache folder to `/.cabal` and it seems to work now. One of the last builds had a build-step time of about 2-3 minutes - where it was always >30 minutes before.

I also added another cache for freebsd `pkg` downloads - not sure how much that will be worth.

I'm still going through a couple of runs to see whether any of the installed packages can be omitted, because there is a comment like `## TODO: Some of these workaround steps might be unneeded now`.

wolfgangwalther:
Ok, done.

The pipeline now takes ~5-6 minutes in total compared to 30-40 mins before. Cirrus still takes like 5 min to schedule it, so it's like 10 min in total right now.

Good to merge.

steve-chavez:
> I found the problem: The cache folder was set to ~/.cabal. Apparently ~ is somehow resolved to / in this environment

Glad you could figure this out! :-)

Merging..

wolfgangwalther:
@monacoremo Just a heads up for you... on my last PR CirrusCI now was finished after 4 min total. nix-build-test is already well over 10 minutes and still running... your nix stuff comes in last now... ;)

monacoremo:
@wolfgangwalther great job! And I'm feeling the pressure! 🔥🔥🔥 

There are definitely some more tweaks that we can do, like using `nix-build-uncached` (reviving https://github.com/PostgREST/postgrest/pull/1581) and parallelizing some tests... could be a fun challenge!

We are doing a lot in the Nix job, including building the fully static executable (which takes *hours* from scratch even on my desktop machine), running tests against all Postgres versions, memory and io tests, building the docker image etc., all fully reproducible, so I don't feel too bad that it takes a few minutes :-)

Issue #1233, opened 2019-01-24
Multiple "hops" for embed(Unnest/flatten embeds)


blainehansen:
### Environment

* PostgreSQL version: (if using docker, specify the image)
docker `postgres:10-alpine`
* PostgREST version: (if using docker, specify the image)
docker `postgrest/postgrest` (pulled fresh this morning)
* Operating system:
Ubuntu 18

### Description of issue (less an issue than a question)

Is it possible to do multiple "hops" to embed something?

For example, I have a schema similar to this (the real thing has more columns and checks)

```sql
create table dimension (
	id serial primary key,
	dimension_type_name text not null,

	"text" text not null
);

create table dimension_relationship (
	ancestor_id int references dimension(id) not null,
	descendent_id int references dimension(id) not null,
	check (ancestor_id != descendent_id),
	primary key (ancestor_id, descendent_id)
	-- other checks and indices to ensure these only go one way
);
```

The basic idea behind this structure is that it can store any arbitrary hierarchy of data. The relationships can only go in one direction, but it doesn't have to be tree-like (a dimension can can multiple parents).

I want to get a particular dimension (let's say with `dimension_type_name` "a"), and all the descendent dimensions with `dimension_type_name` "b", in a json structure similar to this:

```json
[{
	// this is a dimension of type "a"
	"id": 1, "text": "stuff",
	// these should all be dimensions of type "b"
	"attributes": [{
		"id": 2, "text": "other"
	}],
	// ...
}]
```

This gets me close:

```bash
"/dimension?
	select=*,
	relationships:dimension_relationship.ancestor_id(
		attribute:dimension.descendent_id(*)
	)
	&dimension_type_name=eq.a
	&relationships.attribute.dimension_type_name=eq.b"
```

But there's an extra level of nesting that I don't want, and a bunch of null attributes:

```json
[{
	"id": 32,
	"dimension_type_name": "a",
	"text": "tmgxledhjxjlgqh",
	"relationships": [
		{
			"attribute": {
				"id": 76,
				"dimension_type_name": "b",
				"text": "didpwfqkgxyfpkr"
			}
		},
		{
			"attribute": null
		},
		{
			"attribute": null
		}
	]
},
// ...
]
```

Is something akin to this supported?

```bash
"/dimension?
	select=*,
	attributes:dimension_relationship.ancestor_id.dimension.descendent_id(*)
	&dimension_type_name=eq.a
	&attributes.dimension_type_name=eq.b"
```

[...]

PR #1625, opened 2020-10-15
Enable embedding through multiple layers of views recursively

### LINKED ISSUE #1607: Embedding views recursively
Currently, a foreign key relationship is not detected through more than one layer of a view, although that should not be much of a problem theoretically. This can be useful when creating views as an abstraction between the data tables and the public apis views.

So assuming the example schema from the [docs](http://postgrest.org/en/latest/api.html#resource-embedding) and something like this:

```sql
CREATE VIEW german_films AS
SELECT * FROM films WHERE language='de';

CREATE VIEW best_german_films AS
SELECT * FROM german_films WHERE rating>8;
```

It would be possible to the following:
- `GET /films?select=title,directors(*)`
- `GET /german_films?select=title,directors(*)`

But not:
- `GET /best_german_films?select=title,directors(*)`

In the last case, the relationship is not detected. This is because the query in `allSourceColumns` is not recursive:
https://github.com/PostgREST/postgrest/blob/d395bb60522aa228a534d9181339519f171fe6b2/src/PostgREST/DbStructure.hs#L680-L736

It stops once it has found that the `best_german_films.director_id` column is a direct mapping of `german_films.director_id`. But it should be possible to extend this query to recurse, once the "base" table is actually another view.

wolfgangwalther:
Resolves #1607.

Changes the `allSourceColumns` query to recurse, when the base relation is still a view or materialized view. The query now returns the "true" base columns through multiple view layers, enabling the embedding of views of views of... you know!

### DIFF
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -175,6 +175,14 @@ create view orders_view as
 create view consumers_view as
   select * from public.public_consumers;
 
+create view consumers_view_view as
+  select * from consumers_view;
+
+create view private.consumers_private as
+  select * from consumers_view;
+
+create view consumers_private_view as
+  select * from private.consumers_private;
 
 --
 -- Name: getitemrange(bigint, bigint); Type: FUNCTION; Schema: test; Owner: -
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -178,11 +178,11 @@ create view consumers_view as
 create view consumers_view_view as
   select * from consumers_view;
 
-create view private.consumers_private as
+create view public.consumers_extra as
   select * from consumers_view;
 
-create view consumers_private_view as
-  select * from private.consumers_private;
+create view consumers_extra_view as
+  select * from public.consumers_extra;
 
 --
 -- Name: getitemrange(bigint, bigint); Type: FUNCTION; Schema: test; Owner: -
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1470, Allow calling RPC with variadic argument by passing repeated params - @wolfgangwalther
  - #1559, No downtime when reloading the schema cache with SIGUSR1 - @steve-chavez
  - #504, Add `log-level` config option. The admitted levels are: crit, error, warn and info - @steve-chavez
+ - #1607, Enable embedding through multiple views recursively - @wolfgangwalther
 
 ### Fixed
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -244,7 +244,7 @@ connectionStatus pool =
 fillSchemaCache :: P.Pool -> PgVersion -> IORef AppConfig -> IORef (Maybe DbStructure) -> IO ()
 fillSchemaCache pool actualPgVersion refConf refDbStructure = do
   conf <- readIORef refConf
-  result <- P.use pool $ HT.transaction HT.ReadCommitted HT.Read $ getDbStructure (toList $ configSchemas conf) actualPgVersion
+  result <- P.use pool $ HT.transaction HT.ReadCommitted HT.Read $ getDbStructure (toList $ configSchemas conf) (configExtraSearchPath conf) actualPgVersion
   case result of
     Left e -> do
       -- If this error happens it would mean the connection is down again. Improbable because connectionStatus ensured the connection.
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -31,6 +31,7 @@ import qualified Hasql.Session       as H
 import qualified Hasql.Statement     as H
 import qualified Hasql.Transaction   as HT
 
+import Contravariant.Extras          (contrazip2)
 import Data.Set                      as S (fromList)
 import Data.Text                     (breakOn, dropAround, split,
                                       splitOn, strip)
@@ -43,12 +44,12 @@ import Text.InterpolatedString.Perl6 (q, qc)
 import PostgREST.Private.Common
 import PostgREST.Types
 
-getDbStructure :: [Schema] -> PgVersion -> HT.Transaction DbStructure
-getDbStructure schemas pgVer = do
+getDbStructure :: [Schema] -> [Schema] -> PgVersion -> HT.Transaction DbStructure
+getDbStructure schemas extraSearchPath pgVer = do
   HT.sql "set local schema ''" -- This voids the search path. The following queries need this for getting the fully qualified name(schema.name) of every db object
   tabs    <- HT.statement () allTables
   cols    <- HT.statement schemas $ allColumns tabs
-  srcCols <- HT.statement schemas $ allSourceColumns cols pgVer
+  srcCols <- HT.statement (schemas, extraSearchPath) $ allSourceColumns cols pgVer
   m2oRels <- HT.statement () $ allM2ORels tabs cols
   keys    <- HT.statement () $ allPrimaryKeys tabs
   procs   <- HT.statement schemas allProcs
@@ -672,9 +673,9 @@ pkFromRow :: [Table] -> (Schema, Text, Text) -> Maybe PrimaryKey
 pkFromRow tabs (s, t, n) = PrimaryKey <$> table <*> pure n
   where table = find (\tbl -> tableSchema tbl == s && tableName tbl == t) tabs
 
-allSourceColumns :: [Column] -> PgVersion -> H.Statement [Schema] [SourceColumn]
+allSourceColumns :: [Column] -> PgVersion -> H.Statement ([Schema], [Schema]) [SourceColumn]
 allSourceColumns cols pgVer =
-  H.Statement sql (arrayParam HE.text) (decodeSourceColumns cols) True
+  H.Statement sql (contrazip2 (arrayParam HE.text) (arrayParam HE.text)) (decodeSourceColumns cols) True
   -- query explanation at https://gist.github.com/steve-chavez/7ee0e6590cddafb532e5f00c46275569
   where
     subselectRegex :: Text
@@ -707,7 +708,7 @@ allSourceColumns cols pgVer =
         from pg_class c
         join pg_namespace n on n.oid = c.relnamespace
         join pg_rewrite r on r.ev_class = c.oid
-        where c.relkind in ('v', 'm') and n.nspname not in ('pg_catalog', 'information_schema')
+        where c.relkind in ('v', 'm') and n.nspname = ANY($1 || $2)
       ),
       removed_subselects as(
         select
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -686,13 +686,15 @@ allSourceColumns cols pgVer =
                    | otherwise = ":subselect {.*?:stmt_len 0} :location \\d+} :res(no|ult)"
     sql = [qc|
       with recursive
-      any_part_of_fk as (
+      pks_fks as (
+        -- pk + fk referencing col
         select
           conrelid as resorigtbl,
           unnest(conkey) as resorigcol
         from pg_constraint
-        where contype='f'
+        where contype IN ('p', 'f')
         union
+        -- fk referenced col
         select
           confrelid,
           unnest(confkey)
@@ -769,7 +771,7 @@ allSourceColumns cols pgVer =
       join pg_attribute col on col.attrelid = tbl.oid and col.attnum = rec.resorigcol
       join pg_attribute vcol on vcol.attrelid = rec.view_id and vcol.attnum = rec.view_column
       join pg_namespace sch on sch.oid = tbl.relnamespace
-      join any_part_of_fk using (resorigtbl, resorigcol)
+      join pks_fks using (resorigtbl, resorigcol)
       order by view_schema, view_name, view_column_name; |]
 
 getPgVersion :: H.Session PgVersion
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -684,62 +684,92 @@ allSourceColumns cols pgVer =
     subselectRegex | pgVer < pgVersion100 = ":subselect {.*?:constraintDeps <>} :location \\d+} :res(no|ult)"
                    | otherwise = ":subselect {.*?:stmt_len 0} :location \\d+} :res(no|ult)"
     sql = [qc|
-      with
+      with recursive
+      any_part_of_fk as (
+        select
+          conrelid as resorigtbl,
+          unnest(conkey) as resorigcol
+        from pg_constraint
+        where contype='f'
+        union
+        select
+          confrelid,
+          unnest(confkey)
+        from pg_constraint
+        where contype='f'
+      ),
       views as (
         select
+          c.oid       as view_id,
           n.nspname   as view_schema,
           c.relname   as view_name,
           r.ev_action as view_definition
         from pg_class c
         join pg_namespace n on n.oid = c.relnamespace
         join pg_rewrite r on r.ev_class = c.oid
-        where c.relkind in ('v', 'm') and n.nspname = ANY ($1)
+        where c.relkind in ('v', 'm') and n.nspname not in ('pg_catalog', 'information_schema')
       ),
       removed_subselects as(
         select
-          view_schema, view_name,
+          view_id, view_schema, view_name,
           regexp_replace(view_definition, '{subselectRegex}', '', 'g') as x
         from views
       ),
       target_lists as(
         select
-          view_schema, view_name,
-          regexp_split_to_array(x, 'targetList') as x
+          view_id, view_schema, view_name,
+          string_to_array(x, 'targetList') as x
         from removed_subselects
       ),
       last_target_list_wo_tail as(
         select
-          view_schema, view_name,
-          (regexp_split_to_array(x[array_upper(x, 1)], ':onConflict'))[1] as x
+          view_id, view_schema, view_name,
+          (string_to_array(x[array_upper(x, 1)], ':onConflict'))[1] as x
         from target_lists
       ),
       target_entries as(
         select
-          view_schema, view_name,
-          unnest(regexp_split_to_array(x, 'TARGETENTRY')) as entry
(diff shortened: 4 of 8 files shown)


steve-chavez:
Interesting! :eyes:

One thing we need to check with changes to DbStructure is if they considerably slow down our startup time.

We currently test this manually(me and Remo) with a sample big schema we share privately.
Would you like to test this for your self? If so, you can shoot me an email at stevechavezast@gmail.com and I'll share you the instructions.

wolfgangwalther:
I used the "big schema" and did some performance testing. Since I only changed the query, I ran it directly on the database. I used the following function to run 100 times - alternating both queries - and compute the average plan and execution time for each query.

```sql
create or replace function timeit (n int, variadic queries text[], out plan_avg numeric[], out execution_avg numeric[])
language plpgsql as $$
declare
  i int;
  j int;
  explained json;
  plan_sum numeric[];
  execution_sum numeric[];
begin
  for i in 1..n loop
    for j in array_lower(queries, 1)..array_upper(queries, 1) loop
      execute 'explain (analyze, format json) ' || queries[j] into explained;
      plan_sum[j] := coalesce(plan_sum[j], 0) + (explained->0->>'Planning Time')::numeric;
      plan_avg[j] := plan_sum[j] / i;
      execution_sum[j] := coalesce(execution_sum[j], 0) + (explained->0->>'Execution Time')::numeric;
      execution_avg[j] := execution_sum[j] / i;
    end loop;
  end loop;
end
$$;

select * from timeit(100,
$q$ ... old query ... $q$,
$q$ ... new query ... $q$);
```

The results are as follows:
```
query | plan_avg | execution_avg
old   | 0.56 ms  | 395.3 ms
new   | 0.63 ms  | 399.7 ms
```

I repeated that and had the same difference (~+5ms for the new query).

So, as expected the recursion is a bit slower. But really not much.

As a side note:
- The old query returns 2208 rows
- The new query returns 2181 rows
- The new query without `where base_kind not in ('v', 'm')` returns 2490 rows

The new query returns fewer rows, because rows that have a view as a base table are removed - those are useless anyway. However, without removing those rows, the new query returns almost 300 rows more, so there are plenty of "views of views" in that schema.

wolfgangwalther:
> *well not really. not nearly all of those columns are unique or referenced at all in a fk constraint...

Thinking this further... we don't need to return **all** source columns from this query, but only those that are used in any foreign key constraint on either side. Correct?

I changed both the old and the new query to return only those columns:
- The old query now returns 543 rows
- The new query returns 519 rows (the same logic as above applies here)
- The except query returns a difference of 24 columns

Those 24 columns now really represent new opportunities to embed.

And even better - those queries perform better than ever:

```
query        | plan_avg | execution_avg
master       | 0.54ms   | 387.7ms
recursive    | 0.62ms   | 393.0ms
master_fk    | 1.16ms   | 362.2ms
recursive_fk | 1.06ms   | 362.0ms
```

Now, the recursive query is just as fast as the one without recursion - but both are about 7% faster than before.

wolfgangwalther:
Reordered the whole query a bit and replaced `regexp_split_to_array` with `string_to_array`, which is a lot faster. The query now comes it at around 220-230ms, so almost twice as fast.

Should be good to merge.

steve-chavez:
Mindblown! :boom:  :exploding_head:

Amazing work here Wolfgang. I wonder what @monacoremo thinks :smile:.

We might just call this new version **The Performance Release** .

I'm going to give it a review and then merge. It's all looking good though, I'm just going to digest and marvel at how the new queries work :)

monacoremo:
Great stuff!!! Like where we are going with this :-)

wolfgangwalther:
Rebased on current master and did some small renaming for less of a diff and better readability.

Glad you guys like it. I'm not sure whether I'm digging my own grave here with the performance stuff, because now it will be very hard for me to get my "transform to json" approach on the same level... ;)

wolfgangwalther:
Rebased on master.

Added another test as well, which is currently still failing. This is about a chain of views, where one of the views in the middle is not in the public schema. Will have to see how I can fix this the best, while keeping performance.

wolfgangwalther:
Embedding is now possible through hidden views in the middle of the chain as well. This does have the same performance on the big schema, but that's mainly because there are no hidden views in there. The query, as it is now, has to parse all views in the database to detect relationships through hidden views. In a setup with many hidden, but unrelated views this could increase the time for this query.

steve-chavez:
@wolfgangwalther I've finally understood your great work here!

The query is looking solid, but I think we can clarify the code a bit(for example `allSourceColumns` is no more, it should be something like `allSourceFkColumns`).

I'll do some reviews, but first..

> Embedding is now possible through hidden views in the middle of the chain as well.

Would this cover a realistic use case? 

IIUC, It would imply that a user creates **public** views for their API and then creates **private** views based on those **public** views.

That wouldn't make sense according to [schema isolation](http://postgrest.org/en/v7.0.0/schema_structure.html). The idea is to work inside out,  selecting **private** tables or views to be exposed as **public**.

> The query, as it is now, has to parse all views in the database to detect relationships through hidden views. In a setup with many hidden, but unrelated views this could increase the time for this query.

The price to pay seems a bit too high. PostgREST can be used on legacy databases and those can have a potentially large number of views.

wolfgangwalther:
> The query is looking solid, but I think we can clarify the code a bit(for example `allSourceColumns` is no more, it should be something like `allSourceFkColumns`).

I'll change that together with everything else you find in your review :)

> Would this cover a realistic use case?
> 
> IIUC, It would imply that a user creates **public** views for their API and then creates **private** views based on those **public** views.
> 
> That wouldn't make sense according to [schema isolation](http://postgrest.org/en/v7.0.0/schema_structure.html). The idea is to work inside out, selecting **private** tables or views to be exposed as **public**.

I did this, because this was exactly what lead me to this issue. It is the other way around, though. My scenario:
- hidden schema `data` with tables
- hidden schema `hidden` with some views as an abstraction on the data tables (some code that is shared between some of my public views)
- public schema `public` with view that use those in `hidden`

So this is exactly the inside-out approach you're describing.

Before my latest change the view detection would only take those views in the `public` schema, but not those in the `hidden` schema. The table in `data` would never be found.

> The price to pay seems a bit too high. PostgREST can be used on legacy databases and those can have a potentially large number of views.

I think, in fact, it probably only makes sense to do it like this. I would assume that it is more likely to have such a "view-chain" when using something like I did above (using views as another layer between data tables and public views).

But maybe we can do something inbetween: Right now we were either looking for views only in the schemas mentioned in `db-schema` or in all schemas everywhere. We could just look for views that are in the search path, so `db-schema` + `db-extra-search-path`. In this case I could add `hidden` to the search path in my case (which is something that I already did anyway). Users can still control how many views are exposed like this, then.

steve-chavez:
> hidden schema data with tables
> hidden schema hidden 
> public schema public with view that use those in hidden
> So this is exactly the inside-out approach you're describing.

Just got It! I was a bit mislead by the tests. That definitely coincides with the schema isolation approach.

> We could just look for views that are in the search path, so db-schema + db-extra-search-path

That sounds right! I agree with that change.

wolfgangwalther:
Wolfgang:
> Thinking this further... we don't need to return **all** source columns from this query, but only those that are used in any foreign key constraint on either side. Correct?

Steve:
> (for example `allSourceColumns` is no more, it should be something like `allSourceFkColumns`).

Looking at the code, my earlier assumption is wrong. It is not only the FK columns that we need, but also the PK columns of the base tables, because we have this line:
https://github.com/PostgREST/postgrest/blob/4b4a622a1702f29c394f12ad25349956abade878/src/PostgREST/DbStructure.hs#L58

We're missing a test-case here, because my change should have made at least one of them fail. I guess all relevant PKs in our test-schema are also used in some kind of foreign key relationship. I will try to make up a test-case to make sure we're not breaking anything.

wolfgangwalther:
We just don't have any case to test inserting into a view and getting a location header back. At least I couldn't find any. I'm pretty sure this works already, so we need to add that.

steve-chavez:
> it is not only the FK columns that we need, but also the PK columns of the base tables
> We just don't have any case to test inserting into a view and getting a location header back

@wolfgangwalther You're right, I also missed that one. A new test would be great.

wolfgangwalther:
Rebased on #1644, which we should merge first. This shows that the new test is indeed breaking with the current changes here - will fix.

wolfgangwalther:
Rebased on master, fixed the missing pks and did the todos from above.

steve-chavez:
Sorry for the late response here @wolfgangwalther. I wanted to make sure we get a nightly for your PR, so I did https://github.com/PostgREST/postgrest/pull/1648 first.

I'll review your latest changes tomorrow :)

test/fixtures/schema.sql
```diff
@@ -175,6 +175,14 @@ create view orders_view as
...
   select * from public.public_consumers;
 
+create view consumers_view_view as
+  select * from consumers_view;
+
+create view public.consumers_extra as
+  select * from consumers_view;
+
+create view consumers_extra_view as
+  select * from public.consumers_extra;
```
steve-chavez, line 185: @wolfgangwalther Hmm.. still a bit confused about this test setup, perhaps you can clarify. This is what I'm seeing:

- `public.consumers_extra` - the middle(or hidden) view.
- `test.consumers_extra_view` - the exposed view which is based on the middle view.

So far so good. The problem is that the middle view `public.consumers_extra` is based on `test.consumers_view`(filling the explicit schema here) which is is already on a exposed schema.

That doesn't look like the inside-out approach we discussed before. It's more like an in-out-in-out, if that makes sense.

This is a test setup that makes sense to me:

```sql
create view public.consumers_middle as
select * from public.public_consumers;
                                          
create view test.consumers_middle_view as
select * from public.consumers_middle;
```

And the test:

```http
GET /consumers_hidden_view?select=*,orders_view(*)
```

I've tested it and it works good. What do you think? Is the test clearer that way?

steve-chavez: I've also checked that the above test setup didn't work on your first recursive-only approach. That commit got lost because of the force-push, but the git reflog helped me there :). So the middle view query is definitely good, only want to clarify the test before merging :rocket: .

wolfgangwalther: > I've tested it and it works good. What do you think? Is the test clearer that way?

We currently have the following two view chains exposed on the public schema [view name: chain of schema types]:

- consumers_view_view: exposed -> exposed > hidden
- consumers_extra_view: exposed -> extra -> exposed -> hidden

As you noted only the "exposed -> extra" part is relevant for this test-case.

The only reason why I based the extra view on another public view was, because this allows us to implicitly test another level of recursion. So the 2nd chain is now longer compared to the first one. I didn't feel like adding another test-case just for this, so I took the short way out here.

wolfgangwalther: > That commit got lost because of the force-push, but the git reflog helped me there :).

I might have to force-push less... but first I have to see what that reflog is ;)

steve-chavez: > The only reason why I based the extra view on another public view was, because this allows us to implicitly test another level of recursion

Ahh, I see. It's more of an artificial test. Then it's all good! I'll merge :rocket:

steve-chavez: > I might have to force-push less... but first I have to see what that reflog is ;)

I'm guilty of force-pushing as well :grin: . So no need to change any workflows, reflog is good enough :)

steve-chavez:
Just FYI. I've released a [nightly](https://github.com/PostgREST/postgrest/releases/tag/nightly) with this change.

Also, we definitely need to document the `db-extra-search-path` interaction with hidden/middle views. I think a subsection under [Embedding Views](http://postgrest.org/en/v7.0.0/api.html#embedding-views) would do.

wolfgangwalther:
Sweet. Just started testing the nightly in my current project. The first thing that happenend was that some embeddings, that I had in place before, suddenly returned 30x, because there were 5 additional m2m relationships found to embed on.... :D
So this is definitely something that we need to point out for the next release as well, just as discussed in #1593.

steve-chavez:
> suddenly returned 30x, because there were 5 additional m2m relationships

Oh, that's bad. And #1593 is likely to bring more possibilities.

The m2m's is something I was thinking sometime ago. 
Right now, many tables can be treated as [junctions](https://en.wikipedia.org/wiki/Associative_entity) even when they're not really meant to be.

Perhaps we can select junctions for embedding in a stricter way? Like when they only have FKs and no additional columns?

We could even test this with a config option like `db-strict-junctions`.

wolfgangwalther:
> Oh, that's bad. And #1593 is likely to bring more possibilities.

My reaction was more like: "YES! It worked :D". Wasn't that bad, as I have automated tests in place for all embeddings I'm using and it was just a simple hint to add.

> The m2m's is something I was thinking sometime ago.
> Right now, many tables can be treated as [junctions](https://en.wikipedia.org/wiki/Associative_entity) even when they're not really meant to be.
> 
> Perhaps we can select junctions for embedding in a stricter way? Like when they only have FKs and no additional columns?
> 
> We could even test this with a config option like `db-strict-junctions`.

I thought about this a bit. I agree that something needs to be done about m2m relationships and junction tables. Not sure whether I'd go in the same direction, though.

In my scenarios so far, I almost always had some other columns on those junction tables. I can't use m2m embedding for those. I do want a "flat" output and not something nested, so I ended up creating views that join the junction and target table. This view is then easily (now!) embeddable.

However, if there was a way to do this "flat m2m embedding" in the query syntax, that would be even better.

So this boils down to: In the current implementation m2m embedding only makes sense with "strict" junction tables anyway - so the limitation would sure help to decrease ambiguation. But maybe it would be possible to extend the m2m feature to make it more usable with extended junction tables instead?

steve-chavez:
> So this boils down to: In the current implementation m2m embedding only makes sense with "strict" junction tables anyway 

Yes, I also believe that, the non-fk columns are lost anyway.

> But maybe it would be possible to extend the m2m feature to make it more usable with extended junction tables instead?

Maybe we could use the `cardinality` hint I mentioned on https://github.com/PostgREST/postgrest/issues/1643#issuecomment-721417728.

So, by default only strict junctions are automatically detected, but extended junctions can be used for M2M embedding by adding an `m2m` hint.

wolfgangwalther:
> > But maybe it would be possible to extend the m2m feature to make it more usable with extended junction tables instead?
> 
> Maybe we could use the `cardinality` hint I mentioned on [#1643 (comment)](https://github.com/PostgREST/postgrest/issues/1643#issuecomment-721417728).
> 
> So, by default only strict junctions are automatically detected, but extended junctions can be used for M2M embedding by adding an `m2m` hint.

I like that. It would reduce complexity for the most common cases, but still allow all of those we currently do.

Do we need to support adding multiple hints in this case? I think yes, because otherwise we can't tell which m2m to use.

To avoid name collisions with columns or fks named "m2m" / "o2m" / "m2o", maybe we should use another operator for the cardinality hint and separate the two types of hints? Could be easier to parse as well?

> Yes, I also believe that, the non-fk columns are lost anyway.

Do you think there'd be any way we can change this? I think this has been requested in the past as well. Maybe we can just allow selecting columns from the junction table inside a m2m relationship to flatten the output instead of having two nested embeds. Flattening could also be useful for m2o relationships.

Flattening would have to work differently for m2o and m2m relationships:

- in a m2o relationship, we'd want to move the embedded columns to the "outer" level. I suggest something like this:
`GET /films?select=title,director:directors.last_name,directors()`
returns
```json
[
  { "title": "Workers Leaving The Lumière Factory In Lyon",
    "director": "Lumière"
  },
  { "title": "The Dickson Experimental Sound Film",
    "director": "Dickson"
  },
  { "title": "The Haunted Castle",
    "director": "Méliès"
  }
]
```

- in the m2m relationship, we could then use `..` instead of `directors.` to qualify the column, so something like this:
`GET /actors?select=first_name,last_name,films!m2m.roles(title,..character)`
`character` is then taken from the junction table.

steve-chavez:
Regarding flattening, I came up with a proposal here: https://github.com/PostgREST/postgrest/issues/1233#issuecomment-469782724. Would that be enough for the use cases you have in mind?

> Do we need to support adding multiple hints in this case?

Yes, that's also something I considered on the first disambiguation proposal. It was like:

```http
GET /<table>?select=*,<table>!<cardinality>!<fk>(*)
```

> To avoid name collisions with columns or fks named "m2m" / "o2m" / "m2o", maybe we should use another operator for the cardinality hint and separate the two types of hints? 

I thought those names are rare enough to not be used as columns/fks. But we can also warn about this on docs.

Not sure about more operators. I used the `!` because we are short in urlsafe chars to use as operators(it also reminded me of the `!important` in css). IIRC, the only other char that was available was `$`, but pg identifiers can have `$` so I discarded it.

wolfgangwalther:
> Regarding flattening, I came up with a proposal here: [#1233 (comment)](https://github.com/PostgREST/postgrest/issues/1233#issuecomment-469782724). Would that be enough for the use cases you have in mind?

If I read that correctly, that will only work for the embedding of m2o relationships. The example you gave there was a m2o->m2o relationship, right?

Regarding m2o relationships our proposals differ in two aspects:

- use of `->` vs `.`. That's just a matter of choice, I guess. I feel like `.` more closely resembles the SQL ("table.column"), but `->` would work as well of course.

- Your suggestion has the embedding and the flattening together like this: `clients(*)->name`, where I have two fields: `clients.name,clients()`. The empty embedding would be used to do the join, while the `clients.name` / `alias.column` would put one of the columns in the result. This could easily be extended to support multiple columns.

I have not looked in detail at the SQL, yet, but I think it should be possible to do this with multiple columns as well.

---

m2m relationships are different. I don't see how your proposal would work in this case. Taking the same example, but a slightly different request:

`GET /tasks?select=name,clients!projects(*)`

This would give me a list of clients through the m2m relationship. How would I go about adding the columns of the projects table / junction now?

> Yes, that's also something I considered on the first disambiguation proposal. It was like:
> 
> ```httpspec
> GET /<table>?select=*,<table>!<cardinality>!<fk>(*)
> ```

> I thought those names are rare enough to not be used as columns/fks. But we can also warn about this on docs.
> 
> Not sure about more operators. I used the `!` because we are short in urlsafe chars to use as operators(it also reminded me of the `!important` in css). IIRC, the only other char that was available was `$`, but pg identifiers can have `$` so I discarded it.

You're right. Using multiple `!` and making m2o, o2m and m2m keywords should be fine and straightforward.

PR #1627, opened 2020-10-16
remove io tests from stack-test because of random segmentation faults

wolfgangwalther:
this is just to debug - nothing to merge, yet.

Right now, `stack-test` often fails with a segmentation fault during the io-tests. I tried to reproduce locally, but couldn't. Here, I ripped out all the other CI pipelines (no need to run them loads of times) and removed some `> /dev/null` redirections. Hopefully I can reproduce the error when running the CI pipeline a couple of times and get more useful logging output.

Let's see what happens.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,12 +71,6 @@ jobs:
           command: |
             POSTGREST_TEST_CONNECTION=$(test/create_test_db "postgres://circleci@localhost" postgrest_test) \
               stack test
-      - run:
-          name: run io tests
-          command: |
-            export PATH="~/.local/bin:$PATH"
-            stack install --fast -j1
-            test/io-tests.sh
 
   # Run memory usage tests based on stack and docker.
   stack-test-memory:


wolfgangwalther:
Triggered the build 10 times, but the io failure didn't show up. Funny thing is... at the same time it happened in a different PR, but of course the extended logging was not enabled there.

I suggest to temporarily merge this into master and wait until the failure shows up again somewhere. Hopefully the bit of extra logging will then give us an idea how to proceed. The io tests will not look nice during that time, but that should be ok.

wolfgangwalther:
Alternatively, we could just remove the io test suite from `stack-test`. The io tests are run in `nix-build-test` anyway and have never shown that same error there so far. So it seems to be very much an issue with the ci setup, not the tests or even postgrest.

steve-chavez:
> Alternatively, we could just remove the io test suite from stack-test. The io tests are run in nix-build-test anyway and have never shown that same error there so far

I've thought of that as well. I'd agree with that change.

wolfgangwalther:
so be it! changed!

PR #1628, opened 2020-10-16
ci: run spec tests against all pg versions, even if one of them fails

wolfgangwalther:
Sometimes the spec tests fail only for older pg versions. It would be cool, if the spec tests would still be run against all versions, even if one of them fails, to get a quick overview of the problem with different versions.

This change tries to do that. I will temporarily push a couple of errors after the first build to confirm it is working as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,29 +184,41 @@ jobs:
           name: Install testing scripts
           command: nix-env -f default.nix -iA tests tests.ioTests tests.memoryTests
       - run:
+          name: Skip tests on build failure
+          command: circleci-agent step halt
+          when: on_fail
+      - run:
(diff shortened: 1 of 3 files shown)


wolfgangwalther:
This is only about the nix-build-test step and it works well:
- when the build fails no tests are run
- when the build succeeds all tests are run, even if some of them fail

I left the commits in here, so you can still see the pipeline results easily. Please squash this when merging.

Ready to merge.

steve-chavez:
> Sometimes the spec tests fail only for older pg versions. It would be cool, if the spec tests would still be run against all versions, even if one of them fails, to get a quick overview of the problem with different versions.

Yeah! This is the way it used to work before as well with different circle jobs for each pg test.

Nice improvement :)

Merging.. :rocket:

PR #1630, opened 2020-10-16
ci: add coveralls.io integration

wolfgangwalther:
Wouldn't it be good to have code coverage tools in CI as well?

I have found 3 packages that send hpc coverage results to one of the ci services, but only `stack-hpc-coveralls` seems to be compatible to stack. The other 2 use cabal.

I thought it would make sense to use `stack-test` for that purpose. I'm not sure why that was specifically targeting the oldest supported PG version and changed that to the latest. Coverage results seem to be more interesting against the latest version, as that might touch more features, which are not enabled for older versions!?

To make this work, @steve-chavez, you would have to register the postgrest repo with coveralls.io. I'm not sure whether we need to add a token somewhere. The coveralls docs suggest we have to, but the source code for `stack-hpc-coveralls` tells me that the repo_token command line option is only used in case the CI is detected as `travis-ci`.

If you want to do that, I suggest you register the repo and we see what happens.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,8 +47,7 @@ jobs:
             curl -L https://github.com/commercialhaskell/stack/releases/download/v2.3.1/stack-2.3.1-linux-x86_64.tar.gz | tar zx -C /tmp
             sudo mv /tmp/stack-2.3.1-linux-x86_64/stack /usr/bin
             sudo apt-get update
-            # ncat utility from nmap needed to test socket connection with curl <  7.40
-            sudo apt-get install -y libgmp-dev postgresql-client nmap
+            sudo apt-get install -y libgmp-dev postgresql-client
             sudo apt-get install -y --only-upgrade binutils
             stack setup
       - run:
--- a/nix/README.md
+++ b/nix/README.md
@@ -97,7 +97,7 @@ postgrest-test-spec-all             postgrest-test-spec-postgresql-9.6
 
 ```
 
-Note that `postgrest-tests-io` is now also available.
+Note that `postgrest-test-io` is now also available.
 
 To run one-off commands, you can also use `nix-shell --run <command>`, which
 will lauch the Nix shell, run that one command and exit. Note that the tab
@@ -168,7 +168,7 @@ give you some more background and details on how it works.
 
 ### `default.nix`
 
-[`default.nix`](../default.nix) is our 'respository expression' that pulls all
+[`default.nix`](../default.nix) is our 'repository expression' that pulls all
 the pieces that we define with Nix together. It returns a set (like a dict in
 other programming languages), where each attribute is a derivation that Nix
 knows how to build, like the `postgrest` attribute from earlier.
--- a/test/io-tests.sh
+++ b/test/io-tests.sh
@@ -43,7 +43,7 @@ pgrStart(){ postgrest $1 >/dev/null 2>/dev/null & pgrPID="$!"; }
 pgrStartRead(){ postgrest $1 <$2 >/dev/null & pgrPID="$!"; }
 pgrStartStdin(){ postgrest $1 >/dev/null <<< "$2" & pgrPID="$!"; }
 pgrStarted(){ kill -0 "$pgrPID" 2>/dev/null; }
-pgrStop(){ kill "$pgrPID" 2>/dev/null; pgrPID=""; }
+pgrStop(){ kill "$pgrPID" 2>/dev/null; pgrPID=""; sleep 0.1; }
 
 # Utilities to send HTTP requests to the PostgREST server
 rootStatus(){
@@ -273,13 +273,10 @@ replaceConfigValue(){
 }
 
 getSocketStatus() {
-  curl -sL -w "%{http_code}\\n" -o /dev/null localhost:54321
+  curl -sL -w "%{http_code}\\n" -o /dev/null --unix-socket /tmp/postgrest.sock http://localhost/
 }
 
 socketConnection(){
-  # map port 54321 traffic to unix socket as workaround for curl below 7.40
-  # not supporting --unix-socket flag
-  ncat -vlk 54321 -c 'ncat -U /tmp/postgrest.sock' &
   pgrStart "./configs/unix-socket.config"
   while pgrStarted && test "$( getSocketStatus )" -ne 200
   do
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,70 +1,113 @@
-## Travis is only used for building an OSX binary ,
-## no tests are run here.
 language: generic
 
 sudo: false
 
-os:
-- osx
+jobs:
+  include:
+    - name: Build OSX Binary
+      os: osx
+      cache:
+        timeout: 1000
+        directories:
+          - $HOME/.stack
+          - $HOME/.local/bin
+      before_install:
+        - mkdir -p "$HOME/.local/bin"
+        - export PATH="$PATH:$HOME/.local/bin"
+      install:
+        - |
+          if test -f "$HOME/.local/bin/stack"
+            then
+              echo 'Stack is already installed.'
+            else
+              echo "Installing Stack..."
+              travis_retry curl -L https://www.stackage.org/stack/osx-x86_64 > stack.tar.gz
+              gunzip stack.tar.gz
+              tar -x -f stack.tar --strip-components 1
+              mv stack "$HOME/.local/bin/"
+              rm stack.tar
+          fi
+        - |
+          if test -f "$HOME/.local/bin/ghr"
+          then
+            echo 'ghr is already installed.'
+          else
+            echo "Installing ghr..."
+            travis_retry curl -L https://github.com/tcnksm/ghr/releases/download/v0.5.4/ghr_v0.5.4_darwin_386.zip > ghr.zip
+            unzip ghr.zip -d "$HOME/.local/bin"
+            rm ghr.zip
+          fi
+      script:
+        ## Building the whole project can take longer than 50 minutes. Since Travis has a global timeout of 50 minutes
+        ## we compile for 30 minutes tops(`gtimeout 1800`) and quit compiling with no error.
+        ## Since we CACHE the compile results we can continue compiling from where we left off
+        ## on the next commit.
+        - gtimeout 1800 stack build --no-terminal --only-snapshot --install-ghc || (($?==124))
+        - |
+          if test ! "$TRAVIS_TAG"
+          then
+            echo 'No tag pushed. Skip building binary.'
+          else
+            stack build --no-terminal --copy-bins --local-bin-path .
+          fi
+        - |
+          if test ! "$TRAVIS_TAG"
+          then
+            echo 'No tag pushed. Skipping release.'
+          else
+            OWNER="$(echo "$TRAVIS_REPO_SLUG" | cut -f1 -d/)"
+            REPO="$(echo "$TRAVIS_REPO_SLUG" | cut -f2 -d/)"
+            START=$(echo $TRAVIS_TAG | cut -c2-)
+            END='## \['
+            BODY=$(sed -n "1,/$START/d;/$END/q;p" CHANGELOG.md)
+            strip postgrest
+            tar cJf postgrest-$TRAVIS_TAG-osx.tar.xz postgrest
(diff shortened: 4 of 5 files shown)


wolfgangwalther:
Currently failing with:

```
Error: Coveralls returned HTTP 422 Unprocessable Entity
{"message":"Couldn't find a repository matching this job.","error":true}
```

So that looks ok so far, since it's not registered, yet.

steve-chavez:
> Wouldn't it be good to have code coverage tools in CI as well?

@wolfgangwalther Yes! That looks interesting. I'll see about enabling coveralls.

steve-chavez:
I've added the repo to coveralls: https://coveralls.io/github/PostgREST/postgrest.

I'm re-running the checks to see if something happens.
However, I'm not sure ifthis is all we need. For showing coveralls on the checks, wouldn't we need a `.coveralls.yml` file?

steve-chavez:
> Error: Coveralls returned HTTP 422 Unprocessable Entity
{"message":"Couldn't find a repository matching this job.","error":true}

Same error. There must be something missing..

wolfgangwalther:
I do have coveralls set up for [vue-postgrest](https://github.com/technowledgy/vue-postgrest) without the `.coveralls.yml` file. But I remember, I had to set up that token.

I changed the setup to use some other tools: The haskell package `hpc-lcov` will convert from hpc to lcov format and then from there `node-coveralls` should take over. This does have support for the `COVERALLS_REPO_TOKEN` environment variable. You will have to set this in CircleCI to the token you can get from coveralls, citing their docs:

> The secret repo token for your repository, found at the bottom of your repository’s page on Coveralls.

I changed the base docker image for `stack-test` to `circleci/node` to support coveralls. Let's see whether everything still builds. Maybe we have to add some dependencies.

wolfgangwalther:
Ok, the new solution now throws the same error as before - once you create that environment variable `COVERALLS_REPO_TOKEN`, that should change though!

.circleci/config.yml
```diff
@@ -47,10 +47,11 @@ jobs:
             curl -L https://github.com/commercialhaskell/stack/releases/download/v2.3.1/stack-2.3.1-linux-x86_64.tar.gz | tar zx -C /tmp
             sudo mv /tmp/stack-2.3.1-linux-x86_64/stack /usr/bin
             sudo apt-get update
-            # ncat utility from nmap needed to test socket connection with curl <  7.40
-            sudo apt-get install -y libgmp-dev postgresql-client nmap
+            sudo apt-get install -y libgmp-dev postgresql-client
             sudo apt-get install -y --only-upgrade binutils
             stack setup
+            stack install hpc-lcov
```
wolfgangwalther, line 53: where and how exactly do I add this to any of the cabal or stack files instead of calling this manually here? would be important to be reflected in the cache key through that as well.

.circleci/config.yml
```diff
@@ -47,10 +47,11 @@ jobs:
             curl -L https://github.com/commercialhaskell/stack/releases/download/v2.3.1/stack-2.3.1-linux-x86_64.tar.gz | tar zx -C /tmp
             sudo mv /tmp/stack-2.3.1-linux-x86_64/stack /usr/bin
             sudo apt-get update
-            # ncat utility from nmap needed to test socket connection with curl <  7.40
```
wolfgangwalther, line 50: that was left over from the io tests

steve-chavez:
> once you create that environment variable COVERALLS_REPO_TOKEN, that should change though!

Done!

wolfgangwalther:
Can you trigger a rebuild? I don't have permission to do that :)

wolfgangwalther:
The build currently takes ages, but that is due to caching and the changed base image I think. Therefore it has to build all deps again everytime.

.circleci/config.yml
steve-chavez: Hmm.. maybe it could go in `stack.yaml` -> `extra-deps`.

Then below use like `stack exec hpc-lcov && coveralls < lcov.info`?

(Not sure if it'd work)

wolfgangwalther: That sounds like it makes sense. But how do I add it there? there are some hashes in there - I am not supposed to put those in by hand, am I? I tried `stack install` locally, but that didn't add anything. Any stack command that adds those as extra-deps?

steve-chavez: IIRC, the hashes are only needed when stack doesn't have the package on its repos.

Looks like hpc-lcov is there though: https://www.stackage.org/lts-15.8/hoogle?q=hpc-lcov

You could try it like:
```
extra-deps:
- hpc-lcov-1.0.1
```

Actually, I've just noticed the hash is here: https://www.stackage.org/lts-16.19/package/hpc-lcov-1.0.1

```
This version can be pinned in stack with:hpc-lcov-1.0.1@sha256:f7831e13530634f0c2f4f2e613ddc15541f78ca45e4214e73f23efcfb90f68cb,2845
```

wolfgangwalther:
I see you have a SSH session into the last build. Can you confirm the env var is correctly set? Maybe something like `coveralls -v < lcov.info` can tell you anything.

steve-chavez:
> I see you have a SSH session into the last build.

Sorry, somehow circle was rejecting my ssh key. But I can confirm I have the env var set up on their UI:

2020-10-20-084428_963x70_scrot

(If the error keeps happening I'll retry connecting)

wolfgangwalther:
So I had "export" in there to see all env vars and `COVERALLS_REPO_TOKEN` was not there. It should appear masked, if I understand https://circleci.com/docs/2.0/env-vars/#secrets-masking correctly. I added another `echo $COVERALLS_REPO_TOKEN`.

The `stack exec hpc-lcov` did not work. I think stack did not install the package, even though it was added in stack.yaml. Not sure what's missing.

wolfgangwalther:
> I added another `echo $COVERALLS_REPO_TOKEN`.

That line is just empty. Also the result of coveralls -v looks like there is no repo_token in the json that will be sent. So somehow it's not there.

steve-chavez:
> That line is just empty. Also the result of coveralls -v looks like there is no repo_token in the json that will be sent. So somehow it's not there.

Just managed to SSH to the instance. You're right. None of the env vars show with `echo` though(GITHUB_TOKEN, CACHIX_SIGNING_KEY, etc).

wolfgangwalther:
Are those maybe tied to jobs that run on a specific branch or something?

steve-chavez:
Yes, looks like they're disabled for PRs. See below:

2020-10-20-120131_998x294_scrot

steve-chavez:
I'll enable passing secrets for this PR. I'll disable afterwards.

wolfgangwalther:
Maybe we'll have to set up the coverage job with a different ci provider then, where the coverage token would be the only secret?

When I remember correctly, travis doesn't even need a repo token, they do it differently somehow. So maybe that would be the best way to go forward.

steve-chavez:
> When I remember correctly, travis doesn't even need a repo token, they do it differently somehow. So maybe that would be the best way to go forward.

Yes, that would be better from a security standpoint. Circle has secrets for deploying to docker, cachix, github token for releases, etc.

wolfgangwalther:
Wuhuuu - it works. Travis builds and uploads coverage without token.

wolfgangwalther:
Now, looking at https://coveralls.io/builds/34347135, we might need to fix a couple of small things:
- why is anything in `.stack-work` showing up in that report... that seems wrong
- `main/...` is not showing up. do we need to run the io tests to get coverage here?

wolfgangwalther:
Now that we're running coverage on travis, I tested `stack-hpc-coveralls` again. Works as well.

Both tools produce slightly different results on the conversion from hpc to lcov/coveralls. My understanding is, that this is because hpc coverage is not line based, but lcov/coveralls is. The source-code view on coveralls.io doesn't really help that much, when compared to the html output we get from stack directly. I compared both conversions and the conversion from `stack-hpc-coveralls` seems to make a lot more sense. Just a random example:

```
// ApiRequest.hs #176
  | otherwise = Nothing
```

In the HPC output `otherwise` is marked green here (green = Always True). That makes sense, otherwise IS always true - by definition. So this is not a problem at all.
- `stack-hpc-coveralls` marks this line as green on coveralls (not missed)
- `hpc-lcov | coveralls` marks this line as red on coveralls (missed)

And `stack-hpc-coveralls` does not show stuff from `.stack-work`, so that seems better as well. The folder is still listed in coveralls, because of the first two reports from `hpc-lcov | coveralls` - maybe you can reset the coveralls reports somehow, @steve-chavez?

wolfgangwalther:
The file from `.stack-work` that's showing up is `module postgrest-7.0.1-JBRavZt2lgCYI0yAKBSaw/Paths_postgrest`. It's in the hpc report already, so both tools are sending it to coveralls.

It does not have a corresponding source file in our repo, but is built together with the core app - not any dependencies.

What is this?

wolfgangwalther:
Removed `Paths_postgrest` from postgrest.cabal. Still builds successfully locally. Not sure if the file still shows up in the coverage report, though.

wolfgangwalther:
Ok, I excluded `Paths_postgrest` manually from the reports and it works fine.

I'll have a look into bringing back the IO tests for more coverage of `main/...`.

wolfgangwalther:
Almost there, I guess. The coverage is now higher, so the results of the io-tests are merged with the those from `stack test`. But I still don't see any files from `main/...` showing up, yet... mh.

test/io-tests.sh
```diff
@@ -43,7 +43,7 @@ pgrStart(){ postgrest $1 >/dev/null 2>/dev/null & pgrPID="$!"; }
 pgrStartRead(){ postgrest $1 <$2 >/dev/null & pgrPID="$!"; }
 pgrStartStdin(){ postgrest $1 >/dev/null <<< "$2" & pgrPID="$!"; }
 pgrStarted(){ kill -0 "$pgrPID" 2>/dev/null; }
-pgrStop(){ kill "$pgrPID" 2>/dev/null; pgrPID=""; }
+pgrStop(){ kill "$pgrPID" 2>/dev/null; pgrPID=""; sleep 0.1; }
```
wolfgangwalther, line 46: This should fix the "segmentation fault" problem we had with stack-test as well. Not relevant anymore, because we removed the io-tests there, just noting.

Had a similar problem. Once coverage is enabled, the coverage report file from process A is not closed yet when process B already tries to write to it. This leads to the bash shell running the io-tests file erroring out with segmentation fault. In the case of stack-test I mostly observed it when the port process B tried to liste on was not closed yet from process A.

nix/README.md
```diff
@@ -168,7 +168,7 @@ give you some more background and details on how it works.
 
 ### `default.nix`
 
-[`default.nix`](../default.nix) is our 'respository expression' that pulls all
+[`default.nix`](../default.nix) is our 'repository expression' that pulls all
```
wolfgangwalther, line 171: Just two random typos I ran across and didn't feel like opening up another PR for.

test/io-tests.sh
```diff
@@ -273,13 +273,10 @@ replaceConfigValue(){
 }
 
 getSocketStatus() {
-  curl -sL -w "%{http_code}\\n" -o /dev/null localhost:54321
+  curl -sL -w "%{http_code}\\n" -o /dev/null --unix-socket /tmp/postgrest.sock http://localhost/
```
wolfgangwalther, line 276: this will now fail with curl < 7.40, which does not support --unix-socket.. but is 5 years old. Should be fine, I guess.

test/io-tests.sh
```diff
@@ -273,13 +273,10 @@ replaceConfigValue(){
...
 }
 
 socketConnection(){
-  # map port 54321 traffic to unix socket as workaround for curl below 7.40
-  # not supporting --unix-socket flag
-  ncat -vlk 54321 -c 'ncat -U /tmp/postgrest.sock' &
```
wolfgangwalther, line 282: removed this because travis blocks the nmap package.

wolfgangwalther:
Done!

steve-chavez:
> Paths_postgrest.hs.
> What is this?

Looks like that file contains cabal metadata, which we use to output the version number: https://stackoverflow.com/questions/2892586/how-can-my-haskell-program-or-library-find-its-version-number.

> Done!

@wolfgangwalther Awesome work here!

Merging now.. :rocket:

PR #1631, opened 2020-10-17
perf: change Text queries to ByteString

steve-chavez:
Improves performance by not utf8 encoding the full queries with `encodeUtf8`(done in `unicodeStatement`). It only encodes certain parts like pg identifiers and literals.

I get around ~220 req/s more on my load tests more with this change.

This is also a gradual step needed to use the Snippet type from [hasql-dynamic-statements](http://hackage.haskell.org/package/hasql-dynamic-statements-0.3.1/docs/Hasql-DynamicStatements-Snippet.html). We'll get prepared statements for GET when this library is integrated.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -187,7 +187,7 @@ app dbStructure proc cols conf apiRequest =
                         , Just $ contentRangeH 1 0 $ if shouldCount then Just queryTotal else Nothing
                         , if null pkCols && isNothing (iOnConflict apiRequest)
                             then Nothing
-                            else (\x -> ("Preference-Applied", encodeUtf8 (show x))) <$> iPreferResolution apiRequest
+                            else (\x -> ("Preference-Applied", BS.pack (show x))) <$> iPreferResolution apiRequest
                         ] ++ ctHeaders)) (unwrapGucHeader <$> ghdrs)
                   if contentType == CTSingularJSON && queryTotal /= 1
                     then do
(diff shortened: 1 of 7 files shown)


steve-chavez:
For being safe about escaping and quoting pg identifiers we use `pgFmtIdent` in a number of places. However, it'd be better to have a type like this:

```haskell
newtype PgIdent = PgIdent Text deriving (Show, Eq, Ord)
                                                                                                      
pgIdentToSql :: PgIdent -> SqlFragment
pgIdentToSql (PgIdent x) = encodeUtf8 $ "\"" <> T.replace "\"" "\"\"" (trimNullChars $ toS x) <> "\""
```

That way no identifier can slip by from being escaped.

Doing this change would involve changing the code in a lot of places and there are open PRs that would conflict. So I'll leave it for another time.

Should be good to merge.

wolfgangwalther:
> I get around ~220 req/s more on my load tests more with this change.

How much improvement is that? I.e. how many requests do you get in absolute numbers?

steve-chavez:
> How much improvement is that? I.e. how many requests do you get in absolute numbers?

@wolfgangwalther Sorry, that wasn't clear. Running this [test](https://github.com/steve-chavez/postgrest-benchmark/blob/master/k6/GETSingle.js#L30) on an AWS t3a.nano I got:

|   | req/s |
| ------------- | ------------- |
| Previous version  | 1526.06342633 |
| Bytesring version  | 1748.747079  |

Issue #1179, opened 2018-09-03
Feature request: Being able to manually specify foreign keys in views


photz:
### Description

I am still fairly new to working with PostgREST, but my understanding thus far is that when a database view is exposed through PostgREST, then PostgREST will try to detect any foreign key relationships to enable the kind of resource embedding described in the docs here:
http://postgrest.org/en/v5.0/api.html#resource-embedding

In my limited experience this works very well as long as the view is fairly straightforward, but when the view definition goes beyond a `select * from myTable`, then PostgREST will often fail to detect foreign key relationships.  

Since both resource embedding as well as the ability to work freely with views seem crucial to me, I've been wondering if it would be possible / make sense to let users manually specify which view columns are foreign keys.

steve-chavez:
Try v5.1(released a few days ago), we're now able to detect fks in more complicated views(see this [comment on a previous issue](https://github.com/PostgREST/postgrest/issues/709#issuecomment-405675425)), also the query got a lot faster(noticeable in big schemas).

If the new query detection is good enough(try/share a view that doesn't work) then it may be unnecessary to have this capability.

photz:
Thanks for your response, I'm currently using the subzerocloud/postgrest image which apparently already incorporates the changes you mention.  

At the moment resource embedding on the `state` column in the following view doesn't work:
```sql
create or replace view updates as
select walk.lead as lead,
       0 as pos,
       walk.start_state as state,  -- foreign key to data.states
       walk.created_at as entered_at
from data.walk
union all
(select step.lead as lead,
        step.idx + 1 as pos,
        edges.to_node as state, -- foreign key to data.states
        step.created_at as entered_at
from data.step
inner join data.edges on edges.id = step.edge)
order by 1, 2;
```

steve-chavez:
UNION counts as modifying the origin column as I mentioned [here](https://github.com/PostgREST/postgrest/issues/709#issuecomment-405675425) and even PostgreSQL(by querying pg_rewrite) cannot tell from which table/view the column(state in your case) was coming from.

I had some ideas for manually specifying the fk [here](https://github.com/PostgREST/postgrest/pull/1115#issuecomment-404032018), but I don't see any of those getting implemented anytime soon. 

This as a limitation for now(which will be documented) so I suggest trying to rewrite your view.

purpleKarrot:
I try `/stats?select=*,sites(*)` without success:

```sql
CREATE TABLE projects (
[...]

Issue #1323, opened 2019-06-10
POST-ing invalid json content results in 201 Created


james-callahan:
### Environment

* PostgreSQL version: docker: postgrest/postgrest:v5.2.0-3900baa
* PostgREST version: docker: postgrest/postgrest:v5.2.0-3900baa
* Operating system: Linux

### Description of issue

POST-ing a json string you can get a 201 success; even though it's not a valid row. e.g (strace output):
```
sendto(6, "POST /foo HTTP/1.1\r\nHost: localhost:3001\r\naccept: application/json\r\nAccept-Encoding: gzip, deflate\r\nUser-Agent: Python/3.7 aiohttp/4.0.0a0\r\nContent-Length: 5\r\nContent-Type: application/json\r\n\r\n", 193, 0, NULL, 0) = 193
sendto(6, "\"bar\"", 5, 0, NULL, 0)     = 5
recvfrom(6, "HTTP/1.1 201 Created\r\nTransfer-Encoding: chunked\r\nDate: Wed, 05 Jun 2019 04:42:52 GMT\r\nServer: postgrest/5.2.0 (UNKNOWN)\r\nContent-Range: */*\r\n\r\n0\r\n\r\n", 262144, 0, NULL, NULL) = 149
```

wolfgangwalther:
I can still confirm this in the current nightly.

Once the body is valid json, but not an object, 201 is returned but no row added.

steve-chavez:
Also can reproduce with an empty array:
```bash
http POST localhost:3000/projects <<JSON
[]
JSON

HTTP/1.1 201 Created
```

Adding an `Accept: application/vnd.pgrst.object` makes the request fail.

I think a fix can be done by checking the `queryTotal > 1` for 201. Similarly to how it's done for singlejson:

https://github.com/PostgREST/postgrest/blob/0c25f128250f427f470910104eea6b8bee589083/src/PostgREST/App.hs#L314-L319

wolfgangwalther:
Mh... this makes it a duplicate of #1070. Closing here in favor of the other one, because there's some other ideas to solve that more generally for on conflict update etc.

laurenceisla:
[...]

PR #1632, opened 2020-10-19
Fix embedding through views with subqueries inside function calls

### LINKED ISSUE #1608: Embedding views with sub-selects
When embedding views, the following breaks the detection algorithm:

```sql
-- this works
CREATE VIEW german_films AS
SELECT *,
       (SELECT 'hello') AS world
  FROM films
 WHERE language='de';

-- while this does not work
CREATE VIEW german_films AS
SELECT *,
       COALESCE((SELECT 'hello'), 'bye') AS world
  FROM films
 WHERE language='de';
```

What happens here is:
- the SQL query in `allSourceColumns` tries to parse the view definition in form of a `pg_node_tree` (https://www.postgresql.org/docs/current/catalog-pg-rewrite.html)
- to be able to read the embeddable columns, all sub-selects have to be filtered out
- to filter out the sub-selects a regex is used
- this regex fails when there is a function call around it (e.g. `COALESCE`)
- because the regex fails, no columns are detected at all for embedding

My current solution is to wrap another `SELECT` around the `COALESCE` - that works fine.

I have played a bit with changing up the query in `allSourceColumns` - I think the structure of `pg_node_tree` does not allow a proper regex solution to really get rid of all sub-selects. However, I was able to apply a couple of transformations to the node tree to bring it to JSON format - which can easily be parsed for the embeddable columns. This was only tested with PG12 so far and also not very extensively, yet. (just noted that it fails on some of the pg_catalog views apparently - so there is definitely some more tweaking needed for the `transform_json` part)

The query currently looks like this:

```sql


with
views as (
select
   n.nspname   as view_schema,
   c.relname   as view_name,
   r.ev_action as view_definition
from pg_class c
join pg_namespace n on n.oid = c.relnamespace
join pg_rewrite r on r.ev_class = c.oid
where c.relkind in ('v', 'm') and n.nspname = 'public' -- replace in haskell with ANY ($1)
),
transform_json as (
   select
     view_schema,
     view_name,
     replace(
      regexp_replace(
       regexp_replace(
        regexp_replace(
         replace(
          replace(
           replace(
            replace(
             view_definition,
            '({', '[{'),
           '})', '}]'),
          '} {', '},{'),
         '"', '\"'),
        '{([A-Z]+)', '{"node":"\1"', 'g'),
       ':([A-Za-z_]+) ([^[{][^:}]*)( |})', ',"\1":"\2"\3', 'g'),
      ':([A-Za-z_]+) ', ',"\1":', 'g'),
     '\ ', ' ')::jsonb AS view_definition
   from views
),
target_lists as (
   select
     view_schema,
     view_name,
     view_definition->0->'targetList' AS target_list
   from transform_json
),
target_entries as (
   select
     view_schema,
     view_name,
     jsonb_array_elements(target_list) as target_entry
   from target_lists
   where jsonb_typeof(target_list) = 'array'
),
results as (
   select
     view_schema,
     view_name,
     target_entry->>'resname' AS view_column_name,
     target_entry->>'resorigtbl' AS resorigtbl,
     target_entry->>'resorigcol' AS resorigcol
   from target_entries
)
select
   sch.nspname as table_schema,
   tbl.relname as table_name,
   col.attname as table_column_name,
   res.view_schema,
   res.view_name,
   res.view_column_name
from results res
join pg_class tbl on tbl.oid::text = res.resorigtbl
join pg_attribute col on col.attrelid = tbl.oid and col.attnum::text = res.resorigcol
join pg_namespace sch on sch.oid = tbl.relnamespace
where resorigtbl <> '0'
order by view_schema, view_name, view_column_name;
```

In theory this should allow a much more robust parsing of the view definition, but ONLY IF the json transformation works flawlessly.

Just thought I post that here, in case anyone wants to play with this, before I continue to.

### LINKED ISSUE #1635: Foreign key not detected in view with COAELSCE + subselect
### Environment

* PostgreSQL version: 10.11 (AWS RDS)
* PostgREST version: 7.0.1
* Operating system: Linux

### Description of issue

I have a `person` view with an `addressId` column that is a foreign key to the `address_tbl` table.

And now I have an `address` view like this:

```sql
-- (WORKS)
CREATE VIEW address AS
SELECT
    id,
    (SELECT ARRAY_AGG(foo) FROM address_foo WHERE address_foo."addressId" = address_tbl.id) AS foos
FROM
    address_tbl
```

Now I use the request to PostgREST, which **WORKS**:

    GET /person?select=*,addressId(*)

However, when I change the view to:

```sql
-- (BUGGY!)
CREATE VIEW address AS
SELECT
    id,
    COALESCE((SELECT ARRAY_AGG(foo) FROM address_foo WHERE address_foo."addressId" = address_tbl.id), '{}') AS foos
FROM
    address_tbl
```

Then it appears there is a bug, and when I issue the same PostgREST query as above I get the error:

    "Could not find foreign keys between these entities. No relationship found between person and addressId"

No amount of hints is able to help.

I have found a curious workaround: wrap everything in `SELECT * FROM (...)`, like this:

```sql
-- (WORKS!)
CREATE VIEW address AS
SELECT * FROM (
SELECT
    id,
    COALESCE((SELECT ARRAY_AGG(foo) FROM address_foo WHERE address_foo."addressId" = address_tbl.id), '{}') AS foos
FROM
    address_tbl
) z
```

wolfgangwalther:
Resolves #1608, resolves #1635.

This is on top of the changes in #1625. Will rebase, once the other one is merged.

Background:
Due to the nested nature of `pg_node_tree`, it is not possible to correctly parse all kinds of views with just a regex. This replaces that parsing by using the postgres `json` parser. First the `pg_node_tree` is reformatted to proper json with a series of `replace()` and `regexp_replace()` calls. After that, a cast to json is possible and the TargetEntries can easily be extracted.

The json that is produced is a very much stripped down version of the node tree, that has just two parts left:
- the fields that we require for our query
- the whole structure, i.e. all object and array brackets with some empty string keys, where required

It looks roughly like this:
```json
[
  {
    "": [
      {"":{},"":{}},
      {"":{},"":{}},
      {"":{}},
      {"":{}},
      {"":{}},
      {"":{},"":[{},{},{},{},{},{},{},{},{},{},{},{},{},{}]},
      {"":{},"":[{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{},{}]}
    ],
    "": {
      "":[{"":{},"":{"":{},"":{},"":{"":[{},{}]}},"":{"":[{},{}]}}],
      "":{"":[{},{"":[{},{},{}]}]}
    },
    "targetList": [
      {
        "": {},
        "resno": 1,
        "resorigtbl": 27118,
        "resorigcol": 1
      },
      {
        "": {},
        "resno": 2,
        "resorigtbl": 27118,
        "resorigcol": 2
      },
      {
        "": {},
        "resno": 3,
        "resorigtbl": 27118,
        "resorigcol": 3
      }
    ],
    "": [{},{}]
  }
]
```

Using the big test schema, I get 1 more available column to embed on and the following timings for n=100:

```
query  | plan [ms] | execution [ms]
master | 1.71      | 681.5
#1625  | 2.60      | 417.5
json   | 2.67      | 574.2
```

Those numbers are from a different (less powerful) machine compared to those reported in #1625. Here I don't get a factor of 2x between master and the optimized regex query, more like 1,5x. Considering that I started with ~8000ms for the json approach, I'm quite happy that I was able to make it faster than the current master.

While #1625 is even faster, the json approach has a lot of upside, not only the "subquery inside a function call" that I had fixed initially. Since we now have the full node tree available, we can extend the parsing to support set operations (unions, ...). This will be part of a diffrent PR, though. Once we support UNION, we should be able to give embedding hints (#1179) like this:

```sql
CREATE VIEW xy AS
SELECT a, b, foreign_key_column FROM base_table WHERE false
UNION
SELECT my_regular_view_which_is_not_detected_because_of_crazy_functions
FROM base_table
```

The first select would basically serve as an embedding hint producing no rows. 

_@steve-chavez: Since this is still faster than current master/v7, I think you will still be able to call it "The Performance Release" if you want to, even when you merge both #1625 and this one. ;)_

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -39,7 +39,7 @@ import GHC.Exts                      (groupWith)
 import Protolude                     hiding (toS)
 import Protolude.Conv                (toS)
 import Protolude.Unsafe              (unsafeHead)
-import Text.InterpolatedString.Perl6 (q, qc)
+import Text.InterpolatedString.Perl6 (q)
 
 import PostgREST.Private.Common
 import PostgREST.Types
@@ -49,7 +49,7 @@ getDbStructure schemas extraSearchPath pgVer prepared = do
   HT.sql "set local schema ''" -- This voids the search path. The following queries need this for getting the fully qualified name(schema.name) of every db object
   tabs    <- HT.statement () $ allTables prepared
   cols    <- HT.statement schemas $ allColumns tabs prepared
-  srcCols <- HT.statement (schemas, extraSearchPath) $ pfkSourceColumns cols pgVer prepared
+  srcCols <- HT.statement (schemas, extraSearchPath) $ pfkSourceColumns cols prepared
   m2oRels <- HT.statement () $ allM2ORels tabs cols prepared
   keys    <- HT.statement () $ allPrimaryKeys tabs prepared
   procs   <- HT.statement schemas $ allProcs prepared
@@ -699,18 +699,14 @@ pkFromRow tabs (s, t, n) = PrimaryKey <$> table <*> pure n
   where table = find (\tbl -> tableSchema tbl == s && tableName tbl == t) tabs
 
 -- returns all the primary and foreign key columns which are referenced in views
-pfkSourceColumns :: [Column] -> PgVersion -> Bool -> H.Statement ([Schema], [Schema]) [SourceColumn]
-pfkSourceColumns cols pgVer =
+pfkSourceColumns :: [Column] -> Bool -> H.Statement ([Schema], [Schema]) [SourceColumn]
+pfkSourceColumns cols =
   H.Statement sql (contrazip2 (arrayParam HE.text) (arrayParam HE.text)) (decodeSourceColumns cols)
-  -- query explanation at https://gist.github.com/steve-chavez/7ee0e6590cddafb532e5f00c46275569
+  -- query explanation at:
+  --  * rationale: https://gist.github.com/wolfgangwalther/5425d64e7b0d20aad71f6f68474d9f19
+  --  * json transformation: https://gist.github.com/wolfgangwalther/3a8939da680c24ad767e93ad2c183089
   where
-    subselectRegex :: Text
-    -- "result" appears when the subselect is used inside "case when", see `authors_have_book_in_decade` fixture
-    -- "resno"  appears in every other case
-    -- when copying the query into pg make sure you omit one backslash from \\d+, it should be like `\d+` for the regex
-    subselectRegex | pgVer < pgVersion100 = ":subselect {.*?:constraintDeps <>} :location \\d+} :res(no|ult)"
-                   | otherwise = ":subselect {.*?:stmt_len 0} :location \\d+} :res(no|ult)"
-    sql = [qc|
+    sql = [q|
       with recursive
       pks_fks as (
         -- pk + fk referencing col
@@ -738,36 +734,96 @@ pfkSourceColumns cols pgVer =
         join pg_rewrite r on r.ev_class = c.oid
         where c.relkind in ('v', 'm') and n.nspname = ANY($1 || $2)
       ),
-      removed_subselects as(
+      transform_json as (
         select
           view_id, view_schema, view_name,
-          regexp_replace(view_definition, '{subselectRegex}', '', 'g') as x
+          -- the following formatting is without indentation on purpose
+          -- to allow simple diffs, with less whitespace noise
+          replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            regexp_replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+            replace(
+              view_definition::text,
+            -- This conversion to json is heavily optimized for performance.
+            -- The general idea is to use as few regexp_replace() calls as possible.
+            -- Simple replace() is a lot faster, so we jump through some hoops
+            -- to be able to use regexp_replace() only once.
+            -- This has been tested against a huge schema with 250+ different views.
+            -- The unit tests do NOT reflect all possible inputs. Be careful when changing this!
+            -- -----------------------------------------------
+            -- pattern           | replacement         | flags
+            -- -----------------------------------------------
+            -- `,` is not part of the pg_node_tree format, but used in the regex.
+            -- This removes all `,` that might be part of column names.
+               ','               , ''
+            -- The same applies for `{` and `}`, although those are used a lot in pg_node_tree.
+            -- We remove the escaped ones, which might be part of column names again.
+            ), '\{'              , ''
+            ), '\}'              , ''
+            -- The fields we need are formatted as json manually to protect them from the regex.
+            ), ' :targetList '   , ',"targetList":'
+            ), ' :resno '        , ',"resno":'
+            ), ' :resorigtbl '   , ',"resorigtbl":'
+            ), ' :resorigcol '   , ',"resorigcol":'
+            -- Make the regex also match the node type, e.g. `{QUERY ...`, to remove it in one pass.
+            ), '{'               , '{ :'
+            -- Protect node lists, which start with `({` or `((` from the greedy regex.
+            -- The extra `{` is removed again later.
+            ), '(('              , '{(('
+            ), '({'              , '{({'
+            -- This regex removes all unused fields to avoid the need to format all of them correctly.
+            -- This leads to a smaller json result as well.
+            -- Removal stops at `,` for used fields (see above) and `}` for the end of the current node.
+            -- Nesting can't be parsed correctly with a regex, so we stop at `{` as well and
+            -- add an empty key for the followig node.
+            ), ' :[^}{,]+'       , ',"":'              , 'g'
+            -- For performance, the regex also added those empty keys when hitting a `,` or `}`.
+            -- Those are removed next.
+            ), ',"":}'           , '}'
+            ), ',"":,'           , ','
+            -- This reverses the "node list protection" from above.
+            ), '{('              , '('
+            -- Every key above has been added with a `,` so far. The first key in an object doesn't need it.
+            ), '{,'              , '{'
+            -- pg_node_tree has `()` around lists, but JSON uses `[]`
+            ), '('               , '['
+            ), ')'               , ']'
+            -- pg_node_tree has ` ` between list items, but JSON uses `,`
+            ), ' '             , ','
+            -- `<>` in pg_node_tree is the same as `null` in JSON, but due to very poor performance of json_typeof
+            -- we need to make this an empty array here to prevent json_array_elements from throwing an error
+            -- when the targetList is null.
+            ), '<>'              , '[]'
+          )::json as view_definition
         from views
-      ),
-      target_lists as(
-        select
-          view_id, view_schema, view_name,
-          string_to_array(x, 'targetList') as x
-        from removed_subselects
-      ),
-      last_target_list_wo_tail as(
-        select
-          view_id, view_schema, view_name,
-          (string_to_array(x[array_upper(x, 1)], ':onConflict'))[1] as x
-        from target_lists
       ),
       target_entries as(
         select
           view_id, view_schema, view_name,
-          unnest(string_to_array(x, 'TARGETENTRY')) as entry
-        from last_target_list_wo_tail
+          json_array_elements(view_definition->0->'targetList') as entry
+        from transform_json
       ),
       results as(
         select
           view_id, view_schema, view_name,
-          substring(entry from ':resno (\d+)')::int as view_column,
-          substring(entry from ':resorigtbl (\d+)')::oid as resorigtbl,
-          substring(entry from ':resorigcol (\d+)')::int as resorigcol
+          (entry->>'resno')::int as view_column,
+          (entry->>'resorigtbl')::oid as resorigtbl,
+          (entry->>'resorigcol')::int as resorigcol
         from target_entries
       ),
       recursion as(
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -454,6 +454,12 @@ spec actualPgVersion = do
           [json|[{"title":"1984","first_publisher":"Secker & Warburg","author":{"name":"George Orwell"}}]|]
           { matchHeaders = [matchContentTypeJson] }
 
+      it "works with views that have subselects in a function call" $
+        get "/authors_have_book_in_decade2?select=*,books(title)&id=eq.3"
+          `shouldRespondWith`
+            [json|[ {"id":3,"name":"Antoine de Saint-Exupéry","has_book_in_forties":true,"has_book_in_fifties":false,
+                     "has_book_in_sixties":false,"books":[{"title":"The Little Prince"}]} ]|]
+
       it "works with views that have CTE" $
         get "/odd_years_publications?select=title,publication_year,first_publisher,author:authors(name)&id=in.(1,2,3)" `shouldRespondWith`
           [json|[
--- a/test/fixtures/privileges.sql
+++ b/test/fixtures/privileges.sql
@@ -96,6 +96,7 @@ GRANT ALL ON TABLE
     , jsonb_test
     , authors_books_number
     , authors_have_book_in_decade
+    , authors_have_book_in_decade2
     , forties_and_fifties_books
     , odd_years_publications
     , foos
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -1433,6 +1433,24 @@ select
   end as has_book_in_sixties
 from private.authors x;
 
+create view test.authors_have_book_in_decade2 as
+select
+  id,
+  name,
+  coalesce(
+    (select true from test.forties_books where author_id=x.id limit 1),
+    false
+  ) as has_book_in_forties,
+  coalesce(
+    (select true from test.fifties_books where author_id=x.id limit 1),
+    false
+  ) as has_book_in_fifties,
+  coalesce(
+    (select true from test.sixties_books where author_id=x.id limit 1),
+    false
+  ) as has_book_in_sixties
+from private.authors x;
+
 create view test.forties_and_fifties_books as
 select x.id, x.title, x.publication_year, y.name as first_publisher, x.author_id
 from (


steve-chavez:
@wolfgangwalther You're on fire :fire:!

I was rushing to do #1633, but tomorrow I'll get back to reviewing #1625.

"The Performance Release" name will stay ;)

wolfgangwalther:
rebase on current version of #1625

wolfgangwalther:
rebased on master

wolfgangwalther:
just a rebase... to get rid of the red CI checks. not true! This one is good!

wolfgangwalther:
With some of the other PRs done now - @steve-chavez did you have a chance to look at this one already? Is there any kind of testing that we should still do in your opinion?

I have one point, that we need to think through:
I'm quite confident that the JSON parsing for pg_node_tree should work for all trees we will see. But if for whatever reason one of them fails, the whole query will fail. I don't know a way of casting to JSON with a silent `NULL` fallback if casting fails - short of creating a stored procedure with exception handling. But we don't have our `pgrst` schema, yet.

I'm not entirely sure, but I think that the schema query currently fails silently if an sql exception is raised. I don't think we get log output of that. What we should really do is add proper error handling to those schema queries - so that we get log output on the console when that happens. Since we never expect any errors in those queries, they could include a line similiar to "The following error occured... This is a bug in PostgREST. Please report this at githuburl with this error message".

src/PostgREST/DbStructure.hs
```diff
@@ -738,36 +731,100 @@ pfkSourceColumns cols pgVer =
...
+            ), '{('              , '('
+            -- Every key above has been added with a `,` so far. The first key in an object doesn't need it.
+            ), '{,'              , '{'
+            -- pg_node_tree has `()` around lists, but JSON uses `[]`
+            ), '('               , '['
+            ), ')'               , ']'
+            -- pg_node_tree has ` ` between list items, but JSON uses `,`
+            ), '} {'             , '},{'
+            ), '] ['             , '],['
+            -- `<>` in pg_node_tree is the same as `null` in JSON, but due to very poor performance of json_typeof
```
steve-chavez, line 809: > but due to very poor performance of json_typeof

@wolfgangwalther Oh, how bad is it? We're using json_typeof for our INSERTs/UPDATEs.

src/PostgREST/DbStructure.hs
```diff
@@ -699,18 +699,11 @@ pkFromRow tabs (s, t, n) = PrimaryKey <$> table <*> pure n
   where table = find (\tbl -> tableSchema tbl == s && tableName tbl == t) tabs
 
 -- returns all the primary and foreign key columns which are referenced in views
-pfkSourceColumns :: [Column] -> PgVersion -> Bool -> H.Statement ([Schema], [Schema]) [SourceColumn]
-pfkSourceColumns cols pgVer =
+pfkSourceColumns :: [Column] -> Bool -> H.Statement ([Schema], [Schema]) [SourceColumn]
+pfkSourceColumns cols =
   H.Statement sql (contrazip2 (arrayParam HE.text) (arrayParam HE.text)) (decodeSourceColumns cols)
-  -- query explanation at https://gist.github.com/steve-chavez/7ee0e6590cddafb532e5f00c46275569
```
steve-chavez, line 705: @wolfgangwalther I think this one will be hard to understand by just looking at the code. Even on my prior query I had to think hard to remember what I did.

Could you add a comment here explaining the general process?

> First the pg_node_tree is reformatted to proper json with a series of replace() and regexp_replace() calls. After that, a cast to json is possible and the TargetEntries can easily be extracted.

Even a comment like the above would be helpful.

If you have some sample node trees like on my prior gist, that would be helpful as well. You can also leave a gist of your own if you'd like to.

src/PostgREST/DbStructure.hs
wolfgangwalther: I remember this was really bad. I might be able to do some testing, to confirm that. With `json_typeof` the query looked a bit different - so maybe it wasn't directly json_typeof, but the other query structure in general.

src/PostgREST/DbStructure.hs
wolfgangwalther: Yes that makes a lot of sense. I removed the link to your gist, because big parts of it don't apply anymore (the regex stuff). But the basics still do and you're absolutely right that this could use a bit more explanation here. I tried to comment the json conversion as much as possible - but that does not give any idea of the bigger picture here.

I will try to add more comments.

steve-chavez:
> Using the big test schema, I get 1 more available column to embed on and the following timings for n=100:
query  | plan [ms] | execution [ms]
master | 1.71      | 681.5
json   | 2.67      | 574.2

Bummer, I'm seeing that the new query is about x2 slower than the old one. Same setup with the `timeit` function(n=100):

| query | plan(ms) | execution(ms) |
| ----- | -------- | ------------- |
| master   | 3.38227  | 618.11523     |
| json   | 3.52706  | 1450.29681    |

@wolfgangwalther Have you tested the query again recently? What could be wrong?

steve-chavez:
>  I get 1 more available column to embed 

I get 10 more records on the big schema when running the new query(578 on the old, and 588 on the new), perhaps #1625 changed the perf here?

wolfgangwalther:
I just did another performance test. This is what I did to set up the test:
- **old master**: Took query from commit 65968b532060bc59452e354c96888b85057e67b7 (just before merging #1625), replaced `$1` with `'{apflora}'`, replaced `{subselectRegex}` with the regex for "otherwise", so >= pg10, remember to change `\\d` to `\d` here.
- **current master**: took current master, replaced `$1 || $2` with `'{apflora,public}'` and further down `$1` with `'{apflora}'`, same as for old master regarding the regex
- **json PR**: took this PR, replaced the parameters the same as on current master, no replacement of regex, because this is not there anymore

You can find the whole SQL here: https://pastebin.com/9CYCzCxc

Here are my results (n=100) - running on PG 12.5:

| query | plan | execution | count |
| -- | --: | --: | --: |
| old master | 0.72 | 460.0 | 2243 |
| master | 1.40 | 258.7 | 587 |
| json PR | 1.34 | 337.3 | 588 |

This matches my expectations:
1. **number of rows**: old master returned all columns, we changed that to only return pk and fk columns in #1625. In the big schema there is only 1 new source column detected, this must be because of some subquery in a view that now does not break parsing. If I remember correctly, the difference in rows for you is because you forgot to change `\\d` to `\d` in the regex.
2. **performance**: as previously advertised, the json is not as fast as the previous, optimized regex, that is currently on master. However, this is not a performance degradation, because the big picture is what counts. old master -> json PR is still an improvement. This is kind of similar to the introduction of prepared statements: Even though the `set_config` was a bit slower, in total this is still a nice performance improvement.
3. **relative performance**: Why is the relative execution time ~ 1.3x for me and 2.3x for you, when comparing master and json? My understanding is that the numbers we see are a sum of two components: I call one of them "base" and the other "regex". That might not reflect what is really happening in those two parts, but the conclusion should still hold: We do have different machines and my CPU is quite a bit faster (Ryzen 3950X). My assumption is that the base component of the query (select parsing, IO, ...) does not make that much of a difference between our machines, but the whole regex parsing does differ a lot. I guess this is a lot more dependent on CPU. So the increased complexity in replacing/regexing to the json does not slow me down as much as it does you. I would be really interested in seeing your numbers for **old master** as well though. I really hope they are higher than what you get for the json - so that the "old master -> json" = "improving performance" still holds :).

Of course there could also be other things influencing performance. Which PG version are you testing on?

> > Using the big test schema, I get 1 more available column to embed on and the following timings for n=100:
> > query  | plan [ms] | execution [ms]
> > master | 1.71      | 681.5
> > json   | 2.67      | 574.2
> 
> Bummer, I'm seeing that the new query is about x2 slower than the old one. Same setup with the `timeit` function(n=100):
> query 	plan(ms) 	execution(ms)
> master 	3.38227 	618.11523
> json 	3.52706 	1450.29681
> 
> @wolfgangwalther Have you tested the query again recently? What could be wrong?

Just to make that clear again: **master** is not the same between what I posted back then and what you posted here. Back then master was without #1625, your master is with that included. So it is **expected** that this PR is slower than current master.

wolfgangwalther:
> > I get 1 more available column to embed
> 
> I get 10 more records on the big schema when running the new query(578 on the old, and 588 on the new), perhaps #1625 changed the perf here?

In case it get's lost in the big text: you need to change `\\d` to `\d` in the regex, if I remember correctly.

steve-chavez:
> because you forgot to change \\d to \d in the regex

Argh, forgot about that one.

> I would be really interested in seeing your numbers for old master as well though

Sure. I just ran your https://pastebin.com/9CYCzCxc. I had to delete the count though, it erred with `column "last_count" does not exist`.

```
┌────────────┬──────┬───────────┐
│   query    │ plan │ execution │
├────────────┼──────┼───────────┤
│ old master │ 1.13 │     824.0 │
│ master     │ 3.44 │     627.0 │
│ json PR    │ 3.44 │    1452.8 │
└────────────┴──────┴───────────┘
```

> Which PG version are you testing on?

It's on 11.5.

steve-chavez:
Ran the same test on a newer laptop(system76 oryx pro), pg 12.5:

```
   query    | plan | execution
------------+------+-----------
 old master | 0.57 |     288.7
 master     | 1.06 |     173.6
 json PR    | 1.02 |     258.2
```

<details>
<summary>Oryx pro lscpu(16 CPUs)</summary>
<br>
<pre>
$ lscpu 
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   39 bits physical, 48 bits virtual
CPU(s):                          16
On-line CPU(s) list:             0-15
Thread(s) per core:              2
Core(s) per socket:              8
Socket(s):                       1
NUMA node(s):                    1
Vendor ID:                       GenuineIntel
CPU family:                      6
Model:                           165
Model name:                      Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
Stepping:                        2
CPU MHz:                         1905.241
CPU max MHz:                     5100.0000
CPU min MHz:                     800.0000
BogoMIPS:                        4599.93
Virtualization:                  VT-x
L1d cache:                       256 KiB
L1i cache:                       256 KiB
L2 cache:                        2 MiB
L3 cache:                        16 MiB
NUMA node0 CPU(s):               0-15
Vulnerability Itlb multihit:     KVM: Mitigation: VMX disabled
Vulnerability L1tf:              Not affected
Vulnerability Mds:               Not affected
Vulnerability Meltdown:          Not affected
Vulnerability Spec store bypass: Mitigation; Speculative Store Bypass disabled via prctl and
                                 seccomp
Vulnerability Spectre v1:        Mitigation; usercopy/swapgs barriers and __user pointer sani
                                 tization
Vulnerability Spectre v2:        Mitigation; Enhanced IBRS, IBPB conditional, RSB filling
Vulnerability Srbds:             Not affected
Vulnerability Tsx async abort:   Not affected
</pre>
</details>

wolfgangwalther:
> Sure. I just ran your https://pastebin.com/9CYCzCxc. I had to delete the count though, it erred with `column "last_count" does not exist`.

Ah, yes - that was my upgraded timeit function, sorry. It only adds the count, though.


> Oryx pro lscpu(16 CPUs)
> 
> @wolfgangwalther Seems this one is faster than your Ryzen stuck_out_tongue. I'll try pg 12.5 on the old toshiba just in case.

Whooooot? I gotta get a new CPU!! :D

Small correction, though: I just realized that I was actually running **PG 13**, sorry. But the numbers of 12.4 were exactly the same back then, so I don't think there is a difference.

> Yeah, this one is an old toshiba laptop, running pg 11.5. Here's the lscpu output:
> Toshiba lscpu(4 CPUs)

I do think, however that there is a difference between 11.5 and 12/13? Or maybe it is the machine.

---

So.. let's not think too much about why your laptop is faster than my Ryzen... terrible... but focus on the fact, that they both show the same pattern, kind of.

But the Toshiba / PG 11 shows something entirely different. If you could get some numbers on PG12/13 on that laptop, that would be cool. I'm going to run the test on PG 11 on my machine now. That should give us a good idea of whats happening.

wolfgangwalther:
I went back and did the test for some more versions. I only ran `n=10`, but my experience tells me that the numbers are quite consistent with that as well. I removed the plan time and the count and am focusing on the meat, the execution time:

| query | pg10 | pg11 | pg12 | pg13 |
| -- | --: | --: | --: | --: |
| old master | 311.6 | 295.7 | 458.5 | 479.2 |
| master | 230.3 | 222.2 | 253.0 | 263.0 |
| json PR | 523.9 | 504.8 | 319.5 | 336.6 |

PG just got slower on the regular regex and faster on the JSON stuff between 11 and 12. I think the CPU stuff still plays a role, though, as one can tell from your laptop data. Not only is it faster, but the relations between those queries are also different.

One thing to note here: The json query has been optimized while running PG12. So maybe it could be optimized to run faster with the older versions as well. But I don't think we need that. We should mostly be concerned about performance with newer PG versions, I think. Especially on this query, which is "just" the schema cache and startup time.

steve-chavez:
Can confirm the pattern on the old toshiba with pg 12.4:

| query      | plan | execution  |
| ---------- | ---- | ---------- |
| old master | 2.34 | 1233.7     |
| master     | 4.70 | 723.4      |
| json PR    | 4.99 | 1029.4     |

> We should mostly be concerned about performance with newer PG versions, I think.

Yeah, totally agree. But should we document this change somehow? Like "#1632, expect a faster startup time for pg > 12, a bit slower for pg < 12"? Might get some users to upgrade pg as well(if the change is noticeable for them).

wolfgangwalther:
> Yeah, totally agree. But should we document this change somehow? Like "#1632, expect a faster startup time for pg > 12, a bit slower for pg < 12"? Might get some users to upgrade pg as well(if the change is noticeable for them).

Yes, that makes sense!

steve-chavez:
So this one only needs some comments added.

Besides that, it should be good to go :+1:

src/PostgREST/DbStructure.hs
wolfgangwalther: So I just tested this for this query again. The changes I made where:
```
), '<>'              , '[]'
```
becomes
```
), '<>'              , 'null'
```
(so no target list is encoded as null instead of an empty array)

and then

```
target_entries as(
        select
          view_id, view_schema, view_name,
          json_array_elements(view_definition->0->'targetList') as entry
        from transform_json
        where json_typeof(view_definition->0->'targetList') = 'array' -- this line is added
      ),
```

so empty target lists are now filtered by json_typeof.

The changed replace should have no effect, so only json_typeof makes things slower here. Here are the results:

| query | execution |
| -- | --: |
| old master | 415.0 |
| master | 232.6 |
| json PR | 321.9 |
| json_typeof | 714.1 |

So just the json_typeof call takes more time then everything else combined - the time is more than doubled!

wolfgangwalther: I tested a bit more and even without `json_typeof`, just adding the following make the query just as slow:

```
where view_definition::text is not NULL
```

This is effectively a no-op - but the cast to `text` is taking quite a bit of time. It probably has to go through the whole json object. The same thing will happen here for `json_typeof`.

We're currently using json_typeof here:
https://github.com/PostgREST/postgrest/blob/11d62a80105c21696e65b70acf07b6f47405ca68/src/PostgREST/Private/QueryFragment.hs#L40-L53

Maybe we can just check the first non-whitespace character of the body we're passing in, in haskell. If it's a `[` use the array code, if it's a `{` use the object code. Could give us a performance improvement. Or not. Would have to loadtest this, of course :)

src/PostgREST/DbStructure.hs
wolfgangwalther: I forked your gist and took the regex stuff out. This is part #1. Then I wrote part #2 about the json transformation in detail.

First of all, this is lots of detail, very good to come back to that in the future. But even better - explaining it in detail, I realized that I could skip one of the replaces altogether. My timings are now:

| query | execution |
| -- | --: |
| old master | 406.5 |
| master | 229.8 |
| json before | 317.5 |
| json now | 301.8 |

wolfgangwalther: You probably won't notice on your super fast machine, but on my slow Ryzen it's 16ms faster ;)

src/PostgREST/DbStructure.hs
steve-chavez: Ahh.. I see it now! Maybe we could make the `[ or {` check  on pg?

I feel like messing with the payload on Haskell might bring https://github.com/PostgREST/postgrest/issues/690 again.

Or perhaps we should just accept the `json_typeof` overhead. It's nice to be aware of it though.

src/PostgREST/DbStructure.hs
steve-chavez: Awesome! That's a really nice explanation Wolfgang, I feel we have at least 2 blog posts with those gists :)

> You probably won't notice on your super fast machine, but on my slow Ryzen it's 16ms faster ;)

I'll try the new query to see if I can beat your time again :smile:. Do you have a quick pastebin for me? :)

wolfgangwalther: https://pastebin.com/6x3wTJVX

Here you go :)

src/PostgREST/DbStructure.hs
wolfgangwalther: > Ahh.. I see it now! Maybe we could make the `[ or {` check on pg?

Yeah, I thought so. That was the reason why I tried `where view_definition::text is not NULL`. Because to do the check, we need to cast the json back to text first. Just as slow as `pg_typeof`.

> I feel like messing with the payload on Haskell might bring #690 again.

Yeah, we would need to make sure that we're not parsing the body at all. We just want to read the first bytes of it. Isn't that kind of lazy evaluation one of haskell's strength? If we do it properly of course.

That still does not mean the json is correct in any way. But in that case PG would still throw.

Plus, it would also solve #1323. I tried to debug that today, but I don't see why this is happening. This would be fixed by:
- if `[` -> recordset
- if `{` -> record
- else -> error

wolfgangwalther:
Ready from my side, will just wait for CI before merging.

src/PostgREST/DbStructure.hs
steve-chavez: | query       | execution          |
| ----------- | ------------------ |
| old master  | 282.7              |
| master      | 168.6              |
| json before | 249.2              |
| json now    | 232.0              |

17 ms faster for me. Beat you again :laughing:

wolfgangwalther: Ah, wait a second. Between my latest test and yours... I removed another replace. It didn't show much of a difference in perf, so I didn't post it... but your test was running with both of them removed..

So let me run my tests again... let's see whether I can get another 1 or 2 ms. :D

wolfgangwalther: 319.9-303.5 = 16.4. :(

I'm leaving now to do something else. This is annoying :D

src/PostgREST/DbStructure.hs
wolfgangwalther: Still had this one my todo list - but I guess we can call this resolved. My tests were flawed, it wasn't really `json_typeof` that was performing that bad, it was the "accessing nested json values" part that came with the approach.

`json_typeof` does have a little bit of overhead - but that's to be expected when it's called ~2800 times when querying the schema cache in *the big schema*. I don't see a problem when it's called once per request.

Issue #1578, opened 2020-08-30
Feature request: show SQL query in terminal(log-level=debug)


sbayeta:
Hi guys, congratulations on this excellent software!

Could you please consider adding a configuration option to output the SQL query sent to the db with each request? I was curious about the generated SQL by a nested embedded resource request. I was able to do it by enabling statement logging in the database, but I thought it would be nice to be able to see it in the console running postgREST.

Thanks for your terrific work!

Cheers!
Santiago

swuecho:
You might find this issue helpful. https://github.com/PostgREST/postgrest/issues/1573

steve-chavez:
@sbayeta Thanks!

We've always recommended looking at the pg logs because they are more complete, but I can see how printing the SQL ourselves might give better UX.

I think this would be a "debug" logging level. This level would only show the main SQL query and not our `SET LOCAL` stuff. We'll still recommend looking at the pg logs for more information.

Related to https://github.com/PostgREST/postgrest/issues/540.

sbayeta:
Hi Steve, thank you for your excellent support!

I read the issue you mentioned, as well as swuecho's, and I get there's a lot of nuances regarding this.

I'm not sure how valuable this feature would be for others, specially when weighing the details mentioned in the linked issues. For me, getting the pg logs to work was a bit of a hassle since I haven't done that before, but I guess it's something one has to learn in order to maintain a pb db propperly. Now I know how to do it, but it's still cumbersome to enable and (not forget to) disable logging, and look for the specific entries of interest (in my case, pgadmin4 floods the logs so they were greater than 1MB with only one pgrest request).

So if there's a way to catch at least the simple use case of seeing the generated SQL right in the console by setting the postgrest log level, I would probably be a nice addition.

Thanks again for your support and congratulations.

Cheers,
Santiago

wolfgangwalther:
> I think this would be a "debug" logging level. This level would only show the main SQL query and not our `SET LOCAL` stuff. We'll still recommend looking at the pg logs for more information.

Maybe we can make this orthogonal to the logging level? If this was another logging setting, that would enable the logging of queries, it could be a bit easier to find the correct query.

So, e.g. assume that I only care about the troublesome stuff, so I set the logging level to `error`. It would be good, if I could now output the query only in that case as well, without suddenly dumping ALL requests to the log via logging level `debug`.

Thoughts?

steve-chavez:
[...]

PR #1633, opened 2020-10-21
perf: Prepared statements for GET

steve-chavez:
Use [hasql-dynamic-statements](http://hackage.haskell.org/package/hasql-dynamic-statements) to:

- Parametrize filters(`WHERE name = $1`) instead of inlining and escaping them(`WHERE name = '<escaped>'`).
- Enable prepared statements for GET.

With this change we get more requests/sec on top of the improvement on https://github.com/PostgREST/postgrest/pull/1631#issuecomment-712844918.

|   | req/s |
| ------------- | ------------- |
| GETSingle - Bytestring queries  | 1748.747079  |
| GETSingle - Prepared  | 1980.830129/s  |

([GETSingle](https://github.com/steve-chavez/postgrest-benchmark/blob/master/k6/GETSingle.js#L30) load test on a AWS t3a.nano)

Which gets us close to our previously advertised [2000 req/s](https://github.com/PostgREST/postgrest#performance)!

### Implementation

The magic really is in the [encoderAndParam](http://hackage.haskell.org/package/hasql-dynamic-statements-0.3.1/docs/Hasql-DynamicStatements-Snippet.html#v:encoderAndParam) function. The rest is changing our `SqlQuery` type to the [H.Snippet](http://hackage.haskell.org/package/hasql-dynamic-statements-0.3.1/docs/Hasql-DynamicStatements-Snippet.html#t:Snippet) type, with help from the [H.sql](http://hackage.haskell.org/package/hasql-dynamic-statements-0.3.1/docs/Hasql-DynamicStatements-Snippet.html#v:sql) function.

(It took a lot of work to make the conversion gradual!)

### Pending

- [x] Parametrize LIMIT and OFFSET
- [x] Single parameter for the IN operator. Right now it's like `IN ($1, $2, $3)` and I think we could do `= ANY({$1})`. This would be good for reducing the prepared statements cache size on the Postgres side.
- [x] Parametrize json path index
- [x] Some refactoring. I've tried to not convert all the `SqlQuery` and `SqlFragment` to `Snippet` to keep the change simpler.

### Notes

- Not all statements have to be parametrized.
  - For example `SET LOCAL "my.setting" = 'value'` cannot be parametrized,
  - The `IS TRUE/FALSE/NULL` operator cannot be parametrized.
  - `hasql-dynamic-statements` is very flexible in this regard. So we can have a mix of parametrized/inline inputs if necessary.
- Possible future improvement: limit the prepared statements cache size on Hasql. I've had a discussion on https://github.com/nikita-volkov/hasql-dynamic-statements/issues/3 about why this would be a good safeguard.


wolfgangwalther:
> With this change we get more requests/sec

master ~1500 req/s -> parametrized 2000 req/s? Wow! Great job!

>     * For example `SET LOCAL "my.setting" = 'value'` cannot be parametrized,

Now, I would be very much interested in the `set_config` approach in #1600. How would that perform with a parametrized query?

>     * The `IS TRUE/FALSE/NULL` operator cannot be parametrized.

After a quick test it seems like we could use `IS NOT DISTINCT FROM` instead of `IS` for that in parametrized queries. I didn't think too much about it yet, but it should be the same. Not sure whether it has the same performance, but I don't really see why not.

wolfgangwalther:
> Not sure whether it has the same performance, but I don't really see why not.

I did some tests on that - and it seems like `IS DISTINCT FROM` doesn't use indexes compared to just `IS`. So this does not seem to be an option anyway.

src/PostgREST/Statements.hs
```diff
@@ -77,30 +79,30 @@ createWriteStatement selectQuery mutateQuery wantSingle isInsert asCsv rep pKeys
...
    fromMaybe (Nothing, 0, [], mempty, Right [], Right Nothing) <$> HD.rowMaybe standardRow
 
-createReadStatement :: SqlQuery -> SqlQuery -> Bool -> Bool -> Bool -> Maybe FieldName -> PgVersion ->
+createReadStatement :: H.Snippet -> H.Snippet -> Bool -> Bool -> Bool -> Maybe FieldName -> PgVersion ->
                        H.Statement () ResultsWithCount
 createReadStatement selectQuery countQuery isSingle countTotal asCsv binaryField pgVer =
-  H.Statement sql HE.noParams decodeStandard False
+  H.dynamicallyParameterized snippet decodeStandard True
  where
-  sql = [qc|
```
wolfgangwalther, line 92: I find those quasi-quoted statements a lot easier to read. Do we need to change those?

Reading http://hackage.haskell.org/package/interpolatedstring-perl6-1.0.2/docs/Text-InterpolatedString-Perl6.html I expect this to work as well:

```
snippet = [qc|
  WITH
  {H.sql sourceCTEName} AS ({selectQuery})
...
```

Coolest thing would be to have a `[sql|` quasi-quoter that takes snippets and strings and returns a snippet or something like that?

wolfgangwalther: There seems to be `[uncheckedSql|` here: https://hackage.haskell.org/package/hasql-th-0.4.0.8/docs/Hasql-TH.html#g:4

Not sure if that helps with DynamicStatements though.

steve-chavez: Yes, I agree. They were easier to read. Also they looked better on pg logs.

I changed them because interpolating the `Snippet` didn't work. We also had a discussion a long time ago about removing `qc` because it had a perf impact(didn't confirm).

> Coolest thing would be to have a [sql| quasi-quoter that takes snippets and strings and returns a snippet or something like that?

Looks possible! I'll check that out.

dwagin:
Awesome feature!

But I have some question about it:
1. Does this completely free us from escaping parameters?
2. Do params send as string or binary representation?
3. What about custom pg-types and its caching?
4. Connection poolers (ex, [PgBouncer](https://www.pgbouncer.org/), [Odyssey](https://github.com/yandex/odyssey)) in transaction mode don't support prepared statements.
Can you add option for disable preparing?

steve-chavez:
> Now, I would be very much interested in the set_config approach in #1600. How would that perform with a parametrized query?

@wolfgangwalther Yes, I think we could optimize #1600 by using a prepared statement for `set_config`. I was thinking something along the lines of:

```sql
with pgrst_configs as(
  $1 -- somehow grab whole value here
)
select set_config(pgrst_configs.key, pgrst_configs.val, true) from pgrst_configs;
```
We could have all of the configs converted to a single JSON value. Though having multiple parameters would work fine(the statement would vary and get larger with more headers, but maybe that's not a problem).

Just an idea because Aeson has proven to be slow before. Currently looking at https://github.com/nikita-volkov/jsonifier for a possible enhancement there.

steve-chavez:
@dwagin Thanks!

> 1. Does this completely free us from escaping parameters?

Not completely as there are some inputs that can't be parametrized like the `IS TRUE/FALSE/NULL` and `SET LOCAL`(which could change to `set_config` and be parametrized). But all the rest of the filters will be parametrized.

> 2. Do params send as string or binary representation?
> 3. What about custom pg-types and its caching?

The params are sent using the pg protocol 'text' format and we let pg infer the value's type. This is done by using
the [Hasql unknown encoder](http://hackage.haskell.org/package/hasql-1.4.4.2/docs/Hasql-Encoders.html#v:unknown).

Do you foresee any issues with that? What kind of values would not work?

> 4. Connection poolers (ex, PgBouncer, Odyssey) in transaction mode don't support prepared statements.

Ah.. you're right. That makes sense since prepared statements are cached at the session level.

> Can you add option for disable preparing?

Sure, I can add a config like `db-prepared-statements`. I think it should be `true` by default though, as the perf gains would benefit most of the users. We can have a warning for connection poolers.

Curious here, our POST/PUT/PATCH/RPC have been prepared(just parametrizing the payload) for some time, how did you managed to make PostgREST work there? Or do you only do GETs?

wolfgangwalther:
> Curious here, our POST/PUT/PATCH/RPC have been prepared(just parametrizing the payload) for some time, how did you managed to make PostgREST work there? Or do you only do GETs?

[This](https://github.com/nikita-volkov/hasql-dynamic-statements/issues/3) reads like the queries so far had been parametrized, but not prepared. Or is that not related in this case?

steve-chavez:
@wolfgangwalther No, INSERT/UPDATE/RPC have been parametrized and prepared for some time. Check the True at the end:

https://github.com/PostgREST/postgrest/blob/eed0d3a66c48f7fec8d3b31708aa8be6f7bbc0be/src/PostgREST/Statements.hs#L48-L49

https://github.com/PostgREST/postgrest/blob/eed0d3a66c48f7fec8d3b31708aa8be6f7bbc0be/src/PostgREST/Statements.hs#L134-L135


Now that I remember, I think @dwagin has been patching PostgREST  for not preparing them https://github.com/PostgREST/postgrest/pull/1542#issuecomment-64608170

dwagin:
> Do you foresee any issues with that? What kind of values would not work?

I had problems with Go language (pgx driver)

> Now that I remember, I think @dwagin has been patching PostgREST for not preparing them [#1542 (comment)](https://github.com/PostgREST/postgrest/pull/1542#issuecomment-64608170)

this patch completely removes the ability to use prepared queries

dwagin:
> > Can you add option for disable preparing?
> 
> Sure, I can add a config like `db-prepared-statements`. I think it should be `true` by default though, as the perf gains would benefit most of the users. We can have a warning for connection poolers.
> 
> Curious here, our POST/PUT/PATCH/RPC have been prepared(just parametrizing the payload) for some time, how did you managed to make PostgREST work there? Or do you only do GETs?

It is not enough to fix PostgREST to support connection poolers, need also fix [hasql-transaction](https://github.com/nikita-volkov/hasql-transaction).

dwagin:
For example, these lines at [hasql-transaction](https://github.com/nikita-volkov/hasql-transaction) are hardcoded in permanent mode:

https://github.com/nikita-volkov/hasql-transaction/blob/56dafd8ac0ea5abd70b93167dec72ee51a826a63/library/Hasql/Transaction/Private/Statements.hs#L17
https://github.com/nikita-volkov/hasql-transaction/blob/56dafd8ac0ea5abd70b93167dec72ee51a826a63/library/Hasql/Transaction/Private/Statements.hs#L21
https://github.com/nikita-volkov/hasql-transaction/blob/56dafd8ac0ea5abd70b93167dec72ee51a826a63/library/Hasql/Transaction/Private/Statements.hs#L25

nix/overlays/haskell-packages.nix
```diff
@@ -17,6 +17,33 @@ let
...
+      # TODO: We need to patch upstream for unbreaking hasql-dynamic-statements, hasql-implicits, ptr
+      hasql-dynamic-statements =
+        self.haskell.lib.dontCheck (prev.callHackageDirect
+          {
+            pkg = "hasql-dynamic-statements";
+            ver = "0.3.1";
+            sha256 = "1zjv91xlfkyxwq6mhzj7rsfm4kjvs9ygkgbl6jbbg19jihcn2kiy";
+          }
+          { }
+        );
```
steve-chavez, line 30: @monacoremo This might be one for you. Whenever you have the time :)

Btw, I learned the `lib.dontCheck` trick thanks to this post: https://discourse.nixos.org/t/nix-haskell-development-2020/6170.

src/PostgREST/Statements.hs
steve-chavez: @wolfgangwalther What do you think about leaving the refactor for another PR? I'd like to merge and see if I can improve #1600 with hasql-dynamic.

We'll be definitely working on the SQL fragments and we'll notice if it's hard to work with these new Snippets.

steve-chavez:
@dwagin I've added the `db-prepared-statements` config option! I noticed our schema cache queries were also prepared so now those are optionally prepared as well.

> For example, these lines at hasql-transaction are hardcoded in permanent mode:

I see what you mean, while trying to come with a reproducible test for prepared statements I noticed:

```bash
curl 'l:3000/prepared_statements?select=name,statement'

[{"name":"2","statement":"ABORT"},
 {"name":"0","statement":"BEGIN ISOLATION LEVEL READ COMMITTED READ ONLY"},
 {"name":"1","statement":"COMMIT"}]
```

This is despite setting `db-prepared-statements=false`. So as you mention, hasql-transaction is currently hardcoding those queries.

I wonder if preparing those short statements is really a perf advantage. Anyway, that should be discussed on a hasql-transaction issue later on.

src/PostgREST/Statements.hs
wolfgangwalther: > What do you think about leaving the refactor for another PR? I'd like to merge and see if I can improve #1600(the SET LOCALs) with hasql-dynamic.

Yes, of course.

> We'll be definitely working on the SQL fragments and we'll notice if it's hard to work with these new Snippets.

If this also makes them less nice to look at, we might want to do something about it for #1578. I'm sure we'll pick up on this.

steve-chavez:
CirrusCI keeps failing - doesn't have enough time for building. I think we'll need to define a timeout similarly to the other CIs.

wolfgangwalther:
Hm. I didn't want to commit that immediately. I wanted to "suggest" a commit like you do it sometimes, @steve-chavez. How does that work?

steve-chavez:
@wolfgangwalther I use this suggestion button https://til.hashrocket.com/posts/szgdi4at59-github-insert-a-suggestion-. It shows when leaving a comment.

steve-chavez:
I see what you mean regarding the config help. I left the `db-prepared-statements` at the end because it's going to be rarely used(only when behind connection poolers). Also wanted to "hide it" a bit. But I guess it's fine with the warning I left there.

steve-chavez:
Cirrus seems fine with the timeout now. Should be good to merge.

PR #1636, opened 2020-10-28
perf: shortcut for proc with no variadic arg 

steve-chavez:
On https://github.com/PostgREST/postgrest/pull/1603, logic for using VARIADIC functions was added. However that logic is also executed for every non-variadic function.

This PR changes that. By adding a cached `pdHasVariadic` to `ProcDescription` we avoid extra steps for non-variadic functions.

Besides that:

- Isolates the logic for changing the RPC params `/rpc/func?arg1=val1&arg2=val2` to json in a single function. This makes that part of the code more understandable IMO.
- Includes some refactors done on https://github.com/PostgREST/postgrest/pull/1582.
- Fixes appending an erroneous `charset=uft-8` on binary output. Reported on [gitter chat](https://gitter.im/begriffs/postgrest?at=5e4ed61189f30b12651cad0f).


test/Feature/RpcSpec.hs
```diff
@@ -522,8 +523,8 @@ spec actualPgVersion =
               `shouldRespondWith`
                 [json|["hi", "there"]|]
 
-    it "returns first value for repeated params without VARIADIC" $
-      get "/rpc/sayhello?name=world&name=ignored"
+    it "returns last value for repeated params without VARIADIC" $
+      get "/rpc/sayhello?name=ignored&name=world"
```
steve-chavez, line 527: @wolfgangwalther I've changed this behavior. It would be a breaking change otherwise.

wolfgangwalther: Oh yes. It was my intention to not introduce a breaking change here - that's why I wrote the test. :+1:

src/PostgREST/ApiRequest.hs
```diff
@@ -70,6 +70,32 @@ data Target = TargetIdent QualifiedIdentifier
...
+jsonRpcParams proc prms =
+  if not $ pdHasVariadic proc then -- if proc has no variadic arg, save steps and directly convert to json
+    ProcessedJSON (JSON.encode $ M.fromList $ second JSON.toJSON <$> prms) (S.fromList $ fst <$> prms)
+  else
+    let paramsMap = M.fromListWith mergeParams $ toRpcParamValue proc <$> prms in
+    ProcessedJSON (JSON.encode paramsMap) (S.fromList $ M.keys paramsMap)
+  where
+    mergeParams :: RpcParamValue -> RpcParamValue -> RpcParamValue
+    mergeParams (Variadic a) (Variadic b) = Variadic $ b ++ a
+    mergeParams _ v                       = v -- repeated params for non-variadic arguments are not merged
```
wolfgangwalther, line 97: I think the order of this needs to be changed to `v _`. This is because you changed the test-case from first to last repetition for non-repeated args.

However, since you're change with `hasVariadic` this code path is not taken anymore. So we actually need another test case that has a function with regular argument + variadic argument, where the non-variadic one is repeated as well. And then this part kicks in and should show the same behaviour as the regular case.

src/PostgREST/ApiRequest.hs
```diff
@@ -70,6 +70,32 @@ data Target = TargetIdent QualifiedIdentifier
...
+toRpcParamValue :: ProcDescription -> (Text, Text) -> (Text, RpcParamValue)
+toRpcParamValue proc (k, v) | argIsVariadic k = (k, Variadic [v])
+                            | otherwise       = (k, Fixed v)
+  where
+    argIsVariadic arg = isJust $ find (\PgArg{pgaName, pgaVar} -> pgaName == arg && pgaVar) $ pdArgs proc
+
+-- | Convert rpc params `/rpc/func?a=val1&b=val2` to json `{"a": "val1", "b": "val2"}
+jsonRpcParams :: ProcDescription -> [(Text, Text)] -> PayloadJSON
+jsonRpcParams proc prms =
+  if not $ pdHasVariadic proc then -- if proc has no variadic arg, save steps and directly convert to json
```
wolfgangwalther, line 89: I'm curious - did you test the performance difference of this? How many req/s does the variadic parsing take away?

steve-chavez: You know what, the perf difference is actually negligible.

I did a load test before and I remember getting around `100 req/s` less but I must have made a mistake. I was calling an `add_them` func with 10 args.

It's still a nice refactor I guess :smile:

steve-chavez: Btw, I get `~1990 req/s` for an `/rpc/add_them?a=1&b=2&c=3&d=4&e=5`.

src/PostgREST/ApiRequest.hs
steve-chavez: @wolfgangwalther Hmm.. could you help me with that on another PR?

I mostly wanted to isolate the bits for non-variadic functions here. If the variadic part behaves differently it wouldn't be a breaking change for now since we haven't been supporting variadic.

wolfgangwalther: Sure, see #1639.

src/PostgREST/ApiRequest.hs
wolfgangwalther: With the code you have in this PR, you could compare the following 3 functions/requests:

```
-- call with: GET /rpc/add_them1?a=1&b=2&c=3&d=4&e=5
-- will take the "if not pdHasVariadic" path
CREATE FUNCTION add_them1(a int, b int, c int, d int, e int) RETURNS int
LANGUAGE SQL AS $$
  SELECT a+b+c+d+e
$$;

-- call with: GET /rpc/add_them2?a=1&b=2&c=3&d=4&e=5
-- will take the "else" path - but is actually not a variadic function (the variadic arg will be ignored, it has a default value)
CREATE FUNCTION add_them2(a int, b int, c int, d int, e int, VARIADIC v int[] DEFAULT '{}') RETURNS int
LANGUAGE SQL AS $$
  SELECT a+b+c+d+e
$$;

-- call with: GET /rpc/add_them3?v=1&v=2&v=3&v=4&v=5
-- will take the "else" path - and is really variadic
CREATE FUNCTION add_them3(VARIADIC v int[]) RETURNS int
LANGUAGE SQL AS $$
  -- no SUM or anything on purpose to be as much comparable to the other funcs as possible
  SELECT v[1]+v[2]+v[3]+v[4]+v[5]
$$;
```

This should give us a really good idea whether the variadic branch is in fact slower or not.

wolfgangwalther: > It's still a nice refactor I guess smile

With most of the refactor I absolutely agree, yes! But if the variadic branch was not really slower, then this branching (and adding the `hasVariadic` stuff) would just increase the complexity of the code.

steve-chavez: @wolfgangwalther Great idea for the tests!

> GET /rpc/add_them2?a=1&b=2&c=3&d=4&e=5

**addthem2** is not working (need to add `v=5`). Should it work with https://github.com/PostgREST/postgrest/pull/1639?

addthem 1 and 3 do work, these are my results:

## addthem1

(2004 + 2011 + 1983 + 1995 + 1994 + 1996 + 2004)/7 = 1998

## addthem3

(1950 + 1942 + 1959 + 1928 + 1939 + 1923 + 1955)/7 = 1942

So looks like the `non-variadic` path is indeed a bit faster. `~56 req/s` more.

steve-chavez: @wolfgangwalther WDYT? Should we keep the non-variadic branch?

wolfgangwalther: > @wolfgangwalther WDYT? Should we keep the non-variadic branch?

Well... we gotta get the `addthem2` working, I guess. That one will really tell us where the difference comes from (although it is very likely that the difference is indeed in the haskell, like you suggest). If the difference is there - of course keep the non-variadic branch. Non-variadic is going to be used so much more often, I think it's worth it.

> **addthem2** is not working (need to add `v=5`). Should it work with #1639?

No, I think we found a bug then. We do have a test-case for the n=0 case, but only without other arguments. Maybe there is a difference. I will try to reproduce this and report back.

In the meantime you can compare the following two functions against each other - that should work around the limitation, but yield comparable results:

```sql
-- call with: GET /rpc/add_them1x?a=1&b=2&c=3&d=4&e=5&x=0
-- will take the "if not pdHasVariadic" path
CREATE FUNCTION add_them1x(a int, b int, c int, d int, e int, x int) RETURNS int
LANGUAGE SQL AS $$
  SELECT a+b+c+d+e
$$;

-- call with: GET /rpc/add_them2x?a=1&b=2&c=3&d=4&e=5&x=0
-- will take the "else" path - but is actually not a variadic function (the variadic arg will be ignored, it has a default value)
CREATE FUNCTION add_them2x(a int, b int, c int, d int, e int, VARIADIC x int[]) RETURNS int
LANGUAGE SQL AS $$
  SELECT a+b+c+d+e
$$;
```

src/PostgREST/Types.hs
```diff
@@ -45,6 +49,7 @@ toMime CTTextCSV         = "text/csv"
 toMime CTTextPlain       = "text/plain"
 toMime CTOpenAPI         = "application/openapi+json"
 toMime CTSingularJSON    = "application/vnd.pgrst.object+json"
+toMime CTUrlEncoded      = "application/x-www-form-urlencoded"
```
wolfgangwalther, line 52: Code coverage drops because of this line. Is this ever going to be used? This would only be needed to **return** this mimetype, right? I doubt that will ever be the case.

BTW, that's the same case for the `CTAny` line below.

If our goal was to increase code coverage (not only technically but also for a nice number :D), this would be some simple things to do.

steve-chavez: Hmm.. yeah, it's unused.

How would I go about that? Just delete the lines? Not sure if the compiler would like that..

wolfgangwalther: Yeah, you're right. I didn't think that to the end. But for the code coverage stuff to be useful, we'll have to find a way to "remove" those false alarms, otherwise it's a pain to check for the important stuff.

It looks like the tool for the job are [hpc "overlays"](http://wiki.haskell.org/Haskell_program_coverage#Example). I will dig into this further once I find a bit of time and see what we can do.

src/PostgREST/ApiRequest.hs
steve-chavez: Made new 7 load tests with addthem2:

### addthem2

(1923 + 1908 + 1897 + 1892 + 1911 + 1915 + 1904)/7 = 1907

Seems to bit the slowest code path(not sure).

So the non-variadic path remains to be the fastest.

steve-chavez: > Filed a bug report here: https://www.postgresql.org/message-id/5a80bcdb-e92a-5c21-d923-f3813543771b@technowledgy.de

Nice! I see you already got some replies.

So if it's a pg bug, would there be anything to add on this PR?

wolfgangwalther: > Seems to bit the slowest code path(not sure why).

This might be because it's now 6 params instead of 5 that have to be parsed. So you can't really compare it to the first addthem1 you ran, but only to the 2nd modified example I gave, that has 6 params as well.

Anyway, I think it's fine to keep the non-variadic path.

> So if it's a pg bug, would there be anything to add on this PR?

Not sure whether they will agree to call it a bug or rather a "limitation". It's currently not possible to call the function in PG that way, so I think we are fine with where we are. If you want to use such a function you can always add an overload that does not have the variadic part - and then everything will work.

Nothing to do here for us right now.

steve-chavez: Got it. Then I'll merge!

src/PostgREST/ApiRequest.hs
steve-chavez: Thanks a lot Wolfgang :)

PR #1640, opened 2020-11-02
run stylish-haskell v0.12.1.0

wolfgangwalther:
I updated stylish-haskell again (after it was broken in #1616 at first) and now I only get a small diff, that doesn't look too bad.

Couldn't get nix-shell to run yet, so I'm not sure whether the older stylish-haskell will revert this change... Let's see whether CI complains!


wolfgangwalther:
Now the only problem is circleci not doing anything. https://status.circleci.com says it's having problems to trigger.

wolfgangwalther:
So it seems like this change makes both older and newer versions of stylish haskell happy!

PR #1654, opened 2020-11-18
Fix rpc return part1

### LINKED ISSUE #1469: Incorrect behavior of header `Prefer: params=single-object` on overloaded function

wolfgangwalther:
So #1615 is rather lengthy. Most of that stuff you already approved, @steve-chavez.

So this is the first 4 commits of that branch, so we can merge them already. Had quite a few merge conflicts when rebasing on master after development has progressed there, and I don't need to repeat that when possible ;)

Resolves #1469.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -153,6 +153,7 @@ decodeProcs =
     parseArgs :: Text -> [PgArg]
     parseArgs = mapMaybe parseArg . filter (not . isPrefixOf "OUT" . toS) . map strip . split (==',')
 
+    -- TODO: does parseArg properly handle unnamed "character varying" arguments or arguments with spaces in their names?
     parseArg :: Text -> Maybe PgArg
     parseArg arg =
       let isVariadic = isPrefixOf "VARIADIC " $ toS arg
@@ -161,7 +162,9 @@ decodeProcs =
           (body, def) = breakOn " DEFAULT " argNoMode
           (name, typ) = breakOn " " body in
       if T.null typ
-         then Nothing
+         -- Handle unnamed args. TODO: refactor to types
+         then Just $
+           PgArg mempty (strip name) (T.null def) isVariadic
          else Just $
            PgArg (dropAround (== '"') name) (strip typ) (T.null def) isVariadic
 
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -140,7 +140,7 @@ decodeProcs =
                   <$> column HD.text
                   <*> column HD.text
                   <*> column HD.bool
-                  <*> column HD.char)
+                  <*> column HD.bool)
               <*> (parseVolatility <$> column HD.char)
               <*> pure False
 
@@ -165,18 +165,15 @@ decodeProcs =
          else Just $
            PgArg (dropAround (== '"') name) (strip typ) (T.null def) isVariadic
 
-    parseRetType :: Text -> Text -> Bool -> Char -> RetType
-    parseRetType schema name isSetOf typ
+    parseRetType :: Text -> Text -> Bool -> Bool -> RetType
+    parseRetType schema name isSetOf isComposite
       | isSetOf   = SetOf pgType
       | otherwise = Single pgType
       where
         qi = QualifiedIdentifier schema name
-        pgType = case typ of
-          'c' -> Composite qi
-          'p' -> if name == "record" -- Only pg pseudo type that is a row type is 'record'
-                   then Composite qi
-                   else Scalar qi
-          _   -> Scalar qi -- 'b'ase, 'd'omain, 'e'num, 'r'ange
+        pgType
+          | isComposite = Composite qi
+          | otherwise   = Scalar qi
 
     parseVolatility :: Char -> ProcVolatility
     parseVolatility v | v == 'i' = Immutable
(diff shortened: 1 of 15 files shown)


test/fixtures/schema.sql
```diff
@@ -925,6 +925,19 @@ CREATE FUNCTION setprojects(id_l int, id_h int, name text) RETURNS SETOF project
     update test.projects set name = $3 WHERE id >= $1 AND id <= $2 returning *;
 $_$;
 
+-- domains on tables are only supported from pg 11 on
+DO $do$BEGIN
+  IF (SELECT current_setting('server_version_num')::INT >= 110000) THEN
+      CREATE DOMAIN projects_domain AS projects;
```
steve-chavez, line 931: :O I didn't know this was possible. Do you use domains for tables in your projects? I guess that would allow adding additional check constraints to a table.

test/Feature/RpcSpec.hs
```diff
@@ -197,6 +197,16 @@ spec actualPgVersion =
           ]|]
           { matchHeaders = [matchContentTypeJson] }
 
+      when (actualPgVersion >= pgVersion110) $
+        it "can embed if rpc returns domain of table type" $ do
+          post "/rpc/getproject_domain?select=id,name,client:clients(id),tasks(id)"
+              [json| { "id": 1} |]
+            `shouldRespondWith`
+              [json|[{"id":1,"name":"Windows 7","client":{"id":1},"tasks":[{"id":1},{"id":2}]}]|]
+          get "/rpc/getproject_domain?id=1&select=id,name,client:clients(id),tasks(id)"
```
steve-chavez, line 206: These tests lead me to believe that it'd be possible to `GET /projects_domain`. But of course that doesn't work because `select * from projects_domain` also doesn't. I thought the DOMAIN could be used as an alias.

test/Feature/RpcSpec.hs
```diff
@@ -258,6 +268,12 @@ spec actualPgVersion =
       it "cannot return composite type in hidden schema" $
         post "/rpc/ret_point_3d" [json|{}|] `shouldRespondWith` 401
 
+      when (actualPgVersion >= pgVersion110) $
+        it "returns domain of composite type" $
+          post "/rpc/ret_composite_domain" [json|{}|] `shouldRespondWith`
+            [json|[{"x": 10, "y": 5}]|]
```
steve-chavez, line 274: Hmm.. but why should this return an array instead of an object? Shouldn't the output depend on the function `setof`?

If `setof` array, else object? This is the behavior I'm seeing on 7.0.1, with this change I see we always return an array.

src/PostgREST/DbStructure.hs
```diff
@@ -161,22 +162,20 @@ decodeProcs =
           (body, def) = breakOn " DEFAULT " argNoMode
           (name, typ) = breakOn " " body in
       if T.null typ
-         then Nothing
+         then Just $
+           PgArg mempty (strip typ) (T.null def) isVariadic
```
steve-chavez, line 166: Ah.. I see what you did here. A comment would be nice - the logic about the unnamed arg is not obvious by just looking at the code.

test/fixtures/schema.sql
wolfgangwalther: I haven't used those, I think this was just for completeness. Must have stumbled upon those while reading the docs. My use case so far is just domains of "explicit" composite types.

test/Feature/RpcSpec.hs
wolfgangwalther: I think the domain is just on top of the implicit composite type associated with every table. So it does not really have any connection to the table itself. What you could do with that, is e.g. adding additional constraints to the return value of a function returning a table type. But those constraints would not have any relevance for the actual data stored. Does that make sense?

test/Feature/RpcSpec.hs
wolfgangwalther: Absolutely, this should return an object. But this is fixed in part2 of #1615. In 7.0.1 only scalars are treated properly to not return an array. But composite types are not. The fix in here is just to treat domains of composites the same as composites.

So, 7.0.1 does the following:
- `RETURNS composite` returns array
- `RETURNS domain_of_composite` returns object

This fix:
- `RETURNS composite` returns array
- `RETURNS domain_of_composite` returns array

#1615:
- `RETURNS composite` returns object
- `RETURNS domain_of_composite` returns object

src/PostgREST/DbStructure.hs
wolfgangwalther: Hm.... could you explain to me what I did there? It's been a while :D.

So this is the `T.null typ` branch, that means the argument is unnamed and the type is actually in `name` per the `breakOn` above. That means it actually has to be `PgArg mempty (strip name) ...`, right?

test/Feature/RpcSpec.hs
steve-chavez: Got it! Just checked 7.0.1 and that holds. Then this new test is good :+1:.

src/PostgREST/DbStructure.hs
```diff
@@ -195,22 +194,46 @@ accessibleProcs = H.Statement (toS sql) (param HE.text) decodeProcs True
 
 procsSqlQuery :: SqlQuery
 procsSqlQuery = [q|
+  WITH RECURSIVE
+  rec_types AS (
+    SELECT
+      oid,
+      typbasetype,
+      COALESCE(NULLIF(typbasetype, 0), oid) AS base
+    FROM pg_type
```
steve-chavez, line 203: Did you checked this change against the `big schema` and see if there is a downgrade in perf?

From the looks of it, it doesn't seem so. But double checking would be good.

steve-chavez: Just ran the new query on the big schema and it's about `7 ms` slower than the previous query. So the perf drop is negligible, it should be fine.

wolfgangwalther: You guessed right:

```
query  | plan | execution | rows 
master | 0.37 | 6.3       | 3152
#1654  | 0.56 | 7.9       | 3152
```

It's a tiny bit slower. That makes sense, because the query is a bit more complex. But the 1.5ms here is nothing against the source columns query.

src/PostgREST/DbStructure.hs
steve-chavez: Oh, the only thing I noticed here was the lack of a name in the PgArg, so unnamed arg. I was actually hoping you clarify the other parts with the comment :D.

steve-chavez: I do think this logic can be improved, but for now a simple comment would help:
```suggestion
-- Handle unnamed args. TODO: refactor to types
         then Just $
           PgArg mempty (strip typ) (T.null def) isVariadic
```

src/PostgREST/DbStructure.hs
steve-chavez: Cool! How about adding a comment that this part is related to solving domains of domains:
```suggestion
 -- Recursively get the base types of domains
  WITH RECURSIVE
  rec_types AS (
    SELECT
      oid,
      typbasetype,
      COALESCE(NULLIF(typbasetype, 0), oid) AS base
    FROM pg_type
```

test/Feature/RpcSpec.hs
steve-chavez: Yes, makes sense. I've not had this use case, but it's great that we can support it transparently.

wolfgangwalther:
Applied the changes as discussed into the relevant commits.

PR #1658, opened 2020-11-20
Split StructureSpec into OpenApiSpec and OptionsSpec

wolfgangwalther:
I went looking for the OpenApi tests, because I made a small change but it didn't break any tests. Didn't find them at first. Felt like `StructureSpec` could very well be split into `OpenApiSpec` and `OptionsSpec`. Makes the OPTIONS tests a bit more visible as well.

Didn't want to change anything except splitting it up, to make it easier to review with that big of a diff.


steve-chavez:
> I went looking for the OpenApi tests. Didn't find them at first

Something I've experimented as well. Nice change :)

Issue #1656, opened 2020-11-20
Add explicit Prefer return=headers


wolfgangwalther:
Currently, a POST request returns either:
- both a body and a location header, with `return=representation`
- no body and no location header, with `return=minimal`
- only a location header, without Prefer `return`

It's not possible to explicitly make a request for `return=headers`.

My use-case is that I default to `return=representation` in my library, but want the user to be able to override this to only return the header. Allowing something like `return=headers` explicitly is much easier to handle and understand.

I think this would need to be extended here:

https://github.com/PostgREST/postgrest/blob/8618ffa5fc51ba43c8998f1acda013cc05ff5dc0/src/PostgREST/Types.hs#L87-L90

Is it possible to add some kind of alias here, so that `'return=headers'` and `mempty` both result in `HeadersOnly`?

steve-chavez:
> My use-case is that I default to return=representation in my library, but want the user to be able to override this to only return the header. Allowing something like return=headers explicitly is much easier to handle and understand.

I agree. With this logic we could say that PATCH/DELETE default to `return=minimal` while `POST` defaults to `return=headers-only`.

steve-chavez:
I wonder if we should change the POST default to `return=minimal`.

I've noticed that when newcomers are learning about postgres privileges(RLS, grants), they always have to jump through this hoop. It's always a head-scratcher, and makes postgrest not transparent to PostgreSQL.

I think most users are using `Prefer: return=representation` instead of the `Location` header anyway. Should we change the default and perhaps offer a config for backwards compatibility?

wolfgangwalther:
> I wonder if we should change the POST default to `return=minimal`.
> 
> I've noticed that when newcomers are learning about postgres privileges(RLS, grants), they always have to jump through this hoop. It's always a head-scratcher, and makes postgrest not transparent to PostgreSQL.

I can see your point from the SQL perspective. However from the HTTP perspective, we should definitely default to returning a `Location` header for `POST`, given that the RFC asks for it: https://tools.ietf.org/html/rfc7231#section-4.3.3

> I think most users are using `Prefer: return=representation` instead of the `Location` header anyway. Should we change the default and perhaps offer a config for backwards compatibility?

If we were to change the default, I'd suggest to just change and document it - but I don't think we'd need a config option.

steve-chavez:
> given that the RFC asks for it: https://tools.ietf.org/html/rfc7231#section-4.3.3

[...]

Issue #740, opened 2016-11-13
Postgrest does not respond with a Preference-Applied header


jackfirth:
HTTP Preferences are intended to be optional requests by clients that servers can choose to ignore at whim. Because of this, if a server chooses to honor a preference it may respond with a `Preference-Applied` header listing the preferences the server applied to the request. Postgrest does not return this header, meaning clients cannot tell immediately what preferences were applied without analyzing the response body.

This behavior is documented by [section 3 of RFC 7240](https://tools.ietf.org/html/rfc7240#section-3). Inclusion of the header is optional, but the RFC specifically points out this is a good idea for the "return=representation" preference since it can be difficult for clients to determine whether this was applied or not.

ruslantalpa:
As things stand now there is not a single case of preference header that postgrest ignores so there is no practical reason for adding this response besides "respecting RFC". When a client sends a header, it's 100% applied, no need to look at the body.
Unless there is a specific problem this solves, it's just additional bytes in the response

begriffs:
True, the RFC states

> Use of the Preference-Applied header is only necessary when it is not
   readily and obviously apparent that a server applied a given
   preference and such ambiguity might have an impact on the client's
   handling of the response.

The question is, is it readily apparent which Prefer headers are applied? As Ruslan pointed out, we always apply any Prefer header that we understand. Thus it's generally clear. There are two situations I can see where it might be unclear, but may not be worth the extra overhead on every response to address them:

1. Detecting when new versions of PostgREST drop support for Prefer headers. For instance `Prefer: count=none` is now default behavior so the server no longer "applies" that header. However if this is a confusion we could alternately return an error to the client and that would be an even more effective (albeit brash) way to let them know about use of an unsupported header.
2. Debugging requests. Sometimes people don't really know what they are sending, or a proxy strips out request information. Preference-Applied would reliably indicate which options made it to the server. However an alternative way to handle debugging is to allow verbose server logging, assuming the person debugging the client is also involved with the API server.

jackfirth:
@ruslantalpa: You're correct that if a client *assumes* it's speaking to a Postgrest server (more specifically, a Postgrest server of a particular known version), than the Preference-Applied header isn't useful. Preferences are meant to be general enough that a client can use them without knowing details about the specific server requests are being sent to. When using a standard preference like "return=representation" which many different servers may implement special handling for, I think it's worthwhile to indicate that the preference was applied so clients without Postgrest-specific knowledge can take advantage.

[...]

PR #1659, opened 2020-11-20
Add support for Prefer tx=rollback

### LINKED ISSUE #1598: Suggestion: Rollback transaction by setting HTTP header
I'm doing some automated testing in my projects, where I send requests through PostgREST to my database layer. Some of them are `POST`, `PATCH` and `DELETE` requests. It would be cool, if I was able to set a http header for those requests to make them rollback the transaction after execution. So the requests would execute just the same and also return the same response - but only at the end the transaction is rolled back. This would make it easier to write tests that depend on a clean database and are executed multiple times.

This could also make some of the postgrest internal unit tests easier to write. I think quite a few of them "depend" on each other in a sense that each test must be careful what PK values to insert into tables, because those might already be used by other tests.

I couldn't find any standard header that would be suitable for this, so it would probably be something like `X-PGRST-TX: commit|rollback`, where `commit` would be the default if the header was omitted?

Thoughts?

wolfgangwalther:
Resolves #1598.

In that issue we discussed `Prefer: rollback` vs. `Prefer: tx=rollback`. I implemented the latter and also added `Prefer: tx=commit` (the default) for completeness (see #1656 as well).

---

This would also allow adding another config option to use `rollback` as the default, to be overwritten with `Prefer: tx=commit`. This would be something that I would run the test environment in my projects with, because those automated tests should almost never change the database for real.

Now that I write this, I wonder whether we should change the config option `allow-rollback` to something like `allow-prefer-tx` now. This would allow us to allow/deny the use of `tx=commit`, when the default is `rollback`.

The combo of `tx-default = rollback` and `allow-prefer-tx = false` could be a cool thing to setup a demo postgrest server, where everyone can make requests against a test schema immediately, maybe even from a web-interface?

Would `rollback-by-default = true | false` or `tx-default = commit | rollback` be better? Any other ideas?

---

Back to the current PR. `Prefer: tx=rollback` is only applied for POST, PUT, PATCH and DELETE requests and only when those do not return with an error. Not for GET and HEAD. The `Preference-Applied` header is set accordingly. When a request returns an error, the transaction is rolled back anyway, the error is returned regularly and the `Preference-Applied` header is not set.

TODO:

- [x] Add `Preference-Applied` header for explicit `tx=commit`.
- [x] Rename `allow-rollback` to `allow-prefer-tx` or something better
- [x] Add test for `ActionInvoke`
- [x] Add `rollback-by-default` config option

And then for another PR: Make `rollback` the default for our spec-tests. This will require quite a bit of cleanup. My goal would be to be able to run the whole test suite multiple times against the same database without recreating the schema.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -86,7 +86,15 @@ postgrest logLev refConf refDbStructure pool getTime connWorker =
                 Right claims -> do
                   let
                     authed = containsRole claims
-                    handleReq = runPgLocals conf claims (app dbStructure conf) apiRequest
+                    shouldCommit   = configTxAllowOverride conf && iPreferTransaction apiRequest == Just Commit
+                    shouldRollback = configTxAllowOverride conf && iPreferTransaction apiRequest == Just Rollback
+                    preferenceApplied
+                      | shouldCommit    = addHeadersIfNotIncluded [(hPreferenceApplied, BS.pack (show Commit))]
+                      | shouldRollback  = addHeadersIfNotIncluded [(hPreferenceApplied, BS.pack (show Rollback))]
+                      | otherwise       = identity
+                    handleReq = do
+                      when (shouldRollback || (configTxRollbackAll conf && not shouldCommit)) HT.condemn
+                      mapResponseHeaders preferenceApplied <$> runPgLocals conf claims (app dbStructure conf) apiRequest
                   dbResp <- P.use pool $ HT.transaction HT.ReadCommitted (txMode apiRequest) handleReq
                   return $ either (errorResponseFor . PgError authed) identity dbResp
         -- Launch the connWorker when the connection is down. The postgrest function can respond successfully(with a stale schema cache) before the connWorker is done.
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -97,6 +97,9 @@ data AppConfig = AppConfig {
   , configJWKS              :: Maybe JWKSet
 
   , configLogLevel          :: LogLevel
+
+  , configTxRollbackAll     :: Bool
+  , configTxAllowOverride   :: Bool
   }
 
 configPoolTimeout' :: (Fractional a) => AppConfig -> a
@@ -196,6 +199,15 @@ readPathShowHelp = customExecParser parserPrefs opts
           |
           |## logging level, the admitted values are: crit, error, warn and info.
           |# log-level = "error"
+          |
+          |## rollback all transactions by default, use for test environments
+          |## disabled by default
+          |# tx-rollback-all = false
+          |
+          |## allow overriding the tx-rollback-all setting for a request by
+          |## setting the Prefer: tx=[commit|rollback] header
+          |## disabled by default
+          |# tx-allow-override = false
           |]
 
 -- | Parse the config file
@@ -225,9 +237,9 @@ readAppConfig cfgPath = do
         <*> (fmap unpack <$> optString "server-unix-socket")
         <*> parseSocketFileMode "server-unix-socket-mode"
         <*> (fromMaybe "pgrst" <$> optString "db-channel")
-        <*> ((Just True ==) <$> optBool "db-channel-enabled")
+        <*> (fromMaybe False <$> optBool "db-channel-enabled")
         <*> (fmap encodeUtf8 <$> optString "jwt-secret")
-        <*> ((Just True ==) <$> optBool "secret-is-base64")
+        <*> (fromMaybe False <$> optBool "secret-is-base64")
         <*> parseJwtAudience "jwt-aud"
         <*> (fromMaybe 10 <$> optInt "db-pool")
         <*> (fromMaybe 10 <$> optInt "db-pool-timeout")
@@ -240,6 +252,8 @@ readAppConfig cfgPath = do
         <*> (maybe [] (fmap encodeUtf8 . splitOnCommas) <$> optValue "raw-media-types")
         <*> pure Nothing
         <*> parseLogLevel "log-level"
+        <*> (fromMaybe False <$> optBool "tx-rollback-all")
+        <*> (fromMaybe False <$> optBool "tx-allow-override")
 
     parseSocketFileMode :: C.Key -> C.Parser C.Config (Either Text FileMode)
     parseSocketFileMode k =
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1559, No downtime when reloading the schema cache with SIGUSR1 - @steve-chavez
  - #504, Add `log-level` config option. The admitted levels are: crit, error, warn and info - @steve-chavez
  - #1607, Enable embedding through multiple views recursively - @wolfgangwalther
+ - #1598, Allow rollback of the transaction with Prefer tx=rollback - @wolfgangwalther
 
 ### Fixed
  
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -154,8 +154,10 @@ test-suite spec
                       Feature.DeleteSpec
                       Feature.EmbedDisambiguationSpec
                       Feature.ExtraSearchPathSpec
+                      Feature.HtmlRawOutputSpec
                       Feature.InsertSpec
                       Feature.JsonOperatorSpec
+                      Feature.MultipleSchemaSpec
                       Feature.NoJwtSpec
                       Feature.NonexistentSchemaSpec
                       Feature.OpenApiSpec
@@ -164,6 +166,8 @@ test-suite spec
                       Feature.QueryLimitedSpec
                       Feature.QuerySpec
                       Feature.RangeSpec
+                      Feature.RawOutputTypesSpec
+                      Feature.RollbackSpec
                       Feature.RootSpec
                       Feature.RpcPreRequestGucsSpec
                       Feature.RpcSpec
@@ -171,9 +175,6 @@ test-suite spec
                       Feature.UnicodeSpec
                       Feature.UpdateSpec
                       Feature.UpsertSpec
-                      Feature.RawOutputTypesSpec
-                      Feature.HtmlRawOutputSpec
-                      Feature.MultipleSchemaSpec
                       SpecHelper
                       TestTypes
   hs-source-dirs:     test
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -114,6 +114,7 @@ data ApiRequest = ApiRequest {
   , iPreferParameters     :: Maybe PreferParameters           -- ^ How to pass parameters to a stored procedure
   , iPreferCount          :: Maybe PreferCount                -- ^ Whether the client wants a result count
   , iPreferResolution     :: Maybe PreferResolution           -- ^ Whether the client wants to UPSERT or ignore records on PK conflict
+  , iPreferTransaction    :: Maybe PreferTransaction          -- ^ Whether the clients wants to commit or rollback the transaction
   , iFilters              :: [(Text, Text)]                   -- ^ Filters on the result ("id", "eq.10")
   , iLogic                :: [(Text, Text)]                   -- ^ &and and &or parameters used for complex boolean logic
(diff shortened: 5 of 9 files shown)


steve-chavez:
Oh.. wait. I remember the rollback logic being simple. On:

https://github.com/PostgREST/postgrest/blob/944efb3a6ab910711d9151c6241a48111762e1bd/src/PostgREST/App.hs#L89-L91

Do:
```haskell
handleReq =  HT.condemn >> runPgLocals conf claims (app dbStructure proc cols conf) apiRequest
```

I remember getting almost all the tests green, besides the ones that depended on each other. It should work.

---

> My goal would be to be able to run the whole test suite multiple times against the same database without recreating the schema.

Are you using Nix? This is something I already do with `postgrest-test-spec`. You can even switch versions or run the tests on all of them.

wolfgangwalther:
> Oh.. wait. I remember the rollback logic being simple. 

Thanks. I was going to ask about that, the current implementation felt a bit repetitive :D

> Are you using Nix? This is something I already do with `postgrest-test-spec`. You can even switch versions or run the tests on all of them.

Yes, I do. Since yesterday :). I was a bit short with the explanation here or maybe my logic was kind of backwards. With `postgrest-test-spec` (and also `make test` which I used before), the database is recreated on each run. That works fine, of course.

Let me start somewhere else: When I did #1658 I felt like sorting the following lines alphabetically while trying to figure out where to insert my new lines:

https://github.com/PostgREST/postgrest/blob/944efb3a6ab910711d9151c6241a48111762e1bd/test/Main.hs#L103-L114

However, when I do that - suddenly some tests turn red. This is because some of those tests depend on each other across specs. That's bad.

Getting the tests to run without recreating the database would prove to a certain degree, that this coupling across tests is removed.

Then I noticed:

https://github.com/PostgREST/postgrest/blob/944efb3a6ab910711d9151c6241a48111762e1bd/test/Main.hs#L125

So maybe it could help with performance while running the tests as well.

wolfgangwalther:
So I did that and learned a bit about Monads as well. Also did all the other TODOs and simplified the tests a alot. I think they read really nicely now.

src/PostgREST/Config.hs
```diff
@@ -196,6 +198,14 @@ readPathShowHelp = customExecParser parserPrefs opts
           |
           |## logging level, the admitted values are: crit, error, warn and info.
           |# log-level = "error"
+          |
+          |## rollback all transactions by default, use for test environments
+          |## disabled by default
+          |# transaction-rollback = false
```
steve-chavez, line 204: How about having this one only for internal use?

It would only be declared on the `Config` type. As an example, we used to have this setting:

https://github.com/PostgREST/postgrest/blob/96a16a377f8dab58a72cb5c6d7ddffee2316334e/src/PostgREST/Config.hs#L92

That one disabled logging and it was only set to True on the tests.

steve-chavez: Though I have a question here. Would the rollback only make sense in a global way? Or do we need it to be on demand?

Perhaps we just need one of these options.

wolfgangwalther: > How about having this one only for internal use?

But.. but... but... I want to use this in my project's dev environment as well. I run automated tests there. Some via `pgtap` directly in the database, but also some that make requests via `newman` through PostgREST. Those should not modify the database by default. I don't want to put `tx=rollback` on *every* request.

> Though I have a question here. Would the rollback only make sense in a global way? Or do we need it to be on demand?
> 
> Perhaps we just need one of these options.

But then I have some requests, that I need to commit. So I need default=rollback / override=commit for sure.

And I can very well imagine both the following scenarios as well:
- default=rollback / no overrides: Public PostgREST demo server to test requests against. Maybe even from a webinterface, linked to in the docs. Aka "read the docs, try it out immediately". That would be really cool.

- default=commit / override=rollback: Maybe I do want to make a request on a production server for diagnostic purposes. It must be rolled back, but all the other requests that other users make, should go through. I can even imagine using that somehow to validate requests before they are actually made. Think of a form on the client, where the user can enter stuff. You could even "test" the request in the background, before they submit, and use error messages to display validation hints. Only if the user hits submit, you submit the real request, though.

I am sure people will find more use-cases for this! :)

steve-chavez: Oh, cool cool :). The demo server idea sounds amazing!

Then I think what we should do is to group the options, by prefix. How about naming them like:

- rollback-all: all requests will be rolled back
- rollback-on-prefer: rollback on demand with Prefer

wolfgangwalther: I like grouping them, but I don't think `rollback-` is correct. Because on `rollback-all = true`, the `prefer` setting actually allows `prefer tx=commit`, not rollback.

Maybe:
- `prefer-tx-allowed` and
- `prefer-tx-default`

But this is not a 100%. This could be interpreted as `default` only kicking in when `allowed: true` or something like that.

What about just one config option `transactions = force-commit | commit | rollback | force-rollback`, where `force-commit` is the default? Better name for "transactions" maybe?

wolfgangwalther: `tx-prefer-allowed` and `tx-rollback-by-default` could also work.

steve-chavez: >  Because on rollback-all = true, the prefer setting actually allows prefer tx=commit, not rollback.

I see. Nice, that's a sensible behavior.

> tx-prefer-allowed and tx-rollback-all

I prefer this one. It's more obvious. The `transactions` option is not so straightforward.

`tx-prefer-allowed` seems related to all prefers though, but I don't have a better name for now. It should be enough to clarify it with the option description.

(Or maybe `tx-override-on-prefer`?)

wolfgangwalther: I went with `tx-rollback-all` and `tx-allow-override` and improved the help output about it a bit.

src/PostgREST/Config.hs
```diff
@@ -240,6 +252,8 @@ readAppConfig cfgPath = do
         <*> (maybe [] (fmap encodeUtf8 . splitOnCommas) <$> optValue "raw-media-types")
         <*> pure Nothing
         <*> parseLogLevel "log-level"
+        <*> ((Just False ==) <$> optBool "tx-rollback-all")
+        <*> ((Just False ==) <$> optBool "tx-allow-override")
```
steve-chavez, line 256: Both of these default to False right?

I remember this part being confusing because of Hlint. Check this: https://github.com/ndmitchell/hlint/issues/970#issuecomment-643646635

Maybe newer versions don't show this hint.

src/PostgREST/App.hs
```diff
@@ -86,7 +86,15 @@ postgrest logLev refConf refDbStructure pool getTime connWorker =
...
-                    handleReq = runPgLocals conf claims (app dbStructure conf) apiRequest
+                    shouldCommit   = configTxAllowOverride conf && iPreferTransaction apiRequest == Just Commit
+                    shouldRollback = configTxAllowOverride conf && iPreferTransaction apiRequest == Just Rollback
+                    preferenceApplied
+                      | shouldCommit    = addHeadersIfNotIncluded [(hPreferenceApplied, BS.pack (show Commit))]
+                      | shouldRollback  = addHeadersIfNotIncluded [(hPreferenceApplied, BS.pack (show Rollback))]
+                      | otherwise       = identity
+                    handleReq = do
+                      when (shouldRollback || (configTxRollbackAll conf && not shouldCommit)) HT.condemn
+                      mapResponseHeaders preferenceApplied <$> runPgLocals conf claims (app dbStructure conf) apiRequest
```
steve-chavez, line 97: TIL about [mapeResponseHeaders](https://hackage.haskell.org/package/wai-3.2.2.1/docs/Network-Wai.html#v:mapResponseHeaders). Maybe that can be used for the `Prefer: return` as well. That would solve https://github.com/PostgREST/postgrest/issues/740.

steve-chavez:
Looks good! Don't forget to add an entry to the CHANGELOG :)

src/PostgREST/Config.hs
wolfgangwalther: Uh, eh. I have no idea. This is not covered by the tests, I guess I'll need to add to the io-tests.

I just copied from a couple more lines above, I think and thought False must be False. But according to the link you're giving... I am completely lost now :D

steve-chavez: Actually when trying the `tx-rollback-all=true` while running the postgrest binary I don't get the `all ROLLBACK` behavior. When changing these to `(Just True ==)` I do get the behavior. Maybe this needs to be corrected?

wolfgangwalther: Hm. Ok. I had a brief look at the IO tests, but I don't think it's simple to add such a test there. What I would really like to have is a command line argument to call the postgrest binary with, which just exits and dumps the effectively loaded config. This could be useful for debugging and the io tests.

wolfgangwalther: Since this is **very** confusing, I added some **very** explicit comments on each of those boolean options to make that clear.

The IO test is something for later.

steve-chavez: Yes, **very** confusing. We should look at the latest hlint version and see if that still prevents us from using a `fromMaybe False`.

monacoremo: That might be very useful! Related topic if we add an exit command line option (or env variable): It would also be useful to have an option for postgrest to exit once the dbstrucutre is loaded. That would allow us to benchmark the startup time (including in CI) as a basis for refactorings/improvements

wolfgangwalther: Those should definitely be command line options and not environment variables.

Something like `--dump-config` could work. Is there anything useful that we could output for the dbstructure? What if `--dump-structure` would output a json with all the query results?

wolfgangwalther: > Yes, **very** confusing. We should look at the latest hlint version and see if that still prevents us from using a `fromMaybe False`.

I just changed all of them to a `fromMaybe False` and hlint is not complaining. Maybe this is fixed already with what we have now. Let's see what CI says.

monacoremo: Yes, that would be the relevant output. Will need to discuss in more detail at some other point I think, just wanted to flag that we might want to consider additional exit points if we add that command line option

steve-chavez:
Had an idea for saving effort on the CHANGELOG: https://github.com/PostgREST/postgrest-docs/issues/340#issuecomment-731863705

steve-chavez:
After this change I'm getting a drop in performance of about `100 req/s` on the [GETSingle](https://github.com/steve-chavez/postgrest-benchmark/blob/master/k6/GETSingle.js) load test:

Some runs with prepared statements without tx rollback logic:

- 2069.639507/s
- 2073.134588/s
- 2078.380637/s
- 2067.949048/s

With tx rollback logic:

- 1908.421302/s
- 1927.784673/s
- 1924.697279/s
- 1930.679238/s

https://github.com/PostgREST/postgrest/pull/1609 would have definitely helped in detecting this.

--- 

I'm not sure what part of the code is causing the regression(I thought `mapResponseHeaders`).

My idea was to shortcut this code path in case `tx-end=commit` is not modified, but no dice: https://github.com/steve-chavez/postgrest/commit/ea3f58218e98ef10cb7d99d685ced594aacc8777

Removing the logic does bring back the previous perf results: https://github.com/steve-chavez/postgrest/commit/18057c10c4f633518c1db4e802ee5e8c748f0c59. 

Not sure how to fix for now.

wolfgangwalther:
> **Edit**: Nevermind. I cannot reproduce anymore(I recreated the AWS VPC, not sure what went wrong). I now get around `15 req/s` less. which is not a big deal.

Puh. I will still try to have a go at using #1609 and see what I get there locally to confirm.

wolfgangwalther:
So I did a few test runs across 30 min with `postgrest-loadtest`. The commit with the `tx=rollback` change came it at 347 req/s and 350 reqs/. The commit immediately before that came in at 349 req/s. I couldn't see any drop in performance.

steve-chavez:
That's good. The tests I did before were 1 min only. In general I've found that when there's around 100req/s less on short tests then usually that means a drop in perf. This case was weird for me, the load tests on a newly created AWS VPC tended to be somewhat reliable.

dsyrstad:
This should be added to the doc. I just spent 30 minutes searching for something like this and ran across this PR.

PR #1662, opened 2020-11-21
Update README.md

yvmarques:
Added "Since" for the security row level PostgreSQL has it since 9.5


steve-chavez:
Thanks for clarifying!

PR #1663, opened 2020-11-21
Make transaction-rollback=true the default for test-suite

wolfgangwalther:
Follow-up on #1659:

> And then for another PR: Make rollback the default for our spec-tests. This will require quite a bit of cleanup. My goal would be to be able to run the whole test suite multiple times against the same database without recreating the schema.

This is based on the other PR, so the diff is a bit longer than it should be. Still, it's huge. I hope it looks better with "Hide whitespace changes" turned on.

Quick guide on how to look at this:
- Everything in `src/...` is from the other PR. Ignore here.
- Changes have been made to `test/Main.hs` and `test/SpecHelper.hs`. Removed the schema reloading, re-ordered the tests and enabled `transaction-rollback = true`. This caused a lot of broken tests.
- Everything in `test/Feature/...` + `test/fixtures/...` is fixing those tests, so that they don't depend on the result of other tests anymore. As always, every test-case I touched, I also reformatted / refactored. This might increase the diff here, but I hope to be able to make all those test-cases consistently structured down the road. This helps me a lot to understand the tests quickly when reading them. The changes to the tests, that I had to make to account for the rolled back transactions are often quite similar. For every kind of change, I will comment on an example.

When refactoring / reformatting the tests, I did the following:
- When possible, replaced code like this:
```haskell
p <- request ...
liftIO $do
  simpleBody p `shouldBe` ...
  simpleStatus p `shouldBe` ...
... lookup hLocation $ simpleHeaders p
...
```
with:
```haskell
request ...
  `shouldRespondWith`
    <body>
    { matchStatus = ...
    , matchHeaders = ... }
```

Much easier to read, because it follows the same structure that almost all the tests have.

- When possible replaced `request methodPost ...` with the short form `post ...` (for all http methods). Fewer empty arrays and empty bodies that way.

- Formatted the requests around `shouldRespondWith` with consistent indentation and newlines, like this:
```haskell
request <method> <uri>
    <headers>
    <body>
  `shouldRespondWith`
    <body>
    <status/headers>
```

For me, it's much easier to immediately see what's going on, compared to when the `shouldRespondWith` is sometimes hidden somewhere on the same line as other things.

- Changed `[str|...json data...|]` to `[json|...json data...|]`. This is essential, because `shouldRespondWith` will then do a json-comparison and ignore formatting, which sometimes causes trouble. It does also include a check for the `application/json` content-type by default. I don't think there were any changes of this, in this PR.

- Removed obsolete `matchContentTypeJson` (see last point why, it's included already) and `matchStatus = 200` (the default check).

@steve-chavez do you think this formatting / structure for tests is something that we can "formalize" a bit more, even without a linter / style-check that covers for it? Any suggestions for improvement?

### DIFF
--- a/test/Feature/AndOrParamsSpec.hs
+++ b/test/Feature/AndOrParamsSpec.hs
@@ -33,13 +33,24 @@ spec actualPgVersion =
               {"id": 1, "child_entities": [ { "id": 1 }, { "id": 2 } ] }, { "id": 2, "child_entities": []},
               {"id": 3, "child_entities": []}, {"id": 4, "child_entities": []}
             ]|] { matchHeaders = [matchContentTypeJson] }
+
         it "can do logic on the third level" $
-          get "/entities?child_entities.grandchild_entities.or=(id.eq.1,id.eq.2)&select=id,child_entities(id,grandchild_entities(id))" `shouldRespondWith`
-            [json|[
-              {"id": 1, "child_entities": [ { "id": 1, "grandchild_entities": [ { "id": 1 }, { "id": 2 } ]}, { "id": 2, "grandchild_entities": []}]},
-              {"id": 2, "child_entities": [ { "id": 3, "grandchild_entities": []} ]},
-              {"id": 3, "child_entities": []}, {"id": 4, "child_entities": []}
-            ]|] { matchHeaders = [matchContentTypeJson] }
+          get "/entities?child_entities.grandchild_entities.or=(id.eq.1,id.eq.2)&select=id,child_entities(id,grandchild_entities(id))"
+            `shouldRespondWith`
+              [json|[
+                {"id": 1, "child_entities": [
+                  { "id": 1, "grandchild_entities": [ { "id": 1 }, { "id": 2 } ]},
+                  { "id": 2, "grandchild_entities": []},
+                  { "id": 4, "grandchild_entities": []},
+                  { "id": 5, "grandchild_entities": []}
+                ]},
+                {"id": 2, "child_entities": [
+                  { "id": 3, "grandchild_entities": []},
+                  { "id": 6, "grandchild_entities": []}
+                ]},
+                {"id": 3, "child_entities": []},
+                {"id": 4, "child_entities": []}
+              ]|]
 
       context "and/or params combined" $ do
         it "can be nested inside the same expression" $
@@ -210,12 +221,15 @@ spec actualPgVersion =
     context "used with POST" $
       it "includes related data with filters" $
         request methodPost "/child_entities?select=id,entities(id)&entities.or=(id.eq.2,id.eq.3)&entities.order=id"
-          [("Prefer", "return=representation")]
-          [json|[{"id":4,"name":"entity 4","parent_id":1},
-                 {"id":5,"name":"entity 5","parent_id":2},
-                 {"id":6,"name":"entity 6","parent_id":3}]|] `shouldRespondWith`
-          [json|[{"id": 4, "entities":null}, {"id": 5, "entities": {"id": 2}}, {"id": 6, "entities": {"id": 3}}]|]
-          { matchStatus = 201, matchHeaders = [matchContentTypeJson] }
+            [("Prefer", "return=representation")]
+            [json|[
+              {"id":7,"name":"entity 4","parent_id":1},
+              {"id":8,"name":"entity 5","parent_id":2},
+              {"id":9,"name":"entity 6","parent_id":3}
+            ]|]
+          `shouldRespondWith`
+            [json|[{"id": 7, "entities":null}, {"id": 8, "entities": {"id": 2}}, {"id": 9, "entities": {"id": 3}}]|]
+            { matchStatus = 201 }
 
     context "used with PATCH" $
       it "succeeds when using and/or params" $
@@ -228,9 +242,10 @@ spec actualPgVersion =
     context "used with DELETE" $
       it "succeeds when using and/or params" $
         request methodDelete "/grandchild_entities?or=(id.eq.1,id.eq.2)&select=id,name"
-          [("Prefer", "return=representation")] "" `shouldRespondWith`
-          [json|[{ "id": 1, "name" : "updated grandchild entity"},{ "id": 2, "name" : "updated grandchild entity"}]|]
-          { matchHeaders = [matchContentTypeJson] }
+            [("Prefer", "return=representation")]
+            ""
+          `shouldRespondWith`
+            [json|[{ "id": 1, "name" : "grandchild entity 1" },{ "id": 2, "name" : "grandchild entity 2" }]|]
 
     it "can query columns that begin with and/or reserved words" $
       get "/grandchild_entities?or=(and_starting_col.eq.smth, or_starting_col.eq.smth)" `shouldRespondWith` 200
--- a/test/Feature/DeleteSpec.hs
+++ b/test/Feature/DeleteSpec.hs
@@ -57,14 +57,6 @@ spec =
           , matchHeaders = ["Content-Range" <:> "*/*"]
           }
 
-      it "actually clears items ouf the db" $ do
-        _ <- request methodDelete "/items?id=lt.15" [] ""
-        get "/items"
-          `shouldRespondWith` [json|[{"id":15}]|]
-          { matchStatus  = 200
-          , matchHeaders = ["Content-Range" <:> "0-0/*"]
-          }
-
     context "known route, no records matched" $
       it "includes [] body if return=rep" $
         request methodDelete "/items?id=eq.101"
--- a/test/Feature/EmbedDisambiguationSpec.hs
+++ b/test/Feature/EmbedDisambiguationSpec.hs
@@ -194,9 +194,9 @@ spec =
             { matchHeaders = [matchContentTypeJson] }
 
         it "can request two parents with fks" $
-          get "/articleStars?select=createdAt,article(owner),user(name)&limit=1" `shouldRespondWith`
-            [json|[{"createdAt":"2015-12-08T04:22:57.472738","article":{"owner": "postgrest_test_authenticator"},"user":{"name": "Angela Martin"}}]|]
-            { matchHeaders = [matchContentTypeJson] }
+          get "/articleStars?select=createdAt,article(id),user(name)&limit=1"
+            `shouldRespondWith`
+              [json|[{"createdAt":"2015-12-08T04:22:57.472738","article":{"id": 1},"user":{"name": "Angela Martin"}}]|]
 
         it "can specify a view!fk" $
           get "/message?select=id,body,sender:person_detail!message_sender_fkey(name,sent),recipient:person_detail!message_recipient_fkey(name,received)&id=lt.4" `shouldRespondWith`
--- a/test/Feature/InsertSpec.hs
+++ b/test/Feature/InsertSpec.hs
@@ -135,18 +135,14 @@ spec actualPgVersion = do
 
     context "with no pk supplied" $ do
       context "into a table with auto-incrementing pk" $
-        it "succeeds with 201 and link" $ do
-          p <- post "/auto_incrementing_pk" [json| { "non_nullable_string":"not null"} |]
-          liftIO $ do
-            simpleBody p `shouldBe` ""
-            simpleHeaders p `shouldSatisfy` matchHeader hLocation "/auto_incrementing_pk\\?id=eq\\.[0-9]+"
-            simpleStatus p `shouldBe` created201
-          let Just location = lookup hLocation $ simpleHeaders p
-          r <- get location
-          let [record] = fromJust (JSON.decode $ simpleBody r :: Maybe [IncPK])
-          liftIO $ do
-            incStr record `shouldBe` "not null"
-            incNullableStr record `shouldBe` Nothing
+        it "succeeds with 201 and location header" $ do
+          post "/auto_incrementing_pk"
+              [json| { "non_nullable_string":"not null"} |]
+            `shouldRespondWith`
+              ""
+              { matchStatus  = 201
+              , matchHeaders = [ "Location" <:> "/auto_incrementing_pk?id=eq.2" ]
+              }
 
       context "into a table with simple pk" $
         it "fails with 400 and error" $
@@ -197,21 +193,14 @@ spec actualPgVersion = do
 
     context "with compound pk supplied" $
       it "builds response location header appropriately" $ do
-        let inserted    = [json| { "k1":12, "k2":"Rock & R+ll" } |]
-            expectedObj = CompoundPK 12 "Rock & R+ll" Nothing
-            expectedLoc = "/compound_pk?k1=eq.12&k2=eq.Rock%20%26%20R%2Bll"
-        p <- request methodPost "/compound_pk"
-                     [("Prefer", "return=representation")]
-                     inserted
-        liftIO $ do
-          JSON.decode (simpleBody p) `shouldBe` Just [expectedObj]
-          simpleStatus p `shouldBe` created201
-          lookup hLocation (simpleHeaders p) `shouldBe` Just expectedLoc
-
-        r <- get expectedLoc
-        liftIO $ do
-          JSON.decode (simpleBody r) `shouldBe` Just [expectedObj]
-          simpleStatus r `shouldBe` ok200
+        request methodPost "/compound_pk"
+            [("Prefer", "return=representation")]
+            [json| { "k1":12, "k2":"Rock & R+ll" } |]
+          `shouldRespondWith`
+            [json|[ { "k1":12, "k2":"Rock & R+ll", "extra": null } ]|]
+            { matchStatus  = 201
+            , matchHeaders = [ "Location" <:> "/compound_pk?k1=eq.12&k2=eq.Rock%20%26%20R%2Bll" ]
+            }
 
     context "with bulk insert" $
       it "returns 201 but no location header" $ do
@@ -248,9 +237,9 @@ spec actualPgVersion = do
     context "attempting to insert a row with the same primary key" $
       it "fails returning a 409 Conflict" $
         post "/simple_pk"
-            [json| { "k":"k1", "extra":"e1" } |]
+            [json| { "k":"xyyx", "extra":"e1" } |]
           `shouldRespondWith`
-            [json|{"hint":null,"details":"Key (k)=(k1) already exists.","code":"23505","message":"duplicate key value violates unique constraint \"simple_pk_pkey\""}|]
+            [json|{"hint":null,"details":"Key (k)=(xyyx) already exists.","code":"23505","message":"duplicate key value violates unique constraint \"simple_pk_pkey\""}|]
             { matchStatus  = 409 }
 
     context "attempting to insert a row with conflicting unique constraint" $
@@ -411,15 +400,22 @@ spec actualPgVersion = do
       it "succeeds and returns usable location header" $ do
         let payload = [json| { "k":"圍棋", "extra":"￥" } |]
         p <- request methodPost "/simple_pk?select=extra,k"
-                     [("Prefer", "return=representation")]
-                     payload
+            [("Prefer", "tx=commit"), ("Prefer", "return=representation")]
+            payload
         liftIO $ do
           simpleBody p `shouldBe` "["<>payload<>"]"
           simpleStatus p `shouldBe` created201
 
         let Just location = lookup hLocation $ simpleHeaders p
-        r <- get (location <> "&select=extra,k")
-        liftIO $ simpleBody r `shouldBe` "["<>payload<>"]"
+        get location
+          `shouldRespondWith`
+            [json|[ { "k":"圍棋", "extra":"￥" } ]|]
+
+        request methodDelete location
+            [("Prefer", "tx=commit")]
+            ""
+          `shouldRespondWith`
+            204
 
   describe "Row level permission" $
     it "set user_id when inserting rows" $ do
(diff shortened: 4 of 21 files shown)


test/Feature/AndOrParamsSpec.hs
```diff
@@ -33,13 +33,24 @@ spec actualPgVersion =
...
-              {"id": 2, "child_entities": [ { "id": 3, "grandchild_entities": []} ]},
-              {"id": 3, "child_entities": []}, {"id": 4, "child_entities": []}
-            ]|] { matchHeaders = [matchContentTypeJson] }
+          get "/entities?child_entities.grandchild_entities.or=(id.eq.1,id.eq.2)&select=id,child_entities(id,grandchild_entities(id))"
+            `shouldRespondWith`
+              [json|[
+                {"id": 1, "child_entities": [
+                  { "id": 1, "grandchild_entities": [ { "id": 1 }, { "id": 2 } ]},
+                  { "id": 2, "grandchild_entities": []},
+                  { "id": 4, "grandchild_entities": []},
```
wolfgangwalther, line 44: In some cases I had to add rows to `data.sql` to be available for other tests, so existing tests that depend on what is in those tables had to be adjusted.

test/Feature/AndOrParamsSpec.hs
```diff
@@ -210,12 +221,15 @@ spec actualPgVersion =
...
         request methodPost "/child_entities?select=id,entities(id)&entities.or=(id.eq.2,id.eq.3)&entities.order=id"
-          [("Prefer", "return=representation")]
-          [json|[{"id":4,"name":"entity 4","parent_id":1},
-                 {"id":5,"name":"entity 5","parent_id":2},
-                 {"id":6,"name":"entity 6","parent_id":3}]|] `shouldRespondWith`
-          [json|[{"id": 4, "entities":null}, {"id": 5, "entities": {"id": 2}}, {"id": 6, "entities": {"id": 3}}]|]
-          { matchStatus = 201, matchHeaders = [matchContentTypeJson] }
+            [("Prefer", "return=representation")]
+            [json|[
+              {"id":7,"name":"entity 4","parent_id":1},
```
wolfgangwalther, line 226: Because I had to add more rows to `data.sql`, some PKs were already used. Some requests had to be adjusted to use other PK ids, that were still available.

test/Feature/AndOrParamsSpec.hs
```diff
@@ -228,9 +242,10 @@ spec actualPgVersion =
     context "used with DELETE" $
       it "succeeds when using and/or params" $
         request methodDelete "/grandchild_entities?or=(id.eq.1,id.eq.2)&select=id,name"
-          [("Prefer", "return=representation")] "" `shouldRespondWith`
-          [json|[{ "id": 1, "name" : "updated grandchild entity"},{ "id": 2, "name" : "updated grandchild entity"}]|]
```
wolfgangwalther, line 232: Some tests expected the "updated" state of a row from a previous test, `"updated grandchild entity"` here. Those had to be changed to expect the original, unchanged value.

test/Feature/DeleteSpec.hs
```diff
@@ -57,14 +57,6 @@ spec =
           , matchHeaders = ["Content-Range" <:> "*/*"]
           }
 
-      it "actually clears items ouf the db" $ do
```
wolfgangwalther, line 60: This test is obsolete, because "persistance" of all kinds of requests is now extensively covered in `RollbackSpec`.

test/Feature/InsertSpec.hs
```diff
@@ -135,18 +135,14 @@ spec actualPgVersion = do
...
     context "with no pk supplied" $ do
       context "into a table with auto-incrementing pk" $
-        it "succeeds with 201 and link" $ do
-          p <- post "/auto_incrementing_pk" [json| { "non_nullable_string":"not null"} |]
-          liftIO $ do
-            simpleBody p `shouldBe` ""
-            simpleHeaders p `shouldSatisfy` matchHeader hLocation "/auto_incrementing_pk\\?id=eq\\.[0-9]+"
-            simpleStatus p `shouldBe` created201
-          let Just location = lookup hLocation $ simpleHeaders p
-          r <- get location
```
wolfgangwalther, line 145: Sometimes rows were confirmed to exist or not exist anymore in the same tests with additional requests. Those failed with the rolled back main request.

Some of those, like in this case, could safely be removed. This test is about status code and location header. The location header is now hardcoded instead of a regex (because it's now predictable which ID it will get), so the location URL does not need to be confirmed anymore.

test/Feature/InsertSpec.hs
```diff
@@ -411,15 +400,22 @@ spec actualPgVersion = do
       it "succeeds and returns usable location header" $ do
         let payload = [json| { "k":"圍棋", "extra":"￥" } |]
         p <- request methodPost "/simple_pk?select=extra,k"
-                     [("Prefer", "return=representation")]
-                     payload
+            [("Prefer", "tx=commit"), ("Prefer", "return=representation")]
```
wolfgangwalther, line 403: In this case, I thought it was important to still test the working location header. In other cases it's the same thing about the persistance of the changes made.

In those cases, I explicitely added a `Prefer tx=commit` **and** a cleanup request at the end (also with `tx=commit`).

test/Feature/QueryLimitedSpec.hs
```diff
@@ -15,59 +15,56 @@ spec :: SpecWith ((), Application)
 spec =
   describe "Requesting many items with server limits(max-rows) enabled" $ do
     it "restricts results" $
-      get "/items"
-        `shouldRespondWith` [json| [{"id":1},{"id":2}] |]
-        { matchStatus  = 200
-        , matchHeaders = ["Content-Range" <:> "0-1/*"]
-        }
+      get "/items?order=id"
```
wolfgangwalther, line 18: Rolled back updates are not entirely without consequence. They change the order of rows on disk. That means that requesting `/items` will return those in a different order. Since PG does not give any guarantee about the order of results without `ORDER BY`, this is something that should be changed in any case. Those that were breaking, I added a `?order=` part.

test/Feature/RangeSpec.hs
```diff
@@ -199,62 +197,98 @@ spec = do
...
-          `shouldRespondWith` [json|[{"id":4}, {"id":5}, {"id":6}]|]
-          { matchStatus  = 206
-          , matchHeaders = ["Content-Range" <:> "0-2/4"]
-          }
-        request methodGet "/getallprojects_view?select=id&id=lt.3" [("Prefer", "count=planned")] ""
-          `shouldRespondWith` [json|[{"id":1}, {"id":2}]|]
-          { matchStatus  = 206
-          , matchHeaders = ["Content-Range" <:> "0-1/673"]
-          }
+        request methodGet "/items?select=id&id=gt.8"
```
wolfgangwalther, line 200: Some tests are reformatted, because I touched them intermediately, but then reverted that and added data instead. So some tests technically are not changed but just look better ;)

test/Feature/SingularSpec.hs
```diff
@@ -49,48 +48,52 @@ spec =
...
-        _ <- post "/addresses" [json| { id: 100, address: "xxx" } |]
-        _ <- post "/addresses" [json| { id: 101, address: "yyy" } |]
-        p <- request methodPatch "/addresses?id=gt.0"
-                [("Prefer", "return=representation"), singular]
-                [json| { address: "zzz" } |]
-        liftIO $ do
-          simpleStatus p `shouldBe` notAcceptable406
-          isErrorFormat (simpleBody p) `shouldBe` True
+        request methodPatch "/addresses"
+            [("Prefer", "tx=commit"), ("Prefer", "return=representation"), singular]
```
wolfgangwalther, line 84: Some tests that were expected to fail, but then had additional `get` requests to confirm that no data was changed, I had to add a `Prefer tx=commit` header. I also check explicitely for `Preference-Applied` to make sure that the "non-persistance" is because of the errored request and not because of a rollback.

test/Feature/SingularSpec.hs
```diff
@@ -277,18 +301,24 @@ spec =
                 , matchHeaders = [matchContentTypeSingular]
                 }
 
-      it "executes the proc exactly once per request" $ do
```
wolfgangwalther, line 280: This test description is just non-sense here. I think it was copied from another test, but the test does something else. corrected that.

test/Feature/UpdateSpec.hs
```diff
@@ -91,23 +100,38 @@ spec = do
...
+
+        -- put value back for other tests
+        request methodPatch "/no_pk?b=eq.1"
+            [("Prefer", "tx=commit")]
+            [json| { b: "0" } |]
+          `shouldRespondWith`
+            204
 
       it "can set a column to NULL" $ do
-        _ <- post "/no_pk" [json| { a: "keepme", b: "nullme" } |]
```
wolfgangwalther, line 106: Some "preparation" requests were removed, because in those cases existing data could be used - changes are rolled back anyway.

test/Feature/UpsertSpec.hs
```diff
@@ -233,87 +241,132 @@ spec =
 
       context "Inserting row" $ do
         it "succeeds on table with single pk col" $ do
-          get "/tiobe_pls?name=eq.Go" `shouldRespondWith` "[]"
-          put "/tiobe_pls?name=eq.Go" [json| [ { "name": "Go", "rank": 19 } ]|] `shouldRespondWith` 204
-          get "/tiobe_pls?name=eq.Go" `shouldRespondWith` [json| [ { "name": "Go", "rank": 19 } ]|] { matchHeaders = [matchContentTypeJson] }
+          -- assert that the next request will indeed be an insert
```
wolfgangwalther, line 244: Most of the `PUT` tests had some read requests before the main request to make sure that, what was going to happen next, would indeed be either an `INSERT` or `UPDATE` effectively. I added comments to those as an explanation and also added some of those requests, where they were missing.

test/Feature/UpsertSpec.hs
```diff
@@ -233,87 +241,132 @@ spec =
...
             `shouldRespondWith`
-            [json|{ "name": "Go", "rank": 19 }|]
-            { matchStatus = 200 , matchHeaders = [matchContentTypeSingular] }
+              [json|[ { "name": "Java", "rank": 1 } ]|]
+
+          request methodPut "/tiobe_pls?name=eq.Java"
+              [("Prefer", "return=representation"), ("Accept", "application/vnd.pgrst.object+json")]
+              [json| [ { "name": "Java", "rank": 19 }, { "name": "Swift", "rank": 12 } ] |]
+            `shouldRespondWith`
+              [json|{ "name": "Java", "rank": 19 }|]
```
wolfgangwalther, line 303: Put Java were it belongs....

test/Feature/UpsertSpec.hs
```diff
@@ -233,87 +241,132 @@ spec =
...
-            { matchHeaders = [matchContentTypeJson] }
-          put "/employees?first_name=eq.Susan&last_name=eq.Heidt"
-            [json| [ { "first_name": "Susan", "last_name": "Heidt", "salary": "60000", "company": "Gamma Gas", "occupation": "Railroad engineer" } ]|]
-            `shouldRespondWith` 204
-          get "/employees?first_name=eq.Susan&last_name=eq.Heidt"
+              [json| [ { "first_name": "Frances M.", "last_name": "Roe", "salary": "$24,000.00", "company": "One-Up Realty", "occupation": "Author" } ]|]
+
+          request methodPut "/employees?first_name=eq.Frances M.&last_name=eq.Roe"
+              [("Prefer", "return=representation")]
+              [json| [ { "first_name": "Frances M.", "last_name": "Roe", "salary": "60000", "company": "Gamma Gas", "occupation": "Railroad engineer" } ]|]
```
wolfgangwalther, line 314: Frances has gotten quite a raise in salary here. Good job!

test/Main.hs
```diff
@@ -87,44 +88,40 @@ main = do
...
-      mutSpecs = uncurry describe <$> [
-          ("Feature.DeleteSpec"             , Feature.DeleteSpec.spec)
-        , ("Feature.InsertSpec"             , Feature.InsertSpec.spec actualPgVersion)
-        , ("Feature.SingularSpec"           , Feature.SingularSpec.spec)
-        , ("Feature.UpdateSpec"             , Feature.UpdateSpec.spec)
-        ]
-
   hspec $ do
-    -- Only certain Specs need a database reset, this should be used with care as it slows down the whole test suite.
-    mapM_ (afterAll_ reset . before withApp) mutSpecs
```
wolfgangwalther, line 126: Removal of this improved performance from 55s to 53s for `postgrest-test-spec-all` on my machine. So... basically nothing :)

test/fixtures/schema.sql
```diff
@@ -1819,8 +1819,8 @@ create table v1.parents (
 );
 
 create table v1.children (
-  id       serial primary key
```
wolfgangwalther, line 1822: Next to ordering of rows, this is something else that is not restored after rollback. Tests should be made more independent of auto-generated IDs in the future, I guess.

wolfgangwalther:
And it seems like `stack-test` is failing for two tests in CI. `postgrest-test-spec-all` runs fine on my local machine, so I'll have to see where the difference is.

And I did not run the tests against the same database without recreation yet. That's something that I want to do as well, I'm pretty sure some more failing tests will come up.

test/Feature/EmbedDisambiguationSpec.hs
```diff
@@ -194,8 +194,8 @@ spec =
             { matchHeaders = [matchContentTypeJson] }
 
         it "can request two parents with fks" $
-          get "/articleStars?select=createdAt,article(owner),user(name)&limit=1" `shouldRespondWith`
-            [json|[{"createdAt":"2015-12-08T04:22:57.472738","article":{"owner": "postgrest_test_authenticator"},"user":{"name": "Angela Martin"}}]|]
```
wolfgangwalther, line 198: So what happened here, is that before the rollback changes, it seems like this row was updated and the owner was set permanently to `postgrest_test_authenticator`. Locally, this is still the case, but in CI the row will have been created by the `circleci` sql user. That means we can't test for the owner here, because it might be different in different environments. For this reason I changed it to use the `id` instead.

steve-chavez:
@wolfgangwalther A rebase here would be good. I'll carefully review this one in case something slips by on the modified tests.

wolfgangwalther:
Rebased.

> I'll carefully review this one in case something slips by on the modified tests.

Yes, that would be great, thanks!

If possible, you could also try to run the tests twice against the same database, without re-creating the schema. I haven't been able to do that, yet. And I think the tests about "execute a rpc only once per request" need special attention as well. Not sure right now, if I even touched them, but they for sure need to keep state on the database. So they probably need to clean up as well, not sure they do.

test/Feature/InsertSpec.hs
steve-chavez: This one is much more readable now. I prefer the hardcoded Location.

test/Feature/InsertSpec.hs
steve-chavez: Agree, this one definitely needs persistence.

test/Feature/QueryLimitedSpec.hs
steve-chavez: Agree, I had to do a similar change before.

test/Feature/SingularSpec.hs
```diff
@@ -15,8 +15,7 @@ import SpecHelper
 spec :: SpecWith ((), Application)
 spec =
   describe "Requesting singular json object" $ do
-    let pgrstObj = "application/vnd.pgrst.object+json"
-        singular = ("Accept", pgrstObj)
+    let singular = ("Accept", "application/vnd.pgrst.object+json")
```
steve-chavez, line 18: Nice. Deleting unnecessary variables :+1:

test/Feature/SingularSpec.hs
steve-chavez: Ah.. so for testing errors we should add the `Prefer: tx=commit`. I'll take note.

test/Feature/UpdateSpec.hs
```diff
@@ -91,23 +100,38 @@ spec = do
           }
 
       it "can update multiple items" $ do
-        replicateM_ 10 $ post "/auto_incrementing_pk"
-          [json| { non_nullable_string: "a" } |]
-        replicateM_ 10 $ post "/auto_incrementing_pk"
-          [json| { non_nullable_string: "b" } |]
```
steve-chavez, line 97: This one I don't understand. Where did the `replicateM_` tests went? Or are they not necessary now?

test/Feature/UpsertSpec.hs
```diff
@@ -100,30 +100,38 @@ spec =
...
+              [("Prefer", "return=representation"), ("Prefer", "resolution=ignore-duplicates")]
+              [json| [
+                { "unique_key": 1, "value": "B" },
+                { "unique_key": 2, "value": "C" },
+                { "unique_key": 3, "value": "D" }
+              ]|]
+            `shouldRespondWith`
+              [json| [
+                { "unique_key": 2, "value": "C" },
+                { "unique_key": 3, "value": "D" }
```
steve-chavez, line 113: Hmm.. not figuring this one out. Why does it give a different result than before?

test/Feature/UpsertSpec.hs
```diff
@@ -100,30 +100,38 @@ spec =
...
+              [("Prefer", "return=representation"), ("Prefer", "resolution=ignore-duplicates")]
+              [json| [
+                { "key1": 1, "key2": 1, "value": "B" },
+                { "key1": 1, "key2": 2, "value": "C" },
+                { "key1": 1, "key2": 3, "value": "D" }
+              ]|]
+            `shouldRespondWith`
+              [json| [
+                { "key1": 1, "key2": 2, "value": "C" },
+                { "key1": 1, "key2": 3, "value": "D" }
```
steve-chavez, line 130: :thinking: Same here. Different result.

steve-chavez:
@wolfgangwalther Finally gave this one a review :) Check my above comments.

I like that persistence in the tests has to be explicit now with `Prefer: tx=commit`. Also noted that for testing errors we should now use `Prefer: tx=commit` as well.

test/Feature/SingularSpec.hs
wolfgangwalther: Not all errors need `Prefer: tx=commit`. Once you test for "non-persistance after an error" you need it. Most tests for errors do not have the "non-persistance test" part.

test/Feature/UpdateSpec.hs
wolfgangwalther: Yeah, that's not easy from the diff, because you need context that is not in here.

This test is about `PATCH`ing multiple rows at the same time. The `replicateM_` are just calls to prepare the `auto_incrementing_pk` to hold multiple rows with the same values. I wanted to avoid so many persistant changes like that, because they would also affect the serial pk column a lot, even if undone. So I went looking for another table where we already have a few rows with the same values in one column, to make a "multiple-rows-patch-request" against.

In `data.sql` we have:

```
INSERT INTO no_pk VALUES (NULL, NULL);
INSERT INTO no_pk VALUES ('1', '0');
INSERT INTO no_pk VALUES ('2', '0');
```

Ok, this test does not change 10 rows at the same time anymore, but only 2 - but that's the same.

The new test changes those `0` to `1`. So I test for 0 first, then make the patch request, then test for 1, then make the reverse patch request to put the data back in the original state.

test/Feature/UpsertSpec.hs
wolfgangwalther: So this tests `ignore-duplicates`. The return value should only have those rows that were actually inserted.

The difference is in the 2nd row `{ "unique_key": 2, "value": "C" }`. Before the all-rollback changes this existed already at the start of this test, now it doesn't.

A couple of tests up, there is a test `it "INSERTs and UPDATEs rows on single unique key conflict" $`. This test updates the `unique_key: 1` row, which is created in `data.sql`, and **inserts the `unique_key: 2` row**. Because those changes are rolled back, this row is now not there anymore.

We are only interested in `unique_key: 1` **not** showing up in the output here, so it's alright. I could have removed the `unique_key: 3` row alltogether, I guess...

test/Feature/UpsertSpec.hs
wolfgangwalther: Same reason as above!

wolfgangwalther:
> I like that persistence in the tests has to be explicit now with `Prefer: tx=commit`. Also noted that for testing errors we should now use `Prefer: tx=commit` as well.

I somehow can't see my comment about this above easily, so I repeat what I wrote there, here:

Not all errors need Prefer: tx=commit. Once you test for "non-persistance after an error" you need it. Most tests for errors do not have the "non-persistance test" part.



> > If possible, you could also try to run the tests twice against the same database, without re-creating the schema.
> 
> Not sure how to do that quickly :(.

Yeah, I wasn't sure about that either. I think I now know how: I will just wrap two calls to `cabal v2-test` in the same `test/with_tmp_db` call. That should work. I will let you know what happens.

wolfgangwalther:
So I did this, and it worked:

```
test/with_tmp_db sh -c 'cabal v2-test -f FailOnWarn --test-show-detail=direct && cabal v2-test -f FailOnWarn --test-show-detail=direct'
```

Showing another 6 failures. Will fix those in a 2nd commit, so you can review those independently.

wolfgangwalther:
Ok, this should be good to merge now from my perspective.

test/Feature/UpdateSpec.hs
steve-chavez: Aha, got it!

steve-chavez:
@wolfgangwalther Same here. Looks good to go!

PR #1664, opened 2020-11-22
Add test to ignore functions unnamed arguments for rpc disambiguation

### LINKED ISSUE #1638: Stored Procedure returning a table type only returns one single row if another function has the same name
### Environment

PostgreSQL version: PostgreSQL 12.4 (Debian 12.4-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
PostgREST version: PostgREST 7.0.1
Operating system: Debian 10 (buster) / Linux LABO-BDD-dev 4.15.18-27-pve #1 SMP PVE 4.15.18-55 (Tue, 17 Mar 2020 15:32:02 +0100) x86_64 GNU/Linux

### Description of issue & steps to reproduce

Let's say you have a view:
```
CREATE OR REPLACE VIEW "films" AS
SELECT * FROM ( VALUES ('2001 Space Odissey'::text), ('Interstellar'::text), ('Gone with the Wind'::text)) film_titles;
```

And you create a Stored Procedure that returns this table type:
```
CREATE FUNCTION my_films() RETURNS SETOF films AS $$
BEGIN
  RETURN query SELECT * FROM films;
END;
$$ LANGUAGE plpgsql IMMUTABLE STRICT;
```

Expected behaviour (and working) is that if you ask `/rpc/my_films` (without any parameters) you'll get all the rows from this view.

Now, you create another Function with completely different signature, but same name:
```
CREATE FUNCTION my_films(int, date, text) RETURNS date
AS $$
BEGIN
RETURN null;
END;
$$ LANGUAGE plpgsql IMMUTABLE STRICT;
```

And here the bug comes: now `/rpc/my_films` will only give you the first row of the view.

The workaround I found is to give names to all parameters of this second function:
```
DROP FUNCTION my_films(int,date,text);
CREATE FUNCTION my_films(foo int, bar date, something text) RETURNS date
AS $$
BEGIN
RETURN null;
END;
$$ LANGUAGE plpgsql IMMUTABLE STRICT;
```

And then it's working normally again.

wolfgangwalther:
Resolves #1638.

This has in fact already been fixed in e08bb3a197cbc03e5c623b9847651880955e45b8.

Just added another function with the same name as a set-returning function to the fixtures. Before this commit, this breaks a test, but now it passes. Confirms, that it is fixed already.

### DIFF
--- a/test/fixtures/schema.sql
+++ b/test/fixtures/schema.sql
@@ -970,6 +970,12 @@ create function test.ret_setof_integers() returns setof integer as $$
   values (1), (2), (3);
 $$ language sql;
 
+-- this function does not have named arguments and should be ignored
+-- if it's not ignored, it will break the test for the function before
+create function test.ret_setof_integers(int, int) returns integer AS $$
+  values (1);
+$$ language sql;
+
 create function test.ret_scalars() returns table(
   a text, b test.enum_menagerie_type, c test.one_nine, d int4range
 ) as $$
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1585, Fix error messages on connection failure for localized postgres on Windows - @wolfgangwalther
  - #1636, Fix `application/octet-stream` appending `charset=utf-8` - @steve-chavez
  - #1615, Fix RPC return type handling and embedding for domains with composite base type - @wolfgangwalther
- - #1469, Fix overloading of functions with unnamed arguments - @wolfgangwalther
+ - #1469, #1638 Fix overloading of functions with unnamed arguments - @wolfgangwalther
 
 ### Changed
 


test/fixtures/schema.sql
```diff
@@ -970,6 +970,12 @@ create function test.ret_setof_integers() returns setof integer as $$
   values (1), (2), (3);
 $$ language sql;
 
+-- this function does not have named arguments and should be ignored
+-- if it's not ignored, it will break the test for the function before
+create function test.ret_setof_integers(int, int) returns integer AS $$
+  values (1);
+$$ language sql;
```
steve-chavez, line 977: Got it. How about adding #1638 to the same CHANGELOG entry:

https://github.com/PostgREST/postgrest/commit/e08bb3a197cbc03e5c623b9847651880955e45b8#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR28

wolfgangwalther: Absolutely!

PR #1665, opened 2020-11-22
Add docker files to run nix dev environment on Windows

wolfgangwalther:
There is no nix on Windows. This adds a docker container to run the whole nix environment in. Works surprisingly well :). At least on my linux machine, haha. Windows is a bit slower and currently still building the image.

This could also be useful for those, who have trouble installing nix. It didn't work for me via my distributions package manager, so that delayed my arrival in the nix world a bit.

This will also allow us to remove `test/Dockerfile` and `test/docker-compose.yml`. I used them for development on Windows, when I was still doing that, but I doubt they serve any other purpose now. When I did that, I got a little crazy and thought: What if we just remove all the other dev tools, because everyone should use the nix dev environment anyway? So I went ahead and ripped out a few parts. I know this might be controversial, so this is for discussion :).


Makefile
```diff
@@ -1,43 +1,4 @@
-.PHONY: commit-check check clean lint style test test-watch coverage circleci circleci-memory prompt-clean prompt-long-process cachix-push-all
```
wolfgangwalther, line 1: I guess this will be the most controversial... Thoughts?

test/destroy_test_db
```diff
@@ -1,65 +0,0 @@
-#! /usr/bin/env bash
```
wolfgangwalther, line 1: This file was not used anywhere anymore. Does anyone still use it manually? I think this is handled by just recreating the whole database now, no?

test/fixtures/dumpfixture.sh
```diff
@@ -1 +0,0 @@
-pg_dump --host localhost --port 5432 --username "postgres" --no-password  --format plain --column-inserts --verbose --file "./test/fixtures/schema.sql" "postgrest_test"
```
wolfgangwalther, line 1: I don't think anyone does (or should for that matter!) use this command anymore. Am I wrong?

.github/CONTRIBUTING.md
```diff
@@ -44,7 +44,7 @@ your contributions.
 
 * All code must also pass [hlint](http://community.haskell.org/~ndm/hlint/) and [stylish-haskell](https://github.com/jaspervdj/stylish-haskell)
   with no warnings. This helps enforce a uniform style for all committers. Continuous integration will check this as well on every
-  pull request. There's a useful Makefile that helps with checking this locally. You can run `make commit-check` to do this manually but
+  pull request. There are useful tools in the nix-shell that help with checking this locally. You can run `NIX COMMAND HERE` to do this manually but
```
wolfgangwalther, line 47: Already talked to @monacoremo about that: A replacement for "make commit-check" / "make check" would be cool.

Makefile
monacoremo: Bold, I like it :-) We still want to keep stack as a first class citizen for development, but I don't think this hurts that goal. 

Suggest that we even completely kill the Makefile, `cachix-push-all` can be something that we note in the nix/README or add as e.g. 'postgrest-cachix-push' (to be then also used in CI). Its use will always be reserved to a very limited group anyway.

This will also be an opportunity to streamline the postgrest-docs on running tests.

.github/CONTRIBUTING.md
monacoremo: Yes, it's going to be pretty simple to add a `postgrest-check` that checks everything. We could even run that one in CI, theoretically - but then we would lose some readability in the CI outputs, so best to keep the steps separate.

test/destroy_test_db
monacoremo: postgrest-docs still references it I think, but we will be able to significantly simplify those.

Makefile
wolfgangwalther: > Bold, I like it :-) We still want to keep stack as a first class citizen for development, but I don't think this hurts that goal.

stack for development (i.e. including tools) or just to be able to build from source?

I am not against keeping "stack for development", but mainly because I have no idea about that. So please help my understanding here: For me, developing for PostgREST - what kind of things are not possible for me to do with the nix devtools, that I need stack for?

> Suggest that we even completely kill the Makefile, `cachix-push-all` can be something that we note in the nix/README or add as e.g. 'postgrest-cachix-push' (to be then also used in CI). Its use will always be reserved to a very limited group anyway.

Perfect.

> This will also be an opportunity to streamline the postgrest-docs on running tests.

Even more so!

.github/CONTRIBUTING.md
wolfgangwalther: Yeah, I agree that in CI the steps should be kept separate.

Makefile
monacoremo: The nix dev environment covers everything and is the preferred setup. 

However, we don't want to lock out one-off contributors if they don't use Nix, so we want to keep building and running basic tests with stack functional. We cover this with the stack.yaml and tests/with_tmp_db script, ensuring that it still works with the one CI job. Linting/styling the haskell source code correctly will be a bit more tricky for non-nix-users when we remove the Makefile, but we'll always have our CI to quickly flag issues if needed.

steve-chavez: One more reason to keep stack. robx receives notifications from them to keep dependencies up-to-date. See:

- https://github.com/PostgREST/postgrest/pull/1611#issue-498341270
- https://github.com/commercialhaskell/stackage/issues/5672.

Really helpful for us. stack upgrades GHC on their LTSs so this lets us know what we need to do to upgrade.

steve-chavez: I would not be opposed to killing :skull:  the Makefile though, I only use Nix for development now :)

wolfgangwalther:
I now tested this on Windows - best development experience I ever had there with PostgREST. `postgrest-test-spec-all` ran without a problem. Nice!

Makefile
wolfgangwalther: Just killed the Makefile! Felt good!

.github/CONTRIBUTING.md
```diff
@@ -44,8 +44,8 @@ your contributions.
 
 * All code must also pass [hlint](http://community.haskell.org/~ndm/hlint/) and [stylish-haskell](https://github.com/jaspervdj/stylish-haskell)
   with no warnings. This helps enforce a uniform style for all committers. Continuous integration will check this as well on every
-  pull request. There's a useful Makefile that helps with checking this locally. You can run `make commit-check` to do this manually but
-  we recommend adding it to `.git/hooks/pre-commit` to automatically check this before doing a commit.
+  pull request. There are useful tools in the nix-shell that help with checking this locally. You can run `postgrest-check` to do this manually but
+  we recommend adding it to `.git/hooks/pre-commit` as `nix-shell --run postgrest-check` to automatically check this before doing a commit.
```
wolfgangwalther, line 48: @monacoremo is this the correct command to run it without actually launching `nix-shell`?

.github/CONTRIBUTING.md
```diff
@@ -44,8 +44,8 @@ your contributions.
 
 * All code must also pass [hlint](http://community.haskell.org/~ndm/hlint/) and [stylish-haskell](https://github.com/jaspervdj/stylish-haskell)
   with no warnings. This helps enforce a uniform style for all committers. Continuous integration will check this as well on every
-  pull request. There's a useful Makefile that helps with checking this locally. You can run `make commit-check` to do this manually but
-  we recommend adding it to `.git/hooks/pre-commit` to automatically check this before doing a commit.
+  pull request. There are useful tools in the nix-shell that help with checking this locally. You can run `postgrest-check` to do this manually but
```
wolfgangwalther, line 47: `postgrest-check` is not an exact replacement of `make commit-check`. The latter only did style and lint, while the former also does spec tests in addition.

Guys, do you think it is ok like that or do we need another `postgrest-commit-check` here? @steve-chavez @monacoremo

monacoremo: Would be an improvement IMO!

wolfgangwalther: Which one exactly would be the improvement? The `postgrest-check` as is or the `postgrest-commit-check`? :D

monacoremo: Sorry, that was unclear - I meant leaving it at postgrest-check and having the extended checks by default. If anyone wants to check less, they can always use the specific commands

test/destroy_test_db
wolfgangwalther: Opened https://github.com/PostgREST/postgrest-docs/pull/378 to make sure we're not loosing track of that.

.github/CONTRIBUTING.md
wolfgangwalther: @steve-chavez If you think otherwise, we can still add it later. Everything else is addressed now, so I'm merging this one as well.

PR #1666, opened 2020-11-22
Improve dev tools in the Nix environment

monacoremo:
@wolfgangwalther has started using the nix tooling and had a plethora of excellent ideas to improve the developer experience. Should not be too difficult to realize, so tracking here:

* [x] add a script to watch for changes (e.g. to re-run tests)
  * [x] add docs
* [x] add a script to run all checks
* [x] add a script to push all artifacts to cachix
* [x] on failure in postgrest-test-spec-all, print the failing postgres version before exiting
* [x] run tests against postgest in ascending/chronological order in postgrest-test-spec-all (was alphabetical)
* [x] add docs on postgrest-style-check

Tbd/potentially for later PRs:
* integrate coverage
* option to preserve temporary dbs and explore them
* script to run postgrest against a temporary db

Edit: tracking those in #1671 

* [x] Refactoring: will also use this opportunity to factor out `set -euo pipefail` from writeShellScriptBin uses and lint bash scripts that we define in Nix.

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -59,7 +59,7 @@ Within `nix-shell`, you can run Cabal commands as usual. You can also run
 stack with the `--nix` option, which causes stack to pick up the non-Haskell
 dependencies from the same pinned Nixpkgs version that the Nix builds use.
 
-## Aside: Working with `nix-shell` and the PostgREST utility scripts
+## Working with `nix-shell` and the PostgREST utility scripts
 
 The PostgREST utilities available in `nix-shell` all have names that begin with
 `postgrest-`, so you can use tab completion (typing `postgrest-` and pressing
@@ -74,6 +74,7 @@ postgrest-style-check               postgrest-test-spec-postgresql-13
 postgrest-test-spec                 postgrest-test-spec-postgresql-9.5
 postgrest-test-spec-all             postgrest-test-spec-postgresql-9.6
 postgrest-test-spec-postgresql-10
+...
 
 [nix-shell]$
 
@@ -94,6 +95,7 @@ postgrest-style-check               postgrest-test-spec-postgresql-12
 postgrest-test-io                   postgrest-test-spec-postgresql-13
 postgrest-test-spec                 postgrest-test-spec-postgresql-9.5
 postgrest-test-spec-all             postgrest-test-spec-postgresql-9.6
+...
 
 ```
 
@@ -160,6 +162,20 @@ $ nix-shell --run postgrest-lint
 $ nix-shell --run postgrest-style
 
 ```
+
+There is also `postgrest-style-check` that exits with a non-zero exit code if
+the check resulted in any uncommited changes. It's mostly useful for CI.
+
+## General development tools
+
+Tools like `postgrest-build`, `postgrest-run` etc. are simple wrappers around
+`cabal` and should do what you expect. `postgrest-check` runs most checks that will
+also run in CI, with the exception of the IO and Memory checks that need to be run
+separately.
+
+`postgrest-watch` takes a command as an argument that it will re-run if any source
+file is changed. For example, `postgrest-watch postgrest-test-spec-all` will re-run
+the full spec test suite against all PostgreSQL versions on every change.
 
 ## Tour
 
--- a/nix/devtools.nix
+++ b/nix/devtools.nix
@@ -1,49 +1,58 @@
-{ writeShellScriptBin
-, buildEnv
+{ buildEnv
+, cabal-install
+, checkedShellScript
+, entr
 , git
-, hlint
-, nixpkgs-fmt
 , silver-searcher
-, stylish-haskell
+, style
+, tests
 }:
 let
-  style =
-    writeShellScriptBin "postgrest-style"
+  watch =
+    checkedShellScript "postgrest-watch"
       ''
-        set -euo pipefail
-
         rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
 
-        # Format Nix files
-        ${nixpkgs-fmt}/bin/nixpkgs-fmt "$rootdir" > /dev/null 2> /dev/null
+        ${silver-searcher}/bin/ag -l . "$rootdir" | ${entr}/bin/entr "$@"
+      '';
 
-        # Format Haskell files
-        ${silver-searcher}/bin/ag -l -g '\.l?hs$' "$rootdir" \
-          | xargs ${stylish-haskell}/bin/stylish-haskell -i
+  pushCachix =
+    checkedShellScript "postgrest-push-cachix"
+      ''
+        nix-store -qR --include-outputs "$(nix-instantiate)" \
+          | cachix push postgrest
+      '';
+
+  build =
+    checkedShellScript "postgrest-build"
+      ''exec ${cabal-install}/bin/cabal v2-build -f FailOnWarn "$@"'';
+
+  run =
+    checkedShellScript "postgrest-run"
+      ''exec ${cabal-install}/bin/cabal v2-run postgrest -f FailOnWarn -- "$@"'';
+
+  clean =
+    checkedShellScript "postgrest-clean"
+      ''
+        ${cabal-install}/bin/cabal v2-clean
       '';
 
   check =
-    writeShellScriptBin "postgrest-style-check"
+    checkedShellScript "postgrest-check"
       ''
-        set -euo pipefail
-
-        ${style}/bin/${style.name}
-
-        ${git}/bin/git diff-index --exit-code HEAD -- '*.hs' '*.lhs' '*.nix'
-      '';
-  lint =
-    writeShellScriptBin "postgrest-lint"
-      ''
-        set -euo pipefail
-
-        rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
-
-        # Lint Haskell files
-        ${silver-searcher}/bin/ag -l -g '\.l?hs$' "$rootdir" \
-          | xargs ${hlint}/bin/hlint -X QuasiQuotes -X NoPatternSynonyms
+        ${tests}/bin/postgrest-test-spec-all
+        ${style}/bin/postgrest-lint
+        ${style}/bin/postgrest-style-check
       '';
 in
 buildEnv {
   name = "postgrest-devtools";
-  paths = [ style check lint ];
+  paths = [
+    watch.bin
+    pushCachix.bin
+    build.bin
+    run.bin
+    clean.bin
+    check.bin
+  ];
 }
--- a/nix/devtools.nix
+++ b/nix/devtools.nix
@@ -1,6 +1,7 @@
 { buildEnv
 , cabal-install
 , checkedShellScript
+, devCabalOptions
 , entr
 , git
 , silver-searcher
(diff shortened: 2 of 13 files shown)


wolfgangwalther:
>     * run tests against postgest in ascending/chronological order in postgrest-test-spec-all (was alphabetical)

Actually my suggestion was to do it in descending order, if I remember correctly. This assumes that one develops with the latest version by default and then goes through `postgrest-test-spec-all`. So you already know it passes with pg13. If `all` goes through backwards, you immediately know the first version it breaks on. If you go upwards you know that it doesn't work in pg 9.5, but don't know where the cut is.

wolfgangwalther:
Current situation:
I am in nix-shell. I want to just build the executable without testing it. Ok, I ran `postgrest-test-spec` and have my build, but it took longer than it could. Maybe add a `postgrest-build`? Or am I missing something trivial here?

And now I want to execute the binary to test some cli options. Running `postgrest` tells me it can't be found. Would be good if the built executable would also be on the PATH in nix-shell, right?

wolfgangwalther:
It's probably just me missing any stack or cabal background...

So if I enter `cabal run postgrest`, I get what I want. Instead of changing the path, maybe a tool `postgrest` could be just that - an alias on top of `cabal run postgrest`, so that it builds automatically as well?

steve-chavez:
> Maybe add a postgrest-build
> maybe a tool postgrest-run

Those would be nice for me as well. I usually do:

- cabal v2-build
- cabal v2-run postgrest support/configs/config

monacoremo:
Good idea, especially as cabal is in this weird transition phase with the `v2-*` commands that need to be used.

So let's alias
* `postgrest-build` to `cabal v2-build`
* `postgrest-run` to something like `tests/with_tmp_db cabal v2-run postgrest` (+ one version per Postgres version) - the temporary db can be used if needed, or not
* potentially `clean` and others, tbd

To note: Putting a postgrest build on the nix-shell path immediately would require that we build postgrest through Nix as nix-shell starts up, causing potentially long load times, and would require that nix-shell is restarted when code has changed to get an updated version. Putting the cabal build directory on the path might be tricky, will look into that.

To build a production-ready binary, you would run `nix-build -A postgrestStatic`, but this would be a completely fresh, non-incremental build.

wolfgangwalther:
> To note: Putting a postgrest build on the nix-shell path immediately would require that we build postgrest through Nix as nix-shell starts up, causing potentially long load times, and would require that nix-shell is restarted when code has changed to get an updated version. Putting the cabal build directory on the path might be tricky, will look into that.

I don't think we need that, when we have `postgrest-run`. This is much better.

>     * `postgrest-run` to something like `tests/with_tmp_db cabal v2-run postgrest` (+ one version per Postgres version) - the temporary db can be used if needed, or not

This would run the executable with a default config to use the temp db, right?

nix/devtools.nix
```diff
@@ -1,49 +1,96 @@
...
+    checkedShellScript "postgrest-clean"
+      ''
+        ${cabal-install}/bin/cabal v2-clean
+      '';
+
+  check =
+    checkedShellScript "postgrest-check"
+      ''
+        ${style}
+        ${tests}/bin/postgrest-test-spec-all
```
wolfgangwalther, line 80: Can we flip those two around, and also add `${lint}`?

I think 1. test-spec, 2. lint, 3. style would be the best order. I first want to make sure that the specs pass, then make the linter happy once the code does what it should do, and then let style do some whitespace stuff.

But if the linter fails before the specs are run, doing a `postgrest-watch postgrest-check` might be annoying.

wolfgangwalther:
Ah it's really nice to see shellcheck in immediate action with lots of suggestions. Nice addition! :)

nix/devtools.nix
monacoremo: Yes, good idea! I'll also have to add a note that io and memory tests need to be run separately, as they depend on the postgrest executable as built by Nix (at least for now)

monacoremo:
> Ah it's really nice to see shellcheck in immediate action with lots of suggestions. Nice addition! :)

Oh yes, feels good - also to know that all future scripts and changes will be checked and solid!

nix/devtools.nix
wolfgangwalther: Yeah, I just looked into that as well. Can we not just add a `exec ${cabal-install}/bin/cabal v2-exec` before the memory and io tests scripts? This should already fix this, I think. I remember doing that back in the stack days, with stack exec.

wolfgangwalther:
I see no "shellcheck" tool in devtools.nix - so I assume the check is done at runtime of those tools, right? I follow that about half of our commands are checked in CI as well, but not all of them (release etc.).

monacoremo:
> I see no "shellcheck" tool in devtools.nix - so I assume the check is done at runtime of those tools, right? I follow that about half of our commands are checked in CI as well, but not all of them (release etc.).

I added 'checkedShellScript' in an overlay, which is now used everywhere in our nix setup. It checks the scripts at nix evaluation time, so only at first use e.g. when you launch nix shell.

monacoremo:
I will need to split out the styling/linting scripts, as at this point we are pulling in too many dependencies for our poor styling job...

wolfgangwalther:
Ah, I see. So `nix-shell` evaluates everything, right? Is the same true for those tasks that are triggered in CI? Or could we maybe just add another CI step that just launches the `nix-shell` and exit to trigger those?

monacoremo:
> Ah, I see. So `nix-shell` evaluates everything, right? Is the same true for those tasks that are triggered in CI? Or could we maybe just add another CI step that just launches the `nix-shell` and exit to trigger those?

The exact same checks happen when we do `nix-env -iA ...` or `nix-build` in CI - so it's all already covered!

The checks are not re-run when the exact scriot is already im the nix store or we pull from cachix, but everything in there is guaranteed to have been checked.

nix/devtools.nix
monacoremo: Yes, that should work! The profiled build for memory is a bit tricky though, so let's do that in a separate PR.

nix/README.md
```diff
@@ -161,6 +163,20 @@ $ nix-shell --run postgrest-style
...
 
+There is also `postgrest-style-check` that exits with a non-zero exit code if
+the check resulted in any uncommited changes. It's mostly useful for CI.
+
+## General development tools
+
+Tools like `postgrest-build`, `postgrest-run` etc. are simple wrappers around
+`cabal` and should do what you expect. `postgrest-check` runs most checks that will
+also run in CI, with the exception of the IO and Memory checks that need to be run
+separarely.
```
wolfgangwalther, line 174:
```suggestion
separately.
```

nix/README.md
```diff
@@ -161,6 +163,20 @@ $ nix-shell --run postgrest-style
...
+
+`postgrest-watch` takes a command as an argument that it will rerun if any source
+file is changed. For example, `postgrest-watch postgrest-test-spec-all` will re-run
+the full spec test suite againsg all PostgreSQL version on every change.
```
wolfgangwalther, line 178:
```suggestion
the full spec test suite against all PostgreSQL versions on every change.
```

nix/README.md
```diff
@@ -161,6 +163,20 @@ $ nix-shell --run postgrest-style
...
+the check resulted in any uncommited changes. It's mostly useful for CI.
+
+## General development tools
+
+Tools like `postgrest-build`, `postgrest-run` etc. are simple wrappers around
+`cabal` and should do what you expect. `postgrest-check` runs most checks that will
+also run in CI, with the exception of the IO and Memory checks that need to be run
+separarely.
+
+`postgrest-watch` takes a command as an argument that it will rerun if any source
```
wolfgangwalther, line 176:
```suggestion
`postgrest-watch` takes a command as an argument that it will re-run if any source
```
(same as below)

nix/README.md
monacoremo: Thanks! Completed this on my phone, works surprisingly well! But it's not so great for proof-reading :)

wolfgangwalther:
I would like to give this a test-run before merging by rebasing one of my current branches on it and using the tools. Can we squash all those commits in here into one? I can also do it, while I'm on it, if you want me to. I should be able to force push on this branch. Learned by accident :D

monacoremo:
> Can we squash all those commits in here into one? I can also do it, while I'm on it, if you want me to. I should be able to force push on this branch. Learned by accident :D

Yes, please do! Don't have access to my machine right now. Let's see those superpowers :-)

wolfgangwalther:
So here are a couple of notes. Maybe some of those can be fixed easily, others might take more work and be done later.

- I can't seem to exit `postgrest-watch`. When I hit CTRL+C, it restarts the watched program, but does not pass through sigint (or whatever it is). Could be specific to my setup (terminal) as well, could you give it a try?

- `postgrest-test-spec-all` outputs `Failed` after every run - even if successful. I think the `trap exit` needs to check for exit code as well.

```
Done running spec against postgresql-9.5.23.

Failed on postgresql-9.5.23
```

We might not need to check the return code at all, but just echo `Done` with the trap each time. Done does not imply success and the error will be shown above anyway?

- auto-complete does not work when adding a command after `postgrest-watch`. That would be really cool, I cant remember all those commands that we now have :)

- it seems like `postgrest-build` and `postgrest-run` on the one hand, and `postgrest-test-...` on the other have different build settings. That leads to repeated builds, when switching between those commands. When I repeat `postgrest-run`, it just says "Up to date" and starts immediately. But when I run `postgrest-run` and `postgrest-test-spec` after each other, it always rebuilds. Very quickly, but it's never "Up to date".

- the `postgrest-style`, `postgrest-lint` and the stylecheck are not available in nix-shell anymore? Probably a result of the refactor because of dependencies?

monacoremo:
Thank you for the thorough testing @wolfgangwalther !

>     * I can't seem to exit `postgrest-watch`. When I hit CTRL+C, it restarts the watched program, but does not pass through sigint (or whatever it is). Could be specific to my setup (terminal) as well, could you give it a try?

Same issue here - solved by not handling new files for now, which requires a restart of postgrest-watch. Will look further into a proper solution

>     * `postgrest-test-spec-all` outputs `Failed` after every run - even if successful. I think the `trap exit` needs to check for exit code as well.

Fixed by resetting the trap

>     * auto-complete does not work when adding a command after `postgrest-watch`. That would be really cool, I cant remember all those commands that we now have :)

That's really unfortunate, just noticed myself. Unfortunately this is beyond my bash-fu right now, will also need to solve later
 
>     * it seems like `postgrest-build` and `postgrest-run` on the one hand, and `postgrest-test-...` on the other have different build settings. That leads to repeated builds, when switching between those commands. When I repeat `postgrest-run`, it just says "Up to date" and starts immediately. But when I run `postgrest-run` and `postgrest-test-spec` after each other, it always rebuilds. Very quickly, but it's never "Up to date".

True, I missed the FailOnWarn flag. Should be fixed now I believe

>     * the `postgrest-style`, `postgrest-lint` and the stylecheck are not available in nix-shell anymore? Probably a result of the refactor because of dependencies?

Fixed

wolfgangwalther:
> True, I missed the FailOnWarn flag. Should be fixed now I believe

Hm, no I'm still getting ` - postgrest-7.0.1 (lib) (configuration changed)`, whenever I change between the two commands.



> Same issue here - solved by not handling new files for now, which requires a restart of postgrest-watch. Will look further into a proper solution

I noticed something else as well: When I do a watch of something that runs a bit longer (`test-spec-all` maybe) and I do a change while it is running, the tests are not restarted and also not run again after finishing. That's confusing, because it force me to restart/trigger the tests manually sometimes. Could be that I save multiple files and the tests already start at the first file, or that shortly after saving I see a typo and fix it.

wolfgangwalther:
> > True, I missed the FailOnWarn flag. Should be fixed now I believe
> 
> Hm, no I'm still getting ` - postgrest-7.0.1 (lib) (configuration changed)`, whenever I change between the two commands.

I have just confirmed, that this is because of `--test-show-detail=direct`. It seems to be possible to add this to the `v2-build` and `v2-run` commands without problem, even though it is probably unused. So we can do that easily.

wolfgangwalther:
Looks good! ready to merge from your side?

monacoremo:
Just confirmed that both issues from your last comments are fixed - good to go!

Thank you for reviewing and debugging cabal :-)

wolfgangwalther:
Awesome job remo, I'm looking forward to working with that!

PR #1667, opened 2020-11-22
Cleanup postgrest --help

wolfgangwalther:
Related to https://github.com/PostgREST/postgrest-docs/pull/367.

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -146,13 +146,26 @@ readPathShowHelp = customExecParser parserPrefs opts
 
     exampleCfg :: Doc
     exampleCfg = vsep . map (text . toS) . lines $
-      [str|db-uri = "postgres://user:pass@localhost:5432/dbname"
-          |db-schema = "public" # this schema gets added to the search_path of every request
+      [str|### REQUIRED:
+          |db-uri = "postgres://user:pass@localhost:5432/dbname"
+          |db-schema = "public"
           |db-anon-role = "postgres"
-          |# number of open connections in the pool
+          |
+          |### OPTIONAL:
+          |## number of open connections in the pool
           |db-pool = 10
-          |# Time to live, in seconds, for an idle database pool connection.
+          |
+          |## Time to live, in seconds, for an idle database pool connection.
           |db-pool-timeout = 10
+          |
+          |## extra schemas to add to the search_path of every request
+          |db-extra-search-path = "public"
+          |
+          |## Notification channel for reloading the schema cache
+          |db-channel = "pgrst"
+          |
+          |## Enable or disable the notification channel
+          |db-channel-enabled = false
           |
           |server-host = "!4"
           |server-port = 3000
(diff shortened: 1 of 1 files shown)


src/PostgREST/Config.hs
```diff
@@ -146,50 +146,56 @@ readPathShowHelp = customExecParser parserPrefs opts
 
     exampleCfg :: Doc
     exampleCfg = vsep . map (text . toS) . lines $
-      [str|db-uri = "postgres://user:pass@localhost:5432/dbname"
-          |db-schema = "public" # this schema gets added to the search_path of every request
+      [str|REQUIRED:
```
steve-chavez, line 149: How about adding this as a comment:
```suggestion
      [str|##REQUIRED:
```

#### Reasoning

I'd like to create a sample config in a quick step with `postgrest 2> postgrest.conf`. Right now it's not entirely usable, but it's something.

I also added that trick in http://postgrest.org/en/v7.0.0/install.html#running.

src/PostgREST/Config.hs
```diff
@@ -146,50 +146,56 @@ readPathShowHelp = customExecParser parserPrefs opts
...
+          |db-uri = "postgres://user:pass@localhost:5432/dbname"
+          |db-schema = "public"
           |db-anon-role = "postgres"
-          |# number of open connections in the pool
+          |
+          |OPTIONAL:
```
steve-chavez, line 154: Same suggestion here:
```suggestion
          |##OPTIONAL:
```

src/PostgREST/Config.hs
```diff
@@ -198,16 +204,16 @@ readPathShowHelp = customExecParser parserPrefs opts
...
           |## rollback all transactions by default, use for test environments
           |## disabled by default
-          |# tx-rollback-all = false
+          |tx-rollback-all = false
           |
           |## allow overriding the tx-rollback-all setting for a request by
           |## setting the Prefer: tx=[commit|rollback] header
           |## disabled by default
-          |# tx-allow-override = false
+          |tx-allow-override = false
```
steve-chavez, line 216: @wolfgangwalther Not sure if you'd be up for this. But recently we had feedback on hacker news about our config file: https://news.ycombinator.com/item?id=25170658

> I'd first assumed there would be a config file a mile long but it really is all Postgres.

So in general I think we should keep the number of options short(within reason). That lead me to think about your previous proposal about having a single `tx` config.

Would you be up for revisiting that? Would be another PR of course.

src/PostgREST/Config.hs
wolfgangwalther: Absolutely, I had that in mind as well, but I missed it when adding those to. I'll comment them out. Maybe even with `###` because it's one more level then the regular comments.

src/PostgREST/Config.hs
wolfgangwalther: Yeah sure, why not. I guess with the "In-Database configuration" stuff it would be even more impressive in terms of "size of config file needed".

steve-chavez: Cool. Then I think the config should be along the lines of your previous idea. Maybe like `tx-end=commit|rollback|commit-allow-override|rollback-allow-override`.

wolfgangwalther: See #1668

PR #1668, opened 2020-11-23
Refactor tx-... config options to single tx-end option

wolfgangwalther:
See https://github.com/PostgREST/postgrest/pull/1667#discussion_r528830815.

At some point the distinction between the two settings / bools has to be made. Instead of creating another type for this setting, passing this into Api.hs and then destructuring it there, I decided to do the destructuring immediately at the source, when parsing the config file. Everything else is untouched, the internal config object still has `configTxRollbackAll` and `configTxAllowOverride`.

What do you think - is that fine?

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -167,6 +167,18 @@ readPathShowHelp = customExecParser parserPrefs opts
           |## Enable or disable the notification channel
           |db-channel-enabled = false
           |
+          |## how to terminate database transactions
+          |## possible values are:
+          |## commit (default)
+          |##   transaction is always committed, this can not be overriden
+          |## commit-allow-override
+          |##   transaction is committed, but can be overriden with Prefer tx=rollback header
+          |## rollback
+          |##   transaction is always rolled back, this can not be overriden
+          |## rollback-allow-override
+          |##   transaction is rolled back, but can be overriden with Prefer tx=commit header
+          |db-tx-end = "commit"
+          |
           |server-host = "!4"
           |server-port = 3000
           |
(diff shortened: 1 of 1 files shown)


src/PostgREST/Config.hs
```diff
@@ -206,14 +206,17 @@ readPathShowHelp = customExecParser parserPrefs opts
...
+          |## possible values are:
+          |## commit (default)
+          |##   transaction is always committed, this can not be overriden
+          |## commit-allow-override
+          |##   transaction is committed, but can be overriden with Prefer tx=rollback header
+          |## rollback
+          |##   transaction is always rolled back, this can not be overriden
+          |## rollback-allow-override
+          |##   transaction is rolled back, but can be overriden with Prefer tx=commit header
+          |tx-end = "commit"
```
steve-chavez, line 219: Nice! I like the consolidated option. Also I'm fine with saving work on the Config type :+1: .

wolfgangwalther:
Now that I looked a fair bit at those config options when re-ordering them in my and your PRs ;)... should we rename this to `db-tx-end`? It sure is a db-related config option, so it belongs in that block, right?

wolfgangwalther:
Although that will be a bit harder to change, because of backwards-compatibility, some of those look like they should actually be prefixed as well:

- `secret-is-base64` -> `jwt-secret-is-base64`
- `role-claim-key` -> `jwt-role-claim-key`
- `max-rows` -> `db-max-rows`
- `pre-request ` -> `db-pre-request`
- `root-spec` -> `db-root-spec`

wolfgangwalther:
The 3 prefixes of `db-`, `server-` and `jwt-` would correspond nicely with what PostgREST at the core does - it connects those 3 pieces.

steve-chavez:
> should we rename this to db-tx-end?
> The 3 prefixes of db-, server- and jwt- would correspond nicely with what PostgREST at the core does 

Yeah, makes a lot of sense! Regarding backwards compat, maybe we could create aliases for those configs? Not sure how hard it is with the current code.

wolfgangwalther:
Just changed `tx-end` to `db-tx-end` for now and will do the other stuff in a different PR. Probably need the `--dump-config` option to properly test for the support of both the new and old aliases.

Issue #1671, opened 2020-11-28
[Meta] Developer experience improvements


monacoremo:
Many good ideas are coming together on how we can improve the developer experience,  e.g. in #1666 . Tracking here to work on them step by step:

* [x] integrate coverage into nix tooling (#1694)
* [x] option to preserve temporary dbs and explore them (#1722)
* [x] script to easily run postgrest against a temporary db (#1722)
* [x] run io tests in nix tooling against the incrementally built postgrest, instead of the nix built one (#1673 )
* [ ] run memory tests in nix tooling against the incrementally built postgrest, instead of the nix built one
* [x] autocomplete for postgrest-watch arguments (#1674, #1676)
* [x] have postgrest-watch watch new files also (#1676)
* [ ] accelerate CI by re-enabling caching for the cabal builds
* [x] add a check that runs tests twice against the same db (see https://github.com/PostgREST/postgrest/pull/1663#issuecomment-737900216) (PR: https://github.com/PostgREST/postgrest/pull/1693)
* [x] add docs on the postgrest-* scripts (#1706)
* [x] -O1 for dev builds
* [x] io tests: make the postgrest endpoint dynamic/switch to unix domain sockets by default to allow parallelization (#1704)
* [x] io tests: include https://pypi.org/project/pytest-parallel/ (used xdist for now in #1704)
* [x] io tests: add docs for advanced usage (postgrest-watch postgrest-test-io -k ... etc.) (#1704)

wolfgangwalther:
I took the freedom to split io/memory up in two bullets and mark the io tests as done per the referenced PR.

wolfgangwalther:
Regarding auto-complete for `postgrest-watch`, I found this: https://askubuntu.com/a/854365

Looks fairly simple.

monacoremo:
> Regarding auto-complete for `postgrest-watch`, I found this:

Nice! Should be pretty easy to plug into the shellHook

wolfgangwalther:
>     * [ ]  have postgrest-watch watch new files also

If I remember correctly, you had `-d` and a while loop around it, which broke `CTRL+C`. I just noticed that the exit code of `entr` is 0 when doing `CTRL+C` and 2 when a new file is added.

So something like this (pseudo) should work:

```
do
  entr -d ...
until exitcode = 0
```

wolfgangwalther:
> run memory tests in nix tooling against the incrementally built postgrest, instead of the nix built one

[...]

PR #1673, opened 2020-11-30
Use cabal v2-exec to run io tests, include io tests in nix-shell by default; remove obsolete ncat dependency

wolfgangwalther:
As mentioned in #1666 and #1671, this runs the io-tests with `cabal v2-exec`. I need to use those currently, so went ahead with that.

This does not change the memory tests, those are going to be a bit trickier.

Would love to have your feedback on this @monacoremo.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -8,7 +8,6 @@
 , git
 , haskell
 , lib
-, ncat
 , postgresql
 , postgresqlVersions
 , postgrest
@@ -78,12 +77,14 @@ let
     checkedShellScript
       name
       ''
+        env="$(cat ${postgrest.env})"
+        export PATH="$env/bin:${curl}/bin:$PATH"
+
         rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
         cd "$rootdir"
 
-        export PATH="${curl}/bin:$PATH"
-
-        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-exec ${devCabalOptions} "$rootdir"/test/io-tests.sh
+        ${cabal-install}/bin/cabal v2-build ${devCabalOptions}
+        ${cabal-install}/bin/cabal v2-exec ${withTmpDb postgresql} "$rootdir"/test/io-tests.sh
       '';
 
   testMemory =
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -81,9 +81,9 @@ let
         rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
         cd "$rootdir"
 
-        export PATH="${postgrestStatic}/bin:${curl}/bin:${ncat}/bin:$PATH"
+        export PATH="${curl}/bin:$PATH"
 
-        ${withTmpDb postgresql} "$rootdir"/test/io-tests.sh
+        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-exec ${devCabalOptions} "$rootdir"/test/io-tests.sh
       '';
 
   testMemory =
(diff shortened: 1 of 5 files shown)


wolfgangwalther:
Oh yeah: Because the static build is not a dependency anymore... the io-tests can quickly be loaded by default and don't need an arg to nix-shell anymore.

nix/tests.nix
```diff
@@ -81,9 +81,9 @@ let
         rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
         cd "$rootdir"
 
-        export PATH="${postgrestStatic}/bin:${curl}/bin:${ncat}/bin:$PATH"
```
wolfgangwalther, line 84: The ncat/nmap dependency was removed in #1630 already, so we don't need it anymore.

nix/tests.nix
```diff
@@ -81,9 +81,9 @@ let
...
+        export PATH="${curl}/bin:$PATH"
 
-        ${withTmpDb postgresql} "$rootdir"/test/io-tests.sh
+        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-exec -f FailOnWarn --test-show-detail=direct "$rootdir"/test/io-tests.sh
```
monacoremo, line 86: Lets use the `${devCabalOptions}` here!

wolfgangwalther: Oh boy... I just changed that before pushing, but forgot to commit it! :D

monacoremo (commented):
Will alsk need to remove ioTests from https://github.com/wolfgangwalther/postgrest/blob/9b61e976766ff81b06fcb33ead923c00c937997b/.circleci/config.yml#L197

monacoremo:
Looks good to me overall, great job!

wolfgangwalther:
Ok, cool. Will let run CI through just to make sure and then merge. Thanks for the quick review.

wolfgangwalther:
> Will alsk need to remove ioTests from https://github.com/wolfgangwalther/postgrest/blob/9b61e976766ff81b06fcb33ead923c00c937997b/.circleci/config.yml#L197

Your review was faster to point that out than CI was failing on my first commit. Might have to engage you instead of CI... ;)

monacoremo:
1

> Your review was faster to point that out than CI was failing on my first commit. Might have to engage you instead of CI... ;)

More of a slow CI issue tbh... let's look into accelerating that again, e.g. by re-enabling caching for the cabal builds

wolfgangwalther:
So, @monacoremo, we're not off the hook yet in terms of CI:
https://app.circleci.com/pipelines/github/PostgREST/postgrest/549/workflows/1994150c-cc07-4814-9b81-a0e28212d7ab/jobs/7278

It did work locally, but that might be because.... I didn't run `nix-shell --pure`, maybe?

wolfgangwalther:
Mh, no. It works fine with `nix-shell --pure`.

wolfgangwalther:
@monacoremo Am I missing the following two lines for the io tests maybe?

```
env="$(cat ${postgrest.env})"
export PATH="$env/bin:$PATH"
```

wolfgangwalther:
So, that did change something:
https://app.circleci.com/pipelines/github/PostgREST/postgrest/551/workflows/b3e470a7-cb83-42ac-802b-b35c92c92c2e/jobs/7287

But still not successful.

monacoremo:
Hm, very weird. It fails here https://github.com/wolfgangwalther/postgrest/blob/85f289c6d481a25dc6e993c2038bad7e61c4576b/test/io-tests.sh#L290 , but that could be caused by all kinds of things. Is v2-exec messing with the path somehow?

wolfgangwalther:
Yes, it does in fact. It adds the following in front of it:
```
/path/to/repo/postgrest/dist-newstyle/build/x86_64-linux/ghc-8.8.3/postgrest-7.0.1/x/postgrest/build/postgrest
```

This is, so that it can run the binary that is built. So this would be expected, I guess.

wolfgangwalther:
I changed the order of `cabal v2-exec` and `withTmpDb` here, but I doubt that this will make much of a difference. Both works for me locally.

monacoremo:
I think we might have to explicitly postgrest-build before. Does it still work locally after a postgrest-clean?

monacoremo:
From the cabal manpage on v2-exec:

> Compiler tools will be configured to see the same subset of the store that builds would see. The PATH is modified to make all executables in the dependency tree available (provided they have been built already).

That's probably it?

wolfgangwalther:
I guess so too. `postgrest-clean` made it fail for me as well. I'm pretty sure CI will pass now. Thanks for debugging cabal for me ;)

wolfgangwalther:
Yes, that worked. Only travis is lagging behind, but this is unrelated. Merging now, thanks again!

PR #1674, opened 2020-12-01
Add autocompletion for postgrest-watch

monacoremo:
TIL bash autocompletion. Fixes one item of #1671

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -35,5 +35,10 @@ lib.overrideDerivation postgrest.env (
       ++ lib.optional memoryTests postgrest.tests.memoryTests
       ++ lib.optional docker postgrest.docker
       ++ lib.optional release postgrest.release;
+
+    shellHook =
+      ''
+        complete -c postgrest-watch
+      '';
   }
 )


wolfgangwalther:
That looks a bit... short!

I get the following error when using it: 

```
postgrest-watch postbash: completion: function `_command' not found
```

wolfgangwalther:
That's even shorter! But it works... :D

monacoremo:
Then let's go even shorter! (`-F _command` seems to be non-standard, by `-c` apparently works? https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html specifies it as alias to `-A command`)

monacoremo:
We are going to have the most polished dev experience ever - no excuses now not to do many performance releases!

wolfgangwalther:
haha :)
Absolutely!

( @steve-chavez your job with the performace! )

wolfgangwalther:
So I'm pretty sure the pipeline fails is completely unrelated. Merging!

wolfgangwalther:
I already used postgrest-watch yesterday and it's great!

PR #1675, opened 2020-12-03
improve error output for io-tests and with_tmp_db

wolfgangwalther:
We used to get some "segmentation faults" for the io tests in CI. I am currently trying to set up a nix-based coverage script - and ran into those errors again.

It turns out that the seg fault is just bash complaining, because some traps are set where bash tries to kill itself multiple times. Something similar happenend in `with_tmp_db`, where those calls led to errors regarding the log file, because it didn't exist anymore.

After fixing all those traps and obsolete output I was left with.. nothing. The actual error that I was looking for was still hidden, because all postgrest stderr is piped to `/dev/null` in the `io-tests.sh`. Some errors during the test are expected, so it's fine to filter those out, but in my case I had some startup errors for postgrest due to missing libraries and stuff like this. With the change here, those errors will not be filtered out anymore.

I'm pretty sure that with those changes we will get useful error message next time the io-tests in CI fail that way. So don't be afraid of seg faults, @monacoremo, when trying to set up the circle caching again. Those are a thing of the past now!

### DIFF
--- a/test/io-tests.sh
+++ b/test/io-tests.sh
@@ -21,7 +21,17 @@ export POSTGREST_TEST_CONNECTION=${POSTGREST_TEST_CONNECTION:-"postgres:///postg
 cd "$(dirname "$0")"
 cd io-tests
 
-trap "kill 0" int term exit
+cleanup() {
+  # clean up trap to avoid bash segmentation fault
+  trap - sigint sigterm exit
+
+  # kill without output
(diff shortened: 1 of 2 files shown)


wolfgangwalther:
Since this just changes logging output and nothing of substance, I will wait for CI to hopefully finish and then merge.

PR #1676, opened 2020-12-04
dev: improve postgrest-watch experience

wolfgangwalther:
This PR does:
* speed up io tests
* avoid infinite loop while watching io tests
* improve auto-complete with postgrest-watch postgrest-run
* detect when files are added or deleted

TODO:
* [x] fix `postgrest-watch postgrest-lint`

### DIFF
--- a/nix/devtools.nix
+++ b/nix/devtools.nix
@@ -14,7 +14,9 @@ let
       ''
         rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
 
-        ${silver-searcher}/bin/ag -l . "$rootdir" | ${entr}/bin/entr -r "$@"
+        while true; do
+          (! ${silver-searcher}/bin/ag -l . "$rootdir" | ${entr}/bin/entr -dr "$@")
+        done
       '';
 
   pushCachix =
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -13,6 +13,7 @@
 , postgrest
 , postgrestStatic
 , postgrestProfiled
+, procps
 , runtimeShell
 }:
 let
@@ -78,7 +79,7 @@ let
       name
       ''
         env="$(cat ${postgrest.env})"
-        export PATH="$env/bin:${curl}/bin:$PATH"
+        export PATH="$env/bin:${curl}/bin:${procps}/bin:$PATH"
 
         rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
         cd "$rootdir"
--- a/shell.nix
+++ b/shell.nix
@@ -38,7 +38,8 @@ lib.overrideDerivation postgrest.env (
 
     shellHook =
       ''
-        complete -c postgrest-watch
+        source ${pkgs.bashCompletion}/etc/profile.d/bash_completion.sh
+        complete -F _command postgrest-watch
       '';
   }
 )
--- a/test/io-tests.sh
+++ b/test/io-tests.sh
@@ -88,8 +88,7 @@ readSecretFromFile(){
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+    sleep 0.1
   done
   if pgrStarted
   then
@@ -113,8 +112,7 @@ readDbUriFromStdin(){
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+    sleep 0.1
   done
   if pgrStarted
   then
@@ -131,8 +129,7 @@ reqWithRoleClaimKey(){
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+    sleep 0.1
   done
   authorsJwt=$(psql -qtAX "$POSTGREST_TEST_CONNECTION" -c "select jwt.sign('$2', 'reallyreallyreallyreallyverysafe');")
   httpStatus="$( authorsStatus "$authorsJwt" )"
@@ -151,8 +148,7 @@ invalidRoleClaimKey(){
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+    sleep 0.1
   done
   if pgrStarted
   then
@@ -169,8 +165,7 @@ ensureIatClaimWorks(){
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+    sleep 0.1
   done
   for i in {1..10}; do \
     iatJwt=$(psql -qtAX "$POSTGREST_TEST_CONNECTION" -c "select jwt.sign(row_to_json(r), 'reallyreallyreallyreallyverysafe') from ( select 'postgrest_test_author' as role, extract(epoch from now()) as iat) r")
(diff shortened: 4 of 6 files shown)


nix/devtools.nix
```diff
@@ -14,7 +14,18 @@ let
...
+          # invert error code: if entr returns with an error it's because a new file has been added or deleted
+          # if entr returns with exit code 0 it's because the user canceled via CTRL+C
+          trap 'return 0' ERR
+          ${silver-searcher}/bin/ag -l . "$rootdir" | ${entr}/bin/entr -dr "$@"
+          trap - ERR
+          return 1
+        }
+
+        while true; do
+          watch "$@"
```
wolfgangwalther, line 27: @monacoremo Can you have a look at this?

This allows detection of added and deleted files via `postgrest-watch`.

nix/tests.nix
```diff
@@ -78,7 +79,7 @@ let
       name
       ''
         env="$(cat ${postgrest.env})"
-        export PATH="$env/bin:${curl}/bin:$PATH"
+        export PATH="$env/bin:${curl}/bin:${procps}/bin:$PATH"
```
wolfgangwalther, line 82: just a fix, I forgot yesterday. Only affects running io tests in `nix-shell --pure`.

shell.nix
```diff
@@ -38,7 +38,8 @@ lib.overrideDerivation postgrest.env (
 
     shellHook =
       ''
-        complete -c postgrest-watch
+        source ${pkgs.bashCompletion}/etc/profile.d/bash_completion.sh
+        complete -F _command postgrest-watch
```
wolfgangwalther, line 42: @monacoremo Your first take at this was better. `-c` did not allow to complete filenames after `postgrest-watch postgrest-run` properly anymore, but `_command` does.

I think it only failed on my system, because my host doesn't have `bash-completion` installed. I use zsh. Adding `bash-completion` inside `nix-shell` fixes the `_command not found`.

test/io-tests.sh
```diff
@@ -88,8 +88,7 @@ readSecretFromFile(){
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
```
wolfgangwalther, line 92: @monacoremo I removed the non-standard fallback for sleep - because `nix-shell` does guarantee we're using coreutils' sleep... ehhh this is more of a question. Does it? Or is that taken from the host?

test/io-tests.sh
```diff
@@ -186,16 +181,16 @@ ensureIatClaimWorks(){
...
   pgrStart "./configs/app-settings.config"
   while pgrStarted && test "$( rootStatus )" -ne 200
   do
     # wait for the server to start
-    sleep 0.1 \
-    || sleep 1 # fallback: subsecond sleep is not standard and may fail
+    sleep 0.1
   done
-  sleep 11
+  sleep 2
```
wolfgangwalther, line 193: @steve-chavez Can you have a good look at this? You wrote the test.

I changed the pool timeout to 1s and the sleep to 2s. Was 10s / 11s before. This should achieve the same, but doesn't waste 9s of my life each time the io tests are re-run :D. Does that make sense?

test/io-tests.sh
```diff
@@ -207,15 +202,17 @@ ensureAppSettings(){
 }
 
 checkAppSettingsReload(){
-  pgrStart "./configs/sigusr2-settings.config"
+  configFile=$(mktemp)
+  trap "rm -f $configFile" ERR RETURN
+  cat "./configs/sigusr2-settings.config" > "$configFile"
+  pgrStart "$configFile"
```
wolfgangwalther, line 208: This is a fun one... `postgrest-watch postgrest-test-io` was running in an infinite loop before, because those tests change the config file on disk, which results in a restart, which... you know! :)

I copied the config to a tempfile and then made the changes there. Outside of our source dir -> no problemo!

test/io-tests.sh
steve-chavez: Sure! I just did a `postgrest-test-io` and it worked good.

Changing the pool timeout is much better. Don't know why I didn't thought of that before, I guess because it only wasted my own time  :laughing:

wolfgangwalther: You know - I don't think I would have even noticed without that nice `postgrest-watch postgrest-test-io` infinite loop I fixed here as well. But that kind of stuck out, because it was taking 10 seconds for this test - and then immediately restarted :D.

wolfgangwalther:
I also just noticed that `postgrest-watch postgrest-lint` is broken (it has been before). `hlint` is just printing it's help screen, because silver-searcher does not pass any files to it. I debugged a little bit - when I run silver-searcher inside `postgrest-lint` to list all files (without pattern) it returns a lot when run as `postgrest-lint` - but none, zero, nada when run with `postgrest-watch postgrest-lint`. I have no idea what happens here - any idea @monacoremo?

test/io-tests.sh
monacoremo: That's from before my time here, no idea where this would ever be an issue. LGTM :-) 

Overall those tests are stretching bash to it's limits IMO, it would be a bit easier to meintain as a Python script for example. But it's not too painful as is I guess

shell.nix
monacoremo: Awesome, had no idea about the bashCompletion package - great solution!

nix/devtools.nix
monacoremo: Looks good if it works! Not able to test it right now... I'm a bit puzzeled why entr is behaving like this, the plain loop worked well for me previously

wolfgangwalther: > the plain loop worked well for me previously

Thanks for the hint. This made me have another look. My function is essentially doing nothing else than negating the exit code of entr. TIL about the bash `!` operator - which does exactly the same. So instead of the whole function I should be able to just write:
```
while true; do
  ! ${silver-searcher}/bin/....
done
```

So you were only missing one character in your first approach :).

Note: I just tested it and needed 3 characters, because the whole command has to be wrapped in `(! ...)` to work.

test/io-tests.sh
wolfgangwalther: I agree: Having a proper unit test framework around it, would help. I had a brief look at things to test arbitrary executables. I found DejaGNU and quickly turned away again. But I also found [bats-core](https://github.com/bats-core/bats-core). This is essentially unit testing for bash. The bats syntax is even supported by shellcheck, so this would not block adding linting as well.

The alternative would be to add a python based unit test. Of course this would be more to do for a rewrite. How much upside do you see here @monacoremo? Any experience with python's unit test frameworks?

wolfgangwalther:
Debugging the `postgrest-watch postgrest-lint` situation. I added `ag` and `entr` to the shell and did this:

```
ag -l | entr -dr sh -c "echo hello world; ag -l
```

Whenever I change a file, I get "hello world" as an output. But the 2nd `ag` lists nothing at all..

When I remove the `-r` flag from entr it works. Do we need this @monacoremo? Everything seems to work now without it.

test/io-tests.sh
monacoremo: Py.test is super straight forward - just define a few functions in a file (e.g. ljke this https://github.com/monacoremo/daydream/blob/master/tests/test_api_accounts.py) and run.

Happy to take a stab at that

monacoremo:
Hm, we need the -r for any long running process that needs to be killed (see http://eradman.com/entrproject/ for details). Not sure why it would hurt with shorter running processes. Will need to test at next opportunity

test/io-tests.sh
wolfgangwalther: I the key question here is: How easy is it to set up the `postgrest` process running in the background, and then clean up this process afterwards? The request side seems super simple.

Probably a helper `postgrest(configPath)` to be used in a `with` block?

```
with postgrest('./configs/simple.config'):
  # run my test
```

Ah - that could be very good!

> Happy to take a stab at that

Yes, please go ahead!

wolfgangwalther:
> Hm, we need the -r for any long running process that needs to be killed (see http://eradman.com/entrproject/ for details). Not sure why it would hurt with shorter running processes. Will need to test at next opportunity

Hm. I think this is used for services, that keep running? But all our tools are returning at some point, except maybe `postgrest-run`, when you keep that running. I tested a few of the other tools, it's still possible to change a file mid-run of `test-spec-all` and it starts all over again.

monacoremo:
Ah, ok - if it aborts the running process it should be fine! A bit confused tbh, as I added -r specifically to fix the rest-all restart...

wolfgangwalther:
> Ah, ok - if it aborts the running process it should be fine! A bit confused tbh, as I added -r specifically to fix the test-all restart...

Yeah, scratch that. I don't know what I did earlier, it does not restart `test-spec-all`.

wolfgangwalther:
This might be releated: https://github.com/ggreer/the_silver_searcher/issues/753

I will try adding a path to `postgrest-lint`.

wolfgangwalther:
Adding a path did not work - but adding `--vimgrep` as suggested in the same thread did work. Now we have `-r` in place, but `postgrest-lint` still works.

Issue #1624, opened 2020-10-15
Can't set app settings when using environment variables


LorenzHenk:
### Environment

* PostgreSQL version: 13
* PostgREST version: v7.0.1

### Description of issue

The documentation mentions the `app.settings.*` settings, where one can set custom session variables.

How can this setting be set when configuring PostgREST through environment variables?

I tried:
- `PGRST_APP.SETTINGS.TEST` - this is not allowed in the shell
- `PGRST_APP_SETTINGS_TEST` - this is not recognized by PostgREST

steve-chavez:
@LorenzHenk Are you on docker?

Otherwise the config should accept env vars like:
```
app.settings.test = "$(A_TEST)"
```

(From: http://postgrest.org/en/v7.0.0/configuration.html#app-settings)

LorenzHenk:
Yes I'm using docker & docker-compose. 

I'm not using a config file. Instead, I'm using environment variables (like `PGRST_DB_URI`) for configuration.

steve-chavez:
I see. Right now the config file is the only way to use the `app.settings.*` configs.

You could do it like:

```
docker run -v /absolute/path/to/config:/etc/postgrest.conf postgrest/postgrest
```

This is a limitation because we currently use a static [postgrest.conf](https://github.com/PostgREST/postgrest/blob/master/nix/docker/postgrest.conf) inside Docker.

steve-chavez:
One way we could solve this is by changing the `app.settings` config to use a single value(thinking json for now) for all the settings.

In the regular binary, one would use like:
```
app-settings = "{'one': '$(ONE)', 'two': '$(TWO)'}"
```
On Docker:
```
PGRST_APP_SETTINGS = "{'one': 'value', 'two': 'other value'}"
```

Would that work for you?

steve-chavez:
The other way to solve this would be to make our binary also understand command line arguments instead of only the config file. But I think that would be more work since we'd need to integrate an additional library.

wolfgangwalther:
I think we should just separate the config file from the environment variables. If we read those in a second step manually, we can easily read all of the custom settings with a pattern like `PGRST_APP_SETTINGS_xxx`.

Environment variables should override those values, that were set in the config file. Including environment variables with the current syntax in the config file would still be possible, because that's part of the lib we're using, right?

Since we would be parsing the environment variables ourself, that would solve #1572 as well.

LorenzHenk:
[...]

Issue #1572, opened 2020-08-15
Passing Bool parameters to docker images is a bit broken


ruslantalpa:
For `db-channel-enabled` and `secret-is-base64` if the value is passed as string as in `"true"` as it's done when working with docker images (since there you have `db-channel-enabled = "$(PGRST_DB_CHANNEL_ENABLED)"`) the config will be set to `False` and that is because `readMaybe` will correctly parse only `True` but not `true`
this line
https://github.com/PostgREST/postgrest/blob/bd2160db26210d41821825925b7149bf30566069/src/PostgREST/Config.hs#L289

needs to be changed to
```
coerceBool (C.String b) = readMaybe.toS $ toTitle b
``` 
(toTitle is from Data.Text)

also, after this fix, when passing the value through docker-compose, note this
```
PGRST_DB_CHANNEL_ENABLED=true #will work
PGRST_DB_CHANNEL_ENABLED="true" #will fail
```

PR #1678, opened 2020-12-04
added: cli option --dump-config prints loaded config and exits

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/pull/1659#discussion_r528413899, this adds the CLI option `--dump-config`, which prints the config after loading and then exits. The printed config is itself a valid postgrest config file.

This is mostly useful as a base for future improvements. When adding aliases for some config options (https://github.com/PostgREST/postgrest/pull/1668#issuecomment-732367387) or reading options directly from environment variables (#1624, #1572), this can be used as the base to write proper tests.

Also added colors to our io-tests. They are all nice and green now. Much easier to spot the occasional red one...


test/io-tests.sh
```diff
@@ -36,17 +36,71 @@ trap cleanup sigint sigterm exit
...
-ko(){ result 'not ok' "- $1"; failedTests=$(( $failedTests + 1 )); }
+result(){ echo -e "$1 $currentTest $2${NC}"; currentTest=$(( $currentTest + 1 )); }
+todo(){ result "${YELLOW}ok" "# TODO: $*"; }
+skip(){ result "${YELLOW}ok" "# SKIP: $*"; }
+ok(){ result "${GREEN}ok" "- $1"; }
+ko(){ result "${RED}not ok" "- $1"; failedTests=$(( $failedTests + 1 )); }
 comment(){ echo "# $1"; }
 
+####################
+# FOREGROUND TESTS #
```
wolfgangwalther, line 57: appreciate suggestions for a better name for this section and the "background" section. The distinction here is, that for those tests `postgrest` immediately returns and is not sent to the background.

Maybe it's just not needed at all. But it does give a bit of structure to that file. I'm thinking about running shellcheck against it.... but I am bit afraid of the result as well :D

test/io-tests/configs/no-defaults.config
```diff
@@ -0,0 +1,26 @@
...
+db-channel-enabled = true
+db-tx-end = "rollback-allow-override"
+db-prepared-statements = false
+server-host = "0.0.0.0"
+server-port = 80
+server-unix-socket = "/tmp/pgrst_io_test.sock"
+server-unix-socket-mode = "777"
+openapi-server-proxy-uri = "https://postgrest.org"
+jwt-secret = "c2VjdXJpdHl0aHJvdWdob2JzY3VyaXR5"
+jwt-aud = "postgrest.org"
```
wolfgangwalther, line 17: I tried using `https://postgrest.org` here, because I understood https://tools.ietf.org/html/rfc7519#section-4.1.3 (that's what we are linking in the docs) to support either strings or URIs. But this results in an error when starting postgrest, because the `Crypto.JWT` parser for `StringOrUri` tells me, that this is not a valid URI.

How do I add a URI here? If somebody knows how, we could add a hint to the docs. If not for anyone else, it would at least help me :).

wolfgangwalther:
Lot's of haskell again, @steve-chavez. Looking forward to your feedback! I already learned a bit from `hlint` :D.

wolfgangwalther:
Cirrus seems to be hitting a build timeout of 15 minutes now, saw that for a couple of commits in different PRs now. @steve-chavez you added a timeout of 30 minutes a little time ago. Did they lower the timeout or what's happening?

src/PostgREST/Config.hs
```diff
@@ -62,9 +65,13 @@ import PostgREST.Types            (JSPath, JSPathExp (..),
...
 import Protolude                  hiding (concat, hPutStrLn,
                                    intercalate, null, replace, take,
-                                   toS, (<>))
+                                   toList, toLower, toS, (<>))
 import Protolude.Conv             (toS)
 
+-- | Command line interface options
+data CLI = CLI
+  { cliDumpConfig :: Bool
+  , cliPath       :: FilePath }
```
monacoremo, line 74: As we are hoing to add options here that are mutually exclusive (e.g. dump schema cache), can we make this s sym type instead of bool?

monacoremo: E.g. `data CliMode = Run | DumpConfig` (`| DumpSchemaCache` etc.)

steve-chavez:
> Cirrus seems to be hitting a build timeout of 15 minutes now, saw that for a couple of commits in different PRs now

@wolfgangwalther Hmm.. are you sure it's a timeout? It looks like it's failing because of low memory: https://github.com/haskell/cabal/issues/5505.

Maybe we should try adding `-j1` to cirrus(I remember `-j1` didn't work before, but might work now).

wolfgangwalther:
> @wolfgangwalther Hmm.. are you sure it's a timeout? It looks like it's failing because of low memory: [haskell/cabal#5505](https://github.com/haskell/cabal/issues/5505).

Absolutely, the low memory makes sense. I did not look carefully - I just figured it was not a real build error. I had 3 cirrus builds failing at 15 minutes, so I was a bit too quick to assume it was a timeout thing.

> **Edit**: The`-j1` seems to have worked. I added the change on #1600 (was also getting the cirrus error here).

Sweet, so I won't look at this here and just see green where red is :)

src/PostgREST/Config.hs
wolfgangwalther: Very good point, Remo. I think `--dump-config` as an option is wrong to start with, when those things are mutually exclusive.

Two ideas to make this better:
- Use `--dump=config` and `--dump=schema` - this is mutually exclusive from the start. Then have a `CliDump = DumpConfig | DumpSchema | DumpNothing`.
- Use a "sub command" based approach. So have subcommands like `postgrest run`, `postgrest dump-config` and `postgrest dump-schema`. Of course the default without command would be `run`. Those commands would be mutually exclusive as well by design. Then have `CliCmd = CmdRun | CmdDumpConfig | CmdDumpSchema` like you suggested. This would be a bit more flexible in case we're adding other commands in the future. Not sure how easy the "default to run without command" is with the cli parser.

I think I like the `--dump=...` option better. WDYT?

monacoremo: --dump-config if good I think! Most straight forward. Optparse applicative will be able to enforce a consistent command (see 'alternatives' where a sum type is the basis and 'flags' in https://github.com/pcapriotti/optparse-applicative)

wolfgangwalther: Ok, I went with a `--dump-config` flag and `data Command = CmdRun | CmdDumpConfig` for now.

test/io-tests/configs/no-defaults.config
steve-chavez: Weird.. I'm trying the `jwt-aud="https://postgrest.org"` now(on my set-configs branch) and it works.

steve-chavez: Though there doesn't seem to be any test with an URI.. I've also tried `jwt-aud="//"` and it doesn't fail. Maybe the aud is not being validated at all..

wolfgangwalther: I know that it's only treated as URI, if there is a `:` in it. And then it must be valid. I will try again with a valid and an invalid uri.

steve-chavez:
> The printed config is itself a valid postgrest config file.

I thought this would mean I could do `postgrest --dump-config > pgrst.conf` and generate a default config. But now I get it's mostly for tests. I'll give it a review now :)

test/io-tests/configs/no-defaults.config
wolfgangwalther: Or at least that's what I thought... I can confirm, any valid and invalid URL works right now.

src/PostgREST/Config.hs
```diff
@@ -225,6 +241,62 @@ readPathShowHelp = customExecParser parserPrefs opts
...
           |]
 
+-- | Dump the config
+dumpAppConfig :: AppConfig -> IO ()
+dumpAppConfig conf = do
+  putStrLn dump
+  exitSuccess
+
+  where
+    dump = intercalate "\n" $ (\(k, v) -> k <> " = " <> v) <$>
```
steve-chavez, line 251: I think [unlines](https://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text.html#v:unlines) could replace `intercalate` here?

test/io-tests.sh
steve-chavez: Hmm.. I also don't have a better name, maybe just put the distinction as a comment?

test/io-tests/configs/no-defaults.config
steve-chavez: I think we can open a new issue for this one.

wolfgangwalther:
> I thought this would mean I could do `postgrest --dump-config > pgrst.conf` and generate a default config. But now I get it's mostly for tests.

I think we're missing two things to be able to do that:
- the `FILENAME` argument must be optional. This is something that I have to change anyway to support "all-environment-only" configurations.
- with `--dump-config` enabled, postgrest should not throw when missing one of the required settings. This is something that could happen with the environment variables approach - not 100% yet. Depends on implementation of that.
- A good example config file needs comments, just like we provide them with the help output.

Maybe we can do that a bit different - what if we split the output of `--help` / the error output on stdout and stderr, so that only the example config file is on stdout and everything else on stderr? Then you should be able to do `postgrest --help > example.config` and have our example ready to go.

steve-chavez:
> what if we split the output of --help / the error output on stdout and stderr, so that only the example config file is on stdout and everything else on stderr? Then you should be able to do postgrest --help > example.config and have our example ready to go.

That sounds good! Seems like a simpler implementation.

wolfgangwalther:
> > what if we split the output of --help / the error output on stdout and stderr, so that only the example config file is on stdout and everything else on stderr? Then you should be able to do postgrest --help > example.config and have our example ready to go.
> 
> That sounds good! Seems like a simpler implementation.

Had a look at that.. and the implementation is not as easy as thought. The whole thing is dumped by `Options.Applicative`.

Implementation-wise we could just do the following:
- remove the example config from the options help output
- add another CLI flag `--dump-example-config` to print just that

I think that's also better to use. WDYT?

---

Anyway, this would be something for a different PR. I addressed the other comments, CI is passing -> merge time!

PR #1679, opened 2020-12-05
Rename config settings with fallback aliases

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/pull/1668#issuecomment-732354557 this changes names for some of the config options:
- secret-is-base64 -> jwt-secret-is-base64
- role-claim-key -> jwt-role-claim-key
- max-rows -> db-max-rows
- pre-request -> db-pre-request
- root-spec -> db-root-spec

I also changed db-schema to db-schemas, because that's how it was used internally. I guess this was changed once multi schema support was included.

All of the old names are added as aliases, so there is no breaking changes - existing configs continue to work.

The first commit ("changed: ...") is the renaming of config options (trivial) + a lot of sorting, so that config options with the same prefix are now nicely grouped everywhere. Everywhere except the `--help` output that is, because this follows a slightly different logic, so I did not touch that again. This 3rd commit creates a big diff because of all the sorting, but just compiling and running all the tests should give enough confidence. This will not need a deep review.

The second commit ("added: ...") is the code for the aliases. This is new code and should be reviewed properly.


steve-chavez:
LGTM!

Just tried omitting `db-schemas`. And it shows:

```
Error parsing config file:
        missing key: either db-schemas or db-schema must be set
```

Which is great! I also like the new `reqWithAlias` and `optWithAlias` functions.

steve-chavez (approved):

PR #1680, opened 2020-12-05
Improve parsing of boolean config values

### LINKED ISSUE #1572: Passing Bool parameters to docker images is a bit broken

wolfgangwalther:
Resolves #1572.

Boolean config values can now set with all kinds of:
- strings: `True`, `true`, `TRUE`, `"true"` - non-alphanumeric characters are stripped away and the remainings formatted in Title-case to be parsed by `readMaybe`.
- but also integers: `0` is false, all other integers are true.

This allows to set environment variables like this as well:
- `PGRST_DB_CHANNEL_ENABLED=true`
- `PGRST_DB_CHANNEL_ENABLED="true"`
- `PGRST_DB_CHANNEL_ENABLED=1`


src/PostgREST/Config.hs
```diff
@@ -417,7 +419,12 @@ readAppConfig cfgPath = do
 
     coerceBool :: C.Value -> Maybe Bool
     coerceBool (C.Bool b)   = Just b
-    coerceBool (C.String b) = readMaybe $ toS b
+    coerceBool (C.String s) =
+      -- parse all kinds of text: True, true, TRUE, "true", ...
+      case readMaybe . toS $ toTitle $ filter isAlpha $ toS s of
```
steve-chavez, line 424: This one is really nice @wolfgangwalther. Good to merge!

steve-chavez:
@wolfgangwalther LGTM! By the way, I thought of reminding you about the CHANGELOG. However, I see you're already thinking about generating it from the commits. Nice!

A nitpick though :grin: , perhaps we should use "fix" instead of "fixed" for the message format? To be more in line with [semantic release](https://github.com/semantic-release/semantic-release#commit-message-format)?

wolfgangwalther:
> @wolfgangwalther LGTM! By the way, I thought of reminding you about the CHANGELOG. However, I see you're already thinking about generating it from the commits. Nice!
> 
> A nitpick though grin , perhaps we should use "fix" instead of "fixed" for the message format? To be more in line with [semantic release](https://github.com/semantic-release/semantic-release#commit-message-format)?

Yes, that's something I would have brought up as well. Good timing here, though. So as I see it we have two choices:
1. either do commit messages according to our current changelog format (added, fixed, changed)
2. or do them according to semantic release format (feat, fix, BREAKING CHANGE)

If we stick close to our current changelog, "changed:" would be a direct mapping to "Changed".

---

If we go semantic release format, then:

I think added/feat and fixed/fix match. But I am not sure about "changed" and BREAKING CHANGES. That is, I am not sure about "changed" in our current changelog at all. What does it mean? Is "Changed" about breaking changes?

If we were going full semantic release format, we would probably do something like:
- feat: added config file
- fix: improve parsing of boolean config values
- feat: configuration can be read directly from environment
  BREAKING CHANGE: whatever breaking change

This would result in:

```
Added:
  added config file
  configuration can be read directly from environment

Fixed:
  improve parsing of boolean config values

Changed:
  whatever breaking change
```

So 1 commit message with a breaking change would be split up to both added/fixed and changed sections?

And then I would parse the commit message for any references like `#1234` and add those like we currently do.

WDYT?

steve-chavez:
> That is, I am not sure about "changed" in our current changelog at all. What does it mean? Is "Changed" about breaking changes?

Yes, they're breaking changes.

> WDYT?

Sounds good!

wolfgangwalther:
> Yes, they're breaking changes.

Ok, I think I was confused because of this:

```
Changes
#1522, #1528, #1535, Docker images are now built from scratch based on a the static PostgREST executable (#1494) and with Nix instead of a Dockerfile. This reduces the compressed image size from over 30mb to about 4mb - @monacoremo
```

Although it is technically a "change", it's not a breaking change, right? Maybe I'm just failing to see the breaking aspect of it :D.

It is more of a new feature (reduced image size) and should be in "Added"?

monacoremo:
> Although it is technically a "change", it's not a breaking change, right? Maybe I'm just failing to see the breaking aspect of it :D.
> 
> It is more of a new feature (reduced image size) and should be in "Added"?

I'd argue for 'change', as this didn't add any new functionality, and removed some unneeded features that some might still have relied on (e.g. running bash in the Docker image)

wolfgangwalther:
We should really rename "Changes" to "Breaking Changes".

> I'd argue for 'change', as this didn't add any new functionality,

If we're talking about "semantic release" semantics, this is irrelevant, I think. Reducing the image size is a feature, because it is not a bug fix.

> and removed some unneeded features that some might still have relied on (e.g. running bash in the Docker image)

I see that point. The question is: What constitutes a breaking change? Clearly not everything that one is able to do with the current version, because then every bugfix would be a breaking change as it changes behaviour of the software. I don't think we ever advertised bash inside the docker container as one of the features of PostgREST.

Does the smaller docker image alone make a major version bump from 7.x to 8.x neccessary?

I don't think so.

wolfgangwalther:
And we should rename "Added" to "Features" or so, because I feel like this is really where the confusion comes from. I fully agree that it feels strange to put the smaller docker image under "Added".

PR #1682, opened 2020-12-06
Migrate io-tests from bash to pytest

monacoremo:
As discussed in https://github.com/PostgREST/postgrest/pull/1676#discussion_r536664505, the io-tests are pushing bash to its limits and migrating to a python setup might be much more maintainable

## Tests to migrate (Edit: updated from master)

* [x] ok 1 - dump of config file configs/aliases.config is valid
* [x] ok 2 - dump of config file configs/app-settings.config is valid
* [x] ok 3 - dump of config file configs/base64-secret-from-file.config is valid
* [x] ok 4 - dump of config file configs/boolean-numeric.config is valid
* [x] ok 5 - dump of config file configs/boolean-string.config is valid                                                                                                                      [0/1614]
* [x] ok 6 - dump of config file configs/dburi-from-file.config is valid
* [x] ok 7 - dump of config file configs/defaults.config is valid 
* [x] ok 8 - dump of config file configs/no-defaults.config is valid                                     
* [x] ok 9 - dump of config file configs/role-claim-key.config is valid
* [x] ok 10 - dump of config file configs/secret-from-file.config is valid                               
* [x] ok 11 - dump of config file configs/sigusr2-settings.config is valid                           
* [x] ok 12 - dump of config file configs/simple.config is valid                                         
* [x] ok 13 - dump of config file configs/unix-socket.config is valid                                
* [x] ok 14 - dump of config file configs/aliases.config does match expectation                          
* [x] ok 15 - dump of config file configs/boolean-numeric.config does match expectation
* [x] ok 16 - dump of config file configs/boolean-string.config does match expectation           
* [x] ok 17 - dump of config file configs/defaults.config does match expectation
* [x] ok 18 - dump of config file configs/no-defaults.config does match expectation
* [x] ok 19 - Succesfully connected through unix socket  
* [x] ok 20 - authentication with simple (no EOL) secret read from a file
* [x] ok 21 - authentication with simple secret read from a file
* [x] ok 22 - authentication with ASCII (no EOL) secret read from a file
* [x] ok 23 - authentication with ASCII secret read from a file
* [x] ok 24 - authentication with UTF-8 (no EOL) secret read from a file
* [x] ok 25 - authentication with UTF-8 secret read from a file 
* [x] ok 26 - authentication with binary secret read from a file
* [x] ok 27 - authentication with binary (+EOL) secret read from a file
* [x] ok 28 - authentication with Base64 (simple) secret read from a file
* [x] ok 29 - authentication with Base64 (ASCII) secret read from a file
* [x] ok 30 - authentication with Base64 (UTF-8) secret read from a file
* [x] ok 31 - authentication with Base64 (binary) secret read from a file
* [x] ok 32 - connection with (no EOL) dburi read from stdin / a file
* [x] ok 33 - connection with (EOL) dburi read from stdin / a file
* [x] ok 34 - request with ".postgrest.a_role" role-claim-key for {"postgrest":{"a_role":"postgrest_test_author"}} jwt: 200
* [x] ok 35 - request with ".customObject.manyRoles[1]" role-claim-key for {"customObject":{"manyRoles": ["other", postgrest_test_author"]}} jwt: 200
* [x] ok 36 - request with "."https://www.example.com/roles"[0].value" role-claim-key for {"https://www.example.com/roles":[{"value":"postgrest_test_author"}]} jwt: 200
* [x] ok 37 - request with ".myDomain[3]" role-claim-key for {"myDomain":["other","postgrest_test_author"]} jwt: 401
* [x] ok 38 - request with ".myRole" role-claim-key for {"role":"postgrest_test_author"} jwt: 401
* [x] ok 39 - invalid jspath "role.other": rejected
* [x] ok 40 - invalid jspath ".role##": rejected
* [x] ok 41 - invalid jspath ".my_role;;domain": rejected
* [x] ok 42 - invalid jspath ".#$%&$%/": rejected
* [x] ok 43 - invalid jspath "": rejected
* [x] ok 44 - invalid jspath "1234": rejected
* [x] ok 45 - iat claim accepted
* [x] ok 46 - GET /rpc/get_guc_value: "0123456789abcdef"
* [x] ok 47 - app.settings.name_var config reloaded with SIGUSR2
* [x] ok 48 - jwt-secret config reloaded with SIGUSR2 
* [x] ok 49 - db-schemas config reloaded with SIGUSR2 

## For later PRs:

 * make the postgrest endpoint dynamic/switch to unix domain sockets by default to allow parallelization 
 * include https://pypi.org/project/pytest-parallel/
 * add docs for advanced usage (`postgrest-watch postgrest-test-io -k ...` etc.)


wolfgangwalther:
Ah, the first test is already so much nicer compared to the bash script.

You might want to consider using [pathlib](https://docs.python.org/3/library/pathlib.html). Specifically the following could make stuff even easier to read:
- https://docs.python.org/3/library/pathlib.html#operators
- https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir
- https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

monacoremo:
> You might want to consider using [pathlib](https://docs.python.org/3/library/pathlib.html).

Yes, will be useful to handle the config files! I will also properly use pytest fixtures, avoiding the iteration in the config test function https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures

monacoremo:
I'll add more tests from the original bash file as I can, but probably not before next weekend - so marking as ready to review, please feel free to merge if you want to build on top of this scaffolding. When all tests are migrated, the can remove the old bash file.

test/io-tests/test_io.py
```diff
@@ -0,0 +1,129 @@
...
+    command = ['postgrest', configpath]
+    process = subprocess.Popen(command, stdin=subprocess.PIPE)
+
+    try:
+        if stdin:
+            process.stdin.write(stdin)
+        process.stdin.close()
+
+        waitfor200(BASEURL)
+        yield BASEURL
```
wolfgangwalther, line 63: If `BASEURL` was created dynamically, we could run the tests in parallel, right? Could speed them up. Irrelevant for CI, because they are quite short, but for `postgrest-watch` it'd be nice.

wolfgangwalther: Not sure if pytest can run them in parallel, easily?

monacoremo: Yes - we could even run them via temporary sockets. We would need to change all config files and adjust io-tests.sh a bit, so I'm savin that for last :-)

monacoremo: Yes, there's a plugin for that - easy!

wolfgangwalther: We do have a test for connecting through sockets in place. So if all of the tests are changed to use sockets by default (which I think is good!), we should swap that test around to test connection via localhost.

wolfgangwalther:
CI currently shows the tests as:

```
============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/circleci/project
collecting ... collected 24 items                                                             

test/io-tests/test_io.py ........................                        [100%]

============================== 24 passed in 2.72s ==============================
``` 

Can we get an output with each test-case listed here as well? That would be more green :D



> I'll add more tests from the original bash file as I can, but probably not before next weekend - so marking as ready to review, please feel free to merge if you want to build on top of this scaffolding. When all tests are migrated, the can remove the old bash file.

Looks very good so far. I don't think we need to hurry, but can merge this in one go when it's done. I have added another config io test to the bash script locally already - but that's written for now, so we can rewrite it when we're there.

test/io-tests/test_io.py
monacoremo: Yes, good point! And that one we can also easily put on a random free port, so no more hardcoded and non-parallelizable endpoints

test/io-tests/test_io.py
```diff
@@ -0,0 +1,175 @@
...
+
+    jwt = secretpath.with_suffix('.jwt').read_text()
+    headers = {'Authorization': f'Bearer {jwt}'}
+
+    with run(configfile, stdin=secret) as url:
+        response = requests.get(f'{url}/authors_only', headers=headers)
+        assert response.status_code == 200
+
+
+def test_read_dburi_from_file_withouteol():
```
wolfgangwalther, line 158:
```suggestion
def test_read_dburi_from_file_without_eol():
```

easier to read

test/io-tests/test_io.py
```diff
@@ -0,0 +1,175 @@
...
+    with run(dburifromfileconfig, stdin=dburi.encode('utf-8')) as url:
+        response = requests.get(f'{url}/')
+        assert response.status_code == 200
+
+
+def test_read_dburi_from_file_witheol():
```
wolfgangwalther, line 164:
```suggestion
def test_read_dburi_from_file_with_eol():
```

same

test/io-tests/test_io.py
```diff
@@ -0,0 +1,175 @@
...
+basedir = pathlib.Path(os.path.realpath(__file__)).parent
+configs = [path for path in (basedir / 'configs').iterdir() if path.is_file()]
+expectedconfigs = list((basedir / 'configs' / 'expected').iterdir())
+secrets = [path for path in (basedir / 'secrets').iterdir() if path.suffix != '.jwt']
+dburi = os.getenv('POSTGREST_TEST_CONNECTION')
+dburifromfileconfig = basedir / 'configs' / 'dburi-from-file.config'
+roleclaimkeyconfig = basedir / 'configs' / 'role-claim-key.config'
+
+invalidroleclaimkeys = [
+        'role.other',
```
wolfgangwalther, line 25: What do you think about 2 spaces instead of 4 spaces indentation across the whole python file? I think that's what we're using across the other files in the project.

test/io-tests/test_io.py
monacoremo: you're right, fixed!

test/io-tests/test_io.py
monacoremo: Ah, I was going to add automatic formatting with e.g. [black](https://black.readthedocs.io/en/stable/) (just added in last commit), which enforces the Python standard of [four spaces](https://www.python.org/dev/peps/pep-0008/#indentation). For Python code I think it's better when we stick to that standard, and postgrest-style will now make sure that we stick to that no matter how contributors add new code. WDYT?

wolfgangwalther: I don't think I would have even mentioned it, if it wasn't for the 8 spaces that the array items had in the spot I commented on. I see black has made that 4 spaces only by pulling the `]` out one level. This satisfies my eye enough - I won't object to this :).

test/io-tests/test_io.py
```diff
@@ -0,0 +1,220 @@
...
+        assert response.status_code == 200
+
+
+def test_read_dburi_from_file_with_eol():
+    with run(dburifromfileconfig, stdin=dburi + b"\n") as url:
+        response = requests.get(f"{url}/")
+        assert response.status_code == 200
+
+
+def test_role_claim_key(roleclaimcase):
```
wolfgangwalther, line 206: I'm trying to review the new tests as they come in and realized it's hard to look at one test with everything that goes into it. Currently the file is organized as:

```
var1 =
var2 = 
var3 =

fixture1
fixture2
fixture3

test1 
test2
test3
```

Now that the file is getting longer, I need to scroll more and more to get one test together. I think most the variables and fixtures belong to a small group of tests. What do you think about the following?

```
var1 =
fixture1
test1

var2 =
fixture2
test2

var3 =
fixture3
test3
```

A lot less scrolling needed to look at those tests!

monacoremo: Yes, will refactor as everything comes together!

test/io-tests/test_io.py
```diff
@@ -0,0 +1,242 @@
...
+    with run(roleclaimkeyconfig, moreenv=env) as url:
+        response = requests.get(f"{url}/authors_only", headers=headers)
+        assert response.status_code == roleclaimcase.expected_status
+
+
+def test_invalid_role_claim_key(invalidroleclaimkey):
+    env = {"ROLE_CLAIM_KEY": invalidroleclaimkey}
+
+    with pytest.raises(subprocess.CalledProcessError):
+        dumpconfig(roleclaimkeyconfig, moreenv=env)
```
wolfgangwalther, line 221: is there any way we can show the dumped config here in case of a test failure?

i.e. when an incorrect role claim key does not throw an error and is parsed as correct - it would be helpful to see the dumped config, because this would show how the role claim key was parsed and dumped again.

wolfgangwalther: And maybe I can extend the `--dump-config` to do the following:

- `--dump-config` prints the whole file, just like it does now

- `--dump-config="jwt-role-claim-key"` prints just this one setting.

This would make the output in case of a failure here much more useful. It would also help with the boolean test-cases I have added on master. They dump the whole config file, but it's actually just about 3 values. They can easily be lost in the big picture.

WDYT?

test/io-tests/test_io.py
```diff
@@ -226,17 +232,41 @@ def test_iat_claim():
...
+        url = f"{process.baseurl}/rpc/get_guc_value?name=app.settings.name_var"
+
+        response = requests.get(url)
+        assert response.status_code == 200
+        assert response.text == '"John"'
+
+        # change setting
+        configfile.write_text(config.replace("John", "Jane"))
+        # reload
+        process.process.send_signal(signal.SIGUSR2)
```
wolfgangwalther, line 268: `process.process` :D

Maybe rename `PostgrestProcess` to just `Postgrest` and `as process` to `as postgrest`. Then it would be `postgrest.process` and `postgrest.baseurl` ?

test/io-tests/test_io.py
```diff
@@ -157,7 +164,7 @@ def test_expected_config(expectedconfig):
     assert dumpconfig(basedir / "configs" / expectedconfig) == expected
 
 
-def test_stable_config(configpath):
+def test_stable_config(tmp_path, configpath):
```
wolfgangwalther, line 167: this is the first argument that is snake-cased. all others are just like `tmppath`. it seems like your idea was to only snake-case the test function names?

wolfgangwalther: but tbh: I don't really get where it actually comes from? Is this a pytest built-in fixture or something?

test/io-tests/test_io.py
monacoremo: You're reviewing faster than I can code :-) Yes, I'm trying to see what is needed and how it fits together, in the end several pieces will be simplified/streamlined!

test/io-tests/test_io.py
monacoremo: that's a pytest builtin fixture, have to use that name... I'll put some of the longer names in snake case later, too

monacoremo:
Being able to filter the tests is going to be a great usability improvement I think - e.g. `postgrest-watch postgrest-test-io-new -k reload`

monacoremo:
All tests covered! Now to clean it up, refactor and add some docs

test/io-tests/test_io.py
monacoremo: Fixed, this is what is printed when a role key is seen as valid:

```
========================================== FAILURES ===========================================
_____________________________ test_invalid_role_claim_key[.role] ______________________________

invalidroleclaimkey = '.role'

    @pytest.mark.parametrize("invalidroleclaimkey", FIXTURES["invalidroleclaimkeys"])
    def test_invalid_role_claim_key(invalidroleclaimkey):
        "Given an invalid role-claim-key, Postgrest should exit with a non-zero exit code."
        env = {"ROLE_CLAIM_KEY": invalidroleclaimkey}
                                               
        with pytest.raises(PostgrestError):
>           print(dumpconfig(CONFIGSDIR / "role-claim-key.config", moreenv=env))
E           Failed: DID NOT RAISE <class 'test_io.PostgrestError'>

test/io-tests/test_io.py:235: Failed
------------------------------------ Captured stdout call -------------------------------------
db-uri = "postgresql://postgres?host=/run/user/1001/tmp.iZohhdd0bV/socket&user=postgrest_test_authenticator"
db-schema = "test"                             
db-anon-role = "postgrest_test_anonymous"
db-pool = 1                                    
db-pool-timeout = 10                           
db-extra-search-path = "public"
db-channel = "pgrst"                           
db-channel-enabled = false
db-tx-end = "commit"                           
db-prepared-statements = true
server-host = "127.0.0.1"
server-port = 49421                            
server-unix-socket = ""
server-unix-socket-mode = "660"
openapi-server-proxy-uri = ""
jwt-secret = "reallyreallyreallyreallyverysafe" 
jwt-aud = ""                                   
secret-is-base64 = false
role-claim-key = ".\"role\""
max-rows = ""                                  
pre-request = ""                               
root-spec = ""                                 
raw-media-types = ""                           
log-level = "error"                            

=================================== short test summary info ===================================
FAILED test/io-tests/test_io.py::test_invalid_role_claim_key[.role] - Failed: DID NOT RAISE ...
================================ 1 failed, 43 passed in 12.66s ================================
```

monacoremo: I think it's easier to add the filtering on the tests side - let me check

wolfgangwalther: I see what you mean. Have a look at those files, though:
- https://github.com/PostgREST/postgrest/blob/master/test/io-tests/configs/boolean-numeric.config
- https://github.com/PostgREST/postgrest/blob/master/test/io-tests/configs/expected/boolean-numeric.config

The only values I care about in the expected output are the lines that are given in the config. All the others are just default values and noise.

Maybe we can change the tests that test for "expected" config to make sure that all lines in the expectation file are there, but allow additional lines in the output?

wolfgangwalther:
I had a look at the merge conflict you have with io-tests. I think the only commit that changed io-tests.sh after you branched off was ed58511de3ab964035401f8329b6e06ac99ef36b. This was just a rename from `db-schema` to `db-schemas`, so once you do that in the `test_db_schema_reload` you should be fine.

monacoremo:
I rebased to make sure, should be good now I think! In a last commit I'll remove the old io-tests

monacoremo:
Travis is acting up again, but we shouldn't have impacted anything that concerns it. Ready for review/merge :-)

wolfgangwalther:
> Travis is acting up again, but we shouldn't have impacted anything that concerns it.

Actually, we did. Travis runs the code-coverage currently and that includes the io-tests. They fail with:

```
Did not find executable at specified path: test/io-tests.sh
```

You could have a look at that and see how hard it would be to install those python deps there to be able to run them - or not. I wanted to move the code coverage to a nix tool and to circleci anyway, so that might be the time to do it...

test/io-tests/test_io.py
```diff
@@ -0,0 +1,348 @@
...
+        response = session.get(url, headers=headers)
+        assert response.status_code == 404
+
+        # change setting
+        configfile.write_text(
+            config.replace('db-schemas = "test"', 'db-schemas = "test, v1"')
+        )
+        # reload
+        postgrest.process.send_signal(signal.SIGUSR2)
+        postgrest.process.send_signal(signal.SIGUSR1)
```
wolfgangwalther, line 345: maybe a comment explaining why we send `SIGUSR1` here as well (the test is about `SIGUSR2` after all?) could be added. and it should be separated from `# reload`.

`SIGUSR2` reloads the config file - `SIGUSR1` then reloads the schema cache to verify that the config reload actually happenend. So one is what we want to test, while the other is just boilerplate.

test/io-tests/test_io.py
```diff
@@ -0,0 +1,348 @@
...
+    'configs/expected' directory.
+
+    """
+    expected = expectedconfig.read_text()
+    assert dumpconfig(CONFIGSDIR / expectedconfig.name) == expected
+
+
+@pytest.mark.parametrize(
+    "config",
+    [conf for conf in CONFIGSDIR.iterdir() if conf.is_file()],
```
wolfgangwalther, line 142: this iterates through all files in the `configs` directory. `io-tests.sh` only interated through files with `.config` extension. In #1691 I added a file `configs/no-defaults.env`. This is a configuration, so the `configs/` dir is right, but it sets environment variables because the files not read in as a config file.

It'd be better to use the same filtering here, I think.

test/io-tests/fixtures.yaml
```diff
@@ -0,0 +1,23 @@
+roleclaims:
+  - key: ".postgrest.a_role"
+    data: {"postgrest": {"a_role": "postgrest_test_author"}}
+    expected_status: 200
+  - key: ".customObject.manyRoles[1]"
+    data: {"customObject": {"manyRoles": ["other", "postgrest_test_author"]}}
+    expected_status: 200
+  - key: '."https://www.example.com/roles"[0].value'
+    data: {"https://www.example.com/roles": [{"value": "postgrest_test_author"}]}
```
wolfgangwalther, line 9: what about using only single quotes in this yaml file? more consistent, because of line 8, where we need double quotes inside the string once.

test/io-tests/test_io.py
```diff
@@ -0,0 +1,348 @@
...
+        for line in dump.split("\n"):
+            if "role-claim-key" in line:
+                print(line)
+
+
+def test_iat_claim(session):
+    """
+    A claim with an 'iat' (issued at) attribute should be successful.
+
+    The PostgREST time cache lead to issues here, see:
```
wolfgangwalther, line 245:
```suggestion
    The PostgREST time cache leads to issues here, see:
```

test/io-tests/test_io.py
```diff
@@ -0,0 +1,348 @@
...
+    configfile = tmp_path / "test.config"
+    configfile.write_text(config)
+
+    headers = jwtauthheader({"role": "postgrest_test_author"}, SECRET)
+
+    with run(configfile) as postgrest:
+        url = f"{postgrest.baseurl}/authors_only"
+
+        response = session.get(url, headers=headers)
+        assert response.status_code == 401
```
wolfgangwalther, line 314: This was not part of the test in `io-tests.sh` - it's a nice improvement!

wolfgangwalther:
> @monacoremo travis shot in the dark 

Nice shot. I never make those on the first try. Nor the 2nd. :D

So I went through everything twice. Once just looking at the python stuff, once comparing the old io tests to the new ones. Everything is correct, only nit-picks and small improvements suggested.

Great job, Remo!

monacoremo:
> Nice shot. I never make those on the first try. Nor the 2nd. :D

I'm super surprised myself! :D 😎

Thank you for the great comments! Will add a few fixes

test/io-tests/test_io.py
```diff
@@ -0,0 +1,348 @@
...
+@pytest.mark.parametrize(
+    "roleclaim", FIXTURES["roleclaims"], ids=lambda claim: claim["key"]
+)
+def test_role_claim_key(session, roleclaim):
+    "Authorization should depend on a correct role-claim-key and JWT claim."
+    env = {"ROLE_CLAIM_KEY": roleclaim["key"]}
+    headers = jwtauthheader(roleclaim["data"], SECRET)
+
+    with run(CONFIGSDIR / "role-claim-key.config", moreenv=env) as postgrest:
+        response = session.get(f"{postgrest.baseurl}/authors_only", headers=headers)
```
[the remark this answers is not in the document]
monacoremo: Great idea!

wolfgangwalther:
Not sure if you already noticed: The reload tests fail kind of randomly. One test here, one test there. Maybe we need a sleep after the send_signal or something?

monacoremo:
Yes, just noticed also - seems like we are going too fast. I added short sleeps in the affected tests

wolfgangwalther:
Hmm. We still have an issue with the code coverage though. It's showing only 75%. Compare those two reports:

- master: https://coveralls.io/github/PostgREST/postgrest
- this pr: https://coveralls.io/builds/35672595

This PR is missing all the files that are collected by running the IO tests.

So it seems like the python IO test script is not launching the instrumented executable?

wolfgangwalther:
Hm. Could be some stack issue as well. In the travis coverage job there is a huge debug output from stack - no idea what it means.

Later it also shows this:

```
  # merge the results from `stack test` and the io tests and exclude Paths_postgrest
  stack --no-terminal exec hpc -- sum --union --exclude=Paths_postgrest --output=/tmp/all.tix $_HPC_DIR/combined/all/all.tix test/io-tests/postgrest.tix

  # fix a bug in stack-hpc-coveralls
  mv $_MIX_DIR/hpc/Main.mix $_MIX_DIR/hpc/$_PKG_NAME/Main.mix
  mv $_MIX_DIR/hpc/UnixSocket.mix $_MIX_DIR/hpc/$_PKG_NAME/UnixSocket.mix
  sed -i -r "s/(Main|UnixSocket)/$_PKG_NAME\/\1/g" /tmp/all.tix

  # upload to coveralls
  mv /tmp/all.tix $_HPC_DIR/combined/all/all.tix
  shc combined all

hpc: user error (Pattern match failure in do expression at utils/hpc/HpcCombine.hs:127:3-14)
sed: can't read /tmp/all.tix: No such file or directory
mv: cannot stat '/tmp/all.tix': No such file or directory
```

This script is brittle - part of the reason why I want to switch to a different setup with nix etc.

The first command fails. I would assume it's because the test/io-tests/postgrest.tix file is not there. The simple reason could be that it's just in some other location now. The `io-tests.sh` script changed folder to test/io-tests - so that was the working directory.

The file is probably just in the repo root now.

monacoremo:
> So it seems like the python IO test script is not launching the instrumented executable?

It should also be using the postgrest executable on the path... It might have been an issue with passing the flags, let's see if this works

wolfgangwalther:
I think this should do it:

```
stack --no-terminal exec hpc -- sum --union --exclude=Paths_postgrest --output=/tmp/all.tix $_HPC_DIR/combined/all/all.tix test/io-tests/postgrest.tix
``` 
to
```
stack --no-terminal exec hpc -- sum --union --exclude=Paths_postgrest --output=/tmp/all.tix $_HPC_DIR/combined/all/all.tix postgrest.tix
```

monacoremo:
Yes, had the same thought!

...aand we're back randomly poking at travis again :D

wolfgangwalther:
> ...aand we're back randomly poking at travis again :D

I'm glad about that. So much fun. Some things don't change.

wolfgangwalther:
That worked. Even though the coverage is not shown now.. but that is a different bug. It's happening in other PRs as well, I just realized. Basically coverage is disabled right now...

At least this PR doesn't report a lower number, now. Just none at all, like the others :D

Fine for now.

steve-chavez:
@monacoremo @wolfgangwalther Great work here guys! Liking the new python test suite.

One thing though. When running `postgrest-test-io`, I get a ReadTimeout error from time to time:

```
requests.exceptions.ReadTimeout: HTTPConnectionPool(host='127.0.0.1', port=49421): Read timed out. (read timeout=0.1)

/nix/store/cnp0f7y76izc2kpl6xfkmjwjq6i71cxd-python3-3.8.5-env/lib/python3.8/site-packages/requests/adapters.py:529: ReadTimeout
```

I also get a large stack trace, which I've pasted here: https://pastebin.com/bCqTbvbx

Could that be improved to make the test runs more deterministic?

wolfgangwalther:
If I remember correctly, that would be the `1s` timeout that is currently in place when waiting for `postgrest` to startup. Should probably just increase it. @monacoremo?

Issue #1352, opened 2019-07-15
Issue with Foreign Data Wrapper extension(multicorn) when PostgREST discovers the database


Remi-C:
### Environment

* PostgreSQL version: 
PostgreSQL 11.2 (Ubuntu 11.2-1.pgdg18.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0, 64-bit|
* PostgREST version: 
6.0
* Operating system:
Ubuntu 18.04
### Description of issue
Issue using postgREST with a database using the `multicorn` extension, which is a meta foreign data wrapper.
Regardless if a foreign table is in the postgREST exposed schema or not, postgREST crash-loop. 
 - use the postgres extension multicorn
 - create a foreign data wrapper using multicorn (so first a server then a fdw).
 - attempting to start postgREST results in a nasty infinite loop :
```
    Attempting to connect to the database...
    Failed to query the database. Retrying.
    {"hint":"No function matches the given name and argument types. You might need to add explicit type casts.","details":null,"code":"42883","message":"function multicorn_check_plpython3u() does not exist"}
```
    Of course the function public.multicorn_check_plpython3u() do exist and I use a super user to connect with postgREST to eliminate any chance of improper privilege.
   

The reason of the bug is that postgREST execute a bunch of SQL queries to discover the relations in the database:
[code here](https://github.com/PostgREST/postgrest/blob/master/src/PostgREST/DbStructure.hs)
One of this query calls `pg_relation_is_updatable(c.oid::regclass, FALSE)`,
which in turn calls the foreign data wrapper check function `multicorn_check_plpython3u()`.

This function is installed in  the `public` schema, but at this point, PostgREST can not access to the public schema.

The workaround I found for the moment is to either 
 - install `multicorn` in the `pg_catalog` schema (bad idea)
 - add to the `pg_catalog` the 3 multicorn check functions (`multicorn_check_plpython3u(), ...`)

It would be nice to have some control about what the search_path looks like during the postgREST discovery phase (another parameter?).

Issue was discussed with @ruslantalpa and @steve-chavez

steve-chavez:
You know, I have used [tds_fdw](https://github.com/tds-fdw/tds_fdw) with postgrest before and it didn't pose an issue, so maybe it could be a bug in multicorn.

I'll try installing multicorn next and see what I can find.

Remi-C:
[...]

Issue #1287, opened 2019-04-29
PostgREST fails to start with certain types of foreign tables in different schema


bbarany:
### Environment

* PostgreSQL version:  PostgreSQL 11.2 (Debian 11.2-2) on x86_64-pc-linux-gnu
* PostgREST version: 5.2
* Operating system: Debian testing

### Description of issue

I have a database with a schema (not the api schema) that has foreign tables with the type ogr_fdw. 

I created the api_auth and api_anon roles and the api schema, set up permissions etc. I can connect to the database with api_auth and can select from a test table.

However, postgrest (configured for login with api_auth, using the api_anon role and the api schema) fails to start up, instead it throws the following error repeatedly until terminated:

> Failed to query the database. Retrying.
> {"hint":null,"details":null,"code":"XX000","message":"[1] Server is read-only WFS; no WFS-T feature advertized"}
> 

PostgreSQL log:

> 2019-04-29 14:55:15.235 CEST [1325] api_auth@lww19 ERROR:  [1] Server is read-only WFS; no WFS-T feature advertized
> 2019-04-29 14:55:15.235 CEST [1325] api_auth@lww19 STATEMENT:  
>             SELECT
>               n.nspname AS table_schema,
>               c.relname AS table_name,
>               NULL AS table_description,
>               c.relkind = 'r' OR (c.relkind IN ('v','f'))
>               AND (pg_relation_is_updatable(c.oid::regclass, FALSE) & 8) = 8
>               OR (EXISTS
>                 ( SELECT 1
>                   FROM pg_trigger
>                   WHERE pg_trigger.tgrelid = c.oid
>                   AND (pg_trigger.tgtype::integer & 69) = 69) ) AS insertable
>             FROM pg_class c
>             JOIN pg_namespace n ON n.oid = c.relnamespace
>             WHERE c.relkind IN ('v','r','m','f')
>               AND n.nspname NOT IN ('pg_catalog', 'information_schema')
>             GROUP BY table_schema, table_name, insertable
>             ORDER BY table_schema, table_name 
> 2019-04-29 14:55:15.235 CEST [1325] api_auth@lww19 LOG:  could not send data to client: Broken pipe
> 2019-04-29 14:55:15.235 CEST [1325] api_auth@lww19 FATAL:  connection to client lost
> CPLDestroyMutex: Error = 16 (Device or resource busy)
> 

I understand the the problem is probably with the FDW, it shouldn't throw an error when querying the updatability of the relation. 

[...]

PR #1685, opened 2020-12-09
fix: retry connection on failed schema cache load

steve-chavez:
When the "Failed to load the schema cache" error happens, the connection is not retried. This can be reproduced on https://github.com/PostgREST/postgrest/compare/master...steve-chavez:failed-to-load-schema.

This error is actually a regression that happened on #1542(check this [deleted comment](https://github.com/PostgREST/postgrest/pull/1542/files#diff-aefb86a77d8df9fc673b38b36a8d87d1eb9c6e16ba2cfc2bb3b807d7f005df61L61-L62)).

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -35,7 +35,8 @@ import PostgREST.DbStructure (getDbStructure, getPgVersion)
 import PostgREST.Error       (PgError (PgError), checkIsFatal,
                               errorPayload)
 import PostgREST.Types       (ConnectionStatus (..), DbStructure,
-                              PgVersion (..), minimumPgVersion)
+                              PgVersion (..), SCacheStatus (..),
+                              minimumPgVersion)
 import Protolude             hiding (hPutStrLn, head, toS)
 import Protolude.Conv        (toS)
 
@@ -203,8 +204,11 @@ connectionWorker mainTid pool refConf refDbStructure refIsWorkerOn (dbChannelEna
         NotConnected                -> return ()                                     -- Unreachable because connectionStatus will keep trying to connect
         Connected actualPgVersion   -> do                                            -- Procede with initialization
           putStrLn ("Connection successful" :: Text)
-          loadSuccess <- loadSchemaCache pool actualPgVersion refConf refDbStructure
-          unless loadSuccess work
+          scStatus <- loadSchemaCache pool actualPgVersion refConf refDbStructure
+          case scStatus of
+            SCLoaded    -> pure ()            -- do nothing and proceed if the load was successful
+            SCOnRetry   -> work               -- retry
+            SCFatalFail -> killThread mainTid -- die if our schema cache query has an error
           liftIO $ atomicWriteIORef refIsWorkerOn False
 
 {-|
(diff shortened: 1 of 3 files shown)


steve-chavez:
Ideally we'd have "recovery tests" that could detect these errors. But they're a bit complicated - need to stop/start pg, recover with SIGUSR1 or NOTIFY, check pgrest is alive, etc. Maybe we can look at how to do them after #1682 is merged.

So, I think this one should be good to merge as it is.

wolfgangwalther:
In https://github.com/PostgREST/postgrest/pull/1632#issuecomment-740629768 I assumed that there is no error message when one of the db schema scripts fails. I just tested with this branch by introducing a syntax error in the pfkSourceColumns query - the error is reported. It doesn't look like this changed here, so I am not sure what happened back when I did not get any errors.

But this also leads to flooding my console:

```
Connection successful
{"hint":null,"details":null,"code":"42601","message":"syntax error at or near \"contype\""}
Failed to load the schema cache
Connection successful
{"hint":null,"details":null,"code":"42601","message":"syntax error at or near \"contype\""}
Failed to load the schema cache
Connection successful
{"hint":null,"details":null,"code":"42601","message":"syntax error at or near \"contype\""}
Failed to load the schema cache
Connection successful
{"hint":null,"details":null,"code":"42601","message":"syntax error at or near \"contype\""}
Failed to load the schema cache
...
```
and so on.

I'm not sure we actually want to retry like this? This seems like an error that it's not recoverable from. I think all SQL errors in dbStructure are errors that should not be there at all and are not caused by the user or the environment. They are bugs in PostgREST, so unlikely to change by restarting.

Are there any errors that can be recovered from? I think connection errors are caught before even trying the query, right?

I feel like we should exit on an error like this. Or at least debounce the retry (we do this for the database connection, right?).

steve-chavez:
@wolfgangwalther I think a check for a syntax error can be added here: 

https://github.com/PostgREST/postgrest/blob/11d62a80105c21696e65b70acf07b6f47405ca68/src/PostgREST/Error.hs#L214-L220

I agree it'd be useful, reminds me of an error that happened some time ago with FDWs: https://github.com/PostgREST/postgrest/issues/1352 (not a syntax error though).

I'll check it out.

steve-chavez:
We should be careful in quitting on retrying the schema cache query.

- On https://github.com/PostgREST/postgrest/issues/1287, theres an `XX000` error that looks like it might be fatal to our query. But according to https://stackoverflow.com/a/2447576/4692662, that could be fixed with a reconnection.
- On https://github.com/PostgREST/postgrest/issues/1352, an undefined function(42883) error shows. But that could also be fixed by creating the function? postgrest would reconnect when that's done.

So for now I've only added the strict case of dying on a syntax error.

@wolfgangwalther WDYT?

main/Main.hs
```diff
@@ -244,21 +249,27 @@ connectionStatus pool =
...
     Left e -> do
-      -- If this error happens it would mean the connection is down again. Improbable because connectionStatus ensured the connection.
-      -- It's not a problem though, because App.postgrest would retry the connectionWorker or the user can do a SIGSUR1 again.
-      hPutStrLn stderr . toS . errorPayload $ PgError False e
-      putStrLn ("Failed to load the schema cache" :: Text)
+      let err = PgError False e
+          putErr = hPutStrLn stderr . toS . errorPayload $ err
+      case checkIsFatal err of
+        Just _  -> do
+          hPutStrLn stderr ("A fatal error ocurred when loading the schema cache" :: Text) >> putErr
```
wolfgangwalther, line 263: What do you think about adding wording along the lines of "this is probably a bug in PostgREST. Please report it at <github>"?

steve-chavez: Sure, that sounds nice for the user.

"This is probably a bug in PostgREST. Please report it at https://github.com/PostgREST/postgrest/issues"?

Is that good?

wolfgangwalther:
> We should be careful in quitting on retrying the schema cache query.
> 
>     * On #1287, theres an `XX000` error that looks like it might be fatal to our query. But according to https://stackoverflow.com/a/2447576/4692662, that could be fixed with a reconnection.
> 
>     * On #1352, an undefined function(42883) error shows. But that could also be fixed by creating the function? postgrest would reconnect when that's done.
> 
> 
> So for now I've only added the strict case of dying on a syntax error.
> 
> @wolfgangwalther WDYT?

Of course - if there is any chance to recover from the error without user interaction, then we should retry.

If the user needs to create a function to make this work, they could also trigger a restart of PostgREST? This seems to be a case that's somewhere in the middle. Not sure about that. However both your examples seem to be about fdws - it seems reasonable to restart for fdw stuff, because we can't tell why the fdw failed. It could very well be just a connection loss or whatever, so easily retryable.

Dying has the advantage that the user needs to take action and it's more likely that they will report a bug to us - so for those cases where it's highly likely that we have a bug, this should be the best thing to do.

Starting with the syntax error seems reasonable, even if it's just for convenience while developing (with still hypothetical `postgrest-watch postgrest-run --dump-schema`, I'm sure we will get to that point). We can add more errors along the way, when we have other cases where it makes sense to die.

If in doubt, retrying should be the safe choice.

main/Main.hs
wolfgangwalther: Yes!

wolfgangwalther: You can be sure that I will then report my syntax errors during development to the issue tracker :D

steve-chavez:
> Dying has the advantage that the user needs to take action and it's more likely that they will report a bug to us 

True. Now that you mention it, I think it would be fit even on fdws. As it is, the inifinite loop can discourage the user in trying to make postgrest work.

> We can add more errors along the way, when we have other cases where it makes sense to die.

Yeah. When trying to get a syntax error I got a "42703"(undefined column), we can refine it progressively.

steve-chavez:
The code coverage job is taking some time, but I believe this should be good to merge.

PR #1691, opened 2020-12-13
feat: Read config directly from environment variables

### LINKED ISSUE #1624: Can't set app settings when using environment variables

wolfgangwalther:
resolves #1624

TODO (after #1682 is merged):
- [x] make tooling and tests use env vars directly without config files 
- [x] rewrite io test in python

@steve-chavez The haskell parts are done and ready for review.


nix/docker/default.nix
```diff
@@ -26,27 +26,6 @@ let
 
       config = {
         Cmd = [ "/bin/postgrest" "/etc/postgrest.conf" ];
-        Env = [
-          "PGRST_DB_URI=postgresql://?user=postgres"
-          "PGRST_DB_SCHEMA=public"
-          "PGRST_DB_ANON_ROLE="
-          "PGRST_DB_POOL=100"
```
wolfgangwalther, line 33: I found this gem. This seems to be a typo. According to some of @steve-chavez performance tests, this could have a negative impact on performance for our docker containers, right?

It's back to the default of `10` now.

test/io-tests/configs/no-defaults.env
```diff
@@ -0,0 +1,26 @@
...
+PGRST_JWT_AUD="https://postgrest.org"
+PGRST_SECRET_IS_BASE64=true
+PGRST_ROLE_CLAIM_KEY=".user[0].\"real-role\""
+PGRST_MAX_ROWS=1000
+PGRST_PRE_REQUEST="please_run_fast"
+PGRST_ROOT_SPEC="openapi_v3"
+PGRST_RAW_MEDIA_TYPES="application/vnd.pgrst.config"
+PGRST_LOG_LEVEL="info"
+PGRST_APP_SETTINGS_test="test"
+PGRST_APP_SETTINGS_test2="test"
```
wolfgangwalther, line 26: `PGRST_APP_SETTINGS_xxx` variables are currently read in as-is, without any transformation on the `xxx` part. I am unsure whether we should apply some transformations like making them lowercase and replacing `_` with `.` or something like that.

nix/release/docker-hub-full-description.md
```diff
@@ -15,13 +15,8 @@ write from scratch.
...
 
-You can configure the PostgREST image by setting the enviroment variables used
-in the default `/etc/postgrest.conf` file or overriding that file. This is the
-default configuration file:
-
-```
-$DEFAULT_CONFIG
-```
+You can configure the PostgREST image by setting
+[enviroment variables](https://postgrest.org/en/stable/configuration.html).
```
wolfgangwalther, line 19: This link (and the changed docker hub description) will only be correct/useful, when we have released the next stable version. Once we release the next nightly, this will be out of sync. We should maybe skip the `postgrest-release-dockerhubdescription` step for nightly releases to make sure the description always stays in sync with the stable release?

@steve-chavez @monacoremo WDYT?

monacoremo: Yes, we should skip postgrest-release-dockerhubdescription for the nightly builds

wolfgangwalther:
I now changed our tooling to use environment variables directly and not through config files. So `with_tmp_db` sets `PGRST_DB_URI` etc. now - and `postgrest` ready those.

That means - among other stuff - that `test/with_tmp_db postgrest-run` will now just work without anything else.

wolfgangwalther:
I also changed `test/create_test_db` to be called just like `with_tmp_db`. `.circleci/config.yml` as an example:

```
test/create_test_db "postgres://circleci@localhost" postgrest_test stack test
```

The scripts exports the right env variables from argument 1 and 2 and then calls argument 3+ just like `with_tmp_db`.

nix/docker/default.nix
steve-chavez: Woah! Nice catch!

Yes, too many connections will lead to contention: https://wiki.postgresql.org/wiki/Number_Of_Database_Connections#Summary

(Also confirmed that it's true, throughput decreased on my tests)

src/PostgREST/Config.hs
```diff
@@ -152,10 +158,10 @@ readCLIShowHelp = customExecParser parserPrefs opts
         long "dump-config" <>
         help "Dump loaded configuration and exit"
       ) <*>
-      strArgument (
+      optional (strArgument (
```
steve-chavez, line 161: So now that this is optional, when running the binary without a config file, I get:

```
postgrest-run

Error parsing config file:
        missing key: db-anon-role
```

Before we would get the "missing" and the `help` output:

```
postgrest
Missing: FILENAME

Usage: postgrest FILENAME
  PostgREST 7.0.0 (UNKNOWN) / create a REST API to an existing Postgres database
...
```

Do you think we should keep it like this(it'd be a breaking change)? Or perhaps we can show the `help` output when no env var is defined?

test/io-tests/configs/no-defaults.env
steve-chavez: I think it's good to leave them as-is. That way it corresponds to the config file option - which AFAICT can use lower/uppercase, same for `_` or `.`.

test/QueryCost.hs
```diff
@@ -24,7 +24,7 @@ import Test.Hspec
 
 main :: IO ()
 main = do
-  testDbConn <- getEnvVarWithDefault "POSTGREST_TEST_CONNECTION" "postgres://postgrest_test@localhost/postgrest_test"
+  testDbConn <- getEnvVarWithDefault "PGRST_DB_URI" "postgres://postgrest_test@localhost/postgrest_test"
```
steve-chavez, line 27: Really nice change.  Makes things more consistent :+1:

src/PostgREST/Config.hs
```diff
@@ -292,14 +298,28 @@ dumpAppConfig conf = do
         secret = fromMaybe mempty $ configJwtSecret c
     showSocketMode c = showOct (fromRight' $ configServerUnixSocketMode c) ""
 
+-- This class is needed for the polymorphism of overrideFromEnvironment
+-- because C.required and C.optional have different signatures
+class JustIfMaybe a b where
```
steve-chavez, line 303: Like the comments. I don't see another way. So this one is good.

steve-chavez:
One thing nice about this is that now we can start postgrest without creating a file(avoiding the need for a sample config, as I wanted before):

```
PGRST_DB_ANON_ROLE=postgres \
PGRST_DB_SCHEMA=test \
PGRST_DB_URI="postgres://postgres/localhost:5432/postgrest_test" \
postgrest

Listening on port 3000
Connection successful
Schema cache loaded
```

main/Main.hs
```diff
@@ -125,10 +125,11 @@ main = do
     Catch connWorker
     ) Nothing
 
-  -- Re-read the config on SIGUSR2
-  void $ installHandler sigUSR2 (
-    Catch $ reReadConfig (cliPath opts) refConf
-    ) Nothing
+  -- Re-read the config on SIGUSR2, but only if we have a config file
```
steve-chavez, line 128: So by using env vars we lose the ability to reload configs without downtime right?

According to https://unix.stackexchange.com/a/314913/278242:

> The environment is copied into the process' address space when the process is created. Changes in the environment only affect newly started processes, you can't change the environment of a running program from the outside.

So we should discourage using them for systemd services. Or should they only be used for docker?

wolfgangwalther: I'm not sure whether we should discourage using them, but yes I think the combinations of systemd / config file and docker / environment variables make a lot of sense.

In a docker based setup I would want to use in-database-configuration for those parts that should be reloadable.

src/PostgREST/Config.hs
wolfgangwalther: I agree that when no filename is given and no required configs are set via environment variables, we should show the usage screen. It wouldn't need to contain "Missing: FILENAME" (because it's not required) but tell the user to either set a config file or use environment variables.

Once any of the required configs is set or a file name is missing, we can still error out with the same error as right now. The user should know by now **how** to set this up, they just forgot one of the settings.

main/Main.hs
steve-chavez: Right. Discouraging is not needed if we can be clear about reloading on the docs. So no problem here.

src/PostgREST/Config.hs
steve-chavez: >  "Missing: FILENAME" (because it's not required) but tell the user to either set a config file or use environment variables.

True. A friendly message for that would be great. Also forgot that the help message is also valuable for showing the postgrest version. We definitely need to keep it.

wolfgangwalther: > I agree that when no filename is given and no required configs are set via environment variables, we should show the usage screen. It wouldn't need to contain "Missing: FILENAME" (because it's not required) but tell the user to either set a config file or use environment variables.

I did exactly that. I removed the example config file from the usage screen as well and added another `--example` cli option to show it instead.

wolfgangwalther:
I made the changes to the error messages / usage screen / example config as mentioned above. Also rebased on master and made the newly added `postgrest-dump-schema` use environment variables instead of the config file, too.

Ready to merge, I think.

wolfgangwalther:
The only CI job that is failing is the old coverage job - but that is broken and soon to be replaced. It makes no sense to keep this job, so I removed it for now.

PR #1692, opened 2020-12-13
run spec tests in parallel

wolfgangwalther:
Now that we are able to run the spec tests repeatedly on the same database, I was wondering whether we could run them in parallel as well. With a few more tests adjusted, this is possible.

I only get an improvment of running postgrest-test-spec-all of about 20-25%. It took ~ 1 min before for me, ~45 seconds now. I had hoped for more, but hey it's still an improvement!

### DIFF
--- a/test/Feature/InsertSpec.hs
+++ b/test/Feature/InsertSpec.hs
@@ -293,29 +293,30 @@ spec actualPgVersion = do
         -- reset pk sequence first to make test repeatable
         request methodPost "/rpc/reset_sequence"
             [("Prefer", "tx=commit")]
-            [json|{"name": "items_id_seq", "value": 20}|]
+            [json|{"name": "items2_id_seq", "value": 20}|]
           `shouldRespondWith`
             [json|""|]
 
-        request methodPost "/items" [("Prefer", "return=representation")] "{}"
-          `shouldRespondWith` [json|[{ id: 20 }]|]
-          { matchStatus  = 201,
-            matchHeaders = []
-          }
+        request methodPost "/items2"
+            [("Prefer", "return=representation")]
+            [json|{}|]
+          `shouldRespondWith`
+            [json|[{ id: 20 }]|]
+            { matchStatus  = 201 }
 
       it "successfully inserts a row with all-default columns with prefer=rep and &select=" $ do
         -- reset pk sequence first to make test repeatable
         request methodPost "/rpc/reset_sequence"
             [("Prefer", "tx=commit")]
-            [json|{"name": "items_id_seq", "value": 20}|]
+            [json|{"name": "items3_id_seq", "value": 20}|]
           `shouldRespondWith`
             [json|""|]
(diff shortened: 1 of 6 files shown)


test/Feature/InsertSpec.hs
```diff
@@ -293,29 +293,30 @@ spec actualPgVersion = do
...
+            [json|{"name": "items2_id_seq", "value": 20}|]
           `shouldRespondWith`
             [json|""|]
 
-        request methodPost "/items" [("Prefer", "return=representation")] "{}"
-          `shouldRespondWith` [json|[{ id: 20 }]|]
-          { matchStatus  = 201,
-            matchHeaders = []
-          }
+        request methodPost "/items2"
```
wolfgangwalther, line 300: Because of concurrency issues, we're running all the insert tests that use the auto incrementing PK on `items` on different tables now: items, items2 and items3. No conflict anymore.

test/Feature/UpdateSpec.hs
```diff
@@ -51,32 +51,14 @@ spec = do
 
     context "in a nonempty table" $ do
       it "can update a single item" $ do
-        get "/items?id=eq.42"
-          `shouldRespondWith`
-            [json|[]|]
-
-        request methodPatch "/items?id=eq.2"
-            [("Prefer", "tx=commit")]
```
wolfgangwalther, line 59: Persistance of PATCH is tested in the rollback specs. Here it just leads to more concurrency issues, so I decided to take it out as well. Had already done that for the INSERT on the "all tests rollback" PR, I think.

wolfgangwalther:
@steve-chavez It'd be good, if you could double-check on the test cases, breaking those wouldn't be good.

test/fixtures/schema.sql
```diff
@@ -113,6 +113,14 @@ CREATE TABLE items2 (
...
 );
 
+CREATE TABLE items3 (
+    id bigserial primary key
+);
+
+CREATE TABLE persistance (
+    id bigint primary key
+);
+
```
steve-chavez, line 123: @wolfgangwalther Hmm.. is this table used? I don't see it on the tests..

wolfgangwalther: Ah, good catch. Left over from something I tried. Will remove.

steve-chavez:
Nice! I go from `19.1069 seconds` to  `11.6206 seconds` on `postgrest-test-spec`.

Left a comment on the `persistance` table leftover.

Besides that, looks good to go!

test/Feature/UpdateSpec.hs
steve-chavez: Yes, looks good. I see PATCH persistence is already here:

https://github.com/PostgREST/postgrest/blob/a52c114ff43c4576d491106df11177bcb8a8cdbb/test/Feature/RollbackSpec.hs#L100-L114

PR #1694, opened 2020-12-13
Add postgrest-coverage to show and upload hpc reports to codecov

wolfgangwalther:
`postgrest-coverage` is here.

Some background on the why and how: When I introduced code coverage in CI a while back, I went with a combination of stack, coveralls, `stack-hpc-coveralls` and travis. I turns out that this has many disadvantages. This PR replaces everything, and we are now using cabal, codecov, `hpc-codecov` and circleci.

This has the following advantages:
- Coveralls does not support partial line coverage at all - codecov does. Of course some information is still lost, because the hpc coverage info is very detailed for expressions, etc. - but having that marked as partial line coverage instead of "no coverage at all" will hopefully be much better.
- Only travis supported token-less upload of coverage reports for coveralls. For codecov, we can do the same from circleci. Travis had huge delays for the coverage job in the last couple of months, because of their move to new infrastructure. Most of our jobs are running on circleci and on the nix pipeline, so it's good that we can add the coverage here as well.
- Of course using a cabal based setup instead of stack is better, because our whole nix tooling is cabal-based.
- `hpc-codecov` seems to be much more maintained than `stack-hpc-coveralls`. At least there were some recent commits to the former, while the latter didn't have much done in the last years. Also the new tool is "build tool agnostic", so it does not care whether it's stack or cabal. It just doesn't make any assumptions of directory structure. Since we're copying together various `.tix` files from different tests (spec and io so far), this makes it easier to handle.

I have not had a look at a codecov coverage report for the repo, yet, but I'm sure it will be better than what we have right now.

TODO:
- [x] fix ci failure
- [x] register repo at codecov.io
- [x] add overlay for deliberately untested code

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,6 +234,13 @@ jobs:
           name: Run memory tests
           command: postgrest-test-memory
           when: always
+      - run:
+          name: Run coverage
+          command: postgrest-coverage
+      - run:
+          name: Upload coverage to codecov
+          command: bash <(curl -s https://codecov.io/bash) -f coverage/codecov.json
+
 
 workflows:
   version: 2
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ dist-newstyle
 postgrest.hp
 postgrest.prof
 __pycache__
+*.tix
+coverage
+.hpc
--- a/default.nix
+++ b/default.nix
@@ -66,7 +66,7 @@ let
 
   # Options passed to cabal in dev tools and tests
   devCabalOptions =
-    "-f FailOnWarn --test-show-detail=direct";
+    "-f dev --test-show-detail=direct";
 
   profiledHaskellPackages =
     pkgs.haskell.packages."${compiler}".extend (self: super:
@@ -87,7 +87,7 @@ rec {
   # libraries and documentation. We disable running the test suite on Nix
   # builds, as they require a database to be set up.
   postgrestPackage =
-    lib.dontCheck (lib.enableCabalFlag postgrest "FailOnWarn");
+    lib.dontCheck postgrest;
 
   # Static executable.
   postgrestStatic =
@@ -114,7 +114,11 @@ rec {
 
   # Scripts for running tests.
   tests =
-    pkgs.callPackage nix/tests.nix { inherit postgrest postgrestStatic postgrestProfiled postgresqlVersions devCabalOptions; };
+    pkgs.callPackage nix/tests.nix {
+      inherit postgrest postgrestProfiled postgresqlVersions devCabalOptions;
+      ghc = pkgs.haskell.compiler."${compiler}";
+      hpc-codecov = pkgs.haskell.packages."${compiler}".hpc-codecov;
+    };
 
   # Linting and styling scripts.
   style =
--- a/nix/devtools.nix
+++ b/nix/devtools.nix
@@ -72,6 +72,8 @@ let
       }
       ''
         ${cabal-install}/bin/cabal v2-clean
+        # clean old coverage data, too
+        rm -rf .hpc coverage
       '';
 
   check =
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -5,15 +5,15 @@
 , checkedShellScript
 , curl
 , devCabalOptions
-, diffutils
+, ghc
+, gnugrep
 , haskell
+, hpc-codecov
 , lib
 , postgresql
 , postgresqlVersions
 , postgrest
 , postgrestProfiled
-, postgrestStatic
-, procps
 , python3
 , runtimeShell
 , yq
@@ -164,6 +164,90 @@ let
             postgrest --dump-schema \
             | ${yq}/bin/yq -y .
       '';
+
+  coverage =
+    name: postgresql:
+    checkedShellScript
+      {
+        inherit name;
+        docs = "Run spec and io tests while collecting hpc coverage data.";
+        inRootDir = true;
+      }
+      ''
+        env="$(cat ${postgrest.env})"
+        export PATH="$env/bin:$PATH"
+
+        # clean up previous coverage reports
+        mkdir -p coverage
+        rm -rf coverage/*
+
+        # temporary directory to collect data in
+        tmpdir="$(mktemp -d)"
+
+        # we keep the tmpdir when an error occurs for debugging and only remove it on success
+        trap 'echo Temporary directory kept at: $tmpdir' ERR SIGINT SIGTERM
+
+        # build once before running all the tests
+        ${cabal-install}/bin/cabal v2-build ${devCabalOptions} --enable-tests all
+
+        # collect all tests
+        HPCTIXFILE="$tmpdir"/io.tix \
+        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-exec ${devCabalOptions} \
+          ${ioTestPython}/bin/pytest -- -v test/io-tests
+          
+        HPCTIXFILE="$tmpdir"/spec.tix \
+        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-test ${devCabalOptions}
+
+        # collect all the tix files
+        ${ghc}/bin/hpc sum  --union --exclude=Paths_postgrest --output="$tmpdir"/tests.tix "$tmpdir"/io.tix "$tmpdir"/spec.tix
+
+        # prepare the overlay
+        ${ghc}/bin/hpc overlay --output="$tmpdir"/overlay.tix test/coverage.overlay
+        ${ghc}/bin/hpc sum --union --output="$tmpdir"/tests-overlay.tix "$tmpdir"/tests.tix "$tmpdir"/overlay.tix
+
+        # check nothing in the overlay is actually tested
+        ${ghc}/bin/hpc map --function=inv --output="$tmpdir"/inverted.tix "$tmpdir"/tests.tix
+        ${ghc}/bin/hpc combine --function=sub \
+          --output="$tmpdir"/check.tix "$tmpdir"/overlay.tix "$tmpdir"/inverted.tix
+        # returns zero exit code if any count="<non-zero>" lines are found, i.e.
+        # something is covered by both the overlay and the tests
+        if ${ghc}/bin/hpc report --xml "$tmpdir"/check.tix | ${gnugrep}/bin/grep -qP 'count="[^0]'
+        then
+          ${ghc}/bin/hpc markup --highlight-covered --destdir=coverage/overlay "$tmpdir"/overlay.tix || true
+          ${ghc}/bin/hpc markup --highlight-covered --destdir=coverage/check "$tmpdir"/check.tix || true
+          echo "ERROR: Something is covered by both the tests and the overlay:"
+          echo "file://$(pwd)/coverage/check/hpc_index.html"
+          exit 1
+        else
+          # copy the result .tix file to the coverage/ dir to make it available to postgrest-coverage-draft-overlay, too
+          cp "$tmpdir"/tests-overlay.tix coverage/postgrest.tix
+          # prepare codecov json report
+          ${hpc-codecov}/bin/hpc-codecov --mix=.hpc --out=coverage/codecov.json coverage/postgrest.tix
+
+          # create html and stdout reports
+          # TODO: The markup command fails when run outside nix-shell (i.e. in CI!)
+          # Need to fix it properly in the future instead of adding the || true
+          ${ghc}/bin/hpc markup --destdir=coverage coverage/postgrest.tix || true
+          echo "file://$(pwd)/coverage/hpc_index.html"
+          ${ghc}/bin/hpc report coverage/postgrest.tix "$@"
+        fi
+
+        rm -rf "$tmpdir"
+      '';
+
+  coverageDraftOverlay =
+    name:
+    checkedShellScript
+      {
+        inherit name;
+        docs = "Create a draft overlay from current coverage report.";
+        inRootDir = true;
+      }
+      ''
+        ${ghc}/bin/hpc draft --output=test/coverage.overlay coverage/postgrest.tix
+        sed -i 's|^module \(.*\):|module \1/|g' test/coverage.overlay
+      '';
+
 in
 # Create an environment that contains all the utility scripts for running tests
   # that we defined above.
@@ -179,6 +263,8 @@ buildEnv
         testSpecAllVersions.bin
         (testIO "postgrest-test-io" postgresql).bin
         (dumpSchema "postgrest-dump-schema" postgresql).bin
+        (coverage "postgrest-coverage" postgresql).bin
+        (coverageDraftOverlay "postgrest-coverage-draft-overlay").bin
       ] ++ testSpecVersions;
   }
   # The memory tests have large dependencies (a profiled build of PostgREST)
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -19,12 +19,17 @@ source-repository head
   type:     git
   location: git://github.com/PostgREST/postgrest.git
 
-flag FailOnWarn
+flag dev
   default:     False
   manual:      True
-  description: No warnings allowed
+  description: Development flags
 
 library
+  default-language:   Haskell2010
+  default-extensions: OverloadedStrings
+                      QuasiQuotes
+                      NoImplicitPrelude
+  hs-source-dirs:     src
   exposed-modules:    PostgREST.ApiRequest
                       PostgREST.App
                       PostgREST.Auth
@@ -43,7 +48,6 @@ library
                       PostgREST.Private.Common
                       PostgREST.Private.ProxyUri
                       PostgREST.Private.QueryFragment
-  hs-source-dirs:     src
   build-depends:      base                      >= 4.9 && < 4.15
                     , HTTP                      >= 4000.3.7 && < 4000.4
                     , Ranged-sets               >= 0.3 && < 0.5
@@ -88,15 +92,12 @@ library
                     , wai-extra                 >= 3.0.19 && < 3.2
                     , wai-logger                >= 2.3.2
                     , wai-middleware-static     >= 0.8.1 && < 0.10
-  default-language:   Haskell2010
-  default-extensions: OverloadedStrings
-                      QuasiQuotes
-                      NoImplicitPrelude
-  if flag(FailOnWarn)
-    ghc-options: -O2 -Werror -Wall -fwarn-identities
+  if flag(dev)
+    ghc-options: -O0 -Werror -Wall -fwarn-identities
                  -fno-spec-constr -optP-Wno-nonportable-include-path
+                 -fhpc -hpcdir .hpc
   else
-    ghc-options: -O2 -Wall -fwarn-identities
+    ghc-options: -O2 -Werror -Wall -fwarn-identities
                  -fno-spec-constr -optP-Wno-nonportable-include-path
           -- -fno-spec-constr may help keep compile time memory use in check,
           --   see https://gitlab.haskell.org/ghc/ghc/issues/16017#note_219304
@@ -105,8 +106,12 @@ library
           --   see https://github.com/commercialhaskell/stack/issues/3918
 
 executable postgrest
-  main-is:            Main.hs
+  default-language:   Haskell2010
+  default-extensions: OverloadedStrings
+                      QuasiQuotes
+                      NoImplicitPrelude
   hs-source-dirs:     main
+  main-is:            Main.hs
   build-depends:      base                >= 4.9 && < 4.15
                     , aeson               >= 1.4.7 && < 1.6
                     , auto-update         >= 0.1.4 && < 0.2
@@ -126,17 +131,14 @@ executable postgrest
                     , time                >= 1.6 && < 1.11
                     , wai                 >= 3.2.1 && < 3.3
                     , warp                >= 3.2.12 && < 3.4
-  default-language:   Haskell2010
-  default-extensions: OverloadedStrings
-                      QuasiQuotes
-                      NoImplicitPrelude
-  if flag(FailOnWarn)
+  if flag(dev)
     ghc-options:      -threaded -rtsopts "-with-rtsopts=-N -I2"
-                      -O2 -Werror -Wall -fwarn-identities
+                      -O0 -Werror -Wall -fwarn-identities
                       -fno-spec-constr -optP-Wno-nonportable-include-path
+                      -fhpc -hpcdir .hpc
   else
     ghc-options:      -threaded -rtsopts "-with-rtsopts=-N -I2"
-                      -O2 -Wall -fwarn-identities
+                      -O2 -Werror -Wall -fwarn-identities
                       -fno-spec-constr -optP-Wno-nonportable-include-path
 
   if !os(windows)
@@ -145,6 +147,11 @@ executable postgrest
 
 test-suite spec
   type:               exitcode-stdio-1.0
+  default-language:   Haskell2010
+  default-extensions: OverloadedStrings
+                      QuasiQuotes
+                      NoImplicitPrelude
+  hs-source-dirs:     test
   main-is:            Main.hs
   other-modules:      Feature.AndOrParamsSpec
                       Feature.AsymmetricJwtSpec
@@ -179,7 +186,6 @@ test-suite spec
                       Feature.UpsertSpec
                       SpecHelper
                       TestTypes
-  hs-source-dirs:     test
   build-depends:      base              >= 4.9 && < 4.15
                     , aeson             >= 1.4.7 && < 1.6
                     , aeson-qq          >= 0.8.1 && < 0.9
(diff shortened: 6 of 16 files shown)


nix/tests.nix
```diff
@@ -8,6 +8,7 @@
 , diffutils
 , git
 , haskell
+, haskellPackages
```
wolfgangwalther, line 11: @monacoremo This is certainly not the right way to include this here, I assume. Need it to get `hpc-codecov`.

When I started nix-shell the first time I saw some downloads of ghc 8.8.4 or something like that, while everything else is running on 8.8.3. So I assume, this is wrong?

nix/tests.nix
```diff
@@ -97,7 +98,7 @@ let
 
         ${cabal-install}/bin/cabal v2-build ${devCabalOptions}
         ${cabal-install}/bin/cabal v2-exec ${withTmpDb postgresql} \
-          ${ioTestPython}/bin/pytest -- -v "$rootdir"/test/io-tests "$@"
```
wolfgangwalther, line 100: I think we can skip a few of those `$rootdir`s, because we are doing `cd $rootdir` in those scripts.

nix/tests.nix
```diff
@@ -112,6 +113,58 @@ let
...
+    let
+      covCabalOptions="${devCabalOptions} --enable-coverage --enable-executable-dynamic";
+    in
+    checkedShellScript
+      name
+      ''
+        env="$(cat ${postgrest.env})"
+        export PATH="$env/bin:$PATH"
+
+        rootdir="$(${git}/bin/git rev-parse --show-toplevel)"
```
wolfgangwalther, line 128: @monacoremo Maybe we should refactor the rootdir / cd rootdir out. It seems like this is something that makes sense to do for every script.

nix/tests.nix
```diff
@@ -112,6 +113,58 @@ let
...
+        cd "$rootdir"
+
+        # clean up previous coverage data
+        rm -rf coverage
+        find . -type f -regextype posix-extended -regex ".*\.(tix)" -print0 | xargs -0 rm -f
```
wolfgangwalther, line 133: Note to self: I think I can simplify the regex. It was more complicated before, but doesn't need to be posix-extended anymore...

nix/tests.nix
```diff
@@ -112,6 +113,58 @@ let
 
         ${withTmpDb postgresql} "$rootdir/test/memory-tests.sh"
       '';
+
+  coverage =
+    name: postgresql:
+    let
+      covCabalOptions="${devCabalOptions} --enable-coverage --enable-executable-dynamic";
```
wolfgangwalther, line 120: Without enabling `--enable-executable-dynamic` I had the problem that as soon as I built exe, lib and tests, every part was built twice. HPC flags seemed to change randomly. Whenever I ran the coverage job, it had to rebuild from scratch, even though I did not change any code.

The combination of `--enable-coverage --enable-execute-dynamic` solves that. The executable is now linked against a shared library. Maybe we should enable that for all dev builds? Maybe it is already - I am confused where those settings actually come from. `package.cabal`, `devCabalOptions`, `*.nix`, ...

.circleci/config.yml
```diff
@@ -229,6 +229,13 @@ jobs:
           name: Run memory tests
           command: postgrest-test-memory
           when: always
+      - run:
+          name: Run coverage
+          command: postgrest-coverage
+      - run:
+          name: Upload coverage to codecov
+          command: bash <(curl -s https://codecov.io/bash) -f coverage/codecov.json
```
wolfgangwalther, line 237: This should work without token for circleci, too! Once we're registered there, of course.

nix/tests.nix
monacoremo: Yes, was also thinking about factoring out a postgrest-rootdir - lets do that in a separate PR. I added those for developer experience, as it's nice to have the scripts work consistently in nix shell independently from cwd

nix/tests.nix
monacoremo: Hm, yes... but the setup I chose might be a bit inelegant, and might be better fixed with an overlay that picks the default haskellPackages of our choosing.

To fix this short term: import just `haskell` and pass in our `compiler` variable from default.nix, then use `pkgs.haskell.packages."${compiler}".[...]`

wolfgangwalther: So... I just did this in default.nix:

```
  haskellPackages = pkgs.haskell.packages."${compiler}";
...
  tests =
    pkgs.callPackage nix/tests.nix { inherit haskellPackages ...; };
```

It looked like it worked: When I restarted nix-shell, it downloaded / built hpc again.

Does that make sense, remotely?

nix/tests.nix
```diff
@@ -143,6 +144,58 @@ let
...
+
+        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-exec ${covCabalOptions} \
+          ${coveragePathHelper} -- ${ioTestPython}/bin/pytest -v test/io-tests
+        ${withTmpDb postgresql} ${cabal-install}/bin/cabal v2-test ${covCabalOptions}
+
+        # collect all the hpc files in one location
+        hpcdir="$(find ./dist-newstyle -type d -iname dyn)"
+        mkdir -p coverage
+        cp -r "$hpcdir"/mix/postgrest*/* coverage
+        hpc sum --output=coverage/all.tix --union --exclude=Paths_postgrest postgrest.tix "$hpcdir"/tix/postgrest*/*.tix
```
monacoremo, line 177: These (hpc) should also come from some nix derivation.

nix/tests.nix
monacoremo: It would be even cleaner I think to request a concrete derivation we want (hpc, hpc-codecov?) at the top of this file instead of the broad haskellPackages, and then to pass it in explicitly in default.nix:

```
tests = ... callPackage ... { inherit ...; hpc = pkgs.haskell....; }

```

nix/tests.nix
wolfgangwalther: Ah, yes, thanks for bringing that up again. It's in ghc itself and I couldn't figure out how to get it properly at the time. Now I get it from `pkgs.haskell.compiler."${compiler}"` - that works.

.circleci/config.yml
steve-chavez: Done! I've also added the `CODECOV_TOKEN` env var to CircleCI.

default.nix
```diff
@@ -87,7 +87,7 @@ rec {
   # libraries and documentation. We disable running the test suite on Nix
   # builds, as they require a database to be set up.
   postgrestPackage =
-    lib.dontCheck (lib.enableCabalFlag postgrest "FailOnWarn");
+    lib.dontCheck postgrest;
```
wolfgangwalther, line 90: Have no idea what I did here - tried to remove the "FailOnWarn" flag, because I had enabled it by default now. Didn't test that bit here - @monacoremo does the change make sense?

postgrest.cabal
```diff
@@ -88,15 +92,12 @@ library
...
                     , wai-logger                >= 2.3.2
                     , wai-middleware-static     >= 0.8.1 && < 0.10
-  default-language:   Haskell2010
-  default-extensions: OverloadedStrings
-                      QuasiQuotes
-                      NoImplicitPrelude
-  if flag(FailOnWarn)
-    ghc-options: -O2 -Werror -Wall -fwarn-identities
+  if flag(dev)
+    ghc-options: -O0 -Werror -Wall -fwarn-identities
```
wolfgangwalther, line 96: @steve-chavez Added -O0 for development and it feels faster. Didn't time it, though.

test/Feature/InsertSpec.hs
```diff
@@ -441,9 +437,6 @@ spec actualPgVersion = do
 
   describe "Row level permission" $
     it "set user_id when inserting rows" $ do
-      post "/postgrest/users" [json| { "id":"jdoe", "pass": "1234", "role": "postgrest_test_author" } |]
```
wolfgangwalther, line 444: After changing the ghc flags to `-Werror` for the test suite as well, I got some errors here. I removed those two lines.... and the test still passes. That kind of worries me about the whole test...

I will have to have a closer look at the whole thing.

wolfgangwalther:
Some notes about what I did in this PR:
- Previously we had a `FailOnWarn` flag in `postgrest.cabal`. I now changed this to a `dev` flag. `-Werror` is now enabled for all builds (with and without dev flag). The dev flag is used by our nix devtools and changes `-O2` to `-O1` and adds the hpc flags. Collecting coverage info is now enabled in all devtools. This allows running `postgrest-coverage` without rebuilding everything. Cleaned up the `postgrest.cabal` file along the way (key ordering, capitalization).
- By setting the `-fhpc` flags through `postgrest.cabal` I completely avoid cabal's `--enable-coverage`, which gave me problems earlier. Build times are now short and collecting coverage data is simple, because the data is not burried somewhere in `dist-newstyle/...`, but directly on the top-level in `.hpc`.
- I added `-Werror` to our test-suites as well - this gave a few warnings about unused imports, which I fixed.
- `postgrest-coverage` does the following:
  1. Build and run io and spec tests
  2. Merge coverage data from both tests
  3. Add overlay coverage data (see below) to it.
  4. Check that no code is tested and overlayed at the same time. This is because the overlay is supposed to be used for stuff that we can't test. If we now add tests that cover those parts, we don't need the overlay anymore.
  5. Create json data for codecov and html/stdout reports for the user.
- I added a `test/coverage.overlay` file. This is in hpc overlay format and manually "ticks" those lines that can't be tested (e.g. unreachable code, that is still needed because of pattern matching, etc...). To create this, I went through all untested code. In some occasions, I did small refactors instead of adding an overlay, because those parts were not only untested but also not needed. In most cases this is a `deriving (Show)` that is never used.

default.nix
monacoremo: LGTM!

nix/devtools.nix
```diff
@@ -25,7 +25,7 @@ let
       }
       ''
         while true; do
-          (! ${silver-searcher}/bin/ag -l . | ${entr}/bin/entr -dr "$@")
+          (! ${silver-searcher}/bin/ag -l main src test | ${entr}/bin/entr -dr "$@")
```
wolfgangwalther, line 28: I needed to change this to support `postgrest-watch postgrest-coverage`, since the coverage script adds files on the top level directory. This causes an infinite loop of restarting through the watch script, even though those files are not monitored directly. This is because the `-d` option to entr watches changes to the top-level directory as soon as we watch any file from that directory.

The only file that is relevant to watch for changes would be `postgrest.cabal`. So right now, this is not the case and a restart with watch running has to be triggered manually when changing this file. Since we can only have one or the other, I thought supporting `postgrest-watch postgrest-coverage` makes more sense, because - at least in theory - it should be used more often ;).

nix/tests.nix
```diff
@@ -164,6 +164,78 @@ let
...
+          echo "file://$(pwd)/coverage/check/hpc_index.html"
+          exit 1
+        else
+          # prepare codecov json report
+          ${hpc-codecov}/bin/hpc-codecov --mix=.hpc --out=coverage/codecov.json .hpc/tests-overlay.tix
+
+          # create html and stdout reports
+          # TODO: The markup command fails when run outside nix-shell (i.e. in CI!)
+          # Need to fix it properly in the future instead of adding the || true
+          ${ghc}/bin/hpc markup --destdir=coverage .hpc/tests-overlay.tix || true
```
wolfgangwalther, line 220: This `|| true` is really ugly. Once `postgrest-coverage` is started from outside nix-shell by installing it with nix-env, it starts failing at this line with the following error:

```
Writing: Main.hs.html
hpc: ./main/Main.hs: hGetContents: invalid argument (invalid byte sequence)
```

I have no clue where this comes from. When I add `--exclude Main` to the hpc command it works, but of course the html file for Main is not written. So there seems to be something special about the `Main.hs` file - in combination with running the command outside nix-shell. @monacoremo maybe you can have a look? My nix skills are at their limits...

.circleci/config.yml
wolfgangwalther: Ok, great. Seems to work - even without the token, because "pass secrets to PRs" was not enabled. Therefore I removed the token from CircleCI again - it works without it. Less noise.

wolfgangwalther:
I added a codecov config file to prevent the codecov comment from showing up - I feel like this will quickly become annoying on frequent changes. I tried to enable status checks (just like the other CIs) instead, possibly even including the GitHub checks api.

test/Feature/InsertSpec.hs
wolfgangwalther: I don't think those two lines ever served any purpose. The route `/postgrest/users` certainly does not exist, so this request would always fail. But since the result of the request was never used or checked for, this wasn't a problem - until the compiler complained about the unused result. Should be safe to remove.

wolfgangwalther:
`stack-test` is currently failing randomly. This is a concurrency issue related to the parallel execution of spec tests. I will push a small fix for this in a minute, even though it is unrelated to this PR.

The codecov status check / github check didn't show up, yet, but I think that's because the base branch (master) doesn't have any coverage data, yet. We will be able to look at this in more detail once we merge this PR and then see what happens for other PRs.

This PR is ready for review now. @steve-chavez @monacoremo.

test/coverage.overlay
```diff
@@ -0,0 +1,167 @@
+module "postgrest-7.0.1-inplace/PostgREST.ApiRequest" {
+  inside "userApiRequest" {
+    tick "otherwise" on line 291;
+    tick "Nothing" on line 291;
```
monacoremo, line 4: How will we maintain/generate this file? It seems that if any lines change, we need to update this?

test/io-tests/test_io.py
```diff
@@ -73,15 +73,15 @@ def defaultenv():
     }
 
 
-def dumpconfig(configpath=None, env=None, stdin=None):
+def dumpconfig(configpath=None, env={}, stdin=None):
```
monacoremo, line 76: This is not quite the same, I suggest that we revert. It's good practice to use None for default arguments, as the dict in this new function definition will only be created once. If it's modified anywhere later by mistake, it would impact all other calls of the function.

wolfgangwalther: That's a left-over from some hacks that I had to use at some point inbetween. Figured I'd leave it in... I'm glad I did, learned about that problem now. Will revert it.

test/coverage.overlay
wolfgangwalther: Yes, this is something that I don't know about, yet. I figured we'd need to try and see. It seems like this could become ugly... :o).

This seems to be a challenge in the whole hpc overlay concept. I'm not experienced enough in using it to see what we can do about it, yet.

nix/tests.nix
```diff
@@ -158,12 +161,85 @@ let
...
+      ''
+        env="$(cat ${postgrest.env})"
+        export PATH="$env/bin:$PATH"
+
+        # clean up previous coverage data
+        rm -rf coverage && mkdir coverage
+        find . -type f -iname "*.tix" -delete
+
+        # build once before running all the tests
+        ${cabal-install}/bin/cabal v2-build ${devCabalOptions} --enable-tests all
```
monacoremo, line 189: For me it's building the library and tests twice - I think it's because the later invocations don't have --enable-tests all. Should we define `coverageCabalOptions` that always has that options?

test/io-tests/test_io.py
```diff
@@ -94,7 +94,7 @@ def dumpconfig(configpath=None, env=None, stdin=None):
 
 
 @contextlib.contextmanager
-def run(configpath=None, stdin=None, env=None, port=None):
+def run(configpath=None, stdin=None, env={}, port=None):
```
wolfgangwalther, line 97: @monacoremo 

In this case the `env` dict is modified in the function body - before the `env or {}` would happen. So this would need a fix of some kind - not passing `env` would break it.

We could just remove the default value, to make `env` required. We could also do something like `env = env or {}` at the beginning. Reassigning the argument should not be a problem, right?

monacoremo: Yes, that would work! As when we use `run(env={})`, the `{}` will be recreated on each invocation.

nix/tests.nix
wolfgangwalther: Hm, no I don't think the `--enable-tests` is the reason for that. I had that problem at the beginning as well - it only disappeared once I changed the `postgrest.cabal` file to contain the options in the right place. The underlying reason, I think, is that some parts are not built with coverage and some are - and that causes some kind of strange rebuild cycle.

Can you try to do a `postgrest-clean` before and then rebuild?

test/coverage.overlay
monacoremo: Looks quite messy, would impact pretty much every commit. Leaving it out would mean that we get a lower coverage ratio and less transparency on what is actually testable, right?

wolfgangwalther: Yes. I'm not that much concerned about coverage ratio, but much more about the mental overhead of deciding "should this be tested or not?" every time we look at the coverage reports. This will quickly lead to mistakes, where we don't test something that we should.

I thought a bit more about it - and the overlay tool is only really usable, once we are at 100% coverage. If we're at that point we can just do an automatic `postgrest-coverage-draft-overlay` for each commit and get the proper file, because everything left is "untestable". We can then tell easily from the diff of the overlay file whether just line numbers were changed or whether something was added that should actually be tested.

WDYT?

monacoremo:
I ran `postgrest-clean` and `postgrest-coverage` in `nix-shell --pure` and only got:

```
  2% expressions used (246/9370)
  2% boolean coverage (4/193)
       1% guards (2/104), 3 always True, 1 always False, 98 unevaluated
       2% 'if' conditions (2/72), 3 always False, 67 unevaluated
       0% qualifiers (0/17), 17 unevaluated
  3% alternatives used (28/781)
  2% local declarations used (10/440)
  4% top-level declarations used (22/487)
```

On the second run of `postgrest-coverage`:

```
 87% expressions used (9038/10297)
 63% boolean coverage (136/213)
      52% guards (64/121), 43 always True, 2 always False, 12 unevaluated
      85% 'if' conditions (64/75), 5 always True, 3 always False, 3 unevaluated
      47% qualifiers (8/17), 9 always True
 77% alternatives used (640/829)
 92% local declarations used (449/484)
 80% top-level declarations used (410/509)
```

This was consistent across two tries. Any idea why this happens?

Maybe we could cd to a fresh temporary directory on each run to collect the tix with more consistency?

nix/tests.nix
monacoremo: Multiple builds unfortunately...

wolfgangwalther:
> ```
> 
> ```
> 
> 
> This was consistent across two tries. Any idea why this happens?
> 
> Maybe we could cd to a fresh temporary directory on each run to collect the tix with more consistency?

It's very likely that this is related to the problem with multiple builds. Multiple builds are probably caused by some hpc / converage settings not propagated properly and then some parts of the project are built without collecting tix data. Judging from your numbers, I would expect that the spec tests are not collecting data correctly, most likely because some part of them is built without coverage enabled.

After the first and second run with low and high coverage, respectively - can you have a look at the `.tix` files that are created? There should be 3 files in the repo root: `postgrest.tix` (io tests) `spec.tix` and `spec-querycost.tix`. I am guessing that the last two are missing for the first run?

wolfgangwalther:
Actually... all 3 files are probably missing after the first run. The low numbers are most likely **only** the overlay ticks.

nix/tests.nix
wolfgangwalther: Just shooting in the dark here, but maybe you have some kind of global cabal setting with `--enable-coverage`? The cabal coverage settings always resulted in those double builds for me. Can you try adding `--disable-coverage` to all 3 cabal commands (build, exec, test)? It worked the same for me, because the hpc flags are passed manually.

wolfgangwalther:
> Maybe we could cd to a fresh temporary directory on each run to collect the tix with more consistency?

Indepdenent of the multiple builds or first run / second run problems - this would nicely solve a couple of other things that I had to do: The change to `postgrest-watch` would not be needed, just as a lot of the cleanup commands I added could be removed. Not sure how to do it, yet.

test/coverage.overlay
monacoremo: Yes, that makes sense! It will take some expertise to interpret the coverage reports in the meantime.

Not sure that we should aim for 100% coverage at all cost... Let's see how far we get with reasonable effort.

wolfgangwalther: Ok, so we will merge this without the overlay file first.

Maybe we should not go for 100% "at all cost". But everything that we decide the cost is too high, we can add to the overlay file, too, because we decided against testing it. In that sense we should absolutely go for 100% (test+overlay), because this is the only scenario that will be easily manageable without too much mental overhead.

wolfgangwalther:
> > Maybe we could cd to a fresh temporary directory on each run to collect the tix with more consistency?
> 
> Indepdenent of the multiple builds or first run / second run problems - this would nicely solve a couple of other things that I had to do: The change to `postgrest-watch` would not be needed, just as a lot of the cleanup commands I added could be removed. Not sure how to do it, yet.

Did that, the tix files are now collected in a temp directory. This solves those other cases nicely, but I don't think this will help with your first / second run problem, @monacoremo.

monacoremo:
Ah, nice find with the HPCTIXFILE env variable! Much simpler than I feared.

Will try with this new version.

test/Feature/InsertSpec.hs
steve-chavez: @wolfgangwalther :O These are leftovers of a really old auth implementation. Nice that the coverage caught this. Tests should have failed.

wolfgangwalther:
Nothing changed really, just rebased and moved two commits into separate PRs.

wolfgangwalther:
@steve-chavez Maybe you can give `postgrest-coverage` a test run, too? Remo is describing some problems above and this could help me decide whether my system is the odd one or Remos is.. :o).

Basically two things to watch out for:
- When running `postgrest-coverage` is all the stuff built once or twice?
- When running `postgrest-clean` + `postgrest-coverage` how do the numbers look like?

steve-chavez:
@wolfgangwalther Sure.

> When running postgrest-coverage is all the stuff built once or twice?

Twice. Here's the pastebin: https://pastebin.com/9FnqCbcf

I see two `postgrest-7.0.1 (lib) (configuration changed)` there.

steve-chavez:
> When running postgrest-clean + postgrest-coverage how do the numbers look like?

I got the same results: 

```
 87% expressions used (9044/10297)
 63% boolean coverage (136/213)
      52% guards (64/121), 43 always True, 2 always False, 12 unevaluated
      85% 'if' conditions (64/75), 4 always True, 4 always False, 3 unevaluated
      47% qualifiers (8/17), 9 always True
 77% alternatives used (640/829)
 92% local declarations used (449/484)
 80% top-level declarations used (410/509)
```

pastebin: https://pastebin.com/8ZrEjmpk

(didn't run `nix-shell --pure` though, did it on the same `nix-shell`)

wolfgangwalther:
> Twice. Here's the pastebin: https://pastebin.com/9FnqCbcf
> 
> I see two `postgrest-7.0.1 (lib) (configuration changed)` there.

Thanks! That's actually what I consider **once** :). I get the same output. The numbers are fine, too.

When everythig built twice for me, I had something like the following:

```
Building library for postgrest-7.0.1..
[ 1 of 18] Compiling Paths_postgrest (...)
[ 2 of 18] Compiling PostgREST.Private.Common (...)
[...]
[ 1 of 18] Compiling Paths_postgrest (...) [HPC flags changed]
[ 2 of 18] Compiling PostgREST.Private.Common (...) [HPC flags changed]
[...]
```

(This is written from memory, so might be slightly different)

So every module (exe, lib, spec, ...) was built twice immediately after each other. And that was just the `build` command. And then for the `test` command everything was built again.

This took ages and needed a rebuild everytime I ran `postgrest-coverage`.

Maybe I just misunderstood because of this background.

@monacoremo, which kind of results are you getting?

wolfgangwalther:
> > When running postgrest-clean + postgrest-coverage how do the numbers look like?
> 
> I got the same results:
> 
> pastebin: https://pastebin.com/8ZrEjmpk

Cool, that looks exactly like it does for me.

monacoremo:
@wolfgangwalther on your latest commit, it built only once consistently https://pastebin.com/Ya8Y9faj

The results are a bit different from what Steve had, is that due to the overlay you removed? 

```
 87% expressions used (8984/10299)
 63% boolean coverage (136/213)
      52% guards (64/121), 43 always True, 2 always False, 12 unevaluated
      85% 'if' conditions (64/75), 5 always True, 3 always False, 3 unevaluated
      47% qualifiers (8/17), 9 always True
 75% alternatives used (623/829)
 92% local declarations used (449/484)
 71% top-level declarations used (402/561)
```

nix/devtools.nix
```diff
@@ -72,6 +72,9 @@ let
       }
       ''
         ${cabal-install}/bin/cabal v2-clean
+        # clean old coverage data, too
+        rm -rf .hpc coverage
+        find . -type f -iname "*.tix" -delete
```
monacoremo, line 77: This is probably no longer needed with the temp dir?

.gitignore
```diff
@@ -18,3 +18,6 @@ dist-newstyle
 postgrest.hp
 postgrest.prof
 __pycache__
+*.tix
```
monacoremo, line 21: This is optional now I think, but doesn't hurt to leave it in for further experimentation on coverage

monacoremo (approved):
As noted in https://github.com/PostgREST/postgrest/pull/1694#issuecomment-752405874, the results from my last local run and CI are very slightly off - but it's close enough and we can investigate this further at a later point, good to merge!

nix/devtools.nix
wolfgangwalther: Correct, the tix files don't need to be deleted. I'll leave the `rm -rf .hpc coverage` in, though. Those still exist.

wolfgangwalther:
> The 'expressions used' are off by 6, everything else is identical.

Wow, great observation. I can actually reproduce this: Running it inside `nix-shell` I get the same numbers you have. Running it after installing with `nix-env`, I get the CI numbers.

This kind of matches the error that is thrown when running it outside `nix-shell` (the one that is hidden with `|| true`).

> As noted in [#1694 (comment)](https://github.com/PostgREST/postgrest/pull/1694#issuecomment-752405874), the results from my last local run and CI are very slightly off - but it's close enough and we can investigate this further at a later point, good to merge!

I agree. Right now we need a baseline coverage on master first, debugging everything else against that baseline will then be easier.

wolfgangwalther:
> I can actually reproduce this: Running it inside `nix-shell` I get the same numbers you have. Running it after installing with `nix-env`, I get the CI numbers.

That didn't hold up. I was able to get both numbers both inside and outside of nix-shell. So there seems to be some randomness involved. Most likely related to the io tests - there could be concurrency issues when all those processes are trying to write to the same .tix file. I will try to use the `HPCTIXFILE` environment variable to get every test to write to a separate tix file.

wolfgangwalther:
> The 'expressions used' are off by 6, everything else is identical.

I was able to track it down. The difference is in this line:

https://github.com/PostgREST/postgrest/blob/568477b6afc22c18335db450c28ef77e97690980/src/PostgREST/Config.hs#L133

In one of the cases the condition is marked as "always true" and in the other case it is marked as "always false". The next two lines are either covered or not accordingly, creating the difference in the number of expressions covered.

I have no idea why the gitHash is sometimes detected and sometimes it's not. @steve-chavez @monacoremo anyone?

PR #1696, opened 2020-12-17
fix: catch ReadTimeout in IO tests

monacoremo:
This should fix https://github.com/PostgREST/postgrest/pull/1682#issuecomment-744110866


steve-chavez:
LGTM! :+1: 

I've ran `postgrest-test-io` several times and the timeout error is no more.

Issue #1421, opened 2019-11-29
How to 


ppKrauss:
I not see at documentation about  *"how to show the JSON schema of only one table"*. For example table `public.t`  need only the `"{paths": {"/t": {etc}}}` JSON schema.

steve-chavez:
You could use jq for that. I think maybe like: 
```bash
curl localhost:3000/ | jq '.paths."/t"'
```

ppKrauss:
Hum... but the traffic (and visual pollution) is preserved.  I have ~600 tables to be published... Suggestion for *new issue*: an option in the PostgREST API to access table-specific metadata.

ppKrauss:
@steve-chavez, thanks. Hum... but the traffic (and visual pollution) is preserved.  I have ~600 tables to be published... Suggestion for *new issue*: an option in the PostgREST API to access table-specific metadata.

steve-chavez:
@ppKrauss In theory that would be possible with the `root-spec` config(released but undocumented). See my comment in https://github.com/PostgREST/postgrest/issues/790#issuecomment-495330657.

ppKrauss:
> 
> 
> @ppKrauss In theory that would be possible with the `root-spec` config(released but undocumented). See my comment in [#790 (comment)](https://github.com/PostgREST/postgrest/issues/790#issuecomment-495330657).

Hum... I not see *NGINX config script* as a program language: any super-complex solution is error-prone and an indication that PostgREST is not doing the correct/friendly thing. See my suggestion at #1425

ppKrauss:
> 
> 
> @ppKrauss In theory that would be possible with the `root-spec` config(released but undocumented). See my comment in [#790 (comment)](https://github.com/PostgREST/postgrest/issues/790#issuecomment-495330657).

Seems that you is suggesting there to use the PL function `root()`, and  here a customized one, where we can add the table name as parameter,  something as `root(tableName)`... Comments:

* the use of a custom function will be create a "Babel", no OpenAPI standard, each user will create yours, and it will be not PostgREST standard (I like the "power" of standardize of the project).

[...]

PR #1697, opened 2020-12-17
Add --dump-schema

monacoremo:
Add a `--dump-schema` option that pretty-prints the schema cache and exits.

This does not pretty print at this stage - thinking about outputting JSON, which would require adding some Aeson derivations on the DbStructure types. WDYT?

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -3,9 +3,11 @@
 module Main where
 
 import qualified Data.ByteString            as BS
+import qualified Data.ByteString.Lazy       as LBS
 import qualified Hasql.Connection           as C
 import qualified Hasql.Notifications        as N
 import qualified Hasql.Pool                 as P
+import qualified Hasql.Session              as S
 import qualified Hasql.Transaction.Sessions as HT
 
 import Control.AutoUpdate       (defaultUpdateSettings, mkAutoUpdate,
@@ -17,6 +19,7 @@ import Control.Debounce         (debounceAction, debounceEdge,
 import Control.Retry            (RetryStatus, capDelay,
                                  exponentialBackoff, retrying,
                                  rsPreviousDelay)
+import Data.Aeson               (encode)
 import Data.IORef               (IORef, atomicWriteIORef, newIORef,
                                  readIORef)
 import Data.String              (IsString (..))
@@ -64,9 +67,6 @@ main = do
   -- build the 'AppConfig' from the config file path
   conf <- readValidateConfig $ cliPath opts
 
-  -- dump config and exit if option is set
-  when (cliCommand opts == CmdDumpConfig) $ dumpAppConfig conf
-
   -- These are config values that can't be reloaded at runtime. Reloading some of them would imply restarting the web server.
   let
     host = configServerHost conf
@@ -85,6 +85,17 @@ main = do
     poolSize = configDbPoolSize conf
     poolTimeout = configDbPoolTimeout' conf
     logLevel = configLogLevel conf
+
+  case cliCommand opts of
+    CmdDumpConfig ->
+      dumpAppConfig conf
+    CmdDumpSchema ->
+      do
+        schema <- dumpSchema conf
+        putStrLn schema
+        exitSuccess
+    CmdRun ->
+      pass
 
   -- create connection pool with the provided settings, returns either a 'Connection' or a 'ConnectionError'. Does not throw.
   pool <- P.acquire (poolSize, poolTimeout, dbUri)
@@ -157,6 +168,7 @@ main = do
   whenNothing maybeSocketAddr $ do
     putStrLn $ ("Listening on port " :: Text) <> show port
     runSettings serverSettings postgrestApplication
+
 
 -- Time constants
 _32s :: Int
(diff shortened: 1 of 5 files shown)


steve-chavez:
@monacoremo This would be useful for tests right?

In that case, I think making the output JSON would make the parsing easier on the tests. Let's see what Wolfgang thinks.

Btw, I remember that at one time the DbStructure types had Aeson derivations. I believe this was for the old JSON output(not OpenAPI) at the root.

This also reminds me that it'd be useful to have a way to know the resource relationships before making embedding requests.
I thought this could be done at runtime by doing:

```http
GET /?tablename=eq.mytable

{"relationships": [....], ..}
```
(Mentioned on https://github.com/PostgREST/postgrest/issues/1421#issuecomment-561875346)

Being able to get the relationships through CLI is still really useful though.

monacoremo:
> This would be useful for tests right?

Yes! There are several uses I think:

* test startup time / time we take loading the schema cache
* debug DbStructure issues in development
* write io tests? (not sure there would be cases that are not better handled in the spec tests)
* debug user issues (what does postgrest see? Please provide postgrest --dump-schema)

steve-chavez:
Cool! Those sound like a big plus!

> debug user issues (what does postgrest see? Please provide postgrest --dump-schema)

+1 for JSON output in this case. The output could be big, so we could tell the user to pipe it through `jq` to get only the relevant json.

wolfgangwalther:
> * write io tests? (not sure there would be cases that are not better handled in the spec tests)

Maybe we can start with a simple test that just dumps the test fixtures schema to a file that we commit to the repo. The "test" would then be similar to what we do with style-check: Does running the dump produce a diff?

The effect of this would be that we basically get a changelog of what is happening with the db structure output in PRs, because the changes to this file have to be committed as well.

Might make it easier to review and reason about whats happening.

If we go that way I would suggest to pipe the dump cache output to a json2yaml converter. Yaml is much friendlier for diffs.

> > debug user issues (what does postgrest see? Please provide postgrest --dump-schema)
> 
> +1 for JSON output in this case. The output could be big, so we could tell the user to pipe it through `jq` to get only the relevant json.

Yes, I agree. JSON output seems most flexible to build on. Even though we might use a 2yaml converter later on, we should still output json first. Without doing any research on this, I assume the amount of available tooling would be better for JSON than for other formats.

wolfgangwalther:
Just adding the possibility to easily render the schema cache in json might also help with #1698.

monacoremo:
> Maybe we can start with a simple test that just dumps the test fixtures schema to a file that we commit to the repo. The "test" would then be similar to what we do with style-check: Does running the dump produce a diff?

I like this a lot! One caveat right now:

```bash
$ postgrest-dump-schema | wc -l
31878
```

Are we comfortable with checking in such a large file? It's partly that huge because e.g. tables are referenced mutltiple times, from the tables list and from each column - something that might be quite simple to sort out, but not in this PR.

wolfgangwalther:
I'm not opposed to checking in a big file in principal - because this would be mostly about a changes to that later on - and those could be relatively small.

One thing we could do to improve that would be to split up that one big file into multiple files - one for each top level key?
`dbRelations.yaml`, `dbColumns.yaml`, `dbPrimaryKeys.yaml`....

Maybe this is possible by piping it through jq first somehow?

Each file would be smaller + you could immediately tell from a diff in which part of the whole structure you are currently in from the filename (not as easy to tell from just the diff).

Of course the size of the schema plays a big role here as well. Not in this PR, but down the road we might be able to reorganize our test fixtures a bit so that we can dump not as many tables.

WDYT?

monacoremo:
> Maybe this is possible by piping it through jq first somehow?

Yes, that should be relatively straight forward! Suggest that we do that in a separate PR though. The changes as is are already useful, and it would be good to have the huge file in a separate commit.

monacoremo:
For the record, I had also experiented with `aeson-prettyprint`, but it's singnificantly slower than dumping the raw JSON, and the same can be achieved with `jq`, which you will need anyway to do anything useful with the output.

main/Main.hs
```diff
@@ -322,7 +338,46 @@ reReadConfig path refConf = do
...
+            pgVersion
+            (configDbPreparedStatements conf)
+    Right dbStructure <- S.run getDbStructureTransaction conn
+    C.release conn
+    return $ Aeson.encode dbStructure
+
+
+-- | Print the time taken to run an IO action to stderr with the given printf string
+timeToStderr :: [Char] -> IO a -> IO a
+timeToStderr fmtString a =
```
wolfgangwalther, line 363: Do we need that? I thought we'd just do something like `cabal v2-exec time -- postgrest --dump-schema`. Is there a disadvantage with that, that I don't see?

monacoremo: Intent is to get the time for just the schema cache loading, without the encoding and printing overhead. With the latest commit we should actually achieve that :-) 

The difference is quite significant:

```
[nix-shell:~/Projects/postgrest]$ test/with_tmp_db cabal v2-exec -- bash -c "time postgrest --dump-schema test/io-tests/configs/simple.config > /dev/null"
Loaded schema in 0.011 seconds

real    0m0,073s
user    0m0,050s
sys     0m0,075s
```

This would be useful to monitor the load time in CI, in the past it got out of hand for large schemas without being noticed for a while.

wolfgangwalther: Ok with the changes, I can follow your reasoning. Quite a lot of overhead.

main/Main.hs
```diff
@@ -316,16 +332,60 @@ listener dbUri dbChannel pool refConf refDbStructure mvarConnectionStatus connWo
 
 -- | Re-reads the config at runtime. Invoked on SIGUSR2.
 -- | If it panics(config path was changed, invalid setting), it'll show an error but won't kill the main thread.
+#ifndef mingw32_HOST_OS
 reReadConfig :: FilePath -> IORef AppConfig -> IO ()
 reReadConfig path refConf = do
   conf <- readValidateConfig path
   atomicWriteIORef refConf conf
   putStrLn ("Config file reloaded" :: Text)
+#endif
```
wolfgangwalther, line 341: I'm confused. Why did this not throw errors for the windows build before? What changed here?

monacoremo: I explicitly exported only main in `module Main (main) where` at the top of Main.hs to make sure I didn't have any unused symbols - it didn't make a difference for my Linux builds, but on Windows it then rightfully complained about those symbols being unsused

wolfgangwalther: Ah, ok. Thanks for the explanation!

steve-chavez (approved):
LGTM!

Really liking the json output. I could have used that several times already(used to traceShowId manually :( ).

PR #1699, opened 2020-12-20
Add test for expected schema dump

monacoremo:
As discussed in https://github.com/PostgREST/postgrest/pull/1697#issuecomment-747942662


wolfgangwalther:
Cool! That's already way better than the "one big file". I noticed that all files are arrays on the top level. I think it would be a lot easier to look at those files, if we could add 1 or 2 empty lines between each top level array item.

Maybe something like this?

https://stackoverflow.com/a/44284819

I like the last approach in there via subclassing `Dumper` in general. The implementation linked in the comment seems even better (shorter).

monacoremo:
> I think it would be a lot easier to look at those files, if we could add 1 or 2 empty lines between each top level array item.

Good idea, much more readable!

And what are a few thousand lines extra... :-) `dbRelations.yaml` is almost half a MB with over 20k lines, pretty huge...

wolfgangwalther:
Given the size of the files, I'm thinking about setting up a much smaller schema that covers all the dbStructure aspects, right now - before we merge this. My main reason is that I want to avoid a situation where the diff to `dbRelations.yaml` for a simple change is still so big, that it will not display nicely in a PR. This would render the whole thing kind of useless.

monacoremo:
Having that transparency into DbStructure will be pretty useful - some things that jump out:

* the `dbPrimaryKeys` structure seems to be very redundant, it could be a simple boolean flag on the `dbColumns`; an it seems to be only needed for the OpenAPI output, so we might to remove it completely
* each `dbColumn` holds a reference to its table, the other way round probably makes more sense
* There is some combinatorial explosion going on with the `dbRelations`, which probably also impacts runtime performance
* A lot of the joining in our code is done based on the fully qualified name, getting using the oids might make everything simpler and faster

monacoremo:
> Given the size of the files, I'm thinking about setting up a much smaller schema that covers all the dbStructure aspects, right now - before we merge this. My main reason is that I want to avoid a situation where the diff to `dbRelations.yaml` for a simple change is still so big, that it will not display nicely in a PR. This would render the whole thing kind of useless.

Yes, we can park this for now - maybe we can also get some fixes to `DbStructure` in, which would further reduce the size. And prune the test schema, not sure that we are using everything. I can rebase this periodically to see how we come along.

wolfgangwalther:
> * the `dbPrimaryKeys` structure seems to be very redundant, it could be a simple boolean flag on the `dbColumns`; an it seems to be only needed for the OpenAPI output, so we might to remove it completely

I think we need the primary key information to create the `Location` header as well - but I am not sure whether that's `dbPrimaryKeys` or from somewhere else right now.

> * each `dbColumn` holds a reference to its table, the other way round probably makes more sense

Sounds good.

> * There is some combinatorial explosion going on with the `dbRelations`, which probably also impacts runtime performance

"combinatorial explosion" - love it :)

> * A lot of the joining in our code is done based on the fully qualified name, getting using the oids might make everything simpler and faster

Yes, I noticed that as well in the `pfkSourceColumns` query. I think I changed some of this already to use oids, but there will be plenty more to do.


> Yes, we can park this for now - maybe we can also get some fixes to `DbStructure` in, which would further reduce the size. And prune the test schema, not sure that we are using everything. I can rebase this periodically to see how we come along.

Yes, I like it. This will already help us to keep track of that, while it is still an open PR. I'll copy some very basic numbers here, so we can compare later. Numbers of lines:

- 8,094 test/dump-schema/expected/dbColumns.yaml
- 825 test/dump-schema/expected/dbPrimaryKeys.yaml
- 1,828 test/dump-schema/expected/dbProcs.yaml
- 21,487 test/dump-schema/expected/dbRelations.yaml
- 714 test/dump-schema/expected/dbTables.yaml

test/dump-schema/update.py
```diff
@@ -0,0 +1,38 @@
...
+import json
+import os
+import pathlib
+import shutil
+import subprocess
+import yaml
+
+
+BASEDIR = pathlib.Path(os.path.realpath(__file__)).parent
+# TODO: remove config file as soon as env config is merged
```
wolfgangwalther, line 11: bump!

monacoremo: so much nicer! :-)

wolfgangwalther:
I gave this one a rebase, I hope you don't mind, @monacoremo. I'm curious how our numbers look right now, are Steve merged #1825.

wolfgangwalther:
> Yes, I like it. This will already help us to keep track of that, while it is still an open PR. I'll copy some very basic numbers here, so we can compare later. Numbers of lines:
> 
>  * 8,094 test/dump-schema/expected/dbColumns.yaml
> 
>  * 825 test/dump-schema/expected/dbPrimaryKeys.yaml
> 
> * 1,828 test/dump-schema/expected/dbProcs.yaml
> 
> * 21,487 test/dump-schema/expected/dbRelations.yaml
> 
> * 714 test/dump-schema/expected/dbTables.yaml

We now have:

* 7,324 test/dump-schema/expected/dbColumns.yaml
* 1,000 test/dump-schema/expected/dbPrimaryKeys.yaml
* 1,601 test/dump-schema/expected/dbProcs.yaml
* 23,095 test/dump-schema/expected/dbRelations.yaml
* 814 test/dump-schema/expected/dbTables.yaml
* 3 test/dump-schema/expected/pgVersion.yaml

I have no idea what that means.

steve-chavez:
> From: 21,487 test/dump-schema/expected/dbRelations.yaml
> To:     23,095 test/dump-schema/expected/dbRelations.yaml

Hm, I expected this number to go down. One thing I've noticed from the output is that many fields are unneeded(`colDescription`, `tableDescription`, `tableInsertable`, etc) and they add a lot of verbosity.

What relationships really needs are references to columns and tables `QualifiedIdentifiers`, this could be a potential refactor.

wolfgangwalther:
> Hm, I expected this number to go down. One thing I've noticed from the output is that many fields are unneeded(`colDescription`, `tableDescription`, `tableInsertable`, etc) and they add a lot of verbosity.

Yeah.. we need to get rid of `OpenApi.hs` so much... that will allow us to throw all that out completely!

> What relationships really needs are references to columns and tables `QualifiedIdentifiers`, this could be a potential refactor.

Yeah, although that would only affect dump-size, I think. Internally that should all be references, only, right?

steve-chavez:
>  However there were no PrimaryKey, Column or Table removed at all. I wonder whether that's because of how our test fixtures are set up - or whether we actually don't need the filters for all of those, because there is no such data.

Actually I remember seeing some cases of internal pks/columns/tables when doing the `request.spec`. It's just that they're few, that's why the numbers don't go down, I think. So the filters would be necessary.

(Just in case, the unfiltered `request.spec` [output](https://gist.githubusercontent.com/steve-chavez/bccb5ae9f426fd247bedec3cebdc76a2/raw/e6d59fea66a05b177c6c129a7de5b491d2f60a5d/root_override.json) does prove this, search for `private`)

> Yeah, although that would only affect dump-size, I think. Internally that should all be references, only, right?

Ah, you know I'm not sure, but I think they're copies, References are only possible through `IORef` and the like, from my understanding.

wolfgangwalther:
> > However there were no PrimaryKey, Column or Table removed at all. I wonder whether that's because of how our test fixtures are set up - or whether we actually don't need the filters for all of those, because there is no such data.
> 
> Actually I remember seeing some cases of internal pks/columns/tables when doing the `request.spec`. It's just that they're few, that's why the numbers don't go down, I think. So the filters would be necessary.

Hm. I went through the diff and have only green lines, so no deletions in those files. Didn't check the others, but the committed table file does still have the private tables you mentioned:
https://github.com/PostgREST/postgrest/blob/8edc1adb3243ee5c7b731cacb1cab36a5e4a575a/test/dump-schema/expected/dbTables.yaml

But there's something strange going on, too, because the schema-check failed in CI - but did not fail for me locally. And, when I checked your PR directly I had a decrease in dump file size by 1/3, I think. I'm not sure whether the dump I got and committed here is correct.

wolfgangwalther:
> > 
> 
> Ah, you know I'm not sure, but I think they're copies, References are only possible through `IORef` and the like, from my understanding.
> 
> **Edit**: More info about references:

Yeah, I think references was the wrong word. But they are "not copied". Actually not evaluated until we do the dump. Right?

wolfgangwalther:
Ah, unlike most of the other commands `postgrest-dump-schema-update` does not build automatically before. I didn't run `postgrest-build` before. Now, the files are **a lot** smaller:

* 6,568 test/dump-schema/expected/dbColumns.yaml
* 783 test/dump-schema/expected/dbPrimaryKeys.yaml
* 1,601 test/dump-schema/expected/dbProcs.yaml
* 14,915 test/dump-schema/expected/dbRelationships.yaml
* 674 test/dump-schema/expected/dbTables.yaml
* 3 test/dump-schema/expected/pgVersion.yaml

steve-chavez:
> Now, the change between these numbers here and the numbers I posted earlier today is only the difference from #1825. Good Job, Steve!

Awesome, thanks! The numbers make sense now.

> Actually not evaluated until we do the dump. Right?

Yes, because of [lazy evaluation](https://wiki.haskell.org/Lazy_evaluation).

Issue #1698, opened 2020-12-18
Split off openapi into postgrest-contrib


wolfgangwalther:
# Problem

OpenAPI support is not up to date: We support v2, but v3 is out there for some years now (#932). Currently we have 26 issues open that are tagged "OpenAPI".

I'm sure some of those issues are not that hard to solve in theory. I feel like having the OpenAPI code in haskell is quite a hurdle for community contributions, though. The majority of our users probably does not know haskell at all - but would be able to make contributions in SQL.

Test coverage seems to be quite low for that area, too. I recently played a bit with #1653 - but was surprised to see that I didn't get any failing tests after changing some things.

# Suggestion

I suggest to split off all the openapi related code. Remove it from core postgrest, put it in a separate (still official) `postgrest-contrib` repo and turn all the code into SQL. We already have the `db-root-spec` config option, to be able to use it.

Then we ship PostgREST without openapi support at all. In the docs we are very clear about how to add openapi support from our "official extension". It's basically just running a SQL script to create the function(s) + setting the config option. Without this config option, the root endpoint does not return anything (or maybe a hint on how to add openapi support, if that's better).

I see the following advantages of this:
- Less complex codebase for postgrest core
- Fewer stuff to load with the schema cache, might make startup faster
- Hopefully increased community contributions for the openapi parts, because of lower hurdles to do so
- Ability to provide OpenAPI v3 support in addition to v2 without breaking changes: Users are able to choose which set of functions to install.
- Ability to disable OpenAPI output by not providing a `db-root-spec` config at all
- Ability to extend / change the output specifically for a single project (currently root-spec is available as well, but you would still have to reproduce the whole code in SQL first)
- Ability to add root endpoint filtering (#1421) by treating the `db-root-spec` *mostly* like a regular RPC (with some small tweaks, it should be able to add something powerful)

The `postgrest-contrib` repo could have other SQL-only stuff, too. E.g. #1690 and other SQL snippets we are suggesting in the docs right now. We can have several distribution formats for this repo, e.g. as pg extension or just as one big SQL file (for cloud hosted pg servers). Users can install just some parts of this or everything - and then just enable those functions in the postgrest configuration.

Having this in a separate repo should make it a bit easier to track on the issue tracker. Also, while some SQL parts might be postgrest-version dependent, most parts while likely be usable across more versions. Having a different repo (and multi-version compatibility in mind) should make it easier to perform major version upgrades of postgrest - without the need to update the SQL code at the same time.

[...]

PR #1701, opened 2020-12-21
fix: fix OpenAPI pk/fk note not being overridable

### LINKED ISSUE #1700: "Note: this is a Primary Key"  is not overridable with a comment
If I add a comment to a pk column:

```sql
COMMENT ON COLUMN articles.id IS 'An article primary key';
```

On the OpenAPI output, I get:

```bash
curl -s localhost:3000 | jq .definitions.articles.properties.id

{
  "format": "integer",
  "type": "integer",
  "description": "An article primary key\n\nNote:\nThis is a Primary Key.<pk/>"
}
```

The redundant `Note:\nThis is a Primary Key.<pk/>` at the end should be overridden by the comment.

--- 

According to https://github.com/PostgREST/postgrest/pull/885#issuecomment-681967336, the `<pk>`(and `<fk>`) notes, were a hack for inserting metadata.

Ideally we'd add metadata by using [OpenAPI extensions](https://swagger.io/docs/specification/openapi-extensions/). The property could have an additional `x-primary-key: true` field.

steve-chavez:
Resolves #1700.

I've just allowed overriding the notes with the db comment.

I looked into adding `x-primary-key` or `x-foreign-key` fields to not lose the metadata, but seems this is not easy with our current swagger lib: https://github.com/GetShopTV/swagger2/issues/113#issuecomment-558185318.

Still I think this is a better behavior, I'll leave adding the extensions for our future OpenAPI-in-SQL https://github.com/PostgREST/postgrest/issues/1698.

### DIFF
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -16,7 +16,7 @@ import Data.Aeson                 (decode, encode)
 import Data.HashMap.Strict.InsOrd (InsOrdHashMap, fromList)
 import Data.Maybe                 (fromJust)
 import Data.String                (IsString (..))
-import Data.Text                  (append, breakOn, dropWhile, init,
+import Data.Text                  (breakOn, dropWhile, init,
                                    intercalate, pack, tail, toLower,
                                    unpack)
 import Network.URI                (URI (..), URIAuth (..))
@@ -68,20 +68,19 @@ makeProperty pks c = (colName c, Inline s)
       intercalate "" ["This is a Foreign Key to `", a, ".", b, "`.<fk table='", a, "' column='", b, "'/>"]
     pk :: Bool
     pk = any (\p -> pkTable p == colTable c && pkName p == colName c) pks
-    n = catMaybes
+    notes = catMaybes
       [ Just "Note:"
       , if pk then Just "This is a Primary Key.<pk/>" else Nothing
       , fk <$> colFK c
       ]
-    d =
-      if length n > 1 then
-        Just $ append (maybe "" (`append` "\n\n") $ colDescription c) (intercalate "\n" n)
-      else
-        colDescription c
+    additionalProp = if length notes > 1
+      then Just $ AdditionalPropertiesSchema . Inline $ (mempty :: Schema) & description ?~ intercalate "\n" notes
+      else Nothing
     s =
       (mempty :: Schema)
         & default_ .~ (decode . toS =<< colDefault c)
-        & description .~ d
+        & description .~ colDescription c
+        & additionalProperties .~ additionalProp
         & enum_ .~ e
         & format ?~ colType c
         & maxLength .~ (fromIntegral <$> colMaxLen c)
--- a/test/Feature/MultipleSchemaSpec.hs
+++ b/test/Feature/MultipleSchemaSpec.hs
@@ -217,7 +217,9 @@ spec actualPgVersion =
                     "type" : "object",
                     "properties" : {
                       "id" : {
-                        "description" : "Note:\nThis is a Primary Key.<pk/>",
+                        "additionalProperties": {
+                          "description": "Note:\nThis is a Primary Key.<pk/>"
+                        },
                         "format" : "integer",
                         "type" : "integer"
                       },
@@ -246,7 +248,9 @@ spec actualPgVersion =
                     "type" : "object",
                     "properties" : {
                       "id" : {
-                        "description" : "Note:\nThis is a Primary Key.<pk/>",
+                        "additionalProperties": {
+                          "description": "Note:\nThis is a Primary Key.<pk/>"
+                        },
                         "format" : "integer",
                         "type" : "integer"
                       },
@@ -275,7 +279,9 @@ spec actualPgVersion =
                     "type" : "object",
                     "properties" : {
                       "id" : {
-                        "description" : "Note:\nThis is a Primary Key.<pk/>",
+                        "additionalProperties": {
+                          "description": "Note:\nThis is a Primary Key.<pk/>"
+                        },
                         "format" : "integer",
                         "type" : "integer"
                       },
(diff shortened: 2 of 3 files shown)


wolfgangwalther:
Uh. This immediately breaks my client lib `vue-postgrest`, once a comment on a PK column is added. I rely on `<pk/>` to be there, to identify those columns.

steve-chavez:
@wolfgangwalther Ah, I didn't think anyone was relying on those notes. Would adding an `x-primary-key: true` work for you?

Also, do you use the foreign key `<fk/>` note?

wolfgangwalther:
> @wolfgangwalther Ah, I didn't think anyone was relying on those notes. Would adding an `x-primary-key: true` work for you?

I haven't looked deeper into that, but I guess it would. I just need to know which columns identify records on a route uniquely, so that I can use those for all my `model.$patch(..)`, `model.$delete(...)`, `model.$get(...)`, etc. calls. I am using the nightly in a project currently under development, so it would be cool if we had an immediate replacement when removing this.

> Also, do you use the foreign key `<fk/>` note?

Nope, I can't imagine a use-case. I would hope to have "embedding opportunities" available in the future, that would help. Better to know about those before, than just after a 300 response... But I'm not going to derive those from `<fk/>` ;).

steve-chavez:
Since vendor extensions are not supported https://github.com/GetShopTV/swagger2/issues/4 (and there's no workaround but to patch upstream). I'm thinking about adding the pk/fk notes on the `additionalProperties` field, like:
```json
curl -s localhost:3000 | jq .definitions.articles.properties
{
  "id": {
    "format": "integer",
    "additionalProperties": {
      "description": "Note:\nThis is a Primary Key.<pk/>\n"
    },
    "type": "integer",
    "description": "An article primary key"
  },
  "body": {
    "format": "text",
    "type": "string"
  },
  "owner": {
    "format": "name",
    "type": "string"
  }
}
```

@wolfgangwalther Would that work for you?

steve-chavez:
I think it'd just be a matter of changing:

```js
//from
fieldDef.description?.includes('<pk/>'))

//to 
fieldDef.description.additionalProperties.description?.includes('<pk/>'))
```

On https://github.com/technowledgy/vue-postgrest/blob/4dd955799dddcdffafc022264eb1242bddb2e2eb/src/Route.js#L22.

So I'll go ahead with the change :)

wolfgangwalther:
> Since vendor extensions are not supported [GetShopTV/swagger2#4](https://github.com/GetShopTV/swagger2/issues/4) (and there's no workaround but to patch upstream). I'm thinking about adding the pk/fk notes on the `additionalProperties` field, like:
> 
> ```json
> curl -s localhost:3000 | jq .definitions.articles.properties
> {
>   "id": {
>     "format": "integer",
>     "additionalProperties": {
>       "description": "Note:\nThis is a Primary Key.<pk/>\n"
>     },
>     "type": "integer",
>     "description": "An article primary key"
>   },
>   "body": {
>     "format": "text",
>     "type": "string"
>   },
>   "owner": {
>     "format": "name",
>     "type": "string"
>   }
> }
> ```
> 
> @wolfgangwalther Would that work for you?

I couldn't find any docs on `additionalProperties` in that context yet... but if this field does in fact allow arbitrary keys, like the name suggests, why don't you do it like this?

```
"id": {
    "format": "integer",
    "additionalProperties": {
      "pk": true
    },
    "type": "integer",
    "description": "An article primary key"
  },
```

steve-chavez:
Not possible to add a `pk` field like that, the types are too strict: https://hackage.haskell.org/package/swagger2-2.6/docs/Data-Swagger.html#t:Schema

Also, the description format doesn't change in this PR. That would help other users that also relied on that behavior, easier migration.

Besides that, I don't think it's worth to put much work on this since #1698 is the way forward.

wolfgangwalther:
But to be honest, I don't think `additionalProperties` is to be used in either of those two ways. This is something entirely different, and not valid for a `"type": "integer"` at all.

My understanding is that `additionalProperties` allows to define a schema for additional properties of a field with `"type": "object"`.  Using a description on this non-existent schema to somehow hide the `<pk/>` information.... sounds really fishy.

What is the urgency in removing this from the description? Does the "Note: ..." break any usecase? Why can't it wait until we solve the openapi stuff properly in SQL and we don't have to add some strange hacks here?

wolfgangwalther:
> Not possible to add a `pk` field like that, the types are too strict: https://hackage.haskell.org/package/swagger2-2.6/docs/Data-Swagger.html#t:Schema

Ah yes, those Types match my expectation and just confirm that adding the note in the description of an otherwise non-existent sub-schema is really... not valid openapi at all.

wolfgangwalther:
To summarize: It feels like this PR is just replacing one hack with another, worse, hack. It was never the best idea to add `<pk/>` in the `description` in the first place, but it is certainly a worse idea to add it in `additionalProperties.description`.

wolfgangwalther:
I only need `<pk/>` though, so maybe it would be enough to just remove the whole "Note: This is a Primary Key." thing and **only** add the  `<pk/>`? It would certainly be less obtrusive.

steve-chavez:
> To summarize: It feels like this PR is just replacing one hack with another, worse, hack.

Yeah, my original intention was to remove the hack. But since that breaks `vue-postgrest` I went with the `additionalProperties` workaround. As you mention that seems invalid for an integer type, so I'll not proceed because maybe it could break other stuff as well(like code generators).

I'll just close this one. Let's solve this later as part of #1698.

steve-chavez:
> Does the "Note: ..." break any usecase? 

Ah, yes. It breaks OpenAPI UIs, as the notes don't go away despite adding a comment for the column. The `<pk>`/`<fk/>` look a bit like "garbage" for the end user.

This happens for [supabase.io](https://supabase.io/), which shows our OpenAPI comments.

Not sure how users of Swagger UI are cleaning the comments or if it was ever an issue there. But in Supabase's case, the comments are now cleaned with js on frontend.

robertsosinski:
Adding on here, would it be possible to override this behavior with something like: `comment on column <insert column here>.id is 'Unique id for this table.';`?  I can see the current Note + </pk> as a sensible default (as well as for backwards compatibility). But if I want to change this behavior, would be great to be able to, and if I want the "</pk>" there I can add it in my own comment.

Thanks!

Issue #1672, opened 2020-11-30
Stored Procedure that returns a table type is colliding with another homonym stored procedure though having different parameters signature, resulting in unexpected errors


Pigeo:
### Environment

* PostgreSQL version: PostgreSQL 12.4 (Debian 12.4-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
* PostgREST version: PostgREST 7.0.1
* Operating system: Debian 10 (buster) / Linux LABO-BDD-dev 4.15.18-27-pve #1 SMP PVE 4.15.18-55 (Tue, 17 Mar 2020 15:32:02 +0100) x86_64 GNU/Linux

### Steps to reproduce:

1. In your public schema, create a function with an optional TEXT param, and returning some rows of a table:

```
create function foobar(optional_param TEXT DEFAULT '') RETURNS SETOF films AS
$$
BEGIN
  RETURN QUERY SELECT * from films;
END;
$$ LANGUAGE plpgsql STABLE;
```

2. A query like `/rpc/foobar?select=id,title,actors!characters(name)` works as expected.

3. Now in the same schema, add another function, with same name, but different signature (e.g. mandatory param of row type + returning a JSONB):

```
create function foobar(mandatory_param "REST_API_v1_0".lemmas) RETURNS JSONB
AS $$
BEGIN
   RETURN '{"foo":"bar"}'::jsonb;
END;
$$ LANGUAGE plpgsql STABLE;
```

3. Here comes the bug: the same query `/rpc/foobar?select=id,title,actors!characters(name)` (as above) doesn't work anymore: 

```
{"message": "Could not find foreign keys between these entities. No relationship found between pgrst_source and actors"}
```

Even a simpler query like `/rpc/foobar?select=id,title` doesn't work anymore :

```
{
    "hint": null,
    "details": null,
    "code": "42703",
    "message": "column pgrst_source.id does not exist"
}
```

(Whereas simplest query `/rpc/foobar?select=*` still works)

=> The bug seems related to the process of verifying existence of given column names (and/or existence of relationship through join tables), which fails to correctly select the function which is returning a SETOF rather than the other homonym function.

4. Work-around: even if they are optional, specify at least one optional parameter (with its default value) in the query, thus giving PostgREST a supplemental hint to identify the proper function.

```
{{ThesocUrl}}/rpc/foobar?select=id,title,actors!characters(name)&optional_param=
```

[...]

Issue #1643, opened 2020-11-03
Unable to disambiguate select of embedded table in presence of view(self-reference)


dvalentiate:
### Environment

* PostgreSQL version: postgres 11
* PostgREST version: postgrest/postgrest:v7.0.1
* Operating system: MacOS

### Description of issue

It seems having a view disrupts the ability to do a recursive embedded view on table.

The following is the minimal for our recursive table:

```sql
CREATE TABLE api.test (
  id BIGINT NOT NULL PRIMARY KEY,
  parent_id BIGINT CONSTRAINT parent_test REFERENCES api.test(id)
);
GRANT SELECT ON api.test TO api_anon;

INSERT INTO api.test (id, parent_id) VALUES (1, null);
INSERT INTO api.test (id, parent_id) VALUES (2, 1);
```

And `curl -s 'http://bluedrop.test:3000/test?select=id,parent_id,parent_test:parent_id!parent_test(id)' | jq .` gets:

> ```[
>   {
>     "id": 1,
>     "parent_id": null,
>     "parent_test": null
>   },
>   {
>     "id": 2,
>     "parent_id": 1,
>     "parent_test": {
>       "id": 1
>     }
>   }
> ]

Which is fine and the output we are hoping for, **but** then if we add this view:

```sql
CREATE OR REPLACE VIEW api.view_test AS
  SELECT api.test.id FROM api.test
;
```

The same query is now deemed ambiguous:

> ```{
>   "hint": "By following the 'details' key, disambiguate the request by changing the url to /origin?select=relationship(*) or /origin?select=target!relationship(*)",
>   "details": [
>     {
>       "relationship": "parent_test[parent_id][id]",
>       "origin": "api.test",
>       "cardinality": "m2o",
>       "target": "api.test"
>     },
>     {
>       "relationship": "parent_test[parent_id][id]",
>       "origin": "api.test",
>       "cardinality": "m2o",
>       "target": "api.view_test"
>     }
>   ],
>   "message": "More than one relationship was found for test and parent_id"
> }

I can't figure out a valid target and hint that allows for an embed m2o nested object like is returned when the view doesn't exist. I'm not sure if this is a bug or just something I haven't figured out yet.

wolfgangwalther:
The hint says: `disambiguate the request by changing the url to [...] /origin?select=target!relationship(*)`

This leads me to the following two URLs:

- `/test?select=*,test!parent_test(id)`
- `/test?select=*,view_test!parent_test(id)`

[...]

PR #1702, opened 2020-12-21
add hollywood.sql test fixtures

wolfgangwalther:
Related to https://github.com/PostgREST/postgrest/pull/1699#issuecomment-748620675.

This adds another schema to the test fixtures with the primary goal of having something manageable in size for the "dump schema" test. Taking #1698 into account not everything that might currently be in the schema cache will be needed down the road. I identified 3 core features where postgrest core relies on the schema cache: **embedding relations**, **calling RPCs** and **creating `Location` headers after INSERT**.  Are there any other features I am missing right now?

To achieve proper test coverage with the "dump schema" test, I'd like to cover many of the items in the list below. Open for suggestions about what is missing. Some of those potentially targeted features are not even implemented, yet - so some of those "test cases" could yield interesting behaviour.

Of course there could be secondary goals achieved with those new fixtures:
- Have an example schema for the docs that covers all of the postgrest functionality (that's why I based it on the actors/films examples)
- Use those fixtures in the spec tests

What else comes to mind?

----

### Parsing Base Columns in Views
*Needed for both embedding and INSERT (location headers, on conflict)*

- [x] simple views (`SELECT * FROM table`)
- [x] views with subquery in `SELECT` (https://github.com/PostgREST/postgrest/pull/1251)
- [x] views with subquery inside function call in `SELECT` (https://github.com/PostgREST/postgrest/issues/1608)
- [x] views with `GROUP BY` aggregation
- [ ] views with CTEs
- [x] views with UNION
- [x] recursive views
- [x] base columns from relation in `FROM` (https://github.com/PostgREST/postgrest/issues/1230)
- [x] base columns from relation in `JOIN` (https://github.com/PostgREST/postgrest/issues/1230)
- [x] 2+ level view chains (https://github.com/PostgREST/postgrest/issues/1607)

#### Basics
- [x] O2M with simple key
- [x] M2O with simple key
- [x] M2M with simple keys
- [x] O2M with compound key
- [x] M2O with compound key
- [x] M2M with compound key(s) (https://github.com/PostgREST/postgrest/pull/1593)
- [x] M2M junctions without extra columns
- [x] M2M junctions with extra columns

#### Hints / Query Syntax
- [x] Embedding via column name
- [x] Embedding via fk constraint name (named/unnamed)
- [x] Embedding via table name
- [x] Embedding via cardinality
- [ ] Manual embedding hint via no-row UNION (https://github.com/PostgREST/postgrest/issues/1179)

#### Inheriting Relationships for Views
- [x] views with table in same schema
- [x] views with hidden table and a different name
- [x] views with hidden table and the same name (https://github.com/PostgREST/postgrest/issues/1587)
- [x] m2m junction views with hidden table and the same name (https://github.com/PostgREST/postgrest/issues/1587)
- [x] view chains with middle view in exposed schema (https://github.com/PostgREST/postgrest/pull/1625)
- [x] view chains with middle view in extra.-search-path (https://github.com/PostgREST/postgrest/pull/1625)

#### Self-References
- [x] O2M/M2O self-reference with simple key
- [ ] O2M/M2O self-reference with compound key
- [x] M2M self-reference with simple key
- [ ] M2M self-reference with compound key
- [x] self-references with views of same table (https://github.com/PostgREST/postgrest/issues/1643)

### Location Headers
- [x] simple pk on table
- [x] compound pk on table
- [ ] no pk, but unique on table
- [ ] no pk, no unique on table
- [x] simple pk on view
- [x] compound pk on view
- [x] no pk, but unique on view
- [ ] no pk, no unique on view
- [x] multiple unique on same relation
- [ ] multiple pks on same view

#### Basics
- [x] marked `VOLATILE`
- [x] marked `STABLE`
- [ ] marked `IMMUTABLE`
- [x] returns set of composite
- [x] returns single composite
- [x] returns set of scalar
- [x] returns single scalar
- [x] returns void
- [ ] returns domain of composite
- [ ] returns domain of scalar

#### Function Overloading
- [x] basic arguments with different names
- [x] variadic argument
- [x] arguments with default values
- [ ] only one argument with default value (https://github.com/PostgREST/postgrest/issues/1672)
- [x] ignore functions with unnamed arguments
- [x] work with single unnamed argument function and `Prefer: params=single-object` (https://github.com/PostgREST/postgrest/issues/1469)
- [ ] arguments with same name but different type


steve-chavez:
Nice! This is really comprehensive. 

>  embedding relations, calling RPCs and creating Location headers after INSERT. Are there any other features I am missing right now?

Some other features I remember needing the schema cache:

The ON CONFLICT for INSERT needs the PK:

https://github.com/PostgREST/postgrest/blob/bf141ca13f23818a443a4239eab7085176da02aa/src/PostgREST/DbRequestBuilder.hs#L291-L297

The RETURNING part needs to know about PK/FK(to remove them or add them automatically):

https://github.com/PostgREST/postgrest/blob/bf141ca13f23818a443a4239eab7085176da02aa/src/PostgREST/DbRequestBuilder.hs#L327-L330

wolfgangwalther:
I don't see myself picking up this PR again.

PR #1703, opened 2020-12-22
Upgrade Nixpkgs and Stackage, use GHC 8.8.4

monacoremo:
* Upgraded nixpkgs to the current master and added the following changes:
  * Use GHC 8.8.4
  * Removed our patch for the GHC 8.6.5 bootstrap on `pkgsMusl`, as that was fixed upstream
  * Added a patch for `pkgsMusl.gdb` https://github.com/NixOS/nixpkgs/issues/104133 (reverting the commit that updated gdb and caused the issue)
  * Several packages are upgraded in the new Nixpkgs, uncluding PostgreSQL to the latest patch releases/minor versions
* Upgraded stackage to 16.26, which also uses GHC 8.8.4

Our patch to the OpenSSL package https://github.com/NixOS/nixpkgs/pull/87879 is 'Ready' in [nixpkgs staging](https://github.com/NixOS/nixpkgs/projects/25), so hopefully that will be merged sometime soon (causing a pretty huge rebuild of nixpkgs, as pretty much everything depends on OpenSSL) and we can remove our patch.

The cachix update in the background is courtesy of my brave laptop, who worked very hard for many hours to put everything together, building everything on the static/musl side from scratch, including OpenSSL and GHC.


nix/overlays/haskell-packages.nix
```diff
@@ -18,33 +21,14 @@ let
...
       hasql-dynamic-statements =
-        self.haskell.lib.dontCheck (prev.callHackageDirect
-          {
-            pkg = "hasql-dynamic-statements";
-            ver = "0.3.1";
-            sha256 = "1zjv91xlfkyxwq6mhzj7rsfm4kjvs9ygkgbl6jbbg19jihcn2kiy";
-          }
-          { }
-        );
+        lib.dontCheck (lib.unmarkBroken prev.hasql-dynamic-statements);
```
steve-chavez, line 25: [unmarkBroken](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/lib.nix#L172), nice escape hatch.

steve-chavez (approved):
LGTM!

Laptop's effort won't be in vain :)

nix/overlays/haskell-packages.nix
monacoremo: Or `markUnbroken`, equivalently - all the choices you might want :-)

PR #1704, opened 2020-12-23
Improvements to io-tests

monacoremo:
* [x] isolate the postgrest process from env
* [x] randomize port
* [x] parallelize io-tests

Will rebase on #1691 once merged

Edit: rebased, and refactored a bit using our new env vars. The environment that we run the PostgREST process in is now fully explicit and controlled, much better than the setup we had before.

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -168,7 +168,7 @@ The io-test that test PostgREST as a black box with inputs and outputs can be
 run with `postgrest-test-io`. The test runner under the hood is
 [pytest](https://docs.pytest.org/) and you can pass it the usual options:
 
-```
+```bash
 # Filter the tests to run by name, including all that contain 'config':
 postgrest-test-io -k config
 
--- a/nix/README.md
+++ b/nix/README.md
@@ -160,7 +160,21 @@ $ nix-shell --run postgrest-test-spec-all
 
 # Run the tests against a specific version of PostgreSQL (use tab-completion in
 # nix-shell to see all available versions):
-$ nix-shell --run postgrest-test-spec-postgresql-9.5
+$ nix-shell --run postgrest-test-spec-postgresql-13
+
+```
+
+The io-test that test PostgREST as a black box with inputs and outputs can be
+run with `postgrest-test-io`. The test runner under the hood is
+[pytest](https://docs.pytest.org/) and you can pass it the usual options:
+
+```
+# Filter the tests to run by name, including all that contain 'config':
+postgrest-test-io -k config
+
+# Run tests in parallel using xdist, specifying the number of processes:
+postgrest-test-io -n auto
+postgrest-test-io -n 8
 
 ```
 
--- a/test/io-tests/test_io.py
+++ b/test/io-tests/test_io.py
@@ -97,6 +97,7 @@ def run(configpath=None, stdin=None, env=None, serversocket=None):
     else:
         port = freeport()
         env["PGRST_SERVER_PORT"] = str(port)
+        env["PGRST_SERVER_HOST"] = "localhost"
         baseurl = f"http://localhost:{port}"
 
     command = [POSTGREST_BIN]
--- a/test/io-tests/test_io.py
+++ b/test/io-tests/test_io.py
@@ -8,6 +8,7 @@ import os
 import pathlib
 import shutil
 import signal
+import socket
 import subprocess
 import time
 import urllib.parse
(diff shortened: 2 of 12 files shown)


wolfgangwalther:
You can rebase and use environment variables everywhere now - #1691 is merged!

monacoremo:
@wolfgangwalther Awesome! Was looking forward to the env vars change for a long time. There is more refactoring that we could do in the io-tests with those, but I think it's good for now - ready for review/merge.

nix/README.md
```diff
@@ -150,7 +150,21 @@ $ nix-shell --run postgrest-test-spec-all
...
+The io-test that test PostgREST as a black box with inputs and outputs can be
+run with `postgrest-test-io`. The test runner under the hood is
+[pytest](https://docs.pytest.org/) and you can pass it the usual options:
+
+```
+# Filter the tests to run by name, including all that contain 'config':
+postgrest-test-io -k config
+
+# Run tests in parallel using xdist, specifying the number of processes:
+postgrest-test-io -n auto
```
wolfgangwalther, line 166: can we make auto the default and then override it to `-n 1` for CI (if needed)?

nix/README.md
```diff
@@ -150,7 +150,21 @@ $ nix-shell --run postgrest-test-spec-all
...
-$ nix-shell --run postgrest-test-spec-postgresql-9.5
+$ nix-shell --run postgrest-test-spec-postgresql-13
+
+```
+
+The io-test that test PostgREST as a black box with inputs and outputs can be
+run with `postgrest-test-io`. The test runner under the hood is
+[pytest](https://docs.pytest.org/) and you can pass it the usual options:
+
+```
```
wolfgangwalther, line 161:
```suggestion
```bash
```

nix/README.md
wolfgangwalther: When I run `postgrest-test-io -n auto -n 1` it runs with `-n 1`, so this should work.

wolfgangwalther:
`postgrest-test-io -n auto` runs in ~14s compared to ~23s for me, cool!

But: When I run it with auto every now and then some tests fail with `PostgrestTimedOut` errors. Sometimes it works, sometimes it's 2-3 tests that are failing.

test/memory-tests.sh
```diff
@@ -20,6 +20,7 @@ ok(){ result 'ok' "- $1"; }
 ko(){ result 'not ok' "- $1"; failedTests=$(( $failedTests + 1 )); }
 
 pgrPort=49421
+export POSTGREST_TEST_PORT="$pgrPort"
```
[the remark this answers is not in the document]
monacoremo: Ah, yes - this was a leftover from before the rebase. Fixed and moved pgrPort up, so that we can also use it for the env variable without duplication

test/io-tests/test_io.py
```diff
@@ -87,16 +90,21 @@ def dumpconfig(configpath=None, moreenv=None, stdin=None):
 
 
 @contextlib.contextmanager
-def run(configpath, stdin=None, moreenv=None, socket=None):
+def run(configpath=None, stdin=None, env=None, serversocket=None):
     "Run PostgREST and yield an endpoint that is ready for connections."
-
-    if socket:
-        baseurl = "http+unix://" + urllib.parse.quote_plus(str(socket))
+    if serversocket:
```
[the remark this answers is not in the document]
monacoremo: Switched to run via socket by default, to run over ip you need to provide an explicit port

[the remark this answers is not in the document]
wolfgangwalther: Yes, that's best!

monacoremo:
> But: When I run it with auto every now and then some tests fail with PostgrestTimedOut errors. Sometimes it works, sometimes it's 2-3 tests that are failing.

Yes, also saw that with very large -n , e.g. almost always with -n 50 to run all tests at once. Increasing the read_timeout and the retries fixes the issue. But I'm hesitant to increase them by default, as this would massively increase the time taken to complete the tests if anything else is wrong. Would suggest to leave it non-parallelized by default and leave -n + tweaks to the timeouts to our powerusers.

nix/README.md
monacoremo: Please see the comment below

wolfgangwalther:
Yes, I'm fine with that - also because the output of running the tests in parallel isn't really that nice anymore - lots of noise from those workers.

PR #1705, opened 2020-12-23
nix: add "cd $rootdir" to all shell scripts via checked-shell-script

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/pull/1694#discussion_r545403933

@monacoremo: I felt like a `postgrest-rootdir` was not necessary and just added it to the checked-shell-script wrapper. What do you think?

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -37,7 +37,7 @@ GHC version and all the build dependencies.
 We recommend that you use the PostgREST binary cache on
 [cachix](https://cachix.org/):
 
-```
+```bash
 # Install cachix:
 nix-env -iA cachix -f https://cachix.org/api/v1/install
 
@@ -70,7 +70,7 @@ The PostgREST utilities available in `nix-shell` all have names that begin with
 `postgrest-`, so you can use tab completion (typing `postgrest-` and pressing
 `<tab>`) in `nix-shell` to see all that are available:
 
-```
+```bash
 # Note: The utilities listed here might not be up to date.
 [nix-shell]$ postgrest-<tab>
 postgrest-lint                      postgrest-test-spec-postgresql-11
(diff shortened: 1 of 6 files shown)


monacoremo:
That works great! A few thoughts:

* will we have any scripts that we don't want to always run in the root dir of the project? I can't think of any right now, so it seems to be a sensible default
* this behavior might be surprising, so it's worth to explicitly document it somewhere
* we could use this opportunity (or later) to make this a bit more robust and user friendly, by checking for the existence of our `postgrest.cabal` in the root dir and failing with something like "not in a postgrest project, aborting" if not
* I was thinking of changing the 'name' parameter to a set, with mandatory name and docs attributes, to ensure that all our scripts have docs. This would also be a good place to add an optional inRootDir attribute, that would enable this behavior

wolfgangwalther:
> * will we have any scripts that we don't want to always run in the root dir of the project? I can't think of any right now, so it seems to be a sensible default

I can't think of any right now - but even if we do, we can still introduce the `inRootDir` attribute later.

> * this behavior might be surprising, so it's worth to explicitly document it somewhere

I added a paragraph to `nix/Readme.md`.

> * we could use this opportunity (or later) to make this a bit more robust and user friendly, by checking for the existence of our `postgrest.cabal` in the root dir and failing with something like "not in a postgrest project, aborting" if not

I went the other direction: I replaced the `git` command with a `rootdir` variable set in `default.nix`. This will always point to the same folder that `default.nix` / `shell.nix` are in. Now, once in nix-shell you can run those commands from **anywhere**, including paths completely outside the PostgREST repo.

> * I was thinking of changing the 'name' parameter to a set, with mandatory name and docs attributes, to ensure that all our scripts have docs. This would also be a good place to add an optional inRootDir attribute, that would enable this behavior

I'll leave that for you for later... :o).

monacoremo:
> I went the other direction: I replaced the `git` command with a `rootdir` variable set in `default.nix`. This will always point to the same folder that `default.nix` / `shell.nix` are in. Now, once in nix-shell you can run those commands from **anywhere**, including paths completely outside the PostgREST repo.

Ah, I'm not a fan! This means builds will not be reproducible for different users (meaning rebuilds everywhere, albeit small) and I will be locked into the location where I once had the repo when I use nix-env. I will also not be able to easily switch between checkouts.

wolfgangwalther:
> Ah, I'm not a fan! This means builds will not be reproducible for different users (meaning rebuilds everywhere, albeit small) and I will be locked into the location where I once had the repo when I use nix-env. I will also not be able to easily switch between checkouts.

Yeah... I had a feeling that this wasn't that good of an idea. Changed it back to using `git` - now it exits with error, if the command is run outside the git repo.

monacoremo:
LGTM! :)

PR #1706, opened 2020-12-24
Add docs to all checkShellScripts in Nix

monacoremo:
This adds docs to every `checkedShellScript`-based script that we have in our Nix setup.

Need to figure out how to best use the `docs` attribute in `checkShellScript`, but this is already useful as is I think.

### DIFF
--- a/nix/overlays/checked-shell-script/checked-shell-script.nix
+++ b/nix/overlays/checked-shell-script/checked-shell-script.nix
@@ -2,13 +2,15 @@
 # directly, or use the .bin attribute to get the script in a bin/ directory,
 # to be used in a path for example.
 { git
+, lib
 , runCommand
 , runtimeShell
 , shellcheck
 , stdenv
 , writeTextFile
 }:
-name: text:
+{ name, docs, inRootDir ? false }: text:
+# TODO: do something sensible with docs, e.g. provide automated --help
 let
   bin =
     writeTextFile {
@@ -20,7 +22,8 @@ let
         ''
           #!${runtimeShell}
           set -euo pipefail
-
+        ''
+        + lib.optionalString inRootDir ''
           cd "$(${git}/bin/git rev-parse --show-toplevel)"
 
           if test ! -f postgrest.cabal; then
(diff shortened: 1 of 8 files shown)


monacoremo:
This also makes it more explicit where we use our `inRootDir` feature, hopefully making it a bit less surprising.

nix/tests.nix
```diff
@@ -124,7 +150,11 @@ let
   dumpSchema =
     name: postgresql:
     checkedShellScript
-      name
+      {
+        inherit name;
+        docs = "Dump the DbStructure schema as loaded as yaml file..";
```
wolfgangwalther, line 155:
```suggestion
        docs = "Dump the loaded schema's DbStructure as a yaml file.";
```

nix/overlays/checked-shell-script/checked-shell-script.nix
```diff
@@ -2,13 +2,15 @@
...
+, lib
 , runCommand
 , runtimeShell
 , shellcheck
 , stdenv
 , writeTextFile
 }:
-name: text:
+{ name, docs, inRootDir ? false }: text:
+# TODO: do something sensible with docs, e.g. provide automated --help
```
wolfgangwalther, line 13: If we add `--help`, we'd need a way to pass the `--help` to the underlying command, too. Something like `-- --help` or so. Many of the commands pass arguments through and I sometimes use it.

monacoremo: Yes, that was my worry - if we print the docs when `--help` or `-h` are on the command line, how do we make sure that we don't prevent the commands we use to show their help texts? `--` is the right solution probably. Will have to experiment with it.

wolfgangwalther:
The windows build on AppVeyor is failing everywhere (although a bit randomly) right now - this seems to be a caching issue. The error message doesn't make sense - it reports something, that I temporarily had in the "config from env" PR, but is long gone.

Going to merge this, the `--help` can be added later, if we figure it out.

Issue #699, opened 2016-08-19
Cancel query if user cancels request


begriffs:
@NikolayS described a nightmarish situation where he accidentally sent a DELETE request to a route with no filters, causing PostgREST to start deleting all rows of the table. After a few seconds he realized it and killed the http request, but the query kept on running.

I think we could use [setOnClose](https://hackage.haskell.org/package/warp-3.2.8/docs/Network-Wai-Handler-Warp.html#v:setOnClose) to detect this situation and roll back the current transaction. Not sure about the details of sharing the connection between what might be multiple threads.

steve-chavez:
While reading about the pg protocol, I've noticed that [Canceling Requests in Progress](https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.5.7.9) is possible.

As a side note, protecting from an PATCH/DELETE with no filters has been such a common need(and for all endpoints) for such a long time that I'm wondering why shouldn't we offer a config option that does exactly that.

PUT already has a [safeguard](https://github.com/PostgREST/postgrest/blob/eed0d3a66c48f7fec8d3b31708aa8be6f7bbc0be/src/PostgREST/App.hs#L248-L256), we could do the same for PATCH/DELETE.

The configs can be like:

- `db-protect-update`
- `db-protect-delete`

They would enforce the presence of a filter for PATCH/DELETE to succeed.

The options would be false by default. That's in case users want to handle that at the db level with [pg-safeupdate](https://github.com/eradman/pg-safeupdate).

bmillwood:
Just to mention, I'd definitely prefer the version where the underlying query gets cancelled. I occasionally accidentally ask my postgrest server for a query that will return some astronomical number of rows and take several minutes to process. When it doesn't load immediately, I notice my mistake and try to close the window, but by then it's too late.

steve-chavez:
The implementation of cancel is a bit complicated as mentioned [here](https://stackoverflow.com/a/18248218/4692662)(also might be unreliable according to [this](https://deno.land/x/postgresjs@v3.2.4#canceling-queries-in-progress)).

Right now we have the [max rows hard limit](https://postgrest.org/en/stable/configuration.html#db-max-rows), but maybe we can add a soft limit config with a default(like 1000) since it's almost never good to have such bulky responses at the API level. @bmillwood WDYT?

bmillwood:
I think for now don't do that on my account :) I can always fix my queries so they have inbuilt limits instead.

[...]

Issue #1661, opened 2020-11-21
Expose raw body in request


nathancahill:
It would be useful to expose the raw body of a request the same way that headers are exposed:

```python
current_setting('request.header.x', true);
```

Maybe:

```python
current_setting('request.body', true);
```

Useful for verifying signed payloads.

alexStrickner00:
I was also just looking for a possibility to get the request body in a pre-request function.

wolfgangwalther:
> It would be useful to expose the raw body of a request the same way that headers are exposed:

I'm not sure this is a good idea. GUCs are already a bit abused by us and are not really meant to be used that way. Putting potentially huge request bodies with possibly binary data in there, doesn't sound like a good idea. Plus there would need to be an opt-out.

> I was also just looking for a possibility to get the request body in a pre-request function.

But this is something that I could see happening.

What about:
- if the pre_request function is defined without arguments, don't pass the body
- it the pre_request function has a json argument, pass the json body (in some cases this is the raw body, in other cases this is transformed to json from some of the other types, like csv, form-data or the query string for RPCs)
- if the pre_request function has a text (or better bytea?) argument, pass the raw request body without any changes

This would have opt-in included.

Would something like this solve your use-case, @nathancahill, as well?

I can easily see this extended to allow the pre-request function to `RETURN json` as well. This would even allow some sort of transformation on the input before it is passed to the query. In this case the pre_request function would have to be called in the main query, instead of in the `SET LOCAL` part.

nathancahill:
That would be fantastic, much better than adding it to GUCs.

alexStrickner00:
>What about:
 >   if the pre_request function is defined without arguments, don't pass the body
 >   it the pre_request function has a json argument, pass the json body (in some cases this is the raw body, in other cases this is transformed to json from some of the other types, like csv, form-data or the query string for RPCs)
  >  if the pre_request function has a text (or better bytea?) argument, pass the raw request body without any changes

That would be an optimal solution for my usecase. 👍

steve-chavez:
@wolfgangwalther The `pre-request` can also be used for things like changing the `role` and considering https://github.com/PostgREST/postgrest/pull/1600#issuecomment-741894511 we'd need the `pre-request` to be called before the query right?

So maybe we should use a different function for this case, maybe a `pre-body`.

wolfgangwalther:
[...]

PR #1710, opened 2020-12-28
feat: add request.param guc

steve-chavez:
As discussed on https://github.com/PostgREST/postgrest/issues/915#issuecomment-744736038. This adds the `request.param` GUC.

This is the draft implementation. Not efficient because it leads to a lot of new prepared statements(no reuse).
Each query param `?select=id,name&id=eq.1&name=eq.project&order=id` will get its own parameter:

```sql
select 
  set_config('request.param.id', $1, true),
  set_config('request.param.name', $2, true),
  set_config('request.param.select', $3, true),
  set_config('request.param.order', $4, true);
```

### Ideas

- Convert all the query parameters to a single json, like mentioned in https://github.com/PostgREST/postgrest/pull/1600#issuecomment-738232201. Would lead to reusing prepared statements.
- How about having a single GUC for the filters only?
  - It can be like `request.param.filters`. This would be a JSON object: `{"id": "eq.1", "name": "eq.project"}`. There wouldn't be a `request.param.id` or `request.param.name`.
  - With this we could disable update/delete with no filters globally by using `pre-request`. It'd be easier to validate that filters are present by having them on a single json.

  ```sql
  create or replace function update_delete_restrictions() returns void as $$
  declare
    req_filters text = nullif(current_setting('request.param.filters', true), '{}');
    req_method text = current_setting('request.method', true);
  begin
  if req_method similar to 'PATCH|DELETE' and req_filters is null then
    RAISE EXCEPTION 'UPDATE or DELETE is not allowed without filters'
    USING HINT = 'Add filters to the request';
  end if;
  end $$ language plpgsql;
  ```
  (this would be more flexible than doing it as config option as proposed on https://github.com/PostgREST/postgrest/issues/699#issuecomment-716266488)

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -129,6 +129,7 @@ data ApiRequest = ApiRequest {
   , iMethod               :: ByteString                       -- ^ Raw request method
   , iProfile              :: Maybe Schema                     -- ^ The request profile for enabling use of multiple schemas. Follows the spec in hhttps://www.w3.org/TR/dx-prof-conneg/ttps://www.w3.org/TR/dx-prof-conneg/.
   , iSchema               :: Schema                           -- ^ The request schema. Can vary depending on iProfile.
+  , iParams               :: [(Text, Text)]                   -- ^ All request params
   }
 
 -- | Examines HTTP request and translates it into user intent.
@@ -177,6 +178,7 @@ userApiRequest confSchemas rootSpec dbStructure req reqBody
       , iMethod = method
       , iProfile = profile
       , iSchema = schema
+      , iParams = second (toS . fromMaybe mempty) <$> qParams
       }
  where
   -- queryString with '+' converted to ' '(space)
--- a/src/PostgREST/Middleware.hs
+++ b/src/PostgREST/Middleware.hs
@@ -49,13 +49,14 @@ runPgLocals :: AppConfig   -> M.HashMap Text JSON.Value ->
                ApiRequest  -> H.Transaction Response
 runPgLocals conf claims app req = do
   H.statement mempty $ H.dynamicallyParameterized
-    ("select " <> intercalateSnippet ", " (searchPathSql : roleSql ++ claimsSql ++ [methodSql, pathSql] ++ headersSql ++ cookiesSql ++ appSettingsSql))
+    ("select " <> intercalateSnippet ", " (searchPathSql : roleSql ++ claimsSql ++ [methodSql, pathSql] ++ headersSql ++ cookiesSql ++ paramsSql ++ appSettingsSql))
     HD.noResult (configDbPreparedStatements conf)
   traverse_ H.sql preReqSql
   app req
   where
     methodSql = setConfigLocal mempty ("request.method", toS $ iMethod req)
     pathSql = setConfigLocal mempty ("request.path", toS $ iPath req)
+    paramsSql = setConfigLocal "request.param." <$> iParams req
     headersSql = setConfigLocal "request.header." <$> iHeaders req
     cookiesSql = setConfigLocal "request.cookie." <$> iCookies req
     claimsWithRole =


wolfgangwalther:
Hm. I'm not entirely convinced that I would want to have all the request params set as GUCs. Every GUC has a performance penalty when setting it: I remember from the SQL tests with did with the `SET LOCAL` stuff - the number of gucs had a linear impact on the query time. There is no way to opt out of this.

> It'd be easier to validate that filters are present by having them on a single json.

I agree with that.

---

Taking a step back for a moment I wonder whether we can design something that would allow me to opt-in for all kind of `request.` GUCs, but have nothing as the default - for the best-performing case.

What if we did:
- remove all `set_config` except `role` and `search_path`.
- pass those settings in to hook functions we have (currently only `db-pre-request`, but might become more, see https://github.com/PostgREST/postgrest/issues/1661#issuecomment-743388070).
- make it optional by using named arguments on those hooks, similar to how fixtures work with pytest.

Example:
```sql
CREATE FUNCTION my_pre_request(method TEXT, filters JSON) RETURNS void
LANGUAGE plpgsql AS $$
  ... similar to the function in the opening post ...
$$; 
```

Here, we would parse the function definition together with the schema cache and would then know which arguments we need to pass in.
We could easily add an example pre-request function that uses `set_config` to restore current behaviour to `postgrest-contrib`.

This would have the advantage that we could choose which variables to pass / set and could leave a big chunk out in most cases - improving performance.

We would still only need one round-trip to the database, it should not perform worse.

---

Thinking more about the naming of arguments, we could support something like a json path in those, to be able to select only a sub-key:

```sql
CREATE FUNCTION my_pre_request("headers.my_custom_header" TEXT, filters JSON)
```
This would provide all filters, but only one of the headers.

steve-chavez:
> pass those settings in to hook functions we have (currently only db-pre-request)
> make it optional by using named arguments on those hooks, similar to how fixtures work with pytest.

Defining the gucs in pre-request seems interesting.

A drawback is that if you only want to use restricted views or tables(https://github.com/PostgREST/postgrest/issues/915#issuecomment-744736038), you'd have to define the gucs on pre-request as well, which makes the feature more complicated to use. Ideally one would only touch pre-request for the global cases.

An advantage is that it could allow more GUCs per endpoint and it would not be a global config like we could do on the config file.

Not opposed to the idea, need to have more thought about it.

---

Other stuff that could be interesting for this feature: 

- `request.param.filters.indexed`(json): with this we could reject all requests that don't use at least one indexed filter. We'd have to get INDEXes into our schema cache for this.
- `request.param.distinct` and `request.param.group_by`: if we expose `distinct` and `group by` as discussed on #915, these two would allow to globally disallow them(would solve https://github.com/PostgREST/postgrest/issues/915#issuecomment-744759089) or enable them per endpoint.

steve-chavez:
### New idea for filter restrictions

How about this.

If we (ab)use our in-db config for defining restrictions, we could make them static checks that **don't have to hit the database**.

Example:

```sql
CREATE TABLE clients(
  id int
, name text
);

-- assuming our postgrest-contrib in place
SELECT pgrst.restrict_filter('clients', 'id');
```

Then a request that doesn't include a filter on `id` will get rejected as:

```http
GET /clients

HTTP/1.1 403 Forbidden
{"hint":null,"details":null, "code": null, "message":"id filter must be present"}
```

To make that work, the `pgrst.restrict_filter(..)` would be translated to:

```sql
ALTER ROLE pgrst_authenticator SET parameters_restrictions = $$
{
  "clients": "id"
}
$$;

-- or to a config file as
parameters-restrictions = '{"clients": "id"}'
```

### Advantages

- We don't have to add GUCs to every request.
- As a I mentioned above, we can reject the request at the postgrest level, it doesn't have to hit the db. It would be similar to how we reject non-existent embeds.
- We could use this mechanism not only for tables/views but also for set returning functions.
```sql
SELECT pgrst.restrict_filter('rpc/get_clients', 'id');
```
- Restricting PATCH/DELETE without filters could also be done in a similar manner.
```sql
SELECT pgrst.restrict_method('clients', '{PATCH,DELETE}', '{id,name}');
-- only allow patch/delete if id or name filters are present.
```
- Though the functions interface is not yet defined, this could also aid in allowing them in this way:

```sql
SELECT pgrst.allow_function('clients', 'avg(salary)');
```
 
(Just an idea, the functions form could use more work)

---

@wolfgangwalther What do you think? Do you see any drawbacks with this approach?

wolfgangwalther:
> If we (ab)use our in-db config for defining restrictions, we could make them static checks that **don't have to hit the database**.
> ### Advantages
> * We don't have to add GUCs to every request.
> * As a I mentioned above, we can reject the request at the postgrest level, it doesn't have to hit the db. It would be similar to how we reject non-existent embeds.
> * We could use this mechanism not only for tables/views but also for set returning functions.
> * Restricting PATCH/DELETE without filters could also be done in a similar manner.
> * Allow/deny operators (someone asked for this before on gitter, only choice was proxy)
> * Enable group by/distinct globally or per route
>
> @wolfgangwalther What do you think? Do you see any drawbacks with this approach?

Hm. My first reaction was: Why not do that at the proxy-level? This seems out of scope for PostgREST.

But giving it further thought... I actually like it. When discussing all the content negotiation stuff with `Accept` headers etc., I briefly had the idea of improving our current "router" situation. Right now we basically have a very basic router hardcoded: It just takes the route, maps it to a table or RPC name and then does a little bit of magic with the `Accept` header.

We could implement this in a much more generic and flexible way. And then we could load additional config per route the kind of checks you mention. I'm not sure about using all kinds of GUCs / config options for those kind of settings, though. Maybe we should just have a `db-router-config` setting that takes a table name. This table would have to follow a predefined schema and would hold all kind of those settings you mentioned above.

steve-chavez:
> Why not do that at the proxy-level? This seems out of scope for PostgREST.

Yeah, the line that separated proxy and PostgREST has blurred out with time. Even the previous GUC `request.param` proposal or the whole discussion at https://github.com/PostgREST/postgrest/issues/915#issuecomment-744736038 would be meaningless if we decide to do in the proxy(which we can). But that would keep us from evolving and adding features that are valuable.

Also, we're already a proxy for pg as well, after all we reject some requests(invalid http method, invalid jwt, etc) without touching pg.

Additionally, doing REST restrictions with our functions seems superior to doing it inside RLS or a security-barrier-like condition on views. They're a PostgREST concern after all, they should be stored in our config.

> I briefly had the idea of improving our current "router" situation
> Maybe we should just have a db-router-config setting that takes a table name.

Sounds interesting! But I think we should keep the design that `authenticator` doesn't need a privilege over a table or usage over a schema. So perhaps we should keep doing it in our config(as a JSON field), or maybe later on a table inside our extension.

wolfgangwalther:
> > I briefly had the idea of improving our current "router" situation
> > Maybe we should just have a db-router-config setting that takes a table name.
> 
> Sounds interesting! But I think we should keep the design that `authenticator` doesn't need a privilege over a table or usage over a schema. So perhaps we should keep doing it in our config(as a JSON field), or maybe later on a table inside our extension.

Hm. The table approach is kind of a hack to achieve the same thing as schema variables would do. For schema variables, we'd need to have privileges, too. Yes, I was thinking about putting it in our extension. This idea was mainly targeted at cloud-based users, who can't use GUC-based configuration. Maybe we can even extend this idea:

How about making `db-config` take either a boolean or a string? The meaning of boolean would still be the same. If it's set to a string, this is the schema qualified name of a table, that the authenticator user needs access to. This table can be a simple key/value store of all those config options that we are reading from GUCs right now. This would basically replace the GUC loading with table-loading. This would allow all cloud-based users to still maintain their config in the database.

---

In any case, I agree with keeping the filter settings you mentioned above in a single `JSON` map. This would allow the following generalization of our "routing".

We could create some types roughly along the lines of:

```haskell
-- maps a list of path segments to a route configuration
type Router = M.Map [Text] Route

data Route = Route {
  rtTarget :: Target,
  -- ... other settings
  rtAllowedMethods :: [Method],
  rtAllowedOperators :: [Operator],
  rtAllowAggregation :: Boolean,
  rtAllowedFunctions :: [Text],
  rtRequiredFilters :: [Text]
}
```

(Those types need to be adjusted to properly allow function overloading for RPCs, but the general idea would be the same.)

When parsing the request, we would match this to one of the routes. This would imply, that we can't respond to any unknown routes anymore. I remember we discussed this elsewhere already, though.

The default values for each route are generated from the schema cache. And then we add the json config option mentioned above (e.g. `router-overlay`) on top of the whole `Router` structure. This would allow us to override some of those values selectively. With some helper functions as mentioned above, this could be easily managed. In the future this could also allow much more customization, e.g. moving target tables/views/rpcs to completely different paths, etc.

wolfgangwalther:
Maybe `wai-routing` can be of help here, too. I didn't fully read it, yet, but it sounds like those preconditions could be checked with it: https://hackage.haskell.org/package/wai-routing-0.3/docs/Network-Wai-Routing-Tutorial.html.

steve-chavez:
> Every GUC has a performance penalty when setting it: I remember from the SQL tests with did with the SET LOCAL stuff - the number of gucs had a linear impact on the query time

The above is still true. However now that we use a [single json for our GUCs](https://postgrest.org/en/stable/api.html#accessing-request-headers-cookies-and-jwt-claims), `request.params` would have a lesser impact.

steve-chavez:
The [above ideas](https://github.com/PostgREST/postgrest/pull/1710#issuecomment-808788905) could also be done with a SECURITY LABEL extension, like discussed on https://github.com/PostgREST/postgrest/issues/2442#issuecomment-1228812319

rdlrt:
The option to be able to access limit/offset - even if not other columns (controlled via config) - also helps us build RPCs that can artificially use custom pagination techniques (for my use case, currently modifying RPC to view works , combed with haproxy URL redirection for consistency - but would be nice to have a more consistent option as we predominantly rely on /rpc => /api map - which works for 80% of endpoints - except cases when RPC are not ideal - for being able to leverage keyset pagination)

PR #1711, opened 2020-12-29
Fix "unused-imports" warning for windows build

wolfgangwalther:
When merging #1691 I introduced the failing appveyor windows build. It didn't fail when merging, because the windows build was not finished yet (cancelled after timeout). Now it has been failing ever since. I first thought this was a caching issue, but it turns out it is for real - one of the imports is not used for the Windows build.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -32,11 +32,7 @@ import System.IO                (BufferMode (..), hSetBuffering)
 import Text.Printf              (hPrintf)
 
 import PostgREST.App         (postgrest)
-import PostgREST.Config      (AppConfig (..), CLI (..), Command (..),
-                              Environment, configDbPoolTimeout',
-                              dumpAppConfig, prettyVersion,
-                              readCLIShowHelp, readEnvironment,
-                              readValidateConfig)
+import PostgREST.Config
 import PostgREST.DbStructure (getDbStructure, getPgVersion)
 import PostgREST.Error       (PgError (PgError), checkIsFatal,
                               errorPayload)


main/Main.hs
```diff
@@ -32,11 +32,7 @@ import System.IO                (BufferMode (..), hSetBuffering)
 import Text.Printf              (hPrintf)
 
 import PostgREST.App         (postgrest)
-import PostgREST.Config      (AppConfig (..), CLI (..), Command (..),
-                              Environment, configDbPoolTimeout',
-                              dumpAppConfig, prettyVersion,
-                              readCLIShowHelp, readEnvironment,
-                              readValidateConfig)
+import PostgREST.Config
```
monacoremo, line 35: We should avoid unqualified wildcard imports where possible, as they make the code more difficult to follow. 

But I think we can leave it like this for now, I was going to make a suggestion anyway for refactoring our imports and untangling some of the code (esp. App.hs).

wolfgangwalther: I agree. I wanted to split up the import, to import `Environment` from `PostgREST.Config` only in the `#ifndef ...` block below - but that throws off the linter, because hlint wants to merge those two import statements then...

Qualifying it would make it easier to understand - but that's something we should tackle on a bigger scale, as you say.

steve-chavez: I think the problem here is just the `Environment` import.

Couldn't we wrap that import with `#ifndef mingw32_HOST_OS` and leave the rest explicit?

wolfgangwalther: > Couldn't we wrap that import with `#ifndef mingw32_HOST_OS` and leave the rest explicit?

Cross-posted here, but see above: hlint complains about that. The unqualified import was the only way out I found here that fixes the build error for now, but does not involve more refactoring across the whole file.

steve-chavez: Actually I remembered it wasn't that simple because of `hlint`.

That's why I had to make the `UnixSocket.hs` file. So I agree with leaving it like this for now.

steve-chavez: > Cross-posted here, but see above: hlint complains about that.

Ah, you beat me to it(non real-time comments played me). Yes, the unqualified import is good for now.

PR #1713, opened 2020-12-29
Add test coverage overlay for code that is not tested on purpose or not reachable

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/pull/1694#discussion_r549840741, I'm splitting up the overlay into another PR. We need to increase our overall coverage by a fair bit, before merging this.


wolfgangwalther:
For future reference. Code coverage can show some lines as unused, even though they are used. We had the following cases:
- `deriving ...`: https://gitlab.haskell.org/ghc/ghc/-/issues/17155:

  > So, unless you explicitly call every methods of Ord, deriving Ord is marked as never executed.

- Using `RecordWildCards` will not mark record fields as used: https://gitlab.haskell.org/ghc/ghc/-/issues/17834

steve-chavez:
Just a note for posteriority: make the InternalError(https://github.com/PostgREST/postgrest/pull/2270#issuecomment-1115463986) not reduce coverage.

PR #1714, opened 2020-12-30
Fix io tests not collecting coverage data when using run()

wolfgangwalther:
Before, postgrest was killed after each test and no coverage data was saved.
Now the process is terminated gracefully via SIGTERM first to allow writing the .tix file.

To avoid concurrency issues when executing the io tests in parallel, each io test case yields a separate tix file now.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -184,8 +184,10 @@ let
         # temporary directory to collect data in
         tmpdir="$(mktemp -d)"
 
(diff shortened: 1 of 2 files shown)


monacoremo (approved):

wolfgangwalther:
Just a note: This does still not fix the random differences in coverage between runs. Currently still investigating those.

PR #1717, opened 2020-12-30
Add basic io tests and remove unused code

wolfgangwalther:
Part of #1719.

Best to review commit-by-commit:
1. commit adds io tests for basic cli commands and many config related errors. This will also allow to refactor some of the tests that are currently based on config files (boolean values).
2. commit is taken from #1713. This removes unused code that I stumbled upon when first going through the coverage overlay draft. Most of that is reviewed automatically by the compiler (all `deriving (...)`). In many places I changed `[]` to `mempty` - I did this whenever I added an overlay for it in #1713, because that means the value is never ever used. `mempty` seemed to fit better in that case. The remaining few changes are removed code that is never used - all in codepaths that are well tested. The tests are still passing, so confidence should be high that those are really not needed at all.
3. Is just a small refactor that increases code coverage on paper, because Pattern Matching is more flexible than `elem`.

The io tests probably need the most review, all other stuff is covered mostly by the compiler and existing tests.


wolfgangwalther:
codecov/patch is failing by design. This job measures the coverage across those lines that I touched. Since I mostly touched lines that would appear in the coverage overlay, those are by design untested right now.

wolfgangwalther:
@steve-chavez @monacoremo Should I split this PR up further for easier review or is it good like that?

src/PostgREST/Auth.hs
```diff
@@ -117,7 +115,5 @@ parseSecret str =
 jwkFromSecret :: ByteString -> JWK
 jwkFromSecret key =
   fromKeyMaterial km
-    & jwkUse ?~ Sig
-    & jwkAlg ?~ JWSAlg HS256
```
steve-chavez, line 121: Hmm.. those are default?

wolfgangwalther: I'm not entirely sure. The values were marked as uncovered and all tests still pass without those lines. If those settings are the default, then the values would have still be used.

There must be something else. My assumption is, that those settings don't have any effect at all and are just neither used nor needed.

steve-chavez: I'm a bit scared of the change because the algorithm could default to `None`:

https://hackage.haskell.org/package/jose-jwt-0.8.0/docs/Jose-Jwa.html#t:JwsAlg

And it could be that the tests are just accepting any JWT.

wolfgangwalther: Well, if the value is not used at all, then it doesn't matter what we set it to.

My understanding is that a JWT does have the algorithm it was encoded with included. This is the header of a JWT:
```
{"typ":"JWT",
 "alg":"HS256"}
```

My best bet is, that those settings are only used when actually signing a JWT with this JWK. But we only ever use this JWK for verifying the signature of an existing JWT.

I further assume that the only reason we turn the secret into a JWK here is to be able to use real JWKs instead of text secrets easily with the same code path for the verification.

So basically, this is not a real JWK, but just a container for our secret to have the right type.

wolfgangwalther: Of course, it could also be, that we're just missing a proper test here and that's why it's uncovered. But I think that's unlikely, given that we do use properly signed JWTs in our tests and there is no config option that affects how we use the text secret, right? I.e. which encoding algorithm etc. - that's all part of application that actually signs the JWT.

wolfgangwalther: So I had a look at the code again: We are using the JWK that is created here in `attemptJwtClaims`. Here, we're using `defaultJWTValidationSettings` of which `defaultValidationSettings` is the main part:

https://hackage.haskell.org/package/jose-0.8.4/docs/Crypto-JOSE-JWS.html#v:defaultValidationSettings

There it says:
> The default validation settings.
> - All algorithms except "none" are acceptable.
> - All signatures must be valid (and there must be at least one signature.)

So the algorithm is in fact taken from the JWT, but "none" is not accepted.

wolfgangwalther: I also tracked down the signature verifying function: https://hackage.haskell.org/package/jose-0.8.4/docs/src/Crypto.JOSE.JWS.html#verifySig

It only reads `jwkMaterial` from the key and not `jwkAlg` or `jwkUse`. **If** we really want to limit the allowed algorithms for verifying JWTs, we'd need to change this line:

https://github.com/PostgREST/postgrest/blob/ab6f90aa78b5ff5b4643f9b3ac97ae4ff6b0b589/src/PostgREST/Auth.hs#L65

and change the allowed algorithms in those settings.

wolfgangwalther:
Just a rebase for the merge conflict.

src/PostgREST/Auth.hs
steve-chavez: > All algorithms except "none" are acceptable.
> It only reads jwkMaterial from the key and not jwkAlg or jwkUse

Great! Thanks for digging into this @wolfgangwalther.

So then I'm no longer worried about this change. It's safe to remove these lines.

> If we really want to limit the allowed algorithms for verifying JWTs, we'd need to change this line:

It looks fine to leave as it is for now. Good to know we can limit the algorithms though.

steve-chavez (approved):
LGTM! It's great how coverage pushes us to delete unused code.

test/io-tests/fixtures.yaml
```diff
@@ -1,3 +1,144 @@
...
+    expect: error
+    use_defaultenv: true
+    env:
+      PGRST_SERVER_UNIX_SOCKET_MODE: '778'
+# TODO: Bug needs to be fixed
+#  - name: invalid jwt-aud
+#    expect: error
+#    use_defaultenv: true
+#    env:
+#      PGRST_JWT_AUD: 'htp:/@@localhorst.invalid'
```
taimoorzaeem, line 63: @wolfgangwalther Any string value is allowed in `jwt-aud`, I think this is not a bug.

wolfgangwalther: No, not *any* string value. Allowed values are of type `StringOrURI`, which is defined as this by the RFC:

> A JSON string value, with the additional requirement that while arbitrary string values MAY be used, any value containing a ":" character MUST be a URI [[RFC3986](https://www.rfc-editor.org/rfc/rfc3986)].  StringOrURI values are compared as case-sensitive strings with no transformations or canonicalizations applied.

Thus, the value in this snippet must throw an error, but it didn't back then. Does it now?

taimoorzaeem: > Does it now?

Nope. I guess we can raise issue for this?

wolfgangwalther: Sure

Issue #1719, opened 2020-12-31
Increase test coverage


wolfgangwalther:
To be able to properly use coverage overlays (#1713), we need to increase coverage close to 100%. Many of the files are close to that already.

This issue is to track progress across PRs.

TODO:

- [ ] Main
  - [ ] Add tests for schema reload on notify
  - [ ] Add tests dealing with database connection problems 
- [ ] UnixSocket
  - [ ] Add test to throw when unix socket already exists
- [ ] PostgREST.ApiRequest
  - [ ] Add test with empty query param (`?key=`) (line 171)
  - [ ] Add test with `+` in query string converted to space (line 183)
  - [ ] Debug `"application/x-www-form-urlencoded"` (line 215)
  - [ ] Add test for POST with unsupported Content-Type body (line 240)
  - [ ] Add test for unsupported request method (line 259)
- [ ] PostgREST.App
  - [ ] Make any RPC `STABLE` (line 112) (should we test for read only transaction somehow?)
  - [ ] Add tests that throw with `POST`, `PATCH`, `DELETE` (lines 160, 195, 254)
  - [ ] Check whether count queries can actually throw (lines 138, 170, 204, 237, 264)
  - [ ] Add test to return count on `PATCH` (line 212)
  - [ ] Add `OPTIONS` request to missing endpoint (line 285)
  - [ ] Check whether `pdName` can be covered in line 292
  - [ ] Find out why `tablePKCols dbStructure s tn` is not covered in line 329
- [ ] PostgREST.Auth
  - [ ] Add roleclaimkey tests that try to access objects with arrays presents or vice-versa (line 89)
  - [ ] Check whether constructing a JWT with something different from an object as top-level json value can be constructed (line 82)
  - [ ] Add test with empty `jwt-aud` (line 64)
- [x] PostgREST.Config (#1717)
- [ ] PostgREST.DbRequestBuilder
  - [ ] _Understand whats happening here..._ 
- [ ] PostgREST.DbStructure
  - [ ] Refactor `addHasVariadic` to in-SQL (line 145)
  - [ ] Refactor `parseArg` to in-SQL to solve TODO and coverage (line 161)
  - [ ] Add schema description to fixtures (line 247) 
  - [ ] Refactor `combinations` to allow better Pattern Matching in line 404
  - [ ] Refactor `addViewM2ORels` to remove Pattern Matching in line 342
- [ ] PostgREST.Error
- [ ] PostgREST.Middleware
[...]

PR #1720, opened 2020-12-31
Refactor dbStructure.hs

wolfgangwalther:
Related to #1719.

### DIFF
--- a/src/PostgREST/DbRequestBuilder.hs
+++ b/src/PostgREST/DbRequestBuilder.hs
@@ -10,6 +10,7 @@ A query tree is built in case of resource embedding. By inferring the relationsh
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE RecordWildCards       #-}
 
 module PostgREST.DbRequestBuilder (
   readRequest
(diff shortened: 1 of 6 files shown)


wolfgangwalther:
I moved the `parseArgs` part to SQL now. This does:
- increase coverage, because it avoids some edge cases in the parsers
- allow fool-proof parsing of argument names and types with spaces
- have the following side-effect:
  For functions without arguments, the schema cache seemed to store a single unnamed / untyped argument before:
  ```yaml
  - - qiName: welcome
      qiSchema: test
    - - pdArgs:
        - pgaName: ''
          pgaReq: true
          pgaType: ''
          pgaVar: false
        pdDescription: null
        pdHasVariadic: false
        pdName: welcome
        pdReturnType:
          contents:
            contents:
              qiName: text
              qiSchema: pg_catalog
            tag: Scalar
          tag: Single
        pdSchema: test
        pdVolatility: Volatile
  ```
  taken from https://github.com/PostgREST/postgrest/pull/1699/files#diff-153d54c3e6049f5c3ce9a68a92a10ccebd625e4ed7f13e2a51f908ae52aab1cbR60-R78

  Now, the schema cache has an empty `pdArgs` in that case. This might be related to some of the bugs we still have with overloading RPCs, although I didn't test that, yet. In any case the schema dump yaml files will become shorter ;).

wolfgangwalther:
memory tests have been failing in CI for this PR. I have no idea why - my local build is well short of the thresholds:

```
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(9,728,264 bytes) is less than 13M
ok 2 - POST /leak?columns=blob: with a json key of 1M the memory usage(9,798,432 bytes) is less than 13M
ok 3 - PATCH /leak?id=eq.1&columns=blob: with a json key of 1M the memory usage(9,701,448 bytes) is less than 13M
ok 4 - POST /rpc/leak?columns=blob: with a json key of 10M the memory usage(37,459,200 bytes) is less than 41M
ok 5 - POST /leak?columns=blob: with a json key of 10M the memory usage(37,537,168 bytes) is less than 41M
ok 6 - PATCH /leak?id=eq.1&columns=blob: with a json key of 10M the memory usage(37,439,024 bytes) is less than 41M
ok 7 - POST /rpc/leak?columns=blob: with a json key of 50M the memory usage(160,678,984 bytes) is less than 171M
ok 8 - POST /leak?columns=blob: with a json key of 50M the memory usage(160,754,392 bytes) is less than 171M
ok 9 - PATCH /leak?id=eq.1&columns=blob: with a json key of 50M the memory usage(160,663,608 bytes) is less than 171M
ok 10 - POST /perf_articles?columns=id,body: with a json payload of 32K that has 1000 array values the memory usage(6,828,480 bytes) is less than 11M
ok 11 - POST /perf_articles?columns=id,body: with a json payload of 329K that has 10000 array values the memory usage(7,721,816 bytes) is less than 11M
ok 12 - POST /perf_articles?columns=id,body: with a json payload of 3.4M that has 100000 array values the memory usage(17,046,840 bytes) is less than 21M
```

wolfgangwalther:
Memory seems to be fine in CI again.

A lot of the refactor is just about simplifying some pattern matches. Easier to read - less stale patterns that are not covered.

steve-chavez (approved):
LGTM!

Really like the parseArgs in SQL and the new addM2MRels is short and sweet :candy: .

PR #1721, opened 2020-12-31
nix: Fix build errors with stale .tix files present

wolfgangwalther:
Since we enabled the hpc flags on every build (even without `postgrest-coverage`), `.tix` files are created on every command. Sometimes regular commands (test-spec, test-io, build, run) will fail because of those stale `.tix` files.

Here, I add a tmp `HPCTIXFILE` location for every nix script to avoid those files from ever appearing. This can be overriden either from inside the script by setting the `HPCTIXFILE` location manually (as is done in `postgrest-coverage`) or from the outside by setting `HPCTIXFILE` before running the command.

### DIFF
--- a/nix/overlays/checked-shell-script/checked-shell-script.nix
+++ b/nix/overlays/checked-shell-script/checked-shell-script.nix
@@ -22,6 +22,14 @@ let
         ''
           #!${runtimeShell}
           set -euo pipefail
+
+          # make sure that no script creates conflicting tix files
+          # postgrest-coverage will override HPCTIXFILE to make proper use of it
+          if test ! -v HPCTIXFILE; then
+            hpctixdir=$(mktemp -d)
+            export HPCTIXFILE="$hpctixdir"/postgrest.tix
+            trap 'rm -rf $hpctixdir' EXIT
+          fi
         ''
         + lib.optionalString inRootDir ''
           cd "$(${git}/bin/git rev-parse --show-toplevel)"
@@ -32,7 +40,7 @@ let
             exit 1
           fi
         ''
-        + text;
+        + "(${text})";
 
       checkPhase =
         ''
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -24,6 +24,11 @@ flag dev
   manual:      True
   description: Development flags
 
+flag hpc
+  default:     True
+  manual:      True
+  description: Enable HPC (dev only)
+
 library
   default-language:   Haskell2010
   default-extensions: OverloadedStrings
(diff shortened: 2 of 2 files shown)


nix/overlays/checked-shell-script/checked-shell-script.nix
```diff
@@ -22,6 +22,14 @@ let
         ''
           #!${runtimeShell}
           set -euo pipefail
+
+          # make sure that no script creates conflicting tix files
+          # postgrest-coverage will override HPCTIXFILE to make proper use of it
+          if test ! -v HPCTIXFILE; then
+            hpctixdir=$(mktemp -d)
+            export HPCTIXFILE="$hpctixdir"/postgrest.tix
+            trap 'rm -rf $hpctixdir' EXIT
```
monacoremo, line 31: This trap is overridden if the script passed in as `text` sets its own trap later. I think it would be better if we disabled hpc by default, is this feasible in a reasonable way?

wolfgangwalther: hm. I enabled the hpc flags for all builds to avoid having to rebuild everything when switching between regular commands and the coverage command. It feels a lot smoother in development.

I could just add the good old `rm -rf *.tix` as a simple solution - but I felt, it'd be good to have some kind of way to explicitely tell those commands to keep collecting .tix data.

wolfgangwalther: And once I start collecting tix files in the repo root, it'd be annoying if those are gone all of a sudden...

monacoremo: Maybe we could add a flag to shell.nix to enable a hpc/coverage mode? It feels a bit messy to always have hpc on. It probably also impacts testing and later benchmarking performance, no?

Thinking of it, what would probably work here in the interim is `HPCTIXFILE=/dev/null`?

wolfgangwalther: > Maybe we could add a flag to shell.nix to enable a hpc/coverage mode? It feels a bit messy to always have hpc on.

That would be easily possible, but wouldn't really solve the problem. Even with the flag enabled, either per default or explicitly, I would still want those commands to just work without manually deleting .tix files inbetween.

> It probably also impacts testing and later benchmarking performance, no?

That's something I also thought about, but decided to defer it to later, once we're there.

> Thinking of it, what would probably work here in the interim is `HPCTIXFILE=/dev/null`?

Nope, that was the first thing I tried :). Once the file exists, the hpc-enabled-binary will try to read it and will fail if the file is empty.

nix/overlays/checked-shell-script/checked-shell-script.nix
```diff
@@ -32,7 +40,7 @@ let
             exit 1
           fi
         ''
-        + text;
+        + "(${text})";
```
wolfgangwalther, line 43: This should run the remaining script in a sub-shell. The sub-shell will inherit almost everything - except the trap.

wolfgangwalther:
Added the hpc flag to postgrest.cabal, too. For now it's still enabled by default, so this does not change anything. You should be able to pass `-f -hpc` to any cabal command to be able to disable it, though.

PR #1722, opened 2021-01-01
nix: Add postgrest-with-* tools

wolfgangwalther:
I need to run `postgrest-test-io` or `postgrest-run --dump-schema` against an older version. Right now they always use the latest pg version. If I do `test/with_tmp_db postgrest-run`, it will use my hosts pg version, but that won't really help.

Therefore I introduced `postgrest-with-postgresql-xx` commands. They replace the `postgrest-test-spec-postgresql-xx` commands, but are more flexible, because they take any other command as argument. The same goes for `postgrest-with-all`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,30 +200,26 @@ jobs:
           when: on_fail
       - run:
           name: Run the spec tests against PostgreSQL 9.5
-          command: postgrest-test-spec-postgresql-9.5
+          command: postgrest-with-postgresql-9.5 postgrest-test-spec
           when: always
       - run:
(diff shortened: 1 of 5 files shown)


wolfgangwalther:
I just realized I can even do `postgrest-with-postgresql-9.6 psql` and am launched right into a psql session with the test fixtures all set-up. That's so cool!

monacoremo:
> I just realized I can even do `postgrest-with-postgresql-9.6 psql` and am launched right into a psql session with the test fixtures all set-up. That's so cool!

Yes, this is a with_tmp_db feature :-) something we should document more prominently I guess!

PR #1723, opened 2021-01-01
Add basic error handling to --dump-schema

wolfgangwalther:
I worked with `--dump-schema` during development and had an error in the dbStructure queries. Before those changes I only got a pattern matching error, now the hasql error is shown on stdout. Formatting could probably be improved a little bit, because the query string is displayed with `\n` instead of newlines, but it's much better than before.


wolfgangwalther:
Codecov fails, but basic tests for `--dump-schema` are added in #1717. Errors during schema cache loading can't be tested properly, because those should never happen with checked-in code.

PR #1724, opened 2021-01-04
Refactor Middleware.hs

wolfgangwalther:
This refactoring increases coverage in Middleware.hs. We still need to add some tests for the error logging part, but this will have to wait until #1717 is merged, so we don't get too many merge conflicts on the IO tests. The haskell stuff is ready for review.

I did:
- Refactor the logger middleware avoiding the re-implementation of the upstream apache logger
- Refactor the CORS policy, where the settings were spread out across two functions. Seems like the incorrect `Authentication` header has been used here so far, too, which should be `Authorization`. In fact `Authorization` was used in "defaults", but was immediately overwritten in the other function with `Authentication`, so never used. Reformatted the test cases here, too.
- Added a few extra claims to the role claims tests, to increase test coverage in the `runPgLocals` function.

TODO:
- [x] Add logging tests


src/PostgREST/Middleware.hs
```diff
@@ -103,7 +102,6 @@ pgrstMiddleware logLevel =
     logger
   . gzip def
   . cors corsPolicy
-  . staticPolicy (only [("favicon.ico", "static/favicon.ico")])
```
steve-chavez, line 106: Hmm.. wait, I think this is not unreachable. If it's removed it will lead to 404 errors when doing a request through chrome:

I went to the address bar with: http://localhost:3000/projects

And the logs say:

```
127.0.0.1 - - [04/Jan/2021:11:28:08 -0500] "GET /projects HTTP/1.1" 200 - "" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.87 S
afari/537.36"
127.0.0.1 - - [04/Jan/2021:11:28:08 -0500] "GET /favicon.ico HTTP/1.1" 404 - "http://localhost:3000/projects" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML,
like Gecko) Chrome/78.0.3904.87 Safari/537.36"
```

wolfgangwalther: Hm. I had a feeling that my "test" was not a 100%.

I tried to access `/favicon.ico` with Firefox but got a 404 (with the code still in it). Maybe I didn't use the right accept headers or something?

steve-chavez: You can reproduce by doing a hard refresh(ctrl + shift + r) on chrome.

steve-chavez: Ah.. I've just noticed that this also happens on 7.0.1, so the `staticPolicy` is not doing anything.

wolfgangwalther: I just repeated that test... and the favicon loaded fine...

I probably tested the first time... when I had already removed it :D

I will add it back in and add a basic IO test instead.

I wonder, however: Are we packaging that file in the Docker container, too??

wolfgangwalther: > Ah.. I've just noticed that this also happens on 7.0.1, so the `staticPolicy` is not doing anything.

Wait, now I'm confused. It seemed to work for me on the current main branch, now.

wolfgangwalther: Actually... I might have tested the favicon on the docker container before removing it. Maybe that's the real issue here.

steve-chavez: > Ah.. I've just noticed that this also happens on 7.0.1, so the staticPolicy is not doing anything.

Confused as well, I ran the binary I have installed and the favicon is 404ing. Maybe the favicon.ico is not being packaged into the static binary? Not sure..

wolfgangwalther: Yes, I was only able to make it work by running it with `postgrest-run`. All packages are missing it, because it's only read at run-time right now.

steve-chavez: True, I also get the favicon with postgrest-run. Then that's another issue.


(Btw, the favicon is empty. Perhaps we should replace it with the [docs favicon](https://raw.githubusercontent.com/PostgREST/postgrest-docs/master/_static/favicon.ico))

src/PostgREST/Middleware.hs
```diff
@@ -69,74 +63,50 @@ runPgLocals conf claims app req = do
...
+logger logLevel app req sendResponse = app req $ \res ->
+  let
+    status = responseStatus res
+    logRequest = logStdout app req sendResponse
+  in
+  case logLevel of
+    LogInfo                        -> logRequest
+    LogWarn  | status >= status400 -> logRequest
+    LogError | status >= status500 -> logRequest
+    _                              -> sendResponse res
```
steve-chavez, line 89: Nitpick, more explicit:

```suggestion
    LogCrit                        -> sendResponse res
```

src/PostgREST/Middleware.hs
```diff
@@ -69,74 +63,50 @@ runPgLocals conf claims app req = do
...
-      LogError -> unsafePerformIO $ mkRequestLogger def { outputFormat = CustomOutputFormat $ pgrstFormat status500}
-      LogWarn  -> unsafePerformIO $ mkRequestLogger def { outputFormat = CustomOutputFormat $ pgrstFormat status400}
-      LogInfo  -> logStdout
 
-defaultCorsPolicy :: CorsResourcePolicy
-defaultCorsPolicy =  CorsResourcePolicy Nothing
-  ["GET", "POST", "PATCH", "PUT", "DELETE", "OPTIONS"] ["Authorization"] Nothing
-  (Just $ 60*60*24) False False True
+logger :: LogLevel -> Middleware
+logger logLevel app req sendResponse = app req $ \res ->
```
steve-chavez, line 80: Awesome! I didn't know it was that simple.

wolfgangwalther: tbh, it took me a while to figure that out ;)

src/PostgREST/Middleware.hs
wolfgangwalther: Hm. Just replacing `_` with `LogCrit` won't work, because `_` also catches the failing conditions for `LogWarn` and `LogError`. I think the `_` catchall fits best.

src/PostgREST/Middleware.hs
wolfgangwalther: I see two ways to go about this. We should either:
- serve the docs favicon as static file embedded into the executable
- or just remove the whole thing and add a condition in the logger to never log favicon errors

Given that the icon is empty currently, it seems like the reason to put this in was just about logging?

steve-chavez: > or just remove the whole thing and add a condition in the logger to never log favicon errors

Though it would still show as an error on chrome console.

> serve the docs favicon as static file embedded into the executable

Would this be hard to do?

src/PostgREST/Middleware.hs
steve-chavez: True. Didn't saw that one. It's good to leave it as it is then.

src/PostgREST/Middleware.hs
wolfgangwalther: I found this:
https://hackage.haskell.org/package/wai-middleware-static-embedded

I will give it a try.

src/PostgREST/Middleware.hs
```diff
@@ -69,74 +63,50 @@ runPgLocals conf claims app req = do
...
-    <> toLogStr (show (statusCode status)::Text)
-    <> " "
-    <> toLogStr (maybe "-" show responseSize::Text)
-    <> " \""
-    <> toLogStr (fromMaybe mempty $ requestHeaderReferer req)
-    <> "\" \""
-    <> toLogStr (fromMaybe mempty $ requestHeaderUserAgent req)
-    <> "\"\n"
-  where
-    getSourceFromSocket = BS.pack . showSockAddr . remoteHost
```
steve-chavez, line 99: Just noted that thanks to this refactor, fast-logger and wai-logger can be removed from the cabal file.

Basically undo this change: https://github.com/PostgREST/postgrest/commit/f6b6abe7340d906c698d47a67132017495c0ecbd#diff-eb6a76805a0bd3204e7abf68dcceb024912d0200dee7e4e9b9bce3040153f1e1

src/PostgREST/Middleware.hs
wolfgangwalther: After the discussion in #2092 I am not sure anymore, whether we should add back the favicon. It does "pollute" the api namespace.

And I'm not sure about adding a condition to the logger either. If somebody used this endpoint somehow, it would be really strange to not log requests about it.

I'll just remove the dysfunctional code and stick with that for now.

src/PostgREST/Middleware.hs
wolfgangwalther: It's much better now, that https://github.com/yesodweb/wai/pull/866 is merged.

wolfgangwalther:
Based on #2111 now - need to figure out why those tests are red. After that, this should be good to merge.

wolfgangwalther:
The loadtest is expected to fail, because it can't deal with removing dependencies in `postgrest.cabal`. This is because, the nix environment is started on the PR branch, but during `postgrest-loadtest-against` the main branch is checked out. The main branch will then try to find the now-removed dependency... but can't, because nix doesn't provide it.

src/PostgREST/Middleware.hs
steve-chavez: > It does "pollute" the api namespace.

True. Just realized this as well.

> If it's removed it will lead to 404 errors when doing a request through chrome
> Ah.. I've just noticed that this also happens on 7.0.1

Since nobody complained maybe the 404 was a non-issue from the get go. Removing it should be fine.

steve-chavez (approved):

test/io-tests/test_io.py
```diff
@@ -773,3 +774,43 @@ def test_admin_not_found(defaultenv):
...
         assert response.status_code == 404
+
+
+@pytest.mark.parametrize(
+    "level, has_output",
+    [
+        ("info", [True, True, True]),
+        ("warn", [False, True, True]),
+        ("error", [False, False, True]),
+        ("crit", [False, False, False]),
```
steve-chavez, line 785: Nice concise tests :100:

PR #1364, opened 2019-07-27
LTS 13.29

diogob:
I've done some changes to the `circleci/config.yml` scripts to get a faster feedback loop when trying to tweak the build:
* Updated stack to 2.1.3
* Save the cache a bit earlier so we have usable cache even if the build or linting fails.
* Split the build steps into dependencies and main target so we can save the cache for the dependencies. 

I have also done some changes to the `QueryBuilder` module to avoid the out of memory issues, breaking the module into several smaller internal modules.

Also added the `stack.yaml.lock` which is a new feature in stack to improve build reproducibility (apparently this is really a word 😜 ).

@steve-chavez I have started my branch on top of your `release-6.0.1` branch (I was assuming that you would merge that one first). But let me know if you want me to remove those commits or rebase against some other branch.

Cheers!

PR #1363, opened 2019-07-24
Release 6.0.1

steve-chavez:
I'm inclined to do a new PATCH release since the previous release process failed badly.

Windows was not possible to build with v6.0.0 and I had to use a different GHC and libraries to successfully compile and add a binary to the releases page(which doesn't seem correct).

Also the `text/html` feature(actually a bug for most users I'm sure) introduced in https://github.com/PostgREST/postgrest/issues/1308 was fixed and removed in https://github.com/PostgREST/postgrest/pull/1349. I'd like to properly document v6.0 without that.

I want to call this release a `v6.0.1` instead of `v6.1.0` because I want to communicate the release is more of a PATCH, despite there being 2 new features.

PR #1725, opened 2021-01-04
Refactor App.hs and related changes

monacoremo:
I struggled to follow what was happening in `App.hs`, so I tried to simplify the structure by decomposing the `app` function, which has  grown to become quite huge over time.

* Prefixed most functions and types by their module, making it obvious what is coming from where. No wildcard imports.
* Took out repetition, e.g. failing on non-singular results or setting GUCs is happening in one respective function
* Used ExceptT to combine pure error handling with e.g. IO or Hasql.Transaction, allowing us to avoid the deeply nested Either case analysis

The result is more verbose, but IMO also easier to follow and maintain. I kept the refactor purely to `App.hs` to keep the diff simpler, but there are some obvious simplifications in other modules that result from this.

I'm not a fan of the right-padding on imports and in type declarations, as they are a pain to maintain and lead to unclean diffs. Maybe that's something we can reconsider in our stylishhaskell settings?


wolfgangwalther:
I only skimmed through it on the surface, so far. Here's a couple of notes.

I really like this: `codecov/project 82.97% (+2.27%) compared to 5223082`


> * Prefixed most functions and types by their module, making it obvious what is coming from where. No wildcard imports.

In general, I like that. Although I wonder whether we should use shorter prefixes. At times this almost reads like Java ;)

One thing I noted independently of this PR: I think the whole `Types.hs` file is just a mess. We have some type definitions in this file, but others in the respective module's files where they belong semantically. I find myself jumping between the "feature" module and `Types.hs` quite often when developing one thing. I feel like most of the type definitions could go into one of the module directly. E.g. all the database structure types would be much better off in `DbStructure.hs`. Of course, the types are _used_ somewhere else, too - but they originate from `DbStructure`.

WDYT?

> * Took out repetition, e.g. failing on non-singular results or setting GUCs is happening in one respective function

+1 - I had thought about that one before, too.

> * Used ExceptT to combine pure error handling with e.g. IO or Hasql.Transaction, allowing us to avoid the deeply nested Either case analysis

I really like the flat structure - the deep nesting was hard to follow for me, too.

> I'm not a fan of the right-padding on imports and in type declarations, as they are a pain to maintain and lead to unclean diffs. Maybe that's something we can reconsider in our stylishhaskell settings?

TBH, everything that stylishhaskell does, I don't care about during development - so I don't find it hard to maintain. I never pad imports manually. I just add them and when I'm done `postgrest-style` will do the padding for me. Same for the unclean diffs - you can just enable "ignore whitespace" when displaying diffs (at least here in the GitHub PR) and all those won't show up anymore.

I feel like aligning things often helps with readability, especially in highly repeated patterns like imports, type declarations or pattern matching / case statements. For a couple of months now, I'm aligning things heavily in SQL, too (see #1702 for an example and https://www.sqlstyle.guide/ for the background) and I feel like this really helps readability a lot.

For me readability is more important than the additional effort needed to pad things nicely. Coming back at a piece of code after a longer time to change something, I spent the most time reading and understanding what was done before - not padding things after the change.

wolfgangwalther:
> > * Prefixed most functions and types by their module, making it obvious what is coming from where. No wildcard imports.
> 
> In general, I like that. Although I wonder whether we should use shorter prefixes. At times this almost reads like Java ;)

Maybe we can do something in the middle, too.

For external modules:
- No wildcard imports
- Use short prefixes (we already do in most of the cases, but standardize them a bit more across files) for standard tools like `L` for `Data.List`, `S` for `Data.Set`, `M` for `Data.Map`, `BS` for `Data.ByteString` or `LBS` for the lazy bytestring. `LazyByteString` is just too much.
- At the same time I really like `JSON` for aeson etc. - I think just  `SQL` for all the hasql stuff would be just fine, too.

For internal modules:
- No qualified imports in general. Feels way too much like Java - in other words: too much to read without meaning.
- Exceptions would be utilities like `Common.hs`, that might be used across more files. Those could basically be treated similar to an external module.
- Remove `Types.hs` (and possibly `QueryFragments.hs`) and split them up to the modules files where those are mainly used. This should remove a lot cross-file references.
- Import things explicitely. A huge list of imported functions from one file is probably an indication that we didn't separate our modules properly and should move some stuff around.

steve-chavez:
Took me a while to get used to it. But now I'm linking the new structure :)

(I guess I built [compression tolerance](http://steve-yegge.blogspot.com/2008/02/portrait-of-n00b.html) and got too used to `App.hs` code density)

Since this is a big refactor, one thing I like to see is if this change brings a perf downgrade, doesn't look like it but I want to be certain(maybe fusion matters here). So perhaps run this change with the load tests on https://github.com/PostgREST/postgrest/pull/1609 and compare?

> For external modules:

> No wildcard imports

How about Prelude/Protolude? I think we can make an exception in this case.

> Use short prefixes 

Short prefixes and SQL for the Hasql import sounds good.

> For internal modules:

> No qualified imports in general. Feels way too much like Java - in other words: too much to read without meaning.

Strongly agree on this one. Too verbose. I feel it optimizes too much for the newcomer and makes reading code harder for more experienced postgrest devs.

> Remove Types.hs (and possibly QueryFragments.hs) 

I remember the main reason for putting things in `Types.hs` was to avoid cyclical module dependencies errors. Types.hs is like a Common.hs in this regard. Not sure if it's possible to get the types in their own module.

Also, keep in mind that `QueryFragment.hs` was created to avoid out of memory issues on CI. QueryBuilder was too big to compile, see https://github.com/PostgREST/postgrest/pull/1364#issue-301724669.

wolfgangwalther:
> How about Prelude/Protolude? I think we can make an exception in this case.

For me, without a haskell background, Prelude/Protolude is about the same as every other external module :D. I'm mostly confused by all the `hiding (...)` imports - but I guess those won't happen anymore, once all the other imports are qualified. So I'd be fine with wildcard importing those.



> I remember the main reason for putting things in `Types.hs` was to avoid cyclical module dependencies errors. Types.hs is like a Common.hs in this regard. Not sure if it's possible to get the types in their own module.
> 
> Also, keep in mind that `QueryFragment.hs` was created to avoid out of memory issues on CI. QueryBuilder was too big to compile, see [#1363 (comment)](https://github.com/PostgREST/postgrest/pull/1363#issuecomment-515147114) and [#1364 (comment)](https://github.com/PostgREST/postgrest/pull/1364#issue-301724669).

Ah, thanks for the background on this. Maybe we can still make the cut somewhere else and just put a few types in `Common.hs`.

steve-chavez:
Another thing about Protolude is that symbols would be too verbose prefixed: 
```haskell
<$>, <*>, =<<
-- vs
Protolude.<$>, Protolude.<*>, Protolude.=<<
```

Enforcing the qualifier would defeat the purpose to have symbols, I think.

---

I've noted that Hasql uses qualified imports for internal modules, but some of them are short(like `A`, `B`): https://github.com/nikita-volkov/hasql/blob/master/library/Hasql/Private/Decoders.hs#L9-L17.

Still, I'd prefer not qualifying internal modules.

steve-chavez:
Actually I'm finding some functions(like `Auth.containsRole`) more readable with qualifiers. The main issue is the modules with long names. Shortening those might be another solution.

wolfgangwalther:
> Actually I'm finding some functions(like `Auth.containsRole`) more readable with qualifiers. The main issue is the modules with long names. Shortening those might be another solution.

I feel like this is more an exception than the norm. For many of the exported functions we currently have, it's very easy to see which module they belong to. This is most prominent with all the field lookup functions for records functions. Just look at this:

```haskell
handleRequest context@(RequestContext _ dbStructure apiRequest _) =
  case (ApiRequest.iAction apiRequest, ApiRequest.iTarget apiRequest) of
```

The change I would make here, would be to change the field prefix for `ApiRequest` from `i` to `req`. This would be more in line with all the other types we have. And then I would add `LANGUAGE RecordWildCards` and write it like this:

```haskell
handleRequest context@(RequestContext _ dbStructure ApiRequest{..} _) =
  case (reqAction, reqTarget) of
```

In my opinion this is muuch quicker to "parse" visually and basically just as explicit, because of the prefix in the field name.

monacoremo:
Agree that some parts have become quite verbose and depart a bit from the typical Haskell style. I think this is a very valuable discussion we are having here, as I think there is a huge value in having the code as easy to follow and accessible as possible. PostgREST is one of the more visible open source projects outside the Haskell ecosystem, and a entryway for some new users of the language (at least it was for me).

So let me be the devil's advocate in a few points :-) 

> For many of the exported functions we currently have, it's very easy to see which module they belong to.

I think ideally, we would turn this convention around, avoiding prefixes within modules and using prefixed imports to disambiguate. For example in the Elm ecosystem (another ML language), it's customary to prefix all imported variables (https://elm-lang.org/docs/style-guide). This makes Elm code very easy to follow. It also avoids polluting modules with prefixes internally.

So we would at some point use `Config.jwtRoleClaimKey` instead of `configJwtRoleClaimKey`. This makes it clearer what the origin is, while having the same length, is easier to parse and it will make the `Config` module terser internally.

> And then I would add LANGUAGE RecordWildCards and write it like this:

That looks like a great simplification to make it more readable!

> Short prefixes and SQL for the Hasql import sounds good.

I think `Hasql`, `Aeson`, and a bit longer prefixes than `S` or `L` in general are much more easy to follow for new readers of the code. Looking up `JSON decode` on Google yields nothing useful, while `Aeson decode` does. The same goes for `SQL pool` vs. `Hasql pool`. Having to refer back to the imports all the time is a significant distractions (what is `S` again?), while having a bit longer names within the module is not a significant cost.

> How about Prelude/Protolude? I think we can make an exception in this case.

Agree that this should be one exception for wildcard import, as Protolude conceptionally replaces the Prelude that is available by default. At some point we might want to roll our own Prelude, as e.g. Hasql does.

> Remove Types.hs (and possibly QueryFragments.hs)

Totally agree that we should split this up. If circular imports are a concern, we could still have PostgREST.Types.DbStructure, PostgREST.Types.ContentTypes etc. separately.

wolfgangwalther:
Thanks for adding the refactor for `Auth.hs`, @monacoremo. The file is much smaller than `App.hs` and a lot easier to discuss about regarding imports and general style. I see you did a lot less regarding those in this commit. I feel like the best way would be somewhere inbetween the two styles.

A couple of notes:
- The first thing I note when I go through the `Auth.hs` file is how often I ask myself: "Where does this type come from?" Turns out all of them come from the unqualified `Crypto.JWT` import. I very much feel this needs a prefix, too!
- I think we should remove the `Control.Lens` stuff. To me it looks like we use lenses only very inconsistently across the whole code-base. We can do just the same with basic record syntax here.
- A `where` helps me a lot to immediately tell which functions are only used in a certain scope. `validation`, `jwtClaimsError` and `claims2map` are all only used in `jwtClaims` - and there are more examples in the same file. Putting those into `where` might look like some sort of "nesting" - but it's not. It's just indentation according to context and doesn't increase complexity. The style as it is currently (and was before, too!) is **too flat**!
- In a short function like `jwtClaims` the blank lines are just a waste of screen space, imho. Each of those blank lines just separates a single statement. This is the same thing that indentation does visually - no gain. But the consequence is that all functions are separated by two blank lines now. Even with a where clause it will be harder to fit the whole `jwtClaims` functions (including dependants) on the screen at the same time (well not on my screen, but on a bigger function I would have the same problem). I'm just missing the "one look at the whole function+context" with this style - and that makes it harder for me to read.

wolfgangwalther:
> I think this is a very valuable discussion we are having here, as I think there is a huge value in having the code as easy to follow and accessible as possible.

Fully agree!

> I think ideally, we would turn this convention around, avoiding prefixes within modules and using prefixed imports to disambiguate. [...]
> So we would at some point use `Config.jwtRoleClaimKey` instead of `configJwtRoleClaimKey`. This makes it clearer what the origin is, while having the same length, is easier to parse and it will make the `Config` module terser internally.

Yes, I thought about this briefly, too. It will however:
- increase the likeliness of field name conflicts across different record types (think `name` or something general like this)
- make `LANGUAGE RecordWildCards` harder to understand, because the automatically extracted fields do not have the prefix anymore. This might also create those conflicts mentioned across types in different files - suddenly you might not be able to unpack two unrelated record types in the same context anymore.
- it would also mean that we'd need to prefix all value constructors. `ApiRequest.ApiRequest` and so on. I feel like the code-base is not too big to be able to avoid that. If we avoid wildcard imports, it will be one find in the currently open file max to see where the type comes from.

Using "proper" prefixes instead of "in-name-prefixes" felt more "proper" for me, too - at first. But now I like the field name prefixes that we have.

> I think `Hasql`, `Aeson`, and a bit longer prefixes than `S` or `L` in general are much more easy to follow for new readers of the code. Looking up `JSON decode` on Google yields nothing useful, while `Aeson decode` does. The same goes for `SQL pool` vs. `Hasql pool`. Having to refer back to the imports all the time is a significant distractions (what is `S` again?), while having a bit longer names within the module is not a significant cost.

_"Looking up `JSON decode` on Google yields nothing useful"_ - that might be true, but "haskell JSON decode" will immediately point you to Aeson. To be honest, I had quite some trouble with `JSON` and `Aeson` when starting with haskell in postgrest. Googling "haskell JSON xxx" would always yield this strange "Aeson" thing... :)
But the good thing about the JSON prefix was, that I immediately knew what the code was about. It was about JSON stuff. Without even looking up Aeson, I know whether that's something I care about right now or not, even in the very beginning. It's the same thing with `SQL` vs `Hasql`, but much easier of course, because sql is part of hasql...

Bottom line is: Start out with the PostgREST code-base and you'll have to learn some commonly used prefixes used for import. Once you understand we're using `S` for `Set` and `L` for `List` everywhere and do the same with `JSON` and `SQL` - it will be no problem, even for beginners.

> > How about Prelude/Protolude? I think we can make an exception in this case.
> 
> Agree that this should be one exception for wildcard import, as Protolude conceptionally replaces the Prelude that is available by default. At some point we might want to roll our own Prelude, as e.g. Hasql does.

Absolutely, I agree with both of you. It looks like `liftEither` and `mapLeft` definitely belong in our own Prelude after your refactoring :D.

test/Feature/SingularSpec.hs
```diff
@@ -70,8 +70,7 @@ spec =
           `shouldRespondWith`
             [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned"}|]
             { matchStatus  = 406
-            , matchHeaders = [ matchContentTypeSingular
-                             , "Preference-Applied" <:> "tx=commit" ]
+            , matchHeaders = [ matchContentTypeSingular ]
```
monacoremo, line 73: With the refactored way we handling errors, errors go straight through `optionalRollback` untouched and this header is not being added. I believe this is fine, as the transaction is always rolled back on error and it doesn't matter what the `tx` preference was. As it's clear what happened to the user, the header is not needed according to https://tools.ietf.org/html/rfc7240#page-7 section 3.

monacoremo: @wolfgangwalther WDYT?

monacoremo:
@steve-chavez There will be conflicts with #1729, let's get that PR merged first and then I'll rebase/backport the changes here

src/PostgREST/CLI.hs
```diff
@@ -0,0 +1,351 @@
...
+    let postgrestApplication =
+          postgrest
+            logLevel
+            refConf
+            refDbStructure
+            pool
+            getTime
+            connWorker
+
+    runWithSocket conf maybeSocketAddr serverSettings postgrestApplication
```
wolfgangwalther, line 342: `conf` is only used here to get `configServerUnixSocketMode`. Pass the value of that directly and it will remove another internal dependency (Main->Config, currently).

src/PostgREST/CLI.hs
```diff
@@ -0,0 +1,351 @@
+{-|
+Module      : PostgREST.CLI
+Description : Command line interface.
```
wolfgangwalther, line 3: If we still keep the distinction between executable and library - this module should really be part of the executable, no? I don't think we test any of that in the spec tests, only in the io tests - that seems like a sign that all of this can only be used in the executable.

src/PostgREST/Config.hs
```diff
@@ -42,7 +42,7 @@ import qualified Data.Map.Strict        as M
```
wolfgangwalther, line 42: can't comment on the hidden line above (the `Map` import). If that is still used after the split-up into `CLI`, the same reasoning for naming etc. applies as I commented there.

src/PostgREST/Config.hs
```diff
@@ -20,20 +20,20 @@ Other hardcoded options such as the minimum version number also belong here.
...
+  , prettyVersion
+  , readValidateConfig
+  ) where
+
+import qualified Crypto.JOSE.Types      as JOSE
+import qualified Crypto.JWT             as JWT
+import qualified Data.Aeson             as JSON
 import qualified Data.ByteString        as B
 import qualified Data.ByteString.Base64 as B64
 import qualified Data.ByteString.Char8  as BS
```
wolfgangwalther, line 39: Those seem like they'd need some kind standardization. Are we using `Char8` anywhere else? I suggest we use `Data.ByteString as BS` - and then `BS64` and `BS8` or something?

test/Feature/SingularSpec.hs
wolfgangwalther: Hm. I think it's possible to commit a transaction and still return a `4xx` / `5xx` error code via GUC: https://postgrest.org/en/latest/api.html#setting-response-status-code

This means, that you can not tell from the error code alone whether the transaction was comitted or rolled back.

Is it possible to refactor to avoid this change?

wolfgangwalther:
I glanced over most files, but not `App.hs` so far.

Many of my comments are about import statements - and I guess half of those 
about some statements, that you didn't even change, @monacoremo. So there is a lot of inconsistency about that already currently. I'm not sure how to proceed here.

I know we should get this PR merged rather quickly, because it touches so much code - it will be a pain to keep this open.

postgrest.cabal
```diff
@@ -38,7 +38,9 @@ library
   exposed-modules:    PostgREST.ApiRequest
                       PostgREST.App
                       PostgREST.Auth
+                      PostgREST.CLI
                       PostgREST.Config
+                      PostgREST.Connection
```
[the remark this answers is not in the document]
steve-chavez: I think it was mainly to allow Haskell devs to use postgrest as a lib and extend it. Not that I have ever done that, but I'm not sure if we're invalidating haskell libs such as https://github.com/diogob/postgres-websockets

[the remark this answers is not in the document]
monacoremo: We need both the lib and executable: We need modules to be in a lib in order to test them, and we obviously need an executable to generate `postgrest`. What I see most projects do is to pull as much as possible into the lib (where it can be tested and inspected) and have a minimal one-liner Main module. E.g. Pandoc: https://github.com/jgm/pandoc/blob/master/app/pandoc.hs

In our case, we have the extra wart of having to use CPP macros to conditionally compile in non-Windows features. it's good to isolate those in Main I think, while having as much as possible available for testing in the lib.

main/Main.hs
```diff
@@ -1,398 +1,36 @@
...
-          (\_ -> do -- if the thread dies, we try to recover
-            putStrLn retryMessage
-            connWorker -- assume the pool connection was also lost, call the connection worker
-            start)     -- retry the listener
-        _ ->
-          putStrLn errorMessage -- Should be unreachable. connectionStatus will retry until there's a connection.
-    errorMessage = "Could not listen for notifications on the " <> dbChannel <> " channel" :: Text
-    retryMessage = "Retrying listening for notifications on the " <> dbChannel <> " channel.." :: Text
+main :: IO ()
+main = CLI.app installHandlers runWithSocket
```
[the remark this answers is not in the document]
monacoremo: That will not quite work, as we need to swallow the unused arguments... I had a earlier version with lambdas, which was messier. An alternative would be passing Maybes into app, but that would cause a bit of more complexity there. WDYT?

monacoremo:
> I know we should get this PR merged rather quickly, because it touches so much code - it will be a pain to keep this open.

Yes, but no need to hurry - I'll rebase as needed

main/Main.hs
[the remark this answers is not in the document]
wolfgangwalther: I see. Using Maybe here sounds semantically correct.

postgrest.cabal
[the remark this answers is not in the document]
wolfgangwalther: > I think it was mainly to allow Haskell devs to use postgrest as a lib and extend it. Not that I have ever done that, but I'm not sure if we're invalidating haskell libs such as https://github.com/diogob/postgres-websockets

Doesn't look like anyone is using PostgREST as a lib:
https://packdeps.haskellers.com/reverse

main/Main.hs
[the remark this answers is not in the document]
monacoremo: Fixed!

src/PostgREST/Config.hs
monacoremo: Used BS8 for now

src/PostgREST/CLI.hs
monacoremo: Fixed

src/PostgREST/CLI.hs
```diff
@@ -0,0 +1,351 @@
...
+  , app
+  ) where
+
+import qualified Data.Aeson                 as Aeson
+import qualified Data.ByteString.Lazy       as LBS
+import qualified Data.Map.Strict            as M
+import qualified Hasql.Connection           as C
+import qualified Hasql.Pool                 as P
+import qualified Hasql.Session              as S
+import qualified Hasql.Transaction.Sessions as HT
```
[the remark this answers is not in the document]
monacoremo: Fixed, pulled everything in as SQL

src/PostgREST/Auth.hs
```diff
@@ -12,108 +10,77 @@ Authentication should always be implemented in an external service.
...
 
-import Control.Lens    (set)
-import Data.Time.Clock (UTCTime)
+import Control.Lens            (set)
+import Control.Monad.Except    (liftEither)
+import Data.Either.Combinators (mapLeft)
+import Data.Time.Clock         (UTCTime)
 
-import Crypto.JWT
+import Crypto.JWT (JWTError (..))
```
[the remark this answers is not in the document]
monacoremo: Fixed. The general distinction is not there yet, will have a look at all our imports jn a separate PR

test/Feature/SingularSpec.hs
monacoremo: Thinking a bit about this @wolfgangwalther : It would be relatively simple to always set the preference header (catch any errors, turn them into a response, set the header), but I wonder whether the new behavior that emerged here is more correct.

Setting a custom status via GUC still works and will set the preference header as before, regardless of what the status code is. 

Setting the preference header will only be skipped when a true error happened and the transaction will definitely be rolled back. In those cases, showing tx=commit seems misleading.

wolfgangwalther: > Setting a custom status via GUC still works and will set the preference header as before, regardless of what the status code is.
> 
> Setting the preference header will only be skipped when a true error happened and the transaction will definitely be rolled back. In those cases, showing tx=commit seems misleading.

The whole thing is a bit more complicated, because we have `db-tx-end=commit|commit-allow-override|rollback-allow-override|rollback` settings. This creates a lot of different scenarios.

I understand RFC 7240 as follows: We should only ever return a preference that was asked for with `Preference-Applied`. This means we should never return a `Preference-Applied: tx=rollback` header, when we had a `Prefer: tx=commit` header before.

This basically means that we can not use the `Preference-Applied` header for feedback about the actual result ("committed" or "rolled back") of the transaction. A couple of examples, where this fails:

Assume an error response with `4xx` status code. With `db-tx-end=commit-allow-override` and `Prefer: tx=commit`: Setting a `Preference-Applied: tx=rollback` seems wrong. How would we differ between the "error via RAISE" and "error via GUC" cases then? We could only return nothing for RAISE (i.e. a true rollback) and `Preference-Applied: tx=commit` for the GUC case (i.e. a true commit). This basically assumes that the transaction was rolled back for any `4xx` error code by default. But now assume we have `db-tx-end=commit` and `Prefer: tx=rollback`. For a GUC error response this results in a commit in the end, because the override is not allowed. But we can't return `Preference-Applied: tx=commit` - this was not asked for. And we can't assume a rollback by default either, because that wouldn't be true.

In short: We can only use the `Preference-Applied` header to indicate what our transaction termination command was (`COMMIT` or `ROLLBACK`) - completely independent of whether the transaction was implicitly rolled back before because of an error. Note, that this is also what is implied by the config option `db-tx-end`.

There are a couple of other reasons why it makes a lot of sense to return a `Preference-Applied` header whenever we understand and accept that preference - regardless of whether it might be sensible to assume this behaviour by default in that specific case. See #740.

The current behaviour should be kept. Our goal should be to respond with more `Preference-Applied` headers instead of less.

wolfgangwalther:
> > I know we should get this PR merged rather quickly, because it touches so much code - it will be a pain to keep this open.
> 
> Yes, but no need to hurry - I'll rebase as needed

The problem here is, that it makes reviewing a lot harder, too. Whereas it was reasonable to assume that code was unchanged at the very beginning when you only split some parts in different files - now a reviewer will have to check every commit that was rebased on and see whether those changes have been applied correctly.

I don't have an in-depth idea of how the various refactors are related - do you think it's possible to split some parts (files probably?) off into another PR and merge those independently? The whole `Main.hs` refactor seems independent enough?

monacoremo:
> I don't have an in-depth idea of how the various refactors are related - do you think it's possible to split some parts (files probably?) off into another PR and merge those independently? The whole Main.hs refactor seems independent enough?

Yes, true - all the exploration with `Main` and sorting out imports definitely got too much for one PR. 

I've split out all the `Main.hs` stuff, leaving the `App.hs` refactor and related changes:
* Adjusted `Error` type (which causes some renames in other modules)
* Refactored interfaces of `Auth.hs`, `Middleware.hs`
* `parseSecret` moved to `Config` to resolve an import loop with `Auth.hs`.

The original behavior of Preference-Applied headers is maintained.

Hope that makes it a bit overview.

src/PostgREST/CLI.hs
```diff
@@ -0,0 +1,351 @@
...
+  , readEnvironment
+  , dumpSchema
+  , HandlerInstaller
+  , SocketRunner
+  , app
+  ) where
+
+import qualified Data.Aeson                 as Aeson
+import qualified Data.ByteString.Lazy       as LBS
+import qualified Data.Map.Strict            as M
```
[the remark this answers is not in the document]
monacoremo: Next PR

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
-                      | shouldRollback  = addHeadersIfNotIncluded [(hPreferenceApplied, BS.pack (show Rollback))]
-                      | otherwise       = identity
-                    handleReq = do
-                      when (shouldRollback || (configDbTxRollbackAll conf && not shouldCommit)) HT.condemn
-                      mapResponseHeaders preferenceApplied <$> runPgLocals conf claims (app dbStructure conf) apiRequest
-                  dbResp <- P.use pool $ HT.transaction HT.ReadCommitted (txMode apiRequest) handleReq
-                  return $ either (errorResponseFor . PgError authed) identity dbResp
-        -- Launch the connWorker when the connection is down. The postgrest function can respond successfully(with a stale schema cache) before the connWorker is done.
-        when (responseStatus response == status503) connWorker
+  Middleware.pgrstMiddleware logLev $
```
wolfgangwalther, line 77: It's odd that we're passing in the log level and refConf separately. I think adding the middleware should be done in `Main.hs`. This is not part of the main app - it's just some additional stuff. Also this would remove the whole logging from the spec tests - which makes sense, because it's not used at all.

This might drop coverage by a few lines in `Middleware.hs` temporarily. But we need to add a few logging tests anyway and can do so for the cors stuff again, too.

After that, `\req respond ->` should be part of the main function arguments - reducing nesting/indentation a lot here.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+        Just dbStructure ->
+          return dbStructure
+
+        Nothing ->
+          throwError Error.ConnectionLostError
+
+    apiRequest <-
+      liftEither . mapLeft Error.ApiRequestError $
+        Req.userApiRequest
+          (Config.configDbSchemas conf)
```
wolfgangwalther, line 120: just like suggested in the other files:
Let's import `Config.AppConfig(..)` unqualified and then use record wildcards like `conf@{..}` in this function. Then `configDbSchemas` and `configDbRootSpec` will be available just like that. It works really nice for all the config stuff, because the field prefix just says everything.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+
+    dbStructure <-
+      case maybeDbStructure of
+        Just dbStructure ->
+          return dbStructure
+
+        Nothing ->
+          throwError Error.ConnectionLostError
+
+    apiRequest <-
```
wolfgangwalther, line 117:
```suggestion
    apiRequest@{..} <-
```

and import `ApiRequest.ApiRequest(..)` unqualified. `iJWT` and `iAccepts` will be available like that.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
-                  dbResp <- P.use pool $ HT.transaction HT.ReadCommitted (txMode apiRequest) handleReq
-                  return $ either (errorResponseFor . PgError authed) identity dbResp
-        -- Launch the connWorker when the connection is down. The postgrest function can respond successfully(with a stale schema cache) before the connWorker is done.
-        when (responseStatus response == status503) connWorker
+  Middleware.pgrstMiddleware logLev $
+    \req respond ->
+      do
+        time <- getTime
+        conf <- readIORef refConf
+        maybeDbStructure <- readIORef refDbStructure
```
wolfgangwalther, line 82: Is it possible to handle the `throwError Error.ConnectionLostError` here already and return `dbStructure` instead of `maybeDbStructure`?

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
-        else Left . errorResponseFor $ BinaryFieldError ct
-  | otherwise = Right Nothing
+postgrestResponse
+  :: AppConfig
+  -> Maybe DbStructure
+  -> SQL.Pool
+  -> UTCTime
+  -> Wai.Request
+  -> Handler IO Wai.Response
+postgrestResponse conf maybeDbStructure pool time req =
```
wolfgangwalther, line 105: I feel like this whole function should really be part of the `postgrest` function. Once the middleware is moved out of there, it should not increase nesting that much.

Both of those functions taken together just take all the different pieces that are done before and after handling a request and stitch them together. Right now they are spread across two functions, I think they fit better together.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+        runExceptT handler
+
+    resp <-
+      liftEither . mapLeft Error.PgErr $
+        mapLeft (Error.PgError $ Auth.containsRole jwtClaims) dbResp
+
+    liftEither resp
+
+handleRequest :: RequestContext -> DbHandler Wai.Response
+handleRequest context@(RequestContext _ dbStructure apiRequest _) =
```
wolfgangwalther, line 166:
```suggestion
handleRequest context@(RequestContext _ dbStructure ApiRequest{..} _) =
```

with an unqualified import of `ApiRequest(..)` (+ the Action and Target types) would remove all the `Req.` prefixes, that just waste space.

Would it be possible to bring the pattern and the call on one line, too? Ideally space-aligned. That would really improve readability **a lot**. (Moving the case on the first line would save 2 spaces indent, too.)

Right now it's just a wall of text with not too much info.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+      handleReq apiReq =
+        handleRequest (context apiReq)
+
+    runDbHandler pool (txMode apiRequest) jwtClaims $
+      Middleware.optionalRollback conf apiRequest $
+        Middleware.runPgLocals conf jwtClaims handleReq apiRequest
+
+runDbHandler :: SQL.Pool -> SQL.Mode -> Auth.JWTClaims -> DbHandler a -> Handler IO a
+runDbHandler pool mode jwtClaims handler =
+  do
```
wolfgangwalther, line 154: This is a regular pattern, not only here: I suggest to move the `do` to the previous line everywhere. This reduces indentation of the whole block by 2 spaces, saving screen real estate.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+          dbStructure
+          req
+          body
+
+    -- The JWT must be checked before touching the db
+    jwtClaims <- Auth.jwtClaims conf (toS $ Req.iJWT apiRequest) time
+
+    let
+      reqContentTypes = requestContentTypes conf apiRequest
+      acceptContentType = Req.iAccepts apiRequest
```
wolfgangwalther, line 131:
```suggestion
      reqContentTypes =
        requestContentTypes conf apiRequest
      acceptContentType =
        Req.iAccepts apiRequest
```

It seems like you're following a pattern like this in most other places here. I suggest to remove the blank lines everywhere like in this example however - keeps the `let` block together nicely, but still makes it easy to separate the different functions/variables.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+        Just ct ->
+          return ct
+
+        Nothing ->
+          throwError . Error.ContentTypeError $ map Types.toMime acceptContentType
+
+    let
+      context apiReq =
+        RequestContext conf dbStructure apiReq contentType
+
```
wolfgangwalther, line 144: Just as an example of what I meant with my comment above: Removing this blank line will keep the let block together nicely - much easier to tell the second part of the let (handleReq) and the next step in the do (runDbHandler) apart.

(probably not as nice to see, now that I split everything up with that comment... :D).

src/PostgREST/App.hs
```diff
@@ -9,430 +9,664 @@ Some of its functionality includes:
...
+      handleInfo identifier dbStructure
+    (Req.ActionInvoke invMethod, Req.TargetProc proc _) ->
+      handleInvoke invMethod proc context
+    (Req.ActionInspect headersOnly, Req.TargetDefaultSpec tSchema) ->
+      handleOpenApi headersOnly tSchema context
+    _ ->
+      throwError Error.NotFound
+
+handleRead :: Bool -> QualifiedIdentifier -> RequestContext -> DbHandler Wai.Response
+handleRead headersOnly identifier context@(RequestContext conf dbStructure apiRequest contentType) =
```
wolfgangwalther, line 188:
```suggestion
handleRead headersOnly identifier context@(RequestContext AppConfig{..} dbStructure ApiRequest{..} contentType) =
```

I think for those two, this would be a pattern that makes sense across the whole file. I commented a lot of places already - but it just keeps coming up again, and the more I comment it the more I like it. It just removes boilerplate.

wolfgangwalther:
I gave it a deeper review now. I'm still not done with `App.hs` completely (note to self: need to continue from `handleRead` on down), but already did the first part of it and also all the other files.

Since the whole point of this refactoring PR is readability - I included a lot of nitpicks, too, to improve readability just a tiny bit.

monacoremo:
Thank you for your thorough review @wolfgangwalther ! Will make adjustments :-)

src/PostgREST/OpenAPI.hs
```diff
@@ -25,16 +22,33 @@ import Control.Lens
 import Data.Swagger
 
 import PostgREST.ApiRequest       (ContentType (..))
-import PostgREST.Config           (docsVersion, prettyVersion)
+import PostgREST.Config           (AppConfig,
+                                   configOpenApiServerProxyUri,
+                                   configServerHost, configServerPort,
```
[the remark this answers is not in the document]
monacoremo: Good catch!

src/PostgREST/OpenAPI.hs
```diff
@@ -25,16 +22,33 @@ import Control.Lens
...
 import PostgREST.Private.ProxyUri (isMalformedProxyUri, toURI)
-import PostgREST.Types            (Column (..), ForeignKey (..),
-                                   PgArg (..), PrimaryKey (..),
+import PostgREST.Types            (Column (..), DbStructure,
+                                   ForeignKey (..), PgArg (..),
+                                   PrimaryKey (..),
                                    ProcDescription (..), Proxy (..),
-                                   Table (..), toMime)
+                                   Table (..), dbPrimaryKeys,
+                                   proxyHost, proxyPath, proxyPort,
```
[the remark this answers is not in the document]
monacoremo: Good catch!

src/PostgREST/Auth.hs
```diff
@@ -12,108 +10,77 @@ Authentication should always be implemented in an external service.
...
+        JWT.verifyClaimsAt validation secret time =<< JWT.decodeCompact payload
+    liftEither . mapLeft jwtClaimsError $ claimsMap jspath <$> eitherClaims
+  where
+    validation =
+      set JWT.allowedSkew 1 $ JWT.defaultJWTValidationSettings audienceCheck
+
+    audienceCheck :: JWT.StringOrURI -> Bool
+    audienceCheck = maybe (const True) (==) (Config.configJwtAudience conf)
+
+    jwtClaimsError :: JWTError -> Error
```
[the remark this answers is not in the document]
monacoremo: Got lost in rebase - reapplied!

src/PostgREST/App.hs
monacoremo: Yes, much cleaner.

src/PostgREST/Auth.hs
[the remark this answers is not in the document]
monacoremo: Got lost in rebase, fixed now,

src/PostgREST/App.hs
monacoremo: Yes, but it would be difficult to do without changing the behavior of when the connectionWorker is launched (https://github.com/monacoremo/postgrest/blob/a6ba8784e9096eee2e30f3c049ec7eeec9886a7c/src/PostgREST/App.hs#L90), which I'm not sure is fully correct right now, but I wouldn't want to change it in this PR.

src/PostgREST/App.hs
monacoremo: There is some point in separating the IORef stuff from the actual, `pure` response function. At this point, merging the functions without changing the behavior of https://github.com/monacoremo/postgrest/blob/a6ba8784e9096eee2e30f3c049ec7eeec9886a7c/src/PostgREST/App.hs#L90 will be tricky, so I'd leave it unchanged in this PR; let's review this in a separate one.

We'd also need an extra level of indentation, as we need to run everything in `runExceptT` to avoid the nested case analysis.

src/PostgREST/App.hs
monacoremo: Moving this to Main.hs changes the semantics, several Cors tests failed. Backed out the commit - recommend that we leave this unchanged for this PR.

monacoremo:
@wolfgangwalther Suggest that we defer a few changes to a separate PR, please see the comments above. All others should be reflected.

monacoremo:
@wolfgangwalther The coverage failure happens because we're no longer using the record fields in App.hs directly, instead we always unpack the records. Should probably be one of the coverage exceptions once we have full coverage.

wolfgangwalther:
I now reviewed the remaining parts of `App.hs`.

> Since this is a big refactor, one thing I like to see is if this change brings a perf downgrade, doesn't look like it but I want to be certain(maybe fusion matters here). So perhaps run this change with the load tests on #1609 and compare?

I did that and there is no difference in minimum latency with the vegeta test at all. That's only for a plain `GET` request, but I don't think there's reason to believe the other types of requests would make a difference.

wolfgangwalther:
> I really like this: `codecov/project 82.97% (+2.27%) compared to 5223082`

The initial increase in code coverage was not real, but just an effect of the refactor having a lot more lines of code initially, which increased the percentage.

However I checked the open coverage TODOs in #1719 for App.hs. Because of the changes to error handling, 3 of the 6 are done with this PR.



> @wolfgangwalther The coverage failure happens because we're no longer using the record fields in App.hs directly, instead we always unpack the records. Should probably be one of the coverage exceptions once we have full coverage.

Absolutely, I agree. Others have that problem, too - let's hope somebody will do something about it: https://gitlab.haskell.org/ghc/ghc/-/issues/17834

src/PostgREST/App.hs
```diff
@@ -9,430 +9,549 @@ Some of its functionality includes:
...
+
+  failNotSingular contentType rQueryTotal $
+    if iPreferRepresentation == Types.Full then
+      response HTTP.status200
+        (contentTypeHeaders context ++ [contentRangeHeader])
+        (toS rBody)
+    else
+      response HTTP.status204 [contentRangeHeader] mempty
+
+handleInfo :: Monad m => QualifiedIdentifier -> DbStructure -> Handler m Wai.Response
```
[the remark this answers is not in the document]
monacoremo: I think the idiomatic solution would be to make our `Handler` a `ReaderT RequestContext` - then we wouldn't need to pass the context around everywhere, and the functions could `ask` for exactly what they need. It seems a bit more obscure then what we currently have, though, so I'd save that for a separate PR where we can discuss it.

I typed `handleInfo` as `Monad m => Handler m Wai.Response` to make it clear that `handleInfo` does not (and because of its type, cannot) access the database. Haven't found a better way to express it, happy for any suggestions. I would be reluctant to type it `DbHandler` just for consistency, as it's actually a special case.

src/PostgREST/App.hs
```diff
@@ -9,430 +9,549 @@ Some of its functionality includes:
...
+    (ActionInvoke InvHead, _) ->
+      SQL.Read
+    (ActionInvoke InvPost, TargetProc Types.ProcDescription{Types.pdVolatility=Types.Stable} _) ->
+      SQL.Read
+    (ActionInvoke InvPost, TargetProc Types.ProcDescription{Types.pdVolatility=Types.Immutable} _) ->
+      SQL.Read
+    _ ->
+      SQL.Write
+
+data CreateResult = CreateResult
```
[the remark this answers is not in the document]
monacoremo: Agree, was difficult to follow. Added a comment and renamed.

This type and the `writeQuery` function are some glue to adapt the `DbRequestBuilder` and `Statements` APIs to something that is easier to handle. They should move to one of those modules when we sort out the imports. I didn't want to touch those modules in this PR too, it's unwieldy enough as it is :-)

monacoremo:
Many thanks for your thorough review and helpful comments @wolfgangwalther !

I added a few commits to address your notes.

monacoremo:
The freebsd builds fails with

```
cabal v2-update
ld-elf.so.1: /usr/local/bin/cabal: Undefined symbol "pthread_setname_np@FBSD_1.6"
```

No idea what change would have caused that. Could that be some caching issue @steve-chavez ?

steve-chavez:
@monacoremo I'm also getting the same error. It could be related to caching because we cache freebsd packages in:

https://github.com/PostgREST/postgrest/blob/0ddd676ef04a72d89c0fa734aeca0596841874e4/.cirrus.yml#L7-L13

I'll try removing this cache on https://github.com/PostgREST/postgrest/pull/1760.

(Seems related to https://reviews.freebsd.org/D25117)

wolfgangwalther (approved):

steve-chavez (approved):

monacoremo:
Thank you @wolfgangwalther and @steve-chavez reviewing and helping to get this PR done! I'll merge this now so that we can push ahead on other topics, even though we'll get a CI failure for the unrelated FreeBSD issue.

steve-chavez:
> I'll try removing this cache on #1760.

Bummer, this didn't work:

- https://github.com/PostgREST/postgrest/pull/1760/commits/28bac0c009fe9999b7ac28659bef390dbcfefe4d
- https://cirrus-ci.com/task/4544153243615232

It must be something else.

Perhaps we can try busting the cache with `echo $CIRRUS_OS` like done here:
https://cirrus-ci.org/guide/writing-tasks/#cache-instruction

I'll refrain from doing this on #1760 though, it already got a bit long.

Issue #1562, opened 2020-07-20
In-Database Configuration


wolfgangwalther:
This is a suggestion for in-database configuration, based on ideas in #1429 and https://github.com/PostgREST/postgrest/pull/1559#issuecomment-659597014.

Currently, all configuration is done file-based in `postgrest.conf`. After the recent addition of #1544, reloading the config on the fly is now possible. Not sure whether this is already possible with `NOTIFY` as mentioned in #1542. As mentioned in #1544 this reloads only part of the config, while other parts (like database connection info) need a restart of the server.

The downside of the current implementation is, that we don't have a single source of truth for everything set in the config file, that e.g. references database objects. Stuff like:

- `db-schema` needs to be adjusted together with the schema created/dropped (multi-tenancy #1521)

- `pre-request` needs to be set to the corresponding function

- `jwt-secret` is often copied as `app.settings.jwt-secret` or something like that

It would be much nicer, if we could (optionally) have the whole reloadable configuration in the database as well.

# Using local, session and database variables

We are already using local variables (`SET LOCAL`) to set some values per transaction (`request.*`, `response.*`, `app.settings.*`). We can extend this to use session and database variables as well and add all the reloadable config options as follows:

- Use `ALTER DATABASE xyz SET pgrst.db_schema='public';` to set config options, just like in the current `postgrest.conf`.

- For backwards compatibility: Keep a copy of the config file `postgrest.conf` in memory and add the reloadable parts to each session with `SET SESSION`. Those will overwrite the database variables automatically. (This will also support use-cases where `ALTER DATABASE` permissions are not available).

- Query all those config variables when updating the schema cache. Since the `SET SESSION` will have been executed for this session as well, this will result in the currently active config, with both in-database and file-based merged together. Use this config from the schema cache for operation.

- For each request, add the `request.*` and `response.*` variables as `SET LOCAL` like we do already.

This would allow us to:

- use only one variable for the jwt secret, instead of two

- do stuff like this:
```SQL
CREATE SCHEMA new_customer;
ALTER DATABASE mydb SET pgrst.db_schema='...,new_customer';
NOTIFY pgrst;
```

_Note: When changing database variables, those are only in effect for new sessions. That means, we would need to recreate all existing sessions when reloading the schema cache._

# Extensibility

[...]

PR #1729, opened 2021-01-10
feat: Add db-settings-prefix config 

steve-chavez:
### Feature

For https://github.com/PostgREST/postgrest/issues/1562. Allows configuring postgrest from the db by setting config options on the connection role. For example:

```sql
-- postgrest.conf having:
-- db-settings-prefix = "instance"

ALTER ROLE postgrest_test_authenticator SET pgrst.instance.jwt-secret = "REALLYREALLYREALLYREALLYVERYSAFE"
```

That will set the `jwt-secret` config option accordingly.

SUPERUSER privileges are required for ALTERing role settings(see https://github.com/PostgREST/postgrest/issues/1562#issuecomment-743568080), so this might not work on some cloud-managed databases.

Config options can only be set on `ALTER ROLE .. SET ..` and not `ALTER DATABASE .. SET`. Though the same privileges are required(SUPERUSER), ROLE level settings are better because they're scoped to our `authenticator` role.


main/Main.hs
```diff
@@ -116,6 +105,24 @@ main = do
   -- Config that can change at runtime
   refConf <- newIORef conf
 
+  -- re-read and override the config with db settings if db-settings-prefix is provided
+  when (isJust dbSettingsPrefix) $
+    reReadConfig pool dbSettingsPrefix env (cliPath opts) refConf
```
steve-chavez, line 110: Didn't see a better choice than to re-read here. The db-uri must be obtained before loading the settings from the db, then the config can be overriden with the db settings.

src/PostgREST/App.hs
```diff
@@ -84,23 +84,48 @@ postgrest logLev refConf refDbStructure pool getTime connWorker =
...
         -- Launch the connWorker when the connection is down. The postgrest function can respond successfully(with a stale schema cache) before the connWorker is done.
         when (responseStatus response == status503) connWorker
         respond response
 
+optionalRollback
+  :: AppConfig
+  -> ApiRequest
+  -> HT.Transaction Response
+  -> HT.Transaction Response
+optionalRollback conf apiRequest tx =
```
steve-chavez, line 101: I refactored the `configDbTxEnd` on top of the `optionalRollback` function from https://github.com/PostgREST/postgrest/pull/1725. I updated it with the TxEnd constructors.

I could remove this commit and leave it for a later PR..

steve-chavez, line 101: I refactored the `configDbTxEnd` on top of the `optionalRollback` function from https://github.com/PostgREST/postgrest/pull/1725. I updated it with the TxEnd constructors.

I think I should remove this commit and leave it for a later PR.. might be better.

steve-chavez:
One thing that is missing is a test for reloading the config with the db settings. But it doesn't make much sense to reload with a SIGUSR2 signal. I think we should extend our NOTIFY channel to allow reloading the config from there.

So, how about:

```sql
-- These two for reloading the schema cache 
NOTIFY pgrst;
NOTIFY pgrst, 'SIGUSR1';

-- This one for reloading the config
NOTIFY pgrst, 'SIGUSR2';
```

Thoughts?

src/PostgREST/Statements.hs
```diff
@@ -190,3 +194,19 @@ decodeGucHeaders = first (const GucHeadersError) . JSON.eitherDecode . toS <$> H
...
+dbSettingsStatement :: Bool -> H.Statement Text [(Text, Text)]
+dbSettingsStatement = H.Statement sql (param HE.text) decodeSettings
+  where
+    sql = [q|
+      with
+      role_setting as (
+        select unnest(setconfig) as setting from pg_catalog.pg_db_role_setting where setrole = current_user::regrole::oid
+      ),
+      pgrst_settings as (
+        select replace(setting, 'pgrst.' || $1 || '.', '') as setting from role_setting where setting like 'pgrst.' || $1 || '%'
```
wolfgangwalther, line 208:
```suggestion
        select replace(setting, 'pgrst.' || $1 || '.', '') as setting from role_setting where setting like 'pgrst.' || $1 || '.%'
```

test/io-tests/test_io.py
```diff
@@ -209,6 +209,22 @@ def test_expected_config_from_environment():
     assert dumpconfig(env=env) == expected
 
 
+def test_expected_config_from_db_settings(defaultenv):
+    "Config should be overriden from database settings"
+
+    env = defaultenv
+    env["PGRST_DB_SETTINGS_PREFIX"] = "instance"
+    expected = (CONFIGSDIR / "db-settings-prefix.config").read_text()
```
wolfgangwalther, line 217: This file would better be in the `expected` subdir. I suggest you just re-use one of the files in this directory. The `no-defaults.config` would fit, because it's exactly that. If you change the fixtures to those values, you can use it.

src/PostgREST/App.hs
wolfgangwalther: I agree - this would make things easier in terms of merge conflicts.

main/Main.hs
```diff
@@ -143,7 +150,8 @@ main = do
   -- Re-read the config on SIGUSR2, but only if we have a config file
```
wolfgangwalther, line 150: Now we would need to change that condition, right? If we start up without a config file, but with `PGRST_DB_URI`, do we still want to trigger a reload from the database?

main/Main.hs
wolfgangwalther: I don't see another way right now either.

test/io-tests/test_io.py
wolfgangwalther: Or, alternatively, if that's easier, move the file you have to `expected`. You might change some condition for one of the others tests that loads it's fixtures from the list of files in `expected`, but that should be straight forward.

wolfgangwalther:
> One thing that is missing is a test for reloading the config with the db settings. But it doesn't make much sense to reload with a SIGUSR2 signal. I think we should extend our NOTIFY channel to allow reloading the config from there.
> 
> So, how about:
> 
> ```sql
> -- These two for reloading the schema cache 
> NOTIFY pgrst;
> NOTIFY pgrst, 'SIGUSR1';
> 
> -- This one for reloading the config
> NOTIFY pgrst, 'SIGUSR2';
> ```
> 
> Thoughts?

I think that was suggested / discussed somewhere else already, right?

I am not sure what is currently implemented already and whether that would be any breaking change, but how about:
- `NOTIFY pgrst` - reloads both schema cache and config
- `NOTIFY pgrst, 'reload-schema'`
- `NOTIFY pgrst, 'reload-config'`

While we can use the signals for that too, I think it'd be easier to have some more explicit strings. We're not really sending those signals. And a string like this is easier to extend in the future.

wolfgangwalther:
> ### Feature
> 
> For #1562. Allows configuring postgrest from the db by setting config options on the connection role. For example:
> 
> ```sql
> -- postgrest.conf having:
> -- db-settings-prefix = "instance"
> 
> ALTER ROLE postgrest_test_authenticator SET pgrst.instance.jwt-secret = "REALLYREALLYREALLYREALLYVERYSAFE"
> ```
> 
> That will set the `jwt-secret` config option accordingly.

Awesome! :)

> Config options can only be set on `ALTER ROLE .. SET ..` and not `ALTER DATABASE .. SET`. Though the same privileges are required(SUPERUSER), ROLE level settings are better because they're scoped to our `authenticator` role.

I think the SQL query could easily be extended to support `ALTER DATABASE` too, right? Is this a deliberate decision to not support that or more like "we can do that later"?

wolfgangwalther:
A couple of config options to think about:
- `db-anon-role`. Is currently required on the first load already, so it's not possible to set this only through the database. We should make it optional and block anonymous access without it as discussed elsewhere. This would allow startup without it.
- `db-schemas`. Can we make this optional, too, to allow setting it on the database level? I suggest we just respond with an error to every request when we don't have a valid `db-schemas` setting.
- `server-*` + `db-uri` + `db-pool*` - can we add a deny-list for those settings, so that they are not loaded from the database? Otherwise a `--dump-config` would really be misleading, because it would show overridden settings, that are not in effect.

src/PostgREST/App.hs
wolfgangwalther: I suggest to split into 3 parts:
- refactor of tx end -> after #1725
- the two fixes "Correct ..." to merge right away
- the main commit to work on in this PR

nerfpops:
Just tested in a cloud hosted Postgres on Aiven, and as you've already discussed, it doesn't work unfortunately. I hope you find a way at some point to allow non-superusers to use the db as config. I think it would make it easier to do jwt key rotation by simply updating a config in the db.

steve-chavez:
> I think the SQL query could easily be extended to support ALTER DATABASE too, right? Is this a deliberate decision to not support that or more like "we can do that later"?

@wolfgangwalther Deliberate decision, and now it's on point with what LorenzHenk mentioned on https://github.com/PostgREST/postgrest/issues/1179#issuecomment-758533822:

> uninstalling PostgREST will still leave its trails in the DDL of views, which is harder to track down than e.g. having to delete 1 table & 1 view containing all the metadata

If we keep this isolated to the authenticator role, uninstalling postgrest(not that we want that) would be matter of dropping the roles + schema. I also feel that this would look better on the docs - doing an ALTER DATABASE looks more invasive(though requiring the same privileges unfortunately).

Also, the ROLE approach it's more flexible, we could even remove the `db-settings-prefix` if we tell the user to create different ROLEs for different postgrest instances on the same db. I've noted it's a bit inconvenient to set a prefix when you only have a single postgrest instance. 

So that's why I think we should restrict to ROLEs only.

> I am not sure what is currently implemented already and whether that would be any breaking change, but how about:
NOTIFY pgrst - reloads both schema cache and config
NOTIFY pgrst, 'reload schema'
NOTIFY pgrst, 'reload config'
While we can use the signals for that too, I think it'd be easier to have some more explicit strings. We're not really sending those signals. And a string like this is easier to extend in the future.

I like it! The strings look much better. I'll add that.

> server-* + db-uri + db-pool* - can we add a deny-list for those settings, so that they are not loaded from the database? Otherwise a --dump-config would really be misleading, because it would show overridden settings, that are not in effect.

Totally agree. I'll deny changing those settings.

> db-anon-role. Is currently required on the first load already, so it's not possible to set this only through the database.

I think that one deserves it's own PR, it's a breaking change after all - there would be no panic when it's missing.

> db-schemas. Can we make this optional, too, to allow setting it on the database level? I suggest we just respond with an error to every request when we don't have a valid db-schemas setting.

Same as above. It's also a bit weird to enable running postgrest when it's in a "broken state". Would need more discussion. 

WDYT?

steve-chavez:
@nerfpops Thanks for reporting, too bad about Aiven. Later on we could create a postgrest extension to [register the GUCs](https://github.com/PostgREST/postgrest/issues/1562#issuecomment-743724921) so no SUPERUSER is needed. Or we could just wait for SCHEMA VARIABLES to land on postgres as mentioned by Wolfgang on https://github.com/PostgREST/postgrest/issues/1562#issue-661980151.

wolfgangwalther:
> If we keep this isolated to the authenticator role, uninstalling postgrest(not that we want that) would be matter of dropping the roles + schema. I also feel that this would look better on the docs - doing an ALTER DATABASE looks more invasive(though requiring the same privileges unfortunately).

I'm not really convinced by this argument...

> Also, the ROLE approach it's more flexible, we could even remove the `db-settings-prefix` if we tell the user to create different ROLEs for different postgrest instances on the same db. I've noted it's a bit inconvenient to set a prefix when you only have a single postgrest instance.

... but very much but this one! Even the current implementation doesn't need the prefix at all, you're right. Using different `ROLE`s is all we need. Let's get rid of it right away.
 
> So that's why I think we should restrict to ROLEs only.

I can kind of see use-cases where it would be easier to have the database option, too. But since that's easy to extend, I will come back to this, once I should really hit one of those cases. So see you in a couple of years... - ah wait, we'll have schema variables by then... :o)

> > db-anon-role. Is currently required on the first load already, so it's not possible to set this only through the database.
> 
> I think that one deserves it's own PR, it's a breaking change after all - there would be no panic when it's missing.
> 
> > db-schemas. Can we make this optional, too, to allow setting it on the database level? I suggest we just respond with an error to every request when we don't have a valid db-schemas setting.
> 
> Same as above. [...]
> 
> WDYT?

Yes, absolutely fine to put those into different PRs. Just wanted to bring it up, because those are related here, too.

> It's also a bit weird to enable running postgrest when it's in a "broken state". Would need more discussion.

I think the "broken, but defined state" is better then any undefined state. Assume you're just setting up PostgreSQL and PostgREST. Most setups would probably restart PostgREST once it errors out. Assume you're initializing the database and are **not** creating all the roles and objects etc. in one transaction. PostgREST would kind of magically start serving the schema once the config variable is set - even though the schema might have any number of objects so far (empty, some created, all created - randomly). It'd be much better if, after setting everything up, you just `NOTIFY` PostgREST to reload. Before reload we will serve an error message. After reload we will serve the full schema. But nothing in parts only (undefined). A lot less surprising. Will also teach people a bit earlier about the need and how to reload the schema.

wolfgangwalther:
> > Also, the ROLE approach it's more flexible, we could even remove the `db-settings-prefix` if we tell the user to create different ROLEs for different postgrest instances on the same db. I've noted it's a bit inconvenient to set a prefix when you only have a single postgrest instance.
> 
> ... but very much but this one! Even the current implementation doesn't need the prefix at all, you're right. Using different `ROLE`s is all we need. Let's get rid of it right away.

Do we still have a config option for this? Or do we just load config options from database all the time? Not sure if there's any downside to doing so - I don't see any right now.

steve-chavez:
> Do we still have a config option for this? Or do we just load config options from database all the time? Not sure if there's any downside to doing so - I don't see any right now.

I think not! I would also like removing the prefix config. I'll check it out.

src/PostgREST/App.hs
steve-chavez: Done! I put the refactor on this branch https://github.com/steve-chavez/postgrest/tree/tx-end-refactor for later

test/io-tests/test_io.py
steve-chavez: Done! I've put the config in expected.

steve-chavez:
Actually, one downside to load the db settings always is that it's wasteful for users that don't have superuser over their db(most of our users maybe). The query is simple, but still.

Another one is that it slows down the io-tests, I'm getting a `test_io.PostgrestTimedOut` on all the tests now.

So, how about having this config:

```bash
db-settings = true
```

@wolfgangwalther WDYT?

wolfgangwalther:
Hm. I'm not too concerned about performance, because it's only on reload, not while serving requests. In the context of reloading the overhead should be rather small. IO test performance is a thing however.

So if we can make the default `db-settings = true`, then we could still change it to `false` for the io tests. Users that care for reload performance can do that, too.

Enabling by default seems appealing to me, because the whole point of putting the config in the database is, that I want to have the least amount of config needed in my config file / docker-compose file.

So instead of having this:

```
  environment:
    PGRST_DB_URI: ...
    PGRST_DB_SETTINGS: "true"
```

it would be

```
  environment:
    PGRST_DB_URI: ...
```

It's not much, but it "just works".

Does that make sense?

---

Thinking about the name of the config option. Currently we are only using the term "settings" in the context of `app.settings...`. In the docs we call everything else "configuration" and "configuration parameters". Two of the three boolean config options we currently have, do have some "boolean" indicator, too: "jwt-secret-**is**-base64" and "db-channel-**enabled**".

Given that, I suggest we call it `db-load-guc-config`. We could also rename `db-prepared-statements` to `db-use-prepared-statements` and `db-channel-enabled` to `db-enable-channel` for consistency. I think both of those are currently unreleased, so that would not be a breaking change, yet.

WDYT?

test/fixtures/roles.sql
```diff
@@ -20,4 +20,17 @@ ALTER ROLE postgrest_test_authenticator SET pgrst.instance."db-prepared-statemen
 ALTER ROLE postgrest_test_authenticator SET pgrst.instance."db-pre-request" = 'prereq_func';
 ALTER ROLE postgrest_test_authenticator SET pgrst.instance."db-max-rows" = '1000';
 ALTER ROLE postgrest_test_authenticator SET pgrst.instance."db-extra-search-path" = 'public, extensions';
-ALTER ROLE postgrest_test_authenticator SET pgrst.instance."db-anon-role" = 'postgrest_test_author';
+
+-- non-reloadable configs for io tests
+ALTER ROLE postgrest_test_authenticator SET pgrst.instance."server-host" = 'modified';
```
wolfgangwalther, line 25: I suggest to change the `modified` string to `ignored` in all those cases, to make it clear that this value should never be taken.

test/io-tests/configs/db-settings-prefix.config
```diff
@@ -1,22 +1,23 @@
-db-anon-role = "postgrest_test_author"
+db-uri = "required"
+db-anon-role = "postgrest_test_anonymous"
 db-channel = "pgrst"
 db-channel-enabled = false
-db-extra-search-path = "public,extensions"
+db-extra-search-path = "public"
 db-max-rows = 1000
```
wolfgangwalther, line 6: I think it would be valuable to set all values in this config to a non-default value to make sure that the final value that is dumped does really result from the config and not the default fallback.

Given that, I think you should be able to re-use the `no-defaults.config` file here - this does exactly that. If we use `true` as the default for `db-settings` (or whatever name we choose), the `no-defaults` would then have false. You can then override this in python via `PGRST_DB_SETTINGS` env variable.

So basically:
- with `no-defaults.config` and `db-settings=false` the expected output is `expected/no-defaults.config`
- with `no-defaults.config` and `db-settings=true` the expected output is `expected/db-settings.config`

Makes sense?

test/io-tests/test_io.py
```diff
@@ -212,13 +218,18 @@ def test_expected_config_from_environment():
 def test_expected_config_from_db_settings(defaultenv):
     "Config should be overriden from database settings"
 
-    env = defaultenv
-    env["PGRST_DB_SETTINGS_PREFIX"] = "instance"
-    expected = (CONFIGSDIR / "db-settings-prefix.config").read_text()
+    expected = (CONFIGSDIR / "expected" / "db-settings-prefix.config").read_text()
+    config = CONFIGSDIR / "db-settings-prefix.config"
 
     # remove db-uri from the config file since it varies
```
wolfgangwalther, line 224: WDYT about adding the actual db-uri value to the expected output instead? Maybe you can just add a placeholder in the expected file like `<REPLACED_WITH_DB_URI>` and then use a simple replace?

That would make the test code simpler - and would test the "db-uri can't be overwritten from the database", too.

steve-chavez: > would test the "db-uri can't be overwritten from the database", too.

Hmm.. I don't understand how can this be really proven if I have to replace the expected value.

The `db-uri` I get varies like:

```
db-uri = "postgresql:///postgres?host=/run/user/1000/tmp.MarQgcKD7r/socket&user=postgrest_test_authenticator"
db-uri = "postgresql:///postgres?host=/run/user/1000/tmp.hyIrX2yB1H/socket&user=postgrest_test_authenticator"
db-uri = "postgresql:///postgres?host=/run/user/1000/tmp.YSNO7AmIjZ/socket&user=postgrest_test_authenticator"
```

What I've done instead is to add an additional assert for failing if `db-uri = "postgresql://ignored"`(value from the guc). Which does fail when removing `db-uri` from:

```haskell
          if key `notElem` [
            "server-host", "server-port", "server-unix-socket", "server-unix-socket-mode", "log-level",
            "db-anon-role", "db-uri", "db-channel-enabled", "db-channel", "db-pool", "db-pool-timeout", "db-load-guc-config"]
          then lookup key dbSettings
          else Nothing
```

Additiona

steve-chavez:
> Given that, I suggest we call it db-load-guc-config. 

I like that :+1:, I've renamed the config and made it True by default.

src/PostgREST/Statements.hs
steve-chavez: Outdated now since prefix is no more.

test/io-tests/test_io.py
wolfgangwalther: Values are first read from the config file (1), then from the environment (2) and then from the database (3), right?

In the IO tests, `db-uri` is read from (2).

Now, we want to test that there is **no** (3) for `db-uri`, because it should not be possible to do that.

We do the following to test that:
- We set `db-uri` to `ignored` in the database
- We assert that the dumped value is still the same value that was provided by the environment variable (2)

This assert is done by comparing the dump vs the expected output. Because `db-uri` is not a static value, but dynamically created, I suggested we just adjust the expected output dynamically by loading the file and then replacing the actual value for `db-uri`.

This should show the same behaviour that you are describing in terms of removing `db-uri` from that list.

wolfgangwalther: Or to give a concrete example of what I meant:

```python
expected = expected.replace('<REPLACED_WITH_DB_URI>', defaultenv['PGRST_DB_URI'])
```

If you do that, you only need to assert `dumped == expected` - nothing else.

steve-chavez: Ahh.. makes sense now!

Thanks for elaborating. I've corrected the test.

test/io-tests/configs/db-settings-prefix.config
steve-chavez: Got it. I'm now using the `no-defaults.config` for the test.

test/io-tests/test_io.py
```diff
@@ -182,7 +183,13 @@ def jwtauthheader(claim, secret):
 
 
 @pytest.mark.parametrize(
-    "expectedconfig", (CONFIGSDIR / "expected").iterdir(), ids=attrgetter("name")
+    "expectedconfig",
+    [
+        x
+        for x in (CONFIGSDIR / "expected").iterdir()
+        if "load-db-guc" not in str(x)
```
wolfgangwalther, line 190:
```suggestion
        expectedconfig
        for expectedconfig in (CONFIGSDIR / "expected").iterdir()
        if (CONFIGSDIR / expectedconfig.name).exists()
```

test/io-tests/test_io.py
```diff
@@ -209,6 +216,23 @@ def test_expected_config_from_environment():
...
+def test_expected_config_from_db_settings(defaultenv):
+    "Config should be overriden from database settings"
+
+    config = CONFIGSDIR / "no-defaults.config"
+    env = {
+        **defaultenv,
+        "PGRST_DB_LOAD_GUC_CONFIG": "true",
+    }
+    expected = (
+        (CONFIGSDIR / "expected" / "load-db-guc.config")
```
wolfgangwalther, line 228:
```suggestion
        (CONFIGSDIR / "expected" / "no-defaults-with-db.config")
```

(would need to change the filename, too, of course)

test/io-tests/configs/defaults.config
```diff
@@ -1,3 +1,4 @@
 db-uri = "required"
 db-schemas = "required"
 db-anon-role = "required"
+db-load-guc-config = false
```
wolfgangwalther, line 4:
```suggestion
# Not the default, but only works with proper db-uri
db-load-guc-config = false
```

test/fixtures/roles.sql
```diff
@@ -5,3 +5,31 @@ CREATE ROLE postgrest_test_default_role;
 CREATE ROLE postgrest_test_author;
 
 GRANT postgrest_test_anonymous, postgrest_test_default_role, postgrest_test_author TO :USER;
+
+-- reloadable config options for io tests
+ALTER ROLE postgrest_test_authenticator SET pgrst."jwt-aud" = 'https://example.org';
```
wolfgangwalther, line 10:
```suggestion
ALTER ROLE postgrest_test_authenticator SET pgrst."raw-media-types" = 'application/vnd.pgrst.db-config';
```

(jwt-aud is added twice, but raw-media-types missing here)

test/io-tests/configs/expected/load-db-guc.config
```diff
@@ -0,0 +1,27 @@
...
+db-load-guc-config = "true"
+db-tx-end = "commit-allow-override"
+db-uri = "<REPLACED_WITH_DB_URI>"
+jwt-aud = "https://example.org"
+jwt-role-claim-key = ".\"a\".\"role\""
+jwt-secret = "REALLYREALLYREALLYREALLYVERYSAFE"
+jwt-secret-is-base64 = true
+log-level = "info"
+openapi-server-proxy-uri = "https://example.org/api"
+raw-media-types = "application/vnd.pgrst.config"
```
wolfgangwalther, line 21:
```suggestion
raw-media-types = "application/vnd.pgrst.db-config"
```

test/fixtures/roles.sql
```diff
@@ -5,3 +5,31 @@ CREATE ROLE postgrest_test_default_role;
...
+ALTER ROLE postgrest_test_authenticator SET pgrst."server-port" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."server-unix-socket" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."server-unix-socket-mode" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."log-level" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-anon-role" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-uri" = 'postgresql://ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-channel-enabled" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-channel" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-pool" = 'ignored';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-pool-timeout" = 'ignored';
```
wolfgangwalther, line 35:
```suggestion
ALTER ROLE postgrest_test_authenticator SET pgrst."db-pool-timeout" = 'ignored';
ALTER ROLE postgrest_test_authenticator SET pgrst."db-load-guc-config" = 'ignored';
```

for completeness

wolfgangwalther:
Looking good. Suggested just a few more minor changes.

> One thing that is missing is a test for reloading the config with the db settings. But it doesn't make much sense to reload with a SIGUSR2 signal. I think we should extend our NOTIFY channel to allow reloading the config from there.

Are you going to do that in this PR? If yes, cool. If no, I suggest to still add a test for now that tests the reloading with a `SIGUSR2` signal. This can then easily be switched to a `NOTIFY`, but we should have test coverage good immediately. This is also related to my still open comment above about the condition for reload, that needs to be changed to make test-case like this without config file pass.

wolfgangwalther:
One other note: I wonder why codecov doesn't add any status checks here. The upload part says this:

```
View reports at https://codecov.io/github/steve-chavez/postgrest/commit/684abc0461948f8d60541346dc4229ab36ba70d3
```

@steve-chavez have you enabled codecov for you fork, too? Maybe that's causing this.

It could also be circleci, that's enabled for your fork. I noticed that when you currently open a PR for the first time CircleCI has already run the pipeline in your fork and will not report anything in the PR until you trigger a rebuild somehow. I had the same thing while I had Circle enabled in my own fork. Once I disabled it, it behaved as expected again.

steve-chavez:
>  I noticed that when you currently open a PR for the first time CircleCI has already run the pipeline in your fork and will not report anything in the PR until you trigger a rebuild somehow.

Triggered a rebuild but it didn't seem to work.

> It could also be circleci, that's enabled for your fork. 
>  Once I disabled it, it behaved as expected again.

Yes, I do have it enabled in my fork. It's helpful for running tests before making a PR though, so I would prefer not to disable it.

> have you enabled codecov for you fork, too? Maybe that's causing this.

That one I could disable. I'll try it out.

wolfgangwalther:
> Yes, I do have it enabled in my fork. It's helpful for running tests before making a PR though, so I would prefer not to disable it.

Yeah. I digged further and it seems like disabling it for the fork is in fact the only way to "solve" this. Circle not showing up when I first open the PR was annoying enough for me to do that. I'm totally fine with you keeping it enabled, though. It will eventually show up once anything changes in the PR or if we manually trigger a rebuild.

> > have you enabled codecov for you fork, too? Maybe that's causing this.
> 
> That one I could disable. I'll try it out.

I assume it will not work. This is because we are using the tokenless-upload. Therefore the codecov upload script has to take the repo info from what it gets from CircleCI - and this is running with your fork's repo name. In this case you could try to add the postgrest repo's codecov token to your fork in CircleCI. This way, the upload script should pick up this token when running under your CircleCI user and still upload to the main repo.

wolfgangwalther:
I guess that's just what I assumed:

```
==> Uploading reports
    url: https://codecov.io
    query: branch=in-db-config&commit=684abc0461948f8d60541346dc4229ab36ba70d3&build=3877&build_url=&name=&tag=&slug=steve-chavez%2Fpostgrest&service=circleci&flags=&pr=1729&job=0&cmd_args=f
```

That `slug=steve-chavez/postgrest` is probably taken from one of the circle ci env vars.

wolfgangwalther:
Compare this to the same from #1724. I don't have circle enabled for my fork and it shows this:

```
==> Uploading reports
    url: https://codecov.io
    query: branch=pull%2F1724&commit=4f3aea70a9bb9efdde8ff09af75a72ff93c25329&build=7966&build_url=&name=&tag=&slug=PostgREST%2Fpostgrest&service=circleci&flags=&pr=1724&job=0&cmd_args=f
```

I hope setting up the codecov token in your circleci user will work.

steve-chavez:
> I hope setting up the codecov token in your circleci user will work.

Let's see. I've just copied the CODECOV_TOKEN from https://app.codecov.io/gh/PostgREST/postgrest/settings to my postgrest project env vars on CircleCI.

wolfgangwalther:
> **Edit**: Worked!

Sweet!

`codecov/patch` is currently not hitting the target. This seems to be mainly about parts of the refactor and `--dump-schema`. I'm pretty certain that most of those lines will be covered once you rebase on current main, where I merged the new io tests. The two remaining uncovered lines will be two of the few cases where we need the coverage overlay later on - we can't test the error branches in the settings and dump-schema queries.

test/io-tests/configs/expected/load-db-guc.config
steve-chavez: This test allowed me to fix a bug in the settings query because I was only getting a `"application/vnd.db-config"`. The `pgrst.` stripping from the query was also applied to values :sweat_smile: .

test/io-tests/configs/defaults.config
steve-chavez: The tests fail by adding this line, I'd need to strip the comment for going green. Do you think it's worth it?

wolfgangwalther: How so? I made the change in the config file - not the expected output file. I would expect the output to stay the same?

steve-chavez: Ah, my bad. You're right. I'll add the change.

steve-chavez:
Tried adding a test for SIGUSR2 reload but hit this [nasty pg bug](https://gist.github.com/steve-chavez/e2e4da63afdabecb320c2a094d2ef70c) again. So doing a:

```sql
create or replace function change_max_rows_config(val int) returns void as $$
  alter role postgrest_test_authenticator set pgrst."db-max-rows" = val;
$$ language sql security definer;
```

Won't work because `max-rows` is always set to `val`.

Fortunately `pg_db_role_setting` is UPDATABLE so we can still modify it with:

```sql
update pg_db_role_setting
set setconfig = setconfig || format('pgrst.db-max-rows=%s', 5)
where setrole = 'postgrest_test_authenticator'::regrole::oid
```

The problem with that is that settings can now be duplicated. That didn't happen with the `ALTER ROLE SET` way.

So we'd need to build a query that ensures the array doesn't have duplicates.

Bummer, kinda takes away the simplicity of the solution :/.

wolfgangwalther:
> Tried adding a test for SIGUSR2 reload but hit this [nasty pg bug](https://gist.github.com/steve-chavez/e2e4da63afdabecb320c2a094d2ef70c) again.
> [,,,]
> Won't work because `max-rows` is always set to `val`.

I don't think it's a bug. Are you looking for dynamic SQL like this?

```sql
create or replace function change_max_rows_config(val int) returns void
security definer language plpgsql as $$
  begin
    execute format($f$
      alter role postgrest_test_authenticator set pgrst."db-max-rows" = %L;
    $f$, val);
  end
$$;
```


> Fortunately `pg_db_role_setting` is UPDATABLE so we can still modify it with:
> [...]
> The problem with that is that settings can now be duplicated. That didn't happen with the `ALTER ROLE SET` way.
> So we'd need to build a query that ensures the array doesn't have duplicates.

I don't think we should go that way - modifying those tables directly is discouraged by Postgres.

main/Main.hs
steve-chavez: >  If we start up without a config file, but with PGRST_DB_URI, do we still want to trigger a reload from the database?

We also need the `PGRST_DB_ANON` and `PGRST_DB_SCHEMAS` for starting, but yes - supporting it is good for completeness. Though the common use case would be to reload with NOTIFY.

steve-chavez:
> Are you looking for dynamic SQL like this?

@wolfgangwalther Awesome! You're a lifesaver, I didn't think of going the dynamic SQL route. 

> Are you going to do that in this PR? If no, I suggest to still add a test for now that tests the reloading with a SIGUSR2 signal. This can then easily be switched to a NOTIFY.

I've added the SIGUSR2 test. I'm going to do NOTIFY next, but I'll add that on another PR.

test/io-tests/test_io.py
```diff
@@ -87,6 +87,8 @@ def defaultenv():
         "PGRST_DB_URI": os.environ["PGRST_DB_URI"],
         "PGRST_DB_SCHEMAS": os.environ["PGRST_DB_SCHEMAS"],
         "PGRST_DB_ANON_ROLE": os.environ["PGRST_DB_ANON_ROLE"],
+        "PGRST_DB_LOAD_GUC_CONFIG": "false",
+        "PGRST_LOG_LEVEL": "warn",
```
steve-chavez, line 91: I put a non-existent `root-spec` function in the role settings config. The `wait_until_ready` function then failed, but the logs only showed a couple of 503 and not the 404s. I changed to warn to show this in case a similar error happens.

steve-chavez: This one broke another test. I've just reverted it.

steve-chavez:
> I'm going to do NOTIFY next, but I'll add that on another PR.

@wolfgangwalther Changed my mind. I've added `NOTIFY 'reload config'/'reload schema'` on this PR.

test/io-tests/test_io.py
wolfgangwalther: Yes, that confirms my gut feeling, when I first read it. We'll have to add a few tests for different log level somewhere else anyway.

main/Main.hs
```diff
@@ -310,32 +308,27 @@ loadSchemaCache pool actualPgVersion refConf refDbStructure = do
...
     start = do
       connStatus <- takeMVar mvarConnectionStatus -- takeMVar makes the thread wait if the MVar is empty(until there's a connection).
       case connStatus of
         Connected actualPgVersion -> void $ forkFinally (do -- forkFinally allows to detect if the thread dies
           dbOrError <- C.acquire dbUri
-          -- Debounce in case too many NOTIFYs arrive. Could happen on a migration(assuming a pg EVENT TRIGGER is set up).
-          -- This might not be needed according to pg docs https://www.postgresql.org/docs/12/sql-notify.html:
-          -- "If the same channel name is signaled multiple times from the same transaction with identical payload strings, the database server can decide to deliver a single notification only."
-          -- But we do it to be extra safe.
-          scFiller <- mkDebounce (defaultDebounceSettings {
```
steve-chavez, line 310: I removed the debounce, it isn't necessary because pg does it. Having it makes the functionality and the tests slower - must wait at least 1 second for the reloads to occur.

test/fixtures/roles.sql
```diff
@@ -17,7 +17,7 @@ ALTER ROLE postgrest_test_authenticator SET pgrst."db-tx-end" = 'commit-allow-ov
 ALTER ROLE postgrest_test_authenticator SET pgrst."db-schemas" = 'test, tenant1, tenant2';
 ALTER ROLE postgrest_test_authenticator SET pgrst."db-root-spec" = 'root';
 ALTER ROLE postgrest_test_authenticator SET pgrst."db-prepared-statements" = 'false';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-pre-request" = 'custom_headers';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db-pre-request" = 'test.custom_headers';
```
steve-chavez, line 20: Had to add the schema qualifier because when switching to the v1 schema there's no `custom_headers` func, which have errors.

Btw, thanks to this I noted here that the `root-spec` doesn't support a schema qualifier. Which looks necessary when having multiple schemas.

steve-chavez:
I've squashed some commits to clean up history. I think this one is good to merge.

test/fixtures/roles.sql
wolfgangwalther: Are you tracking the missing schema qualifier for root-spec somewhere?

main/Main.hs
```diff
@@ -64,11 +61,11 @@ main = do
   -- read PGRST_ env variables
   env <- readEnvironment
 
-  -- read path from commad line
-  opts <- readCLIShowHelp env
+  -- read command/path from commad line
+  (command, confPath) <- ((,) <$> cliCommand <*> cliPath) <$> readCLIShowHelp env
```
wolfgangwalther, line 65: I like the idea - but maybe it's possible to use `RecordWildCards` here, so something like this?

```haskell
CLI{..} <- readCLIShowHelp env
```

or `NamedFieldPuns` if you want to be more explicit?

```haskell
CLI{cliCommand, cliPath} <- readCLIShowHelp env
```

In both cases you should be able to use `cliCommand` and `cliPath` directly without the `opts` afterwards.

wolfgangwalther (approved):

wolfgangwalther:
Great work @steve-chavez. There's only one question remaining for me... merging this - can we still call the next release the **performance** release? Or should we call it the **configuration** release? Environment variables, database settings... awesome stuff! :)

steve-chavez:
> can we still call the next release the performance release? Or should we call it the configuration release?

@wolfgangwalther Ah, and there's also more **security** because of the switch from "set locals" to prepared statements. Perhaps we can call it something like **enterprise-grade release**? Not sure for now :)

steve-chavez:
Btw, code coverage is amazing(first time user here). I see we have a badge already on the repo. But maybe we could also list it as a feature on the release since it means more **quality** and **trust** in the postgrest codebase.

test/fixtures/roles.sql
steve-chavez: Mentioned the issue on https://github.com/PostgREST/postgrest/issues/1731. Didn't want to open a full issue as bug because as it is, `root-spec` is incomplete.

src/PostgREST/Statements.hs
```diff
@@ -190,3 +194,19 @@ decodeGucHeaders = first (const GucHeadersError) . JSON.eitherDecode . toS <$> H
...
 decodeGucStatus = first (const GucStatusError) . fmap (Just . toEnum . fst) . decimal <$> HD.text
+
+-- | Get db settings from the connection role. Only used for configuration.
+dbSettingsStatement :: H.Statement () [(Text, Text)]
+dbSettingsStatement = H.Statement sql HE.noParams decodeSettings False
+  where
+    sql = [q|
+      with
+      role_setting as (
+        select unnest(setconfig) as setting from pg_catalog.pg_db_role_setting where setrole = 'postgrest_test_authenticator'::regrole::oid
```
ruslantalpa, line 205: postgrest_test_authenticator hardcoded here does not look right, `current_user` maube?

wolfgangwalther: Yes, this was fixed in #1743.

ruslantalpa:
While the settings are scoped per role, they are readable by everybody that can connect to the database.
This is a bad idea for storing `jwt-secret` specifically.
It is convenient for rotating the secret but it's very insecure.

test/fixtures/roles.sql
```diff
@@ -5,3 +5,32 @@ CREATE ROLE postgrest_test_default_role;
 CREATE ROLE postgrest_test_author;
 
 GRANT postgrest_test_anonymous, postgrest_test_default_role, postgrest_test_author TO :USER;
+
+-- reloadable config options for io tests
+ALTER ROLE postgrest_test_authenticator SET pgrst."jwt-aud" = 'https://example.org';
+ALTER ROLE postgrest_test_authenticator SET pgrst."openapi-server-proxy-uri" = 'https://example.org/api';
```
ruslantalpa, line 11: while naming the settings like `pgrst."openapi-server-proxy-uri"`, the alter statement works, and they are stored in the pg config tables, however, all the settings named like this can not be read using `show setting` or `current_setting`, and it should work under normal conditions (i bet that is why you used that semi complicated statement to get them back).

The problem here is the `-` character, it it were named like `pgrst.openapi_server_proxy_uri` (or maybe `PGRST.OPENAPI_SERVER_PROXY_URI` like the env vars) then `show` and `current_setting` functions work properly and the `dbSettingsStatement` statement can be a bit more clear/explicit as to what it's fetching from the database

ruslantalpa: discussion here https://www.postgresql.org/message-id/flat/20210209144059.GA21360%40depesz.com

wolfgangwalther: It might not work with `SHOW`, but it does work with `current_setting`.

```sql
set app."test-a-b-c" = 'hello world';
select current_setting('app.test-a-b-c');
```

returns "hello world" for me. No problem there.

wolfgangwalther: PG 13.1 here.

wolfgangwalther: Hm. I'm probably mis-understanding something. The same thing works for me with `SHOW`.

wolfgangwalther: Hm. Happens only when using `ALTER ROLE ... SET` - then loading the guc on login is broken once there is a `-` in it. I'd consider this a bug in Postgres.

steve-chavez: Ah, can reproduce as well on pg 11.5 - not even `current_setting` works when the guc has a dash. I guess this should be fine for our config since we get it through `pg_db_role_setting`. But not for using the guc configs through functions, which should be the case for `jwt-secret`.

Not sure if we should change the dashes to underscores, looks like this is already been considered as a bug on the pg mailing list.

ruslantalpa:
While it might be a bug, it will be fixed only for 13+ which means postgrest will not work properly for any other version. This is not released yet so its the perfect time to switch to _  just like the env vars

steve-chavez:
> This is a bad idea for storing jwt-secret specifically (unless you really tightly control who can connect to the database and run queries).

True. That's something the `app.settings.jwt-secret` solution avoided since it was transaction scoped. I don't see a way to restrict access unless schema variables(as mentioned on https://github.com/PostgREST/postgrest/issues/1562#issue-661980151) land on pg.

It's not a risk if you don't allow a direct SQL connection for external users though. Which I think it's the case for most PostgREST users.

test/fixtures/roles.sql
wolfgangwalther: Looking at the latest post on the mailing list, the bug seems to be that those variables are actually available with `_` instead of `-` already.

So for all PG versions that have the bug, `SELECT current_setting('pgrst.openapi_server_proxy_uri')` should work even without changes from our side.

steve-chavez: @wolfgangwalther Great! I've also confirmed this works. 

Though I guess it'd be good to keep the config option names consistent. I'll change the dashes to underscores on another PR.

ruslantalpa:
the `jwt-secret` thing..., the convenience is too good to pass on. A good middle ground is, when the docs are written for this feature, make a big note in bold/red next to jwt warning people that it's on them to balance security vs convenience. In a world where everybody connects with a superuser to the database, this is not a big deal (and maybe with some work, access to pg_catalog can be restricted).

steve-chavez:
> (and maybe with some work, access to pg_catalog can be restricted).

I think this would be enough for protecting the setttings:

```sql
-- don't allow reading the settings for every role
REVOKE SELECT ON pg_catalog.pg_db_role_setting FROM public;

-- explicitly allow the authenticator to read the settings
GRANT SELECT ON pg_catalog.pg_db_role_setting TO authenticator;
```

Then every non-SUPERUSER role that connects to the db cannot read from `pg_db_role_setting`. Access would have to be granted explicitly.

The above would have to be done for every postgrest-enabled database.

wolfgangwalther:
> If we keep this isolated to the authenticator role, uninstalling postgrest(not that we want that) would be matter of dropping the roles + schema. I also feel that this would look better on the docs - doing an ALTER DATABASE looks more invasive(though requiring the same privileges unfortunately).
> 
> Also, the ROLE approach it's more flexible, we could even remove the `db-settings-prefix` if we tell the user to create different ROLEs for different postgrest instances on the same db. I've noted it's a bit inconvenient to set a prefix when you only have a single postgrest instance.
> 
> So that's why I think we should restrict to ROLEs only.

So `ALTER ROLE` is better than `ALTER DATABASE` - agreed. I just noticed, however, that it only works with `ALTER ROLE ... IN DATABASE SET ...`, but does not when using `ALTER ROLE ... SET`.

In my current project, I don't know the name of the database in advance (could be development, staging, ...) and there is nothing like `ALTER ROLE ... IN CURRENT DATABASE ...`, yet.

We could probably read both the settings for the current database and the settings for all databases at the same time from `pg_db_role_setting`, right? We'd just need to make sure to merge properly in case of conflicts.

WDYT?

test/fixtures/roles.sql
wolfgangwalther: This will be our only way forward, too, because the consensus on the PG mailing list now is to don't allow `-` in GUC names anymore.

wolfgangwalther:
> > (and maybe with some work, access to pg_catalog can be restricted).
> 
> I think this would be enough for protecting the setttings:
> 
> ```sql
> -- don't allow reading the settings for every role
> REVOKE SELECT ON pg_catalog.pg_db_role_setting FROM public;
> 
> -- explicitly allow the authenticator to read the settings
> GRANT SELECT ON pg_catalog.pg_db_role_setting TO authenticator;
> ```
> 
> Then every non-SUPERUSER role that connects to the db cannot read from `pg_db_role_setting`. Access would have to be granted explicitly.
> 
> The above would have to be done for every postgrest-enabled database.

Sounds cool. I haven't tested it - does `ALTER ROLE ... SET` still work for a regular user on login with that? Not sure if permissions are checked at login time when `SET`-ing all the stuff.

wolfgangwalther:
> So `ALTER ROLE` is better than `ALTER DATABASE` - agreed. I just noticed, however, that it only works with `ALTER ROLE ... IN DATABASE SET ...`, but does not when using `ALTER ROLE ... SET`.
> 
> In my current project, I don't know the name of the database in advance (could be development, staging, ...) and there is nothing like `ALTER ROLE ... IN CURRENT DATABASE ...`, yet.
> 
> We could probably read both the settings for the current database and the settings for all databases at the same time from `pg_db_role_setting`, right? We'd just need to make sure to merge properly in case of conflicts.
> 
> WDYT?

Looking at the test-cases in here a bit more closely, I guess my problem is somewhere else. Still.. looking at the code I wonder what happens when somebody sets both `ALTER ROLE ... SET` and `ALTER ROLE ... IN DATABASE ... SET`? I don't think this will be handled properly.

And the query seems to assume that all settings in `pg_db_role_setting` apply to the current database, too. But that's not the case, this table is cluster-wide - so we'd need at least a filter for `setdatabase` to filter for the current database and `NULL`s only. Right now, I think it's possible to add a setting in a completely unrelated database and have it affect the postgrest instance running in this database.

steve-chavez:
> I just noticed, however, that it only works with ALTER ROLE ... IN DATABASE SET ..., but does not when using ALTER ROLE ... SET.

Oh, that shouldn't be the case. According to tests:

https://github.com/PostgREST/postgrest/blob/0ddd676ef04a72d89c0fa734aeca0596841874e4/test/fixtures/roles.sql#L10-L15

> In my current project, I don't know the name of the database in advance (could be development, staging, ...) and there is nothing like ALTER ROLE ... IN CURRENT DATABASE ..., yet.
>  what happens when somebody sets both ALTER ROLE ... SET and ALTER ROLE ... IN DATABASE ... SET? I don't think this will be handled properly.
> And the query seems to assume that all settings in pg_db_role_setting apply to the current database

I definitely intended for the ROLE guc settings to be global - all databases. I didn't know this was only being applied for the current db, I'll check it out.

> does ALTER ROLE ... SET still work for a regular user on login with that? Not sure if permissions are checked at login time when SET-ing all the stuff.

If the `authenticator` role doesn't have `GRANT SELECT ON pg_catalog.pg_db_role_setting`, this is what happens:

```bash
An error ocurred when trying to query database settings for the config parameters:
SessionError (QueryError "\n      with\n      role_setting as (\n        select unnest(setconfig) as setting from pg_catalog.pg_db_role_setting where setrole = current_user::regrole::oid\n      ),\n      kv_settings as (\n        select split_part(setting, '=', 1) as key, split_part(setting, '=', 2) as value from role_setting\n      )\n      select replace(key, 'pgrst.', '') as key, value from kv_settings where key like 'pgrst.%';\n    " [] (ResultError (ServerError "42501" "permission denied for table pg_db_role_setting" Nothing Nothing)))
Listening on port 3000
Connection successful
Schema cache loaded
```

So postgrest still starts as usual, but it cannot get the guc settings from the db, and it sticks to the config file settings.

If the `authenticator` is GRANTed SELECT, then the error is cleared and the guc settings are loaded as usual.

wolfgangwalther:
> I definitely intended for the ROLE guc settings to be global - all databases. I didn't know this was only being applied for the current db, I'll check it out.

Not sure whether there is a mis-understanding. The problem is this:

- postgrest runs on database A
- `ALTER ROLE authenticator IN DATABASE B SET pgrst.xxx TO zzz;`

When postgrest (connected to database A) reads `pg_db_role_setting` it will get a row for `pgrst.xxx` back - even though that is set in database B only. The table `pg_db_role_setting` is cluster-wide instead of database-wide, so we need to account for that by selecting for the current db only! And for global settings (db column = NULL), of course.

> If the `authenticator` role doesn't have `GRANT SELECT ON pg_catalog.pg_db_role_setting`, this is what happens:
> [...]
> If the `authenticator` is GRANTed SELECT, then the error is cleared and the guc settings are loaded as usual.

Definitely a mis-understanding.

I wanted to know if everything still works correctly after `REVOKE SELECT`:
```sql
create role alice login;
alter role alice set a.b to 'c';
revoke select on pg_catalog.pg_db_role_setting from public;
```
Is it still possible to login as user alice without error? I just tested it and it still works. `SHOW a.b;` works, too - the value is correctly set on login. So no problem here.

wolfgangwalther:
And another problem with loading the config:

I have a docker setup with one container for the database and one container for postgrest. When I launch both containers at the same time (`docker-compose up`), the database container will seed itself and take a bit of time to start up. While that happens PostgREST tries to connect already, but fails at first. It then reconnects sucessfully - but at that time the config is not reloaded from the database. This means that the database config is never loaded.

This is the postgrest output:

```
postgrest_1  | An error ocurred when trying to query database settings for the config parameters:
postgrest_1  | ConnectionError (Just "could not connect to server: Connection refused\n\tIs the server running on host \"database\" (172.19.0.2) and accepting\n\tTCP/IP connections on port 5432?\n")
postgrest_1  | Attempting to connect to the database...
postgrest_1  | Listening on port 3000
postgrest_1  | {"details":"could not connect to server: Connection refused\n\tIs the server running on host \"database\" (172.19.0.2) and accepting\n\tTCP/IP connections on port 5432?\n","code":"","message":"Database connection error. Retrying the connection."}
postgrest_1  | Attempting to reconnect to the database in 0 seconds...
postgrest_1  | Connection successful
postgrest_1  | Schema cache loaded
```

What's basically missing is a `Config reloaded` before the `Schema cache loaded`, right?

Should we reload the config whenever the connection is re-tried? And possibly the schema cache, too? (If not already the case)

wolfgangwalther:
Summarizing a TODO list from all the points above, so we're not missing anything:
- [ ] Document security concerns with `pgrst.jwt-secret` and how to protect the setting table
- [ ] Change `pgrst.xxx-yyy` variables to use `_` instead of `-`. Or maybe `.`, would be more consistent: `pgrst.jwt.secret = 'xyz'`
- [ ] Fix config reloading after reconnect
- [ ] Ignore settings from other databases (`ALTER ROLE ... IN DATABASE B SET ...` should not affect postgrest in database A)
- [ ] Merge database specific and global settings consistently (`ALTER ROLE ... IN DATABASE ... SET` should override `ALTER ROLE ... SET`)

steve-chavez:
Working this on https://github.com/PostgREST/postgrest/pull/1760.

> Change pgrst.xxx-yyy variables to use _ instead of -. Or maybe ., would be more consistent: pgrst.jwt.secret = 'xyz'

I went with underscore - `pgrst.jwt_secret`. They look similar to env var settings and I thought names like `pgrst.server.unix.socket.mode` looked weird, only server is a namespace there.

PR #1737, opened 2021-01-21
Fix notify config

steve-chavez:
## Bug fix

Fixes a bug in the recently added https://github.com/PostgREST/postgrest/pull/1729.

When altering a config with an invalid setting and reloading:

```sql
ALTER ROLE postgrest_test_authenticator SET pgrst."jwt-role-claim-key" = 'invalid';
NOTIFY pgrst, 'reload config';
```

postgrest will respond with:

```
Retrying listening for notifications on the pgrst channel..
Connection successful
Listening for notifications on the pgrst channel
Schema cache loaded
```

This is because the `readConfig` function panicked and killed the `listener` thread.

## Solution

Don't panic on `readConfig` and instead do it on the `Main.hs` module.

So with this change, postgrest will respond with:

```
Failed config reload. Error in config: "\"failed to parse role-claim-key value (asdf)\" (line 1, column 1):\nunexpected \"a\"\nexpecting period (.)"
```

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -44,7 +44,6 @@ import Control.Lens            (preview)
 import Control.Monad           (fail)
 import Crypto.JWT              (JWKSet, StringOrURI, stringOrUri)
 import Data.Aeson              (encode, toJSON)
-import Data.Either.Combinators (fromRight', whenLeft)
 import Data.List               (lookup)
 import Data.List.NonEmpty      (fromList, toList)
 import Data.Maybe              (fromJust)
@@ -108,7 +107,7 @@ data AppConfig = AppConfig {
   , configDbUri                 :: Text
   , configJWKS                  :: Maybe JWKSet
   , configJwtAudience           :: Maybe StringOrURI
-  , configJwtRoleClaimKey       :: Either Text JSPath
+  , configJwtRoleClaimKey       :: JSPath
   , configJwtSecret             :: Maybe B.ByteString
   , configJwtSecretIsBase64     :: Bool
   , configLogLevel              :: LogLevel
@@ -117,7 +116,7 @@ data AppConfig = AppConfig {
   , configServerHost            :: Text
   , configServerPort            :: Int
   , configServerUnixSocket      :: Maybe FilePath
-  , configServerUnixSocketMode  :: Either Text FileMode
+  , configServerUnixSocketMode  :: FileMode
   }
 
 configDbPoolTimeout' :: (Fractional a) => AppConfig -> a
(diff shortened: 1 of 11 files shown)


src/PostgREST/Config.hs
```diff
@@ -331,22 +330,18 @@ instance JustIfMaybe a (Maybe a) where
...
+readAppConfig :: [(Text, Text)] -> Environment -> Maybe FilePath -> IO (Either Text AppConfig)
 readAppConfig dbSettings env optPath = do
   -- Now read the actual config file
   conf <- case optPath of
-    Just cfgPath -> catches (C.load cfgPath)
-      [ Handler (\(ex :: IOError)    -> exitErr $ "Cannot open config file:\n\t" <> show ex)
-      , Handler (\(C.ParseError err) -> exitErr $ "Error parsing config file:\n" <> err)
+    Just cfgPath -> C.load cfgPath `catches`
+      [ Handler (\(ex :: IOError)    -> panic $ "Cannot open config file: " <> show ex)
+      , Handler (\(C.ParseError err) -> panic $ "Error parsing config file: " <> err)
```
steve-chavez, line 339: Ideally these panics would go in Main, but I wasn't able to change them to Lefts.

They're only likely to occur on startup though(not during reload), so I thought it's fine to leave them for now.

steve-chavez:
I tried addding a test for this. But I couldn't figure out how to do something like:

```python

    postgrest.session.post(
        "/rpc/change_role_claim_key", data={"val": "invalid", "notify": True}
    )

   ## this part
   postgrest.process.stdout == "Failed config reload. Error in config:.."
```

When trying to print `postgrest.process.stdout`, the test hanged.

test/io-tests/fixtures.yaml
```diff
@@ -173,5 +173,4 @@ invalidroleclaimkeys:
   - '.role##'
   - '.my_role;;domain'
   - '.#$$%&$%/'
-  - ''
```
steve-chavez, line 176: empty string is treated as Nothing now, so the default `.role` applies. This makes it easier to reuse the `optString` function. I think it's fine.

steve-chavez:
> But I couldn't figure out how to do something like

Managed to add a python test now. It uses `postgrest.process.stderr`.

steve-chavez:
I think it's good to merge. It corrects a previous PR and adds it one more test.

src/PostgREST/Config.hs
wolfgangwalther: > They're only likely to occur on startup though(not during reload), so I thought it's fine to leave them for now.

Not sure I understand that reasoning. Whenever somebody deletes the config file and then tries to reload - or when they change the config file to something with invalid syntax and then reloads - those will occur. This seems equally likely on startup and reload.

wolfgangwalther:
I like the refactoring that happenend in `Config.hs`. A lot clearer now.

src/PostgREST/Config.hs
steve-chavez: You're right. I've found that I can replace the `catches` with [try](https://hackage.haskell.org/package/base-4.14.1.0/docs/Control-Exception.html#v:try). I'll make a PR with that change.

Issue #1740, opened 2021-01-25
Hardcoded postgrest_test_authenticator in latest nightly


LorenzHenk:
### Environment

* PostgreSQL version: 13.1
* PostgREST version: nightly-2021-01-22-15-56-6557f1f

### Description of issue

In the latest nightly, when trying to run `postgrest --dump-config`, I get the following error:

```
$ postgrest --dump-config

An error ocurred when trying to query database settings for the config parameters:
SessionError (QueryError "\n      with\n      role_setting as (\n        select unnest(setconfig) as setting from pg_catalog.pg_db_role_setting where setrole = 'postgrest_test_authenticator'::regrole::oid\n      ),\n      kv_settings as (\n        select split_part(setting, '=', 1) as key, split_part(setting, '=', 2) as value from role_setting\n      )\n      select replace(key, 'pgrst.', '') as key, value from kv_settings where key like 'pgrst.%';\n    " [] (ResultError (ServerError "42704" "role \"postgrest_test_authenticator\" does not exist" Nothing Nothing)))
```

Seems like the role "postgrest_test_authenticator" was hardcoded here: https://github.com/PostgREST/postgrest/blob/125ea8f6d9ecfe40ed4a1c5a9291ae23bb695972/src/PostgREST/Statements.hs#L205
Related: #1729

The same error also appears on `postgrest --dump-schema`, but the schema JSON is still printed.

wolfgangwalther:
Wow. I wonder how I missed that in my review of that PR :D.

Should be enough to replace `'postgrest_test_authenticator'` with `current_user`.

steve-chavez:
@LorenzHenk Set `db-load-guc-config = false` to fix this error.

LorenzHenk:
Thanks for the quick answers!

steve-chavez:
@LorenzHenk The latest `postgrest-nightly-2021-01-25-18-38-c93e8f9` on [Nightly builds](https://github.com/PostgREST/postgrest/releases/tag/nightly) should have this one fixed!

PR #1743, opened 2021-01-25
Correct hardcoded postgrest_test_authenticator

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/1740.

Made sure the patch is working by adding a test for another authenticator role. This test fails on current master.


src/PostgREST/Statements.hs
```diff
@@ -202,7 +202,7 @@ dbSettingsStatement = H.Statement sql HE.noParams decodeSettings False
     sql = [q|
       with
       role_setting as (
-        select unnest(setconfig) as setting from pg_catalog.pg_db_role_setting where setrole = 'postgrest_test_authenticator'::regrole::oid
+        select unnest(setconfig) as setting from pg_catalog.pg_db_role_setting where setrole = current_user::regrole::oid
```
steve-chavez, line 205: I'm pretty sure I started the notify feature with `current_user`. I must have changed it when doing tests and then I squashed the commits locally plus force pushed :man_facepalming: .

steve-chavez: (did that here: https://github.com/PostgREST/postgrest/pull/1729#issuecomment-762938712)

test/io-tests/test_io.py
```diff
@@ -283,6 +283,24 @@ def test_expected_config_from_db_settings(defaultenv):
 
     assert dumpconfig(configpath=config, env=env) == expected
 
+def test_expected_config_from_db_settings_other_authenticator(defaultenv):
+    "NOTIFY reload config should show an error if role-claim-key is invalid"
```
wolfgangwalther, line 287: The description seems wrong. Also the `PGRST_DB_CHANNEL_ENABLED` looks like this test was copied from somewhere ;).

I suggest to just parametrize the `test_expected_config_from_db_settings` test with a db role parameter. Then this will run twice with just a different user.

Set env to something like:
```python
db_uri = defaultenv["PGRST_DB_URI"].replace("user=postgrest_test_authenticator", f"user={role}")
[...]
  "PGRST_DB_URI": db_uri
```
(in the regular case this will just replace `postgrest_test_authenticator` with `postgrest_test_authenticator`)

And then load the the expected config file from either by using multiple parameters with `@pytest.mark.parametrize` or by putting the role name into the filename.

src/PostgREST/Config.hs
```diff
@@ -334,14 +333,12 @@ readAppConfig :: [(Text, Text)] -> Environment -> Maybe FilePath -> IO (Either T
 readAppConfig dbSettings env optPath = do
   -- Now read the actual config file
   conf <- case optPath of
-    Just cfgPath -> C.load cfgPath `catches`
-      [ Handler (\(ex :: IOError)    -> panic $ "Cannot open config file: " <> show ex)
-      , Handler (\(C.ParseError err) -> panic $ "Error parsing config file: " <> err)
-      ]
+    -- Both C.ParseError and IOError are shown here
+    Just cfgPath -> mapLeft (("Bad config file. " <>) . show) <$> (try $ C.load cfgPath :: IO (Either SomeException C.Config))
```
wolfgangwalther, line 337: Hm. It looks like we currently don't have a test that checks the output of that. Is the error message still different somehow for `ParseError` and `IOError` - i.e. is it possible to tell from the error message what the problem is? (not sure what `show` will tell us here)

steve-chavez: Yeah, I already tested that manually. The errors are different:

```
-- For ParseError
postgrest: FatalError {fatalErrorMessage = "Bad config file. ParseError \"support/configs/config:1:14:\\n  |\\n1 | qwerqwerqwer qwer qwer ewqr qew rewq q
ewr\\n  |              ^\\nunexpected 'q'\\nexpecting '=' or '{'\\n\""}

-- For IOError. unexistent file
postgrest: FatalError {fatalErrorMessage = "Bad config file. Cannot open config file: support/configs/configx: openFile: does not exist (No such file or directory)"}
```

wolfgangwalther: Ah cool. You know what: I think we don't even need the `Bad config file.` prefix. Those two error messages are good like they are? Almost at least, it would be good to have a nicer `show` instance for `ParseError` without all the `\"` and `\n`?

steve-chavez: How would I test the error message?

I tried with the `cli` function but I get a PostgrestError exception.

Also, seems `fixtures.yaml` already test this somehow? I see:
```
  - name: non-existant config file
    expect: error
    args: ['does_not_exist.conf']
  - name: invalid config file
```

wolfgangwalther: Actually `FatalError` is probably the one that needs better `show`?

wolfgangwalther: > How would I test the error message?
> 
> I tried with the `cli` function but I get a PostgrestError exception.
> 
> Also, seems `fixtures.yaml` already test this somehow? I see:
> 
> ```
>   - name: non-existant config file
>     expect: error
>     args: ['does_not_exist.conf']
>   - name: invalid config file
> ```

Yes, this was mainly for coverage. I didn't care too much about the actual output in those cases, yet. This is something that is not implemented, yet. I suggest to leave that for later, because we should add a bit of test infrastructure here and test other error messages, too.

steve-chavez: >   I suggest to leave that for later, because we should add a bit of test infrastructure here and test other error messages, too.

Cool. Then I'll just remove the `Bad config file` prefix for now.

test/io-tests/test_io.py
steve-chavez: Added the pytest parameter. Thanks a lot for your help Wolfgang :). I'm still getting used to the python tests.

steve-chavez:
Done with the reviews. I'll merge this now, the bug on nightly is quite bad.

src/PostgREST/Config.hs
wolfgangwalther: I left a note in #1719, so we can remember to improve the test-cases here.

PR #1751, opened 2021-02-06
Add a script to generate a graph of Haskell imports

monacoremo:
Current output:
imports

This could be a useful tool to maintain the module structure.

### DIFF
--- a/nix/devtools.nix
+++ b/nix/devtools.nix
@@ -3,6 +3,7 @@
 , checkedShellScript
 , devCabalOptions
 , entr
+, graphviz
 , silver-searcher
 , style
 , tests
@@ -96,6 +97,29 @@ let
         ${style}/bin/postgrest-lint
         ${style}/bin/postgrest-style-check
       '';
+
+  importsgraph =
+    checkedShellScript
+      {
+        name = "postgrest-importsgraph";
+        docs =
+          ''
+            Render the imports between PostgREST modules as a graph.
+
+            Output is written to 'imports.png' in the root directory of the project.
+          '';
+        inRootDir = true;
+      }
+      ''
+        grep -rE 'import .*PostgREST\.' src \
+          | sed -E \
+              -e 's|/|\.|g' \
+              -e 's/src\.(.*)\.hs:import .*(PostgREST\.\S+)( .*)?/"\1" -> "\2"/' \
+              -e '1 i digraph \{' \
+              -e '$ a \}' \
+          | ${graphviz}/bin/dot -Tpng \
+          > imports.png
+      '';
 in
 buildEnv {
   name = "postgrest-devtools";
@@ -106,5 +130,6 @@ buildEnv {
     run.bin
     clean.bin
     check.bin
+    importsgraph.bin
   ];
 }
--- a/nix/devtools.nix
+++ b/nix/devtools.nix
@@ -111,14 +111,26 @@ let
         inRootDir = true;
       }
       ''
-        grep -rE 'import .*PostgREST\.' src \
-          | sed -E \
-              -e 's|/|\.|g' \
-              -e 's/src\.(.*)\.hs:import .*(PostgREST\.\S+)( .*)?/"\1" -> "\2"/' \
-              -e '1 i digraph \{' \
-              -e '$ a \}' \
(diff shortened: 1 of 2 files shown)


steve-chavez (approved):
I like it! I think it could be useful for new contributors as well(for an overview of the modules) and it could even go somewhere in our docs.

nix/devtools.nix
```diff
@@ -96,6 +97,29 @@ let
...
+      }
+      ''
+        grep -rE 'import .*PostgREST\.' src \
+          | sed -E \
+              -e 's|/|\.|g' \
+              -e 's/src\.(.*)\.hs:import .*(PostgREST\.\S+)( .*)?/"\1" -> "\2"/' \
+              -e '1 i digraph \{' \
+              -e '$ a \}' \
+          | ${graphviz}/bin/dot -Tpng \
+          > imports.png
```
steve-chavez, line 121: Would it be possible to print the module Description(top comment of the file)? I wonder how would that look.

nix/devtools.nix
```diff
@@ -96,6 +97,29 @@ let
...
+        docs =
+          ''
+            Render the imports between PostgREST modules as a graph.
+
+            Output is written to 'imports.png' in the root directory of the project.
+          '';
+        inRootDir = true;
+      }
+      ''
+        grep -rE 'import .*PostgREST\.' src \
```
wolfgangwalther, line 114: You might want to have a look at GHC's `-ddump-minimal-imports` option. This should return stuff in a better to parse way. Parsing the output of that could also allow us to add more data about those imports to the graph. You could e.g. add the number of functions/types used in this import on the edges of the graph?

wolfgangwalther:
The tool is great - but looking at the image it's so much chaos... There is a lot of potential for improvement in the number of imports, I think.

Can we add `main/...`, too?

nix/devtools.nix
monacoremo: Just added in a new commit - new output:

imports

monacoremo:
> Can we add `main/...`, too?

Yes, latest commit:

imports

monacoremo:
> The tool is great - but looking at the image it's so much chaos... There is a lot of potential for improvement in the number of imports, I think.

Absolutely! Looks quite like spaghetti right now... We should be able to simplify the structure a lot, making the codebase easier to understand.

nix/devtools.nix
monacoremo: To get the dump, I ran:

```bash
mkdir dump
cabal v2-build --ghc-option=-ddump-minimal-imports --ghc-option=-dumpdir=dump
```

Example output: https://gist.github.com/monacoremo/d72726e7afad5c05f95cc34301c62a9f

Suggest that we keep the current approach for now, as parsing the number of imports could be tricky and running the build to generate the dump adds a long extra step. But we should keep it in mind for later improvements!

monacoremo: @wolfgangwalther Experimented on https://github.com/monacoremo/postgrest/tree/symbolimports

This yields a map of all imported symbols:
symbols

Many symbols are only imported from one module, might be good leads for refactoring!

monacoremo:
Merged this first version that should be useful already. @wolfgangwalther thanks for the pointer to -ddump-minimal-imports! The experiment linked above has a bug for long lists of constructors, but should be easy to fix - will submit in a further PR for discussion.

wolfgangwalther:
> > Can we add `main/...`, too?
> 
> Yes, latest commit:

I think `UnixSocket.hs` is currently missing from the output. I'm not too concerned about that, though, because you had some refactoring in the pipeline that would move all of this to the postgrest lib, if I remember correctly.

> Suggest that we keep the current approach for now, as parsing the number of imports could be tricky and running the build to generate the dump adds a long extra step. 

Ah, yes, sorry. I remember looking into this, because the output of `-ddump-minimal-imports` could help us in other ways, too - standardizing imports across files, building our own protolude. I figured we'd need to parse the import somehow - but forgot about that we still need to do that.

If we could add a tool to dump those minimal imports and parse them in a structured way (e.g. CSV or JSON, one file for all imports across the project, one item for each imported symbol), we could use this output for a number of use cases: creating graphs like you do here, but also manually checking for consistent alias usage across files and stuff like that.

PR #1756, opened 2021-02-09
Fix "function string_agg(integer, unknown) does not exist"

### LINKED ISSUE #1755: ERROR:  function string_agg(integer, unknown) does not exist
### Environment

* PostgreSQL version: 11
* PostgREST version: 7.0.1 (d6050c8)
* Operating system: Ubuntu 20.04.2 LTS

### Description of issue

I am calling a function via the RPC endpoint having all arguments as TEXT of varying length. However, the PostgreSQL logs indicate that the query being executed by PostgREST calls the function `string_agg(integer, unknown)`, which does not exist.

The portion of the query that is buggy is as follows.
`	      SELECT
	        coalesce(string_agg(_postgrest_t."pgrst_scalar", ''), '') AS body,
	        coalesce(nullif(current_setting('response.headers', true), ''), '[]') AS response_headers
	      FROM (SELECT "pg_source".* FROM "pg_source"    ) _postgrest_t;`

Over here, the call to `string_agg` throws the error: `ERROR:  function string_agg(integer, unknown) does not exist at character `...

The hint provided by the PostgreSQL server is: `HINT:  No function matches the given name and argument types. You might need to add explicit type casts.`

The offending line of code is:
`asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldName <> ", ''), '')"` in the file `src/PostgREST/Private/QueryFragment.hs`

anirudh-ramesh:
This commit type-casts arguments to the PostgreSQL function `string_agg` in order to avoid `string_agg(integer, unknown)` and instead appropriately call `string_agg(text, text)`.

Closes #1755

### DIFF
--- a/src/PostgREST/Private/QueryFragment.hs
+++ b/src/PostgREST/Private/QueryFragment.hs
@@ -101,7 +101,7 @@ asJsonSingleF returnsScalar
   | otherwise     = "coalesce(string_agg(to_json(_postgrest_t)::text, ','), '')::character varying"
 
 asBinaryF :: FieldName -> SqlFragment
-asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldName <> ", ''), '')"
+asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldName <> "::text, ''::text), '')"
 
 locationF :: [Text] -> SqlFragment
 locationF pKeys = [qc|(


wolfgangwalther:
We need `string_agg` to concat rows as-is when a custom mimetype is requested. `string_agg` has two signatures: once taking `TEXT` and once taking `BYTEA` arguments. We certainly don't want RPCs that return binary data to do the text conversion here. Let's discuss in the issue how to modify your RPC to fix this.

Issue #1689, opened 2020-12-11
Make `db-anon-role` optional


wolfgangwalther:
Came up in https://github.com/PostgREST/postgrest-docs/issues/381.

If we make `db-anon-role` optional, we can block any anonymous access when it is not set.

wolfgangwalther:
Taken from #1823:

> we should print a warning when neither anonymous access nor JWT is set up

This would make the server pretty much disfunctional.

steve-chavez:
Taken from https://github.com/PostgREST/postgrest/issues/2104#issuecomment-1001175822:

> I expected that if I set the jwt-secret parameter in the configuration, it would automatically set that only jwt-authorization would work.

I like the above, when the `jwt-secret` is set, force the JWT for all requests and basically disable `db-anon-role`. It's good for a quickstart because all users start with an empty `jwt-secret`. It kinda goes in line with our tutorials as well, tut0 main topic is the anonymous role and for tut1 is the  JWT secret.

Having `db-anon-role` by default is still a problem because it requires a proxy for disallowing public access, something we can do ourselves.

wolfgangwalther:
> I like the above, when the `jwt-secret` is set, force the JWT for all requests and basically disable `db-anon-role`. It's good for a quickstart because all users start with an empty `jwt-secret`. It kinda goes in line with our tutorials as well, tut0 main topic is the anonymous role and for tut1 is the JWT secret.

I don't see how that's related to `jwt-secret`?

You can't disable anonymous requests, when `db-anon-role` is given - but you can't enable them either, when it's not (once that's possible).

A given `jwt-secret` will just enable **authenticated** requests. If it's not there, no authenticated requests are possible. But I don't see any interaction with `db-anon-role` / anonymous requests, other than https://github.com/PostgREST/postgrest/issues/1689#issuecomment-822251955.

cemremengu:
Technically if anon role is not defined on db side no requests would be allowed so it is good.

But still I am thinking on the necessity of anon role. I am pretty sure a rest api for database will almost always need some kind of auth and believe it is a security issue to have an anon role.

Going with secure by default, instead of anon `basic authentication` should be provided as the default and then jwt could optionally be enabled.

wolfgangwalther:
> Going with "secure by default" mantra, instead of anon `basic authentication` should be provided as the default and then jwt could optionally be enabled.

Could you elaborate on that? What would basic authentication do? And how would that be more secure?

[...]

PR #1760, opened 2021-02-18
db settings: reload on connection recovery, only get them from global/current-database scope and change dash to underscore

steve-chavez:
Corrections of db settings discussed on: https://github.com/PostgREST/postgrest/pull/1729#issuecomment-779436633

- [x] Change `pgrst.xxx-yyy` variables to use `_` instead of `-`.
- [x] Ignore settings from other databases (`ALTER ROLE ... IN DATABASE B SET ...` should not affect postgrest in database A
- [x] Merge database specific and global settings consistently (`ALTER ROLE ... IN DATABASE ... SET` should override `ALTER ROLE ... SET`)
- [x] Fix config reloading after reconnect


src/PostgREST/Statements.hs
```diff
@@ -195,18 +195,23 @@ decodeGucHeaders = first (const GucHeadersError) . JSON.eitherDecode . toS <$> H
...
+          and setdatabase in (0, (select oid from pg_catalog.pg_database where datname = current_catalog))
       ),
       kv_settings as (
-        select split_part(setting, '=', 1) as key, split_part(setting, '=', 2) as value from role_setting
+        select setdatabase, split_part(setting, '=', 1) as k, split_part(setting, '=', 2) as value from role_setting
       )
-      select replace(key, 'pgrst.', '') as key, value from kv_settings where key like 'pgrst.%';
+      select distinct on (key) replace(k, 'pgrst.', '') as key, value
+      from kv_settings
+      where k like 'pgrst.%'
```
wolfgangwalther, line 214: The `where` here could be moved into one of the two CTEs. No need to split all the other settings, that are not `pgrst.xxx`.

test/fixtures/roles.sql
```diff
@@ -7,46 +7,50 @@ CREATE ROLE postgrest_test_author;
...
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_tx_end" = 'commit-allow-override';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_schemas" = 'test, tenant1, tenant2';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_root_spec" = 'root';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_prepared_statements" = 'false';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_pre_request" = 'test.custom_headers';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_max_rows" = '1000';
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_extra_search_path" = 'public, extensions';
+
+-- override with database specific setting
+ALTER ROLE postgrest_test_authenticator SET pgrst."db_max_rows" = '1111';
```
wolfgangwalther, line 25: This line doesn't match the comment - it's not database specific. Was it supposed to be `IN DATABASE`, too?

test/fixtures/roles.sql
```diff
@@ -7,46 +7,50 @@ CREATE ROLE postgrest_test_author;
...
+ALTER ROLE postgrest_test_authenticator IN DATABASE postgres SET pgrst."db_extra_search_path" = 'public, extensions, private';
```
wolfgangwalther, line 26:
```suggestion
ALTER ROLE postgrest_test_authenticator IN DATABASE :DBNAME SET pgrst."db_extra_search_path" = 'public, extensions, private';
```

I think you can use `:DBNAME` here, because the sql file will be loaded by `psql`. This way the database name is not hardcoded here.

test/fixtures/roles.sql
wolfgangwalther, line 25: Maybe we can do something like this, too?

```suggestion
DROP DATABASE IF EXISTS other;
CREATE DATABASE other;
ALTER ROLE postgrest_test_authenticator IN DATABASE other SET pgrst."db_max_rows" = '1111';
```

And then test that `db-max-rows` is **not** 1111, but still 1000.

steve-chavez: True. I put `IN DATABASE` there and other tests failed.

steve-chavez: I though this couldn't be done because of `CREATE DATABASE cannot be executed inside a transaction block`. But I've tried and it works! We only got `\set ON_ERROR_STOP on` set for the scripts.

wolfgangwalther:
It would be really nice, if we could add a test for reload recovery. This would go a long way in terms of test coverage. In general, I think we're not testing any of the "recover from an error" situations, yet.

However, I can't come up with a way to do this with our current io test setup. `with_tmp_db` makes sure that the database is ready to go, before the tests are run - and I don't know how we could force a failed connection at first, that will succeed later, without sending any reload signal to postgrest (sending a signal doesn't make sense, because it wouldn't test self recovery). @monacoremo any idea?

steve-chavez:
> Fix config reloading after reconnect

I think this is more or less fixed and I'm in the process of adding python tests. However these tests will fail with the latest commit:

https://github.com/PostgREST/postgrest/blob/0ddd676ef04a72d89c0fa734aeca0596841874e4/test/io-tests/test_io.py#L341-L347

This is because the config file for the tests uses `/dev/stdin`:

https://github.com/PostgREST/postgrest/blob/0ddd676ef04a72d89c0fa734aeca0596841874e4/test/io-tests/configs/secret-from-file.config#L3-L4

And since the config file reload is now done in a subthread, stdin will not reach it. Because AFAICT, stdin only reaches the parent process: https://stackoverflow.com/questions/55602283/how-to-write-data-to-stdin-to-be-consumed-by-a-separate-thread-waiting-on-input

Seems the workaround is to create a named pipe somehow for stdin to work, but it looks too complicated to fix this in postgrest(also not sure if possible).

How about if I just change the tests for not using stdin?

@wolfgangwalther Thoughts?

wolfgangwalther:
> How about if I just change the tests for not using stdin?
> 
> @wolfgangwalther Thoughts?

Was "reading the secret from stdin" just a hack to test "reading the secret from file"? So we don't actually support/advertise reading from stdin here? In that case I agree: Just use a proper temp file created with python (have a look at some of the other tests, I think there is a nice pytest-way of doing it).

steve-chavez:
> Just use a proper temp file created with python

Cool. I'll do that.

> However, I can't come up with a way to do this with our current io test setup. 

I've managed to create a couple of utilitarian functions that seem to be working.

```py
def pg_stop():
    "Stop postgresql"
    subprocess.run(["pg_ctl", "stop", "-m", "i"])
    time.sleep(1)

def pg_start():
    "Start postgresql"
    PGHOST = os.getenv("PGHOST")
    subprocess.run(["pg_ctl", "start", "-o", f'-F -c listen_addresses="" -k {PGHOST}'])
```

I'm not sure if they'll hold up when writing the actual tests though. I'm all ears for a better way to do this.

wolfgangwalther:
> > However, I can't come up with a way to do this with our current io test setup.
> 
> I've managed to create a couple of utilitarian functions that seem to be working.
> 
> ```python
> def pg_stop():
>     "Stop postgresql"
>     subprocess.run(["pg_ctl", "stop", "-m", "i"])
>     time.sleep(1)
> 
> def pg_start():
>     "Start postgresql"
>     PGHOST = os.getenv("PGHOST")
>     subprocess.run(["pg_ctl", "start", "-o", f'-F -c listen_addresses="" -k {PGHOST}'])
> ```
> 
> I'm not sure if they'll hold up when writing the actual tests though. I'm all ears for a better way to do this.

I think the problem with that approach is, that it will break running the io tests in parallel. We're using the same database instance for all tests, so those other tests will have a stopped postgres, too.

Maybe we can add some kind of a postgres proxy - pgbouncer or something? I am not familiar with that, but that could be a base to test support for that in general, too.

If we keep postgres running, but only start the proxy/pgbouncer after we started postgrest.. that could work. WDYT?

steve-chavez:
> So we don't actually support/advertise reading from stdin here?

Not that I'm aware of. Seems this was inherited from previous bash tests: https://github.com/PostgREST/postgrest/blob/0c25f128250f427f470910104eea6b8bee589083/test/io-tests.sh#L373

Though it does seem to be a legitimate use case: https://github.com/kubernetes/kubernetes/issues/54200#issuecomment-337880729

(Personally I've never done it)

steve-chavez:
> If we keep postgres running, but only start the proxy/pgbouncer after we started postgrest.. that could work. WDYT?

It does seem interesting. But if the problem is the parallelism of the io tests, perhaps I'd be easier to separate the recovery tests(in a `recovery-tests.py`) and force them to run sequentially. Can that be done with pytest?

(I'm definitely up for including pgbouncer later though)

monacoremo:
It should be possible to nest with_tmp_db instances with no issues. So what we could do is:

* Add an option like `withtmpdb=False` to `run`
* If `withtmpdb` is true, we use test/with_tmp_db postgrest` as the command, join the arguments with `shlex.join` and run the combined command string with `Popen(..., shell=True)`

Then it should be possible to stop just that db without affecting the parent db.

wolfgangwalther:
> It should be possible to nest with_tmp_db instances with no issues. So what we could do is:
> 
>     * Add an option like `withtmpdb=False` to `run`
> 
>     * If `withtmpdb` is true, we use `["test/with_tmp_db", "postgrest"]` as the command
> 
> 
> Then it should be possible to stop just that db without affecting the parent db.
> 
> Edit: simpler solution

I don't think this will work like that. Now we have 2 postgrest instances running - which instance will `pg_ctl` start and stop? Probably depends on environment variables and since the function is called from the io tests, it will still start and stop the global instance - not the `with_tmp_db` that runs together with postgrest.

wolfgangwalther:
Maybe instead of running a full pgbouncer as a proxy, we can just set up a port forwarding via `netcat` (https://stackoverflow.com/a/2150342).

So in the test, we'd:
- take note of the unix socket that `with_tmp_db` provides
- start postgrest with some `localhost:xxx` db-uri setting instead of the socket
- after the first connection failed, we open the port forwarding from `localhost:xxx` to the unix socket from `with_tmp_db` via netcat
- the connection will now succeed

No need to fiddle with restarting postgres etc. either.

steve-chavez:
> Was "reading the secret from stdin" just a hack to test "reading the secret from file"? So we don't actually support/advertise reading from stdin here?

I didn't want to break any use cases(perhaps UNIX-savy users use `stdin`), so I've just fixed the issue. No test was changed.

The main problem was that the files(db-uri, jwt-secret) were read several times and reading `/dev/stdin` makes the thread wait for input, so our tests would have needed repeating the same input several times. It was also weird to see this behavior when running postrest manually. 

What I've changed is that external config files are now only read one time. Also untangled the logic a bit, I think the code is also clearer this way.

> If withtmpdb is true, we use ["test/with_tmp_db", "postgrest"] as the command
> Maybe instead of running a full pgbouncer as a proxy, we can just set up a port forwarding via netcat

Hm, could you guys help me with the recovery tests on another PR? @monacoremo @wolfgangwalther. I can put a list of the manual tests I make for ensuring the recovery is working.

Still have to do some changes here(like verify SIGUSR2 reloads the external config files) and it would take me a while to figure out how to change the setup for the recovery tests.

wolfgangwalther:
> > Was "reading the secret from stdin" just a hack to test "reading the secret from file"? So we don't actually support/advertise reading from stdin here?
> 
> I didn't want to break any use cases(perhaps UNIX-savy users use `stdin`), so I've just fixed the issue. No test was changed.
> 
> The main problem was that the files(db-uri, jwt-secret) were read several times and reading `/dev/stdin` makes the thread wait for input, so our tests would have needed repeating the same input several times. It was also weird to see this behavior when running postrest manually.
> 
> What I've changed is that external config files are now only read one time. Also untangled the logic a bit, I think the code is also clearer this way.

I didn't go through the latest commit, yet, but is my understanding correct as follows?

- config file (and other files) are read once on startup and whenever a config re-read is triggered manually (signal / notify)
- env+file config (without database) is cached
- database config is re-read whenever a connection is (re-)established - same as the schema cache (?)
- database config is applied onto the cached env+file-config
- the files are read in the main thread, but the database config in each worker (?)

If so, that's a lot better, I agree!



> Hm, could you guys help me with the recovery tests on another PR? @monacoremo @wolfgangwalther. I can put a list of the manual tests I make for ensuring the recovery is working.

Absolutely. A list of all your manual tests to reproduce would be great!

src/PostgREST/Config.hs
```diff
@@ -514,90 +539,6 @@ readAppConfig dbSettings env optPath = do
...
-      where
-        chomp bs = fromMaybe bs (BS.stripSuffix "\n" bs)
-    --
-    -- Turns the Base64url encoded JWT into Base64
-    transformString :: Bool -> ByteString -> IO ByteString
-    transformString False t = return t
-    transformString True t =
-      case B64.decode $ encodeUtf8 $ strip $ replaceUrlChars $ decodeUtf8 t of
-        Left errMsg -> panic $ pack errMsg
-        Right bs    -> return bs
```
steve-chavez, line 578: base64 logic was tangled inside `loadSecretFile`, this is now moved to `parseJwtSecret`. `loadSecretFile` just loads the file now.

test/fixtures/roles.sql
```diff
@@ -7,46 +7,55 @@ CREATE ROLE postgrest_test_author;
...
-ALTER ROLE postgrest_test_authenticator SET pgrst."jwt-secret-is-base64" = 'true';
-ALTER ROLE postgrest_test_authenticator SET pgrst."jwt-role-claim-key" = '."a"."role"';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-tx-end" = 'commit-allow-override';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-schemas" = 'test, tenant1, tenant2';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-root-spec" = 'root';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-prepared-statements" = 'false';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-pre-request" = 'test.custom_headers';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-max-rows" = '1000';
-ALTER ROLE postgrest_test_authenticator SET pgrst."db-extra-search-path" = 'public, extensions';
+ALTER ROLE postgrest_test_authenticator SET pgrst.jwt_aud = 'https://example.org';
```
steve-chavez, line 10: Just removed the now unnecessary double quotes here.

test/io-tests/test_io.py
```diff
@@ -88,6 +88,7 @@ def defaultenv():
         "PGRST_DB_SCHEMAS": os.environ["PGRST_DB_SCHEMAS"],
         "PGRST_DB_ANON_ROLE": os.environ["PGRST_DB_ANON_ROLE"],
         "PGRST_DB_LOAD_GUC_CONFIG": "false",
+        "PGRST_LOG_LEVEL": "info",
```
steve-chavez, line 91: Added `info` because it's always useful for failing tests.

main/Main.hs
```diff
@@ -358,12 +358,12 @@ reReadConfig startingUp pool gucConfigEnabled env path refConf dbUriFile secretF
     Left err   ->
       if startingUp
         then panic err -- die on invalid config if the program is starting up
-        else hPutStrLn stderr $ "Failed config load. " <> err
+        else hPutStrLn stderr $ "Failed loading in-database config. " <> err
```
steve-chavez, line 361: The term in-database config is much clearer.
Perhaps the `db-load-guc-config` option should be changed to `config-in-db`?

steve-chavez:
@wolfgangwalther This should be ready for a final review.

----

These are the scenarios for the recovery tests. Hope they're useful!

### 1. postgrest started with a pg connection, then pg becomes unavailable

- worker starts only after a request to postgrest, postgrest should respond with a `{"details":"no connection to the server\n","message":"Database client error. Retrying the connection."}`
  - if db-channel-enabled=true, worker starts immediately, not necessary to prove this in tests though.
- pg becomes available, postgrest succeeds reconnecting, reloads the schema cache and responds with 200
- if db-load-guc-config=true, it should also re-read the db config.
  - test with an `ALTER ROLE postgrest_test_authenticator SET pgrst.db_schemas = 'public';` and try a `GET /public_consumers` which should give a 404 if the db config isn't re-read.

### 2. unavailable pg, postgrest started

- worker starts immediately, postgrest should respond with a `503 {"message":"Database connection lost. Retrying the connection."}`
  - Bug: if db-channel-enabled=true, postgrest doesn't reply and curl gives `Connection refused`. This must be because of the `mvarConnectionStatus` MVar, it doesn't happen on 1 though.
- pg becomes available, postgrest succeeds reconnecting, reloads the schema cache and responds with 200
- if db-load-guc-config=true, it should also re-read the db config.
  - Same test as 1

### 3. SIGUSR1 - NOTIFY reload schema

- when postgrest has a pg connection, both SIGUSR1 and NOTIFY will reload the schema cache
  - if db-load-guc-config=true, it should also re-read the db config.
- when postgrest loses the connection, and no db-channel-enabled(no NOTIFY, only SIGUSR1)
  - SIGUSR1 starts the worker, only one can run at a time. Ensured by `refIsWorkerOn`, this can be confirmed by doing several SIGUSR1 and just noting one `Attempting to reconnect to the database in 1 seconds...` message. If `refIsWorkerOn` is removed, there will be several `Attempting to reconnect to the database in 1 seconds...` mesagges.
    - Not sure how to test this, maybe count the number of threads?
  - pg becomes available, postgrest succeeds reconnecting, reloads the schema cache and responds with 200

main/Main.hs
wolfgangwalther: Hm, I see your point about `db-load-guc-config` - it's clumsy. I'd like to keep the `db-` prefix, though.

What about `db-has-config = true|false` ? Or just `db-config`.

wolfgangwalther (approved):

steve-chavez:
Deleted the previous comment. I've now opened another issue for the recovery tests #1766.

main/Main.hs
steve-chavez: The `has` looks a bit weird inside a config option. I'll go with `db-config`, it's already much better than `db-load-guc-config`.

steve-chavez: @wolfgangwalther What do you think of having `db-config` to be `false` by default

The main argument for it being `true` was to have the least amount of config options: https://github.com/PostgREST/postgrest/pull/1729#issuecomment-759564831. But that isn't possible atm, because `db-anon-role` and `db-schemas` are required.

However I'm thinking that only defining `db-uri` could be done and still maintain the required fields restriction. This by having an `DbUriAppConfig` that is used previous to `AppConfig`. With that even non-reloadable config options(like `db-pool`) could be set with the `db-config`. 

But anyway the above is still not done, so `db-config` is not an advantage now regarding few config options.

It's just that I feel `db-config=false` is right because most users will not use it(or won't be able, as mentioned on https://github.com/PostgREST/postgrest/pull/1729#issuecomment-758751747). It's a special use case. Having it by default gives potential for errors(because of the query done, twice) on all instances.

WDYT?

steve-chavez: > db-anon-role and db-schemas are required.
> This by having an DbUriAppConfig that is used previous to AppConfig

Actually, a better solution for this would be to:

- make `db-anon-role` optional. This has some value as we discussed before(some users want only JWT access).
- make `db-schemas` default to `public`. This seems like a nice default that will always work.

wolfgangwalther: > > db-anon-role and db-schemas are required.
> > This by having an DbUriAppConfig that is used previous to AppConfig
> 
> Actually, a better solution for this would be to:
> 
>     * make `db-anon-role` optional. #1689
> 
>     * make `db-schemas` default to `public`. This seems like a nice default that will always work. We can put a message on stdout to make it clear when the default is used.

Much, much better, yes!

wolfgangwalther: > It's just that I feel `db-config=false` is right because most users will not use it(or won't be able, as mentioned on [#1729 (comment)](https://github.com/PostgREST/postgrest/pull/1729#issuecomment-758751747)). It's a special use case. Having it by default gives potential for errors(because of the query done, twice) on all instances.
> 
> WDYT?

I think the risk involved is very small. There is no overhead either - even though it's a separate statement, it's basically just part of reloading the schema cache. And in the big picture, loading config options from the database is just a tiny piece of reading the whole schema.

tbh: I actually wonder whether we really need the config option at all. It could just be enabled all the time. It's probably most useful for writing tests - but that's about it.

steve-chavez: > And in the big picture, loading config options from the database is just a tiny piece of reading the whole schema.

Ah, you're totally right here! I remember trying to do this at first, but I had some issues with the code.

I think it could be done now. We'd have to enable reloading only a part of the schema cache(the guc settings) for SIGUSR2/NOTIFY. Also make the `--dump-config` wait for the `connectionWoker` to finish(maybe an MVar) and only print the config contents to stdout(no `Attempt to connect to db..` or tests fail).

> tbh: I actually wonder whether we really need the config option at all. It could just be enabled all the time. 

I think the same way now. If the guc settings are part of the schema cache, we can remove the config option.

PR #1764, opened 2021-02-27
freebsd ci: attempting to fix build failure

robx:
We were using FreeBSD 12.1, which is EOL: https://www.freebsd.org/security/#sup. (That appears like it will happen regularly within 3 months of a point release.)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image: freebsd-12-1-release-amd64
+  image: freebsd-12-2-release-amd64
 
 build_task:
   env:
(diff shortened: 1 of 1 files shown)


robx:
Hmm, this seems to be getting closer to the issue:

```
pkg update
Updating FreeBSD repository catalogue...
Fetching meta.txz: . done
Fetching packagesite.txz: .......... done
Processing entries: 
Newer FreeBSD version for package glew:
To ignore this error set IGNORE_OSVERSION=yes
- package: 1202000
- running kernel: 1201000
Ignore the mismatch and continue? [Y/n]: pkg: repository FreeBSD contains packages for wrong OS version: FreeBSD:12:amd64
Processing entries... done
Unable to update repository FreeBSD
Error updating repositories!

Exit status: 70
```

steve-chavez:
Hey @robx. Thanks a lot for looking into this :pray: .

PR #1768, opened 2021-03-07
Nix: Add a tool for checking Haskell imports and exports

monacoremo:
This uses the actual GHC parser, so it should be fairly robust. Works on both our Haskell modules and the ones generated with `-ddump-minmal-imports`.

To use:

```bash
nix-build -A haskellimports
./result src/PostgREST/App.hs
```

Example output so far:

```csv
source,moduleName,impQualified,impAs,impHiding,symbolName
src/PostgREST/App.hs,Control.Monad.Except,false,,false,liftEither
src/PostgREST/App.hs,Data.Either.Combinators,false,,false,mapLeft
src/PostgREST/App.hs,Data.IORef,false,,false,IORef
src/PostgREST/App.hs,Data.IORef,false,,false,readIORef
src/PostgREST/App.hs,Data.List,false,,false,union
src/PostgREST/App.hs,Data.Time.Clock,false,,false,UTCTime
src/PostgREST/App.hs,Data.ByteString.Char8,true,BS8,,
src/PostgREST/App.hs,Data.ByteString.Lazy,true,LBS,,
src/PostgREST/App.hs,Data.Set,true,Set,,
src/PostgREST/App.hs,Hasql.DynamicStatements.Snippet,true,SQL,,
src/PostgREST/App.hs,Hasql.Pool,true,SQL,,
src/PostgREST/App.hs,Hasql.Transaction,true,SQL,,
src/PostgREST/App.hs,Hasql.Transaction.Sessions,true,SQL,,
src/PostgREST/App.hs,Network.HTTP.Types.Header,true,HTTP,,
src/PostgREST/App.hs,Network.HTTP.Types.Status,true,HTTP,,
src/PostgREST/App.hs,Network.HTTP.Types.URI,true,HTTP,,
src/PostgREST/App.hs,Network.Wai,true,Wai,,
src/PostgREST/App.hs,PostgREST.ApiRequest,true,ApiRequest,,
src/PostgREST/App.hs,PostgREST.Auth,true,Auth,,
src/PostgREST/App.hs,PostgREST.DbRequestBuilder,true,ReqBuilder,,
src/PostgREST/App.hs,PostgREST.DbStructure,true,DbStructure,,
src/PostgREST/App.hs,PostgREST.Error,true,Error,,
src/PostgREST/App.hs,PostgREST.Middleware,true,Middleware,,
src/PostgREST/App.hs,PostgREST.OpenAPI,true,OpenAPI,,
src/PostgREST/App.hs,PostgREST.QueryBuilder,true,QueryBuilder,,
src/PostgREST/App.hs,PostgREST.RangeQuery,true,RangeQuery,,
src/PostgREST/App.hs,PostgREST.Statements,true,Statements,,
src/PostgREST/App.hs,PostgREST.ApiRequest,false,,false,Action
src/PostgREST/App.hs,PostgREST.ApiRequest,false,,false,ApiRequest
src/PostgREST/App.hs,PostgREST.ApiRequest,false,,false,InvokeMethod
src/PostgREST/App.hs,PostgREST.ApiRequest,false,,false,Target
src/PostgREST/App.hs,PostgREST.Config,false,,false,AppConfig
src/PostgREST/App.hs,PostgREST.Error,false,,false,Error
src/PostgREST/App.hs,PostgREST.Types,false,,,
src/PostgREST/App.hs,Protolude,false,,true,Handler
src/PostgREST/App.hs,Protolude,false,,true,toS
src/PostgREST/App.hs,Protolude.Conv,false,,false,toS
```

Example usage in nix/devtools.nix, WIP.

@wolfgangwalther : This might be useful for many automated checks and analysis on our imports. Any comments or ideas?

### DIFF
--- a/nix/hsie/Main.hs
+++ b/nix/hsie/Main.hs
@@ -14,12 +14,12 @@
 
 module Main (main) where
 
-import qualified Data.Aeson                              as JSON
-import qualified Data.ByteString.Lazy.Char8              as LBS8
-import qualified Data.Csv                                as Csv
-import qualified Data.Map                                as Map
-import qualified Data.Set                                as Set
-import qualified Data.Text                               as T
+import qualified Data.Aeson                 as JSON
+import qualified Data.ByteString.Lazy.Char8 as LBS8
+import qualified Data.Csv                   as Csv
+import qualified Data.Map                   as Map
+import qualified Data.Set                   as Set
+import qualified Data.Text                  as T
 import qualified Data.Text.IO                            as T
 import qualified Dot
 import qualified GHC
@@ -79,7 +79,7 @@ data ImportedSymbol =
     , impModule     :: Text
     , impQualified  :: ImportQualified
     , impAs         :: Maybe Text
-    , impHiding     :: ImportHiding
+    , impType       :: ImportType
     , impSymbol     :: Maybe Text
     }
     deriving (Generic, Csv.ToNamedRecord, Csv.DefaultOrdered, JSON.ToJSON)
@@ -93,16 +93,16 @@ instance Csv.ToField ImportQualified where
   toField Qualified    = "qualified"
   toField NotQualified = "not qualified"
 
-data ImportHiding
+data ImportType
   = Wildcard
   | Hiding
-  | NotHiding
+  | Explicit
   deriving (Eq, Generic, JSON.ToJSON)
 
-instance Csv.ToField ImportHiding where
-  toField Wildcard  = "wildcard"
-  toField Hiding    = "hiding"
-  toField NotHiding = "not hiding"
+instance Csv.ToField ImportType where
+  toField Wildcard = "wildcard"
+  toField Hiding   = "hiding"
+  toField Explicit = "explicit"
 
 -- | Mapping of modules to their aliases and to the files they are found in
 type ModuleAliases = [(Text, [(Text, [Text])])]
@@ -222,7 +222,7 @@ importSymbols _ _  (GHC.XImportDecl _) = mempty
 importSymbols source filepath GHC.ImportDecl{..} =
   case ideclHiding of
     Just (hiding, syms) ->
-      symbol (if hiding then Hiding else NotHiding) . Just . GHC.unLoc <$> GHC.unLoc syms
+      symbol (if hiding then Hiding else Explicit) . Just . GHC.unLoc <$> GHC.unLoc syms
     Nothing ->
       [ symbol Wildcard Nothing ]
   where
@@ -234,7 +234,7 @@ importSymbols source filepath GHC.ImportDecl{..} =
         , impModule = T.pack . moduleNameString . GHC.unLoc $ ideclName
         , impQualified = if ideclQualified then Qualified else NotQualified
         , impAs = T.pack . moduleNameString . GHC.unLoc <$> ideclAs
-        , impHiding = hiding
+        , impType = hiding
         , impSymbol = T.pack . occNameString . rdrNameOcc . GHC.ieName <$> sym
         }
     moduleFromPath =
@@ -301,7 +301,7 @@ wildcards okModules =
   groupByFile . filter isWildcard . filter (not . isOkModule)
   where
     isWildcard ImportedSymbol{..} =
-      impQualified == NotQualified && impHiding /= NotHiding
+      impQualified == NotQualified && impType /= Explicit
     isOkModule = flip Set.member (Set.fromList okModules) . impModule
     groupByFile = Map.toList . fmap Set.toList . foldr insertMap Map.empty
     insertMap ImportedSymbol{..} =
--- a/nix/hsie/Main.hs
+++ b/nix/hsie/Main.hs
@@ -14,12 +14,12 @@
 
 module Main (main) where
 
-import qualified Data.Aeson                 as JSON
-import qualified Data.ByteString.Lazy.Char8 as LBS8
-import qualified Data.Csv                   as Csv
-import qualified Data.Map                   as Map
-import qualified Data.Set                   as Set
-import qualified Data.Text                  as T
+import qualified Data.Aeson                              as JSON
+import qualified Data.ByteString.Lazy.Char8              as LBS8
+import qualified Data.Csv                                as Csv
+import qualified Data.Map                                as Map
+import qualified Data.Set                                as Set
+import qualified Data.Text                               as T
 import qualified Data.Text.IO                            as T
 import qualified Dot
 import qualified GHC
@@ -27,7 +27,6 @@ import qualified Language.Haskell.GHC.ExactPrint.Parsers as ExactPrint
 import qualified Options.Applicative                     as O
 import qualified System.FilePath                         as FP
 
-import Control.Applicative        ((<|>))
 import Data.Aeson.Encode.Pretty   (encodePretty)
 import Data.Function              ((&))
 import Data.List                  (intercalate)
@@ -46,21 +45,8 @@ import System.Exit                (exitFailure)
 data Options =
   Options
     { command :: Command
-    , sources :: [Source]
+    , sources :: [FilePath]
     }
-
-data Source
-  = SourceDir FilePath
-  | ImportsDir FilePath
-  deriving (Generic, JSON.ToJSON)
-
-sourceDir :: Source -> FilePath
-sourceDir (SourceDir path)  = path
-sourceDir (ImportsDir path) = path
-
-instance Csv.ToField Source where
-  toField (SourceDir _)  = "source"
-  toField (ImportsDir _) = "imports"
 
 data Command
   = Dump OutputFormat
@@ -73,14 +59,14 @@ data OutputFormat = OutputCsv | OutputJson
 
 data ImportedSymbol =
   ImportedSymbol
-    { impSourceFile :: Text
-    , impSourceType :: Source
-    , impFromModule :: Text
+    { impFromModule :: Text
     , impModule     :: Text
     , impQualified  :: ImportQualified
     , impAs         :: Maybe Text
     , impType       :: ImportType
     , impSymbol     :: Maybe Text
+    , impSource     :: FilePath
+    , impFile       :: FilePath
     }
     deriving (Generic, Csv.ToNamedRecord, Csv.DefaultOrdered, JSON.ToJSON)
 
@@ -105,10 +91,10 @@ instance Csv.ToField ImportType where
   toField Explicit = "explicit"
 
 -- | Mapping of modules to their aliases and to the files they are found in
-type ModuleAliases = [(Text, [(Text, [Text])])]
+type ModuleAliases = [(Text, [(Text, [FilePath])])]
 
 -- | Mapping of modules to files
-type WildcardImports = [(Text, [Text])]
+type WildcardImports = [(FilePath, [Text])]
 
 
 -- MAIN
@@ -124,24 +110,15 @@ main =
         <> O.header "hsie - Swiss army knife for HaSkell Imports and Exports"
         <> O.progDesc "Parse Haskell code to analyze it's imports and exports"
     opts =
-      Options <$> commandOption <*> O.some (srcOption <|> importsOption)
+      Options <$> commandOption <*> O.some srcOption
     srcOption =
-      O.option (SourceDir <$> O.str) $
-        O.long "src"
-        <> O.short 's'
-        <> O.metavar "SRCDIR"
+      O.argument O.str $
+        O.metavar "SRCDIR"
         <> O.help "Haskell source directory"
-        <> O.action "directory"
-    importsOption =
-      O.option (ImportsDir <$> O.str) $
-        O.long "imports"
-        <> O.short 'i'
-        <> O.metavar "IMPORTSDIR"
-        <> O.help "Imports directory generated with GHC's '--ddump-minimal-imports'"
         <> O.action "directory"
     commandOption =
       O.subparser $
-        command "dump" "Dump imported symbols as CSV or JSON"
+        command "dump-imports" "Dump imported symbols as CSV or JSON"
           (Dump <$> jsonOutputFlag)
         <> command "graph-modules" "Print dot graph of module imports"
              (pure GraphModules)
@@ -188,14 +165,9 @@ run Options{command, sources} =
 -- SYMBOLS
 
 -- | Parse all imported symbols from a source of Haskell source files
-sourceSymbols :: Source -> IO [ImportedSymbol]
+sourceSymbols :: FilePath -> IO [ImportedSymbol]
 sourceSymbols source = do
-  files <-
-    case source of
-      SourceDir dirpath ->
-        filterExts [".hs"] <$> getFilesRecursive dirpath
-      ImportsDir dirpath ->
-        filterExts [".imports"] <$> getFilesRecursive dirpath
+  files <- filterExts [".hs", "imports"] <$> getFilesRecursive source
   concat <$> mapM moduleSymbols files
   where
     filterExts exts = filter $ flip elem exts . FP.takeExtension
@@ -217,7 +189,7 @@ parseModule filepath = do
 --
 -- If the import is a wildcard, i.e. no symbols are selected for import, then
 -- only one item is returned.
-importSymbols :: Source -> FilePath -> GHC.ImportDecl GhcPs -> [ImportedSymbol]
+importSymbols :: FilePath -> FilePath -> GHC.ImportDecl GhcPs -> [ImportedSymbol]
 importSymbols _ _  (GHC.XImportDecl _) = mempty
 importSymbols source filepath GHC.ImportDecl{..} =
   case ideclHiding of
@@ -228,8 +200,8 @@ importSymbols source filepath GHC.ImportDecl{..} =
   where
     symbol hiding sym =
       ImportedSymbol
-        { impSourceFile = T.pack filepath
-        , impSourceType = source
+        { impFile = relativePath filepath
+        , impSource = source
         , impFromModule = T.pack $ moduleFromPath filepath
         , impModule = T.pack . moduleNameString . GHC.unLoc $ ideclName
         , impQualified = if ideclQualified then Qualified else NotQualified
@@ -239,7 +211,7 @@ importSymbols source filepath GHC.ImportDecl{..} =
         }
     moduleFromPath =
       intercalate "." . FP.splitDirectories . FP.dropExtension . relativePath
-    relativePath = FP.makeRelative $ sourceDir source
+    relativePath = FP.makeRelative source
 
 
 -- DUMP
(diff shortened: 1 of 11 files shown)


wolfgangwalther:
That's really cool!

Some ideas we can use this for:
- Automated check that no modules have different aliases between files (consistent aliases across the projects would increase readability)
- Automated check for certain rules regarding unqualified imports, aliases etc. - would have to make up those rules first, though.
- Manually use with `-ddump-minmal-imports` to find a set of commonly used imports across different files, that we could pack into our own prelude/protolude.
- Manually use with `-ddump-minmal-imports` to find functions in the `PostgREST` namespace, that are only imported from 1 file. Could be an indication that those functions could better be moved to the importing file.
- Maybe we can develop some kind of simple metric to "count" the imports within our namespace - to easily check the "complexity" of imports between different files. The goal would be to reduce this complexity of course, so that our dependency graphs are straight-forward to read, too.

WDYT?

monacoremo:
Those are great use cases! Thinking a bit about the best way to implement them, it will probably be easier to express them directly in Haskell instead of fiddling with the dump using nix/bash/python etc. We could even build a small tool around this with optparse-applicative:

```
hsimps dump --src ./src --imports /tmp/imports -s main -i imports2 Mod.hs Mod2.hs # dump imports as CSV (or Json with --json?)

hsimps graph-symbols … # print a dot file showing which symbols are imported from which modules

hsimps graph-modules … # print a dot file showing module dependencies (would replace the tooling I added earlier)

hsimps check-aliases … # check that aliases are consistent

hsimps check-wildcards … # check that imports are either qualified or explicit

…
```

Where 

* --src/-s includes a source directory
* --imports/-i includes an imports directory created with -ddump-minimal-imports
* Individual files (zero or more) can also be included, module name is parsed from file if available

This would help getting the module names right based on the filenames (as the module declaration is not included in the ddump, for example)

The nix/haskellimports/Main.hs would grow a bit, but it should be low maintenance and simplify everything else a lot. Does that make sense?

wolfgangwalther:
> Those are great use cases! Thinking a bit about the best way to implement them, it will probably be easier to express them directly in Haskell instead of fiddling with the dump using nix/bash/python etc. We could even build a small tool around this with optparse-applicative:
> [...]
> The nix/haskellimports/Main.hs would grow a bit, but it should be low maintenance and simplify everything else a lot. Does that make sense?

Yes. Sounds better than dumping, reloading the file in python etc.

> `hsimps dump --src ./src --imports /tmp/imports -s main -i imports2 Mod.hs Mod2.hs # dump imports as CSV (or Json with --json?)`

Personally, I like CSV more here. I can load this directly into Calc/Excel and play around.

> --imports/-i includes an imports directory created with -ddump-minimal-imports

Not sure if that's possible with `GHC` as a lib - but can we remove the extra step of doing a dump to a tmp directory? Maybe just have a flag `--use-minimal-imports` or something like that to make the script operate on the output of that. If the flag is not given it operates on the "raw" imports in the source files.

monacoremo:
Still work in progress, but the basic idea works:

```bash
[nix-shell]$ hsie check-aliases --src src
The following imports have inconsistent aliases:

Module 'Data.ByteString' has the aliases:
  'B' in file:
    src/PostgREST/Config.hs
  'BS' in file:
    src/PostgREST/ApiRequest.hs
    src/PostgREST/Types.hs

Module 'Data.ByteString.Char8' has the aliases:
  'BS' in files:
    src/PostgREST/Config.hs
    src/PostgREST/Middleware.hs
    src/PostgREST/Private/QueryFragment.hs
    src/PostgREST/QueryBuilder.hs
    src/PostgREST/RangeQuery.hs
    src/PostgREST/Statements.hs
  'BS8' in file:
    src/PostgREST/App.hs

Module 'Data.ByteString.Lazy' has the aliases:
  'BL' in files:
    src/PostgREST/ApiRequest.hs
    src/PostgREST/Private/QueryFragment.hs
    src/PostgREST/Types.hs
  'LBS' in file:
    src/PostgREST/App.hs
    src/PostgREST/OpenAPI.hs

Module 'Data.HashMap.Strict' has the aliases:
  'HM' in file:
    src/PostgREST/Private/QueryFragment.hs
  'HashMap' in file:
    src/PostgREST/OpenAPI.hs
  'M' in files:
    src/PostgREST/ApiRequest.hs
    src/PostgREST/Auth.hs
    src/PostgREST/DbRequestBuilder.hs
    src/PostgREST/DbStructure.hs
    src/PostgREST/Middleware.hs
    src/PostgREST/Parsers.hs
    src/PostgREST/Types.hs

Module 'Data.Set' has the aliases:
  'S' in files:
    src/PostgREST/ApiRequest.hs
    src/PostgREST/DbRequestBuilder.hs
    src/PostgREST/DbStructure.hs
    src/PostgREST/Parsers.hs
    src/PostgREST/QueryBuilder.hs
    src/PostgREST/Types.hs
  'Set' in file:
    src/PostgREST/App.hs

Module 'Hasql.DynamicStatements.Snippet' has the aliases:
  'H' in files:
    src/PostgREST/Private/Common.hs
    src/PostgREST/Private/QueryFragment.hs
    src/PostgREST/QueryBuilder.hs
    src/PostgREST/Statements.hs
  'SQL' in file:
    src/PostgREST/App.hs

Module 'Hasql.Pool' has the aliases:
  'P' in file:
    src/PostgREST/Error.hs
  'SQL' in file:
    src/PostgREST/App.hs

Module 'Hasql.Transaction' has the aliases:
  'H' in file:
    src/PostgREST/Middleware.hs
  'HT' in file:
    src/PostgREST/DbStructure.hs
  'SQL' in file:
    src/PostgREST/App.hs

Module 'Network.HTTP.Types.Status' has the aliases:
  'HT' in file:
    src/PostgREST/Error.hs
  'HTTP' in file:
    src/PostgREST/App.hs

```

monacoremo:
Current outputs from `hsie dump`:
https://gist.github.com/monacoremo/a0b8277f9baa45238ca36610d924ecaf

monacoremo:
> Personally, I like CSV more here. I can load this directly into Calc/Excel and play around.

We get both! The `--json` flag is optional, CSV is the default

> Not sure if that's possible with GHC as a lib - but can we remove the extra step of doing a dump to a tmp directory? Maybe just have a flag --use-minimal-imports or something like that to make the script operate on the output of that. If the flag is not given it operates on the "raw" imports in the source files.

That's extremely tricky, as we would need to run the whole compiler machinery, compile all dependencies etc. I'll add a bash wrapper script that provides the `--ddump-minimal-imports` as an input

wolfgangwalther:
Just skimmed through the source code: It doesn't look like the exit code is any different for `check-aliases`, when there are any errors. Can we get a non-zero exit code here to possibly break tests?

monacoremo:
> Just skimmed through the source code: It doesn't look like the exit code is any different for check-aliases, when there are any errors. Can we get a non-zero exit code here to possibly break tests?

Yes, great idea! Added.

Also added `check-wildcards`:

```
[nix-shell]$ hsie check-wildcards --ok Protolude --src src
Modules in the following files were imported as wildcards:

In src/PostgREST/ApiRequest.hs:
  Data.Ranged.Boundaries
  PostgREST.Types

In src/PostgREST/App.hs:
  PostgREST.Types

In src/PostgREST/Config.hs:
  Control.Applicative
  Data.Monoid
  Options.Applicative

In src/PostgREST/DbRequestBuilder.hs:
  Control.Applicative
  Data.Tree
  PostgREST.Parsers
  PostgREST.Types

In src/PostgREST/DbStructure.hs:
  PostgREST.Private.Common
  PostgREST.Types

In src/PostgREST/Error.hs:
  Network.HTTP.Types.Header
  PostgREST.Types

In src/PostgREST/Middleware.hs:
  Network.Wai
  Network.Wai.Middleware.RequestLogger
  PostgREST.Private.Common

In src/PostgREST/OpenAPI.hs:
  Control.Lens
  Data.Swagger

In src/PostgREST/Parsers.hs:
  Data.Tree
  PostgREST.Types
  Text.Parsec.Error
  Text.ParserCombinators.Parsec

In src/PostgREST/Private/Common.hs:
  Data.Maybe

In src/PostgREST/Private/QueryFragment.hs:
  Data.Maybe
  PostgREST.Private.Common
  PostgREST.Types

In src/PostgREST/QueryBuilder.hs:
  Data.Maybe
  PostgREST.Private.Common
  PostgREST.Private.QueryFragment
  PostgREST.Types

In src/PostgREST/RangeQuery.hs:
  Control.Applicative
  Data.Ranged.Boundaries
  Data.Ranged.Ranges
  Network.HTTP.Types.Header
  Network.HTTP.Types.Status

In src/PostgREST/Statements.hs:
  Data.Aeson
  Data.Maybe
  Network.HTTP.Types.Status
  PostgREST.Error
  PostgREST.Private.Common
  PostgREST.Private.QueryFragment
  PostgREST.Types

In src/PostgREST/Types.hs:
  Data.Tree

```

Nearly done with this :-)

monacoremo:
Left for later PRs:

* CI integration, will need to fix actual issues flagged by `check-aliases` and `check-wildcards`. For now, you can run `hsie check-aliases -s src -s main` in `nix-shell`
* Analyzing exports (e.g., wildcard exports, unused exported symbols)

nix/hsie/Main.hs
```diff
@@ -0,0 +1,349 @@
...
+        O.long "json" <> O.short 'j' <> O.help "Output JSON"
+    okModuleOption =
+      O.option O.auto $
+        O.long "ok"
+        <> O.short 'o'
+        <> O.metavar "OKMODULE"
+        <> O.help "Module that is ok to import as wildcard"
+
+run :: Options -> IO ()
+run Options{..} =
```
wolfgangwalther, line 169: In this case, where the fields don't have a prefix, I find it much more readable to use `NamedFieldPuns` instead of `RecordWildcards`:

```suggestion
run Options{command, sources} =
```

nix/hsie/Main.hs
```diff
@@ -0,0 +1,370 @@
...
+
+module Main (main) where
+
+import qualified Data.Aeson                 as Aeson
+import qualified Data.ByteString.Lazy.Char8 as LBS8
+import qualified Data.Csv                   as Csv
+import qualified Data.Map                   as Map
+import qualified Data.Set                   as Set
+import qualified Data.Text                  as T
+--import qualified Data.Text.Lazy.Builder                  as T
```
wolfgangwalther, line 22: Seems like this can be fully removed?

wolfgangwalther (commented):
I didn't have the chance, yet, to actually run the tools, but will do so soon.

Doing the review, I will admit that I didn't go through the meat of the hsie functions in detail. The code looks very organized and well readable, though.

I thought briefly about possibly renaming the `dump` command, because I was initially a bit confused when reading the part about `dump-minimal-imports`. But `dump` is probably the best way to describe it...

nix/hsie/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+hsie dump -j -s src -s main
+```
+
+If you dumped minimal imports into the `imports` directory using GHC's
+`--ddump-minimal-imports`, you can analyse them using `--imports` or `-i`:
+
+```
+hsie dump -i imports
+```
```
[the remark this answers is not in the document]
wolfgangwalther: Reading the source, I now know the difference is in the file extension. I wonder whether we need to make that distinction, or could just take both extensions all the time and make this only `-s/--src`.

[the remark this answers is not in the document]
monacoremo: It's also how we derive the module name from the file path - although the function I came up with happens to handle both. The distinction should be useful later, e.g. when analyzing exports, which are missing in the dump, so I suggest we leave it in for now.

nix/hsie/Main.hs
monacoremo: yes!

monacoremo:
Many thanks for your review @wolfgangwalther !

> I thought briefly about possibly renaming the dump command, because I was initially a bit confused when reading the part about dump-minimal-imports. But dump is probably the best way to describe it...

Maybe `dump-imports`? Could make sense once we also analyze exports.

wolfgangwalther:
> Maybe `dump-imports`? Could make sense once we also analyze exports.

I don't think I would have been less confused. But if we plan to have `dump-exports`, then that might make sense. But now `dump-imports --imports xxx` really is confusing :).

While trying the tools (see below), I thought about this a bit more and I think I now know where my confusion is coming from. When we look at everything that is done with `--ddump-minimal-imports` we are actually looking at **usage** data. So we have 3 parts to look at:
- (real) import of symbols
- export of symbols
- usage of symbols

This will also be important later on, when you want to analyze exports - because you actually don't care about imports but usage of those.

I don't think it ever makes sense to dump imports (`--src`) and usage (`--import`) at the same time. I suggest the following interface, that should be a lot less confusing **for me**:
```bash
hsie dump-imports main src
hsie dump-exports main src
hsie dump-usage usage # assume --ddump-minimal-imports is saved in "usage"
```
Passing the folders to work on as arguments also avoids the ambiguity between `--src` and `--imports`. Basically abstract the whole `--ddump-minimal-imports` away as much as possible, except for mentioning it in the readme as the input to `dump-usage`.

This takes away the possibility of running the `check-` or `graph-` against the usage data. But I really don't think this is useful anyway - running those against the imports made does give us a good picture of how our code-base is set up.

---

I just tried the tools and noted the following:
- Running `postgrest-hsie-minimal-import dump`, I get output like this:
  ```csv
  impSourceFile,impSourceType,impFromModule,impModule,impQualified,impAs,impHiding,impSymbol
  /run/user/1000/tmp.TmIMph9zFs/UnixSocket.imports,imports,UnixSocket,Network.Socket,not qualified,,not hiding,Family
  ....
  ```
  The first column is not really useful, I think - especially with the temporary path. Since `impFromModule` is taken from the filename, I don't think we lose anything by removing `impSourceFile` from the output. At least the first thing I do, when loading the CSV is removing the column :).
- If possible, another column to tell whether the symbol is internal (aka within our code-base) or external, would be helpful. I can filter by `PostgREST.` prefix, but have to manually include `UnixSocket` all the time. This would also allow to add this as a cli flag.
- Assuming the interface is changed as suggested above `postgrest-hsie-minimal-import` could be renamed `postgrest-hsie-dump-usage` and **include** the `dump-usage` subcommand (+ folder to work on) already.
- The `impHiding` column naming is a bit off, when it includes "wildcard". Maybe something like this:
  ```haskell
  data ImportType
    = Wildcard
    | Explicit
    | Hiding
  ```

monacoremo:
> This takes away the possibility of running the check- or graph- against the usage data. But I really don't think this is useful anyway - running those against the imports made does give us a good picture of how our code-base is set up.

It should be useful e.g. when we analyze exports - we need the full internal imports from ddump-minimal-imports, and the exports from the actual source code. To graph the full symbol imports, we also need ddump-minimal-imports.

But agree that having `-i` and `-s` separated does not add value at this point, so I removed it and commands like `hsie dump-imports src main` work (as do `hsie src dump-imports main` for that matter based on the optparse-applicative magic..).

> The first column is not really useful, I think - especially with the temporary path. Since impFromModule is taken from the filename, I don't think we lose anything by removing impSourceFile from the output. At least the first thing I do, when loading the CSV is removing the column :).

We need it to display the file that issues are found in, e.g. with `check-aliases`. Hiding them would add a bit of complexity which I'd rather avoid. I moved the fields to the end of the record, where they are less obtrusive hopefully!

> If possible, another column to tell whether the symbol is internal (aka within our code-base) or external, would be helpful. I can filter by PostgREST. prefix, but have to manually include UnixSocket all the time. This would also allow to add this as a cli flag.

Great idea, and needed anyway e.g. for the graphing commands. Added.

> Assuming the interface is changed as suggested above postgrest-hsie-minimal-import could be renamed postgrest-hsie-dump-usage and include the dump-usage subcommand (+ folder to work on) already.

`dump-usage` would be redundant, as `dump-imports` works as-is. Suggest that we keep the minimal-imports wording from GHC, as it's for specialized use and we avoid introducing new terminology. `postgrest-hsie-minimal-import` works with any command as is.

> The impHiding column naming is a bit off, when it includes "wildcard". Maybe something like this: 

Yes, much better! That's one up from what our GHC friends came up with ([`ideclHiding`](http://hackage.haskell.org/package/ghc-8.10.1/docs/GHC-Hs-ImpExp.html#t:ImportDecl), which was super confusing for me :-) )

nix/hsie/Main.hs
```diff
@@ -0,0 +1,358 @@
...
+          (Dump <$> jsonOutputFlag)
+        <> command "graph-modules" "Print dot graph of module imports"
+             (pure GraphModules)
+        <> command "graph-symbols" "Print dot graph of symbol imports"
+             (pure GraphSymbols)
+        <> command "check-aliases"
+             "Check that aliases of imported modules are consistent"
+             (pure CheckAliases)
+        <> command "check-wildcards"
+             "Check that no unwanted modules are imported as wildcards"
```
wolfgangwalther, line 141:
```suggestion
             "Check that no modules are imported as unqualified wildcards"
```

I think the `--ok` explains itself from using it, but the default is no modules at all.

nix/hsie/Main.hs
```diff
@@ -0,0 +1,358 @@
...
+      O.command name . O.info (O.helper <*> options) $ O.progDesc desc
+    jsonOutputFlag =
+      O.flag OutputCsv OutputJson $
+        O.long "json" <> O.short 'j' <> O.help "Output JSON"
+    okModuleOption =
+      O.option O.auto $
+        O.long "ok"
+        <> O.short 'o'
+        <> O.metavar "OKMODULE"
+        <> O.help "Module that is ok to import as wildcard"
```
wolfgangwalther, line 153:
```suggestion
        <> O.help "Module that is ok to import as unqualified wildcard"
```

shell.nix
```diff
@@ -47,6 +48,7 @@ lib.overrideDerivation postgrest.env (
         complete -F _command postgrest-with-postgresql-10
         complete -F _command postgrest-with-postgresql-9.6
         complete -F _command postgrest-with-postgresql-9.5
+        source ${postgrest.hsie.bashCompletion}
```
wolfgangwalther, line 51: This looks like I should be getting bash completion for `hsie`. But it doesn't work for me. I can only complete filenames, but not commands etc. - is this supposed to work?

wolfgangwalther (commented):
Just tried `check-wildcards` and it gives me an error:

```bash
$ hsie check-wildcards --ok Protolude src main
option --ok: cannot parse value `Protolude'

Usage: hsie check-wildcards [-o|--ok OKMODULE]
  Check that no unwanted modules are imported as wildcards
```

---

Just noting for now:
There is also a reverse approach for `check-aliases`. Some aliases are used for different modules in different files. Example:
| file | import | alias |
| --- | --- | --- |
| PostgREST.Config | Data.Configurator | C |
| Main | Hasql.Connection | C |

This is probably not something that we can auto-detect easily, because there could be lots of false-positives, too. Think the different Hasql packages - could all be imported with `SQL.`.

shell.nix
wolfgangwalther: I played around with the completion script's content a bit and I get this:
```bash
$ hsie --bash-completion-index dump-imports
option --bash-completion-index: cannot parse value `dump-imports'
```

Not sure whether that's the right way to call the completion-index subcommand - but the error looks like it might be related to the error I get for `check-wildcards`?

wolfgangwalther: Ah, this is because it needs a number, not a string. `--bash-completion-index 0` etc. works.

monacoremo:
> Just tried check-wildcards and it gives me an error:

Good catch - not sure when that broke. Fixed! 

We're reaching a point where we need a test suite, might be worth to spin this off at some point.

shell.nix
monacoremo: Should work with the latest commit, now that it completes for the right executable name... :-)

wolfgangwalther (approved):

monacoremo:
Many thanks for your review @wolfgangwalther !

PR #1770, opened 2021-03-13
fix tests to be independent of JSON ordering

robx:
This is prompted by a report that postgrest tests fail with newest hashable due to unstable order in aeson output: https://github.com/commercialhaskell/stackage/issues/5878. Stackage doesn't actually run postgrest tests, so this change isn't required, but the underlying issue *does* cause test failures.

The PR addresses two issues:
- cryptonite-0.27 build failure on the legacy circleci base image
- test failures due to changed JSON ordering with hashable-1.3.1.0

There's a bit more detail in the commit messages.

### DIFF
--- a/test/SpecHelper.hs
+++ b/test/SpecHelper.hs
@@ -8,7 +8,7 @@ import qualified Data.Set               as S
 import qualified System.IO.Error        as E
 
 import Data.Aeson           (Value (..), decode, encode)
-import Data.CaseInsensitive (CI (..))
+import Data.CaseInsensitive (CI (..), original)
 import Data.List            (lookup)
 import Data.List.NonEmpty   (fromList)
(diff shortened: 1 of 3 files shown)


test/SpecHelper.hs
```diff
@@ -33,6 +33,12 @@ matchContentTypeJson = "Content-Type" <:> "application/json; charset=utf-8"
 matchContentTypeSingular :: MatchHeader
 matchContentTypeSingular = "Content-Type" <:> "application/vnd.pgrst.object+json; charset=utf-8"
 
+matchHeaderAbsent :: HeaderName -> MatchHeader
+matchHeaderAbsent name = MatchHeader $ \headers _body ->
```
steve-chavez, line 37: Nice! Makes the tests clearer :+1:

steve-chavez (approved):
LGTM! Nice commit messages :100:

PR #1778, opened 2021-03-26
cirrusci: update FreeBSD pkg

drTr0jan:
Updated CI for using PostgreSQL 12 (as a default FreeBSD version) instead 9.5 (as an EOS version).


steve-chavez (approved):
Cool! Thanks a lot @drTr0jan.

PR #1795, opened 2021-04-07
Void functions return null instead of empty body

### LINKED ISSUE #1762: void functions return an empty body, which is invalid json
### Environment

* PostgREST version: nightly

### Description of issue

The nightly version returns an empty body for a function like:

```sql
create function test.ret_void() returns void as '' language sql;
```

This breaks libraries that use `JSON.parse`: https://github.com/supabase/supabase/issues/762 (try doing `JSON.parse("")` on the browser console).
It also doesn't seem right from the [RFC](https://tools.ietf.org/html/rfc7159#section-1) standpoint:

> JSON can represent four primitive types (strings, numbers, booleans, and null) and two structured types (objects and arrays)

Since we're asking for `application/json`, we should return a `null` or an empty string `""` at worst.

I've also noted this happens when you want to return a `null` explicit:

```sql
create or replace function ret_null() returns int as $$
  select null::int;
$$ language sql;
```

```bash
curl localhost:3000/rpc/ret_null
## no body
```

laurenceisla:
Fixes #1762. Includes a test for functions returning an integer with null value.

### DIFF
--- a/src/PostgREST/Private/QueryFragment.hs
+++ b/src/PostgREST/Private/QueryFragment.hs
@@ -98,7 +98,7 @@ asJsonF returnsScalar
 
 asJsonSingleF :: Bool -> SqlFragment --TODO! unsafe when the query actually returns multiple rows, used only on inserting and returning single element
 asJsonSingleF returnsScalar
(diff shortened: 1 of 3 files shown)


steve-chavez (approved):
@laurenceisla Nice one! The PR is on point :+1:

PR #1805, opened 2021-04-11
nix: Add postgrest-changelog-* tools

wolfgangwalther:
Data is taken from `git log`, formatted as YAML and then parsed with yq. The JSON output is then parsed for author and reference data.

Still WIP. Todo:
- [x] Parse `git log` as JSON.
- [ ] Add `postgrest-changelog` to pass json data to template engine
- [ ] Add changelog template for GitHub release page
- [ ] Add changelog template for docs (in postgrest-docs)
- [ ] Add special handling for v8 (including the data from `CHANGELOG.md` for both templates)
- [ ] Add auto-generated changelog to `postgrest-release-github`
- [ ] Add proper bash completion

### DIFF
--- a/changelog/HEAD.yaml
+++ b/changelog/HEAD.yaml
@@ -1,0 +1,972 @@
+
+_current: HEAD
+
+_previous: v7.0.1
+
+222a530:
+  author: steve-chavez
+  subject: Fix Dockerfile
+  type: other
+
+67344c8:
+  author: steve-chavez
+  subject: 'cabal: fix upload to hackage (#1532)'
+  refs:
+  - 1532
+  type: other
+
+289bb66:
+  author: monacoremo
+  subject: Reduce size of the static docker image by patching the Nix openssl package (#1528)
+  refs:
+  - 1528
+  type: other
+
+e6874c8:
+  author: monacoremo
+  subject: Fix UNKNOWN being shown as git commit hash (#1533)
+  refs:
+  - 1533
+  type: other
+
+48c9ac3:
+  author: monacoremo
+  subject: Bump nixpkgs and simplify/clean up nix setup (#1529)
+  refs:
+  - 1529
+  type: other
+
+3a1844e:
+  author: monacoremo
+  subject: Fix the static build (#1534)
+  refs:
+  - 1534
+  type: other
+
+d4aba5c:
+  author: steve-chavez
+  subject: Add notice about single Linux static executable
+  type: other
+
+08186ea:
+  author: monacoremo
+  subject: Nixify CircleCI setup (#1535)
+  refs:
+  - 1535
+  type: other
+
+784ebe5:
+  author:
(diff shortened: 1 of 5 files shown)


wolfgangwalther:
The idea is to have a two-step process:

1. First the `git log` history is transformed to a yaml document. This document is stored in the repo in `changelog/vX.Y.Z.yaml` and allows manual corrections of changelog entries, categories, etc. - and also to collaborate on it in a PR. Doing this step in YAML allows to easily render the final changelog to different formats (github releases, docs) without copy-editing multiple files manually. It should also make it easy to add old `CHANGELOG` info for the transition phase.

2. When a release is made, the corresponding yaml file is automatically rendered to the github release template and uploaded there. The same thing, but fully automated without the intermediate yaml file, will be done for a nightly build, to keep the "release a nightly" simple enough. For a stable release, the same changelog can then be manually rendered with the corresponding template from `postgrest-docs`. The result will be a `.rst` file for the docs - this might be manually edited afterwards to add references in the docs, but that should be about it.


The category of the commit (fix, feat, breaking change) is parsed from the commit message:
  * `fix` and `feat` from prefix of subject
  * `breaking` from a `BREAKING CHANGE` anywhere in the message (subject + body)
  * all other are categorized as `other` and are not rendered in the final changelog

For the case that we forget to properly mark a commit and need to adjust it, we can edit the yaml file after step 1. Manual changes to the changelog entries in the yaml file will be kept, even if `postgrest-changelog` is run again for the same version - this will then just add any new commits to the file. This will allow us to use a regular PR + rebase workflow, when collaborating on the changelog for a new version.

There is a sample changelog yaml file added to this PR in `changelog/HEAD.yaml`. This is the output when I currently run it on this branch. It's obviously a lot of commits since our last release.

steve-chavez:
@wolfgangwalther Check this tool: https://github.com/orhun/git-cliff. I think it's exactly what we need for this PR.

wolfgangwalther:
> @wolfgangwalther Check this tool: https://github.com/orhun/git-cliff. I think it's exactly what we need for this PR.

Had a first look at it. It looks nice and might be able to replace a part of the workflow. However, I think, the challenge is that I'd like to support a process where:
- we can easily generate multiple different changelogs from one source of truth
- have a 2-step process, where the actual source of truth is not the git history itself, but a yaml file that is checked in. This is because, while we try hard to write good commit messages, we can't edit them later.

So commit messages would only serve as auto-generated input to that yaml file, which can then be modified as needed - and then multiple changelogs (repo, release page, docs) can be generated easily.

It might be possible to write a template for `git-cliff` to output things in yaml. But I also found a link to [clog-cli](https://github.com/clog-tool/clog-cli) on git-cliff's repo. It seems like this has an option to output things in json, so that might be a better fit.

wolfgangwalther:
Tracked in #4590 instead.

PR #1806, opened 2021-04-11
Carve CLI out of Config

monacoremo:


wolfgangwalther:
I remember seeing this commit already. Was that not part of the other PR? Or was that taken out together with App-stuff?

monacoremo:
Taken out with the App stuff. I know this PR seems small, but that's really it :)

src/PostgREST/CLI.hs
```diff
@@ -0,0 +1,72 @@
...
+  ( CLI (..)
+  , Command (..)
+  , readCLIShowHelp
+  ) where
+
+import qualified Data.Map.Strict     as M
+import qualified Options.Applicative as O
+import qualified Protolude.Conv      as Conv
+
+import PostgREST.Config (Environment, example, prettyVersion)
```
wolfgangwalther, line 11: I think `CLI` should not depend on `Config`. `example`, yes, maybe, although I'd be absolutely fine with moving that to `CLI`, too. I see why you left it in there, because it has some kind of "natural" connection. But code-wise it does not belong at all, it has no connection. And when changing something about config options, one has to touch many files anyway, so it doesn't do anything for maintenance either.

Versioning and environment stuff just seems to be wrong in `Config`. It might seem like overkill, but `Environment` should really be in its own file, because it is referenced from `Main`, `Config` and `CLI`. Version stuff is referenced from a few places, too, right?

wolfgangwalther: Or: Let's say `Environment` is only really needed for `Config` - let's keep it in there. But then the interface to `readCLIShowHelp` should not pass in the whole environment, but just a Bool `hasEnvironment`. That should be enough for the purpose of deciding whether the argument is optional or not. In fact, I like this more!

Versioning stuff still seems to be wrong in `Config`.

monacoremo: Yes, versioning is only used in the CLI I think, let's move it there.

On the `example`: I thought that it would be good to add a test that it actually parses. But we can also do that from CLI.

Usage of `Environment` in `Main` will disappear once we clean `Main` out. Let's keep it in Config for now, but keep in mind to potentially change it later one.

wolfgangwalther: > On the `example`: I thought that it would be good to add a test that it actually parses. But we can also do that from CLI.

That sounds like an io-test for me. Not sure if we actually have one like that already?
In any case, the io test wouldn't care where the example is ;)

monacoremo: > That sounds like an io-test to me.

Or a wonderful unit/doctest! Requires just `Config.parse` and `configFileExample`.

> Yes, versioning is only used in the CLI I think, let's move it there.

It turns out that our good friend `OpenApi` also uses the version strings. While we're at it, let's also carve them out as `Version`? Would make sense in light of their TemplateHaskell requirement and the weirder dependencies those require.

wolfgangwalther: > It turns out that our good friend `OpenApi` also uses the version strings. While we're at it, let's also carve them out as `Version`? Would make sense in light of their TemplateHaskell requirement and the weirder dependencies those require.

I noticed that, too. My judgement was: `OpenApi` does not exist ;). At least at some point in the future, so let's pretend it's not there and don't care about it too much. From this perspective, I think `CLI` would be just fine.

monacoremo: > OpenApi does not exist

True! :-) still, I might try the `Version` route. The TemplateHaskell stuff majorly slows down compilation of Config right now I think, so it might be worth to isolate it

wolfgangwalther: > > That sounds like an io-test to me.
> 
> Or a wonderful unit/doctest! Requires just `Config.parse` and `configFileExample`.

I thought about this a bit more. While a unit/doctest can be very valuable in other places, it's not as valuable here. An IO-Test would be better. That's because it's important that whatever is output on stdout is parseable as a config file. Before I changed it to `--dump-example`, this config was printed together with the usage info. In that case the unittest would have passed, but it wouldn't really have worked.

Since we already have all the machinery in place for the IO-tests with re-dumping config files etc., it'd should be straightforward to add a test there.

wolfgangwalther:
> Or: Let's say Environment is only really needed for Config - let's keep it in there. But then the interface to readCLIShowHelp should not pass in the whole environment, but just a Bool hasEnvironment. That should be enough for the purpose of deciding whether the argument is optional or not. In fact, I like this more!

Shall we do this in here or in a different PR?

Other than that it LGTM.

src/PostgREST/CLI.hs
```diff
@@ -29,54 +27,58 @@ data Command
...
+      O.infoOption exampleConfigFile $
+        O.long "example"
+        <> O.short 'e'
+        <> O.help "Show an example configuration file"
+
+    cliParser :: O.Parser CLI
+    cliParser =
+      CLI
+        <$> (dumpConfigFlag <|> dumpSchemaFlag)
+        <*> optionalIf hasEnvironment configFileOption
```
wolfgangwalther, line 59: :heart:

PR #1808, opened 2021-04-12
art or garbage?

wolfgangwalther:
Or, for our fellow german readers: ["Ist das Kunst oder kann das weg?"](https://forum.duolingo.com/comment/3852423/Interesting-German-Idioms)

I'm not exactly sure what `Setup.hs` is meant to do - but it has not been touched since.... the "Initial commit" like 7 years ago!

I found [this](https://cabal.readthedocs.io/en/3.4/setup-commands.html) and stopped reading after:

> The low-level Cabal interface is implemented using Setup.hs scripts. You should prefer using higher level interface provided by nix-style builds.

Also removed a few static assets that are not used anymore.


steve-chavez:
> https://stackoverflow.com/questions/13876047/why-cabal-tool-doesnt-use-setup-lhs-setup-hs

> 1) You also ask about the rationale for requiring the Setup file anyway: actually, it isn't required if you use the Simple build type. The cabal binary will happily configure and install it without. 

> 2) However, it is considered good style to include a Setup file for any package, because it will allow users to install the package who have the Cabal libary available, but not the cabal-install tool (and Hackage enforces the presence of a Setup file for this reason)

From 1, I'd agree in throwing out the Setup.hs file, but 2 makes me doubt. However, I assume that most Haskell devs are using nix nowadays so handling the `cabal-install` dependency shouldn't be a problem.

So yeah, It's garbage by now :D

wolfgangwalther:
I worry a bit about "and Hackage enforces the presence of a Setup file for this reason". Will that be a problem? I just checked a few packages on Hackage - all have a `Setup.hs` or `Setup.lhs`. Some of those files are just as old as ours... :D

Maybe we should leave that one untouched.

monacoremo:
> Maybe we should leave that one untouched.

Yes. Would be good to add a comment in that file so that our successors don't need to wonder again why we kept that around :-)

PR #1813, opened 2021-04-13
Allow headers only POST request and default the request to minimal

### LINKED ISSUE #1656: Add explicit Prefer return=headers

laurenceisla:
Fixes #1656.

Allows returning only the Location header using "Prefer return=headers-only".
Defaults the POST request made without Prefer header to minimal instead of headers only.

### DIFF
--- a/test/Feature/InsertSpec.hs
+++ b/test/Feature/InsertSpec.hs
@@ -111,12 +111,21 @@ spec actualPgVersion = do
                            , "Content-Range" <:> "*/*" ]
           }
 
-    context "requesting no representation" $
+    context "requesting headers only representation" $
       it "should not throw and return location header when selecting without PK" $
-        request methodPost "/projects?select=name,client_id" []
+        request methodPost "/projects?select=name,client_id" [("Prefer", "return=headers-only")]
           [json|{"id":11,"name":"New Project","client_id":2}|] `shouldRespondWith` ""
           { matchStatus  = 201
           , matchHeaders = [ "Location" <:> "/projects?id=eq.11"
+                           , "Content-Range" <:> "*/*" ]
+          }
+    
+    context "requesting no representation" $
+      it "should not throw and return location header when selecting without PK" $
+        request methodPost "/projects?select=name,client_id" []
+          [json|{"id":12,"name":"New Project","client_id":2}|] `shouldRespondWith` ""
+          { matchStatus  = 201
+          , matchHeaders = [ "Location" <:> "/projects?id=eq.12"
                            , "Content-Range" <:> "*/*" ]
           }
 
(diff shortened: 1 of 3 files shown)


wolfgangwalther:
Note about commit messages:
So far, I have used and also implemented in #1805 the convention to use a `BREAKING CHANGE` anywhere in the commit message to indicate a breaking change - not a prefix `breaking:`. This way, we are a bit more flexible - e.g. when fixing a bug, we can still prefix it with `fix:` and then add a `BREAKING CHANGE`, too.

In this case it would be something like:
```
feat: Add explicit headers-only POST request using Prefer header

Allows using header "Prefer=headers-only" explicitly to get a response with only a Location header.

BREAKING CHANGE: Change default for POST request from headers-only to minimal

Resolves #1656
```

test/Feature/InsertSpec.hs
```diff
@@ -508,13 +516,24 @@ spec actualPgVersion = do
...
-          }
+  describe "Inserting into VIEWs" $ do
+    context "requesting no representation" $
+      it "succeeds with 201" $
+        post "/compound_pk_view"
+            [json|{"k1":1,"k2":"test","extra":2}|]
+          `shouldRespondWith`
+            ""
+            { matchStatus  = 201
+            , matchHeaders = [ "Content-Range" <:> "*/*" ]
```
wolfgangwalther, line 527: Ah I really like that `matchHeaders = [ matchHeaderAbsent hLocation ]` above.

Can we add that here, too?

wolfgangwalther (changes requested):

test/Feature/InsertSpec.hs
laurenceisla: Sure, will do. Should I leave the "Content-Range" in the matchHeaders? or replace it entirely with the matchHeaderAbsent?

wolfgangwalther: I think we can remove that. Keeps the test-cases more focused on the important details.

wolfgangwalther (approved):

wolfgangwalther:
@steve-chavez @monacoremo Any idea what we're looking at in the cirrus build failure here? Is this related to this PR or can I merge?

wolfgangwalther:
> It seems to be this: `(($?==124))`. I think I wrote that to handle a timeout gracefully. Clearly not the case.

It was added here: https://github.com/PostgREST/postgrest/commit/9adec12a678afceca78f6bfebba236b031e63487

I wrote it for travis and then we copied it to cirrus. It seems like the shell that is run in cirrus is different and doesn't support `((...))`.

PR #1814, opened 2021-04-13
refactor: carve out modules from `Main.hs`

monacoremo:
This will reduce the amount of code under conditional CCP compilation, making the code easier to reason about. Carving out modules will also increase unit testing opportunities.

Very WIP.

Edit: I converged to the following structure, to summarize:

* `PostgREST.AppState` encapsulate all stateful pieces of our application. It's now the only module handling `MVars`, `IORefs` etc. This makes it very obvious where state is used in other modules, and makes it easier to refactor the state further.
* `PostgREST.Workers` contains all current workers. They seem a bit convoluted right now, I tried to clarify the current setup by avoiding passing around `connWorker :: IO ()` etc. More work to be done here to simplify
* `PostgREST.Config` has a simplified interface - parsing the config is now one function, not puzzling together with the right incantation of multiple functions. Config could be signifcantly refactored now that most (soon: all?) options are optional and we have several config sources (env, config file, db..)
* `PostgREST.Config.Database` contains a few of the config from db pieces (e.g. the statement, that was a bit out of place with the request/query stuff) - more to come as we refactor Config down the line
* `PostgREST.Unix` now contains all the unix-specific functionality (from UnixSocket and things in old Main)
* `PostgREST.CLI` absorbed a bit more of the miscellaneous functionality of the executable. To be further sorted out as we go 
* `Main` now is quite minimal, also regarding external dependencies. Minimal code is under conditional CPP compilation


main/Main.hs
```diff
@@ -94,138 +87,167 @@ main = do
...
+  -- ^ Used to sync the listener(NOTIFY reload) with the connectionWorker. No
+  -- connection for the listener at first. Only used if dbChannelEnabled=true.
+  , stateDbStructure      :: IORef (Maybe DbStructure)
+  -- ^ No schema cache at the start. Will be filled in by the connectionWorker
+  , stateIsWorkerOn       :: IORef Bool
+  -- ^ Helper ref to make sure just one connectionWorker can run at a time
+  , stateConf             :: IORef AppConfig
+  -- ^ Config that can change at runtime
+  , stateGetTime          :: IO UTCTime
+  }
```
wolfgangwalther, line 228: Cool. Is this essentially what I had down as `Context` in my graph? i.e. - do you plan to move this out and nest config/schema etc. under `AppState`? Or is this just an internal helper here?

monacoremo: Yes, will be split out! As a fully opaque type even I think

src/PostgREST/Unix.hs
```diff
@@ -38,3 +44,29 @@ runAppInSocket settings app socketFileMode sockPath = do
...
   close sock
+
+
+-- SIGNALS
+
+data SignalHandlers = SignalHandlers
+  { fullConfigReReader :: IO ()
+  , connWorker         :: IO ()
+  , releasePool        :: IO ()
+  }
```
wolfgangwalther, line 55: I don't think this record really improves the interface of `installSignalHandlers`. Why not just 3 regular arguments?

src/PostgREST/Config.hs
```diff
@@ -390,18 +386,52 @@ parseSecret bytes =
...
 -- | Read database uri from a separate file if `db-uri` is a filepath.
-readDbUriFile :: Text -> IO (Maybe Text)
-readDbUriFile dbUri = case T.stripPrefix "@" dbUri of
-  Nothing       -> return Nothing
-  Just filename -> Just . T.strip <$> readFile (toS filename)
+readDbUriFile :: AppConfig -> IO AppConfig
+readDbUriFile conf =
+  case T.stripPrefix "@" $ configDbUri conf of
+    Nothing ->
+      return conf
```
wolfgangwalther, line 427:
```suggestion
    Nothing       -> return conf
```
(consistency with `readSecretFile`)

src/PostgREST/Config.hs
```diff
@@ -390,18 +386,52 @@ parseSecret bytes =
...
+  where
+    decodeB64 = B64.decode . encodeUtf8 . T.strip . replaceUrlChars . decodeUtf8
+    replaceUrlChars = T.replace "_" "/" . T.replace "-" "+" . T.replace "." "="
+
+decodeJWKS :: AppConfig -> AppConfig
+decodeJWKS conf =
+  case configJwtSecret conf of
+    Nothing -> conf
+    Just secret ->
+      conf { configJWKS = Just $ parseSecret secret }
```
wolfgangwalther, line 420:
```suggestion
decodeJWKS conf@AppConfig{configJwtSecret} =
  conf { configJWKS = fmap parseSecret configJwtSecret }
```

Something like that should work, too, right? fmapping over the Maybe removes a bit of the boilerplate around the maybe pattern matching.

But I don't know whether that works the same way with the other functions around here, returning `IO AppConfig`. If not, maybe it's better to leave as is for consistency.

src/PostgREST/Config.hs
```diff
@@ -165,77 +160,78 @@ instance JustIfMaybe a a where
...
-            -- replace because the JWT is actually base64url encoded which must be turned into just base64 before decoding.
-            replaceUrlChars = T.replace "_" "/" . T.replace "-" "+" . T.replace "." "="
-            willBeFile = isPrefixOf "@" (toS secStr) && isNothing secretFile
-        in
-        if isB64 && not willBeFile -- don't decode in bas64 if the secret will be a file or it will err. The secFile will be filled with the file contents in a later stage.
-          then case B64.decode . encodeUtf8 . T.strip . replaceUrlChars $ decodeUtf8 secFile of
-            Left errMsg -> fail errMsg
-            Right bs    -> pure $ Just bs
-          else pure $ Just secFile
+      Just s  -> pure . Just $ encodeUtf8 s
```
wolfgangwalther, line 234: Those two can probably be (simplified and) moved just in the list above, similar to all the others?

src/PostgREST/Config.hs
```diff
@@ -75,6 +73,7 @@ data AppConfig = AppConfig {
   , configDbTxAllowOverride     :: Bool
   , configDbTxRollbackAll       :: Bool
   , configDbUri                 :: Text
+  , configFilePath              :: Maybe FilePath
```
wolfgangwalther, line 76: Clever...!! :)

src/PostgREST/Config.hs
```diff
@@ -165,77 +160,78 @@ instance JustIfMaybe a a where
...
+      confWithDbUri <-
+        case previousConf of
+          Nothing ->
+            readDbUriFile parsedConfig
+          Just prevConf ->
+            pure $ parsedConfig { configDbUri = configDbUri prevConf }
+
+      Right . decodeJWKS <$> (decodeSecret =<< readSecretFile confWithDbUri)
+    Left err ->
+      return . Left $ "Error in config: " <> err
```
wolfgangwalther, line 186: Left first? Maybe? Ehh Either? ah whatever. Left Right!

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,114 @@
...
+releasePool AppState{..} = P.release statePool >> throwTo stateMainThreadId UserInterrupt
+
+pgVersion :: AppState -> IO PgVersion
+pgVersion = takeMVar . statePgVersion
+
+putPgVersion :: AppState -> PgVersion -> IO ()
+putPgVersion appState pgVer = void $ tryPutMVar (statePgVersion appState) pgVer
+
+dbStructure :: AppState -> IO (Maybe DbStructure)
+dbStructure = readIORef . stateDbStructure
```
wolfgangwalther, line 92: Commenting on a random one of those here:

So I understand we have some getters, because we need to read the `IORef`s in some cases - that's why it doesn't really help to use `RecordWildcards` or whatever at the callsite. I guess you added those other getters (`pool` looks quite simple) for consistency.

How about naming them consistently something like `getDbStructure`, `getPool` or whatever?

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,114 @@
...
+  -- connection for the listener at first. Only used if dbChannelEnabled=true.
+  , stateDbStructure  :: IORef (Maybe DbStructure)
+  -- ^ No schema cache at the start. Will be filled in by the connectionWorker
+  , stateIsWorkerOn   :: IORef Bool
+  -- ^ Helper ref to make sure just one connectionWorker can run at a time
+  , stateConf         :: IORef AppConfig
+  -- ^ Config that can change at runtime
+  , stateGetTime      :: IO UTCTime
+  , stateMainThreadId :: ThreadId
+  }
```
wolfgangwalther, line 51: Weird, to have the comments **below** the fields. Everywhere else we put them above.

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,114 @@
...
+  , pgVersion
+  , isWorkerOn
+  , time
+  , config
+  , dbStructure
+  , putDbStructure
+  , putIsWorkerOn
+  , putPgVersion
+  , putConfig
+  , killMainThread
```
wolfgangwalther, line 18: `:sort` :D

src/PostgREST/Config.hs
monacoremo: True, much better! All these funcs are still quite messy, I was furiously trying to get the tests back to green somehow ... :-)

main/Main.hs
```diff
@@ -1,455 +1,50 @@
...
-  hSetBuffering stderr NoBuffering
-
-  -- read PGRST_ env variables
-  env <- readEnvironment
-
-  -- read command/path from commad line
-  CLI{cliCommand, cliPath} <- readCLIShowHelp . not $ M.null env
+  setBuffering
+  hasPGRSTEnv <- not . M.null <$> readPGRSTEnvironment
+  opts <- CLI.readCLIShowHelp hasPGRSTEnv
```
wolfgangwalther, line 26: I think you could put those into `CLI` now, too!

src/PostgREST/CLI.hs
```diff
@@ -1,19 +1,72 @@
...
+  run cliCommand appState
+  where
+    run :: Command -> AppState -> IO ()
+    run CmdDumpConfig appState = putStr . dumpAppConfig =<< AppState.config appState
+    run CmdDumpSchema appState = putStrLn =<< dumpSchema appState
+    run CmdRun appState = App.run installSignalHandlers runAppInSocket appState
+
+-- | Dump DbStructure schema to JSON
+dumpSchema :: AppState -> IO LBS.ByteString
+dumpSchema appState = do
```
wolfgangwalther, line 53: Similar to how `dumpAppConfig` is in `Config` - this would belong in `DbStructure`, right?

main/Main.hs
monacoremo: I was thinking that this would go away if we have defaults for everything sometime soon, also for dburi etc - not sure though where our discussion stood there @steve-chavez @wolfgangwalther ?

wolfgangwalther: Hm. I don't think we discussed having a default for `db-uri`, yet, or did we? I'd expect that to always remain required.

Although automatic service discovery would be kind of cool...

monacoremo: Might have imagined it :-) Thinking was that we could fit in well with psql and do exactly what it does with defaults (and maybe even with the other env vars it uses). That would mean that in some cases, you could run postgrest without any config. Or in the cases where you've set up a psql environment, as we do in withtmpdb

monacoremo: Continued here #1823

src/PostgREST/CLI.hs
monacoremo: Hm, part of it maybe, e.g. the encode function. Other than with dumpConfig, this is an impure function with error handling that is specific to this use case. That felt like it fits best with CLI..

Actually, most of it would not be needed if we loaded the dbStructure synchronously on startup, before launching the web server. That would mean that we bind to the port/socket a bit later. But no request can succeed anyway before we have the dbStructure. Could be a slightly breaking change... @steve-chavez is there a strong reason why we load the dbStructure on a separate thread? It's clear why we do that on reloads, but on startup?

steve-chavez: > is there a strong reason why we load the dbStructure on a separate thread? It's clear why we do that on reloads, but on startup?

Yes. The db connection can fail at startup. For example, the db can be ready on another machine but DNS resolution can fail(and postgrest errs). While connection recovery happens, it's useful for the web server to reply with a 503 so its health can be monitored. This is important when having many postgrest instances on production.

> dumpSchema appState 

Would it be possible to remove the `dumpSchema` and instead rely on the `request.spec`(https://github.com/PostgREST/postgrest/pull/1794) for debugging? Seems both would give the same information.

wolfgangwalther: > Would it be possible to remove the `dumpSchema` and instead rely on the `request.spec`(#1794) for debugging? Seems both would give the same information.

-1. `dumpSchema` is a much more low-level view on things - ideally. Like right now it should not be filtered at all, while `request.spec` already is. Down the road, the `request.spec` stuff might not even return the full schema cache at all - depending on what we figure out is actually needed for the root endpoint. It'd be better if we didn't limit us in reducing `request.spec` later on - if we use it for debugging, we basically need to expose everything.

wolfgangwalther: > Hm, part of it maybe, e.g. the encode function. Other than with dumpConfig, this is an impure function with error handling that is specific to this use case. That felt like it fits best with CLI..

Ah. I thought about the purity argument briefly, too. I agree. I suggest to rename `dumpConfig` then. The analogy between the two functions is just misleading. I think `dump` would be the IO action. Maybe `Config.toText conf` or something would be better.

wolfgangwalther: > Yes. The db connection can fail at startup. For example, the db can be ready on another machine but DNS resolution can fail(and postgrest errs). While connection recovery happens, it's useful for the web server to reply with a 503 so its health can be monitored. This is important when having many postgrest instances on production.

Ah. I see your point. However, I wonder how that plays with this:

```haskell
when configDbConfig $ reReadConfig True appState
```

Doesn't this wait on the main thread until the connection to the DB could be made, too? Otherwise `--dump-config` wouldn't really work. Maybe I'm completely off, though, because I don't really understand, yet, how all those workers interact and which parts are synchronous or asynchronous.

steve-chavez: > when configDbConfig $ reReadConfig True appState
> Doesn't this wait on the main thread until the connection to the DB could be made, too? 

Ah, no. If that one fails, it just prints an error - the connection is not retried there(but the in-db config is retried again later on the connection worker, noted in the comment). Really I added that part so the tests can still pass with the cli dump config.

> -1. dumpSchema is a much more low-level view on things

Really my main motivation to not use `dumpSchema` was so I can merge the in-db config with the schema cache, as we discussed on https://github.com/PostgREST/postgrest/pull/1760#discussion_r589070303. Then rely on endpoints for both dumpSchema/dumpConfig(not sure if possible) and avoid synchronization work. But now I think we should go with the MVar alternative I mentioned on that comment. The CLI would wait for the connection worker to be done.

wolfgangwalther: > > when configDbConfig $ reReadConfig True appState
> > Doesn't this wait on the main thread until the connection to the DB could be made, too?
> 
> Ah, no. If that one fails, it just prints an error - the connection is not retried there(but the in-db config is retried again later on the connection worker, noted in the comment). Really I added that part so the tests can still pass with the cli dump config.

Ah, I mis-understood the following bit:

```haskell
      if startingUp then
        panic err -- die on invalid config if the program is starting up
```

This only fails when the config is invalid - but not when the connection failed, right?

wolfgangwalther: > But now I think we should go with the MVar alternative I mentioned on that comment. The CLI would wait for the connection worker to be done.

I think after Remo removed the timing bit for `--dump-schema`, it doesn't really matter any more in terms of performance. If we want to have a performance test of our schema queries, we should probably add something like that as a unit test or so.

I have no idea what a MVar is - but would that mean that we'd open the socket/port for `--dump-config` / `--dump-schema`, too? I don't think that would be a good idea.

steve-chavez: > This only fails when the config is invalid - but not when the connection failed, right?

Yes, correct. I can see that part is convoluted. Once the in-db config is inside the schema cache, that code can be removed.

> I think after Remo removed the timing bit for --dump-schema, it doesn't really matter any more in terms of performance.

Sounds good, the MVar would be viable then.

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,111 @@
...
+import Protolude      hiding (toS)
+import Protolude.Conv (toS)
+
+
+data AppState = AppState
+  -- | Connection pool, either a 'Connection' or a 'ConnectionError'
+  { statePool         :: P.Pool
+  -- | Used to sync the listener(NOTIFY reload) with the connectionWorker. No
+  -- connection for the listener at first. Only used if dbChannelEnabled=true.
+  , statePgVersion    :: MVar PgVersion
```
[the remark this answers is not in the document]
wolfgangwalther: Ah, it comes from:

```haskell
  -- Used to sync the listener(NOTIFY reload) with the connectionWorker. No connection for the listener at first. Only used if dbChannelEnabled=true.
  mvarConnectionStatus <- newEmptyMVar
```

So where is that gone?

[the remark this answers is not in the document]
monacoremo: That's mostly inherent to the bit of worker spaghetti we currently have... The db `listener` only starts doing its thing once the connectionWorker has got a connection and the pgVersion. Before the reactor this saved a whole `ConnectionStatus`, but only the pgVersion is actually used. We'll be able to sort this out once we refactor the workers. I wouldn't want to change too much in this PR.

src/PostgREST/AppState.hs
monacoremo: I renamed all getters that need IO to `get...`. I left `pool` as is, as it does not run in IO at directly returns the pool.

monacoremo: At some point `AppState` might become a monad with a few small fixes - but I think for this PR it's fine and usable as is.

src/PostgREST/AppState.hs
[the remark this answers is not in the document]
monacoremo: > So where is that gone?

That's still the same, except that only the pgVersion was ever used from the ConnectionStatus, so we use that

monacoremo:
I've added a summary of what changed to the top comment. Suggest that we make a cut here, WDYT @wolfgangwalther ? Thank you for all the review!!

src/PostgREST/Config.hs
monacoremo: I've cleaned up a little bit - but I think we can simplify Config a lot now that we have more optional settings and the several sources. So I'd leave it at that for this PR.

src/PostgREST/Config.hs
```diff
@@ -371,37 +358,65 @@ readAppConfig dbSettings env optPath dbUriFile secretFile = do
...
 -- | Read database uri from a separate file if `db-uri` is a filepath.
-readDbUriFile :: Text -> IO (Maybe Text)
-readDbUriFile dbUri = case T.stripPrefix "@" dbUri of
-  Nothing       -> return Nothing
-  Just filename -> Just . T.strip <$> readFile (toS filename)
+readDbUriFile :: Maybe AppConfig -> AppConfig -> IO AppConfig
+readDbUriFile maybePrevConf conf =
+  case maybePrevConf of
+    Just prevConf ->
+      pure $ conf { configDbUri = configDbUri prevConf }
```
wolfgangwalther, line 409: Can you expand on this a little bit? I think this is a change of behaviour - it's not possible to reload a different db-uri from the config file anymore, right? What is the background for this?

wolfgangwalther (commented):
I did it! My head is burning a bit, because this a bit harder to review, but I'm still alive.

It's a bit annoying that this is one of the areas where we're really missing a few tests. Need to review a bit more carefully to make sure there is no change in behaviour. Refactoring with 100% coverage is a lot more fun ;)

wolfgangwalther:
> Suggest that we make a cut here, WDYT @wolfgangwalther ?

Yes, please! :D

src/PostgREST/Config.hs
monacoremo: Yes, good question! The behavior should be unchanged. The intention is to explicitly model what happens implicitly here: https://github.com/PostgREST/postgrest/blob/7ca0d469367d480804b2b8b91bdceb653ada754a/main/Main.hs#L146

Also note that the dbUri in the original code is only ever used [here on startup](https://github.com/PostgREST/postgrest/blob/4b46c4eff329c91dc13b26c78b266c7361ecace4/main/Main.hs#L125) and [here in the `listener`](https://github.com/PostgREST/postgrest/blob/4b46c4eff329c91dc13b26c78b266c7361ecace4/main/Main.hs#L359). It can never change once the dbUriFile has been read and Config parsed.

src/PostgREST/Config.hs
```diff
@@ -165,78 +161,69 @@ instance JustIfMaybe a a where
...
+readAppConfig dbSettings optPath previousConf = do
+  env <- readPGRSTEnvironment
+  -- if no filename provided, start with an empty map to read config from environment
+  conf <- maybe (return $ Right M.empty) loadConfig optPath
+
+  case C.runParser (parser optPath env dbSettings) =<< mapLeft show conf of
+    Right parsedConfig ->
+      Right <$> decodeLoadFiles parsedConfig
+    Left err ->
+      return . Left $ "Error in config " <> err
```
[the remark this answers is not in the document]
monacoremo: I'm fine either way - intuitively for me it makes sense to show the happy path first, but this is also fine

src/PostgREST/Workers.hs
```diff
@@ -0,0 +1,249 @@
...
+  = SCLoaded
+  | SCOnRetry
+  | SCFatalFail
+
+-- Time constants
+_32s :: Int
+_32s = 32000000 :: Int -- 32 seconds
+
+_1s :: Int
+_1s  = 1000000  :: Int -- 1 second
```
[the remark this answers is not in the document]
monacoremo: Those are from the original code - I think _1s is used twice. But makes sense to move them to `where`s!

src/PostgREST/Workers.hs
```diff
@@ -0,0 +1,249 @@
...
+
+-- | The purpose of this worker is to obtain a healthy connection to pg and an
+-- up-to-date schema cache(DbStructure).  This method is meant to be called
+-- multiple times by the same thread, but does nothing if the previous
+-- invocation has not terminated. In all cases this method does not halt the
+-- calling thread, the work is preformed in a separate thread.
+--
+-- Note: 'atomicWriteIORef' is essentially a lazy semaphore that prevents two
+-- threads from running 'connectionWorker' at the same time.
```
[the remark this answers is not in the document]
monacoremo: I'll adjust this comment to document that getPgVersion is used as a semaphore, and also adjust the docs in AppState!

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,110 @@
...
+initPool :: AppConfig -> IO P.Pool
+initPool AppConfig{..} =
+  P.acquire
+    ( configDbPoolSize
+    , fromRational . toRational $ configDbPoolTimeout
+    , toS configDbUri
+    )
+
+pool :: AppState -> P.Pool
+pool = statePool
```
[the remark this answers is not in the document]
monacoremo: Ok - I'll rename it to getPool!

main/Main.hs
```diff
@@ -1,455 +1,48 @@
...
-    Catch connWorker
-    ) Nothing
-
-  -- Re-read the config on SIGUSR2
-  void $ installHandler sigUSR2 (
-    Catch fullConfigReReader
-    ) Nothing
+runAppInSocket = Unix.runAppInSocket
+#else
+runAppInSocket _ _ _ _ = pass
```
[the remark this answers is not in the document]
monacoremo: currently it exits silently - does not make too much sense ofcourse. Happy to add a `panic` here!

main/Main.hs
monacoremo: The complexity will be fully handled by our config parser expect - either there is a valid dbUri from one of the config sources, or the parse and therefore the app fails. Pretty sure the `readPGRSTEnv` will disappear! At the same time, I've seen as good practice to have something like `run =<< parseCLI` as the Main of a Haskell app. Ok to leave it like this for a little bit?

monacoremo:
> > Suggest that we make a cut here, WDYT @wolfgangwalther ?
> 
> Yes, please! :D

Thank you so much for your thorough review! Working on the fixes

src/PostgREST/Config.hs
wolfgangwalther: I see. The reason why I looked at that so carefully, is that my passing in `previousConf` into `readAppConfig` that opens up all kind of doors where stuff could go wrong. E.g., if the full prev config was taken and new values applied on it - that would not allow resetting db-settings by just removing them and so on.

From the comment you linked, I understand that not reloading was the easy way out back then. But in our case it makes things more complicated - the easy way out after the refactor would be to just reload it anyway. We do have a few other settings that are not reloadable either (correct?), so after reloading those settings will not reflect reality anymore anyway - which is not a problem.

I suggest to just remove the whole `previousConf` stuff and reload it all the time. Makes it easier to understand, less-error prone for future "improvements", etc.

main/Main.hs
wolfgangwalther: Sure!

src/PostgREST/Config.hs
monacoremo: Currently we can't re-read the dbUri, as it might be read from `stdin`, where it's only available once. I've restricted the arguments to `readAppConfig` for now, without changing the behavior

src/PostgREST/Workers.hs
```diff
@@ -0,0 +1,249 @@
...
+      if startingUp then
+        panic err -- die on invalid config if the program is starting up
+      else
+        hPutStrLn stderr $ "Failed loading in-database config. " <> err
+    Right newConf -> do
+      AppState.putConfig appState newConf
+      if startingUp then
+        pass
+      else
+        putStrLn ("In-database config loaded" :: Text)
```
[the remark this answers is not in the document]
monacoremo: Yes, you're right - most call sites are conditional on `configDbConfig`, but the SIGUSR2 signal handler calls `reReadConfig` regardless. I adjusted the messages.

src/PostgREST/Config.hs
wolfgangwalther: A lot better, thanks! That makes it much easier to see what's happening.

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,110 @@
...
+    <*> newIORef False
+    <*> newIORef conf
+    <*> mkAutoUpdate defaultUpdateSettings { updateAction = getCurrentTime }
+    <*> myThreadId
+
+initPool :: AppConfig -> IO P.Pool
+initPool AppConfig{..} =
+  P.acquire
+    ( configDbPoolSize
+    , fromRational . toRational $ configDbPoolTimeout
```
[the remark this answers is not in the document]
monacoremo: I don't know why this happens. The following change:

```patch
index 3b88022..0b293e5 100644
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -46,6 +46,7 @@ import Data.List               (lookup)
 import Data.List.NonEmpty      (fromList, toList)
 import Data.Maybe              (fromJust)
 import Data.Scientific         (floatingOrInteger)
+import Data.Time.Clock         (NominalDiffTime)
 import Numeric                 (readOct, showOct)
 import System.Environment      (getEnvironment)
 import System.Posix.Types      (FileMode)
@@ -66,7 +67,7 @@ data AppConfig = AppConfig
   , configDbExtraSearchPath     :: [Text]
   , configDbMaxRows             :: Maybe Integer
   , configDbPoolSize            :: Int
-  , configDbPoolTimeout         :: Int
+  , configDbPoolTimeout         :: NominalDiffTime
   , configDbPreRequest          :: Maybe Text
   , configDbPreparedStatements  :: Bool
   , configDbRootSpec            :: Maybe Text
@@ -195,7 +196,7 @@ parser optPath env dbSettings =
     <*> optWithAlias (optInt "db-max-rows")
                      (optInt "max-rows")
     <*> (fromMaybe 10 <$> optInt "db-pool")
-    <*> (fromMaybe 10 <$> optInt "db-pool-timeout")
+    <*> (fromIntegral <$> parsePoolTimeout)
     <*> optWithAlias (optString "db-pre-request")
                      (optString "pre-request")
     <*> (fromMaybe True <$> optBool "db-prepared-statements")
@@ -224,6 +225,9 @@ parser optPath env dbSettings =
     <*> (fmap T.unpack <$> optString "server-unix-socket")
     <*> parseSocketFileMode "server-unix-socket-mode"
   where
+    parsePoolTimeout :: C.Parser C.Config Int
+    parsePoolTimeout = fromMaybe 10 <$> optInt "db-pool-timeout"
+
     parseAppSettings :: C.Key -> C.Parser C.Config [(Text, Text)]
     parseAppSettings key = addFromEnv . fmap (fmap coerceText) <$> C.subassocs key C.value
       where
```

Raise the following error in io tests:

```
postgrest: FatalError {fatalErrorMessage = "Error in config ParseError \"/run/user/1001/pytest-of-remo/pytest-30/test_stable_config_sigusr2_set1/config:7:21:\\n  |\\n7 | db-pool-timeout = 10s\\n  |                     ^\\nunexpected 's'\\nexpecting '.', 'E', 'e', digit, end of input, or end of line\\n\""}
```

This change should not change anything about how "10s" is parsed, no?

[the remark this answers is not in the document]
wolfgangwalther: Since this is the "stable config" test, I would assume this is about `--dump-config`. You didn't adjust the `show configDbPoolTimeout`. I would think you need to convert it to integer there first, because we can't parse the `10s` that are `show`n from the `NominalDiffTime` type.

src/PostgREST/Unix.hs
```diff
@@ -0,0 +1,61 @@
...
+
+
+-- | Run the PostgREST application with user defined socket.
+runAppInSocket :: Warp.Settings -> Application -> FileMode -> FilePath -> IO ()
+runAppInSocket settings app socketFileMode socketFilePath = do
+  sock <- createAndBindSocket
+  putStrLn $ ("Listening on unix socket " :: Text) <> show socketFilePath
+  Socket.listen sock Socket.maxListenQueue
+  Warp.runSettingsSocket settings sock app
+  Socket.close sock
```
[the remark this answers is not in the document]
monacoremo: Ah, `runSettingsSocket` probably never returns. I added a Control.Exception.bracket to make sure release is run.

main/Main.hs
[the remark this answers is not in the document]
monacoremo: fixed

main/Main.hs
```diff
@@ -1,455 +1,48 @@
...
-  -- This is passed to the connectionWorker method so it can kill the main thread if the PostgreSQL's version is not supported.
-  mainTid <- myThreadId
-
-  let connWorker = connectionWorker mainTid pool refConf refDbStructure refIsWorkerOn (dbChannelEnabled, mvarConnectionStatus) $
-                     dbConfigReReader False
-
-  -- Sets the initial refDbStructure
-  connWorker
 
+runAppInSocket :: App.SocketRunner
```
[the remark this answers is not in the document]
monacoremo: fixed

src/PostgREST/AppState.hs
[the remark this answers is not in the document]
monacoremo: Ah, it was failing on the reload! Thank you @wolfgangwalther ! Fixed in latest commit

> Edit: Ah, really cool to see the "stable config test" providing value here :)

Yes, awesome!

src/PostgREST/AppState.hs
```diff
@@ -0,0 +1,110 @@
...
+getConfig = readIORef . stateConf
+
+putConfig :: AppState -> AppConfig -> IO ()
+putConfig = atomicWriteIORef . stateConf
+
+getTime :: AppState -> IO UTCTime
+getTime = stateGetTime
+
+killMainThread :: AppState -> IO ()
+killMainThread = killThread . stateMainThreadId
```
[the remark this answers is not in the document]
monacoremo: I like the idea! Right now it wouldn't fully work without some major changes - e.g. pgVersion is never set if configDbConfig is not enabled. For now I moved the thread killing out of AppState, but let's keep it in mind!

monacoremo:
@wolfgangwalther I think that's it! All review comments should be adressed, in individual commits since you reviewed. I didn't touch the history. Shall we merge?

src/PostgREST/Workers.hs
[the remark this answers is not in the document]
wolfgangwalther: I think you adjusted the comments to pgVersion, but that specific part of the comment that I commented on, is not addressed yet. "atomicWriteIORef", which is referenced here, is not used in the function below.

Either remove the comment or replace "atomicWriteIORef" with "isWorkerOn", wdyt?

wolfgangwalther:
> @wolfgangwalther I think that's it! All review comments should be adressed, in individual commits since you reviewed. I didn't touch the history. Shall we merge?

Apart from the still wrong comment, everything else looks good.

Could you rebase/squash in a way, that everything is in one `refactor:` commit, except:
- `fix: Improve log message on config reload`
- `fix: Close unix socket properly`
- `fix: Panic when running on windows with unix socket enabled`

(I hope I got all of those?)

Would be good to separate those in the history, because they change behaviour. And we can then put them in the changelog, too, when they are properly prefixed and separated.

src/PostgREST/Workers.hs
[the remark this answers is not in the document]
monacoremo: Goof point! done

wolfgangwalther (approved):

monacoremo:
> Would be good to separate those in the history, because they change behaviour. 

Yes, that's good! Everything else was just reshuffling stuff around.

@wolfgangwalther, I hope you will not be too sad, I think this was the last mega-refactor for now. Now that we have split up a lot of stuff, the further changes will be more localized. Thank you so much for bearing with me on this!

wolfgangwalther:
Something went wrong with rebasing. The `always pass` is back on the windows branch.

monacoremo:
> Something went wrong with rebasing. The always pass is back on the windows branch.

Yes, weird - must have slipped by as I reordered the commits, would have expected a rebase conflict. Added it back in.

PR #1817, opened 2021-04-15
ci: split nix-build-test into nix-build nix-test and introduce circle ci contexts

wolfgangwalther:
As discussed via email. This is to secure our cachix, docker and github tokens better. Contexts will still have to be filled with environment variables.

The split in `nix-build` and `nix-test` should also make our pipeline faster, because they can both run in parallel. @monacoremo Did I miss anything in there? I don't think the `nix-build` step will have any problems, but will that be OK for `nix-test`? Or is it missing anything that I removed in that step?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,7 +176,13 @@ jobs:
           when: on_fail
       - run:
           name: Upload coverage to codecov
-          command: bash <(curl -s https://codecov.io/bash) -f coverage/codecov.json
+          command: |
+            # Modified from:
+            # https://docs.codecov.io/docs/about-the-codecov-bash-uploader#validating-the-bash-script
+            curl -s https://codecov.io/bash > codecov;
+            VERSION=$(grep 'VERSION=\".*\"' codecov | cut -d'"' -f2);
+            shasum -a 512 -c <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA512SUM | grep codecov)
+            bash codecov -f coverage/codecov.json
       - run:
           name: Run the spec tests against PostgreSQL 12
           command: postgrest-with-postgresql-12 postgrest-test-spec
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
(I cancelled the nix steps for now, because the cache is not built up again, yet)

wolfgangwalther:
Just noticed that we seem to have a limit of 4 jobs running in parallel on circle. Not too much of a problem because `style-check` is done pretty quickly. Still I wondered whether we should move the two `stack-` jobs into one, maybe. Or if we'd just kill the `stack-test-memory` one. We're not running the full test suite (no io tests) anyway with stack, so maybe the basic spec test is enough for testing basic functionality of the build that was made with stack?

steve-chavez:
> Or if we'd just kill the stack-test-memory one.

:+1: For that. A single stack job should be enough, we just need to make sure it builds(so stackage can notify robx of dependency changes).

wolfgangwalther:
@monacoremo `nix-build` on the first commit of this PR was running for > 1h so far: https://app.circleci.com/pipelines/github/PostgREST/postgrest/911/workflows/f1b6b8cf-2f66-486f-8811-dcb3ac232bfb/jobs/8737. I cancelled it now, because it would not push anything to cachix anyway.

Are we still missing some of the cache?

monacoremo:
Yes, noticed that some of the static build assets were not pushed yet. Running that now

wolfgangwalther:
Also added a checksum check for the codecov bash script and reordered `nix-test` a bit to report coverage back earlier.

The tokens are still going to be available through regular environment variables for now. @steve-chavez once you have added the new tokens to the contexts, you can remove the old ones from the env vars. I suggest you then trigger a nightly build - that should test whether context + tokens work fine.

Since we can't test it in the PR anyway, because the tokens won't be used here, I'm going to merge this now.

wolfgangwalther:
> Since we can't test it in the PR anyway, because the tokens won't be used here, I'm going to merge this now.

Actually, I will wait for @monacoremo to build the remaining cache and then merge to avoid a long running build job on main. Let me know!

Issue #1771, opened 2021-03-16
Embedding filter gets silently ignored if the resource doesn't exist


yyqgituser:
### Environment

* PostgreSQL version: PostgreSQL 13.2 (docker)
* PostgREST version: 7.0.1
* Operating system: Linux

### Embedding filter gets silently ignored if the resource doesn't exist
Embedded resources can be filtered by a filter parameter, which is prefixed  with the name of the embedded resource, for example,
```
GET /films?select=*,roles(*)&roles.character=in.(Chico,Harpo,Groucho) HTTP/1.1
```
This is expected, but when trying to filter with a invalid resource name, PostgREST just silently ignored the filter and acted as no filter at all. Following query will return all roles without filtering.
```
GET /films?select=*,roles(*)&resource_not_exist.character=in.(Chico,Harpo,Groucho) HTTP/1.1
```
Though it doesn't make sense to specify an invalid resource name, in practice it may be just a typo, so returning 400 and some error message can help use to correct their query, for example:
```
{
    "hint": null,
    "details": null,
    "code": "xxxxx",
    "message": "resource resource_not_exist does not exist"
}
```

Issue #249, opened 2015-08-02
Customize query timeouts


begriffs:
Either add a server command line argument to specify the timeout for all connections, or establish it per-user.

For the first option we can make another middleware that composes with `authenticated`:
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Main.hs#L83

For the second option (per user) we can add another column to our auth table and update `setRole`:
https://github.com/begriffs/postgrest/blob/master/src/PostgREST/Auth.hs#L56

We'll use the command `SET SESSION STATEMENT_TIMEOUT TO [ms];`

/cc @smilliken

steve-chavez:
I see this as a requirement for #286, I think we need to restrict the clients api usage for that case.

I think this kind of measure would be better enabled at the database level than at the api level, say the dba needs to set the timeout for each user of the api as in:
```sql
ALTER ROLE webuser SET statement_timeout = 30000
```

Just gonna throw an idea here, maybe PostgREST could have a `force-timeout=true`(by default true) config param, that could make it query the user `current_setting('statement_timeout')` for each transaction and if it is undefined or not `> 0` it would give an error response.   

Now another option I've seen is to enforce a `statement_cost_limit` because as mentioned in this [article](http://www.databasesoup.com/2015/02/why-you-might-need-statementcostlimit.html):

> PostgreSQL has had statement_timeout for a while, which can be set on a per-user basis (or other places) to prevent application errors from running queries for hours.  However, this doesn't really solve the "overload" issue, because the query runs for that length of time, gobbling resources until it's terminated.  What you really want to do is return an error immediately if a query is going to be too costly.

It seems like a safer option, but it relies on a loadable module and it may be inconvenient for the user to setup, so for now maybe we could discuss `timeout` as an alternative.

steve-chavez:
More thoughts on this, I think restricting the statement cost is the best choice, so I've been testing the [plan_filter](https://github.com/pgexperts/pg_plan_filter) module, originally I thought that it would be enough to setup the roles each with it's own cost in the db without changing PostgREST code but that doesn't work. 

Say we enable the `plan_filter` module and then do `ALTER ROLE .. SET plan_filter.statement_cost_limit = ..` for some roles, then we have:

| Role  | plan_filter.statement_cost_limit |
| ------------- | ------------- |
| authenticator  | 4000(has to fill `dbStructure`)  |
| anonymous  | 500(can only `login`)  |
| privileged  | 10000  |

[...]

PR #1818, opened 2021-04-16
refactor: make Junction non-recursive

steve-chavez:
While working on https://github.com/PostgREST/postgrest/pull/1794, I noticed that we can have recursive Junctions:

https://github.com/PostgREST/postgrest/blob/4b46c4eff329c91dc13b26c78b266c7361ecace4/src/PostgREST/DbStructure/Relation.hs#L43-L53

Which didn't seem right, as discussed in https://github.com/PostgREST/postgrest/commit/522308217a03b912e858bdb3311641b6c65c1954#r49568150.

This PR fixes that, also Cardinality now contains the constraint or junction that defines it:

```haskell
data Cardinality
  = O2M ConstraintName -- ^ one-to-many cardinality
  | M2O ConstraintName -- ^ many-to-one cardinality
  | M2M Junction       -- ^ many-to-many cardinality
  deriving (Eq, Generic, JSON.ToJSON)
```

That makes sense IMO, it clears up the code in some parts.

### DIFF
--- a/src/PostgREST/DbStructure/Relation.hs
+++ b/src/PostgREST/DbStructure/Relation.hs
@@ -25,11 +25,11 @@ import Protolude
 --
 -- TODO merge relColumns and relFColumns to a tuple or Data.Bimap
 data Relation = Relation
-  { relTable    :: Table
-  , relColumns  :: [Column]
-  , relFTable   :: Table
-  , relFColumns :: [Column]
-  , relCard     :: Cardinality
+  { relTable          :: Table
+  , relColumns        :: [Column]
+  , relForeignTable   :: Table
+  , relForeignColumns :: [Column]
+  , relCardinality    :: Cardinality
   }
   deriving (Eq, Generic, JSON.ToJSON)
 
@@ -37,25 +37,25 @@ data Relation = Relation
 -- | https://en.wikipedia.org/wiki/Cardinality_(data_modeling)
 -- TODO: missing one-to-one
 data Cardinality
-  = O2M ConstraintName -- ^ one-to-many cardinality
-  | M2O ConstraintName -- ^ many-to-one cardinality
-  | M2M Junction       -- ^ many-to-many cardinality
+  = O2M FKConstraint -- ^ one-to-many cardinality
+  | M2O FKConstraint -- ^ many-to-one cardinality
+  | M2M Junction     -- ^ many-to-many cardinality
   deriving (Eq, Generic, JSON.ToJSON)
 
-type ConstraintName = Text
+type FKConstraint = Text
 
 -- | Junction table on an M2M relationship
 data Junction = Junction
   { junTable       :: Table
-  , junConstraint1 :: ConstraintName
+  , junConstraint1 :: FKConstraint
   , junColumns1    :: [Column]
-  , junConstraint2 :: ConstraintName
+  , junConstraint2 :: FKConstraint
   , junColumns2    :: [Column]
   }
   deriving (Eq, Generic, JSON.ToJSON)
 
 isSelfReference :: Relation -> Bool
-isSelfReference r = relTable r == relFTable r
+isSelfReference r = relTable r == relForeignTable r
 
 data PrimaryKey = PrimaryKey
   { pkTable :: Table
--- a/src/PostgREST/DbStructure/Relation.hs
+++ b/src/PostgREST/DbStructure/Relation.hs
@@ -46,11 +46,11 @@ type ConstraintName = Text
 
 -- | Junction table on an M2M relationship
 data Junction = Junction
-  { junTable :: Table
-  , junCons1 :: ConstraintName
-  , junCols1 :: [Column]
-  , junCons2 :: ConstraintName
-  , junCols2 :: [Column]
+  { junTable       :: Table
+  , junConstraint1 :: ConstraintName
+  , junColumns1    :: [Column]
+  , junConstraint2 :: ConstraintName
+  , junColumns2    :: [Column]
   }
   deriving (Eq, Generic, JSON.ToJSON)
 
--- a/src/PostgREST/DbStructure/Relation.hs
+++ b/src/PostgREST/DbStructure/Relation.hs
@@ -3,11 +3,10 @@
 
 module PostgREST.DbStructure.Relation
   ( Cardinality(..)
-  , Constraint
   , ForeignKey(..)
-  , Link(..)
   , PrimaryKey(..)
   , Relation(..)
+  , Junction(..)
   , isSelfReference
   ) where
 
@@ -15,8 +14,6 @@ import qualified Data.Aeson as JSON
 
 import PostgREST.DbStructure.Table (Column (..), ForeignKey (..),
                                     Table (..))
-
-import qualified GHC.Show (show)
 
 import Protolude
 
@@ -32,45 +29,36 @@ data Relation = Relation
   , relColumns  :: [Column]
   , relFTable   :: Table
   , relFColumns :: [Column]
-  , relType     :: Cardinality
-  , relLink     :: Link -- ^ Constraint on O2M/M2O, Junction for M2M Cardinality
+  , relCard     :: Cardinality
   }
+  deriving (Eq, Generic, JSON.ToJSON)
+
+-- | The relationship cardinality
+-- | https://en.wikipedia.org/wiki/Cardinality_(data_modeling)
+-- TODO: missing one-to-one
+data Cardinality
+  = O2M ConstraintName -- ^ one-to-many cardinality
+  | M2O ConstraintName -- ^ many-to-one cardinality
+  | M2M Junction       -- ^ many-to-many cardinality
   deriving (Eq, Generic, JSON.ToJSON)
 
 type ConstraintName = Text
 
 -- | Junction table on an M2M relationship
-data Link
-  = Constraint
-      { constName :: ConstraintName }
-  | Junction
-      { junTable :: Table
-      , junLink1 :: Link
-      , junCols1 :: [Column]
-      , junLink2 :: Link
-      , junCols2 :: [Column]
-      }
+data Junction = Junction
+  { junTable :: Table
+  , junCons1 :: ConstraintName
+  , junCols1 :: [Column]
+  , junCons2 :: ConstraintName
+  , junCols2 :: [Column]
+  }
   deriving (Eq, Generic, JSON.ToJSON)
+
+isSelfReference :: Relation -> Bool
+isSelfReference r = relTable r == relFTable r
 
 data PrimaryKey = PrimaryKey
   { pkTable :: Table
   , pkName  :: Text
   }
   deriving (Generic, JSON.ToJSON)
-
--- | The relationship
--- [cardinality](https://en.wikipedia.org/wiki/Cardinality_(data_modeling)).
--- TODO: missing one-to-one
-data Cardinality
-  = O2M -- ^ one-to-many,  previously known as Parent
-  | M2O -- ^ many-to-one,  previously known as Child
-  | M2M -- ^ many-to-many, previously known as Many
-  deriving (Eq, Generic, JSON.ToJSON)
-
-instance Show Cardinality where
-  show O2M = "o2m"
-  show M2O = "m2o"
-  show M2M = "m2m"
-
-isSelfReference :: Relation -> Bool
-isSelfReference r = relTable r == relFTable r
--- a/src/PostgREST/Request/DbRequestBuilder.hs
+++ b/src/PostgREST/Request/DbRequestBuilder.hs
@@ -126,7 +126,7 @@ addRels :: Schema -> [Relation] -> Maybe ReadRequest -> ReadRequest -> Either Ap
 addRels schema allRels parentNode (Node (query@Select{from=tbl}, (nodeName, _, alias, hint, depth)) forest) =
   case parentNode of
     Just (Node (Select{from=parentNodeQi}, _) _) ->
-      let newFrom r = if qiName tbl == nodeName then tableQi (relFTable r) else tbl
+      let newFrom r = if qiName tbl == nodeName then tableQi (relForeignTable r) else tbl
           newReadNode = (\r -> (query{from=newFrom r}, (nodeName, Just r, alias, Nothing, depth))) <$> rel
           rel = findRel schema allRels (qiName parentNodeQi) nodeName hint
       in
@@ -158,7 +158,7 @@ findRel schema allRels origin target hint =
     -- foreign key and relTable/relFtable but with different
     -- cardinalities(m2o/o2m) We output the O2M rel, the M2O rel can be
     -- obtained by using the origin column as an embed hint.
-    rs@[rel0, rel1]  -> case (relCard rel0, relCard rel1, relTable rel0 == relTable rel1 && relFTable rel0 == relFTable rel1) of
+    rs@[rel0, rel1]  -> case (relCardinality rel0, relCardinality rel1, relTable rel0 == relTable rel1 && relForeignTable rel0 == relForeignTable rel1) of
       (O2M cons1, M2O cons2, True) -> if cons1 == cons2 then Right rel0 else Left $ AmbiguousRelBetween origin target rs
       (M2O cons1, O2M cons2, True) -> if cons1 == cons2 then Right rel1 else Left $ AmbiguousRelBetween origin target rs
       _                            -> Left $ AmbiguousRelBetween origin target rs
@@ -175,16 +175,16 @@ findRel schema allRels origin target hint =
     rel = filter (
       \Relation{..} ->
         -- Both relationship ends need to be on the exposed schema
-        schema == tableSchema relTable && schema == tableSchema relFTable &&
+        schema == tableSchema relTable && schema == tableSchema relForeignTable &&
         (
           -- /projects?select=clients(*)
           origin == tableName relTable  &&  -- projects
-          target == tableName relFTable ||  -- clients
+          target == tableName relForeignTable ||  -- clients
 
           -- /projects?select=projects_client_id_fkey(*)
           (
-            origin == tableName relTable &&       -- projects
-            matchConstraint (Just target) relCard -- projects_client_id_fkey
+            origin == tableName relTable &&              -- projects
+            matchConstraint (Just target) relCardinality -- projects_client_id_fkey
           ) ||
           -- /projects?select=client_id(*)
           (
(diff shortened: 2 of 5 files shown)


src/PostgREST/DbStructure/Relation.hs
```diff
@@ -32,45 +29,36 @@ data Relation = Relation
   , relColumns  :: [Column]
   , relFTable   :: Table
   , relFColumns :: [Column]
-  , relType     :: Cardinality
-  , relLink     :: Link -- ^ Constraint on O2M/M2O, Junction for M2M Cardinality
+  , relCard     :: Cardinality
```
wolfgangwalther, line 32:
```haskell
  , relForeignTable   :: Table
  , relForeignColumns :: [Column]
  , relCardinality    :: Cardinality
```

could we rename it like that? Would improve readability a lot, I think. I started reading the code above, before I came by the type definition - and I first wondered, what that "card" of a relation was.

I think it would be great if down the road we could make the naming of our record fields consistently:
- have a good, abbreviated prefix to help with `RecordWildCards`. `Relation` already has a good prefix, but for example we will need to improve on `ApiRequest.iXXX`.
- use a non-abbreviated name itself. `relCard` is kinda ookaaay in itself. But the `F` is really hard to spot.

src/PostgREST/DbStructure/Relation.hs
```diff
@@ -32,45 +29,36 @@ data Relation = Relation
...
-      , junCols1 :: [Column]
-      , junLink2 :: Link
-      , junCols2 :: [Column]
-      }
+data Junction = Junction
+  { junTable :: Table
+  , junCons1 :: ConstraintName
+  , junCols1 :: [Column]
+  , junCons2 :: ConstraintName
+  , junCols2 :: [Column]
```
wolfgangwalther, line 53: (see above)

```suggestion
  , junConstraint1 :: ConstraintName
  , junColumns1    :: [Column]
  , junConstraint2 :: ConstraintName
  , junColumns2    :: [Column]
```

src/PostgREST/DbStructure/Relation.hs
```diff
@@ -32,45 +29,36 @@ data Relation = Relation
...
-data Cardinality
-  = O2M -- ^ one-to-many,  previously known as Parent
-  | M2O -- ^ many-to-one,  previously known as Child
-  | M2M -- ^ many-to-many, previously known as Many
-  deriving (Eq, Generic, JSON.ToJSON)
-
-instance Show Cardinality where
-  show O2M = "o2m"
-  show M2O = "m2o"
-  show M2M = "m2m"
```
wolfgangwalther, line 73: Ah, good to get rid of that show instance. @monacoremo will be happy. :)

src/PostgREST/Request/DbRequestBuilder.hs
```diff
@@ -152,21 +153,25 @@ findRel schema allRels origin target hint =
...
-      --
-      -- Here we handle a self reference relationship to not cause a breaking
-      -- change: In a self reference we get two relationships with the same
-      -- foreign key and relTable/relFtable but with different
-      -- cardinalities(m2o/o2m) We output the O2M rel, the M2O rel can be
-      -- obtained by using the origin column as an embed hint.
-      let [rel0, rel1] = take 2 rs in
-      if length rs == 2 && relLink rel0 == relLink rel1 && relTable rel0 == relTable rel1 && relFTable rel0 == relFTable rel1
-        then note (NoRelBetween origin target) (find (\r -> relType r == O2M) rs)
-        else Left $ AmbiguousRelBetween origin target rs
```
wolfgangwalther, line 167: Could you explain that change a little bit?

Where is the `note (NoRelBetween...)` gone? Was that dead code?

wolfgangwalther (commented):
The new typing is **much** better than a revert would have been. And also better than all my suggestions before. This also avoids to have a `Junction` with `O2M` or something like that. Nice!

I'm not seeing any codecov reports here, yet. Probably because CircleCI was running with your fork when you first uploaded without a PR, @steve-chavez. But didn't you add the codecov token or something like that in your own fork's CircleCI settings to avoid that? I remember we discussed that, at least. Anyway, I will trigger a re-run of the pipeline and then we will see whether how code coverage is with those new types.

wolfgangwalther:
> Anyway, I will trigger a re-run of the pipeline and then we will see whether how code coverage is with those new types.

Only... that I can't seem to do that. Can't find a way to trigger a build for a PR, that doesn't have one, yet. And I can't rerun the pipeline because it's in your workspace, not postgrest.

src/PostgREST/Request/DbRequestBuilder.hs
steve-chavez: This is the code related to: https://github.com/PostgREST/postgrest/issues/1643#issuecomment-721417728. It will be deleted when I fix that issue.

> note (NoRelBetween...) gone?

Yes, the `note` was just to handle the case were the `find  relType r == O2M` returned Nothing(which couldn't happen). Now that there's a pattern match for 2 exact elements in the list that's not needed anymore.

wolfgangwalther: > This is the code related to: [#1643 (comment)](https://github.com/PostgREST/postgrest/issues/1643#issuecomment-721417728). It will be deleted when I fix that issue.

Ok, I had a feeling that this piece of code is something that will possibly change once we fix some things.

My main concern was, that it still is only refactor at this point and no implied feature or fix. Just to double check - is that correct?


> Yes, the `note` was just to handle the case were the `find relType r == O2M` returned Nothing(which couldn't happen). Now that there's a pattern match for 2 exact elements in the list that's not needed anymore.

I think those impossible branches / pattern matches are kind of a result of wrong typing. Wrong in the sense of "typing would allow cases, that just don't happen in reality". Now that we fixed it for these types - I'm really curious how the codecov report comes back. There should really not be any impossible branches showing up.

src/PostgREST/DbStructure/Relation.hs
```diff
@@ -32,45 +29,36 @@ data Relation = Relation
...
+-- | The relationship cardinality
+-- | https://en.wikipedia.org/wiki/Cardinality_(data_modeling)
+-- TODO: missing one-to-one
+data Cardinality
+  = O2M ConstraintName -- ^ one-to-many cardinality
+  | M2O ConstraintName -- ^ many-to-one cardinality
+  | M2M Junction       -- ^ many-to-many cardinality
+  deriving (Eq, Generic, JSON.ToJSON)
+
 type ConstraintName = Text
```
wolfgangwalther, line 45: One more thing about naming came to mind: Now, that we don't have `Constraint` anymore, why don't we just:

```suggestion
type Constraint = Text
```

wolfgangwalther:
> I'm not seeing any codecov reports here, yet. Probably because CircleCI was running with your fork when you first uploaded without a PR, @steve-chavez. But didn't you add the codecov token or something like that in your own fork's CircleCI settings to avoid that? I remember we discussed that, at least. Anyway, I will trigger a re-run of the pipeline and then we will see whether how code coverage is with those new types.

What might have happenend is that Codecov has reverted the token you have set up in your own CircleCI, as a consequence of the security incident.

src/PostgREST/Request/DbRequestBuilder.hs
steve-chavez: > My main concern was, that it still is only refactor at this point and no implied feature or fix. Just to double check - is that correct?

Yes, I've made sure the logic is the same, no break(there are tests as well).

> Now that we fixed it for these types - I'm really curious how the codecov report comes back. 

You know, by mistake I removed the secrets of my own CircleCI setup yesterday. I've now added the codecov token back, hope the report shows now.

src/PostgREST/DbStructure/Relation.hs
```diff
@@ -32,45 +29,36 @@ data Relation = Relation
...
   deriving (Eq, Generic, JSON.ToJSON)
 
+-- | The relationship cardinality
+-- | https://en.wikipedia.org/wiki/Cardinality_(data_modeling)
+-- TODO: missing one-to-one
+data Cardinality
+  = O2M ConstraintName -- ^ one-to-many cardinality
+  | M2O ConstraintName -- ^ many-to-one cardinality
+  | M2M Junction       -- ^ many-to-many cardinality
+  deriving (Eq, Generic, JSON.ToJSON)
```
wolfgangwalther, line 43: Ah, you can probably remove one of them. You can see in the html coverage report (when you run `postgrest-coverage` locally), which one it is.

Just had a look - that would be `Eq`. Not need on any of the three types (`Junction`, `Cardinality` or `Relation`).

src/PostgREST/DbStructure/Relation.hs
```diff
@@ -32,45 +29,36 @@ data Relation = Relation
...
-      , junCols1 :: [Column]
-      , junLink2 :: Link
-      , junCols2 :: [Column]
-      }
+data Junction = Junction
+  { junTable       :: Table
+  , junConstraint1 :: ConstraintName
+  , junColumns1    :: [Column]
+  , junConstraint2 :: ConstraintName
+  , junColumns2    :: [Column]
```
wolfgangwalther, line 53: This is something not to worry about - those missing record fields are caused by `RecordWildCard` and will be solved with the coverage overlay later.

src/PostgREST/Request/DbRequestBuilder.hs
```diff
@@ -152,21 +153,25 @@ findRel schema allRels origin target hint =
...
+    -- Here we handle a self reference relationship to not cause a breaking
+    -- change: In a self reference we get two relationships with the same
+    -- foreign key and relTable/relFtable but with different
+    -- cardinalities(m2o/o2m) We output the O2M rel, the M2O rel can be
+    -- obtained by using the origin column as an embed hint.
+    rs@[rel0, rel1]  -> case (relCard rel0, relCard rel1, relTable rel0 == relTable rel1 && relFTable rel0 == relFTable rel1) of
+      (O2M cons1, M2O cons2, True) -> if cons1 == cons2 then Right rel0 else Left $ AmbiguousRelBetween origin target rs
```
wolfgangwalther, line 162: So this is interesting - the pattern O2M / M2O seems not to be hit at all. Why?

And can we just delete the whole row? Or will this be changed as part of the issue you mentioned anyway?

Note, that this impossible branch was not there before - before it was only the `note` that was not touched. So I wonder whether we can improve on that.

src/PostgREST/DbStructure/Relation.hs
steve-chavez: Going to rename it ot `FKConstraint`, seems clearer. Also, otherwise:

```
    Ambiguous occurrence ‘Constraint’
    It could refer to
       either ‘Protolude.Constraint’,
```

src/PostgREST/Request/DbRequestBuilder.hs
```diff
@@ -358,14 +357,10 @@ returningCols rr@(Node _ forest) pkCols
...
-    -- for the other relType's.
+    -- succeeds, result would be `RETURNING name, client_id`.
     fkCols = concat $ mapMaybe (\case
-        Node (_, (_, Just Relation{relColumns=cols, relType=relTyp}, _, _, _)) _ -> case relTyp of
-          O2M -> Just cols
-          M2O -> Just cols
-          M2M -> Just cols
-        _ -> Nothing
+        Node (_, (_, Just Relation{relColumns=cols}, _, _, _)) _ -> Just cols
+        _                                                        -> Nothing
```
wolfgangwalther, line 363: This one was uncovered before, too, so no need to worry about for now.

wolfgangwalther (commented):
Just had a look at the uncovered lines.

src/PostgREST/Request/DbRequestBuilder.hs
wolfgangwalther: Or... are we just missing a test-case for a self-reference embedding "the other way around"?

wolfgangwalther: Ah, I now understood:

> It will be deleted when I fix that issue.

So the whole thing will not be there anymore at all. All good then :)

src/PostgREST/DbStructure/Relation.hs
steve-chavez: First time using checking the html, that's really cool!

Is there a nix command for that? I just did a `python -mSimpleHTTPServer 8080` on `/coverage`.

wolfgangwalther: Fair question! We have nix tools for everything. Except that.

I just open the index file directly in the browser via file manager. Works just as fine.

wolfgangwalther: Actually. Exactly for that case I have added the `file://` URI to the CLI output. My terminal has clickable-URI support... so I should not even use a file manager :)

src/PostgREST/Request/DbRequestBuilder.hs
steve-chavez: > So this is interesting - the pattern O2M / M2O seems not to be hit at all. Why?

This could be because when building the DbStructure, the M2O rels are always built first. So when finding the rel with `filter` on the `Relation` list, the `M2O` will always be first and thus the O2M-M2O case will not happen.

> So the whole thing will not be there anymore at all.

Yeah, this will be removed anyway.

src/PostgREST/DbStructure/Relation.hs
steve-chavez: Cool. I'll have to level up my terminal game then :)

>  that would be Eq. Not need on any of the three types (Junction, Cardinality or Relation).

Can't seem to do that:

```haskell
src/PostgREST/Request/DbRequestBuilder.hs:306:61: error:
    • No instance for (Eq Relation) arising from a use of ‘delete’
    • In the second argument of ‘(:)’, namely ‘delete tn forest’
      In the second argument of ‘Node’, namely
        ‘(addProperty f (remainingPath, a) tn : delete tn forest)’
      In the expression:
        Node rn (addProperty f (remainingPath, a) tn : delete tn forest)
    |
306 |     Just tn -> Node rn (addProperty f (remainingPath, a) tn:delete tn forest)
```

steve-chavez: Weird because I also see the same "never executed" notice on the Eqs, on the coverage html.

wolfgangwalther: Well, if the only place where `Eq` is used is uncovered, too.... then we need it of course :)

I should have been more precise. It's possible that one of the 3 `Eq`s is unused - you can find out by removing. If no error, it can be safely removed.

wolfgangwalther: > Well, if the only place where `Eq` is used is uncovered, too.... then we need it of course :)

Which does not seem to be the case here, I think. That line you referenced, does seem to be covered.

I think the only thing that can happen here is, that `delete` is only ever run with an empty list here. That's why it never calls `Eq` internally. But it still depends on it.

wolfgangwalther: I don't pretend to understand what the `addProperty` function does, so I don't know whether that makes any sense, whether we're missing a test-case here or whether this is kind of an impossible branch.

wolfgangwalther (approved):

src/PostgREST/DbStructure/Relation.hs
steve-chavez: > Well, if the only place where Eq is used is uncovered, too...

Ah, but the line where it fails it doesn't show as uncovered:

https://github.com/PostgREST/postgrest/blob/1f206a560bedf0d7971e9db5e475c37ec311acda/src/PostgREST/Request/DbRequestBuilder.hs#L307

> I should have been more precise. It's possible that one of the 3 Eqs is unused - you can find out by removing. If no error, it can be safely removed.

No dice for Eq. I remove it on Cardinality or Junction and Relation fails compiling, I remove the Eq from Relation and it gives the [above error](https://github.com/PostgREST/postgrest/pull/1818#discussion_r615012352). Maybe the coverage "never executed" is related to the record fields?

Anyway, I guess I'll just leave it as it is.

steve-chavez: > I don't pretend to understand what the addProperty function does, so I don't know whether that makes any sense, whether we're missing a test-case here or whether this is kind of an impossible branch.

No, that part is related to [embedded filters](https://postgrest.org/en/stable/api.html#embedded-filters). When having `&roles.character=in.(Chico,Harpo,Groucho)`, addProperty adds the `.character` part if it's a valid embed or drops it silently(ref: https://github.com/PostgREST/postgrest/issues/1771). Code definitely needs cleaning.

wolfgangwalther: > No dice for Eq. I remove it on Cardinality or Junction and Relation fails compiling, I remove the Eq from Relation and it gives the [above error](https://github.com/PostgREST/postgrest/pull/1818#discussion_r615012352). Maybe the coverage "never executed" is related to the record fields?

Ah, of course that makes sense. If we want `Eq` on `Relation` we need it on both the others, too - otherwise a comparison on `Relation` could never be done.

The coverage line is for sure not related to the record fields, I had seen many of those already.

I just noted the `delete` is actually part of a recursive call. Maybe the problem is that, we don't have a test for an embedded filter on a nested embedding? Something like `films.roles.character=in...`. Eh, I keep questioning myself whether that's actually possible to do, because I never find any docs on it. But we **can** do nested embeddings, right?

steve-chavez:
cachix push failed: https://app.circleci.com/pipelines/github/PostgREST/postgrest/917/workflows/e3706800-a3aa-4fac-8a55-2573faf1d726/jobs/8771.

Hm, circle is missing the cachix auth token right?

src/PostgREST/DbStructure/Relation.hs
steve-chavez: > Maybe the problem is that, we don't have a test for an embedded filter on a nested embedding? Something like films.roles.character=in...

True, I think we don't have such test.

> Eh, I keep questioning myself whether that's actually possible to do, because I never find any docs on it. But we can do nested embeddings, right?

Yes, correct. To an N depth, which always reminds me of https://github.com/PostgREST/postgrest/issues/249#issuecomment-334630205. I've put an idea on https://github.com/PostgREST/postgrest/pull/1710#issuecomment-808788905 (at the bottom) to solve that one.

wolfgangwalther:
It says:

> Binary cache doesn't exist or it's a private cache and you don't have access. Please check the name for typos or provide an auth token.

I remember @monacoremo suggesting to switch from signing key to token or something? In CircleCI there still seems to be signing key, not a token? Do we only need to change the environment variable - or do we need to change the CI script, too?

steve-chavez:
> Do we only need to change the environment variable - or do we need to change the CI script, too?

Ah, yes, I think so. What puzzled me is that the previous commit was green: https://github.com/PostgREST/postgrest/tree/1f206a560bedf0d7971e9db5e475c37ec311acda.

wolfgangwalther:
> Ah, yes, I think so. What puzzled me is that the previous commit was green: https://github.com/PostgREST/postgrest/tree/1f206a560bedf0d7971e9db5e475c37ec311acda.

Yeah, that didn't touch any code, so didn't have to upload to cachix.

monacoremo:
We'll need to add a line before we push:

```
cachix authtoken "$CACHIX_AUTH_TOKEN"
```

@steve-chavez can you set the env variable up with a separate CI auth token? Happy to fix the CI then

steve-chavez:
@monacoremo Done! I set up the `CACHIX_AUTH_TOKEN` under the cachix context.

monacoremo:
@steve-chavez perfect, thank you! Just added #1821

PR #1812, opened 2021-04-12
Add `postgrest-loadtest` based on vegeta

wolfgangwalther:
This is my take on #1609. As discussed in that PR, I ran a few tests with [vegeta](https://github.com/tsenart/vegeta).

Compared to locust, as initially tested in #1609, I identified the following advantages:
- Support for unix sockets out of the box
- Running single-threaded is possible (needed for CI with limited no of cores)
- Supports HTTP/2, once we need it
- No boilerplate at all for tests - using the plain-text HTTP message format
- Written in a compiled language for more req/s

I have only written one very basic test, so far. We are not doing too much with the results either, yet. However, I have added a nice framework around the test runner. We have:
- `postgrest-loadtest` to run the loadtests in the current working tree
- `postgrest-loadtest-against <commit-ish>` to run the loadtests twice for A/B testing. Once on the current working tree and once on the branch that is given by the first argument. It defaults to main, that means `postgrest-loadtest-against` will give you a test run comparing your current branch against main. This, with some fancy number crunching of results, could be a great tool for CI.
- As a side-effect of implementing that, I learned about `git worktree` (quite awesome!) and implemented `postgrest-with-git <commit-ish> <sub-command>`. This will checkout commit-ish in a temporary directory and run the sub-command in it. This not only allows to compare loadtests on different branches, but also things like:
  - I am currently working on a fancy new nix feature (this one here) and [some reviewer suggests a change to one of my open PRs](https://github.com/PostgREST/postgrest/pull/1808#issuecomment-817993627). To make the change without doing anything to my current working tree, I just ran `postgrest-with-git remove-setup-hs nix-shell`. This started a new nix-shell on that branch. Made the change, pushed, `exit`, done. Ok, that's really what you would do with `git worktree` anyway - but I didn't know, so that was quite an awesome thing to do :D.
  - Remo is going on another refactor-spree? Not sure what that does to performance/tests/whatever? Ah, just a quick `postgrest-with-git remo/untangle-the-world postgrest-test-spec-all` will answer that.
  - Just developed some cool new nix tooling, like we always do, but it's not yet committed and you want to run this tool against a different branch? `postgrest-with-git that-other-branch postgrest-my-new-tool` has you covered, because even though it runs in the other branch, it still takes the `postgrest-` command from the current instance of nix-shell. Of course, you can do by just running `nix-shell` on one branch and then checking out the other branch. Once you have uncommitted changes, it becomes a bit complicated with possible merge conflicts, though... `postgrest-with-git` has you covered!

[...]

PR #1819, opened 2021-04-17
nix: Add argbash support + some refactors

wolfgangwalther:
This adds [argbash](https://argbash.readthedocs.io/en/latest/) support to our nix-shell scripts. None of the current scripts use any options or flags, yet, but I'd like to do so in #1805 and #1812. All scripts now have a `--help` output to display the `docs` string we had already in place.

Also added some smaller refactors I did on the other branches already - and some new ones, too.

We still need to figure out how to add argbash auto-complete support, but we can do that later.


nix/overlays/checked-shell-script/checked-shell-script.nix
```diff
@@ -1,17 +1,48 @@
...
+          ${lib.strings.concatMapStrings (arg: "# " + arg) args}
+          # ARG_DEFAULTS_POS()
+          # ARGBASH_GO
+        '';
+
+      checkPhase =
+        ''
+          # run argbash to turn the template into a shell script
+          ${argbash}/bin/argbash $out/parser.sh -o $out/parser.sh
+        '';
```
wolfgangwalther, line 43: This is a bit of a hack, since we're modifying the written script in the `checkPhase`. Should we use a different builder?

monacoremo: It would be a bit cleaner with a separate `runCommand` (we already use that in a few places). It could create a file directly without a directory, putting the output into $out. Same could be done with the textfile, no destination required.

Is there a difference between the sourcing method you chose and the `# [` and `# ]` weirdness I saw in the argbash docs?

monacoremo: Just read the docs, makes sense!

wolfgangwalther: > It would be a bit cleaner with a separate `runCommand` (we already use that in a few places). It could create a file directly without a directory, putting the output into $out. Same could be done with the textfile, no destination required.

So you suggest one textfile `argTemplate` and a runCommand `argParser` or something? We could also make it one runCommand and pipe the template into stdin of `argbash`. WDYT?

> Is there a difference between the sourcing method you chose and the `# [` and `# ]` weirdness I saw in the argbash docs?

That weirdness is about the ability to run argbash against the output file **again**, e.g. after modification. The sourcing method does play a small role, but not that much. Since we're starting from scratch all the time, due to how nix works, we don't need to worry about that.

nix/overlays/checked-shell-script/checked-shell-script.nix
```diff
@@ -1,17 +1,48 @@
...
+      name = "${name}-args";
+      destination = "/parser.sh";
+
+      text =
+        ''
+          # BASH_ARGV0 sets $0 - which is used in parser.sh for usage information
+          # stripping the /nix/store/... path for nicer display
+          BASH_ARGV0="$(basename "$0")"
+
+          # ARG_HELP([${name}], [${lib.strings.stringAsChars (c: if c == "\n" then " \\n" else c) docs}])
```
wolfgangwalther, line 33: I think we'd need to escape `[` and `]`, too, to not make the m4 macro fail. Short of adding some crazy nested `if`s... any other utility function I could use for that?

monacoremo: builtins.stringReplace I think

nix/overlays/checked-shell-script/checked-shell-script.nix
monacoremo: Well need the template for the completion also, think it will be good to have it separare!

wolfgangwalther: Hm. I think I'd run the completion command on the final script, not the "to-be-sourced-args-script", because from [the docs](https://argbash.readthedocs.io/en/latest/usage.html#bash-completion):

> Notice that in this case, the completion file has to “know” the basename of the script the completion is meant for. The basename is inferred either from the source filename, or from the destination by stripping the .m4 suffix if applicable

monacoremo: Hm, not sure - will be best to try it out. It will need some version of the file with the macros in it, no?

> So you suggest one textfile argTemplate and a runCommand argParser or something? We could also make it one runCommand and pipe the template into stdin of argbash. WDYT?

I'd go with the first, but both is fine!

wolfgangwalther: > Hm, not sure - will be best to try it out. It will need some version of the file with the macros in it, no?

Yes, but the macros are copied to the final script, too.

I'm currently looking at running `argbash --type complete` in the `script` phase of checked-shell-script, i.e. where the symbolic link is created - and I wonder, what the reason is to create that symbolic link in the first place? Currently this just creates another `.drv` files to add the link - why are we not executing the script directly?

monacoremo: We are returning two versions of the script at the end: `script // { inherit bin; }`

`script` is easier to use within our scripts, as you can just go `${script}` instead of `${script}/bin/script`. `bin` has the same script in a `bin` subdirectory, which is needed to put our scripts on PATHs. We access that attribute to use it.

When we add bash completion, I would imagine that we have an additional `bashCompletion` attribute (as we do e.g. for `hsie`). The last line will then be:

```
`script // { inherit bin bashCompletion; }`
```

wolfgangwalther: > We are returning two versions of the script at the end: `script // { inherit bin; }`
> 
> `script` is easier to use within our scripts, as you can just go `${script}` instead of `${script}/bin/script`. `bin` has the same script in a `bin` subdirectory, which is needed to put our scripts on PATHs. We access that attribute to use it.

Ah! That's where the magic comes from. Ok, got it.

> When we add bash completion, I would imagine that we have an additional `bashCompletion` attribute (as we do e.g. for `hsie`).

Yes, I came to that conclusion, too. Now, I just need to export all those to `shell.nix`. I think I will create a single completion script for each "module" (devtools, style, ...), which sources all the tools. This would also to move the `_command` completions we currently have in `shell.nix` closer to their shell-scripts.

wolfgangwalther: > > Hm, not sure - will be best to try it out. It will need some version of the file with the macros in it, no?
> 
> Yes, but the macros are copied to the final script, too.

But we're sourcing that part only... argh! I see the problem now ;)

wolfgangwalther:
> We still need to figure out how to add argbash auto-complete support, but we can do that later.

Did it now.

TODO:
- [ ] Add required environment variables for some scripts via `ARG_USE_ENV`
- [ ] Use `ARG_POSITIONAL_DOUBLEDASH` everywhere with requirement to use optional before positional arguments (improves passing arguments to cabal etc)

wolfgangwalther:
Unless CI tells me otherwise... I think I'm done with this for now.

Feedback very much welcome, especially from maybe just running `nix-shell` on this branch, then switching somewhere else and developing regularly for a bit. Might have missed something obvious.

steve-chavez:
Really cool, didn't know about argbash. I'm liking the new `-h` commands and how some commands automatically output the help when they require arguments and 0 are given.

The `-h` doesn't seem to be working on some commands though:

```bash
postgrest-release-github -h
line 25: GITHUB_TOKEN: unbound variable

postgrest-release-dockerhub -h
line 15: DOCKER_REPO: unbound variable

postgrest-release-dockerhubdescription -h
line 12: DOCKER_USER: unbound variable
```

wolfgangwalther:
Hm. I am getting the following output:

```bash
$ postgrest-release-github -h
postgrest-release-github
Usage: postgrest-release-github [-h|--help] [--] <version>
	<version>: git version tag to make release for
	-h, --help: Prints help

Environment variables that are supported:
	GITHUB_TOKEN: GitHub token.
	GITHUB_USERNAME: GitHub user name.
	GITHUB_REPONAME: GitHub repository name.

Push a new release to GitHub.
```

Are you sure, you are on the latest commit of this branch? The last few commits changed some stuff for the `postgrest-release-` commands.

monacoremo:
I love it! works great for me. That will bring our dev environment to another level!

monacoremo (approved):

wolfgangwalther:
Had a few more small changes and refactors to the nix code. Importing stuff into `shell.nix` is now a lot cleaner and all modules have a common interface.

Two changes, for consistency:
- `postgrest-docker-login` is now `postgrest-release-docker-login` (same namespace as the other release tools)
- `nix-shell --arg memoryTests true` is now `nix-shell --arg memory true` because I always forget whether it's `memoryTest` or `memoryTests` - and nix-shell does not complain.

Also fixed `postgrest-push-cachix`, which used system-binaries and didn't work with `nix-shell --pure`.

wolfgangwalther:
Short of CI failures, I'll leave it at that and start rebasing my other two nix PRs on this one.

monacoremo:
https://github.com/PostgREST/postgrest/pull/1819/commits/6d1d0012abce4241a55c3366f6d26c8679fa648e not sure about that one... 'flat is better than nested'. Should we leave it flat? Or alternatively, name the directory 'tools'?

wolfgangwalther:
> [6d1d001](https://github.com/PostgREST/postgrest/commit/6d1d0012abce4241a55c3366f6d26c8679fa648e) not sure about that one... 'flat is better than nested'. Should we leave it flat? Or alternatively, name the directory 'tools'?

Flat is fine - until I can't find things anymore :D. It was a lot of files by now in `nix/`, and since they all share the same kind of interface now, I figured it made sense. Also with loadtest and changelog there's two more files coming in.

I had `tools` first and then wasn't sure about hsie. It's a tools, but not a script. I'd be fine with `tools`, though.

monacoremo:
> and since they all share the same kind of interface now, I figured it made sense. 

Yes, good point.

`hsie` is a special case, might be spun out into a separate repo at some point. Let's go with tools?

PR #1821, opened 2021-04-17
Use cachix authtoken in CI

monacoremo:


monacoremo:
Will need to merge in order to see whether it works.

.circleci/config.yml
```diff
@@ -140,9 +140,16 @@ jobs:
...
+            if [ -n "${CACHIX_AUTH_TOKEN:-""}" ]; then
               echo "Building and caching all derivations..."
+              cachix authtoken "$CACHIX_AUTH_TOKEN"
+
+              # Push new builds as we go
               nix-build | cachix push postgrest
+
+              # Make sure that everything, including .drv files, is pushed
+              nix-env -f default.nix -iA devtools
+              postgrest-push-cachix
```
wolfgangwalther, line 152: Does it need to call cachix twice?

monacoremo: To be both thorough and fast yes - the first one pushes to cachix while we build. That misses some .drv files etc though, so we need the second pass. We could only have the second, but that would mean we waste upload bandwidth while we build

wolfgangwalther: Ah, I see. Cool!

PR #1822, opened 2021-04-18
ci: merge nix-test and style-check; split nix-memory

wolfgangwalther:
I am annoyed by CI performance - especially in the light of #1812, where I'd want to increase loadtest runtime in CI by a fair bit. Hope to be able to make other stuff faster before, otherwise it will be a pain.

First try is reorganizing a few of the jobs we have right now. `style-check` is always done almost immediately, so I thought we could just move it into `nix-test` and split off the whole memory-test stuff instead. Let's see how the fewer required dependencies speed up `nix-test`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,23 @@ jobs:
             cachix use postgrest
       - run:
           name: Install testing scripts
-          command: nix-env -f default.nix -iA style tests withTools
+          command: nix-env -f default.nix -iA cabalTools style tests withTools
+      - restore_cache:
+          keys:
+          - v2-cabal-{{ .Branch }}-{{ .Revision }}
+          - v2-cabal-{{ .Branch }}
+          - v2-cabal-main
+      - run:
+          name: Build postgrest
+          command: postgrest-build exe:postgrest lib:postgrest test:spec test:spec-querycost
+      - run:
+          name: Skip tests on build failure
+          command: circleci-agent step halt
+          when: on_fail
+      - save_cache:
+          paths:
+            - "./dist-newstyle"
+          key: v2-cabal-{{ .Branch }}-{{ .Revision }}
       - run:
           name: Run style check
           command: postgrest-style-check
@@ -180,10 +196,6 @@ jobs:
           name: Run coverage (io tests and spec tests against PostgreSQL 13)
           command: postgrest-coverage
           when: always
-      - run:
-          name: Skip tests on build or primary test failure
-          command: circleci-agent step halt
-          when: on_fail
       - run:
           name: Upload coverage to codecov
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,21 +150,11 @@ jobs:
           path: /tmp/postgrest
 
   # Run tests
-  # Needs to run on a virtual machine, because PostgreSQL needs to run as non-root
   nix-test:
-    machine: true
+    docker:
+      - image: technowledgy/nix:latest
     steps:
       - checkout
-      - run:
-          name: Install Nix
-          command: |
-            curl -L https://nixos.org/nix/install | sh
-            echo "source $HOME/.nix-profile/etc/profile.d/nix.sh" >> $BASH_ENV
-      - run:
-          name: Install and use the Cachix binary cache
-          command: |
-            nix-env -iA cachix -f https://cachix.org/api/v1/install
-            cachix use postgrest
       - run:
           name: Install testing scripts
           command: nix-env -f default.nix -iA cabalTools style tests withTools
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,27 +1,6 @@
 version: 2
 
 jobs:
-  # Make sure that there are no outstanding linting hints and that
-  # auto-formatting does not result in any changes.
-  style-check:
-    docker:
-      - image: nixos/nix:2.3
-    steps:
-      - checkout
-      - run:
-          name: Install linting and styling scripts
-          command: nix-env -f default.nix -iA style
-      - run:
-          name: Run linter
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -157,22 +136,18 @@ jobs:
 
   # Run tests
   nix-test:
-    machine: true
+    docker:
+      - image: nixos/nix:2.3.10
```
monacoremo, line 140: That doesn't work because postgresql needs to run as non-root - easiest way to get there was to use a plain linux machine

wolfgangwalther: Ah, I remember you already told me about that. I remembered *something*, but wasn't sure what exactly. I had hoped that by moving all the "docker" stuff (which might be a problem, too, with docker-in-docker) out to `nix-build`, I'd be lucky... :D

monacoremo: We probably should add a comment :-) the nix docker stuff runs completly without docker, cooking the image from scratch. So no risk there!

wolfgangwalther: > We probably should add a comment :-) 

Already did so.

> the nix docker stuff runs completly without docker, cooking the image from scratch. So no risk there!

ah.. ha! Does that mean I can switch to the docker executor for `nix-build`?

monacoremo: Yes, that should be fine!

monacoremo: More info on the from scratch building here: https://github.com/PostgREST/postgrest/blob/main/nix/tools/docker/README.md

That will yield our 3-4mb docker image next release!

wolfgangwalther:
I tested caching `/nix/store` in circle CI cache. Didn't really succeed completely, but the numbers I got were not very promising. ~15 sec improvment (down from 2 min) when reloading the cache compared to cachix. Now take into account an additional >3 min to upload that cache again as one big archive.... no!

One thing that we could possibly do is creating a custom docker CI image. This would be a nixos based docker image with all the dependencies we need for CI pre-installed. We can even reference this image in our `nix/Dockerfile` as a base. Should speed up things for users of the nix-in-docker based environment. This might not always be up2date, but then nix would only have to pull differences from cachix. We should update this image at least once we update the nixpkgs-version, but maybe even after adding a few dependencies.

To be able to do that, we'd first need to solve the "postgres needs to be non-root inside docker" issue.

monacoremo:
> To be able to do that, we'd first need to solve the "postgres needs to be non-root inside docker" issue.

Yes, I had hacked together an image like that here: 
https://github.com/monacoremo/nix-docker

Used for example here - note: runs in Docker, uses postgres, nix stuff is pre-installed. Haven't maintained it though as the building process was a bit sketchy and not fully worth the hassle

wolfgangwalther:
You know... I guess I must have solved that already, too. Otherwise the `nix/Dockerfile` would not work at all. I think this might be as simple as adding `nix-env -f default.nix -iA cabalTools style tests withTools` to `nix/Dockerfile` and then publishing this...

monacoremo:
> You know... I guess I must have solved that already, too. 

Indeed you did! Sorry for overloading your head with large PRs...

.circleci/config.yml
```diff
@@ -214,23 +201,42 @@ jobs:
...
           command: postgrest-test-spec-idempotence
           when: always
+      - run:
+          name: Run linter
+          command: postgrest-lint
+          when: always
+      - run:
+          name: Run style check
+          command: postgrest-style-check
+          when: always
```
monacoremo, line 211: Can we run those as early as possible? Always a pain to have everything pass but some whitespace is wrong at the end.

wolfgangwalther: I suggest you add a pre-push hook to your git repo running at least style-check - would that be early enough? :D

I'll put them first after the build.

wolfgangwalther: > I suggest you add a pre-push hook to your git repo running at least style-check - would that be early enough? :D

Pre-commit is even earlier!

monacoremo: Snarky! But you're right.. Could we get a `postgrest-install-git-hooks` that sets up the recommended Wolfgang config?

wolfgangwalther:
@monacoremo So, I gave the docker approach a try. I created that docker image and uploaded it to my dockerhub account. Now `nix-test` is using it. The first load of that image took 1:24 minutes, which is ~2/3 of the time it took to load stuff from cachix before. Not sure, yet, how often we will actually get a cache hit, so that the docker image will already be there, but this would improve things even more. This is something, that we'd have to try with a few more runs.

But I'm getting an error in CI during the "Install scripts" step now:
```
error: packages '/nix/store/p2693g42x34j013f81adp7gyjnsij1kz-postgrest-cabal/bin/postgrest-build' and '/nix/store/3ap2wbsy78ibcj4g6p9fi3nmk3fmsxmf-postgrest-cabaltools/bin/postgrest-build' have the same priority 5; use 'nix-env --set-flag priority NUMBER INSTALLED_PKGNAME' to change the priority of one of the conflicting packages (0 being the highest priority)
```

I understand why this happens. I built the docker image on a different local branch, which is slightly different regarding the tooling. Running the following command twice, with conflicting changes, will result in that error:

```
nix-env -f nix/default.nix -iA cabalTools style tests withTools
```

What is the best way to solve that? Can I give a "force" flag to `nix-env` somehow, so that it will overwrite any conflicts on the second run? Or, alternatively, what would happen if I run a `nix-env -e '.*'` during my docker build? This would still keep all the dependencies in the nix-store, but create a clean environment to install tools in, right?

monacoremo:
I adopted the idea of splitting memory-tests in #1901. With the move to GitHub Actions, CI performance is significantly better. Closing this for now as CircleCI has become obsolete, but happy to revisit for the performance tests!

Issue #1647, opened 2020-11-05
Allow option to only generate certain HTTP methods for a DB view


zwennesm:
### Description of issue

As discussed earlier in the Gitter, it would be nice if there would be an option to avoid generation of HTTP methods for certain views. 

### Example:

If you have 1 table which is called `src.orders` and two views on it which are named `api.orders_per_hour` and `api.orders_per_day`. It would be a nice addition if there would be an option to only allow certain HTTP methods (e.g. only GET). So that it's clear for the users of the API that certain methods are not to be used at all. This especially applies if you communicate the API spec with something such as `Swagger`.

(Expected behavior vs actual behavior)

Currently when a view is created, PostgREST automatically generates a GET, POST, DELETE, and PATCH. In cases where the view contains a `group by` it does seem to only create a GET.

wolfgangwalther:
I think this currently depends on whether the view is updateable or not:

https://github.com/PostgREST/postgrest/blob/302d4e15ad9d4689caa132b0ad5cdab828de19da/src/PostgREST/App.hs#L279

A view is updateable if either you have some triggers defined for that or it is auto-updateable. See the [postgres docs](https://www.postgresql.org/docs/current/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS) for when that is the case - the `GROUP BY` is one of the cases where it's not.

So your view seems to be auto-updateable right now. The question then is: How is it possible to prevent auto-updateability?

It seems like a view is not updateable anymore when it contains the `OFFSET` clause. The [docs](https://www.postgresql.org/docs/current/queries-limit.html) for that say:

> OFFSET 0 is the same as omitting the OFFSET clause, as is OFFSET with a NULL argument.

So maybe you can just append `OFFSET NULL` to your view definition. The view should be the same, but it should only have `GET` in the api output.

zwennesm:
Thanks for the suggestion @wolfgangwalther. I've tested your suggestion but it looks as the methods are still being generated. 

Example query:
```
create view api.citizen_crowd_observation as
  select
    a,
    b,
    c
  from
    public.test;
  offset null
```

Switching to a `MATERIALIZED VIEW` does seem to only generate a `GET`. However in this case we would manually need to refresh the `VIEW`

zwennesm:
@wolfgangwalther never mind. I was misreading the docs you suggested. 

[...]

PR #1824, opened 2021-04-19
fix: Allow options to generate certain HTTP methods for a DB view

laurenceisla:
WIP.  It will resolve issue #1647. 

My approach is to check first if the role has the permissions assigned to the DB table/view, only then check if it's insertable, etc.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -358,7 +358,9 @@ handleInfo identifier RequestContext{..} =
     allOrigins = ("Access-Control-Allow-Origin", "*")
     allowH table =
       ( HTTP.hAllow
-      , if tableInsertable table then "GET,POST,PATCH,DELETE" else "GET"
+      , "GET" <> (if tableInsertable table then ",POST" else "")
+              <> (if tableUpdatable table then ",PATCH" else "")
+              <> (if tableDeletable table then ",DELETE" else "")
       )
     tableMatches table =
       tableName table == qiName identifier
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -359,12 +359,11 @@ handleInfo identifier RequestContext{..} =
     allowH table =
       ( HTTP.hAllow
       , BS8.intercalate "," $
-          ["GET"]
+          ["OPTIONS,GET,HEAD"]
           ++ ["POST" | tableInsertable table]
           ++ ["PUT" | tableInsertable table && tableUpdatable table && hasPK]
           ++ ["PATCH" | tableUpdatable table]
           ++ ["DELETE" | tableDeletable table]
-          ++ ["HEAD,OPTIONS"]
       )
     tableMatches table =
       tableName table == qiName identifier
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -358,13 +358,19 @@ handleInfo identifier RequestContext{..} =
     allOrigins = ("Access-Control-Allow-Origin", "*")
     allowH table =
       ( HTTP.hAllow
-      , "GET" <> (if tableInsertable table then ",POST" else "")
-              <> (if tableUpdatable table then ",PATCH" else "")
-              <> (if tableDeletable table then ",DELETE" else "")
+      , BS8.intercalate "," $
+          ["GET"]
+          ++ ["POST" | tableInsertable table]
+          ++ ["PUT" | tableInsertable table && tableUpdatable table && hasPK]
+          ++ ["PATCH" | tableUpdatable table]
+          ++ ["DELETE" | tableDeletable table]
+          ++ ["HEAD,OPTIONS"]
       )
     tableMatches table =
       tableName table == qiName identifier
       && tableSchema table == qiSchema identifier
+    hasPK =
+      not $ null $ tablePKCols ctxDbStructure (qiSchema identifier) (qiName identifier)
 
 handleInvoke :: InvokeMethod -> ProcDescription -> RequestContext -> DbHandler Wai.Response
 handleInvoke invMethod proc context@RequestContext{..} = do
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -311,20 +311,27 @@ accessibleTables =
       relname as table_name,
       d.description as table_description,
       (
-        c.relkind in ('r', 'v', 'f')
-        and (pg_relation_is_updatable(c.oid::regclass, false) & 8) = 8
-        -- The function `pg_relation_is_updateable` returns a bitmask where 8
-        -- corresponds to `1 << CMD_INSERT` in the PostgreSQL source code, i.e.
-        -- it's possible to insert into the relation.
-        or (exists (
-          select 1
-          from pg_trigger
-          where
-            pg_trigger.tgrelid = c.oid
-            and (pg_trigger.tgtype::integer & 69) = 69)
-            -- The trigger type `tgtype` is a bitmask where 69 corresponds to
-            -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_INSERT
-            -- in the PostgreSQL source code.
+        (
+          has_table_privilege(c.oid, 'INSERT,UPDATE,DELETE,TRUNCATE')
+          or has_any_column_privilege(c.oid, 'INSERT, UPDATE')
+        ) 
+        and 
+        (
+          c.relkind in ('r', 'v', 'f')
+          and (pg_relation_is_updatable(c.oid::regclass, false) & 8) = 8
+          -- The function `pg_relation_is_updateable` returns a bitmask where 8
+          -- corresponds to `1 << CMD_INSERT` in the PostgreSQL source code, i.e.
+          -- it's possible to insert into the relation.
+          or (exists (
+            select 1
+            from pg_trigger
+            where
+              pg_trigger.tgrelid = c.oid
+              and (pg_trigger.tgtype::integer & 69) = 69)
+              -- The trigger type `tgtype` is a bitmask where 69 corresponds to
+              -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_INSERT
+              -- in the PostgreSQL source code.
+          )
         )
       ) as insertable
     from
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -118,6 +118,8 @@ decodeTables =
                  <*> column HD.text
                  <*> nullableColumn HD.text
                  <*> column HD.bool
+                 <*> column HD.bool
+                 <*> column HD.bool
 
 decodeColumns :: [Table] -> HD.Result [Column]
 decodeColumns tables =
@@ -312,8 +314,8 @@ accessibleTables =
       d.description as table_description,
       (
         (
-          has_table_privilege(c.oid, 'INSERT,UPDATE,DELETE,TRUNCATE')
-          or has_any_column_privilege(c.oid, 'INSERT, UPDATE')
+          has_table_privilege(c.oid, 'INSERT')
+          or has_any_column_privilege(c.oid, 'INSERT')
         ) 
         and 
         (
@@ -333,7 +335,46 @@ accessibleTables =
               -- in the PostgreSQL source code.
           )
         )
-      ) as insertable
+      ) as insertable,
+      (
+        (
+          has_table_privilege(c.oid, 'UPDATE')
+          or has_any_column_privilege(c.oid, 'UPDATE')
+        )
+        and
+        (
+          c.relkind in ('r', 'v', 'f')
+          and (pg_relation_is_updatable(c.oid::regclass, false) & 20) = 20
+          -- (1 << CMD_UPDATE) + (1 << CMD_DELETE) = 20
+          or (exists (
+            select 1
+            from pg_trigger
+            where
+              pg_trigger.tgrelid = c.oid
+              and (pg_trigger.tgtype::integer & 81) = 81)
+              -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_UPDATE
+          )
+        )
+      ) as updatable,
+      (
+        (
+          has_table_privilege(c.oid, 'DELETE,TRUNCATE')
+        )
+        and
+        (
+          c.relkind in ('r', 'v', 'f')
+          and (pg_relation_is_updatable(c.oid::regclass, false) & 20) = 20
+          -- (1 << CMD_UPDATE) + (1 << CMD_DELETE) = 20
+          or (exists (
+            select 1
+            from pg_trigger
+            where
+              pg_trigger.tgrelid = c.oid
+              and (pg_trigger.tgtype::integer & 73) = 73)
+              -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_UPDATE
+          )
+        )
+      ) as deletable
     from
       pg_class c
       join pg_namespace n on n.oid = c.relnamespace
@@ -459,6 +500,11 @@ allTables =
       c.relname AS table_name,
       NULL AS table_description,
       (
+        has_table_privilege(c.oid, 'INSERT,UPDATE,DELETE,TRUNCATE')
+        or has_any_column_privilege(c.oid, 'INSERT, UPDATE')
+      )
+      and
+      (
         c.relkind IN ('r', 'v','f')
         AND (pg_relation_is_updatable(c.oid::regclass, FALSE) & 8) = 8
         OR EXISTS (
@@ -468,12 +514,14 @@ allTables =
             pg_trigger.tgrelid = c.oid
             AND (pg_trigger.tgtype::integer & 69) = 69
         )
-      ) AS insertable
+      ) AS insertable,
+      true as updatable,
+      true as deletable
     FROM pg_class c
     JOIN pg_namespace n ON n.oid = c.relnamespace
     WHERE c.relkind IN ('v','r','m','f')
       AND n.nspname NOT IN ('pg_catalog', 'information_schema')
-    GROUP BY table_schema, table_name, insertable
+    GROUP BY table_schema, table_name, insertable, updatable, deletable
     ORDER BY table_schema, table_name |]
 
 allColumns :: [Table] -> Bool -> H.Statement [Schema] [Column]
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -313,66 +313,38 @@ accessibleTables =
       relname as table_name,
       d.description as table_description,
       (
-        (
-          has_table_privilege(c.oid, 'INSERT')
-          or has_any_column_privilege(c.oid, 'INSERT')
+        c.relkind IN ('r', 'v','f')
+        AND (pg_relation_is_updatable(c.oid::regclass, FALSE) & 8) = 8
+        OR EXISTS (
+          SELECT 1
+          FROM pg_trigger
+          WHERE
+            pg_trigger.tgrelid = c.oid
+            AND (pg_trigger.tgtype::integer & 69) = 69
         )
-        and
-        (
-          c.relkind in ('r', 'v', 'f')
-          and (pg_relation_is_updatable(c.oid::regclass, false) & 8) = 8
-          -- The function `pg_relation_is_updateable` returns a bitmask where 8
-          -- corresponds to `1 << CMD_INSERT` in the PostgreSQL source code, i.e.
-          -- it's possible to insert into the relation.
-          or (exists (
-            select 1
-            from pg_trigger
-            where
-              pg_trigger.tgrelid = c.oid
-              and (pg_trigger.tgtype::integer & 69) = 69)
-              -- The trigger type `tgtype` is a bitmask where 69 corresponds to
-              -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_INSERT
-              -- in the PostgreSQL source code.
-          )
-        )
-      ) as insertable,
+      ) AS insertable,
       (
-        (
-          has_table_privilege(c.oid, 'UPDATE')
-          or has_any_column_privilege(c.oid, 'UPDATE')
-        )
-        and
-        (
-          c.relkind in ('r', 'v', 'f')
-          and (pg_relation_is_updatable(c.oid::regclass, false) & 20) = 20
-          -- (1 << CMD_UPDATE) + (1 << CMD_DELETE) = 20
-          or (exists (
-            select 1
-            from pg_trigger
-            where
-              pg_trigger.tgrelid = c.oid
-              and (pg_trigger.tgtype::integer & 81) = 81)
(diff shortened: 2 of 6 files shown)


src/PostgREST/DbStructure.hs
```diff
@@ -461,12 +514,14 @@ allTables =
...
-      ) AS insertable
+      ) AS insertable,
+      true as updatable,
+      true as deletable
     FROM pg_class c
     JOIN pg_namespace n ON n.oid = c.relnamespace
     WHERE c.relkind IN ('v','r','m','f')
       AND n.nspname NOT IN ('pg_catalog', 'information_schema')
-    GROUP BY table_schema, table_name, insertable
+    GROUP BY table_schema, table_name, insertable, updatable, deletable
```
wolfgangwalther, line 524: I wonder whether we need this `GROUP BY` at all. I don't think so. We don't have any aggregates and I can't see how we'd get multiple lines from `pg_class JOIN pg_namespace`. You should be able to safely delete the whole line.

src/PostgREST/DbStructure.hs
```diff
@@ -451,6 +499,11 @@ allTables =
       n.nspname AS table_schema,
       c.relname AS table_name,
       NULL AS table_description,
+      (
+        has_table_privilege(c.oid, 'INSERT,UPDATE,DELETE,TRUNCATE')
+        or has_any_column_privilege(c.oid, 'INSERT, UPDATE')
+      )
```
wolfgangwalther, line 505: I don't think this bit should be added here. That's in the `allTables` query, which should not be concerned about permissions - just about the general insertability, updatability and deletability. Those values should then be used in the `OPTIONS` request in the `Allow` header: https://tools.ietf.org/html/rfc7231#section-7.4.1

I read this as the returned values should only depend on the methods supported by the entity - not limited to those with privileges for the current user. @steve-chavez WDYT?

src/PostgREST/DbStructure.hs
```diff
@@ -311,22 +313,68 @@ accessibleTables =
...
-          where
-            pg_trigger.tgrelid = c.oid
-            and (pg_trigger.tgtype::integer & 69) = 69)
-            -- The trigger type `tgtype` is a bitmask where 69 corresponds to
-            -- TRIGGER_TYPE_ROW + TRIGGER_TYPE_INSTEAD + TRIGGER_TYPE_INSERT
-            -- in the PostgreSQL source code.
+        (
+          has_table_privilege(c.oid, 'INSERT')
+          or has_any_column_privilege(c.oid, 'INSERT')
+        )
```
wolfgangwalther, line 319: I'm not sure about adding those lines here, either. I think this would only affect OpenAPI output anyway - but I'm not even sure in which way. I had a quick glance at `OpenApi.hs` but didn't understand immediately what was going on with the insertable flag.

I suggest to leave any changes to OpenApi out of this PR and focus on the `OPTION` request's `Allow` header (see other comment).

If you decide to make this query similar to the `allTables` query, there's also a nice possibility to refactor:
What about making `accessibleTables` just take the query from `allTables` and then adding the two `WHERE` conditions, e.g. via argument or so? Would remove a lot of duplicated code here. Kind of similar to what `accessibleProcs` does.

src/PostgREST/DbStructure.hs
```diff
@@ -311,22 +313,68 @@ accessibleTables =
...
+        (
+          has_table_privilege(c.oid, 'UPDATE')
+          or has_any_column_privilege(c.oid, 'UPDATE')
         )
-      ) as insertable
+        and
+        (
+          c.relkind in ('r', 'v', 'f')
+          and (pg_relation_is_updatable(c.oid::regclass, false) & 20) = 20
+          -- (1 << CMD_UPDATE) + (1 << CMD_DELETE) = 20
```
wolfgangwalther, line 348: Shouldn't this be 4 or 16 instead of 20? (Just guessing - but not a combo of update and delete, right?)

src/PostgREST/DbStructure.hs
```diff
@@ -311,22 +313,68 @@ accessibleTables =
...
+      ) as updatable,
+      (
+        (
+          has_table_privilege(c.oid, 'DELETE,TRUNCATE')
+        )
+        and
+        (
+          c.relkind in ('r', 'v', 'f')
+          and (pg_relation_is_updatable(c.oid::regclass, false) & 20) = 20
+          -- (1 << CMD_UPDATE) + (1 << CMD_DELETE) = 20
```
wolfgangwalther, line 367: Same here:
Shouldn't this be 4 or 16 instead of 20?

src/PostgREST/DbStructure.hs
wolfgangwalther: I don't even think it will work here as expected, because those privileges will only be checked for the authenticator role.

steve-chavez: > I read this as the returned values should only depend on the methods supported by the entity - not limited to those with privileges for the current user.

@wolfgangwalther But according to your comment here https://github.com/PostgREST/postgrest/issues/1647#issuecomment-731846689, we do need to look at the privileges for restricting certain http methods, right?

Hm, besides restricting OPTIONS, would we also need to disallow HTTP methods and respond with `405 Method Not Allowed`? That would avoid hitting the db in case of missing privileges, but I'm not sure if we should do it. It would make development a bit more complicated - the user would have to refresh PostgREST schema cache when changing privileges on the db.

Perhaps the `postgrest-contrib` approach mentioned on https://github.com/PostgREST/postgrest/pull/1710#issuecomment-808788905 would be more fit for users that want to restrict the http methods.

src/PostgREST/DbStructure.hs
laurenceisla: I was using the [information schema](https://github.com/postgres/postgres/blob/master/src/backend/catalog/information_schema.sql) in postgres source code as reference, had the idea it returned both updatable and deletable together, but checking in more detail [the source code](https://doxygen.postgresql.org/rewriteHandler_8c.html#a52a3ea381d9db1b6e917b7280e3e6d71) it may discriminate the UPDATE/INSERT and DELETE events for views. 
In any case, I agree, it should be as you say, I'll fix it.

src/PostgREST/DbStructure.hs
laurenceisla: > That's in the allTables query, which should not be concerned about permissions

Yes, I had doubts about modifying it. Won't add privileges check here.

src/PostgREST/DbStructure.hs
laurenceisla: > I suggest to leave any changes to OpenApi out of this PR and focus on the OPTION request's Allow header (see other comment).

OK, I get the idea now. The issue mentioned the problem communicating the Api spec using Swagger though, that's why I tried modifying OpenApi too.

> What about making accessibleTables just take the query from allTables and then adding the two WHERE conditions

Seems good, I'll do that.

src/PostgREST/DbStructure.hs
wolfgangwalther: > @wolfgangwalther But according to your comment here [#1647 (comment)](https://github.com/PostgREST/postgrest/issues/1647#issuecomment-731846689), we do need to look at the privileges for restricting certain http methods, right?

No. The whole comment is about `OPTIONS`, which is handled by `handlerInfo`. This does not access the database at all, so it's role-independent. The question here is not whether insert/update/delete are **allowed**, but if they are technically **possible** at all. This is only relevant for views, that are not auto-updateable. In this case, the possibilty to do anything is dependent on the availability of a `INSTEAD OF` trigger.

> Hm, besides restricting OPTIONS, would we also need to disallow HTTP methods and respond with 405 Method Not Allowed? That would avoid hitting the db in case of missing privileges, but I'm not sure if we should do it. It would make development a bit more complicated - the user would have to refresh PostgREST schema cache when changing privileges on the db.

Hm. I think any privilege-related thing needs to be determined at run-time, not by schema cache. We're not keeping track of different schema caches for different users. Even the OpenAPI output is making a separate query on each request to determine the current permissions. So if you want to take permissions into account, you'd need to hit the database anyway. But looking at just the technical possibility to do an `UPDATE`, `DELETE` or `INSERT` - yeah, we could block those before they even get to the database - and that would not be as inconveniet, because you'd need to refresh the schema cache after adding a trigger. That seems reasonable. I suggested a refactor of our whole routing somewhere else - this would be easily included in that kind of refactor. So I would not go for it independently right now - not worth the effort right now.

wolfgangwalther: > > That's in the allTables query, which should not be concerned about permissions
> 
> Yes, I had doubts about modifying it. Won't add privileges check here.

Following my argument above, we should not add them anywhere. Otherwise the fields on `Table` would have different meaning depending on whether the query was made with `accessibleTables` or `allTables` - that seems really odd.

src/PostgREST/DbStructure.hs
wolfgangwalther: Those constants are defined here:

https://github.com/postgres/postgres/blob/e717a9a18b2e34c9c40e5259ad4d31cd7e420750/src/include/nodes/nodes.h#L682-L693

So, `CMD_UPDATE` seems to be 4 and `CMD_DELETE` 16. Agree?

src/PostgREST/DbStructure.hs
wolfgangwalther: > OK, I get the idea now. The issue mentioned the problem communicating the Api spec using Swagger though, that's why I tried modifying OpenApi too.

Background regarding OpenAPI: Every change we make here now is basically a waste of time, because want to split this off into `postgrest-contrib` (to be created) as SQL, see #1698.

src/PostgREST/DbStructure.hs
wolfgangwalther: > Ah, got it. So if it's about INSTEAD OF triggers, then only querying `information_schema.triggers`(or an internal pg table) would be needed right? No need to look into the view's privileges.

Yeah, the query for that is all sorted (it worked fine for `INSERT` before and just needed to be adapted for `UPDATE` or `DELETE`). My question was actually not about the PG side of things but only about the best HTTP semantics here - basically whether my interpretation of the RFC I cited above was correct or not:

Should an `OPTIONS` request return the `Allow` header dependent on user privileges or not? My answer was no - do you agree?

steve-chavez: > Should an OPTIONS request return the Allow header dependent on user privileges or not? My answer was no - do you agree?

Agree. This reminds me of this [comment from Roy Fielding](https://roy.gbiv.com/untangled/2008/rest-apis-must-be-hypertext-driven#comment-732):

> HTTP operations are generic: they are allowed or not, per resource, but they are always **valid**. Hypertext doesn’t usually tell you all the operations allowed on any given resource; it tells you which operation to use for each potential transition.

And the [Allow RFC](https://tools.ietf.org/html/rfc7231#section-7.4.1) says:

> The purpose of this field is strictly to inform the recipient of **valid** request methods associated with the resource.

So yes, we're informing the recipient of the **valid** methods. Allowance is defined at run-time and if the resource replies with 401, then the recipient could request higher privileges.

This also makes me think about RPC, it should not include GET(when it's not IMMUTABLE/STABLE) to avoid `on a read-only transaction` errors. Actually I've realized that right now RPC doesn't allow OPTIONS.

wolfgangwalther: > This also makes me think about RPC, it should not include GET(when it's not IMMUTABLE/STABLE) to avoid `on a read-only transaction` errors. Actually I've realized that right now RPC doesn't allow OPTIONS.

Ah good point. And it's not too easy to add, either, because of overloaded functions. However, with the routing-refactor it should be easily doable, too. Just need to make sure we keep track of that!

src/PostgREST/DbStructure.hs
```diff
@@ -311,22 +313,40 @@ accessibleTables =
...
+          SELECT 1
+          FROM pg_trigger
+          WHERE
+            pg_trigger.tgrelid = c.oid
+            and (pg_trigger.tgtype::integer & 81) = 81
+        )
+      ) as updatable,
+      (
+        c.relkind IN ('r', 'v','f')
+        AND (pg_relation_is_updatable(c.oid::regclass, FALSE) & 16) = 16
```
steve-chavez, line 340: These `c.relkind` and `pg_relation_is_updatable` parts are repeated in all the subqueries. Would it be possible to put them in a CTE?

src/PostgREST/App.hs
```diff
@@ -358,7 +358,9 @@ handleInfo identifier RequestContext{..} =
     allOrigins = ("Access-Control-Allow-Origin", "*")
     allowH table =
       ( HTTP.hAllow
-      , if tableInsertable table then "GET,POST,PATCH,DELETE" else "GET"
+      , "GET" <> (if tableInsertable table then ",POST" else "")
+              <> (if tableUpdatable table then ",PATCH" else "")
+              <> (if tableDeletable table then ",DELETE" else "")
```
steve-chavez, line 363: How about:
```haskell
intercalate "," $
   ["GET"] ++ ["POST" | tableInsertable] ++ ["PATCH" | tableUpdatable] ++ ["DELETE" | tableDeletable]
```

src/PostgREST/DbStructure.hs
wolfgangwalther: How would you do that? They all compare against a different bitmask.

src/PostgREST/DbStructure/Table.hs
```diff
@@ -22,6 +22,8 @@ data Table = Table
   , tableName        :: TableName
   , tableDescription :: Maybe Text
   , tableInsertable  :: Bool
+  , tableUpdatable   :: Bool
+  , tableDeletable   :: Bool
```
steve-chavez, line 26: Perhaps there should be a comment mentioning that these fields are not related to privileges but to view capabilities

src/PostgREST/App.hs
laurenceisla: Looks cleaner, I'll give it a try.

src/PostgREST/DbStructure/Table.hs
laurenceisla: Yes, I was confused at first when looking at it. I think a comment won't hurt.

src/PostgREST/DbStructure.hs
laurenceisla: > They all compare against a different bitmask.

I also think it wouldn't be possible in this case. The `c.relkind` on the otherhand may be, but I don't know if it's worth the use of a CTE only for a line of code.

On a side note, when looking at the [source code,](https://doxygen.postgresql.org/rewriteHandler_8c.html#a52a3ea381d9db1b6e917b7280e3e6d71) postgres treats ordinary tables, that is `c.relkind = 'r'`, as always updatable with all events. Maybe it could be better to verify first if it's an ordinary table and use `pg_relation_is_updatable` only for views, something like [this](https://github.com/postgres/postgres/blob/master/src/backend/catalog/information_schema.sql#L1989-L1993).

src/PostgREST/App.hs
wolfgangwalther: How are we going to handle `PUT`? Should we just say `insertable + updatable = PUT`?

wolfgangwalther: Actually.. we'd need the presence of a PK, too, right?

wolfgangwalther: And... for completeness, we should probably add `HEAD` and `OPTIONS` by default, as well!

src/PostgREST/DbStructure.hs
laurenceisla: > If you decide to make this query similar to the allTables query, there's also a nice possibility to refactor:
What about making accessibleTables just take the query from allTables and then adding the two WHERE conditions, e.g. via argument or so? Would remove a lot of duplicated code here. Kind of similar to what accessibleProcs does.

The refactor result won't be as smooth as `accessibleProcs`, `accessibleTables` has an extra Join to `pg_catalog.pg_description` and uses the description field [here](https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/DbStructure.hs#L330), in `allTables` it's a [NULL value](https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/DbStructure.hs#L471). 

Would it be OK for that extra JOIN to be used by `allTables` too? Or should it be conditionally added to `accessibleTables` only?

wolfgangwalther: I was about to say that it's no problem to add the description to the `allTables` query, because it will go away completely once we remove OpenApi from core. But the same holds throw for the whole `accessibleTables` query. So... it probably doesn't make too much sense to actually refactor it, when we will remove it soon.

laurenceisla: Makes sense. Will leave it as it is.

src/PostgREST/App.hs
```diff
@@ -358,11 +358,19 @@ handleInfo identifier RequestContext{..} =
...
     allowH table =
       ( HTTP.hAllow
-      , if tableInsertable table then "GET,POST,PATCH,DELETE" else "GET"
+      , BS8.intercalate "," $
+          ["GET"]
+          ++ ["POST" | tableInsertable table]
+          ++ ["PUT" | tableInsertable table && tableUpdatable table && hasPK]
+          ++ ["PATCH" | tableUpdatable table]
+          ++ ["DELETE" | tableDeletable table]
+          ++ ["HEAD,OPTIONS"]
```
wolfgangwalther, line 367:
```suggestion
          ["OPTIONS,GET,HEAD"]
          ++ ["POST" | tableInsertable table]
          ++ ["PUT" | tableInsertable table && tableUpdatable table && hasPK]
          ++ ["PATCH" | tableUpdatable table]
          ++ ["DELETE" | tableDeletable table]
```

wolfgangwalther:
Nice work, I like it! :)

test/Feature/OptionsSpec.hs
```diff
@@ -16,11 +16,40 @@ spec = describe "Allow header" $ do
...
+        matchHeader "Allow" "GET,POST,PUT,PATCH,DELETE,HEAD,OPTIONS"
 
   it "includes read verbs for read-only table" $ do
     r <- request methodOptions "/has_count_column" [] ""
     liftIO $
       simpleHeaders r `shouldSatisfy`
-        matchHeader "Allow" "GET"
+        matchHeader "Allow" "GET,HEAD,OPTIONS"
 
+  it "includes read verbs for non auto updatable views" $ do
```
wolfgangwalther, line 27: I think we need a test for an auto-updatable view, too - otherwise this code-path in the SQL query is not really tested.

test/Feature/OptionsSpec.hs
```diff
@@ -16,11 +16,40 @@ spec = describe "Allow header" $ do
...
+    r <- request methodOptions "/projects_view_without_triggers" [] ""
+    liftIO $
+      simpleHeaders r `shouldSatisfy`
+        matchHeader "Allow" "GET,HEAD,OPTIONS"
+
+  it "includes read/write verbs for insertable, updatable and deletable views" $ do
+    r <- request methodOptions "/projects_view_with_all_triggers" [] ""
+    liftIO $
+      simpleHeaders r `shouldSatisfy`
+        matchHeader "Allow" "GET,POST,PUT,PATCH,DELETE,HEAD,OPTIONS"
```
wolfgangwalther, line 37: And we need a test-case for a view that is updateable, but does not have a PK (just don't select the PK columns) - this should then return `POST` and `PATCH`, but not `PUT`.

laurenceisla: I organized the tests to make them more readable, and added the ones that were missing for views which are insertable and updatable but without pk.

test/Feature/OptionsSpec.hs
```diff
@@ -12,44 +12,66 @@ import SpecHelper
...
-        matchHeader "Allow" "GET,DELETE,HEAD,OPTIONS"
+  context "a table" $ do
+    it "includes read/write verbs for writeable table" $ do
+      r <- request methodOptions "/items" [] ""
+      liftIO $
+        simpleHeaders r `shouldSatisfy`
+          matchHeader "Allow" "OPTIONS,GET,HEAD,POST,PUT,PATCH,DELETE"
+
+    it "includes read verbs for read-only table" $ do
+      r <- request methodOptions "/has_count_column" [] ""
```
steve-chavez: `has_count_column` is a view though, right?

wolfgangwalther: Yeah I think this test-case is redundant now. It's the same as: `it "includes read verbs for non auto updatable views" $ do`

laurenceisla: True, that shouldn't be there. I'll remove it entirely, that case is included in views tests.

test/fixtures/schema.sql
```diff
@@ -1544,6 +1544,73 @@ $$A materialized view for projects
...
+    for each row execute procedure test_for_views_with_triggers();
+
+create trigger projects_view_with_all_triggers_without_pk_delete
+    instead of delete on test.projects_view_with_all_triggers_without_pk
+    for each row execute procedure test_for_views_with_triggers();
+
+create view test.projects_view_with_insert_trigger as
+select distinct id, name, client_id from test.projects;
+
+create trigger projects_view_with_update_trigger_insert
```
steve-chavez, line 1596: Shouldn't this trigger be called `projects_view_with_insert_trigger_insert`?

steve-chavez:
@laurenceisla The Allow header is looking good :+1:. However, I think we're missing something. I've just tried this:

```bash
http POST localhost:3000/projects_view_with_delete_trigger <<JSON
{"id": 6, "name": "client 6"}
JSON

{
    "code": "55000",
    "details": "Views containing DISTINCT are not automatically updatable.",
    "hint": "To enable inserting into the view, provide an INSTEAD OF INSERT trigger or an unconditional ON INSERT DO INSTEAD rule.",
    "message": "cannot insert into view \"projects_view_with_delete_trigger\""
}
```

According to the [RFC](https://tools.ietf.org/html/rfc7231#section-7.4.1), we should reply with a `405 Method Not Allowed` instead.

If we do this, the client request wouldn't not reach the db. The disadvantage I see is that if users add a INSTEAD OF INSERT trigger, they'll need to refresh our schema cache to make the request succeed.

WDYT? Shall we do this?

wolfgangwalther:
> According to the [RFC](https://tools.ietf.org/html/rfc7231#section-7.4.1), we should reply with a `405 Method Not Allowed` instead.
> 
> If we do this, the client request wouldn't not reach the db. The disadvantage I see is that if users add a INSTEAD OF INSERT trigger, they'll need to refresh our schema cache to make the request succeed.
> 
> WDYT? Shall we do this?
> 
> **Edit**:
> 
> I was thinking that perhaps we can just catch the `55000`(object_not_in_prerequisite_state) error and convert it to a `405 Method Not Allowed`, but seems this code is used for [other](https://stackoverflow.com/questions/16444622/postgresql-trigger-raises-error-55000) [internal](https://stackoverflow.com/questions/47332752/error-55000-object-not-in-prerequisite-state-when-trying-to-refresh-a-material) errors as well.

Ah, an easy way out would have been cool. But given that this seems not possible to do, I think this will be out of scope for this PR, as discussed here:
- https://github.com/PostgREST/postgrest/pull/1824#discussion_r616777035
- https://github.com/PostgREST/postgrest/pull/1824#discussion_r616798461

steve-chavez:
> But given that this seems not possible to do, I think this will be out of scope for this PR, as discussed here

Cool. I do think that anything that makes PostgREST less transparent to the db requires more discussion anyway. So lets leave that for another time.

steve-chavez (approved):

Issue #1587, opened 2020-09-22
Resource embedding on views doesn't work on N-to-N relations tables


Pigeo:
### Environment

* PostgreSQL version: PostgreSQL 12.4 (Debian 12.4-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
* PostgREST version: PostgREST 7.0.1
* Operating system: Debian 10 (buster) / Linux LABO-BDD-dev 4.15.18-27-pve #1 SMP PVE 4.15.18-55 (Tue, 17 Mar 2020 15:32:02 +0100) x86_64 GNU/Linux

### Description of issue

Trying to reproduce the same database of films than [in the documentation](http://postgrest.org/en/v7.0.0/api.html#resource-embedding), it turns out that resource embedding on views doesn't always work as explained in the documentation (however, when replacing the views with tables, it then works fine and as expected / explained in the documentation)

### Steps to reproduce:

1. In your private schema, create the following tables (mimicking the database of films from PostgREST's documentation) :

```sql
CREATE TABLE "actors" (
    "id" integer NOT NULL,
    "name" character varying(255) NOT NULL,
    CONSTRAINT "actors_id" PRIMARY KEY ("id")
);

INSERT INTO "actors" ("id", "name") VALUES
('1',	'john'),
('2',	'mary'),
('3',	'bob');

CREATE TABLE "films" (
    "id" integer NOT NULL,
    "title" character varying(255) NOT NULL,
    CONSTRAINT "films_id" PRIMARY KEY ("id")
);

INSERT INTO "films" ("id", "title") VALUES
('12',	'douze commandements'),
('2001',	'odyssée de l''espace');

CREATE TABLE "personnages" (
    "film_id" integer NOT NULL,
    "role_id" integer NOT NULL,
    "character" character varying(255) NOT NULL,
    CONSTRAINT "personnages_film_id_role_id" PRIMARY KEY ("film_id", "role_id"),
    CONSTRAINT "personnages_film_id_fkey" FOREIGN KEY (film_id) REFERENCES films(id) NOT DEFERRABLE,
    CONSTRAINT "personnages_role_id_fkey" FOREIGN KEY (role_id) REFERENCES actors(id) NOT DEFERRABLE
);

CREATE INDEX "personnages_film_id" ON "personnages" USING btree ("film_id");

[...]

PR #1825, opened 2021-04-19
fix: disregard internal junction when embedding

steve-chavez:
Fixes https://github.com/PostgREST/postgrest/issues/1736. Also fixes https://github.com/PostgREST/postgrest/issues/1587.

While working on the pure json schema cache, we noticed that the schema cache contains private db objects: https://github.com/PostgREST/postgrest/pull/1794#issuecomment-815264688.

The solution was to filter the private objects from the json output https://github.com/PostgREST/postgrest/pull/1794#issuecomment-815264688. However, I noticed that this fix would be better applied at a lower level(DbStructure), and thus also solve other issues, like #1736.

It should also increase embedding speed, since there are less relationships to search.

#### Pending

- [x] Add a test case for solving https://github.com/PostgREST/postgrest/issues/1587 as well.
- [x] Quick test on the big schema to see if there's a noticeable perf loss on startup.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -43,19 +43,19 @@ import Data.Set                      as S (fromList)
 import Data.Text                     (split)
 import Text.InterpolatedString.Perl6 (q)
 
-import PostgREST.DbStructure.Identifiers (QualifiedIdentifier (..),
-                                          Schema, TableName)
-import PostgREST.DbStructure.PgVersion   (PgVersion (..))
-import PostgREST.DbStructure.Proc        (PgArg (..), PgType (..),
-                                          ProcDescription (..),
-                                          ProcVolatility (..),
-                                          ProcsMap, RetType (..))
-import PostgREST.DbStructure.Relation    (Cardinality (..),
-                                          ForeignKey (..),
-                                          Junction (..),
-                                          PrimaryKey (..),
-                                          Relation (..))
-import PostgREST.DbStructure.Table       (Column (..), Table (..))
+import PostgREST.DbStructure.Identifiers  (QualifiedIdentifier (..),
+                                           Schema, TableName)
+import PostgREST.DbStructure.PgVersion    (PgVersion (..))
+import PostgREST.DbStructure.Proc         (PgArg (..), PgType (..),
+                                           ProcDescription (..),
+                                           ProcVolatility (..),
+                                           ProcsMap, RetType (..))
+import PostgREST.DbStructure.Relationship (Cardinality (..),
+                                           ForeignKey (..),
+                                           Junction (..),
+                                           PrimaryKey (..),
+                                           Relationship (..))
+import PostgREST.DbStructure.Table        (Column (..), Table (..))
 
 import Protolude        hiding (toS)
 import Protolude.Conv   (toS)
@@ -63,12 +63,12 @@ import Protolude.Unsafe (unsafeHead)
 
 
 data DbStructure = DbStructure
-  { dbTables      :: [Table]
-  , dbColumns     :: [Column]
-  , dbRelations   :: [Relation]
-  , dbPrimaryKeys :: [PrimaryKey]
-  , dbProcs       :: ProcsMap
-  , pgVersion     :: PgVersion
+  { dbTables        :: [Table]
+  , dbColumns       :: [Column]
+  , dbRelationships :: [Relationship]
+  , dbPrimaryKeys   :: [PrimaryKey]
+  , dbProcs         :: ProcsMap
+  , pgVersion       :: PgVersion
   }
   deriving (Generic, JSON.ToJSON)
 
@@ -104,7 +104,7 @@ getDbStructure schemas extraSearchPath pgVer prepared = do
   return DbStructure {
       dbTables = tabs
     , dbColumns = cols'
-    , dbRelations = rels
+    , dbRelationships = rels
     , dbPrimaryKeys = keys'
     , dbProcs = procs
     , pgVersion = pgVer
@@ -135,7 +135,7 @@ decodeColumns tables =
       <*> nullableColumn HD.text
       <*> nullableColumn HD.text
 
-decodeRels :: [Table] -> [Column] -> HD.Result [Relation]
+decodeRels :: [Table] -> [Column] -> HD.Result [Relationship]
 decodeRels tables cols =
   mapMaybe (relFromRow tables cols) <$> HD.rowList relRow
  where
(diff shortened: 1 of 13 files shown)


steve-chavez:
After this change, I expected to be able to delete this line on `findRel`:

https://github.com/PostgREST/postgrest/blob/7ca0d469367d480804b2b8b91bdceb653ada754a/src/PostgREST/Request/DbRequestBuilder.hs#L177-L178

However, deleting it makes 6 tests fail(most pass though); these are related to a `sourceCTEName("pgrst_source")` hack we do here:

https://github.com/PostgREST/postgrest/blob/7ca0d469367d480804b2b8b91bdceb653ada754a/src/PostgREST/Request/DbRequestBuilder.hs#L72-L88

I didn't want to make an unrelated change, so I left that line untouched.

src/PostgREST/DbStructure.hs
```diff
@@ -63,12 +63,12 @@ import Protolude.Unsafe (unsafeHead)
...
 data DbStructure = DbStructure
-  { dbTables      :: [Table]
-  , dbColumns     :: [Column]
-  , dbRelations   :: [Relation]
-  , dbPrimaryKeys :: [PrimaryKey]
-  , dbProcs       :: ProcsMap
-  , pgVersion     :: PgVersion
+  { dbTables        :: [Table]
+  , dbColumns       :: [Column]
+  , dbRelationships :: [Relation]
```
wolfgangwalther, line 68: What's the reason for that change? If it was my suggestion to not use abbreviations for field names: "Relation" is not an abbreviation. It's a full word. And our type is named Relation, too, so I'd leave it at that!

test/Feature/EmbedDisambiguationSpec.hs
```diff
@@ -431,3 +431,9 @@ spec =
           [json|[{"id":1,"users":[{"id":1},{"id":3}]},{"id":2,"users":[{"id":1}]},{"id":3,"users":[{"id":1}]},{"id":4,"users":[{"id":1}]},{"id":5,"users":[{"id":2},{"id":3}]},{"id":6,"users":[{"id":2}]},{"id":7,"users":[{"id":2}]},{"id":8,"users":[]}]|]
           { matchHeaders = [matchContentTypeJson] }
 
+    context "m2m embed when the junction table is in an internal schema" $ do
+      -- https://github.com/PostgREST/postgrest/issues/1736
+      it "works by disregarding the internal junction" $
```
wolfgangwalther, line 436: I think we need another test-case for an internal junction with a different name (that has no view with the same name) - this should not be embeddable at all.

src/PostgREST/DbStructure.hs
```diff
@@ -101,15 +101,33 @@ getDbStructure schemas extraSearchPath pgVer prepared = do
...
-    , dbRelations = rels
+    , dbRelationships = rels
     , dbPrimaryKeys = keys'
     , dbProcs = procs
     , pgVersion = pgVer
     }
 
+-- | Remove db objects that belong to an internal schema(not exposed through the API) from the DbStructure.
+removeInternal :: [Schema] -> DbStructure -> DbStructure
+removeInternal schemas dbStruct =
```
wolfgangwalther, line 115: Hm... I'm not a big fan of filtering this in haskell. But don't see a way to do this on the SQL side already - short of refactoring a lot of the queries in here. Reminds of the the megaquery approach!

src/PostgREST/DbStructure.hs
steve-chavez: Ah, the thing is that Relation means a table or view under pg(also under Relational algebra, I believe) and I thought the term was confusing. The "Relationship" name is more fit in this case. This is currently under a comment:

```haskell
-- | "Relation"ship between two tables.
--
data Relation = Relation
```

The `dbRelationships` would actually show on the resulting json schema cache, so that's why I only changed that. But you're right, I can leave this unrelated change out of this PR.

Besides that, WDYT about renaming Relation to Relationship?

monacoremo: > Besides that, WDYT about renaming Relation to Relationship?

Agree to not use Relation, it's confusing in the relational db context!

src/PostgREST/DbStructure.hs
steve-chavez: Yeah, IIRC getting the relationships with one big query is possible. But it's a big change, not only the queries need to be changed but also the types/functions. This is definitely the simplest way(not that bad I think, just one function).

src/PostgREST/DbStructure.hs
wolfgangwalther: Absolutely agree! I was just confused by the inconsistency. If we change it, which we should, we should change it everywhere - especially on the type itself.

steve-chavez: Cool, done!

src/PostgREST/Error.hs
```diff
@@ -55,18 +56,18 @@ data ApiRequestError
   | InvalidBody ByteString
   | ParseRequestError Text Text
   | NoRelBetween Text Text
-  | AmbiguousRelBetween Text Text [Relation]
+  | AmbiguousRelBetween Text Text [Relationship]
   | InvalidFilters
   | UnacceptableSchema [Text]
-  | UnknownRelation                -- Unreachable?
+  | UnknownRelationship                -- Unreachable?
```
wolfgangwalther, line 62: `grep` tells me we're really not using that at all - let's remove it right away, while it shows up here in the diff? :)

wolfgangwalther:
> After this change, I expected to be able to delete this line on `findRel`:
> 
> https://github.com/PostgREST/postgrest/blob/7ca0d469367d480804b2b8b91bdceb653ada754a/src/PostgREST/Request/DbRequestBuilder.hs#L177-L178

I don't think you can remove this line, because we do not have a separate schema cache for different exposed schemas. So using the `Accept-Profile` stuff - wouldn't removing this line lead to possibly embedding tables of **other** exposed schemas?

src/PostgREST/Error.hs
steve-chavez: Done!

steve-chavez:
>  schema == tableSchema relTable && schema == tableSchema relForeignTable 
>  So using the Accept-Profile stuff - wouldn't removing this line lead to possibly embedding tables of other exposed schemas?

True. I'll refrain from doing this, I mostly wanted to try and simplify the `findRel` function. But clearly this is not the way.

test/Feature/EmbedDisambiguationSpec.hs
steve-chavez: Here you refer to this case you put up [here](https://github.com/PostgREST/postgrest/issues/1587#issuecomment-734995669) right?

> So just create two separate tables, both with the same fks and the same name, but obviously only one of them in the exposed schema. The error will be the same "could not find any relationship". Drop one of the tables and it works again.

wolfgangwalther: No, but that's a very good one to add, too!

The current test is just testing whether the view in the exposed schema can be embedded without an ambiguous error. You solved the bug, that this was not the case by removing the private junction completely. But you could have also solved it by giving "priority" to the public junction - and still allowing the private junction to be embedded somehow (the easiest case would be "only when no public junction with the same name exists"). I think this was a (bad) suggestion I made in the discussion in that issue.

To be 100% we're not breaking this again, you need to add a test:
```
TABLE public.A
TABLE public.B
JUNCTION private.C (A, B)
```
without any direct or indirect (through a view) exposure of `C`. This should fail with "No Relationship found".

steve-chavez: Now it's clear as day. Done!

steve-chavez:
Added test cases for both https://github.com/PostgREST/postgrest/issues/1587 (main issue, which was the same as #1736) and https://github.com/PostgREST/postgrest/issues/1587#issuecomment-734995669 (different one that's also solved by filtering internal junctions).

wolfgangwalther (approved):
I tested `dump-schema` on main vs. this PR: The dump is now ~15% smaller.

PR #1827, opened 2021-04-23
Some refactoring and fixes in the nix development environment

wolfgangwalther:
There are a couple of things I needed to do for #1805, #1812 and while playing around with CI in #1822 - some of which I needed to have fixed in multiple of those PRs. To avoid juggling the same commits in the different PRs, I want to merge them ahead of time.

This PR does the following:
- Moves `default.nix` and `shell.nix` into the `nix/` subfolder. This is to have all nix-related files in one directory, which allows to set up the context for a ci-nix-Dockerfile much easier. This adds a convenience wrapper for `shell.nix` in the root directory, so that `nix-shell` can still be called the same way as before.
- Splits `cabalTools` from `devTools` to allow installing `postgrest-build` in CI without the overhead of many `devtools` dependencies.
- Fix `postgrest-dump-schema` currently not working `nix-shell --pure`. `yq` needs `jq` under the hood - but it seems like it currently doesn't import it properly in it's environment. Not sure if that's fixed already upstream, but just adding `jq` to the PATH fixes it for now.
- Fix `postgrest-coverage` needing a rebuild while running the tests, because of changed flags. The build step had `--enable-tests all`, but the test step did not and so started a rebuild. Now the build targets are listed explicitly and no rebuild is needed.
- A small improvement to the way we create our `withTmpDir` temporary directories. They are now created in a common directory `/tmp/postgrest/...` - this allows to upload all of the kept temp dirs as artifacts in CI in the case of a failed step. Also each individual temp directory is now prefixed with the name of the shell script that was run and not a generic `tmp.xxxx`. Makes it much easier to debug things.
- Makes `with_tmp_db` a true `checkedShellScript` as discussed in https://github.com/PostgREST/postgrest/pull/1812#issuecomment-822023148. This also helps with the Dockerfile context (see above). We still need to agree on how we handle the stack situation. The 3 suggestions we have so far are:
a) Auto-generate a checked-in copy of with_tmp_db
b) just keep the original file or
c) create `postgrest-stack-build` and `postgrest-stack-test` commands and keep everything in nix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,8 @@ jobs:
           name: run spec tests
           command: |
             test/create_test_db "postgres://circleci@localhost" postgrest_test stack test
+      - store_artifacts:
+          path: /tmp/postgrest
 
   # Publish a new release. This only runs when a release is tagged (see
   # workflow below).
@@ -111,6 +113,8 @@ jobs:
             then
               postgrest-release-dockerhub-description
             fi
+      - store_artifacts:
+          path: /tmp/postgrest
 
   # Build everything in default.nix and push to the Cachix binary cache if running on main
   nix-build:
@@ -154,6 +158,8 @@ jobs:
               echo "Building all derivations (caching skipped for outside pull requests)..."
               nix-build
             fi
+      - store_artifacts:
+          path: /tmp/postgrest
 
   # Run tests
   nix-test:
@@ -218,7 +224,8 @@ jobs:
           name: Run memory tests
           command: postgrest-test-memory
           when: always
-
+      - store_artifacts:
+          path: /tmp/postgrest
 
 workflows:
   version: 2
--- a/default.nix
+++ b/default.nix
@@ -113,9 +113,12 @@ rec {
 
   ### Tools
 
+  cabalTools =
+    pkgs.callPackage nix/tools/cabalTools.nix { inherit devCabalOptions postgrest; };
+
   # Development tools.
-  devtools =
-    pkgs.callPackage nix/tools/devtools.nix { inherit tests style devCabalOptions hsie; };
+  devTools =
+    pkgs.callPackage nix/tools/devTools.nix { inherit tests style devCabalOptions hsie; };
 
   # Docker images and loading script.
   docker =
@@ -127,7 +130,7 @@ rec {
 
   # Utility for updating the pinned version of Nixpkgs.
   nixpkgsTools =
-    pkgs.callPackage nix/tools/nixpkgstools.nix { };
+    pkgs.callPackage nix/tools/nixpkgsTools.nix { };
 
   # Scripts for publishing new releases.
   release =
@@ -149,5 +152,5 @@ rec {
     };
 
   withTools =
-    pkgs.callPackage nix/tools/withtools.nix { inherit postgresqlVersions; };
+    pkgs.callPackage nix/tools/withTools.nix { inherit postgresqlVersions; };
 }
--- a/default.nix
+++ b/default.nix
@@ -102,19 +102,43 @@ rec {
       )
     );
 
+  env =
+    postgrest.env;
+
+  # Tooling for analyzing Haskell imports and exports.
+  hsie =
+    pkgs.callPackage nix/hsie {
+      ghcWithPackages = pkgs.haskell.packages."${compiler}".ghcWithPackages;
+    };
+
+  ### Tools
+
+  # Development tools.
+  devtools =
+    pkgs.callPackage nix/tools/devtools.nix { inherit tests style devCabalOptions hsie; };
+
   # Docker images and loading script.
   docker =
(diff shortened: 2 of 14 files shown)


monacoremo:
> Moves default.nix and shell.nix into the nix/ subfolder. This is to have all nix-related files in one directory, which allows to set up the context for a ci-nix-Dockerfile much easier.

This breaks my [workflow](https://xkcd.com/1172/), or at least make it less convenient and a bit suprising to any nix developer. Having the `default.nix` at the root allows us to run `nix-build`, `nix-build -A postgrestStatic`, `nix run ...`, `nix-instantiate` (postgrest-push-cachix is broken in this branch right  now) in the root of the project without needing to specify a file or directory. I'm very against changing this and giving up the expected behaviour, unless we have an exceedingly strong reason for it. I imagine that it would be possible to create the nix-ci-dockerfile  as-is also, maybe with 1-2 extra lines? Can you elaborate on why it is needed?

monacoremo:
> The 3 suggestions we have so far are:
a) Auto-generate a checked-in copy of with_tmp_db
b) just keep the original file or
c) create postgrest-stack-build and postgrest-stack-test commands and keep everything in nix.

Suggest that we go with b) in this PR.

wolfgangwalther:
> This breaks my [workflow](https://xkcd.com/1172/), or at least makes it less convenient and a bit suprising to any nix developer. Having the `default.nix` at the root allows us to run `nix-build`, `nix-build -A postgrestStatic`, `nix run ...`, `nix-instantiate` (postgrest-push-cachix is broken in this branch right now) in the root of the project without needing to specify a file or directory. I'm very against changing this and giving up the expected behaviour, unless we have an exceedingly strong reason for it. I imagine that it would be possible to create the nix-ci-dockerfile as-is also, maybe with 1-2 extra lines? Can you elaborate on why it is needed?

Love the xkcd :D. You are right, it is not technically needed anymore. I actually wanted to add a question "Remo, I am sure there are problems with this change, what am I missing?" to that part. The background was, that I started with the `nix/Dockerfile` as-is. To be able to run `nix-env -f default.nix -iA ...` during the build - all the nix files had to be in the docker file context. I started with just `docker build .` and those files were out of context. Only much later I realized, that I needed to copy `postgrest.cabal`, too (to make it install build dependencies with `cabal2nix`) and would easily be able to change the context directory for the docker command. In my latest setup, I have a checkedShellScript (not pushed anywhere, yet) to create that Dockerfile and the `nix/Dockerfile` is gone. So... it's not at all inconvenient anymore.

Happy to remove that change :).

wolfgangwalther:
> > This breaks my [workflow](https://xkcd.com/1172/), or at least makes it less convenient and a bit suprising to any nix developer. Having the `default.nix` at the root allows us to run `nix-build`, `nix-build -A postgrestStatic`, `nix run ...`, `nix-instantiate` (postgrest-push-cachix is broken in this branch right now) in the root of the project without needing to specify a file or directory. I'm very against changing this and giving up the expected behaviour, unless we have an exceedingly strong reason for it. I imagine that it would be possible to create the nix-ci-dockerfile as-is also, maybe with 1-2 extra lines? Can you elaborate on why it is needed?
> 
> Love the xkcd :D. You are right, it is not technically needed anymore. I actually wanted to add a question "Remo, I am sure there are problems with this change, what am I missing?" to that part. The background was, that I started with the `nix/Dockerfile` as-is. To be able to run `nix-env -f default.nix -iA ...` during the build - all the nix files had to be in the docker file context. I started with just `docker build .` and those files were out of context. Only much later I realized, that I needed to copy `postgrest.cabal`, too (to make it install build dependencies with `cabal2nix`) and would easily be able to change the context directory for the docker command. In my latest setup, I have a checkedShellScript (not pushed anywhere, yet) to create that Dockerfile and the `nix/Dockerfile` is gone. So... it's not at all inconvenient anymore.
> 
> Happy to remove that change :).

Ah, one more idea:
Not sure if that works, but it should: What about creating just another convenience wrapper similar to `shell.nix` in the root directory for `default.nix`? That should still allow the same workflow, but would have all nix code nicely packaged in the `nix/` directory. WDYT? What's the up or down side for this?

monacoremo:
> Not sure if that works, but it should: What about creating just another convenience wrapper similar to shell.nix in the root directory for default.nix? That should still allow the same workflow, but would have all nix code nicely packaged in the nix/ directory. WDYT? What's the up or down side for this?

Yes, a file with just `import ./nix' in it should work. I see the value of having everything nix related in the nix directory, but this would also add two extra files, indirection and a bit of extra complexity, I'm not sure it's worth it. How much would it simplify the ci docker?

wolfgangwalther:
Kept `test/with_tmp_db` and kept `default.nix` and `shell.nix` in the root directory. Are all other changes good to merge for you, @monacoremo?

monacoremo (approved):

monacoremo:
Merge time!

PR #1829, opened 2021-04-24
Add stack support to nix environment

wolfgangwalther:
https://github.com/PostgREST/postgrest/pull/1812#issuecomment-822247012 got me thinking... and primarily out of curiosity I wanted to try to make our nix dev tools "build tool agnostic" by:
- making all tests and devtools use `cabalTools` instead of `cabal-install` directly
- adding `stackTools` similar to `cabalTools`
- make all tests use `buildTools`, which are either `cabalTools` or `stackTools` depending on command line argument `stack` (boolean, default false)

Some things that I am not sure about:
- The dynamic derivation `postgrest` is built with `cabal2nix` - I am using it right now through `postgrest.env` in `stackTools`, too. I don't really have an idea what `postgrest.env` contains. I guess I only need GHC.
- I understand that `postgrest` is about creating the haskell dependencies via nix packages - and loading them via cachix, right? I guess this doesn't work with `cabal2nix` and stack. I found https://input-output-hk.github.io/haskell.nix/, which seems to support something like `stack2nix`, but... I don't really have a clue.

Unfortunately I couldn't test locally, yet, because I get this when building with stack:
```
cassava > error: #error **INVARIANT BROKEN** Detected invalid combination of `text-short` and
`bytestring` versions. Please verify the `pre-bytestring-0.10-4` flag-logic in the .cabal file wasn't elided.
```
And I have no idea what to do :).

I have a small hope that dependencies are already built in CI and cached, and `postgrest-build` might go through.

Happy about any feedback regarding the idea in general, and the specifics, too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,19 +24,10 @@ jobs:
 
   # Run tests based on stack
   stack-test:
-    machine: true
+    docker:
+      - image: technowledgy/nix:latest
     steps:
       - checkout
-      - run:
-          name: Install Nix
-          command: |
-            curl -L https://nixos.org/nix/install | sh
-            echo "source $HOME/.nix-profile/etc/profile.d/nix.sh" >> $BASH_ENV
-      - run:
-          name: Install and use the Cachix binary cache
-          command: |
-            nix-env -iA cachix -f https://cachix.org/api/v1/install
-            cachix use postgrest
       - run:
           name: Install testing scripts
           command: nix-env -f default.nix --arg stack true -iA buildTools tests withTools
@@ -55,8 +46,37 @@ jobs:
           name: Build src and tests
           command: postgrest-build -j1
       - run:
-          name: Run spec tests
-          command: postgrest-test-spec
+          name: Skip tests on build or primary test failure
+          command: circleci-agent step halt
+          when: on_fail
+      - run:
+          name: Run coverage (io tests and spec tests against PostgreSQL 13)
+          command: postgrest-coverage
+          when: always
+      - run:
+          name: Run the spec tests against PostgreSQL 12
+          command: postgrest-with-postgresql-12 postgrest-test-spec
+          when: always
+      - run:
+          name: Run the spec tests against PostgreSQL 11
+          command: postgrest-with-postgresql-11 postgrest-test-spec
+          when: always
+      - run:
+          name: Run the spec tests against PostgreSQL 10
+          command: postgrest-with-postgresql-10 postgrest-test-spec
+          when: always
+      - run:
+          name: Run the spec tests against PostgreSQL 9.6
+          command: postgrest-with-postgresql-9.6 postgrest-test-spec
+          when: always
+      - run:
+          name: Run the spec tests against PostgreSQL 9.5
+          command: postgrest-with-postgresql-9.5 postgrest-test-spec
+          when: always
+      - run:
+          name: Check the spec tests for idempotence
+          command: postgrest-test-spec-idempotence
+          when: always
       - store_artifacts:
           path: /tmp/postgrest
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,51 +22,41 @@ jobs:
             # 'Note: For checking this locally, use `nix-shell --run postgrest-style`
             postgrest-style-check
 
-  # Run tests based on stack and docker against the oldest PostgreSQL version
-  # that we support.
+  # Run tests based on stack
   stack-test:
-    docker:
-      - image: cimg/base:2021.03
-        environment:
-          - PGHOST=localhost
-      - image: circleci/postgres:9.5
-        environment:
-          - POSTGRES_USER=circleci
-          - POSTGRES_DB=circleci
-          - POSTGRES_HOST_AUTH_METHOD=trust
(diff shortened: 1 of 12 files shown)


monacoremo:
> I don't really have an idea what postgrest.env contains. I guess I only need GHC.

It contains a cabal-install with GHC and builds of all postgrest dependencies. It comes with absolutely everything that is needed to build the project, even in a sandbox.

> I found https://input-output-hk.github.io/haskell.nix/, which seems to support something like stack2nix, but... I don't really have a clue.

Haskell.nix is a very different approach to packaging and building Haskell projects. It has some advantages, like making all components of a project (e
g. our library, executable, tests) individually accessible and buildable. I couldn't get the static build (which was our gateway drug into the Nix ecosystem) to work with it, so I went with the standard nixpkgs Haskell infrastructure - which is also pretty good!

Cabal2nix parses our .cabal file, figures out what the corresponding Nix packages are for the dependencies and generates a Nix derivation to build our project. It uses the versions of our dependencies that are available in our nixpkgs pin, which corresponds, with some modifications, to a recent stackage snapshot (or at least it did, there are efforts I think to decouple from stackage).

As nixpkgs provides us with a consistent set of Haskell packages, stack (and stack2nix) has no additional value over cabal for us.

The only purpose of our stack test is to make sure that the project can still be built and tested without Nix. That's releva t for our builds for any platform that does not support Nix (Windows, the BSDs etc.) and any potential contributors that only use stack (of which there are likely more than Nix users).

Given that, I don't think there will be value in integrating stack into our Nix setup. It's clearer that it still works independently if we keep it separate.

wolfgangwalther:
> > I don't really have an idea what postgrest.env contains. I guess I only need GHC.
> 
> It contains a cabal-install with GHC and builds of all postgrest dependencies. It comes with absolutely everything that is needed to build the project, even in a sandbox.
> 
> > I found https://input-output-hk.github.io/haskell.nix/, which seems to support something like stack2nix, but... I don't really have a clue.
> 
> Haskell.nix is a very different approach to packaging and building Haskell projects. It has some advantages, like making all components of a project (e
> g. our library, executable, tests) individually accessible and buildable. I couldn't get the static build (which was our gateway drug into the Nix ecosystem) to work with it, so I went with the standard nixpkgs Haskell infrastructure - which is also pretty good!
> 
> Cabal2nix parses our .cabal file, figures out what the corresponding Nix packages are for the dependencies and generates a Nix derivation to build our project. It uses the versions of our dependencies that are available in our nixpkgs pin, which corresponds, with some modifications, to a recent stackage snapshot (or at least it did, there are efforts I think to decouple from stackage).
> 
> As nixpkgs provides us with a consistent set of Haskell packages, stack (and stack2nix) has no additional value over cabal for us.

Thanks, I understand a bit more now. I absolutely agree that stack itself does not provide any more value compared to cabal-in-nix on linux.

Turning this around, however, means that I don't need stack2nix or something - because stack itself takes care of dependencies "properly". I found `stack --nix`, which solved all of the cabal / stack / `postgrest.env` confusion for me. So, currently, the `stackTools` are independent of `postgrest.env`. The build environment is created by stack - but that's fine for me. I just want the tooling to be inside nix, not the build itself.

> The only purpose of our stack test is to make sure that the project can still be built and tested without Nix. That's releva t for our builds for any platform that does not support Nix (Windows, the BSDs etc.) [...]

Yes, I agree...

> [...] and any potential contributors that only use stack (of which there are likely more than Nix users).

... but I don't understand this. In some way stack and nix solve the same problem when looking at dependencies. But nix is much more than that. So I don't see the "either stack or nix user" thing. You can be both. You can be none (using cabal without nix).

I think there would be much value when our dev tools were to support stack, too - for exactly those potential contributors that like to work with stack. And also for us - how do we debug a stack issue right now? Only in CI? I don't have stack installed - and it would make things much easier, if I had a reproducible setup for stack in our devtools.

One example: For the error I am getting right now - I don't know whether my stack setup is at fault or whether we have a stack dependency problem. But to be able to debug a stack issue, I'd need to have a known-to-be-working stack build setup.

Plus, it would be really cool if we could run the same CI pipeline we currently run with `nix-test` for cabal with stack. That would not only prove that building with stack is possible, but also how we support stack first class in our dev tools.

wolfgangwalther:
This works quite well locally now. Just run `nix-shell --arg stack true` and everything you do will be with stack. Tests, etc. all run just the same as with cabal.

wolfgangwalther:
And it works in CI, too. The failing test is because of the docker image I used as executor, same as in #1822. I have not been able to find the reason why the IO tests fail with this docker image, yet.

monacoremo:
There are some great ideas in these commits, but I don’t think integrating Stack into our Nix setup makes sense or is worth it. 

We need to build with Stack on all the platforms that Nix does not support (Windows, FreeBSD) and CI continuously checks that this works. Installing stack for testing is super simple (with Nix: nix-shell -p stack) and it brings its own GHC etc. @wolfgangwalther, do you agree that we close this?

wolfgangwalther:
> @wolfgangwalther, do you agree that we close this?

Yes, in principle, I'd like to take two of the commits, that are unrelated to the stack stuff, out of this first, though. I'll close when I get to it.

PR #1830, opened 2021-04-25
refactor: Move accept content type code from App into ApiRequest

monacoremo:
This is a smaller intermediate step to carving out a small API to `Request`. This moves the code that finds an acceptable response content type into ApiRequest.

### DIFF
--- a/src/PostgREST/Request/ApiRequest.hs
+++ b/src/PostgREST/Request/ApiRequest.hs
@@ -138,7 +138,6 @@ data ApiRequest = ApiRequest {
   , iRange                :: M.HashMap ByteString NonnegRange -- ^ Requested range of rows within response
   , iTopLevelRange        :: NonnegRange                      -- ^ Requested range of rows from the top level
   , iTarget               :: Target                           -- ^ The target, be it calling a proc or accessing a table
-  , iAccepts              :: [ContentType]                    -- ^ Content types the client will accept, [CTAny] if no Accept header
   , iPayload              :: Maybe PayloadJSON                -- ^ Data sent by client and used for mutation actions
   , iPreferRepresentation :: PreferRepresentation             -- ^ If client wants created items echoed back
   , iPreferParameters     :: Maybe PreferParameters           -- ^ How to pass parameters to a stored procedure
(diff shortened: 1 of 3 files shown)


src/PostgREST/Request/ApiRequest.hs
```diff
@@ -158,22 +159,25 @@ data ApiRequest = ApiRequest {
...
   , iProfile              :: Maybe Schema                     -- ^ The request profile for enabling use of multiple schemas. Follows the spec in hhttps://www.w3.org/TR/dx-prof-conneg/ttps://www.w3.org/TR/dx-prof-conneg/.
   , iSchema               :: Schema                           -- ^ The request schema. Can vary depending on iProfile.
+  , iAcceptContentType    :: ContentType
   }
 
 -- | Examines HTTP request and translates it into user intent.
-userApiRequest :: NonEmpty Schema -> Maybe Text -> DbStructure -> Request -> RequestBody -> Either ApiRequestError ApiRequest
-userApiRequest confSchemas rootSpec dbStructure req reqBody
+userApiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> Either ApiRequestError ApiRequest
+userApiRequest conf dbStructure req reqBody
```
wolfgangwalther, line 167:
```suggestion
userApiRequest AppConfig{..} dbStructure req reqBody
```
The where list is long enough and `confSchemas` is used immediately on the next line - so using `configDbSchemas` is easier to follow.

wolfgangwalther (approved):

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -158,22 +159,25 @@ data ApiRequest = ApiRequest {
...
-  | otherwise = Right ApiRequest {
+  | otherwise = do
+     acceptContentType <- findAcceptContentType conf action target accepts
+     return ApiRequest {
       iAction = action
       , iTarget = target
       , iRange = ranges
       , iTopLevelRange = topLevelRange
-      , iAccepts = maybe [CTAny] (map ContentType.decodeContentType . parseHttpAccept) $ lookupHeader "accept"
+      , iAccepts = accepts
```
wolfgangwalther, line 180: I think this field is not needed at all anymore, because `accepts` is used right in here.

wolfgangwalther:
Cirrus is failing with:
```
src/PostgREST/App.hs:163:34: error: [-Wunused-record-wildcards, -Werror=unused-record-wildcards]
    No variables bound in the record wildcard match are used
      Possible fix: omit the ‘..’
    |
163 | postgrestResponse conf@AppConfig{..} maybeDbStructure pool time req = do
```

I wonder why none of our other builds fail here? I would expect that this should fail everywhere.

src/PostgREST/Request/ApiRequest.hs
monacoremo: True, was not needed anymore!

monacoremo:
> I wonder why none of our other builds fail here? I would expect that this should fail everywhere.

```
Build profile: -w ghc-8.10.4 -O1
```

Apparently it builds with GHC 8.10 that added this warning. Wouldn't that require a different stack resolver though?..

Issue #1804, opened 2021-04-11
[dev] Refactoring roadmap


monacoremo:
With #1793, we've started to process of streamlining the codebase. This is a purely internal concern that will yield no additional features. It will, however:

* ease maintenance
* help with the onboarding of new contributors
* enable quicker implementation of new features
* further increase confidence on correctness 

Next steps include (in no particular order and to be completed)

* [ ] Encapsulate the `Request` modules behind one API
* [ ] Encapsulate the `Query` modules behind on API
* [ ] Carve-out `Response` concerns from `App`, i.a.
* [ ] Stop using `GHC.Show` instances for string conversion
* [ ] Stop using partial functions (ones from `Protolude.Conv`, `unsafeHead` etc.)
* [ ] Improve string handling in Query generation for additional safety and performance
* [ ] ...

monacoremo:
@steve-chavez decision needed!

This works in principle, with some positives and negatives. 

Positives:

* It works. We have a `postgrest-test-doctest` command and it's integrated in CI
* Doctests are quite easy to write and make for great in-module documentation  (see example in the Preferences module)
* Using doctests drives us to create clean, usable module APIs (as would unit tests, but not the integration tests we have)
* Added complexity is quite low - essentially a section in `postgrest.cabal` and a `Main.hs` file. There is no impact on other tests
* We get extra safety through the additional tests

Negatives:

* We won't get coverage out of the doctests, probably ever - this is an inherent limitation of the GHCi based implementation and the fact that GHCi does not support HPC. At least, we will not worsen coverage
* Running the doctests outside nix-shell right now is flaky and I'm not quite sure why. This should be fixable later, in the meantime it takes about 30s extra time in CI

Should we still merge this? I think the benefits outweigh the negatives and recommend yes.

steve-chavez:
`unsafeHead` went away on https://github.com/PostgREST/postgrest/pull/2254

steve-chavez:
> Request: purely parse request, at most taking config into account. `parseRequest :: Wai.Request -> Either RequestError Request

> Plan: Use db structure to validate and plan the request, including embeds. plan :: DbStructure -> Request -> Either PlanError QueryPlan

To do the above, ApiRequest should remove its dependency on DbStructure - which is needed by the variadic logic introduced on https://github.com/PostgREST/postgrest/pull/1603.  

[...]

PR #1831, opened 2021-04-25
refactor: Split Request, Query and Response interfaces from App

monacoremo:
Work in progress: Create a small API for all the request parsing we do - this will simplify `App`, enable more unit testing and allow us to simplify/refactor `Request` and it's submodules more easily.


monacoremo:
@wolfgangwalther it's becoming obvious that https://github.com/PostgREST/postgrest/pull/1814#issuecomment-826120200 was a lie... 

This splits up pieces from `App` into `Request`, `Query` and `Response`, as we discussed back in #1793. There will still be many unwanted interlinks between these modules, but it's a first step. The split of `Query` follows your recommendation, much better than the earlier attempt. WDYT?

wolfgangwalther:
> @wolfgangwalther it's becoming obvious that [#1814 (comment)](https://github.com/PostgREST/postgrest/pull/1814#issuecomment-826120200) was a lie...

All good. I was a bit bored by #1830 already.

> This splits up pieces from `App` into `Request`, `Query` and `Response`, as we discussed back in #1793. There will still be many unwanted interlinks between these modules, but it's a first step. The split of `Query` follows your recommendation, much better than the earlier attempt. WDYT?

Hm. I had hoped we could get to a point where all IO handling was in `App` - and `Request`, `Query` and `Response` were all pure. The execution of the query would then be in `App` - while `Query` would construct and return it. Right now, it's a bit too much of `App` that is moved into `Query`, if I understood correctly.

Do you think that's possible / would make things easier? My hope is that unit testing of query generation would be easier to do.

---

Also what strikes me as odd is the fact, that `App` pattern matches for each different request type - and then we have query and response handlers for each request type individually. I guess this is mostly a "we're not there, yet" - do you think it will be possible to have the IO part in `App` made generic? I.e. pass `Request` into `createQuery` (or whatever the name would be) - then pattern match there. Then pass `Request` and `QueryResult` in `createResponse` (or whatever ...) etc.

---

I really like those interfaces:
```haskell
updateResponse :: ApiRequest -> WriteQueryResult -> Wai.Response

singleUpsertResponse :: ApiRequest -> WriteQueryResult -> Wai.Response

deleteResponse :: ApiRequest -> WriteQueryResult -> Wai.Response
```

Those are still a bit heavy:
```haskell
readResponse :: Bool -> QualifiedIdentifier -> AppConfig -> ApiRequest -> Statements.ResultsWithCount -> Maybe Int64 -> [GucHeader] -> Maybe HTTP.Status -> Wai.Response

createResponse :: QualifiedIdentifier -> DbStructure -> ApiRequest -> WriteQueryResult -> Wai.Response

invokeResponse :: InvokeMethod -> ApiRequest -> Statements.ProcResults -> [GucHeader] -> Maybe HTTP.Status -> Wai.Response
```

This looks like we need to either extend `ApiRequest` with a bit more things - or wrap it in a `Request` together with other stuff.

src/PostgREST/Response.hs
```diff
@@ -0,0 +1,195 @@
...
+                                          PreferRepresentation (..))
+
+import qualified PostgREST.ContentType as ContentType
+
+import Protolude      hiding (Handler, toS)
+import Protolude.Conv (toS)
+
+
+readResponse :: Bool -> QualifiedIdentifier -> AppConfig  -> ApiRequest -> Statements.ResultsWithCount -> Maybe Int64 -> [GucHeader] -> Maybe HTTP.Status -> Wai.Response
+readResponse headersOnly identifier AppConfig{..} apiRequest@ApiRequest{..} (_, queryTotal, _ , body, _, _) total gucHeaders gucStatus =
```
wolfgangwalther, line 50:
```suggestion
readResponse :: Bool -> QualifiedIdentifier -> ApiRequest -> Statements.ResultsWithCount -> Maybe Int64 -> [GucHeader] -> Maybe HTTP.Status -> Wai.Response
readResponse headersOnly identifier apiRequest@ApiRequest{..} (_, queryTotal, _ , body, _, _) total gucHeaders gucStatus =
```

At first glance this is not used. Correct?

Either I missed something entirely.. or GHC is not picking up the unused wildcard matching here. Didn't we have one of the CI's just complain exactly about such a thing?

monacoremo:
> Also what strikes me as odd is the fact, that App pattern matches for each different request type - and then we have query and response handlers for each request type individually.

Thought about this also, and how it would be nice to chain one set of three functions for all cases. I think this way seems a bit less elegant, but makes more impossible states impossible: It's obvious that no `update` query will ever be rendered into a `singleUpsert` response just by looking at `App`. Let's see, but it might be the stable outcome.

> This looks like we need to either extend ApiRequest with a bit more things - or wrap it in a Request together with other stuff

Yes, definitely - the signatures of both the `Query` and `Response` functions are a mess. I'm trying to see what the overall pattern is, it's not the `RequestContext` we previously had. I might completely wrap ApiRequest and only expose Request to get there...

wolfgangwalther:
> Hm. I had hoped we could get to a point where all IO handling was in `App` - and `Request`, `Query` and `Response` were all pure. The execution of the query would then be in `App` - while `Query` would construct and return it. Right now, it's a bit too much of `App` that is moved into `Query`, if I understood correctly.
> 
> Do you think that's possible / would make things easier? My hope is that unit testing of query generation would be easier to do.

Hm. Everything from `handleRequest` on is already pure, right? I'm still confused by Monads in general... and `runDbHandler` was hidden in the diff. Aha!

monacoremo:
> Hm. I had hoped we could get to a point where all IO handling was in App - and Request, Query and Response were all pure

We are pretty close actually!

* Request returns Either
* Query returns a DbHandler (more below)
* Response returns a pure Wai.Response

A DbHandler is a Hasql.Transaction whose execution (and theoretically creation, need to check whether we have such cases left) can fail with an Error.

We should be able to split this out further within Query, but as a first step this makes sense I think.

src/PostgREST/Query.hs
```diff
@@ -0,0 +1,234 @@
...
+           QueryBuilder.limitedQuery countQuery ((+ 1) <$> configDbMaxRows)
+         else
+           countQuery
+        )
+        (iAcceptContentType == CTSingularJSON)
+        (shouldCount iPreferCount)
+        (iAcceptContentType == CTTextCSV)
+        bField
+        (pgVersion dbStructure)
+        configDbPreparedStatements
```
wolfgangwalther, line 76:
```suggestion
    lift . SQL.statement mempty $
      Statements.createReadStatement
        (QueryBuilder.readRequestToQuery req)
        (if iPreferCount == Just EstimatedCount then
           -- LIMIT maxRows + 1 so we can determine below that maxRows was surpassed
           QueryBuilder.limitedQuery countQuery ((+ 1) <$> configDbMaxRows)
         else
           countQuery
        )
        (iAcceptContentType == CTSingularJSON)
        (shouldCount iPreferCount)
        (iAcceptContentType == CTTextCSV)
        bField
        (pgVersion dbStructure)
      configDbPreparedStatements
```

The prepared statements option seems to be an argument to `SQL.statement`, not `Statements.createReadStatement`. Might come up in other places, too?

monacoremo:
> DbHandler seems to be defined twice. Once in App.hs and once in Query.hs. Is that on purpose?

Was not on purpose! Fixed. I think that the whole `runDbHandler` stuff will move to Query later on, but I'm leaving it as is fow now.

monacoremo:
I think we should make a cut here! The dispatch in [`App`](https://github.com/monacoremo/postgrest/blob/a22d914e00a6a849312af44c335298ea7e41cb98/src/PostgREST/App.hs#L138-L147) should be pretty stable, as are the interfaces of `Request`, `Query` and `Response`. Each of those modules and their submodules still have significant refactoring potential. This will be easier to do now that they have clear interfaces.

src/PostgREST/Auth.hs
```diff
@@ -57,6 +64,15 @@ jwtClaims AppConfig{..} payload time = do
     jwtClaimsError JWT.JWTExpired = JwtTokenInvalid "JWT expired"
     jwtClaimsError e              = JwtTokenInvalid $ show e
 
+headerJWT :: Request -> Text
```
monacoremo, line 67: I split this out of `ApiRequest` as it's a separate concern from all the other request parsing. This allows us to check the JWT before we even load the body of the request, which should improve DoS resistance a bit.

src/PostgREST/Middleware.hs
```diff
@@ -51,16 +49,14 @@ import PostgREST.Request.Preferences
...
 
--- | Runs local(transaction scoped) GUCs for every request, plus the pre-request function
-runPgLocals :: AppConfig   -> M.HashMap Text JSON.Value ->
-               (ApiRequest -> ExceptT Error H.Transaction Wai.Response) ->
-               ApiRequest  -> ExceptT Error H.Transaction Wai.Response
-runPgLocals conf claims app req = do
+-- | Set local(transaction scoped) GUCs for every request, plus the pre-request function
+setPgLocals :: AppConfig -> M.HashMap Text JSON.Value -> ApiRequest
+  -> ExceptT Error H.Transaction ()
+setPgLocals conf claims req = do
```
monacoremo, line 55: This is not really middleware, and does not really wrap the application. I renamed it to `setPgLocals` to reflect its function better, will probably move to `App` later on.

src/PostgREST/Query.hs
```diff
@@ -0,0 +1,255 @@
...
+
+type DbHandler = ExceptT Error SQL.Transaction
+
+data ReadQueryResult = ReadQueryResult
+  { rqRequest    :: ReadRequestInfo
+  , rqQueryTotal :: Int64
+  , rqBody       :: BS8.ByteString
+  , rqTableTotal :: Maybe Int64
+  , rqGucHeaders :: [GucHeader]
+  , rqGucStatus  :: Maybe HTTP.Status
```
monacoremo, line 64: Coverage fails as we only use those with `RecordWildCards`

src/PostgREST/App.hs
```diff
@@ -161,438 +113,47 @@ postgrestResponse
...
-  let
-    handleReq apiReq =
-      handleRequest $ RequestContext conf dbStructure apiReq
-
-  runDbHandler pool (txMode apiRequest) jwtClaims .
-    Middleware.optionalRollback conf apiRequest $
-      Middleware.runPgLocals conf jwtClaims handleReq apiRequest
+  jwtClaims <- Auth.jwtClaims conf req time
+  body <- lift $ Wai.strictRequestBody req
+  dbStructure <- maybe (throwError Error.ConnectionLostError) pure maybeDbStructure
```
wolfgangwalther, line 119: Do we need to read the body before getting the dbStructure? I don't think so. Let's swap those 2 lines.

No need to read the body before throwing a `ConnectionLostError`.

Probably makes sense to do:
- `dbStructure`
- `jwtClaims`
- `body`
- `request`

In this order?

src/PostgREST/App.hs
```diff
@@ -161,438 +113,47 @@ postgrestResponse
...
+    UpdateRequest requestInfo ->
+      Response.updateResponse <$> Query.updateQuery requestInfo
+    SingleUpsertRequest requestInfo ->
+      Response.singleUpsertResponse <$> Query.singleUpsertQuery requestInfo
+    DeleteRequest requestInfo ->
+      Response.deleteResponse <$> Query.deleteQuery requestInfo
+    InfoRequest dbStructure _ QualifiedIdentifier{..} ->
+      case find tableMatches $ dbTables dbStructure of
+        Just table -> return $ Response.infoResponse hasPK table
+        Nothing    -> throwError Error.NotFound
```
wolfgangwalther, line 151: In the discussion in #1839 we realized that we need to parse the request (arguments in querystring etc.) to properly handle the `OPTIONS` request.

Does it make sense to change the interface here right now already, so that `requestInfo` is also passed into `Response.infoResponse`?

src/PostgREST/Auth.hs
```diff
@@ -20,33 +20,40 @@ module PostgREST.Auth
...
-import Control.Monad.Except    (liftEither)
-import Data.Either.Combinators (mapLeft)
-import Data.Time.Clock         (UTCTime)
+import Control.Lens              (set)
+import Control.Monad.Except      (liftEither)
+import Data.Either.Combinators   (mapLeft)
+import Data.List                 (lookup)
+import Data.Time.Clock           (UTCTime)
+import Network.HTTP.Types.Header (hAuthorization)
+import Network.Wai               (Request (..))
```
wolfgangwalther, line 32: Uh. This needs to be imported qualified. I am terribly confused by `Request` below, which is **not** our `Request`. `Wai.Request` would make that clear.

src/PostgREST/Auth.hs
```diff
@@ -20,33 +20,40 @@ module PostgREST.Auth
...
-  AppConfig -> LByteString -> UTCTime -> ExceptT Error m JWTClaims
-jwtClaims _ "" _ = return M.empty
-jwtClaims AppConfig{..} payload time = do
-  secret <- liftEither . maybeToRight JwtTokenMissing $ configJWKS
-  eitherClaims <-
-    lift . runExceptT $
-      JWT.verifyClaimsAt validation secret time =<< JWT.decodeCompact payload
-  liftEither . mapLeft jwtClaimsError $ claimsMap configJwtRoleClaimKey <$> eitherClaims
+  AppConfig -> Request -> UTCTime -> ExceptT Error m JWTClaims
+jwtClaims AppConfig{..} request time =
```
wolfgangwalther, line 46: We need some kind of naming convention for `PostgREST.Request`s and `Wai.Request`s. In `App.hs` it was `request` and `req`. Now `request` is the other one.

I wonder whether we should maybe not use `Request` and `Response` at all, given that those clash with wai like that.

What about `RestRequest` and `RestResponse`?

And then consistently use `request` for `Wai.Request` and `restRequest` for `PostgREST.RestRequest`.

src/PostgREST/Auth.hs
wolfgangwalther: Yes, that makes a lot of sense!

src/PostgREST/Query/Statements.hs
```diff
@@ -10,7 +10,9 @@ This module constructs single SQL statements that can be parametrized and prepar
 TODO: Currently, createReadStatement is not using prepared statements. See https://github.com/PostgREST/postgrest/issues/718.
```
wolfgangwalther, line 10: This looks like it is solved, wdyt?

wolfgangwalther (commented):
Reviewed "everything" except `Query.hs`, `Request.hs` and `Response.hs` so far. So... basically nothing, yet. :D

src/PostgREST/Auth.hs
wolfgangwalther: Yes, that makes a lot of sense!

src/PostgREST/Query/Statements.hs
monacoremo: Hm, not sure - it doesn't look like we are explicitly using prepared statements here. @steve-chavez did we change this since that comment?

src/PostgREST/App.hs
monacoremo: Yes, that order makes most sense! Fixed.

src/PostgREST/App.hs
monacoremo: There actually are different `requestInfo` types depending on the request type. This is explicit in `Request` but is not visible in `App`, so I've changed the naming of the `requestInfo` variables to indicate this.

We already pass and ignore (`_`) ApiRequest to the `infoResponse` dispatch, which would be easy to pass through. But it will be best (and easy) to create a custom `requestInfo` for that case then.

src/PostgREST/Auth.hs
monacoremo: Hm, I think it would be cleanest to use `PostgREST.Request`and `PostgREST.Response` if possible in our codebase. The most impacted module is `PostgREST.App`, let me see if I can make it clearer there.

src/PostgREST/Auth.hs
monacoremo: Fixed!

src/PostgREST/Query/Statements.hs
wolfgangwalther: It was my impression that we are using prepared statements everywhere by now.

src/PostgREST/Auth.hs
wolfgangwalther: `request` and `waiRequest` for names then?

src/PostgREST/Query/Statements.hs
steve-chavez: Yeah, this is solved, the comment is a leftover.

https://github.com/PostgREST/postgrest/blob/082c91c85518d4b2ea97b848bba6e2981d962743/src/PostgREST/Query/Statements.hs#L92-L95

[dynamicallyParameterized](https://hackage.haskell.org/package/hasql-dynamic-statements-0.3.1/docs/Hasql-DynamicStatements-Statement.html#v:dynamicallyParameterized) takes a Bool argument for preparing the statement.

Which is set here(below, needs scrolling) with `configDbPreparedStatements`:

https://github.com/PostgREST/postgrest/blob/082c91c85518d4b2ea97b848bba6e2981d962743/src/PostgREST/App.hs#L237-L250

wolfgangwalther (commented):
This second review is mostly about the interfaces between `App`, `Request`, `Query` and `Response`. I haven't really looked into the 3 modules in detail, yet - that's something I will do in a third round, to make sure that no unrelated changes sneaked in.

From what I can tell right now, I really like how things have been split up and where the different parts ended up.

However, I don't like the interface between those modules, too much, yet. Specifically:
- `App` currently needs to know about a few implementation details, namely:
  - The explicit pass-through of `xxxRequestInfo`. This type is purely internal actually and just a workaround to the partiality problem for constructors with different record fields.
  - The mere fact that each "path" (read, create, ...) is separated in each module. Whether some of those paths are handled by the same function internally or not, should be up to the module. It would be better to have one entrypoint per module - just like `Request.parse`.
- On the other side some part of the interface are hidden, because they are passed on through the `xxxRequestInfo` and `xxxQueryResult` types: config, version, structure in both cases, and the request in `QueryResult`. The response function's interface should really be something like `respond :: Context -> Request -> Result -> Wai.Response`. Right now each step kind of takes their input and wraps it in the output. Strange!

I know that this interface resulted from "making impossible states impossible", i.e. stronger typing - at compile time. That's an absolute positive.

But, I think we can get both. Generalize `App`, make the interfaces really nice - and have strong typing. Here's how:

(Note: this includes suggestions in some of the other comments below, might need to read those first - mainly about naming etc.)

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE KindSignatures #-}

data Context = Context
  { ctxConfig      :: AppConfig
  , ctxPgVersion   :: PgVersion
  , ctxDbStructure :: DbStructure
  }

data RequestType
  = Read
  | Create
  | Update
  | Upsert
  | Delete
  | Invoke
  | Info
  | OpenApi

data Request (t :: RequestType) where
  ReadRequest :: ApiRequest -> ReadTree -> Request 'Read
  CreateRequest :: ApiRequest -> ReadTree -> MutateQuery -> Request Create
  UpdateRequest :: ApiRequest -> ReadTree -> MutateQuery -> Request Update
  UpsertRequest :: ApiRequest -> ReadTree -> MutateQuery -> Request Upsert
  DeleteRequest :: ApiRequest -> ReadTree -> MutateQuery -> Request Delete
  InvokeRequest :: ApiRequest -> ReadTree -> Request Invoke
  InfoRequest :: ApiRequest -> Request Info
  OpenApiRequest :: ApiRequest -> Request OpenApi

data Result (t :: RequestType) = Result
  { resBody       :: ByteString
  , resGucHeaders :: [GucHeader]
  , resGucStatus  :: Maybe HTTP.Status
  , resQueryTotal :: Int64
  , resTableTotal :: Maybe Int64
}

-- public interfaces (towards App)
Request.parse :: Context -> Wai.Request -> LBS.ByteString -> Either Error (Request t)
Query.build :: Context -> Request t -> DbHandler (Result t)
Response.send :: Context -> Request t -> DbHandler (Result t) -> Wai.Response

-- examples for internal interfaces
readQuery :: Context -> Request Read -> DbHandler (Result Read)
createQuery :: Context -> Request Create -> DbHandler (Result Create)
-- ...
readResponse :: Context -> Request Read -> DbHandler (Result Read) -> Wai.Response
createResponse :: Context -> Request Create -> DbHandler (Result Create) -> Wai.Response
-- ...
```

This would even be more type-safe than the current interface, where you could pass a `MutateQueryResult` from e.g. `createQuery` into `updateResponse`  or so...

We can still easily see from the public interfaces, that it's not possible to mix different request and response types (`Request t -> Result t`). `App` is completely generic and just puts the different parts together. None of the 3 modules is limited regarding internal refactoring, because they each have 1 entrypoint only.

steve-chavez:
I like the idea of a single entrypoint per module but I'm having trouble with getting GADTs to compile.

Assuming the following reduced example, how would the `parse` function work?

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE KindSignatures #-}

data RequestType
  = Create
  | Update

data Request (t :: RequestType) where
  CreateRequest :: ApiRequest -> ReadRequest -> MutateRequest -> Request Create
  UpdateRequest :: ApiRequest -> ReadRequest -> MutateRequest -> Request Update

-- single entrypoint for Request.hs
parse :: ApiRequest -> ReadRequest -> MutateRequest -> Request t
parse apiReq readReq mutateReq =
  case iAction apiReq of
    ActionMutate MutationUpdate -> UpdateRequest apiReq readReq mutateReq
    _                           -> CreateRequest apiReq readReq mutateReq

```

This produces

```
$ postgrest-build

src/PostgREST/App.hs:102:36: error:
    • Couldn't match type ‘t’ with ‘'Update’
      Expected: Request t
        Actual: Request 'Update
      ‘t’ is a rigid type variable bound by
        the type signature for:
          parse :: forall (t :: RequestType).
                   ApiRequest -> ReadRequest -> MutateRequest -> Request t
        at src/PostgREST/App.hs:99:1-64
    • In the expression: UpdateRequest apiReq readReq mutateReq
      In a case alternative:
          ActionMutate MutationUpdate
            -> UpdateRequest apiReq readReq mutateReq
      In the expression:
        case iAction apiReq of
          ActionMutate MutationUpdate
            -> UpdateRequest apiReq readReq mutateReq
          _ -> CreateRequest apiReq readReq mutateReq
    • Relevant bindings include
        parse :: ApiRequest -> ReadRequest -> MutateRequest -> Request t
          (bound at src/PostgREST/App.hs:100:1)
    |
102 |     ActionMutate MutationUpdate -> UpdateRequest apiReq readReq mutateReq
```

If the type is changed to

```haskell
data Request (t :: RequestType) where
  CreateRequest :: ApiRequest -> ReadRequest -> MutateRequest -> Request t
  UpdateRequest :: ApiRequest -> ReadRequest -> MutateRequest -> Request t
```

It does compile but that would defeat the purpose of GADTs.

src/PostgREST/Request.hs
```diff
@@ -0,0 +1,170 @@
...
+      let
+        fldNames = Types.fstFieldNames readReq
+        fieldName = headMay fldNames
+      in
+      if length fldNames == 1 && fieldName /= Just "*" then
+        return fieldName
+      else
+        Left $ Error.BinaryFieldError iAcceptContentType
+  | otherwise =
+      return Nothing
```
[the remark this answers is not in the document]
steve-chavez: Solved on https://github.com/PostgREST/postgrest/pull/2496

steve-chavez:
@wolfgangwalther Thanks for the explanation, It's clear now!

I've linked your comment on https://github.com/PostgREST/postgrest/issues/1804. I think that's better left for another PR since `ApiRequest` needs other changes too(like https://github.com/PostgREST/postgrest/issues/1804#issuecomment-1267962989).

I'll close this one since `Query`, `Response` and `Plan`(https://github.com/PostgREST/postgrest/pull/2497, which replaces the `Request` module here) have being added in other PRs.

Issue #1826, opened 2021-04-20
OPTIONS doesn't work for RPC


steve-chavez:
This happens currently:

```http
OPTIONS /rpc/func_name

HTTP/1.1 405 Method Not Allowed
{"message":"Bad Request"}
```

Considering the [Allow RFC](https://tools.ietf.org/html/rfc7231#section-7.4.1), we should respond with an `Allow: GET, POST` header depending on the function volatility — fail with 405 for a GET on a VOLATILE function.

This would prevent confusion regarding GET on RPC, see this [so question](https://stackoverflow.com/questions/66582787/get-method-erroneously-working-with-volatile-postgrest-rpc-function).

PR #1839, opened 2021-04-29
fix: Allow OPTIONS requests to RPC

laurenceisla:
This PR handles OPTIONS requests to functions, mentioned here: #1826

Won't add the 405 response to a GET for a volatile function feature due to the routing refactor as discussed [here ](https://github.com/PostgREST/postgrest/pull/1824#discussion_r616857705)

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -28,6 +28,7 @@ import System.Posix.Types       (FileMode)
 
 import qualified Data.ByteString.Char8           as BS8
 import qualified Data.ByteString.Lazy            as LBS
+import qualified Data.HashMap.Strict             as M
 import qualified Data.Set                        as Set
 import qualified Hasql.DynamicStatements.Snippet as SQL
 import qualified Hasql.Pool                      as SQL
@@ -210,7 +211,9 @@ handleRequest context@(RequestContext _ _ ApiRequest{..}) =
     (ActionDelete, TargetIdent identifier) ->
       handleDelete identifier context
     (ActionInfo, TargetIdent identifier) ->
-      handleInfo identifier context
+      handleInfoTable identifier context
+    (ActionInfo, TargetProc proc _) ->
+      handleInfoProc proc context
     (ActionInvoke invMethod, TargetProc proc _) ->
       handleInvoke invMethod proc context
     (ActionInspect headersOnly, TargetDefaultSpec tSchema) ->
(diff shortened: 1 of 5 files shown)


wolfgangwalther:
Right now you're returning `GET` for a `/rpc/some_overloaded_functions` endpoint, when any of the overloaded functions are non-volatile. I don't think that's correct in the HTTP/REST context.

URIs in HTTP/REST should be completely replaceable - i.e. `/rpc/some_overloaded_functions` and `/rpc/some_overloaded_functions?arg=1&arg=2` are completely independent of each other. You should not need to make assumptions about the URI with arguments from the `OPTIONS` request to the URI without arguments.

Right now, an `OPTIONS /rpc/options_test_overloaded_at_least_one_non_volatile` will return `OPTIONS,GET,HEAD,POST` - but that's wrong. You can only call this endpoint with `POST`, when you pass the arguments in the body. Once you want to make a `GET` request, you need to pass arguments in the query string - thereby changing the URI. You could request `GET /rpc/options_test_overloaded_at_least_one_non_volatile?num=1` - but that's something that `OPTIONS /rpc/options_test_overloaded_at_least_one_non_volatile?num=1` needs to return. And this options request should then return `OPTIONS,GET,HEAD` only - no `POST`, because with `POST` you can't pass arguments via query string.

This makes the whole thing a lot more complicated, because you need to take the actual arguments passed via query string into account. The routing-refactor will make this a lot easier, so I suggest we wait for that.

steve-chavez:
So the assumption of the PR is that OPTIONS only identifies the URL-path(`/path`) capabilities. But OPTIONS actually identifies the resource capabilities(the whole URL with query parameters). The [RFC](https://tools.ietf.org/html/rfc7231#section-4.3.7) does mention this.

I guess the assumption comes from seeing examples in the wild, like this one from [Spring REST](https://howtodoinjava.com/spring-rest/http-options-request-handler/):

```http
OPTIONS /SpringRestExample/api/rest/employee-management/employees/1
 
Allow:           GET,DELETE,PUT,OPTIONS
```

> You could request GET /rpc/options_test_overloaded_at_least_one_non_volatile?num=1 - but that's something that OPTIONS /rpc/options_test_overloaded_at_least_one_non_volatile?num=1 needs to return. And this options request should then return OPTIONS,GET,HEAD only - no POST, because with POST you can't pass arguments via query string.

This makes me realize that OPTIONS for our table/view paths are also incomplete, because currently:

```http
OPTIONS /projects?id=eq.1

Allow: OPTIONS,GET,HEAD,POST,PUT,PATCH,DELETE
```

And the `Allow` should omit the POST there.

wolfgangwalther:
> So the assumption of the PR is that OPTIONS only identifies the URL-path(`/path`) capabilities. But OPTIONS actually identifies the resource capabilities(the whole URL with query parameters). The [RFC](https://tools.ietf.org/html/rfc7231#section-4.3.7) does mention this.

Thanks for putting it into much better words than I did :).

> This makes me realize that OPTIONS for our table/view paths are also incomplete, because currently:
> 
> ```httpspec
> OPTIONS /projects?id=eq.1
> 
> Allow: OPTIONS,GET,HEAD,POST,PUT,PATCH,DELETE
> ```
> 
> And the `Allow` should omit the POST there.

Uh. Absolutely right. And PUT should only be there, when the PK is given in the query string, too.

laurenceisla:
According to the discussion above, the routing refactor is needed to fix the issue. Closing.

Issue #1670, opened 2020-11-28
Feature request: Add ability to cast an input argument in a POST before INSERT


a-mckinley:
Hi,

I have a custom postgres type that is nested and relatively complex so I used ```jsonb_populate_record()``` server-side to create instances of this type from a json representation.

Currently to INSERT to this table from I would use a custom written server-side function to wrap the call to ```jsonb_populate_record()```.

It would be very convenient and versatile if PostgREST allowed you to apply a cast to input arguments e.g.

```
POST /projects?columns=id,thing::custom_type
``` 
which would hopefully apply a server-side cast created with

```sql
create function to_custom_type(json(b) object) returns custom_type as 
$$
    select json(b)_populate_record(null::custom_type,object);
$$ language sql immutable parallel safe;

create cast (json(b) as custom_type) with function to_custom_type(json(b)) as implicit;
```

Is such an enhancement possible? I think it could rather cleanly reduce the amount of specific server-side code needed in this case, but potentially has other uses that I haven't thought of yet.

wolfgangwalther:
`json[b]_populate_record[set]` work recursively on composite types: https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING-TABLE

But you already know that - that's why you are using it.

We are using it as well to process the json body:
https://github.com/PostgREST/postgrest/blob/9adec12a678afceca78f6bfebba236b031e63487/src/PostgREST/QueryBuilder.hs#L74

If your target tables column is set to your custom type - there is no reason to assume this wouldn't work out of the box. No cast needed.

Does it not work for you?

steve-chavez:
Related to this [gitter thread](https://gitter.im/begriffs/postgrest?at=5fbfe6167850f66b6045d6ee). In general I think it'd be simple to add. But I'm failing to come up with a use case because, as Wolfgang mentions, casting should already work thanks to `json_populate_recordset`.

wolfgangwalther:
> In general I think it'd be simple to add.

I don't think so. We don't touch the json before `json_populate_recordset`. And once this call is over, everything has already been converted to the target table's column types. Now, at this time we could add a cast, but this would either be a cast to the same type the column is already in - or to a type different from the target table's column type. There is literally no point in doing so... :)

We can't hook into the casting that `json_populate_recordset` does here.

@a-mckinley Did you try just inserting without any custom casts?

steve-chavez:
[...]

PR #1841, opened 2021-05-03
feat: Show comprehensive error when an RPC is not found in a stale schema cache

### LINKED ISSUE #1688: The stale schema cache
Sometimes the schema cache is not up to date anymore. This can happen during startup, or after changes to the database schema.

PostgREST's current approach is to continue to serve requests on a best effort basis - this is most notably the case for RPCs. RPC functions can be overloaded and need to be identified properly to set the correct transaction isolation level, create the right embedding queries and return the result appropriately formatted. For RPCs the stale schema cache issue is further amplified because `findProc` also tries to find the right function on a best effort basis - both with and without up to date schema cache.

This leads to all kinds of hard to unterstand (for the user) and hard to debug (for developers) issues, like #1469, #1638 and #1672. Also our issue template tells quite a story about it:

```
Before reporting a bug:
If your database schema has changed while the PostgREST server is running,
send the server a SIGUSR1 signal or restart it(http://postgrest.org/en/stable/admin.html#schema-reloading)
to ensure the schema cache is not stale. This sometimes fixes apparent bugs.
```

I propose to get rid of the whole "best effort" approach and do the following:

1. **find exact matches for RPCs or error out**: `findProc` should find an exact match for the function arguments/types. If no match is found, it should throw an error. If multiple matches are found, it could return some kind of 300 response, similar to the "hey I don't know which embedding you want" message. To be able to tell functions with the same name but different argument types apart we could add "hints" in the form of "virtual casts" for the arguments. Assume we have a two functions with the same name and same argument name - but one them has the argument as int, the other one as text. We could make requests like this:

  ```
  GET /rpc/func?arg::int=123
  GET /rpc/func?arg::text=asd
  ```

2. **Make calls to RPCs 100% explicit.** No "unknown" type arguments. We cast them all exactly the way we expect the function to be according to our request and schema cache. This avoids any "whoops, we called the wrong function here" situations.

_(There was a part here about updating the schema cache automatically on failed requests. This is useless, because those requests would error out anyway. New database objects could not be detected by this, so we can as well skip it.)_

This can be extended to requests to tables. Fail early if unknown columns are requested - no need to hit the database with it, trust the schema cache.

If the schema cache is out of date, we might return errors for requests that could be served, because the a function or table has been added in the meantime. This is not a problem. Before the function is created, we will get an error. After the schema cache is updated we get no error. This approach just delays the ability to call the function until the schema cache is up to date. This should happen anyway.

Force people to update their schema cache. Everybody will quickly understand "table/function not found" errors after adding stuff. Plus, since this error will be thrown directly from haskell, we can add "please update your schema cache, if this function should exist, bla bla" to it.

laurenceisla:
Implements the enhacement mentioned in #1688 (Closes #1688)

It will only cover the error messages when the schema is stale, not the explicit argument types casting for the overloaded functions.

- [x] Return an error message when the RPC's name and arguments are not found at all
- [x] Return the RPC's name and schema in the error message (handles a Nothing type now instead of a default ProcDescription)
- [x] Return a 300 message when the RPC is overloaded and the arguments have the same name but different types
- [x] Refactor / optimize code
- [x] Tests for the new features


src/PostgREST/App.hs
```diff
@@ -403,8 +403,10 @@ handleInfo identifier RequestContext{..} =
     hasPK =
       not $ null $ tablePKCols ctxDbStructure (qiSchema identifier) (qiName identifier)
 
-handleInvoke :: InvokeMethod -> ProcDescription -> RequestContext -> DbHandler Wai.Response
-handleInvoke invMethod proc context@RequestContext{..} = do
+handleInvoke :: InvokeMethod -> Maybe ProcDescription -> RequestContext -> DbHandler Wai.Response
+handleInvoke invMethod maybeProc context@RequestContext{..} = do
+  proc <- maybe (throwError Error.RpcNotFound) return maybeProc
```
wolfgangwalther, line 408: I think it would be better to throw an `ApiRequestError` in `userApiRequest` instead and revert the change to `TargetProc` regarding the Maybe. That would be much cleaner.

src/PostgREST/Error.hs
```diff
@@ -289,6 +291,9 @@ instance JSON.ToJSON Error where
   toJSON (JwtTokenInvalid message) = JSON.object [
     "message" .= (message :: Text)]
   toJSON NotFound = JSON.object []
+  toJSON RpcNotFound = JSON.object [
+    "hint"    .= ("No function matches the given name and argument types. If a new function was created in the database with this name, try updating the schema cache." :: Text),
```
wolfgangwalther, line 295:
```suggestion
    "hint"    .= ("No function matches the given name and arguments. If a new function was created in the database with this name, try reloading the schema cache." :: Text),
```

steve-chavez, line 295: Hm. The `hint` should serve as a guide for how to clear the error.  So the `No function matches the given name and arguments` message doesn't belong to the `hint`. It should go in `error`.

src/PostgREST/Error.hs
```diff
@@ -289,6 +291,9 @@ instance JSON.ToJSON Error where
...
+    "message" .= ("This function does not exist" :: Text)]
```
steve-chavez, line 296: To be more precise, the error should be phrased in terms of `Could not find..`:

```haskell
    "message" .= ("Could not find the add_them(a,b) function")
```

src/PostgREST/App.hs
laurenceisla: I agree, that may help to make better error messages too (solving the second bullet point of my first comment)

src/PostgREST/Error.hs
laurenceisla: I'll use this and @steve-chavez 's feedback to change the error message.

laurenceisla: Right now, postgrest shows the `No function matches the given name and arguments` error in the hint when a function is not found (the database raises this error). But I agree that it should be part of the message.

src/PostgREST/Error.hs
laurenceisla: Agree, it's a better description of the issue.

src/PostgREST/DbStructure/Proc.hs
```diff
@@ -77,9 +77,13 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
...
   Ideally, handling overloaded functions should be left to pg itself. But we need to know certain proc attributes in advance.
 -}
-findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> Maybe ProcDescription
-findProc qi payloadKeys paramsAsSingleObject allProcs = bestMatch
+findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> (ProcDescription, Bool)
+findProc qi payloadKeys paramsAsSingleObject allProcs = (fromMaybe fallback bestMatch, isFoundInSchema)
   where
+    -- instead of passing Maybe ProcDescription around, we create a fallback description here when we can't find a matching function
+    -- args is empty, but because "specifiedProcArgs" will fill the missing arguments with default type text, this is not a problem
+    fallback = ProcDescription (qiSchema qi) (qiName qi) Nothing mempty (SetOf $ Composite $ QualifiedIdentifier mempty "record") Volatile False
```
steve-chavez: Hm, I was liking how the fallback went away before; because we no longer fallback and instead err right?

Would it possible to remove the fallback and still give a clean error?

laurenceisla: I think It may be possible, although there are some error validating functions that use `target` with that fallback value, like this one

https://github.com/laurenceisla/postgrest/blob/cd6d6784c2ee34cd5b987f162236cfbe8d929919/src/PostgREST/Request/ApiRequest.hs#L168

I'll check and let you know if it breaks too many things.

src/PostgREST/DbStructure/Proc.hs
```diff
@@ -77,22 +79,31 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
   An overloaded function can have a different volatility or even a different return type.
   Ideally, handling overloaded functions should be left to pg itself. But we need to know certain proc attributes in advance.
 -}
-findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> ProcDescription
-findProc qi payloadKeys paramsAsSingleObject allProcs = fromMaybe fallback bestMatch
+findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> (NonEmpty ProcDescription, Bool)
```
wolfgangwalther, line 82: That return type seems strange.

Can we not just return a list of `ProcDescription` and then later do:
- if length = 0 -> not found error
- if length = 1 -> all good
- if length > 1 -> duplicate error

src/PostgREST/Error.hs
```diff
@@ -91,6 +97,12 @@ instance JSON.ToJSON ApiRequestError where
...
     "message" .= ("More than one relationship was found for " <> parent <> " and " <> child :: Text),
     "details" .= (compressedRel <$> rels) ]
+  toJSON (RpcNotUnique procs)  = JSON.object [
+    "hint"    .= ("Explicit argument type casts are needed. Possible endpoints are: " <> T.intercalate " , " ["/rpc/" <> pdName p <> "?" <> T.intercalate "&" [pgaName a <> "::" <> pgaType a <> "=value" | a <- pdArgs p] | p <- procs] :: Text),
+    "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [pgaName a <> " => " <> pgaType a | a <- pdArgs p] <> ")" | p <- procs])]
+  toJSON (RpcNotFound schema procName payloadKeys)  = JSON.object [
+    "hint"    .= ("If a new function was created in the database with this name and arguments, try reloading the schema cache." :: Text),
+    "message" .= T.unwords ["Couldn't find the", schema <> "." <> procName <> "(" <> T.intercalate ", " payloadKeys <> ")", "function"]]
```
wolfgangwalther, line 105: Similar to the two `xRelBetween` errors, we could name those two:
- `AmbiguousRpc`
- `NoRpc`

WDYT?

laurenceisla: Agree, it should keep the consintency of the names within the `ApiRequestError`.

src/PostgREST/DbStructure/Proc.hs
laurenceisla: It's a problem related with the `fallback` function that we commented before. There are some complications that I need to look into with more detail. After I  finish the argument type casting and the tests, I'll refactor this part of the code as you mentioned.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -230,12 +238,11 @@ userApiRequest conf@AppConfig{..} dbStructure req reqBody
...
-    TargetProc _ _ -> True
-    _              -> False
-  isTargetingDefaultSpec = case target of
-    TargetDefaultSpec _ -> True
-    _                   -> False
+  (isTargetingProc, targetProcIsRootSpec, isTargetingDefaultSpec) = case path of
+    [] | isJust configDbRootSpec -> (True, True, False)
+       | otherwise               -> (False, False, True)
+    ["rpc", _]                   -> (True, False, False)
+    _                            -> (False, False, False)
```
laurenceisla, line 245: Alright, the procedure error handling https://github.com/PostgREST/postgrest/pull/1841#discussion_r633287931 without the fallback procedure https://github.com/PostgREST/postgrest/pull/1841#discussion_r626902462 looks better now. There are some changes that needed to be made, like replacing the `target` function call in some places in order to make it the last error to be checked (in these selected lines for instance).

Need some feedback and suggestions to make this change a bit cleaner.

src/PostgREST/DbStructure/Proc.hs
```diff
@@ -77,30 +79,39 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
...
+                               args -> matchesArg args
+    matchesArg args =
+      case L.partition pgaReq args of
+        (reqArgs, [])      -> payloadKeys == S.fromList (castedArgs reqArgs)
+        ([], defArgs)      -> payloadKeys `S.isSubsetOf` S.fromList (castedArgs defArgs)
+        (reqArgs, defArgs) -> payloadKeys `S.difference` S.fromList (castedArgs defArgs) == S.fromList (castedArgs reqArgs)
+    castedKeys =
+      S.filter (not . T.null . snd) (S.map (T.breakOn "::") payloadKeys)
+    castedArgs args=
+      [if pgaName a `S.member` S.map fst castedKeys then pgaName a <> "::" <> pgaType a else pgaName a | a <- args]
```
laurenceisla, line 106: Looks for the casted keys from the client request and matches it with the converted arguments found for the Rpc.

src/PostgREST/DbStructure/Proc.hs
```diff
@@ -77,30 +79,39 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
...
 
 {-|
   Search the procedure parameters by matching them with the specified keys.
   If the key doesn't match a parameter, a parameter with a default type "text" is assumed.
 -}
 specifiedProcArgs :: S.Set FieldName -> ProcDescription -> [PgArg]
 specifiedProcArgs keys proc =
-  (\k -> fromMaybe (PgArg k "text" True False) (find ((==) k . pgaName) (pdArgs proc))) <$> S.toList keys
+  (\k -> fromMaybe (PgArg k "text" True False) (find ((==) k . pgaName) (pdArgs proc))) <$> S.toList (S.map (fst . T.breakOn "::") keys)
```
laurenceisla, line 114: Filters the existing arguments using the keys from the client request without the explicit casting.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -113,9 +113,10 @@ toRpcParamValue proc (k, v) | argIsVariadic k = (k, Variadic [v])
   where
     argIsVariadic arg = isJust $ find (\PgArg{pgaName, pgaVar} -> pgaName == arg && pgaVar) $ pdArgs proc
 
--- | Convert rpc params `/rpc/func?a=val1&b=val2` to json `{"a": "val1", "b": "val2"}
+-- | Convert rpc params `/rpc/func?a::type=val1&b=val2` to json `{"a": "val1", "b": "val2"}
 jsonRpcParams :: ProcDescription -> [(Text, Text)] -> PayloadJSON
-jsonRpcParams proc prms =
+jsonRpcParams proc castedParams =
+  let prms = first (fst . T.breakOn "::") <$> castedParams in
```
laurenceisla, line 119: The casted params are converted to a json "payload"  without the explicit casting from the client request.

laurenceisla (commented):
Would also appreciate some feedback here, not sure if this is the right way to solve this problem. 
This is my approach to the "/rpc/name?arg::type=value" transformation. Works for GET/HEAD and POST from html forms. For now, json POSTs return a `null` value for the argument (because the payload json body isn't touched).

test/Feature/RpcSpec.hs
```diff
@@ -689,7 +780,7 @@ spec actualPgVersion =
 
     context "only for POST rpc" $ do
       it "gives a parse filter error if GET style proc args are specified" $
-        post "/rpc/sayhello?name=John" [json|{}|] `shouldRespondWith` 400
+        post "/rpc/sayhello?name=John" [json|{name: "John"}|] `shouldRespondWith` 400
```
steve-chavez, line 783: Hm, why did this test change?

test/Feature/RpcSpec.hs
```diff
@@ -96,22 +96,113 @@ spec actualPgVersion =
...
+        { matchStatus  = 404
+        , matchHeaders = [matchContentTypeJson]
+        }
+
+    context "ambiguous overloaded functions with same arguments but different types" $ do
+      it "should fail with 300 Multiple Choises without explicit argument type casts" $
+        get "/rpc/overloaded_same_args?arg=value" `shouldRespondWith`
+          [json| {
+            "hint":"Explicit argument type casts are needed. Possible endpoints are: /rpc/overloaded_same_args?arg::integer=value , /rpc/overloaded_same_args?arg::xml=value , /rpc/overloaded_same_args?arg::text=value&num::integer=value",
+            "message":"Could not choose the best candidate function between: test.overloaded_same_args(arg => integer), test.overloaded_same_args(arg => xml), test.overloaded_same_args(arg => text, num => integer)" } |]
```
steve-chavez, line 158: Typo in `300 Multiple Choises` - should be `Choices`.

Btw, really nice error message! This is much better than the ambiguous error for multiple relationships.

src/PostgREST/DbStructure/Proc.hs
```diff
@@ -77,30 +79,39 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
...
+                               [arg] -> pgaType arg `elem` ["json", "jsonb"]
+                               _     -> False
+      | otherwise            = case pdArgs proc of
+                               []   -> null payloadKeys
+                               args -> matchesArg args
+    matchesArg args =
+      case L.partition pgaReq args of
+        (reqArgs, [])      -> payloadKeys == S.fromList (castedArgs reqArgs)
+        ([], defArgs)      -> payloadKeys `S.isSubsetOf` S.fromList (castedArgs defArgs)
+        (reqArgs, defArgs) -> payloadKeys `S.difference` S.fromList (castedArgs defArgs) == S.fromList (castedArgs reqArgs)
```
steve-chavez, line 102: Lots of work here, it's a bit hard to follow. Perhaps you could add a comment?

src/PostgREST/DbStructure/Proc.hs
```diff
@@ -77,30 +79,39 @@ type ProcsMap = M.HashMap QualifiedIdentifier [ProcDescription]
   An overloaded function can have a different volatility or even a different return type.
   Ideally, handling overloaded functions should be left to pg itself. But we need to know certain proc attributes in advance.
 -}
-findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> ProcDescription
-findProc qi payloadKeys paramsAsSingleObject allProcs = fromMaybe fallback bestMatch
+filterProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> [ProcDescription]
```
steve-chavez, line 82: I think the correct return type here should be an `Either ApiRequestError ProcDescription` and not `[ProcDescription]`. Your work [below](https://github.com/PostgREST/postgrest/pull/1841/files#diff-6f407ba6cfd206036588fcd2d4d9583a65800906f957ecd40d79f9e2fec41bffR316-R318) suggests that should be the case. Also our `findRel` function for finding the relationships has the same type.

test/Feature/RpcSpec.hs
laurenceisla: After the change, postgrest looks for the function existence first. It doesn't find a sayhello function without parameters, it gives a 404 error instead of the 400. Only when the function exists it checks for parse filter errors, that's why I added the parameter in the json payload.

src/PostgREST/DbStructure/Proc.hs
wolfgangwalther: Let me respond do that, as I suggested earlier to make it return `[ProcDescription]` here:

This is in `dbStructure/Proc.hs` - I think this code should not be concerned about returning `ApiRequestError`s. This helper just returns the list of all matching procs we have in the cache.

The lines you referenced are in `ApiRequest.hs`, where it makes a lot of sense to turn the number of procs returned into `ApiRequestError`s, when <> 1.

---

Although, there are a few bits in this function right now that really belong into `ApiRequest` already: The parsing of the cast `::` syntax for one. But also "payload" in `payloadKeys` and `paramsAsSingleObject` are `ApiRequest` concerns.

It would be better overall, if the `filterProc` would have an interface that takes a list of arguments (name + type) to look for.

Maybe create a type like `data Argument = UnnamedArgument | NamedArgument Text | TypedArgument Text Text` or something like that and then make the function take `[Argument]` instead of `S.Set Text -> Bool`.

test/Feature/RpcSpec.hs
steve-chavez: Got it. The change is harmless, so it's cool to leave it like this.

src/PostgREST/DbStructure/Proc.hs
steve-chavez: > This is in dbStructure/Proc.hs - I think this code should not be concerned about returning ApiRequestErrors. 

True, but how about if we move the function to ApiRequest.hs?

I think `findProc` should be similar to [findRel](https://github.com/PostgREST/postgrest/blob/5c75f0dcc295e6bd847af6d9703fad5b9c3d76c9/src/PostgREST/Request/DbRequestBuilder.hs#L151-L165), which has complex logic, but it's all encapsulated in a single function(that also returns the Left or Right). Having that logic split between different functions in different modules would make it harder to understand.

wolfgangwalther: Yeah, that would be fine, too. Although the raw parsing of `::` casts should still happen at an earlier stage, I think.

steve-chavez:
@laurenceisla I see you've implemented the arg casting as proposed on https://github.com/PostgREST/postgrest/issues/1688.

```
GET /rpc/func?arg::int=123
GET /rpc/func?arg::text=asd
```

How would that work for disambiguating POST RPC requests?

IIUC, that's not possible right now. And if we're trying to be 100% correct, we should cover that case.

Some time ago, I [proposed](https://github.com/PostgREST/postgrest/pull/946#issue-136635374) adding an additional `argtypes` query param, e.g. ` argtypes=id::int,name::text`. That could work for both POST and GET.

However, right now I'm finding the `findProc` function to be complex. And the overloaded case where a parameter with the same name has a different type is a bit contrived. So I'm wondering if being 100% correct is worth it here.

Wouldn't be better to err when this contrived case happens? After all, the net win of https://github.com/PostgREST/postgrest/issues/1688 is a better error message.

@laurenceisla @wolfgangwalther WDYT?

wolfgangwalther:
> @laurenceisla I see you've implemented the arg casting as proposed on #1688.
> 
> ```
> GET /rpc/func?arg::int=123
> GET /rpc/func?arg::text=asd
> ```
> 
> How would that work for disambiguating POST RPC requests?
> 
> IIUC, that's not possible right now. And if we're trying to be 100% correct, we should cover that case.
> 
> Some time ago, I [proposed](https://github.com/PostgREST/postgrest/pull/946#issue-136635374) adding an additional `argtypes` query param, e.g. ` argtypes=id::int,name::text`. That could work for both POST and GET.

I think the best way of doing this for `POST` is by using the `columns=` query param, we already have. This is about "describing" the POST-ed body already in terms of "which columns to include", so we could as well add the casts here?

> However, right now I'm finding the `findProc` function to be complex.

Agreed. Not sure where @monacoremo is hiding right now, but we might just try to throw it his way once we're done and he is on another refactor-spree... ;)

> And the overloaded case where a parameter with the same name has a different type is a bit contrived. So I'm wondering if being 100% correct is worth it here.
> 
> Wouldn't be better to err when this contrived case happens?

I think it's great to be able to cover those cases. Yes, when designing a schema from scratch to use it with postgrest - most likely, you'd work around that with different argument names. After all it doesn't make much difference - you would have to change the query string in any case.

But the use-case of exposing an existing schema is different, I think.

steve-chavez:
> I think the best way of doing this for POST is by using the columns= query param, we already have. This is about "describing" the POST-ed body already in terms of "which columns to include", so we could as well add the casts here?

We had a feature request for casting POST columns [before](https://github.com/PostgREST/postgrest/issues/1670), I guess it could be reused for rpc.

Also, the `?arg::int=123` looks too irregular. I don't think we have another syntax where the param(`?param=value`) in the querystring is that loaded. I guess we could reuse `?columns` for just the overloaded case? Not sure for now.

> But the use-case of exposing an existing schema is different, I think.

Hm, I think most overloaded cases would come from pg extensions, not from users schemas.But in any case, any existing schema cannot be exposed just like that, it will need some work as documented on [schema isolation](https://postgrest.org/en/stable/schema_structure.html).

src/PostgREST/Request/ApiRequest.hs
steve-chavez: This will change once you change the `findProc` to return an Either; let's see if it's simpler then. For now, from the previous version, it looks like `isTargetingProc/isTargetingDefaultSpec` could be outside here and go next to `ApiRequest.Target`; and become independent functions.

wolfgangwalther:
> I guess we could reuse `?columns` for just the overloaded case on GET? Not sure for now.

Don't like it. The `arg::type` syntax has everything in one place. No need to split it up in that case. For POST, we already have `columns` and don't have a better way right now - so it makes sense.

laurenceisla:
> > Wouldn't be better to err when this contrived case happens? After all, the net win of #1688 is a better error message.

> I think it's great to be able to cover those cases. Yes, when designing a schema from scratch to use it with postgrest - most likely, you'd work around that with different argument names. After all it doesn't make much difference - you would have to change the query string in any case.

I also think postgrest should be able to cover these cases, but it seems like the implementation of explicit casts is getting too complicated for this PR and needs more discussion. I'll leave it to cover only the error messages.

src/PostgREST/Request/ApiRequest.hs
laurenceisla: Modified the `findProc` to return an `Either ApiRequestError Target` but couldn't see how to refactor this part of the code to make it look cleaner without complicating things after this change. Mostly because these functions are needed before any `Target` is assigned. I'll leave it like this for now and open the PR for reviews.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -448,10 +450,46 @@ requestContentTypes conf action target =
...
+{-|
+  Search a pg procedure by its parameters. Since a function can be overloaded, the name is not enough to find it.
+  An overloaded function can have a different volatility or even a different return type.
+-}
+findTargetProc :: QualifiedIdentifier -> S.Set Text -> Bool -> Bool -> ProcsMap -> Either ApiRequestError Target
+findTargetProc qi payloadKeys paramsAsSingleObject isRootSpec allProcs =
+  case bestMatch of
+    []     -> Left $ NoRpc (qiSchema qi) (qiName qi) (S.toList payloadKeys) paramsAsSingleObject
+    [proc] -> Right $ TargetProc proc isRootSpec
+    procs  -> Left $ AmbiguousRpc (toList procs)
```
steve-chavez, line 470: I like how the `Lefts` are now inside the function, that makes the logic easier to understand.

However, it's weird how it returns a `Target` type - which can be a `TargetDefaultSpec`, among other things.

Is it possible to return a `ProcDescription` instead?

src/PostgREST/Error.hs
```diff
@@ -92,6 +98,12 @@ instance JSON.ToJSON ApiRequestError where
     "hint"    .= ("By following the 'details' key, disambiguate the request by changing the url to /origin?select=relationship(*) or /origin?select=target!relationship(*)" :: Text),
     "message" .= ("More than one relationship was found for " <> parent <> " and " <> child :: Text),
     "details" .= (compressedRel <$> rels) ]
+  toJSON (AmbiguousRpc procs)  = JSON.object [
+    "hint"    .= ("Overloaded functions with the same argument name but different types are not supported" :: Text),
+    "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [pgaName a <> " => " <> pgaType a | a <- pdArgs p] <> ")" | p <- procs])]
+  toJSON (NoRpc schema procName payloadKeys hasPreferSingleObject)  = JSON.object [
+    "hint"    .= ("If a new function was created in the database with this name and arguments, try reloading the schema cache." :: Text),
+    "message" .= ("Couldn't find the " <> schema <> "." <> procName <> if hasPreferSingleObject then " function with a single json or jsonb argument" else "(" <> T.intercalate ", " payloadKeys <> ")" <> " function")]
```
steve-chavez, line 106: I think we could add the "Could not find the function.. **in the schema cache**", similar to the error we have for relationships now: https://postgrest.org/en/latest/schema_cache.html#stale-foreign-key-relationships

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -94,6 +94,16 @@ data Action = ActionCreate       | ActionRead{isHead :: Bool}
...
+-- | The path info that will be mapped to a target (used to handle validations and errors before defining the Target)
+data Path
+  = PathInfo
+      { pSchema        :: Schema,
+        pName          :: Text,
+        pHasRpc        :: Bool,
+        pIsDefaultSpec :: Bool,
+        pIsRootSpec    :: Bool
+      }
+  | PathUnknown
```
steve-chavez, line 106: This type does still carry almost the same info as our Target type, however I'm not seeing a way to refactor it for now; so it's good to leave it like that.

The whole `ApiRequest` module needs a big refactor, it has turned messy over the years.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -448,10 +469,46 @@ requestContentTypes conf action target =
...
+      -- The function's required arguments are separated from the ones with a default value assigned.
+      -- The set of names of those arguments is compared to the set of keys supplied by the client
+      -- 1. If only required arguments are found, the keys must be exactly the same as those arguments
+      -- 2. If only optional arguments are found, the keys must be a subset of those arguments
+      -- 3. If both required and optional arguments are found, the result of taking away the optional arguments
+      --    from the keys must be exactly the same as the required arguments
+      case L.partition pgaReq args of
+        (reqArgs, [])      -> payloadKeys == S.fromList (pgaName <$> reqArgs)
+        ([], defArgs)      -> payloadKeys `S.isSubsetOf` S.fromList (pgaName <$> defArgs)
+        (reqArgs, defArgs) -> payloadKeys `S.difference` S.fromList (pgaName <$> defArgs) == S.fromList (pgaName <$> reqArgs)
```
steve-chavez, line 514: @laurenceisla This is the only part that I still need to check. I'll run some bench tests to see if there's an impact on RPC performance. Ideally, the non-overloaded case shouldn't suffer a perf penalty.

laurenceisla: I've run the benchmark tests and these are the results. 

```
pgrbench-k6 t3anano k6/RPCSimple.js
    http_reqs..................: 55521  1844.785613/s
    http_reqs..................: 56208  1868.872353/s
    http_reqs..................: 56631  1871.617673/s

pgrbench-k6 t3anano k6/RPCGETSingle.js
    http_reqs..................: 61239  2030.219531/s
    http_reqs..................: 60722  2001.174361/s
    http_reqs..................: 60647  2016.992396/s

pgrbench-k6 t3anano k6/RPCGETSingleEmbed.js
    http_reqs..................: 24001  798.671285/s
    http_reqs..................: 24001  794.321514/s
    http_reqs..................: 23997  796.039754/s
```

Compared to the default ones in the benchmark there's a small improvement in the RPCSimple test while the others give similar results. I'll build the latest nightly and see how they compare.

laurenceisla: These are the benchmarks for the latest nightly f3a184af017381129245db6c98743963af9b6d45 . I still need to resolve the merge conflicts with this branch and, if the results are too different, I'll update the benchmark I made here: https://github.com/PostgREST/postgrest/pull/1841#discussion_r661919724

```
pgrbench-k6 t3anano k6/RPCSimple.js
    http_reqs..................: 55615  1848.913494/s
    http_reqs..................: 55428  1842.921351/s
    http_reqs..................: 55743  1842.730914/s


pgrbench-k6 t3anano k6/RPCGETSingle.js
    http_reqs..................: 57051  1896.411094/s
    http_reqs..................: 58782  1940.730047/s
    http_reqs..................: 59174  1967.293803/s

pgrbench-k6 t3anano k6/RPCGETSingleEmbed.js
    http_reqs..................: 24001  798.712055/s
    http_reqs..................: 23979  797.954143/s
    http_reqs..................: 24000  798.706433/s
```

steve-chavez: @laurenceisla So according to your results [here](https://github.com/PostgREST/postgrest/pull/1841#discussion_r661919724), we're actually improving RPC perf!

Why is that? The above logic looked a bit more complicated.

laurenceisla: > These are the benchmarks for the latest nightly f3a184a . I still need to resolve the merge conflicts with this branch and, if the results are too different, I'll update the benchmark I made here: #1841 (comment)

@steve-chavez  I resolved the merge conflicts and redid the benchmarks. The results are now very similar to the nightly I mentioned here. So, it looks like the performance isn't better or worse.

> @laurenceisla So according to your results here, we're actually improving RPC perf!
>
> Why is that? The above logic looked a bit more complicated.

I thought it was because the `matchesArgs` function now verifies using equality for RPCs without optional parameters, instead of using the `isSubsetOf` function all the time. But with this last merge, it seems like it didn't affect the performance in any way.

steve-chavez: Ok, cool. Good to know it doesn't affect performance. I'll merge now.

PR #1855, opened 2021-05-24
Fix regression on null PK

### LINKED ISSUE #1832: Regression: Nightly throws 500 "Unexpected NULL"
### Environment

* PostgreSQL version: 13.2
* PostgREST version: nightly (d3a8b5f6e1f541c5c6cde9398a7ba6a95cdbdb88)
* Operating system: FreeBSD 13

### Description of issue

There's a regression after 343e41c51d6cad33383043abb4ef15dc54a32245

PostgREST throws 500 "Unexpected NULL" instead 201 "Created".

@wolfgangwalther, why had you removed null PK value processing on this? https://github.com/PostgREST/postgrest/blob/6670a3214bc2f74eac10ff32eb69a38db3807eaf/src/PostgREST/Query/SqlFragment.hs#L174

### Testcase
```PLpgSQL
DROP VIEW IF EXISTS test_bug;
DROP TABLE IF EXISTS test_a;
DROP TABLE IF EXISTS test_b;

CREATE TABLE test_b
(
    id int,

    PRIMARY KEY (id)
);

CREATE TABLE test_a
(
    id   int,
    b_id int,

    PRIMARY KEY (id),
    FOREIGN KEY (b_id) REFERENCES test_b
);

CREATE OR REPLACE VIEW test_bug AS
SELECT a.id, b.id AS b_id
FROM test_a AS a
     LEFT JOIN test_b AS b ON a.b_id = b.id;

CREATE RULE test_bug_insert AS ON INSERT TO test_bug DO INSTEAD
    INSERT INTO test_a(id, b_id) VALUES (new.id, new.b_id) RETURNING id, b_id;
```

```HTTP
POST http://postgrest/test_bug

{"id": 1}

...

{
  "message": "Row error: Wrong value parser used",
  "details": "Unexpected NULL"
}
```

drTr0jan:
Closes #1832

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1636, Fix `application/octet-stream` appending `charset=utf-8` - @steve-chavez
  - #1615, Fix RPC return type handling and embedding for domains with composite base type - @wolfgangwalther
  - #1469, #1638 Fix overloading of functions with unnamed arguments - @wolfgangwalther
+ - #1832, #1855 Fix regression on null PK - @drTr0jan
 
 ### Changed
 
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -171,7 +171,7 @@ asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldNam
 locationF :: [Text] -> SqlFragment
 locationF pKeys = [qc|(
   WITH data AS (SELECT row_to_json(_) AS row FROM {sourceCTEName} AS _ LIMIT 1)
-  SELECT array_agg(json_data.key || '=eq.' || json_data.value)
+  SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null'))
   FROM data CROSS JOIN json_each_text(data.row) AS json_data
   WHERE json_data.key IN ('{fmtPKeys}')
 )|]


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1636, Fix `application/octet-stream` appending `charset=utf-8` - @steve-chavez
  - #1615, Fix RPC return type handling and embedding for domains with composite base type - @wolfgangwalther
  - #1469, #1638 Fix overloading of functions with unnamed arguments - @wolfgangwalther
+ - #1832, #1855 Fix regression on null PK - @drTr0jan
```
wolfgangwalther, line 31: We don't need to change the changelog manually, as we'll create it from the commit history.

wolfgangwalther (commented):
Thanks for the PR!

From the issue:

> Unfortunately I'm not good at Haskell for preparing tests, but I've tested the PR in my environment.

That's not hard. Just basically just need to copy the SQL reproduction posted in #1832 and add it to `test/fixtures/schema.sql`. The new view will need privileges in `test/fixtures/privileges.sql`. And then you can copy and adjust one of the location header tests in `tests/Feature/InsertSpec.hs`, e.g. this one:
https://github.com/PostgREST/postgrest/blob/4b46c4eff329c91dc13b26c78b266c7361ecace4/test/Feature/InsertSpec.hs#L114-L121

laurenceisla:
> Unfortunately I'm not good at Haskell for preparing tests, but I've tested the PR in my environment.

Hello, @drTr0jan I grabbed your commit and added a test in #1875

steve-chavez:
Superseded by #1875. Thank you for your work here @drTr0jan.

PR #1875, opened 2021-06-18
fix: Regression on null PK inserts on views for headers only representation

### LINKED ISSUE #1832: Regression: Nightly throws 500 "Unexpected NULL"

laurenceisla:
Closes #1832

Added the PR #1855 missing tests.


test/Feature/InsertSpec.hs
```diff
@@ -537,3 +537,13 @@ spec actualPgVersion = do
             , matchHeaders = [ "Location" <:> "/compound_pk_view?k1=eq.1&k2=eq.test"
                              , "Content-Range" <:> "*/*" ]
             }
+
+      it "should not throw and return location header when a PK is null" $
```
steve-chavez, line 541: I'm really puzzled as to how a PK can be null. Perhaps we can add a short comment of how that happens?

laurenceisla: For instance, in the use case exemplified in this test, PostgREST is recognizing both the `id` and `sponsor_id` columns of the VIEW as PRIMARY KEYS. In this case, a `competitor` without a `sponsor` is possible, this means that the column `sponsor_id` (identified as a PK in the VIEW) could have a NULL value.

With this regression fix, at the moment the Location Header is built, PostgREST handles this NULL value for the `sponsor_id` and returns the header correctly.

steve-chavez: Cool. So it's the inferred PKs on the VIEW that are causing the issue - 2 PKs for this view, and one of them is null.

test/Feature/InsertSpec.hs
```diff
@@ -537,3 +537,13 @@ spec actualPgVersion = do
...
+        request methodPost "/test_null_pk_competitors_sponsors" [("Prefer", "return=headers-only")]
+            [json|{"id":1}|]
+          `shouldRespondWith`
+            ""
+            { matchStatus  = 201
+            , matchHeaders = [ "Location" <:> "/test_null_pk_competitors_sponsors?id=eq.1&sponsor_id=is.null"
```
steve-chavez, line 547: Just to clarify, `test_null_pk_competitors_sponsors` doesn't have`sponsor_id` as a pk column:

```sql
CREATE TABLE competitors (
    id integer primary key,
    sponsor_id integer references sponsors(id),
    full_name varchar(40) not null
);

CREATE VIEW test_null_pk_competitors_sponsors  AS
SELECT c.id, s.id as sponsor_id
FROM competitors c
    LEFT JOIN sponsors s on c.sponsor_id = s.id;
```

So the above would be a bug.

Issue #1622, opened 2020-10-14
SUGGESTION - api "signal handler" to reset cache - SIGUSR1


mpasquini:
It is possible to have an api call to the signal handler ? 

> Before reporting a bug:
> If your database schema has changed while the PostgREST server is running,
> send the server a SIGUSR1 signal or restart it(http://postgrest.org/en/stable/admin.html#schema-reloading)
> to ensure the schema cache is not stale. This sometimes fixes apparent bugs.

From GNU: 

> These signals are used for various other purposes. In general, they will not affect your program unless it explicitly uses them for something.
> 
> Macro: int SIGUSR1
> Macro: int SIGUSR2
> 
>     The SIGUSR1 and SIGUSR2 signals are set aside for you to use any way you want. They’re useful for simple interprocess communication, if you write a signal handler for them in the program that receives the signal.
>

wolfgangwalther:
> It is possible to have an api call to the signal handler ?

I'm not sure I undestand what you mean - but are you maybe looking for #1542 ? (not released yet)

steve-chavez:
I think @mpasquini might be referring to calling SIGUSR1 through an endpoint, like `/internal/sigusr1`.

It kind of reminds to https://github.com/PostgREST/postgrest/issues/1526 in this regard.

wolfgangwalther:
Ah, ok. That could be done with something like the following already?

```sql
CREATE FUNCTION sigusr1() RETURNS void
LANGUAGE plpgsql AS $$
BEGIN
  NOTIFY pgrst;
END$$;
```

And then call it via `POST /rpc/sigusr1`.

Of course `db-channel-enabled=true` needs to be set.

wolfgangwalther:
Can this be closed?

steve-chavez:
> Of course db-channel-enabled=true needs to be set.

Maybe we should default `db-channel-enabled` to true. We disabled it because of connection poolers https://github.com/PostgREST/postgrest/pull/1542#issuecomment-643767295. However dwagin told us(https://github.com/PostgREST/postgrest/pull/1633#issuecomment-718911934) that hasql-transaction still needs to be patched for poolers to work. We can put a warning about disabling `db-prepared-statements` and `db-channel-enabled` for poolers to work later on, when we're sure we work with them.

wolfgangwalther:
[...]

PR #1880, opened 2021-06-23
Change the default db-channel-enabled config value from False to True

laurenceisla:
BREAKING CHANGE

According to the discussion in https://github.com/PostgREST/postgrest/issues/1622#issuecomment-733066644, there's a good explanation as to why the `db-channel-enabled` should be True by default.


steve-chavez:
LGTM. Btw, I changed the BREAKING CHANGE note on the commit because it's not really a breaking change - db-channel is not yet released on a stable version.

PR #1881, opened 2021-06-24
feat: add openapi-mode config

### LINKED ISSUE #1449: OpenAPI output that doesn't consider role privileges
Basically, users would like to get all the schema info through OpenAPI without caring for privileges. This was also discussed in https://github.com/PostgREST/postgrest/issues/1060 and we get this question on gitter chat from time to time. The workaround is cumbersome — adding a hardcoded high privileged JWT for the root endpoint. 

I propose we include a config option called `openapi-follows-acl`, default to true to avoid a breaking change. When this is false we'll use our schema cache info without considering privileges.

steve-chavez:
Fixes #1449. Adds the `openapi-mode` config:

* `openapi-mode="follow-acl"`(default): OpenAPI output follows access control for the JWT role.
* `openapi-mode="ignore-acl"`: OpenAPI output  ignores access control for the JWT role.
* `openapi-mode="disabled"`: disables OpenAPI output, the root endpoint replies with 404 Not Found.

#### Pending

- [ ] Add tests

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -16,6 +16,7 @@ module PostgREST.Config
   , JSPath
   , JSPathExp(..)
   , LogLevel(..)
+  , OpenAPIMode(..)
   , Proxy(..)
   , toText
   , isMalformedProxyUri
@@ -85,6 +86,7 @@ data AppConfig = AppConfig
   , configJwtSecret             :: Maybe B.ByteString
   , configJwtSecretIsBase64     :: Bool
   , configLogLevel              :: LogLevel
+  , configOpenApiMode           :: OpenAPIMode
   , configOpenApiServerProxyUri :: Maybe Text
   , configRawMediaTypes         :: [B.ByteString]
   , configServerHost            :: Text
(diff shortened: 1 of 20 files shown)


src/PostgREST/Config.hs
```diff
@@ -101,6 +103,9 @@ instance Show LogLevel where
   show LogWarn  = "warn"
   show LogInfo  = "info"
 
+data OpenAPIMode = OAIFollowACL | OAIIgnoreACL | OAIDisabled
+  deriving Eq
```
laurenceisla, line 107: For "OpenAPI", I like OA better as an acronym than OAI, e.g. `OAFollowACL`, `OADisabled`. WDYT?

src/PostgREST/Config.hs
```diff
@@ -128,6 +133,7 @@ toText conf =
       ,("jwt-secret-is-base64",          T.toLower . show . configJwtSecretIsBase64)
       ,("log-level",                 q . show . configLogLevel)
       ,("openapi-server-proxy-uri",  q . fromMaybe mempty . configOpenApiServerProxyUri)
+      ,("openapi-server-proxy-uri",  q . fromMaybe mempty . configOpenApiServerProxyUri)
```
laurenceisla, line 136:
```suggestion
```
Looks like this setting is duplicated.

src/PostgREST/Config.hs
steve-chavez: Yeah, that's better. The OAI term from https://github.com/OAI, somehow stuck for me, but it's a different thing.

src/PostgREST/Config.hs
steve-chavez: Right, that was a mistake.

PR #1886, opened 2021-07-04
Correct openapi ignore-privs table comments

steve-chavez:
The new `openapi-mode=ignore-privileges` didn't output `COMMENTs` on tables. Seems this wasn't considered back in https://github.com/PostgREST/postgrest/pull/885.

Comments on columns were already included, so this is just a change to make the output consistent.

I've tested it on the private big schema(327 tables/views) and there's no apparent change in query speed.


src/PostgREST/DbStructure.hs
```diff
@@ -466,7 +466,7 @@ allTables =
     SELECT
       n.nspname AS table_schema,
       c.relname AS table_name,
-      NULL AS table_description,
+      d.description AS table_description,
```
steve-chavez, line 469: This was the issue, the description was just a NULL.

test/Feature/IgnoreAclOpenApiSpec.hs
```diff
@@ -1,65 +0,0 @@
-module Feature.IgnorePrivOpenApiSpec where
```
steve-chavez, line 1: Renamed this file on https://github.com/PostgREST/postgrest/pull/1882. Forgot to delete it :man_facepalming:.

laurenceisla:
> I've tested it on the private big schema(327 tables/views) and there's no apparent change in query speed.

That was my main concern in this case. If it doesn't affect the performance, then looks good to me.

Issue #1873, opened 2021-06-15
prepared statement already exists while connecting to DB or executing requests


Uhta:
### Environment

* PostgreSQL version: 11
* PostgREST version: latest for 06/21
* Operating system: Windows 10

### Description of issue

Expected behaviour: no errors while connecting or executing a request

Actual behaviour: Error 42P05 "prepared statement \"0\" already exists"

### Steps to reproduce:
Install postgREST as docs say
Try to connect from cmd

steve-chavez:
Hey @Uhta,

Try adding `db-prepared-statements=false` to the config. If that doesn't work please try a previous version: https://github.com/PostgREST/postgrest/releases/tag/v7.0.1

Also, are you using a connection pooler? This looks similar to the error in https://github.com/PostgREST/postgrest/issues/1011.

Uhta:
Hello. I don't use a connection pooler.

Is it possible to use db-prepared-statment as an environmental variable? I want to use it in docker. 

I also added the line ```PREPARED_THRESHOLD = "0" ``` to config file and it seems to work. Any chance to use this as an environmental variable as well?

steve-chavez:
> Is it possible to use db-prepared-statment as an environmental variable? I want to use it in docker.

Yes, use it like `PGRST_DB_PREPARED_STATEMENTS=true`.

> I also added the line PREPARED_THRESHOLD = "0" to config file and it seems to work

:thinking: Weird, we don't have such a config option.

laurenceisla:
Transaction pooling is now compatible after commit db95bd1c3733eb979f4a65b3df0eaa16cadb4df2.

PR #1888, opened 2021-07-06
feat: add compatibility with connection poolers on transaction mode 

laurenceisla:
WIP

With the latest update in [`hasql-transaction`](https://github.com/nikita-volkov/hasql-transaction), now in version 1.0.1, there's now an option to not use Prepared Statements and be fully compatible with transaction pooling as it was mentioned in https://github.com/PostgREST/postgrest/pull/1633#issuecomment-718911934.

* [x] Update `hasql-transaction` to new version
* [x] Use `transaction` and `unpreparedTransaction` from the `hasql-transaction` library depending on the `db-prepared-statements` option
* [x] Add a hint to the 'prepared statement "0" already exists' error message (https://github.com/PostgREST/postgrest/issues/1011 and https://github.com/PostgREST/postgrest/issues/1873)
* [x] Add tests
* [x] Fix haskell dependencies for builds and continuous integration platforms


postgrest.cabal
```diff
@@ -87,7 +87,7 @@ library
                     , hasql-dynamic-statements  == 0.3.1
                     , hasql-notifications       >= 0.1 && < 0.2
                     , hasql-pool                >= 0.5 && < 0.6
-                    , hasql-transaction         >= 0.7.2 && < 1.1
+                    , hasql-transaction         >= 1.0.1
```
steve-chavez, line 90: Needs an upper bound:
```suggestion
                    , hasql-transaction         >= 1.0.1 && < 1.1
```

src/PostgREST/App.hs
```diff
@@ -190,14 +190,15 @@ postgrestResponse conf maybeDbStructure jsonDbS pgVer pool time req = do
...
 
-  runDbHandler pool (txMode apiRequest) jwtClaims .
+  runDbHandler pool (txMode apiRequest) jwtClaims (configDbPreparedStatements conf) .
     Middleware.optionalRollback conf apiRequest $
       Middleware.runPgLocals conf jwtClaims handleReq apiRequest jsonDbS
 
-runDbHandler :: SQL.Pool -> SQL.Mode -> Auth.JWTClaims -> DbHandler a -> Handler IO a
-runDbHandler pool mode jwtClaims handler = do
+runDbHandler :: SQL.Pool -> SQL.Mode -> Auth.JWTClaims -> Bool -> DbHandler a -> Handler IO a
+runDbHandler pool mode jwtClaims preparedStatements handler = do
```
steve-chavez, line 198: Hm, I think just `prepared` is good:
```suggestion
runDbHandler pool mode jwtClaims prepared handler = do
```

src/PostgREST/Error.hs
```diff
@@ -228,7 +228,19 @@ checkIsFatal (PgError _ (P.ConnectionError e))
...
+-- Check for a "prepared statement already exists" (Code 42P05: duplicate_prepared_statement).
+-- This would mean that a connection pooler in transaction mode is being used
+-- and incompatible features such as prepared statements or notification channels are enabled in the PostgREST configuration.
+checkIsFatal (PgError _ (P.SessionError (H.QueryError _ _ (H.ResultError (H.ServerError "42P05" _ _ _))))) =
+  Just $
+    "Hint: If you are using connection poolers in transaction mode,"
+    <> "try setting db-prepared-statements and db-channel-enabled to false."
+-- Check for "transaction blocks not allowed in statement pooling mode" errors (Code 08P01: protocol_violation).
+-- This would mean that a connection pooler in statement mode is being used which is not supported in PostgREST
+checkIsFatal (PgError _ (P.SessionError (H.QueryError _ _ (H.ResultError (H.ServerError "08P01" _ _ _))))) =
```
steve-chavez, line 242: @laurenceisla The `08P01` error code is looking too generic, I'm not sure if it could happen in other cases and then we give a misleading error. How about checking for the error message instead?

Does the pooler in statement mode give the exact "transaction blocks not allowed in statement pooling mode" error?

src/PostgREST/Error.hs
```diff
@@ -228,7 +228,19 @@ checkIsFatal (PgError _ (P.ConnectionError e))
   where isAuthFailureMessage = "FATAL:  password authentication failed" `isPrefixOf` toS failureMessage
         failureMessage = fromMaybe mempty e
 -- Chek for a syntax error(42601 is the pg code). This would mean the error is on our part somehow, so we treat it as fatal.
-checkIsFatal (PgError _ (P.SessionError (H.QueryError _ _ (H.ResultError (H.ServerError "42601" e _ _))))) = Just $ toS e
+checkIsFatal (PgError _ (P.SessionError (H.QueryError _ _ (H.ResultError (H.ServerError "42601" _ _ _))))) =
+  Just "Hint: This is probably a bug in PostgREST, please report it at https://github.com/PostgREST/postgrest/issues"
+-- Check for a "prepared statement already exists" (Code 42P05: duplicate_prepared_statement).
+-- This would mean that a connection pooler in transaction mode is being used
+-- and incompatible features such as prepared statements or notification channels are enabled in the PostgREST configuration.
```
steve-chavez, line 235: Hm, when the channel is enabled, does that give the duplicate prepared stmt error? If not, then this comment is not precise. Do we get a particular error for the channel when connecting to the tx pooler?

laurenceisla: You're right, the notification channel with transaction pooling doesn't give an error. I'll fix the comment and the hint.

src/PostgREST/Error.hs
laurenceisla: > Does the pooler in statement mode give the exact "transaction blocks not allowed in statement pooling mode" error?

Yes, it does. I added the message validation too.

monacoremo:
Rebased on main, should all be green now (at least everything built locally with nix and stack)

test/fixtures/schema.sql
```diff
@@ -2041,9 +2041,12 @@ returns setof v2.parents as $$
...
 $$ language sql;
 
--- Only used for manually testing creating prepared statements
-create view prepared_statements as
-select * from pg_catalog.pg_prepared_statements;
+-- Used to test if prepared statements are used
+create function uses_prepared_statements() returns bool as $$
+    select c > 0 from (
+      select count(name) as c from pg_catalog.pg_prepared_statements
+    ) ps;
```
steve-chavez, line 2048: I think this can be shortened to:
```suggestion
  select count(name) > 0 from pg_catalog.pg_prepared_statements;
```

Issue #1820, opened 2021-04-17
io tests fail on pg 9.5


wolfgangwalther:
I just noticed that our io tests currently fail on pg 9.5.

`postgrest-with-postgresql-9.5 postgrest-test-io` returns:

```
test/io-tests/test_io.py::test_read_db_setting FAILED
test/io-tests/test_io.py::test_db_schema_notify_reload FAILED
test/io-tests/test_io.py::test_max_rows_reload FAILED
test/io-tests/test_io.py::test_max_rows_notify_reload FAILED
test/io-tests/test_io.py::test_invalid_role_claim_key_notify_reload FAILED
```

We are not running the io tests against all PG versions in CI right now - that's something we should maybe do.

PG 9.5 had its final relese February 11, 2021, and is not supported upstream anymore - but we still do. My idea was to remove 9.5 support after our next release.

In that case, we'd need to fix those tests now, to make sure everything still works with 9.5. However, I don't know whether it's worth it - should we just drop PG 9.5 right away? @steve-chavez WDYT?

steve-chavez:
This happens because `current_setting(name, bool)` is not available on [pg 9.5](https://www.postgresql.org/docs/9.5/functions-admin.html). Only `current_setting(name)` is available there.

Those python errors are `PostgrestTimedOut` because they keep trying to request the root endpoint until it responds with 200 but it never does(keeps 404ing). pre-request also has the same issue.

https://github.com/PostgREST/postgrest/blob/915d568257b00f66efe6815b00fc8bc93ec9ce3e/test/fixtures/roles.sql#L18-L20

https://github.com/PostgREST/postgrest/blob/915d568257b00f66efe6815b00fc8bc93ec9ce3e/test/fixtures/schema.sql#L1633-L1655

https://github.com/PostgREST/postgrest/blob/915d568257b00f66efe6815b00fc8bc93ec9ce3e/test/fixtures/schema.sql#L1826-L1831

steve-chavez:
I think we could fix this by just using the `current_setting(name)` variant. Not sure if it's worth it though, if we're going to drop 9.5 anyway.

wolfgangwalther:
So this is just a problem with our test-cases, but if a user on 9.5 would use `current_setting` correctly, they'd have no problem?

Then I suggest we just leave as-is, drop 9.5 after the next release - and then add the io tests with all pg versions into CI.

steve-chavez:
[...]

PR #1896, opened 2021-07-14
Update nixpkgs, static-haskell-nix and use GHC 8.10.4

monacoremo:
Time to upgrade nixpkgs again. @nh2 did amazing work that makes this possible, thanks a million! 

* [x] Upgrade `nixpkgs` to a version that includes https://github.com/NixOS/nixpkgs/pull/129289`
* [x] Fix linting errors from new hlint version
* [x] Fix io-tests after change in the `jwt` Python module API
* [x] Revive Postgres 9.5 as it dropped out of support and was removed from nixpkgs
* [x] Upgrade `static-haskell-nix`, leveraging https://github.com/nh2/static-haskell-nix/pull/98
* [x] Set our Nix setup to use GHC 8.10.4 (`hsie` to follow later due to changes in the GHC API)
* [x] Upgrade `stackage` to LTS 18.2 with GHC 8.10.4


nh2:
Let me know if you encounter any problems.

monacoremo:
> Let me know if you encounter any problems.

Thank you @nh2! Seems to work perfectly. Only slight hiccup: The statically built executable was no longer stripped by default, which was easily fixed with a `lib.doStrip`. Are you aware whether that was an intended change somewhere in the Nix Haskell packages or static-haskell-nix?

nix/overlays/postgresql-legacy.nix
```diff
@@ -0,0 +1,20 @@
+self: super:
+# Overlay that adds legacy versions of PostgreSQL that are supported by
+# PostgREST.
+{
+  # PostgreSQL 9.5 was removed from Nixpkgs with
+  # https://github.com/NixOS/nixpkgs/commit/72ab382fb6b729b0d654f2c03f5eb25b39f11fbb
+  # We pin its parent commit to get the last version that was available.
+  postgresql_9_5 =
```
wolfgangwalther, line 8: Not sure if we need that - we agreed in #1820 to remove 9.5 soon anyway.

This boils down to whether we'll release v8 before merging this here or after.

@steve-chavez WDYT? How close are we do getting v8 ready docs-wise?

test/io-tests/test_io.py
```diff
@@ -207,7 +207,7 @@ def authheader(token):
 
 def jwtauthheader(claim, secret):
     "Authorization header with signed JWT."
-    return authheader(jwt.encode(claim, secret).decode("utf-8"))
+    return authheader(jwt.encode(claim, secret))
```
wolfgangwalther, line 210: Is this change related to the other changes?

wolfgangwalther (approved):
I didn't check the changes to the patches for other nix packages - those are over my head ;).

test/io-tests/test_io.py
monacoremo: Yes, the API of the Python jwt module changed with the newer version in the updated nixpkgs, returning a string instead of bytes: https://pyjwt.readthedocs.io/en/latest/changelog.html#jwt-encode-return-type

nix/overlays/postgresql-legacy.nix
monacoremo: Suggest that we merge this before, as we'll bake the latest 13.3 Postgres bindings into the static executable that way (not that there is anything critical there, but can't hurt: https://www.postgresql.org/docs/release/13.3/).

In any case, I would also keep this file around, with commented out content if it's not needed right now - we'll need it now and again later.

steve-chavez: Hey guys, good to see you back!

> This boils down to whether we'll release v8 before merging this here or after.

I don't see a problem with removing 9.5 after v8, so we can merge this.

> How close are we do getting v8 ready docs-wise?

@wolfgangwalther There's still work to be done in the docs. Me and @laurenceisla thought it'd be better to tag/release v8 with just the CHANGELOG in the github release(next week tops, after finishing the last breaking change https://github.com/PostgREST/postgrest/issues/1643) - and once that's done we can take a bit more time with finishing the v8 docs.

steve-chavez (approved):

nh2:
> The statically built executable was no longer stripped by default, which was easily fixed with a `lib.doStrip`. Are you aware whether that was an intended change somewhere in the Nix Haskell packages or static-haskell-nix?

@monacoremo Yes, that is this issue that made `strip` 300x on `.a` files:

https://github.com/nh2/static-haskell-nix/pull/98/commits/da452685fc7c5742c45cec3386631782884fa006

So I enabled `dontStrip` for all `.a` files to not make the build take insanely long. Your approach to strip your final executable as desired is correct.

I have already landed a fix for the 300x slowdown on nixpkgs `staging`, but I'm considering to keep the `dontStrip = true;` the default, so that people can decide at the very end whether they prefer small executable sizes, increased debuggability by keeping debug symbols. Still on the fence which way to go, as I haven't tried GHC's DWARF support for a while.

nh2:
@monacoremo I've pushed another update that should make `doStrip` unneccessary:

https://github.com/nh2/static-haskell-nix/pull/98/commits/cdc1a9a36add24fac3b98e32b03e0e3c41bd17ee

But your approach here is also fine.

monacoremo:
> @monacoremo I've pushed another update that should make `doStrip` unneccessary:
> 
> [nh2/static-haskell-nix@cdc1a9a](https://github.com/nh2/static-haskell-nix/commit/cdc1a9a36add24fac3b98e32b03e0e3c41bd17ee)

Your fix worked perfectly @nh2, many thanks! Again, amazing work on fixing musl GHC with all those interlinked issues that cropped up.

PR #1897, opened 2021-07-15
Manually update changelog for v8 release

monacoremo:
We'll land https://github.com/PostgREST/postgrest/pull/1805 most likely after the v8 release, so I'm updating the changelog once more manually.

For reference, I'm using:

To get the commits since the last major change in CHANGELOG.md:

```
git log 11d62a8010..HEAD --format="%h: %s - %an" 
```

To get a list of closed PRs:

```
curl -H "Accept: application/vnd.github.v3+json" 'https://api.github.com/repos/postgrest/postgrest/pulls?state=closed&per_page=100&page=1' > PRs.json
```

To filter and format the PRs:

```python
import json

with open("PRs") as prsfile:
    prs = json.loads(prsfile.read())

for pr in prs:
    if pr["merged_at"]:
        print(f" * #{pr['number']}, {pr['title']} - @{pr['user']['login']}")
```

...and then sorting everything out manually


CHANGELOG.md
```diff
@@ -26,18 +42,41 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1162, Fix location header for POST request with select= without PK - @wolfgangwalther
  - #1585, Fix error messages on connection failure for localized postgres on Windows - @wolfgangwalther
  - #1636, Fix `application/octet-stream` appending `charset=utf-8` - @steve-chavez
- - #1615, Fix RPC return type handling and embedding for domains with composite base type - @wolfgangwalther
  - #1469, #1638 Fix overloading of functions with unnamed arguments - @wolfgangwalther
+ - #1560, Return 405 Method not Allowed for GET of volatile RPC instead of 500 - @wolfgangwalther
+ - #1608, Fix embedding through views with subqueries inside function calls - @wolfgangwalther
+ - #1631, perf: change Text queries to ByteString - @steve-chavez
+ - #1632, Fix embedding through views with subqueries inside function calls - @wolfgangwalther
```
steve-chavez, line 49: Dup line:
```suggestion
 - #1608, Fix embedding through views with subqueries inside function calls - @wolfgangwalther
 - #1631, perf: change Text queries to ByteString - @steve-chavez
```

steve-chavez (changes requested):
@monacoremo Thanks a lot for this! I've removed some entries because they're not "fixes" per se - they weren't bugs in the previous version but in the nightlies.

After that, it should be good to merge!

monacoremo:
Thank you @steve-chavez !

Observations from manually running this update:

 * Fixes and changes might concern new/unreleased features and need to be eliminated from the changelog, as you also noticed with the ones I had missed
 * Some PRs/issues are related and should be in one combined changelog entry
 * Some additions and fixes also introduce breaking changes and should probably be listed twice, once as addition/fix and once as change
 * Consistent wording in the changelog and making sure everything is comprehensible requires some editing, even if we are very careful with the commit messages
 * Our track record with having consistent commit messages was ok, but not perfect and we would have gaps and inconsistencies if we generated the changelog automatically

All this to say: Maybe we should stick to requiring manual changelog entries with each PR. 

It's simple, shouldn't be tangibly more effort than crafting a good commit message and allows for editing to get a high quality, comprehensible result. It would also maintain the ability to find the rekevant change commits with a git blame on the changelog.

WDYT? Also cc'ing @wolfgangwalther

monacoremo:
Slightly related: Seeing as catching up with the docs is quite painful, potentially delaying releases, I think we could reduce the burden a lot by requiring docs for each PR, if relevant. 

It would be easiest to ensure this if we had the docs in the same repo. That would also aleviate the changelog duplication issue. I know we already discussed this last year, but WDYT?

steve-chavez:
> It's simple, shouldn't be tangibly more effort than crafting a good commit message and allows for editing to get a high quality, comprehensible result.

Also want to add that our new contributors added the entries to the CHANGELOG without issue — see https://github.com/PostgREST/postgrest/pull/1855 and https://github.com/PostgREST/postgrest/pull/1854. So our [PR template](https://raw.githubusercontent.com/PostgREST/postgrest/main/.github/PULL_REQUEST_TEMPLATE.md) reminder was good enough.

> All this to say: Maybe we should stick to requiring manual changelog entries with each PR.

Yes, agree! At least until https://github.com/PostgREST/postgrest/pull/1805 is merged and the new workflow is proven to work, we should keep updating the CHANGELOG manually. Though thinking a bit more about this, as you mention there still would be work to be done to tailor the CHANGELOG, which would lead to more commits on the main branch and more work. So maybe the manual work will stay in the end, it's simpler.

> It would be easiest to ensure this if we had the docs in the same repo. I know we already discussed this last year, but WDYT?

Ah, me and Wolfgang also discussed this a while ago [here](https://github.com/PostgREST/postgrest-docs/pull/367#issuecomment-732380642) and we agreed that docs is in fact another project. For example, on https://github.com/PostgREST/postgrest-docs/issues/411 we're thinking about adding curl/JS snippets to the references in postgrest-docs, doing all that work here would cause bloat in number of open issues and commits. So I think we should still keep them separate.

> That would also aleviate the changelog duplication issue. 

Yes, this is still an issue https://github.com/PostgREST/postgrest-docs/issues/340, I think the solution here is to change the CHANGELOG location to be on the docs [here](https://github.com/PostgREST/postgrest-docs/tree/main/releases)(also a bit more tidy since it's not a big markdown file), we could link to that in our PR template.

monacoremo:
> Yes, this is still an issue PostgREST/postgrest-docs#340, I think the solution here is to change the CHANGELOG location to be on the docs here(also a bit more tidy since it's not a big markdown file), we could link to that in our PR template.

We should definitely have a single source of truth for the changelog! But I'm not sure requiring a separate PR on the -docs repo for all code PRs that make a change would be good, that's significant extra work for any potential contributor and would require a lot of coordination and checking in maintenance.

Thinking about it, I'm not sure there is a good DRY solution as long as there are two separate repos.

## Docs and changelogs in other projects

I just had a look of how other projects handle this:

### Docs I remember for very high quality:

* Postgres: https://github.com/postgres/postgres/tree/master/doc -> docs in tree
* Django: https://github.com/django/django/tree/main/docs  -> docs in tree
* CPython: https://github.com/python/cpython/tree/main/Doc  -> docs in tree
* Flask: https://github.com/pallets/flask/tree/main/docs  -> docs in tree

### Average docs:

* GHC: https://gitlab.haskell.org/ghc/ghc/-/tree/master/docs  -> docs in tree
* Tmux: https://github.com/tmux/tmux/blob/master/tmux.1  -> docs in tree
* Vim: https://github.com/vim/vim/tree/master/runtime/doc  -> docs in tree
* NixOS: https://github.com/NixOS/nixpkgs/tree/master/doc  -> docs in tree

...all have the docs directly in tree.

Trying to find opposite examples I ran:

```bash
curl 'https://raw.githubusercontent.com/EvanLi/Github-Ranking/master/Data/github-ranking-2021-07-17.csv' | grep "\-docs"
```

This yielded one relevant example: https://github.com/microsoft/vscode-docs - from what I can see that's because the code and docs are under different licenses and the docs are pretty low activity. There does not seem to be a changelog in tree and the commit history is a mess. The updates, e.g. https://code.visualstudio.com/updates/v1_58 are written with what seems significant manual effort.

## Docs in separate repo

> For example, on PostgREST/postgrest-docs#411 we're thinking about adding curl/JS snippets to the references in postgrest-docs, doing all that work here would cause bloat in number of open issues and commits. So I think we should still keep them separate.

To summarize my understanding of the pro's and con's of having the docs separate:

### Benefits

1. Dedicated attention to crafting high-quality docs
2. Docs-related issues/PRs separated from other ones
3. Reduced complexity in code repo

### Disadvantages

4. Extra effort to document changes - requires extra PR and coordination
5. Risk that documentation significantly falls behind code, becoming more difficult to catch up and delaying releases
6. No single source of truth for changelog (deriving it from the commit history is not viable as we noticed. We could automate the sync between repos, though) 

Is this accurate?

### Case for merging the -docs repo in the code one

I fully appreciate (1)! Merging the repos would take away some focus on purey the docs. On the other hand, it would allow us to ensure that the basic updates to docs are made as we go, leaving only some overall review and polishing to be done before releases. 

On (2), I think we could still maintain the separate repo for tracking issues, even if we move the docs into the code repo. In that case, we could tag pure docs PRs to keep them separate - although most docs changes should be within the code change PRs. It's also pretty easy to move issues between repos on GitHub, should they be created on the wrong one.

On (3), I think we could handle this well if the repos were merged. The docs would live completely separated in a `docs` directory. We could avoid duplicating any Nix setup and integrate the CI jobs, ensuring that any (nightly) release also has docs released.

On the disadvantages: If we merged the repos, i think we could alleviate all the listed ones.

As a note: We would merge the repos in a way that maintains the full history of the -docs repo.

This got pretty long... and don't need to decide now. But does that make sense @steve-chavez ?

monacoremo:
Out of curiosity I experimented with merging the repos and maintaining the histories.

Approach:

* In `postgrest-docs`, add a commit to move everything to the `docs` subdirectory, push to a branch `move-docs`
* In `postgrest`:
  * Add the postgrest-docs repo with the branch as a remote: `git remote add -f moveddocs git@github.com-monacoremo:monacoremo/postgrest-docs.git`
  * Merge but do not commit yet: `git merge -s ours moveddocs/move-docs --allow-unrelated-histories --no-commit`. This merges the histories, but does not actually add the files
  * Create the files: `git read-tree --prefix mergeddocs -u moveddocs/move-docs: ; mv mergeddocs/docs docs ; rmdir mergeddocs` (there must be a move elegant way with `-m`...)
  * Commit: `git commit -m 'merge postgrest-docs'`

Result: https://github.com/monacoremo/postgrest/tree/move-docs

Observations:
* All commit hashes are preserved
* Git blame works as expected as it follows renamed files: https://github.com/monacoremo/postgrest/blame/move-docs/docs/README.md
* Potential issue: only history for the docs files that is visible at first is the commit that moves them (but git blame works)
* Would need a bit of CI integration to complete the merge

I'm not saying we need to do this! Just checking whether it's feasible :-)

steve-chavez:
> Docs I remember for very high quality(single repo):

One counter-example for this would be the [godot-docs](https://github.com/godotengine/godot-docs), which I consider high-quality(with a great community too), so working on a separate repo doesn't necessarily imply that we can't be high-quality.

> Postgres: https://github.com/postgres/postgres/tree/master/doc -> docs in tree

They do organize differently for doing docs - they have the pgsql-docs [mailing list](https://wiki.postgresql.org/wiki/Mailing_Lists#Overview_of_discussion_lists) - and I assume those mails are all related to docs.
While our workflow is more github-centric(for better or worse) and we only have different repos/issues to organize different work.

(I assume both vim and tmux have a similar mailing list workflow since they're mature projects)

On that topic, I've noticed that the number of Watchers(which I assume are potential contributors) tend to reduce when we have a lot of refactor-related commit activity. I assume this number would decrease more if we have a lot of docs-related activity(say "Fix typo").

A more case in point example, sometime ago I tried to be more involved on [nixpkgs](https://github.com/NixOS/nixpkgs/), so I did a "Watch" on that project and then I was swarmed with lots of notifications of PRs/issues I couldn't help with - so I had to desist and "Unwatch"(just wanted to be involved in postgres/haskell tooling).

> But I'm not sure requiring a separate PR on the -docs repo for all code PRs that make a change would be good, that's significant extra work for any potential contributor and would require a lot of coordination and checking in maintenance.

Even with the low-friction CHANGELOG, we got some reluctancy([here](https://github.com/PostgREST/postgrest/pull/1855#discussion_r637815531) and [here](https://github.com/PostgREST/postgrest/pull/1854#discussion_r636936730), no reply) to change it. I think that's fine, most contributors want to contribute code and docs are always a chore(which is the core team responsibility). So I wouldn't expect much more activity when merging the repos.

> Risk that documentation significantly falls behind code, becoming more difficult to catch up and delaying releases

Ah, but docs are not just about new code/features though, there are tutorials/explanations/how-tos that still need to be added, which take considerable work outside a feature PR.

Many things contributed to the delayed release, I think at first I wanted it to have more PostGIS features on it, then more perf, then I lost track, and now the CHANGELOG is [pretty big](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#unreleased) :open_mouth:. There definitely needs to be a limit for how much time or number of features one release carries, maybe we should commit to a release every 6 months, at least.

> No single source of truth for changelog (deriving it from the commit history is not viable as we noticed. We could automate the sync between repos, though)

This would be fixed if we move the CHANGELOG to the docs, needs to be manual anyway, as we agreed.

> Extra effort to document changes - requires extra PR and coordination

Hm, I think the effort of opening a PR on the docs is little compared to the actual work of actually doing the docs, check the latest PRs from Laurence: https://github.com/PostgREST/postgrest-docs/pull/404, https://github.com/PostgREST/postgrest-docs/pull/407 (lots of back and forth).

@laurenceisla WDYT? Do you think it'd be easier to do all the work in one repo?

(out of counterpoints for now :D)

wolfgangwalther:
> Observations from manually running this update:
> * Fixes and changes might concern new/unreleased features and need to be eliminated from the changelog, as you also noticed with the ones I had missed 
> * Some PRs/issues are related and should be in one combined changelog entry
> * Some additions and fixes also introduce breaking changes and should probably be listed twice, once as addition/fix and once as change
> * Consistent wording in the changelog and making sure everything is comprehensible requires some editing, even if we are very careful with the commit messages
> * Our track record with having consistent commit messages was ok, but not perfect and we would have gaps and inconsistencies if we generated the changelog automatically 
> 
> All this to say: Maybe we should stick to requiring manual changelog entries with each PR.

Agree with the observations, but not with the conclusion. I still think we should try to keep the commit messages as the primary source of information for the changelog - in the end this makes us write much better commit messages, too.

I actually like the process of what you did in this MR - I envision the changelog tools to heavily support this workflow by partly automating things. The workflow would be roughly like this:
- nightly changelogs are created automatically from commit history on release. Those might not be perfect, but the goal is to be able to release quickly.
- all other changelogs would be committed as part of the version bump. So basically we'd create a PR, that bumps the version and adds the changelogs at the same time. We can still discuss/improve the changelog items in the PR. The first commit of that PR is created easily via changelog tools.
- Instead of working on the formatted changelog file, we'd work on the yaml file that the changelog tools create - and then keep the yaml files for each version in the repo, too. This would be our single source of truth for all changelogs.
- The actual changelogs in different places (repo file, release page, docs) would then all be created via templates from that yaml file.

So yeah: The changelog tools will not fully automate things, but help us be consistent instead.

--- 

> It would be easiest to ensure this if we had the docs in the same repo. That would also aleviate the changelog duplication issue. I know we already discussed this last year, but WDYT?

I'm still heavily +1 here.

The changelog duplication issue would not be entirely removed, I still think the changelog tools would help a lot. But being able (and forced) to add docs together with the feature would improve things a lot.

---

> Ah, me and Wolfgang also discussed this a while ago [here](https://github.com/PostgREST/postgrest-docs/pull/367#issuecomment-732380642) and we agreed that docs is in fact another project.

I went back and read it again - and in fact I agreed. I just don't know why. It seems to me like every argument against merging the repos is just a thing of implementing a workflow (CI / issue tags / ...) to handle both in the same repo. If we do that properly, I don't see any problems - but just the huge value of being able to force docs updates together with features/fixes.

---

> They do organize differently for doing docs - they have the pgsql-docs mailing list - and I assume those mails are all related to docs.
> While our workflow is more github-centric(for better or worse) and we only have different repos/issues to organize different work.

They still require all patches on the "-hackers" mailing list to add documentation right away together with the feature. It's also perfectly fine to put stuff on "-hackers" that is about the docs only and happens regularly. The "-docs" mailing list seems to be more of a relict of past times - I guess they would not set it up like that, if they started right now.

---

> On that topic, I've noticed that the number of Watchers(which I assume are potential contributors) tend to reduce when we have a lot of refactor-related commit activity. I assume this number would decrease more if we have a lot of docs-related activity(say "Fix typo").

Well, there's different levels of "watching". People only interested in new features/fixes should just watch "releases only" - and we can improve their experience a lot by:
- generating nightlies more frequently
- using auto-generated changelogs to allow us do that
- have docs in-sync with the latest nightlies - otherwise it's hard to make use of all the latest features. We can enforce this with merging the repos...

---

> and docs are always a chore(which is the core team responsibility).

I disagree.

---

> Hm, I think the effort of opening a PR on the docs is little compared to the actual work of actually doing the docs

Merging the repos would not be about reducing work. It would just be about forcing docs to be merged together with features/fixes. It's much less work to write docs at the time the feature is developed - and much more to write them later.

steve-chavez:
> Well, there's different levels of "watching". People only interested in new features/fixes should just watch "releases only" 

But there is no "only watch this folder" level, which was my point about nixpkgs - no "Watch" only the [postgres](https://github.com/NixOS/nixpkgs/tree/master/pkgs/servers/sql/postgresql) directory(because that's the thing I can help with, I don't want to just be notified on releases).

> I don't see any problems - but just the huge value of being able to force docs updates together with features/fixes.
> Merging the repos would not be about reducing work. It would just be about forcing docs to be merged together with features/fixes. It's much less work to write docs at the time the feature is developed - and much more to write them later.

Hm, I don't see why that cannot be done with the PR template idea I mentioned(link to the docs repo with more clear steps) - you'll do the docs after finishing the feature anyway.

> merging the repos is just a thing of implementing a workflow (CI / issue tags / ...)

Ah, that still seems more work than what we have right now. More issues on a single repo, more tags on a single repo - ensuring if the tag workflow we use for core(nightlies, releases) and for releasing docs(different versions) both can coexist in the same CI runs.

steve-chavez:
Maybe the whole idea of keeping separate repos will make more sense when `postgrest-contrib` lands and perhaps another repo/project(future ideas) that also needs docs. Because really postgrest-docs is not only about postgrest core functionality - there's even systemd/Nginx stuff there.

steve-chavez:
> The "-docs" mailing list seems to be more of a relict of past times - I guess they would not set it up like that, if they started right now.

Just a note here - the above looks false. I've been subscribed to the `pgsql-docs` mailing list for the past days and it seems pretty active(and focused) to me - I don't see any indication of it being a relic or that they want to deprecate it.

> They still require all patches on the "-hackers" mailing list to add documentation right away together with the feature. It's also perfectly fine to put stuff on "-hackers" that is about the docs only and happens regularly. 

Now that I've noticed, subscribing to the `pgsql-hackers` mailing list is similar to "Watching" the nixpkgs repo. There's too much activity, you'll get flooded with emails. I also subscribed to that list a while ago to try and learn about postgres internals, but it was no use with so much activity, I couldn't focus.

wolfgangwalther:
> Just a note here - the above looks false. I've been subscribed to the `pgsql-docs` mailing list for the past days and it seems pretty active(and focused) to me - I don't see any indication of it being a relic or that they want to deprecate it.

I didn't mean to say that `-docs` wasn't used. But there are discussions about docs issues on `-hackers`, too. So it's not clearly separated. Which makes it harder to track stuff, especially if we're thinking "GitHub issue tracker" instead of "mailing list".

I mentioned it, because of this idea:

> On (2), I think we could still maintain the separate repo for tracking issues, even if we move the docs into the code repo.

In that case, we'd have the discussions about docs split up between two repos. That's something I'd want to avoid.

So imho we should fully merge the two repos, including the issue tracker, if we went that way at all.

wolfgangwalther:
> Maybe the whole idea of keeping separate repos will make more sense when `postgrest-contrib` lands and perhaps another repo/project(future ideas) that also needs docs. Because really postgrest-docs is not only about postgrest core functionality - there's even systemd/Nginx stuff there.

Good point. Looking at PostgreSQL itself suggests, we should not even create another repo `postgrest-contrib`, but just add a `contrib` folder to the main repo. Would also make it a lot easier to keep optional stuff in-sync with the postgrest across breaking changes and simplify CI for contrib a lot.

PR #1920, opened 2021-08-19
Allow newer base, hspec, and retry

felixonmars:
Builds fine and all tests are passing here, with the new versions.


monacoremo:
Thank you for the PR @felixonmars! 

Just to note: As we pin all Haskell package versions with Stackage and nixpkgs, no new versions are actually used in the builds and tests here when we update the postgrest.cabal. Any potential issues would pop up when we update the Stack resolver and nixpkgs pin. Still, this is fine to merge.

PR #1926, opened 2021-08-23
Add 'PGRST' error code to differentiate from PostgreSQL errors

### LINKED ISSUE #1917: Differentiate between PostgREST and PostgreSQL errors in the error body
**Issue**

I think the error body that the user gets does not clearly disambiguate if they come from the PostgreSQL database or from PostgREST itself. One way to know the difference is that the `code` key and its value are not present for PostgREST errors.

**Proposals**

1. A source string could be prepended to the PostgREST error message, while leaving the PostgreSQL errors untouched: 
```json
{
  "message": "PostgREST: Could not find...",
  "hint": "...."
}
```
2. Take the unused `code` key value pairs and specify that it's a PostgREST error there, leaving the rest untouched:
```json
{
  "code": "PostgREST",
  "message": "Could not find...",
  "hint": "..."
}
```
Maybe use unique codes like `PostgREST101` or `PGRST101` to identify each error.

laurenceisla:
Closes #1917

Todo:

- [x] Add error codes for PostgREST errors
- [x] Modify test specs with the new error codes


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
  - #1783, Include partitioned tables into the schema cache. Allows embedding, UPSERT, INSERT with Location response, OPTIONS request and OpenAPI support for partitioned tables - @laurenceisla
  - #1878, Add Retry-After hint header when in recovery mode - @gautam1168
+ - #1917, Add 'PostgREST' error code to payload body to differentiate from PostgreSQL errors - @laurenceisla
```
steve-chavez, line 12: A bit more explicit:
```suggestion
 - #1917, Add `"code": "PostgREST"` to the error response body to differentiate PostgREST errors from PostgreSQL errors - @laurenceisla
```

steve-chavez:
True, the PGRST is not helpful because it contains "PG", which could be interpreted as a PostgreSQL error. 

> Or do the same with just code and add a prefix for hasql and postgrest to avoid collisions. No prefix = Postgresql.

I think there are libraries relying on the `code` contents and modifying it would cause a breaking change that we can avoid. 

> One way to achieve both would be to add another field, e.g. called source: postgresql|hasql|postgrest

I like this one better. I'd say just have a `source: PostgREST/PostgreSQL`, because for end users Hasql would be the same as PostgREST. Then we can put Hasql specific errors in the `code` part.

src/PostgREST/Error.hs
```diff
@@ -161,46 +192,52 @@ instance JSON.ToJSON PgError where
...
-    _         -> JSON.object [
+  toJSON (H.ResultError (H.ServerError c m d h)) = JSON.object [
         "code"    .= (toS c      :: Text),
         "message" .= (toS m      :: Text),
         "details" .= (fmap toS d :: Maybe Text),
         "hint"    .= (fmap toS h :: Maybe Text)]
 
   toJSON (H.ResultError (H.UnexpectedResult m)) = JSON.object [
-    "message" .= (m :: Text)]
+    "code"    .= ("PGUNX" :: Text),
```
steve-chavez, line 212: This is more of a internal error - it'd happen if we mess a query. Hm, not sure if it'd valuable to have a special code for it.

src/PostgREST/Error.hs
```diff
@@ -161,46 +192,52 @@ instance JSON.ToJSON PgError where
...
     "details" .= d,
     "hint"    .= (("Row number " <> show i) :: Text)]
   toJSON (H.ResultError (H.UnexpectedAmountOfRows i)) = JSON.object [
+    "code"    .= ("PGUAR" :: Text),
     "message" .= ("Unexpected amount of rows" :: Text),
-    "details" .= i]
+    "details" .= i,
+    "hint"    .= JSON.Null]
   toJSON (H.ClientError d) = JSON.object [
+    "code"    .= ("PGCLI" :: Text),
```
steve-chavez, line 237: The only valuable code to differentiate would be this one I think. Because it's shown on a connection error which is not a PostgREST error per se.

steve-chavez:
> One way to achieve both would be to add another field, e.g. called source: postgresql|hasql|postgrest

Hm, though having another field could break some libraries with strict typing for the error body. That was mostly the main idea for just having two additional codes: `PGRST`(all PostgREST errors) and `PGCON`(anything related to the connection). Which seems simpler imho, and then just document this difference.

steve-chavez:
> Or do the same with just code and add a prefix for hasql and postgrest to avoid collisions. No prefix = Postgresql. 

One problem with prefixing the code, with say `PostgREST: HasqlError`, is that it's not really a "code" - not a [five-character error](https://www.postgresql.org/docs/13/errcodes-appendix.html). Although maybe that doesn't matter for this case.

laurenceisla:
> One way to achieve both would be to add another field, e.g. called `source: postgresql|hasql|postgrest` 
>>I like this one better. I'd say just have a source: PostgREST/PostgreSQL, because for end users Hasql would be the same as PostgREST.

@wolfgangwalther @steve-chavez I like this approach too. My initial suggestion was to add `"PostgREST: ..."` in the message field and not touch the code field at all. What Wolfgang suggested would be more akin to it without needing to modify the message.

> and then use `code` for a different number for each error.

I think we should arrive to an agreement on how to name these error codes, if at all. Otherwise, just the new `source` field without specifying the error code should be enough for the purpose of this PR (which is to know the error source).

steve-chavez:
One thing with adding the `source` field is that we lose some sense of **transparency** to PostgreSQL.

The [docs](https://postgrest--430.org.readthedocs.build/en/430/admin.html#error-source) for this PR made a lot of sense and it was great that we could map our error fields to [pg error fields(MESSAGE, DETAIL, HINT, ERRCODE)](https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html).

With the `source` field we lose this and the user would have to check our docs anyway to know the difference between `source: PostgreSQL/PostgREST`.

So I think we should just go back to `code: "PGRST"` for all of our errors and leave pg errors as is. The `PGRST` prefix is well known since we already use it for [singular or plural](https://postgrest.org/en/latest/api.html#singular-or-plural).

@wolfgangwalther @laurenceisla WDYT?

wolfgangwalther:
> So I think we should just go back to `code: "PGRST"` for all of our errors and leave pg errors as is. The `PGRST` prefix is well known since we already use it for [singular or plural](https://postgrest.org/en/latest/api.html#singular-or-plural).

Well, let's just make it a code prefix then and give all our errors a `PGRST01`, `PGRST02`, ... code? This way we can easily assemble a FAQ-style list of error codes with hints on how to solve them later on. We can probably even add a URI to the error message as hint or something, because those codes would allow static URIs to our docs?

src/PostgREST/Error.hs
```diff
@@ -52,6 +52,16 @@ class (JSON.ToJSON a) => PgrstError a where
...
 
+data ErrorCodePrefix
+  = ApiRequestErrorPrefix Text
+  | HasqlErrorPrefix      Text
+  | GeneralErrorPrefix    Text
+
+instance JSON.ToJSON ErrorCodePrefix where
+  toJSON (GeneralErrorPrefix    c) = JSON.toJSON ("PGRST0" <> c)
+  toJSON (ApiRequestErrorPrefix c) = JSON.toJSON ("PGRST1" <> c)
+  toJSON (HasqlErrorPrefix      c) = JSON.toJSON ("PGRST2" <> c)
```
laurenceisla, line 63: I found convenient to add a prefix for each group of errors that are defined in the code. This is to keep an order in the file and to identify the groups more easily. If no prefix was present, then the sequence of error codes would "jump" from group to group; e. g. if the last `Error` has a `PGRST30` code, and the last `ApiRequestError` has a `PGRST10` then the next `ApiRequestError` added would be `PGRST31`. 

The `GerenalErrorPrefix` corresponds to the `Error` data type and has a prefix of `0` due to being the errors that are not "grouped" by default and has `ApiRequestError` and `PgError` (Hasql errors) as constructors.

steve-chavez: I think that might be useful and it's similar to how pg classifies their errors: https://www.postgresql.org/docs/13/errcodes-appendix.html#ERRCODES-TABLE

So one group could be for JWT related errors, another for API errors, another for connection errors and so on.

src/PostgREST/Error.hs
```diff
@@ -302,27 +349,53 @@ instance PgrstError Error where
...
+    "code"    .= GeneralErrorPrefix "08",
+    "message" .= ("Server lacks JWT secret" :: Text),
+    "details" .= JSON.Null,
+    "hint"    .= JSON.Null]
   toJSON (JwtTokenInvalid message) = JSON.object [
-    "message" .= (message :: Text)]
+    "code"    .= GeneralErrorPrefix "09",
+    "message" .= (message :: Text),
+    "details" .= JSON.Null,
+    "hint"    .= JSON.Null]
```
steve-chavez, line 398: I think JWT errors should have their own group/prefix.

src/PostgREST/Error.hs
```diff
@@ -161,46 +204,52 @@ instance JSON.ToJSON PgError where
...
     "details" .= d,
     "hint"    .= (("Row number " <> show i) :: Text)]
   toJSON (H.ResultError (H.UnexpectedAmountOfRows i)) = JSON.object [
+    "code"    .= HasqlErrorPrefix "06",
     "message" .= ("Unexpected amount of rows" :: Text),
-    "details" .= i]
+    "details" .= i,
+    "hint"    .= JSON.Null]
   toJSON (H.ClientError d) = JSON.object [
+    "code"    .= HasqlErrorPrefix "07",
```
steve-chavez, line 249: As we discussed on https://github.com/PostgREST/postgrest-docs/pull/430#discussion_r696189225, this type of error is a special one that refers to the connection. I think we also need a special prefix for this one. Ideally, this would be a `PGRST000` so it's clearly identifiable.

laurenceisla: Regrouped the errors and this one would be the `PGRST001` error after the Connection Error from the Hasql Pool.

steve-chavez:
> We can probably even add a URI to the error message as hint or something, because those codes would allow static URIs to our docs?

@wolfgangwalther This is a great idea that Laurence is currently working on.

Also, regarding a URL on the error body, I've just seen the [Problem Details for HTTP APIs RFC](https://datatracker.ietf.org/doc/html/rfc7807) that presents a `type` field for this. Maybe we can adopt that.

steve-chavez:
The RFC mentions that:

> Note that both "type" and "instance" accept relative URIs; this means
> that they must be resolved relative to the document's base URI, as
> per [RFC3986], Section 5.

So the above idea about specifying `type` wouldn't be complete/correct unless we allowed to configure the base URI for the errors. That can be left for another PR.

wolfgangwalther:
> So the above idea about specifying `type` wouldn't be complete/correct unless we allowed to configure the base URI for the errors. That can be left for another PR.

I don't think we need any base URI to configure here. The URIs will all point to PostgREST docs, no? That will be an absolute URI in all cases.

Note, that the RFC does not say that URIs "must" be relative. It merely says, relative URIs are accepted, too, and if used, they must resolve in the matter described. There are some examples in the RFC that use absolute URIs and it also has this:

>  New problem type definitions MUST document:
> 
>   1.  a type URI (typically, with the "http" or "https" scheme),

steve-chavez:
> The URIs will all point to PostgREST docs, no? That will be an absolute URI in all cases.

Hm, I don't think it would be right to do that. I assume that external services([example](https://pro.tzkt.io/)) or open data initiatives([example](https://www.lobbyview.org/data-download/api)) that use us for their API would like to keep their users on their own site, so they should be able to override the error base URI.

Also, I'm not sure if we should add the `type` field, OpenAPI also has a way to describe errors https://swagger.io/docs/specification/describing-responses/#status-codes

src/PostgREST/Error.hs
```diff
@@ -343,3 +410,81 @@ invalidTokenHeader m =
...
+  SchemaCacheErrorCode03 -> "203"
+
+  JWTErrorCode00         -> "300"
+  JWTErrorCode01         -> "301"
+
+  HasqlErrorCode00       -> "400"
+  HasqlErrorCode01       -> "401"
+  HasqlErrorCode02       -> "402"
+  HasqlErrorCode03       -> "403"
+  HasqlErrorCode04       -> "404"
```
laurenceisla, line 483: Codecov is giving a warning for these errors codes and their error instances, but I don't think they can be tested because they are caused by unexpected bugs. We might have to tank the score loss here.

src/PostgREST/Error.hs
```diff
@@ -54,74 +54,107 @@ class (JSON.ToJSON a) => PgrstError a where
...
+  = ParseRequestError Text Text
+  | ActionInappropriate
   | InvalidBody ByteString
-  | ParseRequestError Text Text
-  | NoRelBetween Text Text Text
-  | AmbiguousRelBetween Text Text [Relationship]
-  | AmbiguousRpc [ProcDescription]
-  | NoRpc Text Text [Text] Bool ContentType Bool
+  | InvalidRange
+  | UnsupportedVerb -- Unreachable?
```
laurenceisla, line 63: I think this error is actually unreachable as mentioned in the comment. If the verb is not supported it would throw a `404 Not Found` before reaching the [`mutateRequest`](https://github.com/PostgREST/postgrest/blob/c858d158aa587510d8cf57373319f962514888ef/src/PostgREST/Request/DbRequestBuilder.hs#L312-L347) function. Doing this to the function, although less transparent, would omit using this error and give the same result (tests pass):

```haskell
-- mutateRequest function

    -- Change this:
    ActionDelete -> Delete qi <$> combinedLogic <*> pure returnings
    _            -> Left UnsupportedVerb

    -- Into this:
    _ -> Delete qi <$> combinedLogic <*> pure returnings
```

wolfgangwalther: I agree with the observation that we don't need this error. The suggested change, however, I don't agree with. This is really making it much less transparent. The proper solution would be to create an `ActionWrite WriteMethod` data constructor very similar to `ActionInvoke InvokeMethod`. This can the be passed from `writeQuery` in `App.hs` to `mutateRequest` - removing the impossible case.

wolfgangwalther:
> > The URIs will all point to PostgREST docs, no? That will be an absolute URI in all cases.
> 
> Hm, I don't think it would be right to do that. I assume that external services([example](https://pro.tzkt.io/)) or open data initiatives([example](https://www.lobbyview.org/data-download/api)) that use us for their API would like to keep their users on their own site, so they should be able to override the error base URI.

Maybe we can add a base URI for that later, but I don't think we need it right away. A base URI would require to have those external services have the same general URI structure, where just the base can be replaced.

A more flexible approach would be to rewrite those error responses via nginx `sub_filter` or `body_filter_by_lua` to their own URIs - and that's something that works better when we have a `https://postgrest.org/...` prefix there, that they can match.

> Also, I'm not sure if we should add the `type` field, OpenAPI also has a way to describe errors https://swagger.io/docs/specification/describing-responses/#status-codes

If we indeed move the OpenAPI output to "contrib", we shouldn't assume we have OpenAPI at all in the main app. I think linking to our docs via the type field (or any other field, `type` just happens to be in some kind of standard, that we can fall back to) is very valuable.

CHANGELOG.md
```diff
@@ -50,6 +50,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
  - #1783, Include partitioned tables into the schema cache. Allows embedding, UPSERT, INSERT with Location response, OPTIONS request and OpenAPI support for partitioned tables - @laurenceisla
  - #1878, Add Retry-After hint header when in recovery mode - @gautam1168
+ - #1917, Add error codes with the `"PGRST"` prefix to the error response body to differentiate PostgREST errors from PostgreSQL errors and include the `detail` and `hint` fields with `null` values if they are missing from the error response body. - @laurenceisla
```
steve-chavez, line 53: This should be split into two CHANGELOG entries for more clarity.

```suggestion
 - #1917, Add error codes with the `"PGRST"` prefix to the error response body to differentiate PostgREST errors from PostgreSQL errors - @laurenceisla
 - #1917, Normalize the error messages by always having the `detail` and `hint` error fields as `null - @laurenceisla
```

src/PostgREST/Error.hs
```diff
@@ -161,46 +195,52 @@ instance JSON.ToJSON PgError where
...
   toJSON (SQL.QueryError _ _ e) = JSON.toJSON e
 
 instance JSON.ToJSON SQL.CommandError where
-  toJSON (SQL.ResultError (SQL.ServerError c m d h)) = case BS.unpack c of
-    'P':'T':_ -> JSON.object [
-        "details" .= fmap T.decodeUtf8 d,
-        "hint"    .= fmap T.decodeUtf8 h]
-
-    _         -> JSON.object [
+  toJSON (SQL.ResultError (SQL.ServerError c m d h)) = JSON.object [
```
steve-chavez, line 208: So we are fixing https://github.com/PostgREST/postgrest/issues/2176 as well right? This should be added to the CHANGELOG.

CHANGELOG.md
laurenceisla: I modified it slightly to clarify when they will have the null value.

CHANGELOG.md
```diff
@@ -61,6 +60,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + PostgreSQL versions below 14 can opt in to the new JSON GUCs by setting the `db-use-legacy-gucs` config option to false (true by default)
  - #1988, Allow specifying `unknown` for the `is` operator - @steve-chavez
  - #2031, Improve error message for ambiguous embedding and add a relevant hint that includes unambiguous embedding suggestions - @laurenceisla
+ - #1917, Add error codes with the `"PGRST"` prefix to the error response body to differentiate PostgREST errors from PostgreSQL errors - @laurenceisla
+ - #1917, Normalize the error response body by always having the `detail` and `hint` error fields with a `null` value if they are empty - @laurenceisla
+ - #2176, Errors raised with `SQLSTATE` now include the message and the code in the response body.
```
steve-chavez, line 65:
```suggestion
 - #2176, Errors raised with `SQLSTATE` now include the message and the code in the response body - @laurenceisla 
```

PR #1927, opened 2021-08-25
feat: RPC POST for function w/single unnamed param 

### LINKED ISSUE #1735: Allow a function with single unnamed parameter to be called with POST
### Use case

Some external services have webhooks that allow to POST json to an URL. This json doesn't come with a top level key, so our [Prefer: params=single-object](https://postgrest.org/en/v7.0.0/api.html#calling-functions-with-a-single-json-parameter) header is ideal here.

However, some services(like Postmark) don't allow setting HTTP headers for these webhooks. The workaround here is to use the proxy to map an URL query parameter to the header, but this is not convenient.

(Also, I can't find it right now, but there was an issue/gitter comment about an embedded device having the same limitation - crippled http client)

### Proposal

Allow a function with single unnamed parameter to be called with POST, without specifying `params=single-object`:

```sql
create or replace function bulk_email_processing(json) returns void as $_$
-- ...
$_$ language sql;
```

```http
POST /rpc/bulk_email_processing

{"any": "json"}
```

This wouldn't work for GET, since the parameter is unnamed it can't be specified in the URL.

---

Right now we don't allow calling unnamed parameters on functions, so this wouldn't cause a breaking change.

steve-chavez:
Closes #1735. For POST on RPC, allows:

- **Passing a json object without using `Prefer: params=single-object`**. The function must be defined with a single unnamed `json` or `jsonb` param and `Content-Type: application/json` must be specified.

- **Uploading binary to a function**. The function must be defined with a single unnamed `bytea` param and `Content-Type: application/octet-stream` must be specified.

- **Uploading raw text to a function**. The function must be defined with a single unnamed `text` param and `Content-Type: text/plain` must be specified.

---

**BREAKING CHANGE for Overloaded Functions** If there's a function `my_func` having a single unnamed json param and other overloaded pairs(with any number of params), PostgREST won't be able to resolve a POST request to `my_func`. For solving this, you can name the unnamed json param.

```
my_func(json) -> my_func(prm json)
```


test/fixtures/schema.sql
```diff
@@ -2242,4 +2242,28 @@ A test for partitioned tables$$;
...
+  select $1;
+$$ language sql;
+
+create or replace function test.overloaded_unnamed_param(json) returns int as $$
+  select $1;
+$$ language sql;
+
+create or replace function test.overloaded_unnamed_param(x int, y int) returns int as $$
+  select x + y;
+$$ language sql;
```
steve-chavez, line 2269: Having a func with an unnamed json parameter plus any number of overloaded pairs

test/Feature/RpcSpec.hs
```diff
@@ -1059,3 +1060,76 @@ spec actualPgVersion =
...
+        it "will not be able to resolve when a single unnamed json parameter exists and other overloaded functions exist" $
+          request methodPost "/rpc/overloaded_unnamed_param" [("Content-Type", "application/json")]
+              [json|{"x": 1, "y": 2}|]
+            `shouldRespondWith`
+              [json| {
+                "hint":"Try changing the parameters' names or the function name so function overloading can be resolved",
+                "message":"Could not choose the best candidate function between: test.overloaded_unnamed_param( => json), test.overloaded_unnamed_param(x => integer, y => integer)"}|]
+              { matchStatus  = 300
+              , matchHeaders = [matchContentTypeJson]
+              }
```
steve-chavez, line 1135: PostgREST won't be able to resolve the function called with POST because we match the function's parameters with the json keys and the unnamed parameter basically absorbs all the json keys into one parameter.

steve-chavez (commented):
Explanation of the breaking change

test/fixtures/schema.sql
steve-chavez: Somehow it doesn't show where, but the above function is:

```sql
create or replace function test.overloaded_unnamed_param(json) returns int as $$
  select $1;
$$ language sql;
```

test/Feature/RpcSpec.hs
steve-chavez: Unfortunately I don't think there's a way to avoid this breaking change.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -480,37 +484,41 @@ rawContentTypes AppConfig{..} =
...
+        then null argumentsKeys
+      -- If the function is called with post and has a single unnamed parameter
+      -- it can be called depending on content type and the parameter type
+      else if isInvPost && length params == 1 && (ppName <$> headMay params) == Just mempty
+        then case headMay params of
+          Just prm | contentType == CTApplicationJSON -> ppType prm `elem` ["json", "jsonb"]
+                   | contentType == CTTextPlain       -> ppType prm == "text"
+                   | contentType == CTOctetStream     -> ppType prm == "bytea"
+                   | otherwise                        -> False
+          Nothing  -> False
```
steve-chavez, line 514: Match the function with the single unnamed parameter and content types.

It's necessary to only consider POST(`isInvPost`) because otherwise we'd also break GET requests that also have overloaded functions with a single unnamed json param. `contentType` isn't enough to determine we're in a POST because a GET can also send a `Content-Type` header despite having no body. Also, we default to `application/json` for all requests included GET.

src/PostgREST/Error.hs
```diff
@@ -99,11 +99,19 @@ instance JSON.ToJSON ApiRequestError where
...
-    "hint"    .= ("Overloaded functions with the same argument name but different types are not supported" :: Text),
-    "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [pgaName a <> " => " <> pgaType a | a <- pdArgs p] <> ")" | p <- procs])]
-  toJSON (NoRpc schema procName payloadKeys hasPreferSingleObject)  = JSON.object [
-    "hint"    .= ("If a new function was created in the database with this name and arguments, try reloading the schema cache." :: Text),
-    "message" .= ("Could not find the " <> schema <> "." <> procName <> (if hasPreferSingleObject then " function with a single json or jsonb argument" else "(" <> T.intercalate ", " payloadKeys <> ")" <> " function") <> " in the schema cache")]
+    "hint"    .= ("Try changing the parameters' names or the function name so function overloading can be resolved" :: Text),
+    "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [ppName a <> " => " <> ppType a | a <- pdParams p] <> ")" | p <- procs])]
```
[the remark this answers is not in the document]
steve-chavez: > I think it should be specified that the change should be done in the database, otherwise, at first glance, it may look like the change should be done on the query

True, I agree.

> Also, I think the old hint should be kept and moved to the "detail" field. WDYT?

The thing is that the "Overloaded functions with the same argument name but different types are not supported" error is no longer accurate now. This can be noted on the test I've added:

```haskell
it "will not be able to resolve when a single unnamed json parameter exists and other overloaded functions exist" $
  request methodPost "/rpc/overloaded_unnamed_param" [("Content-Type", "application/json")]
      [json|{"x": 1, "y": 2}|]
    `shouldRespondWith`
      [json| {
        "hint":"Try changing the parameters' names or the function name so function overloading can be resolved",
        "message":"Could not choose the best candidate function between: test.overloaded_unnamed_param( => json), test.overloaded_unnamed_param(x => integer, y => integer)"}|]
      { matchStatus  = 300
      }
```

The above is not related to having the same parameter name with different types.

steve-chavez:
stack on FreeBSD is failing with a really weird error:

```
/tmp/cirrus-ci-build/src/PostgREST/Request/Types.hs:44:44: error:
    Module ‘PostgREST.DbStructure.Proc’ does not export ‘ProcParam(..)’
   |
44 | import PostgREST.DbStructure.Proc         (ProcParam (..))
```

Which doesn't make sense because `PostgREST.DbStructure.Proc` does export `ProcParam`.

I'll merge now and revisit that one(if it keeps failing) later.

steve-chavez:
Same failure happening on master. Must be something related to stack. I see [Nix being supported on FreeBSD](https://wiki.freebsd.org/Nix), wonder if we can switch to it there.

PR #1928, opened 2021-08-25
fix: Detect default values for String types in OpenAPI output

### LINKED ISSUE #1871: OpenAPI description is missing the default value of postgres columns(?)
Hello, great work! I'm currently getting to know postgrest, and considering it for my project as a slim "application server" replacement.

### Environment

* PostgreSQL version: docker image 13
* PostgREST version: docker image v7.0.1

* Operating system: ubuntu 20.04 on WSL2

#### Expected behavior vs actual behavior
When a postgres column has a default value, I would wish that the OpenAPI description reflects that (which currently does not seem to be the case?).

#### Steps to reproduce
CREATE TABLE t (i int DEFAULT 42)

See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md : I would have expected the following:

> default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if type is string, then default can be "foo" but cannot be 1.

laurenceisla:
Closes #1871

Detects most string formats excluding "json" and "jsonb".
Arrays are now detected as an "array" type instead of "string".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  
 ### Fixed
 
- - #1871, OpenAPI fix: Show missing default values for String types and identify Array types as "array" instead of "string" - @laurenceisla
+ - #1871, Fix OpenAPI missing default values for String types and identify Array types as "array" instead of "string" - @laurenceisla
 
 ## [8.0.0] - 2021-07-25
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  
 ### Fixed
 
+ - #1871, OpenAPI fix: Show missing default values for String types and identify Array types as "array" instead of "string" - @laurenceisla
```
steve-chavez, line 15:
```suggestion
 - #1871, Fix OpenAPI missing default values for String types and identify Array types as "array" instead of "string" - @laurenceisla
```

Issue #1121, opened 2018-05-28
Feature request: support for filter/condition that is always false (ie. where false)


john-kelly:
First and foremost, **thank you very much for your time**

### Summary
I've discussed this issue in the past on gitter, however, I thought I'd make a more formal request for this functionality. In short, I'm in need of support for a filter/condition that always fails (ie. where false). This request is a matter of elegance/beauty + handling edge cases rather than desiring new functionality.

In sql terms, what I'm requesting is something along the lines of:
```sql
... where false
```

We already have support for:
```sql
... where true
```
through implicit means (specifying no filters).

### Context
I'm the author of [elm-postgrest](https://github.com/john-kelly/elm-postgrest/tree/dev), and I'm under the impression that this functionality would allow my API to be more elegant + simpler (and I assume the same will be true for other client libraries). Let me try to demonstrate (examples will be in elm, however, those familiar with haskell should feel at home).

Let's start by looking at the *current*/*desired* API for a readMany request in elm-postgrest:
```elm
readMany :
    Schema id attributes
    ->
        { select : Selection attributes a
        , where_ : Condition attributes
        , order : List (Order attributes)
        , limit : Maybe Int
        , offset : Maybe Int
        }
    -> Request (List a)
```

and now some example usage:
```elm
getArticles : Request List
getArticles =
    PG.readMany schema
        { select = selection
        , where_ = PG.true
        , order = []
        , offset = Nothing
        , limit = Nothing
        }
```

as you can see `PG.true` has the type:
```elm
PG.true : Condition attributes
```

In addition to `PG.true`, elm-postgrest also supports `PG.not`. This function has the type:

```elm
not : Condition attributes -> Condition attributes
```

And, by extension of having both `PG.true` and `PG.not`, in order for things to remain consistent, we must also have `PG.false`.

Now, to be clear, I understand that I could represent the `PG.true` case with something other than a "true" `Condition attributes`. For example, `readMany` could have the type:
```elm
readMany :
    Schema id attributes
    ->
        { select : Selection attributes a
        , where_ : Maybe (Condition attributes) <----- NOTE: api change
        , order : List (Order attributes)
        , limit : Maybe Int
        , offset : Maybe Int
        }
    -> Request (List a)
```

and example usage being:
```elm
getArticles : Request List
getArticles =
    PG.readMany schema
        { select = selection
        , where_ = Nothing <------ NOTE: api change
        , order = []
        , offset = Nothing
        , limit = Nothing
        }
```

[...]

PR #1929, opened 2021-08-28
Changes to allow limit 0

gautam1168:
Adds #1121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Enables calling a function with a single json parameter without using `Prefer: params=single-object`
    + Enables uploading bytea to a function with `Content-Type: application/octet-stream`
    + Enables uploading raw text to a function with `Content-Type: text/plain`
+ - #1121, Allow `limit=0` to get an empty result - @gautam1168 
 
 ### Fixed
 
--- a/src/PostgREST/Request/ApiRequest.hs
+++ b/src/PostgREST/Request/ApiRequest.hs
@@ -34,8 +34,7 @@ import Data.List                 (last, lookup, partition, union)
 import Data.List.NonEmpty        (head)
 import Data.Maybe                (fromJust)
 import Data.Ranged.Boundaries    (Boundary (..))
-import Data.Ranged.Ranges        (Range (..), emptyRange,
-                                  rangeIntersection)
+import Data.Ranged.Ranges        (Range (..), rangeIntersection)
 import Network.HTTP.Base         (urlEncodeVars)
 import Network.HTTP.Types.Header (hAuthorization, hCookie)
 import Network.HTTP.Types.URI    (parseQueryReplacePlus,
@@ -183,7 +182,7 @@ userApiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> Either A
 userApiRequest conf@AppConfig{..} dbStructure req reqBody
   | isJust profile && fromJust profile `notElem` configDbSchemas = Left $ UnacceptableSchema $ toList configDbSchemas
   | isTargetingProc && method `notElem` ["HEAD", "GET", "POST"] = Left ActionInappropriate
-  | topLevelRange == emptyRange = Left InvalidRange
+  | invalidLimit = Left InvalidRange
   | shouldParsePayload && isLeft payload = either (Left . InvalidBody . toS) witness payload
   | isLeft parsedColumns = either Left witness parsedColumns
   | otherwise = do
@@ -392,6 +391,20 @@ userApiRequest conf@AppConfig{..} dbStructure req reqBody
           l = fromMaybe 0 $ rangeLimit rl
           o = rangeOffset ro
   ranges = M.insert "limit" (rangeIntersection headerRange (fromMaybe allRange (M.lookup "limit" urlRange))) urlRange
+
+  limitFromQparams :: [(Text, Maybe ByteString)] -> [Maybe Integer]
+  limitFromQparams [] = []
+  limitFromQparams ((k, v):qs) = if isLimitQueryParameter then (readMaybe . toS =<< v):remainingLimits else remainingLimits
+    where isLimitQueryParameter = isJust v && endingIn ["limit"] k
+          remainingLimits = limitFromQparams qs
+
+  invalidLimit :: Bool
+  invalidLimit = foldl trueIfNegative False limits
+    where limits = limitFromQparams qParams
+          trueIfNegative = \ acc lim
+                              -> case lim of
+                                  Just n  -> (n < 0) || acc
+                                  Nothing -> acc
 
 {-|
   Find the best match from a list of content types accepted by the
--- a/test/Feature/RangeSpec.hs
+++ b/test/Feature/RangeSpec.hs
@@ -79,11 +79,6 @@ spec = do
             simpleStatus r `shouldBe` ok200
 
       context "of invalid range" $ do
-        it "fails with 416 for offside range" $
-          request methodPost  "/rpc/getitemrange"
-                  (rangeHdrs $ ByteRangeFromTo 1 0) emptyRange
-            `shouldRespondWith` 416
-
         it "refuses a range with nonzero start when there are no items" $
           request methodPost "/rpc/getitemrange"
                   (rangeHdrsWithCount $ ByteRangeFromTo 1 2) emptyRange
@@ -181,17 +176,38 @@ spec = do
             [json|[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}]|]
             { matchHeaders = ["Content-Range" <:> "0-14/*"] }
 
-      it "fails if limit equals 0" $
+      it "returns blank array when limit is 0" $
         get "/items?select=id&limit=0"
-          `shouldRespondWith` [json|{"message":"HTTP Range error"}|]
-          { matchStatus  = 416
+          `shouldRespondWith` [json|[]|]
+          { matchStatus  = 200
           , matchHeaders = [matchContentTypeJson]
           }
 
       it "fails if limit is negative" $
-        get "/items?select=id&limit=-1"
-          `shouldRespondWith` [json|{"message":"HTTP Range error"}|]
+        get "/items?select=id&limit=-1" `shouldRespondWith`
+          [json|{"message":"HTTP Range error"}|]
           { matchStatus  = 416
+          , matchHeaders = [matchContentTypeJson]
+          }
+
+      it "fails if limit is negative in an embedded resource" $
+        get "/clients?select=id,projects(id,tasks(id))&order=id.asc&limit=1&projects.order=id.asc&projects.limit=-1&projects.tasks.order=id.asc&projects.tasks.limit=1" `shouldRespondWith`
+          [json|{"message":"HTTP Range error"}|]
+          { matchStatus  = 416
+          , matchHeaders = [matchContentTypeJson]
+          }
+
+      it "fails if limit is negative in an embedded resource at any level" $
+        get "/clients?select=id,projects(id,tasks(id))&order=id.asc&limit=1&projects.order=id.asc&projects.limit=2&projects.tasks.order=id.asc&projects.tasks.limit=-1" `shouldRespondWith`
+          [json|{"message":"HTTP Range error"}|]
+          { matchStatus  = 416
+          , matchHeaders = [matchContentTypeJson]
+          }
+
+      it "treats an empty limit parameter as no limit parameter being passed" $
+        get "/items?select=id&limit=" `shouldRespondWith`
+          [json|[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}]|]
+          { matchStatus  = 200
           , matchHeaders = [matchContentTypeJson]
           }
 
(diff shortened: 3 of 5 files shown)


gautam1168:
@steve-chavez @monacoremo I have not cleaned this PR yet (lint checks, unit tests, changelog etc.), but I wanted to show you guys what changes I did to add this feature as they are spread out across 3 modules. Please let me know if it looks alright and I should clean it up for merging.

steve-chavez:
@gautam1168 I think the change is pretty minimal, overall it looks good. I'd say go ahead with the cleanup.

One the test failures caught my attention

```haskell
  3) Feature.RangeSpec, GET /items, with limit/offset parameters, fails if limit is negative
       status mismatch:
         expected: 416
         but got:  200
       body mismatch:
         expected: {"message":"HTTP Range error"}
         but got:  []

-- it's doing:
-- get "/items?select=id&limit=-1"
```

Weird that it accepts that the negative limit, in SQL that's also invalid:

```sql
select * from projects limit -1;
ERROR:  2201W: LIMIT must not be negative
```

gautam1168:
I will go through the unit tests and make sure that negative limit is not accepted.

gautam1168:
@steve-chavez I found the reason behind the failure of the unit test you pointed out and fixed it. The reason is that `NonnegRange` is defined as `Range Integer` so even if a negative number is passed to one of the boundaries of a range in `restrictRange` it passes through without throwing an error (and just constructing an `emptyRange`). Since after this change, an `emptyRange` is valid the response comes out with a status of 200 and a body of `[]`. To fix this, I have added another check in `ApiRequest.hs` to make sure that if a negative number is passed as the value of the `limit` queryParameter then the flow gets stopped there itself by considering that as an invalid limit (since even a negative number boundary will only result in an emptyRange if its allowed to flow down further into the pipeline). I have updated the PR with this change now and checked that the unit tests pass on my local.

I did comment out two of the unit tests that were expecting 416 for `emptyRange`'s. Should I change the assertion in those tests instead of commenting them out?

gautam1168:
Can you tell me what the style-check is trying to check? Its error message is not very useful and I don't know how to fix it.

steve-chavez:
Right, the error message in CI is not helpful.

> Run postgrest-style-check
> :100644 100644 0085583761e14b8e37d32a4b4ea431c7f8b1e3de 
> 0000000000000000000000000000000000000000 M	src/PostgREST/Request/ApiRequest.

It just says that ApiRequest is "unstyled".

Could you run `postgrest-style` and then commit? This should clear the error.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -389,6 +389,16 @@ userApiRequest conf@AppConfig{..} dbStructure req reqBody
...
   ranges = M.insert "limit" (rangeIntersection headerRange (fromMaybe allRange (M.lookup "limit" urlRange))) urlRange
 
+  limitFromQparams :: [(Text, Maybe ByteString)] -> Maybe Integer
+  limitFromQparams [] = Nothing
+  limitFromQparams ((k, v):qs) = if isLimitQueryParameter then readMaybe . toS =<< v else limitFromQparams qs
+    where isLimitQueryParameter = isJust v && endingIn ["limit"] k
+
+  invalidLimit :: Bool
+  invalidLimit = case limitFromQparams qParams of 
+      Just n -> n < 0
```
steve-chavez, line 399: Could you add some tests to ensure `?limit=-1` gives an error? Maybe for embedded resources as well(`?clients.limit=-1`)

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -183,7 +182,8 @@ userApiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> Either A
 userApiRequest conf@AppConfig{..} dbStructure req reqBody
   | isJust profile && fromJust profile `notElem` configDbSchemas = Left $ UnacceptableSchema $ toList configDbSchemas
   | isTargetingProc && method `notElem` ["HEAD", "GET", "POST"] = Left ActionInappropriate
-  | topLevelRange == emptyRange = Left InvalidRange
+  -- | topLevelRange == emptyRange = Left InvalidRange
```
steve-chavez, line 185: No commented out code please :pray:

test/Feature/RangeSpec.hs
```diff
@@ -79,10 +79,11 @@ spec = do
...
       context "of invalid range" $ do
-        it "fails with 416 for offside range" $
-          request methodPost  "/rpc/getitemrange"
-                  (rangeHdrs $ ByteRangeFromTo 1 0) emptyRange
-            `shouldRespondWith` 416
+        -- emptyRange is now supported
+        -- it "fails with 416 for offside range" $
+        --   request methodPost  "/rpc/getitemrange"
+        --           (rangeHdrs $ ByteRangeFromTo 1 0) emptyRange
+        --     `shouldRespondWith` 416
```
steve-chavez, line 86: Ditto

test/Feature/RangeSpec.hs
```diff
@@ -330,10 +331,11 @@ spec = do
...
       context "of invalid range" $ do
-        it "fails with 416 for offside range" $
-          request methodGet  "/items"
-                  (rangeHdrs $ ByteRangeFromTo 1 0) ""
-            `shouldRespondWith` 416
+        -- emptyRange is now supported
+        -- it "fails with 416 for offside range" $
+        --   request methodGet  "/items"
+        --           (rangeHdrs $ ByteRangeFromTo 1 0) ""
+        --     `shouldRespondWith` 416
```
steve-chavez, line 338: Ditto

src/PostgREST/Request/ApiRequest.hs
gautam1168: @steve-chavez 
There is actually a test for negative limit already in `RangeSpec.hs` :
```
it "fails if limit is negative" $
        get "/items?select=id&limit=-1"
          `shouldRespondWith` [json|{"message":"HTTP Range error"}|]
          { matchStatus  = 416
          }
```

I found this while I was trying to make it throw a 400 in case of negative limit and adding a test for it in QuerySpec.hs

I will leave it as 416 now, because I don't know what things downstream are depending on a 416 error when a negative limit is passed. But I'll add another test for embedded resources.

gautam1168:
@steve-chavez I have made all the changes you requested above:
- Removed commented out code.
- Added tests for negative limits at top level and embedded resources
- Fixed the postgrest-style

I do not get why the FreeBSD build is failing though. Its failing in a file that has no changes in my pull request and I have rebased my changes to latest upstream/master. The error says that `ProcParam` is not exported by `DbStructure.Proc` module. But I can see it is exported in the `Proc.hs` file. Is this some known thing in BSD?

test/Feature/RangeSpec.hs
gautam1168: removed

test/Feature/RangeSpec.hs
gautam1168: removed

src/PostgREST/Request/ApiRequest.hs
gautam1168: removed

steve-chavez:
> I do not get why the FreeBSD build is failing though.

Me neither :confused:, it started failing on https://github.com/PostgREST/postgrest/pull/1927#issuecomment-908771903. At any rate, the FreeBSD failure is unrelated to this PR.

> I have made all the changes you requested above

Cool! I'll give it another review.

test/Feature/RangeSpec.hs
```diff
@@ -181,20 +176,41 @@ spec = do
...
-          { matchStatus  = 416
+          `shouldRespondWith` [json|[]|]
+          { matchStatus  = 200
           , matchHeaders = [matchContentTypeJson]
           }
 
       it "fails if limit is negative" $
-        get "/items?select=id&limit=-1"
-          `shouldRespondWith` [json|{"message":"HTTP Range error"}|]
+        get "/clients?select=id,projects(id,tasks(id))&order=id.asc&limit=-1&projects.order=id.asc&projects.limit=1&projects.tasks.order=id.asc&projects.tasks.limit=1" `shouldRespondWith`
```
steve-chavez, line 187: @gautam1168 Sorry for the late review. This looks almost done now.

Q: Why do we need the embeds here? We already have some embeds tests below and seems this test would be clearer without them.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Enables calling a function with a single json parameter without using `Prefer: params=single-object`
    + Enables uploading bytea to a function with `Content-Type: application/octet-stream`
    + Enables uploading raw text to a function with `Content-Type: text/plain`
+ - #1121, Feature request: support for filter/condition that is always false (ie. where false) - @gautam1168
```
steve-chavez, line 16: More explicit:
```suggestion
 - #1121, Allow `limit=0` to get an empty result - @gautam1168 
```

steve-chavez:
The FreeBSD error was fixed on https://github.com/PostgREST/postgrest/pull/1939. If you rebase then the PR should be all green on CI.

test/Feature/RangeSpec.hs
gautam1168: Yes these are not needed. I just used the same endpoint for al lthe tests. But I will change this to the previous version.

gautam1168:
@steve-chavez I have rebased the PR and also made the change in CHANGELOG, `RangeSpec.hs` as you suggested. I have not removed any tests so I don't know why codecov/patch is failing. I'm not sure if it was failing before or not.

CHANGELOG.md
gautam1168: Added the change

test/Feature/RangeSpec.hs
gautam1168: Reverted the test to use the non-embedded resource for testing as it was before

laurenceisla:
@gautam1168 Nice job on the PR! I've noticed that this query `/clients?limit=2&offset=-2` returns an "HTTP Range error" on the previous version and in this PR it returns an empty array, which is OK, since it could be interpreted as a limit = 0. However, this query `/clients?limit=2&offset=-3` also returns an empty array in this PR, I think this should throw a range error instead as it could be interpreted as a negative limit or an invalid range.

laurenceisla (commented):
@gautam1168  I checked some more and I think you can avoid using the `limitFromQParams` function after modifying the Ranges involved. In order to make the Ranges accept 0 as their lowest upper limit (and not 1 as it was before), I tried the following changes and got only the limit=0 and Range 1-0 tests to break (the other changes need to be reverted to test this):

test/Feature/RangeSpec.hs
```diff
@@ -79,11 +79,6 @@ spec = do
             simpleStatus r `shouldBe` ok200
 
       context "of invalid range" $ do
-        it "fails with 416 for offside range" $
-          request methodPost  "/rpc/getitemrange"
-                  (rangeHdrs $ ByteRangeFromTo 1 0) emptyRange
-            `shouldRespondWith` 416
-
```
laurenceisla, line 86: I think this test should stay and return an error for `ByteRangeFromTo 2 0`, for example. Another test could be made for Range 1-0 to return an empty array, because it could be interpreted as offset 1 for the element in position 0.

test/Feature/RangeSpec.hs
```diff
@@ -330,11 +346,6 @@ spec = do
             simpleStatus r `shouldBe` ok200
 
       context "of invalid range" $ do
-        it "fails with 416 for offside range" $
-          request methodGet  "/items"
-                  (rangeHdrs $ ByteRangeFromTo 1 0) ""
-            `shouldRespondWith` 416
-
```
laurenceisla, line 337: Ditto.

gautam1168: ok ill make these changes

gautam1168:
@laurenceisla isn't this test ensuring that negative offsets pass as no-op's?
```
it "succeeds if offset is negative as a no-op" $
        get "/items?select=id&offset=-4&order=id"
          `shouldRespondWith`
            [json|[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}]|]
            { matchHeaders = ["Content-Range" <:> "0-14/*"] }
```

I think this means that negative offsets must be ignored. 

Also I talked with @steve-chavez before starting working on this PR and I think he did not want to edit the range module because it has been stable for a while.

laurenceisla:
> I think this means that negative offsets must be ignored.

Yes, I think it's a bit confusing in that regard. Negative offsets are ignored unless they are together with a `limit` in this and previous versions. I don't know if this was intended (there are no tests for this) but it can be inferred from the `restrictRange` function (and `urlRange` in `ApiRequest.hs`), that the parameter `rangeOffset r + limit -1` allows a negative offset. Doing `/clients?limit=4&offset=-2`  does not ignore the offset and works the same as `/client?limit=2` (tested this for v8.0 and v7.0 as well).

> Also I talked with @steve-chavez before starting working on this PR and I think he did not want to edit the range module because it has been stable for a while.

I see, but I think the changes I suggested are mostly on the parameters regarding to the Range limits, not so much on the structure of the functions themselves. For instance, from the example above: `restrictRange` forces the limit to be at least 1 `rangeOffset r + limit - 1`, removing the `-1` will allow a limit 0. I applied a similar reasoning for the other parameters that forced the limit without changing the structure of those functions. With those changes only the limit 0 and Range 1-0 fail, I couldn't find any other thing that breaks, maybe more review is needed.

steve-chavez:
@gautam1168 Sorry for the late reply, this fell off my radar.

> Also I talked with @steve-chavez before starting working on this PR and I think he did not want to edit the range module because it has been stable for a while.
> @steve-chavez can you please comment on whether I should modify my PR to actually edit the range module as @laurenceisla suggested or if this is more desirable?

Yeah, now that I think about it, we shouldn't fear changing the range module - we have solid tests and if they pass then all's good. So yes, I'd say go ahead and change what you need :+1:

laurenceisla:
I've re read all the convo from the Issue https://github.com/PostgREST/postgrest/issues/1121#issuecomment-907616208 and your PR and understood better your approach. Mine was wrong in trying to broaden the Ranges, it would generate more confusion when dealing with those ranges in the future. 

Due to the time and the changes made to main, I opened a new PR #2269. With these modifications I tried to make the changes a little simpler, i.e. instead of looking for limits again and check that match 0, it uses the already found `limitRange`, verifies if it's `limit=0` and uses a `limitZeroRange`, bypassing the intersection with the header ranges.

The `limit=-1` for embedding is a good find, it is sent to the database without verifying the ranges, most likely this behavior was not intended but, regardless, I think that should be left for another PR (if at all), knowing that negative offsets e.g. `offset=-1` also fail in the same way.

Sorry for the very late review.

PR #1931, opened 2021-08-31
fix: Return array of objects for RPCs with single-column TABLE return type

### LINKED ISSUE #1930: Simple objects returned from functions are converted to strings
### Environment

* PostgreSQL version: postgres:11 (docker image)
* PostgREST version: postgrest/postgrest:v8.0.0 (docker image)
* Operating system: Ubuntu 20.04

### Description of issue

The return type changed from version 7.0.1 to version 8.0.0 for functions that return simple object arrays.  In version 7.0.1 this returns the object with its original keys, but in version 8.0.0 the objects are converted to strings.

Defining a test function with permissions for anonymous users to run:

```sql
CREATE FUNCTION public_api.test() RETURNS TABLE(test TEXT) AS $$
    WITH series AS (
        SELECT generate_series(0, 10) AS gen
    )
    SELECT
        'Test ' || series.gen as test
    FROM series;
$$ LANGUAGE sql;

GRANT EXECUTE ON FUNCTION public_api.test() TO anon_user;
```

Using curl to post to this function:

```bash
curl -X POST http://127.0.0.1:3000/rpc/test -H "Accept: application/json"
```

With the PostgREST 7.0.1 image this returns (expected behavior):

```json
[{"test":"Test 0"}, 
 {"test":"Test 1"}, 
 {"test":"Test 2"}, 
 {"test":"Test 3"}, 
 {"test":"Test 4"}, 
 {"test":"Test 5"}, 
 {"test":"Test 6"}, 
 {"test":"Test 7"}, 
 {"test":"Test 8"}, 
 {"test":"Test 9"}, 
 {"test":"Test 10"}]
```

Taking this image down and replacing with the PostgREST 8.0.0 image the same command returns:

```json
["Test 0", "Test 1", "Test 2", "Test 3", "Test 4", "Test 5", "Test 6", "Test 7", "Test 8", "Test 9", "Test 10"]
```

wolfgangwalther:
This is a regression after changing responses for RPCs with INOUT or OUT arguments in #1615.

Resolves #1930.

The main idea of #1615 was "whenever any argument is given a name, we should return it in object style". We forgot about `RETURNS TABLE`, which has a special case when there is only one column given. With multiple columns in `RETURNS TABLE(...)`, PostgreSQL creates a record type behind the scenes, which we handle correctly. With only one column, this is not the case and we need to specifically check the existance of `TABLE`.

From the two tests I added the first one failed before my change, confirming #1930. The second one is working fine and is just added for completeness.

I added a second commit, too, which removes the check for the record pseudo type. I think we fully replaced that with the check for `TABLE` (and no test is failing), but it would be better to double check that one.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -284,10 +284,8 @@ procsSqlQuery = [q|
     COALESCE(comp.relname, t.typname) AS name,
     p.proretset AS rettype_is_setof,
     (t.typtype = 'c'
-     -- Only pg pseudo type that is a row type is 'record'
-     or t.typtype = 'p' and t.typname = 'record'
-     -- if any INOUT or OUT arguments present, treat as composite
-     or COALESCE(proargmodes::text[] && '{b,o}', false)
+     -- if any TABLE, INOUT or OUT arguments present, treat as composite
+     or COALESCE(proargmodes::text[] && '{t,b,o}', false)
     ) AS rettype_is_composite,
     p.provolatile,
     p.provariadic > 0 as hasvariadic
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #1871, Fix OpenAPI missing default values for String types and identify Array types as "array" instead of "string" - @laurenceisla
+ - #1930, Fix RPC return type handling for `RETURNS TABLE` with a single column. Regression of #1615. - @wolfgangwalther
+
 ### Changed
 
  - #1927, Overloaded Functions: If there's a function "my_func" having a single unnamed json param and other overloaded pairs(with any number of params), PostgREST won't be able to resolve a POST request to "my_func". For solving this, you can name the unnamed json param `my_func(json) -> my_func(prm json)`.
--- a/test/Feature/RpcSpec.hs
+++ b/test/Feature/RpcSpec.hs
@@ -572,6 +572,17 @@ spec actualPgVersion =
           `shouldRespondWith`
             [json|{"num":1,"str":"two","b":false}|]
 
+    context "procs with TABLE return" $ do
+      it "returns an object result when there is a single-column TABLE return type" $
+        get "/rpc/single_column_table_return"
+          `shouldRespondWith`
+            [json|[{"a": "A"}]|]
+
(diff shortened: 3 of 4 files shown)


steve-chavez:
Hey Wolfgang, nice to see you back :raised_hands: 

Just made this snippet to understand the bug

```sql
create or replace function haha() returns table(test TEXT) as $$ select 'haha' as test $$ language sql;
create or replace function hehe() returns table(test1 TEXT, test2 TEXT) as $$ select 'haha' as test1, 'hehe' as test2 $$ language sql;

select proc.proname, typ.typname as return_type from pg_proc proc inner join pg_type typ on typ.oid = proc.prorettype 
where proname in ('haha', 'hehe');

┌─────────┬─────────────┐
│ proname │ return_type │
├─────────┼─────────────┤
│ haha    │ text        │
│ hehe    │ record      │
└─────────┴─────────────┘
```

> With multiple columns in RETURNS TABLE(...), PostgreSQL creates a record type behind the scenes, which we handle correctly. With only one column, this is not the case and we need to specifically check the existance of TABLE.

So yes, pg doesn't consider a RETURNS TABLE having a single column as a `record`.

src/PostgREST/DbStructure.hs
```diff
@@ -283,10 +283,8 @@ procsSqlQuery = [q|
     COALESCE(comp.relname, t.typname) AS name,
     p.proretset AS rettype_is_setof,
     (t.typtype = 'c'
-     -- Only pg pseudo type that is a row type is 'record'
-     or t.typtype = 'p' and t.typname = 'record'
-     -- if any INOUT or OUT arguments present, treat as composite
-     or COALESCE(proargmodes::text[] && '{b,o}', false)
+     -- if any TABLE, INOUT or OUT arguments present, treat as composite
+     or COALESCE(proargmodes::text[] && '{t,b,o}', false)
```
steve-chavez, line 287: Looks correct, since we cannot trust on the `t.typname = 'record'` being true, better rely on [proargmodes](https://www.postgresql.org/docs/13/catalog-pg-proc.html)(pg catalog considers the RETURNS part as arguments).

wolfgangwalther:
So, I fixed the errors that I had for older PG versions. CI is still failing for FreeBSD, but I think that's a different issue,  @steve-chavez?

Should be good to merge.

steve-chavez (approved):
Yeah, the error was noted on https://github.com/PostgREST/postgrest/pull/1927#issuecomment-908771903, it's unrelated to this PR.

Issue #1397, opened 2019-10-24
Unexpected compression encoding for release archives


aviaviavi:
### Environment

* PostgreSQL version: NA
* PostgREST version: 6.0.2
* Operating system: MacOS Mojave

### Description of issue

When pulling down the release tarballs for 6.0.2, they are a `.xz` file, but from what I can tell they appear to be bzip compressed. 

```shell
$ file postgrest-v6.0.2-osx.tar.xz
postgrest-v6.0.2-osx.tar.xz: bzip2 compressed data, block size = 900k
```

It's not a huge deal, just made pulling down this binary from within some of my Haskell code a bit tricky :). 

I think at the very least, the compression encoding should match the file extension, and also maybe a more common one like gzip could be used?

steve-chavez:
That's true for the OSX release. Sorry about that.

Here I used bzip(`tar cjf`) instead of xz (`tar cJf`):
https://github.com/PostgREST/postgrest/blob/80f763448f9f84a7b783af79c05ddb92265fb31c/.travis.yml#L68

Already corrected `postgrest-v6.0.2-osx.tar.xz` on the releases page. Will fix the travis file.

aviaviavi:
Thanks!

PR #1948, opened 2021-09-14
pre-release: Bump to v9.0.0-a1

steve-chavez:

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            8.0.0
+version:            9.0.0-a1
 synopsis:           REST API for any Postgres database
 description:        Reads the schema of a PostgreSQL database and creates RESTful routes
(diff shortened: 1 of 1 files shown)


steve-chavez:
@monacoremo First time doing a pre-release with the new process. Mostly worked great!

Some things that I think should be improved:

- The tars are not using the usual `xz` compression. I think we should stick to `xz`, we had an issue about this before: https://github.com/PostgREST/postgrest/issues/1397. Also, I usually did a `tar xvf <release>.tar.xz && ./postgrest` and this worked, don't know why but now `./postgrest` is not executable by default(have to do a `chmod +x`).
- The PostgREST version shows as `Server: postgrest/9.0.0` and `Usage: ... PostgREST 9.0.0`, is it possible to include the `9.0.0-a1` there? This used to work with the nightlies(it showed `PostgREST 20210706` on "Usage" for example).

PR #1951, opened 2021-09-17
Fix the limit proc results test

laurenceisla:
A small fix for this test. The `limit` in the query wasn't doing anything due to the `?` operator.


wolfgangwalther:
So while this is better, the tests still don't really test the `limit` part. I guess the filter should be changed to `id=lt.6` instead of `id=lt.5`, to actually have a different result set in case the limit does not apply.

PR #1957, opened 2021-09-20
CI fixes

monacoremo:
Fix issues noted in https://github.com/PostgREST/postgrest/pull/1948#issuecomment-919547558

According to https://pvp.haskell.org/#version-tags, tags in the package version are not supported anymore. Cabal silently swallows them, which is why the earlier CI hack for pre-releases worked somehow.

I propose that we use version numbers with 4 components (e.g., 1.1.1.1) to identify pre-releases.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            9.0.0-a1
+version:            8.0.0.20210921
 synopsis:           REST API for any Postgres database
 description:        Reads the schema of a PostgreSQL database and creates RESTful routes
                     for the tables and views, supporting all HTTP verbs that security
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,7 +1,14 @@
 name: CI
+
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - main
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
 
 jobs:
   Lint-Style:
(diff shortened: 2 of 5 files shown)


steve-chavez:
> Historically Cabal supported version numbers with string tags at the end, e.g. 1.0-beta. This proved not to work well because the ordering for tags was not well defined. Version tags are no longer supported and mostly ignored, however some tools will fail in some circumstances if they encounter them.

Ah, that's surprising.

> I propose that we use version numbers with 4 components (e.g., 1.1.1.1) to identify pre-releases.

Hm, one issue with that convention is that it doesn't convey a pre-release, unlike the `<version>-a1` convention which kinda did.

How about if we tag the `pre-release` as `9.0.0.20210920`, the latter component being the date of the pre-release. Would that work with cabal?

monacoremo:
> How about if we tag the `pre-release` as `9.0.0.20210920`, the latter component being the date of the pre-release. Would that work with cabal?

Yes, that works! I like it.

Also, this PR adds a `(pre-release)` note to the version when it has four components. So you'll get `9.0.0.20210920 (pre-release)` in the help text and header.

wolfgangwalther:
> How about if we tag the `pre-release` as `9.0.0.20210920`, the latter component being the date of the pre-release. Would that work with cabal?

That's kind of strange, I think - because `9.0.0.20210920` would actually be **older** than `9.0.0`, right? Not sure if all tooling would understand that, I guess the basic assumption is, that "bigger number" = "newer release".

steve-chavez:
> That's kind of strange, I think - because 9.0.0.20210920 would actually be older than 9.0.0, right?

@wolfgangwalther Ah, yes you're right. I think the mistake here was bumping the version number for the pre-release, it should have been `8.0.0.20210920`. Maybe we shouldn't do any version(major, minor or patch) bump for the pre-releases.

postgrest.cabal
```diff
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            9.0.0-a1
+version:            8.0.0.20210921
```
wolfgangwalther, line 2: I see the version bump was correct - but should this not have triggered a new pre-release right here?

On the release page, I can still only find the `9.0.0-a1` release, but not this `8.0.0.20210921`. Same thing on docker hub.

Should we manually remove the `9.0.0-a1` release everywhere to go full circle on the correction?

wolfgangwalther: > but should this not have triggered a new pre-release right here?

Ah, I see - no tag was pushed at the same time...

PR #1968, opened 2021-09-28
Make GUC names compatible with PostgreSQL v14

### LINKED ISSUE #1857: Custom parameter names must be of the form "identifier.identifier"
### Environment

* PostgreSQL version: 14beta1
* PostgREST version: 7.0.1
* Operating system: CentOS 8

### Description of issue

I've been testing my current setup with the latest postgresql v14 beta1. There is a new limitation which does not allow to use current PostgreSQL as is.

The root case can be explained with this snippet:
```
postgres=# set local "request.jwt.claim.role" = 'webuser';
WARNING:  SET LOCAL can only be used in transaction blocks
ERROR:  invalid configuration parameter name "request.jwt.claim.role"
DETAIL:  Custom parameter names must be of the form "identifier.identifier".
```

A workaround was reverting the code which introduced this limitation:
```
git clone https://github.com/postgres/postgres.git
git revert 3db826bd55cd1df0dd8c3d811f8e5b936d7ba1e4
```

More details:
https://github.com/postgres/postgres/commit/3db826bd55cd1df0dd8c3d811f8e5b936d7ba1e4

I'm posting this here so people know that this will be an issue in the upcoming releases which needs to be solved.

laurenceisla:
Closes #1857

Currently a WIP.

- [ ] Add nix package for PostgreSQL v14
- [ ] Fix configuration names using `-`  (use JSON format values).

Following this approach from the comments: https://github.com/PostgREST/postgrest/issues/1857#issuecomment-854113924


laurenceisla:
@monacoremo I tried to update to the newest `nixpkgs` for the PostgreSQL v14beta3 (the beta1 doesn't have the changes made to allow multiple `.` to the GUC), but after updating using `nix-shell --run postgrest-nixpkgs-upgrade` and accessing the `nix-shell` I get this error: 
```
error: attribute 'ghc8104' missing, at /home/laurence/Projects/postgrest/nix/overlays/haskell-packages.nix:39:11
```
 Any tips on how to tackle this?

src/PostgREST/Middleware.hs
```diff
@@ -64,12 +68,12 @@ runPgLocals conf claims app req jsonDbS = do
...
     pathSql = setConfigLocal mempty ("request.path", iPath req)
-    headersSql = setConfigLocal "request.header." <$> iHeaders req
-    cookiesSql = setConfigLocal "request.cookie." <$> iCookies req
+    headersSql = setGuc "request.header." "request.headers" (iHeaders req)
+    cookiesSql = setGuc "request.cookie." "request.cookies" (iCookies req)
     claimsWithRole =
       let anon = JSON.String . toS $ configDbAnonRole conf in -- role claim defaults to anon if not specified in jwt
       M.union claims (M.singleton "role" anon)
-    claimsSql = setConfigLocal "request.jwt.claim." <$> [(toS c, toS $ unquoted v) | (c,v) <- M.toList claimsWithRole]
+    claimsSql = setGuc "request.jwt.claim." "request.jwt.claims" [(toS c, toS $ unquoted v) | (c,v) <- M.toList claimsWithRole]
```
laurenceisla, line 76: I think this could be optimized. Here `M.toList claimsWithRole` it looks like it's doing a conversion from HashMap to List, later the `gucJsonVal` function will do the opposite. Maybe avoid this by encoding the `claimsWithRole` directly.

src/PostgREST/Middleware.hs
```diff
@@ -83,6 +89,16 @@ runPgLocals conf claims app req jsonDbS = do
     setConfigLocal :: ByteString -> (ByteString, ByteString) -> H.Snippet
     setConfigLocal prefix (k, v) =
       "set_config(" <> unknownEncoder (prefix <> k) <> ", " <> unknownEncoder v <> ", true)"
+    -- | Starting from PostgreSQL v14, some characters are not allowed for config names (mostly affecting headers with "-").
+    -- | A JSON format string is used to avoid this problem. See https://github.com/PostgREST/postgrest/issues/1857
+    setGuc :: ByteString -> ByteString -> [(ByteString, ByteString)] -> [H.Snippet]
+    setGuc prefixString prefixJson keyVals
```
steve-chavez, line 95: I think this should be an independent function(outside of the `where`), also maybe rename the args like:

```suggestion
    setGuc prefixPgVerLt14 prefixPgVerGte14 keyVals
```

And a comment mentioning that the prefixes change between versions?

laurenceisla: `setGuc` uses the `setConfigLocal` function, in this case is it okay to also take that function out of the `where` too?

steve-chavez: Yeah, should be ok This `where` part seems too long now.

steve-chavez:
Hm, maybe we should offer a config option to force the new json gucs for users that have pg <=13 - this would ease migration for whenever they upgrade to pg 14.

We could call the config option `db-use-json-gucs`(false by default).

monacoremo:
> Never mind, the compiler in `default.nix` needed to be changed from `ghc8104` to `ghc8107` for it to work.

Yes, that's it! That however creates some complications with the static build. For this PR I would recommend to leave `nix/nixpkgs-version.nix` as-is, and instead add Postgres 14 the way we did in `nix/overlays/postgresql-legacy.nix`.

To do this, just copy and hook up the file in the `default.nix` files, e.g. as `postgresql-future.nix`, update the nixpkgs commit in that file to one that contains version 14 (e.g. https://github.com/NixOS/nixpkgs/commit/76b1e16c6659ccef7187ca69b287525fea133244) and the hash to something random in the right format (e.g. by changing one char); then set the right hash that nix-build expects.

laurenceisla:
> Edit: This is how I recommend to do it @laurenceisla : [monacoremo@9de7c74](https://github.com/monacoremo/postgrest/commit/9de7c74cedacf1cfd7eec91c5bb397d434c2cfe5)

@monacoremo  Cool, thanks for the help, I modified the files as you mentioned here. There's no more Linux static Nix errors now. I wonder if the `ci.yaml` file should be modified to accept tests for PostgreSQL 14 or is it better to add it later?

monacoremo:
> I wonder if the `ci.yaml` file should be modified to accept tests for PostgreSQL 14 or is it better to add it later?

Let's add it here!

wolfgangwalther:
> We could call the config option `db-use-json-gucs`(false by default on pg <= 13? Hmm, maybe just false by default and not considered on pg >= 14).

Maybe `db-use-legacy-gucs` with a default `true` on pg <= 13 and not considered on pg >= 14. This makes it a bit clearer that those gucs are not the way forward.

.github/workflows/ci.yaml
```diff
@@ -34,13 +34,16 @@ jobs:
...
+      - name: Run coverage (IO tests and Spec tests against PostgreSQL 14)
         run: postgrest-coverage
       - name: Upload coverage to codecov
         uses: codecov/codecov-action@v2
         with:
           files: ./coverage/codecov.json
 
+      - name: Run the spec tests against PostgreSQL 13
+        if: always()
+        run: postgrest-with-postgresql-13 postgrest-test-spec
```
laurenceisla, line 46: @monacoremo Running the command `postgrest-coverage` already uses the PostgreSQL 14 in my local environment, does it do the same for the CI or is anything else needed? If so, the missing one would be the tests on v13, which I added here.

Issue #1210, opened 2018-11-10
Feature Request: ability to call pg 11 stored procedure


sunbing81:
### Environment

* PostgreSQL version: 11
* PostgREST version: 0.5.1
* Operating system: not relevant

### Description of issue

1. create the stored procedure (not a function)
```
create procedure api.test_proc(a int) language sql as $$
select 1;
$$;
```

2. make the request
```
POST /api/rpc/test_proc
Content-Type: application/json

{"a": 1}
```

3. postgrest cannot call a stored procedure for now
```
: HTTP/1.1 400 Bad Request
: Server: nginx/1.10.3 (Ubuntu)
: Date: Sat, 10 Nov 2018 09:04:34 GMT
: Content-Type: application/json; charset=utf-8
: Transfer-Encoding: chunked
: Connection: keep-alive
:
: {"hint":"To call a procedure, use CALL.","details":null,"code":"42809","message":"api.test_proc(a => integer) is a procedure"}
```

steve-chavez:
At first I thought PROCEDUREs can't have return values, but turns out they work with INOUT parameters(only OUT parameters don't work btw).
```sql
create procedure public.test_proc(inout a int, inout b int) language sql as $$
select 1, 2;
$$;
 
call test_proc(3, 4);

 a | b 
---+---
 1 | 2
(1 row)
```

Going to have to think this through. Had the idea that a 204 No Content would always be the response.

steve-chavez:
`call` syntax is severely limited. We cannot dynamically construct the input(arguments) and output(result) in SQL as we have done with traditional stored functions.

## Arguments

Not possible to:
```sql
-- use ctes
with smth as(
  select 1
)
call test_proc(3, 4);
-- ERROR:  42601: syntax error at or near "call"

-- use subqueries
call test_proc((select 3), (select 4));
-- ERROR:  0A000: cannot use subquery in CALL argument

-- it's possible to call functions in arguments though
create table myrowtype(a int, b int);
call test_proc((json_populate_record(null::myrowtype, '{"a":1,"b":2}')).a,(json_populate_record(null::myrowtype, '{"a":1,"b":2}')).b);
 a | b
---+---
 1 | 2

-- but not possible to use "AS" inside call
call test_proc((json_to_record('{"a":1,"b": 3}') as x(a int, b int)).a, (json_to_record('{"a":1,"b": 3}') as x(a int, b int)).b);
-- ERROR:  42601: syntax error at or near "as"
```
[...]

PR #1976, opened 2021-10-08
Add support for Stored Procedures

laurenceisla:
Resolves part of #1210. As noted in the issue, nested transactions won't be implemented

- [x] Identify stored procedures in the schema cache
- [ ] Query the data from the procedure (INOUT and OUT), use  https://github.com/steve-chavez/postgrest/commit/8ac5a11b35a98685e288ced786f6377de8f90641 as reference
- [ ] Add tests
- [ ] Optimize code
- [ ] Modify changelog

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -462,13 +462,13 @@ handleInvoke invMethod proc context@RequestContext{..} = do
       (if invMethod == InvHead then mempty else toS body)
 
 handleOpenApi :: Bool -> Schema -> RequestContext -> DbHandler Wai.Response
-handleOpenApi headersOnly tSchema (RequestContext conf@AppConfig{..} dbStructure apiRequest _) = do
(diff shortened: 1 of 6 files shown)


steve-chavez:
We can fully support stored procedures transactions with the idea mentioned [here](https://github.com/PostgREST/postgrest/issues/1210#issuecomment-1180721273).

steve-chavez:
On 9.0.0, when trying to call a procedure, we used to fail with:

```
{
    "hint": "To call a procedure, use CALL.",
    "details": null,
    "code": "42809",
    "message": "api.do_nothing() is a procedure"
}
```

Now we fail with a schema cache error:

```
Could not find the api.do_nothing() function in the schema cache
```

I think both errors are imprecise, ideally we should throw a [501 Not Implemented](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) and mention procedures are not supported yet. Also maybe suggest moving to a function since we already wrap it with a transaction.

PR #1988, opened 2021-10-23
Make JWT authorization a middleware

DavidLindbom:
I found myself switching the log level on my Postgres server a lot when needing to debug permissions because I wanted to see what role the query ran as. This PR moves the JWT parsing to a middleware such that it is now available for the logging middleware.

Example of new access log output (e.g the `api_user` value)

```
127.0.0.1 api_user - [23/Oct/2021:22:58:32 +0200] "GET /todos HTTP/1.1" 200 - "" "curl/7.64.1"
```

This follows the style of wai-middleware-auth package and makes the JWT claims available anywhere we have a `Request` value through the `vault` mechanism in Wai. 

The existing JWT handling is not yet replaced as I wanted to get some feedback if this is interesting for anyone else.

### DIFF
--- a/src/PostgREST/Middleware.hs
+++ b/src/PostgREST/Middleware.hs
@@ -10,6 +10,8 @@ module PostgREST.Middleware
   , defaultCorsPolicy
   , corsPolicy
   , optionalRollback
+  , getJwtClaims
+  , getRoleClaim
   ) where
 
 import qualified Data.Aeson                           as JSON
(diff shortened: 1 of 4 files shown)


src/PostgREST/Middleware.hs
```diff
@@ -100,7 +104,9 @@ pgrstFormat minStatus date req status responseSize =
   if status < minStatus
     then mempty
   else toLogStr (getSourceFromSocket req)
-    <> " - - ["
+    <> " "
+    <> toLogStr (maybe "-" unquoted (getRoleClaim req))
+    <> " - ["
```
wolfgangwalther, line 109: Since I'd like to remove `pgrstFormat` in #1724, I opened a PR for `wai-logger` to add upstream support for the current user in the apache logger. See https://github.com/kazu-yamamoto/logger/pull/198 for details.

steve-chavez: @wolfgangwalther  Just saw https://github.com/kazu-yamamoto/logger/pull/198 was merged :raised_hands:. Nice work!

DavidLindbom:
Hello again,

I made some updates to remove the duplicated JWT parsing and now it should only use the new middleware.

There are a couple of not so nice parts you might want to look at

1. The failure case in `authMiddleware` and how to propagate this to the next middleware without running the full request function.
2. `Middleware.runPgLocals`. I looked at it but quickly decided it was outside the scope of what I wanted to do. There are some simplifications that should be possible there.

wolfgangwalther:
Now that #1724 is merged, the logger middleware has heavily changed. Therefore, I took this PR and rebased it on latest main. I can't figure out how to push to this branch, so I created #2115, instead. See my comment over there on what I did.

PR #1996, opened 2021-10-28
Allow overloaded functions if one has a single unnamed JSON param

laurenceisla:
WIP

Implementing this idea https://github.com/PostgREST/postgrest/pull/1927#discussion_r737776638 to avoid responding with a `300 Multiple Choices` error when doing a POST request to overloaded functions if one has a single unnamed JSON parameter.

The first commit is an ugly attempt, which is working so far (only the test for the overloaded function is failing as expected).

### DIFF
--- a/src/PostgREST/Request/ApiRequest.hs
+++ b/src/PostgREST/Request/ApiRequest.hs
@@ -490,24 +490,30 @@ rawContentTypes AppConfig{..} =
 findProc :: QualifiedIdentifier -> S.Set Text -> Bool -> ProcsMap -> ContentType -> Bool -> Either ApiRequestError ProcDescription
 findProc qi argumentsKeys paramsAsSingleObject allProcs contentType isInvPost =
   case matchProc of
-    []     -> Left $ NoRpc (qiSchema qi) (qiName qi) (S.toList argumentsKeys) paramsAsSingleObject contentType isInvPost
-    [proc] -> Right proc
-    procs  -> Left $ AmbiguousRpc (toList procs)
+    ([], [])     -> Left $ NoRpc (qiSchema qi) (qiName qi) (S.toList argumentsKeys) paramsAsSingleObject contentType isInvPost
+    -- If there are no procedures with named arguments, fallback to the single unnamed argument procedure
+    ([], [proc]) -> Right proc
+    ([], procs)  -> Left $ AmbiguousRpc (toList procs)
+    -- Matches the procedures with named arguments
+    ([proc], _)  -> Right proc
+    (procs, _)   -> Left $ AmbiguousRpc (toList procs)
   where
-    matchProc =
-      if null (fst findProcPartition) && isInvPost && contentType == CTApplicationJSON
-        then snd findProcPartition -- use the fallback function if none is found with the provided parameters
-        else fst findProcPartition
-    findProcPartition = overloadedProcPartition $ M.lookupDefault mempty qi allProcs -- first find the proc by name
+    matchProc = overloadedProcPartition $ M.lookupDefault mempty qi allProcs -- first find the proc by name
     -- The partition obtained has the form (overloadedProcs,fallbackProcs)
-    -- where fallbackProcs are functions with a single unnamed parameter of type json or jsonb
+    -- where fallbackProcs are functions with a single unnamed parameter
     overloadedProcPartition procs = foldr select ([],[]) procs
     select proc ~(ts,fs)
-      | matchesParams proc             = (proc:ts,fs)
-      | hasSingleUnnamedJsonParam proc = (ts,proc:fs)
-      | otherwise                      = (ts,fs)
-    hasSingleUnnamedJsonParam proc = case pdParams proc of
-      [ProcParam ppName ppType _ _] -> T.null ppName && ppType `elem` ["json", "jsonb"]
+      | matchesParams proc         = (proc:ts,fs)
+      | hasSinglaUnnamedParam proc = (ts,proc:fs)
+      | otherwise                  = (ts,fs)
+    -- If the function is called with post and has a single unnamed parameter
+    -- it can be called depending on content type and the parameter type
+    hasSinglaUnnamedParam proc = isInvPost && case pdParams proc of
+      [ProcParam "" ppType _ _]
+        | contentType == CTApplicationJSON -> ppType `elem` ["json", "jsonb"]
+        | contentType == CTTextPlain       -> ppType == "text"
+        | contentType == CTOctetStream     -> ppType == "bytea"
+        | otherwise                        -> False
       _ -> False
     matchesParams proc =
       let params = pdParams proc in
@@ -517,14 +523,6 @@ findProc qi argumentsKeys paramsAsSingleObject allProcs contentType isInvPost =
       -- If the function has no parameters, the arguments keys must be empty as well
       else if null params
         then null argumentsKeys
-      -- If the function is called with post and has a single unnamed parameter
-      -- it can be called depending on content type and the parameter type
-      else if isInvPost && length params == 1 && (ppName <$> headMay params) == Just mempty
-        then case headMay params of
-          Just prm | contentType == CTTextPlain       -> ppType prm == "text"
-                   | contentType == CTOctetStream     -> ppType prm == "bytea"
-                   | otherwise                        -> False
-          Nothing  -> False
       -- A function has optional and required parameters. Optional parameters have a default value and
       -- don't require arguments for the function to be executed, required parameters must have an argument present.
       else case L.partition ppReq params of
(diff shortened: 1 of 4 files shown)


src/PostgREST/Request/ApiRequest.hs
```diff
@@ -494,7 +494,11 @@ findProc qi argumentsKeys paramsAsSingleObject allProcs contentType isInvPost =
     [proc] -> Right proc
     procs  -> Left $ AmbiguousRpc (toList procs)
   where
-    matchProc = filter matchesParams $ M.lookupDefault mempty qi allProcs -- first find the proc by name
+    hasSingleUnnamedJsonParam proc = case proc of
+          ProcDescription{pdParams=[ProcParam ppName ppType _ _]} -> T.null ppName && ppType `elem` ["json", "jsonb"]
+          _ -> False
+    matchProc = if null matchProc' && isInvPost then filter hasSingleUnnamedJsonParam (M.lookupDefault mempty qi allProcs) else matchProc'
```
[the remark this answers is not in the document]
laurenceisla: > I don't like too much that a second filter is made on the overloaded procedures again though, if it affects performance I'll try to find an alternative.

Addressed that using a `foldr` with a partition to avoid the double filter. It should be ready to review right now.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -494,7 +494,21 @@ findProc qi argumentsKeys paramsAsSingleObject allProcs contentType isInvPost =
...
+    -- The partition obtained has the form (overloadedProcs,fallbackProcs)
+    -- where fallbackProcs are functions with a single unnamed parameter of type json or jsonb
+    overloadedProcPartition procs = foldr select ([],[]) procs
+    select proc ~(ts,fs)
+      | matchesParams proc             = (proc:ts,fs)
+      | hasSingleUnnamedJsonParam proc = (ts,proc:fs)
+      | otherwise                      = (ts,fs)
+    hasSingleUnnamedJsonParam proc = case pdParams proc of
+      [ProcParam ppName ppType _ _] -> T.null ppName && ppType `elem` ["json", "jsonb"]
+      _ -> False
```
wolfgangwalther, line 511: Should we maybe put this bit on the `ProcDescription` type at schema cache creation time already?

We currently have:

```haskell
data ProcDescription = ProcDescription
[...]
  , pdParams      :: [ProcParam]
[...]
```

and could change it to something like

```haskell
data ProcDescription = ProcDescription
[...]
  , pdUnnamedParam     :: Maybe ProcUnnamedParamType
  , pdNamedParams      :: [ProcParam]
[...]

data ProcUnnamedParamType = SingleBytea | SingleJson | SingleText
```

And then filter the procs at schema cache creation time already:
- remove all procs with more than 1 unnamed param
- remove all procs with 1 unnamed param + named params (side note: should we add a test-case for such a situation?)
- remove all unnamed args from `pdNamedParams`

I think this could make the whole logic here (and below) a bit more readable?

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -494,7 +494,21 @@ findProc qi argumentsKeys paramsAsSingleObject allProcs contentType isInvPost =
...
+      if null (fst findProcPartition) && isInvPost && contentType == CTApplicationJSON
+        then snd findProcPartition -- use the fallback function if none is found with the provided parameters
+        else fst findProcPartition
+    findProcPartition = overloadedProcPartition $ M.lookupDefault mempty qi allProcs -- first find the proc by name
+    -- The partition obtained has the form (overloadedProcs,fallbackProcs)
+    -- where fallbackProcs are functions with a single unnamed parameter of type json or jsonb
+    overloadedProcPartition procs = foldr select ([],[]) procs
+    select proc ~(ts,fs)
+      | matchesParams proc             = (proc:ts,fs)
+      | hasSingleUnnamedJsonParam proc = (ts,proc:fs)
```
wolfgangwalther, line 507: I like this!

Should we maybe treat the other single unnamed cases as "fallback", too?

I think that would make sense. Since in those cases, we don't have any keys from the payload to parse, the `overloadedProcs` part of the partition would be empty - and we would always fallback. So maybe this will then be `(namedProcs, singleUnnamedProcs)` instead.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -494,7 +494,21 @@ findProc qi argumentsKeys paramsAsSingleObject allProcs contentType isInvPost =
     [proc] -> Right proc
     procs  -> Left $ AmbiguousRpc (toList procs)
   where
-    matchProc = filter matchesParams $ M.lookupDefault mempty qi allProcs -- first find the proc by name
+    matchProc =
+      if null (fst findProcPartition) && isInvPost && contentType == CTApplicationJSON
+        then snd findProcPartition -- use the fallback function if none is found with the provided parameters
+        else fst findProcPartition
```
wolfgangwalther, line 500: I think we can move this part into the `case matchProc of` above. I feel like that would make the logic a bit clearer. Something like:

```haskell
  case findProcPartition of
    ([], [])     -> Left $ NoRpc (qiSchema qi) (qiName qi) (S.toList argumentsKeys) paramsAsSingleObject contentType isInvPost
    -- no matches with named arguments, fallback to single unnamed argument
    ([], [proc]) -> Right proc
    ([], procs)  -> Left $ AmbiguousRpc (toList procs)
    -- matches with named arguments
    ([proc], _)  -> Right proc
    (procs, _)   -> Left $ AmbiguousRpc (toList procs)
```

wolfgangwalther (changes requested):

src/PostgREST/Request/ApiRequest.hs
laurenceisla: I like the idea of doing the filter at schema cache creation time, but separating the parameters, while it does make the logic more readable here, it complicates it a bit on other parts, like [here](https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Request/DbRequestBuilder.hs#L360) or in the OpenApi for instance. 

Maybe adding the fallback type and using it as the filter if it exists could be a bit simpler?

```haskell
data ProcDescription = ProcDescription
[...]
  , pdFallbackType :: Maybe ProcFallbackType
[...]

data ProcFallbackType = Bytea | Json | Text
```

> - remove all procs with more than 1 unnamed param
> - remove all procs with 1 unnamed param + named params (side note: should we add a test-case for such a situation?)

Removing all procs would make functions like this `create or replace function api.test_overload(c int, json default '{}')` to not work causing a breaking change, although I can't think of a use case right now. A solution would be to remove those where at least one unnamed parameter doesn't have a default value.

wolfgangwalther: Ok, replace "unnamed param" with "unnamed required param" then, in my list above.

> Maybe adding the fallback type and using it as the filter if it exists could be a bit simpler?

Ok, let's do it like that. Let's call it like this:

```haskell
data ProcDescription = ProcDescription
[...]
  , pdPositionalParam :: Maybe ProcPositionalParam
[...]

data ProcPositionalParam = SingleBytea | SingleJson | SingleText
```

I think we need the `Single` prefix, otherwise we'd get conflicts with built-in `Text` very quickly.

Positional is better than "Fallback", because the first is descriptive of what is in the database, while the second is already hinting at how it might be used. But `ProcDescription` is really just that - a description.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -490,11 +490,31 @@ rawContentTypes AppConfig{..} =
...
+    -- The partition obtained has the form (overloadedProcs,fallbackProcs)
+    -- where fallbackProcs are functions with a single unnamed parameter
+    overloadedProcPartition procs = foldr select ([],[]) procs
+    select proc ~(ts,fs)
+      | matchesParams proc         = (proc:ts,fs)
+      | hasSinglaUnnamedParam proc = (ts,proc:fs)
+      | otherwise                  = (ts,fs)
+    -- If the function is called with post and has a single unnamed parameter
+    -- it can be called depending on content type and the parameter type
+    hasSinglaUnnamedParam proc = isInvPost && case pdParams proc of
```
wolfgangwalther, line 511:
```suggestion
      | hasSingleUnnamedParam proc = (ts,proc:fs)
      | otherwise                  = (ts,fs)
    -- If the function is called with post and has a single unnamed parameter
    -- it can be called depending on content type and the parameter type
    hasSingleUnnamedParam proc = isInvPost && case pdParams proc of
```

src/PostgREST/Request/ApiRequest.hs
wolfgangwalther: With the latest changes, I think it's actually quite readable already. Maybe we don't need to change the typing at all for now.

laurenceisla: Yes, I agree, it looks good right now. I'll leave this part as is, fix some tests and then this should be ready for a last review and merge.

wolfgangwalther (approved):
LGTM.

wolfgangwalther:
TODO: Adjust the Changelog entry for #1927.

wolfgangwalther (approved):

wolfgangwalther:
Nice work!

steve-chavez (approved):

PR #2003, opened 2021-11-02
Nix: Add statix linting for nix code

monacoremo:
https://github.com/nerdypepper/statix is a very promising new linting tool for nix code, it already identified and automatically fixed several improvements in our codebase. Hooked into `postgrest-style`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -80,8 +80,7 @@ let
       }
     );
 
-  lib =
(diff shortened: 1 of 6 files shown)


monacoremo:
This should not be contentious I believe as it only impacts our Nix setup - merging!

PR #2016, opened 2021-11-07
Refactor: Remove (mis)uses of GHC.Show instances

monacoremo:
This is an item on our refactoring roadmap #1804 

Using `GHC.Show` to pretty-print or serialize data is an [anti-pattern](https://www.stephendiehl.com/posts/strings.html), show is meant to create a Haskell expression that can be parsed by `read`, see https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-Show.html

This PR removes all uses of `GHC.Show` instances in the codebase. As a side effect, this avoids several string conversions, especially when parsing preference headers, for what is likely a small performance gain.

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -50,7 +50,7 @@ import System.Environment      (getEnvironment)
 import System.Posix.Types      (FileMode)
 
 import PostgREST.Config.JSPath           (JSPath, JSPathExp (..),
-                                          pRoleClaimKey)
+                                          dumpJSPath, pRoleClaimKey)
 import PostgREST.Config.Proxy            (Proxy (..),
                                           isMalformedProxyUri, toURI)
 import PostgREST.DbStructure.Identifiers (QualifiedIdentifier, dumpQi,
@@ -100,10 +100,10 @@ data LogLevel = LogCrit | LogError | LogWarn | LogInfo
 
 dumpLogLevel :: LogLevel -> Text
 dumpLogLevel = \case
-  LogCrit -> "crit"
+  LogCrit  -> "crit"
   LogError -> "error"
-  LogWarn -> "warn"
-  LogInfo -> "info"
+  LogWarn  -> "warn"
+  LogInfo  -> "info"
 
 data OpenAPIMode = OAFollowPriv | OAIgnorePriv | OADisabled
   deriving Eq
@@ -138,7 +138,7 @@ toText conf =
       ,("db-embed-default-join",     q . dumpJoin . configDbEmbedDefaultJoin)
       ,("db-use-legacy-gucs",            T.toLower . show . configDbUseLegacyGucs)
       ,("jwt-aud",                       toS . encode . maybe "" toJSON . configJwtAudience)
-      ,("jwt-role-claim-key",        q . T.intercalate mempty . fmap show . configJwtRoleClaimKey)
+      ,("jwt-role-claim-key",        q . T.intercalate mempty . fmap dumpJSPath . configJwtRoleClaimKey)
       ,("jwt-secret",                q . toS . showJwtSecret)
       ,("jwt-secret-is-base64",          T.toLower . show . configJwtSecretIsBase64)
       ,("log-level",                 q . dumpLogLevel . configLogLevel)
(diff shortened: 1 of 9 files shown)


wolfgangwalther:
I think I get the reasoning for not using `GHC.Show`.

But are there any reasons against using custom classes for this purpose?

Just two examples from the first commit (preferences):
- Why not one `ToAppliedHeader` class for both `resolutionAppliedHeader` and `transactionAppliedHeader`?
- Why not one `ToHeaderValue` class for all of `resolutionStr`, `representationStr`, `parametersStr`, `countStr`, `transactionStr`?

I think this would make it a bit more concise and clearly show that those functions mentioned above all share a common goal / output.

monacoremo:
> I think this would make it a bit more concise and clearly show that those functions mentioned above all share a common goal / output.

True, that's a bit cleaner! I tried to keep it as simple as possible in general, but it makes sense to use typeclasses in this case.

src/PostgREST/Request/Preferences.hs
```diff
@@ -19,36 +72,38 @@ data PreferRepresentation
...
+instance ToHeaderValue PreferCount where
+  toHeaderValue ExactCount     = "count=exact"
+  toHeaderValue PlannedCount   = "count=planned"
+  toHeaderValue EstimatedCount = "count=estimated"
 
 data PreferTransaction
-  = Commit   -- Commit transaction - the default.
-  | Rollback -- Rollback transaction after sending the response - does not persist changes, e.g. for running tests.
+  = Commit   -- ^ Commit transaction - the default.
+  | Rollback -- ^ Rollback transaction after sending the response - does not persist changes, e.g. for running tests.
```
wolfgangwalther, line 102: What is this change about? Does the `^` do anything special - or what does it mean? If not, I'd rather remove it from the other comments...?

src/PostgREST/Request/Preferences.hs
```diff
@@ -1,16 +1,69 @@
...
 import Protolude
 
 
+data Preferences
+  = Preferences
+    { preferResolution     :: Maybe PreferResolution
+    , preferRepresentation :: PreferRepresentation
+    , preferParameters     :: Maybe PreferParameters
+    , preferCount          :: Maybe PreferCount
+    , preferTransaction    :: Maybe PreferTransaction
```
wolfgangwalther, line 25: I don't understand why we're getting a codecov warning here that those 5 lines are not covered. Are those getters not what is used in `ApiRequest`?

src/PostgREST/Config.hs
```diff
@@ -168,6 +169,9 @@ toText conf =
         secret = fromMaybe mempty $ configJwtSecret c
     showSocketMode c = showOct (configServerUnixSocketMode c) mempty
 
+    innerJoin JTInner = "inner"
+    innerJoin JTLeft  = "left"
```
wolfgangwalther, line 173: This should be `dumpJoin` or so, looking at the other values. But Steve is going to remove it soon anyway, so maybe not that important.

wolfgangwalther:
> Edit: see latest commit with those changes for Preferences.

I Like that!

> We should do something similar for dumpConfig, but that's for another PR!

Agreed. This could be something like `ToConfigValue` and should include `showTxEnd` and the other `showXXX` functions we currently have. Maybe we could even extend this class to cover some of the builtin types that we use for other config options. The class could then include the quoting etc...

And then there should be a class for `ToSqlFragment`. I'm sure we will find "a few" other cases where using that would make things consistent.

src/PostgREST/Request/Preferences.hs
monacoremo: Those are for Haddock comments after a declaration, see: https://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields

src/PostgREST/Request/Preferences.hs
monacoremo: I think that's the `RecordWildCards` bug we already struggled with

src/PostgREST/Config.hs
monacoremo: True, fixed!

monacoremo:
Many thanks for reviewing @wolfgangwalther ! The FreeBSD build is doing weird things again due to some cache issues, but otherwise this should be good to merge

src/PostgREST/Request/Preferences.hs
wolfgangwalther: Right, I remember... gotta get used to things again :D

wolfgangwalther (approved):

PR #2017, opened 2021-11-07
Refactor: Remove uses of Protolude.Conv.toS

monacoremo:
This is another item on our refactoring roadmap #1804 and builds on top of the commits in #2016

Currently, we use `toS` from [`Protolude.Conv`](https://hackage.haskell.org/package/protolude-0.3.0/docs/Protolude-Conv.html) in almost all modules. `toS` contains partial functions and expensive transformations, which is why it is not exported by Protolude by default. For example, converting `ByteString` to `Text` involves [`T.decodeUtf8`](https://hackage.haskell.org/package/text-1.2.5.0/docs/Data-Text-Encoding.html#v:decodeUtf8) (see: https://hackage.haskell.org/package/protolude-0.3.0/docs/src/Protolude.Conv.html#StringConv) which will raise an uncatchable exception if the input is not valid utf-8.

This PR replaces all uses of `Protolude.Conv.toS` with the respective explicit transformations that are needed, making transparent where we are doing expensive (e.g., `BS.unpack`), performance critical (`LBS.toStrict`) or unsafe (`T.decodeUtf8`) ones, as a basis for further fixes.

### DIFF
--- a/src/PostgREST/GucHeader.hs
+++ b/src/PostgREST/GucHeader.hs
@@ -10,8 +10,7 @@ import qualified Data.HashMap.Strict  as M
 
 import Network.HTTP.Types.Header (Header)
 
-import Protolude      hiding (toS)
-import Protolude.Conv (toS)
+import Protolude
 
 
 {-|
@@ -21,11 +20,11 @@ import Protolude.Conv (toS)
 newtype GucHeader = GucHeader (CI.CI ByteString, ByteString)
 
 instance JSON.FromJSON GucHeader where
-  parseJSON (JSON.Object o) = case headMay (M.toList o) of
-    Just (k, JSON.String s) | M.size o == 1 -> pure $ GucHeader (CI.mk $ toS k, toS s)
-                            | otherwise     -> mzero
-    _ -> mzero
-  parseJSON _          = mzero
+  parseJSON (JSON.Object o) =
+    case M.toList o of
+      [(k, JSON.String s)] -> pure $ GucHeader (CI.mk $ toUtf8 k, toUtf8 s)
+      _ -> mzero
+  parseJSON _ = mzero
 
 unwrapGucHeader :: GucHeader -> Header
 unwrapGucHeader (GucHeader (k, v)) = (k, v)
--- a/src/PostgREST/Middleware.hs
+++ b/src/PostgREST/Middleware.hs
@@ -14,10 +14,11 @@ module PostgREST.Middleware
 
 import qualified Data.Aeson                           as JSON
 import qualified Data.ByteString.Char8                as BS
-import qualified Data.ByteString.Lazy.Char8           as BSL
+import qualified Data.ByteString.Lazy.Char8           as LBS
 import qualified Data.CaseInsensitive                 as CI
 import qualified Data.HashMap.Strict                  as M
 import qualified Data.Text                            as T
+import qualified Data.Text.Encoding                   as T
 import qualified Hasql.Decoders                       as HD
 import qualified Hasql.DynamicStatements.Snippet      as SQL hiding
                                                              (sql)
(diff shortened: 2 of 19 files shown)


src/PostgREST/GucHeader.hs
```diff
@@ -21,11 +20,11 @@ import Protolude.Conv (toS)
...
 
 instance JSON.FromJSON GucHeader where
-  parseJSON (JSON.Object o) = case headMay (M.toList o) of
-    Just (k, JSON.String s) | M.size o == 1 -> pure $ GucHeader (CI.mk $ toS k, toS s)
-                            | otherwise     -> mzero
-    _ -> mzero
-  parseJSON _          = mzero
+  parseJSON (JSON.Object o) =
+    case M.toList o of
+      [(k, JSON.String s)] -> pure $ GucHeader (CI.mk $ toUtf8 k, toUtf8 s)
```
wolfgangwalther, line 25: Much better... :)

src/PostgREST/Middleware.hs
```diff
@@ -152,15 +151,15 @@ corsPolicy req = case lookup "origin" headers of
   where
     headers = Wai.requestHeaders req
     accHeaders = case lookup "access-control-request-headers" headers of
-      Just hdrs -> map (CI.mk . toS . T.strip . toS) $ BS.split ',' hdrs
+      Just hdrs -> map (CI.mk . BS.dropWhile (' ' ==) . BS.dropWhileEnd (' ' ==)) $ BS.split ',' hdrs
```
wolfgangwalther, line 154: Since `BS` is `Data.ByteString.Char8`, I think there is `BS.strip`, too?

https://hackage.haskell.org/package/bytestring-0.11.1.0/docs/src/Data.ByteString.Char8.html#strip

monacoremo: Ah, good point! Fixed

postgrest.cabal
```diff
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            8.0.0.20210921
+version:            8.0.0.20211109
```
steve-chavez, line 2: @monacoremo Don't know if you noticed this, but somehow the `v9.0.0-a1` pre-release was deleted, its assets are gone but the [tag](https://github.com/PostgREST/postgrest/releases/tag/v9.0.0-a1) remains.

Is there some kind of GitHub rule where there can be only one `pre-release`? If so, could this new pre-release delete the previous one?

monacoremo: Hm, we are deleting an existing pre-release with the same tag here: https://github.com/PostgREST/postgrest/blob/2b22f889bdd191453c3f2798b5ea9da024231c4f/.github/workflows/ci.yaml#L327 - but this shouldn't  have deleted that one, unless something went wrong?

I updated the version here to invalidate the FreeBSD CI cache which is doing weird things, it shoudn't have triggered anything as there is not tag

wolfgangwalther: Here's the log of the last release's CI run:
https://github.com/PostgREST/postgrest/runs/4088373432?check_suite_focus=true

It says:

```
Releasing version v8.0.0.20211102 on GitHub (isprerelease=1)...
release not found
https://github.com/PostgREST/postgrest/releases/tag/v8.0.0.20211102
```

The 2nd line is from the delete command - so it should not have deleted the v9 release.

I noticed the release was gone a few days ago, too - but I assumed somebody deleted it manually because of the accidental version bump?

wolfgangwalther: > Is there some kind of GitHub rule [..] where there can be only one pre-release?

That seems unlikely, because we still have the old nightly release around - so we do have 2 pre-releases at the same time right now, in terms of GitHub.

monacoremo:
Tests pass, let's merge? @steve-chavez @wolfgangwalther

wolfgangwalther:
> Tests pass, let's merge? @steve-chavez @wolfgangwalther

Not sure about version bumping without tagging? Might be confusing. Or are you going to?

postgrest.cabal
steve-chavez: >  but I assumed somebody deleted it manually because of the accidental version bump?

Hm, I didn't delete it. Don't know what might have happened.

monacoremo:
> > Tests pass, let's merge? @steve-chavez @wolfgangwalther
> 
> Not sure about version bumping without tagging? Might be confusing. Or are you going to?

Took it out! Not sure it would be confusing though, we don't need to (pre-)release all new versions...

PR #2019, opened 2021-11-08
Add doctests

monacoremo:
This is another item on our refactorin roadmap #1804, building on top of #2016 

This PR integrates doctests into our Nix tooling and adds examples to the `Request.Preferences` module

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,6 +67,14 @@ jobs:
         if: always()
         run: postgrest-with-postgresql-9.5 postgrest-test-spec
 
+      - name: Run query cost tests against all PostgreSQL versions
+        if: always()
+        run: postgrest-with-all postgrest-test-querycost
+
+      - name: Run doctests
+        if: always()
+        run: nix-shell --run postgrest-test-doctests
+
       - name: Check the spec tests for idempotence
         if: always()
         run: postgrest-test-spec-idempotence
@@ -289,7 +297,7 @@ jobs:
           path: artifacts
       - name: Create release bundle with archives for all builds
         run: |
-          find artifacts -type f -iname postgrest -exec chmod +x {} \; 
+          find artifacts -type f -iname postgrest -exec chmod +x {} \;
 
           mkdir -p release-bundle
 
--- a/nix/tools/tests.nix
+++ b/nix/tools/tests.nix
@@ -24,7 +24,37 @@ let
         withEnv = postgrest.env;
       }
       ''
-        ${withTools.latest} ${cabal-install}/bin/cabal v2-test ${devCabalOptions}
+        ${withTools.latest} ${cabal-install}/bin/cabal v2-run ${devCabalOptions} test:spec
+      '';
+
+  testQuerycost =
+    checkedShellScript
+      {
+        name = "postgrest-test-querycost";
+        docs = "Run the Haskell test suite for query costs";
+        inRootDir = true;
+        withEnv = postgrest.env;
+      }
+      ''
+        ${withTools.latest} ${cabal-install}/bin/cabal v2-run ${devCabalOptions} test:querycost
+      '';
+
+  testDoctests =
+    checkedShellScript
+      {
+        name = "postgrest-test-doctests";
+        docs = "Run the Haskell doctest test suite";
+        inRootDir = true;
+        withEnv = postgrest.env;
+      }
+      ''
+        # For unknown reasons, doctests uses the wrong GHC package database outside
+        # nix-shell and fails, so we set the package path explicitly
+        #ghcWithPackages="$(cat ${postgrest.env})"
+        #ghcVersion="$(ls "$ghcWithPackages/lib")"
+        #export GHC_PACKAGE_PATH="$ghcWithPackages/lib/$ghcVersion/package.conf.d/"
+
+        ${cabal-install}/bin/cabal v2-run ${devCabalOptions} test:doctests
       '';
 
   testSpecIdempotence =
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -67,6 +67,14 @@ jobs:
         if: always()
         run: postgrest-with-postgresql-9.5 postgrest-test-spec
 
+      - name: Run query cost tests against all PostgreSQL versions
+        if: always()
+        run: postgrest-with-all postgrest-test-spec-querycost
+
+      - name: Run doctests
+        if: always()
+        run: nix-shell --run postgrest-test-doctests
```
wolfgangwalther, line 76:
```suggestion
        run: postgrest-test-doctests
```

Any reason why not like this?

nix/tools/tests.nix
```diff
@@ -24,7 +24,31 @@ let
...
       }
       ''
-        ${withTools.latest} ${cabal-install}/bin/cabal v2-test ${devCabalOptions}
+        ${withTools.latest} ${cabal-install}/bin/cabal v2-run ${devCabalOptions} test:spec
+      '';
+
+  testSpecQuerycost =
+    checkedShellScript
+      {
+        name = "postgrest-test-spec-querycost";
```
wolfgangwalther, line 33:
```suggestion
  testQuerycost =
    checkedShellScript
      {
        name = "postgrest-test-querycost";
```

Allows for better autocompletion, because you only have to do continue with `s`, `q` or `d` at `postgrest-test-` once.

Maybe change that everywhere for consistency.

nix/tools/tests.nix
```diff
@@ -105,14 +129,18 @@ let
...
-        ${withTools.latest} ${cabal-install}/bin/cabal v2-test ${devCabalOptions}
+          ${withTools.latest} ${cabal-install}/bin/cabal v2-run ${devCabalOptions} test:spec
+
+        HPCTIXFILE="$tmpdir"/spec-querycost.tix \
+          ${withTools.latest} ${cabal-install}/bin/cabal v2-run ${devCabalOptions} test:spec-querycost
 
         # collect all the tix files
-        ${ghc}/bin/hpc sum  --union --exclude=Paths_postgrest --output="$tmpdir"/tests.tix "$tmpdir"/io*.tix "$tmpdir"/spec.tix
+        ${ghc}/bin/hpc sum  --union --exclude=Paths_postgrest --output="$tmpdir"/tests.tix \
+          "$tmpdir"/io*.tix "$tmpdir"/spec.tix "$tmpdir"/spec-querycost.tix
```
wolfgangwalther, line 143: Should we add the doctests to coverage collection, too?

I see you opted for not including them right now, any specific reasons to do so?

src/PostgREST/Request/Preferences.hs
```diff
@@ -1,54 +1,196 @@
...
+-- | Preferences recognized by the application.
+data Preferences
+  = Preferences
+    { preferResolution     :: Maybe PreferResolution
+    , preferRepresentation :: Maybe PreferRepresentation
+    , preferParameters     :: Maybe PreferParameters
+    , preferCount          :: Maybe PreferCount
+    , preferTransaction    :: Maybe PreferTransaction
+    }
+    deriving Show
```
wolfgangwalther, line 40: Hm, this line is uncovered and only needed for the doctests. That would be one reason to include the doctests in coverage, otherwise we'd have false reports all over the place again.

.github/workflows/ci.yaml
monacoremo: I couldn't get it work properly outside nix-shell yet... It fails because ghci somehow picks up the wrong GHC packages. It works for me locally by setting GHC_PACKAGE_PATH explicitly, but then that fails in CI... need to investigate

nix/tools/tests.nix
monacoremo: yes, makes sense! fixed

nix/tools/tests.nix
monacoremo: Bad news: https://downloads.haskell.org/~ghc/8.10.7/docs/html/users_guide/profiling.html#caveats-and-shortcomings-of-haskell-program-coverage

GHCi does not support hpc, and doctest is built on GHCi - so we won't be able to get coverage from `doctest` in its current incarnation

src/PostgREST/Request/Preferences.hs
monacoremo: See above on why we can't get coverage for doctests at this point. At least we can avoid those unneeded derivations in the actual code, see latest commit

wolfgangwalther: Yes, I like that!

monacoremo:
@steve-chavez decision needed!

This works in principle, with some positives and negatives. 

Positives:

* It works. We have a `postgrest-test-doctest` command and it's integrated in CI
* Doctests are quite easy to write and make for great in-module documentation  (see example in the Preferences module)
* Using doctests drives us to create clean, usable module APIs (as would unit tests, but not the integration tests we have)
* Added complexity is quite low - essentially a section in `postgrest.cabal` and a `Main.hs` file. There is no impact on other tests
* We get extra safety through the additional tests

Negatives:

* We won't get coverage out of the doctests, probably ever - this is an inherent limitation of the GHCi based implementation and the fact that GHCi does not support HPC. At least, we will not worsen coverage
* Running the doctests outside nix-shell right now is flaky and I'm not quite sure why. This should be fixable later, in the meantime it takes about 30s extra time in CI

Should we still merge this? I think the benefits outweigh the negatives and recommend yes.

steve-chavez:
> Should we still merge this? I think the benefits outweigh the negatives and recommend yes.

Sure! I'm all for improving code quality.

steve-chavez (approved):

PR #2025, opened 2021-11-17
nix(feat): Add actionlint to postgrest-lint

wolfgangwalther:
I found `actionlint` yesterday. Thought we could add it. Of course, we have @remo, whose github workflows are flawless - actionlint reports no problems, so far! But just to make sure we don't make things worse in the future, we can still add it..

### DIFF
--- a/nix/tools/style.nix
+++ b/nix/tools/style.nix
@@ -1,4 +1,5 @@
-{ black
+{ actionlint
+, black
 , buildToolbox
 , checkedShellScript
 , git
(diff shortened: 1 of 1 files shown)


Remo:
Thanks for complimenting my flawless workflows, but I think you might have meant @monacoremo 😄

wolfgangwalther:
> Thanks for complimenting my flawless workflows, but I think you might have meant @monacoremo smile

Right... :D

monacoremo:
Great addition!

Issue #1998, opened 2021-10-28
`is` operator not fully parametrized?


wolfgangwalther:
As mentioned in https://github.com/PostgREST/postgrest/issues/1992#issuecomment-953535851, tracking here for better visibility:

> We are still depending on pgFmtLit in the fallback case for the is. operator. That means that this part of the query is not parametrized and in theory still up for sql injection attacks. @steve-chavez should we just disallow other values than NULL, FALSE and TRUE for is. altogether, instead of falling back to passing the value as ::unknown?

steve-chavez:
`IS` cannot be prepared:

```sql
PREPARE boolplan AS SELECT * FROM projects where id IS $1

ERROR:  42601: syntax error at or near "$1"
LINE 1: PREPARE boolplan AS SELECT * FROM projects where id IS $1;
```
This is commented here:

https://github.com/PostgREST/postgrest/blob/a15f988c33b067c283b1e7fa36d9a7add0fa7439/src/PostgREST/Query/SqlFragment.hs#L245-L251

> should we just disallow other values than NULL, FALSE and TRUE for is

Hm, I don't recall if `pgFmtLit` was needed(maybe a test failed) but if not, only allowing NULL, FALSE, TRUE sounds good.

wolfgangwalther:
> only allowing NULL, FALSE, TRUE sounds good.

For completeness, we could also add `UNKNOWN`: https://www.postgresql.org/docs/current/functions-comparison.html#FUNCTIONS-COMPARISON-PRED-TABLE

Although that might be cosmetic only:

> Notice that IS UNKNOWN and IS NOT UNKNOWN are effectively the same as IS NULL and IS NOT NULL, respectively, except that the input expression must be of Boolean type.

steve-chavez:
> For completeness, we could also add UNKNOWN

That one seems nice for expressing three-valued logic(true, false, unknown). I'll add it.

PR #2027, opened 2021-11-18
Allow unknown for is operator

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/1998.

- feat: allow unknown on the is operator 

  Also make the IS values strict at the Parser level since IS cannot be parametrized.

- refactor: remove pgFmtLit

  This is so there's no more logic related to escaping in the codebase.

### DIFF
--- a/src/PostgREST/Request/Types.hs
+++ b/src/PostgREST/Request/Types.hs
@@ -31,6 +31,7 @@ module PostgREST.Request.Types
   , ReadRequest
   , SelectItem
   , SingleVal
+  , TrileanVal(..)
   , fstFieldNames
   ) where
 
@@ -209,6 +210,7 @@ data OpExpr =
 data Operation
   = Op Operator SingleVal
   | In ListVal
+  | Is TrileanVal
   | Fts Operator (Maybe Language) SingleVal
   deriving (Eq)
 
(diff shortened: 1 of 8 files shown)


src/PostgREST/Request/Types.hs
```diff
@@ -220,3 +222,11 @@ type SingleVal = Text
...
 -- | Represents a list value in a filter, e.g. id=in.(val1,val2,val3)
 type ListVal = [Text]
+
+-- | Three-valued logic values
+data TrileanVal
+  = TriTrue
+  | TriFalse
+  | TriNull
+  | TriUnknown
+  deriving Eq
```
wolfgangwalther, line 232:
```suggestion
```

steve-chavez: It gives an error if removed:

```
src/PostgREST/Request/Types.hs:215:13: error:
    • No instance for (Eq TrileanVal)
        arising from the first field of ‘Is’ (type ‘TrileanVal’)
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (Eq Operation)
    |
215 |   deriving (Eq)
```

wolfgangwalther: Ah, I was mislead by the coverage report saying it was unused. I remember reading something about coverage in that case and how the report can be misleading. Might be able to find it later.

laurenceisla (approved):

PR #2030, opened 2021-11-19
refactor: Carve out QueryParams module

monacoremo:
What module is in charge of parsing the query string `?id=eq.1&select=name`? Currently, that concern is spread across many modules, including:

* `ApiRequest`
* `Parsers`
* `DbRequestBuilder`
* `SqlFragment`

This PR carves out one coherent module that is in charge of parsing the query string. It exposes one function `parse` that is also easy to use interactively. Additional type safety is introduced by defining types for the filter operators, which also decouples the query string parsing from SqlFragment.

### DIFF
--- a/src/PostgREST/Request/ApiRequest.hs
+++ b/src/PostgREST/Request/ApiRequest.hs
@@ -55,7 +55,8 @@ import PostgREST.Request.Preferences     (PreferCount (..),
                                           PreferParameters (..),
                                           PreferRepresentation (..),
                                           PreferResolution (..),
-                                          PreferTransaction (..))
+                                          PreferTransaction (..),
+                                          Preferences (..))
 import PostgREST.Request.QueryParams     (QueryParams (..))
 import PostgREST.Request.Types           (ApiRequestError (..))
 
@@ -169,11 +170,19 @@ data ApiRequest = ApiRequest {
 
 -- | Examines HTTP request and translates it into user intent.
 userApiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> Either ApiRequestError ApiRequest
-userApiRequest conf dbStructure req reqBody =
-  apiRequest conf dbStructure req reqBody =<< first QueryParamError (QueryParams.parse (rawQueryString req))
+userApiRequest conf dbStructure req reqBody = do
+  queryParams <- first QueryParamError . parseQueryParams $ rawQueryString req
+  apiRequest conf dbStructure req reqBody preferences queryParams
+  where
+    preferences = Preferences.fromHeaders $ requestHeaders req
+    isUpsert =
+      case (requestMethod req, preferResolution preferences) of
+        ("POST", Just _) -> True
+        _                -> False
+    parseQueryParams = if isUpsert then QueryParams.parseUpsert else QueryParams.parse
 
-apiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> QueryParams.QueryParams -> Either ApiRequestError ApiRequest
-apiRequest conf@AppConfig{..} dbStructure req reqBody queryparams@QueryParams{..}
+apiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> Preferences -> QueryParams.QueryParams -> Either ApiRequestError ApiRequest
+apiRequest conf@AppConfig{..} dbStructure req reqBody Preferences{..} queryparams@QueryParams{..}
   | isJust profile && fromJust profile `notElem` configDbSchemas = Left $ UnacceptableSchema $ toList configDbSchemas
   | isTargetingProc && method `notElem` ["HEAD", "GET", "POST"] = Left ActionInappropriate
   | topLevelRange == emptyRange = Left InvalidRange
@@ -323,7 +332,6 @@ apiRequest conf@AppConfig{..} dbStructure req reqBody queryparams@QueryParams{..
   method          = requestMethod req
   hdrs            = requestHeaders req
   lookupHeader    = flip lookup hdrs
-  Preferences.Preferences{..} = Preferences.fromHeaders hdrs
   auth = fromMaybe "" $ lookupHeader hAuthorization
   tokenStr = case T.split (== ' ') (T.decodeUtf8 auth) of
     ("Bearer" : t : _) -> t
--- a/src/PostgREST/Request/QueryParams.hs
+++ b/src/PostgREST/Request/QueryParams.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE TupleSections #-}
 module PostgREST.Request.QueryParams
   ( parse
+  , parseUpsert
   , QueryParams(..)
   , pRequestRange
   ) where
@@ -124,7 +125,14 @@ data QueryParams =
 -- >>> qsFilters <$> parse "a.b=noop.0"
 -- Left (QPError "\"failed to parse filter (noop.0)\" (line 1, column 6)" "unknown single value operator noop")
 parse :: ByteString -> Either QPError QueryParams
-parse qs =
+parse = parseWith False
+
+-- | Parse query string for upserts with on_conflict as reserved keyword.
+parseUpsert :: ByteString -> Either QPError QueryParams
+parseUpsert = parseWith True
+
+parseWith :: Bool -> ByteString -> Either QPError QueryParams
+parseWith isUpsert qs =
   QueryParams
     canonical
     params
@@ -137,7 +145,7 @@ parse qs =
     <*> pRequestFilter `traverse` filtersRoot
     <*> pRequestFilter `traverse` filtersNotRoot
     <*> pure (S.fromList (fst <$> filters))
-    <*> sequenceA (pRequestOnConflict <$> onConflict)
+    <*> if isUpsert then sequenceA (pRequestOnConflict <$> onConflict) else Right Nothing
   where
     logic = filter (endingIn ["and", "or"] . fst) nonemptyParams
     select = fromMaybe "*" $ lookupParam "select"
@@ -170,7 +178,7 @@ parse qs =
 
     filtersAndParams = filter (isFilterOrParam . fst)  nonemptyParams
     isFilterOrParam k = not (endingIn reservedEmbeddable k) && notElem k reserved
-    reserved = ["select", "columns", "on_conflict"]
+    reserved = ["select", "columns"] ++ ["on_conflict" | isUpsert]
     reservedEmbeddable = ["order", "limit", "offset", "and", "or"]
 
     (filtersNotRoot, filtersRoot) = L.partition isNotRoot filters
--- a/src/PostgREST/Request/QueryParams.hs
+++ b/src/PostgREST/Request/QueryParams.hs
@@ -52,8 +52,9 @@ import PostgREST.Request.Types (EmbedParam (..), EmbedPath, Field,
                                 OpExpr (..), Operation (..),
                                 OrderDirection (..), OrderNulls (..),
                                 OrderTerm (..), QPError (..),
-                                SelectItem, SimpleOperator (..),
-                                SingleVal, TrileanVal (..))
+                                SelectItem, SingleVal,
+                                SingleValOperator (..),
+                                TrileanVal (..))
 
 import Protolude hiding (try)
 
@@ -64,7 +65,7 @@ import Protolude hiding (try)
 -- >>> deriving instance Show QPError
 -- >>> deriving instance Show TrileanVal
 -- >>> deriving instance Show FtsOperator
--- >>> deriving instance Show SimpleOperator
+-- >>> deriving instance Show SingleValOperator
 -- >>> deriving instance Show Operation
 -- >>> deriving instance Show OpExpr
 -- >>> deriving instance Show JsonOperand
@@ -206,7 +207,7 @@ parse qs =
         offsetParams =
           M.fromList [(k, maybe allRange rangeGeq (readMaybe $ toS v)) | (k,v) <- offsets]
 
-operator :: Text -> Maybe SimpleOperator
+operator :: Text -> Maybe SingleValOperator
 operator = \case
   "eq"    -> Just OpEqual
   "gte"   -> Just OpGreaterThanEqual
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -59,7 +59,7 @@ import PostgREST.Request.Types           (Alias, Field, Filter (..),
                                           OrderDirection (..),
                                           OrderNulls (..),
                                           OrderTerm (..), SelectItem,
-                                          SimpleOperator (..),
+                                          SingleValOperator (..),
                                           TrileanVal (..))
 
 import Protolude hiding (cast)
@@ -74,7 +74,7 @@ noLocationF = "array[]::text[]"
 sourceCTEName :: SqlFragment
 sourceCTEName = "pgrst_source"
 
-singleValOperator :: SimpleOperator -> SqlFragment
+singleValOperator :: SingleValOperator -> SqlFragment
 singleValOperator = \case
   OpEqual            -> "="
   OpGreaterThanEqual -> ">="
--- a/src/PostgREST/Request/Types.hs
+++ b/src/PostgREST/Request/Types.hs
@@ -35,7 +35,7 @@ module PostgREST.Request.Types
   , SingleVal
   , TrileanVal(..)
   , fstFieldNames
-  , SimpleOperator(..)
+  , SingleValOperator(..)
   , FtsOperator(..)
   ) where
 
@@ -232,7 +232,7 @@ data OpExpr =
   deriving (Eq)
 
 data Operation
-  = Op SimpleOperator SingleVal
+  = Op SingleValOperator SingleVal
   | In ListVal
   | Is TrileanVal
   | Fts FtsOperator (Maybe Language) SingleVal
@@ -254,7 +254,7 @@ data TrileanVal
   | TriUnknown
   deriving Eq
 
-data SimpleOperator
+data SingleValOperator
   = OpEqual
   | OpGreaterThanEqual
   | OpGreaterThan


monacoremo:
There still are a few rough edges:

* The doctests are proof of concept examples, a lot more could be added
* Many things could be optimized within QueryParams, e.g. factoring out the duplicate ways to identify filter params
* Introduced a separate Error type QueryParamError that is only roughly integrated with the existing error type
* In ApiRequest, we have `userApiRequest` and `userApiRequest'`

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,484 @@
...
+      where lastWord = L.last $ T.split (== '.') key
+
+    (filters, params) = L.partition isParam filtersAndParams
+    isParam (k, v) = isEmbedPath k || hasOperator v || hasFtsOperator v
+
+    filtersAndParams = filter (isFilterOrParam . fst)  nonemptyParams
+    isFilterOrParam k = not (endingIn reservedEmbeddable k) && notElem k reserved
+    reserved = ["select", "columns"]
+    -- TODO: check potential previous bug, on_conflict was missing
+    reservedEmbeddable = ["order", "limit", "offset", "and", "or", "on_conflict"]
```
monacoremo, line 169: `on_conflict` was missing orginally - not sure if that would have created issues previously, but I don't think it hurts to add it here

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,488 @@
...
+      fld <- pField
+      cast' <- optionMaybe (string "::" *> many letter)
+      return (fld, toS <$> cast', alias, Nothing, Nothing)
+  )
+  <|> do
+    s <- pStar
+    return ((s, []), Nothing, Nothing, Nothing, Nothing)
+
+pOpExpr :: Parser SingleVal -> Parser OpExpr
+pOpExpr pSVal = try ( string "not" *> pDelimiter *> (OpExpr True <$> pOperation)) <|> OpExpr False <$> pOperation
```
wolfgangwalther, line 381: There is some refactoring going on in `pOpExpr` already - could this be moved to a second commit, at least? That should make it much easier to review

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -180,12 +168,18 @@ data ApiRequest = ApiRequest {
...
 userApiRequest :: AppConfig -> DbStructure -> Request -> RequestBody -> Either ApiRequestError ApiRequest
-userApiRequest conf@AppConfig{..} dbStructure req reqBody
+userApiRequest conf dbStructure req reqBody =
+  userApiRequest' conf dbStructure req reqBody =<< first QPError (QueryParams.parse (rawQueryString req))
+
+userApiRequest' :: AppConfig -> DbStructure -> Request -> RequestBody -> QueryParams.QueryParams -> Either ApiRequestError ApiRequest
+userApiRequest' conf@AppConfig{..} dbStructure req reqBody querystring
+  | isTargetingProc && method == "POST" && not (L.null qsParams) = Left $ ParseRequestError "" ""
+  | not isTargetingProc && not (L.null qsParams) =
+      Left $ ParseRequestError "Unexpected param or filter missing operator" ("Failed to parse " <> show qsParams)
```
wolfgangwalther, line 178: This actually looks like a feature "Improved error messages" for those cases. Although I don't understand why we have `ParseRequestError "" ""` in there - is this error actually reported somewhere?

Not sure whether it would be possible to split those changes into a separate commit, together with the changed test output, because this is not a refactor anymore.

wolfgangwalther:
Along the lines of the other two comments regarding separate commits: Would it be possible to split the changes in `DbRequestBuilder.hs` into a separate commit? It looks like there's some more refactoring/simplification going on, which is not really related to `QueryParams`. I.e. only make the changes neccessary for the `QueryParams` module in that commit and have another commit with the remaining refactors?

src/PostgREST/Request/QueryParams.hs
wolfgangwalther: Should this not be added to `reserved` one line above that instead?

monacoremo: yes, good point - it does not apply to embedded fields. Fixed!

src/PostgREST/Request/ApiRequest.hs
monacoremo: This is difficult to split from the rest of the refactor, as catching this error case here (params provided although no params are expected as no proc is being called) significantly simplifies the remaining logic of parsing the query string. Providing a better error message is a side effect that is hard to avoid in this case.

src/PostgREST/Request/QueryParams.hs
monacoremo: Ah, I missed that - this PR is already based on https://github.com/PostgREST/postgrest/pull/2027, are there any other changes in progress?

The type safety was useful to ensure the logic is sound and tests pass, the Map-based approach was very brittle.

wolfgangwalther: Ah no, I meant to say, that you've been refactoring `pOpExpr` already a bit in this PR. But it's hard to review, because we don't get a good diff because of moving big chunks around. So if you could move the parsers to this file, mostly unchanged (only to make them work in this context) - and then do those refactors in a second commit, the diff would be much easier to understand.

monacoremo:
Many thanks for reviewing @wolfgangwalther !

> Along the lines of the other two comments regarding separate commits: Would it be possible to split the changes in `DbRequestBuilder.hs` into a separate commit? It looks like there's some more refactoring/simplification going on, which is not really related to `QueryParams`. I.e. only make the changes neccessary for the `QueryParams` module in that commit and have another commit with the remaining refactors?

I kept most of the intermediate steps here: https://github.com/monacoremo/postgrest/tree/refactorquerystring - maybe this is a bit easier to follow?

The changes in DbRequestBuilder are difficult to seperate out, as pulling the query parsing bits out of that module is the core of this PR. The simplifications result from the additional type safety that QueryParams provides, making many edge-cases in DbRequestBuilder impossible. This PR also gets rid of many uses of partial functions like `fromJust` on that basis.

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,488 @@
...
+  where
+    treePath = P.parse pTreePath ("failed to parse tree path (" ++ toS k ++ ")") $ toS k
+    oper = P.parse (pOpExpr pSingleVal) ("failed to parse filter (" ++ toS v ++ ")") $ toS v
+    path = fst <$> treePath
+    fld = snd <$> treePath
+
+pRequestOrder :: (Text, Text) -> Either QueryParamError (EmbedPath, [OrderTerm])
+pRequestOrder (k, v) = mapError $ (,) <$> path <*> ord'
+  where
+    treePath = P.parse pTreePath ("failed to parser tree path (" ++ toS k ++ ")") $ toS k
```
wolfgangwalther, line 260:
```suggestion
    treePath = P.parse pTreePath ("failed to parse tree path (" ++ toS k ++ ")") $ toS k
```

There are a few more below.

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,488 @@
...
+    <*> pure (S.fromList (fst <$> filters))
+    <*> sequenceA (pRequestOnConflict <$> onConflict)
+  where
+    logic = filter (endingIn ["and", "or"] . fst) nonemptyParams
+    select = fromMaybe "*" $ lookupParam "select"
+    onConflict = lookupParam "on_conflict"
+    columns = lookupParam "columns"
+    order = filter (endingIn ["order"] . fst) nonemptyParams
+    limits = filter (endingIn ["limit"] . fst) nonemptyParams
+    offsets = first (replaceLast "limit") <$> filter (endingIn ["offset"] . fst) nonemptyParams
```
wolfgangwalther, line 148: The `replaceLast "limit"` bit requires some explanation, I think. I see from the commit history in your fork, that you were confused about this, too.

Maybe you have a refactor of this in mind anyway - but should we maybe do something like this in the meantime:

```suggestion
    -- Replace .limit and .offset endings to be able to match those params later in a map
    limits = first (replaceLast "") <$> filter (endingIn ["limit"] . fst) nonemptyParams
    offsets = first (replaceLast "") <$> filter (endingIn ["offset"] . fst) nonemptyParams
```

src/PostgREST/Request/QueryParams.hs
wolfgangwalther: Actually.. I think it might not be a good idea to add this here at all. I didn't test that, yet, but I understand that part of the code to block any RPC from using an argument by the name of `on_conflict`. Since this is not strictly needed, because `on_conflict` is only a thing for `POST` to tables/views, it would actually be a breaking change to add this here - in case somebody is using such an argument name.

If that holds true, it's actually the other way around: I think we can relax things a little bit and not reserve `columns` - because those are not used when calling RPCs either... right?

wolfgangwalther:
> I kept most of the intermediate steps here: https://github.com/monacoremo/postgrest/tree/refactorquerystring - maybe this is a bit easier to follow?

That was certainly helpful. I can't say I really reviewed line by line, but I did indeed go over all the commits in that branch.

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,489 @@
...
+    <*> sequenceA (pRequestOnConflict <$> onConflict)
+  where
+    logic = filter (endingIn ["and", "or"] . fst) nonemptyParams
+    select = fromMaybe "*" $ lookupParam "select"
+    onConflict = lookupParam "on_conflict"
+    columns = lookupParam "columns"
+    order = filter (endingIn ["order"] . fst) nonemptyParams
+    -- Replace .limit and .offset endings to be able to match those params later in a map
+    limits = first (replaceLast "") <$> filter (endingIn ["limit"] . fst) nonemptyParams
+    offsets = first (replaceLast "") <$> filter (endingIn ["offset"] . fst) nonemptyParams
```
wolfgangwalther, line 149:
```suggestion
    limits = first (replaceLast "range") <$> filter (endingIn ["limit"] . fst) nonemptyParams
    offsets = first (replaceLast "range") <$> filter (endingIn ["offset"] . fst) nonemptyParams
```

Seems like CI failed with the empty string - maybe like this, instead?

src/PostgREST/Request/QueryParams.hs
monacoremo: Ah, yes, that makes sense! What I did now: First commit is the refactoring but the `QueryParams` module is still called `Parsers` - this creates a legible diff for what happended with the parsers. In a second commit I did the renaming.

src/PostgREST/Request/QueryParams.hs
monacoremo: Hm, there seems to be some particular dependency on this being renamed `limit`, tests fail otherwise. I'll add a comment and we'll investigate in a separate PR

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,489 @@
...
+pRequestOrder (k, v) = mapError $ (,) <$> path <*> ord'
+  where
+    treePath = P.parse pTreePath ("failed to parse tree path (" ++ toS k ++ ")") $ toS k
+    path = fst <$> treePath
+    ord' = P.parse pOrder ("failed to parse order (" ++ toS v ++ ")") $ toS v
+
+pRequestRange :: (Text, NonnegRange) -> Either QueryParamError (EmbedPath, NonnegRange)
+pRequestRange (k, v) = mapError $ (,) <$> path <*> pure v
+  where
+    treePath = P.parse pTreePath ("failed to parser tree path (" ++ toS k ++ ")") $ toS k
```
wolfgangwalther, line 268:
```suggestion
    treePath = P.parse pTreePath ("failed to parse tree path (" ++ toS k ++ ")") $ toS k
```

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,489 @@
...
+    path = fst <$> treePath
+
+pRequestLogicTree :: (Text, Text) -> Either QueryParamError (EmbedPath, LogicTree)
+pRequestLogicTree (k, v) = mapError $ (,) <$> embedPath <*> logicTree
+  where
+    path = P.parse pLogicPath ("failed to parser logic path (" ++ toS k ++ ")") $ toS k
```
wolfgangwalther, line 274:
```suggestion
    path = P.parse pLogicPath ("failed to parse logic path (" ++ toS k ++ ")") $ toS k
```

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,489 @@
...
+  )
+  <|> do
+    s <- pStar
+    return ((s, []), Nothing, Nothing, Nothing, Nothing)
+
+pOpExpr :: Parser SingleVal -> Parser OpExpr
+pOpExpr pSVal = try ( string "not" *> pDelimiter *> (OpExpr True <$> pOperation)) <|> OpExpr False <$> pOperation
+  where
+    pOperation :: Parser Operation
+    pOperation = pIn <|> pIs <|> try pFts <|> pOp <?> "operator (eq, gt, ...)"
```
wolfgangwalther, line 385: Any special reason why pOp is not first anymore, but much later? Does it make a difference? I don't really understand the whole parsing that much, yet, so..

src/PostgREST/Request/QueryParams.hs
```diff
@@ -0,0 +1,489 @@
...
+  "cd"    -> Just OpContained
+  "ov"    -> Just OpOverlap
+  "sl"    -> Just OpStrictlyLeft
+  "sr"    -> Just OpStrictlyRight
+  "nxr"   -> Just OpNotExtendsRight
+  "nxl"   -> Just OpNotExtendsLeft
+  "adj"   -> Just OpAdjacent
+  _       -> Nothing
+
+ftsOperator :: Text -> Maybe FtsOperator
```
wolfgangwalther, line 230: I think the naming of `SingleValOperator` vs `FtsOperator` could be improved. They both really only have a single value, but one has arguments, the other does not, right?

What about:
- `Operator` vs `OperatorWithArg` or
- `OperatorWithoutArg` vs `OperatorWithArg` or
- `SimpleOperator` vs `OperatorWithArg`

This naming thing probably extends to the parser, too, because it's not really relevant that those operators are "fts" or not.

src/PostgREST/Request/QueryParams.hs
monacoremo: It works in any order, e.g. `pOperation = try pOp <|> pIn <|> pIs <|> try pFts <?> "operator (eq, gt, ...)"` also passes. It seemed most sensible to put the most complex parser last.

src/PostgREST/Request/QueryParams.hs
monacoremo: Ah, but it does matter: The full text search operators have an optional language parameter, the single val operators do not. I followed the naming of the existing parsers here, would not change it in this PR.

src/PostgREST/Request/QueryParams.hs
monacoremo: I think it's quite confusing to only have the parameters be reserved in special cases... But let's not change the behavior here, I've added logic to handle the current 'on_conflict' behavior explicitly - let's see if we want to keep it in a separate PR.

monacoremo:
@wolfgangwalther There is so much more to be done in ApiRequest now that QueryParams is carved out (e.g. all the simplification that would fall out of eliminating `PathUnknown`and `TargetUnknown` and the current way there seem to be a circular dependency between action -> path -> schema -> profile -> action which makes my head hurt :-) ), but I suggest we make a cut here an commit this PR as is for now. WDYT?

src/PostgREST/Request/QueryParams.hs
wolfgangwalther: Mh. I think I didn't make my point clear enough. Maybe this code snippet tells the story a bit better:

```haskell
data Operation
  = Op SingleValOperator SingleVal
  | In ListVal
  | Is TrileanVal
  | Fts FtsOperator (Maybe Language) SingleVal
```

Both `SingleValOperator` and `FtsOperator` take a `SingleVal`. That means the `SingleVal` part in `SingleValOperator` is not really describing it well.

I agree with `FtsOperator` - that name was there before already. But `SingleValOperator` was not. It was called `Operator` before. Maybe just rename it to `SimpleOperator` or so - I think that's what you meant to say.

src/PostgREST/Request/QueryParams.hs
wolfgangwalther: I just tested - and `POST`- upsert does not even support filters at all. This is because `POST` will never add a `WHERE` clause:

https://github.com/PostgREST/postgrest/blob/74f6e5225cc2d4c3b8b13bb99a80224592240644/src/PostgREST/Query/QueryBuilder.hs#L89-L90

So this is actually a non-issue. I think you shouldn't add that logic here.

wolfgangwalther:
> @wolfgangwalther There is so much more to be done in ApiRequest now that QueryParams is carved out (e.g. all the simplification that would fall out of eliminating `PathUnknown`and `TargetUnknown` and the current way there seem to be a circular dependency between action -> path -> schema -> profile -> action which makes my head hurt :-) ), but I suggest we make a cut here an commit this PR as is for now. WDYT?

Absolutely. There's still a lot of stuff, but that doesn't need to be done in this PR. Just need to make sure that the new things we are introducing here are well designed. The two issues about `on_conflict` and `SingleValOperator` are still open for me - see comments above.

wolfgangwalther:
To avoid this thing getting outdated, I rebased this on latest main, and merged most of the commits together. I also changed `SingleValOperator` to `SimpleOperator` - but added another commit reverting this change. This is just to clearly demonstrate what I changed - I changed nothing else.

This leaves us with 3 commits:
- The main refactor commit. I agree with merging this.
- The "revert my change of naming". I suggest to drop this, because I think `SimpleOperator` is better. I still left the commit here, to be able to discuss it, if you'd like.
- The commit you added regarding the special case for `on_conflict`. I suggest to drop that, too - with my reasoning mentioned above.

Since we pretty much agree on the first commit, I will merge that into main once CI passes - and keep the other two here until I hear from you!

steve-chavez:
Hm, the QueryParams module has being in main for a while so closing this one.

PR #2031, opened 2021-11-20
Include valid examples in the hint for ambiguous embedding

laurenceisla:
The example given when an Ambiguous Embedding error is thrown shows a generic example `/origin?select=relationship(*)` or `/origin?select=target!relationship(*)`.

This PR adds examples with the real  `target!relationship` names relevant to the query, omitting the `origin` and the `select` clause. 

Edit: Also adds changes to other parts of the error to make it clearer to the user.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -98,7 +98,7 @@ instance JSON.ToJSON ApiRequestError where
     "message" .= ("Could not find a relationship between " <> parent <> " and " <> child <> " in the schema cache" :: Text)]
   toJSON (AmbiguousRelBetween parent child rels) = JSON.object [
     "hint"    .= ("Try changing '" <> child <> "' to one of the following: " <> relHint rels <> ". Find the desired relationship in the 'details' key." :: Text),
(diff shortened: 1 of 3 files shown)


src/PostgREST/Error.hs
```diff
@@ -97,7 +97,7 @@ instance JSON.ToJSON ApiRequestError where
     "hint"    .= ("If a new foreign key between these entities was created in the database, try reloading the schema cache." :: Text),
     "message" .= ("Could not find a relationship between " <> parent <> " and " <> child <> " in the schema cache" :: Text)]
   toJSON (AmbiguousRelBetween parent child rels) = JSON.object [
-    "hint"    .= ("By following the 'details' key, disambiguate the request by changing the url to /origin?select=relationship(*) or /origin?select=target!relationship(*)" :: Text),
+    "hint"    .= ("According to the relationship needed, try changing the value in the query string to one of the following: " <> relHint rels :: Text),
```
steve-chavez, line 100:
```suggestion
    "hint"    .= ("Try changing '" <> child <> "' to one of the following: " <> relHint rels <> ". Follow the 'details' key in the error message to confirm you are specifying your desired relationship.' :: Text),
```

steve-chavez (approved):

PR #2032, opened 2021-11-21
ci: Refactor workflows

wolfgangwalther:
I wanted to get used to GitHub Actions a bit more and played with our workflow. To reduce some of the repetition we currently have, I added a `setup-nix` action and used a matrix strategy to build with stack on different OS. This doesn't really reduce the amount of code, but I feel it's a bit quicker to see which parts are the same between different jobs and which parts are varied.

I also tried two more things, but failed so far:
- Reorganizing the big workflow into multiple files with the "reusable workflows" feature, doesn't really work, yet, because this does not support calling workflow files in the same repository, yet. This might be possible in the future.
- Running the spec tests on Windows requires to have `with_tmp_db` run on Windows. I can see that as a possibility and also made a bit of progress, but gave up on it for now.

### DIFF
--- a/.github/actions/setup-nix/action.yaml
+++ b/.github/actions/setup-nix/action.yaml
@@ -1,0 +1,21 @@
+name: Setup Nix
+
+description: Installs nix, sets up cachix and installs a subset of tooling.
+
+inputs:
+  authToken:
+    description: Token to pass to cachix
+  tools:
+    description: Tools to install with nix-env -iA <tools>
+
+runs:
+  using: composite
+  steps:
+    - uses: cachix/install-nix-action@v13
+    - uses: cachix/cachix-action@v10
+      with:
+        name: postgrest
+        authToken: ${{ inputs.authToken }}
+    - if: ${{ inputs.tools }}
+      run: nix-env -f default.nix -iA ${{ inputs.tools }}
+      shell: bash
--- a/.github/actions/setup-nix/action.yaml
+++ b/.github/actions/setup-nix/action.yaml
@@ -11,7 +11,7 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -282,9 +273,11 @@ jobs:
           path: CHANGES.md
           if-no-files-found: error
 
+
   Release-GitHub:
     name: Release on GitHub
-    permissions: write-all
+    permissions:
+      contents: write
```
wolfgangwalther, line 280: According to https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-contents this should be enough to write to the release endpoint.

monacoremo:
Great stuff! Overall it adds a few more lines than it saves, but I like that this ensures consistency!

.github/actions/setup-nix/action.yaml
```diff
@@ -0,0 +1,21 @@
...
+inputs:
+  authToken:
+    description: Token to pass to cachix
+  tools:
+    description: Tools to install with nix-env -iA <tools>
+
+runs:
+  using: composite
+  steps:
+    - uses: cachix/install-nix-action@v13
```
monacoremo, line 14: This might be a good opportunity to upgrade to https://github.com/cachix/install-nix-action/releases/tag/v16

monacoremo (approved):

.github/actions/setup-nix/action.yaml
wolfgangwalther: Yeah, good point. Did so now. Unfortunately I still can't run the workflow locally with `act`, still breaking at the `install-nix-action` step. Probably need to update to latest master of `act` at some point.

Updated the other actions at the same time, which allowed me to remove one of the parameters to the matrix.

PR #2037, opened 2021-11-24
ci: Add dependabot config for auto updates of github actions

wolfgangwalther:
Now, that we have switched to GitHub Actions and are using actions from other sources, it would make life easier if dependabot gave us automatic PRs for updates to those actions.

This is the config for that - but dependabot needs to be enabled at the same time in the repo settings, for which I don't have access. @steve-chavez

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly


steve-chavez (approved):
These 2 options are enabled now:

> Dependabot alerts
> Receive alerts of new vulnerabilities that affect your dependencies.

> Dependabot security updates
> Easily upgrade to non-vulnerable dependencies.

(Dependabot alerts was already enabled)

PR #2038, opened 2021-11-24
Drop support for PostgreSQL 9.5

wolfgangwalther:
Now that [PG 9.6 is past EOL](https://www.postgresql.org/support/versioning/), it's time to drop support for PG 9.5.

We should still support PG 9.6 for the next release, but maybe we can drop that soon after that, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,15 +112,6 @@ the connection cannot do anything the user themselves couldn't. Other
 forms of authentication can be built on top of the JWT primitive. See
 the docs for more information.
 
-Since PostgreSQL 9.5 supports true [row-level
-security](http://www.postgresql.org/docs/9.5/static/ddl-rowsecurity.html).
-In previous versions it can be simulated with triggers and
-security-barrier views. Because the possible queries to the database
-are limited to certain templates using
-[leakproof](http://blog.2ndquadrant.com/how-do-postgresql-security_barrier-views-work/)
-functions, the trigger workaround does not compromise row-level
-security.
-
 ## Versioning
 
 A robust long-lived API needs the freedom to exist in multiple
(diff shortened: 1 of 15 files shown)


steve-chavez (approved):

steve-chavez:
> We should still support PG 9.6 for the next release, but maybe we can drop that soon after that, too.

Hm, there's value in keeping backwards compat, not sure if we should just follow PostgreSQL EOLs. Users that don't want to migrate will likely stay on an old PostgREST version, which is not good for us.

If the impact on the codebase is low, I'd say we should keep supporting old PostgreSQL versions.

README.md
```diff
@@ -112,15 +112,6 @@ the connection cannot do anything the user themselves couldn't. Other
 forms of authentication can be built on top of the JWT primitive. See
 the docs for more information.
 
-Since PostgreSQL 9.5 supports true [row-level
-security](http://www.postgresql.org/docs/9.5/static/ddl-rowsecurity.html).
-In previous versions it can be simulated with triggers and
-security-barrier views. Because the possible queries to the database
-are limited to certain templates using
-[leakproof](http://blog.2ndquadrant.com/how-do-postgresql-security_barrier-views-work/)
-functions, the trigger workaround does not compromise row-level
```
steve-chavez, line 121: I don't think we have RLS mentioned elsewhere, this paragraph could be included back into the README.

steve-chavez: Nevermind, I think this should be added in the docs instead.

wolfgangwalther: At first I removed all the parts about the trigger workaround, because that's not relevant anymore when we drop pg 9.5. But then.. nothing really remained anymore, except for "PostgreSQL supports true row-level security." :D

Would be great to add more about that in the docs, yes.

PR #2044, opened 2021-11-24
Treat plpgsql assert as invalid input

dwagin:
IMHO: [PL/pgSQL assertions](https://www.postgresql.org/docs/14/plpgsql-errors-and-messages.html#PLPGSQL-STATEMENTS-ASSERT) is invalid input (HTTP 400 Bad request).

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -241,6 +241,7 @@ pgErrorStatus authed (SQL.SessionError (SQL.QueryError _ _ (SQL.ResultError rErr
         'F':'0':_ -> HTTP.status500 -- conf file error
         'H':'V':_ -> HTTP.status500 -- foreign data wrapper error
         "P0001"   -> HTTP.status400 -- default code for "raise"
+        "P0004"   -> HTTP.status400 -- assert failure
         'P':'0':_ -> HTTP.status500 -- PL/pgSQL Error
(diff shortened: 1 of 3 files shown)


wolfgangwalther:
> IMHO: [PL/pgSQL assertions](https://www.postgresql.org/docs/14/plpgsql-errors-and-messages.html#PLPGSQL-STATEMENTS-ASSERT) is invalid input (HTTP 400 Bad request).

Hm. I disagree. I think assertions should be 500 Internal Server Error.

My reasoning: Assertions should be used to make sure your code is passing around the right things internally. They should not be used to check user input. Otherwise disabling assertions wouldn't make much sense - and that's one of the core features of `ASSERT`:

> Testing of assertions can be enabled or disabled via the configuration parameter plpgsql.check_asserts, which takes a Boolean value; the default is on. If this parameter is off then ASSERT statements do nothing.

Or to put it in other words (also from the PG docs):
> Note that ASSERT is meant for detecting program bugs, not for reporting ordinary error conditions. Use the RAISE statement, described above, for that.

So I suggest to change it to 500 - if that wasn't the case before already? Adding a test would still be worth it.

dwagin:
Yes, probably @wolfgangwalther you are right.

wolfgangwalther:
I added the test-case in #2047.

Issue #2009, opened 2021-11-04
Inaccurate total record count returned when top-level filtering with embedded resource filters


dexteradeus:
### Environment

* PostgreSQL version: 9.6.2 (docker image postgres:9.6.2-alpine)
* PostgREST version: 8.0.0.20211102 (docker image postgrest/postgrest:v8.0.0.20211102)
* Operating system: Arch Linux

### Description of issue
When using `Prefer: count=exact` and filtering top-level resources based on embedded resource filters, the total record count shown in the `Content-Range` does not take into account the embedded resource filters. Filters applied to the top level resource itself are still applied, however.

I would expect that when using `count=exact` that the number of total records returned matches the total listed in the `Content-Range` header even when using filters based on embedded resources.

The issue is also present on other `count=` options.

This can be seen using a sample database from https://www.postgresqltutorial.com/postgresql-sample-database/.
The relevant schema is as follows:
```sql
CREATE TABLE public.film (
    film_id integer DEFAULT nextval('public.film_film_id_seq'::regclass) NOT NULL,
    title character varying(255) NOT NULL,
    description text,
    release_year public.year,
    language_id smallint NOT NULL,
    rental_duration smallint DEFAULT 3 NOT NULL,
    rental_rate numeric(4,2) DEFAULT 4.99 NOT NULL,
    length smallint,
    replacement_cost numeric(5,2) DEFAULT 19.99 NOT NULL,
    rating public.mpaa_rating DEFAULT 'G'::public.mpaa_rating,
    last_update timestamp without time zone DEFAULT now() NOT NULL,
    special_features text[],
    fulltext tsvector NOT NULL
);
CREATE TABLE public.category (
    category_id integer DEFAULT nextval('public.category_category_id_seq'::regclass) NOT NULL,
    name character varying(25) NOT NULL,
    last_update timestamp without time zone DEFAULT now() NOT NULL
);
CREATE TABLE public.film_category (
    film_id smallint NOT NULL,
    category_id smallint NOT NULL,
    last_update timestamp without time zone DEFAULT now() NOT NULL
);
ALTER TABLE ONLY public.film_category
    ADD CONSTRAINT film_category_category_id_fkey FOREIGN KEY (category_id) REFERENCES public.category(category_id) ON UPDATE CASCADE ON DELETE RESTRICT;
ALTER TABLE ONLY public.film_category
[...]

PR #1900, opened 2021-07-26
Bumps to v8.0.0

steve-chavez:
Closes #1779.

- Correct/organize the CHANGELOG
- Bumps to v8.0.0

PR #2045, opened 2021-11-25
Bump to v9

steve-chavez:
We're still not PostgreSQL 14 compatible on docker latest, I think it's time to do a new stable release.

@wolfgangwalther @monacoremo @laurenceisla WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Enables uploading raw text to a function with `Content-Type: text/plain`
 - #1938, Allow escaping inside double quotes with a backslash, e.g. `?col=in.("Double\"Quote")`, `?col=in.("Back\\slash")` - @steve-chavez
 - #1075, Allow filtering top-level resource based on embedded resources filters. This is enabled by adding `!inner` to the embedded resource, e.g. `/projects?select=*,clients!inner(*)&clients.id=eq.12`- @steve-chavez, @Iced-Sun
+ - #1857, Make GUC names for headers, cookies and jwt claims compatible with PostgreSQL v14 - @laurenceisla, @robertsosinski
+   + Getting the value for a header GUC on PostgreSQL 14 is done using `current_setting('request.headers')::json->>'name-of-header'` and in a similar way for `request.cookies` and `request.jwt.claims`
+   + PostgreSQL versions below 14 can opt in to the new JSON GUCs by setting the `db-use-legacy-gucs` config option to false (true by default)
 - #1988, Allow specifying `unknown` for the `is` operator - @steve-chavez
 - #2031, Improve error message for ambiguous embedding and add a relevant hint that includes unambiguous embedding suggestions - @laurenceisla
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +21,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Enables uploading raw text to a function with `Content-Type: text/plain`
 - #1938, Allow escaping inside double quotes with a backslash, e.g. `?col=in.("Double\"Quote")`, `?col=in.("Back\\slash")` - @steve-chavez
 - #1075, Allow filtering top-level resource based on embedded resources filters. This is enabled by adding `!inner` to the embedded resource, e.g. `/projects?select=*,clients!inner(*)&clients.id=eq.12`- @steve-chavez, @Iced-Sun
```
wolfgangwalther, line 23: Since we're introducing this feature and already know about #2009, we should solve this by at least disabling `count=exact` in the problematic case for this release - or by fixing it alltogether. But it seems odd to release a known-to-be-broken feature.

steve-chavez: Fair enough! I've opened https://github.com/PostgREST/postgrest/pull/2048 for that.

app.json
```diff
@@ -10,7 +10,7 @@
     },
     "POSTGREST_VER": {
       "description": "Version of PostgREST to deploy",
-      "value": "8.0.0"
+      "value": "9.0.0"
     },
     "DB_URI": {
```
wolfgangwalther, line 15: Can't comment on the lines below - but there are a lot of config options / environment variables in here. We had plenty of changes to those, that are not reflected here - already between v7 and 8.

Should we update all of those here? Does it matter?

CHANGELOG.md
wolfgangwalther, line 23:
```suggestion
 - #1938, Allow escaping inside double quotes with a backslash, e.g. `?col=in.("Double\"Quote")`, `?col=in.("Back\\slash")` - @steve-chavez
 - #1075, Allow filtering top-level resource based on embedded resources filters. This is enabled by adding `!inner` to the embedded resource, e.g. `/projects?select=*,clients!inner(*)&clients.id=eq.12`- @steve-chavez, @Iced-Sun
```

CHANGELOG.md
```diff
@@ -15,6 +21,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+ - #1857, Make GUC names for headers, cookies and jwt claims compatible with PostgreSQL v14 - @laurenceisla, @robertsosinski
+   + Getting the value for a header GUC on PostgreSQL 14 is done using `current_setting('request.headers')::json->>'name-of-header'` and in a similar way for `request.cookies` and `request.jwt.claims`
+   + PostgreSQL versions below 14 can opt in to the new JSON GUCs by setting the `db-use-legacy-gucs` config option to false (true by default)
 - #1988, Allow specifying `unknown` for the `is` operator - @steve-chavez
 - #2031, Improve error message for ambiguous embedding and add a relevant hint that includes unambiguous embedding suggestions - @laurenceisla
```
wolfgangwalther, line 28:
```suggestion
 - #1988, Allow specifying `unknown` for the `is` operator - @steve-chavez
 - #2031, Improve error message for ambiguous embedding and add a relevant hint that includes unambiguous embedding suggestions - @laurenceisla
```

app.json
steve-chavez: Hm, right. I think I'll remove this change. A postgrest-heroku user could do it later.

(Ideally we'd move this into postgrest-heroku but IIRC that can't be done)

app.json
```diff
@@ -10,7 +10,7 @@
     },
     "POSTGREST_VER": {
       "description": "Version of PostgREST to deploy",
-      "value": "8.0.0"
+      "value": "9.0.0"
```
steve-chavez, line 13:
```suggestion
      "value": "8.0.0"
```

wolfgangwalther:
I pushed the missing `v9.0.0` tag to trigger a release.

app.json
wolfgangwalther: Let's discuss in #2049 how we're going to deal with this.

wolfgangwalther:
> I pushed the missing `v9.0.0` tag to trigger a release.

The release job failed:
https://github.com/PostgREST/postgrest/runs/4337370284?check_suite_focus=true

Seems like that was only the docker description, so the images should have been pushed fine, as well as the release on github.

steve-chavez:
> Seems like that was only the docker description

Ah, we had the same issue on v8: https://github.com/PostgREST/postgrest/pull/1900#issuecomment-886915127. Forgot to check that, maybe my dockerhub user is missing privileges.

wolfgangwalther:
> > Seems like that was only the docker description
> 
> Ah, we had the same issue on v8: [#1900 (comment)](https://github.com/PostgREST/postgrest/pull/1900#issuecomment-886915127). Forgot to check that, maybe my dockerhub user is missing privileges.

Ah, right. I remember. This is discussed here: https://github.com/docker/hub-feedback/issues/1927

So we should probably just disable the dockerhub description update job, because it's not currently supported anymore.

Issue #1809, opened 2021-04-12
Update `app.json`


wolfgangwalther:
I don't know what it actually does, but [app.json](https://github.com/PostgREST/postgrest/blob/main/app.json) looks heavily out-of-date.

I assume we need to bring this up-to-date for the next release?

steve-chavez:
The "Deploy to Heroku" button needs that file, not sure if this is being actively used though.

> I assume we need to bring this up-to-date for the next release?

Yep, it would need a version bump(ref: [Release Checklist](https://github.com/PostgREST/postgrest/wiki/Release-Checklist))

wolfgangwalther:
> Yep, it would need a version bump(ref: [Release Checklist](https://github.com/PostgREST/postgrest/wiki/Release-Checklist))

Version bump is one thing... but that file looks like it has our config options listed. Shouldn't we add all the new config options, too?

steve-chavez:
> Shouldn't we add all the new config options, too?

True, forgot about that. 

Maybe we can move the `app.json` to `postgrest-heroku`.

One of our users wrote this: https://postgrest.org/en/stable/install.html#deploying-to-heroku.

If that's the commonly used workflow, then perhaps we don't need to have the `app.json` here.

wolfgangwalther:
Not sure if removing `app.json` wouldn't break that workflow, too. I just had a quick glance at the [heroku docs](https://devcenter.heroku.com/articles/setting-up-apps-using-the-heroku-platform-api) and it seems like the file is expected in the source repo:

> This article describes how to use the app-setups resource of the Platform API. The resource, together with an app.json manifest file embedded in the source code, can be used to programmatically setup an app.

PR #2049, opened 2021-11-26
For discussion: Remove heroku manifest

wolfgangwalther:
The heroku manifest (`app.json`) has been heavily outdated for quite some time. It seems like we, the core team, lack the capacity and knowledge to keep it up to date. For me, it makes no sense in keeping something, that we can't guarantee to be updated and work properly.

I see two options:
- either somebody, who knows and/or uses PostgREST with Heroku steps up and takes care of maintaining this or
- we should just go ahead and remove the outdated stuff.

(Note: We'd also need to remove heroku stuff from `postgrest-docs`)


steve-chavez:
Related to https://github.com/PostgREST/postgrest/issues/1809.

I'm still unsure if the `app.json` is required for anything else besides the "Deploy to Heroku" button.

An example of PostgREST+Heroku can be seen here https://github.com/web3-storage/web3.storage/pull/545 and it only seems to use the [postgrest-heroku repo](https://github.com/web3-storage/web3.storage/pull/545/files#diff-0e74d87b5e30086da84bd10dd29ecf0f6eac056722407859377818c7ce089a66R35-R36), with its own config file.

I'd say we should remove it.

> (Note: We'd also need to remove heroku stuff from postgrest-docs)

I think we'd only need to remove step 3 from https://postgrest.org/en/v8.0/install.html?highlight=heroku#deploying-to-heroku

> 3. Add the require Config Vars in Heroku (see https://github.com/PostgREST/postgrest/blob/main/app.json for more details)

wolfgangwalther:
> I think we'd only need to remove step 3 from https://postgrest.org/en/v8.0/install.html?highlight=heroku#deploying-to-heroku
> 
> > 3. Add the require Config Vars in Heroku (see https://github.com/PostgREST/postgrest/blob/main/app.json for more details)

I removed the reference to this file in all docs branches down to v7.

Since no one had any objections, I will merge this now.

PR #2054, opened 2021-11-28
fix: Link to latest docs in pre-release

### LINKED ISSUE #2018: Pre-Release should point to latest docs
I just updated from the accidental `v9.0.0` pre-release to `v8.0.0-20211102` and have the following diff on the OpenAPI schema output:

```diff
@@ -1,7 +1,7 @@
 {
   "swagger": "2.0",
   "info": {
-    "version": "9.0.0",
+    "version": "8.0.0.20211102 (pre-release)",
     "title": "PostgREST API",
     "description": "standard public schema"
   },
@@ -4740,7 +4740,7 @@
     }
   },
   "externalDocs": {
-    "url": "https://postgrest.org/en/v9.0/api.html",
+    "url": "https://postgrest.org/en/v8.0/api.html",
     "description": "PostgREST Documentation"
   }
 }
```

While the version number accounts for the pre-release correctly, I think the docs url should point to `https://postgrest.org/en/latest/api.html` in this case?

(Side-note: One more reason why not to bump the major version on a pre-release - the v9 docs never existed anyway, yet)

wolfgangwalther:
Resolves #2018

### DIFF
--- a/src/PostgREST/Version.hs
+++ b/src/PostgREST/Version.hs
@@ -30,10 +30,12 @@ prettyVersion =
 
 
 -- | Version number used in docs.
+-- Pre-release versions link to the latest docs
 -- Uses only the two first components of the version. Example: 'v1.1'
 docsVersion :: Text
-docsVersion =
-  "v" <> (T.intercalate "." . map show . take 2 $ versionBranch version)
+docsVersion
+  | isPreRelease = "latest"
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
I tested this manually, by changing `postgrest.cabal` to have a pre-release version and then check the openapi output. This is impossible to unit test with our infrastructure right now, so codecov/patch will be failing.

steve-chavez (approved):
LGTM!

PR #2058, opened 2021-11-30
Fix: Returning 204 No Content where appropriate

### LINKED ISSUE #2001: Inconsistent return values for functions returning void
### Environment

* PostgreSQL version: 12.8
* PostgREST version: 8.0.0
* Operating system: Ubuntu 20.04

### Description of issue

Expected behavior: When calling functions declared returning void, postgrest returns a consistent value reflecting this (preferrably `204 No Content`).

Actual behavior: postgrest returns a empty string or null, depending on the language of the function.

### Example

```
CREATE FUNCTION x() RETURNS VOID AS $$ BEGIN
END $$ LANGUAGE plpgsql;
CREATE FUNCTION z() RETURNS VOID AS $$ $$ LANGUAGE sql;
```

For the first function, postgrest returns `""`:

```
GET /rpc/x HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: localhost:3000
User-Agent: HTTPie/1.0.3



HTTP/1.1 200 OK
Content-Range: 0-0/*
Date: Fri, 29 Oct 2021 19:48:17 GMT
Server: postgrest/8.0.0
Transfer-Encoding: chunked

""
```

For the sql function, postgrest returns `null`:

```
GET /rpc/z HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: localhost:3000
User-Agent: HTTPie/1.0.3



HTTP/1.1 200 OK
Content-Range: 0-0/*
Date: Fri, 29 Oct 2021 19:49:38 GMT
Server: postgrest/8.0.0
Transfer-Encoding: chunked

null
```

### Proposal

Postgrest should return `204 No Content` as result for functions that return void. (This also was suggested in #1762, but not implemented.)

wolfgangwalther:
While trying to fix #2001, I started with making our expectance for a absent or present `Content-Type` header more explicit in the spec tests. While doing that, I noticed that a `PUT` was inconsistently returning a `Content-Type` when returning with `204 No Content`.

The first commit improves those tests and fixes PUT.

The second commit solves a TODO in a related part of the code - it should be self-explanatory.

The third commit then implements the breaking change to return `204 No Content` without body for `VOID` RPCs.

I consider only the part about RPCs a breaking change and am thinking about merging the other fix ahead of time. I'd like us to be able to get a bugfix release out, once we fix a couple of the open, non-breaking bugs, so I'd like to hold back on merging the breaking change for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2135, Remove trigger functions from schema cache and OpenAPI output, because they can't be called directly anyway. - @wolfgangwalther
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
 
+### Changed
+
+ - #2001, Return 204 No Content without Content-Type for RPCs returning VOID - @wolfgangwalther
+   + Previously, those RPCs would return "null" as a body with Content-Type: application/json.
+
 ## [9.0.0] - 2021-11-25
 
 ### Added
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -469,9 +469,12 @@ handleInvoke invMethod proc context@RequestContext{..} = do
       RangeQuery.rangeStatusHeader iTopLevelRange queryTotal tableTotal
 
   failNotSingular iAcceptContentType queryTotal $
-    response status
-      (contentTypeHeaders context ++ [contentRange])
-      (if invMethod == InvHead then mempty else LBS.fromStrict body)
+    if Proc.procReturnsVoid proc then
+      response HTTP.status204 [contentRange] mempty
+    else
+      response status
+        (contentTypeHeaders context ++ [contentRange])
+        (if invMethod == InvHead then mempty else LBS.fromStrict body)
 
 handleOpenApi :: Bool -> Schema -> RequestContext -> DbHandler Wai.Response
 handleOpenApi headersOnly tSchema (RequestContext conf@AppConfig{..} dbStructure apiRequest ctxPgVersion) = do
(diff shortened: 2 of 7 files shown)


test/Feature/IgnorePrivOpenApiSpec.hs
```diff
@@ -19,8 +19,14 @@ spec :: SpecWith ((), Application)
 spec = describe "OpenAPI Ignore Privileges" $ do
   it "root path returns a valid openapi spec" $ do
     validateOpenApiResponse [("Accept", "application/openapi+json")]
-    request methodHead "/" (acceptHdrs "application/openapi+json") ""
-      `shouldRespondWith` "" { matchStatus  = 200 }
+    request methodHead "/"
+        (acceptHdrs "application/openapi+json")
+        ""
```
steve-chavez, line 24: This one is just reformatting right?

test/Feature/RangeSpec.hs
```diff
@@ -109,11 +109,10 @@ spec = do
...
           request methodGet "/menagerie"
-                  [("Prefer", "count=none")] ""
-            `shouldRespondWith` "[]"
-            { matchStatus  = 200
-            , matchHeaders = ["Content-Range" <:> "*/*"]
-            }
+              [("Prefer", "count=none")] ""
+            `shouldRespondWith`
+              [json|[]|]
+              { matchHeaders = ["Content-Range" <:> "*/*"] }
```
steve-chavez, line 115: Hm, why does 200 go away here?

steve-chavez:
> The second commit solves a TODO in a related part of the code - it should be self-explanatory.

Nice one, that TODO was there ever since I can remember.

> I'd like us to be able to get a bugfix release out, once we fix a couple of the open, non-breaking bugs, so I'd like to hold back on merging the breaking change for now.

Cool, I agree. With that it should be good to merge.

PR #2059, opened 2021-11-30
fix: Make POST and PATCH with keys longer than PostgreSQL's NAMEDATALEN work as expected

### LINKED ISSUE #1937: PATCH column with name longer than NAMEDATALEN will cause data loss in column
### Environment

* PostgreSQL version: 13
* PostgREST version: Ubuntu
* Operating system: 20.04

### Description of issue

Because postgresql have 64 bytes fixed length of table and column names, postgresql truncates long names.
Although in Latin-based languages 64 bytes is enough for column name, when we name column in other language (such as Cyrillic) we will have only 32 symbols instead of 64 because Unicode.

So, if we name some column in Russian like "НаименованиеПлательщикаПриПеречисленииНалогов" (what means name of tax payer when he pays taxes) it will truncate to "НаименованиеПлательщикаПриПереч".

Anyway, PostgreSQL works fine with truncated names. Inserting, Updating - all good. It will autotruncate long names in sql-scripts and properly CRUD data.

But when we try to PATCH view with long column name in PostGREST by it's full name - it will cause to data loss (value will be null)

(Steps to reproduce: Include a minimal SQL definition plus how you make the request to PostgREST and the response body)

1. Create table "test" with only one column - "VeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongName" (72 chars)

It will be shrinked to "VeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongNameVer" (63 chars)

2. Insert any data into it.
3. Try to update data via SQL.
4. Then try to PATCH data via PostGREST by full name:
{
 "VeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongName": true
}
5. Value will become null

wolfgangwalther:
This is my attempt at #1937, with the approach of truncating keys to 64 chars length.

The query is significantly more complex and I think I am seeing a slight regression in performance when testing this locally with `postgrest-loadtest`. Not sure whether that shows in CI. Unfortunately we don't separate the loadtest results by request type, yet, and this would only slow down the `POST` and `PATCH` requests.

At the same time, this leads to a problem, when you have a payload with two keys that will be truncated to the same shorter key:
```json
{
  "VeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongName": "haha",
  "VeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLong": "hehe"
}
```

This results in:

```json
{
  "hint":null,
  "details":null,
  "code":"42701",
  "message":"column \"VeryLongNameVeryLongNameVeryLongNameVeryLongNameVeryLongNameVer\" specified more than once"
}
```

That's probably the same reason, why the `json_populate_recordset` function behaves as it does, and is not fixed.

Note, that this does not happen, when you add two equal keys that are shorter than `NAMEDATALEN` to the json body: The earlier keys will be discarded by the json parsing already.

Overall, I think, we should just consider the fact that some operations seem to work with column names longer than `NAMEDATALEN` a conincidence and maybe even point out in the docs, that we are not supporting those. Let's close #1937 as "won't fix".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #2020, Execute deferred constraint triggers when using `Prefer: tx=rollback` - @wolfgangwalther
+ - #1937, Make POST and PATCH with keys longer than PostgreSQL's NAMEDATALEN work as expected - @wolfgangwalther
 
 ## [9.0.0] - 2021-11-25
 
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -88,7 +88,7 @@ mutateRequestToQuery (Insert mainQi iCols body onConflct putConditions returning
   "FROM (" <>
     "SELECT rec.* " <>
     "FROM json_array_elements(" <> SQL.sql selectBody <> ") AS elem, " <>
-         "json_populate_record(null::" <> SQL.sql (fromQi mainQi) <> ", elem) AS rec) _ " <>
+         "json_populate_record(null::" <> SQL.sql (fromQi mainQi) <> ", " <> (SQL.sql $ truncateJsonKeysF "elem") <> ") AS rec) _ " <>
   -- Only used for PUT
   (if null putConditions then mempty else "WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "_") <$> putConditions)) <>
   SQL.sql (BS.unwords [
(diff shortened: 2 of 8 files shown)


steve-chavez:
>  maybe even point out in the docs, that we are not supporting those. Let's close #1937 as "won't fix".

Fully agree, I think this should be addressed in PostgreSQL itself.

> The query is significantly more complex and I think I am seeing a slight regression in performance when testing this locally with postgrest-loadtest

That reminds me that I noticed something similar while benching inserts with`?columns`([doc](https://postgrest.org/en/v9.0/api.html#specifying-columns)), somehow it made throughput reduce noticeably - and this despite we only do a simple `json_typeof` inside the query. So yes, in principle I see how doing `json_array_elements` could reduce performance.

(The above needs further research but it could be that `?columns` is only apt for few requests with big payloads but not for many requests with small payloads)

PR #2071, opened 2021-12-03
nix: Add weeder to postgrest-coverage

wolfgangwalther:
Found [weeder](https://hackage.haskell.org/package/weeder) and gave it a try. It found some dead code in `SpecHelper.hs`. Removed that and added it to `postgrest-coverage`. `postgrest-coverage` instead of `postgrest-lint`, because it needs to have the whole project built to operate.


monacoremo:
Very nice!

PR #2081, opened 2021-12-14
fix: `is` not working with upper/mixed case values

### LINKED ISSUE #2077: `is` operator no longer working with an uppercase value
Unintended effect of https://github.com/PostgREST/postgrest/pull/2027:
```http
GET /projects?id=is.NULL
```

```js
{"details":"unexpected \"N\" expecting null or trilean value (unknown, true, false)","message":"\"failed to parse filter (is.NULL)\" (li
ne 1, column 4)"}
```

steve-chavez:
Fixes #2077.

### DIFF
--- a/test/Feature/QuerySpec.hs
+++ b/test/Feature/QuerySpec.hs
@@ -86,6 +86,10 @@ spec actualPgVersion = do
         { matchHeaders = [matchContentTypeJson] }
 
     it "matches with trilean values in upper or mixed case" $ do
+      get "/chores?done=is.NULL" `shouldRespondWith`
+        [json| [{"id": 3, "name": "wash the dishes", "done": null }] |]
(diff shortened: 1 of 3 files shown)


test/Feature/QuerySpec.hs
```diff
@@ -85,6 +85,19 @@ spec actualPgVersion = do
         [json| [{"id": 3, "name": "wash the dishes", "done": null }] |]
         { matchHeaders = [matchContentTypeJson] }
 
+    it "matches with trilean values in upper or mixed case" $ do
```
laurenceisla, line 88: Maybe it needs a test for `NULL` (not here, but in the "matches nulls..." section) for completion.

laurenceisla (approved):

PR #2082, opened 2021-12-14
Add hint to verify table and foreign key existence on a relationship error

laurenceisla:
This PR adds an extra hint that recommends the user to verify the existence of the tables and the relationship between them in the database before restarting the schema cache.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -94,8 +94,8 @@ instance JSON.ToJSON ApiRequestError where
   toJSON InvalidRange = JSON.object [
     "message" .= ("HTTP Range error" :: Text)]
   toJSON (NoRelBetween parent child schema) = JSON.object [
-    "hint"    .= ("Make sure that the tables '" <> parent <> "' and '" <> child <> "' exist in the database schema '" <> schema <> "' and that there is a foreign key relationship between them. If both conditions are met, try reloading the schema cache." :: Text),
-    "message" .= ("Could not find a relationship between '" <> parent <> "' and '" <> child <> "' in the schema cache." :: Text)]
+    "hint"    .= ("Verify that '" <> parent <> "' and '" <> child <> "' exist in the schema '" <> schema <> "' and that there is a foreign key relationship between them. If a new relationship was created, try reloading the schema cache." :: Text),
+    "message" .= ("Could not find a relationship between '" <> parent <> "' and '" <> child <> "' in the schema cache" :: Text)]
   toJSON (AmbiguousRelBetween parent child rels) = JSON.object [
     "hint"    .= ("Try changing '" <> child <> "' to one of the following: " <> relHint rels <> ". Find the desired relationship in the 'details' key." :: Text),
     "message" .= ("Could not embed because more than one relationship was found for '" <> parent <> "' and '" <> child <> "'" :: Text),
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -59,7 +59,7 @@ data ApiRequestError
   | InvalidRange
   | InvalidBody ByteString
   | ParseRequestError Text Text
-  | NoRelBetween Text Text
+  | NoRelBetween Text Text Text
   | AmbiguousRelBetween Text Text [Relationship]
   | AmbiguousRpc [ProcDescription]
   | NoRpc Text Text [Text] Bool ContentType Bool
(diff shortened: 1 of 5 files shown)


src/PostgREST/Error.hs
```diff
@@ -93,9 +93,9 @@ instance JSON.ToJSON ApiRequestError where
     "message" .= T.decodeUtf8 errorMessage]
   toJSON InvalidRange = JSON.object [
     "message" .= ("HTTP Range error" :: Text)]
-  toJSON (NoRelBetween parent child) = JSON.object [
-    "hint"    .= ("If a new foreign key between these entities was created in the database, try reloading the schema cache." :: Text),
-    "message" .= ("Could not find a relationship between " <> parent <> " and " <> child <> " in the schema cache" :: Text)]
+  toJSON (NoRelBetween parent child schema) = JSON.object [
+    "hint"    .= ("Make sure that the tables '" <> parent <> "' and '" <> child <> "' exist in the database schema '" <> schema <> "' and that there is a foreign key relationship between them. If both conditions are met, try reloading the schema cache." :: Text),
```
steve-chavez, line 97: - Removing the tables mention, since the resources can be views as well
- Removing the database mention, which I think is unnecessary

```suggestion
    "hint"    .= ("Make sure that '" <> parent <> "' and '" <> child <> "' exist in the schema '" <> schema <> "' and that there is a foreign key relationship between them. If both conditions are met, try reloading the schema cache." :: Text),
```

steve-chavez: > If both conditions are met, try reloading the schema cache.

The above doesn't seem right, it tells the user to reload the schema cache after verifying.

steve-chavez:
Maybe we should remove the mention to "reloading the schema cache", it's unnecessary and confusing when [automatic schema cache reloading](https://postgrest.org/en/v9.0/schema_cache.html#automatic-schema-cache-reloading) is enabled.

steve-chavez:
Or perhaps we could try do detect the `pgrst_watch` event trigger in the schema cache and add the hint conditionally. The effort doesn't seem worth the trouble though. The user still has to go to docs to find how to reload the schema cache.

I think it would be better to include an error code(https://github.com/PostgREST/postgrest/pull/1926) for this.

src/PostgREST/Error.hs
wolfgangwalther: > [...] it tells the user to reload the schema cache after verifying.

Which seems like a perfectly valid procedure to me. Why do you think it's wrong?

steve-chavez: Ah, in the case the user detects a typo(`useRs` instead of `users`) then he/she would correct it and then proceed to reload the schema cache(following the hint verbatim) instead of just retrying the request.

steve-chavez: Also, I don't think where doing the hints right for some cases.

When `AmbiguousRelBetween` happens, we give a hint that's actionable from the frontend(change the url), while reloading the schema cache must be done from the backend - not actionable from the client side.

wolfgangwalther: > Ah, in the case the user detects a typo(useRs instead of users) then he/she would correct it and then proceed to reload the schema cache(following the hint verbatim) instead of just retrying the request.

... which would be fine, because it would be successful. I'd rather have somebody fix their queries **and** reload the schema cache, even if not needed, and be successful - then missing one of both an fail to get it right.

> When AmbiguousRelBetween happens, we give a hint that's actionable from the frontend(change the url), while reloading the schema cache must be done from the backend - not actionable from the client side.

I like that thought in principal. However, if we take the part about the schema cache out, then this might really leave people in the dark about what a possible solution would be.

laurenceisla: > Ah, in the case the user detects a typo(useRs instead of users) then he/she would correct it and then proceed to reload the schema cache(following the hint verbatim) instead of just retrying the request.

In that case,  we could keep the original hint at the end to avoid this.

```suggestion
    "hint"    .= ("Verify that '" <> parent <> "' and '" <> child <> "' exist in the schema '" <> schema <> "' and that there is a foreign key relationship between them. If a new relationship was created, try reloading the schema cache." :: Text),
```
Or maybe move the explanation to the `detail` key, like:

```suggestion
    "hint"    .= ("If a new foreign key between these entities was created in the database, try reloading the schema cache." :: Text),
    "detail"  .= ("There is no foreign key relationship between '" <> parent <> "' and '" <> child <> "' or they do not exist in the schema '"<> schema <> "'." :: Text),

> When AmbiguousRelBetween happens, we give a hint that's actionable from the frontend(change the url), while reloading the schema cache must be done from the backend - not actionable from the client side.

I think that's inevitable due to the hints that may come from database errors.

steve-chavez: > However, if we take the part about the schema cache out, then this might really leave people in the dark about what a possible solution would be.

True, I don't have a better suggestion ATM, so I think the schema cache should stay.

> In that case, we could keep the original hint at the end to avoid this.

@laurenceisla Yeah, giving the user multiple hints seems much better.

steve-chavez (approved):

Issue #2070, opened 2021-12-03
Help with embedded ambiguity, upgrading from v6.0.2 to v9.0.0


bolerodan:
So we are currently still on version 6.0.2 of PostgREST and a while back there was lots of talk about resource ambiguity, duct typing, specifying foreign key constraints, etc.

This has held us from upgrading due to these changes, but version 9.0 has introduced some great features we actually want, so we are upgrading, but that requires quite the trickle down effect since duct typing seems to be removed. In fact it's amazing we managed to get so far with the duct typing, but I digress.

Issues I'm having. For starters here is an example table structure 

```
                                                Table "api.plate"
      Column      |           Type           | Collation | Nullable |                   Default                   
------------------+--------------------------+-----------+----------+---------------------------------------------
 plate_id         | integer                  |           | not null | nextval('api.plate_plate_id_seq'::regclass)

Indexes:
    "plate_pkey" PRIMARY KEY, btree (plate_id)


                                           Table "api.well"
     Column     |       Type       | Collation | Nullable |                  Default                  
----------------+------------------+-----------+----------+-------------------------------------------
 well_id        | integer          |           | not null | nextval('api.well_well_id_seq'::regclass)
 plate_id       | integer          |           |          | 
 extract_id     | integer          |           |          | 
 parent_well_id | integer          |           |          | 

Indexes:
    "well_pkey" PRIMARY KEY, btree (well_id)
Foreign-key constraints:
    "well_extract_id_fkey" FOREIGN KEY (extract_id) REFERENCES api."extract"(extract_id) ON DELETE SET NULL
    "well_parent_well_id_fkey" FOREIGN KEY (parent_well_id) REFERENCES api.well(well_id) ON DELETE SET NULL
    "well_plate_id_fkey" FOREIGN KEY (plate_id) REFERENCES api.plate(plate_id) ON DELETE CASCADE


                                              Table "api.plate_plan_step"
       Column       |  Type   | Collation | Nullable |                             Default                             
--------------------+---------+-----------+----------+-----------------------------------------------------------------
 plate_plan_step_id | integer |           | not null | nextval('api.plate_plan_step_plate_plan_step_id_seq'::regclass)
 plate_plan_id      | integer |           |          | 
 from_well_id       | integer |           |          | 
 to_well_id         | integer |           |          | 
 to_plate_id        | integer |           |          | 
Indexes:
    "plate_plan_step_pkey" PRIMARY KEY, btree (plate_plan_step_id)
[...]

PR #2085, opened 2021-12-16
Ignore impossible disambiguations examples in error hint

laurenceisla:
* Ignores m2m relationships where the target and hint are the same table (self reference)
* Ignores m2m relationships where an extra hint is needed to disambiguate

Resolves part of #2070

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -148,20 +148,21 @@ compressedRel Relationship{..} =
 relHint :: [Relationship] -> Text
 relHint =  T.intercalate ", " . unique . foldr hint []
   where
-    sg :: Ord a => [a] -> [[a]]
-    sg = group . sort
-    filterByLength :: Ord a => (Int -> Bool) -> [a] -> [[a]]
-    filterByLength p = filter (p . length) . sg
-    unique :: Ord a => [a] -> [a]
-    unique = concat . filterByLength (==1)
+    unique = concat . filter ((==1) . length) . group
     hint Relationship{..} hintList =
-      let buildHint rel = ("'" <> tableName relForeignTable <> "!" <> rel <> "'"):hintList in
+      let
+        buildHint isSelfM2M isSelfM2O rel
+          -- Do not include a M2M self reference case where target and hint are the same table.
+          -- TODO: Omit this conditional when aliases are implemented in the query builder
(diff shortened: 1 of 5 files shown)


wolfgangwalther:
This only removes the mentioned relationships from the hint for the error message - but does not remove from the actual relationship discovery. I don't really understand the purpose of that at all - can you explain what the idea here is?

laurenceisla:
@wolfgangwalther  Yes, this PR is centered only the `hint` key to make it accurate with the examples it gives as a fix to #2031 (to take into consideration the special self reference cases). It does not limit the result in the `details` key as to let the user know that relationships still exist, even if it cannot be disambiguated e.g. the user may ask themselves why the relationship is not detected even if it exists in the database: the details show them that they exist,

From this perspective maybe another key could  be added in the `details` to let the user know where does the relationship in the hint come from:

```json
{
  "hint": "Try changing 'dependencies' to one of the following: 'dependencies!dependencies_locations_fk'. Find the desired relationship in the 'details' key.",
  "details": [
    {
      "relationship": "test.dependencies[dependencies_locations_fk][dependencies_parent_fk]",
      "embedding": "locations with dependencies",
      "cardinality": "many-to-many",
      "disambiguation": "Cannot disambiguate" // here
    },
    {
      "relationship": "dependencies_locations_fk[id][location_id]",
      "embedding": "locations with dependencies",
      "cardinality": "one-to-many"
      "disambiguation": "dependencies!dependencies_locations_fk" // and here
    }
  ],
  ...
}
```

In summary I took it as a simple fix to #2031 and leaves these comments to other PRs: https://github.com/PostgREST/postgrest/issues/2070#issuecomment-996073047 and https://github.com/PostgREST/postgrest/issues/2070#issuecomment-996274638.

wolfgangwalther:
I think that's the wrong approach. If our current implementation of the query language does not allow to disambiguate a request in a way that a certain relationship can actually be used, then we should just not create that relationship in the first place. Listing something in the output for `300 Multiple Choices` that can't actually be chosen is just wrong - and we should solve that. Tweaking the hint provides little value.

steve-chavez:
@laurenceisla Since #2070 is closed perhaps this one should too?

laurenceisla:
@steve-chavez Yes! There is no need for this PR anymore, as the relationship discovery is now restricted in #2262

Issue #1766, opened 2021-03-03
Add recovery tests


steve-chavez:
Currently recovery tests are done manually, it'd be great to have them as automated tests.

These are the main scenarios:

(the connection recovery worker is referred as just "worker")

### 1. postgrest started with a pg connection, then pg becomes unavailable

- worker starts only after a request to postgrest, postgrest should respond with a `{"details":"no connection to the server\n","message":"Database client error. Retrying the connection."}`
  - if db-channel-enabled=true, worker starts immediately, not necessary to prove this in tests though.
- pg becomes available, postgrest succeeds reconnecting, reloads the schema cache and responds with 200
- if db-load-guc-config=true, it should also re-read the in-db config.
  - test with an `ALTER ROLE postgrest_test_authenticator SET pgrst.db_schemas = 'public';` and try a `GET /public_consumers` which should give a 404 if the in-db config isn't re-read.

### 2. unavailable pg, postgrest started

- worker starts immediately, postgrest should respond with a `503 {"message":"Database connection lost. Retrying the connection."}`
  - Bug: if db-channel-enabled=true, postgrest doesn't reply and curl gives `Connection refused`. This must be because of the `mvarConnectionStatus` MVar, it doesn't happen on 1 though.
- pg becomes available, postgrest succeeds reconnecting, reloads the schema cache and responds with 200
- if db-load-guc-config=true, it should also re-read the in-db config.
  - Same test as 1

### 3. SIGUSR1 - NOTIFY reload schema

- when postgrest has a pg connection, both SIGUSR1 and NOTIFY will reload the schema cache
  - if db-load-guc-config=true, it should also re-read the in-db config.
  - ensure SIGUSR1 starts the worker when db-channel-enabled=true(got it to lock before, and worker was not starting, so this must be ensured)
- when postgrest loses the connection, and db-channel-enabled=false(only SIGUSR1)
  - SIGUSR1 starts the worker, only one can run at a time. Ensured by `refIsWorkerOn`, this can be confirmed by doing several SIGUSR1 and just noting one `Attempting to reconnect to the database in 1 seconds...` message. If `refIsWorkerOn` is removed, there will be several `Attempting to reconnect to the database in 1 seconds...` mesagges.
    - Not sure how to test this, maybe count the number of threads?
  - pg becomes available, postgrest succeeds reconnecting, reloads the schema cache and responds with 200
- when postgrest loses the connection, and db-channel-enabled=true
[...]

PR #2107, opened 2021-12-29
Consider schema cache state on the health check

steve-chavez:
As suggested in https://github.com/PostgREST/postgrest/pull/2092#issuecomment-1000776098

> However, what we can and should do is: Keep track of the status of the last schema cache refresh - and if that fails, e.g. because of something like #2024, we should return 503. Currently we are still returning 200, when the schema cache fails.

This change does prevent a healthy response in case something like #2024 happens, but only at startup. If the schema cache fails to load after startup, the main app will continue working and the health check will respond accordingly.

The main app failure happens at startup because of:

https://github.com/PostgREST/postgrest/blob/4f7ee6f36aae6ed36ecfec4ac3c1aa0b656bdddb/src/PostgREST/App.hs#L203-L208

Should we turn the schema cache to Nothing if it fails loading? That would make the main app fail for all requests, I think it's better to leave the schema cache stale in that case, as it happens currently.

**Edit**: On second thought, I think it'd be better to void the schema cache, otherwise the app will continue running with a stale schema cache and the developers won't know about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
  - #1933, #2109, Add a minimal health check endpoint - @steve-chavez
    + For enabling this, the `admin-server-port` config must be set explictly
-   + The check is at the `<host>:<admin_server_port>/live` endpoint. A 200 OK status will be returned if postgrest is alive, otherwise a 503 will be returned.
+   + A `<host>:<admin_server_port>/live` endpoint is available for checking if postgrest is running on its port/socket. 200 OK = alive, 503 = dead.
    + A `<host>:<admin_server_port>/ready` endpoint is available for checking a correct internal state(the database connection plus the schema cache). 200 OK = ready, 503 = not ready.
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


steve-chavez:
>      Nothing -> 
>       throwError Error.ConnectionLostError 

Fixed the above, it gave a misleading connection error(`"Database connection lost. Retrying the connection."`) while the error was with the schema cache query(error now is `"Could not query the database for the schema cache. Retrying."`).

wolfgangwalther:
> **Edit**: On second thought, I think it'd be better to void the schema cache, otherwise the app will continue running with a stale schema cache and the developers won't know about it.

Agreed!

steve-chavez:
I had difficulties with trying to fix the endless loop schema cache loading here.

I left a comment on https://github.com/PostgREST/postgrest/issues/1766#issuecomment-1007109348 to work on it on a later PR.

wolfgangwalther:
I tested the ready endpoint by adding it as the `wait_for_startup` uri in the io tests in #2116. Before this PR here, using ready would fail half the io tests randomly. Now, they are passing.

wolfgangwalther (approved):

PR #2109, opened 2021-12-31
Connect to main app socket on health check

steve-chavez:
As discussed on https://github.com/PostgREST/postgrest/pull/2092#issuecomment-1003260934

### Tasks

- [x] Add io test(remove unix socket and do a health check)
- [x] Change `/health` to `/ready`
- [x] Add `/live` endpoint to the admin server

---

Will finish https://github.com/PostgREST/postgrest/pull/2107 after this one.

### DIFF
--- a/src/PostgREST/Admin.hs
+++ b/src/PostgREST/Admin.hs
@@ -24,7 +24,7 @@ postgrestAdmin appState appConfig req respond  = do
   isMainAppReachable <- isRight <$> reachMainApp appConfig
 
   case Wai.pathInfo req of
-    ["health"] ->
+    ["ready"] ->
       if configDbChannelEnabled appConfig then do
         listenerOn <- AppState.getIsListenerOn appState
         respond $ Wai.responseLBS (if listenerOn && isMainAppReachable then HTTP.status200 else HTTP.status503) [] mempty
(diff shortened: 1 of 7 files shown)


src/PostgREST/Admin.hs
```diff
@@ -0,0 +1,53 @@
...
+        sock <- socket AF_UNIX Stream 0
+        connect sock $ SockAddrUnix path
+        return sock
+      Nothing -> do
+        let hints = defaultHints { addrSocketType = Stream }
+        addr:_ <- getAddrInfo (Just hints) (Just . T.unpack $ configServerHost appConfig) (Just . show $ configServerPort appConfig)
+        sock <- socket (addrFamily addr) (addrSocketType addr) (addrProtocol addr)
+        connect sock $ addrAddress addr
+        return sock
+    sendEmpty sock = void $ send sock mempty
```
steve-chavez, line 53: Sending an empty message.. this could be removed. I found that just `connect` is necessary for getting an error.

steve-chavez: A note aside: I wasn't able to send a valid HTTP message with the `Network.socket` lib. I tried with `send sock "GET / HTTP/1.1\n"`. Not sure what could be wrong there but that was only for testing purposes, connect seems enough.

steve-chavez:
All green(including codecov), seems good to merge.

PR #2111, opened 2022-01-03
Refactor IO tests

wolfgangwalther:
While hacking on other stuff, I made a couple of improvements to our io tests. These changes:
- improve io test performance by ~20% on my machine
- improve diagnostic log output when tests fail
- improve maintainability by:
  - removing two now redundant test-cases
  - separating io test SQL fixtures from spec fixtures
- refactor the new admin tests to not block potential parallel running* of io tests (@steve-chavez watch out - you'll need to rebase your new admin test in #2107)

*parallel running not working, yet, due to other problems, but getting there


test/io-tests/configs/unix-socket.config
```diff
@@ -1,4 +0,0 @@
-db-pool = 1
-server-unix-socket = "$(POSTGREST_TEST_SOCKET)"
-jwt-secret = "reallyreallyreallyreallyverysafe"
-db-config = false
```
wolfgangwalther, line 4: This test is not used anymore, now that we are using unix sockets in io tests everywhere.

test/io-tests/test_io.py
```diff
@@ -578,16 +593,15 @@ def test_db_schema_reload(tmp_path, defaultenv):
...
     configfile.write_text(config)
 
-    headers = {"Accept-Profile": "v1"}
     env = {key: value for key, value in defaultenv.items() if key != "PGRST_DB_SCHEMAS"}
 
     with run(configfile, env=env) as postgrest:
-        response = postgrest.session.get("/parents", headers=headers)
-        assert response.status_code == 404
+        response = postgrest.session.get("/rpc/get_guc_value?name=search_path")
+        assert response.text == '"public, public"'
```
wolfgangwalther, line 600: Instead of testing the changed db-schema implicitly, I opted to test it explicitly by looking at the search path. It didn't make sense to me to copy over the `/parents` table just for that kind of test, when we can instead do it like this.

test/io-tests/test_io.py
```diff
@@ -299,23 +334,6 @@ def test_expected_config_from_db_settings(defaultenv, role, expectedconfig):
...
-    """
-    env = {
-        **defaultenv,
-        "PGRST_DB_CONFIG": "true",
-    }
-    with run(env=env) as postgrest:
-        uri = "/rpc/get_guc_value?name=pgrst.jwt_secret"
-        response = postgrest.session.get(uri)
-
-        assert response.text == '"OVERRIDEREALLYREALLYREALLYREALLYVERYSAFE"'
```
wolfgangwalther, line 316: I don't think we need this test-case anymore, because now that PG 14 has arrived, and we are running tests on it, we can't create any of those "non-showable-gucs" anymore, that were a problem in the referenced discussion.

test/io-tests/db_config.sql
```diff
@@ -0,0 +1,55 @@
+CREATE ROLE db_config_authenticator LOGIN NOINHERIT;
```
wolfgangwalther, line 1: The db-config tests are now running with a completely separate user and not the "main" `postgrest_test_authenticator`, with which all other IO tests are running. This prevents a lot of interaction between the config options that are set here and the other tests that don't rely on it. This will make some follow-up PRs much easier to work on.

test/io-tests/fixtures.sql
```diff
@@ -0,0 +1,75 @@
...
+
+GRANT postgrest_test_anonymous, postgrest_test_author TO :USER;
+
+CREATE SCHEMA v1;
+GRANT USAGE ON SCHEMA v1 TO postgrest_test_anonymous;
+
+CREATE TABLE authors_only ();
+GRANT SELECT ON authors_only TO postgrest_test_author;
+
+CREATE TABLE projects AS SELECT FROM generate_series(1,5);
```
wolfgangwalther, line 14: Ever wanted to see a table without columns but with some rows in action? Here we go :D

steve-chavez: That is some black magic :magic_wand: !

It's great that we now isolate the tables used in the io tests from the api tests :clap:

steve-chavez (approved):
Nice spring cleaning :blossom: !

PR #2115, opened 2022-01-06
Add current user to the request log

wolfgangwalther:
See discussion in #1988.

TODO:
- [ ] Update docs: "Once this PR is merged, we should update the [log format in the docs](https://postgrest.org/en/latest/admin.html#logging)."

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -12,40 +12,54 @@ very simple authentication system inside the PostgreSQL database.
 -}
 {-# LANGUAGE RecordWildCards #-}
 module PostgREST.Auth
-  ( containsRole
-  , jwtClaims
-  , JWTClaims
+  ( AuthResult (..)
+  , getResult
+  , getRole
+  , middleware
   ) where
 
-import qualified Crypto.JWT          as JWT
-import qualified Data.Aeson          as JSON
-import qualified Data.HashMap.Strict as M
-import qualified Data.Vector         as V
+import qualified Crypto.JWT                      as JWT
+import qualified Data.Aeson                      as JSON
+import qualified Data.Aeson.Types                as JSON
+import qualified Data.ByteString.Lazy.Char8      as LBS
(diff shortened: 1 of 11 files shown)


wolfgangwalther:
Compared to #1988, I tied up a few loose ends and did some more refactoring:
- Moved the logger and cors middlewares out of `Middleware.hs` into their own files - and put the new auth middleware into `Auth.hs`
- The failure-case in the auth middleware is now handled by storing the error in the vault and then responding with that in the main app. This was the only way for me to deal with the circular dependency of auth and logger middlewares: "auth needs to come first to have the user available when logging, but logging needs to come first to be able to log failed auth attempts..."
- The role is now stored independently from the claims. This will allow to remove the "write role back into claims" thing we have right now, which is buggy imho. I will create an issue to discuss this.

The PR is now split into a refactor commit (the new auth middleware) and a feature commit (adding the user to the log).

src/PostgREST/Auth.hs
```diff
@@ -57,26 +71,49 @@ jwtClaims AppConfig{..} payload time = do
...
+  claims <- runExceptT $ parseToken conf (LBS.fromStrict token) time
+
+  let
+    authResult = mapRight (parseClaims conf) claims
+    req' = req { Wai.vault = Wai.vault req & Vault.insert authResultKey authResult }
+  app req' respond
+
+authResultKey :: Vault.Key (Either Error AuthResult)
+authResultKey = unsafePerformIO Vault.newKey
+{-# NOINLINE authResultKey #-}
```
steve-chavez, line 113: Why is `NOINLINE` needed here?

wolfgangwalther: To make sure it's only run once and it does not in fact generate multiple different keys:
https://www.yesodweb.com/blog/2015/10/using-wais-vault

steve-chavez: Cool, I see now that this is a well-known way to use vault with wai.

steve-chavez (approved):

PR #2116, opened 2022-01-07
Make io tests use admin/ready to wait for startup

wolfgangwalther:
Based on #2107.


steve-chavez:
It's so awesome how it all fits together :smiley:

PR #2119, opened 2022-01-07
update cabal description

steve-chavez:
Trying a test commit because somehow codecov failed on https://github.com/PostgREST/postgrest/commit/a134d8a493e572e32d30f0c21e51b6f467ab24fb.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -2,7 +2,7 @@ name:               postgrest
 version:            9.0.0.20220107
 synopsis:           REST API for any Postgres database
 description:        Reads the schema of a PostgreSQL database and creates RESTful routes
-                    for the tables and views, supporting all HTTP verbs that security
+                    for tables, views, and functions, supporting all HTTP verbs that security
                     permits.
 license:            MIT
 license-file:       LICENSE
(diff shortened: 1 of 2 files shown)


steve-chavez:
Keeps failing with:

```
25hhpc: can not find postgrest-9.0.0.20211220-inplace/PostgREST.Middleware in ./.hpc
CallStack (from HasCallStack):
  error, called at libraries/hpc/Trace/Hpc/Mix.hs:122:15 in hpc-0.6.1.0:Trace.Hpc.Mix
Temporary directory kept at: /tmp/postgrest/postgrest-coverage-qeI
Error: Process completed with exit code 1.
```

@wolfgangwalther Any pointers as to why coverage fails?

steve-chavez:
Ah, I see it now https://github.com/PostgREST/postgrest/blob/e13d912a79ae5dd2b11e5069562c13ec93d650ae/test/coverage.overlay#L1-L5

wolfgangwalther:
Yes, that should solve it. I just thought about the overlay file a couple of hours ago, and the hardcoded version string is one problem. The other problem is that in #2115 the relevant code ('corsPolicy`) is moved to a different file - but the code coverage is not failing.

So basically the overlay file currently makes CI fail when it shouldn't (version update) and makes it not fail when it should (ticked line is moved).

Not very helpful.

Let's remove the content of the file for now (we have to keep the file, because `postgest-coverage` depends on it) until I find a better way to use that.

steve-chavez:
> Let's remove the content of the file for now (we have to keep the file, because postgest-coverage depends on it) until I find a better way to use that.

Cool, I'll do that.

PR #2122, opened 2022-01-10
Detecting space leaks (stack size) in CI

### LINKED ISSUE #387: Automate the space leak detection introduced by new code
http://neilmitchell.blogspot.co.uk/2015/09/detecting-space-leaks.html?m=1
In short it means adding 

```
ghc-options: -with-rtsopts=-K1K
```

to force stack overflow on bad code

wolfgangwalther:
Resolves #387.

Related:
- https://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html?m=1
- https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/runtime_control.html#rts-flag--K%20%E2%9F%A8size%E2%9F%A9

`-K1K` will make the spec tests fail - the OpenAPI output causing the stack overflow here. I raised it up to `-K33K` to make it pass. This is just added to prevent things from using more stack than right now. Maybe we can lower it to `-K1K` at some point in the future.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -152,6 +152,8 @@ executable postgrest
 
   if flag(dev)
     ghc-options: -O0 -fwrite-ide-info
+                 -- https://github.com/PostgREST/postgrest/issues/387
+                 -with-rtsopts=-K1K
     if flag(hpc)
       ghc-options: -fhpc -hpcdir .hpc
   else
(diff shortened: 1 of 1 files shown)


steve-chavez (approved):

steve-chavez:
>  the OpenAPI output causing the stack overflow here. 
> Maybe we can lower it to -K1K at some point in the future.

True, OpenAPI is expensive and even more on big schemas. I think this mostly comes from the json generation we do on Haskell code.

PR #2127, opened 2022-01-19
Add CI for ARM architectures

### LINKED ISSUE #2010: Integrate Github Actions with self-hosted runner for ARM64
### Problem

Currently I have to manually compile PostgREST for ARM64 on an AWS Ubuntu instance - this has gotten too distracting and tedious.

And the same would have to be done for Docker https://github.com/PostgREST/postgrest/issues/1117#issuecomment-949260970.

### Proposal

Use Github Actions [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) to automate building on ARM64.

There's a list of examples in [jonico/awesome-runners](https://github.com/jonico/awesome-runners) from where https://github.com/machulav/ec2-github-runner looks exactly what we need.

laurenceisla:
Closes #2010

Explanation on what this PR does so far:

Requirements

- A remote Ubuntu server with Docker installed and a SSH key added to the `.ssh/authorized_keys` (currently using a AWS EC2 t4g.xlarge instance, lower RAM gave errors). The username is `ubuntu` and has `sudo` privileges.
- The following repository secrets:
  - `SSH_ARM_HOST`: The domain/IP of the remote server
  - `SSH_ARM_PRIVATE_KEY`:  The PEM private key. Its public key must be added to the `authorized_keys` of the remote server
  - `SSH_ARM_FINGERPRINT`: The fingerprint that the remote server uses (can be obtained using `ssh-keygen -l -f /etc/ssh/ssh_host_ed25519_key.pub | cut -d ' ' -f2` in the remote server)

Steps

- The scripts in `.github/scripts/arm` are copied to the remote server and executed there afterwards.
- The `build.sh` script uses Docker (`.github/scripts/arm/docker-env/Dockerfile`) to build the aarch64 and armv7 architectures. Then the binaries are downloaded and added as artifacts. This is executed only when merging to main because forks cannot use the secrets needed for the Job.
- The `docker-publish.sh`  uses the same Dockerfile and adds a separate ARM image for that version, then adds it to the manifest of the already pushed image for the amd64 architecture. This only executes for full and pre releases.

Tests

- I used another repository to test and simulate a release. There, I adapted this CI and hard coded some variables to make it work. This is the latest functional action that simulates a full release https://github.com/laurenceisla/github-actions-test/actions/runs/2160203915 . The Docker Hub i used was also updated accordingly: https://hub.docker.com/r/laurenceisla/postgrest/tags

Drawbacks

- As mentioned above, in order to add the ARM versions to the manifest, a separate ARM image must be pushed. This image needs to be manually deleted after the process is finished.

Observations

- With the Dockerfile, the build *could* be done by a Docker GitHub action without connecting to the remote server, but it would need emulation which is very slow [according to some reviews](https://github.com/haskell/bytestring/blob/master/.github/workflows/ci.yml#L114-L118) (didn't test it for our case).


.github/scripts/arm/build.sh
```diff
@@ -0,0 +1,59 @@
...
+cd ~/$DOCKER_BUILD_PATH
+
+# Build ARM versions
+sudo docker buildx build --build-arg PGRST_GITHUB_COMMIT=$PGRST_GITHUB_COMMIT \
+                         --build-arg BUILDKIT_INLINE_CACHE=1 \
+                         --platform linux/arm/v7,linux/arm64 \
+                         --cache-from $DOCKER_REPO/postgrest:postgrest-build-arm \
+                         --target=postgrest-build \
+                         -t $DOCKER_REPO/postgrest:postgrest-build-arm \
+                         --push .
```
steve-chavez, line 36: We also have a `docker login` and `push` on:

https://github.com/PostgREST/postgrest/blob/cb145271ac8136354df0800e689601c5debd7bfd/.github/workflows/ci.yaml#L366-L370

Would be nice to DRY this up into a nix script.

Not sure if it can be done for `push` since you use it inline with `buildx` here(haven't used that before) but seems doable for `login`.

.github/scripts/arm/build.sh
```diff
@@ -0,0 +1,57 @@
...
+cd ~/$DOCKER_BUILD_PATH
+
+# Build ARM versions
+sudo docker buildx build --build-arg PGRST_GITHUB_COMMIT=$PGRST_GITHUB_COMMIT \
+                         --build-arg BUILDKIT_INLINE_CACHE=1 \
+                         --platform linux/arm/v7,linux/arm64 \
+                         --cache-from $DOCKER_USER/$DOCKER_REPO:postgrest-build-arm \
+                         --target=postgrest-build \
+                         -t $DOCKER_USER/$DOCKER_REPO:postgrest-build-arm \
+                         --push .
```
laurenceisla, line 42: This uses a `postgrest/postgrest:postgrest-build-arm` image which is needed for caching the build process. I think maybe it shouldn't be considered as a postgrest version but a different repo instead, like `postgrest/postgrest-build-arm`.

.github/scripts/arm/build.sh
laurenceisla: In this case, the docker login in `build.sh` is used to publish the intermediate ARM build image for caching purposes. The login in `docker-publish.sh` is the one that is duplicated as you mentioned. Applying a DRY here would solve the need to manually delete the final ARM PostgREST image, but I can't see how to make it work with nix and multiple architectures: the `amd64` version only needs a scratch image while the `arm` ones need a bigger ecosystem. 

Maybe it should be left for later or someone can find a better solution here.

steve-chavez (approved):

steve-chavez:
https://www.haskell.org/ was down and hence https://get-ghcup.haskell.org/ too - that's why CI was failing.. these seem normal now?

https://status.haskell.org/

Issue #1526, opened 2020-05-16
Feature Request: metrics or instrumentation


sevagh:
It would be nice to have a `/metrics` endpoint that exposes Prometheus-style metrics.

Note that I might be opinionated for working mostly with the Prometheus ecosystem - perhaps a more general metrics library, with different exposition formats to choose from, could work? I also don't know which existing Haskell metrics libraries or Prometheus libraries exist or are good (but hackage shows there might be some).

I'm not sure what metrics are the best to expose. I bet PostgREST developers know more about essential PostgREST KPIs. But things like:

```
postgrest_query_execution_time (histogram)
postgrest_http_requests
postgrest_schema_reloads
...
```

steve-chavez:
Needs more discussion, but It could be a good idea!

One thing I'd like to note for now is that we can't use a `metrics` endpoint because it would conflict with users routes(one example of using a `metrics` table [here](https://gitter.im/begriffs/postgrest?at=5f036594bb149531edf2ec74)). So maybe we can come up with a prefix, like `/pgrst/metrics` or `/internal/metrics`.

sevagh:
That sounds good. Given PostgREST + NGINX is the recommended/common deployment, one could easily expose the internal metrics prefix at their desired location.

steve-chavez:
On https://github.com/PostgREST/postgrest/issues/1933, we were thinking of using a special header for this and avoid creating an extra route.

Seems Prometheus doesn't support adding headers for scraping though :disappointed: https://github.com/prometheus/prometheus/issues/1724

> PostgREST + NGINX is the recommended/common deployment

But since Nginx would be present, then I guess it's not a problem because it can map the header to a url. Another option in the future might be https://github.com/PostgREST/postgrest/issues/1909 as well. So a special header would do.

darora:
Hmm at least for liveness/health checks, we'd want to be hitting the postgrest instances directly, rather than going through nginx or a similar rewriting layer.

Even with an nginx setup, I'd imagine we generally have it load balancing between multiple postgrest instances transparently, so even for metrics you'd likely want to be hitting each instance directly, rather than going over nginx.

One alternative would be to use a secondary port to host endpoints for liveness/metrics etc. This would avoid creating a breaking change where we reserve e.g. `/internal` or a similar base path, and trivially allow users to not expose these endpoints externally.

rupurt:
[...]

PR #2129, opened 2022-01-20
Add pool stats to `/metrics` admin endpoint

steve-chavez:
As mentioned on https://github.com/PostgREST/postgrest/issues/2042#issuecomment-1012703514, this adds a `/metrics` endpoint on the admin server.

For now, it contains the database pool stats:

```http
GET /metrics

{"dbPoolStats":{"currentUsage":20,"creates":43,"highwaterUsage":20,"createFailures":0,"takes":102097}}
```

New metrics would be namespaced on their own json field.

With the addition of https://github.com/avanov/pool/pull/3 we'd also get the waiters in the queue (discussed on https://github.com/PostgREST/postgrest/issues/2042#issuecomment-1027621704)

```js
{ 
  "dbPoolStats": {
    "currentUsage":10,"creates":421,
    "highwaterUsage":10,"createFailures":0,
    "totalWaiters":483,"takes":190445
  }
}
```

### DIFF
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -50,6 +50,15 @@ let
           }
           { };
 
+      resource-pool-fork-avanov =
+        prev.callHackageDirect
+          {
+            pkg = "resource-pool-fork-avanov";
+            ver = "0.2.4.0";
+            sha256 = "0y5hk4wi2n5xzdb11jvb9f8mh3lmycjfyxii81kl6s412ir5gpm5";
+          }
+          { };
+
       hasql-dynamic-statements =
         lib.dontCheck (lib.unmarkBroken prev.hasql-dynamic-statements);
 
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -56,6 +56,7 @@ library
                       PostgREST.Logger
                       PostgREST.Middleware
                       PostgREST.OpenAPI
+                      PostgREST.Pool
                       PostgREST.Query.QueryBuilder
                       PostgREST.Query.SqlFragment
                       PostgREST.Query.Statements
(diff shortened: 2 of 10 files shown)


wolfgangwalther:
Before inventing our own metrics output format, I guess it would be better to use something that is already in-use. Note: I have no idea about those formats - so you might already be doing so?

I'm sure you have #1526 on your radar, i.e. Prometheus-style metrics.

Are there any other "standard" formats for metrics out there?

If not, we might as well do Prometheus-style if no other suggestion comes up.

And when we do that - I wonder whether https://github.com/fimad/prometheus-haskell can be of any help?

wolfgangwalther:
So it looks like prometheus uses a text-based format:
https://prometheus.io/docs/instrumenting/exposition_formats/

There's also [OpenMetrics](https://github.com/OpenObservability/OpenMetrics) which builds on the prometheus format and tries to standardize it.

I briefly thought about whether we could provide different output depending on an `Accept` header - but I don't really know which header prometheus will send. It doesn't seem to be configurable: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

I'd assume it sends a `Accept: text/plain` header or something similar - at least almost certainly not a `Accept: application/json`. So we could - in theory - implement a json metrics endpoint only when this is sent.

But I guess it would make more sense to evaluate the `prometheus-haskell` way.

steve-chavez:
I recall trying to use `prometheus-haskell` but I had issues because https://github.com/avanov/pool (and other `resource-pool` variants) maintained its own counters and it was not possible to sync them. The newest `hasql-pool` version doesn't use `resource-pool`(we're migrating to that one on https://github.com/PostgREST/postgrest/pull/2391/), so this PR is no longer relevant.

PR #2131, opened 2022-01-23
Fix `--example` configuration file

wolfgangwalther:
I mentioned `postgrest --example` in https://github.com/PostgREST/postgrest-docs/pull/493 - and then realized that I forgot to update the example config file in #2112. This PR does that - and improves the example file a bit more.

### DIFF
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -150,10 +150,6 @@ exampleConfigFile =
       |## Stored proc to exec immediately after auth
       |# db-pre-request = "stored_proc_name"
       |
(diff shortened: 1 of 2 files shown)


steve-chavez (approved):

wolfgangwalther:
Added a commit to fix the memory tests as mentioned in https://github.com/PostgREST/postgrest/pull/2112#issuecomment-1019843888.

wolfgangwalther:
The windows build currently fails, but I'm not sure whether we can actually do anything about that right now. Certainly not in this PR, as it's unrelated here.

PR #2133, opened 2022-01-24
Fix embedded filter silently getting ignored

### LINKED ISSUE #1771: Embedding filter gets silently ignored if the resource doesn't exist

steve-chavez:
Fixes #1771

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2107, Clarify error for failed schema cache load. - @steve-chavez
    + From `Database connection lost. Retrying the connection` to `Could not query the database for the schema cache. Retrying.`
  - #2120, Fix reading database configuration properly when `=` is present in value - @wolfgangwalther
+ - #1771, Fix silently ignoring filter on a non-existent embedded resource - @steve-chavez
 
 ## [9.0.0] - 2021-11-25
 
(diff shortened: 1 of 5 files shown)


laurenceisla (approved):
To keep consistency with other hints.

steve-chavez:
Windows build is failing but that's unrelated to this PR. Same issue noted on https://github.com/PostgREST/postgrest/pull/2131#issuecomment-1020125984.


```
error: mingw-w64-x86_64-postgresql: signature from "David Macek <david.macek.0@gmail.com>" is unknown trust
:: File /var/cache/pacman/pkg/mingw-w64-x86_64-postgresql-13.1-1-any.pkg.tar.zst is corrupted (invalid or corrupted package (PGP signature)).
error: failed to commit transaction (invalid or corrupted package (PGP signature))
Do you want to delete it? [Y/n] 
Errors occurred, no packages were upgraded.
```

Maybe it's a temporary error.

PR #2135, opened 2022-01-24
fix: Remove trigger functions from schema cache and OpenAPI output

wolfgangwalther:
Trigger functions can't be called directly from SQL and can't be called
via the /rpc prefix either - it makes no sense to expose them in the
OpenAPI output. And we don't need to cache them in the schema cache
either.

Best practice would be to keep the trigger functions in a non-exposed
schema anyway.


wolfgangwalther:
Since we don't have a CI test of `--dump-schema`, yet, I tested it manually. And I also confirmed it does not show up in the OpenAPI output either. Some kind of snapshot test for that wouldn't be bad in general, I guess.

steve-chavez (approved):
Fully agree! It doesn't make sense to expose trigger functions.

Issue #1089, opened 2018-04-01
Shortcomings in HTTP compliance


vfaronov:
Thank you for making PostgREST. It clearly attempts to comply with the HTTP protocol, but it falls short in a few places. I’m reporting these all in one issue, because I don’t know if you’re interested in fixing them. These problems are somewhat theoretical: they might trip up generic HTTP components (like caches), but it’s likely that none of your current users really suffer from them. If you prefer, I can split this issue into smaller ones.

I found all this playing with PostgREST 0.4.4.0 (f9e770b).

1\. PostgREST uses the [`Range`](https://tools.ietf.org/html/rfc7233#section-3.1) and [`Content-Range`](https://tools.ietf.org/html/rfc7233#section-4.2) headers defined in [RFC 7233](https://tools.ietf.org/html/rfc7233), but it doesn’t implement RFC 7233 properly. Firstly, PostgREST sends `Content-Range` on all kinds of responses, including 200 (OK) to `GET`, 201 (Created) to `POST`, and 204 (No Content) to `DELETE`. But in RFC 7233 `Content-Range` is only a feature of special *partial* responses, which are indicated by status code [206 (Partial Content)](https://tools.ietf.org/html/rfc7233#section-4.1), which in turn can’t happen unless the request has a `Range`. [RFC 7233 § 4.2](https://tools.ietf.org/html/rfc7233#section-4.2):

>    The Content-Range header field has no meaning for status codes that
   do not explicitly describe its semantic.

Secondly, the `Range` and `Content-Range` headers as used by PostgREST are invalid: they don’t include the `range-unit`.

This can confuse software like HTTP caches, which might not see how one 200 response to `GET /tasks` differs from another. (Conversely, if you fix it, a theoretical smart cache might even be able to combine different ranges on the fly.) One way to fix this might be to switch into “standards mode” upon receiving something like `Prefer: rfc7233`.

2\. PostgREST accepts `PATCH` requests with `Content-Type: application/json`. As explained in [RFC 5789 errata](https://www.rfc-editor.org/errata/eid3169), such patches are bad. A better patch format would be [`application/merge-patch+json`](https://tools.ietf.org/html/rfc7396), which has the semantics used by PostgREST, but only in the case of a single object (it doesn’t apply to arrays). PostgREST currently rejects patches with `Content-Type: application/merge-patch+json`. It might be a good idea to allow them, at least for single objects.

[...]

PR #2139, opened 2022-01-28
Return 405 Method Not Allowed for unsupported http verbs

wolfgangwalther:
Previously, an unsupported http method would return a 404 Not Found error. With this PR, those requests now return a proper 405 Method Not Allowed.

Related to #1089. Might also make #1826 easier to do.

Also removes the impossible error case for the previous `UnsupportedVerb` error, mentioned in https://github.com/PostgREST/postgrest/pull/1926#discussion_r774974102.

Along the way I did the following as well:
- Fixed `postgrest-git-hooks`
- Replaced a couple of cases of `elem` with pattern matching. Using `elem` will not show missing branches in code coverage. Using pattern matching will do so.
- Moved the spec tests to a few subfolders to make things hopefully easier to find.
- Sorted a few error types - again to make things easier to find.


wolfgangwalther:
(just rebase to fix the windows build)

steve-chavez (approved):
Nice. The new `Mutation` type is suitable.

Issue #1543, opened 2020-06-15
Composite type filtering 


jamesmstone:
### Environment

* PostgreSQL version: `12.2` (docker)
* PostgREST version: `7.0` (docker)
* Operating system:  docker

### Description of the feature request / issue

Is it possible to vertically filter ( and I guess also horizontally - although not needed for my use case ) on Composite Types: [Postgres Docs](https://www.postgresql.org/docs/current/rowtypes.html)

Imagine you have the following schema

```
CREATE TYPE complex AS (
    r       double precision,
    i       double precision
);

CREATE TABLE fav_numbers (
    num      complex,
    person     text
);

INSERT INTO fav_numbers VALUES (ROW(0.5, 0.5), 'A'),  (ROW(0.6, 0.6), 'B');
```

I am trying to filter `/fav_numbers` `where i > 0.5`

An example SQL query that would match looks like this:

`SELECT num, person from fav_numbers where i > 0.5`

I first tried the following request but it returns all rows:

`/fav_numbers?num.i=gt.0.5`

In a way, this is a bit like resource embedding

jamesmstone:
I guess another thing to consider with this is ordering composite type by sub field

steve-chavez:
Hey @jamesmstone.

> SELECT num, person from fav_numbers where i > 0.5

Hmm.. that one works for you? I had to do this:

(also according to pg [ROWTYPES-ACCESSING docs](https://www.postgresql.org/docs/13/rowtypes.html#ROWTYPES-ACCESSING))
```sql
SELECT num, person from fav_numbers where (num).i > 0.5;

-- this one also works
SELECT num, person from fav_numbers where num > '(0.5, 0)'::complex;
```

Thought the latter SQL statement would work through pgrst, but unfortunately it doesn't:
```
curl 'localhost:3000/fav_numbers?num=gt.(0,5,0)'
{"hint":null,"details":null,"code":"0A000","message":"input of anonymous composite types is not implemented"}
```
We would need to cast the filter to `::complex` internally for that to work.

> I guess another thing to consider with this is ordering composite type by sub field

That one also works in pg by using the `(column)` special syntax:

```sql
SELECT "test"."fav_numbers".* FROM "test"."fav_numbers" ORDER BY (num).r;
```

For now I'm thinking the solution would be to allow the special syntax in our Parsers. So the requests would be like this:

```bash
curl 'localhost:3000/fav_numbers?(num).i=gt.0.5'
[...]

PR #2145, opened 2022-02-02
feat: Support accessing array items and fields of composite types through json operators

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/issues/1543#issuecomment-1028162880.

Works fine... almost.

---

The array part is failing two tests. Filtering works, but select and order do not - and I don't understand why.

The following request (the combination of both failing tests):
```http
GET /arrays?select=a:numbers->0,b:numbers->1,c:numbers_mult->1->1,d:numbers_mult->2->3&order=numbers->0.desc
```
results in the following query:

```sql
WITH pgrst_source AS (
  SELECT
    to_jsonb("test"."arrays"."numbers")->$1 AS "a",
    to_jsonb("test"."arrays"."numbers")->$2 AS "b",
    to_jsonb("test"."arrays"."numbers_mult")->$3::int->$4 AS "c",
    to_jsonb("test"."arrays"."numbers_mult")->$5::int->$6::int AS "d"
  FROM "test"."arrays"
  ORDER BY to_jsonb("test"."arrays"."numbers")->$7 DESC
)
-- [...]
```

Note the missing `::int` for some of those query parameters. This will cause the unknown literal to be interpreted as `text` - and the `->` and `->>` operators to fail accessing the array properly.

I just have no idea why the parser does that. And it looks random, too, because some of the `::int` are there...

### DIFF
--- a/test/spec/Feature/Query/JsonOperatorSpec.hs
+++ b/test/spec/Feature/Query/JsonOperatorSpec.hs
@@ -103,7 +103,7 @@ spec actualPgVersion = describe "json and jsonb operators" $ do
 
       it "can get array of objects" $ do
         get "/json_arr?select=data->0->>a&id=in.(5,6)" `shouldRespondWith`
-          [json| [{"a":"A"}, {"a":"[1,2,3]"}] |]
+          [json| [{"a":"A"}, {"a":"[1, 2, 3]"}] |]
           { matchHeaders = [matchContentTypeJson] }
         get "/json_arr?select=data->0->a->>2&id=in.(5,6)" `shouldRespondWith`
           [json| [{"a":null}, {"a":"3"}] |]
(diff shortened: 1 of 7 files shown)


wolfgangwalther:
> The array part is failing two tests. Filtering works, but select and order do not - and I don't understand why.

I do now, and it's fixed.

There was a bug in the `JsonPath` parser we before: When a numeric value for `->` or `->>` was followed by a `,` (as in the `?select=` case) or by a `.` (as in the `?order=` case) - it was treated as a text literal instead. Added some doctests to while fixing it.

test/spec/Feature/Query/JsonOperatorSpec.hs
```diff
@@ -103,7 +103,7 @@ spec actualPgVersion = describe "json and jsonb operators" $ do
 
       it "can get array of objects" $ do
         get "/json_arr?select=data->0->>a&id=in.(5,6)" `shouldRespondWith`
-          [json| [{"a":"A"}, {"a":"[1,2,3]"}] |]
+          [json| [{"a":"A"}, {"a":"[1, 2, 3]"}] |]
```
wolfgangwalther, line 106: The test output changes slighty because `to_jsonb` adds a bit of whitespace.

steve-chavez: I see, and `to_jsonb` is used because of:

> -- Using to_jsonb instead of to_json to avoid missing operator errors when filtering:
> -- "operator does not exist: json = unknown"

steve-chavez (approved):
Simple and elegant solution :100:

wolfgangwalther:
Added docs in https://github.com/PostgREST/postgrest-docs/pull/501

PR #2152, opened 2022-02-05
Remove non-callable functions from schema cache and OpenAPI output

wolfgangwalther:
This allows to create virtual columns in the exposed schema, as suggested in our docs, without "polluting" the OpenAPI output with some fake RPCs. To do this the virtual columns need a single **unnamed** argument.

Example:

```sql
CREATE TABLE people (lastname TEXT, firstname TEXT);

-- This virtual column will have an RPC listed in the OpenAPI output
CREATE FUNCTION fullname (p people) RETURNS TEXT
RETURN p.firstname || p.lastname;

-- Before this PR: This virtual column would have had a fake RPC listed, but couldn't be called directly.
-- After this PR: This virtual column will have no fake RPC listed
CREATE FUNCTION initials (people) RETURNS TEXT
RETURN left($1.firstname, 1) || left($1.lastname, 1);
```

Tested locally by observing changes to the `--dump-schema` output. We had a few fake RPCs still in there, that are now gone.

TODO:
- [x] Mention this in the docs (https://github.com/PostgREST/postgrest-docs/pull/503)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1771, Fix silently ignoring filter on a non-existent embedded resource - @steve-chavez
  - #2135, Remove trigger functions from schema cache and OpenAPI output, because they can't be called directly anyway. - @wolfgangwalther
  - #2101, Remove aggregates, procedures and window functions from the schema cache and OpenAPI output. - @wolfgangwalther
+ - #2152, Remove functions, which are uncallable because of unnamend arguments from schema cache and OpenAPI output. - @wolfgangwalther
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
  - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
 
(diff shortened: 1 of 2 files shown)


steve-chavez:
> this virtual column will have a fake RPC output - but can't be called directly

Perhaps it's not that useful, but the above it's not true as the function can be called by using a record literal:

```bash
http POST "localhost:3000/rpc/fullname" <<JSON
{"p": "(Doe,John)"}
JSON
```

wolfgangwalther:
> > this virtual column will have a fake RPC output - but can't be called directly
> 
> Perhaps it's not that useful, but the above it's not true as the function can be called by using a record literal:
> 
> ```shell
> http POST "localhost:3000/rpc/fullname" <<JSON
> {"p": "(Doe,John)"}
> JSON
> ```

Hm. I can't remember why I put the comment there. Because... those are exactly the function that we **keep**, right?

The whole term "fake RPC" is actually from.. **before** the changes I made here.

wolfgangwalther:
I updated my comments in the SQL snippets. They should now reflect what is really going on.

steve-chavez (approved):

PR #2153, opened 2022-02-05
Remove aggregates, procedures and window functions from schema cache and OpenAPI

### LINKED ISSUE #2101: Stored procedures not supported, but still shown in OpenApi output
Although postgrest picks up the RPC name for a stored procedure (as opposed to a function, which is different in PostgreSQL), it does not appear to invoke it correctly.
For example, I have the following stored procedure which when called, returns an error:

```sql
CREATE OR REPLACE PROCEDURE public.items_create_new_item_for_category(IN category_id bigint, IN item_name character varying)
    LANGUAGE 'sql'
    
AS $BODY$
with new_item as (
  insert into items (id, name) values (default, item_name)
  returning id
)
insert into categories_items (category_id, item_id) 
  select (category_id) as category_id, new_item.id as item_id
	from new_item;
$BODY$;
```

The generated types based on OpenAPI show the expected rpc endpoint like this:
```JS
  "/rpc/items_create_new_item_for_category": {
    post: {
      parameters: {
        body: {
          args: {
            item_name: string;
            category_id: number;
          };
        };
        header: {
          /** Preference */
          Prefer?: parameters["preferParams"];
        };
      };
      responses: {
        /** OK */
        200: unknown;
      };
    };
  };
```

But issuing an RPC I get this error:
URL: http://localhost:54321/rest/v1/rpc/items_create_new_item_for_category
Payload: `{"category_id":"1","item_name":"th"}`

Response: 
```JS
{
code: "42809",
details: null,
hint: "To call a procedure, use CALL.",
message: "public.items_create_new_item_for_category(category_id=> bigint, item_name=> character varying) is a procedure"
}
```

I've searched the documentation and issues on github but haven't found anything saying that Stored Procedures in particular are not supported, so I expected them to work. Can someone please point me to the err of my ways, or otherwise help figure out where this bit should be documented?

wolfgangwalther:
Aggregates and Window functions can't be called as RPCs in a useful way.

Until procedures are implemented in #1976 it makes no sense to list them in the OpenAPI output.

Resolves #2101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2120, Fix reading database configuration properly when `=` is present in value - @wolfgangwalther
  - #1771, Fix silently ignoring filter on a non-existent embedded resource - @steve-chavez
  - #2135, Remove trigger functions from schema cache and OpenAPI output, because they can't be called directly anyway. - @wolfgangwalther
+ - #2101, Remove aggregates, procedures and window functions from the schema cache and OpenAPI output. - @wolfgangwalther
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
  - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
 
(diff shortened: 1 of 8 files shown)


wolfgangwalther:
The change made another bug visible, we hadn't noticed before: `--dump-schema` was always running with the `minimalPgVersion`, because the pg version is only queried in the connection worker, but not in the CLI case. I fixed this by querying the pg version in the db schema cache queries to make sure they **always** run with the right version, and therefore the right queries.

steve-chavez (approved):

PR #2158, opened 2022-02-07
fix: Keep working when EMFILE is reached

### LINKED ISSUE #2042: Network.Socket.accept: resource exhausted (Too many open files)

steve-chavez:
Closes #2042.

### TODO

- [ ] Admin docs about timeouts and EMFILE

### DIFF
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -50,6 +50,15 @@ let
           }
           { };
 
+      warp =
+        lib.dontCheck (prev.callHackageDirect
+          {
+            pkg = "warp";
+            ver = "3.3.19";
+            sha256 = "0y3jj4bhviss6ff9lwxki0zbdcl1rb398bk4s80zvfpnpy7p94cx";
+          }
+          { });
+
       hasql-dynamic-statements =
         lib.dontCheck (lib.unmarkBroken prev.hasql-dynamic-statements);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2101, Remove aggregates, procedures and window functions from the schema cache and OpenAPI output. - @wolfgangwalther
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
  - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
+ - #2042, Keep working when EMFILE(Too many open files) is reached. - @steve-chavez
 
 ### Changed
 
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -113,7 +113,7 @@ library
                     , wai                       >= 3.2.1 && < 3.3
                     , wai-cors                  >= 0.2.5 && < 0.3
                     , wai-extra                 >= 3.1.8 && < 3.2
-                    , warp                      >= 3.2.12 && < 3.4
+                    , warp                      >= 3.3.19 && < 3.4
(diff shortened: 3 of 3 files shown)


nix/overlays/haskell-packages.nix
```diff
@@ -50,6 +50,15 @@ let
           }
           { };
 
+      warp =
+        lib.dontCheck (prev.callHackageDirect
```
wolfgangwalther, line 54: Just curious - what does the `lib.dontCheck` do?

I see we are using it on the `hasql-` packages - but we are not using it on the `wai-` packages.

wolfgangwalther (approved):

wolfgangwalther:
Do you think we can craft an io test that runs into the EMFILE problem without the patch?

steve-chavez:
We'd be testing Warp if we do that, so I think we shouldn't. However, basically we need to dynamically change the file descriptor limit(can be done with [prlimit](https://unix.stackexchange.com/a/422732/278242)) to something like 1, make requests and see postgrest fail.

steve-chavez:
> make requests and see postgrest fail.

Doing a `prlimit --pid <pid> --nofile=10:1024` and making a new request:

- old version: postgrest dies and prints the `Network.Socket.accept: resource exhausted (Too many open files)` error.
- new version: new requests wait and when increasing the fds(`prlimit --pid <pid> --nofile=1024:1024`) they get a response.

(Ideally the EMFILE error would bubble up to the client instead of waiting but at least we keep running)

steve-chavez:
The above looks like a recovery test, so I've listed it on #1766 to add it later if it's worth it.

wolfgangwalther:
> (Ideally the EMFILE error would bubble up to the client instead of waiting but at least we keep running)

Agree with both!

Issue #2125, opened 2022-01-15
Implement new `QUERY` method to pass arguments in request body for read request


telewak:
Thank you for this wonderful and magical api...  i would like to send alot of arguments while doing a GET request.
Does PostgREST as an API have a limit on the url length ? Is there possibility to POST a json object to the GET endpoint?

wolfgangwalther:
See https://github.com/PostgREST/postgrest-docs/issues/417.

`POST` to `GET` is not implemented. Will keep this open to track the `QUERY` proposal by @steve-chavez in the thread above.

steve-chavez:
Even the new QUERY http method won't help with cases where `DELETE` has a big `in` filter. PATCH would suffer from this too.

Perhaps we should have a POST with a special body + content type(similar to https://github.com/PostgREST/postgrest/issues/465) for these cases?

Come to think of it, QUERY would help because it will contain the DELETE/UPDATE inside.

steve-chavez:
> Come to think of it, QUERY would help because it will contain the DELETE/UPDATE inside.
> I don't understand what you're saying here.

The example [here](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html#section-4.2):

> QUERY /contacts HTTP/1.1
> Host: example.org
> Content-Type: example/query
> Accept: text/csv

> select surname, givenname, email limit 10

Hints that we could have a `delete where in (1,2,3,4,...)` or `update set name = 'xxx', status = 'yy' where in (1,2,3,4,..)` in the body? Maybe I'm wrong..

steve-chavez:
> We'd use that for QUERY and DELETE by default -

You mean we would do an HTTP DELETE with a body? I thought that was frowned upon according to [this](https://stackoverflow.com/a/60357432/4692662).

steve-chavez:
> Allow sending a POST to the root endpoint. This request would contain basically the whole request, including method/action, filters, headers, body etc.

Not adamantly opposed, but that solution feels like "chucking it all".

> Is it only the filters that go in there or the full path? I feel like it's hard to draw a line here

The problem really is the `in` filter, the only I've seen reported to grow that big(like a thousand values) for both GET and DELETE.

So perhaps we could do something like:

```http
POST /projects?status=eq.active
Content-Type: application/vnd.pgrst.big-in-filter+json

{
  "method": "DELETE"
, "in": [1,2,3,4,5,6]
}
```

[...]

Issue #2066, opened 2021-12-03
Improvements to query language


wolfgangwalther:
We have discussed a few shortcomings of our current query language (expressions in the query string). Taking those suggestions together - and adding a few more points, that seem to be inconsistent to me in the current implementation.

I wonder whether we should try to implement all those improvements as non-breaking separated changes - or just invent a query language v2, where we can chose via config option which language to use. Maybe #1804 can lead us to have a cleanly separated `QueryLanguage` module that we can replace with a new module.

Here we go:
- Change operator syntax from `col=op.val` to `col.op=val`: #1970. This will allow to use `swagger-ui` to send requests. It will also allow plain-html requests without javascript like this:
   ```html
   <form action="/table?select=html">
      <input type="text" name="title.eq" />
    </form>
    ```
    This was not possible before, because the value would have to be constructed dynamically.

- Separating filter logic from values: #1970. This will allow doing more complex queries in `swagger-ui` involving logic conditions, too. It will also simplify a lot of our qouting requirements already. Let's put all logic into a `logic=...` parameter. Maybe `filters=` is even better.

- Replace current `in` with `col.eq.any=<array>` as suggested in #1943. I'd go one step further and remove `in` completely. We are using `ANY` under the hood anyway and together with `logic=`, we would **remove all manual quoting** we'd need to do. We'd rely only on PG array quoting.

- Move `select` and `order` to parameters of the mimetype in the `Accept` header. Both parameters only shape the response - but the underlying **entity** is the same. As such, I think they don't really fit in the query string, which should describe which entity to receive. This would also remove the strange interaction we have between the `select=` and the `Accept` header already: When you request a custom mimetype, e.g. `application/octet-stream` or `text/plain`, the whole `select=` doesn't make any sense. So the availability of `select` is already dependent on the mediatype.
    Example: Both of the following requests will return the same person. Returning fewer or more columns is not much different to just returning the same entity in a different format.
   ```http
   GET /people?uid.eq=123&select=*
   GET /people?uid.eq=123&select=lastname,firstname
   ```
   would become
   ```http
   GET /people?uid.eq=123
   Accept: application/json;select=*

   GET /people?uid.eq=123
   Accept: application/json;select=lastname,firstname
   ``` 
[...]

Issue #1970, opened 2021-09-30
On swagger with uuid id: failed to parse filter


artsra:
### Environment

* PostgreSQL version: (if using docker, specify the image)
* PostgREST version: (if using docker, specify the image)
* Operating system: MacOS Big Sur 11.6

### Description of issue

On swagger I try to GET a record from a table where the primary key is a UUID.
As a result I get:
```
{
  "details": "unexpected \"3\" expecting \"not\" or operator (eq, gt, ...)",
  "message": "\"failed to parse filter (3922488c-18bc-45d8-baa2-bdb10cd90e37)\" (line 1, column 1)"
}
```

I think this error originates from postgrest. 

On the other hand if I try a filter like 'eq.3922488c-18bc-45d8-baa2-bdb10cd90e37' the action is refused by swagger, because the string with filter does not follow the uuid format.

How do I resolve this?

wolfgangwalther:
This will not work right now. We had some ideas in https://github.com/PostgREST/postgrest/pull/1949#issuecomment-922875327 and https://github.com/PostgREST/postgrest/pull/1949#issuecomment-923376956 that would make this work - but those are far from being implemented, yet.

steve-chavez:
I see it now. So as Wolfgang mentioned, our current syntax is a problem for swagger-ui. Because it's like:

```
id=eq.2e1183bf-123b-4a48-982f-d9c78584ce08
```

The value being `eq.2e1183bf-123b-4a48-982f-d9c78584ce08`, which is an invalid uuid. If we were to switch our operators to the left side:

```
id.eq=2e1183bf-123b-4a48-982f-d9c78584ce08
```

Then the value would be valid.

Function calls should work fine with swagger-uid though, because they use the required form of `arg=val`.

---

Seems that `or`(`or=(age.gte.14,age.lte.18)`) will always be a problem, even if we switched the syntax. Complex boolean logic for swagger-ui would have to be done via functions.

steve-chavez:
> Seems that or(or=(age.gte.14,age.lte.18)) will always be a problem, even if we switched the syntax. Complex boolean logic for swagger-ui would have to be done via functions.

I think the above problem with `or` could be fixed as well.

We could change this:

```http
GET /people?or=(grade.gte.90,student.is.true)
``` 

To:

```http
GET /people?grade.gte=90&student.is=true&or=grade.gte,student.is
```

Or as a shorthand:

```http
GET /people?grade.gte=90&student.is=true&or=*
[...]

Issue #1909, opened 2021-08-13
Provide a way to customize the URL structure


steve-chavez:
### Problem

We recommend customizing the URL structure [through Nginx](https://postgrest.org/en/v8.0/admin.html#alternate-url-structure), however doing so is complicated and requires additional tools. This can be seen on [ppKrauss/PostgREST-writeAPI](https://github.com/ppKrauss/PostgREST-writeAPI), which uses [mustache templating](https://github.com/ppKrauss/PostgREST-writeAPI/blob/master/nginx-tpl/tpl01-baseBreak.mustache) plus a [nodejs script](https://github.com/ppKrauss/PostgREST-writeAPI/blob/master/writeApi.js) to generate the Nginx rewrite config.

Additionally, changing the URL in the proxy will not be reflected in our OpenAPI spec - which would need rewriting on the proxy as well, which is also more work.

Relying solely on the proxy for this **hurts our extensibility**, ideally PostgREST should be able to adapt to other standards, like more classical REST with nested routes or maybe more complex ones like [Odata](https://www.odata.org/getting-started/basic-tutorial/); and do so outside of PostgREST core.

Note that we do have a way forward to customize the response body(see [this discussion regarding JSON-LD](https://github.com/PostgREST/postgrest/discussions/1790)), but not the URL structure.

### Proposal

Provide a way to customize the URL through the config file or through postgrest-contrib. It could be like so:

```sql
select pgrst.url_rewrites(
  -- more classic REST
  -- /persons/1
  'persons/${value}', 'people?id=eq.${value}'
  -- /persons/1/companies
, 'persons/${value}/companies', '/people?select=*,companies(*)&id=eq.${value}'
  -- bring an RPC to root endpoint
, 'articles/${year}', 'rpc/get_articles?year=${year}'
  -- special route
, 'christmas_2021', 'rpc/get_articles?year=2021&month=december'
);
```

### Related

- https://github.com/PostgREST/postgrest/issues/1086

---

@wolfgangwalther I think you had an idea about a "routing refactor", this may be related.

wolfgangwalther:
> @wolfgangwalther I think you had an idea about a "routing refactor", this may be related.

Absolutely, I had customizing of the URLs for endpoints in mind with that, too.

gautam1168:
[...]

PR #2164, opened 2022-02-10
feat: max-changes prefer header to limit mutations

### LINKED ISSUE #2156: Conditional delete/update based on rows affected
### Problem

Even with [safeupdate](https://github.com/eradman/pg-safeupdate), a `DELETE /employees?salary=gt.1000` can still delete a large amount of rows.

We already have code for handling a special case of 1 affected row for update/delete, when using `application/vnd.pgrst.object+json`(CTSingularJSON).

https://github.com/PostgREST/postgrest/blob/bd62008e1c6e752c11c4c5a1a9bcebd7a439e69d/src/PostgREST/App.hs#L569-L574

We can extend this logic to accept an arbitrary number of affected rows and reject the request if it doesn't comply. With this a user can ensure that an expected number of rows will be updated/deleted.

### Proposal

Use a new Prefer header to specify the affected rows, a couple of options:

1) `Prefer: rows-affected=<number>`

Specifies the affected rows in the header value. The default behavior we have being like `Prefer: rows-affected=unlimited/infinity`. A `Prefer: rows-affected=1` would be equivalent to `application/vnd.pgrst.object+json`.

2) `Prefer: rows-affected=same-as-limit`

This would reuse the `?limit=<number>` query param value. The default behavior would also be "unlimited". To do an equivalent `application/vnd.pgrst.object+json` the request would be `curl -X DELETE "host/tbl?limit=1" -H "Prefer: rows-affected=same-as-limit"`. We could also have a `rows-affected=less-than-limit` to use the limit as a threshold instead of an exact value. Though maybe just a `Prefer: rows-affected=limited` with a behavior of `rows affected <= limit` would be enough.

With this option, to globally disallow unlimited deletes/updates, we could change the default Prefer to `rows-affected=limited` and combine it with `max-rows-mutated=<number>`(https://github.com/PostgREST/postgrest/issues/2155). If users wants to do a full table delete, they'd need to be explicit and specify a `Prefer: rows-affected=unlimited`.

### Notes

The [Prefer RFC says](https://datatracker.ietf.org/doc/html/rfc7240#section-2):

>  A server that does not recognize or is unable to comply with
   particular preference tokens in the Prefer header field of a request
   MUST ignore those tokens and continue processing instead of signaling
   an error.

Before I understood that as "Prefer cannot reject a request", but [handling=strict](https://datatracker.ietf.org/doc/html/rfc7240#section-4.4) says:

>  The "handling=strict"
   preference can be used to indicate that, while any particular error
   may be recoverable, the client would prefer that the server reject
   the request.

So I think the first paragraph above is mostly saying that we should ignore unrecognized/conflicting prefers, we can reject requests with Prefer.

steve-chavez:
Closes #2156. Basically:

```bash
http PATCH localhost:3000/projects "Prefer: max-changes=4" <<JSON
{"name": "PopOS"}
JSON

HTTP/1.1 400 Bad Request

{
    "details": "Results contain 57 rows changed but the maximum number allowed is 4",
    "message": "The maximum number of rows changed per operation was surpassed"
}
```

### Tasks

- [ ] Decide whether a header or query parameter will be used
- [x] max-changes for PATCH
- [ ] max-changes for DELETE
- [ ] max-changes for POST?
- [ ] Docs


wolfgangwalther:
I wonder whether `Prefer` is really the best choice here. For everything else we do with `Prefer`, there is not much harm done when the server doesn't understand / use it. But here, we really want the server to **not** proceed, when it doesn't understand the request. It's missing the "optional" part, [for which `Prefer` was designed](https://datatracker.ietf.org/doc/html/rfc7240#section-2).

> The Prefer request header field is used to indicate that particular
> server behaviors are preferred by the client but are not required for
> successful completion of the request. Prefer is similar in nature to
> the Expect header field defined by [Section 6.1.2 of [RFC7231]](https://datatracker.ietf.org/doc/html/rfc7231#section-6.1.2) with
> the exception that servers are allowed to ignore stated preferences.

~Unfortunately I don't have a better way to express this right now~. `Expect` headers have the "requiredness" factor - but they require each hop (proxy) to understand those headers. So that won't ever work.

---

*thinking a little bit*

So what is this feature actually about?

I guess the idea is to add a safe-guard to a request, for the case that the query string does not describe the "resource" that we actually want to target, well enough.

A resource could be a single entity - or a number of entities. And especially in the latter case, the descriptors for that resource, that we use through filters could be wrong. So basically we want to add more information about the resource we're requesting. Basically some more restrictions that are not regular filters.

Of course this leads us to `?limit` directly, but we already discussed that in the issue: It's too close to both SQL `LIMIT` and the `Range` header semantically, which is something that this is not about.

So this is basically a way of saying "please change the 5 entities that have ids between X and Y to the following" instead of "please change the entities that have ids between X and Y to the following".

I'm now thinking the answer is somewhere in the query string.

We don't need to support any nesting here, because those mutations only apply to the top-level resource.

This reminds me of one my ideas in #2066:

> Change the `Accept: application/vnd.pgrst.object+json` way to request a single resource to a [path parameter](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) something like this:
>
> ```http
> GET /people;single?uid.eq=123
> GET /people;one?uid.eq=123
> ```

This also makes sense somehow, because the `application/vnd.pgrst.object+json` currently does something very similar - it rejects every request that would affect fewer or more than 1 row, right?

So maybe something like:

```http
PATCH /people;1?uid=eq.123
PATCH /people;0-3?age=gt.100
```

The first one would be an exact match, much like `application/vnd.pgrst.object+json` - but more flexible. The second one would be a range of acceptable number of rows.

steve-chavez:
Hm, on https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1031734903 we agreed that using `Prefer` for rejecting requests was fine, since the [handling=strict](https://datatracker.ietf.org/doc/html/rfc7240#section-4.4) in the RFC already mentions it can be used for that.

> Expect headers have the "requiredness" factor - but they require each hop (proxy) to understand those headers. So that won't ever work

Yeah, the Expect header has been useless for some time now because it was confirmed Nginx doesn't proxy pass it https://github.com/PostgREST/postgrest/issues/748#issuecomment-263037419.

steve-chavez:
> PATCH /people;1?uid=eq.123
> PATCH /1-3,people?age=gt.100

[Historically](https://postgrest.org/en/stable/api.html)(`Why not provide nested routes?..`) we have frowned upon using path parameters, I'm surprised to see the [RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) containing a section on using "," and ";" on a path segment:

> URI producing applications
   **often** use the reserved characters allowed in a segment to delimit
   scheme-specific or dereference-handler-specific subcomponents.  For
   example, the semicolon (";") and equals ("=") reserved characters are
   often used to delimit parameters and parameter values applicable to
   that segment.  The comma (",") reserved character is often used for
   similar purposes.  For example, one URI producer might use a segment
   such as "name;v=1.1" to indicate a reference to version 1.1 of
   "name", whereas another might use a segment such as "name,1.1" to
   indicate the same.

This [blog post](https://doriantaylor.com/policy/http-url-path-parameter-syntax#:~:text=A%20URI%20path%20parameter%20is,be%20constructed%20out%20of%20band.) also mentions a path segment can be used like that.

Do note that the RFC says "applications often use", which means it's taking inspiration from previous cases. Have you ever seen an API that uses parameters in a path segment? I have not and I doubt anyone would call that design "elegant", I think it breaks clients expectations(and perhaps some libraries/tools as well). 

> I wonder whether Prefer is really the best choice here. For everything else we do with Prefer, there is not much harm done when the server doesn't understand / use it. 

If this is about consistency, then diverging from not using path parameters is much more inconsistent.

I also think `Prefer` is not ideal, but unless we invent a new header, I don't see any other better way around it.

steve-chavez:
> and perhaps some libraries/tools as well

One such case might be OpenAPI, I don't see a way on [their docs](https://swagger.io/docs/specification/paths-and-operations/) to specify a path segment parameter.

Just found this [comment](https://github.com/OAI/OpenAPI-Specification/issues/2653#issuecomment-1034034385) that indicates this style is legacy.

wolfgangwalther:
> Hm, on [#2156 (comment)](https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1031734903) we agreed that using `Prefer` for rejecting requests was fine, since the [handling=strict](https://datatracker.ietf.org/doc/html/rfc7240#section-4.4) in the RFC already mentions it can be used for that.

Yes, I agreed with you that throwing errors on `Prefer` is fine. It still is. This is not the point here, though. The point here is, that when you make a request with this header, you **expect** the server to error out when it **does not support this prefer header**. I.e. assume you are making such a request to an older PostgREST version, because you read about the feature in the docs. You **don't** want to have PostgREST silently ignore this, because then you are at risk of loosing data.

wolfgangwalther:
> Just found this [comment](https://github.com/OAI/OpenAPI-Specification/issues/2653#issuecomment-1034034385) that indicates this style is legacy.

It does **not** say the style is legacy. It does say the API is legacy (and therefore they can't change it) and it uses those parameters.

steve-chavez:
Ah, I've also found the `handling=strict` [suggested](https://github.com/PostgREST/postgrest/issues/748#issuecomment-265102653) in that old thread.

Perhaps we could enforce its presence for `max-changes` to work. So `Prefer: max-changes=20, handling=strict` would err as proposed here but `Prefer: max-changes=20, handling=lenient`(or just `Prefer: max-changes=20`) would always make the request succeed and ignore `max-changes`.

wolfgangwalther:
> One such case might be OpenAPI, I don't see a way on [their docs](https://swagger.io/docs/specification/paths-and-operations/) to specify a path segment parameter.

The link you gave lead me to https://swagger.io/docs/specification/describing-parameters/#path-parameters.

This specifically mentions a "simple" style:
> simple-style – comma-delimited, such as /users/12,34,56

wolfgangwalther:
> Hm, wait, I don' see the point here. Say you're specifying a `Prefer: tx=rollback` on an old postgrest, it just ignores it and the client will lose data. Isn't that the same problem?

Hm. Yes, nice example. I would agree - it is the same problem. Although, hopefully, less likely to occur, because that feature is meant for things like unit testing the api etc. - while the feature we are discussing here is for production use.

But yes, in hindsight, this has the same problem. If I knew back then what I know now, I might have looked for a different approach. Not sure whether I would have found one or not.

steve-chavez:
If anything, perhaps we should have handled `handling=strict` since the beginning. So `Prefer: unknown=val, handling=strict` will always fail. That would solve the problem, we could do it from now.

wolfgangwalther:
> If anything, perhaps we should have handled `handling=strict` since the beginning. So `Prefer: unknown=val, handling=strict` will always fail. That would solve the problem, we could do it from now.

No, that's not an option. The [RFC](https://datatracker.ietf.org/doc/html/rfc7240#section-2) specifically says:

> A server that does not recognize or is unable to comply with
> particular preference tokens in the Prefer header field of a request
> MUST ignore those tokens and continue processing instead of signaling
> an error.

steve-chavez:
> I understand you might be opposed to using path parameters here, because it's:

It's just more complexity for the users... having more ways to do the same thing...

> Although, hopefully, less likely to occur, because that feature is meant for things like unit testing the api etc. - while the feature we are discussing here is for production use.

The tx=rollback is already being used by [Supabase users](https://github.com/supabase/postgrest-js/pull/244), once the feature it's out you never know how the feature might be exploited.

> Say you're specifying a Prefer: tx=rollback on an old postgrest, it just ignores it and the client will lose data. Isn't that the same problem?
> Hm. Yes, nice example. I would agree - it is the same problem.

That's why we have the `Preference-Applied` on the response right? I don't see an issue with old versions not supporting new Prefer headers then. We can have a more detailed section on the docs about "Preferences".

> tbh, I think it is elegant.
> It reads nicely. It uses more of what the URI spec allows us to do. It's consistent and conceptually sound. Yeah - that's elegant.

It reminds me of my critique against Odata [here](https://github.com/PostgREST/postgrest/issues/1398#issuecomment-548354861), those `;max-updated=123` are more unreadable/unfamiliar in URLs than the `$` symbols. But I guess the point here is that this is subjective, so no need to discuss it further.

> I wouldn't want to put them into the query string, wasting even more keywords there, though - but that would be a similarly consistent approach but that would be a similarly consistent approach.

Yeah, a new query string param would be more consistent. However I don't see any problem with the new Prefer, considering  the `Preference-Applied` and all.

steve-chavez:
I've been looking at some additions to SQL for DELETE. I've found that for example MySQL has [DELETE LIMIT](https://www.techonthenet.com/mysql/delete_limit.php) and SQL Server has [DELETE TOP](https://docs.microsoft.com/en-us/sql/t-sql/statements/delete-transact-sql?view=sql-server-ver15)(you can even specify a percentage of rows there).

Nothing for failing in case a threshold is exceeded though. However, SQL Server also has the [OPTION clause](https://docs.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=sql-server-ver15), could we use the same idea for our own options?

steve-chavez:
Another option might be borrowing [ROWCOUNT](https://docs.microsoft.com/en-us/sql/t-sql/functions/rowcount-transact-sql?view=sql-server-ver15)(`number of rows affected or read. Rows may or may not be sent to the client`) from SQL Server, this has less chance of conflicting with an existing column name. So we could say:

```http
DELETE /projects?rowcount=lte.10
```

It's a really special case though, because we'd fail if the filter doesn't pass.

steve-chavez:
Those seem like the path forward for now(header or new query param). In general, I think a new URI syntax(whether adding nested routes or path segments paremeters) needs much more discussion(like on https://github.com/PostgREST/postgrest/issues/2066) and consensus. I don't see myself trying to add new syntax here.

steve-chavez:
On this particular case though, I think we'd have to ping begriffs so he can chime in on the REST syntax. For me, I just see elegance/simplicity in how he [originally designed](https://begriffs.com/posts/2014-12-30-intro-to-postgrest.html)(despite we know now that `id=eq.val` should be changed) it and adding path parameters just adds complexity, for both users and us. That's why I'm opposed to adding it.

Now, if there would be no time limit for this discussion. I think the right way to do it would be to propose a new header to IANA, a replacement for `Expect` that doesn't ignore values as the `Prefer` header <- which is the main opposition to the `Prefer: max-changes`.

steve-chavez:
> You assume this represents the resource [6,7,8,9,10]. But you didn't know that somebody else added some more items, so the same query now represents a different resource instead: [6,7,8,9,10,11]. Yeah URIs change, that's fine.
> DELETE /5items?id=gt.5
> This is not meant to represent the syntax I proposed but just a different URI with a different meaning: that the resource we are requesting is described by length 5.
> What would be the answer to this query? A 404 Not Found of course, because this resource does not exist anymore. There are resources /6items?id=gt.6 and /5items?id=gt.5&id=lte.10 - but not the one we requested.

Makes a lot of sense, seems that mechanism could be useful for GETs as well. I think this concept now fits exactly with the `row_count` I mentioned above.

```http
GET /items?id=gt.5
[6,7,8,9,10,11]
```
```http
GET /items?id=gt.5&row_count=eq.5
404
```

I think `row_count` is unlikely to cause a breaking change, note that it's already used [here](https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-CURRENT-DIAGNOSTICS-VALUES).

wolfgangwalther:
I thought about this a little bit more, and even before your latest comments - which are well in line with what I am proposing now - I came to the following conclusion: I actually suggested two separate things, namely:
- Making $subject a part of the URI due to "what is a resource" considerations
- Using different parts of the URI (path parameters) for things in the query string that are "not a filter".

The second suggestion is not something that we can solve for just this specific case, but we need to discuss it in a bigger context. The basic question is "how many reserved keywords can we remove from the querystring?". If we were to start from scratch, the best answer would be "all" - so that we could use arbitrary names for filtering. But this is for sure better discussed in #2066 - and not here.

For this PR my main point is about headers vs URI. I have thought about this a little bit more. Our current usage of URI features and syntax certainly means we should implement this as a query string parameter - this is how we do those things right now. Even though I don't know any specific examples right now, I can see this general concept of "higher level filters", so to say, being extended in the future with more than just `row_count`. For this reason, I would like to find a way to in the query string that will avoid reserving even more keywords in the future - maybe we can create some kind of a namespace now.

I thought a little bit more of what that "higher level filters" concept is about here and I came up with the following. Our SQL queries have a general simplified structure of:

```sql
SELECT json_agg(<select>)
FROM <endpoint>
WHERE <filters>;
```

I.e. we have an aggregation at the end, and before that we have our regular filters.

I think it would help to think about those new filters like `row_count` as something like a `HAVING` clause, i.e. filters that are applied **after** the aggregation phase:

```sql
SELECT json_agg(<select>)
FROM <endpoint>
WHERE <filters>
HAVING count(*) BETWEEN x AND y;
```

In fact, this very much looks like a query similar to what we need for $subject.

Maybe we can find a syntax for this, that would allow extending it to more general aggregation scenarios and `HAVING` filters later on? I guess we should think about how the syntax for https://github.com/PostgREST/postgrest/issues/915#issuecomment-1026720780 should end up looking. We don't need to implement that right away, but if we can implement $subject in a way that can easily be extended to support other aggregation cases, too... that would be great.

TLDR: I think the correct solution to $subject is actually aggregation + `HAVING` filters.

---

> I think `row_count` is unlikely to cause a breaking change, note that it's already used [here](https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-CURRENT-DIAGNOSTICS-VALUES).

This is a much better argument for `row_count` than before. I didn't buy the "SQL Server does this" argument at all. But if that's a postgres term...

I was about to agree to `row_count` - but while writing up the above, it became obvious that just `count` with some kind of `HAVING` notion is much better.

> on_confict comes to mind, it's not a filter too and we haven't had any complaints about that, so maybe not a problem.

Well, we do have other non-filter query string parameters, too - e.g. `select`. But I guess that doesn't matter anymore.. because if we're talking `HAVING`, then those things **are** filters, so it's good when they look the same.

---

I think we have basically two options of implementing the `HAVING`:
- with prefix, e.g. `GET /items?id=gt.5&having.count=eq.5`
- without prefix, but some kind of smartness about "where vs having": `GET /items?id=gt.5&count=eq.5`

I feel we should discuss the details of that in #915. Then we can come back and implement something that looks the same for now in this PR.

wolfgangwalther:
> I feel we should discuss the details of that in #915. Then we can come back and implement something that looks the same for now in this PR.

I'm back from #915. I'm quite satisfied with my suggestion there. If we were to agree on that, the syntax to implement $subject in this PR would look like:

```http
GET /items?id=gt.5&count()=eq.5
```

wolfgangwalther:
One thing we'd need to do to actually support the "let's raise a 404 error if those filters don't apply" case we are looking for:

When the outer query (the one that aggregates via `json_agg`) returns 0 rows, because of the `HAVING` clause - we need to **rollback** the transaction and return 404. Otherwise we'd still be running the mutations behind the scenes.

steve-chavez:
> HAVING count(*) BETWEEN x AND y

Wow, that makes lot of sense :exploding_head:

> One thing we'd need to do to actually support the "let's raise a 404 error if those filters don't apply" case we are looking for:

Right now we do:

```http
GET /projects?id=eq.unexistent
200 OK
[]
```

Since the `count()` is basically a new filter, its behavior would have to be the same. So if we want 404 now, we'd need to change that for all the filters, right? (probably needs a new issue - big breaking change) Or should we treat `count()` as a special case(high level filter) since it's an aggregation? I wouldn't oppose to that.

> When the outer query (the one that aggregates via json_agg) returns 0 rows, because of the HAVING clause - we need to rollback the transaction and return 404

Another option could be to treat the `count()=eq.5` as syntax sugar for mutations and internally just operate in the row count logic we have, that way we can reuse some of the code which already does the rollback. Just a thought, I haven't looked at the actual implementation of HAVING.

steve-chavez:
> I'm back from https://github.com/PostgREST/postgrest/issues/915. I'm quite satisfied with my suggestion there. If we were to agree on that, the syntax to implement $subject in this PR would look like
> GET /items?id=gt.5&count()=eq.5

Agree :handshake:. Just had one question on #915. 

> One thing we need to be aware of: When using this syntax in a query that has aggregates as part of the select=, we basically don't need the json_agg in our main query anymore - because it's replaced with the aggregates provided

That makes sense. Will consider for the implementation.

wolfgangwalther:
> > One thing we need to be aware of: When using this syntax in a query that has aggregates as part of the select=, we basically don't need the json_agg in our main query anymore - because it's replaced with the aggregates provided
> 
> That makes sense. Will consider for the implementation.

Ah, one thing I missed is: What happens when we have a combination of aggregates and non-aggregates in `?select=`? The idea was to turn those non-aggregates into a `GROUP BY` - and that makes a lot of sense.

But that would require us to run two aggregation stages, something roughly like this:

```http
GET /employees?select=gender,salary.avg()&position=eq.department+lead
```

```sql
WITH pgrst_source AS (
  SELECT
    gender,
    avg(salary)
  FROM employees
  WHERE position='department lead'
  GROUP BY gender
)
SELECT json_agg(pgrst_source)
FROM pgrst_source
```

But now, adding a `&count()=gt.10` to the query string is ambiguous. Should it go into a `HAVING` in the inner aggregation or the outer one?

And that would change semantics quite a bit. If it was in the inner aggregation it's very similar to a regular filter - the same thing you noted here:

> Since the `count()` is basically a new filter, its behavior would have to be the same.

But if we put it into the outer aggregation, it's not really the same as a regular filter anymore. The result of a filter in the outer aggregation will always be either: One row or no row. So there is no "filtering down the number of rows". And that's why I would answer this:

> So if we want 404 now, we'd need to change that for all the filters, right?

with no.

> Or should we treat `count()` as a special case(high level filter) since it's an aggregation? I wouldn't oppose to that.

Yes, but only those on the **outer** aggregation as mentioned above. Those on the inner aggregation should behave like regular filters.

---

How to separate the two cases?

I think anything that depends on context (i.e. depending on whether there are non-aggregate columns, `&count()=` could mean inner or outer) is going to be hard to understand and work with. So we need to add something to the query string.

My best idea right now is to use a keyword prefix (the one thing I didn't want to do in the first place...) and I would call it `when`:

```http
GET /employees?select=gender,salary.avg()&position=eq.department+lead&count()=gt.10
GET /employees?select=gender,salary.avg()&position=eq.department+lead&when.count()=gt.10
```

The first one would be a `HAVING` in the inner aggregation. The second one would be on the outer aggregation.

`when` will have the same semantics as a `WHEN` on a trigger: If the condition doesn't apply, the whole query is not run at all.

This could be extended on the server-side, too. Suppose we support a config option `db-when="<function-name>"` (similar to `db-pre-request`). We could call this function in the outer aggregations `HAVING` clause. This function could then do, for example, an `EXPLAIN` to implement a cost threshold. Or it could implement a `db-max-rows`-like solution. Those functions could then throw custom errors - so we can still default to `404 Not Found` for any client-side-when.

This would be a breaking change, because `when` could not be used as an alias for embeddings anymore.

---

> Another option could be to treat the count()=eq.5 as syntax sugar for mutations and internally just operate in the row count logic we have, that way we can reuse some of the code which already does the rollback. Just a thought, I haven't looked at the actual implementation of HAVING. If it's a bit difficult, then this could be a poor's man implementation that leaves room for improvement without causing a breaking change later.

I think the `when` based solution should be easy enough to implement, because it doesn't need to touch aggregation in general at all. I suggest to hardcode the `when.count()` in the parser for now, so no flexibility for other aggregates. And then you can add a simple `HAVING` clause in that case.

But yeah, anything that doesn't cause a breaking change later will work.

steve-chavez:
Been trying the HAVING query mentioned above https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1039489939

The generated query would be like this:

```sql
WITH pgrst_source AS ( SELECT "test"."projects".*FROM "test"."projects")
SELECT
  coalesce(json_agg(_postgrest_t), '[]')::character varying AS body
FROM ( SELECT * FROM pgrst_source ) _postgrest_t
HAVING count(_postgrest_t) = 5;

-- or

WITH pgrst_source AS ( SELECT "test"."projects".*FROM "test"."projects")
SELECT
  coalesce(json_agg(_postgrest_t), '[]')::character varying AS body
FROM ( SELECT * FROM pgrst_source ) _postgrest_t
HAVING count(*) = 5;
```

So far seems to be working great!

steve-chavez:
> GET /employees?select=gender,salary.avg()&position=eq.department+lead&when.count()=gt.10

Hm, I would prefer not to expose a PostgREST-specific thing but as you mention it seems necessary to differentiate the outer level query from the inner one. It's still a a bit weird using the embedded resource syntax for the 'when'.

How about using my [previously proposed syntax for marking filters](https://github.com/PostgREST/postgrest/issues/915#issuecomment-987585190)(`?select=average,amount&average=avg.amount`) for this. So like:

```http
GET /employees?select=gender,salary.avg()&position=eq.department+lead&count()=gt.10&count()=source.outer
```

With `count()=source.outer` we can specify the `count` is on the outer query or on the inner one(`count()=source.inner`) and it doesn't invade the embedded resources namespace - more verbose but doesn't cause a breaking change.

I could also just implement it like `count()=eq.10` for this PR and then later we can add the `source` filter, assuming the default to be `outer` for later on.. when we add `having` support to the aggregates then the `source` filter can be introduced.

@wolfgangwalther WDYT?

wolfgangwalther:
> ```http
> GET /employees?select=gender,salary.avg()&position=eq.department+lead&count()=gt.10&count()=source.outer
> ```
> With count()=source.outer we can specify the count is on the outer query or on the inner one(count()=source.inner) and it doesn't invade the embedded resources namespace - more verbose but doesn't cause a breaking change.

Hm, I see some problems with that:
- We don't have the concept of multiple query parameters with the same key being joined together for the same filter (here both `count()=` parameters are combined into one filter, a `HAVING COUNT(*) >= 10` on the outer level. I think we realized that we are in fact supporting multiple separate filters joined by `AND` with the same name right now - that means we should support something like `GET ...&count()=gt.10&count()=lt.20`. How would you know in this case, where both filters belong (inner or outer query)? What if you wanted to do something like: `GET ...&count()=gt.10&when.count()=eq.1`, i.e. have a `count()` filter on both the inner and outer queries? How would you express that with the `source.inner` / `source.outer` syntax?
- I don't like `outer` and `inner` at all. First of all, it could be confused with `OUTER JOIN` and `INNER JOIN`. And then, to actually understand what both values mean, the user needs to understand how the query that PostgREST runs is built for those to make any sense. I don't think we should require any knowledge about that. We should find a way to have the user express their **intent**, not how the query should look like. After all, we expose a REST API - and not just a query builder. This is also why I don't really like the `!inner` hint for embeddings - and think that the `exists` filter would be much better. The fact that `exists` matches with SQL `EXISTS` is merely a coincidence, but it carries the right semantics for the http request. `!inner` does not carry any of that for the HTTP request itself - it only has meaning because people know the underlying concept of an `INNER JOIN`.

> I could also just implement it like `count()=eq.10` for this PR and then later we can add the `source` filter, assuming the default to be `outer` for later on.. when we add `having` support to the aggregates then the `source` filter can be introduced.

Hm. I don't like that either. The default (i.e. no `outer` or `when` or whatever we end up using) should be similar in behavior to other filters - and that's reducing the number of rows **within** the resource without returning an error. So the default really needs to be `inner` in your terms. Otherwise it will be confusing when using regular filters vs. inner having filters. Outer having filters are the special thing here.

`when.count()=gt.10` can easily be thought of as "succeed only **when** the number of rows is greater than 10" for people without SQL background. And it shares a similar behavior as a `WHEN` clause in triggers for those folks who know SQL, as in "run this trigger only when..." and "run this mutation request only when...".

steve-chavez:
> What if you wanted to do something like: GET ...&count()=gt.10&when.count()=eq.1, i.e. have a count() filter on both the inner and outer queries? How would you express that with the source.inner / source.outer syntax?

I thought about filter aliasing as we discussed [before](https://github.com/PostgREST/postgrest/issues/1970#issuecomment-985259291), we need that for the improved `or` as well, the source filter would specify these aliases.

> it could be confused with OUTER JOIN and INNER JOIN
> After all, we expose a REST API - and not just a query builder. 

All true, but could be solved with better naming I guess.

> when.count()=gt.10 can easily be thought of as "succeed only when the number of rows is greater than 10" for people without SQL background. 

Yeah, I get that and it makes a lot of sense. I'm just trying to see if there's a way to avoid having a non embedded resource using the embedded resource syntax. Internally we do have the `_postgrest_t` CTE(as noted [above](https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1065925310)), it seems that's the `when` for this case. So I guess it's kinda like an internal embedding that's always included.

steve-chavez:
With the above thought, I think I'm fine implementing the `when` as sort of an special embedding, this could be explained on an "under the hood" doc as well.

steve-chavez:
On second thought, I'm not seeing how this `when.count()` syntax would be any better than the previous [row_count](https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1039489939), the latter seems way more intuitive. I got the `db-when` possibility but that looks a bit like an artificial use case for now, it seems it could be done in other ways too. Also I'm not seeing more useful things to do with HAVING on the outside CTE.

So how about this: we do `row_count` as the special column and internally we translate it to `HAVING count(*)`. WDYT?

wolfgangwalther:
> On second thought, I'm not seeing how this `when.count()` syntax would be any better than the previous [row_count](https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1039489939), the latter seems way more intuitive. I got the `db-when` possibility but that looks a bit like an artificial use case for now, it seems it could be done in other ways too. Also I'm not seeing more useful things to do with HAVING on the outside CTE.
> 
> So how about this: we do `row_count` as the special column and internally we translate it to `HAVING count(*)`. WDYT?

Even if you can't see any more useful things to do with the outer having there might very well be some in the future. `with.count()` will block a whole namespace - and one where I think the probability of breaking somebody's application is less likely than with the column name `row_count`: I just think it's far more likely that somebody already named a column `row_count` in some aggregation view or such compared to somebody renaming their embedding to `when`. And even if they did so, this would not be a breaking change immediately - it would just mean that **before** adding `when.count()`, those users would need to rename their embedding. But existing code would just work fine. This is not the case with `row_count`.

Furthermore, I can very well imagine having other filters on the outer having. Especially custom aggregates - you could do **a lot** of things with those.

steve-chavez:
> Even if you can't see any more useful things to do with the outer having there might very well be some in the future.

Yes, but the main keyword here is "in the future" :D. We don't have those use cases right now, once we have safe aggregations then I could see some popping but that's a different issue.

> with.count() will block a whole namespace 

You meant `when.count()` right :D? That's kinda my point, that syntax is hard to remember. The `row_count` seems more obvious.

> where I think the probability of breaking somebody's application is less likely 

Yeah, I somewhat agree but we just can't be sure unless we have all the data regarding endpoint usage. It's going to be a breaking change anyway, users will need to be careful before upgrading.

Not opposed to `when.count` at all, it's just that I find the `row_count` more focused/suitable for this particular issue.

wolfgangwalther:
> Yes, but the main keyword here is "in the future" :D. We don't have those use cases right now, once we have safe aggregations then I could see some popping but that's a different issue.

I don't understand this. When you reserve `when.count()` now, you can easily extend it to use `when.xyz()` in the future - without any more breaking changes / adding of reserved key words.

This is not the case for `row_count`. row_count is exactly one thing and not extensible at all.

> It's going to be a breaking change anyway, users will need to be careful before upgrading.

Actually.. no. It's not a breaking change with `when.count()`. Because this syntax does not exist, yet. But somebody could very well use `row_count=eq.10` for a regular column name already.

steve-chavez:
> Actually.. no. It's not a breaking change with when.count(). Because this syntax does not exist, yet. 

Hm, yeah, even if someone named their table `when` they could not use `count()` before so no breaking change.

steve-chavez:
Should we worried about calling different aggregates on `when.xyz()` without still implementing the cost treshold? It seems to me that `count(*)` is already safe because it's the same as the page count but not sure about other aggregates.

In that way I guess the `row_count` would be better because it's more restricted.

Also, not sure why should we be this worried on `row_count` causing a breaking change when it's clearly a reserved keyword as mentioned [before](https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-CURRENT-DIAGNOSTICS-VALUES).

steve-chavez:
> Also, not sure why should we be this worried on `row_count` causing a breaking change when it's clearly a reserved keyword as mentioned before.

So if a user decided to call a column row_count it would be the same as doing:

```sql
create table tabel(int int, float8 float8, text text, money money);
CREATE TABLE
```

Which does work but you'd expect for some things to break along the line.

steve-chavez:
Just found the reserved keyword list, `row_count` is clearly included:

https://www.postgresql.org/docs/current/sql-keywords-appendix.html

wolfgangwalther:
> Should we worried about calling different aggregates on `when.xyz()` without still implementing the cost treshold? It seems to me that `count(*)` is already safe because it's the same as the page count but not sure about other aggregates.

My understanding is that we're not going to implement different aggregates right out of box, right? I assumed we'd be hardcoding a `when.count()` param in the parser right now - and then we can extend it later.

In terms of "outer having" aggregates, I don't think we need any cost threshold, because the aggregation is happening anyway with `json_agg`. So that's not much more to do.

When implementing the cost threshold, we might actually be able to implement it as an outer having aggregate itself. Kind of a server-side `db-when-cost=lt.1000` or so.

> Also, not sure why should we be this worried on `row_count` causing a breaking change when it's clearly a reserved keyword as mentioned [before](https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-CURRENT-DIAGNOSTICS-VALUES).

I don't think reserved keywords in SQL have any meaning for us. You can use every reserved keyword as a name, as long as you quote it. And we quote everything. So there is no restriction in that sense.

> Which does work but you'd expect for some things to break along the line.

No, certainly not. The expectation is for sure, that those things work. Because those names are valid names.

steve-chavez:
> You can use every reserved keyword as a name, as long as you quote it. And we quote everything. So there is no restriction in that sense.

Yeah, that's true.

Ok, this is a bit unrelated but we have a really big chunk of users asking/complaining about better typing(through Supabase). I feel it'd be easier to expose one magical field(`row_count`) through OpenAPI for every resource that is an `integer` type(same goes for a `statement_cost` field, instead of a `when.cost`). Any idea of how would the `when.xyz()` OpenAPI output be like?

steve-chavez:
If we consider `when` as a magical embedding then we cannot add typing to it since AFAICT OpenAPI still doesn't support references: https://github.com/OAI/OpenAPI-Specification/issues/1646.

steve-chavez:
> Ok, this is a bit unrelated but we have a really big chunk of users asking/complaining about better typing 

I've always wondered if this should be an integral part of our REST API, because of the [Code-On-Demand](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm#sec_5_1_7) constraint. IIUC, that means that we should provide client libraries(say requesting an `application/javascript` media type at the root endpoint). In that case, we should always think of typing whenever adding new syntax.

wolfgangwalther:
> Ok, this is a bit unrelated but we have a really big chunk of users asking/complaining about better typing(through Supabase).

The one big thing that limits our ability to provide better hints for typing is #1970. As long as we have filters like `id=eq.10`, the type for that in OpenAPI will always be `string`. But once we move to `id.eq=10`, we can return `int` (or whatever the correct type is) here.

> I feel it'd be easier to expose one magical property(`row_count`) through OpenAPI for every resource that is an `integer` type(same goes for a `statement_cost` property, instead of a `when.cost`). Any idea of how would the `when.xyz()` OpenAPI output be like?

I think it would look exactly the same as the output you imagine for `row_count`, except for `s/row_count/when.count()/`.

> If we consider `when` as a magical embedding then we cannot add typing to it since AFAICT OpenAPI still doesn't support references: [OAI/OpenAPI-Specification#1646](https://github.com/OAI/OpenAPI-Specification/issues/1646).

I don't understand what is meant by "magical embedding". There is no embedding for me. `when.` is just a prefix. A namespace, to collect all "outer having filters" and avoid collisions with other filters. The only way this is related to embeddings is, because with this an embedding named `when` could not be used with "inner having filters".

What am I missing?

> I've always wondered if typing should be an integral part of our REST API, because of the [Code-On-Demand](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm#sec_5_1_7) constraint. IIUC, that means that we should provide client libraries(say requesting an `application/javascript` media type at the root endpoint should be possible).

I'm not sure whether we should provide client libraries - but in the whole discussion about moving OpenAPI out of haskell and into SQL, I always had in mind that we could have alternative root endpoint implementations, too. One of the things I think many users would benefit from would be typescript typings for the api (i.e. a `.d.ts` file). Generating them directly from the schema should give much better results than going through an intermediate OpenAPI output.

> In that case, we should always think of typing whenever adding new syntax.

Fully agree, thinking about that makes a ton of sense.

steve-chavez:
> I don't understand what is meant by "magical embedding".
> What am I missing?

Just updated the [query grammar](https://github.com/PostgREST/postgrest-docs/issues/228#issuecomment-346981443)(still missing some things). Maybe not so much of an issue but the `when.` creates ambiguity - the `embed-path` is no longer, I guess it's a prefix now. 

> There is no embedding for me. when. is just a prefix. A namespace, to collect all "outer having filters" and avoid collisions with other filters.

If it's just meant to be a prefix, what would be the downside of just using it as `when_count`? That seems unlikely to be a column name and cause a breaking change.

steve-chavez:
> but in the whole discussion about moving OpenAPI out of haskell and into SQL
>  One of the things I think many users would benefit from would be typescript typings for the api (i.e. a .d.ts file). Generating them directly from the schema should give much better results than going through an intermediate OpenAPI output.

Spot on. I wonder how hard/easy is to do this in Haskell itself as an initial version.

steve-chavez:
Suggesting a different approach on https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1069748529.

wolfgangwalther:
> If it's just meant to be a prefix, what would be the downside of just using it as `when_count`? That seems unlikely to be a column name and cause a breaking change.

This is not a prefix by syntax, but just by convention. It does conflict with column names. It's in the same namespace as all column names, so all extensions here would be breaking changes, too.

> > The basic question is "how many reserved keywords can we remove from the querystring?". If we were to start from scratch, the best answer would be "all" - so that we could use arbitrary names for filtering. But this is for sure better discussed in #2066 - and not here.
> 
> Maybe we're mixing a goal of #2066 with the current syntax. I'm not sure how I would have introduced the `on_conflict` or `columns`(even worse because it's not a reserved keyword) query parameters if we really want to avoid using new reserved query params at all costs. I guess with the #2066 syntax those could be changed to `when.on_conflict` and `when.columns`.

I don't see any connection here. `when.on_conflict` and `when.columns` doesn't make any sense to me. Those are not part of the outer having filters, and they do not impact the query the same way. In that same issue I suggested to move them elsewhere, and I still think that those fit better in headers.

> Still feel like we did great progress here, the HAVING and the aggregations inside the inner query are definitely the right design and I'm planning to implement that next. I feel however that the `when` adds some complexity and we can avoid that with the new approach.

I don't think batching is an alternative solution to the same problem - it solves a different problem. And tbh, I don't really see that complexity with `when.`?

steve-chavez:
> I don't think batching is an alternative solution to the same problem - it solves a different problem.

Hm, it does help because a client can now limit an update to only one row, it doesn't enforce an error but still helps.

> And tbh, I don't really see that complexity with when.?

Yeah, I worded that wrongly, I think it's mostly a bit of inconsistency because of reusing the embedded resource syntax. Will come back to this one later though.

steve-chavez:
Would like to make a new release after https://github.com/PostgREST/postgrest/pull/2195 is done. Not merging this one when we already have the implementation is a bit of a shame though.

I was thinking of a way out. How about if we offer a config like this one:

```ini
renamed-reserved-query-parameters = "row_count:another_count, select:my_select, another_reserved_qp:my_qp"
```

With that, if a user used `row_count` as a column name then he/she can change it and prevent a breaking change. This config will also help with the introduction of any new query parameter.

(A bit like https://github.com/PostgREST/postgrest/issues/1909 but with much lower capacity, just for this use case)

@wolfgangwalther WDYT?

steve-chavez:
> renamed-reserved-query-parameters = "row_count:another_count, select:my_select, another_reserved_qp:my_qp"

Not like we're going to introduce any name for new query parameters with this change though.

I just that I think that `row_count` is the better choice here. It's more to the point than `when.count()` and it seems safe enough for us to use(I'm considering APIs with columns named `row_count` edge cases, I'd use a more friendly name for a view col aggregation).

steve-chavez:
Guess this can go into the next release so no hurry.

wolfgangwalther:
> How about if we offer a config like this one:
> 
> ```ini
> renamed-reserved-query-parameters = "row_count:another_count, select:my_select, another_reserved_qp:my_qp"
> ```
> [...]
> @wolfgangwalther WDYT?

A new config option?

-1.

-100.

> I was thinking of a way out.

Summarizing:
- We agree that the query string is the right place to put this piece of information due to resource considerations.
- We disagree about how exactly - both our concerns are about naming collisions.

Maybe we should pick up this thought again:

> Using different parts of the URI (path parameters) for things in the query string that are "not a filter".

What we are describing here is not a regular filter, because it behaves differently. Very similar btw to the "path" component of the URI: When you do `GET /not_existing?...` you'll get a 404 error, too!

I'm not saying my original idea was best, I'm open for other suggestions. But looking at "something in the path component" makes a lot of sense...

> Not merging this one when we already have the implementation settled is a bit of a shame though.

We should not merge it, when we still disagree.

steve-chavez:
> We should not merge it, when we still disagree.

Yeah definitely not, the implementation is not even complete here. Just want to merge https://github.com/PostgREST/postgrest/pull/2195 and then make a release.

wolfgangwalther:
> > We should not merge it, when we still disagree.
> 
> Yeah definitely not, the implementation is not even complete here. Just want to merge #2195 and then make a release.

Ah, "this" as #2195. Got it.

steve-chavez:
> What we are describing here is not a regular filter, because it behaves differently. Very similar btw to the "path" component of the URI: When you do GET /not_existing?... you'll get a 404 error, too!

That's true. Let's explore this path for now. Assume we use [this format mentioned in the OpenAPI docs](https://swagger.io/docs/specification/describing-parameters/#path-parameters):

> label expansion – dot-prefixed, such as /color.R=100.G=200.B=150

So for the `row_count` we'd use:

```http
GET /projects.5
DELETE /projects.5?id=gt.100
````

Still a bit weird for me but it doesn't look that bad. It seems it could be extended to embedded resources as well.

```http
GET /projects.5?select=clients.3(*)
```

And to extend that to other special filters we'd do:

```http
GET /projects.row_count=5.statement_cost=500?id=gt.5&is_active=is.true
```

That does start to look confusing, path parameters look the same as query parameters. For me, we might as well just use query parameters then.

Unless the `row_count` would be the only path parameter we had..

steve-chavez:
Also, another thing we lose with `row_count` as a path parameter is that we can no longer use our filters on it. Like:

```http
DELETE /users?status=eq.inactive&row_count=gt.20
DELETE /users?status=eq.inactive&row_count=eq.20
DELETE /users?status=eq.inactive&row_count=lt.20
```

So with that we can say delete "at most"/"at least"/exactly this amount of rows.

With path parameters I guess we could do:

```http
GET /projects.row_count=gt.20?status=eq.inactive
```

But again that form is just the same we already have with our query parameters.

steve-chavez:
Ok, so let's step back to the `when.count()` option. How about this, since we already have a namespace there, we do:

```http
GET /projects?when.row_count=lt.5
```

It seems more clear than doing `count()` for me. WDYT?

Then we could have the `when.cost` as well.

wolfgangwalther:
> Ok, so let's step back to the `when.count()` option. How about this, since we already have a namespace there, we do:
> 
> ```httpspec
> GET /projects?when.row_count=lt.5
> ```
> 
> It seems more clear than doing `count()` for me. WDYT?

Namespace-wise, this is better than using a plain `row_count`.

However I don't see why we should use a renaming construct here? Introducing an arbitrary mapping makes it really hard to possibly extend this to using arbitrary aggregation functions later on. The function used is called `count` - we should use that name.

---

> With path parameters I guess we could do:
> 
> ```http
> GET /projects.row_count=gt.20?status=eq.inactive
> ```
> 
> But again that form is just the same we already have with our query parameters.

That would be very nice actually. It looks the same as a query parameter, because it's a filter. But there is a difference: Query parameters are **optional**, in the sense that not matching them will still return a response and not an error. Path parameters are **required**. This maps very nicely to the fact that query parameters are put **after the `?`**, which makes them optional. Very similar to what `?` means in a regex :). Taking this one step further it would be great if we could mark filters as **required by using `!`**.

Something like:
```http
GET /projects!count=gt.20?status=eq.inactive
```

[`!` is allowed in that part of the URI](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3), because it's a `sub-delim`, just like `;` and `,`.

This also matches up very nicely with embedding hints: When you specify a `!hint` for an embedding, it is a ''requirement** that needs to be matched - otherwise the query will fail.

---

> Still a bit weird for me but it doesn't look that bad. It seems it could be extended to embedded resources as well.
> 
> GET /projects.5?select=clients.3(*)

In a sense a "required filter" for an embedding would turn the embedding into an `INNER JOIN`. Throwing an error is not appropriate in that case. But from a query perspective the regular optional filters would be inside the subquery, while the required filters would be in the `WHERE` part of the outer query. However that kind of problem would already be better solved with https://github.com/PostgREST/postgrest/issues/1414#issuecomment-1028220016, so I don't think we need to implement those required filters on embeddings.

wolfgangwalther:
> > Ok, so let's step back to the `when.count()` option. How about this, since we already have a namespace there, we do:
> > ```
> > GET /projects?when.row_count=lt.5
> > ```
> > It seems more clear than doing `count()` for me. WDYT?
> 
> Namespace-wise, this is better than using a plain `row_count`.
> 
> However I don't see why we should use a renaming construct here? Introducing an arbitrary mapping makes it really hard to possibly extend this to using arbitrary aggregation functions later on. The function used is called `count` - we should use that name.

One more thing to note here: Using `when.count()` is not a breaking change **only** because of using `()`. Without those, it would break current embeddings which are called `when` and have a column called `count` or `row_count`.

If we were ok to introduce a breaking change, we could do the following:
- Block the entire `when.` namespace for embedding now (breaking change)
- Then use `row_count=eq.whatever`, `cost=lt.whatever` etc for postgrest built-in calls
- And use `count()=eq.whatever`, `other_aggregate()=lt.whatever` for custom aggregates later on

In this case using a mapping of `row_count` -> `count()` would not prevent us from implementing custom aggregates later on. And it would also solve the "`cost` is not implemented as an aggregate anyway" problem.

If we do it like this, I'd be fine with using `when.row_count=op.val` here. If custom aggregates were implemeted, you could achieve the same with `when.count()=op.val`, but there would be no collision.

However, tbh: I like the `!count()=op.val` path parameter approach more. We could take the same approach there:
- `!row_count=op.val` and later `!cost=op.val` (without `()`) for pgrst built-ins
- have the freedom to implement `!any_agg()=op.val` later on

steve-chavez:
> However, tbh: I like the !count()=op.val path parameter approach more. We could take the same approach there:

The path parameters have the advantage of not colliding with another namespace as well, unlike the `when`.

However I still feel this choice will give us trouble in the future. Like when we chose making the `+` symbol significant(it was used instead of `!` at one time) at then we found out some proxies automatically changed it to a whitespace. I think the same thing would happen with a GET/DELETE with a body as we discussed on other issue.

For example, on curl you can use the `-G` and `-d` options to pass query parameters more cleanly:

```bash
curl -X DELETE "localhost:3000/limited_update_items" -G \
   -d order=id \ 
   -d limit=1
```

There's no option for doing this with the new path parameters. Likely other tools/libraries also have facilities for dealing with regular query parameters as well.

---

I have another proposal that would solve the [Prefer problem](https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1034954143)(older PostgREST versions will err on it and not just ignore it), the namespacing ones and wouldn't require a new syntax.

Basically use another vendored [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types):

```http
Accept: application/vnd.pgrst.mutation+json;rowcount=lte.50
```

The media type parameter already has our namespace, so it doesn't conflict with existing embeds or columns. Also older PostgREST versions will just fail if this media type is sent.

The existing `vnd.pgrst.object` would be equivalent to:

```http
Accept: application/vnd.pgrst.mutation+json;rowcount=eq.1
```

This namespace would also allow us to set a `cost` in the future.

So far I think this is the most consistent way of adding this feature. @wolfgangwalther WDYT?

wolfgangwalther:
> However I still feel this choice will give us trouble in the future. Like when we chose making the `+` symbol significant(it was used instead of `!` at one time) at then we found out some proxies automatically changed it to a whitespace.

That's because `+` is part of `application/x-www-form-urlencoded`: https://www.w3.org/TR/html4/interact/forms.html#h-17.13.4.1

I don't think `!` is treated in a special way anywhere. And if it was, we would have found out already, because we're using it in hints.

> I think the same thing would happen with a GET/DELETE with a body as we discussed on other issue.

Could you elaborate on that? I don't see a problem, yet, with the path parameter syntax and what we've discussed around this matter.


> For example, on curl you can use the `-G` and `-d` options to pass query parameters more cleanly:
> 
> ```shell
> curl -X DELETE "localhost:3000/limited_update_items" -G \
>    -d order=id \ 
>    -d limit=1
> ```
> 
> There's no option for doing this with the new path parameters. Likely other tools/libraries also have facilities for dealing with regular query parameters as well.

Hm. However, tools/library support for mimetype parameters in the accept header is even worse, right?


> ```httpspec
> Accept: application/vnd.pgrst.mutation+json;rowcount=lte.50
> ```
> 
> The media type parameter already has our namespace, so it doesn't conflict with existing embeds or columns. Also older PostgREST versions will just fail if this media type is sent.
> 
> The existing `vnd.pgrst.object` would be equivalent to:
> 
> ```httpspec
> Accept: application/vnd.pgrst.mutation+json;rowcount=eq.1
> ```
> [...]
> So far I think this is the most consistent way of adding this feature.

Hm. It is consistent with what was previously done with `vnd.pgrst.object`. However, I remember you saying elsewhere, that this wasn't the best choice, because some tools don't even allow to set a mimetype at all. A path, however, can be set everywhere. Might not have the best tooling support, but certainly better than mimetypes.

---

> This namespace would also allow us to set a `cost` in the future.

Uh, mh. I mean `cost` is already out of place in the path. But in the mimetype? It's even more out of place. The existing `vnd.pgrst.object` is first of all about the shape of the json response: object vs array. It's only a secondary effect, that responses with fewer or more rows throws errors, because they can't be represented as an object.

But what does `vnd.pgrst.mutation` mean? How does this describe the shape of the response body? It does not at all.

Even worse: When you do prefer not to return a representation, which is especially likely on a `DELETE` request - how would the accept header even make sense? We can freely ignore the accept header if we choose not to return a body at all.

In terms of HTTP semantics, the path parameter is really the right place to put it.

wolfgangwalther:
> > Uh, mh. I mean cost is already out of place in the path. But in the mimetype? It's even more out of place
> 
> Yeah, noticed that as well for the path. Then I think the cost only makes sense as a header, seems it'd fit as a Prefer.

Hm. Assume we have no cost limit by default and want to enforce one from the client-side: In this case, we'll just have the same problem with the `Prefer` header: It can by ignored, if not understood. So it's not a safe-guard at all.

But if we think about cost the other way around: Let's say we have a rather low server-side default limit for cost ("soft-limit"). And then we send a `Prefer` header to set the cost limit **higher** (up to a max - "hard limit"). In that case a `Prefer` header would be a great fit. If the header was ignored - fine, we'd still work with the lower limit. That's safe to do.


> > Even worse: When you do prefer not to return a representation, which is especially likely on a DELETE request - how would the accept header even make sense? We can freely ignore the accept header if we choose not to return a body at all.
> 
> On [#1417 (comment)](https://github.com/PostgREST/postgrest/pull/1417#issuecomment-560497408), we concluded that `Accept` can be enforced whether the response returns a body or not.
> 
> > But what does vnd.pgrst.mutation mean? How does this describe the shape of the response body? It does not at all.
> 
> How about a different name then:
> 
> ```httpspec
> Accept: application/vnd.pgrst.array+json;size=lte.50
> ```

Yeah, that would be very consistent with how we use `vnd.pgrst.object` now. But as mentioned in https://github.com/PostgREST/postgrest/issues/2066#issue-1070332545, I don't think this was the best implementation in the first place.

> Then it does describe the shape of the body. I was thinking of `mutation` as a generic term to do `csv` as well, but a name for that can be thought later(maybe `rows` instead of `array`).

Wait, what? A generic mimetype that would return either json or csv? How, if not by mimetype, would I then even tell PostgREST to return json or csv? A mimetype should describe the output format.

The csv argument is a very good one: We should not do this via mimetype. Otherwise, we'd need a special mimetype for each kind of response. And suddenly it will be very hard to support this for **custom** mimetypes. Why should this feature not be available, once we support something like `Accept: text/yaml` with custom functions?

steve-chavez:
> Wait, what? A generic mimetype that would return either json or csv? How, if not by mimetype, would I then even tell PostgREST to return json or csv? 

Isn't that possible by having suffixes in the mime type like the `+json` suffix we've been using in `vnd.pgrst.object+json`?

https://datatracker.ietf.org/doc/html/draft-ietf-appsawg-media-type-suffix-regs-02#section-3.1

>The csv argument is a very good one: We should not do this via mimetype. Otherwise, we'd need a special mimetype for each kind of response

So with the above we could do json and csv with: 

```http
Accept: application/vnd.pgrst.rows+json;rowcount=lte.50

Accept: application/vnd.pgrst.rows+csv;rowcount=lte.50
```

steve-chavez:
The [iana registered media types](https://www.iana.org/assignments/media-types/media-types.xhtml) seem pretty liberal. For example, the subtype of this [application/conference-info+xml](https://www.iana.org/assignments/media-types/application/conference-info+xml) doesn't define a structure on his own but the suffix does. It would be the same case with our `vnd.pgrst.rows` subtype.

steve-chavez:
> Accept: application/vnd.pgrst.array+json;

Though I think we need the above as well.

We discussed this somewhere else but the `json_typeof` we run on mutations is expensive on the db side. Having both `application/vnd.pgrst.array+json` and `application/vnd.pgrst.object+json` would let us run `json_populate_recordset` and `json_populate_record` precisely. These options could then be abstracted on a client library to gain performance.

wolfgangwalther:
> So with the above we could do json and csv with:
> 
> ```httpspec
> Accept: application/vnd.pgrst.rows+json;rowcount=lte.50
> 
> Accept: application/vnd.pgrst.rows+csv;rowcount=lte.50
> ```

But how would you do that with custom mimetypes that we don't handle out of the box but that could be supported by #1582?

This will limit the ability to use custom mimetypes heavily.

---

> > Accept: application/vnd.pgrst.array+json;
> 
> Though I think we need the above as well.
> 
> We discussed this somewhere else but the `json_typeof` we run on mutations is expensive on the db side. Having both `application/vnd.pgrst.array+json` and `application/vnd.pgrst.object+json` would let us run `json_populate_recordset` and `json_populate_record` precisely. These options could then be abstracted on a client library to gain performance.

This is not a requirement for `Accept`, though - but for `Content-Type`.

steve-chavez:
Hm, considering all the alternatives here, I think we shouldn't let the [the perfect be enemy of the good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good). 

The `Prefer` approach doesn't conflict with the columns namespace, doesn't require new syntax in the URL and has the [problem of not being strict](https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1034954143) in its application: a problem in the case someone uses an outdated postgREST, for which there are other workarounds(explicit docs, warnings, etc). In light of the unavailability of the `Expect` header, `Prefer`  is the best HTTP affordance we can adopt.

```http
PATCH /tbl

Prefer: rowcount=eq.0, rowcount=lt.50
```

(We need to make sure the value syntax accepts two row counts, otherwise we'd need to collapse into a single val)

---

Related https://github.com/PostgREST/postgrest/issues/2343#issuecomment-1165019537

steve-chavez:
Taking the dollar namespace idea on https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1376533758

> when.count()=op.val

How about if we do it like this:

```http
GET /tbl?$rowcount=eq.3
```

Since pg identifiers cannot start with a `$` this should be safe - well maybe not completely safe because we quote the columns.

In this case I wouldn't be opposed to a namespace like `when` since we could have other dollar namespaces and is clear that this is a postgREST-specific thing.

```http
GET /tbl?$when.rowcount=eq.3
```

`$when` is not that clear though, but I don't have a better idea for now.

@wolfgangwalther WDYT? No more debate on the `Prefer` header :smile_cat:

steve-chavez:
> GET /tbl?$when.rowcount=eq.3

The above still seems a bit unwieldy to me. Really the lack of the `Expect` header is what really originates this whole discussion. 

If Kong can introduce such non-standard headers as `apikey`([ref](https://docs.konghq.com/hub/kong-inc/key-auth/#make-a-request-with-the-key)), why cannot we introduce our own header?

It could be called `X-Expect` or `PG-Expect`(Cloudflare has its `CF-` headers).

Personally I'd vote for `X-Expect`. We have needed this capability for quite some time.

steve-chavez:
I think https://github.com/PostgREST/postgrest/issues/2887 is now a better fit of all the alternatives here. 

It will require less parsing too (which will help for https://github.com/PostgREST/postgrest/issues/2816)

PR #2168, opened 2022-02-15
fix: Using GET with certain Content-Type headers now correctly calls the no parameter function if it exists

### LINKED ISSUE #2147: GET requests to functions with no parameters and "text/plain" or "application/octet-stream" as "Content-Type" header return "Not Found" error
Bug found in [this conversation in Gitter](https://gitter.im/begriffs/postgrest?at=61fb1330b8b082448215afa0)

It returns the error even if the function exists in the database. It should ignore the header and access the function without problem.

Looks like this line is the culprit: https://github.com/PostgREST/postgrest/blob/58f76f3d6d28c676af68c5be59efa1fde4b29ac6/src/PostgREST/Request/ApiRequest.hs#L468

How to reproduce:

```sql
create function api.noparam() returns integer
  immutable
  language sql
as
$$
select 1;
$$;

select api.noparam();
```

```console
$ # Returns 1 as expected
$ curl "localhost:3000/rpc/noparam"
1

$ # Returns 404 Not Found instead of returning 1
$ curl "localhost:3000/rpc/noparam" -H "Content-Type: text/plain"
{
  "hint":"If a new function was created in the database with this name and parameters, try reloading the schema cache.",
  "message":"Could not find the api.noparam() function in the schema cache"
}
```

laurenceisla:
Closes #2147.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
  - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
  - #2042, Keep working when EMFILE(Too many open files) is reached. - @steve-chavez
+ - #2147, Using GET with `text/plain` or `application/octet-stream` as `Content-Type` headers no longer returns "Not Found" when the requested function with no parameters exists
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
  - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
  - #2042, Keep working when EMFILE(Too many open files) is reached. - @steve-chavez
- - #2147, Using GET with `text/plain` or `application/octet-stream` as `Content-Type` headers no longer returns "Not Found" when the requested function with no parameters exists
+ - #2147, Ignore `Content-Type` headers for `GET` requests when calling RPCs. Previously, `GET` without parameters, but with `Content-Type: text/plain` or `Content-Type: application/octet-stream` would fail with `404 Not Found`, even if a function without arguments was available.
+ - ```
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1244,6 +1244,20 @@ spec actualPgVersion =
             let respBody = simpleBody r
             respBody `shouldBe` file
 
+        it "should call the function with no parameters and not fallback to the single unnamed parameter function when using GET with Content-Type headers" $ do
+          request methodGet "/rpc/overloaded_unnamed_param" [("Content-Type", "application/json")] ""
+            `shouldRespondWith`
+              [json| 1|]
+              { matchStatus  = 200 }
```
steve-chavez, line 1251: Hm, did `Content-Type: application/json` also failed?

https://github.com/PostgREST/postgrest/issues/2147 only mentions `text/plain` and `octet-stream`.

steve-chavez, line 1251: Hm, `Content-Type: application/json` also failed?

https://github.com/PostgREST/postgrest/issues/2147 only mentions `text/plain` and `octet-stream`.

laurenceisla: True that. I removed it from the test.

steve-chavez (approved):

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2145, Fix accessing json array fields with -> and ->> in ?select= and ?order=. - @wolfgangwalther
  - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
  - #2042, Keep working when EMFILE(Too many open files) is reached. - @steve-chavez
+ - #2147, Using GET with `text/plain` or `application/octet-stream` as `Content-Type` headers no longer returns "Not Found" when the requested function with no parameters exists
```
wolfgangwalther, line 39: I had to read it a few times to understand.

What about the following?

```suggestion
 - #2147, Ignore `Content-Type` headers for `GET` requests when calling RPCs. Previously, `GET` without parameters, but with `Content-Type: text/plain` or `Content-Type: application/octet-stream` would fail with `404 Not Found`, even if a function without arguments was available.
 - ```

wolfgangwalther: > somehow make the Content-Type value part of the [Payload](https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Request/ApiRequest.hs#L71) type so it doesn't leak on GET logic?

Yeah, I think that's the right direction. We need stronger typing for our whole `ApiRequest`, eliminating many impossible cases.

Issue #2174, opened 2022-02-19
Health check returns 503 even if PostgREST is running correctly unless the `server-host` config is set to `localhost`


laurenceisla:
### Environment

* PostgreSQL version: 13.5
* PostgREST version: 9.0.0.20220211
* Operating system: EndeavourOS - Linux 5.16.10-arch1-1

### Description of issue

After setting the configuration variable `admin-server-port` to `3001` I tested the health check by doing a `curl -I "http://localhost:3001/live` and the response I got was `HTTP/1.1 503 Service Unavailable` (it was the same for the `/ready` endpoint), even though  the PostgREST service at the `3000` port was working correctly.

I did a debug trace to this line of code:

https://github.com/PostgREST/postgrest/blob/dcf7ade5bc70a31fe0e083e1527b74125a94aeda/src/PostgREST/Admin.hs#L24

and the PostgREST service returns this as a result of the health check:

```haskell
Left Network.Socket.getAddrInfo 
  (called with preferred socket type/protocol: 
    AddrInfo {addrFlags = [], 
              addrFamily = AF_UNSPEC, 
              addrSocketType = Stream, 
              addrProtocol = 0, 
              addrAddress = <assumed to be undefined>, 
              addrCanonName = <assumed to be undefined>}, 
              host name: Just "!4", 
              service name: Just "3000"): 
  does not exist (Name or service not known)
```

Changing the `server-host` to `*` gave a different error:

```haskell
Left Network.Socket.connect: <socket: 16>: does not exist (Connection refused)
```

What worked was to set the `server-host` to `localhost` or `127.0.0.1`, only then the health check returned `200 OK` as expected.

Perhaps this is an issue with my distro, but still, maybe some logs are required to inform admins of the possible error?

wolfgangwalther:
> Perhaps this is an issue with my distro

Don't think so, I can reproduce easily.

The problem is in the special values for `server-host`. Those are handled by [`bindPortTCP`](https://hackage.haskell.org/package/streaming-commons-0.2.2.4/docs/Data-Streaming-Network.html#g:8) - but obviously not by the code that tries to connect to the main socket.

I think the proper solution would be to:
- Replace [`runSettings` with `runSettingsSocket`](https://hackage.haskell.org/package/warp-3.3.19/docs/Network-Wai-Handler-Warp.html#g:1) - and call `bindPortTCP` ourselves.
[...]

PR #2182, opened 2022-03-04
Correct live/ready checks to consider special host values

steve-chavez:
Fixes https://github.com/PostgREST/postgrest/issues/2174.

Took a different route from the one suggested in https://github.com/PostgREST/postgrest/issues/2174#issuecomment-1045978929. I avoided including the [streaming-commons](https://hackage.haskell.org/package/streaming-commons) dependency and instead just handled the special host values in the admin server logic by stealing some of the code from [bindPortGenEx](https://hackage.haskell.org/package/streaming-commons-0.2.2.4/docs/src/Data.Streaming.Network.html#bindPortGenEx). That seemed like a simpler change.

### TODO

- [x] Refactor
- [x] Make tests green

### DIFF
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -140,7 +140,7 @@ def dumpconfig(configpath=None, env=None, stdin=None):
 
 
 @contextlib.contextmanager
-def run(configpath=None, stdin=None, env=None, port=None):
+def run(configpath=None, stdin=None, env=None, port=None, host=None):
     "Run PostgREST and yield an endpoint that is ready for connections."
     env = env or {}
     env["PGRST_DB_POOL"] = "1"
@@ -149,7 +149,7 @@ def run(configpath=None, stdin=None, env=None, port=None):
     with tempfile.TemporaryDirectory() as tmpdir:
         if port:
             env["PGRST_SERVER_PORT"] = str(port)
-            env["PGRST_SERVER_HOST"] = "localhost"
+            env["PGRST_SERVER_HOST"] = host or "localhost"
             baseurl = f"http://localhost:{port}"
         else:
             socketfile = pathlib.Path(tmpdir) / "postgrest.sock"
@@ -883,6 +883,17 @@ def test_admin_live_dependent_on_main_app(defaultenv):
         response = postgrest.admin.get("/live")
         assert response.status_code == 503
 
+@pytest.mark.parametrize("specialhostvalue", FIXTURES["specialhostvalues"])
+def test_admin_works_with_host_special_values(specialhostvalue, defaultenv):
+    "Should get a success from the admin live and ready endpoints when using special host values for the main app"
+
+    with run(env=defaultenv, port=freeport(), host=specialhostvalue) as postgrest:
+
+      response = postgrest.admin.get("/live")
+      assert response.status_code == 200
+
+      response = postgrest.admin.get("/ready")
+      assert response.status_code == 200
 
 @pytest.mark.parametrize(
     "level, has_output",
--- a/src/PostgREST/Admin.hs
+++ b/src/PostgREST/Admin.hs
@@ -21,7 +21,7 @@ import Protolude
 -- | PostgREST admin application
 postgrestAdmin :: AppState.AppState -> AppConfig -> Wai.Application
 postgrestAdmin appState appConfig req respond  = do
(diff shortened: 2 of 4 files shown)


steve-chavez:
localhost is somehow not resolving.. not sure why.

wolfgangwalther:
Hm. I'm not 100% convinced by this. When I run multiple instances of PostgREST on the same port, but different interfaces, one instance might go down, but the healthcheck could still show 200 OK, because it connects to the other instance instead.

So this has two problems:
- It tries multiple addresses, which might lead to the above.
- It determines those addresses at request-time, i.e. when the request to the live endpoint is made. At this time there might be other interfaces with other IPs.

A more reliable approach would be to run this once, when the main app port is assigned. And then only take the first available interface. But that will quickly lead to the same solution as mentioned in https://github.com/PostgREST/postgrest/issues/2174#issuecomment-1045978929.

test/io/test_io.py
```diff
@@ -883,6 +884,17 @@ def test_admin_live_dependent_on_main_app(defaultenv):
         response = postgrest.admin.get("/live")
         assert response.status_code == 503
 
+def test_admin_works_with_host_special_values(defaultenv):
+    "Should get a success from the admin live and ready endpoints when using special host values for the main app"
+
+    for host in ["*4", "!4", "*6", "!6", "*"]:
```
wolfgangwalther, line 890: This is not how multiple almost identical test-cases are created in pytest. Have a look at some of the other test cases in this file that use `@pytest.mark.parametrize` to achieve it in a better way.

test/io/test_io.py
```diff
@@ -140,16 +140,17 @@ def dumpconfig(configpath=None, env=None, stdin=None):
...
     env = env or {}
     env["PGRST_DB_POOL"] = "1"
     env["PGRST_DB_POOL_TIMEOUT"] = "1"
+    host = host or "localhost"
 
     with tempfile.TemporaryDirectory() as tmpdir:
         if port:
             env["PGRST_SERVER_PORT"] = str(port)
-            env["PGRST_SERVER_HOST"] = "localhost"
+            env["PGRST_SERVER_HOST"] = host
```
wolfgangwalther, line 153: Right you change the default host for all tests that don't specify `host`.

To not accidentally change unrelated things, it'd be better to do:

```suggestion
            if host:
                env["PGRST_SERVER_HOST"] = host
```

And then you don't need the `host = host or ...` line above.

wolfgangwalther:
@laurenceisla There's a lot of unrelated changes that are already on `main` in here. I think you need to rebase properly to get rid of them; I think you somehow merged those in without rebase.

laurenceisla:
> @laurenceisla There's a lot of unrelated changes that are already on `main` in here. I think you need to rebase properly to get rid of them; I think you somehow merged those in without rebase.

Fixed, I reverted it to the state before I commited. I will push the tests again after I fix them with the revisions you gave.

steve-chavez:
> When I run multiple instances of PostgREST on the same port, but different interfaces, one instance might go down, but the healthcheck could still show 200 OK, because it connects to the other instance instead

Hm, this seems like an edge case but yeah it seems it can happen. In the case you have a machine with multiple network interfaces you'd be better using an specific host name for each postgrest service rather than the special values(!4, *, etc) though right? And in that case the edge case won't happen?

wolfgangwalther:
> Hm, this seems like an edge case but yeah it seems it can happen. In the case you have a machine with multiple network interfaces you'd be better using an specific host name for each postgrest service rather than the special values(!4, *, etc) though right? And in that case the edge case won't happen?

True. Should be enough to point that out in the docs. Using multiple interfaces is only safe when setting specific values. And I think that's true for the main app as well, not just for the admin app.

---

This is one of the MRs where we would really benefit from the ability to build docker images for MRs, including upload to docker hub. I would really like to test this in my k8s setup, but don't know how.

steve-chavez:
> True. Should be enough to point that out in the docs.

Cool. Then this is looking good to merge.

> This is one of the PRs where we would really benefit from the ability to build docker images for PRs, including upload to docker hub.

That would be nice, like the previews in readthedocs.. users could try new features too.. I guess the idea would be tagging an image for each CI run and the tag would be the PR's number..

Issue #2155, opened 2022-02-07
`max-rows` affects bulk insert


steve-chavez:
If `max-rows=1000`, [bulk inserting](https://postgrest.org/en/stable/api.html#bulk-insert) 1500 rows with `Prefer: return=representation` will only return 1000 rows, which is misleading.

I think the correct behavior should be ignoring `max-rows` for POST, maybe for PATCH/DELETE as well.

**Edit 1**: Should POST, PATCH and DELETE have their own `max-rows` as a way to limit inserted, updated and deleted rows?

**Edit 2**: We could have a `max-rows-read` and `max-rows-mutated`.

steve-chavez:
Yes. Let's use this issue to disallow `?limit/offset` on POST/PATCH/DELETe as we agreed on https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1034004013.

aranibatta:
This a trip

Ron Bhattacharyay
EECS UC Berkeley Regent's and Chancellor's Scholar '18
***@***.*** | 3104916828

wolfgangwalther:
> Yes. Let's use this issue to disallow `?limit/offset` on POST/PATCH/DELETe as we agreed on [#2156 (comment)](https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1034004013).

Does that still include the following?

> I think the correct behavior should be ignoring `max-rows` for POST, maybe for PATCH/DELETE as well.

So the suggestion would be to apply both `db-max-rows` (server-side limit) and `?limit`+`Range` headers (client-side limit) **only** to `GET` requests, correct?

steve-chavez:
> So the suggestion would be to apply both db-max-rows (server-side limit) and ?limit+Range headers (client-side limit) only to GET requests, correct?
> Does that mean that db-max-rows is essentially an upper limit on ?limit? So kind of a db-max-limit in that sense?

Yes, yes and yes. Though `db-max-limit` doesn't really convey that it is the maximum number of rows returned.

steve-chavez:
Will just not apply `max-rows` to POST/PATCH/DELETE to solve this. Users haven't asked for a `max-mutations` config so that's still not needed.

Issue #1959, opened 2021-09-21
Support proper bulk update via PATCH


yourfavouriteuncle:
### Environment

* PostgreSQL version: 11
* PostgREST version: 8.0.0
* Operating system: Linux

### Description of issue

Schema:

```sql
drop table if exists example;
create table if not exists example (
	pk integer PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
	name   varchar(40) NOT NULL CHECK (name <> ''),
	value	varchar(40) NOT NULL CHECK (value <> '')
);

insert into example (name, value) values ('one', '1');
insert into example (name, value) values ('two', '2');
insert into example (name, value) values ('three', '3');
```

Postgrest usage:

```shell
#!/bin/bash

set -e

data='
    [
        {
            "pk": 1,
            "name": "example"
        },
        {
            "pk": 2,
            "name": "second"
        }
    ]
'

# returns all three
curl \
    -X GET "http://localhost:9000/example"

echo -e ""
echo -e "--"

# null value in column \"value\" violates not-null constraint
curl \
    -X POST "http://localhost:9000/example" \
    -H "Prefer: resolution=merge-duplicates" \
    -d "${data}"

echo -e ""
echo -e "--"

# null value in column \"value\" violates not-null constraint
curl \
    -X POST "http://localhost:9000/example?columns=name" \
    -H "Prefer: resolution=merge-duplicates" \
    -d "${data}"
```

Bulk update on partial set of columns triggers the `NOT NULL` constraint of the rest of the columns which are set as `NOT NULL`. Currently we relaxed the constraints in our database to fix that, and looked at doing this via a stored procedure, however it would be nice to have an official suggestion / fix for this.

wolfgangwalther:
The reason is that you are not doing a bulk **update** at all. You are doing an `INSERT`. I guess you wanted to use `PUT` instead of `POST` to make it a `INSERT ... ON CONFLICT DO UPDATE ...`, but that would still yield the same error. The error happens at the `INSERT` stage, before PostgreSQL ever reaches the `UPDATE` part.

`PUT` is not a proper "bulk update".

Related: #1945.

yourfavouriteuncle:
Thanks for the clarification, makes perfect sense. Should I close this issue and move the discussion over at the Q&A?

wolfgangwalther:
> Is a PATCH the HTTP verb that should be used for this?

[...]

PR #2195, opened 2022-03-21
Add limited UPDATE/DELETE

steve-chavez:
As discussed in https://github.com/PostgREST/postgrest/issues/2156.

Applies this [technique](https://blog.crunchydata.com/blog/simulating-update-or-delete-with-limit-in-postgres-ctes-to-the-rescue) to limit rows affected by UPDATE/DELETEs.

Sample generated query:

```http
PATCH /projects?limit=1
```

```sql
WITH
  pgrst_payload AS ( SELECT $1::json AS json_data),
  pgrst_update_body AS (SELECT * FROM json_populate_recordset (null::"test"."projects" , (SELECT val FROM pgrst_body) ) LIMIT 1),
  pgrst_affected_rows AS (SELECT "test"."projects"."id" FROM "test"."projects" ORDER BY "test"."projects"."id" LIMIT $2 OFFSET $3)
UPDATE "test"."projects"
SET "name" = (SELECT "name" FROM pgrst_update_body)
FROM pgrst_affected_rows
WHERE "test"."projects"."id" = "pgrst_affected_rows"."id"
RETURNING 1;
```
### TODO

- [x] Limited UPDATE
- [ ] Limited DELETE
- [ ] Fix https://github.com/PostgREST/postgrest/issues/2155


test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -386,3 +386,71 @@ spec = do
...
+    it "works with the limit query param plus a filter" $
+      request methodPatch "/projects?limit=1&id=gt.3&select=id,name"
+          [("Prefer", "return=representation")]
+          [json| {"name": "PopOS"} |]
+        `shouldRespondWith`
+          [json|[{"id": 4, "name": "PopOS" }]|]
+          { matchStatus = 200 }
+
+    it "works with the limit and offset query params" $
+      request methodPatch "/projects?limit=2&offset=2&select=id,name"
```
steve-chavez, line 410: Offset working is on purpose, I don't see why not give users this flexbility that already comes for free.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -105,24 +107,52 @@ mutateRequestToQuery (Insert mainQi iCols body onConflct putConditions returning
...
+    SQL.sql $ "SELECT " <> emptyBodyReturnedColumns <> " FROM " <> fromQi mainQi <> " WHERE false"
+
+  | range == allRange =
+    let
+      cols = BS.intercalate ", " (pgFmtIdent <> const " = _." <> pgFmtIdent <$> S.toList uCols) in
+    "WITH " <> normalizedBody body <> " " <>
+    "UPDATE " <> mainTbl <> " SET " <> SQL.sql cols <> " " <>
+    "FROM (SELECT * FROM json_populate_recordset (null::" <> mainTbl <> " , " <> SQL.sql selectBody <> " )) _ " <>
+    whereLogic <> " " <>
+    SQL.sql (returningF mainQi returnings)
```
steve-chavez, line 130: Keeping the old UPDATE as-is to avoid any breaking change.

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -386,3 +386,71 @@ spec = do
...
+      request methodPatch "/articles?limit=1&offset=1&select=id,body"
+          [("Prefer", "return=representation")]
+          [json| {"body": "This is a row changing. Do you see?"} |]
+        `shouldRespondWith`
+          [json| [ { "body": "This is a row changing. Do you see?", "id": 2 } ] |]
+          { matchStatus = 200 }
+
+    it "works on a table without a pk" $
+      request methodPatch "/no_pk?limit=1&offset=1"
+          [("Prefer", "return=representation")]
```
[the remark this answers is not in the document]
steve-chavez: Seems easier to document as well - only PK needed for this feature to work.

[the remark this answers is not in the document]
steve-chavez: Removing the `ctid` also would let us show a better error message in case there's no PK.

Will go ahead on this one.

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -115,3 +115,81 @@ spec =
...
+        request methodDelete "/projects?limit=1&offset=4&select=id,name"
+            [("Prefer", "return=representation")]
+            mempty
+          `shouldRespondWith`
+            [json|[
+              {
+                  "id": 5,
+                  "name": "Orphan"
+              }]|]
+            { matchStatus = 200 }
```
wolfgangwalther, line 130: I haven't had the time to look at the code, yet. But those test cases seem like they don't really test what we need to test.

>  - #2156, `limit/offset` now limits the affected rows on UPDATE/DELETE  - @steve-chavez
>   + Previously, `limit/offset` only limited the returned rows but not the actual updated rows

The returned rows are not really changed. They were the same before and after your change.

We'll need to test that the actual `UPDATE` or `DELETE` really affected only those rows, too. I guess we'll need some requests with `Prefer: tx=commit`, then check for which rows exist in the database, then rollback the changes manually. Because it's actual mutations to the database, it would be better to create a new table for those tests then, to not conflict with other test cases running on the same fixtures.

steve-chavez: True. Fair enough, I'll update the tests.

steve-chavez: Ok done. I must say, doing tests for real mutations is too tedious/painful.. we need utilitarian functions for this. Not willing to do that here though, could be added later.

wolfgangwalther:
Will look at this later today.

wolfgangwalther:
The docs say:

> It will work on a view if the PK or `ctid <https://www.postgresql.org/docs/current/ddl-system-columns.html>`_ is included as one of its columns.

Yet, I can't see a test-case with a view that exposes the ctid column. Did I just miss it? I would really like to see an example.

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -334,3 +335,12 @@ unknownLiteral = unknownEncoder . encodeUtf8
...
 intercalateSnippet frag snippets = foldr1 (\a b -> a <> SQL.sql frag <> b) snippets
+
+-- the "ctid" system column is always available to tables
+mutRangeF :: QualifiedIdentifier -> [FieldName] -> (SqlFragment, SqlFragment)
+mutRangeF mainQi rangeId = (
+    BS.intercalate " AND " $
+          (\col -> pgFmtColumn mainQi col <> " = " <> pgFmtColumn (QualifiedIdentifier mempty "pgrst_affected_rows") col) <$>
+          (if null rangeId then ["ctid"] else rangeId)
+  , if null rangeId then pgFmtColumn mainQi "ctid" else BS.intercalate ", " (pgFmtColumn mainQi <$> rangeId)
+  )
```
wolfgangwalther, line 346: Do I understand correctly that for a view, the `ctid` column will just be tried, too? And if it's not exposed, then the query will fail?

This is potentially dangerous, because a user could just create a view with a column named `ctid`, which does **not** select from the underlying table. Maybe it's not even unique. And then, the whole limited updates/delete feature would not work anymore - it would suddenly mutate more rows than requested.

steve-chavez: > Do I understand correctly that for a view, the ctid column will just be tried, too? And if it's not exposed, then the query will fail?

Yes, exactly that, I just let it fail.

> with a column named ctid, which does not select from the underlying table. Maybe it's not even unique. And then, the whole limited updates/delete feature would not work anymore - it would suddenly mutate more rows than requested.

Seems you cannot do that:

```
postgres=# create table hehe(ctid int primary key);
ERROR:  column name "ctid" conflicts with a system column name


```

steve-chavez: Also not for views:

```
postgres=# create table hehe_view as select 'hehe' as ctid, name from test.projects;
ERROR:  column name "ctid" conflicts with a system column name
```

steve-chavez: Argh, the above is for tables but for views it does work:

```sql
postgres=# create view hehe_view as select 'hehe' as ctid, name from test.projects;
CREATE VIEW
postgres=# table hehe_view ;
 ctid |    name
------+------------
 hehe | Windows 7
 hehe | Windows 10
 hehe | IOS
 hehe | OSX
 hehe | Orphan
(5 rows)
```

wolfgangwalther:
I don't think it's possible to support this feature for all views from which we can infer PKs, but only for **auto-updateable** views. Those views are required to have only one table in the `FROM` by PostgreSQL, among others. In those cases, there is a strict 1-to-1 relationship between rows in the view and in the underlying table, and this feature can work. That's why PostgreSQL makes those views auto-updateable by default.

But as soon as you have multiple tables in the `FROM` or have set-returning functions in the `SELECT` etc. - this is not guaranteed to work anymore: The columns in the view might be based on the tables PK columns - but they are not neccessarily unique anymore.

Example:

```sql
create table t (p int primary key);
create view v as
select p, generate_series(1, 2) from t;
-- then create some basic instead of triggers for DELETE
```

When you run a request on this view to delete `limit=2` rows, it might actually delete 1 rows, or 2 rows. Depending on whether the same `p` is returned from the CTE twice or not.

It's possible to construct other examples that do the opposite: Deleting **more** rows than were requested.

src/PostgREST/Query/SqlFragment.hs
wolfgangwalther: Yes, just tried that, too.

So it seems we could extend the view parser to tell us whether a column is in-fact a ctid column on the base table. It wouldn't even matter which name it had in the view. That would be a much better solution, I think - although much more effort.

If you want to do something quick, I'd suggest to skip the `ctid` part for now?

steve-chavez: Hm, so what's the move for this one, remove the `ctid` fallback and only work for PKs?

steve-chavez:
All true with the above, nice catch! :baseball: 

> but only for auto-updateable views.

We can determine that by checking that all of the `tableInsertable`/`tableUpdatable`/`tableDeletable` are true right?

https://github.com/PostgREST/postgrest/blob/1d3db3563e93a48623a8cec27a2bc9a15fbbe681/src/PostgREST/DbStructure/Table.hs#L19-L28

https://github.com/PostgREST/postgrest/blob/1d3db3563e93a48623a8cec27a2bc9a15fbbe681/src/PostgREST/DbStructure.hs#L336-L351

Or those somehow will also be true if there are INSTEAD OF triggers defined for the view? In which case we also need to check that there are no INSTEAD OF triggers defined?

wolfgangwalther:
> Or those somehow will also be true if there are INSTEAD OF triggers defined for the view? In which case we also need to check that there are no INSTEAD OF triggers defined?

Those are currently determined by `pg_relation_is_updatable`, which is an pg-internal function as it's not documented. However:
https://github.com/postgres/postgres/blob/f6f0db4d62400ff88f523dcc4d7e25f9506bc0d8/src/backend/utils/adt/misc.c#L611-L625

The second argument is a boolean `include_triggers`. We call it with `TRUE`, so the fields you mentioned do include `INSTEAD OF` triggers. You'd need the result of the call with `FALSE` to tell whether it's **auto**-updateable or not, I think.

wolfgangwalther:
> We can determine that by checking that all of the `tableInsertable`/`tableUpdatable`/`tableDeletable` are true right?

And once you have a version of that, that is only true for "auto", then it would be enough to only check for the update case for `UPDATE` and the delete case for `DELETE`.

You could have an auto-updateable view that still defines an instead of trigger - but maybe only for one of the two. Although I don't know what `pg_relation_is_updateable` returns in that case anyway.

steve-chavez:
Hm, so if we are using the `pg_relation_is_updatable` with TRUE as mentioned [above](https://github.com/PostgREST/postgrest/pull/2195#issuecomment-1079272200) then the following tests are not true?

https://github.com/PostgREST/postgrest/blob/1d3db3563e93a48623a8cec27a2bc9a15fbbe681/test/spec/Feature/OptionsSpec.hs#L39-L50

Because it's not testing for the `auto` on "auto-updatable"? They're just "updatable" thanks to the INSTEAD OF triggers?

steve-chavez:
Ok, so what I'm thinking is we should have an additional atribute in our `Table` structure for `isAutoUpdatableView` that will use the `pg_relation_is_updateable` with FALSE on the 3 operations(maybe only one is necessary though).

wolfgangwalther:
> Ok, so what I'm thinking is we should have an additional atribute in our `Table` structure for `isAutoUpdatableView` that will use the `pg_relation_is_updateable` with FALSE on the 3 operations(maybe only one is necessary though).

Maybe we can just switch the existing fields from a boolean to something like: `Manually | Automatic | No`. "Manually" would be the case when triggers were created - but also with `CREATE RULE`, that's why I didn't use "trigger" in it.

steve-chavez:
> You could have an auto-updateable view that still defines an instead of trigger - but maybe only for one of the two. Although I don't know what pg_relation_is_updateable returns in that case anyway.

According to @laurenceisla the `pg_relation_is_updateable` is not accurate in the presence of RULEs so detecting only auto-updatable views is going to take more work.

I'll just disable this feature for views now and leave it for a later enhancement.

steve-chavez:
The [501 Not Implemented](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) status is good for missing implementations, whenever PATCH/DELETE is tried on views it will respond with that now. If the view is not present in the internal cache, if will also reject the request.

I've also left the `ctid` fallback since it will work now that we're dealing only with tables.

src/PostgREST/Query/SqlFragment.hs
steve-chavez: Leaving the `ctid` now that the feature only works for tables.

laurenceisla (approved):

src/PostgREST/App.hs
```diff
@@ -391,7 +395,10 @@ handleSingleUpsert identifier context@(RequestContext _ _ ApiRequest{..} _) = do
       response HTTP.status204 [] mempty
 
 handleDelete :: QualifiedIdentifier -> RequestContext -> DbHandler Wai.Response
-handleDelete identifier context@(RequestContext _ _ ApiRequest{..} _) = do
+handleDelete identifier context@(RequestContext _ ctxDbStructure ApiRequest{..} _) = do
+  when (iTopLevelRange /= RangeQuery.allRange && findIfView identifier (dbTables ctxDbStructure)) $
+    throwError $ Error.NotImplemented "limit/offset is not implemented for views"
+
```
wolfgangwalther, line 401: Having this in `App.hs` really seems the wrong place. This should go into `ApiRequest.hs`, I think - because this module is concerned with parsing the user's request and matching it up with the schema cache.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -105,28 +106,63 @@ mutateRequestToQuery (Insert mainQi iCols body onConflct putConditions returning
...
+
+  | otherwise =
+    "WITH " <> normalizedBody body <> ", " <>
+    "pgrst_update_body AS (SELECT * FROM json_populate_recordset (null::" <> mainTbl <> " , " <> SQL.sql selectBody <> " ) LIMIT 1), " <>
+    "pgrst_affected_rows AS (" <>
+      "SELECT " <> SQL.sql rangeIdF <> " FROM " <> mainTbl <>
+       whereLogic <> " " <>
+      "ORDER BY " <> SQL.sql rangeIdF <> " " <> limitOffsetF range <>
```
wolfgangwalther, line 130: Do we need to order this by `rangeIdF` here? Shouldn't we order this by whatever was put in the query string as `?order=...`? Otherwise the result of limit will be unpredictable.

wolfgangwalther (commented):
None of the test-cases tests the interaction between `order=..` and `limit=...` - and I think this is still broken.

wolfgangwalther:
> [](https://github.com/steve-chavez)@[steve-chavez](https://github.com/steve-chavez) steve-chavez requested a review from [wolfgangwalther](https://github.com/wolfgangwalther) 7 hours ago

...

> [](https://github.com/steve-chavez)@[steve-chavez](https://github.com/steve-chavez) steve-chavez merged commit [f3f9030](https://github.com/PostgREST/postgrest/commit/f3f9030c4897e93e50e6f5dcecb317f2dca31a1b) into main [2 hours ago](https://github.com/PostgREST/postgrest/pull/2195#event-6311891762)

Really?

I feel pressurized to give quick reviews, because you seem to be in a hurry. I do have plenty of other things on my plate right now. I don't think this was ready to merge, yet.

And I really don't think we should merge a feature as quick as this, and then immediately create a release. We should have time to review **and test** stuff like this, before immediately bumping major...

**I don't like this!**

wolfgangwalther:
> None of the test-cases tests the interaction between `order=..` and `limit=...` - and I think this is still broken.

More specifically, I would expect a request like the following:
```http
DELETE /items?order=non_pk&limit=2
```

On a table like this:

| pk | non_pk |
| - | - |
| 1 | d |
| 2 | c |
| 3 | b |
| 4 | a |

To delete the 3/4 rows. However, I think right now the 1/2 rows are deleted.

This is what I read from the code. I wasn't able to test this myself, because I fear you're going to release this before I will be able to do so.

steve-chavez:
> We should have time to review and test stuff like this, before immediately bumping major...
> I don't like this!

Yeah, really sorry about this @wolfgangwalther. I also decided against doing a major release, so no worries. I'll only make a pre-release for this one.

steve-chavez:
> More specifically, I would expect a request like the following:
> DELETE /items?order=non_pk&limit=2

So basically the `DELETE /items?order=non_pk&limit=2` should delete the same items obtained through `GET /items?order=non_pk&limit=2`. True, that's not working right now, because the `order` is only applied to the returned rows and not the affected ones.

src/PostgREST/Query/SqlFragment.hs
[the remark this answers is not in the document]
steve-chavez: Given what [I mentioned below](https://github.com/PostgREST/postgrest/pull/2195#issuecomment-1079736464), it does make a lot of sense to order by `ctid` only and not by PK, because doing this:

```sql
INSERT INTO projects VALUES (1, 'Windows 7', 1);
INSERT INTO projects VALUES (2, 'Windows 10', 1);
INSERT INTO projects VALUES (5, 'Orphan', NULL);
INSERT INTO projects VALUES (4, 'OSX', 2);
INSERT INTO projects VALUES (3, 'IOS', 2);
```

Gives through http:

```http
GET /projects

 {"id":5,"name":"Orphan","client_id":null},
 {"id":3,"name":"IOS","client_id":2}]
```

Which is the same as doing `GET /projects?order=ctid` but not the same as doing `GET /projects?order=id`.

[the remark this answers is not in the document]
steve-chavez: So basically with `ctid` we can ensure that `DELETE /projects?limit=2` deletes the same rows as the ones obtained through `GET /projects?limit=2`.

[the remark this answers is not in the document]
steve-chavez: The only thing bad about preferring `ctid` is that you have to give explicit SELECT privilege to the column for it work(or a full wide SELECT) and that is weird - giving the PK columns a SELECT privilege is a more usual operation.

src/PostgREST/App.hs
steve-chavez: >This should go into ApiRequest.hs, I think - because this module is concerned with parsing the user's request and matching it up with the schema cache.

Hmm, I never thought of ApiRequest like that, I mean the only reason why the schema cache is used there is because of the function overloading which needs the Proc info and that is something that I thought be better moved to a different module.

If you check `handleSingleUpsert` it has a similar logic:

https://github.com/PostgREST/postgrest/blob/a3b0908548dba9f2c6df977c7076f46ec9b19ba8/src/PostgREST/App.hs#L373-L376

So to be consistent I thought `App.hs` was the better place for this logic.

src/PostgREST/Query/SqlFragment.hs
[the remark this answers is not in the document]
steve-chavez: So because of the above - the need to do an special `GRANT SELECT(ctid)` for `limit` to work - I think in fact `ctid` should be left out and `limit` should depend on PKs.

Later on, it would also be weird to tell users to expose `ctid` in ther VIEWs for `limit` to work.

src/PostgREST/Query/QueryBuilder.hs
wolfgangwalther: I'd say we just use the `?order` and add the `ctid` column at the end of that. If there is no `?order`, it's **just** the `ctid` column.

src/PostgREST/Query/SqlFragment.hs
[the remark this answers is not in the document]
wolfgangwalther: > The only thing bad about preferring ctid is that you have to give explicit SELECT privilege to the column for it to work(or a full wide SELECT) and that is weird - giving the PK columns a SELECT privilege is a more usual operation.

I don't think that's weird.

It should be a pretty descriptive error message anyway (lacking permission to select ctid) and we can add a hint about that on the feature page.

The advantage of using ctid:
- Consistent ordering with `GET` for both views with pks and without.
- Easier to explain ordering in the docs, if we do at all.
- More flexible, because you might want to explicitly **not** give `SELECT` on (a part of) the primary key. `ctid` is surely not sensitive, so granting a select for those bulk actions on it shouldn't be a problem.

src/PostgREST/App.hs
wolfgangwalther: Interesting. The `handleSingleUpsert` doesn't even depend on the schema cache, so it's even more out of place. It's an invalid request, so that should really go into `AppRequest`.

src/PostgREST/Query/SqlFragment.hs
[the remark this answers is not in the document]
steve-chavez: > ctid is surely not sensitive

Yeah, this is a nice advantage.

> Later on, it would also be weird to tell users to expose ctid in ther VIEWs for limit to work.

What about the above though? For views, it would be invasive to tell the user to explicitly expose the `ctid`. Usually views are used for exposing just the columns needed for the frontend.

Or perhaps we can have an additional rule, that for views you have to explicitly `order` by a column for the limited `update/delete` to work?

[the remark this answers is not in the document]
wolfgangwalther: Actually... it just occured to me, that we should just not use `ctid` for ordering at all. Without `?order=` this will then be exactly the same as `GET` - we don't use `ctid` there either. And PG sort order is not guaranteed, so we don't know whether it will always match up with `ctid`.

We can use `ctid` for the CTE join right now. And we can extend that later to views with whatever columns we use for that. But we don't need either for ordering.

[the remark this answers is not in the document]
steve-chavez: > And PG sort order is not guaranteed, so we don't know whether it will always match up with ctid.

True. And I also like the idea of not doing an implicit `order`. So using the `ctid` as default for both tables/views it is.

> We can use ctid for the CTE join right now. And we can extend that later to views with whatever columns we use for that.

How about if for both tables and views we use the explicit `order` column(s) for the join, that way is up to the client to decide the criteria. This would remove the need for smartness from the internal cache and also provide a way out of requiring `ctid` on views.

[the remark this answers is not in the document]
wolfgangwalther: This is quite an interesting idea. I'm not sure about whether I like the following suggestion, but if we take it a step further, we could do the following:
- Disallow any mutation request with `?limit` but without `?order`.
- Always use the `?order` columns for both the "inside the CTE `ORDER BY`" **and** the join condition

The great benefit at this point would be: This would support both tables and views without a problem.

The downside: To work properly, this would require the user to always use columns in `order` that form a unique identifier for a single row. Otherwise the query could mutate more rows than requested. As a first step, we could warn very explicitly about this behavior in the docs and tell the user that they need to use unique columns.

In a later step we could add an implicit ["required row count filter"](https://github.com/PostgREST/postgrest/pull/2164) to force a hard-limit on the mutated rows using something like `!row_count=lt.<limit>`. This would basically throw an error if the user used some non-unique set of columns, which would lead to a delete bigger than the requested limit. We should certainly not implement that second step right now, but wait for #2164 - but if we end up with this, this could be a very nice solution to support the whole thing on views properly - without any of those "need to expose column XYZ" challenges. It would be up to the user to know which columns fulfill the uniqueness requirement.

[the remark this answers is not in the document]
wolfgangwalther: > > The downside: To work properly, this would require the user to always use columns in order that form a unique identifier for a single row.
> 
> I think I can a put a nice error and even suggest trying the pks available on the schema cache, that should help with DX.

You can only show this error message in the case of "`?limit` without `?order` specified". When a user chooses to `order` on non-pk columns, you can't show this error, because you'd need to detect the uniqueness of those columns again - which is exactly what we have trouble with for views. The answer to that is my suggestion below...
 
> > In a later step we could add an implicit "required row count filter" to force a hard-limit on the mutated rows using something like !row_count=lte.
> 
> So like a `db-max-rows` for the `row_count` right? That's something I've been wanting as well. Unlike `db-max-rows` it would make sense to override it client side so if an operation requires mutating a large amount of rows the client must be explicit about that.

Now, I don't really understand what you mean by "`db-max-rows` for `row_count`, but I have a feeling there is mis-understanding here.

What I am thinking is the following:

Given this schema/data:

```sql
create view v as select pk, 'static' AS static from some_table;
-- insert 3 rows into some_table
```

Those 3 rows would have different PKs, but the same `static` column.

A query like this would work fine:

```http
DELETE /v?limit=2&order=pk
```

It would delete exactly 2 rows all the time, because `pk` is unique.

However, a query like this would not work as expected:

```http
DELETE /v?limit=2&order=static
```

Since `static` is the same for all 3 rows, it would select two of those rows in the CTE - but then join on that column and actually `DELETE` all 3 rows.

However, if we instead run this query implicitly as the following:
```http
DELETE /v!row_count=lte.2?limit=2&order=static
```

Then we would receive an error, because the actual mutation included 3 rows, which is bigger than the limit requested. In this case the transaction would be rolled back anyway.

If we can catch this error properly, and then show a better hint like "you need to use unique columns in `order`" - great.

Is this what you understood, too?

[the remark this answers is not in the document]
wolfgangwalther: I see. A server-side default, which is overridable from the client-side, would be completely unrelated to this issue here, though. The implicit row-count here would just be to avoid mistakes in the `order` clause - while the default you are proposing would be a protection against unlimited delete.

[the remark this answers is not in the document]
steve-chavez: > However, if we instead run this query implicitly as the following:
> DELETE /v!row_count=lte.2?limit=2&order=static

@wolfgangwalther Now that I've realized, this would mean that `limit` needs `row_count` and that for this feature both might as well be the same.

So how about this. I'll go ahead and apply the `row_count` logic for the `limit` but will not expose the actual `row_count` special filter. We can add that one later on when we settle on a syntax.

(I do see `row_count` being useful in other ways but for this feature we can apply it implicitly as you mentioned)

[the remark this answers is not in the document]
steve-chavez: And now that I think about it. If we enforce `row_count` implicitly on `limit`, then it would be safe to assume the PK cols for the `order/join` on views as well?

So the [above problem](https://github.com/PostgREST/postgrest/pull/2195#issuecomment-1079249217) would be no more?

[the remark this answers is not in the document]
wolfgangwalther: > So how about this. I'll go ahead and apply the `row_count` logic for the `limit` but will not expose the actual `row_count` special filter.

Ok.

> I mean because we'd be already fulfilling our part of the contract - the affected rows would not exceed the `limit`. And if the VIEW query reduces the amount of rows mapped to the PKs this would not be an issue.

I think it would still have very odd interactions with `offset`. For me, it'd be much simpler to implement and explain to always require and use the columns in `order` - and hint in the docs and error messages at the requirements for that. We'd need that for views without inferred PK columns anyway - and this way we don't have a complicated mix of what to do in which situation.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -105,28 +106,63 @@ mutateRequestToQuery (Insert mainQi iCols body onConflct putConditions returning
...
-      "FROM (SELECT * FROM json_populate_recordset (null::" <> SQL.sql (fromQi mainQi) <> " , " <> SQL.sql selectBody <> " )) _ " <>
-      (if null logicForest then mempty else "WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree mainQi <$> logicForest)) <> " " <>
-      SQL.sql (returningF mainQi returnings)
+    SQL.sql $ "SELECT " <> emptyBodyReturnedColumns <> " FROM " <> fromQi mainQi <> " WHERE false"
+
+  | range == allRange =
+    "WITH " <> normalizedBody body <> " " <>
+    "UPDATE " <> mainTbl <> " SET " <> SQL.sql nonRangeCols <> " " <>
+    "FROM (SELECT * FROM json_populate_recordset (null::" <> mainTbl <> " , " <> SQL.sql selectBody <> " )) _ " <>
+    whereLogic <> " " <>
```
wolfgangwalther, line 121: Looks like this change always adds a `WHERE` clause to the `UPDATE` query - even if the request is completely unconditional.

This breaks using `pg-safeupdate`. The request now goes through instead of failing because of a previously unconditional `UPDATE` query.

steve-chavez: Note that `pg-safeupdate` is not required anymore because we're disallowing [unfiltered patch](https://github.com/PostgREST/postgrest/issues/1959#issuecomment-1152813181)/[unfiltered delete](https://github.com/PostgREST/postgrest/issues/2314) unless a `limit` is applied.

> Looks like this change always adds a WHERE clause to the UPDATE query - even if the request is completely unconditional.

Yeah, because of the above, if no filter is added then `WHERE false` is done.

One thing better with the `pg-safeupdate` safeguard is that it fails with an error while the `WHERE false` will not do anything.
That's only the case of unfiltered UPDATE though. For unfiltered DELETE, [we're going to fail with an error](https://github.com/PostgREST/postgrest/blob/47a599371fc65d89da782ad31ce003ea0088b863/test/spec/Feature/Query/DeleteSpec.hs#L304-L318).

(I need to go back to this one because it will be inconsistent otherwise)

Would we need to keep compatibility with `pg-safeupdate` now that we have this? Why?

---

Not sure but maybe we could do `LIMIT 0` instead of `WHERE false`.

steve-chavez: > Yeah, because of the above, if no filter is added then WHERE false is done.

Oh, hell. I thought this was a review on https://github.com/PostgREST/postgrest/pull/2311.

steve-chavez: Addressing this in https://github.com/PostgREST/postgrest/pull/2405/

laurenceisla: > Looks like this change always adds a WHERE clause to the UPDATE query - even if the request is completely unconditional.

I think the change in this PR still allows the full table update without conditionals. The one that disallowed this was #2311 if I'm not mistaken.

steve-chavez: Compare https://github.com/PostgREST/postgrest/pull/2418

PR #2197, opened 2022-03-23
Addition of Content Range unit for RFC compliance

Oquile:
The Content Range response HTTP header is missing a unit declaration. It should be in the form :
`Content-Range: <unit> <range-start>-<range-end>/<size>`

Supplementary docs-
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
https://httpwg.org/specs/rfc7233.html#header.content-range

This change makes Postgrest compatible with Varnish.


Oquile:
Will re-roll and try to address failing tests.

PR #2199, opened 2022-03-23
Content Range unit for RFC 7233 compliance

Oquile:
The Content Range response HTTP header is missing a unit declaration. It should be in the form :
`Content-Range: <unit> <range-start>-<range-end>/<size>`

Supplementary docs-
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
https://httpwg.org/specs/rfc7233.html#header.content-range

This change also makes PostgREST compatible with Varnish HTTP Cache (https://varnish-cache.org/).

### DIFF
--- a/src/PostgREST/RangeQuery.hs
+++ b/src/PostgREST/RangeQuery.hs
@@ -93,7 +93,7 @@ contentRangeH :: (Integral a, Show a) => a -> a -> Maybe a -> Header
 contentRangeH lower upper total =
     ("Content-Range", toUtf8 headerValue)
     where
-      headerValue   = rangeString <> "/" <> totalString :: Text
+      headerValue   = "bytes" rangeString <> "/" <> totalString :: Text
       rangeString
         | totalNotZero && fromInRange = show lower <> "-" <> show upper
         | otherwise = "*"


Oquile:
This breaks build and tests. Will close and re-roll another

steve-chavez:
> This change also makes PostgREST compatible with Varnish HTTP Cache (https://varnish-cache.org/).

@Oquile Definitely something we want, the lack of range-unit was also reported in https://github.com/PostgREST/postgrest/issues/1089.

steve-chavez:
@Oquile You still interested in this one? Since it's a breaking change I'd be better to merge it now, I'll be doing a new release later today.

steve-chavez:
Seeing https://github.com/PostgREST/postgrest/issues/1089#issuecomment-515184613, I think we should go with `items` instead of `bytes` for our `<range-unit>`.

There's also a `Content-Range: rows..` [I found in the wild](https://www.codeproject.com/Articles/10256/Querying-a-SharePoint-Portal-Se) but I don't think the name matters as much, let's go with the most common one(which seems to be `items`).

Oquile:
@steve-chavez - It is (currently!) beyond my ablity to do a succesfully patch for this quickly.

wolfgangwalther:
> Since it's a breaking change I'd be better to merge it now, I'll be doing a new release later today.

I think it would be better to be able to discuss breaking changes - and not have a workflow of doing them and then releasing immediately.

If you want to make a release soon, this should not go in.

Issue #2188, opened 2022-03-09
feat: support SOAP?


fjf2002:
Is there any interest in this project to support creating SOAP endpoints? I don't think it would be hard.

I am building upon #1927, especially:
_Uploading raw text to a function. The function must be defined with a single unnamed text param and Content-Type: text/plain must be specified._

### My investigation until now with postgrest 9.0: It works with...
1. I have a function `soap_endpoint(text) RETURNS text ...`
2.  The HTTP request must contain header "Content-Type: text/plain"
3.  The HTTP request must contain header "Accept: text/plain" (otherwise the body seems to get enclosed in double quotes)
4. My function currently sets `PERFORM set_config('response.headers', '[{"content-type": "text/xml; charset=utf-8"}]', true);`
5. If my function returned data type xml, this would fail with Postgres error "Function string_agg(xml, unknown) does not exist".

But it would be nicer to be able to write a function `soap_endpoint(xml) RETURNS xml ...` and especially not having to deal with the HTTP header stuff.

### Suggested changes:

- To remedy my point 2, I think we would need (analogous to #1927):
_Uploading **XML** to a function. The function must be defined with a single unnamed **xml** param and Content-Type: **text/xml or application/xml or application/soap+xml** must be specified._

- To remedy my points 3 and 4, I think we would need: If function return type is xml, automatically set response content-type to text/xml (or something like that)

- To remedy my point 5, I think we would need: Allow returning xml data type, i. e. solve the string_agg issue - perhaps use xmlagg, or only allow scalar xml return values?

steve-chavez:
Yeah, we'd like to support XML endpoints as well.

I get your point about messing with headers. We had a long discussion on https://github.com/PostgREST/postgrest/pull/1582 about how to send raw bodies with the right content-type headers.

> Uploading XML to a function. The function must be defined with a single unnamed xml param and Content-Type: text/xml or application/xml or application/soap+xml must be specified.
> If function return type is xml, automatically set response content-type to text/xml (or something like that)
> perhaps use xmlagg, or only allow scalar xml return values??

That would be another option yes, it's less flexible than the approach in #1582 but seems much simpler to implement. Would you like to give it a shot?

For me the right approach would be the generic one in #1582 but I don't think there's any harm in providing a shortcut for XML responses in postgrest's core.

wolfgangwalther:
[...]

PR #2202, opened 2022-03-24
Allow returning XML from RPCs

fjf2002:
This PR would allow RPCs to return xml.

Part of #2188.

- Request must contain header 'Accept: text/xml'
- Will work for scalar RPC return type xml
- Will work for RPC table return type - everything gets concatenated (assuming every column is kind of xml)
- Will fail if request header 'Accept: text/xml' is ~~not~~ set and  RPC return type is scalar but is not xml.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -271,7 +271,9 @@ pgErrorStatus authed (SQL.SessionError (SQL.QueryError _ _ (SQL.ResultError rErr
         "P0001"   -> HTTP.status400 -- default code for "raise"
         'P':'0':_ -> HTTP.status500 -- PL/pgSQL Error
         'X':'X':_ -> HTTP.status500 -- internal Error
-        "42883"   -> HTTP.status404 -- undefined function
+        "42883"-> if BS.isPrefixOf "function xmlagg(" m
+          then HTTP.status406
+          else HTTP.status404 -- undefined function
         "42P01"   -> HTTP.status404 -- undefined table
         "42501"   -> if authed then HTTP.status403 else HTTP.status401 -- insufficient privilege
         'P':'T':n -> fromMaybe HTTP.status500 (HTTP.mkStatus <$> readMaybe n <*> pure m)
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -13,6 +13,7 @@ module PostgREST.Query.SqlFragment
   , asCsvF
   , asJsonF
   , asJsonSingleF
+  , asXmlF
   , countF
   , fromQi
   , limitOffsetF
@@ -175,6 +176,9 @@ asJsonSingleF :: Bool -> SqlFragment
 asJsonSingleF returnsScalar
   | returnsScalar = "coalesce((json_agg(_postgrest_t.pgrst_scalar)->0)::text, 'null')"
   | otherwise     = "coalesce((json_agg(_postgrest_t)->0)::text, 'null')"
+
+asXmlF :: FieldName -> SqlFragment
+asXmlF fieldName = "coalesce(xmlagg(_postgrest_t." <> pgFmtIdent fieldName <> "), '')"
 
 asBinaryF :: FieldName -> SqlFragment
 asBinaryF fieldName = "coalesce(string_agg(_postgrest_t." <> pgFmtIdent fieldName <> ", ''), '')"
--- a/src/PostgREST/Query/Statements.hs
+++ b/src/PostgREST/Query/Statements.hs
@@ -129,9 +129,9 @@ standardRow = (,,,,,) <$> nullableColumn HD.int8 <*> column HD.int8
 type ProcResults = (Maybe Int64, Int64, ByteString, Either Error [GucHeader], Either Error (Maybe Status))
 
 callProcStatement :: Bool -> Bool -> SQL.Snippet -> SQL.Snippet -> SQL.Snippet -> Bool ->
-                     Bool -> Bool -> Bool -> Maybe FieldName -> Bool ->
+                     Bool -> Bool -> Bool -> Bool -> Maybe FieldName -> Bool ->
                      SQL.Statement () ProcResults
-callProcStatement returnsScalar returnsSingle callProcQuery selectQuery countQuery countTotal asSingle asCsv multObjects binaryField =
+callProcStatement returnsScalar returnsSingle callProcQuery selectQuery countQuery countTotal asSingle asCsv asXml multObjects binaryField =
   SQL.dynamicallyParameterized snippet decodeProc
   where
     snippet =
@@ -151,6 +151,7 @@ callProcStatement returnsScalar returnsSingle callProcQuery selectQuery countQue
     bodyF
      | asSingle           = asJsonSingleF returnsScalar
      | asCsv              = asCsvF
+     | isJust binaryField && asXml = asXmlF $ fromJust binaryField
      | isJust binaryField = asBinaryF $ fromJust binaryField
      | returnsSingle
        && not multObjects = asJsonSingleF returnsScalar
--- a/test/spec/Feature/Query/RpcSpec.hs
+++ b/test/spec/Feature/Query/RpcSpec.hs
@@ -970,6 +970,37 @@ spec actualPgVersion =
             , matchHeaders = ["Content-Type" <:> "text/plain; charset=utf-8"]
             }
 
+        it "can get raw xml output with Accept: text/xml" $
+          request methodGet "/rpc/return_scalar_xml" (acceptHdrs "text/xml") ""
+            `shouldRespondWith`
+            "<my-xml-tag/>"
+            { matchStatus = 200
+            , matchHeaders = ["Content-Type" <:> "text/xml; charset=utf-8"]
+            }
+
+        it "can get raw xml output with Accept: text/xml" $
+          request methodGet "/rpc/welcome.xml" (acceptHdrs "text/xml") ""
+            `shouldRespondWith`
+            "<html>\n  <head>\n    <title>PostgREST</title>\n  </head>\n  <body>\n    <h1>Welcome to PostgREST</h1>\n  </body>\n</html>"
+            { matchStatus = 200
+            , matchHeaders = ["Content-Type" <:> "text/xml; charset=utf-8"]
+            }
+
+        it "should fail with function returning text and Accept: text/xml" $
+          request methodGet "/rpc/welcome" (acceptHdrs "text/xml") ""
+            `shouldRespondWith`
+            [json|
+              {
+                "hint":"No function matches the given name and argument types. You might need to add explicit type casts.",
+                "details":null,
+                "code":"42883",
+                "message":"function xmlagg(text) does not exist"
+              }
+            |]
+            { matchStatus = 406
+            , matchHeaders = ["Content-Type" <:> "application/json; charset=utf-8"]
+            }
+
       context "Proc that returns set of scalars" $
         it "can query without selecting column" $
           request methodGet "/rpc/welcome_twice"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #1917, Normalize the error response body by always having the `detail` and `hint` error fields with a `null` value if they are empty - @laurenceisla
  - #2176, Errors raised with `SQLSTATE` now include the message and the code in the response body - @laurenceisla
  - #2236, Support POSIX regular expression operators for row filtering - @enote-kane
+ - #2202, Allow returning XML from RPCs - @fjf2002
 
 ### Fixed
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -466,6 +466,7 @@ handleInvoke invMethod proc context@RequestContext{..} = do
         (shouldCount iPreferCount)
         (iAcceptContentType == CTSingularJSON)
         (iAcceptContentType == CTTextCSV)
+        (iAcceptContentType == CTTextXML)
         (iPreferParameters == Just MultipleObjects)
         bField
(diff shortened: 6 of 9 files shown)


src/PostgREST/Query/SqlFragment.hs
```diff
@@ -173,6 +174,12 @@ asJsonSingleF returnsScalar
   | returnsScalar = "coalesce((json_agg(_postgrest_t.pgrst_scalar)->0)::text, 'null')"
   | otherwise     = "coalesce((json_agg(_postgrest_t)->0)::text, 'null')"
 
+asXmlSingleF :: Bool -> SqlFragment
+asXmlSingleF returnsScalar
+  | returnsScalar = "coalesce(xmlagg(_postgrest_t.pgrst_scalar)::text, '')"
+  -- RPC not returning scalar: Hope every column in the table is xml and concat everything:
+  | otherwise     = "coalesce(string_agg((SELECT string_agg(json_each_text.value, '') FROM json_each_text(row_to_json(_postgrest_t))), '\n'), '')"
```
steve-chavez, line 181: Interesting :eyes:! Do you have a sample output of this query?

Btw, the codecov check will get green once you add tests.

fjf2002: The support of scalar xml return type is surely helpful. Whether the support of table return type is helpful, I am not sure. Nevertheless, the latter works like that:

The `RETURNS TABLE` function:
```sql
create or replace function public.x_table() returns table(xml xml, xml2 xml) language sql as $$
	select '<xml1/>'::xml, '<b1/>'::xml
	union all
	select '<xml2/>'::xml, '<b2/>'::xml
$$
```

The output on `curl --header 'Accept: text/xml' localhost:3000/rpc/x_table`:
```
<xml1/><b1/>
<xml2/><b2/>
```
Synergies with #1582 expected.

wolfgangwalther: Hm, with a composite return type the column names should not be lost. How about something based on `xml_forest`?

https://stackoverflow.com/a/65245107

fjf2002:
I have tried adding 2 test cases, see schema.sql in this PR. They fail with

`Final statement returns xml instead of test.xml at column 1.`

resp.

`return type mismatch in function declared to return test.xml`

What is PostgreSQL doing here, interpreting the well-known built-in `xml` type as hypothetical `test.xml` type, which obviously is missing? Can somebody help please?

wolfgangwalther:
> I have tried adding 2 test cases, see schema.sql in this PR. They fail with
> 
> `Final statement returns xml instead of test.xml at column 1.`
> 
> resp.
> 
> `return type mismatch in function declared to return test.xml`
> 
> What is PostgreSQL doing here, interpreting the well-known built-in `xml` type as hypothetical `test.xml` type, which obviously is missing? Can somebody help please?

We do have a type `test.xml`, because we have a view `test.xml`.

So I guess your function definition uses that, maybe try using `pg_catalog.xml` instead?

fjf2002:
> So I guess your function definition uses that, maybe try using `pg_catalog.xml` instead?

Yes, thank you. I fixed that in latest commit.

src/PostgREST/Query/SqlFragment.hs
fjf2002: Interesting idea. I can't manage it to get to work: Both xmlforest and xmlelement seem to need hardcoded parameter list resp. element names, thus would need some dynamic sql. How could that be integrated into the haskell code?

Meanwhile, the poor man's solution would be text-concatenation:

```sql
coalesce(string_agg((SELECT string_agg('<' || json_each_text.key || '>' || json_each_text.value || '</' || json_each_text.key || '>', '') FROM json_each_text(row_to_json(pg_user))), '\n'), '')
```

The json_each_text.key and json_each_text.value should probably be xml-escaped?

fjf2002: Note to myself: Above code will give undesired result if json_each_text.value is null.

wolfgangwalther: I thought a bit more about this and I think the basic question is: What kind of level do we want to have for XML support?

We support three levels of "content formatting" or whatever we call it:
- JSON does format multiple columns and also embedded resources into one big proper JSON document
- CSV formats the top-level nicely, but all nested resources pose a challenge. However, this seems to be a limitation of the CSV format itself, not strictly our implementation.
- TEXT/BYTEA formats only a single column - but does not support multiple columns in a useful way.

Even if we can make the multi-column XML approach work like suggested in this comment thread, it will still only be similar to the CSV style, right? So embedded resources would not be properly formatted as XML etc.

Since XML does in fact allow the nesting there is no point in implementing a half-baked solution like CSV does. However to properly implement a full XML representation including columns and nested resources, considerably more work would need to be done, correct?

I would suggest to focus this PR on only implementing XML support similar to TEXT/BYTEA, because everything else seems quite complex.

This means it could be enough to just add the xml media type here:

https://github.com/PostgREST/postgrest/blob/f3f9030c4897e93e50e6f5dcecb317f2dca31a1b/src/PostgREST/App.hs#L620-L622

And then treat it as a binary field. Not sure if that works with the built-in casting, but you can try.

fjf2002: > I would suggest to focus this PR on only implementing XML support similar to TEXT/BYTEA

I agree, as I said previously,

> Whether the support of table return type is helpful, I am not sure.

fjf2002: > This means it could be enough to just add the xml media type here [...] And then treat it as a binary field.

This still fails with Postgres error "Function string_agg(xml, unknown) does not exist" as reported in #2188.  I guess this happens in `SqlFragment.asBinaryF`?

So I suppose we need to fiddle "the information that the response shall be xml" though the call stack, from `App.handleInvoke` to `Statements.callProcStatement`, and add a function `SqlFragment.asXmlF`, analogous to `asBinaryF`?

See my updated PR, thank you.

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -970,6 +970,29 @@ spec actualPgVersion =
...
+                "hint":"No function matches the given name and argument types. You might need to add explicit type casts.",
+                "details":null,
+                "code":"42883",
+                "message":"function xmlagg(text) does not exist"
+              }
+            |]
+            { matchStatus = 404
+            , matchHeaders = ["Content-Type" <:> "application/json; charset=utf-8"]
+            }
+
```
wolfgangwalther, line 995: Hm. I'm not sure how, but I think we should answer with `406 Not Acceptable` and not `404 Not Found` in this case. It's not that we can't find `/rpc/welcome` - but we can't satisfy the `Accept: text/xml`.

fjf2002: @wolfgangwalther: That goes beyond my "pattern-recognition" haskell abilities. Can you help?

wolfgangwalther: Hm. Maybe we don't need to throw that error at all. Do we actually need `xmlagg()`? Or could we just use `string_agg()` with an explicit cast to `text`? xmlagg is not doing anything other that concatenating, right? And we'll just return text anyway.

That would allow returning non-xml with xml header, though. So maybe we just do `string_agg(col::pg_catalog.xml::text)`. On a real xml column, the first cast would be a no-op. But with a different column type (or invalid xml content for that matter) it would throw an error.

fjf2002: >  xmlagg is not doing anything other that concatenating, right?

Correct.

> That would allow returning non-xml with xml header, though. 

Yes, unfortunately. For example: I think a text of "four < eight" is not valid XML, the "<" would have to be escaped.

wolfgangwalther: > Yes, unfortunately. For example: I think a text of "four < eight" is not valid XML, the "<" would have to be escaped.

Yes, but I think my suggestion of `string_agg(col::pg_catalog.xml::text)` would prevent that.

fjf2002: Thanks, now I understand. I have committed the changes.

fjf2002:
Oh my. Does the failed codecov/patch check really prevent my PR from being considered, or how can I increase code coverage?

wolfgangwalther:
> Oh my. Does the failed codecov/patch check really prevent my PR from being considered [...]?

What makes you believe we wouldn't consider this PR?

If you feel progress is slow here, this would be for two reasons:
- My time is very limited right now, so I can't give feedback as quickly.
- We have not discussed implementing XML support thoroughly so far. So while I am very much convinced that we will end up with a good solution here, we're still kind of exploring our options along the way. This takes a few iterations.

The one thing that is currently still a question for me is, how to improve the error message / status code. From a HTTP perspective, we would want to have a `406 Not Acceptable` when XML is requested but the cast fails. We have already made progress, because 400 is certainly better than 404 and I feel we're in a better situation to catch the error we currently get and map it to a 406. I'm not 100% sure how to do it, yet, because ideally we wouldn't catch any other errors that are not the result of using `asXmlF`. I think what could happen right now is, that inside a RPC function the same casting error could be thrown and ideally we would not map that. Still have to think about this.

---


> how can I increase code coverage?

Not sure why the codecov job does not add annotations to the Changes tab right now, normally it highlights all changed lines that are not covered, yet. However you can also run `postgrest-coverage` locally and then examine the results to see which lines you changed are not covered, yet. We can then go from there to add tests.

In general, our goal is to increase coverage to 100% eventually, so we'd like to cover everything we add right now. This might not always be possible, so far, because of our testing infrastructure, but those cases are very likely found in IO tests. We're looking at api / spec tests here and in this case it's always possible to achieve 100% coverage.

test/spec/Feature/Query/RpcSpec.hs
steve-chavez: How about just catching the `function xmlagg(*) does not exist` error in:

https://github.com/PostgREST/postgrest/blob/057e8d63bfa351e5ba0518109dd9058aa113da4b/src/PostgREST/Error.hs#L250-L254

The 2nd parameter(`m`) is the message I believe - then you can map that to a 406 status code.

I also think we should go back to `xmlagg` instead of `stringagg`, the strictness will likely lead to better error messages for malformed xml.

steve-chavez: The above would be the simpler solution I think. Note that `octet-stream` and `text/plain` are also giving a 400 error instead of a 406 currently. 

The ideal way to solve all of these issues is to add more smartness to our internal cache(DbStructure). Likely by detecting that the function `returns xml` before executing it <- but this will be more complicated and can be done in a later PR.

steve-chavez:
I found the coverage detection of dead code useful before and it also failed for me on another PR but now I think it just went crazy, I'm getting:

```
[nix-shell:~/Projects/postgrest]$ postgrest-coverage

src/PostgREST/Config/PgVersion.hs:42: pgVersion109
src/PostgREST/Config/PgVersion.hs:45: pgVersion110
src/PostgREST/Config/PgVersion.hs:48: pgVersion112
src/PostgREST/Config/PgVersion.hs:51: pgVersion114
src/PostgREST/Config/PgVersion.hs:54: pgVersion121
src/PostgREST/Config/PgVersion.hs:57: pgVersion130
src/PostgREST/Config/PgVersion.hs:42: pgVersion109
src/PostgREST/Config/PgVersion.hs:45: pgVersion110
src/PostgREST/Config/PgVersion.hs:48: pgVersion112
src/PostgREST/Config/PgVersion.hs:51: pgVersion114
src/PostgREST/Config/PgVersion.hs:54: pgVersion121
src/PostgREST/Config/PgVersion.hs:57: pgVersion130
Found dead code: Check file list above.
```

(For this PR)

Which makes no sense since the pgVersion's are actually used.

@fjf2002 Check my suggestion [above](https://github.com/PostgREST/postgrest/pull/2202#discussion_r855521641) about the 406 error and then rebase - maybe the coverage error will go away then and we can merge.

fjf2002:
@[steve-chavez](https://github.com/steve-chavez): Done.

src/PostgREST/Query/Statements.hs
```diff
@@ -151,8 +151,10 @@ callProcStatement returnsScalar returnsSingle callProcQuery selectQuery countQue
     (countCTEF, countResultF) = countF countQuery countTotal
 
     bodyF
+     | asXml              = asXmlSingleF returnsScalar
```
steve-chavez, line 154: @fjf2002 There's no `asXmlSingleF` defined, that's why all compilation/tests are failing.

fjf2002: Thanks. I am not used to `git rebase`, took me quite a while and - well, this code line was re-introduced.

src/PostgREST/Error.hs
```diff
@@ -271,7 +271,9 @@ pgErrorStatus authed (SQL.SessionError (SQL.QueryError _ _ (SQL.ResultError rErr
         "P0001"   -> HTTP.status400 -- default code for "raise"
         'P':'0':_ -> HTTP.status500 -- PL/pgSQL Error
         'X':'X':_ -> HTTP.status500 -- internal Error
-        "42883"   -> HTTP.status404 -- undefined function
+        "42883"-> if BS.unpack m == "function xmlagg(text) does not exist"
```
steve-chavez, line 274: Hm, this will not catch other parameter types. How about using [isPrefixOf](https://hackage.haskell.org/package/bytestring-0.11.3.0/docs/Data-ByteString-Char8.html#v:isPrefixOf) and catching `function xmlagg(`?

fjf2002: OK, done.

steve-chavez:
@fjf2002 Nice work! All that's left to do is to add the feature to the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#added).

fjf2002:
I have added a line to the CHANGELOG.

fjf2002:
Ok, thanks! - And thanks for the support on my first PR.

PR #2203, opened 2022-03-24
fix: json/jsonb columns should have type "object" in OpenAPI spec

### LINKED ISSUE #2165: JSONB column should not have type "string" but "{}"
Issue:
I have a table with a jsonb column and I would expect the swagger editor to work correctly and allow sending a nested json object of any kind.

To reproduce:
Create a table with jsonb column, e.g.
```
create table sample (
                        id serial primary key,
                        data jsonb,
                        insertDate timestamptz
);
```

The swagger editor provides a sample data like this for the POST endpoint:
```
{
  "id": 0,
  "data": "string",
  "insertdate": "string"
}
```
Changing it to something like this, makes the swagger editor not send the request anymore:
```
{
  "data": {"a":"1"},
  "insertdate": "2022-02-07T13:49:51.141+00:00"
}
```

When I change the type of data from
```
  data:
        format: jsonb
        type: string
```
to
```
  data:
        format: jsonb
        type: {}
```

everything works fine!

alvare:
Fixes #2165

I couldn't find a way to type `toSwaggerType` that didn't involve `DataKinds` and importing from `Data.Swagger.Internal` ... 🤔 but glad to change that if someone knows how.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
- - #2165, Fix json/jsonb columns should have type "object" in OpenAPI spec
+ - #2165, Fix json/jsonb columns should not have type in OpenAPI spec
  - #2020, Execute deferred constraint triggers when using `Prefer: tx=rollback` - @wolfgangwalther
  - #2058, Return 204 No Content without Content-Type for PUT - @wolfgangwalther
  - #2077, Fix `is` not working with upper or mixed case values like `NULL, TrUe, FaLsE` - @steve-chavez
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+ - #2165, Fix json/jsonb columns should have type "object" in OpenAPI spec
  - #2020, Execute deferred constraint triggers when using `Prefer: tx=rollback` - @wolfgangwalther
  - #2058, Return 204 No Content without Content-Type for PUT - @wolfgangwalther
  - #2077, Fix `is` not working with upper or mixed case values like `NULL, TrUe, FaLsE` - @steve-chavez
--- a/src/PostgREST/OpenAPI.hs
+++ b/src/PostgREST/OpenAPI.hs
@@ -25,7 +25,6 @@ import Network.URI                (URI (..), URIAuth (..))
 import Control.Lens (at, (.~), (?~))
 
 import Data.Swagger
-import Data.Swagger.Internal (SwaggerKind (SwaggerKindSchema))
 
 import PostgREST.Config                   (AppConfig (..), Proxy (..),
                                            isMalformedProxyUri, toURI)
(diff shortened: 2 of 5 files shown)


steve-chavez:
Thanks for the contrib! 

I ran the styler+linter locally and added a commit. Besides that it's all looking good.

Just waiting for the github actions which seem to be in maintenance right now: https://www.githubstatus.com/incidents/qt1dj8v3whg9

wolfgangwalther:
As mentioned in https://github.com/PostgREST/postgrest/issues/2165#issuecomment-1039562684, I think adding a `type: object` is **wrong**. json/jsonb columns can hold much more than just objects.

alvare:
> As mentioned in [#2165 (comment)](https://github.com/PostgREST/postgrest/issues/2165#issuecomment-1039562684), I think adding a `type: object` is **wrong**. json/jsonb columns can hold much more than just objects.

Okay, just pushed some changes for that. LMK what you think.

src/PostgREST/OpenAPI.hs
```diff
@@ -2,6 +2,7 @@
 Module      : PostgREST.OpenAPI
 Description : Generates the OpenAPI output
 -}
+{-# LANGUAGE DataKinds       #-}
```
wolfgangwalther, line 5: left-over?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+ - #2165, Fix json/jsonb columns should have type "object" in OpenAPI spec
```
wolfgangwalther, line 24: Needs to be adjusted without "object". And add yourself, as in the other entries.

test/io/test_io.py
```diff
@@ -883,17 +883,19 @@ def test_admin_live_dependent_on_main_app(defaultenv):
...
 
     with run(env=defaultenv, port=freeport(), host=specialhostvalue) as postgrest:
 
-      response = postgrest.admin.get("/live")
-      assert response.status_code == 200
+        response = postgrest.admin.get("/live")
+        assert response.status_code == 200
+
+        response = postgrest.admin.get("/ready")
+        assert response.status_code == 200
```
wolfgangwalther, line 897: I'm not sure what happened here. I guess you ran `postgrest-style`? I just don't understand why that code was formatted like that. It should have stopped CI from passing and never been committed like that.

wolfgangwalther (approved):
@steve-chavez, WDYT? Returning no type is the best we can do for json/jsonb, right?

test/io/test_io.py
alvare: 🤷  @steve-chavez ran the linter and commited that

steve-chavez: This one is mine actually, I ran `postgrest-style` and commited. Not sure what happened too, but I assumed no harm is done if it's added here.

steve-chavez:
> @steve-chavez, WDYT? Returning no type is the best we can do for json/jsonb, right?

Yes, all good. I thought `object` was an improvement over `string` anyways but `any` is even better.

test/io/test_io.py
wolfgangwalther: Hm. It's certainly worth more investigation. As I said, I would have expected CI to pick that up and fail in main already. It doesn't. What else is CI missing to notice right now? :confused:

wolfgangwalther:
> > @steve-chavez, WDYT? Returning no type is the best we can do for json/jsonb, right?
> 
> Yes, all good. I thought `object` was an improvement over `string` anyways but `any` is even better.

Ah, is there something like `type: "any"`? Or does the absence of type have that meaning?

steve-chavez (approved):

PR #2204, opened 2022-03-25
fix: compliant range headers

steve-chavez:
Addresses https://github.com/PostgREST/postgrest/issues/1089#issuecomment-515184613.

Previous PR on https://github.com/PostgREST/postgrest/pull/2199.

### DIFF
--- a/src/PostgREST/RangeQuery.hs
+++ b/src/PostgREST/RangeQuery.hs
@@ -3,8 +3,7 @@ Module      : PostgREST.RangeQuery
 Description : Logic regarding the `Range`/`Content-Range` headers and `limit`/`offset` querystring arguments.
 -}
 module PostgREST.RangeQuery (
-  rangeParse
-, rangeRequested
+  rangeRequested
 , rangeLimit
 , rangeOffset
 , restrictRange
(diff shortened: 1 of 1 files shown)


steve-chavez:
@Oquile Maybe give this one a try and see if it helps with Varnish.

wolfgangwalther:
If we implement the response header part, we should also implement the request header part (`Range: items=1-2`...).

I'm not sure, though, what we're doing there right now, given this:
https://github.com/PostgREST/postgrest/blob/a3c1d9977f60de1849a9697d65bfbe2bea1aa1df/test/spec/SpecHelper.hs#L203-L204

Are we using a `Range-Unit` header? Is that actually a thing? I can't find it anywhere.

So it looks like all of this needs more cleanup - and definitely some tests!

steve-chavez:
> If we implement the response header part, we should also implement the request header part (Range: items=1-2...).

That's already done here actually.

> So it looks like all of this needs more cleanup - and definitely some tests!

Yeah, will leave it for a next release though.

wolfgangwalther:
> > If we implement the response header part, we should also implement the request header part (Range: items=1-2...).
> 
> That's already done here actually.

Perfect. I only had the other PR in mind.

> Yeah, will leave it for a next release though.

Just thinking.. will it be a breaking change at all? If a unit was required, then our implementation would just be buggy - and I wouldn't consider that a breaking change then.

PR #2211, opened 2022-03-30
Limited update/delete part 2

steve-chavez:
Continuing the work on https://github.com/PostgREST/postgrest/pull/2195.

As agreed on https://github.com/PostgREST/postgrest/pull/2195#discussion_r835779754, `limit` now works on views if they do include an explicit `order`.

### Steps

- [x] Works on views with explicit `?order`
- [x] Remove the default `ctid` fallback(tables will also need an explicit `?order`)
- [x] Apply an implicit `row_count` equal to the `limit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + #1689, Add the ability to run without `db-anon-role` disabling anonymous access. - @wolfgangwalther
  - #1543, Allow access to fields of composite types in select=, order= and filters through JSON operators -> and ->>. - @wolfgangwalther
  - #2075, Allow access to array items in ?select=, ?order= and filters through JSON operators -> and ->>. - @wolfgangwalther
- - #2156, Allow applying `limit/offset` to UPDATE/DELETE to only affect a subset of rows - @steve-chavez
-   + Uses the table primary key, so it needs a select privilege on the primary key columns
-   + If no primary key is available, it will fallback to using the "ctid" system column(will also require a select privilege on it)
-   + Doesn't work on views and it will throw an error if tried
+ - #2156, #2211, Allow applying `limit/offset` to UPDATE/DELETE to only affect a subset of rows - @steve-chavez
+   + It requires an explicit `order` on a unique column(s)
  - #1917, Add error codes with the `"PGRST"` prefix to the error response body to differentiate PostgREST errors from PostgreSQL errors - @laurenceisla
  - #1917, Normalize the error response body by always having the `detail` and `hint` error fields with a `null` value if they are empty - @laurenceisla
  - #2176, Errors raised with `SQLSTATE` now include the message and the code in the response body - @laurenceisla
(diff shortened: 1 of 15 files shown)


steve-chavez:
> Apply an implicit row_count equal to the limit

Added the above mechanism but I'm failing to come up with a test that proves when can it happen.

> It's possible to construct other examples that do the opposite: Deleting **more** rows than were requested.

@wolfgangwalther Can you come up with a view that does the above(discussed in https://github.com/PostgREST/postgrest/pull/2195#issuecomment-1079249217)

steve-chavez:
If the order is on a non-unique set of columns we cannot know for sure which items will be deleted.. however the LIMIT will still apply.

So for now it's looking like it's not necessary to apply the `row_count`.

steve-chavez:
Btw, this one seems to need https://github.com/PostgREST/postgrest/pull/1929. It's likely that an user will test `limit=0` and it will be weird to fail with the `HTTP Range error` message.

wolfgangwalther:
> If the order is on a non-unique set of columns we cannot know for sure which items will be deleted.. however the LIMIT will still apply.

That's not true, I gave an example already: https://github.com/PostgREST/postgrest/pull/2195#discussion_r837405117

The problem is in the join: Even though the CTE might be limited correctly, the join will target more rows, because the join condition is not on unique columns.

wolfgangwalther:
> > It's possible to construct other examples that do the opposite: Deleting **more** rows than were requested.
> 
> @wolfgangwalther Can you come up with a view that does the above(discussed in [#2195 (comment)](https://github.com/PostgREST/postgrest/pull/2195#issuecomment-1079249217))

It's certainly possible with a "not-a-simple-view", i.e. one that is not auto-updateable. In this case, to even support mutations, we need some `INSTEAD OF` triggers - and then behavior is pretty much undefined anyway, because PostgREST doesn't know what will happen. Even more important in that case, that the responsibility for the correctly `limit`ed request is on the user, because only they know what columns they need to make the ordering/join on.

Here's an example:

```sql
create table ta (ca int primary key);
create table tb (cb int primary key);
create view v as select * from ta, tb;

create function instead_of_delete() returns trigger
language plpgsql as $$
begin
  delete from ta where ta.ca = old.ca;
  return old;
end
$$;

create trigger ... instead of delete ...;

-- Suppose v has:
-- ca | cb
-- 10 | 10
-- 11 | 10
-- 10 | 11
-- 11 | 11
```

Run a delete with `limit=1` and you will delete one row from `ta` - but two rows from `v`.

steve-chavez:
> That's not true, I gave an example already: https://github.com/PostgREST/postgrest/pull/2195#discussion_r837405117

Nice, I included that example in a test.

> Here, it's not really the "limited delete" feature that is causing this. It's just the instead of trigger doing that.
> So in this case the row_count filter would not even work, because it doesn't know about those extra rows deleted.

Cool, thanks for clarifying that out. So a test on a view with that kind of instead trigger is not necessary.

steve-chavez:
By moving the order/limit conditionals to the ApiRequest, we gained some performance:

[head vs main now](https://github.com/PostgREST/postgrest/pull/2211/checks?check_run_id=6244463559)

throughput | 387 | 388
-- | -- | --

[head vs main before](https://github.com/PostgREST/postgrest/pull/2211/checks?check_run_id=6238033542)

throughput | 377 | 382
-- | -- | --

PR #2239, opened 2022-04-08
Clarify the relationship in the embedding disambiguation error

laurenceisla:
Clarifies the content of the `relationship` key error. The content in square brackets `[]` are not part of the disambiguation syntax and may be mistakenly added to the request when trying to fix this error as seen here: https://github.com/supabase/supabase/discussions/6274#discussioncomment-2504906

> `select('id,from(id,username),to(id,username),series!adaptation_notifications_series_fkey[series](id,title,novel),status,when')`

This suggestion changes, for instance:

```
"relationship": "people_country_fk[country_id][id]"
```

to:

```
"relationship": "people_country_fk using people(country_id) and country(id)"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2042, Keep working when EMFILE(Too many open files) is reached. - @steve-chavez
  - #2147, Ignore `Content-Type` headers for `GET` requests when calling RPCs. Previously, `GET` without parameters, but with `Content-Type: text/plain` or `Content-Type: application/octet-stream` would fail with `404 Not Found`, even if a function without arguments was available.
  - #2155, Ignore `max-rows` on POST, PATCH, PUT and DELETE - @steve-chavez
- - #2239, Fix misleading disambiguation error for functions where the content of the `relationship` key looked like valid syntax - @laurenceisla
+ - #2239, Fix misleading disambiguation error where the content of the `relationship` key looks like valid syntax - @laurenceisla
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


test/spec/Feature/Query/EmbedDisambiguationSpec.hs
```diff
@@ -103,22 +103,22 @@ spec =
               "details": [
                 {
                   "cardinality": "many-to-many",
-                  "relationship": "test.whatev_jobs[whatev_jobs_site_id_1_fkey][whatev_jobs_project_id_1_fkey]",
+                  "relationship": "whatev_jobs using whatev_jobs_site_id_1_fkey and whatev_jobs_project_id_1_fkey",
```
steve-chavez, line 106: Can we add the fk columns for the m2m cases too? That would make the error more complete. It would look something like:

```json
"relationship": "whatev_jobs using whatev_jobs_site_id_1_fkey(job_id) and whatev_jobs_project_id_1_fkey(site_id)"
```

laurenceisla: Just to confirm, the columns will be the ones in the intermediate table, right? In this case the test would be:

```
"relationship": "whatev_jobs using whatev_jobs_site_id_1_fkey(site_id_1) and whatev_jobs_project_id_1_fkey(project_id_1)"
```

steve-chavez: Yes, correct!

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2042, Keep working when EMFILE(Too many open files) is reached. - @steve-chavez
  - #2147, Ignore `Content-Type` headers for `GET` requests when calling RPCs. Previously, `GET` without parameters, but with `Content-Type: text/plain` or `Content-Type: application/octet-stream` would fail with `404 Not Found`, even if a function without arguments was available.
  - #2155, Ignore `max-rows` on POST, PATCH, PUT and DELETE - @steve-chavez
+ - #2239, Fix misleading disambiguation error for functions where the content of the `relationship` key looked like valid syntax - @laurenceisla
```
steve-chavez, line 49: Not related to functions right?
```suggestion
 - #2239, Fix misleading disambiguation error where the content of the `relationship` key looked like valid syntax - @laurenceisla
```

PR #2250, opened 2022-04-18
Refactor/Cleaning

steve-chavez:
Some refactors needed for https://github.com/PostgREST/postgrest/issues/2070#issuecomment-1093547838.

The general idea is that we want to have the Relationships and PKCols as Table attributes in our internal cache. This will avoid unnecessary searches - increasing perf and making it easier to do some logic with table attributes.

It also gets us closer to only needing to execute a single query for the internal cache.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -57,14 +57,14 @@ import PostgREST.DbStructure.Relationship (Cardinality (..),
                                            Junction (..),
                                            PrimaryKey (..),
                                            Relationship (..))
-import PostgREST.DbStructure.Table        (Column (..), Table (..))
+import PostgREST.DbStructure.Table        (Column (..), Table (..), TablesMap)
 
 import Protolude
 import Protolude.Unsafe (unsafeHead)
 
 
 data DbStructure = DbStructure
-  { dbTables        :: [Table]
+  { dbTables        :: TablesMap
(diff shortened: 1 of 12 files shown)


laurenceisla (approved):

src/PostgREST/DbStructure.hs
```diff
@@ -721,50 +681,52 @@ allPrimaryKeys tabs =
...
+            information_schema._pg_expandarray(c.conkey) AS x
+          FROM pg_namespace nr
+          JOIN pg_class r
+            ON nr.oid = r.relnamespace
+          JOIN pg_constraint c
+            ON r.oid = c.conrelid
+          JOIN pg_namespace nc
+            ON c.connamespace = nc.oid
+          WHERE
+            c.contype in ('p', 'u')
```
wolfgangwalther, line 705: Why was `'f'` removed here?

steve-chavez: @wolfgangwalther I elaborated on that on this commit message: https://github.com/PostgREST/postgrest/pull/2250/commits/37fa43c879f32e5c9351d45a605a6257bbfa0431

This query was meant for getting all the primary key cols and the 'f'(foreign key) doesn't make sense there.

I assumed this happened because this query was copied from `information_schema.key_column_usage` <- which searches all constraints.

'u' should likely be removed as well.

PR #2256, opened 2022-04-20
test: add big schema openapi test

steve-chavez:
The big_schema(`apflora`) we use internally for perf tests is OSS https://github.com/barbalex/apf2/tree/master/sql/apflora so we can include it in our CI tests.

I've also asked barbalex to be sure - alles gut.

---

apflora stats:

```
select count(*)  from information_schema.routines where specific_schema = 'apflora';
 count
-------
    12

select count(*)  from information_schema.tables where table_schema = 'apflora' and table_type = 'BASE TABLE';
 count
-------
    45
(1 row)

select count(*)  from information_schema.views where table_schema = 'apflora';
 count
-------
   281

```

It doesn't have that many tables, the views are likely what makes our internal cache so big; probably the relationships generated between them(strict m2m junctions should help a bit here).

### DIFF
--- a/test/io/big_schema.sql
+++ b/test/io/big_schema.sql
@@ -1,0 +1,11321 @@
+/*
+This is a 2018 version of the apflora schema https://github.com/barbalex/apf2/tree/master/sql/apflora - latest version likely has differing contents
+
+We use it to test our metadata generation because it contains a good amount of views(281).
+
+Custom roles and privileges where removed.
+
+postgrest-with-postgresql-14 -f test/io/big_schema.sql psql
+
+select count(*)  from information_schema.routines where specific_schema = 'apflora';
+ count
+-------
+    12
+
+select count(*)  from information_schema.tables where table_schema = 'apflora' and table_type = 'BASE TABLE';
+ count
+-------
+    45
+(1 row)
+
+select count(*)  from information_schema.views where table_schema = 'apflora';
+ count
+-------
+   281
+(1 row)
+*/
+
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
+SET client_encoding = 'UTF8';
+SET standard_conforming_strings = on;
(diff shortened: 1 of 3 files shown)


steve-chavez:
So this now reveals something I've discovered locally, doing a `curl localhost:3000`(openapi) fails for `apflora` with:

```
[nix-shell:~/Projects/postgrest]$ PGRST_OPENAPI_MODE="ignore-privileges" PGRST_DB_ANON_ROLE="postgrest_test_authenticator" PGRST_DB_SCHEMAS='apflora' postgrest-with-postgresql-14 -f test/io/big_schema.sql  postgrest-run
20/Apr/2022:17:59:35 -0500: Attempting to connect to the database...
20/Apr/2022:17:59:35 -0500: Connection successful
20/Apr/2022:17:59:35 -0500: Listening on port 3000
20/Apr/2022:17:59:35 -0500: Listening for notifications on the pgrst channel
20/Apr/2022:17:59:35 -0500: Config reloaded
20/Apr/2022:17:59:35 -0500: Schema cache loaded
stack overflow
```

Just a `stack overflow` error.

I've managed to reproduce it with https://github.com/PostgREST/postgrest/commit/bc7d54c7d2df309a6d84a9b04ccda0912c2112f2 (before recent refactors) as well.

steve-chavez:
Ah, this is because of https://github.com/PostgREST/postgrest/pull/2122 - there's a stack limit(`-with-rtsopts=-KxK`) on `postgrest-run`(1K) and on the spec tests(33K).

steve-chavez:
So for now I'm skipping the io test, I've confirmed it passes with 200K stack size but it'd be sad to lose the 1K safeguard for the rest of the codebase. It's still nice to have apflora in the codebase for quick tests.

Once OpenAPI stops doing things like searching all table columns over a list for each table(having Column inside Table would prevent this) I'll reenable the test - hopefully it requires much less stack then.

PR #2260, opened 2022-04-22
refactor: merge Column list into Table

steve-chavez:
Merges the SQL query of `allColumns` into `allTables`.

I thought with this we'd get a considerable reduction of the apflora `dump-schema` output but:

```
du -sh old.json new.json

68M    old.json
67M     new.json
```

Just 1M reduction. Which goes on to show the big reduction [before](https://github.com/PostgREST/postgrest/pull/2254#issuecomment-1104163025) was because the bulk of the size comes from the Relationship list.

PR #2262, opened 2022-04-23
fix: restrict generated m2m relationships

steve-chavez:
Fixes https://github.com/PostgREST/postgrest/issues/2070.

Now we only add m2m relationship based on [strict junctions](https://en.wikipedia.org/wiki/Associative_entity):

- A table has FKs to two other tables
- **These FK columns are part of the table's PK columns.**

### DIFF
--- a/test/spec/fixtures/schema.sql
+++ b/test/spec/fixtures/schema.sql
@@ -791,18 +791,13 @@ CREATE TABLE files (
 );
 
 CREATE TABLE touched_files (
-    user_id integer NOT NULL,
-    task_id integer NOT NULL,
-    project_id integer NOT NULL,
-    filename text NOT NULL,
-    CONSTRAINT fk_users_tasks
-	FOREIGN KEY (user_id, task_id)
-	REFERENCES users_tasks (user_id, task_id)
-	ON DELETE CASCADE ON UPDATE CASCADE,
-    CONSTRAINT fk_upload
-	FOREIGN KEY (project_id, filename)
-	REFERENCES files (project_id,filename)
-	ON DELETE CASCADE ON UPDATE CASCADE
+  user_id integer NOT NULL,
+  task_id integer NOT NULL,
+  project_id integer NOT NULL,
+  filename text NOT NULL,
+  constraint fk_users_tasks foreign key (user_id, task_id) references users_tasks (user_id, task_id) on delete cascade on update cascade,
+  constraint fk_upload foreign key (project_id, filename) references files (project_id,filename) on delete cascade on update cascade,
+  primary key(user_id, task_id, project_id, filename)
 );
 
 create table private.articles (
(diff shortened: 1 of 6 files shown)


steve-chavez:
Some stats regarding the big schema(apflora):

- On the main branch, we have 179940 relationships detected - where 144396 are m2m relationships.
- On this PR, we now have 35544 relationships detected - where none are m2ms. It makes sense since no table has a composite PK.

This now reduces the `dump-schema` output to `8.6M` from the previous [67M](https://github.com/PostgREST/postgrest/pull/2260#issue-1212641135).

steve-chavez:
> 35544 relationships

These are still a lot considering `apflora` has 45 tables and 281 views. Moving forward we probably need a way to annotate(through a `COMMENT` or other mechanism) a VIEW and only in that case generation relationships for it.

test/spec/fixtures/schema.sql
```diff
@@ -1955,12 +1952,13 @@ create table whatev_sites (
...
 
 create table whatev_jobs (
-  job_id        uuid    primary key
+  job_id        uuid
 , name text
 , site_id_1     int not null references whatev_sites (id)
 , project_id_1  int not null references whatev_projects (id)
 , site_id_2     int not null references whatev_sites (id)
 , project_id_2  int not null references whatev_projects (id)
+, primary key(job_id, site_id_1, project_id_1, site_id_2, project_id_2)
```
steve-chavez, line 1961: This PK was added so this test keeps passing:

https://github.com/PostgREST/postgrest/blob/9bdb4e5c4f723c01f1a77f7b03575c7f8efb368f/test/spec/Feature/Query/EmbedDisambiguationSpec.hs#L121-L157

I do still wonder if this type of modeling can happen in the real world. Still, keeping this for completion.

test/spec/fixtures/schema.sql
```diff
@@ -2294,7 +2292,8 @@ A test for partitioned tables$$;
         car_dealer_city varchar(64) not null,
         quantity int not null,
         foreign key (car_model_name, car_model_year) references test.car_models (name, year),
-        foreign key (car_dealer_name, car_dealer_city) references test.car_dealers (name, city)
+        foreign key (car_dealer_name, car_dealer_city) references test.car_dealers (name, city),
+        primary key (car_model_name, car_model_year, car_dealer_name, car_dealer_city, quantity)
```
steve-chavez, line 2296: I believe this new detection makes perfect sense. It also accounts for cases where the PK has an additional column(`quantity`) that is not part of an FK.

laurenceisla (approved):

bolerodan:
Are there any builds or nightly builds where I can test these recent changes (like there was for the Ltree merge)?

steve-chavez:
The loadtest reports a drop in performance here: https://github.com/PostgREST/postgrest/runs/6142982808. 

Rate: 445(main) vs 407(head).

Not sure why though. The change here is only related to the schema cache and it reduces the amount of relationships searched on each request.

PR #2266, opened 2022-04-26
perf: improve embedding performance

steve-chavez:


test/spec/Feature/Query/EmbedDisambiguationSpec.hs
```diff
@@ -99,17 +99,17 @@ spec =
...
                 {
-                    "cardinality": "one-to-many",
-                    "relationship": "departments_head_id_fkey using agents(id) and departments(head_id)",
+                    "cardinality": "many-to-one",
+                    "relationship": "agents_department_id_fkey using agents(department_id) and departments(id)",
                     "embedding": "agents with departments"
                 }
               ],
-              "hint": "Try changing 'departments' to one of the following: 'departments!agents_department_id_fkey', 'departments!departments_head_id_fkey'. Find the desired relationship in the 'details' key.",
+              "hint": "Try changing 'departments' to one of the following: 'departments!departments_head_id_fkey', 'departments!agents_department_id_fkey'. Find the desired relationship in the 'details' key.",
```
steve-chavez, line 112: The order changed, but tests the same.

test/spec/Feature/Query/EmbedDisambiguationSpec.hs
```diff
@@ -99,17 +99,17 @@ spec =
...
               "message": "Could not embed because more than one relationship was found for 'agents' and 'departments'",
```
steve-chavez, line 113: The order changed, but tests the same.

steve-chavez:
Seems this didn't actually cause a performance increase - in fact it might have caused a slight performance decrease:

Doing the [GetSingleEmbed](https://github.com/supabase/benchmarks/blob/master/postgrest/k6/GETSingleEmbed.js) 5 times on:

- v9.0.0: 1330.618783/s, 1374.860516/s, 1373.573841/s, 1359.105488/s, 1344.366727/s
- PR: 1306.490444/s, 1295.490369/s, 1314.2843/s, 1316.905683/s, 1302.176807/s

Which also coincides with the [LoadTest result on CI](https://github.com/PostgREST/postgrest/runs/6168647103)(rate 366 vs 361)

Seems really weird to me considering that a HashMap should be faster than the List. Perhaps it's because the amount of relationships is not that big on our fixtures to matter.

I guess these commits can be considered a refactor/cleaning then.

PR #2269, opened 2022-04-28
Allow limit=0 in query params to return an empty array

### LINKED ISSUE #1121: Feature request: support for filter/condition that is always false (ie. where false)

laurenceisla:
Closes #1121.

Continues what was discussed in PR #1929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2176, Errors raised with `SQLSTATE` now include the message and the code in the response body - @laurenceisla
  - #2236, Support POSIX regular expression operators for row filtering - @enote-kane
  - #2202, Allow returning XML from RPCs - @fjf2002
+ - #2269, Allow `limit=0` in the request query to return an empty array - @gautam1168, @laurenceisla
 
 ### Fixed
 
--- a/src/PostgREST/RangeQuery.hs
+++ b/src/PostgREST/RangeQuery.hs
@@ -10,6 +10,8 @@ module PostgREST.RangeQuery (
 , restrictRange
 , rangeGeq
 , allRange
+, limitZeroRange
+, hasLimitZero
 , NonnegRange
 , rangeStatusHeader
 , contentRangeH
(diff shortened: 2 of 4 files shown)


steve-chavez (approved):
Nice small change :+1:

PR #2270, opened 2022-04-29
Reorganize error groups

laurenceisla:
Followed the example of [PostgreSQL internal errors](https://www.postgresql.org/docs/current/errcodes-appendix.html) and separated the Hasql errors from the group sequence (now starts with X instead of 4) and changed the name to Internal errors.

Also moved some General errors that should be in the Api Request group.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -226,7 +226,7 @@ instance JSON.ToJSON SQL.CommandError where
   toJSON (SQL.ResultError resultError) = JSON.object [
     "code"    .= InternalErrorCode00,
     "message" .= (show resultError :: Text),
-    "details" .= ("Internal error." :: Text),
+    "details" .= JSON.Null,
     "hint"    .= JSON.Null]
 
   toJSON (SQL.ClientError d) = JSON.object [
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -223,27 +223,27 @@ instance JSON.ToJSON SQL.CommandError where
         "hint"    .= (fmap T.decodeUtf8 h :: Maybe Text)]
 
   toJSON (SQL.ResultError (SQL.UnexpectedResult m)) = JSON.object [
-    "code"    .= HasqlErrorCode00,
+    "code"    .= InternalErrorCode00,
     "message" .= (m :: Text),
     "details" .= JSON.Null,
     "hint"    .= JSON.Null]
   toJSON (SQL.ResultError (SQL.RowError i SQL.EndOfInput)) = JSON.object [
-    "code"    .= HasqlErrorCode01,
+    "code"    .= InternalErrorCode01,
     "message" .= ("Row error: end of input" :: Text),
     "details" .= ("Attempt to parse more columns than there are in the result" :: Text),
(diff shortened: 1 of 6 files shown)


steve-chavez:
Link to docs: https://postgrest--530.org.readthedocs.build/en/530/errors.html#postgrest-error-codes

steve-chavez:
Code | Description
-- | --
PGRST400 | Related to Setting Response Headers.
PGRST401 | The status code must be a positive integer. See Setting Response Status Code.
PGRST402 | Related to Binary Output. See Providing images for  for an example on requesting images.
PGRST403 | The requested feature is not implemented in some way.

I think those could be considered ApiRequest errors too.

For example `PGRST400` happens when the response.headers are not in the format exposed mentioned in https://postgrest--530.org.readthedocs.build/en/530/api.html#setting-response-headers. So it depends on the db state but the rule is static unlike failures from the schema cache. `PGRST401` is the same.

`PGRST402` happens when a single column is not selected on a `GET /tbl` request or a set returning function IIRC. So it can be considered an Api Request error as well.

`PGRST403` is the same. Though it will be gone once #2211 is gone.

With that I think we can get rid of the General group.

steve-chavez:
Now that https://github.com/PostgREST/postgrest/pull/2211 is merged, `PGRST403` can be removed.

steve-chavez:
Codecov fails because there's no way to prove that the [InternalErrors](https://postgrest--530.org.readthedocs.build/en/530/errors.html#group-x-internal) happen.

I've also never seen these errors reported during my time here.

Instead of having each type of [ResultError](https://hackage.haskell.org/package/hasql-1.5.0.2/docs/Hasql-Session.html#t:ResultError) separated, how about if we just do a `show` of `ResultError` - hasql already has a Show instance for that.

This should decrease the lost percentage in coverage, perhaps codecov will turn green with that.

steve-chavez:
To be clear, with the above change we'd only have a single InternalError.

laurenceisla:
> This should decrease the lost percentage in coverage, perhaps codecov will turn green with that.

It didn't, but the overall coverage % of the project increased (+0.64%).

src/PostgREST/Error.hs
```diff
@@ -217,36 +218,17 @@ instance JSON.ToJSON SQL.QueryError where
...
-    "details" .= i,
+    "code"    .= (T.decodeUtf8 c      :: Text),
+    "message" .= (T.decodeUtf8 m      :: Text),
+    "details" .= (fmap T.decodeUtf8 d :: Maybe Text),
+    "hint"    .= (fmap T.decodeUtf8 h :: Maybe Text)]
+
+  toJSON (SQL.ResultError resultError) = JSON.object [
+    "code"    .= InternalErrorCode00,
+    "message" .= (show resultError :: Text),
+    "details" .= ("Internal error." :: Text),
```
steve-chavez, line 229: I think a `null` would be preferable to this since it doesn't really add more "details".

steve-chavez:
> It didn't, but the overall coverage % of the project increased (+0.64%).

Cool, I think that's good enough. Otherwise the other option would be to eat this error and not expose it to the client.

Will add a todo on https://github.com/PostgREST/postgrest/pull/1713 to work around this.

steve-chavez (approved):

PR #2274, opened 2022-05-04
Fix Windows build using Stack in CI

### LINKED ISSUE #2271: Drop Stack support
> As nixpkgs provides us with a consistent set of Haskell packages, stack (and stack2nix) has no additional value over cabal for us.

(previously mentioned on https://github.com/PostgREST/postgrest/pull/1829#issuecomment-826153676)

We've put a lot of effort into our [Nix tooling](https://github.com/PostgREST/postgrest/tree/main/nix#nix-development-and-build-environment) and we no longer use [stack](https://docs.haskellstack.org/en/stable/README/) for development.

If we can build with [cabal](https://www.haskell.org/cabal/download.html) on [Windows](https://github.com/PostgREST/postgrest/blob/main/.github/workflows/ci.yaml#L164-L176) and [FreeBSD CIs](https://github.com/PostgREST/postgrest/blob/main/.cirrus.yml) then we can drop Stack support.

---

[Stack is currently failing for our Windows build](https://github.com/PostgREST/postgrest/runs/6267208827?check_suite_focus=true), preventing our release process to fire.

laurenceisla:
Related to  #2271

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -161,19 +161,19 @@ jobs:
             pgdir: /usr/local/Cellar/postgresql
             artifact: postgrest-macos-x64
 
-          - name: Windows
-            runs-on: windows-latest
-            cache: |
-              ~\AppData\Roaming\stack
-              ~\AppData\Local\Programs\stack
-              .stack-work
-            deps: |
-              stack exec -- pacman -Syu --noconfirm
-              stack exec -- pacman -S msys2-keyring --noconfirm
-              stack exec -- pacman -S mingw64/mingw-w64-x86_64-postgresql --noconfirm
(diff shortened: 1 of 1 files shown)


laurenceisla:
Couldn't make it work replacing stack with cabal. Got the same error mentioned here: https://stackoverflow.com/questions/71891129/the-configure-script-cant-find-grep-on-windows-on-github-actions where `grep` couldn't be found to build some libraries (it looks like `msys2` doesn't expose that binary when building with cabal).

Ended up removing the dependencies on the `msys2-keyring`, if they were just added to work with the `libpq` library then replacing the by adding the `PGBIN` to the path should work in the same way (that is the case, it seems).

steve-chavez (approved):

PR #2279, opened 2022-05-09
fix: allow cast on types with underscores/numbers

steve-chavez:
Now possible:

```http
GET /tbl?select=oid_array::_int4
```

`_int4` is the base name of the `int[]` cast.

Allowing brackets `[]` in the parsers would be a bit more effort and also they need to be urlencoded.

### DIFF
--- a/test/spec/fixtures/schema.sql
+++ b/test/spec/fixtures/schema.sql
@@ -2589,7 +2589,7 @@ CREATE TABLE test.xmltest (
     xml pg_catalog.xml NOT NULL
 );
 
-INSERT INTO test.xmltest VALUES
-(1, '<myxml>foo</myxml>'),
-(2, 'bar'),
-(3, '<foobar><baz/></foobar>');
+CREATE TABLE oid_test(
+  id int,
+  oid_col oid,
+  oid_array_col oid[]);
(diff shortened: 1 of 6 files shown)


test/spec/fixtures/schema.sql
```diff
@@ -2589,7 +2589,7 @@ CREATE TABLE test.xmltest (
     xml pg_catalog.xml NOT NULL
 );
 
-INSERT INTO test.xmltest VALUES
-(1, '<myxml>foo</myxml>'),
-(2, 'bar'),
-(3, '<foobar><baz/></foobar>');
```
steve-chavez, line 2595: @fjf2002 Just FYI, I put this on `data.sql`

steve-chavez, line 2595: @fjf2002 Just FYI, I've moved this to `test/spec/fixtures/data.sql`

skulltech:
Hi @steve-chavez I'm running into an issue and it seems like it's related to casting to array types, i.e. using brackets. Has that been added yet, or is there any progress on that?

If not, is there any workaround?

For example, this something I'd like to be supported

```
select member_ids::text[] from table
```

steve-chavez:
> If not, is there any workaround?

@skulltech Yes, you can use `member_ids::_text` instead. This also works with plain SQL:

```sql
select pg_typeof('{foo}'::_text);
 pg_typeof
-----------
 text[]
```

skulltech:
@steve-chavez 

Got it, thanks!

Also another question, I want to run a query like the following

```sql
select (profile).id::text, (profile).name, is_following from search_profiles();
```

But PostgREST is giving me the following error

```
unexpected "(" expecting field name (* or [a..z0..9_]) or "*"
```

Any workaround? Or something I'm missing?

Thank you so much for the help!!

PR #2284, opened 2022-05-12
Add doctest for pField

laurenceisla:
The `pField` function does not ignore trailing spaces when a column is selected:
```
/table?select=name,join (*) 
```
It does work for leading spaces though: 
```
/table?select=name, join(*)
``` 
Both trailing and leading spaces will be ignored in this PR.


laurenceisla:
Closed and left discussion open in issue #2285

Issue #2288, opened 2022-05-17
GHC 9.2.2 support


arrowd:
### Environment

* PostgREST version: 9.0.0 release

### Description of issue

The package doesn't build with GHC 9.2.2 because of `base < 4.16` dependency line in the `.cabal` file, while GHC 9.2.2 ships `base-4.16`. There might be more version bumps required to build PostgREST with GHC 9.2.2.

arrowd:
New protolude is on Hackage now.

arrowd:
All pieces seem to be in place. Hoping for a new release!

arrowd:
Can we get a new release containing this change?

PR #2292, opened 2022-05-25
Upgrade aeson to 2.0.*

robx:
Aeson 2.0 abstracts the object type (from `Data.HashMap.Strict` to `Data.Aeson.KeyMap`). From version 2.0.1.0, the default map implementation is switched to `Data.Map.Strict`, addressing the hash flooding vulnerability. This is also one step towards compatibility with GHC 9.2 (#2288).

Compare https://frasertweedale.github.io/blog-fp/posts/2021-10-12-aeson-hash-flooding-protection.html for some background.

Besides fixing the vulnerability, the change to `Data.Map.Strict` from `Data.HashMap.Strict` is likely to have a performance impact for some work loads.

Overview:
- update cabal file to require aeson-2, and a number of related bound update, specifically
  * jose-0.9
  * swagger-2.8
  * hasql-1.5
  These don't appear to have important changes.
- update sources to work with the abstract aeson `KeyMap` type, and change auth claims and CSV row maps accordingly
- fix IO tests for updated wai-logger
- update stack configuration for updated dependencies
- nix updates:
  * update nixpkgs to a version from 2022-03-30, which comes with aeson-2
  * update haskell package overrides
  * update static nix build from static-haskell-nix to build with updated nixpkgs
  * explicitly keep the old version of hlint in order to not deal with changed linter complaints in this PR

The load test fails as per https://github.com/PostgREST/postgrest/issues/2308 due to the nix dependency update; this will go away after merging.


nix/overlays/haskell-packages.nix
```diff
@@ -59,6 +59,15 @@ let
...
           { });
 
+      aeson =
+        lib.dontCheck (prev.callHackageDirect
+          {
+            pkg = "aeson";
+            ver = "2.0.3";
+            sha256 = "1y3jj4bhviss6ff9lwxki0zbdcl1rb398bk4s80zvfpnpy7p94cx";
+          }
+          { });
```
robx, line 69: I think I naively copied the warp entry correctly (with a wrong hash), but it results in an "infinite recursion"...

steve-chavez: Ah, seems the problem is that `callHackageDirect` depends on `cabal2nix`, which in turn depends on aeson.

https://github.com/NixOS/nixpkgs/issues/83098

robx: ah that makes sense, thanks!

robx: fixed this one by adding an extra cabal2nix overlay

robx:
I (think I) fixed the cabal2nix-related infinite recursion with the extra cabal2nix overlay, but now I'm getting another infinite recursion that I can't make heads or tails of.

The curious thing is that even after I remove `aeson` entirely from `overlays/haskell-packages.nix`, I see the following:

```
$ nix-build -A pkgs.haskell.packages.ghc8107.OneTuple --show-trace
error: infinite recursion encountered

       at /nix/store/hjaixgpg8m0xafqixjdgrd4i9gyq6irh-source/pkgs/stdenv/generic/make-derivation.nix:254:7:

          253|       depsHostHostPropagated      = lib.elemAt (lib.elemAt propagatedDependencies 1) 0;
          254|       propagatedBuildInputs       = lib.elemAt (lib.elemAt propagatedDependencies 1) 1;
             |       ^
          255|       depsTargetTargetPropagated  = lib.elemAt (lib.elemAt propagatedDependencies 2) 0;

       … while evaluating the attribute 'propagatedBuildInputs' of the derivation 'aeson-2.0.3.0'

       at /nix/store/hjaixgpg8m0xafqixjdgrd4i9gyq6irh-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while evaluating the attribute 'propagatedBuildInputs' of the derivation 'aeson-2.0.3.0'

       at /nix/store/hjaixgpg8m0xafqixjdgrd4i9gyq6irh-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let
```

I have no idea where that `aeson-2.0.3.0` even comes from...

robx:
> I have no idea where that aeson-2.0.3.0 even comes from...

That was because I'd copied over the aeson source hash. So, since it has the aeson sources in the store as a fixed output thingy, it tries building the aeson sources as OneTuple which... can't work.

steve-chavez:
Whoa, this looks like hard work. Unfortunately I'm also lost and don't know a simple way to solve the many dependency errors.

Another option(not sure how feasible), would be to try and replace aeson? I've been eyeing [jsonify](https://github.com/nikita-volkov/jsonifier) for a while.

robx:
> Whoa, this looks like hard work. Unfortunately I'm also lost and don't know a simple way to solve the many dependency errors.
> 
> Another option(not sure how feasible), would be to try and replace aeson? I've been eyeing [jsonifier](https://github.com/nikita-volkov/jsonifier) for a while.

Ah don't worry, I think it's a simple change that's just made complicated by a combination of nixpkg's rigidity and my lack of understanding. I *think* I can probably sort it out by also updating the nixpkgs pin, was just hoping to do only one thing at a time.

steve-chavez:
> Any idea what happened there? Is the test fragile

Yeah, that test can be changed, no problem about it. I've been doing the same on recent commits. In fact that root spec feature is not documented and will be dropped once we have https://github.com/PostgREST/postgrest/issues/2299.

robx:
@steve-chavez I think this is at a point now where it's "stable" from the point of view of the underlying nixpkgs change -- would it make sense to update cachix (`nix-shell --run postgrest-push-cachix`) to save on CI resources?

steve-chavez:
>  would it make sense to update cachix (nix-shell --run postgrest-push-cachix) to save on CI resources?

@robx Do you want me to do this manually for this PR? Because `postgrest-push-cachix` [runs on CI already](https://github.com/PostgREST/postgrest/blob/main/.github/workflows/ci.yaml#L140) and the cache gets refreshed when it runs on the main branch.

robx:
> > would it make sense to update cachix (nix-shell --run postgrest-push-cachix) to save on CI resources?
> 
> @robx Do you want me to do this manually for this PR? Because `postgrest-push-cachix` [runs on CI already](https://github.com/PostgREST/postgrest/blob/main/.github/workflows/ci.yaml#L140) and the cache gets refreshed when it runs on the main branch.

I was thinking about manually doing this, yeah, but not sure it's a good idea... Right now every single CI job takes 20 minutes to set up the nix environment. (I don't mind the time, just seems a bit expensive while iterating.)

Besides requiring manual work, a good argument against is that the nixpkgs version *might* need to change again after all. There's two outstanding issues here, the load test encoding failure and the static build. I expect they should be fixable without changing the nixpkgs version, but it's not really clear at this point.

steve-chavez:
>  the load test encoding failure

We can ignore that one here. It will be gone when merging to main. As seen on https://github.com/PostgREST/postgrest/pull/2302#issuecomment-1150170068

default.nix
```diff
@@ -141,7 +141,23 @@ rec {
...
+
+      nixpkgsPrev =
+        builtins.fetchTarball {
+          url = "https://github.com/nixos/nixpkgs/archive/${nixpkgsVersionPrev.rev}.tar.gz";
+          sha256 = nixpkgsVersionPrev.tarballHash;
+        };
+      pkgsPrev = import nixpkgsPrev { };
+      inherit (pkgsPrev) hlint;
+    in
+    pkgs.callPackage nix/tools/style.nix { inherit hlint hsie; };
```
robx, line 160: I'd file an issue and/or a follow-up PR to make postgrest hlint-clean for the updated hlint version, and to drop this shim again. But would rather not add linter fixes into the mix here, the PR ballooned enough already.

nix/hsie/default.nix
```diff
@@ -28,8 +28,8 @@ let
         mkdir -p $out/bin
         ln -s $hsie $out/bin/$name
       '';
-  bashCompletion =
+  bash-completion =
```
robx, line 31: The updated nixpkgs came with deprecation warnings against spelling it `bashCompletion` :shrug:

nix/patches/static-haskell-nix-isexecutable.patch
```diff
@@ -0,0 +1,18 @@
...
+   # Function that tells us if a given Haskell package has an executable.
+   # Pass only Haskell packages to this!
+   # Filter away other stuff with `isProperHaskellPackage` first.
+-  isExecutable = pkg:
+-    (pkgs.haskell.lib.overrideCabal pkg (drv: {
+-      passthru.isExecutable = drv.isExecutable or false;
+-    })).isExecutable;
++  # FIXME postgrest: original fails with our nixpkgs version, we
++  # just hardcode the check to work for a static postgrest build.
++  isExecutable = pkg: pkg.name == "postgrest";
```
robx, line 15: This isn't particularly elegant, but it works.

src/PostgREST/Request/ApiRequest.hs
```diff
@@ -18,11 +18,14 @@ module PostgREST.Request.ApiRequest
...
 
 import qualified Data.Aeson            as JSON
+import qualified Data.Aeson.Key        as K
+import qualified Data.Aeson.KeyMap     as KM
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy  as LBS
 import qualified Data.CaseInsensitive  as CI
 import qualified Data.Csv              as CSV
 import qualified Data.HashMap.Strict   as M
+import qualified Data.Map.Strict       as MS
```
robx, line 28: I would have preferred calling this `M`, but renaming `HashMap` to `HM` causes linter problems with all the `M` imports elsewhere in the codebase.

robx:
Assuming the tests pass 🤞, this should be good for review now.

The one thing I can't evaluate is if the switch to `Data.Map.Strict` has noticeable performance downsides. Do we run some benchmarks to track performance regressions?

src/PostgREST/Request/ApiRequest.hs
robx: Had to rename after all to placate the linters...

nix/hsie/default.nix
steve-chavez: That's weird. it's not about using a camelcase key right? Just about this particular identifier.

steve-chavez:
>The one thing I can't evaluate is if the switch to Data.Map.Strict has noticeable performance downsides. Do we run some benchmarks to track performance regressions?

Ah, that would be the `loadtest`, which unfortunately doesn't run here. But let me try some manual load tests.

nix/hsie/default.nix
robx: Yes, that particular one. It was changed from a warning to an exception here: https://github.com/NixOS/nixpkgs/pull/161146

steve-chavez (approved):
Wasn't able to do manual load tests(it takes too long to compile the static build :disappointed:), will do it when cachix gets updated. At any rate, we have to upgrade eventually. So this LGTM.

robx:
Here are load test results post-merge on main: https://github.com/PostgREST/postgrest/runs/6874664942
And the previous commit on main: https://github.com/PostgREST/postgrest/runs/6813060905

robx:
Randomly pulling the `HEAD` column from both reports, we get:

| metric | main | pr | pr/main |
| ------ | ----- | ----- | ----- |
| bytes_in.mean | 841 | 841 | 1.0 |
| bytes_in.total | 14688631 | 16883495 | 1.14 |
| bytes_out.mean | 18 | 18 | 1.0 |
| bytes_out.total | 309042 | 355416 | 1.15 |
| duration | 60000099777 | 60018094347 | 1.0 |
| latencies.50th | 1025381 | 929754 | 0.9 |
| latencies.90th | 13953938 | 10120705 | 0.72 |
| latencies.95th | 23995719 | 21513839 | 0.89 |
| latencies.99th | 26633982 | 22525434 | 0.84 |
| latencies.max | 43676099 | 38462247 | 0.88 |
| latencies.mean | 3436057 | 2987338 | 0.86 |
| latencies.min | 515510 | 429336 | 0.83 |
| latencies.total | 59997005490 | 59991731686 | 0.99 |
| rate | 291 | 335 | 1.15 |
| requests | 17461 | 20082 | 1.15 |
| status_codes.200 | 10477 | 12050 | 1.15 |
| status_codes.201 | 1746 | 2008 | 1.15 |
| status_codes.204 | 5238 | 6024 | 1.15 |
| success | 1 | 1 | 1.0 |
| throughput | 291 | 335 | 1.15 |
| wait | 25301762 | 949680 | 0.03 |

The `wait` value for pre-merge seems to be an outlier -- the `main` column is similar  to the post-merge result.

Other than that, it looks like the upgrade (whether aeson2 or some of the other dependency changes) rather improved results, looks to me like 15% throughput increase, matching latency decrease.

steve-chavez:
@robx Loadtest results look great! I've been trying to run other loadtests but..

Unfortunately, I no longer seem to be able to produce a static binary:

```
nix-build --attr postgrestStatic

ldd result/bin/postgrest
        linux-vdso.so.1 (0x00007ffe5bb03000)
        libz.so.1 => /nix/store/srww288f7sb0f8hb5399431k8ldg2pss-zlib-1.2.11/lib/libz.so.1 (0x00007fda1a070000)
        libpq.so.5 => /nix/store/8y5nh8zx42qbzw0ki5rc9ywgi7z3axqb-postgresql-14.2-lib/lib/libpq.so.5 (0x00007fda19d2c000)
        libgmp.so.10 => /nix/store/wxyn3pv3kvqrl2lnfzpaa82vsd6fizka-gmp-6.2.1/lib/libgmp.so.10 (0x00007fda19c8c000)
        libc.so => /nix/store/88nqrkcal1mmiqbx10ax6wkyncf1h1gk-musl-1.2.2/lib/libc.so (0x00007fda19bd9000)
        libffi.so.8 => /nix/store/wb3ji9r5b2f747phif5lfx7np45fl9px-libffi-3.4.2/lib/libffi.so.8 (0x00007fda19bcc000)
```

[9.0.1](https://github.com/PostgREST/postgrest/releases/tag/v9.0.1) shows:

```
ldd postgrest
$       not a dynamic executable
```

We need a test on CI to ensure the above output.

PR #2296, opened 2022-05-30
Cherry pick non-breaking fixes for a PATCH version release

laurenceisla:
Testing cherry picking fixes to release the PATCH version v9.0.1

Applied most commits in this range: ab90bd24e0fd9cf4fdd8af76990823c4c183f867 - 6b5370f145ad2f346ec803c0ddaa3f47ab86ed76
 excluding the features, breaking changes and other commits related to those. After that last commit there are many features implemented, so I only cherry-picked some small fixes and the relevant commits to the CI files (for this PR to run the tests).

Fixes included:

```markdown
 - #2165, Fix json/jsonb columns should not have type in OpenAPI spec
 - #2020, Execute deferred constraint triggers when using `Prefer: tx=rollback` - @wolfgangwalther
 - #2077, Fix `is` not working with upper or mixed case values like `NULL, TrUe, FaLsE` - @steve-chavez
 - #2024, Fix schema cache loading when views with XMLTABLE and DEFAULT are present - @wolfgangwalther
 - #1724, Fix wrong CORS header Authentication -> Authorization - @wolfgangwalther
 - #2120, Fix reading database configuration properly when `=` is present in value - @wolfgangwalther
 - #2135, Remove trigger functions from schema cache and OpenAPI output, because they can't be called directly anyway. - @wolfgangwalther
 - #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
 - #2101, Remove aggregates, procedures and window functions from the schema cache and OpenAPI output. - @wolfgangwalther
 - #2294, Disable parallel GC for better performance on higher core CPUs - @steve-chavez
 - #1076, Fix using CPU while idle - @steve-chavez
```


steve-chavez:
@laurenceisla Would it be possible to include these non-breaking fixes too?

```markdown
- #2107, Clarify error for failed schema cache load. - @steve-chavez
From Database connection lost. Retrying the connection to Could not query the database for the schema cache. Retrying.
- #2101, Remove aggregates, procedures and window functions from the schema cache and OpenAPI output. - @wolfgangwalther
- #2152, Remove functions, which are uncallable because of unnamend arguments from schema cache and OpenAPI output. - @wolfgangwalther
- #2153, Fix --dump-schema running with a wrong PG version. - @wolfgangwalther
- #2278, Allow casting to types with underscores and numbers(e.g. select=oid_array::_int4) - @steve-chavez
- #2294, Disable parallel GC for better performance on higher core CPUs - @steve-chavez
- #1076, Fix using CPU while idle - @steve-chavez
```

If some of them depend on a commit that included a feature, it's fine to leave them for v10.

laurenceisla:
Added many of the fixes you mentioned, but these ones may be more complicated so I excluded them:

> #2107, Clarify error for failed schema cache load. - @steve-chavez
From Database connection lost. Retrying the connection to Could not query the database for the schema cache. Retrying.

This adds a test for the health check feature here: [`51ee07`](https://github.com/PostgREST/postgrest/commit/51ee072f84b764796e8b52b9cf03ff4ac18c2428#diff-476936d59c75a5951c72d4e5d14e760c0ee78fe31c3a90b676df4ff9af69194dR798-R800). It can be fixed but may cause problems if a new minor version arises from this branch.

> #2152, Remove functions, which are uncallable because of unnamend arguments from schema cache and OpenAPI output. - @wolfgangwalther

It depends on d9f7f6f509e95a963dc0257b98fb52e172c7732a, which in turn looks like it depends on [`52d628`](https://github.com/PostgREST/postgrest/commit/52d628f1ed1393daeab401cc0cbe9b8c973c91a5#diff-32990f96190cb6f977c8b3a5937c81ec9fa5c9fa9874a46959ce8e08ed158ea6R292). It breaks tests for VARIADIC functions.

> #2278, Allow casting to types with underscores and numbers(e.g. select=oid_array::_int4) - @steve-chavez

It uses files created after a refactor.

steve-chavez:
Including https://github.com/PostgREST/postgrest/pull/2239 as it was straightforward.

steve-chavez:
> A practical approach to ease part of the problem would be to immediately release a new pre-release 9.0.1.20220630. This would then supersede 9.0.1 again.

Done: https://github.com/PostgREST/postgrest/releases/tag/v9.0.1.20220630

> But automatic dependency update tools like renovate or dependabot or whatever won't pick up on that.. so they will suggest an "update" from 9.0.0.xxx to 9.0.1 - which breaks things.

Hm, does dependabot work for upgrades on binaries/executables as well? I thought it was only for libraries. Also, does it pull pre-releases? If it sticks to releases only then there shouldn't be a problem.

> We discussed elsewhere (can't find it right now) why we need to tag the pre-release version numbers the way we do now, i.e. 
> This is quite annoying. We shouldn't not make any releases out-of-order and stick with a linear branch history.

I thought it would be preferable to have an out-of-order patch release than two consecutive major releases. I guess we just need to be more careful in merging breaking changes.

I do agree in sticking to linear history. This was an exceptional case and it does require some extra work to make it happen, I don't think we'll do it again.

> We discussed elsewhere (can't find it right now)

Maybe we should document the release rules/process in https://github.com/PostgREST/postgrest/blob/main/.github/CONTRIBUTING.md

wolfgangwalther:
> Done: https://github.com/PostgREST/postgrest/releases/tag/v9.0.1.20220630

Thanks.

> Hm, does dependabot work for upgrades on binaries/executables as well? I thought it was only for libraries.

I'm using renovate, which is highly configurable and can update **a lot** of stuff. I'm using an "infrastructure as code" approach so everything that runs on my servers is defined in a git repo. The repo is updated via PRs by renovate and synced with flux to the kubernetes cluster. So whenever a new postgrest docker image is released, I get automated updates / PRs.

> Also, does it pull pre-releases? If it sticks to releases only then there shouldn't be a problem.

Well, how would it be able to tell that it's a pre-release? It's just another regular version, but with 4 numbers. And I actually **want** to use pre-releases. Otherwise I'd be missing a lot of features, due to our low release frequency. I just don't want to have an automated downgrade inbetween...

PR #2300, opened 2022-06-02
RPC POST for function w/single unnamed XML param

fjf2002:
RPC POST for function w/single unnamed XML param

- extends #1927 for data type xml
- part of #2188

### DIFF
--- a/test/spec/Feature/Query/RpcSpec.hs
+++ b/test/spec/Feature/Query/RpcSpec.hs
@@ -1202,6 +1202,13 @@ spec actualPgVersion =
             `shouldRespondWith`
             [str|unnamed text arg|]
 
+        it "can insert xml directly" $
+          request methodPost "/rpc/unnamed_xml_param"
+            [("Content-Type", "text/xml"), ("Accept", "text/xml")]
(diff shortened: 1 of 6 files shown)


test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1202,6 +1202,13 @@ spec actualPgVersion =
             `shouldRespondWith`
             [str|unnamed text arg|]
 
+        it "can insert xml directly" $
+          request methodPost "/rpc/unnamed_xml_param"
+            [("Content-Type", "text/xml"), ("Accept", "text/xml")]
+            [str|unnamed xml arg|]
+            `shouldRespondWith`
+            [str|unnamed xml arg|]
```
steve-chavez, line 1210: Oh, didn't know this was valid XML. How about a more representative example? Like:

```suggestion
            [str|<note><from>John</from><to>Jane</to><message>Remember me</message></note>|]
            `shouldRespondWith`
            [str|<note><from>John</from><to>Jane</to><message>Remember me</message></note>|]
```

(Got it from https://www.w3schools.com/xml/xml_server.asp)

fjf2002: Ok, done.

PR #2302, opened 2022-06-07
Fix ambiguity on cabal builds

### LINKED ISSUE #2301: Builds using cabal throw errors (ambiguous types)
### Environment

* PostgreSQL version: 14.3
* PostgREST version: 9.0.0.20220531/ main 807364bd7b2adfcfc0e2ebc5efccd5150dc45652
* Operating system: ubuntu-arm64 / EndeavourOS

### Description of issue

Running `cabal v2-update && cabal v2-install` throws errors on some modules as seen on the [latest ARM builds in the GitHub CI](https://github.com/PostgREST/postgrest/runs/6729509695?check_suite_focus=true#step:8:123). Also, testing on older versions (which already run without problems before) are now giving the same errors.

Looks like Cabal needs explicit casting to avoid the ambiguity, but can't figure out why now.

laurenceisla:
Closes #2301.

Tested a single module with the explicit casting, and it seems to solve the ambiguity when building using Cabal. Maybe a CI for cabal should be added for testing, the ARM build does not run on PRs.


src/PostgREST/Config.hs
```diff
@@ -359,17 +359,17 @@ parser optPath env dbSettings =
 
     coerceInt :: (Read i, Integral i) => C.Value -> Maybe i
     coerceInt (C.Number x) = rightToMaybe $ floatingOrInteger x
-    coerceInt (C.String x) = readMaybe $ toS x
+    coerceInt (C.String x) = readMaybe (toS x :: [Char])
```
robx, line 362: I think we'd typically call this `String` instead of `[Char]`?

robx (commented):
Looks like a fine change to me, and a good fix for the issue.

An alternative (slightly preferable looking forward?) would be a diff like the following:

```
-                    , protolude                 >= 0.3 && < 0.4
+                    , protolude                 >= 0.3.1 && < 0.4
                     , regex-tdfa                >= 1.2.2 && < 1.4
...
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -359,17 +359,17 @@ parser optPath env dbSettings =
 
     coerceInt :: (Read i, Integral i) => C.Value -> Maybe i
     coerceInt (C.Number x) = rightToMaybe $ floatingOrInteger x
-    coerceInt (C.String x) = readMaybe $ toS x
+    coerceInt (C.String x) = readMaybe x
     coerceInt _            = Nothing
 
     coerceBool :: C.Value -> Maybe Bool
     coerceBool (C.Bool b)   = Just b
     coerceBool (C.String s) =
       -- parse all kinds of text: True, true, TRUE, "true", ...
-      case readMaybe . toS $ T.toTitle $ T.filter isAlpha $ toS s of
+      case readMaybe $ T.toTitle $ T.filter isAlpha $ toS s of
         Just b  -> Just b
         -- numeric instead?
-        Nothing -> (> 0) <$> (readMaybe $ toS s :: Maybe Integer)
+        Nothing -> (> 0) <$> (readMaybe s :: Maybe Integer)
     coerceBool _            = Nothing
```

That simplifies the code a bit, but loses backwards-compatibility with old protolude versions. It would require adding an override for protolude to `nix/overlays/haskell-packages.nix`

src/PostgREST/Config.hs
laurenceisla: > An alternative (slightly preferable looking forward?) would be a diff like the following:

Yes, I think that's the way to go. I modified the protolude version and the affected lines.

stack.yaml.lock
```diff
@@ -5,40 +5,54 @@
 
 packages:
 - completed:
-    hackage: hasql-dynamic-statements-0.3.1@sha256:c3a2c89c4a8b3711368dbd33f0ccfe46a493faa7efc2c85d3e354c56a01dfc48,2673
     pantry-tree:
       sha256: b1b9a6a26ec765e5fe29f9a670a5c9ec7067ea00dee8491f0819284ff0201b6f
       size: 641
+    hackage: hasql-dynamic-statements-0.3.1@sha256:c3a2c89c4a8b3711368dbd33f0ccfe46a493faa7efc2c85d3e354c56a01dfc48,2673
```
laurenceisla, line 11: Don't know why Stack modifies the order of the `hackage` line in my environment. Is this normal or should I change the version (I'm using  Stack ` 2.7.5`).

robx: Yeah I got those diffs too, they're a bit annoying, but I wouldn't worry about them.

robx (approved):
This looks good to me.

There's the load-test failure which I've seen myself on #2992. It doesn't seem likely that it would be related to this change. Any idea @steve-chavez, have you seen this before?

steve-chavez:
Not sure why it happens, but I can reproduce the error locally for this branch:

```
[nix-shell:~/Projects/postgrest]$ postgrest-loadtest-against main

Running loadtest on "main"...

Requests      [total, rate, throughput]         19522, 325.35, 325.35
Duration      [total, attack, wait]             1m0s, 1m0s, 1.094ms
Latencies     [min, mean, 50, 90, 95, 99, max]  380.989µs, 3.072ms, 1.126ms, 9.401ms, 20.049ms, 25.89ms, 50.827ms
Bytes In      [total, mean]                     16420679, 841.14
Bytes Out     [total, mean]                     345504, 17.70
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:11714  201:1952  204:5856
Error Set:

Done running on "main".


Running loadtest on HEAD...

2022/06/08 11:26:48 encode: can't detect encoding of "stdin"
```

steve-chavez:
Loadtest succeeded on main https://github.com/PostgREST/postgrest/runs/6798040491?check_suite_focus=true

PR #2309, opened 2022-06-09
nix: more explicit logging in load test (fixes #2308)

### LINKED ISSUE #2308: load tests fail obscurely when nix dependencies change
The `Loadtest (Nix)` CI check compares Postgrest versions on the `main` and PR branches. If the nix dependencies are changed in a PR, this will typically cause build failures of the `main` branch, causing the check to fail.

E.g., in PRs #2306 and #2292, we've seen the `Loadtest (Nix)` CI check failing with

```
Running loadtest on "main"...
2022/06/07 23:54:11 encode: can't detect encoding of "stdin"
Error: Process completed with exit code 1.
```

(e.g. here: https://github.com/PostgREST/postgrest/runs/6784583539?check_suite_focus=true)

- [x] improve the test output so that it's clear what's happening #2309
- [ ] skip the CI check or make it optional when the cabal file changes
- [ ] (stretch goal) fix the load test to work even when dependencies change

robx:
* Write vegeta output to file explicitly.
  The previous version using `tee` captured the standard
  output of the `withPgrst` helper, too, causing weird
  errors when those had unexpected output.

* Log and fail the postgrest build explicitly.
  In situations such as changing nix dependencies, the
  `cabal build` run could fail, but this wasn't visible
  in CI.

With these changes, the load test now fails with a nice
error message when the build fails.


wolfgangwalther:
Yes, that's much better.

As proposed in #2308 I suggest we keep the issue open to run the loadtest conditionally later on. Maybe we just need to add some condition that a failure of the loadtest is fine. We don't have an automated evaluation of the loadtest results so far anyway - so the only failure mode is when the test itself didn't succeed.

wolfgangwalther (approved):

PR #2311, opened 2022-06-12
Bulk update

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/1959.

### TODO

- [x] Tests


test/spec/fixtures/schema.sql
```diff
@@ -2634,3 +2634,26 @@ CREATE TABLE test.test (
...
+);
+
+create function reset_bulk_items(tbl_name text default '') returns void as $_$ begin
+  execute format(
+  $$
+    delete from %I;
+    insert into %I values (1, 'item-1'), (2, 'item-2'), (3, 'item-3');
+  $$::text,
+  tbl_name, tbl_name);
+end; $_$ language plpgsql volatile;
```
laurenceisla, line 2659: This is repeating the code from the `reset_limited_items` RPC, maybe rename the first one to something like `reset_items_tables` and use one function for both cases (limited and bulk)?

steve-chavez: Yeah, that sounds good

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -388,6 +388,17 @@ spec = do
...
+      request methodPatch "/limited_update_items"
+          [("Prefer", "tx=commit"), ("Prefer", "count=exact")]
+          [json| {"name": "updated-item"} |]
+        `shouldRespondWith`
+          ""
+          { matchStatus  = 404
+          , matchHeaders = [ matchHeaderAbsent hContentType
+                           , "Content-Range" <:> "*/0" ]
+          }
+
```
steve-chavez, line 401: This error was just a lucky side-effect of https://github.com/PostgREST/postgrest/issues/2343. This needs a clear error message plus a 422 status - like done for [batch delete](https://github.com/PostgREST/postgrest/pull/2355#issuecomment-1176973562).

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -607,8 +619,186 @@ spec = do
...
+            { "id": 1, "name": "item-1 - 1st", "observation": "Lost item" }
+          , { "id": 3, "name": "item-3 - 3rd", "observation": null }
+          ]|]
+        `shouldRespondWith`
+          ""
+          { matchStatus  = 204
+          , matchHeaders = [ matchHeaderAbsent hContentType
+                           , "Content-Range" <:> "0-1/*"
+                           , "Preference-Applied" <:> "tx=commit" ]
+          }
```
steve-chavez, line 650: Using the pks as filters without the option to reference other key in the body is too arbitrary. We need to be able to refer to every key, like [bulk delete](https://github.com/PostgREST/postgrest/pull/2355#issue-1291906948) does.

For this, we'd need [underscore operators](https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1200054135).

PR #2324, opened 2022-06-14
Update GHC to 9.2.2

robx:
- relax cabal bounds to allow compiling with GHC
- some refactoring to fix changed compiler warnings
- nix:
  - change compiler to GHC 9.2.2
  - update hsie for GHC 9.2.2
  - add new versions of weeder and ptr to overlay


steve-chavez (approved):
The required separation into the new ReadQuery/MutateQuery modules looks good!

Switching to GHC 9.2 LGTM!

PR #2332, opened 2022-06-16
experiment: try using nix's pkgs.pkgsMusl for the static build

robx:


robx:
After rebasing, this is currently failing with:

```
       > /nix/store/gzms61swp55fg5qbvshyqv5jfsnfvybz-x86_64-unknown-linux-musl-binutils-2.38/bin/x86_64-unknown-linux-musl-ld: /nix/store/pcdy4i08qq7w4ipxa3xan6rqdpdvh8hs-icu4c-71.1-x86_64-unknown-linux-musl/lib/libicuuc.a(parsepos.ao):(.data.rel.ro._ZTIN6icu_7113ParsePositionE[_ZTIN6icu_7113ParsePositionE]+0x0): undefined reference to `vtable for __cxxabiv1::__si_class_type_info'
       > collect2: error: ld returned 1 exit status
       > make[2]: *** [Makefile:66: postgres] Error 1
       > make[2]: Leaving directory '/build/postgresql-14.4/src/backend'
       > make[1]: *** [Makefile:42: all-backend-recurse] Error 2
       > make[1]: Leaving directory '/build/postgresql-14.4/src'
       > make: *** [GNUmakefile:16: world-src-recurse] Error 2
       For full logs, run 'nix log /nix/store/wb0p273zxsa5f1wspn4173l7wxz9kr95-postgresql-static-x86_64-unknown-linux-musl-14.4.drv'.
```

This appears to be an issue linking `libicuuc`, while building postgresql itself statically.

We shouldn't actually need to build the postgresql backend statically, this is just happening because we need static `libpq` (but chances are similar issues would show up anyway).

brandonros:
What would it take to add cross-compilation support to this as per https://nix.dev/tutorials/cross-compilation?

robx:
> What would it take to add cross-compilation support to this as per https://nix.dev/tutorials/cross-compilation?

I'm not sure I understand what you're going for here, could you elaborate?

(The issue https://github.com/PostgREST/postgrest/issues/2478 collects some more info about the static build, might be the better place to discuss this.)

steve-chavez:
> We shouldn't actually need to build the postgresql backend statically, this is just happening because we need static libpq (but chances are similar issues would show up anyway).

There's some progress for packaging only libpq on nixpkgs: https://github.com/NixOS/nixpkgs/pull/234470/files

Perhaps we could copy that expression into our repo?

PR #2333, opened 2022-06-16
experiment: use a lazy bytestring hasql 'unknown'

robx:
compare #2261

PR #2336, opened 2022-06-17
nix: update bytestring-strict-builder to 0.4.5.6

robx:
This comes with a rewrite rule that hopefully allows us to avoid
a copy of the request body.

### DIFF
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -89,6 +89,15 @@ let
           }
           { });
 
+      bytestring-strict-builder =
+        lib.dontCheck (prev.callHackageDirect
+          {
+            pkg = "bytestring-strict-builder";
+            ver = "0.4.5.6";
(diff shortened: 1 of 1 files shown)


steve-chavez:
Hm, would this change be mutually exclusive with the change in https://github.com/PostgREST/postgrest/pull/2333? That one seemed to give better results.

robx:
The rule doesn't trigger, yet, unfortunately. If it worked, it would be equivalent as a first step, and allow us to avoid a second copy if we manage to read the body into a strict bytestring directly.

The other advantage of this approach is that it's something that doesn't require potentially unwanted upstream changes, compare discussion in https://github.com/nikita-volkov/hasql/pull/146 and https://github.com/nikita-volkov/bytestring-strict-builder/pull/11.

But if this doesn't work out quickly, we can definitely go with the forked version of hasql (#2333) momentarily and follow up after.

robx:
The rewrite rule turned out not to trigger for us, closing this.

PR #2345, opened 2022-06-22
Allow uppercase, spaces and other special characters in schema names

### LINKED ISSUE #2341: Computed columns not working anymore when the SCHEMA name contains uppercase characters
### Environment

* PostgreSQL version: 12.10 (Debian 12.10-1.pgdg100+1)
* PostgREST version: both 8.0 and 9.0.1 are impacted (linux-static-x64)
* Operating system: Debian Buster (10.12)

### Description of issue

Trying to use a computed column: `/my_film?select=*,thumbnail` (where thumbnail(my_film) is a computed column)

When the SCHEMA name contains any uppercase characters, PostgREST will raise an error:

 `column my_film.thumbnail does not exist`

However if I rename the SCHEMA name in order to remove any uppercase characters, it then works normally.

Note this is a regression bug: I'm using PostgREST version 7 in production with a SCHEMA name "REST_API_v1_0" and it is working properly. Only when I try to upgrade to PostgREST 8 or 9 then the problem arises.

### Steps to reproduce

```sql
CREATE SCHEMA "REST_api_v2_0" AUTHORIZATION anonymous;

CREATE TABLE "REST_api_v2_0".film  (
    film_id integer NOT NULL,
    title character varying(255) NOT NULL
);

CREATE VIEW "REST_api_v2_0".my_film AS
    SELECT film.film_id AS id,
    film.title AS titre
   FROM "REST_api_v2_0".film
  WHERE film_id<100;

GRANT SELECT ON TABLE "REST_api_v2_0".film TO public;
GRANT SELECT ON TABLE "REST_api_v2_0".my_film TO public;

CREATE FUNCTION "REST_api_v2_0".thumbnail(c "REST_api_v2_0".my_film) RETURNS text AS $$
BEGIN
  RETURN 'REST_API_v2_0 bar';
END;
$$ LANGUAGE plpgsql IMMUTABLE    RETURNS NULL ON NULL INPUT   PARALLEL SAFE;

GRANT EXECUTE ON FUNCTION "REST_api_v2_0".thumbnail TO public;
```

=> Trying `/my_film?select=*,thumbnail` won't work (error "column does not exist").

Now execute the following SQL line:

```sql
ALTER SCHEMA "REST_api_v2_0" RENAME TO "rest_api_v2_0";
```

(And don't forget to update your `db-schema` param in the `postgrest.conf` file)

=> Now it works properly (no more error)

laurenceisla:
Closes #2341

- [x] Allow uppercase/spaces/special characters in schema names
- [x] Add spec tests
- [x] Add unit tests

### DIFF
--- a/src/PostgREST/Middleware.hs
+++ b/src/PostgREST/Middleware.hs
@@ -14,7 +14,6 @@ import qualified Data.Aeson.Key                    as K
 import qualified Data.Aeson.KeyMap                 as KM
 import qualified Data.ByteString.Lazy.Char8        as LBS
 import qualified Data.HashMap.Strict               as HM
-import qualified Data.Text                         as T
 import qualified Data.Text.Encoding                as T
 import qualified Hasql.Decoders                    as HD
 import qualified Hasql.DynamicStatements.Snippet   as SQL hiding (sql)
@@ -32,7 +31,7 @@ import PostgREST.Config.PgVersion   (PgVersion (..), pgVersion140)
 import PostgREST.Error              (Error, errorResponseFor)
 import PostgREST.GucHeader          (addHeadersIfNotIncluded)
 import PostgREST.Query.SqlFragment  (fromQi, intercalateSnippet,
-                                     unknownEncoder)
+                                     pgFmtIdentList, unknownEncoder)
 import PostgREST.Request.ApiRequest (ApiRequest (..), Target (..))
 
 import PostgREST.Request.Preferences
@@ -64,8 +63,8 @@ runPgLocals conf claims role app req jsonDbS actualPgVersion = do
     roleSql = [setConfigLocal mempty ("role", toUtf8 role)]
     appSettingsSql = setConfigLocal mempty <$> (join bimap toUtf8 <$> configAppSettings conf)
     searchPathSql =
-      let schemas = "\"" <> T.intercalate "\", \"" (iSchema req : configDbExtraSearchPath conf) <> "\"" in
-      setConfigLocal mempty ("search_path", toUtf8 schemas)
+      let schemas = pgFmtIdentList (iSchema req : configDbExtraSearchPath conf) in
+      setConfigLocal mempty ("search_path", schemas)
     preReqSql = (\f -> "select " <> fromQi f <> "();") <$> configDbPreRequest conf
     specSql = case iTarget req of
       TargetProc{tpIsRootSpec=True} -> [setConfigLocal mempty ("request.spec", jsonDbS)]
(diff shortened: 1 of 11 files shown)


src/PostgREST/Middleware.hs
```diff
@@ -64,7 +64,7 @@ runPgLocals conf claims role app req jsonDbS actualPgVersion = do
     roleSql = [setConfigLocal mempty ("role", toUtf8 role)]
     appSettingsSql = setConfigLocal mempty <$> (join bimap toUtf8 <$> configAppSettings conf)
     searchPathSql =
-      let schemas = T.intercalate ", " (iSchema req : configDbExtraSearchPath conf) in
+      let schemas = "\"" <> T.intercalate "\", \"" (iSchema req : configDbExtraSearchPath conf) <> "\"" in
```
steve-chavez, line 67: This a great fix @laurenceisla, really nice tests!

WDYT about having a separate function for this logic, it would be similar to `pgFmtIdent`

https://github.com/PostgREST/postgrest/blob/d87b622d64d993a240be9a28b7e6f515a489c161/src/PostgREST/Query/SqlFragment.hs#L152-L153

laurenceisla: > WDYT about having a separate function for this logic, it would be similar to pgFmtIdent

Thanks to this I noticed that double quotes `"` still weren't being accepted. I used the `pgFmtIdent` function to build the list now.

test/spec/Feature/Query/MultipleSchemaSpec.hs
```diff
@@ -111,7 +122,7 @@ spec =
         request methodPost "/children" [("Content-Profile", "unknown")]
           [json|{"name": "child 4", "parent_id": 4}|]
           `shouldRespondWith`
-          [json|{"message":"The schema must be one of the following: v1, v2","code":"PGRST106","details":null,"hint":null}|]
+          [json|{"message":"The schema must be one of the following: v1, v2, SPECIAL \"@/\\#~_-","code":"PGRST106","details":null,"hint":null}|]
```
steve-chavez, line 125: Oh, how would it look here if the `SPECIAL \"@/\\#~_-` schema had a comma(`,`)?

laurenceisla: I added the comma to the tests and worked without issues. But I think the problem is that the configuration file does not accept the comma in the value, because it uses it as a splitter character: https://github.com/PostgREST/postgrest/blob/870f7a39b077269faebdc32006e7c41acff79b4a/src/PostgREST/Config.hs#L375-L377
Or am I missing a way to send commas as part of the schema name?

steve-chavez (approved):

PR #2351, opened 2022-06-24
Return Location header only when using Prefer: return=header-only

### LINKED ISSUE #2312: Unnecessary computing of Location header on `Prefer: return=representation`
When doing:

```http
POST /table_name HTTP/1.1
Prefer: return=representation

{ "col1": "value1", "col2": "value2" }
```

[return=headers-only](https://postgrest.org/en/stable/api.html#insertions-updates) is assumed, so the computing of the Location header happens anyway.

This is wasteful considering the representation is already obtained, the user doesn't need to make another request and visit the link in the Location header.

The generated query is:

```sql
WITH pgrst_source AS (
  WITH
  pgrst_payload AS (SELECT $1::json AS json_data),
  INSERT INTO "test"."projects"("id", "name")
  SELECT "id", "name"
  FROM json_populate_recordset (null::"test"."projects", (SELECT val FROM pgrst_body)) _
  RETURNING "test"."projects".*)
SELECT
  '' AS total_result_set,
  CASE WHEN pg_catalog.count(_postgrest_t) = 1
    THEN coalesce((
          WITH data AS (SELECT row_to_json(_) AS row FROM pgrst_source AS _ LIMIT 1)
          SELECT array_agg(json_data.key || '=' || coalesce('eq.' || json_data.value, 'is.null'))
          FROM data CROSS JOIN json_each_text(data.row) AS json_data
          WHERE json_data.key IN ('id')
        ), array[]::text[])
    ELSE array[]::text[] END AS header,
FROM (SELECT "projects".*FROM "pgrst_source" AS "projects"    ) _postgrest_t;
```

If we only obtain the Location header on `return=headers-only`, we can omit the `coalesce(( .. CROSS JOIN` part and gain some perf.

laurenceisla:
Closes #2312

### DIFF
--- a/test/spec/Feature/Query/InsertSpec.hs
+++ b/test/spec/Feature/Query/InsertSpec.hs
@@ -96,7 +96,7 @@ spec actualPgVersion = do
           [json|{"id":6,"name":"New Project","client_id":2}|] `shouldRespondWith` [json|[{"id":6,"name":"New Project","clients":{"id":2,"name":"Apple"}}]|]
           { matchStatus  = 201
           , matchHeaders = [ matchContentTypeJson
-                           , "Location" <:> "/projects?id=eq.6"
+                           , matchHeaderAbsent hLocation
                            , "Content-Range" <:> "*/1" ]
           }
 
@@ -107,7 +107,7 @@ spec actualPgVersion = do
           [json|[{"pId":"7","pName":"New Project","cId":"2"}]|]
           { matchStatus  = 201
           , matchHeaders = [ matchContentTypeJson
-                           , "Location" <:> "/projects?id=eq.7"
+                           , matchHeaderAbsent hLocation
                            , "Content-Range" <:> "*/*" ]
           }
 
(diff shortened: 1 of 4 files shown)


test/spec/Feature/Query/InsertSpec.hs
```diff
@@ -489,13 +489,21 @@ spec actualPgVersion = do
...
 
-    context "with unicode values" $
+    context "with unicode values" $ do
+      it "succeeds and returns full representation" $
+        request methodPost "/simple_pk2?select=extra,k"
+            [("Prefer", "return=representation")]
+            [json| { "k":"圍棋", "extra":"￥" } |]
+        `shouldRespondWith`
+          [json|[ { "k":"圍棋", "extra":"￥" } ]|]
+          { matchStatus = 201 }
```
laurenceisla, line 499: Had to separate this test into two: Test returning unicode characters in the body with `return=representation` and in the Location header with `return=headers-only`.

Also, I noticed this is under the `describe "CSV insert"` group of tests, shouldn't the payload be in CSV format?

steve-chavez: Yeah, the payload can be changed or the test should be moved.

laurenceisla: There's a similar test for Updates under "Patching record" that was added at the same time: 

https://github.com/PostgREST/postgrest/blob/f7745e1569f47dc02cfd0e1e526b8be42293db6c/test/spec/Feature/Query/UpdateSpec.hs#L283-L289

So I moved this one under the "Posting new record" group for consistency.

steve-chavez (approved):

Issue #2352, opened 2022-06-28
Send a `499 Client Closed Request` status when the proxy times out a request


steve-chavez:
Right now we don't log a canceled request.

Needs confirmation.

steve-chavez:
Yeah, this can be reproduced without a proxy too:

```sql
-- Append on test/spec/fixtures/schema.sql
CREATE FUNCTION test.sleep() RETURNS void AS $$
  SELECT pg_sleep(5);
  INSERT INTO test.projects VALUES (111, 'A canceled request');
$$ LANGUAGE sql;
```
```bash 
# In one terminal
PGRST_DB_ANON_ROLE='postgrest_test_anonymous' postgrest-with-postgresql-14 postgrest-run

## On another terminal
http POST localhost:3000/rpc/sleep

## cancel the request immediately
## Then after 5 seconds
http localhost:3000/projects

...
{
    "client_id": null,
    "id": 111,
    "name": "A canceled request"
}
]
```

Related to https://github.com/PostgREST/postgrest/issues/699

steve-chavez:
On #699, it was proposed: 

> I think we could use [setOnClose](https://hackage.haskell.org/package/warp-3.2.8/docs/Network-Wai-Handler-Warp.html#v:setOnClose) to detect this situation and roll back the current transaction.

But a `rollback` wouldn't cancel a query immediately, we need to [cancel the query using libpq](https://www.postgresql.org/docs/current/libpq-cancel.html) or by calling `pg_cancel_backend`(not sure if possible/easy).

Cancelling the query should be possible because on `psql`:

```sql
select pg_sleep(300);
-- Pressing Ctrl + C
Cancel request sent
ERROR:  57014: canceling statement due to user request
LOCATION:  ProcessInterrupts, postgres.c:3329
```

steve-chavez:
Cancel support in other libraries:

- psycopg2: https://stackoverflow.com/questions/2041978/cancel-query-execution-in-pyscopg2 (doesn't seem possible)
- java jdbc: https://stackoverflow.com/questions/1526211/how-to-cancel-a-postgres-query-in-java-jdbc (possible?)
- How to do it with libpq in C: https://stackoverflow.com/questions/18015203/postgresql-cancel-query-from-c-c-program (seems complicated)
- Deno: https://deno.land/x/postgresjs@v3.2.4#canceling-queries-in-progress (unreliable)

steve-chavez:
[...]

PR #2355, opened 2022-07-01
Add support for bulk-delete

### LINKED ISSUE #2314: Bulk DELETE, restrict full table DELETE without filters
The DELETE method now has two problems:

- It allows a full table DELETE without filters - an expensive operation(can take a long time on big tables) and it can occur by accident
- A DELETE with a big in query parameter(`?col=in.(1,2,3,4,5,....)`) can err because of URL size limitations.

We can have the same interface as the new [Bulk UPDATE](https://github.com/PostgREST/postgrest/issues/1959#issuecomment-1152813181) for DELETE and solve the above problems.

This would involve having a body for a DELETE method, which seems fine since [other well-known APIs do it](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-scroll.html#_clear_scroll_api). There was also a previous discussion in https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1046306418.

laurenceisla:
Closes #2314

- [ ] Full support for bulk-delete
- [ ] Tests
- [ ] Changelog


src/PostgREST/App.hs
```diff
@@ -203,7 +205,7 @@ postgrestResponse conf@AppConfig{..} maybeDbStructure jsonDbS pgVer pool AuthRes
 
   apiRequest <-
     liftEither . mapLeft Error.ApiRequestError $
-      ApiRequest.userApiRequest conf dbStructure req body
+      ApiRequest.userApiRequest conf dbStructure req (BUG.traceShowId body)
```
laurenceisla, line 208: Need help here. Debugging this line shows that the body is empty, does the library ignore it for DELETEs?

steve-chavez: That looks like [Lazy evaluation](https://wiki.haskell.org/Lazy_evaluation). There will be no trace since the body variable is unused.

You need to add the MutateDelete case here for the body to be passed:

https://github.com/PostgREST/postgrest/blob/fa8dee2f90e82c28f3d390945a275b3025a4cbbb/src/PostgREST/Request/ApiRequest.hs#L312-L318

laurenceisla: Oops, right. That fixed it.

steve-chavez:
I think we can close https://github.com/PostgREST/postgrest/issues/699 once this is merged(and continue the cancel request part on https://github.com/PostgREST/postgrest/issues/2352)

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -151,13 +151,25 @@ mutateRequestToQuery (Update mainQi uCols body logicForest pkFlts range ordts re
     rangeCols = BS.intercalate ", " ((\col -> pgFmtIdent col <> " = (SELECT " <> pgFmtIdent col <> " FROM pgrst_update_body) ") <$> S.toList uCols)
     (whereRangeIdF, rangeIdF) = mutRangeF mainQi (fst . otTerm <$> ordts)
 
-mutateRequestToQuery (Delete mainQi logicForest range ordts returnings)
+mutateRequestToQuery (Delete mainQi body logicForest pkFlts range ordts returnings)
+  -- The body is not mandatory, although a delete without filters is not possible.
   | range == allRange =
+    let whereLogic | null logicForest = if hasEmptyBody || null pkFlts then "FALSE" else pgrstDeleteBodyF
+                   | otherwise        = (if hasEmptyBody || null pkFlts then mempty else pgrstDeleteBodyF <> "AND") <> logicForestF in
```
steve-chavez, line 158:
```suggestion
                   | otherwise        = (if hasEmptyBody || null pkFlts then mempty else pgrstDeleteBodyF <> " AND ") <> logicForestF in
```

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -90,7 +243,7 @@ spec =
 
       it "succeeds deleting the row when return=representation and selecting only the privileged columns" $
         request methodDelete "/app_users?id=eq.1&select=id,email" [("Prefer", "return=representation")]
-          [json| { "password": "passxyz" } |]
+          mempty
```
steve-chavez, line 246: What happens when there are filters applied and a request body too?

steve-chavez:
As it is, one deficiency of the new bulk delete(and [bulk update](https://github.com/PostgREST/postgrest/pull/2311) suffers from something similar too) is that it doesn't fail with an appropriate message when no filters or body are specified, it just returns 204. This is desirable as discussed on the issues, but it is surprising for someone that hasn't read our docs.

Do you think it would be possible to respond with a [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) when no filters, limits or body are present?

test/spec/Feature/Query/DeleteSpec.hs
laurenceisla: Yeah, I did not add a test for that. It treats the request body as the main source for PKs and adds the query string as filters (works like an intersection of PKs between the body and the query string). The WHERE clause for the delete looks something like this:

```sql 
...
WHERE app_users.id IN <request_body>.id AND id = <query_string.id> AND <other_filters>
```

Before changing this test, there were no ids in the body (only a password column) so it didn't delete anything.

steve-chavez: I see. Hmm, for this case I recall the bulk update did not consider the body PKs as filters. Would it be possible to do this for DELETE as well? I think we should have the same behavior for both bulk update and delete.

steve-chavez: > Now, to keep consistency, as soon as any filter (not only PKs) exist in the query string when deleting, then the body should be ignored (there's no data to be used for updating here), correct?

Yes, exactly. That would be consistent with PATCH.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -171,7 +183,9 @@ mutateRequestToQuery (Delete mainQi logicForest range ordts returnings)
     SQL.sql (returningF mainQi returnings)
 
   where
-    whereLogic = if null logicForest then mempty else " WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree mainQi <$> logicForest)
+    hasEmptyBody = body == mempty
+    logicForestF = intercalateSnippet " AND " (pgFmtLogicTree mainQi <$> logicForest)
+    pgrstDeleteBodyF = SQL.sql (BS.intercalate " AND " $ (\x -> pgFmtColumn mainQi x <> " = " <> pgFmtColumn (QualifiedIdentifier mempty "pgrst_delete_body") x) <$> pkFlts)
```
laurenceisla, line 188: I'm not sure about the use of `columns` in the query string with DELETE.  For POST and PATCH it's used as a way to select data from the body to be inserted and updated (in the SET clause). But for DELETE it would be used to select PKs from the body to be used in the WHERE clause, which is redundant since it already filters using the PKs (as seen in this line). I can see these options:
- Do not use `columns` (but it won't have an option to use the body "as is", without JSON parsing)
- Use it and filter only PKs (will need a warning "Only PKs allowed" if other columns are selected)
- Could be used as a way to add more filters in the body.

src/PostgREST/Request/DbRequestBuilder.hs
```diff
@@ -342,6 +350,9 @@ mutateRequest mutation schema tName ApiRequest{..} pkCols readReq = mapLeft ApiR
     rootOrder = maybe [] snd $ find (\(x, _) -> null x) qsOrder
     combinedLogic = foldr addFilterToLogicForest logic qsFiltersRoot
     body = payRaw <$> iPayload -- the body is assumed to be json at this stage(ApiRequest validates)
+    bodyHasPks = case iPayload of
+      Just ProcessedJSON{payKeys} -> all (`elem` payKeys) pkCols
+      _                           -> True
```
laurenceisla, line 355: Was looking for a workaround for this. It looks like the `_ -> True` part is never executed for Deletes (because no `columns` are implemented). But this will be used to add the `422` response on missing/empty bodies for `PATCH`, so I think it could be left as is right now.

src/PostgREST/Query/QueryBuilder.hs
steve-chavez: > Could be used as a way to add more filters in the body.

Hm, this would be useful but it would be inconsistent with bulk PATCH since `columns` couldn't do the same there.

> Use it and filter only PKs (will need a warning "Only PKs allowed" if other columns are selected)

I think this would be the best way for now. Also leaves room for an enhancement later without a breaking change.

src/PostgREST/Request/DbRequestBuilder.hs
laurenceisla: Solved this in the last commit, no more Codecov warning.

src/PostgREST/Query/QueryBuilder.hs
laurenceisla: After deliberation on another platform, we agreed on using the third option for DELETE:

> Could be used as a way to add more filters in the body

Works for single and bulk deletes.

src/PostgREST/Request/DbRequestBuilder.hs
```diff
@@ -342,6 +345,7 @@ mutateRequest mutation schema tName ApiRequest{..} pkCols readReq = mapLeft ApiR
     rootOrder = maybe [] snd $ find (\(x, _) -> null x) qsOrder
     combinedLogic = foldr addFilterToLogicForest logic qsFiltersRoot
     body = payRaw <$> iPayload -- the body is assumed to be json at this stage(ApiRequest validates)
+    noBodyFilterLimit = null qsColumns && null qsLogic && null qsFilters && null qsRanges && (isNothing body || S.null iColumns)
```
steve-chavez, line 348: This check would be better at the `ApiRequest` level since it doesn't depend on the db. This way we avoid taking a pool connection and opening an empty transaction when this error happens.

steve-chavez:
Hm, I've been thinking that this one could also be done with [underscore operators](https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1200054135). Like:

```http
DELETE /tbl?id=_in.body->ids

{ "ids": [1,2,3,4]}
```

I guess one problem with that approach would be how to compare two values that are on the same json object, which this PR already does. For underscore operators, I'm not sure if this would be possible:

```http
DELETE /tbl?col1=_eq.body->*->col1&col2=_eq.body->*->col2


[
  {"col1": "val1", "col2": "val2"}
, {"col1": "val3", "col2": "val4"}
]
```

The `*` is borrowed from [JSON path's](https://www.postgresql.org/docs/12/functions-json.html#FUNCTIONS-SQLJSON-PATH) `[*]` which means get keys from all array elements.

laurence-isla:
Closed in favor of https://github.com/PostgREST/postgrest/pull/2724

Issue #2356, opened 2022-07-05
OpenAPI is missing "This is a Primary key.<pk/>" description for views


wolfgangwalther:
This is happening on the latest pre-release and was introduced in d2719420f46a537a75d76a4666f8bffec582d4f8 / #2254.

PK columns have a description field in the OpenAPI output like this: `This is a Primary key.<pk/>`. This is still present for tables as tested here:

https://github.com/PostgREST/postgrest/blob/3e83bef9c4b1f105ad0f261d27f8c16eb9377776/test/spec/Feature/OpenApi/OpenApiSpec.hs#L125-L158

However after the above mentioned change this is now missing for views. The test-case demonstrated in ee906da99cd918e4940f9cc3d4570e526f71b851 is now failing but was passing before.

This breaks client libraries, which depend on `<pk/>`. For me that's https://github.com/technowledgy/vue-postgrest - and it blocks me from updating to the latest pre-release.

PR #2357, opened 2022-07-05
test: Add test for openapi output of views

wolfgangwalther:
This demonstrates a breaking change in the latest pre-release by adding a new test for openapi output of views.

Related to #2356.


steve-chavez:
Another alternative that seems simpler. On `accessibleTables/Procs` we don't query all the metadata again but just the list of tables/procs that are accessible by the role, then we filter the tables/procs from the cache with that list in Haskell code.

wolfgangwalther:
> @wolfgangwalther As a workaround, if you use `PGRST_OPENAPI_MODE="ignore-privileges"` then the PK will be detected.

I tried that and it seems like it works... but not quite, yet. It looks like PKs are not detected through multiple views recursively anymore. So one view selects from another view which selects a PK from a table.

Does this make sense? Is this expected, because of changes to embedding? Or should I try to put together a test-case?

wolfgangwalther:
> > @wolfgangwalther As a workaround, if you use `PGRST_OPENAPI_MODE="ignore-privileges"` then the PK will be detected.
> 
> I tried that and it seems like it works... but not quite, yet. It looks like PKs are not detected through multiple views recursively anymore. So one view selects from another view which selects a PK from a table.
> 
> Does this make sense? Is this expected, because of changes to embedding? Or should I try to put together a test-case?

Hm. The few ones I'm still missing might also be related to multi column PKs. Will have to put a test-case together, eventually, I guess.

steve-chavez:
> Does this make sense? Is this expected, because of changes to embedding? Or should I try to put together a test-case?

Hm, no. PK detection on views recursively should work. I didn't expect to break that. A test case would be great, yes.

steve-chavez:
Approach looks good! Could you make the change independent of #2526? I still don't understand that one and it might take me more time but 6b5f916fb145d7be05d8d566ae6a75fccd6a6bf5 can be merged

PR #2363, opened 2022-07-08
Add security options for OpenApi

### LINKED ISSUE #1082: Add securityDefinitions to openapi/swagger spec

laurenceisla:
Closes #1082.

- [x] Support security options
- [x] Refactor code
- [x] Add configuration option
- [x] Add changelog


laurenceisla:
Tested for Swagger and the options `securityDefinitions` and `security` are enough for the `Authorize` button to appear and handle authorized requests. Although this applies to all the endpoints, I don't see an easy way for now to differentiate between an endpoint that needs authorization and one that doesn't.

steve-chavez:
> I don't see an easy way for now to differentiate between an endpoint that needs authorization and one that doesn't.

Hm, let's just make it configurable for now.

```
openapi-security-definitions = ".."
```

disabled by default.

test/io/configs/expected/no-defaults-with-db-other-authenticator.config
```diff
@@ -19,6 +19,7 @@ jwt-secret = "ODERREALLYREALLYREALLYREALLYVERYSAFE"
 jwt-secret-is-base64 = true
 log-level = "info"
 openapi-mode = "disabled"
+openapi-security-active = true
```
steve-chavez, line 22: Hm, I think this one required changing the in-db config(like below). Would be good to test this too.

https://github.com/PostgREST/postgrest/blob/bdf1cbe11126bae043b754009a1a7fe19f9c7e9f/test/io/db_config.sql#L56-L57

steve-chavez (approved):

steve-chavez:
Hm, somehow a test failed on main branch https://github.com/PostgREST/postgrest/runs/7333244906?check_suite_focus=true

laurenceisla:
> Hm, somehow a test failed on main branch https://github.com/PostgREST/postgrest/runs/7333244906?check_suite_focus=true

It passed now.

Issue #2354, opened 2022-06-28
Get the EXPLAIN output of a request


steve-chavez:
Sometimes it's difficult to understand why a simple request like `/tbl?id=eq.1` can be slow while it's equivalent in SQL is fast. This is usually due to a bad RLS policy that only applies to a postgREST role.

We could support an `Accept: application/vnd.pgrst.explain+json` media type that gets the `EXPLAIN (FORMAT JSON)` output into a response. That would clarify that the query is expensive and is not because of us doing some extra processing.

An `EXPLAIN ANALYZE` could also be supported with `Accept: application/vnd.pgrst.explain-analyze+json`, coupled with `Prefer: tx=rollback` it would be safe to do. We might also consider applying `tx=rollback` by default for it.

---

References: 
- https://www.postgresql.org/docs/current/sql-explain.html
- [Previous discussion of using the http TRACE method for debugging](https://github.com/PostgREST/postgrest/issues/1573#issuecomment-677899436)
- [Previous discussion of using `Accept: application/sql` to obtain the generated query](https://github.com/PostgREST/postgrest/issues/1573#issuecomment-677865905)

steve-chavez:
Actually `EXPLAIN (FORMAT JSON) ANALYZE` is not allowed in pg(gives syntax error). So we only need a single media type.

Searching [IANA types](https://www.iana.org/assignments/media-types/media-types.xhtml), I found some related to cost(`alto-endpointcost+json`, `alto-costmap+json`, etc) but those doesn't map to our use case.

So we should go with our own vendor media type: `application/vnd.pgrst.plan`. Taking advantage of the different formats in EXPLAIN, we can do `application/vnd.pgrst.plan+{text,xml,json,yaml}`.

steve-chavez:
> [Previous discussion of using the http TRACE method for debugging](https://github.com/PostgREST/postgrest/issues/1573#issuecomment-677899436)

Mentioned before, but repeating it here: 

>  [A client MUST NOT send a message body in a TRACE request.](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.8)

So unfortunately that leaves TRACE out. Still, because of the different formats supported, the media type seems like the right interface.

steve-chavez:
Getting the [pg 14 Query ID](https://blog.rustprooflabs.com/2021/10/postgres-14-query-id) would be good for debugging too. That way a request can be easily mapped to a query in the logs.

[...]

PR #2368, opened 2022-07-14
Add media type for explain

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/2354

### TODO

- [x] Plan for reads
- [x] Parse options(SUMMARY, TIMING, etc)
- [x] Plan for rpc
- [x] Plan for writes
- [x] Refactor
- [x] Tests
  + [x] reads with all options
  + [x] writes
  + [x] rpc
- [x] Add config for enabling EXPLAIN
- [x] Allow text and json output
- [x] CHANGELOG


steve-chavez:
Some observations.

### The query identifier is included with the VERBOSE option

<details>
<summary>curl 'localhost:3000/projects?id=lt.5' -H "Accept: application/vnd.pgrst.plan; options=<b>verbose</b>"</summary>
<pre>
[
  {
    "Plan": {
      "Node Type": "Aggregate",
      "Strategy": "Plain",
      "Partial Mode": "Simple",
      "Parallel Aware": false,
      "Async Capable": false,
      "Startup Cost": 24.25,
      "Total Cost": 24.28,
      "Plan Rows": 1,
      "Plan Width": 144,
      "Output": ["NULL::bigint", "count(ROW(projects.id, projects.name, projects.client_id))", "'{}'::text[]", "(COALESCE(json_agg(ROW(projects.id, projects.name, projects.client_id)), '[]'::j
son))::character varying", "NULLIF(current_setting('response.headers'::text, true), ''::text)", "NULLIF(current_setting('response.status'::text, true), ''::text)"],
      "Plans": [
        {
          "Node Type": "Bitmap Heap Scan",
          "Parent Relationship": "Outer",
          "Parallel Aware": false,
          "Async Capable": false,
          "Relation Name": "projects",
          "Schema": "test",
          "Alias": "projects",
          "Startup Cost": 7.25,
          "Total Cost": 22.25,
          "Plan Rows": 400,
          "Plan Width": 40,
          "Output": ["projects.id", "projects.name", "projects.client_id"],
          "Recheck Cond": "(projects.id < 5)",
          "Plans": [
            {
              "Node Type": "Bitmap Index Scan",
              "Parent Relationship": "Outer",
              "Parallel Aware": false,
              "Async Capable": false,
              "Index Name": "projects_pkey",
              "Startup Cost": 0.00,
              "Total Cost": 7.15,
              "Plan Rows": 400,
              "Plan Width": 0,
              "Index Cond": "(projects.id < 5)"
            }
          ]
        }
      ]
    },
    "Query Identifier": 8960408738112779943
  }
]
</pre>
</details>

Somehow the query identifier doesn't change unless the filter is changed. So doing an `lt.3` or `lt.4` give the same identifier, `eq.1` or `eq.5` give a different identifier. This is more or less explained in the [pg_stat_statements docs](https://www.postgresql.org/docs/current/pgstatstatements.html):

> When a constant's value has been ignored for purposes of matching the query to other queries, the constant is replaced by a parameter symbol, such as $1

> However, it is important to understand that there are only limited guarantees around the stability of the queryid hash value. Since the identifier is derived from the post-parse-analysis tree, its value is a function of, among other things, the internal object identifiers appearing in this representation.

So looks more like a plan identifier than a query identifier. Not really reliable to trace a request back to a query.

### The search_path is included with the SETTINGS option

<details>
<summary>curl 'localhost:3000/projects?id=lt.5' -H "Accept: application/vnd.pgrst.plan; options=<b>settings</b>"</summary>
<pre>
[
  {
    "Plan": {
      "Node Type": "Aggregate",
      "Strategy": "Plain",
      "Partial Mode": "Simple",
      "Parallel Aware": false,
      "Async Capable": false,
      "Startup Cost": 24.25,
      "Total Cost": 24.28,
      "Plan Rows": 1,
      "Plan Width": 144,
      "Plans": [
        {
          "Node Type": "Bitmap Heap Scan",
          "Parent Relationship": "Outer",
          "Parallel Aware": false,
          "Async Capable": false,
          "Relation Name": "projects",
          "Alias": "projects",
          "Startup Cost": 7.25,
          "Total Cost": 22.25,
          "Plan Rows": 400,
          "Plan Width": 40,
          "Recheck Cond": "(id < 5)",
          "Plans": [
            {
              "Node Type": "Bitmap Index Scan",
              "Parent Relationship": "Outer",
              "Parallel Aware": false,
              "Async Capable": false,
              "Index Name": "projects_pkey",
              "Startup Cost": 0.00,
              "Total Cost": 7.15,
              "Plan Rows": 400,
              "Plan Width": 0,
              "Index Cond": "(id < 5)"
            }
          ]
        }
      ]
    },
    "Settings": {
      "search_path": "\"test\", \"public\""
    }
  }
]
</pre>
</details>

> SETTINGS
> Include information on configuration parameters. Specifically, include options affecting query planning with value different from the built-in default value. This parameter defaults to FALSE.

This seems really useful for debugging as we have had questions about `search_path` in the past. However, I'm worried that this might leak our [in-db config](https://postgrest.org/en/stable/configuration.html#in-database-configuration) settings, specifically the `jwt-secret`. For now I haven't been able to expose settings other than `search_path` but this seems like it could break in the future. If an anonymous user is able to specify the plan media type, that would be disastrous.

I guess this concern only applies if a production instance gets our EXPLAIN config enabled, which is likely. Perhaps we can recommend an additional measure: a `db-pre-request` that only enables the plan media type from certain IPs, otherwise rejects it.

It might also be worth considering tying the two together, i.e. enabling the plan media type like `db-plan-enabled = "172.168.1.x, 172.168.1.y"`.

### Specifying TIMING or WAL without ANALYZE gets a nice error message

```
curl 'localhost:3000/projects?id=lt.5' -H "Accept: application/vnd.pgrst.plan; options=timing"

HTTP/1.1 400 Bad Request
{"code":"22023","details":null,"hint":null,"message":"EXPLAIN option TIMING requires ANALYZE"}
```

---

Overall all the options seem pretty useful. I don't see why we should not expose all of them if we can ensure the callers are restricted.

steve-chavez:
Exposing COSTS is unnecessary because is already TRUE by default for all.

> COSTS
Include information on the estimated startup and total cost of each plan node, as well as the estimated number of rows and the estimated width of each row. This parameter defaults to TRUE.

TIMING as well because requires ANALYZE and it's TRUE by default here.

> TIMING
Include actual startup time and time spent in each node in the output. The overhead of repeatedly reading the system clock can slow down the query significantly on some systems, so it may be useful to set this parameter to FALSE when only actual row counts, and not exact times, are needed. Run time of the entire statement is always measured, even when node-level timing is turned off with this option. This parameter may only be used when ANALYZE is also enabled. It defaults to TRUE.

SUMMARY ditto. It doesn't add any info on non-ANALYZE and it's already included for ANALYZE.

> SUMMARY
Include summary information (e.g., totaled timing information) after the query plan. Summary information is included by default when ANALYZE is used but otherwise is not included by default, but can be enabled using this option. Planning time in EXPLAIN EXECUTE includes the time required to fetch the plan from the cache and the time required for re-planning, if necessary.

steve-chavez:
Just noted that with this feature we can remove the [QueryCost](https://github.com/PostgREST/postgrest/blob/main/test/spec/QueryCost.hs) test and instead include the cost tests on the PlanSpec; which is good because QueryCost was always brittle(added on https://github.com/PostgREST/postgrest/pull/1385).

steve-chavez:
Added a plain text output to be compatible with https://explain.depesz.com/. The text output is really cool, same one as `psql`:

```
curl 'localhost:3000/projects?id=in.(1,2,3)' -H "Accept: application/vnd.pgrst.plan+text;options=analyze"

Aggregate  (cost=15.61..15.63 rows=1 width=112) (actual time=0.024..0.025 rows=1 loops=1)
  ->  Bitmap Heap Scan on projects  (cost=8.48..15.59 rows=3 width=40) (actual time=0.008..0.009 rows=3 loops=1)
        Recheck Cond: (id = ANY ('{1,2,3}'::integer[]))
        Heap Blocks: exact=1
        ->  Bitmap Index Scan on projects_pkey  (cost=0.00..8.48 rows=3 width=0) (actual time=0.005..0.005 rows=3 loops=1)
              Index Cond: (id = ANY ('{1,2,3}'::integer[]))
Planning Time: 0.093 ms
Execution Time: 0.077 ms
```

src/PostgREST/MediaType.hs
```diff
@@ -27,7 +29,28 @@ data MediaType
...
+  MTTextCSV == MTTextCSV                 = True
+  MTTextPlain == MTTextPlain             = True
+  MTTextXML == MTTextXML                 = True
+  MTOpenAPI == MTOpenAPI                 = True
+  MTUrlEncoded == MTUrlEncoded           = True
+  MTOctetStream == MTOctetStream         = True
+  MTAny == MTAny                         = True
+  MTOther bs1 == MTOther bs2             = bs1 == bs2
+  MTPlan {} == MTPlan {}                 = True
+  _ == _                                 = False
```
steve-chavez, line 47: Added this for keeping the `mutuallyAgreeable` function working for the MTPlan. It's kinda icky but I didn't saw a better way without changing too much code.

steve-chavez: @laurenceisla Pointed me out [to this technique](https://stackoverflow.com/a/5956095/4692662) so hacking the MediaType Eq is no longer necessary.

laurenceisla (approved):

wolfgangwalther:
I just skimmed the tests quickly, but couldn't find it: Do we already have the `for=` parameter implemented as mentioned, [here](https://github.com/PostgREST/postgrest/issues/2354#issuecomment-1174448658)? Or did you conclude that it's not actually necessary?

steve-chavez:
@wolfgangwalther Ah, no. I just left a note in the CHANGELOG(docs would be the same) that `for` is a TODO(next version).

> Limited to generating the plan of a json representation(application/json) but can be extended later to allow other representations.

steve-chavez:
Adding `for` in https://github.com/PostgREST/postgrest/pull/2400

PR #2389, opened 2022-07-27
raise timeout in failing schema cache reload test

robx:
Locally, this make the test hang. Filing this PR to see if the same happens in CI, and to ask for help.

This is a test that's causing me trouble while upgrading hasql-pool.


robx:
Found the problem, filed as #2390.

Issue #2348, opened 2022-06-24
Add `db-pool-query-wait-timeout` config


steve-chavez:
## Problem

- PostgREST doesn't have a timeout for how long a request should wait for obtaining a connection from the db pool.
- Nginx(or other proxy) has a request timeout.

When a timeout occurs, this usually causes an error on the client side that has poor information: usually just a 502 Bad Gateway or 408 Request Timeout. This happens on servers with high amount of traffic.

The problem can be reproduced with an RPC that has a `pg_sleep()` inside plus Nginx with the req timeout.

(I haven't checked if the request will still reach the database once the proxy timeout occurs, it's possible)

## Proposal

Have a `db-pool-query-wait-timeout` config that produces a timeout from our side, similar to the pgbouncer [query_wait_timeout](https://www.pgbouncer.org/config.html#query_wait_timeout). The pgbouncer config has a default of 120 but in our case it should be lower since [Nginx has a 60 second default](https://tecadmin.net/increase-request-timeout-in-nginx.). Maybe it should be like 50 seconds.

On the http side, when this happens we should return a `529 Site is overloaded` status. Optionally with a `Retry-After` header, maybe we can use the [pool waiter queue size](https://github.com/PostgREST/postgrest/pull/2129) to determine a good value.

**Edit**: For now maybe we should just keep it simple and return a `504 Gateway Timeout` with a descriptive error message.

---

This would require a patch in [hasql-pool](https://github.com/nikita-volkov/hasql-pool).

steve-chavez:
## To Reproduce

Start a postgrest with a sleep function:

```bash
cat << EOF > test/spec/fixtures/schema.sql

CREATE OR REPLACE FUNCTION sleep() returns void as $$
  select pg_sleep(50);
$$ language sql;

EOF

PGRST_DB_POOL=1 PGRST_DB_ANON_ROLE='postgrest_test_anonymous' postgrest-with-postgresql-14 postgrest-run
```

On other terminals do:

```bash
# take the single pool connection
http GET 'localhost:3000/rpc/sleep'
```

```bash
# This one will wait 50 seconds until the sleep request is done

http GET 'localhost:3000/projects'
```

PR #2391, opened 2022-07-27
Upgrade hasql-pool

### LINKED ISSUE #2401: Database connections can outlive SIGUSR1 reconfiguration
The intended behaviour when receiving `SIGUSR1` is for the PostgREST process to refresh the state of the database (reload schema cache); the new state is expected to take effect for any requests received from that point onwards.

E.g., in the sequence
1. `alter role authenticator set statement_timeout = ...`
2. `kill -USR1 $postgrest`
3. `curl http://postgrest/...`
the updated timeout should be in effect for the request in step 3.

To that aim, the signal handler "releases" the postgresql connection pool. However, with the present version of `hasql-pool`, releasing the pool merely destroys idle connections. Any busy connections survive, and can be returned to the pool at a later time.

Thus, we get e.g. the following failure scenario:
- any time before step 2 above, issue a request to postgrest that results in a slow postgresql query (e.g. `pg_sleep(...)`); suppose this request uses postgresql connection `C`
- suppose this request finishes between steps 2 & 3 above; `C` is returned to the pool
- suppose the request of step 3 picks up connection `C` from the pool; `C` still has the old timeout setting, violating the expectation.

(This should be fixed during the upgrade to a newer release of `hasql-pool` #2391, where we have to touch the pool release logic anyhow. This issue is just for tracking the bug.)

robx:
Preparatory work for #2348, as discussed with @steve-chavez.

This loses the `db-pool-timeout` config option (and timing out of idle connections at all), and fixes #2401.


test/io/test_io.py
```diff
@@ -545,26 +544,6 @@ def test_iat_claim(defaultenv):
             time.sleep(0.1)
 
 
-def test_app_settings(defaultenv):
-    """
-    App settings should not reset when the db pool times out.
```
robx, line 550: This test still passes, but it no longer makes sense.

test/io/test_io.py
```diff
@@ -844,6 +823,7 @@ def test_admin_ready_wo_channel(defaultenv):
         assert response.status_code == 200
 
 
+@pytest.mark.skip
```
robx, line 826: This test now fails, because we no longer flush the pool on `SIGUSR1`.

The fact that this test relies on that behaviour seems to rather be an implementation detail of the test -- it just needs schema cache reloading to fail, which we used to achieve by setting the session timeout.

Could there be a way to get the introspection query to fail more directly, by messing up the schema or permissions?

src/PostgREST/Workers.hs
```diff
@@ -99,18 +99,13 @@ connectionWorker appState = do
...
--- Which might not happen if the server is busy with requests. No idle
--- connection, no pool timeout.
+-- database.
 --
 -- The connection tries are capped, but if the connection times out no error is
 -- thrown, just 'False' is returned.
 connectionStatus :: AppState -> IO ConnectionStatus
 connectionStatus appState =
-  retrying retrySettings shouldRetry $
-    const $ SQL.release pool >> getConnectionStatus
```
steve-chavez, line 113: Hmm.. one thing that was useful with releasing the pool. pg role settings kick in at login time, so by doing `ALTER ROLE rol SET statement_timeout = 5` and then `SIGUSR1`(or `NOTIFY pgrst`) we enforced the new `statement_timeout` for all pool connections.

I think it would be possible to apply the role settings ourselves but it'd be easier to let pg do that.

Why is necessary to remove `SQL.release`? Doing that cause errors now?

test/io/test_io.py
steve-chavez: I think it might be possible by revoking privileges on one of the pg_catalog objects.

src/PostgREST/Workers.hs
robx: Yes, `SQL.release` essentially destroys the pool.

Without patching `hasql-pool`, getting the same behaviour would require replacing the pool. That's quite possible but a bit of a hassle, I would have avoided it if not necessary.

E.g. something like
- in `AppState`, have a `TVar Pool` instead of `Pool`
- where we currently call `release`, instead
  1. create new pool
  2. place it in the tvar
  3. release the old pool

Do you think that's a good way to go? An alternative could be to patch `hasql-pool` to provide a variant of `release` (`flush`?) that keeps the pool alive and merely closes idle connections.

(It would be good to have a dedicated test for the `statement_timeout` functionality if that's a feature we want -- might be this test is basically that, but by its name etc. it seems to be aiming at a different feature. I'm imagining a test similar to this one that
- runs something like `pg_sleep(n)` successfully
- lowers the statement timeout to `< n` similarly to this test, then sends `USR1`
- and now the same `pg_sleep(n)` rpc times out
Does that seem sensible?)

test/io/test_io.py
steve-chavez: Right. So really the whole app.settings feature was meant to add GUCs at the session level(https://github.com/PostgREST/postgrest/pull/1063) but those got lost on timeouts. So I made a lazy fix(https://github.com/PostgREST/postgrest/pull/1169, [this test was in bash at that time](https://github.com/PostgREST/postgrest/pull/1169/files#diff-20d0d2179c4a0dea3feea7ebcc9627b8296ef717828184176520808957f36906R146-R165)) to move them to the transaction level instead - which is wasteful considering they don't change on each transaction.

Now that we don't have timeouts, perhaps we can move `app.settings` to the session level. This would also pair up with the `statement_timeout` feature [discussed below](https://github.com/PostgREST/postgrest/pull/2391#discussion_r931279472), we could reuse the same code to set it at the session level ourselves.

src/PostgREST/Workers.hs
steve-chavez: > 1. Without patching hasql-pool, getting the same behaviour would require replacing the pool. That's quite possible but a bit of a hassle, I would have avoided it if not necessary.
> 2. An alternative could be to patch hasql-pool to provide a variant of release (flush?) that keeps the pool alive and merely closes idle connections.

Yeah 1 seems too complicated. 2 wouldn't help because we'd need to close all connections(not only idle ones) to keep the same behavior.

> I think it would be possible to apply the role settings ourselves but it'd be easier to let pg do that.

Another option would be getting the statement_timeout from `pg_db_role_settings` and applying it at the session level ourselves. To reset the current `statement_timeout`, perhaps we can run a `SET statement_timeout` on all connections when `SIGUSR1/NOTIFY` is done.

steve-chavez: > runs something like pg_sleep(n) successfully
> lowers the statement timeout to < n similarly to this test, then sends USR1
> and now the same pg_sleep(n) rpc times out
> Does that seem sensible?)

Yeah, that sounds good.

> Another option would be getting the statement_timeout from pg_db_role_settings and applying it at the session level ourselves.

I can help with that one btw, if there's no simpler alternative.

robx: > 2 wouldn't help because we'd need to close all connections(not only idle ones) to keep the same behavior.

Just to clarify this one point: Currently, does releasing the pool interrupt any in-progress connections to the database? At first glance that would seem like a bit of a questionable design to me... E.g. that would mean HTTP error responses for requests that might have resulted in a successful database update. (Which isn't terrible, just something I'd have thought we'd avoid for this kind of configuration reload.) Intuitively, I wouldn't expect a statement timeout configuration change to apply to in-progress requests, just to requests that start after the change.

It does make sense to me that we'd want the configuration change to take effect immediately, so yeah the current state of the PR is no good, we need to close at least the idle connections on `SIGUSR1`.

steve-chavez: >  Currently, does releasing the pool interrupt any in-progress connections to the database?

Yes, this can be reproduced with a sleep function and a pool size of 1, doing a SIGUSR1 will then abort the sleep request.

> I wouldn't expect a statement timeout configuration change to apply to in-progress requests, just to requests that start after the change.

Hm, I think this would be a better behavior as well. Though in a busy server with all connections taken, a `flush` function wouldn't find any idle connections and then no new `statement_timeout` would be applied right? Or there could be a way to queue this `flush` operation?

steve-chavez: > Yes, this can be reproduced with a sleep function and a pool size of 1, doing a SIGUSR1 will then abort the sleep request.

Wait, I'm wrong. The above is not true if doing a `notify pgrst, 'reload schema';`(which also calls `release`) on a pool size of 2 and a request with `sleep`. Something wrong with my test, SIGUSR1 is somehow killing the database.

steve-chavez: So current `release` is indeed not interrupting in-progress connections to the db. I was wrong all along :man_facepalming: .

test/io/test_io.py
robx: It works again in its original form now that we flush the pool properly.

src/PostgREST/Workers.hs
robx: The updated version addresses this -- we now have a dedicated `flushPool` called here which replaces the pool with a fresh one.

test/io/test_io.py
robx: I don't quite follow the session vs transaction level setting, if you think that's a change we should make would you file an issue so we don't lose it? I think for this PR we don't need to make a change (and I've updated the test now so that it's in some way still useful).

src/PostgREST/AppState.hs
```diff
@@ -46,7 +46,9 @@ import Protolude
 
 
 data AppState = AppState
-  { statePool         :: SQL.Pool -- | Connection pool, either a 'Connection' or a 'ConnectionError'
+  -- | Database connection pool; 'Nothing' indicates the pool has been destroyed
+  { statePool         :: IORef (Maybe SQL.Pool)
```
robx, line 50: This is all (i.e., `IORef (Maybe SQL.Pool)` as well as `usePool`, `destroyPool`, `flushPool` below) a *bit* messy. It might be nicer to hide all of that in a separate `PostgREST.Pool` module with

```
module PostgREST.Pool
   ( Pool
   , init
   , use
   , flush
   , destroy
   ) where

newtype Pool = Pool (IORef (Maybe SQL.Pool))

...
```

and then moving back to offering `AppState.getPool :: AppState -> Pool` and calling it via e.g. `Pool.use (AppState.getPool appState) xxx` all over the place.

That approach would also lend itself to unit testing our pool wrapper. On the other hand I'm not sure it's worth the effort right now so I decided ot file the PR as it is for the moment. What do you think?

src/PostgREST/AppState.hs
```diff
@@ -91,17 +94,54 @@ initWithPool newPool conf =
...
-releasePool :: AppState -> IO ()
-releasePool AppState{..} = SQL.release statePool
+usePool appState@AppState{..} session = do
+  pool <- readIORef statePool
+  case pool of
+    Nothing -> return $ Left SQL.PoolIsReleasedUsageError
+    Just p -> do
+      res <- SQL.use p session
+      case res of
+        Left SQL.PoolIsReleasedUsageError -> usePool appState session
```
robx, line 121: This potentially unlimited recursion is a bit scary. I claim that the only situation in which this code path is hit is in the sequence

1. call `usePool`, read `pool1` from the `IORef`
2. `SIGUSR1` hits (or similar), calling `flushPool`
3. `flushPool` creates a new pool `pool2`, switches out the `IORef`, releases `pool1`
4. the `usePool` call from step 1 proceeds to call `SQL.use`, which finds a released pool and thus causes this error.

And now calling `usePool` anew is fine, because we'll necessarily get `pool2` (or later) on the next try.

We could make the code more obviously safe by using `retry` in some form. Or cover it by unit tests.

src/PostgREST/Workers.hs
steve-chavez: @robx Really sorry for the misdirection, I forgot about a majorly important detail here.

It's critical that [Automatic Schema Cache Reloading](https://postgrest.org/en/latest/schema_cache.html#automatic-schema-cache-reloading) can be done continuously on a production instance. For that, we can't have ​a `NOTIFY pgrst, 'reload schema'` or `SIGUSR1` interrupting running requests(commented that pending test in https://github.com/PostgREST/postgrest/issues/1766).
 
Under that restriction, it'd be preferable to lose the eventual application(due to the old pool timeout) of the new `statement_timeout`. 

Another option I see would be to move the `flushPool` to another unix signal like SIGHUP(we had trouble with using this one before). However, I don't see the real benefit over a `systemctl restart postgrest` since we'll be inoperative anyway.

It might be possible to apply `statement_timeout` with session level settings later. So it's not that bad to lose this now.

robx: This doesn't interrupt running requests, though. What `flushPool` does is 1. create a new pool 2. switch out the pools in the `IORef`, so any new requests will be handled by the new pool 3. releases the old pool, which just means that idle connections from the old pool are closed.

Running requests aren't affected; the `use` call for a running request will just find that its pool has been released, so the connection is closed instead of returning it to the pool.

steve-chavez: That is awesome. Can that be remarked as a comment on the `flushPool` function? A test that confirms we don't interrupt running requests would also be really nice.

steve-chavez: Comment for posterity, I was being fooled by this manual test:

```
PGRST_DB_ANON_ROLE='postgrest_test_anonymous' postgrest-with-postgresql-14
postgrest-run

05/Aug/2022:12:07:49 -0500: Attempting to connect to the database...
05/Aug/2022:12:07:49 -0500: Connection successful
05/Aug/2022:12:07:49 -0500: Listening on port 3000
05/Aug/2022:12:07:49 -0500: Config reloaded
05/Aug/2022:12:07:49 -0500: Listening for notifications on the pgrst channel
05/Aug/2022:12:07:49 -0500: Schema cache loaded

curl localhost:3000/rpc/sleep

# on another terminal I did
# kill -USR1 $(pgrep postgrest)
# and then the request was interrupted

{"code":"PGRST001","details":"no connection to the server\n","hint":null,"message":"Database client error. Retrying the connection."}
```

My mistake was omitting the `-n(newest)` flag in pgrep(should have been a `pgrep -n postgrest`) and I was killing postgresql instead. TIL that [SIGUSR1 triggers a quickdie for pg](https://www.postgresql.org/docs/7.0/signals.htm).

src/PostgREST/AppState.hs
steve-chavez: Sure, that's fine by me. The Pool module can be added later in another PR.

src/PostgREST/Workers.hs
steve-chavez: > 2. switch out the pools in the IORef, so any new requests will be handled by the new pool
> 3. releases the old pool, which just means that idle connections from the old pool are closed.

Question here, does that mean the old pool connections will continue to live on when the running requests are finished? Now that we don't have the idle timeout those will not be closed right?

src/PostgREST/Error.hs
```diff
@@ -216,12 +216,17 @@ instance JSON.ToJSON PgError where
...
     "message" .= ("Database connection error. Retrying the connection." :: Text),
     "details" .= (T.decodeUtf8With T.lenientDecode $ fromMaybe "" e :: Text),
     "hint"    .= JSON.Null]
-  toJSON (SQL.SessionError e) = JSON.toJSON e -- SQL.Error
+  toJSON (SQL.SessionUsageError e) = JSON.toJSON e -- SQL.Error
+  toJSON SQL.PoolIsReleasedUsageError = JSON.object [
+    "code"    .= InternalErrorCode00,
+    "message" .= ("Use of released pool" :: Text),
+    "details" .= JSON.Null,
+    "hint"    .= JSON.Null]
```
steve-chavez, line 229: Is it possible for the user to get this error? AIUI, on `usePool` we try to recover from `PoolIsReleasedUsageError` immediately.

If it's possible, perhaps this momentary error should include a "Retrying.." message and a `503` status to clarify it's temporary.

src/PostgREST/AppState.hs
steve-chavez: Yeah, true :fearful:. I think at minimum we should log the retry with `AppState.logWithZTime`.

src/PostgREST/Error.hs
robx: It's not expected to be visible to the user except during shutdown, so not temporary in that sense.

src/PostgREST/Workers.hs
robx: They'll be closed as soon as their requests are done. When `use` tries to return them to the released pool, it closes them and discards them instead.

robx:
@steve-chavez I believe I've addressed the review comments, could you have another look?

src/PostgREST/AppState.hs
```diff
@@ -91,17 +94,58 @@ initWithPool newPool conf =
...
+      res <- SQL.use p session
+      case res of
+        Left SQL.PoolIsReleasedUsageError -> do
+          logWithZTime appState "Attempted to used flushed pool, retrying"
+          usePool appState session
+        _                                 -> return res
+
+-- | Flush the connection pool so that any future use of the pool will
+-- use connections freshly established after this call. In-use
+-- connections aren't affected, they just won't be reused anymore.
```
robx, line 128: There's one issue with the way we replace pools, in that there's a timespan where two pools are around (the old one with the in-process requests, and the new empty one). Until all the old requests are done, it's possible to exceed the configured pool size.

robx: I've implemented this as a patch to `hasql-pool` here https://github.com/nikita-volkov/hasql-pool/pull/16.

src/PostgREST/Error.hs
steve-chavez: I see. In that case maybe we should a have more user-facing error message, like: 

```suggestion
  toJSON SQL.PoolIsReleasedUsageError = JSON.object [
    "code"    .= InternalErrorCode00,
    "message" .= ("The database connection pool is being released. Shutting down.." :: Text),
    "details" .= JSON.Null,
    "hint"    .= JSON.Null]
```

src/PostgREST/AppState.hs
steve-chavez: Just to be clear, does https://github.com/nikita-volkov/hasql-pool/pull/16 prevent exceeding the configured pool size?

steve-chavez:
Q: With the new `flushPool`, do see any performance impact in comparison to the old `releasePool`?

src/PostgREST/AppState.hs
robx: Yes.

> Just to be clear, does [nikita-volkov/hasql-pool#16](https://github.com/nikita-volkov/hasql-pool/pull/16) prevent exceeding the configured pool size?

Yes (well unless there's a bug in the PR).

robx:
> Q: With the new `flushPool`, do see any performance impact in comparison to the old `releasePool`?

I don't expect a meaningful performance impact for `flushPool`. The extra cost should be just looking up an `IORef`.

There's more likely to be performance regressions lurking in the `hasql-pool` rewrite, though the new version is simple enough that I don't expect anything there either. (The current `hasql-pool` version relies on an old `resource-pool` release, which uses STM just the same as the new `hasql-pool`, so I don't see any reason for new overhead as compared to the old version. Judging by the recent rewrite of `resource-pool` with non-STM for "better throughput and latency" there might be some room for improvement though.)

steve-chavez (approved):

steve-chavez:
@robx I think this should be good to merge. Or should we wait on closing https://github.com/nikita-volkov/hasql-pool/pull/16 or https://github.com/PostgREST/postgrest/issues/2422?

robx:
> @robx I think this should be good to merge. Or should we wait on closing [nikita-volkov/hasql-pool#16](https://github.com/nikita-volkov/hasql-pool/pull/16) or #2422?

I think it's time to fork (not just for this issue but those other hasql issues too). Here's a clone of nikita-volkov/hasql-pool#16: https://github.com/PostgREST/hasql-pool/pull/1.

robx:
Closing in favour of #2454.

PR #2397, opened 2022-08-01
fix: fix some race conditions running connection worker

robx:
Previously, it was quite possible to have two connection workers
running, or to get into a state where a failed connection worker
is still considered running preventing new connection workers
from starting.


steve-chavez (approved):
> preventing new connection workers from starting.

Not too long ago I got a case in production where a NOTIFY wasn't reloading the schema cache, this was likely it. Nice work!

steve-chavez:
Oh btw, this one should be listed in the [CHANGELOG](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#fixed)

PR #2402, opened 2022-08-02
Add test for #2401

robx:
This primarily adds a test to prove #2401.


test/io/test_io.py
```diff
@@ -86,14 +106,24 @@ def dburi():
 
 
 @pytest.fixture
-def defaultenv():
-    "Default environment for PostgREST."
+def baseenv():
+    "Base environment to connct to PostgreSQL"
```
steve-chavez, line 110:
```suggestion
    "Base environment to connect to PostgreSQL"
```

steve-chavez (approved):

PR #2403, opened 2022-08-02
refactor: move pool release on shutdown out of signal handler

robx:
This changes behaviour somewhat in that:
- We now consistently release the pool on shutdown, even on non-Unix
  platforms, and including for CmdDumpConfig.
- We release the pool *after* interrupting `App.run`, which will
  rather cause more than fewer connection to be closed properly.
  (Previously any in-use connections would not have been caught by
  `releasePool`, though *maybe* the `UserInterrupt` handling in
  the web handler ends up closing the connections properly already
  anyway).
    
(The main aim of the change is to make it clearer when and why the
pool is released.)

### DIFF
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -34,13 +34,19 @@ main :: App.SignalHandlerInstaller -> Maybe App.SocketRunner -> CLI -> IO ()
 main installSignalHandlers runAppWithSocket CLI{cliCommand, cliPath} = do
   conf@AppConfig{..} <-
     either panic identity <$> Config.readAppConfig mempty cliPath Nothing
-  appState <- AppState.init conf
-  case cliCommand of
-    CmdDumpConfig -> do
-      when configDbConfig $ reReadConfig True appState
-      putStr . Config.toText =<< AppState.getConfig appState
-    CmdDumpSchema -> putStrLn =<< dumpSchema appState
-    CmdRun -> App.run installSignalHandlers runAppWithSocket appState
+
+  -- Per https://github.com/PostgREST/postgrest/issues/268, we want to
+  -- explicitly close the connections to PostgreSQL on shutdown.
+  -- 'AppState.destroy' takes care of that.
+  bracket
+    (AppState.init conf)
+    AppState.destroy
+    (\appState -> case cliCommand of
+      CmdDumpConfig -> do
+        when configDbConfig $ reReadConfig True appState
+        putStr . Config.toText =<< AppState.getConfig appState
+      CmdDumpSchema -> putStrLn =<< dumpSchema appState
+      CmdRun -> App.run installSignalHandlers runAppWithSocket appState)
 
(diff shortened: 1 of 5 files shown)


src/PostgREST/CLI.hs
```diff
@@ -34,13 +34,19 @@ main :: App.SignalHandlerInstaller -> Maybe App.SocketRunner -> CLI -> IO ()
...
-  case cliCommand of
-    CmdDumpConfig -> do
-      when configDbConfig $ reReadConfig True appState
-      putStr . Config.toText =<< AppState.getConfig appState
-    CmdDumpSchema -> putStrLn =<< dumpSchema appState
-    CmdRun -> App.run installSignalHandlers runAppWithSocket appState
+
+  -- Per https://github.com/PostgREST/postgrest/issues/268, we want to
+  -- explicitly close the connections to PostgreSQL on shutdown.
+  -- 'AppState.destroy' takes care of that.
```
robx, line 40: The apparent reason that we want to release the pool is that otherwise postgresql keeps around workers. I *guess* that's because postgresql takes a while to notice a socket is gone when it's been closed by process shutdown on the other hand? I can't quite reproduce that problem though, so it's a bit hard for me to verify this continues to do what it needs to do.

steve-chavez: > otherwise postgresql keeps around workers

Yeah [backends](https://www.interdb.jp/pg/pgsql02.html#_2.1.2.) remained open on shutdown a long time ago. 

I think one way to prove this would be to query [pg_stat_activity](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW) like:

```
select count(*) = 0 from pg_stat_activity where usename = 'postgrest_test_authenticator';
# or
select count(*) = 0 from pg_stat_activity where application_name = 'postgrest';
# this would work if the connection string is `postgres://authenticator:mysecretpassword@localhost/postgres?application_name=postgrest`
```

..after shutdown. Not sure if that would be too hard to set up though.

robx: I've tried reproducing the original issue to no avail. Even if I remove the call to release pool in a test that creates a bunch of long-lived idle connections, they go away as soon as postgrest exits.

- the difference between releasing the pool or not would be whether the connections are actively terminated by sending a termination message via `PQfinish()`, or whether we leave it to the OS to close the socket
- even if we leave it to the OS, the postgresql server *should* notice the socket going away; there's a decent chance that the handling of this on the postgresql side has improved since the original issue, so maybe we'd need an old postgres version to verify it...
- there might also well be a difference between unix domain sockets and tcp sockets here

Overall, I'm pretty sure the change in this PR is good, and we handle active termination a bit better with this change then before. It's just a bit hard to verify, due to a mix of being unsure the underlying issue even exists anymore, and not knowing how to cause it...

steve-chavez (approved):

Issue #1105, opened 2018-05-09
Feature request: column to column comparison


priyank-purohit:
### Description of issue
Currently PostgREST supports only column-to-value(s) comparison, however column-to-column comparisons are extremely common and too important to not support.

### Implementation
This is a somewhat tricky feature to implement because you need to differentiate values from column names. The best solution I came up with is to add a `.c.` prefix between the operator and the column.

For example, `?and=(first_name.c.eq.last_name)` vs `?and=(first_name.eq.last_name)` to search for people with same first and last names vs. people whose first name is in the db as literal string 'last_name'.

> Another potential solution: duplicate each operator (where it makes sense the user wants to do column comparison), and this duplicate operator (with a slightly different name) will specifically interpret the value supplied by the user as a column name. If the value supplied is not a column name, 404 response. For example, the `.eq.` operator's column comparison equivalent could be `.ceq.`

steve-chavez:
Why not define that in the view you're exposing?, say like:
```sql
create view api.persons as (
  select id, first_name, last_name, first_name = last_name as is_tautonym 
  from private.persons
);
```
A computed column will also work, then you could do:
```http
/persons?is_tautonym=is.true
```

For now I'd be against implementing this, as it's more clean to represent that kind of comparisons in the db.

priyank-purohit:
@steve-chavez  That works for that one particular query, but not for any other queries (you can create any column-to-value comparison query otherwise). Is there any other reason besides the cleanliness of the PostgREST code?

Alternatively, I could implement something like this in a custom version of PostgREST; but I'm wondering if would this be a difficult undertaking for someone with 0 Haskell knowledge?

steve-chavez:
Considering the difficulty required towards implementing this and defining views/computed columns in a few lines of SQL, I'd say the latter is a much more easy option.

I think this kind of comparisons are core business logic, and as so, they should be defined at the db level with an appropriate name/comment on the column so the REST API can have proper documentation, as your project matures you're gonna need this.

Also there needs to be a limit in what kind of operations PostgREST exposes, having this somehow feels like a step closer to exposing custom functions in filters values.

priyank-purohit:
I would say that something like 'regular expression comparison operator' is a use-case specific feature, as in not everyone will want/need that (and it increases risk of ReDoS). Hence, that is something you can expose via custom functions if you need.

[...]

PR #2405, opened 2022-08-05
Add underscore equality operator to filter using keys in the request body

laurenceisla:
Using this idea https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1200054135, the underscore operators will filter using the keys in the body. This PR implements the following:

- The `_eq` operator to verify for equality with items in the body
- Support to select any identifier in the body (not only PKs). This  gives more freedom in the case of bulk updates, e.g.:
  ```http
  PATCH /user?id=_eq.id HTTP/1.1
  
  [
    { "username": "userA", "is_active": true },
    { "username": "userB", "is_active": false }
  ]
  ```
- Full table updates are allowed once again, addressing this issue: https://github.com/PostgREST/postgrest/pull/2195#discussion_r932946412

### DIFF
--- a/src/PostgREST/Request/Types.hs
+++ b/src/PostgREST/Request/Types.hs
@@ -31,6 +31,7 @@ module PostgREST.Request.Types
   , TrileanVal(..)
   , SimpleOperator(..)
   , FtsOperator(..)
+  , BodyOperator(..)
   ) where
 
 import qualified Data.ByteString.Lazy as LBS
@@ -179,6 +180,7 @@ data Operation
   | In ListVal
   | Is TrileanVal
   | Fts FtsOperator (Maybe Language) SingleVal
+  | BodOp BodyOperator SingleVal
   deriving (Eq)
 
 type Language = Text
@@ -224,4 +226,9 @@ data FtsOperator
   | FilterFtsPlain
   | FilterFtsPhrase
   | FilterFtsWebsearch
+  deriving Eq
+
+-- | Operators for filtering using the request body
+data BodyOperator
+  = BodyOpEqual
   deriving Eq
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -110,7 +110,7 @@ mutateRequestToQuery (Insert mainQi iCols body onConflct putConditions returning
     cols = BS.intercalate ", " $ pgFmtIdent <$> S.toList iCols
 
 -- An update without a limit is always filtered with a WHERE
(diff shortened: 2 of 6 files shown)


src/PostgREST/Request/Types.hs
```diff
@@ -225,3 +227,8 @@ data FtsOperator
   | FilterFtsPhrase
   | FilterFtsWebsearch
   deriving Eq
+
+-- | Operators for filtering using the request body
+data BodyOperator
+  = BodyOpEqual
+  deriving Eq
```
steve-chavez, line 234: I think it's fine to have this as a special case for now, we can add underscore versions of the other operators later.

One question though, will this work for DELETE and RPC as well?

laurenceisla: It does not work for DELETE in the PR because the body is ignored, but it should work with the changes in #2355 (unless those changes are done here to avoid breaking changes in the future). 

RPC are a different story in which the filter of the result set is done outside of the `pgrst_source` CTE in the QueryBuilder.hs file, so it won't work out of the box unless the filter is done inside the CTE (may break some things). The resulting query  looks something like:

```sql
WITH pgrst_source AS (
  WITH pgrst_payload AS (...).
            pgrst_body AS (...),
            pgrst_args AS (...)
           SELECT ... -- The WHERE should be done here for the filter to get the data from pgrst_body
SELECT ...
FROM
  (SELECT "table".*
   FROM "pgrst_source" AS "table"
   WHERE "table"."id" = "pgrst_recordset_body"."id") _postgrest_t  -- it is done outside instead, which gives an error
```

wolfgangwalther:
Wait, the comment in https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1200054135 says:

> Underscore operators could be used to operate on columns besides referencing values on the request body. This would allow doing https://github.com/PostgREST/postgrest/issues/1105#issuecomment-395238192 without computed columns:
>
> ```httpspec
> SEARCH /PredictedGenes?Strand=eq.+&GeneStart=_gt.GeneEnd
> ```
>
> Operating on a body value would still be allowed, body would be a reserved identifier:
>
> ```httpspec
> SEARCH /PredictedGenes?Strand=_eq.body->strand&GeneStart=_lt.GeneEnd
> { "Strand": "-"}
> ```

With the implementation in this PR, the ability to do column to column matches will be lost, right?

laurenceisla:
> With the implementation in this PR, the ability to do column to column matches will be lost, right?

Yes. It was one of the reasons I turned this into a draft for now. It should use a syntax like `?col=_eq.$.col` to get filters from the body and allow column matches too.

Issue #2410, opened 2022-08-08
Bug "UNION types \"char\" and text cannot be matched" with v9.0.1.+++ versions,


yevon:
### Environment

* PostgREST docker image version: **v9.0.1.20220802** or **v9.0.1.20220714**

### Description of issue

I was trying versions above v9.0.1, from v9.0.1.20220714, but when I change the docker image I just receive errors in the docker logs. Any idea what could it be? If I just go back to the stable v9.0.1 it works ok. I tried a schema reload without success.

![image](https://user-images.githubusercontent.com/121157/183502027-0834d51e-1b93-4c92-b135-1a533dcfd8e1.png)

wolfgangwalther:
Which PostgreSQL version are you using?

Can you try to put together a reproducible example? There's probably something in your schema that triggers this.

wolfgangwalther:
You could also have a look at the PostgreSQL logs - the statement that failed should be logged there, could you show us that, too?

yevon:
I'm using the postgres 15 beta 2. I think that I found the query that fails, if I try to manually execute it it fails:

Found here:

[DbStructure.hs](https://github.com/PostgREST/postgrest/blob/bac63f339daa4a9292870b0674b41cde2bcfee74/src/PostgREST/DbStructure.hs)

`with recursive
      pks_fks as (
        -- pk + fk referencing col
        select
          contype,
          conname,
          conrelid as resorigtbl,
          unnest(conkey) as resorigcol
        from pg_constraint
        where contype IN ('p', 'f')
        union
        -- fk referenced col
        select
          concat(contype, '_ref') as contype,
          conname,
          confrelid,
          unnest(confkey)
        from pg_constraint
        where contype='f'
      ),
      views as (
        select
          c.oid       as view_id,
          n.nspname   as view_schema,
          c.relname   as view_name,
          r.ev_action as view_definition
        from pg_class c
        join pg_namespace n on n.oid = c.relnamespace
        join pg_rewrite r on r.ev_class = c.oid
        where c.relkind in ('v', 'm') and n.nspname = ANY($1 || $2)
      ),
      transform_json as (
        select
          view_id, view_schema, view_name,
          -- the following formatting is without indentation on purpose
          -- to allow simple diffs, with less whitespace noise
          replace(
            replace(
            replace(
            replace(
            replace(
            replace(
            replace(
            regexp_replace(
            replace(
            replace(
            replace(
            replace(
            replace(
            replace(
            replace(
[...]

PR #2413, opened 2022-08-08
Fix bug "UNION types \"char\" and text cannot be matched" with v9.0.1.+++ versions

yevon:
The query in:

https://github.com/PostgREST/postgrest/blob/bac63f339daa4a9292870b0674b41cde2bcfee74/src/PostgREST/DbStructure.hs#L655-L809

Doesn't properly consider the type of the column contype (char) of the table pg_constraint. So when it tries to make a union with the concat function (text), fails with a missmatch type error. This fixes the issue. 

Related issue here:

[Reported issue](https://github.com/PostgREST/postgrest/issues/2410)

#2410

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -656,7 +656,7 @@ allViewsKeyDependencies =
       pks_fks as (
         -- pk + fk referencing col
         select
-          cast(contype as text) contype,
+          contype::text as contype,
           conname,
           conrelid as resorigtbl,
           unnest(conkey) as resorigcol
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -656,7 +656,7 @@ allViewsKeyDependencies =
       pks_fks as (
         -- pk + fk referencing col
         select
-          contype,
+          cast(contype as text) contype,
           conname,
           conrelid as resorigtbl,
           unnest(conkey) as resorigcol
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2342, Fix inaccurate result count when an inner embed was selected after a normal embed in the query string - @laurenceisla
  - #2376, OPTIONS requests no longer start an empty database transaction - @steve-chavez
  - #2395, Allow using columns with dollar sign($) without double quoting in filters and `select` - @steve-chavez
+ - #2410, Fix loop crash error on startup in Postgres 15 beta 2. Log: "UNION types \"char\" and text cannot be matched". - @yevon
 
 ### Changed
 


steve-chavez:
I don't have a pg 15 to test it, but LGTM!

Could you add an entry to the [Fixed section in the CHANGELOG](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#fixed)?

yevon:
> I don't have a pg 15 to test it, but LGTM!
> 
> Could you add an entry to the [Fixed section in the CHANGELOG](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#fixed)?

Done!

src/PostgREST/DbStructure.hs
```diff
@@ -656,7 +656,7 @@ allViewsKeyDependencies =
       pks_fks as (
         -- pk + fk referencing col
         select
-          contype,
+          cast(contype as text) contype,
```
wolfgangwalther, line 659:
```suggestion
          contype::text,
```

That's the style we use around there, I think.

wolfgangwalther:
I think this is part of a bigger issue: PG15 support. It makes little sense to me to commit something as isolated as this now.

We need to ensure this doesn't break again and that we don't have any other problems with PG15. So we should add PG15 to our postgres versions in nix. I don't think this version is currently in the nixpkgs, so we'd need to figure out how to install it. But I guess we should do that once, and then we can always use the latest beta version ahead of time to confirm everything still works.

wolfgangwalther:
> I don't think this version is currently in the nixpkgs, so we'd need to figure out how to install it.

Looks like pg14 was added as beta1 already to nixpkgs last year: https://github.com/NixOS/nixpkgs/commit/285fa468b19d8280fa0b46c9003daec9d5c6bb32

So maybe we just need somebody to put up a PR for pg15 beta 2?

wolfgangwalther:
> So maybe we just need somebody to put up a PR for pg15 beta 2?

That somebody was me, I guess: https://github.com/NixOS/nixpkgs/pull/185751

yevon:
I haven't checked if this is working with previous versions of postgres. But it seems that it should also fail because this union is wrong anyway in any postgres version. Is it possible that the column contype was added just after stable 9.0.1? 9.0.1 is working nicely with postgres 15 beta 2.

wolfgangwalther:
> I haven't checked if this is working with previous versions of postgres. But it seems that it should also fail because this union is wrong anyway in any postgres version.

Right, that makes sense. Can you test this with pg14, too?

If that fails as well, we need to find out what you have in your schema that triggers this, because we certainly want to test this to avoid a regression.

> Is it possible that the column contype was added just after stable 9.0.1? 9.0.1 is working nicely with postgres 15 beta 2.

No, not really. The same construct was there before, but it was moved around to a different query since.

yevon:
I will test this with postgres 14

wolfgangwalther:
> > Is it possible that the column contype was added just after stable 9.0.1? 9.0.1 is working nicely with postgres 15 beta 2.
> 
> No, not really. The same construct was there before, but it was moved around to a different query since.

Scratch that, I confused this with something else. The `contype` was indeed added in the meantime in d2719420f46a537a75d76a4666f8bffec582d4f8.

wolfgangwalther:
I'm pretty sure this issue came up after the change discussed here https://www.postgresql.org/message-id/flat/2216388.1638480141%40sss.pgh.pa.us

So this should really be a pg15 issue only.

yevon:
You are right then, postgres 15 seems to not consider char as text anymore as it is deprecated and this should not affect past versions of postgres. Anyway it is working fine until postgREST 9.0.1 stable so this is due to recent changes only. Do I have a way to build a docker image locally to try this out? It was trying to find a dockerfile, but it seems that you use Nix. I don't know anything about nix but it should be pretty the same?

yevon:
Could you build a docker image with this change so I can try this with postgres 15?

steve-chavez:
Seeing that https://github.com/NixOS/nixpkgs/pull/185751 is blocked, I think we should just merge here.

I've added a note [here](https://github.com/PostgREST/postgrest/issues/1766#issuecomment-1216052378) for testing pg15 when it's available.

src/PostgREST/DbStructure.hs
laurenceisla: Yes, we use this style more (`::`), can confirm that tests run successfully for docker `postgres 15beta3` btw.

steve-chavez: @yevon Can you make this change? Somehow we cannot commit the suggestion.

laurenceisla:
I've run the tests and they work successfully for docker postgres 15 beta2 and [beta3](https://hub.docker.com/layers/postgres/library/postgres/15beta3/images/sha256-7e8c76773c37c58e2254a46207381244c921a4e0224669dabea82088498bd3fb?context=explore).

yevon:
Suggested syntax change done! Also modified the column alias syntax "as contype" as it seems that this is always used for column alias also.

yevon:
Confirm is working nicely version v10.0.0!

rrbagautdinov:
> 

Hi, i have similar problem on pg:latest (15) tag (docker)

yevon:
> > 
> 
> Hi, i have similar problem on pg:latest (15) tag (docker)

Upgrade postgrest to at least v10.0.0, this is not happening anymore, already solved.

PR #2414, opened 2022-08-09
Allow full table updates again

laurenceisla:
Full table updates where disallowed in 86574c8. This PR brings back that functionality because of the reasons mentioned here: https://github.com/PostgREST/postgrest/pull/2195#discussion_r932946412. This issue is also related, mentions discussions on how to handle bulk updates: https://github.com/PostgREST/postgrest/issues/1959

- [x] Allow full table update
- [x] Add tests


src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -111,12 +111,12 @@ mutateRequestToQuery (Update mainQi uCols body logicForest pkFlts range ordts re
     SQL.sql $ "SELECT " <> emptyBodyReturnedColumns <> " FROM " <> fromQi mainQi <> " WHERE false"
 
   | range == allRange =
-    let whereLogic | null logicForest = if null pkFlts then "FALSE" else pgrstUpdateBodyF
-                   | otherwise        = logicForestF in
+    let whereLogic | null logicForest = if null pkFlts || not colsHavePk then mempty else " WHERE " <> pgrstUpdateBodyF
```
steve-chavez, line 114: I'm puzzled by this. By looking at this block, seems like a `WHERE` is applied all the time. Can it be clarified?

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE MultiWayIf #-}
```
steve-chavez, line 2: This is unused right?

laurenceisla: Yes, was testing something and forgot to remove it.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -137,6 +138,7 @@ mutateRequestToQuery (Update mainQi uCols body logicForest pkFlts range ordts re
 
   where
     mainTbl = SQL.sql (fromQi mainQi)
+    pkInCols = all (`elem` uCols) pkFlts
```
steve-chavez, line 141:
```suggestion
    pkInCols = (S.fromList pkFlts) `S.isSubsetOf` pkFlts
```

test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -37,40 +137,248 @@ spec =
...
+            , { "id": 3, "name": "updated-item", "observation": null }
+            ]|]
+
+        request methodPost "/rpc/reset_items_tables"
+          [("Prefer", "tx=commit")]
+          [json| {"tbl_name": "unsafe_update_items"} |]
+          `shouldRespondWith` ""
+          { matchStatus  = 204 }
+
+      it "works wihtout filters, taking only the first item in the json array body" $ do
```
steve-chavez, line 207: Hm, this test here seems unrelated to pg-safeupdate. Doesn't it belong to the UpdateSpec, on the bulk patch section?

test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -37,40 +137,248 @@ spec =
...
+            , { "id": 3, "name": "item-all", "observation": "Damaged item" }
+            ]|]
+
+        request methodPost "/rpc/reset_items_tables"
+          [("Prefer", "tx=commit")]
+          [json| {"tbl_name": "unsafe_update_items"} |]
+          `shouldRespondWith` ""
+          { matchStatus  = 204 }
+
+      it "works without filters and no pk in ?columns, taking only the first item in the json array body" $ do
```
steve-chavez, line 244: Ditto

test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -37,40 +137,248 @@ spec =
...
+            , { "id": 3, "name": "item-3", "observation": "Damaged item" }
+            ]|]
+
+        request methodPost "/rpc/reset_items_tables"
+          [("Prefer", "tx=commit")]
+          [json| {"tbl_name": "unsafe_update_items"} |]
+          `shouldRespondWith` ""
+          { matchStatus  = 204 }
+
+      it "works without filters and no composite pk in ?columns, taking only the first item in the json array body" $ do
```
steve-chavez, line 281: Ditto.

test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -37,40 +137,248 @@ spec =
...
+            , { "id": 1, "name": "item-3", "observation": "Damaged item" }
+            ]|]
+
+        request methodPost "/rpc/reset_items_tables"
+          [("Prefer", "tx=commit")]
+          [json| {"tbl_name": "unsafe_update_items_cpk"} |]
+          `shouldRespondWith` ""
+          { matchStatus  = 204 }
+
+      it "works with no pk defined in the table, taking only the first item in the json array body" $ do
```
steve-chavez, line 318: Ditto.

test/spec/Feature/Query/QueryLimitedSpec.hs
```diff
@@ -93,13 +93,16 @@ spec =
               { "id": 8, "name": "HaikuOS" } ]|]
             { matchStatus  = 201 }
 
-      it "doesn't affect updates(2 rows would be modified if it did)" $
+      it "doesn't affect updates" $
```
steve-chavez, line 96: The comment in the parens, it still applies right? I think that was related to https://github.com/PostgREST/postgrest/issues/2155

If true then that should be brought back.

laurenceisla: True, I was misunderstanding the test. Fixed.

steve-chavez:
As it is, the PR causes that the following:

```http
PATCH /table

{"id": 1, "name": "new-name"}
```

No longer does a full table update(doesn't trigger a pg-safeupdate error if installed) but instead it would update the row with id 1. However if we do:

```http
PATCH /table

{"name": "new-name"}
```

This would do a full table update because there's no id in the body.

Somehow this feels like surprising behavior to me, the url no longer denotes the operation. It could be very confusing for users.

I still think that doing the bulk patch(also delete) would be more explicit and consistent with underscore operators. However we cannot add a single one like on https://github.com/PostgREST/postgrest/pull/2405, we need a consistent design for all <- this could take a while and it'd be better for a next release.

---

What if we reopen https://github.com/PostgREST/postgrest/issues/1959, revert back https://github.com/PostgREST/postgrest/pull/2311 and leave bulk patch for the next release?

wolfgangwalther:
> What if we reopen https://github.com/PostgREST/postgrest/issues/1959, revert back https://github.com/PostgREST/postgrest/pull/2311 and leave bulk patch for the next release?

Yes, I came to the same conclusion yesterday, after reading through everything.

PR #2418, opened 2022-08-09
Add tests for the pg-safeupdate library

laurenceisla:
Adds tests to verify if PostgREST is allowing full table updates/deletes and the interaction with the `pg-safeupdate` library.

This will serve as basis for #2414. The full table update tests will be implemented there and the full table deletes, here and in #2355.


test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+      it "does not update and throws no error if no condition is present" $
+        request methodPatch "/safe_update"
+            [("Prefer", "tx=commit"), ("Prefer", "count=exact")]
+            [json| {"name": "New name"} |]
+          `shouldRespondWith`
+            ""
+            { matchStatus  = 404
+            , matchHeaders = [ matchHeaderAbsent hContentType
+                             , "Content-Range" <:> "*/0" ]
+            }
```
steve-chavez, line 26: So this proves that `pgsafeupdate` is broken for full table updates. It should throw a `UPDATE requires a WHERE clause`.

Related https://github.com/PostgREST/postgrest/pull/2195#discussion_r932946412

test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -0,0 +1,74 @@
...
+      it "deletes and does not throw error if no condition is present" $ do
+        request methodDelete "/unsafe_delete"
+            [("Prefer", "count=exact")]
+            mempty
+          `shouldRespondWith`
+            ""
+            { matchStatus  = 204
+            , matchHeaders = [ matchHeaderAbsent hContentType
+                             , "Content-Range" <:> "*/3" ]
+            }
```
steve-chavez, line 74: Hm, this needs a `Prefer: tx=commit` with checking the body before and after to make sure the delete went through.

steve-chavez: Just adding `Prefer: tx=commit` would be good though. Those type of tests that check the body need a lot of code.

steve-chavez: I realize now that with `tx=commit` this would require a reset after the request so the idempotence tests can pass. Let's just leave this as it is.

steve-chavez (approved):

Issue #2144, opened 2022-02-01
Computed relationships - First-class computed/virtual columns - a generalization of foreign-key-based resource embedding


wolfgangwalther:
# Problem

This builds on some ideas we had earlier:

@steve-chavez [understood resource embedding as functions](https://github.com/PostgREST/postgrest/issues/915#issuecomment-776060792):

> So, currently we support embedding - or joining - with the embed(col1,col2) syntax. A join is an operator(according to relational algebra) and thus: a function.

And I [gave a practical example](https://github.com/PostgREST/postgrest/issues/915#issuecomment-780035727), showing the similarities:

> ```sql
> create function clients (projects) returns clients
> stable language sql as $$
>   select * from clients where id = $1.client_id
> $$;
> ```
> Because this function has only 1 argument of type projects it can be selected as a virtual column from the projects table. The url would look like this: `GET /projects?select=*,clients`. The result is the same as the embedding we provide: `GET /projects?select=*,clients(*)`.

@steve-chavez then suggested this as [the way forward to do custom embeddings](https://github.com/PostgREST/postgrest/issues/1179#issuecomment-851069227).

However, the current support for virtual columns is only implicit through [field notation](https://www.postgresql.org/docs/current/rowtypes.html#ROWTYPES-USAGE). This has several [drawbacks for embedding](https://github.com/PostgREST/postgrest/issues/915#issuecomment-780035727), i.e. the following can only be done through fk-based embedding right now:

> - ability to select only a subset of columns
> - nesting
> - performance: inlining of o2m and m2m queries (doesn't work with a real function, because the aggregation would prevent it)

Additionally it's currently impossible to use a virtual column that has the same name as a real column of the same relation - field notation will always choose the latter.

---

# Basic implementation

To solve these issues, I propose to introduce first-class support for virtual columns: When creating the schema cache, we should parse function definitions to detect virtual columns and then make them callable **the same way as resource embedding is currently done**.

The good thing: This is actually much simpler than it sounds. We need to add relationships based on function definitions in:

https://github.com/PostgREST/postgrest/blob/8934190d4359bdf3597339fd4da4b4b250c8dbb9/src/PostgREST/DbStructure.hs#L96

Those relationships would have a cardinality of `M2O` or `O2M` depending whether the function returns multiple rows or not. This would generate the right query for aggregation here:

[...]

PR #2419, opened 2022-08-10
Add computed relationships

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/2144

## TODO

- [x] Override detected relationships
- [x] Tests
  - [x] o2m and m2o relationship
  - [x] rpc
  - [x] mutations
  - [x] override detected relationships
  - [x] self relationships
- [x] Refactor
- [x] CHANGELOG

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -90,18 +90,36 @@ queryDbStructure schemas extraSearchPath prepared = do
   keyDeps <- SQL.statement (schemas, extraSearchPath) $ allViewsKeyDependencies prepared
   m2oRels <- SQL.statement mempty $ allM2ORels pgVer prepared
   procs   <- SQL.statement schemas $ allProcs pgVer prepared
+  cRels   <- SQL.statement mempty $ allComputedRels prepared
 
   let tabsWViewsPks = addViewPrimaryKeys tabs keyDeps
-      rels          = relsToMap $ addO2MRels $ addM2MRels tabsWViewsPks $ addViewM2ORels keyDeps m2oRels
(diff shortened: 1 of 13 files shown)


steve-chavez:
Should be ready. Overall I feel confident that this one doesn't break existing functionality since `ComputedRelationShip` is added as a new data constructor of `Relationship`.

src/PostgREST/DbStructure/Relationship.hs
```diff
@@ -26,6 +26,13 @@ data Relationship = Relationship
   , relTableIsView  :: Bool
   , relFTableIsView :: Bool
   }
+  | ComputedRelationShip
```
laurenceisla, line 29:
```suggestion
  | ComputedRelationship
```

src/PostgREST/DbStructure.hs
```diff
@@ -643,6 +662,46 @@ allM2ORels pgVer =
...
+      where relkind in ('v','r','m','f','p')
+    ),
+    computed_rels as (
+      select
+        p.pronamespace::regnamespace::text as schema,
+        p.proname as name,
+        split_part((p.proargtypes[0])::regtype::text, '.', 1) as rel_table_schema,
+        split_part((p.proargtypes[0])::regtype::text, '.', 2) as rel_table_name,
+        split_part(p.prorettype::regtype::text, '.', 1) as rel_ftable_schema,
+        split_part(p.prorettype::regtype::text, '.', 2) as rel_ftable_name,
```
laurenceisla, line 683: It looks like this will break when a table has a `.` in its name (probably it's the same with the schema name).

steve-chavez: Good catch. Fixed.

laurenceisla (approved):

PR #2426, opened 2022-08-12
Simplified tests where mutations need to be verified in the db

laurenceisla:
Some mutations tests (updates/deletes) need to verify if the change did happen in the databas, which makes them bulky and difficult to read. 

This PR moves that logic to `SpecHelper.hs` and uses it as a global function for the tests that need it.

### DIFF
--- a/test/spec/Feature/Query/PgSafeUpdateSpec.hs
+++ b/test/spec/Feature/Query/PgSafeUpdateSpec.hs
@@ -36,23 +36,15 @@ spec =
             { matchStatus  = 400 }
 
       it "allows full table update if a filter is present" $
-        verifyMutation "safe_update_items" tblDataBefore
-          [json|[
-            { "id": 1, "name": "updated-item", "observation": null }
-          , { "id": 2, "name": "updated-item", "observation": null }
-          , { "id": 3, "name": "updated-item", "observation": null }
-          ]|]
-          $
-          request methodPatch "/safe_update_items?id=gt.0"
-              [("Prefer", "tx=commit"), ("Prefer", "count=exact")]
-              [json| {"name": "updated-item"} |]
-            `shouldRespondWith`
-              ""
-              { matchStatus  = 204
-              , matchHeaders = [ matchHeaderAbsent hContentType
-                               , "Content-Range" <:> "0-2/3"
-                               , "Preference-Applied" <:> "tx=commit" ]
-              }
+        baseTable "safe_update_items" "id" tblDataBefore
+        `mutatesWith`
+        requestMutation methodPatch "/safe_update_items?id=gt.0" [json| {"name": "updated-item"} |]
+        `shouldMutateInto`
+        [json|[
+          { "id": 1, "name": "updated-item", "observation": null }
+        , { "id": 2, "name": "updated-item", "observation": null }
+        , { "id": 3, "name": "updated-item", "observation": null }
+        ]|]
 
     context "Full table delete" $ do
       it "does not delete and throws error if no condition is present" $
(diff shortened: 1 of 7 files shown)


test/spec/SpecHelper.hs
```diff
@@ -243,3 +244,20 @@ isErrorFormat s =
...
+-- | Follows these steps to verify if the table data changed in the db:
+--  * Verifies the table data in the db before the change
+--  * Does the mutation
+--  * Verifies that the table data changed in the db
+--  * Resets the table with the original data
+verifyMutation :: ByteString -> Value -> ResponseMatcher -> WaiExpectation m -> WaiSession m ()
+verifyMutation tblName dataBefore dataAfter mutation = do
+  get ("/" <> tblName) `shouldRespondWith` [json|#{dataBefore}|]
+  mutation
+  get ("/" <> tblName <> "?order=id") `shouldRespondWith` dataAfter
```
steve-chavez, line 257: `?order=id` being hidden is a bit surprising. How about specifying this in each request instead?

test/spec/Feature/Query/PgSafeUpdateSpec.hs
```diff
@@ -74,101 +66,56 @@ spec =
...
+          request methodDelete "/unsafe_delete_items"
+              [("Prefer", "tx=commit"), ("Prefer", "count=exact")]
+              mempty
+            `shouldRespondWith`
+              ""
+              { matchStatus  = 204
+              , matchHeaders = [ matchHeaderAbsent hContentType
+                               , "Content-Range" <:> "*/3"
+                               , "Preference-Applied" <:> "tx=commit" ]
+              }
```
steve-chavez, line 121: We're always specifying the `tx=commit` plus checking the `Prefernce-Applied`, could this be abstracted away?

I also note that in every test the `shouldRespondWith` is comparing to `""`.  Maybe we can ignore the status/headers and just focus on the result body? This should make the tests even shorter.

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -514,102 +477,56 @@ spec = do
...
+        $
+        request methodPatch "/limited_update_items_no_pk?order=ctid&limit=1"
+            [("Prefer", "tx=commit")]
+            [json| {"name": "updated-item"} |]
+          `shouldRespondWith`
+            ""
+            { matchStatus  = 204
+            , matchHeaders = [ matchHeaderAbsent hContentType
+                             , "Preference-Applied" <:> "tx=commit" ]
+            }
```
steve-chavez, line 532: With the above improvements it could look like this:
```haskell
    it "works on a table without a pk by ordering by 'ctid'" $
      verifyMutation "limited_update_items_no_pk?order=id" tblDataBefore
        methodPatch "/limited_update_items_no_pk?order=ctid&limit=1" [json| {"name": "updated-item"} |] $
        [json|[
          { "id": 1, "name": "updated-item" }
        , { "id": 2, "name": "item-2" }
        , { "id": 3, "name": "item-3" }
        ]|]
```

steve-chavez: Also an idea, since there are 3 steps to this check, perhaps we can use [infix operators](https://stackoverflow.com/questions/6354401/anyone-know-how-to-use-a-partially-applied-three-argument-function-infix-haskel) to make it clearer. Like:

```haskell
"limited_update_items_no_pk?order=id" tblDataBefore `mutationBefore` 
methodPatch "/limited_update_items_no_pk?order=ctid&limit=1" `mutationAfter`        
[json|[
          { "id": 1, "name": "updated-item" }
        , { "id": 2, "name": "item-2" }
        , { "id": 3, "name": "item-3" }
        ]|]
```

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -514,102 +477,50 @@ spec = do
...
+      baseTable "limited_update_items_no_pk" "order=id" tblDataBefore 
+      `mutatesWith`
+      requestMutation methodPatch "/limited_update_items_no_pk?order=ctid&limit=1"
+        [json| {"name": "updated-item"} |]
+      `shouldMutateInto`
+        [json|[
+          { "id": 1, "name": "updated-item" }
+        , { "id": 2, "name": "item-2" }
+        , { "id": 3, "name": "item-3" }
+        ]|]
```
laurenceisla, line 526: I managed to make infix work by adding a `baseTable` function (only two parameters must be used for the infix to work this way). The `BaseTable` datatype could be used directly though.

steve-chavez: Looks much better! So `baseTable` takes a table name plus an `id` to ensure the order of the result. I believe the `order=` can be omitted in that case.

steve-chavez (approved):

PR #2427, opened 2022-08-15
Nix: allow building from pure-eval mode

mvnetbiz:
I am using PostgREST in a flake, which uses Nix pure evaluation mode by default, which cannot use builtins.currentSystem.
Also updated nix-gitignore to include this fix to stop attempting to read ~/.gitignore https://github.com/hercules-ci/gitignore.nix/pull/58

The type of default.nix is changed from an attrset to a function. nix-build still works (nix-build . -A postgrestStatic), but if other people import default.nix in their own expression maybe default.nix should be renamed to pure.nix, and default.nix change to
```nix
import ./pure.nix { }
```
??

https://github.com/hercules-ci/gitignore.nix/pull/58

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,3 +1,5 @@
+{ system ? builtins.currentSystem }:
+
 let
   name =
     "postgrest";
@@ -42,7 +44,7 @@ let
 
   # Evaluated expression of the Nixpkgs repository.
   pkgs =
-    import nixpkgs { inherit overlays; };
+    import nixpkgs { inherit overlays system; };
 
   postgresqlVersions =
     [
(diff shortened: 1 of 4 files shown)


steve-chavez:
The change looks simple however it's causing the following error on CI:

```
Run nix-shell --run postgrest-test-doctests
error: value is a function while a set was expected

       at /home/runner/work/postgrest/postgrest/shell.nix:12:[4](https://github.com/PostgREST/postgrest/runs/7835051291?check_suite_focus=true#step:6:5):

           11| }:
           12| let
             |    ^
           13|   postgrest =
```

Any idea why?

mvnetbiz:
There was an import of default.nix in shell.nix, I added an empty attrset parameter.

PR #2435, opened 2022-08-18
bump version to v10.0.0

steve-chavez:


steve-chavez:
Will let the `postgrest-release` command do this instead.

Issue #1984, opened 2021-10-18
Detect one-to-one mappings in joins to avoid sub-arrays


colearendt:
### Description of issue

In some table setups, there are unique keys on both tables that establish / enforce a one-to-one mapping. Right now PostgREST joins these as an array with a single entry: `"child": [{}]`

It would be nice to have the option to join one-to-one tables as `"child": {}`

Discussed here: https://github.com/PostgREST/postgrest/pull/1978#discussion_r731153211

https://github.com/PostgREST/postgrest/blob/42d7cf8ae5b966dd901bf3695732423a9e9e2e52/src/PostgREST/DbStructure/Relationship.hs#L36-L41

madeleineostoja:
Is there any movement or consensus on implementing this? Would be great to get a PR going

steve-chavez:
Sure, just opened https://github.com/PostgREST/postgrest/pull/2439.

---

Adding this one later will be pain because it's strictly a breaking change(since the shape of the json output will change) and it would require a new major version bump. Users will need to revise their relationships on this release anyway(because m2m relationships changed) so it's apt to include it now.

PR #2439, opened 2022-08-19
feat: add one to one relationship for resource embedding

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/1984. One to one relationships are detected when:

- There's a foreign key which is also a primary key:

```sql
-- one-to-one rel between students and students_info
CREATE TABLE test.students(
  id int
, code text
, name text
, primary key(id, code)
);

CREATE TABLE test.students_info(
  id int
, code text
, address text
, primary key(id, code)
, foreign key (id, code) references test.students(id, code)
);
```

- There's a foreign key with an unique constraint

```sql
-- one-to-one rel between country and capital
CREATE TABLE test.country(
  id int primary key
, name text
);

CREATE TABLE test.capital(
  id int primary key
, name text
, country_id int unique
, foreign key (country_id) references test.country(id)
);
```

([There's a blog post by Vertabelo that makes this clearer](https://vertabelo.com/blog/one-to-one-relationship-in-database))

This works across the board(tables, views, rpc, mutations, computed rels can override them). O2O rels are a refinement of M2O rels, so no news rels are added and there is no potential for new disambiguation conflicts.

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -88,12 +88,12 @@ queryDbStructure schemas extraSearchPath prepared = do
   pgVer   <- SQL.statement mempty pgVersionStatement
   tabs    <- SQL.statement schemas $ allTables pgVer prepared
   keyDeps <- SQL.statement (schemas, extraSearchPath) $ allViewsKeyDependencies prepared
-  m2oRels <- SQL.statement mempty $ allM2ORels pgVer prepared
+  m2oRels <- SQL.statement mempty $ allM2OandO2ORels pgVer prepared
   procs   <- SQL.statement schemas $ allProcs pgVer prepared
   cRels   <- SQL.statement mempty $ allComputedRels prepared
 
   let tabsWViewsPks = addViewPrimaryKeys tabs keyDeps
-      rels          = addO2MRels $ addM2MRels tabsWViewsPks $ addViewM2ORels keyDeps m2oRels
+      rels          = addInverseRels $ addM2MRels tabsWViewsPks $ addViewM2OAndO2ORels keyDeps m2oRels
 
   return $ removeInternal schemas $ DbStructure {
       dbTables = tabsWViewsPks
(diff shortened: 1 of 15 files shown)


src/PostgREST/Error.hs
```diff
@@ -188,6 +188,10 @@ compressedRel Relationship{..} =
             "cardinality" .= ("many-to-one" :: Text)
           , "relationship" .= (cons <> " using " <> qiName relTable <> fmtEls (fst <$> relColumns) <> " and " <> qiName relForeignTable <> fmtEls (snd <$> relColumns))
           ]
+        O2O cons relColumns -> [
+            "cardinality" .= ("one-to-one" :: Text)
+          , "relationship" .= (cons <> " using " <> qiName relTable <> fmtEls (fst <$> relColumns) <> " and " <> qiName relForeignTable <> fmtEls (snd <$> relColumns))
```
steve-chavez, line 193: No idea how to appease codecov here. A disambiguation error usually could be triggered by defining an M2M plus M2O to the same table, but since it's a O2O, a junction table in between the tables won't define a M2M.

laurenceisla: Maybe this helps. I get a disambiguation error by doing this:

```sql
create table api.first (
  id int primary key,
  name text,
  second_id_1 int references api.second unique,
  second_id_2 int references api.second unique
);

create table api.second (
  id int primary key,
  name text
);
```

This query returns the error:

```http
GET /first?select=second(*)
```

steve-chavez: Awesome :100: . Also good thing that it can be disambiguated with `/first?select=second!second_id_1(*)`

src/PostgREST/DbStructure.hs
```diff
@@ -634,22 +642,45 @@ tablesSqlQuery getAll pgVer =
       )|]
     relIsPartition = if pgVer >= pgVersion100 then " AND not c.relispartition " else mempty
 
-allM2ORels :: PgVersion -> Bool -> SQL.Statement () [Relationship]
-allM2ORels pgVer =
+
+-- | Gets many-to-one relationships and one-to-one(O2O) relationships, which are a refinement of the many-to-one's
+allM2OandO2ORels :: PgVersion -> Bool -> SQL.Statement () [Relationship]
+allM2OandO2ORels pgVer =
```
laurenceisla, line 648: Just to comment that this query runs on pg 15 beta 3 and builds correctly. Did not check the code, though.

PR #2444, opened 2022-08-24
Upgrade hasql-pool to forked 0.7.2

robx:
This is a take on #2391 that uses our fork of hasql-pool with flushing support.


robx:
I would ike #2445 to be merged first, because I expect that there's something to be worked out between the plain cabal build and the nix build. (I.e., I'll want to add a `cabal.project` file pointing at the fork, without that interfering with the nix build which also uses cabal.)

And then this PR should wait for https://github.com/PostgREST/hasql-pool/pull/2 to be merged, to ensure (a) that change is reviewed and (b) keep things somewhat clean, by only ever depending on main branch commits of the forked repo.

robx:
This should be good to go now.

I was wondering whether I should write up some guidelines on dealing with forked dependencies such as this one (where? a new `DEVELOPMENT.md`?), but maybe let's see how this plays out first and follow up then?

steve-chavez:
> I was wondering whether I should write up some guidelines on dealing with forked dependencies such as this one (where? a new DEVELOPMENT.md?), but maybe let's see how this plays out first and follow up then?

Yeah, I see new activity on `hasql-pool` so let's see if we still maintain our forked dependencies for a longer time.

steve-chavez (approved):

robx:
Chances are we'll merge https://github.com/nikita-volkov/hasql-pool/pull/16 soon enough with a slightly simpler interface (`PoolIsReleasedUsageError` goes away, in particular.) But probably it's best to just go ahead with this and update/simplify later, what do you think?

steve-chavez:
> Chances are we'll merge https://github.com/nikita-volkov/hasql-pool/pull/16 soon enough with a slightly simpler interface (PoolIsReleasedUsageError goes away, in particular.)

Nice :tada: 

> But probably it's best to just go ahead with this and update/simplify later, what do you think?

Agree.

bhupixb:
Is this ideal to remove pool timeout?
We have a sudden spike in traffic at different times of the day in most of our workloads, so we have configured the DB Pool size to a higher value(100) so that by the time HPA scales up the pods, the existing pod have enough connections available to serve requests. 
After the traffic slows down, we want the ideal connections to be closed after some time.

But if we remove this flag, then we will be wasting a lot of resources on the Postgres server per connection unnecessarily where there's minimal/no traffic. Does the upgraded `hasql-pool` automatically closes ideal connections?

steve-chavez:
@bhupixb Could you test the pool behavior with `idle_session_timeout` as mentioned above by Wolfgang?

robx:
Some quick random ideas on this:
- you could actively flush the pool every couple of minutes using `SIGUSR1`, e.g. via cron
- if we wanted to implement this in postgrest without touching hasql-pool, we might do the same, i.e. periodically release the pool

But if `idle_session_timeout` works, that's probably preferable, because it also covers scenarios such as the postgrest server being disconnected from the postgresql server.

bhupixb:
> @bhupixb Could you test the pool behavior with `idle_session_timeout` as mentioned above by Wolfgang?
This will work IMO(haven't tested yet), but an RDS can have multiple applications running their DB on it, so setting the `idle_session_timeout` RDS level is a bit less flexible in our case. 
Because not all applications want to clean up idle connections as frequently as we do(because we know in Postgrest we can clean up).

> * you could actively flush the pool every couple of minutes using `SIGUSR1`, e.g. via cron
@robx Can you please elaborate a little here, say my postgrest container is running and I added a sidecar to close idle connections. So what command should I send to postgrest for doing so?

robx:
> @robx Can you please elaborate a little here, say my postgrest container is running and I added a sidecar to close idle connections. So what command should I send to postgrest for doing so?

Here: https://postgrest.org/en/stable/schema_cache.html#schema-cache-reloading. A side-effect of schema-cache reloading is that connections are released.

steve-chavez:
>  We'd expect the idle connections to be closed for sure - but what we don't know is whether PostgREST will restore all those idle connections immediately, or whether it will wait reconnecting until it needs them.

Just tried that, postgrest recovers but it gets noisy on the logs and the `LISTEN` channel gets killed every time.

```
PGRST_DB_ANON_ROLE='postgrest_test_anonymous' postgrest-with-postgresql-14   postgrest-run

04/Nov/2022:17:06:33 -0500: Attempting to connect to the database...
04/Nov/2022:17:06:33 -0500: Connection successful
04/Nov/2022:17:06:33 -0500: Listening on port 3000
04/Nov/2022:17:06:33 -0500: Config reloaded
04/Nov/2022:17:06:33 -0500: Listening for notifications on the pgrst channel
04/Nov/2022:17:06:33 -0500: Schema cache loaded
FATAL:  terminating connection due to idle-session timeout
04/Nov/2022:17:06:38 -0500: Retrying listening for notifications on the pgrst channel..
04/Nov/2022:17:06:38 -0500: Attempting to connect to the database...
04/Nov/2022:17:06:38 -0500: Connection successful
04/Nov/2022:17:06:38 -0500: Config reloaded
04/Nov/2022:17:06:38 -0500: Listening for notifications on the pgrst channel
04/Nov/2022:17:06:38 -0500: Schema cache loaded
FATAL:  terminating connection due to idle-session timeout
04/Nov/2022:17:06:43 -0500: Retrying listening for notifications on the pgrst channel..
04/Nov/2022:17:06:43 -0500: Attempting to connect to the database...
04/Nov/2022:17:06:43 -0500: Connection successful
04/Nov/2022:17:06:43 -0500: Listening for notifications on the pgrst channel
04/Nov/2022:17:06:43 -0500: Config reloaded
04/Nov/2022:17:06:43 -0500: Schema cache loaded
```

robx:
> Surely, in that case we'd want PostgREST to just delay the request until the connection is up again?

I tend to agree. It seems reasonable (and mostly unavoidable -- we can't know if an old connection is still good) to automatically retry for certain error classes.

It *might* make sense to do this in hasql-pool directly, not sure there.

robx:
>     * Connections that are found dead should return their resources to the pool (not sure about the right terminoloy here). Right now it seems like hasql-pool still thinks those are open and makes them available to the pool user? Until it runs a query and then realizes the connection loss.

You won't be able to tell that the connection is closed without writing to or reading from the socket, which needs to happen actively. It would be possible to preemptively try to read from the connection to see if the server terminated the connection abnormally (compare https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.6.7.11), but that's quite complicated with little fundamental benefit: Even if we have some reaper thread waiting for connections to terminate in order to remove them from the pool, there's a window where the server terminates the connection just after we took it from the pool, so for correct behaviour, we'd have to implement the retry anyway.

The straightforward way (what feels like the "right" way to me), is to expect certain errors, such as a dead connection, and retry silently.

>     * Whenever a new connection is made, because there is no idle connection in the pool, the request should wait for that connection to happen. I'm pretty sure this is the case already, because when I start with 1 connection and run concurrent requests.. the number of connections increases without any of those error messages.

Yes, I think that's correct.

> > we can't know if an old connection is still good
> 
> Why is that? When the connection is terminated via `idle_session_timeout` I would expect the client to realize this immediately, no?

As noted above, only when reading from the socket. And then this still only covers certain active shutdown scenarios. It can also happen that network disruption causes a TCP socket to be in a state where we only see that the connection is broken once we write to it.

So in general, no, the client doesn't realize this immediately. If it does something like `select()` on the socket, it *could* realize immediately when a shutdown message is received, but that would still require some thread actively waiting for such notifications.

(There's a number of connection settings that can play into this, e.g. around TCP keepalives, see https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS.)

PR #2449, opened 2022-08-26
Add db-pool-acquisition-timeout

### LINKED ISSUE #2348: Add `db-pool-query-wait-timeout` config

robx:
From the commit message:

```
feat: introduce pool acquisition timeout (fixes #2348)

The configuration option db-pool-acquisition-timeout
specifies the time in seconds to wait for the pool to
free up a connection slot. Otherwise, a 504 error is
returned. By default, there is no timeout.
```


src/PostgREST/CLI.hs
```diff
@@ -148,8 +148,8 @@ exampleConfigFile =
       |## Number of open connections in the pool
       |db-pool = 10
       |
-      |## Time to live, in seconds, for an idle database pool connection
-      |db-pool-timeout = 3600
+      |## Time in seconds to wait to acquire a slot from the connection pool
+      |# db-pool-acquisition-timeout = 10
```
steve-chavez, line 152: Yeah, agree with the config name.

test/io/test_io.py
```diff
@@ -993,6 +993,37 @@ def sleep(i=i):
...
+def test_pool_acquisition_timeout(defaultenv, metapostgrest):
+    "Verify that PGRST_DB_POOL_ACQUISITON_TIMEOUT times out when the pool is empty"
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_POOL": "1",
+        "PGRST_DB_POOL_ACQUISITION_TIMEOUT": "1",  # 1 second
+    }
+
+    with run(env=env) as postgrest:
```
steve-chavez, line 1005: run has a `no_pool_connection_available` parameter, could that be used for this test instead of opening a thread?

Mostly to be consistent with other tests already using that.

src/PostgREST/AppState.hs
```diff
@@ -97,7 +97,7 @@ destroy = destroyPool
 
 initPool :: AppConfig -> IO SQL.Pool
 initPool AppConfig{..} =
-  SQL.acquire configDbPoolSize Nothing $ toUtf8 configDbUri
+  SQL.acquire configDbPoolSize ((1000000 *) <$> configDbPoolAcquisitionTimeout) $ toUtf8 configDbUri
```
steve-chavez, line 100: Just for clarity
```suggestion
  let microSecond = 1000000 in
  SQL.acquire configDbPoolSize ((microSecond *) <$> configDbPoolAcquisitionTimeout) $ toUtf8 configDbUri
```

test/io/test_io.py
robx: Unfortunately, that parameter doesn't actually do anything, compare #2456 #2457. It's a good point that we should be using this though. It shouldn't be too hard to fix it, using a similar approach as here.

src/PostgREST/AppState.hs
robx: I've addressed this (with slightly different naming, `microSecond = 1000000` didn't feel quite right because that's not literally a microsecond, but rather a second in microseconds...).

test/io/test_io.py
robx: I've rebased this on top of #2457 (which now has a fix for the issue), should work now!

steve-chavez (approved):

PR #2454, opened 2022-08-29
upgrade to hasql-1.6, hasql-pool-0.8

robx:
- hasql-pool-0.8
  * 'release' now flushes the pool, and no longer destroys it;
    'PoolIsReleasedUsageError' is gone compared to our fork
  * now supports an acquisition timeout, which we don't use
    yet
- hasql-1.6
  introduces a position parameter to ServerError, which we
  ignore

hasql-notifications has an upper bound on hasql-pool, which
we override for the moment, pending
  https://github.com/diogob/hasql-notifications/issues/12


steve-chavez:
Hm, on CI I see we fail building with `stack` for all non-linux platforms. Is that expected?

robx:
> Hm, on CI I see we fail building with `stack` for all non-linux platforms. Is that expected?

Should be fixed now.

src/PostgREST/Error.hs
```diff
@@ -259,13 +259,14 @@ instance JSON.ToJSON SQL.CommandError where
...
+
 pgErrorStatus :: Bool -> SQL.UsageError -> HTTP.Status
 pgErrorStatus _      (SQL.ConnectionUsageError _) = HTTP.status503
-pgErrorStatus _      SQL.PoolIsReleasedUsageError = HTTP.status500
+pgErrorStatus _      SQL.AcquisitionTimeoutUsageError = HTTP.status504
 pgErrorStatus _      (SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _)))      = HTTP.status503
 pgErrorStatus authed (SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ResultError rError))) =
   case rError of
-    (SQL.ServerError c m _ _) ->
+    (SQL.ServerError c m _ _ _) ->
```
steve-chavez, line 269: Ok so we don't need this parameter as it looks more fit for detecting an error when sending raw SQL queries.

https://hackage.haskell.org/package/hasql-1.6.0.1/docs/Hasql-Session.html#t:ResultError

([Maybe](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Maybe.html#t:Maybe) [Int](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Int.html#t:Int))	
Position. Error cursor position as an index into the original statement string. Positions are measured in characters not bytes.

steve-chavez (approved):

src/PostgREST/Error.hs
robx: Yes, I had it in first but it resulted mostly in `null` values in the tests, and where it did give a number it didn't seem particularly useful information, should have mentioned that.

PR #2472, opened 2022-09-13
Add error message when Prefer: count=exact is used and offset is out of bounds

### LINKED ISSUE #2470: Empty error when range is not satisfiable
No error message on the response body on

```bash
$ PGRST_DB_ANON_ROLE='postgrest_test_anonymous' postgrest-with-postgresql-14   postgrest-run

curl -i "http://localhost:3000/items?offset=16" -H "Prefer: count=exact"
HTTP/1.1 416 Requested Range Not Satisfiable

[]
```

The above doesn't happen when there's no `count=exact` because here we don't know the max range

```bash
curl -i "http://localhost:3000/items?offset=16"
HTTP/1.1 200 OK

[]
```

When the limit is negative there is an error message, with or without count

```bash
curl -i "http://localhost:3000/items?limit=-1"

HTTP/1.1 416 Requested Range Not Satisfiable
{"code":"PGRST103","details":null,"hint":null,"message":"HTTP Range error"}
```

Since we cannot always reply with an error, perhaps all wrong ranges should reply with 200? Not sure, but adding an error message for all 416 is definitely better.

laurenceisla:
Closes #2470

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -291,9 +291,12 @@ handleRead headersOnly identifier context@RequestContext{..} = do
             )
           ]
           ++ contentTypeHeaders context
+        rsOrErrBody = if status == HTTP.status416
+          then Error.errorPayload (Error.ApiRequestError ApiRequestTypes.InvalidRange)
+          else LBS.fromStrict rsBody
 
       failNotSingular iAcceptMediaType rsQueryTotal . response status headers $
-        if headersOnly then mempty else LBS.fromStrict rsBody
+        if headersOnly then mempty else rsOrErrBody
 
     RSPlan plan ->
       pure $ Wai.responseLBS HTTP.status200 (contentTypeHeaders context) $ LBS.fromStrict plan
@@ -501,6 +504,9 @@ handleInvoke invMethod proc context@RequestContext{..} = do
       let
         (status, contentRange) =
           RangeQuery.rangeStatusHeader iTopLevelRange rsQueryTotal rsTableTotal
+        rsOrErrBody = if status == HTTP.status416
+          then Error.errorPayload (Error.ApiRequestError ApiRequestTypes.InvalidRange)
+          else LBS.fromStrict rsBody
 
       failNotSingular iAcceptMediaType rsQueryTotal $
         if Proc.procReturnsVoid proc then
@@ -508,7 +514,7 @@ handleInvoke invMethod proc context@RequestContext{..} = do
         else
           response status
             (contentTypeHeaders context ++ [contentRange])
-            (if invMethod == InvHead then mempty else LBS.fromStrict rsBody)
+            (if invMethod == InvHead then mempty else rsOrErrBody)
 
     RSPlan plan ->
       pure $ Wai.responseLBS HTTP.status200 (contentTypeHeaders context) $ LBS.fromStrict plan
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -292,7 +292,8 @@ handleRead headersOnly identifier context@RequestContext{..} = do
           ]
           ++ contentTypeHeaders context
         rsOrErrBody = if status == HTTP.status416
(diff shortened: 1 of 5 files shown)


src/PostgREST/App.hs
```diff
@@ -291,9 +291,12 @@ handleRead headersOnly identifier context@RequestContext{..} = do
             )
           ]
           ++ contentTypeHeaders context
+        rsOrErrBody = if status == HTTP.status416
+          then Error.errorPayload (Error.ApiRequestError ApiRequestTypes.InvalidRange)
+          else LBS.fromStrict rsBody
```
steve-chavez, line 296: I see the 416 error body on GET and RPC but POST/PATCH also have some range logic, would it make sense to handle the 416 in the other methods too?

laurenceisla: I'm a bit confused about the [RFC](https://httpwg.org/specs/rfc9110.html#field.range) for `Range`:

> A server MUST ignore a Range header field received with a request method that is unrecognized or for which range handling is not defined. For this specification, [GET](https://httpwg.org/specs/rfc9110.html#GET) is the only method for which range handling is defined.

I interpret it as if we allow Range for PATCH, for instance, we do not need to ignore it, right? Then I'd agree that the same 416 error logic should be applied for these methods. 

---

On the other hand, the [416 status error](https://httpwg.org/specs/rfc9110.html#status.416) says:

> The 416 ([Range](https://httpwg.org/specs/rfc9110.html#field.range) Not Satisfiable) status code indicates that the set of ranges in the request's Range header field ([Section 14.2](https://httpwg.org/specs/rfc9110.html#field.range)) has been rejected...

Would this mean that the 416 error raised by  `offset` in the query string is not accurate since an error with `Range` is expected but no such header is present? The new tests I added would not be OK, then.

steve-chavez: I think the problem stems from using the same logic for `limit/offset` and `Range`. One has postgresql semantics which don't err while the HTTP semantic can err.

That being said, it's useful for web clients to get an error. Also from the [416 status error](https://httpwg.org/specs/rfc9110.html#status.416):

> or because the client has requested an excessive number of small or overlapping ranges (a potential denial of service attack).

Which is interesting because that could be extrapolated to: reply with an error whenever we have a `db-max-rows=1000` and a client requests a `limit=2000`. Right now we don't err and just prefer the 1000 limit.

---

I think the PR is good as it is, we can open a new issue for deciding what to do with the Range semantics on `limit/offset`.

src/PostgREST/Error.hs
```diff
@@ -92,7 +92,7 @@ instance JSON.ToJSON ApiRequestError where
     "hint"    .= JSON.Null]
   toJSON InvalidRange = JSON.object [
     "code"    .= ApiRequestErrorCode03,
-    "message" .= ("HTTP Range error" :: Text),
+    "message" .= ("Requested range not satisfiable" :: Text),
```
steve-chavez, line 95: Could we further clarify this with a message like `an offset with 1000 was requested but there are only 500 rows`?

laurenceisla: I added this to the "details". It also kinda forced me to add details for other 416 errors like negative limits and where the lower boundary is greater than the upper one in the Range header.

steve-chavez (approved):

src/PostgREST/App.hs
```diff
@@ -291,9 +291,13 @@ handleRead headersOnly identifier context@RequestContext{..} = do
             )
           ]
           ++ contentTypeHeaders context
+        rsOrErrBody = if status == HTTP.status416
+          then Error.errorPayload $ Error.ApiRequestError $ ApiRequestTypes.InvalidRange
+            $ ApiRequestTypes.OutOfBounds (show $ RangeQuery.rangeOffset iTopLevelRange) (maybe "0" show total)
```
steve-chavez, line 296: Giving this a second look. It's a really long function chain.. could this logic be abstracted in Errors.hs?

PR #2476, opened 2022-09-14
Fix embedding with computed relationships not working with schema names with special characters

### LINKED ISSUE #2473: Embedding with computed relationships breaks when the schema name has special characters
### Environment

* PostgreSQL version: 13.4, 13.8, 14.5
* PostgREST version: v10.0.0
* Operating system: Ubuntu 22.04

### Description of issue

With the snippet of
```sql
create schema "api";
create table "api".premieres (
  id integer primary key,
  location text,
  "date" date,
  film_id integer
);

insert into "api".premieres values(2001, 'loc 1', '2022-01-01', 1);
insert into "api".premieres values(2002, 'loc 2', '2022-02-01', 2);
insert into "api".premieres values(2003, 'loc 3', '2022-03-01', 3);
insert into "api".premieres values(2004, 'loc 4', '2022-04-01', 4);
insert into "api".premieres values(2005, 'loc 5', '2022-05-01', 5);

create table "api".films(
    id integer  primary key,
    director_id integer,
    title text,
    year integer,
    rating decimal(3,2),
    language text
);

insert into "api".films values(1, 1, 'film 1', 2010, 4.3, 'EN');
insert into "api".films values(2, 1, 'film 2', 2011, 4.9, 'EN');
insert into "api".films values(4, 3, 'film 4', 2012, 4.1, 'CN');
insert into "api".films values(5, 3, 'film 5', 2013, 4.6, 'CN');

create function "api".computed_films("api".premieres) returns setof "api".films rows 1 as $$
  select * from "api".films where id = $1.film_id
$$ stable language sql;
```
a query of `curl -H 'Accept-Profile: api' 'localhost:3000/premieres?select=*,computed_films(*)'` just works as expected.

If we change the schema name from `api` to `api-v1`, the query `curl -H 'Accept-Profile: api-v1' 'localhost:3000/premieres?select=*,computed_films(*)'` suddenly breaks with the `PGRST200` message.

A schema name of `api+v2` breaks the rel detection as well.

laurenceisla:
Closes #2473

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -710,13 +710,13 @@ allComputedRels =
     ),
     computed_rels as (
       select
-        p.pronamespace::regnamespace::text as schema,
-        p.proname::text                    as name,
-        arg_schema.nspname::text           as rel_table_schema,
-        arg_name.typname::text             as rel_table_name,
-        ret_schema.nspname::text           as rel_ftable_schema,
-        ret_name.typname::text             as rel_ftable_name,
-        p.prorows = 1                      as single_row
+        (parse_ident(p.pronamespace::regnamespace::text))[1] as schema,
+        p.proname::text          as name,
+        arg_schema.nspname::text as rel_table_schema,
+        arg_name.typname::text   as rel_table_name,
+        ret_schema.nspname::text as rel_ftable_schema,
+        ret_name.typname::text   as rel_ftable_name,
+        p.prorows = 1            as single_row
       from pg_proc p
         join pg_type      arg_name   on arg_name.oid = p.proargtypes[0]
         join pg_namespace arg_schema on arg_schema.oid = arg_name.typnamespace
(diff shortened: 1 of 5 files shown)


src/PostgREST/DbStructure.hs
```diff
@@ -710,13 +710,13 @@ allComputedRels =
...
     computed_rels as (
       select
-        p.pronamespace::regnamespace::text as schema,
-        p.proname::text                    as name,
-        arg_schema.nspname::text           as rel_table_schema,
-        arg_name.typname::text             as rel_table_name,
-        ret_schema.nspname::text           as rel_ftable_schema,
-        ret_name.typname::text             as rel_ftable_name,
-        p.prorows = 1                      as single_row
+        (parse_ident(p.pronamespace::regnamespace::text))[1] as schema,
```
laurenceisla, line 713: I could not find another way of removing the double quote without using a join to `pg_namespace`.

steve-chavez (approved):

src/PostgREST/DbStructure.hs
steve-chavez: That one is pretty good actually. TIL about `parse_ident`

PR #2489, opened 2022-09-27
Control user rights

jyritali:
Control if user had rights to INSERT, UPDATE OR DELETE on table

Please add releveant entry to changelog

### DIFF
--- a/src/PostgREST/DbStructure.hs
+++ b/src/PostgREST/DbStructure.hs
@@ -595,29 +595,41 @@ tablesSqlQuery getAll pgVer =
     d.description AS table_description,
     c.relkind IN ('v','m') as is_view,
     (
+      (
       c.relkind IN ('r','p')
+      AND exists(SELECT * FROM information_schema.table_privileges where table_name=c.relname and privilege_type='INSERT' and grantee=current_user)
+      )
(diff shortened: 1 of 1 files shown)


jyritali:
My PostgRest usage problem, is that  I Wan't to allow only Select (GET) commands on tables to PostgRest.

jyritali:
Sry i'm not so good in Haskell or Phyton to write test for it.

steve-chavez:
> My PostgRest usage problem, is that I Wan't to allow only Select (GET) commands on tables to PostgRest.

Can you elaborate? Is this only for filtering OpenAPI?

jyritali:
> > My PostgRest usage problem, is that I Wan't to allow only Select (GET) commands on tables to PostgRest.
> 
> Can you elaborate? Is this only for filtering OpenAPI?
Yes I can, but this is needed for views to, because you can insert data trough view.

Yes, I don't want to postgRest shows POST, DELETE and PATCH operations, when postgRest postgres role not have permissionts to do these operations.
some pictures:
Swagger shows:
image
postgrest rights to table:
image

When postgRest have only select right to table, then postgRest should show only GET operation for this table.

PR #2531, opened 2022-10-26
Avoid infinite recursion when loading schema cache with self-referencing views

### LINKED ISSUE #2283: Views with infinite recursion causes a crashloop
### Problem

A broken view with an infinite recursion will cause us to crashloop while making PostgreSQL consuming a lot of memory/CPU.

So a view with infinite recursion It's a really edge case. To reproduce, first we have to create a valid view:

```sql
create view test.view_projects as 
select * from test.projects;
```

Then replace its body using the same view name, creating the infinite recursion:

```sql
create or replace view test.view_projects as 
select * from test.view_projects;
```

```sql
select * from test.view_projects;
ERROR:  infinite recursion detected in rules for relation "view_projects"
```

This will make the following query to go into an infinite loop:

https://github.com/PostgREST/postgrest/blob/2be63b36d6e051611f800ba29577346267851a74/src/PostgREST/DbStructure.hs#L638-L802

### Proposal

Tune the query to recurse a maximum number of times. 

@wolfgangwalther Do you think that would be possible?

wolfgangwalther:
Based on #2357 and #2526 - only the last commit is relevant here.

Fixes #2283.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2518, Fix a regression when embedding views where base tables have a different column order for FK columns - @wolfgangwalther
  - #2458, Fix a regression with the location header when inserting into views with PKs from multiple tables - @wolfgangwalther
  - #2356, Fix a regression in openapi output with mode follow-privileges - @wolfgangwalther
+ - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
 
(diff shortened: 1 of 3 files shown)


steve-chavez:
4a61d7cf9468ee221cb9309a4f9695dbe79c4ebd LGTM.

As I mentioned on the comment [here](https://github.com/PostgREST/postgrest/pull/2357#issuecomment-1292493816) - could you remove #2526 from this PR? Then it can be merged.

PR #2535, opened 2022-10-26
Avoid unnecessary UPDATE on conflicted column when upserting

### LINKED ISSUE #2446: Unnecessary UPDATE on PK col on upsert
From https://gitter.im/begriffs/postgrest?at=6302565ef4d7a323de64aa37

I have found that upserting {id: 42, comment: 'foo'} into

```sql
create table mytable (
  id bigserial primary key,
  comment text);
```

generates this following [pseudo]query

```sql
insert into mytable (id, comment) values (request.id, request.comment)
on conflict (id) do update id=excluded.id, comment=excluded.comment;
```

the do `update id=excluded.id` part is causing issues for me as i wanted to disallow updating the id once inserted so i through simple

```sql
GRANT select, insert(id, comment), update(comment) on mytable to api
```

would work.
With the above permissions and set up i can do INSERT followed by UPDATE as expected, but it turns out that i am getting 403 for upsert.

I tend to think this is a bug, as there is no real reason to include the do update id=excluded.id if the id is the default constraint. Or am i missing something?

wolfgangwalther:
This is supposed to fix #2446.

I'm not sure whether we should go ahead, though. For one, it's a breaking change, so we shouldn't merge it right away. That's because of this test failing:

```
  test/spec/Feature/Query/UpsertSpec.hs:96:13:
  1) Feature.Query.UpsertSpec.UPSERT, with POST, when Prefer: resolution=merge-duplicates is specified, succeeds if the table has only PK cols and no other cols
       body mismatch:
         expected: [{"id":1},{"id":2},{"id":4}]
         but got:  [{"id":4}]
```

results for `Prefer: return=representation` will be different, because fewer rows are actually updated - and thus returned.

---

Furthermore, the change doesn't seem to be as simple as assumed, because now two other tests fail:

```
  test/spec/Feature/Query/UpsertSpec.hs:389:13:
  2) Feature.Query.UpsertSpec.UPSERT, with PUT, Updating row, succeeds if the table has only PK cols and no other cols
       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: [{"id":1}]
         but got:  {"code":"PGRST115","details":null,"hint":null,"message":"Payload values do not match URL in primary key column(s)"}

  To rerun use: --match "/Feature.Query.UpsertSpec/UPSERT/with PUT/Updating row/succeeds if the table has only PK cols and no other cols/"

  test/spec/Feature/RpcPreRequestGucsSpec.hs:65:9:
  3) Feature.RpcPreRequestGucsSpec, GUC headers on all methods via pre-request, succeeds setting the headers on PUT
       status mismatch:
         expected: 204
         but got:  400
       unexpected header: Content-Type
       missing header:
         X-Custom-Header: mykey=myval
       the actual headers were:
       body mismatch:
         expected: ""
         but got:  "{\"code\":\"PGRST115\",\"details\":null,\"hint\":null,\"message\":\"Payload values do not match URL in primary key column(s)\"}"
```

Since this is not a regression, but potentially a breaking change, we should push this back after the next patch release for sure - if we decide to do it at all.

### DIFF
--- a/src/PostgREST/Query.hs
+++ b/src/PostgREST/Query.hs
@@ -145,7 +145,6 @@ actionQuery (DbCrud plan@MutateReadPlan{mrMutation=MutationUpdate, ..}) conf api
 
 actionQuery (DbCrud plan@MutateReadPlan{mrMutation=MutationSingleUpsert, ..}) conf apiReq _ _ = do
   resultSet <- writeQuery mrReadPlan mrMutatePlan mrMedia mrHandler apiReq conf
-  failPut resultSet
   optionalRollback conf apiReq
   pure $ DbCrudResult plan resultSet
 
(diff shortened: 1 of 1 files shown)


steve-chavez:
> Since this is not a regression, but potentially a breaking change, we should push this back after the next patch release for sure - if we decide to do it at all

Agree. I didn't expect the patch to be complicated.

(been wrong on beginner tags many times now)

steve-chavez:
Linking this [comment](https://github.com/PostgREST/postgrest/issues/2446#issuecomment-1427354403), an alternative solution would be to UPDATE according to the role privileges.

PR #2536, opened 2022-10-27
fix: embedding computed with normal relationship

### LINKED ISSUE #2534: Generated SQL missing aliases with computed rels
### Environment

* PostgreSQL version: 14
* PostgREST version: Build artifact from pull request #2519
* Operating system: Ubuntu

### Description of issue

So I'm testing a build of Postgrest that contains the changes up in pull request #2519 that has fixed a bunch of my heavily nested queries due to missing aliases.

However I think I found another layer of aliasing that is missed. Given the following url query

```
select=well_id,container(root_container(*))
```

where  `root_container(*)` is a computed rel function. The details of that function dont matter other than it returns one row of type container

```
create or replace function api.root_container(api.container) returns setof api.container rows 1 as $$
  select * from api.container limit 1
$$ stable language sql;
```

Postgrest generates the following query

```
WITH pgrst_source AS ( 
  SELECT "api"."well"."well_id", row_to_json("well_container_1".*) AS "container" 
  FROM "api"."well" 
  LEFT JOIN LATERAL ( 
    SELECT row_to_json("container_root_container_2".*) AS "root_container" 
    FROM "api"."container" AS "container_1" 
    LEFT JOIN LATERAL ( 
      SELECT "container_2".* FROM "api"."root_container"("container") AS "container_2"     
    ) AS "container_root_container_2" ON TRUE 
    WHERE "container_1"."container_id" = "api"."well"."container_id"   
  ) AS "well_container_1" ON TRUE   
  LIMIT $1 OFFSET $2 
)
```

and produces the following error in Postgresql

`ERROR:  column "container" does not exist`

The issue comes from this part of the query, that references the computed rel (function) and the passed in value

```
FROM "api"."root_container"("container")
```

However there is an alias for the outer query we actually want to reference which is `container_1`
```
FROM "api"."container" AS "container_1" 
```
So if we change to
```
FROM "api"."root_container"("container_1")
```

the query executes

Thank you again for the absolute amazing work, communication and progress on PostgREST.

steve-chavez:
Closes #2534

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
  - #2343, Return status code 200 for PATCH requests which don't affect any rows - @wolfgangwalther
  - #2481, Treat computed relationships not marked SETOF as M2O/O2O relationship - @wolfgangwalther
+ - #2534, Fix embedding a computed relationship with a normal relationship - @steve-chavez
 
 ### Changed
 
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -107,7 +107,7 @@ initReadRequest qi@QualifiedIdentifier{..} =
   foldr (treeEntry rootDepth) $ Node defReadPlan{from=qi, relName=qiName, depth=rootDepth} []
   where
     rootDepth = 0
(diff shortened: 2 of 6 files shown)


steve-chavez:
Yeah, should be ready now.

bolerodan:
Damn this work is amazing guys, just wanted to post and say thanks (also now a Patreon sub), so far this build has shown no issues with our massive, highly interconnected schema.

thanks again

Issue #2475, opened 2022-09-14
Confused on computed columns and relationships


Iced-Sun:
### Environment

* PostgreSQL version: not relevant
* PostgREST version: v10.0.0
* Operating system: not relevant

### Description of issue

The current implementation allows the mixed usage of computed columns and relationships, i.e. both the queries of `/premieres?select=*,film(*)` and ` /premieres?select=*,film` return something.

After some experiments, I got a bit confused since some of the results seem unexpected and unreasonable to me.

With the fixture
```sql
create schema api;

create table api.premieres (
  id text primary key,
  film_id text -- references api.films(id)
);
insert into api.premieres values
  ('P1', 'F1'),
  ('P2', 'F1'),
  ('P3', 'F2'),
  ('P4', 'F3')
 ;

create table api.films(id text primary key);
insert into api.films values('F1'), ('F2'), ('F4');
```
and the functions
```sql
create or replace function api.computed_films_m2o(api.premieres) returns setof api.films as $$
  select * from api.films where id = $1.film_id
$$ stable language sql rows 1;

create or replace function api.computed_films_scalar(api.premieres) returns api.films as $$
  select * from api.films where id = $1.film_id
$$ stable language sql;

create or replace function api.computed_films_m2m(api.premieres) returns setof api.films as $$
  select * from api.films
$$ stable language sql;

create or replace function api.computed_films_m2m_marked_as_m2o(api.premieres) returns setof api.films as $$
  select * from api.films
$$ stable language sql rows 1;

create or replace function api.computed_premiers_o2m(api.films) returns setof api.premieres as $$
  select * from api.premieres where film_id = $1.id
$$ stable language sql;

create or replace function api.computed_premiers_o2m_marked_as_o2o(api.films) returns api.premieres as $$
  select * from api.premieres where film_id = $1.id
$$ stable language sql;
```
let's access the functions with both the column and embedding syntaxes.

Here are some test results:

```http
GET /premieres?select=*,computed_films_m2o(*)

-- expected usage and expected results
[{"id":"P1","film_id":"F1","computed_films_m2o":{"id":"F1"}}, 
 {"id":"P2","film_id":"F1","computed_films_m2o":{"id":"F1"}}, 
 {"id":"P3","film_id":"F2","computed_films_m2o":{"id":"F2"}}, 
 {"id":"P4","film_id":"F3","computed_films_m2o":null}]
```

```http
GET /premieres?select=*,computed_films_m2o

[...]

PR #2537, opened 2022-10-27
Stricter query string parsing

### LINKED ISSUE #2285: Trailing spaces in columns are treated as part of the column name
### Environment

* PostgreSQL version: 14.2
* PostgREST version: 9.0.0.20220211

### Description of issue

The `pField` function does not ignore trailing spaces when a column is selected:
```
/table?select=name,join (*) 
```
It does work for leading spaces though: 
```
/table?select=name, join(*)
``` 
To keep consistency it should either allow both or ignore both (I think the latter is the better option).

### LINKED ISSUE #2362: Bad error message when `[]` is used inside `select`
Our parsers are failing to give a good error message for this case:

```python
http GET 'localhost:3000/projects?select=id,clients(name[])'

{
    "code": "42703",
    "details": null,
    "hint": "Perhaps you meant to reference the column \"projects.client_id\".",
    "message": "column projects.clients does not exist"
}
```

The error message should come from postgresql and should be like:

```sql
select id, "name[]" from test.projects;

ERROR:  column "name[]" does not exist
HINT:  Perhaps you meant to reference the column "projects.name".
```

wolfgangwalther:
There's probably a lot more potential to make the parser more strict. This just addresses what came up in several issues recently (Fixes #2285, Fixes #2362, [#2475](https://github.com/PostgREST/postgrest/issues/2475#issuecomment-1251516120)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2534, Fix embedding a computed relationship with a normal relationship - @steve-chavez
  - #2362, Fix error message when [] is used inside select - @wolfgangwalther
  - #2475, Disallow !inner on computed columns - @wolfgangwalther
+ - #2285, Ignore leading and trailing spaces in column names when parsing the query string - @wolfgangwalther
 
 ### Changed
 
--- a/src/PostgREST/ApiRequest/QueryParams.hs
+++ b/src/PostgREST/ApiRequest/QueryParams.hs
@@ -352,15 +352,21 @@ pStar = string "*" $> "*"
 -- >>> P.parse pFieldName "" ":"
 -- Left (line 1, column 1):
 -- unexpected ":"
--- expecting field name (* or [a..z0..9_ $])
+-- expecting field name (* or [a..z0..9_$])
 --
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
  - #2343, Return status code 200 for PATCH requests which don't affect any rows - @wolfgangwalther
  - #2481, Treat computed relationships not marked SETOF as M2O/O2O relationship - @wolfgangwalther
+ - #2362, #2475, #2537, Stricter parsing of query string. Instead of silently ignoring, the parser now throws on invalid syntax like json paths for embeddings, hints for regular columns, empty casts or fts languages, etc. - @wolfgangwalther
```
steve-chavez, line 24: Love :heart: this change. Should we also include an entry on `### Changed` for this?

steve-chavez: I mean, some requests could start failing for some users. I got the same effect when doing https://github.com/PostgREST/postgrest/issues/1771.

steve-chavez (approved):

PR #2538, opened 2022-10-27
fix: Return status code 200 when PATCHing without changing rows

### LINKED ISSUE #2343: Return an error message for 404 Not Found on PATCH/DELETE
When doing:

```bash
http PATCH 'localhost:3000/projects?id=eq.33' "Prefer: return=representation" <<JSON
{"name": "any"}
JSON
```

This is returned:

```bash
HTTP/1.1 404 Not Found
Content-Range: */*
Date: Tue, 21 Jun 2022 18:27:29 GMT
Server: postgrest/9.0.0.20220531 (pre-release) (fe8b32d)
Transfer-Encoding: chunked

[]
```

The empty array is not useful, it'd be better to return an error message like "No rows where updated".

This would be useful for debugging RLS as well, a policy might deny updates. 

Mentioned in: https://twitter.com/AxelVaindal/status/1539159747353624576

---

Semi-related: https://github.com/PostgREST/postgrest/issues/1655

wolfgangwalther:
This partially reverts #1257 / #1272 / 553531711ba9e8e23cbbf50dd5ef819db1527642 where the 404 was introduced.

A 406 error is still returned when requesting a single object via accept header.

Returning an error when no rows are changed can be introduced through a different syntax again, see the discussion in #2164.

Fixes #2343

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,10 +19,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2458, Fix a regression with the location header when inserting into views with PKs from multiple tables - @wolfgangwalther
  - #2356, Fix a regression in openapi output with mode follow-privileges - @wolfgangwalther
  - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
+ - #2343, Return status code 200 for PATCH requests which don't affect any rows - @wolfgangwalther
 
 ### Changed
 
  - #2444, Removed `db-pool-timeout` option, because this was removed upstream in hasql-pool. - @robx
+ - #2343, PATCH requests that don't affect any rows no longer return 404 - @wolfgangwalther
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2458, Fix a regression with the location header when inserting into views with PKs from multiple tables - @wolfgangwalther
  - #2356, Fix a regression in openapi output with mode follow-privileges - @wolfgangwalther
  - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
+ - #2343, Return status code 200 for PATCH requests which don't affect any rows - @wolfgangwalther
```
steve-chavez, line 18: I think we should add an entry on `### Changed` too.

steve-chavez: Like `PATCH requests that don't affect rows no longer return 404`

steve-chavez (approved):

PR #2539, opened 2022-10-27
Test non existing config option / clean up CHANGELOG

wolfgangwalther:
As discussed in https://github.com/PostgREST/postgrest/issues/2530#issuecomment-1293550900 this confirms that we can make the next release a minor release without breaking changes.

PR #2540, opened 2022-10-27
Fix computed relationships with non SETOF return

### LINKED ISSUE #2481: Unexpected behavior for computed relationships using scalar functions (`returns <tablename>)
With the fixture

```sql
create schema api;

create table api.premieres (
  id text primary key,
  film_id text -- references api.films(id)
);
insert into api.premieres values
  ('P1', 'F1'),
  ('P2', 'F1'),
  ('P3', 'F2'),
  ('P4', 'F3')
 ;

create table api.films(id text primary key);
insert into api.films values('F1'), ('F2'), ('F4');
```
and the functions
```sql
create or replace function api.computed_premiers_o2m(api.films) returns setof api.premieres as $$
  select * from api.premieres where film_id = $1.id
$$ stable language sql;

create or replace function api.computed_premiers_scalar(api.films) returns api.premieres as $$
  select * from api.premieres where film_id = $1.id
$$ stable language sql;
```

```http
GET /premieres?select=*,computed_films_scalar(*)

-- questionable usage, expected results (taken as m2m relationship)
[{"id":"P1","film_id":"F1","computed_films_scalar":[{"id":"F1"}]}, 
 {"id":"P2","film_id":"F1","computed_films_scalar":[{"id":"F1"}]}, 
 {"id":"P3","film_id":"F2","computed_films_scalar":[{"id":"F2"}]}, 
 {"id":"P4","film_id":"F3","computed_films_scalar":[{"id":null}]}]
```

If we want a rowtype-returning scalar function used as a computed relationship, my suggestion would be marking it as in `rows 1`, because such a function always returns one row, no more, no less.

```sql
-- check the film existence
select f from api.films f where id = 'F1';
--   f   
-- ------
--  (F1)
-- (1 row)

select f from api.films f where id = 'F3';
-- f 
-- ---
-- (0 rows)

--- the differences between set-returning function and scalar function
-- scalar function returns one row even there are two matching rows
select * from api.computed_premiers_scalar((select f from api.films f where id = 'F1'));
--  id | film_id 
-- ----+---------
--  P1 | F1
-- (1 row)

-- set-returning function returns all the matching rows
select * from api.computed_premiers_o2m((select f from api.films f where id = 'F1'));
--  id | film_id 
-- ----+---------
--  P1 | F1
--  P2 | F1
-- (2 rows)

-- scalar function returns one row even there are no matching row
select * from api.computed_premiers_scalar((select f from api.films f where id = 'F3'));
--  id | film_id 
-- ----+---------
--  ¤  | ¤
-- (1 row)

-- set-returning function returns no rows
select * from api.computed_premiers_o2m((select f from api.films f where id = 'F3'));
-- id | film_id 
-- ----+---------
-- (0 rows)
```

_Originally posted by @Iced-Sun in https://github.com/PostgREST/postgrest/issues/2475#issuecomment-1251843074_

wolfgangwalther:
Fixes #2481.

We should still advertise the usage of `SETOF` and `ROWS 1` in the docs, because that's the better choice for inlining. Still, making the non-SETOF case "just work" seems like a good idea.

~~(Based on #2539, only last commit is relevant)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2356, Fix a regression in openapi output with mode follow-privileges - @wolfgangwalther
  - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
  - #2343, Return status code 200 for PATCH requests which don't affect any rows - @wolfgangwalther
+ - #2481, Treat computed relationships not marked SETOF as M2O/O2O relationship - @wolfgangwalther
 
 ### Changed
 
--- a/src/PostgREST/SchemaCache.hs
+++ b/src/PostgREST/SchemaCache.hs
@@ -727,12 +727,12 @@ allComputedRels =
     computed_rels as (
       select
         (parse_ident(p.pronamespace::regnamespace::text))[1] as schema,
-        p.proname::text          as name,
-        arg_schema.nspname::text as rel_table_schema,
-        arg_name.typname::text   as rel_table_name,
-        ret_schema.nspname::text as rel_ftable_schema,
-        ret_name.typname::text   as rel_ftable_name,
-        p.prorows = 1            as single_row
+        p.proname::text                  as name,
+        arg_schema.nspname::text         as rel_table_schema,
+        arg_name.typname::text           as rel_table_name,
+        ret_schema.nspname::text         as rel_ftable_schema,
+        ret_name.typname::text           as rel_ftable_name,
+        not p.proretset or p.prorows = 1 as single_row
       from pg_proc p
         join pg_type      arg_name   on arg_name.oid = p.proargtypes[0]
         join pg_namespace arg_schema on arg_schema.oid = arg_name.typnamespace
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2458, Fix a regression with the location header when inserting into views with PKs from multiple tables - @wolfgangwalther
  - #2356, Fix a regression in openapi output with mode follow-privileges - @wolfgangwalther
  - #2283, Fix infinite recursion when loading schema cache with self-referencing view - @wolfgangwalther
+ - #2481, Treat computed relationships not marked SETOF as M2O/O2M relationship instead of M2M - @wolfgangwalther
```
steve-chavez, line 22: Hm, shouldn't this be

```suggestion
 - #2481, Treat computed relationships not marked SETOF as M2O/O2O relationship instead of M2M - @wolfgangwalther
```

A `RETURNS table` always points to a to-one relationship right?

steve-chavez: Yeah, that's confusing. How about just removing the M2M mention(I've edited my suggestion above).

steve-chavez: And this should be a feature(in `### Added`) I think

steve-chavez (approved):

PR #2523, opened 2022-10-21
Data Representations

aljungberg:
# Data Representations

Present and receive API fields with custom formatting rules.

For certain APIs, the default conversion to and from JSON performed by PostgreSQL may not be right. For example you might wish to represent dates in UNIX epoch timestamp format, fixed precision decimals as strings rather than JSON floats, colours as CSS hex strings, binary blobs as base64 data and so forth. Perhaps you even need different representations in different API endpoints. 

Data Representations allow you to create custom, two-way convertible representations of your data on a per field level using standard PostgreSQL functions. Just create a `domain`, which we can view as a kind of type alias that does nothing by default, and then define casts to and from JSON. (Normally PostgreSQL ignores domain casts, but Data Representations detects those casts and applies them automatically as needed.)

Features:

- Lets you define custom data types and attendant formatters and parsers.
- Use them in your base tables to apply custom representations globally in all APIs using those tables.
- Or create views with different representations for the same underlying columns.
- Presents data: fields transparently formatted into JSON when using for example `GET`.
- Receive data: automatically parses custom data types in incoming JSON payloads when using `POST` or `PATCH`.
- Filters: parses query parameters sent as plain text if you also specify a `TEXT->my custom type` cast.
- Works for data `RETURNING` operations, like if you request the full body of your patched resource with `return=presentation`.
- Works for batch editing.
- Completely opt in. If you don't want it, don't define any domain casts. (There's no reason to do so in ordinary usage since PostgreSQL doesn't support such casts.)
- Fully integrated: unlike a naive implementation where you create a view with computed columns that formats the data the way you want it, with Data Representations you don't lose the ability to write to those fields. They are two-way.
- High performant: since your filters are parsed to their native data type *before* being applied for filtering, you don't need a second index of your computed presentation value. Querying is as fast as querying the underlying true data type.

## Example

PostgreSQL has a built-in formatter for dates:

```bash
curl "http://mypostgrest/todo?select=id,due_at&id=eq.1"
[{"id":1, "due_at":"2018-01-02T00:00:00+00:00"}]
```

Let's replace the `+00:00` suffix with `Z`, an equivalent but shorter ISO 8601 date:

```sql
CREATE DOMAIN public.isodate AS timestamp with time zone;
CREATE OR REPLACE FUNCTION json(public.isodate) RETURNS json AS $$
    SELECT to_json(replace(to_json($1)#>>'{}', '+00:00', 'Z'));
$$ LANGUAGE SQL IMMUTABLE;
[...]

Issue #1567, opened 2020-08-05
Feature Request: Updating Columns to Default Values


Exahilosys:
Currently, there is no way to update tables, setting columns to their default values. I have found myself wanting to allow "resetting" of columns back to how they were when the row was created, but to no avail since such a feature is not supported.

I suggest adding a `defaults` or `revert` query parameter for `PATCH` requests that accepts a comma-separated list of column names (like the `columns` param for `POST`) to be set to their default values using `column = DEFAULT`.

Thank you for this amazing piece of software.

wolfgangwalther:
I think we don't even need another parameter for this - we could just use the existing `columns`, like this:

- every column in the `?columns=...` list gets added to the `UPDATE SET` part

- when a column is not defined in the json body, the column is set to `DEFAULT` instead

- all keys in the json that are not in the list are ignored (as is currently the case)

This can also be extended to `POST` - and would improve batch inserts as well, because some inserted records could have a value set, while others would use the default value for the same key.

Currently both `PATCH` and `POST` insert `NULL` when a column in `?columns=` doesn't have a key in the json body. But since `null` is a valid json value as well and can therefore be set explicitely, this wouldn't need to be like this. While I couldn't find anything in the docs or in the tests about this, I guess it would still have to be considered a breaking change.

steve-chavez:
@wolfgangwalther POST has an issue since it's not possible to do `INSERT INTO ... SELECT DEFAULT, col1..`. Check this [question](https://dba.stackexchange.com/questions/134399/how-to-insert-values-into-a-table-with-default-values-from-a-select-query-in-pos).

That being said, I think your idea should work for PATCH because of the `UPDATE .. SET = DEFAULT` form. To do this we'd have to pass all the table/view columns to QueryBuilder.hs and make a diff with `?columns` for the defaults.

wolfgangwalther:
Down the road it would be good to be able to support this for `POST` as well - would be good if `PATCH` and `POST` behave the same here.

[...]

PR #2542, opened 2022-10-27
Json to recordset

aljungberg:
Switches from `json_populate_recordset` to `json_to_recordset`, which we will need for data representations. See #2523 for details.

The impossible panic we discussed in #2523 has been eliminated by creating a `TypedField` type which doesn't allow that situation.

### DIFF
--- a/src/PostgREST/ApiRequest/Types.hs
+++ b/src/PostgREST/ApiRequest/Types.hs
@@ -81,7 +81,7 @@ data ApiRequestError
   | SpreadNotToOne Text Text
   | UnacceptableSchema [Text]
   | UnsupportedMethod ByteString
-  | ColumnNotFound Text
+  | ColumnNotFound Text Text
 
 data QPError = QPError Text Text
 data RangeError
--- a/src/PostgREST/ApiRequest/Types.hs
+++ b/src/PostgREST/ApiRequest/Types.hs
@@ -81,6 +81,7 @@ data ApiRequestError
   | SpreadNotToOne Text Text
   | UnacceptableSchema [Text]
   | UnsupportedMethod ByteString
+  | ColumnNotFound Text
 
 data QPError = QPError Text Text
 data RangeError
@@ -144,7 +145,7 @@ type JsonPath = [JsonOperation]
 data JsonOperation
   = JArrow { jOp :: JsonOperand }
   | J2Arrow { jOp :: JsonOperand }
-  deriving (Eq)
+  deriving (Eq, Ord)
 
 -- | Represents the key(`->'key'`) or index(`->'1`::int`), the index is Text
 -- because we reuse our escaping functons and let pg do the casting with
@@ -152,7 +153,7 @@ data JsonOperation
 data JsonOperand
   = JKey { jVal :: Text }
   | JIdx { jVal :: Text }
-  deriving (Eq)
+  deriving (Eq, Ord)
 
 -- | Boolean logic expression tree e.g. "and(name.eq.N,or(id.eq.1,id.eq.2))" is:
 --
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -199,7 +199,7 @@ instance JSON.ToJSON ApiRequestError where
     "hint"    .= ("Try renaming the parameters or the function itself in the database so function overloading can be resolved" :: Text)]
   toJSON (ColumnNotFound colName) = JSON.object [
     "code"    .= ApiRequestErrorCode18,
-    "message" .= ("Could not find '" <> colName <> "' in the target table" :: Text),
+    "message" .= ("Column '" <> colName <> "' does not exist" :: Text),
     "details" .= JSON.Null,
     "hint"    .= JSON.Null]
 
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -76,6 +76,7 @@ instance PgrstError ApiRequestError where
   status UnacceptableSchema{}    = HTTP.status406
   status UnsupportedMethod{}     = HTTP.status405
   status LimitNoOrderError       = HTTP.status400
+  status ColumnNotFound{}        = HTTP.status400
 
   headers _ = [MediaType.toContentType MTApplicationJSON]
 
@@ -196,6 +197,11 @@ instance JSON.ToJSON ApiRequestError where
     "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [ppName a <> " => " <> ppType a | a <- pdParams p] <> ")" | p <- procs]),
     "details" .= JSON.Null,
     "hint"    .= ("Try renaming the parameters or the function itself in the database so function overloading can be resolved" :: Text)]
+  toJSON (ColumnNotFound colName) = JSON.object [
+    "code"    .= ApiRequestErrorCode18,
+    "message" .= ("Could not find '" <> colName <> "' in the target table" :: Text),
+    "details" .= JSON.Null,
+    "hint"    .= ("If a new column was created in the database with this name, try reloading the schema cache." :: Text)]
 
 compressedRel :: Relationship -> JSON.Value
 -- An ambiguousness error cannot happen for computed relationships TODO refactor so this mempty is not needed
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -197,9 +197,9 @@ instance JSON.ToJSON ApiRequestError where
     "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [ppName a <> " => " <> ppType a | a <- pdParams p] <> ")" | p <- procs]),
     "details" .= JSON.Null,
     "hint"    .= ("Try renaming the parameters or the function itself in the database so function overloading can be resolved" :: Text)]
-  toJSON (ColumnNotFound colName) = JSON.object [
+  toJSON (ColumnNotFound relName colName) = JSON.object [
     "code"    .= ApiRequestErrorCode18,
-    "message" .= ("Column '" <> colName <> "' does not exist" :: Text),
+    "message" .= ("Column '" <> colName <> "' of relation '" <> relName <> "' does not exist" :: Text),
     "details" .= JSON.Null,
     "hint"    .= JSON.Null]
 
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -201,7 +201,7 @@ instance JSON.ToJSON ApiRequestError where
     "code"    .= ApiRequestErrorCode18,
     "message" .= ("Could not find '" <> colName <> "' in the target table" :: Text),
     "details" .= JSON.Null,
-    "hint"    .= ("If a new column was created in the database with this name, try reloading the schema cache." :: Text)]
+    "hint"    .= JSON.Null]
 
 compressedRel :: Relationship -> JSON.Value
 -- An ambiguousness error cannot happen for computed relationships TODO refactor so this mempty is not needed
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -380,9 +380,9 @@ mutatePlan :: Mutation -> QualifiedIdentifier -> ApiRequest -> SchemaCache -> Re
 mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
   case mutation of
     MutationCreate ->
-      mapRight (\typedColumns -> Insert qi (S.fromList typedColumns) body ((,) <$> iPreferResolution <*> Just confCols) [] returnings pkCols) typedColumnsOrError
+      mapRight (\typedColumns -> Insert qi typedColumns body ((,) <$> iPreferResolution <*> Just confCols) [] returnings pkCols) typedColumnsOrError
     MutationUpdate ->
-      mapRight (\typedColumns -> Update qi (S.fromList typedColumns) body combinedLogic iTopLevelRange rootOrder returnings) typedColumnsOrError
+      mapRight (\typedColumns -> Update qi typedColumns body combinedLogic iTopLevelRange rootOrder returnings) typedColumnsOrError
     MutationSingleUpsert ->
         if null qsLogic &&
            qsFilterFields == S.fromList pkCols &&
@@ -390,7 +390,7 @@ mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
            all (\case
               Filter _ (OpExpr False (Op OpEqual _)) -> True
               _                                      -> False) qsFiltersRoot
-          then mapRight (\typedColumns -> Insert qi (S.fromList typedColumns) body (Just (MergeDuplicates, pkCols)) combinedLogic returnings mempty) typedColumnsOrError
+          then mapRight (\typedColumns -> Insert qi typedColumns body (Just (MergeDuplicates, pkCols)) combinedLogic returnings mempty) typedColumnsOrError
         else
           Left InvalidFilters
     MutationDelete -> Right $ Delete qi combinedLogic iTopLevelRange rootOrder returnings
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -27,7 +27,7 @@ import qualified Data.HashMap.Strict        as HM
 import qualified Data.Set                   as S
 import qualified PostgREST.SchemaCache.Proc as Proc
 
-import Data.Either.Combinators (mapLeft)
+import Data.Either.Combinators (mapLeft, mapRight)
 import Data.List               (delete)
 import Data.Tree               (Tree (..))
 
@@ -54,14 +54,14 @@ import PostgREST.SchemaCache.Relationship (Cardinality (..),
                                            Relationship (..),
                                            RelationshipsMap,
                                            relIsToOne)
-import PostgREST.SchemaCache.Table        (tablePKCols)
-
-import PostgREST.Plan.CallPlan
-import PostgREST.Plan.MutatePlan
-import PostgREST.Plan.ReadPlan   as ReadPlan
+import PostgREST.SchemaCache.Table        (Table, tablePKCols)
 
 import PostgREST.ApiRequest.Preferences
 import PostgREST.ApiRequest.Types
+import PostgREST.Plan.CallPlan
+import PostgREST.Plan.MutatePlan
+import PostgREST.Plan.ReadPlan          as ReadPlan
+import PostgREST.Plan.Types
 
 import qualified PostgREST.ApiRequest.QueryParams as QueryParams
 
@@ -380,8 +380,9 @@ mutatePlan :: Mutation -> QualifiedIdentifier -> ApiRequest -> SchemaCache -> Re
 mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
   case mutation of
     MutationCreate ->
-      Right $ Insert qi iColumns body ((,) <$> iPreferResolution <*> Just confCols) [] returnings pkCols
-    MutationUpdate -> Right $ Update qi iColumns body combinedLogic iTopLevelRange rootOrder returnings
+      mapRight (\typedColumns -> Insert qi (S.fromList typedColumns) body ((,) <$> iPreferResolution <*> Just confCols) [] returnings pkCols) typedColumnsOrError
+    MutationUpdate ->
+      mapRight (\typedColumns -> Update qi (S.fromList typedColumns) body combinedLogic iTopLevelRange rootOrder returnings) typedColumnsOrError
     MutationSingleUpsert ->
         if null qsLogic &&
            qsFilterFields == S.fromList pkCols &&
@@ -389,7 +390,7 @@ mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
            all (\case
               Filter _ (OpExpr False (Op OpEqual _)) -> True
               _                                      -> False) qsFiltersRoot
-          then Right $ Insert qi iColumns body (Just (MergeDuplicates, pkCols)) combinedLogic returnings mempty
+          then mapRight (\typedColumns -> Insert qi (S.fromList typedColumns) body (Just (MergeDuplicates, pkCols)) combinedLogic returnings mempty) typedColumnsOrError
         else
           Left InvalidFilters
     MutationDelete -> Right $ Delete qi combinedLogic iTopLevelRange rootOrder returnings
@@ -405,6 +406,21 @@ mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
     rootOrder = maybe [] snd $ find (\(x, _) -> null x) qsOrder
     combinedLogic = foldr addFilterToLogicForest logic qsFiltersRoot
     body = payRaw <$> iPayload -- the body is assumed to be json at this stage(ApiRequest validates)
+    typedColumnsOrError = mapStopOnLeft (resolveOrError (HM.lookup qi $ dbTables sCache)) $ S.toList iColumns
+
+resolveOrError :: Maybe Table -> FieldName -> Either ApiRequestError TypedField
+resolveOrError Nothing _ = Left NotFound
+resolveOrError (Just table) field =
+  case resolveTableField table field of
+    Nothing         -> Left $ ColumnNotFound field
+    Just typedField -> Right typedField
+
+mapStopOnLeft :: (a -> Either b c) -> [a] -> Either b [c]
+mapStopOnLeft _ [] = Right []
+mapStopOnLeft f (x:xs) =
+  case f x of
+    Left e  -> Left e
+    Right y -> (y:) <$> mapStopOnLeft f xs
 
 callPlan :: ProcDescription -> ApiRequest -> ReadPlanTree -> CallPlan
 callPlan proc apiReq readReq = FunctionCall {
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -406,7 +406,8 @@ mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
     rootOrder = maybe [] snd $ find (\(x, _) -> null x) qsOrder
     combinedLogic = foldr addFilterToLogicForest logic qsFiltersRoot
     body = payRaw <$> iPayload -- the body is assumed to be json at this stage(ApiRequest validates)
-    typedColumnsOrError = mapStopOnLeft (resolveOrError (HM.lookup qi $ dbTables sCache)) $ S.toList iColumns
+    tbl = HM.lookup qi $ dbTables sCache
+    typedColumnsOrError = resolveOrError tbl `traverse` S.toList iColumns
 
 resolveOrError :: Maybe Table -> FieldName -> Either ApiRequestError TypedField
 resolveOrError Nothing _ = Left NotFound
@@ -414,13 +415,6 @@ resolveOrError (Just table) field =
   case resolveTableField table field of
     Nothing         -> Left $ ColumnNotFound field
     Just typedField -> Right typedField
-
-mapStopOnLeft :: (a -> Either b c) -> [a] -> Either b [c]
-mapStopOnLeft _ [] = Right []
-mapStopOnLeft f (x:xs) =
-  case f x of
-    Left e  -> Left e
-    Right y -> (y:) <$> mapStopOnLeft f xs
 
 callPlan :: ProcDescription -> ApiRequest -> ReadPlanTree -> CallPlan
 callPlan proc apiReq readReq = FunctionCall {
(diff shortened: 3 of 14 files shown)


src/PostgREST/ApiRequest/Types.hs
```diff
@@ -92,6 +98,23 @@ type Cast = Text
 type Alias = Text
 type Hint = Text
 
+
+-- | A TypedField is a field with sufficient information to be read from JSON with `json_to_recordset`. In a future PR
+-- type fields will also be key for the data representations feature.
+-- Typed fields should have a nonempty field name and a nonempty type. We enforce this invariant in the `resolveField`
+-- function and keep the real constructor private. (It's not foolproof, you can still construct a `TypedField` using an
+-- existing TypedField and the Haskell `{ tfFieldName = "" }` syntax, so don't do that.)
```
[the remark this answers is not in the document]
aljungberg: I think there are n ways to create an invalid field description like this, not just an empty string but invalid identifiers and so on. So there's no need to go to extremes with validation, which is why I didn't. At the same time a little validation can save some headache down the road, fail fast style.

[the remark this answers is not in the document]
aljungberg: Actually I managed to do it while making this change with a typo in the SQL such that the field type ended up indeed empty! I can also imagine some request with `columns=,b` or whatever.

Anyhow I don't have a strong opinion either way. Happy to make the constructor public if that's preferred.

src/PostgREST/Plan.hs
```diff
@@ -109,7 +110,7 @@ initReadRequest qi@QualifiedIdentifier{..} =
     rootDepth = 0
     defReadPlan = ReadPlan [] (QualifiedIdentifier mempty mempty) Nothing [] [] allRange mempty Nothing [] Nothing Nothing Nothing rootDepth
     treeEntry :: Depth -> Tree SelectItem -> ReadPlanTree -> ReadPlanTree
-    treeEntry depth (Node fld@((fldName, _),_,alias, hint, joinType) fldForest) (Node q rForest) =
+    treeEntry depth (Node fld@((fldName,_),_,alias, hint, joinType) fldForest) (Node q rForest) =
```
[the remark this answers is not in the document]
aljungberg: Agreed. Would be nice if the linter caught such inconsistencies.

src/PostgREST/SchemaCache/Table.hs
```diff
@@ -40,10 +49,21 @@ data Column = Column
...
 type TablesMap = HM.HashMap QualifiedIdentifier Table
+type ColumnMap = HMI.InsOrdHashMap FieldName Column
+
+instance Ord ColumnMap where
+  compare = compare `on` HMI.keys
+
+resolveTableField :: Table -> FieldName -> Maybe TypedField
+resolveTableField table fieldName =
+  case HMI.lookup fieldName (tableColumns table) of
+    Just column -> resolveField (fieldName, []) (colNominalType column)
```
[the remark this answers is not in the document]
aljungberg: Great, I like it.

src/PostgREST/Plan.hs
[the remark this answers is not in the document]
aljungberg: Fixed in the next commit.

src/PostgREST/ApiRequest/Types.hs
```diff
@@ -92,6 +98,23 @@ type Cast = Text
...
+data TypedField = PrivateTypedField
+   { tfFieldName :: FieldName
+   , tfJsonPath  :: JsonPath
+   , tfIRType    :: Text
+   } deriving (Eq, Ord)
```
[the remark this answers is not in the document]
aljungberg: Hmm yes. I wanted it near `Field`. We will end up with some analogs such as `Filter` vs `TypedFilter` (which leads to `TypedLogicTree` etc too unfortunately). It is true though that like the defining difference is sort of that in `ApiRequest` we have kind of raw user defined data and later we've distilled that into augmented, better validated data. So based on that extinction, certainly putting it in `Plan/` makes a lot of sense because it will be `Plan` that's responsible for that distillation.

src/PostgREST/ApiRequest/Types.hs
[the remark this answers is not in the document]
aljungberg: Right, isn't that the idea with `TypedField` though, to be better validated and therefore less able to cause runtime errors down the line? The nature of bugs is that they seem impossible, otherwise you wouldn't have them!

Just making the point for the sake of rational debate, again, not married to having extra error checking here.

src/PostgREST/Plan/Types.hs
```diff
@@ -0,0 +1,32 @@
...
+   } deriving (Eq, Ord)
+
+resolveField :: Field -> Text -> TypedField
+resolveField (fieldName, jsonPath) = TypedField fieldName jsonPath
+
+resolveTableField :: Table -> FieldName -> Maybe TypedField
+resolveTableField table fieldName =
+  case HMI.lookup fieldName (tableColumns table) of
+    Just column -> Just $ resolveField (colName column, []) (colNominalType column)
+    Nothing     -> Nothing
```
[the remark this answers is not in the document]
aljungberg: `resolveField` is used by `resolveTableField` to deconstruct the `Field`. Keeps it cleaner, and I expect to use it when we do need the `jsonPath` in TypedField in the next PR (for selects).

aljungberg:
Regarding the code coverage on `Ord`, what's the best approach there? I could stick a few tests in that do some `compare` exercises, but didn't see anywhere else we do that. Could just drop `Ord` from table, seems to be unused, which would mean we don't need it on `ColumnMap` either. We only need `Hashable` for the maps. Thoughts?

src/PostgREST/ApiRequest/Types.hs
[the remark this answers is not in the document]
aljungberg: Has now been moved.

src/PostgREST/ApiRequest/Types.hs
```diff
@@ -92,6 +98,23 @@ type Cast = Text
 type Alias = Text
 type Hint = Text
 
+
+-- | A TypedField is a field with sufficient information to be read from JSON with `json_to_recordset`. In a future PR
+-- type fields will also be key for the data representations feature.
+-- Typed fields should have a nonempty field name and a nonempty type. We enforce this invariant in the `resolveField`
+-- function and keep the real constructor private. (It's not foolproof, you can still construct a `TypedField` using an
+-- existing TypedField and the Haskell `{ tfFieldName = "" }` syntax, so don't do that.)
+data TypedField = PrivateTypedField
```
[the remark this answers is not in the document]
aljungberg: I made the constructor public while moving everything to `Plan`.

src/PostgREST/ApiRequest/Types.hs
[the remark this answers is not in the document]
aljungberg: I dropped the sanity checking from `TypedField`.

aljungberg:
Not entirely sure what this means:

Screenshot 2022-10-28 at 16 09 00

Clearly `insCol` is tested since it's used by the insertion tests?

steve-chavez:
> Switches from json_populate_recordset to json_to_recordset

Would it be possible to fallback to `json_populate_recordset` in case the columns of a relation are not present on the schema cache?

For example, this could happen if a new table is added and the schema cache is not refreshed. We remained operational before for these cases, what would happen now?

aljungberg:
> > Switches from json_populate_recordset to json_to_recordset
> 
> Would it be possible to fallback to `json_populate_recordset` in case the columns of a relation are not present on the schema cache?
> 
> For example, this could happen if a new table is added and the schema cache is not refreshed. We remained operational before for these cases, what would happen now?

The new behaviour is that you get an error when trying to insert data into a table or a column we believe not to exist.

Personally I think that's actually an improvement over the old behaviour. Clear and immediate messaging for a situation that is more likely to be an error than not.

Like how often do you expect to make API calls to tables so fresh off the presses they're not even in the cache? The only time I can imagine is if you're live developing your schema. In that case can't you do `NOTIFY pgrst, 'reload schema'` at the end of your edit transaction as well?

What's the venn diagram between people live editing their schema yet unwilling to `NOTIFY`? I imagine it's pretty small.

- If we make it a stated goal to always allow interaction with unknown tables and columns, data representations will have confusing failure modes (e.g. they work some of the time).
- Even if you don't use data representations, you can have confusing behaviour with regards to relations, embedding and so forth if you are in the habit of changing your schema without reloading the cache. 
- There's more complexity in our codebase having fallbacks.
- To support `json_populate_recordset` as a fallback, the constraints on `TypedField` would have to be relaxed. The record started out relaxed and I made it more strict due to the feedback on [the data rep PR](https://github.com/PostgREST/postgrest/pull/2523). To be clear, I'm happy to do the work, but we might be pulling in two directions at once here.

All that said, happy to change it. I'm really excited about this feature and I think it'll be a great addition to PostgREST, so keen to get it merged. We're going to be using this probably in every single API on our end.

src/PostgREST/Error.hs
```diff
@@ -180,6 +181,11 @@ instance JSON.ToJSON ApiRequestError where
     "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [ppName a <> " => " <> ppType a | a <- pdParams p] <> ")" | p <- procs]),
     "details" .= JSON.Null,
     "hint"    .= ("Try renaming the parameters or the function itself in the database so function overloading can be resolved" :: Text)]
+  toJSON (ColumnNotFound colName) = JSON.object [
+    "code"    .= ApiRequestErrorCode18,
+    "message" .= ("Could not find '" <> colName <> "' in the target table" :: Text),
+    "details" .= JSON.Null,
+    "hint"    .= ("If a new column was created in the database with this name, try reloading the schema cache." :: Text)]
```
steve-chavez, line 188: Should we include on `details` the list of detected columns? Similarly to how we do it for an error on a relationship?

https://github.com/PostgREST/postgrest/blob/b2935ef7990107642e261b25b6f30a2aed932864/test/spec/Feature/Query/EmbedDisambiguationSpec.hs#L17-L44

steve-chavez:
Cool, I agree that becoming more strict is good overall. 

Not sure if stuffing schema cache info on the `details` key of the error message is the right approach. The schema cache is becoming more relevant with this change and maybe we should make it more visible. 

How about we add a special header on the endpoints to get their schema cache entry(discussed before on https://github.com/PostgREST/postgrest/issues/1421).

```http
GET /projects
Accept: application/vnd.pgrst.schema-cache

{
  "columns": [".."],
  "relationships": [".."]
}
```

---

The above can be done on another PR but I'd consider it a must for a stable release after this feature is merged.

src/PostgREST/Error.hs
aljungberg: I like it! Can't see a security problem either, the columns of tables/views exposed through PostgREST are already published in other ways.

aljungberg:
> Not sure if stuffing schema cache info on the `details` key of the error message is the right approach. The schema cache is becoming more relevant with this change and maybe we should make it more visible.
> 
> How about we add a special header on the endpoints to get their schema cache entry(discussed before on #1421).
> 
> ```httpspec
> GET /projects
> Accept: application/vnd.pgrst.schema-cache
> 
> {
>   "columns": [".."],
>   "relationships": [".."]
> }
> ```
> 
> The above can be done on another PR but I'd consider it a must for a stable release after this feature is merged.

Having a per relation info API would be a nice addition. 

As you rightfully point out in the referenced PR, the schema cache is already visible through the root endpoint so perhaps such an API would be more of a nice to have? 

Maybe there's something we can do there to present more information about the available data representations for each column. Haven't fully thought that through yet, but if we add support for CSV and binary data representations down the line it might be nice to see which column custom formatters and/or parsers are available, and which ones will rely on the base type native PostgreSQL casting.

steve-chavez:
> How about making a request to the root endpoint with a different accept header, i.e. the one you mentioned.. and then returning schema cache instead of OpenAPI output?

Yeah, that could be another option. I've branched off the discussion to https://github.com/PostgREST/postgrest/issues/1421#issuecomment-1300683598 to not make this thread longer.

aljungberg:
> Are we throwing the same kind of error messages before? In that case the feature is probably "throwing those errors **without hitting the database**"?

The essence of the error is the same but the error message now has a PostgREST error code and indeed happens without hitting the database.

```
# Before
{"code":"42703","details":null,"hint":null,"message":"column \"helicopter\" of relation \"articles\" does not exist"}
# After
{"code":"PGRST118","details":null,"hint":"If a new column was created in the database with this name, try reloading the schema cache.","message":"Could not find 'helicopter' in the target table"}
```

src/PostgREST/Plan.hs
```diff
@@ -363,6 +364,21 @@ mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
...
+  case resolveTableField table field of
+    Nothing         -> Left $ ColumnNotFound field
+    Just typedField -> Right typedField
+
+mapStopOnLeft :: (a -> Either b c) -> [a] -> Either b [c]
+mapStopOnLeft _ [] = Right []
+mapStopOnLeft f (x:xs) =
+  case f x of
+    Left e  -> Left e
+    Right y -> (y:) <$> mapStopOnLeft f xs
```
steve-chavez, line 381: This looks like [traverse](https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#v:traverse)

> (a -> f b) -> t a -> f (t b)

You might be able to use it like 

```haskell
typedColumnsOrError = resolveOrError (HM.lookup qi $ dbTables sCache) `traverse` S.toList iColumns
```

---

There is an example on https://github.com/PostgREST/postgrest/blob/78d45b4e32e1129f25c2b6739dbd587ef873822c/src/PostgREST/Plan.hs#L310

src/PostgREST/Plan.hs
```diff
@@ -363,6 +364,21 @@ mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
     rootOrder = maybe [] snd $ find (\(x, _) -> null x) qsOrder
     combinedLogic = foldr addFilterToLogicForest logic qsFiltersRoot
     body = payRaw <$> iPayload -- the body is assumed to be json at this stage(ApiRequest validates)
+    typedColumnsOrError = mapStopOnLeft (resolveOrError (HM.lookup qi $ dbTables sCache)) $ S.toList iColumns
```
steve-chavez, line 367: This looks up the table for each column right? Maybe we should cache the table like so
 
```suggestion
    tbl = HM.lookup qi $ dbTables sCache
    typedColumnsOrError = mapStopOnLeft (resolveOrError tbl) $ S.toList iColumns
```

---

Checking the [load tests](https://github.com/PostgREST/postgrest/pull/2542/checks?check_run_id=9363756197)

param | main | head
-- | -- | --
throughput | 317.3776065823794 | 308.3469937995588

We're losing some perf on the PR. Perhaps the above will help.

src/PostgREST/Error.hs
steve-chavez: A more useful `hint` for this one. 

When referring to a bad column name, postgres tries to detect a typo:

```
$ curl 'localhost:3000/projects?select=id,nam'
{"code":"42703","details":null,"hint":"Perhaps you meant to reference the column \"projects.name\".","message":"column projects.nam does not exist"}

$ curl 'localhost:3000/projects?select=id,namee'
{"code":"42703","details":null,"hint":"Perhaps you meant to reference the column \"projects.name\".","message":"column projects.namee does not exist"}

$ curl 'localhost:3000/projects?select=id,nae'
{"code":"42703","details":null,"hint":"Perhaps you meant to reference the column \"projects.name\".","message":"column projects.nae does not exist"}

$ curl 'localhost:3000/projects?select=id,xxname'
{"code":"42703","details":null,"hint":"Perhaps you meant to reference the column \"projects.name\".","message":"column projects.xxname does not exist"}

## here it no longer suggest a column
$ curl 'localhost:3000/projects?select=id,yyxxname'
{"code":"42703","details":null,"hint":null,"message":"column projects.yyxxname does not exist"}
```

PostgreSQL does some fuzzy logic for this: [errorMissingColumn](https://github.com/postgres/postgres/blob/5035c93c8a5ac6804da79c67403460348b381924/src/backend/parser/parse_relation.c#L3625-L3696).

Perhaps we can use https://hackage.haskell.org/package/fuzzyfind-3.0.0/docs/Text-FuzzyFind.html for something similar?

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -81,12 +82,25 @@ getSelectsJoins rr@(Node ReadPlan{relName, relToParent=Just rel, relAggAlias, re
...
 
+pgSelectFromJson :: S.Set TypedField -> SQL.Snippet
+pgSelectFromJson fields =
+  SQL.sql "SELECT " <> parsedCols <> " " <>
+  (if S.null fields
+    -- When we are inserting no columns (e.g. using default values), we can't use our ordinary `json_to_recordset`
+    -- because it can't extract records with no columns (there's no valid syntax for the `AS (colName colType,...)`
+    -- part). But we still need to ensure as many rows are created as there are array elements.
+    then SQL.sql ("FROM json_array_elements (" <> selectBody <> ") _ ")
+    else SQL.sql ("FROM json_to_recordset (" <> selectBody <> ") AS _ " <> "(" <> typedCols <> ") ")
```
steve-chavez, line 93: This reminds me of https://github.com/PostgREST/postgrest/issues/1567#issuecomment-739478169, not having defaults is still a problem.

Would `json_to_recordset` help or make it harder to do the above suggestion later on?

src/PostgREST/Plan.hs
aljungberg: Great comment, glad to see this level of focus on performance. That said, I think no, it doesn't look up the table more than once, more on that below.

First let me just comment on the performance impact overall of this part of the PR: I think the impact will be limited since the operation just adds a small constant time lookup per field, which should be dominated by the time already spent per field. The stuff postgresql does like parsing the JSON, parsing the SQL, doing the actual work of writing to the table; all of these have complexity linearish to _n_, _n_ being the column count, and each component alone is likely much larger than the time for this op (a hash map lookup in RAM).

Back to the table lookup. I'd expect GHC to only evaluate the `HM.lookup qi $ dbTables sCache` node once while partially applying `resolveOrError`. I'm no Haskell guy so might just be flopping about like a fish out of water here, but take a look at the generated core (`ghc -fforce-recomp -O2 -ddump-simpl`):

```
typedColumnsOrError_a2geX
  = hpc<PostgREST.Plan,179>
    hpc<PostgREST.Plan,178>
    $ (hpc<PostgREST.Plan,175>
       mapStopOnLeft_r2g9q
         (let {
            ds20_d2gYE :: Maybe Table
            [LclId]
            ds20_d2gYE
              = hpc<PostgREST.Plan,173>
                $ (hpc<PostgREST.Plan,170>
                   HM.lookup
                     PostgREST.SchemaCache.Identifiers.$fEqQualifiedIdentifier
                     PostgREST.SchemaCache.Identifiers.$fHashableQualifiedIdentifier
                     (hpc<PostgREST.Plan,169> qi_a2geh))
                  (hpc<PostgREST.Plan,172>
                   dbTables (hpc<PostgREST.Plan,171> sCache1_a2geC)) } in
          hpc<PostgREST.Plan,174>
          \ (ds21_d2gYF :: FieldName) ->
            hpc<PostgREST.Plan,168>
            case ds20_d2gYE of {
              Nothing ->
                hpc<PostgREST.Plan,157>
                Data.Either.Left
                  (hpc<PostgREST.Plan,156> PostgREST.ApiRequest.Types.NotFound);
              Just table_a2gf2 ->
                hpc<PostgREST.Plan,167>
                hpc<PostgREST.Plan,160>
                case resolveTableField
                       (hpc<PostgREST.Plan,158> table_a2gf2)
                       (hpc<PostgREST.Plan,159> ds21_d2gYF)
                of {
                  Nothing ->
                    hpc<PostgREST.Plan,164>
                    $ (hpc<PostgREST.Plan,161>
                       \ (ds22_d2gYL :: ApiRequestError) ->
                         Data.Either.Left ds22_d2gYL)
                      (hpc<PostgREST.Plan,163>
                       PostgREST.ApiRequest.Types.ColumnNotFound
                         (hpc<PostgREST.Plan,162> ds21_d2gYF));
                  Just typedField_a2gf4 ->
                    hpc<PostgREST.Plan,166>
                    Data.Either.Right (hpc<PostgREST.Plan,165> typedField_a2gf4)
                }
            }))
      (hpc<PostgREST.Plan,177>
       S.toList (hpc<PostgREST.Plan,176> ds11_d2h02)) } in
```

So `typedColumnsOrError_a2geX` applies `mapStopOnLeft_r2g9q` on a lambda which captures `ds20_d2gYE` from a `let` statement. `ds20_d2gYE` is basically the same as your `tbl = ...`. The first time we evaluate `case ds20_d2gYE of ...` in the lambda, I believe the "thunk" for `ds20_d2gYE` on the heap [should be overwritten](https://downloads.haskell.org/~ghc/7.8.3/docs/html/users_guide/informal-semantics.html) by a "WHNF", and so not evaluated again on subsequent calls to the same lambda. (And if I'm wrong here, the proposed version shouldn't make any difference anyhow since the only difference is that the binding happens a little earlier but still outside of the lambda).

src/PostgREST/Plan.hs
```diff
@@ -338,16 +338,17 @@ mutatePlan :: Mutation -> QualifiedIdentifier -> ApiRequest -> SchemaCache -> Re
 mutatePlan mutation qi ApiRequest{..} sCache readReq = mapLeft ApiRequestError $
   case mutation of
     MutationCreate ->
-      Right $ Insert qi iColumns body ((,) <$> iPreferResolution <*> Just confCols) [] returnings pkCols
-    MutationUpdate -> Right $ Update qi iColumns body combinedLogic iTopLevelRange rootOrder returnings
+      mapRight (\typedColumns -> Insert qi (S.fromList typedColumns) body ((,) <$> iPreferResolution <*> Just confCols) [] returnings pkCols) typedColumnsOrError
```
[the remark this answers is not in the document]
aljungberg: Nice. Yeah, just conforming to the existing structure here, agree converting back is wasted CPU cycles.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -81,12 +82,25 @@ getSelectsJoins rr@(Node ReadPlan{relName, relToParent=Just rel, relAggAlias, re
...
+  )
+  where
+    parsedCols = SQL.sql $ BS.intercalate ", " $ pgFmtIdent . tfFieldName <$> S.toList fields
+    typedCols = BS.intercalate ", " $ pgFmtIdent . tfFieldName <> const " " <> encodeUtf8 . tfIRType <$> S.toList fields
```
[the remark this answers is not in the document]
aljungberg: Yes, I think you're right.

aljungberg:
> I have most of my suggested changes locally already. If you want to save some effort, you can check the "allow edits of maintainers" box (not sure what the name is exactly). Then I can push those to your branch.

Let's revisit that shortly. I'm trying to minimise the amount of rebasing/conflict resolution I have to do on the follow up branch.

(On that note, looking like RPC support will be easy after the rewrite. Some stuff got a little more wordy and verbose but the core will now be a "type X -> type Y" conversion engine with user defined values. This generalises to RPC parameters and results in theory. I'm not implementing that ATM but should be easy when we get there.)

src/PostgREST/Plan.hs
aljungberg: Oh. Newbie mistake. `traverse` makes perfect sense, thanks!

(As a non-Haskell guy, it's not always obvious to me what's implicit in the behaviours of various types. Like in this case how the Functor and Applicative implementations for `Either` lead to a nice short-circuit behaviour when traversing and hitting that first `Left`.)

src/PostgREST/Plan.hs
aljungberg: Broke out `tbl` anyhow, doesn't hurt to be more explicit that we want single eval on that expression.

src/PostgREST/Error.hs
aljungberg: Oh yeah, nice, quite helpful indeed! Save that work for another PR down the road?

aljungberg: Done and pushed.

src/PostgREST/Plan/Types.hs
```diff
@@ -0,0 +1,31 @@
...
+
+import PostgREST.SchemaCache.Identifiers (FieldName)
+import PostgREST.SchemaCache.Table       (Column (..), Table (..))
+
+import Protolude
+
+-- | A TypedField is a field with sufficient information to be read from JSON with `json_to_recordset`.
+data TypedField = TypedField
+   { tfFieldName :: FieldName
+   , tfJsonPath  :: JsonPath
```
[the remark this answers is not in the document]
aljungberg: Done.

src/PostgREST/Plan/Types.hs
```diff
@@ -0,0 +1,31 @@
...
+   , tfIRType    :: Text
+   } deriving (Eq, Ord)
```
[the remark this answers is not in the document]
aljungberg: Done including the rename to `tfName`.

aljungberg:
I believe all feedback up to this point has been addressed.

src/PostgREST/Error.hs
steve-chavez: Sorry for the late reply here.

This is something that still worries me, I've got many users at Supabase that require hand-holding and removing this error is no good for them.

Could you try to add a hint for columns with similar names? On https://github.com/PostgREST/postgrest/pull/2575 we're using the `FuzzySet` library, that might help you get started.

src/PostgREST/Error.hs
```diff
@@ -196,6 +197,11 @@ instance JSON.ToJSON ApiRequestError where
     "message" .= ("Could not choose the best candidate function between: " <> T.intercalate ", " [pdSchema p <> "." <> pdName p <> "(" <> T.intercalate ", " [ppName a <> " => " <> ppType a | a <- pdParams p] <> ")" | p <- procs]),
     "details" .= JSON.Null,
     "hint"    .= ("Try renaming the parameters or the function itself in the database so function overloading can be resolved" :: Text)]
+  toJSON (ColumnNotFound colName) = JSON.object [
+    "code"    .= ApiRequestErrorCode18,
+    "message" .= ("Could not find '" <> colName <> "' in the target table" :: Text),
```
steve-chavez, line 202: Making the message resemble postgresql (`column projects.xxname does not exist"`):

```suggestion
    "message" .= ("Column '" <> colName <> "' does not exist" :: Text),
```

Adding the table name would be great too.

src/PostgREST/Error.hs
aljungberg: Sorry, I just noticed this: are we talking about errors when using unknown columns with `?select`? This PR only changes the error for unknown `?columns`. The behaviour for `select` remains the same:

```
curl 'localhost:3000/todos?select=id,xdone'
# before
{"code":"42703","details":null,"hint":"Perhaps you meant to reference the column \"todos.done\".","message":"column todos.xdone does not exist"}
# after
{"code":"42703","details":null,"hint":"Perhaps you meant to reference the column \"todos.done\".","message":"column todos.xdone does not exist"}
```

This PR only changes the message for bad `columns`:

```
curl --request POST \
  --url 'http://localhost:3000/todos?columns=xtask' \
  --header 'Prefer: return=representation' \
  --data '{"id":5, "task": "buy eggs", color": "#001100"}'
# before
{"code":"42703","details":null,"hint":null,"message":"column \"xtask\" of relation \"todos\" does not exist"}
# after
{"code":"PGRST118","details":null,"hint":null,"message":"Column 'xtask' does not exist"}
```

Indeed the table name was there before, but other than that pretty much the same, no fuzzy suggestions from PostgreSQL in this case.

So are we okay in that case? I certainly don't want to add to your support burden and I feel like this PR should not since the behaviour is the same before and after.

src/PostgREST/Error.hs
aljungberg: I've made the change in message style and added the table name, both (separate from our talk about fuzzy matching above). Thanks!

src/PostgREST/Error.hs
steve-chavez: Oh, I see. I was thinking that PostgreSQL suggestions for `SELECT` would apply to insertions as well.

The table name on the error message would still be nicer but not critical as for now we can only do an insert on a single table at a time.

> So are we okay in that case?

Yes, it's all good now.

aljungberg: Table name already added! Agreed, more of a nice to have but it's more consistent with the select error.

steve-chavez (approved):
LGTM! The error message now is clear as the one from PostgreSQL!

steve-chavez:
@wolfgangwalther Do you have more input on this PR? I think it's looking good to merge.

steve-chavez:
>  It's on my list - as well as a lot of other things ;)

Oh, no problem!

>  Feel free to merge, if you think it's good.

I hesitate because I'm a bit lost on https://github.com/PostgREST/postgrest/pull/2523 (which IIUC is the main goal of the PR ), so will let you review and merge this one.

aljungberg:
> I hesitate because I'm a bit lost on #2523 (which IIUC is the main goal of the PR ), so will let you review and merge this one.

Anything I can do to help explain or detail it for you? If you'd like I could discuss it over a video call.

steve-chavez:
@aljungberg Sorry for the delay! Let's move forward with this one - we'll also need it for #2594, so I've just merged it.

(I'll start reviewing #2523 and I'll let you know any doubts I have there)

PR #2560, opened 2022-11-04
fix: Fatal error when requesting limit=0 and db-max-rows is set

### LINKED ISSUE #2558: fatal: range without lower bound
To reproduce:

```bash
PGRST_DB_MAX_ROWS=100 PGRST_DB_ANON_ROLE='postgrest_test_anonymous' postgrest-with-postgresql-14   postgrest-run
03/Nov/2022:17:53:37 -0500: Attempting to connect to the database...
03/Nov/2022:17:53:37 -0500: Connection successful
03/Nov/2022:17:53:37 -0500: Listening on port 3000
03/Nov/2022:17:53:37 -0500: Listening for notifications on the pgrst channel
03/Nov/2022:17:53:37 -0500: Config reloaded
03/Nov/2022:17:53:37 -0500: Schema cache loaded
FatalError {fatalErrorMessage = "range without lower bound"}
```

(Needs `max-rows`)

```
curl -X GET "localhost:3000/projects?limit=0"
Something went wrong
```

Bug introduced on https://github.com/PostgREST/postgrest/pull/2269.

The major issue is that every fatal error consumes a pool connection permanently, so if `PGRST_DB_POOL=1` is specified PostgREST becomes unresponsive.

laurenceisla:
Partially addresses #2558

- [x] Fix error
- [x] Changelog

~~Remove panic from `rangeOffset` function~~ . Left this one for later since it may take more time to fix.

### DIFF
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -40,8 +40,7 @@ import PostgREST.Config                   (AppConfig (..))
 import PostgREST.Error                    (Error (..))
 import PostgREST.Query.SqlFragment        (sourceCTEName)
 import PostgREST.RangeQuery               (NonnegRange, allRange,
-                                           hasLimitZero,
-                                           limitZeroRange,
+                                           convertToLimitZeroRange,
                                            restrictRange)
 import PostgREST.SchemaCache              (SchemaCache (..))
 import PostgREST.SchemaCache.Identifiers  (FieldName,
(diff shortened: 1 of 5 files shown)


src/PostgREST/Plan.hs
```diff
@@ -123,7 +125,7 @@ treeRestrictRange _ (ActionMutate _) request = Right request
 treeRestrictRange maxRows _ request = pure $ nodeRestrictRange maxRows <$> request
   where
     nodeRestrictRange :: Maybe Integer -> ReadPlan -> ReadPlan
-    nodeRestrictRange m q@ReadPlan{range_=r} = q{range_=restrictRange m r }
+    nodeRestrictRange m q@ReadPlan{range_=r} = q{range_= if hasLimitZero r then limitZeroRange else restrictRange m r }
```
steve-chavez, line 128: Hm, I see the same logic on ApiRequest
https://github.com/PostgREST/postgrest/blob/c8e4f38984d9194d939bf8b3c31010974f560be4/src/PostgREST/ApiRequest.hs#L357

Could this be wrapped inside a function for the Ranges module?

steve-chavez: Also maybe a comment in the function on why is it needed

steve-chavez (approved):

PR #2562, opened 2022-11-05
fix: bump hasql-pool to 0.8.0.6

robx:
This update ensures that connections aren't lost if they throw an exception. Compare #2558.


robx:
Thanks for checking!

PR #2564, opened 2022-11-09
feat: add spread embeds

steve-chavez:
Allows unnesting json objects. Closes https://github.com/PostgREST/postgrest/issues/1233.

Instead of getting

```http
GET /grandchild_entities?select=name,child_entities(parent_name:name,entities(grandparent_name:name))&limit=3'

[
{"name":"grandchild entity 1","child_entities":{"parent_name":"child entity 1","entities":{"grandparent_name":"entity 1"}}},
{"name":"grandchild entity 2","child_entities":{"parent_name":"child entity 1","entities":{"grandparent_name":"entity 1"}}},
{"name":"grandchild entity 3","child_entities":{"parent_name":"child entity 2","entities":{"grandparent_name":"entity 1"}}}
]
```

You can get

```http
GET /grandchild_entities?select=name,..child_entities(parent_name:name,..entities(grandparent_name:name))&limit=3

[
{"name":"grandchild entity 1","parent_name":"child entity 1","grandparent_name":"entity 1"},
{"name":"grandchild entity 2","parent_name":"child entity 1","grandparent_name":"entity 1"},
{"name":"grandchild entity 3","parent_name":"child entity 2","grandparent_name":"entity 1"}
]
```

Only for many-to-one and one-to-one relationships.

### Tasks

- [x] Initial implementation and test
- [x] Refactor - DRY
- [x] Validate to-one relationship
- [x] More tests

### DIFF
--- a/test/spec/Feature/Query/SpreadQueriesSpec.hs
+++ b/test/spec/Feature/Query/SpreadQueriesSpec.hs
@@ -1,0 +1,86 @@
+module Feature.Query.SpreadQueriesSpec where
+
+import Network.Wai (Application)
+
+import Test.Hspec
+import Test.Hspec.Wai
+import Test.Hspec.Wai.JSON
+
+import Protolude  hiding (get)
+import SpecHelper
+
+spec :: SpecWith ((), Application)
+spec =
+  describe "spread embeds" $ do
+    it "works on a many-to-one relationship" $ do
+      get "/projects?select=id,..clients(client_name:name)" `shouldRespondWith`
+        [json|[
+          {"id":1,"client_name":"Microsoft"},
+          {"id":2,"client_name":"Microsoft"},
+          {"id":3,"client_name":"Apple"},
+          {"id":4,"client_name":"Apple"},
+          {"id":5,"client_name":null}
+        ]|]
+        { matchStatus  = 200
+        , matchHeaders = [matchContentTypeJson]
+        }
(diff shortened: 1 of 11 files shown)


test/spec/Feature/Query/SpreadQueriesSpec.hs
```diff
@@ -0,0 +1,86 @@
...
+        , matchHeaders = [matchContentTypeJson]
+        }
+
+    it "fails when is not a to-one relationship" $ do
+      get "/clients?select=*,..projects(*)" `shouldRespondWith`
+        [json|{
+          "code":"PGRST118",
+          "details":null,
+          "hint":null,
+          "message":"'clients' and 'projects' do not form a many-to-one or one-to-one relationship"
```
laurenceisla, line 72: Maybe the error should be more specific in the message? And move the explanation to the "details". Something like:

```suggestion
          "code":"PGRST118",
          "details":"'clients' and 'projects' do not form a many-to-one or one-to-one relationship",
          "hint":null,
          "message":"Spread operator '..' is not compatible with this relationship"
```

And use another message for the related orders to differentiate them.

laurenceisla (approved):

steve-chavez:
The spread operator(`...`) is even used in GraphQL here: https://spec.graphql.org/June2018/#sec-Language.Fragments.

> But two dots seems fine.

I think it would be better to keep that syntax since it's already a popular convention, so will go back to 3 dots in another PR.

Issue #2571, opened 2022-11-22
Wrong paramters when calling a RPC function give a misleading error message


ff6347:
Thank you for this great tool I really love it. I just hit a bump yesterday and wanted to report it to give some feedback:

---

Let's say I have a function called `login(user text, password text)` in my schema called `api` and I make a wrong call like this one:

```shell
 curl --request POST \
  --url http://localhost:3000/rpc/login \
  --header 'content-type: application/json' \
  --data '{"user": "me", "pass": "123456"}'
```

I get the following response.

```
{'hint': 'If a new function was created in the database with this name and parameters, try reloading the schema cache.', 'message': 'Could not find the api.login() function in the schema cache'}
```

This mislead me to think that I need to reload the cache, restart the instance but did not hint me to check if my parameters are the right ones. 

### Environment

* PostgREST version: latest
* Operating system: linux

steve-chavez:
With https://github.com/PostgREST/postgrest/pull/2575 now you have a clearer suggestion!

```bash
 curl --request POST \
  --url http://localhost:3000/rpc/login \
  --header 'content-type: application/json' \
  --data '{"user": "me", "pass": "123456"}'
```

```json
{
 "code":"PGRST202","details":null,
 "hint":"Perhaps you meant to call the function test.login(password, user)",
 "message":"Could not find the test.login(pass, user) function or the test.login function with a single unnamed json or jsonb parameter in the schema cache"}
```

@laurenceisla Btw, I've noted that the order of function parameters changes on the hint, which could cause some confusion. Perhaps we could obtain the position of the parameter on `ProcParam` and use it on the hint?

ff6347:
Awesome! I love it.

Issue #2451, opened 2022-08-27
When a function is not found, show the overloaded functions in error hint


laurenceisla:
### Environment

* PostgreSQL version: 14
* PostgREST version: 10
* Operating system: EndeavourOS

### Description of issue

I think it would be nice to show all possible overloaded functions in the error `hint` when no function is found. Right now it shows that the `function(param type)` does not exist and, while it's useful, showing other possibilities may help the users chose the correct function.

Not sure if it's a problem to do so because all functions regardless of permissions will be shown.

steve-chavez:
> Not sure if it's a problem to do so because all functions regardless of permissions will be shown.

Hm, if we were to do this correctly we'd need an array of roles on the `ProcDescription` and check if the JWT role is in that array. Seems complicated.

How about just checking if the EXECUTE privilege is PUBLIC(meaning all roles) on the function? Then you can add the suggestions to the hint.

steve-chavez:
Experimenting with PostgreSQL we realized that it suggests the correct column name despite not having privileges on it:

```json
{
    "code": "42703",
    "details": null,
    "hint": "Perhaps you meant to reference the column \"films.rating\".",
    "message": "column films.ratin does not exist"
}
```

Which should be fine since using the correct column name will fail at the SQL level.

However we could configure the error verbosity later on to suppress all hints/details and give terse error messages.

PR #2575, opened 2022-11-25
Add hint for similar function parameter names when they don't match

laurenceisla:
Related to #2571 and #2451.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -17,7 +17,7 @@ module PostgREST.Error
 
 import qualified Data.Aeson                as JSON
 import qualified Data.ByteString.Char8     as BS
-import qualified Data.FuzzySet             as F
+import qualified Data.FuzzySet             as Fuzzy
 import qualified Data.HashMap.Strict       as HM
 import qualified Data.Text                 as T
 import qualified Data.Text.Encoding        as T
@@ -200,9 +200,11 @@ noRpcHint schema procName params allProcs overloadedProcs =
         (("Perhaps you meant to call the function " <> schema <> ".") <>)
         possibleProcs
   where
+    fuzzySetOfProcs  = Fuzzy.fromList [qiName k | k <- HM.keys allProcs, qiSchema k == schema]
+    fuzzySetOfParams = Fuzzy.fromList ["(" <> T.intercalate ", " [ppName prm | prm <- pdParams ov] <> ")" | ov <- overloadedProcs]
     possibleProcs
-      | null overloadedProcs = F.getOne (F.fromList [qiName k | k <- HM.keys allProcs, qiSchema k == schema]) procName
-      | otherwise            = (procName <>) <$> F.getOne (F.fromList ["(" <> T.intercalate ", " [ppName prm | prm <- pdParams ov] <> ")" | ov <- overloadedProcs]) params
+      | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
+      | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams params
 
 compressedRel :: Relationship -> JSON.Value
 -- An ambiguousness error cannot happen for computed relationships TODO refactor so this mempty is not needed
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -187,6 +187,7 @@ instance JSON.ToJSON ApiRequestError where
         _                            -> fmtParams) <>
       " in the schema cache"),
     "details" .= JSON.Null,
+    -- The hint will be null in the case of single unnamed parameter functions
     "hint"    .= if hasPreferSingleObject || (isInvPost && contentType `elem` [MTTextPlain, MTTextXML, MTOctetStream])
                  then Nothing
                  else noRpcHint schema procName argumentKeys allProcs overloadedProcs ]
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -45,6 +45,8 @@ import PostgREST.SchemaCache.Relationship (Cardinality (..),
                                            Relationship (..))
 import Protolude
 
+-- >>>
+-- >>>
 
 class (JSON.ToJSON a) => PgrstError a where
   status   :: a -> HTTP.Status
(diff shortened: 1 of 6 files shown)


src/PostgREST/Error.hs
```diff
@@ -17,6 +17,8 @@ module PostgREST.Error
 
 import qualified Data.Aeson                as JSON
 import qualified Data.ByteString.Char8     as BS
+import qualified Data.FuzzySet             as F
```
steve-chavez, line 20: To make the code more clear
```suggestion
import qualified Data.FuzzySet             as Fuzzy
```

src/PostgREST/Error.hs
```diff
@@ -183,13 +186,24 @@ instance JSON.ToJSON ApiRequestError where
...
+-- TODO: adapt for a Content-Type other than application/json or limit only to that header
+noRpcHint :: Text -> Text -> Text -> ProcsMap -> [ProcDescription] -> Text
+noRpcHint schema procName params allProcs overloadedProcs =
+  maybe ("No functions with a similar name found in the schema " <> schema) -- Should be null and adapt to function found but no similar parameters found
+        (("Perhaps you meant to call the function " <> schema <> ".") <>)
+        possibleProcs
+  where
+    possibleProcs
+      | null overloadedProcs = F.getOne (F.fromList [qiName k | k <- HM.keys allProcs, qiSchema k == schema]) procName
+      | otherwise            = (procName <>) <$> F.getOne (F.fromList ["(" <> T.intercalate ", " [ppName prm | prm <- pdParams ov] <> ")" | ov <- overloadedProcs]) params
```
steve-chavez, line 205: Some variables would make this clearer

src/PostgREST/Error.hs
```diff
@@ -37,12 +39,14 @@ import qualified PostgREST.MediaType        as MediaType
...
 import PostgREST.SchemaCache.Proc         (ProcDescription (..),
-                                           ProcParam (..))
+                                           ProcParam (..), ProcsMap)
 import PostgREST.SchemaCache.Relationship (Cardinality (..),
                                            Junction (..),
                                            Relationship (..))
 import Protolude
 
+-- >>>
+-- >>>
```
steve-chavez, line 49: Hm, why is this needed? Can it be deleted?

laurenceisla: Oops, forgot to remove those.

src/PostgREST/Error.hs
```diff
@@ -170,26 +174,78 @@ instance JSON.ToJSON ApiRequestError where
...
+-- in the same schema and shows the best match as hint.
+--
+-- >>> :set -Wno-missing-fields
+-- >>> let procsMap = HM.fromList [(QualifiedIdentifier "api" "test" , []), (QualifiedIdentifier "api" "another" , []), (QualifiedIdentifier "private" "other" , [])]
+--
+-- >>> noRpcHint "api" "testt" "(val, param, name)" procsMap []
+-- Just "Perhaps you meant to call the function api.test"
+--
+-- >>> noRpcHint "api" "other" "()" procsMap []
+-- Just "Perhaps you meant to call the function api.another"
```
steve-chavez, line 213: Ah, this one doesn't work because `other` is on a non-exposed schema.

src/PostgREST/Error.hs
```diff
@@ -170,26 +174,78 @@ instance JSON.ToJSON ApiRequestError where
...
+--
+-- >>> noRpcHint "api" "noclosealternative" "()" procsMap []
+-- Nothing
+--
+-- If a function is found with the given name, but no params match, then it does a fuzzy searh
```
steve-chavez, line 218:
```suggestion
-- If a function is found with the given name, but no params match, then it does a fuzzy search
```

src/PostgREST/Error.hs
```diff
@@ -170,26 +174,78 @@ instance JSON.ToJSON ApiRequestError where
...
+-- to all the overloaded functions' params using the form "(param1, param2, param3, ...)"
+-- and shows the best match as hint.
+--
+-- >>> let procsDesc = [ProcDescription {pdParams = [ProcParam {ppName="val"}, ProcParam {ppName="param"}, ProcParam {ppName="name"}]}, ProcDescription {pdParams = [ProcParam {ppName="id"}, ProcParam {ppName="attr"}]}]
+--
+-- >>> noRpcHint "api" "test" "(vall, pqaram, nam)" procsMap procsDesc
+-- Just "Perhaps you meant to call the function api.test(val, param, name)"
```
steve-chavez, line 225: This is really cool!

Q: Why do the params use a Text type`"(vall, pqaram, nam)"`? Why can't they be a List?

laurenceisla: The text type is needed because fuzzy search looks for a Text in a set of possible matches, that's why the solution I saw was formatting the parameters as Text. So, behind the scenes it does this:

`Fuzzy.getOne (Fuzzy.fromList ["(val, param, name)", "(id, attr)"])  "(vall, pqaram, nam)"`

And returns the first one directly to be used by the hint. Although, the arguments could be formatted into a Text inside the function itself, but I wanted to reuse the `prms` function that already did that.

steve-chavez: Hm, I still don't understand. From the [fuzzy set docs](https://hackage.haskell.org/package/fuzzyset-0.2.3/docs/Data-FuzzySet.html) I don't see that they use that format.
 
```haskell
Fuzzy.getOne (Fuzzy.fromList ["(val, param, name)", "(id, attr)"]) "(vall, pqaram, nam)"
```

So the fuzzy search is ignoring the parentheses?

src/PostgREST/Error.hs
laurenceisla: Yes.

src/PostgREST/Error.hs
laurenceisla: Yes, it's doing that. I separated that login in other function and added a comment to hopefully make it clearer.

src/PostgREST/Error.hs
```diff
@@ -170,26 +171,82 @@ instance JSON.ToJSON ApiRequestError where
...
-        (_, True, MTApplicationJSON) -> prms <> " function or the " <> schema <> "." <> procName <>" function with a single unnamed json or jsonb parameter"
-        _                            -> prms <> " function") <>
+        (_, True, MTApplicationJSON) -> fmtParams <> " or the " <> func <>" function with a single unnamed json or jsonb parameter"
+        _                            -> fmtParams) <>
       " in the schema cache"),
     "details" .= JSON.Null,
-    "hint"    .= ("If a new function was created in the database with this name and parameters, try reloading the schema cache." :: Text)]
+    "hint"    .= if hasPreferSingleObject || (isInvPost && contentType `elem` [MTTextPlain, MTTextXML, MTOctetStream])
+                 then Nothing
+                 else noRpcHint schema procName argumentKeys allProcs overloadedProcs ]
```
steve-chavez, line 192: Here the assumption is that if we're in the case of the single unnamed params we don't add the `hint` right? This should be clarified with a comment.

steve-chavez (approved):

PR #2583, opened 2022-12-10
Add suggestions with fuzzy text search when no relationship is found

### LINKED ISSUE #2569: reload the schema cache error message is misleading
Whenever https://postgrest.org/en/stable/schema_cache.html#automatic-schema-cache-reloading is enabled

```bash
$ curl 'localhost:3000/x?select=a,y(*)'

{
"code":"PGRST200",
"details":null,
"hint":"Verify that 'x' and 'y' exist in the schema 'test' and that there is a foreign key relationship between them.
If a new relationship was created, try reloading the schema cache.",
"message":"Could not find a relationship between 'x' and 'y' in the schema cache"
}
```

The `try reloading the schema cache` message is misleading. Many users take the hint literally without checking for typos or other errors in their request - then report that the reload is not working. This creates a lot of support burden.

The message could be avoided if we detect the event triggers and the notification channel are set.

---

The error messages will become more relevant once https://github.com/PostgREST/postgrest/pull/2542 is merged, see https://github.com/PostgREST/postgrest/pull/2542#pullrequestreview-1163724719.

laurenceisla:
Closes #2569 

Contemplating two options:

- Suggest either the `parent` or `child` one at a time. If no parent is found, then suggest the closest match and stop there, otherwise, continue  looking for the child and suggest its closest match.
- ~Suggest both `parent` and `child`, if neither is found then suggest the closest relationship between them.~

Done so far:

- [x] Simple suggestions for parents or children (one at a time)
- [x] Doctests
- [x] Spectests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #2565, Fix bad M2M embedding on RPC - @steve-chavez
- - #2575, Replace misleading error message by function and parameter name suggestions when no function is found - @laurenceisla
- - #2569, Replace misleading error message by parent and child name suggestions when no relationship is found - @laurenceisla
+ - #2575, Replace misleading error message when no function is found with a hint containing functions/parameters names suggestions - @laurenceisla
+ - #2569, Replace misleading error message when no relationship is found with a hint containing parent/child names suggestions - @laurenceisla
 ## [10.1.1] - 2022-11-08
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #2565, Fix bad M2M embedding on RPC - @steve-chavez
+ - #2575, Replace misleading error message by function and parameter name suggestions when no function is found - @laurenceisla
  - #2569, Replace misleading error message by parent and child name suggestions when no relationship is found - @laurenceisla
 ## [10.1.1] - 2022-11-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #2565, Fix bad M2M embedding on RPC - @steve-chavez
-
+ - #2569, Replace misleading error message by parent and child name suggestions when no relationship is found - @laurenceisla
 ## [10.1.1] - 2022-11-08
 
 ### Fixed
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -176,10 +176,10 @@ instance JSON.ToJSON ApiRequestError where
     "details" .= ("Only is null or not is null filters are allowed on embedded resources":: Text),
     "hint"    .= JSON.Null]
 
-  toJSON (NoRelBetween parent child hint schema allRels) = JSON.object [
+  toJSON (NoRelBetween parent child embedHint schema allRels) = JSON.object [
     "code"    .= SchemaCacheErrorCode00,
     "message" .= ("Could not find a relationship between '" <> parent <> "' and '" <> child <> "' in the schema cache" :: Text),
-    "details" .= ("Searched for a foreign key relationship between '" <> parent <> "' and '" <> child <> maybe "" ("' using the hint '" <>) hint <> "' in the schema '" <> schema <> "', but no matches were found."),
+    "details" .= ("Searched for a foreign key relationship between '" <> parent <> "' and '" <> child <> maybe mempty ("' using the hint '" <>) embedHint <> "' in the schema '" <> schema <> "', but no matches were found."),
     "hint"    .= noRelBetweenHint parent child schema allRels]
 
   toJSON (AmbiguousRelBetween parent child rels) = JSON.object [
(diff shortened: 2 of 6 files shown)


src/PostgREST/Error.hs
```diff
@@ -167,11 +169,11 @@ instance JSON.ToJSON ApiRequestError where
     "details" .= ("'" <> origin <> "' and '" <> target <> "' do not form a many-to-one or one-to-one relationship" :: Text),
     "hint"    .= JSON.Null]
 
-  toJSON (NoRelBetween parent child schema) = JSON.object [
+  toJSON (NoRelBetween parent child hint schema allRels) = JSON.object [
     "code"    .= SchemaCacheErrorCode00,
     "message" .= ("Could not find a relationship between '" <> parent <> "' and '" <> child <> "' in the schema cache" :: Text),
-    "details" .= JSON.Null,
-    "hint"    .= ("Verify that '" <> parent <> "' and '" <> child <> "' exist in the schema '" <> schema <> "' and that there is a foreign key relationship between them. If a new relationship was created, try reloading the schema cache." :: Text)]
+    "details" .= ("Searched for a foreign key relationship between '" <> parent <> "' and '" <> child <> maybe "" ("' using the hint '" <>) hint <> "' in the schema '" <> schema <> "', but no matches were found."),
```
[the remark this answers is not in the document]
steve-chavez:
```suggestion
    "details" .= ("Searched for a foreign key relationship between '" <> parent <> "' and '" <> child <> maybe mempty ("' using the hint '" <>) hint <> "' in the schema '" <> schema <> "', but no matches were found."),
```

src/PostgREST/Error.hs
```diff
@@ -214,6 +216,51 @@ instance JSON.ToJSON ApiRequestError where
...
+-- Nothing
+--
+-- >>> noRelBetweenHint "films" "noclosealternative" "api" rels
+-- Nothing
+--
+-- >>> noRelBetweenHint "films" "noclosealternative" "noclosealternative" rels
+-- Nothing
+--
+noRelBetweenHint :: Text -> Text -> Text -> RelationshipsMap -> Maybe Text
+noRelBetweenHint parent child schema allRels = ("Perhaps you meant '" <>) <$>
```
steve-chavez, line 252:
```suggestion
noRelBetweenHint :: Text -> Text -> Schema -> RelationshipsMap -> Maybe Text
noRelBetweenHint parent child schema allRels = ("Perhaps you meant '" <>) <$>
```

src/PostgREST/Error.hs
```diff
@@ -214,6 +216,51 @@ instance JSON.ToJSON ApiRequestError where
...
+  if isJust findParent
+    then (<> "' instead of '" <> child <> "'.") <$> suggestChild
+    else (<> "' instead of '" <> parent <> "'.") <$> suggestParent
+  where
+    findParent = HM.lookup (QualifiedIdentifier schema parent, schema) allRels
+    fuzzySetOfParents  = Fuzzy.fromList [qiName (fst p) | p <- HM.keys allRels, snd p == schema]
```
steve-chavez, line 258: I wonder if this is expensive if `allRels` is big, in which case perhaps the hint could be disabled by looking at the map length?

Maybe this could be tested with the apflora schema, by doing `time curl ..` on a request with bad names.

src/PostgREST/Error.hs
```diff
@@ -214,6 +216,51 @@ instance JSON.ToJSON ApiRequestError where
     "details" .= JSON.Null,
     "hint"    .= ("Try renaming the parameters or the function itself in the database so function overloading can be resolved" :: Text)]
 
+-- |
+-- If no relationship is found between a parent and a child, then it looks for the parent first.
+-- If the parent is not found then it does a fuzzy search to all the parents in the schema cache and
+-- gives the best match as suggestion. Otherwise, it does a fuzzy search to all the corresponding children
+-- of that parent and gives the best match as suggestion. If both are found, then no suggestion is given.
```
steve-chavez, line 223: Just to make sure I'm understanding this.

If no rel is found:

- Looks for parent suggestions if parent not found
- Looks for child suggestions if children not found and parent is found
- No suggestions if both are found

Is that correct?

steve-chavez:
> Suggest both parent and child, if neither is found then suggest the closest relationship between them.

That one is really tricky and likely not worth it. Good that you didn't pursue it further.

src/PostgREST/Error.hs
laurenceisla: It is much slower than when a relationship is found, not sure if it's too slow, though. For instance these are the times that I got:

1. Relationship is found:
    ```bash
    $ curl -w "%{time_total}s" -o /dev/null -s "http://localhost:3000/ziel?select=ap(*)"
    0.002485s
    0.002125s
    ```

2. Relationship not found (parent not found)
    ```bash
    $ curl -w "%{time_total}s" -o /dev/null -s "http://localhost:3000/ziels?select=ap(*)"
    0.072776s
    0.056077s
    ``` 

3. Relationship not found (child not found)
    ```bash
    $ curl -w "%{time_total}s" -o /dev/null -s "http://localhost:3000/ziel?select=app(*)"
    0.035553s
    0.027014s
    ```

src/PostgREST/Error.hs
laurenceisla: Yes, if both are found then there's an error with the hint, which is more complex to look for.

src/PostgREST/Error.hs
steve-chavez: > (fuzzy search for parent)
> 0.072776s
> 0.056077s
> (fuzzy search for child)
> 0.035553s
>0.027014s

So parent takes longer than child because it searches on the full `allRels`, while child only searches on the relationships that have parent. Makes sense.

CHANGELOG.md
```diff
@@ -23,7 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #2565, Fix bad M2M embedding on RPC - @steve-chavez
-
+ - #2575, Replace misleading error message by function and parameter name suggestions when no function is found - @laurenceisla
+ - #2569, Replace misleading error message by parent and child name suggestions when no relationship is found - @laurenceisla
```
steve-chavez, line 27:
```suggestion
 - #2575, Replace misleading error message when no function is found with a hint containing functions/parameters names suggestions - @laurenceisla
 - #2569, Replace misleading error message when no relationship is found with a hint containing parent/child names suggestions - @laurenceisla
```

PR #2604, opened 2023-01-09
build with nix on macos in CI (and push to cachix)

robx:
The primary point is to cache the nix environment for macos development, too. Running some tests *might* be a good idea, too.


robx:
Closing in favor of #2610.

PR #2608, opened 2023-01-10
ci: authenticate GitHub API request (fixes #2607)

### LINKED ISSUE #2607: "Get FreeBSD ..." CI job fails due to GitHub API rate limiting
We get frequent CI failures on that one job.

The `curl` calls to the GitHub API in https://github.com/PostgREST/postgrest/blob/main/.github/get_cirrusci_freebsd fail with

```
HTTP/2 403
...
x-ratelimit-limit: 60
x-ratelimit-remaining: 0
x-ratelimit-reset: 1673356097
x-ratelimit-resource: core
x-ratelimit-used: 60
...

{
  "message":"API rate limit exceeded for 104.45.204.51. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"
}
```

(see https://github.com/PostgREST/postgrest/actions/runs/3883439292/jobs/6624750236, from the debugging PR https://github.com/PostgREST/postgrest/pull/2606)

`x-ratelimit-limit: 60` indicates that 60 requests are allowed per hour. The script retries every 60s after one extra initial call, so we're likely to hit the limit ourselves, even disregarding possible other jobs running on the same runner.

Some things we could do:
- sleep longer
- sleep and retry on 403 / rate limit response
- authenticate the request (probably the best solution)

In addition, it would be neat if the `curl` would write the error message to stderr if they fail.

robx:
It turns out this is (hopefully) easier than I thought, because we already have an API token available. Let's see...

### DIFF
--- a/.github/get_cirrusci_freebsd
+++ b/.github/get_cirrusci_freebsd
@@ -7,12 +7,13 @@ set -euo pipefail
 # https://docs.github.com/en/rest/reference/checks#list-check-suites-for-a-git-reference
 
 cirrus_artifact_name=bin
+gh_auth_header="Authorization: Bearer $GITHUB_TOKEN"
 gh_accept_header="Accept: application/vnd.github.v3+json"
 
 get_gh_check_runs_url() {
   gh_checks_list_url="https://api.github.com/repos/$GITHUB_REPOSITORY/commits/$GITHUB_COMMIT/check-suites"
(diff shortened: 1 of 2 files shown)


steve-chavez (approved):
:100: :100: :100:

PR #2611, opened 2023-01-12
ci: use postgrest-push-cachix independently of other tools

### LINKED ISSUE #2609: nix: building postgrest-push-cachix shouldn't require various postgresql versions to be built
Due to the bundling of `postgrest-push-cachix` in a "build tool box" `devTools` which pulls in `withTools` for `withGit`, where `withTools` itself is a "build tool box" that also includes the various `withPg` tools which depend on postgresql versions, it's not possible to run postgrest-push-cachix before building essentially the whole dependency tree. That's a mess in general, but particularly it makes it hard to push partial build results to cachix to seed cachix when we run into CI timeouts for big upgrades.

It seems to me that the bundling into build toolboxes at all is a poor approach except at the highest level. I.e., `postgrest-push-cachix` should be exposed directly as an attribute of `default.nix`, and it should depend on `withGit` directly instead of on `withTools`.

robx:
This cuts down the dependency footprint of postgrest-push-cachix massively, allowing us to invoke it for incrementally updating cachix from CI.

This fixes #2609 (in a sense -- it doesn't really make anything simpler and things still feel a bit overengineered, but...).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -135,7 +135,7 @@ jobs:
         if: ${{ github.ref == 'refs/heads/main' }}
         run: |
           nix-build
-          nix-env -f default.nix -iA devTools
+          nix-env -f default.nix -iA devTools.pushCachix.bin
           postgrest-push-cachix
 
 
@@ -152,7 +152,7 @@ jobs:
       - name: Build everything
         run: |
           nix-build
-          nix-env -f default.nix -iA devTools
+          nix-env -f default.nix -iA devTools.pushCachix.bin
 
       - name: Push to Cachix (main branch only)
         if: ${{ github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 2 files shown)


robx:
It doesn't really fix the dependency situation satisfactorily. What I'm seeing now on #2612 is that `nix-env -iA devTools.pushCachix.bin` builds half the world including e.g. libgmp, presumably, because it uses `checkedShellCheck` which relies on `shellcheck` which is a Haskell tool that's presumably rebuilt compared to upstream nixpkgs because it comes from the nixpkgs with our postgrest-specific Haskell overrides.

steve-chavez:
Would suggest the same as on https://github.com/PostgREST/postgrest/pull/2612#issuecomment-1381023695 - temporarily drop `shellcheck`.

robx:
Closing in favour of #2615.

PR #2625, opened 2023-01-20
fix: consider authentication failure as a fatal error

### LINKED ISSUE #2622: PostgreSQL Authentication failure not considered as fatal
### Environment

* PostgreSQL version: `postgres:12.11-alpine`
* PostgREST version: `postgrest/postgrest:v10.0.0` (also reproduced with compiled v10.0.0 and the latest HEAD version at the time of writing be6c30a0fbe11eee4c680ef2dda635370e59a999)
* Operating system: NixOS unstable

### Description of issue

It seems that PostgREST is not detecting PostgreSQL authentication error properly and keeps on trying when it is supposed to consider it as a fatal error and exit immediately.

### Technical info

Currently, when PostgREST tries to log into PostgreSQL with an incorrect password, it will just wait and retry later. The following are logs with the HEAD revision:

```
backend-postgrest-1  | 19/Jan/2023:18:38:30 +0000: {"code":"PGRST000","details":"connection to server at \"postgres\" (172.18.0.2), port 5432 failed: FATAL:  password authentication failed for user \"xxx\"\n","hint":null,"message":"Database connection error. Retrying the connection."}
backend-postgrest-1  | 19/Jan/2023:18:38:30 +0000: Attempting to reconnect to the database in 1 seconds...
```

It seems that the intended behavior was to fail and exit. Here is the associated code:

https://github.com/PostgREST/postgrest/blob/be6c30a0fbe11eee4c680ef2dda635370e59a999/src/PostgREST/Error.hs#L449-L454

It seems the code tries to identify authentication failure by checking if the error message starts with `FATAL:  password authentication failed`. But the authentication failure message in the latest version seems to be `connection to server at "postgres" (172.18.0.2), port 5432 failed: FATAL:  password authentication failed for user[...]`. So authentication failures are never matched.

### Expected behavior

A possible solution would be: since there's no corresponding error code, try and check for `FATAL:  password authentication failed` in the whole error message. Using `isInfixOf` for example seems to result in properly identified authentication failures and immediate exit. It would still be vulnerable to change in the message though.

Also, there is the issue of whether authentication failure is fatal or not. Possible scenarios:

- Consider authentication failure as fatal, having PostgREST exit once it fails to log in. This would be ideal for our use case. But it will break backward compatibility in a way.
- Consider authentication failure as temporary and wait to retry again. It would preserve backward compatibility but can hide infrastructure issues, one that we want to detect specifically.
- Use either one of those scenario by introducing a flag that will determine the outcome of the authentication failure. Ideal but requires more change.

### To reproduce

1. Write a PostgREST configuration, pointing to a valid PostgreSQL database but with an invalid password. For example:

```
db-uri = "postgresql://authenticator:something@localhost:5432/mydb"
db-schemas = "something"
db-anon-role = "web_anon"
server-port = 3000
admin-server-port = 3001
```
2. Run PostgREST, pointing it to the configuration file
3. Observe that it fails to connect and retries

michivi:
Fixes #2622

To detect whether PostgreSQL authentication has failed or not, the `FATAL: password authentication failed"` is looked throughout the whole message details. It was only tested at the beginning of the message beforehand and the format of the message has changed at some point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+ - #2622, Consider any PostgreSQL authentication failure as fatal and exit immediately - @michivi
+
 ### Added
 
  - #1414, Add related orders - @steve-chavez
--- a/nix/tools/withTools.nix
+++ b/nix/tools/withTools.nix
@@ -54,6 +54,10 @@ let
         export PGDATABASE
         export PGRST_DB_SCHEMAS
 
+        HBA_FILE="$tmpdir/pg_hba.conf"
+        echo "local $PGDATABASE some_protected_user password" > "$HBA_FILE"
+        echo "local $PGDATABASE all trust" >> "$HBA_FILE"
+
         log "Initializing database cluster..."
         # We try to make the database cluster as independent as possible from the host
         # by specifying the timezone, locale and encoding.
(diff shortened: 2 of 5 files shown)


steve-chavez (approved):
Perfect :100:. Thanks!

Issue #1601, opened 2020-10-03
Discard columns with null values


jsommr:
Say I have a polymorphic table `contact` that can be either `person` or `company` (each in their own table):

```sql
create type contact_type as enum (
  'company',
  'person'
);

create table contact (
  contact_id   uuid         primary key default uuid_time_nextval(),
  contact_type contact_type not null
);

create table company (
  company_id   uuid primary key references contact,
  company_name text not null
);

create table person (
  person_id      uuid primary key references contact,
  preferred_name text not null,
  email          text not null
);
```

With insert triggers on company and person ensuring a contact with the same id is inserted as well.

We create a view that combines them:

```sql
create view v1.contact as
  select 
    contact.*,
    c.company_name,
    p.preferred_name,
    p.email
  from contact
  left join company c on contact_id = company_id
  left join person p on contact_id = person_id;
```

With a little sample data, requesting this via `/contact?select=*` returns:

```json
[
  {
    "contact_id": "546137e0-cda5-4e24-b684-94d030bfcf42",
    "contact_type": "company",
    "company_name": "Nerfpops Incorporated",
    "preferred_name": null,
    "email": null
  },
  {
    "contact_id": "54772635-7feb-438a-b4ee-251e23684e51",
    "contact_type": "person",
    "company_name": null,
    "preferred_name": "Nerfpops",
    "email": "email@example.com"
  }
]
```

Is there any way to tell PostgREST not to include fields that are null in the JSON response? Even if I explicitly asked for the field, I'd still prefer not to have it in the response.

steve-chavez:
Hey @nerfpops. Looks like this might be possible with [json_strip_nulls](https://www.postgresql.org/docs/current/functions-json.html#id-1.5.8.22.5.11.2.2.16.1.1.1).

Right now we use a query similar to:
```sql
WITH
pgrst_source AS (SELECT "test"."projects".* FROM "test"."projects"    )
SELECT coalesce(json_agg(_postgrest_t), '[]')::character varying AS body
FROM ( SELECT * FROM pgrst_source) _postgrest_t;

-- and this gives
[{"id":1,"name":"Windows 7","client_id":1},
 {"id":2,"name":"Windows 10","client_id":1},
 {"id":3,"name":"IOS","client_id":2},
 {"id":4,"name":"OSX","client_id":2},
 {"id":103,"name":"hey","client_id":null},
 {"id":104,"name":"hey","client_id":null},
[...]

PR #2637, opened 2023-01-26
Add feature : Discard columns with null values v2

AnthonyFisi:
I will implement a solutions to discard columns with null values and my reference is the following issue #1601

### DIFF
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -53,7 +53,7 @@ data MTNullAttrs = MTNullAttrs MTNullFormat deriving Show
 instance Eq MTNullAttrs where
   MTNullAttrs {} == MTNullAttrs{} = True
 
-data MTNullFormat 
+data MTNullFormat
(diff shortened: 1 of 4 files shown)


src/PostgREST/MediaType.hs
```diff
@@ -22,6 +22,7 @@ import Protolude
 -- | Enumeration of currently supported media types
 data MediaType
   = MTApplicationJSON
+  | MTApplicationNullJSON
```
laurenceisla, line 25: This should have another name, maybe `MTArrayJSON` similar to `MTSingularJSON`. The thing is that the `;nulls=stripped` attribute will also be added to the `MTSingularJSON` according to  https://github.com/PostgREST/postgrest/issues/1601#issuecomment-801038777, so you can add it as a field. You could check the `MTPlanAttrs` and follow it as example, it should be similar to that.

AnthonyFisi: Hi @laurenceisla i've followed your advice, i hope you can review my code.

Issue #2638, opened 2023-01-27
v10.1.1: Not having a pool timeout causes PostgreSQL connections to leak memory


steve-chavez:
## Problem

Idle connections can take up memory and not release it, here's an example `htop` output showing this:

![idle_conns](https://user-images.githubusercontent.com/1829294/215181422-35fcd32f-a719-48cf-b297-d85c261b39fb.png)

(`authenticator` is the connection role used by us, this follows the [docs](https://postgrest.org/en/stable/auth.html#authentication-sequence))

This gets serious whenever the `db-pool` size is big, the connections can consume all memory.

More details on why a connection takes up memory in this [SO question](https://stackoverflow.com/questions/20514569/idle-postgres-processes-taking-up-lots-of-memory). Basically it has a cache of the catalog and this gets bigger as more database objects are added.

Now, this wasn't an issue until v10.0.0, because we had a [db-pool-timeout](https://postgrest.org/en/stable/configuration.html#db-pool-timeout) which timed out idle connections. But on [v10.1.0](https://github.com/PostgREST/postgrest/releases/tag/v10.1.0) this was removed(https://github.com/PostgREST/postgrest/pull/2444) due to the upstream library not supporting it.

## Solution

We should bring back the timeout or alternatively seems we can use the [DISCARD ALL](https://www.postgresql.org/docs/current/sql-discard.html) statement whenever a connection is returned back to the pool(not sure of the drawbacks) according to this [AWS blog post](https://aws.amazon.com/blogs/database/resources-consumed-by-idle-postgresql-connections/).

cc @robx

steve-chavez:
[v10.0.0](https://github.com/PostgREST/postgrest/releases/tag/v10.0.0) might also be affected by this because the timeout was [increased to 1 hour](https://github.com/PostgREST/postgrest/pull/2326).

robx:
There's been quite a bit of discussion of this following this comment: https://github.com/PostgREST/postgrest/pull/2444#issuecomment-1302980367. It would be useful to summarize this on this ticket. E.g., I'm guessing that setting the session idle timeout wouldn't be sufficient because the connections might all stay busy?

(I find the title of the issue a bit misleading because at first read, it seems like the postgrest process is leaking memory, while instead the postgresql server processes are growing caches indefinitely.)

- As noted in the linked discussion, periodically releasing the pool could be an option, at least as a work-around (postgrest could do this in-process on a timer)
[...]

Issue #2592, opened 2022-12-20
Missing Swagger Prefer Option Preventing Upserts via Swagger UI


HuntosShuntos:
### Environment

* PostgreSQL version: 15.1-1 (Image postgres:15)
* PostgREST version: Docker v10.1.1 (Image Id: 9fef4519eb00e2c359fa0dafecacf389b14ee1c523bf4c134146bb5e830e4235)
* Operating system: macOS 12.5.1

### Description of issue

Swagger Documentation for post requests don't have the drop down to enable upserts. Currently missing an option for `Prefer: resolution=merge-duplicates`. Expected behaviour is to be able to perform an upsert via Swagger UI 

See below image

<img width="447" alt="image" src="https://user-images.githubusercontent.com/111670981/208721633-4d320f96-3298-4ad0-bc71-e7e7e3aca648.png">

PR #2582, opened 2022-12-09
Clarify error messages for functions

laurenceisla:
Moves the explanations about "single unnamed parameters" to the details instead of the message part of the error. It also removes any mention to resetting the schema cache from other error hints.

PR #2642, opened 2023-01-31
v10.1.2 PATCH release

steve-chavez:
v10.1.1 is currently not stable for production as mentioned on https://github.com/PostgREST/postgrest/issues/2638#issuecomment-1410864086.

We'll need to make a v10.1.2 that includes some fixes.

Included fixes:

- [x] #2565, Fix bad M2M embedding on RPC cb99270a8f14cfe3838408a5a3f9f9205e740323
- [x] #2575, Replace misleading error message when no function is found with a hint containing functions/parameters names suggestions 9e567216e9968134c814048375dce08bfb3878ca
- [x] #2582, Clarify error messages for functions aecc53d8f9ec8d80447a8c4d2854559adca22c9a
- [x] #2569, Replace misleading error message when no relationship is found with a hint containing parent/child names suggestions 5a660e0cbe405a28093366f0fb4b6c5ba6ac4eac
- [x] #1405, Add the required OpenAPI items object when the parameter is an array 1f697578358942949b66bf6f37de155d212a184e
- [x] #2592, Add upsert headers for POST requests to the OpenAPI output 4ddd5b4a76e076f8c34f56bdfc82e32ce3da613e
- [x] #2623, FK pointing to VIEW instead of TABLE in OpenAPI output 775c0068068a9612c06d3186acfdbf20c7414cb8
- [x] #2622, Consider any PostgreSQL authentication failure as fatal and exit immediately 8aa79086d2f0134b4f40e06cdf5cdc27f63eced3
- [x] #2620, Fix NOTIFY pgrst not reloading the db connections catalog cache 7874bee87927ed73ef8d32c9377c53ea97f44101

The only critical fix is really 7874bee87927ed73ef8d32c9377c53ea97f44101, so if some fixes turn out to be hard to merge(due to refactors and dependency to other commits) we can leave them for v11.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+ - #2565, Fix bad M2M embedding on RPC - @steve-chavez
+
 ## [10.1.1] - 2022-11-08
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #2565, Fix bad M2M embedding on RPC - @steve-chavez
-
+ - #2575, Replace misleading error message when no function is found with a hint containing functions/parameters names suggestions - @laurenceisla
+ - #2569, Replace misleading error message when no relationship is found with a hint containing parent/child names suggestions - @laurenceisla
 ## [10.1.1] - 2022-11-08
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+## [10.1.2] - 2023-02-01
 
 ### Fixed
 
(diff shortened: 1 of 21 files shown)


steve-chavez:
I realize this does cause an issue again for users of pre-releases, as mentioned on https://github.com/PostgREST/postgrest/pull/2296#issuecomment-1171165411

> The out-of-order 9.0.1 patch release causes our version numbers for pre-releases to break the "higher version number means more recent" promise.

I think a solution for this will be reserving a `minor` for pre-releases. 

So if we had a v10.2.1.20221215 instead of a v10.1.1.20221215 releasing a v10.1.2 would not break the above promise.

And then if we want to go for a minor release, we'd have to jump two numbers, from v10.1 to v10.3, I think this is fine. A major v11 will work as well.

Will make this fix in the `postgrest-release` command.

cc @wolfgangwalther

steve-chavez:
Sorry for the late reply here.

>This is basically the same thing that for example nodejs does - only that they do it with major releases: Only even major releases are considered stable. Odd major releases are considered experimental.
> What if we instead do the same?

I didn't know that, it's an interesting idea. But that would basically break semver right? Seems risky as we have been committing to semver and it will be surprising to users that a new version causes a breakage.

> As soon as we commit a new feature, we bump to 11.1.0.

That could be an option, just releasing more frequently. The downside is that that could cause something like "notification fatigue"(users will then ignore updates). Personally, I prefer the way PostgreSQL does releases, once a year I can keep up with learning the new stuff. That might be too extreme for us, but you get the idea.

> You'd still have the freedom to go back and release patch versions for the latest stable version. It happened twice already, so I don't think we can continue to pretend it will never happen.

Yeah, true. Right now I'm willing to put in the work and create branches from time to time, rebase the fixes and do releases there.

Another option could also be to never do releases off the main branch, only on other branches(like `rel-`). I've seen projects(like pgx) that commit to a `devel` branch(not `main`) a do the releases on other branches.

PR #2651, opened 2023-02-06
Add 'get' to the path item for functions

### LINKED ISSUE #1222: OpenApi not showing GET method for immutable functions

laurenceisla:
Closes #1222 (and maybe #1151?)

- [x] Add 'get' to the function's path item
- [x] Refactor repetitive code
- [x] Add tests
- [x] Add CHANGELOG

### DIFF
--- a/src/PostgREST/Response/OpenAPI.hs
+++ b/src/PostgREST/Response/OpenAPI.hs
@@ -73,9 +73,15 @@ toSwaggerType colType             = case T.takeEnd 2 colType of
   "[]" -> Just SwaggerArray
   _    -> Just SwaggerString
 
-makeSwaggerItemType :: Maybe (SwaggerType t) -> Text -> Maybe (Referenced Schema)
-makeSwaggerItemType itemType colType = case itemType of
-  Just SwaggerArray -> Just $ Inline (mempty & type_ .~ toSwaggerType (T.dropEnd 2 colType))
+typeFromArray :: Text -> Text
+typeFromArray = T.dropEnd 2
+
+toSwaggerTypeFromArray :: Text -> Maybe (SwaggerType t)
+toSwaggerTypeFromArray arrType = toSwaggerType $ typeFromArray arrType
+
+makeSwaggerItemTypeSchema :: Text -> Maybe (Referenced Schema)
+makeSwaggerItemTypeSchema arrType = case toSwaggerType arrType of
+  Just SwaggerArray -> Just $ Inline (mempty & type_ .~ toSwaggerTypeFromArray arrType)
   _                 -> Nothing
 
 parseDefault :: Text -> Text -> Text
@@ -129,7 +135,6 @@ makeProperty tbl rels col = (colName col, Inline s)
         Just $ T.append (maybe "" (`T.append` "\n\n") $ colDescription col) (T.intercalate "\n" n)
       else
         colDescription col
-    pType = toSwaggerType (colType col)
     s =
       (mempty :: Schema)
         & default_ .~ (JSON.decode . toUtf8Lazy . parseDefault (colType col) =<< colDefault col)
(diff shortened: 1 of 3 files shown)


src/PostgREST/Response/OpenAPI.hs
```diff
@@ -307,20 +308,23 @@ makePathItem t = ("/" ++ T.unpack tn, p $ tableInsertable t || tableUpdatable t
     ref = Ref . Reference
 
 makeProcPathItem :: ProcDescription -> (FilePath, PathItem)
-makeProcPathItem pd = ("/rpc/" ++ toS (pdName pd), pe)
+makeProcPathItem pd = ("/rpc/" ++ toS (pdName pd), pe $ pdVolatility pd == Volatile)
```
laurenceisla, line 311: I have a doubt about this. The only way to check the volatility of a function is through this parameter, but PostgreSQL makes all the functions `VOLATILE by default so the `get` param won't show for those functions even if the request using `GET` is allowed. 

If we go with this route the user will be obligated to specify the volatility of a function in order for the OpenAPI output to be accurate (needs to be mentioned in the docs). Or we could add the `get` parameter disregarding the volatility and risk getting `405 Method Not Allowed` even when `get` is present in the OpenAPI output.

steve-chavez: > so the get param won't show for those functions even if the request using GET is allowed.

You mean the function will still work through GET because it will be compatible with a read-only transaction right?

> Or we could add the get parameter disregarding the volatility and risk getting 405 Method Not Allowed even when get is present in the OpenAPI output.

Yeah, I guess this would be most correct because it's what we actually support.

---

If we did https://github.com/PostgREST/postgrest/issues/1826 and indeed disable GET/HEAD for volatile functions, then conditionally enabling GET would be accurate. But this can be done later and it'd be a breaking change.

src/PostgREST/Response/OpenAPI.hs
```diff
@@ -175,8 +175,25 @@ makePreferParam ts =
...
+    & name .~ n
+    & required ?~ r
+    & schema .~ ParamOther ((mempty :: ParamOtherSchema)
+      & in_ .~ ParamQuery
+      & type_ .~ toSwaggerType t
+      & items .~ if not v then Nothing else Just $ SwaggerItemsPrimitive
+          (Just CollectionMulti)
+          ((mempty :: ParamSchema x)
+            & type_ .~ toSwaggerType (T.dropEnd 2 t)
+            & format ?~ t))
```
laurenceisla, line 190: This allows for `variadic` parameters to be repeated in the Swagger UI. 

image

I tried to show the format for special types (for instance a SQL `ENUM`) as a visual help, but while adding it outside `items` shows the format, the parameters can no longer be repeated. Adding it inside `items` (like it is right now) doesn't change the behavior, but the visual help is not shown. I think I'll remove it in the refactor since it doesn't seem to help.

steve-chavez (approved):
Great work! :100:

steve-chavez:
Seems CI is failing because of https://github.com/haskell/actions/issues/187

Issue #2664, opened 2023-02-14
sporadic postgrest-test-spec failure in Feature.Query.PlanSpec


robx:
### Environment

* PostgreSQL version: 15 (via nix `withPg`)
* PostgREST version: current HEAD
* Operating system: macOS 12.5 (x86)

### Description of issue

I often see one test failure when running `postgrest-test-spec`.

```
Feature.Query.PlanSpec.spec
  read table/view plan
    outputs blocks info when using the buffers option [✘]
```

This fails because the value "Shared Hit Blocks" is non-zero, contrary to expectation. The precise value varies.

```
Failures:

  test/spec/Feature/Query/PlanSpec.hs:68:20: 
  1) Feature.Query.PlanSpec.spec, read table/view plan, outputs blocks info when using the buffers option
       expected: Just (Object (fromList [("Local Dirtied Blocks", Number 0.0), ("Local Hit Blocks", Number 0.0), ("Local Read Blocks", Number 0.0), ("Local Written Blocks", Number 0.0), ("Shared Dirtied Blocks", Number 0.0), ("Shared Hit Blocks", Number 0.0), ("Shared Read Blocks", Number 0.0), ("Shared Written Blocks", Number 0.0), ("Temp Read Blocks", Number 0.0), ("Temp Written Blocks", Number 0.0)]))
        but got: Just (Object (fromList [("Local Dirtied Blocks", Number 0.0), ("Local Hit Blocks", Number 0.0), ("Local Read Blocks", Number 0.0), ("Local Written Blocks", Number 0.0), ("Shared Dirtied Blocks", Number 0.0), ("Shared Hit Blocks", Number 39.0), ("Shared Read Blocks", Number 0.0), ("Shared Written Blocks", Number 0.0), ("Temp Read Blocks", Number 0.0), ("Temp Written Blocks", Number 0.0)]))

  To rerun use: --match "/Feature.Query.PlanSpec.spec/read table/view plan/outputs blocks info when using the buffers option/"

Randomized with seed 1855785399

Finished in 0.0948 seconds
```

robx:
Hmm, this shows in CI now on the PR #2663. I *think* I got it to happen reliably on `main`, but there's a chance a `hasql-pool` updated slipped in and that causes the test failures...

steve-chavez:
Hm, I thought these values would be stable when doing the tests. Maybe they should be changed to just test the keys and not the values..

steve-chavez:
[...]

PR #2665, opened 2023-02-14
nix: bump hasql-pool to 0.8.0.7

robx:
bump hasql-pool dependency to the latest hackage release to narrow down the test failure in #2664.


robx:
Closing. It turns out the test is broken regardless of hasql-pool

PR #2667, opened 2023-02-15
fix: log to stderr on AcquisitionTimeoutUsageError

steve-chavez:
When reaching the timeout on `db-pool-acquisition-timeout`, we don't log an error to stderr. This is useful for admins since they will know that the pool size has to be increased or that queries are taking too long.

- [x] Debounce(since many requests could trigger this error)
- [x] Test

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -16,6 +16,7 @@ module PostgREST.AppState
   , init
   , initWithPool
   , logWithZTime
+  , logPgrstError
   , putConfig
   , putSchemaCache
   , putIsListenerOn
@@ -27,8 +28,11 @@ module PostgREST.AppState
   , waitListener
   ) where
 
-import qualified Hasql.Pool    as SQL
-import qualified Hasql.Session as SQL
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.Text.Encoding   as T
+import qualified Hasql.Pool           as SQL
+import qualified Hasql.Session        as SQL
+import qualified PostgREST.Error      as Error
 
 import Control.AutoUpdate (defaultUpdateSettings, mkAutoUpdate,
                            updateAction)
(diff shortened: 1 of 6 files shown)


src/PostgREST/AppState.hs
```diff
@@ -47,29 +53,31 @@ import Protolude
...
-  , stateGetZTime     :: IO ZonedTime
+  , stateGetZTime                 :: IO ZonedTime
   -- | Used for killing the main thread in case a subthread fails
-  , stateMainThreadId :: ThreadId
+  , stateMainThreadId             :: ThreadId
   -- | Keeps track of when the next retry for connecting to database is scheduled
-  , stateRetryNextIn  :: IORef Int
+  , stateRetryNextIn              :: IORef Int
+  -- | Logs a pool error with a debounce
+  , debounceLogAcquisitionTimeout :: IO ()
```
laurenceisla, line 80: Is the naming consistency needed here? If so, perhaps it should be changed to something like `stateDebounceLogAcquisitionTO`.

src/PostgREST/AppState.hs
```diff
@@ -91,6 +99,17 @@ initWithPool pool conf =
     <*> mkAutoUpdate defaultUpdateSettings { updateAction = getZonedTime }
     <*> myThreadId
     <*> newIORef 0
+    <*> pure (pure ())
+
+  deb <-
+    let oneSecond = 1000000 in
+    mkDebounce defaultDebounceSettings
+       { debounceAction =  logPgrstError appState SQL.AcquisitionTimeoutUsageError
```
laurenceisla, line 107:
```suggestion
       { debounceAction = logPgrstError appState SQL.AcquisitionTimeoutUsageError
```

src/PostgREST/AppState.hs
steve-chavez: It's a bit different, the `debounceLogAcquisitionTimeout` function is meant to be called outside the module, this is also why it's exported.

The `state*` variables OTOH are not accessed directly but through other wrapper functions.

Ideally we should have another Logger module for this to be clearer, but adding it is much work for now.

laurenceisla (approved):

Issue #2627, opened 2023-01-23
Lack of `Show` on many types hurts explorability of the codebase


steve-chavez:
Say I want to explore some functions of the `MediaType` module with the cabal repl, the following will fail:

```
$ cabal v2-repl

ghci> import PostgREST.MediaType

ghci> decodeMediaType "application/json"

<interactive>:2:1: error:
    • No instance for (Show MediaType)
        arising from a use of ‘System.IO.print’
      There are instances for similar types:
        instance [safe] Show
                          http-media-0.8.0.0:Network.HTTP.Media.MediaType.Internal.MediaType
          -- Defined in ‘http-media-0.8.0.0:Network.HTTP.Media.MediaType.Internal’
    • In a stmt of an interactive GHCi command: System.IO.print it
```

If I add the required Show instances(manually, which turns into a chore quickly):

```diff
diff --git a/src/PostgREST/MediaType.hs b/src/PostgREST/MediaType.hs
index 39840a80..ebb5cad6 100644
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -33,17 +33,17 @@ data MediaType
   | MTAny
   | MTOther ByteString
   | MTPlan MTPlanAttrs
-  deriving Eq
+  deriving (Show, Eq)
 
-data MTPlanAttrs = MTPlanAttrs (Maybe MediaType) MTPlanFormat [MTPlanOption]
+data MTPlanAttrs = MTPlanAttrs (Maybe MediaType) MTPlanFormat [MTPlanOption] deriving Show
 instance Eq MTPlanAttrs where
   MTPlanAttrs {} == MTPlanAttrs {} = True -- we don't care about the attributes when comparing two MTPlan media types
 
 data MTPlanOption
-  = PlanAnalyze | PlanVerbose | PlanSettings | PlanBuffers | PlanWAL
+  = PlanAnalyze | PlanVerbose | PlanSettings | PlanBuffers | PlanWAL deriving Show
 
 data MTPlanFormat
-  = PlanJSON | PlanText
+  = PlanJSON | PlanText deriving Show
 
 -- | Convert MediaType to a Content-Type HTTP Header
 toContentType :: MediaType -> Header
```

It will now succeed:

```
$ cabal v2-repl

ghci> import PostgREST.MediaType
ghci> decodeMediaType "application/json"
MTApplicationJSON
```

[...]

PR #2671, opened 2023-02-18
coverage: Use OverloadedRecordDot instead of RecordWildCards

wolfgangwalther:
As mentioned in https://github.com/PostgREST/postgrest/issues/2627#issuecomment-1435642297 and elsewhere before, we currently have quite a few false-negatives for coverage: Fields in record types are marked as uncovered, because we use `RecordWildCard`, which isn't supported by HPC.

While `RecordWildCard` is not supported, I just found out that `OverloadedRecordDot` actually is.

This is just a small example of what it could look like.

The key change is the following. Instead of

```haskell
handleRequest :: AuthResult -> [...]
handleRequest AuthResult{..} [...]
[...]
Query.setPgLocals [...] authClaims authRole [...]
```

we do

```haskell
handleRequest :: AuthResult -> [...]
handleRequest authResult [...]
[...]
Query.setPgLocals [...] authResult.claims authResult.role
```

I'm pretty sure the dot syntax will feel more natural to a lot of people.

Imho, it only makes sense when we remove all the prefixes for record fields, as done in the example above. `authResult.authClaims` etc. would just read very clumsy.

At the same time, this also allows us to use the type itself in a namespaced way, too. We previously had to:

```haskell
import PostgREST.Auth (AuthResult (..))
```

just to have the record field selectors available for `RecordWildCards`. But now we don't need to do that anymore - and we can just use `Auth.AuthResult` from the qualified import. Of course, this then allows us to drop the prefix from the type, too, so we can export just `Result` from `Auth.hs` and use it as `Auth.Result`.

If we were to use this style consistently, we'd get code coverage of our type definitions fixed. Before this MR, the `authClaims` field in the following was uncovered:

```haskell
data AuthResult = AuthResult
  { authClaims :: KM.KeyMap JSON.Value
  , authRole   :: Text
  }
```

Now, this line is covered, too.

WDYT?

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE MultiWayIf          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
@@ -303,6 +304,12 @@ getConfig appState = readIORef appState.conf
 putConfig :: AppState -> AppConfig -> IO ()
 putConfig appState = atomicWriteIORef appState.conf
 
+instance HasField "getConnStatus" AppState (IO ConnectionStatus) where
+  getField appState = readIORef appState.connStatus
+
+instance HasField "putConnStatus" AppState (ConnectionStatus -> IO ()) where
+  getField appState = atomicWriteIORef appState.connStatus
+
 getIsListenerOn :: AppState -> IO Bool
 getIsListenerOn appState = do
   AppConfig{..} <- getConfig appState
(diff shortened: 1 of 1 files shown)


steve-chavez:
LGTM!

> Any specific reason why we still support GHC 8.10?

Not one from my side.

cc @robx

robx:
> LGTM!
> 
> > Any specific reason why we still support GHC 8.10?
> 
> Not one from my side.
> 
> cc @robx

None that I'm aware of either.

steve-chavez:
> Any specific reason why we still support GHC 8.10?

Removed on https://github.com/PostgREST/postgrest/pull/2683

steve-chavez:
There's a blocker for [OverloadedRecordDot](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_record_dot.html) on https://github.com/PostgREST/postgrest/pull/2834.

That will be solved once we offer a static binary for ARM.

Issue #2602, opened 2022-12-29
ON CONFLICT specify Action


vivianshub:

steve-chavez:
Or as suggested on https://github.com/PostgREST/postgrest/issues/2066, we could set `on_conflict` on the `Prefer` header instead:

```http
Prefer: resolution=merge-duplicates;columns=col1,col2,col3
```

That seems simpler to implement. To make it clearer it could also be:

```http
Prefer: on-conflict=update;columns=col1,col2,col3
```

steve-chavez:
We could also try to combine this with https://github.com/PostgREST/postgrest/issues/465.

```http
Prefer: resolution=merge-duplicates;col1=set.col1;col2=set.col2;col3=set.col3


Prefer: resolution=merge-duplicates;col1=set.excluded.col1;col2=set.excluded.col2;col3=set.excluded.col3
```

It's longer to type but it would support using operators and other things that `set` does.

PR #2672, opened 2023-02-19
feat: undefined json keys as defaults when using `?columns` and a `Prefer` header

### LINKED ISSUE #1567: Feature Request: Updating Columns to Default Values

steve-chavez:
Closes #1567.

(`field-with_sep` has default 1 on the following)

- Allows bulk insert with defaults

```bash
http POST "localhost:3000/complex_items?columns=id,name,field-with_sep,arr_data" "Prefer: return=representation" <<JSON
[
  {"id": 4, "name": "Vier"},
  {"id": 5, "name": "Funf", "arr_data": null},
  {"id": 6, "name": "Sechs", "field-with_sep": 6, "arr_data": [1,2,3]}
]
JSON

[
    { "arr_data": null, "field-with_sep": 1, "id": 4, "name": "Vier", "settings": null },
    { "arr_data": null, "field-with_sep": 1, "id": 5, "name": "Funf", "settings": null },
    { "arr_data": [ 1, 2, 3 ], "field-with_sep": 6, "id": 6, "name": "Sechs", "settings": null }
]
```

- Allows updating with defaults

```bash
http PATCH "/complex_items?id=eq.3&columns=name,field-with_sep" "Prefer: return=representation" <<JSON
{"name": "Tres"}
JSON


[ { "arr_data": [ 1, 2, 3 ], "field-with_sep": 1, "id": 3, "name": "Tres", "settings": { "foo": { "bar": "baz", "int": 1 } } } ]
```

- It works on views as well assuming they have a default set up using: `ALTER VIEW test.complex_items_view ALTER COLUMN name SET DEFAULT 'Default'`

## Pending

- [x] Optionally apply the feature with a Prefer header(since this new insert query is more expensive)
- [x] CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + When the client sends the request header specified in the config it will be included in the response headers.
  - #2694, Make `db-root-spec` stable. - @steve-chavez
    + This can be used to override the OpenAPI spec with a custom database function
+ - #1567, On bulk inserts with `?columns`, undefined json keys can get columns' DEFAULT values by using the `Prefer: undefined-keys=apply-defaults` header - @steve-chavez
 
 ### Fixed
 
--- a/src/PostgREST/ApiRequest/Preferences.hs
+++ b/src/PostgREST/ApiRequest/Preferences.hs
@@ -9,6 +9,7 @@
 module PostgREST.ApiRequest.Preferences
   ( Preferences(..)
   , PreferCount(..)
+  , PreferDefaults(..)
   , PreferParameters(..)
   , PreferRepresentation(..)
   , PreferResolution(..)
@@ -43,6 +44,7 @@ data Preferences
     , preferParameters     :: Maybe PreferParameters
     , preferCount          :: Maybe PreferCount
     , preferTransaction    :: Maybe PreferTransaction
+    , preferDefaults       :: Maybe PreferDefaults
     }
 
 -- |
@@ -109,6 +111,7 @@ fromHeaders headers =
     , preferParameters = parsePrefs [SingleObject, MultipleObjects]
     , preferCount = parsePrefs [ExactCount, PlannedCount, EstimatedCount]
     , preferTransaction = parsePrefs [Commit, Rollback]
+    , preferDefaults = parsePrefs [ApplyDefaults, IgnoreDefaults]
     }
   where
     prefHeaders = filter ((==) HTTP.hPrefer . fst) headers
@@ -204,3 +207,14 @@ instance ToHeaderValue PreferTransaction where
   toHeaderValue Rollback = "tx=rollback"
 
 instance ToAppliedHeader PreferTransaction
+
+data PreferDefaults
+  = ApplyDefaults
+  | IgnoreDefaults
+  deriving Eq
+
+instance ToHeaderValue PreferDefaults where
+  toHeaderValue ApplyDefaults  = "defaults=apply"
+  toHeaderValue IgnoreDefaults = "defaults=ignore"
+
+instance ToAppliedHeader PreferDefaults
--- a/src/PostgREST/ApiRequest/Preferences.hs
+++ b/src/PostgREST/ApiRequest/Preferences.hs
@@ -94,13 +94,14 @@ data Preferences
 --
 -- Preferences can be separated by arbitrary amounts of space, lower-case header is also recognized:
 --
--- >>> pPrint $ fromHeaders [("prefer", "count=exact,    tx=commit   ,return=representation")]
+-- >>> pPrint $ fromHeaders [("prefer", "count=exact,    tx=commit   ,return=representation , defaults=apply")]
 -- Preferences
 --     { preferResolution = Nothing
 --     , preferRepresentation = Full
 --     , preferParameters = Nothing
 --     , preferCount = Just ExactCount
 --     , preferTransaction = Just Commit
+--     , preferDefaults = Just ApplyDefaults
 --     }
 --
 fromHeaders :: [HTTP.Header] -> Preferences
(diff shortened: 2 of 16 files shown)


steve-chavez:
> It works on views as well assuming they have a default set up using: ALTER VIEW test.complex_items_view ALTER COLUMN name SET DEFAULT 'Default'

The above is not ideal. When omitting a column on a view INSERT:

```sql
INSERT INTO test.complex_items_view(id,name) VALUES (11, 'Eleven') RETURNING *;
 id |  name  | settings | arr_data | field-with_sep
----+--------+----------+----------+----------------
 11 | Eleven |          |          |              1
```

PostgreSQL correctly infers the `field-with_sep` default which is defined on the source table but when leaving an undefined key on the json we don't infer the default. I'm leaving this for a later enhancement.

steve-chavez:
> Check if it makes sense for upsert(might need additional logic)

So I was thinking that doing this:

(consider that `id:3` already exists)

```bash
http POST "localhost:3000/complex_items?columns=id,name,field-with_sep,arr_data" "Prefer: return=representation, resolution=merge-duplicates" <<JSON
[
  {"id": 3, "name": "Drei"},
  {"id": 4, "name": "Vier"},
  {"id": 5, "name": "Funf", "arr_data": null},
  {"id": 6, "name": "Sechs", "field-with_sep": 6, "arr_data": [1,2,3]}
]
JSON
```

Would only UPDATE the `name` column for the `id: 3` column, currently it updates the `field-with_sep` to the default and `arr_data` to `null`.

I was thinking to do this automatically but:

- doing so would not allow to update a column to a default on upsert
- it'd be better to be explicit about the updated columns: https://github.com/PostgREST/postgrest/issues/2602#issuecomment-1382753901

steve-chavez:
So I just made a pgbench test on the old insert query compared to the new one:

### old

```bash
$ postgrest-with-postgresql-15 -f test/pgbench/fixtures.sql pgbench -T 30 -f test/pgbench/1567/old.sql

pgbench (15.1)
starting vacuum...pgbench: error: ERROR:  relation "pgbench_branches" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
pgbench: error: ERROR:  relation "pgbench_tellers" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
pgbench: error: ERROR:  relation "pgbench_history" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
end.
transaction type: test/pgbench/1567/old.sql
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 30 s
number of transactions actually processed: 146236
latency average = 0.205 ms
initial connection time = 1.234 ms
tps = 4874.710773 (without initial connection time)
```

### new

```
$ postgrest-with-postgresql-15 -f test/pgbench/fixtures.sql pgbench -T 30 -f test/pgbench/1567/new.sql

pgbench (15.1)
starting vacuum...pgbench: error: ERROR:  relation "pgbench_branches" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
pgbench: error: ERROR:  relation "pgbench_tellers" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
pgbench: error: ERROR:  relation "pgbench_history" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
end.
transaction type: test/pgbench/1567/old.sql
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 30 s
number of transactions actually processed: 146236
latency average = 0.205 ms
initial connection time = 1.234 ms
tps = 4874.710773 (without initial connection time)
pgbench (15.1)
starting vacuum...pgbench: error: ERROR:  relation "pgbench_branches" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
pgbench: error: ERROR:  relation "pgbench_tellers" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
pgbench: error: ERROR:  relation "pgbench_history" does not exist
pgbench: detail: (ignoring this error and continuing anyway)
end.
transaction type: test/pgbench/1567/new.sql
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 30 s
number of transactions actually processed: 121645
latency average = 0.247 ms
initial connection time = 1.253 ms
tps = 4054.990532 (without initial connection time)
```

## results

So there's `16.8%` loss in perf.

>  Shortcut for tables that don't have defaults(since this new insert query is more expensive)

The above is adequate but I'm also thinking maybe we should add a new `Prefer` header for this feature.

steve-chavez:
> Like Prefer: infer-defaults=true.

The above wouldn't be right if this feature is JSON specific, it would have to be like:

- `Content-Type: application/vnd.pgrst.object+json; apply-defaults=true`
- `Content-Type: application/vnd.pgrst.array+json; apply-defaults=true`

Maybe it could be added to CSV as well though.. in which case the `Prefer` would not be incorrect. It's also less ergonomic to switch the whole content type for an optional behavior.

Also both options could be done.

steve-chavez:
So I've added a `Prefer: defaults=apply/ignore` header. Maybe to be more precise this should be `Prefer: undefined-keys = apply-defaults/ignore-defaults` .

steve-chavez:
Having to add both `columns` and the `Prefer` header also seems like a complicated interface. Maybe it won't matter later when we get rid of the Haskell json processing.

steve-chavez:
So since this feature requires `jsonb`, we no longer would be able to insert json with duplicates keys, if we were to switch completely to `jsonb`. This would be a breaking change.

To avoid this, I'm only switching to `jsonb` whenever the new Prefer header is used. I've also added tests on for this(on https://github.com/PostgREST/postgrest/pull/2672/commits/dc3335798f884f95a1b4e77bf2ff5947d9005b93).

steve-chavez:
> Having to add both columns and the Prefer header also seems like a complicated interface.

To avoid complicating the interface, I've enabled this Prefer header to work without `?columns`. I see no harm in doing so. We can document that this is only required with `?columns`.

steve-chavez:
Tried to appease codecov with the method mentioned on https://github.com/PostgREST/postgrest/pull/2671/, it cleared up one warning but not the other 2, no idea how to solve those.

Will leave it like that, they're only record fields which codecov doesn't understand.

steve-chavez:
Looking good to merge now.

steve-chavez:
> What about something like Prefer: missing-values=apply-defaults|ignore-defaults or Prefer: missing=default|null. I actually like the latter more.

Great idea! It looks more generic.

Doing that on https://github.com/PostgREST/postgrest/pull/2723.

Issue #1652, opened 2020-11-17
limits don't seem to apply directly to the underlying query, causing poor performance


ric2b:
### Environment

* PostgreSQL version: 12.5 (docker image `postgres:12`)
* PostgREST version: 7.0.1 (docker image `postgrest/postgrest:v7.0.1`)
* Operating system: Ubuntu `20.10`

### Description of issue

~~(My best guess is that it is related to https://github.com/PostgREST/postgrest/issues/621, but I might be misdiagnosing the issue)~~

Nope, see the following comments, the use of CTE's doesn't seem to be related.

#### Some context
I have the following simple function to do a text search for products and return the rows ordered by similarity (`<->` is from the [pg_trgm](https://www.postgresql.org/docs/current/pgtrgm.html) extenstion) of the text query to the concatenated and unaccented `brand`, `name` and `description` fields:

```sql
create or replace function search_products(query text) returns setof product as $$
    select *
    from product
    order by query <-> unaccent_concat_brand_name_desc(brand, name, description)::text asc
$$ stable language sql;
```

(By the way, the reason this is a function is that I don't think it's possible to represent it in PostgREST's query syntax, but please correct me if I'm wrong)
 
`unaccent_concat_brand_name_desc` is a function I created because `unaccent` and `concat` aren't immutable (for some very specific reasons I looked into and don't concern me) and were blocking me from creating an index to help with this query. With that function (marked as immutable) I created the following index:

```sql
create index product_trigram_idx on product using gist (unaccent_concat_brand_name_desc(brand, name, description) gist_trgm_ops);
```

With that index I can run the following query in ~100ms:

```sql 
select store_id, id, brand, name, description
from search_products('potatoes')
where store_id='some_store_id'
limit 10;
```

I have run the query above with `explain analyse` and the plan does use the index (plus the performance difference is very noticeable when I remove the index).

#### The issue
I'm trying to get equivalent results from PostgREST, so I call `/rpc/search_products?query=potatoes&store_id=eq.some_store_id&select=store_id,id,brand,name,description&limit=10`.

However this takes nearly 2 seconds to run, this is the query that actually hits PostgreSQL:

```sql
db_1          | 2020-11-17 11:37:14.524 UTC [31] LOG:  execute 0: BEGIN ISOLATION LEVEL READ COMMITTED READ ONLY
[...]

Issue #2676, opened 2023-02-21
Optimizations for INSERT


steve-chavez:
While working on https://github.com/PostgREST/postgrest/pull/2672, I noticed some optimizations we can do for INSERTs.

Our current query on `POST /complex_items Prefer: return=representation` is:

```sql
WITH pgrst_source AS (
  WITH
  pgrst_payload AS (SELECT '[{"id": 4, "name": "Vier"}, {"id": 5, "name": "Funf", "arr_data": null}, {"id": 6, "name": "Sechs", "arr_data": [1, 2, 3], "field-with_sep": 6}]'::json AS json_data),
  pgrst_body AS ( SELECT CASE WHEN json_typeof(json_data) = 'array' THEN json_data ELSE json_build_array(json_data) END AS val FROM pgrst_payload)
  INSERT INTO "test"."complex_items"("arr_data", "field-with_sep", "id", "name")
  SELECT "arr_data", "field-with_sep", "id", "name"
  FROM json_to_recordset ((SELECT val FROM pgrst_body)) AS _ ("arr_data" integer[], "field-with_sep" integer, "id" bigint, "name" text)
  RETURNING "test"."complex_items".*
)
SELECT
  '' AS total_result_set,
  pg_catalog.count(_postgrest_t) AS page_total,
  array[]::text[] AS header,
  coalesce(json_agg(_postgrest_t), '[]')::character varying AS body,
  nullif(current_setting('response.headers', true), '') AS response_headers,
  nullif(current_setting('response.status', true), '') AS response_status
FROM (SELECT "complex_items".* FROM "pgrst_source" AS "complex_items") _postgrest_t;
```

### Optimizations

- [x]  `pg_catalog.count(_postgrest_t) AS page_total` is always included, but if remove it, this grants us an increase in 12% TPS. AFAICT this is only required for:
  + PUT
  + When using `application/vnd.pgrst.object+json`
  + When using `Prefer: count=exact`
  + Cannot be done. See [below](https://github.com/PostgREST/postgrest/issues/2676#issuecomment-1981675643).

- [x] Change the CTEs(`pgrst_payload, pgrst_body`) to LATERAL instead. This increases TPS by 6%. I got the hint of this optimization from https://github.com/PostgREST/postgrest/issues/1652#issuecomment-730358520.
  + Done on https://github.com/PostgREST/postgrest/pull/2677
[...]

PR #2677, opened 2023-02-22
perf: use LATERAL instead of CTE for json body

steve-chavez:
Increases transactions per second for RPC on 16% and for INSERTs on 6%.

It uses the fact that [inlining](https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions) can be done on function calls when:

> none of the actual arguments contain volatile expressions or subselects

So basically it goes from:

```sql
WITH
pgrst_payload AS (SELECT '[{"id": 4, "name": "Vier"}, {"id": 5, "name": "Funf", "arr_data": null}, {"id": 6, "name": "Sechs", "arr_data": [1, 2, 3], "field-with_sep": 6}]'::json AS json_data),
INSERT INTO "test"."complex_items"("arr_data", "field-with_sep", "id", "name")
SELECT "arr_data", "field-with_sep", "id", "name"
FROM json_to_recordset ((SELECT val FROM pgrst_body)) AS _ ("arr_data" integer[], "field-with_sep" integer, "id" bigint, "name" text)
RETURNING "test"."complex_items".*
```

To:

```sql
INSERT INTO "test"."complex_items"("arr_data", "field-with_sep", "id", "name")
SELECT pgrst_body."arr_data", pgrst_body."field-with_sep", pgrst_body."id", pgrst_body."name"
FROM (
  SELECT '[{"id": 4, "name": "Vier"}, {"id": 5, "name": "Funf", "arr_data": null}, {"id": 6, "name": "Sechs", "arr_data": [1, 2, 3], "field-with_sep": 6}]'::json as json_data
) pgrst_payload,
LATERAL (
  SELECT CASE WHEN json_typeof(pgrst_payload.json_data) = 'array' THEN pgrst_payload.json_data ELSE json_build_array(pgrst_payload.json_data) END AS val
) pgrst_uniform_json,
LATERAL (
  SELECT * FROM json_to_recordset (pgrst_uniform_json.val) AS _ ("arr_data" integer[], "field-with_sep" integer, "id" bigint, "name" text)
) pgrst_body
RETURNING "test"."complex_items".*
```

---

Fixes https://github.com/PostgREST/postgrest/issues/1652 and does optimization 2 mentioned on https://github.com/PostgREST/postgrest/issues/2676.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + `PGRST204`: Column is not found
    + `PGRST003`: Timed out when acquiring connection to db
  - #2667, Fix `db-pool-acquisition-timeout` not logging to stderr when the timeout is reached - @steve-chavez
+ - #1652, Fix function call with arguments not inlining - @steve-chavez
 
 ## [10.1.2] - 2023-02-01
 
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -82,13 +82,12 @@ getSelectsJoins rr@(Node ReadPlan{select, relName, relToParent=Just rel, relAggA
 
 mutatePlanToQuery :: MutatePlan -> SQL.Snippet
 mutatePlanToQuery (Insert mainQi iCols body onConflct putConditions returnings _) =
-  "WITH " <> normalizedBody body <> " " <>
   "INSERT INTO " <> SQL.sql (fromQi mainQi) <> SQL.sql (if null iCols then " " else "(" <> cols <> ") ") <>
-  pgFmtSelectFromJson iCols <>
+  fromJsonBodyF body iCols True False <>
   -- Only used for PUT
-  (if null putConditions then mempty else "WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "_") <$> putConditions)) <>
+  (if null putConditions then mempty else "WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "pgrst_body") <$> putConditions)) <>
   SQL.sql (BS.unwords [
-    maybe "" (\(oncDo, oncCols) ->
+    maybe mempty (\(oncDo, oncCols) ->
       if null oncCols then
         mempty
       else
@@ -114,15 +113,14 @@ mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings)
     SQL.sql $ "SELECT " <> emptyBodyReturnedColumns <> " FROM " <> fromQi mainQi <> " WHERE false"
 
   | range == allRange =
-    "WITH " <> normalizedBody body <> " " <>
     "UPDATE " <> mainTbl <> " SET " <> SQL.sql nonRangeCols <> " " <>
-    "FROM (" <> pgFmtSelectFromJson uCols <> ") AS _ " <>
+    fromJsonBodyF body uCols False False <>
     whereLogic <> " " <>
     SQL.sql (returningF mainQi returnings)
 
   | otherwise =
-    "WITH " <> normalizedBody body <> ", " <>
-    "pgrst_update_body AS (" <> pgFmtSelectFromJson uCols <> " LIMIT 1), " <>
+    "WITH " <>
+    "pgrst_update_body AS (" <> fromJsonBodyF body uCols True True <> "), " <>
     "pgrst_affected_rows AS (" <>
       "SELECT " <> SQL.sql rangeIdF <> " FROM " <> mainTbl <>
       whereLogic <> " " <>
@@ -138,7 +136,7 @@ mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings)
     whereLogic = if null logicForest then mempty else " WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree mainQi <$> logicForest)
     mainTbl = SQL.sql (fromQi mainQi)
     emptyBodyReturnedColumns = if null returnings then "NULL" else BS.intercalate ", " (pgFmtColumn (QualifiedIdentifier mempty $ qiName mainQi) <$> returnings)
-    nonRangeCols = BS.intercalate ", " (pgFmtIdent . tfName <> const " = _." <> pgFmtIdent . tfName <$> uCols)
+    nonRangeCols = BS.intercalate ", " (pgFmtIdent . tfName <> const " = " <> pgFmtColumn (QualifiedIdentifier mempty "pgrst_body") . tfName <$> uCols)
     rangeCols = BS.intercalate ", " ((\col -> pgFmtIdent (tfName col) <> " = (SELECT " <> pgFmtIdent (tfName col) <> " FROM pgrst_update_body) ") <$> uCols)
     (whereRangeIdF, rangeIdF) = mutRangeF mainQi (fst . otTerm <$> ordts)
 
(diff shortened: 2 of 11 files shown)


steve-chavez:
One thing to note is that the `loadtest` is not capturing this kind of slight improvements.

A first run shows an improvement in throughput: https://github.com/PostgREST/postgrest/pull/2677/checks?check_run_id=11510644266

While the second shows a decrease: https://github.com/PostgREST/postgrest/pull/2677/checks?check_run_id=11526098776

steve-chavez:
Will leave the pgbench test on the codebase. It doesn't run on CI but I think it's useful to have it handy(for copy/pasting for other benchs too).

steve-chavez:
Took at stab at https://github.com/PostgREST/postgrest/issues/1652. Using the queries on https://github.com/PostgREST/postgrest/pull/2677/commits/cc9cfe2344564e7bfbf2ebba4066ce6c96ed7d4a.

```bash
$ postgrest-with-postgresql-15 -f test/pgbench/fixtures.sql pgbench -n -T 10 -f test/pgbench/1652/old.sql 

pgbench (15.1)
transaction type: test/pgbench/1652/old.sql
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 10 s
number of transactions actually processed: 45046
latency average = 0.222 ms
initial connection time = 1.254 ms
tps = 4505.073033 (without initial connection time)

$ postgrest-with-postgresql-15 -f test/pgbench/fixtures.sql pgbench -n -T 10 -f test/pgbench/1652/new.sql
pgbench (15.1)
transaction type: test/pgbench/1652/new.sql
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 10 s
number of transactions actually processed: 52286
latency average = 0.191 ms
initial connection time = 1.274 ms
tps = 5229.205542 (without initial connection time)
```

So a 16% TPS increase. This is a case where the cost plan is misleading because the old query has a lower cost than the new one:

```sql
explain WITH pgrst_source AS (
  WITH
  pgrst_payload AS (SELECT '{"id": 4}'::json AS json_data),
  pgrst_args AS ( SELECT * FROM json_to_recordset((SELECT val FROM pgrst_body)) AS _("id" integer) )
  SELECT "get_projects_below".*
  FROM "test"."get_projects_below"("id" := (SELECT "id" FROM pgrst_args LIMIT 1))
)
SELECT
FROM (SELECT "projects".* FROM "pgrst_source" AS "projects") _postgrest_t;

                                          QUERY PLAN
----------------------------------------------------------------------------------------------
 Aggregate  (cost=25.28..25.30 rows=1 width=112)
   ->  Function Scan on get_projects_below  (cost=0.28..10.28 rows=1000 width=40)
         InitPlan 2 (returns $1)
           ->  Limit  (cost=0.02..0.03 rows=1 width=4)
                 InitPlan 1 (returns $0)
                   ->  Result  (cost=0.00..0.01 rows=1 width=32)
                 ->  Function Scan on json_to_recordset _  (cost=0.00..1.00 rows=100 width=4)
```

```sql
explain WITH pgrst_source AS (
  SELECT "get_projects_below".*
  FROM (
    SELECT '{"id": 4}'::json as json_data
  ) pgrst_payload,
  LATERAL (
    SELECT CASE WHEN json_typeof(pgrst_payload.json_data) = 'array' THEN pgrst_payload.json_data ELSE json_build_array(pgrst_payload.json_data) END AS val
  ) pgrst_uniform_json,
  LATERAL (
    SELECT * FROM json_to_recordset(pgrst_uniform_json.val) AS _("id" integer) LIMIT 1
  ) pgrst_args,
  LATERAL "test"."get_projects_below"("id" := pgrst_args.id)
)
SELECT
FROM (SELECT "projects".* FROM "pgrst_source" AS "projects") _postgrest_t;

                                         QUERY PLAN
--------------------------------------------------------------------------------------------
 Aggregate  (cost=28.28..28.30 rows=1 width=112)
   ->  Nested Loop  (cost=7.26..26.28 rows=400 width=40)
         ->  Limit  (cost=0.01..0.01 rows=1 width=4)
               ->  Function Scan on json_to_recordset _  (cost=0.01..1.00 rows=100 width=4)
         ->  Bitmap Heap Scan on projects  (cost=7.25..22.25 rows=400 width=40)
               Recheck Cond: (id < _.id)
               ->  Bitmap Index Scan on projects_pkey  (cost=0.00..7.15 rows=400 width=0)
                     Index Cond: (id < _.id)

```

I made the `test.get_projects_below` STABLE for the EXPLAINs above. As it can be seen the new query has inlining enabled since the `Index Cond` applies while the old query does a function scan. So I believe this fixes https://github.com/PostgREST/postgrest/issues/1652.

steve-chavez:
This is failing for pg 11 and 12:

```haskell
      when (actualPgVersion >= pgVersion110) $
        it "can embed if rpc returns domain of table type" $ do
          post "/rpc/getproject_domain?select=id,name,client:clients(id),tasks(id)"
              [json| { "id": 1} |]
            `shouldRespondWith`
              [json|[{"id":1,"name":"Windows 7","client":{"id":1},"tasks":[{"id":1},{"id":2}]}]|]
          get "/rpc/getproject_domain?id=1&select=id,name,client:clients(id),tasks(id)"
            `shouldRespondWith`
              [json|[{"id":1,"name":"Windows 7","client":{"id":1},"tasks":[{"id":1},{"id":2}]}]|]
```

```
  test/spec/Feature/Query/RpcSpec.hs:316:13: 
  1) Feature.Query.RpcSpec, remote procedure call, foreign entities embedding, can embed if rpc returns domain of table type
       status mismatch:
         expected: 200
         but got:  500
       body mismatch:
         expected: [{"client":{"id":1},"id":1,"name":"Windows 7","tasks":[{"id":1},{"id":2}]}]
         but got:  {"code":"XX000","details":null,"hint":null,"message":"could not find attribute 3 in subquery targetlist"}
```

It seems to be a bug in PostgreSQL itself, because of inlining:

```sql
WITH pgrst_source AS (
  SELECT "pgrst_call"."client_id", "pgrst_call"."id", "pgrst_call"."name"
  FROM (SELECT '{ "id": 1}'::json AS json_data) pgrst_payload,
  LATERAL (SELECT CASE WHEN json_typeof(pgrst_payload.json_data) = 'array' THEN pgrst_payload.json_data ELSE json_build_array(pgrst_payload.json_data) END AS val) pgrst_uniform_json,
  LATERAL (SELECT * FROM json_to_recordset(pgrst_uniform_json.val) AS _("id" integer) LIMIT 1) pgrst_body ,
  LATERAL "test"."getproject_domain"("id" := pgrst_body."id") pgrst_call
)
SELECT
FROM (SELECT "projects"."id", "projects"."name", row_to_json("projects_client_1".*) AS "client", COALESCE( "projects_tasks_1"."projects_tasks_1", '[]') AS "tasks" FROM "pgrst_source" AS "proje
cts" LEFT JOIN LATERAL ( SELECT "clients_1"."id" FROM "test"."clients" AS "clients_1"  WHERE "clients_1"."id" = "projects"."client_id"   ) AS "projects_client_1" ON TRUE LEFT JOIN LATERAL ( SE
LECT json_agg("projects_tasks_1") AS "projects_tasks_1"FROM (SELECT "tasks_1"."id" FROM "test"."tasks" AS "tasks_1"  WHERE "tasks_1"."project_id" = "projects"."id"   ) AS "projects_tasks_1" )
AS "projects_tasks_1" ON TRUE   ) _postgrest_t;
ERROR:  type "projects_domain" does not exist
LINE 2:           SELECT projects::projects_domain FROM test.project...
                                   ^
QUERY:
          SELECT projects::projects_domain FROM test.projects WHERE id = $1;

CONTEXT:  SQL function "getproject_domain" during inlining
```

I'll just enable the test to run on pg >= 13.

steve-chavez:
To prevent regressions, I added tests to ensure inlining happens.

steve-chavez:
Right now we're considering `void` as `Nothing:

https://github.com/PostgREST/postgrest/blob/caaa9a394495b936b87d75c1f151ea39ebe86c06/src/PostgREST/SchemaCache.hs#L258-L267

steve-chavez:
> I don't think we can claim to support PG11 and PG12 and not fix it.

Yeah, you're right.

> I think this would actually be technically correct, because void behaves exactly like a scalar always returning NULL.

Let me try and make this fix on https://github.com/PostgREST/postgrest/pull/2723

steve-chavez:
> I think this would actually be technically correct, because void behaves exactly like a scalar always returning NULL.

That did work and actually ProcDescription looks better by having a `RetType` instead of a `Maybe RetType`.

3 test failures now:

```
  test/spec/Feature/Query/RpcSpec.hs:627:11:
  1) Feature.Query.RpcSpec, remote procedure call, procs with OUT/INOUT params, returns an object result when there is a single OUT param
       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: {"num_plus_one":6}
         but got:  {"code":"42809","details":null,"hint":null,"message":"type integer is not composite"}

  To rerun use: --match "/Feature.Query.RpcSpec/remote procedure call/procs with OUT/INOUT params/returns an object result when there is a single OUT param/"

  test/spec/Feature/Query/RpcSpec.hs:636:11:
  2) Feature.Query.RpcSpec, remote procedure call, procs with OUT/INOUT params, returns an object result when there is a single INOUT param
       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: {"num":3}
         but got:  {"code":"42809","details":null,"hint":null,"message":"type integer is not composite"}

  To rerun use: --match "/Feature.Query.RpcSpec/remote procedure call/procs with OUT/INOUT params/returns an object result when there is a single INOUT param/"

  test/spec/Feature/Query/RpcSpec.hs:652:11:
  3) Feature.Query.RpcSpec, remote procedure call, procs with TABLE return, returns an object result when there is a single-column TABLE return type
       status mismatch:
         expected: 200
         but got:  400
       body mismatch:
         expected: [{"a":"A"}]
         but got:  {"code":"42809","details":null,"hint":null,"message":"type text is not composite"}
```

Somehow these are being taken as scalars:

```sql
create function test.single_out_param(num int, OUT num_plus_one int) AS $$
  select num + 1;
$$ language sql;

create function test.single_inout_param(INOUT num int) AS $$
  select num + 1;
$$ language sql;

create function test.single_column_table_return () returns table (a text) AS $$
  select 'A'::text;
$$ language sql;
```

There's a blocker in fixing the `returns table (a text)` one. 

```haskell
procReturnsScalar :: ProcDescription -> Bool
procReturnsScalar proc = case proc of
  ProcDescription{pdReturnType = Single (Scalar _)} -> True
  ProcDescription{pdReturnType = SetOf (Scalar _)}  -> True
  _                                                    -> False
```

The binary output on a single column logic is relying on the `SetOf (Scalar _) -> True` :confused:

steve-chavez:
Pretty sure there is an assumption somewhere for why the first 2 functions are considered scalar too.

steve-chavez:
> The binary output on a single column logic is relying on the SetOf (Scalar _) -> True confused

This would be easy to clear if I can break the current binary output feature(already did that on https://github.com/PostgREST/postgrest/pull/2701) instead of relying on workarounds which are hard to understand in the codebase.

I think I'll continue with https://github.com/PostgREST/postgrest/pull/1582 so I can break binary output and then fix this one.

steve-chavez:
> If I understood correctly, then RPCs returning a composite value are entirely broken on PG11 and PG12, right?

@wolfgangwalther Not really, the following test is passing for example:

https://github.com/PostgREST/postgrest/blob/032f07f3f05eca99b3174ea6517ce0c0f54ff4fd/test/spec/Feature/Query/RpcSpec.hs#L79-L84

Returning a domain on RPC is also passing:

https://github.com/PostgREST/postgrest/blob/032f07f3f05eca99b3174ea6517ce0c0f54ff4fd/test/spec/Feature/Query/RpcSpec.hs#L384-L389

The only thing that's broken is returning a domain alias for a table type:

https://github.com/PostgREST/postgrest/blob/032f07f3f05eca99b3174ea6517ce0c0f54ff4fd/test/spec/fixtures/schema.sql#L1067-L1079

This seems to be fixed on pg here https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=a9ae99d0190960ce2d3dd3e5f10e7f4adc3cf203 but I don't think it has been backpatched(just tried 11.19, which should also come on https://github.com/PostgREST/postgrest/pull/2718, it doesn't work)

Other references:

- https://trac.osgeo.org/postgis/ticket/4511
- https://www.postgresql.org/message-id/flat/402356c32eeb93d4fed01f66d6c7fe2d%40postgrespro.ru

> I don't think we can claim to support PG11 and PG12 and not fix it.

So that looks a bit extreme. At worst, I think the limitation on returning DOMAIN alias for table on RPC could be documented. And for those cases we can suggest to upgrade pg version.

> +   LATERAL (SELECT ("test"."getproject_domain"("id" := pgrst_body."id")).*) pgrst_call;

Doing the above also inflates the plan cost tremendously.

steve-chavez:
> The only thing that's broken is returning a domain alias for a table type

Actually that gave me a clean idea on how to solve this, followed up on https://github.com/PostgREST/postgrest/pull/2737

PR #2691, opened 2023-02-28
perf (?): make pre-request preparable

robx:
The previous use of [`Hasql.Transaction.sql`](https://hackage.haskell.org/package/hasql-transaction-1.0.1.2/docs/Hasql-Transaction.html#v:sql) means that the pre-request statement can't be prepared, regardless of configuration setting.

This change makes it so that it can be prepared.

I'm a bit unsure whether this is actually an improvement -- the pre-request statement is just a parameter-less `SELECT pre_req_func()`, so we'd only be saving the parsing of that server-side.

Load test results seem to indicate a slight improvement, but they're not really conclusive: Locally, I see a request rate of ~380 with plain `postgrest-loadtest`, ~370 with `PGRST_DB_PRE_REQUEST=call_me_prereq postgrest-loadtest` on the `main` branch, and ~375 with this change. (`call_me_prereq` is from this commit: https://github.com/robx/postgrest/commit/5bcb0aef39b9437380b96ad0f3d68015628f772f.)

### DIFF
--- a/src/PostgREST/Query.hs
+++ b/src/PostgREST/Query.hs
@@ -23,6 +23,8 @@ import qualified Data.Text.Encoding                as T
 import qualified Hasql.Decoders                    as HD
 import qualified Hasql.DynamicStatements.Snippet   as SQL (Snippet)
 import qualified Hasql.DynamicStatements.Statement as SQL
+import qualified Hasql.Encoders                    as HE
+import qualified Hasql.Statement                   as SQL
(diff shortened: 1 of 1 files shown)


steve-chavez (approved):
Great idea! LGTM :+1:

steve-chavez:
@robx Do you have any concerns on this one? It's looking good to merge.

PR #2693, opened 2023-03-01
Support bulk updates using PATCH

### LINKED ISSUE #1959: Support proper bulk update via PATCH

laurenceisla:
Closes #1959. It allows bulk updates using PATCH only when the header `Prefer: params=multiple-objects` is applied as suggested here https://github.com/PostgREST/postgrest/issues/1959#issuecomment-1428789816.

- [x] Allow bulk updates
- [x] Allow different keys in the body 
  Needs: https://github.com/PostgREST/postgrest/pull/2672
- [ ] Clean and refactor
- [x] Add tests
- [x] Add changelog

### DIFF
--- a/src/PostgREST/ApiRequest/Preferences.hs
+++ b/src/PostgREST/ApiRequest/Preferences.hs
@@ -115,7 +115,7 @@ fromHeaders headers =
     , preferParameters = parsePrefs [SingleObject, MultipleObjects]
     , preferCount = parsePrefs [ExactCount, PlannedCount, EstimatedCount]
     , preferTransaction = parsePrefs [Commit, Rollback]
-    , preferUndefinedKeys = parsePrefs [ApplyDefaults, IgnoreDefaults]
+    , preferUndefinedKeys = parsePrefs [ApplyDefaults, IgnoreValues, ApplyNulls]
     }
   where
     prefHeaders = filter ((==) HTTP.hPrefer . fst) headers
@@ -216,12 +216,14 @@ instance ToAppliedHeader PreferTransaction
 -- How to handle the insertion/update when the keys specified in ?columns are not present
 -- in the json body.
 data PreferUndefinedKeys
-  = ApplyDefaults  -- ^ Use the default column value for the unspecified keys.
-  | IgnoreDefaults -- ^ Inserts: null values / Updates: the keys are not SET to any value
+  = ApplyDefaults  -- ^ Inserts and Updates: Use the default column value for the unspecified keys.
+  | IgnoreValues   -- ^ Inserts: null values | Updates: the keys are not SET to any value
+  | ApplyNulls     -- ^ Inserts: null values | Updates: SET the values to null
   deriving Eq
 
 instance ToHeaderValue PreferUndefinedKeys where
-  toHeaderValue ApplyDefaults  = "undefined-keys=apply-defaults"
-  toHeaderValue IgnoreDefaults = "undefined-keys=ignore-defaults"
+  toHeaderValue ApplyDefaults = "undefined-keys=apply-defaults"
+  toHeaderValue IgnoreValues  = "undefined-keys=ignore-values"
+  toHeaderValue ApplyNulls    = "undefined-keys=apply-nulls"
 
 instance ToAppliedHeader PreferUndefinedKeys
--- a/src/PostgREST/ApiRequest/Preferences.hs
+++ b/src/PostgREST/ApiRequest/Preferences.hs
@@ -173,10 +173,10 @@ instance ToHeaderValue PreferRepresentation where
   toHeaderValue None        = "return=minimal"
   toHeaderValue HeadersOnly = "return=headers-only"
 
--- | How to pass parameters to stored procedures.
+-- | How to pass parameters to stored procedures or handle bulk UPDATEs.
 data PreferParameters
   = SingleObject    -- ^ Pass all parameters as a single json object to a stored procedure.
-  | MultipleObjects -- ^ Pass an array of json objects as params to a stored procedure.
+  | MultipleObjects -- ^ Pass an array of json objects as params to a stored procedure or as items in a bulk UPDATE.
   deriving Eq
 
 -- TODO: Deprecate params=multiple-objects in next major version
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -83,7 +83,7 @@ getSelectsJoins rr@(Node ReadPlan{select, relName, relToParent=Just rel, relAggA
 mutatePlanToQuery :: MutatePlan -> SQL.Snippet
 mutatePlanToQuery (Insert mainQi iCols body onConflct putConditions returnings _ applyDefaults) =
   "INSERT INTO " <> SQL.sql (fromQi mainQi) <> SQL.sql (if null iCols then " " else "(" <> cols <> ") ") <>
-  fromJsonBodyF body iCols True False applyDefaults <>
+  fromJsonBodyF body iCols True False applyDefaults False <>
   -- Only used for PUT
   (if null putConditions then mempty else "WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "pgrst_body") <$> putConditions)) <>
   SQL.sql (BS.unwords [
@@ -105,7 +105,7 @@ mutatePlanToQuery (Insert mainQi iCols body onConflct putConditions returnings _
     cols = BS.intercalate ", " $ pgFmtIdent . tfName <$> iCols
 
 -- An update without a limit is always filtered with a WHERE
-mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings applyDefaults pkFlts isBulk)
+mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings applyDefaults applyIgnoreValues pkFlts isBulk)
   | null uCols =
     -- if there are no columns we cannot do UPDATE table SET {empty}, it'd be invalid syntax
     -- selecting an empty resultset from mainQi gives us the column names to prevent errors when using &select=
@@ -113,20 +113,20 @@ mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings a
     SQL.sql $ "SELECT " <> emptyBodyReturnedColumns <> " FROM " <> fromQi mainQi <> " WHERE false"
 
   | isBulk =
-    "UPDATE " <> mainTbl <> " SET " <> SQL.sql bulkCols <> " " <>
-    fromJsonBodyF body uCols False False applyDefaults <>
+    "UPDATE " <> mainTbl <> " SET " <> SQL.sql (if applyIgnoreValues then bulkCols else nonRangeCols) <> " " <>
+    fromJsonBodyF body uCols False False applyDefaults applyIgnoreValues <>
     whereLogicBulk <> " " <>
     SQL.sql (returningF mainQi returnings)
 
   | range == allRange =
     "UPDATE " <> mainTbl <> " SET " <> SQL.sql nonRangeCols <> " " <>
-    fromJsonBodyF body uCols False False applyDefaults <>
+    fromJsonBodyF body uCols False False applyDefaults applyIgnoreValues <>
     whereLogic <> " " <>
     SQL.sql (returningF mainQi returnings)
 
   | otherwise =
     "WITH " <>
-    "pgrst_update_body AS (" <> fromJsonBodyF body uCols True True applyDefaults <> "), " <>
+    "pgrst_update_body AS (" <> fromJsonBodyF body uCols True True applyDefaults applyIgnoreValues <> "), " <>
     "pgrst_affected_rows AS (" <>
       "SELECT " <> SQL.sql rangeIdF <> " FROM " <> mainTbl <>
       whereLogic <> " " <>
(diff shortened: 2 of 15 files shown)


test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -535,11 +541,136 @@ spec = do
...
+--        [json|[
+--          { "id": 4, "name": "item-4", "observation": "Damaged item" }
+--        , { "id": 3, "name": "item-3 - 3rd", "observation": null }
+--        ]|]
+--      `shouldMutateInto`
+--      [json|[
+--        { "id": 2, "name": "item-2", "observation": null }
+--      , { "id": 3, "name": "item-3", "observation": null }
+--      , { "id": 4, "name": "item-4", "observation": "Lost item" }
+--      ]|]
```
laurenceisla, line 629: I'm not sure if both bulk update and limits should be applied at the same time. Can't find a use case that may use this either. These tests fail btw, they're just an idea of how the interaction may work.

steve-chavez: @laurenceisla Yeah, limits and bulk update doesn't make sense IMO. With bulk update the user would have to limit the amount of array elements sent.

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -535,11 +541,137 @@ spec = do
...
+          }
+
+    it "does not update when no pk is specified in the body" $ do
+      request methodPatch "/bulk_update_items"
+          [("Prefer", "tx=commit"), ("Prefer", "params=multiple-objects")]
+          [json|[
+            { "name": "Item 1" }
+          , { "name": "Item 2" }
+          ]|]
+        `shouldRespondWith` 400
```
laurenceisla, line 668: Here, the error is returned by the database and has this form:

```json
{
    "code": "42703",
    "details": null,
    "hint": null,
    "message": "column pgrst_body.id does not exist"
}
```

It looks a bit confusing, maybe an `SchemaCache` error could be created to specify that this is an issue with bulk updates and no PK is found in the body. @steve-chavez WDYT?

steve-chavez: @laurenceisla That would be much better :+1: 

Why does it assume an `id` column on the error? Is it hardcoded?

laurenceisla: > Is it hardcoded?

No, in this case, the PK for the table is `id`. The next test example has a composite PK of `id,name` and only `id` is specified in the body. In that case, the message would be:

```json
{
    "code": "42703",
    "details": null,
    "hint": null,
    "message": "column pgrst_body.name does not exist"
}
```

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -133,7 +133,9 @@ mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings)
     SQL.sql (returningF mainQi returnings)
 
   where
-    whereLogic = if null logicForest then mempty else " WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree mainQi <$> logicForest)
+    whereLogic = if null logicForestF && (not isBulk || null pkFlts) then mempty else " WHERE " <> intercalateSnippet " AND " (logicForestF <> pgrstUpdateBodyF)
+    logicForestF = pgFmtLogicTree mainQi <$> logicForest
+    pgrstUpdateBodyF = if not isBulk then mempty else pgFmtBodyFilter mainQi (QualifiedIdentifier mempty "pgrst_body") <$> pkFlts
```
steve-chavez, line 138:
```suggestion
    whereLogic = 
      if null logicForestF && (not isBulk || null pkFlts) 
        then mempty 
        else " WHERE " <> intercalateSnippet " AND " (logicForestF <> pgrstUpdateBodyF)
    logicForestF = pgFmtLogicTree mainQi <$> logicForest
    pgrstUpdateBodyF = 
      if not isBulk 
         then mempty 
         else pgFmtBodyFilter mainQi (QualifiedIdentifier mempty "pgrst_body") <$> pkFlts
```

steve-chavez: Also, would it be possible to refactor this to only execute the `not isBulk` conditional once? That would be clearer I think.

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -213,6 +214,9 @@ pgFmtField table (c, []) = SQL.sql (pgFmtColumn table c)
 -- "operator does not exist: json = unknown"
 pgFmtField table (c, jp) = SQL.sql ("to_jsonb(" <> pgFmtColumn table c <> ")") <> pgFmtJsonPath jp
 
+pgFmtBodyFilter :: QualifiedIdentifier -> QualifiedIdentifier -> FieldName -> SQL.Snippet
+pgFmtBodyFilter table cte f = SQL.sql (pgFmtColumn table f <> " = " <> pgFmtColumn cte f)
```
steve-chavez, line 218: Since this function is only used on `mutatePlanToQuery`, how about scoping it to just that function?

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -535,11 +541,137 @@ spec = do
...
+          { "name": "item", "observation": "Lost item" }
+        , { "name": "item-2 - 2nd", "observation": "Damaged item" }
+        , { "name": "item-3 - 3rd", "observation": null }
+        ]|]
+      `shouldMutateInto`
+      [json|[
+        { "id": 1, "name": "item-1", "observation": null }
+      , { "id": 2, "name": "item", "observation": "Lost item" }
+      , { "id": 3, "name": "item", "observation": "Lost item" }
+      ]|]
```
steve-chavez, line 615: A bit confused by this one. So any value on  `limit` and `offset` will have the same outcome?

test/spec/Feature/Query/UpdateSpec.hs
```diff
@@ -535,11 +541,137 @@ spec = do
...
+          [json|[
+            { "id": 1, "name": "Item 1" }
+          , { "id": 2, "name": "Item 2" }
+          , { "id": 3, "name": "Item 3", "observation": "New item" }
+          ]|]
+        `shouldRespondWith`
+          [json| {"message":"All object keys must match","code":"PGRST102","hint":null,"details":null} |]
+          { matchStatus  = 400
+          , matchHeaders = [matchContentTypeJson]
+          }
```
steve-chavez, line 659: Hm, this is definitely undesirable. Users will want to update different columns for each array element. I think we'll need https://github.com/PostgREST/postgrest/pull/2672 for solving this.

steve-chavez: Maybe it can be solved without #2672. So assume you specify `columns` here or another test. You could do:
```sql
UPDATE bulk_update_items SET observation = COALESCE(pgrst_body.observation, observation) WHERE ..
```
 Meaning that a value will be updated to the new value in the body or to its old value.

This will allow updating rows with a changing number of columns. WDYT?

test/spec/Feature/Query/UpdateSpec.hs
steve-chavez: Ah, got it. Q: What does happen when specifying `?columns=id` and the body doesn't have an `id`?

laurenceisla: It does not give an error unlike normal `PATCH`. That's because the `WHERE` with the primary key is forced in the bulk update:

```sql
WHERE "test"."bulk_update_items"."id" = "pgrst_body"."id"
```

In this case, the `pgrs_body.id` is `null`, so the query works but it won't update any row.

src/PostgREST/Query/SqlFragment.hs
laurenceisla: Ah, I was planning to reuse it for bulk DELETEs. But I could leave it inside UPDATE for now and, if needed, modify it in the other PR.

test/spec/Feature/Query/UpdateSpec.hs
laurenceisla: In other words, it works as a normal update (without bulk): it takes the first element in the body and updates applying limits/offsets. Now that you mention it, it could be confusing if the user expects a bulk update of some kind, maybe a custom error should be expected here, like the one where limit/offset are not allowed for `PUT`?

test/spec/Feature/Query/UpdateSpec.hs
laurenceisla: After implementing this I noticed that setting a value to null will fail, e.g.:

```http
/bulk_update_items?columns=id,name,observation
```
```json
[
            { "id": 1, "name": "Item 1" }
          , { "id": 3, "name": "Item 3", "observation": null }
]
```

The item with `id=3` will keep its old `observation` and won't be set to `null`. Is there a conditional that could be used here, to only apply `COALESCE` in these cases? Although, I see it difficult due to the fact that `columns` does not check the body, right?

steve-chavez: Right. Wolfgang suggested a way to do this with the `?` operator on `jsonb` [here](https://github.com/PostgREST/postgrest/issues/1567#issuecomment-739478169).

On https://github.com/PostgREST/postgrest/pull/2672 I had the same problem as well, but I decided to go for the alternative of using the `||` operator on `jsonb`. 

Maybe those approaches could help.

test/spec/Feature/Query/UpdateSpec.hs
steve-chavez: Just to be clear, `limit`/`offset` are ignored right? So `limit=100&offset=400` would result on the same?

> Now that you mention it, it could be confusing if the user expects a bulk update of some kind, maybe a custom error should be expected here, like the one where limit/offset are not allowed for PUT?

Yeah, I agree. I think we can ban `limit/offset` for bulk updates.

test/spec/Feature/Query/UpdateSpec.hs
steve-chavez: I think that's fine. 

> In this case, the pgrs_body.id is null, so the query works but it won't update any row.

I guess one way to prevent this would be to count the array elements and compare it to the updated items. If they're equal then we can assume a unique key was passed for every row, if not, we could throw an error. 

Not sure if worth it though, it could be confusing as well because we can't tell the user which array element would have the incorrect unique.

test/spec/Feature/Query/UpdateSpec.hs
laurenceisla: > Just to be clear, limit/offset are ignored right? So `limit=100&offset=400` would result on the same?

No, it does not ignore limits and the result would be different depending on the values. In other words. It works as if it was this test https://github.com/PostgREST/postgrest/blob/b05ea14122298d2b00c4f00b28efa0c36091966f/test/spec/Feature/Query/UpdateSpec.hs#L511-L521

laurence-isla: I implemented the Error handling for this case and I noticed that the `PATCH` and  `DELETE` with limits and without `order` in the query string throw this error:

https://github.com/PostgREST/postgrest/blob/b05ea14122298d2b00c4f00b28efa0c36091966f/src/PostgREST/ApiRequest.hs#L253

The thing is that using the `Range` header with `order` in the query string, work as if we were using `limit` and  `offset`. Also, using `Range` without `order` throws a different error that reaches the database, with this message:

```
"message": "syntax error at or near \"RETURNING\""
```

@steve-chavez Is this behavior expected? This is unrelated to this PR, btw.

test/spec/Feature/Query/UpdateSpec.hs
laurence-isla: > Not sure if worth it though, it could be confusing as well because we can't tell the user which array element would have the incorrect unique.

Yeah, I'm not sure about this one either, maybe it's better to discuss later for another PR.

> It looks a bit confusing, maybe an `SchemaCache` error could be created to specify that this is an issue with bulk updates and no PK is found in the body.

I mentioned this but, as you said on another chat, using it with `?columns` won't allow to verify if the JSON body has the `id`. So I'll go with the names you suggested maybe `json_body` or `bulk_update_body` instead of `pgrst_body`.

test/spec/Feature/Query/UpdateSpec.hs
steve-chavez: > Also, using Range without order throws a different error that reaches the database, with this message:
> "message": "syntax error at or near \"RETURNING\""

Yeah that seems like a bug. Fixing it in another PR would be great.

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -216,12 +216,14 @@ instance ToAppliedHeader PreferTransaction
 -- How to handle the insertion/update when the keys specified in ?columns are not present
 -- in the json body.
 data PreferUndefinedKeys
-  = ApplyDefaults  -- ^ Use the default column value for the unspecified keys.
-  | IgnoreDefaults -- ^ Inserts: null values / Updates: the keys are not SET to any value
+  = ApplyDefaults  -- ^ Inserts and Updates: Use the default column value for the unspecified keys.
+  | IgnoreValues   -- ^ Inserts: null values | Updates: the keys are not SET to any value
+  | ApplyNulls     -- ^ Inserts: null values | Updates: SET the values to null
```
laurence-isla, line 221: I noticed that the comment I added in the other PR is wrong. 

```haskell
| IgnoreDefaults -- ^ Inserts: null values / Updates: the keys are not SET to any value
```

When no header is sent, it updates applying `null` values instead of "ignoring" the key and keeping the old value. 

I added other header as a proposal:
- `IgnoreDefaults` is now `ApplyNulls` and is the default behavior when no header is sent, that is, it updates the unspecified columns to `null` values.
- `IgnoreValues` does not set to any value (in reality it sets to the original value but the result is the same).

test/spec/Feature/Query/UpdateSpec.hs
laurence-isla: I ended up using `?` instead. Although it works now, I know that it isn't too performant. But when i used `||` , I got this query:

```sql
UPDATE "test"."bulk_update_items"
SET "id" = "pgrst_body"."id",
    "name" = "pgrst_body"."name",
    "observation" = "pgrst_body"."observation"
FROM
    (SELECT $1::jsonb AS json_data) pgrst_payload
    CROSS JOIN LATERAL
        (SELECT CASE WHEN jsonb_typeof(pgrst_payload.json_data) = 'array' THEN pgrst_payload.json_data ELSE jsonb_build_array(pgrst_payload.json_data) END AS val) pgrst_uniform_json
    CROSS JOIN LATERAL
        (SELECT jsonb_agg(jsonb_build_object('id', "test"."bulk_update_items"."id",'name', "test"."bulk_update_items"."name",'observation', "test"."bulk_update_items"."observation") || elem) AS val
         FROM jsonb_array_elements(pgrst_uniform_json.val) elem) pgrst_json_old_vals
    CROSS JOIN LATERAL
        (SELECT *
         FROM jsonb_to_recordset(pgrst_json_old_vals.val) AS _("id" integer, "name" text, "observation" text)) pgrst_body
WHERE "test"."bulk_update_items"."id" = "pgrst_body"."id" RETURNING 1;
```

Which gives an error when accessing `test.bulk_update_items` inside the `pgrst_json_old_vals` subquery due to the `CROSS JOIN`. Couldn't find a way to make it work that way, maybe I'm missing something.

steve-chavez:
Will close the PR because it got too stale. Also the design of the feature changed on https://github.com/PostgREST/postgrest/issues/1959#issuecomment-1722383580

Issue #2295, opened 2022-05-28
pipeline mode


steve-chavez:
Currently we send 2 queries for each request, one for the [http context](https://postgrest.org/en/stable/api.html#http-context)(+ role auth + search_path) and another one for the crud operation.

Removing the http context query grants about 33% increase in TPS with plain `pgbench` tests.

The 2 queries cannot be merged into one because the crud query needs the `role` + `search_path` setting beforehand.

I believe [libpq pipeline mode](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html) could help us gain perf here. With pipeline mode we wouldn't need to wait for the result of the first query(which we don't care) before sending the second.

steve-chavez:
Here's how psycopg implemented this: https://github.com/psycopg/psycopg/pull/89. 

```python
with conn.pipeline() as pipeline:
   cur.execute(...)
   cur2.execute(...)
   pipeline.sync()  # calls PQpipelineSync
   r1 = cur.fetchone()
   r2 = cur2.fetchall()
```

We might have to pitch this feature on hasql.

wolfgangwalther:
Supporting pipeline mode would be **great**!

steve-chavez:
Maybe this could also help us with doing an EXPLAIN before the request is executed. Related to https://github.com/PostgREST/postgrest/issues/915#issuecomment-676487851

steve-chavez:
Out of curiosity I pgbenched pipeline mode on https://github.com/PostgREST/postgrest/pull/2672/commits/fa50be2f1a064bba7039d01de913d79f30154c49. I got:

```
postgrest-with-postgresql-15 -f test/pgbench/fixtures.sql pgbench -M extended -n -T 10 -f test/pgbench/pipeline_mode/old.sql
pgbench (15.1)
transaction type: test/pgbench/pipeline_mode/old.sql
scaling factor: 1
query mode: extended
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 10 s
number of transactions actually processed: 30443
number of failed transactions: 0 (0.000%)
latency average = 0.328 ms
initial connection time = 1.232 ms
tps = 3044.579797 (without initial connection time)

postgrest-with-postgresql-15 -f test/pgbench/fixtures.sql pgbench -M extended -n -T 10 -f test/pgbench/pipeline_mode/new.sql
pgbench (15.1)
transaction type: test/pgbench/pipeline_mode/new.sql
scaling factor: 1
[...]

PR #2707, opened 2023-03-16
experiment: use pipeline mode

robx:
This pulls in an exploratory implementation of pipeline mode (#2295). The bulk of the change is in the supporting libraries:

- postgresql-libpq is extended to wrap the libpq pipeline mode API
- hasql is hacked to allow queueing up pipelined statements, with pipeline synchronization and result reading (ignoring) deferred until the next result is required

Some notes as to the state of this:
- the hasql change is very much just a minimal hack to allow us to evaluate pipeline mode in postgrest -- it makes little sense as some kind of "pipeline mode for a hasql" feature
- the implementation seems mostly correct, but error handling is at least a bit broken (we don't treat aborted pipelines quite right), and there's a decent chance that some failure scenarios actually mess up the connection state, though I haven't seen that
- this includes #2682 to allow simulating slow postgres

### DIFF
--- a/src/PostgREST/Query.hs
+++ b/src/PostgREST/Query.hs
@@ -25,7 +25,7 @@ import qualified Hasql.Decoders                    as HD
 import qualified Hasql.DynamicStatements.Snippet   as SQL (Snippet)
 import qualified Hasql.DynamicStatements.Statement as SQL
 import qualified Hasql.Encoders                    as HE
-import qualified Hasql.Session                     as Session
+import qualified Hasql.Session                     as SQL (queuePipelineStatement)
 import qualified Hasql.Statement                   as SQL
 import qualified Hasql.Transaction                 as SQL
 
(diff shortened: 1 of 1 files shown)


robx:
The library changes:
- https://github.com/nikita-volkov/hasql/compare/master...robx:hasql:pipeline2
- https://github.com/nikita-volkov/hasql-transaction/compare/master...robx:hasql-transaction:pipeline
- https://github.com/PostgREST/postgresql-libpq/compare/master...robx:postgresql-libpq:pipeline

The `postgresql-libpq` change is essentially good to go upstream, but I haven't filed it yet.

src/PostgREST/Query.hs
```diff
@@ -233,11 +236,19 @@ optionalRollback AppConfig{..} ApiRequest{iPreferences=Preferences{..}} = do
     shouldRollback =
       configDbTxAllowOverride && preferTransaction == Just Rollback
 
+usePipeline :: Bool
+usePipeline = True
+
+queuePipelineStatement :: params -> SQL.Statement params () -> SQL.Transaction ()
+queuePipelineStatement params stmt =
+  if usePipeline then SQL.inTransaction $ Session.queuePipelineStatement params stmt
+                 else SQL.statement params stmt
```
steve-chavez, line 245: So awesome that is such a small change.

> the implementation seems mostly correct, but error handling is at least a bit broken (we don't treat aborted pipelines quite right), and there's a decent chance that some failure scenarios actually mess up the connection state, though I haven't seen that

So since there could be some unknown unknowns(pipeline mode also new to me), I think we should make this configurable. Name could be `db-pipeline-mode`.

robx: I mean, the large change is in `hasql`. And I don't think it's a particularly sensible implementation there API-wise so far.

But yes gating this by a configuration option makes sense.

PR #2710, opened 2023-03-21
Ignore Range header when the method is different than GET

### LINKED ISSUE #2705: Fix bug with limited `PATCH` or `DELETE` with `Range` header
The `PATCH` and  `DELETE` with limits and without `order` in the query string throw this error:

https://github.com/PostgREST/postgrest/blob/b05ea14122298d2b00c4f00b28efa0c36091966f/src/PostgREST/ApiRequest.hs#L253

The thing is that using the `Range` header with `order` in the query string, works as if we were using `limit` and  `offset`. Also, using `Range` without `order` throws a different error that reaches the database, with this message:

```
"message": "syntax error at or near \"RETURNING\""
```

_Originally posted by @laurence-isla in https://github.com/PostgREST/postgrest/pull/2693#discussion_r1136417969_

laurence-isla:
Closes #2705

- [x] Ignore Range header when method is different than GET
- [x] Add/modify tests
- [x] Refactor/clean code
- [x] Add changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,11 +33,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + `PGRST003`: Timed out when acquiring connection to db
  - #2667, Fix `db-pool-acquisition-timeout` not logging to stderr when the timeout is reached - @steve-chavez
  - #1652, Fix function call with arguments not inlining - @steve-chavez
- - #2705, Fix `Range` header not being ignored on request methods different from `GET` - @laurenceisla
-
+ - #2705, Fix bug when using the `Range` header on `PATCH/DELETE` - @laurenceisla
+   + Fix the`"message": "syntax error at or near \"RETURNING\""` error
+   + Fix doing a limited update/delete when an `order` query parameter was present
+    
 ### Changed
 
- - #2705, The `Range` header is now ignored when doing an RPC request with a method different from `GET` - @laurenceisla
+ - #2705, The `Range` header is now only considered on `GET` requests and is ignored for any other method - @laurenceisla
+   + Other methods should use the `limit/offset` query parameters for sub-ranges
+   + `PUT` requests no longer return an error when this header is present (using `limit/offset` still triggers the error)
 
 ## [10.1.2] - 2023-02-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + `PGRST003`: Timed out when acquiring connection to db
  - #2667, Fix `db-pool-acquisition-timeout` not logging to stderr when the timeout is reached - @steve-chavez
  - #1652, Fix function call with arguments not inlining - @steve-chavez
+ - #2705, Fix `Range` header not being ignored on request methods different from `GET` - @laurenceisla
+
+### Changed
+
+ - #2705, The `Range` header is now ignored when doing an RPC request with a method different from `GET` - @laurenceisla
 
 ## [10.1.2] - 2023-02-01
 
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -240,6 +240,8 @@ getRanges method QueryParams{qsOrder,qsRanges} hdrs
   | method == "PUT" && topLevelRange /= allRange = Left PutLimitNotAllowedError
   | otherwise = Right (topLevelRange, ranges)
   where
+    -- According to the RFC (https://www.rfc-editor.org/rfc/rfc9110.html#name-range),
+    -- the Range header must be ignored for all methods other than GET
     headerRange = if method == "GET" then rangeRequested hdrs else allRange
     limitRange = fromMaybe allRange (HM.lookup "limit" qsRanges)
     headerAndLimitRange = rangeIntersection headerRange limitRange
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -237,10 +237,10 @@ getRanges :: ByteString -> QueryParams -> RequestHeaders -> Either ApiRequestErr
 getRanges method QueryParams{qsOrder,qsRanges} hdrs
   | isInvalidRange = Left $ InvalidRange (if rangeIsEmpty headerRange then LowerGTUpper else NegativeLimit)
   | method `elem` ["PATCH", "DELETE"] && not (null qsRanges) && null qsOrder = Left LimitNoOrderError
-  | method == "PUT" && topLevelRange /= allRange = Left PutRangeNotAllowedError
+  | method == "PUT" && topLevelRange /= allRange = Left PutLimitNotAllowedError
   | otherwise = Right (topLevelRange, ranges)
   where
-    headerRange = rangeRequested hdrs
+    headerRange = if method == "GET" then rangeRequested hdrs else allRange
     limitRange = fromMaybe allRange (HM.lookup "limit" qsRanges)
     headerAndLimitRange = rangeIntersection headerRange limitRange
     -- Bypass all the ranges and send only the limit zero range (0 <= x <= -1) if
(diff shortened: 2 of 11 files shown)


laurence-isla:
My current approach is based on [this convo](https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1033481329) which led to the conclusion that we should disallow the Range header altogether for mutation requests (mostly for `POST` requests).

Now, I have a doubt with [what the RFC mentions](https://www.rfc-editor.org/rfc/rfc9110.html#name-range):

> A server MUST ignore a Range header field received with a request method that is unrecognized or for which range handling is not defined. For this specification, [GET](https://www.rfc-editor.org/rfc/rfc9110.html#GET) is the only method for which range handling is defined.

I understand that "ignore" means we could continue the request as if the header wasn't there? But it could be argued that we do define Range handling by returning a `400` status like we do with `PUT` and now, in this PR.  Or am I wrong in any of those cases?

steve-chavez:
> My current approach is based on https://github.com/PostgREST/postgrest/issues/2156#issuecomment-1033481329 which led to the conclusion that we should disallow the Range header altogether for mutation requests (mostly for POST requests).

Agree.

> A server MUST ignore a Range header field received with a request method that is unrecognized or for which range handling is not defined. For this specification, [GET](https://www.rfc-editor.org/rfc/rfc9110.html#GET) is the only method for which range handling is defined.

Agree as well. MUST for RFCs mean it's REQUIRED, so we should follow that behavior.

Really `Range` is quite broken even now at the master version. Which reminds me of https://github.com/PostgREST/postgrest/pull/2204, we should also merge that one for this release and label the breaking change.

So yes, `Range` should be ignored for anything besides GET. The `?limit/offset` thing can behave differently.

src/PostgREST/ApiRequest.hs
```diff
@@ -237,10 +237,10 @@ getRanges :: ByteString -> QueryParams -> RequestHeaders -> Either ApiRequestErr
 getRanges method QueryParams{qsOrder,qsRanges} hdrs
   | isInvalidRange = Left $ InvalidRange (if rangeIsEmpty headerRange then LowerGTUpper else NegativeLimit)
   | method `elem` ["PATCH", "DELETE"] && not (null qsRanges) && null qsOrder = Left LimitNoOrderError
-  | method == "PUT" && topLevelRange /= allRange = Left PutRangeNotAllowedError
+  | method == "PUT" && topLevelRange /= allRange = Left PutLimitNotAllowedError
   | otherwise = Right (topLevelRange, ranges)
   where
-    headerRange = rangeRequested hdrs
+    headerRange = if method == "GET" then rangeRequested hdrs else allRange
```
steve-chavez, line 243: A comment here citing the RFC would be great.

```suggestion
    headerRange = if method == "GET" then rangeRequested hdrs else allRange -- According to the RFC(https://www.rfc-editor.org/rfc/rfc9110.html#name-range), the Range header must be ignored for all methods other than GET 
```

CHANGELOG.md
```diff
@@ -33,6 +33,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + `PGRST003`: Timed out when acquiring connection to db
  - #2667, Fix `db-pool-acquisition-timeout` not logging to stderr when the timeout is reached - @steve-chavez
  - #1652, Fix function call with arguments not inlining - @steve-chavez
+ - #2705, Fix `Range` header not being ignored on request methods different from `GET` - @laurenceisla
+
+### Changed
+
+ - #2705, The `Range` header is now ignored when doing an RPC request with a method different from `GET` - @laurenceisla
```
steve-chavez, line 40: I think once https://github.com/PostgREST/postgrest/pull/2204 is merged we can say this is a fix. Like "The Range header is now RFC 9110 compliant...". Right now is just a change.

```suggestion
 
### Changed

 - #2705, The `Range` header is now ignored when doing an RPC request with a method different from `GET` - @laurenceisla
```

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -220,9 +220,55 @@ spec =
...
+          (rangeHdrs (ByteRangeFromTo 0 0)) mempty
+         `shouldMutateInto`
+         [json|[]|]
+
+        baseTable "limited_delete_items" "id" tblDataBefore
+         `mutatesWith`
+         requestMutation methodDelete "/limited_delete_items?order=id"
+          (rangeHdrs (ByteRangeFromTo 0 0)) mempty
+         `shouldMutateInto`
+         [json|[]|]
```
steve-chavez, line 243: Why is this operation repeated twice?

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -220,9 +220,55 @@ spec =
...
+
+        baseTable "limited_delete_items" "id" tblDataBefore
+         `mutatesWith`
+         requestMutation methodDelete "/limited_delete_items?order=id&limit=1&offset=1"
+          (rangeHdrs (ByteRangeFromTo 0 1)) mempty
+         `shouldMutateInto`
+         [json|[
+           { "id": 1, "name": "item-1" }
+         , { "id": 3, "name": "item-3" }
+         ]|]
```
steve-chavez, line 274: Maybe only one of these tests is necessary?

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -24,50 +24,65 @@ spec :: PgVersion -> SpecWith ((), Application)
...
-            , matchHeaders = [ matchContentTypeJson
-                             , "Content-Range" <:> "0-0/*" ]
-            }
+      context "returns paginated results" $ do
+        it "using the Range header" $
+          request methodGet "/rpc/getitemrange?min=2&max=4"
+                  (rangeHdrs (ByteRangeFromTo 1 1)) mempty
+             `shouldRespondWith` [json| [{"id":4}] |]
+              { matchStatus = 200
+              , matchHeaders = ["Content-Range" <:> "1-1/*"]
```
steve-chavez, line 33: Just noticed that our `Content-Range` also doesn't comply the spec https://www.rfc-editor.org/rfc/rfc9110.html#name-content-range. This can be fixed in another PR.

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -24,50 +24,65 @@ spec :: PgVersion -> SpecWith ((), Application)
...
-            ""
-            { matchStatus = 206
-            , matchHeaders = [ matchContentTypeJson
-                             , "Content-Range" <:> "0-0/2" ]
-            }
+        it "using limit and offset" $ do
+          post "/rpc/getitemrange?limit=1&offset=1" [json| { "min": 2, "max": 4 } |]
+             `shouldRespondWith` [json| [{"id":4}] |]
+              { matchStatus = 200
+              , matchHeaders = ["Content-Range" <:> "1-1/*"]
```
steve-chavez, line 40: Fortunately the spec doesn't restrict the `Content-Range` to only GET https://www.rfc-editor.org/rfc/rfc9110.html#name-content-range

So this response should be fine.

test/spec/Feature/Query/DeleteSpec.hs
laurence-isla: Ah, I added that specifically for the Limited Update/Delete features. Before this change, if the Range header is used with `order=id`, then limited update/delete works as if the ranges where limits. I think it should have it's own description instead of removing it. Because `limit` and `Range` are kinda tied together, maybe some changes could bypass this.

test/spec/Feature/Query/RpcSpec.hs
steve-chavez: Forgot that this was addressed on https://github.com/PostgREST/postgrest/pull/2204. Nevermind.

test/spec/Feature/Query/DeleteSpec.hs
laurence-isla: I forgot to separate it to a different test as I did with Updates. Before the changes the first one throws an "invalid range error".

test/spec/Feature/Query/DeleteSpec.hs
steve-chavez: Ah ok.. yeah a comment or a test with description would clarify.

steve-chavez (approved):

PR #2713, opened 2023-03-23
ci: Update ARM build script to use GHC 9.2.4

### LINKED ISSUE #2675: Update ARM script to use GHC 9.2.4
### Description of issue

As mentioned in https://github.com/PostgREST/postgrest/pull/2671#discussion_r1110994203, we won't use GHC 8.10.7 anymore, so the ARM build script should reflect that. The way the GHC version is detected should be changed, maybe doing `ghcup set ghc <ghc_version>` and if it fails install it. These are the relevant lines: 

https://github.com/PostgREST/postgrest/blob/2a2889020a1d4a822c77d4c6b77d4d8884b8691a/.github/scripts/arm/build.sh#L33-L45

laurence-isla:
Closes #2675 

Changes working on a test repository that connects to the ARM server: https://github.com/laurence-isla/github-actions-test/actions/runs/4504899339/jobs/7929954628

### DIFF
--- a/.github/scripts/arm/build.sh
+++ b/.github/scripts/arm/build.sh
@@ -9,6 +9,8 @@ PGRST_GITHUB_COMMIT="$1"
 SCRIPT_DIR="$2"
 
 DOCKER_BUILD_DIR="$SCRIPT_DIR/docker-env"
+CABAL_VERSION="3.6.0.0"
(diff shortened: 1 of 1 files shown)


steve-chavez (approved):
:fire:

PR #2718, opened 2023-03-26
nix: Update nixpkgs to 2023-03-25

wolfgangwalther:
Because... why not?

### DIFF
--- a/nix/nixpkgs-version.nix
+++ b/nix/nixpkgs-version.nix
@@ -1,6 +1,6 @@
 # Pinned version of Nixpkgs, generated with postgrest-nixpkgs-upgrade.
 {
-  date = "2023-01-12";
-  rev = "92f9580a4c369b4b51a7b6a5e77da43720134c9f";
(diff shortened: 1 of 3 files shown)


robx:
> So the MacOS nix build fails. @robx I think you added that job. I have no way to debug that, could you have a look?

Looks like the build is timing out. In https://github.com/PostgREST/postgrest/pull/2615, I added a github workflow that can be triggered manually to incrementally seed cachix -- you could try running that on this PR branch? Possibly it also passes in the meantime with hydra having caught up -- I just triggered a rerun of the failing job.

robx:
> > In #2615, I added a github workflow that can be triggered manually to incrementally seed cachix -- you could try running that on this PR branch?
> 
> Aha. How do I do that? When I go to the actions page for this workflow, I can't select my branch. Do I need to create a branch in the main repo instead of my fork? Or do I need to do run the action in my fork instead, setting up cachix tokens before that?

Yeah, I think pushing a branch to the main repo is probably the most straightforward way, that's what I did for the last upgrade.

PR #2723, opened 2023-03-27
Change to `Prefer: missing=default`

steve-chavez:
Following up on https://github.com/PostgREST/postgrest/pull/2672#issuecomment-1484050151.

Also do the feedback from https://github.com/PostgREST/postgrest/pull/2677#discussion_r1148521792


src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -215,13 +215,13 @@ instance ToAppliedHeader PreferTransaction
 -- |
 -- How to handle the insertion/update when the keys specified in ?columns are not present
 -- in the json body.
-data PreferUndefinedKeys
-  = ApplyDefaults  -- ^ Use the default column value for the unspecified keys.
-  | IgnoreDefaults -- ^ Inserts: null values / Updates: the keys are not SET to any value
+data PreferMissing
+  = ApplyDefaults  -- ^ Use the default column value for missing values.
+  | ApplyNulls     -- ^ Use the null value for missing values. Inserts: null values / Updates: the keys are not SET to any value
```
laurence-isla, line 220:
```suggestion
  | ApplyNulls     -- ^ Use the null value for missing values.
```
I wrongly added this here. For updates, this sets to `null` values and does not ignore them (I proposed to add another header value for those cases here: https://github.com/PostgREST/postgrest/pull/2693#discussion_r1148326675)

PR #2724, opened 2023-03-28
Add support for bulk deletes

laurence-isla:
Closes https://github.com/PostgREST/postgrest/issues/2314

- [ ] Full support for Bulk DELETE
- [ ] Tests
- [ ] Cleaning and refactor
- [ ] Changelog


laurence-isla:
This PR changes the logic used for bulk deletes that was specified in https://github.com/PostgREST/postgrest/pull/2355#issue-1291906948.

As it is mentioned in https://github.com/PostgREST/postgrest/issues/2314#issuecomment-1428792805, bulk deletes could also use the same header as bulk updates `Prefer: params=multiple-objects; unique_keys=id`. That would mean the following:

- Deletes won't detect the body unless that header is used.
- PostgREST detects only the `PK` inside the JSON body and delete accordingly. 
- An error is returned when limits are used.

Now, I'm not so sure about how to handle the `?columns` query parameter, since it's different for updates and inserts, where they are used to modify/add values. In the previous PR, we used any key in the JSON body as filters and `?columns` helped to restrict the body and avoid parsing it. In this case, we could add that behavior: any key in `?columns` is added to the filter in the `WHERE.

Although that would get in conflict  with the future implementation of the header params: `unique_keys=id.

steve-chavez:
> Although that would get in conflict with the future implementation of the header params: `unique_keys=id.

Really that was just a loose idea, even the naming is wrong for a header.

So the problem with using `?columns` for filters is that while it would work for delete, it won't for update - we need to come up with a consistent interface.

I'd say keep `columns` working as it is, it shouldn't affect this feature or bulk updates. You can scope it to PKs on the body for now.

If say a pk is not included in `columns`, that should be validated in a way that's not dependent on `columns`. I think we discussed how to do this before, similarly to PUT, a query can ensure every row has a PK.

PR #2726, opened 2023-03-29
refactor: remove character varying and text casting

steve-chavez:
Following discussion on https://github.com/PostgREST/postgrest/pull/1582#issuecomment-1488954603.

So our `ResultSet` type(used for all results of our generated queries) has a body which is a ByteString:

https://github.com/PostgREST/postgrest/blob/7629eff51d7df69bfa7dedfabd5beea18cb7a8cc/src/PostgREST/Query/Statements.hs#L38-L55

This is decoded as a `bytea` with Hasql:

https://github.com/PostgREST/postgrest/blob/7629eff51d7df69bfa7dedfabd5beea18cb7a8cc/src/PostgREST/Query/Statements.hs#L177-L187

At some point I assume it was decoded as `text`, but this is no longer the case. I assume that's why it was needed to cast the result body.

### DIFF
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -180,13 +180,13 @@ asCsvF = asCsvHeaderF <> " || '\n' || " <> asCsvBodyF
 
 asJsonF :: Bool -> SqlFragment
 asJsonF returnsScalar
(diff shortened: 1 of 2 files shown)


steve-chavez:
Github Actions is downgraded, that should explain the errors here.

https://www.githubstatus.com/

steve-chavez:
Seems safe to merge.

PR #2737, opened 2023-04-05
RPC returning table alias now works for pg 11/12

steve-chavez:
The new LATERAL query used for calling the function, introduced on https://github.com/PostgREST/postgrest/pull/2677, failed on functions that returned a domain like `CREATE DOMAIN projects_domain AS projects`.

Work around that by changing the query conditionally, by obtaining a bool that represents the composite alias on the SchemaCache and only do this on pg 11 and 12.

### DIFF
--- a/src/PostgREST/SchemaCache/Proc.hs
+++ b/src/PostgREST/SchemaCache/Proc.hs
@@ -13,6 +13,7 @@ module PostgREST.SchemaCache.Proc
   , procReturnsSingleComposite
   , procReturnsVoid
   , procTableName
+  , procReturnsCompositeAlias
   ) where
 
 import qualified Data.Aeson          as JSON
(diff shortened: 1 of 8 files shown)


src/PostgREST/SchemaCache/Proc.hs
```diff
@@ -25,7 +26,7 @@ import Protolude
 
 data PgType
   = Scalar Bool -- True if the type is void
-  | Composite QualifiedIdentifier
+  | Composite QualifiedIdentifier Bool -- True if the composite is a domain alias(used to workaround a bug in pg 11 and 12, see QueryBuilder.hs)
```
laurence-isla, line 29:
```suggestion
  | Composite QualifiedIdentifier Bool -- True if the composite is a domain alias(used to work around a bug in pg 11 and 12, see QueryBuilder.hs)
```

laurence-isla (approved):

PR #2739, opened 2023-04-06
feat: Allow filtering by IS DISTINCT FROM

### LINKED ISSUE #2501: Add support for other comparison predicates such as `IS DISTINCT FROM`
### Environment

* PostgreSQL version: v14.0
* PostgREST version: v10.0

### Description of issue

For completion, it would be nice to have the other [comparison predicates](https://www.postgresql.org/docs/current/functions-comparison.html) such as ~~`BETWEEN` and~~ `IS DISTINCT FROM`.  Based on [this SO question](https://stackoverflow.com/q/73941708/17265105).

laurence-isla:
Closes #2501.

As it was mentioned in https://github.com/PostgREST/postgrest/issues/2501#issuecomment-1496618749, the `BETWEEN` operator won't be implemented.

### DIFF
--- a/src/PostgREST/ApiRequest/Types.hs
+++ b/src/PostgREST/ApiRequest/Types.hs
@@ -229,6 +229,7 @@ data SimpleOperator
   | OpAdjacent
   | OpMatch
   | OpIMatch
+  | OpIsDistinctFrom
   deriving Eq
 
 -- | Operators for full text search operators
--- a/src/PostgREST/ApiRequest/Types.hs
+++ b/src/PostgREST/ApiRequest/Types.hs
@@ -191,6 +191,7 @@ data Operation
   = Op SimpleOperator SingleVal
   | In ListVal
   | Is TrileanVal
+  | IsDistinctFrom SingleVal
   | Fts FtsOperator (Maybe Language) SingleVal
   deriving (Eq)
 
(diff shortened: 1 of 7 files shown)


src/PostgREST/ApiRequest/Types.hs
```diff
@@ -229,6 +229,7 @@ data SimpleOperator
   | OpAdjacent
   | OpMatch
   | OpIMatch
+  | OpIsDistinctFrom
```
steve-chavez, line 232: One thing here. So on https://github.com/PostgREST/postgrest/pull/2738 I'm implementing any/all modifiers for operators, and I've realized this fits in pretty well with the `SimpleOperator` types.. because they all can use them like:

```sql
select 1 = any('{3,2,1}');
select 1 > any('{3,2,1}');
```

However `IS DISTINCT FROM` cannot be used in that way.

Could you make it instead be a part of `Operation`? Like:

```haskell
data Operation
  = Op SimpleOperator (Maybe OpQuantifier) SingleVal
  | In ListVal
  | Is TrileanVal
  | IsDistinctFrom SingleVal
  | Fts FtsOperator (Maybe Language) SingleVal
  deriving (Eq)
```

steve-chavez (approved):

PR #2740, opened 2023-04-06
Add access to the PostgREST version through SQL

### LINKED ISSUE #2647: Obtain the PostgREST version through SQL
It would be nice to obtain the version like this:

```sql
select distinct application_name from pg_stat_activity where application_name like '%postgrest%';

 application_name
------------------
postgrest-10.1.2
```

The [fallback_application_name](https://www.postgresql.org/docs/9.4/libpq-connect.html#LIBPQ-CONNECT-FALLBACK-APPLICATION-NAME) fits this use case exactly, we can set it and still allow users to override it with `application_name`.

To implement this, we'd have to add `?fallback_application_name=val` to `configDbUri` here:

https://github.com/PostgREST/postgrest/blob/09249f97ab301f5c46e7430b06ebbf93b5f5c071/src/PostgREST/AppState.hs#L100

Where `val` would be the current postgREST version.

The `?` should be added conditionally. If the uri already contains `?`, then add `&fallback_application_name=val` else add `?fallback_application_name=val`.

---

This function could be used for testing:

```sql
CREATE FUNCTION test.get_version() RETURNS pg_stat_activity.application_name%TYPE
AS $$
  select distinct application_name from pg_stat_activity where application_name like '%postgrest%';
$$ LANGUAGE sql;
```

laurenceisla:
Closes #2647.

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -34,6 +34,7 @@ import qualified Hasql.Pool                 as SQL
 import qualified Hasql.Session              as SQL
 import qualified Hasql.Transaction.Sessions as SQL
 import qualified PostgREST.Error            as Error
+import           PostgREST.Version          (prettyVersion)
 
 import Control.AutoUpdate (defaultUpdateSettings, mkAutoUpdate,
                            updateAction)
@@ -137,7 +138,7 @@ initPool AppConfig{..} =
     (fromIntegral configDbPoolAcquisitionTimeout)
     (fromIntegral configDbPoolMaxLifetime)
     (fromIntegral configDbPoolMaxIdletime)
-    (toUtf8 $ addPgrstVerToDbUri configDbUri)
+    (toUtf8 $ addPgrstVerToDbUri prettyVersion configDbUri)
 
 -- | Run an action with a database connection.
 usePool :: AppState -> SQL.Session a -> IO (Either SQL.UsageError a)
@@ -419,7 +420,7 @@ listener appState = do
 
   -- forkFinally allows to detect if the thread dies
   void . flip forkFinally (handleFinally dbChannel) $ do
-    dbOrError <- acquire $ toUtf8 (addPgrstVerToDbUri configDbUri)
+    dbOrError <- acquire $ toUtf8 (addPgrstVerToDbUri prettyVersion configDbUri)
     case dbOrError of
       Right db -> do
         logWithZTime appState $ "Listening for notifications on the " <> dbChannel <> " channel"
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -47,6 +47,7 @@ import Data.Time          (ZonedTime, defaultTimeLocale, formatTime,
 import Data.Time.Clock    (UTCTime, getCurrentTime)
 
 import PostgREST.Config                  (AppConfig (..),
+                                          addPgrstVerToDbUri,
                                           readAppConfig)
 import PostgREST.Config.Database         (queryDbSettings,
                                           queryPgVersion,
@@ -136,7 +137,7 @@ initPool AppConfig{..} =
     (fromIntegral configDbPoolAcquisitionTimeout)
     (fromIntegral configDbPoolMaxLifetime)
     (fromIntegral configDbPoolMaxIdletime)
-    (toUtf8 configDbUri)
+    (toUtf8 $ addPgrstVerToDbUri configDbUri)
 
 -- | Run an action with a database connection.
 usePool :: AppState -> SQL.Session a -> IO (Either SQL.UsageError a)
@@ -418,7 +419,7 @@ listener appState = do
 
   -- forkFinally allows to detect if the thread dies
   void . flip forkFinally (handleFinally dbChannel) $ do
-    dbOrError <- acquire $ toUtf8 configDbUri
+    dbOrError <- acquire $ toUtf8 (addPgrstVerToDbUri configDbUri)
     case dbOrError of
       Right db -> do
         logWithZTime appState $ "Listening for notifications on the " <> dbChannel <> " channel"
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -48,7 +48,7 @@ import Data.Time          (ZonedTime, defaultTimeLocale, formatTime,
 import Data.Time.Clock    (UTCTime, getCurrentTime)
 
 import PostgREST.Config                  (AppConfig (..),
-                                          addPgrstVerToDbUri,
+                                          addFallbackAppName,
                                           readAppConfig)
 import PostgREST.Config.Database         (queryDbSettings,
                                           queryPgVersion,
@@ -138,7 +138,7 @@ initPool AppConfig{..} =
     (fromIntegral configDbPoolAcquisitionTimeout)
     (fromIntegral configDbPoolMaxLifetime)
     (fromIntegral configDbPoolMaxIdletime)
-    (toUtf8 $ addPgrstVerToDbUri prettyVersion configDbUri)
+    (toUtf8 $ addFallbackAppName prettyVersion configDbUri)
 
 -- | Run an action with a database connection.
 usePool :: AppState -> SQL.Session a -> IO (Either SQL.UsageError a)
@@ -420,7 +420,7 @@ listener appState = do
 
   -- forkFinally allows to detect if the thread dies
   void . flip forkFinally (handleFinally dbChannel) $ do
-    dbOrError <- acquire $ toUtf8 (addPgrstVerToDbUri prettyVersion configDbUri)
+    dbOrError <- acquire $ toUtf8 (addFallbackAppName prettyVersion configDbUri)
     case dbOrError of
       Right db -> do
         logWithZTime appState $ "Listening for notifications on the " <> dbChannel <> " channel"
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -491,6 +491,6 @@ addFallbackAppName version dbUri = dbUri <>
     _        -> "&" <> uriFmt
   where
     uriFmt = toS $ escapeURIString isUnescapedInURI $ toS (pKeyWord <> pgrstVer)
-    keyValFmt = pKeyWord <> "'" <> pgrstVer <> "'"
+    keyValFmt = pKeyWord <> "'" <> T.replace "'" "\\'" pgrstVer <> "'"
     pKeyWord = "fallback_application_name="
     pgrstVer = "PostgREST " <> T.decodeUtf8 version
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -467,6 +467,7 @@ readPGRSTEnvironment =
 -- | Adds a `fallback_application_name` value to the connection string. This allows querying the PostgREST version on pg_stat_activity.
 --
 -- >>> let ver = "11.1.0 (5a04ec7)"::ByteString
+-- >>> let strangeVer = "11'1&0@#$%,.:\"[]{}?+^()asdfqwer"::ByteString
 --
 -- >>> addFallbackAppName ver "postgres://user:pass@host:5432/postgres"
 -- "postgres://user:pass@host:5432/postgres?fallback_application_name=PostgREST%2011.1.0%20(5a04ec7)"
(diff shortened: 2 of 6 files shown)


src/PostgREST/AppState.hs
```diff
@@ -111,10 +113,17 @@ destroy = destroyPool
...
 initPool :: AppConfig -> IO SQL.Pool
 initPool AppConfig{..} =
-  SQL.acquire configDbPoolSize timeoutMicroseconds $ toUtf8 configDbUri
+  SQL.acquire configDbPoolSize timeoutMicroseconds configDbUriAndAppName
   where
     timeoutMicroseconds = (* oneSecond) <$> configDbPoolAcquisitionTimeout
     oneSecond = 1000000
+    configDbUriAndAppName = toUtf8 configDbUri <>
+      case uriQuery <$> parseURI (toS configDbUri) of
+        Nothing  -> " " <> appNameParam
```
laurence-isla, line 122: Testing this for Keyword/Value connection strings. Not fail proof, though. E.g. this string works and connects to the database, ignoring the `?` in password:

```
"postgres://authenticator:mysecr?????etpassword@localhost:5432/postgres?"
```

But `Network.URI` does not recognize it as a valid URI.

src/PostgREST/AppState.hs
```diff
@@ -111,10 +113,17 @@ destroy = destroyPool
...
+        Just ""  -> "?" <> appNameParam
+        Just "?" -> appNameParam
+        _        -> "&" <> appNameParam
+    appNameParam = "fallback_application_name='PostgREST " <> prettyVersion <> "'"
```
laurence-isla, line 126: I think a pretty print could be used instead of kebab-case. For instance, my setup returns this when doing:

```sql
select distinct application_name from pg_stat_activity;
```

|application_name|
|-|
|IntelliJ IDEA 2022.2.3|
|PostgREST 10.2.0.20230209 (pre-release) (16f2849)|

It looks good, only the commit hash and perhaps the "(pre-release)" should go.

src/PostgREST/AppState.hs
steve-chavez: Hmm, it's a bit hard to follow. Maybe we can have some doctests for this function?

steve-chavez:
Check https://github.com/PostgREST/postgrest/issues/2647#issuecomment-1536704518. It might not be necessary to mess with the conn URI.

steve-chavez:
There's an issue with the env var approach on https://github.com/PostgREST/postgrest/issues/2647#issuecomment-1581414068.

Also, seems https://github.com/openstreetmap/osm2pgsql/issues/867 also adds the `fallback_application_name` by parsing the connstring. Maybe check their implementation.

laurenceisla:
> Also, seems [openstreetmap/osm2pgsql#867](https://github.com/openstreetmap/osm2pgsql/issues/867) also adds the `fallback_application_name` by parsing the connstring. Maybe check their implementation.

They don't use a URI connection string but a Key-Value one instead. Also they receive each connection parameter separately by the user (user, password, database,  etc.) and then build the connstring themselves, so it's easier to add the fallback parameter there.

steve-chavez:
> They don't use a URI connection string but a Key-Value one instead.
> and then build the connstring themselves

Got it. We have it a bit more complicated. But we should be fine if we doctest(maybe io test too) our connstring mutation.

src/PostgREST/Config.hs
```diff
@@ -219,7 +221,7 @@ readAppConfig dbSettings optPath prevDbUri roleSettings roleIsolationLvl = do
     decodeLoadFiles :: AppConfig -> IO AppConfig
     decodeLoadFiles parsedConfig =
       decodeJWKS <$>
-        (decodeSecret =<< readSecretFile =<< readDbUriFile prevDbUri parsedConfig)
+        (decodeSecret =<< readSecretFile =<< addPgrstVerToDbUri =<< readDbUriFile prevDbUri parsedConfig)
```
laurenceisla, line 224: Maybe not the place to add the variable, a `--dump-config` shows that the parameter is included twice. IO-tests also fail because the `fallback_application_name` is now present in the querystring.

laurenceisla:
I need some confirmation in the way this feature behaves:

1. When doing a

    ```sql
    select application_name 
    from pg_stat_activity;
    ```
    It returns the version twice:
    ```
    application_name
    ----------------------------
     PostgREST 11.1.0 (11a9849)    
     PostgREST 11.1.0 (11a9849)
    ```
   One due to the [initial connection](https://github.com/PostgREST/postgrest/blob/c1a8661ab38e6878f24441a9ce98b7c9c4127a0c/src/PostgREST/AppState.hs#L132) and other for the [listener](https://github.com/PostgREST/postgrest/blob/c1a8661ab38e6878f24441a9ce98b7c9c4127a0c/src/PostgREST/AppState.hs#LL408C1-L408C1). 
2. After that, the initial connection is released once the `db-pool-acquisition-timeout` is reached and it will show only one `PostgREST 11.1.0 (11a9849)` due to the listener.
3. If the listener is deactivated by `db-channel-enabled=false` then no `PostgREST <version>` will be shown after the acquisition timeout. That is, selecting the `application_name` will show no PostgREST version (no active connection is in the pool).
4. After that timeout, only doing a request that reaches the database will show the version again after querying the `application_name` in the database. Once it's released it won't show it anymore.

steve-chavez:
@laurenceisla Those are not problems IMO. This is just an extra nicety for debugging. We can document it's not reliable if there's no pg connection(and by default there will always be one if the listen channel is up). For 1, the user can do a distinct(https://github.com/PostgREST/postgrest/issues/2647, this will be documented too).

laurenceisla:
I'm having a problem with the doctest. Running `postgrest-run-doctests` in a local environment and the CI throws this error:

```bash
src/PostgREST/Config.hs:469: failure in expression `addPgrstVerToDbUri "postgres://user:pass@host:5432/postgres"'
expected: "postgres://user:pass@host:5432/postgres?fallback_application_name=PostgREST%20..."
 but got: 
          ^
          <interactive>:27:1: error:
              Variable not in scope: addPgrstVerToDbUri :: t0 -> t

Examples: 150  Tried: 147  Errors: 0  Failures: 1
```

But `stack test` works normally:

```bash
postgrest> test (suite: doctests)
                       
Examples: 150  Tried: 150  Errors: 0  Failures: 0

postgrest> Test suite doctests passed
postgrest> test (suite: spec)
```

steve-chavez:
@laurenceisla So the full error is:

```
src/PostgREST/Version.hs:12:1: error:
    Could not find module ‘Paths_postgrest’
    Perhaps you meant Paths_doctest
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
src/PostgREST/Config.hs:469: failure in expression `addPgrstVerToDbUri "postgres://user:pass@host:5432/postgres"'
expected: "postgres://user:pass@host:5432/postgres?fallback_application_name=PostgREST%20..."
 but got:
          ^
          <interactive>:27:1: error:
              Variable not in scope: addPgrstVerToDbUri :: t0 -> t

Examples: 150  Tried: 147  Errors: 0  Failures: 1
```

See:
```
    Could not find module ‘Paths_postgrest’
    Perhaps you meant Paths_doctest
```

Seems the doctest environment can find `Paths_postgrest`. This is bc you import `PostgREST.Version` in `Config.hs` and that is dependent on `Paths_postgrest`.

steve-chavez:
Tried adding 

```
postgrest.cabal

test-suite doctests
  other-modules:      Paths_postgrest
```

But no dice

steve-chavez:
Related https://github.com/sol/doctest/pull/307

steve-chavez:
Fixed the doctests by making the `addPgrstVerToDbUri` pure, i.e. not depend on `Paths_postgrest`.

src/PostgREST/Config.hs
```diff
@@ -460,3 +462,34 @@ type Environment = M.Map [Char] Text
...
+-- "postgresql://?fallback_application_name=PostgREST%20..."
+addFallbackAppName :: ByteString -> Text -> Text
+addFallbackAppName version dbUri = dbUri <>
+  case uriQuery <$> parseURI (toS dbUri) of
+    Nothing  -> " " <> keyValFmt -- Assume key/value connection string if the uri is not valid
+    Just ""  -> "?" <> uriFmt
+    Just "?" -> uriFmt
+    _        -> "&" <> uriFmt
+  where
+    uriFmt = T.replace " " "%20" $ pKeyWord <> pgrstVer
```
steve-chavez, line 492: @laurenceisla Instead of manually encoding the spaces here, how about using https://hackage.haskell.org/package/uri-encode-1.5.0.7/docs/Network-URI-Encode.html#v:encode?

Our `prettyVersion` should only contain spaces as non-url safe chars but let's be extra safe here.

src/PostgREST/Config.hs
```diff
@@ -460,3 +462,34 @@ type Environment = M.Map [Char] Text
 readPGRSTEnvironment :: IO Environment
 readPGRSTEnvironment =
   M.map T.pack . M.fromList . filter (isPrefixOf "PGRST_" . fst) <$> getEnvironment
+
+-- | Adds a `fallback_application_name` value to the connection string. This allows querying the PostgREST version on pg_stat_activity.
+--
+-- >>> let ver = "11.1.0 (5a04ec7)"::ByteString
+--
+-- >>> addFallbackAppName ver "postgres://user:pass@host:5432/postgres"
+-- "postgres://user:pass@host:5432/postgres?fallback_application_name=PostgREST%20..."
```
steve-chavez, line 471: Btw, I've noticed these `...` work as a wildcard. Do you have a link for this behavior?

laurenceisla: Yes, found it in their repo's README: https://github.com/sol/doctest#matching-arbitrary-output

src/PostgREST/Config.hs
```diff
@@ -460,3 +463,34 @@ type Environment = M.Map [Char] Text
...
+-- "host=localhost port=5432 dbname=postgres fallback_application_name='PostgREST 11.1.0 (5a04ec7)'"
+--
+-- >>> addFallbackAppName ver "postgresql://"
+-- "postgresql://?fallback_application_name=PostgREST%2011.1.0%20(5a04ec7)"
+addFallbackAppName :: ByteString -> Text -> Text
+addFallbackAppName version dbUri = dbUri <>
+  case uriQuery <$> parseURI (toS dbUri) of
+    Nothing  -> " " <> keyValFmt -- Assume key/value connection string if the uri is not valid
+    Just ""  -> "?" <> uriFmt
+    Just "?" -> uriFmt
```
steve-chavez, line 490: How about adding a case where `?` is present so codecov doesn't complain? I think we can reuse some of:

https://github.com/PostgREST/postgrest/blob/0a1564ba5aaa8175dcb738271967944652cf4ce9/test/io/test_io.py#L69-L75

src/PostgREST/Config.hs
```diff
@@ -460,3 +463,41 @@ type Environment = M.Map [Char] Text
...
+-- "host=localhost port=5432 dbname=postgres fallback_application_name='PostgREST 11.1.0 (5a04ec7)'"
+--
+-- >>> addFallbackAppName ver "postgresql://"
+-- "postgresql://?fallback_application_name=PostgREST%2011.1.0%20(5a04ec7)"
+--
+-- >>> addFallbackAppName strangeVer "host=localhost port=5432 dbname=postgres"
+-- "host=localhost port=5432 dbname=postgres fallback_application_name='PostgREST 11\\'1&0@#$%,.:\"[]{}?+^()asdfqwer'"
+--
+-- >>> addFallbackAppName strangeVer "postgres:///postgres?host=server&port=5432"
+-- "postgres:///postgres?host=server&port=5432&fallback_application_name=PostgREST%2011'1&0@#$%25,.:%22[]%7B%7D?+%5E()asdfqwer"
```
steve-chavez, line 491: Shouldn't this result urlencode the `&`? Seems this strange version would break the connstring.

laurenceisla: Yes. I replaced it with the `isUnescapedInURIComponent`, it now escapes every special char not including the allowed ones (e.g. `.`), better safe than sorry.

test/io/test_io.py
```diff
@@ -998,3 +998,38 @@ def test_openapi_in_big_schema(defaultenv):
...
+    defaultenv_without_libpq = {
+        key: value
+        for key, value in defaultenv.items()
+        if key not in ["PGDATABASE", "PGHOST", "PGUSER"]
+    }
+    env = {**defaultenv_without_libpq, "PGRST_DB_URI": uri}
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.post("/rpc/get_pgrst_version")
+        assert response.text.startswith('"PostgREST ')
```
steve-chavez, line 1035: Now that we escape the version, it'd be good to test the version is actually formatted in the way we expect.

I suggest putting the full version here and comparing it by obtaining the version from `postgrest.cabal`. The `commit` still will not be possible to compare without breaking later so that part can be left out.

laurenceisla: I saw that the `prettyVersion` is also used to return the version in the `Server` header. Maybe we could use that to compare (already added an example to the tests).

steve-chavez (approved):

PR #2744, opened 2023-04-11
v10.2.0 MINOR release

laurence-isla:
Cherry picked the following fixes/features for a stable 10.2.0 version:

```markdown
## Added

 - #2663, Limit maximal postgresql connection lifetime - @robx
   + New option `db-pool-max-lifetime` (default 30m)
   + `db-pool-acquisition-timeout` is no longer optional and defaults to 10s
   + Fixes postgresql resource leak with long-lived connections (#2638)

### Fixed

 - #2667, Fix `db-pool-acquisition-timeout` not logging to stderr when the timeout is reached - @steve-chavez
```


steve-chavez (approved):
Looks good!

Releasing off the main branch does take some work but I think it's worth it since we make v10 stable.

Our pre-releases numbering also look good this time, since 10.2.0 can be considered lower than v10.2.0.20230407.

PR #2746, opened 2023-04-13
ci: bring back freebsd build to release bundle

laurenceisla:
Since the MR #2608 fixed the issue with FreeBSD builds in CI, then it should be safe to bring back these builds for the release bundle.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -313,7 +313,7 @@ jobs:
       - Test-Memory-Nix
       - Build-Static-Nix
       - Build-Stack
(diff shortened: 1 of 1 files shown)


laurence-isla:
It looks like some builds are still failing  https://github.com/PostgREST/postgrest/actions/runs/4659956760/jobs/8247447694 :

This artifact downloading step coincidentally does not have the authentication headers. Maybe that's the problem here?

https://github.com/PostgREST/postgrest/blob/e572d1d1a2ea14cd4505be2fa0c2c08e46d32997/.github/get_cirrusci_freebsd#L48-L49

Issue #1139, opened 2018-07-06
JWTIssuedAtFuture failure for valid JWT token


dudleycarr:
### Environment

* PostgreSQL version: docker:latest
* PostgREST version: 0.5
* Operating system: GCP Kubernetes

### Description of issue

Postgrest intermittently returns ```{"message":"JWTIssuedAtFuture"}``` for valid JWT tokens. I verified that there was no clock drift between the client and the PostgREST server.

(Steps to reproduce: Include a minimal SQL definition plus how you make the request to PostgREST and the response body)

The easiest way to reproduce is to generate a new JWT token for each request. The issue seems to stem from the change to cache the system timestamp for 1 second within PostgREST. This was a non-issue prior to 0.5.

steve-chavez:
Reproduced. You're generating a JWT with an `iat` claim right?(provide a sample JWT if possible).
Now, the bug is also present in `v0.4.4.0`(see if you can test this version to double check), this was prior to the time caching addition, seems the problem appears when we upgrade the version of our jwt lib(this [PR](https://github.com/PostgREST/postgrest/pull/997)), I'll have to dig further into that.

Steps to reproduce:
```bash
## Must have the pgjwt extension and reallyreallyreallyreallyverysafe as jwt-secret in postgrest.conf
for i in {1..10}; do \
  JWT=$(psql -X -t -c "select sign(row_to_json(r), 'reallyreallyreallyreallyverysafe') from ( select 'postgres' as role, extract(epoch from now())::integer as iat) r" | xargs);\
  curl -s -H "Authorization: Bearer $JWT" localhost:3000/authors_only; echo;\
  sleep .5;\
done

{"message":"JWTIssuedAtFuture"}
[]
{"message":"JWTIssuedAtFuture"}
[]
{"message":"JWTIssuedAtFuture"}
[]
[]
{"message":"JWTIssuedAtFuture"}
[]
{"message":"JWTIssuedAtFuture"}
```

dudleycarr:
@steve-chavez Yes, the JWT does include the `iat` claim. I'll produce a rejected JWT later today. My bad on not checking the previous version for the existence of the bug.

steve-chavez:
@miaojianxin That seems more related to #991, check if your secret is at least 32 characters long.

steve-chavez:
Can confirm that by removing the time cache the issue still persists.

dudleycarr:
[...]

PR #2772, opened 2023-04-28
fix: skew of 30 seconds for JWT validation

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/1139.

Use a 30 seconds skew by default as per google recommendations [here](https://cloud.google.com/iap/docs/signed-headers-howto#verifying_the_jwt_payload). Previous 1 second skew was too strict. 

Hard to test for now. We'll just monitor if #1139 gets more reports.


steve-chavez:
So weird why suddenly the pg 15 tests fail.

PR #2786, opened 2023-05-15
feat: use hasql-pool-0.10, add db-pool-max-idletime

robx:
- new option db-pool-max-idletime limits the maximal idle time of a postgresql connection, defaults to ~~10 minutes~~ 30 seconds
- this essentially restores the db-pool-timeout option from versions <=10, with more specific name
- ~~temporary overrides to allow building hasql-notifications with hasql-pool-0.10 (see https://github.com/diogob/hasql-notifications/issues/15)~~


robx:
10 minute default is pretty arbitrary. I just wanted something shorter than the max-lifetime default since it's pointless otherwise, but I'm not sure that's a good reason.

steve-chavez:
Not sure what the correct default is for now.

On `db-pool-timeout` it used to be 10 seconds for us until v9.0.1, then it went to 1 hour(too high) for v10.0.0.

---

> 10 minute default is pretty arbitrary. I just wanted something shorter than the max-lifetime default since it's pointless otherwise, but I'm not sure that's a good reason.

10 mins is also the default on [HikariCP](https://github.com/brettwooldridge/HikariCP#frequently-used).

It also says:

> This setting only applies when minimumIdle is defined to be less than maximumPoolSize

Which I think we could document as well.

steve-chavez:
Ok, so I recall seeing some cases where users set https://postgresqlco.nf/doc/en/param/idle_session_timeout/ to 60 seconds as best practice.

I believe we should set the idle timeout lower than that. If anything we can prevent some errors showing in the logs when we recover.

So I'm thinking 30 seconds is a good number.

steve-chavez (approved):

PR #2795, opened 2023-05-22
feat: add db-pool-timeout alias

steve-chavez:
For db-pool-max-idletime. 

Follow up on #2786.

On https://github.com/PostgREST/postgrest/releases/tag/v10.1.0 we removed the `db-pool-timeout`. While removing the config didn't cause an error,  it still was a change in behavior. With this alias we can bring the old behavior back.

Planning to release a v11.0.2 and v10.2.1(backpatch) after this.


steve-chavez:
> and v10.2.1(backpatch) 

Don't feel comfortable doing this since this time we already have a major version after it. Will only release a new v11.

Issue #2703, opened 2023-03-11
Allow reading configuration indirectly using SQL provided as configuration parameter(`pre-config` hook)


mkleczek:
Currently configuration is read from authenticator role attributes. CREATEROLE privilege is needed to modify them.

It would be much more flexible to be able to configure an SQL query that PostgREST would use to retrieve configuration from the database.
That way some dynamic configuration scenarios might be supported (for example a list of schemas might be read from a table/view implementing a specific policy)

steve-chavez:
Hmm, one issue is that the `authenticator` role would need privilege over tables and that would go against the design of our role system, see [here](https://postgrest.org/en/stable/auth.html).

I get the disadvantage of the CREATEROLE privilege though. What other alternatives do we have?

mkleczek:
Maybe another configuration property with a role name to read the settings?

steve-chavez:
Hmm, another option could be a SECURITY DEFINER function. That way the `authenticator` wouldn't need any privilege on a table, only the privilege to execute the function. By default the EXECUTE priv is public too.

mkleczek:
Ok, so that goes back to the original idea: have a configuration parameter with a query to read configuration from the db and let the user configure the database roles with appropriate privileges (a query might be a SELECT from a SECURITY DEFINER function in particular). The feature is for advanced users and use cases anyway.

Some documentation with best practices would be a good idea.

steve-chavez:
It'd be better to be restrictive in this case, so I was thinking just a function name on the configuration.

Another alternative could be that the function only does a bunch of 

```sql
SET LOCAL pgrst.db_anon_role TO ..
SET LOCAL pgrst.jwt_secret TO .. 
```

That way we don't need to agree upon a RETURN type on the function, plus we don't need any privilege on a table.

We also keep the code for the config more or less the same.

mkleczek:
I am not really sure if it changes anything in terms of required privileges as the values of configuration parameters have to be read from somewhere so all necessary privileges to do that are required anyway.

I am not sure if restricting it to a function name is a good idea as it requires creating this function in the database (making it somewhat cumbersome).

I can imagine scenarios where one might want to perform a select on catalog views to expose schemas based on some criteria (ie. naming conventions and filtered by USAGE privileges).

mkleczek:
I am not sure but I don't think you need any special privileges to execute for example:

`WITH RECURSIVE cte AS (
[...]

Issue #2791, opened 2023-05-19
"reload schema" notifications are dropped when a schema reload is already in progress


gaborcs:
### Environment

* PostgreSQL version: public.ecr.aws/supabase/postgres:15.1.0.73
* PostgREST version: public.ecr.aws/supabase/postgrest:v10.1.2
* Operating system: macOS

### Description of issue

Schema reloading doesn't always work well for us. I investigated the issue, and it seems to be caused by the fact that a "reload schema" notification "does nothing if the previous invocation has not terminated": https://github.com/PostgREST/postgrest/blob/f9f0f79fa914ac00c11fbf7f4c558e14821e67e2/src/PostgREST/Workers.hs#L54-L58

This can be problematic in some cases, such as the following scenario:
1. A schema reload is triggered.
2. PostgREST requests the schema.
3. The DB receives the request and responds with the schema.
4. The DB schema is changed again, that triggers a new "reload schema" notification.
5. PostgREST receives the new "reload schema" notification, but it drops it because there's another schema reload in progress.
6. PostgREST receives the schema sent at step 3, which is outdated at this point.

This doesn't occur very commonly, but since we switched to using Flyway and its afterVersioned callbacks to drop and recreate all of our DB functions, it does happen about once out of 5 or 10 times, as each afterVersioned callback triggers a schema reload. As a workaround, I've added the following afterMigrate callback to reload the schema after everything else has completed:
```
SELECT pg_sleep(1);
NOTIFY pgrst, 'reload schema';
```

This fixes the issue for us, but I think schema reloading should be more reliable in PostgREST.

PR #2803, opened 2023-05-26
Add `pre-config` hook

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/2703.

Allows using the [in-database configuration](https://postgrest.org/en/v11.0/references/configuration.html#in-database-configuration) without SUPERUSER. You can have a function that configures postgREST like so:

```sql
returns void as $$
begin
  if current_user = 'other_authenticator' then
    perform
    , set_config('pgrst.db_schemas', 'schema1, schema2', true)
    , set_config('pgrst.jwt_role_claim_key', '."other"."pre_config_role"', true)
    , set_config('pgrst.db_anon_role', 'pre_config_role', true)
    , set_config('pgrst.db_tx_end', 'rollback-allow-override', true);
  else
    null;
  end if;
end $$ language plpgsql;
```

The function name and schema is arbitrary. It should be kept separate from public APIs(not unsafe to expose it bc it doesn't return anything, but for maintenance).

A single function can be used for different postgREST instances. Each instance can use a different connection role and different settings can be applied(like the above example with the condition on `current_user`).


src/PostgREST/Config/Database.hs
```diff
@@ -36,37 +64,53 @@ pgVersionStatement = SQL.Statement sql HE.noParams versionRow
     sql = "SELECT current_setting('server_version_num')::integer, current_setting('server_version')"
     versionRow = HD.singleRow $ PgVersion <$> column HD.int4 <*> column HD.text
 
-queryDbSettings :: Bool -> Session [(Text, Text)]
-queryDbSettings prepared =
+-- | Query the in-database configuration. The settings have the following priorities:
+--
+-- 1. Role + with database-specific settings: ALTER ROLE authenticator IN DATABASE postgres SET <prefix>jwt_aud = 'val';
+-- 2. Role + with settings: ALTER ROLE authenticator SET <prefix>jwt_aud = 'overridden';
+-- 3. pre-config function: CREATE FUNCTION pre_config() .. PERFORM set_config(<prefix>jwt_aud, 'pre_config_aud'..)
```
laurenceisla, line 71:
```suggestion
-- 1. Role + with database-specific settings:
--    ALTER ROLE authenticator IN DATABASE postgres SET <prefix>jwt_aud = 'val';
-- 2. Role + with settings:
--    ALTER ROLE authenticator SET <prefix>jwt_aud = 'overridden';
-- 3. pre-config function: 
--    CREATE FUNCTION pre_config() .. PERFORM set_config(<prefix>jwt_aud, 'pre_config_aud'..)
```

test/io/db_config.sql
```diff
@@ -49,9 +53,6 @@ ALTER ROLE other_authenticator SET pgrst.openapi_server_proxy_uri = 'https://oth
 ALTER ROLE other_authenticator SET pgrst.raw_media_types = 'application/vnd.pgrst.other-db-config';
 ALTER ROLE other_authenticator SET pgrst.jwt_secret = 'ODERREALLYREALLYREALLYREALLYVERYSAFE';
 ALTER ROLE other_authenticator SET pgrst.jwt_secret_is_base64 = 'true';
-ALTER ROLE other_authenticator SET pgrst.jwt_role_claim_key = '."other"."role"';
-ALTER ROLE other_authenticator SET pgrst.db_anon_role = 'other';
-ALTER ROLE other_authenticator SET pgrst.db_tx_end = 'rollback-allow-override';
```
laurenceisla, line 54: Ah, you removed these because they wouldn't be set by the function otherwise, right? Perhaps it needs another authenticator user for testing the overrides with the pre config function?

test/io/db_config.sql
```diff
@@ -63,6 +64,27 @@ ALTER ROLE other_authenticator SET pgrst.openapi_mode = 'disabled';
...
+grant usage on schema postgrest to db_config_authenticator;
+grant usage on schema postgrest to other_authenticator;
+
+-- pre-config hook
+create or replace function postgrest.pre_config()
+returns void as $$
+begin
+  if current_user = 'db_config_authenticator' then
+    null;
+  elsif current_user = 'other_authenticator' then
```
laurenceisla, line 77:
```suggestion
  if current_user = 'other_authenticator' then
```

Or is the `db_config_authenticator` included to make it more explicit?

test/io/db_config.sql
steve-chavez: Yes, correct. Hm, I thought of a new role but `other_authenticator` already has some redundant testing with `db_config_authenticator`. Applying it to an existing role also ensures the logic is correct.

test/io/db_config.sql
steve-chavez: Yeah, just for explicitness. But maybe it looks confusing, so I'd be fine with removing it from the conditional.

steve-chavez:
Just tried this for reproducing https://github.com/PostgREST/postgrest/issues/2791. No dice.

It happens before the schema cache reload on Workers.hs, which makes sense(schema cache needs the db-schema config).

We need a way to inject a function into the schema cache for reproducing https://github.com/PostgREST/postgrest/issues/2791. This would also be useful for inducing a Hasql error and increase code coverage(currently we cannot cover the `InternalErrorCode00`/`PGRSTX00` errors). This should be an internal only config parameter bc there's no feature that needs this.

Issue #2781, opened 2023-05-12
LISTEN/NOTIFY doesn't work on read replicas


steve-chavez:
## Problem

Read replicas don’t support LISTEN/NOTIFY([ref](https://stackoverflow.com/questions/70293290/can-you-perform-listen-unlisten-on-a-postgresql-read-replica)). Thus, PostgREST [NOTIFY reloading](https://postgrest.org/en/stable/schema_cache.html#reloading-with-notify) won’t work in them.

As a workaround, we can kill the pool connections:

```sql
select pg_terminate_backend(pid) 
from pg_stat_activity 
where application_name = 'postgrest';
```

(can be enabled conditionally with `pg_is_in_recovery()`)

This will cause a schema cache reload because of [Automatic Recovery](https://postgrest.org/en/latest/references/connection_pool.html#automatic-recovery).

However, the recovery won't be graceful. Two requests have to be done to start recovery.

```bash
$ curl localhost:3000/projects -i
HTTP/1.1 500 Internal Server Error
Transfer-Encoding: chunked
Date: Fri, 12 May 2023 02:13:02 GMT
Server: postgrest/11.0.1 (4197d2f)
Content-Type: application/json; charset=utf-8

{"code":"57P01","details":null,"hint":null,"message":"terminating connection due to administrator command"}

$ curl localhost:3000/projects -i
HTTP/1.1 503 Service Unavailable
Transfer-Encoding: chunked
Date: Fri, 12 May 2023 02:13:10 GMT
Server: postgrest/11.0.1 (4197d2f)
Retry-After: 0
Content-Type: application/json; charset=utf-8

{"code":"PGRST001","details":"no connection to the server\n","hint":null,"message":"Database client error. Retrying the connection."}

$ curl localhost:3000/projects -i
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Date: Fri, 12 May 2023 02:13:14 GMT
Server: postgrest/11.0.1 (4197d2f)
Content-Range: 0-4/*
Content-Location: /projects
Content-Type: application/json; charset=utf-8

[{"id":1,"name":"Windows 7","client_id":1},
 {"id":2,"name":"Windows 10","client_id":1},
 {"id":3,"name":"IOS","client_id":2},
 {"id":4,"name":"OSX","client_id":2},
 {"id":5,"name":"Orphan","client_id":null}]
```

## Solution

[...]

PR #2810, opened 2023-06-01
Fix dropping schema cache reload notifications

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/2791

## Tasks

- [x] Reproduce the failure by introducing a `internal-schema-cache-sleep` config that calls `pg_sleep` at the end of the schema cache load.
- [x] Fix by introducing a debouncer instead of a semaphore for the connection worker

## Notes

- To introduce the debounce(which has to store state in AppState), I had to move the `connectionWorker`(and others) to the AppState module. It had to be done this way otherwise it caused a dependency cycle and compiling failed.
  + I think this move makes sense. Workers.hs was filled with AppState logic.
  + In doing so the code shortens a bit bc the AppState prefix is removed.
  + State is easier to see and this will help to solve https://github.com/PostgREST/postgrest/issues/2781.
  + A clearer Admin.hs module is introduced.


steve-chavez:
All the CI failures now show that the failure is reproduced:

- https://github.com/PostgREST/postgrest/actions/runs/5149959715/jobs/9273574222?pr=2810
- https://github.com/PostgREST/postgrest/actions/runs/5149959715/jobs/9273574343?pr=2810
- https://github.com/PostgREST/postgrest/actions/runs/5149959715/jobs/9273574816?pr=2810

test/io/test_io.py
```diff
@@ -946,6 +946,44 @@ def test_isolation_level(defaultenv):
...
+        response = postgrest.session.post("/rpc/migrate_function")
+        assert response.status_code == 204
+
+        time.sleep(
+            2 * internal_sleep
+        )  # wait enough time to ensure the broken state remains
+
+        response = postgrest.session.get("/rpc/mult_them?c=3&d=4")
+        assert response.text == "12"
+        assert response.status_code == 200
```
steve-chavez, line 984: This was easy to test manually but really tricky in the io tests :sweat:

steve-chavez:
Tests that prove the failure:

- https://github.com/PostgREST/postgrest/actions/runs/5151261608/jobs/9276259589?pr=2810
- https://github.com/PostgREST/postgrest/actions/runs/5151261608/jobs/9276259650?pr=2810
- (other pg versions are the same)

steve-chavez:
Codecov failing bc of all the missing tests mentioned on https://github.com/PostgREST/postgrest/issues/1766

src/PostgREST/AppState.hs
```diff
@@ -96,15 +107,24 @@ initWithPool pool conf = do
...
     let oneSecond = 1000000 in
     mkDebounce defaultDebounceSettings
        { debounceAction = logPgrstError appState SQL.AcquisitionTimeoutUsageError
        , debounceFreq = 5*oneSecond
        , debounceEdge = leadingEdge -- logs at the start and the end
        }
 
-  return appState { debounceLogAcquisitionTimeout = deb }
+  debWorker <-
+    let decisecond = 100000 in
```
steve-chavez, line 120: This is enough to ensure schema cache reloading is still fast(io tests pass) but we still debounce.

Also tried manually with a bash loop `for ((;;)) { kill -s SIGUSR1 $(pgrep -n postgrest); }` and it's working fine.

steve-chavez:
Tests in main are flaky now. Will fix on a new PR.

PR #2811, opened 2023-06-02
nix: use similar config to tests on postgrest-run 

steve-chavez:
Otherwise manual tests can be deceiving. As commented on https://github.com/PostgREST/postgrest/pull/2810#discussion_r1213841669.

### DIFF
--- a/nix/tools/cabalTools.nix
+++ b/nix/tools/cabalTools.nix
@@ -41,6 +41,8 @@ let
         args =
           [
             "ARG_USE_ENV([PGRST_DB_ANON_ROLE], [postgrest_test_anonymous], [PostgREST anonymous role])"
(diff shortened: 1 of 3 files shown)


steve-chavez:
So strange, after removing the big_schema.sql from the io tests, I get the following error:

```bash
Run postgrest-with-postgresql-9.6 -f test/io/fixtures.sql postgrest-test-io -vv
psql:test/io/fixtures.sql:9[4](https://github.com/PostgREST/postgrest/actions/runs/5160106490/jobs/9295833509?pr=2811#step:5:5): ERROR:  return type mismatch in function declared to return text
DETAIL:  Actual return type is unknown.
CONTEXT:  SQL function "hello"
```

It's a simple fix. But it's strange that we didn't got this error before.

laurenceisla (approved):

Issue #2763, opened 2023-04-21
Add a HTML table response type


saolof:
With [HTMX](https://htmx.org/) making HATEOAS RESTful endpoints that return data as HTML a bit more mainstream, I personally often wish that postgrest had a way to make lists return a sequence of lines with each line wrapped in a <tr> and each field in a <td> tag. It is a fairly simple string substitution, but it would be a nice quality of life improvement to have postgrest do it.

This is mainly useful for implementing something like [this example](https://htmx.org/examples/active-search/) without any need to write any actual code outside of the database and the HTML, without needing any CRUD backend other than postgrest. With that said, this is something that already can be done now from a JSON response by pulling in the htmx client-side-templates extension and using a mustache template that iterates over a top level array, so this would be more of a quality of life improvement.

steve-chavez:
Yes, we're working on returning different media types(like HTML) on https://github.com/PostgREST/postgrest/pull/1582. I'm making this a priority for the next release.

I've also long wanted to use HTMX and in general want to simplify the frontend stack.

steve-chavez:
@saolof There's now a how-to for HTMX! https://postgrest.org/en/latest/how-tos/providing-html-content-using-htmx.html

Let us know if you have any feedback on how to improve the integration.

Issue #2699, opened 2023-03-09
geojson structure: crs inside each feature instead on level type and features


BirGie123:
Hello, 
we are using postgREST for crud functions on our ai-geodata.  

we saw that the geo+json header responses a featurecollection with coordinate reference systems (crs). 

the crs is in each feature instead outside the features. 

can you reproduce this behavior and if yes is it possible that the structure of the response is adjusted by you?

### Environment

* PostgreSQL version: 12
* PostgREST version: 10
* PostGIS version: 3.1 USE_GEOS=1 USE_PROJ=1 USE_STATS=1
* Operating system: Kubernetes/Ubuntu20

#### request: 
`curl -X GET https://url.com//result_frame_field?run_uuid=eq.11341ed6-4713-c28c-0d70-d290032abd73 -H "Accept: application/geo+json"`

#### response:

`{"type" : "FeatureCollection", 
"features" : [{
"type": "Feature", 
"geometry": {"type":"Polygon",
"crs":{"type":"name","properties":{"name":"EPSG:25832"}},"coordinates":[[[489633.9,5888000],[489634.042248535,5887995.982175827],[489636.885974121,5887995.468318176],[489637.14420166,5887997.19841671],[489637.2,5888000],[489633.9,5888000]]]}, "properties": {"id": 321386, "run_uuid": "11341ed6-4713-c28c-0d70-d290032abd73", "confidence": 0.8387449445279349, "date": "2023-03-03T13:58:01+00:00", "path": null}}, {"type": "Feature", "geometry": {"type":"Polygon","crs":{"type":"name","properties":{"name":"EPSG:25832"}},"coordinates":[[[489999.9,5887952.3],[489998.616845703,5887949.44701233],[489998.868896484,5887949.294552612],[489997.822460938,5887946.764746094],[489998.847167969,5887945.881781006],[489999.085253906,5887944.549810791],[489999.283007812,5887946.699108886],[489999.9,5887949.4],[489999.9,5887952.3]]]}, "properties": {"id": 321387, "run_uuid": "11341ed6-4713-c28c-0d70-d290032abd73", "confidence": 0.5394721098729639, "date": "2023-03-03T13:58:01+00:00", "path": null}}]}`

#### request: 
`curl -X GET https://url.com//result_frame_field?run_uuid=eq.11341ed6-4713-c28c-0d70-d290032abd73 -H "Accept: application/geo+json"`

#### response:

`{
"type": "FeatureCollection",
[...]

Issue #2170, opened 2022-02-17
Don't fail on RPC with`Accept: text/plain, application/json`


steve-chavez:
A request like :

```http
GET /rpc/jsonfunc
Accept: text/plain, application/json
```

Gives `406 Not Acceptable`. We prefer `text/plain` in this case because both `q` values are 1.0 by default and there's an order preference. If it were `Accept: text/plain; q=0.2, application/json; q=0.8`, then `application/json` would be picked and the request would succeed.

The RFC doesn't specify that ordering is relevant according to this [so answer](https://stackoverflow.com/a/47196708/4692662). So we could give preference to `application/json` over `text/plain`, because the latter is only supported on certain conditions while the former is supported on every request.

The problem comes from https://github.com/PostgREST/postgrest/discussions/2162#discussioncomment-2193532.

This might require patching [parseHttpAccept](https://hackage.haskell.org/package/wai-extra-3.1.8/docs/Network-Wai-Parse.html#v:parseHttpAccept).

wolfgangwalther:
I think we need to separate two issues here:
- The bug in #2162 seems to be, that we are treating content negotiation (deciding which mediatype to respond to) and "finding the right function to call for this request" separately - while they need to merged into one. Before deciding for any of the mimetypes listed in the `Accept` header, we need to see which of those are actually actionable for the given set of arguments etc. - in this case, it would only be `application/json`, because there is no function found that can handle a `text/plain` type. But we settle on `text/plain` way too early right now - before we even look into the list of functions.
- The other question is what should happen, when both of the given mimetypes are indeed acceptable, i.e. we have overloaded functions that could handle both mimetypes and they do have the same priority. You suggest that we should give priority to `application/json` in this case - and I don't think we should. From a set of equally weighted types, it would be absolutely fine to choose the first one that is possible to choose.

steve-chavez:
> Before deciding for any of the mimetypes listed in the Accept header, we need to see which of those are actually actionable for the given set of arguments etc
> From a set of equally weighted types, it would be absolutely fine to choose the first one that is possible to choose.

Agree with both :+1:

Issue #1548, opened 2020-06-29
Feature request: allow user-defined function to handle application-specific Accept headers


W1M0R:
Consider a request `/accounts` that returns a list of accounts and their details. 

Request header `Accept: application/json` will return the information as JSON. However, some clients might want the data returned in some other application-specific representation, e.g. protobuf base64, flatbuffer base64, msgpack base64, etc.

Currently, one would have to implement a generic stored procedure (say `/rpc/msgpack`)  that will parse the body to determine what content is requested (since we lose the `/accounts` query url info), then to call the stored procedure that implements `/accounts`, convert the result to json and then again to msgpack. However, with this approach we lose the benefit of having PostgREST handling filtering and resource selection with query parameters.

Ideally, the client should be able to query `/accounts` but use an application-defined Accept header, let's say `Accept: application/msgpack`. PostgREST can be started with a configuration option that indicates the name of a user-defined Postgres function (e.g. `myschema.myhandler(accept text, result jsonb)`) that will handle any unknown Accept headers. When PostgREST receives the `/accounts` request, it should determine that the provided Accept header is not part of the default PostgREST supported headers, and having confirmed that `myschema.myhandler)` exists, it will fulfill the `/accounts` query, but before sending the result to the user, it will forward the results to `myschema.myhandler` and return the result of that function instead. 

The `myschema.myhandler` function will use its `accept` parameter to determine that the user requested `application/msgpack` data, and then convert the `result` parameter to msgpack and return that instead.

Concerning the return value of `myschema.myhandler`, it should probably also be a user-defined custom type.

steve-chavez:
@W1M0R I think the rpc part could be handled by https://github.com/PostgREST/postgrest/pull/1546#issuecomment-649094238.

That PR also includes some thought for regular `GET /table` Accept logic via `pre-request`, perhaps you can chime in there.

wolfgangwalther:
Not sure where to post this, as this topic seems to be spread across multiple issues and PR now. Posting it in here, because I think this is the best approach so far.

I would like to suggest a couple of additions to that as well:

# Content Negotiation (RPC)

I don't really like the whole "raw-media-endpoints config option" approach - I think it would be a lot better, if we could somehow pass the information about which accept headers a function might respond to in the function definition itself. The best I could come up with in this regard would be to do it like this:

```SQL
CREATE FUNCTION my_rpc_with_custom_accept_header (...) RETURNS bytea AS $$
  [...]
[...]

PR #2825, opened 2023-06-15
SQL handlers for custom media types

steve-chavez:
## Summary

Custom media types are now possible on RPC and tables by using scalar functions and aggregates respectively.

Docs in progress: https://github.com/PostgREST/postgrest-docs/pull/689

## Examples

- Create custom `text/csv` handler that includes a BOM (byte order mark) plus `Content-Disposition` header:

```sql
create domain "text/csv" as text;

-- override CSV with BOM plus attachment
create or replace function test.bom_csv_trans (state text, next test.lines)
returns text as $$
  select state || next.id::text || ',' || next.name || ',' || next.geom::text || E'\n';
$$ language sql;

create or replace function test.bom_csv_final (data text)
returns "text/csv" as $$
  select set_config('response.headers', '[{"Content-Disposition": "attachment; filename=\"lines.csv\""}]', true);
  -- EFBBBF is the BOM in UTF8 https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8
  select convert_from (decode (E'EFBBBF', 'hex'),'UTF8') || (E'id,name,geom\n' || data));
$$ language sql;

drop aggregate if exists test.bom_csv_agg(test.lines);
create aggregate test.bom_csv_agg (test.lines) (
  initcond = ''
, stype = text
, sfunc = bom_csv_trans
, finalfunc = bom_csv_final
);
```

```bash
curl 'localhost:3000/lines?id=in.(1,2)' -H "Accept: text/csv"  -i

HTTP/1.1 200 OK
Content-Disposition: attachment; filename="lines.csv"

id,name,geom
1,line-1,0102000020E610000002000000000000000000F03F000000000000F03F00000000000014400000000000001440
2,line-2,0102000020E6100000020000000000000000000040000000000000004000000000000018400000000000001840

# works for RPC that returns a table type too
curl 'localhost:3000/rpc/get_lines?id=eq.1' -H "Accept: text/csv"  -i

HTTP/1.1 200 OK
Content-Disposition: attachment; filename="lines.csv"

id,name,geom
1,line-1,0102000020E610000002000000000000000000F03F000000000000F03F00000000000014400000000000001440
```

- Create a custom `application/geo+json` generic handler (`anyelement` can work thanks to `ST_AsGeoJSON` being generic):

```sql
create domain "application/geo+json" as jsonb;

create or replace function test.lines_geojson_trans (state jsonb, next anyelement)
returns jsonb as $$
  select state || extensions.ST_AsGeoJSON(next)::jsonb;
$$ language sql;

create or replace function test.lines_geojson_final (data jsonb)
returns "application/geo+json" as $$
  select jsonb_build_object(
    'type', 'FeatureCollection',
    'crs',  json_build_object(
        'type',      'name',
        'properties', json_build_object(
            'name', 'EPSG:4326'
         )
     ),
    'features', data
  );
$$ language sql;

drop aggregate if exists test.lines_geojson_agg(anyelement);
create aggregate test.lines_geojson_agg (anyelement) (
  initcond = '[]'
, stype = jsonb
, sfunc = lines_geojson_trans
, finalfunc = lines_geojson_final
);
```

```bash
curl 'localhost:3000/lines?id=in.(1,2)' -H "Accept: application/geo+json" 

{"crs": {"type": "name", "properties": {"name": "EPSG:4326"}}, "type": "FeatureCollection", "features": [{"type": "Feature", "geometry": {"type": "LineString", "coordinates": [[1, 1], [5, 
5]]}, "properties": {"id": 1, "name": "line-1"}}, {"type": "Feature", "geometry": {"type": "LineString", "coordinates": [[2, 2], [6, 6]]}, "properties": {"id": 2, "name": "line-2"}}]}
```

- Create a custom html handler response

```sql
create domain "text/html" as text;

create or replace function welcome() returns "text/html" as $_$
select $$
<html>
  <head>
    <title>Custom handler for HTML</title>
  </head>
  <body>
    <h1>Welcome!</h1>
  </body>
</html>
$$;
$_$ language sql;
```

```
curl 'localhost:3000/rpc/welcome' -H "Accept: text/html"

<html>
  <head>
    <title>Custom handler for HTML</title>
  </head>
  <body>
    <h1>Welcome!</h1>
  </body>
</html>
```

## Breaks

 - Removed [raw-media-types config](https://postgrest.org/en/v11.1/references/configuration.html#raw-media-types)
 - Removed `application/octet-stream`, `text/plain`, `text/xml` [builtin support for scalar results](https://postgrest.org/en/v11.1/references/api/resource_representation.html#scalar-function-response-format)
 - Removed default `application/openapi+json` media type for [db-root-spec](https://postgrest.org/en/v11.1/references/configuration.html#db-root-spec)
 - All can be migrated without impacting API clients by using custom handlers.

## Closes

- Closes https://github.com/PostgREST/postgrest/issues/1548
- Closes https://github.com/PostgREST/postgrest/issues/2699
- Closes https://github.com/PostgREST/postgrest/issues/2763
- Closes https://github.com/PostgREST/postgrest/issues/2170
- Closes https://github.com/PostgREST/postgrest/issues/1462
- Closes https://github.com/PostgREST/postgrest/issues/1102
- Closes https://github.com/PostgREST/postgrest/issues/1374
- Closes https://github.com/PostgREST/postgrest/issues/2901. No longer think it's necessary to use an external module.

## Pending

- Both https://github.com/PostgREST/postgrest/issues/2188 and  https://github.com/PostgREST/postgrest/issues/1043 would need https://github.com/PostgREST/postgrest/issues/2826


test/spec/Feature/Query/PlanSpec.hs
```diff
@@ -241,7 +242,7 @@ spec actualPgVersion = do
           resStatus  = simpleStatus r
 
       liftIO $ do
-        resHeaders `shouldSatisfy` elem ("Content-Type", "application/vnd.pgrst.plan+text; charset=utf-8")
+        resHeaders `shouldSatisfy` elem ("Content-Type", "application/vnd.pgrst.plan+text; for=\"application/json\"; charset=utf-8")
```
steve-chavez, line 245: Will move the plan media type changes to another PR. Overall this media type was hard to get right for this feature.

steve-chavez: Done in https://github.com/PostgREST/postgrest/pull/2833

steve-chavez:
There's a problem in solving https://github.com/PostgREST/postgrest/issues/2699 with the current aggregate interface.

So having a replacement for our geojson is simple:

```sql
create domain "application/geo+json" as jsonb;

create or replace function test.geoson_trans (state jsonb, next anyelement)
returns jsonb as $$
  select state || extensions.ST_AsGeoJSON(next)::jsonb;
$$ language sql;

create or replace function test.geojson_final (data jsonb)
returns "application/geo+json" as $$
  select jsonb_build_object('type', 'FeatureCollection', 'features', data);
$$ language sql;

create or replace aggregate test.geojson_agg (anyelement) (
  initcond = '[]'
, stype = jsonb
, sfunc = geoson_trans
, finalfunc = geojson_final
);

curl "localhost:3000/shops" -H "Accept: application/geo+json"

{"type": "FeatureCollection", "features": [{"type": "Feature", "geometry": {"type": "Point", "coordinates": [-71.10044, 42.373695]}, "properties": {"id": 1, "address": "1369 Cambridge St"}}, {
"type": "Feature", "geometry": {"type": "Point", "coordinates": [-71.10543, 42.366432]}, "properties": {"id": 2, "address": "757 Massachusetts Ave"}}, {"type": "Feature", "geometry": {"type":
"Point", "coordinates": [-71.081924, 42.36437]}, "properties": {"id": 3, "address": "605 W Kendall St"}}]}
```

If we want to add a `crs` to the final geojson, we could do it like:

```sql
create or replace function test.geojson_final (data jsonb)
returns "application/geo+json" as $$
  select jsonb_build_object(
    'type', 'FeatureCollection', 
    'crs',  json_build_object(
        'type',      'name', 
        'properties', json_build_object(
            'name', 'EPSG:4326'  
         )
     ),
    'features', data
  );
$$ language sql;
```

That's not a generic way though, ideally we use [Find_SRID](https://postgis.net/docs/manual-2.0/Find_SRID.html) as recommend on #2699.

For this, we could use FINALFUNC_EXTRA:

```sql
create or replace function test.geojson_final (data jsonb, typ anyelement)
returns "application/geo+json" as $$
  -- here, assume we enhance pg_typeof with what's required for find_SRID to work, i.e. getting the schema, name and column. Which is possible.
  select jsonb_build_object('type', 'FeatureCollection', 'type', pg_typeof(typ), 'features', data);
$$ language sql;

create or replace aggregate test.geojson_agg (anyelement) (
  initcond = '[]'
, stype = jsonb
, sfunc = geoson_trans
, finalfunc = geojson_final
, finalfunc_extra = true
);

curl "localhost:3000/shops" -H "Accept: application/geo+json"
{"type": "FeatureCollection", "type": "record", "features": [{"type": "Feature", "geometry": {"type": "Point", "coordinates": [-71.10044, 42.373695]}, "properties": {"id": 1, "address": "1369
 Cambridge St"}}, {"type": "Feature", "geometry": {"type": "Point", "coordinates": [-71.10543, 42.366432]}, "properties": {"id": 2, "address": "757 Massachusetts Ave"}}, {"type": "Feature", "g
eometry": {"type": "Point", "coordinates": [-71.081924, 42.36437]}, "properties": {"id": 3, "address": "605 W Kendall St"}}]}
```

Notice there's a `type: record`. This is bc of how we call the aggregation:

```haskell
customAggF :: QualifiedIdentifier -> SqlFragment
customAggF qi = fromQi qi <> "(_postgrest_t)"
```

If we casted `_postgrest_t::test.shops`, then the correct `type: shops` would show. 

The casting won't work if resource embedding is done. So maybe we can cast it only when embedding is not used?

src/PostgREST/SchemaCache.hs
```diff
@@ -1025,6 +1044,71 @@ allViewsKeyDependencies =
...
       |]
 
+initialMediaAggregates :: MediaAggregateMap
+initialMediaAggregates =
+  HM.insert (RelAnyElement, MediaType.MTAny            ) BuiltinAggJson $
+  HM.insert (RelAnyElement, MediaType.MTApplicationJSON) BuiltinAggJson $
+  HM.insert (RelAnyElement, MediaType.MTSingularJSON   ) BuiltinAggSingleJson $
+  HM.insert (RelAnyElement, MediaType.MTTextCSV        ) BuiltinAggCsv $
+  HM.insert (RelAnyElement, MediaType.MTGeoJSON        ) BuiltinAggGeoJson
+  HM.empty
```
steve-chavez, line 1054: The implementation is similar to the idea on https://github.com/PostgREST/postgrest/pull/1582#issuecomment-1483019360

steve-chavez:
I was under the wrong impression that we could do everything with aggregates. For instance, creating an aggregate that also strips nulls https://github.com/PostgREST/postgrest/issues/1601. For this we would only need to apply a function to the final function of the aggregate.

`json_agg` uses `json_agg_finalfn` and `json_agg_transfn`. So:

```sql
create or replace function test.appjson_agg_final (data json)
returns "application/json" as $$
  select json_strip_nulls(pg_catalog.json_agg_finalfn($1));
$$ language sql;

-- ERROR:  function pg_catalog.json_agg_finalfn(json) does not exist
-- LINE 3:   select json_strip_nulls(pg_catalog.json_agg_finalfn($1));
-- HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

create or replace aggregate test.appjson_agg (anyelement) (
  initcond = '[]'
, stype = json
, sfunc = pg_catalog.json_agg_transfn
, finalfunc = appjson_agg_final
);
```

The above won't work bc `json_agg_finalfn` argument is `internal`:

```sql
postgres=# \df+ json_agg_finalfn
List of functions
-[ RECORD 1 ]-------+------------------------------
Schema              | pg_catalog
Name                | json_agg_finalfn
Result data type    | json
Argument data types | internal
```



Wolfgang idea with CASTs would solve this case https://github.com/PostgREST/postgrest/pull/1582#issuecomment-1483817669. So looks we'll need CASTs anyway.

```sql
CREATE CAST ("application/json" AS json) WITH FUNCTION  json_strip_nulls;
```

I'm thinking we should merge https://github.com/PostgREST/postgrest/pull/2523 before going further on this feature.

steve-chavez:
> I was under the wrong impression that we could do everything with aggregates. For instance, creating an aggregate that also strips nulls https://github.com/PostgREST/postgrest/issues/1601

Alternatively, I think this could be solved later with the `postgrest-contrib` extension. We might be able to define custom aggregates there. Even some that could integrate with our tx settings (request.method, etc). With this we could avoid the need for CASTs too (though I won't preclude the possibility of implementing that in the future), which were deemed complicated by a user recently ([ref](https://github.com/PostgREST/postgrest/discussions/2924#discussioncomment-6865121)) .

test/spec/fixtures/schema.sql
```diff
@@ -2631,12 +2635,12 @@ create table plate_plan_step (
 	  REFERENCES well(well_id)
 );
 
-CREATE FUNCTION test.return_scalar_xml() RETURNS pg_catalog.xml
+CREATE FUNCTION test.return_scalar_xml() RETURNS "text/xml"
 LANGUAGE sql AS $$
   SELECT '<my-xml-tag/>'::pg_catalog.xml
 $$;
```
steve-chavez, line 2641: This means https://postgrest.org/en/stable/how-tos/create-soap-endpoint.html will have to be changed to:

```sql
CREATE OR REPLACE FUNCTION my_soap_endpoint(xml) RETURNS "text/xml" AS $$
```

In theory we could assume `RETURNS xml` means `text/xml` but that would only work for scalars. It would be messy in the code.

The change is not so bad anyway, just needs a new line for `CREATE DOMAIN "text/xml" AS xml`

test/spec/fixtures/schema.sql
```diff
@@ -3471,3 +3519,60 @@ stable
...
+
+    drop aggregate if exists test.twkb_agg(test.lines);
+    create aggregate test.twkb_agg (test.lines) (
+      initcond = ''
+    , stype = "application/vnd.twkb"
+    , sfunc = test.twkb_handler_transition
+    , finalfunc = test.twkb_handler_final
+    );
+
+  END IF;
```
steve-chavez, line 3571: So to be compatible with pg <= 12, the domain needs to be present as `stype` plus casting on the transition. I guess that should be ok for perf since the types are binary compatible? 

Otherwise we'll get the following errors:

- https://github.com/PostgREST/postgrest/actions/runs/6583374462/job/17886321550?pr=2825
- https://github.com/PostgREST/postgrest/actions/runs/6583374462/job/17886321667?pr=2825 

If we need to do this anyway, I wonder if we should drop the requirement for the `finalfunc`? It would be one function less, which seems to be boilerplate most of the time. We could get the media type from the `stype`?

@wolfgangwalther WDYT?

steve-chavez: @wolfgangwalther You're right! Thanks for the hint!


So in the end we can use both ways by only querying `pg_proc`. My query in the schema cache was looking at `pg_aggregate` unnecessarily.

```sql
postgres=# \df+ twkb_agg
                                                                                  List of functions
 Schema |   Name   |    Result data type    | Argument data types | Type | Volatility | Parallel |  Owner   | Security | Access privileges | Language |   Source code   | Description 
--------+----------+------------------------+---------------------+------+------------+----------+----------+----------+-------------------+----------+-----------------+-------------
 test   | twkb_agg | "application/vnd.twkb" | lines               | agg  | immutable  | unsafe   | postgres | invoker  |                   | internal | aggregate_dummy | 
(1 row)
```

The resulting data type is preserved in both of the above queries (compatible with pg>=13000 and down).

With this the user is free to use a final func or not.

test/spec/Feature/Query/CustomMediaSpec.hs
```diff
@@ -0,0 +1,176 @@
...
+        { matchStatus  = 200
+        , matchHeaders = ["Content-Type" <:> "application/vnd.geo2+json"]
+        }
+
+      request methodGet "/rpc/get_shop_bles" (acceptHdrs "application/vnd.geo2+json") ""
+        `shouldRespondWith`
+        "\SOH\"anyelement overridden\""
+        { matchStatus  = 200
+        , matchHeaders = ["Content-Type" <:> "application/vnd.geo2+json"]
+        }
```
steve-chavez, line 68: For some reason, these tests are giving a `\SOH` ([Start of Heading](https://www.ascii-code.com/character/%E2%90%81)) output. I've no idea if this comes from pg or Haskell, I think it's the latter.

steve-chavez:
As a consequence of https://github.com/PostgREST/postgrest/pull/2894, `application/vnd.pgrst.object` and `application/vnd.pgrst.object` are no longer overridable per our original design, since they take parameters (`nulls=stripped`) now.

Maybe it's not so bad to make vendored media types non-overridable. After all users cannot override `vnd.pgrst.plan` too, this one has a lot of special parameters.

test/spec/Feature/Query/CustomMediaSpec.hs
```diff
@@ -0,0 +1,221 @@
...
+    it "will match image/png according to q values" $ do
+      r1 <- request methodGet "/rpc/ret_image" (acceptHdrs "image/png, */*") ""
+      liftIO $ do
+        simpleBody r1 `shouldBe` readFixtureFile "A.png"
+        simpleHeaders r1 `shouldContain` [("Content-Type", "image/png")]
+
+      r2 <- request methodGet "/rpc/ret_image" (acceptHdrs "text/html,application/xhtml+xml,application/xml;q=0.9,image/png,*/*;q=0.8") ""
+      liftIO $ do
+        simpleBody r2 `shouldBe` readFixtureFile "A.png"
+        simpleHeaders r2 `shouldContain` [("Content-Type", "image/png")]
```
steve-chavez, line 197: Clears the case in https://github.com/PostgREST/postgrest/issues/1462, replace `image/png` for `image/webp`. This is more about correct content negotiation.

steve-chavez:
Ok this should be ready to go. It implements the **first stage** idea discussed on https://github.com/PostgREST/postgrest/pull/1582#issuecomment-1592559765. This means that the CASTs idea can be built on top.

There might be some rough edges, but it already solves a lot of issues. Merging it later will cost a lot (it already took a lot of rebases), so I'd rather merge it at this stage and improve it later.

steve-chavez:
Merging. Will continue with https://github.com/PostgREST/postgrest/issues/2826 to solve the remaining issues.

steve-chavez:
With `raw-media-types` gone, https://postgrest.org/en/stable/how-tos/providing-images-for-img.html is not possible now. We'd need a way to dynamically set the accepted media type inside the function.

PR #2828, opened 2023-06-18
refactor: plan media type

steve-chavez:
## Problem

Previously it was impossible to have doctests for the MediaType module due to recursion:

https://github.com/PostgREST/postgrest/blob/78821a8fe7990022bb9036c24321ab411c55a3b1/src/PostgREST/MediaType.hs#L23-L38

Basically a `MediaType` was able to contain another `MediaType`. See the `MTPlan` constructor.

This caused the `deriving instance Show ...`  setup to fail.

## Solution

Eliminate recursion by having non-plan media types as a separate type(`NormalMedia`).

Some doctests are now added for `MediaType`.

### DIFF
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -1,7 +1,9 @@
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE LambdaCase            #-}
 
 module PostgREST.MediaType
   ( MediaType(..)
+  , NormalMedia(..)
   , MTPlanOption (..)
   , MTPlanFormat (..)
   , MTPlanAttrs(..)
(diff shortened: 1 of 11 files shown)


src/PostgREST/MediaType.hs
```diff
@@ -32,83 +49,95 @@ data MediaType
...
+-- | Convert from ByteString to MediaType.
+--
+-- >>> decodeMediaType "application/json"
+-- MTNormal MTApplicationJSON
+--
+-- >>> decodeMediaType "application/vnd.pgrst.plan;"
+-- MTPlan (MTPlanAttrs Nothing PlanText [])
+--
+-- >>> decodeMediaType "application/vnd.pgrst.plan;for=\"application/json\""
+-- MTPlan (MTPlanAttrs (Just MTApplicationJSON) PlanText [])
```
steve-chavez, line 119: Here I tried adding a space like so:

```suggestion
-- >>> decodeMediaType "application/vnd.pgrst.plan; for=\"application/json\""
-- MTPlan (MTPlanAttrs (Just MTApplicationJSON) PlanText [])
```

And the decoding failed to get `Just MTApplicationJSON` (got `Nothing` instead). This clearly works on the feature specs:

https://github.com/PostgREST/postgrest/blob/78821a8fe7990022bb9036c24321ab411c55a3b1/test/spec/Feature/Query/PlanSpec.hs#L119

I thought there must be something wrong with whitespace on the `doctests`. This would have been easy to prove with the repl. But https://github.com/PostgREST/postgrest/issues/2627 continues to be a problem.

PR #2835, opened 2023-06-21
ci: tweak arm builds using cabal

laurenceisla:
Related to #2834

Makes it similar to the Linux build with cabal:
- Remove the hard coded GHC version from the arm build script and puts it in the yaml file instead.
- Uses the [default cabal version by ghcup](https://www.haskell.org/ghcup/install/#which-versions-get-installed).

[Working in test repo](https://github.com/laurenceisla/github-actions-test/actions/runs/5338068914).

### DIFF
--- a/.github/scripts/arm/build.sh
+++ b/.github/scripts/arm/build.sh
@@ -10,7 +10,7 @@ PGRST_GITHUB_COMMIT="$1"
 SCRIPT_DIR="$2"
 
 DOCKER_BUILD_DIR="$SCRIPT_DIR/docker-env"
-CABAL_VERSION="recommended"
+CABAL_VERSION="latest"
 GHC_VERSION="$3"
 
 install_packages() {
--- a/.github/scripts/arm/build.sh
+++ b/.github/scripts/arm/build.sh
@@ -4,13 +4,14 @@
 
 [ -z "$1" ] && { echo "Missing 1st argument: PostgREST github commit SHA"; exit 1; }
 [ -z "$2" ] && { echo "Missing 2nd argument: Build environment directory name"; exit 1; }
+[ -z "$3" ] && { echo "Missing 3rd argument: GHC version"; exit 1; }
 
 PGRST_GITHUB_COMMIT="$1"
 SCRIPT_DIR="$2"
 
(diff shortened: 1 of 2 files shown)


.github/scripts/arm/build.sh
```diff
@@ -4,13 +4,14 @@
...
 [ -z "$2" ] && { echo "Missing 2nd argument: Build environment directory name"; exit 1; }
+[ -z "$3" ] && { echo "Missing 3rd argument: GHC version"; exit 1; }
 
 PGRST_GITHUB_COMMIT="$1"
 SCRIPT_DIR="$2"
 
 DOCKER_BUILD_DIR="$SCRIPT_DIR/docker-env"
-CABAL_VERSION="3.6.0.0"
-GHC_VERSION="9.2.4"
+CABAL_VERSION="recommended"
```
steve-chavez, line 13: This value is weird. Is its interpretation documented somewhere?

laurenceisla: Yes, according to ghcup, it's the default installation: https://www.haskell.org/ghcup/install/#which-versions-get-installed and it can be used as an installation tag https://www.haskell.org/ghcup/guide/#tags-and-shortcuts. 

Although, now that I check it, the `ubuntu-latest` runner image uses Cabal 3.10.1.0, which is the `latest` version in ghcup. I'll change it to `latest` to keep consistency between both builds.

.github/scripts/arm/build.sh
```diff
@@ -4,13 +4,14 @@
...
 [ -z "$2" ] && { echo "Missing 2nd argument: Build environment directory name"; exit 1; }
+[ -z "$3" ] && { echo "Missing 3rd argument: GHC version"; exit 1; }
 
 PGRST_GITHUB_COMMIT="$1"
 SCRIPT_DIR="$2"
 
 DOCKER_BUILD_DIR="$SCRIPT_DIR/docker-env"
-CABAL_VERSION="3.6.0.0"
-GHC_VERSION="9.2.4"
+CABAL_VERSION="latest"
```
steve-chavez, line 13:
```suggestion
# latest is a shortcut documented on https://www.haskell.org/ghcup/guide/#tags-and-shortcuts
CABAL_VERSION="latest"
```

Issue #2840, opened 2023-06-27
Non-null domain types that have a default value are coerced to null


imageck:
### Environment

* PostgreSQL version: PostgreSQL 15.1 (Ubuntu 15.1-1.pgdg20.04+1) on aarch64-unknown-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, 64-bit
* PostgREST version: Supabase doesn't disclose it anywhere (`Server` header reads `cloudflare`)
* Operating system: Linux

### Description of issue

Domain constraints should be respected but they aren't. `json_to_recordset()` fills remaining columns with nulls, thus violating the non-null constraint.

https://stackblitz.com/edit/stackblitz-starters-6sexes?file=pages%2Findex.js

imageck:
Note that:

1. Supabase doesn't disclose their version of PostgREST. The `Server` header is overridden by their cloud provider.
2. It is not possible to extract the error logs through the API.

The final query looks like this:
```sql
WITH pgrst_source AS (WITH pgrst_payload AS (SELECT $1 AS json_data), pgrst_body AS ( SELECT CASE WHEN json_typeof(json_data) = 'array' THEN json_data ELSE json_build_array(json_data) END AS val FROM pgrst_payload) INSERT INTO \"public\".\"people\"(\"name\") SELECT \"name\" FROM json_populate_recordset (null::\"public\".\"people\", (SELECT val FROM pgrst_body)) _  RETURNING \"public\".\"people\".*) SELECT '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header, coalesce((json_agg(_postgrest_t)->0)::text, 'null') AS body, nullif(current_setting('response.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status FROM (SELECT \"people\".* FROM \"pgrst_source\" AS \"people\"    ) _postgrest_t
```

steve-chavez:
```sql
CREATE DOMAIN uint AS int
  DEFAULT 0
  NOT NULL
  CHECK (VALUE >= 0);

CREATE TABLE people (
  id int GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
  name text,
  count uint
);
```

Using https://postgrest.org/en/stable/references/api/tables_views.html#bulk-insert-with-default-values.

```bash
[...]

PR #2843, opened 2023-06-30
fix: Prefer: missing=default with DOMAIN defaults

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/2840.

### DIFF
--- a/src/PostgREST/SchemaCache.hs
+++ b/src/PostgREST/SchemaCache.hs
@@ -571,8 +571,7 @@ tablesSqlQuery pgVer =
           c.relname::name AS table_name,
           a.attname::name AS column_name,
           d.description AS description,
-  |] <> columnDefault <>
-  [q|
+  |] <> columnDefault <> [q| AS column_default,
           not (a.attnotnull OR t.typtype = 'd' AND t.typnotnull) AS is_nullable,
               CASE
                   WHEN t.typtype = 'd' THEN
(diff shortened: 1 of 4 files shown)


src/PostgREST/SchemaCache.hs
```diff
@@ -760,19 +759,25 @@ tablesSqlQuery pgVer =
...
+          END|]
       | pgVer >= pgVersion100 = [q|
           CASE
+            WHEN t.typbasetype  != 0  THEN pg_get_expr(t.typdefaultbin, 0)
             WHEN a.attidentity = 'd' THEN format('nextval(%s)', quote_literal(seqsch.nspname || '.' || seqclass.relname))
             ELSE pg_get_expr(ad.adbin, ad.adrelid)::text
-          END AS column_default,|]
-      | otherwise  = "pg_get_expr(ad.adbin, ad.adrelid)::text as column_default,"
+          END|]
+      | otherwise  = [q|
```
steve-chavez, line 776: Don't get why this line is not covered in codecov. It runs when using `postgrest-with-postgresql-9.6 postgrest-test-spec`. I've double checked the test fails when removing it.

steve-chavez: Ah, code coverage only covers pg 15:

https://github.com/PostgREST/postgrest/blob/0a1564ba5aaa8175dcb738271967944652cf4ce9/.github/workflows/ci.yaml#L46-L51

steve-chavez: Noted this on https://github.com/PostgREST/postgrest/issues/1719#issuecomment-1615254184

PR #2852, opened 2023-07-06
ci: fix ARM docker containers not stopping immediately

### LINKED ISSUE #2851: Docker image takes 10s to stop
### Environment

* PostgreSQL version: supabase/postgres:15.1.0.90
* PostgREST version: image: postgrest/postgrest:latest
* Operating system: Mac

### Description of issue

(Expected behavior vs actual behavior)
* `docker compose down` usually runs quickly (1-2s) on most stateless containers but takes 10.5s on postgREST
* `docker compose restart` also faces the same problem
* Just kind of a slight annoyance in development, doesn't seem urgent, but perhaps is an underlying issue with the docker setup?

### Docker compose config to reproduce:
```# Usage
#   Start:          docker compose up
#   Stop:           docker compose down

version: "3.8"

services:
    rest:
        container_name: supabase-rest
        image: postgrest/postgrest:latest
        depends_on:
            db:
                # Disable this if you are using an external Postgres database
                condition: service_healthy
        restart: unless-stopped
        environment:
            PGRST_DB_URI: postgres://authenticator:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
            PGRST_DB_SCHEMAS: ${PGRST_DB_SCHEMAS}
            PGRST_DB_ANON_ROLE: anon
            PGRST_JWT_SECRET: ${JWT_SECRET}
            PGRST_DB_USE_LEGACY_GUCS: "false"

    db:
        container_name: supabase-db
        image: supabase/postgres:15.1.0.90
        healthcheck:
            test: pg_isready -U postgres -h localhost
            interval: 5s
            timeout: 5s
            retries: 10
        command:
            - postgres
            - -c
            - config_file=/etc/postgresql/postgresql.conf
            - -c
            - log_min_messages=fatal # prevents Realtime polling queries from appearing in logs
        restart: unless-stopped
        ports:
            # Pass down internal port because it's set dynamically by other services
            - ${POSTGRES_PORT}:${POSTGRES_PORT}
        environment:
            POSTGRES_HOST: /var/run/postgresql
            PGPORT: ${POSTGRES_PORT}
            POSTGRES_PORT: ${POSTGRES_PORT}
            PGPASSWORD: ${POSTGRES_PASSWORD}
            POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
            PGDATABASE: ${POSTGRES_DB}
            POSTGRES_DB: ${POSTGRES_DB}
```

laurenceisla:
Closes #2851

### DIFF
--- a/.github/scripts/arm/docker-env/Dockerfile
+++ b/.github/scripts/arm/docker-env/Dockerfile
@@ -13,6 +13,6 @@ EXPOSE 3000
 
 USER 1000
 
-# Use array form to avoid running the command using bash, which does not handle `SIGTERM` properly. 
+# Use the array form to avoid running the command using bash, which does not handle `SIGTERM` properly. 
 # See https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop 
 CMD ["postgrest"]
--- a/.github/scripts/arm/docker-env/Dockerfile
+++ b/.github/scripts/arm/docker-env/Dockerfile
@@ -13,4 +13,6 @@ EXPOSE 3000
 
 USER 1000
 
-CMD postgrest
+# Use array form to avoid running the command using bash, which does not handle `SIGTERM` properly. 
+# See https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop 
+CMD ["postgrest"]


.github/scripts/arm/docker-env/Dockerfile
```diff
@@ -13,4 +13,6 @@ EXPOSE 3000
 
 USER 1000
 
-CMD postgrest
+# Use array form to avoid running the command using bash, which does not handle `SIGTERM` properly. 
+# See https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop 
+CMD ["postgrest"]
```
steve-chavez, line 18:
```suggestion
# Use the array form to avoid running the command using bash, which does not handle `SIGTERM` properly. 
# See https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop 
CMD ["postgrest"]
```

steve-chavez: This was really surprising btw. Nice catch!

steve-chavez (approved):

Issue #2846, opened 2023-07-04
Null filtering on embedded resources throws `column does not exist` error when using `Prefer: count=exact`


zachblume:
### Environment

* PostgreSQL version: (if using docker, specify the image)
* Docker image: supabase/postgres:15.1.0.90
* PostgREST version: (if using docker, specify the image)
* image: postgrest/postgrest:latest (A little tricky here, I'm selfhosting supabase with the latest postgREST!)
* Operating system: Mac

### Description of issue

Ex.
```
-- Create the "people" table
CREATE TABLE people (
  id SERIAL PRIMARY KEY,
  name VARCHAR(255),
  email VARCHAR(255),
  -- Add other columns as needed
);

-- Create the "donations" table
CREATE TABLE donations (
  id SERIAL PRIMARY KEY,
  person_id INTEGER REFERENCES people(id),
  amount NUMERIC,
  -- Add other columns as needed
);

```

```
curl 'http://localhost:8000/rest/v1/people?select=*%2Cdonations%28*%29&or=%28and%28donations.not.is.null%29%29&donations.or=(and%28amount.eq.1%29)'
```

...responds with the correct items.

```
curl 'http://localhost:8000/rest/v1/people?select=*%2Cdonations%28*%29&or=%28and%28donations.not.is.null%29%29&donations.or=(and%28amount.eq.1%29)'
-H "Prefer: count=exact"
```

Responds:

`{code: "42703", details: null, hint: null, message: "column "people_donations_1" does not exist"}`

This seems to hold for count estimated or exact and other combinations or filtering against related tables.

So in effect we cannot count the rows alongside a filter against a joined table.

laurenceisla:
To simplify the error reproduction, a minimal example that causes this to fail is:

```bash
curl 'http://localhost:3000/people?select=*,donations(*)&donations=not.is.null' \
  -H "Prefer: count=exact"
```

Removing the null filtering from your example works OK, so the bug happens only in the case mentioned above:

```bash
curl 'http://localhost:3000/people?select=*,donations(*)&donations.or=(and(amount.eq.1))' \
  -H "Prefer: count=exact"
```

steve-chavez:
Note: here the behavior should be the same as https://github.com/PostgREST/postgrest/issues/2009. We need to add an EXISTS to the count query.

PR #2853, opened 2023-07-06
test: doctest for addNullEmbedFilters

steve-chavez:
Towards fixing #2846

### DIFF
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -598,7 +598,6 @@ resolveLogicTree ctx (Expr b op lts) = CoercibleExpr b op (map (resolveLogicTree
 
 resolveFilter :: ResolverContext -> Filter -> CoercibleFilter
 resolveFilter ctx (Filter fld opExpr) = CoercibleFilter{field=resolveQueryInputField ctx fld, opExpr=opExpr}
-resolveFilter _ (FilterNullEmbed isNot fieldName) = CoercibleFilterNullEmbed isNot fieldName
 
 -- Validates that spread embeds are only done on to-one relationships
 validateSpreadEmbeds :: ReadPlanTree -> Either ApiRequestError ReadPlanTree
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -82,6 +82,10 @@ import qualified PostgREST.ApiRequest.QueryParams as QueryParams
 
 import Protolude hiding (from)
 
+-- $setup
+-- Setup for doctests
+-- >>> import Data.Ranged.Ranges (fullRange)
+
 data WrappedReadPlan = WrappedReadPlan {
   wrReadPlan :: ReadPlanTree
 , wrTxMode   :: SQL.Mode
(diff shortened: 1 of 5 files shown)


src/PostgREST/Plan.hs
```diff
@@ -552,24 +556,92 @@ addRelatedOrders (Node rp@ReadPlan{order,from} forest) = do
...
--- It fails if operators other than is.null or not.is.null are used.
+-- | Searches for null filters on embeds, e.g. filter for `projects` on `GET /clients?select=*,projects(*)&projects=not.is.null`
+--
+-- Setup:
+--
+-- >>> let nullOp    = OpExpr True (Is TriNull)
+-- >>> let nonNullOp = OpExpr False (Is TriNull)
+-- >>> let notEqOp   = OpExpr True (Op OpNotEqual "val")
+-- >>> :{
+-- -- this resembles the `projects(*)` part on `/clients?select=*,projects(*)`
```
laurenceisla, line 567:
```suggestion
-- -- this represents the `projects(*)` part on `/clients?select=*,projects(*)`
```

src/PostgREST/Plan.hs
```diff
@@ -552,24 +556,92 @@ addRelatedOrders (Node rp@ReadPlan{order,from} forest) = do
...
+--         relJoinConds = [],
+--         relAlias = Nothing, relAggAlias = "clients_projects_1", relHint = Nothing, relJoinType = Nothing, relIsSpread = False, depth = 1
+--       },
+--       subForest = []
+--     }
+--   ]
+-- :}
+--
+-- >>> :{
+-- -- this will resemble the full URL `/clients?select=*,projects(*)&projects=not.is.null`
```
laurenceisla, line 587:
```suggestion
-- -- this represents the full URL `/clients?select=*,projects(*)&projects=not.is.null`
```

src/PostgREST/Plan.hs
```diff
@@ -552,24 +556,92 @@ addRelatedOrders (Node rp@ReadPlan{order,from} forest) = do
...
+--         )
+--       ],
+--       order = [], range_ = fullRange, relName = "clients", relToParent = Nothing, relJoinConds = [], relAlias = Nothing, relAggAlias = "", relHint = Nothing,
+--       relJoinType = Nothing, relIsSpread = False, depth = 0
+--     },
+--     subForest = subForst
+--   }
+-- :}
+--
+-- Don't do anything to the filter if there's no embedding (a subtree) on projects. Assume it's a normal filter.
```
laurenceisla, line 610:
```suggestion
-- It doesn't do anything to the filter if there's no embedding (a subtree) on projects. Assume it's a normal filter.
```

src/PostgREST/Plan.hs
```diff
@@ -552,24 +556,92 @@ addRelatedOrders (Node rp@ReadPlan{order,from} forest) = do
...
+--
+-- >>> ReadPlan.where_ . rootLabel <$> addNullEmbedFilters (readPlanTree nullOp [])
+-- Right [CoercibleStmnt (CoercibleFilter {field = CoercibleField {cfName = "projects", cfJsonPath = [], cfIRType = "", cfTransform = Nothing, cfDefault = Nothing}, opExpr = OpExpr True (Is TriNull)})]
+--
+-- If there's an embedding on projects, then change the filter to use the internal aggregate name (`clients_projects_1`) so the filter can succeed later.
```
laurenceisla, line 615:
```suggestion
-- If there's an embedding on projects, then it changes the filter to use the internal aggregate name (`clients_projects_1`) so the filter can succeed later.
```

src/PostgREST/Plan.hs
steve-chavez: Hm, the imperative word is shorter. I think we should keep it as it is.

steve-chavez: The hemingway app also always recommends the "active voice".

src/PostgREST/Plan.hs
steve-chavez: Ditto as above

laurenceisla (approved):

PR #2856, opened 2023-07-07
Add the -v/--version option to show the version information

laurenceisla:
Now that we can check the version from many sources, it makes sense to also allow it from the command line. This PR adds the `-v` / `--version` option and changes the following:

```bash
$ postgrest --version
Invalid option `--version'

...
```

To:

```bash
$ postgrest --version
PostgREST 11.1.0 (e752224)
```

### DIFF
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -80,13 +80,19 @@ readCLIShowHelp =
   where
     prefs = O.prefs $ O.showHelpOnError <> O.showHelpOnEmpty
     opts = O.info parser $ O.fullDesc <> progDesc
-    parser = O.helper <*> exampleParser <*> cliParser
+    parser = O.helper <*> versionParser <*> exampleParser <*> cliParser
 
     progDesc =
       O.progDesc $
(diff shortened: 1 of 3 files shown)


steve-chavez (approved):

src/PostgREST/CLI.hs
```diff
@@ -80,14 +80,20 @@ readCLIShowHelp =
...
         "PostgREST "
         <> BS.unpack prettyVersion
         <> " / create a REST API to an existing Postgres database"
 
+    versionParser =
+      O.infoOption ("PostgREST " <> BS.unpack prettyVersion) $
+        O.long "version"
+        <> O.short 'v'
+        <> O.help "Show the version information"
+
```
steve-chavez, line 96: To keep consistency with the other flags:

```suggestion
    versionFlag =
      O.infoOption ("PostgREST " <> BS.unpack prettyVersion) $
        O.long "version"
        <> O.short 'v'
        <> O.help "Show the version information"

```

Also this doesn't seem to do parsing.

PR #2867, opened 2023-07-22
fix: paramaters of type character and bit not ignoring length

### LINKED ISSUE #1586: postgrest: value too long for type character(1)
### Environment

* PostgreSQL version: 11.7
* PostgREST version: 7.0
* Operating system: Raspbian 10

### Description of issue
I have this table
 
CREATE TABLE api.user (
        id serial NOT NULL,
        name varchar(80) NOT NULL,
        pin char(4) NOT NULL,
        CONSTRAINT pk_user PRIMARY KEY (id)
);

and this function:
create or replace function functions.register_user(username varchar(80),
                                                 pin char(4))
returns void as
$$
insert into api.user values(DEFAULT, username, pin);
$$ language sql;

Within the psql client I can easily run for example 
'insert into api.user values(DEFAULT, "test", "0000");'

Whenever I try something similar with curl, I get the following response:
{"details":null,"code":"22001","message":"value too long for type character(1)","hint":null}

It works only if I POST 'pin' with at most 1 char

laurenceisla:
Closes #1586.

### DIFF
--- a/src/PostgREST/SchemaCache/Routine.hs
+++ b/src/PostgREST/SchemaCache/Routine.hs
@@ -69,10 +69,11 @@ instance JSON.ToJSON Routine where
     ]
 
 data RoutineParam = RoutineParam
-  { ppName :: Text
-  , ppType :: Text
-  , ppReq  :: Bool
-  , ppVar  :: Bool
+  { ppName          :: Text
+  , ppType          :: Text
+  , ppTypeMaxLength :: Text
+  , ppReq           :: Bool
+  , ppVar           :: Bool
   }
   deriving (Eq, Show, Ord, Generic, JSON.ToJSON)
 
--- a/test/spec/Feature/Query/RpcSpec.hs
+++ b/test/spec/Feature/Query/RpcSpec.hs
@@ -1447,3 +1447,15 @@ spec actualPgVersion =
               { matchStatus  = 400
               , matchHeaders = [matchContentTypeJson]
               }
+
+    -- https://github.com/PostgREST/postgrest/issues/1586#issuecomment-696345442
+    context "a proc with bit and char parameters" $ do
+      it "modifies the param type from character to character varying" $ do
+        get "/rpc/char_param?char_=abcdefg&char_arr={abc,abcdefg}" `shouldRespondWith`
+          [json| [{ "char_": "abcdefg", "char_arr": [ "abc", "abcdefg" ] }] |]
+          { matchHeaders = [matchContentTypeJson] }
+
+      it "modifies the param type from bit to bit varying" $ do
+        get "/rpc/bit_param?bit_=101010&bit_arr={101,101010}" `shouldRespondWith`
+          [json| [{ "bit_": "101010", "bit_arr": [ "101", "101010" ] }] |]
+          { matchHeaders = [matchContentTypeJson] }
--- a/test/spec/Feature/Query/RpcSpec.hs
+++ b/test/spec/Feature/Query/RpcSpec.hs
@@ -1451,11 +1451,21 @@ spec actualPgVersion =
     -- https://github.com/PostgREST/postgrest/issues/1586#issuecomment-696345442
     context "a proc with bit and char parameters" $ do
       it "modifies the param type from character to character varying" $ do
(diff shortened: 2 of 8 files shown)


laurenceisla:
According to [PostgreSQL docs on character types](https://www.postgresql.org/docs/current/datatype-character.html):

> If specified, the length must be greater than zero and cannot exceed 10485760. character without length specifier is equivalent to character(1)

But, the [docs for CREATE FUNCTION](https://www.postgresql.org/docs/15/sql-createfunction.html) say this:

> The full SQL type syntax is allowed for declaring a function's arguments and return value. However, parenthesized type modifiers (e.g., the precision field for type numeric) are discarded by CREATE FUNCTION. Thus for example CREATE FUNCTION foo (varchar(10)) ... is exactly the same as CREATE FUNCTION foo (varchar) ...

It made me think that `CREATE FUNCTION foo(character(10))` is the same as `CREATE FUNCTION foo(character)`, where `character` is of length 1 by default, as the docs on chars say. But that's not the case, because it allows to use parameters with a maximum length specified for a character which is the same of a varchar and text (10485760).

I couldn't find in the source code how they handle this change, so I casted both `character` and `bit` to their `character varying` and `bit varying` respectively. I'm not sure if it's a bit opinionated on my part.

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1447,3 +1447,15 @@ spec actualPgVersion =
...
               }
+
+    -- https://github.com/PostgREST/postgrest/issues/1586#issuecomment-696345442
+    context "a proc with bit and char parameters" $ do
+      it "modifies the param type from character to character varying" $ do
+        get "/rpc/char_param?char_=abcdefg&char_arr={abc,abcdefg}" `shouldRespondWith`
+          [json| [{ "char_": "abcdefg", "char_arr": [ "abc", "abcdefg" ] }] |]
+          { matchHeaders = [matchContentTypeJson] }
+
+      it "modifies the param type from bit to bit varying" $ do
```
steve-chavez, line 1458: How about a test that proves failure if the char exceeds the size?

steve-chavez:
> so I casted both character and bit to their character varying and bit varying respectively

If they fail when exceeding the length I guess it's find. (maybe the error message changes but it wouldnt be a dealbreaker)

src/PostgREST/SchemaCache/Routine.hs
```diff
@@ -69,10 +69,11 @@ instance JSON.ToJSON Routine where
     ]
 
 data RoutineParam = RoutineParam
-  { ppName :: Text
-  , ppType :: Text
-  , ppReq  :: Bool
-  , ppVar  :: Bool
+  { ppName          :: Text
+  , ppType          :: Text
+  , ppTypeMaxLength :: Text
```
steve-chavez, line 74: Is `ppType` now unused? Would it make sense to have a single `ppType` with the added length for these cases?

laurenceisla: No, `ppType` is still used on every other place (e.g. errors, OpenAPI, etc.) Only when building the Query it transforms e.g. `character` -> `character varying`.

The thing is, the length is ignored by PostgreSQL, it's not specified in the function definition, so no length can be added, it's the `character` -> `character varying` change that's done here.

test/spec/Feature/Query/RpcSpec.hs
laurenceisla: Added the test for the use case mentioned in the issue.

steve-chavez (approved):

laurenceisla:
Interestingly, only the PG14 io-test keeps failing. Could not find a reason why for now. Maybe a change needs to be done in the schema query.

steve-chavez:
@laurenceisla I don't see the pg 14 failure from CI. Can you link to the error?

laurenceisla:
@steve-chavez Ah, true. Strangely it failed twice in a row (1st and 2nd attempt) https://github.com/PostgREST/postgrest/actions/runs/5661060491/attempts/2?pr=2867, then the last ones run without problems.

Issue #1655, opened 2020-11-18
Return 404 instead of 406 for json object response with no rows?


wolfgangwalther:
Currently when making a request with `Accept: application/vnd.pgrst.object+json`, we get the following response, when `<> 1` rows are returned:

```json
{
    "details": "Results contain 0 rows, application/vnd.pgrst.object+json requires 1 row",
    "message": "JSON object requested, multiple (or no) rows returned"
}
```

This is status code 406.

With no rows returned, I would expect to get a 404 Not Found. Maybe we can return 404 for no rows, and 406 for multiple rows?

This would help a lot in differentiating those two cases. 404 (no rows) is something that I expect in my case and can handle on the client side. 406 (multiple rows) would mean something went terribly wrong.

steve-chavez:
Not sure if changing the status would be a good idea, specially when [406 Not Acceptable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) is the best semantic fit for a failed `Accept`.

Perhaps we could add a field(or use the `hint`) on the json error to make it easier for clients to detect the number of rows affected?

I've also looked at other REST mechanisms. I found this REST API: https://www.hl7.org/fhir/http.html#patch. It seems it responds 404 for zero rows and `412 Precondition Failed` for multiple rows. Perhaps we can have another Prefer that makes the operations stricter?

(Their implementation also has a `Prefer: return=OperationOutcome`, interesting to find another option besides `minimal/representation`)

wolfgangwalther:
> Not sure if changing the status would be a good idea, specially when [406 Not Acceptable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) is the best semantic fit for a failed `Accept`.

I think the underlying problem is, that we don't really have a concept of "requesting a single resource". We do have a concept of "requesting a list of resources", with a special case of n=1, where we can use a different output format. At least in terms of what `Accept` etc. means.

However, when using `Accept: application/vnd.pgrst.object+json`, I assume almost everyone using PostgREST to think about requesting a single resource.

In terms of Content-Negotiation and `Accept` headers, 406 is a semantically better response. In terms of requesting a single resource, it doesn't fit at all. In this case the 404/412 mentioned below are a much better fit.

> Perhaps we could add a field(or use the `hint`) on the json error to make it easier for clients to detect the number of rows affected?

[...]

PR #2874, opened 2023-07-25
Add the number of rows affected in hint field of response of SingularityError

taimoorzaeem:
Fixes issue #1655.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Note: data representations require Postgres 10 (Postgres 11 if using `IN` predicates); data representations are not implemented for RPC
  - #2647, Allow to verify the PostgREST version in SQL: `select distinct application_name from pg_stat_activity`. - @laurenceisla
  - #2856, Add the `--version` CLI option that prints the version information - @laurenceisla
+ - #1655, Add the number of rows affected to the `hint` field of response of `SingularityError`. - @taimoorzaeem
 
 ### Fixed
 
--- a/test/spec/Feature/Query/SingularSpec.hs
+++ b/test/spec/Feature/Query/SingularSpec.hs
@@ -69,7 +69,7 @@ spec =
             [("Prefer", "tx=commit"), singular]
             [json| { address: "zzz" } |]
           `shouldRespondWith`
-            [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":null}|]
+            [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":"4"}|]
             { matchStatus  = 406
             , matchHeaders = [ matchContentTypeSingular
                              , "Preference-Applied" <:> "tx=commit" ]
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Note: data representations require Postgres 10 (Postgres 11 if using `IN` predicates); data representations are not implemented for RPC
  - #2647, Allow to verify the PostgREST version in SQL: `select distinct application_name from pg_stat_activity`. - @laurenceisla
  - #2856, Add the `--version` CLI option that prints the version information - @laurenceisla
+ - #1655, Add the number of rows affected to the `hint` field of response of `SingularityError`. - @taimoorzaeem
```
laurenceisla, line 25:
```suggestion
 - #1655, Add the number of rows affected to the `hint` field of a singular response error. - @taimoorzaeem
```

`SingularityError` may confuse users, because it's an internal identifier in the code.

test/spec/Feature/Query/SingularSpec.hs
```diff
@@ -69,7 +69,7 @@ spec =
             [("Prefer", "tx=commit"), singular]
             [json| { address: "zzz" } |]
           `shouldRespondWith`
-            [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":null}|]
+            [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":"4"}|]
```
steve-chavez, line 72: Hm, just noticing that we already have the affected rows on `details`. @laurenceisla Perhaps they were added after #1655 was opened and we forgot to close?

test/spec/Feature/Query/SingularSpec.hs
```diff
@@ -85,7 +85,7 @@ spec =
             [("Prefer", "tx=commit"), ("Prefer", "return=representation"), singular]
             [json| { address: "zzz" } |]
           `shouldRespondWith`
-            [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":null}|]
+            [json|{"details":"Results contain 4 rows, application/vnd.pgrst.object+json requires 1 row","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":"4"}|]
```
steve-chavez, line 88: @taimoorzaeem Besides the hint (which is likely unnecessary now), we can make an improvement to the error message:

```suggestion
            [json|{"details":"The result contains 4 rows","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116"}|]
```

The reasoning is that I've been noticing that users ignore the error message when they see the `application/vnd.pgrst.object+json`. This happens when they use PostgREST through client libraries, like https://github.com/supabase/postgrest-js.

steve-chavez: Also technically the error message is not accurate since it should refer to json array elements or array length and not "rows". It could be reworded.

taimoorzaeem: @steve-chavez "rows" is more consistent with other error messages and also since "rows" refer to the rows returned from the database so it is accurate IMO. Perhaps maybe no need to reword that.

taimoorzaeem:
Because we are making changes in a different direction now, I am close this pull request.

test/spec/Feature/Query/SingularSpec.hs
laurenceisla: Hmm, I understood that it was already present in the details, as shown in the issue's OP, but that an easy access to the number of rows was needed (as an alternative to the `404`, `412`), hence only the number of rows in the hint. The number of rows is present in the details even in version 7.0.0, which was launched before the issue was opened,  see [this diff from 2017](https://github.com/PostgREST/postgrest/commit/104a7ed4fa247e30cd5a3e150ebdf3d7515b1953#diff-328c2f3103b25df9c32be37811eabc32ea0f39cea4f860b33651b3d470adf974R46-R55).

Issue #2176, opened 2022-02-20
Raise sqlstate with `message` does not return `message` (used with custom http status codes).


robertsosinski:
* PostgreSQL version: 13.1
* PostgREST version: 9.0.0
* Operating system: Docker on Win 10

When I raise an error with `sqlstate` I specify a `message`:

```
RAISE sqlstate 'PT402' using
  message = 'Payment Required',
  detail = 'Quota exceeded',
  hint = 'Upgrade your plan';
```

However, the `message` does not come along with the response:

```
{
    "hint":"Upgrade your plan",
    "details":"Quota exceeded"
}
```

If I raise with a built in error, the message comes through though:

```
RAISE  insufficient_privilege using 
  detail = 'Quota exceeded',
  hint = 'Upgrade your plan';
```

You can see here:

```
{
    "hint": "Upgrade your plan",
    "details": "Quota exceeded",
    "code": "42501",
    "message": "insufficient_privilege"
}
```

Would be great to get the `message` specified in the `raise` statement when specifying custom http status codes.

Thanks!

steve-chavez:
> RAISE sqlstate 'PT402' using
>   message = 'Payment Required',
>   detail = 'Quota exceeded',
>   hint = 'Upgrade your plan';

IIRC the `message` there is passed as the status message. So if you do:

```sql
 RAISE sqlstate 'PT452' using
   message = 'My Special Status',
   detail = 'Quota exceeded',
   hint = 'Upgrade your plan';
```

It would respond with:

```
HTTP/1.1 452 My Special Status 
```

I thought this could be used for defining new status codes/messages but perhaps the most useful behavior is to auto-generate the status messages for known status codes(similar to [response.status](https://postgrest.org/en/stable/api.html#setting-response-status-code)) and leave them blank for unknown status codes.

---

Relevant code:

https://github.com/PostgREST/postgrest/blob/dcf7ade5bc70a31fe0e083e1527b74125a94aeda/src/PostgREST/Error.hs#L172-L182

robertsosinski:
That makes sense, I just took a look and indeed the `message` field is visible with the HTTP code. I think this works, but if we could also get the same `message` and `code` in the JSON response body that built in errors get, that would be more consistent.  The response would look like:

plpgsql
```
 RAISE sqlstate 'PT452' using
   message = 'My Special Status',
   detail = 'Quota exceeded',
   hint = 'Upgrade your plan';
```

response
```
[...]

PR #2876, opened 2023-07-26
Remove hint field and improve details field of the singular error response.

taimoorzaeem:
Made changes from the discussion on the pull request #2874.
This should close issue #1655.


src/PostgREST/Error.hs
```diff
@@ -530,8 +530,7 @@ instance JSON.ToJSON Error where
   toJSON (SingularityError n) = JSON.object [
     "code"    .= ApiRequestErrorCode16,
     "message" .= ("JSON object requested, multiple (or no) rows returned" :: Text),
-    "details" .= T.unwords ["Results contain", show n, "rows,", T.decodeUtf8 (MediaType.toMime MTSingularJSON), "requires 1 row"],
-    "hint"    .= JSON.Null]
```
steve-chavez, line 534: For consistency, we should keep the `hint` as null. Check a previous discussion about this here: https://github.com/PostgREST/postgrest/issues/2176#issuecomment-1049982360

Once that's corrected, it'll be good to merge.

PR #2884, opened 2023-07-31
Add Preference-Applied header in response for Prefer: return=representation/headers-only/minimal

taimoorzaeem:
Fixes issue #740.

PR #2888, opened 2023-08-02
refactor: Response rm Paths_postgrest dependency

steve-chavez:
This allows using doctests on the Response.hs module. Should solve the error mentioned on https://github.com/PostgREST/postgrest/pull/2884#discussion_r1282393438.


laurenceisla (approved):
Yep, that does it. Tested a simple doctest without problems.

PR #2889, opened 2023-08-03
fix: error when a function returns RECORD or SETOF RECORD

### LINKED ISSUE #2881: v11.0.0:  RPC with function returning setof record no longer works
## Problem

The following:

```sql
create or replace function setof_record() returns setof record as $$
  select * from projects;
$$ language sql;
```

Fails on RPC:

```
curl localhost:3000/rpc/setof_record
{"code":"42601","details":null,"hint":null,"message":"a column definition list is required for functions returning \"record\""}
```

On v10.2.0, it worked:

```
curl localhost:3000/rpc/setof_record
 {"id":3,"name":"IOS","client_id":2}, 
 {"id":5,"name":"Orphan","client_id":null}]
```

Because it was treated as a scalar, i.e. it was called like `select setof_record()`.

## Solution

Treat `setof record` as scalar again. However this has the drawback of not being able to `select` a column:

```
curl 'localhost:3000/rpc/setof_record?select=id'
{"code":"42703","details":null,"hint":null,"message":"column setof_record.id does not exist"}
```

We should document this capability plus the limitation.

laurenceisla:
Fixes #2881

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2862, Fix null filtering on embedded resource when using a column name equal to the relation name - @steve-chavez
  - #1586, Fix function parameters of type character and bit not ignoring length - @laurenceisla
    + Fixes the error "value too long for type character(1)" when the char length of the parameter was bigger than one.
+ - #2881, Fix error when a function returns `RECORD` or `SET OF RECORD` - @laurenceisla
 
 ## [11.1.0] - 2023-06-07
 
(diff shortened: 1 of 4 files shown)


laurenceisla:
This gives an error for PostgreSQL versions 9.6 and 10 when using `RECORD`. It would require an exception only for those versions so maybe a not so trivial restructuring is needed here, which may not be worth it (since these are unsupported versions).

https://github.com/PostgREST/postgrest/blob/0fce7ca3618b616454450a288ddb53596d6f4671/src/PostgREST/Query/SqlFragment.hs#L199-L200

PR #2894, opened 2023-08-07
feat: add optional nulls=stripped parameter for mediatypes applicatio…

taimoorzaeem:
…n/vnd.pgrst.array+json and application/vnd.pgrst.object+json

Fixes issue #1601.

@steve-chavez The codecov is not working for me yet. Let me know if any warnings show up. Also, I am not exactly sure what combinations of testcases need to be written for this feature so please let me know if I need to add more.

### DIFF
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -19,7 +19,8 @@ import Protolude
 -- | Enumeration of currently supported media types
 data MediaType
   = MTApplicationJSON
-  | MTSingularJSON
+  | MTArrayJSONStrip
+  | MTSingularJSON Bool
   | MTGeoJSON
   | MTTextCSV
   | MTTextPlain
@@ -33,19 +34,20 @@ data MediaType
   | MTPlan MediaType MTPlanFormat [MTPlanOption]
   deriving Show
 instance Eq MediaType where
-  MTApplicationJSON == MTApplicationJSON = True
-  MTSingularJSON    == MTSingularJSON    = True
-  MTGeoJSON         == MTGeoJSON         = True
-  MTTextCSV         == MTTextCSV         = True
-  MTTextPlain       == MTTextPlain       = True
-  MTTextXML         == MTTextXML         = True
-  MTOpenAPI         == MTOpenAPI         = True
-  MTUrlEncoded      == MTUrlEncoded      = True
-  MTOctetStream     == MTOctetStream     = True
-  MTAny             == MTAny             = True
-  MTOther x         == MTOther y         = x == y
-  MTPlan{}          == MTPlan{}          = True
-  _                 == _                 = False
+  MTApplicationJSON    == MTApplicationJSON = True
+  MTArrayJSONStrip     == MTArrayJSONStrip  = True
+  MTSingularJSON x     == MTSingularJSON y  = x == y
+  MTGeoJSON            == MTGeoJSON         = True
+  MTTextCSV            == MTTextCSV         = True
+  MTTextPlain          == MTTextPlain       = True
+  MTTextXML            == MTTextXML         = True
+  MTOpenAPI            == MTOpenAPI         = True
+  MTUrlEncoded         == MTUrlEncoded      = True
+  MTOctetStream        == MTOctetStream     = True
+  MTAny                == MTAny             = True
+  MTOther x            == MTOther y         = x == y
+  MTPlan{}             == MTPlan{}          = True
+  _                    == _                 = False
 
 data MTPlanOption
   = PlanAnalyze | PlanVerbose | PlanSettings | PlanBuffers | PlanWAL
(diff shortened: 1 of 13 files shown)


src/PostgREST/MediaType.hs
```diff
@@ -19,7 +19,9 @@ import Protolude
 -- | Enumeration of currently supported media types
 data MediaType
   = MTApplicationJSON
+  | MTArrayJSONStrip
   | MTSingularJSON
+  | MTSingularJSONStrip
```
steve-chavez, line 24: How about turning the `nulls=stripped` to a parameter of the existing media types?

```suggestion
  | MTSingularJSON Bool
```

src/PostgREST/MediaType.hs
```diff
@@ -33,19 +35,21 @@ data MediaType
...
+  MTTextCSV            == MTTextCSV           = True
+  MTTextPlain          == MTTextPlain         = True
+  MTTextXML            == MTTextXML           = True
+  MTOpenAPI            == MTOpenAPI           = True
+  MTUrlEncoded         == MTUrlEncoded        = True
+  MTOctetStream        == MTOctetStream       = True
+  MTAny                == MTAny               = True
+  MTOther x            == MTOther y           = x == y
+  MTPlan{}             == MTPlan{}            = True
+  _                    == _                   = False
```
steve-chavez, line 52: I realize this is quite ugly. I think the whole media type handling could be refactored to use https://hackage.haskell.org/package/http-media-0.8.0.0/docs/Network-HTTP-Media.html.

Then we could remove this part.

src/PostgREST/MediaType.hs
```diff
@@ -106,26 +112,44 @@ toMimePlanFormat PlanText = "text"
 --
 -- >>> decodeMediaType "application/vnd.pgrst.plan+json;for=\"text/csv\""
 -- MTPlan MTTextCSV PlanJSON []
+--
+-- >>> decodeMediaType "application/json;nulls=stripped"
+-- MTArrayJSONStrip
```
steve-chavez, line 117: As mentioned on https://github.com/PostgREST/postgrest/issues/1601#issuecomment-703323701, we cannot add parameters to existing media types (`application/json`). At least not without an RFC. 

So `application/json` should not accept nulls=stripped.

steve-chavez:
@taimoorzaeem The end result looks good :+1:. But the code needs some refactoring. I've added some comments.

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -204,6 +211,17 @@ asJsonF rout
...
+asJsonArrayStripF rout
+  | returnsSingleComposite    = "coalesce(json_strip_nulls(json_agg(_postgrest_t)->0), 'null')"
+  | returnsScalar             = "coalesce(json_strip_nulls(json_agg(_postgrest_t.pgrst_scalar)->0), 'null')"
+  | returnsSetOfScalar        = "coalesce(json_strip_nulls(json_agg(_postgrest_t.pgrst_scalar)), '[]')"
+  | otherwise                 = "coalesce(json_strip_nulls(json_agg(_postgrest_t)), '[]')"
+  where
+    (returnsSingleComposite, returnsScalar, returnsSetOfScalar) = case rout of
+      Just r  -> (funcReturnsSingleComposite r, funcReturnsScalar r, funcReturnsSetOfScalar r)
+      Nothing -> (False, False, False)
+
```
steve-chavez, line 224: Ideally we wouldn't need these 2 new functions. We could just add a Bool parameter to the existing ones to add `json_strip_nulls`.

src/PostgREST/MediaType.hs
taimoorzaeem: Yes, that is what I thought at first, but we would still need `MTArrayJSONStrip` for `Accept: application/vnd.pgrst.array+json;nulls=stripped`. Additionally, with `MTSingularJSON Bool`, we would have to make changes at all the places where `MTSingularJSON` is used, for example in `OpenAPI.hs` which I don't understand yet:
```haskell
./Response/OpenAPI.hs:353:  & produces ?~ makeMimeList [MTApplicationJSON, MTSingularJSON <True/False>]
./Response/OpenAPI.hs:409:  & produces .~ makeMimeList [MTApplicationJSON, MTSingularJSON <True/False>, MTTextCSV]
./Response/OpenAPI.hs:410:  & consumes .~ makeMimeList [MTApplicationJSON, MTSingularJSON <True/False>, MTTextCSV]
```
However, still I think it is much simpler. Let me push the changes.

src/PostgREST/MediaType.hs
taimoorzaeem: Perhaps, maybe we should move this refactoring to a new separate issue to work on?

test/spec/Feature/Query/NullsStrip.hs
```diff
@@ -0,0 +1,80 @@
...
+
+spec :: SpecWith ((), Application)
+spec =
+  describe "Stripping null values from JSON response" $ do
+    let arrayStrip = ("Accept", "application/vnd.pgrst.array+json;nulls=stripped")
+    let singularStrip = ("Accept", "application/vnd.pgrst.object+json;nulls=stripped")
+
+    context "strip nulls from response" $ do
+      it "strips nulls when Accept: application/vnd.pgrst.array+json;nulls=stripped" $
+        request methodGet  "/organizations?select=*"
```
steve-chavez, line 21: Let's also add a test for `POST` and `PATCH`. You can test the result by adding `Prefer: return=representation`. Also, all writes are rolled back in the test suite ([db-tx-end=rollback](https://postgrest.org/en/stable/references/configuration.html#db-tx-end)) so you don't have to worry about the other tests breaking.

test/spec/Feature/Query/NullsStrip.hs
```diff
@@ -0,0 +1,80 @@
...
+          , matchHeaders = [matchContentTypeSingular]
+          }
+
+    context "strip nulls from response even if explicitly selected" $ do
+      it "strips nulls when Accept: application/vnd.pgrst.array+json;nulls=stripped" $
+        request methodGet  "/organizations?select=id,referee,auditor"
+          [arrayStrip]
+          ""
+          `shouldRespondWith`
+          [json|[{"id":1},{"id":2},{"id":3,"referee":1,"auditor":2},{"id":4,"referee":1,"auditor":2},{"id":5,"referee":3,"auditor":4},{"id":6,"referee":3,"auditor":4}]|]
```
steve-chavez, line 56: Nice test here :+1:. Definitely the behavior users would want.

steve-chavez:
@taimoorzaeem Just found a bug here. So currently we accept doing:

```
Accept: application/vnd.pgrst.object
Accept: application/vnd.pgrst.object+json
```

But the array fails on

```
Accept: application/vnd.pgrst.array
```

It always requires the `+json` suffix, which is inconvenient.

taimoorzaeem:
@steve-chavez Oh, I can't believe how I missed this. It's an easy fix, let me send a PR.

src/PostgREST/MediaType.hs
steve-chavez: @taimoorzaeem Could you open that one :pray: ?

taimoorzaeem: Yeah sure!

PR #2900, opened 2023-08-11
feat: add db-pool-automatic-recovery configuration to disable connection retrying

### LINKED ISSUE #1614: Does `postgrest` have a "fail fast" mode?
(I'm a new user of PostgREST so I may not have learned all the possible options/features but have tried my best.)

### Environment

* PostgreSQL version: Docker image [`postgres` 13](https://github.com/docker-library/postgres/blob/8d281112a872380578e37a92a698ce59d66473d7/13/Dockerfile)
* PostgREST version: 7.0.1, downloaded at [this link](https://github.com/PostgREST/postgrest/releases/download/v7.0.1/postgrest-v7.0.1-linux-x64-static.tar.xz).
* Operating system: Ubuntu 18.04

### Description of question

**Does `postgrest` have a _"fail fast"_ mode instead of retrying by itself?** Specifically, I'm hoping that if `postgrest` cannot connect to the database with the `db-uri`, it would exit with an error code instead of retrying to connect over and over again.

Here is **more context**:

- In my project environment, our PostgreSQL database is a Docker container that does **not** publish its service at a pre-determined port. Instead, it publishes the service at a **random port** and `Registrator` registers the service to `Consul`.
- When the machine is restarted thus the PostgreSQL container is restarted, the published port is changed, too. But with the help of `Consul`, we don't have to hard-code any specific host address or port in our code.

However, it seems that `postgres` only reads the configuration `db-uri` once and keeps using it for connection. The PR ["Enable config file re-reading with SIGUSR2" (and the related issues)](https://github.com/PostgREST/postgrest/pull/1544) says the change to `db-uri` requires a restart of `postgrest`.

Besides, it seems that `postgrest` automatically keeps retrying to connect to the database at `db-uri` but doesn't have an option to exit with an error code.

**But what I want to implement is**:

- I want to start `postgrest` as a `systemd` service.
- It is possible that, when `postgrest` service is started due to restarting the machine, the PostgreSQL container is not fully up yet (or at least not re-registered to `Consul` so `Consul` may still have the out-dated address and port of the previous PostgreSQL container).
- So, if `postgrest` fails to connect the database using whatever is in its current configuration file, **I hope `postgrest` could exit with an error code that can be captured by `systemd` so `systemd` would automatically restart it.**
- Unlike what is documented in [Daemonizing](http://postgrest.org/en/v7.0.0/admin.html#daemonizing), I will wrap up the call of `postgrest` with a shell script that obtains from `Consul` the latest address and port of the PostgreSQL container, updates the `postgrest` configuration file, and then starts it.

So I'm wondering if `postgrest` already implements this "fail fast" behavior or not. Any help/suggestions/thoughts would be appreciated!

taimoorzaeem:
This should close #1614.

@steve-chavez I haven't written any tests because I am not sure exactly how configurations are being tested here. And so Codecov is complaining. However, I have tested it locally by giving it an invalid `db-uri` and when the `pg server` is not available for which recovery disables successfully. Please review.


taimoorzaeem:
@steve-chavez Also, I am thinking maybe we should call the configuration option `disable-retrying` instead of `disable-recovery`. It seems more consistent with the internal identifiers and as well as the error messages.

steve-chavez:
@taimoorzaeem Since [recovery is a feature of the connection pool](https://postgrest.org/en/stable/references/connection_pool.html#automatic-recovery), how about calling it `db-pool-automatic-recovery`? It would be true by default.

Note that all our [pool configs have the db-pool prefix](https://postgrest.org/en/stable/references/configuration.html#db-pool).

src/PostgREST/AppState.hs
```diff
@@ -281,8 +281,9 @@ internalConnectionWorker appState = work
           -- Fatal error when connecting
           logWithZTime appState reason >> killThread (getMainThreadId appState)
         NotConnected ->
-          -- Unreachable because establishConnection will keep trying to connect
-          return ()
+          -- Unreachable because establishConnection will keep trying to connect, unless disable-recovery is turned on
+          when configDisableRecovery
+              $ logWithZTime appState "Retrying disabled, killing thread..." >> killThread (getMainThreadId appState)
```
steve-chavez, line 286: Mostly to be consistent with the [docs](https://postgrest.org/en/stable/references/connection_pool.html#automatic-recovery). Also I think we don't need the `"..."`, which "convey" that we're going to keep working and in this case we die.

```suggestion
              $ logWithZTime appState "Automatic recovery disabled, exiting." >> killThread (getMainThreadId appState)
```

test/io/db_config.sql
```diff
@@ -35,6 +35,7 @@ ALTER ROLE db_config_authenticator SET pgrst.server_port = 'ignored';
 ALTER ROLE db_config_authenticator SET pgrst.server_unix_socket = 'ignored';
 ALTER ROLE db_config_authenticator SET pgrst.server_unix_socket_mode = 'ignored';
 ALTER ROLE db_config_authenticator SET pgrst.admin_server_port = 'ignored';
+ALTER ROLE db_config_authenticator SET pgrst.disable_recovery = 'true';
```
steve-chavez, line 38: For the [in-db configs](https://postgrest.org/en/stable/references/configuration.html#in-database-configuration) to work, we need to add them here: 

https://github.com/PostgREST/postgrest/blob/c820efb64a2839c5cb7a6a82b9145c3d7065474c/src/PostgREST/Config/Database.hs#L42-L66

That being said, I don't think we need to have an in-db config for this. It's a bit like changing the [db-uri](https://postgrest.org/en/stable/references/configuration.html#db-uri) (doesn't have an in-db config), which has the potential to kill the process.

steve-chavez:
> However, I have tested it locally by giving it an invalid db-uri and when the pg server is not available for which recovery disables successfully.

@taimoorzaeem Hm, I would expect the disable to work in this case:

```
$ PGRST_DB_POOL_AUTOMATIC_RECOVERY="false" postgrest-with-postgresql-15 postgrest-run

postgrest-with-postgresql-15: You can connect with: psql 'postgres:///postgres?host=/run/user/1000/postgrest/postgrest-with-postgresql-15-OV7/socket' -U postgres
postgrest-with-postgresql-15: You can tail the logs with: tail -f /run/user/1000/postgrest/postgrest-with-postgresql-15-OV7/db.log
14/Aug/2023:14:33:05 -0500: Starting PostgREST 11.2.0 (bd5e361)...
```

```
$ psql 'postgres:///postgres?host=/run/user/1000/postgrest/postgrest-with-postgresql-15-OV7/socket' -U postgres

select pg_terminate_backend(pid) from pg_stat_activity  where application_name iLIKE '%postgrest%';
 pg_terminate_backend 
----------------------
 t
```

But it still recovers:

```
FATAL:  terminating connection due to administrator command
14/Aug/2023:14:36:04 -0500: Retrying listening for notifications on the pgrst channel..
14/Aug/2023:14:36:04 -0500: Starting PostgREST 11.2.0 (bd5e361)...
```

For this I think we can catch the exception here and also kill the main thread:

https://github.com/PostgREST/postgrest/blob/87d6a0d0feaab6a85aaa5ba7ec6c0e9f97855fe4/src/PostgREST/AppState.hs#L144-L150

---

Is the env var not getting applied? I've tried:

```
$ PGRST_DB_POOL_AUTOMATIC_RECOVERY="false" postgrest-run -- --dump-config | grep automatic -c
0
 ```


---

The new config is missing from:

https://github.com/PostgREST/postgrest/blob/87d6a0d0feaab6a85aaa5ba7ec6c0e9f97855fe4/src/PostgREST/CLI.hs#L124-L166

---

> I haven't written any tests because I am not sure exactly how configurations are being tested here. And so Codecov is complaining. 

Here's an example of how configs are modified per io test:

https://github.com/PostgREST/postgrest/blob/87d6a0d0feaab6a85aaa5ba7ec6c0e9f97855fe4/test/io/test_io.py#L1056-L1061

Here's an example that catches the exit code:

https://github.com/PostgREST/postgrest/blob/87d6a0d0feaab6a85aaa5ba7ec6c0e9f97855fe4/test/io/test_io.py#L69-L75

Might be possible to have an io test based on those

**Comment by taimoorzaeem** [comment]
> For this I think we can catch the exception here and also kill the main thread:
> 
> https://github.com/PostgREST/postgrest/blob/87d6a0d0feaab6a85aaa5ba7ec6c0e9f97855fe4/src/PostgREST/AppState.hs#L144-L150

@steve-chavez This didn't work. Alternatively, I handled that in `listener` function.

**Comment by laurenceisla** [comment]
The static build is failing, maybe related to #2905? Rebasing with the main branch should fix this.

**Comment by taimoorzaeem** [comment]
@steve-chavez @laurenceisla I was thinking of adding a test case for `select pg_terminate_backend(pid) from pg_stat_activity  where application_name iLIKE '%postgrest%';`. Can we make `rpc` calls from IO tests? I am thinking of adding the above sql query to a stored procedure and then make a rpc call to test the functionality.

**Comment by steve-chavez** [comment]
@taimoorzaeem Yes, that's possible. There are some examples on `test_io.py`:

https://github.com/PostgREST/postgrest/blob/32b77cae6c4227216b193f6834304e90de3bd7c3/test/io/fixtures.sql#L118-L123

https://github.com/PostgREST/postgrest/blob/32b77cae6c4227216b193f6834304e90de3bd7c3/test/io/test_io.py#L853-L857

**Comment by taimoorzaeem** [comment]
@steve-chavez I have tried doing the above in the latest push, however calling such an `rpc` immediately kills postgrest and an exception is thrown, which the code coverage does not detect.

File: test/io/test_io.py [hunk 3]
```diff
@@ -1063,3 +1063,33 @@ def test_succeed_w_role_having_superuser_settings(defaultenv):
...
+    env = {
+        **defaultenv,
+        "PGRST_DB_POOL_AUTOMATIC_RECOVERY": "false",
+    }
+
+    try:
+        with run(env=env) as postgrest:
+            postgrest.session.get("/rpc/terminate_pgrst")
+    except ConnectionError:
+        return True
```
laurenceisla, line 1095: This test does not fail on the main branch when it should. I think verifying if postgrest stopped running is best:

```suggestion
    with run(env=env) as postgrest:
        postgrest.session.post("/rpc/terminate_pgrst")

        exitCode = wait_until_exit(postgrest)
        assert exitCode == 1
```

Still, doing the above does not close the connection and the test fails. It gives a  "Database client error. Retrying the connection." message and reloads PostgREST afterwards.  This *could* be considered "Automatic recovery", so maybe this shouldn't be happening when `db-pool-automatic-recovery=true`? If so, I think the thread must die here too:

https://github.com/PostgREST/postgrest/blob/531a183b44b36614224fda432335cdaa356b4a0a/src/PostgREST/App.hs#L122-L125

test/spec/fixtures/schema.sql
```diff
@@ -3368,3 +3368,7 @@ $$ language sql;
 create function returns_setof_record_params(id int, name text) returns setof record as $$
 select * from projects p where p.id >= $1 and p.name like $2;
 $$ language sql;
+
+create function terminate_pgrst() returns setof record as $$
+select pg_terminate_backend(pid) from pg_stat_activity where application_name iLIKE '%postgrest%';
+$$ language sql;
```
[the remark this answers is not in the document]
laurenceisla: Clever usage of `pg_terminate_backend` btw, TIL.

test/io/test_io.py
taimoorzaeem: > This _could_ be considered "Automatic recovery", so maybe this shouldn't be happening when `db-pool-automatic-recovery=true`? If so, I think the thread must die here too:

@laurenceisla I think you mean when `db-pool-automatic-recovery=false`. Yes, this shouldn't be happening. However, testing locally, the `terminate_pgrst()` terminates the postgrest without retries. Maybe there is something specific related to IO tests that we are missing?

taimoorzaeem: > select terminate_pgrst();

Yes, this works for me too. This is what I meant when I wrote "testing locally".

> Does using the curl command stop PostgREST on your end?

No. It reloads the connection.

laurenceisla: > No. It reloads the connection.

Yes, this happens with IO tests too, because they do the same request to PostgREST and reload the connection after that.

You're right that the IO tests are missing something, in this case a direct PostgreSQL execution. Luckily Python allows executing commands directly using `os.sytem("command")`. Then we can use `psql` to execute the SQL command like this:

```python
os.system(f'psql -d {defaultenv["PGDATABASE"]} -U {defaultenv["PGUSER"]} -h {defaultenv["PGHOST"]} --set ON_ERROR_STOP=1 -a -c "SELECT terminate_pgrst()"')
```

I replaced this in my original suggestion. This would solve the reloading when closing the connections outside PostgREST.

---

Now, there's the debate with reloading while doing a request. If we consider that it should not reload then the change must be done where I specified earlier.

taimoorzaeem: > ```os.system(f'psql -d {defaultenv["PGDATABASE"]} -U {defaultenv["PGUSER"]} -h {defaultenv["PGHOST"]} --set ON_ERROR_STOP=1 -a -c "SELECT terminate_pgrst()"')```

The above worked smoothly.

```haskell
if configDbPoolAutomaticRecovery appConf
  then when (Response.isServiceUnavailable response) connWorker
else when (Response.isServiceUnavailable response) (killThread $ AppState.getMainThreadId appState)
```
For the `rpc` call, I tried above as you suggested earlier, however this didn't work either. It throws a `ConnectionError` exception and then reloads PostgREST.

laurenceisla: > For the rpc call, I tried above as you suggested earlier, however this didn't work either. It throws a ConnectionError exception and then reloads PostgREST.

Hmmm... Yes, now that I see it, the request is not finished there, thus the error. The try/catch that you did before may be needed here. 

However, now that I see it more carefully, I think we could leave it as is (not modifying `App.hs`), because what this PR is looking for is to stop PostgREST when the DB connection is closed from outside. If the connection is really lost, the `connWorker` will fail with the `NotConnected` status anyways, so this extra verification is not needed here. In summary, closing PostgreSQL from within PostgREST is a special case that is not contemplated in the issue.

laurenceisla (approved):

steve-chavez:
@taimoorzaeem Nice work! How about documenting this new config on:

- https://postgrest.org/en/stable/references/connection_pool.html#automatic-recovery
- https://postgrest.org/en/stable/references/configuration.html#list-of-parameters

Repo at https://github.com/PostgREST/postgrest-docs.

taimoorzaeem:
> @taimoorzaeem Nice work! How about documenting this new config on:

Sure! I would love to. :)

taimoorzaeem:
@yaobinwen-mvs Cheers! Most likely, we will have this feature in our next release which would be `PostgREST 11.3.0`.

yaobinwen-mvs:
@taimoorzaeem Cool! Thanks! Do you know when the new version would be released? (I looked around in the repository but didn't seem to find any document about the release schedule.)

steve-chavez:
@yaobinwen-mvs We'll make a new release once we fix https://github.com/PostgREST/postgrest/issues/2846, which should be completed in the next few weeks.

PR #2906, opened 2023-08-17
Fix schema reloading via notice not working on Windows

### LINKED ISSUE #2524: Schema reloading via notice not working(Windows)
### Environment

* PostgreSQL version: V12.10
* PostgREST version: 10.0.0
* Operating system: Windows

### Description of issue

When I make schema changes and execute a notice to pgrst I expect the schema to reload an the new function to work. However it remains not working until I stop and restart PostgREST.

For any new function added to my api schema and after a notice is raised I still get the below error

"code": "PGRST202",
    "details": null,
    "hint": "If a new function was created in the database with this name and parameters, try reloading the schema cache.",
    "message": "Could not find the api.test_funct2(srid, x1, x2, y1, y2) function or the api.test_funct2 function with a single unnamed json or jsonb parameter in the schema cache"

laurenceisla:
Attempt to fix #2524

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -28,7 +28,8 @@ import qualified Data.ByteString.Char8      as BS
 import qualified Data.ByteString.Lazy       as LBS
 import           Data.Either.Combinators    (whenLeft)
 import qualified Data.Text.Encoding         as T
-import           Hasql.Connection           (acquire)
+import           Hasql.Connection           (acquire, Connection, withLibPQConnection)
+import qualified Database.PostgreSQL.LibPQ as PQ
 import qualified Hasql.Notifications        as SQL
 import qualified Hasql.Pool                 as SQL
 import qualified Hasql.Session              as SQL
@@ -427,7 +428,7 @@ listener appState = do
         logWithZTime appState $ "Listening for notifications on the " <> dbChannel <> " channel"
         putIsListenerOn appState True
         SQL.listen db $ SQL.toPgIdentifier dbChannel
-        SQL.waitForNotifications handleNotification db
+        waitForNotifications_ handleNotification db
       _ ->
         die $ "Could not listen for notifications on the " <> dbChannel <> " channel"
   where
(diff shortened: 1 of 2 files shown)


src/PostgREST/AppState.hs
```diff
@@ -476,3 +477,23 @@ checkIsFatal(SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ResultError serverEr
...
+    pqFetch pqCon = do
+      mNotification <- PQ.notifies pqCon
+      case mNotification of
+        Nothing -> do
+          mfd <- PQ.socket pqCon
+          case mfd of
+            Nothing  -> panic "Error checking for PostgreSQL notifications"
+            Just _ -> do
+              -- void $ threadWaitRead fd
+              void $ threadDelay 5000000
```
laurenceisla, line 496: I modified `waitForNotifications` from the Hasql library and it looks like this `threadWaitRead` is not working on Windows. Changing to `threadDelay` works, for instance, and reloads the schema without problems. This limitation for Windows systems is documented in the [function definition](https://hackage.haskell.org/package/base-4.16.4.0/docs/src/Control.Concurrent.html#threadWaitRead) and specifies that builds must use `-thread`, which is already added: https://github.com/PostgREST/postgrest/blob/32b77cae6c4227216b193f6834304e90de3bd7c3/postgrest.cabal#L161).

Not sure where to go from here, maybe the problem is with `stack` or somewhere in the build process.

steve-chavez: @laurenceisla Nice catch :fire:! So if `threadDelay` works how about changing `threadWaitRead` conditionally (`#if defined(mingw32_HOST_OS)..`) on a PR to https://github.com/diogob/hasql-notifications/?

laurenceisla: Yes, `threadDelay` works but I'm not sure if it's the correct solution here. Instead of waiting for input, it verifies every quantity of seconds (5s in my example). I only used it to test if the problem was indeed the `threadWaitRead`.

steve-chavez: > This limitation for Windows systems is documented in the [function definition](https://hackage.haskell.org/package/base-4.16.4.0/docs/src/Control.Concurrent.html#threadWaitRead) and specifies that builds must use -threaded, which is already added:

hasql-notifications has threaded too https://github.com/diogob/hasql-notifications/blob/master/hasql-notifications.cabal#L34

steve-chavez: Maybe report it on hasql-notifications anyway, Diogo might be able to help.

diogob: I have just setup my brand new Windows development environment and I verified that hasql-notifications' tests are green. I'm going now to build PostgREST and try to reproduce the issue there.

But regardless of the investigation, I can apply the `threadDelay` fix there so we wouldn't need to merge anything here.

I'll continue posting updates on this investigation on the issue https://github.com/diogob/hasql-notifications/issues/17

PR #2911, opened 2023-08-21
feat: allow full response control when raising exceptions

### LINKED ISSUE #2492: Allow full reponse control when raising exceptions (HTTP body + HTTP return headers)
Now when you raise exceptions you can just control HTTP reponse code and exception details, but you cannot have control about the body response or returned HTTP headers, as all previous set_config calls are not persisted after raising an exception. This limits the possibilities of the pre-request procedure as it can only control http code in case of an exception.

@wolfgangwalther solution to this would be:

Add support for a custom PGRST SQLSTATE error, where both message and detail take JSON. message would contain the response body for this error (i.e. message, detail, hint etc. all in there) and detail would contain a list of headers to set.

### Discussed in https://github.com/PostgREST/postgrest/discussions/2486

<sup>Originally posted by **yevon** September 24, 2022</sup>
I'm trying to implement a pre-request procedure that performs an exception with a 302, how can this be achieved? I don't see a mapped postgres error for 302 responses in the docs:

<html>
<body>

08* | 503 | pg connection err
09* | 500 | triggered action exception
0L* | 403 | invalid grantor
0P* | 403 | invalid role specification
23503 | 409 | foreign key violation
23505 | 409 | uniqueness violation
25006 | 405 | read only sql transaction
25* | 500 | invalid transaction state
28* | 403 | invalid auth specification
2D* | 500 | invalid transaction termination
38* | 500 | external routine exception
39* | 500 | external routine invocation
3B* | 500 | savepoint exception
40* | 500 | transaction rollback
53* | 503 | insufficient resources
54* | 413 | too complex
55* | 500 | obj not in prerequisite state
57* | 500 | operator intervention
58* | 500 | system error
F0* | 500 | config file error
HV* | 500 | foreign data wrapper error
P0001 | 400 | default code for “raise”
P0* | 500 | PL/pgSQL error
XX* | 500 | internal error
42883 | 404 | undefined function
42P01 | 404 | undefined table
42501 | if authenticated 403,else 401 | insufficient privileges
other | 400

<br class="Apple-interchange-newline">
</body>
</html></div>

taimoorzaeem:
This should fix #2492.

@steve-chavez The code needs some refactoring. But before that, please review this new interface and suggest improvements.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -27,7 +27,7 @@ import qualified Hasql.Pool                as SQL
 import qualified Hasql.Session             as SQL
 import qualified Network.HTTP.Types.Status as HTTP
 
-import Data.Aeson  ((.!=), (.:), (.:?), (.=))
+import Data.Aeson  ((.:), (.:?), (.=))
 import Network.Wai (Response, responseLBS)
 
 import Network.HTTP.Types.Header (Header)
@@ -373,7 +373,6 @@ instance PgrstError PgError where
        then [MediaType.toContentType MTApplicationJSON, ("WWW-Authenticate", "Bearer") :: Header]
        else [MediaType.toContentType MTApplicationJSON]
 
-
 instance JSON.ToJSON PgError where
   toJSON (PgError _ usageError) = JSON.toJSON usageError
 
@@ -400,11 +399,11 @@ instance JSON.ToJSON SQL.CommandError where
       (Just r, Just _) -> JSON.object [
         "code"     .= getCode r,
         "message"  .= getMessage r,
-        "details"  .= checkEmpty (getDetails r),
-        "hint"     .= checkEmpty (getHint r)]
+        "details"  .= checkMaybe (getDetails r),
+        "hint"     .= checkMaybe (getHint r)]
       _ -> JSON.toJSON JSONParseError
     where
-      checkEmpty det = if det == "" then JSON.Null else JSON.String det
+      checkMaybe = maybe JSON.Null JSON.String
 
   toJSON (SQL.ResultError (SQL.ServerError c m d h _p)) = JSON.object [
     "code"     .= (T.decodeUtf8 c      :: Text),
@@ -423,7 +422,6 @@ instance JSON.ToJSON SQL.CommandError where
     "message" .= ("Database client error. Retrying the connection." :: Text),
     "details" .= (fmap T.decodeUtf8 d :: Maybe Text),
     "hint"    .= JSON.Null]
-
 
 pgErrorStatus :: Bool -> SQL.UsageError -> HTTP.Status
 pgErrorStatus _      (SQL.ConnectionUsageError _) = HTTP.status503
@@ -466,12 +464,11 @@ pgErrorStatus authed (SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ResultError
         'P':'T':n -> fromMaybe HTTP.status500 (HTTP.mkStatus <$> readMaybe n <*> pure m)
         "PGRST"   ->
           case (parseMessage m, parseDetails d) of
-            (Just _, Just r) -> HTTP.mkStatus (getStatus r) "CustomHTTPResponse"
+            (Just _, Just r) -> maybe (toEnum $ getStatus r) (HTTP.mkStatus (getStatus r) . T.encodeUtf8) (getStatusText r)
             _                -> status JSONParseError
         _         -> HTTP.status400
 
     _                       -> HTTP.status500
-
 
 
 data Error
@@ -500,7 +497,7 @@ instance PgrstError Error where
   status (PgErr err)             = status err
   status PutMatchingPkError      = HTTP.status400
   status SingularityError{}      = HTTP.status406
-  status JSONParseError          = HTTP.status400
+  status JSONParseError          = HTTP.status500
 
   headers (ApiRequestError err)  = headers err
   headers (JwtTokenInvalid m)    = [MediaType.toContentType MTApplicationJSON, invalidTokenHeader m]
@@ -584,13 +581,14 @@ singularityError = SingularityError . toInteger
 data PgRaiseErrMessage = PgRaiseErrMessage {
   getCode    :: Text,
   getMessage :: Text,
-  getDetails :: Text,
-  getHint    :: Text
+  getDetails :: Maybe Text,
+  getHint    :: Maybe Text
 }
 
 data PgRaiseErrDetails = PgRaiseErrDetails {
-  getStatus  :: Int,
-  getHeaders :: Map Text Text
+  getStatus     :: Int,
+  getStatusText :: Maybe Text,
+  getHeaders    :: Map Text Text
 }
 
 instance JSON.FromJSON PgRaiseErrMessage where
@@ -598,8 +596,8 @@ instance JSON.FromJSON PgRaiseErrMessage where
     PgRaiseErrMessage
       <$> m .: "code"
       <*> m .: "message"
-      <*> m .:? "details" .!= ""
-      <*> m .:? "hint"    .!= ""
+      <*> m .:? "details"
+      <*> m .:? "hint"
 
   parseJSON _ = mzero
 
@@ -607,16 +605,16 @@ instance JSON.FromJSON PgRaiseErrDetails where
   parseJSON (JSON.Object d) =
     PgRaiseErrDetails
       <$> d .: "status"
+      <*> d .:? "status_text"
       <*> d .: "headers"
 
   parseJSON _ = mzero
 
 parseMessage :: ByteString -> Maybe PgRaiseErrMessage
-parseMessage m = JSON.decodeStrict m :: Maybe PgRaiseErrMessage
+parseMessage = JSON.decodeStrict
 
 parseDetails :: Maybe ByteString -> Maybe PgRaiseErrDetails
-parseDetails (Just d) = JSON.decodeStrict d :: Maybe PgRaiseErrDetails
-parseDetails Nothing  = Nothing
+parseDetails d = JSON.decodeStrict =<< d
 
 -- Error codes are grouped by common modules or characteristics
 data ErrorCode
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -16,8 +16,10 @@ module PostgREST.Error
 
 import qualified Data.Aeson                as JSON
 import qualified Data.ByteString.Char8     as BS
+import qualified Data.CaseInsensitive      as CI
 import qualified Data.FuzzySet             as Fuzzy
 import qualified Data.HashMap.Strict       as HM
+import qualified Data.Map.Internal         as M
 import qualified Data.Text                 as T
 import qualified Data.Text.Encoding        as T
 import qualified Data.Text.Encoding.Error  as T
@@ -25,7 +27,7 @@ import qualified Hasql.Pool                as SQL
 import qualified Hasql.Session             as SQL
 import qualified Network.HTTP.Types.Status as HTTP
 
-import Data.Aeson  ((.=))
+import Data.Aeson  ((.!=), (.:), (.:?), (.=))
 import Network.Wai (Response, responseLBS)
 
 import Network.HTTP.Types.Header (Header)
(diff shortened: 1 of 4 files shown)


src/PostgREST/Error.hs
```diff
@@ -587,6 +629,8 @@ data ErrorCode
   | JWTErrorCode02
   -- Internal errors related to the Hasql library
   | InternalErrorCode00
+  -- JSON parse error, related to invalid json sent via RAISE for custom response
+  | JSONParseErrorCode00
```
steve-chavez, line 633: There's a pretty similar error code for an incorrect format on `set_config('response.headers',..`, see:

- https://postgrest.org/en/stable/references/transactions.html#guc-resp-hdrs
- https://postgrest.org/en/stable/references/errors.html#pgrst111

So this error should belong to the same group, likely a `PGRST121`.

src/PostgREST/Error.hs
```diff
@@ -533,6 +569,12 @@ instance JSON.ToJSON Error where
     "details" .= T.unwords ["The result contains", show n, "rows"],
     "hint"    .= JSON.Null]
 
+  toJSON JSONParseError = JSON.object [
+    "code"    .= JSONParseErrorCode00,
+    "message" .= ("The message and detail field of RAISE error expects JSON" :: Text),
```
steve-chavez, line 574:
```suggestion
    "message" .= ("The message and detail field of RAISE 'PGRST' error expects JSON" :: Text),
```

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1490,3 +1490,25 @@ spec actualPgVersion =
...
+
+      it "returns JSONParseError for invalid JSON in RAISE Message field" $
+        get "/rpc/raise_sqlstate_invalid_json_message" `shouldRespondWith`
+          [json|{"code":"PGRSTJ00","message":"The message and detail field of RAISE error expects JSON","details":null,"hint":null}|]
+          { matchStatus = 400 }
+
+      it "returns JSONParseError for invalid JSON in RAISE Details field" $
+        get "/rpc/raise_sqlstate_invalid_json_details" `shouldRespondWith`
+          [json|{"code":"PGRSTJ00","message":"The message and detail field of RAISE error expects JSON","details":null,"hint":null}|]
+          { matchStatus = 400 }
```
steve-chavez, line 1514: @taimoorzaeem The end result looks good :+1:

As you mention, some refactoring is needed. Ideally we don't introduce attoparsec and stick with parsec.

taimoorzaeem: @steve-chavez For parsec, I would have to write a complete JSON parser from scratch which seems rather unnecessary.

src/PostgREST/Error.hs
```diff
@@ -633,3 +684,47 @@ buildErrorCode code = "PGRST" <> case code of
...
+      <$> d .: "status"
+      <*> d .: "headers"
+
+  parseJSON _ = mzero
+
+parseMessage :: ByteString -> Maybe PGRSTRaiseErrMessage
+parseMessage m =
+  case ATP.parse JSON.json m of
+    (ATP.Done _ r) -> JSON.parseMaybe JSON.parseJSON r :: Maybe PGRSTRaiseErrMessage
+    _          -> Nothing
```
laurenceisla, line 723: I think this can be done directly with Aeson using `decode` or [`decodeStrict`](https://hackage.haskell.org/package/aeson-2.2.0.0/docs/Data-Aeson.html#v:decodeStrict) for this type of ByteString?

```suggestion
parseMessage :: ByteString -> Maybe PGRSTRaiseErrMessage
parseMessage m = JSON.decodeStrict m :: Maybe PGRSTRaiseErrMessage
```

taimoorzaeem: @laurenceisla Thanks! Now we can also remove `attoparsec`.

src/PostgREST/Error.hs
```diff
@@ -359,11 +361,24 @@ type Authenticated = Bool
...
+    let
+      intoHeader (k,v) = ((CI.mk . T.encodeUtf8) k, T.encodeUtf8 v)
+    in
+      case parseMessage m of -- make sure both parses successfully
+        Just _ ->
+          case parseDetails d of
+            Just obj -> headers JSONParseError
+                          ++ map intoHeader (M.toList $ getHeaders obj)
+            Nothing -> headers JSONParseError
+        Nothing -> headers JSONParseError
```
laurenceisla, line 374: This could be refactored by using a Tuple:

```suggestion
    case (parseMessage m, parseDetails d) of -- make sure both parses successfully
      (Just _, Just obj) -> headers JSONParseError ++ map intoHeader (M.toList $ getHeaders obj)
      _                  -> headers JSONParseError
    where
      intoHeader (k,v) = ((CI.mk . T.encodeUtf8) k, T.encodeUtf8 v)
```

Or if you have something better, go for it. It could be applied to similar code below.

taimoorzaeem: Great suggestion! Much clearer and shorter.

src/PostgREST/Error.hs
```diff
@@ -473,6 +500,7 @@ instance PgrstError Error where
   status (PgErr err)             = status err
   status PutMatchingPkError      = HTTP.status400
   status SingularityError{}      = HTTP.status406
+  status JSONParseError          = HTTP.status400
```
laurenceisla, line 503:
```suggestion
  status JSONParseError          = HTTP.status500
```

This is similar to the `GucStatusError`, so this should be a Server error (500) instead of Client (400).

(BTW, I made a commit to the branch, so you should pull before making these changes).

src/PostgREST/Error.hs
```diff
@@ -442,6 +464,10 @@ pgErrorStatus authed (SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ResultError
         "42P01"   -> HTTP.status404 -- undefined table
         "42501"   -> if authed then HTTP.status403 else HTTP.status401 -- insufficient privilege
         'P':'T':n -> fromMaybe HTTP.status500 (HTTP.mkStatus <$> readMaybe n <*> pure m)
+        "PGRST"   ->
+          case (parseMessage m, parseDetails d) of
+            (Just _, Just r) -> HTTP.mkStatus (getStatus r) "CustomHTTPResponse"
```
steve-chavez, line 469: Hm, what does `CustomHTTPResponse` mean in this context?

src/PostgREST/Error.hs
```diff
@@ -359,11 +361,19 @@ type Authenticated = Bool
...
+      _                -> headers JSONParseError
+    where
+      intoHeader (k,v) = (CI.mk $ T.encodeUtf8 k, T.encodeUtf8 v)
+
   headers err =
     if status err == HTTP.status401
        then [MediaType.toContentType MTApplicationJSON, ("WWW-Authenticate", "Bearer") :: Header]
        else [MediaType.toContentType MTApplicationJSON]
 
+
```
steve-chavez, line 376:
```suggestion
```

src/PostgREST/Error.hs
taimoorzaeem: I was thinking since we are can return any HTTP status and headers through this interface, so just called it `CustomHTTPResponse` without thinking too much. How about `PgRaiseResponse` or maybe just leave it blank `""`?

taimoorzaeem: @steve-chavez Please review and let me know.

laurenceisla: This should be using the default HTTP status messages, e.g. if the user specifies `{"status": 404,...}", then the message should be "Not Found". It can be achieved by doing:

```haskell
(Just _, Just r) -> toEnum (getStatus r)
```

But maybe we could also add further extensibility by letting the user specify the status description if they wish. For instance:

```json
{"status": 332, "description": "My Custom Status", ...}
```

Then you can do something like (didn't test this, just an idea):

```haskell
(Just _, Just r) -> maybe  (toEnum $ getStatus r) (HTTP.mkStatus $ getStatus r) (getDescription r)
```

For that, the `PgRaiseErrDetails` needs another field like:

```haskell
getDescription :: Maybe ByteString
```

Something like that (Maybe ByteString is not allowed, use Text instead).

test/spec/fixtures/schema.sql
```diff
@@ -3368,3 +3368,72 @@ $$ language sql;
...
 select * from projects p where p.id >= $1 and p.name like $2;
 $$ language sql;
+
+create function raise_sqlstate_test1() returns void
+  language plpgsql
+  as $$
+begin
+    raise sqlstate 'PGRST' USING
+      message = '{"code":"123","message":"ABC","details":"DEF","hint":"XYZ"}',
+      detail = '{"status":332,"description":"My Custom Status","headers":{"X-Header":"str"}}';
```
steve-chavez, line 3378: @taimoorzaeem @laurenceisla `description` by itself feels a bit loose. The official term for the status description is "reason phrase". See:

- https://www.rfc-editor.org/rfc/rfc7230#section-3.1.2
- https://www.ibm.com/docs/en/cics-ts/5.4?topic=concepts-status-codes-reason-phrases

So I think there are two options for making this more correct:

1) One single field for the status line, the reason phrase being optional: `"status": "332"` or `"status": "332 My Custom Status"`. Harder to implement but better DX for the user.
2) Change `description` to `status_text` (also a known term for it). Simpler to implement but less DX for the user. Though the case where they'd want custom status codes is too rare, so it doesn't matter as much.

Personally I'd go with option 2.

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1490,3 +1490,40 @@ spec actualPgVersion =
...
+      it "gets message and details from raise sqlstate PGRST" $ do
+        get "/rpc/raise_sqlstate_test1" `shouldRespondWith`
+          [json|{"code":"123","message":"ABC","details":"DEF","hint":"XYZ"}|]
+          { matchStatus = 332
+          , matchHeaders = ["X-Header" <:> "str"] }
+
+        get "/rpc/raise_sqlstate_test2" `shouldRespondWith`
+          [json|{"code":"123","message":"ABC","details":null,"hint":null}|]
+          { matchStatus = 332
+          , matchHeaders = ["X-Header" <:> "str"] }
```
steve-chavez, line 1504: Btw, we can also test the status message. There's an example on:

https://github.com/PostgREST/postgrest/blob/531a183b44b36614224fda432335cdaa356b4a0a/test/spec/Feature/Query/PlanSpec.hs#L26-L40

See `statusMessage="OK"`.

src/PostgREST/Error.hs
```diff
@@ -546,6 +577,45 @@ requiredTokenHeader = ("WWW-Authenticate", "Bearer")
...
+  getCode    :: Text,
+  getMessage :: Text,
+  getDetails :: Maybe Text,
+  getHint    :: Maybe Text
+}
+
+data PgRaiseErrDetails = PgRaiseErrDetails {
+  getStatus      :: Int,
+  getDescription :: Maybe Text,
+  getHeaders     :: Map Text Text
```
laurenceisla, line 591:
```suggestion
  status      :: Int,
  description :: Maybe Text,
  headers     :: Map Text Text
```

Now that I see it, I think avoiding the "get" here is better as a data type.

src/PostgREST/Error.hs
```diff
@@ -546,6 +577,45 @@ requiredTokenHeader = ("WWW-Authenticate", "Bearer")
 singularityError :: (Integral a) => a -> Error
 singularityError = SingularityError . toInteger
 
+-- For parsing byteString to JSON Object, used for allowing full response control
+data PgRaiseErrMessage = PgRaiseErrMessage {
+  getCode    :: Text,
+  getMessage :: Text,
+  getDetails :: Maybe Text,
+  getHint    :: Maybe Text
```
laurenceisla, line 585: Ditto.

test/spec/fixtures/schema.sql
laurenceisla: Cool! I'd choose option 2 too.

src/PostgREST/Error.hs
steve-chavez: `description` to `statusText` would be better as well.

taimoorzaeem: @laurenceisla `status` and `headers` are already used by `PgrstError`, so I can't use those here otherwise variable name shadowing warning shows up.

laurenceisla (approved):

steve-chavez:
@taimoorzaeem Great work! You can document it on https://postgrest.org/en/stable/references/errors.html#raise-errors-with-http-status-codes

yevon:
Great news! Really looking forward to this one!

src/PostgREST/Error.hs
taimoorzaeem: Thanks @wolfgangwalther. I'll change this in a future refactoring.

PR #2916, opened 2023-08-24
fix: duplicate headers in response

### LINKED ISSUE #2915: Regression: Using `SET response.headers` for Content-Type returns headers twice
### Environment

* PostgreSQL version: any
* PostgREST version: 11.2.0
* Operating system:

### Description of issue

Using this:

```sql
create function response_headers() returns text
language plpgsql as $$
begin
  set local response.headers to '[{"Content-Type": "custom/type"}]';
  return 'hello world';
end$$;
```

returns the following:

```
% curl localhost:3000/rpc/response_headers -I
HTTP/1.1 200 OK
Date: Wed, 23 Aug 2023 15:30:08 GMT
Server: postgrest/11.2.0 (b8b3145)
Content-Range: 0-0/*
Content-Type: custom/type
```

Note the duplicated header `Content-Type`. This leads to browsers taking the first header for granted, which breaks all kinds of file downloads in my app.

Bisecting leads me to d490bf0 and #2884 - seems to have been introduced there.

taimoorzaeem:
This should close #2915.

### DIFF
--- a/src/PostgREST/Response.hs
+++ b/src/PostgREST/Response.hs
@@ -297,11 +297,36 @@ optionalRollback AppConfig{..} ApiRequest{iPreferences=Preferences{..}} resp = d
       | otherwise =
           identity
 
--- | Add headers not already included to allow the user to override them instead of duplicating them
+-- | Add headers not already included to allow the user to override them instead of duplicating them. The exception here is the Preference-Applied header, which will be duplicated here, but later get combined into a single header
+--
+-- >>> :{
+--  addHeadersIfNotIncluded
+--     [("Content-Type","application/json"),
+--      ("Preference-Applied","tx=commit"),
+--      ("Content-Range","*/*")]
+--     [("Content-Type","custom/type"),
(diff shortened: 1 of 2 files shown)


src/PostgREST/Response.hs
```diff
@@ -297,11 +297,36 @@ optionalRollback AppConfig{..} ApiRequest{iPreferences=Preferences{..}} resp = d
...
+--   because preferRepresentation is only added once in the request processing
+--   pipeline. Thus, no need to add an extra filter. In case this becomes a
+--   problem in the future, we can always change this
+--
+-- >>> :{
+--  addHeadersIfNotIncluded
+--      [("Preference-Applied","return=minimal")]
+--      [("Preference-Applied","return=representation")]
+-- :}
+-- [("Preference-Applied","return=minimal"),("Preference-Applied","return=representation")]
```
steve-chavez, line 322: Yeah, this seems fine for now :+1:

steve-chavez (approved):

PR #2921, opened 2023-08-28
fix: range request with first position same as length return status 206

### LINKED ISSUE #2824: Range request with first position the same as length full response should return 416
Adding the following test-case to `RangeSpec.hs`:

```haskell
        it "refuses a range with first position the same as number of items" $
          request methodGet "/rpc/getitemrange?min=1&max=2"
                  (rangeHdrsWithCount $ ByteRangeFromTo 1 2) mempty
            `shouldRespondWith`
              [json| {
                "message":"Requested range not satisfiable",
                "code":"PGRST103",
                "details":"An offset of 1 was requested, but there are only 1 rows.",
                "hint":null
              }|]
            { matchStatus  = 416
            }
```

fails, because the response is a an empty 206 response:

```
       status mismatch:
         expected: 416
         but got:  206
       body mismatch:
         expected: {"code":"PGRST103","details":"An offset of 1 was requested, but there are only 1 rows.","hint":null,"message":"Requested range not satisfiable"}
         but got:  []
```

This is incorrect.

The `/rpc/getitemrange?min=1&max=2` resource has 1 item.

The first item is `0-1`. Everything with a higher first position, [is not satisfiable](https://www.rfc-editor.org/rfc/rfc9110#name-byte-ranges):

> For a GET request, a valid bytes range-spec is satisfiable if it is [...] an int-range with a first-pos that is **less** than the current length of the selected representation [...]

(emphasis mine)

taimoorzaeem:
This should fix #2824.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2899, Fix `application/vnd.pgrst.array` not accepted as a valid mediatype - @taimoorzaeem
  - #2524, Fix schema cache and configuration reloading with `NOTIFY` not working on Windows - @diogob, @laurenceisla
  - #2915, Fix duplicate headers in response - @taimoorzaeem
+ - #2824, Fix range request with first position same as length return status 206 - @taimoorzaeem
 
 ## [11.2.0] - 2023-08-10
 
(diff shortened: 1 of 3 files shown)


steve-chavez (approved):
LGTM! The fix looks simple enough.

Issue #2730, opened 2023-04-01
Sort By Count Aggregation


jdgamble555:
While there is a need for `GROUP BY` in PostgREST, this feature would take care a lot of the use cases:

https://github.com/PostgREST/postgrest/issues/158
https://github.com/PostgREST/postgrest/issues/167
https://github.com/PostgREST/postgrest/issues/915
https://github.com/PostgREST/postgrest/issues/158

It would be nice to be able to say:

```graphql
GET /films?select=title,directors(id,last_name,_aggregate.count)?order=last_name(_aggregate.count) HTTP/1.1
```
I'm thinking you should be able to sort by the top level table OR by the foreign key / child table.

It should work similar to GraphQL like Hasura:

https://hasura.io/docs/latest/queries/postgres/sorting/#for-array-relationships

That way, we don't need to think about Group By, but think about the actual data and how you want to sort it.

So, from a regular data perspective, let's say we want to display the Top Votes of a Post:

```sql
CREATE TABLE posts (
  id uuid
  vote_id uuid
  ...
)
CREATE TABLE votes (
  post_id uuid REFERENCES posts(vote_id)
  user_id uuid REFERNCES users(id)
  PRIMARY KEY(post_id, user_id)
  ...
)
```

So we want something like this:
```
?order=posts(votes._count)
```
OR
```
?order=posts(votes._aggregate.count)
```

There are different ways to accomplish this, but basically you wouldn't need to think about `GROUP BY`, and it would write this SQL code automatically (maybe with Group By).  

J

arbela-cyberillium:
I would love to just be able to get the count

`?select=title,directors(_aggregate.count)`

so that way I can know how many directors are there for the film (I didn't found any work around that will support filters )

steve-chavez:
>?select=title,directors(_aggregate.count)

Ah, actually the above already works like:

```http
GET /tbl?select=title,directors(count)
```

Caveats:

- You can't include any other columns inside `directors(..)`.
- The `count` will be inside an array.

steve-chavez:
To elaborate on the above:

```bash
curl 'localhost:3000/clients?select=id,name,projects(count)'

[
 {"id":1,"name":"Microsoft","projects":[{"count":3}]},
 {"id":2,"name":"Apple","projects":[{"count":2}]}
]
```

---

Addressing the OP issue.

Using the new related order feature: https://github.com/PostgREST/postgrest/pull/2511

I tried to do:

```bash
[...]

PR #2925, opened 2023-08-30
Add aggregate functions

timabdulla:
This is an initial attempt to add support for aggregate functions, using the call pattern suggested here: https://github.com/PostgREST/postgrest/issues/915#issuecomment-1039522570 by @wolfgangwalther.

Note that I am aware that this PR is missing a number of things (at the very least, `count()`, support for `HAVING`, the proposed limit on query cost to prevent denial-of-service failures, and, of course, automated tests), and I haven't fully tested this yet beyond some initial sanity checks.

What I would really like to understand at this point is a) is the technical direction I am taking one that is appropriate, and b) what are the must-haves before this feature could be considered merge-able.

Thank you for your time and attention!


steve-chavez:
@timabdulla Thanks for taking the initiative!

> the proposed limit on query cost to prevent denial-of-service failures

This part seems different now that we have [Impersonated Role Settings](https://postgrest.org/en/stable/references/auth.html#impersonated-role-settings), we could leave the job to https://github.com/pgexperts/pg_plan_filter. For users that don't want to deal with aggregates, we could add a config option to enable/disable them (better save this as a last step of the implementation though).

One other thing that we should consider is avoiding failure from clients whenever they add an aggregate but forget to specify GROUP BY. Using a window function for no GROUP BY (as mentioned on https://github.com/PostgREST/postgrest/issues/2066#issuecomment-1699707246, sorry this is all over the place) might be a good solution for this.

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -496,6 +519,7 @@ setConfigLocalJson prefix keyVals = [setConfigLocal mempty (prefix, gucJsonVal k
     arrayByteStringToText :: [(ByteString, ByteString)] -> [(Text,Text)]
     arrayByteStringToText keyVal = (T.decodeUtf8 *** T.decodeUtf8) <$> keyVal
 
+-- Investigate this
```
steve-chavez, line 522: @timabdulla You don't need to worry about this part.

We could say that these are "final aggregates", they're used to send an "interchange format" to clients. i.e. json, csv, xml, binary, etc.

For this feature we're dealing with regular aggregates, used to fold a number of rows into a postgres native format.

steve-chavez:
> Using a window function for no GROUP BY (as mentioned on https://github.com/PostgREST/postgrest/issues/2066#issuecomment-1699707246, sorry this is all over the place) might be a good solution for this.

Considering the above, it might possible to isolate a first pass of the feature to only supporting aggregates and not tie them to GROUP BY.

timabdulla:
Hey @steve-chavez,

Really appreciate your quick feedback!

In terms of `GROUP BY`, I was following the guidance from this comment, though not sure if it's still relevant:

https://github.com/PostgREST/postgrest/issues/915#issuecomment-1026720780

Specifically, this part:

> I thought a bit more about the proposed addition to the query string in the form of groupby=.... I think we can do without.
> As long as we detect the possible aggregation functions via schema cache, we should know whenever an aggregate is called. Once that's the case, we can add all columns mentioned in the select= part, that are not aggregates to the GROUP BY. They'd need to be added anyway, if they were not functionally dependent on each other - and in that case it doesn't hurt to add them either.

The `GROUP BY` clause is automatically computed based on the columns selected. If only aggregates are selected, it defaults collapsing down to one row with aggregation occurring across the whole table (i.e. no `GROUP BY`.)

Let me know what you think. I'll also take a look at Impersonated Role Settings to get a better understanding of that feature.

steve-chavez:
> In terms of GROUP BY, I was following the guidance from this comment, though not sure if it's still relevant:
> https://github.com/PostgREST/postgrest/issues/915#issuecomment-1026720780

Ah, I missed the above.

Regarding inferring the GROUP BY columns, I think it would be better to default to doing aggregates with the window function approach. 

This way the elements in `select` are just transformations, they don't filter rows. Otherwise we might run into the same issue as `/a?select=b!inner(*)` which also filters rows (and later we found out that `/a?select=b(*)&b=not.is.null` was better as it's more explicit).

@wolfgangwalther Please give your feedback whenever you have the time.

timabdulla:
Ok, got it. Thanks for explaining. That seems a simple enough change to make: Basically without an explicit GROUP BY, default to doing an aggregate over an "empty" window (e.g. `SUM(salary) OVER ()`).

We would then need a new query parameter for `groupby`, but as you said, this could perhaps be kept for a separate PR. If we do that, that would also imply `HAVING` being a separate PR too, along with `GROUP BY`.

I also still need to add support for `count()` with no associated field, which seems like it may require its own separate bit of handling. Will do some thinking on that.

steve-chavez:
> A "group by" by default would be very useful

It is very useful, I don't disagree with that.

> The same rows still go into the resultset, they are just aggregated into a single value. But they have been part of the computation and have not been filtered out before.

Hm, the fact remains that adding an element in `?select=` reduces the number of rows.

I fear that violating a semantic we agreed upon before (`select` shouldn't affect the number of rows) will result in inconsistency and tech debt later. This is what makes me think that `groupby` should be explicit.

How about special syntax for auto-derive group by though? Could be like

```bash
groupby=*
#or
groupby=!
```

Idea taken from https://stackoverflow.com/a/416764/4692662

src/PostgREST/Query/SqlFragment.hs
timabdulla: Thanks for your comment @wolfgangwalther. To be honest, I don't have as much context on the bigger-picture elements of the architecture and the like, so I am not totally clear on the practical implications of your comment as it relates to this PR. Any examples or illustrations would be very helpful! Thank you.

steve-chavez:
> Doing a query without select will fetch the full response, while adding a single column to select to make it `select=count` changes the query heavily. And we should certainly not require a group by clause for a query with only aggregates columns, because that's not required in SQL either.

You're right. My bad, I forgot about the `count` case.

> I doubt that anyone really needs to either:
> - hide columns in the ouput that they want to group on
> - select more columns, which are not grouped on, only to receive an error by postgres about that

> One advantage of that is, that we will also have the case of `select=count` covered, which already works right now, although not on purpose. 

Great points. Agree, we should omit `groupby` then.

@timabdulla Sorry for the back and forth on this. Let's stick with Wolfgang's proposal.

timabdulla:
Thanks folks! Then I'll work to get this finished as per the original plan. One thing: I might omit support for `HAVING`, as it doesn't seem necessary for a minimal first release to me -- any thoughts on that?

src/PostgREST/Query/SqlFragment.hs
aljungberg: Commenting without being fully read in on this PR, just to answer the question: are the 'final' aggregates different from other aggregation (in our envisioned future)?

So I think the answer is 'no'. With pervasive data representations, `json_agg`, CSV representation, or even more exotic forms like "encode into a protobuf buffer" are all just pluggable transformations which we would no longer hard-code special cases for. 

So `json_agg` would just be one of the built-in data representations when "what we have" (an intermediate type that needs to be aggregated) differs from "what we want" (output consistent with the negotiated content type `application/json`).

timabdulla:
Hi folks,

I'm getting pretty close with my changes, but I had a few things I wanted to bring up:

## 'Hoisting' Aggregate Functions in the Case of Has-One Spreads

Imagine the following example (and please forgive me if the details are a little contrived): We have a table called `projects` that is related to another table, `project_invoices`, through a has-one relationship, and we have the following query: `GET /projects?select=client_id,...project_invoices(invoice_total.sum())`.

As it stands, the aggregation occurs _within_ the subquery of the lateral join, and thus the sum basically does nothing: There is only one row within the subquery (this being a has-one relationship), and thus the sum of the `invoice_total` is the `invoice_total` itself. 

Given this, I _can_ see an argument for why we might 'hoist' the aggregate function to the top-level in the case of has-one spreads: Aggregate functions would be brought to the top-level (along with the associated `GROUP BY` terms). In the above example, that would mean that we would get the sum of the `invoice_total` by `client_id`.

But I then wonder a bit about consistency, as without resource spreading, this form of hoisting is not possible (or sensible, really), which means that `GET /projects?select=client_id,project_invoices(invoice_total.sum())` would exhibit the same behaviour as shown before. The same would apply to has-many relationships, as spreading is not possible with has-many relationships.

It is important to note that in any case, there is also a workaround, in that if you invert the original query (i.e. ``GET /project_invoices?select=invoice_total.sum(),...projects(client_id)``), you can get the sum of the `invoice_total` grouped by `client_id`.

*The question*: Should aggregate functions be hoisted in the case of has-one spreads? The implementation is not trivial, but it's also not hugely complex.

## Order of Operations on Selects

Currently, the select syntax is implemented like so: `?select=alias:column->jsonField::cast.aggregateFunction()`. With the addition of aggregate functions, the order of operations becomes somewhat ambiguous, specifically with respect to casting. 

I have opted for casting to be performed on the column itself _before_ aggregation, as this seems much more useful than casting the return value of the aggregate function itself. This would be most useful in the case of wanting to perform an aggregation on a `jsonb` value, as, for instance, a number coming from a `jsonb` object would need to be casted to a `numeric` before aggregation can be performed (e.g. `?select=invoice_details->>total::numeric.sum()`).

Please let me know if you have other thoughts or otherwise disagree with this choice.

## `count()`

Support for `count()` is included, both as `count(*)` (`?select=count(),project_id`), or as `count(column)` (`?select=project_id.count()`). That said, I did wonder about the existing (accidental) support for `count` (`?select=count`).

I don't think it should be eliminated in this PR, but I would suggest that it should ultimately be eliminated as it is inconsistent with this feature and is slightly confusing. Perhaps it may be sensible to issue a deprecation notice of some kind, with a plan to eventually remove `count`? Although, I am not sure what form this elimination would take... You can have a column named `count` so maybe it's more trouble than it's worth to try to distinguish between the two cases (accidental count operation vs selecting a real column named `count`).

I am quite close to be finished, and I hope to have a final PR ready for review early next week. Thanks for reading!

timabdulla:
Thanks for the quick reply!

1. I will add the hoisting then, stay tuned.
2. Yes, it should be fairly easy to add an additional cast option. Will add that.
3. Understood, let's leave for future work then.

timabdulla:
Random thought: Perhaps (not in this PR) there could be a possibility of has-many spreads, so long as aggregate functions are involved. Example use case: `GET /invoices?select=customer_id,...invoice_line_items(line_amount.sum())`. Basically, the `line_amount` would get hoisted into the top-level, and what you would end up with is the invoice total (defined to be the sum of all `line_amount`s) by the `customer_id`.

timabdulla:
Ok, I now have something I feel that is pretty feature-complete. Still pending code cleanup (not yet ready for review, I would say), automated tests and docs. Hopefully I can knock that off that off this week, and then bring it for review. The PR has grown a fair bit in size, as handling all the various cases well took a bit of refactoring.

timabdulla:
Ok, this PR is finally ready for review. The only thing I've yet to do is finish rounding out the tests, though I already have some tests in. That said, the actual code itself is more or less complete from my perspective. I welcome your comments and feedback; while I await that, I'll finish up the tests.

Overview
---------

As a reminder, this PR enables support for aggregate functions.

The basic request syntax works like so: `GET /table?select=foo.sum(),bar,baz`. The group by is determined implicitly via the fields in the select list.

You may also optionally provide a cast for the output of the aggregate function, like so: `GET /table?select=foo.sum()::text,bar,baz`. This is distinct from the cast that can be provided on the input column. Here is an example showing both casts in use: `GET /table?select=foo::int.sum()::text,bar,baz`.

You may also provide a `count()` without reference to a specific field, like so: `GET /table?select=count()`. In this case, you can still provide an alias and/or an aggregate cast.

The following aggregate functions are supported: `sum`, `min`, `max`, `avg`, and `count`.

Aggregate functions are also supported in the case of embedded resource spreads by means of 'hoisting'. Further detail provided here: https://github.com/PostgREST/postgrest/pull/2925/files#diff-2295ae0baafabcf593f2991ae456859ad5c3a8190ff532155da6e8760d08411cR611

Notes and observations
----------------------

* Generally speaking, I've been trying to nudge things in the direction of the `ReadPlanTree` being the abstract representation of the query to be generated, with the SQL generation code being more confined to the mechanical process of generation rather than making many decisions of its own. This makes it easier to reflect on and manipulate the query, and I think just makes things simpler to understand. For instance, we already had a function `addDataRepresentationAliases` in the `Plan` module, but automatically determining the alias for a field that arises from an operator on a JSON column was done in the `Query.SqlFragment` module. I've now combined that in a more general `addAliases` function in the `Plan`. This arose out of a specific need I had around needing to know the alias at plan time (i.e. not a proactive refactor), but it turns out I had similar needs multiple times, and this was a pattern I turned to frequently.
* I think there's probably a need for a refactor of some of the core types. There's a parallel type hierarchy that is emerging between `ApiRequest.Types` and `Plan.Types` on a subset of the types (e.g. `CoercibleOrderTerm` and `OrderTerm`, `CoercibleLogicTree` and `LogicTree`, etc.) Additionally, `Plan.Types` ends up importing a lot of things from `ApiRequest.Types`, which makes it tricky to clearly see the delineation between the two.


Thank you folks. I look forward to hearing from you.

src/PostgREST/Plan.hs
```diff
@@ -333,49 +335,86 @@ initReadRequest ctx@ResolverContext{qi=QualifiedIdentifier{..}} =
...
-    expandStars rPlan@ReadPlan{from=fromTable} =
-      expandStarsForTable ctx{qi=fromTable} rPlan
-
-expandStarsForTable :: ResolverContext -> ReadPlan -> ReadPlan
-expandStarsForTable ctx@ResolverContext{representations, outputType} rplan@ReadPlan{select=selectItems} =
-  -- If we have a '*' select AND the target table has at least one data representation, expand.
-  if ("*" `elem` map (\(field, _, _) -> cfName field) selectItems) && any hasOutputRep knownColumns
-    then rplan{select=concatMap (expandStarSelectItem knownColumns) selectItems}
-    else rplan
+    adjustContext context@ResolverContext{qi=ctxQI} (QualifiedIdentifier "" "pgrst_source") (Just a) = context{qi=ctxQI{qiName=a}}
```
aljungberg, line 394: Nice move sticking this bit of the action into `adjustContext`! That helps. Somehow I do miss the straightness of matching on `from=..., fromAlias=...tblAlias` directly in the plan though, to me that really spelled out what condition triggers this special case. This is probably bike-shedding but to get back a little of that one could perhaps imagine renaming `adjustContext` to `resolveTable` since that's basically what it does (if we are accessing indirectly via `pgrst_source` we figure out the true underlying table as documented, otherwise we're already resolved), and `a` in the capture to `fromAlias`.

src/PostgREST/Plan.hs
```diff
@@ -333,49 +335,86 @@ initReadRequest ctx@ResolverContext{qi=QualifiedIdentifier{..}} =
...
+    adjustContext context fromQI _ = context{qi=fromQI}
+
+expandStarsForTable :: ResolverContext -> Bool -> ReadPlan -> ReadPlan
+expandStarsForTable ctx@ResolverContext{representations, outputType} hasAgg rp@ReadPlan{select=selectFields}
+  -- We expand if either of the below are true:
+  -- * We have a '*' select AND there is an aggregate function in this ReadPlan's sub-tree.
+  -- * We have a '*' select AND the target table has at least one data representation.
+  -- We ignore any '*' selects that have an aggregate function attached (i.e for COUNT(*)).
```
aljungberg, line 402: I wonder if there's a case where an aggregate function needs to operate on a data rep domain column. It doesn't seem out of the realm of possibility. In that case we'd need to expand these stars too, wouldn't we? But I guess that would be a future enhancement.

src/PostgREST/Plan.hs
timabdulla: I like the suggestion, will do!

aljungberg:
Agree with the strategy of doing most of the heavy lifting in the planning stage.

And yes, the 'shadow' type structure we're arriving at is perhaps a little awkward. Part of it started with data reps because we had different correctness guarantees in the query versus the plan, if I remember right, so we decided to not just extend the existing types.

timabdulla:
@steve-chavez I see some of my tests are failing due to test data differences for different versions of Postgres. Is that deliberate? It looks like `test.car_model_sales` is missing in PG <= 11.

steve-chavez:
@timabdulla Yes, those fixtures are enabled conditionally because of partitioning support.

https://github.com/PostgREST/postgrest/blob/e977032847fc066a939918af2c3f5aa6e6555eb6/test/spec/fixtures/schema.sql#L2304-L2312

To solve this I think you'd need to add new fixtures

timabdulla:
Got it, I'll look into that, thanks.

timabdulla:
Would love to hear review comments from @steve-chavez and @wolfgangwalther when you all get a chance! Many thanks.

test/spec/Feature/Query/AggregateFunctionsSpec.hs
```diff
@@ -0,0 +1,70 @@
...
+          [json|[{ "total_sold": 4, "car_model_name": "Murcielago"}, { "total_sold": 16, "car_model_name": "DeLorean" }]|] { matchHeaders = [matchContentTypeJson] }
+      it "allows you to cast the result of the aggregate" $
+        get "/car_model_sales?select=total_sold:quantity.sum()::text,car_model_name" `shouldRespondWith`
+          [json|[{ "total_sold": "4", "car_model_name": "Murcielago"}, { "total_sold": "16", "car_model_name": "DeLorean" }]|] { matchHeaders = [matchContentTypeJson] }
+      it "allows you to cast the input argument of the aggregate" $
+        get "/trash_details?select=jsonb_col->>key::integer.sum()" `shouldRespondWith`
+          [json|[{"sum": 24}]|] { matchHeaders = [matchContentTypeJson] }
+      it "allows the combination of an alias, a before cast, and an after cast" $
+        get "/trash_details?select=s:jsonb_col->>key::integer.sum()::text" `shouldRespondWith`
+          [json|[{"s": "24"}]|] { matchHeaders = [matchContentTypeJson] }
```
steve-chavez, line 63: This one is impressive! :exploding_head: 
```bash
curl "localhost:3000/trash_details?select=s:jsonb_col->>key::integer.sum()::text"
[{"s":"24"}]
```

test/spec/Feature/Query/AggregateFunctionsSpec.hs
```diff
@@ -0,0 +1,70 @@
...
+      it "supports use of aggregates on RPC functions that return table values" $
+        get "/rpc/getallprojects?select=id.max()" `shouldRespondWith`
+          [json|[{"max": 5}]|] { matchHeaders = [matchContentTypeJson] }
+    context "performing aggregations on spreaded fields from an embedded resource" $ do
+      xit "supports the use of aggregates on spreaded fields" $
+        get "/car_models?select=car_brand_name,...car_model_sales(quantity.sum())" `shouldRespondWith`
+          [json|[{"car_brand_name": "dog"}]|] { matchHeaders = [matchContentTypeJson] }
```
steve-chavez, line 70: Since this one is marked as [pending](https://hackage.haskell.org/package/hspec-2.11.7/docs/Test-Hspec.html#v:xit), is that intentional?

timabdulla: Hoisting works, but the test here is broken because this relationship doesn't actually support spreads at all, with or without aggregates.

timabdulla: e.g.

```curl --location 'http://localhost:3000/contract?select=tournament%2C...player_view(birth_date.min())'```

**Comment by timabdulla** [comment]
@steve-chavez thanks for your review/comments! any other thoughts or things I should look at, or are we getting close to merge-ready?

**Comment by steve-chavez** [comment]
@timabdulla Sorry for the late reply. This is an interesting feature, though I'm not sure of all the implications yet. That being said, we can merge now and improve it later if issues arise.

Please rebase it and make sure the test pass. Then I'll give it one final pass!

File: src/PostgREST/Query/SqlFragment.hs [hunk 6]
```diff
@@ -392,17 +420,40 @@ pgFmtJsonPath = \case
...
+    noRelSelectsAreAggregated = all (\case Spread sels _ -> all (isNothing . ssSelAggFunction) sels; _ -> True) relSelect
+    groupTermsFromSelect = mapMaybe (pgFmtGroup qi) select
+    groupTermsFromRelSelect = mapMaybe groupTermFromRelSelectField relSelect
+    groupTerms = groupTermsFromSelect ++ groupTermsFromRelSelect
+
+groupTermFromRelSelectField :: RelSelectField -> Maybe SQL.Snippet
+groupTermFromRelSelectField (JsonEmbed { rsSelName }) =
+  Just $ pgFmtIdent rsSelName
+groupTermFromRelSelectField (Spread { rsSpreadSel, rsAggAlias }) =
+  if null groupTerms
```
steve-chavez, line 442: @timabdulla I've helped by making the tests pass. However seems there are parts that are still untested. Code coverage mention this line and some others below.

test/spec/Feature/Query/AggregateFunctionsSpec.hs
steve-chavez: I've added tests for the hoisting.

test/spec/Feature/Query/AggregateFunctionsSpec.hs
```diff
@@ -0,0 +1,143 @@
...
+        get "/trash_details?select=jsonb_col->>key::integer.sum()" `shouldRespondWith`
+          [json|[{"sum": 24}]|] { matchHeaders = [matchContentTypeJson] }
+      it "allows the combination of an alias, a before cast, and an after cast" $
+        get "/trash_details?select=s:jsonb_col->>key::integer.sum()::text" `shouldRespondWith`
+          [json|[{"s": "24"}]|] { matchHeaders = [matchContentTypeJson] }
+      it "supports use of aggregates on RPC functions that return table values" $
+        get "/rpc/getallprojects?select=id.max()" `shouldRespondWith`
+          [json|[{"max": 5}]|] { matchHeaders = [matchContentTypeJson] }
+
+    context "performing hoisted aggregations" $ do
```
steve-chavez, line 95: @timabdulla I've added these tests to prove hoisting works. Now I do realize these would be better with spreading combined. But as you've noticed this doesn't work for to-many relationships.

I think we can solve that on another PR (https://github.com/PostgREST/postgrest/issues/2730 needs it too). The  output is still useful as it is.

src/PostgREST/Query/SqlFragment.hs
steve-chavez: codecov also warns some fields being unused but those are not critical.

src/PostgREST/Plan/Types.hs
```diff
@@ -65,3 +70,37 @@ data CoercibleOrderTerm
     , coNullOrder :: Maybe OrderNulls
     }
   deriving (Eq, Show)
+
+data CoercibleSelectField = CoercibleSelectField
+  { csField       :: CoercibleField
+  , csAggFunction :: Maybe AggregateFunction
+  , csAggCast     :: Maybe Cast
+  , csCast        :: Maybe Cast
+  , csAlias       :: Maybe Alias
```
steve-chavez, line 79: codecov also warns of some fields being unused but those are not critical.

src/PostgREST/Query/SqlFragment.hs
timabdulla: I will add these tests, thanks!

timabdulla:
@steve-chavez thanks for the help, Steve! I will add those missing tests and then ping you once they are in.

test/spec/Feature/Query/AggregateFunctionsSpec.hs
steve-chavez: Tracking spreading for to-many relationships on https://github.com/PostgREST/postgrest/issues/3041

timabdulla:
Hey @steve-chavez,

I think I added the missing test that should hopefully fix the coverage issue. Note that the very helpful tests that you added do not actually cover hoisting, as hoisting is only used in the case of spreads; without spreads, aggregates are applied within the context the joined table (and thus not hoisted.)

I've added a test to cover hoisting/spreads, but I'm afraid hoisting isn't as useful as I'd hoped without support for has-many spreads. That said, I think we can get an easy win once has-many spreads are supported.

I think adding support for domain reps would be good in the future too, but there are some conflicting semantics that need to be worked out.

timabdulla:
@steve-chavez seems like the latest tests fixed the code coverage issue and all checks are now passing. Two questions:

1. Is the typical procedure to squash the PR commits before merge?
2. Is there anything else needed before merge?

steve-chavez:
@timabdulla Great work on adding the remaining tests!

> Is the typical procedure to squash the PR commits before merge?

Yes, I'll squash here. Makes sense when all the commits are related to the same feature.

> Is there anything else needed before merge?

- I was thinking about enabling this feature with `plan_filter.statement_cost_limit` for users that want to control the cost of the generated queries. But that doesn't seem right since we would depend on an specific extension. Additionally it would be messy in the code. 
- Thinking more about it, even regular resource embedding can be expensive. Users that want full control maybe want to disable it altogether. Which reminds of the `db-fk-embedding` config that Wolfgang proposed [here](https://github.com/PostgREST/postgrest/issues/2144#issuecomment-1031720960). WDYT? With that we can release this feature more safely.

> I've added a test to cover hoisting/spreads, but I'm afraid hoisting isn't as useful as I'd hoped without support for has-many spreads. That said, I think we can get an easy win once has-many spreads are supported.

- Once we've added support for has-many, would the response format change? I believe many users will rely on aggregate functions as it is and then it would become a painful breaking change.

timabdulla:
> The only way to make this safe would be with https://github.com/PostgREST/postgrest/pull/1710, applied in the way it's suggested https://github.com/PostgREST/postgrest/issues/915#issuecomment-667710634. WDYT?

What about adding a config option to this PR that makes aggregate functions opt-in by default? That should be a relatively small change (I think) that would allow users to decide whether it's safe in their situation (for instance, they may have a small dataset, or they may have other safeguards, like the `pg_plan_filter` extension or a `statement_timeout`.) Later on, more advanced scenarios could be supported, like the example you gave of using the request variables to determine what is safe or not.

> Once we've added support for has-many, would the response format change? I believe many users will rely on aggregate functions as it is and then it would become a painful breaking change.

No, it wouldn't, I don't think.

steve-chavez:
> What about adding a config option to this PR that makes aggregate functions opt-in by default?

@timabdulla Sounds good!

timabdulla:
> @timabdulla Sounds good! Maybe db-aggregate-functions?

added!

timabdulla:
I went ahead and squashed the commits. I think all the CI steps should pass now. By the way, is there a way to run all the CI steps in the nix shell environment? Like one command to do the specs, the IO tests, linting, etc?

steve-chavez:
@timabdulla There's `postgrest-git-hooks`:

```
postgrest-git-hooks -h
postgrest-git-hooks
Usage: postgrest-git-hooks [-h|--help] [--hook <HOOK>] [--] <operation> [<mode>]
        <operation>: Operation. Can be one of: 'disable', 'enable' and 'run'
        <mode>: Mode. Can be one of: 'basic' and 'full' (default: 'basic')
        -h, --help: Prints help
        --hook: Hook. Can be one of: 'pre-commit' and 'pre-push' (default: 'pre-commit')

Enable or disable git pre-commit and pre-push hooks.

Basic is faster and will only run:
  - pre-commit: postgrest-style
  - pre-push: postgrest-lint

Full takes a lot more time and will run:
  - pre-commit: postgrest-style && postgrest-lint
  - pre-push: postgrest-check

Changes made by postgrest-style will be staged automatically.

Example usage:
  postgrest-git-hooks disable
  postgrest-git-hooks enable basic
  postgrest-git-hooks enable full

The "run" operation and "--hook" argument are only used internally.
```

But maybe it makes sense to add a single `postgrest-test-full` command.

src/PostgREST/Error.hs
```diff
@@ -198,6 +199,9 @@ instance JSON.ToJSON ApiRequestError where
     (Just $ JSON.String $ T.decodeUtf8 ("Invalid preferences: " <> BS.intercalate ", " prefs))
     Nothing
 
+  toJSON AggregatesNotAllowed = toJsonPgrstError
+    ApiRequestErrorCode23 "Use of aggregate functions is not allowed" Nothing Nothing
```
steve-chavez, line 203: Hm, codecov is complaining about this being untested.

timabdulla: yes, I think that's because I enable this setting for the aggregate tests. I can add a test showing it fails when it's disabled, but is there a way to alter a setting on a per-test basis? I could only find examples on a per-module basis (e.g. https://github.com/PostgREST/postgrest/pull/2925/files#diff-f4402037436f4c7b85bee4c8c3d4820e194d9877598aee600b1f29af35abf543R247)

timabdulla: I suppose I could add in a test in another module to show that it fails?

steve-chavez: You can use the same module, there's this example for the `RollbackSpec`:

https://github.com/PostgREST/postgrest/blob/125f10a60f9928981224557464365a723f7fc232/test/spec/Main.hs#L248-L258

timabdulla: thanks! added.

src/PostgREST/Config/Database.hs
```diff
@@ -45,7 +45,8 @@ prefix = "pgrst."
 dbSettingsNames :: [Text]
 dbSettingsNames =
   (prefix <>) <$>
-  ["db_anon_role"
+  ["db_aggregates_enabled"
```
steve-chavez, line 48: This was missing for the [in-db config](https://postgrest.org/en/stable/references/configuration.html#in-database-configuration) to work.

Will refactor it soon so it can be missed.

src/PostgREST/Config.hs
```diff
@@ -137,7 +138,8 @@ toText conf =
   where
     -- apply conf to all pgrst settings
     pgrstSettings = (\(k, v) -> (k, v conf)) <$>
-      [("db-anon-role",              q . T.decodeUtf8 . fromMaybe "" . configDbAnonRole)
+      [("db-aggregates-enabled",         T.toLower . show . configDbAggregates)
```
steve-chavez, line 141: Changed the name so it's more consistent with other boolean configs

steve-chavez:
> But maybe it makes sense to add a single postgrest-test-full command.

@timabdulla Btw, forgot that `postgrest-check` already does that:

https://github.com/PostgREST/postgrest/blob/c3301a165377f218db0abed217cf29ea3bbd9254/nix/tools/devTools.nix#L66-L84

timabdulla:
> @timabdulla Awesome work 🔥 ! Would you help me document this feature?
> 
> It should be in a new page https://postgrest.org/en/latest/references/api.html
> 
> https://github.com/PostgREST/postgrest-docs/tree/main/docs/references/api

Thanks for all your help getting this over the line! It's very much appreciated. Sure thing - I will write something up shortly and leave it for your review.

jdgamble555:
Do the new aggregate functions include `sorting` by aggregate functions?


J

timabdulla:
@jdgamble555 no, not yet, but wouldn't be too tricky to add, I don't think.

With the size of this PR being so large, I felt that was better to defer to a future PR, though it's obviously a super helpful feature.

jaulz:
Does this PR actually support counting on embedded resources? Before it worked like this:
`GET /countries?select=name,languages(count)` 

So is there any "modern" equivalent after this PR? 😊

PR #2926, opened 2023-09-04
fix: HTTP status responses for upserts

### LINKED ISSUE #1070: Respond with HTTP 200 instead of 201 when no rows inserted

taimoorzaeem:
This might fix #1070.

#### Fixed

- `PUT` returns `201` instead of `200` when rows are inserted
- `POST` with `Prefer: resolution=merge-duplicates` returns `200` instead of `201` when no rows are inserted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #3015, Fix unnecessary count() on RPC returning single - @steve-chavez
+ - #1070, Fix HTTP status responses for upserts - @taimoorzaeem
+   + `PUT` returns `201` instead of `200` when rows are inserted
+   + `POST` with `Prefer: resolution=merge-duplicates` returns `200` instead of `201` when no rows are inserted
 
 ### Changed
 
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -87,7 +87,8 @@ mutatePlanToQuery (Insert mainQi iCols body onConflct putConditions returnings _
   "INSERT INTO " <> fromQi mainQi <> (if null iCols then " " else "(" <> cols <> ") ") <>
   fromJsonBodyF body iCols True False applyDefaults <>
   -- Only used for PUT
-  (if null putConditions then mempty else "WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "pgrst_body") <$> putConditions)) <>
+  (if null putConditions then mempty else "WHERE " <> addConfigPgrstInserted True <> " AND " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "pgrst_body") <$> putConditions)) <>
+  (if null putConditions && mergeDups then "WHERE " <> addConfigPgrstInserted True else mempty) <>
   maybe mempty (\(oncDo, oncCols) ->
     if null oncCols then
       mempty
@@ -98,11 +99,12 @@ mutatePlanToQuery (Insert mainQi iCols body onConflct putConditions returnings _
       MergeDuplicates  ->
         if null iCols
            then "DO NOTHING"
-           else "DO UPDATE SET " <> intercalateSnippet ", " ((pgFmtIdent . cfName) <> const " = EXCLUDED." <> (pgFmtIdent . cfName) <$> iCols)
+           else "DO UPDATE SET " <> intercalateSnippet ", " ((pgFmtIdent . cfName) <> const " = EXCLUDED." <> (pgFmtIdent . cfName) <$> iCols) <> (if null putConditions && not mergeDups then mempty else "WHERE " <> addConfigPgrstInserted False)
     ) onConflct <> " " <>
-  returningF mainQi returnings
+    returningF mainQi returnings
   where
     cols = intercalateSnippet ", " $ pgFmtIdent . cfName <$> iCols
+    mergeDups = case onConflct of {Just (MergeDuplicates,_) -> True; _ -> False;}
 
 -- An update without a limit is always filtered with a WHERE
 mutatePlanToQuery (Update mainQi uCols body logicForest range ordts returnings applyDefaults)
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -24,6 +24,8 @@ module PostgREST.Query.SqlFragment
   , fromJsonBodyF
   , responseHeadersF
   , responseStatusF
+  , addConfigPgrstInserted
+  , currentSettingF
   , returningF
   , singleParameter
   , sourceCTE
@@ -432,6 +434,11 @@ responseHeadersF = currentSettingF "response.headers"
 
 responseStatusF :: SQL.Snippet
 responseStatusF = currentSettingF "response.status"
+
+addConfigPgrstInserted :: Bool -> SQL.Snippet
+addConfigPgrstInserted add =
+  let (symbol, num) =  if add then ("+", "0") else ("-", "-1") in
+  "set_config('pgrst.inserted', (coalesce(" <> currentSettingF "pgrst.inserted" <> "::int, 0) " <> symbol <> " 1)::text, true) <> '" <> num <> "'"
 
 currentSettingF :: SQL.Snippet -> SQL.Snippet
 currentSettingF setting =
--- a/src/PostgREST/Query/Statements.hs
+++ b/src/PostgREST/Query/Statements.hs
@@ -51,7 +51,7 @@ data ResultSet
   , rsGucStatus  :: Maybe Text
   -- ^ the HTTP status to be added to the response
   , rsInserted   :: Maybe Int64
-  -- ^ the number of rows inserted (Only used with PUT)
+  -- ^ the number of rows inserted (Only used for upserts)
   }
   | RSPlan BS.ByteString -- ^ the plan of the query
 
--- a/src/PostgREST/Query/Statements.hs
+++ b/src/PostgREST/Query/Statements.hs
@@ -50,15 +50,19 @@ data ResultSet
   -- ^ the HTTP headers to be added to the response
   , rsGucStatus  :: Maybe Text
   -- ^ the HTTP status to be added to the response
+  , rsInserted   :: Maybe Int64
+  -- ^ the number of rows inserted (Only used with PUT)
   }
   | RSPlan BS.ByteString -- ^ the plan of the query
 
 
-prepareWrite :: QualifiedIdentifier -> SQL.Snippet -> SQL.Snippet -> Bool -> MediaType -> MediaHandler ->
-                Maybe PreferRepresentation -> [Text] -> Bool -> SQL.Statement () ResultSet
-prepareWrite qi selectQuery mutateQuery isInsert mt handler rep pKeys =
+prepareWrite :: QualifiedIdentifier -> SQL.Snippet -> SQL.Snippet -> Bool -> Bool -> MediaType -> MediaHandler ->
+                Maybe PreferRepresentation -> Maybe PreferResolution -> [Text] -> Bool -> SQL.Statement () ResultSet
+prepareWrite qi selectQuery mutateQuery isInsert isPut mt handler rep resolution pKeys =
   SQL.dynamicallyParameterized (mtSnippet mt snippet) decodeIt
  where
+  checkUpsert snip = if isInsert && (isPut || resolution == Just MergeDuplicates) then snip else "''"
+  pgrstInsertedF = checkUpsert "nullif(current_setting('pgrst.inserted', true),'')::int"
   snippet =
     "WITH " <> sourceCTE <> " AS (" <> mutateQuery <> ") " <>
     "SELECT " <>
@@ -67,7 +71,8 @@ prepareWrite qi selectQuery mutateQuery isInsert mt handler rep pKeys =
       locF <> " AS header, " <>
       handlerF Nothing qi handler <> " AS body, " <>
       responseHeadersF <> " AS response_headers, " <>
-      responseStatusF  <> " AS response_status " <>
+      responseStatusF  <> " AS response_status, " <>
+      pgrstInsertedF <> " AS response_inserted " <>
     "FROM (" <> selectF <> ") _postgrest_t"
 
   locF =
@@ -87,7 +92,7 @@ prepareWrite qi selectQuery mutateQuery isInsert mt handler rep pKeys =
   decodeIt :: HD.Result ResultSet
   decodeIt = case mt of
     MTVndPlan{} -> planRow
-    _        -> fromMaybe (RSStandard Nothing 0 mempty mempty Nothing Nothing) <$> HD.rowMaybe (standardRow False)
+    _           -> fromMaybe (RSStandard Nothing 0 mempty mempty Nothing Nothing Nothing) <$> HD.rowMaybe (standardRow False)
 
 prepareRead :: QualifiedIdentifier -> SQL.Snippet -> SQL.Snippet -> Bool -> MediaType -> MediaHandler -> Bool -> SQL.Statement () ResultSet
 prepareRead qi selectQuery countQuery countTotal mt handler =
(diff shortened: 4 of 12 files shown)


src/PostgREST/Response.hs
```diff
@@ -153,9 +153,13 @@ singleUpsertResponse ctxApiRequest@ApiRequest{iPreferences=Preferences{..}} resu
   RSStandard {..} -> do
     let
       response = gucResponse rsGucStatus rsGucHeaders
+      returnStatus =
+        case (rsInserted, rsUpdated) of
+          (Just i, Just u) -> if i > u then HTTP.status201 else HTTP.status200
+          _                -> HTTP.status200
```
laurenceisla, line 159: I updated my suggestion here: https://github.com/PostgREST/postgrest/issues/1070#issuecomment-1705901014. I think that we don't need to know exactly how many rows are updated/inserted. Anyways, this comparison only worked when we had two different settings, so it does not work when working with one. Only needs to know if rsInserted > 0.

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -438,6 +441,12 @@ responseHeadersF = currentSettingF "response.headers"
 responseStatusF :: SQL.Snippet
 responseStatusF = currentSettingF "response.status"
 
+pgrstInsertedF :: SQL.Snippet
+pgrstInsertedF = "count(nullif(current_setting('pgrst.mutation'), 'inserted'))"
```
laurenceisla, line 445: Hmm... I think this `current_setting('pgrst.mutation')` only returns the last value that was set here? Or does it compare with every insert/update?

Also the `nullif(..., 'inserted')` means that it will turn all 'inserted' into `null` and won't count it, so it should be in `pgrstUpdatedF` instead (and vice versa).

Anyways, this would change with the suggestion I gave in the other comment.

test/spec/Feature/Query/UpsertSpec.hs
```diff
@@ -395,6 +400,7 @@ spec actualPgVersion =
               [json| [ { "name": "Java", "rank": 5 } ]|]
             `shouldRespondWith`
               [json| [ { "name": "Java", "rank": 5 } ]|]
+              { matchStatus = 201 }
```
laurenceisla, line 403: Since this is an update only, it should be `200` (the same for others).

src/PostgREST/Response.hs
laurenceisla: I saw your response there with your new approach, so you can ignore this suggestions now.

src/PostgREST/Response.hs
```diff
@@ -153,9 +153,13 @@ singleUpsertResponse ctxApiRequest@ApiRequest{iPreferences=Preferences{..}} resu
   RSStandard {..} -> do
     let
       response = gucResponse rsGucStatus rsGucHeaders
+      returnStatus =
+        case rsInserted of
+          Just i -> if i > 0 then HTTP.status201 else HTTP.status200
+          _      -> HTTP.status200
```
taimoorzaeem, line 159: I tried to debug the what's wrong by check the actual value of `i` here.
```suggestion
      returnStatus =
        case rsInserted of
          Just i -> if i > 0 then HTTP.mkStatus (fromIntegral i) "" else HTTP.status200
          _      -> HTTP.status200
```

After running the tests again, I am getting the value of `i` as `48` and `49`, which is why it always responds with `201` which is very strange. 
```
test/spec/Feature/Query/UpsertSpec.hs:400:13: 
  10) Feature.Query.UpsertSpec.UPSERT, with PUT, Updating row, ignores the Range header
       status mismatch:
         expected: 200
         but got:  48

test/spec/Feature/Query/UpsertSpec.hs:310:15: 
  3) Feature.Query.UpsertSpec.UPSERT, with PUT, Inserting row, succeeds on a partitioned table with composite pk
       status mismatch:
         expected: 201
         but got:  49

```

laurenceisla: Sorry for the late reply. Below is why it's giving those values:

src/PostgREST/Query/Statements.hs
```diff
@@ -49,15 +49,19 @@ data ResultSet
...
 
-prepareWrite :: SQL.Snippet -> SQL.Snippet -> Bool -> MediaType -> ResultAggregate ->
+prepareWrite :: SQL.Snippet -> SQL.Snippet -> Bool -> Bool -> MediaType -> ResultAggregate ->
                 Maybe PreferRepresentation -> [Text] -> Bool -> SQL.Statement () ResultSet
-prepareWrite selectQuery mutateQuery isInsert mt rAgg rep pKeys =
+prepareWrite selectQuery mutateQuery isInsert isPut mt rAgg rep pKeys =
   SQL.dynamicallyParameterized (mtSnippet mt snippet) decodeIt
  where
+  checkUpsert snip = if isInsert && isPut then snip else "''"
+  pgrstInsertedF = checkUpsert "current_setting('pgrst.inserted', true)"
```
laurenceisla, line 64: This needs to be converted to an `int`. Also needs the wrap with `nullif` or `''::int` may give errors.

```suggestion
  pgrstInsertedF = checkUpsert "nullif(current_setting('pgrst.inserted', true),'')::int"
```

Or maybe you could take it as a boolean (and change `rsInserted` to `Maybe Bool`).

```suggestion
  pgrstInsertedF = checkUpsert "current_setting('pgrst.inserted', true) = '1"
```

laurenceisla:
Also, I'm wondering if the `pgrst.inserted` guc could/should be initialized to '0' here:

https://github.com/PostgREST/postgrest/blob/0703f27d2b0ffe60edfc6053ea5df10a3ed2c0dd/src/PostgREST/Query.hs#L234-L236

It would avoid using `coalesce(...,0)` in the query.

src/PostgREST/Query/Statements.hs
taimoorzaeem: > This needs to be converted to an `int`. Also needs the wrap with `nullif` or `''::int` may give errors.

Hmm, I have made these changes, however `PUT` is still not responding with `201` on inserts. It seems very strange since I can't figure out anything wrong in the implementation.

taimoorzaeem:
With this change, the cost of single upsert becomes:
```
test/spec/Feature/Query/PlanSpec.hs:195:19: 
  5) Feature.Query.PlanSpec.spec, writes plans, outputs the total cost for a single upsert
       expected: 1.29
        but got: 3.55
```
@laurenceisla @steve-chavez Is it still worth it?

src/PostgREST/Query/Statements.hs
laurenceisla: Ah, this should be inside `singleUpsertResponse` instead of `updateResponse`.

steve-chavez:
> outputs the total cost for a single upsert

@taimoorzaeem To be extra sure, we would need to test upserts with more than one single row. These are done with [resolution=merge-duplicates](https://postgrest.org/en/stable/references/api/tables_views.html#upsert). If the extra cost is constant, then it would be worth it, I think.

So how about adding some more tests below here:

https://github.com/PostgREST/postgrest/blob/a9d6c318fdb1513d93c9b1911ed9cfaaa84c65dd/test/spec/Feature/Query/PlanSpec.hs#L183-L195

I'm thinking: 

```haskell
 it "outputs the total cost for an upsert with 10 rows" $ do

 it "outputs the total cost for an upsert with 100 rows" $ do

 it "outputs the total cost for an upsert with 1000 rows" $ do
```

There should be a helper function that generates the payload. Rough idea:

```
$ postgrest-repl

ghci> [("Lang " <> show i, i) | i <- [20..120] ] :: [(Text, Int)]

[("Lang 20",20),("Lang 21",21)....
```

And maybe using [Aeson.toJSONList](https://hackage.haskell.org/package/aeson-2.2.1.0/docs/Data-Aeson.html#v:toJSONList) to convert to JSON. 

---

These tests can be added in another PR. In fact that would be better to see the resulting costs there to compare.

Sometimes we do commits that only add tests, like https://github.com/PostgREST/postgrest/commit/add10bd10ca9e40c1b31601a0be0b1083fe04aae

taimoorzaeem:
> These tests can be added in another PR. In fact that would be better to see the resulting costs there to compare.

Alright. I'll open another PR to add these tests.

laurenceisla:
As far as I can see, there were no existing tests for `Prefer: merge-duplicates` where all the elements in the payload already exists, thus generating only updates and no inserts.

Adding this test to the beginning of `UpsertSpec.hs` makes the test fail:

```haskell
it "UPDATEs rows and returns 200 on pk conflict on all the items" $
  request methodPost "/tiobe_pls" [("Prefer", "return=representation"), ("Prefer", "resolution=merge-duplicates")]
    [json| [
      { "name": "Python", "rank": 6 },
      { "name": "Java", "rank": 2 },
      { "name": "C", "rank": 1 }
    ]|] `shouldRespondWith` [json| [
      { "name": "Python", "rank": 6 },
      { "name": "Java", "rank": 2 },
      { "name": "C", "rank": 1 }
    ]|]
    { matchStatus = 200
    , matchHeaders = ["Preference-Applied" <:> "resolution=merge-duplicates, return=representation", matchContentTypeJson]
    }
```

All of those already exist in the database, so a `200` should be returned since no insert was done. The good thing is that the changes you made in SQL already allow this, only needs implementation for Merge Duplicates as you did with PUT.

So far so good, we're almost there! This wasn't a simple issue by any means.

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -439,6 +442,12 @@ responseHeadersF = currentSettingF "response.headers"
 responseStatusF :: SQL.Snippet
 responseStatusF = currentSettingF "response.status"
 
+coalesceAddPgrstInsertedF :: SQL.Snippet
+coalesceAddPgrstInsertedF = "set_config('pgrst.inserted', (coalesce(" <> currentSettingF "pgrst.inserted" <> "::int, 0) + 1)::text, true) <> '0'"
+
+coalesceSubPgrstInsertedF :: SQL.Snippet
+coalesceSubPgrstInsertedF = "set_config('pgrst.inserted', (coalesce(" <> currentSettingF "pgrst.inserted" <> "::int, 0) - 1)::text, true) <> '-1'"
+
```
laurenceisla, line 450: This could be refactored to avoid repetition, something like:

```haskell
setConfigPgrstInserted add =
  let (sym, num) =  if add then ("+", "0") else ("-", "-1") in
  "set_config('pgrst.inserted', (coalesce(" <> currentSettingF "pgrst.inserted" <> "::int, 0) " <> sym <> " 1)::text, true) <> '" <> num <> "'"'"
```

test/spec/Feature/Query/PlanSpec.hs
```diff
@@ -192,7 +192,7 @@ spec actualPgVersion = do
       liftIO $ do
         resHeaders `shouldSatisfy` elem ("Content-Type", "application/vnd.pgrst.plan+json; for=\"application/json\"; charset=utf-8")
         resStatus `shouldBe` Status { statusCode = 200, statusMessage="OK" }
-        totalCost `shouldBe` 1.29
+        totalCost `shouldBe` 3.55
```
laurenceisla, line 195: I was wondering why it only changed for single upserts and not multiple ones. The `resolution=merge-duplicates` was not being applied in these tests, I commented in that PR (may need to be fixed in a new one): https://github.com/PostgREST/postgrest/pull/3005/files#r1366092283

Good news is the costs are still constant.

laurenceisla:
Lastly, what's missing from the original issue is to verify that:

- It returns `200` when the payload is `[]`
- It returns `200` when `Prefer: ignore-duplicates` is sent and all the elements in the body are duplicates.

taimoorzaeem:
@laurenceisla 
> It returns `200` when `Prefer: ignore-duplicates` is sent and all the elements in the body are duplicates.

Hmm, for `ignore-duplicates`, it's getting tricky.
```haskell
mutatePlanToQuery :: Maybe PreferResolution -> MutatePlan -> SQL.Snippet
mutatePlanToQuery resolution (Insert mainQi iCols body onConflct putConditions returnings _ applyDefaults) =
  "INSERT INTO " <> fromQi mainQi <> (if null iCols then " " else "(" <> cols <> ") ") <>
  fromJsonBodyF body iCols True False applyDefaults <>
  -- Only used for PUT
  (if null putConditions then mempty else "WHERE " <> addConfigPgrstInserted True <> " AND " <> intercalateSnippet " AND " (pgFmtLogicTree (QualifiedIdentifier mempty "pgrst_body") <$> putConditions)) <>
  (if null putConditions && isJust resolution then "WHERE " <> addConfigPgrstInserted True else mempty) <>
  maybe mempty (\(oncDo, oncCols) ->
    if null oncCols then
      mempty
    else
      " ON CONFLICT(" <> intercalateSnippet ", " (pgFmtIdent <$> oncCols) <> ") " <> case oncDo of
      IgnoreDuplicates ->
        "DO NOTHING"
      MergeDuplicates  ->
        if null iCols
           then "DO NOTHING"
           else "DO UPDATE SET " <> intercalateSnippet ", " ((pgFmtIdent . cfName) <> const " = EXCLUDED." <> (pgFmtIdent . cfName) <$> iCols) <> (if null putConditions && isNothing resolution then mempty else "WHERE " <> addConfigPgrstInserted False)
    ) onConflct <> " " <>
    returningF mainQi returnings
```
I first add 1 when row is inserted, but if there is `resolution=ignore-duplicates`, I need to subtract that.
In other words, I need to do something like this:
```haskell
 IgnoreDuplicates ->
        "DO NOTHING" <> " WHERE " <> addConfigPgrstInserted False -- not possible i guess
```

laurenceisla:
> Hmm, for ignore-duplicates, it's getting tricky.

True that. I noticed that [PostgreSQL allows](https://www.postgresql.org/docs/current/sql-insert.html) `ON CONFLICT WHERE index_predicate ...`, but it didn't work for me to set the `pgrst.inserted` value there. It may work in a different way.

We could leave that outside of this PR to check for a better solution later on, it won't close the issue completely but it's good progress.

CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2824, Fix regression by reverting fix that returned 206 when first position = length in a `Range` header - @laurenceisla, @strengthless
  - #3015, Fix unnecessary count() on RPC returning single - @steve-chavez
 
+### Fixed
+
+ - #2824, Fix range request with 0 rows and 0 offset return status 416 - @strengthless
+ - #1070, Fix PUT responding with 200 when rows are inserted - @taimoorzaeem
```
laurenceisla, line 26:
```suggestion
 - #1070, Fix PUT responding with 200 when rows are inserted - @taimoorzaeem
```

This must be under a single "Fixed" subtitle.  Rebase the latest change in main first and then check that only the current fix is added.

laurenceisla: Also, it needs to mention that `resolution=merge-duplicates` now returns `200` when no insert was done.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -82,12 +82,13 @@ getSelectsJoins rr@(Node ReadPlan{select, relName, relToParent=Just rel, relAggA
   in
   (if null select && null forest then selects else sel:selects, joi:joins)
 
-mutatePlanToQuery :: MutatePlan -> SQL.Snippet
-mutatePlanToQuery (Insert mainQi iCols body onConflct putConditions returnings _ applyDefaults) =
+mutatePlanToQuery :: Bool -> MutatePlan -> SQL.Snippet
+mutatePlanToQuery mergeDups (Insert mainQi iCols body onConflct putConditions returnings _ applyDefaults) =
```
laurenceisla, line 86: Maybe I'm being too nitpicky, but we could check if the preference is applied without passing `mergeDups` as a parameter, by using `Insert{onConflict}`:

```haskell
where
  -- ...
  mergeDups = case onConflct of {Just (MergeDuplicates, _) -> True; _ -> False;}
```

It would make the `mutatePlanToQuery` function cleaner in my eyes, WDYT?

---

Those are the only reviews I have. After this, then this PR is ready to be merged.

taimoorzaeem: @laurenceisla No, you're not being nitpicky. It is actually a really nice refactor. Code cleanup should always be done, better now than later.

laurenceisla (approved):
Good job! I clarified a message and rebased against the changes made in main to be ready to merge.

PR #2928, opened 2023-09-07
feat: implement JWT caching

### LINKED ISSUE #2698: JWT caching
In our set-up we use asymmetric encryption to authenticate JWTs. As crypto operations are quite resource intensive we would like to cache results of JWT validation.

From my limited understanding of Haskell code JWTs are validated and parsed upon every request.

Do you think it would be a good idea to provide a time limited cache of JWT validation/parsing results?

taimoorzaeem:
This should fix #2698


src/PostgREST/AppState.hs
```diff
@@ -87,6 +92,14 @@ data AppState = AppState
   , stateRetryNextIn              :: IORef Int
   -- | Logs a pool error with a debounce
   , debounceLogAcquisitionTimeout :: IO ()
+  -- | JWT Cache
+  , jwtCache                      :: C.Cache ByteString AuthResult
+  }
```
taimoorzaeem, line 97: @steve-chavez Here, I have made `jwtCache` a part of `AppState`. Is `AppState` persisted across two subsequent requests from the `spec-tests`? 

Currently I have added a new module `JwtCaching.hs` in the spec-tests where I am making two requests with the same `JWT` token. The `JWT` token is parsed in both requests and the cache doesn't respond with the inserted token in the first request. Hence, I was thinking maybe `AppState` is reset and loses it's contents (jwtCache).

steve-chavez:
@taimoorzaeem Would it be possible to separate the `Server-Timing` feature in another PR?

That one could be merged standalone. It would the caching feature easier to review.

taimoorzaeem:
> @taimoorzaeem Would it be possible to separate the `Server-Timing` feature in another PR?

Yes, that actually would be much better. Let me work on that.

steve-chavez:
Let's consider warmup (https://github.com/PostgREST/postgrest/pull/2937#issuecomment-1720881628) for testing.

I guess first we need two requests and then measure that the third will have lower duration thanks to jwt-caching? Hm, we might need to test this manually first.

taimoorzaeem:
```
PGRST_JWT_SECRET="reallyreallyreallyreallyverysafe" PGRST_DB_PLAN_ENABLED="true" PGRST_JWT_CACHING="true" postgrest-with-postgresql-15 postgrest-run
```

```
export JWT="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjk5OTk5OTk5OTksInJvbGUiOiJwb3N0Z3Jlc3RfdGVzdF9hdXRob3IiLCJpZCI6Impkb2UiLCJhdWQiOiJ5b3VyYXVkaW
VuY2UifQ.fJ4tLKSmolWGWehWN20qiU9dMO-WY0RI2VvacL7-ZGo"

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=495.1

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=17.7

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=10.9
```

@steve-chavez The caching feature is surely working as shown in local testing above. However the same is not the case when I test it with `spec-tests`. Maybe there is something missing in spec-tests that I wrote?

steve-chavez:
@taimoorzaeem Around 90% time reduction compared to https://github.com/PostgREST/postgrest/pull/2937#issuecomment-1716617086 :tada: :tada: 

```
curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Date: Tue, 19 Sep 2023 19:53:09 GMT
Server: postgrest/11.2.0 (c5a9ed1)
Content-Range: */*
Content-Location: /authors_only
Server-Timing: jwt;dur=14.1
```

At startup, sometimes I've gotten higher values than before:

```
$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
Server-Timing: jwt;dur=516.8

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
Server-Timing: jwt;dur=903.6

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
Server-Timing: jwt;dur=402.8
```

But the next durations are always low (~15), so no problem there.

test/spec/Feature/Auth/JwtCachingSpec.hs
```diff
@@ -0,0 +1,39 @@
...
+spec = describe "JWT Caching" $ do
+  let auth = authHeaderJWT "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoicG9zdGdyZXN0X3Rlc3RfYXV0aG9yIn0.Xod-F15qsGL0WhdOCr2j3DdKuTw9QJERVgoFD3vGaWA"
+
+  it "jwt Cache warmup" $
+    request methodPost "/rpc/privileged_hello" [auth] [json|{"name": "jdoe"}|]
+      `shouldRespondWith` [json|"Privileged hello to jdoe"|]
+      { matchStatus = 200
+      , matchHeaders = ["Server-Timing" <:> "jwt;dur="
+                       , matchContentTypeJson]
+      }
```
steve-chavez, line 23: >@steve-chavez The caching feature is surely working as shown in local testing above. However the same is not the case when I test it with spec-tests. Maybe there is something missing in spec-tests that I wrote?

Hm, I see on CI that the tests are working fine:

```haskell
  test/spec/Feature/Auth/JwtCachingSpec.hs:19:7: 
  1) Feature.Auth.JwtCachingSpec, JWT Caching, jwt Cache warmup
       missing header:
         Server-Timing: jwt;dur=
       the actual headers were:
         Content-Range: 0-0/*
         Server-Timing: jwt;dur=148.4
```

See `Server-Timing: jwt;dur=148.4`, but you're not matching the value correctly. You'd only need to check that the value is present with

https://github.com/PostgREST/postgrest/blob/c83befee454bcba29f4e6010f104d260a3058ae0/test/spec/SpecHelper.hs#L49-L53

steve-chavez: Now I guess you mean why the duration doesn't decrease on the spec tests.

> My guess is that spec-tests resets the AppState after a single request, so that Cache is "reset" after every request.

Yeah, there's no real webserver running in the spec tests. I'd suggest moving these tests to io tests in python.

test/io/test_io.py
```diff
@@ -1095,3 +1095,55 @@ def test_fail_with_automatic_recovery_disabled_and_terminated_using_query(defaul
...
+        second_dur_text = postgrest.session.get(
+            "/authors_only", headers=headers
+        ).headers["Server-Timing"]
+
+        first_dur = float(first_dur_text[8:])  # skip "jwt;dur="
+        second_dur = float(second_dur_text[8:])
+
+        # their difference should be atleast 200, implying
+        # that JWT Caching is working as expected
+        assert (first_dur - second_dur) > 200.0
```
steve-chavez, line 1126: Clever test! :brain:

src/PostgREST/Auth.hs
```diff
@@ -107,16 +106,49 @@ middleware appState app req respond = do
...
+
+    (False, True)  -> do
+          authResult <- getJWTFromCache appState token parseJwt
+          let req' = req { Wai.vault = Wai.vault req & Vault.insert authResultKey authResult }
+          app req' respond
+
+    (False, False) -> do
+          authResult <- parseJwt
+          let req' = req { Wai.vault = Wai.vault req & Vault.insert authResultKey authResult }
+          app req' respond
```
steve-chavez, line 130: Hm, can this part be deduplicated? The `app req' respond` looks redundant.

src/PostgREST/Auth.hs
```diff
@@ -107,16 +106,49 @@ middleware appState app req respond = do
...
+
+-- Used to extract JWT exp claim and add to JWT Cache
+getTimeSpec :: AuthResult -> Maybe TimeSpec
+getTimeSpec res = do
+  let sciToInt = fromMaybe 0 . Sci.toBoundedInteger
+  expireJSON <- KM.lookup "exp" (authClaims res)
+  case expireJSON of
+    JSON.Number seconds -> Just $ TimeSpec (sciToInt seconds) 0
+    _                   -> Just $ TimeSpec 0 0 -- set timeSpec to 0 so it expires immediately, hence not cached
```
steve-chavez, line 139: The `0` here means that when there's no exp claim, the JWT will not be cached right?

How about an io test that confirms this. I guess the test would be that the duration is not down by 200 on the second request?

This would appease codecov here too.

src/PostgREST/CLI.hs
```diff
@@ -205,6 +205,9 @@ exampleConfigFile =
       |# jwt-secret = "secret_with_at_least_32_characters"
       |jwt-secret-is-base64 = false
       |
+      |## Allows JWT caching to avoid validation on every request
+      |# jwt-caching = false
```
steve-chavez, line 209: Good thinking in making this false by default. For next major version we can make it true.

src/PostgREST/Auth.hs
steve-chavez: > If there is no exp claim, a JWT is valid infinitely, iirc. Imho, it should be chached without expiry, too.

Hm, one advantage of only caching JWTs with `exp` claims is that the memory consumption will always be bounded.

If a server gets too many infinitely valid JWTs, we could hit OOM errors and then crash.

How should we deal with this case? Should we force a cache lifetime for infinitely valid JWTs?

steve-chavez: Cool, agreed. 

>  jwt-cache-for = seconds

An `exp` might have a shorter time than the config, thus last less in the cache. So I think a more apt name would be: `jwt-cache-max-lifetime`. `0` for disabling would still stand.

taimoorzaeem: Just to be clear, is the `0` here relative to current time? I am not sure because `exp` is relative to the `POSIX` time AFAIK.

steve-chavez: Hm, no. Should just be a time in seconds for the cache max lifetime. So a lifetime of 24 hours would be:

```
jwt-cache-max-lifetime = 86400
```

steve-chavez:
@taimoorzaeem Really cool feature! Don't forget to add an entry to the CHANGELOG.

steve-chavez (approved):

steve-chavez (changes requested):
My bad, I see the memory test failing. I need to give it another review.

src/PostgREST/Auth.hs
```diff
@@ -107,16 +106,52 @@ middleware appState app req respond = do
...
+getTimeSpec res maxLifetime = do
+  expireJSON <- KM.lookup "exp" (authClaims res)
+  case expireJSON of
+    JSON.Number seconds -> Just $ unsafePerformIO (posixToCurrentTime seconds)
+    _                   -> Just $ CL.TimeSpec (fromIntegral maxLifetime :: Int64) 0 -- set timeSpec to 0 so it expires immediately, hence not cached
+
+posixToCurrentTime :: Sci.Scientific -> IO CL.TimeSpec
+posixToCurrentTime secs = do
+  let sciToInt = fromMaybe 0 . Sci.toBoundedInteger
+  cur <- CL.getTime CL.Monotonic
```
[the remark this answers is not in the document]
steve-chavez: With the above it might be possible to remove the `IO` from the function too.

src/PostgREST/Auth.hs
```diff
@@ -107,16 +106,52 @@ middleware appState app req respond = do
...
+
+  app req' respond
+
+-- Used to extract JWT exp claim and add to JWT Cache
+getTimeSpec :: AuthResult -> Int -> Maybe CL.TimeSpec
+getTimeSpec res maxLifetime = do
+  expireJSON <- KM.lookup "exp" (authClaims res)
+  case expireJSON of
+    JSON.Number seconds -> Just $ unsafePerformIO (posixToCurrentTime seconds)
+    _                   -> Just $ CL.TimeSpec (fromIntegral maxLifetime :: Int64) 0 -- set timeSpec to 0 so it expires immediately, hence not cached
```
steve-chavez, line 136: Hm, this part is not clear to me. So as we discussed before, if there's no `exp`claim, the cache lifetime should be `maxLifetime` and not 0 right?

steve-chavez: Adding a test for ensuring caching works for a JWT without exp would also be good.

taimoorzaeem: Oh, my bad. It is actually `maxLifetime`. I forgot to remove the comment after changing that.

test/io/test_io.py
```diff
@@ -1095,3 +1095,93 @@ def test_fail_with_automatic_recovery_disabled_and_terminated_using_query(defaul
...
+
+    headers = jwtauthheader(
+        {
+            "role": "postgrest_test_author",
+            "exp": int(
+                (datetime.now(timezone.utc) + timedelta(minutes=30)).timestamp()
+            ),
+        },
+        SECRET,
+    )
```
taimoorzaeem, line 1119: @steve-chavez I am adding `exp` here because we don't have one by default as far as I understand. However, in other tests where I didn't add any `exp` (see other test cases below in `test_io.py`) the following code that should execute never gets executed:
```haskell
expireJSON <- lookup "exp" (authClaims res)
case expireJSON of
  JSON.Number seconds <- Just $ CL.TimeSpec ...
  _                   <- Just $ CL.TimeSpec ... -- this never gets executed

```
Somehow, it is always finding an `exp` key in `authClaims`. Do we have an `exp` by default in our tokens?

src/PostgREST/Auth.hs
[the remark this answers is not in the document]
taimoorzaeem: @steve-chavez I have changed the code and now using the time from the `AppState` but memory tests are still failing.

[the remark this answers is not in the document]
steve-chavez: @taimoorzaeem Try updating the `16M` to `23M` here:
https://github.com/PostgREST/postgrest/blob/ce96c88ba0e64e9d0868479365523a45417d30da/test/memory/memory-tests.sh#L105

You could also reproduce this locally with `postgrest-test-memory`, check https://github.com/PostgREST/postgrest/tree/main/nix#working-with-nix-shell-and-the-postgrest-utility-scripts

test/memory/memory-tests.sh
```diff
@@ -102,7 +102,7 @@ postJsonArrayTest(){
 
 echo "Running memory usage tests.."
 
-jsonKeyTest "1M" "POST" "/rpc/leak?columns=blob" "16M"
+jsonKeyTest "1M" "POST" "/rpc/leak?columns=blob" "23M"
```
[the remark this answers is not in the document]
steve-chavez: Yeah, I don't understand why that happened but it didn't seem that much extra memory to me. 

Another thing that's puzzling is that the memory tests don't have the jwt caching enabled.

Maybe the [Data.Cache](https://hackage.haskell.org/package/cache-0.1.3.0/docs/Data-Cache.html) library does some allocation by default?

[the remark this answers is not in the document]
steve-chavez: Just tried `postgrest-test-memory` locally on 90e3a5e29fa1f73807e474dc4fc1342667feffce

```
Running memory usage tests..
not ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(22,978,640 bytes) is more than 16M
ok 2 - POST /leak?columns=blob: with a json key of 1M the memory usage(14,167,624 bytes) is less than 16M
ok 3 - PATCH /leak?id=eq.1&columns=blob: with a json key of 1M the memory usage(14,203,336 bytes) is less than 16M
ok 4 - POST /rpc/leak?columns=blob: with a json key of 10M the memory usage(23,663,344 bytes) is less than 44M
ok 5 - POST /leak?columns=blob: with a json key of 10M the memory usage(23,568,720 bytes) is less than 44M
ok 6 - PATCH /leak?id=eq.1&columns=blob: with a json key of 10M the memory usage(23,610,888 bytes) is less than 44M
ok 7 - POST /rpc/leak?columns=blob: with a json key of 50M the memory usage(65,453,568 bytes) is less than 172M
ok 8 - POST /leak?columns=blob: with a json key of 50M the memory usage(65,355,608 bytes) is less than 172M
ok 9 - PATCH /leak?id=eq.1&columns=blob: with a json key of 50M the memory usage(65,398,104 bytes) is less than 172M
ok 10 - POST /perf_articles?columns=id,body: with a json payload of 32K that has 1000 array values the memory usage(13,176,328 bytes) is less than 14M
ok 11 - POST /perf_articles?columns=id,body: with a json payload of 329K that has 10000 array values the memory usage(13,484,600 bytes) is less than 14M
ok 12 - POST /perf_articles?columns=id,body: with a json payload of 3.4M that has 100000 array values the memory usage(16,688,424 bytes) is less than 24M
```

Seems the memory increase was not related to this PR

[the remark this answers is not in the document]
steve-chavez: Now on a6e3eda5b20817f134beb8d660029bd2875c280f, I got `14M` too:

```
Running memory usage tests..
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(14,271,344 bytes) is less than 23M
ok 2 - POST /leak?columns=blob: with a json key of 1M the memory usage(14,162,648 bytes) is less than 16M
ok 3 - PATCH /leak?id=eq.1&columns=blob: with a json key of 1M the memory usage(14,195,632 bytes) is less than 16M
ok 4 - POST /rpc/leak?columns=blob: with a json key of 10M the memory usage(23,667,280 bytes) is less than 44M
ok 5 - POST /leak?columns=blob: with a json key of 10M the memory usage(23,577,504 bytes) is less than 44M
ok 6 - PATCH /leak?id=eq.1&columns=blob: with a json key of 10M the memory usage(23,613,160 bytes) is less than 44M
ok 7 - POST /rpc/leak?columns=blob: with a json key of 50M the memory usage(65,462,040 bytes) is less than 172M
ok 8 - POST /leak?columns=blob: with a json key of 50M the memory usage(65,358,784 bytes) is less than 172M
ok 9 - PATCH /leak?id=eq.1&columns=blob: with a json key of 50M the memory usage(65,399,112 bytes) is less than 172M
ok 10 - POST /perf_articles?columns=id,body: with a json payload of 32K that has 1000 array values the memory usage(13,169,752 bytes) is less than 14M
ok 11 - POST /perf_articles?columns=id,body: with a json payload of 329K that has 10000 array values the memory usage(13,486,992 bytes) is less than 14M
ok 12 - POST /perf_articles?columns=id,body: with a json payload of 3.4M that has 100000 array values the memory usage(16,699,496 bytes) is less than 24M
```

And consistenly:

```
Running memory usage tests..
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(14,266,608 bytes) is less than 23M

Running memory usage tests..
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(14,260,440 bytes) is less than 23M

Running memory usage tests..
ok 1 - POST /rpc/leak?columns=blob: with a json key of 1M the memory usage(14,266,984 bytes) is less than 23M
```

So it seems to be some kinda flakiness on the CI runners.

steve-chavez:
Just benched jwt caching with the [setup](https://github.com/steve-chavez/benchmarks/blob/master/postgrest/deploy.nix) (t3a.nano pg + t3a.nano pgrst + m5a.16xlarge k6 client) and this [script](https://github.com/steve-chavez/benchmarks/blob/master/postgrest/k6/GETSingleJWT.js)

```
req/s
2477.315611/s  no JWT sent              (direct anon request)
1973.09754/s   JWT sent without caching (~20% decrease)
2432.181056/s  JWT sent with caching    (~1% decrease)
```

With jwt caching Perf is almost as good as with no JWT sent :tada: 

---

Note: CPU consumption didn't seem to vary that much, during the above load test:

```
%CPU  %MEM    
148.5   7.6  Without JWT caching
154.8   8.0  With JWT caching
```

~%6 less CPU usage. I wonder if it's us mostly consuming CPU or the Warp server.

PR #2934, opened 2023-09-11
refactor: move media type logic to Plan module

steve-chavez:
Part of https://github.com/PostgREST/postgrest/pull/2825.

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -15,6 +15,7 @@ module PostgREST.ApiRequest
   , Action(..)
   , Target(..)
   , Payload(..)
+  , PathInfo(..)
   , userApiRequest
   ) where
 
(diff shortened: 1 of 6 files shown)


steve-chavez:
According to the loadtest results we lose some throughput with this change https://github.com/PostgREST/postgrest/pull/2934/checks?check_run_id=16660131983

param | head | main
-- | -- | --
throughput | 307.90351639998954 | 315.3078228340249

Which is weird because it's the same logic, just on a different module. I've seen this case before, will check what's wrong with the Plan module.

steve-chavez:
Thanks a lot for the help Wolfgang,

> Always repeat those tests locally, while keeping load constant.

Yeah, actually I did that and I always got worse results for this PR. Now, after https://github.com/PostgREST/postgrest/pull/2940, I don't get bad results, in fact they seem slightly better:

- main: 565.19
- this branch: 574.37

Maybe there was some perf improvement on #2940 after all. Anyway, this is looking good to merge now. Postponing this change will make #2825 harder otherwise.

Issue #2771, opened 2023-04-27
`Server-Timing` response header


steve-chavez:
## Problem

Currently there's no way to debug the time we take to parse the request and the time we take to receive the query response. The [EXPLAIN execution plan](https://postgrest.org/en/stable/api.html#execution-plan), only obtains the query time, but not the aforementioned two.

## Solution

Add the standard `Server-Timing` header. This can communicate all types of metrics for a given request-response cycle. 

An example response would be like:

```http
GET /projects?id=eq.3

Server-Timing: jwt;dur=34, app;dur=47.2, db;dur=53, render;dur=15
```

Which would mean the jwt ("jwt") decoding time was 34 ms, the parsing + planning("app") took 47.2ms, the database ("db") query time was 53 ms, and we took 15 ms to obtain the query response("render", this is relevant to https://github.com/PostgREST/postgrest/issues/2770).

To avoid doing this work for all the requests:

- We can add `Server-Timing` whenever the `Prefer: server-timing=include` header is specified.
- Or enable it whenever [db-plan-enabled](https://postgrest.org/en/stable/configuration.html#db-plan-enabled) is turned on.

References:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing
- https://www.w3.org/TR/server-timing/

steve-chavez:
Seems this might be simple with the timeit package: https://hackage.haskell.org/package/timeit-1.0.0.0/docs/System-TimeIt.html. Since the functions are [so small](https://hackage.haskell.org/package/timeit-1.0.0.0/docs/src/System-TimeIt.html#timeItT) we could also vendor the functionality.

Got that from this blog post: https://chrisdone.com/posts/measuring-duration-in-haskell/

steve-chavez:
If parsing(parsec) time is high, we can look into switching to https://github.com/AndrasKovacs/flatparse

steve-chavez:
Though the implementation might be simple with [timeit](https://hackage.haskell.org/package/timeit-1.0.0.0/docs/System-TimeIt.html), doing this in the current codebase would be repetitive:

```hs
     (ActionRead headersOnly, TargetIdent identifier) -> do 
       wrPlan <- liftEither $ Plan.wrappedReadPlan identifier conf sCache apiReq 
       -- timeit
       resultSet <- runQuery roleIsoLvl (Plan.wrTxMode wrPlan) $ Query.readQuery wrPlan conf apiReq 
      -- timeit
[...]

Issue #2450, opened 2022-08-26
PostgREST takes a lot of time to make first embedding request with big schema


laurenceisla:
### Environment

* PostgreSQL version: v14
* PostgREST version: v10.0.0
* Operating system: EndeavourOS

### Description of issue

Using the [big schema](https://github.com/PostgREST/postgrest/blob/main/test/io/big_schema.sql) in this repo as an example, the first PostgREST start up takes a little while, after that, doing this query:

```bash
curl 'localhost:3000/projects?select=clients(*)'
```

takes a very long time. This may have to do with searching inside the schema cache. The subsequent requests are fast though.

I tested this for PosgREST v8, v9 and v10 with the same results.

wolfgangwalther:
I assume parsing the query result when loading the schema cache is actually done lazily - so only happens when the schema cache is used the first time.

steve-chavez:
I don't think there's something to be done here. If it only takes a while the first time it should be fine.

steve-chavez:
> for instance if you read in a file, do some processing, and then print out some of the results, it's likely that any processing not needed by the output won't be evaluated. However, the entire file will be read, even parts you never use.

From https://stackoverflow.com/a/2777842/4692662

The above is what happening on our case. Just replace "reading from file" with "querying the db".

I'm working on a logging improvement that clears the difference between "querying the schema cache" and "parsing the schema cache".

```
PGRST_DB_SCHEMAS="apflora" PGRST_DB_PLAN_ENABLED=1 PGRST_SERVER_TIMING_ENABLED=1  PGRST_ADMIN_SERVER_PORT=3001 PGRST_SERVER_HOST='*6' postgrest-with-post
gresql-15 -f test/io/big_schema.sql  postgrest-run
postgrest-with-postgresql-15: You can connect with: psql 'postgres:///postgres?host=/run/user/1000/postgrest/postgrest-with-postgresql-15-Acv/socket' -U postgres
postgrest-with-postgresql-15: You can tail the logs with: tail -f /run/user/1000/postgrest/postgrest-with-postgresql-15-Acv/db.log
06/Feb/2024:23:16:31 -0500: Starting PostgREST 11.2.0 (eaa1d81)...
06/Feb/2024:23:16:31 -0500: Attempting to connect to the database...
06/Feb/2024:23:16:31 -0500: Connection successful
06/Feb/2024:23:16:31 -0500: Admin server listening on port 3001
[...]

PR #2937, opened 2023-09-11
feat: add Server-Timing header with JWT duration

taimoorzaeem:
Part of #2771.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -114,6 +114,7 @@ library
                     , swagger2                  >= 2.4 && < 2.9
                     , text                      >= 1.2.2 && < 1.3
                     , time                      >= 1.6 && < 1.12
+                    , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
                     , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
@@ -203,13 +204,14 @@ test-suite spec
                       Feature.Query.DeleteSpec
                       Feature.Query.EmbedDisambiguationSpec
                       Feature.Query.EmbedInnerJoinSpec
-                      Feature.Query.PlanSpec
+                      Feature.Query.ErrorSpec
                       Feature.Query.HtmlRawOutputSpec
                       Feature.Query.InsertSpec
                       Feature.Query.JsonOperatorSpec
                       Feature.Query.MultipleSchemaSpec
-                      Feature.Query.ErrorSpec
+                      Feature.Query.NullsStrip
                       Feature.Query.PgSafeUpdateSpec
+                      Feature.Query.PlanSpec
                       Feature.Query.PostGISSpec
                       Feature.Query.QueryLimitedSpec
                       Feature.Query.QuerySpec
@@ -217,8 +219,8 @@ test-suite spec
                       Feature.Query.RawOutputTypesSpec
                       Feature.Query.RelatedQueriesSpec
                       Feature.Query.RpcSpec
+                      Feature.Query.ServerTimingSpec
                       Feature.Query.SingularSpec
-                      Feature.Query.NullsStrip
                       Feature.Query.SpreadQueriesSpec
                       Feature.Query.UnicodeSpec
                       Feature.Query.UpdateSpec
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -209,7 +209,7 @@ test-suite spec
                       Feature.Query.InsertSpec
                       Feature.Query.JsonOperatorSpec
                       Feature.Query.MultipleSchemaSpec
-                      Feature.Query.NullsStrip
+                      Feature.Query.NullsStripSpec
                       Feature.Query.PgSafeUpdateSpec
                       Feature.Query.PlanSpec
                       Feature.Query.PostGISSpec
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -14,6 +14,7 @@ very simple authentication system inside the PostgreSQL database.
 module PostgREST.Auth
   ( AuthResult (..)
   , getResult
+  , getJwtDur
   , getRole
   , middleware
   ) where
@@ -37,6 +38,7 @@ import Data.Either.Combinators (mapLeft)
 import Data.List               (lookup)
 import Data.Time.Clock         (UTCTime)
 import System.IO.Unsafe        (unsafePerformIO)
+import System.TimeIt           (timeItT)
 
 import PostgREST.AppState (AppState, getConfig, getTime)
 import PostgREST.Config   (AppConfig (..), JSPath, JSPathExp (..))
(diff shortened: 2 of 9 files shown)


steve-chavez:
@taimoorzaeem Interesting feature! So I've been playing a bit with it and I've noticed that:

```bash
$ PGRST_JWT_SECRET="reallyreallyreallyreallyverysafe" PGRST_DB_PLAN_ENABLED="true" postgrest-with-postgresql-15 postgrest-run
```

```bash
export JWT="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjk5OTk5OTk5OTksInJvbGUiOiJwb3N0Z3Jlc3RfdGVzdF9hdXRob3IiLCJpZCI6Impkb2UiLCJhdWQiOiJ5b3VyYXVkaW
VuY2UifQ.fJ4tLKSmolWGWehWN20qiU9dMO-WY0RI2VvacL7-ZGo"

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=335.1

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=147.7

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=145.9

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
Server-Timing: jwt;dur=146.2

# restarting the postgrest server

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=327.1

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=141.3

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=143.9

$ curl 'localhost:3000/authors_only' -H "Authorization: Bearer $JWT" -D -
HTTP/1.1 200 OK
Server-Timing: jwt;dur=142.5
```

The first jwt decode always takes longer than the subsequent ones. Is this because of laziness? Related: https://github.com/PostgREST/postgrest/issues/2450

steve-chavez:
@taimoorzaeem Also please rebase. Sorry, I wanted to do the refactoring for completing all the other durations (https://github.com/PostgREST/postgrest/issues/2771#issuecomment-1696590401) but I got carried away and did other changes.

test/spec/Feature/Query/ServerTiming.hs
```diff
@@ -0,0 +1,80 @@
+module Feature.Query.ServerTiming where
```
steve-chavez, line 1: It'd be better named as `ServerTimingSpec`, to maintain consistency with the other tests plus it's easier to find that way.

steve-chavez:
Strange that the idempotence test fails. Maybe some mutations on the new tests are not rolled back (which is the default for all tests `configDbTxRollbackAll  = True`).

src/PostgREST/Auth.hs
```diff
@@ -102,12 +104,12 @@ middleware appState app req respond = do
...
 
-  let token = fromMaybe "" $ Wai.extractBearerAuth =<< lookup HTTP.hAuthorization (Wai.requestHeaders req)
-  authResult <- runExceptT $
-    parseToken conf (LBS.fromStrict token) time >>=
-    parseClaims conf
+  let token  = fromMaybe "" $ Wai.extractBearerAuth =<< lookup HTTP.hAuthorization (Wai.requestHeaders req)
+      parseJwt = runExceptT $ parseToken conf (LBS.fromStrict token) time >>= parseClaims conf
 
-  let req' = req { Wai.vault = Wai.vault req & Vault.insert authResultKey authResult }
+  (dur,authResult) <- timeItT parseJwt
```
steve-chavez, line 110: This always runs right? Regardless of `configDbPlanEnabled`. I think it should be conditional to that config.

taimoorzaeem:
> The first jwt decode always takes longer than the subsequent ones. Is this because of laziness? Related: #2450

Hmm, I'll get into it. Would that be okay to test this using strict evaluation using `seq`? I am not sure because AFAIK `GHC` optimizes code much better with laziness than with strict evaluation.

steve-chavez:
> Would that be okay to test this using strict evaluation using seq? I am not sure because AFAIK GHC optimizes code much better with laziness than with strict evaluation. So maybe getting rid of laziness might not help us.

Yeah, not necessary to be strict. I mostly wanted to check if it's really because of laziness and why would it take more than double the time the first time. On https://github.com/PostgREST/postgrest/issues/2450, it's clear it takes a while because the schema cache loading is an expensive process. On this case I don't know what initialization takes place.

taimoorzaeem:
> Strange that the idempotence test fails. Maybe some mutations on the new tests are not rolled back (which is the default for all tests `configDbTxRollbackAll = True`).

I think this is more than just mutations not being rollbacked. For instance:
```
test/spec/Feature/Query/QuerySpec.hs:64:34: 
  1) Feature.Query.QuerySpec, Filtering response, matches nulls using not operator
       body mismatch:
         expected: "[{\"a\":\"1\",\"b\":\"0\"},{\"a\":\"2\",\"b\":\"0\"}]"
         but got:  "[{\"a\":\"1\",\"b\":\"1\"}, \n {\"a\":\"2\",\"b\":\"1\"}]"
```

Where is the `\n` coming from? So yeah, strange indeed.

test/spec/Feature/Query/ServerTimingSpec.hs
```diff
@@ -0,0 +1,80 @@
...
+          `shouldRespondWith`
+          [json|[{"id":7,"name":"John","referee":null,"auditor":null,"manager_id":6}]|]
+          { matchStatus  = 201
+          , matchHeaders = [ matchContentTypeJson
+                           , matchHeaderPresent "Server-Timing"]
+          }
+
+      it "works with patch request" $
+        request methodPatch "/no_pk?b=eq.0"
+            [("Prefer", "tx=commit")]
```
steve-chavez, line 42: I think this `tx=commit` is what's causing the error, remove?
```suggestion
      it "works with patch request" $
        request methodPatch "/no_pk?b=eq.0" mempty
```

postgrest.cabal
```diff
@@ -203,22 +204,23 @@ test-suite spec
...
                       Feature.Query.EmbedDisambiguationSpec
                       Feature.Query.EmbedInnerJoinSpec
-                      Feature.Query.PlanSpec
+                      Feature.Query.ErrorSpec
                       Feature.Query.HtmlRawOutputSpec
                       Feature.Query.InsertSpec
                       Feature.Query.JsonOperatorSpec
                       Feature.Query.MultipleSchemaSpec
-                      Feature.Query.ErrorSpec
+                      Feature.Query.NullsStrip
```
steve-chavez, line 212: If `NullsStrip` mutates the db (not sure), then this change of order might also affect the idempotence test.

taimoorzaeem: I don't think the order matters here since, `postgrest.cabal` simply lists the build and test dependencies.

test/spec/Feature/Query/ServerTimingSpec.hs
taimoorzaeem: Thanks @steve-chavez. This fixed the idempotence tests fails.

postgrest.cabal
steve-chavez: Ah my bad, I thought this list was somewhere in the SpecHelper.

steve-chavez (approved):

steve-chavez:
@taimoorzaeem Btw, could we make the JWT evaluation stricter with bang patterns?

- https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/strict.html
- https://wiki.haskell.org/Performance/Strictness

Pros of doing this:

- Will not confuse users when reading `Server-Timing` durations.
- Make JWT caching easier to test.

taimoorzaeem:
> @taimoorzaeem Btw, could we make the JWT evaluation stricter with bang patterns?

Hmm, I tried `BangPatterns` but it made no difference to the duration.

PR #2940, opened 2023-09-12
fix: inconsistent Preference-Applied

### LINKED ISSUE #2939: Inconsistencies with the `Preference-Applied` header
- `Prefer: tx=commit` gets inside `Preference-Applied` despite an uncommitted transaction:

```bash
$ PGRST_DB_TX_END="rollback-allow-override" postgrest-with-postgresql-15 postgrest-run
```

```bash
$ curl -D - -H "Content-Type: application/json" \-D - -H "Content-Type: application/json" \
    -H "Prefer: tx=commit" -H "Accept: application/vnd.pgrst.object"\
    'localhost:3000/rpc/setprojects' \
    -d @- <<JSON
{"id_l": 1, "id_h": 2, "name": "changed"}
JSON

HTTP/1.1 406 Not Acceptable
Transfer-Encoding: chunked
Date: Tue, 12 Sep 2023 04:24:23 GMT
Server: postgrest/11.2.0 (c5a9ed1)
Preference-Applied: tx=commit
Content-Type: application/vnd.pgrst.object+json; charset=utf-8

{"code":"PGRST116","details":"The result contains 2 rows","hint":null,"message":"JSON object requested, multiple (or no) rows returned"}
```

- `Prefer: count=exact` never gets into `Preference-Applied`:

```bash
$ curl -D - -H "Content-Type: application/json" \
  -H "Prefer: count=exact, return=representation" \
 "localhost:3000/projects" -d @- <<JSON
{"id":180, "name": "new"}
JSON

HTTP/1.1 201 Created
Transfer-Encoding: chunked
Date: Tue, 12 Sep 2023 04:21:52 GMT
Server: postgrest/11.2.0 (c5a9ed1)
Content-Range: */1
Preference-Applied: return=representation

[{"id":180,"name":"new","client_id":null}]
```

steve-chavez:
Closes #2939.

* Don't apply `tx=commit` if the transaction doesn't commit
* Apply `count=exact`
* Also simplifies  the `Preference-Applied` logic, removing the need for some functions.


test/spec/Feature/Query/SingularSpec.hs
```diff
@@ -72,8 +72,7 @@ spec =
           `shouldRespondWith`
             [json|{"details":"The result contains 4 rows","message":"JSON object requested, multiple (or no) rows returned","code":"PGRST116","hint":null}|]
             { matchStatus  = 406
-            , matchHeaders = [ matchContentTypeSingular
-                             , "Preference-Applied" <:> "tx=commit" ]
+            , matchHeaders = [ matchContentTypeSingular ]
```
steve-chavez, line 75: These are the tests for part 1 of the fix, `tx=commit` no longer added.

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -38,7 +38,7 @@ spec =
           `shouldRespondWith` [json|[{"id":2}]|]
           { matchStatus  = 200
           , matchHeaders = ["Content-Range" <:> "*/1"
-                           , "Preference-Applied" <:> "return=representation"]
+                           , "Preference-Applied" <:> "return=representation, count=exact"]
```
steve-chavez, line 41: For part 2, `count=exact` is now present on `Preference-Applied`.

src/PostgREST/Response.hs
```diff
@@ -98,6 +96,9 @@ createResponse QualifiedIdentifier{..} MutateReadPlan{mrMutatePlan} ctxApiReques
     let
       pkCols = case mrMutatePlan of { Insert{insPkCols} -> insPkCols; _ -> mempty;}
       response = gucResponse rsGucStatus rsGucHeaders
+      prefHeader = prefAppliedHeader $
+        Preferences (if null pkCols && isNothing (qsOnConflict iQueryParams) then Nothing else preferResolution)
+        preferRepresentation Nothing preferCount preferTransaction preferMissing
```
steve-chavez, line 101: This logic would be better centralized on the `Plan` module, mostly ensuring that prefers make sense for the operation (e.g., `return=representation` doesn't make sense on GET). For now I'm leaving it here as it's already a good change.

test/spec/Feature/Query/SingularSpec.hs
steve-chavez: > Also this would be one very specific case of not sending that header... but would it be the same for all kinds of errors that are thrown by pg itself during the transaction?

Currently only the tx=commit would be included in pref-applied in case of an error. None of the others preferences get inside. To fix this we need to handle it in a different way than the current `optionalRollback`.

 > The preference applied header should be considered more of a "did the server understand the requested preference / value and would apply it if relevant"?
> When I set tx=commit and get a response without that preference applied... I would assume I am talking to either a PostgREST version that does not support this preference, yet (or anymore), o

That use case would be much better handled by `handling=strict` https://www.rfc-editor.org/rfc/rfc7240.html#section-4.4

Because currently pref-applied is unreliable, just on this PR it can be seen that `count=exact` was never applied. So I don't think any user (even us maintainers) should rely on just pref-applied to guarantee a PostgREST version understands certain preference. But `handling-strict` can be used to guarantee that.

Can work on `handling=strict` after this PR. I think the Plan module approach for `Prefer` I mentioned below would make the implementation simpler, we could build the Prefer and fail at the Plan level.

@wolfgangwalther How does that sound?

---

Merging this seems like a big win because of how much code is deleted (the prefer stuff was getting complicated).

PR #2946, opened 2023-09-13
ci: loadtest commits to main against latest release

### LINKED ISSUE #2945: The `loadtest` CI job in `main` branch should run against latest release instead of itself
This issue is a follow-up to findings in https://github.com/PostgREST/postgrest/issues/2842#issuecomment-1717783040 

The description says it

develop7:
Introduces loadtest job for main branch that runs loadtest against latest release instead of itself

fixes #2945

### DIFF
--- a/.github/workflows/loadtest.yaml
+++ b/.github/workflows/loadtest.yaml
@@ -11,8 +11,9 @@ on:
       - main
 
 jobs:
-  Loadtest-Nix:
-    name: Loadtest (Nix)
+  Loadtest-PR-Nix:
+    name: Loadtest PR (Nix)
+    if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


steve-chavez:
This is definitely an improvement. The `loadtest` on main was not useful really.

I was thinking we should run it against `HEAD^1` instead but maybe that wouldn't be helpful in case we merge multiple commits? @wolfgangwalther WDYT? `loadtest` against latest tag LGTM.

steve-chavez:
Let's merge as it is. No harm done and it is an improvement.

steve-chavez:
@develop7 CI is always skipping this job on `main`: https://github.com/PostgREST/postgrest/actions/runs/6193207166/job/16814376470

Why would that be?

develop7:
The condition in L39 seems to evaluate to false
Try dropping the whole line with the `if:` and see what happens (should run always)

PR #2951, opened 2023-09-15
fix: Bug when Null Filtering on embedded resources

### LINKED ISSUE #2800: Null filtering on Embedded Resources bug
### Environment

* PostgreSQL version: 14
* PostgREST version: [11.0.1](https://github.com/PostgREST/postgrest/releases/tag/v11.0.1)
* Operating system: Windows

### Description of issue

Trying to use new feature : [Null filtering on Embedded Resources](https://postgrest.org/en/stable/references/api/resource_embedding.html#null-filtering-on-embedded-resources)

Table & Data : 

Table1 : 
| row | FieldA  | Table2(FieldB) -> foreign key to Table2 |
|-|--------|--------|
|1| valueA1 | valueB1 |
|2| valueA2 | valueB2| 

Table2
| row| FieldB | FieldC | FieldD|
|-|--------|--------|--------|
|1| valueB1 | valueC1 | valueD1|
|2| valueB2 | ValueC2 | NULL| 

doing : 
`GET /table1?select=*,table2(*)&table2=not.is.null`

Result : 

```json
[
    {
        "FieldA": "valueA1",
        "Table2": {
            "FieldB": "valueB1",
            "FieldC": "valueC1",
            "FieldD": "valueD1"
        }
    }
]
```

expected : 

```json
[
    {
        "FieldA": "valueA1",
        "Table2": {
            "FieldB": "valueB1",
            "FieldC": "valueC1",
            "FieldD": "valueD1"
        }
    },
    {
        "FieldA": "valueA2",
        "Table2": {
            "FieldB": "valueB2",
            "FieldC": "valueC2",
            "FieldD": NULL
        }
    }
]
```

I'm not getting the second row because FieldD is null 

it seems to be a problem from the postgresql's request,
 here the one i got through the log : (simplified)

```sql
select * table1
left join lateral (
	select * from table2
	where id = table1.table2_id ) as "table1_table2_id" on true
where "table1_table2_id" is not null

```

Any idea how i could get a workaround ? 

any idea how i could bypass this problem ?

laurenceisla:
Closes #2800

### DIFF
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -336,7 +336,7 @@ pgFmtArrayLiteralForField values _ = unknownLiteral (pgBuildArrayLiteral values)
 
 
 pgFmtFilter :: QualifiedIdentifier -> CoercibleFilter -> SQL.Snippet
-pgFmtFilter _ (CoercibleFilterNullEmbed hasNot fld) = pgFmtIdent fld <> " IS " <> (if hasNot then "NOT" else mempty) <> " NULL"
+pgFmtFilter _ (CoercibleFilterNullEmbed hasNot fld) = pgFmtIdent fld <> " IS " <> (if not hasNot then "NOT " else mempty) <> "DISTINCT FROM NULL"
 pgFmtFilter _ (CoercibleFilter _ (NoOpExpr _)) = mempty -- TODO unreachable because NoOpExpr is filtered on QueryParams
 pgFmtFilter table (CoercibleFilter fld (OpExpr hasNot oper)) = notOp <> " " <> pgFmtField table fld <> case oper of
    Op op val  -> " " <> simpleOperator op <> " " <> pgFmtUnknownLiteralForField (unknownLiteral val) fld
--- a/test/spec/Feature/Query/RelatedQueriesSpec.hs
+++ b/test/spec/Feature/Query/RelatedQueriesSpec.hs
@@ -257,13 +257,20 @@ spec = describe "related queries" $ do
         , matchHeaders = [matchContentTypeJson]
         }
 
-    -- "?table=not.is.null" does a "NOT table IS NULL" instead of a "table IS NOT NULL"
-    -- https://github.com/PostgREST/postgrest/issues/2800#issuecomment-1562620508
-    it "embeds even if the target relationship has a row with a NULL value in any of its columns" $
+    -- "?table=not.is.null" does a "table IS DISTINCT FROM NULL" instead of a "table IS NOT NULL"
+    -- https://github.com/PostgREST/postgrest/issues/2800#issuecomment-1720315818
+    it "embeds verifying that the entire target table row is not null" $ do
       get "/table_b?select=name,table_a(name)&table_a=not.is.null" `shouldRespondWith`
         [json|[
-          {"name":"Test 1","table_a":{"name":"Not null"}},
+          {"name":"Test 1","table_a":{"name":"Not null 1"}},
           {"name":"Test 2","table_a":{"name":null}}
+        ]|]
+        { matchStatus  = 200
+        , matchHeaders = [matchContentTypeJson]
+        }
+      get "/table_b?select=name,table_a()&table_a=is.null" `shouldRespondWith`
+        [json|[
+          {"name":"Test 3"}
         ]|]
         { matchStatus  = 200
         , matchHeaders = [matchContentTypeJson]
--- a/test/spec/Feature/Query/RelatedQueriesSpec.hs
+++ b/test/spec/Feature/Query/RelatedQueriesSpec.hs
@@ -256,3 +256,15 @@ spec = describe "related queries" $ do
         { matchStatus  = 200
         , matchHeaders = [matchContentTypeJson]
         }
+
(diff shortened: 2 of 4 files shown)


steve-chavez (approved):

steve-chavez:
> Or even better... maybe just translate table=not.is.null to an INNER JOIN entirely?

That might be difficult as joins much change the shape of the query.

steve-chavez:
> The correct check would be to translate table=not.is.null to something like table.join_column IS NOT NULL.

What would be the advantage of that one? `DISTINCT FROM` null is more similar to `table=not.is.null`.

laurenceisla:
> Or even better... maybe just translate `table=not.is.null` to an `INNER JOIN` entirely?

It could be possible for anti joins but the OR queries wouldn't work this way. Although I get that it's not 1-1 with PostgreSQL (`table=isdistinct.null` might be more precise). With that in mind, using `is not null` gives the errors from the issue: it checks if there's a null value for every column in the row instead of the existence of the row.

laurenceisla:
I did some benchmarks using `pgbench` for each of the approaches to the resulting query of:

```bash
curl "http://localhost:3000/projects?select=name,clients(name)&clients=not.is.null"
```

<details>
<summary>Queries:</summary>

```SQL


-- IS DISTINCT FROM NULL

WITH pgrst_source AS
  (SELECT "test"."projects"."name",
          row_to_json("projects_clients_1".*) AS "clients"
   FROM "test"."projects"
   LEFT JOIN LATERAL
     (SELECT "clients_1"."name"
      FROM "test"."clients" AS "clients_1"
      WHERE "clients_1"."id" = "test"."projects"."client_id") AS "projects_clients_1" ON TRUE
   WHERE "projects_clients_1" IS NOT NULL)
       nullif(current_setting('response.status', TRUE), '') AS response_status
FROM
  (SELECT *
   FROM pgrst_source) _postgrest_t;


-- IS DISTINCT FROM

WITH pgrst_source AS
  (SELECT "test"."projects"."name",
          row_to_json("projects_clients_1".*) AS "clients"
   FROM "test"."projects"
   LEFT JOIN LATERAL
     (SELECT "clients_1"."name"
      FROM "test"."clients" AS "clients_1"
      WHERE "clients_1"."id" = "test"."projects"."client_id") AS "projects_clients_1" ON TRUE
   WHERE "projects_clients_1" IS DISTINCT FROM NULL)
       nullif(current_setting('response.status', TRUE), '') AS response_status
FROM
  (SELECT *
   FROM pgrst_source) _postgrest_t;


-- INNER JOIN

WITH pgrst_source AS
  (SELECT "test"."projects"."name",
          row_to_json("projects_clients_1".*) AS "clients"
   FROM "test"."projects"
   INNER JOIN LATERAL
     (SELECT "clients_1"."name"
      FROM "test"."clients" AS "clients_1"
      WHERE "clients_1"."id" = "test"."projects"."client_id") AS "projects_clients_1" ON TRUE)
       nullif(current_setting('response.status', TRUE), '') AS response_status
FROM
  (SELECT *
   FROM pgrst_source) _postgrest_t;
```

</details>

<details>
<summary>The `pgbench` for each of the queries:</summary>

```log
-- IS NOT NULL (Before the change)

duration: 25 s
number of transactions actually processed: 152304
latency average = 0.164 ms
initial connection time = 1.737 ms
tps = 6092.564790 (without initial connection time)


 -- IS DISTINCT FROM NULL

duration: 25 s
number of transactions actually processed: 158569
latency average = 0.158 ms
initial connection time = 2.025 ms
tps = 6343.240056 (without initial connection time)


-- INNER JOIN

duration: 25 s
number of transactions actually processed: 169379
latency average = 0.148 ms
initial connection time = 1.733 ms
tps = 6775.588488 (without initial connection time)
```

</details>

> Performance. I don't think the distinct from approach will be optimized very well, while comparing only the join columns is very likely optimized nicely.

The `INNER JOIN` is indeed the most performant, but I don't see how it can be rewritten to allow the `OR` between `?or=(table1.not.is.null,table2.not.is.null)`, for instance.

> The correct check would be to translate table=not.is.null to something like table.join_column IS NOT NULL.

I didn't know how to implement this for the test (and allow the OR across joins), just used the `IS NOT NULL` before the change in this PR.

steve-chavez:
> -- IS NOT NULL (Before the change)
> tps = 6092.564790 (without initial connection time)

> -- IS DISTINCT FROM NULL
> tps = 6343.240056 (without initial connection time)

So we getter perf now! :tada: 

> The INNER JOIN is indeed the most performant, but I don't see how it can be easily rewritten to allow the OR 

Yeah, INNER JOIN cannot be done and this goes back to the introduction of top-level filtering: https://github.com/PostgREST/postgrest/pull/1949#issuecomment-924191966. Iced-Sun's query helped us there.

laurenceisla:
> But in my latest comment I didn't say "write as JOIN", but "check only join columns".

Yes, at the end of the tests I added that I couldn't find a way on how to do this, that's why it's not included there.

> If you test the join columns, PG might be able to rewrite this to an inner join automatically.

Cool, then the problem right now is to rewrite the queries to allow filtering on the JOIN columns. It does not seem to be trivial: as it is now, we select from the LEFT JOIN LATERAL subquery, if the join column is not selected by the user `?select=joincol`, then it's not included in that subquery, hence the filter cannot be done. We need to find a way around that, if anyone have some ideas I could give them a try.

PR #2960, opened 2023-09-20
ci: loadtest PRs against latest releases

develop7:
Lets `postgrest-loadtest-against` accept multiple refs performing loadtests against all of them; and 
introduces loadtesting PRS against latest release as well.

The downside is we're now unable to pass addtional parameters to vegeta (due to limitation of argbash's support of infinite parameters), please let me know if this is a problem (/cc um… @wolfgangwalther?)

### DIFF
--- a/.github/workflows/loadtest.yaml
+++ b/.github/workflows/loadtest.yaml
@@ -23,9 +23,13 @@ jobs:
         uses: ./.github/actions/setup-nix
         with:
           tools: loadtest
+      - uses: actions-ecosystem/action-get-latest-tag@v1
+        id: get-latest-tag
+        with:
+          semver_only: true          
       - name: Run loadtest
         run: |
-          postgrest-loadtest-against main
+          postgrest-loadtest-against main ${{ steps.get-latest-tag.outputs.tag }}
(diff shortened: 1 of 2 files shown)


steve-chavez:
Interesting! The load test result: https://github.com/PostgREST/postgrest/pull/2960/checks?check_run_id=16970655907

Another data point will also help clear some random noise, as it was the case on https://github.com/PostgREST/postgrest/pull/2934#issuecomment-1713773902

develop7:
@wolfgangwalther from what I've gathered during the load testing in CI is the impression that the resources budget allocated to the VM the CI job runs in is probably consistent across VM's lifetime. OTOH, I agree that multiple runs would let us know how consistently is postgrest performing under the load. Shouldn't be hard to implement, will do with copy&paste for clarity's sake.

And since we're having multiple runs, I'd suggest adding some extra stats, i.e. mean & sigma and/or min & max. Or, even better, since there's not very much runs against the same revision, add delta against mean value for each run. Of course this one seems to deserve for separate issue at least.

steve-chavez (approved):

Issue #2816, opened 2023-06-09
Implement "Prepared Requests"


steve-chavez:
## Problem

There are several steps we do for each request to become a generated query. e.g. for a GET:

https://github.com/PostgREST/postgrest/blob/1f13e43abec9fb7acda3967d3645d28e6a1f112a/src/PostgREST/App.hs#L167-L173

Before that:

https://github.com/PostgREST/postgrest/blob/1f13e43abec9fb7acda3967d3645d28e6a1f112a/src/PostgREST/App.hs#L148-L153

## Solution

Akin to "prepared statements", implement "prepared requests". We can make a hash of our input parameters and store the generated query. Then just execute the query using the arguments. 

## Notes

- Avoiding the ApiRequest step to parse the query parameters(`?id=eq.3&col=lt.4`) is not possible. If we switched to a grammar like `?id.eq=3&col.lt=4`(https://github.com/PostgREST/postgrest/issues/2066`) it might be possible to avoid our query grammar parsing and only do some lighter parsing.
- If we use the SEARCH method(https://github.com/PostgREST/postgrest/issues/2125#issuecomment-1376533758) and only have placeholder parameters on the query params then it should be possible to completely avoid parsing.   
- use a library like https://github.com/cocreature/lrucaching/
- check how postgresql does prepared statements to see if we can use similar modules
- similar idea to https://github.com/PostgREST/postgrest/issues/2698

steve-chavez:
> hash only the "structure", so kind of like a "parametrized request", similar to the prepared statement?
> If not... then you'd need to parse way too much of the actual structure of the request before you could actually calculate the hash - you'd almost be there in that case.

Yes, only the structure. But as you mention that's not worth it. SEARCH looks good. 

My motivation is  that with https://github.com/PostgREST/postgrest/pull/2523 and https://github.com/PostgREST/postgrest/pull/1582 we're going to be putting more work at the Plan level. That will certainly not be zero cost in terms of perf.

---

https://github.com/PostgREST/postgrest/issues/2771 will prove how much time we take for parsing/planning.

taimoorzaeem:
A cache for `QueryParams` sounds good. Maybe something like `LRU ByteString QueryParams` using [lrucache](https://hackage.haskell.org/package/lrucache-1.2.0.1/docs/Data-Cache-LRU.html)? I believe this would reduce parse time significantly for the same query.

PR #2964, opened 2023-09-25
Introduce ServerTiming for queries

develop7:
Roadmap:

- [ ] `db` — database response time
- [ ] `render` — resultset render time
- [ ] `plan` — :question: :question: :question: do we need it? is it worth it? does it get stuck?
- [x] `jwt` — authentication time

quoting https://github.com/PostgREST/postgrest/issues/2771#issuecomment-1696590401:
> For now I'm thinking of the following steps, the main idea is to make the Response and Query modules pure; this will also enable those to be "doctested":
> 
* [x]  Make the Response module pure (wip on [steve-chavez@edbc4f9](https://github.com/steve-chavez/postgrest/commit/edbc4f92ad51b7c672251e6d9d1269caaad61bd5))
* [ ]  Make the Query module pure
	* This will also help implement [Feature request: show SQL query in terminal(log-level=debug) #1578 (comment)](https://github.com/PostgREST/postgrest/issues/1578#issuecomment-1705737599)
	* The Query module has functions, like `failNotSingular`, that are impure in nature (they need to rollback), these can be separated into another `AbortQuery` module.
* [ ]  Then have a single function entrypoint for each module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2939, Fix `count=exact` not being included in `Preference-Applied` - @steve-chavez
  - #2800, Fix not including to-one embed resources that had a `NULL` value in any of the selected fields when doing null filtering on them - @laurenceisla
  - #2846, Fix error when requesting `Prefer: count=<type>` and doing null filtering on embedded resources - @laurenceisla
+ - #2846, Fix setting `default_transaction_isolation` unnecessarily - @steve-chavez
 
 ## [11.2.0] - 2023-08-10
 
--- a/src/PostgREST/Config/Database.hs
+++ b/src/PostgREST/Config/Database.hs
@@ -154,7 +154,7 @@ queryRoleSettings prepared =
       select
         kv.rolname,
         i.value as iso_lvl,
-        array_agg(row(kv.key, kv.value)) filter (where key <> 'default_transation_isolation') as role_settings
+        coalesce(array_agg(row(kv.key, kv.value)) filter (where key <> 'default_transaction_isolation'), '{}') as role_settings
       from kv_settings kv
       join pg_settings ps on ps.name = kv.key and ps.context = 'user'
       left join iso_setting i on i.rolname = kv.rolname


develop7:
hey @steve-chavez do we need to make `Response.info{Ident,Info,Proc}Response` pure as well? My guess is we do, but not right now

https://github.com/PostgREST/postgrest/blob/a6e3eda5b20817f134beb8d660029bd2875c280f/src/PostgREST/Response.hs#L194-L214

steve-chavez:
Hey @develop7, yeah so `infoIdentResponse` is the response for the OPTIONS method and this one is special because it doesn't have a Plan or Query stage - it doesn't execute any query.

Hm, maybe it's not that relevant for `Server-Timing` but I think it would be good to have it pure for completion and consistency.

steve-chavez:
>  plan — ❓ ❓ ❓ do we need it? is it worth it? does it get stuck?

Yeah, so during [resource embedding](https://postgrest.org/en/stable/references/api/resource_embedding.html) I believe the plan could get slow. If that's true then we might try to optimize the code or cache the results somehow ([idea here](https://github.com/PostgREST/postgrest/issues/2816)).

But first we would need to confirm if that is indeed the case.

steve-chavez:
Might make sense to combine both the `ApiRequest.userApiRequest` and the Plan module work into the same `Server-Timing` duration too.

develop7:
Gotchu, on it

steve-chavez:
@develop7 The current refactor seem good to merge. Maybe you'd like to open another PR for that and leave this one open as sort of an "epic"?

Otherwise you'll have to rebase each time the Response module is modified, that's currently happening on https://github.com/PostgREST/postgrest/pull/2969/

develop7:
> it doesn't execute any query

Hmm, that means we need to make the `db` timing optional at least here, which requires all/some metrics to be optional. 

I'm wondering if we could get away with something like 

```haskell
data ResponseTiming =  RMJwt | RMRender | RMDb

data ServerTimings = Map ResponseTiming Double
```

then pass the value of ServerTimings along, throwing metrics' values in as needed; then render the header somewhere near our newly introduced `pgrstResponse`, wrapper around WAI.Response. That should enable us to add extra metrics relatively easily.

What do you guys think?

develop7:
> Maybe you'd like to open another PR for that and leave this one open as sort of an "epic"?

Good point, let me shuffle commits a bit then.

steve-chavez:
> Hmm, that means we need to make the db timing optional at least here, which requires all/some metrics to be optional.

True.

> That should enable us to add extra metrics relatively easily.
> What do you guys think?

@develop7 Sounds good!

develop7:
> Good point, let me shuffle commits a bit then.

Shuffling done, the PR got accidentally closed though

PR #2971, opened 2023-09-27
refactor: make Response module pure

develop7:
Part of #2771

created per the suggestion at https://github.com/PostgREST/postgrest/pull/2964#issuecomment-1737441167


develop7:
@steve-chavez done

steve-chavez (approved):

develop7:
> I don't see an easy fix for those now so LGTM.

um, why, the rest of module doesn't depend on them, so they could reside somewhere in/under `App`

steve-chavez:
> um, why, the rest of module doesn't depend on them, so they could reside somewhere in/under App just as well

Ah right, simply moving them to another module would be fine.

PR #2972, opened 2023-09-27
fix: unnecessary set default_transaction_isolation

### LINKED ISSUE #2959: There's a typo in "default_transation_isolation" with a missing c
https://github.com/PostgREST/postgrest/blob/1b4dae5a7a48842880a7ed11b9a7b6487023e4ad/src/PostgREST/Config/Database.hs#L157

steve-chavez:
Closes #2959

### DIFF
--- a/src/PostgREST/Config/Database.hs
+++ b/src/PostgREST/Config/Database.hs
@@ -154,7 +154,7 @@ queryRoleSettings prepared =
       select
         kv.rolname,
         i.value as iso_lvl,
-        array_agg(row(kv.key, kv.value)) filter (where key <> 'default_transation_isolation') as role_settings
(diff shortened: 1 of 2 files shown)


src/PostgREST/Config/Database.hs
```diff
@@ -154,7 +154,7 @@ queryRoleSettings prepared =
       select
         kv.rolname,
         i.value as iso_lvl,
-        array_agg(row(kv.key, kv.value)) filter (where key <> 'default_transation_isolation') as role_settings
+        coalesce(array_agg(row(kv.key, kv.value)) filter (where key <> 'default_transaction_isolation'), '{}') as role_settings
```
steve-chavez, line 157: I tried putting the `'default_transaction_isolation'` into a variable and interpolate with `qc` - this fails to compile.

This has happened a number of times now, we need a new quasiquoting lib.

PR #2977, opened 2023-09-29
fix: incorrect parsing for fallback_application_name

### LINKED ISSUE #2970: Connection error - Database does not exists after patching to a newer version
### Environment

* PostgreSQL version: (if using docker, specify the image)
*  15.4(container)
* PostgREST version: (if using docker, specify the image)
* 11.2.0 and 11.1.0(container)
* Operating system:

### Description of issue

From PostgREST version 11.0.1 the api can connect to the database via the config:

PGRST_SERVER_PORT=2888
PGRST_DB_URI=postgres://forecasting:<PW_HERE>@<HOST_HERE>:5432/forecasting
PGRST_DB_SCHEMA=api
PGRST_DB_ANON_ROLE=forecasting

But after patching to the next version I see errors:

26/Sep/2023:13:14:51 +0200: {"code":"PGRST000","details":"connection to server at \"<HOST_HERE>\" (10.8.106.143), port 5432 failed: FATAL:  database \"forecasting fallback_application_name='PostgREST 11.2.0'\" does not exist\n","hint":null,"message":"Database connection error. Retrying the connection."}

I see the same errors in the logging in the postgresql database logging. The config has not changed between versions.

steve-chavez:
Fixes #2970

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -49,8 +49,7 @@ import Data.List.NonEmpty      (fromList, toList)
 import Data.Maybe              (fromJust)
 import Data.Scientific         (floatingOrInteger)
 import Network.URI             (escapeURIString,
-                                isUnescapedInURIComponent, parseURI,
-                                uriQuery)
+                                isUnescapedInURIComponent)
 import Numeric                 (readOct, showOct)
 import System.Environment      (getEnvironment)
 import System.Posix.Types      (FileMode)
@@ -472,6 +471,8 @@ readPGRSTEnvironment =
   M.map T.pack . M.fromList . filter (isPrefixOf "PGRST_" . fst) <$> getEnvironment
 
 -- | Adds a `fallback_application_name` value to the connection string. This allows querying the PostgREST version on pg_stat_activity.
+-- | This follows the same logic as libpq for parsing the URI https://github.com/postgres/postgres/blob/a829b704015104f49a11ea007957ef03a03dc0d4/src/interfaces/libpq/fe-connect.c#L5880-L5921
+-- | If there's a postgres URI scheme designator prefix ("postgres://") then the connstring is considered an URI, if not, it's in key/value format.
 --
 -- >>> let ver = "11.1.0 (5a04ec7)"::ByteString
 -- >>> let strangeVer = "11'1&0@#$%,.:\"[]{}?+^()=asdfqwer"::ByteString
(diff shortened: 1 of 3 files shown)


src/PostgREST/Config.hs
```diff
@@ -496,15 +497,25 @@ readPGRSTEnvironment =
...
+-- "postgres:///postgres:pass=()[]&#?host=server&port=5432&fallback_application_name=PostgREST%2011.1.0%20%285a04ec7%29"
 addFallbackAppName :: ByteString -> Text -> Text
 addFallbackAppName version dbUri = dbUri <>
-  case uriQuery <$> parseURI (toS dbUri) of
-    Nothing  -> " " <> keyValFmt -- Assume key/value connection string if the uri is not valid
-    Just ""  -> "?" <> uriFmt
-    Just "?" -> uriFmt
-    _        -> "&" <> uriFmt
+  if isPgURI dbUri
+    then case snd $ T.breakOn "?" dbUri of
```
laurenceisla, line 507: Does it work when the password has a `?` character inside? It works OK when it's present in `v11.0.1`.

steve-chavez: On main I get:

```
ghci> addFallbackAppName "11.0.0" "postgres:///postgres:pass=qwer?xyz"
"postgres:///postgres:pass=qwer?xyz&fallback_application_name=PostgREST%2011.0.0"
```

Which seems to be broken? Or does the `&` work without needing the `?` first?

---

On this branch it gives the same result:

```
ghci> addFallbackAppName "11.0.0" "postgres:///postgres:pass=qwer?xyz"
"postgres:///postgres:pass=qwer?xyz&fallback_application_name=PostgREST%2011.0.0"
```

steve-chavez: Yeah, just confirmed that works on v11.0.1 and also works for:

```
PGRST_DB_URI='postgres://id:mysecret?password@localhost:5432/postgres?application_name=postgrest' postgrest-run
```

And the `application_name` does get applied to the connection (confirmed by querying pg_Stat_activity)

---

The libpq parsing for the password follows the condition that it is contained between `:` and `@`:

https://github.com/postgres/postgres/blob/a829b704015104f49a11ea007957ef03a03dc0d4/src/interfaces/libpq/fe-connect.c#L6522-L6549

I think we can do something similar here.

steve-chavez: Or we might do the simplest thing: refuse to append fallback app name if the uri has more than one `?` character.

steve-chavez: We can document that as: "make sure you're URI is correctly url-encoded to ensure the fallback_application_name gets added".

steve-chavez: This should be good enough until we have a good connstring parser: https://github.com/jfischoff/postgres-options/issues/8

steve-chavez:
This approach is fraught. As @laurenceisla mentioned, we also need to consider that `?` can be inside the password.

A quick fix without a correct parser still has the potential to mess the uri.

Conclusion: only append the fallback_app_name on valid URIs, do not apply it for key/value formats. Document this as a limitation on the docs. This until we get a stable parsing library.

PR #2983, opened 2023-10-02
Add more data to `Server-Timing` header

### LINKED ISSUE #2771: `Server-Timing` response header

develop7:
Introduces the performance metrics' collection and rendering in `Server-Timing` header 

fixes #2771 

- [x] metrics
	* `query`
	* `jwt`
	* `plan`
	* `render`
- [x] tests 
- [x] CHANGELOG

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -45,23 +45,27 @@ import qualified PostgREST.Plan             as Plan
 import qualified PostgREST.Query            as Query
 import qualified PostgREST.Response         as Response
 
-import PostgREST.ApiRequest          (Action (..), ApiRequest (..),
-                                      Mutation (..), Target (..))
-import PostgREST.AppState            (AppState)
-import PostgREST.Auth                (AuthResult (..))
-import PostgREST.Config              (AppConfig (..))
-import PostgREST.Config.PgVersion    (PgVersion (..))
-import PostgREST.Error               (Error)
-import PostgREST.Query               (DbHandler)
-import PostgREST.Response            (ServerTimingParams (..))
-import PostgREST.SchemaCache         (SchemaCache (..))
-import PostgREST.SchemaCache.Routine (Routine (..))
-import PostgREST.Version             (docsVersion, prettyVersion)
+import PostgREST.ApiRequest           (Action (..), ApiRequest (..),
+                                       Mutation (..), Target (..))
+import PostgREST.AppState             (AppState)
+import PostgREST.Auth                 (AuthResult (..))
+import PostgREST.Config               (AppConfig (..))
+import PostgREST.Config.PgVersion     (PgVersion (..))
+import PostgREST.Error                (Error)
+import PostgREST.Query                (DbHandler)
+import PostgREST.Response.Performance (ServerMetric (..),
+                                       ServerTimingData,
+                                       renderServerTimingHeader)
+import PostgREST.SchemaCache          (SchemaCache (..))
+import PostgREST.SchemaCache.Routine  (Routine (..))
+import PostgREST.Version              (docsVersion, prettyVersion)
 
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.List             as L
+import qualified Data.Map              as Map (fromList)
 import qualified Network.HTTP.Types    as HTTP
 import           Protolude             hiding (Handler)
+import           System.TimeIt         (timeItT)
 
 type Handler = ExceptT Error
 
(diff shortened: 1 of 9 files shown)


develop7:
dead code error at https://github.com/PostgREST/postgrest/actions/runs/6418699702/job/17427023445?pr=2983#step:4:119, the matchHeaderPresent gets unused after my PR; I'd say it's a handy matcher, I'd leave it, but don't know how

steve-chavez:
@develop7 Interesting! So doing:

```bash
# PGRST_DB_PLAN_ENABLED="true" postgrest-with-postgresql-15 postgrest-run

$ curl 'localhost:3000/projects?select=*,clients(*)'  -I
Server-Timing: jwt;dur=16.0, render;dur=8.8, plan;dur=16956.1, query;dur=763.5

$ curl 'localhost:3000/projects?select=*,clients(*)'  -I
Server-Timing: jwt;dur=13.3, render;dur=7.5, plan;dur=118.7, query;dur=1156.1

$ curl 'localhost:3000/projects?select=*,clients(*)'  -I
Server-Timing: jwt;dur=34.3, render;dur=11.9, plan;dur=283.7, query;dur=1121.3

$ curl 'localhost:3000/projects?select=*,clients(*)'  -I
Server-Timing: jwt;dur=15.4, render;dur=5.0, plan;dur=119.0, query;dur=478.7

$ curl 'localhost:3000/projects?select=*,clients(*)'  -I
Server-Timing: jwt;dur=30.4, render;dur=9.8, plan;dur=274.1, query;dur=1165.8
```

Shows that:

- The plan takes a lot the first time, this must be because of https://github.com/PostgREST/postgrest/issues/2450
- The plan time varies and is not negligible.
- The bulk of the time goes to the query though.
- render time is low

> dead code error at https://github.com/PostgREST/postgrest/actions/runs/6418699702/job/17427023445?pr=2983#step:4:119, the matchHeaderPresent gets unused after my PR; I'd say it's a handy matcher, I'd leave it, but don't know how

Yeah, maybe just comment it for now?

src/PostgREST/App.hs
```diff
@@ -170,63 +174,73 @@ runDbHandler appState isoLvl mode authenticated prepared handler = do
...
+handleRequest AuthResult{..} conf appState authenticated prepared pgVer apiReq@ApiRequest{..} sCache jwtTime =
   case (iAction, iTarget) of
     (ActionRead headersOnly, TargetIdent identifier) -> do
-      wrPlan <- liftEither $ Plan.wrappedReadPlan identifier conf sCache apiReq
-      resultSet <- runQuery roleIsoLvl (Plan.wrTxMode wrPlan) $ Query.readQuery wrPlan conf apiReq
-      pgrst <- liftEither $ Response.readResponse wrPlan headersOnly identifier apiReq resultSet serverTimingParams
-      return $ pgrstResponse pgrst
+      (planTime, wrPlan) <- timeItT $ liftEither $ Plan.wrappedReadPlan identifier conf sCache apiReq
+      (rsTime, resultSet) <- timeItT $ runQuery roleIsoLvl (Plan.wrTxMode wrPlan) $ Query.readQuery wrPlan conf apiReq
+      (renderTime, pgrst) <- timeItT $ liftEither $ Response.readResponse wrPlan headersOnly identifier apiReq resultSet
```
steve-chavez, line 183: `timeItT` always gets done despite `configDbPlanEnabled=false` right? Ideally we don't time the action if it's not necessary.

develop7: That what I was going to do, but this way it seemed to be easier to implement at one hand, and not introduce significant overhead due to the measurement, on the other. Do you have any particular improvements in mind?

develop7:
> Please don't. The dead code checker is there for exactly this reason: To remove dead code. If you comment it out... the code is still dead. The point of removing dead code is to reduce maintenance effort. And commenting out does not help with it.

Good point, although the above matcher still can be handy later, can it? Ideally, it belongs if not `hspec-wai` itself, then `hspec-wai-extra` for certain. This `SpecHelper` module serves as a project-specific lightweight library, rather than stores business logic, so I'm not sure deleting helpers from here would decrease maintenance efforts significantly. On the other hand, the scenario when one day we delete it and next month we need it again, seems quite feasible to me.

develop7:
@wolfgangwalther agree. Deleted it.

src/PostgREST/App.hs
steve-chavez: > and not introduce significant overhead due to the measurement

I'm not sure if this is the case. A more realistic load testing would settle it. 

We have this repo now: https://github.com/PostgREST/postgrest-benchmark. Try running the [quickstart](https://github.com/PostgREST/postgrest-benchmark/#quickstart) which includes a load test. Then:

- [produce a static bin](https://github.com/PostgREST/postgrest/tree/main/nix#building-postgrest) of this branch.
- use the static bin on the [postgrest derivation](https://github.com/PostgREST/postgrest-benchmark/blob/master/postgrest/postgrest.nix).
- rerun the load test and compare the results to the previous.

If they don't vary much, then I guess we can keep it as is.

---

Might be easier to do the `timeIt` conditionally than running the above benchmark :smile:

develop7: Okay, made timing conditional, please review again

steve-chavez (approved):
@develop7 Great work! The new modules and test helpers are really clean too :+1:.

Don't forget to add an entry to the CHANGELOG!

develop7:
@steve-chavez changelog done, rebasing done. Hit it!

Issue #2776, opened 2023-05-02
Provide an accurate way to determine if the returned response is complete or partial


Pigeo:
### Environment

* PostgreSQL version: probably any
* PostgREST version: probably any
* Operating system: any

### Description of issue

The [db-max-rows](https://postgrest.org/en/stable/configuration.html#db-max-rows) is an important setting that limits the number of rows PostgREST will fetch from a view, table, or stored procedure (against accidental or malicious requests).

On my web application, I need to know wether the content returned by PostgREST is partial (`db-max-rows` has effectively limited the number of results) or complete (i.e. number of returned results < `db-max-rows`).

Since PostgREST is returning HTTP status code 206 (Partial Content) + the `Content-Range` (when passing HTTP header `Prefer: count=estimated`), I initially thought I could rely on one of these 2 hints to determine if the returned content is partial or complete, but that doesn't work, for one simple reason:

in case of [resource embedding](https://postgrest.org/en/stable/api.html#resource-embedding), the `db-max-rows` setting also applies to embedded tables (which I understand is a very good thing, for safety reasons), but the HTTP 206 doesn't apply to embedded resources (when only the embedded resource is effectively limited to `db-max-rows`, I always get an HTTP 200, even though I correctly passed the `Prefer: count=estimated` HTTP header), and the `Content-Range` is of no use here (because it doesn't give a clue about the embedded resource's rows)

### Expected behavior / suggestions of improvement

1. as long as the `db-max-rows` has effectively limited the response in *any* way (including on embedded resources), the HTTP 206 status code should always be returned (at least when `Prefer: count=estimated` HTTP header is present)
2. if you're reluctant to change the actual behaviour of the HTTP status code (for breaking-change reason), you could alternatively add another new HTTP header `Response-Is-Complete: false` that would pursue this purpose.
3. going further, it would also be nice to have size of results not only for the main table but also for each embedded resources => here I'm thinking about a small hierarchical json object, having the same hierarchical structure as the `select=` parameter of the request (with all its embedded resources), that would give us the retrieved rows / total numbers for each embedded resource (at each node of this hierarchical object).
4. another very quick and simple way to answer my initial need would simply consist of returning the `db-max-rows` value as an HTTP header of PostgREST's response, so that I can check for myself if any of the embedded ressource arrays has reached this size limit.

Pigeo:
[...]

PR #2984, opened 2023-10-02
feat: add config to disable Prefer: count=exact

### LINKED ISSUE #2777: Config flag for disabling `Prefer: count=exact`
We currently rely on this guide for protecting against full counts on big tables: https://postgrest.org/en/v10.2/admin.html#count-header-dos

Which is incomplete because AFAICT Nginx can't remove a header based on a value OOTB. It will always remove the whole header. We'd need OpenResty for this.

Now, this problem has already been solved with the [estimated](https://postgrest.org/en/v10.2/api.html#estimated-count) and [planned](https://postgrest.org/en/v10.2/api.html#planned-count) count.

We just need to disable the `Prefer: count=exact`. We can do this with a config option.

```
db-exact-count-enable = false
```

taimoorzaeem:
This should fix #2777.


steve-chavez:
@taimoorzaeem Oh, #2777  was labeled as `idea`. I've just added a description for it: `still in need of discussion, do not implement`. Sorry about that :sweat_smile:. Generally I just open them to see if someone has any feedback.

So exposing `count: exact` is still a problem, however I'm not sure if disabling it is the correct solution. 

- `count=planned` is immediate but inaccurate if the statistics are out of date.
- `count=estimated` depends on max-rows being set, which causes an issue https://github.com/PostgREST/postgrest/issues/2776
  + (another pending issue) I've detected that it's not working as intended when having high counts. The query is still slow.

---

The point of #2777 remains that it's still cumbersome to remove this value at the proxy level.

Since `Prefer: tx=rollback` can be disabled with a config maybe we can have a single config for all the preferences? I still need to think if this make sense but that would be better since we don't need to have one config for each preference.

Let's leave the PR open for now.

steve-chavez:
> Since Prefer: tx=rollback can be disabled with a config maybe we can have a single config for all the preferences? 

FYI, added this idea on https://github.com/PostgREST/postgrest/issues/2987

PR #2986, opened 2023-10-03
feat: add config to specify CORS origins

### LINKED ISSUE #2441: Is there a way to setup CORS origins? Nginx config overwritten by postgrest
I was trying to setup nginx CORS rules for my project in nginx ingress in kubernetes, until I realized a weird behaviour, and it is that postgrest is always adding the "access-control-allow-origin: *" header. Is there a way to completely disable this? or modify this header?  As nginx is returning this header when it shoudn't. It would be nice being able to disable this behaviour, or just being able to modify the origins returned by the access-control-allow-origin header.

taimoorzaeem:
This should fix #2441.


src/PostgREST/Response.hs
```diff
@@ -224,17 +225,18 @@ infoIdentResponse identifier sCache = do
...
+infoRootResponse corsAllowedOrigins = respondInfo corsAllowedOrigins "OPTIONS,GET,HEAD"
 
-respondInfo :: ByteString -> Either Error.Error PgrstResponse
-respondInfo allowHeader =
-  let allOrigins = ("Access-Control-Allow-Origin", "*") in
-  Right $ PgrstResponse HTTP.status200 [allOrigins, (HTTP.hAllow, allowHeader)] mempty
+respondInfo :: Text -> ByteString -> Either Error.Error PgrstResponse
+respondInfo corsAllowedOrigins allowHeader =
+  let allowedOrigins = ("Access-Control-Allow-Origin", T.encodeUtf8 corsAllowedOrigins) in
+  Right $ PgrstResponse HTTP.status200 [allowedOrigins, (HTTP.hAllow, allowHeader)] mempty
```
laurenceisla, line 239: To include the CORS validation for all the methods (`GET`, `POST`, etc.), the allowed origins should be handled here:

https://github.com/PostgREST/postgrest/blob/0703f27d2b0ffe60edfc6053ea5df10a3ed2c0dd/src/PostgREST/Cors.hs#L24

If you add the content of `cors-allowed-origins` to  [`corsOrigins`](https://hackage.haskell.org/package/wai-cors-0.2.7/docs/Network-Wai-Middleware-Cors.html#t:CorsResourcePolicy), then the library will verify if it's a valid Origin and return the preflight headers accordingly.

---

What I find strange is that the `OPTIONS` implementation in these lines is not done for preflight. Once the request has both `Origin` and `Access-Control-Request-Method` headers (preflight), then the `wai-cors` library will bypass this and send a default request. I couldn't find a way to make the library handle CORS requests for `OPTIONS` (not preflight, that is, only with the `Origin` header) and return by itself the headers. 

In other words, I think you'll need to handle this manually:
- Verify if there's an `Origin: <origin>` sent (otherwise leave as it is)
- Then verify if the `<origin>` is present in `cors-allowed-origins`
- If it is then include `Access-Control-Allow-Origin: <origin>` and (maybe?) `Access-Control-Allow-Credentials: true`
- If not then do not include any of those headers (CORS will fail in the browser).

If you can make the library do it, then all the better.

test/io/test_io.py
```diff
@@ -1214,3 +1214,26 @@ def test_jwt_cache_with_no_exp_claim(defaultenv):
...
+        "Origin": "http://example.com",
+        "Access-Control-Allow-Method": "POST",
+        "Access-Control-Allow-Headers": "Content-Type",
+    }
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.options("/items", headers=headers)
+        assert (
+            response.headers["Access-Control-Allow-Origin"]
+            == "http://example.com, http://example2.com"
```
laurenceisla, line 1238: This response header [must have only one origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#directives), which is the same as the one in the `Origin` request header (if it's allowed). Having both as it is right now will invalidate the preflight (the `wai-cors` library should handle this nicely, except for `OPTIONS` as I mentioned before).

test/io/test_io.py
```diff
@@ -1214,3 +1214,26 @@ def test_jwt_cache_with_no_exp_claim(defaultenv):
...
+    env = {
+        **defaultenv,
+        "PGRST_CORS_ALLOWED_ORIGINS": "http://example.com, http://example2.com",
+    }
+
+    headers = {
+        "Accept": "*/*",
+        "Origin": "http://example.com",
+        "Access-Control-Allow-Method": "POST",
+        "Access-Control-Allow-Headers": "Content-Type",
```
laurenceisla, line 1231:
```suggestion
        "Access-Control-Request-Method": "POST",
        "Access-Control-Request-Headers": "Content-Type",
```

src/PostgREST/Config.hs
```diff
@@ -233,6 +235,7 @@ parser :: Maybe FilePath -> Environment -> [(Text, Text)] -> RoleSettings -> Rol
 parser optPath env dbSettings roleSettings roleIsolationLvl =
   AppConfig
     <$> parseAppSettings "app.settings"
+    <*> (fromMaybe "*" <$> optString "cors-allowed-origins")
```
laurenceisla, line 238: Maybe it's better to use the `splitOnCommas` function here (see other examples for other configs below). The type of this AppConfig could be `Maybe [Text]`, if it's not set then it means `Nothing` for all origns (`*`).

test/io/test_io.py
```diff
@@ -1214,3 +1214,47 @@ def test_jwt_cache_with_no_exp_claim(defaultenv):
...
+    headers = {
+        "Accept": "*/*",
+        "Origin": "http://example.com",
+    }
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.options("/items", headers=headers)
+        assert (
+            response.headers["Access-Control-Allow-Origin"]
+            == "*"  # not sure, maybe "http://example.com" ?
```
laurenceisla, line 1259: Yeah, if the allowed origins are only "http://example.com, http://example2.com" then `*`  is not accurate.

test/io/test_io.py
```diff
@@ -1214,3 +1214,47 @@ def test_jwt_cache_with_no_exp_claim(defaultenv):
...
+    with run(env=env) as postgrest:
+        response = postgrest.session.options("/items", headers=headers)
+        assert (
+            response.headers["Access-Control-Allow-Origin"] == "http://example.com"
+            and response.headers["Access-Control-Allow-Credentials"] == "true"
+        )
+
+
+def test_options_no_preflight_request_with_cors_allowed_origin_config(defaultenv):
+    "OPTIONS no preflight request should return Access-Control-Allow-Origin equal to origin"
```
laurenceisla, line 1243: I think a `GET` should also be tested for a no-preflight request. Maybe just add it to this very test and just change its name.

Also, I think we need another test verifying that the "Acess-Control-Allow-Origin" is not returned when an unknown Origin is sent in the request.

taimoorzaeem:
@laurenceisla 

With config's type as `Maybe [Text]` if the config is not specified by the user then it would evaluate to `Nothing`. However if the user specifies this config as `cors-allowed-origins = ""` it would evaluate to `Just []`.

Could this be a problem? If 
```haskell
{ Wai.corsOrigins = Just ([], True)
```
and
```haskell
{ Wai.corsOrigins = Nothing
```
both return `Access-Control-Allow-Origin: *` then it should be fine I think. WDYT?

laurenceisla (approved):
> both return Access-Control-Allow-Origin: * then it should be fine I think. WDYT?

Yes, I agree. Not setting anything or leaving it empty means "*", or any origin allowed. This also needs to be documented to avoid any confusion.

--- 

LGTM! Good job! The only thing that is missing is to tidy up the OPTIONS requests to better use the middleware. But I believe that is a bigger problem with the implementation of OPTIONS itself, so it's outside of the scope of this PR.

test/io/configs/no-defaults.config
```diff
@@ -1,3 +1,4 @@
+cors-allowed-origins = "http://example.com"
```
steve-chavez, line 1: @taimoorzaeem It would be very useful to have this as an [in-database config](https://postgrest.org/en/stable/references/configuration.html#in-database-configuration) too.

steve-chavez: Also, I think I might have given the config a bad name on the issue. My bad. We have the `server-` prefix for all server related configs. So it would make more sense as `server-cors-allowed-origins`, then in-db `server_cors_allowed_origins`.

steve-chavez (approved):

steve-chavez:
@taimoorzaeem Awesome work! :rocket: 

You can document it on https://postgrest.org/en/stable/references/api/cors.html

PR #2991, opened 2023-10-05
fix: range request with 0 rows and 0 offset return status 416

### LINKED ISSUE #2824: Range request with first position the same as length full response should return 416

Strengthless:
Closes #2824.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2771, Add `Server-Timing` header with JWT duration - @taimoorzaeem
  - #2698, Add config `jwt-cache-max-lifetime` and implement JWT caching - @taimoorzaeem
  - #2943, Add `handling=strict/lenient` for Prefer header - @taimoorzaeem
+ - #2824, Fix range request with 0 rows and 0 offset return status 416 - @strengthless
 
 ## [11.2.1] - 2023-10-03
 
--- a/src/PostgREST/RangeQuery.hs
+++ b/src/PostgREST/RangeQuery.hs
@@ -104,9 +104,9 @@ rangeStatusHeader topLevelRange queryTotal tableTotal =
     rangeStatus :: Integer -> Integer -> Maybe Integer -> Status
     rangeStatus _ _ Nothing = status200
     rangeStatus lower upper (Just total)
-      | lower >= total && lower /= upper = status416 -- 416 Range Not Satisfiable
-      | (1 + upper - lower) < total      = status206 -- 206 Partial Content
-      | otherwise                        = status200 -- 200 OK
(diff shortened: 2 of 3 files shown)


taimoorzaeem:
@Strengthless How about adding a test case for this in `test/spec/Feature/Query/RangeSpec.hs`?
```haskell
it "does not throw error when offset is 0 and and total is 0" $
          request methodGet "/rpc/getitemrange?min=0&max=0"
                  (rangeHdrsWithCount $ ByteRangeFromTo 0 1) mempty
            `shouldRespondWith`
            [json|[]|]
            { matchStatus  = 200
            , matchHeaders = ["Content-Range" <:> "*/0"]
            }
```

steve-chavez (approved):

steve-chavez:
No idea why but after merging this PR several jobs started failing on CI:

- https://github.com/PostgREST/postgrest/actions/runs/6438678312/job/17491700303
- https://github.com/PostgREST/postgrest/actions/runs/6438678312/job/17491700050
- https://github.com/PostgREST/postgrest/actions/runs/6438678312/job/17491700181

There must be some non-determinism going on those builds.

cc @laurenceisla

laurenceisla:
@steve-chavez It looks like it's failing to build the [deferred-folds 0.9.18.4]
(https://hackage.haskell.org/package/deferred-folds) library. The NixOS version is 0.9.18.3, so maybe this error isn't present there yet?

There's an open issue in its repo with the same error logs as our builds: https://github.com/nikita-volkov/deferred-folds/issues/6, so the fix should be upstream.

PR #3003, opened 2023-10-12
ci: introduce nightly release job

### LINKED ISSUE #2997: Revive nightly builds
## Problem

Stable releases take longer to release but users might want to try new features right away ([ref1](https://github.com/PostgREST/postgrest/pull/2900#issuecomment-1706993994), [ref2](https://github.com/PostgREST/postgrest/issues/2824#issuecomment-1751705392)).

Pre-releases ([example](https://github.com/PostgREST/postgrest/releases/tag/v10.2.0.20230407)) require manual intervention and also generate some "notification fatigue".

## Solution

Revive nightly builds, do them on every merge to main. We can add the binaries to the nightly pre-release [here](https://github.com/PostgREST/postgrest/releases/tag/nightly) plus docker images. They have the form of `postgrest-nightly-<date>-<shortcommit>-<platform>.tar.xz`.

Examples:

- https://github.com/PostgREST/postgrest/releases/tag/nightly
- https://hub.docker.com/layers/postgrest/postgrest/nightly-2021-07-06-11-19-0961a58/images/sha256-5b55b97db41dbc8f2cae21cef17225f44a274aac3b43501db3f56dc0a35d86f6

## Notes

Currently the binaries are uploaded to the artifacts page: https://github.com/PostgREST/postgrest/actions/runs/6436125768. But somehow they cannot be downloaded unless you're logged in to github, also they're not as visible. So they could be removed from that page and only be uploaded to the nightly pre-release.

develop7:
This PR introduces packaging and uploading successfully built artifacts to `nightly` GitHub release right away, before/without passing test suite.

fixes #2997

### DIFF
--- a/.github/actions/release-nightly/action.yaml
+++ b/.github/actions/release-nightly/action.yaml
@@ -1,0 +1,50 @@
+name: Release Nightly
+
+inputs:
+  artifact:
+    description: 'The name of the artifact to process'
+    required: true
+  release:
+    description: 'The name of the release to upload to'
+    required: false
+    default: 'nightly'
+  platform:
+    description: 'The platform to use for the archive name'
+    required: true
+  format:
+    description: 'The format to use for the archive name'
+    required: false
+    default: 'tar'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Download artifact 
+      uses: actions/download-artifact@v2
+      with:
+        name: ${{ inputs.artifact }}
+    - name: Chmod artifact
+      run: 'chmod +x ${{ inputs.artifact }}'
+    - name: Get current date & time
+      id: datetime
+      run: echo "datetime=$(date +%Y-%m-%d-%H-%M)" >> "$GITHUB_OUTPUT"
+    - name: Render archive name
+      id: archive_name
+      run: echo "::set-output name=archive_name::postgresql-nightly-${{ steps.datetime.outputs.datetime }}-${{ github.sha }}-${{ inputs.platform }}.${{ inputs.format }}"
+    - name: Compress tarball
+      if: ${{ inputs.format == 'tar' }}
+      run: 'tar -cJvf ${{ steps.archive_name.outputs.archive_name }}.tar.xz ${{ inputs.artifact }}'
+    - name: Compress zip
+      if: ${{ inputs.format == 'zip' }}
+      run: 'zip ${{ steps.archive_name.outputs.archive_name }}.zip ${{ inputs.artifact }}'
+    - name: Upload to release
+      uses: softprops/action-gh-release@v0.1.15
+      with:
+        files: |
+          ${{ steps.archive_name.outputs.archive_name }}.tar.xz
+          ${{ steps.archive_name.outputs.archive_name }}.zip
+        name: ${{ inputs.release }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/actions/release-nightly/action.yaml
+++ b/.github/actions/release-nightly/action.yaml
@@ -34,7 +34,8 @@ runs:
       shell: bash
     - name: Render archive name
       id: archive_name
-      run: echo "::set-output name=archive_name::postgresql-nightly-${{ steps.datetime.outputs.datetime }}-${{ github.sha }}-${{ inputs.platform }}.${{ inputs.format }}"
+      run: |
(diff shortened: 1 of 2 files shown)


develop7:
> * uploading the asset fails like this — [PostgREST/postgrest/actions/runs/6576825916/job/17867090122?pr=3003#step:7:119](https://github.com/PostgREST/postgrest/actions/runs/6576825916/job/17867090122?pr=3003#step:7:119). I've created a ticket with GitHub support

Support got back to me  to me and pointed out that PR actions are executed in context of PR creator, so no write access to postgrest/postgrest's releases. However, I can write to my own releases, which I've just tried in a6177ab

develop7:
Anyway, I've just tested uploading arbitrary files to specific release with GH API and personal access token with "Contents" permission and it worked like a charm, see `README.md` in https://github.com/develop7/postgrest/releases/tag/nightly

The "zip" issue might require extracting "upload to nightly release" to separate step, so it would run on latest ubuntu all the time and have `zip` all the time too.

laurenceisla:
> so no write access to postgrest/postgrest's releases.

Yes, the release should not be executed in that part of the process, otherwise, this will be executed any time a PR is done for instance. It should only trigger when it's merged to `main`. You could add a conditional there, but I think it's better to adapt the `Prepare-Release` and `Release-Github` jobs.

For example, `Prepare-Release` needs an additional check to verify that we're on the main branch: 

```yaml
if: {{ startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main' }}
```

In its body you can use extra conditionals to verify if it's a nightly release.

```bash
if [ $GITHUB_REF == "refs/heads/main" ]; then
    echo "isnightly=1" >> "$GITHUB_OUTPUT" #this can be used as output for other jobs
fi

# Then conditionally set version to 'nightly' using

if [ -n "$isnightly" ]; then
    echo "version=nightly-$(date +%Y-%m-%d-%H-%M)" >> "$GITHUB_OUTPUT"
elif [ "$tag_version" != "v$cabal_version" ]; then
    echo "Tagged version ($tag_version) does not match the one in postgrest.cabal (v$cabal_version). Aborting release..."
    exit 1
else

#...
```

Then you can use that output for the `Release-Github` job (remember that the Github release won't be `nightly-<date>`, only `nightly`). You won't need the extra `action.yaml` for nightly builds which avoids repeating code (`tar` and `zip` processes). I think it's OK that it waits for tests to pass.

I'll give you a hand with the docker release, since I have access to the ARM server in [my test repo](https://github.com/laurenceisla/github-actions-test).

laurenceisla:
BTW, I added a PR #3051 that supports docker deployments for nightly releases. WDYT about my proposal? Feel free to let me know if you find something amiss, or if you have any doubts.

develop7:
@laurenceisla thank you for the review and the input. Let me cut to the chase first — CI jobs of this PR cannot upload files to this repo's nightly release because the creator of this PR, me, doesn't have sufficient permissions for this. Moreover, somehow this job is also unable to upload files to my personal fork's releases. In order to have those permissions, I think I have to be promoted to, I guess, a maintainer role (or a custom role with release management permissions, but it requires GitHub Enterprise). Since it's probably too soon for me, I suggest adopting this PR and respective code branch by https://github.com/PostgREST/postgrest/ and someone with enough permission to test it to work.

> Yes, the release should not be executed in that part of the process, otherwise, this will be executed any time a PR is done for instance.

I've had issues setting up local run/debugging the pipeline, so triggering a run here is kind of a way to test changes and debug them. Although there seem to be a way to run https://github.com/nektos/act with rootless `podman`, and I'll give it a try later today.

> It should only trigger when it's merged to `main`.

Absolutely, I intend to add the check for `main` branch last thing before merging it. 

> I think it's better to adapt the `Prepare-Release` and `Release-Github` jobs.

I've tried to do this initially, but subtle differences were piling up, so I've decided to pursue it in isolation to see if that's possible at all. And it is, _almost_.

> BTW, I added a PR #3051 that supports docker deployments for nightly releases.

Looks great, it's definitely something I wasn't covered yet.

laurenceisla:
> Although there seem to be a way to run https://github.com/nektos/act with rootless podman, and I'll give it a try later today.

That's really cool! I'll also try it, maybe that'll save me the hurdle of testing these actions by pushing in a different repo.

> I've tried to do this initially, but subtle differences were piling up, so I've decided to pursue it in isolation to see if that's possible at all. And it is, almost.

I still think it's better that way, mostly due to repetition of steps. But yeah, you can test it in isolation and then I could help migrating to these steps, if it is possible.

laurenceisla:
@develop7 Is it OK if I continue from here? Since I have access/permissions it'll be easier for me to test any changes.

develop7:
Yes, absolutely

laurenceisla:
> Is this superseded by #3051 and #3108?

Yes, it is! Closing rn.

Issue #2478, opened 2022-09-19
Move away from static-haskell-nix


robx:
We currently build a static PostgREST executable using https://github.com/nh2/static-haskell-nix. I think we should try moving away from that because

- the project appears to be abandoned
- the dependency makes it very difficult to upgrade nixpkgs

(This isn't really a new idea, I just wanted to file a central issue for tracking this.)

Some references:
- https://github.com/PostgREST/postgrest/issues/2327
- https://github.com/NixOS/nixpkgs/issues/43795
- https://github.com/NixOS/nixpkgs/issues/61575
- https://discourse.haskell.org/t/portable-haskell-executables-anyone-doing-that-possibly-using-nix/4816
- https://github.com/PostgREST/postgrest/pull/2332

robx:
I've done a bit of digging into `pkgs.pkgsStatic` today:

- https://github.com/NixOS/nixpkgs/pull/191915
- https://github.com/NixOS/nixpkgs/issues/191920

It looks like `pkgsStatic` goes quite a bit further than what we need to build a static PostgREST -- it builds a fully static package set, so it requires all PostgREST dependencies to build in a fully static environment, while all we need is the static build artifacts of all dependencies. E.g., this builds `postgresql` in an environment where only static libraries exist, which the build system seems not prepared for. We only need `dontDisableStatic = true` for our dependencies.

benz0li:
The [`glcr.b-data.ch/ghc/ghc-musl`](https://gitlab.b-data.ch/ghc/ghc-musl/container_registry) images may be an alternative.  
👉 https://github.com/benz0li/ghc-musl

robx:
static-haskell-nix has recently seen some work again, with an update to nixos-23.05: https://github.com/nh2/static-haskell-nix/pull/116

PR #3004, opened 2023-10-12
build: add cabal.project.freeze

### LINKED ISSUE #3000: Cabal builds in CI are failing
### Description of issue

https://github.com/PostgREST/postgrest/actions/runs/6482755722/job/17602892437#step:7:1503, https://github.com/PostgREST/postgrest/actions/runs/6472292459/job/17572620592#step:7:1503

Also confirming the issue is reproduced locally:

image

Suggestions for remediation: 

1. ship `cabal.freeze` with deferred-bounds working for us (from previous `cabal update`)
2. limit `deferred-folds`' bounds?
3. um, upgrade what, `hasql-something` that depends on offending package?

develop7:
Introduces `cabal.project.freeze` to fix build errors caused by `deferred-folds` upgrade, "GHC on Linux" jobs in particular

fixes #3000

### DIFF
--- a/cabal.project.freeze
+++ b/cabal.project.freeze
@@ -1,1 +1,1 @@
-constraints: any.deferred-folds ==0.9.18.3
+index-state: hackage.haskell.org 2023-10-13T13:54:33Z
--- a/cabal.project.freeze
+++ b/cabal.project.freeze
@@ -1,1 +1,1 @@
-index-state: hackage.haskell.org 2023-03-25T00:00:00Z
+index-state: hackage.haskell.org 2023-03-25T23:59:59Z
--- a/cabal.project.freeze
+++ b/cabal.project.freeze
@@ -1,1 +1,1 @@
-index-state: hackage.haskell.org 2023-03-25T23:59:59Z
+index-state: hackage.haskell.org 2023-10-19T09:31:29Z
--- a/cabal.project.freeze
+++ b/cabal.project.freeze
@@ -1,0 +1,1 @@
+constraints: any.deferred-folds ==0.9.18.3
--- a/cabal.project.freeze
+++ b/cabal.project.freeze
@@ -1,1 +1,1 @@
-index-state: hackage.haskell.org 2023-10-19T09:31:29Z
+index-state: hackage.haskell.org 2023-10-13T13:54:33Z
--- a/cabal.project.freeze
+++ b/cabal.project.freeze
@@ -1,1 +1,1 @@
-index-state: hackage.haskell.org 2023-10-13T13:54:33Z
+index-state: hackage.haskell.org 2023-03-25T00:00:00Z


steve-chavez:
@develop7 Q: Why did CI on main started succeeding (all green) ?

develop7:
@steve-chavez there's https://hackage.haskell.org/package/deferred-folds-0.9.18.5, that must be the case

develop7:
yep, https://github.com/nikita-volkov/deferred-folds/commit/93639847f7031c809ad071c15e2d0fcc5849f437

steve-chavez:
@develop7 Ah, I see. Do we stil need the cabal.freeze then?

develop7:
we kind of do to prevent the issue from reappearing in future, but can't deny it did work before quite smoothly

we might just get away with the single `index-state: <now>` stanza in cabal.project.freeze, which should work for both cabal-only builds

let me see

develop7:
@steve-chavez just tested that `index-state: <yesterday>` picks up `deferred-folds-0.9.18.4`. Let's got with today's state for now. 

I also probably need to make a note about this file somewhere around README or something, right? What would be a good place for that?

steve-chavez:
> just tested that index-state: <yesterday> picks up deferred-folds-0.9.18.4. Let's got with today's state for now.

Would it make sense to pin it to the same date as the nixpkgs version?

https://github.com/PostgREST/postgrest/blob/dc01c748ae8da241c8690a3b56a2e2d56f31096d/nix/nixpkgs-version.nix#L2-L6

---

Btw, `index-state` seems like a good practice ([ref](https://stackoverflow.com/questions/75950258/benefits-of-mentioning-index-state-in-cabal-project-file)).

develop7:
> Would it make sense to pin it to the same date as the nixpkgs version?

Absolutely, I've been thinking of syncing with nixpkgs as well, and this seems the way.

Will do tomorrow

develop7:
> Would it make sense to pin it to the same date as the nixpkgs version?

It would, but it causes GHC builds fail: https://github.com/PostgREST/postgrest/actions/runs/6534032019/job/17740379678?pr=3004 https://github.com/PostgREST/postgrest/actions/runs/6534032019/job/17740380228?pr=3004

develop7:
@steve-chavez somehow nixpkgs from `2023-03-25` (_Mar_ 25, 2023),  satisfies the `hasql-pool >= 0.10` requirement, while `hasql-pool` 0.10 was [uploaded **May** 13, 2023](https://hackage.haskell.org/package/hasql-pool-0.10).

develop7:
Yep, it's 0.8.0.7, see below

https://github.com/NixOS/nixpkgs/blob/dbf5322e93bcc6cfc52268367a8ad21c09d76fea/pkgs/development/haskell-modules/hackage-packages.nix#L133705-L133712

steve-chavez:
@develop7 Hm, I'm not sure. I would really like to go all the way with Nix and drop the cabal builds on CI. That way we don't have to worry about these sort of issues again. 

The main thing stopping us is the the `aarch64` static build, which didn't work before (don't know how are things now).

Related: https://github.com/PostgREST/postgrest/issues/2478

steve-chavez:
> Let's got with today's state for now.

As for this PR. It should be fine to go back to the previously working index-state ^

PR #3005, opened 2023-10-12
test: add tests for batch upserts

taimoorzaeem:
Related to #2926.

### DIFF
--- a/test/spec/Feature/Query/PlanSpec.hs
+++ b/test/spec/Feature/Query/PlanSpec.hs
@@ -194,6 +194,62 @@ spec actualPgVersion = do
         resStatus `shouldBe` Status { statusCode = 200, statusMessage="OK" }
         totalCost `shouldBe` 1.29
 
+    it "outputs the total cost for 2 upserts" $ do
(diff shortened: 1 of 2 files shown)


steve-chavez (approved):
LGTM!

test/spec/Feature/Query/PlanSpec.hs
```diff
@@ -194,6 +194,62 @@ spec actualPgVersion = do
         resStatus `shouldBe` Status { statusCode = 200, statusMessage="OK" }
         totalCost `shouldBe` 1.29
 
+    it "outputs the total cost for 2 upserts" $ do
+      r <- request methodPost "/tiobe_pls"
+            [("Prefer","handling=strict; resolution=merge-duplicates"), ("Accept","application/vnd.pgrst.plan+json")]
```
laurenceisla, line 199: There seems to be a typo here. It should be `,` (multiple headers) instead of `;` (for parameters). The resolution is not being applied here.

steve-chavez: I think we don't need the `handling=strict` for this case too

PR #3006, opened 2023-10-13
fix: regression by reverting fix that returned 206 when first-pos=length in Range header

laurenceisla:
Related to: https://github.com/PostgREST/postgrest/issues/2824#issuecomment-1760490793

### DIFF
--- a/test/spec/Feature/Query/RangeSpec.hs
+++ b/test/spec/Feature/Query/RangeSpec.hs
@@ -145,6 +145,14 @@ spec = do
         it "returns whole range with status 200" $
           get "/items" `shouldRespondWith` 200
 
+      context "count with an empty body" $ do
+        it "returns empty body with Content-Range */0" $
+          request methodGet "/items?id=eq.0"
+            [("Prefer", "count=exact")] ""
+            `shouldRespondWith`
+              [json|[]|]
+              { matchHeaders = ["Content-Range" <:> "*/0"] }            
+
       context "when I don't want the count" $ do
         it "returns range Content-Range with /*" $
           request methodGet "/menagerie"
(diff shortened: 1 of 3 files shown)


steve-chavez:
@laurenceisla Just so I understand. Once we merge this. The next step would be to do https://github.com/PostgREST/postgrest/issues/3007 to fix https://github.com/PostgREST/postgrest/issues/2824 ?

test/spec/Feature/Query/RangeSpec.hs
```diff
@@ -145,6 +145,14 @@ spec = do
         it "returns whole range with status 200" $
           get "/items" `shouldRespondWith` 200
 
+      context "count with an empty body" $ do
+        it "returns empty body with Content-Range */0" $
```
laurenceisla, line 149: This test breaks on v11.2.1

test/spec/Feature/Query/RangeSpec.hs
```diff
@@ -211,11 +219,24 @@ spec = do
...
-            { matchHeaders = ["Content-Range" <:> "0-14/*"] }
+      context "succeeds if offset equals 0 as a no-op" $ do
+        it  "no items" $ do
+          get "/items?offset=0&id=eq.0"
+            `shouldRespondWith`
+              [json|[]|]
+              { matchHeaders = ["Content-Range" <:> "*/*"] }
+
+          request methodGet "/items?offset=0&id=eq.0"
+            [("Prefer", "count=exact")] ""
```
laurenceisla, line 230: This also breaks on v11.2.1

test/spec/Feature/Query/RangeSpec.hs
```diff
@@ -211,11 +219,24 @@ spec = do
...
             }
 
-      it "succeeds if offset equals 0 as a no-op" $
-        get "/items?select=id&offset=0&order=id"
-          `shouldRespondWith`
-            [json|[{"id":1},{"id":2},{"id":3},{"id":4},{"id":5},{"id":6},{"id":7},{"id":8},{"id":9},{"id":10},{"id":11},{"id":12},{"id":13},{"id":14},{"id":15}]|]
-            { matchHeaders = ["Content-Range" <:> "0-14/*"] }
+      context "succeeds if offset equals 0 as a no-op" $ do
+        it  "no items" $ do
+          get "/items?offset=0&id=eq.0"
```
laurenceisla, line 224: But this does **not** break on v11.2.1

laurenceisla (commented):
It looks like `ra-data-postgrest` uses the `Prefer: count=exact` header by default. 

https://github.com/raphiniert-com/ra-data-postgrest/blob/cc3d412c3789bdd87bfd792f629631d9bdab9f54/src/index.ts#L137-L144

That's why the test I added for only `offset=0` did not break for v11.2.1, but I'll leave it just in case.

steve-chavez (approved):

PR #3009, opened 2023-10-14
perf: make the Null Filtering query more performant

### LINKED ISSUE #2961: Make the Null Filtering query more performant
After the discussion that started in
https://github.com/PostgREST/postgrest/pull/2951#issuecomment-1720780468, the conclusion is that changing `table IS DISTINCT FROM NULL` to `table.join_column IS NOT NULL` should be more performant because PostgreSQL would treat that last one as an INNER JOIN.

Perhaps we need to change how we build the embedding sub queries to allow this. The `IS DISTINCT FROM` is specified here:

https://github.com/PostgREST/postgrest/blob/add4dfeed5b1d5aa6028a053c921953f04af825d/src/PostgREST/Query/SqlFragment.hs#L340

laurenceisla:
Closes #2961

- [x] Convert 'table is distinct from null' to 'join_col is not null' 
- [ ] Adapt subquery to new filter
- [ ] Verify if it can only be done for `to-one` relationships

### DIFF
--- a/test/spec/Feature/Query/RelatedQueriesSpec.hs
+++ b/test/spec/Feature/Query/RelatedQueriesSpec.hs
@@ -258,8 +258,10 @@ spec = describe "related queries" $ do
         , matchHeaders = [matchContentTypeJson]
         }
 
-    -- "?table=not.is.null" does a "table IS DISTINCT FROM NULL" instead of a "table IS NOT NULL"
-    -- https://github.com/PostgREST/postgrest/issues/2800#issuecomment-1720315818
+    -- "?table=not.is.null" translates to:
+    -- * "table.join_column IS NOT NULL" for to-one relationships
+    -- * "table IS NOT NULL" for to-many relationships
+    -- https://github.com/PostgREST/postgrest/issues/2961
     it "embeds verifying that the entire target table row is not null" $ do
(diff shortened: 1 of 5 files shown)


test/spec/Feature/Query/RelatedQueriesSpec.hs
```diff
@@ -258,8 +258,10 @@ spec = describe "related queries" $ do
         , matchHeaders = [matchContentTypeJson]
         }
 
-    -- "?table=not.is.null" does a "table IS DISTINCT FROM NULL" instead of a "table IS NOT NULL"
-    -- https://github.com/PostgREST/postgrest/issues/2800#issuecomment-1720315818
+    -- "?table=not.is.null" translates to:
+    -- * "table.join_column IS NOT NULL" for to-one relationships
+    -- * "table IS NOT NULL" for to-many relationships
```
laurenceisla, line 263: The query for "to-many" rels forces me to use a `GROUP BY`, the original related issue also happened for these cases only. If no better query exists then it will be left as `table IS NOT NULL` or `table IS DISTINCT FROM NULL`.

laurenceisla:
I'm failing to make a clean implementation for this. I'll close this PR for now.

PR #3024, opened 2023-10-24
feat: add timezone in Prefer header

### LINKED ISSUE #2799: `Prefer: timezone` header
## Problem

No current easy way to change the timezone unless you create new roles:

```sql
alter role la_customers set time zone 'America/Los_Angeles';
```

([Ref](https://github.com/orgs/supabase/discussions/14363))

## Solution

**Edit**: `Prefer: timezone` would be better as mentioned on https://github.com/PostgREST/postgrest/issues/2799#issuecomment-1745835316

Use the `Accept-Timezone: America/Los_Angeles` to change the timezone. 

The implementation is easy as pg does all the work. Just get the header and:

```sql
begin;
set local time zone 'America/Los_Angeles';
```

Reject with a `406 Not Acceptable` if the timezone is not applicable.

## Notes

The `Accept-Timezone` is not standard. Its inclusion has been debated for a long time:

> You are suggesting that this be sent on every HTTP request? Is that correct?
> If you multiply 46 bytes (the line above) up by the number of HTTP
> requests issued across the web every day (billions? trillions?), you can
> imagine how much more data transfer this proposal would require.

https://groups.google.com/g/mozilla.dev.platform/c/hyQQbC3OC-I

That might not be a concern for us...

Alternatively we can use a custom `X-Timezone`  header.

taimoorzaeem:
This should fix #2799.


taimoorzaeem:
```haskell
    context "check behaviour of Prefer: timezone=America/Los_Angeles" $
          it "should change timezone" $
            request methodGet "/timezone_values?select=*"
            [("Prefer", "handling=strict, timezone=America/Los_Angeles")]
            ""
            `shouldRespondWith`
            [json|[{"t":"2023-10-18T05:37:59.611-07:00"}, {"t":"2023-10-18T07:37:59.611-07:00"}, {"t":"2023-10-18T09:37:59.611-07:00"}]|]
            { matchStatus = 200
            , matchHeaders = [matchContentTypeJson] }
```
Strangely, this test is failing. But, if I run this test locally with `curl` it works fine.
```
$ postgrest-with-postgresql-15 postgrest-run
```
```
$ curl 'localhost:3000/timezone_values' -H "Prefer: handling=strict, timezone=America/Los_Angeles"

[{"t":"2023-10-18T05:37:59.611-07:00"}, 
 {"t":"2023-10-18T07:37:59.611-07:00"}, 
 {"t":"2023-10-18T09:37:59.611-07:00"}]

$ curl 'localhost:3000/timezone_values' -H "Prefer: handling=strict, timezone=America/New_York"

[{"t":"2023-10-18T08:37:59.611-04:00"}, 
 {"t":"2023-10-18T10:37:59.611-04:00"}, 
 {"t":"2023-10-18T12:37:59.611-04:00"}]
```

@steve-chavez @laurenceisla Is there something missing in the `spec` tests? Should I move these to `io-tests`.

src/PostgREST/Query.hs
```diff
@@ -273,3 +276,13 @@ runPreReq conf = lift $ traverse_ (SQL.statement mempty . stmt) (configDbPreRequ
...
       HD.noResult
       (configDbPreparedStatements conf)
+
+-- | Sets Timezone from prefer header
+setTimezone :: TimezoneNames -> Preferences -> DbHandler ()
+setTimezone allowedTimezones preferences = 
+  lift $
+    when (isJust $ L.find (== timezone) allowedTimezones)
+    $ do
+      SQL.sql $ "set local time zone '" <> T.encodeUtf8 timezone <> "'"
```
steve-chavez, line 286: To avoid another query (another roundtrip from PostgREST to the db), this can be on [setPgLocals](https://github.com/PostgREST/postgrest/blob/1e0b205e01312b191959a70bd63eb8089c2fe0a3/src/PostgREST/Query.hs#L238-L240). As:

```sql
select set_config('timezone', 'Europe/Berlin', true);
```

taimoorzaeem:
@steve-chavez Hmm, the `io-tests` are running fine locally however, on CI they are failing. Strange!

taimoorzaeem:
@steve-chavez @laurenceisla 
```python
def test_prefer_timezone_with_invalid_timezone(defaultenv):
    "timezone=America/XXX should set timezone to default timezone"

    env = {**defaultenv, "PGRST_DB_CONFIG": "true", "PGRST_JWT_SECRET": SECRET}

    headers = {
        "Prefer": "handling=strict, timezone=America/XXX",
    }

        response = postgrest.session.get("/timezone_values", headers=headers)
        response_body = '[{"t":"2023-10-18T17:37:59.611+05:00"}, \n {"t":"2023-10-18T19:37:59.611+05:00"}, \n {"t":"2023-10-18T21:37:59.611+05:00"}]'
        assert response.text == response_body
```
This test sets the timezone to default timezone for invalid timezones. On my system, we have PKT timezone GMT+05 and so the tests are working fine locally. However, on github CI the default timezone is `+00:00` apparently. The local timezone is different for every developer here and so this would cause the `io-tests` to fail locally. 
How should I write this test case so it matches the response body irrespective of timezone?

laurenceisla:
@taimoorzaeem 

> How should I write this test case so it matches the response body irrespective of timezone?

I think the easiest way is to set the database to UTC. The timezone should be the same for everyone (+00:00:00) after adding this to the `db_config.sql` file:

```sql
ALTER DATABASE :DBNAME SET TIMEZONE TO 'UTC';
```

src/PostgREST/Query.hs
```diff
@@ -278,3 +281,15 @@ runPreReq conf = lift $ traverse_ (SQL.statement mempty . stmt) (configDbPreRequ
...
       (configDbPreparedStatements conf)
+
+-- | Sets Timezone from prefer header
+setTimezone :: TimezoneNames -> Preferences -> DbHandler ()
+setTimezone allowedTimezones preferences =
+  lift $
+    when (isJust $ L.find (== timezone) allowedTimezones)
+    $ do
+        SQL.sql $ "set local time zone '" <> T.encodeUtf8 timezone <> "'"
+        SQL.sql $ "select set_config('timezone', '" <> T.encodeUtf8 timezone <> "', true)"
```
laurenceisla, line 292: Isn't `set_config(timezone,...)` the only setting required here? As Steve suggested, this should be set alongside the other configurations, here:

https://github.com/PostgREST/postgrest/blob/b2352271194cd0051905fdd73232c6b23b29f825/src/PostgREST/Query.hs#L239-L247

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -138,6 +143,7 @@ fromHeaders allowTxEndOverride headers =
 
     prefHeaders = filter ((==) HTTP.hPrefer . fst) headers
     prefs = fmap BS.strip . concatMap (BS.split ',' . snd) $ prefHeaders
+    timezonePref = [ BS.drop 9 p | p <- prefs, BS.take 9 p == "timezone="]
```
laurenceisla, line 146: To keep it DRY, maybe use:

```suggestion
    timezonePref = [ BS.drop 9 p | p <- prefs, hasTimezone p]
    hasTimezone p = BS.take 9 p == "timezone="
```

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -123,7 +127,8 @@ fromHeaders allowTxEndOverride headers =
     , preferTransaction    = if allowTxEndOverride then parsePrefs [Commit, Rollback] else Nothing
     , preferMissing        = parsePrefs [ApplyDefaults, ApplyNulls]
     , preferHandling       = parsePrefs [Strict, Lenient]
-    , invalidPrefs         = filter (`notElem` acceptedPrefs) prefs
+    , preferTimezone       = listToMaybe timezonePref -- In "timezone=America/Los_Angeles", drop timezone= and get "America/Los_Angeles"
+    , invalidPrefs         = filter ((/= "timezone=") . BS.take 9) $ filter (`notElem` acceptedPrefs) prefs
```
laurenceisla, line 131: Then:

```suggestion
    , invalidPrefs         = filter (not . hasTimezone) $ filter (`notElem` acceptedPrefs) prefs
```

src/PostgREST/Query.hs
```diff
@@ -235,21 +237,23 @@ optionalRollback AppConfig{..} ApiRequest{iPreferences=Preferences{..}} = do
...
+    methodSql = setConfigWithConstantName ("request.method", iMethod)
+    pathSql = setConfigWithConstantName ("request.path", iPath)
+    headersSql = setConfigWithConstantNameJSON "request.headers" iHeaders
+    cookiesSql = setConfigWithConstantNameJSON "request.cookies" iCookies
     claimsSql = [setConfigWithConstantName ("request.jwt.claims", LBS.toStrict $ JSON.encode claims)]
     roleSql = [setConfigWithConstantName ("role", role)]
     roleSettingsSql = setConfigWithDynamicName <$> roleSettings
     appSettingsSql = setConfigWithDynamicName <$> (join bimap toUtf8 <$> configAppSettings)
+    getTimezone tz = if isJust $ L.find (== T.decodeUtf8 tz) configTimezoneNames then [setConfigWithConstantName ("timezone", tz)] else mempty
+    timezoneSql = maybe mempty getTimezone $ preferTimezone iPreferences
```
laurenceisla, line 254: I think this filter should be done in the Preferences module.

src/PostgREST/ApiRequest/Preferences.hs
laurenceisla, line 146: Right here.

laurenceisla:
Looking good. Just those refactors and it's ready to merge.

taimoorzaeem:
> BTW. Do you know why is it giving those IO errors?

Nope, I don't know why IO-tests is failing on some pg versions.

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -6,25 +6,30 @@
...
-  , PreferHandling(..)
   , fromHeaders
   , shouldCount
   , prefAppliedHeader
   ) where
 
 import qualified Data.ByteString.Char8     as BS
+import qualified Data.List                 as L
 import qualified Data.Map                  as Map
+import qualified Data.Text.Encoding        as T
```
laurenceisla, line 28: Just FYI, you don't need to import these since they are already included by Protolude. Calling the functions directly works OK.

test/io/postgrest.py
```diff
@@ -19,17 +19,17 @@
...
 def sleep_until_postgrest_config_reload():
     "Sleep until config reload"
-    time.sleep(0.1)
+    time.sleep(0.2)
 
 
 def sleep_until_postgrest_full_reload():
     "Sleep until schema cache plus config reload"
-    time.sleep(0.2)
+    time.sleep(0.3)
```
laurenceisla, line 32: It looks like the extra query increases a bit the reloading time for the schema cache. Adding the sleep time made the io-tests pass. I'm approving since it doesn't look like a lot. @steve-chavez WDYT?

laurenceisla (approved):

test/io/postgrest.py
steve-chavez: @laurenceisla Seems fine for now.

Although I've been noticing each change to the in-db config query does make the io-tests fail. I wonder if we need to join these queries into one and make them more efficient. 

Anyway, that could be left for later.

steve-chavez: > It looks like the extra query increases a bit the reloading time for the schema cache.

Maybe it's not correct to rerun the in-db config query for the schema cache too.

test/io/test_io.py
```diff
@@ -1294,3 +1294,33 @@ def test_no_preflight_request_with_CORS_config_should_not_return_header(defaulte
...
+    env = {**defaultenv, "PGRST_DB_CONFIG": "true", "PGRST_JWT_SECRET": SECRET}
+
+    headers = {
+        "Prefer": "handling=strict, timezone=Invalid/XXX",
+    }
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.get("/timezone_values", headers=headers)
+        response_body = '[{"t":"2023-10-18T12:37:59.611+00:00"}, \n {"t":"2023-10-18T14:37:59.611+00:00"}, \n {"t":"2023-10-18T16:37:59.611+00:00"}]'
+        assert response.text == response_body
```
steve-chavez, line 1326: @taimoorzaeem Just tried this and it doesn't seem correct according to `Prefer: handling=strict`, we should respond with an error in this case.

For the case of `Prefer: handling=lenient` or when no `handling` is specified, then we can just ignore the `Invalid/XXX` as done here.

test/io/fixtures.sql
```diff
@@ -178,3 +178,13 @@ $$;
...
+
+create table timezone_values (
+  t timestamp with time zone
+);
+grant all on timezone_values to postgrest_test_anonymous;
+
+truncate table timezone_values cascade;
+insert into timezone_values values ('2023-10-18 12:37:59.611000+0000');
+insert into timezone_values values ('2023-10-18 14:37:59.611000+0000');
+insert into timezone_values values ('2023-10-18 16:37:59.611000+0000');
```
steve-chavez, line 190: I found `timezone_values` a bit confusing when reading the tests, I thought the timezones were somehow stored in this table.

How about changing it to `timestamp_values` or just `timestamps`?

taimoorzaeem:
> @taimoorzaeem Just tried this and it doesn't seem correct according to Prefer: handling=strict semantics, we should respond with an error in this case.

@steve-chavez Good catch! The only thing left here I guess is that these tests don't belong to `io-tests`. We should move them to `spec-tests` but that needs refactoring as you stated earlier which could be done in another PR.

laurenceisla:
> The only thing left here I guess is that these tests don't belong to io-tests. We should move them to spec-tests but that needs refactoring as you stated earlier which could be done in another PR.

I'm thinking about this, can the result of the `queryTimezones` be handled by the Schema Cache, instead of using the App Config as it is now? It could be alongside the queries in:

https://github.com/PostgREST/postgrest/blob/f10d4139feb8ff2cd619126a14bc4df559e36e25/src/PostgREST/SchemaCache.hs#L132-L133

If I'm not mistaken, the Spec Tests should work OK then.

steve-chavez:
> I'm thinking about this, can the result of the queryTimezones be handled by the Schema Cache, instead of using the App Config as it is now? Or is there a reason against doing so?

@laurenceisla You're right. So the only reason to use the in-db config for db settings is to not expose them in the OpenAPI output (which uses the scheme cache). Some things like the `roleSettings` we cache in AppConfig should definitely not be exposed but the available timezones should.

---

@taimoorzaeem One more thing, the `Preference-Applied` header is not including the timezone currently:

```
curl localhost:3000/projects -H "Prefer: handling=strict, timezone=Europe/Berlin" -i
HTTP/1.1 200 OK
Preference-Applied: handling=strict
```

This might be easier to test once you move the io tests to spec tests.

steve-chavez:
@taimoorzaeem Also a note. When moving the timezones to the Schema Cache, you can just leave the `toJSON` output as empty for now. Something like:

```haskell
instance JSON.ToJSON SchemaCache where
  toJSON (SchemaCache tabs rels routs reps _ _) = JSON.object [
      "dbTables"          .= JSON.toJSON tabs
    , "dbRelationships"   .= JSON.toJSON rels
    , "dbRoutines"        .= JSON.toJSON routs
    , "dbRepresentations" .= JSON.toJSON reps
    , "dbMediaHandlers"   .= JSON.emptyArray
    , "dbTimezones"       .= JSON.emptyArray
    ]
```

We're working on a pure SQL generator for OpenAPI on https://github.com/PostgREST/postgrest-openapi/. So it's not worth the effort to do it on the Haskell side for now.

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -110,34 +116,39 @@ data Preferences
...
+    , preferTransaction    = if allowTxDbOverride then parsePrefs [Commit, Rollback] else Nothing
     , preferMissing        = parsePrefs [ApplyDefaults, ApplyNulls]
     , preferHandling       = parsePrefs [Strict, Lenient]
+    , preferTimezone       = getTimezoneFromPrefs
     , invalidPrefs         = filter (`notElem` acceptedPrefs) prefs
     }
   where
     mapToHeadVal :: ToHeaderValue a => [a] -> [ByteString]
     mapToHeadVal = map toHeaderValue
+    timezonePrefs = map ((<>) "timezone=" . encodeUtf8) tzNames
```
steve-chavez, line 139: @taimoorzaeem Hm, this looks inefficient. There's like 1K timezones and we're mapping over them every time right? I assume this should be avoidable as we only need a search?

src/PostgREST/Config/Database.hs
```diff
@@ -29,6 +30,7 @@ import Protolude
 
 type RoleSettings     = (HM.HashMap ByteString (HM.HashMap ByteString ByteString))
 type RoleIsolationLvl = HM.HashMap ByteString SQL.IsolationLevel
+type TimezoneNames    = [Text] -- cache timezone names for prefer timezone=
```
steve-chavez, line 33: @taimoorzaeem How about turning this into a [Set](https://hackage.haskell.org/package/containers-0.7/docs/Data-Set.html). I think you only need to apply `member` over it. Should be faster than a List since it uses balanced trees.

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -49,15 +51,17 @@ data Preferences
     , preferTransaction    :: Maybe PreferTransaction
     , preferMissing        :: Maybe PreferMissing
     , preferHandling       :: Maybe PreferHandling
+    , preferTimezone       :: Maybe ByteString
```
steve-chavez, line 54: @taimoorzaeem It'd be more consistent to turn this into a `Maybe PreferTimezone`. Then we can have:

```hs
newtype PreferTimezone = PreferTimezone ByteString

instance ToHeaderValue PreferTimezone where
  toHeaderValue (PreferTimezone bs) = "timezone=" <> bs
```

src/PostgREST/ApiRequest/Preferences.hs
steve-chavez: Avoiding the encoding `encodeUtf8` would be good for perf too. I've suggested to convert `TimezoneNames` to a `Set ByteString` below.

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -139,6 +151,10 @@ fromHeaders allowTxEndOverride headers =
     prefHeaders = filter ((==) HTTP.hPrefer . fst) headers
     prefs = fmap BS.strip . concatMap (BS.split ',' . snd) $ prefHeaders
 
+    hasTimezone p = BS.take 9 p == "timezone="
+    getTimezoneFromPrefs = listToMaybe [ PreferTimezone (BS.drop 9 p) | p <- prefs, hasTimezone p && S.member (BS.drop 9 p) acceptedTzNames]
+    checkPrefs p = p `notElem` acceptedPrefs && BS.drop 9 p `S.notMember` acceptedTzNames
```
steve-chavez, line 156: Will DRY this a bit. I think the `BS.drop 9/BS.take 9` can be removed.

test/memory/memory-tests.sh
```diff
@@ -114,8 +114,8 @@ jsonKeyTest "50M" "POST" "/rpc/leak?columns=blob" "172M"
 jsonKeyTest "50M" "POST" "/leak?columns=blob" "172M"
 jsonKeyTest "50M" "PATCH" "/leak?id=eq.1&columns=blob" "172M"
 
-postJsonArrayTest "1000" "/perf_articles?columns=id,body" "14M"
-postJsonArrayTest "10000" "/perf_articles?columns=id,body" "14M"
+postJsonArrayTest "1000" "/perf_articles?columns=id,body" "15M"
+postJsonArrayTest "10000" "/perf_articles?columns=id,body" "15M"
```
steve-chavez, line 118: I get this locally too. I think the increased memory is because of the cached timezones (~1K).

PR #3032, opened 2023-10-30
Clarify the Transaction-Scoped settings in the logs

### LINKED ISSUE #3019: Transaction-Scoped Settings are unclear on the logs
## Problem

[Transaction-Scoped Settings](https://postgrest.org/en/stable/references/transactions.html#transaction-scoped-settings) show on the logs as:

```sql
select set_config($1, $2, true), set_config($3, $4, true), set_config($5, $6, true), set_config($7, $8, true), set_config($9, $10, true), set_config($11, $12, true), set_config($13, $14, true);
```

Details do show the parameters:

```
2023-10-21 13:40:10.955 -05 [498923] DETAIL:  parameters: $1 = 'search_path', $2 = '"test", "public"', $3 = 'role', $4 = 'postgrest_test_anonymous', $5 = 'request.jwt.claims', $6 = '{"role":"p
ostgrest_test_anonymous"}', $7 = 'request.method', $8 = 'GET', $9 = 'request.path', $10 = '/projects', $11 = 'request.headers', $12 = '{"host":"localhost:3000","accept":"*/*","user-agent":"cur
l/7.81.0"}', $13 = 'request.cookies', $14 = '{}'
```

But this level of verbosity is not always enabled. Also they're not clear in `pg_stat_statements`.

## Solution

The first parameter of the `set_config` doesn't need to be parametrized, as it's always defined by us.

So we can convert them to:

```sql
select set_config('search_path', $1, true), set_config('role', $2, true), set_config('request.jwt.claims', $3, true), set_config('request.method', $4, true) ...
```

laurenceisla:
Closes #3019.

As it was mentioned in https://github.com/PostgREST/postgrest/issues/3019#issuecomment-1785725881, the first commit removes support for Legacy GUCs by removing the `db-use-legacy-gucs` config.


src/PostgREST/Query.hs
```diff
@@ -238,37 +234,23 @@ optionalRollback AppConfig{..} ApiRequest{iPreferences=Preferences{..}} = do
...
-    roleSettingsSql = setConfigLocal mempty <$> roleSettings
-    appSettingsSql = setConfigLocal mempty <$> (join bimap toUtf8 <$> configAppSettings)
+    methodSql = setConfigLocalHardcode ("request.method", iMethod req)
+    pathSql = setConfigLocalHardcode ("request.path", iPath req)
+    headersSql = setConfigLocalJson "request.headers" (iHeaders req)
+    cookiesSql = setConfigLocalJson "request.cookies" (iCookies req)
+    claimsSql = [setConfigLocalHardcode ("request.jwt.claims", LBS.toStrict $ JSON.encode claims)]
+    roleSql = [setConfigLocalHardcode ("role", role)]
+    roleSettingsSql = setConfigLocalParam <$> roleSettings
+    appSettingsSql = setConfigLocalParam <$> (join bimap toUtf8 <$> configAppSettings)
```
steve-chavez, line 250: These settings should be safe too as they come from the DBA input.

To be extra safe, I think we can use:

https://github.com/PostgREST/postgrest/blob/3599024e93f6194762c0da158236a5c2dc24d6d2/src/PostgREST/Query/SqlFragment.hs#L158-L166

Maybe let's do this on this PR to close #3019 and add it to the CHANGELOG.

steve-chavez: Doing more thinking though. This code is on the fast path and seems wasteful to do that escaping.

Instead, how about:

- `roleSettings` comes from the db ([here](https://github.com/PostgREST/postgrest/blob/3599024e93f6194762c0da158236a5c2dc24d6d2/src/PostgREST/Config/Database.hs#L128-L161)). We could validate that they don't contain any `'`.
```sql
-- Note that a postgrest identifier ("name") can contain "'", but we can enforce they don't, this should be rarely needed anyway
select E'aa.bb.cc\'.dd'::name;
     name     
--------------
 aa.bb.cc'.dd

-- we just need to filter them in the query so we don't complicate ourselves with failure handling
```

- For `configAppSettings`, we can validate and fail if they contain a `'`.

steve-chavez: It would be nice if we could type those settings as a `newtype SafeLiteral = SafeLiteral Text`.

steve-chavez: Agree. What I suggested seems too complicated and parametrized queries already handle it better.

Then this PR should be enough for solving https://github.com/PostgREST/postgrest/issues/3019 as it is.

steve-chavez: >  For me, it is too much.

Why? Can you elaborate?

Right now I see all settings names being constants.

steve-chavez: @wolfgangwalther I don't see any value being hardcoded, only the keys. This is the result of this PR:

```sql
LOG:  execute 12: select set_config('search_path', $1, true), set_config('role', $2, true), set_config('request.jwt.claims', $3, true), set_config('request.method', $4, true), set_config('request.path', $5, true), set_config('request.headers', $6, true), set_config('request.cookies', $7, true)

DETAIL:  parameters: $1 = '"test", "public"', $2 = 'postgrest_test_anonymous', $3 = '{"role":"postgrest_test_anonymous"}', $4 = 'GET', $5 = '/projects'
, $6 = '{"host":"localhost:3000","accept":"*/*","user-agent":"curl/7.81.0"}', $7 = '{}'
```

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - Removed [raw-media-types config](https://postgrest.org/en/v11.1/references/configuration.html#raw-media-types) - @steve-chavez
  - Removed `application/octet-stream`, `text/plain`, `text/xml` [builtin support for scalar results](https://postgrest.org/en/v11.1/references/api/resource_representation.html#scalar-function-response-format) - @steve-chavez
  - Removed default `application/openapi+json` media type for [db-root-spec](https://postgrest.org/en/v11.1/references/configuration.html#db-root-spec) - @steve-chavez
+ - Removed [db-use-legacy-gucs](https://postgrest.org/en/v11.2/references/configuration.html#db-use-legacy-gucs) - @laurenceisla
```
steve-chavez, line 32: @laurenceisla I think we need a fix entry in the changelog since this already solves https://github.com/PostgREST/postgrest/issues/3019

steve-chavez (approved):

PR #3034, opened 2023-11-01
Display an actual TCP port app is bound to

### LINKED ISSUE #2420: Bogus `Listening on port 0` message, provide useful info
I'm willing to find a random available port for my postgrest service, so I run:
```
PGRST_SERVER_PORT=0 postgrest postgrest.conf
```

It works fine, except then I'm left alone in figuring out which port the kernel assigned to the socket, becuase postgrest just reports:
```
10/Aug/2022:17:31:57 +0200: Listening on port 0
```

Maybe the message should advertise the actual port being used ?

develop7:
Documents and properly handles the special value of `0` for `server-port` config. Before this PR, the value was passed through and handled by the platform's `getaddrinfo` implementation, seemingly resulting in binding to a random port. This PR makes it official and binds to random port explicitly.

Other changes it introduces:

* Simplifies app starting logic, moving platform-dependent POSIX-specific code to `PostgREST.Unix`, making it the only file with the `LANGUAGE CPP` pragma and reducing the amount of indirection. Pretty sure the indirection in question was here on purpose, let me know if that's still the case — tests do pass, things work fine, so I've decided to give it a shot.
* Recent releases of `network` and all Windows released since 5 years ago actually do support Unix domain sockets, so I've taken a chance to prepare code for making use of those: replaced `setFileMode` with its twin function from `unix-compat` and used `Network.Socket.isUnixDomainSocketAvailable` to check such sockets'  availability in runtime. All that's left to pull the plug is to upgrade `network`.
* makes `PostgREST.Unix.installSignalHandlers` more explicit, avoiding potential circular dependency between `PostgREST.Unix` and `PostgREST.AppState`
* slightly decreases test coverage. The cases missed seem to be trivial to me and would require considerable effort to be tested properly.

Fixes #2420

<details><summary>Old TODO</summary>
TODO:

- [ ] test 
- [x] document special `server-port` value of `0`; done in https://github.com/PostgREST/postgrest-docs/pull/695
- [x] get runtime error for Unix socket on unsupported platforms back (right now it degraded to compile error)
- [x] fix Windows build
- [x] ~fix the crash in, presumably, `Admin.reachMainApp`~ I was `bind`ing a socket, but didn't `listen` it :confounded: 

</details>

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -122,7 +122,7 @@ library
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
-                    , unix-compat
+                    , unix-compat >= 0.6 && < 0.7
                     , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
                     , wai                       >= 3.2.1 && < 3.3
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -120,7 +120,7 @@ library
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
-                    , unix-compat               >= 0.7 && < 0.8
+                    , unix-compat               == 0.5.4
                     , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
                     , wai                       >= 3.2.1 && < 3.3
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -120,6 +120,7 @@ library
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
+                    , unix-compat               >= 0.7 && < 0.8
                     , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
                     , wai                       >= 3.2.1 && < 3.3
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -122,7 +122,7 @@ library
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
-                    , unix-compat >= 0.6 && < 0.7
+                    , unix-compat               >= 0.5.4 && < 0.6
                     , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
                     , wai                       >= 3.2.1 && < 3.3
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -114,6 +114,7 @@ library
                     , regex-tdfa                >= 1.2.2 && < 1.4
                     , retry                     >= 0.7.4 && < 0.10
                     , scientific                >= 0.3.4 && < 0.4
+                    , streaming-commons         >= 0.1.1 && < 0.3
                     , swagger2                  >= 2.4 && < 2.9
                     , text                      >= 1.2.2 && < 1.3
                     , time                      >= 1.6 && < 1.12
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -162,7 +162,6 @@ executable postgrest
   main-is:            Main.hs
   build-depends:      base                >= 4.9 && < 4.17
                     , containers          >= 0.5.7 && < 0.7
-                    , network                   >= 2.6 && < 3.2
                     , postgrest
                     , protolude           >= 0.3.1 && < 0.4
   ghc-options:        -threaded -rtsopts "-with-rtsopts=-N -I0 -qg"
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -64,6 +64,7 @@ library
                       PostgREST.Plan.ReadPlan
                       PostgREST.Plan.Types
                       PostgREST.RangeQuery
+                      PostgREST.Unix
                       PostgREST.ApiRequest
                       PostgREST.ApiRequest.Preferences
                       PostgREST.ApiRequest.QueryParams
@@ -120,6 +121,7 @@ library
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
+                    -- pinning unix-compat to match one's version in warp
                     , unix-compat               == 0.5.4
                     , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
@@ -151,8 +153,6 @@ library
     build-depends:
       unix
       , directory >= 1.2.6 && < 1.4
-    exposed-modules:
-      PostgREST.Unix
 
 executable postgrest
   default-language:   Haskell2010
@@ -162,6 +162,7 @@ executable postgrest
   main-is:            Main.hs
   build-depends:      base                >= 4.9 && < 4.17
                     , containers          >= 0.5.7 && < 0.7
+                    , network                   >= 2.6 && < 3.2
                     , postgrest
                     , protolude           >= 0.3.1 && < 0.4
   ghc-options:        -threaded -rtsopts "-with-rtsopts=-N -I0 -qg"
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -122,8 +122,7 @@ library
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
-                    -- pinning unix-compat to match one's version in warp
-                    , unix-compat               == 0.5.4
+                    , unix-compat
                     , vault                     >= 0.3.1.5 && < 0.4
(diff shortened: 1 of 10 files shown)


steve-chavez:
@develop7 Good initiative fixing this!

I see the CI failures. That happens because the `Admin.hs` doesn't keep track of the assigned random port.

I believe the correct fix is the one that Wolfgang mentioned on https://github.com/PostgREST/postgrest/issues/2174#issuecomment-1045978929. Basically store the socket on `AppState` and use it on `Admin.hs`. Then some code can be removed from Admin, like the `getAddrInfo` part.

(I don't recall why I didn't make like that on the first place, but seems it should work)

develop7:
> I believe the correct fix is the one that Wolfgang mentioned on [#2174 (comment)](https://github.com/PostgREST/postgrest/issues/2174#issuecomment-1045978929). Basically store the socket on `AppState` and use it on `Admin.hs`. Then some code can be removed from Admin, like the `getAddrInfo` part.

Yep, on it; the 47df6c2 is more of a PoC

develop7:
BTW since `Warp.runSettings` passes the port number through, it should handle special values like `0` properly as well, by returning actual port number.

develop7:
So `Warp.runSettings` does pass the port number to `streaming-commons`' [`bindPortTCP`](https://www.stackage.org/haddock/lts-21.19/streaming-commons-0.2.2.6/Data-Streaming-Network.html#v:bindPortTCP), which has sister function `bindRandomPortTCP`, which returns port actually bound; we could use it to handle the special case of `0` on our side. Let me explore this for a bit.

develop7:
Not sure whether it's better to store port number along in the `AppState`, along with app socket, or read port number directly out of socket. I chose the latter for the places where actual Warp instances are run, but `initSockets` is currently a bit inconsistent on this matter

steve-chavez:
@develop7 It's starting to look better! Seems you need to add `streaming-commons` to cabal to clear the errors on CI.

> Not sure whether it's better to store port number along in the AppState, along with app socket, or read port number directly out of socket. I chose the latter..

Yeah, the less state we can keep the better.

src/PostgREST/Unix.hs
```diff
@@ -1,44 +1,14 @@
...
--- | Run the PostgREST application with user defined socket.
-runAppWithSocket :: Warp.Settings -> Application -> FileMode -> FilePath -> IO ()
-runAppWithSocket settings app socketFileMode socketFilePath =
-  bracket createAndBindSocket Socket.close $ \socket -> do
-    Socket.listen socket Socket.maxListenQueue
-    Warp.runSettingsSocket settings socket app
-  where
-    createAndBindSocket = do
-      deleteSocketFileIfExist socketFilePath
-      sock <- Socket.socket Socket.AF_UNIX Socket.Stream Socket.defaultProtocol
```
steve-chavez, line 30: Hm, not that this is gone. How are we handling unix sockets?

develop7: We do, in https://github.com/PostgREST/postgrest/pull/3034/files#diff-f6074642dc32a94e9e1997b21b0df631b2d3f098e0a81673edc22b89a86f49edR176

develop7:
Okay, I've just reproduced the crash in `Admin.reachMainApp`. It's probably caused by reusing the `Socket` from `AppState` to `send` to it; one shouldn't do that.

develop7:
https://hackage.haskell.org/package/network-3.1.4.0/docs/Network-Socket.html#v:isUnixDomainSocketAvailable says 

> [AF_UNIX](https://hackage.haskell.org/package/network-3.1.4.0/docs/Network-Socket.html#v:AF_UNIX) is supported on Windows since 3.1.3.0. So, this variable is [True](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-Bool.html#v:True) on all platforms.

Does it mean we can have UNIX sockets on Windows?

develop7:
> Does it mean we can have UNIX sockets on Windows?

Per https://github.com/haskell/network/pull/549#issuecomment-1523930822: <q>any supported version of Windows has it</q>. It's a "yes".

src/PostgREST/Unix.hs
develop7: I've had to this code back in order to be able to handle `server-unix-socket` config on unsupported platforms

postgrest.cabal
```diff
@@ -148,9 +153,6 @@ library
   if !os(windows)
     build-depends:
       unix
-      , directory >= 1.2.6 && < 1.4
-    exposed-modules:
-      PostgREST.Unix
```
steve-chavez, line 153: @develop7 Any concerns on windows? Maybe we should allow running tests on windows to be extra sure (other PR).

develop7: That'd be only fair, since we do release windows binaries

postgrest.cabal
```diff
@@ -114,11 +116,14 @@ library
...
                     , retry                     >= 0.7.4 && < 0.10
                     , scientific                >= 0.3.4 && < 0.4
+                    , streaming-commons         >= 0.1.1 && < 0.3
                     , swagger2                  >= 2.4 && < 2.9
                     , text                      >= 1.2.2 && < 1.3
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
+                    -- pinning unix-compat to match one's version in warp
+                    , unix-compat               == 0.5.4
```
steve-chavez, line 126: This dependency pinning seems like it could give trouble later. Can we avoid the dep or the pinning? Why is it needed?

develop7: Good point, unpinning it altogether. Compilation seems to succeed and tests seem to pass

postgrest.cabal
develop7: We cannot enable all tests on Windows right off the bat, but specs & doctests should be fine

postgrest.cabal
```diff
@@ -114,11 +116,13 @@ library
                     , regex-tdfa                >= 1.2.2 && < 1.4
                     , retry                     >= 0.7.4 && < 0.10
                     , scientific                >= 0.3.4 && < 0.4
+                    , streaming-commons         >= 0.1.1 && < 0.3
                     , swagger2                  >= 2.4 && < 2.9
                     , text                      >= 1.2.2 && < 1.3
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
+                    , unix-compat
```
steve-chavez, line 125: Let's try with:
```suggestion
                    , unix-compat >= 0.6 && < 0.7
```

develop7: https://github.com/PostgREST/postgrest/actions/runs/6960807521/job/18941061764#step:4:19 doesn't build

postgrest.cabal
```diff
@@ -114,11 +116,13 @@ library
                     , regex-tdfa                >= 1.2.2 && < 1.4
                     , retry                     >= 0.7.4 && < 0.10
                     , scientific                >= 0.3.4 && < 0.4
+                    , streaming-commons         >= 0.1.1 && < 0.3
                     , swagger2                  >= 2.4 && < 2.9
                     , text                      >= 1.2.2 && < 1.3
                     , time                      >= 1.6 && < 1.12
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
+                    , unix-compat >= 0.6 && < 0.7
```
steve-chavez, line 125:
```suggestion
                    , unix-compat               >= 0.5.4 && < 0.6
```

postgrest.cabal
steve-chavez: Oops, corrected below.

postgrest.cabal
develop7: Yup, that should do it

develop7: That'd did it

src/PostgREST/AppState.hs
```diff
@@ -144,6 +161,39 @@ initWithPool pool conf = do
...
+    Nothing -> do
+      (_, sock) <-
+        if cfg'port /= 0
+          then do
+            sock <- bindPortTCP cfg'port (fromString $ T.unpack cfg'host)
+            pure (cfg'port, sock)
+          else do
+            -- explicitly bind to a random port, returning bound port number
+            (num, sock) <- bindRandomPortTCP (fromString $ T.unpack cfg'host)
+            pure (num, sock)
```
steve-chavez, line 186: @develop7 Codecov is complaining about these lines. Any way we can increase coverage?

How about an io test with `PGRST_SERVER_PORT = 0`?

develop7:
Oh, that would do it? Sure, will try.

src/PostgREST/AppState.hs
develop7: Oh, that would do it? Sure, will try.

develop7: Woo, it's green!

src/PostgREST/Admin.hs
```diff
@@ -53,35 +52,16 @@ admin appState appConfig req respond  = do
 -- Note that it doesn't even send a valid HTTP request, we just want to check that the main app is accepting connections
 -- The code for resolving the "*4", "!4", "*6", "!6", "*" special values is taken from
 -- https://hackage.haskell.org/package/streaming-commons-0.2.2.4/docs/src/Data.Streaming.Network.html#bindPortGenEx
```
steve-chavez, line 54: This is no longer the case right? Will remove.

PR #3044, opened 2023-11-08
fix: typos

omahs:
fix: typos


steve-chavez (approved):
@omahs Thanks for the contrib!

PR #3052, opened 2023-11-16
fix: handle CORS pre-flight requests correctly

### LINKED ISSUE #3027: OPTIONS does not actually handle CORS pre-flight requests completely
## Problem

While adding Origin validation for CORS requests, we found out that our `OPTIONS` implementation does not actually handle pre-flight requests, at least not completely https://github.com/PostgREST/postgrest/pull/2986#discussion_r1346732033.

This is how it works right now:

### Successful request
```bash
$ curl -X OPTIONS "http://localhost:3000/projects" \
        -H "Access-Control-Request-Method: POST" \
        -H "Access-Control-Request-Headers: Content-Type" \
        -H "Origin: http://localhost" -i
```
```http
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Date: Fri, 27 Oct 2023 19:05:19 GMT
Server: postgrest/11.2.0 (8c9213d)
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, PATCH, PUT, OPTIONS, HEAD, POST
Access-Control-Allow-Headers: Authorization, Content-Type, Accept, Accept-Language, Content-Language
Access-Control-Max-Age: 86400
```
The actual response is given by the `wai-cors` library. We can notice because our implementation does not return pre-flight headers, only `Access-Control-Allow-Origin`. In summary, our `OPTIONS` does nothing here, the library does the response.

### Failing request
```bash
$ curl -X OPTIONS "http://localhost:3000/projects" \
        -H "Access-Control-Request-Method: TRACE" \
        -H "Access-Control-Request-Headers: Content-Type" \
        -H "Origin: http://localhost" -i
```
```http
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Date: Fri, 27 Oct 2023 19:16:15 GMT
Server: postgrest/11.2.0 (8c9213d)
Access-Control-Allow-Origin: *
Allow: OPTIONS,GET,HEAD,POST,PUT,PATCH,DELETE
```
Only when it fails, our `OPTIONS` implementation goes through and returns this info. This happens because we set the library to not respond with an error and let the request continue:
https://github.com/PostgREST/postgrest/blob/5c9c7f4ff4dd0ef659206f1cb17502bb126d8f25/src/PostgREST/Cors.hs#L37
This still fails the pre-flight (in Firefox and Chrome at least) because `Access-Control-Allow-Methods` is not set (different from `Allow` which has no meaning in CORS pre-flight).

### Solution

We have two options here:

- Let the library return the error which is a `400` with a predefined message indicating the problem.
- Implement our own, which would mean to verify if the `OPTIONS` CORS request is pre-flight (with the required headers) and respond with a custom error indicating the problem.

Any of these would liberate the simple `OPTIONS` CORS request (not preflight) to work as we want. For instance, it could bring back the resource structure mentioned in #790.

laurenceisla:
Closes #3027

Lets the library handle errors when a pre-flight request fails.


src/PostgREST/Cors.hs
```diff
@@ -44,3 +44,7 @@ corsPolicy corsAllowedOrigins req = case lookup "origin" headers of
        -- Impossible case, Middleware.Cors will not evaluate this when
        -- the Access-Control-Request-Headers header is not set.
       Nothing   -> []
+    -- The library makes "OPTIONS" requests fail when the "Origin" header is present
```
steve-chavez, line 47: Hm, which library?

laurenceisla: Ah, I failed to mention the `wai-cors` library. 

I added more info here: https://github.com/PostgREST/postgrest/issues/3027#issuecomment-1815405884. The solution is as I mentioned there or patch the library to remove that validation.

laurenceisla: Just confirmed that there's no way to bypass the `OPTIONS` requests without re-implementing the CORS functionality for that method only. So, I opened an issue upstream https://github.com/larskuhtz/wai-cors/issues/37 to see what they think about it.

cabal.project
```diff
@@ -2,3 +2,7 @@ packages: postgrest.cabal
 tests: true
 package *
   ghc-options: -split-sections
+source-repository-package
+  type: git
+  location: https://github.com/laurenceisla/wai-cors.git
+  tag: e2da0d7a3e56592a5f4b35fd3aa6ccd3175525cc
```
laurenceisla, line 8: Using a temporary fork, ideally this would be merged upstream and it should fix allowing using `OPTIONS` with the `Origin` header without returning errors.

PR #3056, opened 2023-11-17
feat: add statement_timeout set on functions

### LINKED ISSUE #3001: Apply functions `statement_timeout` for RPC
## Problem

Pg itself doesn't apply a `statement_timeout` set on a function.

```sql
create or replace function my_func() returns void as $$
  select pg_sleep(3);
$$ language sql set statement_timeout to '1s';

begin;
select my_func();
 my_func 
---------
 
(1 row)
commit;

-- vs

begin;
set local statement_timeout to '2s';
select pg_sleep(3);
ERROR:  canceling statement due to statement timeout
```

It only does it for LOGIN roles plus we do it for [impersonated roles](https://postgrest.org/en/stable/references/auth.html#impersonated-role-settings). 

## Solution

Apply the function `statement_timeout` at the start of the transaction on RPC. This would override the role `statement_timeout`.

We already have similar logic for `default_transaction_isolation`: https://postgrest.org/en/stable/references/transactions.html#isolation-level.

taimoorzaeem:
This should fix #3001.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2825, SQL handlers for custom media types - @steve-chavez
    + Solves #1548, #2699, #2763, #2170, #1462, #1102, #1374, #2901
  - #2799, Add timezone in Prefer header - @taimoorzaeem
+ - #3001, Add `statement_timeout` set on functions - @taimoorzaeem
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


steve-chavez (approved):
Perfect! :100:

steve-chavez:
Will help in documenting this. We need to move https://postgrest.org/en/stable/references/auth.html#impersonated-role-settings into https://postgrest.org/en/stable/references/transactions.html.

steve-chavez:
@taimoorzaeem Would be better to apply all settings instead of only a special case with `statement_timeout`. Opened https://github.com/PostgREST/postgrest/issues/3061 for tracking that.

PR #3064, opened 2023-11-22
feat: implement `server-timing-enabled` config parameter

### LINKED ISSUE #3062: Configuration for server-timing
## Problem

Right now `Server-Timing` header is enabled by [db-plan-enabled](https://postgrest.org/en/stable/references/configuration.html#db-plan-enabled) but it shouldn't. The [execution plan](https://postgrest.org/en/stable/references/admin.html#execution-plan) can be used independently.

https://github.com/PostgREST/postgrest/blob/91d53852370f5bf5ef1ecaa69b9ee7b8435b63a1/src/PostgREST/App.hs#L261-L267

## Solution

Enable `Server-Timing` with a `server-timing-header` configuration instead.

(We have a similar config https://postgrest.org/en/stable/references/configuration.html#server-trace-header)

develop7:
Implements `server-timing-enabled` config parameter controlling the `Server-Timing` header 

Fixes #3062
Documented in https://github.com/PostgREST/postgrest-docs/pull/700


steve-chavez:
@develop7 This looks good! Anything missing to undraft?

develop7:
No, not really, besides undrafting itself

test/io/configs/expected/no-defaults-with-db-other-authenticator.config
```diff
@@ -30,6 +30,7 @@ server-cors-allowed-origins = "http://example.com"
 server-host = "0.0.0.0"
 server-port = 80
 server-trace-header = "traceparent"
+server-timing-enabled = false
```
steve-chavez, line 33: These `test/io/configs/expected/no-defaults..` ones should give true (no default). For these we use settings in `test/io/db_config.sql`.

test/io/configs/expected/no-defaults-with-db.config
```diff
@@ -30,6 +30,7 @@ server-cors-allowed-origins = "http://example.com"
 server-host = "0.0.0.0"
 server-port = 80
 server-trace-header = "CF-Ray"
+server-timing-enabled = false
```
steve-chavez, line 33: ditto for in-db config

src/PostgREST/Config/Database.hs
```diff
@@ -64,6 +64,7 @@ dbSettingsNames =
   ,"openapi_server_proxy_uri"
   ,"raw_media_types"
   ,"server_trace_header"
+  ,"server_timing_enabled"
```
steve-chavez, line 67: Added this for the in-db config to work.

I do realize that it's easy to miss it since it's not part of the Config types. Will open an issue to fix this.

steve-chavez (approved):
Great work! :100:

laurenceisla:
Only thing missing is an entry for the `--dump-config`:

https://github.com/PostgREST/postgrest/blob/df97a5071f243a70226a68cd1cde04cb899f1428/src/PostgREST/CLI.hs#L124-L131

PR #3068, opened 2023-11-24
Change timing name from 'query' to 'transaction' and add ApiRequest timing

### LINKED ISSUE #3055: Split `Server-Timing` query duration
## Problem

Currently there's no way to know how much time the [pre-request](https://postgrest.org/en/stable/references/transactions.html#pre-request) and [tx-settings](https://postgrest.org/en/stable/references/transactions.html#transaction-scoped-settings) are taking.

`Server-Timing` englobes all of them into a single `query` metric:  

```
$ curl 'localhost:3000/projects' -i

HTTP/1.1 200 OK
Server-Timing: jwt;dur=14.6, render;dur=5.7, plan;dur=89.5, query;dur=523.2
```

And the [execution plan](https://postgrest.org/en/stable/references/admin.html#execution-plan) it's only obtained from the main query.

## Solution

Split the `query` metric into `tx-settings`, `pre-request` (only when it's configured) and `main-query`:

```
$ curl 'localhost:3000/projects' -i

HTTP/1.1 200 OK
Server-Timing: tx-settings;dur=x, pre-request;dur=y, main-query;dur=z
```

Basically involves adding timings separately for these steps:

https://github.com/PostgREST/postgrest/blob/a8387e0ad3860e0c6d956fdae9554745e4453f6d/src/PostgREST/App.hs#L253-L256

laurenceisla:
Closes #3055.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -169,51 +169,51 @@ handleRequest AuthResult{..} conf appState authenticated prepared pgVer apiReq@A
   case (iAction, iTarget) of
     (ActionRead headersOnly, TargetIdent identifier) -> do
       (planTime', wrPlan) <- withTiming $ liftEither $ Plan.wrappedReadPlan identifier conf sCache apiReq
-      (rsTime', resultSet) <- withTiming $ runQuery roleIsoLvl Nothing (Plan.wrTxMode wrPlan) $ Query.readQuery wrPlan conf apiReq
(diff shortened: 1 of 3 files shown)


steve-chavez:
Interesting. Looks like the parse time is considerable, even higher than the plan for some cases:

```bash
curl 'localhost:3000/entities?select=*,child_entities(*,grandchild_entities(*))' -I
HTTP/1.1 200 OK

Server-Timing: jwt;dur=14.9, parse;dur=226.5, render;dur=4.6, plan;dur=141.1, transaction;dur=538.0
```

PR #3070, opened 2023-11-28
chore: change server timing render to response

steve-chavez:
`render` is a loaded term than might be thought as the generation of a full HTML page. While for this phase we only process the status and the headers.

Changing it to `response` so is not misleading at least. Users can check the docs for clarification.

### DIFF
--- a/test/spec/Feature/Query/ServerTimingSpec.hs
+++ b/test/spec/Feature/Query/ServerTimingSpec.hs
@@ -14,7 +14,7 @@ spec :: SpecWith ((), Application)
 spec =
   describe "Show Duration on Server-Timing header" $ do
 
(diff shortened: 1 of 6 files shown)


test/spec/Feature/Query/ServerTimingSpec.hs
```diff
@@ -14,15 +14,15 @@ spec :: SpecWith ((), Application)
 spec =
   describe "Show Duration on Server-Timing header" $ do
 
-    context "responds with Server-Timing header" $ do
+    context "responseonds with Server-Timing header" $ do
```
laurenceisla, line 17:
```suggestion
    context "responds with Server-Timing header" $ do
```

laurenceisla (approved):

PR #3071, opened 2023-11-28
ci: merge cached Nix stores

develop7:
Introduces a nix cache warm-up job, which prefetches all the derivations we ever need to the Nix store, which is cached right away. Such cache is reused then in further actions, reducing in-job downloads to zero.

Before this PR every nix-involving job (13 of them on Linux) took 3 to 4 minutes to do "nix setup" due to downloading 121 to 1409 paths, 300 to 1360 MB big. While these jobs cache Nix store here and there, there's a handful of them, multiplied by 2.3-ish GB in size, and thus they exceed the upper cache limit for GH Actions of 10 GB. 
After merging this PR, a cold run takes ~6 minutes to warm up the cache, then every nix-involving job takes 1:30 mins to unpack the cache and have all the derivations installed. A warm run takes 1 minute of a no-op cache warm up step and the same 1:30 to unpack the cache. This produces one Nix store cache per OS of 2.3 GB in size, which fit the storage nicely and would probably be a benefit for every build, for `main` and PR branches as well.

TODO: make the `loadtest` workflow use a common cache as well, maybe by extracting the "warm up" job to a separate workflow.

### DIFF
--- a/.github/actions/setup-nix/action.yaml
+++ b/.github/actions/setup-nix/action.yaml
@@ -20,10 +20,9 @@ runs:
     - name: Restore and cache Nix store
       uses: nix-community/cache-nix-action@v4.0.3
       with:
-        key: cache-nix-${{ runner.os }}-id-${{ inputs.cache-id }}-${{ hashFiles('nix/**/*.nix') }}
+        key: cache-nix-${{ runner.os }}-id-${{ inputs.cache-id }}-${{ hashFiles('nix/**/*.nix', '.github/actions/setup-nix/*') }}
         restore-keys: |
           cache-nix-${{ runner.os }}-common-
-        restore-key-hit: true
     - uses: cachix/cachix-action@v13
       with:
         name: postgrest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,16 +13,32 @@ on:
       - rel-*
 
 jobs:
-  Lint-Style:
-    name: Lint & check code style
+  Prepopulate-Nix-Cache-Linux:
+    name: Prepopulate Nix cache for Linux runners
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Setup Nix Environment
         uses: ./.github/actions/setup-nix
         with:
+          cache-id: common
+      - name: Put all tools to store to be cached afterwards
+        run: |
+          # shellcheck disable=SC2046
+          nix-store -v --realize $( nix-instantiate default.nix )
+        shell: bash
+    
+  Lint-Style:
+    name: Lint & check code style
+    runs-on: ubuntu-latest
+    needs: [Prepopulate-Nix-Cache-Linux]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Nix Environment
+        uses: ./.github/actions/setup-nix
+        with:
           tools: style
-          cache-id: style
+          cache-id: common
       - name: Run linter (check locally with `nix-shell --run postgrest-lint`)
         run: postgrest-lint
       - name: Run style check (auto-format with `nix-shell --run postgrest-style`)
@@ -32,6 +48,7 @@ jobs:
   Test-Nix:
     name: Test (Nix)
     runs-on: ubuntu-latest
+    needs: [Prepopulate-Nix-Cache-Linux]
     defaults:
       run:
         # Hack for enabling color output, see:
@@ -43,7 +60,7 @@ jobs:
         uses: ./.github/actions/setup-nix
         with:
           tools: tests
-          cache-id: test-pg
+          cache-id: common
 
       - name: Run coverage (IO tests and Spec tests against PostgreSQL 15)
         run: postgrest-coverage
@@ -68,6 +85,7 @@ jobs:
         pgVersion: [9.6, 10, 11, 12, 13, 14, 15, 16]
     name: Test PG ${{ matrix.pgVersion }} (Nix)
     runs-on: ubuntu-latest
+    needs: [Prepopulate-Nix-Cache-Linux]
     defaults:
       run:
         # Hack for enabling color output, see:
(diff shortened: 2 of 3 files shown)


develop7:
@laurenceisla @steve-chavez could you guys adopt this PR so we could test if purging individual caches actually works (and cache space is saved) because it looks like I'm being hit by the lack of permissions again (see [here](https://github.com/PostgREST/postgrest/actions/runs/7022126405/job/19105984902?pr=3071#step:12:15))?

.github/actions/setup-nix/action.yaml
```diff
@@ -23,7 +23,6 @@ runs:
         key: cache-nix-${{ runner.os }}-id-${{ inputs.cache-id }}-${{ hashFiles('nix/**/*.nix') }}
         restore-keys: |
           cache-nix-${{ runner.os }}-common-
-        restore-key-hit: true
```
laurenceisla, line 26: Why is this removed? If active a cache won't be created if it hits the key, does that cause a problem?

develop7:
In example workflow at c-n-a repository this parameter is set in "merge
caches" jobs only, so I've followed it. I think the reason was to produce
the individual cache still, so the "merge" job had caches to merge from.

.github/actions/setup-nix/action.yaml
develop7: > it's actually in the `‎make-similar-caches‎`, which is equivalent to our `Setup Nix` I think?

Exactly. My goal was to call the cache job as early as possible so it'll unpack cached store and prevent redownloading store entries, but also after installing nix (it's a prerequisite).

> I tried the PR and gives me the same error:[PostgREST/postgrest/actions/runs/7035767110/job/19147207413#step:12:14](https://github.com/PostgREST/postgrest/actions/runs/7035767110/job/19147207413#step:12:14). It seems that this won't/shouldn't be executed in PRs but when merging on main only.

I think the reason it fails is the `common` cache reference/ID is reserved somewhere (very) early in the run and isn't let go afterward. The fix could be to extract the job to a separate workflow and have it triggered by the, um, schedule, for instance (there are other triggers that look promising, such as a [check_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#check_run)).

laurenceisla: Hmm... I see, I'm kinda lost with how the merging is done or even if it's needed in this case, mostly because only one version of linux `ubuntu-latest` is used (correct me if I'm wrong here, not quite familiarized with that process). 

> I think the reason it fails is the common cache reference/ID is reserved somewhere (very) early in the run and isn't let go afterward.

Yeah, I don't think the `common` cache should be removed (at least in PRs), it is used later in the release too (Release-Docker).

laurenceisla: Ah I assumed it was more relevant for versions due to them having similar caches, it looks like it's clever in finding similar stores alongside different Job steps then. I'll try to update my PR later and see how i works.

develop7:
So prepopulating job takes 6 minutes, produces one cache 2.3G big, all the "nix setup" steps in the nix-involving jobs on Linux take 1:30 mins tops.

.github/actions/setup-nix/action.yaml
```diff
@@ -20,14 +20,19 @@ runs:
...
     - uses: cachix/cachix-action@v12
       with:
         name: postgrest
         authToken: ${{ inputs.authToken }}
+    - name: Put all tools to store to be cached afterwards
+      run: |
+        echo "::group::Downloading tools"
+        nix-store -v --realize $( nix-instantiate default.nix )
+        echo "::endgroup::"
+      shell: bash
```
laurenceisla, line 35: This is executed for every job that uses the setup-nix action. Since it's only needed to generate the first cache, maybe it should be moved to the Prepopulate job instead?

develop7: Good point! Will do

laurenceisla (approved):
Looks great! Took 8 mins to complete all the tests, and that's mostly due to the macOS build. GJ!

laurenceisla:
It's faster now! But I noticed these logs are starting to show (they're also present in this PR's Actions), e.g.:

```log
/usr/bin/tar: ../../../../../nix/store/jad0xjhpmbb7hhsi3r0sp9llbhcy1qqb-aws-c-mqtt-0.8.14/include/aws/mqtt/v5/mqtt5_listener.h: Cannot open: File exists
/usr/bin/tar: ../../../../../nix/store/jad0xjhpmbb7hhsi3r0sp9llbhcy1qqb-aws-c-mqtt-0.8.14/include/aws/mqtt/v5/mqtt5_listener.h: Cannot open: File exists
/usr/bin/tar: ../../../../../nix/store/jad0xjhpmbb7hhsi3r0sp9llbhcy1qqb-aws-c-mqtt-0.8.14/include/aws/mqtt/v5/mqtt5_client.h: Cannot open: File exists
/usr/bin/tar: ../../../../../nix/store/jad0xjhpmbb7hhsi3r0sp9llbhcy1qqb-aws-c-mqtt-0.8.14/include/aws/mqtt/v5/mqtt5_types.h: Cannot open: File exists
...
```

https://github.com/PostgREST/postgrest/actions/runs/7105291857/job/19342398874#step:3:68


That didn't happen before:

https://github.com/PostgREST/postgrest/actions/runs/7092294352/job/19303274920#step:3:68

I believe the Jobs could be even faster if that didn't happen. Not sure where the error is though, the only difference I see for that step is that `restore-key-hit` is `false` now.

develop7:
> But I noticed these logs are starting to show (they're also present in this PR's Actions), e.g.:

[Noticed those too](https://github.com/nix-community/cache-nix-action/issues/16,) it's appears to be fine: https://github.com/nix-community/cache-nix-action/#limitations

laurenceisla:
Ah, it's a known issue then. OK, let's leave it at that.

PR #3073, opened 2023-11-28
chore: fix server-timing metric order and doctest

laurenceisla:
Sends the `response` metric to the end of the `Server-Timing` header and makes the doctest work.

### DIFF
--- a/test/doc/Main.hs
+++ b/test/doc/Main.hs
@@ -16,6 +16,7 @@ main =
     , "src/PostgREST/Query/SqlFragment.hs"
     , "src/PostgREST/ApiRequest/Preferences.hs"
     , "src/PostgREST/ApiRequest/QueryParams.hs"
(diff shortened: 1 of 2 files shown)


steve-chavez (approved):

test/doc/Main.hs
```diff
@@ -16,6 +16,7 @@ main =
     , "src/PostgREST/Query/SqlFragment.hs"
     , "src/PostgREST/ApiRequest/Preferences.hs"
     , "src/PostgREST/ApiRequest/QueryParams.hs"
+    , "src/PostgREST/Response/Performance.hs"
```
steve-chavez, line 19: Oh, so doctests weren't running for this module?

laurenceisla: Yeah, they weren't.

PR #3075, opened 2023-11-28
refactor: use a data type instead of Map for Server Timing

laurenceisla:
Refactor the `Performance.hs` module.


steve-chavez (approved):
Nice! Less code and the order of the metrics is now ensured. :100:

PR #3083, opened 2023-12-03
feat: add max-affected preference to prefer header

### LINKED ISSUE #2887: `Prefer: handling=strict; affected` for conditional requests based on number of affected resources
## Problem

Currently checking the number of affected resources for a request is tied to the `application/vnd.pgrst.object+json` media type and it only checks for a  value of 1. 

## Solution

Now that we have `Prefer: handling=strict`, we can err on unmet conditions. This gives us an equivalent of the `Expect` header and can be used independently of the media type and request method.

We can have an `affected` preference, which would be a count of the operation affected resources. Its syntax would be:

```
Prefer: affected=<lt/lte/eq/gte/gt>.<number>
```

Assuming `items` has 1000 items:

```python
DELETE /items
Prefer: handling=strict; affected=gt.3,lte.20

400 Bad Request
{.. new PGRST code with message..}
```

Correcting the request with a filter:

```python
DELETE /items?id=in.(11,23,49,80)
Prefer: handling=strict; affected=gt.3,lte.20

200 OK
```

If `handling=lenient` is used or if `handling=strict` is not specified, `affected=..` won't have any effect.

## Notes

- It can be implemented with a `HAVING clause` as mentioned [here](https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1039489939).
- Previous issue trying to achieve the same with a `Prefer` header: https://github.com/PostgREST/postgrest/issues/2156
- Previous discussion (long thread) about achieving the same with a new URL query grammar: https://github.com/PostgREST/postgrest/pull/2164
- We can drop [limited-udpate-delete](https://postgrest.org/en/stable/references/api/tables_views.html#limited-update-delete) after this is supported. This feature would supersede it.

taimoorzaeem:
This should fix #2887.

### DIFF
--- a/test/spec/Feature/Query/PreferencesSpec.hs
+++ b/test/spec/Feature/Query/PreferencesSpec.hs
@@ -119,3 +119,75 @@ spec =
           { matchStatus = 200
           , matchHeaders = [matchContentTypeJson
                            , "Preference-Applied" <:> "handling=lenient"]}
+
+    context "test Prefer: max-affected with handling=strict" $ do
+      it "should fail if items deleted more than 10" $
+        request methodDelete "/items?id=lt.15"
+          [("Prefer", "handling=strict, max-affected=10")]
+          ""
+          `shouldRespondWith`
+          [json|{"code":"PGRST124","details":"The query affects 14 rows","hint":null,"message":"Query result exceeds max-affected preference constraint"}|]
+          { matchStatus = 400 }
+
+      it "should succeed if items deleted less than 10" $
+        request methodDelete "/items?id=lt.10"
+          [("Prefer", "handling=strict, max-affected=10")]
+          ""
+          `shouldRespondWith`
+          ""
+          { matchStatus = 204
+          , matchHeaders = ["Preference-Applied" <:> "handling=strict, max-affected=10"]}
+
+      it "should fail if items updated more than 0" $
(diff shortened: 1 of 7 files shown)


taimoorzaeem:
@steve-chavez I am not sure if we need to test this with batch upserts because that would involve the `POST` method and we only have this feature for updates and deletes.

steve-chavez:
@taimoorzaeem Yeah, I think it's fine to not do POST/Upsert

This feature can be like the strict counterpart of [Limited Update/Delete](https://postgrest.org/en/stable/references/api/tables_views.html#limited-update-delete).

test/spec/Feature/Query/PreferencesSpec.hs
```diff
@@ -119,3 +119,75 @@ spec =
...
+        request methodDelete "/tiobe_pls?name=eq.Java"
+          [("Prefer", "handling=strict, max-affected=1")]
+          [json| [{"name":"Java", "rank":19}] |]
+          `shouldRespondWith`
+          ""
+          { matchStatus = 204
+          , matchHeaders = ["Preference-Applied" <:> "handling=strict, max-affected=1"]}
+
+    context "test Prefer: max-affected with handling=lenient" $ do
+      it "should fail if items deleted more than 10" $
```
steve-chavez, line 159: it should be succeed here right?

PR #3102, opened 2023-12-12
fix: add `jwt_cache_max_lifetime` as an in-database configuration option

### LINKED ISSUE #3100: No in-db config for `jwt-cache-max-lifetime`
## Problem

There's no in-db config for https://postgrest.org/en/stable/references/configuration.html#jwt-cache-max-lifetime

## Solution

Add it to:

https://github.com/PostgREST/postgrest/blob/0c65bf494d75fbb2b8650eba1385b8d2c5d042d2/src/PostgREST/Config/Database.hs#L44-L69

laurenceisla:
Closes #3100


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+
+ - #3100, Add `jwt_cache_max_lifetime` to the in-database configuration options - @laurenceisla
+
 ### Fixed
 
  - #3054, Fix not allowing special characters in JSON keys - @laurenceisla
```
steve-chavez, line 14: Sorry, I think this was a fix rather than enhancement. Can you move to Fixed?

test/io/db_config.sql
```diff
@@ -6,6 +6,7 @@ ALTER ROLE db_config_authenticator SET pgrst.openapi_server_proxy_uri = 'https:/
 ALTER ROLE db_config_authenticator SET pgrst.jwt_secret = 'REALLY=REALLY=REALLY=REALLY=VERY=SAFE';
 ALTER ROLE db_config_authenticator SET pgrst.jwt_secret_is_base64 = 'false';
 ALTER ROLE db_config_authenticator SET pgrst.jwt_role_claim_key = '."a"."role"';
+ALTER ROLE db_config_authenticator SET pgrst.jwt_cache_max_lifetime = '3600';
```
steve-chavez, line 9: How about a different value to ensure the in-db config works?

laurenceisla: Ah, but the default value is `0`. That's why I set `3600` and `7200` for different authenticators.

steve-chavez: No, I mean these values are the same as the `no-defaults-*.config` . Try removing this line, the io tests will still pass right?

laurenceisla: Hmm... It actually fails, it gives me this error:

```log
E           jwt-secret-is-base64 = false
E         - jwt-cache-max-lifetime = 3600
E         ?                          ^
E         + jwt-cache-max-lifetime = 86400
E         ?                          ^ +
E           log-level = "info"
```

Because the `config/no-defaults.config` has `86400` and the `config/expected/no-defaults-with-db.config` has `3600`. The same happens with the other authenticator.

steve-chavez: My bad, I didn't see that the `expected` configs changed. It's all good.

PR #3105, opened 2023-12-13
nix: flakify the project

develop7:
* nix flake init 'srid/haskell-flake'
* add dependencies

`nix build` fails (???)


mkleczek:
I can submit a PR with https://github.com/mkleczek/postgrest/blob/temp-opentelemetry/flake.nix

BUT
I did not spend time on making sure all build types work correctly - especially static binary build.

Issue #3126, opened 2023-12-18
Error on Handler Function


steve-chavez:
## Problem

On https://postgrest.org/en/latest/references/api/media_type_handlers.html#handler-function

```sql
create or replace function get_line (id int)
returns "application/vnd.twkb" as $$
  select extensions.st_astwkb(geom) from lines where id = get_line.id;
$$ language sql;
```

When obtaining a null row:

```bash
curl "localhost:3000/rpc/get_line?id=3" -H "Accept: application/vnd.twkb" 
{"code":"PGRSTX00","details":null,"hint":null,"message":"RowError 0 2 UnexpectedNull"}
```

non-null rows work normally:

```
curl "localhost:3000/rpc/get_line?id=1" -H "Accept: application/vnd.twkb" 
HTTP/1.1 200 OK
```

PR #3129, opened 2023-12-19
fix: empty row on handler function 

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/3126

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
   - #3124, Fix table's media type handlers not working for all schemas - @steve-chavez
+  - #3126, Fix empty row on media type handler function - @steve-chavez
 
 ## [12.0.1] - 2023-12-12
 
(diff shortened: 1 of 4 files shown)


laurenceisla (approved):

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -223,7 +223,7 @@ asGeoJsonF = "json_build_object('type', 'FeatureCollection', 'features', coalesc
 
 customFuncF :: Maybe Routine -> QualifiedIdentifier -> QualifiedIdentifier -> SQL.Snippet
 customFuncF rout funcQi target
-  | (funcReturnsScalar <$> rout) == Just True = fromQi funcQi <> "(_postgrest_t.pgrst_scalar)"
+  | (funcReturnsScalar <$> rout) == Just True = "coalesce(" <> fromQi funcQi <> "(_postgrest_t.pgrst_scalar), '')"
```
steve-chavez, line 226: The `coalesce` was a bad approach because all the arguments have to be of the same type. This breaks when doing `coalesce(<json>, '')`  because `''` is not valid json for example.

PR #3137, opened 2023-12-28
feat: add in-database configuration for pool settings

### LINKED ISSUE #3110: Enable in-database configuration for pool settings
## Problem

Currently the following settings have no in-db option:

- https://postgrest.org/en/stable/references/configuration.html#db-pool
- https://postgrest.org/en/stable/references/configuration.html#db-pool-acquisition-timeout
- https://postgrest.org/en/stable/references/configuration.html#db-pool-max-idletime
- https://postgrest.org/en/stable/references/configuration.html#db-pool-max-lifetime

So whenever they need to be updated the `postgrest.service` (if using systemd) has to be restarted. During this lapse Nginx (or other proxy) will give uninformative error messages like `An invalid response was received from the upstream server`.

## Solution

Enable in-db options for the above configs. While there will be some downtime during the pool restart, we can give a better error message and include a `Retry-After` header.

laurenceisla:
Closes #3110.


laurenceisla:
@steve-chavez 

Managed to make it reload the pool configurations, but I'm not very familiarized with the process: I'm not sure if it needs to destroy the old pool or where. For example, I have doubts on what would happen here:

https://github.com/PostgREST/postgrest/blob/b2095a8ef601f4540f46e148ac038014720ea462/src/PostgREST/CLI.hs#L40-L42

I'm also not familiarized with the `debWorker` and if it needs to be updated too.

On a side note, I added a test that checks if the `db-pool` configuration is reloaded correctly. It fails in the main branch.... although now, the `test_no_pool_connection_available`  is failing...

PR #3142, opened 2024-01-06
feat: apply all function settings as transaction-scoped settings

### LINKED ISSUE #3061: Apply all function settings as transaction-scoped settings
## Problem

Currently, only `statement_timeout` is applied https://postgrest.org/en/latest/references/transactions.html#function-settings.

Applying other settings like `plan_filter.statement_cost_limit` is not possible.

```sql
CREATE OR REPLACE FUNCTION myfunc()
RETURNS void as $$
  SELECT pg_sleep(3); -- simulating some long-running process
$$
LANGUAGE SQL
SET plan_filter.statement_cost_limit TO 1300;
```  

## Solution

Apply all function settings.

## Notes

Not necessary to test with `plan_filter.statement_cost_limit`, we can use some others like:
https://github.com/PostgREST/postgrest/blob/91d53852370f5bf5ef1ecaa69b9ee7b8435b63a1/test/io/fixtures.sql#L19-L20

Also consider that applying some settings require privileges, see https://github.com/PostgREST/postgrest/pull/3058

taimoorzaeem:
Fixes #3061.

### DIFF
--- a/src/PostgREST/SchemaCache.hs
+++ b/src/PostgREST/SchemaCache.hs
@@ -74,7 +74,6 @@ import qualified PostgREST.MediaType as MediaType
 
 import Protolude
 
-
 data SchemaCache = SchemaCache
   { dbTables          :: TablesMap
   , dbRelationships   :: RelationshipsMap
@@ -297,7 +296,7 @@ decodeFuncs =
               <*> (parseVolatility <$> column HD.char)
               <*> column HD.bool
               <*> nullableColumn (toIsolationLevel <$> HD.text)
-              <*> nullableColumn HD.text
+              <*> compositeArrayColumn ((,) <$> compositeField HD.text <*> compositeField HD.text) -- function setting
 
     addKey :: Routine -> (QualifiedIdentifier, Routine)
     addKey pd = (QualifiedIdentifier (pdSchema pd) (pdName pd), pd)
@@ -432,7 +431,7 @@ funcsSqlQuery pgVer = [q|
     p.provolatile,
     p.provariadic > 0 as hasvariadic,
     lower((regexp_split_to_array((regexp_split_to_array(iso_config, '='))[2], ','))[1]) AS transaction_isolation_level,
-    lower((regexp_split_to_array((regexp_split_to_array(timeout_config, '='))[2], ','))[1]) AS statement_timeout
+    coalesce(func_settings.kvs, '{}') as kvs
   FROM pg_proc p
   LEFT JOIN arguments a ON a.oid = p.oid
   JOIN pg_namespace pn ON pn.oid = p.pronamespace
@@ -442,7 +441,15 @@ funcsSqlQuery pgVer = [q|
   LEFT JOIN pg_class comp ON comp.oid = t.typrelid
   LEFT JOIN pg_description as d ON d.objoid = p.oid
   LEFT JOIN LATERAL unnest(proconfig) iso_config ON iso_config like 'default_transaction_isolation%'
-  LEFT JOIN LATERAL unnest(proconfig) timeout_config ON timeout_config like 'statement_timeout%'
+  LEFT JOIN LATERAL (
+    SELECT
+      array_agg(row(
+        substr(setting, 1, strpos(setting, '=') - 1),
+        lower(substr(setting, strpos(setting, '=') + 1))
+      )) as kvs
+    FROM unnest(proconfig) setting
+    WHERE setting not LIKE 'default_transaction_isolation%'
+  ) func_settings ON TRUE
   WHERE t.oid <> 'trigger'::regtype AND COALESCE(a.callable, true)
 |] <> (if pgVer >= pgVersion110 then "AND prokind = 'f'" else "AND NOT (proisagg OR proiswindow)")
 
--- a/src/PostgREST/SchemaCache/Routine.hs
+++ b/src/PostgREST/SchemaCache/Routine.hs
@@ -6,6 +6,7 @@ module PostgREST.SchemaCache.Routine
   , Routine(..)
   , RoutineParam(..)
   , FuncVolatility(..)
+  , FuncSettings
   , RoutineMap
   , RetType(..)
   , funcReturnsScalar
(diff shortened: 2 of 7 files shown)


taimoorzaeem:
```haskell
queryFuncSettings :: PgVersion -> Bool -> Session FunctionSettings
queryFuncSettings pgVer prepared =
  let transaction = if prepared then SQL.transaction else SQL.unpreparedTransaction in
  transaction SQL.ReadCommitted SQL.Read $ SQL.statement mempty $ SQL.Statement sql HE.noParams (processRows <$> rows) prepared
 where
    sql = [q|
      WITH
      func_setting AS (
        SELECT current_user AS rolname, unnest(p.proconfig) as setting
        FROM pg_proc p
      ),
      kv_settings AS (
        SELECT
          rolname,
          substr(setting, 1, strpos(setting, '=') - 1) AS key,
          lower(substr(setting, strpos(setting, '=') + 1)) AS value
        FROM func_setting
      ),
      SELECT
        kv.rolname,
        array_agg(row(kv.key, kv.value)) AS func_setting
        FROM kv_settings kv
      JOIN pg_settings ps on ps.name = kv.key |] <>
      (if pgVer >= pgVersion150
        then "and (ps.context = 'user' or has_parameter_privilege(current_user::regrole::oid, ps.name, 'set')) "
        else "and ps.context = 'user' ") <> [q|
      GROUP BY kv.rolname;
```

@steve-chavez With the function settings defined as:
```haskell
type FunctionSettings = (HM.HashMap ByteString (HM.HashMap ByteString ByteString))
```

The first `ByteString` here is `role name`. How are we going to obtain this? Unlike Impersonated Role Settings which have a rolename in `pg_roles`, we don't have a role name in `pg_proc`. Is the query alright?

steve-chavez:
> The first ByteString here is role name. How are we going to obtain this? Unlike Impersonated Role Settings which have a rolename in pg_roles, we don't have a role name in pg_proc. Is the query alright?

@taimoorzaeem We don't need a role here. Since the settings are defined in the function, we only need the settings from `proconfig`.

src/PostgREST/Config/Database.hs
```diff
@@ -181,6 +184,43 @@ queryRoleSettings pgVer prepared =
     rows :: HD.Result [(Text, Maybe Text, [(Text, Text)])]
     rows = HD.rowList $ (,,) <$> column HD.text <*> nullableColumn HD.text <*> compositeArrayColumn ((,) <$> compositeField HD.text <*> compositeField HD.text)
 
+
+queryFuncSettings :: PgVersion -> Bool -> Session FunctionSettings
+queryFuncSettings pgVer prepared =
```
steve-chavez, line 189: Perhaps the query can be inside the SchemaCache? Since this is not meant to be exposed to external users, we could remove the `funcSettings` attribute here:

https://github.com/PostgREST/postgrest/blob/6c3d7a946daef91014328bc50c0f63744fada8ad/src/PostgREST/SchemaCache.hs#L187-L189

steve-chavez: Then `Routine` can have `funcSettings` (or routSettings) as an additional attribute.

taimoorzaeem:
@steve-chavez The type of FuncSettings is:
```haskell
type FuncSettings = [(ByteString,ByteString,ByteString)]
```
I think a hashmap would be better performance-wise but I'm not sure.

steve-chavez:
@taimoorzaeem How about having the function settings as part of:

https://github.com/PostgREST/postgrest/blob/d2fb67f596a8ded54911662aab445b4d7aaae8bd/src/PostgREST/SchemaCache/Routine.hs#L52-L62

> I think a hashmap would be better performance-wise but I'm not sure.

Then you wouldn't need to worry about search because the Function is already searched on `Plan.callReadPlan`, you would just need to get the `funcSettings` attribute.

src/PostgREST/SchemaCache.hs
```diff
@@ -1203,6 +1206,34 @@ timezones = SQL.Statement sql HE.noParams decodeTimezones
...
+          lower(substr(setting, strpos(setting, '=') + 1)) as value
+        FROM func_setting
+      )
+      SELECT
+        proname, kv.key AS key, kv.value AS value
+        FROM kv_settings kv
+      JOIN pg_settings ps ON ps.name = kv.key |] <>
+      (if pgVer >= pgVersion150
+        then "and (ps.context = 'user' or has_parameter_privilege(current_user::regrole::oid, ps.name, 'set'));"
+        else "and ps.context = 'user';")
```
steve-chavez, line 1231: For putting FuncSettings inside `Routine` you would need to move this query to `funcsSqlQuery`.

src/PostgREST/SchemaCache.hs
```diff
@@ -432,7 +439,10 @@ funcsSqlQuery pgVer = [q|
     p.provolatile,
     p.provariadic > 0 as hasvariadic,
     lower((regexp_split_to_array((regexp_split_to_array(iso_config, '='))[2], ','))[1]) AS transaction_isolation_level,
-    lower((regexp_split_to_array((regexp_split_to_array(timeout_config, '='))[2], ','))[1]) AS statement_timeout
+    -- the proconfig is returned as text[], which returns something like
+    -- e.g "\NUL\NUL\DB4statement_timeout=1s", this is solved by using
+    -- selecting first element
+    (p.proconfig)[1] AS func_setting
```
steve-chavez, line 445: This won't work if the function has two settings like:

```sql
create or replace function four_sec_timeout() returns void as $$
  select pg_sleep(3);
$$ language sql
set statement_timeout = '4s'
set log_min_duration_sample = '5s';
```

steve-chavez: Just tested this and the query on this part should be like:

```
    func_settings.kvs
```

And below..

src/PostgREST/SchemaCache.hs
```diff
@@ -442,7 +452,6 @@ funcsSqlQuery pgVer = [q|
   LEFT JOIN pg_class comp ON comp.oid = t.typrelid
   LEFT JOIN pg_description as d ON d.objoid = p.oid
   LEFT JOIN LATERAL unnest(proconfig) iso_config ON iso_config like 'default_transaction_isolation%'
-  LEFT JOIN LATERAL unnest(proconfig) timeout_config ON timeout_config like 'statement_timeout%'
```
steve-chavez, line 445: Use this query here:

```sql
  LEFT JOIN LATERAL (
    SELECT
      coalesce(array_agg(row(
        substr(setting, 1, strpos(setting, '=') - 1),
        lower(substr(setting, strpos(setting, '=') + 1))
      )), '{}') as kvs
    FROM unnest(proconfig) setting
    WHERE setting not like 'default_transaction_isolation%'
  ) func_settings ON TRUE
```

The decoding would then need to use `compositeArrayColumn`, like done here:

https://github.com/PostgREST/postgrest/blob/49584728b6423131712518e2f0803347c80069a1/src/PostgREST/Config/Database.hs#L181-L182

src/PostgREST/SchemaCache/Routine.hs
```diff
@@ -58,12 +61,12 @@ data Routine = Function
   , pdVolatility  :: FuncVolatility
   , pdHasVariadic :: Bool
   , pdIsoLvl      :: Maybe SQL.IsolationLevel
-  , pdTimeout     :: Maybe Text
+  , pdFuncSetting :: Maybe FuncSetting
```
steve-chavez, line 64: With the above queries the type should be like:

```haskell
type FuncSetting = [(Text,Text)]
```

src/PostgREST/SchemaCache.hs
```diff
@@ -317,6 +318,12 @@ decodeFuncs =
                       | v == 's' = Stable
                       | otherwise = Volatile -- only 'v' can happen here
 
+    parseFuncSetting :: Text -> FuncSetting
+    parseFuncSetting txt = toTuple $ T.splitOn "=" txt
+      where
+        toTuple [x,y] = (x,y)
+        toTuple _     = ("error","parsing")
+
```
steve-chavez, line 326: The decoding would then need to use `compositeArrayColumn`, like done here:

https://github.com/PostgREST/postgrest/blob/49584728b6423131712518e2f0803347c80069a1/src/PostgREST/Config/Database.hs#L181-L182

This won't need text splitting on the Haskell code.

src/PostgREST/SchemaCache/Routine.hs
```diff
@@ -49,6 +50,8 @@ data FuncVolatility
   | Immutable
   deriving (Eq, Show, Ord, Generic, JSON.ToJSON)
 
+type FuncSetting = (Text,Text)
```
steve-chavez, line 53: With the above changes then this type should be:
```suggestion
type FuncSetting = [(Text,Text)]
```

src/PostgREST/SchemaCache/Routine.hs
steve-chavez, line 64:
```suggestion
  , pdFuncSettings :: FuncSettings
```

test/io/fixtures.sql
```diff
@@ -198,3 +198,9 @@ $$ language sql set statement_timeout = '4s';
 create function get_postgres_version() returns int as $$
   select current_setting('server_version_num')::int;
 $$ language sql;
+
+GRANT SET ON PARAMETER log_min_duration_sample TO postgrest_test_anonymous;
+
+create or replace function log_min_duration_test() returns text as $$
+  select current_setting('log_min_duration_sample',false);
+$$ language sql set log_min_duration_sample = '5s';
```
steve-chavez, line 206: Since this causes an error on older pg versions, maybe you can use [work_mem](https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-WORK-MEM) instead.

IIRC it won't require a GRANT.

```suggestion
create or replace function work_mem_test() returns text as $$
  select current_setting('work_mem',false);
$$ language sql set work_mem = '6000';
```

steve-chavez:
Hm, the [Build Linux static (Nix)](https://github.com/PostgREST/postgrest/actions/runs/7846927486/job/21414731007#logs) CI job is failing with `error: writing to file: No space left on device`. I thought this would clear on main but didn't.

PR #3169, opened 2024-01-18
Use pkgsStatic instead of static-haskell-nix, bump to libpq 16 and support GSSAPI

### LINKED ISSUE #2478: Move away from static-haskell-nix

### LINKED ISSUE #2815: GSSAPI authorization to the database
### Environment

* PostgreSQL version: 9.6
* PostgREST version:  9.0.0
* Operating system:  debian

### Description of issue
Is it possible to use postgrest with the GSSAPI authorization method https://postgrespro.com/docs/postgresql/14/gssapi-auth

Steps to reproduce: Include a minimal SQL definition plus how you make the request to PostgREST and the response body:
1. sudo nano ph_hba.conf
2. host        all         all                 0.0.0.0/0       gss include_realm=0
3. sudo systemctl restart postgresql
4. su domainuser
5. psql -h example.com -d example_db 
psql (9.6.10)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bit: 256, compression: off)
Enter "help" to get help.
example_db=>
6. there is a connection! )

7. nano postgrest.conf
db-uri="postgres://example.com:5432/example_db"
db-schema="public"
db-anon-role="domainuser"
server-port="3006"
db-channel-enabled="false"
db-prepared-statements="false"
8. ./postgres postgrest.conf 
9. An error ocurred when trying to query database settings for the config parameters
ConnectionError (Just "connection to server at "example.com" (xx.xx.xx.xx), port 5432 failed: GSSAPI authentication not supported")
10. there is no connection (

wolfgangwalther:
The first commit replaces static-haskell-nix with a build of the static executable via the `pkgsStatic` overlay of nixpkgs. This supersedes #2332 to resolve #2478.

Based on this, I also tried to cross-compile the static executable to aarch64, but this quickly fails because of https://github.com/NixOS/nixpkgs/issues/36200. Cross-compiling Template Haskell is not easy. Will investigate further.

In the meantime, I also did the following:
- Update to libpq v16 in the second commit.
- Build libpq with GSSAPI support in the third commit, which resolves #2815. Had to patch PG a little bit for it to correctly link to the static krb5 libraries, but it builds fine now.

There is room for improvement with the libpq package, which is basically just a stripped down postgresql package right now. It still builds the server, but without systemd, which makes it pull in a lot fewer dependencies - and makes it succeed on cross-platform builds, too. We could still do more here to really only build libpq and nothing else - which would surely be much quicker. Also we currently build it twice (once for the dynamic overlay and once for the static overlay), which can possibly be reduced to one. Furthermore, I suspect that the current way of overriding the "default postgresql version" package, effectively leads to having to rebuild some other dependencies that we use for our devtools, because they depend on that postgresql package for their version of libpq. TLDR: If we create a separate, small derivation just for libpq and pass that only to postgresql-lib (both for dynamic and static), that should be **much** more efficient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
  - #3149, Misleading "Starting PostgREST.." logs on schema cache reloading - @steve-chavez
+ - #2815, Build static executable with GSSAPI support - @wolfgangwalther
 
 ### Deprecated
 
--- a/nix/libpq.nix
+++ b/nix/libpq.nix
@@ -25,12 +25,13 @@ stdenv.mkDerivation {
     "--without-gssapi"
     "--without-icu"
     "--without-readline"
+    "--with-gssapi"
     "--with-openssl"
     "--with-system-tzdata=${tzdata}/share/zoneinfo"
   ];
 
   nativeBuildInputs = [ pkg-config tzdata ];
-  buildInputs = [ openssl zlib ];
+  buildInputs = [ libkrb5 openssl zlib ];
 
   buildFlags = [ "submake-libpq" "submake-libpgport" ];
 
@@ -46,6 +47,12 @@ stdenv.mkDerivation {
     rm -rfv $out/share
 
     runHook postInstall
+  '';
+
+  # To avoid linking errors in the static build with gssapi
+  postInstall = ''
+    substituteInPlace $out/lib/pkgconfig/libpq.pc\
+      --replace "Requires.private:" "Requires.private: krb5-gssapi,"
   '';
 
   outputs = [ "out" ];
--- a/nix/static.nix
+++ b/nix/static.nix
@@ -36,8 +36,11 @@ let
           # static libs for libpq.
           nativeBuildInputs = prevAttrs.nativeBuildInputs ++ [ pkgs.pkgsStatic.pkg-config ];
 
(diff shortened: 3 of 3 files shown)


develop7:
@wolfgangwalther didn't help. I've filed nix-community/cache-nix-action#27 and about to revert `cache-nix-action` back to 4.x. Honestly, I didn't quite expect them to break the cache restore part :)

steve-chavez:
Perhaps https://github.com/nlewo/nix2container could help? I've heard it's lighter on the nix store. References:

- https://blog.eigenvalue.net/2023-nix2container-everything-once/
- https://lewo.abesis.fr/posts/nix-build-container-image/
- also see https://github.com/pdtpartners/nix-snapshotter

PR #3181, opened 2024-01-22
Sync with Postgrest/main

Gab0:


steve-chavez:
@Gab0 Interesting work on SAML authentication. Are you planning to upstream it?

PR #3187, opened 2024-01-25
ci: fix previous caches not being reused

develop7:
With this update, any previously existing nix cache would be restored too, avoiding a full rebuild caused by updating files in `nix/` or `setup-nix` action itself.


develop7:
@wolfgangwalther yeah, still relevant. Couldn't wait for the #3190 to be merged.

develop7:
Indeed, I've lost extra keys for nix during rebase. Thank you for the heads-up. Let me fix it in a sec.

develop7:
Exactly, (while I was toying multi-layer cache idea in my head) only one cache is restored, not according to the docs, but also to the logs — https://github.com/PostgREST/postgrest/actions/runs/7797378199/job/21263851394#step:5:83

My guess is Github cut their disc space allowance for action runners

PR #3191, opened 2024-01-26
Check for static executable in nix derivation

wolfgangwalther:
As mentioned in https://github.com/PostgREST/postgrest/pull/3169#issuecomment-1911177795 and more: `postgrest-check-static` is moved into the nix build itself, providing a guarantee that we only build static binaries that way. Ever.


steve-chavez (approved):
:rocket:

PR #3194, opened 2024-01-28
Prevent nix/store references in static builds

wolfgangwalther:
This adds a check to confirm that the static executable we build via nix does not have any references to /nix/store anymore. This currently fails with paths to the following derivations, as noted in https://github.com/PostgREST/postgrest/pull/3169#issuecomment-1913362540:

```
# haskell
/nix/store/j9pp5jp32z3mp84p2fsnwdg4si7qgm0z-postgrest-static-x86_64-unknown-linux-musl-11.2.0
/nix/store/a35zz3304b1fzv4613mfr5lwgz1n0wnj-warp-static-x86_64-unknown-linux-musl-3.3.25
/nix/store/icwvcmmdblflirrnbkhs73hlr2z1ix0b-HTTP-static-x86_64-unknown-linux-musl-4000.4.1

# non-haskell
/nix/store/mkig2hb4ksyk0j006rwyfq8b84nhh2pb-libpq-x86_64-unknown-linux-musl-16.1/etc/postgresql
/nix/store/bf84fd2c0kdr70chri3d21mn8za1ab8n-libkrb5-static-x86_64-unknown-linux-musl-1.20.2
/nix/store/qk1hy5cv7fjf44g3xyw8d94imvhl04xf-openssl-static-x86_64-unknown-linux-musl-3.0.12-etc
/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-openssl-static-x86_64-unknown-linux-musl-3.0.12
```

Most of those were introduced in 259d97acee94a5fadd895a1ef054e03025b953ed and c94aa9ccd9c56b7c8a96aaa6f9c15a9778528d2a - but the openssl-etc reference was there before, already. Before the switch to pkgsStatic, this didn't blow up closure size that much, so went unnoticed.

But even before, the references that OpenSSL was storing to the sysconfdir (/etc) would not have worked on a non-nix system. This should be possible to fix with a `--sysconfdir` configure flag. Not sure, yet, about the other openssl reference, which has already been replaced by nixpkgs' `remove-references-to`.

The libpq reference is just an oversight when I introduced the new `libpq` derivation which I fixed in the second commit. The libkrb5 references can probably be solved similarly - we just didn't build with GSSAPI support, so that's new.

This leaves us with three haskell dependencies - and at first glance, this seems quite random. The thing here is, that in the new executable, the auto-generated `Paths_xxx` modules for those packages are included. Those three are likely the only dependencies we use that have such a module and that's why they show up. IIRC those Paths modules are actually Template Haskell - and should not be referenced in the final bundle at all. The fact they show up means that GHC is linking way too much stuff into our executable now. The unpacked size of the build artifact is a whopping 51 MB for it now - compared to 17 MB before the switch.

How what happened here and why? I don't exactly know, yet, but looking at the last CI run before the switch at https://github.com/PostgREST/postgrest/actions/runs/7660083430 we have:
- A statically linked artifact for linux of around 5 MB.
- A dynamically linked artifact for macos of around 3-4 MB.

Those sizes are zipped, but correspond to the sizes mentioned above. Now, those two seem to match up nice: We expect the dynamic build to be smaller than the static build, right?

But note that we also have zipped sizes of:
- 13 MB for freebsd
- 17 MB for ubuntu-aarch64
- 13 MB for ubuntu-x64
- 14 MB for windows

Those are very similar to the **new** static build, which is around 12-13 MB zipped: https://github.com/PostgREST/postgrest/actions/runs/7670588137. In fact, those dynamically linked builds also contain all the Paths modules, just without a `/nix/store` path, because they were build outside. Example:

```
/home/runner/work/postgrest/postgrest/.stack-work/install/x86_64-linux-tinfo6/7fb34a25383af7e6721ce2222081f7f4af7b1ed5f3f45443771c11e72a32092d/9.4.5/etc
```

This means, that the new static build didn't "break" this in an isolated way, but rather:
- The macos and static-haskell-nix builds did "something right".
- All the other builds are missing something.

Interestingly, the dynamic package build via `nix-build -A postgrestPackage` (not exposed as artifcat anywhere) is also small.

I have not been able to find out exactly why this happens, but here are my observations and ideas:
- For any haskell dependency we always get a `.so` file, in the static overlay we also get a `.a` file.
- ~Because nix enables `--enable-split-sections`, we also get a bunch of `.hi`, `.p_hi` and `dyn_hi` files - one for each module a package exposes.~
- GHC actually links the haskell dependencies statically anyway - even in the very small MacOS case. The difference between the static and dynamic builds is only in the non-haskell libs and how they are linked.

My best guess is that somehow GHC looses the ability to ~use the `.hi` files directly, so that it~ only links in the code we need. Instead we seem to link in all of the `.a` files, so basically each package in full.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -28,6 +28,12 @@ let
       sha256 = nixpkgsVersion.tarballHash;
     };
 
+  nixpkgs-patched = (import nixpkgs { inherit overlays system; }).applyPatches {
+    name = "nixpkgs-patched";
+    src = nixpkgs;
+    patches = [ nix/split-sections-cross.patch ];
+  };
+
   allOverlays =
     import nix/overlays;
 
@@ -46,7 +52,7 @@ let
 
   # Evaluated expression of the Nixpkgs repository.
   pkgs =
-    import nixpkgs { inherit overlays system; };
+    import nixpkgs-patched { inherit overlays system; };
 
   postgresqlVersions =
     [
@@ -82,7 +88,7 @@ let
   inherit (pkgs.haskell) lib;
 in
 rec {
-  inherit nixpkgs pkgs;
+  inherit nixpkgs-patched pkgs;
 
   # Derivation for the PostgREST Haskell package, including the executable,
   # libraries and documentation. We disable running the test suite on Nix
--- a/nix/split-sections-cross.patch
+++ b/nix/split-sections-cross.patch
@@ -1,0 +1,14 @@
+This works around https://github.com/NixOS/nixpkgs/issues/286285 by passing --enable-split-sections instead of ghc-options.
+---
+--- a/pkgs/development/haskell-modules/generic-builder.nix
++++ b/pkgs/development/haskell-modules/generic-builder.nix
+@@ -245,8 +245,7 @@ let
+     (enableFeature doBenchmark "benchmarks")
+     "--enable-library-vanilla"  # TODO: Should this be configurable?
+     (enableFeature enableLibraryForGhci "library-for-ghci")
+-  ] ++ optionals (enableDeadCodeElimination && (lib.versionOlder "8.0.1" ghc.version)) [
+-     "--ghc-option=-split-sections"
++    (enableFeature enableDeadCodeElimination "split-sections")
+   ] ++ optionals dontStrip [
+     "--disable-library-stripping"
+     "--disable-executable-stripping"


steve-chavez:
> A better option might be to not build the dynamic ubuntu executable with stack anymore, but take one of the cabal builds that run anyway instead. 

Yeah, that LGTM. We don't publish that ubuntu executable anyway. IIRC it's just for verifying stack works.

PR #3197, opened 2024-01-29
ci: nix: more concurrent downloads

develop7:
Playing with nix.conf's [`http-connections`](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-http-connections) parameter, might improve the downloading time of many small files.


develop7:
> Any result, so far? I have no idea what to look at, tbh :)

That improved output discovery time across a handful of Nix substituters for me, but in my case, network latency also could be the cause.

I'd just close it for now

PR #3211, opened 2024-02-06
ci: Improve caching

wolfgangwalther:
Plenty of text in the commit messages. Hopefully this also fixes the currently failing loadtest, but I'm not sure about that.

---

TODO:
- [x] Save nix cache only on main
- [x] Split stack/cabal caches into dependency and working directory caches
- [x] Reduce cache size
- [ ] ~~Add PR caching for postgrest-build~~
- [ ] ~~Garbage collect nix/stack/cabal caches?~~
- [ ] ~~Purge old caches proactively when saving new cache?~~

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -119,6 +119,12 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
     steps:
       - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      - uses: haskell-actions/setup@64aa5146ed5bd91f181552b8b5fbe50fbd2bd3fe # v2.7.1
+        with:
+          ghc-version: 9.4.5
+          enable-stack: true
+          stack-no-global: true
+          stack-setup-ghc: true
       - name: Cache ~/.stack
         uses: ./.github/actions/cache-on-main
         with:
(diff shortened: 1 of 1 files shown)


develop7 (approved):
Everything seems to be taken care of, LGTM!

develop7:
> we can additionally cache just the dist-newstyle folder between the "prepopulate job" and all the test jobs

Ooh, this one's good. Should've thought of this myself. Looks like something `action/cache` can easily handle.

> We'd just need to make sure that those caches get some kind of priority over the main caches in terms of eviction.

The current eviction policy is that the oldest caches are deleted first, so all we could do is to create the big Nix cache last somehow. Let's see if we do hit that limit first.

> * Garbage collect nix/stack/cabal caches?

On the one hand, it makes sense for Nix at lease, on the other — there were full Nix cache rebuilds and from what I've noticed, the cache size doesn't seem to increase significantly, if ever; it was 3-something GB all the time.

> * Stack/cabal caches between 300MB and 1GB each

Speaking of Stack caches, somehow Linux cache is 380M, MacOS is 680M and Windows is 1GB, I wonder why?

develop7:
> Maybe we can do better with some combination of ghcup and `--system-ghc` or so.

We could even set stack up with `ghcup`, and [configure stack in a way it'll use `ghcup` to install GHC](https://www.haskell.org/ghcup/guide/#strategy-1-stack-hooks-new-recommended) (actually `ghcup` bootstrap script does it unless you ask it to), and if we cache `.ghcup` separately, we could _share cached GHC_ between `cabal` and `stack` builds. With GHC tarballs are 200 to 300M big, times 3 platforms, this would result in another ~1GB saved cache.

develop7:
> Since all caches are evicted after 7 days anyway, we could just run a daily workflow to restore (and save?) the current caches on main. This way they'd always be "used last" and hopefully not be evicted instead of PR caches.

Yeah, I think I've suggested it somewhere in comments to nix-cache-related PRs. The eviction policy doesn't mention access time, though, but that seems like the best effort.

develop7:
> Not really, they use different GHC versions

Why do we stick to 9.4.5 for Stack, BTW? 9.4.8 is in Stackage for a month or so

develop7:
Excellent point on caching GHC and dependencies in non-nix builds separately

develop7:
> do you see any way of "not storing the cache" with cache-nix-action?

@wolfgangwalther use [`restore` action](https://github.com/nix-community/cache-nix-action/tree/v4.0.3/restore#only-restore-cache) only? That's what the [`cache-nix-action` is using internally](https://github.com/nix-community/cache-nix-action/blob/v4.0.3/action.yml#L68) anyway.

develop7:
> I start to wonder why do we bother with caching this stuff via github actions at all?

Nix caching does shave off a minute or two per Nix-involving job, which shortens the feedback loop accordingly; usually to these two minutes, due to CI jobs running in parallel. For the macOS case, where there is a single cache consumer job and low-performing runner, it certainly does make sense to drop cache altogether.

This is my first approach to GitHub Actions and their cache in particular, and, being honest, it is performing worse than I've expected, especially Windows/Mac runners (download speed is barely 100Mb/sec (down to abysmal 20MB/sec even), while in Linux runners it's 380Mb/s sometimes). With the dedicated infrastructure, I expect the improvement to be more significant, but currently it might not be worth the effort.

Issue #3212, opened 2024-02-06
Bloated `pg_depend` can make the schema cache queries slow and fail due to `statement_timeout`


steve-chavez:
## Problem

Having a bloated `pg_depend`:

```sql
\dit+ pg_catalog.pg_depend*
                                                     List of relations
   Schema   |           Name            | Type  |  Owner   |   Table   | Persistence | Access method | Size  | Description 
------------+---------------------------+-------+----------+-----------+-------------+---------------+-------+-------------
 pg_catalog | pg_depend                 | table | postgres |           | permanent   | heap          | 62 MB | 
 pg_catalog | pg_depend_depender_index  | index | postgres | pg_depend | permanent   | btree         | 54 MB | 
 pg_catalog | pg_depend_reference_index | index | postgres | pg_depend | permanent   | btree         | 41 MB | 
(3 rows)

SELECT * FROM pg_stat_sys_tables WHERE relname = 'pg_depend';

-[ RECORD 1 ]-------+------------------------------
relid               | 2608
schemaname          | pg_catalog
relname             | pg_depend
seq_scan            | 12
seq_tup_read        | 39326
idx_scan            | 52127
idx_tup_fetch       | 146580
n_tup_ins           | 13594
n_tup_upd           | 0
n_tup_del           | 13588
n_tup_hot_upd       | 0
n_live_tup          | 3279
n_dead_tup          | 1072005  <- many dead tuples
n_mod_since_analyze | 0
n_ins_since_vacuum  | 0
last_vacuum         | 2024-02-06 22:49:26.772597+00
last_autovacuum     | 2024-02-06 22:50:50.93451+00
last_analyze        | 2024-02-06 22:49:26.808658+00
last_autoanalyze    | 2024-02-06 22:16:31.48843+00
vacuum_count        | 3
autovacuum_count    | 638
analyze_count       | 3
autoanalyze_count   | 25

```

Makes the [tablesSqlQuery](https://github.com/PostgREST/postgrest/blob/49584728b6423131712518e2f0803347c80069a1/src/PostgREST/SchemaCache.hs#L595-L817) slow:

<details>
<summary> Execution Time: 4657.589 ms</summary>

<pre>
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   QUERY PLAN                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
[...]

PR #3213, opened 2024-02-07
feat: log schema cache stats to stderr

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/3212.

Also related to https://github.com/PostgREST/postgrest/issues/2450.


steve-chavez:
Ah, my bad. I merged this without CI running the pg tests. I thought only the load tests were failing :man_facepalming:

PR #3223, opened 2024-02-11
Minimize nix dependencies for CI

wolfgangwalther:
As discussed in #3211, one way to make CI is faster is to just reduce the number of dependencies we pull in for all our tooling. The heaviest tool is `withTools` - it includes the full postgrest environment (GHC + all deps) and postgresql in many different versions, including all it's dependencies.

The closure of `withTools` is currently 7.9G on main. After all the changes in this PR, it sits at around 2.1G. This should make a big difference.

I did the following:
- Removed pg9.6 and pg10 as prepared for in #2052.
- Removed pg11 on top of that as proposed in https://github.com/PostgREST/postgrest/pull/2052#issuecomment-1937343692 - this removes the duplicated buildchain from the older nixpkgs version.
- Removed postgis, which gives us somewhere between 1G and 1.5G as well. Heavy dependencies here. Maybe we can replace our postgis support with the new domain-based mimetype handling anyway? Not sure whether that's discussed somewhere already.
- Reconfigured postgresql without many optional dependencies. No systemd, no icu, no gssapi, no debug symbols. None of this affects the libpq we are building to actually link against, though. This is just about the 5-8 copies of postgresql we have laying around for our tests.
- Reconfigured GHC to not create profiled libs or dynamic libs - it uses static libs only anyway. We'd need the profiled libs for the memory test - but this can be a separate build of GHC including *only* the profiled libs. Then the memory test would only have to download those and not everything as well.

This is still heavily WIP. I made no effort, so far, to make the static build work, nor did I actually make sure the tests pass etc. This first test is just about evaluating how much time we potentially saving for CI jobs.


wolfgangwalther:
I don't think the added complexity this would add would be worth the gains. It's better to rely on the upstream Nixpkgs GHC as much as possible.

PR #3224, opened 2024-02-12
fix: Avoid casting to table type when select= and media type handler are used

### LINKED ISSUE #3160: Returning mimetypes via aggregates only works for select=*
Using the query from https://github.com/PostgREST/postgrest/issues/2826#issuecomment-1889897084 as an example:

```sql
WITH pgrst_source AS ( SELECT "test"."lines".* FROM "test"."lines")
SELECT
    twkb_agg(_postgrest_t::"test"."lines") AS body,
    '' AS response_inserted
FROM ( SELECT * FROM pgrst_source ) _postgrest_t;
```

The cast to `::"test"."lines"` makes it impossible to use anything else than `select=*`.

I see why the cast is implemented: This allows the aggregate function to be matched on the argument type for different endpoints. But it does not allow selecting only a subset of columns or computed columns.

In my case, I just tried to implement a generic CSV handler like this:

```sql
CREATE DOMAIN "text/csv" AS text;

CREATE FUNCTION jsonb_state_transfn(state jsonb, elem anyelement) RETURNS jsonb
STABLE
PARALLEL SAFE
LANGUAGE plpgsql AS $$
BEGIN
  RETURN state || to_jsonb(elem);
END$$;

CREATE FUNCTION text_csv_agg_finalfn(state jsonb) RETURNS "text/csv"
TRANSFORM FOR TYPE jsonb
STABLE
PARALLEL SAFE
LANGUAGE plpython3u AS $$
  from pandas import DataFrame
  df = DataFrame(state)
  return df.to_csv()
$$;

CREATE AGGREGATE text_csv_agg(anyelement) (
  INITCOND  = '[]',
  STYPE     = jsonb,
  SFUNC     = jsonb_state_transfn,
  FINALFUNC = text_csv_agg_finalfn
);
```

But this quickly fails, because I want to select different columns, or just in a different order - and then the cast fails.

wolfgangwalther:
Previously using a generic mimetype handler failed when any kind of select= was given, because we tried to cast the select-result to the original table type. With this change, this cast is only applied when select=* is given implicitly or explicitly. This is the only case where this makes sense, because this guarantees that correct columns are selected in the correct order for this cast to succeed.

Resolves #3160

### DIFF
--- a/test/spec/Feature/Query/CustomMediaSpec.hs
+++ b/test/spec/Feature/Query/CustomMediaSpec.hs
@@ -230,6 +230,76 @@ spec = describe "custom media types" $ do
         simpleHeaders r `shouldContain` [("Content-Type", "text/csv; charset=utf-8")]
         simpleHeaders r `shouldContain` [("Content-Disposition", "attachment; filename=\"lines.csv\"")]
 
+  -- https://github.com/PostgREST/postgrest/issues/3160
+  context "using select query parameter" $ do
+    it "without select" $ do
+      request methodGet "/projects?id=in.(1,2)" (acceptHdrs "pg/outfunc") ""
+        `shouldRespondWith`
+        [str|(1,"Windows 7",1)
+            |(2,"Windows 10",1)
+            |]
+        { matchStatus  = 200
+        , matchHeaders = ["Content-Type" <:> "pg/outfunc"]
+        }
+
+    it "with fewer columns selected" $ do
+      request methodGet "/projects?id=in.(1,2)&select=id,name" (acceptHdrs "pg/outfunc") ""
+        `shouldRespondWith`
+        [str|(1,"Windows 7")
+            |(2,"Windows 10")
+            |]
+        { matchStatus  = 200
+        , matchHeaders = ["Content-Type" <:> "pg/outfunc"]
+        }
+
+    it "with columns in different order" $ do
+      request methodGet "/projects?id=in.(1,2)&select=name,id,client_id" (acceptHdrs "pg/outfunc") ""
+        `shouldRespondWith`
+        [str|("Windows 7",1,1)
+            |("Windows 10",2,1)
+            |]
+        { matchStatus  = 200
+        , matchHeaders = ["Content-Type" <:> "pg/outfunc"]
+        }
+
+    it "with computed columns" $ do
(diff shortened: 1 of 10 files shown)


test/spec/Feature/Query/CustomMediaSpec.hs
```diff
@@ -230,6 +230,59 @@ spec = describe "custom media types" $ do
...
+
+    it "with fewer columns selected" $ do
+      request methodGet "/projects?id=in.(1,2)&select=id,name" (acceptHdrs "pg/outfunc") ""
+        `shouldRespondWith`
+        [str|(1,"Windows 7")
+            |(2,"Windows 10")
+            |]
+        { matchStatus  = 200
+        , matchHeaders = ["Content-Type" <:> "pg/outfunc"]
+        }
```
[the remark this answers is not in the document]
steve-chavez: > IIUC, 415 is for content-type, while 406 is for Accept. So I guess we are returning the wrong error code here?

Right, my mistake there. Should be `406`.

test/spec/Feature/Query/CustomMediaSpec.hs
```diff
@@ -230,6 +230,76 @@ spec = describe "custom media types" $ do
...
+        `shouldRespondWith` 406
+
+    it "will fail for specific aggregate with columns in different order" $ do
+      request methodGet "/lines?select=name,id,geom" (acceptHdrs "application/vnd.twkb") ""
+        `shouldRespondWith` 406
+
+    -- This is just because it would be hard to detect this case, so we better error in this case, too.
+    it "will fail for specific aggregate with columns in same order" $ do
+      request methodGet "/lines?select=id,name,geom" (acceptHdrs "application/vnd.twkb") ""
+        `shouldRespondWith` 406
```
steve-chavez, line 301: Yes, these look good :+1:

steve-chavez: @wolfgangwalther LGTM. The only purpose was to allow having a generic and particular handler for a media type. If the function name has to change for that it's fine.

> Would you agree to just remove those tests (or change the expected outcome to the error message),

How about just renaming the functions and not removing the tests?

steve-chavez: > How about just renaming the functions and not removing the tests?

Just tried that and it does work.

Issue #3219, opened 2024-02-09
Server-Timing values are incorrect?


steve-chavez:
## Problem

[timeItT](https://hackage.haskell.org/package/timeit-2.0/docs/System-TimeIt.html#v:timeItT) is used for getting the values of the durations of the Server-Timing header, its docs say:

> Wrap a [MonadIO](https://hackage.haskell.org/package/base-4.9.1.0/docs/Control-Monad-IO-Class.html#t:MonadIO) computation so that it returns execution time **in seconds**, as well as the result value.

And on https://postgrest.org/en/stable/references/admin.html#server-timing-header, we say that the durations are in milliseconds

Yet the seconds are multiplied by 1000000 https://github.com/PostgREST/postgrest/blob/5424be76d9df64fd3cab846d9e9a6e621c540769/src/PostgREST/Response/Performance.hs#L29

cc @develop7

develop7:
> It should by `*1000` right?

That seems accurate. Thank you for spotting this. Will fix in a jiff.

develop7:
@steve-chavez #3227

PR #3227, opened 2024-02-14
fix: fixes server timings' precision

develop7:
Make `Server-Timing` durations match the documentation (see #3219) and document measurement units in source.


src/PostgREST/Response/Performance.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE NumericUnderscores #-}
```
steve-chavez, line 1: TIL. This is a nice extension.

Issue #2688, opened 2023-02-27
Selecting computed columns from set-returning RPCs broken


wolfgangwalther:
### Environment

* PostgreSQL version: 15
* PostgREST version: 10.1.2
* Operating system:

### Description of issue

I have an RPC `x` returning `SETOF some_exposed_view`. I have a computed column like `CREATE FUNCTION comp(some_exposed_view) ...`.

The following request works:
```http
GET /rpc/x?select=col,comp
```

But the following request does not work and returns a `column some_exposed_view.comp not found`:

```http
GET /rpc/x?select=*,comp
```

This is because of this:

https://github.com/PostgREST/postgrest/blob/0d5d209bfb9db3af02d153820d9b6a076ca0f105/src/PostgREST/Plan.hs#L478-L479

This basically throws away all computed columns, when `*` is used.

I was able to fix the query in my own project by adding explicit column names. But somehow, I was not able to reproduce this in the spec tests here, yet. The fix should be simple, by returning `fldNames` instead of  `["*"]`. But we'd need a working test-case first.

laurenceisla:
> But somehow, I was not able to reproduce this in the spec tests here, yet. The fix should be simple, by returning fldNames instead of ["*"]. But we'd need a working test-case first.

It seems that this happens only when the function is overloaded. In this PR, I added the failing test: 

https://github.com/PostgREST/postgrest/pull/3228/files#diff-9ef4627ce82e6bbe2a7e41c43686707bcf24dcdddfd94a79bf077e91d2a0397aR544-R546

Is this the same case that you described? That PR should solve this if that's the case.

PR #3228, opened 2024-02-14
fix: selecting overloaded computed columns alongside `*` when returing tables from INSERT and RPC

### LINKED ISSUE #3145: Computed fields not selected during INSERT
### Environment

* PostgreSQL version: 16.1
* PostgREST version: 12.0.2
* Operating system: linux amd64

### Description of issue

Given the schema:
```sql
-- table
create table if not exists app.office (
    id uuid not null primary key default uuid_generate_v7(), -- from pg_uuidv7, irrelevant
    name text not null
);
-- computed field
create or replace function app.created_at(app.office) returns timestamptz strict immutable return uuid_v7_to_timestamptz($1.id);
```

I fail to select computed fields just after successful insert:
```curl
POST /office?select=*,created_at
Prefer: missing=default, return=representation

{
  "name": "o1"
}

{
  "code": "42703",
  "details": null,
  "hint": null,
  "message": "column office.created_at does not exist"
}
```

while the following works just like expected:
```curl
###
POST /office?select=*
Prefer: missing=default, return=representation

{
  "name": "o1"
}
###
GET /office?name=eq.o1&select=*,created_at

[
  {
    "id": "018cecc4-0793-71fd-bf11-eb956fd48099",
    "name": "o1",
    "created_at": "2024-01-09T05:48:39.187+00:00"
  }
]
```

I tried to debug and obtained the stripped-down code:
```sql
with o as (
    insert into app.office (name) values ('o1') returning *
),
base as (
    select o.id, o.name, o.created_at from o -- NB: here `o` is somehow seen by the engine as _just a record_, not `app.office`
) select * from base;
```

Manual cast does the trick:
```sql
with o as (
    insert into app.office (name) values ('o1') returning *
),
base as (
    select o.id, o.name, (o::app.office).created_at from o
) select * from base;
```

TIA

laurenceisla:
Closes #3145
It should also work for #2688 (needs confirmation)

### DIFF
--- a/test/spec/Feature/Query/QuerySpec.hs
+++ b/test/spec/Feature/Query/QuerySpec.hs
@@ -533,9 +533,15 @@ spec actualPgVersion = do
         get "/items2?id=eq.1&select=id,computed_overload" `shouldRespondWith`
           [json|[{"id":1,"computed_overload":true}]|]
           { matchHeaders = [matchContentTypeJson] }
+        get "items?id=eq.1&select=*,computed_overload" `shouldRespondWith`
+          [json|[{"id":1,"computed_overload":true}]|]
+          { matchHeaders = [matchContentTypeJson] }
 
-      it "overloaded computed column on rpc" $
+      it "overloaded computed column on rpc" $ do
         get "/rpc/search?id=1&select=id,computed_overload" `shouldRespondWith`
+          [json|[{"id":1,"computed_overload":true}]|]
+          { matchHeaders = [matchContentTypeJson] }
+        get "/rpc/search?id=1&select=*,computed_overload" `shouldRespondWith`
           [json|[{"id":1,"computed_overload":true}]|]
           { matchHeaders = [matchContentTypeJson] }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3149, Misleading "Starting PostgREST.." logs on schema cache reloading - @steve-chavez
  - #2815, Build static executable with GSSAPI support - @wolfgangwalther
  - #3205, Fix wrong subquery error returning a status of 400 Bad Request - @steve-chavez
+ - #2688, #3145 Fix selecting overloaded computed columns alongside `*` when returning tables from INSERT and RPC - @wolfgangwalther, @laurenceisla
 
 ### Deprecated
 
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -947,7 +947,7 @@ inferColsEmbedNeeds :: ReadPlanTree -> [FieldName] -> [FieldName]
 inferColsEmbedNeeds (Node ReadPlan{select} forest) pkCols
   -- if * is part of the select, we must not add pk or fk columns manually -
   -- otherwise those would be selected and output twice
-  | "*" `elem` fldNames = ["*"]
+  | "*" `elem` fldNames = fldNames
   | otherwise           = returnings
   where
     fldNames = cfName . csField <$> select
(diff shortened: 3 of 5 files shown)


test/spec/Feature/Query/QuerySpec.hs
```diff
@@ -533,11 +533,17 @@ spec actualPgVersion = do
         get "/items2?id=eq.1&select=id,computed_overload" `shouldRespondWith`
           [json|[{"id":1,"computed_overload":true}]|]
           { matchHeaders = [matchContentTypeJson] }
+        get "items?id=eq.1&select=*,computed_overload" `shouldRespondWith`
```
laurenceisla, line 536: This didn't give an error before. Added it for completion, just in case.

laurenceisla:
@wolfgangwalther Yeah, so I just checked that it's an issue with the way we do the selection, filtering, etc. when a mutation or function call returns a table value.

The insertion/function call is done inside the top CTE, while the selection, filtering, etc. is done in the last subquery that gets all the values. 

For example, querying a function `get_projects` that returns a set of `projects`, using the `created_at` overloaded computed column from this PR:

```bash
curl "http://localhost:3000/rpc/get_projects?id=1&select=*,created_at&order=created_at"
```

Results in this query (added explanations as comments):
  
```sql
WITH pgrst_source AS
  (SELECT "pgrst_call".*,
    -- The fix adds the computed column here, but it's retrieved again in _postgrest_t, duplicating the result:
    "pgrst_call"."created_at"
   FROM
     (SELECT '{"id": 1}'::json AS json_data) pgrst_payload,
   LATERAL
     (SELECT CASE WHEN json_typeof(pgrst_payload.json_data) = 'array' THEN pgrst_payload.json_data ELSE json_build_array(pgrst_payload.json_data) END AS val) pgrst_uniform_json,
   LATERAL
     (SELECT "id"
      FROM json_to_recordset(pgrst_uniform_json.val) AS _("id" integer) LIMIT 1) pgrst_body,
   LATERAL "api"."get_projects"("id" := pgrst_body."id") pgrst_call)
       1 AS page_total,
       coalesce(json_agg(_postgrest_t)->0, 'null') AS body,
       nullif(current_setting('response.status', TRUE), '') AS response_status,
       '' AS response_inserted
FROM
  (SELECT "projects".*,
          -- This duplicates the result
          "projects"."created_at"
   -- This subquery does not know that `pgrst_source` is an `projects` table, that's why it errs with "column created_at does not exist"
   FROM "pgrst_source" AS "projects"
   -- The order by also gives the error for the same reason
   ORDER BY "projects"."created_at" LIMIT 1000
   OFFSET 0) _postgrest_t
```

In fact, the filter `created_at=eq.something`  or order `order=created_at` does not need a `select=*`  to fail, it does directly on the main branch and here:

```bash
curl "http://localhost:3000/rpc/get_projects?i=1&order=created_at"
{"code":"42703","details":null,"hint":null,"message":"column projects.created_at does not exist"}
```

A solution would be to somehow make the subquery now that the top CTE is of a `projects` type in this example or do the filtering in the same CTE (this would open other problems I'm sure). I'll look for alternatives later, let me know if you have any ideas.

laurenceisla:
Yes, it's the same read plan for simple selects in the sub query. 

> This might be tricky or it might not be. It will certainly need a bit of a refactor. But ultimately, this would be the most consistent.

 I have the same idea, this would be the go to. I'm going to give it a try and see if it's not too complicated. If it is, I'll close this PR since I'm prioritizing other issues right now.

laurenceisla:
Since what we want is to do the filtering/ordering/embedding in the main query (inside the CTE), then this won't work for the INSERT case. That's because the `INSERT ... RETURNING *` needs a CTE to do all those operations, which will inevitable generate the "columns not found" errors. It would need to cover the special cases 1-by-1.

For RPCs doing the operations inside the CTE is more feasible, but the solution for INSERTS would work here too, not sure if duplicating the effort is worth it. So I'm closing this for now.

PR #3233, opened 2024-02-17
feat: dump schema cache through admin API

### LINKED ISSUE #3210: Dump Schema Cache through Admin API
## Problem

No way to dump the schema cache of a running instance, it's only possible to do it at startup with:

```
postgrest-with-postgresql-16 postgrest-run --dump-schema
```

## Solution

Use the admin api ([ref](https://postgrest.org/en/stable/references/admin.html#health-check)) to output the schema cache of a running instance:

```bash
$ curl localhost:3001/schema_cache

200 OK

{..}
```

taimoorzaeem:
Fixes #3210.


steve-chavez (approved):
LGTM :+1:

steve-chavez:
@wolfgangwalther You're right. Let's do that from now on.

PR #3235, opened 2024-02-18
ci: Improve github actions workflows

wolfgangwalther:
The goal of this PR is to make Pull Requests run only those jobs they need to run, i.e. a docs-only PR should only run the docs jobs. A PR only touching tests, should only run the test-related jobs, and a PR touching code should run both Build and Test jobs.

GitHub Actions only allows `paths` on the workflow level, thus to achieve this, I had to split the whole thing into multiple workflows. This comes with it's own set of challenges, though, because especially the release jobs need to wait for the other jobs to finish, so they can't just run in parallel. Also, artifacts need to be available from build to release. To achieve this, we are using "Reusable Workflows" - i.e. the build, docs and test workflows are now run on their own in a PR - but as a nested workflow on any push event.

TLDR:
- Pull Requests may only run Build, Docs or Test - depending on what has changed.
- Pushes to main / release branches always run everything.
- Pushes to tags run everything + release jobs.

TODO:
- [ ] Split release jobs into separate workflow file to prevent all the skipped release jobs from showing up in regular branch pushes.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -32,6 +32,7 @@ jobs:
     - run: postgrest-docs-build
     - run: git diff --exit-code HEAD locales || echo "Please commit changes to the locales/ folder after running postgrest-docs-build."
 
+
   spellcheck:
     name: Spellcheck
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


laurenceisla:
The FreeBSD build is still executing when only the docs are changed.

https://github.com/PostgREST/postgrest/pull/3296/checks?check_run_id=22142036339

PR #3238, opened 2024-02-19
fix: Dump media handlers and timezones with --dump-schema

### LINKED ISSUE #3237: `--dump-schema` does not dump media handlers
### Environment

* PostgreSQL version: 16
* PostgREST version: 12.1 (development)
* Operating system: Linux

### Description of issue

While trying to debug another possible regression in the current development version, I wanted to see the media handlers via `--dump-schema`.

(Edit: The following commands only work after 4a796e0)

Running `postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run` responds with:

```
...
Schema cache loaded 264 Relations, 221 Relationships, 141 Functions, 15 Domain Representations, 45 Media Type Handlers
```

Nice!

But, `postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run --dump-schema | jq '.dbMediaHandlers'` then gives me:

```
[]
```

Something is broken there.

In fact, counting the array items of the json response:

```
{
  "dbMediaHandlers": 0,
  "dbRelationships": 221,
  "dbRepresentations": 15,
  "dbRoutines": 141,
  "dbTables": 264,
  "dbTimezones": 0
}
```

Not sure if timezones is broken as well?

wolfgangwalther:
Those were left out of the schema dump when the features were introduced, probably because ByteString doesn't have a toJSON instance. Changing the type to Text solves this easily.

Resolves #3237

### DIFF
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveAnyClass        #-}
 {-# LANGUAGE DeriveGeneric         #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 
@@ -10,6 +11,7 @@ module PostgREST.MediaType
   , decodeMediaType
   ) where
 
+import qualified Data.Aeson               as JSON
 import qualified Data.ByteString          as BS
 import qualified Data.ByteString.Internal as BS (c2w)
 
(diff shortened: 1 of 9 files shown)


src/PostgREST/MediaType.hs
```diff
@@ -28,23 +30,23 @@ data MediaType
...
 data MTVndPlanOption
   = PlanAnalyze | PlanVerbose | PlanSettings | PlanBuffers | PlanWAL
-  deriving (Eq, Show, Generic)
+  deriving (Eq, Show, Generic, JSON.ToJSON)
 instance Hashable MTVndPlanOption
 
 data MTVndPlanFormat
   = PlanJSON | PlanText
-  deriving (Eq, Show, Generic)
+  deriving (Eq, Show, Generic, JSON.ToJSON)
```
steve-chavez, line 49: How about a test on `test/io/test_cli.py` to clear these codecov warnings

test/io/test_cli.py
```diff
@@ -225,3 +235,12 @@ def test_invalid_openapi_mode(invalidopenapimodes, defaultenv):
         for line in dump.split("\n"):
             if line.startswith("openapi-mode"):
                 print(line)
+
+
+# If this test is failing, run pytest with --snapshot-update.
+def test_schema_cache_snapshot(baseenv, snapshot_yaml):
+    "Dump of schema cache should match snapshot."
+
+    schema_cache = yaml.load(cli(["--dump-schema"], env=baseenv), Loader=yaml.Loader)
```
steve-chavez, line 244: This converts the JSON to YAML right?

src/PostgREST/MediaType.hs
steve-chavez: Hm, maybe it's because there are no media handlers on `test/io/fixtures.sql`?

steve-chavez: Checking how it can be refactored so this test works..

PR #3239, opened 2024-02-19
docs: add using `or` across embedded resources

laurenceisla:
Adds an example on using `or` across multiple resources to the Null Embedding section.

### DIFF
--- a/docs/references/api/resource_embedding.rst
+++ b/docs/references/api/resource_embedding.rst
@@ -852,7 +852,7 @@ An ``or`` filter can be used for a similar operation:
 
   curl "http://localhost:3000/films?select=*,roles(*)&roles.or=(character.eq.Gummo,character.eq.Zeppo)"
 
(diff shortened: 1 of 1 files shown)


docs/references/api/resource_embedding.rst
```diff
@@ -951,6 +953,13 @@ Both ``is.null`` and ``not.is.null`` can be included inside the `or` operator. F
 
   curl "http://localhost:3000/films?select=title,actors(*),directors(*)&or=(actors.is.null,directors.is.null)"
 
+You can also use ``not.is.null`` as a workaround for an ``or`` filter across multiple resources.
+For instance, to show the films whose actors **or** directors are named John:
```
steve-chavez, line 957: @laurenceisla We need a subtitle here, like "or across embedded resources". Also, "workaround" should be removed. Supporting `or` like this is intentional.. although not succinct.

steve-chavez (approved):

PR #3250, opened 2024-02-20
refactor: make observation messages pure

steve-chavez:
Removes the observation messages from the Logger.

Also, this makes it possible to test the error messages without IO (using a different `observer` in the tests).


steve-chavez:
Codecov failures are still related to the missing recovery tests https://github.com/PostgREST/postgrest/issues/1766

PR #3316, opened 2024-03-08
perf: remove json_typeof

### LINKED ISSUE #2676: Optimizations for INSERT

steve-chavez:
Closes #2676. Removes `json_typeof/jsonb_typeof` by using some light validation on the json body using lazy bytestring functions.

- grants ~10% perf increase. 
  + can be confirmed with the PlanSpec change.
  + can be confirmed manually with pgbench like:
```
$ postgrest-with-postgresql-16 -f test/pgbench/fixtures.sql pgbench -U postgres -n -T 10 -f test/pgbench/2676/old.sql 
tps = 3480.654744 (without initial connection time)

$ postgrest-with-postgresql-16 -f test/pgbench/fixtures.sql pgbench -U postgres -n -T 10 -f test/pgbench/2676/new.sql 
tps = 3973.138881 (without initial connection time)
```
- memory usage is kept the same
- no need for a new `content-type` as per a previous idea on https://github.com/PostgREST/postgrest/issues/2676. This is an improvement for every write case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3061, Apply all function settings as transaction-scoped settings - @taimoorzaeem
  - #3171, #3046, Log schema cache stats to stderr - @steve-chavez
  - #3210, Dump schema cache through admin API - @taimoorzaeem
+ - #2677, Performance improvement on bulk json inserts, around 10% increase on requests per second by removing `json_typeof` from write queries - @steve-chavez
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3061, Apply all function settings as transaction-scoped settings - @taimoorzaeem
  - #3171, #3046, Log schema cache stats to stderr - @steve-chavez
  - #3210, Dump schema cache through admin API - @taimoorzaeem
- - #2677, Performance improvement on bulk json inserts, around 10% increase on requests per second by removing `json_typeof` from write queries - @steve-chavez
+ - #2676, Performance improvement on bulk json inserts, around 10% increase on requests per second by removing `json_typeof` from write queries - @steve-chavez
 
 ### Fixed
 
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -300,19 +300,17 @@ fromJsonBodyF :: Maybe LBS.ByteString -> [CoercibleField] -> Bool -> Bool -> Boo
 fromJsonBodyF body fields includeSelect includeLimitOne includeDefaults =
   (if includeSelect then "SELECT " <> namedCols <> " " else mempty) <>
   "FROM (SELECT " <> jsonPlaceHolder <> " AS json_data) pgrst_payload, " <>
-  -- convert a json object into a json array, this way we can use json_to_recordset for all json payloads
-  -- Otherwise we'd have to use json_to_record for json objects and json_to_recordset for json arrays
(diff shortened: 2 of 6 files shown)


steve-chavez:
> grants ~14% perf increase.

The above is true in pure SQL. But I haven't been able to reproduce a notable gain through http calls.

```bash
# Using https://github.com/PostgREST/postgrest-benchmark

$ postgrest-bench-k6 10 k6/POSTBulk.js # with 12.0.2

    http_reqs..................: 25959  822.396728/s

$ postgrest-bench-k6 10 k6/POSTBulk.js # with master

    http_reqs..................: 28556  906.419157/s
```

The loadtest in CI also shows the same.

steve-chavez:
Right, my bad. I was looking at some `k6/POSTSingle.js` (posts single object) results I had when I finished writing that. There isn't a noticeable gain there. But thinking more about it, it makes sense since `json_typeof` usage should mostly affect large payloads.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3061, Apply all function settings as transaction-scoped settings - @taimoorzaeem
  - #3171, #3046, Log schema cache stats to stderr - @steve-chavez
  - #3210, Dump schema cache through admin API - @taimoorzaeem
+ - #2677, Performance improvement on bulk json inserts, around 10% increase on requests per second by removing `json_typeof` from write queries - @steve-chavez
```
laurenceisla, line 15:
```suggestion
 - #2676, Performance improvement on bulk json inserts, around 10% increase on requests per second by removing `json_typeof` from write queries - @steve-chavez
```

laurenceisla (approved):

PR #3321, opened 2024-03-09
add CLI to docs

### LINKED ISSUE #3248: Add CLI page
The CLI is undocumented right now. 

We need this page to document https://github.com/PostgREST/postgrest/pull/2856/.

There's some extra work here:

- `--dump-schema` should be called `--dump-schema-cache`. Or perhaps `--dump-scache` as an alias.
- `--example` only shows a config file example. It should show env vars and maybe in-db config.
- A mention about using `--dump-schema` with `jq`.

taimoorzaeem:
Fixes #3248.

- [x] Add more config examples to CLI
- [x] Add CLI docs page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    - Exposes connection pool metrics, schema cache metrics
  - #3248, Add more config examples to CLI - @taimoorzaeem
    + Now accepts ``--example-file``, ``--example-db``, and ``--example-env``
+   + Add docs page for CLI
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3435, Add log-level=debug, for development purposes - @steve-chavez
  - #1526, Add `/metrics` endpoint on admin server - @steve-chavez
    - Exposes connection pool metrics, schema cache metrics
+ - #3248, Add more config examples to CLI - @taimoorzaeem
+   + Now accepts ``--example-file``, ``--example-db``, and ``--example-env``
 
 ### Fixed
 
--- a/docs/references/cli.rst
+++ b/docs/references/cli.rst
@@ -1,0 +1,41 @@
+.. _cli:
+
+CLI
+===
+
+PostgREST provides a CLI to start and run your postgrest service. The CLI provides the commands listed below:
+
+.. _cli_commands:
+
+CLI Commands
+------------
+
+Help and Version
+~~~~~~~~~~~~~~~~
+
+.. code:: bash
+
+  $ postgrest [-h|--help]
+  $ postgrest [-v|--version]
+
+
+Example Config
+~~~~~~~~~~~~~~
+
+.. code:: bash
+
+  $ postgrest [--example-file]
+  $ postgrest [--example-db]
+  $ postgrest [--example-env]
+
+These commands show the example configuration file of the selected type.
+
+
+Config or Schema Cache
+~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: bash
+  
+  $ postgrest [--dump-config|--dump-schema-cache] [FILENAME]
+
+Here ``FILENAME`` is the path to configuration file.
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -44,7 +44,7 @@ main CLI{cliCommand, cliPath} = do
       CmdDumpConfig -> do
         when configDbConfig $ AppState.reReadConfig True appState
         putStr . Config.toText =<< AppState.getConfig appState
-      CmdDumpSchema -> putStrLn =<< dumpSchema appState
+      CmdDumpSchemaCache -> putStrLn =<< dumpSchema appState
       CmdRun -> App.run appState)
 
 -- | Dump SchemaCache schema to JSON
@@ -61,6 +61,7 @@ dumpSchema appState = do
       exitFailure
     Right sCache -> return $ JSON.encode sCache
 
+
 -- | Command line interface options
 data CLI = CLI
   { cliCommand :: Command
@@ -70,7 +71,9 @@ data CLI = CLI
 data Command
   = CmdRun
   | CmdDumpConfig
-  | CmdDumpSchema
+  | CmdDumpSchemaCache
+
+data Example = ExampleFile | ExampleDb | ExampleEnv
 
 -- | Read command line interface options. Also prints help.
 readCLIShowHelp :: IO CLI
@@ -93,16 +96,28 @@ readCLIShowHelp =
         <> O.short 'v'
         <> O.help "Show the version information"
 
-    exampleParser =
-      O.infoOption exampleConfigFile $
-        O.long "example"
+    exampleParser = exampleParserFile <|> exampleParserDb <|> exampleParserEnv
+
+    exampleParserFile =
+      O.infoOption (exampleConfig ExampleFile) $
+        O.long "example-file"
         <> O.short 'e'
-        <> O.help "Show an example configuration file"
+        <> O.help "Show an example of a configuration file"
+
+    exampleParserDb =
+      O.infoOption (exampleConfig ExampleDb) $
+        O.long "example-db"
+        <> O.help "Show an example of in-database configuration"
+
+    exampleParserEnv =
+      O.infoOption (exampleConfig ExampleEnv) $
+        O.long "example-env"
+        <> O.help "Show an example of environment variables configuration"
 
     cliParser :: O.Parser CLI
     cliParser =
       CLI
-        <$> (dumpConfigFlag <|> dumpSchemaFlag)
+        <$> (dumpConfigFlag <|> dumpSchemaCacheFlag)
         <*> O.optional configFileOption
 
     configFileOption =
@@ -115,13 +130,13 @@ readCLIShowHelp =
         O.long "dump-config"
         <> O.help "Dump loaded configuration and exit"
 
-    dumpSchemaFlag =
-      O.flag CmdRun CmdDumpSchema $
-        O.long "dump-schema"
+    dumpSchemaCacheFlag =
+      O.flag CmdRun CmdDumpSchemaCache $
+        O.long "dump-schema-cache"
         <> O.help "Dump loaded schema as JSON and exit (for debugging, output structure is unstable)"
 
-exampleConfigFile :: [Char]
-exampleConfigFile =
+exampleConfig :: Example -> [Char]
+exampleConfig ExampleFile =
   [str|## Admin server used for checks. It's disabled by default unless a port is specified.
       |# admin-server-port = 3001
       |
@@ -198,6 +213,7 @@ exampleConfigFile =
       |## Choose a secret, JSON Web Key (or set) to enable JWT auth
       |## (use "@filename" to load from separate file)
       |# jwt-secret = "secret_with_at_least_32_characters"
+      |
       |jwt-secret-is-base64 = false
       |
       |## Enables and set JWT Cache max lifetime, disables caching with 0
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3171, #3046, Log schema cache stats to stderr - @steve-chavez
  - #3210, Dump schema cache through admin API - @taimoorzaeem
  - #2676, Performance improvement on bulk json inserts, around 10% increase on requests per second by removing `json_typeof` from write queries - @steve-chavez
+ - #3248, Add CLI page to docs - @taimoorzaeem
+   + Add config type to ``--example`` in CLI
```
laurenceisla, line 17: This should mention that `--example` now accepts the `file,db,env` parameters.

Also, the "Add CLI page to docs" should be in a separate line, I think. We used to have a [Documentation Improvements](https://postgrest.org/en/v10/releases/v10.0.0.html#documentation-improvements) section in the docs. Maybe a new `### Docs` subtitle is needed here?

src/PostgREST/CLI.hs
```diff
@@ -96,16 +117,21 @@ readCLIShowHelp =
...
+    exampleParser :: O.Parser CLI
+    exampleParser = 
+      CLIExample 
+        <$> showExample
+        <*> exampleConfig
+
+    exampleConfig =
+      O.strArgument $
+        O.metavar "ExampleType"
+        <> O.help "Type of config file/db/env"
```
laurenceisla, line 129: Doing this does not show `ExampleType` as a parameter of `--example` and it shows these options:

```
Usage: postgrest [-v|--version] 
                 [[-e|--example] ExampleType | 
                   [--dump-config | --dump-schema-cache] [FILENAME]]
```

When it should look something like:

```
Usage: postgrest-12.0.2 [-v|--version] [-e|--example EXAMPLETYPE]
                        [--dump-config | --dump-schema] [FILENAME]
```

You'll need to use directly `strOption` or `strArgument` (not sure which is better) and add `file` as a default value to avoid breaking changes. Here's an example on how to use this: https://www.fpcomplete.com/haskell/library/optparse-applicative/

src/PostgREST/CLI.hs
```diff
@@ -46,8 +46,21 @@ main CLI{cliCommand, cliPath} = do
...
+
+  loggerState <- Logger.init
+  bracket
+    (AppState.init conf $ Logger.logObservation loggerState)
+    AppState.destroy
+    (\_ -> case exampleType of
+      "file" -> putStrLn exampleConfigFile
+      "db" -> putStrLn exampleConfigDb
+      "env" -> putStrLn exampleConfigEnv
+      _     -> putStrLn ("Usage: postgrest-run --example/-e file/db/env" :: [Char]))
```
laurenceisla, line 63: I don't think you need to duplicate this code. You could keep `main CLI{...}` and maybe add another `cliCommand` like `CMDExample`, then the code to print the examples could be done in a `printExample` function (like `dumpSchema`, for instance).

src/PostgREST/CLI.hs
```diff
@@ -233,3 +267,157 @@ exampleConfigFile =
       |## When none is provided, 660 is applied by default
       |# server-unix-socket-mode = "660"
       |]
+
+-- Shown in CLI "--example db"
+exampleConfigDb :: [Char]
+exampleConfigDb = 
+  [str|## reloadable config options
+      |ALTER ROLE db_config_authenticator SET pgrst.admin_server_port = '3001';
```
laurenceisla, line 275:
```suggestion
      |ALTER ROLE authenticator SET pgrst.admin_server_port = '3001';
```

We use `authenticator` in the docs, so it's better to use it here.

src/PostgREST/CLI.hs
```diff
@@ -233,3 +267,157 @@ exampleConfigFile =
...
+
+-- Shown in CLI "--example env"
+exampleConfigEnv :: [Char]
+exampleConfigEnv = 
+  [str|## example env config
+      |PGRST_APP_SETTINGS_test2: test
+      |
+      |PGRST_APP_SETTINGS_test: test
+      |
+      |PGRST_DB_AGGREGATES_ENABLED: true
```
laurenceisla, line 348:
```suggestion
      |export PGRST_DB_AGGREGATES_ENABLED=false
```

Since it's an example on how to set it, it should follow that form. Also, the values must be the default ones, and the optional values should be commented (the same goes for the in-db variables). That's how the `files` example is already built.

taimoorzaeem:
@laurenceisla How about the interface now? It is as best as I could make yet. The only problem is that ``postgrest-run`` stopped working with this change. Could you help me figure this out? I'll add the comments etc in the following commits.

src/PostgREST/CLI.hs
```diff
@@ -64,25 +68,33 @@ dumpSchema appState = do
...
+  { example :: Example
+  , command :: Commands
+  }
+
+type Example = [Char]
+
+data Commands = Commands
   { cliCommand :: Command
   , cliPath    :: Maybe FilePath
   }
```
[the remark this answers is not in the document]
taimoorzaeem: The problem with adding ``CmdExample ExampleType`` to ``Command`` is that this is a value option. The other enums ``CmdDumpConfig`` and ``CmdDumpSchema`` are handled as flags. Combining them will complicate things in implementation as well. But maybe it is doable. I'll try this and get back to you.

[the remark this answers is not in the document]
taimoorzaeem: Also using this would make the ``interface help`` look like this:
```
Usage: postgrest-12.0.2 [-v|--version]
                        [--example EXAMPLETYPE | --dump-config | --dump-schema] [FILENAME]

```
instead of this more nice looking one:
```
Usage: postgrest-12.0.2 [-v|--version] [-e|--example EXAMPLETYPE]
                        [--dump-config | --dump-schema] [FILENAME]

```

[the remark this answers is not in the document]
laurenceisla: > Also using this would make the interface help look like this:
> 
> ```
> Usage: postgrest-12.0.2 [-v|--version]
>                         [--example EXAMPLETYPE | --dump-config | --dump-schema] [FILENAME]
> ```
> 

This means that `--example EXAMPLETYPE`, `--dump-config` and `--dump-schema` are mutually exclusive, which makes sense now. Before, you could do something like:

```sh
postgrest --example --dump-config
# or
postgrest --dump-config --example
```

And it will ignore the `--dump-config` and print the example directly. But now that `--example` has an argument, it complicates things as Wolfgang mentioned. So I think it's OK to have them mutually exclusive like that.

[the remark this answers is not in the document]
taimoorzaeem: @laurenceisla  I think the above is not possible to implement as it is inconsistent. Adding an argument option in `Command` that was previously just enum does not work.
As I have looked further into this, a pattern like:
```haskell
data Command
  = ...
  | CmdExample ExampleType
  | Something Argument
```
is used in subparsers. [ref](https://github.com/pcapriotti/optparse-applicative?tab=readme-ov-file#commands). Hence, implementing the above is inconsistent if not impossible. How about using subparsers to lift the interface something like:
```
$ postgrest dump [--dump-config | --dump-schema-cache]
$ postgrest example [--file | --db | --env]
```
I think it seems much simpler, extendible and easier to implement.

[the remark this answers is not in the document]
taimoorzaeem: @laurenceisla 
I have made the interface like you said earlier:
```
Usage: postgrest-12.0.2 [-v|--version]
                        [--example EXAMPLETYPE | --dump-config | --dump-schema] [FILENAME]

```
Please confirm and review this design so I can move to testing.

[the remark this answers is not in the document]
laurenceisla: Argh... this gets complicated due to the need of a parameter and default values for the `--example`. It's in conflict with the `CLI` type expected by the parser and generates the problems that both of you encountered along the way.

> How about using subparsers to lift the interface something like:

I think this may be the way to go in the future. It would need a rework of the CLI, I suppose.

> Please confirm and review this design so I can move to testing.

I'm sorry to keep going back and forth but what if, instead of using an option with value, we do the same as with `--dump-*`. I mean, to use: 

 `--example-file`, `--example-env`, `--example-db`. 

This would work as an extension of the already used `--example` and wouldn't change nor break the way the CLI is currently used. The drawback is that it's more verbose.

@taimoorzaeem @wolfgangwalther WDYT? I'm adding how the parsing would go in a review below.

src/PostgREST/CLI.hs
```diff
@@ -73,7 +75,10 @@ data CLI = CLI
 data Command
   = CmdRun
   | CmdDumpConfig
-  | CmdDumpSchema
+  | CmdDumpSchemaCache
+  | CmdExample ExampleType
+
+data ExampleType = ExampleFile | ExampleDb | ExampleEnv
```
laurenceisla, line 81:
```suggestion

data Example = ExampleFile | ExampleDb | ExampleEnv
```

src/PostgREST/CLI.hs
```diff
@@ -82,7 +87,8 @@ readCLIShowHelp =
   where
     prefs = O.prefs $ O.showHelpOnError <> O.showHelpOnEmpty
     opts = O.info parser $ O.fullDesc <> progDesc
-    parser = O.helper <*> versionFlag <*> exampleParser <*> cliParser
+    -- TODO: use subparser for mulitple commands
+    parser = O.helper <*> versionFlag <*> cliParser
```
laurenceisla, line 91:
```suggestion
    parser = O.helper <*> versionFlag <*> exampleParser <*> cliParser
```

src/PostgREST/CLI.hs
```diff
@@ -96,35 +102,46 @@ readCLIShowHelp =
...
+        readString "env"  = Just $ CmdExample ExampleEnv
+        readString _      = Just $ CmdExample ExampleFile -- default "file"
+
+
+    dumpExampleConfigFile =
+      O.option parseExampleType $
+        O.long "example"
+        <> O.short 'e'
+        <> O.metavar "EXAMPLETYPE"
+        <> O.help "Type of config file/db/env"
```
laurenceisla, line 130:
```suggestion
    exampleParser =
      exampleParserFile <|> exampleParserEnv <|> exampleParserDb

    exampleParserFile =
      O.infoOption (exampleConfig ExampleFile) $
        O.long "example"
        <> O.long "example-file"
        <> O.short 'e'
        <> O.help "Show an example of a configuration file"

    exampleParserEnv =
      O.infoOption (exampleConfig ExampleEnv) $
        O.long "example-env"
        <> O.help "Show an example of environment variables configuration"

    exampleParserDb =
      O.infoOption (exampleConfig ExampleDb) $
        O.long "example-db"
        <> O.help "Show an example of in-database configuration"
```

laurenceisla (commented):
So this is how my suggestion in https://github.com/PostgREST/postgrest/pull/3321#discussion_r1558458869  would be implemented (other minor changes may be needed too):

src/PostgREST/CLI.hs
```diff
@@ -230,3 +246,227 @@ exampleConfigFile =
...
+      |# ALTER ROLE authenticator SET pgrst.db_pool_acquisition_timeout = '10';
+      |
+      |## Time in seconds after which to recycle pool connections
+      |# ALTER ROLE authenticator SET pgrst.db_pool_max_lifetime = '1800';
+      |
+      |## Time in seconds after which to recycle unused pool connections
+      |# ALTER ROLE authenticator SET pgrst.db_pool_max_idletime = '30';
+      |
+      |## Allow automatic database connection retrying
+      |# ALTER ROLE authenticator SET pgrst.db_pool_automatic_recovery = 'true';
```
steve-chavez, line 290: Oh, these don't work. The available in-db settings are the ones here https://github.com/PostgREST/postgrest/blob/85a4e35a2de869ddf9b6c5f6893a3154767cbc76/src/PostgREST/Config/Database.hs#L45-L71

docs/references/cli.rst
```diff
@@ -0,0 +1,41 @@
...
+.. code:: bash
+
+  $ postgrest [--example-file]
+  $ postgrest [--example-db]
+  $ postgrest [--example-env]
+
+These commands show the example configuration file of the selected type.
+
+
+Config or Schema Cache
```
steve-chavez, line 34: How about separating these? One section for config and another for the scache.

docs/references/cli.rst
```diff
@@ -0,0 +1,41 @@
...
+
+
+Example Config
+~~~~~~~~~~~~~~
+
+.. code:: bash
+
+  $ postgrest [--example-file]
+  $ postgrest [--example-db]
+  $ postgrest [--example-env]
```
steve-chavez, line 29: @taimoorzaeem Btw, if you'd like you could split this and do another PR with just the docs commit. That one is already mergeable.

Of course you'd need to remove the `--example-db/--example/env` in that commit since they belong to this PR.

taimoorzaeem: Hmm, you're right. That would be much more manageable. Closing this PR.

PR #3339, opened 2024-03-19
perf: fix space leaks by downgrading fuzzyset to v0.2.4

### LINKED ISSUE #3329: Stack Overflow error in dev enviroment when table and embed resource do not exist
### Environment

* PostgreSQL version: 16.1
* PostgREST version: main branch
* Operating system: NixOS 23.11

### Description of issue

When requesting a table that does not exist with any embedded resource, it shows a `stack overflow` error in the logs and fails the request. To reproduce:

```console
$ nix-shell --pure --run "postgrest-with-postgresql-16 -f ./test/spec/fixtures/load.sql postgrest-run"
```
```console
$ curl "localhost:3000/x?select=y(*)"
```
Expected:
```
$ curl "localhost:3000/x?select=y(*)"
{
  "code":"PGRST200",
  "details":"Searched for a foreign key relationship between 'x' and 'y' in the schema 'public', but no matches were found.",
  "hint":null,
  "message":"Could not find a relationship between 'x' and 'y' in the schema cache"
}
```

This only happens in the Nix development environment  for now (not for spec tests), since it enables the leak detection with `-K1K`:

https://github.com/PostgREST/postgrest/blob/ec7ab271d69e8ad907328b34c7d1a4aed1c60af9/postgrest.cabal#L185

It appears that the `fuzzyset` package, used to build the "hint" of the error, is the culprit. The evidence is that it is marked as "broken" in the latest update of nixpkgs/GHC in commit 6682baeb (previous commits do not throw the error even with `-K1K`):

https://github.com/PostgREST/postgrest/blob/ec7ab271d69e8ad907328b34c7d1a4aed1c60af9/nix/overlays/haskell-packages.nix#L48-L49

laurenceisla:
Closes #3329.

### DIFF
--- a/test/io/test_big_schema.py
+++ b/test/io/test_big_schema.py
@@ -70,3 +70,21 @@ def test_openapi_in_big_schema(defaultenv):
     with run(env=env) as postgrest:
         response = postgrest.session.get("/")
         assert response.status_code == 200
+
+
+# See: https://github.com/PostgREST/postgrest/issues/3329
+def test_should_not_fail_with_stack_overflow(defaultenv):
+    "requesting a non-existent relationship should not fail with stack overflow due to fuzzy search of candidates"
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_SCHEMAS": "apflora",
+        "PGRST_DB_POOL": "2",
+        "PGRST_DB_ANON_ROLE": "postgrest_test_anonymous",
+    }
+
+    with run(env=env, wait_max_seconds=30) as postgrest:
+        response = postgrest.session.get("/unknown-table?select=unknown-rel(*)")
+        assert response.status_code == 400
+        data = response.json()
+        assert data["code"] == "PGRST200"
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -45,8 +45,15 @@ let
           }
           { };
 
-      # Marked as broken (?)
-      fuzzyset = lib.markUnbroken prev.fuzzyset;
+      # Before upgrading fuzzyset to 0.3, check: https://github.com/PostgREST/postgrest/issues/3329
+      fuzzyset =
+        prev.callHackageDirect
+          {
+            pkg = "fuzzyset";
+            ver = "0.2.4";
+            sha256 = "sha256-lpkrTFcR0B4rT/P6x7ui31Twgq7BBj6KIvjKyqXKdpc=";
+          }
+          { };
 
       postgresql-libpq = lib.dontCheck
         (prev.postgresql-libpq_0_10_0_0.override {
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -104,7 +104,7 @@ library
                     , directory                 >= 1.2.6 && < 1.4
                     , either                    >= 4.4.1 && < 5.1
                     , extra                     >= 1.7.0 && < 2.0
-                    , fuzzyset                  >= 0.3.0
+                    , fuzzyset                  >= 0.2.4 && < 0.3
                     , gitrev                    >= 1.2 && < 1.4
                     , hasql                     >= 1.6.1.1 && < 1.7
                     , hasql-dynamic-statements  >= 0.3.1 && < 0.4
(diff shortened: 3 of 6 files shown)


steve-chavez:
@laurenceisla How about a test for this? We now have the `test/io/test_big_schema.py`. You should be able to reproduce the `stack overflow` there, I think.

test/io/test_big_schema.py
```diff
@@ -70,3 +70,21 @@ def test_openapi_in_big_schema(defaultenv):
     with run(env=env) as postgrest:
         response = postgrest.session.get("/")
         assert response.status_code == 200
+
+
+# See: https://github.com/PostgREST/postgrest/issues/3329
+def test_should_not_fail_with_stack_overflow(defaultenv):
```
laurenceisla, line 76: This test fails on the `main` branch with "stack overflow".

steve-chavez:
> Don't know the details, but it's similar in that it takes more space in memory than it's necessary, but it's freed later.

Just to clarify. It's not a memory leak, is just that we limit the stack size during development to `1K` with:

https://github.com/PostgREST/postgrest/blob/0d4441852e1c2383227f03ee189e563c611b0966/postgrest.cabal#L185

Which forces a `stack overflow` error.

More details at https://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html?m=1

---

Don't know `fuzzyset` details and why it consumes noticeable memory now. It'd be better to open an issue upstream to see if it can be fixed.

steve-chavez:
> What is a "space leak"? 

Actually the term was correctly used:

> A space leak occurs when a computer program uses more memory than necessary. In contrast to memory leaks, where the leaked memory is never released, the memory consumed by a space leak is released, but later than expected.

References:

- https://queue.acm.org/detail.cfm?id=2538488
- https://stackoverflow.com/questions/46007746/whats-a-space-leak
- https://www.reddit.com/r/haskell/comments/pvosen/how_can_haskell_programmers_tolerate_space_leaks/
- https://chshersh.com/space-leak
- https://softwareengineering.stackexchange.com/questions/306226/does-it-make-sense-to-use-the-term-space-leak-with-regard-to-java

laurenceisla:
> But maybe we can / should downgrade temporarily while still reporting the problem upstream.

Agree, opened an issue there: https://github.com/laserpants/fuzzyset-haskell/issues/9

taimoorzaeem:
@laurenceisla With this downgrade, `postgrest-test-doctests` fails with the following:
```
src/PostgREST/Error.hs:294:26: error:
    Not in scope: ‘Fuzzy.fromList’
    NB: the module ‘Data.FuzzySet’ does not export ‘fromList’.

src/PostgREST/Error.hs:295:26: error:
    Not in scope: ‘Fuzzy.fromList’
    NB: the module ‘Data.FuzzySet’ does not export ‘fromList’.

src/PostgREST/Error.hs:296:21: error:
    Not in scope: ‘Fuzzy.getOne’
    NB: the module ‘Data.FuzzySet’ does not export ‘getOne’.

src/PostgREST/Error.hs:298:43: error:
    Not in scope: ‘Fuzzy.get’
    NB: the module ‘Data.FuzzySet’ does not export ‘get’.
.
.
.
```

taimoorzaeem:
> Strange. Why did this not fail in CI. I think this was the latest run here: https://github.com/PostgREST/postgrest/actions/runs/8349585526/job/22854079725. Doctests seem fine here.

On more thought, I think it is some local issue on my side. My bad.

laurenceisla:
@steve-chavez Strange, I'm not getting those errors:

```console
[nix-shell:~/Projects/postgrest]$ postgrest-test-doctests
Warning: Both /home/laurence/.cabal and /home/laurence/.config/cabal/config
exist - ignoring the former.
It is advisable to remove one of them. In that case, we will use the remaining
one by default (unless '$CABAL_DIR' is explicitly set).

src/PostgREST/Version.hs:12:1: error:
    Could not find module ‘Paths_postgrest’
    Perhaps you meant Paths_doctest
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
Examples: 153  Tried: 153  Errors: 0  Failures: 0
```

Maybe your nix-shell is caching something? `postgrest-clean` could help.

steve-chavez:
Ah, right. Just reentering `nix-shell` made those errors go.

PR #3346, opened 2024-03-22
fix: in-db config values not loading for `pgrst.server_trace_header` and `pgrst.server_cors_allowed_origins`

### LINKED ISSUE #3345: `pgrst.server_trace_header` in-database config not working
## Problem

The https://postgrest.org/en/v12/references/configuration.html#server-trace-header is not working via:

```sql
ALTER ROLE postgrest_test_authenticator SET pgrst.server_trace_header = 'X-Y';
```

But it does through config or environment variable.

laurenceisla:
Closes #3345


laurenceisla:
I tried using `reReadConfig` here, since it looks like the in-db configuration is not loaded:

https://github.com/PostgREST/postgrest/blob/f1f01f1f5c9d55ff7b61e43472f0b79de0dd9f25/src/PostgREST/App.hs#L70-L74

But after reloading the `config` the values will not update, since it will keep the old config values specified above.

---

Still, I'm not entirely sure if this is the best solution but it works in using the updated in-db configuration values.

test/io/db_config.sql
```diff
@@ -1,75 +1,77 @@
 CREATE ROLE db_config_authenticator LOGIN NOINHERIT;
 
 -- reloadable config options
-ALTER ROLE db_config_authenticator SET pgrst.jwt_aud = 'https://example.org';
-ALTER ROLE db_config_authenticator SET pgrst.openapi_server_proxy_uri = 'https://example.org/api';
-ALTER ROLE db_config_authenticator SET pgrst.jwt_secret = 'REALLY=REALLY=REALLY=REALLY=VERY=SAFE';
-ALTER ROLE db_config_authenticator SET pgrst.jwt_secret_is_base64 = 'false';
-ALTER ROLE db_config_authenticator SET pgrst.jwt_role_claim_key = '."a"."role"';
-ALTER ROLE db_config_authenticator SET pgrst.jwt_cache_max_lifetime = '3600';
+-- these settings will override the values in configs/no-defaults.config, so they must be different
```
laurenceisla, line 4: I noticed that some values used here were the same than the ones in the `no-defaults.config`. Since we want to test if they are overridden correctly, then the values must be different from each other, that's why I changed them (also reordered the settings alphabetically while I'm at it).

steve-chavez (approved):
LGTM. Nice work :100:

steve-chavez:
[server-timing-enabled](https://postgrest.org/en/v12/references/configuration.html#server-timing-enabled) makes a lot of sense for instance. If you'd want to quickly get timings for a request. Changing that through the db is much quicker than through the config.

`server_cors_allowed_origins` and `server-trace-header` might not be that useful as in-db config. But I don't see why not.

IMO the only things that shouldn't are `server-host`, port, etc.. things that could bring down the postgREST service.

steve-chavez:
> Maybe I feel

Perhaps we can settle this with a shared design principle. 

I'd like to pursue extreme-database centric architecture ([pdf](https://revistas.um.edu.uy/index.php/ingenieria/article/view/1147/1496)), which goes a bit beyond "db as source of truth". Right now we're kinda db-centric, since we're two tier but I'd like to get closer to the extreme part. 

Right now for PostgREST this means that everything should be configurable through the db. Right now we cannot do the `server-host`, `server-port` and others but there could be a future where we do.

WDYT?

Issue #3361, opened 2024-03-29
Inserts and Updates now require an up-to-date schema cache


steve-chavez:
## Problem

Start postgREST:

```
$ postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run
```

Add a new column:

```sql
alter table test.projects add column new text;
```

Then:

```bash
http POST localhost:3000/projects <<JSON
{"id": 343, "new": "foo"}
JSON

HTTP/1.1 400 Bad Request
{
    "code": "PGRST204",
    "details": null,
    "hint": null,
    "message": "Column 'new' of relation 'projects' does not exist"
}

http POST "localhost:3000/projects?columns=id,new" <<JSON
{"id": 343, "new": "foo"}
JSON

HTTP/1.1 400 Bad Request
{
    "code": "PGRST204",
    "details": null,
    "hint": null,
    "message": "Column 'new' of relation 'projects' does not exist"
}

http PATCH "localhost:3000/projects?id=eq.1" <<JSON
{"new": "foo"}
JSON

HTTP/1.1 400 Bad Request
{
    "code": "PGRST204",
    "details": null,
    "hint": null,
    "message": "Column 'new' of relation 'projects' does not exist"
}
```

And this happens when using `columns` or not.

Seems this was introduced on https://github.com/PostgREST/postgrest/pull/2839. The problem is that it's not documented on https://postgrest.org/en/v12/references/schema_cache.html.

## Solution

Update the docs. While at it, make the error message more clear. Right now it seems it comes from PostgreSQL.

## Notes

Maybe it's a good thing to behave like this now since we'll reject the request at the plan level and save pool connections. This whenever the user mistypes a column.

PR #3362, opened 2024-03-29
fix: clarify PGRST204 error message

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/3361


steve-chavez:
We already have the "schema cache" message in others errors.

https://github.com/PostgREST/postgrest/blob/594b41aed4752d521da09dfaaebebbc269110355/test/spec/Feature/Query/QuerySpec.hs#L617

https://github.com/PostgREST/postgrest/blob/594b41aed4752d521da09dfaaebebbc269110355/test/spec/Feature/Query/RpcSpec.hs#L213

> but it's also not always an improvement. Not sure what to do.

Yeah, but at least being consistent is a bit better.

PR #3382, opened 2024-04-07
fix: remove rejected mediatype application/vnd.pgrst.object+json from…

### LINKED ISSUE #3373: Rejected `application/vnd.pgrst.object+json` returns the media type sucessfully
## Problem

```
$ postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run
```
Doing:

```
curl 'localhost:3000/projects' -H "Accept:  application/vnd.pgrst.object+json" -i
```

Returns:

```
HTTP/1.1 406 Not Acceptable
Content-Type: application/vnd.pgrst.object+json; charset=utf-8

{"code":"PGRST116","details":"The result contains 5 rows","hint":null,"message":"JSON object requested, multiple (or no) rows returned"}
```

It contains the rejected media type and there's even an additional `Content-Type`.

## Solution

Remove `application/vnd.pgrst.object+json` from the response. All errors should come in `application/json`.

taimoorzaeem:
Fixes #3373.

### DIFF
--- a/test/spec/SpecHelper.hs
+++ b/test/spec/SpecHelper.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE FlexibleContexts #-}
 module SpecHelper where
 
 import           Control.Lens           ((^?))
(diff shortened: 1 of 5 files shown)


test/spec/SpecHelper.hs
```diff
@@ -36,24 +37,42 @@ import PostgREST.SchemaCache.Identifiers (QualifiedIdentifier (..))
...
     Just hdr -> if val `BS.isInfixOf` hdr then Nothing else Just $ "missing header value: " <> toS val <> "\n"
     Nothing  -> Just $ "missing header: " <> toS (original name) <> "\n"
-
+{-
+matchUnitHeaderPresent :: HeaderName -> MatchHeader
+matchUnitHeaderPresent name = MatchHeader $ \headers _ ->
+  case filter (\(headerName,_) -> headerName == name) headers of
+    [_]   -> Nothing
+    _     -> Just "unexpected header: zero or multiple headers present\n"
+-}
```
steve-chavez, line 75: Hm, delete this commented code?

taimoorzaeem: Oops, sure.

PR #3411, opened 2024-04-14
refactor: use LogLevel in Logger 

steve-chavez:


src/PostgREST/AppState.hs
```diff
@@ -193,17 +203,16 @@ initPool AppConfig{..} =
     (toUtf8 $ addFallbackAppName prettyVersion configDbUri)
 
 -- | Run an action with a database connection.
-usePool :: AppState -> AppConfig -> SQL.Session a -> IO (Either SQL.UsageError a)
-usePool AppState{..} AppConfig{configLogLevel, configObserver=observer} sess = do
+usePool :: AppState -> SQL.Session a -> IO (Either SQL.UsageError a)
+usePool AppState{stateObserver=observer,..} sess = do
   res <- SQL.use statePool sess
 
-  when (configLogLevel > LogCrit) $ do
```
steve-chavez, line 200: Log level logic used to be spread on AppState. Now is only on `Logger.hs`.

PR #3229, opened 2024-02-15
feat: log connection pool events on log-level=info

steve-chavez:
For https://github.com/PostgREST/postgrest/issues/3214.

The logs look like this (only for `log-level=info`):

```
14/Apr/2024:22:03:37 -0500: Starting PostgREST 12.1 (36ba330)...
14/Apr/2024:22:03:37 -0500: Attempting to connect to the database...
14/Apr/2024:22:03:37 -0500: Connection b63f8c6d-1423-4f7e-bd81-73ff25244033 is being established
14/Apr/2024:22:03:37 -0500: Connection b63f8c6d-1423-4f7e-bd81-73ff25244033 is available

...
14/Apr/2024:22:05:43 -0500: Connection b63f8c6d-1423-4f7e-bd81-73ff25244033 is terminated due to max idletime

...
14/Apr/2024:22:05:43 -0500: Connection f0813512-f4e6-4795-b156-ba140d2bc602 is being established
14/Apr/2024:22:05:43 -0500: Connection f0813512-f4e6-4795-b156-ba140d2bc602 is terminated due to network error

...
14/Apr/2024:21:47:44 -0500: Connection 63b40544-5a6d-4ec8-890a-c9c68f181e70 is terminated max lifetime

...
14/Apr/2024:21:50:33 -0500: Connection ad9b94c1-c6d7-4faa-bcc6-5edfb53d1119 is terminated due to release
```

Based on the new https://github.com/nikita-volkov/hasql-pool/pull/40

Issue #3214, opened 2024-02-07
Empty error message, `db-pool` size surpassed 


steve-chavez:
### Environment

* PostgreSQL version: PostgreSQL 14.1 
* PostgREST version: 10.2.0
* Operating system: Ubuntu

## Problem

A running instance with a `db-pool = 60` reports the following problems.

1. PostgREST replies with an empty error message.

```
$ curl localhost:3000/x  -i

400 Bad Request
{ code: '', details: null, hint: null, message: '' }
```

A subsequent request succeeds since PostgREST recovers though.

2. The authenticator role surpases db-pool connections.

Here it can be seen it reaches 71 connections.

```sql
select usename as rolname, count(*) as total_connections, sum(case when state = 'active' then 1 else 0 end) as active_connections, sum(case when state = 'idle' then 1 else 0 end) as idle_connections from pg_stat_activity group by usename;


     rolname       | total_connections | active_connections | idle_connections
---------------------+-------------------+--------------------+------------------
authenticator       |                71 |                  0 |               71
```

At some other time it reports 130 connections and surpasses pg `max_connections`:

```
01/Feb/2024:17:43:04 +0000: {"code":"PGRST000","details":"FATAL:  remaining connection slots are reserved for non-replication superuser connections\n","hint":null,"message":"Database connection error. Retrying the connection."}
```

I wasn't able to find the root cause of this. This instance in particular just went through a migration from ipv4 to ipv6 only.

However it's clear that the hasql pool is not recycling the pool connections. An empty error message shouldn't happen too.

## Proposal

We need to increase the pool log traces. Log whenever the hasql pool reaper works ([ref](https://github.com/nikita-volkov/hasql-pool/blob/master/library/Hasql/Pool.hs#L102-L117)).

## Workarounds

A connection limit can be set on the authenticator role so the `db-pool` max is guaranteed.

```sql
alter role authenticator connection limit <db-pool + 1 for listen channel>
```

An idle timeout can be set in case the pool doesn't recycle idle connections.

```sql
alter role authenticator set idle_session_timeout to 1800000;
```

## Related

- https://github.com/PostgREST/postgrest/pull/2129
- https://github.com/PostgREST/postgrest/issues/1526

steve-chavez:
Update on https://github.com/nikita-volkov/hasql-pool/issues/43, we now have https://github.com/nikita-volkov/hasql-pool/pull/45.

So with those new observations, we should be able to add two more metrics:

[...]

PR #3412, opened 2024-04-15
test: use postgrest.read_stdout in io tests

steve-chavez:
It's easier to maintain this way in case there are new log lines.

Related to https://github.com/PostgREST/postgrest/pull/3229#issuecomment-2054760951

### DIFF
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -772,16 +772,8 @@ def test_admin_works_with_host_special_values(specialhostvalue, defaultenv):
         assert response.status_code == 200
 
 
-@pytest.mark.parametrize(
-    "level, has_output",
-    [
-        ("info", [True, True, True]),
-        ("warn", [False, True, True]),
-        ("error", [False, False, True]),
-        ("crit", [False, False, False]),
-    ],
-)
-def test_log_level(level, has_output, defaultenv):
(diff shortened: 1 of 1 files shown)


steve-chavez:
> This takes away from the previous test, because the log format itself is not tested anymore, right?

Just added back the previous regexes. I thought the user agent and roles on the log lines were irrelevant for this test but I'm cool with adding them back.

steve-chavez:
> How about just doing snapshot tests for each log level, and thus saving the full log output for each? This way we'd have a much better picture of the log level differences.

I'd rather not invest more time on this for now. This was just a step for https://github.com/PostgREST/postgrest/pull/3229, which I need for clearing an unknown pool issue https://github.com/PostgREST/postgrest/issues/3214.

steve-chavez:
This is still an improvement over the current tests, I think. Will merge.

steve-chavez:
> How about just doing snapshot tests for each log level, and thus saving the full log output for each? This way we'd have a much better picture of the log level differences.

Wanted to open an issue for this, but giving it more thought.. maybe snapshot testing is not adequate for logs because the lines order is not deterministic. The snapshots could change frequently and we'd have to update them everytime (more maintenance).

Currently we do `sorted()` on the log lines to make the output deterministic, which has been working fine.

PR #3416, opened 2024-04-17
feat: add db-channel-uri config to allow reload using NOTIFY for read replicas

### LINKED ISSUE #3414: Event triggers don't work on read replicas
## Problem

See https://stackoverflow.com/questions/29931136/postgresql-replica-where-i-can-create-triggers/29932655#29932655.

This means that PostgREST [NOTIFY reloading](https://postgrest.org/en/stable/schema_cache.html#reloading-with-notify) won’t work in read replicas.

Previously, I thought this was only a problem with just LISTEN/NOTIFY on read replicas (https://github.com/PostgREST/postgrest/issues/2781), but it's worse.

## Solution

Enable a configurable connection string for the LISTEN channel.

```
db-channel-uri = "postgres://authenticator:mysecretpassword@localhost:5433/postgres?parameters=val"
```

Similar to the https://postgrest.org/en/v12/references/configuration.html#db-uri. By default it will use the same value as the `db-uri`.

This way the LISTEN channel can point to the primary instead of the replica and notify reloading can work normally.

---

- This should also close https://github.com/PostgREST/postgrest/issues/2781
- Testing is not possible right now because we lack the tooling. See https://github.com/PostgREST/postgrest/issues/1766). We can note it as a TODO in that issue.

laurenceisla:
Should close #3414. 

Applies to the cases where a PostgREST instance connects to a Read Replica. Then, the `db-channel-uri` config can be used to set a connection to the Primary DB, allowing PostgREST to reload when an Event Trigger is executed in that Primary DB.


src/PostgREST/AppState.hs
```diff
@@ -515,6 +516,7 @@ listener appState@AppState{stateObserver=observer} conf@AppConfig{..} = do
       putIsListenerOn appState False
       -- assume the pool connection was also lost, call the connection worker
       connectionWorker appState
+      -- TODO: debounce
       -- retry the listener
       listener appState conf
```
laurenceisla, line 521: This needs a debounce for when the connection is lost or the `db-channel-uri` has an error.

Since this should apply only when `db-channel-uri` is set, perhaps it should run conditionally, like:

```suggestion
      when (isNothing dbChannel) $
        connectionWorker appState
      -- TODO: debounce
      -- retry the listener
      listener appState conf
```

steve-chavez:
Let's try Wolfgang's approach.

PR #3419, opened 2024-04-18
fix: OpenAPI now tags a FK correctly on O2O relationships

### LINKED ISSUE #3418: OpenAPI does not tag a column as FK if it also has a UNIQUE constraint
### Environment

* PostgreSQL version: 15.6
* PostgREST version: 12.0.2
* Operating system: NixOS

### Description of issue

When a Foreign Key column is detected, OpenAPI adds the description "Note: This is a Foreign Key..." to it. But if that column also has a UNIQUE constraint, then it does not show that description anymore, e.g.:

This column shows the message:
```sql
todo_id int references api.todos(id)
```

While this doesn't:
```sql
todo_id int references api.todos(id) unique
```

laurenceisla:
Closes #3418

### DIFF
--- a/src/PostgREST/Response/OpenAPI.hs
+++ b/src/PostgREST/Response/OpenAPI.hs
@@ -115,6 +115,7 @@ makeProperty tbl rels col = (colName col, Inline s)
         -- Finds the relationship that has a single column foreign key
         rel = find (\case
           Relationship{relCardinality=(M2O _ relColumns)} -> [colName col] == (fst <$> relColumns)
+          Relationship{relCardinality=(O2O _ relColumns)} -> [colName col] == (fst <$> relColumns)
           _                                               -> False
           ) relsSortedByIsView
         fCol = (headMay . (\r -> snd <$> relColumns (relCardinality r)) =<< rel)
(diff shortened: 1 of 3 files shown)


steve-chavez (approved):

src/PostgREST/Response/OpenAPI.hs
```diff
@@ -115,6 +115,7 @@ makeProperty tbl rels col = (colName col, Inline s)
         -- Finds the relationship that has a single column foreign key
         rel = find (\case
           Relationship{relCardinality=(M2O _ relColumns)} -> [colName col] == (fst <$> relColumns)
+          Relationship{relCardinality=(O2O _ relColumns)} -> [colName col] == (fst <$> relColumns)
           _                                               -> False
```
steve-chavez, line 119: Should we remove this wildcard match and be more explicit?

softmarshmallow:
Hi. can we get a example description output when the fk is unique ?

Before:
```json
"description": "Note:\nThis is a Foreign Key to `foreign.id`.<fk table='foreign' column='id'/>",
```

After:
```
?
```

PR #3423, opened 2024-04-19
feat: PGRST121 error messages now show the failed MESSAGE or DETAIL

### LINKED ISSUE #3404: Misleading error PGRST121
### Environment

* PostgreSQL version: 16
* PostgREST version: 12.0.2

### Description of issue

Response on error is `The message and detail field of RAISE 'PGRST' error expects JSON` but it apparently requires a specific schema for that JSON, fx. the detail JSON must be an object with status and headers fields.

laurenceisla:
Closes #3404 .

### DIFF
--- a/test/spec/Feature/Query/RpcSpec.hs
+++ b/test/spec/Feature/Query/RpcSpec.hs
@@ -1452,17 +1452,29 @@ spec actualPgVersion =
           resHeaders `shouldSatisfy` elem ("X-Header", "str")
           resBody `shouldBe` [json|{"code":"123","message":"ABC","details":null,"hint":null}|]
 
-      it "returns error for invalid JSON in RAISE Message field" $
+      it "returns error for invalid JSON in the MESSAGE option of the RAISE statement" $
         get "/rpc/raise_sqlstate_invalid_json_message" `shouldRespondWith`
-          [json|{"code":"PGRST121","message":"The message and detail field of RAISE 'PGRST' error expects JSON","details":null,"hint":null}|]
+          [json|{
+            "code":"PGRST121",
+            "message":"Could not parse JSON in the \"RAISE SQLSTATE 'PGRST'\" error",
+            "details":"Invalid JSON value for MESSAGE: 'INVALID'",
+            "hint":null}|]
           { matchStatus = 500 }
 
-      it "returns error for invalid JSON in RAISE Details field" $
+      it "returns error for invalid JSON in the DETAIL option of the RAISE statement" $
         get "/rpc/raise_sqlstate_invalid_json_details" `shouldRespondWith`
-          [json|{"code":"PGRST121","message":"The message and detail field of RAISE 'PGRST' error expects JSON","details":null,"hint":null}|]
+          [json|{
+            "code":"PGRST121",
+            "message":"Could not parse JSON in the \"RAISE SQLSTATE 'PGRST'\" error",
+            "details":"Invalid JSON value for DETAIL: 'INVALID'",
+            "hint":null}|]
           { matchStatus = 500 }
 
-      it "returns error for missing Details field in RAISE" $
+      it "returns error for missing DETAIL option in the RAISE statement" $
         get "/rpc/raise_sqlstate_missing_details" `shouldRespondWith`
-          [json|{"code":"PGRST121","message":"The message and detail field of RAISE 'PGRST' error expects JSON","details":null,"hint":null}|]
+          [json|{
+            "code":"PGRST121",
+            "message":"Could not parse JSON in the \"RAISE SQLSTATE 'PGRST'\" error",
+            "details":"DETAIL is missing in the RAISE statement",
+            "hint":null}|]
           { matchStatus = 500 }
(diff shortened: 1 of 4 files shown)


test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1452,17 +1452,29 @@ spec actualPgVersion =
...
 
-      it "returns error for invalid JSON in RAISE Message field" $
+      it "returns error for invalid JSON in the MESSAGE option of the RAISE statement" $
         get "/rpc/raise_sqlstate_invalid_json_message" `shouldRespondWith`
-          [json|{"code":"PGRST121","message":"The message and detail field of RAISE 'PGRST' error expects JSON","details":null,"hint":null}|]
+          [json|{
+            "code":"PGRST121",
+            "message":"Could not parse JSON in the \"RAISE SQLSTATE 'PGRST'\" error",
+            "details":"Invalid JSON value for MESSAGE: 'INVALID'",
+            "hint":null}|]
```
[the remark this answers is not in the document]
laurenceisla: Good idea, will add them.

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1452,17 +1452,29 @@ spec actualPgVersion =
...
 
-      it "returns error for invalid JSON in RAISE Details field" $
+      it "returns error for invalid JSON in the DETAIL option of the RAISE statement" $
         get "/rpc/raise_sqlstate_invalid_json_details" `shouldRespondWith`
-          [json|{"code":"PGRST121","message":"The message and detail field of RAISE 'PGRST' error expects JSON","details":null,"hint":null}|]
+          [json|{
+            "code":"PGRST121",
+            "message":"Could not parse JSON in the \"RAISE SQLSTATE 'PGRST'\" error",
+            "details":"Invalid JSON value for DETAIL: 'INVALID'",
+            "hint":"DETAIL must be a JSON object with obligatory keys: 'status', 'headers' and optional key: 'status_text'."}|]
```
steve-chavez, line 1470: Some thoughts. This message is for the developer, not for the API user. We do this in other cases too IIRC, but it doesn't seem correct.

- We do it this way because it's the most direct way of feedback. Raising a log error is not that visible.
- Is it really a 500 if it's a developer error? Shouldn't it be a 400? :thinking: 
- Since, we have 3 types of users interacting with PostgREST ([ref](https://postgrest.org/en/latest/explanations/architecture.html#bird-s-eye-view)). I wonder if we should show this error only in a case where postgREST is in development mode (say on `log-level=debug`), then here the status would be 4xx. Then for other cases we reply with a generic error and the status should be 500.

@wolfgangwalther @laurenceisla WDYT?

Note: This doesn't need to be solved in this PR.

steve-chavez: One motivation I have for 4xx errors in this cases. When considering Observability, 5xx errors can be seen as PostgREST failures but in this case it really isn't. So if we later expose a metric for failures (5xx errors), it would be flawed.

steve-chavez: Hm, if we would have a `599 Operator Error` status code, this would be easy to solve. Metrics could just exclude this status and the error message could add something like "hint: contact your API administrator".

Discussion about custom status codes:
https://stackoverflow.com/questions/7996569/can-we-create-custom-http-status-codes

Besides that, I'm out of ideas.

steve-chavez (approved):

test/spec/Feature/Query/RpcSpec.hs
steve-chavez: Ah, that's a great idea. In practice it shouldn't matter much since that status code should not hit production. Will open an issue.

test/spec/Feature/Query/RpcSpec.hs
```diff
@@ -1452,17 +1452,29 @@ spec actualPgVersion =
...
-      it "returns error for invalid JSON in RAISE Message field" $
+      it "returns error for invalid JSON in the MESSAGE option of the RAISE statement" $
         get "/rpc/raise_sqlstate_invalid_json_message" `shouldRespondWith`
-          [json|{"code":"PGRST121","message":"The message and detail field of RAISE 'PGRST' error expects JSON","details":null,"hint":null}|]
+          [json|{
+            "code":"PGRST121",
+            "message":"Could not parse JSON in the \"RAISE SQLSTATE 'PGRST'\" error",
+            "details":"Invalid JSON value for MESSAGE: 'INVALID'",
+            "hint":"MESSAGE must be a JSON object with obligatory keys: 'code', 'message' and optional keys: 'details', 'hint'."}|]
           { matchStatus = 500 }
```
steve-chavez, line 1462: @laurenceisla Btw, the docs show [PGRST121](https://postgrest.org/en/latest/references/errors.html#pgrst121) as a 400

laurenceisla: Thanks for the heads up, fixed in https://github.com/PostgREST/postgrest/commit/85a4e35a2de869ddf9b6c5f6893a3154767cbc76.

Issue #2814, opened 2023-06-08
Merge postgrest-docs with main repo


steve-chavez:
## Problem

Keeping postgrest-docs in sync with the main repo is quite painful. Duplicate work is done in tagging a release on main, duplicating the CHANGELOG on docs, etc. This makes us slower overall.

## Solution

Merge postgrest-docs with main repo while preserving history. It's possible as previously discussed on: https://github.com/PostgREST/postgrest/pull/1897#issuecomment-881964560

We still need to figure out the steps for CI.

steve-chavez:
If we only make releases through non-main branches, like we have been doing for some patch/minor versions:

- https://github.com/PostgREST/postgrest/tree/rel-10.1.2
- https://github.com/PostgREST/postgrest/tree/rel-10.2.0
- https://github.com/PostgREST/postgrest/tree/rel-11.2.1

Then we would be compatible with the way we release docs too, which is done exclusively on branches: https://github.com/PostgREST/postgrest-docs/branches

Then our main here would point to the latest docs, which is also done for postgrest-docs.

steve-chavez:
Since CI for docs is fast, we could run it on every PR. 

We would have to drop some old version of the docs though, which don't have equivalent branch releases here.

I think this would be the smoother transition.

steve-chavez:
On  https://github.com/PostgREST/postgrest/pull/1897#issuecomment-881957275

> I think we could still maintain the separate repo for tracking issues, even if we move the docs into the code repo
> Benefits
> Docs-related issues/PRs separated from other ones

We agreed to keep the docs issues on its own repo. I think we should keep that for now.

laurenceisla:
Yes, I think moving them to core is the best if archiving the docs repo gives problems.

> It would make the most sense to move them here, but we can't do that 1-by1...

Someone mentioned that bulk issue migration can be done using GitHub CLI (it doesn't keep the labels though): 

https://github.com/orgs/community/discussions/9695#discussioncomment-5423988

steve-chavez:
> So we need to rename our release branches. The obvious choice is rel-X.Y -> vX.Y.
> maybe we should just use the major only and nothing more?

Agree with both :+1:

steve-chavez:
We'll need a redirect on RTD too. Links like https://postgrest.org/en/v12.0/references/api/aggregate_functions.html are not working. We used those on the release page https://github.com/PostgREST/postgrest/releases/tag/v12.0.0.

[...]

PR #3443, opened 2024-04-24
Testing patch v12.0.3

laurenceisla:
Just testing some cherry-picks with minimal conflicts for a patch release. 

The original strategy is to force push the changes in `main` to the `v12` branch. Since only some commits are applied, I based it on the existing `v12` branch instead.


steve-chavez:
Even with just some fixes, a new patch release LGTM. If the others are hard to merge, we can leave them for the next minor.

laurenceisla:
> I don't understand what that means. You are proposing to merge this into v12, which was exactly the plan, right? What am I missing?

No, no, there shouldn't be a problem. Mostly mentioned it because I didn't know if the release workflow may be different between force pushing the `main` branch and forking and merging the `v12` one.

laurenceisla:
> I'm not sure what you mean by force pushing the main branch - but I don't ever want to do that. Maybe there is a mis-understanding.

Ah cool, I get it now. I misunderstood here: https://github.com/PostgREST/postgrest/issues/2814#issuecomment-1950237197, I now notice that you never mentioned the `main` branch. 

> But then do the cherry-picking + PR semi-automatically and also auto-release bugfix releases after commits to the back-branches, similar to nightly releases on the main branch.

Got it.

laurenceisla:
> But GHC_VERSION is not $GHC_VERSION - and thus ghcup was always throwing errors when trying to install that literal GHC_VERSION string.

Ooph, nice catch.

> If you happen to have a chance to look into this before I do again tomorrow, you can also restart the failed jobs in this pipeline: https://github.com/PostgREST/postgrest/actions/runs/9022807863/job/24793212610 - this should then continue the release pipeline. I'll look into it tomorrow again.

That's strange, the server doesn't seem to have older build folders there... maybe with the GHC fix, it added more libraries to `.cabal`. I'll check it out and will restart when I find the culprit.

laurenceisla:
It was GHC itself that had GBs of tmp files while installing, I think both CI jobs for `main` and `v12.0.3` collided in installing the new GHC version (not exactly sure). Anyways, I nuked the whole ghcup/cabal and reinstalled again. I'm also restarting the jobs with the one in `main` first.

laurenceisla:
All done! it's working correctly now :tada:

Confirmed that both of the issues mentioned here are fixed https://github.com/PostgREST/postgrest/pull/3443#issuecomment-2103263050.

PR #3460, opened 2024-04-26
ci: DRY the cleanup job for ARM servers

laurenceisla:
DRYed the job according to the [GitHub Actions docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds):

> If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue. [...] If you would like a job to run even if a job it is dependent on did not succeed, use the always() conditional expression in jobs.<job_id>.if.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,18 +90,6 @@ jobs:
           fingerprint: ${{ secrets.SSH_ARM_FINGERPRINT }}
           source: "${{ steps.Remote-Dir.outputs.remotepath }}/result.tar.xz"
           target: "result.tar.xz"
-      - name: Remove uploaded files from server
-        if: ${{ always() && !startsWith(github.ref, 'refs/tags/v') }}
-        uses: appleboy/ssh-action@master
-        env:
-          REMOTE_DIR: ${{ steps.Remote-Dir.outputs.remotepath }}
-        with:
-          host: ${{ secrets.SSH_ARM_HOST }}
-          username: ubuntu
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -315,7 +303,7 @@ jobs:
     needs:
       - arm
       - docker-arm
-    if: ${{ always() && startsWith(github.ref, 'refs/tags/v') }}
+    if: ${{ always() && vars.SSH_ARM_ENABLED) }}
```
laurenceisla, line 306: If I understood correctly, `always()` allows to run this job even after `arm` or `docker-arm` succeeds, fails or are cancelled/skipped. So it only needs to add a condition to run when at least the `arm` job runs, which is the `SSH_ARM_ENABLED` variable.

.github/workflows/ci.yaml
```diff
@@ -90,18 +90,6 @@ jobs:
...
-        uses: appleboy/ssh-action@master
-        env:
-          REMOTE_DIR: ${{ steps.Remote-Dir.outputs.remotepath }}
-        with:
-          host: ${{ secrets.SSH_ARM_HOST }}
-          username: ubuntu
-          key: ${{ secrets.SSH_ARM_PRIVATE_KEY }}
-          fingerprint: ${{ secrets.SSH_ARM_FINGERPRINT }}
-          envs: REMOTE_DIR
-          script: rm -rf $REMOTE_DIR
```
steve-chavez, line 104: Less code is always good! :raised_hands:

PR #3462, opened 2024-04-30
feat: force listener to work on primary cluster 

### LINKED ISSUE #3414: Event triggers don't work on read replicas

steve-chavez:
Closes #3414

### DIFF
--- a/test/io/test_replica.py
+++ b/test/io/test_replica.py
@@ -20,12 +20,7 @@ def test_sanity_replica(replicaenv):
         response = postgrest.session.get("/items?select=count")
         assert response.text == '[{"count":10}]'
 
-    working_replica_env = {
-        **replicaenv["replica"],
-        "PGRST_DB_CHANNEL_ENABLED": "false",  # LISTEN doesn't work on read replicas
-    }
-
-    with run(env=working_replica_env) as postgrest:
+    with run(env=replicaenv["replica"]) as postgrest:
         response = postgrest.session.get("/rpc/is_replica")
         assert response.text == "true"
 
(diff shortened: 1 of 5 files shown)


steve-chavez:
New problem with our way to modify the connection string, comma separated hosts:

```
"postgresql://host1:123,host2:456/somedb"
```

Are not valid URIs for [parseURI](https://hackage.haskell.org/package/network-uri-2.6.4.2/docs/Network-URI.html#v:parseURI)

steve-chavez:
New problem on https://github.com/PostgREST/postgrest/pull/3533, when the LISTEN fails it does so silently.

test/io/test_replica.py
```diff
@@ -20,12 +20,7 @@ def test_sanity_replica(replicaenv):
         response = postgrest.session.get("/items?select=count")
         assert response.text == '[{"count":10}]'
 
-    working_replica_env = {
-        **replicaenv["replica"],
-        "PGRST_DB_CHANNEL_ENABLED": "false",  # LISTEN doesn't work on read replicas
-    }
-
-    with run(env=working_replica_env) as postgrest:
+    with run(env=replicaenv["replica"]) as postgrest:
```
steve-chavez, line 23: LISTEN now works and this test proves it

PR #3468, opened 2024-05-01
feat: add metrics for used and created connections

### LINKED ISSUE #3214: Empty error message, `db-pool` size surpassed

laurenceisla:
WIP

Closes #3214

### DIFF
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -57,21 +57,27 @@ let
           }
           { });
 
-      hasql-pool = lib.dontCheck prev.hasql-pool_1_0_1;
+      hasql-pool = lib.dontCheck (prev.callHackageDirect
+        {
+          pkg = "hasql-pool";
+          ver = "1.1";
+          sha256 = "sha256-TJZdUwsBo4pLLYydc8FgxYFI37Tj5K+E7idf4fQYEjU=";
+        }
+        { }
+      );
 
       postgresql-libpq = lib.dontCheck
         (prev.postgresql-libpq.override {
           postgresql = super.libpq;
         });
 
-      hasql-notifications = lib.dontCheck (prev.callHackageDirect
-        {
-          pkg = "hasql-notifications";
-          ver = "0.2.1.1";
-          sha256 = "sha256-oPhKA/pSQGJvgQyhsi7CVr9iDT7uWpKUz0iJfXsaxXo=";
-        }
-        { }
-      );
+      hasql-notifications = lib.dontCheck
+        (prev.callCabal2nixWithOptions "hasql-notifications" (super.fetchFromGitHub {
+          owner = "laurenceisla";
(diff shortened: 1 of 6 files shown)


steve-chavez:
New https://hackage.haskell.org/package/hasql-notifications now allows the new hasql pool

steve-chavez:
@laurenceisla What was the problem for this? IIRC there was a mismatch between the new hasql-pool and hasql-notifications? Can you point out the exact versions? Maybe I can raise this on hasql-notifications.

laurenceisla:
The libraries `hasql-dynamic-statements` and `hasql-transaction` didn't have versions supporting `hasql` >= 1.7. But I think that's fixed now upstream, I'll try relaxing the version limit for those libraries and see how it goes.

wolfgangwalther:
This should now have all the required dependencies already on main, I think.

PR #2985, opened 2023-10-02
Fix regression that rejects URI connection strings with certain unescaped characters in the password

laurenceisla:
Closes #2970. Continues what's specified in https://github.com/PostgREST/postgrest/pull/2977#issuecomment-1743836200

PR #3480, opened 2024-05-03
fix: not adding application_name on all URIs

steve-chavez:
Lifts the limitation added on https://github.com/PostgREST/postgrest/pull/2985.

Related https://github.com/PostgREST/postgrest/pull/3462#issuecomment-2089414110


src/PostgREST/Config.hs
```diff
@@ -502,19 +501,37 @@ readPGRSTEnvironment =
...
+  case pgConnString dbUri of
     Nothing  -> mempty
-    Just ""  -> "?" <> uriFmt
-    Just "?" -> uriFmt
-    _        -> "&" <> uriFmt
+    Just PGKeyVal -> " " <> keyValFmt
+    Just PGURI    -> case T.breakOn "?" dbUri of
+      (_, "")  -> "?" <> uriFmt
+      (_, "?") -> uriFmt
+      (_, _)   -> "&" <> uriFmt
```
laurenceisla, line 519: On PostgREST 12.0.2 an URI without a percent encoded `?` works OK:

```
db-uri = "postgres://authenticator:pass?special@localhost:5432/postgres"
```

But, unless it's percent encoded, it gives this error with this change:

```
FATAL:  database \"postgres&fallback_application_name=PostgREST 12.1 (557d008)\" does not exist\n"
```

I think it still needs to check that the `breakOn` is done after the first appearance of `/`?  (after the `uriDesignator` of course)

steve-chavez:
Argh, io tests got stuck again https://github.com/PostgREST/postgrest/actions/runs/8943572333/job/24568588561?pr=3480.

Seems https://github.com/PostgREST/postgrest/pull/3465 wasn't enough?

src/PostgREST/Config.hs
steve-chavez: Good catch. Fixed that.

steve-chavez: Turns out that logic wasn't enough in cases like this:

```
"postgresql://?dbname=postgres&host=/run/user/1000/postgrest/postgrest-with-postgresql-16-BuR/socket&user=some_protected_user&password=invalid_pass"
```

That's currently mangled to:
```
"postgresql://?dbname=postgres&host=/run/user/1000/postgrest/postgrest-with-postgresql-16-BuR/socket&user=some_protected_user&password=invalid_pass?fallback_application_name=PostgREST%201
1.1.0%20%285a04ec7%29"
```

PR #3481, opened 2024-05-03
feat: add more config examples to CLI

### LINKED ISSUE #3248: Add CLI page

taimoorzaeem:
Closes #3248.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3184, Log full pg version to stderr on connection - @steve-chavez
  - #3242. Add config `db-hoisted-tx-settings` to apply only hoisted function settings - @taimoorzaeem
  - #3214, #3229 Log connection pool events on log-level=debug - @steve-chavez, @laurenceisla
+ - #3248, Add more config examples to CLI - @taimoorzaeem
+   + Now accepts ``--example-file``, ``--example-db``, and ``--example-env``
 
 ### Fixed
 
@@ -41,6 +43,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3414, Force listener to connect to read-write instances using `target_session_attrs` - @steve-chavez
  - #3255, Fix incorrect `413 Request Entity Too Large` on pg errors `54*` - @taimoorzaeem
  - #3549, Remove verbosity from error logs starting with "An error occurred..." and replacing it with "Failed to..." - @laurenceisla
+
+### Changed
+ - #3248, Modified some CLI options - @taimoorzaeem
+   + Replace ``-e`` and ``--example`` with ``--example-file``
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
  - #3248, Modified some CLI options - @taimoorzaeem
    + Replace ``-e`` and ``--example`` with ``--example-file``
+   + Rename option ``--dump-schema`` to ``--dump-schema-cache``
 
 ### Deprecated
 
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -73,6 +73,8 @@ data Command
   | CmdDumpConfig
   | CmdDumpSchema
 
+data Example = ExampleFile | ExampleDb | ExampleEnv
+
 -- | Read command line interface options. Also prints help.
 readCLIShowHelp :: IO CLI
 readCLIShowHelp =
@@ -94,11 +96,22 @@ readCLIShowHelp =
         <> O.short 'v'
         <> O.help "Show the version information"
 
-    exampleParser =
-      O.infoOption exampleConfigFile $
-        O.long "example"
-        <> O.short 'e'
-        <> O.help "Show an example configuration file"
+    exampleParser = exampleParserFile <|> exampleParserDb <|> exampleParserEnv
+
+    exampleParserFile =
+      O.infoOption (exampleConfig ExampleFile) $
+        O.long "example-file"
+        <> O.help "Show an example of a configuration file"
+
+    exampleParserDb =
+      O.infoOption (exampleConfig ExampleDb) $
+        O.long "example-db"
+        <> O.help "Show an example of in-database configuration"
+
+    exampleParserEnv =
+      O.infoOption (exampleConfig ExampleEnv) $
+        O.long "example-env"
+        <> O.help "Show an example of environment variables configuration"
 
     cliParser :: O.Parser CLI
     cliParser =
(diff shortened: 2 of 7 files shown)


taimoorzaeem:
@wolfgangwalther Hmm, I have noticed an inconsistency between `docs` and `dbSettingsNames`. The `dbSettingsNames` include `db_tx_end` but the docs configuration [page](https://postgrest.org/en/v12/references/configuration.html#db-tx-end) says that `db_tx_end` is `n/a` for in-db config. More the `dbSettingsNames` include `raw_media_types` but there is no detail available around this config in the documentation.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Shows the correct JSON format in the `hints` field
  - #3340, Log when the LISTEN channel gets a notification - @steve-chavez
  - #3184, Log full pg version to stderr on connection - @steve-chavez
+ - #3248, Add more config examples to CLI - @taimoorzaeem
+   + Now accepts ``--example-file``, ``--example-db``, and ``--example-env``
+   + Replace ``-e`` and ``--example`` with ``--example-file``
+   + Rename option ``--dump-schema`` to ``--dump-schema-cache``
```
laurenceisla, line 29: Since they're kinda breaking changes, these should go under a separate subtitle (after the `### Fixed` entries), something like:

```markdown
### Changes
 - #3248, Modified some CLI options - @taimoorzaeem
   + Replace ``-e`` and ``--example`` with ``--example-file``
   + Rename option ``--dump-schema`` to ``--dump-schema-cache``
```

src/PostgREST/CLI.hs
```diff
@@ -171,6 +187,9 @@ exampleConfigFile =
       |## When disabled, statements will be parametrized but won't be prepared.
       |db-prepared-statements = true
       |
+      |## Function to override OpenAPI response
+      |db-root-spec = ""
```
laurenceisla, line 191:
```suggestion
      |# db-root-spec = "root"
```

src/PostgREST/CLI.hs
```diff
@@ -219,6 +241,9 @@ exampleConfigFile =
       |server-host = "!4"
       |server-port = 3000
       |
+      |## Trace HTTP request header
+      |server-trace-header = ""
```
laurenceisla, line 245:
```suggestion
      |# server-trace-header = "X-Request-Id"
```

src/PostgREST/CLI.hs
```diff
@@ -230,3 +255,203 @@ exampleConfigFile =
...
+      |
+      |-- Stored proc to exec immediately after auth
+      |-- ALTER ROLE authenticator SET pgrst.db_pre_request = 'stored_proc_name';
+      |
+      |-- Enable or disable prepared statements. disabling is only necessary when behind a connection pooler.
+      |-- When disabled, statements will be parametrized but won't be prepared.
+      |ALTER ROLE authenticator SET pgrst.db_prepared_statements = 'true';
+      |
+      |-- Function to override the OpenAPI response
+      |ALTER ROLE authenticator SET pgrst.db_root_spec = '';
```
laurenceisla, line 285:
```suggestion
      |-- ALTER ROLE authenticator SET pgrst.db_root_spec = 'root';
```

src/PostgREST/CLI.hs
```diff
@@ -230,3 +255,203 @@ exampleConfigFile =
...
+      |ALTER ROLE authenticator SET pgrst.openapi_security_active = 'false'
+      |
+      |-- Base url for the OpenAPI output
+      |ALTER ROLE authenticator SET pgrst.openapi_server_proxy_uri = '';
+      |
+      |-- Configurable CORS origins
+      |-- ALTER ROLE authenticator SET pgrst.server_cors_allowed_origins = '';
+      |
+      |-- Trace HTTP request header
+      |-- ALTER ROLE authenticator SET pgrst.server_trace_header = '';
```
laurenceisla, line 331:
```suggestion
      |-- ALTER ROLE authenticator SET pgrst.server_trace_header = 'X-Request-Id';
```

src/PostgREST/CLI.hs
```diff
@@ -230,3 +255,203 @@ exampleConfigFile =
...
+      |
+      |## Stored proc to exec immediately after auth
+      |# PGRST_DB_PRE_REQUEST='stored_proc_name'
+      |
+      |## Enable or disable prepared statements. disabling is only necessary when behind a connection pooler.
+      |## When disabled, statements will be parametrized but won't be prepared.
+      |PGRST_DB_PREPARED_STATEMENTS=true
+      |
+      |## Function to override the OpenAPI response
+      |PGRST_DB_ROOT_SPEC=''
```
laurenceisla, line 390:
```suggestion
      |# PGRST_DB_ROOT_SPEC='root'
```

src/PostgREST/CLI.hs
```diff
@@ -230,3 +255,203 @@ exampleConfigFile =
...
+      |PGRST_OPENAPI_SERVER_PROXY_URI=''
+      |
+      |## Configurable CORS origins
+      |# PGRST_SERVER_CORS_ALLOWED_ORIGINS=''
+      |
+      |PGRST_SERVER_HOST=!4
+      |PGRST_SERVER_PORT=3000
+      |
+      |## Trace HTTP request header
+      |PGRST_SERVER_TRACE_HEADER=''
```
laurenceisla, line 445:
```suggestion
      |# PGRST_SERVER_TRACE_HEADER='X-Request-Id'
```

laurenceisla (commented):
These should be commented since they do not have a default value, and should also show a sample value (similar to how `db-anon-role` works). Only checked for the changes in the `--example-file`, maybe others are missing (verify in the docs: if they have "Default: n/a" then they should be commented).

CHANGELOG.md
steve-chavez: Is it possible to make these options non-breaking changes?

steve-chavez:
Hm, the newly added env vars and db options look hard to maintain. 

I'm thinking we should somehow auto-generate them.

CHANGELOG.md
laurenceisla: I initially proposed to not break the options in https://github.com/PostgREST/postgrest/pull/3321#pullrequestreview-1990458404, but we agreed to only use the new ones at the end of that review... if we'd like to not break the options, then we could go with that implementation.

laurenceisla:
> Hm, the newly added env vars and db options look hard to maintain.
> 
> I'm thinking we should somehow auto-generate them.

That's a good idea, we could use a sample configuration and the function used to detect valid in-db configs. Although I believe that it would be a bit tricky to implement (perhaps in another PR?).

> Also the in-db config format is not really future-proof as we're transitioning to a pre_config function: https://postgrest.org/en/v12/references/configuration.html#in-database-configuration

Hmm... then maybe printing a sample function instead of modifying a role would be the best here? or maybe both?

CHANGELOG.md
steve-chavez: Since another breaking change is planned on https://github.com/PostgREST/postgrest/issues/3501#issuecomment-2139656953, I think this should be fine.

taimoorzaeem:
Closing as no longer relevant.

Issue #3424, opened 2024-04-20
IO tests depend too heavily on timing / sleeping


wolfgangwalther:
I just ran the IO tests on a heavily overloaded system.. and had 32 tests failing, because of some timings related issues. Those pass just fine without load. This is annoying, because I can't continue working on PostgREST while I have other heavy tasks (rebuilding the world for nixpkgs...) running. This is also potentially an issue on much slower systems.

steve-chavez:
Maybe we can separate those tests (`test_io_timing.py`), so they're not always ran. Like what we do with `postgrest-test-memory`.

steve-chavez:
Hm, perhaps some of the pool timeout errors could be solved by using the new `/metrics` endpoint instead of checking the logs.

steve-chavez:
Additionally I think the acq timeout can be less than a second https://hackage.haskell.org/package/hasql-pool-1.1/docs/Hasql-Pool-Config.html#v:acquisitionTimeout.. but not sure if our config allows it now

steve-chavez:
I've also noted sometimes the io tests get stuck: https://github.com/PostgREST/postgrest/actions/runs/8857677646/job/24325423347?pr=3383

I had that happen just one time locally after https://github.com/PostgREST/postgrest/pull/3229. That should be solved by using metrics, I think.

steve-chavez:
io test got stuck again https://github.com/PostgREST/postgrest/actions/runs/8857677646/job/24325423347?pr=3383

We do some waiting when reading the logs https://github.com/PostgREST/postgrest/blob/7e5fd317e6b6060e31e7120837cec95ee4028869/test/io/postgrest.py#L60-L70

But it's capped at 1 second.. it's weird why it gets stuck

---

This looks to be the test that fails
https://github.com/PostgREST/postgrest/blob/7e5fd317e6b6060e31e7120837cec95ee4028869/test/io/test_io.py#L1361-L1389

steve-chavez:
@wolfgangwalther Great idea! I agree.

steve-chavez:
@wolfgangwalther Looking into this now. Admin API has another bug too (connection drop not detected), so will fix that first.

steve-chavez:
> - 200 when ready
> - 503 Service Unavailable while PostgREST is in the process of doing something: During connection to the database, during config reload, during schema cache reload.
> - 500 Internal Server Error as soon as one of the above fails.

[...]

PR #3497, opened 2024-05-09
Make IO tests more stable

wolfgangwalther:
This tries to address #3424 by preventing `test_admin_ready_includes_schema_cache_state` from creating 100s of connection retries per second and thus somehow making postgrest unresponsive. This does not fix the underlying problem, but should at least make the tests not cause random failures anymore.

One thing that is not clear to me is: Why are the IO tests only failing recently and not before? The test here has been in there for a long time already. It might be connected to the much extended logging, maybe the huge number of connection retries only becomes a problem due to the big number of logs generated that way.

### DIFF
--- a/src/PostgREST/SchemaCache.hs
+++ b/src/PostgREST/SchemaCache.hs
@@ -145,6 +145,9 @@ type SqlQuery = ByteString
 
 querySchemaCache :: AppConfig -> SQL.Transaction SchemaCache
 querySchemaCache AppConfig{..} = do
+  _       <-
+    let sleepCall = SQL.Statement "select pg_sleep($1 / 1000.0)" (param HE.int4) HD.noResult prepared in
+    whenJust configInternalSCSleep (`SQL.statement` sleepCall) -- only used for testing
   SQL.sql "set local schema ''" -- This voids the search path. The following queries need this for getting the fully qualified name(schema.name) of every db object
   pgVer   <- SQL.statement mempty $ pgVersionStatement prepared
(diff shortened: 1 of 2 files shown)


src/PostgREST/SchemaCache.hs
```diff
@@ -155,9 +158,6 @@ querySchemaCache AppConfig{..} = do
   reps    <- SQL.statement schemas $ dataRepresentations prepared
   mHdlers <- SQL.statement schemas $ mediaHandlers pgVer prepared
   tzones  <- SQL.statement mempty $ timezones prepared
-  _       <-
-    let sleepCall = SQL.Statement "select pg_sleep($1)" (param HE.int4) HD.noResult prepared in
-    whenJust configInternalSCSleep (`SQL.statement` sleepCall) -- only used for testing
```
steve-chavez, line 160: I remember I put this explicitly at the end to add a test for https://github.com/PostgREST/postgrest/pull/2810. Is it necessary to change it to the beginning?

I wonder if this test https://github.com/PostgREST/postgrest/blob/main/test/io/test_io.py#L1024-L1051 continues to work as intended.

If it doesn't, maybe I can add another sleep here.

---

I think there's a nice way to maintain these sleeps, in case we also need in other place of the code. We can keep using this "observation" pattern I've been doing for logs and metrics. We would just need an `observer $ Observation StartQuerySCache/EndQuerySCache` and do the sleeps in another module.

steve-chavez: I will make a note to do this change in another PR.

PR #3507, opened 2024-05-11
feat: add metrics=timings to prefer header

### LINKED ISSUE #3410: On-demand server-timing
## Problem

Setting [server-timing-enabled](https://postgrest.org/en/v12/references/configuration.html#server-timing-enabled) to true has  a ~6.5% RPS decrease. Using GETSimple on https://github.com/PostgREST/postgrest-benchmark:

no timings = 2176.4712194 rps
with timings = 2033.3584706 rps

I don't think there's a way around it since calculating the timings requires using clock time which needs a system call (`clock_gettime`).

## Solution

Allow getting server timing headers through a new preference header.

```http
Prefer: metrics=timings
```

`metrics=timings` would be the only value for now, but we could extend it later on.

This can later be enabled/disabled with https://github.com/PostgREST/postgrest/issues/2987.

While at it, deprecate [server-timing-enabled](https://postgrest.org/en/v12/references/configuration.html#server-timing-enabled) since it has a perf impact.

taimoorzaeem:
Closes #3410.

@steve-chavez What exactly do I need to do to deprecate the `server-timings-enabled` config? Should I just remove it from the codebase?


laurenceisla:
> What exactly do I need to do to deprecate the `server-timings-enabled` config? Should I just remove it from the codebase?

Deprecating the config does not mean to remove it from the code (right now), only that it's not recommended anymore (could be removed in the future). But it should be mentioned that it's deprecated in comments and in the docs in favor of the new feature. See, for example: https://github.com/PostgREST/postgrest/commit/6c3d7a946daef91014328bc50c0f63744fada8ad and https://postgrest.org/en/latest/references/api/preferences.html#single-json-object-as-function-parameter

src/PostgREST/App.hs
```diff
@@ -141,27 +144,29 @@ postgrestResponse appState conf@AppConfig{..} maybeSchemaCache pgVer authResult@
 
   body <- lift $ Wai.strictRequestBody req
 
-  let jwtTime = if configServerTimingEnabled then Auth.getJwtDur req else Nothing
+-- the preference metrics=timings cant be used before it is parsed, hence
+-- parseTime will be calculated for all requests
+  (parseTime, apiReq@ApiRequest{..}) <- withTiming True $ liftEither . mapLeft Error.ApiRequestError $ ApiRequest.userApiRequest conf req body sCache
```
laurenceisla, line 149: Ah, so it's not possible to disable ALL timings with `Prefer: metrics=timings`, without parsing the headers first. So it's not a perfect replacement of `server-timings-enabled = false`, since some timings will be calculated even if the header is not sent: the "jwt" and "parse" timings as far as I can see.

Not sure if it's OK to always calculate these timings, I believe it will still affect the performance as mentioned in the issue. cc: @steve-chavez

taimoorzaeem: @steve-chavez Any update on this?

steve-chavez: Hm, so here we're forcing to always calculate the time for parsing? That would defeat the whole purpose of https://github.com/PostgREST/postgrest/issues/3410

But I understand that we first need to parse to get the header. I guess we would need to do this in a step separate from `userApiRequest`, but that's messy.

steve-chavez: So far the `parse` and `response` timings ([ref](https://docs.postgrest.org/en/v12/references/observability.html#server-timing-header)) have not been that useful because they're always fast. 

Another option could be to omit the `parse` timing when doing `Prefer: metrics=timings` or we could just remove support for the above mentioned. Then we wouldn't have this problem.

steve-chavez: I was thinking to do a breaking change and remove the `parse` time now. But then I thought: wouldn't it be simpler to omit the `parse` time when `Prefer: metric=timings` is requested?

It would just be:

```
Server-Timing: jwt;dur=14.9, plan;dur=109.0, transaction;dur=353.2, response;dur=4.4
```

We could point that out in the docs.

@taimoorzaeem I believe that should unblock this feature.

taimoorzaeem: Yes, that would be much simpler. Let's implement it this way.

CHANGELOG.md
```diff
@@ -57,6 +58,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Return new `PGRST303` error when jwt claims decoding fails
  - #3906, Return `PGRST125` and `PGRST126` errors instead of empty json - @taimoorzaeem
 
+### Deprecated
+
+- #3410, The config `server-timing-enabled` is deprecated. Use `metrics=timings` prefer header instead - @taimoorzaeem
```
steve-chavez, line 63: Now that `server-timing-enabled` has different functionality than `Prefer: metrics=timings`, I don't think we can deprecate it/remove it in the future. At least for now.

docs/references/api/preferences.rst
```diff
@@ -264,3 +265,27 @@ To illustrate the use of this preference, consider the following scenario where
...
+
+  HTTP/1.1 200 OK
+
+  Server-Timing: jwt;dur=14.9, plan;dur=109.0, transaction;dur=353.2, response;dur=4.4
+
+See :ref:`server-timing_header` for details about the meaning of these timings.
+
+.. note::
+
+  When timings are requested with this header, we can't tell the ``parse`` stage timings. This is because we are past the "parse" stage when this preference itself is parsed.
```
steve-chavez, line 291: This didn't look good. So I thought of a simpler way: https://github.com/PostgREST/postgrest/pull/4057.

With this, we won't measure the parse time of the Preferences header, but that's fine as its parsing is simple and its input small.

The bulk of the `parse` stage is really in `QueryParams.parse` (URL query string parsing) which is still maintained.

steve-chavez: To be clear, once you rebase on top of #4057, the `Prefer: metric=timings` resposne now should contain the same information as when the `server-timing-enabled` config is set.

src/PostgREST/Auth.hs
```diff
@@ -163,6 +163,8 @@ middleware appState app req respond = do
       jwtCacheState = getJwtCacheState appState
 
 -- If ServerTimingEnabled -> calculate JWT validation time
+      -- Now that we also calculate the timings when Prefer: metrics=timings,
+      -- how do we get that preference here?
```
taimoorzaeem, line 167: @steve-chavez We need the preference here too. I am not sure about this but I think we should think about storing the `ApiRequest` result in `AppState`. This way it would be easier to access already parsed request information at any stage in the code.

steve-chavez: I think we should just refactor to remove the `Auth.middleware` approach. I've never liked that one as it hides side effects and obscures logic. It might allow us to remove the `unsafePerformIO` from Auth.hs too:

https://github.com/PostgREST/postgrest/blob/e4a2095b5a997e721299c24280ab785317b50fa0/src/PostgREST/Auth.hs#L190-L198

Once that's done you should be able to use/get the prefer value value before validating the JWT.

taimoorzaeem: Agreed. Let me refactor this.

PR #3512, opened 2024-05-13
fix: bind admin server to the same host

### LINKED ISSUE #3508: Server port and admin server port can end up with the same value
As found out in https://github.com/PostgREST/postgrest/issues/3424#issuecomment-2106000902.

Do this:

```
PGRST_SERVER_PORT=3000 PGRST_SERVER_HOST="localhost" PGRST_ADMIN_SERVER_PORT=3000 postgrest-with-postgresql-16 postgrest-run
```

and observe:

```
11/May/2024:21:30:27 +0200: Admin server listening on port 3000
11/May/2024:21:30:27 +0200: Listening on port 3000
```

One of them is listening on ipv4, the other on ipv6.

We should probably reject this case on the config level? Admin port should always be different from server port?

develop7:
Reuse main socket address to bind admin socket there too by resolving the address main socket is bound at and binding admin socket to it explicitly.

Introduces extra dependency to `iproute` for `HostAddress -> IP -> String` conversions, following the reference in https://www.stackage.org/haddock/lts-22.21/network-3.1.4.0/src/Network.Socket.Info.html#line-415

Fixes #3508

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -116,6 +116,7 @@ library
                     , http-types                >= 0.12.2 && < 0.13
                     , insert-ordered-containers >= 0.2.2 && < 0.3
                     , interpolatedstring-perl6  >= 1 && < 1.1
+                    , iproute                   >= 1.7.0 && < 1.8
                     , jose                      >= 0.8.5.1 && < 0.12
                     , lens                      >= 4.14 && < 5.3
                     , lens-aeson                >= 1.0.1 && < 1.3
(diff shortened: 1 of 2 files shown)


laurenceisla:
Since this will use the same host, then it would log this when the same port is used, right?

```
postgrest: Network.Socket.bind: resource busy (Address already in use)
```

It solves the issue of not allowing the same ports, but I think it's better/simpler to just not execute when `server-port` and `admin-server-port` are the same as it was mentioned in https://github.com/PostgREST/postgrest/issues/3508#issuecomment-2113162578 (perhaps this comment wasn't addressed yet in this PR?)

develop7:
Closing this in favor of #3559

Issue #3523, opened 2024-05-15
The schema cache load is not retried with exponential backoff


steve-chavez:
## Problem

Adding the following on `test/spec/fixtures/schema.sql`:
```sql
alter role postgrest_test_authenticator set statement_timeout to '10'
```

Makes the schema cache retry load on an endless loop without backoff:

```bash
# $ postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run 
# ...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
08/May/2024:21:06:46 +0200: Attempting to connect to the database...
```

Previously reported on https://github.com/PostgREST/postgrest/issues/3424#issuecomment-2101258619

steve-chavez:
Still an issue, this was incorrectly closed by #3536

PR #3524, opened 2024-05-15
refactor: move checkIsFatal logic to usePool

steve-chavez:
Towards solving https://github.com/PostgREST/postgrest/issues/3523

The fatal logic is now inside `usePool`. It centralizes the logic which is better for Locality of Behavior.

Removes:

- The need to do checkIsFatal on other parts of the code
- SCFatalFail/ConnFatalFail states which are no longer needed.

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -122,14 +122,12 @@ data AppState = AppState
 data SchemaCacheStatus
   = SCLoaded
   | SCPending
-  | SCFatalFail
   deriving Eq
 
 -- | Current database connection status
 data ConnectionStatus
   = ConnEstablished
   | ConnPending
-  | ConnFatalFail Text
   deriving Eq
 
 type AppSockets = (NS.Socket, Maybe NS.Socket)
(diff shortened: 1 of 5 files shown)


src/PostgREST/AppState.hs
```diff
@@ -226,15 +224,57 @@ initPool AppConfig{..} observer =
...
+          observer $ ExitDBFatalError ServerPgrstBug err
+          killThread mainThreadId
+        SQL.UnexpectedAmountOfRows{} -> do
+          observer $ ExitDBFatalError ServerPgrstBug err
+          killThread mainThreadId
+        -- Check for a syntax error (42601 is the pg code) only for queries that don't have `WITH pgrst_source` as prefix.
+        -- This would mean the error is on our schema cache queries, so we treat it as fatal.
+        -- TODO have a better way to mark this as a schema cache query
+        SQL.ServerError "42601" _ _ _ _ ->
+          unless ("WITH pgrst_source" `BS.isPrefixOf` tpl) $ do
```
steve-chavez, line 257: This was added on https://github.com/PostgREST/postgrest/pull/1685/files#diff-328c2f3103b25df9c32be37811eabc32ea0f39cea4f860b33651b3d470adf974R221. To preserve behavior I'm marking it here to only act on schema cache queries because on older pg versions `42601` also appears on API queries, in particular when inserting on generated columns.

https://github.com/PostgREST/postgrest/blob/d25df459ea88538249dd275e55765b350eafa84e/test/spec/Feature/Query/InsertSpec.hs#L550-L571

src/PostgREST/AppState.hs
```diff
@@ -373,24 +408,25 @@ loadSchemaCache appState@AppState{stateObserver=observer} = do
...
-          unless configDbPoolAutomaticRecovery
-              $ observer ExitDBNoRecoveryObs >> killThread (getMainThreadId appState)
+          unless configDbPoolAutomaticRecovery $ do
+            observer ExitDBNoRecoveryObs
+            killThread mainThreadId
         ConnEstablished -> do
+          actualPgVersion <- getPgVersion appState
+          when (actualPgVersion < minimumPgVersion) $ do
+            observer $ ExitUnsupportedPgVersion actualPgVersion minimumPgVersion
+            killThread mainThreadId
```
steve-chavez, line 426: coverage failing because we have no way of testing when we quit

PR #3533, opened 2024-05-18
fix: listener silent fail on replica

steve-chavez:
Update hasql-notifications to include the fix on https://github.com/diogob/hasql-notifications/issues/24.

Which now reveals the following error:

```
$ postgrest-with-postgresql-16 --replica -f test/spec/fixtures/load.sql postgrest-run

17/May/2024:18:35:38 -0500: Successfully connected to PostgreSQL 16.2 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit
17/May/2024:18:35:38 -0500: Failed listening for notifications on the "pgrst" channel. ERROR:  cannot execute LISTEN during recovery
17/May/2024:18:35:38 -0500: Retrying listening for notifications...
```

This is still not good because the LISTEN channel will be retried forever (with an unnecessary schema cache reload too) without a backoff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3424, Admin `/live` and `/ready` now differentiates a failure as 500 status - @steve-chavez
     + 503 status is still given when postgREST is in a recovering state
  - #3478, Media Types are parsed case insensitively - @develop7
+ - #2781, Fix listener silently failing on read replica - @steve-chavez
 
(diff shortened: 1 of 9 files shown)


steve-chavez:
> This is still not good because the LISTEN channel will be retried forever (with an unnecessary schema cache reload too) without a backoff.

Tried to fix that here but it's causing some test failures that only occur on CI (hard to debug). Will make the backoff in another PR to have less changes to track.

Linkcheck fails Patreon permanently.

PR #3534, opened 2024-05-18
fix: fix incorrect 413 error on pg 54* errors

### LINKED ISSUE #3255: Incorrect `413 Request Entity Too Large` on pg `statement_too_complex`
## Problem

Having an infinite loop on insert:

```sql
create table test.ins(
  id int
, name text
);

create or replace function ins()
returns trigger as $$ begin
  insert into ins values (NEW.id, NEW.name);
end $$ language plpgsql;

create trigger do_ins
after insert
on ins
for each row
execute procedure ins();
```

We reply with:

```bash
http POST localhost:3000/ins <<JSON
> {"id": 3, "name": "qwer"}
> JSON

HTTP/1.1 413 Request Entity Too Large
Date: Fri, 23 Feb 2024 21:32:00 GMT
Server: postgrest/12.1 (160caaf)
Transfer-Encoding: chunked

{
    "code": "54001",
    "details": null,
    "hint": "Increase the configuration parameter \"max_stack_depth\" (currently 2048kB), after ensuring the platform's stack depth limit is adequate.",
    "message": "stack depth limit exceeded"
}
```

Which is incorrect. In general because a 4xx is a client side error.

This is done on:

https://github.com/PostgREST/postgrest/blob/1a141c19df43aecf36afd69be0c3d9df739be1fc/src/PostgREST/Error.hs#L479

## Solution

Reply with a 500 status instead for every `54xxx`.

It's tempting to turn `54xxx` into a [508 Loop Detected](https://http.dev/508) but maybe that's not correct for every case.

taimoorzaeem:
Fixes #3255.


laurenceisla:
Rebased CHANGELOG. Should be good to merge after CI runs.

laurenceisla:
> I have seen the linkcheck failure repeatedly now:
> 
> ```
>  (       ecosystem: line   23) broken    https://blog.frankel.ch/poor-man-api - 403 Client Error: Forbidden for url: https://blog.frankel.ch/poor-man-api
> ```
> 
> But the URL is still accessible...

It may similar to the issue in https://github.com/PostgREST/postgrest/pull/3535#issue-2304726238. I like your proposal in #3544, it should also solve the other links.

laurenceisla (approved):

PR #3536, opened 2024-05-19
fix: listener retries with exponential backoff

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/3523.

Now it works like:

```
$ postgrest-with-postgresql-16 --replica -f test/spec/fixtures/load.sql postgrest-run

19/May/2024:20:24:28 -0500: Retrying listening for notifications in 1 seconds...
19/May/2024:20:24:28 -0500: Failed listening for notifications on the "pgrst" channel. ERROR:  cannot execute LISTEN during recovery
19/May/2024:20:24:30 -0500: Retrying listening for notifications in 2 seconds...
19/May/2024:20:24:30 -0500: Failed listening for notifications on the "pgrst" channel. ERROR:  cannot execute LISTEN during recovery
19/May/2024:20:24:34 -0500: Retrying listening for notifications in 4 seconds...
19/May/2024:20:24:34 -0500: Failed listening for notifications on the "pgrst" channel. ERROR:  cannot execute LISTEN during recovery
```

## Limitation

Once the listen channel is recovered, the retry status is not reset. So if the last backoff was 4 seconds, the next time recovery kicks in the backoff will be 8 seconds. If a recovery reaches 32 seconds, the backoff will stay there.

This is a problem with the interaction between `hasql-notifications` and `retry`. The `Hasql.Notifications.waitForNotifications` uses a forever loop that only finishes when it throws an exception, retry recovers on an exception and succeeds (here it restarts the retry status) when the function finishes .

I've left this as a TODO for now, it's still better than retrying without backoff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3424, Admin `/live` and `/ready` now differentiates a failure as 500 status - @steve-chavez
     + 503 status is still given when postgREST is in a recovering state
  - #3478, Media Types are parsed case insensitively - @develop7
- - #2781, Fix listener silently failing on read replica - @steve-chavez
+ - #3533, #3536, Fix listener silently failing on read replica - @steve-chavez
+    + If the LISTEN connection fails, it's retried with exponential backoff
 
 ### Deprecated
 
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -70,7 +70,7 @@ run appState = do
   AppState.connectionWorker appState -- Loads the initial SchemaCache
   Unix.installSignalHandlers (AppState.getMainThreadId appState) (AppState.connectionWorker appState) (AppState.reReadConfig False appState)
   -- reload schema cache + config on NOTIFY
-  AppState.runListener conf appState
+  AppState.runListener appState
 
   Admin.runAdmin appState (serverSettings conf)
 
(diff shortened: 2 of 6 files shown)


steve-chavez:
Really puzzled by this error that only happens on CI:

```
def set_statement_timeout(postgrest, role, milliseconds):
        """Set the statement timeout for the given role.
        For this to work reliably with low previous timeout settings,
        use a postgrest instance that doesn't use the affected role."""
    
        response = postgrest.session.post(
            "/rpc/set_statement_timeout", data={"role": role, "milliseconds": milliseconds}
        )
>       assert response.text == ""
E       assert '{"code":"57P01","details":null,"hint":null,"message":"terminating connection due to administrator command"}' == ''
E         
E         + {"code":"57P01","details":null,"hint":null,"message":"terminating connection due to administrator command"}
```

Whenever io tests fail, they're really hard to debug.

steve-chavez:
It's like something is calling `pg_terminate_backend`. AFAICT there's no relationship between statement_timeout and LISTEN.

steve-chavez:
> It's like something is calling pg_terminate_backend

That was it, I just skipped this test:

https://github.com/PostgREST/postgrest/blob/44160502f423190978efe678c941357ac761b782/test/io/test_io.py#L1212-L1226

And the tests no longer fail.

The thing that has changed in this PR is that the `listener` no longer waits for the `connectionWorker` (since the listener now has its own backoff), so looks like the `pg_terminate_backend`  was affecting the LISTEN connection too.

---

I think this can be fixed by isolating `terminate_pgrst` with a particular application name.

Issue #3290, opened 2024-02-27
Hard to upgrade from `v9.0.1`


steve-chavez:
## Problem

v10.0.0 introduced one-to-one relationship detection https://github.com/PostgREST/postgrest/pull/2439, which changed the json output of responses using resource embedding.  For upgrading, adding computed relationships was recommended but this is considerable work for projects that have many tables. For this reason some projects are stuck on `v9.0.1` .

## Proposal

The following config can be used to determine which relationships will be represented as json objects. By default it will be:

```
db-resource-embedding-single = "many-to-one, one-to-one"
```

So projects stuck on `9.0.1` can do:

```
db-resource-embedding-single = "many-to-one"
```

## Related

- https://github.com/PostgREST/postgrest/issues/2066

steve-chavez:
Yeah, I don't like it either. It also provides an easy way to have incompatible outputs between postgREST APIs, that could cause trouble.

> We should provide better migration paths, instead: Docs, Tools, anything that helps the actual migration.
> Could this be (partly) automated with dynamic SQL to create the computed relationships?

I was thinking we could have a `postgrest --dump-relationships=one-to-one` command, its output could help with generating the dynamic SQL.

steve-chavez:
Just using `jq` looks like a good idea.

Removing some unnecessary keys:

```bash
postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run --dump-schema \
  | jq '.dbRelationships | .[] | .[1] | .[] | select(.relCardinality.tag == "O2O")' \
  | jq 'del(.relFTableIsView,.relTableIsView,.tag,.relIsSelf)'
```

We get an output like:

```json
{
  "relCardinality": {
    "relColumns": [
      [
        "id",
        "id"
      ],
      [
        "code",
        "code"
      ]
    ],
    "relCons": "students_info_id_code_fkey",
    "tag": "O2O"
  },
  "relForeignTable": {
    "qiName": "students",
    "qiSchema": "test"
  },
  "relTable": {
    "qiName": "students_info",
    "qiSchema": "test"
  }
}

{
  "relCardinality": {
    "relColumns": [
      [
        "id",
        "id"
      ],
      [
        "code",
        "code"
      ]
    ],
    "relCons": "students_info_id_code_fkey",
    "tag": "O2O"
  },
  "relForeignTable": {
    "qiName": "students_info",
    "qiSchema": "test"
  },
  "relTable": {
    "qiName": "students",
[...]

PR #3546, opened 2024-05-21
refactor: add isParent flag to O2O rels

laurenceisla:
It allows to identify the side with the FK when isParent == False.

Follow up on https://github.com/PostgREST/postgrest/issues/3290#issuecomment-1982165331


src/PostgREST/Plan.hs
```diff
@@ -533,20 +533,20 @@ findRel schema allRels origin target hint =
     rs  -> Left $ AmbiguousRelBetween origin target rs
   where
     matchFKSingleCol hint_ card = case card of
-      O2M _ [(col, _)] -> hint_ == col
-      M2O _ [(col, _)] -> hint_ == col
-      O2O _ [(col, _)] -> hint_ == col
-      _                -> False
+      O2M _ [(col, _)]   -> hint_ == col
```
steve-chavez, line 536: To reduce these wildcards, how about:

```suggestion
      O2M{relColumns={(col, _)}}   -> hint_ == col
```

laurenceisla: Yes, it looks better that way.

steve-chavez (approved):

PR #3549, opened 2024-05-21
fix: remove verbosity from some error logs

laurenceisla:
Error logs starting with "An error occured..." are replaced with "Failed to..."

### DIFF
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -14,11 +14,11 @@ import qualified Data.ByteString.Lazy       as LBS
 import qualified Hasql.Transaction.Sessions as SQL
 import qualified Options.Applicative        as O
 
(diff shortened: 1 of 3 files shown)


src/PostgREST/CLI.hs
```diff
@@ -57,7 +57,7 @@ dumpSchema appState = do
       (transaction SQL.ReadCommitted SQL.Read $ querySchemaCache conf)
   case result of
     Left e -> do
-      hPutStrLn stderr $ "An error ocurred when loading the schema cache:\n" <> show e
+      hPutStrLn stderr $ "Failed to load the schema cache:\n" <> show e
```
steve-chavez, line 60: How about using `observer $ SchemaCacheErrorObs e` here? Just to have all the error messages in one place.

steve-chavez (approved):

PR #3552, opened 2024-05-22
docs: add "config" and "schema_cache" endpoints to the admin server

### LINKED ISSUE #3501: Write docs for v12.2.0 minor release
This is already listed in https://github.com/PostgREST/postgrest/issues/3113#issuecomment-1948631330, but to give this more visibility I'll create an issue about it.

Ever since merging the postgrest-docs repo into the core repo, we are supposed to merge docs together in the same PR as the corresponding feature. This didn't always happen initially, but we are getting better at it. However, we still have some unreleased feature commits on the main branch, which happened before the docs merge and thus have no corresponding docs written, yet.

This is blocking the next minor release, so needs to be taken care of.

I went through the changelog on main again, and I think the following still need to be documented:
- [ ] #3171, Add an ability to dump config via admin API - @skywriter
- [ ] #3210, Dump schema cache through admin API - @taimoorzaeem
- [ ] Deprecated: `Prefer: params=single-object` is deprecated. Use [a function with a single unnamed JSON parameter](https://postgrest.org/en/latest/references/api/stored_procedures.html#s-proc-single-json) instead. - @steve-chavez

@steve-chavez since you merged the first two and wrote the last yourself - can you take care of the docs for those to unblock a release?

There is also #3061, which is still undocumented, but also blocking a new minor release is #3242 which will be fixed in #3358. This essentially contains the docs for #3061, too.

laurenceisla:
Closes #3501

Adds missing documentation.

### DIFF
--- a/docs/references/admin_server.rst
+++ b/docs/references/admin_server.rst
@@ -54,3 +54,40 @@ Metrics
 =======
 
 Provides :ref:`metrics`.
+
+Runtime Configuration
+=====================
+
+Provides a ``config`` endpoint that returns the runtime :ref:`configuration`.
+
+.. code-block:: bash
+
+  curl "http://localhost:3001/config"
+
+.. code-block::
+
+  db-aggregates-enabled = false
+  db-anon-role = "web_anon"
+  db-channel = "pgrst"
+  db-channel-enabled = false
+  ...
+
+Runtime Schema Cache
+====================
+
+Provides the ``schema_cache`` endpoint that prints the runtime :ref:`schema_cache`.
+
+.. code-block:: bash
+
+  curl "http://localhost:3001/schema_cache"
+
+.. code-block:: json
+
+  {
+    "dbMediaHandlers": ["..."],
+    "dbRelationships": ["..."],
+    "dbRepresentations": ["..."],
+    "dbRoutines": ["..."],
+    "dbTables": ["..."],
+    "dbTimezones": ["..."]
+  }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Deprecated
 
- - `Prefer: params=single-object` is deprecated. Use [a function with a single unnamed JSON parameter](https://postgrest.org/en/latest/references/api/stored_procedures.html#s-proc-single-json) instead. - @steve-chavez
+ - `Prefer: params=single-object` is deprecated. Use [a function with a single unnamed JSON parameter](https://postgrest.org/en/latest/references/api/functions.html#function-single-json) instead. - @steve-chavez
 
 ### Documentation
 


docs/references/admin_server.rst
```diff
@@ -54,3 +54,45 @@ Metrics
 =======
 
 Provides :ref:`metrics`.
+
+Debugging
```
laurenceisla, line 58: Not sure if it's the best title. I remember it was mentioned somewhere that `--dump-config` and `--dump-schema` were used for debugging purposes.

steve-chavez: I think we should solve it like this.

- Remove the Healthcheck subtitle with its paragraph
- Put the Important note under Live
- Remove this Debugging title.

docs/references/admin_server.rst
```diff
@@ -54,3 +54,45 @@ Metrics
...
+=========
+
+You can also verify the configuration and schema cache values applied to a running PostgREST instance.
+
+Configuration
```
steve-chavez, line 63:
```suggestion
config
```

docs/references/admin_server.rst
steve-chavez: Ok, there's another option. Just remove this Debugging title and rename the new subtitles according to the below suggestions

docs/references/admin_server.rst
steve-chavez, line 63:
```suggestion
Runtime Configuration
```

docs/references/admin_server.rst
```diff
@@ -54,3 +54,45 @@ Metrics
...
+-------------
+
+Provides a ``config`` endpoint that returns all the :ref:`configuration values <configuration>` applied to the current instance.
```
steve-chavez, line 66:
```suggestion
Provides a ``config`` endpoint that returns the runtime :ref:`configuration`
```

docs/references/admin_server.rst
```diff
@@ -54,3 +54,45 @@ Metrics
...
+
+.. code-block::
+
+  db-aggregates-enabled = false
+  db-anon-role = "web_anon"
+  db-channel = "pgrst"
+  db-channel-enabled = false
+  ...
+
+Schema Cache
```
steve-chavez, line 80:
```suggestion
Runtime Schema Cache
```

docs/references/admin_server.rst
```diff
@@ -54,3 +54,45 @@ Metrics
...
+------------
+
+Provides the ``schema_cache`` endpoint that prints the :ref:`schema_cache` generated by the current instance.
```
steve-chavez, line 83:
```suggestion
Provides the ``schema_cache`` endpoint that prints the runtime :ref:`schema_cache`.
```

docs/references/admin_server.rst
laurenceisla: Seems good. I modified it accordingly.

steve-chavez (approved):

PR #3554, opened 2024-05-24
fix: log connection pool events on log-level="debug" instead of "info"

laurenceisla:
... since `log-level=debug` didn't exist when #3229 was added.

### DIFF
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -612,7 +612,7 @@ def test_pool_acquisition_timeout(level, defaultenv, metapostgrest):
 
         if level == "crit":
             assert len(output) == 0
-        elif level in ["info", "debug"]:
+        else:
             assert " 504 " in output[0]
             assert "Timed out acquiring connection from connection pool." in output[2]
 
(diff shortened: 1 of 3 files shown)


laurenceisla:
Needs work on tests, locally they sometimes fail on `test_log_postgrest_version` or `test_pool_acquisition_timeout[warn]`.

test/io/test_io.py
```diff
@@ -1164,9 +1178,9 @@ def test_log_postgrest_version(defaultenv):
     with run(env=defaultenv, no_startup_stdout=False) as postgrest:
         version = postgrest.session.head("/").headers["Server"].split("/")[1]
 
-        output = sorted(postgrest.read_stdout(nlines=5))
+        output = postgrest.read_stdout(nlines=1)
 
-        assert "Starting PostgREST %s..." % version in output[4]
+        assert "Starting PostgREST %s..." % version in output[0]
```
laurenceisla, line 1183: Since this test only cares about the first line in the output, I don't think it's necessary to check for more than one line and then order them. Sometimes, the "Listening for notifications..." message does not appear until later, messing up the `sort()` done and failing the test.

> Needs work on tests, locally they sometimes fail on `test_log_postgrest_version` or `test_pool_acquisition_timeout`.

This change fixes the `test_log_postgrest_version`. The other one seems related to #3424.

steve-chavez (approved):

laurenceisla:
Oops, there was a mistake in the commit prefix. It shouldn't be a `fix:` since it's not fixing a previous release, but it's a continuation of #3229 and should have a `feat:` prefix instead (or maybe a new one).

The changelog is OK though, it's included under the "Added" subtitle.

PR #3558, opened 2024-05-29
config: introduce admin-server-host config parameter

develop7:
Introduces optional `admin-server-host` config setting, which falls back to `server-host` value


test/io/configs/expected/no-defaults.config
```diff
@@ -35,6 +35,7 @@ server-trace-header = "X-Request-Id"
 server-timing-enabled = true
 server-unix-socket = "/tmp/pgrst_io_test.sock"
 server-unix-socket-mode = "777"
+admin-server-host = "!4"
```
laurenceisla, line 38: In this test, the values should be different than the default (different from `"!4"`).

test/io/configs/no-defaults-env.yaml
```diff
@@ -38,4 +38,5 @@ PGRST_SERVER_TRACE_HEADER: X-Request-Id
 PGRST_SERVER_TIMING_ENABLED: true
 PGRST_SERVER_UNIX_SOCKET: /tmp/pgrst_io_test.sock
 PGRST_SERVER_UNIX_SOCKET_MODE: 777
+PGRST_ADMIN_SERVER_HOST: "!4"
```
laurenceisla, line 41: Ditto.

laurenceisla (commented):
The new config should be tested in `test/io/db_config.sql` alongside the non-reloadable configs.

Also needs tests for the no-default configs (check below).

docs/references/configuration.rst
```diff
@@ -146,6 +146,21 @@ To reload the configuration from within the database, you can use the ``NOTIFY``
...
+
+  =============== =======================
+  **Type**        String
+  **Default**     `server-host` value
+  **Reloadable**  N
+  **Environment** PGRST_ADMIN_SERVER_HOST
+  **In-Database** `n/a`
+  =============== =======================
+
+  Specifies the host for the :ref:`admin_server`. Defaults to :ref:`server-host` value.
```
laurenceisla, line 162: Here, I understand that if I set only the `server-host`, like:

```
server-host = "127.0.0.1"
```

Then the `admin-server-host` should also be `127.0.0.1`?  In the code I see that it defaults to the fallback `!4` not the `server-host` right now.

If so, it would also need a test.

develop7: @wolfgangwalther good call, implemented it.

test/io/configs/expected/no-defaults.config
develop7: Thanks, done

src/PostgREST/Config.hs
```diff
@@ -282,6 +287,7 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
     <*> (fromMaybe False <$> optBool "server-timing-enabled")
     <*> (fmap T.unpack <$> optString "server-unix-socket")
     <*> parseSocketFileMode "server-unix-socket-mode"
+    <*> optString "admin-server-host"
```
laurenceisla, line 290: I think we could treat `configAdminServerHost` as a `Text` directly (without `Maybe`) and fallback to the default value at the moment of parsing the Config. This should avoid the extra `fromMaybe` done in other parts of the code.

You could use something like:

```suggestion
    <*> parseAdminServerHost "admin-server-host" "server-host"
```

And the implementation would be:

```haskell
    -- This should be replaced above: <*> parseServerHost "server-host"
    parseServerHost :: C.Key -> C.Parser C.Config Text
    parseServerHost key =
      fromMaybe "!4" <$> optString key

    parseAdminServerHost :: C.Key -> C.Key -> C.Parser C.Config Text
    parseAdminServerHost mainKey fallbackKey =
      optString mainKey >>= \case
        Just ash -> pure ash
        Nothing  -> parseServerHost fallbackKey
```

test/io/configs/expected/no-defaults-with-db-other-authenticator.config
```diff
@@ -35,6 +35,7 @@ server-trace-header = "traceparent"
 server-timing-enabled = true
 server-unix-socket = "/tmp/pgrst_io_test.sock"
 server-unix-socket-mode = "777"
+admin-server-host = "127.0.0.1"
```
laurenceisla, line 38: Btw, I also pushed a commit that changed the `no-default` tests. Since it uses it as default, `admin-server-host` must be different from `server-host`.

Do a `git pull` before making any changes.

src/PostgREST/Config.hs
develop7: That seems a very good approach, actually. We could get rid of other maybes in AppState this way as well.

test/io/configs/expected/no-defaults-with-db-other-authenticator.config
develop7: Thank you!

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+
+ - #3558, Add the `admin-server-host` config to set the host for the admin server - @develop7
```
laurenceisla, line 10: Moved changelog to Unreleased.

laurenceisla (approved):
LGTM!

steve-chavez (approved):

src/PostgREST/Config.hs
```diff
@@ -276,12 +278,13 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
...
     <*> parseOpenAPIServerProxyURI "openapi-server-proxy-uri"
     <*> parseCORSAllowedOrigins "server-cors-allowed-origins"
-    <*> (fromMaybe "!4" <$> optString "server-host")
+    <*> parseServerHost "server-host"
     <*> (fromMaybe 3000 <$> optInt "server-port")
     <*> (fmap (CI.mk . encodeUtf8) <$> optString "server-trace-header")
     <*> (fromMaybe False <$> optBool "server-timing-enabled")
     <*> (fmap T.unpack <$> optString "server-unix-socket")
     <*> parseSocketFileMode "server-unix-socket-mode"
+    <*> parseAdminServerHost "admin-server-host" "server-host"
```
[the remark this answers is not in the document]
laurenceisla: Oh cool, just noticed that there was a similar function already. Will try it out.

PR #3559, opened 2024-05-29
config forbid same server-port and admin-server-port

### LINKED ISSUE #3508: Server port and admin server port can end up with the same value

develop7:
Forbids `server-port` and `admin-server-port` from being equal altogether, despite they might not conflict at all in case admin and app are bound to different addresses. Implemented as per the discussion at
https://github.com/PostgREST/postgrest/issues/3508#issuecomment-2125123633

Fixes #3508

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2052, Dropped support for PostgreSQL 9.6 - @wolfgangwalther
  - #2052, Dropped support for PostgreSQL 10 - @wolfgangwalther
  - #2052, Dropped support for PostgreSQL 11 - @wolfgangwalther
- - #3508, #3559, Server port and admin server port can end up with the same value - @develop7
-    + The app fails to start when the server port and admin server port are the same
+ - #3508, PostgREST now fails to start when `server-port` and `admin-server-port` config options are the same - @develop7
 
 ## [12.2.0] - 2024-06-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3414, Force listener to connect to read-write instances using `target_session_attrs` - @steve-chavez
  - #3255, Fix incorrect `413 Request Entity Too Large` on pg errors `54*` - @taimoorzaeem
  - #3549, Remove verbosity from error logs starting with "An error occurred..." and replacing it with "Failed to..." - @laurenceisla
- - #3508, Server port and admin server port can end up with the same value - @develop7
+ - #3508, #3559, Server port and admin server port can end up with the same value - @develop7
     + The app is fails to start when the server port and admin server port are the same
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3414, Force listener to connect to read-write instances using `target_session_attrs` - @steve-chavez
  - #3255, Fix incorrect `413 Request Entity Too Large` on pg errors `54*` - @taimoorzaeem
  - #3549, Remove verbosity from error logs starting with "An error occurred..." and replacing it with "Failed to..." - @laurenceisla
+ - #3508, Server port and admin server port can end up with the same value - @develop7
+    + The app is fails to start when the server port and admin server port are the same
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3414, Force listener to connect to read-write instances using `target_session_attrs` - @steve-chavez
  - #3255, Fix incorrect `413 Request Entity Too Large` on pg errors `54*` - @taimoorzaeem
  - #3549, Remove verbosity from error logs starting with "An error occurred..." and replacing it with "Failed to..." - @laurenceisla
+ - #3508, #3559, Server port and admin server port can end up with the same value - @develop7
+    + The app is fails to start when the server port and admin server port are the same
```
steve-chavez, line 53: Should be moved to the UNRELEASED section https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#unreleased

develop7: Done!

src/PostgREST/Config.hs
```diff
@@ -222,12 +222,21 @@ readAppConfig dbSettings optPath prevDbUri roleSettings roleIsolationLvl = do
     Left err ->
       return . Left $ "Error in config " <> err
     Right parsedConfig ->
-      Right <$> decodeLoadFiles parsedConfig
+      case processForbiddenValues parsedConfig of
+        Left err' ->
+          return . Left $ "Error in config " <> err'
```
steve-chavez, line 227: How about adding an io test for this? Maybe like:

https://github.com/PostgREST/postgrest/blob/86a48b5b4152e4a5a92e60128cca4eb50d1230d2/test/io/test_io.py#L1235-L1245

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #2052, Dropped support for PostgreSQL 9.6 - @wolfgangwalther
  - #2052, Dropped support for PostgreSQL 10 - @wolfgangwalther
  - #2052, Dropped support for PostgreSQL 11 - @wolfgangwalther
+ - #3508, #3559, Server port and admin server port can end up with the same value - @develop7
+    + The app fails to start when the server port and admin server port are the same
```
laurenceisla, line 18:
```suggestion
 - #3508, #3559, PostgREST now fails to start when `server-port` and `admin-server-port` config options are the same - @develop7
```

src/PostgREST/Config.hs
develop7: thanks for the suggestion, I'm debugging the test now :)

src/PostgREST/Config.hs
```diff
@@ -222,12 +222,21 @@ readAppConfig dbSettings optPath prevDbUri roleSettings roleIsolationLvl = do
...
+          Right <$> decodeLoadFiles config
   where
     -- Both C.ParseError and IOError are shown here
     loadConfig :: FilePath -> IO (Either SomeException C.Config)
     loadConfig = try . C.load
 
+    processForbiddenValues :: AppConfig -> Either Text AppConfig
+    processForbiddenValues cfg@AppConfig{..}
+      | configAdminServerPort == Just configServerPort = Left "admin-server-port cannot be the same as server-port"
+      | otherwise = Right cfg
```
laurenceisla, line 238: Could this be done in the `parser`? You could create and reuse a `parseServerPort "server-port"`  function and do the validation when parsing the `admin-server-port`. Something like (an idea, may have errors):

```haskell
 parseAdminServerPort k = do
      serverPort <- parseServerPort "server-port"
      optString k >>= \case
        Nothing -> pure Nothing
        Just asp | asp == serverPort -> fail "admin-server-port cannot be the same as server-port"
                 | otherwise -> pure $ Just asp
```

develop7: Done!

steve-chavez (approved):
Great work! :100:

Issue #3569, opened 2024-06-03
"threads.c" error when running binaries or docker image for amd64 devel version


laurenceisla:
### Environment

* PostgreSQL version: PostgreSQL 15.6 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 12.3.0, 64-bit
* PostgREST version: v12.1 (devel)
* PostgREST docker version: devel
* Operating system: NixOS 23.11

### Description of issue

Executing the binary or docker release [for the `devel` branch ](https://github.com/PostgREST/postgrest/releases/tag/devel), returns an error when trying to connect to the database.

Local binary:

```
PGRST_DB_URI=postgres://postgres@localhost:5432/postgres \
PGRST_DB_ANON_ROLE=postgres \
PGRST_DB_LOG_LEVEL=info \
postgrest-run
```
```console
03/Jun/2024:13:19:01 -0500: Starting PostgREST 12.1...
03/Jun/2024:13:19:01 -0500: Listening on port 3000
03/Jun/2024:13:19:01 -0500: Attempting to connect to the database...
Assertion failed: *(&once->n) != 4 (threads.c: k5_once: 564)
fish: Job 1, 'PGRST_DB_URI=postgres://postgre…' terminated by signal PGRST_DB_ANON_ROLE=postgres \ (PGRST_DB_LOG_LEVEL=info \)
fish: Job ./postgrest, 'SIGABRT' terminated by signal Abort ()
```

Docker:

```
sudo docker run --net=host --rm --name test-db-conn -p 3000:3000 \
            -e PGRST_DB_URI="postgres://postgres@localhost:5432/postgres" \
            -e PGRST_DB_ANON_ROLE="postgres" \
            -e PGRST_LOG_LEVEL="info" \
            postgrest/postgrest:devel
```
```console
03/Jun/2024:18:21:03 +0000: Starting PostgREST 12.1...
03/Jun/2024:18:21:03 +0000: Listening on port 3000
03/Jun/2024:18:21:03 +0000: Attempting to connect to the database...
Assertion failed: *(&once->n) != 4 (threads.c: k5_once: 564)
```

Running the `main` branch from a Nix environment does not return this error:

```bash
PGRST_DB_URI=postgres://postgres@localhost:5432/postgres \
PGRST_DB_ANON_ROLE=postgres \
PGRST_DB_LOG_LEVEL=info \
postgrest-run
```
```console
[...]

PR #3572, opened 2024-06-05
correct exponential backoff on Listener

steve-chavez:
Clears the limitation mentioned on https://github.com/PostgREST/postgrest/pull/3536

The Listener no longer uses the https://hackage.haskell.org/package/retry package and instead uses a much simpler IORef in AppState for the delays.

Additionally it no longer uses exception throwing/catching, which is rather messy and brings some
concerns (https://github.com/PostgREST/postgrest/issues/3569#issuecomment-2146013327).

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -120,7 +120,7 @@ postgrest logLevel appState connWorker =
         -- the connWorker is done.
         when (isServiceUnavailable response) connWorker
         resp <- do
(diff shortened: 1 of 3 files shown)


steve-chavez:
codecov fails because of missing tests for LISTEN recovery. I've put this is a TODO on https://github.com/PostgREST/postgrest/issues/1766#issuecomment-2150021160

PR #3573, opened 2024-06-05
prevent GSSAPI error between Listener and pool

steve-chavez:
Brings back the the signaling/waiting between the connection pool and the Listener.

This behavior was on the previous release, but was removed on https://github.com/PostgREST/postgrest/commit/3cf565614d6529b4f691699b254fd3a680dfcda0. See https://github.com/PostgREST/postgrest/issues/3569#issuecomment-2146336484. 

Prevents the GSSAPI error shown on https://github.com/PostgREST/postgrest/issues/3569

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -29,6 +29,7 @@ module PostgREST.AppState
   , getObserver
   , isLoaded
   , isPending
+  , waitForListenerCanStart
   ) where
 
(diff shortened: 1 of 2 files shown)


steve-chavez:
No, not at all. It's only makes the listener wait for a functional connection pool.

This was already in before as discussed on https://github.com/PostgREST/postgrest/issues/3569#issuecomment-2146336484

PR #3581, opened 2024-06-11
docs: deprecate EOL pg versions

steve-chavez:
As agreed on https://github.com/PostgREST/postgrest/issues/3501#issuecomment-2134016429


steve-chavez:
Dropping support for 3 pg versions in one go is quite harsh. But I'm not doing maintenance work on the nixpkgs side (and I don't have the bandwith), so it's only fair.

PR #3592, opened 2024-06-14
docs: improve architecture diagram

laurenceisla:
Does the following:

- SVG format instead of PNG
- The components now have links to ~~the source code~~ their reference in the Docs
- Supports dark mode


docs/explanations/architecture.rst
```diff
@@ -6,7 +6,18 @@ This page describes the architecture of PostgREST.
 Bird's Eye View
 ===============
 
-.. image:: ../_static/arch.png
+  .. container:: img-dark
+
+    .. raw:: html
+
+      <object width="100%" data="../_static/arch-dark.svg" type="image/svg+xml"></object>
```
laurenceisla, line 13: Using `.. image` does not print an interactive image. This is a workaround I found in https://github.com/sphinx-doc/sphinx/issues/2240#issuecomment-187366626.

docs/_diagrams/arch.uml
```diff
@@ -67,4 +65,20 @@ end note
...
+url of Auth is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Auth.hs]]
+url of ApiRequest is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/ApiRequest.hs]]
+url of Plan is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Plan.hs]]
+url of Query is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Query.hs]]
+url of "Schema Cache" is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/SchemaCache.hs]]
+url of Admin is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Admin.hs]]
+url of Config is [[https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Config.hs]]
+
+url of HTTPAPI is [[https://aosabook.org/en/posa/warp.html]]
+url of HTTPADMIN is [[https://aosabook.org/en/posa/warp.html]]
```
laurenceisla, line 80: This links to the Warp library, not sure if that's OK or if it should be removed.

steve-chavez:
Looks awesome! 

One thing though, should we have links in the arch diagram to the components or to the reference docs?

If we did it that way, we could also link:

- Proxy -> https://postgrest--3592.org.readthedocs.build/en/3592/explanations/nginx.html
- Authorization -> https://postgrest--3592.org.readthedocs.build/en/3592/explanations/db_authz.html
- API schema -> https://postgrest--3592.org.readthedocs.build/en/3592/explanations/schema_isolation.html
- Admin -> https://postgrest--3592.org.readthedocs.build/en/3592/references/admin_server.html
- ... (all major modules have a reference page)
- (except ApiRequest and Plan, here I think there would be no links, for now)

This would be great for end users, since you have all the PostgREST parts in one single diagram.

docs/_diagrams/arch.uml
steve-chavez: I think it's good :+1:

laurenceisla:
> Do you think there is any way to make the background transparent in both svgs instead of a sliiightly different color than readthedocs uses?

Yes, it is possible, will do that.

> One thing though, should we have links in the arch diagram to the components or to the reference docs?

I had the same doubt, but I like that each component/interface is linked to its respective module in PostgREST source code. We could somehow do both if the components keep linking to the source code and the Notes on those components link to their docs counterpart. Also those that do not link to the source code (like API schema or Connection Pool) could link directly to the docs without needing Notes.

Not sure WDYT, maybe it could be untidy or confusing to do it that way?

steve-chavez:
> We could somehow do both if the components keep linking to the source code and the Notes on those components link to their docs counterpart

Sounds good to me, if the diagram maintains a nice layout with the added notes. That could be trouble, too much detail would make it less useful.

>  but I like that each component/interface is linked to its respective module in PostgREST source code.

I's less useful for end users that way, maybe more useful for us devs.. but only slightly so (the codemap below already contains the same info). The point of making the arch diagram part of the docs (instead of being a markdown file on the repo) was to make it more generally useful.

steve-chavez:
> but I like that each component/interface is linked to its respective module in PostgREST source code.

If we want to keep it that way, I'd say remove the architecture page from the docs and just move it back to a markdown file on the root of the repo. That way is clear that is for developers only.

laurenceisla:
> I's less useful for end users that way, maybe more useful for us devs.. but only slightly so (the codemap below already contains the same info)

Ahhh, yes, I see your point.

> I'd say remove the architecture page from the docs and just move it back to a markdown file on the root of the repo

No, I think it's OK as you mentioned, with the links to the reference docs. It's a lot better for end users and devs can check below for the components.

laurenceisla:
Done:

- Transparent background
- Links now reference the documentation counterpart of the component instead of the source code

> (except ApiRequest and Plan, here I think there would be no links, for now)

I think Query also doesn't have a reference?

docs/_diagrams/arch-dark.uml
```diff
@@ -0,0 +1,3 @@
+@startuml
+!include arch.uml
+@enduml
```
steve-chavez, line 3: Q: Is the `arch-dark.uml` really necessary? Can't we do something like:

```
${plantuml}/bin/plantuml -tsvg -darkmode arch.uml -o ../_static/arch-dark.svg
```

docs/explanations/architecture.rst
steve-chavez: How about adding that link a a comment here? So we have the reference

docs/_diagrams/arch-dark.uml
laurenceisla: Ah, no, it won't let me. Doing that creates the SVG in this path:

```
/_static/arch-dark.svg/arch.svg
```

I tried using `-filename` ([see the docs](https://plantuml.com/en/command-line)) before and after the file, but it doesn't seem to work.

```
${plantuml}/bin/plantuml -tsvg -darkmode -filename "arch-dark.uml" arch.uml -o ../_static/arch-dark.svg
```

Since it creates a graphic per file I had to create another file for the dark mode.

steve-chavez:
Nitpick: the clickable area for the Authorization and API Schema links is only limited to the text, not the whole rectangle.
Not a big deal if not possible.

On "Bird’s Eye View" we could also say that the boxes are clickable, otherwise it's easy to miss.

On "Schema Isolation", the "This page is a work in progress" note can be deleted.

docs/_diagrams/arch-dark.uml
steve-chavez: Ok cool. Please comment that here on this file. Otherwise it's not obvious why it was done this way.

laurenceisla:
> Nitpick: the clickable area for the Authorization and API Schema links is only limited to the text, not the whole rectangle. Not a big deal if not possible.

It seems to use the whole rectangle when the it's small. I could not find an option that lets me do that for bigger rectangles.

docs/_diagrams/arch-dark.uml
laurenceisla: I added the info to the [diagram's README](https://github.com/PostgREST/postgrest/blob/f8c7f3b6cee9b8ee4558197fbb17f744ef10e4f7/docs/_diagrams/README.md#uml).

steve-chavez (approved):

Issue #3147, opened 2024-01-09
LISTEN channel stops working due to postgres bug


steve-chavez:
## Problem

This just happened on a production instance. `NOTIFY pgrst` wasn't triggering a schema cache reload while according to `pg_stat_activity` the `LISTEN` channel was up for several days:

```sql
select * from pg_stat_activity where application_name ilike '%postgrest%' and query ilike '%listen%';
-[ RECORD 1 ]----+------------------------------
datid            | 12974
datname          | postgres
pid              | 1368652
leader_pid       | 
usesysid         | 16665
usename          | authenticator
application_name | postgrest
client_addr      | 127.0.0.1
client_hostname  | 
client_port      | 53768
backend_start    | 2023-12-28 11:23:50.967734+00
xact_start       | 
query_start      | 2023-12-28 11:23:50.985951+00
state_change     | 2023-12-28 11:23:50.986375+00
wait_event_type  | Client
wait_event       | ClientRead
state            | idle
backend_xid      | 
backend_xmin     | 
query_id         | 7372465010149019010
query            | LISTEN "pgrst"
backend_type     | client backend
``` 

Further investigation lead to finding this PostgreSQL bug:

```sql
postgres=# listen pgrst;
ERROR:  could not access status of transaction 1044524795
DETAIL:  Could not open file "pg_xact/03E4": No such file or directory.

postgres=# listen pgrst;
ERROR:  could not access status of transaction 1044524795
DETAIL:  Could not open file "pg_xact/03E4": No such file or directory.

postgres=# listen xx;
ERROR:  could not access status of transaction 1044524795
DETAIL:  Could not open file "pg_xact/03E4": No such file or directory.

postgres=# listen yy;
ERROR:  could not access status of transaction 1044524795
DETAIL:  Could not open file "pg_xact/03E4": No such file or directory.
```

More details:

- https://dba.stackexchange.com/questions/325104/error-could-not-access-status-of-transaction-could-not-open-file-pg-xact-0029
- https://www.postgresql.org/message-id/flat/VE1PR03MB531295B1BDCFE422441B15FD92499%40VE1PR03MB5312.eurprd03.prod.outlook.com#7e36d1fdca921b5292e92c7017984ffa

So seems the LISTEN channel is broken. I can't reproduce this locally though.

The only way to repair this error is by restarting the PostgreSQL instance AFAICT.

## Workaround

Killing the LISTEN channel will force PostgREST to reload.

```sql
[...]

PR #3620, opened 2024-06-26
fix: don't reload cache on every listener fail

steve-chavez:
As mentioned on https://github.com/PostgREST/postgrest/issues/3147#issuecomment-2181596749.

Revert "prevent GSSAPI error between Listener and pool"

This reverts commit 4beac10d3d0116e0fa79324782b1baba62bb1368.


steve-chavez:
Yeah, I actually was intending to release a `v12.2.1` with this change.

Issue #1840, opened 2021-05-03
GET request failing with "message": "Server lacks JWT secret" error


priyadarshiniguptan:
Hi,
I am using Postgrest on docker and the below is my docker compose - 

postgREST:
    container_name: middleware-rdbms-auto-baas_postgREST
    image: postgrest/postgrest
    restart: always
    ports:
      - '${POSTGREST_PGDCP_EXPOSE_PORT:-3000}:3000'
    environment:
      PGRST_DB_URI: >-
        postgres://${POSTGRESQLENGINE_USER}:${POSTGRESQLENGINE_PASSWORD}@${POSTGRESQLENGINE_HOST}:${POSTGRESQLENGINE_PORT}/${POSTGRESQLENGINE_DB}
      PGRST_DB_SCHEMA: '${PGDCP_SCHEMA}'
      PGRST_DB_ANON_ROLE: '${POSTGRESQLENGINE_USER}'
      PGRST_JWT_SECRET: 'pgdcpsecurewebtoken'

The login user is the authenticator role and I have also specified a jwt-secret that is being used by a function in Postgres that uses the sign method of pgjwt extension to generate the jwt.

However, when I try to 
GET http://10.10.11.124:4432/<api name> with the authorization header set to Bearer <jwt token> I get the error below:

{
    "message": "Server lacks JWT secret"
}

Please let me know where could I possibly be going wrong.

wolfgangwalther:
> PGRST_JWT_SECRET: 'pgdcpsecurewebtoken'

We need a secret of at least 32 characters, as [stated in the docs](https://postgrest.org/en/stable/configuration.html#jwt-secret).

However, we should try to throw a better error message in that case - possibly already at startup.

priyadarshiniguptan:
Thanks for clarifying however I tried giving a secret longer than 32 characters and gave a restart. But still the issue persists. 
The secret I am using now is something like 'pgdcpsecurewebtokenpgdcpsecurewebtoken'

Secondly , I also tried generating a token on jwt.io using this secret and I got a jwt as below:

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoibWVkaWd5X2FkbWluX3JlYWRfb25seSIsImV4cCI6MTYyMDEwNDI4MH0.LmtOC1igJ1-ilu0ra3d99FnxlbJviJpFnNlLJ5UV1Ug

Using this in my get request fails with the same error 
{
[...]

PR #3628, opened 2024-06-29
feat: add message when JWT secret is less than 32 characters long

### LINKED ISSUE #3607: Re-implement minimal length for jwt-secret
It seems that after the change to `jose-jwt` in 465170c7d6c51dace3faf68fbb4040353d5571c0, we don't require a minimum of 32 characters for `jwt-secret` anymore.

The question is: Should we keep it like that or re-implement that limit ourselves? If we re-implement it, we can surely do https://github.com/PostgREST/postgrest/issues/1840#issuecomment-1034857931.

I tend to say the limit is useful and we should implement it again with a proper error message.

laurenceisla:
Closes #3607  and addresses part of #1840.


laurenceisla:
IO tests are failing in "no-default config" tests:

https://github.com/PostgREST/postgrest/blob/40ed349bba160247ee2d10b7a48e836077a5ac7e/test/io/configs/no-defaults.config#L24-L25

This is because, while the base64 encoded `jwt-secret` is at least 32 chars, once it's decoded it has less than 32 characters. If I'm not mistaken, this is how it worked before, but now we're failing before starting PostgREST.

Also, could this be considered a breaking change or just a feature/fix? PostgREST will fail and won't run or reload the config (if it's already running) when the secret has less than 32 characters. I think this is better because it's mostly a developer issue with the config, not for the end user.

laurenceisla:
This is related to #3629. When setting, for example, `set_config('pgrst.jwt_secret', 'small-secret', true);`, then it will happen what's mentioned in that issue.

laurenceisla:
> the limit was enforced after decoding.

Ah I worded it wrongly, that's what I meant to say.

> It might even be a breaking change, but that's not really important at this stage, because we have merged a few things already which will require a major version bump anyway.

:+1:

steve-chavez (approved):

PR #3634, opened 2024-07-02
fix: handle when file in pgrst.jwt_secret does not exist

### LINKED ISSUE #3629: PostgREST partially starts and does not return error when selecting an invalid file in `pgrst.jwt_secret`
### Description of issue

PostgREST does not fully load and does not print a useful error when an invalid file is selected for `pgrst.jwt_secret`. E.g. with this configuration:

```sql
returns void as $$
  select set_config('pgrst.jwt_secret', '@/does/not/exist', true);
$$ language sql;
```

Running PostgREST prints:

```bash
28/Jun/2024:21:22:50 -0500: Starting PostgREST 12.2.0...
28/Jun/2024:21:22:50 -0500: Admin server listening on port 3001
28/Jun/2024:21:22:50 -0500: Attempting to connect to the database...
28/Jun/2024:21:22:50 -0500: Listening on port 3000
28/Jun/2024:21:22:50 -0500: Successfully connected to PostgreSQL 15.6 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 12.3.0, 64-bit
28/Jun/2024:21:22:50 -0500: Listening for notifications on the "pgrst" channel
```

No schema cache/config logs and no error. A simple request returns this error:

```json
{"code":"PGRST002","details":null,"hint":null,"message":"Could not query the database for the schema cache. Retrying."}
```

This only fails for In-Database configurations, for ENV variables and Config files it prints the error correctly:

```
postgrest: /does/not/exist: withBinaryFile: does not exist (No such file or directory)
```

laurenceisla:
Closes #3629

- PostgREST now starts/reloads config correctly with old jwt-secret
- The error is logged to stderr

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -222,7 +222,8 @@ readAppConfig dbSettings optPath prevDbUri roleSettings roleIsolationLvl = do
     Left err ->
       return . Left $ "Error in config " <> err
     Right parsedConfig ->
-      Right <$> decodeLoadFiles parsedConfig
+      catch (Right <$> decodeLoadFiles parsedConfig)
+            (\e -> pure $ Left (show (e :: IOException)))
   where
     -- Both C.ParseError and IOError are shown here
     loadConfig :: FilePath -> IO (Either SomeException C.Config)


steve-chavez:
:thinking: Hm, it doesn't make sense to refer to a file in the in-db config right? 

I guess being consistent and no silent fail is still good though.

laurenceisla:
> 🤔 Hm, it doesn't make sense to refer to a file in the in-db config right?

Ah, got it.

> I guess being consistent and no silent fail is still good though.

I'll add this in #3628, then. Since handling the issue is now more relevant there.

PR #3643, opened 2024-07-08
perf: Pass arguments to RPCs called via GET directly

### LINKED ISSUE #2858: Postgrest calling function gives very different query plan from console
Before reporting a bug:
send the server a SIGUSR1 signal or restart it(http://postgrest.org/en/stable/admin.html#schema-reloading)
to ensure the schema cache is not stale. This sometimes fixes apparent bugs.
- Not sure how I can do this using Supabase, but would love to try it.

### Environment

* PostgreSQL version: (if using docker, specify the image): 15.1.0.37
* PostgREST version: (if using docker, specify the image): 10.2 
* Operating system: Supabase instance, not sure what OS

# Description of issue

## Expected
Running a function in SQL console or js give the same query plan.

## Actual
I get very different query plans: the console query plan uses the index I created while the postgrest query plan does something else. 

### Function:
```sql
create or replace function get_notifications (uid text, unseen_only boolean, max_num integer) 
    returns setof user_notifications language sql stable parallel SAFE as $$
select *
from user_notifications as n
where n.user_id = uid and (not unseen_only or not ((n.data->'isSeen')::boolean))
order by ((n.data->'createdTime')::bigint) desc
limit max_num
$$;
```
### SQL console:
```sql
explain analyze select * from get_notifications('AJwLWoo3xue32XIiAVrL5SyR1WB2', true, 300);
```
Results:
```
Subquery Scan on get_notifications  (cost=0.56..126.51 rows=300 width=1093) (actual time=6.352..38.031 rows=1 loops=1)
  ->  Limit  (cost=0.56..123.51 rows=300 width=1101) (actual time=6.350..38.027 rows=1 loops=1)
        ->  Index Scan using user_notifications_unseen_created_time_idx on user_notifications n  (cost=0.56..6313.57 rows=15404 width=1101) (actual time=6.349..38.025 rows=1 loops=1)
              Index Cond: ((user_id = 'AJwLWoo3xue32XIiAVrL5SyR1WB2'::text) AND (((data -> 'isSeen'::text))::boolean = false))
Planning Time: 0.513 ms
Execution Time: 38.076 ms
```

### JS:
```js
const explanation = await db
    .rpc('get_notifications', {
      uid:'AJwLWoo3xue32XIiAVrL5SyR1WB2,
      unseen_only: true,
      max_num: 300,
    })
    .explain({ analyze: true })
```
Results:
```
Aggregate  (cost=36.33..36.35 rows=1 width=112) (actual time=70.956..70.960 rows=1 loops=1)
  ->  Subquery Scan on _postgrest_t  (cost=1.32..31.32 rows=1000 width=128) (actual time=70.563..70.587 rows=13 loops=1)
        ->  Limit  (cost=1.32..21.32 rows=1000 width=104) (actual time=70.553..70.562 rows=13 loops=1)
              ->  Function Scan on get_notifications  (cost=1.32..11.32 rows=1000 width=104) (actual time=70.552..70.559 rows=13 loops=1)
                    CTE pgrst_args
                      ->  Function Scan on json_to_recordset _  (cost=0.02..1.01 rows=100 width=37) (actual time=0.023..0.024 rows=1 loops=1)
                            InitPlan 1 (returns $0)
                              ->  Result  (cost=0.00..0.01 rows=1 width=32) (actual time=0.003..0.004 rows=1 loops=1)
                    InitPlan 3 (returns $2)
                      ->  Limit  (cost=0.00..0.02 rows=1 width=32) (actual time=0.026..0.027 rows=1 loops=1)
                            ->  CTE Scan on pgrst_args  (cost=0.00..2.00 rows=100 width=32) (actual time=0.025..0.026 rows=1 loops=1)
                    InitPlan 4 (returns $3)
                      ->  Limit  (cost=0.00..0.02 rows=1 width=1) (actual time=0.001..0.002 rows=1 loops=1)
                            ->  CTE Scan on pgrst_args pgrst_args_1  (cost=0.00..2.00 rows=100 width=1) (actual time=0.001..0.001 rows=1 loops=1)
                    InitPlan 5 (returns $4)
                      ->  Limit  (cost=0.00..0.02 rows=1 width=4) (actual time=0.001..0.001 rows=1 loops=1)
                            ->  CTE Scan on pgrst_args pgrst_args_2  (cost=0.00..2.00 rows=100 width=4) (actual time=0.001..0.001 rows=1 loops=1)
Planning Time: 0.304 ms
Execution Time: 71.051 ms
```

(Steps to reproduce: Include a minimal SQL definition plus how you make the request to PostgREST and the response body)
```sql
-- seed.sql
create table if not exists
  user_notifications (
    user_id text not null,
    notification_id text not null,
    data jsonb not null,
    fs_updated_time timestamp not null,
    primary key (user_id, notification_id)
  );

create index if not exists user_notifications_unseen_created_time_idx on user_notifications (
  user_id,
  ((data->'isSeen')::boolean),
  ((data->'createdTime')::bigint) desc
    );

create
or replace function get_notifications (uid text, unseen_only boolean, max_num integer) returns setof user_notifications language sql stable parallel SAFE as $$
select *
from user_notifications as n
where n.user_id = uid and (not unseen_only or not ((n.data->'isSeen')::boolean))
order by ((n.data->'createdTime')::bigint) desc
limit max_num
$$;
```
```js
-- notifications.ts

// db initialized like:
// import {  createClient } from '@supabase/supabase-js'
// const db = createClient(supabaseInstanceId, supabaseAnonKey) as SupabaseClient

export async function getUnseenNotifications(
  db: SupabaseClient,
  userId: string,
  limit: number
) {
  const explanation = await db
    .rpc(`get_notifications`, {
      uid: userId,
      unseen_only: true,
      max_num: limit,
    })
    .explain({ analyze: true })
  console.log('notifications explanation:', explanation)
}

### PR DESCRIPTION
**Pull request by wolfgangwalther**
Previously they were passed as a JSON payload. This results in a LATERAL join for the calling expression, which prevents LIMIT from being pushed into the inlined function call, making some requests really slow.

Resolves #2858

I'm actually quite surprised that all tests were still passing when I made the build pass. I have not tried to confirm that this makes the respective RPC calls faster, but I'm pretty certain it should do so. I observed the same thing as @steve-chavez did in https://github.com/PostgREST/postgrest/issues/2858#issuecomment-1629999941 in one of my projects.

### DIFF
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE RecordWildCards       #-}
 {-|
 Module      : PostgREST.Query.QueryBuilder
 Description : PostgREST SQL queries generating functions.
@@ -16,8 +17,11 @@ module PostgREST.Query.QueryBuilder
   , limitedQuery
   ) where
 
+import qualified Data.Aeson                      as JSON
 import qualified Data.ByteString.Char8           as BS
+import qualified Data.HashMap.Strict             as HM
 import qualified Hasql.DynamicStatements.Snippet as SQL
+import qualified Hasql.Encoders                  as HE
 
 import Data.Maybe (fromJust)
 import Data.Tree  (Tree (..))
@@ -190,14 +194,19 @@ mutatePlanToQuery (Delete mainQi logicForest range ordts returnings)
     (whereRangeIdF, rangeIdF) = mutRangeF mainQi (cfName . coField <$> ordts)
 
 callPlanToQuery :: CallPlan -> PgVersion -> SQL.Snippet
-callPlanToQuery (FunctionCall qi params args returnsScalar returnsSetOfScalar returnsCompositeAlias returnings) pgVer =
+callPlanToQuery (FunctionCall qi params arguments returnsScalar returnsSetOfScalar returnsCompositeAlias returnings) pgVer =
   "SELECT " <> (if returnsScalar || returnsSetOfScalar then "pgrst_call.pgrst_scalar" else returnedColumns) <> " " <>
   fromCall
   where
+    jsonArgs = case arguments of
+      DirectArgs args -> Just $ JSON.encode args
+      JsonArgs json   -> json
     fromCall = case params of
-      OnePosParam prm -> "FROM " <> callIt (singleParameter args $ encodeUtf8 $ ppType prm)
+      OnePosParam prm -> "FROM " <> callIt (singleParameter jsonArgs $ encodeUtf8 $ ppType prm)
       KeyParams []    -> "FROM " <> callIt mempty
-      KeyParams prms  -> fromJsonBodyF args ((\p -> CoercibleField (ppName p) mempty False (ppTypeMaxLength p) Nothing Nothing) <$> prms) False True False <> ", " <>
+      KeyParams prms  -> case arguments of
+        DirectArgs args -> "FROM " <> callIt (fmtArgs prms args)
+        JsonArgs json   -> fromJsonBodyF json ((\p -> CoercibleField (ppName p) mempty False (ppTypeMaxLength p) Nothing Nothing) <$> prms) False True False <> ", " <>
                          "LATERAL " <> callIt (fmtParams prms)
 
     callIt :: SQL.Snippet -> SQL.Snippet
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by steve-chavez** [comment]
> The new query looks like this

Looks good! I think the plan cost should improve with that new query. If so, perhaps we can test it on `PlanSpec.hs`?

File: src/PostgREST/Query/QueryBuilder.hs [hunk 1]
```diff
@@ -209,6 +218,23 @@ callPlanToQuery (FunctionCall qi params args returnsScalar returnsSetOfScalar re
...
+          " := " <>
+          encodeArg (HM.lookup ppName args) <>
+          "::" <>
+          SQL.sql (encodeUtf8 ppTypeMaxLength)
+        encodeArg :: Maybe RpcParamValue -> SQL.Snippet
+        encodeArg (Just (Variadic v)) = SQL.encoderAndParam (HE.nonNullable $ HE.foldableArray $ HE.nonNullable HE.text) v
+        encodeArg (Just (Fixed v)) = SQL.encoderAndParam (HE.nonNullable HE.unknown) $ encodeUtf8 v
+        -- Currently not supported: Calling functions without some of their arguments without DEFAULT.
+        -- We could fallback to providing this NULL value in those cases.
+        encodeArg Nothing = "NULL"
```
steve-chavez, line 236: How about a test for this case? Would help in understanding and proving the limitation

steve-chavez: > But you could argue that if there is no exact match, then we can just fallback to partial matches and call rpc1(1, NULL) in this case

Oh, this might be a problem for typing and OpenAPI. cc @laurenceisla 

Also not sure if we want that behavior. In postgres, doing:

```sql
-- postgrest-with-postgresql-15  -f test/spec/fixtures/load.sql psql -U postgres

select add_them(a:=3);
ERROR:  function add_them(a => integer) does not exist
LINE 1: select add_them(a:=3);
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

-- select add_them(a:=3, b:=9); succeeds
```

Results in an error. The query string should be equivalent to named arguments, I believe.

steve-chavez (approved):

PR #3646, opened 2024-07-09
Backport to v12

wolfgangwalther:
Backports the fixes from #3639 and #3644 to v12.

I think we should make it a policy to backport fix commits immediately after merging them to main. We don't need to release on the backbranch immediately, but we should at least open the PR with the backport. This is because at this time, the changes are much more recent and easier to deal with in case they fail tests or so on the backbranch.


steve-chavez:
> I think we should make it a policy to backport fix commits immediately after merging them to main

Fully agree with that.

steve-chavez:
Any reason why shouldn't we release a new patch version now? The fixes and the perf increase are really good.

PR #3654, opened 2024-07-11
fix: schema cache retrying without backoff

steve-chavez:
Fixes https://github.com/PostgREST/postgrest/issues/3523.

Now if there's a failure when obtaining the pg version OR schema cache, we do the same retrying process. This way we don't add two retries.

Refactors and renames the "connectionWorker" to "schemaCacheLoader". This makes more sense since what we really want is the schema cache, the version is the pre-requisite for ensuring our schema cache queries work.

Additionally, we no longer log ` Attempting to connect to the database...` at startup unnecessarily. This is only logged whenever there's a retry attempt.

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -24,8 +24,8 @@ module PostgREST.AppState
   , putPgVersion
   , putIsListenerOn
   , usePool
-  , reReadConfig
-  , connectionWorker
+  , readInDbConfig
+  , schemaCacheLoader
   , getObserver
   , isLoaded
   , isPending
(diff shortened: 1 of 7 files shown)


test/io/test_io.py
```diff
@@ -853,7 +853,7 @@ def test_metrics_include_schema_cache_fails(defaultenv, metapostgrest):
                 r'pgrst_schema_cache_loads_total{status="FAIL"} (\d+)', response.text
             ).group(1)
         )
-        assert metrics > 3.0
+        assert metrics == 1.0
```
steve-chavez, line 856: This test proves that this works. I didn't want to rely on the test mentioned in https://github.com/PostgREST/postgrest/issues/3523#issuecomment-2215053565 because it relies on a new feat and I want to backport this fix on v12.

Manually it can be confirmed that the above is fixed though.

```
$ PGRST_DB_SCHEMAS="x" postgrest-with-postgresql-15  -f test/spec/fixtures/load.sql postgrest-run

10/Jul/2024:20:56:36 -0500: Starting PostgREST 12.3 (pre-release)...
10/Jul/2024:20:56:36 -0500: Listening on port 3000
10/Jul/2024:20:56:36 -0500: Listening for notifications on the "pgrst" channel
10/Jul/2024:20:56:36 -0500: Successfully connected to PostgreSQL 15.7 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit
10/Jul/2024:20:56:36 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"x\" does not exist"}

10/Jul/2024:20:56:37 -0500: Attempting to reconnect to the database in 1 seconds...
10/Jul/2024:20:56:37 -0500: Successfully connected to PostgreSQL 15.7 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit
10/Jul/2024:20:56:37 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"x\" does not exist"}

10/Jul/2024:20:56:39 -0500: Attempting to reconnect to the database in 2 seconds...
10/Jul/2024:20:56:39 -0500: Successfully connected to PostgreSQL 15.7 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit
10/Jul/2024:20:56:39 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"x\" does not exist"}
```

steve-chavez:
Some failures on master CI.. that's strange:

- https://github.com/PostgREST/postgrest/actions/runs/9884375643/job/27300648912
- https://github.com/PostgREST/postgrest/actions/runs/9884375643/job/27300649127

src/PostgREST/AppState.hs
```diff
@@ -339,167 +326,120 @@ getSocketAdmin = stateSocketAdmin
...
+  return $ scacheStatus == SCLoaded && connEstablished
 
 isPending :: AppState -> IO Bool
 isPending x = do
   scacheStatus <- readIORef $ stateSCacheStatus x
-  connStatus <- readIORef $ stateConnStatus x
-  listenerOn <- getIsListenerOn x
-  return $ scacheStatus == SCPending || connStatus == ConnPending || not listenerOn
+  connEstablished <- isConnEstablished x
+  return $ scacheStatus == SCPending || not connEstablished
```
steve-chavez, line 350: I think it could be this change of Admin conditions for the `/ready` state. 

The logic should be the same.. but perhaps the previous one was faster. But I cannot reproduce the CI failures locally.

Issue #3504, opened 2024-05-10
Row Filter parameter format is displayed as `integer`


dantheman2865:
Hello, I've been using the OpenAPI spec in tandem with a code generator for Typescript files. I have found that the `rowFilter` parameters are generating incorrectly. For a table like:
```
CREATE TABLE IF NOT EXISTS api.providers
(
    id integer NOT NULL DEFAULT nextval('api.providers_id_seq'::regclass),
    name character varying(256) COLLATE pg_catalog."default" NOT NULL,
    active boolean NOT NULL DEFAULT true,
    created_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP,
    updated_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP,
    CONSTRAINT providers_pkey PRIMARY KEY (id)
)
```
It makes sense that `definitions.providers.properties.id` is:
```
{
    "description": "Note:\nThis is a Primary Key.<pk/>",
    "format": "integer",
    "type": "integer"
},
```
HOWEVER, the definition for `rowFilter` is wrong:
```
"rowFilter.providers.id": {
    "name": "id",
    "required": false,
 -->"format": "integer",<--
    "in": "query",
    "type": "string"
},
```
Because, the query syntax for filtering is `id=eq.<number>`, NOT `id=<number>`.

I believe it should omit the `format` field as below:
```
"rowFilter.providers.id": {
    "name": "id",
    "required": false,
    "in": "query",
    "type": "string"
},
```
Thanks!

laurenceisla:
Closed in #3661

Issue #1031, opened 2017-12-01
OpenAPI format uuid


danilobuerger:
If I have a uuid column it appears like 
```
"rowFilter.test.id": {
	"name": "id",
	"required": false,
	"format": "uuid",
	"in": "query",
	"type": "string"
},
```
in the spec. However when using with something like swagger ui, I can't enter the uuid with an operator:
```
eq.2e094f21-7a6f-4268-b1ff-c2a376de35ad
```
This will throw an error in swagger ui (as intended) as it doesn't match the format.

begriffs:
Thanks for reporting. What is the format that Swagger-UI is expecting for UUID?

danilobuerger:
Looking at the source code I am guessing its this:
```
export const validateGuid = (val) => {
    if (!/^[{(]?[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}[)}]?$/.test(val)) {
        return "Value must be a Guid"
    }
}
```

danilobuerger:
@begriffs any news on this? Wouldn't it be enough to just remove the `"format": "uuid"` then we could enter any string.

begriffs:
> However when using with something like swagger ui, I can't enter the uuid with an operator:
> `eq.2e094f21-7a6f-4268-b1ff-c2a376de35ad`

Oh, because the `eq.` part at the beginning is considered part of the value? That's annoying. Wouldn't the same logic apply to types like integers too since `eq.42` is not itself an int?

danilobuerger:
I just checked with `smallint`, those get rendered as 
```"field": {
  "format": "smallint",
  "type": "string"
},```
Which does accept something like `eq.42` as its type is string and a format called `smallint` doesn't exist.

sarumont:
@begriffs Your comment above is correct - the OpenAPI spec [leaves room for interpretation](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types) of the `format` field on a parameter, but there are some well-defined `format` values as well as some conventionally used ones (i.e. `uuid`).

The fact that `swagger-ui` accepts `eq.42` in a `smallint` field is probably due to the fact that there is no Swagger-defined (or generally accepted) `string / smallint` combo.

[...]

PR #3661, opened 2024-07-15
fix: Remove OpenAPI format for rowFilter params

dantheman2865:
See #3504 for more details.

When using OpenAPI docs with some code generators, rowFilter parameters will be incorrectly typed. Because Row Filter parameters are always a string (e.g. `eq.1001`), the `format` field should not be included.

Note: I could not see any tests to update for this case. Please let me know if I should add a test (presumably to `test/spec/Feature/OpenApi/OpenApiSpec.hs`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
  - #3091, Broken link in OpenAPI description `externalDocs` - @salim-b
  - #3659, Embed One-to-One relationship with different column order properly - @wolfgangwalther
+ - #3504, Eliminate `format` parameter on `rowFilter` parameters in OpenAPI - @dantheman2865
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+ - #3504, Eliminate `format` parameter on `rowFilter` parameters in OpenAPI - @dantheman2865
```
laurenceisla, line 8:
```suggestion
 - #3504, Remove `format` from `rowFilter` parameters in OpenAPI - @dantheman2865
```

This was OK under "Fixed".

laurenceisla:
Thanks for the PR!

> Note: I could not see any tests to update for this case. Please let me know if I should add a test (presumably to `test/spec/Feature/OpenApi/OpenApiSpec.hs`).

That's OK, no need for a test here.

> the `format` field should not be included.

I agree that it's not correct to specify this as `format`. I just noticed that this is related to issue #1031 where it was also suggested to remove it: https://github.com/PostgREST/postgrest/issues/1031#issuecomment-557100774.

We would still need to document the format of the filters, but I would consider that another feature, while this PR is fixing an incorrect `format` issue. So it's a  :+1: from me.

CHANGELOG.md
dantheman2865: I pushed a incorporating this feedback and moving the message back to "Fixed". Thanks!

laurenceisla (approved):

PR #3685, opened 2024-08-08
feat: control tabel/view currentuser insert, update,delete rights

jyritali:
feature: control tabel/view currentuser insert, update,delete rights.
add to PostgREST - user tabel and view rights control and show methods reagarding user rights, not table or view one

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-
+- #3684 - control tabel/view insert,update,delete rights regarding currentuser rights
+  
 ### Added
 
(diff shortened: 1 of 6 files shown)


jyritali:
Please merge these commits to one commit

laurenceisla:
I remember tackling this in PR #1824, we had the following discussion there: https://github.com/PostgREST/postgrest/pull/1824#discussion_r616409225

In summary, the HTTP methods show what CAN be done to the table/view, not the privileges granted to them.

https://github.com/PostgREST/postgrest/blob/6c7963c1c4b419bff14f1f6e5df17f57f3070897/src/PostgREST/SchemaCache/Table.hs#L30-L33

This may be different for OpenAPI though, but we didn't follow up with it because we were in the plans of removing it from the core repo.

All in all, thanks for the initiative, but these changes are not compatible with what we agreed in the PR I mentioned.

PR #3640, opened 2024-07-05
feat: WIP allow spread operators in to-many relationships

laurenceisla:
WIP on #3041

PR #3693, opened 2024-08-20
Fixes when using aggregates inside Spread Embedded Resources

laurenceisla:
Missing:

- [x] Tests
- [x] Refactor
- [x] Changelog

These fixes are also needed for https://github.com/PostgREST/postgrest/pull/3640

### DIFF
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -332,8 +332,8 @@ readPlan qi@QualifiedIdentifier{..} AppConfig{configDbMaxRows, configDbAggregate
   in
     mapLeft ApiRequestError $
     treeRestrictRange configDbMaxRows (iAction apiRequest) =<<
-    validateAggFunctions configDbAggregates =<<
     hoistSpreadAggFunctions =<<
+    validateAggFunctions configDbAggregates =<<
     addRelSelects =<<
     addNullEmbedFilters =<<
     validateSpreadEmbeds =<<
@@ -721,8 +721,7 @@ hoistIntoRelSelectFields _ r = r
 
 validateAggFunctions :: Bool -> ReadPlanTree -> Either ApiRequestError ReadPlanTree
 validateAggFunctions aggFunctionsAllowed (Node rp@ReadPlan {select} forest)
-  | aggFunctionsAllowed = Node rp <$> traverse (validateAggFunctions aggFunctionsAllowed) forest
-  | any (isJust . csAggFunction) select = Left AggregatesNotAllowed
+  | not aggFunctionsAllowed && any (isJust . csAggFunction) select = Left AggregatesNotAllowed
   | otherwise = Node rp <$> traverse (validateAggFunctions aggFunctionsAllowed) forest
 
 addFilters :: ResolverContext -> ApiRequest -> ReadPlanTree -> Either ApiRequestError ReadPlanTree
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -270,7 +270,7 @@ data ResolverContext = ResolverContext
   }
 
 resolveColumnField :: Column -> CoercibleField
-resolveColumnField col = CoercibleField (colName col) mempty False (colNominalType col) Nothing (colDefault col)
+resolveColumnField col = CoercibleField (colName col) mempty False (colNominalType col) Nothing (colDefault col) False
 
 resolveTableFieldName :: Table -> FieldName -> CoercibleField
 resolveTableFieldName table fieldName =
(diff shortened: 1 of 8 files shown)


src/PostgREST/Plan.hs
```diff
@@ -376,15 +377,24 @@ initReadRequest ctx@ResolverContext{qi=QualifiedIdentifier{..}} =
...
+      | isJust alias = field
+      | isJust aggFun = fieldAliasForSpreadAgg isSpread field
       | isJsonKeyPath fieldDetails, Just key <- lastJsonKey fieldDetails = field { csAlias = Just key }
       | isTransformPath fieldDetails = field { csAlias = Just (cfName fieldDetails) }
       | otherwise = field
 
+    fieldAliasForSpreadAgg isSpread field@CoercibleSelectField{csField=fieldDetails, csAggFunction=aggFun} =
+      -- A request like: `/top_table?select=...middle_table(...nested_table(count()))` will `SELECT` the full row instead of `*`,
+      -- because doing a `COUNT(*)` in `top_table` would not return the desired results.
+      -- So we use the "count" alias if none is present since the field name won't be selected.
```
laurenceisla, line 393: I used "count" as an alias because PostgreSQL also returns that name when no alias is given for `SELECT COUNT(...)`. This would conflict when a table has a `count` column or when selecting a `count()` in each nested resource, so the solution would be to use aliases. This is not a breaking change since those requests are not working right now in v12.2.3:

```bash
curl 'localhost:3000/process_supervisor?select=...processes(count,...process_costs(count()))'
curl 'localhost:3000/process_supervisor?select=...processes(count(),...process_costs(count()))'
```
```json
{"code":"42803","details":null,"hint":null,"message":"aggregate functions are not allowed in GROUP BY"}
```

With these changes, the error is much nicer now:

```bash
curl 'localhost:3000/process_supervisor?select=...processes(count,...process_costs(count()))'
```
```json
{"code":"42702","details":null,"hint":null,"message":"column reference \"count\" is ambiguous"}
```

test/spec/Feature/Query/AggregateFunctionsSpec.hs
```diff
@@ -152,6 +152,131 @@ allowed =
...
+              {"factory": 2, "process_costs_count": 4},
+              {"factory": 3, "process_costs_count": 1}]|]
+            { matchHeaders = [matchContentTypeJson] }
+        it "works on different levels of the nested resources at the same time" $
+          get "/process_supervisor?select=...processes(factory:factory_id,processes_count:count(),...process_costs(process_costs_count:count()))&order=processes(factory)" `shouldRespondWith`
+            [json|[
+              {"factory": 1, "processes_count": 2, "process_costs_count": 2},
+              {"factory": 2, "processes_count": 4, "process_costs_count": 4},
+              {"factory": 3, "processes_count": 2, "process_costs_count": 1}]|]
+            { matchHeaders = [matchContentTypeJson] }
```
laurenceisla, line 279: This test illustrates why to use `COUNT(<row>)` instead of `COUNT(*)`. Here, there's a `process` that does not have a `process_cost` assigned. If we used `COUNT(*)` for the `process_costs` table, then the top level resource would not have an idea on which are the relevant columns to count from the intermediate resource (`processes`). Passing the full row solves this issue.

laurenceisla (commented):
For Spread Resources I decided to use `COUNT(<row>)` instead of `COUNT(*)`. When there are nested spread resources, doing a `COUNT(*)` may not return the expected values, since it would count all the columns `*` from the intermediate resources.

For example, for this query:

```bash
curl 'localhost:3000/process_supervisor?select=...processes(factory_id,...process_costs(count()))'
```

The resulting SQL would be:

```sql
SELECT "process_supervisor_processes_1"."factory_id",
       COUNT("process_supervisor_processes_1"."count") AS "count"
FROM "test"."process_supervisor"
LEFT JOIN LATERAL
  (SELECT "processes_1"."factory_id",
          "processes_process_costs_2"."count"
   FROM "test"."processes" AS "processes_1"
   LEFT JOIN LATERAL
     (SELECT "process_costs_2" AS "count"
      FROM "test"."process_costs" AS "process_costs_2"
      WHERE "process_costs_2"."process_id" = "processes_1"."id") AS "processes_process_costs_2" ON TRUE
   WHERE "processes_1"."id" = "test"."process_supervisor"."process_id") AS "process_supervisor_processes_1" ON TRUE
GROUP BY "process_supervisor_processes_1"."factory_id"
```

Using `COUNT("process_supervisor_processes_1".*)`  would not count correctly since there are `processes` that do not have `process_count` assigned.

steve-chavez (approved):
LGTM!

laurenceisla:
Did not backport the fixes to v12 because the last one depends on a feature/performance commit, so it has a conflict with the code there.

PR #3703, opened 2024-08-22
Update functions.rst

jasonmclose:
This change adds an explanation of how to handle an array of JSON objects within an RPC call.  To pass multiple objects, an array of JSON objects must be the JSON value, with the key being the json or jsonb variable name of the Postgres function.

For people who want to perform multiple tasks/inserts/updates within a single API call, this is a needed explanation for that use-case.

I recently had this problem, and after solving it on my own, I realized that I would have been saved several hours of troubleshooting had this simple example been included in the API documentation.

### DIFF
--- a/docs/references/api/functions.rst
+++ b/docs/references/api/functions.rst
@@ -71,6 +71,40 @@ The function parameter names match the JSON object keys in the POST case, for th
 
 .. _function_single_json:
 
+Functions with an array of JSON objects
+----------------------------------------------
+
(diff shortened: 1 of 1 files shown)


steve-chavez (approved):
This is very useful. Many thanks for the contribution!

steve-chavez:
Cherry-picked this into the main branch https://github.com/PostgREST/postgrest/commit/f21053dbee524bc12fe3e68fcee687c4a31a962d

PR #3750, opened 2024-10-19
Fix: handle column default types correctly in SchemaCache

### LINKED ISSUE #3706: Doing a bulk insert with `missing=default` uses the default value of the domain instead of the column
### Environment

* PostgreSQL version: 15.7
* PostgREST version: 12.2.3
* Operating system: NixOS

### Description of issue

As the title says. According to [the PostgreSQL docs](https://www.postgresql.org/docs/current/sql-createdomain.html):

> The default expression will be used in any insert operation that does not specify a value for the column. If a default value is defined for a particular column, it overrides any default associated with the domain. In turn, the domain default overrides any default value associated with the underlying data type.

To reproduce this:

```sql
create domain defdom as integer default 1;

create table test_defdom (
  id int primary key,
  val api.defdom default 10
);
```
```bash
curl -X POST 'localhost:3000/test_defdom?columns=id,val' \
  -H 'Content-Type: application/json' \
  -H 'Prefer: missing=default, return=representation' \
  -d '[{"id": 1}]'
```

Expected:

```json
[{"id":1,"val":10}]
```

Got:
```json
[{"id":1,"val":1}]
```

### Solution

Check if the defaults of the columns are not null first, otherwise use the domain default instead.

https://github.com/PostgREST/postgrest/blob/f21053dbee524bc12fe3e68fcee687c4a31a962d/src/PostgREST/SchemaCache.hs#L624-L629

mursidx:
This pull request addresses issue #3706 by improving the handling of column default values in the SchemaCache. 
The previous implementation did not correctly account for certain data types, which could lead to incorrect metadata retrieval. 
This fix ensures that all relevant cases are covered, including scenarios where `ad.adbin` is not NULL.

Additionally, an entry has been added to the CHANGELOG.md for this fix.

Fixes #3706

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1095,3 +1095,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Make filter position match docs, e.g. `?order=col.asc` rather
   than `?order=asc.col`.
+
+
+### Unreleased
+- **fix**: Improved handling of column default values in `SchemaCache`. This addresses issue #3706 by ensuring all relevant cases are covered, including scenarios where `ad.adbin` is not NULL.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-
+### Fixed
+- **fix**: Improved handling of column default values in `SchemaCache`. This addresses issue #3706 by ensuring all relevant cases are covered, including scenarios where `ad.adbin` is not NULL.
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1095,3 +1095,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Make filter position match docs, e.g. `?order=col.asc` rather
   than `?order=asc.col`.
+
+
+### Unreleased
+- **fix**: Improved handling of column default values in `SchemaCache`. This addresses issue #3706 by ensuring all relevant cases are covered, including scenarios where `ad.adbin` is not NULL.
```
laurenceisla, line 1101: There is already an "Unreleased"  section at the beginning of the document. The entry should go under the "Fixed" subtitle. Check the other entries for the format.

laurenceisla (commented):
Thanks for the PR. Here are my observations:

mursidx:
@laurenceisla ,
I have updated the SchemaCache.hs to prioritize column defaults as suggested. I also added a corresponding test case in InsertSpec.hs to verify the functionality and updated the CHANGELOG.md accordingly.

PR #3757, opened 2024-10-27
remove support for Prefer: params=single-object

joelonsql:
Using this preference was deprecated in 6c3d7a9, in favor of Functions with an array of JSON objects.

This change removes support for this feature, planned for the next major version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Removed
+
+ - #????, Remove support for `Prefer: params=single-object` - @joelonsql
+   + This preference was deprecated in favor of Functions with an array of JSON objects
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


steve-chavez (approved):
Nice work :+1:

steve-chavez:
Ah, my bad, I thought it was addressed. How about this?

- I revert the previous commit.
- Edit the CHANGELOG.
- Make a commit with the `break: ..` prefix. We had some of these in the past.

All in one push to main?

steve-chavez:
> Also, @steve-chavez, I think we loosely agreed on some form of conventional commits, i.e. using prefixes. The squashed commit message has none.

Btw, maybe we can enforce this somehow? https://dev.to/mbarzeev/a-git-hook-for-commit-messages-validation-no-husky-just-js-1hni

steve-chavez:
> We'd have to think about the proper commit prefix for something like this. I used feat: for the old PG versions removal, but maybe we should just introduce remove:?

`remove:` LGTM. 

Fixed the changelog on https://github.com/PostgREST/postgrest/commit/b821857861301aecdb897c57a3c701e04aa4e837

PR #3761, opened 2024-10-29
docs: fix deprecation of analytics in RTD

laurenceisla:
It seems to be working in the `latest` docs. This should make it work for the `stable` version.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -313,6 +313,12 @@ ogp_use_first_image = True
 ogp_enable_meta_description = True
 ogp_description_length = 300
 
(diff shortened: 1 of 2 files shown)


laurenceisla:
> Are you going to open PRs for v11 and v10 as well or should I backport it there, too?

Hmm... yes, I don't think there'd be an issue if backported to the v11 and v10 too. Go ahead.

laurenceisla:
> I tried enabling the addon, let's see whether this does anything to v12 docs without merging this PR first.

Ah, cool, let's check it out first. Strange that the RTD blog/changelog didn't mention that option.

PR #3775, opened 2024-11-10
Update nixpkgs and GHCs for cabal / stack, test with PG 17

wolfgangwalther:
Update and align everything. Let's see which parts pass and which don't.

### DIFF
--- a/nix/tools/cabalTools.nix
+++ b/nix/tools/cabalTools.nix
@@ -15,6 +15,7 @@ let
         withEnv = postgrest.env;
       }
       ''
+        ${cabal-install}/bin/cabal v2-update
         exec ${cabal-install}/bin/cabal v2-build ${devCabalOptions} "''${_arg_leftovers[@]}"
       '';
 
--- a/nix/tools/devTools.nix
+++ b/nix/tools/devTools.nix
@@ -246,6 +246,7 @@ let
       }
       ''
         mkdir -p "$_arg_dumpdir"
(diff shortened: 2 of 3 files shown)


test/spec/fixtures/schema.sql
```diff
@@ -257,7 +257,7 @@ CREATE FUNCTION varied_arguments(
   enum enum_menagerie_type,
   arr text[],
   "integer" integer default 42,
-  json json default '{}',
+  "json" json default '{}',
```
steve-chavez, line 260: So pg 17 finally fails on these identifiers? Or is it just a cleanup?

steve-chavez:
Q: I thought I saw GHC 9.10.1 in this PR before, why was it removed? Something broke?

I was hoping to do some experiments with Linear Types, GHC 9.10.1 added support for using it with let/where ([ref](https://discourse.haskell.org/t/ghc-9-10-1-is-now-available/9523)).

steve-chavez:
I think we discussed this before, but couldn't we just delete the `cabal.project.freeze`? I don't recall why we needed it if we have Nix.

develop7:
> I *think* this would silently pull in new dependencies for cabal-based builds, which we still do. This makes CI not reproducible anymore. Not sure whether that was the original reason why it was added, though.

It's exactly the reason I've introduced it in https://github.com/PostgREST/postgrest/pull/3004. Should've added a comment there too, though.

Also I wonder if I could fix these more explicitly by introducing upper bound in `cabal.project`

PR #3792, opened 2024-11-16
docs: replace broken links for older GUC settings

laurenceisla:
Use wayback machine links for the PostgREST v8.0 docs. The old info is needed for users that want to migrate to newer versions.


laurenceisla:
> Should we maybe just put the part about the "old style GUC handling" in the respective branch as well?

Yeah, that's better. Done that.

PR #3727, opened 2024-09-16
feat: log pool maximum size

steve-chavez:
It's important for observability to have an historic trace of the pool size. Currently we expose it on the metrics endpoint, but not all deployments use it.

This logs the pool size after the successful connection log to make it more visible:

```
<timestamp>: Connection Pool initialized with a maximum size of 4 connections
```

PR #3795, opened 2024-11-19
fix: clarify PGRST116 error message

steve-chavez:
Currently it's redundant.

```JSON
{"message":"JSON object requested, multiple (or no) rows returned",
"details":"The result contains 0 rows"}
```

Now:

```json
{"message":"Cannot coerce the result to a single JSON object",
"details":"The result contains 0 rows"}
```

Also correct docs which had a wrong error code.


steve-chavez:
Hm, I did a fix commit on https://github.com/PostgREST/postgrest/pull/3727/commits/da1150dc1a12580f702a02e6e8b3dd517904a5bb, which was about changing messages too.

> Is this really a fix? I don't think so, nothing is broken. I wouldn't backport it.

I do think it's a "fix", but not so critical to backport it. Definitely not a feature though, I would want to keep those clearly differentiated in commits and CHANGELOG entries because they correlate to breaking changes sometimes.

PR #3801, opened 2024-11-22
fix: jwt cache is not purged

### LINKED ISSUE #3788: JWT cache is not purged?
## Problem

The `Data.Cache` says:

> Note that items are not purged automatically in the background when they expire. You have to manually call [lookup](https://hackage.haskell.org/package/cache-0.1.3.0/docs/Data-Cache.html#v:lookup) to purge a single item, or call [purgeExpired](https://hackage.haskell.org/package/cache-0.1.3.0/docs/Data-Cache.html#v:purgeExpired) to purge all expired items.

But `purgeExpired` nor [purge](https://hackage.haskell.org/package/cache-0.1.3.0/docs/Data-Cache.html#v:purge) are called anywhere on the [Auth module](https://github.com/PostgREST/postgrest/blob/80a4edbd2dc5b7788f1d7b220cc9b084ba0fab5e/src/PostgREST/Auth.hs).

taimoorzaeem:
Fixes #3788.

Fixed the JWT cache not purging expired tokens in cache. This is tested by adding a new metric `pgrst_jwt_cache_size`.

@steve-chavez I can also divide this into two commits, one for adding the new metric and the other for bug fix. Please review and let me know.

Edit: I'll continue development once #3802 gets merged.

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -181,10 +181,33 @@ getJWTFromCache appState token maxLifetime parseJwt utc = do
   authResult <- maybe parseJwt (pure . Right) checkCache
 
   case (authResult,checkCache) of
-    (Right res, Nothing) -> C.insert' (getJwtCache appState) (getTimeSpec res maxLifetime utc) token res
+    -- From comment:
+    -- https://github.com/PostgREST/postgrest/pull/3801#discussion_r1857987914
+    --
+    -- We purge expired cache entries on a cache miss
+    -- The reasoning is that:
+    --
+    -- 1. We expect it to be rare (otherwise there is no point of the cache)
+    -- 2. It makes sure the cache is not growing (as inserting new entries
+    --    does garbage collection)
+    -- 3. Since this is time expiration based cache there is no real risk of
+    --    starvation - sooner or later we are going to have a cache miss.
+
+    (Right res, Nothing) -> do -- cache miss
+
+      let timeSpec = getTimeSpec res maxLifetime utc
+
+      -- purge expired cache entries
+      C.purgeExpired jwtCache
+
+      -- insert new cache entry
+      C.insert' jwtCache timeSpec token res
+
     _                    -> pure ()
 
   return authResult
+    where
+      jwtCache = getJwtCache appState
 
 -- Used to extract JWT exp claim and add to JWT Cache
 getTimeSpec :: AuthResult -> Int -> UTCTime -> Maybe TimeSpec
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1653,7 +1653,7 @@ def test_schema_cache_startup_load_with_in_db_config(defaultenv, metapostgrest):
 def test_jwt_cache_purges_expired_entries(defaultenv):
     "test expired cache entries are purged on cache miss"
 
-    # The verification of actual cache size reduction is done locally
+    # The verification of actual cache size reduction is done manually, see https://github.com/PostgREST/postgrest/pull/3801#issuecomment-2620776041
     # This test is written for code coverage of purgeExpired function
 
     relativeSeconds = lambda sec: int(
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1648,3 +1648,51 @@ def test_schema_cache_startup_load_with_in_db_config(defaultenv, metapostgrest):
     response = metapostgrest.session.post("/rpc/reset_db_schemas_config")
     assert response.text == ""
     assert response.status_code == 204
+
+
+def test_jwt_cache_purges_expired_entries(defaultenv):
+    "test expired cache entries are purged on cache miss"
+
+    # The verification of actual cache size reduction is done locally
+    # This test is written for code coverage of purgeExpired function
+
+    relativeSeconds = lambda sec: int(
+        (datetime.now(timezone.utc) + timedelta(seconds=sec)).timestamp()
+    )
+
+    headers = lambda sec: jwtauthheader(
+        {"role": "postgrest_test_author", "exp": relativeSeconds(sec)},
+        SECRET,
+    )
+
+    env = {
+        **defaultenv,
+        "PGRST_JWT_CACHE_MAX_LIFETIME": "86400",
+        "PGRST_JWT_SECRET": SECRET,
+        "PGRST_DB_CONFIG": "false",
+    }
+
+    with run(env=env) as postgrest:
+
+        # Generate two unique JWT tokens
+        # The 1 second sleep is needed for it generate a unique token
+        hdrs1 = headers(5)
+        postgrest.session.get("/authors_only", headers=hdrs1)
+
+        time.sleep(1)
+
+        hdrs2 = headers(5)
+        postgrest.session.get("/authors_only", headers=hdrs2)
+
+        # Wait 5 seconds for the tokens to expire
+        time.sleep(5)
+
+        hdrs3 = headers(5)
+
+        # Make another request which should cause a cache miss and so
+        # the purgeExpired function will be triggered.
+        #
+        # This should remove the 2 expired tokens but adds another to cache
+        response = postgrest.session.get("/authors_only", headers=hdrs3)
+
+        assert response.status_code == 200
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3795, Clarify `Accept: vnd.pgrst.object` error message - @steve-chavez
  - #3779, Always log the schema cache load time - @steve-chavez
  - #3706, Fix insert with `missing=default` uses default value of domain instead of column - @taimoorzaeem
+ - #3788, Fix jwt cache does not remove expired entries - @taimoorzaeem
 
 ### Changed
 


test/io/test_io.py
```diff
@@ -1648,3 +1650,58 @@ def test_schema_cache_startup_load_with_in_db_config(defaultenv, metapostgrest):
...
+
+        time.sleep(1)
+
+        hdrs3 = headers(5)
+        postgrest.session.get("/authors_only", headers=hdrs3)
+
+        # the cache should now have three tokens
+        response = postgrest.admin.get("/metrics")
+        assert response.status_code == 200
+        assert "pgrst_jwt_cache_size 3.0" in response.text
```
steve-chavez, line 1694: @taimoorzaeem  Is it possible to have the size measured on bytes?

taimoorzaeem: Maybe. There is no built-in function for this in `Data.Cache`. I have tried using [https://hackage.haskell.org/package/ghc-datasize](https://hackage.haskell.org/package/ghc-datasize) to calculate size in bytes at runtime, but I have run into problems with that too.

We can try manually writing some logic to calculate the size in bytes, but I guess that would require some effort.

src/PostgREST/Auth.hs
```diff
@@ -163,14 +164,25 @@ middleware appState app req respond = do
 -- | Used to retrieve and insert JWT to JWT Cache
 getJWTFromCache :: AppState -> ByteString -> Int -> IO (Either Error AuthResult) -> UTCTime -> IO (Either Error AuthResult)
 getJWTFromCache appState token maxLifetime parseJwt utc = do
-  checkCache <- C.lookup (getJwtCache appState) token
+
+  -- Call first to delete the expired JWTs
+  C.purgeExpired jwtCache
```
mkleczek, line 169: I am wondering whether it is a good idea to add this in the hot path of request processing unless we are sure it is a constant time operation (and fast).

Otherwise it should be moved to a separate thread.

src/PostgREST/Auth.hs
```diff
@@ -163,14 +164,25 @@ middleware appState app req respond = do
 -- | Used to retrieve and insert JWT to JWT Cache
 getJWTFromCache :: AppState -> ByteString -> Int -> IO (Either Error AuthResult) -> UTCTime -> IO (Either Error AuthResult)
 getJWTFromCache appState token maxLifetime parseJwt utc = do
-  checkCache <- C.lookup (getJwtCache appState) token
+
+  -- Purge expired tokens in a separate thread
+  _ <- forkIO $ C.purgeExpired jwtCache
```
steve-chavez, line 169: Hm, this is going to start a new thread on every request? That doesn't look efficient.

I think we should have a background thread that is notified for doing the purge.

mkleczek: To be honest - for starters I would implement a simple solution that executes purgeExpired periodically ( frequency subject to configuration possibly - but not necessarily ).

Triggering purge upon every request might mean the purging thread is running constantly wasting cycles as there is nothing to do most of the times.

The ultimate solution would be to have some kind of scheduler that triggers purging upon next nearest expiry - but that would require more complex bookkeeping.

mkleczek: Thinking about it some more - it seems to me the best moment to trigger purge is upon a **cache miss** - just before/after inserting a new entry.

The reasoning is that:

1. We expect it to be rare (otherwise there is no point of the cache)
2. It makes sure the cache is not growing (as inserting new entries does garbage collection)
3. Since this is time expiration based cache there is no real risk of starvation - sooner or later we are going to have a cache miss.

taimoorzaeem: This seems like a much better idea. This way, we wouldn't need to do maintain any new thread state, hence avoiding extra overhead. @steve-chavez WDYT?

test/io/test_io.py
steve-chavez: Btw, this metric could be a new PR so it's easier to review.

src/PostgREST/Auth.hs
steve-chavez: Agree, looks like great idea :+1:

mkleczek: @taimoorzaeem not sure how to help with that - do you think it might be ready soon? Seems to me without it JWT caching is kind of useless (ie. has memory leak). I think it is a pretty good candidate for quick patch release.

taimoorzaeem: @mkleczek Yes, I agree that JWT caching is not production ready yet. @steve-chavez WDYT?

steve-chavez: Yes, we should definitely do a patch release. We need to merge https://github.com/PostgREST/postgrest/pull/3802 first for testing, I'll review it tomorrow.

steve-chavez:
Looking at the latest loadtest results https://github.com/PostgREST/postgrest/actions/runs/13016585172. There's a dropdown in perf wrt v12.2.5, but this is because of https://github.com/PostgREST/postgrest/issues/3882, compared to `main` the numbers are roughly the same, so no perf loss in this PR.

steve-chavez:
I'm trying to manually test this by cherry-picking https://github.com/PostgREST/postgrest/pull/3802 (which won't be merged before this because https://github.com/PostgREST/postgrest/pull/3802#issuecomment-2619472741).

So then I do:

```bash
curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt --exp 1 postgrest_test_author)"
# .. 3 more times

# then
curl localhost:3001/metrics
# HELP pgrst_jwt_cache_size_bytes The JWT cache size in bytes
# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 480.0
```

I wait for like a minute, do new requests and the jwt cache size never goes down.

```
curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt --exp 1 postgrest_test_author)"
# 3 more times

curl localhost:3001/metrics
# HELP pgrst_jwt_cache_size_bytes The JWT cache size in bytes
# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 840.0
```

@taimoorzaeem Can you double check this and see what's the problem?

steve-chavez:
For some reason, doing the request with the same JWT also increase the jwt cache size every time:

```
export JWT=$(postgrest-gen-jwt --exp 1 postgrest_test_author)

curl localhost:3000/authors_only -H "Authorization: Bearer $JWT"
[]

curl localhost:3001/metrics
# HELP pgrst_jwt_cache_size_bytes The JWT cache size in bytes
# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 120.0

curl localhost:3000/authors_only -H "Authorization: Bearer $JWT"
[]

curl localhost:3001/metrics
# HELP pgrst_jwt_cache_size_bytes The JWT cache size in bytes
# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 240.0

curl localhost:3000/authors_only -H "Authorization: Bearer $JWT"
[]
curl localhost:3001/metrics
# HELP pgrst_jwt_cache_size_bytes The JWT cache size in bytes
# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 360.0
```

And that shouldn't happen at all

taimoorzaeem:
@steve-chavez Oh, sorry. This happened because I avoided recalculating entire cache size for performance reasons in #3802. So, the cache size was only ever increasing (because there was no purging yet). However, now that our fix workflow has changed. I have fixed the metric accordingly.

taimoorzaeem:
Now we can check that the cache size goes down. Doing 3 request with different jwt with 5 second jwt expiry:
```bash
curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt --exp 5 postgrest_test_author)"
# wait 1 sec
curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt --exp 5 postgrest_test_author)"
# wait 1 sec
curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt --exp 5 postgrest_test_author)"

# now check metrics
curl localhost:3001/metrics

# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 360.0

# now wait at most 5 seconds for JWTs to expire and do another call
curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt --exp 5 postgrest_test_author)"

# check metrics again
curl localhost:3001/metrics
# TYPE pgrst_jwt_cache_size_bytes gauge
pgrst_jwt_cache_size_bytes 120.0
```

test/io/test_io.py
```diff
@@ -1648,3 +1648,51 @@ def test_schema_cache_startup_load_with_in_db_config(defaultenv, metapostgrest):
     response = metapostgrest.session.post("/rpc/reset_db_schemas_config")
     assert response.text == ""
     assert response.status_code == 204
+
+
+def test_jwt_cache_purges_expired_entries(defaultenv):
+    "test expired cache entries are purged on cache miss"
+
+    # The verification of actual cache size reduction is done locally
```
steve-chavez, line 1656:
```suggestion
    # The verification of actual cache size reduction is done manually, see https://github.com/PostgREST/postgrest/pull/3801#issuecomment-2620776041
```

steve-chavez (approved):
Looks good now. Great work! :rocket:

PR #3813, opened 2024-12-06
feat: support string comparison for jwt-role-claim-key

### LINKED ISSUE #1536: jwt: iterate over roles
### Environment

* PostgreSQL version: 12
* PostgREST version: 7.0.0
* Operating system: Linux

### Description of issue

Hello,
I configured postgrest authentication through jwt with Keycloak and here is what I get for my roles in the jwt token:

```json
"realm_access": {
    "roles": [
      "role1",
      "role2"
    ]
  }
```

In postgrest.conf, I have set `role-claim-key = ".realm_access.roles"` but postgrest takes it as a simple string and I get 
```
{"hint":null,"details":null,"code":"22023","message":"role \"[\"role1\",\"role2\"]\" does not exist"}
```

I tried to look at parsers.hs but I don't know haskell so I was unable to tell if it's feasible or not. 

As `role1` exists in my DB, I got it working with `role-claim-key = ".realm_access.roles[0]"` but this not good enough for production as the list ordering may depend on the user.

Do you have any clue how I can achieve this?

Thanks!

taimoorzaeem:
Fixes #1536.

@steve-chavez Please review.


test/io/fixtures.yaml
```diff
@@ -153,6 +153,41 @@ roleclaims:
       role: postgrest_test_author
       other: true
     expected_status: 401
+  - key: '.realm_access.roles[?(@ == "postgrest_test_author")]'
```
steve-chavez, line 156: Note: These values get used on this test https://github.com/PostgREST/postgrest/blob/60d92f64dcab681ffc1674cb87935cb0a638603b/test/io/test_io.py#L191-L205

steve-chavez:
@taimoorzaeem How about some docs for this feature?

I think we should:

- Have a section after https://docs.postgrest.org/en/v12/references/auth.html#jwt-claims-validation called something like "Role Claim Extraction"
- Move the details about JSPath on [jwt-role-claim-key](https://docs.postgrest.org/en/v12/references/configuration.html#jwt-role-claim-key) there.
-  Add the new comparison operators on the new section. 
- Finally, put a link on [jwt-role-claim-key](https://docs.postgrest.org/en/v12/references/configuration.html#jwt-role-claim-key) to this new section.

docs/references/configuration.rst
```diff
@@ -628,7 +628,12 @@ jwt-role-claim-key
     # the DSL accepts characters that are alphanumerical or one of "_$@" as keys
     jwt-role-claim-key = ".postgrest.roles[1]"
 
-    # {"https://www.example.com/role": { "key": "author }}
+    # {"postgrest":{"roles": ["other", "author"]}}
+    # accepts string comparison operators, supported operators are "==", "!=", "^==", "$==", "*=="
```
steve-chavez, line 632: These look sufficient to solve all the cases on https://github.com/PostgREST/postgrest/issues/1536.

test/io/fixtures.yaml
```diff
@@ -153,6 +153,41 @@ roleclaims:
...
+          - postgrest_test_author
+    expected_status: 200
+  - key: '.realm_access.roles[?(@ != "other")]'
+    data:
+      realm_access:
+        roles:
+          - other
+          - postgrest_test_author
+    expected_status: 200
+  - key: '.realm_access.roles[?(@ ^== "postgrest_te")]'
```
steve-chavez, line 170: Q: Where is documented that `@` means the left value? It does look good, but would like to know.

taimoorzaeem: It comes from the original source of JSONPath [https://goessner.net/articles/JsonPath/](https://goessner.net/articles/JsonPath/) and [https://github.com/JSONPath-Plus/JSONPath](https://github.com/JSONPath-Plus/JSONPath). We can also document this in our docs.

test/io/fixtures.yaml
taimoorzaeem: Yes, exactly! That's why I have added them here to test the feature.

src/PostgREST/Config/JSPath.hs
```diff
@@ -1,51 +1,98 @@
...
+  | NotEqualsCond Text
+  | StartsWithCond Text
+  | EndsWithCond Text
+  | ContainsCond Text
 
 dumpJSPath :: JSPathExp -> Text
 -- TODO: this needs to be quoted properly for special chars
 dumpJSPath (JSPKey k) = "." <> show k
 dumpJSPath (JSPIdx i) = "[" <> show i <> "]"
+dumpJSPath (JSPFilter cond) = "[?(@" <> expr <> "]"
```
steve-chavez, line 38: How about a test to pass codecov here?

steve-chavez: For this I think you can add some configs for the io tests:

- `test/io/configs/jwt-role-claim-key.config`
- `test/io/configs/expected/jwt-role-claim-key.config`

src/PostgREST/Config/JSPath.hs
```diff
@@ -1,51 +1,98 @@
...
+
+pJSPIdx :: P.Parser JSPathExp
+pJSPIdx = do
+  P.char '['
+  num <- read <$> P.many1 P.digit
+  P.char ']'
+  return $ JSPIdx num
 
-pJSPKey :: P.Parser Text
-pJSPKey = toS <$> P.many1 (P.alphaNum <|> P.oneOf "_$@") <|> pQuotedValue <?> "attribute name [a..z0..9_$@])"
```
steve-chavez, line 45: Is it possible to reincorporate this error message?

src/PostgREST/Config/JSPath.hs
```diff
@@ -1,51 +1,98 @@
...
+pJSPFilter = do
+  P.try $ P.string "[?("
+  condition <- pFilterConditionParser
+  P.char ')'
+  P.char ']'
+  P.eof -- this should be the last jspath expression
+  return $ JSPFilter condition
 
-pJSPIdx :: P.Parser Int
-pJSPIdx = P.char '[' *> (read <$> P.many1 P.digit) <* P.char ']' <?> "array index [0..n]"
```
steve-chavez, line 48: Ditto error message.

src/PostgREST/Config/JSPath.hs
```diff
@@ -1,51 +1,98 @@
...
 type JSPath = [JSPathExp]
 
--- | jspath expression, e.g. .property, .property[0] or ."property-dash"
+-- | jspath expression
 data JSPathExp
-  = JSPKey Text
-  | JSPIdx Int
+  = JSPKey Text         -- .property or ."property-dash"
+  | JSPIdx Int          -- [0]
+  | JSPFilter FilterExp -- [?(@ == "match")], [?(@ ^== "match-prefix")], etc
```
steve-chavez, line 25: How about a test that also has two conditions like the example here?

steve-chavez: Or if it doesn't make sense to have 2 conditions of this type.. should we restrict it to just 1?

taimoorzaeem: Hmm, I am not sure what you are asking here. We are already testing all possible conditions in `io-tests` added in `fixtures.yaml`.

src/PostgREST/Config/JSPath.hs
taimoorzaeem: Sure. Thanks, this actually helped me spot an issue with the `dumpJSPath` (fixed now). `code-coverage` is SO useful 🚀.

src/PostgREST/Config/JSPath.hs
steve-chavez: Yes, please add.

src/PostgREST/Config/JSPath.hs
steve-chavez: I mean, is it possible to have:

```
jwt-role-claim-key = '[?(@ == "match")], [?(@ ^== "match-prefix")]'
```

If so, does it make sense to support that?

steve-chavez:
Codecov still complains about some lines, can you try and clear those?

src/PostgREST/Config/JSPath.hs
```diff
@@ -1,51 +1,98 @@
...
+  condOp <- P.choice $ map P.string ["==", "!=", "^==", "$==", "*=="]
+  P.spaces
+  value <- pQuotedValue
+  return $ case condOp of
+     "=="  -> EqualsCond value
+     "!="  -> NotEqualsCond value
+     "^==" -> StartsWithCond value
+     "$==" -> EndsWithCond value
+     "*==" -> ContainsCond value
+     _     -> EqualsCond value -- Impossible case
```
steve-chavez, line 95: Seems it's possible to remove this impossible case if we use `try` like on:


https://github.com/PostgREST/postgrest/blob/60d92f64dcab681ffc1674cb87935cb0a638603b/src/PostgREST/ApiRequest/QueryParams.hs#L183-L194

src/PostgREST/Config/JSPath.hs
steve-chavez: Oh, ok. I misread the comment. I suggest just having one value to avoid confusion.
```suggestion
  | JSPFilter FilterExp -- [?(@ == "match")]
```

Additionally, from the `type JSPath = [JSPathExp]`, it looks like we can have multiple `JSPFilter`? Could we make it so it only accepts one? 

Or how do we ensure that currently? At the parser level? If so, and if the above is too complicated, maybe add a comment here mentioning it.

taimoorzaeem: Yes, this is ensured at the parser level. It only accepts `JSPFilter` at the end of input. I understand it is not the best design but it would have gotten too complicated otherwise. Sure, I'll add a comment mentioning this.
```haskell
pJSPFilter :: P.Parser JSPathExp
pJSPFilter = do
  P.try $ P.string "[?("
  condition <- pFilterConditionParser
  P.char ')'
  P.char ']'
  P.eof -- this should be the last jspath expression
  (return $ JSPFilter condition) <?> "pJSPFilter: JSPath filter exp"

```

taimoorzaeem:
@steve-chavez I have changed the `$==` operator with `==^`. Turns out that the `$` character does not work well inside environment variable values which was causing some issues during testing. I could have overlooked this if it wasn't for `code-cov` not passing enough. Huh, another win for `code-cov` testing  🚀.

steve-chavez:
> I have changed the $== operator with ==^

@taimoorzaeem Are those functionally equivalent?

docs/references/auth.rst
```diff
@@ -161,6 +161,43 @@ JWT Claims Validation
 
 PostgREST honors the :code:`exp` claim for token expiration, rejecting expired tokens.
 
+.. _jwt_role_claim_key_extract:
+
+JWT Role Claim Key Extraction
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A JSPath DSL that specifies the location of the :code:`role` key in the JWT claims. This can be used to consume a JWT provided by a third party service like Auth0, Okta or Keycloak.
```
steve-chavez, line 169: Let's add Microsoft Entra (mentioned on https://github.com/PostgREST/postgrest/issues/1536#issuecomment-2482169503) for SEO purposes.

```suggestion
A JSPath DSL that specifies the location of the :code:`role` key in the JWT claims. This can be used to consume a JWT provided by a third party service like Auth0, Okta, Microsoft Entra or Keycloak.
```

taimoorzaeem:
> @taimoorzaeem Are those functionally equivalent?

Yes, absolutely.

steve-chavez (approved):
Great work! :rocket:

PR #3818, opened 2024-12-16
feat: allow not_null value for the is operator

### LINKED ISSUE #3747: Does postgrest not support "not null" as valud `is` value?
```
failed to parse filter (is.not null)
```

Forcing negate on quering `not null` value seems like a mistake, `not is null` and `is not null` are identical, but as I am building a dynamic query builder, having to transform (negating) the operator by its value is quite not straight forward.

This might seem like a nich-cased suggestion, but
supporting value `not+null` or `not_null` seems like a generally good idea (also for others)

: )

E.g.

image

taimoorzaeem:
Fixes #3747.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3560, Log resolved host in "Listening on ..." messages - @develop7
  - #3727, Log maximum pool size - @steve-chavez
  - #1536, Add string comparison feature for jwt-role-claim-key - @taimoorzaeem
+ - #3747, Allow `not_null` value for the `is` operator - @taimoorzaeem
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3560, Log resolved host in "Listening on ..." messages - @develop7
  - #3727, Log maximum pool size - @steve-chavez
  - #1536, Add string comparison feature for jwt-role-claim-key - @taimoorzaeem
+ - #3747, Add `not_null` filter value to pgrst query - @taimoorzaeem
```
steve-chavez, line 16:
```suggestion
 - #3747, Allow `not_null` value for the `is` operator - @taimoorzaeem
```

docs/references/api/resource_embedding.rst
```diff
@@ -1042,6 +1042,10 @@ Both ``is.null`` and ``not.is.null`` can be included inside the `or` operator. F
     -d select=title,actors(*),directors(*)" \
     -d "or=(actors.is.null,directors.is.null)"
 
+.. note::
+  
+  PostgREST also provides ``is.not_null`` filter which works the same as ``not.is.null``. See `#3747 <https://github.com/PostgREST/postgrest/issues/3747>`_.
+
```
steve-chavez, line 1048:
```suggestion
```

I think we should only document this on https://github.com/PostgREST/postgrest/blob/60d92f64dcab681ffc1674cb87935cb0a638603b/docs/references/api/tables_views.rst#L75

PR #3829, opened 2024-12-23
fix: insert with missing=default uses column default before using domain default

### LINKED ISSUE #3706: Doing a bulk insert with `missing=default` uses the default value of the domain instead of the column

taimoorzaeem:
Fixes #3706.

### DIFF
--- a/test/spec/Feature/Query/InsertSpec.hs
+++ b/test/spec/Feature/Query/InsertSpec.hs
@@ -572,6 +572,15 @@ spec actualPgVersion = do
               , matchHeaders = ["Preference-Applied" <:> "missing=default, return=representation"]
               }
 
+        it "inserts a COLUMN default before a DOMAIN default with missing=default" $
+          request methodPost "/evil_friends_with_column_default?columns=id,name" [("Prefer", "return=representation"), ("Prefer", "missing=default")]
+              [json| { "name": "Demon" } |]
+            `shouldRespondWith`
+              [json| [{"id": 420, "name": "Demon"}] |]
+              { matchStatus  = 201
+              , matchHeaders = ["Preference-Applied" <:> "missing=default, return=representation"]
+              }
+
     it "inserts json that has duplicate keys" $ do
       request methodPost "/tbl_w_json" [("Prefer", "return=representation")]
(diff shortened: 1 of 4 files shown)


test/spec/Feature/Query/InsertSpec.hs
```diff
@@ -572,6 +572,15 @@ spec actualPgVersion = do
               , matchHeaders = ["Preference-Applied" <:> "missing=default, return=representation"]
               }
 
+        it "inserts a COLUMN default before a DOMAIN default with missing=default" $
+          request methodPost "/evil_friends_with_column_default?columns=id,name" [("Prefer", "return=representation"), ("Prefer", "missing=default")]
```
steve-chavez, line 576: Are there tests that ensure a DOMAIN default will be applied in other cases?

taimoorzaeem: Yes, it is present just above this new test.

steve-chavez (approved):

PR #3845, opened 2025-01-14
feat: apply `to_tsvector` to the filtered columns when a `fts` operator is used

### LINKED ISSUE #2255: Change FTS operators to explicitly call to_tsvector()
### Environment

* PostgreSQL version: PostgreSQL 13.6 (Ubuntu 13.6-0ubuntu0.21.10.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.2.0-7ubuntu2) 11.2.0, 64-bit
* PostgREST version: postgrest/postgrest:v9.0.0 
* Operating system: Ubuntu 21.10 

### Description of issue

I cannot perform a full-text search on json or jsonb column types.

#### Steps to reproduce
1. Perform a query such as `GET /contacts?select=jsonb_field&jsonb_field_1=fts.jsonvalue`

#### Actual result
```json
{
    "hint": "No operator matches the given name and argument types. You might need to add explicit type casts.",
    "details": null,
    "code": "42883",
    "message": "operator does not exist: jsonb @@ tsquery"
}
```
The query produced contains the SQL:

```sql
WHERE  "public"."contacts"."jsonb_field" @@ to_tsquery($1) 
```

#### Expected Result

I'm able to perform a full-text search on a json/jsonb column. There is no workaround with PostgREST at present.

### Possible solution

Instead of generating:
```sql
WHERE  "public"."contacts"."jsonb_field" @@ to_tsquery($1) 
```
...generate:
```sql
WHERE  to_tsvector("public"."contacts"."jsonb_field") @@ to_tsquery($1) 
```
I believe this is to be exactly equivalent. According to the doc, `to_tsvector()` is implicitly called with text:

> text @@ tsquery → boolean
>
>    Does text string, after implicit invocation of to_tsvector(), match tsquery?

By explicitly calling `to_tsvector()`, json/jsonb columns would also be handled.

laurenceisla:
Closes #2255 

When an `fts` operator is used, then it applies `to_tsvector` to the filtered column. For example:

```
/table?column=fts(simple).test
```

is equivalent to this in SQL:

```sql
SELECT *
FROM table
WHERE to_tsvector('simple', column) = to_tsquery('simple', 'test')
```

<s>
If the `totsv` modifier (short for `to_tsvector`) is used for `fts` operators then it applies `to_tsvector` to the filtered column. For example:

```
/table?column=totsv.fts(simple).test
```

Is equivalent to this in SQL:

```sql
SELECT *
FROM table
WHERE to_tsvector('simple', column) = to_tsquery('simple', 'test')
```
</s>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3727, Log maximum pool size - @steve-chavez
  - #1536, Add string comparison feature for jwt-role-claim-key - @taimoorzaeem
  - #3747, Allow `not_null` value for the `is` operator - @taimoorzaeem
+ - #2255, Apply `to_tsvector()` explicitly to the full-text search filtered column (excluding `tsvector` types) - @laurenceisla
 
 ### Fixed
 
--- a/docs/references/api/tables_views.rst
+++ b/docs/references/api/tables_views.rst
@@ -193,7 +193,21 @@ The :code:`fts` filter mentioned above has a number of options to support flexib
 
   curl "http://localhost:3000/tsearch?my_tsv=not.wfts(french).amusant"
 
-Using `websearch_to_tsquery` requires PostgreSQL of version at least 11.0 and will raise an error in earlier versions of the database.
+.. _fts_to_tsvector:
+
+Automatic ``tsvector`` conversion
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the filtered column is not of type ``tsvector``, then it will be automatically converted using `to_tsvector() <https://www.postgresql.org/docs/current/functions-textsearch.html#TEXTSEARCH-FUNCTIONS-TABLE>`_.
+This allows using ``fts`` on ``text`` and ``json`` types out of the box, for example.
+
+.. code-block:: bash
+
+  curl "http://localhost:3000/tsearch?my_text_column=fts(french).amusant"
+
+.. code-block:: bash
+
+  curl "http://localhost:3000/tsearch?my_json_column=not.phfts(english).The%20Fat%20Cats"
 
 .. _v_filter:
 
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -265,27 +265,29 @@ data ResolverContext = ResolverContext
   , outputType      :: Text                 -- ^ The output type for the response payload; e.g. "csv", "json", "binary".
   }
 
-resolveColumnField :: Column -> CoercibleField
-resolveColumnField col = CoercibleField (colName col) mempty False (colNominalType col) Nothing (colDefault col) False
+resolveColumnField :: Column -> Maybe ToTsVector -> CoercibleField
+resolveColumnField col toTsV = CoercibleField (colName col) mempty False toTsV (colNominalType col) Nothing (colDefault col) False
 
-resolveTableFieldName :: Table -> FieldName -> CoercibleField
-resolveTableFieldName table fieldName =
+resolveTableFieldName :: Table -> FieldName -> Maybe ToTsVector -> CoercibleField
+resolveTableFieldName table fieldName toTsV=
   fromMaybe (unknownField fieldName []) $ HMI.lookup fieldName (tableColumns table) >>=
-    Just . resolveColumnField
+    Just . flip resolveColumnField toTsV
 
 -- | Resolve a type within the context based on the given field name and JSON path. Although there are situations where failure to resolve a field is considered an error (see `resolveOrError`), there are also situations where we allow it (RPC calls). If it should be an error and `resolveOrError` doesn't fit, ensure to check the `cfIRType` isn't empty.
-resolveTypeOrUnknown :: ResolverContext -> Field -> CoercibleField
-resolveTypeOrUnknown ResolverContext{..} (fn, jp) =
+resolveTypeOrUnknown :: ResolverContext -> Field -> Maybe ToTsVector -> CoercibleField
+resolveTypeOrUnknown ResolverContext{..} (fn, jp) toTsV =
   case res of
     -- types that are already json/jsonb don't need to be converted with `to_jsonb` for using arrow operators `data->attr`
     -- this prevents indexes not applying https://github.com/PostgREST/postgrest/issues/2594
-    cf@CoercibleField{cfIRType="json"}  -> cf{cfJsonPath=jp, cfToJson=False}
-    cf@CoercibleField{cfIRType="jsonb"} -> cf{cfJsonPath=jp, cfToJson=False}
+    cf@CoercibleField{cfIRType="json"}     -> cf{cfJsonPath=jp, cfToJson=False}
+    cf@CoercibleField{cfIRType="jsonb"}    -> cf{cfJsonPath=jp, cfToJson=False}
+    -- Do not apply to_tsvector to tsvector types
+    cf@CoercibleField{cfIRType="tsvector"} -> cf{cfJsonPath=jp, cfToJson=True, cfToTsVector=Nothing}
     -- other types will get converted `to_jsonb(col)->attr`, even unknown types
-    cf                                  -> cf{cfJsonPath=jp, cfToJson=True}
+    cf                                     -> cf{cfJsonPath=jp, cfToJson=True}
   where
     res = fromMaybe (unknownField fn jp) $ HM.lookup qi tables >>=
-          Just . flip resolveTableFieldName fn
+          Just . (\t -> resolveTableFieldName t fn toTsV)
 
 -- | Install any pre-defined data representation from source to target to coerce this reference.
 --
@@ -311,11 +313,15 @@ withJsonParse ctx field@CoercibleField{cfIRType} = withTransformer ctx "json" cf
 
 -- | Map the intermediate representation type to the output type defined by the resolver context (normally json), if available.
 resolveOutputField :: ResolverContext -> Field -> CoercibleField
-resolveOutputField ctx field = withOutputFormat ctx $ resolveTypeOrUnknown ctx field
+resolveOutputField ctx field = withOutputFormat ctx $ resolveTypeOrUnknown ctx field Nothing
 
 -- | Map the query string format of a value (text) into the intermediate representation type, if available.
-resolveQueryInputField :: ResolverContext -> Field -> CoercibleField
-resolveQueryInputField ctx field = withTextParse ctx $ resolveTypeOrUnknown ctx field
+resolveQueryInputField :: ResolverContext -> Field -> OpExpr -> CoercibleField
+resolveQueryInputField ctx field opExpr = withTextParse ctx $ resolveTypeOrUnknown ctx field toTsVector
+  where
+    toTsVector = case opExpr of
+      OpExpr _ (Fts _ lang _) -> Just $ ToTsVector lang
+      _                       -> Nothing
 
 -- | Builds the ReadPlan tree on a number of stages.
 -- | Adds filters, order, limits on its respective nodes.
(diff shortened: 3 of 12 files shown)


docs/references/api/tables_views.rst
```diff
@@ -193,7 +194,16 @@ The :code:`fts` filter mentioned above has a number of options to support flexib
 
   curl "http://localhost:3000/tsearch?my_tsv=not.wfts(french).amusant"
 
-Using `websearch_to_tsquery` requires PostgreSQL of version at least 11.0 and will raise an error in earlier versions of the database.
+For other column types like ``text`` or ``json``, you may want to explicitly convert these columns using ``to_tsvector()``, specially when a language is specified.
+To do so, use the ``totsv`` modifier:
+
+.. code-block:: bash
+
+  curl "http://localhost:3000/tsearch?text_column=totsv.fts(french).amusant"
```
steve-chavez, line 202: Hm, at a glance this looks too ad hoc and leaking internal details through the API.

Could we do the casting automatically with the schema cache approach mentioned on https://github.com/PostgREST/postgrest/issues/2255#issuecomment-1111220493?

laurenceisla: We could. But we wouldn't know if a computed column ([created in the way we mention in the docs](https://docs.postgrest.org/en/v12/references/api/computed_fields.html#computed-fields)) is of `tsvector` type or not.

Hmm... to solve this we could apply `to_tsvector('language',column)` only to those we know are `text` or json types and let PostgREST do the implicit casting to the rest. Still if other computed column with text or json types are present they wouldn't be using the `to_tsvector` and the problem would persist for those.

steve-chavez: > But we wouldn't know if a computed column ([created in the way we mention in the docs](https://docs.postgrest.org/en/v12/references/api/computed_fields.html#computed-fields)) is of tsvector type or not.

Why? Care to elaborate?

I haven't seen computed fields discussed on the issue, are they really relevant for this use case?

laurenceisla: > Why? Care to elaborate?

We cannot detect the computed field because it's not a column of the table, so it won't be present in the schema cache. For example, what would you suggest in those cases? 
  * Use `to_tsvector()` in the computed field anyways, but if it returns a `tsvector` it would err, or
  * Don't add anything and let the issue happen for computed fields only.

> I haven't seen computed fields discussed on the issue, are they really relevant for this use case?

Ah no, they weren't mentioned, just trying to cover all possibilities here. e. g. From the example in the docs, using `?full_name=totsv.fts(simple).and` would work, but using the schema cache (if we don't do anything for computed columns), it wouldn't.

laurenceisla: TBH I also like the approach of auto-detecting the type, but since it didn't cover everything I didn't go for it. We could implement that and wait to see if there are actually problems with users and computed fields or, better yet, document it in some way (since they're already using computed fields, creating a `to_tsvector` one should be the solution). WDYT?

steve-chavez: > Use to_tsvector() in the computed field anyways, but if it returns a tsvector it would err, or

That sounds fine to me for now. Later on we can make it a more clear error if we add computed fields to the schema cache.

laurenceisla: Hmmm... this would break responses for those who are already using `tsvector` computed cols. I'd prefer to not do anything for those cases. Maybe we need @wolfgangwalther to break our stalemate here?

laurenceisla: Cool, I'll keep the computed columns for later then. Tracked here #956

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3795, Clarify `Accept: vnd.pgrst.object` error message - @steve-chavez
  - #3779, Always log the schema cache load time - @steve-chavez
  - #3706, Fix insert with `missing=default` uses default value of domain instead of column - @taimoorzaeem
+ - #2255, Fix `to_tsvector()` not applying explicitly to the full-text search filtered column (on non `tsvector` types) - @laurenceisla
```
laurenceisla, line 29: I'm not entirely sure if this should be considered a new feature or a fix. I think the user expects the `to_tsvector` to be applied under the hood, so I'm adding it here.

src/PostgREST/Plan.hs
```diff
@@ -265,27 +265,29 @@ data ResolverContext = ResolverContext
...
+    cf@CoercibleField{cfIRType="jsonb"}    -> cf{cfJsonPath=jp, cfToJson=False}
+    -- Do not apply to_tsvector to tsvector types
+    cf@CoercibleField{cfIRType="tsvector"} -> cf{cfJsonPath=jp, cfToJson=True, cfToTsVector=Nothing}
     -- other types will get converted `to_jsonb(col)->attr`, even unknown types
-    cf                                  -> cf{cfJsonPath=jp, cfToJson=True}
+    cf                                     -> cf{cfJsonPath=jp, cfToJson=True}
   where
     res = fromMaybe (unknownField fn jp) $ HM.lookup qi tables >>=
-          Just . flip resolveTableFieldName fn
+          Just . (\t -> resolveTableFieldName t fn toTsV)
```
laurenceisla, line 290: There was a performance concern here: https://github.com/PostgREST/postgrest/issues/2255#issuecomment-1111220493

I don't think there will be a difference in performance since it's using this lookup that is already applied for all filters before this PR.

CHANGELOG.md
laurenceisla: > the PR title says "feat:"

Oh woops, I left that from the previous iteration (adding the `totsv` modifier looked more like a feat in that case for me). Yeah, I get it now.

>   I understand you're applying this transformation to every column except those with type tsvector right now. This does have a considerably risk of breaking indexes, we only need to go back as far as when we did something similar with `to_jsonb(...)` around everything to allow accessing fields or composite types or so

Yes, although there's a difference in that the `to_tsvector()`[ is already applied implicitly to the `text` types with the `@@` operator](https://www.postgresql.org/docs/current/functions-textsearch.html#TEXTSEARCH-OPERATORS-TABLE).

```
text @@ tsquery → boolean

Does text string, after implicit invocation of to_tsvector(), match tsquery?

'fat cats ate rats' @@ to_tsquery('cat & rat') → t
```

The other types will still fail but with a different error message, e.g. `function to_tsquery(integer) does not exist` instead of  `operator does not exist: integer @@ tsquery`.

laurenceisla: > This does have a considerably risk of breaking indexes,

In fact, I think it's the contrary, actually. I've been checking more about indexes [and the docs say](https://www.postgresql.org/docs/current/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX):

> Because the two-argument version of to_tsvector was used in the index above, only a query reference that uses the 2-argument version of to_tsvector with the same configuration name will use that index. That is, `WHERE to_tsvector('english', body) @@ 'a & b'` can use the index, but `WHERE to_tsvector(body) @@ 'a & b` cannot. This ensures that an index will be used only with the same configuration used to create the index entries.

As it is right now if we a GIN index to a column of `text` type, PostgREST won't use it. This change (appying to all types, not only JSON) would fix that.

laurenceisla: > Like.. do we actually need this ~fix~ feature, or could we just advise people to create the respective ~cast~ operator?

> It seems to me that suggesting to our users to create generated columns with the tsvector output of the json column would solve the problem nicely.

Yes, that'd be another solution. We'd need to mention that it won't work as expected for `text` or `json` columns and to use `tsvector` computed columns and give the suggestions you list. Nonetheless, I think having an out of the box option to use would be a good experience.

> If instead, we implement as proposed here, I see the next feature request coming very soon: "I need to pass 'english' to the two-argument-version of tsvector, please let PostgREST do that for me."

Do you mean `to_tsvector(config_language, text)`? Cause... that's what it's implemented in this PR: `col=fts(english).val` in SQL would be `to_tsvector('english', col) @@ to_tsquery('english', 'val')` (I updated the original post of the PR to reflect this). I figured it made sense having the same config languages in the tsquery and the tsvector, so using the one in `fts(language)` for both shouldn't be a probem.

---

All in all, I'm on the fence here. I see the value in both having the implementation out of the box and recommending `generated columns` as an alternative and also for complex cases (computed fields are not included in the PR, for instance).

laurenceisla: > Seems like we should maybe do both. This PR and a better explanation of the generated column approach for special cases?

Agree, opened #3870 to track this. Meanwhile this PR should be ready for review/merging.

src/PostgREST/Plan/Types.hs
```diff
@@ -20,6 +21,7 @@ import PostgREST.SchemaCache.Identifiers (FieldName)
 import Protolude
 
 type TransformerProc = Text
+type ToTsVector = Maybe (Maybe Text)
```
steve-chavez, line 24: This is confusing at first glance, why a double Maybe?

docs/references/api/tables_views.rst
```diff
@@ -193,7 +193,7 @@ The :code:`fts` filter mentioned above has a number of options to support flexib
 
   curl "http://localhost:3000/tsearch?my_tsv=not.wfts(french).amusant"
 
-Using `websearch_to_tsquery` requires PostgreSQL of version at least 11.0 and will raise an error in earlier versions of the database.
+This also works with columns of text and JSON types, converting them to ``tsvector`` accordingly.
```
steve-chavez, line 196: A subsection would be better for this. This way we can also highlight the feature on the release notes.

Maybe "automatic tsvector casting"`?

src/PostgREST/Plan/Types.hs
laurenceisla: True, I created a new type to make it make sense.

docs/references/api/tables_views.rst
```diff
@@ -193,7 +193,21 @@ The :code:`fts` filter mentioned above has a number of options to support flexib
...
+If the filtered column is not of type ``tsvector``, then it will be automatically converted using `to_tsvector() <https://www.postgresql.org/docs/current/functions-textsearch.html#TEXTSEARCH-FUNCTIONS-TABLE>`_.
+This allows using ``fts`` on text and JSON types out of the box.
+
+.. code-block:: bash
+
+  curl "http://localhost:3000/tsearch?my_text_column=fts(french).amusant"
+
+.. code-block:: bash
+
+  curl "http://localhost:3000/tsearch?my_json_column=not.phfts(english).The%20Fat%20Cats"
```
steve-chavez, line 210: Just the json example is good, when I see `my_text_column` it makes me scratch my head.

```suggestion
Automatic ``tsvector`` conversion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

If the filtered column is not of type ``tsvector``, then it will be automatically converted using `to_tsvector() <https://www.postgresql.org/docs/current/functions-textsearch.html#TEXTSEARCH-FUNCTIONS-TABLE>`_.
This allows using ``fts`` on JSON and ``text`` types out of the box.

.. code-block:: bash

  curl "http://localhost:3000/tsearch?my_json_column=fts(french).amusant"
```

steve-chavez: Wait, maybe it's enough to highlight the types..

steve-chavez:
```suggestion
Automatic ``tsvector`` conversion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

If the filtered column is not of type ``tsvector``, then it will be automatically converted using `to_tsvector() <https://www.postgresql.org/docs/current/functions-textsearch.html#TEXTSEARCH-FUNCTIONS-TABLE>`_.
This allows using ``fts`` on ``text`` and ``JSON`` types out of the box.

.. code-block:: bash

  curl "http://localhost:3000/tsearch?my_text_column=fts(french).amusant"

.. code-block:: bash

  curl "http://localhost:3000/tsearch?my_json_column=not.phfts(english).The%20Fat%20Cats"
```

steve-chavez (approved):
Great work :100:

PR #3855, opened 2025-01-17
ci: test workflows on new arm runners

wolfgangwalther:
GitHub has now made the ARM runners available for public repositories:

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

As a quick-and-dirty test, I just replaced all the ubuntu runners with their -arm counterparts, let's see which of the jobs build successfully and which don't.

We should be able to remove our current remote arm builder soon. Hopefully with a static binary, but if not - the regular build should still be much easier to do than those via-SSH-builds.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -107,12 +107,9 @@ jobs:
           - name: Windows
             runs-on: windows-2022
             cache: |
-              ~\AppData\Roaming\stack\pantry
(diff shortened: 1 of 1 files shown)


steve-chavez:
> Thus, we don't use the old ARM runner anywhere anymore - you can shut that thing down, @steve-chavez!

Done!

PR #3866, opened 2025-01-19
Build static executable for linux aarch64 (w/o TH)

wolfgangwalther:
This is based on #3865, so the first few commits are from there.

Right now, this is a POC: What happens when we remove all the Template Haskell, update to GHC 9.6/9.8, and then try building the static aarch64 executable again?

This happens:

```
% file /nix/store/[...]-postgrest-static-aarch64-unknown-linux-musl-12.3/bin/postgrest
/nix/store/[...]-postgrest-static-aarch64-unknown-linux-musl-12.3/bin/postgrest:
  ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, stripped
```

Once we have that, we could once more simplify our docker image and ship the minimal, single executable for both platforms. Drastically reduced size.

I also tried to build the static executable on aarch64-darwin and x86_64-freebsd. There's still a bit more work to do for those, but it looks much better than... a year ago. In any case, we need to do our part first.

In #3597 I introduced neat-interpolation as the quasi quoter, because it cross-compiles better. It does - but any project actually *using* it, will still have the same "external interpreter" problem that any Template Haskell causes. So the only way forward is to remove **all** use of Quasi Quoters. Yes - the change look a bit ugly, with all the quotes etc. - but that's the best we can do right now.

GHC 9.12 has [native multiline strings](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/multiline_strings.html), which could go a long way already. We're still a bit away from making 9.12 our minimum version, though. There's also a [GHC proposal for native string interpolation](https://github.com/ghc-proposals/ghc-proposals/pull/570), which would give us everything we need, once agreed on and implemented. But yeah - even further out in the future.

Let's see how long we need to replace OpenAPI - maybe native string interpolation is already widely supported by then ;).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,6 @@
 { system ? builtins.currentSystem
 
-, compiler ? "ghc966"
+, compiler ? "ghc984"
 
 , # Commit of the Nixpkgs repository that we want to use.
   nixpkgsVersion ? import nix/nixpkgs-version.nix


steve-chavez:
> So the only way forward is to remove all use of Quasi Quoters. Yes - the change look a bit ugly, with all the quotes etc. - but that's the best we can do right now.

Completely fine with that too.

> Let's see how long we need to replace OpenAPI - maybe native string interpolation is already widely supported by then ;)

Could we fork (or vendor) [swagger2](https://hackage.haskell.org/package/swagger2) and do the same process to get rid of the template haskell dependency? Looks like that would be the quickest way to unblock this.

taimoorzaeem:
> same "external interpreter" problem that any Template Haskell causes

@wolfgangwalther I would like to know more about the external interpreter problem. I am looking for external links or issue threads related to this.

taimoorzaeem:
@wolfgangwalther Some questions:

> Could we fork (or vendor) [swagger2](https://hackage.haskell.org/package/swagger2) and do the same process to get rid of the template haskell dependency? Looks like that would be the quickest way to unblock this.

Does getting rid of the template haskell dependency mean that we only need to remove it as a **direct** dependency and not if the package is **indirectly** dependent on it?

Does our goal of cross-compilation mean that we eventually need to get rid of template haskell from all the packages that we depend on?

> Plus, we'd end up with some dependencies which are not on hackage, again, which works against our goal of eventually having an up2date PostgREST available in nixpkgs itself as well.

How sustainable or worth it does this sound if I say that I can write a `swagger2` like library (maybe patch it) which is not dependent on template haskell and I can also upload this on hackage and maintain it?

taimoorzaeem:
> > Does our goal of cross-compilation mean that we eventually need to get rid of template haskell from all the packages that we depend on?
> 
> Yes, correct.

Hm, this does not seem very possible to me. Even the very basic packages like `aeson` and `lens` depend on TH. Infact, most workable packages on hackage directly or indirectly depend on TH.

taimoorzaeem:
> IIRC, aeson only uses TH for its test-suite. That can just be disabled for cross-compiling - done

`aeson` uses TH for a [QuasiQuoter](https://hackage.haskell.org/package/aeson-2.2.3.0/docs/Data-Aeson-QQ-Simple.html) for JSON in their library. Can we disable this too?

Issue #3046, opened 2023-11-10
Large schema cache reloads cause 100% CPU spike when schema cache metadata is accessed.


colophonemes:
### Environment

* PostgreSQL version: 13.10 (also tested 13.11)
* PostgREST version: Known issue since v9.0, recently tested in 11.1.0 and 11.2.1
* Operating system: MacOS, Docker (ECS)

### Description of issue

Our **production** instance of PostgREST exhibits the following behaviour:
- Schema cache is reloaded (either via app restart or `NOTIFY pgrst reload`)
- Things work as normal for a few minutes
- Something actually needs to look at the updated schema cache (the most reliable way to reproduce is to load the OpenAPI spec from the `/` endpoint, but the behaviour seems to eventually happen regardless)
- PostgREST starts spinning CPU, pegging at 100% for about five mins
	- If you triggered the spike by attempting to load the OpenAPI schema endpoint, we at first get a partial response (always 135,063 bytes in my testing) that loads within a few seconds, with a truncated form of the JSON (it's not a valid JSON object that contains fewer entries, but a longer response that has been truncated). Eventually this response will time out
	- Subsequent reloads of the OpenAPI schema endpoint trigger a CPU spike of similar duration
	- There's usually two spikes (one per OpenAPI refresh), and then by the end of the second one, the OpenAPI schema request finishes buffering and the full JSON loads; occasionally there'll be a third spike that's a bit shorter than the previous two, and the OpenAPI schema resolves after that one.
	- Once the OpenAPI schema request has loaded fully, subsequent requests return ~immediately, until such time as the schema cache reloads again.

Here's me reproducing the issue in a local Docker container, connected to our production DB. The time it takes for the CPU spike to calm down is shorter, presumably because my laptop is beefier than the ECS instances we run our production PostgREST on.

![image](https://github.com/PostgREST/postgrest/assets/6322017/c82b7fbd-f0cb-45c1-aafe-fc8d9b481e22)

We've observed this issue since at least v9.0.0 (though possibly it was present earlier), through to the present day. I've just run through the testing steps above on both v11.1.0 and v11.2.1, and the behaviour is there in both.

#### Possibly an interaction with TimescaleDB?

[...]

PR #3869, opened 2025-01-21
fix: handle queries on non-existing table gracefully

### LINKED ISSUE #3602: Return 404 on none existing endpoint with resource embedding
### Environment

* PostgreSQL version: 15.7
* PostgREST version: 12.0.3
* Operating system: Ubuntu 24.04

### Description of issue

When making a request to a none existing endpoint that includes resource embedding, it returns an error 400.

```HTTP
GET /none_existing_endpoint?select=title,directors(id,last_name) HTTP/1.1

Response: 400 Bad Request
{
    "code": "PGRST200",
    "details": "Searched for a foreign key relationship between 'none_existing_endpoint' and 'directors' in the schema 'exposed', but no matches were found.",
    "hint": null,
    "message": "Could not find a relationship between 'none_existing_endpoint' and 'directors' in the schema cache"
}
```

I would expect an 404 that tells me that the endpoint itself does not exist like so:

```HTTP
GET /none_existing_endpoint?select=title,directors(id,last_name) HTTP/1.1

Response: 404 Bad Request
{
    "code": "42P01",
    "details": null,
    "hint": null,
    "message": "relation \"exposed.none_existing_endpoint\" does not exist"
}

### LINKED ISSUE #3697: Insert into non-existing table raises empty APIErrror
# Insert into non-existing table raises empty APIErrror

## Description

Found this bug while using Supabase client, where I originally opened this [issue](https://github.com/supabase/supabase-py/issues/901#issuecomment-2296659874).

If `client.table("table_name")` does not exist (e.g. typo), `insert` and `upsert` operations raise empty APIError. 

## To Reproduce

1. Create reference to non-existing table
2. Attempt to insert data
3. Raises empty error with no information provided.

```
 def __init__(self, client: AsyncClient, user: AuthUser):
        self.client = client
        self.repository = self.client.table("not_existing_table")
        self.user = user

        try:
            insert_response = await self.repository.upsert(
                [mod.dict() for mod in db_models],
                count=CountMethod.exact,
                on_conflict="id",
                default_to_null=False,
            ).execute()
        except APIError as e:
            logger.error(f"Failed to bulk insert", user=self.user.id, **e.__dict__)
            # HERE the error is empty
```

## Expected behavior

The error should contain information about the table not present in the scheme.

## Screenshots
Screenshot 2024-08-18 at 23 36 41

## System information

- OS: macOS
- Version of supabase-py: ^2.6.0
- Version of postgrest: 0.16.9
- Version of PostgreSQL: 15.6

### PR DESCRIPTION
**Pull request by taimoorzaeem**
Add new `TableNotFound` error and add json error
message for this error. Closes #3697, closes #3602.

(Couldn't think of a better changelog/commit message. Feel free to suggest a new/better message).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Fixed `"column <json_aggregate>.<alias> does not exist"` error when selecting `?select=...table(aias:count())`
  - #3727, Clarify "listening" logs - @steve-chavez
  - #3795, Clarify `Accept: vnd.pgrst.object` error message - @steve-chavez
+ - #3697, #3602, Handle queries on non-existing table gracefully - @taimoorzaeem
 
 ### Changed
 
@@ -40,6 +41,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + This preference was deprecated in favor of Functions with an array of JSON objects
  - #3013, Drop support for Limited updates/deletes
    + The feature was complicated and largely unused.
+ - #3697, #3602, Querying non-existent table now returns `PGRST205` error instead of empty json - @taimoorzaeem
 
 ## [12.2.8] - 2025-02-10
 
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -45,6 +45,7 @@ import PostgREST.SchemaCache.Relationship (Cardinality (..),
                                            RelationshipsMap)
 import PostgREST.SchemaCache.Routine      (Routine (..),
                                            RoutineParam (..))
+import PostgREST.SchemaCache.Table        (Table (..))
 import Protolude
 
 
@@ -82,6 +83,7 @@ data ApiRequestError
   | UnacceptableSchema [Text]
   | UnsupportedMethod ByteString
   | ColumnNotFound Text Text
+  | TableNotFound Text Text [Table]
   | GucHeadersError
   | GucStatusError
   | PutMatchingPkError
@@ -128,6 +130,7 @@ instance PgrstError ApiRequestError where
   status UnacceptableSchema{}        = HTTP.status406
   status UnsupportedMethod{}         = HTTP.status405
   status ColumnNotFound{}            = HTTP.status400
+  status TableNotFound{}             = HTTP.status404
   status GucHeadersError             = HTTP.status500
   status GucStatusError              = HTTP.status500
   status PutMatchingPkError          = HTTP.status400
@@ -285,6 +288,12 @@ instance JSON.ToJSON ApiRequestError where
   toJSON (ColumnNotFound relName colName) = toJsonPgrstError
     SchemaCacheErrorCode04 ("Could not find the '" <> colName <> "' column of '" <> relName <> "' in the schema cache") Nothing Nothing
 
+  toJSON (TableNotFound schemaName relName tbls) = toJsonPgrstError
+    SchemaCacheErrorCode05
+    ("Could not find the table '" <> schemaName <> "." <> relName <> "' in the schema cache")
+    Nothing
+    (JSON.String <$> tableNotFoundHint schemaName relName tbls)
+
 -- |
 -- If no relationship is found then:
 --
@@ -381,6 +390,16 @@ noRpcHint schema procName params allProcs overloadedProcs =
     possibleProcs
       | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
       | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params)
+
+-- |
+-- Do a fuzzy search in all tables in the same schema and return closest result
+tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
+tableNotFoundHint schema tblName tblList
+  = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
+    where
+      perhapsTable = Fuzzy.getOne fuzzyTableSet tblName
+      fuzzyTableSet = Fuzzy.fromList [ tableName tbl | tbl <- tblList, tableSchema tbl == schema]
+
 
 compressedRel :: Relationship -> JSON.Value
 -- An ambiguousness error cannot happen for computed relationships TODO refactor so this mempty is not needed
@@ -672,6 +691,7 @@ data ErrorCode
   | SchemaCacheErrorCode02
   | SchemaCacheErrorCode03
   | SchemaCacheErrorCode04
+  | SchemaCacheErrorCode05
   -- JWT authentication errors
   | JWTErrorCode00
   | JWTErrorCode01
@@ -719,6 +739,7 @@ buildErrorCode code = case code of
   SchemaCacheErrorCode02 -> "PGRST202"
   SchemaCacheErrorCode03 -> "PGRST203"
   SchemaCacheErrorCode04 -> "PGRST204"
+  SchemaCacheErrorCode05 -> "PGRST205"
 
   JWTErrorCode00         -> "PGRST300"
   JWTErrorCode01         -> "PGRST301"
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -153,18 +153,20 @@ dbActionPlan dbAct conf apiReq sCache = case dbAct of
 
 wrappedReadPlan :: QualifiedIdentifier -> AppConfig -> SchemaCache -> ApiRequest -> Bool -> Either Error CrudPlan
 wrappedReadPlan  identifier conf sCache apiRequest@ApiRequest{iPreferences=Preferences{..},..} headersOnly = do
-  rPlan <- readPlan identifier conf sCache apiRequest
-  (handler, mediaType)  <- mapLeft ApiRequestError $ negotiateContent conf apiRequest identifier iAcceptMediaType (dbMediaHandlers sCache) (hasDefaultSelect rPlan)
+  qi <- mapLeft ApiRequestError $ findTable identifier (dbTables sCache)
+  rPlan <- readPlan qi conf sCache apiRequest
+  (handler, mediaType)  <- mapLeft ApiRequestError $ negotiateContent conf apiRequest qi iAcceptMediaType (dbMediaHandlers sCache) (hasDefaultSelect rPlan)
   if not (null invalidPrefs) && preferHandling == Just Strict then Left $ ApiRequestError $ InvalidPreferences invalidPrefs else Right ()
-  return $ WrappedReadPlan rPlan SQL.Read handler mediaType headersOnly identifier
+  return $ WrappedReadPlan rPlan SQL.Read handler mediaType headersOnly qi
 
 mutateReadPlan :: Mutation -> ApiRequest -> QualifiedIdentifier -> AppConfig -> SchemaCache -> Either Error CrudPlan
 mutateReadPlan  mutation apiRequest@ApiRequest{iPreferences=Preferences{..},..} identifier conf sCache = do
-  rPlan <- readPlan identifier conf sCache apiRequest
-  mPlan <- mutatePlan mutation identifier apiRequest sCache rPlan
+  qi <- mapLeft ApiRequestError $ findTable identifier (dbTables sCache)
+  rPlan <- readPlan qi conf sCache apiRequest
+  mPlan <- mutatePlan mutation qi apiRequest sCache rPlan
   if not (null invalidPrefs) && preferHandling == Just Strict then Left $ ApiRequestError $ InvalidPreferences invalidPrefs else Right ()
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3795, Clarify `Accept: vnd.pgrst.object` error message - @steve-chavez
  - #3779, Always log the schema cache load time - @steve-chavez
  - #3706, Fix insert with `missing=default` uses default value of domain instead of column - @taimoorzaeem
+ - #3697, Handle queries on non-existing table gracefully - @taimoorzaeem
```
steve-chavez, line 29: Seeing the change in the tests output, I think this should also be listed in `### Changed`.

test/io/test_io.py
```diff
@@ -973,11 +973,10 @@ def test_log_level(level, defaultenv):
...
                 output[2],
             )
+
+            assert len(output) == 5
             assert "Connection" and "is available" in output[3]
-            assert "Connection" and "is available" in output[4]
-            assert "Connection" and "is used" in output[5]
-            assert "Connection" and "is used" in output[6]
-            assert len(output) == 7
+            assert "Connection" and "is used" in output[4]
```
steve-chavez, line 979: Hm, these lines I don't understand. Why does `len(output)` change from 7 to 5?

test/spec/Feature/Query/InsertSpec.hs
```diff
@@ -477,7 +477,7 @@ spec actualPgVersion = do
             {"id": 204, "body": "yyy"},
             {"id": 205, "body": "zzz"}]|]
           `shouldRespondWith`
-          [json|{} |]
+          [json|{"code":"PGRST205","details":null,"hint":null,"message":"Could not find relation 'test.garlic' in the schema cache"} |]
```
steve-chavez, line 480: Hm, so the empty body `{}` on an insert error (https://github.com/PostgREST/postgrest/issues/3697) was tested. The behavior is a bit weird, it must have escaped review before.

test/spec/Feature/Query/QuerySpec.hs
```diff
@@ -617,14 +622,12 @@ spec = do
...
+      -- the existence of first table
+      it "table not found error if first table does not exist" $
         get "/car_model_sales_202101?select=id,name,car_models(id,name)&order=id.asc" `shouldRespondWith`
-          [json|
-            {"hint":"Perhaps you meant 'car_model_sales' instead of 'car_model_sales_202101'.",
-             "details":"Searched for a foreign key relationship between 'car_model_sales_202101' and 'car_models' in the schema 'test', but no matches were found.",
-             "code":"PGRST200",
-             "message":"Could not find a relationship between 'car_model_sales_202101' and 'car_models' in the schema cache"} |]
-          { matchStatus  = 400
+          [json| {"code":"PGRST205","details":null,"hint":null,"message":"Could not find relation 'test.car_model_sales_202101' in the schema cache"} |]
```
steve-chavez, line 629: Could we add a hint that suggest a table name?. This is good when typos happen. IIRC there's already some fuzzy search library in place for this. @laurenceisla Did some previous work for this.

steve-chavez:
Now that we rely on the schema cache for simple table filters, we need to make sure these requests are fast when a schema cache load for relationships is slow (https://github.com/PostgREST/postgrest/issues/3046).

There's a test that validates the above behavior (a request with resource embedding is blocked until the schema cache finishes loading):

https://github.com/PostgREST/postgrest/blob/52591c0352080736ffe0ad1bb37dcbba99b86a35/test/io/test_big_schema.py#L10-L34

We need an additional test that asserts that a table request (with no resource embedding) finishes in a short `plan_dur`  after reloading the schema cache (calling `/rpc/notify_pgrst` in the test).

CHANGELOG.md
steve-chavez: > Does that mean you would not consider backpatching this?

Yeah, I was thinking no backpatch. Just to be extra careful.

>  So this is hardly breaking any existing workflows, right?

I know some users that rely on table requests with no resource embedding being fast (in presence of slow relationships loads), if that has changed (test mentioned on https://github.com/PostgREST/postgrest/pull/3869#issuecomment-2605438040) then it would certainly break existing workflows.

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -109,7 +109,15 @@ spec =
 
     context "totally unknown route" $
       it "fails with 404" $
-        request methodDelete "/foozle?id=eq.101" [] "" `shouldRespondWith` 404
+        request methodDelete "/foozle?id=eq.101" [] ""
+          `shouldRespondWith` [json|
+            { "hint": null, "details":null,
+              "code":"PGRST205",
+              "message":"Could not find relation 'test.foozle' in the schema cache"
```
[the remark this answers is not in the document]
taimoorzaeem: Couldn't find any tests where we do it for tables, but we do it for functions.

https://github.com/PostgREST/postgrest/blob/e2dd4354d5e462ec9364c6237a43c7be650a9803/test/spec/Feature/Query/RpcSpec.hs#L1145

test/spec/Feature/Query/QuerySpec.hs
taimoorzaeem: > Could we add a hint that suggest a table name?. This is good when typos happen.

Sounds great!

> IRC there's already some fuzzy search library in place for this. @laurenceisla Did some previous work for this.

Oh yes,  found it: #2575

test/spec/Feature/Query/InsertSpec.hs
taimoorzaeem: A quick `git blame` reveals that it was introduced in #2542.

test/io/test_io.py
taimoorzaeem: I guess this changed because the output of
```python
response = postgrest.session.get("/unknown")
```
changed in this test. Not sure exactly how it got from 7 to 5.

laurenceisla: Should be because it's now a PGRST error and it doesn't hit the database anymore (no need to use a connection).

taimoorzaeem:
We are still using `NotFound` error in a couple places, which returns an empty error json:

https://github.com/PostgREST/postgrest/blob/e2dd4354d5e462ec9364c6237a43c7be650a9803/src/PostgREST/ApiRequest/Types.hs#L82

https://github.com/PostgREST/postgrest/blob/e2dd4354d5e462ec9364c6237a43c7be650a9803/src/PostgREST/Error.hs#L135

This seems fine for now, but maybe should be improved later (add error msg, details etc) for better UX.

taimoorzaeem:
Could this get a review? Seems good to merge.

src/PostgREST/Error.hs
```diff
@@ -253,6 +255,12 @@ instance JSON.ToJSON ApiRequestError where
   toJSON (ColumnNotFound relName colName) = toJsonPgrstError
     SchemaCacheErrorCode04 ("Could not find the '" <> colName <> "' column of '" <> relName <> "' in the schema cache") Nothing Nothing
 
+  toJSON (TableNotFound schemaName relName tbls) = toJsonPgrstError
+    SchemaCacheErrorCode05
+    ("Could not find relation '" <> schemaName <> "." <> relName <> "' in the schema cache")
```
steve-chavez, line 260: The "relation" term might be confused with "relationship". I think we can be more precise here and instead refer to "table" or "view", we have a `tableIsView :: Bool` field already.

steve-chavez: Although it doesn't make sense to discriminate table or view  if we don't know what we're looking for. Maybe you could omit the "relation" term altogether, or change it to "table or view".

src/PostgREST/Error.hs
```diff
@@ -350,6 +358,16 @@ noRpcHint schema procName params allProcs overloadedProcs =
       | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
       | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params)
 
+-- |
+-- Do a fuzzy search in all tables in the same schema and return closest result
+tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
+tableNotFoundHint schema tblName tblList
+  = fmap (\tbl -> "Perhaps you meant the relation '" <> schema <> "." <> tbl <> "'") perhapsTable
```
steve-chavez, line 365: Ditto on "relation"

src/PostgREST/Plan.hs
```diff
@@ -326,8 +326,8 @@ resolveQueryInputField ctx field opExpr = withTextParse ctx $ resolveTypeOrUnkno
 -- | Builds the ReadPlan tree on a number of stages.
 -- | Adds filters, order, limits on its respective nodes.
 -- | Adds joins conditions obtained from resource embedding.
-readPlan :: QualifiedIdentifier -> AppConfig -> SchemaCache -> ApiRequest -> Either Error ReadPlanTree
-readPlan qi@QualifiedIdentifier{..} AppConfig{configDbMaxRows, configDbAggregates} SchemaCache{dbTables, dbRelationships, dbRepresentations} apiRequest  =
+readPlan :: Bool -> QualifiedIdentifier -> AppConfig -> SchemaCache -> ApiRequest -> Either Error ReadPlanTree
+readPlan fromCallPlan qi@QualifiedIdentifier{..} AppConfig{configDbMaxRows, configDbAggregates} SchemaCache{dbTables, dbRelationships, dbRepresentations} apiRequest  =
```
steve-chavez, line 330: This `Bool` here is confusing, looks like it breaks the abstraction. Is there another way to do this?

taimoorzaeem: > Let's try and do it correctly now, even if it requires a refactor. Otherwise tech debt accrues and refactors become more laborious later.

Agreed.

src/PostgREST/Plan.hs
```diff
@@ -153,18 +153,20 @@ dbActionPlan dbAct conf apiReq sCache = case dbAct of
 
 wrappedReadPlan :: QualifiedIdentifier -> AppConfig -> SchemaCache -> ApiRequest -> Bool -> Either Error CrudPlan
 wrappedReadPlan  identifier conf sCache apiRequest@ApiRequest{iPreferences=Preferences{..},..} headersOnly = do
-  rPlan <- readPlan identifier conf sCache apiRequest
-  (handler, mediaType)  <- mapLeft ApiRequestError $ negotiateContent conf apiRequest identifier iAcceptMediaType (dbMediaHandlers sCache) (hasDefaultSelect rPlan)
+  qi <- mapLeft ApiRequestError $ findTable identifier (dbTables sCache)
+  rPlan <- readPlan qi conf sCache apiRequest
```
taimoorzaeem, line 157: Similar to how to we `findProc` before building the call plan, I think this approach should suffice without needing a lot of refactoring for a quick fix.

steve-chavez: Good thinking! :100: Code looks good now.

src/PostgREST/Error.hs
```diff
@@ -285,6 +288,12 @@ instance JSON.ToJSON ApiRequestError where
   toJSON (ColumnNotFound relName colName) = toJsonPgrstError
     SchemaCacheErrorCode04 ("Could not find the '" <> colName <> "' column of '" <> relName <> "' in the schema cache") Nothing Nothing
 
+  toJSON (TableNotFound schemaName relName tbls) = toJsonPgrstError
+    SchemaCacheErrorCode05
+    ("Could not find the table or view '" <> schemaName <> "." <> relName <> "' in the schema cache")
```
steve-chavez, line 293: Looking at the error messages now, `table or view` only adds verbosity and it's not really helpful.

Additionally, I think API users should be kept in the dark about if the server uses a table or a view.

The most generic term would be "resource", but that will cause some head scratching.

So I think we should just go with "table" as the more generic term.

```suggestion
    ("Could not find the table '" <> schemaName <> "." <> relName <> "' in the schema cache")
```

src/PostgREST/Error.hs
```diff
@@ -382,6 +391,16 @@ noRpcHint schema procName params allProcs overloadedProcs =
       | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
       | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params)
 
+-- |
+-- Do a fuzzy search in all tables in the same schema and return closest result
+tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
+tableNotFoundHint schema tblName tblList
+  = fmap (\tbl -> "Perhaps you meant the table or view '" <> schema <> "." <> tbl <> "'") perhapsTable
```
steve-chavez, line 398: Ditto:
```suggestion
  = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
```

steve-chavez (approved):

taimoorzaeem:
I don't think this should be backported anyways as discussed in [#3869 (comment)](https://github.com/PostgREST/postgrest/pull/3869#discussion_r1924409576).

Issue #2529, opened 2022-10-25
Show better errors for naming collisions when embedding


wolfgangwalther:
Against the current spec test schema:

```
GET /organizations?select=id,referee(id,name),referee(id,name)
{"code":"42712","details":null,"hint":null,"message":"table name \"organizations_referee_1\" specified more than once"}
```

```
GET /organizations?select=id,organizations!referee(id,name),organizations!auditor(id,name)
{"code":"42803","details":null,"hint":null,"message":"aggregate functions are not allowed in FROM clause of their own query level"}
```

While the first one is kind of ok, because we can guess what the problem is, the second one is not.

We should error out when the select spec we have results in two columns with the same name instead of trying to build a query.

steve-chavez:
I think it'd be better to err and prevent confusions. Ignoring already caused problems as seen on https://github.com/PostgREST/postgrest/issues/1771 (wrong queries can remain unnoticed)

> Maybe it would be more consistent to just throw away duplicate values, similar to how query strings are parsed in general.
> GET /rpc/getitemrange?min=1&min=2&max=4&max=5
> This will use min=2&max=5 only.

This is usual behavior on web applications AFAICT but our `select` param doesn't have to follow that - we can make it strict.

steve-chavez:
Yes, agree. It should be done at the planner stage. Maybe by handling that the name or alias of the node is unique on the select tree.

steve-chavez:
Bumped into this error when trying to use the new spread-to-many:

```bash
curl 'localhost:3000/titles?select=people!title_directors(director:primary_name),people!title_cast(actor:primary_name)'

{"code":"42803","details":null,"hint":null,"message":"aggregate functions are not allowed in FROM clause of their own query level"}
```

Note that the embedded table name should not matter here, as it shouldn't be included in the end result.

PR #3885, opened 2025-01-29
feat: allow aliases in spread embeddings

laurenceisla:
As mentioned here: https://github.com/PostgREST/postgrest/pull/3640#discussion_r1931147608


steve-chavez:
> Otherwise, without the aliases, it would return an error (the current implementation).

What's the error message?

> As mentioned here: https://github.com/PostgREST/postgrest/pull/3640#discussion_r1931147608, this is going to be more commonly used in to-many spreads.

Yeah, I haven't seen a use case for to-one spreads.

laurenceisla:
> What's the error message?

This issue mentions it better #2529

laurenceisla:
> Otherwise it wouldn't be possible to know which one to filter.

To give an example, something like this could be possible:

```
curl 'localhost:3000/factories?select=...p1:processes(batch_count:count()),...p2:processes(total_count:count())&p1.category_id=eq.1'
```

steve-chavez:
> To give an example, something like this could be possible:

For filtering it does make sense.

Maybe for another issue, I just found the following error:

```
curl 'localhost:3000/projects?select=*,...clients()'
{"code":"42601","details":null,"hint":null,"message":"syntax error at or near \"FROM\""}
```

PR #3895, opened 2025-02-01
fix: regression on #2524 where schema reload notification ceases to work on windows by updating hasql-notifications

diogob:
The issue was caused by hasql-notifications dropping windows specific code. 
Update hasql-notifications to newly fixed  version


nix/overlays/haskell-packages.nix
```diff
@@ -58,6 +58,14 @@ let
         }
         { });
 
+      hasql-notifications = lib.dontCheck (prev.callHackageDirect
```
develop7, line 61: @wolfgangwalther do we need `lib.dontCheck` here?

PR #3896, opened 2025-02-03
fix: log 503 client error to stderr

### LINKED ISSUE #3841: PostgREST does not log 503 errors to stderr
### Environment

* PostgreSQL version: 15.6
* PostgREST version: 12.2.3
* Operating system: ubuntu

### Description of issue

PostgREST logged the following (requests queries redacted):

```console
127.0.0.1 - service_role [06/Jan/2025:17:44:02 +0000] "GET /rpc/<an_rpc> HTTP/1.1" 503 - "" "node"
06/Jan/2025:17:44:03 +0000: Successfully connected to PostgreSQL 15.6 ...
06/Jan/2025:17:44:03 +0000: Config reloaded
06/Jan/2025:17:44:03 +0000: Schema cache queried in 35.5 milliseconds
06/Jan/2025:17:44:03 +0000: Schema cache loaded ...
127.0.0.1 - service_role [06/Jan/2025:17:44:02 +0000] "GET /rpc/<another_rpc> HTTP/1.1" 503 - "" "node"
06/Jan/2025:17:44:03 +0000: Successfully connected to PostgreSQL 15.6 ...
06/Jan/2025:17:44:03 +0000: Config reloaded
06/Jan/2025:17:44:04 +0000: Schema cache queried in 37.9 milliseconds
06/Jan/2025:17:44:04 +0000: Schema cache loaded ...
```

As seen above it does not log the 503 error (when it should) and retries connection immediately. For instance, the 500 errors are logged correctly:

```console
06/Jan/2025:17:44:59 +0000: {"code":"57014","details":null,"hint":null,"message":"canceling statement due to statement timeout"}
127.0.0.1 - service_role [06/Jan/2025:17:44:59 +0000] "GET /rpc/<an_rpc> HTTP/1.1" 500 - "" "node"
```

---

Now, checking the PostgreSQL logs, it shows that it's an error related to. In particular, it shows these two errors:

```
message:"canceling statement due to conflict with recovery"
detail:"User query might have needed to see row versions that must be removed."
```
```
message:"terminating connection due to conflict with recovery"
detail:"User query might have needed to see row versions that must be removed."
hint:"In a moment you should be able to reconnect to the database and repeat your command."
```

Not sure if PostgREST should try to reconnect or not in these cases, since a solution is to set `max_standby_archive_delay` and `max_standby_streaming_delay` in PG config to avoid the error.

taimoorzaeem:
Closes #3841.

Should this go under feat? Maybe we should introduce new commit prefix for logging like maybe `log: ...`?

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -261,8 +261,9 @@ usePool AppState{stateObserver=observer, stateMainThreadId=mainThreadId, ..} ses
         SQL.ServerError{} ->
           when (Error.status (Error.PgError False err) >= HTTP.status500) $
             observer $ QueryErrorCodeHighObs err
-    SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _)) ->
-      pure ()
+    err@(SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _))) ->
+      -- An error on the client-side, usually indicates problems wth connection
+        observer $ QueryErrorCodeHighObs err
     )
 
   return res
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
    + Fixed `"column <json_aggregate>.<alias> does not exist"` error when selecting `?select=...table(aias:count())`
  - #3727, Clarify "listening" logs - @steve-chavez
  - #3795, Clarify `Accept: vnd.pgrst.object` error message - @steve-chavez
+ - #3841, Log `503` client error to stderr - @taimoorzaeem
 
 ### Changed
 
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1696,3 +1696,22 @@ def test_jwt_cache_purges_expired_entries(defaultenv):
         response = postgrest.session.get("/authors_only", headers=hdrs3)
 
         assert response.status_code == 200
+
+
+def test_pgrst_log_503_client_error_to_stderr(defaultenv):
+    "PostgREST should log 503 errors to stderr"
+
+    env = {
+        **defaultenv,
+        "PGAPPNAME": "test-io",
+    }
+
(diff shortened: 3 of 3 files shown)


steve-chavez:
> Since there is nothing really broken, I wouldn't bother with backpatching

The `crit` level would be kinda broken as mentioned in https://github.com/PostgREST/postgrest/issues/3841#issuecomment-2634666007. So this should be a `fix:..`.

src/PostgREST/AppState.hs
```diff
@@ -261,8 +261,10 @@ usePool AppState{stateObserver=observer, stateMainThreadId=mainThreadId, ..} ses
         SQL.ServerError{} ->
           when (Error.status (Error.PgError False err) >= HTTP.status500) $
             observer $ QueryErrorCodeHighObs err
-    SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _)) ->
-      pure ()
+    err@(SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _))) ->
+      -- logging since https://github.com/PostgREST/postgrest/issues/3841
+      when (Error.status (Error.PgError False err) >= HTTP.status500) $
```
taimoorzaeem, line 266: Hmm, `code-cov` is complaining on `False`. This is because we always return `503` here without evaluating `False`.

https://github.com/PostgREST/postgrest/blob/3f78615dff51f680b141c3063a328487efe29e29/src/PostgREST/Error.hs#L468

laurenceisla: > This is because we always return `503`.

Oh, right, since we always return `503` we don't need the `when...` condition then, only the observer, since this should always be logged.

As a side note, I'd remove the comment `-- logging since...`, or explain directly that [this usually indicates problems with connection](https://hackage.haskell.org/package/hasql-1.9.1.1/docs/Hasql-Session.html#v:ClientError) and returns `503`, so we log it.

steve-chavez (approved):

PR #3889, opened 2025-01-30
perf: Purge JWT cache asynchronously in a separate thread

mkleczek:
A follow-up to the JWT cache purging fix - looks like we still left expensive operation on a hot path affecting response times for request triggering cache miss.

A simple fix is to purge in a separate thread started upon a cache miss.

PR #3802, opened 2024-11-26
feat: add metric pgrst_jwt_cache_size_bytes in admin server

taimoorzaeem:
Add metric `pgrst_jwt_cache_size_bytes` in admin server which shows the cache size in bytes.

PR #3905, opened 2025-02-05
perf: Backport of #3889 to v12

mkleczek:
See https://github.com/PostgREST/postgrest/pull/3889

@steve-chavez @taimoorzaeem
I've prepared a backport of asynchronous JWT cache purge to 12.
Looks like there is consistent (small but noticable) performance gain.

Done this as it is not clear when next major is going to be released (and if any JWT caching improvements are going to be included).

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -5,6 +5,7 @@
 module PostgREST.AppState
   ( AppState
   , AuthResult(..)
+  , JwtCacheState(..)
   , destroy
   , getConfig
   , getSchemaCache
@@ -13,7 +14,7 @@ module PostgREST.AppState
   , getNextDelay
   , getNextListenerDelay
   , getTime
-  , getJwtCache
+  , getJwtCacheState
   , getSocketREST
   , getSocketAdmin
   , init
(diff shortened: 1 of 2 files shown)


mkleczek:
> > Looks like there is consistent (small but noticable) performance gain.
> 
> How did you test that?

https://github.com/PostgREST/postgrest/actions/runs/13158254898

mkleczek:
It is difficult for me to judge the effectiveness of existing load tests.
The results are consistent with tests performed earlier on main:
https://github.com/PostgREST/postgrest/actions/runs/13151068303

So it seems that **even** in case of small JWT cache with fresh tokens taking `purgeExpired` from the hot path provides response time gains.

steve-chavez:
I don't think we should backport this. The `MVar` logic looks risky, it could cause breakage

mkleczek:
@steve-chavez
not questioning your decision but could you elaborate a little on the risks of MVar locking logic? I find it pretty straightforward.

steve-chavez:
> So it seems that even in case of small JWT cache with fresh tokens taking purgeExpired from the hot path provides response time gains.

Yes, here https://github.com/PostgREST/postgrest/pull/3889#issuecomment-2755418144 it's revealed there's indeed perf gains.

>  but could you elaborate a little on the risks of MVar locking logic? I find it pretty straightforward.

Mostly that it doesn't seem good for maintenance to have it like that, I was thinking it should be in a function. It's looking like this logic could be inside the debounce added in https://github.com/PostgREST/postgrest/pull/3802 or perhaps another debounce.

---

Let's continue on https://github.com/PostgREST/postgrest/pull/3889. This should go into v13.

Issue #3013, opened 2023-10-17
Ordering embedded resources for PATCH request is broken


wolfgangwalther:
When embedding resources on a PATCH request, ordering the embedded resources does not work. The following test-case fails on latest main:

```patch
diff --git a/test/spec/Feature/Query/UpdateSpec.hs b/test/spec/Feature/Query/UpdateSpec.hs
index 362a00e3..53fff125 100644
--- a/test/spec/Feature/Query/UpdateSpec.hs
+++ b/test/spec/Feature/Query/UpdateSpec.hs
@@ -435,6 +435,18 @@ spec actualPgVersion = do
           , matchHeaders = [matchContentTypeJson, "Preference-Applied" <:> "return=representation"]
           }
 
+      it "with ordering" $
+        request methodPatch "/web_content?id=eq.0&select=id,name,web_content(name)&web_content.order=name.asc"
+                [("Prefer", "return=representation")]
+          [json|{"name": "tardis-patched"}|]
+          `shouldRespondWith`
+          [json|
+            [ { "id": 0, "name": "tardis-patched", "web_content": [ { "name": "bar" }, { "name": "fezz" }, { "name": "foo" } ]} ]
+          |]
+          { matchStatus  = 200
+          , matchHeaders = [matchContentTypeJson, "Preference-Applied" <:> "return=representation"]
+          }
+
       it "with filters" $
         request methodPatch "/web_content?id=eq.0&select=id,name,web_content(name)&web_content.name=like.f*"
                 [("Prefer", "return=representation")]
```

This started failing between v9 and v10 with commit e90391b7ac.

Similar issues regarding embedded filters were fixed in #2618 and #2752, but I am not sure whether they were introduced at the same time.

I think it would make sense to test all embedding features systematically for all kinds of requests - possibly more is broken than just this?

steve-chavez:
@wolfgangwalther I've been finding that users find the limited updated/delete really confusing and additionally it causes bugs like this. Also it's not that useful, I believe it's more of a backend process for which you can just use a function or maybe a direct pg connection.

How about if we drop the feature? I'm intending to make a breaking change for https://github.com/PostgREST/postgrest/pull/2825

steve-chavez:
With https://github.com/PostgREST/postgrest/pull/3907 done, I think closing this should only be a matter of adding more tests.

taimoorzaeem:
[...]

PR #3907, opened 2025-02-09
break: remove limited updated/delete feature

steve-chavez:
BREAKING CHANGE

As agreed on https://github.com/PostgREST/postgrest/issues/3013#issuecomment-1770186262, this removes the [limited update/delete feature](https://docs.postgrest.org/en/v12/references/api/tables_views.html#limited-update-delete).

The feature was complicated, largely unused and caused other bugs in mutations.

It was added in #2195 and #2211.


steve-chavez:
Any opposition for this one? Want to get this in the upcoming major (let's release it after https://github.com/PostgREST/postgrest/pull/3640 is merged).

src/PostgREST/Error.hs
```diff
@@ -140,15 +138,6 @@ instance JSON.ToJSON ApiRequestError where
     Nothing
     (Just $ JSON.String $ "Verify that '" <> resource <> "' is included in the 'select' query parameter.")
 
-  toJSON LimitNoOrderError = toJsonPgrstError
-    ApiRequestErrorCode09 "A 'limit' was applied without an explicit 'order'" Nothing (Just "Apply an 'order' using unique column(s)")
-
-  toJSON (OffLimitsChangesError n maxs) = toJsonPgrstError
-    ApiRequestErrorCode10
```
laurenceisla, line 147: I think those `ApiRequestErrorCodexx` should be commented out below, in the definition, with "used to be mapped to...".

steve-chavez:
> largely unused
> Can we really tell? I don't think we can make that judgement.

I was mostly going by the fact that the error codes for this feature were never searched on the docs, nor an issue was ever reported. But yeah, it's not possible to determine it 100%. 

Stack and FreeBSD CI jobs are failing for something unrelated. 

Will merge this now, maybe it'll make some other PRs easier to review too.

steve-chavez:
@wolfgangwalther I've noted this change was done quite dangerously for v13. So on v12.2.12, when doing this:

```bash
$ curl -X DELETE 'localhost:3000/items?limit=1&order=id'

$ curl localhost:3000/items
[{"id":2}, 
 {"id":3}, 
 {"id":4}, 
 {"id":5}, 
 {"id":6}, 
 {"id":7}, 
 {"id":8}, 
 {"id":9}, 
 {"id":10}, 
 {"id":11}, 
 {"id":12}, 
 {"id":13}, 
 {"id":14}, 
 {"id":15}]
```

Clearly the DELETE was limited to 1 row. But now on v13 doing the same request:

```bash
$ curl -X DELETE 'localhost:3000/items?limit=1&order=id'

$ curl localhost:3000/items
[]
```

The above succeeds silently :open_mouth: . So any user relying on this behavior won't be alerted of this breaking change when upgrading :crying_cat_face:.

What do you think if we proceed now with a fix that does what you mentioned before:

> Disallow limit and offset on mutation requests entirely. Those are not useful and likely to lead to confusion.
> https://github.com/PostgREST/postgrest/issues/3013#issuecomment-1770186262

So err in case of limit/offset on mutation, and then do a patch release (13.0.7)?

wolfgangwalther:
> So err in case of limit/offset on mutation, and then do a patch release (13.0.7)?

In principle yes, if the change is simple enough.

PR #3908, opened 2025-02-11
docs: add missing jwt claims and clock skew

### LINKED ISSUE #3783: docs: JWT allowed skew is not documented
## Problem

We allow  a 30 second skew when doing JWT `exp` validation:

https://github.com/PostgREST/postgrest/blob/f3aa00a8384981baf9076c26b64ccede02bfa365/src/PostgREST/Auth.hs#L85

But this isn't documented anywhere.

taimoorzaeem:
Closes #3783.

### DIFF
--- a/docs/references/auth.rst
+++ b/docs/references/auth.rst
@@ -159,7 +159,15 @@ You can specify the literal value as we saw earlier, or reference a filename to 
 JWT Claims Validation
 ~~~~~~~~~~~~~~~~~~~~~
 
-PostgREST honors the :code:`exp` claim for token expiration, rejecting expired tokens.
+PostgREST honors the following `JWT claims <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4>`_:
+
+- ``exp`` Expiration Time
+- ``iat`` Issued At
+- ``nbf`` Not Before
+- ``aud`` Audience, see :ref:`jwt-aud`
+
+.. note::
+  PostgREST allows a ``30`` seconds clock skew for validating the time-based claims. This means that it allows an extra 30 seconds for that claim before rejecting it.
 
 .. _jwt_role_claim_key_extract:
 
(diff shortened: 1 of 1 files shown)


docs/references/auth.rst
```diff
@@ -159,7 +159,15 @@ You can specify the literal value as we saw earlier, or reference a filename to
...
-PostgREST honors the :code:`exp` claim for token expiration, rejecting expired tokens.
+PostgREST honors the following `JWT claims <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4>`_:
+
+- ``exp`` Expiration Time
+- ``iat`` Issued At
+- ``nbf`` Not Before
+- ``aud`` Audience, see :ref:`jwt-aud`
+
+.. note::
+  PostgREST allows a ``30`` seconds clock skew for validating the time-based claims. This means that it allows an extra 30 seconds for that claim before rejecting it.
```
steve-chavez, line 170: Making the claims explicit:

```suggestion
  PostgREST allows for a 30-second clock skew when validatiing the `exp` and `iat` claims. In other words, it gives an extra 30 seconds before the token is rejected if there is a slight discrepancy in the timestamps.
```

docs/references/auth.rst
```diff
@@ -159,7 +159,15 @@ You can specify the literal value as we saw earlier, or reference a filename to
...
-PostgREST honors the :code:`exp` claim for token expiration, rejecting expired tokens.
+PostgREST honors the following `JWT claims <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4>`_:
+
+- ``exp`` Expiration Time
+- ``iat`` Issued At
+- ``nbf`` Not Before
+- ``aud`` Audience, see :ref:`jwt-aud`
+
+.. note::
+  PostgREST allows for a 30-second clock skew when validatiing the `exp` and `iat` claims. In other words, it gives an extra 30 seconds before the token is rejected if there is a slight discrepancy in the timestamps.
```
steve-chavez, line 170:
```suggestion
  PostgREST allows for a 30-second clock skew when validating the ``exp`` and ``iat`` claims. In other words, it gives an extra 30 seconds before the token is rejected if there is a slight discrepancy in the timestamps.
```

Issue #3600, opened 2024-06-17
Improve error messages for JWT verification errors


wolfgangwalther:
In #3598 I added a new IO test showing how many JWT errors we currently expose by just showing internal errors via `show`. Those are not very user friendly:

https://github.com/PostgREST/postgrest/blob/465170c7d6c51dace3faf68fbb4040353d5571c0/test/io/test_io.py#L78-L128

For at least a few of them we should improve the error message and maybe differentiate them in https://postgrest.org/en/v12/references/errors.html#group-3-jwt? I.e. give them our own PGRST error code.

laurenceisla:
Related issue: https://github.com/PostgREST/postgrest/issues/1840

> I.e. give them our own PGRST error code.

Agree.

PR #3911, opened 2025-02-12
refactor: move `ApiRequestError` to Error module

taimoorzaeem:
This is first steps to improving our error infrastructure.
### Reasons for this refactor

- All error types should originate from `Error` module
- Removes re-exporting types
- No need to depend on an extra module (namely `ApiRequest/Types`).
- Would help in representing errors internally to something similar to how we group them in [docs](https://docs.postgrest.org/en/v12/references/errors.html#postgrest-error-codes)
- Would eventually help in adding new error types for error related issues (eg. #3600)


steve-chavez (approved):
Looks good! Will merge once you rebase.

PR #3925, opened 2025-02-27
feat: Log PoolRequest and PoolRequestFullfilled observations

### LINKED ISSUE #3903: Log `PoolRequest` and `PoolRequestFullfilled` observations
### Discussed in https://github.com/PostgREST/postgrest/discussions/3899

<sup>Originally posted by **taimoorzaeem** February  4, 2025</sup>
### Problem

While working on #3802, we noticed in [comment#3802](https://github.com/PostgREST/postgrest/pull/3802#issuecomment-2600281530) that we are not logging all observations. Namely `PoolRequest` and `PoolRequestFullfilled` (because we are only using them for metrics). Infact, we are just logging empty here for them here:

https://github.com/PostgREST/postgrest/blob/db85e64971145e80359f586db4ca6e8f2231a53c/src/PostgREST/Observation.hs#L140-L142

This never gets executed because we intercept it in the `Logger.hs`:

https://github.com/PostgREST/postgrest/blob/db85e64971145e80359f586db4ca6e8f2231a53c/src/PostgREST/Logger.hs#L91-L94

### Solution

I think this needs a refactor and one way to solve this is to log these observations for `loglevel >= debug`. So it should be something like:

```haskell
-- Logger.hs
  o@PoolRequest ->
    when (logLevel >= LogDebug) $ do
      logWithZTime loggerState $ observationMessage o
  o@PoolRequestFullfilled ->
    when (logLevel >= LogDebug) $ do
      logWithZTime loggerState $ observationMessage o
```
and
```haskell
-- Observation.hs
 PoolRequest -> 
   "Trying to borrow a connection from pool"
 PoolRequestFullfilled -> 
   "Borrowed a connection from the pool"
```
What would be more important is to remove that mempty there, this to ensure we don't log empty lines accidentally like what happened on https://github.com/PostgREST/postgrest/pull/3802#issuecomment-2600281530
</div>

taimoorzaeem:
Closes #3903.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3747, Allow `not_null` value for the `is` operator - @taimoorzaeem
  - #2255, Apply `to_tsvector()` explicitly to the full-text search filtered column (excluding `tsvector` types) - @laurenceisla
  - #1578, Log the main SQL query to stderr at the current `log-level` when `log-query=main-query` - @laurenceisla
- - #3903, Log `PoolRequest` and `PoolRequestFullfilled` observations - @taimoorzaeem
+ - #3903, Log connection pool borrows on `log-level=debug` - @taimoorzaeem
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3747, Allow `not_null` value for the `is` operator - @taimoorzaeem
  - #2255, Apply `to_tsvector()` explicitly to the full-text search filtered column (excluding `tsvector` types) - @laurenceisla
  - #1578, Log the main SQL query to stderr at the current `log-level` when `log-query=main-query` - @laurenceisla
+ - #3903, Log `PoolRequest` and `PoolRequestFullfilled` observations - @taimoorzaeem
```
steve-chavez, line 19: More user facing:
```suggestion
 - #3903, Log connection pool borrows on `log-level=debug` - @taimoorzaeem
```

PR #3927, opened 2025-02-28
fix: remove caching empty JWT in Auth header

### LINKED ISSUE #3926: Throw better error message when empty JWT is sent in Auth header
## Problem

The request with empty JWT hits the db and fails with the permission denied error. The empty JWT is also cached when caching enabled.

```bash
$ curl localhost:3000/authors_only -H "Authorization: Bearer "
{"code":"42501","details":null,"hint":null,"message":"permission denied for table authors_only"}

```

This happens because we default to an empty string here:

https://github.com/PostgREST/postgrest/blob/1a287edf7ecd2bf1eeb3bc4d764b845dac45814d/src/PostgREST/Auth.hs#L57-L58

https://github.com/PostgREST/postgrest/blob/1a287edf7ecd2bf1eeb3bc4d764b845dac45814d/src/PostgREST/Auth.hs#L151-L152

## Solution

The JWT should be validated before hitting the db and a better error message should be given to the client.

taimoorzaeem:
Closes #3926.

### DIFF
--- a/test/spec/Feature/Auth/AuthSpec.hs
+++ b/test/spec/Feature/Auth/AuthSpec.hs
@@ -159,3 +159,14 @@ spec = describe "authorization" $ do
       `shouldRespondWith` 200
     request methodGet "/authors_only" [authHeader "bearer" token] ""
       `shouldRespondWith` 200
+
+  describe "when JWT is empty string" $ do
+
+    it "raises error on empty token in Auth header" $ do
+      let auth = authHeaderJWT ""
+      request methodGet "/authors_only"
+        [auth]
+        ""
+        `shouldRespondWith`
+        [json| {"code":"PGRST301","details":null,"hint":null,"message":"JWSError (CompactDecodeError Invalid number of parts: Expected 3 parts; got 2)"} |]
+        { matchStatus = 401 }
--- a/test/spec/Feature/Auth/JwtCacheSpec.hs
+++ b/test/spec/Feature/Auth/JwtCacheSpec.hs
@@ -1,0 +1,22 @@
+module Feature.Auth.JwtCacheSpec where
+
+import Network.Wai (Application)
+
+import Test.Hspec
+import Test.Hspec.Wai
+import Test.Hspec.Wai.JSON
+
+import Protolude hiding (get)
+
+spec :: SpecWith ((), Application)
+spec = describe "jwt cache enabled" $ do
+  it "denies access to tables that anonymous does not own" $
+    get "/authors_only" `shouldRespondWith`
+      [json| {
+        "hint":null,
+        "details":null,
+        "code":"42501",
+        "message":"permission denied for table authors_only"} |]
+      { matchStatus = 401
+      , matchHeaders = ["WWW-Authenticate" <:> "Bearer"]
+      }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3727, Clarify "listening" logs - @steve-chavez
  - #3795, Clarify `Accept: vnd.pgrst.object` error message - @steve-chavez
  - #3697, #3602, Handle queries on non-existing table gracefully - @taimoorzaeem
+ - #3926, Remove caching empty JWT when `Authorization: Bearer` is empty - @taimoorzaeem
 
 ### Changed
 
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -207,6 +207,7 @@ test-suite spec
                       Feature.Auth.AudienceJwtSecretSpec
                       Feature.Auth.AuthSpec
                       Feature.Auth.BinaryJwtSecretSpec
+                      Feature.Auth.JwtCacheSpec
                       Feature.Auth.NoAnonSpec
                       Feature.Auth.NoJwtSpec
                       Feature.ConcurrentSpec
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -54,9 +54,9 @@ import Protolude
 
 -- | Receives the JWT secret and audience (from config) and a JWT and returns a
 -- JSON object of JWT claims.
-parseToken :: AppConfig -> ByteString -> UTCTime -> ExceptT Error IO JSON.Value
-parseToken _ "" _ = return JSON.emptyObject
-parseToken AppConfig{..} token time = do
+parseToken :: AppConfig -> Maybe ByteString -> UTCTime -> ExceptT Error IO JSON.Value
+parseToken _ Nothing _ = return JSON.emptyObject
+parseToken AppConfig{..} (Just token) time = do
   secret <- liftEither . maybeToRight JwtTokenMissing $ configJWKS
   eitherContent <- liftIO $ JWT.decode (JWT.keys secret) Nothing token
   content <- liftEither . mapLeft jwtDecodeError $ eitherContent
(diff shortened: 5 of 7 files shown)


steve-chavez:
Hm, wasn't this supposed to be a `fix`?

taimoorzaeem:
@steve-chavez It didn't seem like a usual fix which is worth a changelog entry and a backpatch so just thought I should make it a refactor. This is a very little change which the users shouldn't need to worry about IMO. Should I change it to `fix:`?

steve-chavez:
I went from the "bug" type assigned in https://github.com/PostgREST/postgrest/issues/3926.

So there is no change of behavior and this only affects the JWT cache implementation?

taimoorzaeem:
> So there is no change of behavior and this only affects the JWT cache implementation?

Precisely.

steve-chavez:
This is confusing. If the JWT cache used to cache empty JWTs and now with this change it won't, why is it not a bug fix?

taimoorzaeem:
> This is confusing. If the JWT cache used to cache empty JWTs and now with this change it won't, why is it not a bug fix?

So, the only change is that:
```haskell
--- previously, jwt cache had an entry like
--- [key: "", value: {}]
--- this used to hit the db and failed

-- now it fails early and the cache will have one less entry
```

It is technically a very small fix. Just not an important one so as to have its own changelog entry because it has almost no effect.

src/PostgREST/Auth/JwtCache.hs
```diff
@@ -36,8 +36,9 @@ init = do
   return $ JwtCacheState cache
 
 -- | Used to retrieve and insert JWT to JWT Cache
-lookupJwtCache :: JwtCacheState -> ByteString -> Int -> IO (Either Error AuthResult) -> UTCTime -> IO (Either Error AuthResult)
-lookupJwtCache JwtCacheState{jwtCache} token maxLifetime parseJwt utc = do
+lookupJwtCache :: JwtCacheState -> Maybe ByteString -> Int -> IO (Either Error AuthResult) -> UTCTime -> IO (Either Error AuthResult)
+lookupJwtCache _ Nothing _ _ _ = return $ Left JwtTokenRequired
```
taimoorzaeem, line 40: Hmm, this line is blocks anonymous access. I'll fix it.

steve-chavez:
> But - is that the same for the user? Would any user ever complain that memory usage is a few bytes to high, because an empty JWT is cached? I don't think so. No user would ever report this to us. For them, this change is mostly irrelevant.

If it's a correction in behavior, I'd label is it a bug fix. I don't think it matters if it's benign or if a developer (and I do think I qualify as user too) reported it ([ref](https://github.com/PostgREST/postgrest/pull/3802#issuecomment-2689268756)).

It would have been different if this commit was done before a release of the JWT cache, then maybe it'd be a refactor (although in the past I've used a ["correct" commit message](https://github.com/PostgREST/postgrest/commit/1a8b6972a8e940bb59ec9240daf6c57fff9b5c09) for cases like this), but this is not the case.

> It didn't seem like a usual fix which is worth a changelog entry and a backpatch 

I would backpatch it btw, but not adamant on it. From a user perspective, I see patch releases as a sign of progress and I think that's always good.

test/spec/Feature/Auth/AuthSpec.hs
```diff
@@ -159,3 +159,14 @@ spec = describe "authorization" $ do
...
+  describe "when JWT is empty string" $ do
+
+    it "raises error on empty token in Auth header" $ do
+      let auth = authHeaderJWT ""
+      request methodGet "/authors_only"
+        [auth]
+        ""
+        `shouldRespondWith`
+        [json| {"code":"PGRST301","details":null,"hint":null,"message":"JWSError (CompactDecodeError Invalid number of parts: Expected 3 parts; got 2)"} |]
+        { matchStatus = 401 }
```
steve-chavez, line 172: This is a change in behavior from previous versions?

On `main` right now I'm doing:

```
$ curl localhost:3000/authors_only -H "Authorization: Bearer"
{"code":"42501","details":null,"hint":null,"message":"permission denied for table authors_only"}

$ curl localhost:3000/authors_only -H "Authorization: Bearer "
{"code":"42501","details":null,"hint":null,"message":"permission denied for table authors_only"}
```

Is that equivalent to the above test?

steve-chavez: > This is a change in behavior from previous versions?

If so, I think this is an improvement. Users could have been passing an empty JWT by accident and getting the db permission error instead of a decoding error.

> JWSError (CompactDecodeError Invalid number of parts: Expected 3 parts; got 2)

The above is a mysterious error, I wonder if it could be improved.

taimoorzaeem: > If so, I think this is an improvement.

Yes, it is an improvement.

> The above is a mysterious error, I wonder if it could be improved.

We have an improvement issue tracked here (#3600). But for now we can improve this one. How about?

```haskell
"JWSError (CompactDecodeError Invalid number of parts: Expected 3 parts in the JWT)"
```

test/spec/Feature/Auth/JwtCacheSpec.hs
```diff
@@ -0,0 +1,22 @@
...
+  it "denies access to tables that anonymous does not own" $
+    get "/authors_only" `shouldRespondWith`
+      [json| {
+        "hint":null,
+        "details":null,
+        "code":"42501",
+        "message":"permission denied for table authors_only"} |]
+      { matchStatus = 401
+      , matchHeaders = ["WWW-Authenticate" <:> "Bearer"]
+      }
```
steve-chavez, line 22: I don't understand the purpose of this test on this PR. Isn't this already tested here?

https://github.com/PostgREST/postgrest/blob/d0173e71923e15b412fd459e35e1d5b97161a363/test/spec/Feature/Auth/AuthSpec.hs#L17-L26

Is it a way to test the code path of the JWT cache being enabled and that this permission error happens?

I think what we need to test here is that the JWT cache metric doesn't increase when we add a new empty JWT. And that should be an io test. The permission error message can be tested there too.

test/spec/Feature/Auth/AuthSpec.hs
steve-chavez: > Yes, it is an improvement.

Ok cool. So then this is not really tied to the JWT cache issue and it is an important fix. How about a new focused PR for the "empty JWT case giving a good error message"?

test/spec/Feature/Auth/JwtCacheSpec.hs
taimoorzaeem: > Is it a way to test the code path of the JWT cache being enabled and that this permission error happens?

Yes , I mainly added this for code-coverage reasons. Hmm, this PR is only for the empty JWT, and the metric is on the #3802, so we can't test that here.

test/spec/Feature/Auth/AuthSpec.hs
taimoorzaeem: > How about a new focused PR for the "empty JWT case giving a good error message"?

Sure, as soon as this gets merged, I'll open the other PR.

taimoorzaeem: Also, this bad error message isn't just for "empty" JWT. We get the same for `Authorization: Bearer abc`. So a PR just for empty JWT would be enough.

taimoorzaeem: @steve-chavez  Since we are doing #3600 anyway, lets close this PR and move it to the aforementioned PR?

steve-chavez: > Since we are doing https://github.com/PostgREST/postgrest/issues/3600 anyway, lets close this PR and move it to the aforementioned PR?

Yes, sounds good.

taimoorzaeem:
Closing in favor of a new PR.

PR #3930, opened 2025-03-04
refactor: group jwt errors

taimoorzaeem:
Refactor for #3600. It would be easier adding new JWT errors with this refactor.


steve-chavez (approved):
LGTM :+1:

PR #3933, opened 2025-03-05
fix: improve jwt errors

### LINKED ISSUE #3600: Improve error messages for JWT verification errors

### LINKED ISSUE #3926: Throw better error message when empty JWT is sent in Auth header

taimoorzaeem:
BREAKING CHANGE

We were exposing internal errors via `Show`. This commit improves error messages for better user experience.
Closes #3600, closes #3926.

- [x] Implementation
- [x] Tests
- [x] Docs

I'll add docs once this design/change is approved.

### DIFF
--- a/src/PostgREST/Auth.hs
+++ b/src/PostgREST/Auth.hs
@@ -23,6 +23,7 @@ import qualified Data.Aeson.Key                  as K
 import qualified Data.Aeson.KeyMap               as KM
 import qualified Data.Aeson.Types                as JSON
 import qualified Data.ByteString                 as BS
+import qualified Data.ByteString.Internal        as BS
 import qualified Data.ByteString.Lazy.Char8      as LBS
 import qualified Data.Scientific                 as Sci
 import qualified Data.Text                       as T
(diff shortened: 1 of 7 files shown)


src/PostgREST/Auth.hs
```diff
@@ -54,33 +54,36 @@ import Protolude
...
-      jwtDecodeError (JWT.BadAlgorithm _) = JwtTokenInvalid "JWSError JWSNoSignatures"
-      jwtDecodeError e                    = JwtTokenInvalid $ show e
+      jwtDecodeError (JWT.KeyError _)     = JwtDecodeError "No suitable key or wrong key type"
+      jwtDecodeError (JWT.BadAlgorithm _) = JwtDecodeError "Wrong or unsupported encoding algorithm"
+      jwtDecodeError (JWT.BadDots dots)   = JwtDecodeError ("Wrong number of '.' periods in JWT: Expected 3, got " <> show dots)
+      jwtDecodeError (JWT.BadHeader _)    = JwtDecodeError "Header couldn't be decoded or contains bad data"
+      jwtDecodeError JWT.BadClaims        = JwtClaimsError "JWT claims couldn't be decoded or contains bad data"
+      jwtDecodeError JWT.BadSignature     = JwtDecodeError "The JWT signature couldn't be decoded or is invalid"
+      jwtDecodeError JWT.BadCrypto        = JwtDecodeError "Cryptographic operation while decoding failed"
+      jwtDecodeError (JWT.Base64Error _)  = JwtDecodeError "A base64 decoding error has occured"
```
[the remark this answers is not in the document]
steve-chavez: Thinking more about it, maybe some of these error messages can be more generic? Some of them look too technical for API users.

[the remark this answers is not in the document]
laurenceisla: > Some of them look too technical for API users.

Specially the "Cryptographic..." error, since it also happens when the JWS is invalid. Kinda scary for the user to get that message when the JWT structure is not correct. To take another example, note that we do not raise the [`JWT.BadDots` error ](https://github.com/search?q=repo%3Atekul%2Fjose-jwt+BadDots&type=code) anywhere since we don't use the `decodeClaims` function from the library afaict, so I wouldn't bother in giving it a custom message.

>  Not sure yet, I think we would need some tests to see under what scenarios each one happen.

Yes, all in all, we need to check where are the errors raised in our context and add the message accordingly. Our tests should help in pinpointing them.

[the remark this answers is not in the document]
taimoorzaeem: > Some of them look too technical for API users

That's exactly why I posed this question in the comment
```haskell
-- should we validate JWT ourselves before decoding or let the library (not very user friendly with errors) handle it?
```

The library errors are just not very user friendly. So maybe let's validate them ourselves so we can give better and precise error messages. But yeah, we need tests before we can decide further.

> Yes, all in all, we need to check where are the errors raised in our context and add the message accordingly. Our tests should help in pinpointing them

Agreed.

src/PostgREST/Auth.hs
```diff
@@ -54,33 +55,48 @@ import Protolude
...
-      -- TODO: Improve errors, those were just taken as-is from hs-jose to avoid
-      -- breaking changes.
+      hasThreeParts :: ByteString -> Either Error ByteString
+      hasThreeParts token = case length $ BS.split (BS.c2w '.') token of
+        3 -> Right token
+        n -> Left $ JwtErr $ JwtDecodeError ("Expected 3 parts in JWT; got " <> show n)
+      hasMin32Chars :: ByteString -> Either Error ByteString
+      hasMin32Chars token
+          | BS.length token >= 32 = Right token
+          | otherwise = Left $ JwtErr $ JwtDecodeError "JWT must be at least 32 characters"
```
steve-chavez, line 77: Wasn't the `jwt-secret` the one that needed to be at least 32 chars? Not the JWT that's sent per request?

taimoorzaeem: Hmm, now that I think of it, it is unnecessary. I just thought it would be a little nice to give this error, I'll remove it.

test/spec/Feature/Auth/JwtCacheSpec.hs
```diff
@@ -0,0 +1,22 @@
...
+  it "denies access to tables that anonymous does not own" $
+    get "/authors_only" `shouldRespondWith`
+      [json| {
+        "hint":null,
+        "details":null,
+        "code":"42501",
+        "message":"permission denied for table authors_only"} |]
+      { matchStatus = 401
+      , matchHeaders = ["WWW-Authenticate" <:> "Bearer"]
+      }
```
steve-chavez, line 22: PR looks great overall :+1: 

However this test puzzles me, why is it needed here? Isn't this the same as:

https://github.com/PostgREST/postgrest/blob/4348cb205738a8497fadcaed4ab6015aa465b87d/test/spec/Feature/Auth/AuthSpec.hs#L17-L26

taimoorzaeem: Yes, it's the same but written for code-coverage with `jwt cache: enabled` and `server-timing: enabled|disabled`. I should better mention this in the comments.

steve-chavez: And the rest of the tests related to the JWT cache are io tests in python right? Maybe mention that in the comments. Also, for consistency why not make this an io test? Otherwise you have two files to look at for jwt cache tests.

taimoorzaeem: > Also, for consistency why not make this an io test? Otherwise you have two files to look at for jwt cache tests.

Agreed. I was thinking the same.

steve-chavez (approved):

PR #3955, opened 2025-03-18
feat: add Proxy-Status header for better error response

### LINKED ISSUE #2967: `Proxy-Status` header for better error messages
## Problem

HEAD requests hide the error and sometimes the status is not descriptive enough.

For example, doing the following on a table with a high count (say 30_000_000) plus having a short `statement_timeout` (say 3s):

```bash
$ curl 'localhost:3000/table' -I -H "Prefer: count=exact"

Date: Tue, 26 Sep 2023 21:24:38 GMT
Server: postgrest/11.1.0 (1f13e43)
```

Doing a normal GET reveals the error:

```
$ curl 'localhost:3000/table' -H "Prefer: count=exact"

{"code":"57014","details":null,"hint":null,"message":"canceling statement due to statement timeout"}
```

## Proposal

Use the [Proxy-Status header](https://www.rfc-editor.org/rfc/rfc9209.html#name-the-proxy-status-http-field) to communicate the error clearly, even on HEAD requests. 

> Typically, HTTP intermediaries forward requests towards the origin server (inbound) and then forward their responses back to clients (outbound). However, if an error occurs before a response is obtained from an inbound server, the response is often generated by the intermediary itself.
> HTTP accommodates these types of errors with a few status codes -- for example, 502 (Bad Gateway) and 504 (Gateway Timeout). However, experience has shown that more information is necessary to aid debugging and communicate what's happened to the client. 

So for the above case:

```
Proxy-Status: PostgREST; error=statement_timeout
```

## Notes

- Not sure if I'm reading the RFC right and we can just add any error type (we might need to register it first). If not possible, we could use error=[connection_read_timeout](https://www.rfc-editor.org/rfc/rfc9209.html#name-connection-read-timeout).
- According to the RFC, the header can include details too. So this might be another option:

  ```
  Proxy-Status: proxy.example.net; error="http_protocol_error";
                details="Malformed response header: space before colon"
  ```
  https://www.rfc-editor.org/rfc/rfc9209.html#section-2.1.5

taimoorzaeem:
With this feature, PostgREST responds with error code in the `Proxy-Status` header.

Closes #2967.

### DIFF
--- a/docs/references/errors.rst
+++ b/docs/references/errors.rst
@@ -5,6 +5,8 @@ Errors
 
 PostgREST error messages follow the PostgreSQL error structure. It includes ``MESSAGE``, ``DETAIL``, ``HINT``, ``ERRCODE`` and will add an HTTP status code to the response.
 
+.. _postgresql_errors:
+
 Errors from PostgreSQL
 ======================
 
@@ -348,6 +350,8 @@ Internal errors. If you encounter any of these, you may have stumbled on a Postg
 | PGRSTX00      |             |                                                             |
 +---------------+-------------+-------------------------------------------------------------+
 
+.. _custom_errors:
+
 Custom Errors
 =============
 
@@ -444,3 +448,24 @@ For non standard HTTP status, you can optionally add ``status_text`` to describe
   detail = '{"status":419,"status_text":"Page Expired","headers":{"X-Powered-By":"Nerd Rage"}}';
 
 If PostgREST can't parse the JSON objects ``message`` and ``detail``, it will throw a ``PGRST121`` error. See :ref:`Errors from PostgREST<pgrst1**>`.
+
+.. _proxy-status_header:
+
+Proxy-Status Header
+===================
+
+For error cases, `Proxy-Status <https://www.rfc-editor.org/rfc/rfc9209.html#name-the-proxy-status-http-field>`_ header is returned with the error code. The error code comes from either :ref:`PostgREST <pgrst_errors>`, :ref:`PostgreSQL <postgresql_errors>` or :ref:`Custom <custom_errors>` errors. This is useful when doing ``HEAD`` requests where HTTP status is not descriptive enough.
+
+For example, doing a request on a table with high count (say 30_000_000), we get:
+
+.. code-block:: http
+
+  HEAD /table HTTP/1.1
+  Prefer: count=exact
+
+.. code-block:: http
+
+  HTTP/1.1 500 Internal Server Error
+  Proxy-Status: PostgREST; error=57014
+
+The error code ``57014`` reveals that the error is due to a short ``statement_timeout`` value.
--- a/docs/references/errors.rst
+++ b/docs/references/errors.rst
@@ -454,7 +454,7 @@ If PostgREST can't parse the JSON objects ``message`` and ``detail``, it will th
 Proxy-Status Header
 ===================
 
-For error cases, `Proxy-Status <https://www.rfc-editor.org/rfc/rfc9209.html#name-the-proxy-status-http-field>`_ header is returned with the error code. The error code comes from either :ref:`PostgREST <pgrst_errors>`, :ref:`PostgreSQL <postgresql_errors>` or :ref:`Custom <custom_errors>` errors. This is useful when doing ``HEAD`` requests where HTTP status is not descriptive enough.
+For error cases, the standard `Proxy-Status <https://www.rfc-editor.org/rfc/rfc9209.html#name-the-proxy-status-http-field>`_ header is returned with the error code. The error code comes from either :ref:`PostgREST <pgrst_errors>`, :ref:`PostgreSQL <postgresql_errors>` or :ref:`Custom <custom_errors>` errors. This is useful when doing ``HEAD`` requests where the HTTP status is not descriptive enough.
 
 For example, doing a request on a table with high count (say 30_000_000), we get:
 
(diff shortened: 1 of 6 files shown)


src/PostgREST/Error.hs
```diff
@@ -459,17 +460,26 @@ type Authenticated = Bool
 instance PgrstError PgError where
   status (PgError authed usageError) = pgErrorStatus authed usageError
 
-  headers (PgError _ (SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ResultError (SQL.ServerError "PGRST" m d _ _p))))) =
-    case parseRaisePGRST m d of
-      Right (_, r) -> map intoHeader (M.toList $ getHeaders r)
-      Left e       -> headers e
+  headers err@(PgError _ (SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ResultError (SQL.ServerError code m d _ _p))))) =
+    case code of
+      "57014" -> proxyStatusHeader (Just "57014") : statusBasedHeaders (status err)
```
steve-chavez, line 465: Hm, why just this code and not all the codes? The discussion on https://github.com/PostgREST/postgrest/issues/2967#issuecomment-2713224061 only mentioned `57014` as an example.

taimoorzaeem: Oh, okay. So it's for all `pg` errors. Got it.

steve-chavez:
Loadtest looking good, no perf loss https://github.com/PostgREST/postgrest/actions/runs/13937206620?pr=3955

test/io/test_io.py
```diff
@@ -1830,3 +1830,23 @@ def test_log_pool_req_observation(level, defaultenv):
         else:
             output = postgrest.read_stdout(nlines=4)
             assert len(output) == 0
+
+
+def test_proxy_status_header(defaultenv, metapostgrest):
+    "Test Proxy-Status header in statement timeout error"
```
steve-chavez, line 1836: Are [PostgREST error codes](https://docs.postgrest.org/en/v12/references/errors.html#postgrest-error-codes) also exposed on the `Proxy-Status`? If so, we'd need some tests for them.

taimoorzaeem: I have currently exposed [PostgreSQL errors](https://docs.postgrest.org/en/v12/references/errors.html#errors-from-postgresql) only. I am not sure about the `pgrst` errors, do we need them to be exposed as well?

steve-chavez: Yes, all error codes should be exposed. Otherwise it would be inconsistent.

taimoorzaeem: Hmm, not all errors have a code which goes against this feature. I guess we would need a refactor first to have a code associated with all errors.

steve-chavez: I thought all of our errors had a code :thinking: . Yes, if you prefer open another issue/PR to clarify that.

steve-chavez:
> So, I think we need a refactor here where we assign codes somewhere else. I am thinking of refactoring in this way:

@taimoorzaeem LGTM, please go ahead.

taimoorzaeem:
@steve-chavez Is returning `Proxy-Status` header on **every** response critically important? If so we would need to test this extensively. Currently I have written some tests but that's not ideal I guess. Any ideas on how I can test this in a better way?

steve-chavez:
> @steve-chavez Is returning Proxy-Status header on every response critically important? If so we would need to test this extensively. 

It should be only on error responses. Given your refactors on https://github.com/PostgREST/postgrest/pull/3987, I think it's not necessary to have so much extra testing since we already prove in code that all error responses will contain a code?

Given that, only some tests would suffice. Say just one proxy-status error for each different group? (not adamant on this if there are some error groups that are hard to repro, like the X00 ones).

steve-chavez:
@taimoorzaeem Great work! :rocket: 

We should document this. Maybe on https://docs.postgrest.org/en/v12/references/observability.html#traces, after trace header.

steve-chavez:
This is awesome. I just played it with it and it even works with [custom errors](https://postgrest.org/en/latest/references/errors.html#custom-errors):

```
# postgrest-with-postgresql-16  -f test/spec/fixtures/load.sql postgrest-run

$ curl localhost:3000/rpc/raise_pt402 -i
..
Proxy-Status: PostgREST; error=PT402

$ curl localhost:3000/rpc/raise_sqlstate_test1 -i
..
Proxy-Status: PostgREST; error=123
```

@taimoorzaeem How about adding tests for these two cases, to ensure they don't break?

---

Also, I've noticed that the docs for Proxy-Status would better fit in the errors page too. It should mention that it will contain error codes from PostgREST, PostgreSQL and Custom Errors. 

We can still keep a reference (link) to it in the Observability page, same place; for more visilibity.

docs/references/errors.rst
```diff
@@ -444,3 +448,24 @@ For non standard HTTP status, you can optionally add ``status_text`` to describe
...
+
+  HEAD /table HTTP/1.1
+  Prefer: count=exact
+
+.. code-block:: http
+
+  HTTP/1.1 500 Internal Server Error
+  Proxy-Status: PostgREST; error=57014
+
+The error code ``57014`` reveals that the error is due to a short ``statement_timeout`` value.
```
steve-chavez, line 471:
```suggestion
The PostgreSQL error code ``57014`` (`ref <https://www.postgresql.org/docs/current/errcodes-appendix.html>`_) reveals that the error is due to a short ``statement_timeout`` value.
```

PR #3958, opened 2025-03-19
nix: Disable building profiled or dynamic libraries by default

### LINKED ISSUE #3881: Nix loadtests and memory tests fail when adding `GHC.DataSize`
Yeah, running the loadtest in CI is not working when dependencies are changed, because it needs to run against the base branch, which doesn't have the dependencies, yet... and then cabal just breaks it somehow.

You can run something like this locally to get the same markdown output:

```
          postgrest-loadtest-against main v12.2.5
          postgrest-loadtest-report
```

(but it's likely that it fails the same way... :D)

_Originally posted in https://github.com/PostgREST/postgrest/issues/3802#issuecomment-2602971402_

wolfgangwalther:
We never need dynamic haskell libraries, because even the dynamic builds
only dynamically link non-haskell dependencies, but always link haskell
dependencies statically.

Profiled libraries are only required when running the memory test, so
explicitly enable them for the profiled package.

This also means, that we don't need to hide the memory test behind a
feature flag anymore. The reason always was assumed to be the big number
of rebuilds required for it. I assume ever since we moved off of
static-haskell-nix and back to nixpkgs-based builds, we have been
building profiled libraries for all our dependencies anway.

As discussed in https://github.com/PostgREST/postgrest/pull/3802#issuecomment-2737746937

Resolves #3881

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -55,23 +55,19 @@ let
     ];
 
   # Dynamic derivation for PostgREST
-  postgrest =
-    pkgs.haskell.packages."${compiler}".callCabal2nix name src { };
+  postgrest = pkgs.lib.pipe (pkgs.haskell.packages."${compiler}".callCabal2nix name src { }) [
+    # To allow ghc-datasize to be used.
+    lib.disableLibraryProfiling
+    # We are never going to use dynamic haskell libraries anyway. "Dynamic" refers to how
+    # non-haskell deps are linked. All haskell dependencies are always statically linked.
+    lib.disableSharedLibraries
+  ];
 
   staticHaskellPackage = import nix/static.nix { inherit compiler name pkgs src; };
 
   # Options passed to cabal in dev tools and tests
   devCabalOptions =
     "-f dev --test-show-detail=direct";
-
-  profiledHaskellPackages =
-    pkgs.haskell.packages."${compiler}".extend (_: super:
-      {
-        mkDerivation =
-          args:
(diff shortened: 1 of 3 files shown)


steve-chavez (approved):

taimoorzaeem:
Since this, whenever I enter `nix-shell` after a code change, it rebuilds every module again which takes like 2 minutes to complete on my system. This message comes up everytime:
```
these 4 derivations will be built:
  /nix/store/rx51cxfqgxblcvd3ay4cvxc3abpxnw7c-postgrest-12.3.drv
  /nix/store/42j7545wc77ybk3m1fidf9vbnqyqbm0p-postgrest-profiled-run.drv
  /nix/store/yk0a24jj2n8mayd8lrxp3qnvv09dh00k-postgrest-test-memory.drv
  /nix/store/wl4fmm3cxlzsprw0wjnp2qjs1k5rsvjy-postgrest-memory.drv
building '/nix/store/rx51cxfqgxblcvd3ay4cvxc3abpxnw7c-postgrest-12.3.drv'...
```
Can we speed this up to how it was before (entering nix-shell without building)?

steve-chavez:
Can confirm the above.

I believe the memory tests build needed to be explicitly enabled with:
https://github.com/PostgREST/postgrest/blob/d284e9aceab3eac3a56c1fad2f21daca4c085227/nix/README.md#L197-L198

Maybe the "--arg memory" conditional is no longer being applied.

PR #3966, opened 2025-03-24
Add basic flake.nix

### LINKED ISSUE #3026: Nix flake for repo
I would find it very convenient to use a flake interface to install postgrest. I have a fork that makes some minimal changes, to start this, in a backwards compatible way. Would a PR be welcome?

Nixpkgs version upgrades are fairly straightforward — change the revision, then run `nix flake lock --update-input nixpkgs`.

I saw a mention of flakes only in #2350, and that was for a slightly different situation.

wolfgangwalther:
With this basic flake, it's possible to consume postgrest (both dynamic or static) as a flake input. There is no `devShell` via flake or anything else, yet - this can be done later.

Resolves #3026 
Supersedes #3105

@dbaynard @develop7 @mkleczek looking forward to your feedback.

### DIFF
--- a/nix/overlays/postgresql-libpq.nix
+++ b/nix/overlays/postgresql-libpq.nix
@@ -1,6 +1,7 @@
 _: super:
 {
-  libpq = super.callPackage ../libpq.nix {
+  # Depending on which nixpkgs version is pinned, libpq might either be available already - or not.
+  libpq = super.libpq or (super.callPackage ../libpq.nix {
     postgresql = super.postgresql_16;
-  };
+  });
 }


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
```
steve-chavez, line 9: :eyes: what's this about? Does it work on vim?

Also why do we need it here? Isn't the CI style check enough?

flake.nix
```diff
@@ -0,0 +1,47 @@
...
+  inputs = {
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";
+    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+  };
+
+  nixConfig = {
+    extra-substituters = "https://postgrest.cachix.org";
+    extra-trusted-public-keys = "postgrest.cachix.org:icgW4R15fz1+LqvhPjt4EnX/r19AaqxiVV+1olwlZtI=";
+  };
```
steve-chavez, line 13: Q: Does this mean that developers won't need to do an explicit `cachix use postgrest` before compiling?

default.nix
```diff
@@ -3,7 +3,13 @@
...
 
 , # Commit of the Nixpkgs repository that we want to use.
-  nixpkgsVersion ? import nix/nixpkgs-version.nix
+  nixpkgsVersion ? let
+    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  in
+  {
+    inherit (lock.nodes.nixpkgs.locked) owner repo rev;
+    tarballHash = lock.nodes.nixpkgs.locked.narHash;
+  }
```
steve-chavez, line 12: Cool, so this means that doing `nix-build --attr postgrestStatic` to get the static bin will keep working as usual.

flake.nix
steve-chavez: > Later, when we do our development environment via flake,

Once we do that, maybe we shouldn't solely rely on `nix flake check` for tests. I've found it really confusing because it depends on committing files to git and when no files are changed, the tests won't run again. I don't know if there are ways to avoid that.

dbaynard: My comment is redundant given how nixConfig works. I'm glad you tagged me! 

You mention the user being asked: does the option require explicit configuration in non-interactive systems?

steve-chavez: > Turning tests into derivations has benefits - because tests are potentially cached, too. 

TIL that tests are turned into derivations with flakes, now it makes sense. Thanks for elaborating :smiley: 

> We can even add direnv support, then you don't need to run nix develop at all 

IMO direnv should be optional too. I tried it before but didn't have a good experience. I recall it failed sometimes and was slow loading. I'd be fine typing `nix develop/nix-shell` (although I avoid this by having preset projects with nvim-qt).

> and the whole "can I stick with my shell?" topic is done, too.

Cool. Great to know that flakes do bring some simplifications after all.

mkleczek:
I am not really sure if adding flake-parts dependency is a good idea here. It does not look like we are making use of it TBH. Would more widely used flake-utils (like here: https://github.com/sbellem/hello-flake) be enough?

steve-chavez:
>  For what we have right now, we don't really need any of flake-parts or flake-utils.

Related: https://determinate.systems/posts/best-practices-for-nix-at-work/#avoid-flake-helper-libraries-if-possible. +1 on not depending on flake-parts.

dbaynard (approved):

PR #3971, opened 2025-03-25
refactor: cleanup error status and headers

taimoorzaeem:
The error status and headers depend on the authentication parameter. This was not explicitly showed and was hidden in the `PgError` type.

Later, it would be easier to implement new error related features with this change.

For #2967, we need to make `code :: a -> ErrorCode` part of the `PgrstError` typeclass to ensure that every error has a code. This refactor will help implement that cleanly.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -104,7 +104,7 @@ postgrest logLevel appState connWorker =
     -- fromJust can be used, because the auth middleware will **always** add
     -- some AuthResult to the vault.
     \req respond -> case fromJust $ Auth.getResult req of
-      Left err -> respond $ Error.errorResponseFor err
+      Left err -> respond $ Error.errorResponseFor False err
       Right authResult -> do
         appConf <- AppState.getConfig appState -- the config must be read again because it can reload
         maybeSchemaCache <- AppState.getSchemaCache appState
@@ -115,7 +115,7 @@ postgrest logLevel appState connWorker =
           eitherResponse =
             runExceptT $ postgrestResponse appState appConf maybeSchemaCache pgVer authResult req
 
-        response <- either Error.errorResponseFor identity <$> eitherResponse
+        response <- either (Error.errorResponseFor (Auth.getRole req /= configDbAnonRole appConf)) identity <$> eitherResponse
         -- Launch the connWorker when the connection is down.  The postgrest
         -- function can respond successfully (with a stale schema cache) before
         -- the connWorker is done.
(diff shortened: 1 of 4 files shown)


laurenceisla:
Got any idea why so many CodeCov errors are popping up?

taimoorzaeem:
> Got any idea why so many CodeCov errors are popping up?

Well, for starters, all lines of form `instance TypeClass Type where` are "never executed" for some reason. `instance PgrstError Error where` is the only exception. Not sure why coverage fails to see this. 

Secondly, the code of coverage of `Error.hs` module is low already because of many unusual errors and statuses that can't be tested. It would've been so cool if `doctests` were also recognized by code coverage but they are not.

laurenceisla:
> Well, for starters, all lines of form `instance TypeClass Type where` are "never executed" for some reason. `instance PgrstError Error where` is the only exception. Not sure why coverage fails to see this.

The `instance PgrstError Error` is partially covered (it was like that before too), but yes, there are instances that weren't covered before too. Would be nice to check later why this is.

> Secondly, the code of coverage of Error.hs module is low already because of many unusual errors and statuses that can't be tested.

Yeah makes sense.

laurenceisla (approved):

steve-chavez:
> The error status and headers depend on the authentication parameter. This was not explicitly showed and was hidden in the PgError type.

I understand there's some hidden assumption but given the result here wrt to LOCs (`+163 −139`), this doesn't look like a "refactor" in the sense that "it improves the codebase without altering behavior". It looks like the codebase is only deteriorating. I see some options to move this forward:

- Can this be done in a better way?
- If it's a dependency of https://github.com/PostgREST/postgrest/pull/3955, maybe it should be done there instead? This isn't that good, as it will make the PR harder to review.
- If there are some errors that will be left without error codes, maybe these can be corrected after https://github.com/PostgREST/postgrest/pull/3955 is merged? Then the benefit would be clearer, right now on this PR it isn't.

taimoorzaeem:
> * Can this be done in a better way?
> 
> * If it's a dependency of [feat: add Proxy-Status header for better error messages #3955](https://github.com/PostgREST/postgrest/pull/3955), maybe it should be done there instead? This isn't that good, as it will make the PR harder to review.
> 
> * If there are some errors that will be left without error codes, maybe these can be corrected after [feat: add Proxy-Status header for better error messages #3955](https://github.com/PostgREST/postgrest/pull/3955) is merged? Then the benefit would be clearer, right now on this PR it isn't.

Hm, thinking more about this, I can now see that this PR kind of "over-engineer" the code instead of refactoring to make it more clear. I agree that this should be done in a better way.

taimoorzaeem:
Closed as no longer needed.

PR #3983, opened 2025-03-28
refactor: add pg error and custom error to ErrorCode

taimoorzaeem:
Our current `ErrorCode` type wasn't giving us a full picture of how many different types of errors we are handling. This PR makes this explicit by including the error code type for all types of errors.

This is a small change but makes our code consistent and removes ambiguity. It also ensures that all errors use `toJsonPgrstError` to create the error json.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -502,20 +502,18 @@ instance JSON.ToJSON SQL.CommandError where
   -- Special error raised with code PGRST, to allow full response control
   toJSON (SQL.ResultError (SQL.ServerError "PGRST" m d _ _p)) =
     case parseRaisePGRST m d of
-      Right (r, _) -> JSON.object [
-        "code"     .= getCode r,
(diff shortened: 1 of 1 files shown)


laurenceisla (approved):
I like it, I always had in my backlog reorganize the error codes and having all under one type is a step to it. I should've done this when creating the `PGRST` errors, so LGTM.

steve-chavez:
Does this PR supersede https://github.com/PostgREST/postgrest/pull/3971?

steve-chavez:
> by including the error code type for all types of errors.

Q: So are there now responses that give error codes and before they weren't? If so, it would be good to have a new test that showcases this.

steve-chavez (approved):
Code looks much better :+1:

Great to see the explicitness of the error code.

taimoorzaeem:
> Does this PR supersede #3971?

No, its different, but I will close it because we might not need it anymore.

PR #3985, opened 2025-03-30
nix: avoid rebuilding memory tests when entering nix-shell

wolfgangwalther:
Attempt to solve https://github.com/PostgREST/postgrest/pull/3958#issuecomment-2750300101.

The memory consumption increases slightly, because the binary is now built slightly differently. Couldn't figure out how, though. Shouldn't be a problem for the test though.

@taimoorzaeem @steve-chavez please test and let me know what you think


taimoorzaeem:
Working good for me! :tada:

steve-chavez (approved):
Works great for me too :tada:

PR #3987, opened 2025-04-01
refactor: create error body with ErrorBody typeclass

taimoorzaeem:
The old error design didn't allow us to reuse error `code`, `message` etc. With this refactor, these parts of the error body can be reused for other potential features. We need this for #3955.

Edit: With this change, proxy-status header would be simply:
```haskell
[("Proxy-Status", "PostgREST; error=" <> code err)]
```

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -25,7 +25,6 @@ import qualified Data.HashMap.Strict       as HM
 import qualified Data.Map.Internal         as M
 import qualified Data.Text                 as T
 import qualified Data.Text.Encoding        as T
-import qualified Data.Text.Encoding.Error  as T
 import qualified Hasql.Pool                as SQL
 import qualified Hasql.Session             as SQL
 import qualified Network.HTTP.Types.Status as HTTP
@@ -50,7 +49,7 @@ import PostgREST.SchemaCache.Table        (Table (..))
 import Protolude
 
 
-class (JSON.ToJSON a) => PgrstError a where
+class (ErrorBody a, JSON.ToJSON a) => PgrstError a where
   status   :: a -> HTTP.Status
   headers  :: a -> [Header]
 
@@ -61,6 +60,12 @@ class (JSON.ToJSON a) => PgrstError a where
   errorResponseFor err =
     let baseHeader = MediaType.toContentType MTApplicationJSON in
     responseLBS (status err) (baseHeader : headers err) $ errorPayload err
+
+class ErrorBody a where
+  code    :: a -> Text
+  message :: a -> Text
+  details :: a -> Maybe JSON.Value
+  hint    :: a -> Maybe JSON.Value
 
 data ApiRequestError
   = AggregatesNotAllowed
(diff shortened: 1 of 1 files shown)


src/PostgREST/Error.hs
```diff
@@ -784,8 +816,8 @@ buildErrorCode code = case code of
   JWTErrorCode02         -> "PGRST302"
   JWTErrorCode03         -> "PGRST303"
 
-  InternalErrorCode00    -> "PGRSTX00"
+  InternalErrorCode00    -> "PGRSTX00" -- TODO: Ponder - We never really return this error, we kill pgrst early in App.hs
```
laurenceisla, line 819: I think killing PostgREST only happens at startup? I don't think we kill it if it's already running and some unknown Hasql error happens. We've already found an error this way: https://github.com/PostgREST/postgrest/issues/3126

src/PostgREST/Error.hs
```diff
@@ -143,166 +148,141 @@ instance PgrstError ApiRequestError where
...
-    ApiRequestErrorCode03
-    "Requested range not satisfiable"
-    (Just $ case rangeError of
+instance ErrorBody ApiRequestError where
+  -- CODE: ErrorCode
+  code QueryParamError{}           = ApiRequestErrorCode00
+  code InvalidRpcMethod{}          = ApiRequestErrorCode01
+  code InvalidBody{}               = ApiRequestErrorCode02
+  code InvalidRange{}              = ApiRequestErrorCode03
+  code InvalidFilters              = ApiRequestErrorCode05
```
laurenceisla, line 157: I like that the error codes can be defined in order (except for the Connection errors, but that's OK) and grouped by their error type. Maybe we could now remove the, frankly ugly, definition of errors at the bottom. I mean all these definitions:

```haskell
data ErrorCode
  -- PostgreSQL connection errors
  = ConnectionErrorCode00
  | ConnectionErrorCode01
  | ConnectionErrorCode02
  | ConnectionErrorCode03
  -- API Request errors
  | ApiRequestErrorCode00
  | ApiRequestErrorCode01
  ...
```

The objective was to easily check which error code was the last one and add a new one accordingly. This can be easily done now that it's grouped by type of error, that is: `ApiRequestError`, `JwtError` and `SchemaCacheError`. Again, the Connection errors are not ordered but it shouldn't be a problem since adding new errors for those is rare.

So, my proposed change is to remove the `data ErrorCode` definition altogether and use the code as `Text` directly. For example, this part of the code would end as:

```suggestion
  code QueryParamError{}           = "PGRST000"
  code InvalidRpcMethod{}          = "PGRST001"
  code InvalidBody{}               = "PGRST002"
  code InvalidRange{}              = "PGRST003"
--code ParseRequestError           = "PGRST004" -- no longer used
  code InvalidFilters              = "PGRST005"
```

It should also show the "no longer used" error codes. WDYT?

laurenceisla: This also has the added benefit of doing a fast "CTRL+F" search for "PGRSTxxx" code, and have the corresponding error directly.

src/PostgREST/Error.hs
taimoorzaeem: Oh, I now understand that it only happens in rare cases indicating a bug, that's why it is not covered by tests.

src/PostgREST/Error.hs
taimoorzaeem: Hmm, I was not sure about this. For once, I like the idea of seeing all errors codes under one error type. But I also agree that the extra indirection looks kind of ugly and we don't need this because we can define them directly in order (except for connection error but those are rare anyway). So this is fine with me :+1:.

src/PostgREST/Error.hs
```diff
@@ -143,166 +148,159 @@ instance PgrstError ApiRequestError where
...
+  code InvalidResourcePath         = "PGRST125"
+  code OpenAPIDisabled             = "PGRST126"
+  code NotImplemented{}            = "PGRST127"
+
+  code NoRelBetween{}              = "PGRST200"
+  code AmbiguousRelBetween{}       = "PGRST201"
+  code NoRpc{}                     = "PGRST202"
+  code AmbiguousRpc{}              = "PGRST203"
+  code ColumnNotFound{}            = "PGRST204"
+  code TableNotFound{}             = "PGRST205"
```
steve-chavez, line 200: This is great work, things are already looking much better :tada: 

Here I was wondering if we should have a type besides `ApiRequestError` for these schema cache errors? That way the group prefixes are clearly separated?

steve-chavez: Ok, cool. Let's leave that for later.

PR #3993, opened 2025-04-02
docs: updates support specifying columns and missing preference

### LINKED ISSUE #3522: Documentation does not mention that Updates can use the `columns` parameter or `missing=default` preference
Noticed in #3521 that the Update subsection does not mention that [Specifying Columns](https://postgrest.org/en/latest/references/api/tables_views.html#specifying-columns) and [`Prefer: missing=default`](https://postgrest.org/en/v12/references/api/tables_views.html#bulk-insert-with-default-values) can be used.

taimoorzaeem:
Closes #3522.


docs/references/api/tables_views.rst
```diff
@@ -539,45 +539,6 @@ To bulk insert JSON post an array of objects having all-matching keys
     ]
   EOF
 
-.. _bulk_insert_default:
```
steve-chavez, line 542: @taimoorzaeem Note that when we break an anchor link, we add a redirect to the new one with javascript.

For an example of how it works, try https://postgrest.org/en/latest/references/api/tables_views.html#bulk-insert-default.

Just added the redirect on https://github.com/PostgREST/postgrest/commit/d89e7a21735c0a72297cc1467c5f281663539fb3. Forgot to mention it in this PR.

taimoorzaeem: Alright, gotcha.

PR #3999, opened 2025-04-08
docs: add note that ordering of columns is not enforced

### LINKED ISSUE #3962: Ordering of columns using related resources doesn't match query
### Environment

* PostgreSQL version: postgres:17.2
* PostgREST version: postgrest/postgrest:v12.2.3
* Operating system: Docker

### Description of issue

When making a request to PostgREST with a related resource, I'd expect the order of the columns in the response to maintain the same order as the request. For example, if I made the following request:
```
http://localhost:3000/films?select=directors(last_name,id),title
```
I would expect a response of the format:
```json
[
  {
    "directors": {
      "last_name": "name",
      "id": 5
    },
    "title": "title"
  }
]
```
Instead, I get a response that looks like this:
```json
[
  {
    "title": "title",
    "directors": {
      "id": 5,
      "last_name": "name"
    }
  }
]
```
Note that the title is first, even though it's last in the select statement, and the ordering of the fields within `directors` doesn't match what was specified.

If spread the embedded resource, I get the following:
```json
[
  {
    "title": "title",
    "last_name": "name",
    "id": 5
  }
]
```
The ordering of the related fields is correct, but the title is still first.

Ideally, it would be nice for the ordering to match what is put in the select statement, since that's what I'd expect using SQL.

taimoorzaeem:
Closes #3962.

### DIFF
--- a/docs/references/api/resource_representation.rst
+++ b/docs/references/api/resource_representation.rst
@@ -16,6 +16,32 @@ Use the Accept request header to specify the acceptable format (or formats) for 
   curl "http://localhost:3000/people" \
     -H "Accept: application/json"
 
+.. note::
+  
+  The ordering of columns in the response isn't guaranteed to align with the order specified in the ``select`` clause. For example, with resource embedding:
+
+  .. code-block:: bash
+    
+    http://localhost:3000/films?select=directors(last_name,id),title
+  
+  We may get:
+
+  .. code-block:: bash
+
+    [
+      {
+        "title": "title",
+        "directors": {
+          "id": 5,
+          "last_name": "name"
+        }
+      }
+    ]
+
+  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core>`_:
+
+    *JSON Schema is defined over "application/json" or compatible documents, including media types with the "+json" structured syntax suffix.*
+
 .. _builtin_media:
 
 Builtin Media Type Handlers
--- a/docs/references/api/resource_representation.rst
+++ b/docs/references/api/resource_representation.rst
@@ -38,9 +38,9 @@ Use the Accept request header to specify the acceptable format (or formats) for 
       }
     ]
 
-  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core>`_:
+  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model>`_:
 
-    *JSON Schema is defined over "application/json" or compatible documents, including media types with the "+json" structured syntax suffix.*
+    object: An **unordered** set of properties mapping a string to an instance, from the JSON "object" value
 
 .. _builtin_media:
 
--- a/docs/references/api/resource_representation.rst
+++ b/docs/references/api/resource_representation.rst
@@ -40,7 +40,7 @@ Use the Accept request header to specify the acceptable format (or formats) for 
 
   This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model>`_:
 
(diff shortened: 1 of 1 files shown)


docs/references/api/resource_representation.rst
```diff
@@ -16,6 +16,30 @@ Use the Accept request header to specify the acceptable format (or formats) for
...
+      {
+        "title": "title",
+        "directors": {
+          "id": 5,
+          "last_name": "name"
+        }
+      }
+    ]
+
+  See: `github comment <https://github.com/PostgREST/postgrest/issues/3962#issuecomment-2748441060>`_ for why the ordering is not enforced.
```
steve-chavez, line 41: It would be better to inline this explanation:

```suggestion
This is in line with the [JSON schema spec](https://json-schema.org/draft/2020-12/json-schema-core):

> JSON Schema is defined over "application/json" or compatible documents, including media types with the "+json" structured syntax suffix.
```

Note: this is not valid rst, needs to be edited.

docs/references/api/resource_representation.rst
```diff
@@ -16,6 +16,32 @@ Use the Accept request header to specify the acceptable format (or formats) for
...
+          "id": 5,
+          "last_name": "name"
+        }
+      }
+    ]
+
+  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core>`_:
+
+    *JSON Schema is defined over "application/json" or compatible documents, including media types with the "+json" structured syntax suffix.*
+
```
steve-chavez, line 44: Argh, I suggested the wrong section here. Should be:
```suggestion
  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model>`_:

    object: An **unordered** set of properties mapping a string to an instance, from the JSON "object" value

```

docs/references/api/resource_representation.rst
```diff
@@ -16,6 +16,32 @@ Use the Accept request header to specify the acceptable format (or formats) for
...
+        "directors": {
+          "id": 5,
+          "last_name": "name"
+        }
+      }
+    ]
+
+  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model>`_:
+
+    object: An **unordered** set of properties mapping a string to an instance, from the JSON "object" value
```
steve-chavez, line 43:
```suggestion
    *object: An **unordered** set of properties mapping a string to an instance, from the JSON "object" value*
```

docs/references/api/resource_representation.rst
```diff
@@ -16,6 +16,32 @@ Use the Accept request header to specify the acceptable format (or formats) for
...
+        "directors": {
+          "id": 5,
+          "last_name": "name"
+        }
+      }
+    ]
+
+  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model>`_:
+
+    *"object: An **unordered** set of properties mapping a string to an instance, from the JSON "object" value"*
```
steve-chavez, line 43:
```suggestion
    *"object: An unordered set of properties mapping a string to an instance, from the JSON "object" value"*
```

docs/references/api/resource_representation.rst
```diff
@@ -16,6 +16,32 @@ Use the Accept request header to specify the acceptable format (or formats) for
...
+        "directors": {
+          "id": 5,
+          "last_name": "name"
+        }
+      }
+    ]
+
+  This is in line with the `JSON schema spec <https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model>`_:
+
+    *"object: An unordered set of properties mapping a string to an instance, from the JSON "object" value"*
```
steve-chavez, line 43:
```suggestion
    *"object: An unordered set of properties mapping a string to an instance"*
```

Issue #3164, opened 2024-01-16
`Prefer:resolution/missing` options should be inside a `Content-Type`


steve-chavez:
## Problem

Currently there are disparate prefers headers to do a [bulk insert](https://postgrest.org/en/stable/references/api/tables_views.html#bulk-insert-default) and [upsert](https://postgrest.org/en/stable/references/api/tables_views.html#prefer-resolution). Ideally this would be done through a single content type so it later can be overridden with https://github.com/PostgREST/postgrest/issues/2826.

A similar idea was previously discussed on https://github.com/PostgREST/postgrest/issues/2066#issue-1070332545

## Solution

Use a vendor media type that contains both options:

```http
POST /employees
Content-Type: application/vnd.pgrst.bulk+json;resolution=merge-duplicates;missing=default
```

---

Later on we can include the `on_conflict`  and `columns` query parameters as parameters of the same `content-type`.

PR #4005, opened 2025-04-10
fix: Fix parsing of the `for` parameter of plan media type

### LINKED ISSUE #3498: Proper Media Types parser
### Description of issue

Current parser breaks on media types like `application/vnd.pgrst.plan;wat="application/json;text/csv"` producing nonsense like `("application","vnd.pgrst.plan",[("wat","application/json"),("text/csv\"","")])` and doesn't handle escaped quotes well too.

We do use parsec in `QueryParams` BTW, so it's an obvious candidate to use here as well.

taimoorzaeem:
The media type parser had some bugs. It didn't allow space separated parameters and the handling of quoted text was clunky. This commit improves the media type parser to fix these bugs.

Closes #3498 without too many changes. More refactoring is needed but that's a task for the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3697, #3602, Handle queries on non-existing table gracefully - @taimoorzaeem
  - #3600, #3926, Improve JWT errors - @taimoorzaeem
  - #3013, Fix `order=` with POST, PATCH, PUT and DELETE requests - @taimoorzaeem
+ - #3498, Fix incorrect parsing of the `for` parameter of the `application/vnd.pgrst.plan` media type - @taimoorzaeem
 
 ### Changed
 
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE DeriveAnyClass        #-}
 {-# LANGUAGE DeriveGeneric         #-}
 {-# LANGUAGE DuplicateRecordFields #-}
-
+{-# OPTIONS_GHC -Wno-unused-do-bind #-}
 module PostgREST.MediaType
   ( MediaType(..)
   , MTVndPlanOption (..)
@@ -11,17 +11,16 @@ module PostgREST.MediaType
   , decodeMediaType
   ) where
 
-import qualified Data.Aeson      as JSON
-import qualified Data.ByteString as BS
+import qualified Data.Aeson                    as JSON
+import qualified Data.ByteString               as BS
+import qualified Data.Text                     as T
+import qualified Text.ParserCombinators.Parsec as P
 
+import Data.Map                  (fromList, (!?))
+import Data.Text.Encoding        (decodeLatin1)
 import Network.HTTP.Types.Header (Header, hContentType)
 
-import           Data.Map           (fromList, (!?))
-import qualified Data.Text          as T (break, drop, dropWhile,
-                                          dropWhileEnd, null, splitOn,
-                                          toLower)
-import           Data.Text.Encoding (decodeLatin1)
-import           Protolude
+import Protolude
 
 -- | Enumeration of currently supported media types
 data MediaType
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3697, #3602, Handle queries on non-existing table gracefully - @taimoorzaeem
  - #3600, #3926, Improve JWT errors - @taimoorzaeem
  - #3013, Fix `order=` with POST, PATCH, PUT and DELETE requests - @taimoorzaeem
+ - #3498, Improve media type parser - @taimoorzaeem
+   + Allows space separated parameters
+   + Handles quoted text gracefully
```
steve-chavez, line 39: How about some tests for these fixes?

taimoorzaeem: So, we don't currently have this issue with `Accept` header because we use `parseHttpAccept` library function. But this issue is only for `Content-Type` header. I can't think of a good test `spec` with `Content-Type` as we currently don't use parameters for them (we ignore them). I have added the `doctests` for these fixes.

Later with #3164 this behavior will be tested with `Content-Type` header because then, we will use the parameters. So I think I should specify in the CHANGELOG that media parsing is improved for `Content-Type` header?

src/PostgREST/MediaType.hs
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE DeriveAnyClass        #-}
 {-# LANGUAGE DeriveGeneric         #-}
 {-# LANGUAGE DuplicateRecordFields #-}
-
+{-# OPTIONS_GHC -Wno-unused-do-bind #-}
```
steve-chavez, line 4: First time seeing this one, why is it needed?

taimoorzaeem: GHC just raises a warning when:

```haskell
P.char '/' -- unused do bind
```

GHC suggest to alternatively use underscore:

```haskell
_ <- P.char '/'
```
This looks kind of ugly. So this option removes that warning. We also use this in:

https://github.com/PostgREST/postgrest/blob/6b4648d2e4ea37e9ebea2b9ab0ee6acaa5a09d4f/src/PostgREST/Config/JSPath.hs#L1

CHANGELOG.md
taimoorzaeem: > I still don't get is how is this a user-facing fix

Thinking more about it, I guess this isn't user facing at all **yet**. But it's benefit will be seen in #3164. I'll change this to a `refactor: ...` instead.

> Perhaps you can share some fixed cases here to help me understand?

Sure. So, previously we had this which breaks when their is a semicolon in quotes like `\" ; \"`:
```haskell
>>> tokenizeMediaType "application/vnd.pgrst.plan;wat=\"application/json;text/csv\""
("application","vnd.pgrst.plan",[("wat","application/json"),("text/csv\"","")])
```

This is fixed now with:

```haskell
>>> P.parse tokenizeMediaType "" "application/vnd.pgrst.plan;wat=\"application/json;text/csv\""
Right ("application","vnd.pgrst.plan",[("wat","application/json;text/csv")])
```

steve-chavez: Oh, I get it now. Given that the plan media type has the ability to include another media type inside:

> Accept: application/vnd.pgrst.plan; for="text/xml"
> From https://docs.postgrest.org/en/v12/references/observability.html#execution-plan

I think then this could be labeled as a fix for the plan media type. A `fix:` prefix is correct.

On the CHANGELOG it could be like:

```
Fix incorrect parsing of the `for` parameter on the ` application/vnd.pgrst.plan` media type.
```

src/PostgREST/MediaType.hs
```diff
@@ -150,7 +152,10 @@ decodeMediaType mt = decodeMediaType' $ decodeLatin1 mt
         ("*","*",_)                                 -> MTAny
         _                                           -> MTOther mt'
       where
-        (mainType, subType, params') = tokenizeMediaType mt'
+        mediaTypeOrError = P.parse tokenizeMediaType "parsec: tokenizeMediaType failed" $ T.unpack mt'
+        (mainType, subType, params') = case mediaTypeOrError of
+          Right mt'' -> mt''
+          Left _     -> ("*", "*", []) -- TODO: Throw mediatype error, would need refactoring because currently Error module depend on MediaType module
```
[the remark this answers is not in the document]
taimoorzaeem: Yes, `("*", "*", [])` is actually wrong here. If parsing fails then we should just return that media type as is, so I think the correct expression here should be:
```haskell
Left _ -> (mt', "", [])
```

PR #4008, opened 2025-04-12
feat: change JWT cache to limited LRU based cache

### LINKED ISSUE #4003: LRU based cache for JWT
## Problem

Our current JWT cache implementation has no upper bound for number of cache entries. This can cause OOM errors. Additionally, the purge mechanism for expired entries is quite slow.

## Solution

Switch to [lrucache](https://hackage.haskell.org/package/lrucache) implementation for limited number of cache entries. We'll need a new config option (say `jwt-max-cache-entries`) for this to work.

taimoorzaeem:
BREAKING CHANGE

Closes #4003.

- [x] Implementation
- [x] Tests
- [x] Docs


taimoorzaeem:
- How should the two configs `jwt-cache-max-lifetime` and `jwt-cache-max-entries` behave relative to each other?
- What should be the default value for `jwt-cache-max-entries` config?
- How should we handle the case where the config `jwt-cache-max-entries` is set to 0? The library throws an **exception**:

```haskell
-- | Make an LRU.  If a size limit is specified, the LRU is guaranteed
-- to not grow above the specified number of entries.
newLRU :: (Ord key) => Maybe Integer -- ^ the optional maximum size of the LRU
       -> LRU key val
newLRU (Just s)  | s <= 0 = error "non-positive size LRU"
newLRU s = LRU Nothing Nothing Map.empty
```

Should we fail at startup?

taimoorzaeem:
> What exactly is the breaking part here? Just the different config options?

I think the breaking part is not the interface here (because we didn't remove the old config) but the part where we changed the implementation. Is that considered breaking?

taimoorzaeem:
> Not changing the interface, but only the implementation is pretty much the definition of "not breaking", I think :D

Thanks :+1:, then I guess this PR **could** be breaking **if** we decide to remove the config `jwt-cache-max-lifetime` which is unlikely, but still not sure. If we are only adding a new config then I guess this is just another feature and hence "not breaking".

steve-chavez:
> jwt-cache-max-lifetime and jwt-cache-max-entries

Does it make sense to have `jwt-cache-max-lifetime` now? Since the main motivation of that config was to keep memory usage at bay. I believe `jwt-cache-max-entries` should be enough now, then we could remove the other config.

> What should be the default value for jwt-cache-max-entries config?

I think it should start at 1000, I don't have the reference handy but I commented before an average JWT size.

> How should we handle the case where the config jwt-cache-max-entries is set to 0? The library throws an exception:
> Should we fail at startup?

I believe that 0 should mean the JWT cache is disabled.

taimoorzaeem:
How should we handle this when `config-jwt-secret` is changed in a reload? I think we should invalidate all cache entries because they were cached with the previous secret.

taimoorzaeem:
> How should we handle this when `config-jwt-secret` is changed in a reload? I think we should invalidate all cache entries because they were cached with the previous secret.

We are currently not doing this in v12. Does that mean we have a bug in v12 that should be fixed? Not sure if this is a severe issue from a security POV.

steve-chavez:
@taimoorzaeem Please rebase :pray: , I'll review then.

taimoorzaeem:
Waiting for #4023 to be fixed.

src/PostgREST/Auth/JwtCache.hs
```diff
@@ -4,96 +4,99 @@ Description : PostgREST Jwt Authentication Result Cache.
...
+isExpClaimExpired :: AuthResult -> UTCTime -> Expired
+isExpClaimExpired result utc =
+    case expireJSON of
+      Nothing                      -> False -- if exp not present then it is valid
+      Just (JSON.Number expiredAt) -> (sciToInt expiredAt - now) < 0
+      Just _                       -> False -- if exp is not a number then valid
+      where
+        expireJSON = KM.lookup "exp" (authClaims result)
+        now = (floor . nominalDiffTimeToSeconds . utcTimeToPOSIXSeconds) utc :: Int
+        sciToInt = fromMaybe 0 . Sci.toBoundedInteger
```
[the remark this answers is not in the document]
steve-chavez: We might want to do https://github.com/PostgREST/postgrest/issues/4035 to allow better testing of the clock skew on the cache.

src/PostgREST/Auth/JwtCache.hs
```diff
@@ -4,96 +4,99 @@ Description : PostgREST Jwt Authentication Result Cache.
...
+
+            if isExpClaimExpired result utc then do
+                -- delete token -> update cache -> parse token
+                let (jwtCache'',_) = C.delete token jwtCache'
+                I.writeIORef jwtCacheIORef jwtCache''
+                parseJwt
+            else do
+                -- update cache afte lookup -> return result
+                I.writeIORef jwtCacheIORef jwtCache'
+                return $ Right result
```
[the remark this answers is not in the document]
taimoorzaeem: When we lookup, the state of cache changes (cache hit updates that entry because its LRU cache now that entry is used "recently"). So, we need to write the new cache state.

taimoorzaeem:
> > How should we handle this when `config-jwt-secret` is changed in a reload? I think we should invalidate all cache entries because they were cached with the previous secret.
> 
> Could we have a similar case with the `jwt-aud` setting?
> 
> Also, since we're caching `AuthResult`, I guess we have the same thing with `jwt-role-claim-key`, too. If that changes, the `AuthResult` for a given JWT must change as well - so we need to invalidate the cache.
> 
> (those comments are probably more about the existing state, than this PR specifically)

@wolfgangwalther Yes, this is important. We should discuss config reloading in more detail. Could you open an issue with this?

src/PostgREST/Auth/JwtCache.hs
```diff
@@ -4,96 +4,100 @@ Description : PostgREST Jwt Authentication Result Cache.
...
+            if isExpClaimExpired result utc then do
+                -- delete token -> update cache -> parse token
+                let (jwtCache'',_) = C.delete token jwtCache'
+                I.writeIORef jwtCacheIORef jwtCache''
+                parseJwt
+            else do
+                -- update cache after lookup -> return result
+                I.writeIORef jwtCacheIORef jwtCache'
+                return $ Right result
+
```
mkleczek, line 89: The above code is not thread safe - it is _read-modify-write_ and causes lost updates.
Granted - in case of cache lost updates might not be that much of a deal but maybe TRef instead of IORef would be better here.

src/PostgREST/AppState.hs
```diff
@@ -105,8 +104,8 @@ data AppState = AppState
   , stateSocketAdmin       :: Maybe NS.Socket
   -- | Observation handler
   , stateObserver          :: ObservationHandler
-  -- | JWT Cache
-  , stateJwtCache          :: JwtCache.JwtCacheState
+  -- | JWT Cache, disabled when config jwt-cache-max-entries is set to 0
+  , stateJwtCache          :: IORef JwtCache.JwtCacheState
```
mkleczek, line 108: I am not sure if IORef is needed here - we already have a mutable variable in JwtCacheState? Why do we need another one here?
A cleaner solution would be to export _reset :: JwtCacheState -> IO ()_ function from JwtCache module and call it upon config reload.

src/PostgREST/AppState.hs
```diff
@@ -120,14 +119,14 @@ data SchemaCacheStatus
 type AppSockets = (NS.Socket, Maybe NS.Socket)
 
 init :: AppConfig -> IO AppState
-init conf@AppConfig{configLogLevel, configDbPoolSize} = do
+init conf = do
   loggerState  <- Logger.init
-  metricsState <- Metrics.init configDbPoolSize
-  let observer = liftA2 (>>) (Logger.observationLogger loggerState configLogLevel) (Metrics.observationMetrics metricsState)
+  metricsState <- Metrics.init (configDbPoolSize conf)
```
mkleczek, line 124: Does this change have anything to do with JWT cache?

src/PostgREST/Auth/JwtCache.hs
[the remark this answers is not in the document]
mkleczek: See my comments about lack of thread safety here and possible lost updates - which means we do not have up to date information about cache usage and the whole idea of LRU cache does not make too much sense any more.

OTOH updating a shared variable upon each cache hit is going to have a penalty because of contention under load.

To be honest - looking at this I am not convinced LRU cache is a good idea - at least with current immutable LRU cache implementation.

Not really sure what to do with this but Nikita Volkov implemented https://hackage.haskell.org/package/stm-containers which should provide a map implementation that minimizes contention.

Or we could stick with current implementation but use a better cache implementation: https://hackage.haskell.org/package/psqueues comes to mind as it should provide efficient purging.

src/PostgREST/Auth/JwtCache.hs
taimoorzaeem: I think we should run the entire `lookupJwtCache` function atomically. This would make it more thread safe.

mkleczek: If lookupJwtCache is atomic then you introduce very high contention as **all** threads wait for each other to just perform lookup. The problem is even worse because with LRU cache lookups are not read-only anymore, so the shared variable has to be updated upon **every access**.

I don't think there is a good solution to this without a proper **mutable** concurrent high performance LRU cache implementation.

@steve-chavez @wolfgangwalther - what do you think?

taimoorzaeem:
Concurrency and atomic operations are important for correctness and consistency but we also need a mutable reference for `LRU` cache type. So I think we would need to use both `TVar` and `IORef` for transactional operations and mutable state respectively. Considering this, I think this would need quite a bit of refactoring.

src/PostgREST/Auth/JwtCache.hs
taimoorzaeem: > If lookupJwtCache is atomic then you introduce very high contention as **all** threads wait for each other to just perform lookup. The problem is even worse because with LRU cache lookups are not read-only anymore, so the shared variable has to be updated upon **every access**.

With many threads, performance for any shared resource will take a hit no matter what because of mutual exclusion. 

> I don't think there is a good solution to this without a proper **mutable** concurrent high performance LRU cache implementation.

So is that even possible to have "high" performance with concurrency when waiting will be involved? I think the only way to maintain performance here would be to decrease waiting which means we need to use a distributed the cache which will get complex. :thinking:

mkleczek: @wolfgangwalther yes, that's exactly the case. In case of LRU cache it is amplified because write is performed upon **every access** (contrary to current solution which only requires a lock upon cache miss, which is rare).

@taimoorzaeem we don't want to have global lock here.

taimoorzaeem: For any cache with LRU eviction policy, the state of the cache will change on a lookup.

> However the suggested alternative only locks part of the cache when writing, thus allowing much better concurrency.

Could you explain what is the suggested alternative here? I am a little lost on this thread.

mkleczek: No really. I could only find one low contention Map implementation: https://hackage.haskell.org/package/stm-containers.

But it is just a simple key-value mapping.

https://hackage.haskell.org/package/psqueues OTOH is a pure data structure meaning it also requires a global lock. But it could be an alternative to current cache implementation as the one we have today has O(n^2) complexity of purge operation.

taimoorzaeem: I think `stm-containers` can be used as a simple cache + a simple frequency map to track frequency/least-recent entries. It shouldn't be too hard.

mkleczek: > Ah, ok, then I misunderstood as well.
> 
> Ok, so with the current implementation (global lock) we should still compare the performance of lrucache and psqueues.
> 

It is still more complex :)

Current implementation is based on Data.Cache + global lock. Replacing Data.Cache with PSQueues would supposedly have the benefit of faster cache purging (but global lock would stay).

Replacing Data.Cache with lrucache while **leaving global lock** is problematic because lrucache requires taking a lock upon **each access** (contrary to both current and psqueues based implementation).

I am not too sure about roll-our-own cache implementation to be honest. It is not trivial to do it correctly. I would also suggest to implement it as a separate library outside of PostgREST.

taimoorzaeem: >  I would also suggest to implement it as a separate library outside of PostgREST. 

Maintaining a separate library would have been possible if we only used hackage+cabal, but we don't. It takes months for a new version of a library to be available on Stack and NixOS. I think our own implementation in postgrest would be our best bet.

taimoorzaeem: > It does not support expiration though. It would be good to add it otherwise the cache is vulnerable to cache thrashing attacks by continuously sending requests with expired JWTs.

That is not an issue at all. We don't need a built-in expiration support. We can just check the `exp` claim on lookup. It would just be an `O(1)` task.

mkleczek: > This implementation is pure as well, so we'd still need to write the new cache on a lookup 😢.
> 

It is not - see this link: https://hackage.haskell.org/package/lrucaching-0.3.4/docs/Data-LruCache-IO.html#v:stripedCached

taimoorzaeem: The `stripedCached` does not have any interface for insert,update,delete etc. Internally, it is still `LruCache`.

taimoorzaeem:
Because the already existing libraries don't satisfy our requirements yet, I guess our best bet is to come up with our own implementation of cache. It might take a while and it would probably delay the v13 release. @steve-chavez WDYT?

taimoorzaeem:
Closing as superseded by #4084.

PR #4009, opened 2025-04-12
v12:  fix: Fix parsing of the for parameter of plan media type

wolfgangwalther:
#4005 applies cleanly on v12.

I have not looked into it at all, should we backport it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


steve-chavez (approved):
Looks good to me to backport

steve-chavez:
> Any idea why the memory test fails repeatedly, i.e. why this causes more memory consumption?

Hm, no. But it has increased very slightly, it should be fine to increase the upper bounds IMO.

steve-chavez:
Maybe Parsec consumes more memory, seems reasonable (because the backtracking and all).

steve-chavez (approved):

Issue #2563, opened 2022-11-08
Related `and/or` conditions


steve-chavez:
Do you mean that if `order=clients.name,name,clients.rank` were used, then we should also consider to support `or=(clients.name.like.Chris*,and(name.eq.TopProject,clients.rank.gt.1)`?

It is indeed hard to read, and `or=(clients(name).like.Chris*,and(name.eq.TopProject,clients(rank).gt.1)` seems slightly better.

_Originally posted by @Iced-Sun in https://github.com/PostgREST/postgrest/issues/1414#issuecomment-986187388_
      
Also previously discussed on https://github.com/PostgREST/postgrest/discussions/2014

steve-chavez:
Having the `client`, `clientinfo` and `contact` tables:
```
$ postgrest-with-postgresql-15 psql

postgres=# table test.clientinfo;
 id | clientid |      other
----+----------+------------------
  1 |        1 | 123 Main St
  2 |        2 | 456 South 3rd St
  3 |        3 | 789 Palm Tree Ln
(3 rows)

postgres=# table test.clientinfo;
 id | clientid |      other
----+----------+------------------
  1 |        1 | 123 Main St
  2 |        2 | 456 South 3rd St
  3 |        3 | 789 Palm Tree Ln
(3 rows)

postgres=# table test.contact;
 id |     name      | clientid
----+---------------+----------
  1 | Wally Walton  |        1
  2 | Wilma Wellers |        1
  3 | Tabby Targo   |        2
  4 | Bobby Bots    |        3
  5 | Bonnie Bits   |        3
  6 | Billy Boats   |        3
(6 rows)

```

I'd like to get the `client` rows where `clientinfo.name LIKE %Tabby% OR contact.other LIKE %Main%`.

Keeping the shape of our query, this could be done like:

```sql
  WITH pgrst_source AS (
    SELECT
      "test"."client".*
    FROM "test"."client"
    LEFT JOIN LATERAL (
      SELECT json_agg("client_clientinfo_1") AS "client_clientinfo_1"
      FROM (
        SELECT
          "clientinfo_1".*
        FROM "test"."clientinfo" AS "clientinfo_1"
        WHERE
        "clientinfo_1"."clientid" = "test"."client"."id" AND
        "clientinfo_1"."other" LIKE '%Main%'
      ) AS "client_clientinfo_1"
    ) AS "client_clientinfo_1" ON TRUE
    LEFT JOIN LATERAL (
[...]

PR #4011, opened 2025-04-13
feat: add related filters

laurenceisla:
Implement related AND/OR according to https://github.com/PostgREST/postgrest/issues/2563#issuecomment-1361775232

- [ ] Related conditions as query parameters `embed(col)=eq.val`.
- [x] Related conditions inside OR/AND `or=(embed1(col1).eq.val1,embed2(col2).eq.val2)`
- [ ] Tests


test/spec/Feature/Query/RelatedQueriesSpec.hs
```diff
@@ -378,3 +378,14 @@ spec = describe "related queries" $ do
...
+  context "related conditions using the embed(column) syntax" $ do
+    it "can do an or across embeds" $
+      get "/client?select=*,clientinfo(),contact()&or=(clientinfo(other).like.*Main*,contact(name).like.*Tabby*)" `shouldRespondWith`
+        [json|[
+          {"id":1,"name":"Walmart"},
+          {"id":2,"name":"Target"}
+        ]|]
+        { matchStatus  = 200
+        , matchHeaders = [matchContentTypeJson]
+        }
```
laurenceisla, line 391: Currently working for the simplest case, e.g. (didn't use empty embeds to illustrate):

```bash
curl 'localhost:3000/client?select=*,clientinfo(other),contact(name)&or=(clie
ntinfo(other).like.*Main*,contact(name).like.*Tabby*)'
```

Gives the same result as:

```bash
curl 'localhost:3000/client?select=*,clientinfo(other),contact(name)&clientin
fo.other=like.*Main*&contact.name=like.*Tabby*&or=(clientinfo.not.is.null,contact.not.is.null)'
```

```json
[
 {"id":1, "name":"Walmart", "clientinfo":[{"other": "123 Main St"}], "contact":[]},
 {"id":2, "name":"Target", "clientinfo":[], "contact":[{"name": "Tabby Targo"}]}
]
```

There are other possible cases that need to be considered. One of those is how to use this syntax to filter nested relationships https://github.com/PostgREST/postgrest/discussions/3726#discussioncomment-10685122

Issue #3934, opened 2025-03-05
Log all generated queries with `log-query=full`


steve-chavez:
## Problem

Currently [log-query](https://postgrest.org/en/latest/references/observability.html#sql-query-logs) doesn't log:

- [transaction-scoped-settings](https://docs.postgrest.org/en/v12/references/transactions.html#transaction-scoped-settings)
- `explain` query whenever `Prefer: count=estimated` is done ([ref](https://github.com/PostgREST/postgrest/blob/d0173e71923e15b412fd459e35e1d5b97161a363/src/PostgREST/Query.hs#L234-L236)).
- optional rollback query ([ref](https://github.com/PostgREST/postgrest/blob/d0173e71923e15b412fd459e35e1d5b97161a363/src/PostgREST/Query.hs#L256-L265))

## Solution

Log all generated queries for `log-query = full`.

## Implementation details

- log-query was added with some problems as mentioned on https://github.com/PostgREST/postgrest/pull/3904#issue-2832454798
- To implement this cleanly, I think we need to stop using `Hasql.Snippet` in our internal modules and only use it at the top level (`App.hs`).
- Then we would have something like:
```haskell
  apiReq    <- ApiRequest.userApiRequest conf req body sCache
  plan      <- Plan.actionPlan iAction conf apiReq sCache
  queries   <- Statement.generateSQL plan
  -- here we can now log the queries as they would be simple `ByteString`s
  resultset <- Decoders.decode $ -- since generateSQL won't generate Hasql.Snippet, we would now need an explicit decoding step (it would be good to expose this as a new timing too)
               Query.execute queries $ -- here we now execute it with Hasql
 -- then build the response as usual

-- code simplified from https://github.com/PostgREST/postgrest/blob/d0173e71923e15b412fd459e35e1d5b97161a363/src/PostgREST/App.hs#L148-L166
```

steve-chavez:
> Log all generated queries for log-query = full.

Thinking more about this, to be really `full` we would need to also log the Schema cache queries at startup. That would need a similar treatment of "purifying" our SQL generation from Hasql too, I believe.

steve-chavez:
> To implement this cleanly, I think we need to stop using Hasql.Snippet in our internal modules and only use it at the top level (App.hs). This would have the nice side-effect of making our query generation "pure" too.

A simpler step without touching `Hasql.Snippet` might be to split these `decodeIt` steps into their own module:

[...]

PR #4012, opened 2025-04-14
feat: [WIP] parameters tracked snippets, and application/json+sql response

### LINKED ISSUE #3580: Returning generated SQL
In issue #2354 it is discussed multiple times that displaying the generated SQL might be interesting too. It seems like this feature was in the end not implemented.

I unterstand that there are other ways to retrieve the SQL query (for example through logs and using EXPLAIN VERBOSE to get the query ID to make sure to find it easily...

However, any explain plan returned through the plan accept header implemented is always going to be easier to understand, judge and also act upon with the SQL query present in the same data without having to implement extra steps which enable logging queries. It seems a lot more efficient to have the option to just return the SQL query right before the plan in the same reply?

I understand security considerations need to be taken, but this could be implemented via an extra config as db-allow-generated-sql or db-allow-plan-sql if it is only usable in combination with plan.

fauh45:
This should closes #3580 

This PR is intended to add Raw SQL output when the user queries for certain Accept media type on their request. The changes are pretty big, mostly because the as far as my knowledge goes, there's isn't really any way to get the parameters value from `Hasql` snippets (somewhat related to #3934). In my mind there's only two option to get the parameters.

1. Modify `Hasql`, and I don't think I'm ready to do that
2. Or, add a wrapper to `Hasql.Snippet` where it could save the encoded value as well, I've done this through `TrackedSnippets`.

All `TrackedSnippets` do is handle the Snippets as usual, but also save the parameters as `Maybe ByteString` (from what I see all the parameters are `ByteString` or `Maybe ByteString` in the repository). Now because there's a lot of function depending on `Hasql.Snippet` there's a lot of changes just to change the parameter or return type to `TrackedSnippet`, which also means snippets of SQL especially in query building needs to be converted to `TrackedSnippet`to work. Thus a lot of the changes are basically just converting to `TrackSnippet` and the other way around.

There's also an interface to get both of the query and also the parameters with `Accept: application/json+sql`. This will return a response with `{ "sql": "<raw sql generated>", "params": ["params1", "params2"] }`, to be noted that the params here is string representation of the params. So in case of JSON placeholder that uses `jsonbLazyBytes` I'm not sure how it will handle it (also I seems to be lost on how the JSON params could be a jsonb?).

### TODO

Below is list that I'm sure there's some changes needed to make this work.
- [ ] Is this the best way to achieve this goals?
- [ ] Does the CHANGELOG need to be more descriptive?
- [ ] Which part of the Docs you guys think it should be added?
- [ ] Better testing for the `application/json+sql`
- [ ] Make serialized parameters can work with the sql returned. Especially with JSON placeholder as it seems it is encoded as `json` or `jsonb` direcly by Hasql. While the serialized parameters is currently still serialized to string. Code snippet that handles it on `SqlFragment.hs` (`jsonPlaceHolder = TrackedSnippet (SQL.encoderAndParam (HE.nullable $ if includeDefaults then HE.jsonbLazyBytes else HE.jsonLazyBytes) body) [LBS.toStrict <$> body]`)

### DIFF
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -26,6 +26,7 @@ import           Protolude
 -- | Enumeration of currently supported media types
 data MediaType
   = MTApplicationJSON
+  | MTApplicationSQL
   | MTGeoJSON
   | MTTextCSV
   | MTTextPlain
@@ -65,6 +66,7 @@ toContentType ct = (hContentType, toMime ct <> charset)
 -- | Convert from MediaType to a ByteString representing the mime type
 toMime :: MediaType -> ByteString
 toMime MTApplicationJSON      = "application/json"
+toMime MTApplicationSQL       = "application/sql"
 toMime MTVndArrayJSONStrip    = "application/vnd.pgrst.array+json;nulls=stripped"
 toMime MTGeoJSON              = "application/geo+json"
 toMime MTTextCSV              = "text/csv"
@@ -133,6 +135,7 @@ decodeMediaType mt = decodeMediaType' $ decodeLatin1 mt
     decodeMediaType' mt' =
       case (T.toLower mainType, T.toLower subType, params) of
         ("application", "json", _)                  -> MTApplicationJSON
+        ("application", "sql", _)                   -> MTApplicationSQL
         ("application", "geo+json", _)              -> MTGeoJSON
         ("text", "csv", _)                          -> MTTextCSV
         ("text", "plain", _)                        -> MTTextPlain
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -26,7 +26,7 @@ import           Protolude
 -- | Enumeration of currently supported media types
 data MediaType
   = MTApplicationJSON
-  | MTApplicationSQL
+  | MTApplicationJSONSQL
   | MTGeoJSON
   | MTTextCSV
   | MTTextPlain
@@ -66,7 +66,7 @@ toContentType ct = (hContentType, toMime ct <> charset)
 -- | Convert from MediaType to a ByteString representing the mime type
 toMime :: MediaType -> ByteString
 toMime MTApplicationJSON      = "application/json"
-toMime MTApplicationSQL       = "application/sql"
+toMime MTApplicationJSONSQL   = "application/json+sql"
 toMime MTVndArrayJSONStrip    = "application/vnd.pgrst.array+json;nulls=stripped"
 toMime MTGeoJSON              = "application/geo+json"
 toMime MTTextCSV              = "text/csv"
@@ -135,7 +135,7 @@ decodeMediaType mt = decodeMediaType' $ decodeLatin1 mt
     decodeMediaType' mt' =
       case (T.toLower mainType, T.toLower subType, params) of
         ("application", "json", _)                  -> MTApplicationJSON
-        ("application", "sql", _)                   -> MTApplicationSQL
+        ("application", "json+sql", _)              -> MTApplicationJSONSQL
         ("application", "geo+json", _)              -> MTGeoJSON
         ("text", "csv", _)                          -> MTTextCSV
         ("text", "plain", _)                        -> MTTextPlain
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -1088,7 +1088,7 @@ negotiateContent conf ApiRequest{iAction=act, iPreferences=Preferences{preferRep
     -- TODO: despite no aggregate, these are responding with a Content-Type, which is not correct.
     (ActDb (ActRelationRead _ True),             Just (_, mt)) -> Right (NoAgg, mt)
     (ActDb (ActRoutine  _ (InvRead True)), Just (_, mt))             -> Right (NoAgg, mt)
-    (_, Just (_, MTApplicationSQL))                          -> Right (NoAgg, MTApplicationSQL)
+    (_, Just (_, MTApplicationJSONSQL))                          -> Right (NoAgg, MTApplicationJSONSQL)
     (_, Just (x, mt))                                        -> Right (x, mt)
   where
     firstAcceptedPick = listToMaybe $ mapMaybe matchMT accepts -- If there are multiple accepted media types, pick the first. This is usual in content negotiation.
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -1088,6 +1088,7 @@ negotiateContent conf ApiRequest{iAction=act, iPreferences=Preferences{preferRep
     -- TODO: despite no aggregate, these are responding with a Content-Type, which is not correct.
     (ActDb (ActRelationRead _ True),             Just (_, mt)) -> Right (NoAgg, mt)
     (ActDb (ActRoutine  _ (InvRead True)), Just (_, mt))             -> Right (NoAgg, mt)
+    (_, Just (_, MTApplicationSQL))                          -> Right (NoAgg, MTApplicationSQL)
     (_, Just (x, mt))                                        -> Right (x, mt)
   where
     firstAcceptedPick = listToMaybe $ mapMaybe matchMT accepts -- If there are multiple accepted media types, pick the first. This is usual in content negotiation.
--- a/test/spec/Feature/Query/RawSQLSpec.hs
+++ b/test/spec/Feature/Query/RawSQLSpec.hs
@@ -12,32 +12,32 @@ import Protolude
 import SpecHelper (acceptHdrs)
 
 spec :: SpecWith ((), Application)
-spec = describe "When accept header is set to application/sql" $ do
-  let mtApplicationSQLHdrs = acceptHdrs "application/sql"
+spec = describe "When accept header is set to application/json+sql" $ do
+  let mtApplicationJSONSQLHdrs = acceptHdrs "application/json+sql"
 
   it "responds raw sql to a GET request" $ do
-    r <- request methodGet "/items?id=eq.1" mtApplicationSQLHdrs ""
+    r <- request methodGet "/items?id=eq.1" mtApplicationJSONSQLHdrs ""
     liftIO $ do
       simpleStatus r `shouldBe` status200
-      simpleHeaders r `shouldContain` [("Content-Type", "application/sql; charset=utf-8")]
+      simpleHeaders r `shouldContain` [("Content-Type", "application/json; charset=utf-8")]
 
   it "responds raw sql to a POST request for insert" $ do
     r <- request methodPost "/projects"
-             (acceptHdrs "application/sql") [json|{"id":100, "name": "Project 100"}|]
+             mtApplicationJSONSQLHdrs [json|{"id":100, "name": "Project 100"}|]
     liftIO $ do
       simpleStatus r `shouldBe` status200
-      simpleHeaders r `shouldContain` [("Content-Type", "application/sql; charset=utf-8")]
+      simpleHeaders r `shouldContain` [("Content-Type", "application/json; charset=utf-8")]
 
   it "responds raw sql to a create POST for update" $ do
     r <- request methodPatch "/projects?id=eq.3"
-             (acceptHdrs "application/sql") [json|{"name": "Patched Project"}|]
+             mtApplicationJSONSQLHdrs [json|{"name": "Patched Project"}|]
     liftIO $ do
       simpleStatus r `shouldBe` status200
-      simpleHeaders r `shouldContain` [("Content-Type", "application/sql; charset=utf-8")]
+      simpleHeaders r `shouldContain` [("Content-Type", "application/json; charset=utf-8")]
 
   it "responds raw sql to a create POST for delete" $ do
     r <- request methodDelete "/projects?id=in.(1,2,3)"
-             (acceptHdrs "application/sql") ""
+             mtApplicationJSONSQLHdrs ""
     liftIO $ do
       simpleStatus r `shouldBe` status200
-      simpleHeaders r `shouldContain` [("Content-Type", "application/sql; charset=utf-8")]
+      simpleHeaders r `shouldContain` [("Content-Type", "application/json; charset=utf-8")]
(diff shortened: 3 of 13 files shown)


fauh45:
@wolfgangwalther this should be my current solution for the issue, the scope is a little bit more than #3580 as now it also returns the placeholder values

steve-chavez:
@fauh45 How about clearing all CI failures? Then we can further review.

fauh45:
@steve-chavez will do! 🫡

fauh45:
@steve-chavez fixed formatting, also added `MTApplicationJSONSQL` to `dbMediaHandler` so that the test case reflects the current media handler list. Tested locally with `postgrest-check` and it seems to passed.

fauh45:
Oh nvm, forgot to push minor fix to remove unused functions. Sorry!

src/PostgREST/MediaType.hs
```diff
@@ -65,6 +66,7 @@ toContentType ct = (hContentType, toMime ct <> charset)
 -- | Convert from MediaType to a ByteString representing the mime type
 toMime :: MediaType -> ByteString
 toMime MTApplicationJSON      = "application/json"
+toMime MTApplicationJSONSQL   = "application/json+sql"
```
steve-chavez, line 69: `application/json` suffixes have to be registered on IANA to be valid ([ref](https://en.wikipedia.org/wiki/Media_type#Suffix)).

> Unregistered suffixes should not be used

So we would need a vendored media type. I suggest `application/vnd.pgrst.sql+json`.

test/spec/Feature/Query/RawSQLSpec.hs
```diff
@@ -0,0 +1,43 @@
...
+
+spec :: SpecWith ((), Application)
+spec = describe "When accept header is set to application/json+sql" $ do
+  let mtApplicationJSONSQLHdrs = acceptHdrs "application/json+sql"
+
+  it "responds raw sql to a GET request" $ do
+    r <- request methodGet "/items?id=eq.1" mtApplicationJSONSQLHdrs ""
+    liftIO $ do
+      simpleStatus r `shouldBe` status200
+      simpleHeaders r `shouldContain` [("Content-Type", "application/json; charset=utf-8")]
```
steve-chavez, line 22: We would need some samples of the response bodies in these tests.

For now I've checked it locally and it looks like:

```bash
curl "localhost:3000/projects?id=eq.1&name=eq.Joe" -H "Accept: application/json+sql"

{"params":["1","Joe"],"sql":"WITH pgrst_source AS ( SELECT \"test\".\"projects\".* FROM \"test\".\"projects\" WHERE  \"test\".\"projects\".\"id\" = $1 AND  \"test\".\"projects\".\"name\" =
 $2  )  SELECT null::bigint AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, ''::text AS body, nullif(current_setting('response.headers', true), '') AS response_headers, 
nullif(current_setting('response.status', true), '') AS response_status, '' AS response_inserted FROM ( SELECT * FROM pgrst_source ) _postgrest_t"}
```

It's an interesting functionality for sure.

src/PostgREST/Plan.hs
```diff
@@ -1088,6 +1088,7 @@ negotiateContent conf ApiRequest{iAction=act, iPreferences=Preferences{preferRep
     -- TODO: despite no aggregate, these are responding with a Content-Type, which is not correct.
     (ActDb (ActRelationRead _ True),             Just (_, mt)) -> Right (NoAgg, mt)
     (ActDb (ActRoutine  _ (InvRead True)), Just (_, mt))             -> Right (NoAgg, mt)
+    (_, Just (_, MTApplicationJSONSQL))                          -> Right (NoAgg, MTApplicationJSONSQL)
```
steve-chavez, line 1091: This media type should be opt in, I suggest reusing the https://docs.postgrest.org/en/v12/references/configuration.html#db-plan-enabled (`configDbPlanEnabled`) for this. See how it's done on:

https://github.com/PostgREST/postgrest/blob/c732591f37533171fd5bfd4d5c2566f578c03d84/src/PostgREST/Plan.hs#L1098-L1104

It's a bit messy now unfortunately.

test/spec/Feature/Query/RawSQLSpec.hs
```diff
@@ -0,0 +1,43 @@
...
+
+  it "responds raw sql to a POST request for insert" $ do
+    r <- request methodPost "/projects"
+             mtApplicationJSONSQLHdrs [json|{"id":100, "name": "Project 100"}|]
+    liftIO $ do
+      simpleStatus r `shouldBe` status200
+      simpleHeaders r `shouldContain` [("Content-Type", "application/json; charset=utf-8")]
```
steve-chavez, line 29: Looks like the parameters are not complete when having both POST body and query string parameters:

```bash
http POST "localhost:3000/projects?select=*&id=eq.1&name=eq.Joe" "Accept: application/json+sql" "Prefer:return=representation" <<EOF
{"id":2}
EOF
HTTP/1.1 200 OK
Date: Tue, 22 Apr 2025 05:25:48 GMT
Server: postgrest/12.3 (pre-release)
Transfer-Encoding: chunked

{
    "params": [
        "{\"id\":2}\n"
    ],
    "sql": "WITH pgrst_source AS (INSERT INTO \"test\".\"projects\"(\"id\") SELECT \"pgrst_body\".\"id\" FROM (SELECT $1 AS json_data) pgrst_payload, LATERAL (SELECT \"id\" FROM json_to_re
cord(pgrst_payload.json_data) AS _(\"id\" integer) ) pgrst_body  RETURNING \"test\".\"projects\".*) SELECT '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, array[]::te
xt[] AS header, ''::text AS body, nullif(current_setting('response.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status, '' AS 
response_inserted FROM (SELECT * FROM pgrst_source) _postgrest_t"
}
```

id=1 and name=Joe are missing.

src/PostgREST/MediaType.hs
steve-chavez: > application/sql+json

IIUC, we would have to register the `+json` to the SQL media type if we want to use it like that.

That being said, I like that media type too and it doesn't seem "potentially damaging", unlike adding a suffix to `application/json`.

fauh45: I do like the idea of vendoring the media type, though prefixing `sql` makes it more clear though IMO. Just to be safe, I'll probably go with the vendored media type for now

test/spec/Feature/Query/RawSQLSpec.hs
fauh45: Will got it added!

src/PostgREST/Plan.hs
fauh45: Ah so in a way on the Plan there should be a check to allow it if the `configDbPlanEnabled`? Honestly I totally understand that, enabling this does open up lots of internal details

test/spec/Feature/Query/RawSQLSpec.hs
fauh45: CMMIW here, isn't `POST` (insert operation) doesn't really take query string to make the query at all (`INSERT` cannot have `WHERE`)?

steve-chavez: Hm, this is weird. I recall we were able to filter the returned inserted rows, this made sense on bulk inserts. Like:

```bash
http POST "localhost:3000/projects?id=eq.15" "Prefer:return=representation" <<EOF
[{"id":14, "name": "asdf"}, {"id":15, "name": "qer"}]
EOF
```

But looks like that's not working anymore.

steve-chavez: Ok, got it. Filtering does work for POST but only for embedded resources:

```bash
http POST "localhost:3000/projects?select=*,clients(*)" "Prefer:return=representation" <<EOF
[{"id":14, "name": "asdf", "client_id": 1}, {"id":15, "name": "qer", "client_id": 2}]
EOF

[
    {
        "client_id": 1,
        "clients": {
            "id": 1,
            "name": "Microsoft"
        },
        "id": 14,
        "name": "asdf"
    },
    {
        "client_id": 2,
        "clients": {
            "id": 2,
            "name": "Apple"
        },
        "id": 15,
        "name": "qer"
    }
]

```

With filtering:

```bash
http POST "localhost:3000/projects?select=*,clients(*)&clients.id=eq.2" "Prefer:return=representation" <<EOF

[
    {
        "client_id": 1,
        "clients": null,
        "id": 14,
        "name": "asdf"
    },
    {
        "client_id": 2,
        "clients": {
            "id": 2,
            "name": "Apple"
        },
        "id": 15,
        "name": "qer"
    }
]
```

steve-chavez: That's because we remove the root filters for all mutations:

https://github.com/PostgREST/postgrest/blob/c732591f37533171fd5bfd4d5c2566f578c03d84/src/PostgREST/Plan.hs#L803-L807

Only `qsFiltersNotRoot` are applied

steve-chavez: That logic is so strange because in the end we're still able to filter the inserted rows by applying `!inner`:

```bash
http POST 'localhost:3000/projects?select=*,clients!inner(*)&clients.id=eq.2' "Prefer:return=representation" <<EOF
[{"id":14, "name": "asdf", "client_id": 1}, {"id":15, "name": "qer", "client_id": 2}]
EOF

[
    {
        "client_id": 2,
        "clients": {
            "id": 2,
            "name": "Apple"
        },
        "id": 15,
        "name": "qer"
    }
]
```

steve-chavez: Will open an issue about the above. Wrt to this PR, there is a problem when considering that case.

The generated query does contain a `$2` placeholder referring to the filter:

```bash
$ PGRST_LOG_LEVEL="info" PGRST_LOG_QUERY="main-query" postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run
```

```bash
# request
http POST 'localhost:3000/projects?select=*,clients!inner(*)&clients.id=eq.2' "Prefer:return=representation"  <<EOF                              
[{"id":14, "name": "asdf", "client_id": 1}, {"id":15, "name": "qer", "client_id": 2}]
EOF

# logs
23/Apr/2025:13:35:17 -0500: WITH pgrst_source AS (INSERT INTO "test"."projects"("client_id", "id", "name") SELECT "pgrst_body"."client_id", "pgrst_body"."id", "pgrst_body"."name" FROM (SE
LECT $1 AS json_data) pgrst_payload, LATERAL (SELECT "client_id", "id", "name" FROM json_to_recordset(pgrst_payload.json_data) AS _("client_id" integer, "id" integer, "name" text) ) pgrst
_body  RETURNING "test"."projects".*) SELECT '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header, coalesce(json_agg(_postgrest_t), '[]') AS bod
y, nullif(current_setting('response.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status, '' AS response_inserted FROM (SELECT
 "projects".*, row_to_json("projects_clients_1".*)::jsonb AS "clients" FROM "pgrst_source" AS "projects" INNER JOIN LATERAL ( SELECT "clients_1".* FROM "test"."clients" AS "clients_1" WHE
RE  "clients_1"."id" = $2 AND "clients_1"."id" = "projects"."client_id"  ) AS "projects_clients_1" ON TRUE ) _postgrest_t
127.0.0.1 - postgrest_test_anonymous [23/Apr/2025:13:35:17 -0500] "POST /projects?select=*,clients!inner(*)&clients.id=eq.2 HTTP/1.1" 201 - "" "HTTPie/3.2.2"
```

But when using the new header, the `$2` placeholder is not present:

```bash
## request
http POST 'localhost:3000/projects?select=*,clients!inner(*)&clients.id=eq.2' "Prefer:return=representation" "Accept: application/json+sql" <<EOF
[{"id":14, "name": "asdf", "client_id": 1}, {"id":15, "name": "qer", "client_id": 2}]
EOF

#logs
23/Apr/2025:13:36:17 -0500: WITH pgrst_source AS (INSERT INTO "test"."projects"("client_id", "id", "name") SELECT "pgrst_body"."client_id", "pgrst_body"."id", "pgrst_body"."name" FROM (SE
LECT $1 AS json_data) pgrst_payload, LATERAL (SELECT "client_id", "id", "name" FROM json_to_recordset(pgrst_payload.json_data) AS _("client_id" integer, "id" integer, "name" text) ) pgrst
_body  RETURNING "test"."projects".*) SELECT '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, array[]::text[] AS header, ''::text AS body, nullif(current_setting('res
ponse.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status, '' AS response_inserted FROM (SELECT * FROM pgrst_source) _postgre
st_t
127.0.0.1 - postgrest_test_anonymous [23/Apr/2025:13:36:17 -0500] "POST /projects?select=*,clients!inner(*)&clients.id=eq.2 HTTP/1.1" 200 - "" "HTTPie/3.2.2"

# same thing on the client
```

Maybe the same problem will be present on PATCH or PUT.

fauh45: Okay this is something I don't really understand that deeply about the codebase just yet haha. That's kinda weird... Is it in any way related to the Plan somehow or is more part of the Query builder part? Because technically it should still parse the query string, then build the read plan, lastly build the query based on the plan right?

steve-chavez: > That's kinda weird... Is it in any way related to the Plan somehow or is more part of the Query builder part?

The `MutateReadPlan` has a `ReadPlanTree` besides the `MutatePlan`:

https://github.com/PostgREST/postgrest/blob/d005ed341f0e715a56326fe60ea3291d58692135/src/PostgREST/Plan.hs#L106-L114

> Because technically it should still parse the query string, then build the read plan, lastly build the query based on the plan right?

Yes, this still holds. The above is still a single query.

PR #4013, opened 2025-04-14
fix: Fix regression that replaces an unknown media type with */*

taimoorzaeem:
Solves regression introduced in #4005.


taimoorzaeem:
Does this need a CHANGELOG entry? We haven't released it yet, except the `dlevel` so I think probably not?

PR #4018, opened 2025-04-16
feat: add content-length header

### LINKED ISSUE #4016: Set `Content-Length` header
## Problem

Currently the standard [Content-Length](https://httpwg.org/specs/rfc9110.html#field.content-length) is not calculated. This is useful for observability since operators can know which responses are the largest in size.

Proxies like Nginx don't calculate the content-length for upstream services since it would have to buffer the whole response size, which is undesirable for performance reasons in most cases. So this responsability is left to the upstream service.

## Solution 1

**Edit**: too complicated, go to Solution 2.

PostgreSQL has the [octet_length](https://www.postgresql.org/docs/current/functions-binarystring.html) function for this case, it could be used like:

```sql
WITH pgrst_source AS ( SELECT "test"."projects".* FROM "test"."projects"  )
SELECT
-- here
    octet_length(json_agg(_postgrest_t)::text) as content_length_header,
    '' AS response_inserted
FROM ( SELECT * FROM pgrst_source ) _postgrest_t;
```

And then set the `Content-Length` header with the `content_length_header` value.

We would need to avoid repeating the `json_agg`, maybe by putting it into a CTE.

There's also the [pg_column_size](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-DBOBJECT) function, but not sure if it shows the real size (maybe it doesn't decompress the TOASTed values) 

Since this could be expensive, we could:

- Add a config option to enable it.
- Or enable it per request with a new `Prefer` header value.

## Solution 2

- Maybe its cheaper and simpler to calculate this on Haskell.

laurenceisla:
Closes #4016

### DIFF
--- a/docs/references/observability.rst
+++ b/docs/references/observability.rst
@@ -15,14 +15,14 @@ Observability allows measuring a system's current state based on the data it gen
 Logs
 ====
 
-PostgREST logs basic request information to ``stdout``, including the authenticated user if available, the requesting IP address and user agent, the URL requested, and HTTP response status.
+PostgREST logs basic request information to ``stdout``, including the authenticated user if available, the requesting IP address and user agent, the URL requested, the HTTP response status and the response body size in bytes if available.
 
 With :ref:`log-level` set to ``info``, we get:
 
 .. code::
 
-   127.0.0.1 - user [26/Jul/2021:01:56:38 -0500] "GET /clients HTTP/1.1" 200 - "" "curl/7.64.0"
-   127.0.0.1 - anonymous [26/Jul/2021:01:56:48 -0500] "GET /unexistent HTTP/1.1" 404 - "" "curl/7.64.0"
+   127.0.0.1 - user [26/Jul/2021:01:56:38 -0500] "GET /clients HTTP/1.1" 200 56 "" "curl/7.64.0"
+   127.0.0.1 - anonymous [26/Jul/2021:01:56:48 -0500] "GET /unexistent HTTP/1.1" 404 162 "" "curl/7.64.0"
 
 For diagnostic information about the server itself, PostgREST logs to ``stderr``:
 
@@ -73,7 +73,7 @@ This will be logged by PostgREST:
 .. code::
 
   17/Feb/2025:17:28:15 -0500: WITH pgrst_source AS ( SELECT "public"."protected_table".* FROM "public"."protected_table"  )  SELECT null::bigint AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, coalesce(json_agg(_postgrest_t), '[]') AS body, nullif(current_setting('response.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status, '' AS response_inserted FROM ( SELECT * FROM pgrst_source ) _postgrest_t
-  127.0.0.1 - web_anon [17/Feb/2025:17:28:15 -0500] "GET /protected_table HTTP/1.1" 401 - "" "curl/8.7.1"
+  127.0.0.1 - web_anon [17/Feb/2025:17:28:15 -0500] "GET /protected_table HTTP/1.1" 401 99 "" "curl/8.7.1"
 
 Database Logs
 -------------
@@ -269,6 +269,27 @@ This header communicates metrics of the different phases in the request-response
 .. note::
 
   We're working on lowering the duration of the ``parse`` and ``plan`` stages on https://github.com/PostgREST/postgrest/issues/2816.
+
+.. _content-length_header:
+
+Content-Length Header
+---------------------
+
+You can verify the response body size in bytes in the `Content-Length header <https://httpwg.org/specs/rfc9110.html#field.content-length>`_.
+
+.. code-block:: bash
+
+  curl -i 'localhost:3000/users'
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Content-Length: 104
+
+Note that this header won't be returned on ``HEAD`` requests for optimization purposes (see :ref:`head_req`).
+This is in line with `RFC 9110 <https://httpwg.org/specs/rfc9110.html#field.content-length>`_.
+
+The body size is also present in the :ref:`PostgREST logs <pgrst_logging>`.
 
 .. _explain_plan:
 
--- a/src/PostgREST/Response.hs
+++ b/src/PostgREST/Response.hs
@@ -68,6 +68,7 @@ actionResponse (DbCrudResult WrappedReadPlan{wrMedia, wrHdrsOnly=headersOnly, cr
     RSStandard{..} -> do
       let
         (status, contentRange) = RangeQuery.rangeStatusHeader iTopLevelRange rsQueryTotal rsTableTotal
+        cLHeader = if headersOnly then mempty else [contentLengthHeaderStrict rsBody]
         prefHeader = maybeToList . prefAppliedHeader $ Preferences Nothing Nothing preferCount preferTransaction Nothing preferHandling preferTimezone Nothing []
         headers =
           [ contentRange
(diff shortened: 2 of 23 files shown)


src/PostgREST/Response.hs
```diff
@@ -152,27 +154,28 @@ actionResponse (DbCrudResult MutateReadPlan{mrMutation=MutationUpdate, mrMedia}
...
+      cLHeader = [contentLengthHeader rsBody]
       cTHeader = contentTypeHeaders mrMedia ctxApiRequest
 
     let isInsertIfGTZero i = if i > 0 then HTTP.status201 else HTTP.status200
         upsertStatus       = isInsertIfGTZero $ fromJust rsInserted
         (status, headers, body) =
           case preferRepresentation of
-            Just Full -> (upsertStatus, cTHeader ++ prefHeader, LBS.fromStrict rsBody)
+            Just Full -> (upsertStatus, cLHeader ++ cTHeader ++ prefHeader, LBS.fromStrict rsBody)
             Just None -> (HTTP.status204,  prefHeader, mempty)
```
laurenceisla, line 171: [According to the spec](https://httpwg.org/specs/rfc9110.html#field.content-length), the `204` responses must not return `Content-Length`. 
> A server MUST NOT send a Content-Length header field in any response with a status code of [1xx (Informational)](https://httpwg.org/specs/rfc9110.html#status.1xx) or [204 (No Content)](https://httpwg.org/specs/rfc9110.html#status.204).

steve-chavez:
Loadtest looks good https://github.com/PostgREST/postgrest/actions/runs/14481781607?pr=4018, no perf loss.

src/PostgREST/Response.hs
```diff
@@ -68,6 +68,7 @@ actionResponse (DbCrudResult WrappedReadPlan{wrMedia, wrHdrsOnly=headersOnly, cr
     RSStandard{..} -> do
       let
         (status, contentRange) = RangeQuery.rangeStatusHeader iTopLevelRange rsQueryTotal rsTableTotal
+        cLHeader = if headersOnly then mempty else [contentLengthHeader rsBody]
```
laurenceisla, line 71: The body is not returned and processed in `HEAD` requests, so the length would be `0` here. The same resource using `GET` would return a length greater than `0`. [According to the spec](https://httpwg.org/specs/rfc9110.html#field.content-length) the `HEAD` and `GET` requests must return the same value in `Content-Length`, so we must not include the header altogether.

Paraphrased from:
> A server MAY send a Content-Length header field in a response to a HEAD request ([Section 9.3.2](https://httpwg.org/specs/rfc9110.html#HEAD)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a response if the same request had used the GET method.

test/io/test_io.py
```diff
@@ -964,45 +964,45 @@ def test_log_level(level, defaultenv):
             assert len(output) == 0
         elif level == "error":
             assert re.match(
-                r'- - - \[.+\] "GET / HTTP/1.1" 500 - "" "python-requests/.+"',
+                r'- - - \[.+\] "GET / HTTP/1.1" 500 \d+ "" "python-requests/.+"',
```
laurenceisla, line 967: This is a neat free addition. The logs also add the value of the `Content-Length` here (after the HTTP status).

laurenceisla:
Should be ready for review.

For tests I tried to add the `Content-Length` validation for  those that are most relevant, e.g verify that there's no header for `204` responses and head requests; or verify that it's present on each type of request/feature functions, explain plan,  `GET`, `POST`, etc. I tried to add at least one test for each case and not repeat it for similar ones.

test/spec/Feature/Query/PlanSpec.hs
```diff
@@ -33,6 +33,7 @@ spec actualPgVersion = do
 
       liftIO $ do
         resHeaders `shouldSatisfy` elem ("Content-Type", "application/vnd.pgrst.plan+json; for=\"application/json\"; charset=utf-8")
+        resHeaders `shouldSatisfy` notZeroContentLength
```
laurenceisla, line 36: It returns different values for different PG versions, so we just validate that it's present and it's not `0`.

test/io/test_io.py
steve-chavez: Super cool! Before it used to be 0 or blank?

laurenceisla: It used to be `-`, like `500 - - ""`. The modified docs have an example of this.

docs/references/observability.rst
```diff
@@ -362,6 +362,25 @@ For example, to only allow requests from an IP address to get the execution plan
...
+.. code-block:: bash
+
+  curl -i 'localhost:3000/users'
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Content-Length: 104
+
+Note that this header won't be returned on ``HEAD`` requests for optimization purposes (see :ref:`head_req`).
```
steve-chavez, line 381: I believe we need to link the RFC here
```suggestion
Note that this header won't be returned on ``HEAD`` requests for optimization purposes (see :ref:`head_req`). This is in line with RFC XXX.
```

steve-chavez: It might save some issues/discussions

docs/references/observability.rst
```diff
@@ -362,6 +362,25 @@ For example, to only allow requests from an IP address to get the execution plan
...
+The body size is also present in the :ref:`logs <pgrst_logging>` if they are enabled.
```
steve-chavez, line 382: `enabled` sounds like they need a config toggle, we use this term in other places with that meaning. Can we paraphrase?

src/PostgREST/Response.hs
```diff
@@ -205,28 +209,32 @@ actionResponse (DbCallResult CallReadPlan{crMedia, crInvMthd=invMethod, crProc=p
...
 
 actionResponse (MaybeDbResult InspectPlan{ipHdrsOnly=headersOnly} body) _ versions conf sCache schema negotiatedByProfile =
-  Right $ PgrstResponse HTTP.status200
-    (MediaType.toContentType MTOpenAPI : maybeToList (profileHeader schema negotiatedByProfile))
-    (maybe mempty (\(x, y, z) -> if headersOnly then mempty else OpenAPI.encode versions conf sCache x y z) body)
+  let
+    rsBody = maybe mempty (\(x, y, z) -> if headersOnly then mempty else OpenAPI.encode versions conf sCache x y z) body
+    cLHeader = if headersOnly then mempty else [contentLengthHeaderLazy rsBody]
+  in
+  Right $ PgrstResponse HTTP.status200 (MediaType.toContentType MTOpenAPI : cLHeader ++ maybeToList (profileHeader schema negotiatedByProfile)) rsBody
```
steve-chavez, line 237: I can see that we include the header in openapi now

src/PostgREST/Response.hs
```diff
@@ -268,6 +276,12 @@ decodeGucStatus :: Maybe Text -> Either Error.Error (Maybe HTTP.Status)
 decodeGucStatus =
   maybe (Right Nothing) $ first (const . Error.ApiRequestError $ Error.GucStatusError) . fmap (Just . toEnum . fst) . decimal
 
+contentLengthHeader :: BS.ByteString -> HTTP.Header
+contentLengthHeader body = (,) "Content-Length" (show $ BS.length body)
+
+contentLengthHeaderLazy :: LBS.ByteString -> HTTP.Header
```
steve-chavez, line 282: Two versions of the same logic is weird. Can we deduplicate here?

src/PostgREST/Response.hs
laurenceisla: Yes, it's now also included in RPC and errors too.

docs/references/observability.rst
laurenceisla: I modified to mention it directly "in the PostgREST logs", clicking the link should give the necessary info.

docs/references/observability.rst
```diff
@@ -362,6 +362,27 @@ For example, to only allow requests from an IP address to get the execution plan
  -- set this function on your postgrest.conf
  -- db-pre-request = filter_plan_requests
 
+.. _content-length_header:
```
steve-chavez, line 365: @laurenceisla Just checked RTD and this should go before https://postgrest--4018.org.readthedocs.build/en/4018/references/observability.html#execution-plan after https://postgrest--4018.org.readthedocs.build/en/4018/references/observability.html#server-timing-header. 

This is more important than Execution plan (which is not that well positioned)

PR #3140, opened 2024-01-04
feat: produce OpenTelemetry traces with `hs-opentelemetry`

develop7:
This PR introduces producing OpenTelemetry traces containing, among others, metrics same as in `ServerTiming` header from before.

## TODO:

- [x] document oTel configuration
- [x] make code location attributes show call sites instead of helpers
- [x] ~~build with Nix as well (for now Stack only)~~ DONE for free
- [x] ~make an example of exporting log messages~ `hs-opentelemetry` doesn't support logging, per https://github.com/iand675/hs-opentelemetry/issues/100
- [x] ~~make `getTracer` available globally: we're interested in using as many different spans as it makes sense, so `getTracer` should be available everywhere, as described in `hs-opentelemetry-sdk`'s README~~ seems like impossible without major refactoring
- [x] make use of `hs-opentelemetry-wai` middleware
- [x] ~~look into failing Windows builds~~ `hs-opentelemetry-sdk` depends on `unix`, tracking in https://github.com/iand675/hs-opentelemetry/issues/109
- [x] fix "memory test" failures (https://github.com/PostgREST/postgrest/actions/runs/7888534966/job/21526302784?pr=3140)
- [x] fix dependency issues on 9.8 (https://github.com/PostgREST/postgrest/actions/runs/7888534966/job/21526300453?pr=3140)

## Running:

I sort of gave up deploying and configuring all the moving bits locally, so you'd need to create the honeycomb.io account for this one (or ask me for the invite). After that, it's quite straightforward:

1. Build PostgREST executable with `stack build`, and get its path with `stack exec -- which postgrest`
1. Get a PostgreSQL server running (e.g. run `nix-shell`, then `postgrest-with-postgresql-15 --fixture ./test/load/fixture.sql -- cat`). Note the server URL, you'll need it when running PostgREST server
2. get a JWT token with default secret by running `postgrest-jwt --exp 36000 postgrest_test_anonymous`
3. Run PostgREST server with 
    ```shell
	OTEL_EXPORTER_OTLP_ENDPOINT='https://api.honeycomb.io/' \
	OTEL_EXPORTER_OTLP_HEADERS="x-honeycomb-team=<honeycomb_api_key>"  \
	OTEL_SERVICE_NAME='PostgREST' OTEL_LOG_LEVEL='debug' OTEL_TRACES_SAMPLER='always_on' \
	PGRST_DB_URI='<postgresql_server_url>'   \
	postgrest-run 
	```
[...]

PR #4020, opened 2025-04-16
bump version to 13.0.0

wolfgangwalther:
We should really make a new release.

Which breaking changes are still in the pipeline that you would like to include, @steve-chavez?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [13.0.0] - 2025-04-16
+
 ### Added
 
  - #3558, Add the `admin-server-host` config to set the host for the admin server - @develop7
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            12.3
+version:            13.0.0
 synopsis:           REST API for any Postgres database
 description:        Reads the schema of a PostgreSQL database and creates RESTful routes
                     for tables, views, and functions, supporting all HTTP methods that security


steve-chavez:
I agree.

Could we wait for https://github.com/PostgREST/postgrest/pull/4008 since it's going to be a breaking change?

steve-chavez:
We also need to upgrade the postgresql compatibility on https://docs.postgrest.org/en/latest/explanations/install.html#supported-postgresql-versions, now that we don't support various pg versions.

develop7:
Would #3140 require a major version bump or a minor one is okay?

steve-chavez:
@develop7 #3140 should be ok in a minor. We still need to review that though, I recall there were some perf concerns.

jwt cache might already have perf implications so it'd be better to split otel in a different release.

steve-chavez:
> If you want a postgrest / postgresql versions compatibility matrix, I think the place to put it would be the github README, but not the version-specific docs

Fair, will do.

steve-chavez:
> Given that this turns out to be https://github.com/PostgREST/postgrest/pull/4008#issuecomment-2850907564, I think we should not wait for that.

I agree on not waiting for #4008. Although I think we need to merge https://github.com/PostgREST/postgrest/pull/4043 before releasing, otherwise it would cause a breaking change? @laurenceisla Can you clarify?

laurenceisla:
> Although I think we need to merge https://github.com/PostgREST/postgrest/pull/4043 before releasing

This is already merged, so it shouldn't be a problem to go ahead with the release. Unless there are other issues to tackle first, I see there's another JWT error in #4048, but not sure if it's a blocker.

steve-chavez:
I agree with releasing. 

> I see there's another JWT error in https://github.com/PostgREST/postgrest/issues/4048, but not sure if it's a blocker.

I think we can do a patch release for that one, once we can reproduce.

steve-chavez:
Just FYI. I edited the [release notes for v13](https://github.com/PostgREST/postgrest/releases/tag/v13.0.0) manually. It has [some differences with the CHANGELOG](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#1300---2025-05-08).

- I thought we were too strict in labeling breaking changes, so I moved some of them into fixes. 
- Some fixes related to errors messages, are actually improvements. We've considered those as features [before](https://github.com/PostgREST/postgrest/releases/tag/v12.2.0).

PR #4024, opened 2025-04-18
perf: Purge JWT cache asynchronously in a separate thread

steve-chavez:
Rebased and taken from https://github.com/PostgREST/postgrest/pull/3889

### DIFF
--- a/nix/tools/generate_targets.py
+++ b/nix/tools/generate_targets.py
@@ -10,7 +10,7 @@ import sys
 SECRET = b"reallyreallyreallyreallyverysafe"
 URL = "http://postgrest"
 JWT_DURATION = 120
-TOTAL_TARGETS = 30000  # tuned by hand to reduce result variance
+TOTAL_TARGETS = 200_000  # tuned by hand to reduce result variance
 
 
 def base64url_encode(data: bytes) -> str:
@@ -56,7 +56,7 @@ def main():
     lines = []
     for _ in range(TOTAL_TARGETS):
         token = generate_jwt()
-        lines.append(f"GET {URL}/authors_only")
+        lines.append(f"OPTIONS {URL}/authors_only")
         lines.append(f"Authorization: Bearer {token}")
         lines.append("")  # blank line to separate requests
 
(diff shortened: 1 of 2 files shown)


steve-chavez:
Hm, this is no good. The results on https://github.com/PostgREST/postgrest/actions/runs/14539118966/job/40793417597?pr=4024 don't show a clear improvement on HEAD over main or v12.2.9. So I'm not reproducing the results of the manual tests on https://github.com/PostgREST/postgrest/pull/3889#issuecomment-2787889413.

Locally I did get good results as [mentioned before](https://github.com/PostgREST/postgrest/pull/4023#issuecomment-2814495037), but maybe there is something wrong with https://github.com/PostgREST/postgrest/pull/4023.

steve-chavez:
Now that there's a summary report the jwt loadtest is showing some errors https://github.com/PostgREST/postgrest/actions/runs/14539562328?pr=4024#summary-40794699307

steve-chavez:
Maybe we need to reduce more noise from the JWT loadtest.

Thinking more about it, we don't need to hit the database for this. The OPTIONS requests don't do that, so it should be enough to do:

```
curl localhost:3000/authors_only -X OPTIONS -H "Authorization: Bearer <jwt>"
```

Hopefully `vegeta` supports OPTIONS requests.

steve-chavez:
No good results unfortunately https://github.com/PostgREST/postgrest/actions/runs/14542544456?pr=4024

The jwt loadtest shows the same throughput on all the runs.

At this point I'll consider https://github.com/PostgREST/postgrest/pull/4023 a failure and will just remove since it doesn't help at all.

steve-chavez:
> Edit: And most importantly, the JWT loadtest is not testing the cache at all, because it only ever contains two entries, see https://github.com/PostgREST/postgrest/pull/4023/files#r2051457587.

My bad there. I had checked the JWTs were unique but I was shuffling around the python file (tried to inline in `loadtest.nix`) and somehow missed the line that made them unique :man_facepalming: :

```python
    payload = {
        "sub": f"user_{random.getrandbits(32)}",
```

Will retry the loadtest with the new suggestions.

PR #4026, opened 2025-04-19
v12 fix: purge JWT cache asynchronously in a separate thread

steve-chavez:
Backport of https://github.com/PostgREST/postgrest/pull/3889


steve-chavez:
I commented on https://github.com/PostgREST/postgrest/pull/3889#issuecomment-2816315580

> I'll just merge here since the manual tests proved that this improved perf. Also I'll consider this a "fix" since perf was unnecessarily impacted.

So I think it's a "fix", I believe accidental downgrade in perf is a "bug".

PR #4032, opened 2025-04-20
fix: regression with parameter charset=utf-8 in mediatype

### LINKED ISSUE #4030: Content-Type not acceptable error with "charset=utf-8"
### Environment

* PostgreSQL version: 17
* PostgREST version: v12.2.10
* Operating system: Official docker image

### Description of issue

Requests in v12.2.10 that have a charset in the Content-Type header, eg:
```
"Content-Type": "application/json; charset=utf-8"
```

Are failing with the following error:
> Content-Type not acceptable: application/json; charset=utf-8

Requests without the charset are successful, eg;
```
"Content-Type": "application/json"
```

This seems to have been introduced after v12.2.8 as the same request is successful in that version.

taimoorzaeem:
Closes #4030.

### DIFF
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -183,14 +183,14 @@ decodeMediaType mt = decodeMediaType' $ decodeLatin1 mt
 -- >>> P.parse tokenizeMediaType "" "application/vnd.pgrst.plan+text; for=\"text/xml\"; options=analyze|verbose|settings|buffers|wal"
 -- Right ("application","vnd.pgrst.plan+text",[("for","text/xml"),("options","analyze|verbose|settings|buffers|wal")])
 
+-- TODO: Improve mediatype parser as per RFC 2045 https://datatracker.ietf.org/doc/html/rfc2045#section-5.1
 tokenizeMediaType :: P.Parser (Text, Text, [(Text, Text)])
 tokenizeMediaType = do
   mainType <- P.many1 (P.alphaNum <|> P.oneOf ".*")
   P.char '/'
   subType <- P.many1 (P.alphaNum <|> P.oneOf ".*+-")
   params <- P.many pSemicolonSeparatedKeyVals
-  P.optional $ P.try $ P.spaces *> P.char ';' -- ending semicolon
-  P.eof
+  P.optional $ P.try $ P.spaces *> P.char ';' -- ending semicolon, discard input after that because it has already failed or we have hit EOF
   return (T.pack mainType, T.pack subType, params)
     where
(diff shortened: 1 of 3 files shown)


src/PostgREST/MediaType.hs
```diff
@@ -198,12 +198,14 @@ tokenizeMediaType = do
...
             P.spaces
             P.char '='
             P.spaces
             val <- P.try pQuoted <|> P.try pUnQuoted
             return (T.pack key, T.pack val)
               where
-                pUnQuoted = P.many1 (P.alphaNum <|> P.oneOf "|")
+                pUnQuoted = P.many1 (P.alphaNum <|> P.oneOf specialChars)
                 pQuoted = P.char '\"' *> P.manyTill P.anyChar (P.char '\"')
+                -- allowed chars as per RFC 2045
```
steve-chavez, line 210:
```suggestion
                -- allowed chars as per RFC 2045 https://datatracker.ietf.org/doc/html/rfc2045#section-5.1
```

test/spec/Feature/Query/CustomMediaSpec.hs
```diff
@@ -382,3 +382,26 @@ spec = describe "custom media types" $ do
...
+        [json|{"val":1}|]
+        { matchStatus = 200
+        , matchHeaders = ["Content-Type" <:> "application/json; charset=utf-8"]
+        }
+
+    it "allows special chars as specified in RFC 2045 in the key and value of parameter" $
+      request methodPost "/rpc/overloaded_default"
+                                        --        key         =     value
+                                        --   ----------------- -----------------
+        [("Content-Type", "application/json; !#$%^&*-_+/'.{}~`=!#$%^&*-_+/'.{}~`")]
```
steve-chavez, line 401: If the key values are unknown, should we respond with an error? Or should we be lenient? Anything on the RFC about this?

steve-chavez: Not that anything should be changed now, but we could leave it as TODO or issue.. being too strict might break other clients.

taimoorzaeem: > Or should we be lenient? Anything on the RFC about this?

RFC says:
> MIME implementations must ignore any parameters whose names they do not recognize.

Thanks for pointing the leniency out. I actually wrote it to be lenient. The actual mistake was this:
```haskell
P.eof
```
After failing and backtracking, it looked for `EOF` and then failed. So, removing this one line should fix this issue entirely.

steve-chavez (approved):

PR #4033, opened 2025-04-20
fix: filter on unselected columns in a table-valued function

### LINKED ISSUE #3965: Cannot use filters that are not selected in Table-Valued Functions
> [...] when using an rpc i cant filter for fields not in the select. I wanted to build an csv out of an filtered rpc, but i only want to include cols which are not part of a filter. 

 _Originally posted by @pebosi in [#3627](https://github.com/PostgREST/postgrest/issues/3627#issuecomment-2742476686)_

For example:

```bash
curl 'localhost:3000/rpc/getallprojects?select=id,client_id&name=like.OSX'
```
```json
{"code":"42703","details":null,"hint":null,"message":"column projects.name does not exist"} 
```

taimoorzaeem:
Closes #3965.

### DIFF
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -1019,6 +1019,7 @@ callPlan proc ApiRequest{} paramKeys args readReq = FunctionCall {
 , funCScalar = funcReturnsScalar proc
 , funCSetOfScalar = funcReturnsSetOfScalar proc
 , funCRetCompositeAlias = funcReturnsCompositeAlias proc
+, funCFilterFields = getFilterFieldNames readReq
 , funCReturning = inferColsEmbedNeeds readReq []
 }
   where
@@ -1027,6 +1028,22 @@ callPlan proc ApiRequest{} paramKeys args readReq = FunctionCall {
       [prm] | ppName prm == mempty -> OnePosParam prm
             | otherwise            -> KeyParams $ specifiedParams [prm]
       prms  -> KeyParams $ specifiedParams prms
+
+-- | Get filter fields/column names from read plan
+getFilterFieldNames :: ReadPlanTree -> Set FieldName
+getFilterFieldNames rpt = S.fromList $ foldr (\rp names -> names <> rpToFieldNames rp) [] rpt
+  where
+    rpToFieldNames :: ReadPlan -> [FieldName]
+    rpToFieldNames = logicTreesToFieldName . ReadPlan.where_
+
+    logicTreesToFieldName :: [CoercibleLogicTree] -> [FieldName]
+    logicTreesToFieldName = concatMap coLogicTreeToFieldNames
+
+    coLogicTreeToFieldNames :: CoercibleLogicTree -> [FieldName]
+    coLogicTreeToFieldNames = \case
+      CoercibleStmnt (CoercibleFilter{field=CoercibleField{cfName}}) -> [cfName]
+      CoercibleStmnt (CoercibleFilterNullEmbed _ cfName) -> [cfName] -- needs test coverage
+      CoercibleExpr _ _ clts -> concatMap coLogicTreeToFieldNames clts
 
 -- | Infers the columns needed for an embed to be successful after a mutation or a function call.
 inferColsEmbedNeeds :: ReadPlanTree -> [FieldName] -> S.Set FieldName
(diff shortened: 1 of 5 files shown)


src/PostgREST/Plan.hs
```diff
@@ -1028,6 +1029,22 @@ callPlan proc ApiRequest{} paramKeys args readReq = FunctionCall {
...
+    rpToFieldNames = logicTreesToFieldName . ReadPlan.where_
+
+    logicTreesToFieldName :: [CoercibleLogicTree] -> [FieldName]
+    logicTreesToFieldName = concatMap coLogicTreeToFieldNames
+
+    coLogicTreeToFieldNames :: CoercibleLogicTree -> [FieldName]
+    coLogicTreeToFieldNames = \case
+      CoercibleStmnt (CoercibleFilter{field=CoercibleField{cfName}}) -> [cfName]
+      CoercibleStmnt (CoercibleFilterNullEmbed _ cfName) -> [cfName]
+      CoercibleExpr _ _ clts -> concatMap coLogicTreeToFieldNames clts
```
taimoorzaeem, line 1046: @laurenceisla I can't think of a test to `code-cov` these two lines? Any ideas?

laurenceisla: If I understand correctly, `CoercibleStmnt (CoercibleFilterNullEmbed ...)` is for `IS NULL` filters done to the embedded resource, like: `/rpc/getallprojects?select=client()&client=not.is.null`. The `CoercibleExpr` is for logical filters like: `/rpc/getallprojects?or=(id.eq.1,id.eq.2)` (just examples to illustrate, not sure if relevant to this PR).

src/PostgREST/Plan.hs
```diff
@@ -1028,6 +1029,22 @@ callPlan proc ApiRequest{} paramKeys args readReq = FunctionCall {
...
+    rpToFieldNames :: ReadPlan -> [FieldName]
+    rpToFieldNames = logicTreesToFieldName . ReadPlan.where_
+
+    logicTreesToFieldName :: [CoercibleLogicTree] -> [FieldName]
+    logicTreesToFieldName = concatMap coLogicTreeToFieldNames
+
+    coLogicTreeToFieldNames :: CoercibleLogicTree -> [FieldName]
+    coLogicTreeToFieldNames = \case
+      CoercibleStmnt (CoercibleFilter{field=CoercibleField{cfName}}) -> [cfName]
+      CoercibleStmnt (CoercibleFilterNullEmbed _ cfName) -> [cfName]
```
taimoorzaeem, line 1045: I tried `/rpc/getallprojects?select=id,clients(id)&clients.name=not.is.null` to cover this line but it didn't work. Hmm :thinking:.

laurenceisla: Ah, the filter should be done to the embedding itself, not to a column of the embedding, i.e. replace `clients.name=not.is.null` with `clients=not.is.null`.

laurenceisla: You may need to adapt it to your PR needs, so instead of replacing it, you could add another one (the current one seems relevant to your changes).

taimoorzaeem:
I think we should also include an example for this in the docs, maybe in [horizontal-filtering](https://docs.postgrest.org/en/v12/references/api/tables_views.html#horizontal-filtering) or [embedded-filters](https://docs.postgrest.org/en/v12/references/api/resource_embedding.html#embedded-filters)?

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -210,10 +211,17 @@ callPlanToQuery (FunctionCall qi params arguments returnsScalar returnsSetOfScal
...
-    returnedColumns
-      | null returnings = "*"
-      | otherwise       = intercalateSnippet ", " (pgFmtColumn (QualifiedIdentifier mempty "pgrst_call") <$> returnings)
+    returnedColumns = case returnings of
+      []    -> "*"
+      ["*"] -> pgFmtColumn (QualifiedIdentifier mempty "pgrst_call") "*"
+      _     -> intercalateSnippet ", " (pgFmtColumn (QualifiedIdentifier mempty "pgrst_call") <$> returnedColumns')
+        where
+          returnedColumns' = removeDups $ returnings ++ filterFields
+          removeDups = L.nub -- function to remove duplicates from list (possibly impact performance for very large lists)
```
steve-chavez, line 224: Since these fields are meant to be unique. Couldn't we use `Data.Set` instead to avoid this deduplication?

So `funCReturning` (returnings here) could be a Set instead of a List, then here we would just do an union with `S.fromList filterFields`?

That might be cleaner.

Additionally, not sure if large lists are a concern here since we don't expect so much fields thanks to the URL length limit.

taimoorzaeem: Yes, I think it would be cleaner to this in a refactor first to change the `returnings` to a `Set`.

steve-chavez (approved):

Issue #4039, opened 2025-04-22
Issue when using `.aud[0]` inside `jwt-role-claim-key`


laurenceisla:
## Problem

PostgREST version: v12.2.8

A user cannot modify the JWT, they want to use the `aud` as `role`. The format of the `aud` sent in the JWT is:

```javascript
{
  "aud": ["some_value"],
  ...
}
```

So they set `pgrst.jwt_role_claim_key = '.aud[0]'`. But this doesn't work and a request with JWT defaults to the `anonymous` role. However, setting `pgrst.jwt_role_claim_key = '.aud'` works and sets "some_value" as role.

Now, if the `aud` has more than one key then the original setup works:

```javascript
{
  "aud": ["some_value", "other_value"],
  ...
}
```

That is, `pgrst.jwt_role_claim_key = '.aud[0]'` sets "some_value" as role, and `pgrst.jwt_role_claim_key = '.aud'` no longer works.

We need to determine if we should support operating the `aud` claim on `jwt-role-claim-key`. 

If we don't, I think we should fail if the config contains the `aud` claim.

steve-chavez:
> PostgREST version: v12.2.8

@laurenceisla This doesn't happen on `main`? Maybe we should add tests confirming `aud` works in `role-claim-key` if so.

laurenceisla:
> According to this, the v13 behavior should be to always return this as an array, even if the original token had a bare string.
> 
> Thus, I'd say this should work with v13 and `.aud[0]`.

Correct, can confirm. 

> We need to determine if we should support operating the aud claim on jwt-role-claim-key.

Yes, I don't think we should bother in restricting this. Not sure if a test would be good to add just to know we allow it, otherwise this issue should be closed.

steve-chavez:
> Not sure if a test would be good to add just to know we allow it

Yes, tests are definitely needed to ensure no regressions happen.

PR #4043, opened 2025-04-24
fix: not handling array values in `aud` claim correctly

laurenceisla:
Fixes regression and adds missing tests.

### DIFF
--- a/docs/references/auth.rst
+++ b/docs/references/auth.rst
@@ -166,10 +166,27 @@ PostgREST honors the following `JWT claims <https://datatracker.ietf.org/doc/htm
 - ``exp`` Expiration Time
 - ``iat`` Issued At
 - ``nbf`` Not Before
-- ``aud`` Audience, see :ref:`jwt-aud`
+- ``aud`` :ref:`Audience <jwt_aud_validation>`
 
 .. note::
-  PostgREST allows for a 30-second clock skew when validating the ``exp`` and ``iat`` claims. In other words, it gives an extra 30 seconds before the token is rejected if there is a slight discrepancy in the timestamps.
+  PostgREST allows for a 30-second clock skew when validating the ``exp``, ``iat`` and ``nbf`` claims.
+  In other words, it gives an extra 30 seconds before the token is rejected if there is a slight discrepancy in the timestamps.
+
+.. _jwt_aud_validation:
+
+JWT ``aud`` Claim Validation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+PostgREST has built-in validation of the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_.
+It works this way:
+
+- If :ref:`jwt-aud` is not set (the default), PostgREST identifies with all audiences and allows the JWT for any ``aud`` claim.
+- If :ref:`jwt-aud` is set to a specific audience, PostgREST will check if this audience is present in the ``aud`` claim:
+
+  + If the ``aud`` value is a JSON string, it will match it to the :ref:`jwt-aud`.
+  + If the ``aud`` value is a JSON array of strings, it will search every element for a match.
+  + If the match fails or if the ``aud`` value is not a string or array of strings, then the token will be rejected with a :ref:`401 Unauthorized <pgrst303>` error.
+  + If the ``aud`` key **is not present** or if its value is ``null`` or ``[]``, PostgREST will interpret this token as allowed for all audiences and will complete the request.
 
 .. _jwt_role_claim_key_extract:
 
--- a/docs/references/configuration.rst
+++ b/docs/references/configuration.rst
@@ -599,11 +599,7 @@ jwt-aud
   **In-Database** pgrst.jwt_aud
   =============== =================================
 
-  Specifies the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_. If this claim is present in the client provided JWT then you must set this to the same value as in the JWT, otherwise verifying the JWT will fail.
-
-  .. warning::
-
-     Using this setting will only reject tokens with a different audience claim. Tokens **without** audience claim will still be accepted.
+    Specifies an audience for the JWT ``aud`` claim. See :ref:`jwt_aud_validation`.
 
 .. _jwt-role-claim-key:
 
--- a/test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
+++ b/test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
@@ -3,43 +3,238 @@ module Feature.Auth.AudienceJwtSecretSpec where
 import Network.Wai (Application)
 
 import Network.HTTP.Types
+import Protolude           hiding (get)
+import SpecHelper
 import Test.Hspec
 import Test.Hspec.Wai
-
-import Protolude  hiding (get)
-import SpecHelper
+import Test.Hspec.Wai.JSON
 
 spec :: SpecWith ((), Application)
-spec = describe "test handling of aud claims in JWT" $ do
+spec = describe "test handling of aud claims in JWT when the jwt-aud config is set" $ do
 
-  -- this test will stop working 9999999999s after the UNIX EPOCH
-  it "succeeds with jwt token containing with an audience claim" $ do
-    {-  This is the decoded contents of authHeaderJWT
+  context "when the audience claim is a string" $ do
+    -- this test will stop working 9999999999s after the UNIX EPOCH
+    it "succeeds when the audience claim matches" $ do
+      let jwtPayload =
+            [json|{
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": "youraudience"
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith` 200
 
-        {
-          "exp": 9999999999,
-          "role": "postgrest_test_author",
-          "id": "jdoe",
-          "aud": "youraudience"
-        }
+    it "fails when the audience claim does not match" $ do
+      let jwtPayload = [json|
+            {
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": "notyouraudience"
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith`
+          [json|{"code":"PGRST303","details":null,"hint":null,"message":"JWT not in audience"}|]
+          { matchStatus = 401 }
 
-    -}
-    let auth = authHeaderJWT "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjk5OTk5OTk5OTksInJvbGUiOiJwb3N0Z3Jlc3RfdGVzdF9hdXRob3IiLCJpZCI6Impkb2UiLCJhdWQiOiJ5b3VyYXVkaWVuY2UifQ.fJ4tLKSmolWGWehWN20qiU9dMO-WY0RI2VvacL7-ZGo"
-    request methodGet "/authors_only" [auth] ""
-      `shouldRespondWith` 200
+    it "fails when the audience claim is empty" $ do
+      let jwtPayload = [json|
+            {
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": ""
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith`
+          [json|{"code":"PGRST303","details":null,"hint":null,"message":"JWT not in audience"}|]
+          { matchStatus = 401 }
 
-  it "succeeds with jwt token that does not contain an audience claim" $ do
-    {- This is the decoded contents of authHeaderJWT
+  context "when the audience claim is an array of strings" $ do
+    it "succeeds when the audience claim has 1 element and it matches" $ do
+      let jwtPayload = [json|
+            {
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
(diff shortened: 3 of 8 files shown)


test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
```diff
@@ -5,41 +5,127 @@ import Network.Wai (Application)
...
+
+          {
+            "exp": 9999999999,
+            "role": "postgrest_test_author",
+            "id": "jdoe",
+            "aud": "youraudience"
+          }
+
+      -}
+      let auth = authHeaderJWT "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjk5OTk5OTk5OTksInJvbGUiOiJwb3N0Z3Jlc3RfdGVzdF9hdXRob3IiLCJpZCI6Impkb2UiLCJhdWQiOiJ5b3VyYXVkaWVuY2UifQ.fJ4tLKSmolWGWehWN20qiU9dMO-WY0RI2VvacL7-ZGo"
```
steve-chavez, line 29: How about having a util function in tests for generating the JWT?

laurenceisla: K, will add one.

steve-chavez:
Should we use a special prefix for these cases? Before I've used the word "correct":

- https://github.com/PostgREST/postgrest/commit/3e53796120513d5fa560c59f1f4d393d957f108e
- https://github.com/PostgREST/postgrest/commit/1a8b6972a8e940bb59ec9240daf6c57fff9b5c09

Maybe that would be good to not cause confusion later on.

laurenceisla:
> Maybe that would be good to not cause confusion later on.

Yeah that's better, I used the same syntax as your examples. Not sure about using a prefix like "correct: ..."  but maybe we should.

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
laurenceisla: Done, added a simple JWT generation in the SpecHelper. This could be re-used for other tests that currently hardcode the JWT too... in a future refactor.

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
```diff
@@ -3,43 +3,242 @@ module Feature.Auth.AudienceJwtSecretSpec where
...
+      let jwtPayload = [json|
+            {
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": []
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith` 200
```
laurenceisla, line 216: Funnily enough, this test fails for v12 (`jwt-aud` not set and an empty array `aud`). Once the array had elements, it worked. (Just to mention, this should behave the same as discussed above).

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
```diff
@@ -3,43 +3,242 @@ module Feature.Auth.AudienceJwtSecretSpec where
...
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": []
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith`
+          [json|{"code":"PGRST303","details":null,"hint":null,"message":"JWT not in audience"}|]
+          { matchStatus = 401 }
```
laurenceisla, line 123: This also fails in `v12`, but should we keep failing here? In this case `jwt-aud = youraudience` and the `aud` claim has an empty array `[]`. The logic is that the `aud` is "present" and our audience does not match, so we reject it. Or could we interpret it as "there's no audience" and allow the JWT? The RFC seems to be flexible here, I think:

> The interpretation of audience values is generally application specific.

Unless this means that **once** the `aud` claim is validated we can chose what to do with it?

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
```diff
@@ -3,43 +3,242 @@ module Feature.Auth.AudienceJwtSecretSpec where
...
+      let jwtPayload = [json|
+            {
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": { "youraudience": true }
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith` 200
```
laurenceisla, line 136: This passes here, but it returns an error in v12:

```json
{
  "code":"PGRST301",
  "details":null,
  "hint":null,
  "message":"JWTClaimsSetDecodeError \"Error in $.aud: parsing StringOrURI failed, expected String, but encountered Object\""
}
```

Not sure if we should go back to returning an error or keep ignoring the `aud`. The latter needs documentation.

laurenceisla (commented):
@steve-chavez @wolfgangwalther I have some doubts here we should agree on before merging. Some things are not clear enough for me in v12 due to lack of tests and docs.

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
steve-chavez: Does the RFC say anything about an object?

I think we shoud fail and maintain the v12 behavior.

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
```diff
@@ -3,43 +3,242 @@ module Feature.Auth.AudienceJwtSecretSpec where
...
+      let jwtPayload =
+            [json|{
+              "exp": 9999999999,
+              "role": "postgrest_test_author",
+              "id": "jdoe",
+              "aud": "youraudience"
+            }|]
+          auth = authHeaderJWT $ generateJWT jwtPayload
+      request methodGet "/authors_only" [auth] ""
+        `shouldRespondWith` 200
```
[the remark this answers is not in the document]
laurenceisla: > I'd phrase it differently: "If jwt-aud is not set, we identify with all audiences.".

That's much better! And yes, that would follow it.

taimoorzaeem:
This would close #4039, right?

laurenceisla:
> This would close #4039, right?

No, this is a different issue. That one is already fixed actually, but we still need to add tests to verify that it's working correctly, then we can close it.

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
laurenceisla: >  I agree, we should return an error.

This is also valid for other registered claims, so I added the error handling for those too in a different commit.

test/spec/Feature/Auth/AudienceJwtSecretSpec.hs
steve-chavez: @laurenceisla Since we have:

> The interpretation of audience values is generally application specific.

How about documenting our interpretation? I think more docs about `aud` would fit in https://docs.postgrest.org/en/v12/references/auth.html#jwt-claims-validation

laurenceisla: It looks like most of the info for `aud` is [in the configuration page](https://postgrest--4043.org.readthedocs.build/en/4043/references/configuration.html#jwt-aud). I added the interpretation there.

docs/references/configuration.rst
```diff
@@ -599,11 +599,13 @@ jwt-aud
   **In-Database** pgrst.jwt_aud
   =============== =================================
 
-  Specifies the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_. If this claim is present in the client provided JWT then you must set this to the same value as in the JWT, otherwise verifying the JWT will fail.
+  Specifies the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_.
+  If this claim is present in the client provided JWT then you must set this to the same value as in the JWT, otherwise verifying the JWT will fail.
```
steve-chavez, line 603: >   If this claim is present in the client provided JWT then you must set this to the same value as in the JWT

Can you clarify this? Don't quite get it.

laurenceisla: Yeah, I attempted an improvement there. Thanks to this I noticed that I was forgetting to add what we mentioned here: https://github.com/PostgREST/postgrest/pull/4043#discussion_r2061257712.

docs/references/configuration.rst
```diff
@@ -599,11 +599,17 @@ jwt-aud
   **In-Database** pgrst.jwt_aud
   =============== =================================
 
-  Specifies the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_. If this claim is present in the client provided JWT then you must set this to the same value as in the JWT, otherwise verifying the JWT will fail.
+  Specifies the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_.
+
+  By default, PostgREST identifies with all audiences and allows the JWT for any ``aud`` claim.
+
+  Once this config is set, PostgREST will compare this value to the one in the JWT ``aud`` claim.
+  If this comparison fails, then the token is rejected.
```
steve-chavez, line 607: This doesn't mention the case of aud arrays (` "aud": ["notyouraudience", "youraudience", "anotheraudience"]`), where we'll look for one matching element. I think that should be mentioned as it's "our interpretation"?

docs/references/configuration.rst
```diff
@@ -599,11 +599,17 @@ jwt-aud
...
 
   .. warning::
 
-     Using this setting will only reject tokens with a different audience claim. Tokens **without** audience claim will still be accepted.
+     Using this setting will only reject tokens with a different audience claim.
+     Tokens will still be accepted if the ``aud`` claim **is not present** or if its value is ``null`` or ``[]``.
```
steve-chavez, line 612: Instead of a warning note, I think we should structure the `aud` rules as bulletpoints. Also taking into account the above feedback. That'll be more complete and explicit.

steve-chavez: Also, I think these rules would make more sense in a subsection called `"aud validation"` in `auth.rst`, not on the config page.

docs/references/configuration.rst
```diff
@@ -599,11 +599,8 @@ jwt-aud
   **In-Database** pgrst.jwt_aud
   =============== =================================
 
-  Specifies the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_. If this claim is present in the client provided JWT then you must set this to the same value as in the JWT, otherwise verifying the JWT will fail.
-
-  .. warning::
-
-     Using this setting will only reject tokens with a different audience claim. Tokens **without** audience claim will still be accepted.
+  Specifies an audience for the `JWT aud claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_.
+  See :ref:`jwt_aud_validation`.
```
steve-chavez, line 603: Let's deduplicate links:
```suggestion
  Specifies an audience for the ``aud`` JWT claim. :ref:`jwt_aud_validation`.
```

docs/references/auth.rst
```diff
@@ -166,10 +166,26 @@ PostgREST honors the following `JWT claims <https://datatracker.ietf.org/doc/htm
...
+  PostgREST allows for a 30-second clock skew when validating the ``exp``, ``iat`` and ``nbf`` claims.
+  In other words, it gives an extra 30 seconds before the token is rejected if there is a slight discrepancy in the timestamps.
+
+.. _jwt_aud_validation:
+
+JWT ``aud`` Claim Validation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+PostgREST has a built in validation of the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_.
+It works this way:
```
steve-chavez, line 181:
```suggestion
PostgREST has built in validation for the `JWT audience claim <https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3>`_. It works this way:
```

docs/references/auth.rst
```diff
@@ -166,10 +166,26 @@ PostgREST honors the following `JWT claims <https://datatracker.ietf.org/doc/htm
...
+
+- By default, PostgREST identifies with all audiences and allows the JWT for any ``aud`` claim.
+- To change this, you can set a specific audience in the :ref:`jwt-aud` configuration.
+  PostgREST will then check if this audience is present in the ``aud`` claim list (or if it's the same value if ``aud`` is a string).
+  If this fails, then it rejects the token and responds with a :ref:`401 Unauthorized <pgrst303>` error.
+- Note that the token will not be rejected if the ``aud`` claim **is not present** or if its value is ``null`` or ``[]``.
+  PostgREST will interpret this token as allowed for all audiences and will complete the request.
```
steve-chavez, line 188: @laurenceisla This seems more complete and understandable to me. WDYT?

```suggestion
- If :ref:`jwt-aud` is not set (the default), PostgREST identifies with all audiences and allows the JWT for any ``aud`` claim.
- If :ref:`jwt-aud` is set to a specific audience, PostgREST will check if this audience is present in the ``aud`` claim.
  + If the ``aud`` value is a JSON string, it will match it to the set :ref:`jwt-aud`.
  + If the ``aud`` value is a JSON array, it will search every element for a match.
  + If a match is not found or if above the ``aud`` value is a JSON object or number, then the token will be rejected with a :ref:`401 Unauthorized <pgrst303>` error.
  + If the ``aud`` key **is not present** or if its value is ``null`` or ``[]``, PostgREST will interpret this token as allowed for all audiences and will complete the request.
```

laurenceisla: > ...or if the ``aud`` value is a JSON object or number, then the token will be rejected ...

Instead of listing invalid values, maybe just mention that "if it's not an string or array of strings, then it will return an error". Or just "if it's an invalid value, then...".

steve-chavez: First one looks good. Also we should add that the JSON array is composed of strings. I forgot that above.

steve-chavez (approved):

Issue #4061, opened 2025-05-05
Enable log buffering


steve-chavez:
## Problem

Currently there's no log buffering, as noted in [docs](https://docs.postgrest.org/en/v12/references/configuration.html#log-level). This means that log levels higher than error cannot be enabled without reduced RPS.

Additionally, this might also cause other issues with the default log-level=error (default). For example on pool timeouts, the RPS will also suffer.

## Solution

Enable log buffering, currently it's all set to line buffering on:

https://github.com/PostgREST/postgrest/blob/bf79766a9bdc32333a4de4db2698790f09732ffe/main/Main.hs#L15-L21

This concern should be centralized on the `Logger.hs` module and possibly receive a config option as input.

## Notes

The logger middleware obfuscates this logic. Ideally we should reuse our Observation module as mentioned on:

https://github.com/PostgREST/postgrest/blob/bf79766a9bdc32333a4de4db2698790f09732ffe/src/PostgREST/Logger.hs#L58-L70

steve-chavez:
One way to test buffering is working as intended would be to enable log-level=info for the loadtests, in that case throughput should be more or less maintained.

steve-chavez:
I've noticed that it's already possible to do this, first put False on:
https://github.com/PostgREST/postgrest/blob/effcc080e58e0e251d718a27e4445e5a44e4a78a/src/PostgREST/Logger.hs#L78

Then change
https://github.com/PostgREST/postgrest/blob/effcc080e58e0e251d718a27e4445e5a44e4a78a/main/Main.hs#L19

To ` hSetBuffering stdout $ BlockBuffering Nothing` and then doing `postgrest-parallel-curl 100 localhost:3000/projects` should confirm that only after many requests the logs will appear.

The `hSetBuffering` API is a bit weird, when putting Nothing the buffer size is calculated automatically. When putting `Just <number>` that's supposed to be the buffer size but I don't quite get if the unit is bytes or something else.

We should not do buffering for `stderr` since those logs are important and must be known fast, also we already guard flooding with debouncing for several of those. 

---

I'm wondering if we should apply buffering for `stdout` automatically when log levels are higher than error, instead of making it configurable.

taimoorzaeem:
Hm, I think one disadvantage of `BlockBuffering` is that some log entries might never get flushed if the server crashes before that. Example: `SIGKILL`.

steve-chavez:
We have unix signal handlers so in theory we can flush the logs on any signal
[...]

PR #4062, opened 2025-05-05
refactor: no middleware for response logging

steve-chavez:
Benefits:

- Removes `unsafePerformIO` from Logger
- Reuses our cached time getter (`stateGetZTime`) instead of having another unnecessary one. See https://github.com/yesodweb/wai//blob/d50e1184600631a114e3d2ad119abdf0ef08834b/wai-extra/Network/Wai/Middleware/RequestLogger/Internal.hs#L22-L37.
- Reuses Observation module, simplified logic
- Will help with log buffering #4061

TODO:

- [ ] Currently we vendor `apacheLogStr`, but might be possible to expose this function upstream. If not, it's not a big problem since the function is small.
- [ ] Time in logging differs:
```
  Old: 127.0.0.1 - postgrest_test_anonymous [05/May/2025:01:39:57 -0500] "GET /projects HTTP/1.1" 200 212 "" "curl/7.81.0"
  New: 127.0.0.1 - postgrest_test_anonymous [2025-05-05 01:39:14.776911752 -05] "GET /projects HTTP/1.1" 200 212 "" "curl/7.81.0"
```
  + Otherwise the rest of the format is the same.
- [ ] Auth error logging relies on old logger middleware
  + We might need to put `observer` in Auth.hs
  + Or otherwise try to combine this with #4059

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -67,6 +67,7 @@ library
                       PostgREST.Error
                       PostgREST.Listener
                       PostgREST.Logger
+                      PostgREST.Logger.Apache
                       PostgREST.MediaType
                       PostgREST.Metrics
                       PostgREST.Network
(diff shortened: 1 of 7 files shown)


taimoorzaeem:
Looks good :+1:. I was thinking of adding some commits to this PR to quickly wrap this one up, but I am not sure how to do that.

steve-chavez:
> I was thinking of adding some commits to this PR to quickly wrap this one up

@taimoorzaeem That would be great! Maybe you can open another PR based on this one?

Issue #3932, opened 2025-03-05
Export some functions as FFI


vadim2404:
## Problem

Hey, I'm Vadim from Neon. And I have one tricky idea in mind which I want to test. I want to embed Postgrest into our proxy component. What is actually needed is if it is possible to export some functions from your codebase as FFIs and compile a library that I can link in my rust code. 

I've already tried some simple Haskell functions, which I can call from the Rust. 

I personally need the thing that converts HTTP requests into SQL. I can build a request object in Rust and pass it to Postgrest, and I would love to receive a response.

@steve-chavez wdyt about it?

steve-chavez:
Hey @vadim2404,

> I personally need the thing that converts HTTP requests into SQL.

The need for exporting the `HTTP -> SQL` module has also come up before in the context of WASM (https://github.com/PostgREST/postgrest/issues/2452, essentially to use postgREST with pglite), so I think it's a good idea.

I believe we would need to export our [Plan.hs](https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Plan.hs) (internal AST) and [Query.hs](https://github.com/PostgREST/postgrest/blob/main/src/PostgREST/Query.hs) (generates SQL) modules as Haskell libraries to support this cleanly.

taimoorzaeem:
@vadim2404 As a haskell developer who hasn't worked with much FFIs yet, I would love to collaborate on this once we have a workable design for this feature.

taimoorzaeem:
@vadim2404 I think a public discussion would be much better for the PostgREST community.

vadim2404:
it's a valid point. 

how I see it. What I expect from my app: I can pass you the HTTP method, query params, body, and headers (if it makes any sense), and expect to receive a valid SQL. I don't want to let you know about real database connection or schema, I want to keep it generic. 

In terms of FFI, we can define a struct Request and function that accepts a Request object and returns a C-string. Something like this

```c
// Define the Request struct
typedef struct {
    char *method;  // HTTP method (GET, POST, etc.)

    char **query_params;  // Array of key-value pairs (each entry is array[2])
    uint32_t query_count;  // Number of query params

    uint8_t *body;  // Raw body data
    uint32_t body_length;  // Length of the body

    char **headers;  // Array of key-value pairs (each entry is array[2])
    uint32_t header_count;  // Number of headers
} Request;

char* request_to_sql(Request *req);
```

[...]

Issue #2506, opened 2022-10-11
add query tags


steve-chavez:
Query tags are comments on the SQL query(`SELECT ... FROM tbl /*key=value/*`) , they're useful for getting context on the database logs.

References:

- https://cloud.google.com/blog/topics/developers-practitioners/enable-query-tagging-sqlcommenter-understand-application-impact-database-performance
- https://docs.vividcortex.com/general-reference/query-tags/
- https://hasura.io/docs/latest/observability/query-tags/

We could include

- A `request_id`(request_id=...). It could be generated by us or configured to be obtained from a header since a proxy already adds a request id: [nginx has request_id](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_id), [cloudflare has cf-ray](https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#cf-ray), etc.
- The HTTP method(request_method=GET)
- The path (request_path=tbl)

steve-chavez:
Just found out that there is a standard HTTP header for doing the above: `traceparent`.

References:

- https://www.w3.org/TR/trace-context/#traceparent-header
- https://uptrace.dev/opentelemetry/opentelemetry-traceparent.html

I found it on the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#traceparent), which uses an [X-Opaque-Id](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#x-opaque-id) for the same purpose.

Also, the [tracestate](https://www.w3.org/TR/trace-context/#tracestate-header) header allows key value pairs for extra vendor information.

We could include this information on the query tags.

---

One problem with traceparent is that it has its own format

```
traceparent: 00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01

The traceparent header uses the version-trace_id-parent_id-trace_flags format where:

version is always 00.
trace_id is a hex-encoded trace id.
span_id is a hex-encoded span id.
trace_flags is a hex-encoded 8-bit field that contains tracing flags such as sampling, trace level, etc.
```

Though `trace_id` is 16 bytes and matches the length of nginx's `request_id` and cloudflare `cf-ray`, those systems would still have to send a modified header for us to consume.

It also requires parsing and validating the header value as mentioned on [the docs](https://www.w3.org/TR/trace-context/#processing-model-for-working-with-trace-context).

[...]

Issue #2452, opened 2022-08-28
PostgREST in WebAssembly ?


wolfgangwalther:
### Discussed in https://github.com/PostgREST/postgrest/discussions/2353

<div type='discussions-op-text'>

<sup>Originally posted by **nileshtrivedi** June 28, 2022</sup>
Has anyone tried porting PostgREST to WASM binary format? This opens the possibility of deploying PostgREST on edge cloud infra like Deno.com - which can be a lot more convenient and cost-effective.

I'm not familiar with Haskell ecosystem. I couldn't find any discussions about WASM in [Stack](https://github.com/commercialhaskell/stack/issues) repo.</div>

steve-chavez:
Related news https://www.tweag.io/blog/2024-11-21-ghc-wasm-th-ghci/

develop7:
Slightly less related https://github.com/electric-sql/pglite

PR #4064, opened 2025-05-06
refactor: [WIP] remove auth and logging middleware

taimoorzaeem:
Combines #4059 and #4062.

- Removes `vault` library (no longer needed).
- [x] Refactor
- [ ] Code cleanup

The log format remains the same except on error cases we wouldn't be able to get the user/role.
```
Old: 127.0.0.1 - postgrest_test_anonymous [05/May/2025:01:39:57 -0500] "GET /projects HTTP/1.1" 200 212 "" "curl/7.81.0"
New: 127.0.0.1 - postgrest_test_anonymous [05/May/2025:01:39:57 -0500] "GET /projects HTTP/1.1" 200 212 "" "curl/7.81.0"

Old: 127.0.0.1 - postgrest_test_anonymous [05/May/2025:01:39:57 -0500] "GET /unknown HTTP/1.1" 404 212 "" "curl/7.81.0"
New: 127.0.0.1 - - [05/May/2025:01:39:57 -0500] "GET /unknown HTTP/1.1" 404 212 "" "curl/7.81.0"
```

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -67,6 +67,7 @@ library
                       PostgREST.Error
                       PostgREST.Listener
                       PostgREST.Logger
+                      PostgREST.Logger.Apache
                       PostgREST.MediaType
                       PostgREST.Metrics
                       PostgREST.Network
@@ -109,6 +110,7 @@ library
                     , directory                 >= 1.2.6 && < 1.4
                     , either                    >= 4.4.1 && < 5.1
                     , extra                     >= 1.7.0 && < 2.0
+                    , fast-logger               >= 3.2.0 && < 3.3
                     , fuzzyset                  >= 0.2.4 && < 0.3
                     , hasql                     >= 1.6.1.1 && < 1.7
                     , hasql-dynamic-statements  >= 0.3.1 && < 0.4
@@ -142,7 +144,6 @@ library
                     , timeit                    >= 2.0 && < 2.1
                     , unordered-containers      >= 0.2.8 && < 0.3
                     , unix-compat               >= 0.5.4 && < 0.8
-                    , vault                     >= 0.3.1.5 && < 0.4
                     , vector                    >= 0.11 && < 0.14
                     , wai                       >= 3.2.1 && < 3.3
                     , wai-cors                  >= 0.2.5 && < 0.3
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -26,6 +26,7 @@ import Network.Wai.Handler.Warp (defaultSettings, setHost, setPort,
 import qualified Data.Text.Encoding       as T
 import qualified Network.Wai              as Wai
 import qualified Network.Wai.Handler.Warp as Warp
+import qualified Network.Wai.Header       as Wai
 
 import qualified PostgREST.Admin      as Admin
 import qualified PostgREST.ApiRequest as ApiRequest
@@ -34,7 +35,6 @@ import qualified PostgREST.Auth       as Auth
 import qualified PostgREST.Cors       as Cors
 import qualified PostgREST.Error      as Error
 import qualified PostgREST.Listener   as Listener
-import qualified PostgREST.Logger     as Logger
 import qualified PostgREST.Plan       as Plan
 import qualified PostgREST.Query      as Query
 import qualified PostgREST.Response   as Response
@@ -43,8 +43,7 @@ import qualified PostgREST.Unix       as Unix (installSignalHandlers)
 import PostgREST.ApiRequest           (ApiRequest (..))
 import PostgREST.AppState             (AppState)
 import PostgREST.Auth.Types           (AuthResult (..))
-import PostgREST.Config               (AppConfig (..), LogLevel (..),
-                                       LogQuery (..))
+import PostgREST.Config               (AppConfig (..), LogQuery (..))
 import PostgREST.Config.PgVersion     (PgVersion (..))
 import PostgREST.Error                (Error)
 import PostgREST.Network              (resolveHost)
@@ -75,7 +74,7 @@ run appState = do
 
   Admin.runAdmin appState (serverSettings conf)
 
-  let app = postgrest configLogLevel appState (AppState.schemaCacheLoader appState)
+  let app = postgrest appState (AppState.schemaCacheLoader appState)
 
   case configServerUnixSocket of
     Just path -> do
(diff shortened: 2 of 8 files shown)


steve-chavez:
> The log format remains the same except on error cases we wouldn't be able to get the user/role.

Hm, that's no good, why is that?

taimoorzaeem:
> > The log format remains the same except on error cases we wouldn't be able to get the user/role.
> 
> Hm, that's no good, why is that?

That's because it's a little harder to get that with the _no middleware_ approach. In the middleware approach, we added the role to the vault that was attached to the request object. Because the request object was ubiquitously available, we could get the role from anywhere. 

Now without middleware, we are doing the authentication in a linear way, it's harder to do it. I think it should be possible if I spend a bit more time on it. I'll try to reinstate it.

taimoorzaeem:
I think we should not remove these middlewares because authentication and logging are generally considered a [cross-cutting concern](https://en.wikipedia.org/wiki/Cross-cutting_concern). This essentially means they should not be implemented in _linear_ fashion and are better suited as middlewares. @wolfgangwalther WDYT?

Even if we do succeed to refactor these out, I think they would still cause issues in the future and we might end up reverting this change.

As for the original issue [ref](https://github.com/PostgREST/postgrest/pull/3507#discussion_r2072334128), I think we can solve it by just storing the parsed preferences in the request vault (just like `authResult`).

steve-chavez:
> I think we should not remove these middlewares because authentication and logging are generally considered a [cross-cutting concern](https://en.wikipedia.org/wiki/Cross-cutting_concern).

I'm fully aware of the cross-cutting concern concept and I'm old enough to have used Aspect-oriented programming (AspectJ) in Java to try and tame it. 

So far, we've been successful in implementing logging and metrics in a maintainable way, thanks to the Observation module. We don't have logging messages nor counter increments scattered everywhere. This is the best way I've seen a system implement these cross-cutting concerns.
 
> This essentially means they should not be implemented in linear fashion and are better suited as middlewares.
> Even if we do succeed to refactor these out, I think they would still cause issues in the future and we might end up reverting this change.

I think the codebase will be much more predictable, reasonable and maintainable if we do it in a linear fashion, middlewares just hide the complexity in a bad way IMO. Just the fact we have to use unsafe and partial functions plus vault to retain state.

For me the ideal end result would be having a type for request stages and then enforcing the order of the steps via GADTs. Something like:

```haskell
data Stage
  = NotCORSValidated   -- Before we've done CORS validation
  | CORSValidated      -- CORS validated
  | Authenticated      -- Auth done

data Request (s :: Stage) where...

validateCORS :: Request 'NotCORSValidated -> Either Error (Request 'CORSValidated)

authenticate :: Request 'CORSValidated -> Either Error (Request 'Authenticated)

runMainLogic :: Request 'Authenticated -> IO ()
```

I believe this is all doable and PostgREST overall will be easier to extend and maintain.

---

@taimoorzaeem That being said, I see this is too complicated now. So for https://github.com/PostgREST/postgrest/pull/3507, maybe we can use vault as you mention. Or otherwise wait until a refactor makes it easier.

steve-chavez:
>If we get rid of middlewares, do we need still need WAI? I'm not opposed to that in principle, I'm just thinking removing middlewares, but keeping WAI is kind of pointless. 

Yeah, it looks like WAI could be removed. An additional benefit of that is that we could integrate with other web servers since PostgREST would pretty much be just pure functions. We had the idea of integrating with Nginx as a module before and there's also https://github.com/PostgREST/postgrest/issues/3932. It might help with https://github.com/PostgREST/postgrest/issues/2452 too.

mkleczek:
> 
> For me the ideal end result would be having a type for request stages and then enforcing the order of the steps via GADTs. Something like:
> 
> ```haskell
> data Stage
>   = NotCORSValidated   -- Before we've done CORS validation
>   | CORSValidated      -- CORS validated
>   | Authenticated      -- Auth done
> 
> data Request (s :: Stage) where...
> 
> validateCORS :: Request 'NotCORSValidated -> Either Error (Request 'CORSValidated)
> 
> authenticate :: Request 'CORSValidated -> Either Error (Request 'Authenticated)
> 
> runMainLogic :: Request 'Authenticated -> IO ()
> ```
> 
> I believe this is all doable and PostgREST overall will be easier to extend and maintain. An observation at the end of these stages can add logging as needed.
> 

IMHO it should be a mix of both transparent middleware and using an effect system (like https://github.com/haskell-effectful/effectful).

There is no reason to couple anything with `traceHeaderMiddleware` artificially by adding a phantom kind `TraceHeaderHandled`. So having it as a middleware seems fine to me. `Cors` is similar as nothing depends on it being present or absent.

OTOH authentication middleware would be better implemented as an _effect handler_ so subsequent stages could have signature like:
`runMainLogic :: (Authenticated :> es, Error Error :> es) => Request -> Eff es ()`

The actual `Authenticated` effect would be something like a `Reader AuthResult` but specialized and provide:
`getAuth :: Authenticated :> es => Eff es AuthResult`.

The signature of authentication effect handler would be:
`authenticated :: (IOE :> es, Error Error :> es) => Eff (Authenticated : es) a -> Eff es a`

(the above are examples - the actual effects and handlers structure should be well thought out - especially in terms of error handling)

Cors and tracę header could be left as "non-discharging effect handlers" (ie. wrappers) or as middlewares.

That would achieve linearity where necessary while still allowing to have really cross cutting concerns independent from any other logic.

steve-chavez:
> There is no reason to couple anything with traceHeaderMiddleware artificially by adding a phantom kind TraceHeaderHandled. So having it as a middleware seems fine to me.

Trace header is cross-cutting but thinking about its nature it can be linearized and be part of our function pipeline:

- It essentially adds an identifier to our requests, so it should be a first step (before `Auth.hs` and `ApiRequest.hs`) that populates an attribute on a type.
- It will end up as part of the query comment (as a query tag, discussed on https://github.com/PostgREST/postgrest/issues/2506), thus will reach the last module, `Query.hs`
- The other modules  in between (like `Plan`) should also include it (it's an id, so it makes sense)

Now it will affect the `Error.hs` module too, but maybe it can be worked in a way there's no much duplication.
Granted, I haven't tried it, but might be possible.

mkleczek:
> > There is no reason to couple anything with traceHeaderMiddleware artificially by adding a phantom kind TraceHeaderHandled. So having it as a middleware seems fine to me.
> 
> Trace header is cross-cutting but thinking about its nature it can be linearized and be part of our function pipeline:
> 
> * It essentially adds an identifier to our requests, so it should be a first step (before `Auth.hs` and `ApiRequest.hs`) that populates an attribute on a type.
> * It will end up as part of the query comment (as a query tag, discussed on [`traceparent` or `X-Opaque-Id` headers to query tags #2506](https://github.com/PostgREST/postgrest/issues/2506)), thus will reach the last module, `Query.hs`
> * The other modules  in between (like `Plan`) should also include it (it's an id, so it makes sense)
> 
> Now it will affect the `Error.hs` module too, but maybe it can be worked in a way there's no much duplication. Granted, I haven't tried it, but might be possible.

Right, so **iff** other code depends on trace header one way or another then it is not transparent cross cutting concern any more.

What do you think about modeling it as effects and effect handlers discharging them though?

taimoorzaeem:
> If we get rid of middlewares, do we need still need WAI? I'm not opposed to that in principle, I'm just thinking removing middlewares, but keeping WAI is kind of pointless. So we'd need to restructure this a bit more from the ground up. But by using WAI we essentially buy into the middleware approach.

How much restructuring are we talking about here? `WAI` also includes our `Warp` server, would we remove that too? If yes, what other options do we have? Yesod?

wolfgangwalther:
> `WAI` also includes our `Warp` server, would we remove that too?

No, I don't think we want to actually change all of our stack. Warp can be used without WAI, right?

taimoorzaeem:
> [...] Warp can be used without WAI, right?

I don't think so. See [runSettingsSocket](https://hackage-content.haskell.org/package/warp-3.4.8/docs/Network-Wai-Handler-Warp.html#v:runSettingsSocket) from `WAI`, we run this with `Wai.Application` as input (postgrest server in our case). So, it seems like it can't be used without `WAI`.

steve-chavez:
>  In the middleware approach, we added the role to the vault that was attached to the request object. Because the request object was ubiquitously available, we could get the role from anywhere.

While continuing this PR on https://github.com/PostgREST/postgrest/pull/4884, I discovered the above is not true and some logs have an empty role https://github.com/PostgREST/postgrest/issues/4882

taimoorzaeem:
Closing as superseded by #4884.

PR #4072, opened 2025-05-09
test: add test when using .aud in jwt-role-claim-key

### LINKED ISSUE #4039: Issue when using `.aud[0]` inside `jwt-role-claim-key`

taimoorzaeem:
Closes #4039.


test/io/test_io.py
```diff
@@ -234,6 +234,26 @@ def test_role_claim_key(roleclaim, defaultenv):
         assert response.status_code == roleclaim["expected_status"]
 
 
+@pytest.mark.parametrize(
+    "jwtAudRoleclaim", FIXTURES["jwtAudRoleclaims"], ids=lambda claim: claim["key"]
```
laurenceisla, line 238: The tests are looking good! But I think we should add an extra identifier, otherwise we get `.aud0` and `.aud1` in the parameterized tests, which can get confusing:

```console
test/io/test_io.py::test_jwt_aud_in_role_claim_key[.aud0] PASSED                                                                                                                           [ 25%]
test/io/test_io.py::test_jwt_aud_in_role_claim_key[.aud1] PASSED                                                                                                                           [ 50%]
test/io/test_io.py::test_jwt_aud_in_role_claim_key[.aud[0]] PASSED                                                                                                                         [ 75%]
test/io/test_io.py::test_jwt_aud_in_role_claim_key[.aud[1]] PASSED
```

Maybe add the HTTP status at the end? It would now show as `.aud_200`, `.aud_401`.

test/io/fixtures.yaml
```diff
@@ -196,6 +196,24 @@ roleclaims:
           - obj_key: obj_value
     expected_status: 401 # fails because it compares an object with a string
 
+jwtAudRoleclaims:
```
laurenceisla, line 199:
```suggestion
jwtaudroleclaims:
```

Also not sure if it's convention or not, but for consistency with the other names let's keep it in lower case.

test/io/test_io.py
taimoorzaeem: Oh, that's where it's coming from. I couldn't understand where `.aud0` and `.aud1` is coming from. Thanks.

> Maybe add the HTTP status at the end? It would now show as `.aud_200`, `.aud_401`.

Sure.

laurenceisla (approved):

PR #4079, opened 2025-05-12
fix: allow db-extra-search-path to accept empty value

### LINKED ISSUE #4074: v13 regression: db-extra-search-path cannot be set to nothing — cannot use non-public schemas in latest version
### Environment

* PostgreSQL version: 13.0.0

### Description of issue

Setting db-extra-search-path = "" causes a parsing error. Not specifying db-extra-search-path at all defaults to db-extra-search-path = "public". We don't have a public schema, so PostgREST fails to start.

taimoorzaeem:
These two settings throw parse error when set to "". This commit allows them to explicitly be set to empty.
Partially addresses #4074.

- [x] Fix
- [x] Docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix jwt error returning HTTP status `400` for invalid role by @taimoorzaeem in #3601
+- Allow `db-extra-search-path` to accept empty value by @taimoorzaeem in #4074
 
 ## [13.0.0] - 2025-05-08
 
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1896,3 +1896,16 @@ def test_invalidate_jwt_cache_when_secret_changes(tmp_path, defaultenv):
         # now the request should fail because the cached token is removed
         response = postgrest.session.get("/authors_only", headers=headers)
         assert response.status_code == 401
+
+
+def test_allow_configs_to_be_set_to_empty(defaultenv):
+    'configs that are explicitly set to empty (= "<empty>") should not throw parse error'
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_EXTRA_SEARCH_PATH": "",
+    }
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.get("/projects")
+        assert response.status_code == 200
--- a/docs/references/configuration.rst
+++ b/docs/references/configuration.rst
@@ -315,6 +315,10 @@ db-extra-search-path
 
   Multiple schemas can be added in a comma-separated string, e.g. ``public, extensions``.
 
+.. important::
+  
+  We default this config to ``public`` because it is the most common schema used to install PostgreSQL extensions such as :ref:`PostGIS <ww_postgis>`. You can disable this by setting this config to ``""``.
+
 .. _db-hoisted-tx-settings:
 
 db-hoisted-tx-settings
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -256,7 +256,7 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
     <*> (fmap encodeUtf8 <$> optString "db-anon-role")
     <*> (fromMaybe "pgrst" <$> optString "db-channel")
     <*> (fromMaybe True <$> optBool "db-channel-enabled")
-    <*> (maybe ["public"] splitOnCommas <$> optString "db-extra-search-path")
+    <*> (maybe ["public"] splitOnCommasEmptyable <$> optStringEmptyable "db-extra-search-path")
     <*> (maybe defaultHoistedAllowList splitOnCommas <$> optString "db-hoisted-tx-settings")
     <*> optWithAlias (optInt "db-max-rows")
                      (optInt "max-rows")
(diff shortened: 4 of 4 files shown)


test/io/test_io.py
```diff
@@ -1897,3 +1897,17 @@ def test_invalidate_jwt_cache_when_secret_changes(tmp_path, defaultenv):
         # now the request should fail because the cached token is removed
         response = postgrest.session.get("/authors_only", headers=headers)
         assert response.status_code == 401
+
+
+def test_allow_configs_to_be_set_to_empty(defaultenv):
+    "configs that are explicitly set to empty (= " ") should not throw parse error"
```
taimoorzaeem, line 1903:
```suggestion
    "configs that are explicitly set to empty (= "") should not throw parse error"
```
I know it should be `""` here, but the style-check changes it to former. I guess I should just remove `(= " ")` part?

steve-chavez: > = ""

Maybe change it to `="<empty>"` to appease the styler?

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+- #4074, Allow `db-extra-search-path` and `db-hoisted-tx-settings` to accept empty value - @taimoorzaeem
```
steve-chavez, line 10: @taimoorzaeem Looks like this would also affect `db-schemas` right? I don't see a problem with that, if the error from the database is good.

taimoorzaeem: > How would an empty `db-schemas` work in practice?

It wouldn't. It would throw a runtime error but I think Steve meant to explicitly add a test for that failure.

steve-chavez: Just tried setting the empty schema:
```
PGRST_DB_SCHEMAS=" " postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run

12/May/2025:13:25:13 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"\" does not exist"}
```
The error message seems clear to me.

The code `3F000` is also precise:

```
Class 3F — Invalid Schema Name
3F000	invalid_schema_name
```
(See https://www.postgresql.org/docs/current/errcodes-appendix.html)

Here I'm wondering if we should consider this FATAL?

> It wouldn't. It would throw a runtime error but I think Steve meant to explicitly add a test for that failure.

If we do, then we can add a test for this and test the exitCode is 1, we already do this in some places:

https://github.com/PostgREST/postgrest/blob/e20dc60e834153843c20434b011c8ceafc043ee8/test/io/test_io.py#L74-L79

steve-chavez: No, there's no FATAL error. We just keep retrying the invalid schema.

```
PGRST_DB_SCHEMAS=" " postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run

12/May/2025:14:21:45 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"\" does not exist"}
12/May/2025:14:21:46 -0500: Attempting to reconnect to the database in 1 seconds...
12/May/2025:14:21:46 -0500: Successfully connected to PostgreSQL 16.4 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.3.0, 64-bit
12/May/2025:14:21:46 -0500: Connection Pool initialized with a maximum size of 1 connections
12/May/2025:14:21:46 -0500: Config reloaded
12/May/2025:14:21:46 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"\" does not exist"}
12/May/2025:14:21:48 -0500: Attempting to reconnect to the database in 2 seconds...
12/May/2025:14:21:48 -0500: Successfully connected to PostgreSQL 16.4 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.3.0, 64-bit
12/May/2025:14:21:48 -0500: Connection Pool initialized with a maximum size of 1 connections
12/May/2025:14:21:48 -0500: Config reloaded
12/May/2025:14:21:48 -0500: Failed to load the schema cache. {"code":"3F000","details":null,"hint":null,"message":"schema \"\" does not exist"}

```

steve-chavez: Hm, it seems inconsistent to validate the schema name in Haskell if PostgreSQL already does it more or less.

If we do want to fatal err for only the empty schema case, how about catching the `\"\"` in the error message? We already check the error message for other fatal errs. 

The advantage of this would be: if you have several deployments of PostgREST, you can query the logs for a single `3F000` code to know all cases of invalid schema name.

WDYT?

steve-chavez: > For example, for the NixOS PostgREST module that I wrote, it would be great if we eventually had a "validate config" mode, which can be run at build time of the system configuration, rather than run time. This would give us much better error reporting for NixOS.

Cool, I'm sold. Then I agree on doing the validation of `db-schemas` at the Haskell level.

steve-chavez:
We should make a patch release once this is merged, looks like it's affecting many users: https://github.com/PostgREST/postgrest/issues/4102

taimoorzaeem:
How about we default `db-extra-search-path` to the value of `db-schemas`?

taimoorzaeem:
In our tests, we set `db-extra-search-path` to `[]`. However, we don't allow it for our users. I think some users will benefit from setting this config to nothing. So, here we should allow this config to be set to empty.

https://github.com/PostgREST/postgrest/blob/da467fe3b69940a811abb63ba59486b3eb4724d9/test/spec/SpecHelper.hs#L121

taimoorzaeem:
> @taimoorzaeem How about adding details to the log as mentioned on [#4074 (comment)](https://github.com/PostgREST/postgrest/issues/4074#issuecomment-2873558649)?
> 
> I think we could just add two parameters to
> [...]

Should I do this in a separate PR? It looks like a separate improvement.

steve-chavez:
@taimoorzaeem If we're to close #4074 here ([as mentioned above](https://github.com/PostgREST/postgrest/pull/4079#issue-3056600086)) I'd say just add a commit here.

steve-chavez:
I'd say the log error is important to know which setting is being applied, so let's not do a new patch release without this new "fix" (should be one as the schema cache failure is obscured).

PR #4034, opened 2025-04-20
test: add loadtest for async purge of JWT cache

steve-chavez:
Closes #4001.

This proves the fix on https://github.com/PostgREST/postgrest/pull/3889, while running on v12.2.9 the loadtest doesn't even finish all the generated targets (50000) and has poor throughput, while HEAD will finish and show better throughput.

This has to be proved locally now since we already merged #3889 and released it, but it's very clear:

```
[nix-shell:~/Projects/postgrest]$ postgrest-loadtest-against -k jwt v12.2.9

Running loadtest on "v12.2.9"...

Preparing worktree (detached HEAD b454f29b)
Created 50000 targets in /home/stevechavez/Projects/postgrest/test/load/gen_targets.http (0.80s)
postgrest-with-postgresql-17: You can connect with: psql 'postgres:///postgres?host=/run/user/1000/postgrest/postgrest-with-postgresql-17-l0c/socket' -U postgres
postgrest-with-postgresql-17: You can tail the logs with: tail -f /run/user/1000/postgrest/postgrest-with-postgresql-17-l0c/db.log
Building postgrest (nix)... done.
Starting postgrest... done.
Requests      [total, rate, throughput]         12078, 201.27, 201.25
Duration      [total, attack, wait]             1m0s, 1m0s, 7.429ms
Latencies     [min, mean, 50, 90, 95, 99, max]  305.751µs, 4.966ms, 5.028ms, 8.792ms, 9.929ms, 11.863ms, 36.904ms
Bytes In      [total, mean]                     0, 0.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:12078  
Error Set:

Done running on "v12.2.9".


Running loadtest on HEAD...

Created 50000 targets in /home/stevechavez/Projects/postgrest/test/load/gen_targets.http (0.90s)
postgrest-with-postgresql-17: You can connect with: psql 'postgres:///postgres?host=/run/user/1000/postgrest/postgrest-with-postgresql-17-MkP/socket' -U postgres
postgrest-with-postgresql-17: You can tail the logs with: tail -f /run/user/1000/postgrest/postgrest-with-postgresql-17-MkP/db.log
Building postgrest (nix)... done.
Starting postgrest... done.
Requests      [total, rate, throughput]         50001, 3427.18, 3427.11
Duration      [total, attack, wait]             14.59s, 14.59s, 1.115µs
[...]

Issue #3101, opened 2023-12-12
(refactor) Too easy to miss enabling a config as in-db and show it on the CLI


steve-chavez:
## Problem

This occurred on:

- `server-timing-enabled` missing on the CLI https://github.com/PostgREST/postgrest/issues/3100
- `jwt-cache-max-lifetime` missing in-db config on https://github.com/PostgREST/postgrest/pull/3069

This happens because the logic for the CLI and the in-db config is disconnected from the `AppConfig` type

https://github.com/PostgREST/postgrest/blob/0c65bf494d75fbb2b8650eba1385b8d2c5d042d2/src/PostgREST/CLI.hs#L124-L233

https://github.com/PostgREST/postgrest/blob/0c65bf494d75fbb2b8650eba1385b8d2c5d042d2/src/PostgREST/Config/Database.hs#L44-L69

https://github.com/PostgREST/postgrest/blob/0c65bf494d75fbb2b8650eba1385b8d2c5d042d2/src/PostgREST/Config.hs#L70-L115

## Solution

Tie the `AppConfig` or other intermediate type to the CLI an in-db logic.

develop7:
The https://hackage.haskell.org/package/configuration-tools (source at https://github.com/alephcloud/hs-configuration-tools) could be the right tool for the job — it combines CLI options parser, *JSON instances, JSON/YAML config parsing (that's where it falls short for us, though — need to investigate how hard adding custom config formats' support could be), remote configs even, if needed.

steve-chavez:
If we're going to change the file format, I think what would make most sense is using an INI format to be compatible with the [pg service file](https://www.postgresql.org/docs/9.1/libpq-pgservice.html) (also see this [blog post](https://www.cybertec-postgresql.com/en/pg_service-conf-the-forgotten-config-file/)).

That way, we could support multiple databases (https://github.com/PostgREST/postgrest/issues/2798) with a config like:

```ini
# pg_service.ini
[serviceone]
port=5432
user=stevechavez
dbname=stevechavez

[serviceone.postgrest]
jwt-secret = xxxx
db-pool = 10

[servicetwo]
port=5432
...

[servicetwo.postgrest]
jwt-secret = yyy
db-pool = 20
other-configs = ".."
```

This also has the advantage of being able to use psql like:

```
PGSERVICE=serviceone PGSERVICEFILE=pg_service.ini psql

psql (14.10 (Ubuntu 14.10-0ubuntu0.22.04.1))
[...]

PR #4084, opened 2025-05-13
feat: JWT cache implementation based on sieve algorithm

mkleczek:
Draft version of JWT cache implementation based on https://cachemon.github.io/SIEVE-website/blog/2023/12/17/sieve-is-simpler-than-lru/ algorithm


mkleczek:
This PR contains:

1. Refactoring and some cleanup of JWT handling code:
    - Moved JWT parsing and validation to a separate module Auth.JWT
    - Split JWT decoding, parsing and signature validation, claims validation into separate functions
    - Instead of caching `AuthResult` cache decoded claims (which signature was verified). Validating claims and determining role is done *after* cache lookup
    - Cleaned up API so that usage of it is simplified: `lookupJwtCache cache key >>= parseClaims configJwtAud time`
    - handling of JwtCacheState initialization and updates of configuration is encapsulated in Auth.JwtCache module
2. Generic scalable cache implementation based on _stm_,  _stm-hamt_ and _sieve_ algorithm. It also provides usage stats (ie. hit ratio, evictions count, size)

steve-chavez:
Very interesting! :eyes: :eyes: 

> It also provides usage stats (ie. hit ratio, evictions count, size)

Maybe we could expose those as [metrics](https://docs.postgrest.org/en/v13/references/observability.html#metrics). That would help with passing code coverage too, since it looks like it's detecting those functions as dead code:

> postgrest-13.1-inplace: src/PostgREST/Cache/Sieve.hs:124:1: delete
> postgrest-13.1-inplace: src/PostgREST/Cache/Sieve.hs:133:1: deleteIO
> postgrest-13.1-inplace: src/PostgREST/Cache/Sieve.hs:136:1: resetIO
> postgrest-13.1-inplace: src/PostgREST/Cache/Sieve.hs:142:1: accessStats
> postgrest-13.1-inplace: src/PostgREST/Cache/Sieve.hs:148:1: evictionsCount

I also see the JWT loadtest failing on CI https://github.com/PostgREST/postgrest/actions/runs/15030111955?pr=4084:

> Options "http://postgrest/authors_only": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

Not sure what's going on there, but you should be able to reproduce it locally with `postgrest-loadtest -k jwt`.

steve-chavez:
> postgrest-loadtest -k jwt gives me:
> delaying data to/from PostgREST by 0ms
> jwt": -c: line 1: unexpected EOF while looking for matching `"'

Weird, it's like you're using and old version of the nix tools and it's ignoring the `-k jwt` arg. On my machine I get:

```bash
$ postgrest-loadtest -k jwt
Created 50000 targets in ./test/load/gen_targets.http (0.79s)
...

## this is the ouput you're getting
$ postgrest-loadtest -k mixed # same as just "postgrest-loadtest"
delaying data to/from postgres by 0ms 
```

Maybe try going out of `nix-shell` and running it again.

mkleczek:
Yeah, that was it, thanks.

I have serious doubts about `postgrest-loadtest -k jwt`. When testing `main` on my machine (beefy MacBook) there is no difference in results between `PGRST_JWT_CACHE_MAX_LIFETIME=86400` and `PGRST_JWT_CACHE_MAX_LIFETIME=0`. I guess that's because symmetric crypto is used to generate/validate tokens. That's probably too fast to be worth caching and show any difference between cached and not cached access. I think meaningful tests have to be based on asymmetric cryptography.

Nevertheless there indeed seems to be an issue with implementation in this PR: all percentiles (ie. median up to 99th) show _better_ results than `main` but _max_ is 30s (which causes timeouts and skews average response time). Looking at it right now.

steve-chavez:
> That's probably too fast to be worth caching and show any difference between cached and not cached access. I think meaningful tests have to be based on asymmetric cryptography.

Agree, we can change that.

> When testing main on my machine (beefy MacBook) there is no difference in results between PGRST_JWT_CACHE_MAX_LIFETIME=86400 and PGRST_JWT_CACHE_MAX_LIFETIME=0.

Note that the jwt loadtest actually generates unique JWTs , it was mainly done to test the jwt decoding perf while also ensuring the JWT cache purging doesn't slow things down (https://github.com/PostgREST/postgrest/pull/4034).

The "mixed" loadtest does have hardcoded JWTs, and the perf looks more or less maintained (or slightly better).

mkleczek:
@steve-chavez @wolfgangwalther @taimoorzaeem

I am not sure what to do with outstanding test coverage issues. This is about `delete` and `deleteIO` not being used. And indeed - right now there is no explicit cache entry removal anywhere. These functions are meant to be used in case of claims validation errors.

The question is: how to handle invalid JWTs? There are several options:

1. Do not cache invalid JWTs at all
2. Do not cache JWTs that cannot be parsed or have invalid signature. Cache all other JWTs, also JWTs containing invalid claims.
3. Cache all token parsing and signature validation results - this is currently implemented option (ie. the type of values in the cache is `Either Error (JSON.Object)`
4. Cache all tokens that can be parsed (but possibly having invalid signature).

I've decided option three makes sense as it also speeds up repeated invalid token requests. OTOH - it opens up possibility of _cache filling_ attacks using randomly generated tokens. It is disputable if that is a bigger problem than overloading CPU with eg. JWTs signed with random key.

It looks to me option 4 would be the best compromise because it would prevent filling up cache with random garbage but would offload signature validation. I haven't implemented it as I don't know how to split parsing and signature validation with JOSE.

WDYT?

mkleczek:
> Very interesting! 👀 👀
> 
> > It also provides usage stats (ie. hit ratio, evictions count, size)
> 
> Maybe we could expose those as [metrics](https://docs.postgrest.org/en/v13/references/observability.html#metrics).

@steve-chavez 

Done.

mkleczek:
> But do we really need to? Parsing a JWT is really simple - but creating parseable tokens, with invalid signatures is just as easy. So I don't really see the additional value of option 4 vs option 3. If somebody wants to fill the cache, they can do so easily.

Indeed - any random bytes are a signature that consumes CPU to validate.

So that leaves us with the choice between caching negative results or caching only valid JWTs.

I've made cache implementation polymorphic over value computation monad so it is easy to change strategies. The latest commits introduce possibility of selecting one of two variants in PostgREST.Auth.JwtCache - it the matter of changing `cachingErrors` to `notCachingErrors` in `newJwtCache`. For now I've left both variants and disabled _unused local bindings_ warning to quiet linter. Once we decide which one is better we can remove the other one (or we can decide to make it configurable and leave both).

mkleczek:
> > So that leaves us with the choice between caching negative results or caching only valid JWTs.
> 
> I wonder whether we can cache all valid JWTs, but including expired ones? Aka when validation fails because of expiry, still cache. When validation fails because of something else, don't.

That’s exactly option 2. Which is currently implemented as notCachingErrors variant (switchable in JwtCache.init).

As described in the description of the PR (first comment) - we don’t cache AuthResults but raw parsed claims that are re-validated for each request.

mkleczek:
> > That’s exactly option 2.
> 
> Well, not exactly, but close enough I agree. We'd still cache those that fail, for example, and audience check or so. But that's totally fine, yes.

Right - it is not exactly the same.

The reason I decided to leave claims checking until after cache lookup are two-fold:

1. Time sensitive claims validation is dependent on well... time :) Not only `exp` but also for example `nbf` - it might become valid in the future.
2. On the other hand `aud` validation is configuration sensitive and I wanted to minimize the number of configuration options which require whole cache reset when changed.

mkleczek:
> > I wanted to minimize the number of configuration options which require whole cache reset when changed.
> 
> Yes, this makese a lot of sense!
> 
> So the only change that requires a reset is the change of secret, right?

That, and - of course - turning off caching alltogether (ie. setting jwt-cache-max-size=0)

mkleczek:
@steve-chavez @wolfgangwalther @taimoorzaeem 

After some more work on this PR I think it is now in a mergeable state (pending documentation changes, changelog adjustments etc. - and of course code review).

I am pretty confident it is working fine as I've added JWT cache behavior tests that verify hits/misses and evictions using metrics.

Please, let me know if there are any adjustments / changes required (or if you think the whole idea is wrong).

src/PostgREST/Config.hs
```diff
@@ -287,7 +289,8 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
     <*> (fromMaybe False <$> optWithAlias
           (optBool "jwt-secret-is-base64")
           (optBool "secret-is-base64"))
-    <*> (fromMaybe 0 <$> optInt "jwt-cache-max-lifetime")
+--    <*> (fromMaybe 0 <$> optInt "jwt-cache-max-lifetime")
+    <*> (fromMaybe 0 <$> optInt "jwt-cache-max-size")
```
taimoorzaeem, line 293: So the `size` in there seems bit confusing as it may be confused with size in memory. How about renaming the config to `jwt-cache-max-entries`? It seems more appropriate, no?

mkleczek: > So the `size` in there seems bit confusing as it may be confused with size in memory. How about renaming the config to `jwt-cache-max-entries`? It seems more appropriate, no?

Fine for me. @steve-chavez WDYT?

steve-chavez: Yes, agree with `jwt-cache-max-entries`.

test/spec/Feature/Auth/JwtCacheSpec.hs
```diff
@@ -0,0 +1,167 @@
...
+spec :: SpecWith (MetricsState, Application)
+spec = describe "Server started with JWT and metrics enabled" $ do
+  it "Should not have JWT in cache" $ do
+    let auth = genToken [json|{"exp": 9999999999, "role": "postgrest_test_author", "id": "jdoe1"}|]
+
+    expectCounters
+      [
+        requests (+ 1)
+      , hits     (+ 0)
+      ] $
```
steve-chavez, line 37: This is a really elegant way to test the metrics :100: So easy to read!

test/spec/Feature/Auth/JwtCacheSpec.hs
```diff
@@ -0,0 +1,167 @@
...
+    expectCounters
+      [
+        requests (+ 2)
+      , hits     (+ 1)
+      ] $
+
+         request methodGet "/authors_only" [auth] "" `shouldRespondWith` 401
+      *> request methodGet "/authors_only" [auth] "" `shouldRespondWith` 401
+
+  it "Should evict entries from the JWT cache" $ do
```
steve-chavez, line 77: Maybe inline the conf for easier readability.
 
```suggestion
  it "Should evict entries from the JWT cache (jwt cache max is 2)" $ do
```

src/PostgREST/CLI.hs
```diff
@@ -203,8 +203,8 @@ exampleConfigFile =
       |# jwt-secret = "secret_with_at_least_32_characters"
       |jwt-secret-is-base64 = false
       |
-      |## Enables and set JWT Cache max lifetime, disables caching with 0
-      |# jwt-cache-max-lifetime = 0
+      |## Enables JWT Cache and sets its max size, disables caching with 0
+      |# jwt-cache-max-size = 0
```
steve-chavez, line 207: I think we should make the default 1000. That means the cache will be enabled by default for next major.

steve-chavez: 1000 is a rough estimation, mentioned before on https://github.com/PostgREST/postgrest/pull/3802#issuecomment-2755566175

mkleczek:
> @mkleczek I've noticed that the [jwt loadtest results](https://github.com/PostgREST/postgrest/actions/runs/15274145867?pr=4084) show a perf drop compared to main and latest version. Is that expected?

Our load test is the worst possible case for this (and I would say any) cache: all JWTs are different so no caching but - if the cache is smaller than the number of requests then we pay the price of eviction and garbage collection.

src/PostgREST/CLI.hs
mkleczek: > I think we should make the default 1000. That means the cache will be enabled by default for next major.

JWT cache makes sense **only** with asymmetric keys so I wouldn't turn it on by default

mkleczek:
@steve-chavez @wolfgangwalther 

Looks like combining the default cache size from https://github.com/PostgREST/postgrest/pull/4084#discussion_r2110593171 with auto-configuration of the cache idea from https://github.com/PostgREST/postgrest/pull/4084#issuecomment-2915276059 allows us to have a pretty good user experience:

1. If max size is **not** configured, perform auto-configuration that turns on JWT cache of size 1000 but **only** if we have asymmetric keys in configuration.
2. Max size of <= 0 disables caching regardless of the configured keys.
3. Max size > 0 forces caching regardless of the configured keys.

Implemented in https://github.com/PostgREST/postgrest/pull/4084/commits/c6e73b13f41c19750aff56e13d56f265d0b68ca7

src/PostgREST/CLI.hs
mkleczek: > I think we should make the default 1000. That means the cache will be enabled by default for next major.

https://github.com/PostgREST/postgrest/pull/4084#issuecomment-2915678736

steve-chavez:
> updating yet another metric (and of questionable value) would eat up additional cycles

Cool. Fair enough.

> This is somewhat surprising - we do not remove from the cache explicitly right now, so it should be always full (hence occupying constant memory). Maybe what you see is GC effects?

Right. It looks like it's indeed GC as RSS and VSZ both decrease simultaneously as seen on:

[Screencast from 05-28-2025 02:12:05 PM.webm](https://github.com/user-attachments/assets/434f1edb-2341-4450-8a9a-12c58408fbbd)

---

I detected a problem with the previous JWT cache that this PR solves https://github.com/PostgREST/postgrest/issues/4107.

mkleczek:
> > Our load test is the worst possible case for this (and I would say any bounded) cache: all JWTs are different so no caching
> 
> With the new default, we should change the JWT load test to use asymmetric keys - and then see how much perf we lose in that worst case.

@wolfgangwalther @steve-chavez  @taimoorzaeem
See https://github.com/PostgREST/postgrest/pull/4084/commits/4d7ccc431409bc8c35335e1609ab3b4b63c23b8a

I've changed load test to use 1000 JWTs signed by RSA4096 key. Also added additional parameter `--jwtcache=off` to turn off JWT caching to compare results.

In case of asymmetric RSA caching gives around 4x performance boost (all response times up to 95th percentile are 4x lower with caching turned on).

steve-chavez:
> I've changed load test to use 1000 JWTs signed by RSA4096 key. Also added additional parameter --jwtcache=off to turn off JWT caching to compare results.

@mkleczek Cool! Can you split those commits to another PR for easier review/merge?

src/PostgREST/Metrics.hs
```diff
@@ -19,22 +20,36 @@ import PostgREST.Observation
...
+    register (gauge (Info "pgrst_db_pool_available" "Available connections in the pool")) <*>
+    register (gauge (Info "pgrst_db_pool_waiting" "Requests waiting to acquire a pool connection")) <*>
+    register (gauge (Info "pgrst_db_pool_max" "Max pool connections")) <*>
+    register (vector "status" $ counter (Info "pgrst_schema_cache_loads_total" "The total number of times the schema cache was loaded")) <*>
+    register (gauge (Info "pgrst_schema_cache_query_time_seconds" "The query time in seconds of the last schema cache load")) <*>
+    register (counter (Info "pgrst_jwt_cache_requests_total" "The total number of JWT cache lookups")) <*>
+    register (counter (Info "pgrst_jwt_cache_hits_total" "The total number of JWT cache hits")) <*>
+    register (counter (Info "pgrst_jwt_cache_evictions_total" "The total number of JWT cache evictions"))
+  setGauge (poolMaxSize metricState) (fromIntegral configDbPoolSize)
+  pure metricState
```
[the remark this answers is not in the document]
mkleczek: https://github.com/PostgREST/postgrest/pull/4120

mkleczek:
> > The commit history in the PR currently represents which steps you took while you were working on it. Now, we need to get it into a reviewable and mergeable shape. That's because we can't squash merge, but also not merge it as-is for a clean history.
> > So, it would be great if you could rebase to create a nice history of self-contained commits. At least with the following:
> > 
> > * refactor for the split of auth module
> 
> The problem with this is that currently in main we don't really have a proper interface for caching and everything is intermingled. I thought it does not really make too much sense to perform the refactoring while leaving current JWT cache implementation. Will think on how to do that without too much effort, if possible.
> 

https://github.com/PostgREST/postgrest/pull/4121

> > * refactor in the Metrics module
> 
> +1
> 

https://github.com/PostgREST/postgrest/pull/4120

> > * improving the jwt loadtest as Steve mentioned
> 
> +1
> 

https://github.com/PostgREST/postgrest/pull/4118

> 
> > We'll need to slice this nicely, otherwise there is no chance to review this properly. I still left a few nits while skimming the code.
> 
> Thanks for this review.

Not sure how to proceed now - should I rebase this PR on top of the three above? Or wait for them to be merged?

wolfgangwalther:
> Not sure how to proceed now - should I rebase this PR on top of the three above? Or wait for them to be merged?

Let's wait for them to be merged, that will make rebasing easier / less often to repeat.

src/PostgREST/Config.hs
mkleczek: Changed to `jwt-cache-max-entries`

mkleczek:
> > Not sure how to proceed now - should I rebase this PR on top of the three above? Or wait for them to be merged?
> 
> Let's wait for them to be merged, that will make rebasing easier / less often to repeat.

@steve-chavez @wolfgangwalther @taimoorzaeem 

I've rebased it on latest `main` and made adjustments based on the comments so far.

I also updated documentation but bear with me, I'm not a native speaker and it might need some work.

src/PostgREST/Config.hs
```diff
@@ -97,7 +97,7 @@ data AppConfig = AppConfig
   , configJwtRoleClaimKey          :: JSPath
   , configJwtSecret                :: Maybe BS.ByteString
   , configJwtSecretIsBase64        :: Bool
-  , configJwtCacheMaxLifetime      :: Int
+  , configJwtCacheMaxSize          :: Maybe Int
```
taimoorzaeem: Not that it matters much, but for the sake of consistency:
```suggestion
  , configJwtCacheMaxEntries       :: Maybe Int
```

test/spec/SpecHelper.hs
```diff
@@ -140,7 +140,7 @@ baseCfg = let secret = encodeUtf8 "reallyreallyreallyreallyverysafe" in
   , configJwtRoleClaimKey           = [JSPKey "role"]
   , configJwtSecret                 = Just secret
   , configJwtSecretIsBase64         = False
-  , configJwtCacheMaxLifetime       = 0
+  , configJwtCacheMaxSize           = Just 10
```
taimoorzaeem: And here as well:
```suggestion
  , configJwtCacheMaxEntries        = Just 10
```

src/PostgREST/Observation.hs
```diff
@@ -151,6 +153,10 @@ observationMessage = \case
     "Trying to borrow a connection from pool"
   PoolRequestFullfilled ->
     "Borrowed a connection from the pool"
+  JwtCacheLookup _ ->
+    "Looked up a JWT in JWT cache"
+  JwtCacheEviction ->
+    "Evicted entry from JWT cache"
```
taimoorzaeem, line 159: I think we should only log these on `logLevel >= LogDebug`. You can add this check here:

https://github.com/PostgREST/postgrest/blob/f85113669d29962699f11cacb06f1953bac23254/src/PostgREST/Logger.hs#L100-L103

src/PostgREST/Config.hs
mkleczek: Good catch, fixed.

src/PostgREST/Observation.hs
mkleczek: Fixed.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   + It now shows the invalid schema in the `message` field.
   + The exposed schemas are now listed in the `hint` instead of the `message` field.
 - Improve error details of `PGRST301` error by @taimoorzaeem in #4051
+- JWT cache based on sieve algorithm
```
wolfgangwalther: Since we already have a JWT cache and the existing implementation, this should be in "Changed", not "Added", right?

When it is a change, should it mention the changed config option? Users might need to migrate that.

Also should this point to this PR at the end, for the release notes?

docs/references/auth.rst
```diff
@@ -98,9 +98,11 @@ The ``Bearer`` header value can be used with or without capitalization(``bearer`
 JWT Caching
 -----------
 
-PostgREST validates ``JWTs`` on every request. We can cache ``JWTs`` to avoid this performance overhead.
+PostgREST validates ``JWTs`` on every request. Asymmetric signature validation (such as RSA) is slow and we can cache ``JWT`` validation results to avoid this performance overhead.
 
-To enable JWT caching, the config :code:`jwt-cache-max-lifetime` is to be set. It is the maximum number of seconds for which the cache stores the JWT validation results. The cache uses the :code:`exp` claim to set the cache entry lifetime. If the JWT does not have an :code:`exp` claim, it uses the config value. See :ref:`jwt-cache-max-lifetime` for more details.
+JWT caching is automatically enabled if ref:`jwt-secret` is set to an asymmetric key. Otherwise it is disabled and can be enabled by setting the config :code:`jwt-cache-max-entries` to a value greater than 0. Setting it to 0 disables caching regardless of ref:`jwt-secret`
```
wolfgangwalther:
```suggestion
JWT caching is automatically enabled if ref:`jwt-secret` is set to an asymmetric key. Otherwise it is disabled and can be enabled by setting the config :code:`jwt-cache-max-entries` to a value greater than 0. Setting it to 0 disables caching regardless of ref:`jwt-secret`.
```

src/PostgREST/Config.hs
wolfgangwalther, line 1: I think we should have the part that sets `configJwtCacheMaxEntries` to the default of 1000 or the default of 0, depending on jwt-secret, in `Config.hs`. Thus, `configJwtCacheEntries` should be `Int`, not `Maybe Int`.

For example, this has the benefit that `--dump-config` will tell us exactly which default was chosen, depending on which kind of key I set etc.

Also, it separates the logic better.

mkleczek: I thought about it but such separation gets messy when handling configuration update. You have to handle both `configJWKS` and `configJwtCacheMaxEntries` in `JwtCache` module anyway. Having this logic **only** there seems better encapsulated (but the price is - as you say - lack of default value in config dump).

Additionally - it would make configuration reading monadic (instead of applicative as it is right now). The question really is: how much logic should be in Config module - should it be "dumb" and only responsible for parsing config values and providing defaults while leaving interpretation and handling of dependencies between configuration properties to other modules? Or should it be "smart" and actually become an interpreter of some configuration language assembling application components based on it?

wolfgangwalther: > I thought about it but such separation gets messy when handling configuration update. You have to handle both `configJWKS` and `configJwtCacheMaxEntries` in `JwtCache` module anyway.

I don't see that, yet. My naive understanding of the logic required in `JwtCache` is something like:
- `if new.configJwtCacheMaxEntries == 0 then disable cache`
- `if old.configJWKS != new.configJWKS then reset cache`

None of the default values, nor inspection of the jwt-secret come into play here. What am I missing?

wolfgangwalther: > Additionally - it would make configuration reading monadic (instead of applicative as it is right now). The question really is: how much logic should be in Config module - should it be "dumb" and only responsible for parsing config values and providing defaults while leaving interpretation and handling of dependencies between configuration properties to other modules?

Being able to handle dependencies between different config options would be a good thing, I think. We also have other cases like `server-host` and the not-yet-existing `admin-server-host`, where the latter should default to the former if unset etc.

mkleczek: > > I thought about it but such separation gets messy when handling configuration update. You have to handle both `configJWKS` and `configJwtCacheMaxEntries` in `JwtCache` module anyway.
> 
> I don't see that, yet. My naive understanding of the logic required in `JwtCache` is something like:
> 
> * `if new.configJwtCacheMaxEntries == 0 then disable cache`
> * `if old.configJWKS != new.configJWKS then reset cache`
> 
> None of the default values, nor inspection of the jwt-secret come into play here. What am I missing?

It is not that simple as there are three cases:
- No JWKS
- JWKS and no cache
- JWKS and cache

See https://github.com/PostgREST/postgrest/blob/5ecb6dbc2925a85d807c0e47b95e1a5665c56c74/src/PostgREST/Auth/JwtCache.hs#L49

and
https://github.com/PostgREST/postgrest/blob/5ecb6dbc2925a85d807c0e47b95e1a5665c56c74/src/PostgREST/Auth/JwtCache.hs#L60

You could argue the "no JWKS" case could be handled in `Auth` module but that would mean JWT configuration handling would be spread among three modules (Config, JwtCache, Auth).
IMHO spreading JWT configuration handling logic between two or more modules is not a good idea and that's why I decided to have it all in a single JwtCache module.

Happy to see suggestions how to change it.

mkleczek: > I think we should have the part that sets `configJwtCacheMaxEntries` to the default of 1000 or the default of 0, depending on jwt-secret, in `Config.hs`. Thus, `configJwtCacheEntries` should be `Int`, not `Maybe Int`.
> 
> For example, this has the benefit that `--dump-config` will tell us exactly which default was chosen, depending on which kind of key I set etc.
> 
> Also, it separates the logic better.

So I've been looking at our `Config` module and, TBH, I am not sure if the way it works now fits the needs of more complex scenarios.

I must say - the most confusing is the way how `--dump-config` is supposed to work and what the actual goal of it is. IMHO showing defaults in `--dump-config` is wrong as the user has to guess whether a particular value is explicitly set somewhere or not (it makes `--dump-config` useless for me as my use-case is to find out if a particular config option is set explicitly in one of the configuration sources).

Then there is also the question about _magic values_ which meaning is not explicit. Like `jwt-cache-max-entries=0` (or `jwt-cache-max-age=0` in `main`) that disables JWT caching, or `jwt-secret` that disables JWT handling at all when not set, or `db-anon-role` that if not set disables anonymous access.

What I would really expect are **two** different functions:

1. Dumping _raw_ configuration (ie. configuration values that are explicitly provided in one of the configuration sources) - if we could provide the actual source of a particular value it would be even better.
2. Providing higher level view of the actual state of various internal objects that depend on provided configuration, eg:
   1. How JWT handling is configured - is JWT enabled at all (since it is disabled if `jwt-secret` is not provided)? Is JWT caching turned on? How big is the JWT cache?
   2. How DB connection pool is configured - the actual connection settings, the actual pool size etc.
   3. Is anonymous access enabled and if so - what is the anonymous role?
   4. etc.

And we do have both actually: one is `--dump-config` and another is metrics/observations. I wouldn't mix the two and I would leave `Config` as _dumb_ configuration reader, `--dump-config` as _dumb_ configuration output, while metrics/observations should be used to find out the actual state of the system (together with the way how it is actually configured at the moment).

steve-chavez:
@mkleczek Just merged https://github.com/PostgREST/postgrest/pull/4142. Once you rebase the loadtest will now include a markdown table showing process monitoring (MEM and CPU) for each elapsed second, comparing head/main/latest-version.

This should prove that your PR now makes the JWT cache memory bound. I've tried this locally and it should show the MEM % staying at 0 % instead of always growing :raised_hands:

mkleczek:
> @mkleczek Just merged #4142. Once you rebase the loadtest will now include a markdown table showing process monitoring (MEM and CPU) for each elapsed second, comparing head/main/latest-version.
> 
> This should prove that your PR now makes the JWT cache memory bound. I've tried this locally and it should show the MEM % staying at 0 % instead of always growing 🙌



> @mkleczek Just merged #4142. Once you rebase the loadtest will now include a markdown table showing process monitoring (MEM and CPU) for each elapsed second, comparing head/main/latest-version.
> 
> This should prove that your PR now makes the JWT cache memory bound. I've tried this locally and it should show the MEM % staying at 0 % instead of always growing 🙌

I've rebased on top of current main - it produces the tables but I am not really sure how to read the results.

steve-chavez:
Yeah, there's something wrong with the script or CI. On main you can clearly see the `jwt-hs` loadtest keeps rising and goes to 3% MEM usage, plus it uses much more %CPU.

https://github.com/PostgREST/postgrest/actions/runs/15698719500

The [results on this PR](https://github.com/PostgREST/postgrest/actions/runs/15706261514?pr=4084) look like they're not from `main` or `v13.0.2`, but rather from HEAD. So there's no comparison really.

steve-chavez:
@mkleczek Actually I don't know if this is because the rebase is somehow wrong, note that it contains other commits (which in turn makes the PR contain unrelated changes - harder to review).

Could you correct the rebase so it only shows the relevant commits?

steve-chavez:
>  but I am not really sure how to read the results.

I've  improved readability of the tables a bit and added a description (https://github.com/PostgREST/postgrest/commit/7dbb8cd9877b27dcb22e3c03a95c8314c68523a5). This might help interpret the results. See an example on https://github.com/PostgREST/postgrest/actions/runs/15719933763

src/PostgREST/Config.hs
steve-chavez: @mkleczek There are several shortcomings of the Config module right now. Some are tracked at https://github.com/PostgREST/postgrest/issues/3101.

> Dumping raw configuration (ie. configuration values that are explicitly provided in one of the configuration sources) - if we could provide the actual source of a particular value it would be even better.

Maybe the source can be put in the comment? As:

```ini
# postgrest --dump-config
...
db-schemas = "public" # api (file), public (env), public (db) 
...
```

mkleczek:
> @mkleczek Actually I don't know if this is because the rebase is somehow wrong, note that it contains other commits (which in turn makes the PR contain unrelated changes - harder to review).
> 
> Could you correct the rebase so it only shows the relevant commits?

Done clean rebase but not sure if it helped.

steve-chavez:
Can reproduce the same locally. It looks like `postgrest-loadtest-against` is not really switching branches and it's only using HEAD for all loadtests.

So maybe our loadtests have been misleading for a while.

**Edit**: I'll try to corroborate this by printing the postgrest version

steve-chavez:
There's a problem with CI or Nix in PRs it seems, opened https://github.com/PostgREST/postgrest/issues/4151

src/PostgREST/Config.hs
steve-chavez: Opened https://github.com/PostgREST/postgrest/issues/4154 for the above

nix/tools/loadtest.nix
```diff
@@ -61,7 +61,6 @@ let
         export PGRST_DB_TX_END="rollback-allow-override"
         export PGRST_LOG_LEVEL="crit"
         export PGRST_JWT_SECRET="reallyreallyreallyreallyverysafe"
-        export PGRST_JWT_CACHE_MAX_LIFETIME="86400"
```
wolfgangwalther, line 62: This line cause the JWT cache to be disabled on main when running the script from this branch against main.

Support for old config needs to be kept in the loadtest in general, to make comparison work backwards. Also see the comment a few lines further up, this line should be *moved there*: 

> previously required settings to make this work with older branches

steve-chavez: Ahhhh, nice catch. This is what caused the confusion on https://github.com/PostgREST/postgrest/issues/4151.

mkleczek: > This line cause the JWT cache to be disabled on main when running the script from this branch against main.
> 
> Support for old config needs to be kept in the loadtest in general, to make comparison work backwards. Also see the comment a few lines further up, this line should be _moved there_:
> 
> > previously required settings to make this work with older branches

Thanks!

Fixed.

steve-chavez:
> What's more: in case of symmetric JWT keys I don't think cache lookup is faster than simply performing JWT verification.

Maybe we should actually confirm if the above is true? It would be less confusing for users to just have one default for both type of keys.

Now the jwt-hs loadtest results (worst case) is proving that the new JWT cache is more efficient than the old one ([results](https://github.com/PostgREST/postgrest/actions/runs/15803495465?pr=4084)).

Additionally on #4151, we were comparing new cache (head) vs old cache disabled (main/v13.0.4) and it showed virtually no difference in throughput or MEM/CPU consumption  ([results](https://github.com/PostgREST/postgrest/actions/runs/15781551670?pr=4153)).

So far it looks we should just enable for both symmetric and asymmetric for simplicity? 

---

> Maybe we should actually confirm if the above is true?

We could add one more loadtest for jwt-hs (not a worst case like it is now) to confirm.

mkleczek:
> > What's more: in case of symmetric JWT keys I don't think cache lookup is faster than simply performing JWT verification.
> 
> Maybe we should actually confirm if the above is true? It would be less confusing for users to just have one default for both type of keys.
> 
> Now the jwt-hs loadtest results (worst case) is proving that the new JWT cache is more efficient than the old one ([results](https://github.com/PostgREST/postgrest/actions/runs/15803495465?pr=4084)).
> 

Possibly, but in both cases the cache is not really used and is just unnecessary overhead.

> Additionally on #4151, we were comparing new cache (head) vs old cache disabled (main/v13.0.4) and it showed virtually no difference in throughput or MEM/CPU consumption ([results](https://github.com/PostgREST/postgrest/actions/runs/15781551670?pr=4153)).
> 
> So far it looks we should just enable for both symmetric and asymmetric for simplicity?
> 
> > Maybe we should actually confirm if the above is true?
> 
> We could add one more loadtest for jwt-hs (not a worst case like it is now) to confirm.

There is `--jwtcache=off` switch available in `jwt-hs` and `jwt-rsa` load tests but I am not sure how to add additional test run.

mkleczek:
> > > What's more: in case of symmetric JWT keys I don't think cache lookup is faster than simply performing JWT verification.
> > > Maybe we should actually confirm if the above is true?
> > 
> > 
> > We could add one more loadtest for jwt-hs (not a worst case like it is now) to confirm.
> 
> There is `--jwtcache=off` switch available in `jwt-hs` and `jwt-rsa` load tests but I am not sure how to add additional test run.

@steve-chavez 
So I've run some experiments locally with 1000 symmetric JWTs and compared caching and non-caching configuration. I found no difference in response times (even though all JWTs were cached). In other words caching did not help but also did not add any noticeable overhead.

I would say it is your call if we should get rid of auto-configuring cache for asymmetric keys and leaving it off for symmetric. (Although I personally like the idea of "smart" defaults and "smart" auto configuration in general - just let the software "do the right thing").
If you think it is more confusing to be "smart" then let's be it.

WDYT?

wolfgangwalther:
We don't need smart defaults, if enabling the cache *can't hurt*. If we're certain of that, then there is no reason to be able to disable the cache at all, right? Would it make the code simpler and thus easier to maintain?

steve-chavez:
> So I've run some experiments locally with 1000 symmetric JWTs and compared caching and non-caching configuration. I found no difference in response times (even though all JWTs were cached). In other words caching did not help but also did not add any noticeable overhead.

@mkleczek Do you think the JWT cache could improve perf for symmetric JWTs in the future? Maybe it can be improved later somehow?

> If we're certain of that, then there is no reason to be able to disable the cache at all, right? Would it make the code simpler and thus easier to maintain?

Right, I think maintainability is a stronger reason to not have smart defaults.

mkleczek:
@wolfgangwalther @steve-chavez 

As requested I removed "smart" auto configuration of JWT cache.
Right now JWT cache of size 1000 is enabled by default regardless of JWT key. It can be turned off by setting `jwt-cache-max-entries = 0`

CHANGELOG.md

mkleczek:
> > So I've run some experiments locally with 1000 symmetric JWTs and compared caching and non-caching configuration. I found no difference in response times (even though all JWTs were cached). In other words caching did not help but also did not add any noticeable overhead.
> 
> @mkleczek Do you think the JWT cache could improve perf for symmetric JWTs in the future? Maybe it can be improved later somehow?

I don't think so. Symmetric encryption is very fast and hardware acceleration of it is widely supported. We are trading cpu for RAM here and memory latency is inherently difficult to overcome.

So the only good reason to keep cache turned on for symmetric JWTs is simplicity/maintainability.

steve-chavez:
> @steve-chavez
So I've run some experiments locally with 1000 symmetric JWTs and compared caching and non-caching configuration. I found no difference in response times (even though all JWTs were cached). In other words caching did not help but also did not add any noticeable overhead.

I've finished the jwt loadtests and now the above can be proven on CI too: https://github.com/PostgREST/postgrest/pull/4175

@mkleczek Could you rebase here and correct the failing io test?

> I also updated documentation but bear with me, I'm not a native speaker and it might need some work.

I believe no one here is a native speaker :smiley:. Code LGTM so next I'll review the docs and then we can merge.

steve-chavez:
Just managed to understand the failure on CI:

```
AssertionError: assert 'Trying to borrow a connection from pool' in '27/Jun/2025:13:06:39 +0000: Looked up a JWT in JWT cache\n'
```

It looks like the JWT cache is not completely disabled when having `PGRST_JWT_CACHE_MAX_ENTRIES: 0` (the `test_log_pool_req_observation` doesn't have this setup but it should).

To reproduce:

```bash
# nix-shell
PGRST_LOG_LEVEL=debug PGRST_JWT_CACHE_MAX_ENTRIES=0 postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run 

curl localhost:3000/authors_only -H "Authorization: Bearer $(postgrest-gen-jwt postgrest_test_author)"

16/Jul/2025:18:10:45 -0500: Schema cache loaded 271 Relations, 232 Relationships, 148 Functions, 15 Domain Representations, 45 Media Type Handlers, 1196 Timezones
16/Jul/2025:18:10:45 -0500: Schema cache loaded in 15.8 milliseconds
16/Jul/2025:18:10:46 -0500: Looked up a JWT in JWT cache
```

src/PostgREST/Auth/JwtCache.hs
```diff
@@ -1,99 +1,112 @@
...
+  forall m v. CacheVariant m v => JwtCache JwkSet (TVar Int) (SC.Cache m ByteString v)
+
+instance CacheVariant IO (Either Error JSON.Object) where
+  cached c = lift . SC.cached c >=> liftEither
+
+instance CacheVariant (ExceptT Error IO) JSON.Object where
+  cached = SC.cached
+
+decode :: JwtCache -> ByteString -> ExceptT Error IO JSON.Object
+decode JwtNoJwks        = const $ throwError (JwtErr JwtSecretMissing)
```
steve-chavez, line 69: The `JwtSecretMissing` doesn't seem right on the `JwtCache` module, it looks to me it should be handled on a previous step (the cache module should only be entered if we have a secret, no secret will always result in failure and no caching right?). 

I just realized some commented out code was left on main: https://github.com/PostgREST/postgrest/blob/2d4967c4b40faea70a1c0464eac0cd85c6e5a88b/src/PostgREST/Auth/Jwt.hs#L102-L106

Done on https://github.com/PostgREST/postgrest/commit/a409a2cb9487216a4dba6d955a79a6f1121d5edf

Which also moved the `JwtSecretMissing` from Jwt.hs, which looked like a better place for it.

Or having it in Auth.hs would also be fine.

This might simplify the `JwtCache` type and remove the `JwtNoJwks` constructor.

src/PostgREST/Auth.hs
```diff
@@ -30,50 +30,32 @@ import System.TimeIt    (timeItT)
...
 middleware :: AppState -> Wai.Middleware
 middleware appState app req respond = do
-  cfg@AppConfig{..} <- getConfig appState
+  conf@AppConfig{..} <- getConfig appState
   time <- getTime appState
 
   let token  = Wai.extractBearerAuth =<< lookup HTTP.hAuthorization (Wai.requestHeaders req)
-      parseAuthToken = maybe (const $ throwError (JwtErr JwtSecretMissing)) parseAndDecodeClaims configJWKS
-      parseJwt = runExceptT $ maybe (pure KM.empty) parseAuthToken token >>= parseClaims cfg time
+      parseJwt = runExceptT $ lookupJwtCache jwtCacheState token >>= parseClaims conf time
```
steve-chavez, line 49: We should omit the `lookupJwtCache` step if `configJwtCacheMaxEntries` is 0, that should avoid the problem on https://github.com/PostgREST/postgrest/pull/4084#issuecomment-3081692509 (this makes sense too).

src/PostgREST/Auth/JwtCache.hs
```diff
@@ -1,99 +1,112 @@
...
+          JwtCache key maxSizeTVar <$>
+            -- select cachingErrors or notCachingErrors
+            notCachingErrors (readTVar maxSizeTVar) key
+
+    cachingErrors :: STM Int -> JwkSet -> IO (SC.Cache IO ByteString (Either Error JSON.Object))
+    cachingErrors maxSize key = SC.cacheIO (SC.CacheConfig maxSize
+            (runExceptT . parseAndDecodeClaims key)
+            (observationHandler . JwtCacheLookup) -- lookup metrics
+            (const . const $ observationHandler JwtCacheEviction) -- evictions metrics
+            alwaysValid) -- no invalidation for now
```
steve-chavez: We should now delete this and remove the top `{-# OPTIONS_GHC -Wno-unused-local-binds #-}` right?

src/PostgREST/Cache/Sieve.hs
```diff
@@ -0,0 +1,218 @@
...
+alwaysValid = pure (const . const Nothing)
+
+cacheIO :: (MonadIO m, Hashable k) => CacheConfig m k v -> IO (Cache m k v)
+cacheIO = atomically . cache
+
+cache :: (MonadIO m, Hashable k) => CacheConfig m k v -> STM (Cache m k v)
+cache cacheConfig = mdo
+    tail <- newTVar (Some head)
+    entries <- SH.new
+    finger <- newTVar tail
```
steve-chavez, line 79: slightly pedantic, this should be `hand` right?  Just understood the algorithm from the blog post and they use the term `hand`.

src/PostgREST/Auth/JwtCache.hs
steve-chavez: Just noticed this was already discussed on https://github.com/PostgREST/postgrest/pull/4084/files#r2134697995

Hm, both independent reviews do confirm that having the `JwtSecretMissing` outside of the cache module makes the most sense.

mkleczek: @steve-chavez
The issue is handling of configuration changes. How should it work in `JwtCache`?
1. It should look only at `jwt-cache-max-entries` config - then we would keep the cache even if JWT keys are not configured
2. It should also check JWT secret config to destroy the cache if it is not configured. But then it would either have to create an empty cache anyway or have a special value in `jwtCacheState` for _missing keys_ case (that's what it does right now). But then any checks in Auth module would be redundant.

src/PostgREST/Cache/Sieve.hs
mkleczek: I've read a couple of posts and seen both _finger_ and _hand_. I liked `finger` better as it _points_ at an entry. Can change it if you think it is more understandable.

steve-chavez: On second thought I think it's fine, let's leave it as is.

src/PostgREST/Auth/JwtCache.hs
steve-chavez: @mkleczek Got it, I agree now considering the config changes that can invalidate the cache.

How about a comment on `JwtCache` to clarify those cases? That would clarify the need for `JwtNoJwks` as part of it.

src/PostgREST/Auth.hs
steve-chavez: Now that we have the agreement on https://github.com/PostgREST/postgrest/pull/4084#discussion_r2213817445, maybe we should now prevent running these: 

```
            (observationHandler . JwtCacheLookup) -- lookup metrics
            (const . const $ observationHandler JwtCacheEviction) -- evictions metrics
```

When `configJwtCacheMaxEntries` is 0? Or how to solve https://github.com/PostgREST/postgrest/pull/4084#issuecomment-3081692509?

src/PostgREST/Auth/JwtCache.hs
mkleczek: @steve-chavez
Sorry about delays in taking care of it - I am on holidays until next week. Will take care of the issues once I'm back.

steve-chavez:
> So I've run some experiments locally with 1000 symmetric JWTs and compared caching and non-caching configuration. I found no difference in response times (even though all JWTs were cached). In other words caching did not help but also did not add any noticeable overhead.

So comparing the [CI results](https://github.com/PostgREST/postgrest/actions/runs/16567892701?pr=4084) for `jwt-hs` vs `jwt-hs-cache` it does look like the cache is still a win for symmetric (more throughput), while CPU usage is reduced and MEM increases slightly (a couple MB). For both `jwt-hs`/`jwt-rsa` vs `jwt-hs-cache`/`jwt-rsa-cache`, `main` and `v13.0.4` had a bit more throughput, but of course in terms of CPU/MEM usage we're much much better with the sieve cache.

Note: ignore `jwt-rsa-cache-worst` for now. It has some `401 Unauthorized` failures, I need to fix/tune the duration of the JWT exp.

mkleczek:
@steve-chavez I believe outstanding issues have been addressed now, could you double check?

test/io/test_io.py
```diff
@@ -1838,10 +1838,10 @@ def test_log_pool_req_observation(level, defaultenv):
...
 
         if level == "debug":
-            output = postgrest.read_stdout(nlines=4)
-            assert pool_req in output[0]
-            assert pool_req_fullfill in output[3]
-            assert len(output) == 4
+            output = postgrest.read_stdout(nlines=5)
+            assert pool_req in output[1]
+            assert pool_req_fullfill in output[4]
+            assert len(output) == 5
```
steve-chavez, line 1844: @mkleczek Just confirmed the problem on https://github.com/PostgREST/postgrest/pull/4084#issuecomment-3081692509 doesn't happen anymore :tada: 

So why do we need to adjust the traces here? I'd suggest just disabling the cache for the `test_log_pool_req_observation` test.

docs/references/auth.rst
```diff
@@ -98,9 +98,11 @@ The ``Bearer`` header value can be used with or without capitalization(``bearer`
 JWT Caching
 -----------
 
-PostgREST validates ``JWTs`` on every request. We can cache ``JWTs`` to avoid this performance overhead.
+PostgREST validates ``JWTs`` on every request. Signature validation (especially asymmetric such as RSA) is slow and we can cache ``JWT`` validation results to avoid this performance overhead.
 
-To enable JWT caching, the config :code:`jwt-cache-max-lifetime` is to be set. It is the maximum number of seconds for which the cache stores the JWT validation results. The cache uses the :code:`exp` claim to set the cache entry lifetime. If the JWT does not have an :code:`exp` claim, it uses the config value. See :ref:`jwt-cache-max-lifetime` for more details.
+JWT caching is automatically enabled unless the config :code:`jwt-cache-max-entries` is set to 0.
+
+See :ref:`jwt-cache-max-entries` for more details.
```
steve-chavez: I think we're missing some details in the docs, I'll help and make a PR to your repo to see if you agree.

test/io/test_io.py
mkleczek: I've added debug logging for JWT cache lookup so the test failed as there was one additional line.

docs/references/auth.rst
mkleczek: That would be great, thanks a lot!

steve-chavez: Added more details on https://postgrest--4228.org.readthedocs.build/en/4228/references/auth.html#jwt-cache

Had to rebase since I shuffled some sections in the auth page, see: https://github.com/PostgREST/postgrest/pull/4228

LMK if it looks good.

mkleczek: @steve-chavez 
I've merged `steve-chavez:jwt-cache-sieve-docs` here.

steve-chavez: @mkleczek Made a PR on https://github.com/mkleczek/postgrest/pull/1

mkleczek: Merged

steve-chavez (approved):
Awesome work! :100:

steve-chavez:
:raised_hands: We're now included in the "Adoption" section of https://cachemon.github.io/SIEVE-website/ :raised_hands:

Issue #3824, opened 2024-12-18
Show only exposed schemas to which the user has access in error message


laurenceisla:
## Problem

When a user uses an incorrect schema, it returns the error message with all exposed schemas:

```bash
$ curl 'localhost:3000/todos' -H 'Accept-Profile: wrong'

{
  ...,
  "message": "The schema must be one of the following: api, auth_api"
}
```

For this example, the anon user has not `USAGE` access to the `auth_api` schema, but it's still shown.

## Solution

It's expected that any exposed schema is already accessible by any request (the burden is on the dev to secure it),  but it wouldn't hurt to show only the allowed schemas for the specific role in the error message, i.e.:

```bash
$ curl 'localhost:3000/todos' -H 'Accept-Profile: wrong'

{
  ...,
  "message": "The schema must be one of the following: api"
}
```

Adding the usage to the cached user profile and using it to filter the schemas should do it:

https://github.com/PostgREST/postgrest/blob/60d92f64dcab681ffc1674cb87935cb0a638603b/src/PostgREST/Config/Database.hs#L134-L135

steve-chavez:
I don't see harm in providing a bit of security through obscurity.

steve-chavez:
>   "message": "The schema must be one of the following: api, auth_api"

Thinking more about this, shouldn't the above message belong in the `hint` part? It seems it could be fitted. (`hint: only the api, auth_api schemas are exposed`).

Then solving this issue would be a matter of providing a config that disables hints on error messages. This would also solve https://github.com/PostgREST/postgrest/issues/3980#issuecomment-2761958774.

laurenceisla:
> Then solving this issue would be a matter of providing a config that disables hints on error messages.

Nice idea! Even though it's "yet another configuration parameter", at a simple glance I see that it could solve those issues more easily, so it's a +1 from me.

Southclaws:
I was looking for a way to completely disable this error message as for our use-case we have per-tenant schemas and this error, if exposed to clients, allows enumeration of schemas and thus identifiers associated with tenants.

For now we're catching this error at our proxy and writing back a stripped error to the caller, so not a huge issue but it could be worth a configuration value to do what the OP said as well as a way to disable exposing any information at all.

steve-chavez:
>   "message": "The schema must be one of the following: api, auth_api"

[...]

PR #4089, opened 2025-05-15
feat: improve error response when the requested schema is invalid

laurenceisla:
As mentioned in https://github.com/PostgREST/postgrest/issues/3824#issuecomment-2881562792

- It now shows the invalid schema in the "message"
- The exposed schemas are now listed in the "hint" instead of the "message"


src/PostgREST/Error.hs
```diff
@@ -191,7 +191,7 @@ instance ErrorBody ApiRequestError where
   message (InvalidBody errorMessage)   = T.decodeUtf8 errorMessage
   message (InvalidRange _)             = "Requested range not satisfiable"
   message InvalidFilters               = "Filters must include all and only primary key columns with 'eq' operators"
-  message (UnacceptableSchema schemas) = "The schema must be one of the following: " <> T.intercalate ", " schemas
+  message (UnacceptableSchema sch _)   = "Invalid schema: " <> sch
```
laurenceisla, line 194: I think it would improve the message by showing the invalid schema there.

steve-chavez (approved):

PR #4090, opened 2025-05-15
fix: PostgREST not failing at startup when db-schemas config is set to e…

taimoorzaeem:
…mpty

As discussed in [#4079 (comment)](https://github.com/PostgREST/postgrest/pull/4079#discussion_r2085385960).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix jwt error returning HTTP status `400` for invalid role by @taimoorzaeem in #3601
+- Fix PostgREST not failing to start when `db-schemas` config is set to empty by @taimoorzaeem in #4089
 
 ## [13.0.0] - 2025-05-08
 
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -272,8 +272,7 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
     <*> (fromMaybe True <$> optBool "db-prepared-statements")
     <*> (fmap toQi <$> optWithAlias (optString "db-root-spec")
                                     (optString "root-spec"))
-    <*> (fromList . maybe ["public"] splitOnCommas <$> optWithAlias (optString "db-schemas")
-                                                                    (optString "db-schema"))
+    <*> parseDbSchemas "db-schemas" "db-schema"
     <*> (fromMaybe True <$> optBool "db-config")
     <*> (fmap toQi <$> optString "db-pre-config")
     <*> parseTxEnd "db-tx-end" snd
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fix jwt error returning HTTP status `400` for invalid role by @taimoorzaeem in #3601
 
+### Changed
+
+- PostgREST now fails to start when `db-schemas` config is set to empty by @taimoorzaeem in #4089
```
steve-chavez, line 20: Hm, is this really a breaking change? We have this for [v13](https://github.com/PostgREST/postgrest/releases/tag/v13.0.0):

```
Fail to start on an invalid db-schemas or db-extra-search-path by @wolfgangwalther in https://github.com/PostgREST/postgrest/pull/3644
```

Isn't this a case of invalid `db-schemas`?

taimoorzaeem: > Isn't this a case of invalid `db-schemas`? If before we didn't fail on `<empty>`, then this should be a fix?

Hm, agreed. This is a fix.

src/PostgREST/Config.hs
```diff
@@ -395,6 +394,16 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
         Nothing   -> pure Nothing
         Just orig -> pure $ Just (T.strip <$> T.splitOn "," orig)
 
+    parseDbSchemas :: C.Key -> C.Key -> C.Parser C.Config (NonEmpty Text)
+    parseDbSchemas k al =
+      optWithAlias (optString k) (optString al) >>= \case
+        Nothing      -> pure $ fromList ["public"]
+        Just ""      -> pure $ fromList ["public"]
```
steve-chavez, line 401: Isn't this `db-schemas = ""`? Shouldn't it be a failure in this case?

taimoorzaeem: We were defaulting to `public` on just `""`. Failing on this required fixing too many io-test cases that started failing after this change. I'll recheck and confirm.

test/io/test_io.py
```diff
@@ -1896,3 +1896,19 @@ def test_invalidate_jwt_cache_when_secret_changes(tmp_path, defaultenv):
...
+    "postgrest should fail at startup with empty string in db-schemas config"
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_SCHEMA": "",
+    }
+
+    with run(env=env, no_startup_stdout=False, wait_for_readiness=False) as postgrest:
+        exitCode = wait_until_exit(postgrest)
+        assert exitCode == 1
```
taimoorzaeem, line 1911: Hm, the `wait_until_exit` function should exit in this case and return `1` but it actually throws a `PostgrestTimedOut` exception. Is this expected?

src/PostgREST/Config.hs
```diff
@@ -395,6 +394,13 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
         Nothing   -> pure Nothing
         Just orig -> pure $ Just (T.strip <$> T.splitOn "," orig)
 
+    parseDbSchemas :: C.Key -> C.Key -> C.Parser C.Config (NonEmpty Text)
+    parseDbSchemas k al =
+      optWithAlias (optString k) (optString al) >>= \case
+        Nothing      -> pure $ fromList ["public"]
+        Just ""      -> fail "Invalid db-schemas. Check your configuration."
+        Just schemas -> pure (fromList $ T.strip <$> T.splitOn "," schemas)
```
steve-chavez, line 402: @wolfgangwalther Can you help double check if this is the correct behavior? :pray:

taimoorzaeem:
Ah, so we are already filtering out `""` and return `Nothing` for empty cases here for all string configs. I think this is the root issue here:

https://github.com/PostgREST/postgrest/blob/520631913feccce45506016049be4df7d5589127/src/PostgREST/Config.hs#L404-L405

taimoorzaeem:
Closed as discussed in [#4095 (comment)](https://github.com/PostgREST/postgrest/issues/4095#issuecomment-2898937874).

PR #4094, opened 2025-05-17
docs: add note in JWT Role Extraction section

### LINKED ISSUE #4092: Docs should clarify that the JSONPath to be used `jwt-role-claim-key` is not RFC compliant
## Problem

> But, our string comparison feature is a custom feature which is not supported in the RFC.

> I always thought we were following the RFC, I see nothing on [jwt-role-claim-key](https://docs.postgrest.org/en/v13/references/auth.html#jwt-role-extract) that says we're not. Perhaps you can edit the docs to clarify that?

## Solution

Clarify that our JSONPath DSL follows a custom syntax and is not [RFC 9535](https://www.rfc-editor.org/rfc/rfc9535.html) compliant.

_Originally posted by @steve-chavez in https://github.com/PostgREST/postgrest/discussions/4065#discussioncomment-13149758_

taimoorzaeem:
Add a note describing that the used JSPath DSL does not strictly follow the JSONPath as described in RFC 9535.

Closes #4092.

### DIFF
--- a/docs/references/auth.rst
+++ b/docs/references/auth.rst
@@ -224,6 +224,9 @@ Usage examples:
     jwt-role-claim-key = ".postgrest.roles[?(@ ==^ \"hor\")]"
     jwt-role-claim-key = ".postgrest.roles[?(@ *== \"utho\")]"
 
+.. note::
+
+  The string comparison operators are implemented as a custom extension to the JSPath and does not strictly follow the `RFC 9535 <https://www.rfc-editor.org/rfc/rfc9535.html>`_.
 
 JWT Security
 ~~~~~~~~~~~~


docs/references/auth.rst
```diff
@@ -224,6 +224,9 @@ Usage examples:
     jwt-role-claim-key = ".postgrest.roles[?(@ ==^ \"hor\")]"
     jwt-role-claim-key = ".postgrest.roles[?(@ *== \"utho\")]"
 
+.. note::
+
+  The JSPath DSL used here is implemented as a custom feature and does not strictly follow the `RFC 9535 <https://www.rfc-editor.org/rfc/rfc9535.html>`_.
```
steve-chavez, line 229: Which feature here specifically does extend the RFC?

taimoorzaeem: The string comparison feature implemented in #3813. Let me mention that here.

PR #4104, opened 2025-05-27
fix: unexpected results when embedding the same table twice

### LINKED ISSUE #4075: Unexpected results when embedding and filtering the same table twice
### Environment

* PostgreSQL version: v17
* PostgREST version: v10.2.0 to v13.0.0
* Operating system: NixOS

### Description of issue

This happened in v10.2.0 but I could reproduce it in later versions including v13.0.0. Taking these tables and data:

```sql
create type visit_type as enum ('vacation', 'work');

create table places (
  id int primary key generated always as identity,
  name text not null
);

create table visits (
  id int primary key generated always as identity,
  place_id int not null references places(id),
  start_time timestamp,
  end_time timestamp,
  visit_type visit_type
);

insert into places (name) values ('Lake'), ('Mountain'), ('Beach');
insert into visits (place_id, start_time, end_time, visit_type)
values (1, '2025-01-01 10:00','2025-01-01 11:00', 'vacation'),
       (1, '2025-01-01 15:00','2025-01-01 16:00', 'vacation'),
       (1, '2025-01-01 20:00', '2025-01-01 21:00', 'work'),
       (2, '2024-11-01 09:00','2024-11-01 10:00', 'vacation'),
       (3, '2024-12-02 13:00','2024-12-02 14:00', 'vacation'),
       (1, '2023-01-02 20:00','2023-01-01 21:00', 'work');
```

If we take `places` and embed `visits` twice (one for work and others for non-work `visits`) we get the following response:

```bash
curl --get 'localhost:3030/places'\
  -d 'select=name,visits(*),work_visits:visits(*)' \
  -d 'id=eq.1' \
  -d 'visits.start_time=gt.20250101+00:00' \
  -d 'visits.end_time=lt.20250101+24:00' \
  -d 'visits.visit_type=neq.work' \
  -d 'work_visits.start_time=gt.20250101+00:00' \
  -d 'work_visits.end_time=lt.20250101+24:00' \
  -d 'work_visits.visit_type=eq.work'
```
```json
[
  {
    "name": "Lake",
    "work_visits": [],
    "visits": [
      {
        "id": 1,
        "end_time": "2025-01-01T11:00:00",
        "place_id": 1,
        "start_time": "2025-01-01T10:00:00",
        "visit_type": "vacation"
      },
      {
        "id": 2,
        "end_time": "2025-01-01T16:00:00",
        "place_id": 1,
        "start_time": "2025-01-01T15:00:00",
        "visit_type": "vacation"
      },
      {
        "id": 3,
        "end_time": "2025-01-01T21:00:00",
        "place_id": 1,
        "start_time": "2025-01-01T20:00:00",
        "visit_type": "work"
      },
      {
        "id": 6,
        "end_time": "2023-01-01T21:00:00",
        "place_id": 1,
        "start_time": "2023-01-02T20:00:00",
        "visit_type": "work"
      }
    ]
  }
]
```

The `work_vists` embed doesn't show anything (it should show the one with `"id": 3`), and the `visits` embed shows work visits and even those outside of the date range. There may be something wrong when building the query and assigning the filters to the joined tables.

<details><summary>SQL output</summary>
<p>

```sql
WITH pgrst_source AS
  (SELECT "public"."places"."name",
          COALESCE("places_work_visits_1"."places_work_visits_1", '[]') AS "work_visits",
          COALESCE("places_visits_1"."places_visits_1", '[]') AS "visits"
   FROM "public"."places"
   LEFT JOIN LATERAL
     (SELECT json_agg("places_work_visits_1")::jsonb AS "places_work_visits_1"
      FROM
        (SELECT "visits_1".*
         FROM "public"."visits" AS "visits_1"
         WHERE "visits_1"."start_time" > $1
           AND "visits_1"."end_time" < $2
           AND "visits_1"."visit_type" <> $3
           AND "visits_1"."start_time" > $4
           AND "visits_1"."end_time" < $5
           AND "visits_1"."visit_type" = $6
           AND "visits_1"."place_id" = "public"."places"."id") AS "places_work_visits_1") AS "places_work_visits_1" ON TRUE
   LEFT JOIN LATERAL
     (SELECT json_agg("places_visits_1")::jsonb AS "places_visits_1"
      FROM
        (SELECT "visits_1".*
         FROM "public"."visits" AS "visits_1"
         WHERE "visits_1"."place_id" = "public"."places"."id") AS "places_visits_1") AS "places_visits_1" ON TRUE
   WHERE "public"."places"."id" = $7)
       nullif(current_setting('response.status', TRUE), '') AS response_status,
       '' AS response_inserted
FROM
  (SELECT *
   FROM pgrst_source) _postgrest_t

-- [36659] DETAIL:  Parameters: $1 = '2025-01-01 00:00:00', $2 = '2025-01-02 00:00:00', $3 = 'work', $4 = '2025-01-01 00:00:00', $5 = '2025-01-02 00:00:00', $6 = 'work', $7 = '1'
```

</p>
</details> 

### Solution

Permanent solution is pending but the workaround is to use an alias for both tables. The alias must be different from the table name, e.g. using `place_visits:place_visits` still fails, but `non_work_visits:place_visits` will work as expected.

laurenceisla:
Closes #4075

~(In stand-by for a major release)~

### DIFF
--- a/test/spec/Feature/Query/QuerySpec.hs
+++ b/test/spec/Feature/Query/QuerySpec.hs
@@ -1070,6 +1070,16 @@ spec = do
             { "id":4,"children":[]}
           ]|] { matchHeaders = [matchContentTypeJson] }
 
+      it "works when embedding the same table more than once" $
+        get "/places?select=name,visits(id,start_time,visit_type),work_visits:visits(id,start_time,visit_type)&id=eq.1&visits.visit_type=neq.work&visits.start_time=gt.20250101+00:00&work_visits.visit_type=eq.work&work_visits.start_time=gt.20250101+00:00" `shouldRespondWith`
+          [json|[
+            {
+              "name":"Lake",
+              "visits":[{"id": 1, "start_time": "2025-01-01T10:00:00", "visit_type": "vacation"}, {"id": 2, "start_time": "2025-01-01T15:00:00", "visit_type": "vacation"}],
+              "work_visits":[{"id": 3, "start_time": "2025-01-01T20:00:00", "visit_type": "work"}]
+            }
+          ]|] { matchHeaders = [matchContentTypeJson] }
+
   describe "ordering response" $ do
     it "by a column asc" $
       get "/items?id=lte.2&order=id.asc"
@@ -1152,7 +1162,7 @@ spec = do
         { matchHeaders = [matchContentTypeJson] }
 
     it "ordering embeded entities with alias" $
-      get "/projects?id=eq.1&select=id, name, the_tasks:tasks(id, name)&tasks.order=name.asc" `shouldRespondWith`
+      get "/projects?id=eq.1&select=id, name, the_tasks:tasks(id, name)&the_tasks.order=name.asc" `shouldRespondWith`
         [json|[{"id":1,"name":"Windows 7","the_tasks":[{"id":2,"name":"Code w7"},{"id":1,"name":"Design w7"}]}]|]
         { matchHeaders = [matchContentTypeJson] }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file. From versio
 - Shutdown should wait for in flight requests by @mkleczek in #4702
 - Fix login with uppercase and mixed case role names by @taimoorzaeem in #4678
 - Remove automatic transaction retries on `40001 (serialization_failure)` errors to prevent replication lag by @laurenceisla in #3673
+- Fix unexpected results when embedding and filtering the same table more than once by @laurenceisla in #4075
 
 ### Changed
 
@@ -29,6 +30,8 @@ All notable changes to this project will be documented in this file. From versio
   + Now fails at startup. Prior to this, it failed with `PGRST205` on requests related to these schemas.
 - Build a static executable for aarch64-linux by @wolfgangwalther in #4193
 - Build the minimal docker image for aarch64-linux by @wolfgangwalther in #4193
+- The name of an embedded table can no longer be used in filters if it has an alias by @laurenceisla in #4075
+  + e.g. `?select=alias:table(*)&table.id=eq.1` is not possible anymore, use `?select=alias:table(*)&alias.id=eq.1` instead.
 
 ## [14.10] - 2026-04-16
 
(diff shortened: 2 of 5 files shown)


test/spec/Feature/Query/QuerySpec.hs
```diff
@@ -1138,7 +1148,7 @@ spec = do
         { matchHeaders = [matchContentTypeJson] }
 
     it "ordering embeded entities with alias" $
-      get "/projects?id=eq.1&select=id, name, the_tasks:tasks(id, name)&tasks.order=name.asc" `shouldRespondWith`
+      get "/projects?id=eq.1&select=id, name, the_tasks:tasks(id, name)&the_tasks.order=name.asc" `shouldRespondWith`
```
laurenceisla, line 1151: I didn't notice that we allow using the table name as filter even when the alias is present. This is the only test that mentions this, but I don't think this is expected, right?

From now on, this would fail with an error like:

```json
{
  "code":"PGRST108",
  "details":null,
  "hint":"Verify that 'tasks' is included in the 'select' query parameter.",
  "message":"'tasks' is not an embedded resource in this request"}
```

If this change is OK then more info to the hint would be needed like: "If it has an alias, then use it instead of the resource name". 

If using the resource name as well as the alias is expected, then this would be a breaking change and the solution would be different. Maybe throw an error when there may be ambiguity (like in #2529).

laurenceisla: > From now on, this would fail with an error like:

This would also fail for filters, limits, etc.

themavik (commented):
findNode now compares fromMaybe relName relAlias to targetNodeName so duplicate embeds of the same table resolve to distinct plan nodes. Updating the ordering spec to the_tasks.order matches alias-qualified order params; places/visits fixtures exercise the regression well.

wolfgangwalther:
> (In stand-by for a major release)

Since we only have major and patch releases by now, there is no need to hold this back. Also we already have some breaking changes on main anyway.

wolfgangwalther (approved):
Rebased to resolve the conflict in CHANGELOG.

steve-chavez:
@laurenceisla @wolfgangwalther This shouldn't haven been merged in this state. 

- This syntax is very old and the change very subtle.
- We have a big amount of deployments now that have lots of requests over their application logic.
- We can't expect users to hunt cases where this happens. It makes the migration very hard.

I agree the change should be done but gradually. I'll push a PR shortly. **Edit**: done https://github.com/PostgREST/postgrest/pull/4905

PR #4111, opened 2025-05-30
refactor: move jwt error messages to Error.hs module

taimoorzaeem:
Refactor as discussed in [#4101 (comment)](https://github.com/PostgREST/postgrest/pull/4101#discussion_r2109820699).


taimoorzaeem:
@steve-chavez WDYT? This is slightly different from what we discussed, but I think this is a better and more extensible design. I want to do the same for `JwtClaimsError`.

steve-chavez:
> @steve-chavez WDYT? This is slightly different from what we discussed, but I think this is a better and more extensible design. I want to do the same for JwtClaimsError.

@taimoorzaeem Looks good :+1:

PR #4114, opened 2025-05-30
fix: regression that makes order by with nulls order not work alongide limits

### LINKED ISSUE #4109: nullslast/nullsfirst not working on GET with pagination (offset/limit)
### Environment

* PostgreSQL version: PostgreSQL 14.17
* PostgREST version: v13.0.0

### Description of issue
After upgrading to PostgREST v13.0.0 GET requests that use both order=...nullsfirst/nullslast in combination with offset= and limit= return BadRequest responses. The same requests used to work as expected in version v12.2.12. 

Example Request on a local setup using (https://github.com/mattddowney/compose-postgrest)

http://localhost:3000/country?order=headofstate.desc.nullslast&limit=10
`{
    "code": "42601",
    "details": null,
    "hint": null,
    "message": "syntax error at or near \"NULLS\""
}`

Without either the .nullslast part or the limit= part the request returns the expected result.

laurenceisla:
Closes #4109


src/PostgREST/Query/SqlFragment.hs
```diff
@@ -501,7 +501,7 @@ returningF qi returnings =
 
 limitOffsetF :: NonnegRange -> SQL.Snippet
 limitOffsetF range =
-  if range == allRange then mempty else "LIMIT " <> limit <> " OFFSET " <> offset
+  if range == allRange then mempty else " LIMIT " <> limit <> " OFFSET " <> offset
```
steve-chavez, line 504: Oof, this is too subtle. How about using [unwords](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:unwords) instead?

We have that on https://github.com/PostgREST/postgrest/blob/0aa397fa1d5ded01633fe123efe17b3bdd19f63e/src/PostgREST/Query/SqlFragment.hs#L356-L358

src/PostgREST/Query/SqlFragment.hs
```diff
@@ -524,7 +524,7 @@ currentSettingF setting =
 
 orderF :: QualifiedIdentifier -> [CoercibleOrderTerm] -> SQL.Snippet
 orderF _ []    = mempty
-orderF qi ordts = "ORDER BY " <> intercalateSnippet ", " (pgFmtOrderTerm qi <$> ordts)
+orderF qi ordts = " ORDER BY " <> intercalateSnippet ", " (pgFmtOrderTerm qi <$> ordts)
```
steve-chavez, line 527: Ditto, `unwords`

test/spec/Feature/Query/RangeSpec.hs
```diff
@@ -223,6 +223,14 @@ spec = do
                              , "Content-Range" <:> "2-4/*" ]
             }
 
+      it "works alongside order by with nullsz order" $
```
steve-chavez, line 226:
```suggestion
      it "works alongside order by with nulls order" $
```

src/PostgREST/Query/SqlFragment.hs
laurenceisla: The unwords would need to be used at the QueryBuilder.hs level, since the separation is between `orderF` and `limitOffsetF` (both are SQL.Snippet)... Hmm... I can add the separation there to be more explicit, it was like that before:

```haskell
-- QueryBuilder.hs
  groupF qi select relSelect <> " " <>
  orderF qi order <> " " <>
  limitOffsetF readRange
  where
```

laurenceisla: Or do you mean having something like:

```suggestion
  if range == allRange then mempty else " " <> BS.unwords ["LIMIT", limit, "OFFSET", offset]
```

laurenceisla: Hmm... still not possible cause `limit` and `offset` are Snippet types. So, I think the only alternative to make the space more explicit would be this one https://github.com/PostgREST/postgrest/pull/4114/files#r2116852368.

steve-chavez: If you convert `" LIMIT "`and `" OFFSET " ` to `Snippet` with `SQL.sql` , won't `unwords` work?

I'm fine with the other alternative otherwise, it's just too dangerous to leave spaces like that.

laurenceisla: OK, couldn't be done, so I applied the alternative.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -53,8 +53,8 @@ readPlanToQuery node@(Node ReadPlan{select,from=mainQi,fromAlias,where_=logicFor
   (if null logicForest && null relJoinConds
     then mempty
     else " WHERE " <> intercalateSnippet " AND " (map (pgFmtLogicTree qi) logicForest ++ map pgFmtJoinCondition relJoinConds)) <> " " <>
-  groupF qi select relSelect <>
-  orderF qi order <>
+  groupF qi select relSelect <> " " <>
+  orderF qi order <> " " <>
```
steve-chavez, line 57: This is better actually. Not ideal but at least the spaces can be clearly seen.

steve-chavez (approved):
LGTM. @laurenceisla Could you do a patch release after merge? :pray:

laurenceisla:
> LGTM. @laurenceisla Could you do a patch release after merge? 🙏

Cool, will do.

Issue #4107, opened 2025-05-28
test: JWT Cache Memory usage grows quickly


steve-chavez:
## Problem

Doing:
```
$ postgrest-bench-k6 10 k6/GETSingleUniqueJWT.js 
```

And monitoring with:
```
watch -n 1 'ps -p $(pgrep postgrest) -o %mem,rss,vsz'
```

Makes the memory usage increase fast:
```
%MEM   RSS    VSZ
54.9 250496 1073793376
```

As seen on:

https://github.com/user-attachments/assets/6ebf899e-27ff-41ca-819d-2ed789f5aa25

## Solution

https://github.com/PostgREST/postgrest/pull/4084 would be a solution, I've manually tested there and memory usage only reaches 12% and is maintained there.

But we really need a way to test this. Try using `psrecord` to do it.

## Notes

- For some reason, using OPTIONS requests `k6/OPTIONSUniqueJWT.js` doesn't have the same memory effect as doing GET `k6/GETSingleUniqueJWT.js`. With OPTIONS the memory increases but quickly goes down and is maintained at ~8%.
- `psrecord` offered a nice `.png` image, but that's not possible to output on github actions summary. See  https://github.com/orgs/community/discussions/101814 and https://github.com/orgs/community/discussions/35932.
  + Although reading https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/ seems it's possible to have XY charts using mermaid. But these won't work locally?

PR #4118, opened 2025-06-02
test: Use RSA 4096 algorithm for JWT load test

mkleczek:
As requested in https://github.com/PostgREST/postgrest/pull/4084#issuecomment-2920351221

Note that after this PR `gen_targets` no longer generates JWTs with short expiration to test current implementation cache cleanup.
That's because generating JWTs takes considerable more time right now and generating 50k JWTs takes more than 2 minutes on my beefy MacBook Pro.

Not sure if it makes sense to merge it without changing the JWT cache to the bounded one (either LRU or Sieve).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -113,7 +113,7 @@ jobs:
   loadtest:
     strategy:
       matrix:
-        kind: ['mixed', 'jwt-hs-50k', 'jwt-rsa-10k']
+        kind: ['mixed', 'jwt-hs-50k', 'jwt-rsa-1k']
     name: Loadtest
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -113,7 +113,7 @@ jobs:
   loadtest:
     strategy:
       matrix:
-        kind: ['mixed', 'jwt-hs-50k']
+        kind: ['mixed', 'jwt-hs-50k', 'jwt-rsa-10k']
     name: Loadtest
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -113,7 +113,7 @@ jobs:
   loadtest:
     strategy:
       matrix:
-        kind: ['mixed', 'jwt']
+        kind: ['mixed', 'jwt-hs-50k']
     name: Loadtest
     runs-on: ubuntu-24.04
     steps:
--- a/nix/tools/generate_targets_rsa.py
+++ b/nix/tools/generate_targets_rsa.py
@@ -1,5 +1,6 @@
 # generates a file to be used by the vegeta load testing tool
-# It runs TOTAL_TARGETS amount of requests with TOTAL_JWTS amount of JWTs randomly spread among them 
+# It runs TOTAL_TARGETS amount of requests with
+# TOTAL_JWTS amount of JWTs randomly spread among them
 import time
 import argparse
 import sys
--- a/nix/tools/generate_targets_rsa.py
+++ b/nix/tools/generate_targets_rsa.py
@@ -1,0 +1,67 @@
+# generates a file to be used by the vegeta load testing tool
+import time
+import argparse
+import sys
+import random
+import jwt
+import jwcrypto.jwk as jwk
+
+SECRET = b"reallyreallyreallyreallyverysafe"
+URL = "http://postgrest"
+JWT_DURATION = 120
+TOTAL_TARGETS = 50000  # tuned by hand to reduce result variance
+
+key = jwk.JWK.generate(kty="RSA", size=4096)
+
+private_key = jwt.algorithms.RSAAlgorithm.from_jwk(key.export_private())
+public_key = key.export_public()
+
+
+def generate_jwt() -> str:
+    """Generate an RS256 JWT"""
+    now = int(time.time())
+    payload = {
+        "sub": f"user_{random.getrandbits(32)}",
+        "iat": now,
+        "role": "postgrest_test_author",
+    }
+
+    return jwt.encode(payload, private_key, "RS256")
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Generate Vegeta targets with unique JWTs"
+    )
+    parser.add_argument(
+        "output",
+        help="Path to write the generated targets file",
+    )
+    parser.add_argument("jwk", help="Path to write the generated JWK")
+    args = parser.parse_args()
+
+    tokens = [generate_jwt() for _ in range(1000)]
+    lines = []
+    start_time = time.time()
+
+    for i in range(TOTAL_TARGETS):
+        token = random.choice(tokens)
+        lines.append(f"OPTIONS {URL}/authors_only")
+        lines.append(f"Authorization: Bearer {token}")
+        lines.append("")  # blank line to separate requests
+
+    try:
+        with open(args.jwk, "w") as jwk, open(args.output, "w") as f:
+            jwk.write(public_key)
+            f.write("\n".join(lines))
+    except IOError as e:
+        print(f"Error writing to {args.output}: {e}", file=sys.stderr)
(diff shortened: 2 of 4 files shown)


steve-chavez:
@mkleczek Could you make this a new loadtest instead of changing the current one? :pray: 

While the current one was done to test the JWT purging, it's also a good test for the worse case of the JWT cache and https://github.com/PostgREST/postgrest/issues/4107 will ensure this case is memory bound (currently it isn't but with your https://github.com/PostgREST/postgrest/pull/4084 it is).

The above would be good for documentation too.

(Implementation can be a new arg on the python script or a new python script)

nix/tools/loadtest.nix
```diff
@@ -41,8 +41,10 @@ let
         args = [
           "ARG_OPTIONAL_SINGLE([output], [o], [Filename to dump json output to], [./loadtest/result.bin])"
           "ARG_OPTIONAL_SINGLE([testdir], [t], [Directory to load tests and fixtures from], [./test/load])"
-          "ARG_OPTIONAL_SINGLE([kind], [k], [Kind of loadtest (mixed: repeat mixed requests, jwt: run once over many requests with unique jwts)], [mixed])"
+          "ARG_OPTIONAL_SINGLE([kind], [k], [Kind of loadtest (mixed: repeat mixed requests, jwt: run once over many requests with 1000 different jwts)], [mixed])"
```
steve-chavez: Maybe the old `jwt` load test can be named as `jwt-cache-worse` and new one `jwt-cache` or just `jwt`.

mkleczek:
> @mkleczek Could you make this a new loadtest instead of changing the current one? 🙏
> 

Done

nix/tools/loadtest.nix
mkleczek: > Maybe the old `jwt` load test can be named as `jwt-cache-worse` and new one `jwt-cache` or just `jwt`.

I've decided to name them: `jwt-hs-50k` and `jwt-rsa-10k` to avoid any misunderstandings.

WDYT?

steve-chavez: Looks good :+1:

.github/workflows/test.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
   loadtest:
     strategy:
       matrix:
-        kind: ['mixed', 'jwt']
+        kind: ['mixed', 'jwt-hs-50k']
```
steve-chavez: Should also include the new load test right?
```suggestion
        kind: ['mixed', 'jwt-hs-50k', 'jwt-rsa-10k']
```

nix/tools/loadtest.nix
```diff
@@ -41,8 +41,10 @@ let
         args = [
           "ARG_OPTIONAL_SINGLE([output], [o], [Filename to dump json output to], [./loadtest/result.bin])"
           "ARG_OPTIONAL_SINGLE([testdir], [t], [Directory to load tests and fixtures from], [./test/load])"
-          "ARG_OPTIONAL_SINGLE([kind], [k], [Kind of loadtest (mixed: repeat mixed requests, jwt: run once over many requests with unique jwts)], [mixed])"
-          "ARG_TYPE_GROUP_SET([KIND], [KIND], [kind], [mixed,jwt])"
+          "ARG_OPTIONAL_SINGLE([kind], [k], [Kind of loadtest (mixed: repeat mixed requests, jwt: run once over many requests with 1000 different jwts)], [mixed])"
```
steve-chavez: Looks like the descriptions are no longer accurate, so I'd say just remove them:
```suggestion
          "ARG_OPTIONAL_SINGLE([kind], [k], [Kind of loadtest], [mixed])"
```

nix/tools/generate_targets_rsa.py
```diff
@@ -0,0 +1,67 @@
...
+import time
+import argparse
+import sys
+import random
+import jwt
+import jwcrypto.jwk as jwk
+
+SECRET = b"reallyreallyreallyreallyverysafe"
+URL = "http://postgrest"
+JWT_DURATION = 120
```
steve-chavez: This looks unused

nix/tools/generate_targets_rsa.py
```diff
@@ -0,0 +1,67 @@
...
+TOTAL_TARGETS = 50000  # tuned by hand to reduce result variance
```
steve-chavez, line 13: Should this be 10k? Going by the name `jwt-rsa-10k`
```suggestion
TOTAL_TARGETS = 10000  # tuned by hand to reduce result variance
```

nix/tools/generate_targets_rsa.py
```diff
@@ -0,0 +1,67 @@
+# generates a file to be used by the vegeta load testing tool
+import time
+import argparse
+import sys
+import random
+import jwt
+import jwcrypto.jwk as jwk
+
+SECRET = b"reallyreallyreallyreallyverysafe"
```
steve-chavez: Ditto. Unused.
```suggestion
```

nix/tools/generate_targets_rsa.py
steve-chavez: Just noticed this is happening on `nix/tools/generate_targets.py` too. Will add python dead code detection.

steve-chavez: Done in https://github.com/PostgREST/postgrest/pull/4125, if you like you can rebase and `postgrest-lint` will now show you the unused code.

mkleczek: Thanks and sorry for this copy/paste noise.

Fixed.

nix/tools/generate_targets_rsa.py
mkleczek: The idea is to run 50k requests but with 10k JWTs randomly spread among them. So we have TOTAL_TARGETS=50k and TOTAL_JWTS=10k

nix/tools/generate_targets_rsa.py
mkleczek: Fixed.

But the linter did not flag it for me, not sure why.

nix/tools/generate_targets_rsa.py
```diff
@@ -0,0 +1,66 @@
+# generates a file to be used by the vegeta load testing tool
+import time
+import argparse
+import sys
+import random
+import jwt
+import jwcrypto.jwk as jwk
+
+URL = "http://postgrest"
+TOTAL_JWTS = 1000
```
steve-chavez, line 12: This should be 10K right? Going from your [previous comment](https://github.com/PostgREST/postgrest/pull/4118/files#r2125613116).
```suggestion
TOTAL_JWTS = 10000
```

nix/tools/generate_targets_rsa.py
```diff
@@ -0,0 +1,66 @@
+# generates a file to be used by the vegeta load testing tool
```
steve-chavez, line 1:
```suggestion
# generates a file to be used by the vegeta load testing tool
# It runs TOTAL_TARGETS amount of requests with TOTAL_JWTS amount of JWTs randomly spread among them 
```

nix/tools/generate_targets_rsa.py
steve-chavez: Cool. Suggesting the above as [top file comment](https://github.com/PostgREST/postgrest/pull/4118/files#r2126796437) as clarification.

nix/tools/generate_targets_rsa.py
mkleczek: My bad - `jwt-rsa-10k` name is wrong. It is supposed to be 1k as suggested by you as the default size of JWT cache.

steve-chavez: Ok cool. Then we should update the `jwt-rsa-10k` loadtest name to `jwt-rsa-1k`.

nix/tools/generate_targets_rsa.py
steve-chavez: Looks like the comment was too long and now it's failing linting :/

PR #4119, opened 2025-06-02
fix: max-affected preference not failing for rpc with strict handling

### LINKED ISSUE #4100: max-affected preference should fail when handling=strict and RPC returns void or scalar
## Problem

As discussed in [#4096 (comment)](https://github.com/PostgREST/postgrest/issues/4096#issuecomment-2902121869), `max-affected` preference should fail when used in conjunction with `handling=strict` preference and the RPC returns void or a scalar value.

Currently it doesn't fail when:

```sql
create or replace function test.delete_items_returns_void() returns void as $$
  delete from items where id <= 15;
$$ language sql;
```

```sh
$ curl "localhost:3000/rpc/delete_items_returns_void" -X POST -H "Prefer: handling=strict, max-affected=10"
$
```

## Solution

It should fail when return type of RPC is void or scalar.

taimoorzaeem:
Closes #4100.

### DIFF
--- a/docs/references/api/preferences.rst
+++ b/docs/references/api/preferences.rst
@@ -292,3 +292,7 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
       "details": "The query affects 14 rows",
       "hint": null
   }
+
+.. note::
+
+  It is important for functions to return ``SETOF`` or ``TABLE`` when called with ``max-affected`` preference. A violation of this would cause a :ref:`PGRST128 <pgrst128>` error.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   + The exposed schemas are now listed in the `hint` instead of the `message` field.
 - Improve error details of `PGRST301` error by @taimoorzaeem in #4051
 
+### Fixed
+
+- Fix `max-affected` preference not failing with RPC when `handling=strict` by @taimoorzaeem in #4100
+
 ## [13.0.2] - 2025-06-02
 
 ### Fixed
--- a/docs/references/errors.rst
+++ b/docs/references/errors.rst
@@ -267,6 +267,10 @@ Related to the HTTP request elements.
 |               |             | implemented.                                                |
 | PGRST127      |             |                                                             |
 +---------------+-------------+-------------------------------------------------------------+
+| .. _pgrst128: | 400         | ``max-affected`` preference is violated with ``RPC`` call.  |
+|               |             | See :ref:`prefer_max_affected`.                             |
+| PGRST128      |             |                                                             |
++---------------+-------------+-------------------------------------------------------------+
 
 
 .. _pgrst2**:
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -98,6 +98,7 @@ data ApiRequestError
   | MaxAffectedViolationError Integer
   | InvalidResourcePath
   | OpenAPIDisabled
+  | MaxAffectedRpcViolation
   deriving Show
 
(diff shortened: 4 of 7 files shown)


docs/references/api/preferences.rst
```diff
@@ -292,3 +292,7 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
       "details": "The query affects 14 rows",
       "hint": null
   }
+
+.. note::
+
+  It is important for functions to return ``SETOF`` when called with ``max-affected`` preference. A violation of this would cause a :ref:`PGRST128 <pgrst128>` error.
```
laurenceisla: Just tested and this also works when the function returns `TABLE` , right? It should also be mentioned here and in the error message, I think.

taimoorzaeem: Yes, good catch! I should also probably add a test with it.

src/PostgREST/Query.hs
```diff
@@ -180,6 +181,7 @@ actionQuery (DbCall plan@CallReadPlan{..}) conf@AppConfig{..} apiReq@ApiRequest{
       optionalRollback conf apiReq
       failNotSingular crMedia resultSet
       failExceedsMaxAffectedPref (preferMaxAffected,preferHandling) resultSet
+      failMaxAffectedRpcReturnsSingle (preferMaxAffected,preferHandling) crProc
```
laurenceisla: Executing this:

```
curl 'localhost:3000/rpc/delete_items_returns_void' -H 'Prefer: max-affected=0, handling=strict'
```

Returns the previous error. 

```json
{"code":"PGRST124","details":"The query affects 1 rows","hint":null,"message":"Query result exceeds max-affected preference constraint"}
```

It may not be a realistic example, but to be consistent, it should return the new error. This could fix it:

```suggestion
      failMaxAffectedRpcReturnsSingle (preferMaxAffected,preferHandling) crProc
      failExceedsMaxAffectedPref (preferMaxAffected,preferHandling) resultSet
```

docs/references/api/preferences.rst
steve-chavez: +1 on a test, also note:

> RETURNS TABLE also implies RETURNS SETOF.
> https://www.postgresql.org/docs/current/sql-createfunction.html

src/PostgREST/Query.hs
```diff
@@ -252,6 +254,13 @@ failExceedsMaxAffectedPref (Just (PreferMaxAffected n), handling) RSStandard{rsQ
   lift SQL.condemn
   throwError $ Error.ApiRequestError . Error.MaxAffectedViolationError $ toInteger queryTotal
 
+-- | Fail max-affected pref when used with handling=strict but rpc returns scalar
+failMaxAffectedRpcReturnsSingle :: (Maybe PreferMaxAffected, Maybe PreferHandling) -> Routine -> DbHandler ()
+failMaxAffectedRpcReturnsSingle (Nothing,_) _ = pure ()
+failMaxAffectedRpcReturnsSingle (Just (PreferMaxAffected _), handling) rout = when (funcReturnsSingle rout && handling == Just Strict) $ do
+    lift SQL.condemn
```
steve-chavez: Q: It looks like this will start an empty transaction?

Since this failure only depends on the schema cache + header, there would be no need to start a transaction right?

steve-chavez: So it seems to me that this failure should be produced at the `Plan` module, not at the `Query` one.

laurenceisla: > So it seems to me that this failure should be produced at the `Plan` module, not at the `Query` one.

Right, doing this would avoid this inconsistency here too: https://github.com/PostgREST/postgrest/pull/4119#discussion_r2121643188

PR #4128, opened 2025-06-05
nix: add commitlint to lint commit messages

### LINKED ISSUE #4122: chore: validate conventional commits
## Problem

Sometimes commits can be merged without following  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and the commit messages can slip review.

Recent examples:

- https://github.com/PostgREST/postgrest/pull/4121#issuecomment-2931716050
- https://github.com/PostgREST/postgrest/pull/4116

## Solution

Enforce commits follow conventional commits.

We can do this with a git hook https://github.com/tapsellorg/conventional-commits-git-hook. 

Also there's https://github.com/amannn/action-semantic-pull-request for enforcing PR titles -- although is not strictly needed IMO.

## Notes

- Other tools: https://www.conventionalcommits.org/en/about/

taimoorzaeem:
Adds a new script `postgrest-commitlint` to validate the commit messages.

Closes #4122.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -37,7 +37,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          fetch-depth: 100  # fetch history (last 100 commits) instead of default shallow clone history
+          fetch-depth: 100  # fetch history (last 100 commits) instead of default shallow clone history, this is deemed enough for a PR history
       - name: Setup Nix Environment
         uses: ./.github/actions/setup-nix
         with:
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -30,3 +30,23 @@ jobs:
         run: postgrest-lint
       - name: Run style check (auto-format with `nix-shell --run postgrest-style`)
         run: postgrest-style-check
+
+  commit:
+    name: Commit
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 100  # fetch history (last 100 commits) instead of default shallow clone history
+      - name: Setup Nix Environment
+        uses: ./.github/actions/setup-nix
+        with:
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          tools: gitTools.commitCheck.bin
+      - name: Run commitlint (check locally with `nix-shell --run postgrest-commitlint`)
+        run: |
+          # Fetch target branch explicitly
+          git fetch origin ${{ github.base_ref }}
+
+          # Run commitlint
+          postgrest-commitlint --from origin/${{ github.base_ref }} --to HEAD
--- a/default.nix
+++ b/default.nix
@@ -119,6 +119,10 @@ rec {
   docs =
     pkgs.callPackage nix/tools/docs.nix { };
 
+  # Git tools.
+  gitTools =
+    pkgs.callPackage nix/tools/gitTools.nix { };
+
   # Load testing tools.
   loadtest =
     pkgs.callPackage nix/tools/loadtest.nix { inherit withTools; };
--- a/nix/tools/gitTools.nix
+++ b/nix/tools/gitTools.nix
@@ -1,0 +1,60 @@
+{ buildToolbox
+, checkedShellScript
+, commitlint
+, writeText
+}:
+let
+  # Rules format: [<severity>, <"always"/"never">, <value>]
+  commitlintConfig = writeText "commitlint.config.mjs" ''
+    export default {
+      rules: {
+        "type-enum": [2, "always", [
+            'add',      // Add a new feature
+            'amend',    // To amend an unrealease commit
+            'change',   // Breaking changes
+            'chore',    // Update sponsors, changelog, readme etc
+            'ci',       // CI configuration files and scripts
+            'docs',     // Documentation
+            'fix',      // Bug fix
+            'nix',      // Related to Nix
+            'perf',     // Performance improvements
+            'refactor', // Refactoring code
+            'remove',   // Remove a feature or fix
+            'test',     // Adding tests
+          ]],
+
+          'subject-case':       [2, 'always', ['lower-case','sentence-case']],
+          'subject-empty':      [2, 'never'],
+          'subject-full-stop':  [2, 'never', '.'],
+          'subject-max-length': [2, 'always', 80],
+          'subject-min-length': [2, 'always', 5],
+
+          'scope-case':         [2, 'always', 'lower-case'],
+
+          'body-leading-blank': [2, 'always'],
+      },
+    };
+  '';
+
+  commitCheck =
+    checkedShellScript
+      {
+        name = "postgrest-commitlint";
+        docs = "Script to validate commit messages";
+        workingDir = "/";
+        args = [
+          "ARG_OPTIONAL_SINGLE([from],, [commit ref start from], [main])"
+          "ARG_OPTIONAL_SINGLE([to],, [commit ref end at], [HEAD])"
+        ];
+      }
+      ''
+        # Run commitlint with the given configuration
+
+        ${commitlint}/bin/commitlint --config ${commitlintConfig} --from "$_arg_from" --to "$_arg_to"
+      '';
+in
+buildToolbox
+{
+  name = "postgrest-commitlint";
+  tools = { inherit commitCheck; };
+}
--- a/shell.nix
+++ b/shell.nix
@@ -21,6 +21,7 @@ let
       postgrest.cabalTools
       postgrest.devTools
       postgrest.docs
+      postgrest.gitTools
       postgrest.loadtest
       postgrest.release
       postgrest.style


nix/tools/commit_msg_check.py
```diff
@@ -0,0 +1,65 @@
...
+        "docs",
+        "refactor",
+        "chore",
+        "nix",
+        "ci",
+        "perf",
+        "revert",
+        "break",
+        "build",
+    }
```
taimoorzaeem: Am I missing any prefixes?

laurenceisla: * The dependencies updates using the Renovate bot use `chore(deps)`.
* I've also seen `deprecate`, `drop`, `correct`, `changelog` (although this is sometimes replaced by `chore`)
* We had some flexibility with adding new prefixes (e.g. `correct`). If we wanted to use a new prefix we'd need to add it here first, right? IMO that's fine.
* Just to mention, this attempt to make the dev remember to use the prefixes wasn't effective in the end 🫠 : https://github.com/PostgREST/postgrest/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1

taimoorzaeem: > * The dependencies updates using the Renovate bot use `chore(deps)`.

This is allowed by the regex actually. For example `fix(config)` is allowed. This only checks the prefix before the bracketed part. The brackets can contain anything so, anything all lowercase is allowed in `(...)`.

> * I've also seen `deprecate`, `drop`, `correct`, `changelog` (although this is sometimes replaced by `chore`)

Oh yes, Let me add these.

> * We had some flexibility with adding new prefixes (e.g. `correct`). If we wanted to use a new prefix we'd need to add it here first, right? IMO that's fine.

Yes, if a new one is necessary, we just add here.

> * Just to mention, this attempt to make the dev remember to use the prefixes wasn't effective in the end 🫠 : https://github.com/PostgREST/postgrest/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1

That's okay. Now, with this script it will be enforced.

.github/workflows/check.yaml
```diff
@@ -30,3 +30,16 @@ jobs:
...
+  commit-check:
+    name: Commit Message Check
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Setup Nix Environment
+        uses: ./.github/actions/setup-nix
+        with:
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          tools: commit.check.bin
```
taimoorzaeem: I guess we need to update the `CACHIX_AUTH_TOKEN` to support this new workflow.

nix/tools/commit_msg_check.py
laurenceisla: > This is allowed by the regex actually.

Oh that's cool! Sorry, I didn't check the regex.

As a side note, having them in alphabetical order would be nicer.

wolfgangwalther: > I've also seen `deprecate`, `drop`, `correct`, `changelog` (although this is sometimes replaced by `chore`)

"have seen" is the wrong argument to put them into the list, imho. We've been all over the place, precisely because we didn't have linting. So we should standardize and allow fewer options.

.github/workflows/check.yaml
wolfgangwalther: Why? In which way would we need to update it?

nix/tools/devTools.nix
```diff
@@ -382,13 +383,27 @@ let
...
+
+  commitCheck =
+    checkedShellScript
+      {
+        name = "postgrest-commit-msg-check";
+        docs = "Validate latest commit message using the python script.";
+        workingDir = "/";
+      }
+      ''
+        git log -1 --pretty=%s | ${commitCheckScript}
```
wolfgangwalther: This only validates the latest commit, which is not nearly enough in a multi-commit PR.

We should use `commitlint`, which will handle this properly.

nix/README.md
```diff
@@ -98,7 +98,7 @@ postgrest-loadtest                postgrest-with-postgresql-16
 postgrest-loadtest-against        postgrest-with-postgresql-17
 postgrest-loadtest-report         postgrest-with-slow-pg
 postgrest-nixpkgs-upgrade         postgrest-with-slow-postgrest
-postgrest-parallel-curl
+postgrest-parallel-curl           postgrest-commit-msg-check
```
wolfgangwalther: There's two options here:
- Don't adjust the output and accept that it's outdated - it will always be a little bit.
- Adjust it to match the actual output of tab completion, which is alphabetically sorted.

But just adding at the end is not an option.

.github/workflows/check.yaml
taimoorzaeem: The CI addition of this is failing with `error: attribute 'commit' in selection path 'commit.check.bin' not found`. So, I made a wild guess that it might related to it, but I am not sure.

nix/README.md
taimoorzaeem: > * Don't adjust the output and accept that it's outdated - it will always be a little bit.

I am cool with this. Let's skip it here and let it be outdated. Later we can update it in a separate PR.

nix/tools/commit_msg_check.py
laurenceisla: > "have seen" is the wrong argument to put them into the list, imho

Fair. From the whole list, these have some inconsistencies (after checking `git log --grep '<type>:'`):

* `drop` and `break`: Most `break`s are used for dropping functionality right now. So some solutions would be:
  * Use `drop` instead and just use `BREAKING CHANGE` in the commit message.
  * Do not use any of them and instead use `fix` or `feat` (or other?). Still the `BREAKING CHANGE` would need to be present.
  * Both of the above imply removing `break`. I think we mentioned somewhere that `BREAKING CHANGE` is enough to imply a major release change and that the commit prefix should be `fix`, `feat` or whichever describes it the best.
*  `changelog` and `chore`: Sometimes `chore` is also used when updating the `changelog`. If we want to keep fewer options, we could just use `chore` here.

I'm in favor of keeping:
* `deprecate`: There are only a couple of commits with this. However I don't see if they can be included in other categories (maybe `feat`?).
* `correct`: When we're "fixing" an unreleased commit (could be `fix`, `feat`, etc.). We didn't know exactly what to put when this happened, so this should be a good option (see https://github.com/PostgREST/postgrest/pull/4043#issuecomment-2828277376).

I think the rest are fine as they are. Adding a description to each one of them would be nice too (maybe `commitlint` allows this?).

laurenceisla: Just noticed that we're not taking advantage of scopes, those could help with the inconsistencies above, like:

```
feat(drop): ...
chore(changelog): ...
```

Perhaps correct could be part of `fix`? Not sure, WDYT? CC. @steve-chavez

```
fix(correct): ...
```

wolfgangwalther: I'd say `drop` is not a scope, but a type (if we want to use it). `changelog`, `ci`, `docs`, `nix` could all be scopes, if we wanted them to be. But, I'd only use any of them with `chore(...)`, so maybe it's not useful.

A few thoughts:
- I don't think we need to stick with whatever "conventional commits" do. We can use our own prefixes, if we want to.
- It would make sense to have consistency between the commit prefix and two things:
  - versioning: Certain commits cause certain types of bumps.
  - changelog: There should be a clear 1:1 mapping between sections in the changelog and the prefix.

Thus, I propose the following:
- `add:` causes minor/major bump and is in the "Added" section in the changelog.
- `remove:` or `drop:` causes minor/major bump and is in the "Removed" section in the changelog.
- `change:` causes minor/major bump and is in the "Changed" section in the changelog.
- `fix:` causes patch bump and is in the "Fixed" section in the changelog. Also backported, if possible.
- `docs:` causes no bump and no changelog, but is backported if possible.
- `refactor:` causes no bump, no changelog and only backported if low-risk and required for a fix.
- `ci:` causes no bump, no changelog. Smaller changes might be backported, bigger ones are likely not to.
- `nix:` causes no bump, no changelog. Mostly not backported, but could be.
- `chore:` causes no bump, no changelog. Could be backported, could not be - depends on what it is.
- `Revert ...` (default style by git) is only allowed for commits, which were not released, yet and will take this commit out of version bumping and changelog considerations.

Not sure about `perf:`. So far, we listed them in the "Added" section in the changelog, which doesn't make much sense. If we want to have that prefix, we should give it a separate section in the changelog. But backport-ability and version bumping is not entirely clear on those. Depends on the risk-factor.

Also what was previously termed `correct:` is still unclear to me. Aka something that will improve/extend/fix something that has not been released, yet, and will not cause another changelog entry. "correct" might not always be right. Not exactly sure how to best group this.

wolfgangwalther: > Adding a description to each one of them would be nice too (maybe `commitlint` allows this?).

Descriptions could be put into "Prompts": https://commitlint.js.org/reference/prompt.html

I think those would allow a step-by-step interactive selection process of writing the commit message. Never used that, though.

.github/workflows/check.yaml
wolfgangwalther: The cachix token only affects caching, so it would only make a difference in terms of how long CI runs.

The error you are getting is because `commit.check.bin` is not correct. You added the tool in `devTools.commitCheck`, so this should be `devTools.commitCheck.bin`.

nix/tools/commit_msg_check.py
taimoorzaeem: > `chore:` causes no bump, no changelog. Could be backported, could not be - depends on what it is.

I think same goes for `test: ...` (no bump, no changelog, backport is unlikely but could be - it depends).

taimoorzaeem:
`postgrest-commitlint` script works fine locally but fails in CI with:
```
Error: fatal: ambiguous argument 'origin/main..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```
Not sure why this is coming up. Maybe on CI we need to separate it differently?

nix/tools/commitlint.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+  // extends: ["@commitlint/config-conventional"],
+
+  /*
+   * Any rules defined here will override rules from @commitlint/config-conventional.
+   * However we are not using @commitlint/config-conventional until above
+   * mentioned issue is resolved.
+   */
+  rules: {
+    "type-enum": [2, "always", [
+        'add',      // Add a new feature
```
steve-chavez: Should this be `feat`? IIRC we've never used `add`

nix/tools/commitlint.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+        'chore',    // Update sponsors, changelog, readme etc
+        'ci',       // CI configuration files and scripts
+        'docs',     // Documentation
+        'drop',     // Drop a feature or fix
+        'fix',      // Bug fix
+        'nix',      // Related to Nix
+        'perf',     // Performance improvements
+        'refactor', // Refactoring code
+        'remove',   // Remove a feature or fix
```
steve-chavez: What's the difference with this one and `drop`?

steve-chavez: If no difference, maybe it should be removed.

nix/tools/commitlint.config.mjs
wolfgangwalther: This comes from https://github.com/PostgREST/postgrest/pull/4128#discussion_r2134105883.

nix/tools/commitlint.config.mjs
wolfgangwalther: I'd drop `drop` instead of removing `remove`, because we use the term "removed" in the changelog, too?

wolfgangwalther:
> `postgrest-commitlint` script works fine locally but fails in CI with:
> 
> ```
> Error: fatal: ambiguous argument 'origin/main..HEAD': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> ```
> 
> Not sure why this is coming up. Maybe on CI we need to separate it differently?

This is very likely because the default `checkout` is shallow in CI. You'll need to fetch enough commits for commitlint to be able to actually work on them. Also it needs to fetch the target branch to compare against, too.

nix/tools/commitlint.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+        'revert',   // Reverting previous commits
```
wolfgangwalther: I think commitlint has support for "native" reverts, which don't follow the conventional commit pattern. Thus, I don't think we need revert as a type.

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,23 @@
+{ buildToolbox
+, checkedShellScript
+, commitlint
+}:
+let
+  commitLint =
```
wolfgangwalther: I really don't like the `commitlint` vs `commitLint` here. Too easy to get wrong.

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,23 @@
...
+    checkedShellScript
+      {
+        name = "postgrest-commitlint";
+        docs = "Lint commit";
+        workingDir = "/";
+      }
+      ''
+        # Run commitlint with the given configuration
+
+        ${commitlint}/bin/commitlint --config nix/tools/commitlint.config.mjs --from origin/main --to HEAD
```
wolfgangwalther: It's better to take the config from the nix store instead of relying on it from the checkout. That's because the tools should always be fully functional inside a nix-shell session.

Example:
- You'll start your nix-shell on main.
- Then switch to a branch which modifies commitlint and it's config.

In this state, `postgrest-commitlint` will take the code from main, but the config from the other branch. It should always take the full tool from the same origin.

You can either do:

```suggestion
        ${commitlint}/bin/commitlint --config ${./nix/tools/commitlint.config.mjs} --from origin/main --to HEAD
```

or (better imho) just inline the config into `gitTools.nix` like we do elsewhere with a `writeText`. I find it easier to navigate when everything for commitlint is in one file / place.

nix/tools/commitlint.config.mjs
```diff
@@ -0,0 +1,64 @@
+const Configuration = {
```
wolfgangwalther:
```suggestion
export default {
```

No need for the indirection.

nix/tools/commitlint.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+  /*
+   * Resolve and load @commitlint/config-conventional from node_modules.
+   * Referenced packages must be installed
+   */
+
+  // TODO: Not available yet. See https://github.com/NixOS/nixpkgs/issues/250767
+  // extends: ["@commitlint/config-conventional"],
```
wolfgangwalther: Do we want that base or define everything ourselves? Which rules would the default bring? Those commented out below?

.github/workflows/check.yaml
```diff
@@ -30,3 +30,16 @@ jobs:
         run: postgrest-lint
       - name: Run style check (auto-format with `nix-shell --run postgrest-style`)
         run: postgrest-style-check
+
+  commitlint:
+    name: Commit Lint
```
wolfgangwalther: The term "Commit Lint" is never used upstream. The tool's name is `commitlint` in the docs, too.

nix/tools/commit_msg_check.py
steve-chavez: I agree with the [above rules](https://github.com/PostgREST/postgrest/pull/4128#discussion_r2134105883), we should document those (on `nix/tools/commitlint.config.mjs` looks fine for now), maybe verbatim even. Also mentioning some of our previous prefixes like `feat` which are now changed to `add`. Otherwise it will be confusing (new contributor bases its commit prefix based on history).

default.nix
```diff
@@ -131,6 +131,10 @@ rec {
   style =
     pkgs.callPackage nix/tools/style.nix { inherit hsie; };
 
+  # Commit lint
```
wolfgangwalther: Commit Lint, Commit lint, commitlint, commitLint? :D

shell.nix
```diff
@@ -24,6 +24,7 @@ let
       postgrest.loadtest
       postgrest.release
       postgrest.style
+      postgrest.gitTools
```
wolfgangwalther: What appeared as a sorted list previously to me, is not anymore :D

nix/tools/commit_msg_check.py
steve-chavez: An extra nice thing would be: "if a commit with `feat:` prefix comes, fail and suggest using `add:` instead". Maybe that's easy with `commitlint`, if not maybe not worth it.

> Just noticed that we're not taking advantage of scopes

Scopes could be useful for release notes to automatically add a subsection (see [v13.0.0](https://github.com/PostgREST/postgrest/releases/tag/v13.0.0), subsections were done manually).

So if we have `add(x):`, x would refer to a docs' [custom anchor](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-a-custom-anchor) (which is unique) and get its text. e.g. `add(admin):` would add `Admin Server` subsection in release notes. `x` is usually short too, so it seems fit for a commit message.

Not something that should be done here, but an idea for later.

steve-chavez: > Also what was previously termed correct: is still unclear to me. Aka something that will improve/extend/fix something that has not been released, yet, and will not cause another changelog entry.

Maybe we could use a `fix(unreleased):` instead of `correct:`? 

**Edit**: although that name doesn't convey it shouldn't be added in the CHANGELOG. Not sure too.

wolfgangwalther: > although that name doesn't convey it shouldn't be added in the CHANGELOG.

Yeah. And it doesn't really need to be a fix either. We merge a feature, realize something is missing and add to it. But don't need a new changelog entry.

steve-chavez: Right, so it could be an addition (feature) as well. Maybe something like `amend:` would be more descriptive.

wolfgangwalther: Yeah. `amend(<short-git-hash>): ...` to improve an unreleased commit (short-git-hash) without a new changelog entry.

wolfgangwalther:
There is also `committed` as a tool. Just happened to accidentally find it. According to https://github.com/crate-ci/committed/blob/master/docs/reference.md it seems to have sane defaults (to me) and also supports the "imperative mood" check. And it's a compiled tool, so maybe less heavy to pull in compared to a node project with dependencies...

Certainly worth a try, I guess?

Edit: I just learned about https://jorisroovers.com/gitlint/latest/ as well, but this doesn't seem to have "imperative mood" rules.

default.nix
taimoorzaeem: These are good observations. I'll fix them once we finalize on the tool. :)

taimoorzaeem:
> There is also `committed` as a tool. Just happened to accidentally find it. According to https://github.com/crate-ci/committed/blob/master/docs/reference.md it seems to have sane defaults (to me) and also supports the "imperative mood" check. And it's a compiled tool, so maybe less heavy to pull in compared to a node project with dependencies...
> 
> Certainly worth a try, I guess?

Yes, certainly worth a try, especially the imperative mood check feature.

taimoorzaeem:
> Certainly worth a try, I guess?

I tried `committed` and although it may be a great tool, I find it hard to work with because it is quite poorly documented (not even an example of a config file). I think it is most useful as library and also documented that way. I think `commitlint` may be our best option for now.

.github/actions/setup-nix/action.yaml
```diff
@@ -7,6 +7,8 @@ inputs:
     description: Token to pass to cachix
   tools:
     description: Tools to install with nix-env -iA <tools>
+  fetch-depth:
+    description: Fetch all commit history instead of default shallow clone history
```
wolfgangwalther: You added an input to the `setup-nix` action, but are not using it. So this won't have any effect.

But it shouldn't - because that's the wrong file. fetch-depth makes no sense for nix - you will need it for `actions/checkout`!

.github/workflows/check.yaml
```diff
@@ -30,3 +30,22 @@ jobs:
...
+    name: Commit
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Setup Nix Environment
+        uses: ./.github/actions/setup-nix
+        with:
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          tools: gitTools.commitCheck.bin
+          fetch-depth: 0   # fetch full history instead of shallow
```
wolfgangwalther: You'll need that here instead:

```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
        with:
          fetch-depth: 0 # fetch full history instead of shallow
      - name: Setup Nix Environment
        uses: ./.github/actions/setup-nix
        with:
          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
          tools: gitTools.commitCheck.bin
```

But maybe we can just use a reasonably high default value instead of `0`? Like.. would anybody ever create a PR with 100 commits? Probably not, so setting it to 100 would still save massively on checkout.

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,75 @@
...
+  commitlintConfig = writeText "commitlint.config.mjs" ''
+    export default {
+
+      /* Rules applied to commit messages */
+
+      rules: {
+
+        /* Rules format: [<severity>, <"always"/"never">, <value>] */
+
+        "type-enum": [2, "always", [
```
wolfgangwalther: In Nix, it's generally better to move comments out of strings and into nix code, if possible. That will avoid useless rebuilds. So here, I'd do:

```suggestion
  # Rules format: [<severity>, <"always"/"never">, <value>]
  commitlintConfig = writeText "commitlint.config.mjs" ''
    export default {
      rules: {
        "type-enum": [2, "always", [
```

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,75 @@
...
+            'add',      // Add a new feature
+            'chore',    // Update sponsors, changelog, readme etc
+            'ci',       // CI configuration files and scripts
+            'docs',     // Documentation
+            'drop',     // Drop a feature or fix
+            'fix',      // Bug fix
+            'nix',      // Related to Nix
+            'perf',     // Performance improvements
+            'refactor', // Refactoring code
+            'remove',   // Remove a feature or fix
```
wolfgangwalther: You still have drop *and* remove, where we only need one (see some previous comments).

Also we had `change` in that thread, which is not here. And we discussed `amend` in https://github.com/PostgREST/postgrest/pull/4128#discussion_r2136390637.

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,75 @@
...
+            'test',     // Adding tests
+          ]],
+
+          'subject-case':       [2, 'always', 'lower-case'],
+          'subject-empty':      [2, 'never'],
+          'subject-full-stop':  [2, 'never', '.'],
+          'subject-max-length': [2, 'always', 60],
+          'subject-min-length': [2, 'always', 2],
```
wolfgangwalther: I'm not 100% sure on the max-length restrictions.. maybe we should look at the existing commit messages we have written, so far? What does the data say?

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,75 @@
...
+
+
+          'body-leading-blank': [1, 'always'],
```
wolfgangwalther: It's funny that there are *two* leading blanks lines before this rule :D

```suggestion

          'body-leading-blank': [1, 'always'],
```

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,75 @@
...
+          //'header-min-length':      [2, 'always', 5],
+          //'body-max-line-length':   [2, 'always', 100],
+          //'body-case':              [1, 'always', 'sentence-case'],
+          //'footer-leading-blank':   [1, 'always'],
+          //'footer-max-line-length': [2, 'always', 100],
+          //'scope-case':             [2, 'always', 'lower-case'],
+          //'scope-max-length':       [2, 'always', 20],
+          //'scope-empty':            [1, 'never'],
+          //'references-empty':       [1, 'never'],
+          //'signed-off-by':          [0, 'always', 'Signed-off-by:']]
```
wolfgangwalther: What about those? Are those defaults from somewhere? Or ideas/suggestions?

.github/workflows/check.yaml
```diff
@@ -30,3 +30,22 @@ jobs:
...
+      - name: Run commitlint (check locally with `nix-shell --run postgrest-commitlint`)
+        run: |
+          # Fetch target branch explicitly
+          git fetch origin main
```
wolfgangwalther: If we do it like this, not sure whether we still need the fetch-depth at all. Did you test that with multiple commits in CI already? If it worked, then the fetch-depth stuff can go away entirely.

.github/workflows/check.yaml
```diff
@@ -30,3 +30,22 @@ jobs:
...
+
+          # Run commitlint with default branches
+          postgrest-commitlint
```
wolfgangwalther: I think we should be smarter about the branches to select. Not every PR targets `main`. So we should use the base branch from GitHub's context here.

nix/tools/gitTools.nix
taimoorzaeem: Yes, these are suggestions commented out, in case we need any of them in future. It would be easy to just "uncomment" them instead of looking up the documentation. We'd probably not be needing most of them I think. Should I just remove them?

nix/tools/gitTools.nix
taimoorzaeem: We've had a [commit](https://github.com/PostgREST/postgrest/commit/de848f64fa3c423b80af36c37733fdfbfc2ab7ec) with 219 characters long subject message. I would say about any length between 80-100 characters seems reasonable to me for the subject line. For the body of commit message, we have this [commit](https://github.com/PostgREST/postgrest/commit/35d36d95c1c4192d99fee19f2cfdffaa92498fe6) which 983 characters (way too long IMO). Not sure what would be reasonable length for this.

nix/tools/gitTools.nix
wolfgangwalther: If we don't want to do them right now, we should remove them. No need to carry dead code around.

nix/tools/gitTools.nix
wolfgangwalther: I looked through the very recent history manually and all subject lines longer than 80 characters look like they could easily have been shortened for clarity.

(80 characters was the size of my git UI's commit list by chance)

Roughly in the 60-70 range I find some commits that might not be too easy to shorten.

So I'm fine with 80, which gives us a bit of room, we don't want this rule to always interfere.

.github/workflows/check.yaml
taimoorzaeem: I tried, it only works out if we do both `fetch-depth` and `git fetch`. Not sure why this is the case.

wolfgangwalther: `fetch-depth` will fetch the history of the current branch. `git fetch` will fetch the other branch. You need both to compare them.

nix/tools/gitTools.nix
```diff
@@ -0,0 +1,60 @@
...
+  '';
+
+  commitCheck =
+    checkedShellScript
+      {
+        name = "postgrest-commitlint";
+        docs = "Script to validate commit messages";
+        workingDir = "/";
+        args = [
+          "ARG_OPTIONAL_SINGLE([from],, [commit ref start from], [origin/main])"
```
taimoorzaeem:
```suggestion
          "ARG_OPTIONAL_SINGLE([from],, [commit ref start from], [origin/12])"
```
Are we backporting this? If yes then `origin/main` shouldn't be our default right? It could be `v12` or `v13` etc.

wolfgangwalther: I'm not concerned about the `main` part, but about the `origin` part. I have two remotes configured locally, `origin` (my fork) and `upstream` (PostgREST repo). My fork's main branch is never up to date, so `origin/main` wouldn't work. My local `main` branch tracks the `upstream` remote.

So just `[main]` would be much better, I think.

Even if we're backporting, I wouldn't change the default. The most likely case is to work on something targeting main locally. CI already handles it in every case.

.github/workflows/check.yaml
```diff
@@ -30,3 +30,23 @@ jobs:
...
       - name: Run style check (auto-format with `nix-shell --run postgrest-style`)
         run: postgrest-style-check
+
+  commit:
+    name: Commit
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 100  # fetch history (last 100 commits) instead of default shallow clone history
```
steve-chavez: This is unclear, should it say the following?
```suggestion
          fetch-depth: 100  # fetch history (last 100 commits) instead of default shallow clone history, this is deemed enough for a PR history
```

steve-chavez:
I think this should be good to go. Any opposition towards merging it now?

steve-chavez (approved):
Great work! :100:

steve-chavez:
Looks the commit linter is failing now because of the commit I made through the github UI https://github.com/PostgREST/postgrest/pull/4128/commits/e681dd6b6e52deae7f2a858f9c27d3d94ddaa12d. This is fine, it should be gone once I squash it into main.

steve-chavez:
Hm, there was a new failure on main: https://github.com/PostgREST/postgrest/actions/runs/15718198849/job/44293156081

taimoorzaeem:
> Hm, there was a new failure on main: https://github.com/PostgREST/postgrest/actions/runs/15718198849/job/44293156081

> Error: fatal: ambiguous argument 'origin/..HEAD': unknown revision or path not in the working tree

I guess the issue occurred here:

> postgrest-commitlint --from origin/${{ github.base_ref }} --to HEAD

`${{ github.base_ref }}` return empty after the merge.

For push events, it should be `${{ github.ref_name }}`, so we should conditionally handle this with: `${{ github.base_ref || github.ref_name }}`.

PR #4136, opened 2025-06-13
fix: regression that breaks fts on domain types based on tsvector

### LINKED ISSUE #4135: Domain type wrapping tsvector breaks full-text search (fts) in PostgREST v13
### Environment

PostgreSQL version: 17.5
PostgREST version: 13.0.2
OS: Ubuntu 22.04 (Dockerized setup)

Regression Introduced In: [#3845](https://github.com/PostgREST/postgrest/pull/3845)
### Description of issue

PostgREST v13 no longer supports domain types that wrap tsvector fields when using full-text search operators like fts.

Previously, this worked correctly in PostgREST 12 and below.

Example Schema
```sql
CREATE DOMAIN public.market_tsvector AS tsvector
  CONSTRAINT "tsvector is required" CHECK (VALUE IS NOT NULL);

CREATE TABLE public.markets (
  id UUID PRIMARY KEY,
  name TEXT,
  tsvector public.market_tsvector
);

```

```curl
GET /api/markets?select=id,name&tsvector=fts.asda
```

```json
{
  "code": "42883",
  "details": null,
  "hint": "No function matches the given name and argument types. You might need to add explicit type casts.",
  "message": "function to_tsvector(public.market_tsvector) does not exist"
}
```

## Expected Behavior

PostgREST should recognize that the column markets is a domain over tsvector and allow the fts operator witouth re-wrapping it in the to_tsvector().

## Actual Behavior

The domain wrapper type prevents PostgREST from resolving to_tsvector() and related text search operations, due to missing unwrap of domain types in operator lookup.

## Suggested Fix

Reinstate or modify the type resolution logic so that domain base types are used during operator compatibility checks — as was previously supported before [#3845](https://github.com/PostgREST/postgrest/pull/3845).

This would restore expected support for semantic domains while preserving the benefits of custom type constraints.

laurenceisla:
Closes #4135


src/PostgREST/SchemaCache.hs
```diff
@@ -614,7 +614,33 @@ tablesSqlQuery =
...
+      FROM recurse t
+      JOIN pg_type b ON t.typbasetype = b.oid
+    )
+    SELECT
+      oid,
+      base_namespace,
+      base_type
+    FROM recurse
+    WHERE typbasetype = 0
+  ),
```
steve-chavez: This should be deduplicated, it's the same as the below right?
https://github.com/PostgREST/postgrest/blob/deb11c3c2afc8dc19a630f6b8bfe1ed053c28ed2/src/PostgREST/SchemaCache.hs#L381-L403

src/PostgREST/Plan.hs
```diff
@@ -926,7 +926,7 @@ addNullEmbedFilters (Node rp@ReadPlan{where_=curLogic} forest) = do
     newNullFilters rPlans = \case
       (CoercibleExpr b lOp trees) ->
         CoercibleExpr b lOp <$> (newNullFilters rPlans `traverse` trees)
-      flt@(CoercibleStmnt (CoercibleFilter (CoercibleField fld [] _ _ _ _ _ _) opExpr)) ->
+      flt@(CoercibleStmnt (CoercibleFilter (CoercibleField fld [] _ _ _ _ _ _ _) opExpr)) ->
```
steve-chavez: Lots of wildcards, perhaps it can work like this?
```suggestion
      flt@(CoercibleStmnt (CoercibleFilter CoercibleField{cfName=fld, cfJsonPath=[]} opExpr)) ->
```

src/PostgREST/SchemaCache.hs
```diff
@@ -375,32 +375,40 @@ accessibleFuncs = SQL.Statement sql params decodeFuncs
...
       JOIN pg_type b ON t.typbasetype = b.oid
     )
     SELECT
       oid,
-      base
+      base_namespace,
+      base_type
     FROM recurse
     WHERE typbasetype = 0
   ),
```
steve-chavez: This comma here would prevent usage in a single CTE, so the comma should be specified where this CTE snippet is applied 
```suggestion
  )
```

src/PostgREST/SchemaCache.hs
```diff
@@ -375,32 +375,40 @@ accessibleFuncs = SQL.Statement sql params decodeFuncs
...
+|]
+
+funcsSqlQuery :: SqlQuery
+funcsSqlQuery = encodeUtf8 [trimming|
+  WITH
+  $baseTypesCte
```
steve-chavez: Like so
```suggestion
  $baseTypesCte,
```

src/PostgREST/SchemaCache.hs
```diff
@@ -615,6 +623,7 @@ tablesSqlQuery =
   -- generated columns are only available from pg >= 10 but the query is agnostic to versions. dep.deptype = 'i' is done because there are other 'a' dependencies on PKs
   encodeUtf8 [trimming|
   WITH
+  $baseTypesCte
```
steve-chavez: Ditto
```suggestion
  $baseTypesCte,
```

src/PostgREST/SchemaCache.hs
laurenceisla: True, done.

PR #4147, opened 2025-06-17
ci: fix commitlint CI failure

taimoorzaeem:
Fixes CI failure of `commitlint` as occurred in [#4128 (comment)](https://github.com/PostgREST/postgrest/pull/4128#issuecomment-2981862857).

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -46,7 +46,9 @@ jobs:
       - name: Run commitlint (check locally with `nix-shell --run postgrest-commitlint`)
         run: |
           # Fetch target branch explicitly
-          git fetch origin ${{ github.base_ref }}
+          # For PRs event  -> github.base_ref
+          # For push event -> github.ref_name
+          git fetch origin ${{ github.base_ref || github.ref_name }}
 
           # Run commitlint
-          postgrest-commitlint --from origin/${{ github.base_ref }} --to HEAD
+          postgrest-commitlint --from origin/${{ github.base_ref || github.ref_name }} --to HEAD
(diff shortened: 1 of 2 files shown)


taimoorzaeem:
One way to verify this works after merging, we can duplicate this scenario in a private fork CI.

wolfgangwalther:
I think we should just never run commitlint on push, but only in PRs. There is no point in validating commit messages on the branches, because nothing can be done about them anymore.

taimoorzaeem:
> There is no point in validating commit messages on the branches, because nothing can be done about them anymore.

True. Only benefit I can think of is that in case of direct pushes (without PR), it could insist people with write access to set-up pre-commit hooks to avoid future commitlint failures.

On that note, we should probably also add it here:
https://github.com/PostgREST/postgrest/blob/63a492291a2181828647cafdf918dbd2ae83d3ba/nix/tools/devTools.nix#L79-L86

wolfgangwalther:
> True. Only benefit I can think of is that in case of direct pushes (without PR), it could insist people with write access to set-up pre-commit hooks to avoid future commitlint failures.

No, that's not good, because that would still break CI on the main branch. It will only push the latest devel tag once all other jobs pass. So we *should not run commitlint* on that branch.

wolfgangwalther:
> in case of direct pushes (without PR), it could insist people with write access

If that's a concern, this can be solved differently, by flat out *disallowing* direct pushes via branch protection rule.

taimoorzaeem:
> No, that's not good, because that would still break CI on the main branch. It will only push the latest devel tag once all other jobs pass. So we _should not run commitlint_ on that branch.

Right agreed, then we should definitely remove it from `push` event and only do it on `pull_request`.

PR #4148, opened 2025-06-18
ci: only run commitlint on github PRs

taimoorzaeem:
The `commitlint` script disrupts our release workflow as mentioned in [#4147 (comment)](https://github.com/PostgREST/postgrest/pull/4147#issuecomment-2984072695) if failed on push event. We also don't need it on push because once pushed, git commit history can't be amended.


.github/workflows/check.yaml
```diff
@@ -32,6 +32,7 @@ jobs:
         run: postgrest-style-check
 
   commit:
+    if: github.event_name == 'pull_request'  # we only run this on PRs, a failure on push disrupts the release workflow
```
wolfgangwalther: This will work. I have a *slight* preference for `github.event_name != 'push'`, because it could be theoretically possible in the future that we switch from `pull_request` to `pull_request_target` - and this would be less likely to break.

But that's certainly not blocking.

taimoorzaeem: Right, exactly what thought initially. Didn't know there was a third one. Thanks for pointing that out!

Issue #4149, opened 2025-06-19
Reduce closure size of PostgREST flake installation


laurenceisla:
## Problem

The closure size of installing PostgREST through flakes takes quite some space (around 5 GB):

```console
$ nix path-info --size --closure-size --human-readable /nix/store/hlvvyk395f9f219w2d3b4sdqmmxplymj-postgrest-13.0.4/bin/postgrest
/nix/store/hlvvyk395f9f219w2d3b4sdqmmxplymj-postgrest-13.0.4	 118.0 MiB	   5.3 GiB
```

To get the list of all the dependencies: 

```console
$ nix path-info --recursive --size --closure-size --human-readable /nix/store/hlvvyk395f9f219w2d3b4sdqmmxplymj-postgrest-13.0.4/bin/postgrest
/nix/store/sk8z3kdy6iaky6sg972k6fkh3iw2rn3d-bash-5.2p37                                  	   2.7 MiB	   2.7 MiB
/nix/store/6by1qm73kwvbcrhb54bbwhdlmw062n5g-ghc-9.4.8-doc                                	 690.3 MiB	 693.0 MiB
/nix/store/17zqzf1acnr7g6h8rp0r7bnjphsh5qvr-prettyprinter-1.7.1-doc                      	   3.0 MiB	 696.1 MiB
/nix/store/9vzdyy9593pangljqcywch206g9lrsyh-colour-2.3.6-doc                             	   2.2 MiB	 695.2 MiB
/nix/store/kvqhp9h33sscrxhf312clqicxlbdikhf-ansi-terminal-types-0.11.5-doc               	 438.4 KiB	 695.6 MiB
/nix/store/igk3vh9fxnnk09i28vlafiwdm2m8mjvz-ansi-terminal-1.0.2-doc                      	1010.5 KiB	 696.6 MiB
...
/nix/store/s7p0sdy29nwwgmg7k3h07lp90h8wh1s3-hasql-pool-1.0.1-doc                         	 513.8 KiB	 767.3 MiB
/nix/store/vmz6zdr4q543cnngas7la8bmkss1vq50-hasql-pool-1.0.1                             	   1.2 MiB	   4.2 GiB
/nix/store/kqdlifdbbxxwf2lw5af8gdr30qjy4sdc-hasql-notifications-0.2.2.2-doc              	 240.2 KiB	 769.7 MiB
[...]

PR #4156, opened 2025-06-21
nix: reduce closure size of default package in flake.nix

wolfgangwalther:
Splitting the executable in a separate output avoids distributing all of the libraries and documentation, which are not needed when just running PostgREST.

Reduces closure size from 4.3G to 73.9M for the flake exported packages.

Resolves https://github.com/PostgREST/postgrest/issues/4149

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -92,6 +92,7 @@ rec {
   postgrestPackage = pkgs.lib.pipe postgrest [
     lib.dontCheck
     lib.enableSeparateBinOutput
+    (haskellPackages.generateOptparseApplicativeCompletions [ "postgrest" ])
   ];
 
   # Profiled dynamic executable.


laurenceisla:
Great! I also get the same results on a NixOS x64 machine:

```console
$ nix path-info --size --closure-size --human-readable /nix/store/h744gd1s13fv0vvwpkzq316qy0pd6g7m-postgrest-13.1-bin/bin/postgrest
/nix/store/h744gd1s13fv0vvwpkzq316qy0pd6g7m-postgrest-13.1-bin	  13.5 MiB	  73.9 MiB
```

However in aarch64 and MacOS systems I still get the closure size in GB, albeit a bit less (around 4.5 GB).

wolfgangwalther:
Hm, for `aarch64-linux` it should be "enough" to update nixpkgs (which is not an easy thing in itself, yet). `aarch64-darwin` (MacOS) has that huge closure size in nixpkgs, too, so an upstream problem.

None of the two can be easily fixed here, I'm afraid.

laurenceisla:
Ah, got it. I'll keep the issue open to not lose track of those systems then.

Issue #4154, opened 2025-06-20
New `--dump-config-sources` CLI option


steve-chavez:
## Problem

When doing `postgrest --dump-config` it's not possible to know the sources of the config values.

Previously discussed on https://github.com/PostgREST/postgrest/pull/4084#discussion_r2138071617

## Solution

Add a new option `--dump-config-sources` that has an output like this:

```config
...
db-schemas = "public" # public (default), api (file), public (env), public (db) 
```

i.e. it should show all the possible sources of a config value. The sequence should be default/file/env/db as the next overrides the previous one.

## Notes

This might help solving https://github.com/PostgREST/postgrest/issues/3101 too.

wolfgangwalther:
How about doing something like `--dump-config=file`, `--dump-config=env`, `--dump-config=db`, `--dump-config=default` instead?

`--dump-config` would continue to do exactly what it does right now. The others would *only* dump the relevant parts. If there are no env variables, `--dump-config=env` would be empty - it would *not* return defaults. Same for the others.

steve-chavez:
Sounds good to me. I wanted a view of all sources in one go to help in debugging, but having it in the comment seems a bit unstructured.

steve-chavez:
One extra nicety could be adding the name of the source as comment:

```bash
$ postgrest --dump-config=env
...
db-schemas = "public" # PGRST_DB_SCHEMAS

$ postgrest --dump-config=db
...
db-schemas = "public" # pgrst.db_schemas
```

PR #4157, opened 2025-06-24
refactor: decouple config names and values

taimoorzaeem:
Decouples (or I should say "de-tuple" :face_with_diagonal_mouth: ) the config names and values in config dumping. This opens up space for easier implementation of #4154. Not sure if that is the best way to refactor this. Would love any suggestions.

### DIFF
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -151,47 +151,91 @@ toText conf =
   unlines $ (\(k, v) -> k <> " = " <> v) <$> pgrstSettings ++ appSettings
   where
     -- apply conf to all pgrst settings
-    pgrstSettings = (\(k, v) -> (k, v conf)) <$>
-      [("db-aggregates-enabled",         T.toLower . show . configDbAggregates)
-      ,("db-anon-role",              q . T.decodeUtf8 . fromMaybe "" . configDbAnonRole)
-      ,("db-channel",                q . configDbChannel)
-      ,("db-channel-enabled",            T.toLower . show . configDbChannelEnabled)
-      ,("db-extra-search-path",      q . T.intercalate "," . configDbExtraSearchPath)
-      ,("db-hoisted-tx-settings",    q . T.intercalate "," . configDbHoistedTxSettings)
-      ,("db-max-rows",                   maybe "\"\"" show . configDbMaxRows)
-      ,("db-plan-enabled",               T.toLower . show . configDbPlanEnabled)
-      ,("db-pool",                       show . configDbPoolSize)
-      ,("db-pool-acquisition-timeout",   show . configDbPoolAcquisitionTimeout)
-      ,("db-pool-max-lifetime",          show . configDbPoolMaxLifetime)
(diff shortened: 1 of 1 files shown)


steve-chavez:
I think we do need a SSoT for the config names, but the refactor is not looking like a gain in maintenance (has more LOCs).

@taimoorzaeem As a shorter step towards #4154, how about first solving the duplication problem for the sample config in the CLI?

https://github.com/PostgREST/postgrest/blob/c1eb8af2db492337dc33d9e390477d2f1f79f8c3/src/PostgREST/CLI.hs#L126-L238

Then the benefit of the refactor will be clear.

taimoorzaeem:
https://github.com/PostgREST/postgrest/blob/f29513054443e85957029b6c8aeb5d1b1ae298ab/src/PostgREST/CLI.hs#L143-L151

The description before each config is understandable but can we afford to lose this trailing `##` and `#` just before in each config here for instance, in `db-pre-config` and `db-max-rows`? Not sure what the means TBH.

My idea is to utilize `Config.toText` and intercalate with the description list. Just these trailing `##` and `#` get in the way.

steve-chavez:
IIRC `# db-max-rows = 1000 ` has the `#` because it's actually not set, there's no default. The configs that don't have a `#` prefix have a default like `db-schemas = "public"`.

So the above means that there's an additional field for a type (that has possible values: default or empty), that would translate to those commented out configs.

I guess we put `##` in the description for some tools (like vim) that automatically un-comment by removing one `#`, so the config remains valid.

It would be just a nicety to keep them as is.

laurenceisla:
> IIRC # db-max-rows = 1000 has the # because it's actually not set, there's no default.

Agree, these are the ones labeled as "Default: n/a" in the docs (or in the case above when: default = not set). Note that this doesn't seem to be consistent right now, cause there are configs that are commented when they have a default value (like the `db-pool...` examples), so they shouldn't be commented anymore. I think these are specified as `Maybe` types in the `AppConfig`, so that could be a way to detect them programmatically.

taimoorzaeem:
> @taimoorzaeem As a shorter step towards #4154, how about first solving the duplication problem for the sample config in the CLI?
> 
> https://github.com/PostgREST/postgrest/blob/c1eb8af2db492337dc33d9e390477d2f1f79f8c3/src/PostgREST/CLI.hs#L126-L128
> 
> Then the benefit of the refactor will be clear.

One way to solve this could be to do something like:

```hs
{-# OPTIONS_GHC -Wunused-record-wildcards #-} -- only available in ghc >= 9.6.1
exampleConfig :: AppConfig
...

toExampleFile :: AppConfig -> [Char]
toExampleFile AppConfig{..} =
```

Now, we want to haskell to complain at compile time if some field is not consumed in the `toExampleFile` function, so for that we can use `-Wunused-record-wildcards`. But, this warning is only available in `GHC >= 9.6.1` according to [haskell errors](https://errors.haskell.org/), so I think we need wait for the upgrade for that.

taimoorzaeem:

PR #4160, opened 2025-06-25
chore: update postgrest scripts in nix README.md

taimoorzaeem:
Adds `postgrest-commitlint` and removes `postgrest-nixpkgs-upgrade`.

As discussed in [#4128 (comment)](https://github.com/PostgREST/postgrest/pull/4128#discussion_r2133958281).

BTW, this reminds me that we also need to update `nix/UPGRADE.md` and `docs/requirements.txt` but I think that should be done in a separate PR.


wolfgangwalther:
Yeah, I think this is actually an oversight: Now that we intend to only do `nix flake update`, we will not have automated updates of `docs/requirements.txt` anymore. I think we should re-introduce `postgrest-nixpkgs-upgrade` to wrap both instead of removing its traces.

steve-chavez (approved):

taimoorzaeem:
>  think we should re-introduce `postgrest-nixpkgs-upgrade` to wrap both instead of removing its traces.

Tracked on #4161.

PR #4162, opened 2025-06-26
refactor: merge transaction end configs to a single config

taimoorzaeem:
This slightly cleans up config module. End goal is closing #3101.

### Notes

One thing I noticed with config module is that the fields in `AppConfig` don't one-to-one map to end-user configs. This interferes with SSoT principle.

### DIFF
--- a/src/PostgREST/ApiRequest.hs
+++ b/src/PostgREST/ApiRequest.hs
@@ -45,6 +45,7 @@ import Web.Cookie                (parseCookies)
 
 import PostgREST.ApiRequest.QueryParams  (QueryParams (..))
 import PostgREST.Config                  (AppConfig (..),
+                                          DbTxEnd (..),
                                           OpenAPIMode (..))
 import PostgREST.Config.Database         (TimezoneNames)
 import PostgREST.Error                   (ApiRequestError (..),
@@ -164,7 +165,13 @@ userApiRequest conf prefs req reqBody = do
 
 -- | Parses the Prefer header
 userPreferences :: AppConfig -> Request -> TimezoneNames -> Preferences.Preferences
-userPreferences conf req timezones = Preferences.fromHeaders (configDbTxAllowOverride conf) timezones $ requestHeaders req
+userPreferences conf req timezones = Preferences.fromHeaders isTxOverrideAllowed timezones $ requestHeaders req
+  where
+    isTxOverrideAllowed = case configDbTxEnd conf of
+      TxCommitAllowOverride   -> True
+      TxRollbackAllowOverride -> True
+      TxCommit                -> False
+      TxRollback              -> False
 
 getResource :: AppConfig -> [Text] -> Either ApiRequestError Resource
 getResource AppConfig{configOpenApiMode, configDbRootSpec} = \case
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -12,6 +12,7 @@ Description : Manages PostgREST configuration type and parser.
 
 module PostgREST.Config
   ( AppConfig (..)
+  , DbTxEnd (..)
   , Environment
   , JSPath
   , JSPathExp(..)
@@ -88,8 +89,7 @@ data AppConfig = AppConfig
   , configDbSchemas                :: NonEmpty Text
   , configDbConfig                 :: Bool
   , configDbPreConfig              :: Maybe QualifiedIdentifier
-  , configDbTxAllowOverride        :: Bool
-  , configDbTxRollbackAll          :: Bool
+  , configDbTxEnd                  :: DbTxEnd
   , configDbUri                    :: Text
   , configFilePath                 :: Maybe FilePath
   , configJWKS                     :: Maybe JwkSet
@@ -116,6 +116,13 @@ data AppConfig = AppConfig
   , configRoleIsoLvl               :: RoleIsolationLvl
   , configInternalSCSleep          :: Maybe Int32
   }
+
+data DbTxEnd
+  = TxCommit
+  | TxCommitAllowOverride
+  | TxRollback
+  | TxRollbackAllowOverride
+  deriving (Eq)
 
 data LogLevel = LogCrit | LogError | LogWarn | LogInfo | LogDebug
   deriving (Eq, Ord)
@@ -171,7 +178,7 @@ toText conf =
       ,("db-schemas",                q . T.intercalate "," . toList . configDbSchemas)
       ,("db-config",                     T.toLower . show . configDbConfig)
       ,("db-pre-config",             q . maybe mempty dumpQi . configDbPreConfig)
-      ,("db-tx-end",                 q . showTxEnd)
+      ,("db-tx-end",                 q . showTxEnd . configDbTxEnd)
       ,("db-uri",                    q . configDbUri)
       ,("jwt-aud",                   q . fromMaybe mempty . configJwtAudience)
       ,("jwt-role-claim-key",        q . T.intercalate mempty . fmap dumpJSPath . configJwtRoleClaimKey)
@@ -200,16 +207,19 @@ toText conf =
     -- quote strings and replace " with \"
     q s = "\"" <> T.replace "\"" "\\\"" s <> "\""
 
-    showTxEnd c = case (configDbTxRollbackAll c, configDbTxAllowOverride c) of
-      ( False, False ) -> "commit"
-      ( False, True  ) -> "commit-allow-override"
-      ( True , False ) -> "rollback"
-      ( True , True  ) -> "rollback-allow-override"
+    showTxEnd :: DbTxEnd -> Text
+    showTxEnd = \case
+      TxCommit                -> "commit"
+      TxCommitAllowOverride   -> "commit-allow-override"
+      TxRollback              -> "rollback"
+      TxRollbackAllowOverride -> "rollback-allow-override"
+
     showJwtSecret c
       | configJwtSecretIsBase64 c = B64.encode secret
       | otherwise                 = secret
       where
         secret = fromMaybe mempty $ configJwtSecret c
+
     showSocketMode c = showOct (configServerUnixSocketMode c) mempty
 
 -- This class is needed for the polymorphism of overrideFromDbOrEnvironment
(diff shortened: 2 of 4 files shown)


src/PostgREST/Query.hs
```diff
@@ -255,7 +256,7 @@ failExceedsMaxAffectedPref (Just (PreferMaxAffected n), handling) RSStandard{rsQ
 -- | Set a transaction to roll back if requested
 optionalRollback :: AppConfig -> ApiRequest -> DbHandler ()
 optionalRollback AppConfig{..} ApiRequest{iPreferences=Preferences{..}} = do
-  lift $ when (shouldRollback || (configDbTxRollbackAll && not shouldCommit)) $ do
+  lift $ when (shouldRollback || (configDbTxEnd `elem` [TxRollback,TxRollbackAllowOverride] && not shouldCommit)) $ do
```
wolfgangwalther: `elem` checks like that are not good for maintainability, imho.

It would be better to create a helper function that pattern matches and returns true/false. This would immediately complain when adding a new value to the type - forcing you to take a decision immediately.

src/PostgREST/ApiRequest.hs
```diff
@@ -164,7 +165,7 @@ userApiRequest conf prefs req reqBody = do
 
 -- | Parses the Prefer header
 userPreferences :: AppConfig -> Request -> TimezoneNames -> Preferences.Preferences
-userPreferences conf req timezones = Preferences.fromHeaders (configDbTxAllowOverride conf) timezones $ requestHeaders req
+userPreferences conf req timezones = Preferences.fromHeaders (configDbTxEnd conf `elem` [TxCommitAllowOverride,TxRollbackAllowOverride]) timezones $ requestHeaders req
```
wolfgangwalther: Same here.

src/PostgREST/Query.hs
taimoorzaeem: I am not convinced by the impact on maintainability, however, it would look cleaner that way.

wolfgangwalther: Well, I have been there in the past on some other types which were checked via elem - and I didn't notice me breaking stuff by introducing new values to the type. It's compile-time errors we really like, right?

wolfgangwalther:
General opinion on the PR: I'm undecided whether that's an overall net-plus and don't see why it's needed, yet.

taimoorzaeem:
> General opinion on the PR: I'm undecided whether that's an overall net-plus and don't see why it's needed, yet.

See, our `AppConfig` has some fields that don't map to end user configs, they are just there for keeping state (it shouldn't be this way). The eventual goal is to have exactly one field in `AppConfig` for each end-user config to make things consistent. 

From a consistent `AppConfig`, we can derive more things. I try to make small PRs, so it's easier for the reviewers. This doesn't always make the benefits clear, but it doesn't mean we shouldn't try making the codebase more maintainable.

src/PostgREST/Query.hs
taimoorzaeem: > Well, I have been there in the past on some other types which were checked via elem - and I didn't notice me breaking stuff by introducing new values to the type. It's compile-time errors we really like, right?

Hm, sometimes pattern matching could be a lot of work, but I see your point now. Yes, we love compile-time errors.

wolfgangwalther:
> See, our `AppConfig` has some fields that don't map to end user configs, they are just there for keeping state (it shouldn't be this way).

I don't see any state in the db-tx fields. There might be a single config option and two fields, but that doesn't mean there is no clear mapping between them.

taimoorzaeem:
> I don't see any state in the db-tx fields. There might be a single config option and two fields, but that doesn't mean there is no clear mapping between them

See in [#3101 (comment)](https://github.com/PostgREST/postgrest/issues/3101#issuecomment-3004318783):

```hs
data AppConfig = AppConfig
  { 
   ...
  , configDbTxAllowOverride  :: Bool
  , configDbTxRollbackAll    :: Bool
   ...
  }
```

Here marking these two configs as `'File` or `'FileAndInDb` is inconsistent with the fact that only one config actually exists but two are being used internally which is unnecessary and kind of naive.

src/PostgREST/Config.hs
```diff
@@ -373,15 +382,14 @@ parser optPath env dbSettings roleSettings roleIsolationLvl =
...
-        Nothing                        -> pure $ f (False,      False)
-        Just "commit"                  -> pure $ f (False,      False)
-        Just "commit-allow-override"   -> pure $ f (False,      True)
-        Just "rollback"                -> pure $ f (True,       False)
-        Just "rollback-allow-override" -> pure $ f (True,       True)
+        Nothing                        -> pure TxCommit -- default
+        Just "commit"                  -> pure TxCommit
+        Just "commit-allow-override"   -> pure TxCommitAllowOverride
+        Just "rollback"                -> pure TxRollback
+        Just "rollback-allow-override" -> pure TxRollbackAllowOverride
```
taimoorzaeem, line 392: Also, see how this change is much cleaner and less confusing.

steve-chavez: Hm, yeah. I always found these booleans a bit confusing.

src/PostgREST/Query.hs
```diff
@@ -255,14 +256,19 @@ failExceedsMaxAffectedPref (Just (PreferMaxAffected n), handling) RSStandard{rsQ
...
   where
     shouldCommit =
       preferTransaction == Just Commit
     shouldRollback =
       preferTransaction == Just Rollback
+    isTxRollback = case configDbTxEnd of
+      TxRollback              -> True
+      TxRollbackAllowOverride -> True
+      TxCommit                -> False
+      TxCommitAllowOverride   -> False
```
steve-chavez, line 271: But this part is not looking like a net gain. Is there any other way to do it?

src/PostgREST/ApiRequest.hs
```diff
@@ -164,7 +165,13 @@ userApiRequest conf prefs req reqBody = do
...
 -- | Parses the Prefer header
 userPreferences :: AppConfig -> Request -> TimezoneNames -> Preferences.Preferences
-userPreferences conf req timezones = Preferences.fromHeaders (configDbTxAllowOverride conf) timezones $ requestHeaders req
+userPreferences conf req timezones = Preferences.fromHeaders isTxOverrideAllowed timezones $ requestHeaders req
+  where
+    isTxOverrideAllowed = case configDbTxEnd conf of
+      TxCommitAllowOverride   -> True
+      TxRollbackAllowOverride -> True
+      TxCommit                -> False
+      TxRollback              -> False
```
steve-chavez, line 174: This part also overall seems more verbose. Maybe at define it next to the types? Perhaps both these boolean results can be shortened somehow.

wolfgangwalther: The shortening is already there - in the code that is removed!

steve-chavez:
> Here marking these two configs as 'File or 'FileAndInDb is inconsistent with the fact that only one config actually exists but two are being used internally which is unnecessary and kind of naive.
>   , configDbTxAllowOverride  :: Bool
>  , configDbTxRollbackAll    :: Bool

I get the point that having two fields for a single config is inconsistent with all of the other ones but I still don't see the need for GADTs. I'd suggest to keep it the simplest way possible for now. I think we would just need a type like:

```haskell
data PgrstConfig= {
   pcFileName :: Text
,  pcEnvName :: Text
,  pcInDbName :: Maybe Text
,  pcDescription :: Text
...
}
```

`pcInDbName` being the only optional name, both file and env are always possible for all. As you can see with this is clear which are the sources of a config, without the use of GADTs.

---

Besides that I think there is a benefit for typing these values instead of being Bools:

```haskell
data DbTxEnd
  = TxCommit
  | TxCommitAllowOverride
  | TxRollback
  | TxRollbackAllowOverride
  deriving (Eq)
```

That being we can deduplicate these descriptions later on:

https://github.com/PostgREST/postgrest/blob/c1eb8af2db492337dc33d9e390477d2f1f79f8c3/src/PostgREST/CLI.hs#L182-L189

https://github.com/PostgREST/postgrest/blob/c1eb8af2db492337dc33d9e390477d2f1f79f8c3/docs/references/configuration.rst#L533-L543

Right now they're not consistent and unnecessarily separate. I think there's a future where we could generate the config docs automatically from our Config module.

@wolfgangwalther WDYT?

wolfgangwalther:
> WDYT?

I think it's impossible to judge a small refactor PR like this without knowing the bigger picture where we want to get to. There are multiple different ideas on how to evolve the config module, but they need to be prototyped to see whether they actually work, where the limitations are etc.

It makes zero sense to start refactoring stuff before that.

steve-chavez:
> I think it's impossible to judge a small refactor PR like this without knowing the bigger picture where we want to get to.

Assuming we want to deduplicate the descriptions as I mentioned above (which to me looks like an obvious benefit), how can we do it gradually?

Feeding the descriptiosn to the CLI seems possible now with interpolation. For docs, I'm thinking we would need a new CLI option like `--dump-config-rst` or `--dump-config=rst` which has the full output for:

https://github.com/PostgREST/postgrest/blob/c1eb8af2db492337dc33d9e390477d2f1f79f8c3/docs/references/configuration.rst#L149-L938

It needs to be in one go since we need to ensure the alphabetical order of the configs.

Once that's done we could concatenate configuration.rst with this dump, maybe once `postgrest-docs-build` is invoked. 

Now to do that in one big PR might be too complicated and hard.

Would a partial `--dump-config=rst` be reasonable? Then I can create a issue (with subtasks) with this idea, and we can merge PRs gradually.

steve-chavez:
> how can we do it gradually?

Even PostgreSQL introduced its new AIO system gradually, not in one single commit ([ref1](https://github.com/postgres/postgres/commit/02844012b304ba80d1c48d51f6fe10bb622490cc), [ref2](https://github.com/postgres/postgres/commit/55b454d0e14084c841a034073abbf1a0ea937a45), there are more). This for easier review. I think we should do that for big changes too, otherwise progress is too slow (plus new commits add the need for rebases and so on).

wolfgangwalther:
I'm not saying we need to introduce it all in one commit, far from it. But before we have something that we know we *want* to work towards, we can't judge whether our steps are in the right direction.

We need a prototype of how the solution should look like. And then we can commit steps towards it.

taimoorzaeem:
> We need a prototype of how the solution should look like. And then we can commit steps towards it.

Agreed. Without a clear design, it is hard to know what might or might not work out.

taimoorzaeem:

PR #4168, opened 2025-06-27
[v13] docs: fix OpenAPI broken docs link

### LINKED ISSUE #4080: v13 docs link in OpenApi schema is broken
### Environment

* PostgreSQL version: 17.5
* PostgREST version: 13.0.0
* Operating system: NixOS

### Description of issue

The OpenAPI output contains an invalid link to the docs:

```
      url: https://postgrest.org/en/v"13"/references/api.html
```

Note the quotes around the version.

taimoorzaeem:
Closes #4080.


test/spec/Feature/OpenApi/OpenApiSpec.hs
```diff
@@ -46,7 +45,7 @@ spec = describe "OpenAPI" $ do
 
     liftIO $ do
       headers `shouldSatisfy` notZeroContentLength
-      docsUrl `shouldBe` Just (String ("https://postgrest.org/en/" <> docsVersion <> "/references/api.html"))
+      docsUrl `shouldBe` Just (String "https://postgrest.org/en/v13/references/api.html")
```
wolfgangwalther: Why would we hardcode the version here? That's annoying, we'd have to fix it every time?

src/PostgREST/Version.hs
```diff
@@ -26,7 +27,7 @@ prettyVersion =
 docsVersion :: Text
 docsVersion
   | isPreRelease = "latest"
-  | otherwise    =  "v" <> (T.intercalate "." . map show . take 1 $ version)
+  | otherwise    =  "v" <> fromJust (listToMaybe version)
```
wolfgangwalther: Why don't we just remove the `show`? That should be enough, right?

wolfgangwalther:
(and we should do that targeting `main` and then backport the exact same fix)

src/PostgREST/Version.hs
taimoorzaeem: True. But why do we even have intercalate when we only just show the major version?

taimoorzaeem:
> (and we should do that targeting `main` and then backport the exact same fix)

Sure.

test/spec/Feature/OpenApi/OpenApiSpec.hs
taimoorzaeem: Would this change on `v13` branch? I don't think so.

wolfgangwalther: But if you hardcode it, then you need to do the same thing for v14 again...

src/PostgREST/Version.hs
wolfgangwalther: I guess that's from when we still had two components to it. It's also a way to deal with the list without using unsafe `fromJust`. Not that it's a particularly good way, but it is a way.

wolfgangwalther: The correct way to do it, would be to make `version` a `List.NonEmpty Text`, I guess. It's odd, though - I can't find any `splitOn` which returns `NonEmpty`. This should be possible in principle..

taimoorzaeem: Yes, but I guess it isn't too necessary for now. Lets just remove `show` for now.

PR #4169, opened 2025-06-27
fix: fix OpenAPI broken docs link

taimoorzaeem:
Prevents future broken links. Related to #4080.

### DIFF
--- a/src/PostgREST/Version.hs
+++ b/src/PostgREST/Version.hs
@@ -26,7 +26,7 @@ prettyVersion =
 docsVersion :: Text
 docsVersion
   | isPreRelease = "latest"
-  | otherwise    =  "v" <> (T.intercalate "." . map show . take 1 $ version)
+  | otherwise    =  "v" <> T.intercalate "." (take 1 version)
 
 
 -- | Versions with two components (e.g., '1.1') are treated as pre-releases.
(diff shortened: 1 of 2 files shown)


src/PostgREST/Version.hs
```diff
@@ -26,7 +26,7 @@ prettyVersion =
 docsVersion :: Text
 docsVersion
   | isPreRelease = "latest"
-  | otherwise    =  "v" <> (T.intercalate "." . map show . take 1 $ version)
+  | otherwise    =  "v" <> T.intercalate "." (take 1 version)
```
steve-chavez, line 29: Look enough going from the discussion on https://github.com/PostgREST/postgrest/pull/4168/files#r2171481594

steve-chavez:
Should this use the `docs:` prefix? Isn't it really a `fix:` since it fixes the OpenAPI output?

taimoorzaeem:
> Should this use the `docs:` prefix? Isn't it really a `fix:` since it fixes the OpenAPI output?

Yes, agreed,

steve-chavez:
> Should this use the docs: prefix? Isn't it really a fix: since it fixes the OpenAPI output?

I wonder if `commitlint` allows a rule like "only allow a prefix if these specific files are touched"? That in theory would let us match a `docs:` prefix with only changes to `*.rst` files.

Not strictly needed, but it would be nice.

taimoorzaeem:
> I wonder if `commitlint` allows a rule like "only allow a prefix if these specific files are touched"? That in theory would let us match a `docs:` prefix with only changes to `*.rst` files.

No it doesn't natively support that. It is possible to write custom plugins with `@commitlint/cli` but that would be too much effort, plus it isn't available in nix anyways.

PR #4174, opened 2025-07-05
Fix OpenAPI spec to respect function volatility for GET methods

joelonsql:
## Description

This PR fixes a bug where the OpenAPI specification incorrectly exposed GET methods for `VOLATILE` functions, even though such functions properly reject GET requests at runtime with "405 Method Not Allowed". This created a mismatch between the advertised API specification and the actual runtime behavior.

### The Problem

PostgREST was generating OpenAPI specifications that included GET methods for ALL functions regardless of their volatility, while the runtime correctly blocked GET requests to volatile functions. This inconsistency caused:

- Client applications attempting invalid GET requests based on the OpenAPI spec
- Unnecessary network traffic and 405 errors
- Confusion for developers using the OpenAPI spec for code generation
- Poor user experience with error logs and delayed page rendering

### The Solution

The fix modifies the OpenAPI generation logic in `makeProcPathItem()` to check the function's volatility (`pdVolatility`) and only include GET methods for non-volatile functions:

- `VOLATILE` functions: Only POST method exposed
- `STABLE` and `IMMUTABLE` functions: Both GET and POST methods exposed

This aligns the OpenAPI specification with the existing runtime behavior and PostgreSQL's function volatility semantics.

## Changes

- Modified `src/PostgREST/Response/OpenAPI.hs` to check function volatility when generating PathItem
- Added comprehensive tests in `test/spec/Feature/OpenApi/OpenApiSpec.hs` to verify correct method exposure based on volatility

## Testing

Added test cases that verify:
- VOLATILE functions only show POST in OpenAPI spec
- STABLE functions show both GET and POST in OpenAPI spec  
- IMMUTABLE functions show both GET and POST in OpenAPI spec

## Impact

This is a bug fix that corrects the OpenAPI documentation to match existing runtime behavior. No breaking changes to the API itself - only the documentation is corrected.

### Benefits for Users

- OpenAPI specs now accurately represent which HTTP methods are allowed
- Client code generators will produce correct API calls
- Eliminates unnecessary failed GET requests to volatile functions
- Reduces confusion when OpenAPI spec doesn't match runtime behavior

## Related Issues

This issue was discovered during real-world usage where frontend applications were automatically calling all endpoints with GET methods listed in the OpenAPI spec, causing multiple failed requests per page load.

The runtime behavior was already correct since PostgREST v8.0.0 (see #1560), but the OpenAPI generation was never updated to match.

## Note on Existing Limitations

This fix does not address the existing limitation where OpenAPI cannot properly represent overloaded PostgreSQL functions (functions with the same name but different parameter types). When multiple functions share the same name, only one variant will appear in the OpenAPI specification, though PostgREST will correctly route calls to the appropriate function based on the provided parameters. This is a fundamental limitation of the OpenAPI specification format, not PostgREST.

## Checklist

- [x] Code changes are minimal and focused on the specific issue
- [x] Tests added to prevent regression
- [x] No breaking changes to existing API behavior
- [x] Follows PostgreSQL's function volatility semantics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix OpenAPI broken docs link by @taimoorzaeem in #4080
+- Fix OpenAPI specification incorrectly exposing GET methods for volatile functions by @joelonsql in #4174
 
 ## [13.0.4] - 2025-06-17
 


joelonsql:
> Changes LGTM, but we need an entry in `CHANGELOG.md` as well.
> 
> I agree on "fix" and think we can easily backport this.

Cool! Let me know if you want me to take care of the backporting as well.

wolfgangwalther:
> Cool! Let me know if you want me to take care of the backporting as well.

No need, this will be a simple cherry-pick on the v13 branch, I'm sure it will apply cleanly.

laurenceisla (approved):
Also LGTM!

wolfgangwalther:
Thank you, @joelonsql!

Cherry-picked it to v13 as well.

Issue #4123, opened 2025-06-02
test: split loadtests by request type


steve-chavez:
## Problem

Currently it's not possible to know accurately if performance is dropping or increasing on a PR based on deltas of the "mixed" loadtest, this is because all the requests types are contained in a single loadtest:

https://github.com/PostgREST/postgrest/blob/dee7d6f39c867a5d141e31192d21ffb9653ffd0a/test/load/targets.http#L1-L41

More details at: https://github.com/PostgREST/postgrest/pull/4097#pullrequestreview-2858786090

## Solution

Split the mixed loadtest into other loadtests by request type:

- [ ] Root openapi
- [ ] Read table (GET/HEAD)
- [ ] Read table with resource embedding
- [ ] Insert table single
- [ ] Insert table bulk
- [ ] Mutate table (PUT/PATCH)
- [ ] Delete table 
- [ ] RPC GET
- [ ] RPC POST
- [ ] OPTIONS on table

steve-chavez:
We should also remove the JWTs from all the above loadtests, those are already covered on https://github.com/PostgREST/postgrest/pull/4175

steve-chavez:
Consider https://github.com/PostgREST/postgrest/pull/4472#issuecomment-3703502623 when splitting:

> Although the CPU usage is almost x3 on the errors load test unlike the mixed one, I wonder why that is?

> Yeah, that's strange. I examined this and I think that's because on mixed ones, we also have have OPTIONS and a GET on root endpoint which make the overall CPU usage looks less. After removing these two cases and re-running the mixed ones, I noticed that the CPU usage is actually quite similar to what we have in the errors test.

We could have a loadtest that ensures the CPU usage stays low on OPTIONS requests.

PR #4175, opened 2025-07-06
nix: complete jwt loadtests

steve-chavez:
Completes the jwt loadtests, now we have non-cached, cached + worst case for the cache. Also clarifies the names of the loadtests:

- `jwt-hs`: now has the cache disabled
- `jwt-hs-cache`: cache enabled for jwt-hs
- `jwt-hs-cache-worst (old jwt-hs)`: worst case of the jwt-hs cache
- `jwt-rsa`: now has the cache disabled
- `jwt-rsa-cache (old jwt-rsa)`: cache enabled for jwt-rsa
- `jwt-rsa-cache-worst`: worst case of the jwt-rsa cache

Also deletes `nix/tools/generate_targets_rsa.py` and uses a single python script.

Should prove what's mentioned on https://github.com/PostgREST/postgrest/pull/4084#issuecomment-2998170423


steve-chavez:
Not sure why the `jwt-rsa-cache-worst` is not finishing:
https://github.com/PostgREST/postgrest/actions/runs/16103335248/job/45435306790?pr=4175

The log traces of the job show its finished.

**Edit**: nvm, forgot it has to run against head and latest version.

steve-chavez:
The good:

- Proves the `*-cache` versions consume less CPU and finish faster than the non-cache versions.
- Proves `jwt-rsa-cache-worst` has the same memleak problem as `jwt-hs-cache-worst`

The bad:

- `jwt-rsa-cache-worst` takes too long to generate the targets file, around 3.5 minutes. Similar numbers locally and on github actions. Plus `jwt-rsa-cache-worst` runs 3 times because of `postgrest-loadtest-against` on github actions, taking more.
  + Maybe we can disable this after the new jwt cache is merged.
 - Github hides some summaries and you have to manually click "Load summary" to show them (see [here](https://github.com/PostgREST/postgrest/actions/runs/16103335248?pr=4175))
   + With more loadtests (https://github.com/PostgREST/postgrest/issues/4123) this will be a bigger problem.
   + Not sure if this is because of the amount of summaries or because of the page size. If it's the former, then we could try running all the fast loadtests in one job and leave the slow ones for independent jobs.

PR #4176, opened 2025-07-07
docs: rename Authentication page to JWT Authentication

### LINKED ISSUE #4110: docs: rename `Authentication` page to `JWT Authentication`
## Problem

The name "Authentication" in the [Authentication"](https://docs.postgrest.org/en/v13/references/auth.html) page is misleading as it's a more loaded term than can be interpreted as validating against users and passwords.

Also, in the code comments we say:

https://github.com/PostgREST/postgrest/blob/800b32a59c048a08787990a76543818250511840/src/PostgREST/Auth.hs#L9-L11

So we need to clarify that "Authentication" is actually external.

## Solution

We should change the page name to "JWT Authentication".

While at it, also document the external authentication component in [Architecture](https://docs.postgrest.org/en/v13/explanations/architecture.html).

taimoorzaeem:
Closes #4110.


steve-chavez:
Needs more than renaming the titles as mentioned on https://github.com/PostgREST/postgrest/issues/4110#issuecomment-3045955571.

taimoorzaeem:
Ah, I see. Closing this as incomplete.

PR #4180, opened 2025-07-08
Update stackage to 22.44 / GHC 9.6.7

wolfgangwalther:
If this works, we are not forced to GHC 9.4 by FreeBSD anymore.

Only blocker to move off of GHC 9.4 would then be the static build, which I am making progress with in Nixpkgs.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -16,7 +16,7 @@ extra-source-files: CHANGELOG.md
 cabal-version:      >= 1.10
 
 tested-with:
-    -- nix, cabal on Ubuntu (arm)
+    -- nix
     GHC == 9.4.8
     -- cabal on Ubuntu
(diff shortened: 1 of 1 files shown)


steve-chavez:
I forget, the main reason we still have stackage is because `cabal` doesn't work on windows right? https://github.com/PostgREST/postgrest/issues/2271#issuecomment-1118138109

wolfgangwalther:
tbh, I don't really know either. I guess building with stack isn't that bad, because it will, eventually, when we can get rid of all the custom overrides..., be the same set of dependencies that upstream nixpkgs builds with. The nixpkgs package set is always based off of the stackage set.

But yeah, when I can get the static build working for both Linux platforms and hopefully also for both MacOS platforms... - then we only have stack for FreeBSD and Windows. Building a static exec for FreeBSD via Nix is not unthinkable, but Windows is probably farther out of reach.

PR #4182, opened 2025-07-08
Update stackage 22.44 -> 23.28 / GHC 9.8.4

wolfgangwalther:
~~This now matches the stackage / GHC major versions Nixpkgs uses.~~ (Edit: Not anymore, Nixpkgs is now at 9.10.3)

Split into a separate PR from the one before, to see all builds pass for both.


wolfgangwalther:
While GHC 9.8.4 is already packaged for FreeBSD, this doesn't seem to be available in the FreeBSD VM images for Cirrus, yet. Thus, this step will have to wait.

wolfgangwalther:
GHC 9.8.4 is now available on Cirrus' FreeBSD images. Should just need to pass CI and then be good to go.

steve-chavez (approved):

PR #4192, opened 2025-07-11
fix: empty spread embeddings return unexpected SQL error

### LINKED ISSUE #3887: Empty spread embeddings return unexpected SQL error
I just found the following error:

```
curl 'localhost:3000/projects?select=*,...clients()'
{"code":"42601","details":null,"hint":null,"message":"syntax error at or near \"FROM\""}
```

_Originally posted by @steve-chavez in https://github.com/PostgREST/postgrest/issues/3885#issuecomment-2622701383_

taimoorzaeem:
Closes #3887.

### DIFF
--- a/src/PostgREST/Query/QueryBuilder.hs
+++ b/src/PostgREST/Query/QueryBuilder.hs
@@ -69,21 +69,21 @@ readPlanToQuery node@(Node ReadPlan{select,from=mainQi,fromAlias,where_=logicFor
 
 getJoinSelects :: ReadPlanTree -> [SQL.Snippet]
 getJoinSelects (Node ReadPlan{relSelect} _) =
-  mapMaybe relSelectToSnippet relSelect
+  join $ map relSelectToSnippet relSelect
   where
-    relSelectToSnippet :: RelSelectField -> Maybe SQL.Snippet
+    relSelectToSnippet :: RelSelectField -> [SQL.Snippet]
     relSelectToSnippet fld =
       let aggAlias = pgFmtIdent $ rsAggAlias fld
       in
         case fld of
           JsonEmbed{rsEmptyEmbed = True} ->
-            Nothing
+            []
           JsonEmbed{rsSelName, rsEmbedMode = JsonObject} ->
-            Just $ "row_to_json(" <> aggAlias <> ".*)::jsonb AS " <> pgFmtIdent rsSelName
+            ["row_to_json(" <> aggAlias <> ".*)::jsonb AS " <> pgFmtIdent rsSelName]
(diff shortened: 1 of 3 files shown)


src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -84,7 +84,12 @@ getJoinSelects (Node ReadPlan{relSelect} _) =
           JsonEmbed{rsSelName, rsEmbedMode = JsonArray} ->
             Just $ "COALESCE( " <> aggAlias <> "." <> aggAlias <> ", '[]') AS " <> pgFmtIdent rsSelName
           Spread{rsSpreadSel, rsAggAlias} ->
-            Just $ intercalateSnippet ", " (pgFmtSpreadSelectItem rsAggAlias <$> rsSpreadSel)
+            case rsSpreadSel of
+              -- for empty spread like ...table() we return Nothing because
+              -- otherwise "Just []" get intercalated by mapMaybe which
+              -- cause an extra "," in the generated SQL
+              [] -> Nothing
+              _  -> Just $ intercalateSnippet ", " (pgFmtSpreadSelectItem rsAggAlias <$> rsSpreadSel)
```
wolfgangwalther: That's confusing and I don't understand why it happens. Surely `intercalate "," [ ]` should return an empty string, right?

And yes, it does:

```
ghci> T.intercalate "," [ ]
""
```

taimoorzaeem: The key point here is the `mapMaybe` function. Yes, the `intercalateSnippet` returns `""` and then the result we get is `Just ""`. The `mapMaybe` convert this to `[""]`, which causes the issue in the `intercalateSnippet` in the `readPlanToQuery` function. To make this more clear, this happens under the hood:
```hs
> intercalateSnippet ", " (["\"test\".\"projects\".*"] ++ [] ++ [""])
"test"."projects".*, -- extra comma
```
I hope that clears the confusion.

wolfgangwalther: Ah, I see. I think the problem is that we have two `intercalateSnippet` calls nested. We shouldn't do that and instead return a list from this part. The `getJoinSelects` function already returns a list - it's just internally, that we return Maybes and then convert them to a list again.

Instead all these conditions could return lists directly. Empty lists instead of Nothing and singletons instead of the other two cases above.

taimoorzaeem:
Shouldn't we backport this? I noticed that we didn't.

steve-chavez:
Yes, I think backporting this would be good. If the cherry picks are clean we can backport all of the [fixes in the changelog](https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#fixed) too

PR #4193, opened 2025-07-11
Build static executable for linux aarch64 (w/ TH)

wolfgangwalther:
Compared to #3866, this also enables building the static executable for Linux aarch64, but keeping Template Haskell around. This is based on upstream work to make Template Haskell work with GHC 9.6+ in Nixpkgs. It's currently based off my personal branch, which is subject to change and disappear eventually. Thus, this is not ready to merge, yet. But it works.

Also, the release process around the docker image still needs to be tested.

Note: This does not mean we can suddenly start using Template Haskell everywhere again. The main problem of Template Haskell is still there: We can't cross-compile it. And we'll still want to cross-compile static executables to FreeBSD, eventually. And maybe even Windows. So the long-term goal should still be to get rid of Template Haskell.

Also, we can only do this update here, because we pin `hsie` to use GHC 9.4 in the second commit. We *will* fall over this eventually, when GHC 9.4 is not supported in Nixpkgs anymore - or when we start using syntax that is GHC 9.6+ only at some point. `hsie` is tightly coupled to internal GHC APIs to do its thing, so non-trivial to update. @taimoorzaeem maybe you can look into that? Since we can separate the pin for `hsie` easily, it might make sense - depending on the required changes - to just jump to e.g. GHC 9.12 immediately and don't even work with the intermediate versions. Or, if it's easier do it one version at a time - whatever works best, I guess.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,8 +30,18 @@ concurrency:
 
 jobs:
   static:
-    name: Nix - Linux x86-64 static
-    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Linux aarch64
+            runs-on: ubuntu-24.04-arm
+            artifact: aarch64
+          - name: Linux x86-64
+            runs-on: ubuntu-24.04
+            artifact: x86-64
+    name: Nix - ${{ matrix.name }} static
+    runs-on: ${{ matrix.runs-on }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Setup Nix Environment
@@ -44,17 +54,17 @@ jobs:
       - name: Save built executable as artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: postgrest-linux-static-x86-64
+          name: postgrest-linux-static-${{ matrix.artifact }}
           path: result/bin/postgrest
           if-no-files-found: error
 
       - name: Build Docker image
-        run: nix-build -A docker.image --out-link postgrest-docker.tar.gz
+        run: nix-build -A docker.image --out-link postgrest-docker-${{ matrix.artifact }}.tar.gz
       - name: Save built Docker image as artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: postgrest-docker-x86-64
-          path: postgrest-docker.tar.gz
+          name: postgrest-docker-${{ matrix.artifact }}
+          path: postgrest-docker-${{ matrix.artifact }}.tar.gz
           if-no-files-found: error
 
 
@@ -87,15 +97,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: Linux aarch64
-            runs-on: ubuntu-24.04-arm
-            cache: |
-              ~/.stack/pantry
-              ~/.stack/snapshots
-              ~/.stack/stack.sqlite3
-            artifact: postgrest-ubuntu-aarch64
-            deps: sudo apt-get update && sudo apt-get install libpq-dev
-
           - name: MacOS aarch64
             runs-on: macos-14
             cache: |
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -75,6 +75,9 @@ jobs:
 
           mkdir -p release-bundle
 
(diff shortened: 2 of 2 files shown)


taimoorzaeem:
> [...] we start using syntax that is GHC 9.6+ only at some point. `hsie` is tightly coupled to internal GHC APIs to do its thing, so non-trivial to update. @taimoorzaeem maybe you can look into that? 

Sure, I can look into that. I'll put this in my backlog. Is nixpkgs dropping support for GHC 9.4 anytime soon?

src/PostgREST/AppState.hs
```diff
@@ -215,21 +215,30 @@ initPool AppConfig{..} observer = do
...
+    whenLeft res (\case
+      SQL.AcquisitionTimeoutUsageError ->
+        observer $ PoolAcqTimeoutObs SQL.AcquisitionTimeoutUsageError
+      err@(SQL.ConnectionUsageError e) ->
+        let failureMessage = BS.unpack $ fromMaybe mempty e in
+        when (("FATAL:  password authentication failed" `isInfixOf` failureMessage) || ("no password supplied" `isInfixOf` failureMessage)) $ do
+          observer $ ExitDBFatalError ServerAuthError err
+          killThread mainThreadId
+      err@(SQL.SessionUsageError (SQL.QueryError tpl _ (SQL.ResultError resultErr))) -> handleResultError err tpl resultErr
+      -- Passing the empty template will not work for schema cache queries, see TODO further below.
```
steve-chavez: Hm, I can't see any TODO below?

IIUC from the lines changed, these new lines are just because Hasql added new constructors.

taimoorzaeem: I think this is the `TODO` they are mentioning. 

https://github.com/PostgREST/postgrest/blob/d31c873cd54d81a5905c5b0508d64cb06d21fb09/src/PostgREST/AppState.hs#L243-L246

steve-chavez: Ah, I see. Thanks for clarifying.

wolfgangwalther:
> Sure, I can look into that. I'll put this in my backlog. Is nixpkgs dropping support for GHC 9.4 anytime soon?

I don't think so, no. Pretty sure that this will not happen before November this year, but maybe even not well into next year. So a lot of time there :)

src/PostgREST/AppState.hs
wolfgangwalther: > IIUC from the lines changed, these new lines are just because Hasql added new constructors.

Correct!

src/PostgREST/AppState.hs
```diff
@@ -215,21 +215,30 @@ initPool AppConfig{..} observer = do
...
+      err@(SQL.SessionUsageError (SQL.PipelineError (SQL.ResultError resultErr)))    -> handleResultError err mempty resultErr
+      SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _)) -> pure ()
```
laurenceisla:
```suggestion
      err@(SQL.SessionUsageError (SQL.QueryError _ _ (SQL.ClientError _)))           -> observer $ QueryErrorCodeHighObs err
```

Error log was missing here, should fix the failing io-tests.

wolfgangwalther:
Moved the update to the latest nixpkgs-unstable version to #4826.

> Also, we can only do this update here, because we pin `hsie` to use GHC 9.4 in the second commit. We _will_ fall over this eventually, when GHC 9.4 is not supported in Nixpkgs anymore - or when we start using syntax that is GHC 9.6+ only at some point. `hsie` is tightly coupled to internal GHC APIs to do its thing, so non-trivial to update. @taimoorzaeem maybe you can look into that? Since we can separate the pin for `hsie` easily, it might make sense - depending on the required changes - to just jump to e.g. GHC 9.12 immediately and don't even work with the intermediate versions. Or, if it's easier do it one version at a time - whatever works best, I guess.

This will be taken care of in #4797, I think.

wolfgangwalther:
> This will be taken care of in #4797, I think.

Nope, `postgrest-lint` fails locally for me, with this branch. Took @mkleczek's hsie changes from #4797 to fix it.

wolfgangwalther:
All jobs passed on the last run, just rebased to resolve a trivial (comment-only) merge conflict. Will merge this now, but will only know whether the actual changes to the release job for the aarch64 artifact work, when the release job on the main branch has run through. Will iterate from there, if that's not the case.

wolfgangwalther:
Had to fix something after merge, but the release has then worked fine. We now have a docker image with both tiny layers consisting of a single executable only. No big ubuntu image for arm anymore.

steve-chavez:
**Edit**: Nvm, already WIP in https://github.com/PostgREST/postgrest/pull/4832
> No big ubuntu image for arm anymore.

I guess we can delete this now https://github.com/PostgREST/postgrest/blob/6b4022b0c1cce9c44dc020c3cc0f4e099a079552/Dockerfile#L1-L21

cc @laurenceisla

PR #4205, opened 2025-07-16
ci: split loadtests into crud and jwt ones

steve-chavez:
There's no need to run the various jwt loadtests in parallel.

Also doing this to see if the jwt summaries are not hidden on Github CI reports.

Loadtests are split on:

- Loadtest CRUD (considering https://github.com/PostgREST/postgrest/issues/4123)
- Loadtest JWT (has all the jwt loadtests)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -110,11 +110,8 @@ jobs:
         run: postgrest-test-memory
 
 
-  loadtest:
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


steve-chavez:
> Also doing this to see if the jwt summaries are not hidden on Github CI reports.

Results don't look good and out of order: https://github.com/PostgREST/postgrest/actions/runs/16331432416?pr=4205

Also the JWT loadtest takes 24 mins which is too long.

To shorten the summaries we should instead show diffs and maybe hide the details with a `<details>` tag.

PR #4207, opened 2025-07-17
nix: avoid updating cabal on every run

### LINKED ISSUE #4206: `postgrest-build` always downloads hackage package list
## Problem

When doing `postgrest-build`:

```bash
$ postgrest-build 
- cabal.project
- cabal.project.freeze
Downloading the latest package list from hackage.haskell.org
```

This downloading takes a while, then:

```
Package list of hackage.haskell.org has been updated.
The index-state is set to 2025-07-16T23:26:44Z.
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2025-07-16T23:14:31Z'
- cabal.project
- cabal.project.freeze
Resolving dependencies...
Build profile: -w ghc-9.4.8 -O1
In order, the following will be built (use -v for more details):
 - postgrest-13.1 (lib) (configuration changed)
 - postgrest-13.1 (test:spec) (dependency rebuilt)
 - postgrest-13.1 (exe:postgrest) (dependency rebuilt)
 - postgrest-13.1 (test:doctests) (dependency rebuilt)
Configuring library for postgrest-13.1...
Warning: [git-protocol] Cloning over git:// might lead to an arbitrary code
execution vulnerability. Furthermore, popular forges like GitHub do not
support it. Use https:// or ssh:// instead.
Preprocessing library for postgrest-13.1...
Building library for postgrest-13.1...
Preprocessing test suite 'doctests' for postgrest-13.1...
Preprocessing executable 'postgrest' for postgrest-13.1...
...
```

Which is the normal part.

This hurts development speed, specially when using `postgrest-watch postgrest-build`. It also happens each time I run `postgrest-test-io`.

@wolfgangwalther Any idea why this is happening? It didn't use to.

wolfgangwalther:
Resolves #4206

@steve-chavez can you try whether that improves it for you?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -44,7 +44,9 @@ jobs:
         uses: ./.github/actions/setup-nix
         with:
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
-          tools: tests.coverage.bin tests.testDoctests.bin tests.testSpecIdempotence.bin
+          tools: tests.coverage.bin tests.testDoctests.bin tests.testSpecIdempotence.bin cabalTools.update.bin
+
+      - run: postgrest-cabal-update
 
       - name: Run coverage (IO tests and Spec tests against PostgreSQL 15)
         run: postgrest-coverage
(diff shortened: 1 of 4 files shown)


steve-chavez (approved):
Works great! :heart: 

The only difference is that now I always get these traces:

```
$ postgrest-build

- cabal.project
- cabal.project.freeze
```

But this looks harmless, it doesn't take too long.

wolfgangwalther:
> The only difference is that now I always get these traces:

These were there before, too - in addition to the same for the v2-update command. Aka they were part of the log twice when I tested earlier today.

PR #4208, opened 2025-07-17
fix: [WIP] jwt claims validation failure on edge cases

taimoorzaeem:
The JWT IO test is failing too often as reported in [#3424 (comment)](https://github.com/PostgREST/postgrest/issues/3424#issuecomment-3077777310)

My guess is that the JWT claims test failure is because we execute the `getTime` action too early in the request pipeline. I have tried moving it a little further to see if that cause the failure to be less frequent. It can be moved even further, but that would take some refactoring and I think we need to discuss this first.

In the meantime, I'll provide another PR to prevent the failure in our CI workflows.


taimoorzaeem:
I have rebased this on #4210 to check if this prevents the failure. With this change, I have run `postgrest-test-io -k test_jwt_errors` for about 20 times and I haven't seen a single failure locally yet. I hope this is the case in CI too.

mkleczek:
> The JWT IO test is failing too often as reported in [#3424 (comment)](https://github.com/PostgREST/postgrest/issues/3424#issuecomment-3077777310)
> 
> My guess is that the JWT claims test failure is because we execute the `getTime` action too early in the request pipeline.

It is certainly **not** the reason as discussed in the other PR.

The reason is that in the previous refactoring code responsible for `exp` validation was changed: the condition was changed from `t + allowedSkewSeconds < now` to `now - allowedSkewSeconds > t` which is not the same because `now` is calculated using `floor`.

https://github.com/PostgREST/postgrest/pull/4212 fixes this.

taimoorzaeem:
> `t + allowedSkewSeconds < now` to `now - allowedSkewSeconds > t`

These two expressions are **mathematically** equivalent.

mkleczek:
> > `t + allowedSkewSeconds < now` to `now - allowedSkewSeconds > t`
> 
> These two expressions are **mathematically** equivalent.

But they are **not** equivalent when `now = toSec time` and `toSec = floor . nominalDiffTimeToSeconds . utcTimeToPOSIXSeconds` because usage of `floor`.

taimoorzaeem:
Closing as no longer needed.

PR #4209, opened 2025-07-17
test: adjust jwt claims error test to avoid failure

taimoorzaeem:
The JWT claims IO test fails too often. This breaks our workflows. This commit temporarily adjusts the test.

We should also backport this to `v13` and `v12` branches.


wolfgangwalther:
> We should also backport this to `v13` and `v12` branches.

We don't run any CI on `v12` anymore, so don't need to go there.

wolfgangwalther:
Your reasoning in #4208 reads as if we're not actually chasing instability of the test-suite here, but rather an actual bug for JWT validation under high load? If that's indeed the case, then I'm not sure whether we should silence it.

taimoorzaeem:
Closing as no longer needed.

taimoorzaeem:
As discussed in #4210, we don't need to strictly test the edge cases. This is enough to test that we tolerate some skew and validation fails outside the allowed range.

PR #4121, opened 2025-06-02
refactor: Extracted JWT validation functions to a separate module.

mkleczek:
As requested in: https://github.com/PostgREST/postgrest/pull/4084#pullrequestreview-2881245976

This change introduces a PostgREST.Auth.Jwt module containing JWT validation functions.

The reason to extract them from Auth is to enable JwtCache module to reuse them without introducing module dependency cycle.

Issue #4035, opened 2025-04-20
make jwt clock skew configurable


steve-chavez:
## Problem

We have a hardcoded 30 second clock skew for JWT validation

https://github.com/PostgREST/postgrest/blob/c087d79f7bb6170ecf0a74303dac9f6111570bd3/docs/references/auth.rst#L171-L172

It would be useful to make it configurable for tests. For example for https://github.com/PostgREST/postgrest/pull/4034#discussion_r2051793740, it would be useful to make the skew shorter (1 second) so the loadtest can run faster (otherwise we would have to wait for 30 seconds before starting the loadtest).

## Solution

Provide a `jwt-skew` config.

steve-chavez:
> Not sure I understand. Can't we just subtract 30 seconds?

Yeah, I just thought it'd be better to not hardcode this value everywhere on tests and instead obtain it from the dump config API endpoint.

Maybe it's for the best to not allow users to change this config too, since it could generate unexpected errors. So I'll close.

PR #4210, opened 2025-07-17
refactor(auth): declutter jwt module

taimoorzaeem:
In #4121 refactor, unnecessary changes like

- type class constraints
- qualified constraints
- language extensions

were added.

This made the code much harder to read and debug. This commit refactors and removes the unnecessary code to debug new issues related to authentication modules. Also reduces the lines of code.

We also need this refactor to continue my work on #4208.


steve-chavez:
> unnecessary changes like
> were added. There is also some dead code in there

Oof.. I wonder how that passed CI, don't we detect dead code for Haskell? (I'm sure we do it for Nix)

Also, is the deleted code used on https://github.com/PostgREST/postgrest/pull/4084? cc @mkleczek

taimoorzaeem:
> don't we detect dead code for Haskell?

I was surprised too. Apparently `postgrest-lint` also **doesn't** detect ALL unused language extensions. :disappointed: 

> Also, is the deleted code used on #4084? 

No, it isn't at all I think. 

#4084 doesn't touch the `Jwt.hs` module. That was good separation of concerns from @mkleczek :rocket:.

src/PostgREST/Auth/Jwt.hs
```diff
@@ -29,87 +28,35 @@ import qualified Jose.Jwt                   as JWT
...
+data JwtTimeClaim = Exp | Iat | Nbf
 
 -- | Receives the JWT secret and audience (from config) and a JWT and returns a
 -- JSON object of JWT claims.
-parseToken :: (MonadError Error m, MonadIO m) => JwkSet -> ByteString -> m JWT.JwtContent
-parseToken _ "" = throwError $ JwtErr $ JwtDecodeErr EmptyAuthHeader
-parseToken secret tkn = do
-  -- secret <- liftEither . maybeToRight (JwtErr JwtSecretMissing) $ configJWKS
-  tknWith3Parts <- hasThreeParts tkn
+parseToken :: AppConfig -> Maybe ByteString -> UTCTime -> ExceptT Error IO JSON.Value
```
mkleczek, line 46: The idea was to make JWT validation as type safe as possible and avoid partial functions.

So missing token cases are handled in another place as this is not really related to JWT validation itself but rather with JWT key set configuration.

So `parseToken` takes `ByteString` as argument and not `Maybe ByteString`

src/PostgREST/Auth/Jwt.hs
```diff
@@ -29,87 +28,35 @@ import qualified Jose.Jwt                   as JWT
...
-import Protolude        hiding (first)
-
-parseAndDecodeClaims :: (MonadError Error m, MonadIO m) => JwkSet -> ByteString -> m JSON.Object
-parseAndDecodeClaims jwkSet token = parseToken jwkSet token >>= decodeClaims
-
-decodeClaims :: MonadError Error m => JWT.JwtContent -> m JSON.Object
-decodeClaims (JWT.Jws (_, claims)) = maybe (throwError (JwtErr $ JwtClaimsErr ParsingClaimsFailed)) pure (JSON.decodeStrict claims)
-decodeClaims _ = throwError $ JwtErr $ JwtDecodeErr UnsupportedTokenType
-
-validateClaims :: MonadError Error m => UTCTime -> Maybe Text -> JSON.Object -> m ()
```
mkleczek, line 55: `validateClaims` is supposed to be separate from parsing and not mix the two.

src/PostgREST/Auth/Jwt.hs
```diff
@@ -29,87 +28,35 @@ import qualified Jose.Jwt                   as JWT
...
-validateClaims time getConfigAud claims = liftEither $ maybeToLeft () (fmap JwtErr . getAlt $ JwtClaimsErr <$> checkForErrors time getConfigAud claims)
```
mkleczek, line 56: Using generic `checkForErrors` allows for easy switching between different error reporting.

For now `getAlt` is used with `Maybe` and only a single claims validation error is reported.

But it is **very** easy now to switch to performing all validations and report a list of errors - it is just a matter of changing `Maybe` to `NotEmpty` monad instance here.

src/PostgREST/Auth/Jwt.hs
```diff
@@ -29,87 +28,35 @@ import qualified Jose.Jwt                   as JWT
...
-
-data ValidAud = VANull | VAString Text | VAArray [Text] deriving Generic
-instance JSON.FromJSON ValidAud where
-  parseJSON JSON.Null = pure VANull
-  parseJSON o = JSON.genericParseJSON JSON.defaultOptions { JSON.sumEncoding = JSON.UntaggedValue } o
-
-checkForErrors :: (Monad m, forall a. Monoid (m a)) => UTCTime -> Maybe Text -> JSON.Object -> m JwtClaimsError
```
mkleczek, line 63: See above - Having `checkForErrors` generic over error reporting monad is a huge win.

src/PostgREST/Auth/Jwt.hs
```diff
@@ -153,3 +140,5 @@ parseClaims AppConfig{configJwtAudience, configJwtRoleClaimKey, configDbAnonRole
     unquoted :: JSON.Value -> BS.ByteString
     unquoted (JSON.String t) = encodeUtf8 t
     unquoted v               = LBS.toStrict $ JSON.encode v
+-- impossible case - just added to please -Wincomplete-patterns
```
mkleczek, line 143: That's exactly the reason why I changed the previous (restored in this PR) implementation - to avoid partial functions.

mkleczek:
> In #4121 refactor, unnecessary changes like
> 
> * type class constraints

I've added them so that the code is generic over:
* error reporting of JWT validation claims (I have a branch with a change to `JwtClaimsError` that reports **all** claims errors instead of only the first one encountered
* `ExceptT IO` where possible (using `MonadError` instead) - not sure what the policy of the project here is but I prefer to code against abstract interface where possible

> * qualified constraints
> * language extensions

This might or might not be a problem - which extensions are problematic?

> 
> were added. There is also some dead code in there. 😢

Hmm... Which lines? The changes I made were supposed to eliminate dead code. Some of it cannot be eliminated due to `Jose` API but it is the same in this version as well?

> 
> This made the code much harder to read and debug. This commit refactors and removes the unnecessary code to debug new issues related to authentication modules. Also reduces the lines of code.
> 
> We also need this refactor to continue my work on #4208.

mkleczek:
I don't believe test flakiness is related to JWT validation code at all to be honest. `getTime` is called in `Auth.hs` and passed to JWT validation functions which are _fast_ (ie. we don't do any IO there and the only CPU intensive logic is actual signature validation that's unavoidable in any case). We also have `30s` clock skew margin which is... a lot for computers. I don't believe JWT validation takes more than 30s either in current or proposed implementations.

The only thing that comes to mind is a bug in time validation logic but that would be caught by unit tests.

taimoorzaeem:
> I don't believe test flakiness is related to JWT validation code at all to be honest. `getTime` is called in `Auth.hs` and passed to JWT validation functions which are _fast_ [...]

The flakiness doesn't happen on #4208 anymore. So, I think that is evidence that the early execution of `getTime` action was indeed the issue.

### Detailed Explanation, what might be happening under the hood:

Notice that the it was always the `exp` test failing. The POSIX seconds value in our haskell code is actually a fractional value on which we call `floor`. Now with millisecond difference, for instance `9.99` is floored to `9` and `10.01` is floored to 10. 

This caused the issue only for the `exp` claim but never for `nbf` and `iat` claims because they are verified in the opposite time direction. So, executing the `getTime` just before the verification time reduces the chance of flooring getting two different values significantly.  I hope this makes sense.

src/PostgREST/Auth/Jwt.hs
taimoorzaeem: > The idea was to make JWT validation as type safe as possible and avoid partial functions.

I don't see any partial functions here. They are total, just not pure, which is the whole point.

mkleczek:
> This caused the issue only for the `exp` claim but never for `nbf` and `iat` claims because they are verified in the opposite time direction. So, executing the `getTime` just before the verification time reduces the chance of flooring getting two different values significantly. I hope this makes sense.

This does not make sense as the probability of JWT being expired is greater when you postpone `getTime`. The earlier you invoke `getTime` the more probable is JWT being valid.

wolfgangwalther:
> This does not make sense as the probability of JWT being expired is greater when you postpone `getTime`. The earlier you invoke `getTime` the more probable is JWT being valid.

Arguably the expiry of a JWT should be checked as close as possible to the actual query of the data - since it's checked before, it should be checked as late as possible. `getTime` *is* essentially that check.

src/PostgREST/Auth/Jwt.hs
```diff
@@ -29,87 +28,35 @@ import qualified Jose.Jwt                   as JWT
...
-  , claim "aud" AudClaimNotStringOrArray checkAud
-  ]
-  where
-      allowedSkewSeconds = 30 :: Int64
-      sciToInt = fromMaybe 0 . Sci.toBoundedInteger
-      toSec = floor . nominalDiffTimeToSeconds . utcTimeToPOSIXSeconds
-      now = toSec time
-
-      inTheFuture = checkTime ((now + allowedSkewSeconds) <)
-      inThePast = checkTime ((now - allowedSkewSeconds) >)
```
mkleczek, line 78: @taimoorzaeem 
I think this might be the reason for flaky tests.
I've changed it from `(now + allowedSkewSeconds) > sciToInt secs`

mkleczek: See:
https://github.com/PostgREST/postgrest/pull/4212

mkleczek:
> > This does not make sense as the probability of JWT being expired is greater when you postpone `getTime`. The earlier you invoke `getTime` the more probable is JWT being valid.
> 
> Arguably the expiry of a JWT should be checked as close as possible to the actual query of the data - since it's checked before, it should be checked as late as possible. `getTime` _is_ essentially that check.

Right - but this is a completely different issue from what we are discussing here.

src/PostgREST/Auth/Jwt.hs
mkleczek: > > The idea was to make JWT validation as type safe as possible and avoid partial functions.
> 
> I don't see any partial functions here. They are total, just not pure, which is the whole point.

They are partial but with dead cases added to make the compiler happy (it is even explicitly documented in the comments)

src/PostgREST/Auth/Jwt.hs
```diff
@@ -121,26 +68,66 @@ parseToken secret tkn = do
       -- Control never reaches here, the decode function only returns the above three
       jwtDecodeError _                    = JwtDecodeErr UnreachableDecodeError
 
-parseClaims :: (MonadError Error m, MonadIO m) => AppConfig -> UTCTime -> JSON.Object -> m AuthResult
-parseClaims AppConfig{configJwtAudience, configJwtRoleClaimKey, configDbAnonRole} time mclaims = do
-  validateClaims time configJwtAudience mclaims
+      verifyClaims :: JWT.JwtContent -> Either JwtError JSON.Value
```
mkleczek, line 71: Input type `JWT.JwtContent` is too general and hence we have mixed claims validation logic with claims parsing logic.

In the previous implementation these are separated into two functions:

`decodeClaims` responsible for decoding `JWT.JwtContent` content into `JSON.Object`, returning error if it is not possible
and
`validateClaims` responsible **only** for validation of decoded claims

src/PostgREST/Auth/Jwt.hs
wolfgangwalther: Was this change in v12 as well? The tests were much more flaky in that branch before I disabled CI there. IIRC, the refactor only came in afterwards, so if anything this change would have helped it.

mkleczek: > Was this change in v12 as well? The tests were much more flaky in that branch before I disabled CI there. IIRC, the refactor only came in afterwards, so if anything this change would have helped it.

No, it didn't land in v12 - only `main`

steve-chavez:
> We also have 30s clock skew margin which is... a lot for computers

@mkleczek The reasoning for 30 seconds is mentioned on https://github.com/PostgREST/postgrest/pull/2772, prior to that we used to get several error reports.. we haven't got any since then.

Since the failing CI test is the one related to the skew, maybe it makes sense to make the skew configurable (https://github.com/PostgREST/postgrest/issues/4035) so the test is faster? Maybe this could solve the issue too?

wolfgangwalther:
> The reasoning for 30 seconds is mentioned on #2772, prior to that we used to get several error reports.. we haven't got any since then.

Looks like it doesn't really matter whether the tolerated skew is 29, 30 or 31 seconds. I mean.. it's expired *anyway*.

Maybe we should just test cases of something like 15 and 45 seconds, so well before and well after. We don't need precision here, we just need to test the fact *some amount of skew is tolerated*. That's all.

Edit: Or put differently: By testing for 31 seconds, while tolerating 30 seconds, we are effectively *testing a skew of 1 second*. Which the intent was to *not* have, precisely because it's not enough for real use cases...

steve-chavez:
> Maybe we should just test cases of something like 15 and 45 seconds, so well before and well after. We don't need precision here, we just need to test the fact some amount of skew is tolerated. That's all.

Right, I agree. We just need to test tolerance. Let's fix this by adjusting the tests and not modify core logic then. @taimoorzaeem WDYT?

I think this also goes in line with what @mkleczek concluded here https://github.com/PostgREST/postgrest/pull/4212#issuecomment-3088818842.

taimoorzaeem:
> Right, I agree. We just need to test tolerance. Let's fix this by adjusting the tests and not modify core logic then. @taimoorzaeem WDYT?

Hmm, makes sense. I think the 30 second bracket is a good tolerance range. Let's just rewrite our tests with `-35` and `+35` for validation failing cases and `-25` and `+25` for validation passing cases to verify the tolerance. No need for precise edge case testing.

taimoorzaeem:
Closing as no longer required.

PR #4211, opened 2025-07-18
test: remove jwt decoding from mixed loadtest

steve-chavez:
We now have dedicated JWT loadtests, so it's no longer necessary to mix this here.

First step towards solving https://github.com/PostgREST/postgrest/issues/4123


steve-chavez:
`mixed` throughput looks mostly the same compared to last commit:

- Old: https://github.com/PostgREST/postgrest/actions/runs/16351010465
- New: https://github.com/PostgREST/postgrest/actions/runs/16359471385?pr=4211

Only the memory seems a bit higher on old, which makes sense since both have the JWT cache enabled.

**Edit**: seems fine to merge since no old loadtests wil be impacted.

steve-chavez:
PR failure related to https://github.com/PostgREST/postgrest/pull/4208

PR #4212, opened 2025-07-18
fix: rounding error in exp claim validation causing tests flakiness

mkleczek:
Should fix issues described in: https://github.com/PostgREST/postgrest/issues/3424#issuecomment-3077777310

Changes in JWT module on which https://github.com/PostgREST/postgrest/pull/4208 is based do not look like necessary to fix it.

### DIFF
--- a/src/PostgREST/Auth/Jwt.hs
+++ b/src/PostgREST/Auth/Jwt.hs
@@ -75,9 +75,10 @@ checkForErrors time cfgAud = mconcat
       now = toSec time
 
       inTheFuture = checkTime ((now + allowedSkewSeconds) <)
-      inThePast = checkTime ((now - allowedSkewSeconds) >)
+      -- add 1 to account for floor in toSec (see above)
+      inThePast = checkTime ((now - allowedSkewSeconds + 1) >)
 
-      checkTime cond = checkValue (cond. sciToInt)
+      checkTime cond = checkValue (cond . sciToInt)
 
       checkAud = \case
         (VAString aud)                     -> liftMaybe cfgAud >>= checkValue (aud /=) JWTNotInAudience


mkleczek:
@taimoorzaeem
You added thumbs down emoji here. What is the reasoning? It is a simple change that makes the tests pass without issues.

Any refactorings to the JWT module can be performed in a separately discussed PR. The discussions would not stall fixing the tests flakiness.

taimoorzaeem:
See, passing the test isn't the only goal here. Correctness is more important. Adding `1` doesn't solve the issue here, it justs offsets the values by 1, plus it has no impact on the `floor` function.

We are testing the edge case here. After adding 1, the test isn't an edge case anymore.

mkleczek:
> See, passing the test isn't the only goal here. Correctness is more important. Adding `1` doesn't solve the issue here, it justs offsets the values by 1, plus it has no impact on the `floor` function.
> 
> We are testing the edge case here. After adding 1, the test isn't an edge case anymore.

So I would say - let's fix usage of the `floor` function then as this is **the** root cause.

It does **not** require large refactoring, only change of the definition of `toSec = floor . nominalDiffTimeToSeconds . utcTimeToPOSIXSeconds`.

The problem is present in the codebase for a very long time as `floor` usage is old. The tests used to pass and this change makes them pass again.

Indeed: introduction of `inThePast` binding makes the issue with this logic (and tests!) aparent:
tests pass with `inThePast = (now - allowedSkewSeconds + 1 >)` but fail when it is `inThePast = (now - allowedSkewSeconds >)` - but it is the second definition that seems correct.

Maybe it is time to revisit the tests?
@taimoorzaeem @steve-chavez - WDYT?

steve-chavez:
> Maybe it is time to revisit the tests?

Right, I agree as per reasoning on https://github.com/PostgREST/postgrest/pull/4210#issuecomment-3089357458

wolfgangwalther:
So, if we tolerate 30 seconds skew, does it matter how we use floor to have precision for +-1 second? I don't think so, because we already say that precision is only guaranteed on the half minute scale.

If we want to make the `floor` any more "right", then we should just properly round instead of flooring.

mkleczek:
@wolfgangwalther exactly

I will close this PR and as discussed - we’ll adjust the tests.

wolfgangwalther:
> we’ll adjust the tests.

Done in #4209. +5 seconds should do it.

PR #4219, opened 2025-07-22
add: override PGRST_CMD for postgrest-loadtest from env

develop7:
Enables `PGRST_CMD=postgrest-profiled-run postgrest-loadtest` running loadtest against profiled executable, which wasn't available before

### DIFF
--- a/nix/tools/withTools.nix
+++ b/nix/tools/withTools.nix
@@ -348,20 +348,22 @@ let
       ''
         export PGRST_SERVER_UNIX_SOCKET="$tmpdir"/postgrest.socket
 
-        rm -f result
-        if [ -z "''${PGRST_BUILD_CABAL:-}" ]; then
-          echo -n "Building postgrest (nix)... "
-          # Using lib.getBin to also make this work with older checkouts, where .bin was not a thing, yet.
-          nix-build -E 'with import ./. {}; pkgs.lib.getBin postgrestPackage' > "$tmpdir"/build.log 2>&1 || {
-            echo "failed, output:"
-            cat "$tmpdir"/build.log
-            exit 1
-          }
-          PGRST_CMD=$(echo ./result*/bin/postgrest)
-        else
-          echo -n "Building postgrest (cabal)... "
-          postgrest-build
-          PGRST_CMD=postgrest-run
+        if [ -z "''${PGRST_CMD:-}" ]; then
+          rm -f result
+          if [ -z "''${PGRST_BUILD_CABAL:-}" ]; then
+            echo -n "Building postgrest (nix)... "
+            # Using lib.getBin to also make this work with older checkouts, where .bin was not a thing, yet.
+            nix-build -E 'with import ./. {}; pkgs.lib.getBin postgrestPackage' > "$tmpdir"/build.log 2>&1 || {
+              echo "failed, output:"
+              cat "$tmpdir"/build.log
+              exit 1
+            }
+            PGRST_CMD=$(echo ./result*/bin/postgrest)
+          else
+            echo -n "Building postgrest (cabal)... "
+            postgrest-build
+            PGRST_CMD=postgrest-run
+          fi
         fi
         echo "done."
 


steve-chavez:
> PGRST_CMD=postgrest-profiled-run postgrest-loadtest

The above is real nice, we can use it to improve perf.

Looking at this https://github.com/PostgREST/postgrest/pull/4220#issuecomment-3106755359, is this PR dependent on #4220? Or how do you confirm this PR is working as expected?

develop7:
> Looking at this [#4220 (comment)](https://github.com/PostgREST/postgrest/pull/4220#issuecomment-3106755359), is this PR dependent on #4220? Or how do you confirm this PR is working as expected?

It doesn't depend on #4220, I've merely used `PGRST_CMD=postgrest-profiled-run postgrest-loadtest` to verify it works. One could use `/usr/bin/true` just as well.

nix/tools/withTools.nix
```diff
@@ -348,20 +348,22 @@ let
...
-            echo "failed, output:"
-            cat "$tmpdir"/build.log
-            exit 1
-          }
-          PGRST_CMD=$(echo ./result*/bin/postgrest)
-        else
-          echo -n "Building postgrest (cabal)... "
-          postgrest-build
-          PGRST_CMD=postgrest-run
+        if [ -z "''${PGRST_CMD:-}" ]; then
```
steve-chavez, line 374: How about using argbash `ARG_USE_ENV` instead?

There's an example on https://github.com/PostgREST/postgrest/blob/cc4f92005c401358d2f84e40d45c7a8fc4633faa/nix/tools/withTools.nix#L28-L32

steve-chavez: This way we also document the command has the PGRST_CMD option when doing `--help`

develop7: And have it documented too! Thanks for the suggestion, will do!

nix/tools/withTools.nix
```diff
@@ -339,6 +339,7 @@ let
             "ARG_POSITIONAL_SINGLE([command], [Command to run])"
             "ARG_LEFTOVERS([command arguments])"
             "ARG_OPTIONAL_SINGLE([monitor], [m], [Enable CPU and memory monitoring of the PostgREST process and output to the designated file as markdown])"
+            "ARG_USE_ENV([PGRST_CMD], [postgrest-run], [PostgREST executable to run])"
```
wolfgangwalther: With the default of `postgrest-run`, this will change the logic heavily now. We will now have to explicitly set this to the empty string to get the default behavior from before.

Also, we can't make use of `PGRST_BUILD_CABAL` anymore.

I think we should not control this via env variable at all, but instead add a `--profiled` flag to `postgrest-loadtest` and then use the profiled package in both variants if this is set.

nix/tools/withTools.nix
```diff
@@ -373,6 +376,7 @@ let
         pid=$!
         # shellcheck disable=SC2317
         cleanup() {
+          kill -HUP "$pid" #
```
wolfgangwalther: Did you mean to add a comment here?

Why do we need this?

develop7: Sorry, it leaked from another branch while I've been experimenting with handling multiple features with jujutsu. The experiment is a massive success BTW, https://jj-vcs.github.io/ freaking rocks in no uncertain terms.

nix/tools/withTools.nix
develop7: `PGRST_CMD` env argument doesn't have a default value anymore and should be good to go now

re: separate flag> I agree this is a cleaner behavior, but it'd require to propagate the desired executable name (path) from `pgrst-loadtest` to `postgrest-with-pgrst` script, which I'm yet to wrap my head around of.

nix/tools/withTools.nix
develop7: Done

nix/tools/withTools.nix
develop7: I've updated it to empty value by default, please review

nix/tools/withTools.nix
```diff
@@ -370,20 +371,22 @@ let
...
+              exit 1
+            }
+            PGRST_CMD=$(echo ./result*/bin/postgrest)
+          else
+            echo -n "Building postgrest (cabal)... "
+            postgrest-build
+            PGRST_CMD=postgrest-run
+          fi
         fi
         echo "done."
```
wolfgangwalther: The `done.` message belongs to the `Building postgrest (xxx) ...` opener, so it should only be printed inside the if, right?

develop7: Makes sense, otherwise there could be a lone `echo 'done'` when nothing was done in a hypothetical case of empty `PGRST_CMD`

UPD: Fixed, please review

PR #4222, opened 2025-07-23
fix: do not retry transactions on `40001` SQL error

### LINKED ISSUE #3673: Surprising retrying of SQL statement on a replica causes replication lag
## Problem

On a read replica, this happens:

```
GET /rpc/xx..

Jul 22 23:21:04 [redacted] postgrest[115960]: 22/Jul/2024:23:21:04 +0000: 
{"code":"40001",
"details":"User query might have needed to see row versions that must be removed.",
"hint":null,
"message":"canceling statement due to conflict with recovery"}
```

On the postgres logs, this error is repeated a number of times:

```
..,ERROR,40001,"canceling statement due to conflict with recovery","User query might have needed to see row versions that must be removed.",,
PL/pgSQL function xx() line 3 at RETURN QUERY","WITH pgrst_source AS (SELECT ""pgrst_call"".
...
```

For some reason this causes noticeable replication lag, I'm assuming this error somehow affects the replication process. It can only be solved if postgrest is stopped. 

The problem root cause can be found on `hasql-transaction`:

- https://github.com/nikita-volkov/hasql-transaction/blob/master/CHANGELOG.md
- https://github.com/nikita-volkov/hasql-transaction/blob/2a9a710bd6230cb0abae84f14f852ab14fc3fa85/library/Hasql/Transaction/Private/Sessions.hs#L48

## Solution

Don't retry on 40001. We would need to patch `hasql-transaction` to make the retrying behavior configurable and then disable it.

laurenceisla:
Should close #3673.

- [x] Fix error upstream: https://github.com/nikita-volkov/hasql-transaction/issues/22
- [x] Update dependencies upstream to newer `hasql` version: https://github.com/diogob/hasql-notifications/pull/30
- [x] Update PostgREST to newer library versions and use a non-retryable transaction to fix issue (this is being worked on in: #4829)
- [x] Add tests

### DIFF
--- a/nix/tools/withTools.nix
+++ b/nix/tools/withTools.nix
@@ -105,7 +105,8 @@ let
 
             log "Starting replica on $replica_host"
 
-            pg_ctl -D "$replica_dir" -l "$replica_dblog" -w start -o "-F -c listen_addresses=\"\" -c hba_file=$HBA_FILE -k $replica_host -c log_statement=\"all\" " \
+            # We set a low max_standby_streaming_delay to make the replication conflict fail faster in tests (otherwise it waits for the default 30s)
+            pg_ctl -D "$replica_dir" -l "$replica_dblog" -w start -o "-F -c listen_addresses=\"\" -c hba_file=$HBA_FILE -k $replica_host -c log_statement=\"all\" -c max_standby_streaming_delay=\"3s\" " \
               >> "$setuplog"
 
             >&2 echo "${commandName}: Replica enabled. You can connect to it with: psql 'postgres:///$PGDATABASE?host=$replica_host' -U postgres"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file. From versio
 
 - Shutdown should wait for in flight requests by @mkleczek in #4702
 - Fix login with uppercase and mixed case role names by @taimoorzaeem in #4678
+- Remove automatic transaction retries on `40001 (serialization_failure)` errors to prevent replication lag by @laurence in #3673
 
 ### Changed
 
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -343,7 +343,7 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
     qSchemaCache = do
       conf@AppConfig{..} <- getConfig appState
       (resultTime, result) <-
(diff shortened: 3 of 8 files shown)


steve-chavez:
Should we vendor hasql-notifications for now since there's no response for https://github.com/diogob/hasql-notifications/pull/30 ?

I remember we did this before too. It's a single file so no major maintenance problem there IMO.

wolfgangwalther:
We should not vendor it, for two reasons:
- First of all, we haven't even done our own work, yet. We can't update to hasql 1.8 or 1.9 before the static build is sorted out - and it currently doesn't look like there will be progress soon. The upstream Nixpkgs PR that #4193 depends on is likely not going to get merged - and currently Nixpkgs is not even able to build `pkgsStatic.libpq` anmyore. This only worked by accident. Thus, upgrading to GHC 9.6 and then hasql is quite far out of reach for us right now.
- 3 weeks without reaction is not "no reaction", imho. Especially when there has been no additional ping in the PR, yet, etc.

laurenceisla:
> 3 weeks without reaction is not "no reaction", imho. Especially when there has been no additional ping in the PR, yet, etc.

Thanks for the ping suggestion (I should've done this weeks ago). Changes are now merged and released upstream!

>  First of all, we haven't even done our own work, yet. We can't update to hasql 1.8 or 1.9 before the static build is sorted out - and it currently doesn't look like there will be progress soon.

OK, so this would be the only blocker for this PR for now then. Also some `stack` shenanigans, but will fix after this is sorted.

nix/tools/withTools.nix
```diff
@@ -105,7 +105,7 @@ let
 
             log "Starting replica on $replica_host"
 
-            pg_ctl -D "$replica_dir" -l "$replica_dblog" -w start -o "-F -c listen_addresses=\"\" -c hba_file=$HBA_FILE -k $replica_host -c log_statement=\"all\" " \
+            pg_ctl -D "$replica_dir" -l "$replica_dblog" -w start -o "-F -c listen_addresses=\"\" -c hba_file=$HBA_FILE -k $replica_host -c log_statement=\"all\" -c max_standby_streaming_delay=\"3s\" " \
```
steve-chavez, line 109: @laurenceisla Maybe you can add a comment explaining why do we need this setting?

steve-chavez (approved):

steve-chavez:
@laurenceisla Possible to backport? Would it be hard because it depends on a newer nixpkgs?

laurenceisla:
> Would it be hard because it depends on a newer nixpkgs?

Yes, I'm afraid it will result in a similar case as https://github.com/PostgREST/postgrest/pull/4702#issuecomment-4284702994, due to the dependencies.

Issue #3166, opened 2024-01-16
Partial document update based on JSON Patch


steve-chavez:
## Problem

Currently there's no way to do a partial document update. 

- https://github.com/PostgREST/postgrest/issues/465#issuecomment-1198884375 discusses json patch but there are certain operations (like `test`) which don't make much sense given our horizontal filtering. 
- An alternative is proposed on https://github.com/PostgREST/postgrest/issues/465#issuecomment-1199858002, but it's flexibility makes it harder to implement. Additionally it's not isolated to a single content type.

## Solution

Use a vendored media type based on json patch where we support the operations we lack. As prior art, [azure cosmosdb json patch](https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update#supported-operations) (also see [here](https://learn.microsoft.com/en-us/rest/api/cosmos-db/patch-a-document)) does a similar thing and adds support for `increment` and `set`, which are not defined in the [JSON Patch RFC](https://www.rfc-editor.org/rfc/rfc6902.html).

For starters, we can add support for increment. Like so:

```http
PATCH /items?id=eq.1
Content-Type: application/vnd.pgrst.json-patch

[
  { "op": "incr", "path": "/inventory", "value": 10 },
]
```

Path would refer to any column and the generated SQL would be like `UPDATE items SET inventory = inventory + 10`.

Later on we can extend this to add other operations, like `mult`, `div`, `concat`, etc.

JanKaul:
Am I assuming correctly that the implementation would involve a new [media type handler](https://postgrest.org/en/stable/references/api/media_type_handlers.html#media-type-handlers) and a [postgres implementation of jsonpatch](https://gist.github.com/InfoSec812/b830a9db4c9048552f8c51d7987cc4d0)?

What other steps would be required to implement this functionality as a builtin media type?

JanKaul:
That makes sense. Thanks for the clarification. I will have a look at the implementation of `application/vnd.pgrst.plan`.

steve-chavez:
Taking advantage of the [interval type](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT) in postgres, it would be interesting to have an `interval_pl` operator :

```http
PATCH /items?id=eq.1
Content-Type: application/vnd.pgrst.json-patch

[
  { "op": "interval_pl", "path": "/date", "value": "1 day 15 minute 40 seconds"},
]

``` 

That would get translated roughly to:

[...]

PR #4225, opened 2025-07-25
refactor: move request body to a Payload.hs module

taimoorzaeem:
For #3166, I need to parse the request body. Just like we have a module for parsing `QueryParams`, request body can be separated into a `Payload` module easier implementation of #3166. It would also help not polluting the `ApiRequest` module with too much paring logic.

### DIFF
--- a/src/PostgREST/ApiRequest/Payload.hs
+++ b/src/PostgREST/ApiRequest/Payload.hs
@@ -1,0 +1,138 @@
+-- |
+-- Module      : PostgREST.ApiRequest.Payload
+-- Description : Parser for PostgREST Request Body
+--
+-- This module is in charge of parsing the request body (payload)
+--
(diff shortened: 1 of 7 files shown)


src/PostgREST/ApiRequest/Payload.hs
```diff
@@ -0,0 +1,195 @@
...
+                                          QualifiedIdentifier (..),
+                                          Schema)
+
+import qualified PostgREST.MediaType as MediaType
+
+import Protolude
+
+-- TODO: Maybe move these Types to ApiRequest.Types module to have single source
+--       of import and avoid re-exporting
+data InvokeMethod
```
steve-chavez: Hm, could we do this now? These types look out of place here.

steve-chavez (approved):
LGTM.  Good thinking separating the modules before adding more features :+1:

PR #4227, opened 2025-07-28
add: Pgrst Patch for partial document update

### LINKED ISSUE #3166: Partial document update based on JSON Patch

taimoorzaeem:
This should close #3166.

- [x] Implementation
- [ ] Tests (coverage)
- [ ] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+
+- Improve the `PGRST106` error when the requested schema is invalid by @laurenceisla in #4089
+  + It now shows the invalid schema in the `message` field.
+  + The exposed schemas are now listed in the `hint` instead of the `message` field.
+- Improve error details of `PGRST301` error by @taimoorzaeem in #4051
+- Bounded JWT cache using the SIEVE algorithm by @mkleczek in #4084
+- Implement `PGRST Patch` for partial document update by @taimoorzaeem in #3166
+
 ### Fixed
 
 - Fix not logging OpenAPI queries when `log-query=main-query` is enabled by @steve-chavez in #4226
--- a/src/PostgREST/ApiRequest/Payload.hs
+++ b/src/PostgREST/ApiRequest/Payload.hs
@@ -44,6 +44,7 @@ getPayload reqBody contentMediaType QueryParams{qsColumns} action = do
         (Just ProcessedJSON{payKeys}, _)       -> payKeys
         (Just ProcessedUrlEncoded{payKeys}, _) -> payKeys
         (Just RawJSON{}, Just cls)             -> cls
+        (Just PgrstPatch{payFields}, _)        -> payFields
         _                                      -> S.empty
   return (checkedPayload, cols)
   where
@@ -69,6 +70,12 @@ getPayload reqBody contentMediaType QueryParams{qsColumns} action = do
       (MTTextPlain, True) -> Right $ RawPay reqBody
       (MTTextXML, True) -> Right $ RawPay reqBody
       (MTOctetStream, True) -> Right $ RawPay reqBody
+      (MTVndPgrstPatch, False) ->
+        if isJust columns
+          then Right $ RawJSON reqBody
+          -- Error message too generic?
+          else note "All objects should contain 3 key-vals: 'op','path' and 'value', where op and path must be a string"
+            (pgrstPatchPayloadFields reqBody =<< JSON.decode reqBody)
       (ct, _) -> Left $ "Content-Type not acceptable: " <> MediaType.toMime ct
 
     shouldParsePayload = case action of
@@ -78,10 +85,10 @@ getPayload reqBody contentMediaType QueryParams{qsColumns} action = do
       _                                       -> False
 
     columns = case action of
-      ActDb (ActRelationMut _ MutationCreate) -> qsColumns
-      ActDb (ActRelationMut _ MutationUpdate) -> qsColumns
-      ActDb (ActRoutine     _ Inv)            -> qsColumns
-      _                                       -> Nothing
+      ActDb (ActRelationMut _ MutationCreate)     -> qsColumns
+      ActDb (ActRelationMut _ (MutationUpdate _)) -> qsColumns
+      ActDb (ActRoutine     _ Inv)                -> qsColumns
+      _                                           -> Nothing
 
     isProc = case action of
       ActDb (ActRoutine _ _) -> True
@@ -136,3 +143,45 @@ payloadAttributes raw json =
     _ -> Just emptyPJArray
   where
     emptyPJArray = ProcessedJSON (JSON.encode emptyArray) S.empty
+
+-- Here, we verify the following about pgrst patch body:
+-- 1. The JSON must be a json array.
+-- 2. All objects in the array must have only these three fields:
+--    'op', 'path', 'value'.
+-- 3. Finally, extract the 'path' values as fields
+--
+-- TODO: Return (Either ByteString Payload) for better error messages
+pgrstPatchPayloadFields :: RequestBody -> JSON.Value -> Maybe Payload
+pgrstPatchPayloadFields raw (JSON.Array arr) =
+      if V.all isValidPatchObject arr
+        then PgrstPatch raw . S.fromList <$> getPaths arr
+        else Nothing
+      where
+        isValidPatchObject (JSON.Object o) =
+          KM.member "op" o    &&
+          KM.member "path" o  &&
+          KM.member "value" o &&
+          length (KM.keys o) == 3
+        isValidPatchObject _ = False
+
+        getPaths :: V.Vector JSON.Value -> Maybe [Text]
+        getPaths ar = if any isNothing maybePaths || not (all extractOp $ V.toList ar)
+                        then Nothing
+                        else Just $ catMaybes maybePaths
+          where
+            maybePaths :: [Maybe Text]
+            maybePaths = map extractPath $ V.toList ar
+
+            extractOp (JSON.Object o) =
+              case KM.lookup "op" o of
+                Just op -> op == "set" -- we only have "set" operation, for now
+                Nothing -> False
+            extractOp _ = False
+
+            extractPath (JSON.Object o) =
+              case KM.lookup "path" o of
+                Just (JSON.String path) -> Just path
+                _                       -> Nothing
+            extractPath _ = Nothing
+
+pgrstPatchPayloadFields _ _ = Nothing
--- a/src/PostgREST/ApiRequest/Types.hs
+++ b/src/PostgREST/ApiRequest/Types.hs
@@ -55,8 +55,12 @@ data Mutation
   = MutationCreate
   | MutationDelete
   | MutationSingleUpsert
-  | MutationUpdate
+  | MutationUpdate PgrstPatch
+  -- ^ We have two types of updates, regular updates
+  --   and json patch style updates
   deriving Eq
+
+type PgrstPatch = Bool
 
 data Resource
   = ResourceRelation Text
@@ -91,6 +95,10 @@ data Payload
   | ProcessedUrlEncoded { payArray  :: [(Text, Text)], payKeys :: S.Set Text }
   | RawJSON { payRaw  :: LBS.ByteString }
   | RawPay  { payRaw  :: LBS.ByteString }
+  | PgrstPatch
+      { payRaw    :: LBS.ByteString
+      , payFields :: S.Set Text -- ^ These are columns that are to be patched.
+      }
 
 
 -- | The value in `/tbl?select=alias:field.aggregateFunction()::cast`
--- a/src/PostgREST/MediaType.hs
+++ b/src/PostgREST/MediaType.hs
@@ -37,6 +37,7 @@ data MediaType
   -- vendored media types
   | MTVndArrayJSONStrip
   | MTVndSingularJSON Bool
+  | MTVndPgrstPatch
   -- TODO MTVndPlan should only have its options as [Text]. Its ResultAggregate should have the typed attributes.
   | MTVndPlan MediaType MTVndPlanFormat [MTVndPlanOption]
   deriving (Eq, Show, Generic, JSON.ToJSON)
@@ -72,6 +73,7 @@ toMime MTTextXML              = "text/xml"
 toMime MTOpenAPI              = "application/openapi+json"
 toMime (MTVndSingularJSON True)  = "application/vnd.pgrst.object+json;nulls=stripped"
 toMime (MTVndSingularJSON False) = "application/vnd.pgrst.object+json"
+toMime MTVndPgrstPatch        = "application/vnd.pgrst.patch+json"
 toMime MTUrlEncoded           = "application/x-www-form-urlencoded"
 toMime MTOctetStream          = "application/octet-stream"
 toMime MTAny                  = "*/*"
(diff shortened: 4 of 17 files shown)


src/PostgREST/ApiRequest/Payload.hs
```diff
@@ -136,3 +145,24 @@ payloadAttributes raw json =
...
+  parseJSON (JSON.Object o) = do
+    op    <- parseString o "op"
+    path  <- parseString o "path"
+    -- TODO: We need to decide what JSON "value"s are allowed in our
+    --       our JSON Patch implementation. For now, we only allow string.
+    case op of
+      "incr"        -> Incr     path <$> parseString o "value"
+      "replace"     -> Replace  path <$> parseString o "value"
+      "interval_pl" -> Interval path <$> parseString o "value"
+      _         -> fail $ "Unknown JSON Patch operation " ++ show op
```
steve-chavez: :eyes: I would suggest just focusing on `incr` for this PR, easier implementation and review.

src/PostgREST/MediaType.hs
```diff
@@ -72,6 +73,7 @@ toMime MTTextXML              = "text/xml"
 toMime MTOpenAPI              = "application/openapi+json"
 toMime (MTVndSingularJSON True)  = "application/vnd.pgrst.object+json;nulls=stripped"
 toMime (MTVndSingularJSON False) = "application/vnd.pgrst.object+json"
+toMime MTVndJSONPatch         = "application/vnd.pgrst.json-patch"
```
steve-chavez: It'd be clearer to name this as `application/vnd.pgrst.patch+json`.

Azure cosmosdb uses `application/json_patch+json`
https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update#rest-api-reference-for-partial-document-update

src/PostgREST/Plan/MutatePlan.hs
```diff
@@ -38,3 +39,9 @@ data MutatePlan
       , where_    :: [CoercibleLogicTree]
       , returning :: [FieldName]
       }
+  | JSONPatch -- JSON Patch for partial document update (not fully RFC compliant for now)
```
steve-chavez: I don't think we'll ever be RFC compliant here, we could call this `JSONPatchesque` or just `PgrstPatch`.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -161,13 +163,45 @@ mutatePlanToQuery (Update mainQi uCols body logicForest returnings applyDefaults
...
+        getFieldName (Incr field _) = pgFmtIdent field
+
+    vals :: [SQL.Snippet]
+    vals = map getValAndApplyOp patchBody
+      where
+        intToSnip = SQL.sql . BS.pack . show
+
+        getValAndApplyOp :: PgrstPatchOp -> SQL.Snippet
+        getValAndApplyOp (Incr field val) =
+          pgFmtIdent field <> " + CAST(" <> intToSnip val <> " AS INTEGER)"
```
wolfgangwalther: You should not put the `val` directly into the query string like this. We expect **all** values to be passed as parameters by hasql.

src/PostgREST/Plan/MutatePlan.hs
```diff
@@ -38,3 +38,9 @@ data MutatePlan
       , where_    :: [CoercibleLogicTree]
       , returning :: [FieldName]
       }
+  | PgrstPatch -- A modified version of JSON Patch
+      { in_       :: QualifiedIdentifier
+      , patchBody :: Maybe [PgrstPatchOp]
+      , where_    :: [CoercibleLogicTree]
+      , returning :: [FieldName]
+      }
```
wolfgangwalther: I think we should not add a new `MutatePlan` here, but instead generalize the `Update` plan to contain "operations".

The default "update" that we currently support, should then be a special case "operation". This operation just uses the value from the json body.

Thus, I think these steps would make sense:
- First refactor/generalize the Update method
- Then implement the pgrst patch mimetype with a basic "set" operation (which is the same as the default update, just via the custom mimetype structure)
- Last implement the new "incr" operation.

src/PostgREST/ApiRequest/Payload.hs
```diff
@@ -6,6 +6,7 @@
 --
 {-# LANGUAGE LambdaCase     #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
wolfgangwalther: Why do we need it and can we avoid it? We really should.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   + The exposed schemas are now listed in the `hint` instead of the `message` field.
 - Improve error details of `PGRST301` error by @taimoorzaeem in #4051
 - Bounded JWT cache using the SIEVE algorithm by @mkleczek in #4084
+- Add partial document update using PGRST Patch by @taimoorzaeem in #3166
```
wolfgangwalther: It's entirely unclear what "PGRST Patch" is in this note, when not familiar with the PR.

src/PostgREST/ApiRequest/Payload.hs
taimoorzaeem: That's because of GHC doesn't allow defining instances if the type class or the type itself is also not in the same module. We use this in other places as well. However, yes we really should avoid it and a simple solution is to just move the type from `Types` module to `Payload` module.

src/PostgREST/Plan/MutatePlan.hs
taimoorzaeem: Interesting. 

I went for a separate `MutatePlan` because I didn't see any negatives on this. Generalizing the current update method seems extra work and can likely cause regressions too.

TBH, I see that the `Update` and `Pgrst Patch` are quite similar and the extra code looks like it could be deduplicated. I am not sure either what's the best way to go here. But I would push for the current implementation because it's easier to review and maintain against the cost of some redundancy.

steve-chavez: >Then implement the pgrst patch mimetype with a basic "set" operation (which is the same as the default update, just via the custom mimetype structure)

This makes a lot of sense. I agree.

>  Generalizing the current update method seems extra work and can likely cause regressions too.

@taimoorzaeem Have about doing `set` on a different PR? It would not cause regressions as it would essentially be an "alias" for the current PATCH operation.

taimoorzaeem: > @taimoorzaeem Have about doing `set` on a different PR? It would not cause regressions as it would essentially be an "alias" for the current PATCH operation.

Yes, sure. I'll do in the current PR, we don't have much to lose and I too would like to focus on the `set` op first. We can do `incr` later in a separate PR.

taimoorzaeem:
If we want to do this without creating a separate `PgrstPatch` mutation, then we would have to break abstraction with `Update` mutation. Would it be cool to have `MutatePlan` like:

```haskell
  | Update
      { in_          :: QualifiedIdentifier
      , updCols      :: [CoercibleField]
      , updBody      :: Maybe LBS.ByteString
      , where_       :: [CoercibleLogicTree]
      , returning    :: [FieldName]
      , applyDefs    :: Bool
      , isPgrstPatch :: Bool -- a boolean for pgrst patch
      }
```

My plan is to convert this:

```json
[
  { "op": "set", "path": "id", "value": "10" },
  { "op": "set", "path": "name", "value": "double d" }
]
```

to this when `isPgrstPatch` is `True`:

```json
[
  {"id": "10", "name": "double d"}
]
```

After this, our normal update will take over to build the query. Let me know what you think.

steve-chavez:
@taimoorzaeem As per [above discussion](https://github.com/PostgREST/postgrest/pull/4227#discussion_r2251988499), we agreed that what we have now is a `set` operation under JSON Patch or PgrstPatch.

So instead maybe add a `data UpdateOperation = SetOperation`, just a single constructor for now and then a new field on `Update` that uses this new type.

src/PostgREST/ApiRequest/Payload.hs
```diff
@@ -37,13 +41,37 @@ import qualified PostgREST.MediaType as MediaType
...
+  | RawJSON      { payRaw   :: LBS.ByteString }
+  | RawPay       { payRaw   :: LBS.ByteString }
+  | PgrstPatchPay { payPgrstPatch :: [PgrstPatchOp] } -- ^ PgrstPatchPay is a list of patch updates
+
+-- This type should be in the Types.hs module, but because we
+-- are defining a JSON.FromJSON instance for this type, we need to have
+-- this type here to avoid "no-orphan-instances" warning.
+data PgrstPatchOp
+  = Set FieldName Text
+  -- We can add more operations in the future
```
taimoorzaeem: > So instead maybe add a data UpdateOperation = SetOperation, just a single constructor for now and then a new field on Update that uses this new type.

@steve-chavez We already have this in this implementation. This would make it easy to add new operators in the future (otherwise, having only a `set` operation would not be any more useful than the current PATCH method).

Another reason why I think that `PgrstPatch` and `Update` should be separate is that `PgrstPatch` needs to be parsed before building the plan to know which operator we are using (`set` in this case). We don't need that in `Update`.

steve-chavez: I see, fair enough. Please go ahead.

taimoorzaeem:
@steve-chavez @wolfgangwalther Could this get a review before we add the documentation?

src/PostgREST/ApiRequest/Types.hs
```diff
@@ -77,22 +76,6 @@ data Action
...
-      -- https://github.com/PostgREST/postgrest/pull/1005 for more details
-      , payKeys :: S.Set Text
-      -- ^ Keys of the object or if it's an array these keys are guaranteed to
-      -- be the same across all its objects
-      }
-  | ProcessedUrlEncoded { payArray  :: [(Text, Text)], payKeys :: S.Set Text }
-  | RawJSON { payRaw  :: LBS.ByteString }
-  | RawPay  { payRaw  :: LBS.ByteString }
-
-
```
wolfgangwalther, line 95: Please split refactors in separate commits, here moving the Payload type. Otherwise it's hard to review whether anything change in here.

src/PostgREST/ApiRequest/Payload.hs
```diff
@@ -37,13 +41,37 @@ import qualified PostgREST.MediaType as MediaType
...
+      -- large payloads the encoding had high memory usage, see
+      -- https://github.com/PostgREST/postgrest/pull/1005 for more details
+      , payKeys :: S.Set Text
+      -- ^ Keys of the object or if it's an array these keys are guaranteed to
+      -- be the same across all its objects
+      }
+  | ProcessedUrlEncoded { payArray :: [(Text, Text)], payKeys :: S.Set Text }
+  | RawJSON      { payRaw   :: LBS.ByteString }
+  | RawPay       { payRaw   :: LBS.ByteString }
+  | PgrstPatchPay { payPgrstPatch :: [PgrstPatchOp] } -- ^ PgrstPatchPay is a list of patch updates
```
wolfgangwalther: None of the other types have an indication of "payload" in their name.

Why not just:

```suggestion
  | PgrstPatch { payOps :: [PgrstPatchOp] } -- ^ PgrstPatch is a list of patch updates
```

Also I think the field name should be describing its content.

src/PostgREST/ApiRequest/Payload.hs
```diff
@@ -136,3 +171,22 @@ payloadAttributes raw json =
...
+
+
+instance JSON.FromJSON PgrstPatchOp where
+  parseJSON (JSON.Object o) = do
+    op    <- parseString o "op"
+    path  <- parseString o "path"
+    -- TODO: We need to decide what JSON "value"s are allowed in our
+    --       our Pgrst Patch implementation.
+    case op of
+      "set" -> Set path <$> parseString o "value"
```
wolfgangwalther: Do we only allow string values? I think the values should be arbitrary, i.e. also any other json type, right? They should behave the same way as a regular update/patch, e.g. for composite types etc.

src/PostgREST/Query/QueryBuilder.hs
```diff
@@ -161,13 +164,44 @@ mutatePlanToQuery (Update mainQi uCols body logicForest returnings applyDefaults
...
+-- DELETE: Corresponds to HTTP DELETE method
 mutatePlanToQuery (Delete mainQi logicForest returnings) =
   "DELETE FROM " <> fromQi mainQi <> " " <>
   whereLogic <> " " <>
   returningF mainQi returnings
   where
     whereLogic = if null logicForest then mempty else " WHERE " <> intercalateSnippet " AND " (pgFmtLogicTree mainQi <$> logicForest)
 
+-- PGRST PATCH: HTTP PATCH method with "vnd.pgrst.patch+json" Content-Type
+mutatePlanToQuery (PgrstPatch mainQi patchCols body logicForest returnings)
```
wolfgangwalther: I still don't like the idea of creating different queries for pgrst patch and regular update.

I think they should be treated roughly the same:
- They should both pass the full body in.
- The only difference is in how the values for the body are accessed for the respective operation.

For a PgrstPatch payload, you can still do things like `SET field = pgrst_body->0->'value'` or so. This should remove a lot of duplicate code and allow to implement operations purely as the part after `SET field =` and "doing something with `pgrst_body`".

test/spec/Feature/Query/PgrstPatchSpec.hs
wolfgangwalther, line 1: Does this feature also work with domain representations? It needs to and should be confirmed by a test.

src/PostgREST/Query/QueryBuilder.hs
taimoorzaeem: At this point, I agree. There are a lot of duplications which I noticed when writing the tests. Let me clean this up.

test/spec/Feature/Query/PlanSpec.hs
```diff
@@ -145,6 +145,22 @@ spec actualPgVersion = do
...
+
+      let totalCost  = planCost r
+          resHeaders = simpleHeaders r
+          resStatus  = simpleStatus r
+
+      liftIO $ do
+        resHeaders `shouldSatisfy` elem ("Content-Type", "application/vnd.pgrst.plan+json; for=\"application/json\"; charset=utf-8")
+        resHeaders `shouldSatisfy` notZeroContentLength
+        resStatus `shouldBe` Status { statusCode = 200, statusMessage="OK" }
+        totalCost `shouldBe` 14.21
```
taimoorzaeem, line 162: Hmm, `14.21` compared to a normal update `8.23`. :thinking:

Issue #4106, opened 2025-05-28
docs: catch invalid target references


steve-chavez:
## Problem

On https://github.com/PostgREST/postgrest/pull/4099#discussion_r2111130776, an invalid reference was not caught by `linkcheck`.

```
With `RPC <functions>`_, the preference is honored completely on the basis of the number of rows returned in the result set of the function. This can be useful for complex mutation queries using `data-modifying statements <https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-MODIFYING>`_.
```

Tthe "functions" reference is invalid here and when clicked on rendered docs it redirected to a "404 not found".

## Solution

Looks like we can use `nitpicky` for this:

> In nitpicky mode, Sphinx will warn about all references where the target cannot be found.
> https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-nitpicky-mode

## Notes

If the above doesn't work maybe https://github.com/rstcheck/rstcheck can help?

steve-chavez:
>  It was caught by linkcheck! But we only run linkcheck once a week, not on every PR. I fixed it after receiving the error notification.

Got it, cool. How did you get the error notification? I didn't get any 😿

steve-chavez:
> Maybe we could make the linkcheck post a comment to a designated issue. This way all of us could get the notifications.

That would be great 🚀

taimoorzaeem:
This could work? Maybe we can use this issue for "reporting" the link check failure.

```yaml
jobs:
  linkcheck:
     ...

 on-linkcheck-fail:
    runs-on: ubuntu-latest
    needs: linkcheck
    if: needs.linkcheck.result == 'failure'
    steps:
      - name: Notify on linkcheck failure by commenting
        uses: peter-evans/create-or-update-comment@v3
        with:
          issue-number: 4106
          body: |
            ❌ **Linkcheck Job Failed!**
            
            other details...        
```

steve-chavez:
> issue-number: 4106

@taimoorzaeem Looks good, so we would report the failures here right? That LGTM for starters.

taimoorzaeem:
> so we would report the failures here right? 

Yes, that's right. Looks like we are gonna have to test this by causing a deliberate linkcheck fail?

steve-chavez:
Right, does linkcheck work for commented out rst?

We would need to add something like `https://fake.postgrest.org/` hidden somewhere on `docs/index.rst` or any other page.

[...]

PR #4235, opened 2025-08-01
docs: catch invalid target references

taimoorzaeem:
The scheduled linkcheck job fails without notification. To catch the failure, a github workflow has been added which comments on issue #4106 so that the maintainers get notified.

### DIFF
--- a/.github/workflows/linkcheck.yaml
+++ b/.github/workflows/linkcheck.yaml
@@ -1,8 +1,14 @@
 name: Linkcheck
 
 on:
+  pull_request: # Temporarily test using PR, remove this before merge
+    branches:
+      - main
   schedule:
     - cron: '1 2 * * 3'
+
+permissions:
+  issues: write
 
 jobs:
   linkcheck:
@@ -15,3 +21,15 @@ jobs:
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         tools: docs.linkcheck.bin
     - run: postgrest-docs-linkcheck
+
+  on-linkcheck-fail:
+    runs-on: ubuntu-latest
+    needs: linkcheck
+    if: ${{ always() && needs.linkcheck.result == 'failure' }}
+    steps:
+    - name: Notify on linkcheck failure by commenting
+      uses: peter-evans/create-or-update-comment@v4
+      with:
+        issue-number: 4106
+        body: |
+          ❌ **Linkcheck Job Failed!**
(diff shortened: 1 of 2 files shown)


taimoorzaeem:
The [peter-evans/create-or-update-comment](https://github.com/peter-evans/create-or-update-comment?tab=readme-ov-file#action-inputs) uses github API for this. I think we need a `GITHUB_TOKEN` for this to work.

steve-chavez:
Hm, why does [Linkcheck / linkcheck (pull_request)](https://github.com/PostgREST/postgrest/actions/runs/16670580773/job/47185736333?pr=4235) fail? I thought we only ran it once a week and not per PR?

taimoorzaeem:
> Hm, why does [Linkcheck / linkcheck (pull_request)](https://github.com/PostgREST/postgrest/actions/runs/16670580773/job/47185736333?pr=4235) fail? I thought we only ran it once a week and not per PR?

Yes, you are right. I added a fake link to make it fail. I changed linkcheck temporarily to just run for this PR, to test the commenting workflow. Once this works, I'll remove the fake link before it could be merged.

.github/workflows/linkcheck.yaml
```diff
@@ -1,9 +1,15 @@
 name: Linkcheck
 
 on:
+  pull_request: # Temporarily test using PR, remove this before merge
+    branches:
+      - main
```
wolfgangwalther, line 6: With the `pull_request` event you won't have any permissions for the github token, thus you can't create comments. At least not when the PR comes from a fork.

We'd need to create this PR's branch in the postgrest repo itself to be able to test it that way.

.github/workflows/linkcheck.yaml
```diff
@@ -15,3 +21,15 @@ jobs:
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
         tools: docs.linkcheck.bin
     - run: postgrest-docs-linkcheck
+
+  on-linkcheck-fail:
+    runs-on: ubuntu-latest
+    needs: linkcheck
+    if: ${{ always() && needs.linkcheck.result == 'failure' }}
```
wolfgangwalther, line 28: We should not run a second job, but just a second step in the same job. Otherwise that first job will still fail causing another notification to whomever touched the scheduled workflow file last.

.github/workflows/linkcheck.yaml
taimoorzaeem: > We'd need to create this PR's branch in the postgrest repo itself to be able to test it that way.

If we are testing it that way, then I guess I should leave this for maintainers with write access on this repo.

taimoorzaeem:
Closing as stale and outdated (and in favor of https://github.com/PostgREST/postgrest/pull/4235#discussion_r2251771587).

Issue #4226, opened 2025-07-28
`log-query=main-query` doesn't log queries for OpenAPI


steve-chavez:
## Problem

Doing:

```
$ PGRST_LOG_QUERY=main-query PGRST_LOG_LEVEL=debug PGRST_JWT_CACHE_MAX_ENTRIES=0 postgrest-with-postgresql-16 -f test/spec/fixtures/load.sql postgrest-run

$ curl localhost:3000/ -I
HTTP/1.1 200 OK
```

Doesn't log any of the generated queries:

```
127.0.0.1 - postgrest_test_anonymous [28/Jul/2025:08:01:05 -0500] "HEAD / HTTP/1.1" 200 - "" "curl/7.81.0"
...
```

But in fact we generate 3 queries for the default `openapi-mode = "follow-privileges"`, see 3 `SQL.statement` here:

https://github.com/PostgREST/postgrest/blob/aa58e37f038b06ed7d3d3a98205085b3b319cd38/src/PostgREST/Query.hs#L184-L188

## Solution

Join the 3 queries into one and use hasql-dynamic-statements to log the query the same as the others. Right now it's just a `mempty`:

https://github.com/PostgREST/postgrest/blob/aa58e37f038b06ed7d3d3a98205085b3b319cd38/src/PostgREST/Query.hs#L179

## Notes

- `openapi-mode = "follow-privileges"` also runs a query: 

https://github.com/PostgREST/postgrest/blob/aa58e37f038b06ed7d3d3a98205085b3b319cd38/src/PostgREST/Query.hs#L189-L193

But maybe this can be omitted if we put the schema description into the SchemaCache, it looks more fit and less wasteful there.

- The reason we do queries on `follow-privileges` is to be able to filter to which db objects the role has access to, but perhaps we should put the privileges into the SchemaCache to avoid queries. This is much more involved so it'd need to be done later.

steve-chavez:
We can make OpenAPI requests completely devoid of SQL queries, we just need to cache the `accessibleTables` and `accessibleFuncs` here:

https://github.com/PostgREST/postgrest/blob/aa58e37f038b06ed7d3d3a98205085b3b319cd38/src/PostgREST/Query.hs#L184-L187

This would solve this issue but it would cause a bit of a breaking change. Since we would cache privileges, the event trigger:

https://docs.postgrest.org/en/v13/references/schema_cache.html#finer-grained-event-trigger

Would have to be changed to include `GRANT` too. 

[...]

PR #4237, opened 2025-08-03
perf: cache schema description for OpenAPI

steve-chavez:
Previously we use to query the schema descriptions on each request for OpenAPI.

Partly addresses https://github.com/PostgREST/postgrest/issues/4226

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   + The exposed schemas are now listed in the `hint` instead of the `message` field.
 - Improve error details of `PGRST301` error by @taimoorzaeem in #4051
 - Bounded JWT cache using the SIEVE algorithm by @mkleczek in #4084
+- Cache schema descriptions to avoid a query each time OpenAPI is requested by @steve-chavez in #4226
 
 ### Fixed
 
--- a/src/PostgREST/Query.hs
+++ b/src/PostgREST/Query.hs
@@ -185,12 +185,13 @@ actionQuery (MaybeDb plan@InspectPlan{ipSchema=tSchema}) AppConfig{..} _ sCache 
           MaybeDbResult plan . Just <$> ((,,)
                 (HM.filterWithKey (\qi _ -> S.member qi tableAccess) $ SchemaCache.dbTables sCache)
             <$> SQL.statement ([tSchema], configDbHoistedTxSettings) (SchemaCache.accessibleFuncs configDbPreparedStatements)
-            <*> SQL.statement tSchema (SchemaCache.schemaDescription configDbPreparedStatements))
+            <*> pure (join $ HM.lookup tSchema $ SchemaCache.dbSchemaDescriptions sCache))
         OAIgnorePriv ->
-          (MaybeDbResult plan . Just) . (,,)
-                (HM.filterWithKey (\(QualifiedIdentifier sch _) _ ->  sch == tSchema) $ SchemaCache.dbTables sCache)
-                (HM.filterWithKey (\(QualifiedIdentifier sch _) _ ->  sch == tSchema) $ SchemaCache.dbRoutines sCache)
-            <$> SQL.statement tSchema (SchemaCache.schemaDescription configDbPreparedStatements)
+          pure $ MaybeDbResult plan $ Just (
+                 HM.filterWithKey (\(QualifiedIdentifier sch _) _ ->  sch == tSchema) $ SchemaCache.dbTables sCache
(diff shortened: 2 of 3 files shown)


wolfgangwalther:
With #1698 in mind, we should not do this. Instead, we should try to minimize the contents of the schema cache to everything that is required for *regular* requests and query everything that is only required for OpenAPI output on demand.

steve-chavez:
@wolfgangwalther This is needed for openapi right now (showing the schema comment), additionally it's great for logs too since now we should the total schemas:

```
03/Aug/2025:09:52:34 -0500: Schema cache loaded 271 Relations, 232 Relationships, 148 Functions, 15 Domain Representations, 45 Media Type Handlers, 1196 Timezones, 
14 Schema Descriptions
```

Which helps a lot in debugging.

> Instead, we should try to minimize the contents of the schema cache to everything that is required for regular requests and query everything that is only required for OpenAPI output on demand.

Yeah, but that's a major change. This is much more gradual, how about merging this for unblocking https://github.com/PostgREST/postgrest/issues/4226 and https://github.com/PostgREST/postgrest/issues/3934?

steve-chavez:
Additionally, it was always bad to have a query executing whenever we requested OpenAPI for `ignore-privileges`.. there isn't a need there really.

wolfgangwalther:
> Yeah, but that's a major change. This is much more gradual

Not sure I understand that logic. Yeah, we still have a long way to go towards that goal, but the "gradual" step you are proposing.. is a step *backwards*, away from that goal.

> Additionally, it was always bad to have a query executing whenever we requested OpenAPI for `ignore-privileges`.. there isn't a need there really.

With #1698 there will always be such a query anyway, so I don't think it's worth it to optimize for that.

> it's great for logs too since now we should the total schemas:

IIRC we are not hitting the database for unknown tables anymore, correct? If that's the case, then we already *have* this information in the schema cache, because otherwise we can't tell whether a request to a certain schema/table will be run or not.

steve-chavez:
> With https://github.com/PostgREST/postgrest/issues/1698 there will always be such a query anyway, so I don't think it's worth it to optimize for that.
https://github.com/PostgREST/postgrest/blob/fa09e4af0d74e847b8e1a446825c262320f281a9/src/PostgREST/Query.hs#L183-L193

steve-chavez:
> With https://github.com/PostgREST/postgrest/issues/1698 there will always be such a query anyway, so I don't think it's worth it to optimize for that.

Ok so given the above, the OpenAPI endpoint will always execute a query. So it's fruitless to have this mess were we use some parts of the schema cache + queries for OpenAPI.

https://github.com/PostgREST/postgrest/blob/fa09e4af0d74e847b8e1a446825c262320f281a9/src/PostgREST/Query.hs#L183-L193

I just need some uniformity to be able to do https://github.com/PostgREST/postgrest/issues/4226 and https://github.com/PostgREST/postgrest/issues/3934. So I'll refactor those parts to use a single query instead. That should be uncontroversial I believe.

wolfgangwalther:
> So I'll refactor those parts to use a single query instead. That should be uncontroversial I believe.

Yes, I believe so, too :)

steve-chavez:
> With https://github.com/PostgREST/postgrest/issues/1698 in mind, we should not do this. Instead, we should try to minimize the contents of the schema cache to everything that is required for regular requests and query everything that is only required for OpenAPI output on demand

If #1698 is the goal, then we should instead not use the SchemaCache for the OpenAPI requests. That's doable for Tables and Functions (queries are simple enough) but the problem here is that OpenAPI uses Relationships (queries are complex + requires custom Haskell logic) very poorly (because it only works for single column FKs), that is something I've wanted to break before.. if that would be done it would be doable to remove the SchemaCache dependency and then integrate postgrest-openapi gradually..

steve-chavez:
> if that would be done it would be doable to remove the SchemaCache dependency and then integrate postgrest-openapi gradually..

In the end I guess it's not possible to integrate gradually cleanly, it will need a big breaking change = once `postgrest-openapi` is integrated, there will be no fallback to builtin OpenAPI.

steve-chavez:
I'm finding hard to progress without this improvement here.

> Not sure I understand that logic. Yeah, we still have a long way to go towards that goal, but the "gradual" step you are proposing.. is a step backwards, away from that goal.

I don't think this is a step backwards, the agreement for openapi is still there. The immediate goal is https://github.com/PostgREST/postgrest/issues/4226. 

I don't see at all how is this PR nothing but an improvement in the current state of things, the schema description doesn't need to be queried every time.. it's a performance improvement.

We can later remove all of this, because as mentioned above https://github.com/PostgREST/postgrest/pull/4237#issuecomment-3148573218 we'll have to get rid of the whole openapi logic to integrate postgrest-openapi. But in the meantime, we can keep improving the project.

steve-chavez:
I'll open another PR to solve #4226 in one go so the point is clearer.

Issue #4245, opened 2025-08-05
geojson failure always runs query


steve-chavez:
## Problem

Having:

```bash
PGRST_LOG_LEVEL=info PGRST_LOG_QUERY=main-query PGRST_LOG_LEVEL=debug \
postgrest-with-postgresql-17 -f test/spec/fixtures/load.sql postgrest-run
```

And requesting:

```
curl localhost:3000/shops "Accept: application/geo+json"
```

Will result in an error:

```json
{"code":"42883","details":null,"hint":"No function matches the given name and argument types. You might need to add explicit type casts.","message":"function st_asgeojson(record) does not 
exist"}
```

And it will always run the query (this will show in logs):

```sql
05/Aug/2025:12:54:19 -0500: WITH pgrst_source AS ( SELECT "test"."shops".* FROM "test"."shops"    )  SELECT null::bigint AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total, 
json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]')) AS body, nullif(current_setting('response.headers', true), '') AS res
ponse_headers, nullif(current_setting('response.status', true), '') AS response_status, '' AS response_inserted FROM ( SELECT * FROM pgrst_source ) _postgrest_t
127.0.0.1 - postgrest_test_anonymous [05/Aug/2025:12:54:19 -0500] "GET /shops HTTP/1.1" 404 195 "" "curl/7.81.0"
```

## Solution

Store the PostGIS availability somehow in the Schema Cache, and fail without hitting the db.

steve-chavez:
Btw, the above request works when adding `PGRST_DB_EXTRA_SEARCH_PATH="extensions"` to the `postgrest-run` command.

steve-chavez:
This should also help with the integration of `postgrest-openapi` and `pg_csv`.

steve-chavez:
I'm thinking the query should be:

- Restricted to the extensions in the schemas conformed by `db-schemas` + `db-extra-search-path`
- Restricted to only finding the exact function we need. In this case `st_asgeojson(record)`.
  + This will allow ignoring any logic related to versions of an extension. Keeping things simpler.

steve-chavez:
Given that a function in postgres is identified by its `schema + name + parameters`, we only need to consider those attributes for the extension function schema cache. 

[...]

Issue #3487, opened 2024-05-06
Support structured data in PostgreSQL `COMMENT`S


salim-b:
## Problem

PostgREST allows to specify additional metadata to document the offered RESTful API [via PostgreSQL `COMMENT`s](https://postgrest.org/en/v12/references/api/openapi.html#open-api). Other tools also leverage PostgreSQL `COMMENT`s to specify additional metadata or configuration, e.g.

- pg_graphql via [`@graphql(<JSON>)` "comment directives"](https://supabase.github.io/pg_graphql/configuration/#comment-directives)
- PostGraphile via [`@tag object` "smart tags"](https://postgraphile.org/postgraphile/current/smart-tags)
- schemats/pg-to-ts via [`@type {type}` annotations](https://github.com/danvk/pg-to-ts?tab=readme-ov-file#json-types) which conform to [JSDoc tags](https://jsdoc.app/tags-type)

If one decides to use one of these tools in conjunction with PostgREST, conflicts arise, i.e. PostgREST won't ignore the above tags but include them in the generated OpenAPI description. The underlying problem could be summarized as ***the lack of a common standard to include structured data in PostgreSQL `COMMENT`s***[^s]. 

## Solution

In another discussion @wolfgangwalther [suggested to](https://github.com/PostgREST/postgrest/issues/2028#issuecomment-973832177)

> use a yaml front matter block to add metadata, while keeping the current openapi comments support backwards compatible

YAML headers seem powerful (e.g. multiple tools could read the same keys) and pragmatic at the same time (not too hard to parse and can be implemented in a backwards-compatible way), so other tools would hopefully add support for it, too. The author of pg_graphql [already showed approval and would implement it](https://github.com/supabase/pg_graphql/issues/331#issuecomment-1448567699). YAML headers would also allow to "hide" arbitrary data from PostgREST like the above @-tags by simply defining them inside the YAML header (under a key that has no meaning for PostgREST). In the end, the possible use cases go far beyond PostgREST – but PostgREST could lead the way!

[^s]: A web search did not reveal any kind exisiting initiative to establish such a standard. Please correct me, if I missed something!

## Preliminary discussion

- https://github.com/supabase/pg_graphql/issues/331
- https://github.com/PostgREST/postgrest/issues/2028#issuecomment-973832177

## Further notes

[...]

PR #4246, opened 2025-08-05
fix: geojson invalid query on unavailable PostGIS

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/4245.

It adds a query on schema cache construction, but doesn't add a new attribute to the SchemaCache type. We only need the information to be passed to the `initialMediaHandlers` function.


src/PostgREST/SchemaCache.hs
```diff
@@ -1139,6 +1142,35 @@ timezones = SQL.Statement sql HE.noParams decodeTimezones
...
+        JOIN pg_catalog.pg_extension AS e
+          ON e.oid   = d.refobjid
+        WHERE p.pronamespace   = ANY($$1::regnamespace[] || $$2::regnamespace[])
+          AND p.proname        = 'st_asgeojson'
+          AND e.extname        = 'postgis'
+          AND p.proargtypes[0] = 'record'::regtype
+          AND pg_get_function_result(p.oid) = 'text'
+      );
+    |]
+
```
steve-chavez, line 1173: Simplest query that does the job for now, later we can augment this same query to detect other functions like `pg_csv` aggs or `postgrest-openapi`.

test/io/__snapshots__/test_cli/test_schema_cache_snapshot[dbMediaHandlers].yaml
```diff
@@ -3,11 +3,6 @@
   - - tag: BuiltinOvAggCsv
     - tag: MTTextCSV
 
-- - - tag: RelAnyElement
-    - tag: MTGeoJSON
-  - - tag: BuiltinOvAggGeoJson
-    - tag: MTGeoJSON
-
```
Comment (steve-chavez) [inline test/io/__snapshots__/test_cli/test_schema_cache_snapshot[dbMediaHandlers].yaml:10]: The `test/io/fixures.sql` don't have PostGIS enabled, hence why these lines are deleted. I think these dbMediaHandlers make sense as a default.

`test/spec/Feature/Query/PostGISSpec.hs` already proves that PostGIS works as always.

src/PostgREST/SchemaCache.hs
```diff
@@ -1139,6 +1142,35 @@ timezones = SQL.Statement sql HE.noParams decodeTimezones
     decodeTimezones :: HD.Result TimezoneNames
     decodeTimezones = S.fromList <$> HD.rowList (column HD.text)
 
+
+-- Find the postgis function that has the signature:
+-- st_asgeojson(record,...) returns text
```
steve-chavez, line 1147: https://postgis.net/docs/ST_AsGeoJSON.html

We only need to know that the first parameter is a `record`, the rest of the parameters don't matter as we don't use them.

`returns text` is not strictly needed, but I did it to be safe.

Btw it's weird that this `returns text`. Maybe on a later version they could change it to `returns json` and this would break? We could omit the return type here to avoid that.

steve-chavez:
Looking good to merge.

src/PostgREST/SchemaCache.hs
steve-chavez: We have postgis coming from nixpkgs, that should warn us if a breaking change happens. So this should be fine.

steve-chavez:
> (ah.. why am I even writing this.. it's merged already anyway...)

My bad, will wait longer next time. Should we enforce approvals for PRs?

> PostGIS should not be in initialMediaHandlers. Instead it should be in mediaHandlers, 

Will see how to fix this. I was expecting to redo this once I integrate `pg_csv` (working on this now).

steve-chavez:
> My bad, will wait longer next time. Should we enforce approvals for PRs?

Note: this is now in place.

steve-chavez:
> which looks up some specially hardcoded functions from extensions and then returns the data in the same format as the other media handlers.

Turns out that won't work, `mediaHandlers` has this output:

```
 handler_schema |         handler_name         | target_schema |       target_name        |          media_type          |     resolved_media_type      
----------------+------------------------------+---------------+--------------------------+------------------------------+------------------------------
 test           | */*                          | test          | ret_any_mt               | */*                          | application/octet-stream
 test           | */*                          | test          | ret_some_mt              | */*                          | application/octet-stream
 test           | */*                          | test          | some_final               | */*                          | application/octet-stream
 test           | */*                          | test          | some_trans               | */*                          | application/octet-stream
 test           | application/geo+json         | test          | lines_geojson_final      | application/geo+json         | application/geo+json
 test           | application/geo+json         | test          | lines_geojson_trans      | application/geo+json         | application/geo+json
 test           | application/json             | test          | ov_json_trans            | application/json             | application/json
 test           | application/octet-stream     | test          | overloaded_unnamed_param | application/octet-stream     | application/octet-stream
 test           | application/octet-stream     | test          | unnamed_bytea_param      | application/octet-stream     | application/octet-stream
 test           | application/openapi+json     | test          | root                     | application/openapi+json     | application/openapi+json
 test           | application/vnd.geo2+json    | test          | geo2json_final           | application/vnd.geo2+json    | application/vnd.geo2+json
 test           | application/vnd.geo2+json    | test          | geo2json_trans           | application/vnd.geo2+json    | application/vnd.geo2+json
 test           | application/vnd.pgrst.object | test          | pgrst_obj_json_trans     | application/vnd.pgrst.object | application/vnd.pgrst.object
...
```

i.e. it expects a single aggregate function, not an expression like we have for our builtins, e.g. geojson:

https://github.com/PostgREST/postgrest/blob/0f1ca8faac630518b343f195464beff47fe32c78/src/PostgREST/Query/SqlFragment.hs#L223-L224

Not sure if it's possible to retrofit the geojson query into the media handlers abstraction.

steve-chavez:
> Not sure if it's possible to retrofit the geojson query into the media handlers abstraction.

For now what occurs to me is to add two columns to the media handlers result:

- A `bool is_single_agg`, for the geojson builtin this value would be false.
- A `text expr`, for the geojson builtin this will be the same as our `asGeoJsonF` query
  + This would mean `_postgrest_t` is the hardcoded input that PostgREST expects for transforming the query result
  + Maybe later we can reuse this `_postgrest_t` convention to allow custom media handlers to use expressions on top of aggregates somehow

steve-chavez:
@wolfgangwalther Any thoughts on the above? Want to make sure we agree before I go and implement the idea.

wolfgangwalther:
> i.e. it expects a single aggregate function, not an expression like we have for our builtins, e.g. geojson:

I wonder, whether, at this stage, we're not better off to just rip out in-core PostGIS support and provide docs on how to create a custom aggregate for it instead?

It might not be possible just yet to create such an aggregate, but it would be good to think about what changes we would need to support it.

wolfgangwalther:
> Turns out that won't work, `mediaHandlers` has this output:

I'm confused by this output: Why do we show `..._final` and `..._trans` functions there for each mediatype? This seems wrong, we should not be concerned with the implementation details of these, right? We're calling the aggregate function after all, right?

steve-chavez:
> I wonder, whether, at this stage, we're not better off to just rip out in-core PostGIS support and provide docs on how to create a custom aggregate for it instead?

In light of `pg_csv` and other extensions that we might release, that approach would offer a lesser DX. For me the direction is: "user does CREATE EXTENSION in our detectable schemas" -> "PostgREST has new media type handler". Forcing users to add a migration with a new function or aggregate is inconvenient when we already know how to integrate with an extension. We already provide the means to override these autodetections and the PostGIS query is just one line so there's no big maintenance cost.

So I don't think we should remove PostGIS.

> This seems wrong, we should not be concerned with the implementation details of these, right? 

I don't recall the specifics but maybe they were needed. Will revisit.

wolfgangwalther:
> In light of `pg_csv` and other extensions that we might release, that approach would offer a lesser DX. For me the direction is: "user does CREATE EXTENSION in our detectable schemas" -> "PostgREST has new media type handler". Forcing users to add a migration with a new function or aggregate is inconvenient when we already know how to integrate with an extension. We already provide the means to override these autodetections and the PostGIS query is just one line so there's no big maintenance cost.

If that's your goal, then PostGIS should expose a "proper" aggregate function, that we only need to wrap in a `CREATE DOMAIN` (or auto-detect). Maybe upstreaming such a function would be a better long-term solution?

Having code in PostgREST to deal with specific extensions just doesn't scale.

steve-chavez:
> If that's your goal, then PostGIS should expose a "proper" aggregate function, that we only need to wrap in a CREATE DOMAIN (or auto-detect). Maybe upstreaming such a function would be a better long-term solution?

Dunno if that is feasible since there are some customizations like the one in https://github.com/PostgREST/postgrest/issues/2699#issue-1616621281:

> json_build_object( 'type',      'name',  'properties', json_build_object( 'name', CONCAT('urn:ogc:def:crs:EPSG::',Find_SRID('SCHEMA', 'TABLE_NAME', 'GEOMETRY_COLUMN'))  ) ) as crs

Looks hard to upstream convenience aggregates (and mostly for us) to a mature/established project like PostGIS.

Now convenience aggregates might fit into `postgrest-contrib`, this one might be doable since `json_agg_transfn` is exposed and we could add it a wrapper final function. But that still sounds complicated for the user/us.

Thinking of a more generic way to call a function (not just aggregates) for media type handlers seems more useful.
Even just a [pre_config](https://docs.postgrest.org/en/v13/references/configuration.html#in-database-configuration) with:

```sql
set_config('pgrst.media_handlers', $$
  {application/geo+json,json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]'))}
 $$);
```

Doesn't look that bad IMO.

> Having code in PostgREST to deal with specific extensions just doesn't scale.

Wait, I don't see that many extensions with custom aggregates to be worried about scale right now. I don't think that's a solid argument.

steve-chavez:
> set_config('pgrst.media_handlers', $$
  {application/geo+json,json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]'))}
 $$);

This one actually works: 

```sql
select ARRAY[
  'application/geo+json',
  $$json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]')$$
]::text[];
```

How about that for more generic media handlers?

**Edit**: final type should be a bidimensional array `text[][]`, this works fine on pg:

```sql
select array[array['a','b'],array['c','d']]::text[][];
     array     
---------------
 {{a,b},{c,d}}
(1 row)
```

steve-chavez:
I'm kinda liking this interface more than the current one restricted to aggs, we can even add a 2nd element to add a target(`anyelement` or a particular table):
 
```sql
select ARRAY[
  'application/geo+json',
  'anyelement',
  $$json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]')$$
]::text[];
```

The domain representations abstractions was retrofitted into media type handlers, but it already proved to be insufficient since in postgres we can't compose functions over an aggregate and still be considered an aggregate.

**Edit**: with this we wouldn't need two handlers too ([one for function, one for table](https://docs.postgrest.org/en/v13/references/api/media_type_handlers.html#handler-function)).

wolfgangwalther:
I don't think this should be set via configuration at all. This is completely against the spirit of PostgREST, which is to "expose a database". Now you're starting to put actual SQL code into the configuration, so outside the database - that's exactly **not** how it should be.

steve-chavez:
> This is completely against the spirit of PostgREST, which is to "expose a database". Now you're starting to put actual SQL code into the configuration, so outside the database - that's exactly not how it should be.

I'm proposing to add that on the [pre_config](https://docs.postgrest.org/en/v13/references/configuration.html#in-database-configuration) -- this is inside the database. As:

```sql
returns void as $$
select ARRAY[
  'application/geo+json',
  'anyelement',
  $$json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]')$$
]::text[];
$$ language sql;
```

This is in the same vein as your proposal about enriching the database through comments
https://github.com/PostgREST/postgrest/issues/3487#issuecomment-2096111751.

So I don't see how that's "outside of the database".

Now if you mean we shouldn't enrich the database in any other form than with native database objects (functions, tables, casts, etc), that I can agree. But that would also slash those COMMENT proposals, which I would be fine with.

steve-chavez:
Reverted on https://github.com/PostgREST/postgrest/pull/4266. I'm fine with this as there's no clear direction on how to detect the PostGIS query for now. I was mostly trying to clean a bit before integrating `pg_csv` (which is now clear it can be done by enriching the `mediaHandlers` query).

wolfgangwalther:
> I'm proposing to add that on the [pre_config](https://docs.postgrest.org/en/v13/references/configuration.html#in-database-configuration) -- this is inside the database. As:

So the example you gave is incomplete - the function returns void, but has a pure `SELECT`without side effects in it. I assume you are missing something like a `set_config('pgrst.xxx', ARRAY[...])` or so. This means you are proposing to add a config option to put additional media handlers in - which can contain SQL code.

Config options *can* be set in the database, but they can also be set in the config file, via env variables etc.

SQL code should not be possible to set via these channels. A "config option" is the wrong tool for the job. If you need to keep your config in-sync that closely with the evolution of the database schema, something is wrong.

> This is in the same vein as your proposal about enriching the database through comments [#3487 (comment)](https://github.com/PostgREST/postgrest/issues/3487#issuecomment-2096111751).

Quite the contrary:
- The comments can't be created via configuration - that's exactly what I was opposing back then as well.
- The comments are per definition *in* the database. They are database objects.
- The comments were never supposed to carry arbitrary SQL code.

So there is really no similarity between the two proposals. The only shared pattern is: I dislike configuration.

steve-chavez:
> SQL code should not be possible to set via these channels. A "config option" is the wrong tool for the job. If you need to keep your config in-sync that closely with the evolution of the database schema, something is wrong.

So the underlying problem with SQL code as config is that it will be a maintenance problem since it's not tied to schema evolution right? (e.g. no type check errors when the function doesn't exist or it doesn't match the signature anymore)

That gives me an idea, if we could express `json_build_object('type', 'FeatureCollection', 'features', coalesce(json_agg(ST_AsGeoJSON(_postgrest_t)::json), '[]')` as function composition (using Haskell syntax):

```haskell
json_build_object 'type'  'FeatureCollection' 'features' . flip coalesce '[]' .  json_agg . ::json . st_asgeojson
```

Then in theory we could store this as an array of `regproc`, and then this would be tied to schema evolution -- so no maintenance problem anymore?

This doesn't fully work yet, but could be interesting to explore as an extension.

Issue #3203, opened 2024-02-02
`server-host=*` doesn't bind to ipv6


steve-chavez:
## Problem

[We document](https://postgrest.org/en/stable/references/configuration.html#server-host) that `*` should bind to ipv4 and ipv6. However ipv6 doesn't work, only ipv4:

```bash
$ PGRST_SERVER_HOST='*' postgrest-with-postgresql-16 postgrest-run
```

```bash
$ curl localhost:3000/projects  --ipv4 -I
HTTP/1.1 200 OK

$ curl localhost:3000/projects  --ipv6 -I
curl: (7) Failed to connect to localhost port 3000 after 0 ms: Connection refused
```

Issue #3202, opened 2024-02-02
`server-host=!6` also binds to ipv4 


steve-chavez:
## Problem

[We document](https://postgrest.org/en/stable/references/configuration.html#server-host) that `!6` should only bind to ipv6, however a request using ipv4 succeeds

```bash
$ PGRST_SERVER_HOST='!6' postgrest-with-postgresql-16 postgrest-run
```

```bash
$ curl localhost:3000/projects  --ipv4 -I
HTTP/1.1 200 OK
```

## Notes

Using `!4` works as expected:

```bash
PGRST_SERVER_HOST='!4' postgrest-with-postgresql-15 postgrest-run
```

```bash
$ curl localhost:3000/projects  --ipv6 -I
curl: (7) Failed to connect to localhost port 3000 after 0 ms: Connection refused
```

steve-chavez:
@wolfgangwalther Are you able to reproduce? Not sure if this could be an OS specific.

> Did you find out whether this and https://github.com/PostgREST/postgrest/issues/3203 are actually bugs on our side or in the lib we are using?

It must be the lib, we only do this:
https://github.com/PostgREST/postgrest/blob/71887e4b78aca74adcdabfc61c2901e7e479878f/src/PostgREST/App.hs#L95

- https://hackage.haskell.org/package/warp-3.4.0/docs/Network-Wai-Handler-Warp.html#v:setHost
- https://hackage.haskell.org/package/warp-3.4.0/docs/Network-Wai-Handler-Warp.html#t:HostPreference

laurenceisla:
> So if you can reproduce this with plain warp, then it would probably be better to raise this upstream.

I'm testing the library right now to confirm this. The  "funny" thing is that `*` does not seem to accept `ipv6` requests.

```bash
$ PGRST_SERVER_HOST='*' postgrest-with-postgresql-16 postgrest-run
```

```bash
$ curl localhost:3000/projects  --ipv6 -I
curl: (7) Failed to connect to localhost port 3000 after 0 ms: Couldn't connect to server
```

steve-chavez:
@laurenceisla That looks the same as https://github.com/PostgREST/postgrest/issues/3203

laurenceisla:
@steve-chavez True, didn't check that.

laurenceisla:
Made a gist with the tests I did:

https://gist.github.com/laurenceisla/61af763fb028437f24e5a869225eb1f8

The behavior is the same with Warp only.

[...]

PR #4269, opened 2025-08-17
add: `--ready` flag for postgrest healthcheck

### LINKED ISSUE #4239: New CLI option `--ready`
## Problem

I would like to do a healthcheck in Docker Compose on the PostgREST service. The default image doesn't have anything but the [PostgREST binary](https://github.com/PostgREST/postgrest/tree/main/nix/tools/docker) in it. So a healthcheck using the `/ready` endpoint (as described [here](https://docs.postgrest.org/en/v13/references/admin_server.html#ready)) would require you to expand the image to include `curl`, bloating the image, as you probably need to base it on a distro.

## Solution

A new CLI flag called `--ready` that hooks into the running process and performs the same functionality as the `/ready` endpoint, exiting with status code `0` when ready or `1` otherwise.

Additionally, a similar flag `--live` could be added to duplicate the `/live` endpoint.

taimoorzaeem:
The `--ready` flag is a wrapper around the admin server `/ready` request. This is done through using an http client library in postgrest.

- [x] Implementation
- [x] Refactoring
- [x] Tests
- [x] Docs

Closes #4239.

### DIFF
--- a/docs/explanations/install.rst
+++ b/docs/explanations/install.rst
@@ -142,6 +142,7 @@ To avoid having to install the database at all, you can run both it and the serv
       ports:
         - "3000:3000"
       environment:
+        PGRST_SERVER_HOST: localhost # necessary for `postgrest --ready` flag to work
         PGRST_DB_URI: postgres://app_user:password@db:5432/app_db
         PGRST_OPENAPI_SERVER_PROXY_URI: http://127.0.0.1:3000
       depends_on:
--- a/docs/references/cli.rst
+++ b/docs/references/cli.rst
@@ -7,7 +7,7 @@ PostgREST provides a CLI with the options listed below:
 
 .. code:: text
 
-  Usage: postgrest [-v|--version] [-e|--example] [--dump-config | --dump-schema] 
+  Usage: postgrest [-v|--version] [-e|--example] [--dump-config | --dump-schema | --ready]
                  [FILENAME]
 
     PostgREST / create a REST API to an existing Postgres
@@ -20,6 +20,8 @@ PostgREST provides a CLI with the options listed below:
     --dump-config            Dump loaded configuration and exit
     --dump-schema            Dump loaded schema as JSON and exit (for debugging,
                              output structure is unstable)
+    --ready                  Checks the health of PostgREST by doing a request on
+                             the admin server /ready endpoint
     FILENAME                 Path to configuration file
 
 FILENAME
@@ -71,3 +73,17 @@ Dump Schema
   $ postgrest --dump-schema
 
 Dumps the schema cache in JSON format.
+
+Ready Flag
+----------
+
+Makes a request to the ``/ready`` endpoint of the :ref:`admin_server`. It exits with a return code of ``0`` on success and ``1`` on failure.
+
+.. code-block:: bash
+
+  $ postgrest --ready
+  OK: http://localhost:3001/ready
+
+.. note::
+
+  The ``--ready`` flag cannot be used when :ref:`server-host` is configured with special hostnames. We suggest to change it to ``localhost``.
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -52,6 +52,7 @@ library
                       PostgREST.Auth.Types
                       PostgREST.Cache.Sieve
                       PostgREST.CLI
+                      PostgREST.Client
                       PostgREST.Config
                       PostgREST.Config.Database
                       PostgREST.Config.JSPath
@@ -118,6 +119,7 @@ library
                     , hasql-pool                >= 1.0.1 && < 1.1
                     , hasql-transaction         >= 1.0.1 && < 1.2
                     , heredoc                   >= 0.2 && < 0.3
+                    , http-client               >= 0.7.19 && < 0.8
                     , http-types                >= 0.12.2 && < 0.13
                     , insert-ordered-containers >= 0.2.2 && < 0.3
                     , iproute                   >= 1.7.0 && < 1.8
--- a/src/PostgREST/CLI.hs
+++ b/src/PostgREST/CLI.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase      #-}
 {-# LANGUAGE NamedFieldPuns  #-}
 {-# LANGUAGE QuasiQuotes     #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -24,6 +25,7 @@ import PostgREST.Version     (prettyVersion)
 
 import qualified PostgREST.App      as App
 import qualified PostgREST.AppState as AppState
+import qualified PostgREST.Client   as Client
 import qualified PostgREST.Config   as Config
 
 import Protolude
@@ -31,16 +33,26 @@ import Protolude
 
 main :: CLI -> IO ()
 main CLI{cliCommand, cliPath} = do
-  conf@AppConfig{..} <-
+  conf <-
     either panic identity <$> Config.readAppConfig mempty cliPath Nothing mempty mempty
+  case cliCommand of
+    Client adminCmd -> runClientCommand conf adminCmd
+    Run runCmd      -> runAppCommand conf runCmd
 
+-- | Run command using http-client to communicate with an already running postgrest
+runClientCommand :: AppConfig -> ClientCommand -> IO ()
+runClientCommand conf CmdReady = Client.ready conf
+
+-- | Run postgrest with command
+runAppCommand :: AppConfig -> RunCommand -> IO ()
+runAppCommand conf@AppConfig{..} runCmd = do
   -- Per https://github.com/PostgREST/postgrest/issues/268, we want to
   -- explicitly close the connections to PostgreSQL on shutdown.
   -- 'AppState.destroy' takes care of that.
   bracket
     (AppState.init conf)
     AppState.destroy
-    (\appState -> case cliCommand of
+    (\appState -> case runCmd of
       CmdDumpConfig -> do
         when configDbConfig $ AppState.readInDbConfig True appState
         putStr . Config.toText =<< AppState.getConfig appState
@@ -71,6 +83,13 @@ data CLI = CLI
   }
 
 data Command
+  = Client ClientCommand
+  | Run RunCommand
+
+data ClientCommand
+  = CmdReady
+
+data RunCommand
   = CmdRun
   | CmdDumpConfig
   | CmdDumpSchema
@@ -105,7 +124,7 @@ readCLIShowHelp =
     cliParser :: O.Parser CLI
     cliParser =
       CLI
-        <$> (dumpConfigFlag <|> dumpSchemaFlag)
+        <$> (dumpConfigFlag <|> dumpSchemaFlag <|> readyFlag)
         <*> O.optional configFileOption
 
     configFileOption =
@@ -114,14 +133,20 @@ readCLIShowHelp =
         <> O.help "Path to configuration file"
 
     dumpConfigFlag =
-      O.flag CmdRun CmdDumpConfig $
+      O.flag (Run CmdRun) (Run CmdDumpConfig) $
         O.long "dump-config"
         <> O.help "Dump loaded configuration and exit"
 
     dumpSchemaFlag =
-      O.flag CmdRun CmdDumpSchema $
+      O.flag (Run CmdRun) (Run CmdDumpSchema) $
         O.long "dump-schema"
         <> O.help "Dump loaded schema as JSON and exit (for debugging, output structure is unstable)"
+
+    readyFlag =
+      O.flag (Run CmdRun) (Client CmdReady) $
+        O.long "ready"
+        <> O.help "Checks the health of PostgREST by doing a request on the admin server /ready endpoint"
+
 
 exampleConfigFile :: [Char]
 exampleConfigFile =
--- a/src/PostgREST/Client.hs
+++ b/src/PostgREST/Client.hs
@@ -1,0 +1,100 @@
+{-|
+Module      : PostgREST.Client
+Description : PostgREST HTTP client
+-}
+{-# LANGUAGE NamedFieldPuns #-}
+module PostgREST.Client
+  ( ready
+  ) where
+
+import qualified Data.Text                 as T
+import qualified Network.HTTP.Client       as HC
+import qualified Network.HTTP.Types.Status as HTTP
+
+import Network.HTTP.Client (HttpException (..))
+import System.IO           (hFlush)
+
+import PostgREST.Config  (AppConfig (..))
+import PostgREST.Network (isSpecialHostName)
+
+import Protolude
+
+data PgrstClientError
+  = NoAdminServer
+  | NoSpecialHostNamesAllowed Text
+  | PostgRESTNotReady         Text
+  | HTTPConnectionRefused     Text
+  | HTTPExceptionInvalidURL   Text
+
+-- | This is invoked by the CLI "--ready" flag.
+--   The http-client sends and a request to /ready endpoint
+--   and exits with success or failure.
+ready :: AppConfig -> IO ()
+ready AppConfig{configAdminServerHost, configAdminServerPort} = do
+
+  client <- HC.newManager HC.defaultManagerSettings
+  readyURL <- getURL
+  req <- HC.parseRequest (T.unpack readyURL) `catch` handleHttpException
+  resp <- HC.httpLbs req client `catch` handleHttpException
+
+  let status = HC.responseStatus resp
+
+  if status >= HTTP.status200 && status < HTTP.status300
+    then printAndExitWithSuccess $ "OK: " <> readyURL
+    else printAndExitWithFailure $ clientErrorMsg (PostgRESTNotReady readyURL)
+        where
+          getURL :: IO Text
+          getURL =
+            -- Here, we have three cases:
+            -- 1. If the admin port config is not defined, we exit
+            --      with "no admin server error"
+            -- 2. Otherwise, if admin server is running, then we check if
+            --      postgrest server-host is configured with special hostname like "*4",
+            --      if it is, we fail with "no special hostname allowed with "--ready".
+            --      The reason for this is that we can't know the actual address.
+            -- 3. Finally, if we know the "actual" hostname and the port, then we
+            --      construct the URL and return it.
+            case configAdminServerPort of
+              Nothing   -> printAndExitWithFailure $ clientErrorMsg NoAdminServer
+              Just port ->
+                if isSpecialHostName configAdminServerHost
+                  then printAndExitWithFailure $ clientErrorMsg (NoSpecialHostNamesAllowed configAdminServerHost)
+                  else return $ makeReadyUrl port
+
+          -- NOTE: http-client automatically resolves hostnames
+          makeReadyUrl :: Int -> Text
+          makeReadyUrl p = "http://" <> wrapIfIpv6 configAdminServerHost <> ":" <> (T.pack . show) p <> "/ready"
+            where
+              -- IPv6 needs to wrapped in [], it has ':' as separator
+              wrapIfIpv6 :: Text -> Text
+              wrapIfIpv6 s
+                | T.any (== ':') s = "[" <> s <> "]"
+                | otherwise = s
+
+-- | Handle HTTP exception for "http-client" requests
+handleHttpException :: HttpException -> IO a
+handleHttpException (HttpExceptionRequest req _) = do
+  let url = show (HC.getUri req)
+  printAndExitWithFailure $ clientErrorMsg (HTTPConnectionRefused $ T.pack url)
+handleHttpException (InvalidUrlException url _) = do
+  printAndExitWithFailure $ clientErrorMsg (HTTPExceptionInvalidURL $ T.pack url)
+
+-- | Print the message on stdout and exit with success
+printAndExitWithSuccess :: Text -> IO a
+printAndExitWithSuccess msg = putStrLn (T.unpack msg) >> hFlush stdout >> exitSuccess
+
+-- | Print the message on stderr and exit with failure
+printAndExitWithFailure :: Text -> IO a
+printAndExitWithFailure msg = hPutStrLn stderr (T.unpack msg) >> hFlush stderr >> exitWith (ExitFailure 1)
+
+-- | Pgrst client error to error message
+clientErrorMsg :: PgrstClientError -> Text
+clientErrorMsg err = "ERROR: " <>
+  case err of
+    NoAdminServer -> "Admin server is not running. Please check admin-server-port config."
+    NoSpecialHostNamesAllowed host ->
+      "The `--ready` flag cannot be used when server-host is configured as \"" <> host <> "\". "
+      <> "Please update your server-host config to \"localhost\"."
+    PostgRESTNotReady url -> url
+    HTTPConnectionRefused url -> "connection refused to " <> url
+    HTTPExceptionInvalidURL url -> "invalid url - " <> url
--- a/test/io/test_cli.py
+++ b/test/io/test_cli.py
@@ -1,12 +1,14 @@
 "Unit tests for Input/Ouput of PostgREST seen as a black box."
 
 from operator import attrgetter
+import signal
 import subprocess
 import pytest
 from syrupy.extensions.json import SingleFileSnapshotExtension
 import yaml
 
 from config import *
+from postgrest import *
 
 
 class ExtraNewLinesDumper(yaml.SafeDumper):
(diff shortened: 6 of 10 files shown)


postgrest.cabal
```diff
@@ -117,6 +118,7 @@ library
                     , hasql-pool                >= 1.0.1 && < 1.1
                     , hasql-transaction         >= 1.0.1 && < 1.2
                     , heredoc                   >= 0.2 && < 0.3
+                    , http-client               >= 0.7.19 && < 0.8
```
wolfgangwalther, line 122: I checked whether this builds with Nixpkgs `pkgsStatic` - it does.

taimoorzaeem:
@steve-chavez Where do we document this new flag, I guess in [https://docs.postgrest.org/en/v13/references/cli.html](https://docs.postgrest.org/en/v13/references/cli.html)?

steve-chavez:
@taimoorzaeem Yes, should be in CLI.

postgrest.cabal
taimoorzaeem: Hmm, looks like static build failed here: [https://github.com/PostgREST/postgrest/actions/runs/17230219829/job/48882437057?pr=4269](https://github.com/PostgREST/postgrest/actions/runs/17230219829/job/48882437057?pr=4269).

wolfgangwalther: network errors, probably need to re-run it.

steve-chavez: Already restarted it a couple of times, but the network error persists

steve-chavez:
@taimoorzaeem Just tried:

```bash
$ postgrest-run -- --ready
postgrest: HttpExceptionRequest Request {
  host                 = "0.0.0.0"
  port                 = 3001
  secure               = False
  requestHeaders       = []
  path                 = "/ready"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.connect: <socket: 13>: does not exist (Connection refused))
```

The error is expected, but can we make it clearer? (no Haskell type wrapping)

steve-chavez:
Also noted that on success we don't print anything:

```bash
$ postgrest-run -- --ready

# result is good
$ echo $?
0
```

Maybe we should print something like [traefik](https://doc.traefik.io/traefik/reference/install-configuration/observability/healthcheck/#usage):

```
$ postgrest-run -- --ready
OK: http://:<port>/ready
```

src/PostgREST/CLI.hs
```diff
@@ -114,15 +136,20 @@ readCLIShowHelp =
...
     dumpSchemaFlag =
-      O.flag CmdRun CmdDumpSchema $
+      O.flag (Run CmdRun) (Run CmdDumpSchema) $
         O.long "dump-schema"
         <> O.help "Dump loaded schema as JSON and exit (for debugging, output structure is unstable)"
 
+    readyFlag =
+      O.flag (Run CmdRun) (Admin CmdReady) $
+        O.long "ready"
+        <> O.help "Health check for if PostgREST is running and ready"
```
steve-chavez:
```suggestion
        <> O.help "Checks the health of PostgREST by doing a request on the admin server /ready endpoint"
```

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,37 @@
...
+                   <> ":"
+                   <> show port
+
+      escapeHostName :: Text -> Text
+      escapeHostName "*"  = "0.0.0.0"
+      escapeHostName "*4" = "0.0.0.0"
+      escapeHostName "!4" = "0.0.0.0"
+      escapeHostName "*6" = "0.0.0.0"
+      escapeHostName "!6" = "0.0.0.0"
+      escapeHostName h    = h
```
steve-chavez: Dunno if this logic is correct, can't we reuse `resolveHost` instead?

https://github.com/PostgREST/postgrest/blob/fc1a5ecded68ebfb91b621ee16bc29318ff5f529/src/PostgREST/Network.hs#L11-L17

Seeing how we print the admin server log at startup, it should work:
```
26/Aug/2025:13:18:17 -0500: Admin server listening on 0.0.0.0:3001
```

https://github.com/PostgREST/postgrest/blob/fc1a5ecded68ebfb91b621ee16bc29318ff5f529/src/PostgREST/Admin.hs#L32-L33

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,37 @@
...
+import Protolude
+
+ready :: AppConfig -> IO Status
+ready AppConfig{configAdminServerHost, configAdminServerPort} = do
+
+  client <- HC.newManager HC.defaultManagerSettings
+  req <- HC.parseRequest $ -- Create HTTP Request
+    case configAdminServerPort of
+      Just port -> T.unpack (address port) <> "/ready"
+      Nothing   -> panic "Admin Server is not running. Please check your configuration."
```
steve-chavez: How do I get this error manually?

I try:
```
$ PGRST_ADMIN_SERVER_PORT=" " postgrest-with-postgresql-17 -f test/spec/fixtures/load.sql postgrest-run
```

And `--ready` just gives the generic error message:

```
$ postgrest-run -- --ready
postgrest: HttpExceptionRequest Request {
  host                 = "0.0.0.0"
  port                 = 3001
  secure               = False
  requestHeaders       = []
  path                 = "/ready"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.connect: <socket: 13>: does not exist (Connection refused))
```

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,37 @@
...
+
+  resp <- HC.httpLbs req client     -- HTTP Response
+  return $ HC.responseStatus resp
+    where
+      address port = "http://"
+                   <> escapeHostName configAdminServerHost
+                   <> ":"
+                   <> show port
```
steve-chavez: This concatenation is also done on 

https://github.com/PostgREST/postgrest/blob/fc1a5ecded68ebfb91b621ee16bc29318ff5f529/src/PostgREST/Observation.hs#L72-L73

Maybe a chance for adding a reusable function in `PostgREST/Network.hs` (should be fine to have the `http://` prefix on both places)

test/io/test_cli.py
```diff
@@ -289,3 +289,78 @@ def test_jwt_aud_config_set_to_invalid_uri(defaultenv):
...
+
+    env = {
+        **defaultenv,
+        "PGRST_ADMIN_SERVER_PORT": "3001",
+    }
+
+    # Run postgrest process
+    command = [POSTGREST_BIN]
+    env["HPCTIXFILE"] = hpctixfile()
+    pgrst_process = subprocess.Popen(command, env=env)
```
steve-chavez: Since these lines are repeated in the tests, can we refactor them into a reusable function? Seems like it could fit in `test/io/postgrest.py`.

steve-chavez: Maybe they should be coupled with:

```
    finally:
        pgrst_process.kill()
        pgrst_process.wait()
```

src/PostgREST/Client.hs
taimoorzaeem: You need to do this:

```
$ PGRST_ADMIN_SERVER_PORT=" " postgrest-run --ready
```

> `port                 = 3001`

Hmm, this exception shows the second `postgrest-run` has it's own environment variables. I guess I realize that we have a problem here? How does one postgrest instance know the which admin port the other one is using?

steve-chavez: > PGRST_ADMIN_SERVER_PORT=" " postgrest-run --ready
> Hmm, this exception shows the second postgrest-run has it's own environment variables

Thanks, forgot about that. The error shows as expected now. 

> I guess I realize that we have a problem here? How does one postgrest instance know the which admin port the other one is using?

Inside docker compose the `postgrest --ready` command will also use the PostgREST server env vars, so I think we should be fine here.

docs/references/cli.rst
```diff
@@ -49,3 +49,12 @@ Dump Schema
...
 Dumps the schema cache in JSON format.
+
+Ready Flag
+----------
+
+.. code:: bash
+
+  $ postgrest [--ready]
+
+This is analogous to the ``/ready`` endpoint in :ref:`admin_server`. However, instead of an HTTP response, this exits with return code of ``0`` on success and ``1`` on failure.
```
steve-chavez: A bit shorter:
```suggestion
Ready
--------

.. code:: bash

  $ postgrest [--ready]

Makes a request to the ``/ready`` endpoint of the :ref:`admin_server`. It exits with a return code of ``0`` on success and ``1`` on failure.
```

src/PostgREST/Client.hs
taimoorzaeem: We can't reuse the `resolveHost` because we don't have the `NS.Socket` object. It's part of the `AppState` and the health check process can't get it.

~~I think it would be best if use the [getAddrInfo](https://hackage-content.haskell.org/package/network-3.2.8.0/docs/Network-Socket.html#v:getAddrInfo) function to resolve it.~~

Actually, we don't need explicit resolution. The library `http-client` does it for us. We just need to handle special addresses like `escapeHostName "!6" = "[::1]"` etc.

postgrest.cabal
steve-chavez: The network errors were never cleared despite restarts, and they persist on other PRs too: https://github.com/PostgREST/postgrest/pull/4288

steve-chavez: Maybe it's a problem with Cachix?

wolfgangwalther: Looks like it, yes. I'll try to investigate, but not sure how, yet...

Maybe you can ask their support for whether they have an idea?

steve-chavez: Yes, just emailed their support.

steve-chavez: They said they're having issues, check https://status.cachix.org/

steve-chavez: The problem looks solved now.

test/io/test_cli.py
```diff
@@ -289,3 +290,75 @@ def test_jwt_aud_config_set_to_invalid_uri(defaultenv):
...
+    time.sleep(0)  # No waiting and instantly run the healthcheck process.
+    try:
+        # Run healthcheck process
+        command = [POSTGREST_BIN] + ["--ready"]
+        healthcheck_process = subprocess.run(command, env=env, capture_output=True)
+        assert healthcheck_process.returncode == 1
+        assert (
+            "postgrest: health check request failed - connection refused to"
+            in healthcheck_process.stderr.decode()
+        )
```
taimoorzaeem: This test is flaky and fails sometimes because the `subprocess.run` exits without capturing the stderr output. I guess this happens due to output not being flushed properly. Any ideas on how to handle this cleanly?

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,47 @@
...
+      host = escapeHostName $ T.unpack configAdminServerHost
+      -- We don't need to resolve the host name except the postgrest special
+      -- addresses. The http-client package automatically resolves the other
+      -- host names for us.
+      escapeHostName = \case
+        "*"  -> "0.0.0.0"
+        "*4" -> "0.0.0.0"
+        "!4" -> "0.0.0.0"
+        "*6" -> "[::1]" -- TODO: Not covered by tests
+        "!6" -> "[::1]" -- TODO: Not covered by tests
```
steve-chavez: See:
https://github.com/PostgREST/postgrest/blob/b2ce322d56687fc67000736876571d8123585502/src/PostgREST/Response/OpenAPI.hs#L387-L393

Let's refactor and make this logic part of the `Network.hs` module.

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,47 @@
...
+  resp <- HC.httpLbs req client `catch` \(_ :: HttpException) -> do
+    let url = show (HC.getUri req)
+    hPutStrLn stderr $ T.pack $ "postgrest: health check request failed - connection refused to " <> url
+    exitWith $ ExitFailure 1
+
+  return $ HC.responseStatus resp
+    where
+      host = escapeHostName $ T.unpack configAdminServerHost
+      -- We don't need to resolve the host name except the postgrest special
+      -- addresses. The http-client package automatically resolves the other
```
steve-chavez: > postgrest special addresses

I don't remember the details but this is not a "postgrest" convention, it is handled by a package, it would be good to clarify this and find out which package does this logic and add it as a comment.

test/io/test_cli.py
steve-chavez: Maybe instead try to force a schema cache failure so `/ready` always fails. An idea could be to put a bad database connection uri. Check the io tests too, likely this was done before.

src/PostgREST/Client.hs
taimoorzaeem: I noticed one potential issue with the special addresses which is when we run postgrest with the a special address, say `!4`, the address chosen and the socket created with that is hidden in the main postgrest process. How would the `--ready` flag process know what address to send `/ready` request at?

test/io/test_cli.py
taimoorzaeem: No, this was actually the buffer not being flushed issue. I have managed to resolve it. 

However, yes, we should also test this with a schema cache failure.

src/PostgREST/Client.hs
steve-chavez: Hm, if the postgREST server resolves !4 to X hostname, then resolving !4 again on the HTTP client would also point to X? 

Can we repeat the process in the client? Then that should work?

wolfgangwalther: I don't think we can reliably repeat the resolution of the special hosts. The conditions could be different at that time.

We should just make the client fail with an error when the server is set up like this.

taimoorzaeem: > I don't think we can reliably repeat the resolution of the special hosts. The conditions could be different at that time.

Yes, agreed. We use `!4` directly in [bindPortTCP](https://hackage.haskell.org/package/streaming-commons-0.2.3.0/docs/Data-Streaming-Network.html#v:bindPortTCP) which gives us the `Socket`. It picks up _any_ hostname and client process can't know it or get that information in any reliable way.

steve-chavez: > We should just make the client fail with an error when the server is set up like this.

I agree, but our default is `!4`. The error should be clear saying that for this to work you need to override the default and maybe give an example.

test/io/test_cli.py
```diff
@@ -294,3 +295,16 @@ def test_jwt_aud_config_set_to_invalid_uri(defaultenv):
...
+
+
+def test_cli_ready_flag_success(defaultenv):
+    "PostgREST ready flag succeeds when ready"
+
+    env = {**defaultenv, "PGRST_SERVER_HOST": "localhost"}
+
+    with run(env=env) as postgrest:
+        output = cli(["--ready"], env=env)
+        re.match(r"OK: http://:\d+/ready", output)
```
taimoorzaeem: https://github.com/PostgREST/postgrest/blob/cddfb6cf5ea4bde23df27a2f1dfa115c950ac701/test/io/postgrest.py#L96-L106

I noticed that we are defining a lot default configs in the `run(...)` function which is causing issues when testing for the new flag because I need the same config when running the health check process but it gets used up in the `run(..)` function.

So, I'll refactoring this to set the default configs in the `defaultenv`, instead of the `run(..)` function.

steve-chavez: Looks like a good idea to me. Please go ahead.

taimoorzaeem: Upon looking into this, I realize why we can't have static default config for cases like hosts and ports because their definition depend on the runtime environment. So a better would be to add something like `config=env` to the yield here:

https://github.com/PostgREST/postgrest/blob/0acf5a30e67835f6877f10c8963dd1a60c962c78/test/io/postgrest.py#L142-L146

With this we can just access the config with `postgrest.config` whenever we `run` the postgrest in test. So, no need to refactor this for now.

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,99 @@
...
+
+-- | ClientError to error message
+clientErrorMsg :: ClientError -> Text
+clientErrorMsg NoAdminServer                 = "UNREACHABLE: admin server is not running"
+clientErrorMsg (NoSpecialHostNamesAllowed host)
+  = "REFUSED: The --ready flag doesn't work with special hostnames like \"" <> host <> "\"."
+  <> " Alternatively, use the /ready endpoint on the admin server."
+clientErrorMsg (PostgRESTNotReady url)       = "UNAVAILABLE: " <> url
+clientErrorMsg (HTTPConnectionRefused url)   = "HTTP EXCEPTION: connection refused to " <> url
+clientErrorMsg (HTTPExceptionInvalidURL url) = "HTTP EXCEPTION: invalid url - " <> url
```
taimoorzaeem: @steve-chavez @wolfgangwalther I haven't put too much thought into the error messages, any feedback/improvements?

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,99 @@
...
+-- | Print the message on stderr and exit with failure
+printAndExitWithFailure :: Text -> IO a
+printAndExitWithFailure msg = hPutStrLn stderr (T.unpack msg) >> hFlush stderr >> exitWith (ExitFailure 1)
+
+-- | ClientError to error message
+clientErrorMsg :: ClientError -> Text
+clientErrorMsg NoAdminServer                 = "UNREACHABLE: admin server is not running"
+clientErrorMsg (NoSpecialHostNamesAllowed host)
+  = "REFUSED: The --ready flag doesn't work with special hostnames like \"" <> host <> "\"."
+  <> " Alternatively, use the /ready endpoint on the admin server."
```
steve-chavez: I think here we should recommend to change the hostname.

This means that by default, the client won't work in a docker environment right?

steve-chavez: I've been checking and traefik uses `localhost` by default: https://doc.traefik.io/traefik/reference/install-configuration/configuration-options/
 
We might consider changing our host to `localhost` so this works transparently. Could be discussed in another issue.

I also recalled we had some troubles with these special hosts: https://github.com/PostgREST/postgrest/issues/3203 https://github.com/PostgREST/postgrest/issues/3202

For now maybe suggest changing to `localhost`.

src/PostgREST/Client.hs
steve-chavez: Those LGTM. Just had an observation on a particular message ([on the other thread](https://github.com/PostgREST/postgrest/pull/4269/files#r2352983536))

src/PostgREST/Client.hs
taimoorzaeem: > This means that by default, the client won't work in a docker environment right?

Yes, with default configuration, it wouldn't.

steve-chavez:
Once this is ready we should also update our docker docs, see https://github.com/PostgREST/postgrest/issues/4239#issuecomment-3304011834

docs/explanations/install.rst
```diff
@@ -142,6 +142,7 @@ To avoid having to install the database at all, you can run both it and the serv
       ports:
         - "3000:3000"
       environment:
+        PGRST_SERVER_HOST: 127.0.0.1
```
steve-chavez: To be consistent:
```suggestion
        PGRST_SERVER_HOST: localhost # necessary for the `postgrest --ready` command to work
```

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,97 @@
...
+printAndExitWithFailure msg = hPutStrLn stderr (T.unpack msg) >> hFlush stderr >> exitWith (ExitFailure 1)
+
+-- | Pgrst client error to error message
+clientErrorMsg :: PgrstClientError -> Text
+clientErrorMsg NoAdminServer                 = "UNREACHABLE: Admin server is not running. Please check admin-server-port config."
+clientErrorMsg (NoSpecialHostNamesAllowed host)
+  = "REFUSED: The --ready flag cannot be used when server-host is configured as \"" <> host <> "\". Please update your server-host config to \"localhost\"."
+clientErrorMsg (PostgRESTNotReady url)       = "UNAVAILABLE: " <> url
+clientErrorMsg (HTTPConnectionRefused url)   = "HTTP EXCEPTION: connection refused to " <> url
+clientErrorMsg (HTTPExceptionInvalidURL url) = "HTTP EXCEPTION: invalid url - " <> url
```
steve-chavez: @taimoorzaeem Thinking more about this. We have a message + hint on `REFUSED: ...`.

I'm wondering if we should instead reuse our JSON format for these errors (the [Error module](https://github.com/PostgREST/postgrest/blob/0329cb4d19e0c65a7a210926a85b58eda03c225b/src/PostgREST/Error.hs#L1)).  Seeing that we'll have other functionality (https://github.com/PostgREST/postgrest/issues/4239#issuecomment-3155596979) added to the HTTP client, it'd be better for the long run to reuse this to reduce maintenance.

WDYT?

---

The way I'm thinking about this, is that we should add a new error group (group 4, [ref](https://docs.postgrest.org/en/v13/references/errors.html#group-3-jwt)) that has new codes for these (`NoAdminServer` should be PGRST401, `NoSpecialHostNamesAllowed` PGRST402, and so on).

taimoorzaeem: > WDYT?
> 
> The way I'm thinking about this, is that we should add a new error group (group 4, [ref](https://docs.postgrest.org/en/v13/references/errors.html#group-3-jwt)) that has new codes for these (`NoAdminServer` should be PGRST401, `NoSpecialHostNamesAllowed` PGRST402, and so on).

Yeah, this seems like a great idea! It centralizes the error messages and keeps it much more consistent.

docs/explanations/install.rst
taimoorzaeem: As you mentioned in [#4239 (comment)](https://github.com/PostgREST/postgrest/issues/4239#issuecomment-3324444813), maybe we should stick with a direct address instead of `localhost`? So, for consistency, change to the `localhost ` to the `127.0.0.1` elsewhere?

steve-chavez: The thing with `127.0.0.1` is that it's IPv4 only, while `localhost` works for IPv6 too. I think instead we should try to transition to `localhost` everywhere, gradually (and give a good hint in case it fails).

src/PostgREST/Client.hs
taimoorzaeem: Just so we confirm on this, we will be returning a JSON here _even_ when we are not returning a complete HTTP response with `--ready` option. Other errors from Error module are returning with the HTTP response. With `--ready`, we simply exit with a return code after printing to stdout or stderr.

> we should add a new error group (group 4, [ref](https://docs.postgrest.org/en/v13/references/errors.html#group-3-jwt))

The new error code group still looks good though, we would just not be documenting HTTP status code because we aren't returning any.

@steve-chavez Should we proceed with this design?

steve-chavez: > The new error code group still looks good though, we would just not be documenting HTTP status code because we aren't returning any.

This is a great point, many thanks for pointing that out, I missed that part.

At this point mixing the Error module with client errors doesn't seem correct. For example, we have Config errors which are left on their own module:

```bash
$ PGRST_OPENAPI_MODE="asdf" postgrest-with-postgresql-15 -f test/spec/fixtures/load.sql postgrest-run
postgrest: FatalError {fatalErrorMessage = "Error in config Invalid openapi-mode. Check your configuration."}
```

https://github.com/PostgREST/postgrest/blob/5817a89756f57b665a685dc8961b8db762917db8/src/PostgREST/Config.hs#L236

And these don't have error codes. I think error codes are more fit for tracing server failures, not so much for CLI failures, which are immediate feedback. Client errors are the same.

---

So how about this. My maintenance concern is really about the structure of these errors, in particular the prefixes. Having an `OK: ...` prefix is ok, but having `UNREACHABLE:`, `UNAVAILABLE:` seems too varied. So how about just prefixing these with an `ERROR:`? This seems clearer and easier to maintain.

taimoorzaeem: > So how about just prefixing these with an `ERROR:`? This seems clearer and easier to maintain.

Yeah this seems good, clearer and reusable as compared to the previous ones.

taimoorzaeem: When postgrest is busy (say with schema cache reloading), we would get:

```sh
$ postgrest --ready
ERROR: http://localhost:3001/ready
```

In this case only, shouldn't we print something like `BUSY: ...` which is more indicative that postgrest is infact busy?

steve-chavez: > When postgrest is busy (say with schema cache reloading),

:thinking: How do you reproduce that case?

> In this case only, shouldn't we print something like BUSY: ... which is more indicative that postgrest is infact busy?

I think it's the `ERROR:` prefix what's wrong. Maybe we should have something like `READY: ...` and `NOT READY: ...`?

steve-chavez: Taking inspiration from traefik:
```go
		if errPing != nil {
			fmt.Printf("Error calling healthcheck: %s\n", errPing)
			os.Exit(1)
		}

		if resp.StatusCode != http.StatusOK {
			fmt.Printf("Bad healthcheck status: %s\n", resp.Status)
			os.Exit(1)
		}
		fmt.Printf("OK: %s\n", resp.Request.URL)
``` 
(https://github.com/traefik/traefik/blob/6df82676aaf8186215086a1d9e934170fb5db13f/cmd/healthcheck/healthcheck.go#L25-L46)

Maybe `ERROR` is fine. 

> In this case only, shouldn't we print something like BUSY: ... which is more indicative that postgrest is infact busy?

Can we print something like `ERROR: the schema cache loading is in progress`? Can we tell that condition? 

Really the main idea is to save us the maintenance work in coming up with different prefixes.

taimoorzaeem: > > When postgrest is busy (say with schema cache reloading),
> 
> 🤔 How do you reproduce that case?

Please check the `def test_cli_ready_flag_fail_when_schema_cache_not_loaded(..)` test in `test.cli.py`.

> Maybe `ERROR` is fine.

Thinking more about this - you are right, we should stick to `ERROR: ..` to keep things simple. The main goal of this feature is to _return the exit code_, so we shouldn't think too much on the error message, just a simple `ERROR: ..` is fine.

src/PostgREST/Client.hs
```diff
@@ -0,0 +1,98 @@
...
+
+-- | Pgrst client error to error message
+clientErrorMsg :: PgrstClientError -> Text
+clientErrorMsg NoAdminServer                 = "ERROR: Admin server is not running. Please check admin-server-port config."
+clientErrorMsg (NoSpecialHostNamesAllowed host)
+  = "ERROR: The `--ready` flag cannot be used when server-host is configured as \"" <> host <> "\". "
+  <> "Please update your server-host config to \"localhost\"."
+clientErrorMsg (PostgRESTNotReady url)       = "ERROR: " <> url
+clientErrorMsg (HTTPConnectionRefused url)   = "ERROR: connection refused to " <> url
+clientErrorMsg (HTTPExceptionInvalidURL url) = "ERROR: invalid url - " <> url
```
steve-chavez: How about factoring the ERROR prefix so it's not repeated?

src/PostgREST/CLI.hs
```diff
@@ -71,6 +83,13 @@ data CLI = CLI
   }
 
 data Command
+  = Admin AdminCommand
+  | Run RunCommand
+
+data AdminCommand
+  = CmdReady
+
+data RunCommand
```
steve-chavez: Hm, `Admin AdminCommand` is not clear as this is not a command ran by the Admin Server. I think it should be `ClientCommand`.

PR #4274, opened 2025-08-20
fix(admin): metrics endpoint not responding with Content-Type header

### LINKED ISSUE #4271: /metrics endpoint missing Content-Type: text/plain; version=0.0.4 for Prometheus scraping
## Description

I am using PostgREST as a backend API server and exposing metrics on the /metrics endpoint. However, Prometheus scraping fails with the following error:

`Error scraping target: non-compliant scrape target sending blank Content-Type and no fallback_scrape_protocol specified for target`

After investigating, I found that the /metrics endpoint’s HTTP response does not include the required header:

`Content-Type: text/plain; version=0.0.4`

This header is mandatory for Prometheus to successfully scrape metrics.

## Why is this important?

Without the correct Content-Type header, Prometheus marks the metrics target as down and fails to scrape it.

## Suggestion:

Please consider adding native support in PostgREST to either:
- Serve metrics with the correct Content-Type header, or
- Provide a configuration option to specify additional HTTP headers on /metrics responses.

## Workaround:

Currently, the only workaround is to put a reverse proxy in front of PostgREST to inject the required header. Native support would be preferred for ease of use.

Thanks you

taimoorzaeem:
The prometheus metrics text format requires `Content-Type` header for correct scraping which fails otherwise. Closes #4271.

### DIFF
--- a/src/PostgREST/Admin.hs
+++ b/src/PostgREST/Admin.hs
@@ -16,6 +16,7 @@ import Network.Socket.ByteString
 
 import PostgREST.AppState    (AppState)
 import PostgREST.Config      (AppConfig (..))
+import PostgREST.MediaType   (MediaType (..), toContentType)
 import PostgREST.Metrics     (metricsToText)
 import PostgREST.Network     (resolveHost)
 import PostgREST.Observation (Observation (..))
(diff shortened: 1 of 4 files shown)


src/PostgREST/Admin.hs
```diff
@@ -58,7 +59,9 @@ admin appState req respond  = do
       respond $ Wai.responseLBS HTTP.status200 [] (maybe mempty JSON.encode sCache)
     ["metrics"] -> do
       mets <- metricsToText
-      respond $ Wai.responseLBS HTTP.status200 [] mets
+      -- Content-Type is required for prometheus compliance
+      let ctHeader = (hContentType, "text/plain; charset=utf-8")
```
steve-chavez: Maybe we can reuse this?
```suggestion
      let ctHeader = toContentType MTTextPlain
```
https://github.com/PostgREST/postgrest/blob/f01d5bfeb3a2fd1c36415e7fe9976e0bedf62a28/src/PostgREST/MediaType.hs#L56-L62

taimoorzaeem: Yes, this is perfect.

PR #4273, opened 2025-08-20
fix: log listener connection error messages directly

laurenceisla:
Closes #3588 

Previously the listener connection errors logged this:

```console
20/Aug/2025:03:04:50 -0500: Failed listening for database notifications on the "pgrst" channel. Just "connection to server at \"localhost\" (::1), port 6433 failed: session is read-only\n"
```

Now it logs:

```console
20/Aug/2025:03:07:57 -0500: Failed listening for database notifications on the "pgrst" channel. connection to server at "localhost" (::1), port 6433 failed: session is read-only
```

Decided to leave the `ExitFailure 1` logs since they are related to an exception in the listener not to the connection. Also removing them only for these cases doesn't seem worth the effort.

PR #4281, opened 2025-08-25
v13: fix: logging the Haskell type instead of the listener error message directly

taimoorzaeem:
Backport of #4273.

@steve-chavez @wolfgangwalther We missed including this fix in `v13.0.5`, was that on purpose? If not, would be cool if I take the responsibility of backporting fixes?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+- Fix logging the Haskell type instead of the listener error message directly by @laurenceisla in #3588
+
 ## [13.0.5] - 2025-08-24
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


wolfgangwalther:
In Nixpkgs, we use https://github.com/korthout/backport-action to automate backports via just a label. This works well. I think we should do the same.

taimoorzaeem:
> In Nixpkgs, we use https://github.com/korthout/backport-action to automate backports via just a label. This works well. I think we should do the same.

This is pretty cool and comprehensive. However, this wouldn't work for every fix, and would need some intervention before merging the backport. For example, in some cases changelog wouldn't merge in the right section etc.

taimoorzaeem:
I guess, we should still setup some way to automate this because we haven't backported any fix to `v12` even when some fixes do apply there.

wolfgangwalther:
That's because we only backport to the last major not further back. We don't even intent to backport to v12.

wolfgangwalther:
> However, this wouldn't work for every fix, and would need some intervention before merging the backport. For example, in some cases changelog wouldn't merge in the right section etc.

Well yes, but it would still make things easier:
- This action creates a PR, so the diff can be reviewed before merge.
- The changelog file has append mode enabled for merging, so there won't be a merge conflict, which would prevent the action from doing something.

From my experience, almost all backports apply cleanly and the changelog is in the right spot (assuming we had forward ported the changelog version update and backported all fixes in order).

PR #4288, opened 2025-08-26
refactor: deduplicate host resolution logic and observations

taimoorzaeem:
While working on #4269, I noticed some redundant code in `Network` and `Observation` modules. This commit deduplicates the code in those modules.

### DIFF
--- a/src/PostgREST/Network.hs
+++ b/src/PostgREST/Network.hs
@@ -1,21 +1,37 @@
 module PostgREST.Network
-  ( resolveHost
+  ( resolveSocketToAddress
   ) where
 
-import           Data.IP        (fromHostAddress, fromHostAddress6)
 import           Data.String    (IsString (..))
 import qualified Network.Socket as NS
 
 import Protolude
 
-resolveHost :: NS.Socket -> IO (Maybe Text)
-resolveHost sock = do
+-- | Resolves the socket to an address depending on the socket type. The Show
+--   instance of the socket types automatically resolves it to the correct
+--   address. Example resolution:
+-- -----------------------------------------------------
+-- | IPv4         | IPv6             | Unix            |
+-- -----------------------------------------------------
+-- | 127.0.0.1:80 | [2001:db8::1]:80 | /tmp/pgrst.sock |
+-- -----------------------------------------------------
+resolveSocketToAddress :: NS.Socket -> IO Text
+resolveSocketToAddress sock = do
   sn <- NS.getSocketName sock
-  case sn of
-    NS.SockAddrInet _ hostAddr ->  pure $ Just $ fromString $ show $ fromHostAddress hostAddr
-    -- The IPv6 addresses are wrapped in [] brackets. This is done in accordance
-    -- to RFC 3986 (https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2).
-    -- In short, we did this to have a clear separation between the port and host
-    -- because the components of an IPv6 are separated with the ':' character.
-    NS.SockAddrInet6 _ _ hostAddr6 _ -> pure $ Just $ fromString $ "[" ++ show (fromHostAddress6 hostAddr6) ++ "]"
-    _ -> pure Nothing
+  return $ showSocketAddr sn
+
+-- |
+-- >>> let addr_ipv4 = NS.SockAddrInet 80 (NS.tupleToHostAddress (127,0,0,1))
+-- >>> let addr_ipv6 = NS.SockAddrInet6 80 0 (0,0,0,1) 0
+-- >>> let addr_unix = NS.SockAddrUnix "/tmp/pgrst.sock"
+--
+-- >>> showSocketAddr addr_ipv4
+-- "127.0.0.1:80"
+
+-- >>> showSocketAddr addr_ipv6
+-- "[::1]:80"
+--
+-- >>> showSocketAddr addr_unix
+-- "/tmp/pgrst.sock"
+showSocketAddr :: NS.SockAddr -> Text
+showSocketAddr = fromString . show
--- a/src/PostgREST/Observation.hs
+++ b/src/PostgREST/Observation.hs
@@ -21,19 +21,16 @@ import qualified Hasql.Connection           as SQL
 import qualified Hasql.Pool                 as SQL
 import qualified Hasql.Pool.Observation     as SQL
 import           Network.HTTP.Types.Status  (Status)
-import qualified Network.Socket             as NS
 import           Numeric                    (showFFloat)
 import           PostgREST.Config.PgVersion
 import qualified PostgREST.Error            as Error
 
-import Protolude         hiding (toList)
-import Protolude.Partial (fromJust)
+import Protolude hiding (toList)
 
 data Observation
-  = AdminStartObs (Maybe Text) (Maybe Int)
+  = AdminStartObs Text
   | AppStartObs ByteString
-  | AppServerPortObs Text NS.PortNumber
-  | AppServerUnixObs FilePath
+  | AppServerAddressObs Text
   | ExitUnsupportedPgVersion PgVersion PgVersion
   | ExitDBNoRecoveryObs
   | ExitDBFatalError ObsFatalError SQL.UsageError
@@ -69,14 +66,12 @@ type ObservationHandler = Observation -> IO ()
 
 observationMessage :: Observation -> Text
 observationMessage = \case
-  AdminStartObs host port ->
-    "Admin server listening on " <> fromJust host <> ":" <> show (fromIntegral (fromJust port) :: Integer)
+  AdminStartObs address ->
+    "Admin server listening on " <> address
   AppStartObs ver ->
     "Starting PostgREST " <> T.decodeUtf8 ver <> "..."
-  AppServerPortObs host port ->
-    "API server listening on " <> host <> ":" <> show port
-  AppServerUnixObs sock ->
-    "API server listening on unix socket " <> show sock
+  AppServerAddressObs address ->
+    "API server listening on " <> address
   DBConnectedObs ver ->
     "Successfully connected to " <> ver
   ExitUnsupportedPgVersion pgVer minPgVer ->
--- a/src/PostgREST/Admin.hs
+++ b/src/PostgREST/Admin.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE NamedFieldPuns #-}
-
 module PostgREST.Admin
   ( runAdmin
   ) where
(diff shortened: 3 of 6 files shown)


src/PostgREST/Observation.hs
```diff
@@ -69,12 +67,12 @@ type ObservationHandler = Observation -> IO ()
...
+  AdminStartObs address ->
+    "Admin server listening on " <> address
   AppStartObs ver ->
     "Starting PostgREST " <> T.decodeUtf8 ver <> "..."
-  AppServerPortObs host port ->
-    "API server listening on " <> host <> ":" <> show port
+  AppServerPortObs address ->
+    "API server listening on " <> address
   AppServerUnixObs sock ->
     "API server listening on unix socket " <> show sock
```
taimoorzaeem: If we can combine these two observations, then we can deduplicate the code even more in other places. WDYT?

steve-chavez: @taimoorzaeem Cool, yeah combining them seems fine.

src/PostgREST/Observation.hs
```diff
@@ -69,14 +66,12 @@ type ObservationHandler = Observation -> IO ()
...
-  AdminStartObs host port ->
-    "Admin server listening on " <> fromJust host <> ":" <> show (fromIntegral (fromJust port) :: Integer)
+  AdminStartObs address ->
+    "Admin server listening on " <> address
   AppStartObs ver ->
     "Starting PostgREST " <> T.decodeUtf8 ver <> "..."
-  AppServerPortObs host port ->
-    "API server listening on " <> host <> ":" <> show port
-  AppServerUnixObs sock ->
-    "API server listening on unix socket " <> show sock
```
steve-chavez, line 79: This "unix socket" message has changed right? How do we confirm the user knows the address is a unix socket?

src/PostgREST/Network.hs
```diff
@@ -1,17 +1,13 @@
...
-import           Data.IP        (fromHostAddress, fromHostAddress6)
 import           Data.String    (IsString (..))
 import qualified Network.Socket as NS
 
 import Protolude
 
-resolveHost :: NS.Socket -> IO (Maybe Text)
-resolveHost sock = do
+resolveAddress :: NS.Socket -> IO Text
+resolveAddress sock = do
```
steve-chavez: Could we add doctests here?

Could even be another PR, to ensure this PR doesn't change the expected outcome.

src/PostgREST/Observation.hs
taimoorzaeem: I think unix socket addresses are like this `/tmp/http.sock`. So the message we be like:
```
API server listening on /tmp/http.sock
```
So the user can infer from this that the socket is unix socket. We should have a test for this too.

src/PostgREST/Network.hs
taimoorzaeem: > Could we add doctests here?

Yes sure. 

I am getting this error when I run `postgrest-test-doctests` locally:
```
[nix-shell]$ postgrest-test-doctests
- cabal.project
- cabal.project.freeze
<command line>: libX11.so.6: cannot open shared object file: No such file or directory
doctests: fd:15: hGetLine: end of file
```
@wolfgangwalther Could this be related to [#3865 (comment)](https://github.com/PostgREST/postgrest/pull/3865#issuecomment-3151778318)?

src/PostgREST/Network.hs
```diff
@@ -1,17 +1,13 @@
...
   sn <- NS.getSocketName sock
-  case sn of
-    NS.SockAddrInet _ hostAddr ->  pure $ Just $ fromString $ show $ fromHostAddress hostAddr
-    NS.SockAddrInet6 _ _ hostAddr6 _ -> pure $ Just $ fromString $ show $ fromHostAddress6 hostAddr6
-    _ -> pure Nothing
+  return $ fromString $ show sn -- Example: 0.0.0.0:80 or [2001:db8::1]:80
```
taimoorzaeem: One thing that changes with this PR, is that previously we displayed the `IPv6` address _slightly_ wrong, it was like:
```
28/Aug/2025:23:46:41 +0500: Admin server listening on ::1:3001
```
Now, the correct way, it's like:
```
28/Aug/2025:23:46:41 +0500: Admin server listening on [::1]:3001
```
@steve-chavez I think this is a small but separate improvement. How about we do this in a separate PR? We don't have a test for it, so can add that too.

steve-chavez: > How about we do this in a separate PR? We don't have a test for it, so can add that too.

@taimoorzaeem Nice catch. Yes, please go ahead.

src/PostgREST/Network.hs
steve-chavez: Weird, that runs for me:

```
$ postgrest-test-doctests
Downloading the latest package list from hackage.haskell.org
Package list of hackage.haskell.org has been updated.
The index-state is set to 2025-09-01T07:55:10Z.
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2025-07-17T02:51:41Z'
Resolving dependencies...
...
Examples: 166  Tried: 166  Errors: 0  Failures: 0
```

Still, it's very weird that we get "Downloading.." when using Nix, as we're supposed to be in deterministic land. 

@wolfgangwalther Can you remind me how is the `cabal.project/cabal.project.freeze` helping? Can we kill those and prevent the "Downloading..."?

---

@taimoorzaeem Btw, to enable doctest on the Network you need to add it here:
https://github.com/PostgREST/postgrest/blob/272e2e7535049ad9bf83f613ba84997f61329120/test/doc/Main.hs#L8-L25

wolfgangwalther: > Can you remind me how is the `cabal.project/cabal.project.freeze` helping? Can we kill those and prevent the "Downloading..."?

The freeze file makes out cabal CI at least a little bit predictable. Without it, new dependencies appearing on hackage could break CI, when some packages didn't specify proper bounds.

You should not get this download anymore, it has been moved to a separate command in 25e56f32e462933d425c7aa30e0348924c46f3a8. Are you doing this on an old checkout of main?

wolfgangwalther: doc tests work for me, too, btw. I have a wayland-based window manager, so maybe that's why `libX11.so.6` doesn't play a role for me?

(not that I'd understand why it would try to access any window-manager related libs anyway...)

taimoorzaeem: Ah, I got it fixed now. I think it was nix issue where `ghci` from the nix environment was clashing with my local `ghci` installation. So when I ran `ghci` in `nix-shell`, I got:

```
[nix-shell]$ ghci
Loaded package environment from /home/taimoorz/.ghc/x86_64-linux-9.4.8/environments/default
GHCi, version 9.4.8: https://www.haskell.org/ghc/  :? for help
<command line>: libX11.so.6: cannot open shared object file: No such file or directory
```
After uninstalling ghci and cleaning up `.ghc/` directory, it is now working.

```
[nix-shell]$ ghci
GHCi, version 9.4.8: https://www.haskell.org/ghc/  :? for help
ghci>
Leaving GHCi.

[nix-shell]$ postgrest-test-doctests
- cabal.project
- cabal.project.freeze
Examples: 166  Tried: 166  Errors: 0  Failures: 0
```

steve-chavez: > You should not get this download anymore, it has been moved to a separate command in https://github.com/PostgREST/postgrest/commit/25e56f32e462933d425c7aa30e0348924c46f3a8. Are you doing this on an old checkout of main?

Ah, my bad; I was doing this on the `v13` branch, you're right that on `main` there's no downloading.

> The freeze file makes out cabal CI at least a little bit predictable. Without it, new dependencies appearing on hackage could break CI, when some packages didn't specify proper bounds.

Thanks for elaborating. I get the latter but when you say "cabal CI" do you mean the CI of the cabal external repo? 

(Want to clarify this to add a comment on `cabal.project.freeze`, I keep forgetting why we have it in presence of Nix)

wolfgangwalther: > Thanks for elaborating. I get the latter but when you say "cabal CI" do you mean the CI of the cabal external repo?

No, we have cabal running in CI without any Nix involved:

https://github.com/PostgREST/postgrest/blob/41b6ebea583f140a945f4a41914d27e38c87d615/.github/workflows/build.yaml#L180-L210

steve-chavez: @wolfgangwalther Would you say that only matters on non-Nix builds? 

(I guess right now it doesn't matter as much but when we expose postgREST as library on Hackage it will?)

I'm thinking of the following comment addition:

```
-- cabal.project.freeze
-- The freeze file makes cabal builds more deterministic.
-- Some packages in Hackage don't specify proper bounds so without this, new dependencies appearing on Hackage can break builds.
-- This only matters when building postgrest without Nix, which we do on CI.
index-state: hackage.haskell.org 2025-07-08T15:50:10Z
```

wolfgangwalther: > Would you say that only matters on non-Nix builds?

Define "nix build" and "non nix build".

It certainly does **not** matter when you do `nix-build -A postgrestPackage` (or the static build).

It certainly **does** matter in the above GitHub Actions snippet.

I'm not 100% sure about which effects it might have on building with `postgrest-build` inside the nix-shell (and all the other similar tools like tests etc.). I *think* it might have an effect on the local cabal build cache, not invalidating it too early, but not sure.

steve-chavez: > Define "nix build" and "non nix build".

nix build - the one that uses Haskell packages included on nixpkgs https://github.com/PostgREST/postgrest/blob/9ed85efdc10e49b2a4340785e3a900c36ab96a71/nix/overlays/haskell-packages.nix#L54-L60

non-nix build, the one which uses Hackage or Stackage

Is that correct? This might have a place in a markdown file somewhere.

wolfgangwalther: Well, as I wrote above we essentially have 3 levels of builds: fully nix in the sandbox (the first case), entirely not nix with hackage/stackage (the second case) and the last case, which is kind of a mix, because while cabal is going to use the nix-provided dependencies, the file might still have an effect on how cabal plans the build and invalidates the cache.

steve-chavez: I'm seeing an error similar to the [above](https://github.com/PostgREST/postgrest/pull/4288#discussion_r2308229681), although this is when doing `postgrest-build`:

```
64-linux/ghc-9.4.8/postgrest-13.1/t/spec/build/spec/spec-tmp/SpecHelper.dyn_o ) [Source file changed]

<no location info>: error: [-Wmissed-extra-shared-lib, -Werror=missed-extra-shared-lib]
    libHSpostgrest-13.1-inplace.so: cannot open shared object file: No such file or directory
    It's OK if you don't want to use symbols from it directly.
    (the package DLL is loaded by the system linker
     which manages dependencies by itself).
[11 of 51] Compiling Feature.RollbackSpec ( test/spec/Feature/RollbackSpec.hs, dist-newstyle/build/x86_64-linux/ghc-9.4.8/postgrest-13.1/t/spec/build/spec/spec-tmp/Feature/RollbackSpec.o, 
```

The compiling still finishes though.

---

Another transient error when compiling:

```
<no location info>: error:
    dist-newstyle/build/x86_64-linux/ghc-9.4.8/postgrest-13.1/build/PostgREST/Query.o.tmp: renameFile:renamePath:rename: does not exist (No such file or directory)
Error: [Cabal-7125]
Failed to build postgrest-13.1 (which is required by test:spec from postgrest-13.1, exe:postgrest from postgrest-13.1 and others).
```

wolfgangwalther: These seem similar to https://github.com/PostgREST/postgrest/pull/3865#issuecomment-3151778318.

PR #4291, opened 2025-08-29
fix: format of IPv6 address logged at PostgREST startup

taimoorzaeem:
The IPv6 address logged at the startup like `::1:80` was wrong because the port isn't clearly separated. This commit corrects it, now logging as `[::1]:80` as discussed in [#4288 (comment)](https://github.com/PostgREST/postgrest/pull/4288#discussion_r2308273098).

This also allows the refactor on #4288 to go smoothly without changing any behavior.

### DIFF
--- a/src/PostgREST/Network.hs
+++ b/src/PostgREST/Network.hs
@@ -13,5 +13,9 @@ resolveHost sock = do
   sn <- NS.getSocketName sock
   case sn of
     NS.SockAddrInet _ hostAddr ->  pure $ Just $ fromString $ show $ fromHostAddress hostAddr
-    NS.SockAddrInet6 _ _ hostAddr6 _ -> pure $ Just $ fromString $ show $ fromHostAddress6 hostAddr6
+    -- The IPv6 addresses are wrapped in [] brackets. This is done in accordance
+    -- to RFC 3986 (https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2).
+    -- In short, we did this to have a clear separation between the port and host
+    -- because the components of an IPv6 are separated with the ':' character.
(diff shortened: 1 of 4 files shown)


src/PostgREST/Network.hs
```diff
@@ -13,5 +13,5 @@ resolveHost sock = do
   sn <- NS.getSocketName sock
   case sn of
     NS.SockAddrInet _ hostAddr ->  pure $ Just $ fromString $ show $ fromHostAddress hostAddr
-    NS.SockAddrInet6 _ _ hostAddr6 _ -> pure $ Just $ fromString $ show $ fromHostAddress6 hostAddr6
+    NS.SockAddrInet6 _ _ hostAddr6 _ -> pure $ Just $ fromString $ "[" ++ show (fromHostAddress6 hostAddr6) ++ "]"
```
steve-chavez, line 20: `  [::1]:80`

@taimoorzaeem Can you elaborate why the above is the right format? Does it follow an RFC? If so can we add the reference in a comment?

taimoorzaeem: Yes, sure. FYI, this is in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2). Let me add a comment.

steve-chavez: Aha, I see it now:
> A host identified by an Internet Protocol literal address, version 6
   [[RFC3513](https://datatracker.ietf.org/doc/html/rfc3513)] or later, is distinguished by enclosing the IP literal
   within square brackets ("[" and "]")

steve-chavez (approved):

wolfgangwalther:
Backported to v13.

PR #4297, opened 2025-09-02
refactor: move decoding to Query.hs

steve-chavez:
Towards solving https://github.com/PostgREST/postgrest/issues/3934.

This makes the `Statements.hs` internal module pure now as it only produces `Hasql.Snippet`.


taimoorzaeem:
Unrelated to this PR, but since we are discussing refactoring, I think we should have a common module for encoders, for example, we have the following in three different modules:

https://github.com/PostgREST/postgrest/blob/41b6ebea583f140a945f4a41914d27e38c87d615/src/PostgREST/Config/Database.hs#L186-L205

https://github.com/PostgREST/postgrest/blob/41b6ebea583f140a945f4a41914d27e38c87d615/src/PostgREST/SchemaCache.hs#L1142-L1160

https://github.com/PostgREST/postgrest/blob/41b6ebea583f140a945f4a41914d27e38c87d615/src/PostgREST/Query/Statements.hs#L179-L189

It doesn't matter much though, because this is very slight duplication.

steve-chavez:
> Unrelated to this PR, but since we are discussing refactoring, I think we should have a common module for encoders, for example, we have the following in three different modules:

Yeah, I think we did it that way because those modules are supposed to be isolated, there was some discussion about this a long time ago.

Maybe we can solve that problem by defining our own Prelude.

---

For now the aim of this PR is to have the decoding phase at the top module, which makes sense since they're IO really.

src/PostgREST/Query/Wrapper.hs
```diff
@@ -0,0 +1,100 @@
+{-|
+Module      : PostgREST.Query.Wrapper
+Description : PostgREST wrapper queries
+-}
+module PostgREST.Query.Wrapper
+  ( wwrite
```
laurenceisla:
```suggestion
  ( wWrite
```
This looked like a typo at first glance for me, maybe it would be better to capitalize it?

steve-chavez: Renamed this to `mainWrite`

taimoorzaeem:
> Maybe we can solve that problem by defining our own Prelude.

Hmm, interesting! This could also be a step towards moving away from `protolude` as mentioned in [#1804 (comment)](https://github.com/PostgREST/postgrest/issues/1804#issuecomment-2558540698).

PR #4307, opened 2025-09-04
refactor: move OpenAPI queries outside of SchemaCache module

steve-chavez:
Small steps towards solving https://github.com/PostgREST/postgrest/issues/3934 and https://github.com/PostgREST/postgrest/issues/4226

## Problem 

The Query.hs module uses Hasql Dynamic Statements in all places except in the OpenAPI path:

https://github.com/PostgREST/postgrest/blob/b2ce322d56687fc67000736876571d8123585502/src/PostgREST/Query.hs#L206-L216

This makes it hard to refactor the logic since SQL.Snippet and the SchemaCache queries are incompatible since they're Text.

## Solution

With this change now the the above part looks like:

```haskell
OAFollowPriv -> do
  tableAccess <- SQL.statement mempty $ SQL.dynamicallyParameterized (SqlFragment.accessibleTables tSchema) decodeAccessibleIdentifiers configDbPreparedStatements
  accFuncs <-  SQL.statement mempty $ SQL.dynamicallyParameterized (SqlFragment.accessibleFuncs tSchema) SchemaCache.decodeFuncs configDbPreparedStatements
  schDesc <- SQL.statement mempty $ SQL.dynamicallyParameterized (SqlFragment.schemaDescription tSchema) decodeSchemaDesc configDbPreparedStatements
  let tbls = HM.filterWithKey (\qi _ -> S.member qi tableAccess) $ SchemaCache.dbTables sCache

  pure $ MaybeDbResult plan (Just (tbls, accFuncs, schDesc))
OAIgnorePriv -> do
  schDesc <- SQL.statement mempty (SQL.dynamicallyParameterized (SqlFragment.schemaDescription tSchema) decodeSchemaDesc configDbPreparedStatements)
  let tbls = HM.filterWithKey (\(QualifiedIdentifier sch _) _ ->  sch == tSchema) (SchemaCache.dbTables sCache)
      routs = HM.filterWithKey (\(QualifiedIdentifier sch _) _ ->  sch == tSchema) (SchemaCache.dbRoutines sCache)

  pure $ MaybeDbResult plan (Just (tbls, routs, schDesc))

```

Which is more consistent with the rest of the module and allows for moving the query generation to the top module on a later refactor.

Additionally this isolates the OpenAPI query logic, making it easier to remove it later (using `postgrest-openapi`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # The x86-64 is a single-static-binary image built via Nix, see:
 # nix/tools/docker/README.md
 
-FROM ubuntu:noble@sha256:f3b7f1bdfaf22a0a8db05bb2b758535fe0e70d82bea4206f7549f89aa12922f4 AS postgrest
+FROM ubuntu:noble@sha256:9cbed754112939e914291337b5e554b07ad7c392491dba6daf25eef1332a22e8 AS postgrest
 
 RUN apt-get update -y \
     && apt install -y --no-install-recommends libpq-dev zlib1g-dev jq gcc libnuma-dev \


src/PostgREST/Query/SqlFragment.hs
```diff
@@ -590,3 +593,117 @@ handlerF rout = \case
...
+      FROM pg_type
+      UNION
+      SELECT
+        t.oid,
+        b.typbasetype,
+        b.typnamespace AS base_namespace,
+        COALESCE(NULLIF(b.typbasetype, 0), b.oid) AS base_type
+      FROM recurse t
+      JOIN pg_type b ON t.typbasetype = b.oid
+    )
```
steve-chavez, line 647: There's duplication here considering we have https://github.com/PostgREST/postgrest/blob/b2ce322d56687fc67000736876571d8123585502/src/PostgREST/SchemaCache.hs#L408-L418

But there are also differences since OpenAPI doesn't care for:

https://github.com/PostgREST/postgrest/blob/b2ce322d56687fc67000736876571d8123585502/src/PostgREST/SchemaCache/Routine.hs#L62-L63

This query is a bit shorter, and will later serve to isolate the result to a openapi type or we might just drop this once `postgrest-openapi` is integrated.

It's too hard to reuse the query snippets in schema cache here because of the difference in types. Later on when [MultilineStrings](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/multiline_strings.html) is available it'll be easier. For now I think we should just have this duplication.

PR #4314, opened 2025-09-07
fix: not logging OpenAPI queries

steve-chavez:
Closes https://github.com/PostgREST/postgrest/issues/4226.

This requires moving query generation to the top App.hs module.

At this point is also simple to log the transaction variables + the pre-request function call but this is not done here to reduce scope.

### DIFF
--- a/src/PostgREST/Logger.hs
+++ b/src/PostgREST/Logger.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE RecordWildCards #-}
 {-|
 Module      : PostgREST.Logger
 Description : Logging based on the Observation.hs module. Access logs get sent to stdout and server diagnostic get sent to stderr.
@@ -10,10 +11,16 @@ module PostgREST.Logger
   , LoggerState
   ) where
 
-import           Control.AutoUpdate    (defaultUpdateSettings,
-                                        mkAutoUpdate, updateAction)
+import           Control.AutoUpdate                (defaultUpdateSettings,
+                                                    mkAutoUpdate,
+                                                    updateAction)
 import           Control.Debounce
-import qualified Data.ByteString.Char8 as BS
+import qualified Data.ByteString.Char8             as BS
+import qualified Data.Text.Encoding                as T
+import qualified Hasql.Decoders                    as HD
+import qualified Hasql.DynamicStatements.Snippet   as SQL hiding (sql)
+import qualified Hasql.DynamicStatements.Statement as SQL
+import qualified Hasql.Statement                   as SQL
 
 import Data.Time (ZonedTime, defaultTimeLocale, formatTime,
                   getZonedTime)
@@ -26,6 +33,7 @@ import System.IO.Unsafe          (unsafePerformIO)
 
 import PostgREST.Config      (LogLevel (..))
 import PostgREST.Observation
+import PostgREST.Query       (MainQuery (..))
 
 import Protolude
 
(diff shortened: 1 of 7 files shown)


src/PostgREST/Query.hs
```diff
@@ -84,6 +83,16 @@ data QueryResult
...
 
+-- The Queries that run on every request
+data MainQuery = MainQuery
+  { mqTxVars  :: SQL.Snippet       -- ^ the transaction variables that always run on each query
+  , mqPreReq  :: Maybe SQL.Snippet -- ^ the pre-request function that runs if enabled
+  , mqCount   :: SQL.Snippet       -- ^ this count query is actually a fragment of the main query, but the same count query also runs after the main one in the case of `count=estimated`, so it's cached here.
+  -- TODO only one of the following queries actually runs on each request, once OpenAPI is removed from core it will be easier to refactor this
+  , mqMain    :: SQL.Snippet
+  , mqOpenAPI :: (SQL.Snippet, SQL.Snippet, SQL.Snippet)
+  }
```
steve-chavez, line 94: This type is the main change here, it's now produced at the App.hs level and then passed to the main transaction.

src/PostgREST/Logger.hs
```diff
@@ -113,3 +120,20 @@ logWithZTime :: LoggerState -> Text -> IO ()
 logWithZTime loggerState txt = do
   zTime <- stateGetZTime loggerState
   hPutStrLn stderr $ toS (formatTime defaultTimeLocale "%d/%b/%Y:%T %z: " zTime) <> txt
+
+logMainQ :: LoggerState -> MainQuery -> IO ()
+logMainQ loggerState MainQuery{mqOpenAPI=(x, y, z),..} =
+  let snipts  = renderSnippet <$> [mqMain, x, y, z]
```
steve-chavez, line 126: Logging the transaction vars query and the pre-req query should just be:
```suggestion
logMainQ :: LoggerState -> MainQuery -> IO ()
logMainQ loggerState MainQuery{mqOpenAPI=(x, y, z),..} =
  let snipts  = renderSnippet <$> [mqMain, x, y, z, mqTxVars, mqPreReq]
```

Can be done on a later PR.

laurenceisla:
The logs for the SQL queries generated by the root path are printed in multiple lines. They would need to be printed in a single line. We do something similar for errors in Hasql (remove the line breaks and print it in a single one):
https://github.com/PostgREST/postgrest/blob/d512d5c105a6203c7868192c71376110195a9bbe/src/PostgREST/Observation.hs#L158

steve-chavez:
> They would need to be printed in a single line.

@laurenceisla Done.

---

Cachix errors again, only for the pg 17 run though:

```
warning: error: unable to download 'https://postgrest.cachix.org/nar/1c28989c2ef79b00a7d78821a61d482f7edc3e4fa1b452c865b30a1f682c5762.nar.xz': HTTP error 503; retrying in 338 ms
warning: error: unable to download 'https://postgrest.cachix.org/nar/1c28989c2ef79b00a7d78821a61d482f7edc3e4fa1b452c865b30a1f682c5762.nar.xz': HTTP error 503; retrying in 555 ms
```

wolfgangwalther:
Starting with this PR, running the IO tests locally will hang at `test/io/test_io.py::test_log_query[info]` for me.

(before this PR, they were hanging at `test/io/test_io.py::test_db_error_logging_to_stderr[warn]`)

PR #4315, opened 2025-09-07
fix: hashtables <1.4.1 fail to build with GCC 15

develop7:
Repro: 

1. Install GCC 15 (openSUSE Tumbleweed has it for a while now)
2. run `stack build`

Expected: compilation succeds

Instead: build fails building `hashtables-1.3.1`, reproducing https://github.com/gregorycollins/hashtables/issues/98

This PR makes sure `hashtables`' version is  >= 1.4.1, one with the issue fixed. Stack.yaml specifies 1.4.2 because it's the next version after in the stackage snapshots: https://github.com/commercialhaskell/stackage-snapshots/blob/master/lts/24/0.yaml#L5300

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -2,3 +2,7 @@ packages: postgrest.cabal
 tests: true
 package *
   ghc-options: -split-sections
+
+-- avoid hashtables pre-1.4.1 that fail to build with GCC 15; see https://github.com/gregorycollins/hashtables/issues/98
+constraints:
+  hashtables >= 1.4.1
--- a/stack.yaml
+++ b/stack.yaml
@@ -9,6 +9,7 @@ nix:
   pure: false
 
 extra-deps:
+  - hashtables-1.4.2@sha256:4940cab94a15d469845ccf5225f9cb3d354c15e8127ebb58425c8b681f7721d9,10386 # fixes build with GCC 15-ish; https://github.com/gregorycollins/hashtables/issues/98 for details
   - fuzzyset-0.2.4
   - hasql-pool-1.0.1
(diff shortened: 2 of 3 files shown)


cabal.project
```diff
@@ -2,3 +2,7 @@ packages: postgrest.cabal
 tests: true
 package *
   ghc-options: -split-sections
+
+-- avoid hashtables pre-1.4.1 that fail to build with GCC 15; see https://github.com/gregorycollins/hashtables/issues/98
+constraints:
+  hashtables >= 1.4.1
```
wolfgangwalther: This line causes rebuilds for Nix, too - which we don't need, because we're not running GCC 15 there, yet.

You can see this in the failing CI jobs:

```
hashtables-1.4.2 (lib) (requires download & build)
```

Maybe we should just do something like `allow-newer` here for it or so. The stack deps are pinned anyway, and Nix could happily use the older version.

develop7: or maybe just remove it altogether from `cabal.project`? next nixpkgs upgrade will either have 1.4.2 since it's next version in stackage or won't build in case of GCC upgrade only and will be fixed subsequently

wolfgangwalther: if removal works, then yes.

develop7:
rebased the PR against latest `main`, nix build errors seem to be caused by [recent cloudflare incident](https://www.cloudflarestatus.com/incidents/8gmgl950y3h7)

cabal.project
develop7: done, please review again.

wolfgangwalther (approved):

Issue #4316, opened 2025-09-09
test: some `io-tests` are incorrectly passing


taimoorzaeem:
### Description of issue

In our `io-tests`, some tests are incorrectly passing when they should fail. For example:

https://github.com/PostgREST/postgrest/blob/d512d5c105a6203c7868192c71376110195a9bbe/test/io/test_cli.py#L125-L131

There are multiple issues, with this:

- Using the `cli` function has **nothing** to do with testing an error output.
- This test doesn't belong to `test_cli.py`. It should be in `test_io.py`.
- The test passes **without** running the `assert` statement. To verify this, do something like:

```py
with pytest.raises(PostgrestError):
      dump = cli(["--dump-config"], env=env).split("\n")
      assert "This error message is not in the dump" in dump
```

and the test still passes. So, we don't test the error message at all.

### Solution

This should be tested like:

https://github.com/PostgREST/postgrest/blob/d512d5c105a6203c7868192c71376110195a9bbe/test/io/test_io.py#L69-L79

Additionally, when testing error messages, we redirect the `stderr` to `stdout`. I think it should be tested separately.

wolfgangwalther:
I agree on the other things, but not here:
  
> This test doesn't belong to `test_cli.py`. It should be in `test_io.py`.

`test_cli.py` is about running postgrest in a way that returns immediately. Invalid configuration should throw an error immediately as well, the postgrest process does not need to run. This *is* `test_cli`.

I'd argue that the JWT secret length test is wrong in `test_io` instead - it should be possible to test this with `--dump-config` as well, right?

taimoorzaeem:
> `test_cli.py` is about running postgrest in a way that returns immediately. [...] I'd argue that the JWT secret length test is wrong in `test_io` instead

I see, that makes sense, we can do it this way too, as long as we are consistent with where we write them.

> it should be possible to test this with `--dump-config` as well, right?

Not yet, I was insisting on `run(..)` approach because that worked, while `cli(..)` didn't. Two reasons why `cli(...)` doesn't work for testing config errors:

- It doesn't read the `stderr` output, so the error isn't read.
- We raise the `PostgrestError` on failure (because invalid config), which when caught in the calling function, `assert` line never executes, so it always passes.

So, if we fix this, then `cli(...)` will work correctly.

https://github.com/PostgREST/postgrest/blob/d512d5c105a6203c7868192c71376110195a9bbe/test/io/test_cli.py#L67-L83

[...]

PR #4317, opened 2025-09-09
test(io): correct config error tests in test_cli.py

taimoorzaeem:
Towards fixing #4316. The config error tests were not correctly tested. This commit allows capturing `stderr` output and then checks the error messages properly.

This issue was also a blocker for new features like #4269.

Similar tests in `test_io.py` should be moved to `test_cli.py`, which would be done in a later PR.

### DIFF
--- a/test/io/test_cli.py
+++ b/test/io/test_cli.py
@@ -61,26 +61,39 @@ def itemgetter(*items):
 
 
 class PostgrestError(Exception):
-    "Postgrest exited with a non-zero return code."
+    "Postgrest exited unexpectedly."
 
 
-def cli(args, env=None, stdin=None):
-    "Run PostgREST and return stdout."
+def cli(args, env=None, stdin=None, expect_error=False):
+    "Run PostgREST and return stdout or stderr."
     env = env or {}
 
     command = [POSTGREST_BIN] + args
     env["HPCTIXFILE"] = hpctixfile()
 
     process = subprocess.Popen(
-        command, env=env, stdin=subprocess.PIPE, stdout=subprocess.PIPE
+        command,
+        env=env,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
     )
 
     process.stdin.write(stdin or b"")
     try:
-        result = process.communicate(timeout=5)[0]
-        if process.returncode != 0:
-            raise PostgrestError()
-        return result.decode()
+        (stdout_output, stderr_output) = process.communicate(timeout=5)
+        if expect_error:  # When expected to fail, return stderr, else stdout
+            if process.returncode == 0:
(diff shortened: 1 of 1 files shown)


test/io/test_cli.py
```diff
@@ -72,15 +68,17 @@ def cli(args, env=None, stdin=None):
...
     try:
-        result = process.communicate(timeout=5)[0]
-        if process.returncode != 0:
-            raise PostgrestError()
-        return result.decode()
+        stdout_, stderr_ = process.communicate(timeout=5)
+        return process.returncode, stdout_.decode(), stderr_.decode()
     finally:
         process.kill()
         process.wait()
```
wolfgangwalther: I don't particularly like that we're returning exit codes and both stdout and stderr.

Can we keep throwing an error if the non-expected returncode occured?

Aka by default we expect this to exit with 0. But if we pass some kind of argument to `cli(...)`, then it's flipped.

In one case we return stdout, in the other we return stderr.

Should make the code less verbose when using it - and less likely to use it wrong, e.g. by accidentally doing `returncode, stderr = cli(...)` (which would be wrong!)

So maybe something like `cli(..., fail=True)` or so?

taimoorzaeem: > e.g. by accidentally doing `returncode, stderr = cli(...)` (which would be wrong!)

Doing this would be a python error, so python will complain. So we would be able to correct that.

> So maybe something like `cli(..., fail=True)` or so?

Yes this would be very nice to have. However, one issue I see with this is that, if we raise exception inside `cli(..., fail=True)`, then the assertions afterwards _wouldn't_ run, so for that we would need to pass the error message too.

@wolfgangwalther How about we do something like `cli(..., fail=True, error_msg)`? Then we can run the assertion inside the `cli(..)` when `fail == True`.

wolfgangwalther: > Yes this would be very nice to have. However, one issue I see with this is that, if we raise exception inside `cli(..., fail=True)`, then the assertions afterwards _wouldn't_ run, so for that we would need to pass the error message too.

Why? The exception would only be returned if the return code was 0, so no error. There is no point to run the assertions afterwards, if the returncode was already 0.

There is actually no difference to right now, because your first `assert` is always to test the returncode. Once that fails, the next assert is not run anyway.

taimoorzaeem: I see now. That makes sense. Thanks for clearing that up.

test/io/test_cli.py
```diff
@@ -64,23 +64,32 @@ class PostgrestError(Exception):
     "Postgrest exited with a non-zero return code."
```
wolfgangwalther: The error message needs adjustment because it is now used to return zero return code at some times. Maybe parametrize on the actual exit code? Or have two different errors?

Issue #4048, opened 2025-04-28
JWT Token verifies on latest, not on devel


pebosi:
### Environment

* PostgreSQL version:  13.3
* PostgREST version: latest / devel
* Operating system: Docker 

### Description of issue

Currently using latest images with this JWT Secret Config:
```
{
  "alg":"RS256",
  "e":"AQAB",
  "key_ops":["verify"],
  "kty":"RSA",
  "n":"ryPOQAv29sSO9jbDWkte3exY...."
}
```

is working with latest version, but fails with this error on devel:

```
No suitable key or wrong key type
```

pebosi:
Using Config from Docker Swarm Secrect via this ENV's:
```
PGRST_JWT_SECRET: "@/run/secrets/jwt_sig"
PGRST_JWT_CACHE_MAX_LIFETIME: 900
```

taimoorzaeem:
Agree, the underlying error can be exposed as `details` in the error response.

pebosi:
Tested with the artifacts from the pull request the message is now:

```
{"code":"PGRST301","details":"No suitable key was found to decode the JWT","hint":null,"message":"No suitable key or wrong key type"}
```

pebosi:
Tried setting from file, inline in config, as base64, everything worked on latest, no version worked on devel.

taimoorzaeem:
While I can't understand why this error is coming up, I did a little bit of tracing and this error comes from [here](https://github.com/tekul/jose-jwt/blob/1d59c9fdeb6159431ea12289784d487da86914fd/Jose/Jwk.hs#L112):

```haskell
canDecodeJws :: JwsHeader -> Jwk -> Bool
canDecodeJws hdr jwk = jwkUse jwk /= Just Enc &&
    keyIdCompatible (jwsKid hdr) jwk &&
    algCompatible (Signed (jwsAlg hdr)) jwk &&
    case (jwsAlg hdr, jwk) of
        (EdDSA, Ed25519PublicJwk {}) -> True
        (EdDSA, Ed25519PrivateJwk {}) -> True
        (EdDSA, Ed448PublicJwk {}) -> True
        (EdDSA, Ed448PrivateJwk {}) -> True
        (RS256, RsaPublicJwk {}) -> True
        (RS384, RsaPublicJwk {}) -> True
        (RS512, RsaPublicJwk {}) -> True
        (RS256, RsaPrivateJwk {}) -> True
[...]

PR #4321, opened 2025-09-11
docs: clarify wording on kid validation

steve-chavez:
Also [chore(changelog): add breaking change for JWT kid](https://github.com/PostgREST/postgrest/commit/c900c31e84d3b6827f07e898c79b960c8ba88e7f)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - #3956, Drop `/config` endpoint of admin server - @steve-chavez
    + The endpoint was at risk of being left unprotected when exposing it.
    + The accompanying `admin-server-config-enabled` config was also dropped.
+ - #3598, PostgREST now validates the `kid` parameter of the JWT - @wolfgangwalther
+   + If the JWT contains a ``kid`` parameter, then PostgREST will look for the JSON Web Key in the `jwt-secret`.
+   + If the JWT doesn't contain a `kid`, the behavior should be backwards compatible. PostgREST  will try each key in the `jwt-secret` one by one until it finds one that works.
  - #3697, #3602, Querying non-existent table now returns `PGRST205` error instead of empty json - @taimoorzaeem
  - #3600, #3926, Improve JWT errors - @taimoorzaeem
    + Return `PGRST301` error when `Bearer` in auth header is sent empty


steve-chavez:
Also, it'd be better to merge the error message improvements so we can be clear this doesn't happen for symmetric keys ([ref](https://github.com/PostgREST/postgrest/issues/4048#issuecomment-2954995173)).

In the past I made the argument that error messages should be additions, not fixes.. but now that it's so easy to do patch releases and improve things, I think this policy is wrong.

So how about backpatching these two?

https://github.com/PostgREST/postgrest/blob/c900c31e84d3b6827f07e898c79b960c8ba88e7f/CHANGELOG.md#L14-L17

@wolfgangwalther Any opposition on that?

steve-chavez:
I put the CHANGELOG addition on https://github.com/PostgREST/postgrest/releases/tag/v13.0.0 too.

wolfgangwalther:
> In the past I made the argument that error messages should be additions, not fixes.. but now that it's so easy to do patch releases and improve things, I think this policy is wrong.

I'm not opposed to backporting error message changes in principle - but I *am* opposed to relabel the same thing, just to be able to backport it. That's pointless.

Whether that's an addition or a fix should not depend on whether you want to backport it or not - it should depend on the actual change.

We should only backport stuff that is very low in risk to introduce breakage. That also means we won't backport every bugfix - if the changes are too involved, it might not be worth the risk. Every new feature would be too risky for me: The goal is to *stabilize* a certain release, after all, not to "make it better".

The two changes you mentioned are probably very low risk. But do they help that goal of stabilizing the v13 release? I don't think so - so I don't see a real reason to backport them. But as I said earlier: No objection in general.

steve-chavez:
> The goal is to stabilize a certain release, after all, not to "make it better".

My phrasing was wrong, but really my goal was to stabilize v13. Because as it is now, I'm seeing users run into the `kid` problem and not having the clear error messages is puzzling for both me as maintainer and the user. 

> but I am opposed to relabel the same thing, just to be able to backport it. That's pointless.

If we can't relabel the error messages as fixes and not as features/additions, then I don't see how this can be backported as they would not belong to a patch release? I was thinking to move those changelog entries to "Fixed".

Overall the point I'm arguing here is that the new policy for clearer error messages changes should be to label them as fixes. Of course, backporting them would depend on the risk/complexity they introduce.

wolfgangwalther:
> If we can't relabel the error messages as fixes and not as features/additions, then I don't see how this can be backported as they would not belong to a patch release?

I wouldn't label them as fixes, but would still be happy about them in a patch release. Especially when we don't do semantic versioning soon anymore anyway, and only have major.minor for these kinds of changes.

steve-chavez:
Cool, so this new release would be v13.1. Agree?

wolfgangwalther:
No, we don't use the minor version right now. We only use patch. We can use the minor version only from 14 onwards. (needs adjustments to release scripts / logic / CI etc.)

steve-chavez:
Hm, then this would be v13.0.7? How would we label these new entries then? On https://github.com/PostgREST/postgrest/releases/tag/v13.0.6 and others, we have a section "Fixed".

wolfgangwalther:
I'd just label them the same way as they are right now, so that's "Added".

PR #4339, opened 2025-09-17
test(io): remove unused duplicate freeport() function

taimoorzaeem:
Removes dead code essentially. We have this function in `postgrest.py` already.

https://github.com/PostgREST/postgrest/blob/2b8f8c4c47023174a69a92eab0ebfa89a7b5f1ee/test/io/postgrest.py#L176-L184

**Side note:** Maybe we should set up some kind of linter like `pylint` to detect similar issues. These aren't huge issues though, but I think it adds up to the tech debt a little bit.

### DIFF
--- a/test/io/util.py
+++ b/test/io/util.py
@@ -24,14 +24,6 @@ class Thread(threading.Thread):
             raise self._exception
 
 
-def freeport():
-    "Find a free port on localhost."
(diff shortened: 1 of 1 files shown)


steve-chavez:
> Side note: Maybe we should set up some kind of linter like pylint to detect similar issues. These aren't huge issues though, but I think it adds up to the tech debt a little bit.

That one would be good, maybe open an issue so we don't forget :+1:

PR #4341, opened 2025-09-18
fix: not logging tx variables and pre-request

steve-chavez:
Addresses the first item on https://github.com/PostgREST/postgrest/issues/3934

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fix not logging OpenAPI queries when `log-query=main-query` is enabled by @steve-chavez in #4226
 - Fix not logging explain query when `log-query=main-query` is enabled by @steve-chavez in #4319
+- Fix not logging transaction variables and db-pre-request function when `log-query=main-query` is enabled by @steve-chavez in #3934
 
 ### Added
 
--- a/src/PostgREST/Logger.hs
+++ b/src/PostgREST/Logger.hs
@@ -123,7 +123,7 @@ logWithZTime loggerState txt = do
 
 logMainQ :: LoggerState -> MainQuery -> IO ()
 logMainQ loggerState MainQuery{mqOpenAPI=(x, y, z),..} =
-  let snipts  = renderSnippet <$> [mqMain, x, y, z, fromMaybe mempty mqExplain]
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fix not logging OpenAPI queries when `log-query=main-query` is enabled by @steve-chavez in #4226
 - Fix not logging explain query when `log-query=main-query` is enabled by @steve-chavez in #4319
+- Fix not logging transaction variables and db-pre-request function when `log-query=main-query` is enabled by @steve-chavez in #3934
```
laurenceisla, line 12: Ah, I understood that the `main-query` only included the [definition mentioned in the docs](https://docs.postgrest.org/en/v12/references/transactions.html#main-query) (as mentioned in https://github.com/PostgREST/postgrest/pull/3904#issuecomment-2666684939). Not against it, but it confused me a bit at the beginning.

steve-chavez: Right, so I'm thinking that we should just have one `log-query = true` as originally planned. I believe we have some coercion on the configs, so `log-query = "main-query"` will become `true`. Then this would be a backwards compatible change.

Why do this:

- More log-query values add extra effort when testing all the possibilities. (I had to refactor this test on https://github.com/PostgREST/postgrest/pull/4340, it's considerable effort adding testing for these logs)
- Complicates things for the users and us maintainers.
- Originally we added `main-query` because of a limitation of the previous implementation.

WDYT?

laurenceisla:
Should we add a test without pre-request to appease CodeCov? (I think that's the one that's missing)

steve-chavez:
> Should we add a test without pre-request to appease CodeCov? (I think that's the one that's missing)

Will do that

CHANGELOG.md
steve-chavez: Even if it's not a backwards compatible change it's worth it IMO.

It's more beneficial overall to put the effort into logging every SQL query including the [schema cache queries ](https://github.com/PostgREST/postgrest/issues/3934#issuecomment-2702367462) than to worry about testing these multiple config values.

laurenceisla: Yes, that was my initial idea too, but was just limited by the scope in that PR. So I agree, let's go with that.

steve-chavez:
The codecov failure went away, there's an error with the commitlint but that will go away once I squash-merge.

laurenceisla (approved):

PR #4344, opened 2025-09-19
docs: note index on automatic tsvector

steve-chavez:
On https://docs.postgrest.org/en/v13/references/api/tables_views.html#automatic-tsvector-conversion

### DIFF
--- a/docs/references/api/tables_views.rst
+++ b/docs/references/api/tables_views.rst
@@ -224,7 +224,7 @@ This allows using the ``fts`` operator on ``text`` and ``json`` types out of the
   .. code-block:: postgres
 
     CREATE INDEX idx_people_col ON people
-    USING GIN (to_tsvector(my_text_column));
+    USING GIN (to_tsvector('french', my_text_column));
 
 .. _v_filter:
 


docs/references/api/tables_views.rst
```diff
@@ -217,6 +217,15 @@ This allows using the ``fts`` operator on ``text`` and ``json`` types out of the
...
     -d "my_json_column=not.phfts(english).The%20Fat%20Cats"
 
+.. important::
+
+  To ensure this operation is fast, you need to create an index on the expression:
+
+  .. code-block:: postgres
+
+    CREATE INDEX idx_people_col ON people
+    USING GIN (to_tsvector(my_text_column));
```
laurenceisla:
```suggestion
    USING GIN (to_tsvector('french', my_text_column));
```

According [to the docs](https://www.postgresql.org/docs/current/textsearch-tables.html#TEXTSEARCH-TABLES-INDEX):
> Notice that the 2-argument version of to_tsvector is used. Only text search functions that specify a configuration name can be used in expression indexes ([Section 11.7](https://www.postgresql.org/docs/current/indexes-expressional.html)). This is because the index contents must be unaffected by [default_text_search_config](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-DEFAULT-TEXT-SEARCH-CONFIG).

laurenceisla (approved):

Issue #4342, opened 2025-09-18
nix: Setup a python linter for linting python code


taimoorzaeem:
## Problem

We use `pytest` for running our io-tests. We often put redundant/dead code in our test suite and it frequently gets overlooked and leads to doing manual cleanups later such as #4339. It adds up to the tech debt a little bit.

## Solution

Setup `pylint` or any other linter to detect these issues.

PR #4345, opened 2025-09-20
test(io): remove unused module imports

taimoorzaeem:
Towards #4342. Doing some cleanup, so when we actually integrate `pylint`, it applies cleanly.

### DIFF
--- a/test/io/config.py
+++ b/test/io/config.py
@@ -1,8 +1,6 @@
 import os
 import pathlib
 import shutil
-import signal
-
 import pytest
 import uuid
 import yaml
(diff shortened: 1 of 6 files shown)


taimoorzaeem:
I notice that this [big_schema_test](https://github.com/PostgREST/postgrest/actions/runs/17880039689/job/50846532128?pr=4345) fails often because of time out.

https://github.com/PostgREST/postgrest/blob/bcf3457c287f8535cf58a3c4a14a4a36f2599659/test/io/test_big_schema.py#L21

Maybe we should increase the `30` second limit? Or does this signal some other issue? It could be just flakiness too.

PR #4348, opened 2025-09-22
fix: not logging optional rollback query

steve-chavez:
Addresses item 3 on https://github.com/PostgREST/postgrest/issues/3934.

Should close https://github.com/PostgREST/postgrest/issues/3934.

### DIFF
--- a/src/PostgREST/Query.hs
+++ b/src/PostgREST/Query.hs
@@ -19,7 +19,8 @@ import qualified PostgREST.Query.Statements   as Statements
 
 
 import PostgREST.ApiRequest              (ApiRequest (..))
-import PostgREST.ApiRequest.Preferences  (Preferences (..),
+import PostgREST.ApiRequest.Preferences  (PreferTransaction (..),
+                                          Preferences (..),
                                           shouldExplainCount)
 import PostgREST.Auth.Types              (AuthResult (..))
 import PostgREST.Config                  (AppConfig (..))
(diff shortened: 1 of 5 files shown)


src/PostgREST/Query.hs
```diff
@@ -33,26 +34,37 @@ import Protolude hiding (Handler)
...
-  , mqPreReq  :: Maybe SQL.Snippet -- ^ the pre-request function that runs if enabled
+  { mqTxVars         :: SQL.Snippet       -- ^ the transaction variables that always run on each query
+  , mqPreReq         :: Maybe SQL.Snippet -- ^ the pre-request function that runs if enabled
   -- TODO only one of the following queries actually runs on each request, once OpenAPI is removed from core it will be easier to refactor this
-  , mqMain    :: SQL.Snippet
-  , mqOpenAPI :: (SQL.Snippet, SQL.Snippet, SQL.Snippet)
-  , mqExplain :: Maybe SQL.Snippet     -- ^ the explain query that gets generated for the "Prefer: count=estimated" case
+  , mqMain           :: SQL.Snippet
+  , mqOpenAPI        :: (SQL.Snippet, SQL.Snippet, SQL.Snippet)
+  , mqSetConstraints :: Maybe SQL.Snippet -- ^ the query that gets generated for the "Prefer: tx=rollback" case, https://www.postgresql.org/docs/current/sql-set-constraints.html
```
steve-chavez, line 42: I've ran `postgrest-coverage` locally, and it looks like CI is failing from this new field. The Logger addition is covered.

steve-chavez:
> The code to handle transaction rollback is now oddly split over multiple files, which does not help understanding at all - it's much harder to understand.
I don't think we should log SET CONSTRAINTS ALL IMMEDIATE at all. Are we already logging BEGIN .. and COMMIT / ROLLBACK, too? I can't imagine we do, so this seems very inconsistent.

Fair feedback. `SET CONSTRAINTS` does seem more like transaction control like BEGIN/COMMIT as you say, so it does not belong here. Will close.

> Side note: All these fixes to add logging to more queries... look like the whole logging was implemented in the wrong layer.

Hm, so far I think the refactors/fixes make a lot of sense. The question "how many queries are run for each request?" is clearly answered with the type:

https://github.com/PostgREST/postgrest/blob/0329cb4d19e0c65a7a210926a85b58eda03c225b/src/PostgREST/Query.hs#L34-L42

And the Query module is now pure. I guess the advantages will be more clear later on (and sure the code can still be improved).

wolfgangwalther:
> > Side note: All these fixes to add logging to more queries... look like the whole logging was implemented in the wrong layer.
> 
> Hm, so far I think the refactors/fixes make a lot of sense.

Oh yeah, I'm not saying that the fixes didn't make sense - given the way the feature is implemented. But the fact that we need to put query logging into so many different code paths in the codebase, made me wonder whether query logging should have indeed been implemented in hasql or so - thus catching *all* queries, no matter what.

(This was just a thought, nothing actionable - I wouldn't expect any change here)

PR #4349, opened 2025-09-23
docs: document config file parameter on CLI page

### LINKED ISSUE #4300: docs: CLI page does not document config file parameter
### Description of the issue

The [CLI](https://docs.postgrest.org/en/v13/references/cli.html) page doesn't document config file parameter. 

For example:

```
$ postgrest --dump-schema /path/to/postgrest.conf
```

This is allowed. We should document this too.

taimoorzaeem:
The CLI page had the config file parameter missing. This commit adds that. Also adds the CLI usage output which shows all available options. Closes #4300.

### DIFF
--- a/docs/references/cli.rst
+++ b/docs/references/cli.rst
@@ -3,23 +3,45 @@
 CLI
 ===
 
-PostgREST provides a CLI with the commands listed below:
+PostgREST provides a CLI with the options listed below:
+
+.. code:: text
+
+  Usage: postgrest [-v|--version] [-e|--example] [--dump-config | --dump-schema] 
+                 [FILENAME]
+
+    PostgREST / create a REST API to an existing Postgres
+    database
+
+  Available options:
+    -h,--help                Show this help text
+    -v,--version             Show the version information
+    -e,--example             Show an example configuration file
+    --dump-config            Dump loaded configuration and exit
+    --dump-schema            Dump loaded schema as JSON and exit (for debugging,
+                             output structure is unstable)
+    FILENAME                 Path to configuration file
+
+FILENAME
+--------
+
+Runs PostgREST with the given :ref:`file_config`.
 
 Help
 ----
 
 .. code:: bash
 
-  $ postgrest [-h|--help]
+  $ postgrest --help
 
-Shows all the commands available.
+Shows all the options available.
 
 Version
 -------
 
 .. code:: bash
 
-  $ postgrest [-v|--version]
+  $ postgrest --version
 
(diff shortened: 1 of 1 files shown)


docs/references/cli.rst
```diff
@@ -3,23 +3,49 @@
...
 
-PostgREST provides a CLI with the commands listed below:
+PostgREST provides a CLI with the options listed below:
+
+.. code:: text
+
+  Usage: postgrest [-v|--version] [-e|--example] [--dump-config | --dump-schema] 
+                 [FILENAME]
+
+    PostgREST 13.1 (pre-release) / create a REST API to an existing Postgres
```
steve-chavez: Hm, it says pre-release here.

docs/references/cli.rst
```diff
@@ -3,23 +3,49 @@
...
+    FILENAME                 Path to configuration file
+
+Config
+------
+
+.. code:: bash
+
+  $ postgrest /path/to/postgrest.conf
+
+Runs PostgREST with the given :ref:`file_config`. See :ref:`configuration` for information on configurations.
```
steve-chavez: Otherwise this looks redundant with [Config File](https://postgrest--4349.org.readthedocs.build/en/4349/references/configuration.html#file-config). Also`FILENAME` has relation with the above output.

```suggestion
FILENAME
--------------

Runs PostgREST with the given :ref:`file_config`.
```

docs/references/cli.rst
taimoorzaeem:
```suggestion
    PostgREST / create a REST API to an existing Postgres
```
Maybe like this, so we don't need to change this with every new release?

docs/references/cli.rst
```diff
@@ -3,23 +3,45 @@
...
+
+Runs PostgREST with the given :ref:`file_config`.
 
 Help
 ----
 
 .. code:: bash
 
-  $ postgrest [-h|--help]
+  $ postgrest --help
```
steve-chavez, line 35: Why remove the short forms here and below?

taimoorzaeem: I was thinking with this change, we display all the options above, so no need to display alternatives like `-h | --help` here again. 

Plus, I think square bracket are generally to show that it is _optional_. But here, we actually _running_ it so we shouldn't have brackets here.

steve-chavez: Ok cool. Fair enough.

steve-chavez (approved):

Issue #3704, opened 2024-08-23
Slow schema cache loading and double caching schema


MHC2000:
### Environment

* PostgreSQL version: 13.4
* PostgREST version: 12.2.3
* Operating system: SLES 15

As we use a lot of tables it looks like the initial loading and reloading of the schema cache takes more than 15 minutes.
The following log was created after starting postgrest new.
I see two config reloads, but there was none triggered at this time.

API is very slow at this loading time.

Only if the query to the API is without joins to other tables the query runs in normal time.
If the query includes joins to other tables the first query takes several minutes to be executed.

In parallel I've queried the admin-port. Ready-Endpoint is on 503 still after Schema cache loaded is shown in the log file.
And still after the second reload of the config and schema, it's on 503.
Config and Live are directly on 200.
Schema_Cache is round about 12 MB.
So in total the system needs about 10 to 15 minutes to get the whole structure.

I know our structure is quite large and maybe it's too large for postgrest.
But would be interesting why the config and schema seems to be loaded 2 times?

```
23/Aug/2024:18:57:36 +0200: Starting PostgREST 12.2.3...
23/Aug/2024:18:57:36 +0200: Admin server listening on port 3001
23/Aug/2024:18:57:36 +0200: Listening on port 3000
23/Aug/2024:18:57:36 +0200: Successfully connected to PostgreSQL 13.14 (Ubuntu 13.14-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0, 64-bit
23/Aug/2024:18:57:36 +0200: Config reloaded
23/Aug/2024:18:57:40 +0200: Schema cache queried in 547.9 milliseconds
23/Aug/2024:18:57:40 +0200: Schema cache loaded 1286 Relations, 4921 Relationships, 468 Functions, 0 Domain Representations, 4 Media Type Handlers, 1222 Timezones
23/Aug/2024:19:03:31 +0200: Successfully connected to PostgreSQL 13.14 (Ubuntu 13.14-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0, 64-bit
23/Aug/2024:19:03:31 +0200: Config reloaded
23/Aug/2024:19:03:35 +0200: Schema cache queried in 503.1 milliseconds

```

MHC2000:
[...]

PR #4367, opened 2025-10-02
Fix PGRST002 errors not logging the JSON messages and reloading the schema unnecessarily on startup

### LINKED ISSUE #4129: Requests are rejected immediately when loading the Schema Cache on startup
### Environment

* PostgreSQL version: 17.5
* PostgREST version: Docker 13.0.2
* Operating system: MacOS

### Description of issue

Right now, when the Schema Cache is reloaded, the requests wait until the schema is ready, e. g. this test:

https://github.com/PostgREST/postgrest/blob/69072116be330f0a0a90145ea291f423aa5bc5c2/test/io/test_big_schema.py#L10-L11

However this does not happen on startup. It does not wait for the Schema Cache and rejects the request immediately with a 503 response (as a side note, the JSON body with the error response is not logged):

```bash
06/Jun/2025:00:12:57 +0000: Config reloaded
192.168.215.1 - web_anon [06/Jun/2025:00:12:57 +0000] "GET /films HTTP/1.1" 503 119 "" "curl/8.7.1"
192.168.215.1 - web_anon [06/Jun/2025:00:12:57 +0000] "GET /films HTTP/1.1" 503 119 "" "curl/8.7.1"
192.168.215.1 - web_anon [06/Jun/2025:00:12:57 +0000] "GET /films HTTP/1.1" 503 119 "" "curl/8.7.1"
06/Jun/2025:00:12:57 +0000: Schema cache queried in 28.4 milliseconds
```

The client receives this error message:

```json
{"code":"PGRST002","details":null,"hint":null,"message":"Could not query the database for the schema cache. Retrying."}
```

### Expected behavior

To keep consistency, PostgREST should also wait until the Schema Cache is loaded on startup.

laurenceisla:
Closes #4129 as mentioned in the revision https://github.com/PostgREST/postgrest/issues/4129#issuecomment-3354387811.

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -109,10 +109,14 @@ postgrest logLevel appState connWorker =
             runExceptT $ postgrestResponse appState appConf maybeSchemaCache authResult req
 
         response <- either Error.errorResponseFor identity <$> eitherResponse
-        -- Launch the connWorker when the connection is down.  The postgrest
+        -- Launch the connWorker when the connection is down. The postgrest
         -- function can respond successfully (with a stale schema cache) before
-        -- the connWorker is done.
-        when (isServiceUnavailable response) connWorker
+        -- the connWorker is done. However, when there's an empty schema cache
+        -- postgrest responds with the error `PGRST002`; this means that the schema
+        -- cache is still loading, so we don't launch the connWorker here because
+        -- it would duplicate the loading process, e.g. https://github.com/PostgREST/postgrest/issues/3704
+        -- TODO: this process may be unnecessary when the Listener is enabled. Revisit once https://github.com/PostgREST/postgrest/issues/1766 is done
+        when (isServiceUnavailable response && isJust maybeSchemaCache) connWorker
         resp <- do
           delay <- AppState.getNextDelay appState
           return $ addRetryHint delay response
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1892,6 +1892,28 @@ def test_log_error_when_empty_schema_cache_on_startup_to_stderr(defaultenv):
         assert any(log_err_message in line for line in output_start)
 
 
+def test_no_double_schema_cache_reload_on_empty_schema(defaultenv):
+    "Should only load the schema cache once on a 503 error when there's an empty schema cache on startup"
+
+    env = {
+        **defaultenv,
+        "PGRST_INTERNAL_SCHEMA_CACHE_QUERY_SLEEP": "300",
+    }
+
+    with run(env=env, port=freeport(), wait_for_readiness=False) as postgrest:
+        postgrest.wait_until_scache_starts_loading()
+
+        response = postgrest.session.get("/projects")
+        assert response.status_code == 503
+
+        # Should wait enough time to load the schema cache twice to guarantee that the test is valid
+        time.sleep(1)
+
+        response = postgrest.admin.get("/metrics")
+        assert response.status_code == 200
+        assert 'pgrst_schema_cache_loads_total{status="SUCCESS"} 1.0' in response.text
+
+
 @pytest.mark.parametrize("level", ["crit", "error", "warn", "info", "debug"])
 def test_log_pool_req_observation(level, defaultenv):
     "PostgREST should log PoolRequest and PoolRequestFullfilled observation when log-level=debug"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix not logging transaction variables and db-pre-request function when `log-query=main-query` is enabled by @steve-chavez in #3934
 - Fix loading utf-8 config files with `ASCII` locale set in #4386
 - Fix not logging the JSON message to stderr on a `PGRST002` error by @laurenceisla in #4129
+- Fix reloading the Schema Cache unnecessarily on a `PGRST002` error by @laurenceisla in #4367
 
 ### Added
 


src/PostgREST/Error.hs
```diff
@@ -717,7 +717,7 @@ instance ErrorBody Error where
   hint (ApiRequestError err) = hint err
   hint (SchemaCacheErr err)  = hint err
   hint (JwtErr err)          = hint err
-  hint NoSchemaCacheError    = Nothing
+  hint NoSchemaCacheError    = Just "This usually happens when PostgREST is starting up. Try again later."
```
steve-chavez: Can we get the requests to wait like what happens after the schema cache is loaded? This would be the ideal behavior.

src/PostgREST/App.hs
```diff
@@ -109,10 +109,11 @@ postgrest logLevel appState connWorker =
...
 
         response <- either Error.errorResponseFor identity <$> eitherResponse
-        -- Launch the connWorker when the connection is down.  The postgrest
+        -- Launch the connWorker when the connection is down. The postgrest
         -- function can respond successfully (with a stale schema cache) before
-        -- the connWorker is done.
-        when (isServiceUnavailable response) connWorker
+        -- the connWorker is done. However it won't launch the connWorker if
+        -- the schema cache was not loaded beforehand, i.e. on startup.
+        when (isServiceUnavailable response && isJust maybeSchemaCache) connWorker
```
steve-chavez: This is good. When the scache is Nothing then we don't retry.

src/PostgREST/Error.hs
steve-chavez: @laurenceisla Maybe if we drop the maybe from:

https://github.com/PostgREST/postgrest/blob/706a4894308937fa67bac5cbb0c76a6b0dc8f9d1/src/PostgREST/AppState.hs#L85

The waiting will automatically happen

laurenceisla: > Maybe if we drop the maybe

Tried it out and did not work. There is no `Nothing` schema cache if this is the case, we have to use one with empty values (e.g. empty tables, empty functions, etc.). It doesn't wait automatically, it returns an immediate `404` response with this initial schema instead of a `503` with the `PGRST002` error.

> Can we get the requests to wait like what happens after the schema cache is loaded? This would be the ideal behavior.

I also think this would be ideal, it would also stop flooding the logs with `PGRST002` errors for apps with a lot of requests. However this would be a different behavior from what we currently do when we restart the schema and respond before the schema is loaded. For example:

```
-- On startup:

Config reload
-- Requests are immediately rejected here with `PGRST002` errors
Schema queried
-- Requests wait until schema is loaded
Schema loaded

-- On a Schema Cache Reload:

Config reload
-- Requests are immediately handled using the stale schema
Schema queried
-- Requests wait until schema is loaded
Schema Loaded
```

So instead of immediately reject the request with `PGRST002` we would wait instead. I'll check further on how to do that.

steve-chavez: Got it. Maybe this is too complicated right now. As it is, the PR is an improvement though as it surfaces an error.

The only problem is that the message looks a bit sloppy :thinking:. I can't think of a better choice, maybe leave it as before or add a TODO comment.

```suggestion
```

laurenceisla: > Got it. Maybe this is too complicated right now. 

Yeah, can confirm that it's not so trivial.

> The only problem is that the message looks a bit sloppy 🤔. I can't think of a better choice, maybe leave it as before or add a TODO comment.

Yes, wasn't too convinced either. I'll remove it.

src/PostgREST/App.hs
```diff
@@ -109,10 +109,11 @@ postgrest logLevel appState connWorker =
             runExceptT $ postgrestResponse appState appConf maybeSchemaCache authResult req
 
         response <- either Error.errorResponseFor identity <$> eitherResponse
-        -- Launch the connWorker when the connection is down.  The postgrest
+        -- Launch the connWorker when the connection is down. The postgrest
         -- function can respond successfully (with a stale schema cache) before
-        -- the connWorker is done.
-        when (isServiceUnavailable response) connWorker
+        -- the connWorker is done. However it won't launch the connWorker if
+        -- the schema cache was not loaded beforehand, i.e. on startup.
```
steve-chavez: Please elaborate more on that comment. It says it does something but not **why** (because it's wasteful to rerun the scache queries again if it's not completed anyway?)

You can also link to https://github.com/PostgREST/postgrest/issues/3704. The most detail the better because this part is complex and I'm not sure if it's all covered on tests.

test/io/test_io.py
```diff
@@ -1871,6 +1871,35 @@ def test_pgrst_log_503_client_error_to_stderr(defaultenv):
         assert any(log_message in line for line in output)
 
 
+def test_pgrst_log_503_no_schema_cache_startup_error_to_stderr(
+    defaultenv, metapostgrest
+):
+    "Should log the 503 error message when the schema cache is not loaded on startup and not reload the schema cache"
```
steve-chavez: > Fix reloading the Schema Cache unnecessarily on a `PGRST002` error by @laurenceisla in #4367

Is this proving the above too? If so, why not use https://docs.postgrest.org/en/v13/references/observability.html#pgrst-schema-cache-loads-total instead? It would be much more clear.

steve-chavez: This second part is not too clear, I'd suggest rephrasing or splitting into two tests
```suggestion
    "Should log the 503 error message when the schema cache is not loaded on startup
```

src/PostgREST/App.hs
```diff
@@ -109,10 +109,13 @@ postgrest logLevel appState connWorker =
...
         response <- either Error.errorResponseFor identity <$> eitherResponse
-        -- Launch the connWorker when the connection is down.  The postgrest
+        -- Launch the connWorker when the connection is down. The postgrest
         -- function can respond successfully (with a stale schema cache) before
-        -- the connWorker is done.
-        when (isServiceUnavailable response) connWorker
+        -- the connWorker is done. However, when there's an empty schema cache
+        -- postgrest responds with the error `PGRST002`; this means that the schema
+        -- cache is still loading, so we don't launch the connWorker here because
+        -- it would duplicate the loading process, e.g. https://github.com/PostgREST/postgrest/issues/3704
```
steve-chavez, line 117: Just so we don't forget, this might be true:
```suggestion
        -- it would duplicate the loading process, e.g. https://github.com/PostgREST/postgrest/issues/3704
        -- TODO: this process may be unnecessary when the Listener is enabled. Revisit once https://github.com/PostgREST/postgrest/issues/1766 is done
```

test/io/test_io.py
laurenceisla: > Is this proving the above too?

Yes

> If so, why not use https://docs.postgrest.org/en/v13/references/observability.html#pgrst-schema-cache-loads-total instead? It would be much more clear.

Oh, forgot about that metric, it's much clearer now! I separated it in another test (can confirm that it fails in `main` cause it returns `2.0` reloads there).

PR #4374, opened 2025-10-02
feat: change log-query config string value to bool

steve-chavez:
BREAKING CHANGE

As discussed on https://github.com/PostgREST/postgrest/issues/3934#issuecomment-3362806777, this changes log-query to use a bool value, this:

- Simplifies config for users.
- Reduces effort in testing the different combinations.

Closes https://github.com/PostgREST/postgrest/issues/3934


docs/references/configuration.rst
```diff
@@ -714,23 +714,14 @@ log-query
 ---------
 
   =============== =================================
-  **Type**        String
+  **Type**        Boolean
   **Default**     "disabled"
```
taimoorzaeem:
```suggestion
  **Default**     False
```

taimoorzaeem:
I think for commit message, we should use `change` prefix from here?

https://github.com/PostgREST/postgrest/blob/8abc1559c7efae4c1f7953466e8c85ea88b18742/nix/tools/gitTools.nix#L12-L14

docs/references/configuration.rst
laurenceisla:
```suggestion
  **Default**     False
```

docs/references/observability.rst
```diff
@@ -52,14 +52,13 @@ For diagnostic information about the server itself, PostgREST logs to ``stderr``
 SQL Query Logs
 --------------
 
-To log the :ref:`main SQL query <main_query>` executed for a request, set the :ref:`log-query` to ``main-query``.
-It will be logged based on the current :ref:`log-level` setting.
+To log the :ref:`main SQL query <main_query>` executed for a request, set the :ref:`log-query` to ``main-query``. It will be logged based on the current :ref:`log-level` setting.
```
laurenceisla: Hmm... should we still mention the[ Main Query](https://docs.postgrest.org/en/v13/references/transactions.html#main-query) here? Knowing that we now also print the Pre-Request and other stuff. Perhaps just:

```suggestion
To log the SQL query executed for a request, set the :ref:`log-query` to ``true``. It will be logged based on the current :ref:`log-level` setting.
```

laurenceisla (commented):
Some reviews.

- The first commit message has a typo in "bchange".
- For the second commit:

docs/references/configuration.rst
laurenceisla: Oh, just saw it addressed here: https://github.com/PostgREST/postgrest/pull/4374#discussion_r2400507336, disregard.

PR #4383, opened 2025-10-05
refactor: move example config file to Config.hs 

taimoorzaeem:
We usually miss adding or changing the example config file when we do changes to our configurations. Moving the example config file from CLI module to Config module would make it slightly less easy to miss doing the associated changes to the example config file. Of course, this isn't a long term solution as discussed in #3101, but might help in short term.

This also corrects the `log-query` value in the example config file that we missed doing in #4374.


taimoorzaeem:
Quite an unusual error, only getting this in CI - works fine locally.

```error
test/spec/Feature/Auth/JwtCacheSpec.hs:21:1: error:
    Failed to load dynamic interface file for PostgREST.Metrics:
    Bad interface file: /home/runner/work/postgrest/postgrest/dist-newstyle/build/x86_64-linux/ghc-9.4.8/postgrest-13.1/build/PostgREST/Metrics.dyn_hi
        /home/runner/work/postgrest/postgrest/dist-newstyle/build/x86_64-linux/ghc-9.4.8/postgrest-13.1/build/PostgREST/Metrics.dyn_hi: openBinaryFile: does not exist (No such file or directory)
   |
21 | import PostgREST.Metrics               (MetricsState (..))
```

**EDIT:** I couldn't figure out why this error is coming up, so thought removing `str` quasi-quoter would help, but the error persists. :eyes:

steve-chavez:
> EDIT: I couldn't figure out why this error is coming up, so thought removing str quasi-quoter would help, but the error persists. 👀

I'm puzzled by that error too, it looks like it's related to caching.

I already tried restarting CI twice (yesterday and today).

taimoorzaeem:
> I'm puzzled by that error too, it looks like it's related to caching.

https://github.com/PostgREST/postgrest/blob/ea2c64da541f908af89c7f3630fcdf660196de47/.github/workflows/build.yaml#L200-L206

Maybe related to how we cache `dist-newstyle` here? This is unlikely though because the suffix would have changed with the source file changes and cache must have been invalidated. 

FWIW, maybe we can try deleting the cache as described in [ref1](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches) and [ref2](https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions)?

steve-chavez:
> FWIW, maybe we can try deleting the cache as described in [ref1](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches) and [ref2](https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions)?

Hm, I don't think the github caches matter here. The problem is on Nix caches?

I would suggest closing and opening another PR, chances are it might solve itself.

taimoorzaeem:
Closing because of consistent CI failures. Re-opened in #4390.

PR #4390, opened 2025-10-08
refactor: move example config file to Config.hs

taimoorzaeem:
We usually miss adding or changing the example config file when we do changes to our configurations. Moving the example config file from CLI module to Config module would make it slightly less easy to miss doing the associated changes to the example config file. Of course, this isn't a long term solution as discussed in #3101, but might help in short term.

This also corrects the `log-query` value in the example config file that we missed doing in #4374.


taimoorzaeem:
Hmm, same error again. :eyes: 

```
test/spec/Feature/Auth/JwtCacheSpec.hs:21:1: error:
    Failed to load dynamic interface file for PostgREST.Metrics:
    Bad interface file: /home/runner/work/postgrest/postgrest/dist-newstyle/build/x86_64-linux/ghc-9.4.8/postgrest-13.1/build/PostgREST/Metrics.dyn_hi
        /home/runner/work/postgrest/postgrest/dist-newstyle/build/x86_64-linux/ghc-9.4.8/postgrest-13.1/build/PostgREST/Metrics.dyn_hi: openBinaryFile: does not exist (No such file or directory)
   |
21 | import PostgREST.Metrics               (MetricsState (..))
```

taimoorzaeem:
I was working around with a locally patched version of `configurator-pg` and I got very similar `Failed to load dynamic interface file` error. Naively, I updated the  `cabal.project` file with the following and that fixed the issue locally.

```cabal
packages: postgrest.cabal
tests: true
shared: True -- this fixed the dynamic interface file issue
```

Note sure if this is related here. :confused:

wolfgangwalther:
This seems to be fixed by 32eb6ebf19c3a2ec5b11a3270b570742f66039cc.

PR #4392, opened 2025-10-09
nix: adjust release tool to new versioning scheme

### LINKED ISSUE #4166: New versioning scheme once more
In #3113 we last changed our versioning scheme and release process. A lot of things work really well since we did that:
- Backports of fixes and the related patch releases.
- Versioning is consistently increasing on the main branch for the devel release, bots like renovate are not confused anymore.

Right now we use semantic versioning. With #2275 in mind, this is a problem, because packages on hackage are supposed to follow [PVP](https://pvp.haskell.org/). This is important when we release on hackage, otherwise that will be very confusing for anyone consuming PostgREST as a library. This didn't happen much, yet, but we had some questions asked in similar directions lately.

Also, the current release process still has one problem: We are still blocking breaking changes after a new major release to "not force a new major immediately". For example, if that wasn't the case, #3865 would have long been done already. This is a problem, because updating nixpkgs is critical for the dependencies we provide for the static build. Those are quite outdated again by now. The intent for the new release process was to be able to both provide stable releases with bugfixes (by backporting) *and* to be able to move freely on the main branch.

Also, one more observation: We are not really using minor releases much, if at all. They don't seem to be useful for us. We really need two modes: Shipping bugfixes and shipping new stuff. But we don't need a third mode of "shipping new stuff without breaking changes". Realistically... shipping "non-breaking updates" is about "updating without having to touch code". But to make use of new features... you'd need to touch the code *anyway*!

Also note: PostgreSQL only does major + patch, essentially - no minor version.

Thus, I'd like to propose the following new versioning scheme - a semantic/PVP hybrid:
- We have two versions: The app version and the library version.
- The app version consists of major and patch. For example: 14.0, 14.1, etc.
- The app version is displayed in the CLI output, the OpenAPI output etc. - in all the places, where we currently display the version.
- The library version consists of 4 parts, matching PVP.
- The library version is kept up-to-date in postgresql.cabal.
- The library version is used on releases to hackage.
- The main branch is always updated to the next major version after a major release as the new development version, similar to GHC or Node. v14 will be the next stable version, v15 the next development branch, v16 the next stable version after that. Similar to what we do with the minor version now, uneven numbers are never really released.

This means we would comply to PVP, so could properly release on hackage. But we also have the relevant information for our end users in the two-component version number.

The PVP number would look like this `A.B.C.D`, where:
- `A` is major towards the user and major towards library consumers. Anything can be broken.
- `B` is patch towards the user, but major towards library consumers. That means any bugfix for users can break any library interface, which is important to move freely.
- `C` is *no change* for the user, but can introduce new API for the library in a non-breaking way. Fully PVP.
- `D` is *no change* for the user, and can only contain library-related bugfixes without changing the API. Fully PVP. We'd probably put changes to allow more versions for our haskell dependencies (see the many open PRs...) in this category, so they could be released to hackage easily.

Imho, this should solve all the problems we currently have.

wolfgangwalther:
Resolves #4166

Not sure whether that works on all ends, yet, but that's what I came up with on the spot.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,7 +57,7 @@ jobs:
         run: |
           cabal_version="$(grep -oP '^version:\s*\K.*' postgrest.cabal)"
 
-          if [[ "$cabal_version" == *.*.* ]]; then
+          if [[ "$cabal_version" == *.*.*.* ]]; then
             git fetch --tags
 
             if [ -z "$(git tag --list "v$cabal_version")" ]; then
--- a/nix/tools/release.nix
+++ b/nix/tools/release.nix
@@ -20,27 +20,24 @@ let
         git diff --exit-code HEAD postgrest.cabal > /dev/null
         trap "" ERR
 
+        # TODO: Support C+D bumps when implementing hackage releases
         bump () {
           current_version="$(grep -oP '^version:\s*\K.*' postgrest.cabal)"
           # shellcheck disable=SC2034
-          IFS=. read -r major minor patch <<< "$current_version"
+          IFS=. read -r A B C D <<< "$current_version"
           echo "Current version is $current_version"
 
           case "$1" in
-            major)
-              new_version="$((major+1)).0.0"
-              new_docs_version="$((major+1)).0"
+            A)
+              new_version="$((A+1)).0"
+              new_docs_version="$((A+1))"
               ;;
-            minor)
-              new_version="$major.$((minor+1)).0"
-              new_docs_version="$major.$((minor+1))"
-              ;;
-            patch)
-              new_version="$major.$minor.$((patch+1))"
-              new_docs_version="$major.$minor"
+            B)
+              new_version="$A.$((B+1))"
+              new_docs_version="$A"
               ;;
             devel)
-              new_version="$major.$((minor+1))"
+              new_version="$((A+1))"
               new_docs_version="devel"
               ;;
           esac
@@ -55,13 +52,9 @@ let
 
         today_date_for_changelog="$(date '+%Y-%m-%d')"
         if [[ "$current_branch" == "main" ]]; then
-          if [[ "$_arg_major" == "on" ]]; then
-            bump major
-          else
-            bump minor
-          fi
+          bump A
         else
-          bump patch
+          bump B
         fi
 
         echo "Updating CHANGELOG.md ..."
@@ -75,10 +68,10 @@ let
           bump devel
 
           # The order of operations is important here:
-          # - bump devel is run and $major is upated to the new version
-          # - the branch is created with the new major, but the commit before the devel bump
+          # - bump devel is run and $A is upated to the new version
+          # - the branch is created with the new A, but the commit before the devel bump
           # - the devel bump is committed
-          git branch -f "v$major"
+          git branch "v$A"
 
           echo "Committing (devel bump)..."
           git commit -m "bump version to $new_version" > /dev/null
@@ -90,7 +83,7 @@ let
 
         if [[ "$current_branch" == "main" ]]; then
           push1="git push $remote $current_branch"
-          push2="git push $remote v$major --force"
+          push2="git push $remote v$A"
         else
           push1="git push $remote $current_branch"
           push2=""
--- a/src/PostgREST/Version.hs
+++ b/src/PostgREST/Version.hs
@@ -11,11 +11,11 @@ import Protolude
 version :: [Text]
 version = T.splitOn "." VERSION_postgrest
 
--- | User friendly version number such as '1.1.1'.
--- Pre-release versions are tagged as such, e.g., '1.1 (pre-release)'.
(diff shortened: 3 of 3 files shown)


steve-chavez:
Seems to be working good, I did `postgrest-release` on main and it generated two new commits:

```diff
commit 78f84e9ac3c409ee19bdcea39b78c7bfe1d1e878 (HEAD -> main)
Author: steve-chavez <stevechavezast@gmail.com>
Date:   Thu Oct 9 12:55:48 2025 -0500

    bump version to 15

diff --git a/docs/conf.py b/docs/conf.py
index f4688e28..42eafd57 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,7 +50,7 @@ source_suffix = ".rst"
 master_doc = "index"
 
 # This is overriden by readthedocs with the version tag anyway
-version = "14"
+version = "devel"
 # To avoid repetition in <title> we set this to an empty string.
 release = ""
 
diff --git a/postgrest.cabal b/postgrest.cabal
index 35d21f51..c536056f 100644
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            14.0
+version:            15
 synopsis:           REST API for any Postgres database
 description:        Reads the schema of a PostgreSQL database and creates RESTful routes
                     for tables, views, and functions, supporting all HTTP methods that security

commit 3bb3004f7efaa754a815a007b4d490e16b96f7a8 (v14)
Author: steve-chavez <stevechavezast@gmail.com>
Date:   Thu Oct 9 12:55:48 2025 -0500

    bump version to 14.0

diff --git a/CHANGELOG.md b/CHANGELOG.md

diff --git a/CHANGELOG.md b/CHANGELOG.md
index d8581c3c..d30135ff 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [14.0] - 2025-10-09
+
 ### Fixed
 
 - Fix not logging OpenAPI queries when `log-query=main-query` is enabled by @steve-chavez in #4226
diff --git a/docs/conf.py b/docs/conf.py
index 42eafd57..f4688e28 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,7 +50,7 @@ source_suffix = ".rst"
 master_doc = "index"
 
 # This is overriden by readthedocs with the version tag anyway
-version = "devel"
+version = "14"
 # To avoid repetition in <title> we set this to an empty string.
 release = ""
 
diff --git a/postgrest.cabal b/postgrest.cabal
index d9824df5..35d21f51 100644
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -1,5 +1,5 @@
 name:               postgrest
-version:            13.1
+version:            14.0
 synopsis:           REST API for any Postgres database
 description:        Reads the schema of a PostgreSQL database and creates RESTful routes
                     for tables, views, and functions, supporting all HTTP methods that security

commit 706a4894308937fa67bac5cbb0c76a6b0dc8f9d1 (origin/main)

```

steve-chavez:
After this we need to document this new release model. I'll take care of that.

steve-chavez:
Dunno what failed here, but v14 didn't got published https://github.com/PostgREST/postgrest/commit/23750e644fca895b0153f84c62df64bf16166a78 after doing `postgrest-release`

steve-chavez:
Looks like this didn't fire? (No tag on github)

https://github.com/PostgREST/postgrest/blob/23750e644fca895b0153f84c62df64bf16166a78/.github/workflows/release.yaml#L23-L56

steve-chavez:
Pushed the v14 tag manually https://github.com/PostgREST/postgrest/releases/tag/v14.0

steve-chavez:
I don't know what happened with readthedocs, but https://postgrest.org/ was redirecting to https://docs.postgrest.org/en/v13/.

I went to RTD dashboard and `v13` was highlighted as default. I changed this default manually to v14.

laurenceisla:
> I went to RTD dashboard and `v13` was highlighted as default. I changed this default manually to v14.

Perhaps we should redirect it to `stable`? It would take the latest stable release (v14) that way.

steve-chavez:
> Perhaps we should redirect it to stable? It would take the latest stable release (v14) that way.

Makes sense. Done.

steve-chavez:
Something weird after putting the default to stable, it shows this notice at the top right:

image

steve-chavez:
Reverted back to having the default as v14 and now the above notice doesn't show anymore

wolfgangwalther:
> Looks like this didn't fire? (No tag on github)

Indeed the `tag` job pushed a `devel` tag: https://github.com/PostgREST/postgrest/actions/runs/18793921577/job/53631536567

Will look into that later.

wolfgangwalther:
The `tag` job should be fixed in 6f46233ad7dcb48fac84a05822bef04d2024461a.

PR #4396, opened 2025-10-10
perf: Index various lists in SchemaCache to change complexity from O(n*n) to O(n)

### LINKED ISSUE #3704: Slow schema cache loading and double caching schema

### LINKED ISSUE #4360: Schema cache reload with large relationship count causes 100+ second API blocking
### Environment

* PostgreSQL version: 15.8.1.040 (Supabase local)
* PostgREST version: v12.2.3 (Supabase local)
* Operating system: Docker on macOS 15.7

### Description of issue

**Expected:** Schema cache reloads complete within seconds.
**Actual:** 100-second delay with 100% CPU usage, blocking all API requests.

**This is a production-blocking issue for us.** Every database schema migration triggers a PostgREST schema cache reload, which blocks our entire API for 100+ seconds. This makes deployments extremely disruptive. 

#### Evidence

```
$ docker logs -f --timestamps supabase_rest_arcip

2025-09-26T08:37:29.866368343Z 26/Sep/2025:08:37:29 +0000: Starting PostgREST 12.2.3 (519615d)...
2025-09-26T08:37:29.866582885Z 26/Sep/2025:08:37:29 +0000: Admin server listening on port 3001
2025-09-26T08:37:29.866985885Z 26/Sep/2025:08:37:29 +0000: Listening on port 3000
2025-09-26T08:37:29.896365301Z 26/Sep/2025:08:37:29 +0000: Listening for notifications on the "pgrst" channel
2025-09-26T08:37:29.906531260Z 26/Sep/2025:08:37:29 +0000: Successfully connected to PostgreSQL 15.8 on aarch64-unknown-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit
2025-09-26T08:37:29.933404135Z 26/Sep/2025:08:37:29 +0000: Config reloaded
2025-09-26T08:37:33.235558095Z 26/Sep/2025:08:37:33 +0000: Schema cache queried in 404.1 milliseconds
2025-09-26T08:39:07.222539597Z 26/Sep/2025:08:37:33 +0000: Schema cache loaded 161 Relations, 1026 Relationships, 18 Functions, 0 Domain Representations, 4 Media Type Handlers, 1194 Timezones
```

```
$ docker stats supabase_rest_arcip --no-stream --format "table {{.Container}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}\t{{.NetIO}}\t{{.BlockIO}}"

supabase_rest_arcip   100.29%   224MiB / 2.845GiB   7.69%     209MB / 4.01MB   28.3MB / 0B
```

#### Related Issues
- https://github.com/PostgREST/postgrest/issues/3046
- https://github.com/PostgREST/postgrest/issues/2450

mkleczek:
Fixes #4360
Fixes #3704 

- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix loading utf-8 config files with `ASCII` locale set in #4386
 - Fix not logging the JSON message to stderr on a `PGRST002` error by @laurenceisla in #4129
 - Fix reloading the Schema Cache unnecessarily on a `PGRST002` error by @laurenceisla in #4367
+- Fix schema cache loading taking long time for larche schemas  #4360  #3704
 
 ### Added
 


mkleczek:
Wow, it looks like it is too fast now :) - test_big_schema assertion that schema reloading takes 10s fails quite dramatically.

wolfgangwalther:
Haha, that's nice. We should do some kind of consistency check as well, that the schema cache still returns all the same things. But... looks great so far. It's exactly the kind of problem I assumed to exist with these relationships...

test/io/test_big_schema.py
```diff
@@ -30,7 +30,7 @@ def test_requests_with_resource_embedding_wait_for_schema_cache_reload(defaulten
         plan_dur = parse_server_timings_header(response.headers["Server-Timing"])[
             "plan"
         ]
-        assert plan_dur > 10000.0
+        assert plan_dur < 10000.0
```
wolfgangwalther: hrhr, nice change :D

mkleczek:
> Haha, that's nice. We should do some kind of consistency check as well, that the schema cache still returns all the same things. But... looks great so far. It's exactly the kind of problem I assumed to exist with these relationships...

I am wondering what test we could add... This is a kind of refactoring that does not change any behavior (except performance) and should be covered by existing tests. WDYT?

wolfgangwalther:
I wouldn't add an automated test for it, but it should be simple to load the big schema fixtures from that IO test and then run a `--dump-schema` on it before and after this change. Then look at the diff. Might need to pipe through `jq` to sort keys or so. Might still end up with a diff based on some ordering, not sure. But that should give us a result whether... just all relationships are missing or so. (which is extremely unlikely)

steve-chavez:
I think @MHC2000 will be happy about this

I can confirm the schema privately shared on https://github.com/PostgREST/postgrest/issues/3704#issuecomment-3133324555

Goes from ~7 minutes:

```
06/Oct/2025:16:06:50 -0500: Schema cache queried in 349.6 milliseconds
06/Oct/2025:16:06:50 -0500: Schema cache loaded 919 Relations, 3312 Relationships, 249 Functions, 0 Domain Representations, 4 Media Type Handlers, 1196 Timezones
06/Oct/2025:16:13:43 -0500: Schema cache loaded in 413123.5 milliseconds
```

To ~14 seconds:

```
10/Oct/2025:16:02:15 -0500: Schema cache queried in 416.6 milliseconds
10/Oct/2025:16:02:15 -0500: Schema cache loaded 919 Relations, 3312 Relationships, 249 Functions, 0 Domain Representations, 4 Media Type Handlers, 1196 Timezones
10/Oct/2025:16:02:29 -0500: Schema cache loaded in 14115.5 milliseconds
```

mkleczek:
I've updated the patch to cover more cases:

- addViewM2OAndO2ORels
- addViewPrimaryKeys

mkleczek:
> I think @MHC2000 will be happy about this
> 
> I can confirm the schema privately shared on [#3704 (comment)](https://github.com/PostgREST/postgrest/issues/3704#issuecomment-3133324555)
> 
> Goes from ~7 minutes:
[...] 
> To ~14 seconds:

@steve-chavez can you check again with the latest version of the patch? It should improve the times even more.

steve-chavez:
@mkleczek It's now down to ~2 seconds :rocket: 
```
11/Oct/2025:07:02:14 -0500: Schema cache queried in 448.0 milliseconds
11/Oct/2025:07:02:14 -0500: Schema cache loaded 919 Relations, 3312 Relationships, 249 Functions, 0 Domain Representations, 4 Media Type Handlers, 1196 Timezones
11/Oct/2025:07:02:16 -0500: Schema cache loaded in 2447.1 milliseconds
```

mkleczek:
> @mkleczek It's now down to ~2 seconds 🚀
> 

Amazing, thanks for checking.

I guess it is in mergeable state - patch code coverage is somewhat low but I think the changes touched code not covered by tests originally. It should be probably fixed but I am not sure if this PR is the right one though.
(Interestingly, overall code coverage is better than before)

test/io/test_big_schema.py
wolfgangwalther: While the change is nice, it does break the test expectation. This is not the primary test outcome - we are testing whether other stuff waits on the schema cache reload here. So a high plan duration is a requirement to make the test effective.

We will either need to throw the test away or rewrite it or... I don't know.

src/PostgREST/SchemaCache.hs
```diff
@@ -554,13 +555,14 @@ addViewPrimaryKeys tabs keyDeps =
   where
     findViewPKCols sch vw =
       concatMap (\(ViewKeyDependency _ _ _ _ pkCols) -> takeFirstPK pkCols) $
-      filter (\(ViewKeyDependency _ viewQi _ dep _) -> dep == PKDep && viewQi == QualifiedIdentifier sch vw) keyDeps
+      fold $ HM.lookup (PKDep, QualifiedIdentifier sch vw) indexedDeps
```
wolfgangwalther, line 560: The only diff in the schema-cache I get is here. The big schema has many of these:

```diff
@@ -858435,8 +858435,8 @@
         "tableIsView": true,
         "tableName": "v_pop_ohnekoord",
         "tablePKCols": [
-          "ap_id",
-          "id"
+          "id",
+          "ap_id"
         ],
         "tableSchema": "apflora",
         "tableUpdatable": false
```

Aka, the order of `tablePKCols` is changed.

I'm not exactly sure whether we rely on this order anywhere.

src/PostgREST/SchemaCache.hs
```diff
@@ -477,14 +478,13 @@ addViewM2OAndO2ORels keyDeps rels =
   where
     isM2O card = case card of {M2O _ _       -> True; _ -> False;}
     isO2O card = case card of {O2O _ _ False -> True; _ -> False;}
-    viewRels Relationship{relTable,relForeignTable,relCardinality=card} =
-      if isM2O card || isO2O card then
+    viewRels Relationship{relTable,relForeignTable,relCardinality=card} | isM2O card || isO2O card =
```
wolfgangwalther, line 483: This seems like an unrelated refactor.

src/PostgREST/SchemaCache.hs
```diff
@@ -542,9 +542,10 @@ addM2MRels tbls rels = rels ++ catMaybes
       then Just $ Relationship t ft (t == ft) (M2M $ Junction jt1 cons1 cons2 (swap <$> cols) (swap <$> fcols)) tblIsView fTblisView
       else Nothing
   | Relationship jt1 t  _ (M2O cons1 cols)  _ tblIsView <- rels
-  , Relationship jt2 ft _ (M2O cons2 fcols) _ fTblisView <- rels
-  , jt1 == jt2
```
wolfgangwalther, line 548: I don't understand the removal of `jt1 == jt2`, yet. Is this another unrelated refactor? Or related to the change here?

mkleczek: It is not needed as we lookup in the hash-map using `jt1` as the key (ie. we changed filtering by equality to hash-map lookup) - that's the crux of this PR.

src/PostgREST/SchemaCache.hs
mkleczek: Indeed.
The previous version caused ugly multiple empty list returns (one for the `else` branch in `if` and another for the pattern match case). Fixing it was too tempting to resist :)

Do you think it is worth doing it in a separate PR?

src/PostgREST/SchemaCache.hs
mkleczek: Looked at the source of `HM.fromListWith op` and indeed: it seems like it calls `newValue op existingValue` so it reverses the list order.

Not sure if we depend on this order anyway but will add reversing the list back just in case.

mkleczek: @wolfgangwalther added `fmap reverse` in line 565 - that should fix this issue.

Can you re-check? (or you could provide me with a quick way to generate this diff)

test/io/test_big_schema.py
mkleczek: Got it - I've taken a look at the test and understand its purpose now.

Testing this by measuring response time is brittle, to say politely.

Without deeper changes into the way things work, I don't see a way to keep this test.

One way would be to keep the timestamp of last schema load, and provide it in a response header. But it wouldn't really allow us to test **waiting** of concurrent requests for schema loading.
I tend to think this kind of properties should be tested using some kind of chaos testing or fuzzing that would only provide statistical guarantees of the system behavior.

Not sure what to do in this PR though. Sensible thing would be to simply comment out this test for now.

WDYT?

src/PostgREST/SchemaCache.hs
wolfgangwalther: No need for a separate PR, but a separate commit would be good - that will make things much easier to understand whenever we look at this again in months or so.

src/PostgREST/SchemaCache.hs
wolfgangwalther: The command I used was:

```
PGRST_DB_SCHEMAS=apflora postgrest-with-postgresql-17 -f test/io/big_schema.sql postgrest-run --dump-schema | jq
```

Pipe this to a file on main, then do the same on your branch, then diff the two.

test/io/test_big_schema.py
wolfgangwalther: I'd certainly not comment it out. Either we can fix it or we should just remove it - we can always add it back. But commented out code is just waste.

I'll defer to @steve-chavez for this, though.

src/PostgREST/SchemaCache.hs
mkleczek: Thanks, @wolfgangwalther - run it after https://github.com/PostgREST/postgrest/pull/4396/commits/39b404e3333134d78c23fdec224aa9838225394a and it fixes the issue - no more differences between `main` and this branch.

src/PostgREST/SchemaCache.hs
mkleczek: @wolfgangwalther done - see https://github.com/PostgREST/postgrest/pull/4396/commits/53bef215a6dee576090be50af9d83559ecbb194e

src/PostgREST/SchemaCache.hs
```diff
@@ -554,13 +555,14 @@ addViewPrimaryKeys tabs keyDeps =
...
       concatMap (\(ViewKeyDependency _ _ _ _ pkCols) -> takeFirstPK pkCols) $
-      filter (\(ViewKeyDependency _ viewQi _ dep _) -> dep == PKDep && viewQi == QualifiedIdentifier sch vw) keyDeps
+      fold $ HM.lookup (PKDep, QualifiedIdentifier sch vw) indexedDeps
     -- In the case of multiple reference to the same PK (see comment for ViewKeyDependency) we take the first reference available.
     -- We assume this to be safe to do, because:
     -- * We don't have any logic that requires the client to name a PK column (compared to the column hints in embedding for FKs),
     --   so we don't need to know about the other references.
     -- * We need to choose a single reference for each column, otherwise we'd output too many columns in location headers etc.
     takeFirstPK = mapMaybe (head . snd)
+    indexedDeps = fmap reverse $ HM.fromListWith (++) $ fmap ((keyDepType &&& keyDepView) &&& pure) keyDeps
```
steve-chavez: The `reverse` is quite mysterious, maybe add a comment that it was done to preserve backwards compat?

test/io/test_big_schema.py
steve-chavez: > Testing this by measuring response time is brittle, to say politely.

Yeah, it was the easiest way to prove that at the time. It's most desirable to have requests waiting than to flood server logs with errors or reply quickly with failure (users just see postgREST as unreliable in these cases).

> One way would be to keep the timestamp of last schema load, and provide it in a response header. But it wouldn't really allow us to test waiting of concurrent requests for schema loading.
I tend to think this kind of properties should be tested using some kind of chaos testing or fuzzing that would only provide statistical guarantees of the system behavior.

Another option would be to inject a sleep like we do here:

https://github.com/PostgREST/postgrest/blob/c1d9728dc8a2469c3701e9af18f1ece25b001b7a/src/PostgREST/SchemaCache.hs#L153-L155

But only for the relationships loading part. Currently the above works for the whole schema cache load.

Then this test could be removed from test/io/test_big_schema.py too

src/PostgREST/SchemaCache.hs
mkleczek: Good idea. Done.

wolfgangwalther: Before we decide to keep it, can we first figure out whether we actually *need* it? Do we rely on the order of these or not?

test/io/test_big_schema.py
mkleczek: @steve-chavez - yeah, that makes the test more robust (no longer dependent on sluggish schema loading performance)

Done.

Introduced `internal-schema-cache-relationship-load-sleep` and implemented delayed loading of relationships (see https://github.com/PostgREST/postgrest/pull/4396/commits/ec31bdd881ff6b69dcf3e4f320764cbb1986cf80)

I left the tests in `test_big_schema.py` - moving it to IO tests is probably a good idea but I wouldn't do that as part of this PR - it is getting out of hand anyway.

WDYT?

mkleczek: This was trickier than I thought. Finally I've ended up with three internal configuration properties to introduce delays in various phases of schema cache loading.
See https://github.com/PostgREST/postgrest/pull/4396/commits/f8ff7c8d157c69277dabe181d9735ea5b0a057c1

steve-chavez: @mkleczek I think those are great! :fire: 

How about separating them into another PR? Looks like they can be merged independently. 

I would also suggest prefixing these commits with `test:` instead of `refactor:`
- ec31bdd881ff6b69dcf3e4f320764cbb1986cf80
- f8ff7c8d157c69277dabe181d9735ea5b0a057c1

mkleczek: @steve-chavez done: #4411

src/PostgREST/SchemaCache.hs
steve-chavez: > can we first figure out whether we actually need it?

I don't think we can figure that out. Overall users have been sensitive to OpenAPI changes, so IMO if we can easily maintain backwards compat we should do it.

@wolfgangwalther If you recall https://github.com/PostgREST/postgrest/pull/1701, I tried to remove this legacy hack we have (IIRC, it's incorrect since it doesn't consider composite keys)

https://github.com/PostgREST/postgrest/blob/c1d9728dc8a2469c3701e9af18f1ece25b001b7a/src/PostgREST/Response/OpenAPI.hs#L129-L133

But that broke `vue-postgrest` so we kept it.

wolfgangwalther: I'm pretty sure it does consider composite keys.

But this brings up an interesting question: Does this cause any observable OpenAPI output differences? I have not tested that. I have only tested the schema dump, which is just an internal representation. So my question was not really targeting whether this would change anything in the openapi output, but whether it would change anything *internal*.

But yes, testing the openapi output for differences would be a great idea as well.

mkleczek: I would suggest we leave it like this in this PR. Let's not make perfect the enemy of the good. This PR brings significant performance gains and I think it is worth merging it and possibly taking care of removing `fmap reverse` here and updating the tests appropriately in the future.

laurenceisla: > But yes, testing the openapi output for differences would be a great idea as well.

I compared it against the `main` branch for both `follow-privileges` and `ignore-privileges` settings. It returns the same OpenAPI output in both branches. What's funny it's that it even returns the same output when the `fmap reverse` is removed.

> and possibly taking care of removing `fmap reverse` here and updating the tests appropriately in the future.

So I agree here. Having the same internal schema should be enough and no queries were modified either, so it's not likely that anything else would change (maybe add a TODO so we don't forget to check in the future?).

mkleczek:
Hmm... I've rebased the PR on top of `main` and one test in `postgrest-test-memory` now fails. Got no idea what's going on.

Locally:
- almost all tests in `postgrest-test-memory` fail for me on `main` ( and also on `main` version before #4411 )
- 2 tests in `postgrest-test-memory` fail on this branch

CI reported all green on pre-rebase version of this PR.

Looks like there is some flakiness in `postgrest-test-memory`. @steve-chavez are you able to provide any insights?

steve-chavez:
> Looks like there is some flakiness in postgrest-test-memory. @steve-chavez are you able to provide any insights?

I restarted the CI job and it passed, it's flakiness. Although I've never had all memory tests failing locally.

src/PostgREST/SchemaCache.hs
wolfgangwalther: I'm not sure why we're making things more complicated than they need to. I spent a few minutes looking at the code and I am quite confident now, that we do **not** depend on the order of the pk columns anywhere.

We do an "equality check" in `Plan.hs`, but we convert the list to Set before, so that should be fine.

We use these columns for `ON CONFLICT` clauses, but the order of columns doesn't matter there.

We use them for `RETURNING` - same, order doesn't really matter. (unless we consider changes in ordering of the JSON response a breaking change, but I highly doubt we do)

The `Location` header URL will also change the order of its query params, I think, but that's fine, too.

So I'd say just remove the `reverse` thing. The diff in internal representation is explained and fine. We don't need to keep it together with a comment that indicates that this prevents a breaking change, when there is none.

mkleczek: I've removed commits introducing `fmap reverse` given @wolfgangwalther analysis, @steve-chavez thumbs up and @laurenceisla tests of OpenAPI output.

src/PostgREST/SchemaCache.hs
mkleczek: After some back and forth in the discussion thread below (https://github.com/PostgREST/postgrest/pull/4396#discussion_r2434090391) `fmap reverse` was removed.

mkleczek:
> Pending the changelog entry

Done

steve-chavez:
Added a test in https://github.com/PostgREST/postgrest/pull/4413

Issue #4134, opened 2025-06-11
JWT with aud claim should be rejected if jwt-aud is not set


mkleczek:
Currently:

> If [jwt-aud](https://docs.postgrest.org/en/v13/references/configuration.html#jwt-aud) is not set (the default), PostgREST identifies with all audiences and allows the JWT for any aud claim.

On the other hand RFC 7519 says:

> If the principal processing the claim does not identify itself with a value in the "aud" claim when this claim is present, then the JWT MUST be rejected.

While RFC wording is vague I am convinced we MUST reject JWTs containing aud claim if jwt-aud is empty (ie. empty jwt-aud means we only accept JWTs without aud claim).

My thinking is: we should stop thinking about "protecting the system" and start thinking about "protecting user data processed by the system". In this context, the threat aud claim is supposed to protect against, is that an attacker steals a JWT and tries to use it to access user's data in unrelated context. Empty jwt-aud right now is hence a security hole. We should be secure by default so misconfiguration (ie. missing jwt-aud) should not lead to security breach. If an administrator wants to accept any aud claim then it should be configured explicitly.

The simplest solution I can think of is to make jwt-aud a regular expression with \b\B (ie. reject anything) as a default value. It can be set explicitly to .* to match anything and also to aud1|aud2|aud3 to address https://github.com/PostgREST/postgrest/issues/2099

This was first discussed in https://github.com/PostgREST/postgrest/issues/4132 but I think it is worth creating a separate issue for this.

wolfgangwalther:
I think this is:
- a feature request for the ability to set `jwt-aud` to a value that means "don't identify with *any* aud value"
- a discussion about the default value `jwt-aud`: all or nothing.

I don't think this is a bug, especially because it's documented this way and *does not violate the RFC*.

mkleczek:
I've marked is as bug as in my opinion this is actually a security vulnerability (and something that _can_ be considered as violation of RFC spec).

wolfgangwalther:
If we were to backport this to a patch release, I'm sure we have the next wave of broken setups coming in. So this must be considered a breaking change for a major version, not a bug fix. Even if it's for security.

robertsosinski:
Chiming in, I would be one of those people submitting a problem if this were to change.

My need is being able to configure multiple auds, as per issue #2099, and I've "solved" this by not specifying an aud in the PostgREST config and checking the JWT aud via a `pre_request` function. 

[...]

PR #4397, opened 2025-10-12
change: Make jwt-aud config value a regular expression

### LINKED ISSUE #2099: Allowing multiple aud values in `jwt-aud` configuration
### Environment

* PostgreSQL version: 13.1
* PostgREST version: v9.0.0 docker postgrest/postgrest:v9.0.0
* Operating system: Windows 10, Running PostgREST/PostgreSQL via Docker

### Description of issue

I'm using a service to generate JWTs for my PostgREST app, and they may come with 2 different `"aud"` values. I've tried separating multiple aud strings in the PostgREST configuration setting (including spaces and commas), but it looks like it only allows one aud to be defined at the moment.

Being able to define multiple auds in the `jwt-aud` configuration would be great.

Thanks!

mkleczek:
**DISCLAIMER:**

This change adds flexibility to aud claim validation. jwt-aud configuration property can now be specified as a regular expression. For example, it is now possible to
* configure multiple acceptable aud values with '|' regex operator, eg: 'audience1|audience2|audience3'
* accept any audience from a particular domain, eg: 'https://[a-z0-9]*\\.example\\.com'

Resolves #2099

- [x] Update documentation
- [x] Update changelog

Subsequent #4419 changes default value of jwt-aud to address #4134

### DIFF
--- a/nix/tools/generate_targets.py
+++ b/nix/tools/generate_targets.py
@@ -33,6 +33,7 @@ def generate_jwt(now: int, exp_inc: Optional[int], is_hs: bool) -> str:
     payload = {
         "sub": f"user_{random.getrandbits(32)}",
         "iat": now,
+        "aud": "veryveryveryveryverylonglonglonglonglongaudience",
         "role": "postgrest_test_author",
     }
 
--- a/nix/tools/loadtest.nix
+++ b/nix/tools/loadtest.nix
@@ -63,6 +63,7 @@ let
         # load test works across branches
         # TODO clean once PGRST_JWT_CACHE_MAX_ENTRIES merged and released
         export PGRST_JWT_CACHE_MAX_LIFETIME="86400"
+        export PGRST_JWT_AUD="audience|([a-z]er[a-z])*(long)*aud..nce"
 
         mkdir -p "$(dirname "$_arg_output")"
         abs_output="$(realpath "$_arg_output")"


steve-chavez:
@mkleczek Could you add a summary of the PR? From reading the issues I'm not sure what's the final design here. This would make it easier to review.

Also, is the breaking change avoidable by changing the default configuration?

mkleczek:
> @mkleczek Could you add a summary of the PR? From reading the issues I'm not sure what's the final design here. This would make it easier to review.

Done.

> Also, is the breaking change avoidable by changing the default configuration?

Yes.

I've decided to set default `jwt-aud` to `.*` (accept anything) for now. That means the change no longer addresses #4134 but is not breaking.

docs/references/configuration.rst
```diff
@@ -596,14 +596,14 @@ jwt-aud
 -------
 
   =============== =================================
-  **Type**        String
+  **Type**        String (must be a valid regular expression)
   **Default**     `n/a`
```
laurenceisla:
```suggestion
  **Default**     .*
```

Perhaps also mention in the description something like: "For example, the default value `.*` matches any `aud` claim."?

steve-chavez:
> I've decided to set default jwt-aud to .* (accept anything) for now. That means the change no longer addresses https://github.com/PostgREST/postgrest/issues/4134 but is not breaking.

@mkleczek Just to clarify, then this PR would not be a `change:` (as per the title), just a new feature right?

docs/references/auth.rst
```diff
@@ -193,13 +193,17 @@ PostgREST has built-in validation of the `JWT audience claim <https://datatracke
...
-- If :ref:`jwt-aud` is set to a specific audience, PostgREST will check if this audience is present in the ``aud`` claim:
+- If :ref:`jwt-aud` is set, PostgREST will treat it as a regular expression and check if it matches the ``aud`` claim:
 
   + If the ``aud`` value is a JSON string, it will match it to the :ref:`jwt-aud`.
   + If the ``aud`` value is a JSON array of strings, it will search every element for a match.
   + If the match fails or if the ``aud`` value is not a string or array of strings, then the token will be rejected with a :ref:`401 Unauthorized <pgrst303>` error.
   + If the ``aud`` key **is not present** or if its value is ``null`` or ``[]``, PostgREST will interpret this token as allowed for all audiences and will complete the request.
 
+Examples:
+- To make PostgREST accept ``aud`` claim value from a set ``audience1``, ``audience2``, ``otheraudience``, :ref:`jwt-aud` claim should be set to ``audience1|audience2|otheraudience``.
```
steve-chavez, line 204: Just to wrap my head around this.

By the [docs](https://docs.postgrest.org/en/v13/references/auth.html#jwt-aud-claim-validation), right now we support:

> If the aud value is a JSON array of strings, it will search every element for a match.

But that's just the JWT and not the `jwt-aud` config.

So with this change now `jwt-aud` can have a list of audiences (using or expression), and the JWT can too specify a list of audiences (using JSON array).

docs/references/auth.rst
```diff
@@ -193,13 +193,17 @@ PostgREST has built-in validation of the `JWT audience claim <https://datatracke
...
+- To make PostgREST accept ``aud`` claim value matching any ``https`` URI pointing to a host in ``example.com`` domain, :ref:`jwt-aud` claim should be set to ``https://[a-zA-Z0-9_]*\.example\.com``.
```
steve-chavez, line 205: Is [ReDoS](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS) a possibility? 

What would be the performance impact of this new feature? Does the JWT cache help here?

mkleczek: > Is [ReDoS](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS) a possibility?
> 

We validate `aud` only **after** JWT authentication (ie. we verify JWT signature first). So it is only possible if an attacker can issue tokens.
Short answer: IMHO no 

> What would be the performance impact of this new feature? Does the JWT cache help here?

Performance impact needs to be verified.

Right now we do not cache claims validation results so JWT won't help.
We can change it but that will require reloading JWT cache not only when`jwt-secret` changes but also when `jwt-aud` is modified. I would postpone this until we are sure regex matching really affects performance.

There is also another potential performance related issue: 

https://github.com/PostgREST/postgrest/pull/4397/commits/60c8a981334042e52885572efb06a3398db1c79e introduces syntactic validation of `aud` claim. Before, as implemented by @taimoorzaeem in #4140, we didn't really check if `aud` claim is a valid `StringOrURI` - we only verified that `jwt-aud` config is syntactically valid. So:
* we did not validate `aud` claims syntactically when `jwt-aud` was not set
* we returned wrong error message when `jwt-aud` was set and `aud` claim was invalid URI: instead of "aud syntax error" we returned "JWT not in audience" (that's disputable as both are valid rejection reasons)
* in case of `jwt-aud` config being a regular expression we cannot really check if it is a valid StringOrURI anymore (the main reason to implement it in this PR)

Checking `aud` claim syntactically is additional work to perform upon every request so will affect performance (we don't know if noticeably).

I've implemented it in a separate commit so that we can easily get rid of it (as nothing depends on it). Not sure if syntactical check of `aud` claim is that important anyway. OTOH caching `aud` claims validation in JWT cache would help with this case as well.

docs/references/auth.rst
mkleczek: > Just to wrap my head around this.
> 
> By the [docs](https://docs.postgrest.org/en/v13/references/auth.html#jwt-aud-claim-validation), right now we support:
> 
> > If the aud value is a JSON array of strings, it will search every element for a match.
> 
> But that's just the JWT and not the `jwt-aud` config.
> 

Yes.
(I think this sentence was not changed)

> So with this change now `jwt-aud` can have a list of audiences (using or expression), and the JWT can too specify a list of audiences (using JSON array).

Yes.
That's what #2099 is about.

docs/references/auth.rst
steve-chavez: > There is also another potential performance related issue:
> https://github.com/PostgREST/postgrest/commit/60c8a981334042e52885572efb06a3398db1c79e introduces syntactic validation of aud claim.
(Note: the above was released on [v13.0.4](https://github.com/PostgREST/postgrest/releases/tag/v13.0.4))

@mkleczek Q: The performance hit would only happen if `jwt-aud` is set right? And with this new PR, `jwt-aud` will always be set hence the perf impact will happen for every installation (the `jwt-aud='.*'` regex check will always be done).

> Performance impact needs to be verified.
> Checking aud claim syntactically is additional work to perform upon every request so will affect performance (we don't know if noticeably).

So to check the above we would need new loadtests with the `jwt-aud` enabled right?

mkleczek: > > There is also another potential performance related issue:
> > [60c8a98](https://github.com/PostgREST/postgrest/commit/60c8a981334042e52885572efb06a3398db1c79e) introduces syntactic validation of aud claim.
> > (Note: the above was released on [v13.0.4](https://github.com/PostgREST/postgrest/releases/tag/v13.0.4))
> 
> @mkleczek Q: The performance hit would only happen if `jwt-aud` is set right? And with this new PR, `jwt-aud` will always be set hence the perf impact will happen for every installation (the `jwt-aud='.*'` regex check will always be done).
> 

No. Currently (ie. in `main`) we don't check `aud` claim syntactically at all. We only syntactically check `jwt-aud` config. But we use equality check to validate `aud` claim hence we don't accept syntactically invalid `aud`s **if `jwt-aud` is set** but accept invalid `aud`s if `jwt-aud` is not set.

This change was supposed to change that and validate `aud`s syntactically **always**, before even checking them against `jwt-aud` config.

Nevertheless, I removed commit introducing this check for now.

> > Performance impact needs to be verified.
> > Checking aud claim syntactically is additional work to perform upon every request so will affect performance (we don't know if noticeably).
> 
> So to check the above we would need new loadtests with the `jwt-aud` enabled right?

Nope, see above.

steve-chavez:
> It would be great if it could land in 14 as it is a breaking change so requires a major release.

@mkleczek I think it'd be better to release v14 and leave this one for v16 (shouldn't take long anyway, towards EOY). That way we can see what's the impact of the JWT cache in isolation (since it's enabled by default now) and reduce the amount of breaking changes we have in one release. WDYT?

mkleczek:
> > It would be great if it could land in 14 as it is a breaking change so requires a major release.
> 
> @mkleczek I think it'd be better to release v14 and leave this one for v16 (shouldn't take long anyway, towards EOY). That way we can see what's the impact of the JWT cache in isolation (since it's enabled by default now) and reduce the amount of breaking changes we have in one release. WDYT?

I hoped to have #2099 sooner as we have a need for it, but yeah, maybe it's better not to push it.

mkleczek:
@steve-chavez  @laurenceisla 

I've reworked this PR and I think it is now in quite a good shape.

The prerequisite, the most important and most complex commit is the first one: https://github.com/PostgREST/postgrest/pull/4397/commits/0138fdf4e648205cb5e851cdd0552024f3202449
I've provided quite a bit of information about it in the commit message.

Second commit is comparatively small change in Config module.

Third commit adds `aud` validation to our load tests.


I've spend considerable amount of time testing various caching variants (it was quite easy thanks to the changes in the first commit). Decided to change JwtCache (comparing to `main`) so that it caches properly signed parsed tokens after successful `aud` claim validation.
Thanks to it there is no impact on performance when JWT cache is turned on.

docs/references/configuration.rst
mkleczek: Done

steve-chavez:
@mkleczek Sorry for the late reply.

I've just noted that after d71cb81b2dc833e1418532adeaa002d8e4eb3af3 all JWT loadtests are failing with `401 Unauthorized`.

> The prerequisite, the most important and most complex commit is the first one: https://github.com/PostgREST/postgrest/commit/0138fdf4e648205cb5e851cdd0552024f3202449
I've provided quite a bit of information about it in the commit message.

As first step, maybe you can move that refactor commit in another PR? It looks quite involved and would like to review it separately.

mkleczek:
> @mkleczek Sorry for the late reply.
> 
> I've just noted that after [d71cb81](https://github.com/PostgREST/postgrest/commit/d71cb81b2dc833e1418532adeaa002d8e4eb3af3) all JWT loadtests are failing with `401 Unauthorized`.
> 
> > The prerequisite, the most important and most complex commit is the first one: [0138fdf](https://github.com/PostgREST/postgrest/commit/0138fdf4e648205cb5e851cdd0552024f3202449)
> > I've provided quite a bit of information about it in the commit message.
> 
> As first step, maybe you can move that refactor commit in another PR? It looks quite involved and would like to review it separately.

Yeah, after some more thinking I'm still not happy with it.
What's missing is caching everything that can be safely cached (so signature verification, aud validation and role lookup) while validating everything else (ie. time sensitive claims) after cache lookup.

Marked this PR as draft and will open a new one for refactoring.

Issue #4295, opened 2025-08-31
CI: automate backports


taimoorzaeem:
Nixpkgs uses [https://github.com/korthout/backport-action](https://github.com/korthout/backport-action) to automate backports via just a label. This works well. I think we should do the same.

It would make things easier:
- This action creates a PR, so the diff can be reviewed before merge.
- The changelog file has append mode enabled for merging, so there won't be a merge conflict, which would prevent the action from doing something.

From my experience, almost all backports apply cleanly and the changelog is in the right spot (assuming we had forward ported the changelog version update and backported all fixes in order).

_Originally posted by @wolfgangwalther in https://github.com/PostgREST/postgrest/issues/4281#issuecomment-3239528138_

wolfgangwalther:
This is purely GitHub Actions, has nothing to do with Nix.

PR #4398, opened 2025-10-13
fix: loading utf-8 config files with ascii locale set

### LINKED ISSUE #4386: docs: warn about locale
We should warn about this on the config docs:

> Thank you both for pointing me to the solution, you were absolutely right:
> 
> On my Debian VM, `env | grep LC` was returning:
> ```
> LC_CTYPE=C
> LC_ALL=C
> ```
> 
> In order to fix it, I did `sudo update-locale LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8` and now the problem is gone 👍 
> 
> FYI, `env | grep LC` is now returning:
> ```
> LC_CTYPE=C
> LC_ALL=en_US.UTF-8
> ```
> 
> So it seems the very important setting is the `LC_ALL` (not the `LC_CTYPE`) 

 _Originally posted in [#4379](https://github.com/PostgREST/postgrest/issues/4379#issuecomment-3370161179)_

---

Also, maybe there is a way to avoid depending on the `LC_*` settings?

taimoorzaeem:
Updates configurator-pg to version 0.2.11. This resolves #4386.


wolfgangwalther:
You should update the constraints in `postgrest.cabal` to make *sure* this is always build with that new version.

wolfgangwalther:
Let's add a changelog entry for #4386 and squash all commits in here to a `fix: ...`.

taimoorzaeem:
- This fix seems like a good opportunity to work on #4295.

@wolfgangwalther Please don't backport this fix yet. :pray:

wolfgangwalther:
Are you going to create the backport, @taimoorzaeem?

taimoorzaeem:
> Are you going to create the backport, @taimoorzaeem?

Yes sure. :+1:

PR #4409, opened 2025-10-15
ci: automate backports using backport action

### LINKED ISSUE #4295: CI: automate backports

taimoorzaeem:
Creates a workflow using https://github.com/korthout/backport-action. This allows creating a backport PR by simply labeling the merged PR with a `backport <version>` label.

Closes #4295.

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,52 @@
+name: Backport
+
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-24.04
+    # It triggers only when PR is already merged on either:
+    #
+    # - The merge event itself (action != labeled) or
+    # - A label event with the right label (backport ...).
+    #
+    # The result will be that we can add the label before or after merge,
+    # but the workflow will only run once the PR had been merged.
+    if: >
+      github.event.pull_request.merged &&
+      (
+        github.event.action != 'labeled' ||
+        startsWith(github.event.label.name, 'backport')
+      )
+    steps:
+
+      # This actions creates the github token using the postgrest app secrets
+      - name: Create Github App Token
+        id: app-token
+        uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
+        with:
+          app-id: ${{ secrets.POSTGREST_CI_APP_ID }}
+          private-key: ${{ secrets.POSTGREST_CI_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
+          permission-workflows: write # required when backporting CI changes
+
+      # This is required for backport action to cherry-pick the PR
+      - name: Fetch PR ref
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          token: ${{ steps.app-token.outputs.token }}
+
+      # Backport action that creates the PR with given settings
+      - name: Create backport PR
+        uses: korthout/backport-action@d07416681cab29bf2661702f925f020aaa962997 # v3.4.1
+        with:
+          github_token: ${{ steps.app-token.outputs.token }}
+          pull_description: 'Backport for #${pull_number}.'
+          pull_title: '${target_branch}: ${pull_title}'


taimoorzaeem:
I have tested this in my fork and it works as expected. See https://github.com/taimoorzaeem/postgrest/pull/19

EDIT: Updated this, please checkout again.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,54 @@
...
+            echo "No branch specified in comment"
+            exit 1
+          fi
+
+          # Validate branch exist
+          git fetch --all --quiet
+          if ! git show-ref --verify --quiet "refs/remotes/origin/$BRANCH"; then
+            echo "Invalid branch specified, $BRANCH branch does not exist."
+            exit 1
+          fi
```
wolfgangwalther: Why all this special logic when `korthout/backport-action` can already do it via labels natively?

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,54 @@
+name: Backport
+
+on:
+  pull_request_target:
+    types: [closed]
+  issue_comment:
+    types: [created]
+permissions:
+  contents: write
+  pull-requests: write
```
wolfgangwalther: Using the regular github token won't be enough, because PRs pushed with this *will not run CI*. That's a github "feature" to prevent infinite recursion.

To make this work, we will need to:
- Create a GitHub App for this repo.
- Put the secrets into GHA variables / secrets.
- Create a token for it in the job.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,54 @@
+name: Backport
+
+on:
+  pull_request_target:
+    types: [closed]
```
wolfgangwalther: Whis is this run when closing the PR, when the actual trigger is on a comment? This event seems unused right now. However, with my suggestion of using labels, this would be the right trigger to use.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,54 @@
...
+  issue_comment:
+    types: [created]
+permissions:
+  contents: write
+  pull-requests: write
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
```
wolfgangwalther: Please use the same pattern as in all the other workflows in the repo - i.e. specify the exact, not latest. This is to avoid surprising failures when the -latest tag changes.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,54 @@
...
+      (
+        github.event_name == 'issue_comment' &&
+        github.event.issue.pull_request &&
+        contains(
+            fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'),
+            github.event.comment.author_association
+          )
+      )
+    steps:
+      - uses: actions/checkout@v4
```
wolfgangwalther: Please use the same pattern as in all the other workflows. Pin to a commit, put the version number in a comment behind. This will be used by renovate to update automatically.

(also v5 is latest here)

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+          # Validate branch exist
+          git fetch --all --quiet
+          if ! git show-ref --verify --quiet "refs/remotes/origin/$BRANCH"; then
+            echo "Invalid branch specified, $BRANCH branch does not exist."
+            exit 1
+          fi
+
+      - name: Create backport PR
+        uses: korthout/backport-action@v3
```
wolfgangwalther: Same as with the other action - pin it properly!

.github/workflows/backport.yaml
develop7: to be fair, wouldn't renovate bot offer to pin those properly first thing after this PR is merged?

.github/workflows/backport.yaml
taimoorzaeem: > Using the regular github token won't be enough, because PRs pushed with this _will not run CI_. That's a github "feature" to prevent infinite recursion.

Ahh, I see now. Thanks for the info, I didn't notice that it doesn't run the CI this way.

> Create a GitHub App for this repo. 
> Put the secrets into GHA variables / secrets.
> Create a token for it in the job.

Yeah, this looks like the way to go. Now, although I _can_ create a github app and we could install it on postgrest, but I think it would be _best_ if the members of this organization create the app. That way everyone from this organization would be able to edit or configure it.

wolfgangwalther: Yes. You can create your own app in the fork and test things that way, I will create the respective app in the PostgREST org, once we know all the details.

.github/workflows/backport.yaml
taimoorzaeem: > to be fair, wouldn't renovate bot offer to pin those properly first thing after this PR is merged?

Yes, I think it would, but I can pin it too and we can get it right the first time :innocent:.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,42 @@
...
+    if: github.event.pull_request.merged
+    steps:
+
+      # This actions creates the github token using the postgrest-bot secrets
+      - name: Create Github App Token
+        id: app-token
+        uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
+        with:
+          app-id: ${{ secrets.POSTGREST_TEST_BOT_APP_ID }}
+          private-key: ${{ secrets.POSTGREST_TEST_BOT_PRIVATE_KEY }}
```
taimoorzaeem: I used these for my testing, which should be updated to the new app secrets' names.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,42 @@
...
+          # Fetching full history might take a while, so 100 seems fine for now.
+          # It is unlikely we would backport a PR further than that
+          fetch-depth: 100
+
+      # Backport actions that creates the PR with given configurations
+      - name: Create backport PR
+        uses: korthout/backport-action@d07416681cab29bf2661702f925f020aaa962997 # v3.4.1
+        with:
+          git_committer_name: 'postgrest-test-bot[bot]'
+          git_committer_email: 'postgrest-test-bot[bot]@users.noreply.github.com'
```
taimoorzaeem: Also, these need to be updated to the new app's name.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,42 @@
+name: Backport
+
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+permissions:
+  contents: write
+  pull-requests: write
```
wolfgangwalther: The GITHUB_TOKEN itself doesn't need any permissions anymore.

```suggestion
permissions: {}
```

.github/workflows/backport.yaml
wolfgangwalther: Good practice to be specific about which permissions you need:

```suggestion
        uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
        with:
          app-id: ${{ secrets.POSTGREST_TEST_BOT_APP_ID }}
          private-key: ${{ secrets.POSTGREST_TEST_BOT_PRIVATE_KEY }}
          permission-contents: write
          permission-pull-requests: write
          permission-workflows: write
```

The workflows privilege is required to be able to backport changes to CI itself.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,42 @@
...
+          app-id: ${{ secrets.POSTGREST_TEST_BOT_APP_ID }}
+          private-key: ${{ secrets.POSTGREST_TEST_BOT_PRIVATE_KEY }}
+
+      # This is required for backport action to cherry-pick the PR
+      - name: Fetch git history
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          # Fetching full history might take a while, so 100 seems fine for now.
+          # It is unlikely we would backport a PR further than that
+          fetch-depth: 100
```
wolfgangwalther: In Nixpkgs we're just using this:

```suggestion
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
        with:
          ref: ${{ github.event.pull_request.head.sha }}
```

The fetch-depth doesn't seem required.

.github/workflows/backport.yaml
wolfgangwalther: We don't need to set these at all, they will be configured via the app itself.


```suggestion
```

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-24.04
+    if: github.event.pull_request.merged
```
wolfgangwalther: To avoid running the whole workflow all the time, until the backport action sees it doesn't need to do anything, you can do:


```suggestion
    if: github.event.pull_request.merged && (github.event.action != 'labeled' || startsWith(github.event.label.name, 'backport'))
```

taimoorzaeem: I think it should be this, implying that it only runs when labeled a merged PR with a label that starts with `backport`, no?

```yaml
github.event.pull_request.merged &&
github.event.action == 'labeled' &&
startsWith(github.event.label.name, 'backport')
```

.github/workflows/backport.yaml
taimoorzaeem: Very clever! Perfect. :tada:

.github/workflows/backport.yaml
wolfgangwalther: No, the condition I proposed is correct, I copied it from the battle-tested Nixpkgs workflow.

It triggers only when the PR is already merged on *either*:
- The merge event itself (action != labeled) or
- A label event with the right label.

It's a filter to remove label events for other labels.

The result will be that you can add the label before or after merge, but the workflow will only run once the PR had been merged.

PR #4413, opened 2025-10-21
test: schema cache load max duration

steve-chavez:
Test to ensure no regression happens after https://github.com/PostgREST/postgrest/pull/4396

### DIFF
--- a/test/io/test_big_schema.py
+++ b/test/io/test_big_schema.py
@@ -1,5 +1,7 @@
 "IO tests for PostgREST started on the big schema."
 
+import re
+
 import pytest
 
 from postgrest import run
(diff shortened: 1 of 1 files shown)


mkleczek:
@steve-chavez I think it might be a good idea to move `test_requests_with_resource_embedding_wait_for_schema_cache_reload ` and `test_requests_without_resource_embedding_wait_for_schema_cache_reload ` to IO tests (as discussed earlier). After #4411 they no longer require big schema.

steve-chavez:
@mkleczek Opened https://github.com/PostgREST/postgrest/issues/4417 for that. I looked at doing it now but it was more involved that I thought. (we need new sql fixtures on test/io/fixtures.sql, there's no example for resource embedding there)

PR #4414, opened 2025-10-21
refactor: Encapsulate aud config

mkleczek:
This change is an initial step to change JWT aud configuration to regular expression (#4397). Exporting function
`audMatchesCfg :: AppConfig -> Text -> Bool`
from Config module allows changing the way how JWT aud is configured to be isolated and not affect code in Auth.JWT


mkleczek:
@laurenceisla I've seen you started reviewing #4397. I marked it as draft and would like to split it into smaller and easier to review PRs, that would also be less controversial.

laurenceisla:
> I marked it as draft and would like to split it into smaller and easier to review PRs, that would also be less controversial.

Perfect, thanks for the heads-up.

PR #4418, opened 2025-10-22
remove(feat): fail schema cache lookup with invalid db-schemas config

### LINKED ISSUE #4364: Auto-removal of schemas in `db-schemas` once `DROP SCHEMA ..` is done
## Problem

With the change on https://github.com/PostgREST/postgrest/pull/3644, particularly https://github.com/PostgREST/postgrest/pull/3644/commits/b6863f19c65146dafc38d215947a31379589a9d6, schema-based multi-tenancy becomes more complicated as the user has to remember to remove schemas from `db-schemas` once `DROP SCHEMA` is done, otherwise PostgREST will fail.

## Proposal 1

Automate schema removal by extending the [Finer-Grained Event Trigger](https://docs.postgrest.org/en/v13/references/schema_cache.html#finer-grained-event-trigger).

## Proposal 2

Rollback to becoming lenient and don't fail on nonexistent schemas on db-schemas.

steve-chavez:
Closes #4364 by reverting 86c3257f5496688196be33a465d6cacaf89f62cf

As discussed on #4364, 86c3257f5496688196be33a465d6cacaf89f62cf generates several problems:

1. We diverge from database as source of truth because once a schema is dropped, it's not automatically removed from `db-schemas` and then postgREST fails. Reducing its reliability.
2. We diverge from PostgreSQL `search_path` lenient behavior:
```sql
set search_path to anything, here, even, nonexistent; -- this doesn't fail
```
3. Schema-based multitenancy becomes more complicated because of 1.

There are no noticeable gains from 86c3257f5496688196be33a465d6cacaf89f62cf:

- No gains in performance for the schema cache queries.
- No user report about it ever being a problem. In fact the leniency was a feature.

## Implementation details

Essentially it removes the `$1::regnamespace[]` casting for `db-schemas` that was introduced on 86c3257f5496688196be33a465d6cacaf89f62cf

### DIFF
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -595,8 +595,14 @@ handlerF rout = \case
   NoAgg                      -> "''::text"
 
 schemaDescription :: Text -> SQL.Snippet
-schemaDescription schema =
-  "SELECT pg_catalog.obj_description(" <> encoded <> "::regnamespace, 'pg_namespace')"
+schemaDescription schema = SQL.sql (encodeUtf8 [trimming|
+  SELECT
+    description
+  FROM
+    pg_namespace n
+    left join pg_description d on d.objoid = n.oid
+  WHERE
+    n.nspname = |]) <> encoded
   where
     encoded = SQL.encoderAndParam (HE.nonNullable HE.unknown) $ encodeUtf8 schema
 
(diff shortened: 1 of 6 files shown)


wolfgangwalther:
I don't have the energy to fight about this, so I won't block you, but I also won't approve a change that I think is fundamentally wrong.

> 1. We diverge from database as source of truth because once a schema is dropped, it's not automatically removed from `db-schemas` and then postgREST fails

This is just backwards. With this change, the schema cache becomes the source of truth. Schema still in the schema cache? You get a 200 OK. That's **not** the database as the source of truth.

steve-chavez:
>  Schema still in the schema cache? You get a 200 OK. That's not the database as the source of truth.

In this case we'll get an error coming from the database, should be a 404 according to https://docs.postgrest.org/en/v13/references/errors.html#http-status-codes.

steve-chavez:
That is if you request a table or function. If you request a schema having `openapi-mode=follow-privileges`, it should be an error.

I haven't tried for `openapi-mode=ignore-privileges`, but it's possible that a 200 OK happens in that case. If the schema existed before and then was dropped.

wolfgangwalther:
In https://github.com/PostgREST/postgrest/issues/4364#issuecomment-4806032307 it was mentioned that "this ship has sailed", so let's close this. We're finally on track to find agreement on a way forward in that issue as well.

PR #4419, opened 2025-10-22
Reject non-empty aud when jwt-aud is not configured

### LINKED ISSUE #4134: JWT with aud claim should be rejected if jwt-aud is not set

mkleczek:
**DISCLAIMER:**

Fixes #4134 

- [ ] Update documentation
- [ ] Update changelog

### DIFF
--- a/docs/references/auth.rst
+++ b/docs/references/auth.rst
@@ -203,7 +203,7 @@ It works this way:
 Examples:
 - To make PostgREST accept ``aud`` claim value from a set ``audience1``, ``audience2``, ``otheraudience``, :ref:`jwt-aud` claim should be set to ``audience1|audience2|otheraudience``.
 - To make PostgREST accept ``aud`` claim value matching any ``https`` URI pointing to a host in ``example.com`` domain, :ref:`jwt-aud` claim should be set to ``https://[a-zA-Z0-9_]*\.example\.com``.
-- To make PostgREST accept any ``aud`` claim value , :ref:`jwt-aud` claim should be set to ``.*`` (which is the default).
+- To make PostgREST accept any ``aud`` claim value , :ref:`jwt-aud` claim should be set to ``.*``.
 
 .. _jwt_caching:
 
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -90,7 +90,7 @@ parseCfgAud = fmap CfgAud . (fmap . ParsedValue . Just <*> parseRegex)
     bounded = ("\\`(" <>) . (<> "\\')")
 
 defaultCfgAud :: CfgAud
-defaultCfgAud = CfgAud $ ParsedValue Nothing $ R.makeRegex (".*"::Text)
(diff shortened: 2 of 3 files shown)


steve-chavez:
> The simplest solution I can think of is to make jwt-aud a regular expression with \b\B (ie. reject anything) as a default value. It can be set explicitly to .* to match anything and also to aud1|aud2|aud3 to address https://github.com/PostgREST/postgrest/issues/2099

So with this change we would break if JWTs with an `aud` come with the default jwt-aud configuration (empty).

To return to previous behavior (accept any aud), users would have to explicitly set `jwt-aud = '.*` (accept any).

Some observations:

- If v13 is set to `jwt-aud = '.*'`, this will fail?
- We would need a clear error message plus a new error code `PGRSTxxx` for the case where an empty `jwt-aud` fails with a JWT with an aud
  + This way the user can search the `PGRSTxxx` code on the docs, where we can say "to return to previous behavior use `.*`".

> My reasoning is: we should stop thinking about "protecting the system" and start thinking about "protecting user data processed by the system". In this context, the threat aud claim is supposed to protect against, is that an attacker steals a JWT and tries to use it to access user's data in unrelated context. Empty jwt-aud right now is hence a security hole. We should be secure by default so misconfiguration (ie. missing jwt-aud) should not lead to security breach. If an administrator wants to accept any aud claim then it should be configured explicitly.
> https://github.com/PostgREST/postgrest/issues/4134#issue-3136738644

I think the above makes sense, we would need to mention it on the CHANGELOG.

steve-chavez:
@mkleczek Why was this closed? :thinking:

mkleczek:
> @mkleczek Why was this closed? 🤔

Looks like I’ve deleted the source branch by mistake and the PR was automatically closed.

PR #4420, opened 2025-10-22
fix: db-pre-config function failing with pg reserved words

### LINKED ISSUE #4380: Using PGREST_DB_PRE_CONFIG w/out the underlying config function yields a confusing error
### Environment

* PostgreSQL version: postgres:17
* PostgREST version: postgrest/postgrest:v13.0.7
* Operating system: Docker Desktop 4.26.1 on Mac OS 15.1.1

### Description of issue

Describe the behavior you expected vs the actual behavior. Include:

This on startup:
```
pg         | 2025-10-03 21:48:33.517 UTC [51] LOG:  unexpected EOF on client connection with an open transaction
postgrest exited with code 1
postgrest  | 03/Oct/2025:21:48:35 +0000: Listening for database notifications on the "pgrst" channel
postgrest  | 03/Oct/2025:21:48:35 +0000: Successfully connected to PostgreSQL 17.6 (Debian 17.6-2.pgdg13+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
postgrest  | 03/Oct/2025:21:48:35 +0000: Connection Pool initialized with a maximum size of 15 connections
pg         | 2025-10-03 21:48:35.276 UTC [53] ERROR:  syntax error at or near "true" at character 566
pg         | 2025-10-03 21:48:35.276 UTC [53] STATEMENT:  WITH
pg         |    role_setting AS (
pg         |      SELECT setdatabase as database,
pg         |             unnest(setconfig) as setting
pg         |      FROM pg_catalog.pg_db_role_setting
pg         |      WHERE setrole = CURRENT_USER::regrole::oid
pg         |        AND setdatabase IN (0, (SELECT oid FROM pg_catalog.pg_database WHERE datname = CURRENT_CATALOG))
pg         |    ),
pg         |    kv_settings AS (
pg         |      SELECT database,
pg         |             substr(setting, 1, strpos(setting, '=') - 1) as k,
pg         |             substr(setting, strpos(setting, '=') + 1) as v
pg         |      FROM role_setting
pg         |      UNION
pg         |      SELECT
pg         |        null as database,
pg         |        x as k,
pg         |        current_setting(x, true) as v
pg         |      FROM unnest($1) x
pg         |      JOIN true() _ ON TRUE
pg         |    )
pg         |    SELECT DISTINCT ON (key)
pg         |           replace(k, 'pgrst.', '') AS key,
pg         |           v AS value
pg         |    FROM kv_settings
pg         |    WHERE k = ANY($1) AND v IS NOT NULL
pg         |    ORDER BY key, database DESC NULLS LAST;
postgrest  | 03/Oct/2025:21:48:35 +0000: This is probably a bug in PostgREST, please report it at https://github.com/PostgREST/postgrest/issues. {"code":"42601","details":null,"hint":null,"message":"syntax error at or near \"true\""}
postgrest  | 03/Oct/2025:21:48:35 +0000: Failed to query database settings for the config parameters.{"code":"42601","details":null,"hint":null,"message":"syntax error at or near \"true\""}
postgrest  | postgrest: thread killed
```

Discovered by having ChatGPT through in a gratuitous PGREST_DB_PRE_CONFIG into a docker-compose.yaml where it had no business being.

taimoorzaeem:
When `db-pre-config` is accidentally set to a pg reserved word like "true", it fails with a confusing error. The function names should be properly quoted to avoid such errors. This commit resolves the mentioned issue by quoting the pre-config function name.

Closes #4380.

### DIFF
--- a/src/PostgREST/SchemaCache/Identifiers.hs
+++ b/src/PostgREST/SchemaCache/Identifiers.hs
@@ -10,6 +10,7 @@ module PostgREST.SchemaCache.Identifiers
   , dumpQi
   , escapeIdent
   , isAnyElement
+  , quoteQi
   , toQi
   , trimNullChars
   ) where
@@ -40,6 +41,10 @@ isAnyElement y = QualifiedIdentifier "pg_catalog" "anyelement" == y
 dumpQi :: QualifiedIdentifier -> Text
 dumpQi (QualifiedIdentifier s i) =
   (if T.null s then mempty else s <> ".") <> i
+
+quoteQi :: QualifiedIdentifier -> Text
+quoteQi (QualifiedIdentifier s i) =
+  (if T.null s then mempty else escapeIdent s <> ".") <> escapeIdent i
 
 -- TODO: Handle a case where the QI comes like this: "my.fav.schema"."my.identifier"
 -- Right now it only handles the schema.identifier case
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. From versio
 
 ## Unreleased
 
+## Fixed
+
+- Fix `db-pre-config` function failing when function names are pg reserved words by @taimoorzaeem #4380
+
 ## [14.0] - 2025-10-24
 
 ### Added
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -69,7 +69,7 @@ import PostgREST.Config.PgVersion        (PgVersion (..),
 import PostgREST.SchemaCache             (SchemaCache (..),
                                           querySchemaCache,
                                           showSummary)
-import PostgREST.SchemaCache.Identifiers (dumpQi)
+import PostgREST.SchemaCache.Identifiers (quoteQi)
 import PostgREST.Unix                    (createAndBindDomainSocket)
 
 import Data.Streaming.Network (bindPortTCP, bindRandomPortTCP)
@@ -441,7 +441,7 @@ readInDbConfig startingUp appState@AppState{stateObserver=observer} = do
   pgVer <- getPgVersion appState
   dbSettings <-
     if configDbConfig conf then do
-      qDbSettings <- usePool appState (queryDbSettings (dumpQi <$> configDbPreConfig conf) (configDbPreparedStatements conf))
+      qDbSettings <- usePool appState (queryDbSettings (quoteQi <$> configDbPreConfig conf) (configDbPreparedStatements conf))
       case qDbSettings of
         Left e -> do
           observer $ ConfigReadErrorObs e
--- a/test/io/__snapshots__/test_cli/test_schema_cache_snapshot[dbRoutines].yaml
+++ b/test/io/__snapshots__/test_cli/test_schema_cache_snapshot[dbRoutines].yaml
@@ -461,6 +461,23 @@
       pdSchema: public
       pdVolatility: Volatile
 
+- - qiName: 'true'
+    qiSchema: public
+  - - pdDescription: null
+      pdFuncSettings: []
+      pdHasVariadic: false
+      pdName: 'true'
+      pdParams: []
+      pdReturnType:
+        contents:
+          contents:
+            qiName: bool
+            qiSchema: pg_catalog
+          tag: Scalar
+        tag: Single
+      pdSchema: public
+      pdVolatility: Volatile
+
 - - qiName: create_function
     qiSchema: public
   - - pdDescription: null
--- a/test/io/fixtures.sql
+++ b/test/io/fixtures.sql
@@ -256,3 +256,7 @@ select * from projects;
 
 create or replace view infinite_recursion as
 select * from infinite_recursion;
+
+create or replace function "true"() returns boolean as $_$
+  select true;
+$_$ language sql;
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -2046,3 +2046,29 @@ def test_log_listener_connection_errors(defaultenv):
             in line
             for line in output
         )
+
+
+def test_db_pre_config_with_pg_reserved_words(defaultenv):
+    "The db-pre-config should not fail unexpectedly when function name is a postgres reserved word"
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_PRE_CONFIG": "true",  # call true function
+    }
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.post("/rpc/true")
+        assert response.status_code == 200
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_PRE_CONFIG": "select",  # no "select" function in our fixtures, fail gracefully at startup
+    }
+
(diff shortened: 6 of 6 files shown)


src/PostgREST/Config/Database.hs
```diff
@@ -203,3 +204,14 @@ param = HE.param . HE.nonNullable
...
+-- |
+-- >>> quotePreConfFunc "postgrest.pre_config"
+-- "\"postgrest\".\"pre_config\""
+--
+-- >>> quotePreConfFunc "true"
+-- "\"true\""
+quotePreConfFunc :: Text -> Text
+quotePreConfFunc f = T.intercalate "." $ map escapeIdent $ T.split (== '.') f
+    where
+      escapeIdent id = "\"" <> id <> "\""
```
wolfgangwalther: Don't we have a helper function for this somewhere already?

steve-chavez: Maybe https://github.com/PostgREST/postgrest/blob/3db05612ff3a6abb1d509aac0c2bfdd4f00dad44/src/PostgREST/Query/SqlFragment.hs#L166-L167

taimoorzaeem: https://github.com/PostgREST/postgrest/blob/66161104bf09584dd0bef6539b93ba9f0d0a020e/src/PostgREST/Query/SqlFragment.hs#L245-L249

We have this, but it gives a `Snippet` and not `Text`. Also, the `Config/Database.hs` module is currently free of `Query` and `SchemaCache` modules, so let's not pollute it with that as it would make the dependency graph even weirder.

I think long term solution would be to have our own Prelude as discussed in https://github.com/PostgREST/postgrest/issues/1804#issuecomment-3254810274. For now I have handled it by quoting it by a new function `quoteQi`.

steve-chavez (approved):
Tested it manually. Nice work 💯

src/PostgREST/SchemaCache/Identifiers.hs
```diff
@@ -39,6 +40,12 @@ dumpQi :: QualifiedIdentifier -> Text
 dumpQi (QualifiedIdentifier s i) =
   (if T.null s then mempty else s <> ".") <> i
 
+quoteQi :: QualifiedIdentifier -> Text
+quoteQi (QualifiedIdentifier s i) =
+  (if T.null s then mempty else quoteIdent s <> ".") <> quoteIdent i
+  where
+    quoteIdent ident = "\"" <> ident <> "\""
```
steve-chavez: Oh, one sec. This doesn't escape the `"` right? 

---

If you test it like so:

```
$ PGRST_DB_PRE_CONFIG="sel\"ect" postgrest-with-postgresql-15 -f test/spec/fixtures/load.sql postgrest-run
```

It wil give a bad error:

```sql
24/Oct/2025:13:02:55 -0500: This is probably a bug in PostgREST, please report it at https://github.com/PostgREST/postgrest/issues. {"code":"42601","details":null,"hint":null,"message":"unterm
inated quoted identifier at or near \"\"() _ ON TRUE\n)\nSELECT DISTINCT ON (key)\n       replace(k, 'pgrst.', '') AS key,\n       v AS value\nFROM kv_settings\nWHERE k = ANY($1) AND v IS NOT 
NULL\nORDER BY key, database DESC NULLS LAST;\""}
24/Oct/2025:13:02:55 -0500: Failed to query database settings for the config parameters.{"code":"42601","details":null,"hint":null,"message":"unterminated quoted identifier at or near \"\"() _
 ON TRUE\n)\nSELECT DISTINCT ON (key)\n       replace(k, 'pgrst.', '') AS key,\n       v AS value\nFROM kv_settings\nWHERE k = ANY($1) AND v IS NOT NULL\nORDER BY key, database DESC NULLS LAST
;\""}
postgrest: thread killed

```

src/PostgREST/Config/Database.hs
steve-chavez: > I think long term solution would be to have our own Prelude as discussed in https://github.com/PostgREST/postgrest/issues/1804#issuecomment-3254810274. For now I have handled it by quoting it by a new function quoteQi.

Maybe instead we can move 

https://github.com/PostgREST/postgrest/blob/4643eaf7bf9a3633720e619e7eea0ad3b705034a/src/PostgREST/Query/SqlFragment.hs#L166-L167

To `src/PostgREST/SchemaCache/Identifiers.hs`?

taimoorzaeem: > Maybe instead we can move
> [...] 
> To `src/PostgREST/SchemaCache/Identifiers.hs`?

Yes, lemme refactor this. :+1:

src/PostgREST/SchemaCache/Identifiers.hs
taimoorzaeem: Oh, missed that. Fixed now. :+1:

src/PostgREST/SchemaCache/Identifiers.hs
```diff
@@ -41,6 +42,10 @@ dumpQi :: QualifiedIdentifier -> Text
 dumpQi (QualifiedIdentifier s i) =
   (if T.null s then mempty else s <> ".") <> i
 
+quoteQi :: QualifiedIdentifier -> Text
+quoteQi (QualifiedIdentifier s i) =
+  (if T.null s then mempty else escapeIdent s <> ".") <> escapeIdent i
```
steve-chavez, line 47: Just noticed the `dumpQi`  is too similar to `quoteQi`.

Shouldn't we stay with `quoteQi` and drop `dumpQi`?

https://github.com/PostgREST/postgrest/blob/23750e644fca895b0153f84c62df64bf16166a78/src/PostgREST/SchemaCache/Identifiers.hs#L38-L40

taimoorzaeem: The `dumpQi` function is used in the config dumping, here:

https://github.com/PostgREST/postgrest/blob/75d4131aa67a8169e43082ff5c2e3cd995b576e7/src/PostgREST/Config.hs#L171

With `quoteQi` we set the config to `db-pre-config="true"`, but then get `db-pre-config="\"true\""` in the config dump, which is different, right?

steve-chavez: Right, on second thought we shouldn't change usage on those other places on this PR.

But we definitely some clarification between `dumpQi/quoteQi`, even if it's just a comment on the function saying why we have two functions that are so similar. That can be done on another PR.

(if we can stick to a single function that'd be even better)

taimoorzaeem: Alright. I think we can move `dumpQi` to `Config.hs` module because that's the only place it's used. This should eliminate at least some confusion. Done that on #4435.

taimoorzaeem:
@wolfgangwalther To backport this to `v14`, we need to first backport #4432 right? Could you please create the `backport v14` label? Or we can just do this manually in one PR too.

wolfgangwalther:
> Could you please create the `backport v14` label?

Added that label to the other PR.

PR #4431, opened 2025-10-27
chore(changelog): update versioning scheme description

taimoorzaeem:
The changelog description mentioned that we follow semantic versioning but from now on we don't. Hence updated the description to reflect new versioning policy.


steve-chavez:
This one should be backported to v14 too, just noted we mention semver there https://github.com/PostgREST/postgrest/blob/v14/CHANGELOG.md

PR #4432, opened 2025-10-27
refactor: move escapeIdent function to Identifiers.hs

taimoorzaeem:
Moves the functions `escapeIdent` and `trimNullChars` to `SchemaCache/Identifiers.hs` module.

This should help with #4420.


src/PostgREST/Query/SqlFragment.hs
```diff
@@ -21,28 +29,19 @@ module PostgREST.Query.SqlFragment
...
-  , intercalateSnippet
-  , explainF
-  , setConfigWithConstantName
-  , setConfigWithDynamicName
-  , setConfigWithConstantNameJSON
-  , escapeIdent
-  , escapeIdentList
-  , schemaDescription
-  , accessibleTables
-  , accessibleFuncs
```
steve-chavez, line 45: There's a reordering here that could be split up in another commit.

steve-chavez (approved):

PR #4433, opened 2025-10-28
refactor: Isolate unrelated errors in the `Error` module

nikita-volkov:
This is exclusively a structural refactoring of the `PostgREST.Error` module, which decouples various errors by distributing them across cohesive modules. It also encapsulates the helper definitions.

- No changes to the API of the refactored module.
- No changes in its implementation.
- No renamings.
- Reformatting avoided as much as possible.

I've noticed myself needing this refactoring while working on migrating the codebase to the upcoming release of Hasql. After this refactoring that concern is now clearly located under the `PostgREST.Error.PgError.*` namespace.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,3 +53,15 @@ Check the [development docs](https://github.com/PostgREST/postgrest/blob/main/ni
 ### Running Tests
 
 For instructions on running tests, see the [development docs](https://github.com/PostgREST/postgrest/blob/main/nix/README.md#testing).
+
+### Structuring commits in pull requests
+
+To simplify reviews, make it easy to split pull requests if deemed necessary, and to maintain clean and meaningful history of changes, you will be asked to update your PR if it does not follow the below rules:
+
+* It must be possible to merge the PR branch into target using `git merge --ff-only`, ie. the source branch must be rebased on top of target.
+* No merge commits in the source branch.
+* All commits in the source branch must be self contained, meaning: it should be possible to treat each commit as a separate PR.
+* Commits in the source branch must contain only related changes (related means the changes target a single problem/goal). For example, any refactorings should be isolated from the actual change implementation into separate commits.
+* Tests, documentation, and changelog updates should be contained in the same commits as the actual code changes they relate to. An exception to this rule is when test or documentation changes are made in separate PR.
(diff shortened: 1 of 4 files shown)


wolfgangwalther:
> while working on migrating the codebase to the upcoming release of Hasql

Please note that we are still stuck on the versioning we are currently using, because we still need to support GHC 9.4 for the time being. So there is no quick way forward. Also some other PRs already have code to update hasql, once we get there, for example https://github.com/PostgREST/postgrest/pull/4193/commits/612dfec6a9666d04cad0be5ac1d8733e48a250a3.

nikita-volkov:
> because we still need to support GHC 9.4 for the time being. So there is no quick way forward.

The latest releases of Hasql and the upcoming version is tested to support GHC down to 8.10. So there should be no issue with that.

wolfgangwalther:
> The latest releases of Hasql and the upcoming version is tested to support GHC down to 8.10. So there should be no issue with that.

Interesting. I would have to look into that again, but I don't think it works with the Nixpkgs-provided GHC 9.4. I think this relates to the fact that Nixpkgs always forces the core libraries in the version that the compiler shipped with. IIRC we'd hit bounds for `containers` very quickly.

nikita-volkov:
Let's check. If any issues come up, they should be easy to fix.

Also please notice that this PR is just refactoring and doesn't deal with dependencies.

nikita-volkov:
Pinging to remind about this PR. It's just a refactoring. No behaviour or dependency changes here.

Due to the stalling of this PR, it now has conflicts with master. I can resolve those, but I need assurances that you're willing to merge it and it's not gonna be a waste of time.

wolfgangwalther:
> I've noticed myself needing this refactoring while working on migrating the codebase to the upcoming release of Hasql.

While working on #4829, I didn't see anything wrong right now. So I assume this only comes up with hasql 1.10, which was part of #4434, but not #4829. Since we don't intend to go beyond 1.9.3.1, I assume this PR is not needed anymore.

There's also a concern that this PR could potentially violate our new [AI policy](https://github.com/PostgREST/postgrest/blob/main/CONTRIBUTING.md#ai-policy). Even though I see from looking at the diff locally, that the vast majority of lines is only copied between files, there are still some new lines, that would need to be looked at manually.

> Due to the stalling of this PR, it now has conflicts with master. I can resolve those, but I need assurances that you're willing to merge it and it's not gonna be a waste of time.

I don't think it's worth your time to rebase this. Let's close it instead.

Thank you nevertheless for bringing this (and the other PR that I just closed!) up. The intent to contribute is very much appreciated :)

PR #4446, opened 2025-11-03
[NEEDS DISCUSSION] docs: add note to resource embedding and vertical filtering section

### LINKED ISSUE #4352: DELETE ignores top-level filtering of resource embeddings
### Environment

* PostgreSQL version: 17.6
* PostgREST version: 13.0.6
* Operating system: NixOS

### Description of issue

For details, please see [this RME](https://github.com/rzetterberg/postgrest-delete-resource-embeddings).

My RME has the following DDL:

Image

The data of the RME consists of: 

- 2 directors
  - John
  - Albert
- 5 films
  - 3 directed by John
  - 2 directed by Albert

Imagine that I want to delete all films directed by directors with the name `John`, I would then use the following query parameters:

- `select=*,director:directors!inner(*)`
- `director.name=eq.John`

To double check that the resource embeddings top-level filtering have been specified correctly, I first start by sending a `GET` with the query parameters above:

```
GET http://localhost:8080/films?select=*,director:directors!inner(*)&director.name=eq.John
```

That works, as expected, it only returns 3 of the 5 films, and they all belong to John:

```
[
  {"id":1,"name":"Movie about dogs","director_id":1,"director":{"id": 1, "name": "John"}},
  {"id":2,"name":"Movie about cats","director_id":1,"director":{"id": 1, "name": "John"}},
  {"id":3,"name":"Movie about cows","director_id":1,"director":{"id": 1, "name": "John"}}
]
```

Looking at the PostgreSQL log, I see that the following query was generated by PostgREST:

```sql
WITH pgrst_source AS (
  SELECT "api"."films".*
       , row_to_json("films_director_1".*)::jsonb AS "director"
    FROM "api"."films"
         INNER JOIN LATERAL (
           SELECT "directors_1".*
             FROM "api"."directors" AS "directors_1"
            WHERE "directors_1"."name" = $1
              AND "directors_1"."id" = "api"."films"."director_id"
            LIMIT $2
           OFFSET $3
         ) AS "films_director_1"
      ON TRUE
   LIMIT $4
  OFFSET $5
)

-- Parameters: $1 = 'John', $2 = '100', $3 = '0', $4 = '100', $5 = '0'
```

However, when I send a `DELETE` request with the same query parameters: 

```
DELETE http://localhost:8080/films?select=*,director:directors!inner(*)&director.name=eq.John
```

**All** films are deleted and PostgREST generates the following query: 

```sql
WITH pgrst_source AS (
     DELETE FROM "api"."films"
  RETURNING 1
)
```

I've encountered this problem in both `v12` and `v13,` where the documentation says: 

> Deletions also support Return Representation, **Resource Embedding** and Vertical Filtering.

Let me know if you need any more information, if I can help in some way, or if you are having problems getting the RME working. 

Also, I found out about this problem when `pg-safeupdate` refused to delete all rows in my project, so no data has been harmed in making this bug report :wink:

taimoorzaeem:
Adds a note about resource embedding and vertical filtering sections that they are top-level operations. This resolves #4352.

### DIFF
--- a/docs/references/api/resource_embedding.rst
+++ b/docs/references/api/resource_embedding.rst
@@ -5,6 +5,10 @@ Resource Embedding
 
 PostgREST allows including related resources in a single API call. This reduces the need for many API requests.
 
+.. note::
+
+  It is **not** possible to perform mutations *on* related resources. When combining mutations and resource embedding, the mutation is performed first and then the related resource is returned in combination with :ref:`prefer_return`.
+
 .. _fk_join:
 
 Foreign Key Joins
--- a/docs/references/api/tables_views.rst
+++ b/docs/references/api/tables_views.rst
@@ -246,6 +246,10 @@ When certain columns are wide (such as those holding binary data), it is more ef
 
 The default is ``*``, meaning all columns. This value will become more important below in :ref:`resource_embedding`.
 
+.. note::
+
+  It is **not** possible to perform mutations *on* filtered data. When combining mutations and vertical filtering, the mutation is performed first and then the filter is applied in combination with :ref:`prefer_return`.
+
 .. _renaming_columns:
 
 Renaming Columns


docs/references/api/resource_embedding.rst
```diff
@@ -5,6 +5,10 @@ Resource Embedding
 
 PostgREST allows including related resources in a single API call. This reduces the need for many API requests.
 
+.. note::
+
+  Resource embedding is a **top-level** operation, which means that for mutation queries, it is performed on the affected resource **after** the mutation.
```
wolfgangwalther: I don't understand what "top-level" means here and how that relates to this.

taimoorzaeem: Hmm, I see. Perhaps I should clarify it differently.

docs/references/api/resource_embedding.rst
```diff
@@ -5,6 +5,10 @@ Resource Embedding
 
 PostgREST allows including related resources in a single API call. This reduces the need for many API requests.
 
+.. note::
+
+  It is **not** possible to perform mutations *on* related resources. When combining mutations and resource embedding, the mutation is performed first and then the related resource is returned in combination with :ref:`prefer_return`.
```
wolfgangwalther, line 10: After reading the discussion about the breaking change again, I think we should not document this behavior. This is a bug that needs to be fixed by throwing an error accordingly.

laurenceisla: Can we consider it a bug if we are actively testing for it (and for a while now)?

https://github.com/PostgREST/postgrest/blob/bccf7ac97a8a4d821fe199cab530ec3d37c6c653/test/spec/Feature/Query/DeleteSpec.hs#L73-L78

Not against removing this behavior, but I think it could still be considered a breaking change due to these tests.

taimoorzaeem:
Closing as stale. Discussion can be continued on #4352.

Issue #3873, opened 2025-01-23
Remove unused packages from `postgrest.cabal`?


taimoorzaeem:
### Problem

Using warning `-Wunused-packages` shows that we depend on some unused packages that can be removed for hygiene purposes.

#### Library
```
<no location info>: error: [-Wunused-packages, -Werror=unused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - gitrev-1.3.1 (exposed by flag -package-id gitrev-1.3.1-KhpMNofwsxABRoyEGWB0ci)
      - postgresql-libpq-0.10.1.0 (exposed by flag -package-id postgresql-libpq-0.10.1.0-FSM2HH7QlWiBFY7y3DUKHq)
      - wai-logger-2.5.0 (exposed by flag -package-id wai-logger-2.5.0-58d2260507d3ee2a42fc39045c18cb0226c038aad32e4531b33c165fde01a70a)
```
#### Test Suite
```
<no location info>: error: [-Wunused-packages, -Werror=unused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - auto-update-0.1.6 (exposed by flag -package-id auto-update-0.1.6-4gXBEKjiUIyIgtDokWp2Hk)
```

PR #4450, opened 2025-11-04
chore: remove unused haskell dependencies

taimoorzaeem:
Removes unused haskell dependencies.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -102,10 +102,8 @@ library
                     , auto-update               >= 0.1.4 && < 0.3
                     , base64-bytestring         >= 1 && < 1.3
                     , bytestring                >= 0.10.8 && < 0.13
-                    , cache                     >= 0.1.3 && < 0.2.0
                     , case-insensitive          >= 1.2 && < 1.3
                     , cassava                   >= 0.4.5 && < 0.6
-                    , clock                     >= 0.8.3 && < 0.9.0
                     , configurator-pg           >= 0.2.11 && < 0.3
                     , containers                >= 0.5.7 && < 0.7
                     , cookie                    >= 0.4.2 && < 0.6
@@ -118,11 +116,9 @@ library
                     , hasql-notifications       >= 0.2.2.2 && < 0.2.3
                     , hasql-pool                >= 1.0.1 && < 1.1
                     , hasql-transaction         >= 1.0.1 && < 1.2
-                    , heredoc                   >= 0.2 && < 0.3
                     , http-client               >= 0.7.19 && < 0.8
                     , http-types                >= 0.12.2 && < 0.13
                     , insert-ordered-containers >= 0.2.2 && < 0.3
-                    , iproute                   >= 1.7.0 && < 1.8
                     , jose-jwt                  >= 0.9.6 && < 0.11
                     , lens                      >= 4.14 && < 5.4
                     , lens-aeson                >= 1.0.1 && < 1.3
@@ -132,8 +128,6 @@ library
                     , network-uri               >= 2.6.1 && < 2.8
                     , optparse-applicative      >= 0.13 && < 0.19
                     , parsec                    >= 3.1.11 && < 3.2
-                    -- Technically unused, can be removed after updating to hasql >= 1.7
-                    , postgresql-libpq          >= 0.10
                     , prometheus-client         >= 1.1.1 && < 1.2.0
                     , protolude                 >= 0.3.1 && < 0.4
                     , regex-tdfa                >= 1.2.2 && < 1.4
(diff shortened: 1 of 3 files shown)


steve-chavez:
Is it possible to detect unused dependencies on CI?

steve-chavez (approved):

taimoorzaeem:
> Is it possible to detect unused dependencies on CI?

Not really. The way I found about those is by compiling with `-Wunused-packages`. However, I think the reason we don't have it already is because it gives two "false positives":

```
<no location info>: error: [-Wunused-packages, -Werror=unused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - postgresql-libpq-0.11.0.0 (exposed by flag -package-id postgresql-libpq-0.11.0.0-FYbh1w1IFpzFPSeYd7F80E)
      - wai-logger-2.5.0 (exposed by flag -package-id wai-logger-2.5.0-GgCKOfZCPNy9oBPl1Kfzlq)
```

 We discussed this #3873 as well. For me, locally, it all builds and tests run just fine, even without the above two.

Let me remove them temporarily to check if CI works fine without them. EDIT: It does!!

steve-chavez:
The auto-cancelled CI jobs say:

> This is a scheduled macos-13 brownout. The macOS-13 based runner images are being deprecated. For more details, see https://github.com/actions/runner-images/issues/13046.

I've tried to restart them manually several times, but it doesn't work. Maybe we should fix that in another PR.

postgrest.cabal
```diff
@@ -132,8 +128,6 @@ library
                     , network-uri               >= 2.6.1 && < 2.8
                     , optparse-applicative      >= 0.13 && < 0.19
                     , parsec                    >= 3.1.11 && < 3.2
-                    -- Technically unused, can be removed after updating to hasql >= 1.7
-                    , postgresql-libpq          >= 0.10
```
wolfgangwalther, line 136: The comment explicitly says "can be removed after updating hasql >= 1.7, yet we have *not* updated to that, yet.

This has been explained here already: https://github.com/PostgREST/postgrest/issues/3873

This had been introduced here: https://github.com/PostgREST/postgrest/pull/3177

Removing this change means you might get an older version of postgresql-libpq in certain builds, which will introduce regressions.

Please revert this part...

postgrest.cabal
```diff
@@ -151,11 +145,6 @@ library
                     , wai                       >= 3.2.1 && < 3.3
                     , wai-cors                  >= 0.2.5 && < 0.3
                     , wai-extra                 >= 3.1.8 && < 3.2
-                    -- We already depend on wai-logger >= 2.3.7 indirectly via wai-extra,
-                    -- but we want to depend on 2.4.0 which fixes 'unknownSocket' log output
-                    -- for unix sockets; this is tested in test/io/test_io.py. See
-                    -- https://github.com/kazu-yamamoto/logger/commit/3a71ca70afdbb93d4ecf0083eeba1fbbbcab3fc3
-                    , wai-logger                >= 2.4.0
```
wolfgangwalther, line 158: Wow, does anybody still read comments?

This was explained previously as well: https://github.com/PostgREST/postgrest/issues/3873

Please revert.

wolfgangwalther:
> I've tried to restart them manually several times, but it doesn't work. Maybe we should fix that in another PR.

We will hit these a few times. I would like to wait until December to fix this, though - so that we can still build x86_64-darwin binaries for as long as possible. I'm hoping that we might be able to get some nix static builds going for darwin by then, too.

postgrest.cabal
taimoorzaeem: > Removing this change means you might get an older version of postgresql-libpq in certain builds, which will introduce regressions.

I think we won't get an older version of `postgresql-libpq` because cabal always picks up the maximum version the satisfies the constraint.

You can verify this at your end by running `cabal freeze` and `stack ls dependencies`.

postgrest.cabal
taimoorzaeem: Cabal picks up the maximum version. We are using `wai-logger-2.5.0` already in cabal and `wai-logger-2.4.1` on stack, so no need for this constraint.

taimoorzaeem:
In a rare case in the future, where some other new library has a tighter upper bound, resultantly, the dependency constraint solving could potentially fallback to older versions.

So, to be extra safe, we should restore those pinned dependencies. Thanks @wolfgangwalther for pointing this out. My apologies. Let me restore those.

wolfgangwalther:
> In a rare case in the future, where some other new library has a tighter upper bound, resultantly, the dependency constraint solving could potentially fallback to older versions.

This doesn't need to be at some point in the future. It can also happen by just building with an older GHC or with some other constraint. Maybe because we add a dependency in the future or whatever.

There is a certain knowledge in these constraints, that we are losing by removing them.

postgrest.cabal
steve-chavez: @taimoorzaeem How about adding a clear:

```
DO NOT REMOVE!!!
```

Or something like that in the comments? The reasoning here is easy to miss otherwise.

taimoorzaeem: No, I think we should be more clear about the reasoning here. After reverting this, I'll add a comment here explaining exactly **why** we can't remove them yet.

steve-chavez:
> The auto-cancelled CI jobs say:
> This is a scheduled macos-13 brownout. The macOS-13 based runner images are being deprecated. For more details, see https://github.com/actions/runner-images/issues/13046.
> I've tried to restart them manually several times, but it doesn't work. Maybe we should fix that in another PR.

CI has been failing consistently since https://github.com/PostgREST/postgrest/commit/770c404ebec27afd18c8c5e4c233226556aa92d4 due to the above

wolfgangwalther:
I removed the macos-13 jobs with eb00c1991c67facc7b0e60703825897ba65a40b3. It makes no sense to wait for the Nix thing right now, because Nixpkgs recently bumped the minimum version to 14 as well. Thus, we will not be able to support MacOS 13 anyway.

PR #4452, opened 2025-11-05
chore: pin haskell postgresql-libpq and wai-logger deps

taimoorzaeem:
Partially reverts #4450.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -128,6 +128,9 @@ library
                     , network-uri               >= 2.6.1 && < 2.8
                     , optparse-applicative      >= 0.13 && < 0.19
                     , parsec                    >= 3.1.11 && < 3.2
+                    -- Technically unused, can be removed after updating to hasql >= 1.7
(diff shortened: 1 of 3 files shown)


postgrest.cabal
```diff
@@ -128,6 +128,9 @@ library
                     , network-uri               >= 2.6.1 && < 2.8
                     , optparse-applicative      >= 0.13 && < 0.19
                     , parsec                    >= 3.1.11 && < 3.2
+                    -- Technically unused, can be removed after updating to hasql >= 1.7
+                    -- hasql-1.6.4.4 has postgresql-libpq lower-bound >= 0.9 but we need >= 0.10.1
+                    , postgresql-libpq          >= 0.10.1
```
taimoorzaeem, line 133: Added another comment here for further clarification. Also bumped it to `0.10.1` instead of just `0.10` to match stackage version.

steve-chavez:
@taimoorzaeem How about a clean `git revert` of https://github.com/PostgREST/postgrest/commit/b6bb8382e7a85ed4c5cba1148c0503dca83609b3 instead of a partial revert? Review is easier that way.

Any additions can be done in another commit.

Issue #4462, opened 2025-11-12
Continuous memory growth and OOM kills since PostgREST v13+


bolinocroustibat:
### Description

After upgrading from PostgREST **v12.2.3** → **v13.0.8** and **v14.0.0**, we observe a reproducible memory issue in production:  
RAM usage continuously increases over time until the process is killed by the OOM killer.  
This regression does not occur with v12, where memory usage remains stable even under heavy load.

### Context

We run PostgREST as part of a public API gateway (`tabular-api.data.gouv.fr`), which dynamically exposes a large PostgreSQL database (tens of thousands of tables).  
Each dataset in the DB corresponds to one physical table, so the total number of tables can vary — currently around **55K**, but previously up to **120K**.

PostgREST is deployed via Nomad/Consul (Blue/Green deployments, 2 instances, 6–8 GB RAM each).

### Observed behavior

- With **v13+**, memory usage starts around **25–50%** at startup and grows steadily until the process reaches 100% and is OOM-killed (every 1–2 hours under load).  
- With **v12.2.3**, memory usage may start higher, but remains **stable** over time under identical load and schema size.  
- Reducing the number of tables (from 120K → 55K) does **not** change the behavior.  
- Restarting PostgREST temporarily resets memory, but the same growth pattern repeats.

<img width="563" height="268" alt="Image" src="https://github.com/user-attachments/assets/2d04cfde-ae0e-4372-99dd-460f0a61f81d" />

### Positive change since v14

Startup appears **very significantly smoother and faster** in v14+ — lower initial RAM consumption and faster readiness.  
However, this comes at the cost of **continuous memory growth** over time at least for v13+, which was not observed in previous releases.

### What we’ve tried

- Increasing PostgREST memory allocation (6 GB → 8 GB): no effect, still OOM after a few hours.  
- Purging half of the tables: no effect.  
- Reverting to **v12.2.3**: memory stable again, no OOM.

### Environment

- **PostgreSQL:** 15.8 (Debian package)  
- **PostgREST versions tested:**
  - ✅ 12.2.3 — stable memory behavior  
  - ❌ 13.0.8 — continuous memory growth → OOM  
  - ❌ 14.0.0 — same issue  
- **Deployment:** Nomad + Consul (Blue/Green, 2 instances, 6–8 GB RAM each)  
- **Approx. tables:** ~55,000  
- **Approx. requests:** hundreds/minute (read-only)

### Expected behavior

[...]

Issue #4463, opened 2025-11-13
`ServerDisconnectedError` when querying tables created after PostgREST startup since v13+


bolinocroustibat:
### Description

Since upgrading from PostgREST **v12.2.3** to **v13.0.8** and **v14.0.0**, we observe a reproducible issue where **some recently created tables cannot be queried through PostgREST**.  
Requests to these tables fail with a connection abort (`ServerDisconnectedError` / `RemoteDisconnected`).

This issue does *not* occur on v12.2.3.

### Context

We run PostgREST as part of a public API gateway (`tabular-api.data.gouv.fr`) that dynamically exposes a large PostgreSQL database (tens of thousands of tables).  
New tables are created regularly, while PostgREST continues to run.

Deployment: Nomad + Consul (Blue/Green, typically 2 instances).

### Observed behavior

Starting with **v13**, queries to **some recently created tables** fail, while older tables continue to work normally.

From our observations, the problematic tables appear to be **tables created after PostgREST has started**.

The error is typically:
`requests.exceptions.ConnectionError`: (‘Connection aborted.’, RemoteDisconnected(‘Remote end closed connection without response’))

Notes:

- The issue affects only a subset of newly created tables.
- Tables that already existed at startup behave normally.
- Restarting PostgREST makes the problematic tables work again — until new tables are created.

This behavior is not present in **v12.2.3**.

### What we’ve tried

- Restarting PostgREST: temporarily fixes the issue.
- Purging the number of tables (120k → 55k): no effect.
- Reverting to **v12.2.3**: issue disappears entirely.
- Upgrading to **v14.0.0**: same behavior as v13.

### Environment

- **PostgreSQL:** 15.8 (Debian package)
- **PostgREST versions tested:**
  - ✅ 12.2.3 — works correctly  
  - ❌ 13.0.8 — new tables trigger `ServerDisconnectedError`  
  - ❌ 14.0.0 — same issue  
- **Deployment:** Nomad + Consul
- **Schema size:** ~55K tables (dynamic workload)
- **Requests:** hundreds/minute (read-only)

### Notes

This suggests a regression introduced in v13 regarding table discovery or schema reloading?

[A separate issue we opened](https://github.com/PostgREST/postgrest/issues/4462) describes a continuous memory growth / OOM regression that also appears starting in v13.  
We are not sure whether the two problems share a root cause, but both regressions disappear when reverting to v12.

Happy to provide logs if needed.

steve-chavez:
[...]

PR #4468, opened 2025-11-14
Revert "fix: handle queries on non-existing table gracefully "

steve-chavez:
Temporarily revert commit 390ba199329c75f034e23e3c9029fdb22ec8ff2c to help with debugging https://github.com/PostgREST/postgrest/issues/4462

**Edit**: it was confirmed on #4462 that this revert solves the issue.

### DIFF
--- a/docs/references/errors.rst
+++ b/docs/references/errors.rst
@@ -305,10 +305,6 @@ Related to a :ref:`schema_cache`. Most of the time, these errors are solved by :
 |               |             | in the ``columns`` query parameter is not found.            |
 | PGRST204      |             |                                                             |
 +---------------+-------------+-------------------------------------------------------------+
-| .. _pgrst205: | 404         | Caused when the :ref:`table specified <tables_views>` in    |
-|               |             | the URI is not found.                                       |
-| PGRST205      |             |                                                             |
-+---------------+-------------+-------------------------------------------------------------+
 
 .. _pgrst3**:
 
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -49,7 +49,6 @@ import PostgREST.SchemaCache.Relationship (Cardinality (..),
                                            RelationshipsMap)
 import PostgREST.SchemaCache.Routine      (Routine (..),
                                            RoutineParam (..))
-import PostgREST.SchemaCache.Table        (Table (..))
 import Protolude
 
 
@@ -250,7 +249,6 @@ data SchemaCacheError
   | NoRelBetween Text Text (Maybe Text) Text RelationshipsMap
   | NoRpc Text Text [Text] MediaType Bool [QualifiedIdentifier] [Routine]
   | ColumnNotFound Text Text
-  | TableNotFound Text Text [Table]
   deriving Show
 
 instance PgrstError SchemaCacheError where
@@ -259,7 +257,6 @@ instance PgrstError SchemaCacheError where
   status NoRelBetween{}        = HTTP.status400
   status NoRpc{}               = HTTP.status404
   status ColumnNotFound{}      = HTTP.status400
-  status TableNotFound{}       = HTTP.status404
 
   headers _ = mempty
 
@@ -269,7 +266,6 @@ instance ErrorBody SchemaCacheError where
   code NoRpc{}               = "PGRST202"
   code AmbiguousRpc{}        = "PGRST203"
   code ColumnNotFound{}      = "PGRST204"
-  code TableNotFound{}       = "PGRST205"
 
   message (NoRelBetween parent child _ _ _)  = "Could not find a relationship between '" <> parent <> "' and '" <> child <> "' in the schema cache"
   message (AmbiguousRelBetween parent child _) = "Could not embed because more than one relationship was found for '" <> parent <> "' and '" <> child <> "'"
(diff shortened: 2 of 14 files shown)


mkleczek:
See #4472

steve-chavez:
I wonder why https://github.com/PostgREST/postgrest/pull/3869 would cause https://github.com/PostgREST/postgrest/issues/4462#issue-3617468585? @taimoorzaeem Any ideas?

It would make sense to consume considerable memory for 125K tables at schema cache creation time, but why would it keep increasing as `findTable` is called on each request?

https://github.com/PostgREST/postgrest/blob/d9055921150c9f700cdefa21d0d6bac933fca42b/src/PostgREST/Plan.hs#L813-L817

taimoorzaeem:
> I wonder why #3869 would cause [#4462 (comment)](https://github.com/PostgREST/postgrest/issues/4462#issue-3617468585)? @taimoorzaeem Any ideas?
> 
> It would make sense to consume considerable memory for 125K tables at schema cache creation time, but why would it keep increasing as `findTable` is called on each request?

It's quite strange. I tried tracing it and on error cases, control flow ends on the evaluation of `Fuzzy.getOne` here:

https://github.com/PostgREST/postgrest/blob/50eec773bf1117e1f60a96d807666c0a924b108e/src/PostgREST/Error.hs#L435

I am debugging further to get to the root cause.

taimoorzaeem:
I tried to run the fuzzy search operation on a large set. I got the source tarball of [fuzzyset-0.2.4](https://hackage.haskell.org/package/fuzzyset-0.2.4) and built it with `ghc-9.4.8`. Here is what happened:

```haskell
-- [~/github/fuzzyset]$ cabal v2-repl
ghci> import Data.FuzzySet
ghci> import Data.Text
ghci> fuzzyset = fromList ([ pack ("unknown_table_" <> show (i :: Int)) | i <- [1..100000]])
ghci> getOne fuzzyset (pack "table-with-a-weird-name")
Error: [Cabal-7125]
repl failed for fuzzyset-0.2.4. The build process was killed (i.e. SIGKILL). The typical
reason for this is that there is not enough memory available (e.g. the OS killed a
process using lots of memory).
```

The memory and CPU consumption spiked to 100% and then either the OS kills the process like above or the process nearly kills my OS (a few times it caused my ubuntu OS to kill all processes and logged out :crying_cat_face:). This does not seem safe.

The `top` command output is something like:

<pre>
PID   USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
61689 taimooz   20   0 1024.5g   9.8g 128160 S <b>100.3  85.8</b>   0:25.65 <b>ghc-9.4+</b>
</pre>

#### In Nix Environment

However, if I try the same thing in postgrest nix-shell environment, the CPU spikes to 100% but the MEM usage remains low and the OS does not kill the process the expression eventually (takes about 1-2 minutes) evaluates to `Nothing`.

`top` output looks something like:

<pre>
PID   USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
62098 taimooz   20   0 1024.8g 987124 174764 S <b>100.3   6.3</b>   0:53.05 <b>ghc</b>    
</pre>

Now, this is my analysis so far and from just testing in `ghci`. I am not sure how our warp HTTP server handles this kind of exception.

steve-chavez:
:thinking: @taimoorzaeem So are you saying the OOMs reported in https://github.com/PostgREST/postgrest/issues/4462#issue-3617468585 are not due to `findTable` calls but due to the fuzzy search calls? (i.e. the problem on https://github.com/PostgREST/postgrest/issues/4463#issue-3620709780). I think that would make sense.

If so, the solution would be to disable the fuzzy search instead of fully reverting https://github.com/PostgREST/postgrest/pull/3869

taimoorzaeem:
> So are you saying the OOMs reported in [#4462 (comment)](https://github.com/PostgREST/postgrest/issues/4462#issue-3617468585) are not due to `findTable` calls but due to the fuzzy search calls? (i.e. the problem on [#4463 (comment)](https://github.com/PostgREST/postgrest/issues/4463#issue-3620709780)). I think that would make sense.

Yes, `findTable` lookup is fast, but when it is not found, we do the fuzzy search to generate error `hint` which causes the entire issue.

> If so, the solution would be to disable the fuzzy search instead of fully reverting #3869

Agreed. So, do we remove fuzzy searching entirely or conditionally based on a config as described in #4499?

steve-chavez:
> Agreed. So, do we remove fuzzy searching entirely or conditionally based on a config as described in https://github.com/PostgREST/postgrest/issues/4499?

Could we disable computing the hint when we have a large number of relations? That number could be 4K for starters.

taimoorzaeem:
> Could we disable computing the hint when we have a large number of relations? That number could be 4K for starters.

Yes, absolutely. 4K seems like a good number to me. Will do 👍

mkleczek:
> > Could we disable computing the hint when we have a large number of relations? That number could be 4K for starters.
> 
> Yes, absolutely. 4K seems like a good number to me. Will do 👍

@taimoorzaeem @steve-chavez - I've added this check to https://github.com/PostgREST/postgrest/pull/4472 (see https://github.com/mkleczek/postgrest/blob/8f00c17ec948fb59c866a19c14af88339482c89c/src/PostgREST/Error.hs#L433)

`dbTables` size calculation has to be memoized as well (`HashMap.size` is O(n)) so doing both (ie. memoizing `FuzzySet` and `dbTables` size) makes sense IMHO.

steve-chavez:
In light of https://github.com/PostgREST/postgrest/issues/4873, I think we should do this. 

It's not correct to force failure for all requests if the scache is unavailable, specially if we can serve them without it, as it's the case for single table requests. While this is happening `/ready` can return 503 as it's a degraded service.

Maybe we can look at different ways to solve the issues that https://github.com/PostgREST/postgrest/pull/3869 originally solved. Some ideas on https://github.com/PostgREST/postgrest/issues/4873#issuecomment-4452271952.

steve-chavez:
@taimoorzaeem Could you try reverting https://github.com/PostgREST/postgrest/commit/390ba199329c75f034e23e3c9029fdb22ec8ff2c :pray:? Another PR should be fine, I got lost here.

> Maybe we can look at different ways to solve the issues that https://github.com/PostgREST/postgrest/pull/3869 originally solved.

Then we can look at how to solve those problems. Possibly add a test for https://github.com/PostgREST/postgrest/issues/4613 too.

taimoorzaeem:
Closing this because it is now being discussed in #5120.

PR #4469, opened 2025-11-15
nix: shorten `postgrest-with-postgresql-xx` scripts

### LINKED ISSUE #4461: nix: `postgrest-test-replica` fails locally
### Environment

* PostgREST version: devel, v14
* Operating system: ubuntu

### Description of issue

`postgrest-test-replica` fails with the following error:

```
[nix-shell]$ postgrest-test-replica
- cabal.project
- cabal.project.freeze
Resolving dependencies...
postgrest-with-postgresql-17: You can connect with: psql 'postgres:///postgres?host=/tmp/nix-shell-13818-1957527406/postgrest/postgrest-with-postgresql-17-zz5/socket' -U postgres
postgrest-with-postgresql-17: You can tail the logs with: tail -f /tmp/nix-shell-13818-1957527406/postgrest/postgrest-with-postgresql-17-zz5/db.log
pg_ctl: could not start server
Examine the log output.
Temporary directory kept at: /tmp/nix-shell-13818-1957527406/postgrest/postgrest-with-postgresql-17-zz5
```

It fails with `pg_ctl: could not start server` for me locally. Does it fail for everyone or am I missing something? I checked this both in `main` and `v14`.

taimoorzaeem:
Renames these scripts to `postgrest-with-pg-xx`. The renaming helps reduce the length of temporary filenames. This is needed to ensure that socket file names remain under the maximum allowed length of 107 chars.

Closes #4461.

### DIFF
--- a/nix/tools/withTools.nix
+++ b/nix/tools/withTools.nix
@@ -46,7 +46,7 @@ let
         }
 
         # Avoid starting multiple layers of withTmpDb, but make sure to have the last invocation
-        # load fixtures. Otherwise postgrest-with-postgresql-xx postgrest-test-io would not be possible.
+        # load fixtures. Otherwise postgrest-with-pg-xx postgrest-test-io would not be possible.
         if ! test -v PGHOST; then
 
           mkdir -p "$tmpdir"/{db,socket}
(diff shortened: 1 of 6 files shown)


nix/tools/withTools.nix
```diff
@@ -87,7 +93,7 @@ let
           if test "$_arg_replica" = "on"; then
             replica_slot="replica_$RANDOM"
             replica_dir="$tmpdir/$replica_slot"
-            replica_host="$tmpdir/socket_$replica_slot"
+            replica_host="$tmpdir/sock_$replica_slot"
```
wolfgangwalther: We should just run the replica on the same socket path as the primary, but use a different port (it will then create a different socket). That would save a whole level of the path.

taimoorzaeem: Great idea! I will try to do these changes in a separate PR.

PR #4471, opened 2025-11-16
test(io): move authentication related tests to test_auth.py

taimoorzaeem:
The `test_io.py` module is too bloated (2100+ lines). To logically group related tests, as a first step, this commit separates authentication related IO tests into `test_auth.py` module.

### DIFF
--- a/test/io/test_auth.py
+++ b/test/io/test_auth.py
@@ -1,0 +1,427 @@
+"Auth related IO tests for PostgREST"
+
+from datetime import datetime, timedelta, timezone
+from operator import attrgetter
+import signal
+import time
+import pytest
+
+from config import BASEDIR, CONFIGSDIR, FIXTURES, SECRET
+from util import authheader, jwtauthheader, parse_server_timings_header
+from postgrest import (
+    run,
+    sleep_until_postgrest_config_reload,
+    sleep_until_postgrest_scache_reload,
+    wait_until_exit,
+)
+
+
+@pytest.mark.parametrize(
+    "secretpath",
+    [path for path in (BASEDIR / "secrets").iterdir() if path.suffix != ".jwt"],
+    ids=attrgetter("name"),
+)
+def test_read_secret_from_file(secretpath, defaultenv):
+    "Authorization should succeed when the secret is read from a file."
+
+    env = {**defaultenv, "PGRST_JWT_SECRET": f"@{secretpath}"}
+
+    if secretpath.suffix == ".b64":
+        env["PGRST_JWT_SECRET_IS_BASE64"] = "true"
+
+    secret = secretpath.read_bytes()
+    headers = authheader(secretpath.with_suffix(".jwt").read_text())
+
+    with run(stdin=secret, env=env) as postgrest:
+        response = postgrest.session.get("/authors_only", headers=headers)
+        print(response.text)
+        assert response.status_code == 200
+
+
+def test_read_secret_from_stdin(defaultenv):
+    "Authorization should succeed when the secret is read from stdin."
+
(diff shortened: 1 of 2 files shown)


test/io/test_auth.py
```diff
@@ -0,0 +1,472 @@
...
+        assert response.text == ""
+        assert response.status_code == 204
+        sleep_until_postgrest_config_reload()
+
+        response = postgrest.session.get("/authors_only", headers=headers)
+        assert response.status_code == 401
+
+
+def test_jwt_cache_server_timing(defaultenv):
+    "server-timing duration is exposed for JWT with expiry"
```
steve-chavez, line 312: This one looks more related to observability

test/io/test_auth.py
```diff
@@ -0,0 +1,472 @@
...
+
+        assert first_dur >= 0
+        assert second_dur >= 0
+
+
+def test_jwt_cache_purges_expired_entries(defaultenv):
+    "test expired cache entries are purged on cache miss"
+
+    # The verification of actual cache size reduction is done manually, see https://github.com/PostgREST/postgrest/pull/3801#issuecomment-2620776041
+    # This test is written for code coverage of purgeExpired function
```
steve-chavez: Is this test still relevant after https://github.com/PostgREST/postgrest/pull/4084?

steve-chavez (commented):
Not sure if the test module organization makes the most sense.

Additionally, it's hard to check if all the tests are moved over to the new module and confirm some are not missing.

Can we have something like pgtap `SELECT plan(<number_of_tests>)` ([ref](https://pgtap.org/documentation.html#iloveitwhenaplancomestogether)) in pytest? 

That would ensure we're not deleting any tests unintentionally but comes with the added cost of having to update it every time we add tests (maybe not so bad).

wolfgangwalther:
> Additionally, it's hard to check if all the tests are moved over to the new module and confirm some are not missing. (I'd have to count each test manually)

You can set the git config value [`diff.colorMoved`](https://git-scm.com/docs/diff-config#Documentation/diff-config.txt-diffcolorMoved) and then look at the diff locally. This will give you a very good idea of which lines were moved 1:1 and which might be missing.

I did that and can confirm that no tests have been lost on the transition. No new tests have been added either. It's a clean move. I have not looked at *which* tests were moved, though.

taimoorzaeem:
> You can set the git config value [`diff.colorMoved`](https://git-scm.com/docs/diff-config#Documentation/diff-config.txt-diffcolorMoved) and then look at the diff locally. This will give you a very good idea of which lines were moved 1:1 and which might be missing.

TIL. This is awesome. Thanks for the info :+1: .

test/io/test_auth.py
taimoorzaeem: Yes, this isn't relevant anymore. The test was written for `code-cov` but that part of code was removed in #4084, so I think we can remove it.

taimoorzaeem:
> Additionally, it's hard to check if all the tests are moved over to the new module and confirm some are not missing. (I'd have to count each test manually)

FWIW, I used vim `:%s/def test_//gn` to count the tests.

> Not sure if the test module organization makes the most sense.

I mostly agree on this. My motive for doing this initial separation was to group similar tests somewhat together. Before this, they were spread around randomly in the `test_io.py` module. This grouping can then help a bit in spotting common patterns, irrelevant tests (not so much I think), or reusing previously used logic.

test/io/test_auth.py
taimoorzaeem: Opened #4486 for this for easier review.

test/io/test_auth.py
taimoorzaeem: Agreed. I have added a comment on this test to indicate that it should be moved to a more appropriate test module in the future.

PR #4472, opened 2025-11-17
fix: Performance and high memory usage of relation hint calculation

### LINKED ISSUE #4462: Continuous memory growth and OOM kills since PostgREST v13+

### LINKED ISSUE #4463: `ServerDisconnectedError` when querying tables created after PostgREST startup since v13+

mkleczek:
Calculation of hint message when requested relation is not present in schema cache requires creation of a FuzzySet (to use fuzzy search to find candidate tables). For schemas with many tables it is costly. This patch introduces dbTablesFuzzyIndex in SchemaCache to memoize the FuzzySet creation.

Fixes #4462
Fixes #4463

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file. From versio
 - Fix not returning `Content-Length` on empty HTTP `201` responses by @laurenceisla in #4518
 - Fix inaccurate Server-Timing header durations by @steve-chavez in #4522
 - Fix inaccurate "Schema cache queried" logs by @steve-chavez in #4522
+- Fix performance and high memory usage of relation hint calculation by @mkleczek in #4462 #4463
 
 ## [14.1] - 2025-11-05
 
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -3,6 +3,7 @@ Module      : PostgREST.Error
 Description : PostgREST error HTTP responses
 -}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# LANGUAGE NamedFieldPuns  #-}
 {-# LANGUAGE RecordWildCards #-}
 
 module PostgREST.Error
@@ -41,6 +42,7 @@ import Network.HTTP.Types.Header (Header)
 import           PostgREST.MediaType (MediaType (..))
 import qualified PostgREST.MediaType as MediaType
 
+import PostgREST.SchemaCache              (SchemaCache (SchemaCache, dbTablesFuzzyIndex))
 import PostgREST.SchemaCache.Identifiers  (QualifiedIdentifier (..),
                                            Schema)
 import PostgREST.SchemaCache.Relationship (Cardinality (..),
@@ -49,9 +51,7 @@ import PostgREST.SchemaCache.Relationship (Cardinality (..),
                                            RelationshipsMap)
 import PostgREST.SchemaCache.Routine      (Routine (..),
                                            RoutineParam (..))
-import PostgREST.SchemaCache.Table        (Table (..))
 import Protolude
-
 
 class (ErrorBody a, JSON.ToJSON a) => PgrstError a where
   status   :: a -> HTTP.Status
@@ -250,7 +250,7 @@ data SchemaCacheError
   | NoRelBetween Text Text (Maybe Text) Text RelationshipsMap
   | NoRpc Text Text [Text] MediaType Bool [QualifiedIdentifier] [Routine]
   | ColumnNotFound Text Text
-  | TableNotFound Text Text [Table]
+  | TableNotFound Text Text SchemaCache
   deriving Show
 
 instance PgrstError SchemaCacheError where
@@ -313,7 +313,7 @@ instance ErrorBody SchemaCacheError where
       where
         onlySingleParams = isInvPost && contentType `elem` [MTTextPlain, MTTextXML, MTOctetStream]
   hint (AmbiguousRpc _)      = Just "Try renaming the parameters or the function itself in the database so function overloading can be resolved"
-  hint (TableNotFound schemaName relName tbls) = JSON.String <$> tableNotFoundHint schemaName relName tbls
+  hint (TableNotFound schemaName relName schemaCache) = JSON.String <$> tableNotFoundHint schemaName relName schemaCache
 
   hint _ = Nothing
 
@@ -428,13 +428,11 @@ noRpcHint schema procName params allProcs overloadedProcs =
 
 -- |
 -- Do a fuzzy search in all tables in the same schema and return closest result
-tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
-tableNotFoundHint schema tblName tblList
+tableNotFoundHint :: Text -> Text -> SchemaCache -> Maybe Text
+tableNotFoundHint schema tblName SchemaCache{dbTablesFuzzyIndex}
   = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
     where
-      perhapsTable = Fuzzy.getOne fuzzyTableSet tblName
-      fuzzyTableSet = Fuzzy.fromList [ tableName tbl | tbl <- tblList, tableSchema tbl == schema]
-
+      perhapsTable = (`Fuzzy.getOne` tblName) =<< HM.lookup schema dbTablesFuzzyIndex
 
 compressedRel :: Relationship -> JSON.Value
 -- An ambiguousness error cannot happen for computed relationships TODO refactor so this mempty is not needed
--- a/src/PostgREST/SchemaCache.hs
+++ b/src/PostgREST/SchemaCache.hs
@@ -20,6 +20,7 @@ These queries are executed once at startup or when PostgREST is reloaded.
 
 module PostgREST.SchemaCache
   ( SchemaCache(..)
+  , TablesFuzzyIndex
   , querySchemaCache
   , showSummary
   , decodeFuncs
@@ -66,21 +67,28 @@ import PostgREST.SchemaCache.Table           (Column (..), ColumnMap,
 
 import qualified PostgREST.MediaType as MediaType
 
-import Control.Arrow    ((&&&))
-import Protolude
-import System.IO.Unsafe (unsafePerformIO)
+import           Control.Arrow    ((&&&))
+import qualified Data.FuzzySet    as Fuzzy
+import           Protolude
+import           System.IO.Unsafe (unsafePerformIO)
+
+type TablesFuzzyIndex = HM.HashMap Schema Fuzzy.FuzzySet
 
 data SchemaCache = SchemaCache
-  { dbTables          :: TablesMap
-  , dbRelationships   :: RelationshipsMap
-  , dbRoutines        :: RoutineMap
-  , dbRepresentations :: RepresentationsMap
-  , dbMediaHandlers   :: MediaHandlerMap
-  , dbTimezones       :: TimezoneNames
-  }
+  { dbTables           :: TablesMap
+  , dbRelationships    :: RelationshipsMap
+  , dbRoutines         :: RoutineMap
+  , dbRepresentations  :: RepresentationsMap
+  , dbMediaHandlers    :: MediaHandlerMap
+  , dbTimezones        :: TimezoneNames
+  -- Memoized fuzzy index of table names per schema to support approximate matching
+  -- Since index construction can be expensive, we build it once and store in the SchemaCache
+  -- Haskell lazy evaluation ensures it's only built on first use and memoized afterwards
+  , dbTablesFuzzyIndex :: TablesFuzzyIndex
+  } deriving (Show)
 
 instance JSON.ToJSON SchemaCache where
-  toJSON (SchemaCache tabs rels routs reps hdlers tzs) = JSON.object [
+  toJSON (SchemaCache tabs rels routs reps hdlers tzs _) = JSON.object [
       "dbTables"          .= JSON.toJSON tabs
     , "dbRelationships"   .= JSON.toJSON rels
     , "dbRoutines"        .= JSON.toJSON routs
@@ -90,7 +98,7 @@ instance JSON.ToJSON SchemaCache where
     ]
 
 showSummary :: SchemaCache -> Text
-showSummary (SchemaCache tbls rels routs reps mediaHdlrs tzs) =
+showSummary (SchemaCache tbls rels routs reps mediaHdlrs tzs _) =
   T.intercalate ", "
   [ show (HM.size tbls)       <> " Relations"
   , show (HM.size rels)       <> " Relationships"
@@ -138,6 +146,8 @@ data KeyDep
 -- | A SQL query that can be executed independently
 type SqlQuery = ByteString
 
+maxDbTablesForFuzzySearch :: Int
+maxDbTablesForFuzzySearch = 500
 
 querySchemaCache :: AppConfig -> SQL.Transaction SchemaCache
 querySchemaCache conf@AppConfig{..} = do
@@ -166,6 +176,11 @@ querySchemaCache conf@AppConfig{..} = do
     , dbRepresentations = reps
     , dbMediaHandlers = HM.union mHdlers initialMediaHandlers -- the custom handlers will override the initial ones
     , dbTimezones = tzones
+
+    , dbTablesFuzzyIndex =
+        -- Only build fuzzy index for schemas with a reasonable number of tables
+        -- Fuzzy.FuzzySet is memory heavy we just don't use it for large schemas
+        Fuzzy.fromList <$> HM.filter ((< maxDbTablesForFuzzySearch) . length) (HM.fromListWith (<>) ((qiSchema &&& pure . qiName) <$> HM.keys tabsWViewsPks))
     }
   where
     schemas = toList configDbSchemas
@@ -203,6 +218,7 @@ removeInternal schemas dbStruct =
     , dbRepresentations = dbRepresentations dbStruct -- no need to filter, not directly exposed through the API
     , dbMediaHandlers   = dbMediaHandlers dbStruct
     , dbTimezones       = dbTimezones dbStruct
+    , dbTablesFuzzyIndex = dbTablesFuzzyIndex dbStruct
     }
   where
     hasInternalJunction ComputedRelationship{} = False
--- a/test/io/test_big_schema.py
+++ b/test/io/test_big_schema.py
@@ -70,3 +70,23 @@ def test_should_not_fail_with_stack_overflow(defaultenv):
         assert response.status_code == 404
         data = response.json()
         assert data["code"] == "PGRST205"
+
+
+def test_second_request_for_non_existent_table_should_be_quick(defaultenv):
+    "requesting a non-existent relationship should be quick after the fuzzy search index is loaded (2nd request)"
+
+    env = {
+        **defaultenv,
+        "PGRST_DB_SCHEMAS": "fuzzysearch",
+        "PGRST_DB_POOL": "2",
+        "PGRST_DB_ANON_ROLE": "postgrest_test_anonymous",
+    }
+
+    with run(env=env, wait_max_seconds=30) as postgrest:
+        response = postgrest.session.get("/unknown-table")
+        assert response.status_code == 404
+        data = response.json()
+        assert data["code"] == "PGRST205"
+        first_duration = response.elapsed.total_seconds()
+        response = postgrest.session.get("/unknown-table")
+        assert response.elapsed.total_seconds() < first_duration / 10
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -170,7 +170,7 @@ dbActionPlan dbAct conf apiReq sCache = case dbAct of
 
 wrappedReadPlan :: QualifiedIdentifier -> AppConfig -> SchemaCache -> ApiRequest -> Bool -> Either Error CrudPlan
 wrappedReadPlan  identifier conf sCache apiRequest@ApiRequest{iPreferences=Preferences{..},..} headersOnly = do
-  qi <- findTable identifier (dbTables sCache)
+  qi <- findTable identifier sCache
   rPlan <- readPlan qi conf sCache apiRequest
   (handler, mediaType)  <- mapLeft ApiRequestError $ negotiateContent conf apiRequest qi iAcceptMediaType (dbMediaHandlers sCache) (hasDefaultSelect rPlan)
   if not (null invalidPrefs) && preferHandling == Just Strict then Left $ ApiRequestError $ InvalidPreferences invalidPrefs else Right ()
@@ -178,7 +178,7 @@ wrappedReadPlan  identifier conf sCache apiRequest@ApiRequest{iPreferences=Prefe
 
 mutateReadPlan :: Mutation -> ApiRequest -> QualifiedIdentifier -> AppConfig -> SchemaCache -> Either Error CrudPlan
 mutateReadPlan  mutation apiRequest@ApiRequest{iPreferences=Preferences{..},..} identifier conf sCache = do
-  qi <- findTable identifier (dbTables sCache)
+  qi <- findTable identifier sCache
   rPlan <- readPlan qi conf sCache apiRequest
   mPlan <- mutatePlan mutation qi apiRequest sCache rPlan
   if not (null invalidPrefs) && preferHandling == Just Strict then Left $ ApiRequestError $ InvalidPreferences invalidPrefs else Right ()
@@ -810,10 +810,10 @@ validateAggFunctions aggFunctionsAllowed (Node rp@ReadPlan {select} forest)
   | otherwise = Node rp <$> traverse (validateAggFunctions aggFunctionsAllowed) forest
 
 -- | Lookup table in the schema cache before creating read plan
-findTable :: QualifiedIdentifier -> TablesMap -> Either Error QualifiedIdentifier
-findTable qi@QualifiedIdentifier{..} tableMap =
-  case HM.lookup qi tableMap of
-    Nothing -> Left $ SchemaCacheErr $ TableNotFound qiSchema qiName (HM.elems tableMap)
+findTable :: QualifiedIdentifier -> SchemaCache -> Either Error QualifiedIdentifier
+findTable qi@QualifiedIdentifier{..} sc@SchemaCache{dbTables} =
+  case HM.lookup qi dbTables of
+    Nothing -> Left $ SchemaCacheErr $ TableNotFound qiSchema qiName sc
     Just _ -> Right qi
 
 addFilters :: ResolverContext -> ApiRequest -> ReadPlanTree -> Either Error ReadPlanTree
--- a/src/PostgREST/Response.hs
+++ b/src/PostgREST/Response.hs
@@ -213,10 +213,10 @@ actionResponse (MaybeDbResult InspectPlan{ipHdrsOnly=headersOnly} body) _ versio
   in
   Right $ PgrstResponse HTTP.status200 (MediaType.toContentType MTOpenAPI : cLHeader ++ maybeToList (profileHeader schema negotiatedByProfile)) rsBody
 
-actionResponse (NoDbResult (RelInfoPlan qi@QualifiedIdentifier{..})) _ _ _ SchemaCache{dbTables} _ _ =
+actionResponse (NoDbResult (RelInfoPlan qi@QualifiedIdentifier{..})) _ _ _ sc@SchemaCache{dbTables} _ _ =
   case HM.lookup qi dbTables of
     Just tbl -> respondInfo $ allowH tbl
-    Nothing  -> Left $ Error.SchemaCacheErr $ Error.TableNotFound qiSchema qiName (HM.elems dbTables)
+    Nothing  -> Left $ Error.SchemaCacheErr $ Error.TableNotFound qiSchema qiName sc
   where
     allowH table =
       let hasPK = not . null $ tablePKCols table in
--- a/test/io/fixtures/big_schema.sql
+++ b/test/io/fixtures/big_schema.sql
@@ -11375,12 +11375,34 @@ ALTER TABLE ONLY apflora.zielber
 
 ALTER TABLE apflora."user" ENABLE ROW LEVEL SECURITY;
 
+
+CREATE SCHEMA fuzzysearch;
+
+-- Create many tables to test fuzzy string search
+-- computing hints for non existing tables
+DO
+$$
+DECLARE
+  r record;
+BEGIN
+  FOR r IN
+    SELECT
+      format('CREATE TABLE fuzzysearch.unknown_table_%s ()', n) AS ct
+    FROM
+      generate_series(1, 499) n
+    LOOP
+      EXECUTE r.ct;
+    END LOOP;
+END
+$$;
+
 DROP ROLE IF EXISTS postgrest_test_anonymous;
 CREATE ROLE postgrest_test_anonymous;
 
 GRANT postgrest_test_anonymous TO :PGUSER;
 
 GRANT USAGE ON SCHEMA apflora TO postgrest_test_anonymous;
+GRANT USAGE ON SCHEMA fuzzysearch TO postgrest_test_anonymous;
 
 GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA apflora
 TO postgrest_test_anonymous;


steve-chavez:
Definitely looks related to https://github.com/PostgREST/postgrest/issues/4463. Have you tested the impact of this PR on a schema with lots of tables?

Looks like we'll need the same test setup for https://github.com/PostgREST/postgrest/issues/4462.

steve-chavez:
I think I've managed to repro, with 125K tables:

```
17/Nov/2025:17:56:40 -0500: Schema cache queried in 5230.2 milliseconds
17/Nov/2025:17:56:40 -0500: Schema cache loaded 125271 Relations, 232 Relationships, 149 Functions, 15 Domain Representations, 45 Media Type Handlers, 1196 Timezones
17/Nov/2025:17:56:40 -0500: Schema cache loaded in 237.2 milliseconds
```

The server dies when trying to compute the hint:

```bash
$ curl localhost:3000/xprojects

# correct table names do work
$ curl localhost:3000/projects
 {"id":3,"name":"IOS","client_id":2}, 
 {"id":5,"name":"Orphan","client_id":null}]$ 
```

steve-chavez:
> The server dies when trying to compute the hint:
> $ curl localhost:3000/xprojects
> curl: (52) Empty reply from server

Same result with this PR. I'll add a test case.

mkleczek:
> > The server dies when trying to compute the hint:
> > $ curl localhost:3000/xprojects
> > curl: (52) Empty reply from server
> 
> Same result with this PR. I'll add a test case.

Yeah... this PR won't help with the first hint computation (FuzzySet is created anyway). It would only help with subsequent hints (without this PR FuzzySet is created each time).

Seems like FuzzySet implementation is memory hungry.

Also - it depends on the number of schemas as with this PR there is a separate FuzzySet per schema created (doesn't help if all tables are in a single schema).

mkleczek:
@steve-chavez 
I've spent a little bit of time researching the problem and it seems we do not have any good options for databases with a large number of relations.

My suggestion would be to either:
1. Get rid of fuzzy relation search altogether.
2. Only search when the number of relations is small enough (we need to come up with the right number).
3. Have a configuration property to turn on/off fuzzy search.
4. Hit the database with dummy `SELECT 1 FROM non_existing_table LIMIT 0` or similar and retrieve information from the reported error (if there is no error that means the schema cache is stale).

Aside:
To be honest I have a gut feeling we do something wrong here trying to re-implement catalog search in PostgREST - this is the role of the DBMS. I am not even sure if https://github.com/PostgREST/postgrest/pull/3869 is the right direction as ideally we should be doing just syntactical transformations of requests to queries (and if possible not have the schema cache a all). I can even imagine the implementation that delegates query construction to the database itself (ie. have SQL generating SQL) and only caching them for subsequent executions.

OTOH I understand that if we already have the schema cache then let's use it - trade-offs everywhere :)

wolfgangwalther:
> 1. Get rid of fuzzy relation search altogether.

+1 for this also from a dependency perspective: We have to pin an older version of the fuzzy-thing dependency, because a newer version causes even more regression. I'd like to get rid of that dependency, ideally.

steve-chavez:
> Get rid of fuzzy relation search altogether.

From my experience maintaining a postgREST-as-a-service, before the fuzzy search hint there were lots of support tickets that came from typos but users were quick to to blame postgREST and suspect a malfunction. It was a huge time sink. After the fuzzy search, those complaints just stopped. 

Nowadays with typescript clients and other typed clients we have in the ecosystem, maybe this is not such a big problem but I'd still argue have the fuzzy search is a net win for 99% use cases.

> Only search when the number of relations is small enough (we need to come up with the right number).

This looks like the viable short-term solution. Most use cases don't have that many tables too.

> I am not even sure if https://github.com/PostgREST/postgrest/pull/3869 is the right direction as ideally we should be doing just syntactical transformations of requests to queries (and if possible not have the schema cache a all)

IIRC, #3869 was mostly for better error messages, but we have other features that might also reuse it. To not use the schema cache we'd have to implement resource embedding and other features in PostgreSQL itself (possible but would take a loot of time).

> +1 for this also from a dependency perspective: We have to pin an older version of the fuzzy-thing dependency, because a newer version causes even more regression. I'd like to get rid of that dependency, ideally.

Could we instead vendor the dependency? :thinking: (see https://github.com/laserpants/fuzzyset-haskell/issues/9)

steve-chavez:
> $ curl localhost:3000/xprojects

Besides the above, I'm also getting an "empty reply from server" on `/` (openAPI). I guess the empty replies happens because of the space leak protection https://github.com/PostgREST/postgrest/pull/2122?

>+1 for this also from a dependency perspective: We have to pin an older version of the fuzzy-thing dependency, because a newer version causes even more regression. I'd like to get rid of that dependency, ideally.

Looking at the algorithms used as inspiration on https://github.com/laserpants/fuzzyset-haskell/issues/2#issuecomment-562868913, I wonder if it's optimized for our use case. Maybe we can come up with a better library? (cc @taimoorzaeem)

mkleczek:
> > Get rid of fuzzy relation search altogether.
> 
> From my experience maintaining a postgREST-as-a-service, before the fuzzy search hint there were lots of support tickets that came from typos but users were quick to to blame postgREST and suspect a malfunction. It was a huge time sink. After the fuzzy search, those complaints just stopped.
> 

Makes sense.

[...] 
> > +1 for this also from a dependency perspective: We have to pin an older version of the fuzzy-thing dependency, because a newer version causes even more regression. I'd like to get rid of that dependency, ideally.
>
> Could we instead vendor the dependency? 🤔 (see [laserpants/fuzzyset-haskell#9](https://github.com/laserpants/fuzzyset-haskell/issues/9))

I think it makes sense to merge this PR regardless of these decisions as it is a quite obvious optimization (even though it does not fix the OOM issue).
@steve-chavez @wolfgangwalther WDYT?

steve-chavez:
> I think it makes sense to merge this PR regardless of these decisions as it is a quite obvious optimization (even though it does not fix the OOM issue).
@steve-chavez @wolfgangwalther WDYT?

I don't think we should merge without a test that proves what's being improved (since it doesn't solve #4463).

taimoorzaeem:
> Looking at the algorithms used as inspiration on [laserpants/fuzzyset-haskell#2 (comment)](https://github.com/laserpants/fuzzyset-haskell/issues/2#issuecomment-562868913), I wonder if it's optimized for our use case. Maybe we can come up with a better library? (cc @taimoorzaeem)

I will be happy to write a library :+1: . I would just need to do some feasibility, maybe a better library already exist on hackage? Needs some investigation.

mkleczek:
> > I think it makes sense to merge this PR regardless of these decisions as it is a quite obvious optimization (even though it does not fix the OOM issue).
> > @steve-chavez @wolfgangwalther WDYT?
> 
> I don't think we should merge without a test that proves what's being improved (since it doesn't solve #4463).

Done.
Updated mixed load test to include requests for non-existing relation and the fixture to create 1000 tables.

See: https://github.com/PostgREST/postgrest/actions/runs/19477317041/attempts/1#summary-55740304508

mkleczek:
> > Looking at the algorithms used as inspiration on [laserpants/fuzzyset-haskell#2 (comment)](https://github.com/laserpants/fuzzyset-haskell/issues/2#issuecomment-562868913), I wonder if it's optimized for our use case. Maybe we can come up with a better library? (cc @taimoorzaeem)
> 
> I will be happy to write a library 👍 . I would just need to do some feasibility, maybe a better library already exist on hackage? Needs some investigation.

As I mentioned earlier - I've done some research and I couldn't find any interesting alternatives. Fuzzy search algorithms are divided into two subgroups: online and offline.

Online algorithms do not require additional memory but require scanning the whole list (so are `O(n)` at least).

Offline algorithms require creating an index. Best results are achieved using indexes based on n-grams - this is exactly what `FuzzySet` we use is. The problem with n-grams based index is that it is big (for each word we keep multiple entries: one per n-gram found in the word).

I am skeptical we can come up with a solution to 125k tables...

steve-chavez:
> Done.
> Updated mixed load test to include requests for non-existing relation and the fixture to create 1000 tables.
> See: https://github.com/PostgREST/postgrest/actions/runs/19477317041/attempts/1#summary-55740304508

@mkleczek We should not modify the mixed loadtest for this, that should stay the same across versions so we can see regressions. Additionally, that doesn't prove there's an enhancement here.

Test should be like:

- A new dedicated fixture (like we have for`test/io/replica.sql`).
- A test that triggers a not found error
- The test should measure there's a speed up via `Server-Timing` or other means (for this it might help to increase the number of tables to a big number)

mkleczek:
> > Done.
> > Updated mixed load test to include requests for non-existing relation and the fixture to create 1000 tables.
> > See: https://github.com/PostgREST/postgrest/actions/runs/19477317041/attempts/1#summary-55740304508
> 
> @mkleczek We should not modify the mixed loadtest for this, that should stay the same across versions so we can see regressions.

I am not sure I understand this.

The old `mixed` loadtest does not account for this scenario, and without it it is not a good regression test (the proof is that we didn't catch performance regression introduced by [#3869](https://github.com/PostgREST/postgrest/pull/3869) )

So I would say adding the "table not found" scenario to it is an important addition - it makes it more realistic and covering wider set of scenarios.

>Additionally, that doesn't prove there's an enhancement here.
 
https://github.com/PostgREST/postgrest/actions/runs/19477317041/attempts/1#summary-55740304508
shows more than 20% increase in throughput (108 vs 82 req/s) and around 10% less CPU usage comparing to `main` and `14`.

Am I missing something here?

> Test should be like:
> 
> * A new dedicated fixture (like we have for`test/io/replica.sql`).
> * A test that triggers a not found error
> * The test should measure there's a speed up via `Server-Timing` or other means (for this it might help to increase the number of tables to a big number)

Hmm...
Do we run `io` tests against different branches (similar to load tests)?
If not then I don't really understand how we could show improvement vs `main` or `14`.

I thought load tests (not `io` tests) are supposed to test performance?

@steve-chavez could you advice on the way you would like it to be tested?

steve-chavez:
> I am not sure I understand this.
> The old mixed loadtest does not account for this scenario, and without it it is not a good regression test (the proof is that we didn't catch performance regression introduced by https://github.com/PostgREST/postgrest/pull/3869 )

The problem is that "mixed" is already conflating too many scenarios (see https://github.com/PostgREST/postgrest/issues/4123), with this addition is even worse as we also conflate "successful traffic" with "error traffic" (note that now we have `errors.0: 404 Not Found` in the loadtest results).

Another problem is that loadtests are slow, so we shouldn't add too many of them (IMO a dedicated one for errors doesn't seem worth it) if we can instead have an io test.

> Hmm...
> Do we run io tests against different branches (similar to load tests)?
> If not then I don't really understand how we could show improvement vs main or 14.
> I thought load tests (not io tests) are supposed to test performance?

We do have  perf related tests on:

https://github.com/PostgREST/postgrest/blob/91abcd49e12b2893a8e61234a0dbdf433baa94aa/test/io/test_big_schema.py#L10-L36

It looks enough to have an upper bound of expected time for computing the hint to not have regressions.

> @steve-chavez could you advice on the way you would like it to be tested?

I'd suggest adding the fixtures on `test/io/test_big_schema.py` too, it seems fit. Roughly following the steps I mentioned [above](https://github.com/PostgREST/postgrest/pull/4472#issuecomment-3553559760), I'm thinking measuring `Server-Timing` should be enough. You could also add this test on another PR, to see the current server-timing numbers and then updating the numbers on this PR, to confirm the improvement.

mkleczek:
> > @steve-chavez could you advice on the way you would like it to be tested?
> 
> I'd suggest adding the fixtures on `test/io/test_big_schema.py` too, it seems fit. Roughly following the steps I mentioned [above](https://github.com/PostgREST/postgrest/pull/4472#issuecomment-3553559760), I'm thinking measuring `Server-Timing` should be enough. You could also add this test on another PR, to see the current server-timing numbers and then updating the numbers on this PR, to confirm the improvement.

@steve-chavez
Done. (see `test_second_request_for_non_existent_table_should_be_quick` in test_big_schema.py)

test/io/test_big_schema.py
```diff
@@ -70,3 +70,23 @@ def test_should_not_fail_with_stack_overflow(defaultenv):
...
+    }
+
+    with run(env=env, wait_max_seconds=30) as postgrest:
+        response = postgrest.session.get("/unknown-table?select=unknown-rel(*)")
+        assert response.status_code == 404
+        data = response.json()
+        assert data["code"] == "PGRST205"
+        first_duration = response.elapsed.total_seconds()
+        response = postgrest.session.get("/unknown-table?select=unknown-rel(*)")
+        assert response.elapsed.total_seconds() < first_duration / 20
```
steve-chavez: Is it necessary to use resource embedding?

```suggestion
        response = postgrest.session.get("/unknown-table")
        assert response.status_code == 404
        data = response.json()
        assert data["code"] == "PGRST205"
        first_duration = response.elapsed.total_seconds()
        response = postgrest.session.get("/unknown-table")
        assert response.elapsed.total_seconds() < first_duration / 20
```

test/io/test_big_schema.py
```diff
@@ -70,3 +70,23 @@ def test_should_not_fail_with_stack_overflow(defaultenv):
         assert response.status_code == 404
         data = response.json()
         assert data["code"] == "PGRST205"
+
+
+def test_second_request_for_non_existent_table_should_be_quick(defaultenv):
+    "requesting a non-existent relationship the second time should be quick"
```
steve-chavez: Q: Why is the second time quicker? Does the fuzzy index get populated after hitting an error?

From the code, it looks like the fuzzy index is only populated when the schema cache is built.

mkleczek: > Q: Why is the second time quicker? Does the fuzzy index get populated after hitting an error?
> 
> From the code, it looks like the fuzzy index is only populated when the schema cache is built.

That's because of Haskell laziness.

test/io/test_big_schema.py
mkleczek: > Is it necessary to use resource embedding?

No, copy/paste leftover. Fixed.

test/io/test_big_schema.py
steve-chavez: Ok, then to avoid confusion I suggest:

```suggestion
    "requesting a non-existent relationship should be quick after the schema cache is loaded (2nd request)"
```

steve-chavez: On second thought, not sure if I understand this.

When we do a request with resource embedding the schema cache is already loaded and works on the first request. Why does this change for the fuzzy index and we need to make 2 requests?

src/PostgREST/Error.hs
```diff
@@ -250,7 +248,7 @@ data SchemaCacheError
   | NoRelBetween Text Text (Maybe Text) Text RelationshipsMap
   | NoRpc Text Text [Text] MediaType Bool [QualifiedIdentifier] [Routine]
   | ColumnNotFound Text Text
-  | TableNotFound Text Text [Table]
+  | TableNotFound Text Text (HM.HashMap Schema Fuzzy.FuzzySet)
```
steve-chavez: How about a type alias for this, documenting how it works?

test/io/test_big_schema.py
steve-chavez: Maybe what we need is some comments about this optimization, I suggest adding those on the type definition https://github.com/PostgREST/postgrest/pull/4472#discussion_r2550587447

mkleczek: > On second thought, not sure if I understand this.
> 
> When we do a request with resource embedding the schema cache is already loaded and works on the first request. Why does this change for the fuzzy index and we need to make 2 requests?

This is indeed tricky: schema cache is loaded lazily as well. But there are these two lines in `AppState.retryingSchemaCacheLoad`:
```
(t, _) <- timeItT $ observer $ SchemaCacheSummaryObs $ showSummary sCache
observer $ SchemaCacheLoadedObs t

```
which cause evaluation of `SchemaCache` fields.

I've decided not to add `dbTablesFuzzyIndex` to schema cache summary and leave its evaluation till first use.

src/PostgREST/Error.hs
mkleczek: > How about a type alias for this, documenting how it works?

Yeah, type alias is a good idea but I think comments about lazy evaluation of `SchemaCache. dbTablesFuzzyIndex` should be placed at the field definition (and not the type alias).

test/io/test_big_schema.py
mkleczek: Introduced the type alias, added comments to SchemaCache field and updated the test description.

src/PostgREST/Error.hs
```diff
@@ -428,12 +427,12 @@ noRpcHint schema procName params allProcs overloadedProcs =
 
 -- |
 -- Do a fuzzy search in all tables in the same schema and return closest result
-tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
-tableNotFoundHint schema tblName tblList
+tableNotFoundHint :: Text -> Text -> HM.HashMap Schema Fuzzy.FuzzySet -> Maybe Text
```
taimoorzaeem: Now, that we have a type alias, I think we should use it everywhere for consistency. I think you missed it here:
```suggestion
tableNotFoundHint :: Text -> Text -> TablesFuzzyIndex -> Maybe Text
```

mkleczek: > Now, that we have a type alias, I think we should use it everywhere for consistency. I think you missed it here:

Good catch, thx!
Fixed.

test/io/test_big_schema.py
```diff
@@ -70,3 +70,23 @@ def test_should_not_fail_with_stack_overflow(defaultenv):
...
+    }
+
+    with run(env=env, wait_max_seconds=30) as postgrest:
+        response = postgrest.session.get("/unknown-table")
+        assert response.status_code == 404
+        data = response.json()
+        assert data["code"] == "PGRST205"
+        first_duration = response.elapsed.total_seconds()
+        response = postgrest.session.get("/unknown-table")
+        assert response.elapsed.total_seconds() < first_duration / 20
```
taimoorzaeem: I noticed something strange on my system. When I run the test with get endpoint changed to `/table-with-a-weird-name`, the test fails with an exception.

```
FAILED test/io/test_big_schema.py::test_second_request_for_non_existent_table_should_be_quick 
- requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected(
- 'Remote end closed connection without response'))
```

mkleczek: @taimoorzaeem I'm afraid the same thing happens when running this test against `main`.

This is expected as this PR only makes sure the FuzzySet is created once instead of every time hint is calculated. Hint calculation logic and the data structures stay the same.

IMHO it looks like the library we use for fuzzy search has reliability issues and we should look for other solutions.

mkleczek: @taimoorzaeem I found https://hackage-content.haskell.org/package/fuzzily-0.2.1.0/docs/Text-Fuzzily.html and https://hackage.haskell.org/package/fuzzyfind-3.0.2/docs/Text-FuzzyFind.html but they both implement _online_ fuzzy search. For us means that hint calculation time is at least linear in the number of tables (which I don't think is a good idea as it will almost certainly fail with timeout for large schemas).

taimoorzaeem: I think the most scalable solution would be to implement the `SymSpell` algorithm. See: [ref1](https://github.com/wolfgarbe/SymSpell), [ref2](https://seekstorm.com/blog/1000x-spelling-correction/), [ref3](https://ieeexplore.ieee.org/document/9678171).

I think a haskell package for this would be great for the entire haskell community.

mkleczek: > I think the most scalable solution would be to implement the `SymSpell` algorithm. See: [ref1](https://github.com/wolfgarbe/SymSpell), [ref2](https://seekstorm.com/blog/1000x-spelling-correction/), [ref3](https://ieeexplore.ieee.org/document/9678171).
> 

The problem with SymSpell is that it is memory hungry (ie. for 100,000 words and edit distance of 2 it requires 1,500,000 entries in the dictionary). Building the index is also very costly.
You can minimize memory requirements with perfect hashing but it makes building the index even more costly.

In general, this is a tough problem. It is even tougher if the spelling dictionary is dynamic (relation names in the schema cache can be reloaded and require index rebuilding).

> I think a haskell package for this would be great for the entire haskell community.

Undoubtedly. You can read about SOTA techniques for example here: https://towardsdatascience.com/spelling-correction-how-to-make-an-accurate-and-fast-corrector-dc6d0bcbba5f/
Implementing this in Haskell would be a very interesting task.

Having said that, when you think about the whole architecture from the high level, PostgREST does not seem to be the right place to implement in-memory text search engine. Especially that it is a component that is supposed to be scaled easily (ie. start new instances quickly). In such scenarios you want to externalize state, ie. have a separate component implementing text search algorithms. But you already have such a component: PostgreSQL itself!

IMHO these are possible paths for PostgREST:

1. Stay with current architecture (ie. query building based on in-memory schema cache outside of the database) and implement limited but cheap spell checking.
2. Move query building to the database itself and use database facilities to handle spell checking (somewhat not a viable solution as it would be a complete rewrite).
3. Stay with current architecture and externalize spell checking
   - by invoking external system when calculating hint (there are multiple questions here: should it be some specialized search engine - we don't want to have such a dependency. should it be PostgreSQL - then see below)
   - by letting PostgreSQL handle misspelled relations (that in essence means reverting https://github.com/PostgREST/postgrest/pull/3869)

It just seems to me we've just hit the wall with our current architecture here and the only thing we can do is to admit it and live with deficiency in spell checking or re-architect and rewrite PostgREST.

One mitigation would be to work on `fuzzyset` library and try to optimize it as much as possible (I've taken a quick look at the source and found some minor optimization opportunities). The question is really: what schema sizes are "normal" for PostgREST and what sizes are outside of what PostgREST supports?

@taimoorzaeem @steve-chavez does the above make sense?

taimoorzaeem: Great analysis Michal! Many options to go from here, but I think the first thing we need to do is to have reproducible production level testing setup, only then we can take a sure fire decision.

steve-chavez: Sorry for the late reply here.
 
> by letting PostgreSQL handle misspelled relations

I gave this a shot in pg ([ref](https://www.postgresql.org/message-id/flat/CAGRrpzaBQdPdSRXwHnLnCvs%3Df1a1abcecgBq5f5s%2BXPGU7f_zQ%40mail.gmail.com)), it was much harder than I expected. If the patch gets committed, maybe we should deprecate the fuzzy matching (possibly introducing https://github.com/PostgREST/postgrest/issues/4499) and then drop the feature.

> Move query building to the database itself

This is a direction we've discussed before (see https://github.com/PostgREST/postgrest/pull/1511), the main problem is that we'd lose one our main advantages: statelessness. This because we'd have to create database objects.

@mkleczek Are you envisioning the above to work by creating `plpgsql` functions?

If we could embed SQL without creating functions that could work, we've discussed doing this to embed https://github.com/PostgREST/postgrest-openapi and replace our current Haskell-based openapi.

TBH I would still like to make the above work somehow.

steve-chavez: > I gave this a shot in pg (ref)

@mkleczek How about using a similar logic here and capping the amount of searches? That would certainly fix https://github.com/PostgREST/postgrest/issues/4463. I believe computing the hint is mostly useful for early stages of a project anyway.

mkleczek: > Sorry for the late reply here.
> 
> > by letting PostgreSQL handle misspelled relations
> 
> I gave this a shot in pg ([ref](https://www.postgresql.org/message-id/flat/CAGRrpzaBQdPdSRXwHnLnCvs%3Df1a1abcecgBq5f5s%2BXPGU7f_zQ%40mail.gmail.com)), it was much harder than I expected. If the patch gets committed, maybe we should deprecate the fuzzy matching (possibly introducing #4499) and then drop the feature.
> 

Thanks for the information and reference: very interesting!

It means that Pg does not provide hinting we need and  it faces the same issue as we do (namely: catalog might be huge and that makes hinting infeasible).
I wasn't aware hinting is only `psql` functionality and it is not available in Postgres itself. That makes the discussion moot, I guess :)

There is also a very interesting comment in this thread (at the very end of https://www.postgresql.org/message-id/6B5665E6-14A0-4BA2-97E5-B3A5478490E2%40yesql.se ) which begs the question: do we need to check relation permissions when hinting?

> > Move query building to the database itself
> 
> This is a direction we've discussed before (see #1511), the main problem is that we'd lose one our main advantages: statelessness. This because we'd have to create database objects.
> 
> @mkleczek Are you envisioning the above to work by creating `plpgsql` functions?

Not really. My idea was to have the following pipeline:

request -> construct `query building query` (`qbq`) -> execute `qbq` and retrieve `target query` -> execute `target query`

I guess`qbq` can be static and it really does not matter if it is embedded in the application or deployed in a SQL function.

Target queries must be cached to avoid repeated building. The `target query cache` is a lazily populated analog of current schema cache in this architecture.

src/PostgREST/Error.hs
```diff
@@ -426,14 +426,20 @@ noRpcHint schema procName params allProcs overloadedProcs =
...
+maxDbTablesForFuzzySearch :: Int
+maxDbTablesForFuzzySearch = 4000
+
 -- |
 -- Do a fuzzy search in all tables in the same schema and return closest result
-tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
-tableNotFoundHint schema tblName tblList
-  = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
+tableNotFoundHint :: Text -> Text -> SchemaCache -> Maybe Text
+tableNotFoundHint schema tblName SchemaCache{dbTablesSize, dbTablesFuzzyIndex}
```
taimoorzaeem:
```suggestion
tableNotFoundHint :: Text -> Text -> Int -> TablesFuzzyIndex -> Maybe Text
tableNotFoundHint schema tblName totalTables fuzzyset
```

It seems a bit out of place to pass the entire `SchemaCache` to this function. Maybe this looks cleaner?

src/PostgREST/Error.hs
```diff
@@ -250,7 +250,7 @@ data SchemaCacheError
   | NoRelBetween Text Text (Maybe Text) Text RelationshipsMap
   | NoRpc Text Text [Text] MediaType Bool [QualifiedIdentifier] [Routine]
   | ColumnNotFound Text Text
-  | TableNotFound Text Text [Table]
+  | TableNotFound Text Text SchemaCache
```
taimoorzaeem, line 253:
```suggestion
  | TableNotFound Text Text Int TablesFuzzyIndex
```

Similar to https://github.com/PostgREST/postgrest/pull/4472#discussion_r2596117162.

src/PostgREST/SchemaCache.hs
```diff
@@ -66,21 +67,31 @@ import PostgREST.SchemaCache.Table           (Column (..), ColumnMap,
...
+  , dbMediaHandlers    :: MediaHandlerMap
+  , dbTimezones        :: TimezoneNames
+  -- Memoized size of dbTables
+  -- HashMap.size is O(n) so we store the size once computed
+  , dbTablesSize       :: Int
+  -- Memoized fuzzy index of table names per schema to support approximate matching
+  -- Since index construction can be expensive, we build it once and store in the SchemaCache
+  -- Haskell lazy evaluation ensures it's only built on first use and memoized afterwards
+  , dbTablesFuzzyIndex :: TablesFuzzyIndex
+  } deriving (Show)
```
taimoorzaeem: This looks fine overall :+1:. I would add a `TODO` comment to refactor `dbTablesSize` and `dbTablesFuzzyIndex` some other way later. I think they don't belong to the `SchemaCache` datatype.

src/PostgREST/Error.hs
mkleczek: @taimoorzaeem 

The original (before this PR) was `TableNotFound Text Text [Table]`.

Changing it to `TableNotFound Text Text TablesFuzzyIndex` in the first version of the patch required sweeping changes in several modules (`Plan` and `Response`).
Changing it to `TableNotFound Text Text Int TablesFuzzyIndex` to support conditional hinting calculation required sweeping changes in several modules again.

I've decided to pass `SchemaCache` here because `Plan` and `Response` should be decoupled from implementation of the hinting algorithm (in other words: why should `Plan` and `Response` modules care how error hints are calculated and what is required to do that?).

Ideally `Plan` and `Response` modules should be completely unaware of any metadata required to calculate hint (ie. it should be `TableNotFound Text Text`) and I tried to implement that. Unfortunately the change would need to be very big because of the way `hint` function in `ErrorBody` class  is defined. I guess it should be a separate refactoring if we want to do it.

I remember some discussions about having application monad. It is also hinted in https://github.com/PostgREST/postgrest/issues/1804. Then we could, for example, add necessary constraint to `ErrorBody` class  (ie. `MonadReader SchemaCache m`)

For now I would leave `TableNotFound Text Text SchemaCache` to eliminate tight coupling between `Plan`, `Response` and hint calculation.

WDYT?

test/io/test_big_schema.py
taimoorzaeem: > I noticed something strange on my system. When I run the test with get endpoint changed to `/table-with-a-weird-name`, the test fails with an exception.
> 
> ```
> FAILED test/io/test_big_schema.py::test_second_request_for_non_existent_table_should_be_quick 
> - requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected(
> - 'Remote end closed connection without response'))
> ```

@mkleczek This test is still failing. How about we only just add `< 4000` constraint in this PR? The performance gain with `< 4000` is negligible even with the fuzzy set cache.

mkleczek: @taimoorzaeem 

1. Actually, on my machine the max number of tables that `FuzzySet` can handle is 950. So 4k is waaaay too high (I forgot about it in earlier discussions)
2. The performance gain is actually quite substantial but can only be measured once you enhance `mixed` load test with "missing table scenario". See https://github.com/PostgREST/postgrest/actions/runs/19477317041/attempts/1#summary-55740304508 (upon @steve-chavez request I've reverted the changes to `mixed` load test)

src/PostgREST/SchemaCache.hs
mkleczek: @taimoorzaeem 

What is our policy about leaving TODOs / FIXMEs in the code? I know a lot of projects disallows them preferring to simply report issues in the bug tracker (TODOs in code are not really tracked and become noise).

taimoorzaeem:
@mkleczek @steve-chavez Opened #4523. I think let's fix the bug first because it is impacting our users. We can then work on the performance improvements as they might require more time and work.

src/PostgREST/SchemaCache.hs
taimoorzaeem: As far as I am aware, we don't have a strict policy on this. It would be too much to track if we open an issue for every TODO. Most of our TODOs or FIXMEs are mostly code improvements. I regularly check them by grepping `TODO` and `FIXME`.

mkleczek:
> @mkleczek @steve-chavez Opened #4523. I think let's fix the bug first because it is impacting our users. We can then work on the performance improvements as they might require more time and work.

@taimoorzaeem @steve-chavez - see my comment in #4523 https://github.com/PostgREST/postgrest/pull/4523/files#r2598153417 - such deceiptivly simple "fix" introduces Denial of Service vulnerability in PostgREST.

mkleczek:
@taimoorzaeem @steve-chavez 
I moved limit check to `SchemaCache`. That way we check the limit upon schema (re)loading and make sure `FuzzySet` loading is not triggered anywhere accidentally.

mkleczek:
@steve-chavez @taimoorzaeem 

It would be good to merge this. As discussed in https://github.com/PostgREST/postgrest/pull/4523/changes#r2598153417 - leaving the thing as it is implemented right now is not only affecting stability for large schemas but is a DoS vulnerability (an attacker can request random table names to overload the system).

steve-chavez:
@mkleczek Sorry for the late reply here.

Does this fix both https://github.com/PostgREST/postgrest/issues/4462 and https://github.com/PostgREST/postgrest/issues/4463? If so, maybe we should change the commit prefix to `fix:` and also add an entry to the CHANGELOG?

> . As discussed in https://github.com/PostgREST/postgrest/pull/4523/changes#r2598153417 - leaving the thing as it is implemented right now is not only affecting stability for large schemas but is a DoS vulnerability (an attacker can request random table names to overload the system).
> Re-creating FuzzySet upon each request opens up yet another DoS vulnerability.

It does look like the outcome should fix + patch release. Could we add a test or loadtest for the above?

mkleczek:
> 
> It does look like the outcome should fix + patch release. Could we add a test or loadtest for the above?

@steve-chavez 
We have the test case in big schema tests suite. Updates to `mixed` load tests were reverted upon your request. Not sure what test you have in mind.

taimoorzaeem:
How about a loadtest similar to what I mentioned here: https://github.com/PostgREST/postgrest/pull/4523#issuecomment-3626033233?

> According to my analysis, max `500` number of tables works best for fuzzy search even for `24` character long hashes as table names.

So, maybe the `Loadtest (fuzzy-search)` should be something like:

1. Create maximum `500` tables with random table names (`30` character hashes).
2. Then, send some number of requests (I am not sure what the usual is for other load tests), where the table names are random (`30` character hashes).
3. Then, measure the performance?

There is no reason for it to be `30` characters long. The problem that occurred in #4463 had names `30` characters long, so maybe we should do that?

```
"GET /api/resources/c13d05e5-9e55-4d03-bf7e-042a2ade7e49/data/?page_size=1&DATE_DEBUT__sort=desc&P_SOUSCRITE__exact=6 HTTP/1.1"
```

@steve-chavez @mkleczek WDYT?

mkleczek:
> How about a loadtest similar to what I mentioned here: [#4523 (comment)](https://github.com/PostgREST/postgrest/pull/4523#issuecomment-3626033233)?
> 

@steve-chavez @taimoorzaeem 
- the problem with fuzzy search is obviously affecting the users and has been well analyzed in several discussions
- It is apparent that it is not possible to fix the issue in general and we don't have any other choice than provide mitigations
- performance issues in parts of code responsible for "not found" error handling are very problematic because they are DoS vulnerabilities - closing these vulnerabilities should be a top priority (TBH they are good candidates for CVEs which would be a shame because PostgREST has clean record in CVE databases)
- this PR is straightforward and obviously improves the situation (while not providing a general solution to fuzzy search it addresses deficiencies in current code) - at least it closes DoS vulnerability.

So while I am open to add some specialized load test for non-existent database objects, I think it is not necessary condition to merge this PR.

I would say it is the other way around - it should be merged urgently and - what's more - it should be backported to 14.x.

@steve-chavez @taimoorzaeem my 2 cents

steve-chavez:
Added a loadtest in https://github.com/PostgREST/postgrest/pull/4576, that should help prove #4462 is solved here.

@mkleczek Could you rebase?

mkleczek:
> Added a loadtest in #4576, that should help prove #4462 is solved here.
> 
> @mkleczek Could you rebase?

Done. See https://github.com/PostgREST/postgrest/actions/runs/20556133512/attempts/1#summary-59040541838

taimoorzaeem:
> Done. See https://github.com/PostgREST/postgrest/actions/runs/20556133512/attempts/1#summary-59040541838

From results, this looks like a massive improvement. :rocket:

mkleczek:
> > Done. See https://github.com/PostgREST/postgrest/actions/runs/20556133512/attempts/1#summary-59040541838
> 
> From results, this looks like a massive improvement. 🚀

I haven't yet dig into the test implemented by @steve-chavez but such huge speedup might be misleading. We have a 500 limit on the number of tables that we index in the FuzzySet. So possibly we just don't calculate any hint in the test hence such a difference.

mkleczek:
> > > Done. See https://github.com/PostgREST/postgrest/actions/runs/20556133512/attempts/1#summary-59040541838
> > 
> > 
> > From results, this looks like a massive improvement. 🚀
> 
> I haven't yet dig into the test implemented by @steve-chavez but such huge speedup might be misleading. We have a 500 limit on the number of tables that we index in the FuzzySet. So possibly we just don't calculate any hint in the test hence such a difference.

@steve-chavez @taimoorzaeem 

I've run `postgrest-loadtest-against -k errors main` with the number of tables in `errors.sql` set to `100` (to make sure we are actually computing the hint) and the speedup is massive as well.

What do you think about changing the number of tables in `errors.sql` to `490` (instead of current `20000`)?

steve-chavez:
> From results, this looks like a massive improvement. 🚀

:rocket: Although the CPU usage is almost x3 on the `errors` load test unlike the `mixed` one, I wonder why that is?

> What do you think about changing the number of tables in errors.sql to 490 (instead of current 20000)?

The problem is that just using 490 doesn't prove the high memory usage reported on https://github.com/PostgREST/postgrest/issues/4462. I think we should merge as is and call #4462 solved.

steve-chavez:
> The problem is that just using 490 doesn't prove the high memory usage reported on https://github.com/PostgREST/postgrest/issues/4462

See the [load test result](https://github.com/PostgREST/postgrest/actions/runs/20556133512/attempts/1#summary-59040541838) which shows `~356` MB usage on non-head branches vs the head branch which reaches ~77 MB.

taimoorzaeem:
> Although the CPU usage is almost x3 on the `errors` load test unlike the `mixed` one, I wonder why that is?

Yeah, that's strange. I examined this and I think that's because on `mixed` ones, we  also have have `OPTIONS` and a `GET` on root endpoint which make the overall CPU usage looks less. After removing these two cases and re-running the `mixed` ones, I noticed that the CPU usage is actually quite similar to what we have in the `errors` test.

steve-chavez:
>  we also have have OPTIONS and a GET on root endpoint which make the overall CPU usage looks less. After removing these two cases and re-running the mixed ones, I noticed that the CPU usage is actually quite similar to what we have in the errors test.

Great find. Noted this for later https://github.com/PostgREST/postgrest/issues/4123#issuecomment-3704200611.

> it should be backported to 14.x.

@mkleczek Agree. Let's turn this into a `fix:` commit then?

mkleczek:
> > it should be backported to 14.x.
> 
> @mkleczek Agree. Let's turn this into a `fix:` commit then?

Done.

mkleczek:
> Looks good! We should also add an entry in the `CHANGELOG.md`.

Done.

steve-chavez (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file. From versio
 - Fix not returning `Content-Length` on empty HTTP `201` responses by @laurenceisla in #4518
 - Fix inaccurate Server-Timing header durations by @steve-chavez in #4522
 - Fix inaccurate "Schema cache queried" logs by @steve-chavez in #4522
+- Fix performance and high memory usage of relation hint calculation by @mkleczek in #4462 #4463
```
steve-chavez, line 23: Was just going to release v14 then realized this changelog entry was put in the wrong spot, messing up our `postgrest-release` command.

This was already fixed on https://github.com/PostgREST/postgrest/pull/4589, but please be mindful of the CHANGELOG placement @mkleczek 🙏

test/io/test_big_schema.py
```diff
@@ -70,3 +70,23 @@ def test_should_not_fail_with_stack_overflow(defaultenv):
...
+    }
+
+    with run(env=env, wait_max_seconds=30) as postgrest:
+        response = postgrest.session.get("/unknown-table")
+        assert response.status_code == 404
+        data = response.json()
+        assert data["code"] == "PGRST205"
+        first_duration = response.elapsed.total_seconds()
+        response = postgrest.session.get("/unknown-table")
+        assert response.elapsed.total_seconds() < first_duration / 10
```
taimoorzaeem, line 92: This test is flaky and fails often both locally and in [CI](https://github.com/PostgREST/postgrest/actions/runs/20740716255/job/59546810828?pr=4584) which is not desirable. If the point is to prove that the second request is always faster, I think we should remove the `/ 10` part?

steve-chavez: > I think we should remove the / 10 part?

That looks enough to remove the flakiness. @mkleczek WDYT?

mkleczek: > > I think we should remove the / 10 part?
> 
> That looks enough to remove the flakiness. @mkleczek WDYT?

I would experiment with different dividers so that we test that the second request is really faster. I guess 2 might be safe (on my machine 20 was working fine, made it to 10 but it seems it might be lower still).

OTOH I am wondering if this test is really needed since we have load tests to measure error response scenarios.

taimoorzaeem: > OTOH I am wondering if this test is really needed since we have load tests to measure error response scenarios.

It is a regression test so it is good to keep this test. The test proves and also in a way documents that we build a fuzzy set on the very first request and so that request is slower than the following requests which are "much" faster. 

> I guess 2 might be safe.

Seems good to me. 👍

Issue #3391, opened 2024-04-09
Content-type not set when function returning a domain representing media type is requested with Accept "*/*" header


mkleczek:
### Environment

* PostgreSQL version: 16
* PostgREST version: 12.0.2
* Operating system: Windows

### Description of issue

I have the following function:
```
CREATE DOMAIN "text/javascript" AS text;
CREATE OR REPLACE FUNCTION "module.js"() RETURNS "text/javascript"...
```

When requested by a browser from:

```
<script type="module" src="/rpc/module.js"></script>
```

returned response Content-Type is `application/json`

It seems the browser is setting Accept header to `*/*` and PostgREST does not set proper content type in this case.
As a workaround it is necessary to explicitly set content type in the function using current_setting.

steve-chavez:
This needs the any handler as mentioned on https://github.com/PostgREST/postgrest/discussions/3387#discussioncomment-9192985

steve-chavez:
I agree that this is the ideal behavior and I gave it a shot on https://github.com/steve-chavez/postgrest/commit/760f5ac08d2f7583901eb9593fa865c2ad1a85ac. Adding support for this is simple enough but it's hard to retrofit it with the current table media handlers done with aggregates, many tests break.

Likely we will need to adjust aggregates behavior and cause breaking changes, so I will leave this for a next major version.

steve-chavez:
Yeah, maybe it can be done purely on the Plan side. I mostly was trying to keep that module simple (since it's already pretty complex) by adding extra info on the schema cache.

taimoorzaeem:
@steve-chavez @laurenceisla I am still not sure what should be the correct Content-Type that should be returned here. I think the correct Mediatype would be `text/javascript`. I understand that we need to access this return type from schema cache and return it as the Content-Type. Currently when `Accept: */*`, it returns `Content-Type: application/json` which is because of following code line:
```haskell
HM.insert (RelAnyElement, MediaType.MTAny) (BuiltinOvAggJson, MediaType.MTApplicationJSON) -- SchemaCache.hs:1137:
```
My question is, how are we gonna know that there is domain which is also a mediatype that we need to return? Sorry if that doesn't make any sense.

mkleczek:
@taimoorzaeem - looks like it is not "beginner" difficulty as it requires quite a bit of refactoring to content negotiation  code.

PR #4498, opened 2025-11-24
[WIP] fix: content-type not set with custom domain and accept `*/*`

taimoorzaeem:
BREAKING CHANGE

- [ ] Implementation
- [ ] Tests
- [ ] Docs

EDIT: Unlink the associated issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. From versio
 
 ## Unreleased
 
+## Fixed
+
+- Fix `Content-Type` not set with custom domain type and `Accept: */*` by @taimoorzaeem in #3391
+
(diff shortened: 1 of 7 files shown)


taimoorzaeem:
This is an attempt to fix #3391. About 6 spec tests break because according to this fix, they are a bug.

For the solution, I have added a `*/*` mapping for each resolved media handler type. So, now it would work for both say `Accept: text/javascript` and `Accept: */*`. All other solutions that I tried were not as scalable and performant.

Looking for feedback on this.

mkleczek:
@taimoorzaeem 

Did you look at https://hackage.haskell.org/package/http-media library? It looks to have fairly complete content negotiation implementation.

taimoorzaeem:

PR #4501, opened 2025-11-24
chore: fix small typo in README.md

polmabri:
Fix small typo in README.md


taimoorzaeem:
Hmm, not strictly a typo, but this is fine too I guess.

Normally, this wouldn't be considered a `docs:` change, so I think it should be prefixed with `chore:` instead.

PR #4502, opened 2025-11-25
nix(shell): remove `postgrest/` directory prefix when running pg

taimoorzaeem:
When running postgres from nix-shell, nix creates a directory structure like `postgrest/postgrest-with-pg-17-XXX` in the `/tmp` directory. This commit removes the extra `postgrest/` prefix to shorten length of absolute path length of filenames.

Discussed in [#4469 (comment)](https://github.com/PostgREST/postgrest/pull/4469#pullrequestreview-3468088556).

### DIFF
--- a/nix/overlays/checked-shell-script/checked-shell-script.nix
+++ b/nix/overlays/checked-shell-script/checked-shell-script.nix
@@ -104,7 +104,6 @@ let
         ''
 
         + lib.optionalString withTmpDir ''
-          mkdir -p "''${TMPDIR:-/tmp}"
           tmpdir="$(${coreutils}/bin/mktemp -d --tmpdir ${name}-XXX)"
 
           # we keep the tmpdir when an error occurs for debugging


nix/overlays/checked-shell-script/checked-shell-script.nix
```diff
@@ -104,8 +104,8 @@ let
         ''
 
         + lib.optionalString withTmpDir ''
-          mkdir -p "''${TMPDIR:-/tmp}/postgrest"
-          tmpdir="$(${coreutils}/bin/mktemp -d --tmpdir postgrest/${name}-XXX)"
+          mkdir -p "''${TMPDIR:-/tmp}"
```
wolfgangwalther: I think (hope) we can now assume that TMPDIR already exists, so can remove the whole `mkdir`. Probably had the `-p` in there out of habit before not out of requirement.

taimoorzaeem: Yes, I think so as well. Removed in 17e74685.

wolfgangwalther (approved):

PR #4504, opened 2025-11-26
test(io): move fixtures to `fixtures/` directory

taimoorzaeem:
Organizes `io-tests` fixtures, similar to how we have in the `spec` tests.

```
$ ls
config.py     postgrest.py    test_big_schema.py  test_sanity.py
configs/      secrets/        test_cli.py         util.py
conftest.py   __snapshots__/  test_io.py
fixtures/     test_auth.py    test_replica.py
```

Looks much cleaner now! :v:


nix/tools/tests.nix
```diff
@@ -83,7 +83,7 @@ let
       }
       ''
         ${cabal-install}/bin/cabal v2-build ${devCabalOptions} exe:postgrest
-        ${cabal-install}/bin/cabal v2-exec -- ${withTools.withPg} -f test/io/fixtures.sql \
+        ${cabal-install}/bin/cabal v2-exec -- ${withTools.withPg} -f test/io/fixtures/fixtures.sql \
```
steve-chavez: This one is a bit confusing, maybe we should rename it `load.sql` or just `schema.sql` like we have on the `spec/` tests?

taimoorzaeem: Yes, I agree. I think `load.sql` would suit best because for now it has all kinds of data and not just "schema".

steve-chavez (approved):

PR #4517, opened 2025-12-04
add: log error when db-schemas config contain restricted schemas

### LINKED ISSUE #4359: Emit warning when adding `pg_catalog` on `db-schemas`
> Oh.. That explains it. What was the rationale for this decision?

We did it that way to avoid exposing sensitive information, and all roles have implicit access to `information_schema` and `pg_catalog`, IIRC revoking SELECT privileges to those schemas can break postgres in unexepected ways.

> Additionally it would be nice to get a warning when it is explicitly added to db-schemas instead of silently ignoring it.

Great idea. We should do that and also document this decision.

_Originally posted by @steve-chavez in https://github.com/PostgREST/postgrest/discussions/4252#discussioncomment-14516434_

taimoorzaeem:
We don't allow accessing schemas "pg_catalog" and "information_schema". Instead of silently failing requests on these schemas, this now adds logging an error on startup when these schemas are found in `db-schemas` config.

Closes #4359.

- [x] Implementation
- [x] Tests
- [x] Docs

### DIFF
--- a/docs/references/api/schemas.rst
+++ b/docs/references/api/schemas.rst
@@ -5,6 +5,10 @@ Schemas
 
 PostgREST can expose a single or multiple schema's tables, views and functions. The :ref:`active database role <roles>` must have the usage privilege on the schemas to access them.
 
+.. important::
+  
+  ``pg_catalog`` and ``information_schema`` are not allowed in :ref:`db-schemas`. This is done to prevent leaking sensitive information and hence they cannot be accessed directly. If you wish to expose objects of these schemas, expose another schema that contains wrapper views or functions over ``pg_catalog`` or ``information_schema`` objects.
+
 Single schema
 -------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. From versio
 
 ## Unreleased
 
+### Added
+
+- Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
+
 ### Fixed
 
 - Fix misleading logs on unsupported PostgreSQL versions by @taimoorzaeem in #4519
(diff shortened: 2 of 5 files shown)


taimoorzaeem:
Where should we document this? I am thinking in [schemas.html#restricted-schemas](https://docs.postgrest.org/en/v14/references/api/schemas.html#restricted-schemas) or maybe in the config page?

steve-chavez:
> Where should we document this? I am thinking in [schemas.html#restricted-schemas](https://docs.postgrest.org/en/v14/references/api/schemas.html#restricted-schemas) or maybe in the config page?

Hm, I'm thinking it should be more noticeable, maybe at the begging of https://docs.postgrest.org/en/v14/references/api/schemas.html, likely with an important or warning note.

docs/references/api/schemas.rst
```diff
@@ -5,6 +5,10 @@ Schemas
 
 PostgREST can expose a single or multiple schema's tables, views and functions. The :ref:`active database role <roles>` must have the usage privilege on the schemas to access them.
 
+.. important::
+  
+  Postgres system schemas ``pg_catalog`` and ``information_schema`` are not allowed in :ref:`db-schemas`. This is done to prevent leaking sensitive information and hence they cannot be accessed directly. As a workaround, we suggest that you create a view in your other exposed schema, which selects from these restricted schemas.
```
steve-chavez: Shorter plus a bit more detail:

```suggestion
  ``pg_catalog`` and ``information_schema`` are not allowed in :ref:`db-schemas`. This is done to prevent leaking sensitive information and hence they cannot be accessed directly. If you wish to expose objects of these schemas, expose another schema that contains wrapper views or functions over ``pg_catalog`` or ``information_schemas`` objects .
```

steve-chavez (approved):

steve-chavez (approved):
LGTM. Was about to merge but noticed this is marked as draft.

taimoorzaeem:
> Was about to merge but noticed this is marked as draft.

Yeah, I wanted to rebase this on top of #4531 to avoid any conflicts in changelog. Opened now.

steve-chavez:
@taimoorzaeem I think this one should have a "breaking change" entry on the changelog since before we were lenient.

PR #4518, opened 2025-12-05
fix: regression that truncates error message when offset is out of bounds

### LINKED ISSUE #4455: Malformed response to request with offset larger than resultset in combination with Prefer:count=exact
### Environment

* PostgreSQL version: PostgreSQL 15.13
* PostgREST version: postgrest/postgrest:v13.0.8

### Description of issue

An invalid response is returned to a request that has an offset query parameter specified that is larger than the number of entries in the underlying table. (only if the Prefer:count=exact header is set on the request)

`curl -i "http://localhost:3000/items?offset=2000" -H "Prefer: count=exact"`

Until postgrest version v12.2.12 the following response was returned
HTTP/1.1 416 Requested Range Not Satisfiable

Later postgrest versions return a malformed response that cannot be parsed

laurenceisla:
Closes #4455

### DIFF
--- a/src/PostgREST/Response.hs
+++ b/src/PostgREST/Response.hs
@@ -65,7 +65,7 @@ actionResponse :: DbResult -> ApiRequest -> (Text, Text) -> AppConfig -> SchemaC
 actionResponse (DbCrudResult WrappedReadPlan{pMedia, wrHdrsOnly=headersOnly, crudQi=identifier} RSStandard{..}) ctxApiRequest@ApiRequest{iPreferences=Preferences{..},..} _ _ _ _ _ = do
   let
     (status, contentRange) = RangeQuery.rangeStatusHeader iTopLevelRange rsQueryTotal rsTableTotal
-    cLHeader = if headersOnly then mempty else [contentLengthHeaderLazy bod]
+    cLHeader = if headersOnly then mempty else [ contentLengthHeader bod ]
     prefHeader = maybeToList . prefAppliedHeader $ Preferences Nothing Nothing preferCount preferTransaction Nothing preferHandling preferTimezone Nothing []
     headers =
       [ contentRange
@@ -120,7 +120,7 @@ actionResponse (DbCrudResult MutateReadPlan{mrMutation=MutationCreate, mrMutateP
       Just HeadersOnly -> (headers, mempty)
       Nothing -> (headers, mempty)
 
-  (ovStatus, ovHeaders) <- overrideStatusHeaders rsGucStatus rsGucHeaders status $ contentLengthHeaderLazy bod:headers'
+  (ovStatus, ovHeaders) <- overrideStatusHeaders rsGucStatus rsGucHeaders status $ contentLengthHeader bod:headers'
 
   Right $ PgrstResponse ovStatus ovHeaders bod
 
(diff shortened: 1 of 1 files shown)


src/PostgREST/Response.hs
```diff
@@ -65,7 +65,7 @@ actionResponse :: DbResult -> ApiRequest -> (Text, Text) -> AppConfig -> SchemaC
 actionResponse (DbCrudResult WrappedReadPlan{pMedia, wrHdrsOnly=headersOnly, crudQi=identifier} RSStandard{..}) ctxApiRequest@ApiRequest{iPreferences=Preferences{..},..} _ _ _ _ _ = do
   let
     (status, contentRange) = RangeQuery.rangeStatusHeader iTopLevelRange rsQueryTotal rsTableTotal
-    cLHeader = if headersOnly then mempty else [contentLengthHeaderStrict rsBody]
+    cLHeader = if headersOnly then mempty else [contentLengthHeaderLazy bod]
```
laurenceisla: It was taking the SQL query result `[]` thus setting `Content-Length: 2` and then truncating the error response message. Currently checking if this didn't happen anywhere else for the other responses.

test/spec/Feature/Query/RangeSpec.hs
```diff
@@ -126,7 +126,8 @@ spec = do
                 "hint":null
               }|]
             { matchStatus  = 416
-            , matchHeaders = ["Content-Range" <:> "*/0"]
+            , matchHeaders = [ "Content-Range" <:> "*/0"
+                             , "Content-Length" <:> "144"]
```
laurenceisla, line 130: The test only responds with the `Content-Length` when it's added to `matchHeaders`... this is unexpected for a test since it shouldn't modify how our application works. Unless there's a way to make this same request and avoid getting the "Content-Length" or ignore it?

src/PostgREST/Response.hs
laurenceisla: It looks the 201 responses for INSERTS weren't returning a `Content-Length` either, only for the cases of `Content-Length: 0` (Only `204 No Content` responses are forbidden of doing so according to the spec IIRC). I added that in the second commit.

src/PostgREST/Response.hs
```diff
@@ -249,14 +252,8 @@ decodeGucStatus :: Maybe Text -> Either Error.Error (Maybe HTTP.Status)
...
-contentLengthHeader :: Show b => (a -> b) -> a -> HTTP.Header
-contentLengthHeader lenFn body = ("Content-Length", show (lenFn body))
-
-contentLengthHeaderStrict :: BS.ByteString -> HTTP.Header
-contentLengthHeaderStrict = contentLengthHeader BS.length
-
-contentLengthHeaderLazy :: LBS.ByteString -> HTTP.Header
-contentLengthHeaderLazy = contentLengthHeader LBS.length
+contentLengthHeader :: LBS.ByteString -> HTTP.Header
+contentLengthHeader body = ("Content-Length", show (LBS.length body))
```
steve-chavez, line 256: This change looks good. But I don't fully get why it caused https://github.com/PostgREST/postgrest/issues/4455#issuecomment-3499622589, could this be elaborated in comments?

Also, do we have a test that ensures the regression on https://github.com/PostgREST/postgrest/issues/4455#issuecomment-3499622589 doesn't happen again?

src/PostgREST/Response.hs
steve-chavez: > It was taking the SQL query result [] thus setting Content-Length: 2 and then truncating the error response message.

Ah shoot I see it now, we were using the wrong variable, `bod` had some conditional logic:

https://github.com/PostgREST/postgrest/blob/1724e4551d64fbd26e326201602ef65dc66ee491/src/PostgREST/Response.hs#L81-L84

src/PostgREST/Response.hs
steve-chavez: Nvm, clarified on https://github.com/PostgREST/postgrest/pull/4518#discussion_r2604106077

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file. From versio
 
 ## [14.1] - 2025-11-05
 
+- Fix regression where the `PGRST103` error response was truncated by @laurenceisla in #4455
+  + Happened when an `offset` was greater than the rows requested and `Prefer: count=exact` was sent.
+- Fix not returning `Content-Length` on empty HTTP `201` responses by @laurenceisla in #4518
+
```
taimoorzaeem, line 12: I noticed this entry should go to the `Unreleased` section instead of `14.1` because that is already released.

laurenceisla: Yes, my bad. Was fixed here: https://github.com/PostgREST/postgrest/pull/4531

PR #4523, opened 2025-12-08
fix: out of memory errors for large number of tables

taimoorzaeem:
As discussed in https://github.com/PostgREST/postgrest/pull/4468#issuecomment-3619703503.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -428,9 +428,12 @@ noRpcHint schema procName params allProcs overloadedProcs =
 
 -- |
 -- Do a fuzzy search in all tables in the same schema and return closest result
+-- We have a upper limit for hint generation as 4000 tables to avoid memory
+-- and performance issues.
 tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
 tableNotFoundHint schema tblName tblList
-  = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
+  | length tblList <= 4000 = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
+  | otherwise              = Nothing
     where
       perhapsTable = Fuzzy.getOne fuzzyTableSet tblName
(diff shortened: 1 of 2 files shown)


taimoorzaeem:
I tested this manually using `md5` hashes (cutting length to just `7` chars) as table names. The `4000` is not a good number and out of memory errors still occur with this setup.

According to my analysis, max `500` number of tables works best for fuzzy search even for `24` character long hashes as table names.

src/PostgREST/Error.hs
```diff
@@ -428,9 +428,12 @@ noRpcHint schema procName params allProcs overloadedProcs =
 
 -- |
 -- Do a fuzzy search in all tables in the same schema and return closest result
+-- We have a upper limit for hint generation as 4000 tables to avoid memory
+-- and performance issues.
 tableNotFoundHint :: Text -> Text -> [Table] -> Maybe Text
 tableNotFoundHint schema tblName tblList
-  = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
+  | length tblList <= 4000 = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
```
mkleczek, line 435: There are several problems with this change:

1.
`length tblList` complexity is linear in the list size and it has to iterate the whole list even if it is longer than 4k.

Leaving it like this opens up a DoS vulnerability.

2.
4k is too many from my tests. On my machine PostgREST became unstable with 1k tables.
This is another DoS vulnerability.

3.
Re-creating `FuzzySet` upon each request opens up yet another DoS vulnerability.

I'm afraid there is really no other sensible solution to this than https://github.com/PostgREST/postgrest/pull/4472

taimoorzaeem: > 4k is too many from my tests. On my machine PostgREST became unstable with 1k tables.

Yes, agreed. I was proposing `500` to be the upper limit as mentioned in https://github.com/PostgREST/postgrest/pull/4523#issuecomment-3626033233.

taimoorzaeem:
Hmm, I see that this is not a sufficient enough solution. Closing.

PR #4524, opened 2025-12-08
perf: optimize count=exact when there's no limits, offsets or db-max-rows

### LINKED ISSUE #3957: `Prefer: count=exact` optimization on HEAD request
> Why is the page count slow? Shouldn't it be quick compared to the whole table count?
> 
> It's quicker if `db-max-rows` is set to a lower value. If it's not set or if it's big enough then it will take more time.
> 
> Maybe share EXPLAIN analyze output of these requests?
> 
> (These requests do not have `db-max-rows` set)
> 
> <details><summary>Without `limit=0`</summary>
> <p>
> 
> ```bash
> curl 'localhost:3030/big_table' -H 'Prefer: count=exact' -I
> ```
> ```sql
> EXPLAIN (FORMAT TEXT, ANALYZE, VERBOSE)
> WITH pgrst_source AS
>   (SELECT "public"."big_table".* FROM "public"."big_table") ,
>  pgrst_source_count AS
>   (SELECT 1 FROM "public"."big_table")
> SELECT
>   (SELECT pg_catalog.count(*) FROM pgrst_source_count) AS total_result_set,
>   pg_catalog.count(_postgrest_t) AS page_total,
>   ''::text AS body,
>   nullif(current_setting('response.headers', TRUE), '') AS response_headers,
>   nullif(current_setting('response.status', TRUE), '') AS response_status,
>   '' AS response_inserted
> FROM
>   (SELECT * FROM pgrst_source) _postgrest_t
> ```
> ```bash
> Aggregate  (cost=4562.01..4562.03 rows=1 width=144) (actual time=33.566..33.566 rows=1 loops=1)
>   Output: (InitPlan 1).col1, count(ROW(big_table.id, big_table.name, big_table.val, big_table.num)), ''::text, NULLIF(current_setting('response.headers'::text, true), ''::text), NULLIF(current_setting('response.status'::text, true), ''::text), ''::text
>   InitPlan 1
>     ->  Aggregate  (cost=2281.00..2281.01 rows=1 width=8) (actual time=4.900..4.900 rows=1 loops=1)
>           Output: count(*)
>           ->  Seq Scan on public.big_table big_table_1  (cost=0.00..2031.00 rows=100000 width=0) (actual time=0.009..2.627 rows=100000 loops=1)
>                 Output: big_table_1.id, big_table_1.name, big_table_1.val, big_table_1.num
>   ->  Seq Scan on public.big_table  (cost=0.00..2031.00 rows=100000 width=50) (actual time=0.009..7.667 rows=100000 loops=1)
>         Output: big_table.id, big_table.name, big_table.val, big_table.num
> Planning Time: 0.151 ms
> Execution Time: 33.598 ms
> ```
> 
> </p>
> </details> 
> 
> <details><summary>With `limit=0`</summary>
> <p>
> 
> ```bash
> curl 'localhost:3030/big_table?limit=0' -H 'Prefer: count=exact' -I
> ```
> ```sql
> EXPLAIN (FORMAT TEXT, ANALYZE, VERBOSE)
> WITH pgrst_source AS
>   (SELECT "public"."big_table".* FROM "public"."big_table" LIMIT 0 OFFSET 0) ,
>  pgrst_source_count AS
>   (SELECT 1 FROM "public"."big_table")
> SELECT
>   (SELECT pg_catalog.count(*) FROM pgrst_source_count) AS total_result_set,
>   pg_catalog.count(_postgrest_t) AS page_total,
>   ''::text AS body,
>   nullif(current_setting('response.headers', TRUE), '') AS response_headers,
>   nullif(current_setting('response.status', TRUE), '') AS response_status,
>   '' AS response_inserted
> FROM
>   (SELECT * FROM pgrst_source) _postgrest_t
> ```
> ```bash
> Aggregate  (cost=2281.03..2281.05 rows=1 width=144) (actual time=17.579..17.580 rows=1 loops=1)
>   Output: (InitPlan 1).col1, count(ROW(big_table.id, big_table.name, big_table.val, big_table.num)), ''::text, NULLIF(current_setting('response.headers'::text, true), ''::text), NULLIF(current_setting('response.status'::text, true), ''::text), ''::text
>   InitPlan 1
>     ->  Aggregate  (cost=2281.00..2281.01 rows=1 width=8) (actual time=17.566..17.566 rows=1 loops=1)
>           Output: count(*)
>           ->  Seq Scan on public.big_table big_table_1  (cost=0.00..2031.00 rows=100000 width=0) (actual time=0.015..9.224 rows=100000 loops=1)
>                 Output: big_table_1.id, big_table_1.name, big_table_1.val, big_table_1.num
>   ->  Limit  (cost=0.00..0.02 rows=1 width=50) (actual time=0.001..0.002 rows=0 loops=1)
>         Output: big_table.id, big_table.name, big_table.val, big_table.num
>         ->  Seq Scan on public.big_table  (cost=0.00..2031.00 rows=100000 width=50) (never executed)
>               Output: big_table.id, big_table.name, big_table.val, big_table.num
> Planning Time: 0.289 ms
> Execution Time: 17.657 ms
> ```
> 
> 
> </p>
> </details>  

 _Originally posted by @laurenceisla in [#2849](https://github.com/PostgREST/postgrest/issues/2849#issuecomment-2735908327)_

See possible solution on https://github.com/PostgREST/postgrest/issues/2849#issuecomment-2737476594

laurenceisla:
Closes #3957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ All notable changes to this project will be documented in this file. From versio
 - Ensure Listener connections are released by @mkleczek in #4614
 - Fix incorrectly filtering the returned representation for PATCH requests when using `or/and` filters by @laurenceisla in #3707
 - Fix listener running with exception masked after first failure #4615
+- Optimize requests with `Prefer: count=exact` that do not use ranges or `db-max-rows` by @laurenceisla in #3957
+  + Removed unnecessary double count when building the `Content-Range`.
 
 ## [14.3] - 2026-01-03
 
--- a/src/PostgREST/Query.hs
+++ b/src/PostgREST/Query.hs
@@ -43,16 +43,16 @@ data MainQuery = MainQuery
 
 mainQuery :: ActionPlan -> AppConfig -> ApiRequest -> AuthResult -> Maybe QualifiedIdentifier -> MainQuery
 mainQuery (NoDb _) _ _ _ _ = MainQuery mempty Nothing mempty (mempty, mempty, mempty) mempty
-mainQuery (Db plan) conf@AppConfig{..} apiReq@ApiRequest{iPreferences=Preferences{..}} authRes preReq =
+mainQuery (Db plan) conf@AppConfig{..} apiReq@ApiRequest{iTopLevelRange=range, iPreferences=Preferences{..}} authRes preReq =
   let genQ = MainQuery (PreQuery.txVarQuery plan conf authRes apiReq) (PreQuery.preReqQuery <$> preReq) in
   case plan of
     DbCrud _ WrappedReadPlan{..} ->
       let countQuery = QueryBuilder.readPlanToCountQuery wrReadPlan in
-      genQ (Statements.mainRead wrReadPlan countQuery preferCount configDbMaxRows pMedia wrHandler) (mempty, mempty, mempty)
+      genQ (Statements.mainRead wrReadPlan countQuery preferCount configDbMaxRows range pMedia wrHandler) (mempty, mempty, mempty)
       (if shouldExplainCount preferCount then Just (Statements.postExplain countQuery) else Nothing)
     DbCrud _ MutateReadPlan{..} ->
       genQ (Statements.mainWrite mrReadPlan mrMutatePlan pMedia mrHandler preferRepresentation preferResolution) (mempty, mempty, mempty) mempty
     DbCrud _ CallReadPlan{..} ->
-      genQ (Statements.mainCall crProc crCallPlan crReadPlan preferCount pMedia crHandler) (mempty, mempty, mempty) mempty
+      genQ (Statements.mainCall crProc crCallPlan crReadPlan preferCount configDbMaxRows range pMedia crHandler) (mempty, mempty, mempty) mempty
     MayUseDb InspectPlan{ipSchema=tSchema} ->
       genQ mempty (SqlFragment.accessibleTables tSchema, SqlFragment.accessibleFuncs tSchema, SqlFragment.schemaDescription tSchema) mempty
--- a/src/PostgREST/Query/SqlFragment.hs
+++ b/src/PostgREST/Query/SqlFragment.hs
@@ -487,15 +487,15 @@ pgFmtGroup _  CoercibleSelectField{csAggFunction=Just _} = Nothing
 pgFmtGroup _  CoercibleSelectField{csAlias=Just alias, csAggFunction=Nothing} = Just $ pgFmtIdent alias
 pgFmtGroup qi CoercibleSelectField{csField=fld, csAlias=Nothing, csAggFunction=Nothing} = Just $ pgFmtField qi fld
 
-countF :: SQL.Snippet -> Bool -> (SQL.Snippet, SQL.Snippet)
-countF countQuery shouldCount =
-  if shouldCount
-    then (
-        ", pgrst_source_count AS (" <> countQuery <> ")"
-      , "(SELECT pg_catalog.count(*) FROM pgrst_source_count)" )
-    else (
-        mempty
-      , "null::bigint")
+countF :: SQL.Snippet -> SQL.Snippet -> Bool -> Maybe Integer -> NonnegRange -> (SQL.Snippet, SQL.Snippet)
+countF countQuery pageCountSelect shouldCount maxRows range
+  | shouldCount = if isJust maxRows || range /= allRange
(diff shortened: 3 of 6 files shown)


steve-chavez:
@laurenceisla Could you inline all the context here? :pray: I'm finding difficult to understand the PR.

laurenceisla:
@steve-chavez Yes, so the requests handle two counts when using the `Prefer: count=exact` header in the request. This is to build the two parts of the `Content-Range` header. For example, let's take our items table:

```console
$ curl 'localhost:3000/items?limit=5' -H 'Prefer: count=exact' -I
HTTP/1.1 206 Partial Content
...
Content-Range: 0-4/15
```

To build the `Content-Range`, It counts the limited query to get the `0-4` and the total query to get the `15` (two counts in total).

However if there's nothing that would limit the request (`limit`, `offset` or `db-max-rows`) then we'd get:

```console
$ curl 'localhost:3000/items' -H 'Prefer: count=exact' -I
HTTP/1.1 200 OK
...
Content-Range: 0-14/15
```

The total quantity would always be the same for both parts of the `Content-Range`: `0-14` and `15` (15 items for both).  So, this PR only counts once (the `0-14` part) and uses the same result to build the second part of the `Content-Range` (the `15` part).

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file. From versio
 ### Added
 
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
+- Optimize requests with `Prefer: count=exact` that do not use ranges or `db-max-rows` by @laurenceisla in #3957
+  + The page total is the same as the resource total in these cases, so now it only counts once to build the `Content-Range`.
```
steve-chavez: Maybe we should  be more explicit here and say "get rid of double count.."?

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ All notable changes to this project will be documented in this file. From versio
 - Ensure Listener connections are released by @mkleczek in #4614
 - Fix incorrectly filtering the returned representation for PATCH requests when using `or/and` filters by @laurenceisla in #3707
 - Fix listener running with exception masked after first failure #4615
+- Optimize requests with `Prefer: count=exact` that do not use ranges or `db-max-rows` by @laurenceisla in #3957
+  + Removed unnecessary double count when building the `Content-Range`.
```
steve-chavez, line 20: @laurenceisla I missed this.. isn't this really a feature? I believe we've had an optimization before added as feature.

steve-chavez: Yes, for example in v13: https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#1300---2025-05-08

> #2858, Performance improvements when calling RPCs via GET using indexes in more cases

laurenceisla: Oh, I took the latest "perf" commit as reference and it was under "Fixed". From what I see, a Ctrl+F for `perf` shows that there are others that are either under "Fixed" or "Added". Would all optimizations need to be considered features from now on, or it depends on the nature of the commit?

steve-chavez: > Oh, I took the latest "perf" commit as reference and it was under "Fixed". 

You're talking about be9a0ab467021d99313f8d33efbe4c2f1c8ceb70 right? I think that was a mistake but it was no biggie since it got released in the v14 major.

If you look at the next `perf:` by Wolfgang ee4bfbf2533211f90715fc512dad5f78ef6ccd9d you can see it got listed in features.

From my POV optimizations should be features since they have potential to introduce new bugs.

@wolfgangwalther WDYT?

wolfgangwalther: I don't think perf makes sense in either case. How about a new section "Performance Improvements"?

steve-chavez: > I don't think perf makes sense in either case. How about a new section "Performance Improvements"?

Hm, I think that could be a subsection in "Features", like we've been doing in [v13](https://github.com/PostgREST/postgrest/releases/tag/v13.0.0) (forgot to do it in [v14](https://github.com/PostgREST/postgrest/releases/tag/v14.0) since we had just a couple features but I updated now).

But the point here is mostly about if perf improvements should be backpatched, which I assume we all agree we should not?

wolfgangwalther: > But the point here is mostly about if perf improvements should be backpatched, which I assume we all agree we should not?

Agreed.

PR #4525, opened 2025-12-09
fix: misleading logs on unsupported postgresql versions

### LINKED ISSUE #4519: Misleading logs on unsupported PostgreSQL version
## Problem

This happens when trying a pg 12:

```
postgrest /nix/store/wa7y1f432g0qqph7sjn30ibl1q38l0v9-pgrst.conf
05/Dec/2025:05:18:43 +0000: Starting PostgREST 14.1...
05/Dec/2025:05:18:43 +0000: Admin server listening on 0.0.0.0:3001
05/Dec/2025:05:18:43 +0000: API server listening on /tmp/pgrst.sock
05/Dec/2025:05:18:43 +0000: Listening for database notifications on the "pgrst" channel
05/Dec/2025:05:18:43 +0000: Cannot run in this PostgreSQL version (12.19), PostgREST needs at least 13.0
05/Dec/2025:05:18:43 +0000: Successfully connected to PostgreSQL 12.19 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 13.2.0, 64-bit
05/Dec/2025:05:18:43 +0000: Connection Pool initialized with a maximum size of 1 connections
postgrest: thread killed
```

The last two log lines are confusing, for a while I didn't understand why the `thread killed` but the reason is clear above `Cannot run in this PostgreSQL version...`.

## Solution

The log `05/Dec/2025:05:18:43 +0000: Cannot run in this PostgreSQL version (12.19), PostgREST needs at least 13.0` should be the last one.

taimoorzaeem:
Postgrest fails on unsupported pg versions. However before killing the thread, it continues to print a few more log messages which were misleading. This commit fixes this by making sure that the no log message should be printed after the unsupported pg version observation and kill the thread immediately.

Closes #4519.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. From versio
 
 ## Unreleased
 
+### Fixed
+
+- Fix misleading logs on unsupported PostgreSQL versions by @taimoorzaeem in #4519
+
 ## [14.1] - 2025-11-05
 
 ## Fixed
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -382,14 +382,16 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
             observer ExitDBNoRecoveryObs
             killThread mainThreadId
           return Nothing
-        Right actualPgVersion -> do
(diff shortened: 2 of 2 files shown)


taimoorzaeem:
I forgot to mention that I tested this locally/manually and it works fine :+1:. I couldn't figure out an automated way to test this.

Issue #4381, opened 2025-10-03
Per-request timeout


steve-chavez:
## Problem

Currently we support setting a [statement_timeout](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT) [per role](https://docs.postgrest.org/en/v13/references/transactions.html#impersonated-role-settings) and [per function](https://docs.postgrest.org/en/v13/references/transactions.html#hoisted-function-settings), but we don't have a way to support it per table or view or when resource embedding is used.

## Solution

Send the timeout through an HTTP header.

```http
Timeout: Second-2
```

The above is the Timeout header from WebDAV ([ref](https://www.rfc-editor.org/rfc/rfc4918#section-10.7)), but it has WebDAV specific semantics so I don't think we can reuse it.

A saner option would be a Prefer header:

```http
Prefer: handling=strict, timeout=2
```

Some drawbacks I can see is that this can be used to force failure by end users if the `statement_timeout` is set to a value too low (this in turn can pollute logs). For this maybe we can only allow seconds, instead of milliseconds. So only 1 second as minimum.

This timeout should also have a threshold, it cannot be too high. This threshold could be the role timeout or the function timeout.

mkleczek:
I like this idea.

Besides handling timeout for relation queries, it also somewhat addresses the problem that setting timeout on functions makes them non-inlinable.

steve-chavez:
> it also somewhat addresses the problem that setting timeout on functions makes them non-inlinable.

True, that's relevant for [Table-Valued functions](https://docs.postgrest.org/en/v13/references/api/functions.html#table-valued-functions) + filters but not for scalar functions. We should add a note to https://docs.postgrest.org/en/v13/references/transactions.html#hoisted-function-settings mentioning that drawback.

steve-chavez:
Looks ready for implementation. It's also possible that some cases won't need this functionality, so we should revisit https://github.com/PostgREST/postgrest/issues/2987 afterwards.

taimoorzaeem:
Do we need a new `PGRST` error for this or should we just return the statement timeout error from pg directly?

steve-chavez:
Should be fine to return the error code from pg.

steve-chavez:
I've noticed this can be implemented with the standard `wait` preference https://www.rfc-editor.org/info/rfc7240/#section-4.3 combined with `handling=strict`. `handling=lenient` would do nothing.

[...]

PR #4533, opened 2025-12-11
refactor: create function to handle response preferences

taimoorzaeem:
Centralizes handling of preferences that are used to create `Preference-Applied` header which is returned on responses.

Before working on #4381, I think this is nice refactor to clean things up a bit. This would help in adding/handling the preferences in only one place instead of having to change in every `actionResponse`.

It also makes it more explicit in knowing what preferences are being returned on what type of plan/response.


steve-chavez (approved):
Looks much better ❤️

PR #4553, opened 2025-12-14
fix: hasSingleUnnamedParam incorrectly matching named parameters

joelonsql:
The hasSingleUnnamedParam function was only checking the parameter type but not whether the parameter actually had no name.

This is just a minor bug, since it only cause a less precise error message than if adding the `ppName == mempty` check so functions with named parameters don't incorrectly match the single-param fallback.

```sql
create or replace function test.named_json_param(data json) returns json as $$
  select data;
$$ language sql;

--        it "rejects json body when single param has a name" $
--          post "/rpc/named_json_param"
--              [json|{"A": 1, "B": 2, "C": 3}|]
```

Eror before Fix:

```json
  {
    "code": "42883",
    "message": "function test.named_json_param() does not exist",
    "details": null,
    "hint": "No function matches the given name and argument types..."
  }
```

Error after fix:

```json
  {
    "code": "PGRST202",
    "message": "Could not find the function test.named_json_param(A, B, C) in the schema cache",
    "details": "Searched for the function test.named_json_param with parameters A, B, C or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache.",
    "hint": null
  }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
-- Fix `hasSingleUnnamedParam` incorrectly matching functions with named parameters by @joelonsql in #xxxx
+- Fix `hasSingleUnnamedParam` incorrectly matching functions with named parameters by @joelonsql in #4553
   + Functions with a single named parameter (e.g., `foo(data json)`) no longer incorrectly match the single-param fallback, returning a clean `PGRST202` error instead of a confusing PostgreSQL `42883` error.
 - Fix misleading logs on unsupported PostgreSQL versions by @taimoorzaeem in #4519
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
+- Fix `hasSingleUnnamedParam` incorrectly matching functions with named parameters by @joelonsql in #xxxx
```
steve-chavez:
```suggestion
- Fix `hasSingleUnnamedParam` incorrectly matching functions with named parameters by @joelonsql in #4553
```

steve-chavez (approved):
@joelonsql Nice DX improvement 👍 

(Edit: CI failure for commit message should go away once I squash and merge)

PR #4569, opened 2025-12-17
fix: correct typo in Routine Ord instance comparison

joelonsql:
The condition `schema2 == schema2` was always true. Changed to `schema1 == schema2` to properly compare schemas when ordering overloaded functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
-- Fix incorrect ordering of overloaded functions when schemas differ by @joelonsql in #4554
-  + The `Ord` instance for `Routine` had a typo comparing `schema2 == schema2` instead of `schema1 == schema2`.
+- Fix incorrect ordering of overloaded functions by @joelonsql in #4555
+  + Fixes comparison between two functions in the same schema differing by length of parameters.
 - Fix `hasSingleUnnamedParam` incorrectly matching functions with named parameters by @joelonsql in #4553
   + Functions with a single named parameter (e.g., `foo(data json)`) no longer incorrectly match the single-param fallback, returning a clean `PGRST202` error instead of a confusing PostgreSQL `42883` error.
 - Fix misleading logs on unsupported PostgreSQL versions by @taimoorzaeem in #4519


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
+- Fix incorrect ordering of overloaded functions when schemas differ by @joelonsql in #4554
+  + The `Ord` instance for `Routine` had a typo comparing `schema2 == schema2` instead of `schema1 == schema2`.
```
wolfgangwalther: What would be the observable effect for the user here? We should put that down in the changelog, not the code change.

taimoorzaeem: I think this should do? It's a small change and to explain it completely, we would otherwise need to write a _long_ description which might confuse the users.

```suggestion
- Fix incorrect ordering of overloaded functions by @joelonsql in #4555
  + Fixes comparison between two functions in the same schema differing by length of parameters.
```

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
+- Fix incorrect ordering of overloaded functions by @joelonsql in #4555
+  + Fixes comparison between two functions in the same schema differing by length of parameters.
```
steve-chavez, line 14: 🤔 Could we add a test for this?

CHANGELOG.md
wolfgangwalther: Sorry, I don't want to be pedantic, but as a user I still have no idea what that means for me. As a user, I don't compare functions - why would I? Also what is "length of parameters", do you mean number of parameters? Or the length of their names?

Also.. why are overloaded functions ordered in any way?

(I know the answers to some of these questions, I guess, but most of our users won't)

wolfgangwalther: (and to be clear, the reasons why I am not proposing a specific change is, because I just don't know what the actual difference is, so far. Comparing *schemas* wrong sure is a bit more scary - could it be that we'd be calling a function from a different schema in some cases? Or are we failing to call functions when they exist in the two different exposed schemas with the same name? Or.. I don't know.)

taimoorzaeem: Yeah, now that I think more about it, it is concerning. I will try to find some time to investigate this further.

wolfgangwalther:
Merged as refactor without changelog in 388930336d227442bbe20be2aff95f540bc5ff22 - thank you @joelonsql!

PR #4572, opened 2025-12-19
refactor: move content negotiation logic to Negotiate.hs

taimoorzaeem:
Moves the `negotiateContent` function to `Plan/Negotiate.hs` module. This also adds comments to describe the current approach we have for negotiation, as discussed in https://github.com/PostgREST/postgrest/pull/4510#discussion_r2584907279.

### DIFF
--- a/src/PostgREST/Plan/Negotiate.hs
+++ b/src/PostgREST/Plan/Negotiate.hs
@@ -1,0 +1,82 @@
+{-|
+Module      : PostgREST.Plan.Negotiate
+Description : PostgREST Content Negotiation
+
+This module contains logic for content negotiation.
+RFC: https://datatracker.ietf.org/doc/html/rfc7231#section-3.4
+-}
+
+module PostgREST.Plan.Negotiate
+  ( negotiateContent
+  ) where
+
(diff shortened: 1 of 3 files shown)


src/PostgREST/Plan/Negotiate.hs
```diff
@@ -0,0 +1,82 @@
...
+--    into "mimetypes". We have a default builtin handler that does
+--    "application/json". We can add more handlers via aggregates.
+--
+-- 2. "mime-typed" values, which specifically return a domain type that is
+--    associated to a certain mimetype. e.g, a function returning only
+--    "image/png".
+--
+--    These are simple. If you get a request with a "matching" accept header,
+--    you call the function, otherwise you return an error. If the function
+--    returns image/png, you'd accept */*, image/* and image/png.
```
wolfgangwalther: I don't think we are implementing the wildcard stuff, yet. Certainly not the partial wildcard, not sure about the full wildcard.

taimoorzaeem: > Certainly not the partial wildcard, not sure about the full wildcard.

Yeah, we don't do both as of yet. I have updated the comment.

PR #4576, opened 2025-12-22
nix: add new loadtest for errors

steve-chavez:
Adds a new loadtest that includes only requests that produces errors (misspelled relations, misspelled functions, permission errors).

Tries to repro https://github.com/PostgREST/postgrest/issues/4462

### DIFF
--- a/test/load/errors.sql
+++ b/test/load/errors.sql
@@ -1,0 +1,7 @@
+\ir fixtures.sql
+
+SELECT format('CREATE TABLE test.actors_%s ();', n)
+FROM generate_series(1, 20000) n
+\gexec
+
+-- TODO add many function for fuzzy search (somehow this is making the loadtest start slow)
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -120,7 +120,7 @@ jobs:
   loadtest:
     strategy:
       matrix:
-        kind: ['mixed', 'jwt-hs', 'jwt-hs-cache', 'jwt-hs-cache-worst', 'jwt-rsa', 'jwt-rsa-cache', 'jwt-rsa-cache-worst']
+        kind: ['mixed', 'errors', 'jwt-hs', 'jwt-hs-cache', 'jwt-hs-cache-worst', 'jwt-rsa', 'jwt-rsa-cache', 'jwt-rsa-cache-worst']
     name: Loadtest
     runs-on: ubuntu-24.04
     steps:
--- a/nix/tools/loadtest.nix
+++ b/nix/tools/loadtest.nix
@@ -45,7 +45,7 @@ let
           "ARG_OPTIONAL_SINGLE([testdir], [t], [Directory to load tests and fixtures from], [./test/load])"
           "ARG_OPTIONAL_SINGLE([kind], [k], [Kind of loadtest], [mixed])"
           "ARG_OPTIONAL_SINGLE([method],, [HTTP method used for the jwt loadtests], [OPTIONS])"
-          "ARG_TYPE_GROUP_SET([KIND], [KIND], [kind], [mixed,jwt-hs,jwt-hs-cache,jwt-hs-cache-worst,jwt-rsa,jwt-rsa-cache,jwt-rsa-cache-worst])"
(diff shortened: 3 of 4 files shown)


steve-chavez:
So far this proves that the rate of error processing is much lower than normal successful requests.
Rate of `errors` is ~67 vs and `mixed` is 468, this while using much more CPU, ~100% on `errors` vs ~20% on `mixed`.
See https://github.com/PostgREST/postgrest/actions/runs/20441827293?pr=4576#summary-58736347815.

OOMs of #4462 are not reproduced, this looks only dependent of the number of tables, not the number of continued requests.

steve-chavez:
Later on we can add more types of errors for this load test: https://docs.postgrest.org/en/v14/references/errors.html#postgrest-error-codes. We should ensure error processing stays fast for every case.

test/load/errors.sql
```diff
@@ -0,0 +1,25 @@
...
+      format('CREATE TABLE test.actors_%s ()', n) AS tbls,
+      format($__$CREATE FUNCTION test.call_me_%s (name TEXT)
+              RETURNS TEXT
+              STABLE
+              LANGUAGE SQL
+              AS $func$
+                SELECT 'Hello ' || name;
+              $func$;$__$, n) AS funcs
+    FROM
+      generate_series(1, 950) n
```
steve-chavez: I'm unable to increase this value more because of: 

https://github.com/PostgREST/postgrest/blob/a001108424c50993aa56eb4cc0d6d730b327ad23/postgrest.cabal#L195-L196

Which leads to `Empty reply from server` for the misspelled relations requests.

steve-chavez: Nvm, I was mistaken. The `loadtest` runs the binary by default, it doesn't have the above constraint.

https://github.com/PostgREST/postgrest/blob/3da32e585479aa5d61db2e18229f9d249d8dd95e/nix/tools/withTools.nix#L375-L386

steve-chavez:
This [CI job](https://github.com/PostgREST/postgrest/actions/runs/20448684845?pr=4576#summary-58758748835) proves there's considerable memory consumed when the table/function count is 30K. Unfortunately that number is causing a CI coverage failure which I don't understand. So I'll leave the table/function count at 950.

steve-chavez:
> Unfortunately that number is causing a CI coverage failure which I don't understand. 

And looks like this coverage failure persist even after reducing the table count. The error reported on CI doesn't reveal anything and locally it cannot be reproduced.

nix/tools/withTools.nix
```diff
@@ -81,7 +81,8 @@ let
           # On MacOS, it's 104 chars
           # See: https://serverfault.com/questions/641347/check-if-a-path-exceeds-maximum-for-unix-domain-socket
 
-          pg_ctl -l "$tmpdir/db.log" -w start -o "-F -c listen_addresses=\"\" -c hba_file=$HBA_FILE -k $PGHOST -c log_statement=\"all\" " \
+          # max_locks_per_transaction was modified from the default so we can create a lot of tables in one transaction
+          pg_ctl -l "$tmpdir/db.log" -w start -o "-F -c listen_addresses=\"\" -c max_locks_per_transaction=512 -c hba_file=$HBA_FILE -k $PGHOST -c log_statement=\"all\" " \
```
steve-chavez: No coverage failure on https://github.com/PostgREST/postgrest/pull/4578, looks like the `max_locks_per_transaction` is the culprit here (although the logs don't say anything). It's not really needed so I'll remove.

laurenceisla (approved):

steve-chavez:
Added 20K tables now and left adding many functions as TODO (this was somehow making the loadtest slow). This now proves the memory consumption is high for the fuzzy search, I've also added a `sleep` for 5 seconds so it ensures the high memory consumption doesn't come from the schema cache.

nix/tools/withTools.nix
```diff
@@ -426,6 +427,10 @@ let
           ${monitorPid} "$pid" > "$_arg_monitor" &
         fi
 
+        if [[ -n "$_arg_sleep" ]]; then
+          sleep "$_arg_sleep"
+        fi
```
wolfgangwalther, line 432: Looks like this argument has a default of 0, so the condition will always be true. Should work to just call `sleep` all the time - with 0 as argument if the default is used.

steve-chavez: Fixed in https://github.com/PostgREST/postgrest/commit/76e0e1fbffbce459a3a5182b16cb366bb09e08ac, just removed the default instead.

Issue #4088, opened 2025-05-14
Support `db-error-verbosity` config to customize error verbosity


steve-chavez:
## Problem

Users want to customize the error verbosity in some cases:

- https://github.com/PostgREST/postgrest/issues/3980
- https://github.com/PostgREST/postgrest/issues/3824#issuecomment-2880320331

## Solution

Support a new config, called `db-error-verbosity`, which supports the following values:

```
typedef enum
{
	PQERRORS_TERSE,				/* single-line error messages */
	PQERRORS_DEFAULT,			/* recommended style */
	PQERRORS_VERBOSE,			/* all the facts, ma'am */
	PQERRORS_SQLSTATE			/* only error severity and SQLSTATE code */
} PGVerbosity;
```

Then apply this to the database sessions using [PQsetErrorVerbosity](https://postgrespro.com/docs/postgresql/16/libpq-control.html#LIBPQ-PQSETERRORVERBOSITY).

Lower verbosities should nullify the `hint` and `details` of the JSON error message. I'm lead to believe even the  `message` can be nullified, but I haven't tried it.

taimoorzaeem:
> Then apply this to the database sessions using [PQsetErrorVerbosity](https://postgrespro.com/docs/postgresql/16/libpq-control.html#LIBPQ-PQSETERRORVERBOSITY).

Would this require a `hasql` or `postgresql-libpq` patch?

steve-chavez:
Looks like it's already available in postgresql-libpq:

https://github.com/haskellari/postgresql-libpq/blob/master/src/Database/PostgreSQL/LibPQ/FFI.hs#L113-L114

steve-chavez:
> Wouldn't it be nicer if we have this available through hasql because it is our primary interface for db connection?

Yes, agree.

Note: We already have pending `hasql` improvements like : https://github.com/PostgREST/postgrest/issues/3673 (https://github.com/nikita-volkov/hasql-transaction/issues/22). I've added a new issue label for `hasql`.

taimoorzaeem:
In `postgresql-libpq`, we have `setErrorVerbosity` function which uses the following enum type:

https://github.com/haskellari/postgresql-libpq/blob/55c3fb2cce36883499499a575dd42d093368d4df/src/Database/PostgreSQL/LibPQ/Enums.hsc#L196-L205

```hs
data Verbosity
    = ErrorsTerse
    | ErrorsDefault
    | ErrorsVerbose
[...]

Issue #2987, opened 2023-10-03
Config flag for enabling/disabling preferences


steve-chavez:
## Problem

Certain `Prefer` options need to be disabled in some cases, see https://github.com/PostgREST/postgrest/issues/2777.

We also have a special config for `Prefer: tx=commit`, see [db-tx-end](https://postgrest.org/en/stable/references/configuration.html#db-tx-end).

Ideally we wouldn't need a special config for each case.

## Proposal

Have a single config that enables or disables all preferences:

```
-- by default all preferences are enabled, so empty
server-preferences-disabled = ""

-- disable count: exact
server-preferences-disabled = "count=exact"

-- disable timezone
server-preferences-disabled = "timezone"
```

## Notes

- Maybe we can deprecate `db-tx-end` after this. cc @wolfgangwalther

wolfgangwalther:
> Maybe we can deprecate db-tx-end after this.

This setting does not translate 1:1 to the preferences-disabled, because it also sets a default value.

I'm undecided whether I find a shared preferences-disabled config a good or bad idea.

steve-chavez:
> This setting does not translate 1:1 to the preferences-disabled, because it also set's a default value.

Stretching out our [server-host](https://docs.postgrest.org/en/v14/references/configuration.html#server-host) conventions, perhaps we can represent `db-tx-end` like:

```
server-preferences-allowed = "tx=*commit" -- commit by default but allows overriding
```

I've also changed the config to an allowlist. 

By default it would be:

```
server-preferences-allowed = "tx=commit, timezone, return, ..." 
```

> I'm undecided whether I find a shared preferences-disabled config a good or bad idea

I guess one disadvantage is that it's more error prone, user could forget to put an element inside the list and disable it inadvertently.

PR #4584, opened 2025-12-29
add: per request statement timeout using prefer header

### LINKED ISSUE #4381: Per-request timeout

taimoorzaeem:
Adds a feature to set `statement_timeout` using `Prefer: timeout` header. This also introduces a `PGRST129` error which is returned when the timeout preferred exceeds the per-role timeout, which prevents misuse of this feature.

Closes #4381.

- [ ] Implementation
- [ ] Tests
- [ ] Docs

### DIFF
--- a/docs/references/api/preferences.rst
+++ b/docs/references/api/preferences.rst
@@ -15,6 +15,7 @@ The following preferences are supported.
 - ``Prefer: missing``. See :ref:`prefer_missing`.
 - ``Prefer: max-affected``, See :ref:`prefer_max_affected`.
 - ``Prefer: tx``. See :ref:`prefer_tx`.
+- ``Prefer: timeout``. See :ref:`prefer_timeout`.
 
 .. _prefer_handling:
 
@@ -296,3 +297,62 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
 .. note::
 
   It is important for functions to return ``SETOF`` or ``TABLE`` when called with ``max-affected`` preference. A violation of this would cause a :ref:`PGRST128 <pgrst128>` error.
+
+.. _prefer_timeout:
+
+Timeout
+=======
+
+You can set `statement_timeout <https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT>`_ for the request using this preference. This works in combination with ``handling=strict`` preference in the same header.
+
+The header only accepts integer value indicating the ``seconds`` that are set as timeout value. ``0`` and negative values are ignored and not applied. To demonstrate, see the following example:
+
+.. code-block:: postgres
+
+  CREATE FUNCTION test.sleep(seconds)
+  RETURNS VOID AS $$
+    SELECT pg_sleep(seconds);
+  $$ LANGUAGE SQL;
+
+.. code-block:: bash
+
+  curl -i "http://localhost:3000/rpc/sleep?seconds=5" \
+    -H "Prefer: handling=strict, timeout=2"
+
+.. code-block:: http
+
+  HTTP/1.1 500 Internal Server Error
+
+.. code-block:: json
+
+  {
+      "code": "57014",
+      "details": null,
+      "hint": null,
+      "message": "canceling statement due to statement timeout"
+  }
+
+It is important to note the timeout value cannot exceed the ``statement_timeout`` set :ref:`per-role <impersonated_settings>` or per-database. The role's timeout setting takes precedence over the database level timeout. This restriction prevents misuse of this feature. PostgREST returns a :ref:`PGRST129 <pgrst129>` error in this case.
+
+.. code-block:: postgres
+
+  ALTER ROLE postgrest_test_anonymous SET statement_timeout = '3s';
+
+.. code-block:: bash
+
+  curl -i "http://localhost:3000/rpc/sleep?seconds=4" \
+    -H "Prefer: handling=strict, timeout=5"
+
+.. code-block:: http
+
+  HTTP/1.1 400 Bad Request
+  Content-Type: application/json; charset=utf-8
+
+.. code-block:: json
+
+  {
+      "code": "PGRST129",
+      "message": "Timeout preference exceeded maximum allowed",
+      "details": "The maximum timeout allowed is 3s",
+      "hint": "Reduce the timeout"
+  }
--- a/src/PostgREST/ApiRequest/Preferences.hs
+++ b/src/PostgREST/ApiRequest/Preferences.hs
@@ -6,7 +6,7 @@
 --
 -- [1] https://datatracker.ietf.org/doc/html/rfc7240
 --
-{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
 module PostgREST.ApiRequest.Preferences
   ( Preferences(..)
   , PreferCount(..)
@@ -17,6 +17,7 @@ module PostgREST.ApiRequest.Preferences
   , PreferTransaction(..)
   , PreferTimezone(..)
   , PreferMaxAffected(..)
+  , PreferTimeout(..)
   , fromHeaders
   , shouldCount
   , shouldExplainCount
@@ -43,6 +44,7 @@ import Protolude
 -- >>> deriving instance Show PreferHandling
 -- >>> deriving instance Show PreferTimezone
 -- >>> deriving instance Show PreferMaxAffected
+-- >>> deriving instance Show PreferTimeout
 -- >>> deriving instance Show Preferences
 
 -- | Preferences recognized by the application.
@@ -56,6 +58,7 @@ data Preferences
     , preferHandling       :: Maybe PreferHandling
     , preferTimezone       :: Maybe PreferTimezone
     , preferMaxAffected    :: Maybe PreferMaxAffected
+    , preferTimeout        :: Maybe PreferTimeout
     , invalidPrefs         :: [ByteString]
     }
 
@@ -77,12 +80,13 @@ data Preferences
 --         ( PreferTimezone "America/Los_Angeles" )
 --     , preferMaxAffected = Just
 --         ( PreferMaxAffected 100 )
+--     , preferTimeout = Nothing
 --     , invalidPrefs = []
 --     }
 --
 -- Multiple headers can also be used:
 --
--- >>> pPrint $ fromHeaders True sc [("Prefer", "resolution=ignore-duplicates"), ("Prefer", "count=exact"), ("Prefer", "missing=null"), ("Prefer", "handling=lenient"), ("Prefer", "invalid"), ("Prefer", "max-affected=5999")]
+-- >>> pPrint $ fromHeaders True sc [("Prefer", "resolution=ignore-duplicates"), ("Prefer", "count=exact"), ("Prefer", "missing=null"), ("Prefer", "handling=lenient"), ("Prefer", "invalid"), ("Prefer", "max-affected=5999"), ("Prefer", "timeout=10")]
 -- Preferences
 --     { preferResolution = Just IgnoreDuplicates
 --     , preferRepresentation = Nothing
@@ -93,6 +97,8 @@ data Preferences
 --     , preferTimezone = Nothing
 --     , preferMaxAffected = Just
 --         ( PreferMaxAffected 5999 )
+--     , preferTimeout = Just
+--         ( PreferTimeout 10 )
 --     , invalidPrefs = [ "invalid" ]
 --     }
 --
@@ -124,6 +130,7 @@ data Preferences
 --     , preferHandling = Just Strict
 --     , preferTimezone = Nothing
 --     , preferMaxAffected = Nothing
+--     , preferTimeout = Nothing
 --     , invalidPrefs = [ "anything" ]
 --     }
 --
@@ -138,7 +145,8 @@ fromHeaders allowTxDbOverride acceptedTzNames headers =
     , preferHandling       = parsePrefs [Strict, Lenient]
     , preferTimezone       = if isTimezonePrefAccepted then PreferTimezone <$> timezonePref else Nothing
     , preferMaxAffected    = PreferMaxAffected <$> maxAffectedPref
-    , invalidPrefs         = filter isUnacceptable prefs
+    , preferTimeout        = PreferTimeout <$> timeoutPref
+    , invalidPrefs         = filter (not . isPrefValid) prefs
     }
   where
     mapToHeadVal :: ToHeaderValue a => [a] -> [ByteString]
@@ -159,10 +167,17 @@ fromHeaders allowTxDbOverride acceptedTzNames headers =
     isTimezonePrefAccepted = ((S.member . decodeUtf8 <$> timezonePref) <*> pure acceptedTzNames) == Just True
 
     maxAffectedPref = listStripPrefix "max-affected=" prefs >>= readMaybe . BS.unpack
+    timeoutPref = listStripPrefix "timeout=" prefs >>= readOnlyPositiveInt . readMaybe . BS.unpack
+      where
+        -- 0 and -ve values for timeout are meaningless, we handle them leniently and ignore them
+        readOnlyPositiveInt (Just i) | i > 0 = Just i
+        readOnlyPositiveInt _ = Nothing
 
-    isUnacceptable p = p `notElem` acceptedPrefs &&
-                       (isNothing (BS.stripPrefix "timezone=" p) ||  not isTimezonePrefAccepted) &&
-                       isNothing (BS.stripPrefix "max-affected=" p)
+    isPrefValid p =
+      p `elem` acceptedPrefs ||
+      (isJust (BS.stripPrefix "timezone=" p) && isTimezonePrefAccepted) ||
+      isJust (BS.stripPrefix "max-affected=" p) ||
+      isJust (BS.stripPrefix "timeout=" p)
 
     parsePrefs :: ToHeaderValue a => [a] -> Maybe a
     parsePrefs vals =
@@ -172,7 +187,7 @@ fromHeaders allowTxDbOverride acceptedTzNames headers =
     prefMap = Map.fromList . fmap (\pref -> (toHeaderValue pref, pref))
 
 prefAppliedHeader :: Preferences -> Maybe HTTP.Header
-prefAppliedHeader Preferences {preferResolution, preferRepresentation, preferCount, preferTransaction, preferMissing, preferHandling, preferTimezone, preferMaxAffected } =
+prefAppliedHeader Preferences{..} =
   if null prefsVals
     then Nothing
     else Just (HTTP.hPreferenceApplied, combined)
@@ -187,6 +202,7 @@ prefAppliedHeader Preferences {preferResolution, preferRepresentation, preferCou
       , toHeaderValue <$> preferHandling
       , toHeaderValue <$> preferTimezone
       , if preferHandling == Just Strict then toHeaderValue <$> preferMaxAffected else Nothing
+      , if preferHandling == Just Strict then toHeaderValue <$> preferTimeout else Nothing
       ]
 
 -- |
@@ -289,3 +305,10 @@ newtype PreferMaxAffected = PreferMaxAffected Int64
 
 instance ToHeaderValue PreferMaxAffected where
   toHeaderValue (PreferMaxAffected n) = "max-affected=" <> show n
+
+-- |
+-- Statement Timeout per request
+newtype PreferTimeout = PreferTimeout Int64
+
+instance ToHeaderValue PreferTimeout where
+  toHeaderValue (PreferTimeout n) = "timeout=" <> show n
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -57,7 +57,8 @@ import System.Environment      (getEnvironment)
 import System.Posix.Types      (FileMode)
 
 import PostgREST.Config.Database         (RoleIsolationLvl,
-                                          RoleSettings)
+                                          RoleSettings,
+                                          RoleTimeoutSettings)
 import PostgREST.Config.JSPath           (FilterExp (..), JSPath,
                                           JSPathExp (..), dumpJSPath,
                                           pRoleClaimKey)
@@ -117,6 +118,10 @@ data AppConfig = AppConfig
   , configAdminServerHost          :: Text
   , configAdminServerPort          :: Maybe Int
   , configRoleSettings             :: RoleSettings
+  -- Cached statement timeout settings converted to number of seconds. They
+  -- are never applied, only used to check max allowed timeout for a role
+  -- when "Prefer: timeout=" header is used.
+  , configRoleTimeoutSettings      :: RoleTimeoutSettings
   , configRoleIsoLvl               :: RoleIsolationLvl
   , configInternalSCQuerySleep     :: Maybe Int32
   , configInternalSCLoadSleep      :: Maybe Int32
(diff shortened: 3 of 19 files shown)


docs/references/api/preferences.rst
```diff
@@ -296,3 +297,38 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
...
+
+
+.. _prefer_timeout:
+
+Timeout
+=======
+
+You can set `statement_timeout <https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT>`_ for the request using this preference. This works in combination with ``handling=strict`` preference in the same header.
+
+Currently, the header only accepts integer value indicating the ``seconds`` that are set as timeout value. To demonstrate, see the following example:
```
steve-chavez: I don't think this is going to change:
```suggestion
The header only accepts integer value indicating the ``seconds`` that are set as timeout value. To demonstrate, see the following example:
```

steve-chavez:
We should also fail in case a value exceeds the [per role timeout](https://docs.postgrest.org/en/v13/references/transactions.html#impersonated-role-settings). Otherwise it's abusable.

This failure should happen at the plan level, before hitting the db. I think it will need a new error code.

src/PostgREST/Plan.hs
```diff
@@ -144,29 +146,56 @@ data InfoPlan
...
+  case roleTimeoutSecs of
+    Nothing -> Right ()
+    Just rt -> unless (t <= rt) $ Left $ ApiRequestError $ TimeoutConstraintError t (fromJust roleTimeout) authRole
+    where
+      roleTimeoutSecs = rtValueToSeconds <$> roleTimeout
+      roleTimeout = HM.lookup "statement_timeout" =<< HM.lookup authRole configRoleSettings
+      -- It is safe to use fromJust because the string being read is guaranteed
+      -- to be number after striping the unit.
+      stripUnitAndGetInt unit val = fromJust $ BS.stripSuffix unit val >>= readMaybe
+      rtValueToSeconds :: ByteString -> Int
```
steve-chavez: I'm confused by all of this logic, why do we need to parse a time unit?

taimoorzaeem: > We should also fail in case a value exceeds the [per role timeout](https://docs.postgrest.org/en/v13/references/transactions.html#impersonated-role-settings). Otherwise it's abusable

> I'm confused by all of this logic, why do we need to parse a time unit?

That's because the `statement_timeout` value of the role could be like `1min` or `100ms` depending upon the role. So, we need to throw an error depending on whether the `timeout` exceeds role setting or not and for that, we must parse the value, I think.

steve-chavez: Can we do the conversion in postgres to avoid the parsing? For example with:

```sql
select extract(epoch from current_setting('statement_timeout', false)::interval)::int;
```

That should give you the number of seconds regardless of the input format.

taimoorzaeem: That's great.

However, if we do this conversion at the time of querying the role settings, that would change the original role settings, which we explicitly add to pre-query (BTW, I'm not sure why we do this, wouldn't postgres already know the settings?) here:

https://github.com/PostgREST/postgrest/blob/8f5fe3f46b1f823fa96704cb7c4a024e04b687fe/src/PostgREST/Query/PreQuery.hs#L51

So, we would then have to keep the two sets of role settings, one with the conversion applied on and the original ones without the conversion.

wolfgangwalther: > we explicitly add to pre-query (BTW, I'm not sure why we do this, wouldn't postgres already know the settings?)

The settings set via `ALTER ROLE ...` only apply when *logging in with that role*. PostgreSQL does *not* apply them when doing a `SET ROLE`. That means, these role-specific settings are only set when we do it manually.

steve-chavez: @taimoorzaeem Hm, so we can't avoid this logic with the [above snippet](https://github.com/PostgREST/postgrest/pull/4584/changes#r2679654379)?

taimoorzaeem: Yes. If we try to avoid this logic using the _above snippet_, we would end up with maintaining more state (keep extra pre-computed values, even when they are never used) so I think we should avoid it.

steve-chavez: Sorry for late reply here.

> If we try to avoid this logic using the above snippet, we would end up with maintaining more state (keep extra pre-computed values, even when they are never used) so I think we should avoid it.

IMO we should cache the pg transformed value. 

- Having to repeat this transformation logic for each request at the Plan level would be inefficient.
- It's not much more memory considering we only take child roles of the connection role:
https://github.com/PostgREST/postgrest/blob/ffa393850a461ed02a0906833a52efe5fa844d43/src/PostgREST/Config/Database.hs#L145
- Also later on we could avoid this caching if `Prefer: timeout` is not used with https://github.com/PostgREST/postgrest/issues/2987.
- There is more risk for errors with this Haskell logic.. say if pg changes format of the statement_timeout, very unlikely but overall it feels safer to have the transformation at the db level.

src/PostgREST/Config.hs
```diff
@@ -117,6 +118,10 @@ data AppConfig = AppConfig
   , configAdminServerHost          :: Text
   , configAdminServerPort          :: Maybe Int
   , configRoleSettings             :: RoleSettings
+  -- Cached statement timeout settings converted to number of seconds. They
+  -- are never applied, only used to check max allowed timeout for a role
+  -- when "Prefer: timeout=" header is used.
+  , configRoleTimeoutSettings      :: RoleTimeoutSettings
```
taimoorzaeem, line 124: > IMO we should cache the pg transformed value

@steve-chavez How about an internal config to store the transformed values?

steve-chavez: Yes, this looks much better :+1:

taimoorzaeem:
This looks ready for review.

docs/references/api/preferences.rst
```diff
@@ -296,3 +297,62 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
...
+  $$ LANGUAGE SQL;
+
+.. code-block:: bash
+
+  curl -i "http://localhost:3000/rpc/sleep?seconds=5" \
+    -H "Prefer: handling=strict, timeout=2"
+
+.. code-block:: http
+
+  HTTP/1.1 500 Internal Server Error
```
steve-chavez, line 324: I was thinking about mapping this error to 503 but this isn't correct as 503 indicates temporary state on which a later retry can succeed; this is not true for example when an index is missing on a column, all retries would fail the same. So 500 is the most apt code here.

That also reminds me of https://github.com/PostgREST/postgrest/issues/2967#issuecomment-2713224061, we should put `statement_timeout` on the Proxy-Status.

docs/references/api/preferences.rst
```diff
@@ -296,3 +297,62 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
...
+.. code-block:: http
+
+  HTTP/1.1 400 Bad Request
+  Content-Type: application/json; charset=utf-8
+
+.. code-block:: json
+
+  {
+      "code": "PGRST129",
+      "message": "Timeout preference value cannot exceed statement_timeout of role",
```
steve-chavez: Maybe the message should give less details:
```suggestion
      "message": "Timeout preference exceeded maximum allowed",
```
Otherwise if users want to hide more sensitive details with https://github.com/PostgREST/postgrest/issues/4088, then it would not be possible. (since we're only going to hide the details/hint).

docs/references/api/preferences.rst
```diff
@@ -296,3 +297,62 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
...
   It is important for functions to return ``SETOF`` or ``TABLE`` when called with ``max-affected`` preference. A violation of this would cause a :ref:`PGRST128 <pgrst128>` error.
+
+.. _prefer_timeout:
+
+Timeout
+=======
+
+You can set `statement_timeout <https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-STATEMENT-TIMEOUT>`_ for the request using this preference. This works in combination with ``handling=strict`` preference in the same header.
+
+The header only accepts integer value indicating the ``seconds`` that are set as timeout value. To demonstrate, see the following example:
```
steve-chavez: Should we indicate here that `timeout=0` is a noop?

src/PostgREST/ApiRequest/Preferences.hs
```diff
@@ -171,8 +182,9 @@ fromHeaders allowTxDbOverride acceptedTzNames headers =
     prefMap :: ToHeaderValue a => [a] -> Map.Map ByteString a
     prefMap = Map.fromList . fmap (\pref -> (toHeaderValue pref, pref))
 
+
```
steve-chavez:
```suggestion
```

test/io/test_prefer_header.py
```diff
@@ -0,0 +1,103 @@
...
+        assert response.headers["Preference-Applied"] == "handling=strict, timeout=2"
+
+
+def test_prefer_timeout_header_with_lenient_handling(defaultenv):
+    "Test Prefer: timeout header with handling=lenient"
+
+    with run(env=defaultenv) as postgrest:
+        # Timeout header is ignored for lenient handling
+        headers = {"Prefer": "handling=lenient, timeout=2"}
+        response = postgrest.session.get("/rpc/sleep?seconds=1", headers=headers)
```
steve-chavez: Otherwise this would always succeed even with strict?
```suggestion
        headers = {"Prefer": "handling=lenient, timeout=1"}
        response = postgrest.session.get("/rpc/sleep?seconds=2", headers=headers)
```

src/PostgREST/Config/Database.hs
```diff
@@ -155,33 +157,46 @@ queryRoleSettings pgVer prepared =
         SELECT rolname, value
         FROM kv_settings
         WHERE key = 'default_transaction_isolation'
+      ),
+      role_timeout_setting AS (
+        SELECT rolname, extract(epoch from value::interval)::int as value -- transform to seconds
+        FROM kv_settings
+        WHERE key = 'statement_timeout'
```
steve-chavez: Forgot about this, if there's a global `statement_timeout` (this can be set at the cluster level or at the db level) but no role timeout, it would be bad if we allowed clients to surpass this global too. So we would need to refine the query to obtain the global timeout as implicit default for the role.

docs/references/api/preferences.rst
taimoorzaeem: > That also reminds me of [#2967 (comment)](https://github.com/PostgREST/postgrest/issues/2967#issuecomment-2713224061), we should put `statement_timeout` on the Proxy-Status.

Right, I see that as a separate improvement, opened #4637 for it.

docs/references/api/preferences.rst
taimoorzaeem: IMO, _ideally_ we should throw an error on `0` and `-ve` numbers but that would be too much and a bit needless to handle for now.

> Should we indicate here that `timeout=0` is a noop?

Yes, let's keep it simple and document this for now.

src/PostgREST/Config/Database.hs
taimoorzaeem: I am finding it a bit messy to a write a test with global settings :thinking:. Changing global `statement_timeout` might affect other tests. Any recommendations on how to test this behavior?

**EDIT**: I figured out a way in latest push!!

docs/references/api/preferences.rst
```diff
@@ -296,3 +297,62 @@ With :ref:`RPC <functions>`, the preference is honored completely on the basis o
...
+  Content-Type: application/json; charset=utf-8
+
+.. code-block:: json
+
+  {
+      "code": "PGRST129",
+      "message": "Timeout preference exceeded maximum allowed",
+      "details": "Timeout preferred: 5s, statement_timeout of role 'postgrest_test_anonymous': 3s",
+      "hint": null
+  }
```
taimoorzaeem, line 358: I think maybe we shouldn't reveal the role's actual timeout value due to security concerns.

Alternatively, maybe we can drop the `details` part entirely and add a `hint` like:

```json
"code": "PGRST129",
"message": "Timeout preference exceeded maximum allowed",
"details": null,
"hint": "Make sure the preferred timeout not exceed the database or role level 'statement_timeout'"
```

test/io/test_prefer_header.py
```diff
@@ -0,0 +1,150 @@
...
+def test_prefer_timeout_header_with_global_timeout(defaultenv):
+    "Test Prefer: timeout header with global timeout"
+
+    env = {**defaultenv, "PGRST_JWT_SECRET": SECRET}
+
+    with run(env=env) as postgrest:
+        # set global statement_timeout to 3s using postgres superuser
+        os.system(
+            f'psql -d {env["PGDATABASE"]} -U postgres -h {env["PGHOST"]} --set ON_ERROR_STOP=1 -a -c "ALTER DATABASE postgres SET statement_timeout TO \'3s\';"'
+        )
```
steve-chavez, line 119: Nice, you nailed it 💯

test/io/fixtures/roles.sql
```diff
@@ -1,18 +1,23 @@
...
+  postgrest_test_timeout_s, postgrest_test_timeout_global;
 
 CREATE ROLE postgrest_test_anonymous;
 CREATE ROLE postgrest_test_author;
 CREATE ROLE postgrest_test_serializable;
 CREATE ROLE postgrest_test_repeatable_read;
 CREATE ROLE postgrest_test_w_superuser_settings;
+CREATE ROLE postgrest_test_timeout_ms;
+CREATE ROLE postgrest_test_timeout_s;
+CREATE ROLE postgrest_test_timeout_global; -- no timeout for this role, default to global (db/cluster) statement_timeout
```
steve-chavez: Maybe a more descriptive name? Should help with readability on the tests.
```suggestion
CREATE ROLE postgrest_test_no_timeout; -- no timeout for this role, default to global (db/cluster) statement_timeout
```

src/PostgREST/Config/Database.hs
```diff
@@ -173,8 +176,8 @@ queryRoleSettings pgVer prepared =
     processRows :: [(Text, Maybe Text, [(Text, Text)])] -> (RoleSettings, RoleIsolationLvl)
     processRows rs =
       let
-        rowsWRoleSettings = [ (x, z) | (x, _, z) <- rs ]
-        rowsWIsolation    = [ (x, y) | (x, Just y, _) <- rs ]
+        rowsWRoleSettings = [ (x, z)  | (x, _, z) <- rs ]
+        rowsWIsolation    = [ (x, y)  | (x, Just y, _) <- rs ]
```
steve-chavez: We better not change unrelated lines, helps for having a cleaner git history.
```suggestion
        rowsWRoleSettings = [ (x, z) | (x, _, z) <- rs ]
        rowsWIsolation    = [ (x, y) | (x, Just y, _) <- rs ]
```

src/PostgREST/Config/Database.hs
```diff
@@ -183,6 +186,45 @@ queryRoleSettings pgVer prepared =
...
+queryRoleTimeoutSettings prepared =
+  let transaction = if prepared then SQL.transaction else SQL.unpreparedTransaction in
+  transaction SQL.ReadCommitted SQL.Read $ SQL.statement mempty $ SQL.Statement sql HE.noParams (processRows <$> rows) prepared
+  where
+    sql = encodeUtf8 [trimming|
+      -- For roles without statement_timeout, we fetch global stmnt timeout
+      -- as an implicit timeout for that role.
+      -- IMPORTANT DISTINCTION:
+      --   In pg, 0 indicates no timeout. However, in our approach, the
+      -- cached value 0 indicates timeout < 1s, and NULL indicate no timeout.
```
steve-chavez: I'm a bit confused by this comment.

>  However, in our approach, the cached value 0 indicates timeout < 1s

Is the purpose of the above to describe that we reject `Prefer: timeout=0` at the Plan level?

---

One idea to improve this would be to have a comment on a type alias.

src/PostgREST/Query/PreQuery.hs
```diff
@@ -50,7 +52,12 @@ txVarQuery dbActPlan AppConfig{..} AuthResult{..} ApiRequest{..} =
     roleSql = [setConfigWithConstantName ("role", authRole)]
     roleSettingsSql = setConfigWithDynamicName <$> HM.toList (fromMaybe mempty $ HM.lookup authRole configRoleSettings)
     appSettingsSql = setConfigWithDynamicName . join bimap toUtf8 <$> configAppSettings
-    timezoneSql = maybe mempty (\(PreferTimezone tz) -> [setConfigWithConstantName ("timezone", tz)]) $ preferTimezone iPreferences
+    timezoneSql = maybe mempty (\(PreferTimezone tz) -> [setConfigWithConstantName ("timezone", tz)]) preferTimezone
+    timeoutSql = maybe mempty (\(PreferTimeout t) ->
+      if preferHandling == Just Strict && t > 0 -- 0 and -ve numbers are meaningless
```
steve-chavez: Would 0 and negatives reach this module? They would be rejected at the Plan level before they reach Query right?

docs/references/api/preferences.rst
steve-chavez: > "hint": "Make sure the preferred timeout not exceed the database or role level 'statement_timeout'"

One problem with that `hint` is that it doesn't help the API users (they have no say on the db settings), which are really the target of the client errors. Although it might help an API developer.

---

> I think maybe we shouldn't reveal the role's actual timeout value due to security concerns.

How about:

```
"details": "The maximum timeout allowed is 3s",
"hint": "Reduce the timeout"
```

That way API users are in the dark about the role name or db settings.

src/PostgREST/Query/PreQuery.hs
taimoorzaeem: Yes, these values should not reach this module. We decided to be lenient on `-ve` and `0` values and not fail.

Now, I have handled these values at parse time.

src/PostgREST/Config/Database.hs
taimoorzaeem: Yeah, I realize now that this is confusing. I was trying the explain the query logic. I have now added a detailed comment.

src/PostgREST/Config/Database.hs
```diff
@@ -183,6 +186,59 @@ queryRoleSettings pgVer prepared =
...
+      --
+      -- In our logic, we cache statement_timeout values after transforming
+      -- them to seconds. The transformation results in having a case of:
+      --
+      -- +-----------------------------------------------------------+
+      -- |   Role   |   Timeout Setting  | Cached Timeout in Seconds |
+      -- +-----------------------------------------------------------+
+      -- |   role1  |     0s             |             0             |
+      -- +-----------------------------------------------------------+
+      -- |   role2  |     300ms          |             0             |
```
steve-chavez: Hm, this part confuses me because it should be 0.3 in seconds. Maybe say it's truncated?

src/PostgREST/Config/Database.hs
```diff
@@ -183,6 +186,59 @@ queryRoleSettings pgVer prepared =
...
     rows = HD.rowList $ (,,) <$> column HD.text <*> nullableColumn HD.text <*> compositeArrayColumn ((,) <$> compositeField HD.text <*> compositeField HD.text)
 
+queryRoleTimeoutSettings :: Bool -> Session RoleTimeoutSettings
+queryRoleTimeoutSettings prepared =
+  let transaction = if prepared then SQL.transaction else SQL.unpreparedTransaction in
+  transaction SQL.ReadCommitted SQL.Read $ SQL.statement mempty $ SQL.Statement sql HE.noParams (processRows <$> rows) prepared
+  where
+    sql = encodeUtf8 [trimming|
+      -- For roles without statement_timeout, we fetch database/cluster
+      -- statement_timeout as an implicit timeout for that role.
```
steve-chavez: Could we have this comment at the top of the function? Otherwise I believe the comment will be stored in `pg_stat_statements`, generating some noise for observability.

src/PostgREST/Config/Database.hs
```diff
@@ -183,6 +186,59 @@ queryRoleSettings pgVer prepared =
...
+      -- +-----------------------------------------------------------+
+      -- |   role2  |     300ms          |             0             |
+      -- +-----------------------------------------------------------+
+      --
+      -- For role1, 0s setting mean statement_timeout is "disabled".
+      -- For role2, statement_timeout is 300ms
+      --
+      -- However, both get cached to the same value. So, to differentiate
+      -- these two distinct cases, the query caches "disabled" timeout as NULL
+      -- and < 1s timeouts as 0.
```
steve-chavez: Which line of code does this?

taimoorzaeem: It's this one with the `nullif()` function:

```hs
-- src/PostgREST/Config/Database.hs:241:
      SELECT
        rolname,
        nullif(extract(epoch FROM statement_timeout::interval)::numeric, 0)::int as statement_timeout_in_seconds
      FROM role_timeouts
```

The `< 1s` automatically gets truncated to 0 because of `::int` cast.

src/PostgREST/Plan.hs
```diff
@@ -144,29 +145,42 @@ data InfoPlan
...
+        MTVndPlan{} -> DbCrud True pl
+        _           -> DbCrud False pl
+
+-- | Fail when Prefer: timeout value is greater than the statement_timeout
+--   setting of the role.
+failPreferTimeout :: Preferences -> AppConfig -> AuthResult -> Either Error ()
+failPreferTimeout Preferences{preferTimeout=(Just (PreferTimeout t)), preferHandling=Just Strict} AppConfig{..} AuthResult{..} =
+  case roleTimeout of
+    Nothing -> Right ()
+    Just rt -> when (t > rt) $ Left $ ApiRequestError $ TimeoutConstraintError rt
```
mkleczek, line 180: This check is problematic because it undermines the whole idea of this header.

Who would want to set the timeout header to a value less than the role setting? What for?

The only realistic use case is that a client might want to **increase** `statement_timeout` for a particular query.

So sensible limit is greater than the role setting. Don't know: maybe 2x role setting should be the limit? But then - why not simply set the role setting to a greater value?

Thinking about it: it looks like having any limit is contradictory to the idea of this header. OTOH not having a limit is a security vulnerability. No good choice here, I'm afraid.

@taimoorzaeem @steve-chavez WDYT?

steve-chavez: > This check is problematic because it undermines the whole idea of this header.
> Thinking about it: it looks like having any limit is contradictory to the idea of this header. OTOH not having a limit is a security vulnerability. No good choice here, I'm afraid.

I think you're right. My idea with the limit was to make the header safe to expose to untrusted clients but that makes it less useful. Also, it might never be totally safe because setting a `1s` timeout could force some queries to fail and flood logs on the server.

So this header should fall in the same category as [prefer:tx](https://docs.postgrest.org/en/latest/references/api/preferences.html#transaction-end-preference) and  [Accept: application/vnd.pgrst.plan](https://docs.postgrest.org/en/latest/references/observability.html#explain-plan), it can only be exposed to trusted users.

This means it should be gated by a config and that we should remove the statement_timeout limit.

Thoughts?

---

To expose this header to an untrusted network, a similar thing to https://docs.postgrest.org/en/latest/references/observability.html#securing-the-execution-plan could be used.

wolfgangwalther: > Who would want to set the timeout header to a value less than the role setting? What for?

In way this would be similar to soft and hard limits in Linux, set via `ulimit`. You can vary the soft limit on your own as long as you stay below the hard limit. Just because there is some higher limit available doesn't mean that you will always make full use of it.

I think it only really becomes useful when you have both:
- a low default for the soft limit, which you can override per request, *and*
- a higher hard limit

IIUC, we currently only have the limit defaulting to the hard limit, which then becomes the odd case of "you can only lower your limit".

Edit: This specifically only answers this question - I'm not saying we should implement any of this and also made no judgement on the PR or the related issue.

mkleczek: > > Who would want to set the timeout header to a value less than the role setting? What for?
> 
> In way this would be similar to soft and hard limits in Linux, set via `ulimit`. 

To be honest, I always had a problem with understanding the purpose of soft limits. The only sensible use case IMHO is "grace periods" - ie. exceeding soft limit triggers a warning and requires a corrective action in a specified time period. But that really only makes sense for disk quotas etc. (and is not exactly the use case of user-provided limit).

Our situation is: a user has a query to run. The query will take x seconds to execute or will fail if the limit is lower. As a user I want to make sure my query does not fail so the only sensible strategy (as I don't know how long it will take) is not to set any limit by myself (and deal with possibility that system enforced limit is lower).

Of course - we could implement tracking of resource usage (per user? per tenant?) and have a policy like `allow 10 queries above soft limit per month` - but I don't think we want to do that :)

PR #4596, opened 2026-01-09
nix: add script and ci workflow to enforce changelog entries

### LINKED ISSUE #4590: Add CI workflow to enforce additions in CHANGELOG.md
## Problem

We often forget to add the entry in `CHANGELOG.md` at the right spot which then needs to be corrected later as occurred in #4531 and #4589.

## Solution

Have a CI workflow to check that on certain commit prefixes like `fix:` and `add:`, the CHANGELOG.md should contain the entry/change at the right spot using `sed` or `git` commands.

taimoorzaeem:
Closes #4590.

- [ ] Create nix script.
- [ ] Add this to CI.

I would love some feedback on this.

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -98,7 +98,7 @@ postgrest-lint                    postgrest-with-pg-16
 postgrest-loadtest                postgrest-with-pg-17
 postgrest-loadtest-against        postgrest-with-slow-pg
 postgrest-loadtest-report         postgrest-with-slow-postgrest
-postgrest-nixpkgs-upgrade
+postgrest-nixpkgs-upgrade         postgrest-changelog-check
 ...
 
 [nix-shell]$
--- a/nix/tools/gitTools.nix
+++ b/nix/tools/gitTools.nix
@@ -52,9 +52,56 @@ let
 
         ${commitlint}/bin/commitlint --config ${commitlintConfig} --from "$_arg_from" --to "$_arg_to"
       '';
+
+  changelogCheck =
+    checkedShellScript
+      {
+        name = "postgrest-changelog-check";
+        docs = "Script to validate changelog entries";
+        workingDir = "/";
+        args = [
+          "ARG_OPTIONAL_SINGLE([base],, [base commit ref], [origin/main])"
+          "ARG_OPTIONAL_SINGLE([head],, [head commit ref], [HEAD])"
+        ];
+      }
+      ''
+        # Get all commit hashes and store in ADD_OR_FIX_COMMITS array
+        mapfile -t ADD_OR_FIX_COMMITS < <(git log "$_arg_base".."$_arg_head" --format=%H -E --grep="^(add|fix)")
+
+        # Loop over each "add:" or "fix:" commit and compare the changelog
+        # of those commits with the base(main) commit.
+        for add_or_fix_commit in "''${ADD_OR_FIX_COMMITS[@]}"; do
+          
+          # Get Unreleased section of the selected 'add: ...' or 'fix: ...' commit
+          commit_changelog=$(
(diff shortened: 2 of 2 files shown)


wolfgangwalther:
Years ago I worked a bit on #1805, but this never really lead anywhere. I believe if we want any kind of checks for changelog (or improve the process to maybe only have a single source of truth or whatever), then we need to store the changelog in a *structured way*, not in the freetext format we have right now. I am skeptical of trying to read the current changelog and have any checks on it.

(I did not actually look much at the code itself, only very briefly skimmed it)

steve-chavez:
> I believe if we want any kind of checks for changelog (or improve the process to maybe only have a single source of truth or whatever), then we need to store the changelog in a structured way, not in the freetext format we have right now. I am skeptical of trying to read the current changelog and have any checks on it.

Agree with the structured format. I also skimmed at the PR code before and looked like the check could fail in certain ways.

taimoorzaeem:
I like the idea of a structured format as well. Just like in #1805, maybe we can have a format like:

```yaml
changelog:
  - version: 14.0
    date: 2025-11-05
  added:
    - description: Bounded JWT cache using the SIEVE algorithm
      details:
        - It now uses a fixed size cache instead of arbitrary sized cache.
      author: mkleczek
      refs: [4084]
  fixed:
    - description: Fix not logging OpenAPI queries when `log-query=main-query` is enabled
      author: steve-chavez
      refs: [4226]
    - description: Fix not logging explain query when `log-query=main-query` is enabled
      author: steve-chavez
      refs: [4319]
  changed:
    - description: Drop support for PostgreSQL EOL version 12
      author: wolfgangwalther
      refs: [4319]
```

wolfgangwalther:
> Just like in #1805, maybe we can have a format like:

A few high-level thoughts after having done similar things in the past:
- The structured format is good for tooling, but not to read. Do we want to keep the current changelog file, but essentially recreate it from the structured data? Would we do that as part of a release, so that unreleased changes would not appear in the changelog, yet?
- The proposed format has the same problem as any other format residing in a single file: Merge conflicts are annoying. Some changelog tools work by adding new changelog entries as *separate* file with arbitrary names in a folder, for example `changes/`. Each file represents one change. The changes are then transformed into something else at the time of release, the folder is cleared. This will prevent all kinds of merge conflicts, and will also prevent changes from going into the wrong changelog section after rebasing.

steve-chavez:
> Do we want to keep the current changelog file, but essentially recreate it from the structured data?

I don't think that would work given given old releases didn't use the commit prefixes?

> Some changelog tools work by adding new changelog entries as separate file with arbitrary names in a folder, for example changes/

Sounds interesting, which tool does that? :eyes:

wolfgangwalther:
> I don't think that would work given given old releases didn't use the commit prefixes?

You can always append the existing changelog to the newly generated part.

> Sounds interesting, which tool does that? 👀

There are probably multiple, I can't remember which tool I saw this first on. A google search brings up https://github.com/churchtools/changelogger.

taimoorzaeem:
> Some changelog tools work by adding new changelog entries as separate file with arbitrary names in a folder, for example `changes/`. Each file represents one change. The changes are then transformed into something else at the time of release, the folder is cleared. This will prevent all kinds of merge conflicts, and will also prevent changes from going into the wrong changelog section after rebasing.

> There are probably multiple, I can't remember which tool I saw this first on. A google search brings up https://github.com/churchtools/changelogger.

@wolfgangwalther What you are describing seems very similar to https://github.com/twisted/towncrier, is this the one?

wolfgangwalther:
The name sounds familiar, could be it, not 100% sure.

taimoorzaeem:

PR #4601, opened 2026-01-14
add: Self links

### LINKED ISSUE #4236: HATEOAS Resource Embeddings (_self link)
## Problem

It would be really useful for PostgREST to generate links to embedded resources. That would make it easier to implement UIs allowing navigation between related resources.

## Solution

Provide `_self` syntax, eg:

`http://localhost:3000/films?select=title,director:directors(last_name,_self)` - provide `last_name` and URI to full representation
`http://localhost:3000/films?select=title,_self` - provide title and URI to full representation
`http://localhost:3000/films?select=title,director:directors(last_name,_self(films))` - provide `last_name` and URI to retrieve director's films

mkleczek:
Add row-level self links so clients can discover canonical resource URIs directly from query results, including embedded and spread-shaped payloads. This improves hypermedia-style navigation without requiring clients to reconstruct key-based URLs manually.

Implemented `_self` as a first-class select item in planning/query generation. Reused location/header PK binding logic to build _self URLs. Added validation to reject _self on relations without primary keys (new error code `PGRST129`).

Fixes #4236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file. From versio
 - Log host, port and pg version of listener database connection by @mkleczek in #4617 #4618
 - Optimize requests with `Prefer: count=exact` that do not use ranges or `db-max-rows` by @laurenceisla in #3957
   + Removed unnecessary double count when building the `Content-Range`.
+- Add `_self` links in `select` projections, including aliases, nesting and spread embeds by @mkleczek in #4601
 
 ### Fixed
 


src/PostgREST/Query/SqlFragment.hs
```diff
@@ -272,8 +272,24 @@ pgFmtCoerceNamed CoercibleField{cfName=fn, cfTransform=(Just formatterProc)} = p
...
+          partsArr = "array_remove(ARRAY[" <> intercalateSnippet ", " (buildPart <$> pkCols) <> "], NULL)"
+          -- Use the base table name for the path (strip trailing _N aliases like clients_1 -> clients)
+          tblName = qiName table
+          parts = T.splitOn "_" tblName
+          baseName = case reverse parts of
+            (s:rest) | T.all isDigit s && not (null rest) -> T.intercalate "_" (reverse rest)
+            _ -> tblName
+          urlExpr = "CASE WHEN array_length(" <> partsArr <> ",1) = 0 THEN NULL ELSE " <>
+              SQL.sql (encodeUtf8 $ pgFmtLit ("/" <> baseName)) <>
+              " || '?' || array_to_string(" <> partsArr <> ", '&') END"
```
wolfgangwalther: I believe we already have logic elsewhere to create `Location` headers, which should result in very similar URIs. We should probaly re-use the same URI generator for both?

steve-chavez: I think we should clarify how would this feature be useful ([ref](https://github.com/PostgREST/postgrest/issues/4236#issuecomment-3752720475)) before going to the implementation.

mkleczek: > I believe we already have logic elsewhere to create `Location` headers, which should result in very similar URIs. We should probaly re-use the same URI generator for both?

@wolfgangwalther done in new version

mkleczek:
This PR was implemented with the help of an LLM tool. Since it is inconsistent with the [project policy on AI usage](https://github.com/PostgREST/postgrest/blob/main/CONTRIBUTING.md#ai-policy) (see #4683) I am closing it.

PR #4603, opened 2026-01-15
add: string slicing operator for `jwt-role-claim-key`

### LINKED ISSUE #4599: JWT Role Extraction for WLCG IAM JWT
### Discussed in https://github.com/PostgREST/postgrest/discussions/4594

<sup>Originally posted by **merl1n0** January  8, 2026</sup>
Hello to all, 
I'm trying to derive a `role` from a [WLCG JWT profile](https://zenodo.org/record/3460257) token. 
The role should be extracted from the `wlcg.groups` claim which is defined an ordered list of group names **prepended by a `/`**.
[Jwt Role Extraction](https://docs.postgrest.org/en/stable/references/auth.html#jwt-role-extraction)  cannot be used directly:

```text
# {"postgrest":{"wlcg.groups": ["/groupa", "/groupb", "/groupa/subgroupa1","/groupa/subgroupa2"]}}
jwt-role-claim-key = ".postgrest.wlcg.groups[0]"
```

This would return `/groupa`, instead of `groupa`.  

Is there any way to handle a similar case? (i.e. trim away the first character).

Alternatively, I could extract the role from the `scope` claim:

```text
"scope": "foo bar role:groupa zoo xywz"
```

But again that would need an extraction operator more powerful than those currently provided for JWT Role Extraction. </div>

> merl1n0:
>   What I would need in this case would be something like: `jwt-role-claim-key = '.postgrest.wlcg.groups[0][1:]`, i.e. `ltrim('/groupa','/')`

> taimoorzaeem:
>   I understand the use case now. I think we can support a slice operator like `[1:0]` on string values.

> steve-chavez:
>   Yes, agree.

taimoorzaeem:
Closes #4599.

- [x] Implementation
- [x] Tests
- [x] Docs: https://docs.postgrest.org/en/latest/references/auth.html#jwt-role-extract

### DIFF
--- a/docs/references/auth.rst
+++ b/docs/references/auth.rst
@@ -234,12 +234,16 @@ The DSL follows the `JSONPath <https://goessner.net/articles/JsonPath/>`_ expres
 - ``==^`` selects the first array element that ends with the right operand
 - ``*==`` selects the first array element that contains the right operand
 
-The selected role value can also be sliced using the slice operator ``[a:b]``. Negative index values are also supported. The syntax is as:
+The selected role value can also be sliced using the slice operator ``[a:b]``. It is similar to `slice operator in python <https://docs.python.org/3/library/functions.html#slice>`_. Negative index values are also supported. The syntax is as:
 
 - ``[a:b]`` take slice from index ``a`` up to ``b``
 - ``[a:]`` take slice from index ``a`` to end
 - ``[:b]`` take slice from start to index ``b``
 - ``[:]`` select everything, no slicing
+
+.. important::
+
+  Make sure that you are not taking a slice where the start index comes after the end index like ``[11:2]``. The result of this would be empty string and so no role would get selected.
 
 Usage examples:
 


src/PostgREST/Config/JSPath.hs
```diff
@@ -29,9 +29,10 @@ type JSPath = [JSPathExp]
 -- NOTE: We only accept one JSPFilter expr (at the end of input)
 -- | jspath expression
 data JSPathExp
-  = JSPKey Text         -- .property or ."property-dash"
-  | JSPIdx Int          -- [0]
-  | JSPFilter FilterExp -- [?(@ == "match")]
+  = JSPKey Text                      -- .property or ."property-dash"
+  | JSPIdx Int                       -- [0]
+  | JSPSlice (Maybe Int) (Maybe Int) -- [0:5] or [0:] or [:5] or [:]
```
steve-chavez, line 34: 👀 What does `[:]` mean?

src/PostgREST/Config/JSPath.hs
```diff
@@ -59,16 +61,30 @@ walkJSPath :: Maybe JSON.Value -> JSPath -> Maybe JSON.Value
...
+    EqualsCond txt     -> walkJSPath (findFirstMatch (==) txt ar) rest
+    NotEqualsCond txt  -> walkJSPath (findFirstMatch (/=) txt ar) rest
+    StartsWithCond txt -> walkJSPath (findFirstMatch T.isPrefixOf txt ar) rest
+    EndsWithCond txt   -> walkJSPath (findFirstMatch T.isSuffixOf txt ar) rest
+    ContainsCond txt   -> walkJSPath (findFirstMatch T.isInfixOf txt ar) rest
   where
     findFirstMatch matchWith pattern = find (\case
       JSON.String txt -> pattern `matchWith` txt
       _               -> False)
+
```
steve-chavez: Empty line?
```suggestion
```

src/PostgREST/Config/JSPath.hs
steve-chavez: Oh, just saw it in the tests: it does nothing.

I think we should mention this in the docs.

src/PostgREST/Config/JSPath.hs
```diff
@@ -95,13 +111,25 @@ pJSPIdx = do
...
+      end'   = if isJust endSign   then ((-1) *) <$> endIndex   else endIndex
+  return (JSPSlice start' end') <?> "pJSPSlice: JSPath string slice"
+
 pJSPFilter :: P.Parser JSPathExp
 pJSPFilter = do
   P.try $ P.string "[?("
   condition <- pFilterConditionParser
   P.char ')'
   P.char ']'
-  P.eof -- this should be the last jspath expression
```
steve-chavez, line 104: Do we get a good error message now that `eof` is gone?

src/PostgREST/Config/JSPath.hs
```diff
@@ -80,7 +96,7 @@ pJSPath :: P.Parser JSPath
 pJSPath = P.many1 pJSPathExp <* P.eof
 
 pJSPathExp :: P.Parser JSPathExp
-pJSPathExp = pJSPKey <|> pJSPFilter <|> pJSPIdx
+pJSPathExp = P.try pJSPKey <|> P.try pJSPFilter <|> P.try pJSPIdx <|> P.try pJSPSlice
```
steve-chavez: Hm, shouldn't the last `try` be unnecessary?

src/PostgREST/Config/JSPath.hs
taimoorzaeem: I removed `eof` to allow the slice after it. It wouldn't have much impact on parse error messages.

src/PostgREST/Config/JSPath.hs
taimoorzaeem: Yes, that is redundant. Removed.

docs/references/auth.rst
```diff
@@ -234,6 +234,13 @@ The DSL follows the `JSONPath <https://goessner.net/articles/JsonPath/>`_ expres
 - ``==^`` selects the first array element that ends with the right operand
 - ``*==`` selects the first array element that contains the right operand
 
+The selected role value can also be sliced using the slice operator ``[a:b]``. Negative index values are also supported. The syntax is as:
+
+- ``[a:b]`` take slice from index ``a`` upto ``b``
```
steve-chavez: Looks like this is a misspelling:

```suggestion
- ``[a:b]`` take slice from index ``a`` up to ``b``
```

docs/postgrest.dict
```diff
@@ -182,6 +182,7 @@ Untyped
 UPSERT
 Upsert
 upsert
+upto
```
steve-chavez:
```suggestion
```

src/PostgREST/Config/JSPath.hs
```diff
@@ -59,12 +61,25 @@ walkJSPath :: Maybe JSON.Value -> JSPath -> Maybe JSON.Value
...
+walkJSPath (Just (JSON.String str)) (JSPSlice start end:rest) =
+  let
+    len = T.length str
+
+    norm :: Maybe Int -> Maybe Int -- Normalize negative indices to positive
+    norm = fmap (\i -> max 0 $ min len $ if i < 0 then len + i else i)
+
+    s = fromMaybe 0 $ norm start -- normalized start index
+    e = fromMaybe len $ norm end -- normalized end index
+    slicedString = if s >= e then T.empty else T.take (e-s) $ T.drop s str
```
steve-chavez, line 73: We should better document this, so doing `[11:2]` is absurd and result in no match right?

Ideally we would report an error when that config is applied but not sure if that's hard in code right now.

taimoorzaeem: > We should better document this, so doing `[11:2]` is absurd and result in no match right?

Yes. Agree that we should document this.

> Ideally we would report an error when that config is applied but not sure if that's hard in code right now.

It would take a bit more logic considering we also support negative indices. I am thinking that our slice logic is almost exactly similar to [python slicing](https://docs.python.org/3/library/functions.html#slice), so most users might already be familiar with it, so documenting this case should be enough for most users.

PR #4609, opened 2026-01-20
add: add "Vary" header to responses

### LINKED ISSUE #133: Send "Vary" headers
for Accept and Range (and anything else that makes the response vary)

develop7:
RFC: Introduces hardcoded (for now) `Vary` header for all responses. I'm almost certain the logic should be a bit more complex, so absolute do chime in when that's the case. 

Fixes #133

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,11 @@ All notable changes to this project will be documented in this file. From versio
 - Optimize requests with `Prefer: count=exact` that do not use ranges or `db-max-rows` by @laurenceisla in #3957
   + Removed unnecessary double count when building the `Content-Range`.
 - Add config `client_error_verbosity` to customize error verbosity by @taimoorzaeem in #4088, #3980, #3824
+- Add `Vary` header to responses by @develop7 in #4609
 
 ### Changed
+
+- All responses now include a `Vary` header by @develop7 in #4609
 
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
   + Now fails at startup. Prior to this, it failed with `PGRST205` on requests related to these schemas.
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -60,15 +60,16 @@ import PostgREST.SchemaCache          (SchemaCache (..))
 import PostgREST.TimeIt               (timeItT)
 import PostgREST.Version              (docsVersion, prettyVersion)
 
-import qualified Data.ByteString.Char8  as BS
-import qualified Data.List              as L
-import           Data.Streaming.Network (bindPortTCP,
-                                         bindRandomPortTCP)
-import qualified Data.Text              as T
-import qualified Network.HTTP.Types     as HTTP
-import qualified Network.Socket         as NS
-import           PostgREST.Unix         (createAndBindDomainSocket)
-import           Protolude              hiding (Handler)
+import qualified Data.ByteString.Char8     as BS
+import qualified Data.List                 as L
+import           Data.Streaming.Network    (bindPortTCP,
+                                            bindRandomPortTCP)
+import qualified Data.Text                 as T
+import qualified Network.HTTP.Types        as HTTP
+import qualified Network.HTTP.Types.Header as HTTP (hVary)
+import qualified Network.Socket            as NS
+import           PostgREST.Unix            (createAndBindDomainSocket)
+import           Protolude                 hiding (Handler)
 
 type Handler = ExceptT Error
 
@@ -208,7 +209,17 @@ postgrestResponse appState conf@AppConfig{..} maybeSchemaCache authResult@AuthRe
 
   where
     toWaiResponse :: ServerTiming -> Response.PgrstResponse -> Wai.Response
-    toWaiResponse timing (Response.PgrstResponse st hdrs bod) = Wai.responseLBS st (hdrs ++ ([serverTimingHeader timing | configServerTimingEnabled])) bod
+    toWaiResponse timing (Response.PgrstResponse st hdrs bod) =
+      Wai.responseLBS st (hdrs ++ serverTimingHeaders timing ++ [varyHeader | not $ varyHeaderPresent hdrs]) bod
+
+    serverTimingHeaders :: ServerTiming -> [HTTP.Header]
+    serverTimingHeaders timing = [serverTimingHeader timing | configServerTimingEnabled]
+
+    varyHeader :: HTTP.Header
+    varyHeader = (HTTP.hVary, "Accept, Prefer, Range")
+
+    varyHeaderPresent :: [HTTP.Header] -> Bool
+    varyHeaderPresent = any (\(h, _v) -> h == HTTP.hVary)
 
     withTiming :: Handler IO a -> Handler IO (Maybe Double, a)
     withTiming f = if configServerTimingEnabled
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1811,3 +1811,23 @@ def test_client_error_verbosity_config(defaultenv):
             "details": None,
             "hint": "Perhaps you meant the table 'public.items'",
         }
+
+
+def test_vary_custom_header_set(defaultenv):
+    "Test default Vary header value is overridden in pre-request database function"
+
+    env = {**defaultenv, "PGRST_DB_PRE_REQUEST": "custom_vary_hdr"}
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.get("/projects")
+
+        assert response.headers["Vary"] == "X-Test-Accept"
+
+
+def test_vary_default_header_set(defaultenv):
+    "Test default Vary header value matches default one"
+
+    with run(env=defaultenv) as postgrest:
+        response = postgrest.session.get("/projects")
+
+        assert response.headers["Vary"] == "Accept, Prefer, Range"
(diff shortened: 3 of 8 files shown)


taimoorzaeem:
Linking the [RFC 7231: Vary](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4) for reference.

wolfgangwalther:
From reading up on the RFC only a tiny bit it looks to me like... we might not actually need much more sophisticated logic.

Of course, it'd be great have some kind of sophisticated way of encoding, possibly in the haskell type system or so, which headers are in the list - instead of hardcoding them (and the forgetting to update the list accordingly when changes are made). But this might well be overengineering.

We should certainly not send this header only when we actually varied the response - this defeats the purpose of using this as a cache key on intermediates.

So.. this might already be close to "it"?

develop7:
Let me update the changelog then

develop7:
BTW do I need tests?

wolfgangwalther:
There probably can't be tested much in a meaningful way here. Maybe you can add some expectations for these headers to an existing test somewhere.

Conflicted in the changelog.

src/PostgREST/App.hs
```diff
@@ -178,7 +178,14 @@ postgrestResponse appState conf@AppConfig{..} maybeSchemaCache authResult@AuthRe
 
   where
     toWaiResponse :: ServerTiming -> Response.PgrstResponse -> Wai.Response
-    toWaiResponse timing (Response.PgrstResponse st hdrs bod) = Wai.responseLBS st (hdrs ++ ([serverTimingHeader timing | configServerTimingEnabled])) bod
+    toWaiResponse timing (Response.PgrstResponse st hdrs bod) =
+      Wai.responseLBS st (hdrs ++ serverTimingHeaders timing ++ [varyHeader]) bod
```
mkleczek: I think it is worth filtering `hdrs` to make sure `Vary` is not there.

Not sure what to do in such a case. Error out? Leave them?

develop7: got it, I'd leave it as is

taimoorzaeem:
> BTW do I need tests?

It would be good to add at least one to prevent regression.

mkleczek:
> > BTW do I need tests?
> 
> It would be good to add at least one to prevent regression.

Would be good to also have another one testing that `Vary` header set in the database (eg. in pre-request or rpc function) is not overwritten.

wolfgangwalther:
> Would be good to also have another one testing that `Vary` header set in the database (eg. in pre-request or rpc function) is not overwritten.

What should the behavior be here? Should both be merged? Should a database header replace the default entirely? Not 100% sure right now, but I'm thinking that merging is the better way.

steve-chavez:
> Would be good to also have another one testing that Vary header set in the database (eg. in pre-request or rpc function) is not overwritten.

Is it possible that some users were already setting the Vary header on pre-request? I believe so, in that case this feature would also be a breaking change.

> Not 100% sure right now, but I'm thinking that merging is the better way.

I also think merging is correct because users might read their own `My-Special-Header` and then modify the responses according to that. It would be up to them to modify `Vary` and set `My-Special-Header` while it's our responsibility to set the standard headers we use Prefer, Accept, etc.

src/PostgREST/App.hs
develop7: Oh, it's done BTW

develop7:
@steve-chavez I've updated the PR so it does not change existing `Vary` header. IMO clients can be hacky of different degree — case-sensitive, order-sensitive, whole-content-sensitive even etc, so compatibility-wise I think it'd be better to not change existing header altogether.

src/PostgREST/App.hs
```diff
@@ -200,7 +201,17 @@ postgrestResponse appState conf@AppConfig{..} maybeSchemaCache authResult@AuthRe
...
+      Wai.responseLBS st (hdrs ++ serverTimingHeaders timing ++ [varyHeader | not $ varyHeaderPresent hdrs]) bod
+
+    serverTimingHeaders :: ServerTiming -> [HTTP.Header]
+    serverTimingHeaders timing = [serverTimingHeader timing | configServerTimingEnabled]
+
+    varyHeader :: HTTP.Header
+    varyHeader = (HTTP.hVary, "Accept-Encoding, Accept, Prefer, Range")
+
+    varyHeaderPresent :: [HTTP.Header] -> Bool
+    varyHeaderPresent = any (\(h, _v) -> h == HTTP.hVary)
```
steve-chavez, line 222: I don't think this covers the case mentioned in https://github.com/PostgREST/postgrest/pull/4609#issuecomment-3837688154, if the user set a Vary header at the db level (db-pre-request or function/RPC). If it does we should have a test.

steve-chavez:
We need to add tests here:

- Vary header is present when there's no existing one. Including one coming from the db ([ref](https://docs.postgrest.org/en/v14/references/transactions.html#response-headers)).
- Vary header is not present when there's an existing one.

src/PostgREST/App.hs
develop7: I've added tests including this case, and this implementation does cover pre-requests, as I've expected it to.

develop7:
Added tests, fixed coverage, rebased against main, updated fixtures (took me a bit to figure `postgrest-test-io -- --snapshot-update`).

src/PostgREST/App.hs
```diff
@@ -201,7 +202,17 @@ postgrestResponse appState conf@AppConfig{..} maybeSchemaCache authResult@AuthRe
...
     toWaiResponse :: ServerTiming -> Response.PgrstResponse -> Wai.Response
-    toWaiResponse timing (Response.PgrstResponse st hdrs bod) = Wai.responseLBS st (hdrs ++ ([serverTimingHeader timing | configServerTimingEnabled])) bod
+    toWaiResponse timing (Response.PgrstResponse st hdrs bod) =
+      Wai.responseLBS st (hdrs ++ serverTimingHeaders timing ++ [varyHeader | not $ varyHeaderPresent hdrs]) bod
+
+    serverTimingHeaders :: ServerTiming -> [HTTP.Header]
+    serverTimingHeaders timing = [serverTimingHeader timing | configServerTimingEnabled]
+
+    varyHeader :: HTTP.Header
+    varyHeader = (HTTP.hVary, "Accept-Encoding, Accept, Prefer, Range")
```
steve-chavez: Q: Do we change the response with `Accept-Encoding`? I mean users could specify `Accept-Encoding: gzip` but we don't have that enabled.

Warp does have it builtin it seems: https://hackage-content.haskell.org/package/wai-extra-3.1.18/docs/Network-Wai-Middleware-Gzip.html

test/io/test_io.py
```diff
@@ -1789,3 +1789,19 @@ def test_client_error_verbosity_config(defaultenv):
             "details": None,
             "hint": "Perhaps you meant the table 'public.items'",
         }
+
+
+def test_vary_custom_header_set(defaultenv):
+    env = {**defaultenv, "PGRST_DB_PRE_REQUEST": "custom_vary_hdr"}
```
steve-chavez, line 1819: Could you add a comment here? This is the case were the user overrides `Vary` right?

So far every test in `test/io/test_io.py` has had a comment so we should maintain consistency.

test/io/test_io.py
```diff
@@ -1789,3 +1789,19 @@ def test_client_error_verbosity_config(defaultenv):
...
+
+    with run(env=env) as postgrest:
+        response = postgrest.session.get("/projects")
+
+        assert response.headers["Vary"] == "Accept-Encoding"
+
+
+def test_vary_default_header_set(defaultenv):
+    with run(env=defaultenv) as postgrest:
```
steve-chavez, line 1830: Ditto, a comment.

src/PostgREST/App.hs
mkleczek: @steve-chavez 
I wouldn’t do compression in Haskell, comparing to Nginx and similar reverse proxies it is going to be painfully slow, I am afraid (similarily to SSL/TLS termination).

steve-chavez: Agree, that should be delegated to proxy.

@develop7 So looks like we should remove `Accept-Encoding` here?

develop7: Yep, done

test/io/test_io.py
develop7: Gotchu, done

test/io/test_io.py
develop7: Done, thanks

steve-chavez:
We need docs for this. I think it should be a new page under https://docs.postgrest.org/en/v14/references/api.html

develop7:
@steve-chavez I've added a page under the `api/` section, but I'm not sure whether it's a good place for that

steve-chavez:
@develop7 Looks good :+1: . Make sure the spellcheck passes and then we can merge.

develop7:
spellcheck done

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file. From versio
 - Optimize requests with `Prefer: count=exact` that do not use ranges or `db-max-rows` by @laurenceisla in #3957
   + Removed unnecessary double count when building the `Content-Range`.
 - Add config `client_error_verbosity` to customize error verbosity by @taimoorzaeem in #4088, #3980, #3824
+- Add `Vary` header to responses by @develop7 in #4609
 
 ### Changed
```
steve-chavez, line 18: Let's also list this in changes just to be extra cautious
```suggestion
### Changed

- All responses now include a `Vary` header by @develop7 in #4609
```

develop7:
@steve-chavez done; rebased & resolved conflicts

PR #4614, opened 2026-01-25
fix: Use bracket to release broken listener connections

mkleczek:
retryingListen function potentially leaks database connections. This patch ensures the connections are released in case of listen/notify errors.

### DIFF
--- a/src/PostgREST/Listener.hs
+++ b/src/PostgREST/Listener.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase      #-}
 {-# LANGUAGE MultiWayIf      #-}
 {-# LANGUAGE RecordWildCards #-}
 
@@ -47,8 +48,14 @@ retryingListen appState = do
 
   -- forkFinally allows to detect if the thread dies
   void . flip forkFinally handleFinally $ do
-    bracket (SQL.acquire $ toUtf8 (Config.addTargetSessionAttrs $ Config.addFallbackAppName prettyVersion configDbUri)) (`whenRight` releaseConnection) $ \dbOrError -> do
-      case dbOrError of
+    -- Make sure we don't leak connections on errors
+    bracket
+      -- acquire connection
+      (SQL.acquire $ toUtf8 (Config.addTargetSessionAttrs $ Config.addFallbackAppName prettyVersion configDbUri))
+      -- release connection
+      (`whenRight` releaseConnection) $
+      -- use connection
+      \case
         Right db -> do
           SQL.listen db $ SQL.toPgIdentifier dbChannel
           AppState.putIsListenerOn appState True
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file. From versio
 
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
 
+### Fixed
+
+- Ensure Listener connections are released by @mkleczek in #4614
+
 ## [14.3] - 2026-01-03
 
 ### Fixed


src/PostgREST/Listener.hs
```diff
@@ -49,25 +47,25 @@ retryingListen appState = do
 
   -- forkFinally allows to detect if the thread dies
   void . flip forkFinally handleFinally $ do
-    dbOrError <- SQL.acquire $ toUtf8 (Config.addTargetSessionAttrs $ Config.addFallbackAppName prettyVersion configDbUri)
-    case dbOrError of
-      Right db -> do
-        SQL.listen db $ SQL.toPgIdentifier dbChannel
-        AppState.putIsListenerOn appState True
+    bracket (SQL.acquire $ toUtf8 (Config.addTargetSessionAttrs $ Config.addFallbackAppName prettyVersion configDbUri)) (`whenRight` releaseConnection) $ \dbOrError -> do
```
steve-chavez: > retryingListen function potentially leaks database connections. 

From the `acquire` code can you see how could this happen?

```hs
acquire :: Settings.Settings -> IO (Either ConnectionError Connection)
acquire settings =
  {-# SCC "acquire" #-}
  runExceptT $ do
    pqConnection <- lift (IO.acquireConnection settings)
    lift (IO.checkConnectionStatus pqConnection) >>= traverse throwError
    lift (IO.initConnection pqConnection)
    integerDatetimes <- lift (IO.getIntegerDatetimes pqConnection)
    registry <- lift (IO.acquirePreparedStatementRegistry)
    pqConnectionRef <- lift (newMVar pqConnection)
    pure (Connection pqConnectionRef integerDatetimes registry)
```

src/PostgREST/Listener.hs
```diff
@@ -82,3 +80,5 @@ retryingListen appState = do
 
     cacheReloader =
       AppState.schemaCacheLoader appState
+
+    releaseConnection = void . forkIO . SQL.release
```
steve-chavez: Why do we need the `forkIO` here? Also if the `release` fails can we log an error?

steve-chavez: IMO we should log whenever `releaseConnection` is executed too.

src/PostgREST/Listener.hs
steve-chavez: Maybe we can add a comment saying we're being extra careful with `bracket` here. I do think there's a potential for leaked connections.

mkleczek: It's not `acquire` that leaks connections. It is downstream code calling `SQL.listen` and `SQL.waitForNotifications` that can throw but does not perform clean-up.

src/PostgREST/Listener.hs
mkleczek: > Why do we need the `forkIO` here?

So that we don't block listener thread. What makes blocking even worse is that `bracket` calls clean-up routine with exceptions masked.

> Also if the release fails can we log an error?

Good idea.

mkleczek: > IMO we should log whenever `releaseConnection` is executed too.

Hmm... we are logging all listener errors anyway. If there is a problem with connection clean-up we are logging it as well.
I am not convinced logging connection clean-up attempt would add any useful information... But happy to stand corrected.

mkleczek: > 
> > Also if the release fails can we log an error?
> 
> Good idea.

Done.

src/PostgREST/Listener.hs
mkleczek: > Maybe we can add a comment saying we're being extra careful with `bracket` here. I do think there's a potential for leaked connections.

Looking at it closer it indeed looks like `pqConnection` might leak in `SQL.acquire`. Not sure what to do as we're using quite outdated `hasql` version.

steve-chavez:
@mkleczek Don't forget to add an entry on the CHANGELOG :pray:

Should be like: "ensure broken Listener connection are released.."

PR #4615, opened 2026-01-25
fix: Unmask exceptions in listener retries

mkleczek:
forkFinally calls error handler with exceptions masked. Our handleFinally does not restore exception state before calling retryingListener so after first error the whole listener is running with exceptions masked.
This patch ensures handleFinally is run with exceptions unmasked.

### DIFF
--- a/src/PostgREST/Listener.hs
+++ b/src/PostgREST/Listener.hs
@@ -31,15 +31,16 @@ runListener :: AppState -> IO ()
 runListener appState = do
   AppConfig{..} <- getConfig appState
   when configDbChannelEnabled $
-    void . forkIO $ retryingListen appState
+    void . forkIO . void $ retryingListen appState
 
 -- | Starts a LISTEN connection and handles notifications. It recovers with exponential backoff with a cap of 32 seconds, if the LISTEN connection is lost.
-retryingListen :: AppState -> IO ()
+-- | This function never returns (but can throw) and return type enforces that.
+retryingListen :: AppState -> IO Void
 retryingListen appState = do
   AppConfig{..} <- AppState.getConfig appState
   let
     dbChannel = toS configDbChannel
-    handleFinally err = do
+    onError err = do
       AppState.putIsListenerOn appState False
       observer $ DBListenFail dbChannel (Right err)
       when (isDbListenerBug err) $
@@ -53,10 +54,11 @@ retryingListen appState = do
       threadDelay (delay * oneSecondInMicro)
       unless (delay == maxDelay) $
         AppState.putNextListenerDelay appState (delay * 2)
+      -- loop running the listener
       retryingListen appState
 
-  -- forkFinally allows to detect if the thread dies
-  void . flip forkFinally handleFinally $ do
+  -- Execute the listener with with error handling
+  handle onError $ do
     -- Make sure we don't leak connections on errors
     bracket
       -- acquire connection
@@ -81,7 +83,9 @@ retryingListen appState = do
 
           observer $ DBListenStart pqHost pqPort pgFullName dbChannel
 
-          SQL.waitForNotifications handleNotification db
+          -- wait for notifications
+          -- this will never return, in case of an error it will throw and be caught by onError
+          forever $ SQL.waitForNotifications handleNotification db
 
         Left err -> do
           observer $ DBListenFail dbChannel (Left err)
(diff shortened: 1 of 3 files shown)


src/PostgREST/Listener.hs
```diff
@@ -44,9 +44,11 @@ retryingListen appState = do
       unless (delay == maxDelay) $
         AppState.putNextListenerDelay appState (delay * 2)
       retryingListen appState
+    
+    withHandler action = mask $ \restore -> forkIO $ try (restore action) >>= restore . handleFinally
```
steve-chavez: > This patch ensures handleFinally is run with exceptions unmasked

So the benefit is to run `handleFinally` when  asynchronous exceptions happen right? And this is currently not happening because they're "masked"?

mkleczek: > > This patch ensures handleFinally is run with exceptions unmasked
> 
> So the benefit is to run `handleFinally` when asynchronous exceptions happen right? And this is currently not happening because they're "masked"?

Yes. What's important is that `handleFinally` calls `retryingListen` recursively, so after first failure the whole listener is going to run with async exceptions masked.

steve-chavez: Can you add a comment here explaining that in detail? 🙏 

> Yes. What's important is that handleFinally calls retryingListen recursively, so after first failure the whole listener is going to run with async exceptions masked.

Also, do you have any idea how could we reproduce this? I might be able to help with a test.

mkleczek: > Can you add a comment here explaining that in detail? 🙏
> 

Will do.

Having thought about it though, I think forking a new thread upon each failure is not needed here at all. I think we can just loop in the same listener thread.

> > Yes. What's important is that handleFinally calls retryingListen recursively, so after first failure the whole listener is going to run with async exceptions masked.
> 
> Also, do you have any idea how could we reproduce this? I might be able to help with a test.

It might be somewhat complex to test (but not impossible as there is https://hackage-content.haskell.org/package/base-4.22.0.0/docs/GHC-Conc.html#v:listThreads ).
I am not sure if it is worth testing TBH - listener thread is "private" (ie. its `ThreadId` is not exposed) and we don't `throwTo` it anywhere.
This PR is really low priority and just "cleaning up" - more of aesthetics.

mkleczek: > > Can you add a comment here explaining that in detail? 🙏
> 
> Will do.
> 

Done.

> Having thought about it though, I think forking a new thread upon each failure is not needed here at all. I think we can just loop in the same listener thread.

Changed that as well. Please, take a look.

src/PostgREST/Listener.hs
```diff
@@ -53,10 +53,16 @@ retryingListen appState = do
...
 
-  -- forkFinally allows to detect if the thread dies
-  void . flip forkFinally handleFinally $ do
+  -- Execute the listener with with error handling
+  -- We use mask to make sure onError is invoked in presence of asynchronous exceptions
+  -- but we execute the onError handler in an unmasked state to
+  -- 1. allow it to be interrupted
+  -- 2. make sure recursive call to retryingListen is not masked
+  -- TODO is masking really necessary here? We unmask the onError handler anyway...
+  void $ mask $ \unmask -> handle (unmask . onError) $ unmask $ do
```
steve-chavez: Right, I was thinking the same.. can we avoid the mask/unmask?

FWIW, `forkFinally` code is simple:

```hs
forkFinally :: IO a -> (Either SomeException a -> IO ()) -> IO ThreadId
forkFinally action and_then =
  mask $ \restore ->
    forkIO $ try (restore action) >>= and_then
```

mkleczek: > Right, I was thinking the same.. can we avoid the mask/unmask?
> 

Fixed.

I've also changed return type of `retryingListen` to `IO Void` to make sure no accidental loop exit happens.

Updated changeling as well and I think it is now in a committable shape.

steve-chavez: Nice! `forever` also makes the code look much clearer :100:

taimoorzaeem:
Looks like the backport workflow failed due to conflicts caused by code changes in #4617 and #4618.

steve-chavez:
Fixed the conflict and pushed in https://github.com/PostgREST/postgrest/commit/16c767134c64278128ef9aec0c172d89e36fd6e5

PR #4617, opened 2026-01-27
add: Log actual host and port of listener connection

mkleczek:
Diagnosing problems with listener channel notifications not being handled properly by PostgREST connected to read replicas is difficult. Issues might be related to lost connections and listener not being connected to the right host after failover or database server restarts.

This patch adds logging of actual host:port used by libpq connection opened by the listener. It should make it easier to find out if PostgREST is connected to the right host.

### DIFF
--- a/src/PostgREST/Observation.hs
+++ b/src/PostgREST/Observation.hs
@@ -44,7 +44,7 @@ data Observation
   | SchemaCacheSummaryObs Text
   | SchemaCacheLoadedObs Double
   | ConnectionRetryObs Int
-  | DBListenStart Text
+  | DBListenStart (Maybe ByteString) (Maybe ByteString) Text -- host, port, channel
   | DBListenFail Text (Either SQL.ConnectionError (Either SomeException ()))
   | DBListenRetry Int
   | DBListenBugHint -- https://github.com/PostgREST/postgrest/issues/3147
@@ -110,8 +110,8 @@ observationMessage = \case
     "Attempting to reconnect to the database in " <> (show delay::Text) <> " seconds..."
   QueryPgVersionError usageErr ->
     "Failed to query the PostgreSQL version. " <> jsonMessage usageErr
-  DBListenStart channel -> do
-    "Listening for database notifications on the " <> show channel <> " channel"
+  DBListenStart host port channel -> do
+    "Listener connected to " <> show (fold $ host <> fmap (":" <>) port) <> " and listening for database notifications on the " <> show channel <> " channel"
   DBListenFail channel listenErr ->
     "Failed listening for database notifications on the " <> show channel <> " channel. " <>
       either showListenerConnError showListenerException listenErr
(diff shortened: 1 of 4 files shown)


src/PostgREST/Observation.hs
```diff
@@ -110,8 +110,8 @@ observationMessage = \case
     "Attempting to reconnect to the database in " <> (show delay::Text) <> " seconds..."
   QueryPgVersionError usageErr ->
     "Failed to query the PostgreSQL version. " <> jsonMessage usageErr
-  DBListenStart channel -> do
-    "Listening for database notifications on the " <> show channel <> " channel"
+  DBListenStart host port channel -> do
+    "Listener connected to " <> show (fold $ host <> fmap (":" <>) port) <> " and listening for database notifications on the " <> show channel <> " channel"
```
steve-chavez, line 114: This is nice. So currently we have:

```
27/Jan/2026:09:22:12 -0500: Listening for database notifications on the "pgrst" channel
27/Jan/2026:09:22:12 -0500: Successfully connected to PostgreSQL 14.18 on x86_64-pc-linux-gnu, compiled by clang version 19.1.7, 64-bit
```

I'm wondering if we should make both logs mostly the same. So the Listener also logs the db version and the connection pool also logs the host and port.

If too complicated right now maybe we can add a TODO in the code so we don't forget?

mkleczek: Logging actual host:port does not really make sense for the pool because each connection might be to a different host.

It might make sense to debug log host:port for each new connection created by the pool but I would say it is out of scope of this PR.

What we might add here is Pg version but I am a little bit concerned with the length of the log entry.

WDYT?

steve-chavez: > It might make sense to debug log host:port for each new connection created by the pool but I would say it is out of scope of this PR.

Yes, makes sense. 

> What we might add here is Pg version but I am a little bit concerned with the length of the log entry.

I don't think the log length should be something to worry about. Although getting the full pg version involves doing an additional query:

https://github.com/PostgREST/postgrest/blob/ffa393850a461ed02a0906833a52efe5fa844d43/src/PostgREST/AppState.hs#L377

https://github.com/PostgREST/postgrest/blob/ffa393850a461ed02a0906833a52efe5fa844d43/src/PostgREST/Config/Database.hs#L73-L74

Which might complicate the code a bit.

steve-chavez: Let's merge as it is.

mkleczek: > I don't think the log length should be something to worry about. Although getting the full pg version involves doing an additional query:
> 
> https://github.com/PostgREST/postgrest/blob/ffa393850a461ed02a0906833a52efe5fa844d43/src/PostgREST/AppState.hs#L377
> 
> https://github.com/PostgREST/postgrest/blob/ffa393850a461ed02a0906833a52efe5fa844d43/src/PostgREST/Config/Database.hs#L73-L74
> 
> Which might complicate the code a bit.

Done

PR #4618, opened 2026-01-27
add: Log pg version details of listener connection

mkleczek:
Follow-up to #4617 adding more information to log entry produced upon successful listener connection establishement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file. From versio
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
 - Add a `HINT` when the LISTEN channel stops working due to a PostgreSQL bug by @laurenceisla in #4581
 - Add string slicing operator for `jwt-role-claim-key` by @taimoorzaeem in #4599
-- Log the actual host and port of listener database connection by @mkleczek in #4617
+- Log host, port and pg version of listener database connection by @mkleczek in #4617 #4618
 
 ### Fixed
 
--- a/src/PostgREST/Listener.hs
+++ b/src/PostgREST/Listener.hs
@@ -17,10 +17,13 @@ import           PostgREST.Version     (prettyVersion)
 import qualified PostgREST.AppState as AppState
 import qualified PostgREST.Config   as Config
 
-import           Control.Arrow             ((&&&))
-import           Data.Bitraversable        (bisequence)
-import           Data.Either.Combinators   (whenRight)
-import qualified Database.PostgreSQL.LibPQ as LibPQ
+import           Control.Arrow              ((&&&))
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file. From versio
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
 - Add a `HINT` when the LISTEN channel stops working due to a PostgreSQL bug by @laurenceisla in #4581
 - Add string slicing operator for `jwt-role-claim-key` by @taimoorzaeem in #4599
-- Log the actual host and port of listener database connection by @mkleczek in #4617
+- Log details of listener database connection by @mkleczek in #4617 #4618
```
steve-chavez: More explicit would be better::
```suggestion
- Log host, port and pg version of listener database connection by @mkleczek in #4617 #4618
```

src/PostgREST/Listener.hs
```diff
@@ -74,7 +77,8 @@ retryingListen appState = do
             AppState.putNextListenerDelay appState 1
 
           (pqHost, pqPort) <- SQL.withLibPQConnection db $ bisequence . (LibPQ.host &&& LibPQ.port)
-          observer $ DBListenStart pqHost pqPort dbChannel
+          pgFullName <- SQL.run (queryPgVersion False) db >>= either throwIO (pure . pgvFullName)
```
steve-chavez: Maybe we should log here if the query fails, like done on:

https://github.com/PostgREST/postgrest/blob/ffa393850a461ed02a0906833a52efe5fa844d43/src/PostgREST/AppState.hs#L377-L380

mkleczek: It is going to be logged by `handleFinally`:
```
observer $ DBListenFail dbChannel (Right err)
```

`pgFullName <- SQL.run (queryPgVersion False) db >>= either throwIO (pure . pgvFullName)` will throw `QueryError` as exception. I've decided to do it that way so that it is handled uniformly with failures of `hasql-notification` functions.
TBH mixing exceptions and `MonadError` or `Either` failures handling is a real PITA. Especially when clean-up of resources is required.

steve-chavez: > observer $ DBListenFail dbChannel (Right err)

Right, I missed that. LGTM then.

steve-chavez:
Note: 

The old log line looked like this:
```
27/Jan/2026:09:22:12 -0500: Listening for database notifications on the "pgrst" channel
```

new log line looks like this:
```
27/Jan/2026:15:39:57 -0500: Listener connected to PostgreSQL 14.18 on x86_64-pc-linux-gnu, compiled by clang version 19.1.7, 64-bit on "/tmp/nix-shell-62410-0/postgrest-with-pg-14-VzG/socket:5432" and listening for database notifications on the "pgrst" channel
```

Issue #2323, opened 2022-06-14
PATCH/DELETE based on an embedded table condition


steve-chavez:
With [Embedding with Top-level Filtering](https://postgrest.org/en/stable/api.html#embedding-with-top-level-filtering) is now reasonable to do a PATCH/DELETE based on an embedded table condition, since it's an exact match.

This is not possible currently:

```http
DELETE /projects?select=id,name,clients!inner(*)&clients.name=eq.Apple
Prefer:count=exact

HTTP/1.1 204 No Content
Content-Range: */0
```

We restrict update/delete filters just for the root table [here](https://github.com/PostgREST/postgrest/blob/36ed7e2fb76c587e91a4463200c8ad5993bb3251/src/PostgREST/Request/DbRequestBuilder.hs#L339-L340), but this restriction can be lifted in cases where `!inner` is used.

---

Using the `select` is a bit confusing here, we'd need the `exists` filter for it to be clearer.

steve-chavez:
This is possible with the following query:

```sql
WITH pgrst_source AS (
  WITH
  pgrst_payload AS (SELECT '{"name": "Newnew"}'::json AS json_data),
  pgrst_body AS ( SELECT CASE WHEN json_typeof(json_data) = 'array' THEN json_data ELSE json_build_array(json_data) END AS val FROM pgrst_payload),
  pgrst_update_body AS (SELECT * FROM json_populate_recordset (null::"test"."projects" , (SELECT val FROM pgrst_body) ) LIMIT 1),
  pgrst_affected_rows AS (
    SELECT
      "projects"."id",
      "projects"."name",
      row_to_json("projects_clients_1".*) AS "clients"
    FROM "test"."projects" AS "projects"
    LEFT JOIN LATERAL (
      SELECT "clients_1"."name"
      FROM "test"."clients" AS "clients_1"
      WHERE
        "clients_1"."id" = "projects"."client_id" AND
        "clients_1"."name" = 'Apple'
    ) AS "projects_clients_1" ON TRUE
    WHERE "projects_clients_1" IS NOT NULL
  )
  UPDATE "test"."projects"
  SET "name" = (SELECT "name" FROM pgrst_update_body)
  FROM pgrst_affected_rows
  WHERE "test"."projects"."id" = "pgrst_affected_rows"."id"
  RETURNING "test"."projects"."client_id", "test"."projects"."id", "test"."projects"."name", "pgrst_affected_rows"."clients"
)
SELECT
[...]

PR #4619, opened 2026-01-27
fix: filtering the returned representation for mutations when using or/and filters

### LINKED ISSUE #3707: Invalid query generated when selecting after updating using or operator
### Environment

* PostgreSQL version: public.ecr.aws/supabase/postgres:15.6.1.113
* PostgREST version: public.ecr.aws/supabase/postgrest:v12.2.3

### Description of issue

I faced a problem when trying to update a timestamp column and return the updated value.

Here is the table structure:
```sql
create table jobs (
    id uuid primary key default gen_random_uuid(),
    started_at timestamp with time zone
)
```
I want to update and return a job if it has not been started yet or if it was started more then one minute ago.

But when I make the following query using supabase:
```ts
const min_start_at = new Date(Date.now() - 60000).toISOString()
supabase.from("jobs").update({ started_at: "now()" })
  .eq("id", id).or(`started_at.is.null,started_at.lt.${min_start_at}`)
  .select("id,started_at").maybeSingle()
```
PostgREST generates a query that doesn't returns the updated row:
```sql
WITH pgrst_source AS (
    UPDATE "public"."jobs"
    SET "started_at" = "pgrst_body"."started_at"
    FROM
        (SELECT $1 AS json_data) pgrst_payload,
        LATERAL (SELECT "started_at" FROM json_to_record(pgrst_payload.json_data) AS _("started_at" timestamp with time zone) ) pgrst_body
    WHERE  "public"."jobs"."id" = $2
        AND  ( "public"."jobs"."started_at" IS NULL OR  "public"."jobs"."started_at" < $3)
    RETURNING "public"."jobs"."id", "public"."jobs"."started_at"
)
SELECT
    '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total,
    array[]::text[] AS header, coalesce(json_agg(_postgrest_t)->0, 'null') AS body,
    '' AS response_inserted
FROM (
    SELECT "jobs"."started_at"
    FROM "pgrst_source" AS "jobs"
    WHERE  ( "jobs"."started_at" IS NULL OR  "jobs"."started_at" < $4)
);
```
The problem is the last WHERE clause because at this stage `started_at` has already been set to `now()` which can't be less then a moment in the past.

Furthermore, if you try to remove `started_at` from select:
```ts
const min_start_at = new Date(Date.now() - 60000).toISOString()
supabase.from("jobs").update({ started_at: "now()" })
  .eq("id", id).or(`started_at.is.null,started_at.lt.${min_start_at}`)
  .select("id").maybeSingle()
```
you'll recieve an error because the following query is invalid:
```sql
WITH pgrst_source AS (
    UPDATE "public"."jobs"
    SET "started_at" = "pgrst_body"."started_at"
    FROM (
        SELECT $1 AS json_data) pgrst_payload,
        LATERAL (SELECT "started_at" FROM json_to_record(pgrst_payload.json_data) AS _("started_at" timestamp with time zone) ) pgrst_body
    WHERE  "public"."jobs"."id" = $2
        AND  ( "public"."jobs"."started_at" IS NULL OR  "public"."jobs"."started_at" < $3)
    RETURNING "public"."jobs"."id"
)
SELECT
    '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total,
    array[]::text[] AS header, coalesce(json_agg(_postgrest_t)->0, 'null') AS body,
    '' AS response_inserted
FROM (
    SELECT "jobs"."id"
    FROM "pgrst_source" AS "jobs"
    WHERE  ( "jobs"."started_at" IS NULL OR  "jobs"."started_at" < $4)
);
```
Notice the `pgrst_source` here, it contains only `id` but the WHERE clause still references `started_at`.

Looks like the problem is related to the `or` operator because when I use `is.null`:
```ts
supabase.from("jobs").update({ started_at: "now()" })
  .eq("id", id).is("started_at", null)
  .select("id,started_at").maybeSingle()
```
or `lt.` separately:
```ts
const min_start_at = new Date(Date.now() - 60000).toISOString()
supabase.from("jobs").update({ started_at: "now()" })
  .eq("id", id).lt("started_at", min_start_at)
  .select("id,started_at").maybeSingle()
```
there is no WHERE clause in the generated by PostgREST query:
```sql
WITH pgrst_source AS (
    UPDATE "public"."jobs"
    SET "started_at" = "pgrst_body"."started_at"
    FROM
        (SELECT $1 AS json_data) pgrst_payload,
        LATERAL (SELECT "started_at" FROM json_to_record(pgrst_payload.json_data) AS _("started_at" timestamp with time zone) ) pgrst_body
    WHERE  "public"."jobs"."id" = $2
        AND  "public"."jobs"."started_at" < $3
    RETURNING "public"."jobs"."id", "public"."jobs"."started_at"
)
SELECT
    '' AS total_result_set, pg_catalog.count(_postgrest_t) AS page_total,
    array[]::text[] AS header, coalesce(json_agg(_postgrest_t)->0, 'null') AS body,
    '' AS response_inserted
FROM (
    SELECT "jobs"."started_at"
    FROM "pgrst_source" AS "jobs"
);
```
This is expected behavior and `or` operator should not add WHERE clause as I understand.

laurenceisla:
Closes #3707

### DIFF
--- a/src/PostgREST/Plan.hs
+++ b/src/PostgREST/Plan.hs
@@ -963,9 +963,16 @@ addRanges ApiRequest{..} rReq =
 
 addLogicTrees :: ResolverContext -> ApiRequest -> ReadPlanTree -> Either Error ReadPlanTree
 addLogicTrees ctx ApiRequest{..} rReq =
-  foldr addLogicTreeToNode (Right rReq) qsLogic
+  foldr addLogicTreeToNode (Right rReq) logic
   where
     QueryParams.QueryParams{..} = iQueryParams
+
+    logic =
+      case iAction of
+        ActDb (ActRelationRead _  _) -> qsLogic
+        ActDb (ActRoutine _ _)       -> qsLogic
+        -- For mutations, take the non-root logic filters. These will only affect the embeddings and not the top level of the returned representation.
+        _                            -> filter (not . null . fst) qsLogic
 
     addLogicTreeToNode :: (EmbedPath, LogicTree) -> Either Error ReadPlanTree -> Either Error ReadPlanTree
     addLogicTreeToNode = updateNode (\t (Node q@ReadPlan{from=fromTable, where_=lf} f) -> Node q{ReadPlan.where_=resolveLogicTree ctx{qi=fromTable} t:lf} f)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. From versio
 ### Fixed
 
 - Ensure Listener connections are released by @mkleczek in #4614
+- Fix incorrectly filtering the returned representation for PATCH requests when using `or/and` filters by @laurenceisla in #3707
 
 ## [14.3] - 2026-01-03
 
--- a/test/spec/Feature/Query/AndOrParamsSpec.hs
+++ b/test/spec/Feature/Query/AndOrParamsSpec.hs
@@ -252,21 +252,39 @@ spec =
             [json|[{"id": 7, "entities":null}, {"id": 8, "entities": {"id": 2}}, {"id": 9, "entities": {"id": 3}}]|]
             { matchStatus = 201 }
 
-    context "used with PATCH" $
+    context "used with PATCH" $ do
       it "succeeds when using and/or params" $
         request methodPatch "/grandchild_entities?or=(id.eq.1,id.eq.2)&select=id,name"
           [("Prefer", "return=representation")]
           [json|{ name : "updated grandchild entity"}|] `shouldRespondWith`
           [json|[{ "id": 1, "name" : "updated grandchild entity"},{ "id": 2, "name" : "updated grandchild entity"}]|]
           { matchHeaders = [matchContentTypeJson] }
+      it "succeeds when the filtered column is modified" $
+        request methodPatch "/entities?select=id,name&or=(name.is.null,name.like.*test*)"
+          [("Prefer", "return=representation")]
+          [json|{ "name" : "updated entity" }|] `shouldRespondWith`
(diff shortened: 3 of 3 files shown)


src/PostgREST/Plan.hs
```diff
@@ -963,10 +963,16 @@ addRanges ApiRequest{..} rReq =
...
-  foldr addLogicTreeToNode (Right rReq) qsLogic
+  foldr addLogicTreeToNode (Right rReq) logic
   where
     QueryParams.QueryParams{..} = iQueryParams
 
+    logic =
+      case iAction of
+        ActDb (ActRelationRead _  _) -> qsLogic
+        ActDb (ActRoutine _ _)       -> qsLogic
+        _                            -> filter (not . null . fst) qsLogic
```
steve-chavez, line 975: Could you explain this logic with some comments?

I see that the tests are good and solve the issue but I'm lost on the solution :(

laurenceisla: Added a comment there. Basically, it's the same logic applied to mutation filters: use only the non-root logic filters and apply them to the embeddings on the returned representation.

https://github.com/PostgREST/postgrest/blob/5356f4e973364c3ee39cb042bbc2a63a1c7fe4cb/src/PostgREST/Plan.hs#L824-L828

I didn't create another component like `QueryParams.qsFiltersNotRoot` because I think this won't be used anymore once #2323 is fixed.

steve-chavez: Thanks, I see it now. I wonder if there's a way to reuse this code for both filters and the logic tree.

steve-chavez: Checked the code and reusing doesn't seem trivial. So it LGTM.

Issue #3281, opened 2024-02-26
Static executables on other platforms: Cross compilation via Nix


wolfgangwalther:
We have been building a static executable for x64-linux for quite a while now. From this, the minimal docker image is created.

It would be really cool, if we could build the static executable on other platforms, too - and then create a multi-layer docker image, which provides minimal image size for different platforms.

One way to do this would be to run a nix environment on linux x64 and build the static executable for other platforms via **cross-compiling**.

This issue collects the current state of affairs.

---

I have been building GHC 9.4.8 as a cross-compiler for aarch64 successfully. However, when compiling PostgREST and its dependencies a major challenge comes up quickly: Template Haskell.

Template Haskell is a challenge, because it needs to be **run during compilation**. Because the cross-compiler compiles for the target platform, the compiled TH code can't be executed on the build platform. GHC will then fail like this:

```
Couldn't find a target code interpreter. Try with -fexternal-interpreter
```

There are a few different possible approaches how to deal with that:
1. Remove all Template Haskell from PostgREST and its dependencies. This is not easy, because a few transitive dependencies use TH and for example the whole hasql ecosystem depends on them. We also use it ourselves for the `Paths_` module, although that could probably be worked around.
2. Compile and execute TH natively and then pass the resulting splices to the cross-compiler. The idea is, that most Template Haskell applications are merely "code generation", e.g. helpers to create a lot of similar shaped haskell code, which is then compiled to the actual program. Many of those uses are actually independent of the target system - so it could be possible to compile and execute those splices on the build system and then pass the generated code to the cross-compiler. I experimented with that and I was able to cross-compile almost all our dependencies that way. I was doing it manually and this quickly got very annoying, that's why I didn't finish this, yet. This could possibly be automated, but it would never be the "clean" solution. Once any of our dependencies would use "unsafe" Template Haskell, this might result in hard to debug errors.
[...]

PR #4628, opened 2026-02-03
add: Oidc Connect Discovery

### LINKED ISSUE #1130: JWKS: load public keys from a well known endpoint
Please see https://auth0.com/docs/jwks

> Auth0 exposes a JWKS endpoint for each tenant, which is found at `https://YOUR_AUTH0_DOMAIN/.well-known/jwks.json`. This endpoint will contain the JWK used to sign all Auth0 issued JWTs for this tenant.

It would be great of PostgREST could load the public key from such an endpoint.

### LINKED ISSUE #4137: Possibility to retrieve JWKS using OpenID Connect Discovery
Most OpenID 2.0 IDPs expose `https:[host]/.well-known/openid-configuration` endpoint that provides `jwks_uri` pointing to JWT verification keys.

It would be convenient if PostgREST could use it.

The endpoint is specified in https://openid.net/specs/openid-connect-discovery-1_0.html

mkleczek:
**DISCLAIMER:**

Fixes #1130
Fixes #4137

Currently configuring JWT keys is cumbersome and in practice requires implementing Oidc Connect Discovery in a shell script that writes retrieved keys to a file and reloads PostgREST configuration.
Having this implemented in PostgREST would simplify deployment.

OIDC Connect discovery is widely adopted and implemented by practically all OAuth providers (Supabase as well: https://supabase.com/docs/guides/auth/oauth-server#supported-standards)

Next step is to periodically reload keys to facilitate key rotation.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -164,6 +164,8 @@ library
                     , focus                     >= 1.0 && < 2
                     , some                      >= 1.0.4.1 && < 2
                     , uuid                      >= 1.3 && < 2
+                    , oidc-client               >= 0.8.0.0 && < 0.9
+                    , http-client-tls           >= 0.3.6.4 && < 0.4
                       -- -fno-spec-constr may help keep compile time memory use in check,
                       --   see https://gitlab.haskell.org/ghc/ghc/issues/16017#note_219304
                       -- -optP-Wno-nonportable-include-path
--- a/src/PostgREST/Config.hs
+++ b/src/PostgREST/Config.hs
@@ -73,6 +73,9 @@ import PostgREST.SchemaCache.Identifiers (QualifiedIdentifier (..),
 
 import PostgREST.Version (prettyVersion)
 import Protolude         hiding (Proxy, toList)
+import Web.OIDC.Client (Provider(..), discover)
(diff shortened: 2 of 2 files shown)


steve-chavez:
:eyes: What's the motivation for this?

mkleczek:
> 👀 What's the motivation for this?

Updated description.

postgrest.cabal
```diff
@@ -158,6 +158,8 @@ library
                     , stm-hamt                  >= 1.2 && < 2
                     , focus                     >= 1.0 && < 2
                     , some                      >= 1.0.4.1 && < 2
+                    , oidc-client               >= 0.8.0.0 && < 0.9
+                    , http-client-tls           >= 0.3.6.4 && < 0.4
```
wolfgangwalther, line 168: `http-client-tls` is fine, but `oidc-client` depends on TemplateHaskell, so would block us for https://github.com/PostgREST/postgrest/issues/3281#issuecomment-4283355556. We should have *very* good reasons to introduce that without any sane alternatives, because it would effectively prevent us from ever doing cross compilation to platforms without emulator support.

Issue #4622, opened 2026-01-29
Race condition in pool_available metric causes negative values during network instability


DavraYoung:
## Problem Description

The `pgrst_db_pool_available` metric can drift into negative values during network instability or connection pool disruptions. We observed values as low as **-233** with `pgrst_db_pool_max: 80`, which is clearly incorrect.

## Root Cause

The race condition is in `src/PostgREST/Metrics.hs:35-41`:

```haskell
(HasqlPoolObs (SQL.ConnectionObservation _ status)) -> case status of
  SQL.ReadyForUseConnectionStatus  -> do
    incGauge poolAvailable
  SQL.InUseConnectionStatus        -> do
    decGauge poolAvailable
  SQL.TerminatedConnectionStatus  _ -> do
    decGauge poolAvailable
  SQL.ConnectingConnectionStatus -> pure ()
```

The `incGauge` and `decGauge` operations from `prometheus-client` are **not atomic**. During network instability:

1. Multiple connections transition states simultaneously
2. `decGauge` operations can occur before corresponding `incGauge` operations
3. Connections that terminate before becoming "ready" decrement without ever incrementing
4. The gauge drifts negative over time

## Impact

- **Metrics are unreliable** - monitoring/alerting based on `pool_available` produces false negatives
- **No actual pool impact** - the underlying pool works correctly; only the metric is corrupted
- **Persists until restart** - the counter never self-corrects; requires PostgREST restart

## Environment

- PostgREST version: v12.2.12
- `prometheus-client` constraint: `>= 1.1.1 && < 1.2.0`
- `hasql-pool` constraint: `>= 1.0.1 && < 1.1`
- Trigger: Network instability during Google GCE incident

## Suggested Fix

The gauge updates need to be atomic. Options include:

1. **Use STM** - Wrap the gauge in an `TVar` for atomic updates
2. **Use atomic operations** - If `prometheus-client` supports atomic inc/dec
3. **Track absolute count** - Calculate available from (total - in_use) instead of incrementing
4. **Add mutex/lock** - Protect gauge updates with a lock (less ideal for performance)

## Workaround

Restart PostgREST to reset the counter to correct values.

steve-chavez:
Another thing that likely is wrong is that we're decrementing the gauge for all `SQL.TerminatedConnectionStatus` states:

```hs
-- | Status of a connection.
data ConnectionStatus
  = -- | Connection is being established.
    --
    -- This is the initial status of every connection.
[...]

PR #4630, opened 2026-02-04
fix(metrics): prevent pgrst_db_pool_available drifting below zero

nothankyouzzz:
## Summary

`pgrst_db_pool_available` could drift below zero under connection churn or unstable networks. This happens because the metric was updated using pure inc/dec deltas without tracking per-connection state, so events like Connecting→Terminated or Ready→InUse→Terminated can double-decrement or decrement without a prior increment.

## Fix

Track availability per connection id and set the gauge from the absolute state:
- Maintain a small state map (conn id -> available flag) and a ready counter.
- Update the counter on state transitions and set the gauge to the computed value.
- Clamp the gauge to 0 and pool max to avoid out-of-bounds values.
- Add unit tests for lifecycle transitions and bounds.

## Issue

- https://github.com/PostgREST/postgrest/issues/4622  
  + The issue discussion includes a detailed explanation of the drift root cause and the intended fix.

## Testing

- `postgrest-lint`
- `postgrest-style`
- `postgrest-check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
+- Fix `pgrst_db_pool_available` drifting below zero under connection churn by tracking availability per connection id and setting the gauge from absolute state by @nothankyouzzz in #4622.
+
 ### Changed
 
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359


develop7:
@nothankyouzzz no merge commits please, rebase your branch against `main` instead

nothankyouzzz:
@develop7 done.

src/PostgREST/Metrics.hs
```diff
@@ -6,11 +6,18 @@ Description : Metrics based on the Observation module. See Observation.hs.
...
+  , stepPoolAvailable
+  , poolAvailableGaugeValue
   , observationMetrics
   , metricsToText
   ) where
 
 import qualified Data.ByteString.Lazy   as LBS
+import           Data.IORef             (IORef, atomicModifyIORef',
+                                         newIORef)
+import qualified Data.Map.Strict        as M
```
mkleczek, line 20: Is there a reason why not `Data.HashMap.Strict`?

src/PostgREST/Metrics.hs
```diff
@@ -52,14 +97,8 @@ observationMetrics :: MetricsState -> ObservationHandler
...
-  (HasqlPoolObs (SQL.ConnectionObservation _ status)) -> case status of
-     SQL.ReadyForUseConnectionStatus  -> do
-      incGauge poolAvailable
-     SQL.InUseConnectionStatus        -> do
-      decGauge poolAvailable
-     SQL.TerminatedConnectionStatus  _ -> do
-      decGauge poolAvailable
-     SQL.ConnectingConnectionStatus -> pure ()
+  (HasqlPoolObs (SQL.ConnectionObservation uuid status)) -> do
+    updatePoolAvailable poolAvailableState poolAvailable (show uuid) status
```
mkleczek, line 101: Looking at the code I am thinking it should be probably fixed in `hasql-pool`. The issue is that it is publishing `TerminatedConnectionStatus` for connections that never got to `ReadyForUseConnectionStatus`.

I am not sure if adding tracking logic here as a workaround for an apparent bug in `hasql-pool` is a good idea.

steve-chavez:

Issue #4633, opened 2026-02-06
IO test harness should ignore HTTP(S)_PROXY for local IPv6 readiness checks


nothankyouzzz:
### Environment

* PostgreSQL version: 17.x (nix test env)
* PostgREST version: current main/devel
* Operating system: Linux

### Description of issue

Expected behavior:
- Local readiness probes used by IO/CLI tests should not use `HTTP_PROXY` / `HTTPS_PROXY`.
- Timeout path in `wait_until_status_code` should consistently raise `PostgrestTimedOut`.

Actual behavior:
- With proxy env vars set, local IPv6 readiness probes can be routed through proxy and fail/flap.
- In no-response timeout paths, helper logic can hit an uninitialized `response` path.

A minimal SQL definition:

```sql
-- N/A for reproduction (this issue is in test harness networking/proxy handling).
```

How you make the request to PostgREST (curl command preferred):

```bash
export HTTP_PROXY=http://127.0.0.1:7890
export HTTPS_PROXY=http://127.0.0.1:7890

# local readiness probe (IPv6)
curl -i http://[::1]:3000/ready
```

The PostgREST response:

```
Intermittent readiness timeout/failure in proxied environments for local IPv6 checks.
```

Additional reproduction from test suite:

```bash
postgrest-test-io 'test/io/test_cli.py::test_cli_ready_flag_success[IPv6]'
postgrest-test-io 'test/io/test_io.py::test_admin_works_with_host_special_values[*6]'
postgrest-test-io 'test/io/test_io.py::test_admin_works_with_host_special_values[!6]'
postgrest-test-io 'test/io/test_io.py::test_log_postgrest_host_and_port[IPv6]'
```

taimoorzaeem:
> ```sh
> export HTTP_PROXY=http://127.0.0.1:7890
> export HTTPS_PROXY=http://127.0.0.1:7890
> ```

Anyone running `io-tests` or `loadtests` would have test failures when these environment variables are set like this. From the error, it's not apparent on why tests are failing so it could waste time unnecessarily.

So instead of ignoring these, I think we should print a warning when running `io-tests` only when these env variables are set, so the user would know what's wrong.

PR #4634, opened 2026-02-06
fix(test): harden io readiness polling for local IPv6

nothankyouzzz:
Initialize response in `wait_until_status_code` to avoid `UnboundLocalError` when no response is received.

Disable proxy env usage for test harness sessions so local admin/API probes are not routed through HTTP(S)_PROXY, which broke IPv6 readiness checks in proxied environments.

Validated with targeted tests:

```text
test/io/test_cli.py::test_cli_ready_flag_success[IPv6]
test/io/test_io.py::test_admin_works_with_host_special_values[*6]
test/io/test_io.py::test_admin_works_with_host_special_values[!6]
test/io/test_io.py::test_log_postgrest_host_and_port[IPv6]
```

### DIFF
--- a/test/io/postgrest.py
+++ b/test/io/postgrest.py
@@ -40,8 +40,7 @@ class PostgrestSession(requests_unixsocket.Session):
 
     def __init__(self, baseurl, *args, **kwargs):
         super(PostgrestSession, self).__init__(*args, **kwargs)
-        # Keep local test requests independent from host proxy env vars.
-        self.trust_env = False
+        disable_env_proxy(self)
         self.baseurl = baseurl
 
     def request(self, method, url, *args, **kwargs):
@@ -49,6 +48,20 @@ class PostgrestSession(requests_unixsocket.Session):
         # well with our 'http+unix://' unix domain socket urls.
         fullurl = self.baseurl + url
         return super(PostgrestSession, self).request(method, fullurl, *args, **kwargs)
+
+
+def disable_env_proxy(session):
+    # `trust_env` controls whether proxy/env vars are honored by the client.
+    # Ref: https://www.python-httpx.org/environment_variables/
+    # We disable it so local probes never use HTTP(S)_PROXY/NO_PROXY.
+    session.trust_env = False
+    _ = session.trust_env
+    return session
+
+
+def new_local_session():
+    "Session used by local test helpers, ignoring host proxy env vars."
+    return disable_env_proxy(requests_unixsocket.Session())
 
 
 @dataclasses.dataclass
@@ -188,8 +201,7 @@ def wait_until_exit(postgrest):
 
 def wait_until_status_code(url, max_seconds, status_code):
     "Wait for the given HTTP endpoint to return a status code"
-    session = requests_unixsocket.Session()
-    session.trust_env = False
+    session = new_local_session()
     response = None
 
     for _ in range(max_seconds * 10):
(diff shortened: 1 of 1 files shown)


steve-chavez:
Fixes https://github.com/PostgREST/postgrest/issues/4633? I was puzzled for a while.

test/io/postgrest.py
```diff
@@ -40,6 +40,8 @@ class PostgrestSession(requests_unixsocket.Session):
 
     def __init__(self, baseurl, *args, **kwargs):
         super(PostgrestSession, self).__init__(*args, **kwargs)
+        # Keep local test requests independent from host proxy env vars.
+        self.trust_env = False
```
steve-chavez: So this essentially ignores env vars according to: https://www.python-httpx.org/environment_variables/

Maybe this can be mentioned in the comment, it's not that clear with the field name.

test/io/postgrest.py
```diff
@@ -207,6 +211,7 @@ def wait_until_status_code(url, max_seconds, status_code):
 def sleep_pool_connection(url, seconds):
     "Sleep a pool connection by calling an RPC that uses pg_sleep"
     session = requests_unixsocket.Session()
+    session.trust_env = False
```
steve-chavez: Maybe we can refactor this so `trust_env` is not repeated in many places?

nothankyouzzz:
> Fixes #4633? I was puzzled for a while.

@steve-chavez Yes, exactly. This PR is intended to fix #4633.

The root cause is that our IO test harness sessions were still honoring `HTTP_PROXY/HTTPS_PROXY` from the environment, which could route local readiness probes (notably IPv6 cases) through a proxy and cause failures/flakiness.  
This change disables env-proxy usage for those local harness sessions and also hardens the timeout path in `wait_until_status_code`.

test/io/postgrest.py
nothankyouzzz: Good call. I clarified the comment around `trust_env` and added an explicit reference to the env-var behavior.

The intent is now stated directly: local probes in the test harness should ignore host `HTTP(S)_PROXY` / `NO_PROXY`.

test/io/postgrest.py
nothankyouzzz: Agreed. I refactored this to avoid repetition.

I introduced a shared helper for local test sessions/probes and reused it in:
- `PostgrestSession.__init__`
- `wait_until_status_code`
- `sleep_pool_connection`

So proxy-env disabling is now centralized instead of repeated at each call site.

test/io/postgrest.py
```diff
@@ -49,6 +50,20 @@ def request(self, method, url, *args, **kwargs):
         return super(PostgrestSession, self).request(method, fullurl, *args, **kwargs)
 
 
+def disable_env_proxy(session):
+    # `trust_env` controls whether proxy/env vars are honored by the client.
+    # Ref: https://www.python-httpx.org/environment_variables/
+    # We disable it so local probes never use HTTP(S)_PROXY/NO_PROXY.
+    session.trust_env = False
+    _ = session.trust_env
```
wolfgangwalther, line 58: What is this line supposed to do?

nothankyouzzz: Suppress the linter warning about an unused attribute.

This is a deliberate workaround to handle the dual-linter setup (Ruff and Vulture). Vulture flags this as an "unused attribute" false positive. Since standard `# noqa` suppressions don't work well with Ruff's rules, I used this dummy assignment to explicitly register usage.

I can add a comment explaining this directly in the code if you think it would help clarity.

nothankyouzzz: These PRs were developed with heavy LLM support (GPT-5.3-Codex, Claude Opus 4.6, and Gemini 3.0 Pro), but not “LLM-only”.

Typically the workflow was: an LLM first analyzed the issue and proposed candidate fixes, or I proposed an approach and then had multiple models cross-check it. After we converged on a plan, the LLM produced the initial implementation, and then I reviewed it carefully (and also had multiple models review / sanity-check it) before submitting. I can explain the rationale behind each change and adjust anything that doesn’t match the project’s expectations.

steve-chavez:

PR #4635, opened 2026-02-07
refactor(app): unify QueryObs logging path (follow-up #4226)

nothankyouzzz:
## Summary

Behavior-preserving refactor of `postgrestResponse` in `src/PostgREST/App.hs` to remove duplicate query observation paths and keep a single `QueryObs` call site based on final response status.

## Related issue

- Follow-up to #4226

## Changes

- Keep transaction timing logic unchanged.
- Make tx stage return `Either Error txResult` instead of exiting early on `Left`.
- Build `responseE` from `txResultE` in one place.
- Compute final status once from `responseE`.
- Record `obsQuery` once from that final status.
- Remove the duplicate-observation TODO path in `App.hs`.

## Validation

- `postgrest-check`

### DIFF
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -17,7 +17,7 @@ module PostgREST.App
 
 
 import Control.Monad.Except     (liftEither)
-import Data.Either.Combinators  (mapLeft, whenLeft)
+import Data.Either.Combinators  (mapLeft)
 import Data.Maybe               (fromJust)
 import Data.String              (IsString (..))
 import Network.Wai.Handler.Warp (defaultSettings, setHost, setPort,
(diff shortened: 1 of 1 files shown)


nothankyouzzz:
I tried to reproduce this locally in `nix-shell` by running `postgrest-loadtest -k jwt-rsa-cache-worst` multiple times (2 full runs), but I could not reproduce the failure.

Local results:
- Run 1: 50,000 requests, 100% success, status codes `200:50000` (no `401`)
- Run 2: 50,000 requests, 100% success, status codes `200:50000` (no `401`)

So at the moment this looks more like CI-only flakiness than a deterministic regression from this PR.

Possible reasons:
1. `jwt-*-cache-worst` is intentionally time-sensitive, and token `exp` in `generate_targets.py` is based on estimated timing.
2. CI timing variance (cold Nix build/cache misses, shared-runner contention) can shift the timing window enough for some RSA tokens to expire during the run.
3. `postgrest-loadtest-against` compares multiple refs in one job, which can make one leg hit a less favorable timing window.

steve-chavez:

PR #4638, opened 2026-02-10
refactor: Failure in schema cache load should not set listener delay

mkleczek:
Listener has its own backoff logic independent of schema cache loading. It also uses listener delay value to identify its retries and to trigger schema cache reloads upon failures. If schema cache reloading sets listener delay it might lead to some unexpected reloading loops. That's especially important in set ups where listener connection points to master and the pool to replica(s).

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -364,7 +364,6 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
     when (rsIterNumber > 0) $ do
       let delay = fromMaybe 0 rsPreviousDelay `div` oneSecondInUs
       observer $ ConnectionRetryObs delay
-      putNextListenerDelay appState delay
 
     flushPool appState
 


src/PostgREST/AppState.hs
```diff
@@ -364,7 +364,6 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
     when (rsIterNumber > 0) $ do
       let delay = fromMaybe 0 rsPreviousDelay `div` oneSecondInUs
       observer $ ConnectionRetryObs delay
-      putNextListenerDelay appState delay
```
steve-chavez, line 367: This line was added on https://github.com/PostgREST/postgrest/pull/3654.

I don't really remember why and we lack confidence in these parts of the code because we have no tests: https://github.com/PostgREST/postgrest/issues/1766

steve-chavez: > If schema cache reloading sets listener delay it might lead to some unexpected reloading loops. That's especially important in set ups where listener connection points to master and the pool to replica(s).

Can you elaborate on the above? Maybe we can come up with a failing test.

mkleczek: > This line was added on #3654.
> 
> I don't really remember why and we lack confidence in these parts of the code because we have no tests: #1766

I don't understand why it was added either: listener has its own exponential backoff logic and I am not sure why would it need interference from cache reloading.

mkleczek: > > If schema cache reloading sets listener delay it might lead to some unexpected reloading loops. That's especially important in set ups where listener connection points to master and the pool to replica(s).
> 
> Can you elaborate on the above? Maybe we can come up with a failing test.

The scenario:

1. Master db is working fine
2. Replica is under load
3. PostgREST is configured to connect to replica (but listener is connecting to master)
4. PostgREST starts and runs schema reloading
5. (listener is not yet started)
6. Schema reloading fails and retries
7. Schema retry sets listener delay to a value > 1
8. Schema retry succeeds
9. Listener starts
10. Listener sees delay > 1 so it triggers schema cache reload

Another scenario is when listener triggers cache reload and schema loading fails and retries, next time listener has to retry it will wait too long (because its delay was set by schema reloading retry).

steve-chavez: Thanks for elaborating. I think we can delete but it doesn't look like we can call it a `fix:`? I mean even the current title doesn't make it clear.

So perhaps a `refactor:`?

mkleczek: > Thanks for elaborating. I think we can delete but it doesn't look like we can call it a `fix:`? I mean even the current title doesn't make it clear.
> 
> So perhaps a `refactor:`?

@steve-chavez Done

PR #4639, opened 2026-02-10
fix: Add random delay in Listener.cacheReloader

mkleczek:
Triggering schema cache reload immediately upon receival of notification by the listener leads to thundering herd problem in PostgREST cluster.

This change adds random delay between 0 and 1 seconds between notification retrieval and schema cache reload triggering.

### DIFF
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -158,6 +158,7 @@ library
                     , stm-hamt                  >= 1.2 && < 2
                     , focus                     >= 1.0 && < 2
                     , some                      >= 1.0.4.1 && < 2
+                    , MonadRandom               >= 0.6.2 && < 0.7
                       -- -fno-spec-constr may help keep compile time memory use in check,
                       --   see https://gitlab.haskell.org/ghc/ghc/issues/16017#note_219304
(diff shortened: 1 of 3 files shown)


steve-chavez:
Note: This is a potential solution to https://github.com/PostgREST/postgrest/issues/4642.

steve-chavez:
> This change adds random delay between 0 and 1 seconds between notification retrieval and schema cache reload triggering.

I'm wondering if 1 second is enough considering extreme cases like https://github.com/PostgREST/postgrest/issues/4642.

mkleczek:
> > This change adds random delay between 0 and 1 seconds between notification retrieval and schema cache reload triggering.
> 
> I'm wondering if 1 second is enough considering extreme cases like #4642.

Yeah, I share your concern. But didn't want to add a new config property.

After some more thought: maybe using advisory locks would be a better solution here?

mkleczek:
> After some more thought: maybe using advisory locks would be a better solution here?

@steve-chavez - see alternative solution: https://github.com/PostgREST/postgrest/pull/4643

mkleczek:
Superseded by https://github.com/PostgREST/postgrest/pull/4643

PR #4640, opened 2026-02-10
refactor: Simplify schema cache loading triggering logic

mkleczek:
**DISCLAIMER:**

Using debouncer to trigger schema cache loading makes it difficult to understand when exactly it is triggered.

This change is supposed to make this logic explicit. The idea is to heave a single thread waiting on an MVar and executing schema reload once the MVar is set. To trigger reload tryPutMvar is used.

There are also some issues with the way how `MVar`s are used in `Control.Debounce`. See eg. https://github.com/yesodweb/wai/issues/1027

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -117,13 +117,19 @@ init conf@AppConfig{configLogLevel, configDbPoolSize} = do
   pool <- initPool conf observer
   initWithPool pool conf loggerState metricsState observer --{ stateSocketREST = sock, stateSocketAdmin = adminSock}
 
-simpleDebounce :: IO () -> IO (IO ())
-simpleDebounce act = do
+-- Make a new debouncer action. An internal "worker" thread runs forever ensuring "action" runs when the "trigger" is called. The "action" is only executed once over a burst of calls.
+makeDebouncer :: IO () -> IO (IO ())
+makeDebouncer action = do
   flag <- newEmptyMVar
-  void $ forkIO $ forever $ do
-    takeMVar flag
-    act
-  pure (void $ tryPutMVar flag ())
+
+  let worker = forever $ do
+        takeMVar flag
+        action
+
+  let trigger = void $ tryPutMVar flag ()
+
+  void $ forkIO worker
+  pure trigger
 
 initWithPool :: SQL.Pool -> AppConfig -> Logger.LoggerState -> Metrics.MetricsState -> ObservationHandler -> IO AppState
 initWithPool pool conf loggerState metricsState observer = mdo
@@ -133,7 +139,7 @@ initWithPool pool conf loggerState metricsState observer = mdo
     <*> newIORef Nothing
     <*> newSchemaCacheStatus
     <*> newIORef False
-    <*> simpleDebounce (retryingSchemaCacheLoad appState *> threadDelay 100000)  -- 100ms cooldown
+    <*> makeDebouncer (retryingSchemaCacheLoad appState *> threadDelay 100000)  -- 100ms cooldown
     <*> newIORef conf
     <*> mkAutoUpdate defaultUpdateSettings { updateAction = getCurrentTime }
     <*> myThreadId


src/PostgREST/AppState.hs
```diff
@@ -153,13 +152,12 @@ initWithPool (sock, adminSock) pool conf loggerState metricsState observer = do
...
-       { debounceAction = retryingSchemaCacheLoad appState
-       , debounceFreq = decisecond
-       , debounceEdge = leadingEdge -- runs the worker at the start and the end
-       }
+  deb <- do
+    flag <- newEmptyMVar
+    void $ forkIO $ forever $ do
+      takeMVar flag
+      retryingSchemaCacheLoad appState
+    pure (void $ tryPutMVar flag ())
```
steve-chavez: 🤔 For me this doesn't make the code more readable and we have `Control.Debounce` used in other places too.

Perhaps we can add some comments about its inner workings instead?

It uses MVar internally for example https://hackage.haskell.org/package/auto-update-0.2.6/docs/src/Control.Debounce.Internal.html#mkDebounceInternal

steve-chavez: Another downside is that it reduces coverage: https://app.codecov.io/gh/PostgREST/postgrest/pull/4640?dropdown=coverage&src=pr&el=h1

mkleczek: > 🤔 For me this doesn't make the code more readable 

Fair enough. Need to think how to make it more readable.

> and we have `Control.Debounce` used in other places too.

And I'd say we should consider changing it in other places as well.

> 
> Perhaps we can add some comments about its inner workings instead?
> 
> It uses MVar internally for example https://hackage.haskell.org/package/auto-update-0.2.6/docs/src/Control.Debounce.Internal.html#mkDebounceInternal

It uses multiple `MVar`s which causes race conditions (the one I found is https://github.com/yesodweb/wai/issues/1027 but I suppose there might be more as the same pattern is repeated throughout the codebase).

mkleczek: > Another downside is that it reduces coverage: https://app.codecov.io/gh/PostgREST/postgrest/pull/4640?dropdown=coverage&src=pr&el=h1

Hmm... two lines are false positives.

And the third line is because we do not have any test covering debouncing logic (ie. notifications during cache reload) - we just assume `Control.Debounce` works properly.

src/PostgREST/AppState.hs
```diff
@@ -132,15 +132,23 @@ init conf@AppConfig{configLogLevel, configDbPoolSize} = do
   state' <- initWithPool (sock, adminSock) pool conf loggerState metricsState observer
   pure state' { stateSocketREST = sock, stateSocketAdmin = adminSock}
 
+simpleDebounce :: IO () -> IO (IO ())
+simpleDebounce act = do
+  flag <- newEmptyMVar
+  void $ forkIO $ forever $ do
+    takeMVar flag
+    act
+  pure (void $ tryPutMVar flag ())
```
steve-chavez: Would this debounce only trigger the action once? One thing that `mkDebounce` was missing to me is the ability to just trigger once at the end.

mkleczek: > Would this debounce only trigger the action once? One thing that `mkDebounce` was missing to me is the ability to just trigger once at the end.

I’m not sure I understand the requirement.

This debouncer makes sure only one task is executing in any single moment. It also “eats” multiple triggers encountered during task execution and re-executes the task once.

steve-chavez: Ok cool. Was just wondering how it worked, this shouldn't be a blocker.

src/PostgREST/AppState.hs
steve-chavez: > It uses multiple MVars which causes race conditions (the one I found is https://github.com/yesodweb/wai/issues/1027 but I suppose there might be more as the same pattern is repeated throughout the codebase).

That is worrying, given that I'd agree on replacing the functionality. Specially since the code looks simple.

steve-chavez:
@mkleczek Now that we have https://github.com/PostgREST/postgrest/commit/8d5b1018d3d3f218ea6ff88a4d91279957188408, would it be possible to test that debouncing is done for logs?

Once we do that I think we can merge here and also replace the `Control.Debounce` used in other places too.

mkleczek:
> @mkleczek Now that we have [8d5b101](https://github.com/PostgREST/postgrest/commit/8d5b1018d3d3f218ea6ff88a4d91279957188408), would it be possible to test that debouncing is done for logs?
> 

See #4772

steve-chavez:
This one needs rebasing, it has the commit on https://github.com/PostgREST/postgrest/pull/4772

src/PostgREST/AppState.hs
```diff
@@ -117,15 +117,23 @@ init conf@AppConfig{configLogLevel, configDbPoolSize} = do
   pool <- initPool conf observer
   initWithPool pool conf loggerState metricsState observer --{ stateSocketREST = sock, stateSocketAdmin = adminSock}
 
+simpleDebounce :: IO () -> IO (IO ())
+simpleDebounce act = do
```
steve-chavez: I've noticed this function is actually a closure that carries some internal state.

I think renaming it to `makeDebouncer` would be clearer.

src/PostgREST/AppState.hs
```diff
@@ -117,15 +117,23 @@ init conf@AppConfig{configLogLevel, configDbPoolSize} = do
...
+  flag <- newEmptyMVar
+  void $ forkIO $ forever $ do
+    takeMVar flag
+    act
+  pure (void $ tryPutMVar flag ())
```
steve-chavez:
```suggestion
-- Make a new debouncer action. An internal "worker" thread runs forever ensuring "action" runs when the "trigger" is called. The "action" is only executed once over a burst of calls.
makeDebouncer :: IO () -> IO (IO ())
makeDebouncer action = do
  flag <- newEmptyMVar
  let worker = forever $ do
    takeMVar flag
    action
  let trigger = void $ tryPutMVar flag ()
  void $ forkIO worker
  pure trigger
```

src/PostgREST/AppState.hs
```diff
@@ -117,15 +117,23 @@ init conf@AppConfig{configLogLevel, configDbPoolSize} = do
...
-initWithPool pool conf loggerState metricsState observer = do
+initWithPool pool conf loggerState metricsState observer = mdo
 
   appState <- AppState pool
     <$> newIORef minimumPgVersion -- assume we're in a supported version when starting, this will be corrected on a later step
     <*> newIORef Nothing
     <*> newSchemaCacheStatus
     <*> newIORef False
-    <*> pure (pure ())
+    <*> simpleDebounce (retryingSchemaCacheLoad appState *> threadDelay 100000)  -- 100ms cooldown
```
steve-chavez:
```suggestion
    <*> makeDebouncer (retryingSchemaCacheLoad appState *> threadDelay 100000)  -- 100ms cooldown
```

steve-chavez:
@mkleczek As a follow up we can now remove the last Control.Debounce usage:

https://github.com/PostgREST/postgrest/blob/b610b1ea5199ed97107043a40b324724cefd67dd/src/PostgREST/Logger.hs#L61-L65

steve-chavez:
Also just noticed this commit https://github.com/PostgREST/postgrest/commit/a4c1d945ee2ae1bf92aa96bc8a66a79a3d89f48b was merged with:

```
DISCLAIMER:
This commit was authored entirely by a human without the assistance of LLMs.
```

I didn't notice that.

@mkleczek Please put those disclaimers on PRs only, otherwise it makes git history harder to read. Also gives more work to reviewers since we have to squash/edit the commit message to remove the disclaimers.

mkleczek:
> @mkleczek As a follow up we can now remove the last Control.Debounce usage:
> 
> https://github.com/PostgREST/postgrest/blob/b610b1ea5199ed97107043a40b324724cefd67dd/src/PostgREST/Logger.hs#L61-L65

I would have done it in this PR if it was straightforward :)

The problem is that there is a dependency from `AppState` to `Logger` so `Logger` cannot import `AppState` as it would introduce dep cycle.
It would be awkward to have `makeDebouncer` in `Logger` and use it in `AppState`.
It seems like we need a new module to resolve the above issue but what name should it have? `PostgREST.Debounce` with a single function? Maybe some `Util` module (but that would be smelly).

WDYT @steve-chavez ?

steve-chavez:
> It seems like we need a new module to resolve the above issue but what name should it have? PostgREST.Debounce with a single function? 

@mkleczek I think that would be good for now, later it could be moved to our own Prelude https://github.com/PostgREST/postgrest/issues/4775

PR #4643, opened 2026-02-10
fix: Limit concurrent schema cache loads

### LINKED ISSUE #4642: Thundering herd problem in PostgREST cluster on AWS ECS
## Problem

In a setup where:

- There's a master database and several replicas 
- The master and the replicas each have their own ECS task with a postgREST instance (which can scale and multiply according to demand, this can reach up to 100 instances connecting to the same db), all of them have a LISTEN channel
- A migration occurs and [event trigger for reloading is setup](https://docs.postgrest.org/en/v14/references/schema_cache.html#automatic-schema-cache-reloading), since master + all the replicas LISTEN channel are connected to the master ([ref](https://docs.postgrest.org/en/v14/references/listener.html#listener-on-read-replicas)), the schema cache process for each one is triggered at the same time.
- All the schema cache queries fire at the same time, causing thundering herd -- which makes the schema cache take longer than usual

## Proposal

Having a way to read the schema cache from a file and sharing across instances could potentially mitigate this, also discussed before on https://github.com/PostgREST/postgrest/discussions/2918

mkleczek:
**DISCLAIMER:**

Triggering schema cache reload immediately upon receival of notification by the listener leads to thundering herd problem in PostgREST cluster.

This change adds limiting of number of concurrent schema cache loading queries using advisory locks.

Fixes  #4642

### DIFF
--- a/src/library/PostgREST/AppState.hs
+++ b/src/library/PostgREST/AppState.hs
@@ -1,7 +1,9 @@
-{-# LANGUAGE LambdaCase      #-}
-{-# LANGUAGE NamedFieldPuns  #-}
-{-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE RecursiveDo     #-}
+{-# LANGUAGE LambdaCase       #-}
+{-# LANGUAGE NamedFieldPuns   #-}
+{-# LANGUAGE QuasiQuotes      #-}
+{-# LANGUAGE RecordWildCards  #-}
+{-# LANGUAGE RecursiveDo      #-}
+{-# LANGUAGE TypeApplications #-}
 
 module PostgREST.AppState
   ( AppState
@@ -33,7 +35,8 @@ import qualified Data.ByteString.Char8      as BS
 import           Data.Either.Combinators    (whenLeft)
 import qualified Hasql.Pool                 as SQL
 import qualified Hasql.Pool.Config          as SQL
-import qualified Hasql.Session              as SQL
+import qualified Hasql.Session              as SQL hiding (statement)
+import qualified Hasql.Transaction          as SQL hiding (sql)
 import qualified Hasql.Transaction.Sessions as SQL
 import qualified Network.HTTP.Types.Status  as HTTP
 import qualified PostgREST.Auth.JwtCache    as JwtCache
@@ -59,9 +62,15 @@ import PostgREST.Config.Database         (queryDbSettings, queryPgVersion,
                                           queryRoleSettings)
 import PostgREST.Config.PgVersion        (PgVersion (..), minimumPgVersion)
 import PostgREST.Debounce                (makeDebouncer)
+import PostgREST.Metrics                 (MetricsState (connTrack))
 import PostgREST.SchemaCache             (SchemaCache (..), querySchemaCache,
                                           showSummary)
 import PostgREST.SchemaCache.Identifiers (quoteQi)
+
+import qualified Hasql.Decoders    as HD
+import qualified Hasql.Encoders    as HE
+import qualified Hasql.Statement   as SQL
+import           NeatInterpolation (trimming)
 
 import Protolude
 
@@ -288,7 +297,7 @@ getObserver = stateObserver
 -- + Because connections cache the pg catalog(see #2620)
 -- + For rapid recovery. Otherwise, the pool idle or lifetime timeout would have to be reached for new healthy connections to be acquired.
 retryingSchemaCacheLoad :: AppState -> IO ()
-retryingSchemaCacheLoad appState@AppState{stateObserver=observer} =
+retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMetrics} =
   void $ retrying retryPolicy shouldRetry (\RetryStatus{rsIterNumber, rsPreviousDelay} -> do
     when (rsIterNumber > 0) $ do
       let delay = fromMaybe 0 rsPreviousDelay `div` oneSecondInUs
@@ -327,8 +336,22 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer} =
     qSchemaCache :: IO (Maybe SchemaCache)
     qSchemaCache = do
       conf@AppConfig{..} <- getConfig appState
+      -- Throttle concurrent schema cache loads, guarded by advisory locks.
+      -- This is to prevent thundering herd problem on startup or when many PostgREST
+      -- instances receive "reload schema" notifications at the same time
+      -- See get_lock_sql for details of the algorithm.
+      -- Here we calculate the number of open connections passed to the query.
+      Metrics.ConnStats connected inUse <- Metrics.connectionCounts $ connTrack stateMetrics
+      -- Determine whether schema cache loading will create a new session
+      let
+        -- if all connections in use but pool not full - schema cache loading will create session
+        scLoadingSessions = if connected <= inUse && inUse < configDbPoolSize then 1 else 0
+        withTxLock = SQL.statement
+          (fromIntegral $ connected + scLoadingSessions)
+          (SQL.Statement get_lock_sql get_lock_params HD.noResult configDbPreparedStatements)
+
       (resultTime, result) <-
-        timeItT $ usePool appState (SQL.transactionNoRetry SQL.ReadCommitted SQL.Read $ querySchemaCache conf)
+        timeItT $ usePool appState (SQL.transactionNoRetry SQL.ReadCommitted SQL.Read $ withTxLock *> querySchemaCache conf)
       case result of
         Left e -> do
           markSchemaCachePending appState
@@ -349,6 +372,43 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer} =
           observer $ SchemaCacheLoadedObs loadTime summary
           markSchemaCacheLoaded appState
           return $ Just sCache
+      where
+        -- Recursive query that tries acquiring locks in order
+        -- and waits for randomly selected lock if no attempt succeeded.
+        -- It has a single parameter: this node open connection count.
+        -- It is used to estimate the number of nodes
+        -- by counting the number of active sessions for current session_user
+        -- and dividing it by this node open connections.
+        -- Assuming load is uniform among cluster nodes, all should have
+        -- statistically the same number of open connections.
+        -- Once the number of nodes is known we calculate the number
+        -- of locks as ceil(log(2, number_of_nodes))
+        get_lock_sql = encodeUtf8 [trimming|
+          WITH RECURSIVE attempts AS (
+            SELECT 1 AS lock_number, pg_try_advisory_xact_lock(lock_id, 1) AS success FROM parameters
+            UNION ALL
+            SELECT next_lock_number AS lock_number, pg_try_advisory_xact_lock(lock_id, next_lock_number) AS success
+            FROM
+              parameters CROSS JOIN LATERAL (
+                  SELECT lock_number + 1 AS next_lock_number FROM attempts
+                  WHERE NOT success AND lock_number < locks_count
+                  ORDER BY lock_number DESC
+                  LIMIT 1
+                ) AS previous_attempt
+          ),
+          counts AS (
+            SELECT round(log(2, round(count(*)::double precision/$$1)::numeric))::int AS locks_count
+            FROM
+              pg_stat_activity WHERE usename = SESSION_USER
+          ),
+          parameters AS (
+            SELECT locks_count, 50168275 AS lock_id FROM counts WHERE locks_count > 0
+          )
+          SELECT pg_advisory_xact_lock(lock_id, floor(random() * locks_count)::int + 1)
+          FROM
+            parameters WHERE NOT EXISTS (SELECT 1 FROM attempts WHERE success) |]
+
+        get_lock_params = HE.param (HE.nonNullable HE.int4)
 
     shouldRetry :: RetryStatus -> (Maybe PgVersion, Maybe SchemaCache) -> IO Bool
     shouldRetry _ (pgVer, sCache) = do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file. From versio
 - Fix responding with `Something went wrong` on Admin server when under EMFILE by @mkleczek in #5077
 - Fix schema cache dump missing RPC transaction isolation level by @taimoorzaeem in #5079
 - Fix admin server crashing without a way to recover by @taimoorzaeem in #5096
+- Limit concurrent schema cache loads by @mkleczek in #4643
 
 ### Changed
 
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1,5 +1,6 @@
 "Unit tests for Input/Ouput of PostgREST seen as a black box."
 
+import contextlib
 import os
 import re
 import signal
@@ -30,6 +31,7 @@ from postgrest import (
     sleep_until_postgrest_full_reload,
     sleep_until_postgrest_scache_reload,
     wait_until_exit,
+    wait_until_status_code,
 )
 
 
@@ -1274,6 +1276,93 @@ def test_schema_cache_concurrent_notifications(slow_schema_cache_env):
         response = postgrest.session.get("/rpc/mult_them?c=3&d=4")
         assert response.text == "12"
         assert response.status_code == 200
+
+
+@pytest.mark.parametrize(
+    "instance_count, expected_concurrency", [(2, 2), (4, 3), (6, 4), (8, 4), (16, 5)]
+)
+def test_schema_cache_reload_throttled_with_advisory_locks(
+    instance_count, expected_concurrency, slow_schema_cache_env
+):
+    "schema cache reloads should be throttled across instances"
+
+    internal_sleep_ms = int(
+        slow_schema_cache_env["PGRST_INTERNAL_SCHEMA_CACHE_QUERY_SLEEP"]
+    )
+    lock_wait_threshold_ms = internal_sleep_ms * 2
+    query_log_pattern = re.compile(r"Schema cache queried in ([\d.]+) milliseconds")
+
+    def read_available_output_lines(postgrest):
+        try:
+            output = postgrest.process.stdout.read()
+        except BlockingIOError:
+            return []
+
+        if not output:
+            return []
+        return output.decode().splitlines()
+
+    with contextlib.ExitStack() as stack:
+        instances = [
+            stack.enter_context(
+                run(
+                    env=slow_schema_cache_env,
+                    wait_for=None,
+                    wait_max_seconds=10,
+                )
+            )
+            for _ in range(instance_count)
+        ]
+
+        for postgrest in instances:
+            wait_until_status_code(
+                postgrest.admin.baseurl + "/ready", max_seconds=10, status_code=200
+            )
+
+        # Drop startup logs so only reload logs are parsed.
+        for postgrest in instances:
+            read_available_output_lines(postgrest)
+
+        response = instances[0].session.get("/rpc/notify_pgrst")
+        assert response.status_code == 204
+
+        # Wait long enough for the lock-throttled cache reloads to finish.
+        time.sleep((internal_sleep_ms / 1000) * 2)
+
+        reload_durations_ms = []
+        for postgrest in instances:
+            output_lines = []
+            for _ in range(instance_count * 2):
+                output_lines.extend(read_available_output_lines(postgrest))
+                if any(query_log_pattern.search(line) for line in output_lines):
+                    break
+                time.sleep(0.2)
+
+            durations = []
+            for line in output_lines:
+                match = query_log_pattern.search(line)
+                if match:
+                    durations.append(float(match.group(1)))
+
+            assert durations
+            reload_durations_ms.append(max(durations))
+
+        assert len(reload_durations_ms) == instance_count
+
+        # expected_concurrency instances should have
+        # reload_durations_ms <= lock_wait_threshold_ms
+        # the rest should wait
+        assert (
+            instance_count
+            - len(
+                [
+                    duration
+                    for duration in reload_durations_ms
+                    if duration > lock_wait_threshold_ms
+                ]
+            )
+            == expected_concurrency
+        )
 
 
 def test_schema_cache_query_sleep_logs(defaultenv):


src/PostgREST/AppState.hs
```diff
@@ -401,9 +407,16 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
     qSchemaCache :: IO (Maybe SchemaCache)
     qSchemaCache = do
       conf@AppConfig{..} <- getConfig appState
+      withTxLock <- do
+        -- Allow 10 concurrent schema cache loads, guarded by advisory locks.
+        -- This is to prevent thundering herd problem on startup or when many PostgREST instances receive "reload schema" notifications at the same time
+        lockId <- getRandomR (50168275::Int64, 50168275 + 10)
+        let stmt = SQL.Statement "SELECT pg_catalog.pg_advisory_xact_lock($1)" (HE.param $ HE.nonNullable HE.int8) HD.noResult configDbPreparedStatements
```
steve-chavez: These locks would be released automatically at the end of the transaction right? It does look like it would work for https://github.com/PostgREST/postgrest/issues/4642.

I guess one drawback is that these advisory locks would run and leave a log trace even if the user will never run into https://github.com/PostgREST/postgrest/issues/4642, which are most cases.

WDYT of the solution on https://github.com/PostgREST/postgrest/issues/4642#issuecomment-3882997339? Would that be preferable?

steve-chavez: >I guess one drawback

Also that it's a bit more operational overhead, we would also have to recommend setting `lock_timeout` in addition to `statement_timeout` to avoid waiting for too long? (like on a schema cache load that takes too long due to pg_catalog bloat)

mkleczek: > > I guess one drawback
> 
> Also that it's a bit more operational overhead, we would also have to recommend setting `lock_timeout` in addition to `statement_timeout` to avoid waiting for too long? (like on a schema cache load that takes too long due to pg_catalog bloat)

Is it really an issue? If we get `lock timeout` we are going to retry anyway.

mkleczek: > These locks would be released automatically at the end of the transaction right? It does look like it would work for #4642.

Yeah, they are tx scoped.

> 
> I guess one drawback is that these advisory locks would run and leave a log trace even if the user will never run into #4642, which are most cases.

Maybe we should introduce a config property that activates it then?

> 
> WDYT of the solution on [#4642 (comment)](https://github.com/PostgREST/postgrest/issues/4642#issuecomment-3882997339)? Would that be preferable?

See https://github.com/PostgREST/postgrest/issues/4642#issuecomment-3887355017

I think for now mitigation of thundering herd problem is way more feasible. At least in the short to medium term.

src/PostgREST/AppState.hs
```diff
@@ -401,9 +407,16 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
     qSchemaCache :: IO (Maybe SchemaCache)
     qSchemaCache = do
       conf@AppConfig{..} <- getConfig appState
+      withTxLock <- do
+        -- Allow 10 concurrent schema cache loads, guarded by advisory locks.
+        -- This is to prevent thundering herd problem on startup or when many PostgREST instances receive "reload schema" notifications at the same time
+        lockId <- getRandomR (50168275::Int64, 50168275 + 10)
```
steve-chavez: What's the reasoning behind this magic number?

src/PostgREST/AppState.hs
steve-chavez: > I think for now mitigation of thundering herd problem is way more feasible. At least in the short to medium term.
> Maybe we should introduce a config property that activates it then?

Yes, agree... a config sounds good. Should we parametrize the number of locks? Or should we just hardcode to 10 and expose a boolean config? (Also how do we know 10 is the right number?)

 We would also need to test this, seems doable to ensure only 10 connections can exist at a time when say 20 postgREST instances of db-pool=1 + with a PGRST_INTERNAL_SCHEMA_CACHE_QUERY_SLEEP are spawned.

src/PostgREST/AppState.hs
mkleczek: > What's the reasoning behind this magic number?

It is just randomly generated large number. We need something with low probability of being used by something else than PostgREST here (but it needs to be hardcoded constant so that there is no risk of instances using different locks).

src/PostgREST/AppState.hs
mkleczek: > > I think for now mitigation of thundering herd problem is way more feasible. At least in the short to medium term.
> > Maybe we should introduce a config property that activates it then?
> 
> Yes, agree... a config sounds good. 
> Should we parametrize the number of locks? Or should we just hardcode to 10 and expose a boolean config? (Also how do we know 10 is the right number?)

After giving it some more thought I am not so sure anymore. If the config is opt-in then it imposes operational burden to make sure all instances have the same config for the feature to work.
Making it opt-out means 99% of users won't turn it off but will add cognitive load to configuration.

The question really is about the reasons we would want to turn it off. I can think of two:

1. these advisory locks would run and leave a log trace
2. for large clusters it will make full cluster reloading slower

To fix 1 we could simply make it clear in the documentation that schema reloading uses advisory locks for throttling.

Fixing 2 is more difficult At the end of the day it is a matter of choosing the right number of locks.
My thinking here is that it should be big enough so that for 99.9% of cases it won't throttle anything. But small enough to make throttling fix the problem in 0.1% cases where it matters.
Unfortunately, I don't have statistics from the field (ie. what PostgREST cluster size is most popular). From my experience, 2 scaled to 8 (by k8s autoscaler) instances of PostgREST were more than enough to serve really large traffic - these cases are not affected by throttling at all.
The cluster reported in https://github.com/PostgREST/postgrest/issues/4642 has up to 100 instances connected to the same db. Assuming `0.5 s` querying time, throttling it to 10 concurrent instances would yield `5 s` full cluster reload time - which seems sensible (especially knowing that without throttling the queries started to take `> 5s` time).

I think it is a good case for https://github.com/PostgREST/postgrest/issues/4654 WDYT?

> 
> We would also need to test this, seems doable to ensure only 10 connections can exist at a time when say 20 postgREST instances of db-pool=1 + with a PGRST_INTERNAL_SCHEMA_CACHE_QUERY_SLEEP are spawned.

Will try to think about something.

mkleczek: > 
> We would also need to test this, seems doable to ensure only 10 connections can exist at a time when say 20 postgREST instances of db-pool=1 + with a PGRST_INTERNAL_SCHEMA_CACHE_QUERY_SLEEP are spawned.

Added `test_schema_cache_reload_throttled_with_advisory_locks` io test and updated the lock acquiring logic to first try all locks in order before waiting for randomly chosen one. This is necessary to make sure no throttling takes place in uncontended cases.

Pending decision on introducing a config option, I think this is now in a pretty good shape.

wolfgangwalther: In the absence of data from the field (which would vary a lot between different cases anyway), we should use a "sane" strategy, that doesn't need tuning.

To me, it seems like the best strategy here would be to only allow *a single* schema cache reload at a time. Any number bigger than 1 really makes big assumptions about the environment that this is run in.

Yes, we will probably need other improvements in the future, for example caching the schema cache queries as discussed elsewhere. But any of that also needs a discussion whether running 100 PostgREST instances is really the *right* thing to do in the first place. The fact that we need state in the form of the schema cache also means we are limited in the ability to scale that way - that's just a fact, that others have noted as well.

mkleczek: > In the absence of data from the field (which would vary a lot between different cases anyway), we should use a "sane" strategy, that doesn't need tuning.
> 
> To me, it seems like the best strategy here would be to only allow _a single_ schema cache reload at a time. Any number bigger than 1 really makes big assumptions about the environment that this is run in.
> 

Well... Today we don't have **any** throttling, so a sane default is such that nothing changes for most of the users as long as it protects from thundering herd. And while `1` protects from thundering herd, it is not transparent for many users (ie. users that have multi-instance set-ups).

> Yes, we will probably need other improvements in the future, for example caching the schema cache queries as discussed elsewhere. But any of that also needs a discussion whether running 100 PostgREST instances is really the _right_ thing to do in the first place.

It is actually _wrong_ :-) - we're in agreement here. But from the point of view of PostgREST project - we just learned our software is used in a certain way that we might not like but here we are... We can add guardrails so that even if PostgREST is used that way, it will still work fine.
I actually like software that provides high level of robustness, low operational burden and tolerance to misuse.

> The fact that we need state in the form of the schema cache also means we are limited in the ability to scale that way - that's just a fact, that others have noted as well.

**Any** form of caching will introduce issues with state management, synchronization, staleness etc. Whether it's schema cache or query cache, the problems are going to be the same.

steve-chavez: > To me, it seems like the best strategy here would be to only allow a single schema cache reload at a time. Any number bigger than 1 really makes big assumptions about the environment that this is run in.
> I think it is a good case for https://github.com/PostgREST/postgrest/issues/4654 WDYT?

Agree, I think `1` is a good number.  Given that I think we should avoid a config and just set it. 

> And while 1 protects from thundering herd, it is not transparent for many users (ie. users that have multi-instance set-ups).

Perhaps we can log that the lock is being applied (+ the time it takes releasing it) to make it more transparent?

Also I believe this should be a feature instead of a fix. That way we'd document it and users would be aware of it.

mkleczek: > > To me, it seems like the best strategy here would be to only allow a single schema cache reload at a time. Any number bigger than 1 really makes big assumptions about the environment that this is run in.
> > I think it is a good case for #4654 WDYT?
> 
> Agree, I think `1` is a good number. Given that I think we should avoid a config and just set it.

Hmm... but why? What harm is in allowing _some_ level of concurrency? Especially that right now there is no limit at all?

If the limit is so low (ie. `1`) I am strongly against forcing it on users without a way to opt-out.

For example: if I add a new table/view, I want the schema cache to be reloaded as quickly as possible **on all nodes**. The longer the transition takes the more `404`s is going to be reported (as nodes with stale schema cache will report `404` for the new table requests).
And if I have, let's say, `8` nodes in the cluster - it is now going to take 8 times longer to reload the cache on all nodes.
In case of https://github.com/PostgREST/postgrest/issues/4642 (>100 nodes) - full reload is going to take more than `50s` during which `404`s are going to be reported.

> 
> > And while 1 protects from thundering herd, it is not transparent for many users (ie. users that have multi-instance set-ups).
> 
> Perhaps we can log that the lock is being applied (+ the time it takes releasing it) to make it more transparent?

We really can't as taking the lock has to be done in the same transaction and releasing it is automatic at commit/rollback time.

EDIT: we can try with `unsafePerformIO` as by default we cannot perform IO actions inside an `SQL.Transaction`

> 
> Also I believe this should be a feature instead of a fix. That way we'd document it and users would be aware of it.

+1

mkleczek: > > Agree, I think `1` is a good number. Given that I think we should avoid a config and just set it.
> 
> Hmm... but why? What harm is in allowing _some_ level of concurrency? Especially that right now there is no limit at all?

@steve-chavez @wolfgangwalther 
I think we can actually do better.

How about we adjust the number of locks based on the (estimated) number of nodes connected to the same database?
There are two issues to solve:

1. How do we estimate the number of cluster nodes?
2. What should be the algorithm to calculate the number of locks?

### Node number estimation
The idea is to estimate that based on:

- number of active db sessions opened by the same user as `session_user`
- number of open connections in the pool

The estimate would be: `active_sessions_number / connections_in_the_pool`

This assumes the load is spread evenly among cluster members so all nodes should have the same number of open connections.

### Number of locks calculation
We need a sublinear function and it seems to me logarithm is well fit. The number of locks would be `round(log2(estimated_number_of_nodes))`

That way we can allow concurrent schema loads while protecting from thundering herd issue in large cluster.

I've committed implementation of this idea for you to review. If you don't like it we can easily delete the commit. If you think it is OK, we can split into coherent pieces.
Added the test that verifies the level of concurrency for various cluster sizes and the results are as follows:
| Nodes | Locks |
| --- | --- |
| 2 | 2 |
| 4 | 3 |
6 | 4
8 | 4
16 | 5

WDYT?

steve-chavez: > Hmm... but why? What harm is in allowing some level of concurrency? Especially that right now there is no limit at all?
> If the limit is so low (ie. 1) I am strongly against forcing it on users without a way to opt-out.

Right, so if only one lock can be taken we would be forcing all scache loads to be sequential. If we consider the case of 100 instances (https://github.com/PostgREST/postgrest/issues/4642) then all first 99 have to be loaded before the 100th can even begin right? And yes that doesn't look good for the last instance since it will prolong the time it will have a stale schema cache.

---

> Node number estimation
> Number of locks calculation

Seems complicated. One simpler idea that ocurrs to me:

1. Each postgREST instance has a corresponding LISTEN channel. So we know how many concurrent scache loads will happen.
2. We can also know the time the latest scache query took, since we have the metric https://github.com/PostgREST/postgrest/blob/66fda76644d58e4257490a6123e381b2ace3d8fe/src/PostgREST/Metrics.hs#L29
 
Perhaps we can sample the first scache query time (2) and combined with 1 calculate the right number of locks?

mkleczek: > > Hmm... but why? What harm is in allowing some level of concurrency? Especially that right now there is no limit at all?
> > If the limit is so low (ie. 1) I am strongly against forcing it on users without a way to opt-out.
> 
> Right, so if only one lock can be taken we would be forcing all scache loads to be sequential. If we consider the case of 100 instances (#4642) then all first 99 have to be loaded before the 100th can even begin right? And yes that doesn't look good for the last instance since it will prolong the time it will have a stale schema cache.
> 
> > Node number estimation
> > Number of locks calculation
> 
> Seems complicated. One simpler idea that ocurrs to me:
> 
> 1. Each postgREST instance has a corresponding LISTEN channel. So we know how many concurrent scache loads will happen.

I am afraid I don't understand. How do you want to count the number of Pgrst instances based on LISTEN channel?

> 2. We can also know the time the latest scache query took, since we have the metric https://github.com/PostgREST/postgrest/blob/66fda76644d58e4257490a6123e381b2ace3d8fe/src/PostgREST/Metrics.hs#L29
> 
> Perhaps we can sample the first scache query time (2) and combined with 1 calculate the right number of locks?

OK, but what should be the formula to calculate the number of locks?

IMHO, If you already have the estimated number of nodes, calculating `log2(number_of_nodes)` is as simple as it gets - no additional information required.

The main reason why the proposal in this PR has merit IMHO is all required information is available locally to the node (ie. it is only the number of its opened connections). So acquiring the lock can be done with a single `SELECT` query taking a single parameter.

wolfgangwalther: > Also I believe this should be a feature instead of a fix.

I think we're looking at two separate issues here:
1. It's a bug that multiple PostgREST instances just end up as a thundering herd.
2. There's a performance problem in reloading multiple PostgREST instances at the same time.

We should fix the bug by limiting to 1 concurrent schema cache reloader. We should then discuss how we can best improve performance. This discussion is inflating both.

mkleczek: > 
> 1. It's a bug that multiple PostgREST instances just end up as a thundering herd.
> 2. There's a performance problem in reloading multiple PostgREST instances at the same time.
> 
> We should fix the bug by limiting to 1 concurrent schema cache reloader. 

Such a "fix" would introduce another (major) bug and hence is not acceptable IMO.

> We should then discuss how we can best improve performance. This discussion is inflating both.

Given the above, we must discuss both to come up with the right solution, I'm afraid.

mkleczek:
@steve-chavez @wolfgangwalther rebased on top of #4672 to make review easier.

steve-chavez:
Another option to workaround this issue for now https://github.com/PostgREST/postgrest/issues/4642#issuecomment-4091513699

src/library/PostgREST/AppState.hs
```diff
@@ -359,6 +382,43 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
           observer $ SchemaCacheLoadedObs loadTime summary
           markSchemaCacheLoaded appState
           return $ Just sCache
+      where
+        -- Recursive query that tries acquiring locks in order
+        -- and waits for randomly selected lock if no attempt succeeded.
+        -- It has a single parameter: this node open connection count.
+        -- It is used to estimate the number of nodes
+        -- by counting the number of active sessions for current session_user
```
wolfgangwalther, line 380: So this assumes that all other instances are running with the same user as the current session?

That seems overly specific to a certain use-case for something to put into PostgREST.

mkleczek: > So this assumes that all other instances are running with the same user as the current session?
> 
> That seems overly specific to a certain use-case for something to put into PostgREST.

Is it really overly specific?

I cannot imagine hundreds (or even tens) of PostgREST instances connected to the same database, each configured differently using different user and listening for the same schema changes... What would be the goal of doing that?
The thundering herd issue in reality happens only in clustered environments where identical PostgREST instances are started to horizontally scale the cluster.

This patch prevents thundering herd issue in such cases.

Do you have any other cases in mind?

Also - do you have any ideas how to estimate the size of the PostgREST cluster without this assumption?

@steve-chavez WDYT?

wolfgangwalther: I can easily imagine a multi-tenant use-case, where each tenant gets their own authenticator role and their own PostgREST instance - but all connecting to the same database.

mkleczek: > I can easily imagine a multi-tenant use-case, where each tenant gets their own authenticator role and their own PostgREST instance - but all connecting to the same database.

That would be a schema based multitentancy with shared role structure **but** for some reason stateless PostgREST instances have to be separate **but** listening on the same channel. In other words in such setup one tenant running `NOTIFY pgrst` would force other tenants to reload the schema cache. Why would someone do that???
What's more - how many such PostgREST instances do you imagine in such a setup? So many that it would cause thundering herd problem?

wolfgangwalther: Why would that be schema based multitenancy? I don't see it that way. They'd all be using the same schema, but need different JWT settings.

mkleczek: > Why would that be schema based multitenancy? I don't see it that way. They'd all be using the same schema, but need different JWT settings.

And why would PostgREST use a different user to connect to the database in such a setup?

Also: How many different PostgREST instances would you have (or rather: different database users used by PostgREST to connect to the database.

EDIT:
IMHO, this is such an unusual set-up and, what's more, contradicting what you (rightly) say about multitenancy in other discussions that I wouldn't take it into account when talking about thundering herd issue.

The problem we face is that **any** dynamic limit requires estimation of the number of PostgREST nodes. Any such estimation is imperfect - that's true. That leaves us with a static limit imposed by configuration. IMO that's strictly worse than dynamic limit based on imperfect estimation but I think we might implement both (where static configuration overrides any dynamic algorithm).

wolfgangwalther: > And why would PostgREST use a different user to connect to the database in such a setup?

Because otherwise you can't separate multiple tenants from each other, *if they provide their own JWT signing*. The scenario is:
- Each client has their own infra which signs JWTs.
- Each client has their own authenticator role to which all their user roles are granted.
- Now you need to map public JWT keys and authenticator roles - you can only do this by giving each a separate instance.

Now you might think that you could just provide a JWKS with all the public keys from all clients - but once you do that, any client can impersonate any other client. Doesn't work.

> contradicting what you (rightly) say about multitenancy in other discussions

Eh, what? This is role-based multi-tenancy, just with an additional aspect. I have been advocating for role-based multi-tenancy forever.

> The problem we face is that **any** dynamic limit requires estimation of the number of PostgREST nodes.

Most certainly you can do better than counting connections with the same user as the current user. I think you also pointed out earlier, that this role could also be used to connect to the database from other applications and such.

It should be possible to count "connected PostgREST listener threads" by application name and last client statement or so? Isn't that *precisely* the number of nodes which would affect concurrent schema loads, because the listener thread is what is required to do that?

We could even put something more specific in the application name for that connection, so that it would be enough to filter by that. This has the added benefit, that we'd know that this version *supports the same dynamic limitation algorithm*. So something like "PostgREST" for regular connections and "PostgREST listener thread" for the latter?

mkleczek: > > And why would PostgREST use a different user to connect to the database in such a setup?
> 
> Because otherwise you can't separate multiple tenants from each other, _if they provide their own JWT signing_.

OK.
We are disgressing really.

> > The problem we face is that **any** dynamic limit requires estimation of the number of PostgREST nodes.
> 
> Most certainly you can do better than counting connections with the same user as the current user. 

Sure - I am happy to discuss other ideas!

> It should be possible to count "connected PostgREST listener threads"

Thought about it but it causes huge overestimation in case of single master and multiple replicas.

> by application name

Is `application_name` reliable enough? If that's the case then the query can be changed to use application name instead of current user.

>and last client statement or so?

I am not sure I understand. Could you elaborate?

wolfgangwalther: > Thought about it but it causes huge overestimation in case of single master and multiple replicas.

Not sure what you mean. How so?

> > and last client statement or so?
> 
> I am not sure I understand. Could you elaborate?

I had thought the last client statement (`query` in `pg_stat_activity`) would be `LISTEN` - but just checked, doesn't seem to be the case. I get the "version query".

mkleczek: > > Thought about it but it causes huge overestimation in case of single master and multiple replicas.
> 
> Not sure what you mean. How so?

A very common setup is to have:
* one master db
* multiple read replicas
* `db-uri` pointing to multiple hosts (master and all replicas) and use `target_session` attribute to let `libpq` load balance connections.

In such setups listener of **all** instances connects to _master_ but pool connections of instances with `target_session=readonly` are load balanced between all replicas. Assuming the same number of instances is connected to each db server we would estimate `(number of replicas + 1)*(per db server instance count)`.
Depending on the number of replicas that could mean that the limit we calculate is higher than we want and not effective.

What's more - upon reception of `reload schema` notification all instances would have to query master for the number of listener connections - so we just moved the thundering herd problem really (but this time we are overloading master, which is worse).
Granted - query for the number of listener sessions is very lightweight so it might not really be too much of a problem, but still.

Taking that into account I think it is better to estimate number of instances using the same connection we use to load the schema cache (as that's really the operation we want to throttle).

I am happy to change `session_user` to `application_name` (or something else) when estimating. But I would stick to the above: estimation should use the same connection as schema cache loading.

EDIT:
I think it is important to remember that the goal is to limit concurrent schema loads executed against _a single database server_ - hence it is not feasible to use a different database server for estimation (which would be the case when we estimate against master but execute schema loads agains replicas).

steve-chavez: > I am happy to change session_user to application_name (or something else) when estimating.

I think `application_name` is unreliable, users can override it to anything

> Also - do you have any ideas how to estimate the size of the PostgREST cluster without this assumption?
> @steve-chavez WDYT?

Since what we really want to identify is a replicated/horizontally-scaled postgREST setup, can't we identify each instance by `current_database()` + exposed schemas? That would rule out the multitenant case where each one has a different schema.

(exposed schemas would be easier if we'd base it on GRANT USAGE as discusssed on https://github.com/PostgREST/postgrest/issues/4364)

steve-chavez: >  That seems overly specific to a certain use-case for something to put into PostgREST.

I've been thinking the same about this issue/PR too, but I don't see any other way around it. 

If we do it I believe it should be a feature not a fix, and frame it something like "throttling schema cache under replicated setups like on AWS ECS".

wolfgangwalther: Are we currently reloading the schema on a different connection than where we listen to notifications?

That seems like a bug in itself to me: If we listen on the primary, but load the schema from the replica, there is no guarantee, that the schema changes we actually want to load are already available on the replica. That means we might actually end up with a stale schema cache, even when the replica catches up afterwards.

I had assumed that not to be the case, was my assumption wrong?

steve-chavez: > Are we currently reloading the schema on a different connection than where we listen to notifications?

Yes, it's exactly as you mention, LISTEN on the primary and reload on the replica. I had the same concern on https://github.com/PostgREST/postgrest/issues/4642#issuecomment-4226604101.

mkleczek: > Yes, it's exactly as you mention, LISTEN on the primary and reload on the replica. I had the same concern on https://github.com/PostgREST/postgrest/issues/4642#issuecomment-4226604101.

Loading from master is no better but in the opposite direction: an instance can have newer schema cache than the actual schema on the replica.

There are no good solutions with current `LISTEN/NOTIFY` based architecture but it is unrelated to this PR - see also my comment here: https://github.com/PostgREST/postgrest/issues/4642#issuecomment-4226736602

I've created https://github.com/PostgREST/postgrest/issues/4842 to discuss it further.

wolfgangwalther: > I've created #4842 to discuss it further.

Since the conclusion in that issue is that "loading the schema cache from the primary is likely the best idea", I suggest we assume for now, that this happens / will be the case.

> it causes huge overestimation in case of single master and multiple replicas.

Since we need to start with something and need to make a compromise somewhere, I suggest we ignore this scenario, where we overestimate the schema reloads *per replica*. We can consider it part of the "schema cache reloads should really happen via the primary instead" issue.

So let's try to find a way to reliably count listener threads against the primary. As pointed out, I believe basing it off of the current role is unreliable. Steve pointed out that basing it off of the `application_name` is potentially unreliable as well.

Any other ideas?

mkleczek: > > I've created #4842 to discuss it further.
> 
> Since the conclusion in that issue is that "loading the schema cache from the primary is likely the best idea", I suggest we assume for now, that this happens / will be the case.

Fixing #4642 (possibly with this PR) **must** be a prerequisite to the above (otherwise thundering herd vulnerability is amplified as all PostgREST instances are going to hit master).

> 
> > it causes huge overestimation in case of single master and multiple replicas.
> 
> Since we need to start with something and need to make a compromise somewhere, I suggest we ignore this scenario, where we overestimate the schema reloads _per replica_. We can consider it part of the "schema cache reloads should really happen via the primary instead" issue.

I think we should rather aim for the rule _throttling must be done in **the same**  database transaction as schema cache loading_ (regardless of _which_ connection is used for querying) - thundering herd issue is really per db server instance
This would mean merging this PR (as it implements this strategy) and in subsequent PR changing the connection used for schema cache querying.

These issues are really orthogonal: one is about "what should be the source of schema cache data" and another is about "making sure the source of schema cache data, whatever it is, is not overloaded by concurrent requests".

> 
> So let's try to find a way to reliably count listener threads against the primary. As pointed out, I believe basing it off of the current role is unreliable. Steve pointed out that basing it off of the `application_name` is potentially unreliable as well.
> 
> Any other ideas?

There is also another dimension here: it is safer to _underestimate_ than to _overestimate_ the number of PostgREST instances.
Basing the calculation off the `current_role` can only underestimate instances (ie. there can be other instances connected using different user but they won't be counted towards the limit).
Since all instances, regardless of `session_user` use the same db locks for throttling, it means that it fixes the thundering herd problem in all scenarios (it is just that the limit might be lower than optimal in the scenario @wolfgangwalther described).

So IMHO it is safe to merge this PR as is and we can change the strategy to be more liberal (ie. not underestimating) in the future.

@steve-chavez @wolfgangwalther WDYT?

steve-chavez: > There is also another dimension here: it is safer to underestimate than to overestimate the number of PostgREST instances.

I think we should first solve the problem of counting the postgREST instances accurately (with https://github.com/PostgREST/postgrest/issues/4864 or https://github.com/PostgREST/postgrest/issues/4863)

wolfgangwalther: > Basing the calculation off the `current_role` can only underestimate instances (ie. there can be other instances connected using different user but they won't be counted towards the limit).

Each individual number of instances might be understimated, but the total number of available slots for concurrent reload will be *overestimated*, because instances with different session user will *not* lock each other.

> I think we should first solve the problem of counting the postgREST instances accurately (with [#4864](https://github.com/PostgREST/postgrest/issues/4864) or [#4863](https://github.com/PostgREST/postgrest/issues/4863))

I'll mark this as draft to make it clear we're waiting for one of these.

steve-chavez: Also it's possible that with https://github.com/PostgREST/postgrest/pull/5031 we won't need this PR anymore. Michal had some further ideas about reusing the [admin schema_cache](https://docs.postgrest.org/en/v14/references/admin_server.html#runtime-schema-cache) to initialize other instances (discussed elsewhere), let's continue on #5031.

mkleczek: > Also it's possible that with #5031 we won't need this PR anymore. Michal had some further ideas about reusing the [admin schema_cache](https://docs.postgrest.org/en/v14/references/admin_server.html#runtime-schema-cache) to initialize other instances (discussed elsewhere), let's continue on #5031.

I think they are orthogonal. This one is about reloading the schema cache after receiving listener notification. The other one is about startup.

mkleczek: > Each individual number of instances might be understimated, but the total number of available slots for concurrent reload will be overestimated, because instances with different session user will not lock each other.

They will, we always use the same lock here: https://github.com/mkleczek/postgrest/blob/49a83c1e68a07dde80dc6f8059c54db3e4aca0bd/src/PostgREST/AppState.hs#L409
Underestimating number of active sessions means a node will try locking fewer slots - number of slots is `log(estimated session count)`.

>> I think we should first solve the problem of counting the postgREST instances accurately (with https://github.com/PostgREST/postgrest/issues/4864 or https://github.com/PostgREST/postgrest/issues/4863)

> I'll mark this as draft to make it clear we're waiting for one of these.

Fine for me

wolfgangwalther: > They will, we always use the same lock

Ah, right, my bad. I didn't check the code and somehow assumed the lock would be based on the criteria used for including instances in the count.

In that case I have no objection to *temporarily* count based on the session user, although with https://github.com/PostgREST/postgrest/issues/4863#issuecomment-4816361859 I think we'd be close to be able to count them "properly" as well.

steve-chavez:
With https://github.com/PostgREST/postgrest/pull/5031, maybe we don't have to do this and instead have a mode to trust a master `schema-cache-uri`, which would make the replicas not query for the schema cache.

wolfgangwalther:
> With #5031, maybe we don't have to do this and instead have a mode to trust a master `schema-cache-uri`, which would make the replicas not query for the schema cache.

I think these issues are orthogonal. You can try to reduce the load on the database server by reusing the schema cache (not a fan of all of that, btw - this is just *another* layer on the cache we already have, adding even more complexity...), but you must still guard against the thundering herd which can still happen certain deployment scenarios.

steve-chavez:
@mkleczek Is the above true? I was hoping https://github.com/PostgREST/postgrest/pull/5031 would allow us to side step this fix altogether (it is still an edge case that most users won't hit). Maybe with a mode that only loads schema cache from file/uri and doesn't do db queries?

> You can try to reduce the load on the database server by reusing the schema cache (not a fan of all of that, btw - this is just another layer on the cache we already have, adding even more complexity...),

I'll say that one of the main motivations for #5031 was https://github.com/PostgREST/postgrest/issues/5100 (that is guaranteeing fast starts) but that can be solved in another way. Ideally we make #5031 solve multiple "at scale" problems, otherwise the complexity might not be worth the pay off.

PR #4645, opened 2026-02-11
fix: Flush pool as late as possible during schema cache reloading

mkleczek:
**DISCLAIMER:**

`retryingSchemaCacheLoad` flushes the pool upon every retry before it starts reloading the schema. This is too early as schema reloading might take some time during which new connections might be acquired. The consequence is that:
* upon successful schema cache reload we might have some connections created with the old schema cache
* we close connections upon each retry and under load we will keep closing and re-opening connections until schema cache load succeeds

This change is to make sure we flush the pool only after successful schema cache querying but before loading (so that connections acquired during loading wait for it and do not interfere with timing the loading process).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
-- Pool should be flushed not before but after schema cache reloading by @mkleczek in #4645
+- Fix unnecessary connection pool flushes during schema cache reloading by @mkleczek in #4645
 
 ### Changed
 
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -351,7 +351,7 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
           markSchemaCachePending appState
           putSchemaCache appState $ Just sCache
           -- Flush the pool after loading the schema cache to reset any stale session cache entries
-          -- We do it after successfully querying the schema cache
+          -- We do it after successfully querying the schema cache (because this can fail and during retries we would flush the pool repeatedly unnecessarily)
           -- and after marking sCacheStatus as pending,
           flushPool appState
           observer $ SchemaCacheQueriedObs resultTime


steve-chavez:
The change seems reasonable. Is there a way to test the previous behavior and that this is an improvement?

I'm thinking we have the `schemaCacheLoads` metric but perhaps we should also have one metric for pool flushes? Then we might be able to test that this change reduces the flushes.
https://github.com/PostgREST/postgrest/blob/99984d328e04735b8e745ff4deac72ef2ab3c848/src/PostgREST/Metrics.hs#L22-L33

---

Related, just noted these two functions do the same perhaps we should deduplicate:

https://github.com/PostgREST/postgrest/blob/99984d328e04735b8e745ff4deac72ef2ab3c848/src/PostgREST/AppState.hs#L271-L278

mkleczek:
> The change seems reasonable. Is there a way to test the previous behavior and that this is an improvement?
> 
> I'm thinking we have the `schemaCacheLoads` metric but perhaps we should also have one metric for pool flushes? Then we might be able to test that this change reduces the flushes.
> 

Done:
- Added `pgrst_db_pool_flushes_total` metric.
- Added HSpec module dedicated to metrics tests and a test to verify the new metric (I think it is a good idea to have one - we already have JWT cache spec tests that make use of metric state)
- Added io tests to validate pool flushing behavior (both in normal conditions and with schema loading retries)

> 
> Related, just noted these two functions do the same perhaps we should deduplicate:
> 

After latest changes they are no longer duplicate as `destroyPool` does not emit observations.

steve-chavez:
> Added pgrst_db_pool_flushes_total metric.
Added HSpec module dedicated to metrics tests and a test to verify the new metric (I think it is a good idea to have one - we already have JWT cache spec tests that make use of metric state)
Added io tests to validate pool flushing behavior (both in normal conditions and with schema loading retries)

@mkleczek Could you split those commits into another PR? :pray: (without the fix here) That would make it easier to review.

mkleczek:
> > Added pgrst_db_pool_flushes_total metric.
> > Added HSpec module dedicated to metrics tests and a test to verify the new metric (I think it is a good idea to have one - we already have JWT cache spec tests that make use of metric state)
> > Added io tests to validate pool flushing behavior (both in normal conditions and with schema loading retries)
> 
> @mkleczek Could you split those commits into another PR? 🙏 (without the fix here) That would make it easier to review.

@steve-chavez Done. Stacked this PR on top of https://github.com/PostgREST/postgrest/pull/4658

mkleczek:
> > @mkleczek Could you split those commits into another PR? 🙏 (without the fix here) That would make it easier to review.
> 
> @steve-chavez Done. Stacked this PR on top of #4658

@steve-chavez Rebased on https://github.com/PostgREST/postgrest/pull/4668

steve-chavez:
Needs rebasing now

taimoorzaeem:
Is this a bug fix? If so, then it may also need a `CHANGELOG.md` entry.

steve-chavez:
@mkleczek Could you please address the feedback [here](https://github.com/PostgREST/postgrest/pull/4645#issuecomment-4216107005)?

mkleczek:
> @mkleczek Could you please address the feedback [here](https://github.com/PostgREST/postgrest/pull/4645#issuecomment-4216107005)?

Done

src/PostgREST/AppState.hs
```diff
@@ -352,6 +350,10 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
           -- IORef on putSchemaCache. This is why schema cache status is marked as pending here to signal the Admin server (using isPending) that we're on a recovery state.
           markSchemaCachePending appState
           putSchemaCache appState $ Just sCache
+          -- Flush the pool after loading the schema cache to reset any stale session cache entries
+          -- We do it after successfully querying the schema cache
+          -- and after marking sCacheStatus as pending,
```
steve-chavez, line 355: Adding important fact mentioned on https://github.com/PostgREST/postgrest/pull/4645#issue-3926493680

```suggestion
          -- Flush the pool after loading the schema cache to reset any stale session cache entries
          -- We do it after successfully querying the schema cache (because this can fail and during retries we would flush the pool repeatedly unnecessarily)
          -- and after marking sCacheStatus as pending,
```

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ All notable changes to this project will be documented in this file. From versio
 - Add config `db-timezone-enabled` for optional querying of timezones by @taimoorzaeem in #4751
 - Log when the pool is released during schema cache reload on `log-level=debug` by @mkleczek in #4668
 
+### Fixed
+
+- Pool should be flushed not before but after schema cache reloading by @mkleczek in #4645
```
steve-chavez: More user facing:

```suggestion
- Fix unnecessary connection pool flushes during schema cache reloading by @mkleczek in #4645
```

test/observability/Observation/SchemaCacheSpec.hs
```diff
@@ -27,7 +27,7 @@ spec = describe "Server started with metrics enabled" $ do
       waitFor (1 * sec) "SchemaCacheLoadedObs" $ \x -> [ o | o@SchemaCacheLoadedObs{} <- pure x ]
 
 
-  it "Should flush pool multiple times when schema reloading retries" $ do
+  it "Should flush pool once when schema reloading retries" $ do
```
steve-chavez, line 30: This is great, clearly proves the fix 👍

taimoorzaeem:
If we decide to backport this, we'd need to first backport 5d4f82d70fdcf402b9c02554fbbbc8438a2c4a77 from #4668 (already applied ad907219beb639243774a224ed87ca93bf629009).

steve-chavez:
@taimoorzaeem Yes, those should be fine to backport. This one too.

PR #4647, opened 2026-02-11
fix: don't hide async exceptions in logs

### LINKED ISSUE #4646: Logs are not showing important `AsyncExceptions`
## Problem

This can be reproduced with:

```
$ PGRST_DB_SCHEMAS="apflora" LOG_LEVEL=debug postgrest-with-pg-14 -f test/io/fixtures/big_schema.sql postgrest-run
11/Feb/2026:09:08:23 -0500: Starting PostgREST 15 (pre-release)...
11/Feb/2026:09:08:23 -0500: Schema cache loaded in 252.2 milliseconds

$ curl localhost:3000/

-- No Logs are produced
```

Here a `stack overflow` is happening, yet we don't log it

This is because we use the `defaultsSettings` from Warp:

https://github.com/PostgREST/postgrest/blob/55cd5a23ff52bc688674d5c3b1c0ef751994ec13/src/PostgREST/App.hs#L84-L86

And those have [defaultOnException](https://github.com/yesodweb/wai//blob/8c3882c60f6abe043889fc20c7efd3fa9747fa4a/warp/Network/Wai/Handler/Warp/Settings.hs#L203-L208):

```hs
-- | The default settings for the Warp server. See the individual settings for
-- the default value.
defaultSettings :: Settings
defaultSettings =
    Settings
        { settingsPort = 3000
        , settingsHost = "*4"
        , settingsOnException = defaultOnException
```
Which does [this](https://github.com/yesodweb/wai//blob/8c3882c60f6abe043889fc20c7efd3fa9747fa4a/warp/Network/Wai/Handler/Warp/Settings.hs#L251-L269):

```hs
-- | Apply the logic provided by 'defaultOnException' to determine if an
-- exception should be shown or not. The goal is to hide exceptions which occur
-- under the normal course of the web server running.
--
-- Since 2.1.3
defaultShouldDisplayException :: SomeException -> Bool
defaultShouldDisplayException se
    | Just (_ :: InvalidRequest) <- fromException se = False
    | Just (ioeGetErrorType -> et) <- fromException se
    , et == ResourceVanished || et == InvalidArgument =
        False
    | isAsyncException se = False
    | otherwise = True

-- | Printing an exception to standard error
--   if `defaultShouldDisplayException` returns `True`.
--
-- Since: 3.1.0
defaultOnException :: Maybe Request -> SomeException -> IO ()
defaultOnException _ e =
    when (defaultShouldDisplayException e) $
        TIO.hPutStrLn stderr $
            T.pack $
                show e
```

The ` | isAsyncException se = False` part is hiding all the async exceptions, which look quite important to log.

```hs
-- |Asynchronous exceptions.
data AsyncException
  = StackOverflow
        -- ^The current thread\'s stack exceeded its limit.
        -- Since an exception has been raised, the thread\'s stack
        -- will certainly be below its limit again, but the
        -- programmer should take remedial action
        -- immediately.
  | HeapOverflow
        -- ^The program\'s heap is reaching its limit, and
        -- the program should take action to reduce the amount of
        -- live data it has. Notes:
        --
        --   * It is undefined which thread receives this exception.
        --     GHC currently throws this to the same thread that
        --     receives 'UserInterrupt', but this may change in the
        --     future.
        --
        --   * The GHC RTS currently can only recover from heap overflow
        --     if it detects that an explicit memory limit (set via RTS flags).
        --     has been exceeded.  Currently, failure to allocate memory from
        --     the operating system results in immediate termination of the
        --     program.
  | ThreadKilled
        -- ^This exception is raised by another thread
        -- calling 'Control.Concurrent.killThread', or by the system
        -- if it needs to terminate the thread for some
        -- reason.
  | UserInterrupt
        -- ^This exception is raised by default in the main thread of
        -- the program when the user requests to terminate the program
        -- via the usual mechanism(s) (e.g. Control-C in the console).
  deriving ( Eq  -- ^ @since 4.2.0.0
           , Ord -- ^ @since 4.2.0.0
           )
```

## Notes

Also note that  `TIO.hPutStrLn stderr` doesn't follow our apache log format, we should change that.

steve-chavez:
Fixes #4646. Using the repro on #4646, this now produces the log:

```
11/Feb/2026:09:40:08 -0500: Warp server error: stack overflow
```

When:
```
$ curl localhost:3000/
```

### TODO

- [x] Filtering exceptions, some that are safely ignorable and others that could potentially cause flooding.
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
+- Don't hide async exceptions in logs by @stevechavez in #4646
+
 ### Changed
 
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -9,18 +9,24 @@ Some of its functionality includes:
 - Producing HTTP Headers according to RFCs.
 - Content Negotiation
 -}
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ViewPatterns        #-}
 module PostgREST.App
   ( postgrest
   , run
   ) where
 
 
+import GHC.IO.Exception (IOErrorType (..))
+import System.IO.Error  (ioeGetErrorType)
+
 import Control.Monad.Except     (liftEither)
 import Data.Either.Combinators  (mapLeft, whenLeft)
 import Data.Maybe               (fromJust)
 import Data.String              (IsString (..))
-import Network.Wai.Handler.Warp (defaultSettings, setHost, setPort,
+import Network.Wai.Handler.Warp (defaultSettings, setHost,
+                                 setOnException, setPort,
                                  setServerName)
 
 import qualified Data.Text.Encoding       as T
@@ -63,7 +69,6 @@ type Handler = ExceptT Error
 
 run :: AppState -> IO ()
 run appState = do
-  let observer = AppState.getObserver appState
   conf@AppConfig{..} <- AppState.getConfig appState
 
   AppState.schemaCacheLoader appState -- Loads the initial SchemaCache
@@ -79,7 +84,24 @@ run appState = do
     address <- resolveSocketToAddress (AppState.getSocketREST appState)
     observer $ AppServerAddressObs address
 
-  Warp.runSettingsSocket (serverSettings conf) (AppState.getSocketREST appState) app
+  Warp.runSettingsSocket (serverSettings conf & setOnException onWarpException) (AppState.getSocketREST appState) app
+  where
+    observer = AppState.getObserver appState
+
+    onWarpException :: Maybe Wai.Request -> SomeException -> IO ()
+    onWarpException _ ex =
+      when (shouldDisplayException ex) $
+        observer $ WarpErrorObs $ show ex
+
+    -- Similar to wai defaultShouldDisplayException in
+    -- https://github.com/yesodweb/wai//blob/8c3882c60f6abe043889fc20c7efd3fa9747fa4a/warp/Network/Wai/Handler/Warp/Settings.hs#L251-L258
+    -- but without omitting AsyncException since it's important to log for ThreadKilled, StackOverflow and other cases.
+    -- We want to reuse this to avoid flooding the logs for some transient failure cases.
+    shouldDisplayException :: SomeException -> Bool
+    shouldDisplayException se
+        | Just (_ :: Warp.InvalidRequest) <- fromException se = False
+        | Just (ioeGetErrorType -> et) <- fromException se, et == ResourceVanished || et == InvalidArgument = False
+        | otherwise = True
 
 serverSettings :: AppConfig -> Warp.Settings
 serverSettings AppConfig{..} =
--- a/test/io/test_big_schema.py
+++ b/test/io/test_big_schema.py
@@ -3,6 +3,7 @@
 import re
 
 import pytest
+import requests
 
 from postgrest import run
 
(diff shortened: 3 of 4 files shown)


steve-chavez:
Tests are failing because we have new logs:

```
11/Feb/2026:14:55:01 +0000: Warp server error: Warp: Client closed connection prematurely
11/Feb/2026:14:55:01 +0000: Warp server error: Warp: Client closed connection prematurely
11/Feb/2026:14:55:01 +0000: Schema cache queried in 579.2 milliseconds
```

Maybe we should still surface these for `log-level=debug`. Also remove the unnecessary prefixes.

**Edit**: did not log these for now, as they're transient failures.

src/PostgREST/App.hs
```diff
@@ -71,22 +78,38 @@ run appState = do
...
+
+-- Similar to wai defaultShouldDisplayException in
+-- https://github.com/yesodweb/wai//blob/8c3882c60f6abe043889fc20c7efd3fa9747fa4a/warp/Network/Wai/Handler/Warp/Settings.hs#L251-L258
+-- but without omitting AsyncException.
+-- We want to reuse this to avoid flooding the logs for some transient failure cases.
+shouldDisplayException :: SomeException -> Bool
+shouldDisplayException se
+    | Just (_ :: Warp.InvalidRequest) <- fromException se = False
+    | Just (ioeGetErrorType -> et) <- fromException se, et == ResourceVanished || et == InvalidArgument = False
+    | otherwise = True
```
steve-chavez: Decided to only log AsyncExceptions to minimize risk of flooding logs.

test/io/test_big_schema.py
```diff
@@ -54,6 +55,27 @@ def test_openapi_in_big_schema(defaultenv):
...
+
+    with run(env=env, wait_max_seconds=30, no_startup_stdout=False) as postgrest:
+        with pytest.raises(requests.exceptions.ConnectionError):
+            postgrest.session.get("/")
+
+        output = postgrest.read_stdout(nlines=10)
+        output.extend(postgrest.read_stdout(nlines=10))
+
+        assert any(
+            "Warp server error: stack overflow" in line for line in output
```
steve-chavez: Tested this case as it is the simplest one.

src/PostgREST/App.hs
```diff
@@ -71,22 +78,38 @@ run appState = do
...
+
+-- Similar to wai defaultShouldDisplayException in
+-- https://github.com/yesodweb/wai//blob/8c3882c60f6abe043889fc20c7efd3fa9747fa4a/warp/Network/Wai/Handler/Warp/Settings.hs#L251-L258
+-- but without omitting AsyncException since it's important to log for ThreadKilled, StackOverflow and other cases.
+-- We want to reuse this to avoid flooding the logs for some transient failure cases.
+shouldDisplayException :: SomeException -> Bool
+shouldDisplayException se
+    | Just (_ :: Warp.InvalidRequest) <- fromException se = False
+    | Just (ioeGetErrorType -> et) <- fromException se, et == ResourceVanished || et == InvalidArgument = False
+    | otherwise = True
```
steve-chavez: These two lines are showing as partially covered in codecov, it's too hard to test them for now.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ All notable changes to this project will be documented in this file. From versio
 
 ### Fixed
 
+- Don't hide async exceptions in logs by @stevechavez in #4646
```
steve-chavez, line 18: Considering this a fix as we were hiding logs before. Also, otherwise it's impossible to debug some failure cases behind proxy as mentioned on https://github.com/PostgREST/postgrest/issues/4646#issuecomment-3884819850

src/PostgREST/App.hs
```diff
@@ -71,22 +78,38 @@ run appState = do
 
   Listener.runListener appState
 
-  Admin.runAdmin appState (serverSettings conf)
+  Admin.runAdmin appState (serverSettings conf observer)
```
steve-chavez: This line is also partially covered, I think this can be omitted as we don't need the exception logging in the Admin server.

steve-chavez: Done.

laurenceisla (approved):

PR #4653, opened 2026-02-13
refactor: Remove unnecessary parameter in `PoolAcqTimeoutObs` and lazy initialization of stateLogDebouncePoolTimeout

mkleczek:
Small clean-up of PoolAcqTimeoutObs handling in two commits:

1. PoolAcqTimeoutObs is always created with SQL.AcquisitionTimeoutUsageError so the parameter does not provide any useful information. On the other hand, it complicates debouncing logic in Logger as logWithDebounce has to lazily create the debouncer (which is error prone and hence the logic has a race condition).
2. stateLogDebouncePoolTimeout is an MVar initialized on the first logging of PoolAcqTimeoutObs. The code in logWithDebounce has race condition that could lead to creation of multiple debouncers. This change simplifies logic by getting rid of lazy initialization of debouncer.

### DIFF
--- a/src/PostgREST/Logger.hs
+++ b/src/PostgREST/Logger.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE RecursiveDo     #-}
 {-|
 Module      : PostgREST.Logger
 Description : Logging based on the Observation.hs module. Access logs get sent to stdout and server diagnostic get sent to stderr.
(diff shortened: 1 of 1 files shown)


src/PostgREST/Observation.hs
```diff
@@ -140,8 +140,8 @@ observationMessage = \case
      "Config reloaded"
   PoolInit poolSize ->
      "Connection Pool initialized with a maximum size of " <> show poolSize <> " connections"
-  PoolAcqTimeoutObs usageErr ->
-    jsonMessage usageErr
+  PoolAcqTimeoutObs ->
+     "Timed out acquiring connection from connection pool."
```
steve-chavez: Looks like this is a duplicate of:

https://github.com/PostgREST/postgrest/blob/c5d495006a0a9118273ec810e21f428fd0d22f5b/src/PostgREST/Error.hs#L532

mkleczek: Good catch, thx!

Fixed.

PR #4658, opened 2026-02-17
add(metrics): expose db pool flushes counter

mkleczek:
Emit a dedicated PoolFlushed observation when the DB pool is released during schema cache reload and expose it as `pgrst_db_pool_flushes_total`.

Stacked on top of https://github.com/PostgREST/postgrest/pull/4659 to make tests more readable

### DIFF
--- a/docs/references/observability.rst
+++ b/docs/references/observability.rst
@@ -177,6 +177,16 @@ pgrst_db_pool_timeouts_total
 
 The total number of pool connection timeouts.
 
+pgrst_db_pool_flushes_total
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+======== =======
+**Type** Counter
+======== =======
+
+The total number of times the pool was flushed.
+The pool is flushed after successful schema cache reload.
+
 pgrst_db_pool_available
 ~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -270,10 +270,14 @@ usePool AppState{stateObserver=observer, stateMainThreadId=mainThreadId, ..} ses
 
 -- | Flush the connection pool so that any future use of the pool will
 -- use connections freshly established after this call.
+-- | Emits PoolFlushed observation
 flushPool :: AppState -> IO ()
-flushPool AppState{..} = SQL.release statePool
+flushPool AppState{..} = do
+  SQL.release statePool
+  stateObserver PoolFlushed
 
 -- | Destroy the pool on shutdown.
+-- | Differs from flushPool in not emiting PoolFlushed observation.
 destroyPool :: AppState -> IO ()
 destroyPool AppState{..} = SQL.release statePool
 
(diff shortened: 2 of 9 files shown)


src/PostgREST/AppState.hs
```diff
@@ -271,7 +271,9 @@ usePool AppState{stateObserver=observer, stateMainThreadId=mainThreadId, ..} ses
 -- | Flush the connection pool so that any future use of the pool will
 -- use connections freshly established after this call.
 flushPool :: AppState -> IO ()
-flushPool AppState{..} = SQL.release statePool
+flushPool AppState{..} = do
+  SQL.release statePool
+  stateObserver PoolFlushed
```
steve-chavez, line 277: Should we also add this on 
https://github.com/PostgREST/postgrest/blob/c5d495006a0a9118273ec810e21f428fd0d22f5b/src/PostgREST/AppState.hs#L277-L278

We also discussed about merging both flushPool and destroyPool into a single function.

docs/references/observability.rst
```diff
@@ -177,6 +177,15 @@ pgrst_db_pool_timeouts_total
...
 The total number of pool connection timeouts.
 
+pgrst_db_pool_flushes_total
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+======== =======
+**Type** Counter
+======== =======
+
+The total number of times the pool was flushed.
```
steve-chavez, line 187: Maybe we should mention this means releasing all the connections in the pool, which always happens during schema cache load?

steve-chavez: >  which always happens during schema cache load?

Hm, actually I'm not sure if the above will be true once  https://github.com/PostgREST/postgrest/pull/4645 is merged

src/PostgREST/AppState.hs
mkleczek: > Should we also add this on
> 
> https://github.com/PostgREST/postgrest/blob/c5d495006a0a9118273ec810e21f428fd0d22f5b/src/PostgREST/AppState.hs#L277-L278
> 
> We also discussed about merging both flushPool and destroyPool into a single function.

@steve-chavez Are they really the same any more though? Do we want to generate observation during shutdown?

steve-chavez: > Are they really the same any more though? Do we want to generate observation during shutdown?

You're right. I think we should comment why they're both different functions.

mkleczek: Done

docs/references/observability.rst
mkleczek: > > which always happens during schema cache load?
> 
> Hm, actually I'm not sure if the above will be true once #4645 is merged

@steve-chavez Added a short sentence to the documentation.

steve-chavez:
One question, would this be generally useful? I mean besides for testing https://github.com/PostgREST/postgrest/pull/4645.

I haven't seen this metric on [OTel db metrics](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/#connection-pools) although I have seen [numDestroys in some other place](https://techdocs.broadcom.com/us/en/ca-enterprise-software/it-operations-management/application-performance-management/10-8/using/apm-metrics/resource-metrics/connection-pool-metrics.html#concept.dita_0f49c8c82f677a75ca4e0542e5d0187a122d2abe_ConnectionPoolMetrics#concept.dita_0f49c8c82f677a75ca4e0542e5d0187a122d2abe_ConnectionPoolCountMetrics) (not sure if these are Prom metrics).

steve-chavez:
If not generally useful, maybe this metric can be private and not exposed to users? That way we can use it for tests only.

mkleczek:
> If not generally useful, maybe this metric can be private and not exposed to users? That way we can use it for tests only.

@steve-chavez I agree. There is no need for this metric, only observation.
Closing this PR and raising https://github.com/PostgREST/postgrest/pull/4668 instead.

PR #4661, opened 2026-02-18
refactor: move error data types to `Error/Types.hs`

taimoorzaeem:
The `Error.hs` module is getting bloated. This moves types to another module for better separation of concerns.

Towards #4088.

### DIFF
--- a/src/PostgREST/Error/Types.hs
+++ b/src/PostgREST/Error/Types.hs
@@ -29,7 +29,7 @@ import PostgREST.SchemaCache.Routine      (Routine (..))
 import Protolude
 
 data Error
-  = ApiRequestError ApiRequestError
+  = ApiRequestErr ApiRequestError
   | SchemaCacheErr SchemaCacheError
   | JwtErr JwtError
   | NoSchemaCacheError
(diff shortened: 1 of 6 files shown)


src/PostgREST/Error/Types.hs
```diff
@@ -0,0 +1,139 @@
...
+data SchemaCacheError
+  = AmbiguousRelBetween Text Text [Relationship]
+  | AmbiguousRpc [Routine]
+  | NoRelBetween Text Text (Maybe Text) Text RelationshipsMap
+  | NoRpc Text Text [Text] MediaType Bool [QualifiedIdentifier] [Routine]
+  | ColumnNotFound Text Text
+  | TableNotFound Text Text SchemaCache
+  deriving Show
+
+-- JWT ERRORS: PGRST3XX
```
steve-chavez, line 92: Looks good. I was wondering where are the `PGRSTX00` and the `PGRST00X`?

taimoorzaeem: > I was wondering where are the `PGRSTX00` and the `PGRST00X`?

Those are defined _directly_ on error data types coming from `hasql`. For example:

https://github.com/PostgREST/postgrest/blob/b977ffbb9010cb10f5ba0958b9afba5602c085bf/src/PostgREST/Error.hs#L552-L554

Maybe it would be good to wrap them in our own types in a future refactor to keep things neat.

PR #4664, opened 2026-02-20
refactor: remove commented out code from `Auth/Jwt.hs`

taimoorzaeem:
Some commented out code got left in the codebase in 5196823365c92156816183f02a0fad582bb98850. 

Mentioned in https://github.com/PostgREST/postgrest/pull/4121#discussion_r2831889366.


steve-chavez:
@taimoorzaeem Can we fail on CI on commented out code? Perhaps `hlint` can do this?

taimoorzaeem:
> Can we fail on CI on commented out code? Perhaps `hlint` can do this?

Not sure. I think that would be very challenging to achieve given that we also have `doctests` which is valid _commented out code_.

Maybe something custom can be done using [haskell-src-exts](https://hackage.haskell.org/package/haskell-src-exts). I'll look into it in the future. :)

Issue #4172, opened 2025-07-02
Sanitizing PL/Python Error Messages


DutchGerman:
## Problem

When throwing an error in a PL/Python function, the error message exposes some information about the database by returning `plpy.Error: <message>` as the `message` 

Let's say we have the following function, which is triggered via an API request.

```sql
CREATE FUNCTION raise_custom_exception()
RETURNS void
LANGUAGE plpython3u AS $$
  plpy.error(
      "Method not allowed",
      sqlstate="PT405",
      detail="Because of XYZ this method is not allowed",
      hint="Change XYZ"
  )
$$;
```

The SQL exception raised is:

```
SQL Error [PT405]: ERROR: plpy.Error: This is a custom exception
  Detail: Because of XYZ this message is not allowed
  Hint: Change XYZ
  Where: Traceback (most recent call last):
  PL/Python function "raise_custom_exception", line 2, in <module>
    plpy.error(
PL/Python function "raise_custom_exception"
```

The error message thrown in the client is:

```json
{
    "code": "PT405",
    "details": "Because of XYZ this method is not allowed",
    "hint": "Change XYZ",
    "message": "plpy.Error: Method not allowed"
}
```

## Solution

It would be nice to have a way to strip the `plpy.Error` from the `message` we return to the client.

wolfgangwalther:
Could be related: #2706

DutchGerman:
If I understand [your idea](https://github.com/PostgREST/postgrest/issues/2706#issuecomment-1484037839) correctly, `db-error-handler` would be a setting like `db_pre_request`, that takes in a function that can mutate the error message ?

In that case yes, very related.

steve-chavez:
> It would be nice to have a way to strip the plpy.Error from the message we return to the client.

If NULLifying the message is acceptable then https://github.com/PostgREST/postgrest/issues/4088 might be a simpler solution.

Issue #2706, opened 2023-03-16
Customizing error messages on failed check constraints


steve-chavez:
### Problem

Having:

```sql
CREATE DOMAIN public.filled_text AS text
	CONSTRAINT not_empty CHECK ((NOT (VALUE ~ '^\s*$'::text)));

COMMENT ON DOMAIN public.filled_text IS 'Text value that cannot be empty or whitespace, but doesn''t assume not null.';
```

A failed POST request will return:

```json
{"code":"23514","details":null,"hint":null,"message":"value for domain filled_text violates check constraint \"not_empty\""}
```

Which is not user friendly. We've been telling users to format the error message on the client side, but this is not possible with direct consumers of the API(unless a proxy is involved, which complicates the setup).

### Proposal

Just to kickstart discussion. We could have a `pre-request` function that does:

```sql
create or replace function custom_errors() returns void as $$
declare
  error_message text := current_setting('response.error', true)::json->>'message';
  error_code text := current_setting('response.error', true)::json->>'code';
begin
  if error_code = 23514 and error_message like '%not_empty%'
    then set_config(`response.error`, json_build_object('message', 'cannot be empty'); -- one problem here is that there's no way to know the column
  end if;
end; $$ language plpgsql;
```

mkleczek:
Maybe https://datatracker.ietf.org/doc/rfc7807/ is relevant?

steve-chavez:
Hm, how so? RFC 7807 is for the error message format right? We discussed adopting that on https://github.com/PostgREST/postgrest/pull/1926#issuecomment-916591592

steve-chavez:
On https://github.com/PostgREST/postgrest/issues/3084#issue-2022449376, @dvv mentioned:

> I wonder if we could have a post-request db hook in addition to pre-request one?
The rationale is to sanitize errors or to reconcile just in-place.
It could be given all available info from get stacked diagnostics and raise more meaningful messages or just rewrite the json data.

Hm, I don't think we can use [get stacked diagnostics](https://www.postgresql.org/docs/current/plpgsql-control-structures.html#PLPGSQL-EXCEPTION-DIAGNOSTICS) since that would imply `plpgsql`?

dvv:
@steve-chavez
Right, catching the exception directly implies plpgsql `begin/exception`.

[...]

PR #4667, opened 2026-02-24
add: config `client-error-verbosity` to set error verbosity

### LINKED ISSUE #3824: Show only exposed schemas to which the user has access in error message

### LINKED ISSUE #3980: Information disclosure in RPC function error handling for inaccessible functions in exposedschemas
## Problem

When a user attempts to call RPC functions in an exposed schema they don't have access to, PostgREST sometimes exposes information about the underlying schema through error messages. This creates an inconsistent security boundary and potential information leakage.

### Current Behavior:

1. When calling a slightly misspelled function (`admin.custom_function2`):

```
error: {
  code: 'PGRST202',
  details: 'Searched for the function admin.custom_function2 without parameters or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache.',
  hint: 'Perhaps you meant to call the function admin.custom_function',
  message: 'Could not find the function admin.custom_function2 without parameters in the schema cache'
}
```

2. When calling a completely non-existent function (`admin.missing_function`):

```
error: {
  code: 'PGRST202',
  details: 'Searched for the function admin.missing_functionwithout parameters or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache.',
  hint: null,
  message: 'Could not find the function admin.missing_function without parameters in the schema cache'
}
```

3. When calling a function with the wrong parameters (requires 1 parameter but none passed) :

```
  error: {
    code: 'PGRST202',
    details: 'Searched for the function admin.function_with_params without parameters or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache.',
    hint: null,
    message: 'Could not find the function admin.function_with_params without parameters in the schema cache'
  }
```

4. When calling a function with the correct name (`admin.custom_function`):
```
error: {
  code: '42501',
  details: null,
  hint: null,
  message: 'permission denied for schema admin'
}
```

## Expected Behavior

For security and consistency, PostgREST should return the same permission denied error message in all cases where a user attempts to access a schema they don't have permission for, regardless of whether the function exists or not:
```
error: {
  code: '42501',
  details: null,
  hint: null,
  message: 'permission denied for schema admin'
}
```

This is consistent with the behavior when querying tables or views in protected schemas.

## Related Issue
This issue is related to https://github.com/PostgREST/postgrest/issues/3824

## Additional notes
When trying to access an object (view/table/function) to which no GRANT has been given for the current role in an exposed schema (e.g., `public`), PostgreSQL returns error messages like `permission denied for view xxxx` that may expose details about object existence, function parameters or table columns. This behavior appears to be expected when schema usage privileges are granted but specific object privileges are restricted.

For more granular access control and to prevent information disclosure between roles, it seems the recommended approach is to organize database objects into multiple schemas and only grant `USAGE` privileges to each role for schemas they need to access. This architecture would likely ensure more complete information hiding between roles with different access requirements.

### LINKED ISSUE #4088: Support `db-error-verbosity` config to customize error verbosity

taimoorzaeem:
Set error verbosity using this config. The verbosity can be set to `verbose` or `minimal` for client error responses.

This only affects client side HTTP responses, server side logs are not affected by this config.

Closes #4088, closes #3980 and closes #3824. Other related issues: #2706, #4172.


steve-chavez:
@taimoorzaeem I think this one should close https://github.com/PostgREST/postgrest/issues/4088, https://github.com/PostgREST/postgrest/issues/3980 and https://github.com/PostgREST/postgrest/issues/3824.

src/PostgREST/Error.hs
```diff
@@ -57,26 +58,30 @@ import PostgREST.Error.Types
...
+    toJsonPgrstError Verbose err = JSON.object [
         "code"    .= code err
       , "message" .= message err
       , "details" .= details err
       , "hint"    .= hint err
       ]
+    toJsonPgrstError Minimal err = JSON.object [
+        "code"    .= code err
+      , "message" .= message err
+      ]
```
steve-chavez, line 74: Nice, really this is the gist of the change.

steve-chavez (approved):
The PR is just perfect 🥇

steve-chavez:
@taimoorzaeem Just noticed the docs are not in https://docs.postgrest.org/en/latest/references/errors.html, IMO this is first place anyone would look for.

I think we need a subsection at the end of https://docs.postgrest.org/en/latest/references/errors.html called "Client error verbosity" and transfer the contents of https://docs.postgrest.org/en/latest/references/configuration.html#client-error-verbosity there.

Then https://docs.postgrest.org/en/latest/references/configuration.html#client-error-verbosity should only link to that section.

taimoorzaeem:
> [..] IMO this is first place anyone would look for.

Yes, I agree. Opened #4676.

yilinjuang:
Hi, thanks for adding `client-error-verbosity`!

I saw this is merged into `main` but still under `Unreleased` in the changelog. Is there a rough timeline for a 14.x release including it?

Thanks!

taimoorzaeem:
@yilinjuang We usually release new features in the next major version, so that would be `v16` — not sure about the timeline for that. But, if they are low risk for breaking anything or introducing bugs, then we do include it into a minor release.

@steve-chavez @wolfgangwalther To me, this seems low risk, should we include this in a minor release?

src/PostgREST/App.hs
```diff
@@ -119,30 +119,31 @@ postgrest logLevel appState connWorker =
...
+          -- the connWorker is done. However, when there's an empty schema cache
+          -- postgrest responds with the error `PGRST002`; this means that the schema
+          -- cache is still loading, so we don't launch the connWorker here because
+          -- it would duplicate the loading process, e.g. https://github.com/PostgREST/postgrest/issues/3704
+          -- TODO: this process may be unnecessary when the Listener is enabled. Revisit once https://github.com/PostgREST/postgrest/issues/1766 is done
+          when (isServiceUnavailable response && isJust maybeSchemaCache) connWorker
+          resp <- do
+            delay <- AppState.getNextDelay appState
+            return $ addRetryHint delay response
+          respond resp
```
steve-chavez, line 146: Just a note, I think this big diff could have been avoided? Maybe we should take that into consideration, so if we want to backport something or not the decision is easier.

steve-chavez:
> @steve-chavez @wolfgangwalther To me, this seems low risk, should we include this in a minor release?

@taimoorzaeem It looks harmless to me as it only affects error messages. I'd say yes but let's wait for @wolfgangwalther confirmation too.

src/PostgREST/App.hs
taimoorzaeem: Agree, but the thing is that this is mostly just whitespace/indentation change and github doesn't have a nice way to ignore it. You can add `?w=1` to suppress it.

Try viewing the same in https://github.com/PostgREST/postgrest/pull/4667/changes?w=1 and it looks much better. :sunglasses:

steve-chavez: > Try viewing the same in https://github.com/PostgREST/postgrest/pull/4667/changes?w=1 and it looks much better. 😎

TIL, that's pretty cool! :cool:

wolfgangwalther:
Seems like this is a plain feature that does not fix anything, does not help to debug anything or anything of that sort. I don't think I even need to judge the specific risk of the specific PR, I think this is nothing to backport. Features need to wait for the next release. If people need to wait too long - we will need to make more major releases, not start backporting more.

src/PostgREST/App.hs
wolfgangwalther: > Agree, but the thing is that this is mostly just whitespace/indentation change and github doesn't have a nice way to ignore it. You can add `?w=1` to suppress it.

Or... just use the UI to enable it: Click the gear / settings icon right next to the "submit comments" / "submit reviews" button. There is an option "hide whitespace" that toggles this exact URL parameter.

taimoorzaeem:
> Features need to wait for the next release.

I think there is nothing wrong with releasing harmless features in a minor release. I mean that's what minor releases are for, right?

> If people need to wait too long - we will need to make more major releases, not start backporting more.

From a maintenance POV, I agree here. It would be much easier to do major releases than backporting the feature (resolving conflicts, backporting the refactors first that led to the feature, etc).

One _painful_ thing would be that users would need to adjust to breaking changes on every major instead of getting that one feature right away.

Overall, I agree with doing more major releases.

PR #4668, opened 2026-02-24
add(observation): db pool flushes observation

mkleczek:
**DISCLAIMER:**

Emit a dedicated PoolFlushed observation when the DB pool is released during schema cache reload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. From versio
 - Add config `client-error-verbosity` to customize error verbosity by @taimoorzaeem in #4088, #3980, #3824
 - Add `Vary` header to responses by @develop7 in #4609
 - Add config `db-timezone-enabled` for optional querying of timezones by @taimoorzaeem in #4751
+- Log when the pool is released during schema cache reload on `log-level=debug` by @mkleczek in #4668
 
 ### Changed
 
--- a/test/observability/Observation/SchemaCacheSpec.hs
+++ b/test/observability/Observation/SchemaCacheSpec.hs
@@ -1,0 +1,51 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE MonadComprehensions #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+module Observation.SchemaCacheSpec where
+
+import           Network.Wai           (Application)
+import           ObsHelper
+import qualified PostgREST.AppState    as AppState
+import           PostgREST.Config      (configDbSchemas)
+import           PostgREST.Observation
+import           Protolude
+import           Test.Hspec            (SpecWith, describe, it)
+import           Test.Hspec.Wai        (getState)
+
+spec :: SpecWith (SpecState, Application)
+spec = describe "Server started with metrics enabled" $ do
+
+  it "Should emit PoolFlushed, SchemaCacheQueriedObs and SchemaCacheLoadedObs when schema cache is reloaded" $ do
+    SpecState{specAppState = appState, specObsChan} <- getState
+    let waitFor = waitForObs specObsChan
+
+    liftIO $ do
+      AppState.schemaCacheLoader appState
+
+      waitFor (1 * sec) "PoolFlushed" $ \x -> [ o | o@PoolFlushed <- pure x ]
+      waitFor (1 * sec) "SchemaCacheQueriedObs" $ \x -> [ o | o@SchemaCacheQueriedObs{} <- pure x ]
+      waitFor (1 * sec) "SchemaCacheLoadedObs" $ \x -> [ o | o@SchemaCacheLoadedObs{} <- pure x ]
+
+
+  it "Should flush pool multiple times when schema reloading retries" $ do
+    SpecState{specAppState = appState, specObsChan} <- getState
+    let waitFor = waitForObs specObsChan
+
+    liftIO $ do
+        AppState.getConfig appState >>= \cfg -> do
+          AppState.putConfig appState $ cfg { configDbSchemas = pure "bad_schema" }
+          AppState.schemaCacheLoader appState
+
+          waitFor (1 * sec) "PoolFlushed 1" $ \x -> [ o | o@PoolFlushed <- pure x ]
+          waitFor (1 * sec) "SchemaCacheErrorObs" $ \x -> [ o | o@SchemaCacheErrorObs{} <- pure x ]
+
+          -- Restore configuration
+          AppState.putConfig appState cfg
+
+        -- Wait for 2 seconds so that retry can happen
+        waitFor (2 * sec) "PoolFlushed 2" $ \x -> [ o | o@PoolFlushed <- pure x ]
+        waitFor (1 * sec) "SchemaCacheQueriedObs" $ \x -> [ o | o@SchemaCacheQueriedObs{} <- pure x ]
+        waitFor (1 * sec) "SchemaCacheLoadedObs" $ \x -> [ o | o@SchemaCacheLoadedObs{} <- pure x ]
+
+  where
+    sec = 1000000
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -308,6 +308,7 @@ test-suite observability
   other-modules:      ObsHelper
                       Observation.JwtCache
                       Observation.MetricsSpec
+                      Observation.SchemaCacheSpec
   build-depends:      base              >= 4.9 && < 4.20
                     , base64-bytestring >= 1 && < 1.3
                     , bytestring        >= 0.10.8 && < 0.13
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -225,10 +225,14 @@ usePool AppState{stateObserver=observer, stateMainThreadId=mainThreadId, ..} ses
 
 -- | Flush the connection pool so that any future use of the pool will
 -- use connections freshly established after this call.
+-- | Emits PoolFlushed observation
 flushPool :: AppState -> IO ()
-flushPool AppState{..} = SQL.release statePool
+flushPool AppState{..} = do
+  SQL.release statePool
+  stateObserver PoolFlushed
 
 -- | Destroy the pool on shutdown.
+-- | Differs from flushPool in not emiting PoolFlushed observation.
 destroyPool :: AppState -> IO ()
 destroyPool AppState{..} = SQL.release statePool
 
--- a/src/PostgREST/Logger.hs
+++ b/src/PostgREST/Logger.hs
@@ -111,6 +111,9 @@ observationLogger loggerState logLevel obs = case obs of
   o@PoolRequestFullfilled ->
     when (logLevel >= LogDebug) $ do
       logWithZTime loggerState $ observationMessages o
+  o@PoolFlushed ->
+    when (logLevel >= LogDebug) $ do
+      logWithZTime loggerState $ observationMessages o
   o@JwtCacheEviction ->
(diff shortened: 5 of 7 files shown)


test/io/test_io.py
```diff
@@ -21,6 +21,19 @@
...
+    nextMatcher = next(itm, None)
+    while nextMatcher is not None and (line := next(ito, None)) is not None:
+        if re.match(nextMatcher, line) is not None:
+            nextMatcher = next(itm, None)
+    if nextMatcher is not None:
+        raise AssertionError(
+            f"Expected log line matching {nextMatcher} not found in output"
+        )
+
+
```
steve-chavez, line 36: Can this fit into another independent commit? Looks like a refactor of the tests.

test/spec/Main.hs
```diff
@@ -95,6 +96,13 @@ main = do
       AppState.putSchemaCache appState (Just sCache)
       return (st, postgrest (configLogLevel config) appState (pure ()))
 
+    initObservationsApp sCache config = do
+      obsChan <- newChan
+      appState <- AppState.initWithPool sockets pool config loggerState metricsState (Metrics.observationMetrics metricsState <> writeChan obsChan)
```
steve-chavez: This is interesting, so we use the [newChan](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Concurrent-Chan.html#v:newChan) instead of logging to stdout/stderr right?

What would be the advantage of this over the python tests?

mkleczek: > This is interesting, so we use the [newChan](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Control-Concurrent-Chan.html#v:newChan) instead of logging to stdout/stderr right?

Haskell makes it so easy to add new observers (`Observation -> IO ()` is a `Semigroup`) that it is too tempting not to use it :)

> 
> What would be the advantage of this over the python tests?

More seriously (this is a mix of more or less subjective opinions):

It is not really clear to me what is the project policy on "what tests should go where". My take on this is that I really think it is quite a bad idea to base your business logic tests on debug output.

I spent a considerable amount of time chasing failing `io` tests after adding this new observation and was really annoyed with the experience. So decided to move to `HSpec` tests instead.

I think `HSpec` tests are really well suited for logic testing. They are type safe and make refactoring easy. Also, reading observations as a stream of typed messages in tests (instead of untyped string debug output) makes the tests less brittle and more readable IMHO.

`Chan` is a simple but powerful concurrency primitive and, IMHO, perfect fit for testing asynchronous messaging scenarios. See https://github.com/PostgREST/postgrest/pull/4668/changes#diff-0f4d50d33174e92e7c2b8121202cbcb7923360111854c13c2f4baa9eadb80602R44 in this PR. Thanks to some higher level functions based on `Chan` and `dupChan` we have a simple DSL to validate observations that also provides nice error messages that simplify debugging.

test/io/test_io.py
mkleczek: @steve-chavez Done

test/spec/Main.hs
steve-chavez: > I spent a considerable amount of time chasing failing io tests after adding this new observation and was really annoyed with the experience. So decided to move to HSpec tests instead.

Agree, I've had the same experience before.

> My take on this is that I really think it is quite a bad idea to base your business logic tests on debug output.
> Also, reading observations as a stream of typed messages in tests (instead of untyped string debug output) makes the tests less brittle and more readable IMHO.

Agree too. I like the direction overall.

---

Sometimes we've had cases where the ordering of the log messages change and that was reflected in tests. Say if we decided to move all logging python tests to Haskell, would a change in the order be detected too? (order does matter in some cases, for clarity)

mkleczek: > 
> Sometimes we've had cases where the ordering of the log messages change and that was reflected in tests. Say if we decided to move all logging python tests to Haskell, would a change in the order be detected too? (order does matter in some cases, for clarity)

That depends on how observations order is related to debug log entries order. They should be the same but we don't test it in `HSpec` tests - we only read observations from a `Chan`.
There is a simple helper function `waitFor` which reads from the `Chan` skipping messages until it reads the message meeting provided criteria (or hangs waiting for an observation (or timeout) if none available). I think it is sufficient primitive for most cases.
You can test if required observations happen in right order (but possibly with other observations in between), eg:
```
        waitFor (2 * sec) "PoolFlushed 2" $ \x -> [ o | o@PoolFlushed <- pure x ]
```
This will skip `SchemaCacheQueriedObs` in between.

If necessary we can implement non-lenient validation.

Let's move this discussion to https://github.com/PostgREST/postgrest/pull/4671#issuecomment-3967558822

steve-chavez:
> add(observation): db pool flushes observation

- Should be called `add: log pool flushes`?
- This should only be done on `log-level=debug` it seems?
- Also needs an entry on the CHANGELOG

taimoorzaeem:
Friendly reminder - The commit descriptions in this PR also contain the `DISCLAIMER: ...`. So, just in case you forgot, we don't prefer these as mentioned in https://github.com/PostgREST/postgrest/pull/4640#issuecomment-4201452761.

mkleczek:
> > add(observation): db pool flushes observation
> 
> * Should be called `add: log pool flushes`?

Done

> * This should only be done on `log-level=debug` it seems?

Yes, the observation is logged at debug log level.

> * Also needs an entry on the CHANGELOG

Done

mkleczek:
> Friendly reminder - The commit descriptions in this PR also contain the `DISCLAIMER: ...`. So, just in case you forgot, we don't prefer these as mentioned in [#4640 (comment)](https://github.com/PostgREST/postgrest/pull/4640#issuecomment-4201452761).

Fixed

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. From versio
 - Add config `client-error-verbosity` to customize error verbosity by @taimoorzaeem in #4088, #3980, #3824
 - Add `Vary` header to responses by @develop7 in #4609
 - Add config `db-timezone-enabled` for optional querying of timezones by @taimoorzaeem in #4751
+- Add a dedicated PoolFlushed observation when the DB pool is released during schema cache reload by @mkleczek in #4668
```
steve-chavez: Should be more user facing (observation is more of an internal abstraction):
```suggestion
- Log when the pool is released during schema cache reload on `log-level=debug` by @mkleczek in #4668
```

test/observability/Observation/SchemaCacheSpec.hs
```diff
@@ -0,0 +1,51 @@
...
+          AppState.schemaCacheLoader appState
+
+          waitFor (1 * sec) "PoolFlushed 1" $ \x -> [ o | o@PoolFlushed <- pure x ]
+          waitFor (1 * sec) "SchemaCacheErrorObs" $ \x -> [ o | o@SchemaCacheErrorObs{} <- pure x ]
+
+          -- Restore configuration
+          AppState.putConfig appState cfg
+
+        -- Wait for 2 seconds so that retry can happen
+        waitFor (2 * sec) "PoolFlushed 2" $ \x -> [ o | o@PoolFlushed <- pure x ]
```
steve-chavez, line 46: Curious, what does the `2` in `PoolFlushed 2` mean?

CHANGELOG.md
mkleczek: Done

test/observability/Observation/SchemaCacheSpec.hs
mkleczek: The text in quotes is just description and used for easier tracking the exact spot of failure in case we miss a specific observation.

So in this case it is just to make sure that it is easy to identify which specific `PoolFlushed` observation failed (the second one).

steve-chavez: Ah, so the `msg` in `waitForObs` can be anything really?

I might have misunderstood the function, I was thinking the `msg` needed to match the constructor name of the observation.

https://github.com/PostgREST/postgrest/blob/0e820b9ac569f6907fa230043c5bbd4eea0751e0/test/observability/ObsHelper.hs#L194-L195

steve-chavez:
Github has been so buggy lately, I cannot merge through UI:

image

laurenceisla:
> Github has been so buggy lately, I cannot merge through UI:


Neither can I. Got the same error: 
```
Base branch was modified. Review and try the merge again.
```

Edit: Tried approving and merge again. Didn't work.

test/observability/Observation/SchemaCacheSpec.hs
mkleczek: > Ah, so the `msg` in `waitForObs` can be anything really?

Yes.

> 
> I might have misunderstood the function, I was thinking the `msg` needed to match the constructor name of the observation.

Constructor names and `msg` have different purpose:

`msg` is to distinguish between multiple expectations, eg:

```
waitFor (2 * sec) "PoolFlushed 1" $ \x -> [ o | o@PoolFlushed <- pure x ]
waitFor (2 * sec) "PoolFlushed 2" $ \x -> [ o | o@PoolFlushed <- pure x ]
waitFor (2 * sec) "PoolFlushed 3" $ \x -> [ o | o@PoolFlushed <- pure x ]
```

Upon failure, without `msg`, you would get a message saying "Waiting for an observation failed" and you wouldn't be able to tell where exactly the failure happened (line numbers from `HasCallstack` might help but they are not always precise enough in Haskell).

Constructor names are used to produce output with all observations that were produced until failure (so that it is easier to diagnose the test failures), along the lines of:
```
Failed waiting for "PoolFlushed 2" observation.
Received observations were:
SchemaCacheQueried
OtherObservationConstructorName
SchemaCacheLoaded
```
Looking at this and the test code above, you know one "PoolFlushed" was received (`msg` is "PoolFlushed 2") and after that 3 observations were received but none of them was the one we expected.

mkleczek:
> > Github has been so buggy lately, I cannot merge through UI:
> 
> Neither can I. Got the same error:
> 
> ```
> Base branch was modified. Review and try the merge again.
> ```
> 
> Edit: Tried approving and merge again. Didn't work.

Hmm... I've rebased it on top of current `main` - can you try again?

laurenceisla:
> Hmm... I've rebased it on top of current main - can you try again?

Just tried it and getting the same error again.

steve-chavez:
Same for me too 😞 

Maybe try opening another PR?

taimoorzaeem:
Manually checking the log for this PR:

<pre>
commit adf6802c85c3f31bf882ee7a5a232c79dcb520f7 (HEAD -> pool-flush-obs)
<b>Merge: db998f3c 8d5b1018</b>
Author: Michał Kłeczek <michal@kleczek.org>
Date:   Wed Feb 25 18:36:03 2026 +0100

    add: log pool flushes
    
    Emit a dedicated PoolFlushed observation when the DB pool is released during schema cache reload.

commit db998f3c4ff0fe3669ff22450fd4791f07206615
Author: Michał Kłeczek <michal@kleczek.org>
Date:   Wed Feb 25 06:23:18 2026 +0100

    test(refactor): make debug output testing more robust
    
    This change introduces a match_log function, that implements lenient matching of log output lines. Tests of log output are retrofited to use the function.
</pre>

The top commit is a merge commit (it has two parent commits). I think that might be the reason that the rebase is getting rejected.

mkleczek:
> The top commit is a merge commit (it has two parent commits). I think that might be the reason that the rebase is getting rejected.

Ahh, good catch... That seems an artifact of my use of https://github.com/jj-vcs/jj to manage stacked PRs.

Fixed.

@steve-chavez @laurenceisla - sorry for confusion, will pay more attention to this in the future.

PR #3467, opened 2024-04-30
test: sanity tests for primary and replica

steve-chavez:
* new --replica option to `postgrest-with-postgresql-*`
  + replica process is properly logged to usual temporary log files
* new command `postgrest-test-replica`
  + runs sanity tests on `test_replica.py`
* add `postgrest-test-replica` to `postgrest-check` and `postgrest-coverage`

PR #4671, opened 2026-02-25
refactor(test): provide means to validate metrics and observations

mkleczek:
**DISCLAIMER:**

Some helpers are provided for introspecting metrics already (used in JWT cache tests). This change provides facilities to additionally validate emited Observation events.
A new Spec module is also implemented, adding basic tests of schema cache reloading - their main goal is to excercise the new infrastructure.

This is a prerequisite for:
https://github.com/PostgREST/postgrest/pull/4672 (and subsequent https://github.com/PostgREST/postgrest/pull/4643)
https://github.com/PostgREST/postgrest/pull/4668
https://github.com/PostgREST/postgrest/pull/4645

### DIFF
--- a/test/observability/Main.hs
+++ b/test/observability/Main.hs
@@ -29,6 +29,7 @@ main = do
   -- start a thread that drains the channel
   -- this is necessary because test cases operate on
   -- copies so poolChan is never read from
+  -- this means we have another thread running for the entire duration of the spec but this shouldn't be a problem since Haskell green threads are lightweight
   void $ forkIO $ forever $ readChan poolChan
   metricsState <- Metrics.init (configDbPoolSize testCfg)
   pool <- P.acquire $ P.settings
--- a/test/observability/ObsHelper.hs
+++ b/test/observability/ObsHelper.hs
@@ -38,6 +38,7 @@ import           Test.Hspec
 import           Test.Hspec.Expectations.Contrib (annotate)
 
 -- helpers used to produce observation diagnostics in waitForObs
+-- Implementing the Show instance for Observation is hard due to having many different parameters so instead we use generic programming (`conName`) to obtain the constructor name as `Text`
 class HasConstructor f where
   genericConstrName :: f x -> Text
 


steve-chavez:
FYI I had a question about testing the order of the logs on https://github.com/PostgREST/postgrest/pull/4668#discussion_r2855269526

mkleczek:
> FYI I had a question about testing the order of the logs on [#4668 (comment)](https://github.com/PostgREST/postgrest/pull/4668#discussion_r2855269526)

@steve-chavez - https://github.com/PostgREST/postgrest/pull/4668#discussion_r2859869858 (copying it here as well, I guess that's a better place to discuss it).

That depends on how observations order is related to debug log entries order. They should be the same but we don't test it in HSpec tests - we only read observations from a Chan.
There is a simple helper function waitFor which reads from the Chan skipping messages until it reads the message meeting provided criteria (or hangs waiting for an observation (or timeout) if none available). I think it is sufficient primitive for most cases.
You can test if required observations happen in right order (but possibly with other observations in between), eg:

        waitFor (2 * sec) "PoolFlushed 2" $ \x -> [ o | o@PoolFlushed <- pure x ]
This will skip SchemaCacheQueriedObs in between.

If necessary we can implement non-lenient validation.

test/spec/Feature/MetricsSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+          waitFor (1 * sec) "SchemaCacheErrorObs" $ \x -> [ o | o@(SchemaCacheErrorObs{}) <- pure x]
+          AppState.putConfig appState prev
+
+        -- wait up to 2 secs so that retry can happen
+        waitFor (2 * sec) "SchemaCacheLoadedObs" $ \x -> [ o | o@(SchemaCacheLoadedObs{}) <- pure x]
+
+  where
+    -- prometheus-client api to handle vectors is convoluted
+    schemaCacheLoads label = expectField @"schemaCacheLoads" $
+      (foldMap (Sum . round @Double @Int . snd) . find ((== label) . fst) <$>) . (`getVectorWith` getCounter)
```
steve-chavez: Yes this does look convoluted. 🤔. Perhaps the accumulator can be separated from the getter part? So we have two functions?

mkleczek: > Yes this does look convoluted. 🤔. Perhaps the accumulator can be separated from the getter part? So we have two functions?

@steve-chavez Done

steve-chavez: Looks much better :+1:

test/spec/SpecHelper.hs
```diff
@@ -381,3 +385,45 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+prepareState = getState >>= traverse (liftA2 (<$>) waitFor (liftIO . dupChan))
+  where
+    -- read messages from copy chan and once condition is met drain original to the same point
+    -- upon timeout report error and messages remaining in the original chan
+    -- that way we report messages since last successful read
+    waitFor orig copy t msg f =
+      timeout t (readUntil copy *> readUntil orig) >>= maybe failTimeout mempty
+      where
+        failTimeout =
+          takeUntilTimeout 100000 (readChan orig)
```
steve-chavez: For clarity maybe move this 100000 to a variable.

test/spec/Main.hs
```diff
@@ -68,16 +69,23 @@ import qualified Feature.Query.UpdateSpec
...
 import qualified Feature.RollbackSpec
 import qualified Feature.RpcPreRequestGucsSpec
+import           PostgREST.Observation                 (Observation (HasqlPoolObs))
 
 
 main :: IO ()
 main = do
+  poolChan <- newChan
+  -- make sure poolChan is not growing indefinitely
+  void $ forkIO $ fix (readChan poolChan *>)
```
steve-chavez: Could you elaborate here? I don't quite get how this works.

mkleczek: > Could you elaborate here? I don't quite get how this works.

`fix (readChan poolChan *>)` is convoluted `forever $ readChan poolChan` - updated

Added comment to explain the reasoning behind starting this thread.

test/spec/SpecHelper.hs
mkleczek: > For clarity maybe move this 100000 to a variable.

Couldn't come up with a good name. Updated to at least make the actual time value more visible.

test/spec/SpecHelper.hs
```diff
@@ -381,3 +385,46 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+prepareState = getState >>= traverse (liftA2 (<$>) waitFor (liftIO . dupChan))
+  where
+    -- read messages from copy chan and once condition is met drain original to the same point
+    -- upon timeout report error and messages remaining in the original chan
+    -- that way we report messages since last successful read
+    waitFor orig copy t msg f =
+      timeout t (readUntil copy *> readUntil orig) >>= maybe failTimeout mempty
+      where
+        failTimeout =
+          takeUntilTimeout (100 * millis) (readChan orig)
```
steve-chavez: Is it a decisecond? Maybe the variable can be named like that.

https://github.com/PostgREST/postgrest/blob/2edc44c3520478e5a838534798e5965732fd3b83/src/PostgREST/AppState.hs#L157

test/spec/Main.hs
steve-chavez: Looks much clearer :+1:

test/spec/SpecHelper.hs
```diff
@@ -381,3 +385,46 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+            >>= expectationFailure
+            . ("Timeout waiting for " <> msg <> " at " <> loc <> ". Remaining observations:\n" ++)
+            . foldMap ((++ "\n") . show . observationMessage)
+        readUntil = void . untilM (pure . isJust . f) . readChan
+    loc = fold (head (prettySrcLoc . snd <$> getCallStack callStack))
+    -- execute effectful computation until result meets provided condition
+    untilM cond m = fix $ \loop -> m >>= \a -> ifM (cond a) (pure a) loop
```
steve-chavez: Is it the same as [untilM](https://hackage.haskell.org/package/monad-loops-0.4.3/docs/Control-Monad-Loops.html#v:untilM)? Perhaps we can reuse this package?

test/spec/SpecHelper.hs
mkleczek: > Is it a decisecond? Maybe the variable can be named like that.
> 
> https://github.com/PostgREST/postgrest/blob/2edc44c3520478e5a838534798e5965732fd3b83/src/PostgREST/AppState.hs#L157

@steve-chavez Done

Aside: I was confused when I saw this `decisecond` - had to browse through ISO prefixes in my old brain to understand that it is `100 ms` :)

test/spec/SpecHelper.hs
mkleczek: > Is it the same as [untilM](https://hackage.haskell.org/package/monad-loops-0.4.3/docs/Control-Monad-Loops.html#v:untilM)? Perhaps we can reuse this package?

Not sure if it is worth pulling in the whole package for a one-liner (we don't want `npm` story, I guess :-) ). Besides, [monad-loops](https://hackage.haskell.org/package/monad-loops-0.4.3) looks dormant...

test/spec/SpecHelper.hs
```diff
@@ -381,3 +385,46 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+      -- make sure timer thread is killed on other exceptions
+      -- so that it won't throw TimeoutException later
+      `onException` killThread ttid
+  where
+    onTimeout m a = m `catch` \TimeoutException -> a
+
+
+prepareState :: HasCallStack => Traversable f => WaiSession (f (Chan Observation)) (f (Int -> String -> (Observation -> Maybe a) -> IO ()))
+prepareState = getState >>= traverse (liftA2 (<$>) waitFor (liftIO . dupChan))
+  where
```
taimoorzaeem: I think it would be better to avoid generalizing the code before hand. It hides the type, makes the code a bit difficult to read, review and maintain. We can use type aliases to increase the readability and it would help speed up the review process:

```suggestion
type AppStateAndMetrics = (AppState, Metrics)
type WaitForObsHandler = Int -> String -> (Observation -> Maybe Observation) -> IO ()

prepareObsState :: HasCallStack => WaiSession (AppStateAndMetrics, (Chan Observation)) (AppStateAndMetrics, WaitForObsHandler)
prepareObsState = do
  (appStateAndMetrics, chan) <- getState
  dup <- liftIO (dupChan chan)
  return (appStateAndMetrics, waitFor chan dup)
  where
```

mkleczek: > I think it would be better to avoid generalizing the code before hand. It hides the type, makes the code a bit difficult to read, review and maintain.

@taimoorzaeem thanks for the review and the insight! Indeed - the way it was written was confusing through indirection. I've restructured the code a little

- introduced `ObsChan` data type encapsulating a channel and its copy
- got rid of `prepareState` and introduced top-level `waitForObs` function

I think it is now easier to follow. WDYT?

test/spec/SpecHelper.hs
taimoorzaeem: > Not sure if it is worth pulling in the whole package for a one-liner [..] Besides, [monad-loops](https://hackage.haskell.org/package/monad-loops-0.4.3) looks dormant...

I think it would be worth it. `monad-loops` is a complete library for monadic iteration related functions. It looks dormant because I guess it is _complete_ and _bug free_ for years.

Besides, there are other handy functions in that library which could be useful in the future to improve our test infrastructure.

mkleczek: > > Not sure if it is worth pulling in the whole package for a one-liner [..] Besides, [monad-loops](https://hackage.haskell.org/package/monad-loops-0.4.3) looks dormant...
> 
> I think it would be worth it. `monad-loops` is a complete library for monadic iteration related functions. It looks dormant because I guess it is _complete_ and _bug free_ for years.

I double-checked this library and it does not provide quite the function that we need here (ie. loop until the returned value meets provided criteria). I could reimplement the logic in terms of `monad-loops` `untilJust` function probably, but bear with me: pulling a library and reimplement larger function to adjust it to this library API, just to replace a one-liner, does not meet the bar to use the library IMHO.

> 
> Besides, there are other handy functions in that library which could be useful in the future to improve our test infrastructure.

Quite possibly you're right. But this is out of scope of this PR.

In general, importing a library "just in case it is needed in the future" is not a good idea IMO.

test/spec/Feature/MetricsSpec.hs
```diff
@@ -0,0 +1,55 @@
...
+
+        -- wait up to 2 secs so that retry can happen
+        waitFor (2 * sec) "SchemaCacheLoadedObs" $ \x -> [ o | o@(SchemaCacheLoadedObs{}) <- Just x]
+
+  where
+    -- prometheus-client api to handle vectors is convoluted
+    schemaCacheLoads label = expectField @"schemaCacheLoads" $
+      fmap (sumSamples . findSamples label) . (`getVectorWith` getCounter)
+    sumSamples = getSum . foldMap (Sum . round @Double @Int . snd)
+    findSamples label = find ((== label) . fst)
```
taimoorzaeem: Hm, can we make this part a bit clear? I am confused on why we need `foldMap` and `Sum` type? Without types annotations, it is bit hard to see what underlying data structures are being modified by these functions.

test/spec/Feature/MetricsSpec.hs
```diff
@@ -0,0 +1,55 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE ImpredicativeTypes  #-}
```
taimoorzaeem: This is unused and can be removed.

```suggestion
```

test/spec/Feature/MetricsSpec.hs
mkleczek: > Hm, can we make this part a bit clear? I am confused on why we need `foldMap` and `Sum` type? Without types annotations, it is bit hard to see what underlying data structures are being modified by these functions.

We want `[(label, Double)] -> Int` - used `Sum` monoid to get `0` if label not found.

Changed to direct `maybe` use, hope that's more clear.

BTW, I use `VS Code` with [Haskell Extension Pack](https://marketplace.visualstudio.com/items?itemName=mogeko.haskell-extension-pack) and hovering over a symbol shows me its type.

test/spec/Feature/MetricsSpec.hs
mkleczek: > This is unused and can be removed.

Thanks, fixed.

src/PostgREST/AppState.hs
```diff
@@ -18,6 +18,7 @@ module PostgREST.AppState
   , init
   , initSockets
   , initWithPool
+  , putConfig -- For tests TODO refactoring
```
taimoorzaeem, line 18: Why the `TODO` here? It's fine I think.

```suggestion
  , putConfig
```

test/spec/SpecHelper.hs
```diff
@@ -381,3 +389,54 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+waitForObs (ObsChan orig copy) t msg f =
+  timeout t (readUntil copy *> readUntil orig) >>= maybe failTimeout mempty
+  where
+    failTimeout =
+      takeUntilTimeout decisecond (readChan orig)
+        >>= expectationFailure
+        . ("Timeout waiting for " <> msg <> " at " <> loc <> ". Remaining observations:\n" ++)
+        . foldMap ((++ "\n") . show . observationMessage)
+    readUntil = void . untilM (pure . not . null . f) . readChan
+    loc = fromMaybe "(unknown)" . head $ (prettySrcLoc . snd <$> getCallStack callStack)
```
taimoorzaeem: Hm, `head` is partial and unsafe. Safer alternative:

```suggestion
    loc = maybe "(unknown)" (prettySrcLoc . snd) $ listToMaybe (getCallStack callStack)
```

test/spec/SpecHelper.hs
```diff
@@ -381,3 +389,54 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+-- that way we report messages since last successful read
+waitForObs :: HasCallStack => ObsChan -> Int -> String -> (Observation -> Maybe a) -> IO ()
+waitForObs (ObsChan orig copy) t msg f =
+  timeout t (readUntil copy *> readUntil orig) >>= maybe failTimeout mempty
+  where
+    failTimeout =
+      takeUntilTimeout decisecond (readChan orig)
+        >>= expectationFailure
+        . ("Timeout waiting for " <> msg <> " at " <> loc <> ". Remaining observations:\n" ++)
+        . foldMap ((++ "\n") . show . observationMessage)
```
taimoorzaeem: A bit cleaner by using `unlines` from `Data.List`, no explicit `\n` needed.

```suggestion
    failTimeout = takeUntilTimeout decisecond (readChan orig) >>= expectationFailure . printError
    printError obs  = DL.unlines $
      ["Timeout waiting for " <> msg <> " at " <> loc <> ". Remaining observations:"]
      ++ map (show . observationMessage) obs
```

test/spec/SpecHelper.hs
mkleczek: > Hm, `head` is partial and unsafe. Safer alternative:

It is fine:
`head` returns `Maybe` (otherwise the code would not compile - there is `fromMaybe` there). (We use `Protolude` that has safe version of `head`)
I'd leave `head` as it clearly shows we are interested in the first element of the list.

src/PostgREST/AppState.hs
mkleczek: `putConfig` was not exported before. Relaxing encapsulation introduces risk that it is going to be used incorrectly somewhere and that dependencies will cripple in.
Ideally: if we need to export a function only for testing it means we should split the module into `Internal` and `API` modules (`API` module exporting a subset of `internal` symbols).

I'd leave the decision to @steve-chavez .

test/spec/SpecHelper.hs
mkleczek: > A bit cleaner by using `unlines` from `Data.List`, no explicit `\n` needed.

@taimoorzaeem Good catch, thx! Fixed (a little differently using `Text` instead of `String` and `Protolude.unlines`)

src/PostgREST/AppState.hs
steve-chavez: > Ideally: if we need to export a function only for testing it means we should split the module into Internal and API modules (API module exporting a subset of internal symbols).

That looks good to me.

mkleczek:
@steve-chavez @taimoorzaeem do you think it is now in mergeable state?

taimoorzaeem:
> @steve-chavez @taimoorzaeem do you think it is now in mergeable state?

Almost there!

I am quite sold on new interface for testing metrics and observation. However, the implementation part needs careful review because lots of new stuff there (async exceptions, masking/unmasking are new for me) and it's taking some more time from my side.

test/spec/SpecHelper.hs
```diff
@@ -381,3 +392,72 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+
+instance HasConstructor f => HasConstructor (D1 c f) where
+  genericConstrName (M1 x) = genericConstrName x
+
+instance (HasConstructor x, HasConstructor y) => HasConstructor (x :+: y) where
+  genericConstrName (L1 l) = genericConstrName l
+  genericConstrName (R1 r) = genericConstrName r
+
+instance Constructor c => HasConstructor (C1 c f) where
+  genericConstrName = T.pack . conName
```
taimoorzaeem: Hmm, is this strictly needed? Perhaps maybe we can just use the `Show` instance?

mkleczek: > Hmm, is this strictly needed? Perhaps maybe we can just use the `Show` instance?

@taimoorzaeem
Yes, it is needed: try to implement `Show` instance for `Observation` - I certainly tried that first :)

test/spec/Main.hs
```diff
@@ -68,16 +69,26 @@ import qualified Feature.Query.UpdateSpec
...
 
 
 main :: IO ()
 main = do
+  poolChan <- newChan
+  -- make sure poolChan is not growing indefinitely
+  -- start a thread that drains the channel
+  -- this is necessary because test cases operate on
+  -- copies so poolChan is never read from
+  void $ forkIO $ forever $ readChan poolChan
```
taimoorzaeem: Hmm, does this mean we have another thread running for the entire duration of `postgrest-test-spec`?

mkleczek: > Hmm, does this mean we have another thread running for the entire duration of `postgrest-test-spec`?

Yes. Do you think it is a problem? Haskell threads are green threads so really lightweight.

taimoorzaeem:
Thinking on a broader level, I think maybe we should separate observability tests from our spec test-suite. I have noticed that our spec tests mostly test end-to-end behaviour i.e request and response objects.

Observability testing requires a different setup (threads, timeouts, state checking etc), so maybe we should give it it's own test-suite? Probably also move `JwtCacheSpec.hs` there?

**Pros:**

- Better separation of concerns which means easier to review changes.

**Cons:**

- It would require some initial setup, wrapper scripts like `postgrest-test-observability`, coverage setup etc

Not sure if it would be worth it, thoughts from team members?

CC: @steve-chavez @wolfgangwalther @laurenceisla @mkleczek @develop7

mkleczek:
> Thinking on a broader level, I think maybe we should separate observability tests from our spec test-suite.

I considered it both when implementing JwtCache tests and when working on this PR.
Decided to keep it in the same suite but in separate `Spec` modules.

First reason is, as you say:
> 
> **Cons:**
> 
> * It would require some initial setup, wrapper scripts like `postgrest-test-observability`, coverage setup etc
> 
> Not sure if it would be worth it, thoughts from team members?
> 

IMHO splitting tests into separate `Spec` modules with proper granularity is enough.

Splitting into two separate suites (each with its own setup) would introduce a lot of redundancy but also lead to constant back-and-forth about the proper placement of a particular test (some tests are cross-cutting: require both e2e verification and state checking).

steve-chavez:
> It would require some initial setup, wrapper scripts like postgrest-test-observability, coverage setup etc

Makes perfect sense, also considering that we'll need [custom tooling for otel](https://github.com/PostgREST/postgrest/pull/3140#issuecomment-3954267899) and [for sentry](https://github.com/PostgREST/postgrest/issues/1596#issuecomment-2704402246).. adding those to `postgrest-test-spec` would make the Nix expression more complicated.

 (coverage setup is not that hard, an example [here](https://github.com/PostgREST/postgrest/pull/3467))

> Splitting into two separate suites (each with its own setup) would introduce a lot of redundancy but also lead to constant back-and-forth about the proper placement of a particular test (some tests are cross-cutting: require both e2e verification and state checking).

I think redundancy is fine if it makes the test suites more focused/isolated/reviewable, we already have some redundancy in io and spec schemas ([sleep in test/io/fixtures/schema.sql](https://github.com/PostgREST/postgrest/blob/93db1a5a47ed68d73b93eefb9a326e33a57a2461/test/io/fixtures/schema.sql#L116-L118), [sleep in  test/spec/fixtures/schema.sql](https://github.com/PostgREST/postgrest/blob/93db1a5a47ed68d73b93eefb9a326e33a57a2461/test/spec/fixtures/schema.sql#L3679-L3681)) and that has been working well.

mkleczek:
> > It would require some initial setup, wrapper scripts like postgrest-test-observability, coverage setup etc
> 
> Makes perfect sense, also considering that we'll need [custom tooling for otel]

OK, can it be done as a separate refactoring PR?
I don’t have bandwidth for designing new shape of our test suites. Also, ass this is quite major design I also expect it to take quite some time to reach consensus. But once it is done it should be trivial to move some code around - tests using the new APIs are isolated in their own modules.
OTOH there are multiple PRs dependent on the new test APIs.

This and stacked PRs can also wait for someone to contribute the new test suite structure.

Maybe kaizen mindset would be a good idea?

taimoorzaeem:
> Makes perfect sense, also considering that we'll need [custom tooling for otel](https://github.com/PostgREST/postgrest/pull/3140#issuecomment-3954267899) and [for sentry](https://github.com/PostgREST/postgrest/issues/1596#issuecomment-2704402246).. adding those to `postgrest-test-spec` would make the Nix expression more complicated.
> 
> (coverage setup is not that hard, an example [here](https://github.com/PostgREST/postgrest/pull/3467))

Yes, agree. :+1: 

> [..] I don’t have bandwidth for designing new shape of our test suites.

I will try to do it. Let's see.

mkleczek:
> > Makes perfect sense, also considering that we'll need [custom tooling for otel](https://github.com/PostgREST/postgrest/pull/3140#issuecomment-3954267899) and [for sentry](https://github.com/PostgREST/postgrest/issues/1596#issuecomment-2704402246).. adding those to `postgrest-test-spec` would make the Nix expression more complicated.
> > (coverage setup is not that hard, an example [here](https://github.com/PostgREST/postgrest/pull/3467))
> 
> Yes, agree. 👍
>
> > [..] I don’t have bandwidth for designing new shape of our test suites.
> 
> I will try to do it. Let's see.

Cool, marking this one as draft and will rebase on top of the new structure.

taimoorzaeem:
> Cool, marking this one as draft and will rebase on top of the new structure.

@mkleczek Now that we have #4732, you may continue with this.

test/observability/ObsHelper.hs
```diff
@@ -133,3 +143,72 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+  specObsChan  :: ObsChan
+}
+
+-- helpers used to produce observation diagnostics in waitForObs
+constrName :: (HasConstructor (Rep a), Generic a)=> a -> Text
+constrName = genericConstrName . from
+
+class HasConstructor f where
+  genericConstrName :: f x -> Text
+
```
taimoorzaeem: How about we move the datatypes and type classes definition to the top of the current diff, it helps a bit in seeing the complete picture.

Usually, haskell modules (or code additions) have the following order:

- Type classes and instances
- Data types
- Functions

test/observability/ObsHelper.hs
```diff
@@ -133,3 +143,72 @@ expectCounter :: forall s st m. (KnownSymbol s, HasField s st Counter, MonadIO m
...
+    loc = fromMaybe "(unknown)" . head $ (T.pack . prettySrcLoc . snd <$> getCallStack callStack)
+    -- execute effectful computation until result meets provided condition
+    untilM cond m = fix $ \loop -> m >>= \a -> ifM (cond a) (pure a) loop
+    -- duplicate the provided channel and construct wairFor function binding both channels
+    -- accumulate effecful computation results into a list for specified time
+    takeUntilTimeout t' = fmap reverse . accumulateUntilTimeout t' (flip (:)) []
+    obsDiagMessage (HasqlPoolObs o) = show o
+    obsDiagMessage o@(DBListenStart host port name channel) = constrName o <> show (host, port, name, channel)
+    obsDiagMessage o = constrName o
+    decisecond = 100000
```
taimoorzaeem: Looks like this doesn't have to be an inner function, how about we move it to top-level, it may be reusable.

```suggestion
    decisecond = 100000
    
obsDiagMessage :: Observation -> Text
obsDiagMessage (HasqlPoolObs o) = show o
obsDiagMessage o@(DBListenStart host port name channel) = constrName o <> show (host, port, name, channel)
obsDiagMessage o = constrName o
```

test/observability/ObsHelper.hs
mkleczek: Fixed as requested

test/observability/ObsHelper.hs
mkleczek: Fixed as requested

test/spec/SpecHelper.hs
steve-chavez: I get this now, `Observation` is hard due to having lots of different parameters (dunno if they all support show) so instead we use [generic programming](https://downloads.haskell.org/ghc/7.6.3/docs/html/users_guide/generic-programming.html), like `conName` to just obtain the constructor name.

test/observability/ObsHelper.hs
```diff
@@ -1,32 +1,70 @@
...
+                                                  parseSecret)
+import qualified PostgREST.Metrics               as Metrics
+import           PostgREST.Observation           (Observation (..))
+import           Prometheus                      (Counter, getCounter)
+import           Protolude                       hiding (get, toS)
+import           System.Timeout                  (timeout)
+import           Test.Hspec
+import           Test.Hspec.Expectations.Contrib (annotate)
+
+-- helpers used to produce observation diagnostics in waitForObs
```
steve-chavez, line 40: Adding my comment here in case it helps understanding later
```suggestion
-- helpers used to produce observation diagnostics in waitForObs
-- Implementing the Show instance for Observation is hard due to having many different parameters so instead we use generic programming (`conName`) to obtain the constructor name as `Text`
```

test/observability/Main.hs
```diff
@@ -15,39 +15,54 @@ import qualified PostgREST.Metrics         as Metrics
...
+import Protolude             hiding (toList, toS)
 import Test.Hspec
 
 main :: IO ()
 main = do
+  poolChan <- newChan
+  -- make sure poolChan is not growing indefinitely
+  -- start a thread that drains the channel
+  -- this is necessary because test cases operate on
+  -- copies so poolChan is never read from
```
steve-chavez, line 31:
```suggestion
  -- copies so poolChan is never read from
  -- this means we have another thread running for the entire duration of the spec but this shouldn't be a problem since Haskell green threads are lightweight
```

Issue #4670, opened 2026-02-24
LISTEN channel silently stops receiving notifications due to high `keepalives_idle`


steve-chavez:
## Problem

When using [Listener on read replicas](https://docs.postgrest.org/en/v14/references/listener.html#listener-on-read-replicas), the postgREST listener on the replica has to connect to the pg master. This postgREST instance is usually on a different host than the master (communication can even be cross-DC), so the LISTEN channel is prone to failure due to network conditions.

The problem is that this LISTEN channel can be silently dead (and not receive notifications) and this will only be detected once the [keepalives_idle](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-KEEPALIVES-IDLE)  reach 2 hours (since this is the Linux default).

Related: https://github.com/PostgREST/postgrest/issues/3612

## Solution

Similarly to how we add  `target_session_attrs` to the Listener connection https://github.com/PostgREST/postgrest/blob/c5d495006a0a9118273ec810e21f428fd0d22f5b/src/PostgREST/Listener.hs#L65

Add something like:

```
postgresql://conn?keepalives_idle=30&keepalives_interval=1&keepalives_count=10'
```

The LISTEN channel is meant to be long-lived, so we should ensure it stays up. 

## Notes

- Maybe don't apply these `keepalives` when the connection is under `localhost`.
- We should determine the best defaults for the `keepalives_*` values somehow.

mkleczek:
@steve-chavez it would be good to have testing infrastructure with https://github.com/Shopify/toxiproxy set-up. We could then test it.
I started to work on it earlier but it stalled.

PR #4672, opened 2026-02-25
fix: implement connection tracking in metrics

### LINKED ISSUE #4622: Race condition in pool_available metric causes negative values during network instability

mkleczek:
**DISCLAIMER:**

Right now metrics observation handler does not track database connections but updates a single Gauge based on HasqlPoolObs events.

This is problematic because Hasql pool reports various connection events in various states that make it impossible to predict the state change from the received event. The connection state machine is not simple and to precisely report the number of connections in various states, it is necessary to track their lifecycles.

Fixes #4622

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ All notable changes to this project will be documented in this file. From versio
 ### Fixed
 
 - Fix unnecessary connection pool flushes during schema cache reloading by @mkleczek in #4645
+- Fix race condition in pool_available metric causing negative values during network instability by @mkleczek in #4622
 
 ## [14.9] - 2026-04-10
 
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -163,6 +163,7 @@ library
                     , stm-hamt                  >= 1.2 && < 2
                     , focus                     >= 1.0 && < 2
                     , some                      >= 1.0.4.1 && < 2
+                    , uuid                      >= 1.3 && < 2
                       -- -fno-spec-constr may help keep compile time memory use in check,
                       --   see https://gitlab.haskell.org/ghc/ghc/issues/16017#note_219304
                       -- -optP-Wno-nonportable-include-path
--- a/src/PostgREST/Metrics.hs
+++ b/src/PostgREST/Metrics.hs
@@ -5,7 +5,10 @@ Description : Metrics based on the Observation module. See Observation.hs.
 -}
 module PostgREST.Metrics
   ( init
+  , ConnTrack
+  , ConnStats (..)
   , MetricsState (..)
+  , connectionCounts
   , observationMetrics
   , metricsToText
   ) where
@@ -19,13 +22,18 @@ import qualified Prometheus.Metric.GHC as PMG
 
 import PostgREST.Observation
 
-
-import Protolude
+import           Control.Arrow      ((&&&))
+import           Data.Bitraversable (bisequenceA)
+import           Data.Tuple.Extra   (both)
+import           Data.UUID          (UUID)
+import qualified Focus
+import           Protolude
+import qualified StmHamt.SizedHamt  as SH
 
 data MetricsState =
   MetricsState {
     poolTimeouts         :: Counter,
-    poolAvailable        :: Gauge,
+    connTrack            :: ConnTrack,
     poolWaiting          :: Gauge,
     poolMaxSize          :: Gauge,
     schemaCacheLoads     :: Vector Label1 Counter,
@@ -40,7 +48,7 @@ init configDbPoolSize = do
   whenM getRTSStatsEnabled $ void $ register PMG.ghcMetrics
   metricState <- MetricsState <$>
     register (counter (Info "pgrst_db_pool_timeouts_total" "The total number of pool connection timeouts")) <*>
-    register (gauge (Info "pgrst_db_pool_available" "Available connections in the pool")) <*>
+    register (Metric ((identity &&& dbPoolAvailable) <$> connectionTracker)) <*>
     register (gauge (Info "pgrst_db_pool_waiting" "Requests waiting to acquire a pool connection")) <*>
     register (gauge (Info "pgrst_db_pool_max" "Max pool connections")) <*>
     register (vector "status" $ counter (Info "pgrst_schema_cache_loads_total" "The total number of times the schema cache was loaded")) <*>
@@ -50,20 +58,28 @@ init configDbPoolSize = do
     register (counter (Info "pgrst_jwt_cache_evictions_total" "The total number of JWT cache evictions"))
   setGauge (poolMaxSize metricState) (fromIntegral configDbPoolSize)
   pure metricState
+  where
+    dbPoolAvailable = (pure . noLabelsGroup (Info "pgrst_db_pool_available" "Available connections in the pool") GaugeType . calcAvailable <$>) . connectionCounts
+      where
+        calcAvailable = liftA2 (-) connected inUse
+    toSample name labels = Sample name labels . encodeUtf8 . show
+    noLabelsGroup info sampleType = SampleGroup info sampleType . pure . toSample (metricName info) mempty
 
 -- Only some observations are used as metrics
 observationMetrics :: MetricsState -> ObservationHandler
 observationMetrics MetricsState{..} obs = case obs of
   PoolAcqTimeoutObs -> do
     incCounter poolTimeouts
-  (HasqlPoolObs (SQL.ConnectionObservation _ status)) -> case status of
-     SQL.ReadyForUseConnectionStatus _ -> do
-      incGauge poolAvailable
-     SQL.InUseConnectionStatus        -> do
-      decGauge poolAvailable
-     SQL.TerminatedConnectionStatus  _ -> do
-      decGauge poolAvailable
-     SQL.ConnectingConnectionStatus -> pure ()
+  -- Handle pool observations with connection tracking
+  -- this is necessary because it is not possible
+  -- to accurately maintain open/in use conneciton counts
+  -- statelessly based only on pool observation events.
+  -- The reason is that hasql-pool emits TerminatedConnectionStatus
+  -- both for connections successfully established and failed when connecting.
+  -- When receiving TerminatedConnectionStatus we have to find out
+  -- if we can decrement established connection count. To do that we have to track
+  -- established connections.
+  (HasqlPoolObs sqlObs) -> trackConnections connTrack sqlObs
   PoolRequest ->
     incGauge poolWaiting
   PoolRequestFullfilled ->
@@ -81,3 +97,28 @@ observationMetrics MetricsState{..} obs = case obs of
 
 metricsToText :: IO LBS.ByteString
 metricsToText = exportMetricsAsText
+
+data ConnStats = ConnStats {
+    connected :: Int,
+    inUse     :: Int
+} deriving (Eq, Show)
+
+data ConnTrack = ConnTrack { connTrackConnected :: SH.SizedHamt UUID, connTrackInUse :: SH.SizedHamt UUID }
+
+connectionTracker :: IO ConnTrack
+connectionTracker =  ConnTrack <$> SH.newIO <*> SH.newIO
+
+trackConnections :: ConnTrack -> SQL.Observation -> IO ()
+trackConnections ConnTrack{..} (SQL.ConnectionObservation uuid status) = case status of
+  SQL.ReadyForUseConnectionStatus _ -> atomically $
+    SH.insert identity uuid connTrackConnected *>
+    SH.focus Focus.delete identity uuid connTrackInUse
+  SQL.TerminatedConnectionStatus _ -> atomically $
+    SH.focus Focus.delete identity uuid connTrackConnected *>
+    SH.focus Focus.delete identity uuid connTrackInUse
+  SQL.InUseConnectionStatus -> atomically $
+    SH.insert identity uuid connTrackInUse
+  _ -> mempty
+
+connectionCounts :: ConnTrack -> IO ConnStats
+connectionCounts = atomically . fmap (uncurry ConnStats) . bisequenceA . both SH.size . (connTrackConnected &&& connTrackInUse)
--- a/test/observability/Observation/MetricsSpec.hs
+++ b/test/observability/Observation/MetricsSpec.hs
@@ -6,17 +6,21 @@
 
 module Observation.MetricsSpec where
 
-import           Data.List             (lookup)
-import           Network.Wai           (Application)
+import           Data.List              (lookup)
+import qualified Hasql.Pool.Observation as SQL
+import           Network.Wai            (Application)
 import           ObsHelper
-import qualified PostgREST.AppState    as AppState
-import           PostgREST.Config      (AppConfig (configDbSchemas))
-import qualified PostgREST.Metrics     as Metrics
+import qualified PostgREST.AppState     as AppState
+import           PostgREST.Config       (AppConfig (configDbSchemas))
+import           PostgREST.Metrics      (ConnStats (..),
+                                         MetricsState (..),
+                                         connectionCounts)
 import           PostgREST.Observation
-import           Prometheus            (getCounter, getVectorWith)
-import           Protolude
-import           Test.Hspec            (SpecWith, describe, it)
-import           Test.Hspec.Wai        (getState)
+import           Prometheus             (getCounter, getVectorWith)
+import           Test.Hspec             (SpecWith, describe, it)
+import           Test.Hspec.Wai         (getState)
+
+import Protolude
 
 spec :: SpecWith (SpecState, Application)
 spec = describe "Server started with metrics enabled" $ do
@@ -71,9 +75,40 @@ spec = describe "Server started with metrics enabled" $ do
         -- (there should be none but we need to verify that)
         threadDelay $ 1 * sec
 
+  -- The test verifies we properly count in use connections
+  -- The idea is to fork a worker thread that
+  -- borrows connection from the pool and waits for a signal to release it
+  -- Main thread checks that
+  -- in use connections counter is incremented by worker
+  -- then it signals the worker to release the connection
+  -- and finally verifies that in use connection counter is back to original value
+  it "Should track in use connections" $ do
+    SpecState{specAppState = appState, specMetrics = metrics, specObsChan} <- getState
+    let waitFor = waitForObs specObsChan
+
+    liftIO $ checkState' metrics [
+        -- we expect in use connections to be the same once finished
+        inUseConnections (+ 0)
+      ] $ do
+        signal <- newEmptyMVar
+        -- make sure waiting thread is signaled
+        (`finally` tryPutMVar signal ()) $
+          -- expecting one more connection in use
+          checkState' metrics [
+            inUseConnections (+ 1)
+          ] $ do
+            -- start a thread hanging on a single connection until signaled
+            void $ forkIO $ void $ AppState.usePool appState $ liftIO (readMVar signal)
+            -- main thread waits for ConnectionObservation with InUseConnectionStatus
+            -- after which used connections count should be incremented
+            waitFor (1 * sec) "InUseConnectionStatus" $ \x -> [ o | o@(HasqlPoolObs (SQL.ConnectionObservation _ SQL.InUseConnectionStatus)) <- pure x]
+
+        -- hanging thread was signaled and should return the connection
+        waitFor (1 * sec) "ReadyForUseConnectionStatus" $ \x -> [ o | o@(HasqlPoolObs (SQL.ConnectionObservation _ (SQL.ReadyForUseConnectionStatus _))) <- pure x]
 
   where
     -- prometheus-client api to handle vectors is convoluted
     schemaCacheLoads label = expectField @"schemaCacheLoads" $
       fmap (maybe (0::Int) round . lookup label) . (`getVectorWith` getCounter)
+    inUseConnections = expectField @"connTrack" ((inUse <$>) . connectionCounts)
     sec = 1000000
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -2181,10 +2181,6 @@ def test_vary_default_header_set(defaultenv):
         assert response.headers["Vary"] == "Accept, Prefer, Range"
 
 
-@pytest.mark.xfail(
-    reason="pgrst_db_pool_available should not go negative on pg network failures",
-    strict=True,
-)
 def test_positive_pool_metric(defaultenv):
     "When a network failure is caused on the pg connection, pgrst_db_pool_available stays positive"
 


steve-chavez:
> Fixes https://github.com/PostgREST/postgrest/issues/4622

Fix should have an entry on the CHANGELOG

src/PostgREST/Metrics.hs
```diff
@@ -77,3 +85,28 @@ observationMetrics MetricsState{..} obs = case obs of
...
+trackConnections ConnTrack{..} (SQL.ConnectionObservation uuid status) = case status of
+  SQL.ReadyForUseConnectionStatus -> atomically $
+    SH.insert identity uuid connTrackConnected *>
+    SH.focus Focus.delete identity uuid connTrackInUse
+  SQL.TerminatedConnectionStatus _ -> atomically $
+    SH.focus Focus.delete identity uuid connTrackConnected *>
+    SH.focus Focus.delete identity uuid connTrackInUse
+  SQL.InUseConnectionStatus -> atomically $
+    SH.insert identity uuid connTrackInUse
+  _ -> mempty
```
steve-chavez: Could there be noticeable overhead now that we have the insert/focus logic?

Doesn't look like it from the [mixed loadtest](https://github.com/PostgREST/postgrest/actions/runs/22410711782?pr=4672#summary-64883118221) but want to make sure.

src/PostgREST/Metrics.hs
```diff
@@ -77,3 +85,28 @@ observationMetrics MetricsState{..} obs = case obs of
...
+
+connectionCounts :: ConnTrack -> IO ConnStats
+connectionCounts = atomically . fmap (uncurry ConnStats) . bisequenceA . both SH.size . (connTrackConnected &&& connTrackInUse)
```
steve-chavez, line 124: This is purely used for testing right? If so, a comment here would be good for clarity.

mkleczek: > This is purely used for testing right? If so, a comment here would be good for clarity.

@steve-chavez  It is not. The actual Metric is implementation is based on ConnTrack:
https://github.com/mkleczek/postgrest/blob/7e9df6416195b3d9c534c1f1c4aeb39865869f45/src/PostgREST/Metrics.hs#L59

Also - this function is crucial for #4643 - see: https://github.com/mkleczek/postgrest/blob/f3e06920544b88d26809e75d2652d2ed6a4ef8f3/src/PostgREST/AppState.hs#L421

src/PostgREST/Metrics.hs
steve-chavez: > Could there be noticeable overhead now that we have the insert/focus logic?

Not really because this calculation is outside of the request hot path, `trackConnections` is executed only when `HasqlPoolObs` is emitted and that only happens on

https://github.com/PostgREST/postgrest/blob/0e820b9ac569f6907fa230043c5bbd4eea0751e0/src/PostgREST/AppState.hs#L158-L167

test/observability/Observation/MetricsSpec.hs
```diff
@@ -71,9 +74,33 @@ spec = describe "Server started with metrics enabled" $ do
         -- (there should be none but we need to verify that)
         threadDelay $ 1 * sec
 
+  it "Should track in use connections" $ do
```
steve-chavez, line 85: Q: Does this reproduce the problem on https://github.com/PostgREST/postgrest/issues/4622? If not, could we add a new repro test or is that too complicated?

steve-chavez:
@mkleczek Since this is a `fix`, let's change the commit, PR title and add a fix entry in the CHANGELOG.

test/observability/Observation/MetricsSpec.hs
mkleczek: > Q: Does this reproduce the problem on #4622? If not, could we add a new repro test or is that too complicated?

As far as I understand it might be possible to reproduce once we have `toxiproxy` in place. The issue happens when there connectivity issues.

steve-chavez:
@mkleczek Could you please address the feedback [here](https://github.com/PostgREST/postgrest/pull/4672#issuecomment-3962630756)? I've repeated the same twice now :weary:

mkleczek:
> @mkleczek Could you please address the feedback [here](https://github.com/PostgREST/postgrest/pull/4672#issuecomment-3962630756)? I've repeated the same twice now 😩

Done

test/observability/Observation/MetricsSpec.hs
steve-chavez: For that perhaps you could restart https://github.com/PostgREST/postgrest/pull/4674? Should be easier I think now that we don't have `slocat`.

I also wonder if there's a simpler/focused way by just killing the connection like mentioned on https://github.com/PostgREST/postgrest/issues/1766#issuecomment-1004391470

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file. From versio
 ### Fixed
 
 - Fix unnecessary connection pool flushes during schema cache reloading by @mkleczek in #4645
+- Implement connection tracking in metrics by @mkleczek in #4672
```
wolfgangwalther: This changelog reads like an addition, a new feature. This is in the "fixed" section, so this should describe which bug is fixed for the user. As a user I have no idea here.

mkleczek: > This changelog reads like an addition, a new feature. This is in the "fixed" section, so this should describe which bug is fixed for the user. As a user I have no idea here.

Updated the entry text based on reported issue title. Not the reference as I remember @steve-chavez requesting in another pull request to put PR (not issue) references in the changelog.

Not sure what the policy is.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file. From versio
 ### Fixed
 
 - Fix unnecessary connection pool flushes during schema cache reloading by @mkleczek in #4645
+- Fix race condition in pool_available metric causes negative values during network instability by @mkleczek in #4672
```
steve-chavez:
```suggestion
- Fix race condition in pool_available metric cauing negative values during network instability by @mkleczek in #4622 
```

CHANGELOG.md
steve-chavez: We've been putting the issue when there is one available otherwise just the PR number.

CHANGELOG.md

CHANGELOG.md

postgrest.cabal
```diff
@@ -161,6 +161,7 @@ library
                     , stm-hamt                  >= 1.2 && < 2
                     , focus                     >= 1.0 && < 2
                     , some                      >= 1.0.4.1 && < 2
+                    , uuid                      >= 1.3 && < 2
```
wolfgangwalther, line 164: Still would like to check this dependency for whether it cross-compiles without dependencies on TemplateHaskell (see https://github.com/PostgREST/postgrest/issues/3281#issuecomment-4283355556).

(my builds will take a while, will report back)

test/observability/Observation/MetricsSpec.hs
wolfgangwalther: For a case like this, I wouldn't block on adding an exactly-reproducing test of the original issue: We really changed the entire approach to calculating this metric, from simply reacting to certain events to tracking each connection. I find it's unlikely that someone accidentally introduces a regression here.

That being said, I think we absolutely need some commentary in `Metrics.hs` about *why* we are doing the connection tracking and which problems this prevents. With such commentary, someone looking at this later will know what to look for when rewriting the whole thing again.

mkleczek: > For that perhaps you could restart #4674? Should be easier I think now that we don't have `slocat`.
> 

See #4833 #4836

postgrest.cabal
wolfgangwalther: Builds fine!

test/observability/Observation/MetricsSpec.hs
mkleczek: > 
> That being said, I think we absolutely need some commentary in `Metrics.hs`

Done (see inline comment in `Metrics.observationMetrics`). I am not 100% sure if the comment is in the right place so happy to move it if you think it is necessary.

src/PostgREST/Metrics.hs
```diff
@@ -46,20 +55,28 @@ init configDbPoolSize = do
...
-     SQL.ConnectingConnectionStatus -> pure ()
+  -- Handle pool observations with connection tracking
+  -- this is necessary because it is not possible
+  -- to accurately maintain open/in use conneciton counts
+  -- statelessly based only on pool observation events.
+  -- The reason is that hasql-pool emits TerminatedConnectionStatus
+  -- both for connections successfully established and failed when connecting.
+  -- When receiving TerminatedConnectionStatus we have to find out
+  -- if we can decrement established connection count. To do that we have to track
+  -- established connections.
```
wolfgangwalther, line 78: Thanks for the comment - this helps a lot.

Q: Would this reasoning be different once we vendor hasql, including hasql-pool, and could change the respective events? Would this make the fix here simpler, because we could keep using the same events?

mkleczek: > Thanks for the comment - this helps a lot.
> 
> Q: Would this reasoning be different once we vendor hasql, including hasql-pool, and could change the respective events? Would this make the fix here simpler, because we could keep using the same events?

Yes.
Quite frankly - `hasql-pool` should expose the metrics itself because it already tracks this information anyway. Publishing events is the wrong abstraction here.

EDIT: I mean `hasql-pool` should expose functions providing the actual numbers. Exposing them as `Prometheus` metrics should be done similar to how `ConnTrack` is exposed in this PR.

test/observability/Observation/MetricsSpec.hs
steve-chavez: > For a case like this, I wouldn't block on adding an exactly-reproducing test of the original issue: We really changed the entire approach to calculating this metric, from simply reacting to certain events to tracking each connection.

Yeah, I was thinking we should take this chance to add the required infra to test otherwise we'll end up with a "test debt" (like on https://github.com/PostgREST/postgrest/issues/1766).

But I guess we still have https://github.com/PostgREST/postgrest/issues/4670 as another chance to test an unreliable network.

So I won't block here if you approve here @wolfgangwalther

wolfgangwalther: > We really changed the entire approach to calculating this metric, from simply reacting to certain events to tracking each connection. I find it's unlikely that someone accidentally introduces a regression here.

In light of https://github.com/PostgREST/postgrest/pull/4672#discussion_r3120365020, I think there is a chance that we'll rewrite this whole thing again when we have `hasql-pool` vendored. Thus, I agree with:

> Yeah, I was thinking we should take this chance to add the required infra to test otherwise we'll end up with a "test debt" (like on #1766).

test/observability/Observation/MetricsSpec.hs
```diff
@@ -71,9 +75,33 @@ spec = describe "Server started with metrics enabled" $ do
...
+            inUseConnections (+ 1)
+          ] $ do
+            -- start a thread hanging on a single connection until signaled
+            void $ forkIO $ void $ AppState.usePool appState $ liftIO (readMVar signal)
+            -- main thread waits for ConnectionObservation with InUseConnectionStatus
+            -- after which used connections count should be incremented
+            waitFor (1 * sec) "InUseConnectionStatus" $ \x -> [ o | o@(HasqlPoolObs (SQL.ConnectionObservation _ SQL.InUseConnectionStatus)) <- pure x]
+
+        -- hanging thread was signaled and should return the connection
+        waitFor (1 * sec) "ReadyForUseConnectionStatus" $ \x -> [ o | o@(HasqlPoolObs (SQL.ConnectionObservation _ (SQL.ReadyForUseConnectionStatus _))) <- pure x]
```
wolfgangwalther, line 107: I'm looking at this test-case.. and it just feels like an imperative language like python is much better to represent such a step-by-step test.

I can't even make full sense of the test, but I believe pytest + toxiproxy + metrics endpoint should allow the same test, right?

mkleczek: > I'm looking at this test-case.. and it just feels like an imperative language like python is much better to represent such a step-by-step test.
> 

I personally prefer statically typed language and many people find Haskell [The Best Imperative Language](https://stackoverflow.com/questions/6622524/why-is-haskell-sometimes-referred-to-as-best-imperative-language)

Especially in terms of concurrency support (and it is important in this test) Python is very far behind Haskell.

> I can't even make full sense of the test, but I believe pytest + toxiproxy + metrics endpoint should allow the same test, right?

Not really (well... you could but it is not very convenient).

`Toxiproxy` is not used in this test so does not matter.

The idea of this test is to start a worker thread that will borrow a connection from the pool and keep it used **until signaled** by main thread.
After spawning the worker thread, main thread first waits for `InUseConnectionStatus` so that it is sure connection has been indeed borrowed from the pool. Then it checks it is properly accounted for in the metrics. Then it signals the worker thread, which in turn releases the connection and exits. Main thread then makes sure `ReadyForUseConnectionStatus` was published and then verifies connection release was accounted for in the metrics.

This is a "white box" test - we don't really test PostgREST as a whole here but only `AppState.usePool` function behavior.

There is no easy way to do the same in Python. We would need to somehow use the database advisory locks to be able to achieve "borrow connection from the pool and wait for a signal before releasing it". We also need to be able to wait for specific observations in the test. This was enabled for HSpec tests in #4671 and discussed with @steve-chavez and @taimoorzaeem 

It is probably doable in Python (everything is doable in Turing complete language) but would require a lot of preparation and specific fixtures. It would also require accessing database directly from Python which we don't do in io tests.

steve-chavez: > Toxiproxy is not used in this test so does not matter.

But we do need it to prove the fix for https://github.com/PostgREST/postgrest/issues/4622 as done in https://github.com/PostgREST/postgrest/pull/4855 right?

> I can't even make full sense of the test

I also don't find this readable, if the test in https://github.com/PostgREST/postgrest/pull/4855 is enough maybe the test here is not really necessary?

(Not arguing for python or haskell, but seeing if is possible to save us some lines of code)

mkleczek: > If the test in #4855 is enough maybe the test here is not really necessary?
> 

IMHO this test is necessary regardless. The test in #4855 tests if `poolAvailable` metric does not become negative during adverse network conditions.
This test is about `inUse` count **during** `AppState.usePool` execution.
It is something different and I would say this test is needed to simply make sure we properly calculate connection counts.

mkleczek:
@steve-chavez @wolfgangwalther - what do you think is required to push this forward?

#4622 is being now reported by our support and it is becoming urgent to fix it.

test/observability/Observation/MetricsSpec.hs
wolfgangwalther: > I personally prefer statically typed language and many people find Haskell [The Best Imperative Language](https://stackoverflow.com/questions/6622524/why-is-haskell-sometimes-referred-to-as-best-imperative-language)

From that link:

> However, the false half is the syntax. I find Haskell pretty verbose and awkward to use in an imperative style.

Right on point for me.

> `Toxiproxy` is not used in this test so does not matter.

One of the problems the io tests have is the big dependence on *timing*. Wait a second here, two there, whatever. Of course, we could write it with this approach in pytest as well - but I don't think that's a good idea.

So I thought of using toxiproxy to allow us to do this *properly*, i.e. by pausing the underlying connections, checking the state, then resuming / finishing the connection, checking the state again. But after reading the remainder of your comment about advisory locks, I see that we probably need these to even get to the right spot to pause the connection - and once we're there, we don't need toxiproxy for that specific case anymore.

wolfgangwalther:
> what do you think is required to push this forward?

This does not introduce any new test infrastructure that I'd be opposed to, so I won't block on the long term vision of how our tests should be structured. It uses the existing infrastructure. I might not like the way the test is written, but I don't see a need to block on that either.

To be clear, the question in https://github.com/PostgREST/postgrest/pull/4672#discussion_r3190883302 was asked to get a feeling of how things could be done differently, if we had better test infrastructure elsewhere - and *not* to block this PR's progress.

Imho the only previously blocking comment is https://github.com/PostgREST/postgrest/pull/4672#discussion_r3146244325. Now, since I wrote that comment, I started a major discussion on how we should test in general, which is blocking the *other*, test-infra related, issues/PRs. I don't think we should hold this PR hostage to that either.

TLDR: No blockers for me.

test/observability/Observation/MetricsSpec.hs
steve-chavez: > However, the false half is the syntax. I find Haskell pretty verbose and awkward to use in an imperative style.
> Right on point for me.

It's the same for me but there's also another angle here. @mkleczek On https://github.com/PostgREST/postgrest/pull/4672#issuecomment-4448644124, you mention "support". I've had the experience of support + customers (i.e. users of PostgREST) challenging if a bug is fixed or not. Pointing them to a clear black box python test quickly settled the debate. It won't be the same in Haskell. 

> But after reading the remainder of your comment about advisory locks, I see that we probably need these to even get to the right spot to pause the connection - and once we're there, we don't need toxiproxy for that specific case anymore.

Can we use the advisory locks to achieve a negative connection metric?

steve-chavez:
> what do you think is required to push this forward?

It's gonna be really confusing when we look back in history and we say we fix https://github.com/PostgREST/postgrest/issues/4622 and there isn't a precise test proving it (the current test does not).

Let's not set a precedent here that can later hurt us with tech debt, so we should first clear [the above thread](https://github.com/PostgREST/postgrest/pull/4672#discussion_r3243027363).

test/observability/Observation/MetricsSpec.hs
steve-chavez: Not opposed to keep the current Haskell test. But perhaps we can add another one python io test that proves the Prometheus metrics are correct?

> It would also require accessing database directly from Python which we don't do in io tests.

We do btw, a couple of choices:

1. With psql
https://github.com/PostgREST/postgrest/blob/ed0d9dcd597f29412579b7cce82c35ba3e57fc16/test/io/test_io.py#L703-L722

2. Indirectly with an rpc
https://github.com/PostgREST/postgrest/blob/ed0d9dcd597f29412579b7cce82c35ba3e57fc16/test/io/test_io.py#L426

mkleczek: > It's the same for me but there's also another angle here. @mkleczek On [#4672 (comment)](https://github.com/PostgREST/postgrest/pull/4672#issuecomment-4448644124), you mention "support". I've had the experience of support + customers (i.e. users of PostgREST) challenging if a bug is fixed or not. Pointing them to a clear black box python test quickly settled the debate. It won't be the same in Haskell.

From my experience they don't really care about automated tests you have in your codebase, what language they are written in and whether they instantiate a program as a "black box" or test the module containing the bug - the whole source code and its organization is a black box to them (and they are not programmers most of the time).

They do care that you declare the bug is fixed though, which seems needs to wait.

> 
> > But after reading the remainder of your comment about advisory locks, I see that we probably need these to even get to the right spot to pause the connection - and once we're there, we don't need toxiproxy for that specific case anymore.
> 
> Can we use the advisory locks to achieve a negative connection metric?

No, we can't. This is about **connection errors** - ie. failing to connect to the database server.

------

I'm sorry guys - I've raised #4855 to have a reproducer - I don't have any other idea on how to reproduce it, so if it is not mergeable then it looks like we have to live with #4622 until someone implements a different reproducer (possibly using different testing infrastructure and language). I am happy to wait and rebase this PR once such a reproducer is merged into `main`.

@steve-chavez @wolfgangwalther - you are maintainers of this project and you decide on the overall direction and whether PRs are aligned with your vision. I personally find it discouraging, though, that these grand vision discussions as in #4868 are really just blocking improvements and bug fixing.

I personally prefer continuous improvement and implementing sometimes imperfect solutions to current problems instead of blocking them just because they are imperfect - but that's my personal view.

wolfgangwalther: > I personally find it discouraging, though, that these grand vision discussions as in #4868 are really just blocking improvements and bug fixing.

Just to be clear: I specifically said this discussion is *not* blocking for me. What is blocking for Steve is essentially the backlog of #1766.

Of course, this won't help you in any way and I fully understand that this sucks. :/

wolfgangwalther:
> It's gonna be really confusing when we look back in history and we say we fix #4622 and there isn't a precise test proving it (the current test does not).
> 
> Let's not set a precedent here that can later hurt us with tech debt, so we should first clear [the above thread](https://github.com/PostgREST/postgrest/pull/4672#discussion_r3243027363).

The thread you linked is about the currently existing test, so that doesn't quite match the first sentence about a missing test. If you're concerned about a missing test, then we need to clear https://github.com/PostgREST/postgrest/pull/4672#discussion_r3146244325.

I'd still say the situation now is different compared to when #1766 happened - we are actively working on improving the test situation and we have an open PR to track the addition of the herein-missing test. Thus, I'd say the risk of this getting forgotten is much smaller than earlier.

I'd say we should go ahead with this.

wolfgangwalther (approved):

steve-chavez:
@wolfgangwalther Let's not merge because I have a much simpler test almost ready for PR, let's merge this after that.

steve-chavez (changes requested):

wolfgangwalther:
No worries, I don't intend to merge. I just wanted to make my implicit approval explicit. I am well aware that you still have a thread open (this is now actually blocking the merge as well) - and I'm not just going to override you and resolve that thread. That's for you to decide :)

test/observability/Observation/MetricsSpec.hs
steve-chavez: Added a simpler test on https://github.com/PostgREST/postgrest/pull/4928

mkleczek: Rebased on top of it.

Note that #4928 is equivalent to #4855 using `nginx` instead of `toxiproxy` (let's leave the discussion about merits of each aside).

This one is a different test though - it makes sure we properly track in use connections.

steve-chavez (approved):

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ All notable changes to this project will be documented in this file. From versio
 ### Fixed
 
 - Fix unnecessary connection pool flushes during schema cache reloading by @mkleczek in #4645
+- Fix race condition in pool_available metric causing negative values during network instability by @mkleczek in #4622
```
steve-chavez, line 55: @mkleczek This was added in an old version `Fixed` section https://github.com/PostgREST/postgrest/blob/main/CHANGELOG.md#fixed-2 😕

mkleczek: Ehh... rebasing changelog is inherently tricky. My bad.

Raised #4942

steve-chavez:
I think we should backport this, it will require:

- https://github.com/PostgREST/postgrest/commit/70327cf8694f306d602338da7d58a67b68a004ad
- https://github.com/PostgREST/postgrest/commit/6220ab3f3199688364409f0f2a00a510259f0cef
- https://github.com/PostgREST/postgrest/commit/1eba9855c11f00c6f1749932c78651584c88c7d7

mkleczek:
> I think we should backport this, it will require:
> 
> * [70327cf](https://github.com/PostgREST/postgrest/commit/70327cf8694f306d602338da7d58a67b68a004ad)
> * [6220ab3](https://github.com/PostgREST/postgrest/commit/6220ab3f3199688364409f0f2a00a510259f0cef)
> * [1eba985](https://github.com/PostgREST/postgrest/commit/1eba9855c11f00c6f1749932c78651584c88c7d7)

Hmm... most probably the test in `observability` test suite is a problem for backporting.

Maybe we should split it into 2 PRs then (separate the test in `MetricsSpec`)?

steve-chavez:
> Maybe we should split it into 2 PRs then (separate the test in MetricsSpec)?

Yup, sounds good.

taimoorzaeem:
I think we just follow the order in the git log. It involves 5 commits in order:

- 70327cf8694f306d602338da7d58a67b68a004ad
- 8868d132145cf80b0db9f9a2aa204b91ac1e5da6
- a29739107109e96beada1e175acff4ba56968c9d
- 1eba9855c11f00c6f1749932c78651584c88c7d7
- 6220ab3f3199688364409f0f2a00a510259f0cef

Let me try doing this in 1 PR, so it's easier to revert, just in case.

PR #4673, opened 2026-02-26
add: set keepalives options in listener connection string

mkleczek:
**DISCLAIMER:**

This change adds keepalive options to listener connections string. The value of keepalives_idle is adjusted using a simple algorithm based on idle time tracking.

Keeping it as draft until it is clear how to test it.


steve-chavez:
Related to https://github.com/PostgREST/postgrest/issues/4670

PR #4679, opened 2026-02-27
refactor: get rid of redundant SchemaCacheSummaryObs

mkleczek:
There is unnecessary coupling between observation messages and emited log entries. This causes schema loading logic to emit redundant events: SchemaCacheSummaryObs and SchemaCacheLoadedObs.

Logically - we want to emit a single event containing both summary and timing information. How it is logged is a different matter and should be decoupled.
    
This commit
* changes observationMessage function returning Text to observationMessages returning [Text] so that it is possible to return multiple (or zero) messages to log based on an observation event
* Removes SchemaCacheSummaryObs constructor from Observation type and adds summary text to SchemaCacheLoadedObs

Stacked on top of #4677

### DIFF
--- a/src/PostgREST/AppState.hs
+++ b/src/PostgREST/AppState.hs
@@ -416,8 +416,7 @@ retryingSchemaCacheLoad appState@AppState{stateObserver=observer, stateMainThrea
           putSCacheStatus appState SCPending
           putSchemaCache appState $ Just sCache
           observer $ SchemaCacheQueriedObs resultTime
-          (t, _) <- timeItT $ observer $ SchemaCacheSummaryObs $ showSummary sCache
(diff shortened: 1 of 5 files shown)


steve-chavez (approved):
> Logically - we want to emit a single event containing both summary and timing information. How it is logged is a different matter and should be decoupled.

Agree, I made two Observations because of the inability to have multiple lines. This is much better.

mkleczek:
> > Logically - we want to emit a single event containing both summary and timing information. How it is logged is a different matter and should be decoupled.
> 
> Agree, I made two Observations because of the inability to have multiple lines. This is much better.

@steve-chavez cool - can you merge it? Would rebase the next stacked one.

PR #4680, opened 2026-02-27
refactor: standardize query logging to follow code logging other observations

mkleczek:
This commit addresses the following `TODO` in observationMessage function:
`TODO pending refactor: The logic for printing the query cannot be done here. Join the observationMessages function into observationLogger to avoid this mempty.`

Changing observationMessage to observationMessages (that returns [Text] instead of Text) made it possible to address issue documented as TODO and harmonize query logging with other observations.

Stacked on top of #4679


mkleczek:
@steve-chavez rebased and ready to merge.

PR #4684, opened 2026-03-03
fix(error): leaking table and function names when calculating hint

### LINKED ISSUE #4675: Fuzzy distance is too high for table hint
## Problem

The hints were meant to help the user with typos, but currently:

```http
GET /users?select=* 

{
  "code": "PGRST205",
  "details": null,
  "hint": "Perhaps you meant the table 'public.markers'",
  "message": "Could not find the table 'public.users' in the schema cache"
}

GET /employee_locations?select=*

{
  "code": "PGRST205",
  "details": null,
  "hint": "Perhaps you meant the table 'public.employees'",
  "message": "Could not find the table 'public.employee_locations' in the schema cache"
}

GET /clientsxxxxxxxxxxxxxx

{
  "code":"PGRST205",
  "details":null,
  "hint":"Perhaps you meant the table 'test.clients'",
  "message":"Could not find the table 'test.clientsxxxxxxxxxxxxxx' in the schema cache"
} 
```

And finally we don't get a hint with:

```http
GET /clientsxxxxxxxxxxxxxx

{
  "code":"PGRST205",
  "details":null,
  "hint":null,
  "message":"Could not find the table 'test.clientsxxxxxxxxxxxxxxx' in the schema cache"
}
```

So the fuzzy distance is too high. It looks like it's 14.

This could be considered a vulnerability since even with `openapi-mode=follow-privileges` it can allow a user to enumerate tables in a schema, even when the user doesn't have any privileges on them.

## Solution

Use a max fuzzy distance of 3. Same as on PostgreSQL ([ref](https://github.com/postgres/postgres/blob/851f6649cc18c4b482fa2b6afddb65b35d035370/src/backend/parser/parse_relation.c#L72)).

## Notes

- Same problem should be present on functions and embeddings

taimoorzaeem:
Increase similarity score to `0.75` from `0.33` for table and functions
error hint.

Closes #4675.

### DIFF
--- a/src/PostgREST/Error.hs
+++ b/src/PostgREST/Error.hs
@@ -338,7 +338,7 @@ noRelBetweenHint parent child schema allRels = ("Perhaps you meant '" <>) <$>
 -- Just "Perhaps you meant to call the function api.test"
 --
 -- >>> noRpcHint "api" "other" [] procs []
--- Just "Perhaps you meant to call the function api.another"
+-- Nothing
 --
 -- >>> noRpcHint "api" "noclosealternative" [] procs []
 -- Nothing
@@ -375,8 +375,8 @@ noRpcHint schema procName params allProcs overloadedProcs =
     -- E.g. ["val", "param", "name"] into "(name, param, val)"
     listToText       = ("(" <>) . (<> ")") . T.intercalate ", " . sort
     possibleProcs
-      | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
-      | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params)
+      | null overloadedProcs = getFuzzyHint HintProcedure fuzzySetOfProcs procName
+      | otherwise            = (procName <>) <$> getFuzzyHint HintParams fuzzySetOfParams (listToText params)
 
 -- |
 -- Do a fuzzy search in all tables in the same schema and return closest result
@@ -384,7 +384,21 @@ tableNotFoundHint :: Text -> Text -> SchemaCache -> Maybe Text
 tableNotFoundHint schema tblName SchemaCache{dbTablesFuzzyIndex}
   = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
     where
-      perhapsTable = (`Fuzzy.getOne` tblName) =<< HM.lookup schema dbTablesFuzzyIndex
+      perhapsTable = (\fuzzySet -> getFuzzyHint HintTable fuzzySet tblName) =<< HM.lookup schema dbTablesFuzzyIndex
+
+data HintType
+  = HintTable
+  | HintProcedure
+  | HintParams
+
+-- | Get hint using Fuzzy Search with at least 0.75 similarity score
+getFuzzyHint :: HintType -> Fuzzy.FuzzySet -> Text -> Maybe Text
+getFuzzyHint hintType =
+  let minScore = 0.75 :: Double -- used for table and procedure name hints
+  in case hintType of
+    HintTable     -> Fuzzy.getOneWithMinScore minScore
+    HintProcedure -> Fuzzy.getOneWithMinScore minScore
+    HintParams    -> Fuzzy.getOne -- For params, we stick to `getOne` which defaults to 0.33 min score, not a security risk to reveal params
 
 compressedRel :: Relationship -> JSON.Value
 -- An ambiguousness error cannot happen for computed relationships TODO refactor so this mempty is not needed
--- a/test/spec/Feature/Query/DeleteSpec.hs
+++ b/test/spec/Feature/Query/DeleteSpec.hs
@@ -116,7 +116,7 @@ spec =
       it "fails with 404" $
         request methodDelete "/foozle?id=eq.101" [] ""
           `shouldRespondWith`
-          [json| {"code":"PGRST205","details":null,"hint":"Perhaps you meant the table 'test.foo'","message":"Could not find the table 'test.foozle' in the schema cache"} |]
+          [json| {"code":"PGRST205","details":null,"hint":null,"message":"Could not find the table 'test.foozle' in the schema cache"} |]
           { matchStatus = 404
           , matchHeaders = []
           }
--- a/test/spec/Feature/Query/InsertSpec.hs
+++ b/test/spec/Feature/Query/InsertSpec.hs
@@ -483,7 +483,7 @@ spec actualPgVersion = do
             {"id": 204, "body": "yyy"},
             {"id": 205, "body": "zzz"}]|]
           `shouldRespondWith`
-          [json| {"code":"PGRST205","details":null,"hint":"Perhaps you meant the table 'test.articles'","message":"Could not find the table 'test.garlic' in the schema cache"} |]
+          [json| {"code":"PGRST205","details":null,"hint":null,"message":"Could not find the table 'test.garlic' in the schema cache"} |]
           { matchStatus  = 404
           , matchHeaders = []
           }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ All notable changes to this project will be documented in this file. From versio
   + Removed unnecessary double count when building the `Content-Range`.
 - Add config `client_error_verbosity` to customize error verbosity by @taimoorzaeem in #4088, #3980, #3824
 
+### Fixed
+
+- Fix leaking table and function names when calculating error hint by @taimoorzaeem in #4675
+
 ### Changed
 
 - Log error when `db-schemas` config contains schema `pg_catalog` or `information_schema` by @taimoorzaeem in #4359
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -1768,11 +1768,11 @@ def test_client_error_verbosity_config(defaultenv):
     }
 
     with run(env=env) as postgrest:
-        response = postgrest.session.get("/itemsxx")
+        response = postgrest.session.get("/itemsx")
         assert response.status_code == 404
         assert response.json() == {
             "code": "PGRST205",
-            "message": "Could not find the table 'public.itemsxx' in the schema cache",
+            "message": "Could not find the table 'public.itemsx' in the schema cache",
         }
 
     env = {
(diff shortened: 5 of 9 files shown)


src/PostgREST/Error.hs
```diff
@@ -375,16 +376,27 @@ noRpcHint schema procName params allProcs overloadedProcs =
     -- E.g. ["val", "param", "name"] into "(name, param, val)"
     listToText       = ("(" <>) . (<> ")") . T.intercalate ", " . sort
     possibleProcs
-      | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
-      | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params)
+      | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName >>= checkFuzzyDistance procName
+      | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params) -- no fuzzy distance restriction on parameters hint? too aggresive or is that also a security concern?
```
taimoorzaeem: 3 characters seems too low for parameter hint so I didn't add this restriction on parameters hint. Thoughts?

EDIT: For example, we want to keep the hint when user calls `add_them(a,b,smth_else)` and hint shows `add_them(a,b)`.

steve-chavez: @taimoorzaeem Agree, no problem on the function parameters since there are no privileges on them.

test/spec/Feature/Query/InsertSpec.hs
```diff
@@ -483,7 +483,7 @@ spec actualPgVersion = do
             {"id": 204, "body": "yyy"},
             {"id": 205, "body": "zzz"}]|]
           `shouldRespondWith`
-          [json| {"code":"PGRST205","details":null,"hint":"Perhaps you meant the table 'test.articles'","message":"Could not find the table 'test.garlic' in the schema cache"} |]
+          [json| {"code":"PGRST205","details":null,"hint":null,"message":"Could not find the table 'test.garlic' in the schema cache"} |]
           { matchStatus  = 404
```
steve-chavez, line 487: Shoot, this should have raised an alarm before. Our fuzzy search was so inclusive that it thought `garlic` and `articles` were cousins 😅

wolfgangwalther:
Instead of doing the full fuzzy search first and then checking the levenshtein distance on the result, I think it would be better if the distance check was part of the algorithm already. I think PostgreSQL does the same. Does that mean this kind of change should actually happen further upstream? Could this also help with performance?

steve-chavez:
>  Does that mean this kind of change should actually happen further upstream? Could this also help with performance?

This means on the Data.FuzzySet library right? If so, I agree.

laurenceisla:
> This means on the Data.FuzzySet library right? If so, I agree.

I agree too, hopefully it's possible... the library hasn't seen an update in 3 years (we [have an issue](https://github.com/laserpants/fuzzyset-haskell/issues/9) that hasn't even been triaged too). We should keep looking for alternatives.

>  Could this also help with performance?

It seems so, [according to how they word it](https://www.postgresql.org/docs/current/fuzzystrmatch.html#FUZZYSTRMATCH-LEVENSHTEIN): "levenshtein_less_equal is an accelerated version of the Levenshtein function for use when only small distances are of interest." For future reference, the implementation can be found [here](https://github.com/postgres/postgres/blob/ce4fbe1ac6e9eeae08a825b7ce09efec3614632f/contrib/fuzzystrmatch/fuzzystrmatch.c#L234) and [here](https://github.com/postgres/postgres/blob/ce4fbe1ac6e9eeae08a825b7ce09efec3614632f/src/backend/utils/adt/levenshtein.c#L68) (maybe we can adapt it to Haskell).

taimoorzaeem:
> It seems so, [according to how they word it](https://www.postgresql.org/docs/current/fuzzystrmatch.html#FUZZYSTRMATCH-LEVENSHTEIN): "levenshtein_less_equal is an accelerated version of the Levenshtein function for use when only small distances are of interest." For future reference, the implementation can be found [here](https://github.com/postgres/postgres/blob/ce4fbe1ac6e9eeae08a825b7ce09efec3614632f/contrib/fuzzystrmatch/fuzzystrmatch.c#L234) and [here](https://github.com/postgres/postgres/blob/ce4fbe1ac6e9eeae08a825b7ce09efec3614632f/src/backend/utils/adt/levenshtein.c#L68) (maybe we can adapt it to Haskell).

Agreed. Opened #4686 for this.

To fix the security vulnerability, this PR increases the similarity score threshold from `0.33` to `0.75` as discussed in https://github.com/PostgREST/postgrest/issues/4675#issuecomment-3999181496.

src/PostgREST/Error.hs
```diff
@@ -375,16 +375,30 @@ noRpcHint schema procName params allProcs overloadedProcs =
     -- E.g. ["val", "param", "name"] into "(name, param, val)"
     listToText       = ("(" <>) . (<> ")") . T.intercalate ", " . sort
     possibleProcs
-      | null overloadedProcs = Fuzzy.getOne fuzzySetOfProcs procName
-      | otherwise            = (procName <>) <$> Fuzzy.getOne fuzzySetOfParams (listToText params)
+      | null overloadedProcs = getFuzzyHint Procedure fuzzySetOfProcs procName
+      | otherwise            = (procName <>) <$> getFuzzyHint Params fuzzySetOfParams (listToText params) -- For params, we stick to `getOne` which defaults to 0.33 min score, not a security risk to reveal params
```
steve-chavez:
```
-- For params, we stick to `getOne` which defaults to 0.33 min score, not a security risk to reveal params
```
Nice that we have this comment, should we move it to the `getFuzzyHint` function? (that's where I would read first).

test/spec/Feature/Query/DeleteSpec.hs
```diff
@@ -116,7 +116,7 @@ spec =
       it "fails with 404" $
         request methodDelete "/foozle?id=eq.101" [] ""
           `shouldRespondWith`
-          [json| {"code":"PGRST205","details":null,"hint":"Perhaps you meant the table 'test.foo'","message":"Could not find the table 'test.foozle' in the schema cache"} |]
+          [json| {"code":"PGRST205","details":null,"hint":null,"message":"Could not find the table 'test.foozle' in the schema cache"} |]
```
steve-chavez, line 119: So between `foo` and `foozle` the levenshtein distance is 3, if we were to reuse the postgres algorithm the hint would apply in this case. Perhaps we can tune the score to have a match here?

steve-chavez: Not a deal breaker if not possible.

taimoorzaeem: Yeah it's difficult, we'd have to lower the score too much for this case because the name length is too short. In production, identifiers like table and functions names are usually longer.

And so the 3 distance case will be covered with `0.75` score as soon as name length touches 9 to 12 characters.

src/PostgREST/Error.hs
```diff
@@ -375,16 +375,30 @@ noRpcHint schema procName params allProcs overloadedProcs =
...
+data HintType
+  = Table
+  | Procedure
+  | Params
+
+-- | Get hint using Fuzzy Search with at least 0.75 similarity score
+getFuzzyHint :: HintType -> Fuzzy.FuzzySet -> Text -> Maybe Text
+getFuzzyHint hintType =
+  let minScore = 0.75 :: Double -- used for table and procedure name hints
+  in case hintType of
```
mkleczek, line 398: I would refactor it:

instead of having a dedicated enum and a function doing pattern match on this enum, it is IMO better to simply have three different functions:
* getTableFuzzyHint
* getProcedureFuzzyHint
* getParamsFuzzyHint

Callsites have to know which variant to call anyway (right now they have to pass a proper `HintType` instance) so let them call different functions instead.

steve-chavez: Hm, I don't agree here. I find it more readable as it is. Not sure if one or the other would result in more LOCs or stay the same.

@taimoorzaeem WDYT?

steve-chavez: I guess the intention of this single function was to reuse the `let minScore = 0.75 :: Double`?

If this would be split into different functions, then we would need a global for that or we'd have to repeat the number.

src/PostgREST/Error.hs
```diff
@@ -375,16 +375,30 @@ noRpcHint schema procName params allProcs overloadedProcs =
...
 tableNotFoundHint schema tblName SchemaCache{dbTablesFuzzyIndex}
   = fmap (\tbl -> "Perhaps you meant the table '" <> schema <> "." <> tbl <> "'") perhapsTable
     where
-      perhapsTable = (`Fuzzy.getOne` tblName) =<< HM.lookup schema dbTablesFuzzyIndex
+      perhapsTable = (\fuzzySet -> getFuzzyHint Table fuzzySet tblName) =<< HM.lookup schema dbTablesFuzzyIndex
+
+data HintType
+  = Table
+  | Procedure
+  | Params
```
steve-chavez: Maybe
```suggestion
data HintType
  = HintTable
  | HintProcedure
  | HintParams
```

src/PostgREST/Error.hs
taimoorzaeem: > Hm, I don't agree here. I find it more readable as it is.

Yes, agreed.

> I guess the intention of this single function was to reuse the `let minScore = 0.75 :: Double`?

Yup, that's exactly right.

mkleczek: > Hm, I don't agree here. I find it more readable as it is. Not sure if one or the other would result in more LOCs or stay the same.
> 

The suggestion is not really about readability (that's subjective) or LoC.

It is about coupling/cohesion.

Right now if you want to change logic for one type of object (let's say `Procedure`) you have to change shared unit (function) and risk breaking other (unrelated) objects' handling. The reason is that a single function has low cohesion - it has multiple responsibilities for which it does different things.

Low cohesion also makes refactoring more difficult: if for some (unknown at this moment) reason you would want to split hint processing for different object types into different modules, moving a function is simple, but splitting a function (and supporting type) is way more difficult because it affects all call sites.

Splitting it into different units (functions) increases cohesion and long term maintainability.

Not very important in this case but I try to keep cohesion as high as possible as a rule of thumb.

@taimoorzaeem - I'm fine with whatever you decide.

steve-chavez: > Low cohesion also makes refactoring more difficult: if for some (unknown at this moment) reason you would want to split hint processing for different object types into different modules, moving a function is simple, but splitting a function (and supporting type) is way more difficult because it affects all call sites.

> Splitting it into different units (functions) increases cohesion and long term maintainability.

Not disagreeing but so far hints have been always contained in this Errors module. If the need arises we can refactor to different functions. It's not so much code.

Will merge as is for now and do a patch release since https://github.com/PostgREST/postgrest/issues/4675 can be considered a vulnerability.

PR #4687, opened 2026-03-05
test(refactor): Replace slocat with Toxiproxy

mkleczek:
**DISCLAIMER:**

This is a first PR introducing https://github.com/Shopify/toxiproxy to PostgREST testing infrastructure. It replaces `slocat` with `toxiproxy` and provides basic infrastructure in `nix/withTools.nix`.


nix/tools/withTools.nix
```diff
@@ -6,9 +6,10 @@
 , postgresqlVersions
 , postgrest
 , python3Packages
-, slocat
+, socat
```
steve-chavez: I guess slocat is really being replaced with a combination of `socat` and `toxiproxy`?

Found some references for this project:

- http://www.dest-unreach.org/socat/
- https://github.com/3ndG4me/socat

nix/tools/withTools.nix
```diff
@@ -63,8 +66,15 @@ let
...
           HBA_FILE="$tmpdir/pg_hba.conf"
           echo "local $PGDATABASE some_protected_user password" > "$HBA_FILE"
-          echo "local $PGDATABASE all trust" >> "$HBA_FILE"
-          echo "local replication all trust" >> "$HBA_FILE"
+          {
+            echo "local $PGDATABASE all trust"
+            echo "local replication all trust"
+            echo "host $PGDATABASE some_protected_user localhost scram-sha-256"
+            echo "host $PGDATABASE all localhost trust"
+          } >> "$HBA_FILE"
```
steve-chavez, line 73: > echo "host $PGDATABASE some_protected_user localhost scram-sha-256

Isn't that redundant with the above untouched `echo "local $PGDATABASE some_protected_user password" > "$HBA_FILE"`? Or maybe both lines can be merged?

Also, this change also looks like it includes a refactor (the grouping  `{} >>`). Might be better to separate into another commit.

nix/tools/withTools.nix
mkleczek: > I guess slocat is really being replaced with a combination of `socat` and `toxiproxy`?
> 

The issue with Toxiproxy is that it is strictly TCP socket based tool. And while it is possible to have PostgreSQL listen on both TCP and UNIX socket, PostgREST is capable to only listen on one.
To maintain backwards compatibility with tests depending on Pgrst listening on UNIX socket, I decided to "wrap" Toxiproxy with `socat`.
See the comment in `withSlowPgrst`

nix/tools/withTools.nix
mkleczek: > > echo "host $PGDATABASE some_protected_user localhost scram-sha-256
> 
> Isn't that redundant with the above untouched `echo "local $PGDATABASE some_protected_user password" > "$HBA_FILE"`? Or maybe both lines can be merged?

It is for `host` access (not `local`) so not redundant. Looking at [pg_hba.conf](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) documentation, I don't think they can be merged.

> 
> Also, this change also looks like it includes a refactor (the grouping `{} >>`). Might be better to separate into another commit.

It's not unrelated refactoring but is enforced by the linter. When I added lines in the previous style it ended up with error:
`SC2129 (style): Consider using { cmd1; cmd2; } >> file instead of individual redirects.`

nix/tools/withTools.nix
mkleczek: @steve-chavez fixed in #4688

Issue #4681, opened 2026-02-28
PostgREST returns 400 instead of 503 when database server is temporarily unavailable


mkleczek:
#4674 helped find the following issue:
https://github.com/PostgREST/postgrest/actions/runs/22520100336/job/65243536483?pr=4674#step:5:1901

The test case: https://github.com/mkleczek/postgrest/blob/dd59a0889b1d451ed11547f9ff184efdb06cd7b7/test/spec/Feature/ToxiSpec.hs#L18

The issue is that once a connection is established and available in the pool we do not check for connectivity errors anymore. `hasql-pool` reports it as `UsageError`, `usePool` treats it as client error and returns `400`.

PR #4688, opened 2026-03-06
test(refactor): remove use of socat

mkleczek:
**DISCLAIMER**:

Follow-up to #4687 removing `socat` usage from `withTools.nix`. This required changing of load tests to connect to PostgREST using TCP (instead of UNIX) socket.


steve-chavez:
I just recalled that `slocat` was introduced to test pipeline mode, but that was never completed. See https://github.com/PostgREST/postgrest/pull/2682. (in hindsight merging this was a mistake on my side, I thought pipeline would come soon :man_facepalming:  )

Do we need it now for some tests? Otherwise I think we could just do a clean remove without adding toxyproxy.

mkleczek:
> I just recalled that `slocat` was introduced to test pipeline mode, but that was never completed. See #2682. (in hindsight merging this was a mistake on my side, I thought pipeline would come soon 🤦‍♂️ )
> 
> Do we need it now for some tests? Otherwise I think we could just do a clean remove without adding toxyproxy.

Yeah, I have many tests in the pipeline testing various recovery scenarios. It seems like we have some work to do in error handling. See for example #4681 that was triggered by the very first test I've written to demonstrate Toxiproxy usage.

steve-chavez:
> Yeah, I have many tests in the pipeline testing various recovery scenarios

Yes, but I meant if currently we depend on `slocat` for any tests? If not, a clean commit just removing it globally would be simpler to merge.

mkleczek:
> > Yeah, I have many tests in the pipeline testing various recovery scenarios
> 
> Yes, but I meant if currently we depend on `slocat` for any tests? If not, a clean commit just removing it globally would be simpler to merge.

Mixed loadtest use `withTools.withSlowPg` and `withTools.withSlowPgrst`.
Removing `slocat` would require removal of these two from `withTools` and then potentially re-adding them but this time with `toxiproxy` in yet another commit.

Maybe a better idea is to get rid of `slocat` first and replace it with `toxiproxy`, leaving other refactorings and tests restructuring for another PR? Otherwise we slow down implementation of `toxiproxy` based tests until we do unrelated refactoring.

steve-chavez:
> Mixed loadtest use withTools.withSlowPg and withTools.withSlowPgrst.

Yeah, it's not really needed in those (as I mentioned it was introduced on https://github.com/PostgREST/postgrest/pull/2682/), but we might vary the loadtest results if we remove it, so that's one reason to keep the delay there.

nix/tools/loadtest.nix
```diff
@@ -23,11 +24,12 @@ let
...
         # We manually implement a required check here
         # See also: https://github.com/matejak/argbash/issues/80
-        : "''${PGRST_SERVER_UNIX_SOCKET:?PGRST_SERVER_UNIX_SOCKET is required}"
+        : "''${PGRST_SERVER_HOST:?PGRST_SERVER_HOST is required}"
+        : "''${PGRST_SERVER_PORT:?PGRST_SERVER_PORT is required}"
 
         ${vegeta}/bin/vegeta -cpus 1 attack \
                                      -dns-ttl -1 \
-                                     -unix-socket "$PGRST_SERVER_UNIX_SOCKET" \
+                                     -connect-to "postgrest:80:$PGRST_SERVER_HOST:$PGRST_SERVER_PORT" \
```
steve-chavez, line 32: > This required changing of load tests to connect to PostgREST using TCP (instead of UNIX) socket.

Can we isolate the change to TCP in another PR/commit? (while keeping `slocat`) That way we could see what's the impact just for that on the loadtest results.

steve-chavez:
> DISCLAIMER:
> This commit was authored entirely by a human without the assistance of LLMs.

@mkleczek I would be against adding the above notice, even worse is that it's also added on the commit message? (and that just adds noise to the repo's history)

We don't require any such disclaimers as per the AI policy https://github.com/PostgREST/postgrest/blob/main/CONTRIBUTING.md#ai-policy.

I would suggest to just omit such notices from all PRs/commits. We don't really want to overthink the AI policy. 

@wolfgangwalther WDYT?

mkleczek:
> > DISCLAIMER:
> > This commit was authored entirely by a human without the assistance of LLMs.
> 
> @mkleczek I would be against adding the above notice, even worse is that it's also added on the commit message? (and that just adds noise to the repo's history)
> 

This disclaimer is a form of protection of my intellectual property (which these contributions clearly are as I am the sole owner of their copyright). As an individual I need protection from any accusations of copyright infringement caused by the use of LLMs.

> We don't require any such disclaimers as per the AI policy https://github.com/PostgREST/postgrest/blob/main/CONTRIBUTING.md#ai-policy.
> 
> I would suggest to just omit such notices from all PRs/commits. We don't really want to overthink the AI policy.

I am afraid words have consequences - even if unintended. Especially words in a public policy document. Maybe it would be a good idea to have a disclaimer in the policy clearly stating that proposed contributions are assumed to follow the policy and be created by a human.
That way such individual disclaimers would be less needed.

nix/tools/loadtest.nix
mkleczek: > > This required changing of load tests to connect to PostgREST using TCP (instead of UNIX) socket.
> 
> Can we isolate the change to TCP in another PR/commit? (while keeping `slocat`) 

The reason we have `slocat` at all is that it supports UNIX sockets. Does it make sense to do all the work of switching to TCP while leaving `slocat` in place?

> That way we could see what's the impact just for that on the loadtest results.

Aren't load tests comparable regardless? We are running the same suite against multiple versions anyway.
IMHO we are really interested in relative numbers (absolute numbers are not very meaningful anyway).

wolfgangwalther:
> This disclaimer is a form of protection of my intellectual property (which these contributions clearly are as I am the sole owner of their copyright). As an individual I need protection from any accusations of copyright infringement caused by the use of LLMs.

It's nice of you to tell us that we don't need to worry about LLM content here while deciding whether to review or not - thank you. For that purpose, it is enough to have it in the PR description.

When we hit the merge button, we decide at that moment in time that we assume the policies have been followed. There is no need to keep this disclaimer in the commit messages - it is kept in the PR for future reference, too. The commit message should be written for our future selves, to be able to understand why the change was made, key decisions while writing this etc. - so that we can go back to this in the commit history, for example when git blaming a change or such. This disclaimer is not useful here.

> I am afraid words have consequences - even if unintended. Especially words in a public policy document. Maybe it would be a good idea to have a disclaimer in the policy clearly stating that proposed contributions are assumed to follow the policy and be created by a human. That way such individual disclaimers would be less needed.

This **feels** a bit child-ish **to me** - throwing mud back at us, because you don't agree with the introduction of the policy. I'm sure this is not your intent, though. I guess you are just very principled and want things to be done right - and I fully respect that. I hope we can find a middle way that makes it work for all parties involved.

mkleczek:
> 
> It's nice of you to tell us that we don't need to worry about LLM content here while deciding whether to review or not - thank you. For that purpose, it is enough to have it in the PR description.
> 
> When we hit the merge button, we decide at that moment in time that we assume the policies have been followed. There is no need to keep this disclaimer in the commit messages - it is kept in the PR for future reference, too.

Fine for me. Will amend the commit message.

> 
> > I am afraid words have consequences - even if unintended. Especially words in a public policy document. Maybe it would be a good idea to have a disclaimer in the policy clearly stating that proposed contributions are assumed to follow the policy and be created by a human. That way such individual disclaimers would be less needed.
> 
> This **feels** a bit child-ish **to me** - throwing mud back at us, because you don't agree with the introduction of the policy. I'm sure this is not your intent, though.

Sorry for sounding childish. I absolutely don't want to throw mud back. The disclaimer is just to make sure there are no unnecessary presumptions (which I've already been faced with) which are difficult to deal with after the fact.

wolfgangwalther:
> presumptions (which I've already been faced with) which are difficult to deal with after the fact.

Right, I understand that. I think it's more than enough for us to have a policy in place, you being aware of it and - as demonstrated - OK to follow it. That's enough for me to trust you and review your PRs under the assumption that you wrote this on your own, I don't need a disclaimer for it.

Or put differently: It's not the new policy that poses a practical problem, but it's the unclear state that we were in before. It was not made explicit whether it was allowed or disallowed to use LLMs.

PR #4685, opened 2026-03-04
Add copyright notice for Michał Kłeczek

mkleczek:
There are 22 merged PRs submitted by @mkleczek. The first one touching the actual code is from 2025.

PR #4698, opened 2026-03-09
chore: update license

steve-chavez:
Also including the docs.

cc @begriffs @wolfgangwalther 

Related to https://github.com/PostgREST/postgrest/pull/4685#discussion_r2905333471


LICENSE
```diff
@@ -1,6 +1,4 @@
-Copyright (c) 2014 Joe Nelson
-Copyright (c) 2019 Steve Chavez
-Copyright (c) 2026 Wolfgang Walther
+Copyright (c) 2014-2026 The PostgREST contributors
```
steve-chavez, line 1: One thing is is that now we have to keep updating the year here. I've noticed PostgreSQL does the same https://github.com/postgres/postgres/blob/master/COPYRIGHT

steve-chavez: I've noticed postgres does this:

```
Portions Copyright (c) 1996-2026, PostgreSQL Global Development Group

Portions Copyright (c) 1994, The Regents of the University of California
```
https://github.com/postgres/postgres/blob/master/COPYRIGHT#L4-L6

Maybe we should just keep the existing copyrights and add "The PostgREST contributors"?

wolfgangwalther: I assume they kept the regents bit, because these would not really be covered by "postgresql global development group". In our case all of the existing copyright holders would be, though.

So as long as @begriffs agrees, we should be OK to change it that way.

begriffs: No objection.

begriffs (approved):
Fine by me.

PR #4702, opened 2026-03-10
fix: shutdown should wait for in flight requests

### LINKED ISSUE #4689: Graceful shutdown
Right now PostgREST does not wait for in-flight requests to finish upon shutdown (ie. `SIGTERM`).

It would be good if we used [Warp shutdownHandler](https://hackage-content.haskell.org/package/warp-3.4.12/docs/Network-Wai-Handler-Warp.html#v:setInstallShutdownHandler) to handle shutdown gracefully.

This is crucial to handle upgrades in near-zero downtime fashion.

mkleczek:
Fixes #4689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ All notable changes to this project will be documented in this file. From versio
 - Add config `db-timezone-enabled` for optional querying of timezones by @taimoorzaeem in #4751
 - Log schema cache queries timings on `log-level=debug` by @steve-chavez in #4805
 
+### Fixed
+
+- Shutdown should wait for in flight requests by @mkleczek in #4702
+
 ### Changed
 
 - Drop support for PostgreSQL EOL version 13 by @wolfgangwalther in #4193
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -59,6 +59,43 @@ let
       postgresql-binary = lib.dontCheck (lib.doJailbreak prev.postgresql-binary_0_13_1_3);
       text-builder = prev.text-builder_0_6_10;
       text-builder-dev = prev.text-builder-dev_0_3_10;
+
+      http2 =
+        prev.callHackageDirect
+          {
+            pkg = "http2";
+            ver = "5.4.0";
+            sha256 = "sha256-PeEWVd61bQ8G7LvfLeXklzXqNJFaAjE2ecRMWJZESPE=";
+          }
+          { };
+
+      http-semantics =
+        prev.callHackageDirect
+          {
+            pkg = "http-semantics";
+            ver = "0.4.0";
+            sha256 = "sha256-rh0z51EKvsu5rQd5n2z3fSRjjEObouNZSBPO9NFYOF0=";
+          }
+          { };
+
+      network-run =
+        prev.callHackageDirect
+          {
+            pkg = "network-run";
+            ver = "0.5.0";
+            sha256 = "sha256-vbXh+CzxDsGApjqHxCYf/ijpZtUCApFbkcF5gyN0THU=";
+          }
+          { };
+
+      warp =
+        lib.dontCheck (prev.callHackageDirect
+          {
+            pkg = "warp";
+            ver = "3.4.13";
+            sha256 = "sha256-jmr8kpeSPDkOhT0i9PhozZapX4nUs92cOX7POAGb7/M=";
+          }
+          { });
+
     };
 in
 {
--- a/postgrest.cabal
+++ b/postgrest.cabal
@@ -156,7 +156,7 @@ library
                     -- for unix sockets; this is tested in test/io/test_io.py. See
                     -- https://github.com/kazu-yamamoto/logger/commit/3a71ca70afdbb93d4ecf0083eeba1fbbbcab3fc3
                     , wai-logger                >= 2.4.0
-                    , warp                      >= 3.3.19 && < 3.5
+                    , warp                      >= 3.4.13 && < 3.5
                     , stm                       >= 2.5 && < 3
                     , stm-hamt                  >= 1.2 && < 2
                     , focus                     >= 1.0 && < 2
--- a/stack.yaml
+++ b/stack.yaml
@@ -16,9 +16,12 @@ extra-deps:
   - hasql-notifications-0.2.2.2
   - hasql-pool-1.0.1
   - hasql-transaction-1.1.0.1
+  - http-semantics-0.4.0
+  - http2-5.4.0
   - postgresql-binary-0.13.1.3
   - text-builder-0.6.10
   - text-builder-dev-0.3.10
+  - warp-3.4.13
 
 allow-newer: true
 allow-newer-deps:
--- a/test/io/test_io.py
+++ b/test/io/test_io.py
@@ -128,7 +128,6 @@ def test_flush_pool_no_interrupt(defaultenv):
         t.join()
 
 
-@pytest.mark.xfail(reason="Graceful shutdown is currently failing", strict=True)
 def test_graceful_shutdown_waits_for_in_flight_request(defaultenv):
     "SIGTERM should allow in-flight requests to finish before exiting"
 
--- a/test/memory/memory-tests.sh
+++ b/test/memory/memory-tests.sh
@@ -110,9 +110,9 @@ jsonKeyTest "10M" "POST" "/rpc/leak?columns=blob" "32M"
 jsonKeyTest "10M" "POST" "/leak?columns=blob" "32M"
 jsonKeyTest "10M" "PATCH" "/leak?id=eq.1&columns=blob" "50M"
 
-jsonKeyTest "50M" "POST" "/rpc/leak?columns=blob" "73M"
-jsonKeyTest "50M" "POST" "/leak?columns=blob" "73M"
-jsonKeyTest "50M" "PATCH" "/leak?id=eq.1&columns=blob" "73M"
+jsonKeyTest "50M" "POST" "/rpc/leak?columns=blob" "78M"
+jsonKeyTest "50M" "POST" "/leak?columns=blob" "78M"
+jsonKeyTest "50M" "PATCH" "/leak?id=eq.1&columns=blob" "78M"
 
 postJsonArrayTest "1000" "/perf_articles?columns=id,body" "21M"
 postJsonArrayTest "10000" "/perf_articles?columns=id,body" "22M"
--- a/src/PostgREST/App.hs
+++ b/src/PostgREST/App.hs
@@ -22,6 +22,7 @@ import GHC.IO.Exception (IOErrorType (..))
 import System.IO.Error  (ioeGetErrorType)
 
 import Control.Monad.Except     (liftEither)
+import Control.Monad.Extra      (whenJust)
 import Data.Either.Combinators  (mapLeft, whenLeft)
 import Data.Maybe               (fromJust)
 import Data.String              (IsString (..))
@@ -79,8 +80,10 @@ run appState = do
 
   AppState.schemaCacheLoader appState -- Loads the initial SchemaCache
   (mainSocket, adminSocket) <- initSockets conf
-
-  Unix.installSignalHandlers observer (AppState.getMainThreadId appState) (AppState.schemaCacheLoader appState) (AppState.readInDbConfig False appState)
+  let closeSockets = do
+        whenJust adminSocket NS.close
+        NS.close mainSocket
+  Unix.installSignalHandlers observer closeSockets (AppState.schemaCacheLoader appState) (AppState.readInDbConfig False appState)
 
   Listener.runListener appState
 
--- a/src/PostgREST/Unix.hs
+++ b/src/PostgREST/Unix.hs
@@ -19,10 +19,9 @@ import           System.Directory      (removeFile)
 import           System.IO.Error       (isDoesNotExistError)
 
 -- | Set signal handlers, only for systems with signals
-installSignalHandlers :: Observation.ObservationHandler -> ThreadId -> IO () -> IO () -> IO ()
+installSignalHandlers :: Observation.ObservationHandler -> IO () -> IO () -> IO () -> IO ()
 #ifndef mingw32_HOST_OS
-installSignalHandlers observer tid usr1 usr2 = do
-  let interrupt = throwTo tid UserInterrupt
+installSignalHandlers observer interrupt usr1 usr2 = do
   install Signals.sigINT  $ observer (Observation.TerminationUnixSignalObs "SIGINT") >> interrupt
   install Signals.sigTERM $ observer (Observation.TerminationUnixSignalObs "SIGTERM") >> interrupt
   install Signals.sigUSR1 usr1
--- a/stack.yaml.lock
+++ b/stack.yaml.lock
@@ -61,6 +61,20 @@ packages:
   original:
     hackage: hasql-transaction-1.1.0.1
 - completed:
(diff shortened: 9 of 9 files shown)


test/io/test_io.py
```diff
@@ -105,6 +105,28 @@ def sleep():
...
+
+        t = Thread(target=sleep)
+        t.start()
+
+        # Wait for the request to be in-flight before shutting down.
+        time.sleep(1)
+
+        postgrest.process.terminate()
+
+        t.join()
```
steve-chavez: Had no idea this happened, it would have been great to see it failing on CI first, then do the fix commit.

steve-chavez: Also, do we know what happens to the pg connection when this happens? Does it keep running or it's also closed immediately?

mkleczek: > Also, do we know what happens to the pg connection when this happens? Does it keep running or it's also closed immediately?

It keeps running but the whole pool is destroyed very soon by `bracket` in `CLI.runAppCommand`.

EDIT: today the request handling thread is killed (ie. an asynchronous exception is thrown). Depending on when it happens it might cause tx rollback (if thrown in the middle of tx) or tx commit (if thrown right after tx).

mkleczek:
Marked it as draft because there is a problem caused by https://github.com/yesodweb/wai/issues/853.
What that means is we have to put the hard limit on `gracefulShutdownTimeout` otherwise `warp` will not shutdown at all if there is a reverse proxy keeping connections open (ie. using HTTP keep-alives).
@steve-chavez we need to decide what this limit should be. Unfortunately, `warp` is going to wait exactly that long if there are any keep-alive connections, so this timeout cannot be too long.

mkleczek:
Related: https://github.com/yesodweb/wai/pull/1064

test/io/test_io.py
mkleczek: > Had no idea this happened, it would have been great to see it failing on CI first, then do the fix commit.

The test in the first commit is marked as `xfail` to make sure it is failing. The second commit then provides a fix.
I've opened a draft pr to demonstrate the failure - here is the test result: https://github.com/PostgREST/postgrest/actions/runs/22969464771/job/66681946995?pr=4708

mkleczek: If raised a PR #4714 with the io test marked with `xfail`. I think it is a pretty cool idea to have a PR with failing test first and then another one with a fix.

mkleczek:
@steve-chavez @wolfgangwalther - this is dependent on `warp` fixing https://github.com/yesodweb/wai/issues/853. I've raised a draft PR https://github.com/yesodweb/wai/pull/1064 but don't know if or when it will land.

First commit in this PR is a change to our Haskell overlay pointing to `warp` version from the above PR.

We need to decide what to do with this one.

steve-chavez:
> this is dependent on warp fixing https://github.com/yesodweb/wai/issues/853. I've raised a draft PR https://github.com/yesodweb/wai/pull/1064 but don't know if or when it will land.

Nice, looks like the maintainer is already replying.

> First commit in this PR is a change to our Haskell overlay pointing to warp version from the above PR.
> We need to decide what to do with this one.

Since replies are fast I think we should aim to merge upstream and not fork/vendor for now.

themavik (commented):
Passing closeSockets into installSignalHandlers and dropping the xfail on test_graceful_shutdown_waits_for_in_flight_request lines up with Warp closing the listen socket while draining in-flight work. nit: comment says becase — should be because.

mkleczek:
> Passing closeSockets into installSignalHandlers and dropping the xfail on test_graceful_shutdown_waits_for_in_flight_request lines up with Warp closing the listen socket while draining in-flight work. nit: comment says becase — should be because.

AI slop?

wolfgangwalther:
> AI slop?

Looked at the user's profile earlier today and I think so, yes.

steve-chavez:
Made the ai policy more notorious : https://github.com/PostgREST/postgrest/commit/effcc080e58e0e251d718a27e4445e5a44e4a78a.

mkleczek:
Great news: https://github.com/yesodweb/wai/pull/1071#issuecomment-4260049949

Will restart working on this PR.

steve-chavez:
> Great news: https://github.com/yesodweb/wai/pull/1071#issuecomment-4260049949

Awesome! Would we need to upgrade GHC to use the new warp version?

mkleczek:
> > Great news: [yesodweb/wai#1071 (comment)](https://github.com/yesodweb/wai/pull/1071#issuecomment-4260049949)
> 
> Awesome! Would we need to upgrade GHC to use the new warp version?

Doesn't look like it.

But there are `cabal` build failures caused by some mysterious compilation errors when building `jose-jwt`. I am not sure what's going on but new `warp` introduces `crypton >= 1.1` dependency (indirectly) which might be incompatible with `jose-jwt`. Not sure why `stack` and `nix` builds work though - do they have older `crypton` version pinned somehow?

It also looks like newer `warp` introduces some increased memory usage.

mkleczek:
> But there are `cabal` build failures caused by some mysterious compilation errors when building `jose-jwt`. I am not sure what's going on but new `warp` introduces `crypton >= 1.1` dependency (indirectly) which might be incompatible with `jose-jwt`.

`jose-jwt` got fresh upgrade of `crypton` and migrated to `ram`:
https://github.com/tekul/jose-jwt/commit/f93bd9436d798eb81618fa0f699f677194efcf8c

I wonder when it is going to be released (the latest `0.10.0` was released in 2024)

wolfgangwalther:
> But there are `cabal` build failures caused by some mysterious compilation errors when building `jose-jwt`. I am not sure what's going on but new `warp` introduces `crypton >= 1.1` dependency (indirectly) which might be incompatible with `jose-jwt`.

I believe these mysterious compilation errors should be fixed with #4825 - at least I had some of these and I assume they were the same as yours :)

stack.yaml
```diff
@@ -15,6 +15,13 @@ extra-deps:
   - jose-jwt-0.10.0
   - postgresql-libpq-0.10.1.0
   - streaming-commons-0.2.3.1
+  - warp-3.4.13
+  - auto-update-0.2.6
+  - http2-5.4.0
+  - time-manager-0.3.2
+  - http-semantics-0.4.0
+  - network-control-0.1.7
+  - network-3.2.8.0
```
wolfgangwalther: At least auto-update, network-control and network are now part of the stackage pin we use since #4825. Better check whether the others are still needed as well (ofc, warp is needed).

cabal.project.freeze
```diff
@@ -1 +1 @@
-index-state: hackage.haskell.org 2025-10-29T04:02:18Z
+index-state: hackage.haskell.org 2026-04-16T20:19:07Z
```
wolfgangwalther: Should not be needed anymore, already done in #4825.

postgrest.cabal
```diff
@@ -123,7 +123,7 @@ library
                     , http-client               >= 0.7.19 && < 0.8
                     , http-types                >= 0.12.2 && < 0.13
                     , insert-ordered-containers >= 0.2.2 && < 0.3
-                    , jose-jwt                  >= 0.9.6 && < 0.11
+                    , jose-jwt                  >= 0.11 && < 0.12
```
wolfgangwalther: I think these can be reverted with #4825.

nix/overlays/haskell-packages.nix
wolfgangwalther, line 1: Changes here should also be simplified on a rebase.

Vlix:
> I am not sure what's going on but new `warp` introduces `crypton >= 1.1` dependency (indirectly) which might be incompatible with `jose-jwt`. 

If you use `crypton-x509-1.8.0`, it should work with `crypton < 1.1`. (haven't tried building the new `warp` with `crypton-x509-1.8.0`, so I can't guarantee it works)

cabal.project.freeze
mkleczek: Fixed

postgrest.cabal
mkleczek: Done

nix/overlays/haskell-packages.nix
mkleczek: Left only what's necessary.

stack.yaml
mkleczek: Fixed

test/io/test_io.py
```diff
@@ -128,7 +128,6 @@ def sleep():
         t.join()
 
 
-@pytest.mark.xfail(reason="Graceful shutdown is currently failing", strict=True)
 def test_graceful_shutdown_waits_for_in_flight_request(defaultenv):
```
steve-chavez, line 131: Inlining this for review https://github.com/PostgREST/postgrest/blob/bcc8998e5e122e69a9dc4c8ca895c59175d1ea4a/test/io/test_io.py#L132-L150

steve-chavez: This is pretty cool, so I've tested this behavior on main with a request to `/rpc/sleep?seconds=3`.

- I hit two Ctrl-C:

```
20/Apr/2026:13:15:52 -0500: Received termination unix signal SIGINT
20/Apr/2026:13:15:52 -0500: Received termination unix signal SIGINT
```

PostgREST dies immediately after the second and the `curl` request says "empty reply" which is bad:
```
curl 'localhost:3000/rpc/sleep?seconds=3'
```

- I hit one Ctrl-C and also do a second curl req `/projects`:

```
20/Apr/2026:13:17:28 -0500: Received termination unix signal SIGINT
20/Apr/2026:13:17:30 -0500: Warp server: Thread killed by timeout manager
20/Apr/2026:13:17:30 -0500: Connection 145fa366-9882-4f67-ab98-ec2ca7c604e1 is terminated due to release
```
Same thing with the `curl` request, it hits `empty reply` while the second request waits for postgREST death until it logs:

```curl 'localhost:3000/projects'
curl: (56) Recv failure: Connection reset by peer
```

(did this with a `PGRST_DB_POOL=10`, not sure why it waits)

Now with this PR:

- Hit many Ctrl-C and do a second curl request:
```bash
20/Apr/2026:13:26:16 -0500: Received termination unix signal SIGINT
20/Apr/2026:13:26:16 -0500: Received termination unix signal SIGINT
20/Apr/2026:13:26:16 -0500: Received termination unix signal SIGINT
20/Apr/2026:13:26:19 -0500: Connection 0580d25e-73cd-4157-9f55-0a448737133e is available
20/Apr/2026:13:26:19 -0500: Borrowed a connection from the pool
127.0.0.1 - postgrest_test_anonymous [20/Apr/2026:13:26:19 -0500] "GET /rpc/sleep?seconds=3 HTTP/1.1" 204 - "" "curl/7.81.0"
20/Apr/2026:13:26:19 -0500: Connection 0580d25e-73cd-4157-9f55-0a448737133e is terminated due to release
```

The first `curl` request completes and the second replies after the first SIGINT:

```bash
curl 'localhost:3000/projects'
curl: (7) Failed to connect to localhost port 3000 after 0 ms: Connection refused
```

Which looks like a much better behavior.

steve-chavez: @mkleczek Is there any extra automated test we could add from my above comment? Or do you think it's enough as is.

steve-chavez:
> It also looks like newer warp introduces some increased memory usage.

I see the failures:

```
not ok 7 - POST /rpc/leak?columns=blob: with a json key of 50M the memory usage(77,356,040 bytes) is more than 73M
not ok 8 - POST /leak?columns=blob: with a json key of 50M the memory usage(77,239,768 bytes) is more than 73M
not ok 9 - PATCH /leak?id=eq.1&columns=blob: with a json key of 50M the memory usage(77,304,464 bytes) is more than 73M
```

@mkleczek I think it should be fine to correct these lines to 78M.

https://github.com/PostgREST/postgrest/blob/6b4022b0c1cce9c44dc020c3cc0f4e099a079552/test/memory/memory-tests.sh#L113-L115

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. From versio
 - Add `Vary` header to responses by @develop7 in #4609
 - Add config `db-timezone-enabled` for optional querying of timezones by @taimoorzaeem in #4751
 - Log schema cache queries timings on `log-level=debug` by @steve-chavez in #4805
+- Shutdown should wait for in flight requests by @mkleczek in #4702
```
wolfgangwalther, line 20: commit and PR title say `fix:`, so:


```suggestion

### Fixed

- Shutdown should wait for in flight requests by @mkleczek in #4702
```

test/io/test_io.py
wolfgangwalther: So, you're saying the behavior of rejecting the connection immediately should be tested? Is that a warp-thing or do we do that? If it's warp, I'd say we don't need to test that here and the existing test should be good enough. But quickly glancing at the code it looks like we do that explicitly.

steve-chavez: Yeah agree, warp does it so no need to further test this.

CHANGELOG.md
mkleczek: Fixed

wolfgangwalther (approved):

mkleczek:
> @mkleczek I think it should be fine to correct these lines to 78M.
> 

Done

steve-chavez (approved):
🚀 🚀 🚀

steve-chavez:
It's not possible to backport this because it depends on

- https://github.com/PostgREST/postgrest/commit/051b4799deefa44a0c3aa679f18ac3ef8fd801b5
  + which depends on dropping pg 13 support https://github.com/PostgREST/postgrest/commit/1b74b93466bb9d7dd154fbded931591a12128d97

There will be a showstopping error:

```
       error: postgresql_13 has been removed since it reached its EOL upstream
```

steve-chavez:
WDYT of changing this one to a `feat` instead of `fix`, that way we add docs for it. Some projects have "graceful shutdown" documented like:

- https://trino.io/docs/current/admin/graceful-shutdown.html
- https://static-web-server.net/features/graceful-shutdown/ 

We can mention this is done by `Warp`.

test/memory/memory-tests.sh
```diff
@@ -110,9 +110,9 @@ jsonKeyTest "10M" "POST" "/rpc/leak?columns=blob" "32M"
 jsonKeyTest "10M" "POST" "/leak?columns=blob" "32M"
 jsonKeyTest "10M" "PATCH" "/leak?id=eq.1&columns=blob" "50M"
 
-jsonKeyTest "50M" "POST" "/rpc/leak?columns=blob" "73M"
-jsonKeyTest "50M" "POST" "/leak?columns=blob" "73M"
-jsonKeyTest "50M" "PATCH" "/leak?id=eq.1&columns=blob" "73M"
+jsonKeyTest "50M" "POST" "/rpc/leak?columns=blob" "78M"
+jsonKeyTest "50M" "POST" "/leak?columns=blob" "78M"
+jsonKeyTest "50M" "PATCH" "/leak?id=eq.1&columns=blob" "78M"
```
wolfgangwalther, line 115: In https://github.com/PostgREST/postgrest/pull/4902#issuecomment-4480554289, we noticed that this also makes these kinds of requests a tad slower. Most likely introduced via one of the changed dependencies.

steve-chavez:
> WDYT of changing this one to a feat instead of fix, that way we add docs for it. Some projects have "graceful shutdown" documented like:

Went ahead with the above:

- Moved changelog entry properly to a feature https://github.com/PostgREST/postgrest/commit/876dfad9d9fe04cf0f354a99f845b7f9b5e5a0b8
- Documented the feature  https://github.com/PostgREST/postgrest/commit/142b54158e8b0c16000756c93b86bbd4b28a8f50

